PR #4545, opened 2022-02-13
fix(zoe): get ZCF bundlecap from vatAdminService

warner:
Zoe needs access to the vatAdminSvc, to create a new ZCF vat to host each
contract. `makeZoeKit()` provides an optional argument to control which
bundle is used for this purpose. This 4th argument was a string, defaulting
to 'zcf', and relied upon `E(vatAdminSvc).createVatByName` (which is going
away).

To move everything closer to bundlecaps, the `makeZoeKit()` optional argument
now takes `{ name }` or `{ id }` or `{ bundlecap }`, instead of only a name.
Zoe uses the new `E(vatAdminSvc).getNamedBundleCap()` to convert a name into
a bundlecap, but in the future I expect our chain-side bootstrap() to call
`makeZoeKit()` with a specific bundlecap. In the long run, I'm trying to make
room for Zoe to accomodate multiple versions of ZCF, each indexed by its
bundlecap.

Non-swingset-based zoe-using unit tests generally use `fakeVatAdmin.js` to
build a mock `vatAdminSvc`. This commit changes `fakeVatAdmin.js` to include
an additional `zcfBundlecap` export, which is an empty marker object that
serves as a stand-in for the bundlecap. The fake `vatAdminSvc` can accept
this fake `zcfBundlecap`, and will evaluate the ZCF code appropriately.
However the fake `vatAdminSvc` also knows how to convert the default `'zcf'`
name into that bundlecap, so downstream code should not need to change.

refs #4487

PR #4418, opened 2022-01-28
feat: script to configure .vscode

turadg:
closes: #4722

## Description

As evidenced by https://github.com/Agoric/agoric-sdk/issues/4339#issuecomment-1024581849 we can help people set up VS Code. 

We .gitignore the .vscode directory so that everyone can configure everything how they want. This adds a script that creates the config if absent and shows suggested edits if pressent.

Inspired from https://github.com/git/git/commit/30cf1911e2119eda55d9376c9cc2eda893fe6692 by way of https://stackoverflow.com/a/51920012/46040

### Security Considerations

None.

### Documentation Considerations

Should be mentioned in wiki?

### Testing Considerations

None.

PR #4598, opened 2022-02-18
fix(zoe): add zoe.installBundleID

### LINKED ISSUE #4563: add zoe.installBundleID(bundleID)

As part of the path towards contract/vat upgrade, I need Zoe to treat contract source code as `bundlecap` handles, not (large) bundle objects. With #4372 we have the kernel support, but we need to change the way Zoe manages the "installations" table, as well as change the API.

Zoe distinguishes an "installation" (akin to a class) from an "instance". Each installation has a single contract code bundle. Each instance of that installation has its own vat. The vat is launched with the ZCF bundle, and then Zoe tells ZCF to evaluate the contract bundle.

The current API is: `E(zoe).install(bundle)`. I want to change that to `E(zoe).install(bundleID)`, where `bundleID` is a string that includes a hash of the bundle contents. Zoe uses `vatAdminService` to look up this ID and retrieve a `bundlecap`, and then stores just a bundlecap. When Zoe tells ZCF which contract to use, it sends the bundlecap. ZCF is responsible for retrieving the actual source code (using `bundle = D(bundlecap).getBundle()`) so it can do the `importBundle()`. By leaving the contract's identity as a bundlecap until the last possible moment, we can reduce the size of messages, and give us some wiggle room to get Endo more involved in the process (eventually `importBundle(bundlecap)` should be a service provided by the vat environment, and overlap somehow with JS's native dynamic `import` expression).

For that to work, you must install the bundle first, by talking to the kernel. That means the host application needs to be involved. For the chain, that will require a new cosmos transaction type (#4396), which calls `controller.validateAndInstallBundle`. The contract deployment process must change: the deploy script needs a way to sign and send this transaction first, learn the bundleID, then send the ID to `E(zoe).install` instead of the bundle.

I'm currently expecting that this will need a flag day, which is always horrible. If we were ok with `E(zoe).install(bundleOrID)`, we could avoid it, but that will complicate the necessary Zoe changes. 

Unit test code must be changed too. In looking at the Zoe unit tests, a common pattern is:
```js
  const fakeVatAdminSvc = makeFakeVatAdmin(setJig).admin;
  const { zoeService: zoe } = makeZoeKit(fakeVatAdminSvc);
  const bundle = await bundleSource(contractRoot);
  const installation = await E(zoe).install(bundle);
```

that changes to:
```js
  const { admin: fakeVatAdminSvc, vatAdminState } = makeFakeVatAdmin(setJig);
  const { zoeService: zoe } = makeZoeKit(fakeVatAdminSvc);
  const bundle = await bundleSource(contractRoot);
  vatAdminState.installBundle('b1-zcftester', bundle);
  const installation = await E(zoe).install('b1-atomicswap');
```

(the Bundle ID is normally `b1-` followed by a SHA512 hash of the bundle's compartment map, but the fake vatAdminService allows any string, so we use a made-up ID for the contracts used in the test)

warner:
fix(zoe): add zoe.installBundleID

To transition Zoe from full contract bundles to bundlecaps, this adds a new
install API. `E(zoe).install(bundle)` is unchanged, but the new preferred
approach is `E(zoe).installBundleID(bundleID)`. This requires the
corresponding bundle to be installed with the swingset kernel's
vatAdminService, either before or after `installBundleID()` (zoe will wait
forever for the bundle to be installed).

Zoe `Installation` objects retain their `getBundle()` method to accomodate
dapp tests that have not switched to the new approach, but it throws an error
if used on a new bundleID-based installation. A new method named
`E(zoe).getBundleIDFromInstallation(allegedInstallation)` can be used to both
validate the installation and get back the bundleID, but it throws on the old
bundle-based installations.

Internally, the installationStorage.unwrapInstallation now returns either 
`{ bundle, installation }` or `{ bundlecap, bundleID, installation }`. ZCF's
`evaluateContract()` method accepts either a bundlecap or a full bundle.

closes #4563

### DIFF
--- a/packages/zoe/src/contractFacet/zcfZygote.js
+++ b/packages/zoe/src/contractFacet/zcfZygote.js
@@ -2,7 +2,7 @@
 
 import { assert, details as X, makeAssert } from '@agoric/assert';
 import { E } from '@agoric/eventual-send';
-import { Far, Remotable } from '@endo/marshal';
+import { Far, Remotable, passStyleOf } from '@endo/marshal';
 import { AssetKind, AmountMath } from '@agoric/ertp';
 import { makeNotifierKit, observeNotifier } from '@agoric/notifier';
 import { makePromiseKit } from '@endo/promise-kit';
@@ -275,7 +275,6 @@ export const makeZCFZygote = (
     shutdown: completion => {
       E(zoeInstanceAdmin).exitAllSeats(completion);
       dropAllReferences();
-      // @ts-ignore powers is not typed correctly: https://github.com/Agoric/agoric-sdk/issues/3239s
       powers.exitVat(completion);
     },
     shutdownWithFailure,
@@ -340,7 +339,15 @@ export const makeZCFZygote = (
    * @type {ZCFZygote}
    * */
   const zcfZygote = {
-    evaluateContract: bundle => {
+    evaluateContract: bundleOrBundleCap => {
+      let bundle;
+      if (passStyleOf(bundleOrBundleCap) === 'remotable') {
+        const bundleCap = bundleOrBundleCap;
+        // @ts-expect-error vatPowers is not typed correctly: https://github.com/Agoric/agoric-sdk/issues/3239
+        bundle = powers.D(bundleCap).getBundle();
+      } else {
+        bundle = bundleOrBundleCap;
+      }
       contractCode = evalContractBundle(bundle);
       handlePWarning(contractCode);
     },
--- a/packages/zoe/src/zoeService/installationStorage.js
+++ b/packages/zoe/src/zoeService/installationStorage.js
@@ -3,49 +3,85 @@
 import { assert, details as X } from '@agoric/assert';
 import { Far } from '@endo/marshal';
 import { E } from '@agoric/eventual-send';
+import { makeWeakStore } from '@agoric/store';
+
+/** @typedef { import('@agoric/swingset-vat').BundleID} BundleID */
 
 /**
- *
+ * @param {GetBundleCapForID} getBundleCapForID
  */
-export const makeInstallationStorage = () => {
-  /** @type {WeakSet<Installation>} */
-  const installations = new WeakSet();
+export const makeInstallationStorage = getBundleCapForID => {
+  /** @type {WeakStore<Installation, { bundleCap: BundleCap, bundleID: BundleID }>} */
+  const installationsBundleCap = makeWeakStore('installationsBundleCap');
+  /** @type {WeakStore<Installation, SourceBundle>} */
+  const installationsBundle = makeWeakStore('installationsBundle');
 
   /**
-   * Create an installation by permanently storing the bundle. The code is
-   * currently evaluated each time it is used to make a new instance of a
-   * contract. When SwingSet supports zygotes, the code will be evaluated once
-   * when creating a zcfZygote, then the start() function will be called each
-   * time an instance is started.
+   * Create an installation from a bundle ID or a full bundle. If we are
+   * given a bundle ID, wait for the corresponding code bundle to be received
+   * by the swingset kernel, then store its bundlecap. The code is currently
+   * evaluated each time it is used to make a new instance of a contract.
+   * When SwingSet supports zygotes, the code will be evaluated once when
+   * creating a zcfZygote, then the start() function will be called each time
+   * an instance is started.
    */
-  /** @type {Install} */
-  const install = async bundle => {
-    assert.typeof(bundle, 'object', X`a bundle must be provided`);
+
+  /** @type {InstallBundle} */
+  const installBundle = async bundle => {
+    assert.typeof(bundle, 'object', 'a bundle must be provided');
     /** @type {Installation} */
     const installation = Far('Installation', {
       getBundle: () => bundle,
     });
-    installations.add(installation);
+    installationsBundle.init(installation, bundle);
     return installation;
   };
 
-  const assertInstallation = installation =>
-    assert(
-      installations.has(installation),
-      X`${installation} was not a valid installation`,
-    );
+  /** @type {InstallBundleID} */
+  const installBundleID = async bundleID => {
+    assert.typeof(bundleID, 'string', `a bundle ID must be provided`);
+    // this waits until someone tells the host application to store the
+    // bundle into the kernel, with controller.validateAndInstallBundle()
+    const bundleCap = await getBundleCapForID(bundleID);
+    // AWAIT
+
+    /** @type {Installation} */
+    const installation = Far('Installation', {
+      getBundle: () => {
+        throw Error('bundleID-based Installation');
+      },
+    });
+    installationsBundleCap.init(installation, { bundleCap, bundleID });
+    return installation;
+  };
 
   /** @type {UnwrapInstallation} */
   const unwrapInstallation = installationP => {
     return E.when(installationP, installation => {
-      assertInstallation(installation);
-      const bundle = installation.getBundle();
-      return { bundle, installation };
+      if (installationsBundleCap.has(installation)) {
+        const { bundleCap, bundleID } =
+          installationsBundleCap.get(installation);
+        return { bundleCap, bundleID, installation };
+      } else if (installationsBundle.has(installation)) {
+        const bundle = installationsBundle.get(installation);
+        return { bundle, installation };
+      } else {
+        assert.fail(X`${installation} was not a valid installation`);
+      }
     });
   };
 
+  const getBundleIDFromInstallation = async allegedInstallationP => {
+    const { bundleID } = await unwrapInstallation(allegedInstallationP);
+    // AWAIT
+    assert(bundleID, 'installation does not have a bundle ID');
+    return bundleID;
+  };
+
   return harden({
-    install,
+    installBundle,
+    installBundleID,
     unwrapInstallation,
+    getBundleIDFromInstallation,
   });
 };
--- a/packages/zoe/src/zoeService/internal-types.js
+++ b/packages/zoe/src/zoeService/internal-types.js
@@ -56,12 +56,14 @@
 /**
  * @callback UnwrapInstallation
  *
- * Assert the installation is known, and return the bundle and
+ * Assert the installation is known, and return the bundle/bundlecap and
  * installation
  *
  * @param {ERef<Installation>} installationP
  * @returns {Promise<{
- *   bundle: SourceBundle,
+ *   bundle?: SourceBundle,
+ *   bundleCap?: BundleCap,
+ *   bundleID?: BundleID,
  *   installation:Installation
  * }>}
  */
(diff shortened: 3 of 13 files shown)


warner:
This changes a handful of Zoe tests (to confirm that the new behavior works). It does not yet change the rest of the zoe tests, nor the tests elsewhere in agoric-sdk, nor the tests in dapps that live outside agoric-sdk. I need to change all of those before I can remove the old `E(zoe).install` API.

I also made a small swingset change to export a few types that are now part of the Zoe API (`BundleID` and a module type). This seems necessary to let `packages/vats/` pass typechecking.

packages/SwingSet/src/index.js
```diff
@@ -12,3 +12,6 @@ export { buildTimer } from './devices/timer.js';
 export { buildBridge } from './devices/bridge.js';
 export { default as buildCommand } from './devices/command.js';
 export { buildPlugin } from './devices/plugin.js';
+
+/** @typedef { import('./types').BundleID } BundleID */
```
mhofman: I'm gonna be stubborn, but I'd still prefer using extensions.
```suggestion
/** @typedef { import('./types.js').BundleID } BundleID */
```

packages/zoe/src/contractFacet/zcfZygote.js
```diff
@@ -340,7 +340,15 @@ export const makeZCFZygote = (
    * @type {ZCFZygote}
    * */
   const zcfZygote = {
-    evaluateContract: bundle => {
+    evaluateContract: bundleOrBundlecap => {
+      let bundle;
+      if (passStyleOf(bundleOrBundlecap) === 'remotable') {
+        const bundlecap = bundleOrBundlecap;
+        // @ts-ignore powers is not typed correctly: https://github.com/Agoric/agoric-sdk/issues/3239s
```
erights: @turadg recommends that we prefer `@ts-expect-error` when that works as well. That way, if it stops being necessary, it'll let us know. Also, typo
```suggestion
        // @ts-ignore powers is not typed correctly: https://github.com/Agoric/agoric-sdk/issues/3239
```

packages/zoe/src/zoeService/installationStorage.js
```diff
@@ -3,49 +3,82 @@
...
-    assert(
-      installations.has(installation),
-      X`${installation} was not a valid installation`,
-    );
+  /** @type {InstallBundleID} */
+  const installBundleID = async bundleID => {
+    assert.typeof(bundleID, 'string', `a bundle ID must be provided`);
+    // this waits until someone tells the host application to store the
+    // bundle into the kernel, with controller.validateAndInstallBundle()
+    const bundlecap = await getBundlecapForID(bundleID);
```
erights: I will here admit to an irrational bias, rather than requesting a change.

The idea of a long-term-by-design postponement being done with an `await`, rather than an explicit `E.when` makes me nervous. The callbacks of an `E.when` are distinct js function objects. The implicit callback of an `await` is the unreified continuation of the function. However, I don't see a concrete disadvantage of the long term `await`.
   * both are captured by snapshot with full fidelity.
   * both are reproduced by replay with full fidelity. Therefore
      * both are equally supportive of manchurian mischief
      * including upgrading to a new version of XS that can't read old snapshots
   * neither will ever need to be virtual, because promises can't
   * neither will ever need to survive a normal upgrade, because promises don't.

So, no change suggested.

erights: But please at least put the AWAIT comment after this line.

erights (approved):
Why `*Bundlecap` instead of `*BundleCap`? Since none of this has been merged yet, if we're going to rename it, now is the ideal time.

packages/zoe/src/contractFacet/zcfZygote.js
turadg: 👍, modulo that as yet [sometimes the IDE and CLI/CI will disagree about the set of errors](https://github.com/Agoric/agoric-sdk/issues/4620). In those cases it's better to use ts-ignore than to have either of them complain. After that issue is resolved we could global replace ts-ignore with ts-expect-error and see which are gone.

In this particular case I think we're better off fixing the types: 1b043d284c13aa3679db05d6f1be42b43a2bafc3 I didn't push directly to this branch because I didn't know the norms here around that and I wasn't certain I got it right.

packages/SwingSet/src/index.js
turadg: Note that when VS Code offers to import, it does so without the `.js`, even with the `typescript.preferences.importModuleSpecifierEnding": "js"` setting.

mhofman: Interesting, I swear it does for me. We should find what setting triggers this.

turadg: Oh good! We shouldn't have to decide based on limits of tooling. Once we have a setting we agree on we can distribute with https://github.com/Agoric/agoric-sdk/pull/4418

mhofman: Well apparently I must have muscle memory and add those automatically as I can't seem to reproduce anymore.

packages/zoe/src/zoeService/installationStorage.js
warner: I'm happy to use `.then` here instead. Does your bias sort of feel that `await` is for things that will happen soon, whereas `.then` is for things that happen much later?

I don't think either form is better than the other w.r.t. upgrade or snapshots. Having a name for the second half helps to clarify its purpose, sometimes, but mostly when there might be another code path that leads into it (as in your comments in #4545 and the three different ways to obtain a promise for the ZCF bundle).

One style question: making both `installBundle` and `installBundleID` be `async` seems like a good idea to emphasize their parallelism (also they're both mostly invoked from remote vats, `E(zoe).install*`, so it's not like the caller gets a choice). Once `installBundleID` is `async`, then it seemed awkward to build up the `.then` and return a Promise instead of using `await` inline. If I change the `await` to `.then`, should I also change `installBundleID` to not be `async`? And then should I change `installBundle` to also not be async, and be ok with `installBundle` being actually-sync (and not returning a Promise) but `installBundleID` always returns a Promise? And, now they have different error behaviors if you e.g. pass in a bad argument (synchronous throw vs promise rejection).

packages/SwingSet/src/index.js
warner: Oh, I thought I got red squiggles from vscode when I did that. I prefer extensions too, I'll look into it.

packages/zoe/src/zoeService/installationStorage.js
erights: Please do *not* use `.then`. It is not safe against reentrancy. And least `await` is safe, and so switching from `await` to `.then` is a decrease in robustness. When you would use `.then`, use `E.when` instead.

If `await` and `E.when` are otherwise tied, a tie breaker is that `E.when` gives us deep stacks on errors. We cannot reasonably make that happen for `.then` or `await` without changes to the underlying engine.

erights: I do not consider it awkward for an `async` function to use explicit promise logic rather than `await`. In fact, I think we should do exactly this more often. In the absence of `await`, as you note, the meaning of `async` is to promise-ify the outcome, turning both success and errors into promises.

I will admit that this particular combination is one that other developers may find surprising when they first see it. In this case, I don't think that's a big hurdle.

packages/zoe/src/contractFacet/zcfZygote.js
erights: >  I didn't push directly to this branch because I didn't know the norms here around that and I wasn't certain I got it right.

I defer to @warner on such norms. But my take:

Don't push directly to someone else's branch. Create a PR based on their's, send it to them to review, and mention that PR in a review comment on their original PR. If they like it, they should feel free to merge it into their PR without further consulting you.

warner: yeah, either make a meta-PR, or use github's "suggested change" mechanism.. pushing to my branch would surprise/confuse my local workspace and/or workflow

turadg: 👌 no surprise pushing to PR branches

for something that's too small to make a new PR but too spread out to make a bunch of suggested changes, WDYT of sharing the commit as above? 1b043d284c13aa3679db05d6f1be42b43a2bafc3  Then you can just `git cherry-pick 1b043d2` if you want it.

packages/zoe/src/zoeService/installationStorage.js
warner: Note to folks following along at home (because I had to ask too): the concern with `.then` is a hostile provider of the object that you assume behaves like a Promise. It might be a non-Promise with a `.then` method that does something synchronously. It might be a real Promise but with an overriden `.then` method. Using `await` and `E.when` protects against this malice.

In this case, `getBundleCapForID` is coming from the rest of Zoe, but I'm happy to stick with something that maintains caution in any situation.

I'll stick with `await`, add an `// AWAIT` comment, and leave both functions as `async`.

packages/zoe/src/contractFacet/zcfZygote.js
turadg: @warner I notice TS failures in https://github.com/Agoric/agoric-sdk/runs/5300164567?check_suite_focus=true

If that's due to https://github.com/Agoric/agoric-sdk/commit/1b043d284c13aa3679db05d6f1be42b43a2bafc3 no objections here to punting that.

packages/zoe/src/zoeService/internal-types.js
```diff
@@ -56,12 +56,14 @@
...
- * Assert the installation is known, and return the bundle and
+ * Assert the installation is known, and return the bundle/bundlecap and
  * installation
  *
  * @param {ERef<Installation>} installationP
  * @returns {Promise<{
- *   bundle: SourceBundle,
+ *   bundle?: SourceBundle,
+ *   bundleCap?: BundleCap,
+ *   bundleID?: BundleID,
```
turadg, line 66: non-blocking but FYI, we can use unions to convey mutual exclusion. It has the advantage over optionals of guaranteeing to consumer that one of them is defined. E.g.,
```ts
type BundleOrBundleCap = {bundle: SourceBundle} | {bundleCap: BundleCap, bundleId: BundleID};
```
or in jsdoc,
```js
/** @typedef  { {bundle: SourceBundle} | {bundleCap: BundleCap: bundleId} } BundleOrBundleCap */
```

then intersection for the return type
```ts
Promise<BundleOrBundleCap & {installation: Installation}>
```

warner: aye, thanks, will look into that later, there's another `BundleOrBundleCap` floating around which doesn't include the `bundleID` that I don't want to collide with

turadg: > collide with 

Want to point out that the possibility of collisions is due to ambient types. Because TypeScript is structurally typed, there's usually little problem with redundant definitions. (Besides code maintenance/documentation.)

PR #4519, opened 2022-02-10
Allow AMM liquidity provision at arbitrary rate

Chris-Hibbert:
closes: #4249

## Description

Add a new invitation that allows adding liquidity in an arbitrary ratio.

The virtual pools get an internal facet that allows access to
addLiquidityActual.

In order to allow the rebalancing trade and liquidity addition to be
done in a single transaction, pool methods deal with stagedAllocations
rather than actual allocation.

### Security Considerations

Not particularly security relevant.  The AMM is important, and governance them crucial, but this doesn't affect that much.

### Documentation Considerations

There's a new API `makeAddLiquidityAtRateInvitation`, which should be added to the documentation.

### Testing Considerations

Includes tests for the new functionality. Covers initial liquidity, and rising and falling liquidity ratio.

PR #4600, opened 2022-02-18
dropping the max on the property-based tests led to problems

Chris-Hibbert:
refs: #4519

## Description

As a last step before checkin, I dropped the maximums in the property based tests. This led to failures in CI (https://github.com/Agoric/agoric-sdk/runs/5251436118?check_suite_focus=true).

### Security Considerations

We should investigate whether the CI failures represent actual issues. I suspect it's just that the approximation for 'close enough' was insufficient when the numbers being added to the pool got absurdly large.

No issues.

### Testing Considerations

fixing a test so CI isn't impacted.  We can get this fix in quickly, then we can consider at our leisure whether we need to check more cases.


dckc (approved):
ok; yeah; let's unblock this.

PR #4413, opened 2022-01-28
Import explicit types from `@endo/marshal`

michaelfig:
closes: #XXXX
refs: #XXXX

## Description

Informational for @kriskowal (and @mhofman if interested).  This is the other side of endojs/endo#1025 , which must be merged before this work can continue.

PR #4604, opened 2022-02-18
chore: Sync Endo versions

mhofman:
closes: #4413 

## Description

Syncs the versions of endo packages to the latest published versions.
Removes outdated patch files for endo packages
Includes the changes from https://github.com/Agoric/agoric-sdk/pull/4413 (minus patch files) to bring in compatibility with explicit types for `@endo/marshal`
Remove extraneous eslint global declarations

### Testing Considerations

This change is depends on concurrent fixes which finish dapp migrations from -r esm to Node.js native ESM.

- https://github.com/Agoric/documentation/pull/622
- https://github.com/Agoric/dapp-otc/pull/31


packages/xsnap/src/xsnap.js
```diff
@@ -1,5 +1,5 @@
 /* global process */
-// @ts-check
+// @ts-nocheck
```
mhofman: This is the change we need to revert

michaelfig (approved):
LGTM!

mhofman:
I'd like to do a little yarn surgery tomorrow morning to avoid duplicate babel packages (which may be a reason dapp broke?)

kriskowal:
> I'd like to do a little yarn surgery tomorrow morning to avoid duplicate babel packages (which may be a reason dapp broke?)

This almost certainly contributed, as it would explain why I had to explicitly pre-initialize `@agoric/babel-standalone` in a couple places. It doesn’t account for the problem with `@babel/types` not having a default export when run under `-r esm`, which is a dependency we brought in via `@endo/static-module-record` depending on `recast` depending on `@babel/types`.

PR #4607, opened 2022-02-18
Refresh parameter test on amm liquidity

### LINKED ISSUE #4601: Property-based testing of AMM calc failed in CI
## Describe the bug

In CI, @warner saw this  https://github.com/Agoric/agoric-sdk/runs/5251436118?check_suite_focus=true
It's the result of a new property-based test checked in with #4519.

At the last minute, I removed maximum limit on some values. #4600 reinstated the caps. This issue is for investigating whether the failure represents an actual issue in the code.

## Expected behavior

Property based testing uses random inputs. The version run in CI should probably use a stable seed, but only after sufficient runs to be sure there aren't uncovered corner cases.

## Platform Environment

This ran in CI. Other than reproducing the random seed, it's reproducible locally.

## Screenshots

Screenshot 2022-02-18 at 09-51-11 merge bundle_vat-admin devices add waitForBundlecap() · Agoric_agoric-sdk@9f531c5

Chris-Hibbert:
closes: #4601
 
## Description

#4519  added a test with some failures. #4600 temporarily repaired it. This is a better fix that increases the range of cases we test, with a narrow exclusion for ill-behaved parameters.

### Security Considerations

No actual security issues. The economic security issue is that someone might use the new ability to add liquidity to the AMM in an infelicitous circumstance (the pools have a small amount of collateral to start) and end up giving more money to the arbitrageurs than they intended.

When we explain `makeAddLiquidityAtRateInvitation`, we need to explain where the edge cases are.

### Testing Considerations

This PR repairs a property-based test that by its nature doesn't run with the same values every time, so it can give different results on different runs. The exclusions mentioned above are intended to skip over values with problematic results.

### DIFF
--- a/packages/run-protocol/test/amm/vpool-xyk-amm/test-amm-calc.js
+++ b/packages/run-protocol/test/amm/vpool-xyk-amm/test-amm-calc.js
@@ -14,16 +14,23 @@ const { brand: brandX } = makeIssuerKit('X');
 const { brand: brandY } = makeIssuerKit('Y');
 
 const oneB = 1_000_000_000n;
-// const dec18 = 1_000_000_000_000_000_000n; // 18 decimals as used in ETH
+const dec18 = 1_000_000_000_000_000_000n; // 18 decimals as used in ETH
 
+// The starting balances in the pools will be between 10^4 and  10^27
 const arbPoolX = fc
-  .bigUint({ max: oneB })
-  .map(value => m.make(brandX, 100n + value));
+  .bigUint({ max: oneB * dec18 })
+  .map(value => m.make(brandX, 10_000n + value));
 const arbPoolY = fc
-  .bigUint({ max: oneB })
-  .map(value => m.make(brandY, 100n + value));
-const arbGiveX = fc.bigUint({ max: oneB }).map(value => m.make(brandX, value));
-const arbGiveY = fc.bigUint({ max: oneB }).map(value => m.make(brandY, value));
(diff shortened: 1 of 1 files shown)


packages/run-protocol/test/amm/vpool-xyk-amm/test-amm-calc.js
```diff
@@ -33,6 +47,13 @@ test('balancesToReachRatio calculations are to spec', t => {
         giveY: arbGiveY,
       }),
       ({ poolX, poolY, giveX, giveY }) => {
+        const kBefore = Number(multiply(poolX.value, poolY.value));
+        const poolRatioAfter = Math.max(
+          Number(giveX.value + poolX.value) / Number(giveY.value + poolY.value),
+          Number(giveY.value + poolY.value) / Number(giveX.value + poolX.value),
+        );
+        // skip cases where the target ratio is more than K at the start
+        fc.pre(poolRatioAfter < kBefore);
```
dckc: does this `fc.pre` check subsume the `imbalancedRequest(` check below?

dckc (approved):
looks good!

packages/run-protocol/test/amm/vpool-xyk-amm/test-amm-calc.js
Chris-Hibbert: It's exactly the same calculation.  thanks.

PR #4609, opened 2022-02-19
4602 simpler vault uinotify

### LINKED ISSUE #4602: updateUIState shouldn't be async
@Chris-Hibbert commented on [Fri Mar 05 2021](https://github.com/Agoric/dapp-token-economy/issues/123)

It appears that all the callers have recently checked the collateralization ratio, so `updateUiState()` (in `vault.js`) shouldn't need to await the result of a fresh call. Find a way to store something like `recentCollateralizationRatio` so we can re-use it.

Some of the callers will know that the thing that has changed is the level of collateral rather than the price level, so perhaps they can short cut a call to the priceAuthority.

---

@dtribble commented on [Fri Feb 18 2022](https://github.com/Agoric/dapp-token-economy/issues/123#issuecomment-1045098286)

This will be addressed by removing the computation of the collateralization ratio from the notifier.

dtribble:
closes: #4602
refs: #XXXX

## Description

Delete the function to compute `collateralizationRatio`, remove from the state that gets notified, and fix up tests.

### Security Considerations
None

Update the record definition if it's in the docs.

Vault UI may need to be updated to compute the property.

### Testing Considerations

Deleted most test, and in a few cases checked the notification's debt to confirm that the action in the test had been taken.

### DIFF
--- a/packages/ERTP/src/amountMath.js
+++ b/packages/ERTP/src/amountMath.js
@@ -252,10 +252,12 @@ const AmountMath = {
    *
    * @template {AssetKind} K
    * @param {Brand} brand
-   * @param {AssetKind=} assetKind
+   * @param {K=} [assetKind='nat']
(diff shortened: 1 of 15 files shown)


dtribble:
Because git hates me

PR #4611, opened 2022-02-19
Basic OpenTelemetry tracing support

### LINKED ISSUE #4556: OpenTelemetry for slog files and loadgen

Slog file and loadgen analysis should be more standard and reusable by the whole team.

## Description of the Design

Integrate OpenTelemetry-based exporters with loadgen and slog file ingestion.

michaelfig:
closes: #4556

## Description

This adds basic `SLOGSENDER` support for distributed tracing via the `@opentelemetry` family of packages.  It works as it is, but even with missing features, it's a jumping-off point for future improvements.

Though it is missing expository documentation, in order to prevent folks from stubbing their toes, this support is only activated by setting the `SOLO_SLOGSENDER` or `SLOGSENDER` environment variables appropriately.

I mainly expect a reviewer to confirm that this feature does not interfere with normal usage of the SDK, only with explicit opt-in.

### Security Considerations

The code makes an outbound TCP connection if enabled, otherwise no change in security posture.

An end-to-end tracing walkthrough needs to be recorded and documented.

### Testing Considerations

I expect to enable this for our daily loadgen run, so we can start collecting trace data from it.

### DIFF
--- a/packages/solo/src/start.js
+++ b/packages/solo/src/start.js
@@ -99,6 +99,7 @@ const neverStop = () => {
   });
 };
 
+const TELEMETRY_SERVICE_NAME = 'solo';
 const buildSwingset = async (
   kernelStateDBDir,
   mailboxStateFile,
@@ -165,15 +166,14 @@ const buildSwingset = async (
     plugin: { ...plugin.endowments },
   };
 
-  const {
-    SOLO_OTEL_EXPORTER_PROMETHEUS_PORT,
-    SOLO_OTEL_RESOURCE_ATTRIBUTES,
-    ...restEnv
-  } = process.env;
+  const soloEnv = Object.fromEntries(
+    Object.entries(process.env)
+      .filter(([k]) => k.match(/^SOLO_/)) // narrow to SOLO_ prefixes.
+      .map(([k, v]) => [k.replace(/^SOLO_/, ''), v]), // Replace SOLO_ controls with chain version.
+  );
   const env = {
-    ...restEnv,
-    OTEL_EXPORTER_PROMETHEUS_PORT: SOLO_OTEL_EXPORTER_PROMETHEUS_PORT,
-    OTEL_RESOURCE_ATTRIBUTES: SOLO_OTEL_RESOURCE_ATTRIBUTES,
+    ...process.env,
+    ...soloEnv,
   };
   const { metricsProvider = DEFAULT_METER_PROVIDER } = getTelemetryProviders({
     console,
@@ -181,13 +181,8 @@ const buildSwingset = async (
     serviceName: 'solo',
   });
 
-  const {
-    SOLO_SLOGFILE: slogFile,
-    SOLO_SLOGSENDER,
-    SOLO_LMDB_MAP_SIZE,
-  } = process.env;
-  const mapSize =
-    (SOLO_LMDB_MAP_SIZE && parseInt(SOLO_LMDB_MAP_SIZE, 10)) || undefined;
+  const { SLOGFILE: slogFile, SLOGSENDER, LMDB_MAP_SIZE } = env;
+  const mapSize = (LMDB_MAP_SIZE && parseInt(LMDB_MAP_SIZE, 10)) || undefined;
   const { kvStore, streamStore, snapStore, commit } = openSwingStore(
     kernelStateDBDir,
     { mapSize },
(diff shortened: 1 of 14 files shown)


packages/solo/src/start.js
```diff
@@ -165,29 +166,19 @@ const buildSwingset = async (
...
-  } = process.env;
-  const env = {
-    ...restEnv,
-    OTEL_EXPORTER_PROMETHEUS_PORT: SOLO_OTEL_EXPORTER_PROMETHEUS_PORT,
-    OTEL_RESOURCE_ATTRIBUTES: SOLO_OTEL_RESOURCE_ATTRIBUTES,
-  };
+  const env = Object.fromEntries(
+    Object.entries(process.env)
+      .filter(([k]) => !k.match(/^(OTEL_|SLOG)/)) // Remove chain controls.
+      .map(([k, v]) => [k.replace(/^SOLO_/, ''), v]), // Replace SOLO_ controls with chain version.
```
warner, line 172: Does that `replace` mean that `SOLO_LMDB_MAP_SIZE` and `LMDB_MAP_SIZE` would get mapped to the same thing, and one would randomly replace the other? I think we run both chain and solo from the same "start a test environment" command that their env vars should be disjoint but also tolerate each other's presence without confusion.

I'm wondering if this should start with a `match(/^SOLO_`, but I don't know what else `env` is used for in this app.

packages/telemetry/bin/ingest-slog
```diff
@@ -0,0 +1 @@
+../src/ingest-slog-entrypoint.js
```
warner, line 1: Is this a symlink? (I think this is how github renders a symlink). Those make me nervous in git repos, mostly for cross-platform compatibility (windows) but also because I think they're just confusing in version control. Is there an easy way to make this not a symlink? If not, use the symlink.

packages/telemetry/src/all.js
```diff
@@ -0,0 +1,24 @@
...
+    makeOtelSlogSender(opts),
+  ]);
+
+  const filteredSenders = senders.filter(Boolean);
+  if (!filteredSenders.length) {
+    return undefined;
+  }
+  const slogSender = (...args) => {
+    for (const sender of filteredSenders) {
+      sender?.(...args);
```
warner, line 17: ooh, we can do `?.` now?

packages/telemetry/src/slog-to-otel.js
```diff
@@ -0,0 +1,640 @@
...
+          }
+          case 'resolve':
+          case 'subscribe':
+          case 'vatstoreGet':
+          case 'vatstoreSet':
+          case 'vatstoreDelete':
+          case 'dropImports':
+          case 'retireImports':
+          case 'retireExports': {
+            // TODO: Maybe too noisy and mostly irrelevant?
```
warner, line 507: `resolve` seems important, but yeah we can probably ignore the rest for now. We need to remember that they were skipped when we're analyzing, though, else we're going to be concerned about all the GC that isn't happening, and confused about how the vat is managing virtual objects without vatstore calls.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -0,0 +1,640 @@
...
+          const cause = crankNumToCause.get(slogAttrs.crankNum);
+          ({ attrs, error } = cause);
+          crankNumToCause.delete(slogAttrs.crankNum);
+        }
+        spans.end(getCrankKey(), attrs, error);
+        break;
+      }
+      case 'console': {
+        // We don't care about console messages.  They are out of consensus and
+        // can be really huge.
```
warner, line 620: true, although they may be important to debugging things, if/when we think about telemetry as a debugging tool more than a performance-measuring tool

warner (approved):
seems good to me, modulo the env name question

packages/solo/src/start.js
michaelfig: Thanks for pointing this out.  I've now separated the `SOLO_`-prefixed variables and used them to override `env`.  `process.env` is used directly by the chain, so now at least it's predictable what will happen, and shouldn't cause conflicts (`SOLO_${var}` is used to override `${var}` only in the ag-solo).

packages/telemetry/bin/ingest-slog
michaelfig: Yeah, it's a choice between writing a shell script wrapper or using a symlink.  I used to choose the wrapper, but found later (from @kriskowal's example) that the symlink is more portable to windows.

packages/telemetry/src/all.js
michaelfig: Yippers!

PR #4612, opened 2022-02-19
Prometheus update for loadgen telemetry

michaelfig:
refs: #4556

## Description

This upgrades to the latest `@opentelemetry` packages, including updating the Prometheus support.  It also fixes some minor issues that interfered with the loadgen's support for metrics.


packages/solo/src/start.js
```diff
@@ -144,6 +151,18 @@ const buildSwingset = async (
     plugin: { ...plugin.endowments },
   };
 
+  const env = {
+    ...process.env,
+    OTEL_EXPORTER_PROMETHEUS_PORT:
+      process.env.SOLO_OTEL_EXPORTER_PROMETHEUS_PORT,
+    OTEL_RESOURCE_ATTRIBUTES: process.env.SOLO_OTEL_RESOURCE_ATTRIBUTES,
+  };
```
mhofman: Shouldn't we also remove the `SOLO_` envs as well?

For example:

```suggestion
  const {
    SOLO_OTEL_EXPORTER_PROMETHEUS_PORT,
    SOLO_OTEL_RESOURCE_ATTRIBUTES,
    ...restEnv
  } = process.env;
  const env = {
    ...restEnv,
    OTEL_EXPORTER_PROMETHEUS_PORT: SOLO_OTEL_EXPORTER_PROMETHEUS_PORT,
    OTEL_RESOURCE_ATTRIBUTES: SOLO_OTEL_RESOURCE_ATTRIBUTES,
  };
```

packages/solo/src/start.js
```diff
@@ -161,6 +180,12 @@ const buildSwingset = async (
     snapStore,
   };
 
+  // Not to be confused with the gas model, this meter is for OpenTelemetry.
+  const metricMeter = metricsProvider.getMeter('ag-solo');
+  const slogCallbacks = makeSlogCallbacks({
+    metricMeter,
+  });
```
mhofman, line 205: `labels` is not marked as optional, why is this not causing a type error?

packages/agoric-cli/src/start.js
```diff
@@ -393,6 +393,11 @@ export default async function startMain(progname, rawArgs, powers, opts) {
       create(hashFile, gci),
     ]);
 
+    if (!popts.restart) {
+      // Don't actually run the chain.
+      return 0;
+    }
```
mhofman, line 399: 🎉

packages/cosmic-swingset/src/kernel-stats.js
```diff
@@ -56,10 +63,10 @@ const recordToKey = record =>
 /**
  * @param {{
  *   metricMeter: import('@opentelemetry/sdk-metrics-base').Meter,
- *   labels: Record<string, string>
+ *   attributes: import('@opentelemetry/api-metrics').Attributes,
```
mhofman: Looks like [this type](https://github.com/open-telemetry/opentelemetry-js/blob/f59c5b268bd60778d7a0d185a6044688f9e3dd51/experimental/packages/opentelemetry-api-metrics/src/types/Metric.ts#L108) is no more optional than `Record`.

packages/solo/src/start.js
```diff
@@ -88,6 +89,14 @@ const atomicReplaceFile = async (filename, contents) => {
   }
 };
 
+const neverStop = () => {
+  return harden({
+    vatCreated: () => true,
+    crankComplete: () => true,
+    crankFailed: () => true,
+  });
```
mhofman, line 99: I added an `emptyCrank` to the policy since this PR was started

mhofman (changes requested):
I have a some confusion why type checks are not bringing up certain issues, such as an apparently non-optional `attributes` (formerly `labels`) missing param, or the lack of `emptyCrank` on the `neverStop` policy.

The blocker here is the latter as if it gets used, it'll crash.

mhofman:
At least for the `emptyCrank`, the missing type error might be due to an out of date base commit.

packages/solo/src/start.js
michaelfig: `// @ts-check` was not yet enabled in this module.  I've done that now.

packages/solo/src/start.js
michaelfig: Thanks for the heads-up.  The run policy type doesn't appear to be exposed outside SwingSet.

michaelfig: Oh, found it... ambient `RunPolicy`.

mhofman: Yeah it's currently ambient :(

packages/cosmic-swingset/src/kernel-stats.js
```diff
@@ -8,6 +8,8 @@ import {
   KERNEL_STATS_UPDOWN_METRICS,
 } from '@agoric/swingset-vat/src/kernel/metrics.js';
 
+import '@agoric/swingset-vat';
```
mhofman: Interesting. I think swingset follows the `exported.js` pattern. It might be a side effect of the recent hybrid ambient / exported types approach that this actually even works.

```suggestion
import '@agoric/swingset-vat/exported.js';
```

mhofman (approved):
Thanks for addressing those.

packages/cosmic-swingset/src/kernel-stats.js
michaelfig: > Interesting. I think swingset follows the `exported.js` pattern. It might be a side effect of the recent hybrid ambient / exported types approach that this actually even works.

Argh.  There's no satisfactory way to get the type of the `RunPolicy` unless I shuffle SwingSet types around.  I think I'll leave it as `any`.

`exported.js` currently only pulls in the timer vat's types, and even just that results in:

```
snow:cosmic-swingset michael$ yarn lint
yarn run v1.22.17
$ run-s --continue-on-error lint:*
$ tsc --maxNodeModuleJsDepth 3 -p jsconfig.json
../SwingSet/src/vats/timer/types.js:13:23 - error TS2304: Cannot find name 'ERef'.

13  *             waker: ERef<TimerWaker>
                         ~~~~

../SwingSet/src/vats/timer/types.js:17:23 - error TS2304: Cannot find name 'ERef'.

17  * @property {(waker: ERef<TimerWaker>) => Array<Timestamp>} removeWakeup
                         ~~~~

../SwingSet/src/vats/timer/types.js:35:9 - error TS2304: Cannot find name 'Notifier'.

35  * ) => Notifier<Timestamp>} makeNotifier
           ~~~~~~~~

../SwingSet/src/vats/timer/types.js:66:23 - error TS2304: Cannot find name 'ERef'.

66  * @property {(waker: ERef<TimerWaker>) => Timestamp} schedule
                         ~~~~


Found 4 errors.
```

mhofman: Ugh ambient types need to go!

PR #4613, opened 2022-02-19
Notify broken master tests via e-mail

### LINKED ISSUE #4459: Prevent or notify broken master tests

We recently had some build check failures for master:

> https://github.com/Agoric/agoric-sdk/runs/5069159963?check_suite_focus=true
> ```
> lerna ERR! EPUBLISHCONFLICT Cannot publish @agoric/cosmos@undefined over existing version.
> ```
>
> https://github.com/Agoric/agoric-sdk/runs/5069182650?check_suite_focus=true
> ```
> error @typescript-eslint/parser@5.10.2: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.15.5"
> error Found incompatible module.
> ```

They persisted for some days because there was no notification. The checks that were failing were defined for some purpose, so we should know when they fail (or if it doesn't matter then remove them).

## Description of the Design

TBD, some way to prevent persistent failures of master checks.

## Security Considerations
--
## Test Plan

TBD

michaelfig:
closes: #4459

## Description

Update CI to avoid redundant checks on push to master.  Also, send e-mail to a private list when a push to master CI check fails.

Our Github branch protection rules ensure that both CI passes and the merged branch is up-to-date:

image

### Security Considerations

Our e-mail sending depends upon a SendGrid API key.  It is controlled by @warner and @michaelfig.

### DIFF
--- a/.github/workflows/after-merge.yml
+++ b/.github/workflows/after-merge.yml
@@ -1,0 +1,157 @@
+# Run after merge to trunk
+on:
+  push:
+    branches: [master] # $default-branch
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: ['14.x', '16.x']
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: 'true'
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: cache node modules
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/yarn
+          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      # 'yarn install' must be done at the top level, to build all the
+      # cross-package symlinks
+      - name: yarn install
+        run: yarn install
+      - name: check if git is dirty
+        run: git diff --stat --exit-code
+      - name: check dependencies
+        run: yarn check-dependencies
+      # 'yarn build' loops over all workspaces
+      - name: yarn build
+        run: yarn build
+
+      - uses: ./.github/actions/save-node
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: notify on failure
+        if: failure()
+        uses: ./.github/actions/notify-status
+        with:
+          from: ${{ secrets.NOTIFY_EMAIL_FROM }}
+          to: ${{ secrets.NOTIFY_EMAIL_TO }}
+          password: ${{ secrets.NOTIFY_EMAIL_PASSWORD }}
+
+  dev-canary:
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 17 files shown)


.github/workflows/after-merge.yml
```diff
@@ -0,0 +1,157 @@
...
+          to: ${{ secrets.NOTIFY_EMAIL_TO }}
+          password: ${{ secrets.NOTIFY_EMAIL_PASSWORD }}
+
+  coverage:
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # note: only use one node-version
+        node-version: ['14.x']
```
turadg, line 69: with only one node version, why use the matrix strategy?

.github/workflows/after-merge.yml
```diff
@@ -0,0 +1,157 @@
+# Run after merge to trunk
```
turadg, line 1: I think this only works as long as *Require branches to be up to date before merging* is enabled in the branch protection rules. (Without that someone could merge to master a state that wasn't tested in the PR.)

Would be good to note that requirement.

turadg (approved):
👏

.github/workflows/after-merge.yml
```diff
@@ -0,0 +1,157 @@
...
+            ${{ runner.os }}-yarn-
+
+      # 'yarn install' must be done at the top level, to build all the
+      # cross-package symlinks
+      - name: yarn install
+        run: yarn install
+      - name: check if git is dirty
+        run: git diff --stat --exit-code
+      - name: check dependencies
+        run: yarn check-dependencies
```
mhofman: Drive by, but why use `check-dependencies` instead of `--frozen-lockfile`

.github/workflows/after-merge.yml
```diff
@@ -0,0 +1,157 @@
+# Run after merge to trunk
+on:
+  push:
+    branches: [master] # $default-branch
+
+jobs:
+  build:
```
mhofman, line 9: Is there a way to keep this job in sync with the `test-all-packages.yml` version?

mhofman (approved):
Small nit on the job definition duplication for `build`, but otherwise, LGTM

> Our merge bot ensures that CI passes before merging to master

Technically one can still merge without going through mergebot. However GitHub branch protection are still enforced, and I believe that it should enforce that both the branch is up to date, and that all tests pass, right?

.github/workflows/after-merge.yml
michaelfig: So that the job has the node version string in it.  I like it that way. :)

.github/workflows/after-merge.yml
michaelfig: In the version of Yarn we are using, `--frozen-lockfile` exited successfully, even if the lockfile would have been changed.

.github/workflows/after-merge.yml
michaelfig: I introduced a new composite action.  That should help, though it is more indirect.

PR #4615, opened 2022-02-19
feat(pegasus): implement correct result and denom trace handling

### LINKED ISSUE #4614: Need to handle denom prefixes of `<ibc-port>/<ibc-channel>/`
## To Reproduce

From `@simpletrontdip` on Discord (https://discord.com/channels/585576150827532298/766087740336635924/943026098986422314):

Pegasus seems to send an IBC package with non-prefixed denom  (uphoton instead of pegasus/channel-0/uphoton)
```
{"amount":"1000000","denom":"uphoton","receiver":"cosmos1h68l7uqw255w4m2v82rqwsl6p2qmkrg08u5mye","sender":"pegasus"}
```

This makes the target chain treat it as a remote one (although it was sent back)

```
gaiad q bank balances cosmos1h68l7uqw255w4m2v82rqwsl6p2qmkrg08u5mye --node https://rpc.testnet.cosmos.network:443 --chain-id cosmoshub-testnet
balances:
- amount: "1000000"
  denom: ibc/91CC51904C9E34CD841008CDA73F92811A235501A556EAC7F81B32FAA7F706A7
- amount: "334995400"
  denom: uphoton
pagination:
  next_key: null
  total: "0"

gaiad query ibc-transfer denom-trace 91CC51904C9E34CD841008CDA73F92811A235501A556EAC7F81B32FAA7F706A7 --node https://rpc.testnet.cosmos.network:443

denom_trace:
  base_denom: uphoton
  path: transfer/channel-134
```

## Expected behavior

Pegasus implements denom prefixing semantics for IBC connections, as described in the [ICS 20 spec](https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer#data-structures)

Notably, when IBC-Pegasus sends remotePegged tokens back to their source, the sent denom string must be prefixed with their source's port and channel, such as: `{ics20Port}/{ics20Channel}/{denom}`.  When IBC-Pegasus receives a packet, localPegged tokens must only be matched if they use the channel's local prefix.

| token source |  packet sender  |  action |
| --- | --- |  ---  |
| local | us | no prefix |
| local | them | verify and strip their prefix |
| remote | us |  add our prefix |
| remote | them | no prefix |

## Platform Environment

## Screenshots

michaelfig:
closes: #4614
refs: cosmos/ibc#661

## Description

Implement IBC denom trace transforms for Pegasus, as well as use the acknowledgement packet's actual `result` and `error` properties instead of `success`.

### Security Considerations

The incorrect parsing of acknowledgement packets caused refunds to be given even though the transfer went through.  This led to a practically infinite supply of tokens on Pegasus connections to other chains.

### DIFF
--- a/packages/pegasus/src/courier.js
+++ b/packages/pegasus/src/courier.js
@@ -33,7 +33,7 @@ export const getCourierPK = (key, keyToCourierPK) => {
  * @property {ContractFacet} zcf
  * @property {ERef<BoardDepositFacet>} board
  * @property {ERef<NameHub>} namesByAddress
- * @property {Denom} remoteDenom
+ * @property {Denom} sendDenom
  * @property {Brand} localBrand
  * @property {(zcfSeat: ZCFSeat, amounts: AmountKeywordRecord) => void} retain
  * @property {(zcfSeat: ZCFSeat, amounts: AmountKeywordRecord) => void} redeem
@@ -47,7 +47,7 @@ export const makeCourierMaker =
     zcf,
     board,
     namesByAddress,
-    remoteDenom,
+    sendDenom,
     localBrand,
     retain,
     redeem,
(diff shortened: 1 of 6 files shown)


dckc (approved):
looks good.

Thanks for the clear issue write-up and unit test.

PR #4618, opened 2022-02-21
fix: virtualize payments, purses, ledger

erights:
Virtualize payment, paymentLedger, and purse,depositFacet together as a multi-faceted unit.

This does not virtualize everything that will need to be durable to support upgrade.

This does not even virtualize everything of high arity, needed for scaling, as it does not virtualize the per-purse notifier.

### DIFF
--- a/packages/ERTP/src/payment.js
+++ b/packages/ERTP/src/payment.js
@@ -1,15 +1,21 @@
 // @ts-check
 
-import { Far } from '@endo/marshal';
+import { defineKind } from '@agoric/swingset-vat/src/storeModule.js';
 
 /**
  * @template {AssetKind} K
  * @param {string} allegedName
  * @param {Brand<K>} brand
- * @returns {Payment<K>}
+ * @returns {() => Payment<K>}
  */
-export const makePayment = (allegedName, brand) => {
-  return Far(`${allegedName} payment`, {
-    getAllegedBrand: () => brand,
-  });
+export const makePaymentMaker = (allegedName, brand) => {
+  const makePayment = defineKind(
+    `${allegedName} payment`,
+    () => ({}),
+    () => ({
+      getAllegedBrand: () => brand,
+    }),
+  );
+  return makePayment;
 };
+harden(makePaymentMaker);
--- a/packages/ERTP/src/purse.js
+++ b/packages/ERTP/src/purse.js
@@ -3,6 +3,12 @@ import { defineKind } from '@agoric/swingset-vat/src/storeModule.js';
 import { AmountMath } from './amountMath.js';
 
 export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
+  // - This kind is a pair of purse and depositFacet that have a 1:1
+  //   correspondence.
+  // - They are virtualized together to share a single state record.
+  // - An alternative design considered was to have this return a Purse alone
+  //   that created depositFacet as needed. But this approach ensures a constant
+  //   identity for the facet and exercises the multi-faceted object style.
   const makePurseKit = defineKind(
     allegedName,
     () => {
--- a/packages/ERTP/src/purse.js
+++ b/packages/ERTP/src/purse.js
@@ -1,43 +1,60 @@
 import { makeNotifierKit } from '@agoric/notifier';
-import { Far } from '@endo/marshal';
+import { defineKind } from '@agoric/swingset-vat/src/storeModule.js';
 import { AmountMath } from './amountMath.js';
 
-export const makePurse = (allegedName, assetKind, brand, purseMethods) => {
-  let currentBalance = AmountMath.makeEmpty(brand, assetKind);
+export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
+  const makePurseKit = defineKind(
+    allegedName,
+    () => {
+      const currentBalance = AmountMath.makeEmpty(brand, assetKind);
 
-  /** @type {NotifierRecord<Amount>} */
-  const { notifier: balanceNotifier, updater: balanceUpdater } =
-    makeNotifierKit(currentBalance);
+      /** @type {NotifierRecord<Amount>} */
+      const { notifier: balanceNotifier, updater: balanceUpdater } =
+        makeNotifierKit(currentBalance);
 
-  const updatePurseBalance = newPurseBalance => {
-    currentBalance = newPurseBalance;
-    balanceUpdater.updateState(currentBalance);
-  };
-
-  /** @type {Purse} */
-  const purse = Far(`${allegedName} purse`, {
-    deposit: (srcPayment, optAmountShape = undefined) => {
-      // Note COMMIT POINT within deposit.
-      return purseMethods.deposit(
+      return {
         currentBalance,
-        updatePurseBalance,
-        srcPayment,
-        optAmountShape,
-      );
+        balanceNotifier,
+        balanceUpdater,
+      };
     },
-    withdraw: amount =>
-      // Note COMMIT POINT within withdraw.
-      purseMethods.withdraw(currentBalance, updatePurseBalance, amount),
-    getCurrentAmount: () => currentBalance,
-    getCurrentAmountNotifier: () => balanceNotifier,
-    getAllegedBrand: () => brand,
-    // eslint-disable-next-line no-use-before-define
-    getDepositFacet: () => depositFacet,
-  });
+    state => {
+      const { balanceNotifier, balanceUpdater } = state;
+      const updatePurseBalance = newPurseBalance => {
+        state.currentBalance = newPurseBalance;
+        balanceUpdater.updateState(state.currentBalance);
+      };
 
-  const depositFacet = Far(`${allegedName} depositFacet`, {
-    receive: purse.deposit,
-  });
-
-  return purse;
+      /** @type {Purse} */
+      const purse = {
+        deposit: (srcPayment, optAmountShape = undefined) => {
+          // Note COMMIT POINT within deposit.
+          return purseMethods.deposit(
+            state.currentBalance,
+            updatePurseBalance,
+            srcPayment,
+            optAmountShape,
+          );
+        },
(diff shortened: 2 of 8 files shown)


erights:
Assuming it passes CI enough, this is now ready for review. Thanks!

packages/ERTP/src/payment.js
```diff
@@ -1,15 +1,21 @@
 // @ts-check
 
-import { Far } from '@endo/marshal';
+import { defineKind } from '@agoric/swingset-vat/src/storeModule.js';
```
turadg, line 3: this deep import is unfortunate. I'll prioritize https://github.com/Agoric/agoric-sdk/issues/4255

packages/ERTP/src/purse.js
```diff
@@ -1,43 +1,60 @@
 import { makeNotifierKit } from '@agoric/notifier';
-import { Far } from '@endo/marshal';
+import { defineKind } from '@agoric/swingset-vat/src/storeModule.js';
 import { AmountMath } from './amountMath.js';
 
-export const makePurse = (allegedName, assetKind, brand, purseMethods) => {
-  let currentBalance = AmountMath.makeEmpty(brand, assetKind);
+export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
+  const makePurseKit = defineKind(
+    allegedName,
```
turadg, line 13: Shouldn't the allegedName include a description of the kind?

What is the kind being defined here? From `makePurseKit` I read this as a "purse kit" kind, but I'm surprised to see a kit be a durable thing with state. I thought we used kit as a collection of related functionality.

It looks like this was made to have a place outside `purse` to the deposit facet but I don't see the motivation for that. Especially since `purse` still has `getDepositFacet`.

erights:
Solving a CI failure led me down a rabbit hold, resulting in https://github.com/Agoric/agoric-sdk/pull/4618/commits/981bc1e95bf6a4914398def26ad3b450b38690cb which we should definitely discuss before merging. I would like to avoid merging this as part of this PR.

However, without it, the generated file ui-components/compiled/test/components/test-NatAmountInput.js fails, apparently because its imports include ERTP, which now depends on storeModule, which depends on the global variable `VatData`. However, apparently this generated file is run in an environment in which no one installs `VatData`.

See #4255

packages/ERTP/src/payment.js
erights: Thanks! See also my comment https://github.com/Agoric/agoric-sdk/pull/4618#issuecomment-1074462354 below.

packages/ERTP/src/purse.js
erights: > Shouldn't the allegedName include a description of the kind?

The "kind" is the pair of the `purse` and `depositFacet`, of which only the `purse` is immediately exposed.

Since we're calling (the maker returned by) `defineKind` with (a function that returns) a record with both `purse` and `depositFacets`, and only these objects are declared `Far`, `defineKind` will call `Far` with farNames of, respectively 
```js
`${allegedName} purse`
```
and
```js
`${allegedName} depositFacet`
```
which, somewhat fortuitously, actually mimics its previous behavior.

erights: > It looks like this was made to have a place outside purse to the deposit facet but I don't see the motivation for that. Especially since purse still has getDepositFacet.

The `purse` and `depositFacet` need each other, but both are exposed to the outside world as independent `Far` objects. However, between the two of them, only one state record is needed. This solution doesn't pay any storage in the `state` record for the `depositFacet` because it only needs to exist at runtime, reactivated at the same time the purse is. But it still preserves the identity of the per-purse depositFacet as a Far object.

Do you have another solution in mind?

erights:
Yay! CI passes! But only with the https://github.com/Agoric/agoric-sdk/commit/981bc1e95bf6a4914398def26ad3b450b38690cb kludge. Is there a better way?

packages/ERTP/src/purse.js
turadg: > Do you have another solution in mind?

Not yet; trying to understand this:

> both are exposed to the outside world as independent Far objects

In `master`, `depositFacet` is exposed only via `purse.getDepositFacet()`. Why the change to expose it as an independent Far object?  https://github.com/Agoric/agoric-sdk/blob/45723b48823ed59c647d4b1f4b4d7ae16c421209/packages/ERTP/src/purse.js#L35-L43

erights: The `depositFacet` is exposed *only* by `purse.getDepositFacet()`. But since it is exposed, it must be a far object. Since it has the same cardinality as purses, it must also be virtualized. But since it has no state independent of its purse, it need not be separately virtualized. It can just be virtualized as one of the cohort pair activated for reviving that state record.

warner: IIRC, we figured that `getDepositFacet()` *could* create a new facet on demand, which would just wrap the Purse and expose a deposit method, but 1: it made a good candidate for practicing the multi-facet objects, and 2: there may have been some maintain-a-constant-identity behavior that was easier/only-possible to achieve with the multi-facet approach.

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -213,7 +213,7 @@ test('Invitation', async t => {
   t.is(paramManager.getAmount('Amt'), drachmaAmount);
   const invitationActualAmount =
     paramManager.getInvitationAmount('Invite').value;
-  t.is(invitationActualAmount, invitationAmount.value);
+  t.deepEqual(invitationActualAmount, invitationAmount.value);
```
warner, line 217: Huh, why did this need to change?

packages/SwingSet/src/storeModule.js
```diff
@@ -1,4 +1,39 @@
...
+  makeScalarWeakMapStore,
+  makeScalarSetStore,
+  makeScalarWeakSetStore,
+};
+
+let MyVatData;
+if ('VatData' in globalThis) {
+  MyVatData = VatData;
+} else {
+  console.log('Emulating VatData');
```
warner, line 25: this makes me nervous, but it sounds like other people have a plan to rationalize access to `VatData`, so I'm not going to worry about it

warner (approved):
This seems like a good start. I'm not excited about the VatData thing, but I figure we'll find a solution for that sooner or later.

While reviewing this, I realized we have a GC/determinism problem, which I wrote up in #4892. It's not specific to this conversion though, so it shouldn't hold up this PR.

packages/ERTP/src/purse.js
turadg: Ah. Since this code is meant to serve as an example I'd ask that some of the above be documented in it. If I've got this right,
```
export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
  // - This kind is a pair of purse and depositFacet that have a 1:1 correspondence.
  // - They are virtualized together to share a single state record.
  // - An alternative design considered was to have this return a Purse alone
  //   that created depositFacet as needed but this approach ensures a constant
  //   identity for the facet and exercises the multi-faceted object style.
  const makePurseKit = defineKind(
```

packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -434,7 +434,7 @@ test('issuer.combine bad payments', async t => {
   await t.throwsAsync(
     () => E(issuer).combine(payments),
     {
-      message: '"payment" not found: "[Alleged: other fungible payment]"',
+      message: /.* "\[Alleged: other fungible payment\]"/,
```
turadg: there's no `^` so it the leading wildcard necessary?

packages/governance/test/unitTests/test-buildParamManager.js
turadg: Surprising to me too. Running the test without this change,
```
  Values are deeply equal to each other, but they are not the same:

  [
    {
      description: 'simple',
      handle: Object @Alleged: InvitationHandle {},
      installation: Object @Alleged: Installation {
        getBundle: Function getBundle {},
      },
      instance: Object @Alleged: InstanceHandle {},
    },
  ]

  › packages/governance/test/unitTests/test-buildParamManager.js:216:5
```
Comparing each named key, the are the same value,
```
  for (const k of Object.keys(invitationActualAmount[0])) {
    t.is(invitationActualAmount[0][k], invitationAmount.value[0][k]);
  }
```
But the object containing them is different. Do you know why that is @erights ?

packages/ERTP/src/purse.js
erights: Done, thanks.

packages/governance/test/unitTests/test-buildParamManager.js
erights: > But the object containing them is different. Do you know why that is @erights ?

Yes. The amount is a copyRecord, which is pass-by-copy. It is serialized into persistent storage, and restored from persistent storage by unserializing. This preserves `keyEQ` equivalence which happens to be similar enough to AVA's `deepEqual` equivalence. We do not preserve JavaScript object identity for pass-by-copy objects.

erights: I added a `keyEQ` test to clarify.

packages/SwingSet/src/storeModule.js
erights: Thanks. Resolving.

packages/governance/test/unitTests/test-buildParamManager.js
turadg: Got it. Tests also serve as documentation of how things are expected to work, so would it be more accurate to use `keyEQ` here?

```suggestion
  t.true(keyEQ(invitationActualAmount, invitationAmount.value));
```

packages/ERTP/test/unitTests/test-issuerObj.js
erights: Done.

turadg (approved):
Golden!

PR #4619, opened 2022-02-21
Implement the overhauled virtual object API

### LINKED ISSUE #4606: Overhaul virtual objects API to address various problems found adapting code to it

Recently, @erights and @dtribble undertook an exercise to estimate the difficulty of adapting our various contracts from the currently ubiquitous objects-as-closures style to instead use virtual objects. The exercise was to take one of the existing contract objects, convert it, and see how hard the conversion effort actually ended up being in practice. We anticipate quite a bit of this kind of conversion will be needed in preparation for MN-1 and so we want to get a handle on how much effort all this will take.  The exercise went mostly smoothly, but the code choked on a bug when they tried to test it.  @erights and @FUDCo traced the bug to a re-entrancy issue in virtual object initialization.  In the course of sorting this out we identified a way to change the `makeKind` API to prevent the problem situation from even being possible, as well as to make it much more obvious to a developer how to rework the code to accomplish the job that the problematic code had been trying to do in the first place.  Since we anticipate some of the other contracts may find themselves in a similar situation, it made sense to examine this API change concept more closely, which we did.  The outcome of this discussion was a strong consensus between @erights, @FUDCo, and @warner that this revised design was superior to the present one in almost every aspect, and so made the decision to proceed with it.

## Description of the Design

The revised design for `makeKind` replaces the current single, rather complicated parameter (I won't restate the existing API design here; consult the SwingSet `virtual-objects.md` document for details) with three parameters: a documentary tag string, an `init` function, and an `actualization` function (the names of the parameters are subject to revision if/when we develop better terminology; they are not material to the published API as these names are principally for documentation purposes).  We also decided, while we were at it, to rename `makeKind` to `defineKind`, to better reflect its purpose and to slightly reduced the bewildering variety of things we have named `makeXXX`.  The new API is thus:
```
makerFunction = defineKind(tagString, init, actualize);
```
where:
* `tagString` is a documentary tag string, exactly as used in our `Far` function (and we mean "exactly" because that is what it will be used for)
* `init` is an initialization function, taking as args whatever initialization arguments are desired by the kind developer and returning a plain JS object representing an initialized state for the object.  This replaces the `init` function returned by the instance kit maker that the current API uses.  It differs in that it generates and returns the state object itself rather than closing over an empty one that is provided for it that it just populates.
* `actualize` is a function that provides the behavior for instances of the virtual object.  It takes one argument, `state`, which represents the state of the virtual object, and returns an object full of methods that becomes the actual in-memory representative of the virtual object.  It replaces the `self` property returned by the instance kit maker that the current API uses.  It differs from `self` in that it is a function that returns a value rather than a value directly.

The return value from `defineKind` is an instance maker function exactly the same as that currently returned by the present `makeKind`.

A notable consequence of this design is that the `init` function has no access to the actualized self and `actualize` function only ever sees a fully initialized object state (which was the proximal cause of the reentrancy problem mentioned above).

## Security Considerations

There may be undiagnosed, slippery determinism/GC visibility issues associated with the virtual object machinery in general (though we have taken care to prevent these and are reasonably confident that we have been successful, that is still an unproven hypothesis).  However, we are fairly confident that this API change does not introduce any new such issues that were not already present.

## Test Plan

Once the existing body of virtual object implementation tests are rewritten to make use of this new API, those tests should be sufficient to exercise this API as well.

FUDCo:
This implements the revised virtual object API design as described in #4606

I'm very happy with the code simplifications this enabled.  In particular, an entire large and complicated test went away completely because the largish set of weird edge cases that it checked to ensure were handled correctly are no longer even expressible.

Closes #4606

### DIFF
--- a/packages/SwingSet/docs/virtual-objects.md
+++ b/packages/SwingSet/docs/virtual-objects.md
@@ -15,62 +15,62 @@ A virtual object is an object with a durable identity whose state is automatical
 
 A virtual object has a "kind", which defines what sort of behavior and state it will possess.  A kind is not exactly a data type, since it comes with a concrete implementation, but it indicates a family of objects that share a set of common behaviors and a common state template.
 
-A vat can define new kinds of virtual object by calling the `makeKind` function provided as a vat global:
+A vat can define new kinds of virtual object by calling the `defineKind` function provided as a vat global:
 
-  `makeKind(instanceKitMaker)`
+  `defineKind(descriptionTag, init, actualize, finish)`
 
-The return value from `makeKind` is a maker function which the vat can use to create instances of the newly defined object kind.
+The return value from `defineKind` is a maker function which the vat can use to create instances of the newly defined object kind.
 
-The `instanceKitMaker` parameter is a user provided function that binds an in-memory instance of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.
+The `descriptionTag` parameter is a short description string for the kind.  This is the same kind of tag string you would use in a call to `Far`.  It will appear on `.toString()` representations of corresponding `Presence` objects that are exported to remote vats.
 
-The `instanceKitMaker` function is passed the virtual object's state as a parameter, and is expected to return a kit object with two properties: `init`, an initialization function; and `self`, a new Javascript object with methods that close over the given state, that will become the body of the new instance.
+The `init` parameter is a function that will be called when new instances are first created. It is expected to return a simple JavaScript object that represents the initialized state for the new virtual object instance.  Any parameters passed to the maker function returned by `defineKind` are passed directly to the `init` function.
 
-The `init` function is called when new instances are first created, and is expected to initialize the state object.  Any parameters passed to the maker function returned by `makeKind` are passed directly to the `init` function.
+The `actualize` parameter is a function that binds an in-memory instance (the "Representation) of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return a new Javascript object with methods that close over the given state.  This returned object will become the body of the new instance.  It is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.
+
+The `finish` parameter will, if present (it is optional), be called exactly once as part of instance initialization.  It will be invoked immediately after the `actualize` function is called for the first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function and thus becomes available to whoever requested its creation.  `finish` is passed two parameters: the virtual object's state (exactly as passed to the `actualize` function) and the virtual object itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other virtual objects.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
 
 For example:
 
 ```javascript
-  function counterKitMaker(state) {
+  function initCounter(name) {
     return {
-      init(name = 'thing') {
-        state.counter = 0;
-        state.name = name;
-      },
-      self: Far('counter', {
-        inc() {
-          state.counter += 1;
-        },
-        dec() {
-          state.counter -= 1;
-        },
-        reset() {
-          state.counter = 0;
-        },
-        rename(newName) {
-          state.name = newName;
-        },
-        getCount() {
-          return state.counter;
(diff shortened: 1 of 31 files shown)


packages/SwingSet/docs/virtual-objects.md
```diff
@@ -15,62 +15,57 @@ A virtual object is an object with a durable identity whose state is automatical
...
+A vat can define new kinds of virtual object by calling the `defineKind` function provided as a vat global:
 
-  `makeKind(instanceKitMaker)`
+  `defineKind(descriptionTag, init, actualize)`
 
-The return value from `makeKind` is a maker function which the vat can use to create instances of the newly defined object kind.
+The return value from `defineKind` is a maker function which the vat can use to create instances of the newly defined object kind.
 
-The `instanceKitMaker` parameter is a user provided function that binds an in-memory instance of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.
+The `descriptionTag` parameter is a short description string for the kind.  This is the same kind of tag string you would use in a call to `Far`.
```
warner: suggest adding "it will appear on `.toString` representations of the corresponding `Presence` objects that appear in remote vats"

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -15,62 +15,57 @@ A virtual object is an object with a durable identity whose state is automatical
...
 
-The `init` function is called when new instances are first created, and is expected to initialize the state object.  Any parameters passed to the maker function returned by `makeKind` are passed directly to the `init` function.
+The `actualize` parameter is a function that binds an in-memory instance of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return a new Javascript object with methods that close over the given state.  This returned object will become the body of the new instance.  It is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.
```
warner: let's introduce the term `Representative` here, maybe ".. that binds an in-memory instance (the "Representative") of the virtual object .."

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -15,62 +15,57 @@ A virtual object is an object with a durable identity whose state is automatical
...
+      },
+      getCount() {
+        return state.counter;
+      },
+      getName() {
+        return state.name;
+      },
     };
   }
+  const makeCounter = defineKind('counter', initCount, actualizeCounter);
```
warner: `'initCounter' !== 'initCount'`, not sure which one you want to pick

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -80,9 +75,11 @@ and then use it like this:
   console.log(`${barCounter.getName()} count is ${barCounter.getCount()`); // "new bar count is 1"
 ```
 
+Note that the `init` and `actualize` functions are defined explicitly here for clarity of exposition, but in practice you'd usually declare them inline as arrow functions in the parameters of the `defineKind` call.
+
 Additional important details:
 
-- The set of state properties is determined by the `init` function.  The properties that exist when `init` returns _are_ the properties of the object.  State properties cannot thereafter be added or removed.
+- The set of state properties is fully determined by the `init` function.  The of the object that `init` returns _are_ the properties of the virtual object's state.  State properties cannot thereafter be added or removed.
```
warner: typo, "The of the object".. maybe say something more like:

The set of properties on `state` are completely determined by the enumerable properties of the object that `init` returns. After initialization, the values of `state`'s properties can be modifed, but new properties cannot be added, nor existing ones removed.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -96,22 +93,22 @@ Additional important details:
...
-      state.foo = 47;
+  const makeThing = defineKind(
+    'thing',
+    () => ({
+      someProperty: 47,
+    }),
+    state => {
+      const self = {
+        sendMeTo(somebodyElse) {
+          somebodyElse.hereIAm(state.someProperty, self);
```
warner: I *think* this runs into the same problem we previously had, if someone decides to make that cross-linkage `hereIAm` call at actualization time rather than waiting for a method to be invoked, and `self` is not yet registered anywhere. OTOH, if we're doing the `Far` marker inside the VOM, then `self` at this point is not passable (it isn't even hardened yet, nor marked as Far), which might catch this mistake early. So that's a clear improvement.

packages/SwingSet/test/virtualObjects/test-representatives.js
```diff
@@ -24,132 +24,6 @@ function slot0(iface, kid) {
   };
 }
 
-test('virtual object representatives', async t => {
```
warner, line 27: I see a whole big test was deleted.. is this functionality now covered by something else?

warner (approved):
Looks good, just some minor docs suggestions

packages/SwingSet/test/virtualObjects/test-representatives.js
FUDCo: No, this test went away because the set of weird edge cases whose proper handling it tested are no longer expressible in the API.  A corollary to that is that the test itself is no longer expressible.

Issue #506, opened 2020-02-05
[pixel-demo, cosmic-swingset, agoric-cli, documentation] Shelve the Pixel Demo For Now


katelynsills:
Our current priority is getting an MVP of ERTP and Zoe on the testnet with developer tools like the agoric-cli to create contracts and apps and deploy them. 

The pixel-demo doesn't help us do that, and it needs a significant amount of attention to make it a good experience. We would need to rewrite it to use Zoe rather than the contractHost (#495) for example, in addition to UI work. Thus, we decided to temporarily shelve it until it can get the attention that it needs. In its place, we will promote the wallet frontend and autoswap frontend and encourage people to interact with contracts running on Zoe

We recently moved the pixel-demo into its own package in the mono-repo, taking it out of ERTP. Shelving would mean leaving the package in the mono-repo, but snipping off the imports in cosmic-swingset. We can also move any pixel-demo configs to the pixel-demo package, since we aren't trying to support it in ERTP any longer.

katelynsills:
Also incidentally closes #252

PR #4623, opened 2022-02-21
refactor: Remove bundle helpers

kriskowal:
refs: [#506](https://github.com/endojs/endo/issues/506)

This change removes the old copies of `import-bundle` and `bundle-source`, which have since moved to the `@endo` org in npm. All references have already been pointed away, but the packages remained because our tested dapps still depended upon them directly and couldn’t get the correct `@agoric` versions from npm.


michaelfig (approved):
LGTM!  Negative code day!

Feel free to merge after resolving the conflicts I caused and you reviewed. :)

Issue #4624, opened 2022-02-21
Robust Cosmos SDK state rollback


michaelfig:
## What is the Problem Being Solved?

The Cosmos SDK modules frequently perform speculative execution of transactions/messages, capture a result or error, then throw away the accumulated state to roll back the execution.  If SwingSet state is also mutated and not also rolled back before commit, this breaks the correspondence of Cosmos to SwingSet state.

This happens at least for:
- to process an `x/gov` proposal submission, Cosmos validates the proposal by speculatively executing the enactment handler, then discarding the enactment state
- during transaction simulation (at any time on an RPC node), Cosmos speculatively executes the transaction, then records the consumed gas, and finally, discards the execution state

## Description of the Design

Instead of calling JS functions from arbitrary Cosmos modules, only do side-effects within the `x/swingset` lifecycle events `BEGIN_BLOCK`, `END_BLOCK`, and `COMMIT_BLOCK`.

All other Cosmos interactions with SwingSet will be in the form of adding an action record to the `actionQueue` in the Cosmos kvstore state, for later dequeuing (and deletion) by cosmic-swingset during one of the lifecycle events.  Because this state is kept and updated in the kvstore, it will be properly rolled back if the kvstore changes are discarded before the lifecycle event.

The above queuing will be the coordination mechanism for any other per-message state changes.  This also enforces that responses from SwingSet are guaranteed to be asynchronous, and no leakage of nondeterministic behaviour results can propagate from cosmic-swingset back to the Golang code doing the queuing.

### Large data

Some transactions, like the bundlecaps upload transaction, require more storage than easily fits in the kvstore.

As an example of out-of-band data communication, we will queue an action describing a deterministic filename (indicated by a kvstore counter) and what to do with it, capturing additional data into that filename in a single-block-only staging directory.  Cosmic-swingset will delete this staging directory's contents upon `BEGIN_BLOCK` before consuming a new `actionQueue`.  If the action is rolled back, then the deletion step ensures that disk space is recovered.

It may be desirable to use this large data feature for any kind of data that is user-generated, in order to reduce memory pressure on the kvstore in exchange for potentially slower throughput.

## Security Considerations

Better partitioning between Cosmos and JS.

dckc:
closed per @michaelfig

PR #4625, opened 2022-02-21
Make the Cosmos-to-JS communication path transactional

michaelfig:
refs: #4624

## Description

This PR adds the beginning of wallet transactions, CoreEval governance, and notably, improves the Cosmos-to-JS path by ensuring rolled-back Cosmos transactions cannot have any effect on SwingSet.

@JimLarson please review Golang code, skipping over the generated files.  @warner please review the JS code, especially the block manager and how it's invoked.


JimLarson:
Ping me when the golang tests are fixed.

michaelfig:
> Ping me when the golang tests are fixed.

Oops, sorry about that!  Ready now, @JimLarson.

golang/cosmos/app/app.go
```diff
@@ -548,18 +553,26 @@ func NewAgoricApp(
 		slashingtypes.ModuleName,
 		evidencetypes.ModuleName,
 		stakingtypes.ModuleName,
+		authtypes.ModuleName, banktypes.ModuleName, govtypes.ModuleName, crisistypes.ModuleName, genutiltypes.ModuleName,
```
JimLarson: One per line, please. Here and for EndBlockers and InitGenesis too.

golang/cosmos/x/lien/keeper/keeper.go
```diff
@@ -33,20 +33,20 @@ type keeperImpl struct {
...
-	callToController func(ctx sdk.Context, str string) (string, error)
+	pushAction func(ctx sdk.Context, action interface{}) error
 }
 
 // NewKeeper returns a new Keeper.
 // The ak must be the same accout keeper that the bk and sk use.
-func NewKeeper(key sdk.StoreKey, cdc codec.Codec, ak *types.WrappedAccountKeeper, bk types.BankKeeper, sk types.StakingKeeper,
-	callToController func(ctx sdk.Context, str string) (string, error)) Keeper {
+func NewKeeper(cdc codec.Codec, key sdk.StoreKey, ak *types.WrappedAccountKeeper, bk types.BankKeeper, sk types.StakingKeeper,
+	pushAction func(ctx sdk.Context, action interface{}) error) Keeper {
```
JimLarson: Please document that `action` must be a type that `json.Marshal` understands. Alternatively, create and "action-pusher" type abstraction that contains this documentation. I'm really allergic to Go's practice of declaring `interface{}` types with undocumented expectations on the values you can use.

```
// actionPusher enqueues data for later transmission over the controller.
// The action must be a type that json.Marshal() understands.
type actionPusher func(ctx sdk.Context, action interface{}) error
```

Alternatively

```
// Jsonable is a data type that json.Marshal() understands.
type Jsonable interface{}
```

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -68,10 +69,49 @@ func NewKeeper(
...
 		bankKeeper:       bankKeeper,
 		feeCollectorName: feeCollectorName,
-		CallToController: callToController,
+		callToController: callToController,
 	}
 }
 
+// PushAction appends an action to the controller's action queue.  This queue is
+// kept in the kvstore so that changes to it are properly reverted if the
+// kvstore as rolled back.  By the time the block manager runs, it can commit
```
JimLarson:
```suggestion
// kvstore is rolled back.  By the time the block manager runs, it can commit
```

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -68,10 +69,49 @@ func NewKeeper(
 		accountKeeper:    accountKeeper,
 		bankKeeper:       bankKeeper,
 		feeCollectorName: feeCollectorName,
-		CallToController: callToController,
+		callToController: callToController,
 	}
 }
 
+// PushAction appends an action to the controller's action queue.  This queue is
```
JimLarson, line 77: Specify that `action` must be a type that `json.Marshal` can handle. Consider using type abstraction to do so.

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -68,10 +69,49 @@ func NewKeeper(
...
+	k.SetStorage(ctx, fmt.Sprintf("actionQueue.%d", tail), string(bz))
+	bz, err = json.Marshal(tail + 1)
+	if err != nil {
+		return err
+	}
+	k.SetStorage(ctx, "actionQueue.tail", string(bz))
+	return nil
+}
+
+// BlockingSend sends a message to the controller and blocks the Golang process
```
JimLarson, line 110: Same with `obj` as with `action` above.

golang/cosmos/x/swingset/types/msgs.go
```diff
@@ -1,14 +1,16 @@
...
+
 	"github.com/Agoric/agoric-sdk/golang/cosmos/vm"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
 const RouterKey = ModuleName // this was defined in your key.go file
 
-var _, _ sdk.Msg = &MsgDeliverInbound{}, &MsgProvision{}
+var _, _, _, _ sdk.Msg = &MsgDeliverInbound{}, &MsgProvision{}, &MsgWalletAction{}, &MsgWalletSpendAction{}
```
JimLarson: This is getting ungainly. Consider splitting into separate lines within a `var(...)` block.

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -68,10 +69,49 @@ func NewKeeper(
...
+// kvstore as rolled back.  By the time the block manager runs, it can commit
+// its SwingSet transactions without fear of side-effecting the world with
+// intermediate transaction state.
+func (k Keeper) PushAction(ctx sdk.Context, obj interface{}) error {
+	bz, err := json.Marshal(obj)
+	if err != nil {
+		return err
+	}
+	tailStr := k.GetStorage(ctx, "actionQueue.tail")
+	var tail uint64
```
JimLarson: The way `actionQueue.tail` is implicitly initialized by this code is subtle and worthy of a comment. Please also somewhere describe the full queue describing which side is responsible for what, and reference that description from the producer and the consumer.

golang/cosmos/x/swingset/abci.go
```diff
@@ -47,13 +44,11 @@ func BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock, keeper Keeper) erro
 		ChainID:     ctx.ChainID(),
 		Params:      keeper.GetParams(ctx),
 	}
-	b, err := json.Marshal(action)
-	if err != nil {
-		return sdkerrors.Wrap(sdkerrors.ErrJSONMarshal, err.Error())
+	_, err := keeper.BlockingSend(ctx, action)
+	if err == nil {
+		err = keeper.PushAction(ctx, action)
```
JimLarson: I'm confused that this action is being both send over the bridge and stored in the action queue - will it be double-interpreted on the JS side? Worthy of a comment either way. Also comment on the significance of sending first here, enqueueing first in `EndBlock`, and not enqueueing for `CommitBlock`.

JimLarson (changes requested):
Mostly requests for cleanup and comments - but I'm confused enough by `abci.go` that I'd like to see the answer before approving.

golang/cosmos/app/app.go
michaelfig: I have to push back against this specifically.  The `cosmos-sdk/simapp/app.go` groups these modules, too, and I find it easier to visually compare the order when they are grouped similarly.

golang/cosmos/x/lien/keeper/keeper.go
michaelfig: Thanks, I implemented both of them.

golang/cosmos/x/swingset/abci.go
michaelfig: You identified what was a wart in the block-manager.js.  I've updated it now.

golang/cosmos/x/swingset/client/cli/tx.go
```diff
@@ -113,3 +116,89 @@ func GetCmdProvisionOne() *cobra.Command {
...
 	return cmd
 }
+
+func NewCmdSubmitCoreEvalProposal() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "swingset-core-eval [[permit.json] [code.js]]...",
+		Args:  cobra.MinimumNArgs(2),
+		Short: "Submit a proposal to evaluate code in the SwingSet core",
+		Long: `Submit a SwingSet evaluate core Compartment code proposal along with an initial deposit.
+Specify at least one pair of permit.json and code.js files`,
```
warner, line 126: What format is `code.js` required to be? I glanced through the PR and didn't see any docs. It'll depend on how it gets evaluated, of course, but this `Long:` seems like a good place to explain it. Is it the output of `bundleSource` (in which case, we need to identify what exported function will be invoked). Or is it supposed to be a plain string which is supposed to cause side-effects during evaluation?

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -68,8 +71,57 @@ func NewKeeper(
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Set the storage corresponding to the queue entry for the current tail.
+	k.SetStorage(ctx, fmt.Sprintf("actionQueue.%d", tail), string(bz))
+
+	// Update the tail to point to the next available entry.
+	bz, err = json.Marshal(tail + 1)
```
warner: using JSON to write, but `strconv` to read, seems a bit weird, is there a reason to not use the same tool in both directions?

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -79,6 +82,71 @@ const makeChainStorage = (call, prefix = '', imp = x => x, exp = x => x) => {
...
+ *
+ * Parallel access is not supported, only a single outstanding operation at a
+ * time.
+ *
+ * @param {(obj: any) => any} call send a message to the chain's storage API and
+ * receive a reply
+ * @param {string} [prefix] string to prepend to the queue's storage keys
+ */
+const makeChainQueue = (call, prefix = '') => {
+  const storage = makeChainStorage(call, prefix);
```
warner, line 111: I'm slightly nervous about the pre-existing format-confusion opportunity in `makeChainStorage`. It'd be nice if `prefix` were forbidden to include some specific delimiter (maybe `:`), and the `${prefix}${key}` concatenation were replaced with `${prefix}:${key}`, so there was no way for someone to mess up and confuse `{ prefix: 'abc', key: 'def' }` with `{ prefix: 'ab', key: 'cdef' }`.

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -162,34 +173,32 @@ export default function makeBlockManager({
         verboseBlocks && console.info('block', action.blockHeight, 'begin');
 
         // Start a new block, or possibly replay the prior one.
-        const leftoverActions = currentActions.filter(
-          a => a.blockHeight !== action.blockHeight,
-        );
-        if (leftoverActions.length) {
+        const leftovers = new Map();
```
warner: Maybe rename to `leftoverTypes` to make it clear that we're aware of the possibility of e.g. multiple `DELIVER_INBOUND` in a single block. This code doesn't need to see all the actions, it's only gathering the type names to build a debugging message.

warner (approved):
JS part looks pretty good

golang/cosmos/app/app.go
JimLarson: Ok

golang/cosmos/x/swingset/keeper/keeper.go
JimLarson: Still needs a `vm.Jsonable` here.

golang/cosmos/x/vbank/keeper/keeper.go
```diff
@@ -6,6 +6,8 @@ import (
 
 	"github.com/Agoric/agoric-sdk/golang/cosmos/x/vbank/types"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+
+	vm "github.com/Agoric/agoric-sdk/golang/cosmos/vm"
```
JimLarson, line 10: No need for the package alias :-).

golang/cosmos/x/vbank/vbank.go
```diff
@@ -69,10 +69,10 @@ type vbankBalanceUpdate struct {
 	Updated vbankManyBalanceUpdates `json:"updated"`
 }
 
-func marshalBalanceUpdate(ctx sdk.Context, keeper Keeper, addressToBalance map[string]sdk.Coins) ([]byte, error) {
+func getBalanceUpdate(ctx sdk.Context, keeper Keeper, addressToBalance map[string]sdk.Coins) interface{} {
```
JimLarson: Consider returning `vm.jsonable`.

golang/cosmos/x/vbank/vbank.go
```diff
@@ -97,7 +97,14 @@ func marshalBalanceUpdate(ctx sdk.Context, keeper Keeper, addressToBalance map[s
 
 	// Ensure we have a deterministic order of updates.
 	sort.Sort(event.Updated)
-	return json.Marshal(&event)
+	return event
+}
+
+func marshal(event interface{}) ([]byte, error) {
```
JimLarson: Consider accepting `vm.Jsonable`.

golang/cosmos/x/swingset/client/cli/tx.go
michaelfig: It will be an evaluable string.  Understanding the format of it (as well as the `permit.json`) requires understanding the `@agoric/vats/src/core/` system.  There will certainly be documentation, but this help string is not the place for it.

golang/cosmos/x/swingset/keeper/keeper.go
michaelfig: Thanks for this.  I will use `fmt.Sprintf` instead of `json.Marshal`.

packages/cosmic-swingset/src/chain-main.js
michaelfig: I will gladly assert that the `prefix` must either be `''` or end in `'.'`.

Format confusion is avoided by convention because `makeChainStorage` and `makeChainQueue` are closely held by the manager of the namespace (the chain host).  If that host creates collisions in the key namespace, they get what they pay for.  What attack are you concerned about?

golang/cosmos/x/vbank/vbank.go
```diff
@@ -227,9 +234,6 @@ func (ch portHandler) Receive(ctx *vm.ControllerContext, str string) (ret string
 	return
 }
 
-func (am AppModule) CallToController(ctx sdk.Context, send string) (string, error) {
-	// fmt.Println("vbank.go upcall", send)
-	reply, err := am.keeper.CallToController(ctx, send)
-	// fmt.Println("vbank.go upcall reply", reply, err)
-	return reply, err
+func (am AppModule) PushAction(ctx sdk.Context, action interface{}) error {
```
JimLarson: Jsonable

golang/cosmos/x/vibc/ibc.go
```diff
@@ -149,11 +149,10 @@ func (ch portHandler) Receive(ctx *vm.ControllerContext, str string) (ret string
 	return
 }
 
-func (am AppModule) CallToController(ctx sdk.Context, send string) (string, error) {
+func (am AppModule) PushAction(ctx sdk.Context, action interface{}) error {
```
JimLarson: Jsonable

JimLarson (approved):
Some more opportunities for Jsonable, but otherwise great!

PR #4629, opened 2022-02-21
DO NOT MERGE: Bug snapshot for FUDCo

erights:
Snapshot of the current state of #4618 because it reliably provokes a kernel bug that @FUDCo wishes to investigate.

To reproduce

```
$ yarn && yarn build
$ cd packages/ERTP
$ yarn test test/unitTests/test-interfaces.js
```

If you do this under a debugger, there's already a `debugger;` statement at a place the bug appears.


erights:
No longer relevant. Closing.

PR #4631, opened 2022-02-22
test: Ensure access to reduced counterexamples from fast-check

gibson042:
## Description

Updates use of fast-check assert to conform with its documentation: https://dubzzz.github.io/fast-check/index.html#assert

> WARNING: Has to be awaited
> …
> Returns Promise\<void>

### Security Considerations

n/a

n/a

### Testing Considerations

Noted above.

### DIFF
--- a/packages/ERTP/test/unitTests/test-amountProperties.js
+++ b/packages/ERTP/test/unitTests/test-amountProperties.js
@@ -4,6 +4,7 @@ import fc from 'fast-check';
 
 import { AmountMath as m, AssetKind } from '../../src/index.js';
 import { mockBrand } from './mathHelpers/mockBrand.js';
+import { assertionPassed } from '../../../store/test/test-store.js';
 
 // Perhaps makeCopyBag should coalesce duplicate labels, but for now, it does
 // not.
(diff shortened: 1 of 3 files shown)


Chris-Hibbert (approved):
Thanks for noticing this, Richard.

FYI, @dckc, who introduced these tests.

dckc:
Weird. Their [tutorial example](https://github.com/dubzzz/fast-check/blob/main/documentation/HandsOnPropertyBased.md#hands-on-fast-check) treats `fc.assert` as sync.

gibson042:
It took some digging, but I think I got to the bottom of it—`fc.assert` has multiple signatures, and is asynchronous if and only if the property being tested is asynchronous. So landing this PR is safe (and possibly future-proof) but unnecessary.

gibson042:
I also opened an issue in fast-check: https://github.com/dubzzz/fast-check/issues/2749

packages/ERTP/test/unitTests/test-amountProperties.js
```diff
@@ -28,8 +28,8 @@ const arbAmount = arbBagContents.map(contents =>
 // Note: we write P => Q as !P || Q since JS has no logical => operator
 const implies = (p, q) => !p || q;
 
-test('isEqual is a (total) equivalence relation', t => {
-  fc.assert(
+test('isEqual is a (total) equivalence relation', async t => {
+  await fc.assert(
```
dubzzz, line 33: Not needed given the fact that you are using a sync property.

packages/ERTP/test/unitTests/test-amountProperties.js
```diff
@@ -43,9 +43,9 @@ test('isEqual is a (total) equivalence relation', t => {
   );
 });
 
-test('isGTE is a partial order with empty as minimum', t => {
+test('isGTE is a partial order with empty as minimum', async t => {
   const empty = m.makeEmpty(mockBrand, AssetKind.COPY_BAG);
-  fc.assert(
+  await fc.assert(
```
dubzzz, line 64: Same

packages/ERTP/test/unitTests/test-amountProperties.js
```diff
@@ -56,9 +56,9 @@ test('isGTE is a partial order with empty as minimum', t => {
   );
 });
 
-test('add: closed, commutative, associative, monotonic, with empty identity', t => {
+test('add: closed, commutative, associative, monotonic, with empty identity', async t => {
   const empty = m.makeEmpty(mockBrand, AssetKind.COPY_BAG);
-  fc.assert(
+  await fc.assert(
```
dubzzz, line 86: Same

packages/ERTP/test/unitTests/test-amountProperties.js
```diff
@@ -76,8 +76,8 @@ test('add: closed, commutative, associative, monotonic, with empty identity', t
   );
 });
 
-test('subtract: (x + y) - y = x; (y - x) + x = y if y >= x', t => {
-  fc.assert(
+test('subtract: (x + y) - y = x; (y - x) + x = y if y >= x', async t => {
+  await fc.assert(
```
dubzzz, line 127: Same

packages/run-protocol/test/amm/vpool-xyk-amm/test-amm-calc.js
```diff
@@ -37,8 +37,8 @@ const withinEpsilon = (left, right) =>
   multiply(right, 105) >= multiply(left, 100) &&
   multiply(left, 105) >= multiply(right, 100);
 
-test('balancesToReachRatio calculations are to spec', t => {
-  fc.assert(
+test('balancesToReachRatio calculations are to spec', async t => {
+  await fc.assert(
```
dubzzz, line 41: Same

dubzzz:
Only `asyncProperty` has to be awaited. Normal `property` can be but it is not needed. Anyway, always awaiting can avoid forgetting about it and will work for both (while not awaiting the async one will miss bugs).

Issue #4249, opened 2022-01-07
Atomically adjust pool ratio and add liquidity


Chris-Hibbert:
## What is the Problem Being Solved?

When the RUN Protocol wants to add liquidity, the pool's ratio may not match the intended price.

## Description of the Design

Add an API to the AMM that allows adding liquidity at a stated ratio. Presumably, this will take the ratio of the of the offered collaterals and trade with the pool to match that ratio, then take the remaining amounts in give and add them as liquidity.

I haven't been able to find a way to calculate the trade values that adjusts for the impact of the fees, but it looks like that doesn't make a big difference, so I expect to ignore that effect.

## Security Considerations

Handling this in a single call means that no other transaction will interfere. It's necessary to do a trade with fees, because the alternative would allow ostensible liquidity providers to trade without paying complete fees.

## Test Plan

Verify that the resulting pool balance is close to the ratio offered by the liquidity provider.

Tartuffo:
How is this different from #4200?

Chris-Hibbert:
#4200 was a previous approach. This supersedes it.

PR #4635, opened 2022-02-22
feat!: add collateral Reserve to hold collateral and add to AMM under governance control

### LINKED ISSUE #4188: Collateral Reserve

We need a mechanism to manage crypto assets to improve the economy's liquidity and stability.  Multiple streams of assets must be disbursed or managed in different ways. Here it is called it the PCV (protocol-controlled value) Exchange. It has been referred to in other places as "The RUN protocol".

## Description of the Design

A Zoe contract holds onto various assets, and supports governance mechanisms to direct it to disburse them in various ways.
The assets include
 * crypto assets, which might have been deposited directly or might be the result of other operations.
    . For now, there is no intent to track sources, or ascribe partial ownership to any party.
	. Most of these will be disbursed by using them to fund AMM pools.
 * BLD, which can be sold directly.
 * Liquidity tokens from funding AMM pools.
   . For now, the expectation is that the contract will hold onto these. Some future upgrade may enable redeeming them in some cases.

A book order exchange contract for the sale of BLD for various asset types
 * The contract's creator can offer BLD for sale at specified times, velocities, tranche sizes
 * At time of sale, will sell up to a limit to the best offers better than some floor.
 * proceeds go in the pools

A governance-controlled mechanism for withdrawing (non-BLD) assets from pools, minting RUN, and creating and funding new AMM pools

A separate governance-controlled mechanism for withdrawing assets from pools and funding existing AMM pools.
This would require work on the AMM to allow adding arbitrary an mix of RUN/collateral to a pool

## Security Considerations

* takes no action without governance direction
* Book Order exchange controlled by Zoe guarantees
* Need an upgrade mechanism, since this list of functionality isn't complete.
* All the components need to support governance-style legibility
* The contract's holdings must be visible.

## Test Plan

Extensive, but the specifics will be details with each component.

Chris-Hibbert:
closes: #4188
refs: #4249

## Description
    
Includes new functionality for Governance to be able to invoke a method on the creator facet of a governed contract.
    
The CollateralReserve accepts donations of collateral, and responds to governance decisions jto add collateral to the AMM (along with freshly minted RUN.)
    
**Happy path tests pass.**

**type declarations are not stable yet; tsc isn't happy.**

### Security Considerations

Intended to be a crucial component of RUN protocol.

The changes to Contract Governor need to maintain all the security properties of governance, and ensure continued legibility.

This new contract (the collateral reserve) is not intended for normal users of the chain to interact with, but it is a core part of the economy, so we need to write it up so people understand its role and how it works.

The new governance functionality for invoking methods on contracts also needs to be written up.

### Testing Considerations

Tests are incomplete. We may want to include this in a branch for Restival.


packages/run-protocol/test/reserve/setup.js
```diff
@@ -59,6 +59,14 @@ const setupReserveBootstrap = async (
   };
 };
 
+/**
+ * @typedef {{
```
Chris-Hibbert, line 63: When is this better than 
```
  * @typedef {Object} ReserveKit
  * @property {
```

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -273,6 +273,7 @@ const start = async (zcf, privateArgs) => {
   );
 
   const creatorFacet = wrapCreatorFacet(
+    // @ts-expect-error ??? is there any constraint on what can be wrapped? Reading elsewhere I thought getGovernedApis() was required
```
Chris-Hibbert: My best approach at the moment is to require it, and in contracts that don't have any governed APIs (no methods that should be callable under governance) `getGovernedApis()` returns `empty`or throws.  Is there a better way to implement this or a better way to declare it?

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -198,9 +205,13 @@ const start = async (zcf, privateArgs) => {
...
+  // @ts-expect-error cast
   return harden({ creatorFacet, publicFacet });
 };
 
 harden(start);
 
 export { start };
+
+/** @typedef {Required<ReturnType<typeof start>>['publicFacet']} CollateralReservePublicFacet */
+/** @typedef {Required<ReturnType<typeof start>>['creatorFacet']} CollateralReserveCreatorFacet */
```
Chris-Hibbert: what does `Required<>` do here?

packages/governance/test/swingsetTests/contractGovernor/governedContract.js
```diff
@@ -63,7 +63,8 @@ const start = async (zcf, privateArgs) => {
 
   return {
     publicFacet: wrapPublicFacet({}),
-    creatorFacet: wrapCreatorFacet({}),
+    // ??? is this a requirement for contract being governed? (let's distinguish between "contract governed" (inner) and "governed contract" (outer))
+    creatorFacet: wrapCreatorFacet({ getGovernedApis: () => {} }),
```
Chris-Hibbert: I thought it could be added by default in `contractHelper`.  It seems like a mistake to require that  contracts that don't have governed APIs to declare an empty list. (Though it's not the worst choice.) 
Have I declared things so that the governed contract has to provide it?

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -15,8 +15,15 @@ const makeLiquidityKeyword = keyword => `${keyword}Liquidity`;
...
- * GovernedCreatorFacet<any>,
+ * GovernedPublicFacet<{
+ *  makeAddCollateralInvitation: any,
+ * }>,
+ * GovernedCreatorFacet<{
+ *  addIssuer: (i: Issuer, b: string) => void,
+ *  getAllocations: any,
+ *  makeAddCollateralInvitation: any,
+ *  getGovernedApis: () => unknown,
+ * }>,
```
Chris-Hibbert: Isn't this 
```
GovernedPublicFacet<CollateralReservePublicFacet>,
GovernedCreatorFacet<CollateralReserveCreatorFacet>,
```
Do we need to repeat the declarations?

Chris-Hibbert (commented):
Thanks for the clean-ups. I have some questions, mostly for my own edification.

packages/run-protocol/test/reserve/setup.js
turadg: Style question. IMO we should use the TS syntax whenever we don't need to document with JSDoc. In this particular case I found the `import` calls to be hard to read when embedded in JSDoc.

packages/run-protocol/src/reserve/collateralReserve.js
turadg: Make every optional member no longer be optional.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: Let's hash out the design https://docs.google.com/document/d/1tePGqiu-c0wXjIdR9_IWqWP9UKIgHQeoujg7eW7lh70/edit?usp=sharing

Chris-Hibbert: `getGovernedApis` is optional. The type system should understand that.

Chris-Hibbert:
@turadg   tests don't seem to have re-started. I'm not sure why.

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -132,6 +132,81 @@ export const setupAmm = async ({
...
+}) => {
+  const bundle = await reserveBundle;
+  const reserveInstallation = E(zoe).install(bundle);
+
+  const poserInvitationP = E(committeeCreator).getPoserInvitation();
+  const [poserInvitation, poserInvitationAmount] = await Promise.all([
+    poserInvitationP,
+    E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
+  ]);
+  const timer = await chainTimerService; // avoid promise for legibility
```
turadg, line 169: how does returning a promise affect legibility?

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,212 @@
...
+const start = async (zcf, privateArgs) => {
+  const {
+    main: {
+      [CONTRACT_ELECTORATE]: electorateParam,
+      [AMM_INSTANCE_KEY]: ammInstance,
+    },
+  } = zcf.getTerms();
+
+  // we need a keyword for each issuer, but issuers don't have unique names. The
+  // first brand with any alleged name gets that as its keyword. Others get a
```
turadg: nit: so that the keywords don't ever appear to be authoritative, WDYT of always transforming the name? The note about "aren't visible to users" lessens the need for this but the special case for the first brand smells to me.

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+ *   governedApis: ['addLiquidityToAmmPool'],
+ * }
+ * >} zcf
+ * @param {{feeMintAccess: FeeMintAccess, initialPoserInvitation: Payment}} privateArgs
+ */
+const start = async (zcf, privateArgs) => {
+  const {
+    main: {
+      [CONTRACT_ELECTORATE]: electorateParam,
+      [AMM_INSTANCE_KEY]: ammInstance,
```
turadg: `_KEY` is implied by the context

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+  /** @type {MapStore<Keyword, [Brand, Brand]>} */
+  const brandForKeyword = makeStore('brands');
+
+  /**
+   *
+   * @param {*} issuer
+   * @param {string} keyword unique
+   * @returns {Promise<[argumentIssuer: Issuer, liquidityIssuer: Issuer]>}
+   */
+  const ammPublicFacet = E(zcf.getZoeService()).getPublicFacet(
```
turadg: Looks like a stray doc comment.

```suggestion
  /** @type {XYKAMMPublicFacet} */
  const ammPublicFacet = E(zcf.getZoeService()).getPublicFacet(
```
And I think `XYKAMMPublicFacet` will need `makeAddLiquidityAtRateInvitation` added to it. We could define `XYKAMMPublicFacet` by the the return type of the contract `start`. E.g. in `multipoolMarketMaker.js`,
```
/** @typedef {ContractOf<start>['publicFacet']} XYKAMMPublicFacet */
```
but I don't know how to do that without losing the JSDoc on the current `XYKAMMPublicFacet` typedef.

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+    return Promise.all([
+      zcf.saveIssuer(issuer, keyword),
+      zcf.saveIssuer(liquidityIssuer, makeLiquidityKeyword(keyword)),
+    ]);
+  };
+
+  const getKeywordsForBrand = brand => {
+    assert(
+      keywordForBrand.has(brand),
+      `Please call addIssuer for brand ${brand}`,
```
turadg:
```suggestion
      `Issuer not defined for brand ${brand}; first call addIssuer()`,
```

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+    zcf.getZoeService(),
+    ammInstance.value,
+    initialPoserInvitation,
+  );
+
+  const { wrapPublicFacet, wrapCreatorFacet, getElectorate } =
+    handleParamGovernance(zcf, paramManager);
+
+  const electorateInvAmt = getElectorate();
+  // compare invitation amount from privateArgs with electorateParam from terms
```
turadg, line 90: it's clear that the line is comparing the amounts but not why. is this "amount" is the set of the electors?
```suggestion
  // ensure set of electors in privateArgs matches terms
```

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+  assert(
+    keyEQ(electorateInvAmt, electorateParam.value),
+    X`electorate amount (${electorateParam.value} didn't match ${electorateInvAmt}`,
+  );
+
+  // const brandToKeyword = makeStore('keyword');
+  // should liquidiity tokens go in a separate seat?
```
turadg: I don't know offhand. why or why not?

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+    return collateralSeat.getCurrentAllocation();
+  };
+
+  // Anyone can deposit collateral to the reserve.
+  const addCollateralHook = async seat => {
+    const {
+      give: { Collateral: amountIn },
+    } = seat.getProposal();
+
+    const keyword = getKeywordsForBrand(amountIn.brand);
```
turadg: Took me a couple reads to understand:
```
// transfer the Collateral from the proposal to the appropriate keyword
// for that collateral type in this reserve
```

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+
+    brandForKeyword.init(keyword, [brand, liquidityBrand]);
+    keywordForBrand.init(brand, keyword);
+    return Promise.all([
+      zcf.saveIssuer(issuer, keyword),
+      zcf.saveIssuer(liquidityIssuer, makeLiquidityKeyword(keyword)),
+    ]);
+  };
+
+  const getKeywordsForBrand = brand => {
```
turadg: single keyword
```suggestion
  const getKeywordForBrand = brand => {
```

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+    return 'added Collateral to the Reserve';
+  };
+
+  const makeAddCollateralInvitation = () =>
+    zcf.makeInvitation(addCollateralHook, 'Add Collateal');
+
+  /** @type {ZCFMint} */
+  const runMint = await zcf.registerFeeMint('RUN', feeMintAccess);
+
+  const addLiquidityToAmmPool = async (collateralAmount, runAmount) => {
```
turadg, line 126: From "add" in method name I thought the arguments were passing in an asset. Reading through I understand that the requested `collateralAmount` has to be moved from the reserve's `collateralSeat`. I think "move" might lead the reader to this faster.

Though it's not really moving liquidity… `provideLiquidity`? 

Or if "liquidity" is implied by `AmmPool` so maybe `fundAmmPool`.

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+    ) {
+      throw new Error('insufficient reserves for that transaction');
+    }
+
+    // create the RUN
+    const offerToSeat = runMint.mintGains(harden({ RUN: runAmount }));
+    offerToSeat.incrementBy(
+      collateralSeat.decrementBy(
+        harden({
+          [keyword]: collateralAmount,
```
turadg: to disambiguate in the context consider
```suggestion
          [collateralBrand]: collateralAmount,
```

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+    const proposal = harden({
+      give: {
+        Central: runAmount,
+        Secondary: collateralAmount,
+      },
+      want: { Liquidity: AmountMath.makeEmpty(liqBrand) },
+    });
+
+    await E.get(offerTo(zcf, invitation, mapping, proposal, offerToSeat))
+      .deposited;
```
turadg, line 169: fyi: first read I thought `deposited` was a record and I thought `await E(offerTo(zcf, invitation, mapping, proposal, offerToSeat))` would work but I see it's a promise for what's deposited.

might be worth a clarifying comment,
```
// chain await the offer and its deposit completing

File: packages/run-protocol/src/reserve/collateralReserve.js [hunk 18]
```diff
@@ -0,0 +1,218 @@
...
+
+  // Anyone can deposit collateral to the reserve.
+  const addCollateralHook = async seat => {
+    const {
+      give: { Collateral: amountIn },
+    } = seat.getProposal();
+
+    const keyword = getKeywordsForBrand(amountIn.brand);
+    seat.decrementBy(harden({ Collateral: amountIn }));
+    collateralSeat.incrementBy(harden({ [keyword]: amountIn }));
```
turadg: surprised to see a generic `keyword`. down below the `Liquidity / liquiditKeyword` pairing helped. consider `collateralKeyword`.

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+        [liquidityKeyword]: liquidityAmount.Liquidity,
+      }),
+    );
+    zcf.reallocate(offerToSeat, collateralSeat);
+  };
+
+  const creatorFacet = wrapCreatorFacet(
+    {
+      makeAddCollateralInvitation,
+      // TODO: makeRedeemLiquidityTokensInvitation,
```
turadg: why todo and why not now?

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+      getAllocations,
+      addIssuer,
+    },
+    { addLiquidityToAmmPool },
+  );
+
+  /** @typedef {typeof creatorFacet} ReserveCreatorFacet */
+  const publicFacet = wrapPublicFacet(
+    Far('Collateal Reserve public', {
```
turadg:
```suggestion
    Far('Collateral Reserve public', {
```

packages/run-protocol/src/reserve/collateralReserve.js
```diff
@@ -0,0 +1,218 @@
...
+      makeAddCollateralInvitation,
+      // getAllocations,  perhaps this can't be public without a fee?
```
turadg: because expensive to compute? yeah

we could drop this line and assume it will be handled by a non-authoritative off-chain analytics capability

packages/run-protocol/src/reserve/params.js
```diff
@@ -0,0 +1,41 @@
+// @ts-check
+
+import {
+  CONTRACT_ELECTORATE,
+  makeParamManager as makeBaseParamManager,
+  ParamTypes,
+} from '@agoric/governance';
+
+export const AMM_INSTANCE_KEY = 'AmmInstance';
```
turadg:
```suggestion
export const AMM_INSTANCE = 'AmmInstance';
```

packages/run-protocol/src/reserve/params.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @param {ERef<ZoeService>} zoe
+ * @param {Instance} ammInstance
+ * @param {Invitation} poserInvitation - invitation for the question poser
+ */
+const makeParamManager = async (zoe, ammInstance, poserInvitation) => {
+  return makeBaseParamManager(
```
turadg: non-blocking: let `makeParamManager` be the general one
```suggestion
const makeReserveParamManager = async (zoe, ammInstance, poserInvitation) => {
  return makeParamManager(
```

Either way please make the exports either both have `Reserve` or neither.

packages/run-protocol/src/reserve/params.js
```diff
@@ -0,0 +1,41 @@
...
+const makeReserveTerms = (poserInvitationAmount, ammInstance, timer) => ({
+  timer,
+  main: harden({
+    [AMM_INSTANCE_KEY]: { type: ParamTypes.INSTANCE, value: ammInstance },
+    [CONTRACT_ELECTORATE]: {
+      type: ParamTypes.INVITATION,
+      value: poserInvitationAmount,
+    },
+  }),
+  governedApis: ['addLiquidityToAmmPool'],
```
turadg: since "governed apis" is used elsewhere to mean the callable methods themselves, consider `governedApiNames` or `governedMethodNames`

turadg (commented):
Have now looked through all the code except tests.

packages/run-protocol/src/econ-behaviors.js
Chris-Hibbert: copied from line 96.  I dunno.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done. 

I we had had imputed types earlier, I might have added the doc on the individual methods. It doesn't seem worth distributing them now, though not having to keep the type declaration synced may make it worthwhile.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: An invitation is a particular kind of Zoe Payment. Its amount contains all the invitation details. This is verifying that the invitation in privateArgs (an actual payment) represents the same seat in the same contract as the invitation amount (just a description) in the terms.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: It's using the Amm's `AddLiquidity` API.  I'll add a comment on the method.

Argh! It won't let me add a comment without overriding the type declaration.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: I changed it to `[collateralKeyword]`.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: YAGNI.  It hasn't been prioritized yet.

I added a TODO to remind readers that the liquidity tokens are an assets that's stored but not managed.  I'lll convert it to more explicit prose.

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/collateralReserve.js
Chris-Hibbert: "expensive" is too strong. Since it's non-negligible to compute it, we can't allow the general public to call it without limit.  

But you're right that we can now presume off-chain will handle it.

packages/run-protocol/src/reserve/params.js
Chris-Hibbert: Done

packages/run-protocol/src/reserve/params.js
Chris-Hibbert: It's now added automatically (named `getGovernedApiNames`

turadg (approved):

PR #4575, opened 2022-02-17
Execute vat creation and root object initialization in a crank

FUDCo:
Run vat creation and initialization in a crank, so module initialization and `buildRootObject` have access to syscalls and have their actions logged in the vat transcript.

Closes #2910

Issue #4542, opened 2022-02-12
syscall.send to resolved promise should run-queue to object, not to promise


warner:
## What is the Problem Being Solved?

While discussing #3465 with @mhofman yesterday, our plan (kobjs have priority, promise *subscriptions* have priority, deliveries have priority of their `send` target or `notify` subscription, vats can `syscall.elevate(koids)` exported objects up to the current delivery priority, `syscall.subscribe` inherits the current delivery priority) makes it awkward to have run-queue events that send a message to a *promise*, rather than an object.

Currently, `syscall.send(target=object)` and `syscall.send(target=promise)` do exactly the same thing: they push a `send` event onto the run-queue with the given target. When this event reaches the front of the run-queue, `deliverToTarget(objectOrPromise)` looks at the target to see whether it's an object or a promise. If it's a promise, it looks at the status. If the promise is resolved to an object, the code delegates (recurses just once) to `deliverToTarget(object)`. If the promise resolved to a non-object or rejected, it rejects the result promise. And if the promise is unresolved, the message is pushed onto the kernel promise table's `pendingMessages` queue.

Later, if/when the promise is resolved, all those `pendingMessages` are pushed onto the back of the run-queue (unmodified, still targeting the promise), and they make a second journey through the queue. When they reach the front, they are handed to the same `deliverToTarget` as before, but this time the promise-status check reports "resolved (to object)", and the messages are delivered to a vat object.

I'm pretty sure this extra loop is unnecessary and bad for performance. I remember feeling lazy about it when I implemented it that way, but not being sure of what ordering properties were involved, so I've left it alone until now.

But more importantly, if the run-queue can hold messages aimed at a Promise, we don't have a story for what priority those messages should have (in this new plan). It would be restrict the run-queue to messages aimed at just objects.

## Description of the Design

So the plan would be:
* `syscall.send()` looks at the target immediately, and if it is a promise, check the resolution status:
  * if resolved-to-object, retrieve the `koid`, and push a run-queue event *to that object*
  * if resolved-to-non-object or rejected, resolve the result promise immediately
    * (note: this will happen before the vat has a change to `syscall.subscribe()`, so make sure the now-rejected promise doesn't get deleted or collected or anything too early, and make sure subscribe-after-resolve works well)
  * if unresolved, push the message onto the promise table's `pendingMessages` queue, not the run-queue

I need to think about what ordering properties the old approach had, and how this new approach might be different, and if anything could/should care.

## Security Considerations

[...]

Issue #3465, opened 2021-07-12
Per vat priority (was: two run queues)


warner:
## What is the Problem Being Solved?

As a baby step towards #23, we're going to start with a simple two-level (high-priority + low-priority) scheduler. Combined with per-vat queues (https://github.com/Agoric/agoric-sdk/issues/5025), this should provide a starting point to enable high priority processing of economy elements (https://github.com/Agoric/agoric-sdk/issues/4318).

## Description of the Design

Vats will be marked with a priority, either low or high.

The scheduler used by `controller.step()`/`controller.run()` will then process the active vats with high priority before considering vats with lower priority. To ensure that high priority processing doesn't starve low priority work, the scheduler may occasionally pick a low priority vat.

At first all vats will be created with the same priority. A later change will enable vats with different priorities.

warner:
In today's kernel meeting, we figured that a good starting point would be to have `vat-zoe`, the AMM vat, and the Treasury vat be marked "high priority", and have everything else use the default. (Maybe the oracle vat needs to be high-priority too). The most important activity to prioritize is topping up a Vault (to reduce the risk of default and liquidation).. effectively anything that adds liquidity to the system is worth prioritizing over anything else.

@dtribble pointed out that a lot of recent DEX rollouts have left users asking the question "what is the status of my trade?", and that this could provoke some interesting engineering goals. So one thought is:
* when the Mailbox device receives messages from the outside world, it also accepts a "transaction ID" argument
* we augment the run-queue to include a "txnID" next to each queued item
* when a delivery occurs, any syscalls it provokes are also tagged with this txnID (unless overridden by some future mechanism)
* we copy the run-queue into the merklized cosmos state vector, or at least the sorted list of txnIDs and maybe some notion of their scheduling properties
  * ideally enough information to let an outsider predict when the item will run
* wallets will remember the txnID of their signed cosmos message when they first post it
* clients could do an RPC query to fetch this list, and show users a sort of queue diagram, with their txnID-tagged entries close or far from the top
* users watch their deliveries bubble up in this diagram to get a sense for how long it will take until their trade is complete

Later, we'd give e.g. the AMM contract a way to distinguish between the ACK it sends back when the trade is complete (which runs on the user's txnID) from the price-change signal it sends to any subscribers (which, while caused by the user's action, is not really a part of it, and should be given a distinct txnID, or none at all).

[...]

PR #4638, opened 2022-02-22
Add acceptance queue

mhofman:
refs: #3465

Best reviewed commit by commit

## Description

This is a preliminary refactor splitting the run-queue in 2 before #3465. All `send`, `resolve` and `subscribe` syscalls, as well as kernel initiated deliveries are first enqueued to an "acceptance" queue, which is drained into the usual "run" queue in its own cranks.

The next step (follow-up PR) is to change the handling of the `resolve` and `subscribe` syscalls so that they are enqueued onto the acceptance queue first instead of mutating promise state immediately (except for reference counting). Processing those and `send` events from the acceptance queue would then be as follow:
- message `send` onto promises are enqueued onto the promise queue or run-queue, depending on the state of the promise, instead of going through the run-queue as-is (see #4542). `send` onto objects are simply moved to the run-queue
- `resolve` would queue notifications on the run-queue for subscribers, and move `send` from the promise queue to the run-queue
- `subscribe` would either add the vat to the promise subscriber list, or enqueue a notify back to the vat, depending on the promise state

The goal is to only have messages with a known destination vat pending on the run-queue, so that the run (aka delivery) and acceptance queues can be split per vat according to https://github.com/Agoric/agoric-sdk/issues/3465#issuecomment-1041052494

The `maybeProcessNextMessage` is awkward and a bit of a temporary hack. I'll rethink handling of messages once processing the acceptance queue does more than simply re-queueing the message.

This change also introduces an `emptyCrank` call to the run-policy. It was already possible to have empty cranks (that did no delivery before), e.g. in the case of a send queued on an unresolved promise. With acceptance queue processing this is now much more common.

This PR does not change anything to vat termination, which will still be processed at the end of the current crank. Currently the delivery to vat-admin informing of the vat termination is enqueued, like every other delivery generated from the kernel, onto the acceptance queue, thus after any pending syscall events generated by the exited vat. In the future with queues per vat, we can imagine that kernel initiated deliveries use their own acceptance queue, or are queued directly onto the target's queue.

This PR also contains a fix for a test inconsistency described in https://github.com/Agoric/agoric-sdk/pull/4575/files#r809781876

### Security Considerations

None

This change should only impact the number of cranks necessary to process messages, but nothing else.

### Testing Considerations

Tests updated to account for different amount of cranks, and the new state of the queue, but no specific tests were added since the existing tests seem to cover the updated code already.

### DIFF
--- a/packages/SwingSet/docs/run-policy.md
+++ b/packages/SwingSet/docs/run-policy.md
@@ -33,12 +33,15 @@ The kernel will invoke the following methods on the policy object (so all must e
 * `policy.vatCreated()`
 * `policy.crankComplete({ computrons })`
 * `policy.crankFailed()`
+* `policy.emptyCrank()`
 
 All methods should return `true` if the kernel should keep running, or `false` if it should stop.
 
 The `computrons` argument may be `undefined` (e.g. if the crank was delivered to a non-`xs worker`-based vat, such as the comms vat). The policy should probably treat this as equivalent to some "typical" number of computrons.
 
 `crankFailed` indicates the vat suffered an error during crank delivery, such as a metering fault, memory allocation fault, or fatal syscall. We do not currently have a way to measure the computron usage of failed cranks (many of the error cases are signaled by the worker process exiting with a distinctive status code, which does not give it an opportunity to report back detailed metering data). The run policy should assume the worst.
+
+`emptyCrank` indicates the kernel processed a queued messages which didn't result in a delivery.
 
 More arguments may be added in the future, such as:
 * `vatCreated:` the size of the source bundle
@@ -52,7 +55,7 @@ The run policy should be provided as the first argument to `controller.run()`. I
 
 ## Typical Run Policies
 
-A basic policy might simply limit the block to 100 cranks and two vat creations:
+A basic policy might simply limit the block to 100 cranks with deliveries and two vat creations:
 
 ```js
 function make100CrankPolicy() {
@@ -70,6 +73,9 @@ function make100CrankPolicy() {
     crankFailed() {
       cranks += 1;
       return (cranks < 100);
+    },
+    emptyCrank() {
+      return true;
     },
   });
 }
@@ -109,6 +115,9 @@ function makeComputronCounterPolicy(limit) {
       total += 1000000; // who knows, 1M is as good as anything
       return (total < limit);
     },
+    emptyCrank() {
+      return true;
+    }
   });
 }
 ```
@@ -126,6 +135,7 @@ function makeWallclockPolicy(seconds) {
     vatCreated: () => Date.now() < timeout,
     crankComplete: () => Date.now() < timeout,
     crankFailed: () => Date.now() < timeout,
+    emptyCrank: () => Date.now() < timeout,
   });
 }
 ```
--- a/packages/SwingSet/src/kernel/initializeKernel.js
+++ b/packages/SwingSet/src/kernel/initializeKernel.js
@@ -89,7 +89,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
       vatKeeper.setSourceAndOptions({ bundleID }, creationOptions);
       vatKeeper.initializeReapCountdown(creationOptions.reapInterval);
       if (!creationOptions.enableSetup) {
-        kernelKeeper.addToRunQueue(harden({ type: 'startVat', vatID }));
+        kernelKeeper.addToAcceptanceQueue(harden({ type: 'startVat', vatID }));
       }
       if (name === 'vatAdmin') {
         // Create a kref for the vatAdmin root, so the kernel can tell it
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1192,6 +1192,7 @@ export default function buildKernel(
           policyOutput = policy.crankFailed(policyInput[1]);
           break;
         case 'none':
+          policyOutput = policy.emptyCrank();
           break;
         default:
           assert.fail(`unknown policyInput type in ${policyInput}`);
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -150,9 +150,10 @@ export default function buildKernel(
     return kernelSlog.vatConsole(vatID, origConsole);
   }
 
-  // runQueue entries are {type, vatID, more..}. 'more' depends on type:
-  // * deliver: target, msg
-  // * notify: kernelPromiseID
+  // runQueue and acceptanceQueue entries are {type, more..}. 'more' depends on type:
+  // * send: target, msg
+  // * notify: vatID, kpid
+  // * create-vat: vatID, source, dynamicOptions
 
   // in the kernel table, promises and resolvers are both indexed by the same
   // value. kernelPromises[promiseID] = { decider, subscribers }
@@ -1113,7 +1114,7 @@ export default function buildKernel(
     }
 
     if (!kernelKeeper.isRunQueueEmpty()) {
-      return kernelKeeper.getNextMsg();
+      return kernelKeeper.getNextRunQueueMsg();
     }
     return undefined;
   }
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -732,15 +732,8 @@ export default function buildKernel(
 
   const gcMessages = ['dropExports', 'retireExports', 'retireImports'];
 
-  function processAcceptanceQueue() {
-    while (!kernelKeeper.isAcceptanceQueueEmpty()) {
-      const acceptanceMessage = kernelKeeper.getNextAcceptanceQueueMsg();
-      kernelKeeper.addToRunQueue(acceptanceMessage);
-    }
-  }
-
   let processQueueRunning;
-  async function processQueueMessage(message) {
+  async function processDeliveryMessage(message) {
     kdebug(`processQ ${JSON.stringify(message)}`);
     kdebug(legibilizeMessage(message));
     if (processQueueRunning) {
@@ -804,7 +797,7 @@ export default function buildKernel(
             // delivered again on the next crank.  If we don't want that, then
             // we need to remove it again.
             // eslint-disable-next-line no-use-before-define
-            getNextMessage();
+            getNextDeliveryMessage();
           }
         }
         // state changes reflecting the termination must also survive, so
@@ -817,6 +810,38 @@ export default function buildKernel(
         // eslint-disable-next-line no-use-before-define
         await vatWarehouse.maybeSaveSnapshot();
       }
+      kernelKeeper.processRefcounts();
+      kernelKeeper.saveStats();
+      const crankNum = kernelKeeper.getCrankNumber();
+      kernelKeeper.incrementCrankNumber();
+      const { crankhash, activityhash } = kernelKeeper.commitCrank();
+      kernelSlog.write({
+        type: 'crank-finish',
+        crankNum,
+        crankhash,
+        activityhash,
+      });
+    } finally {
+      processQueueRunning = undefined;
+    }
+    return harden(policyInput);
+  }
+
+  async function processAcceptanceMessage(message) {
+    kdebug(`processAcceptanceQ ${JSON.stringify(message)}`);
+    kdebug(legibilizeMessage(message));
+    if (processQueueRunning) {
+      console.error(`We're currently already running at`, processQueueRunning);
+      assert.fail(X`Kernel reentrancy is forbidden`);
+    }
+    kernelSlog.write({ type: 'crank-start', message });
+    /** @type { PolicyInput } */
+    const policyInput = ['none'];
+    try {
+      processQueueRunning = Error('here');
+
+      kernelKeeper.addToRunQueue(message);
+
       kernelKeeper.processRefcounts();
       kernelKeeper.saveStats();
       const crankNum = kernelKeeper.getCrankNumber();
@@ -1111,7 +1136,7 @@ export default function buildKernel(
     kernelKeeper.loadStats();
   }
 
-  function getNextMessage() {
+  function getNextDeliveryMessage() {
     const gcMessage = processNextGCAction(kernelKeeper);
     if (gcMessage) {
       return gcMessage;
@@ -1127,6 +1152,28 @@ export default function buildKernel(
     return undefined;
   }
 
+  function getNextAcceptanceMessage() {
+    if (!kernelKeeper.isAcceptanceQueueEmpty()) {
+      return kernelKeeper.getNextAcceptanceQueueMsg();
+    }
+    return undefined;
+  }
+
+  function startProcessingNextMessageIfAny() {
+    /** @type {Promise<PolicyInput> | undefined} */
+    let resultPromise;
+    let message = getNextAcceptanceMessage();
+    if (message) {
+      resultPromise = processAcceptanceMessage(message);
+    } else {
+      message = getNextDeliveryMessage();
+      if (message) {
+        resultPromise = processDeliveryMessage(message);
+      }
+    }
+    return { resultPromise };
+  }
+
   async function step() {
     if (kernelPanic) {
       throw kernelPanic;
@@ -1134,11 +1181,10 @@ export default function buildKernel(
     if (!started) {
       throw new Error('must do kernel.start() before step()');
     }
-    processAcceptanceQueue();
+    const { resultPromise } = startProcessingNextMessageIfAny();
     // process a single message
-    const message = getNextMessage();
-    if (message) {
-      await processQueueMessage(message);
+    if (resultPromise) {
+      await resultPromise;
       if (kernelPanic) {
         throw kernelPanic;
       }
@@ -1167,15 +1213,14 @@ export default function buildKernel(
     }
     let count = 0;
     for (;;) {
-      processAcceptanceQueue();
-      const message = getNextMessage();
-      if (!message) {
+      const { resultPromise } = startProcessingNextMessageIfAny();
+      if (!resultPromise) {
         break;
       }
       count += 1;
       /** @type { PolicyInput } */
       // eslint-disable-next-line no-await-in-loop
-      const policyInput = await processQueueMessage(message);
+      const policyInput = await resultPromise;
       if (kernelPanic) {
         throw kernelPanic;
       }
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -783,7 +783,6 @@ export default function buildKernel(
       } else {
         assert.fail(X`unable to process message.type ${message.type}`);
       }
-      processAcceptanceQueue();
       let didAbort = false;
       if (terminationTrigger) {
         // the vat is doomed, either voluntarily or from meter/syscall fault
@@ -811,7 +810,6 @@ export default function buildKernel(
         // state changes reflecting the termination must also survive, so
         // these happen after a possible abortCrank()
         terminateVat(vatID, shouldReject, info);
-        processAcceptanceQueue();
         kernelSlog.terminateVat(vatID, shouldReject, info);
         kdebug(`vat terminated: ${JSON.stringify(info)}`);
       }
@@ -1167,9 +1165,9 @@ export default function buildKernel(
     if (!started) {
       throw new Error('must do kernel.start() before run()');
     }
-    processAcceptanceQueue();
     let count = 0;
     for (;;) {
+      processAcceptanceQueue();
       const message = getNextMessage();
       if (!message) {
         break;
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -150,6 +150,7 @@ export default function buildKernel(
     return kernelSlog.vatConsole(vatID, origConsole);
   }
 
+  // message flow: vat -> syscall -> acceptanceQueue -> runQueue -> delivery -> vat
   // runQueue and acceptanceQueue entries are {type, more..}. 'more' depends on type:
   // * send: target, msg
   // * notify: vatID, kpid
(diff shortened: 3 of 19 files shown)


packages/SwingSet/src/kernel/initializeKernel.js
```diff
@@ -89,7 +89,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
       vatKeeper.setSourceAndOptions({ bundleID }, creationOptions);
       vatKeeper.initializeReapCountdown(creationOptions.reapInterval);
       if (!creationOptions.enableSetup) {
-        kernelKeeper.addToRunQueue(harden({ type: 'startVat', vatID }));
+        kernelKeeper.addToSyscallQueue(harden({ type: 'startVat', vatID }));
```
FUDCo: I totally don't understand the concept of "syscall queue", but maybe it's just a really bad name for whatever it actually is.  The run queue is a queue of things that go into vats, i.e., things that will result in deliveries, or things that are done _to_ vats..  Syscalls are things that come out of a vat or done _by_ a vat.  Syscalls are also synchronous, so the idea of a queue involving them makes no sense to me.

FUDCo (changes requested):
I can't make heads or tails of this with something called "syscall queue".  I tried just treating it as an arbitrary, meaningless label like "cromfelter" but the cognitive interference with the actual words overwhelmed me and I just couldn't pull it off.  Please either rename the abstraction or provide a deeper analysis to help me understand why that name makes any kind of sense.  In particular, I don't understand the demarcation that the second queue is trying to capture.

mhofman:
Yes I struggle with the naming as well, but couldn't come up with anything better. Suggestions welcome.

In the discussion with @dtribble and @warner, and summarized in https://github.com/Agoric/agoric-sdk/issues/3465#issuecomment-1041052494, we were talking about inbound and outbound queues. However if you don't know the reference point, that terminology is even more confusing (the vat was the reference point, so inbound were pending message deliveries into the vat, and outbound were pending messages generated by the vat).

When I looked at syscalls, I realized there are conceptually 2 kinds of syscalls:
- Synchronous operations which affect the state of a data structure for the usage of the vat, and potentially return a value dependent on previous synchronous syscalls. E.g. store and device operations.
- Asynchronous operations which only affect data structures internal to the kernel, and for which the vat doesn't get an immediate result, or cannot sense the effect until another crank. E.g. `send`, `resolve`, `subscribe`.

The idea is to put all the asynchronous syscall messages as-is into a queue (to be named). The kernel then pops those messages from the queue and routes or process them appropriately. For a message send, the message may be put on the delivery queue or to a promise queue, depending on the type and state of the target. For a resolve, the promise state would be updated and notifications queued for delivery to subscribers. For a subscribe, either the subscribing vat would be added to the list of subscribers, or get a notification queued back, depending on the state of the promise.

For a single pair of queues, like in this PR, this does not really[^1] change the order of message delivery. However once we switch to different queues per vat (https://github.com/Agoric/agoric-sdk/issues/3465), and the kernel is free to process messages from vats independently from each other, then this allows limiting the effect a delivery has on the overall system, since asynchronous activity from a vat is queued and processed one at a time.

I actually considered also renaming the current `runQueue` as `deliveryQueue` to highlight it should (ultimately) only contain messages for a known target vat. That doesn't help with the naming of the other queue I currently named `syscallQueue`, but maybe "sorting" or "routing" queue would be better? In my mind, the delivery and syscall/sorting/routing queue collectively form the run queues.

[^1]: Given that message send on promises currently go directly onto the runqueue before being processed, moving to an early queuing onto the promise queue for unresolved promise could delay the actual delivery since the message would be moved to the delivery queue only once the promise is resolved (and the target vat known).

FUDCo:
"Routing queue" is better, especially when paired with "delivery queue", but possibly we can do even better.  That pairing at least hints at a functional demarcation.  I remain a little concerned that whatever thought process generated the term "syscall queue" in the first place was at least partially confused about the conceptualization behind this and therefore what this code does might also be.  I know that *I'm* certainly confused about it.  You also used the phrase "syscall messages" in your comment, which makes no sense at all to me and suggests that there is something foundational here that is wrong.  Probably this is a worthy agenda item for our next kernel meeting.

erights:
> I can't make heads or tails of this with something called "syscall queue". I tried just treating it as an arbitrary, meaningless label like "cromfelter" but the cognitive interference with the actual words overwhelmed me and I just couldn't pull it off. Please either rename the abstraction or provide a deeper analysis to help me understand why that name makes any kind of sense. In particular, I don't understand the demarcation that the second queue is trying to capture.

Btw, I saw this and wondered what a "cromfelter" might actually be. So I googled and the first hit was http://habitatchronicles.com/2018/01/pool-abuse/ . Now I know ;)

mhofman:
Another issue arising is what to do regarding the `exit` syscall, in particular if `send` or `resolve` calls were made in the same crank.

In the case of a clean exit, I'm imagining we could process these previous syscalls first (not what this PR does), so can afford to delay the termination until a later crank. However in case of a failure, I'm not sure if we need the termination to be recorded in the same crank. I assume as long as all pending "syscalls" are processed, it doesn't matter if the vat is terminated or not?

mhofman:
Renamed `syscallQueue` -> `acceptanceQueue`

warner:
> Another issue arising is what to do regarding the `exit` syscall, in particular if `send` or `resolve` calls were made in the same crank.

`vatPowers.exit` aborts the current delivery, unwinding all state changes, resetting to a world in which the vat has spontaneously terminated just before the delivery would have been made. The next thing the kernel does (with the current single run-queue) is to attempt the delivery again, which splats against the dead vat and gets rejected, precisely like any subsequent messages which are headed towards that same vat.

With a distinct vat-inbound queue, we can do the same unwinding, but then reject everything in the inbound queue immediately. No need to wait for a redelivery, which (given multiple queues) could take an unknown amount of time to be serviced.

The real question is what about messages that are already in the vat-outbound queue. The vat observed them to be sent successfully, and has no control over how the outbound queue is serviced, so if the kernel deletes the queue upon termination, basically some random portion of the sent message would get deleted. I don't see how programmers could reason about that.

But the alternative is to keep the vat-outbound queue around (occasionally servicing the remaining messages) even though the vat itself is terminated, which is pretty awkward. The vat lifecycle would grow from "non-existent -> active -> non-existent" to "non-existent -> active -> lingering-death -> non-existent". Maybe if we track vat-outbound queues in one data structure, and vats themselves in a separate one, it wouldn't be too hard to manage deleting the vat but retaining the queue until it is empty (and then deleting it).

mhofman:
> `vatPowers.exit` aborts the current delivery, unwinding all state changes, resetting to a world in which the vat has spontaneously terminated just before the delivery would have been made.

That is not my reading of https://github.com/Agoric/agoric-sdk/blob/a1fcbebffe029142f6cfcff47849fe1343185300/packages/SwingSet/src/kernel/kernel.js#L309-L318

FUDCo:
> > `vatPowers.exit` aborts the current delivery, unwinding all state changes, resetting to a world in which the vat has spontaneously terminated just before the delivery would have been made.
> 
> That is not my reading of
> 
> https://github.com/Agoric/agoric-sdk/blob/a1fcbebffe029142f6cfcff47849fe1343185300/packages/SwingSet/src/kernel/kernel.js#L309-L318

I think @mhofman is right on this.  A non-aborting exit shuts down the vat after the state changes made during the crank are committed.  This means that although the state of the vat itself goes away, any consequences to the kernel data structures (notably the runQueue) are retained, i.e., messages that were sent are actually sent.

However, I don't know that this introduces any new complications above and beyond the (non-trivial!) complications that have already been raised.

warner:
Ah, ok, you're right, `kernelSyscall.js exit()` calls it with `shouldAbortCrank = false`. Vats doing `vatPowers.exit` are asking to self-terminate cleanly, committing all state changes from their current delivery. I think they should expect that all previously-sent messages (from earlier cranks) will be delivered, which means their output queue should outlive the vat.

Vats which somehow perform an illegal operation (or simply an infinite loop) and get terminated non-cleanly should not expect that any messages they sent within the same delivery will be committed, and unwinding the state takes care of that. The question of whether previous-cranks' messages get deleted or eventually delivered remains, and should have the same answer as for `vatPowers.exit`.

The same is true for vats that are killed externally, by `adminNode~.terminateWithFailure()`, but the vat in question doesn't know exactly when their parent will invoke that, making it even more random-looking to them (or to their ghost, I guess).

packages/SwingSet/docs/run-policy.md
```diff
@@ -33,13 +33,16 @@ The kernel will invoke the following methods on the policy object (so all must e
...
 * `policy.crankFailed()`
+* `policy.emptyCrank()`
 
 All methods should return `true` if the kernel should keep running, or `false` if it should stop.
 
 The `computrons` argument may be `undefined` (e.g. if the crank was delivered to a non-`xs worker`-based vat, such as the comms vat). The policy should probably treat this as equivalent to some "typical" number of computrons.
 
 `crankFailed` indicates the vat suffered an error during crank delivery, such as a metering fault, memory allocation fault, or fatal syscall. We do not currently have a way to measure the computron usage of failed cranks (many of the error cases are signaled by the worker process exiting with a distinctive status code, which does not give it an opportunity to report back detailed metering data). The run policy should assume the worst.
 
+`emptyCrank` indicates the kernel processed a queued messages which didn't result in a delivery.
```
warner, line 44: Half of me wants to not tell the runPolicy about these. I don't know how the host application author should react: how many are too many? I guess a basic approach would be to say "no more than 200 `emptyCrank`s in a block". Would that protect against some sort of infinite cyclic loop or DoS attack?

Let's expand on the comment a bit, add another sentence or two about what situation can cause this (`syscall.send` to a rejected promise?). It seems like anything pulling from a vat-outbound-queue will never result in a delivery, so maybe this happens like 50% of the time, and if so the host app author should set their threshold a lot higher.

Hm, maybe we need to rename `policy.crankComplete` to `deliveryComplete`, change `crankFailed` to `deliveryFailed`, and then add a `crankHappened` which means the kernel did work but not necessarily a vat.

warner: But, I'm ok with this PR landing without that change, if we make a ticket to improve it later.

packages/SwingSet/src/kernel/initializeKernel.js
```diff
@@ -89,7 +89,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
       vatKeeper.setSourceAndOptions({ bundleID }, creationOptions);
       vatKeeper.initializeReapCountdown(creationOptions.reapInterval);
       if (!creationOptions.enableSetup) {
-        kernelKeeper.addToRunQueue(harden({ type: 'startVat', vatID }));
+        kernelKeeper.addToAcceptanceQueue(harden({ type: 'startVat', vatID }));
```
warner, line 92: If I understand correctly that "acceptance queue" is the vat-outbound-queue (populated by vat syscalls), then I'd think `startVat` goes on the other one (the vat-inbound queue).

Also, it's critical that `startVat` gets delivered to a vat before anything else is delivered to that vat, regardless of whatever fairness sampling we might implement later, otherwise messages could arrive at an incomplete vat. Putting `startVat` on the vat-inbound queue should guarantee this.

packages/SwingSet/tools/vat.js
```diff
@@ -45,6 +45,8 @@ async function main() {
       deepLog(d.promises);
       console.log('Run Queue:');
       deepLog(d.runQueue);
+      console.log('Acceptance Queue:');
```
warner, line 48: Oh, how nostalgic, I'd completely forgotten about this file. Some day I look forward to defining a `/usr/bin/vat` in the way this originally aspired to. But I bet Endo will beat me to it by a mile.

packages/SwingSet/test/test-kernel.js
```diff
@@ -303,13 +303,17 @@ test('outbound call', async t => {
     },
   ]);
 
+  // Move the send to the run-queue
+  await kernel.step();
+  // Deliver the send
```
warner, line 323: We should test the new functionality here: after the `step()` that moves the send from the acceptance queue to the run-queue, perform a `kernel.dump()` and check that the message now exists in the runQueue (i.e. make a copy of the original code, using `.runQueue`, here). We only need to test this once.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -150,9 +150,10 @@ export default function buildKernel(
     return kernelSlog.vatConsole(vatID, origConsole);
   }
 
-  // runQueue entries are {type, vatID, more..}. 'more' depends on type:
-  // * deliver: target, msg
-  // * notify: kernelPromiseID
+  // runQueue and acceptanceQueue entries are {type, more..}. 'more' depends on type:
```
warner, line 154: This might be a good place for a note on which queue is which.

```
// message flow: vat -> syscall -> acceptanceQueue -> runQueue -> delivery -> vat
```

mhofman:
Right, I think the summary is that this PR changes nothing to vat termination since the acceptance queue is simply the tail of the previous run-queue. 

Vat exit will only become a question once we hold a queue pair per vat. At that point, to preserve existing delivery guarantees (message sent in previous cranks get delivered regardless if vat terminates in a later crank), the vat outbound queue (whatever has been committed to) needs to get drained after the vat is marked as terminated.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1121,22 +1154,44 @@ export default function buildKernel(
...
+      return kernelKeeper.getNextAcceptanceQueueMsg();
+    }
+    return undefined;
+  }
+
+  function maybeProcessNextMessage() {
+    /** @type {Promise<PolicyInput> | undefined} */
+    let resultPromise;
+    let message = getNextAcceptanceMessage();
+    if (message) {
```
warner, line 1166: nit: I'd prefer a short-circuiting style:
```js
let message = getNextAcceptanceMessage();
if (message) {
  return processAcceptanceMessage(message);
}
message = getNextDeliveryMessage();
if (message) {
  return processAcceptanceMessage(message);
}
```

But the fact that you're returning a record, not just a promise, tells me that you've got something in mind for this, so ignore me if your plans don't align with my aesthetics.

BTW the return value of `getNextMessage` has an extra significance that needs to be kept in mind (it looks like your code does the right thing, but I wanted to be explicit about the requirement). If there's no work to be done, we don't commit any state changes. If e.g. `maybeProcessNextMessage` pulled something off a queue and *then* decided that we didn't need to do anything, we'd have a problem, because now there are outstanding state changes that need to be committed. If the kernel doesn't `commitCrank` before returning control to the host app, we're going to have consistency problems (sometimes the pop-from-queue changes get committed, sometimes they don't, depending upon what happens next time).

I think you're headed this way, but we should aim for a pattern in which crank commits happen at the end of every *crank*, regardless of whether that crank includes a delivery or not. When deliveries are unwound, we unwind the entire crank buffer, and we need to make sure that we still pop off whatever queue entry started the process. The `getNextMessage` call the `discardFailedDelivery` branch of `processQueueMessage` does this, and we need to make sure it still does the right thing no matter what queue it came from.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1013,7 +1046,7 @@ export default function buildKernel(
         const source = { bundle };
         const vatID = kernelKeeper.allocateUnusedVatID();
         const event = { type: 'create-vat', vatID, source, dynamicOptions };
-        kernelKeeper.addToRunQueue(harden(event));
+        kernelKeeper.addToAcceptanceQueue(harden(event));
```
warner, line 1053: Hm.. FYI, this call only happens while the `vatAdmin` vat is running, because it's (normally) the only one with access to `devices.vatAdmin`, and this is an endowment provided only to the device. Putting it on the acceptance queue kinda claims to rate-limit the creation of new vats, but that's already limited by the ability to deliver `vatAdminService~.createVat()` deliveries into vatAdmin.

I think it might be better to drop this on the other queue. Or to remember (maybe add a note) that this wants to go onto its own queue (not associated with any particular vat). The `create-vat` events are not vat syscalls (even though they're triggered by a `syscall.callNow(vatAdminDevice)` from vatAdminVat), and `devices.vatAdmin` does not know which vat is asking it to use its kernel endowment, and we're not considering having device-outbound queues to match the vat-outbound queues.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -825,6 +826,38 @@ export default function buildKernel(
...
+    }
+    kernelSlog.write({ type: 'crank-start', message });
+    /** @type { PolicyInput } */
+    const policyInput = ['none'];
+    try {
+      processQueueRunning = Error('here');
+
+      kernelKeeper.addToRunQueue(message);
+
+      kernelKeeper.processRefcounts();
```
warner, line 845: my feeling is that we could skip this `processRefcounts` because nothing could change them during a simple queue-to-queue transfer, but it's safer to leave it in, and in the future the transfer function might cause promises to be rejected or messages to go splat, and those certainly *would* change refcounts

warner (approved):
Looks good, small changes suggested (and slightly-expanded test requested).

packages/SwingSet/src/kernel/initializeKernel.js
mhofman: Yes probably `startVat` could go straight to the `runQueue`. However I didn't want to change any message ordering in this PR, so I always enqueue everything at the end of the `acceptanceQueue` for now. Happy to change the potential order already in this case.

packages/SwingSet/src/kernel/kernel.js
mhofman: > returning a record, not just a promise

In general I don't like have a return value of `Promise | undefined` as it would mean something is maybe async. I also prefer short circuiting, but in this case it didn't feel right.

I am planning on ripping this ugly function apart in the next PR, and didn't want to spend time making something that was right, but happy to revisit.

> I think you're headed this way, but we should aim for a pattern in which crank commits happen at the end of every _crank_, regardless of whether that crank includes a delivery or not

I believe that's already the case. The only time we don't create a crank is if there is no message at all. I will keep making sure that a crank is always created and committed or aborted if a message is popped from a queue.

packages/SwingSet/src/kernel/kernel.js
mhofman: > Or to remember (maybe add a note) that this wants to go onto its own queue (not associated with any particular vat).

Planning on revisiting all enqueuing in the next PR. As stated I'm trying not to change any order in this PR.

packages/SwingSet/src/kernel/kernel.js
mhofman: I am planning on revisiting refcounts in the next PR: currently popping a `send` from the runQueue decrements refcounts, and re-increments them when re-queueing on the promise queue. Obviously that's extra work we won't usually need when popping a `send` from the outbound / acceptance queue since it will likely move to another queue (except as you say if sending to an invalid promise).

packages/SwingSet/docs/run-policy.md
mhofman: I think being more explicit to which kind of crank happened would indeed be good. Probably makes sense to include as part of the next step when changing the processing of the acceptance queue.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1016,7 +1050,7 @@ export default function buildKernel(
         const source = { bundle };
         const vatID = kernelKeeper.allocateUnusedVatID();
         const event = { type: 'create-vat', vatID, source, dynamicOptions };
-        kernelKeeper.addToRunQueue(harden(event));
+        kernelKeeper.addToAcceptanceQueue(harden(event));
```
FUDCo, line 1053: I think I'm unclear on the demarcation criteria that determine which deliveries get initially put on the acceptanceQueue and which get put on the runQueue directly.  My intuition says that `'startVat'` should be one of the latter, but I'm not sure what standard I should be judging on.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1113,22 +1147,44 @@ export default function buildKernel(
...
   }
 
+  function getNextAcceptanceMessage() {
+    if (!kernelKeeper.isAcceptanceQueueEmpty()) {
+      return kernelKeeper.getNextAcceptanceQueueMsg();
+    }
+    return undefined;
+  }
+
+  function maybeProcessNextMessage() {
```
FUDCo: Not sure about the name.  I take the "maybe" here to mean "maybe there will be a message to process, maybe there won't, but if there is I will" rather than "maybe I'll process the next message, maybe I won't process it", but name hints at the latter.

packages/SwingSet/src/kernel/kernel.js
mhofman: In the future, I believe the standard should be based on where the message comes from, and if we know the destination. For kernel initiated events, we could imagine having an "outbound queue" for the kernel with high priority. Or we could bypass that and enqueue directly onto the target.

Regardless, it's a problem I'm punting on until the next PR, as here I'm taking the stance that everything goes through the acceptance queue first, so that this PR doesn't change any delivery order.

packages/SwingSet/src/kernel/kernel.js
mhofman: `processNextMessageIfAny` ?

packages/SwingSet/src/kernel/initializeKernel.js
FUDCo: > so I always enqueue everything at the end of the acceptanceQueue for now

Ah!  That totally makes sense and explains a large amount of my confusion.  Now I understand why I didn't understand what the demarcation criteria were.

packages/SwingSet/test/test-controller.js
```diff
@@ -118,7 +119,9 @@ test('bootstrap', async t => {
   const c = await buildVatController(config);
   for (let i = 0; i < 5; i += 1) {
     // eslint-disable-next-line no-await-in-loop
-    await c.step(); // vat starts
+    await c.step(); // vat start acceptance
+    // eslint-disable-next-line no-await-in-loop
+    await c.step(); // vat start deliver
```
FUDCo, line 124: @gibson042 has a PR #4678 that's about to land that gets rid of some of this fussy `c.step()` stuff, which should subsume some of the changes you've had to make here.

FUDCo (approved):
Now that I understand that passage through the acceptance queue is currently unconditional, this makes a lot more sense.  I'm looking forward to the next step in this process.

Issue #4415, opened 2022-01-28
Audit vaultFactory package for consistency


turadg:
## What is the Problem Being Solved?

vaultFactory has grown somewhat organically and it's time to take a look holistically to tie it all together so that it's easy to comprehend and maintain.

Examples of things to look into:
- Consistent patterns
-- invitations
- Contract modularity
-- AMM from vaultFactory to bootstrap

## Description of the Design

Output of ticket will be proposal for design changes.

## Security Considerations
n/a
## Test Plan

n/a

dckc:
I suggest closing this as a dup of #4199 (resolved in #4209). I don't see any new information that motivates reconsidering that decision.

turadg:
This was filed at the request of @dtribble in late January when he had some thoughts on things to revisit.

> I don't see any new information that motivates reconsidering that decision.

I don't think that decision is in doubt. However "new information" on whether to make changes before Mainnet 1 is actually the output of this work, not the input.

turadg:
Example to come back to: https://github.com/Agoric/agoric-sdk/commit/fcdd7a03747c8b5fedf7b97994e7a9c55989cdab

turadg:
Some observations from @erights https://github.com/Agoric/agoric-sdk/pull/4553/

Tartuffo:
Closing this because it seems to be done.

turadg:
Reopening because it's not done and it doesn't delay MN-1 because it's not scheduled for MN-1.

turadg:
Obsoleted with the sunsetting of Inter Protocol

PR #4640, opened 2022-02-22
feat(run-protocol): distinguish LIQUIDATED from CLOSED vault state

### LINKED ISSUE #4539: Ability for vault observers to know liquidated vs closed

The vault's notification stream has a `liquidated` flag that can be used to determine whether a vault was liquidated, but it currently is `state === CLOSED`. A user will want the ability to distinguish between vaults closed by them vs by liquidation.

That used to be achieved by `getLiquidationPromise()` from the vault, but https://github.com/Agoric/agoric-sdk/pull/4527 removed it because,
- it was much harder to implement in the redesign
- it created a second stream for notifications from the vault
- it wasn't used anywhere yet so handling that use case could be deferred (to this ticket)

## Description of the Design

Tentatively: add fourth vault state 'LIQUIDATED' and change the name of CLOSED to distinguish.

## Security Considerations
--
## Test Plan

--

turadg:
closes: #4539

## Description

- rename VaultState to VaultPhase in anticipation of virtual object refactor (in which vault "state" encompasses all state)
- make a distinct state for "liquidated"
- runtime validation of state transitions

### Security Considerations

--

--

### Testing Considerations

Tests updated.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -263,8 +263,6 @@ export const makeInnerVault = (
       debtSnapshot,
       locked: getCollateralAmount(),
       debt: getDebtAmount(),
-      // XXX uses closure value instead of argument
-      liquidated: phase === VaultPhase.LIQUIDATED,
       vaultState: newPhase,
     });
   };
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -28,18 +28,16 @@ const trace = makeTracer('Vault');
 // collateral, and lending RUN to the borrower
 
 /**
- * Constants for vault state.
+ * Constants for vault phase.
  *
- * @typedef {'active' | 'liquidating' | 'closed' | 'transfer'} VAULT_STATE
- *
- * @type {{ ACTIVE: 'active', LIQUIDATING: 'liquidating', CLOSED: 'closed', TRANSFER: 'transfer' }}
+ * @typedef {VaultPhase[keyof typeof VaultPhase]} VAULT_PHASE
  */
-export const VaultState = {
+export const VaultPhase = /** @type {const} */ ({
   ACTIVE: 'active',
   LIQUIDATING: 'liquidating',
   CLOSED: 'closed',
   TRANSFER: 'transfer',
-};
+});
 
 const makeOuterKit = inner => {
   const { updater: uiUpdater, notifier } = makeNotifierKit();
@@ -102,11 +100,11 @@ export const makeInnerVault = (
   const { zcfSeat: liquidationZcfSeat, userSeat: liquidationSeat } =
     zcf.makeEmptySeatKit(undefined);
 
-  /** @type {VAULT_STATE} */
-  let vaultState = VaultState.ACTIVE;
+  /** @type {VAULT_PHASE} */
+  let phase = VaultPhase.ACTIVE;
 
   const assertVaultIsOpen = () => {
-    assert(vaultState === VaultState.ACTIVE, X`vault must still be active`);
+    assert(phase === VaultPhase.ACTIVE, X`vault must still be active`);
   };
 
   let outerUpdater;
@@ -239,7 +237,7 @@ export const makeInnerVault = (
       : getCollateralAllocated(vaultSeat);
   };
 
-  const snapshotState = vstate => {
+  const snapshotState = newPhase => {
     /** @type {VaultUIState} */
     return harden({
       // TODO move manager state to a separate notifer https://github.com/Agoric/agoric-sdk/issues/4540
@@ -249,8 +247,9 @@ export const makeInnerVault = (
       locked: getCollateralAmount(),
       debt: getDebtAmount(),
       // TODO state distinct from CLOSED https://github.com/Agoric/agoric-sdk/issues/4539
-      liquidated: vaultState === VaultState.CLOSED,
-      vaultState: vstate,
+      // XXX uses closure value instead of argument
+      liquidated: phase === VaultPhase.CLOSED,
+      vaultState: newPhase,
     });
   };
 
@@ -260,26 +259,26 @@ export const makeInnerVault = (
       return;
     }
     /** @type {VaultUIState} */
-    const uiState = snapshotState(vaultState);
+    const uiState = snapshotState(phase);
     trace('updateUiState', uiState);
 
-    switch (vaultState) {
-      case VaultState.ACTIVE:
-      case VaultState.LIQUIDATING:
+    switch (phase) {
+      case VaultPhase.ACTIVE:
+      case VaultPhase.LIQUIDATING:
         outerUpdater.updateState(uiState);
         break;
-      case VaultState.CLOSED:
+      case VaultPhase.CLOSED:
         outerUpdater.finish(uiState);
         break;
       default:
-        throw Error(`unreachable vaultState: ${vaultState}`);
+        throw Error(`unreachable vaultState: ${phase}`);
     }
   };
   // XXX Echo notifications from the manager through all vaults
   // TODO move manager state to a separate notifer https://github.com/Agoric/agoric-sdk/issues/4540
   observeNotifier(managerNotifier, {
     updateState: () => {
-      if (vaultState !== VaultState.CLOSED) {
+      if (phase !== VaultPhase.CLOSED) {
         updateUiState();
       }
     },
@@ -293,15 +292,15 @@ export const makeInnerVault = (
   const liquidated = newDebt => {
     updateDebtSnapshot(newDebt);
 
-    vaultState = VaultState.CLOSED;
+    phase = VaultPhase.CLOSED;
     updateUiState();
   };
 
   const liquidating = () => {
-    if (vaultState === VaultState.LIQUIDATING) {
+    if (phase === VaultPhase.LIQUIDATING) {
       throw new Error('Vault already liquidating');
     }
-    vaultState = VaultState.LIQUIDATING;
+    phase = VaultPhase.LIQUIDATING;
     updateUiState();
   };
 
@@ -342,7 +341,7 @@ export const makeInnerVault = (
     runMint.burnLosses(harden({ RUN: currentDebt }), burnSeat);
     seat.exit();
     burnSeat.exit();
-    vaultState = VaultState.CLOSED;
+    phase = VaultPhase.CLOSED;
     updateDebtSnapshot(AmountMath.makeEmpty(runBrand));
     updateUiState();
 
@@ -670,7 +669,7 @@ export const makeInnerVault = (
     makeCloseInvitation,
     makeTransferInvitation: () => {
       if (outerUpdater) {
-        outerUpdater.finish(snapshotState(VaultState.TRANSFER));
+        outerUpdater.finish(snapshotState(VaultPhase.TRANSFER));
         outerUpdater = null;
       }
       return zcf.makeInvitation(makeTransferInvitationHook, 'TransferVault');
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -33,18 +33,9 @@ const trace = makeTracer('Vault');
  * ACTIVE       - vault is in use and can be changed
  * LIQUIDATING  - vault is being liquidated by the vault manager, and cannot be changed by the user
  * TRANSFER     - vault is released from the manager and able to be transferred
+ * TRANSFER     - vault is able to be transferred (payments and debits frozen until it has a new owner)
  * CLOSED       - vault was closed by the user and all assets have been paid out
  * LIQUIDATED   - vault was closed by the manager, with remaining assets paid to owner
- *
- * These are the valid state transitions:
- * Active -> Liquidating
- * Active -> Transferrable
- * Active -> Closed
- * Liquidating -> Liquidated
- * Transferrable -> Active
- * Transferrable -> Liquidating
- *
- * (Liquidated and Closed cannot be changed)
  *
  * @typedef {VaultPhase[keyof typeof VaultPhase]} VAULT_PHASE
  */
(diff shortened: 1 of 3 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -268,9 +300,8 @@ export const makeInnerVault = (
       locked: getCollateralAmount(),
       debt: getDebtAmount(),
       collateralizationRatio,
-      // TODO state distinct from CLOSED https://github.com/Agoric/agoric-sdk/issues/4539
-      liquidated: vaultState === VaultState.CLOSED,
-      vaultState: vstate,
+      liquidated: phase === VaultPhase.LIQUIDATED,
```
Chris-Hibbert: I suspect the UI actually wants to know "is the vault closed or liquidated?".  "closed" is a better term for that than "liquidated", but maybe the UI needs to distinguish `[ [Active], [Closed, Liquidated], [Transferring]]`.  I think `Liquidating` is most like `Closed` and `Liquidating` among those, but we should probably check with @samsiegart.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -285,26 +316,27 @@ export const makeInnerVault = (
...
-    const uiState = snapshotState(vaultState, collateralizationRatio);
+    const uiState = snapshotState(vaultPhase, collateralizationRatio);
     trace('updateUiState', uiState);
 
-    switch (vaultState) {
-      case VaultState.ACTIVE:
-      case VaultState.LIQUIDATING:
+    switch (vaultPhase) {
+      case VaultPhase.ACTIVE:
+      case VaultPhase.LIQUIDATING:
```
Chris-Hibbert, line 309: add `TRANSFERABLE` here.

If no tests broke, then there aren't any tests for transfering. Please add some tests in that case.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -29,17 +29,37 @@ const trace = makeTracer('Vault');
 // collateral, and lending RUN to the borrower
 
 /**
- * Constants for vault state.
+ * Constants for vault phase.
  *
- * @typedef {'active' | 'liquidating' | 'closed' | 'transfer'} VAULT_STATE
+ * ACTIVE       - vault is in use and can be changed
+ * LIQUIDATING  - vault is being liquidated by the vault manager, and cannot be changed by the user
+ * TRANSFERABLE - vault is released from the manager and able to be transferred
```
dtribble: What does "released from the manager" mean?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -103,11 +123,23 @@ export const makeInnerVault = (
...
 
   const assertVaultIsOpen = () => {
-    assert(vaultState === VaultState.ACTIVE, X`vault must still be active`);
+    assert(vaultPhase === VaultPhase.ACTIVE, X`vault must still be active`);
+  };
+
+  /**
+   * @param {VAULT_PHASE} newPhase
+   */
+  const assignPhase = newPhase => {
```
dtribble, line 133: If we want a state machine...then we have a library for it, and can potentially integrate that with the `enum` proposal

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -258,7 +290,7 @@ export const makeInnerVault = (
     return makeRatioFromAmounts(collateralValueInRun, getDebtAmount());
   };
 
-  const snapshotState = (vstate, collateralizationRatio) => {
+  const snapshotState = (phase, collateralizationRatio) => {
```
dtribble: not sure I like "phase" when it's really the state for a state machine

packages/run-protocol/src/vaultFactory/vault.js
dtribble: I like "transfering" better than "tranferable" I think

packages/run-protocol/src/vaultFactory/vault.js
dtribble: this should be named "isClosed"

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -285,26 +316,27 @@ export const makeInnerVault = (
     // [https://github.com/Agoric/dapp-token-economy/issues/123]
     const collateralizationRatio = await getCollateralizationRatio();
     /** @type {VaultUIState} */
-    const uiState = snapshotState(vaultState, collateralizationRatio);
+    const uiState = snapshotState(vaultPhase, collateralizationRatio);
```
dtribble: I'd use "phase" rather than "vaultPhase" if we are using it

packages/run-protocol/src/vaultFactory/vault.js
dtribble: There are tests for transferring. "Transfer" is only visible in `finish` event that is generated when the transfer is started

packages/run-protocol/src/vaultFactory/vault.js
turadg: I'm assuming the end user would want to know all of these states. I'd really like to remove this flag entirely since it's redundant with the richer option, but I was trying to maintain backwards compatibility for now. To that end, I suppose `CLOSED` achieves that more!

@samsiegart is this backwards compatibility necessary or can we drop the `liquidated` flag in this PR? If not now then when would be good?

packages/run-protocol/src/vaultFactory/vault.js
turadg: 👍 https://github.com/Agoric/agoric-sdk/pull/4640/commits/837381e6a078af878c7c4ba25eb14c851cf0d593

packages/run-protocol/src/vaultFactory/vault.js
turadg: This is a smell though that the transfer case is so special. Refactored in [29b997](https://github.com/Agoric/agoric-sdk/pull/4640/commits/29b9973948516b025bbbb905a2b8391cb5629af3)

packages/run-protocol/src/vaultFactory/vault.js
turadg: The motivation is for when we're on virtual objects. Writing `state.state = VaultState.ACTIVE` was worse :)

packages/run-protocol/src/vaultFactory/vault.js
turadg: Nice. I was considering a fuller state machine with transition functions and figured this was a good incremental step that wasn't entirely out of scope for the ticket :)

I'd like to hear more about the library we have and this `enum` proposal and consider refactoring this in the future to use them.

packages/run-protocol/src/vaultFactory/vault.js
turadg: "released from manager" was when I thought the vault freezes for transfer and can't be liquidated, but that's wrong so I'll fix this. What's the best way to describe this? "Able to be transferred (payments and debits frozen until it has a new owner)" ?

Regarding the term, "transferable" feels clunky but "transfering" isn't necessarily true, right? It could sit inert in a transferable state indefinitely. How about SEALED? It can be sealed for transfer and unsealed upon reception.

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: It's not much: `.../zoe/src/contractSupport/stateMachine.js`. It started approximately like this. Just enough for present purposes, with a hope for more when we need it.

turadg: Thanks for the link. I don't think that version is going to work well for state that need to be serialized into virtual objects because the `stateMachine` assumes the state is in its own closure and it's got to be in the `state` object.

We could isolate the actual state from the transition mechanics with a design like https://kentcdodds.com/blog/implementing-a-simple-state-machine-library-in-javascript that would allow:
```
state.phase = machine.transition(state.phase, 'liquidate')
```

However I think that's out of scope of this PR and the `assignState` method closes the gap to a full machine if that proves worthwhile.

packages/run-protocol/src/vaultFactory/vault.js
turadg: back to `TRANSFER` but I don't like that it's a verb and the rest are adjectives https://github.com/Agoric/agoric-sdk/pull/4640/commits/792c388943a743cc4fc4b0440298f99a34740d4b

packages/run-protocol/src/vaultFactory/vault.js
turadg: from convo today, decision was to drop this boolean and rely just on the phase/state value 797a9cffc684d8fd89d6dfda6353b088f358eb4d

packages/run-protocol/src/vaultFactory/vault.js
turadg: I hit some snags so intend to punt https://github.com/Agoric/agoric-sdk/commit/fcdd7a03747c8b5fedf7b97994e7a9c55989cdab until https://github.com/Agoric/agoric-sdk/issues/4415

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1657,15 +1657,15 @@ test('mutable liquidity triggers and interest', async t => {
   await waitForPromisesToSettle();
   aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
   bobUpdate = await E(bobNotifier).getUpdateSince();
-  t.truthy(aliceUpdate.value.liquidated);
+  t.is(aliceUpdate.value.vaultState, 'liquidated');
```
```suggestion
  t.is(aliceUpdate.value.vaultState, VaultPhase.LIQUIDATED);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1657,15 +1657,15 @@ test('mutable liquidity triggers and interest', async t => {
...
 
   for (let i = 0; i < 5; i += 1) {
     manualTimer.tick();
   }
   await waitForPromisesToSettle();
   bobUpdate = await E(bobNotifier).getUpdateSince();
 
-  t.truthy(bobUpdate.value.liquidated);
+  t.is(bobUpdate.value.vaultState, 'liquidated');
```
```suggestion
  t.is(bobUpdate.value.vaultState, VaultPhase.LIQUIDATED);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2159,13 +2159,13 @@ test('mutable liquidity triggers and interest sensitivity', async t => {
   await waitForPromisesToSettle();
   aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
   bobUpdate = await E(bobNotifier).getUpdateSince();
-  t.falsy(aliceUpdate.value.liquidated);
+  t.is(aliceUpdate.value.vaultState, 'active');
```
```suggestion
  t.is(aliceUpdate.value.vaultState, VaultPhase.ACTIVE);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2159,13 +2159,13 @@ test('mutable liquidity triggers and interest sensitivity', async t => {
...
 
   for (let i = 0; i < 5; i += 1) {
     manualTimer.tick();
   }
   await waitForPromisesToSettle();
   bobUpdate = await E(bobNotifier).getUpdateSince();
 
-  t.truthy(bobUpdate.value.liquidated);
+  t.is(bobUpdate.value.vaultState, 'liquidated');
```
```suggestion
  t.is(bobUpdate.value.vaultState, VaultPhase.LIQUIDATED);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -519,22 +519,22 @@ test('price drop', async t => {
...
   await manualTimer.tick();
   notification = await E(uiNotifier).getUpdateSince(notification.updateCount);
   t.falsy(notification.updateCount);
-  t.truthy(notification.value.liquidated);
+  t.is(notification.value.vaultState, 'liquidated');
 
   const debtAmountAfter = await E(vault).getDebtAmount();
   const finalNotification = await E(uiNotifier).getUpdateSince();
-  t.truthy(finalNotification.value.liquidated);
+  t.is(finalNotification.value.vaultState, 'liquidated');
```
```suggestion
  t.is(finalNotification.value.vaultState, VaultPhase.LIQUIDATED);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -519,22 +519,22 @@ test('price drop', async t => {
...
-  t.falsy(notification.value.liquidated);
-  t.is(notification.value.vaultState, VaultState.LIQUIDATING);
+  t.is(notification.value.vaultState, 'liquidating');
+  t.is(notification.value.vaultState, VaultPhase.LIQUIDATING);
 
   await manualTimer.tick();
   notification = await E(uiNotifier).getUpdateSince(notification.updateCount);
   t.falsy(notification.updateCount);
-  t.truthy(notification.value.liquidated);
+  t.is(notification.value.vaultState, 'liquidated');
```
```suggestion
  t.is(notification.value.vaultState, VaultPhase.LIQUIDATED);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -519,22 +519,22 @@ test('price drop', async t => {
...
   notification = await E(uiNotifier).getUpdateSince();
-  t.falsy(notification.value.liquidated);
+  t.is(notification.value.vaultState, 'active');
 
   await manualTimer.tick();
   notification = await E(uiNotifier).getUpdateSince(notification.updateCount);
   trace('price changed to liquidate', notification.value.vaultState);
-  t.falsy(notification.value.liquidated);
-  t.is(notification.value.vaultState, VaultState.LIQUIDATING);
+  t.is(notification.value.vaultState, 'liquidating');
```
```suggestion
  t.is(notification.value.vaultState, VaultPhase.LIQUIDATING);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -519,22 +519,22 @@ test('price drop', async t => {
   );
   await manualTimer.tick();
   notification = await E(uiNotifier).getUpdateSince();
-  t.falsy(notification.value.liquidated);
+  t.is(notification.value.vaultState, 'active');
```
```suggestion
  t.is(notification.value.vaultState, VaultPhase.ACTIVE);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -508,7 +508,7 @@ test('price drop', async t => {
   let notification = await E(uiNotifier).getUpdateSince();
   trace('got notificaation', notification);
 
-  t.falsy(notification.value.liquidated);
+  t.is(notification.value.vaultState, 'active');
```
```suggestion
  t.is(notification.value.vaultState, VaultPhase.ACTIVE);
```

Chris-Hibbert (approved):

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Agree it's better to use a constant instead of the literal. I'll define the constant in this test so that if the actual string VaultFactory returns changes this test will break.

PR #4641, opened 2022-02-22
feat(cosmic-swingset): grant addVaultType based on addr

### LINKED ISSUE #4634: production bootstrap does not provide trading assets needed by loadgen
## Describe the bug

The loadgen uses the `local-chain` / `local-solo` scenario; as of addressing #4165, this no longer provides the demo USDC that the loadgen expects to have available for trading.

## To Reproduce

We discovered this in ci:

```
Tue, 22 Feb 2022 03:30:46 GMT
stage-0: loadgen: RemoteError(error:captp:[http://127.0.0.1#20001)#1:](http://127.0.0.1/#20001)#1:) "petname" not found: (a string)
```
-- https://github.com/Agoric/agoric-sdk/runs/5282602699?check_suite_focus=true#step:10:4824

I reproduced it locally, so we can see the obscured petname is `USDC`:

```
2022-02-22T16:07:56.589Z SwingSet: ls: v10: Error#1: petname not found: IE:"USDC"
```

## Expected behavior

loadgen can trade USDC on the AMM.

In particular, it can get the USDC issuer and brand via its wallet:

https://github.com/Agoric/testnet-load-generator/blob/8939609fd21151858cdff16bc83f684dedfb60fd/loadgen/contract/agent-trade-amm.js#L19-L21

Meanwhile, production bootstrap code paths remain free of mechanisms that we don't want in production.

## Platform Environment

 - _what OS are you using?_ Ubuntu 20.04.3 LTS _what version of Node.js?_ v16.13.0
 - _is there anything special/unusual about your platform?_ it fails in ci too
 - _what version of the Agoric-SDK are you using?_ 22356c2b4 locally

## Additional context

This is evidently a consequence of the **BREAKING CHANGE** in 2cbf293f4d5f #4541

dckc:
closes: #4634

I'd like to be sure @mhofman can reproduce my results before moving beyond draft.

## Description

Grant access to the VaultFactory creatorFacet to up to one user based on address.

### usage

in a dapp dir, after `agoric install`:

```
agoric start local-solo --reset -v > ,solo.log 2>&1
addr=$(cat _agstate/agoric-servers/local-solo-8000/ag-cosmos-helper-address)
VAULT_FACTORY_CONTROLLER_ADDR=$addr \
  CHAIN_BOOTSTRAP_VAT_CONFIG=@agoric/vats/decentral-loadgen-config.json \
    agoric start local-chain --reset -v
```

then you should see `home.vaultFactoryCreatorFacet` and be able to call `E(home.vaultFactoryCreatorFacet).addVaultType()` (cf [vaultFactory.js](https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/vaultFactory/vaultFactory.js)).

### Security Considerations

This mixes a demo / test mechanism in with production `econ-behaviors.js`, which we aim to avoid. But I think it's clear how the least authority needed is granted.

_IOU an update to env.md_

### Testing Considerations

It's lightly tested; `packages/vats/test/test-boot.js` does cover it to some extent.

cc @michaelfig


packages/vats/decentral-loadgen-config.json
```diff
@@ -0,0 +1,65 @@
...
+          "template": {
+            "agoricNames": true,
+            "bank": true,
+            "namesByAddress": true,
+            "myAddressNameAdmin": true,
+            "board": true,
+            "faucet": true,
+            "zoe": true,
+            "vaultFactoryCreatorFacet": true
+          }
```
dckc: @michaelfig the keys of this template are supposed to be the things that the client manager waits for before resolving the chainbundle. Is this obsolete as of the notifier-based protocol for `home` properties?

Can the loadgen get that notifier?

michaelfig: > @michaelfig the keys of this template are supposed to be the things that the client manager waits for before resolving the chainbundle. Is this obsolete as of the notifier-based protocol for `home` properties?

It's mostly orthogonal.  Those things are the items that must be present before the *first* time the notifier fires.

> Can the loadgen get that notifier?

Not currently.  That would be a good change (always include the notifier as one of the properties on `agoric`).

Until then, the template ensures deploy scripts can have access to those objects.  Without the template and without the notifier, there is only polling.  So let's always ensure one of those is in force.

mhofman:
I have verified that this change, granting both `priceAuthorityAdmin` and `vaultFactoryCreator` to the loadgen client, is sufficient to rewrite the loadgen provisioning and re-enable the amm and vault tasks.

mhofman (approved):

Issue #3154, opened 2021-05-21
Make types of `home` available to deploy scripts


katelynsills:
Is there a reason we can't somehow import these types from the SDK instead of redefining them?

_Originally posted by @mhofman in https://github.com/Agoric/dapp-fungible-faucet/pull/21#discussion_r637245951_

dckc:
In discussion with @michaelfig , we agreed `@agoric/solo` is a good home for relevant types.

dckc:
note: `home` no longer subsumes `agoric` as of #4642

cc @michaelfig

PR #4642, opened 2022-02-23
feat(vats): implement "big hammer" governance

### LINKED ISSUE #4352: "Big hammer" Cosmos governance via bootstrap code evaluation

Cosmos-level chain governance should have in-depth control of many privileged ocaps, so that unanticipated problems can be fixed and new features can be installed.

## Description of the Design

Allow Cosmos governance to propose and approve a (set of) JS function source code with permits that will be run by the bootstrap vat.

michaelfig:
closes: #4352 

## Description

Implement `'core'` bridge handler to enable governance to evaluate core boot code.  Update solo to allow new `home` objects to be pushed.

Remove "leftover actions" assertion, since it is fragile in the face of governance and has thus outlived its usefulness (all actions that make it through the `actionQueue` are precious, not just heuristically-determined ones).

Drive-by update `faucet-helper.sh` not to give away so many assets.

### Demo

This is a demo of the end-to-end: passing a governance action to add something to every ag-solo's `home`:

```sh
cd packages/cosmic-swingset
make scenario2-setup scenario2-run-chain # in one console
make scenario2-run-client # in another console
# Wait for chain and client to boot, then type at the REPL: `agoric`
# propose and vote on gov-permit.json and gov-code.js:
make scenario2-core-eval scenario2-vote
agd query gov proposals
# Wait for the voting_end_time to happen, then voila, typing `agoric` at REPL results in `youAreGoverned`:
```

image

### Security Considerations

No new security dependencies beyond what 2/3 validators could do anyway.

### Testing Considerations

Will be tested in earnest on the next Devnet.


packages/cosmic-swingset/gov-code.js
```diff
@@ -0,0 +1,7 @@
+/* global E, powers */
+const {
+  consume: { client },
+} = powers;
+
+console.info('hello from the governed code!');
+E(client).assignBundle([addr => ({ youAreGoverned: addr })]);
```
dckc: move this under `/test/` maybe? Somehow clarify its purpose and that it's not part of the production code.

packages/deployment/ansible/roles/cosmos-genesis/files/faucet-helper.sh
```diff
@@ -30,12 +30,12 @@ IFS=, read -r -a origRpcAddrs <<<"$(AG_SETUP_COSMOS_HOME=$thisdir ag-setup-cosmo
 
 rpcAddrs=(${origRpcAddrs[@]})
 while [[ ${#rpcAddrs[@]} -gt 0 ]]; do
-  r=$(( $RANDOM % ${#rpcAddrs[@]} ))
+  r=$(( RANDOM % ${#rpcAddrs[@]} ))
```
dckc, line 33: TIL: `echo $(( RANDOM ))` is a bash-ism
TIDNWTLMB. (_Today I Did Not Want To Learn More Bash-isms_:)

packages/deployment/ansible/roles/cosmos-genesis/files/faucet-helper.sh
```diff
@@ -30,12 +30,12 @@ IFS=, read -r -a origRpcAddrs <<<"$(AG_SETUP_COSMOS_HOME=$thisdir ag-setup-cosmo
...
   selected=${rpcAddrs[$r]}
   rpcAddrs=( ${rpcAddrs[@]/$selected} )
 
   # echo "Checking if $selected is alive"
-  if [[ $(curl -s http://$selected/status | jq .result.sync_info.catching_up) == false ]]; then
+  if [[ $(curl -s http://"$selected"/status | jq .result.sync_info.catching_up) == false ]]; then
```
dckc, line 38: you didn't happen to run this thru shellcheck, did you?

packages/solo/src/vat-http.js
```diff
@@ -123,8 +128,17 @@ export function buildRootObject(vatPowers) {
       }
 
       if (decentralObjects) {
+        // Assign the complete decentralObjects to `agoric`.
         Object.assign(replObjects.agoric, decentralObjects);
         doneLoading(['agoric']);
+
+        // Prune out the demo governance stuff from `home`; they're noisy.
```
dckc, line 135: I made a documentation note in #3154 that this means `home` no longer subsumes `agoric`.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -32,6 +33,58 @@ const { keys } = Object;
 
 const NUM_IBC_PORTS = 3;
 
+/**
+ * This is the code triggered by `agd tx gov submit-proposal swingset-core-eval
```
dckc:
```suggestion
 * This registers the code triggered by `agd tx gov submit-proposal swingset-core-eval
```

packages/vats/src/core/chain-behaviors.js
```diff
@@ -32,6 +33,58 @@ const { keys } = Object;
...
+  }
+
+  // Register a coreEval handler over the bridge.
+  const handler = Far('coreHandler', {
+    async fromBridge(_srcID, obj) {
+      switch (obj.type) {
+        case 'CORE_EVAL': {
+          const { evals } = obj;
+          return Promise.all(
+            evals.map(({ json_permits: permit, js_code: code }) =>
```
dckc: I can haz static type plz?

Or a comment cross-referencing a .proto file?

packages/vats/src/core/chain-behaviors.js
```diff
@@ -32,6 +33,58 @@ const { keys } = Object;
...
+              // Run in a new turn to avoid crosstalk of the evaluations.
+              Promise.resolve().then(() => {
+                const powers = extractPowers(permit, allPowers);
+
+                // Inspired by ../repl.js:
+                const globals = harden({
+                  ...farExports,
+                  console,
```
dckc, line 79: `console`... hm... I guess that's reasonable.

How about `assert` while we're at it?

packages/vats/src/core/chain-behaviors.js
```diff
@@ -32,6 +33,58 @@ const { keys } = Object;
...
+              }),
+            ),
+          );
+        }
+        default:
+          assert.fail(X`Unrecognized request ${obj.type}`);
+      }
+    },
+  });
+  await E(bridgeManager).register('core', handler);
```
dckc: I feel even more strongly about magic strings than I did before getting into this bootstrap stuff.

```suggestion
  const CORE_BRIDGE_ID = 'core';
  await E(bridgeManager).register(CORE_BRIDGE_ID, handler);
```

with the definition at the top. and exported.

Well... I dunno... Maybe I can stand to leave it alone until we have another piece of code that has to use the same string.

packages/vats/src/core/client-behaviors.js
```diff
@@ -147,8 +147,17 @@ export const startClient = async ({
...
+    // Observe any configuration changes and update the chain bundle
+    // accordingly.
+    const configNotifier = await E(chainProvider).getConfiguration();
+    await observeNotifier(configNotifier, {
+      updateState(state) {
+        const { clientHome } = state;
+        chainBundle = clientHome;
+        updatePresences();
+      },
+    });
```
dckc, line 160: and the crowd went wild!

packages/vats/src/core/manifest.js
```diff
@@ -1,5 +1,6 @@
 // @ts-check
 const SHARED_BOOTSTRAP_MANIFEST = harden({
+  makeCoreEval: true,
```
dckc: is `makeCoreEval` useful on the sim-chain?

dckc (commented):
Looks great!

where is the .proto thingy?

I should probably test this a bit.

packages/deployment/ansible/roles/cosmos-genesis/files/faucet-helper.sh
michaelfig: > you didn't happen to run this thru shellcheck, did you?

I sure did.  Can you tell?

packages/deployment/ansible/roles/cosmos-genesis/files/faucet-helper.sh
michaelfig: Shellcheck made me do it.

packages/vats/src/core/chain-behaviors.js
michaelfig: Ah, that helped me catch a bug.  The `json_permits` needed parsing.

packages/vats/src/core/manifest.js
michaelfig: I've renamed it to bridgeCoreEval and moved it to just the chain manifest.

packages/vats/src/core/chain-behaviors.js
michaelfig: > How about `assert` while we're at it?

Added here and in `repl.js` for the pleasure of the ag-solo REPL.

packages/vats/src/core/chain-behaviors.js
michaelfig: > Well... I dunno... Maybe I can stand to leave it alone until we have another piece of code that has to use the same string.

That day is today.  I've created `packages/vats/src/bridge-ids.js`.

dckc (dismissed):
I verified that this works as advertised:

```
command[4] agoric.youAreGoverned
history[4] "agoric15ccjs5hmhrnsszd55npfr22gkfjeu4les90l4a"
```

I see the .proto was introduced in #4625 .

There I find discussion of documentation burden:

> There will certainly be documentation, but this help string is not the place for it.
-- https://github.com/Agoric/agoric-sdk/pull/4625#discussion_r812314469

I prefer that any such statements about the future come with references to issues by number, but I don't suppose that's critical to landing this PR.

I'm also a little hesitant that the bulk of the testing here is manual integration testing, but again, `test-boot.js` does seem to cover it at least a little.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -21,16 +22,77 @@ import {
...
+                // Inspired by ../repl.js:
+                const globals = harden({
+                  ...farExports,
+                  assert,
+                  console,
+                  powers,
+                });
+
+                // Evaluate the code in the context of the globals.
+                return new Compartment(globals).evaluate(code);
```
dckc: Let's have the evaluated code return a function and pass the powers to that function, so that the evaluated code can be organized according to POLA rather than having powerful globals available ambiently.

dckc:
I was able to start the RUN protocol, though it's a little twisty. I'll be thinking about ways to straighten it out.

dckc (approved):
feel free to add powerless stuff such as `utils`, `bootBehaviors`, ... to the globals.

Issue #4645, opened 2022-02-23
package dependency graph has cycles


turadg:
## Describe the bug
Running a Lerna command warns,
```
lerna notice cli v3.20.2
lerna WARN ECYCLE Dependency cycles detected, you should fix these!
lerna WARN ECYCLE @agoric/swingset-vat -> @agoric/notifier -> @agoric/swingset-vat
lerna WARN ECYCLE @agoric/deploy-script-support -> @agoric/vats -> @agoric/pegasus -> @agoric/deploy-script-support
lerna WARN ECYCLE @agoric/governance -> @agoric/zoe -> @agoric/governance
lerna WARN ECYCLE @agoric/run-protocol -> (nested cycle: @agoric/deploy-script-support -> @agoric/vats -> @agoric/pegasus -> @agoric/deploy-script-support) -> @agoric/run-protocol
lerna WARN ECYCLE @agoric/store -> (nested cycle: @agoric/swingset-vat -> @agoric/notifier -> @agoric/swingset-vat) -> @agoric/store
lerna WARN ECYCLE @agoric/swing-store -> @agoric/swing-store
lerna WARN ECYCLE @agoric/wallet-backend -> (nested cycle: @agoric/run-protocol -> (nested cycle: @agoric/deploy-script-support -> @agoric/vats -> @agoric/pegasus -> @agoric/deploy-script-support) -> @agoric/run-protocol) -> @agoric/wallet-backend
lerna WARN ECYCLE (nested cycle: @agoric/store -> (nested cycle: @agoric/swingset-vat -> @agoric/notifier -> @agoric/swingset-vat) -> @agoric/store) -> (nested cycle: @agoric/store -> (nested cycle: @agoric/swingset-vat -> @agoric/notifier -> @agoric/swingset-vat) -> @agoric/store)
lerna WARN ECYCLE (nested cycle: @agoric/wallet-backend -> (nested cycle: @agoric/run-protocol -> (nested cycle: @agoric/deploy-script-support -> @agoric/vats -> @agoric/pegasus -> @agoric/deploy-script-support) -> @agoric/run-protocol) -> @agoric/wallet-backend) -> (nested cycle: @agoric/wallet-backend -> (nested cycle: @agoric/run-protocol -> (nested cycle: @agoric/deploy-script-support -> @agoric/vats -> @agoric/pegasus -> @agoric/deploy-script-support) -> @agoric/run-protocol) -> @agoric/wallet-backend)
```

## To Reproduce
[...]

PR #4416, opened 2022-01-28
ci: continuing linting when one kind errors

turadg:
## Description

It's helpful to developers that when CI fails you are informed what you need to do to make it pass. Presently when one kind of lint check fails, the logic fails fast and you get no information about subsequent ones. This was due to the `steps` in Github workflow and the `&&` in the `yarn lint` scripts.

This makes the `yarn lint` command check all the defined `lint:foo` commands and continue when any error. It accomplishes this using https://github.com/mysticatea/npm-run-all . 

It also:
- update CI to use the one `yarn lint` so its steps don't fail fast
- remove `lint-check` command since that's what `lint` does (checks)
- removes `tsc` from `lint-fix` since it can't do any fixing.
- remove the commands that used `.eslintrc-jessie.js` because that files doesn't exist (superceded by `plugin:@jessie.js/recommended`)

### Security Considerations

Introduction of a new package. It's only a devDependency and is maintained by a well-regarded project lead who's a member of ESLint, Babel, and Vue.js.

### Documentation Considerations

Notify developers to that `yarn lint-check` no longer works and to use `yarn lint` instead. 

### Testing Considerations

Locally tested that the different jobs run and failure of any is conveyed for the lint overall.

CI passes.

PR #4646, opened 2022-02-23
build: no-bail when testing/linting whole repo

turadg:
## Description

The errors CI reports in linting aren't exhaustive because it fails-fast. #4416 fixed this within package, but the workspace iterator still stops when an error is encountered (and linting failure necessarily returns a non-zero exit code).

The change here is to use https://npm.io/package/@lerna/run to use it's `--no-bail` option. (Yarn "classic" and "berry" do no support such an option).

Demonstrating with some hardcoded failures:

### Before
Screen Shot 2022-02-23 at 6 34 46 AM

### After
Screen Shot 2022-02-23 at 6 35 41 AM

The latter is more verbose because the default log level is "info". The next log level down looks like the following, but I thought it would be better to keep the output of time spent.

Screen Shot 2022-02-23 at 6 37 34 AM

You might also notice the warning about cycles, filed here: https://github.com/Agoric/agoric-sdk/issues/4645

### Security Considerations

None.

Works now more as expected.

### Testing Considerations

Improved.


package.json
```diff
@@ -41,20 +41,20 @@
     "singleQuote": true
   },
   "scripts": {
-    "OFF-clean": "yarn workspaces run clean",
+    "clean": "yarn lerna run --no-bail clean",
```
turadg: Note this also restores the `yarn clean` command. I thought it was disabled because not all packages have it, but when I run locally it's hanging after `> rm -rf xsnap-native/xsnap/build`. Any ideas why?

package.json
```diff
@@ -41,20 +41,20 @@
...
     "create-agoric-cli": "node ./scripts/create-agoric-cli.cjs",
     "format": "yarn prettier --write .github golang packages",
     "lint:format": "yarn prettier --check .github golang packages",
-    "lint-fix": "yarn workspaces run lint-fix",
+    "lint-fix": "yarn lerna run --no-bail lint-fix",
     "lint": "run-s --continue-on-error lint:*",
-    "lint:packages": "yarn workspaces run lint",
-    "test": "yarn workspaces run test",
+    "lint:packages": "yarn lerna run --no-bail lint",
+    "test": "yarn lerna run --no-bail lint",
```
michaelfig:
```suggestion
    "test": "yarn lerna run --no-bail test",
```

michaelfig (commented):
Do note, I wasn't able to get `yarn lerna run ...` working consistently in the past.  One problem I ran into was that the default concurrency settings may have caused problems.

You're welcome to explore this solution, but be aware, here be dragons!

warner:
Does the github CI viewer know how to find the errors this emits if it doesn't fail fast? I wouldn't want to scroll through the mountains of logged-exception noise in our test output to find the one part that failed.

turadg:
> Do note, I wasn't able to get yarn lerna run ... working consistently in the past. One problem I ran into was that the default concurrency settings may have caused problems.

What were the symptoms? I noticed `yarn clean` had an infinite loop which I think is addressed with [1aa55aa](https://github.com/Agoric/agoric-sdk/pull/4646/commits/1aa55aafb0807126f4a50488f6e20acb5893ee76)

> You're welcome to explore this solution, but be aware, here be dragons!

What would give you confidence to approve the PR?

warner (approved):
@michaelfig knows more about the concerns with `lerna` than me, but if he's good with it, I'm good with it.

package.json
turadg: Figured this out. It was an infinite loop due to Lerna running this clean command again after the children.

turadg: And the reason is that we had `.` (root) listed as a package in lerna.json. @michaelfig could that be the dragon?

c71baaf69db51629c29a23adb8081ef3c4e93179 fixes that and goes ahead to DRY out the list of packages using https://github.com/lerna/lerna/blob/main/commands/bootstrap/README.md#--use-workspaces

michaelfig:
> Does the github CI viewer know how to find the errors this emits if it doesn't fail fast?

Only the `lint` CI output will be structured any differently.  And that needs a scroll-through to find the multiple errors, but at least they won't be intermingled with test output.

turadg:
@kriskowal I heard you might have some thoughts/plans for the Lerna update.

michaelfig:
> What would give you confidence to approve the PR?

Passing existing CI tests, then a once-over would be enough.  Can you ping me for a review when CI is happy?

mhofman:
The `yarn.lock` churn is high on this one! Are these really all due to the bump of lerna version? Put another way, does removing lerna from dependencies really only result in removal from the `yarn.lock` (which unfortunately also translates into changed lines for package entries with multiple matching semver).

turadg:
> The yarn.lock churn is high on this one! Are these really all due to the bump of lerna version?

Not sure I understand the hypothesis. I can tell you I changed the `lerna` version specifier and then ran `yarn install`.

> Put another way, does removing lerna from dependencies really only result in removal from the yarn.lock (which unfortunately also translates into changed lines for package entries with multiple matching semver).

All removals, yes, though as you acknowledge some version removals translate into changed lines.

Well except for the mysterious change in resolution,
```
-ast-types@agoric-labs/ast-types#Agoric-built:
+"ast-types@github:agoric-labs/ast-types#Agoric-built":
```
Maybe due to how `Agoric-built` is being resolved? https://github.com/agoric-labs/recast/commit/d9316d0f0cc6bd68360c9f36927d471850558147
@michaelfig could we do this with `patch-package` instead of forking the repo?

package.json
```diff
@@ -70,6 +70,7 @@
     "patch-package": "^6.2.2"
   },
   "resolutions": {
+    "**/ast-types": "github:agoric-labs/ast-types#Agoric-built",
```
mhofman: I'd like to understand why this is necessary now, and wasn't necessary before?

kriskowal:
> @kriskowal I heard you might have some thoughts/plans for the Lerna update.

I don’t. Michael’s our resident Lerna expert and I only know enough to do releases from Endo.

package.json
kriskowal: Curious as well. `ast-types` is brought in via `recast` via `@endo/static-module-record` via `@endo/compartment-mapper` via bundle utilities. We had to modify it to make it work under SES. It can’t be a patch package because the patch would need to be distributed to dapps and any other dependee transitively. We may need to also patch `ast-types` so it runs under `-r esm` if we end up spending a lot of time supporting contract authors who stub their toe on `ast-types does not have a default export` errors often enough or if rolling forward is hard enough.

There shouldn’t be a need for `ast-types` to be deduplicated. Perhaps there’s another dependency path that brings in the wrong `ast-types` and `yarn` is deduping them to our detriment?

turadg: I'd also like to know! This resolution is just to make my laptop write out the same yarn.lock that CI does. Incidentally, why do we have a `git diff` check instead of `yarn install --frozen-lockfile`? 

Thanks for pointing out why it can't be a patch. How about a versioned release? 

> There shouldn’t be a need for ast-types to be deduplicated. Perhaps there’s another dependency path that brings in the wrong ast-types and yarn is deduping them to our detriment?

That would make sense, but there's just one thing depending on `ast-types`:
```
recast@agoric-labs/recast#Agoric-built:
  version "0.20.5"
  resolved "https://codeload.github.com/agoric-labs/recast/tar.gz/879398a55cd50a53ade179de203706a25c53fb49"
  dependencies:
    ast-types "github:agoric-labs/ast-types#Agoric-built"
    esprima "~4.0.0"
    source-map "~0.6.1"
    tslib "^2.0.1"
```

mhofman: > How about a versioned release?

That was suggested be @kriskowal as an alternative. Given the release complexity we ran into with github, and now this, we should seriously consider this again.



> Incidentally, why do we have a `git diff` check instead of `yarn install --frozen-lockfile`?

Apparently that doesn't result in a non-zero exit code if it fails...

turadg: > Apparently that doesn't result in a non-zero exit code if it fails...

I've heard rumors of this but not witnessed it. [The docs](https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) say it does and a test locally does,
Screen Shot 2022-02-23 at 4 03 30 PM

It's also what I used for the past several years without incident.

michaelfig: I don't object to your questioning my judgement.  But I don't want to be the one to have to repeat diagnosing the problem I ran into where `yarn --frozen-lockfile` didn't fail with an error but the lockfile still differed.

turadg: Nothing personal here. My last comment  above was in response to the assertion "doesn't result in a non-zero exit code if it fails". 

My questioning whether we really need to have our own diff check instead of `frozen-lock` preceded your answer in chat,
> In the testing I did around this, the changes I was able to induce without --frozen-lockfile failing were (material) semantic ones.

That information leads me to a different questions:
- what is a way to reproduce that bug in frozen-lockfile? Not because I doubt that you encountered it, but because 1) it might have been fixed in yarn since then and 2) if it hasn't I'd like to file it as a bug
- whether there are downsides to using `--frozen-lockfile` anyway, to convey that intent (keeping the `if git is dirty` as a backup for this bug)

In the interests of reducing work in flight I'll make that change optimistically and request a code review.

michaelfig: > Nothing personal here.

No worries!  I was being unnecessarily short.

> My last comment above was in response to the assertion "doesn't result in a non-zero exit code if it fails".

Steps to reproduce:

1. delete, say, the `@agoric/nat^4.1.0` section from `yarn.lock`.
2. run `yarn install --frozen-lockfile`.  It doesn't put it back (check the `git diff`), but it also doesn't error.
3. run `yarn install`, it puts the `@agoric/nat^4.1.0` section back

So the issue I found was that `--frozen-lockfile` doesn't complain about the things that a normal `yarn install` would have changed.  I'm more interested in making sure that `yarn install` wouldn't have changed anything, not just that the `yarn.lock` isn't modified.

patches/@lerna+conventional-commits+3.18.5.dev.patch
```diff
@@ -1,22 +0,0 @@
...
---- a/node_modules/@lerna/conventional-commits/lib/recommend-version.js
-+++ b/node_modules/@lerna/conventional-commits/lib/recommend-version.js
-@@ -51,7 +51,16 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath, tagPrefix, pre
- 
-         // result might be undefined because some presets are not consistent with angular
-         // we still need to bump _something_ because lerna saw a change here
--        const releaseType = data.releaseType || "patch";
-+        let releaseType = data.releaseType || "patch";
-+        
-+        // Don't gratuitously break compatibility with clients using `^0.x.y`.
```
michaelfig: Is this logic in `lerna@4.0.0`?  We needed it so that our pre-`1.x.x` packages didn't get bumped up to `1.x.x` just because they had a breaking change.

michaelfig (commented):
Looking good.  Just one more question from me.

kriskowal:
We also want to be sure that no committed files change after yarn build. There is one generated file in agoric-cli that changes if we fail to run yarn build and commit too.

patches/@lerna+conventional-commits+3.18.5.dev.patch
turadg: Yes, I mentioned this in [e78169b](https://github.com/Agoric/agoric-sdk/pull/4646/commits/e78169b7b8206973b9a15924996f4d1ac312bb0e) but should have mentioned in PR too:
> it includes what we were patching: https://github.com/lerna/lerna/pull/2486

turadg:
> We also want to be sure that no committed files change after yarn build. There is one generated file in agoric-cli that changes if we fail to run yarn build and commit too.

Does that mean the `gif diff --exit-code` should be moved to after `yarn build`?

Either way I want to punt the changes in Lerna per se to https://github.com/Agoric/agoric-sdk/issues/4660  so I can get this more immediate pain point solved (CI doesn't show all lint/test errors)

patches/@lerna+conventional-commits+3.18.5.dev.patch
kriskowal: Vindication!

PR #4659, opened 2022-02-24
test(store): Verify the properties of compareRank

### LINKED ISSUE #4675: compareRank properties should be subject to property testing

The reflexivity/transitivity/etc. of `compareRank` functions are not verified as general properties, but should be.

## Description of the Design

Use [fast-check](https://www.npmjs.com/package/fast-check), which is already in use for other agoric-sdk packages.

## Security Considerations

n/a

## Test Plan

See above.

gibson042:
Fixes #4675

## Description

Employs fast-check to verify the generic behavior of compareRank (reflexivity, transitivity, etc.).

### Security Considerations

n/a

n/a

### Testing Considerations

See above.


erights (commented):
IIUC, this PR pays a large notational cost to be compat with both ava 3 and ava 4. I see @turadg has a PR in process to upgrade us to ava 4. If this one is not in a hurry and we'll be on ava 4 soon, does it make sense to wait for that one and they remove the `assertionPassed` kludges? If so, perhaps you can stage this PR behind that one and do the simplifications now?

Or am I misunderstanding the purpose of `assertionPassed`?

gibson042:
You understand correctly, but `assertionPassed` is already in master from a previous PR and the cleanup once we have ava 4 is straightforward (`assertionPassed(t[$method](…), ()=>…)` → `t[$method](…)`). I don't think there's a need to wait, but on the other hand there's no rush to land either.

turadg:
> If […] we'll be on ava 4 soon

I don't expect https://github.com/Agoric/agoric-sdk/pull/4727 will be done soon. Two weeks at the earliest. Possibly after MN-1.

erights:
> I don't expect https://github.com/Agoric/agoric-sdk/pull/4727 will be done soon. Two weeks at the earliest. Possibly after MN-1.

@gibson042 in that case, I'll return to continuing this review under the current assumptions.

packages/store/test/test-rankOrder.js
```diff
@@ -22,6 +23,177 @@ const alice = Far('alice', {});
...
+test('compareRank totally orders ranks', async t => {
+  await fc.assert(
+    fc.property(passable, passable, (a, b) => {
+      const ab = compareRank(a, b);
+      const ba = compareRank(b, a);
+      if (ab === 0) {
+        return assertionPassed(t.is(ba, 0), () => ba === 0);
+      }
+      return assertionPassed(t.is(Math.sign(ba), -Math.sign(ab)), () => {
+        return Object.is(Math.sign(ba), -Math.sign(ab));
```
erights: If `ab` and `ba` are both `NaN`, this will pass but should not. This is a realistic edge case because our partial order comparisons like `compareKeys` return `NaN` to indicate incomparable.

packages/store/test/test-rankOrder.js
```diff
@@ -22,6 +23,177 @@ const alice = Far('alice', {});
...
+      });
+    }),
+  );
+});
+
+test('compareRank is transitive', async t => {
+  const repr = v => {
```
erights: What is the purpose of this helper function? If it is only quoting for diagnostic purposes (as seems to be the case) consider using `q` instead, which I see you're already using elsewhere. Did you choose not to use `q`?

erights (approved):
Sorry it took so long. LGTM.

Eager to learn to do more property based testing. This will help us get started doing that. Thanks!

packages/store/test/test-rankOrder.js
gibson042: Isn't `compareRank` documented in packages/store/src/types.js to only return -1, 0, or 1?

packages/store/test/test-rankOrder.js
gibson042: Ah, I should have realized this need was already satisfied. Updated.

gibson042 (commented):
Thanks, @erights. Back to you.

packages/store/test/test-rankOrder.js
erights: Correct. But I'd still prefer the test to be robust against that case.

erights (approved):
LGTM modulo NaN.

Uh oh. Anything modulo NaN is NaN ;)

Issue #1925, opened 2020-10-26
require Node >=14 for agoric-sdk


warner:
Hm, it looks like the test failures are because Node 12.x does not expose `WeakRef` by default, but Node 14.x does. This might prompt us to increase our required Node.js version to 14, rather than instruct users to add whatever command-line option might enable weakrefs in 12 (if there even is such a thing.. I see `--harmony-weak-refs`, but it's labelled as "in progress").

_Originally posted by @warner in https://github.com/Agoric/agoric-sdk/pull/1924#issuecomment-716682755_

warner:
@michaelfig noted the [Node.js Release Calendar](https://nodejs.org/en/about/releases/) which says Node 14 enters its LTS phase tomorrow (27-Oct-2020). He said v12 is not end-of-life until 30-Apr-2022, and it would be nice to not force people to upgrade to v14.

@erights says SES-shim would probably like to retain Node 12 compatibility. He suggested it would be easy to polyfill/stub `WeakRef` on platforms where it is not available (basically disabling GC), but was also disinclined to spend significant engineering time on support of an old release.

warner:
Let's hold off on this for now, I'll look into a GC-disabling shim for Node v12.

michaelfig:
At least Node v12.14.1 (which is our current minimum requirement) implements `--harmony-weak-refs` to enable WeakRef support.  That flag is compatible with v14.x.  So, I think it would be fairly straightforward to require that flag for all our top-level scripts and just use WeakRef unconditionally in SwingSet.

warner:
In the meeting this morning, we decided to drop support for Node 12. @kriskowal floated the idea of dropping Node 14 as well (to make several `-r esm` / native-esm-support work better), but we agreed we can't do that until Node 16 reaches LTS.

turadg:
> In the meeting this morning, we decided to drop support for Node 12. @kriskowal floated the idea of dropping Node 14 as well (to make several -r esm / native-esm-support work better), but we agreed we can't do that until Node 16 reaches LTS.

Now that [Node v16 is LTS](https://nodejs.org/ko/blog/release/v16.13.0/) shall we drop v14?

michaelfig:
> Now that [Node v16 is LTS](https://nodejs.org/ko/blog/release/v16.13.0/) shall we drop v14?

I would prefer a "last 2 releases" policy.  Is node 18 a thing yet?

turadg:
[...]

PR #4666, opened 2022-02-24
clean up from Node 12 deprecation

turadg:
[no ticket, this was a quick one while listening into a meeting]

## Description

https://github.com/Agoric/agoric-sdk/commit/62411f535c8eca6ea3689e32463fa5b6cf09c0bc (for #1925) dropped support for Node < 14 but some concessions to Node 12 remained. This removes them.

It also updates the non-root package.json to indicate the required version. (Unfortunately they can't inherit from root:  https://github.com/lerna/lerna/issues/2820 )

### Security Considerations

--

Already documented.

### Testing Considerations

--


packages/SwingSet/test/test-vpid-kernel.js
```diff
@@ -3,7 +3,7 @@ import { test } from '../tools/prepare-test-env-ava.js';
 
 import anylogger from 'anylogger';
 import { assert, details as X } from '@agoric/assert';
-import { WeakRef, FinalizationRegistry } from '../src/weakref.js';
+/* global WeakRef, FinalizationRegistry */
```
warner: does the `global` comment need to be on the first line of the file?

warner (approved):
looks good, modulo the question, thanks for jumping on this

packages/SwingSet/test/test-vpid-kernel.js
turadg: Not to pass lint but I'd say yes for style. I meant to move these after find-replace but missed this one.

PR #4669, opened 2022-02-25
Minor tweaks for devnet-8 deployment

michaelfig:
## Description

- Update MAINTANERS docs and scripts
- Default to governance voting period of 36h.


dckc (approved):

dckc (approved):
I don't really grok how this change in shell quoting makes a difference. And I don't really want to :) I'd sooner rewrite this in JS.

But I trust this fixes the faucet to go back to limiting clients to one tap apiece.

PR #4670, opened 2022-02-25
fix: restore simpler setStore type

erights:
The setStore api grew more api and type than I intended. Specifically, it should have `.keys` but not `.values` and `.entries`.

### DIFF
--- a/packages/SwingSet/src/kernel/collectionManager.js
+++ b/packages/SwingSet/src/kernel/collectionManager.js
@@ -595,11 +595,6 @@ export function makeCollectionManager(
       snapshot,
       clear,
     } = collection;
-    function* entries(patt) {
-      for (const k of keys(patt)) {
-        yield [k, k];
-      }
-    }
     function addAll(elems) {
       for (const elem of elems) {
         init(elem, null);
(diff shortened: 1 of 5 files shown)


FUDCo:
It has those because the JavaScript `Set` has those.  You should leave them in.

FUDCo:
Further thought after double checking with MDN to verify that I wasn't just talking out of my hat with that last comment: `Set` has `values` and `entries` but it does not have `keys`.  We should remove `keys` from SetStore but leave `values` and `entries` in.  Or come up with a different name that doesn't make JS programmers think it's related to `Set`.

FUDCo (changes requested):
I think this change is mistaken in principle.  We should discuss before proceeding.

gibson042:
> Further thought after double checking with MDN to verify that I wasn't just talking out of my hat with that last comment: `Set` has `values` and `entries` but it does not have `keys`. We should remove `keys` from SetStore but leave `values` and `entries` in. Or come up with a different name that doesn't make JS programmers think it's related to `Set`.

Set does have `keys`: https://tc39.es/ecma262/multipage/keyed-collections.html#sec-set.prototype.keys

I don't know why it's missing from MDN, but I'll open a PR to fix that.

gibson042:
> I don't know why it's missing from MDN, but I'll open a PR to fix that.

It appears to be an accidental consequence of an intentional redirect: https://github.com/mdn/content/issues/13286

I'll let someone else sort out MDN.

erights:
I still think this would have been an improvement. But it is too late to consider. The existing API is, at this point, entrenched enough that it is not worth changing for this purpose. Closing.

Issue #4655, opened 2022-02-23
@agoric/vats defines types "owned" by @agoric/swingset-vat


dckc:
## Describe the bug

As noted in https://github.com/Agoric/agoric-sdk/pull/4590#issuecomment-1048526336 , @agoric/vats includes type declarations such as...

```
 * @typedef { Device<ReturnType<typeof
 *   import('@agoric/swingset-vat/src/devices/mailbox-src.js').buildRootDeviceNode>> } MailboxDevice
```

These should be exported by the `@agoric/swingset-vat` package.

## Expected behavior

Clients of swingset should rely on swingset to define its types.

## Additional context

https://github.com/Agoric/agoric-sdk/pull/4590#issuecomment-1048526336

warner:
I've got a partial thing for this in my attempts to get #4950 passing, will post it here when I come up for air.

dckc:
I just realized that a lot of the redundancy in the `consume:` / `produce:` types can go away using...

```js
/**
 * TODO: refactor vats/core/types.js like this
 *
 * @template T
 * @typedef {{
 *   consume: { [P in keyof T]: ERef<T[P]> },
 *   produce: { [P in keyof T]: Producer<T[P]> },
 * }} PromiseMarket
 */
```

for example:

```js
/**
 * @typedef { PromiseMarket<{
 *   getRUNBundle: SourceBundle,
 *   client: ClientManager,
 *   lienBridge: StakingAuthority,
 * }>} GetRunBootstrapPowers
 */
```

dckc:
In working on getRUN, I'm adding `Attestation` to the list of well-known issuers (6a640fe). Should this list be in `run-protocol` rather than `vats`?

PR #4680, opened 2022-02-26
rearrange swingset files

### LINKED ISSUE #4502: refactor swingset source code to match runtime artifacts

It's not particularly obvious how the swingset source code maps to the various Compartments and bundles and workers that get used at runtime. We've mused about a better layout, but now I wanted to write up a proposal and implement it.

Swingset is a library, meant to be imported into a host application, which interacts with a "controller". The library consists of several pieces:
* "controller": this code is imported into the host application, and thus runs in the "Start Compartment" (where it has access to the standard Node.js facilities like `fs` and `crypto`)
* "kernel": the controller creates a bundle of all the kernel code and stores it in the database, so a given swingset instance will keep using the same kernel code until deliberately upgraded (#4375). It uses `importBundle` to create a new Compartment and evaluate the kernel bundle inside it.
* "worker": most vats run inside an "xsnap worker", which is a separate (unix) process launched from the `xsnap` binary. The kernel/host-app process sends commands over a pipe to evaluate code bundles and process commands. The first bundle evaluated inside the worker is the `lockdown` bundle, which establishes a SES environment. The second is the `supervisor` bundle, which installs the command handler. The command handler currently imports liveslots (which is therefore incorporated into the supervisor bundle), but we intend (#4376) to make that a separate bundle and deliver it over the command pipe. Finally the supervisor is instructed to evaluate the vat source bundle, to personalize the worker into a specific vat.
* "devices"/"vats": swingset defines a number of vats and devices for internal use (comms, vatAdmin, bundle), which are present in all swingset instances, and need no external endowments. Their source code lives inside the swingset tree, and are bundled by the controller, then stored in the DB. Swingset also provides the source code for some common devices (timer, mailbox, network) that are used by most host applications. These common devices are only enabled if the host configures them (and provides the necessary endowments like a clock). If configured, their sources are bundled by the controler and stored in the DB.
* "development tools": we have a few exports that are intended to help write unit tests of code that expects to run inside a vat environment, without actually starting a whole swingset kernel. We have a debug flavor and a non-debug flavor. By importing one of these files from your `test-*.js` file, before importing anything else, the test program is provided with a suitable SES environment that includes adequate mocks for the special vat facilities like `VatData.makeKind` (for virtual objects).

Bundles include a copy of all the files transitively imported by their entry-point file. The bundler (Endo) will refuse to follow imports to non-source modules, like Node's builtins (`fs`) or modules that include C/C++ -compiled `.so` binary libraries. So if two components reference a common source file, and one component is included in a bundle, that common source file must not import any builtins or compiled modules.

## Description of the Design

I'm thinking of the following layout:

* `src/lib/*.js` : small files which are used by multiple places, like `parseVatSlot.js`. These should not import anything (at least not outside of `lib/`), and must not import builtins or binary modules. These are safe to use from bundled code.
* `src/kernel/*.js`: files which get included in the kernel bundle. Excludes liveslots and vat-supervisor code. Includes vat-manager code. Includes vat-warehouse.
* `src/supervisor/*.js`: files which get included in the supervisor bundle. Maybe break this out into subdirs for the different worker types.
* `src/liveslots/*.js`: files which get included in the liveslots bundle. References `lib/` but nothing else. Includes all the virtual object/collection code.
* `src/builtin-vats-devices/*.js`: code to be bundled into built-in devices and vats: comms, vat-tp
* `src/devices/`: code for devices that userspace must enable and configure, not included automatically

I need to understand `src/vats/network/*.js` and the "plugin manager" to see where they fit. We might need a `src/vats/` if swingset provides source code for vats that must be enabled by the host application.

## Security Considerations

none I can can think of, probably makes things easier to audit

## Test Plan

If the unit tests (which will need a lot of import-path surgery) still work, we'll know this was successful.

cc @FUDCo is there any other housekeeping you'd like to see done while we're tidying up?

warner:
Big cleanup of how swingset is organized.

Main directories now correspond to the bundles that are created and/or the
environments into which they are loaded

* everything in `src/kernel/` goes into the kernel bundle and loaded into the
kernel Compartment
* everything in `src/liveslots/` goes into the liveslots bundle
* `src/supervisors/` contains one directory per worker type, whose contents
get loaded into the corresponding supervisor bundle and loaded into a child
process or Node.js worker thread
* `src/vats/` contains one directory per built-in vat
* `src/devices/` contains one directory per built-in device

All code is free to import from `src/lib/` and from other `@agoric/*` packages within the
agoric-sdk monorepo, under the assumption that they match our general style:
all modules are pure, none import platform builtins, none import third-party
packages (which might violate the other rules).

`src/controller/` builds the controller, which is imported by the host
application into the start compartment.

All imports of platform builtins are under `src/lib-nodejs/` (or, for now,
`src/controller/`), and hopefully this will help us improve ocap discipline
over time by making the primary "build a swingset" API take a batch of powers
instead of importing the platform builtins by itself.

closes #4502


warner:
@FUDCo this will be 100x easier if you review the commits individually. The second commit only moves files, and should have absolutely no changes to their contents. The third commit only updates the import paths, and does not move any files around, nor does it move any lines around (in fact, I recommend you tell github to hide whitespace changes, to make the visible delta even smaller and easier to confirm the nothing-but-imports claim).

(note to everybody: splitting up file-rearranging PRs this way also makes rebasing a lot easier, git doesn't get confused about what delta is a rename and which delta is modifying the import paths. It's non-trivial to set up, though)

@michaelfig please review the `packages/cosmic-swingset/` change, and note that I copied `json-stable-stringify.js` out of swingset and into `cosmic-swingset`. The kernel doesn't use it anymore.. the kernel uses a copy of `djson.js` for its transcript comparison needs. `cosmic-swingset` might want to do this too, I forget why we switched (maybe performance?), but that would probably be a compatibility break with `agd` runtimes and this PR should not cause one.

@dckc please review the `packages/vats/` change, I had to change a handful of deep import paths (all involving typedefs, I think). We should think about how to avoid both those and the actual code deep imports, or figure out some considered/committed-to set of swingset exports, so later refactorings won't accidentally break downstream users like `vats/`.

dckc (commented):
> @dckc ... I had to change a handful of deep import paths (all involving typedefs, I think).

Looks fine.

> We should think about how to avoid both those and the actual code deep imports

Yes, we should (as noted in #4655).

michaelfig (commented):
`cosmic-swingset` LGTM!  I'll leave it for @FUDCo to give the approving review.

FUDCo (approved):
Remarkably straightforward for something so sweeping.  Looks good.

PR #4682, opened 2022-02-26
fix(swingset): add stub vat-upgrade calls (always fails)

warner:
This implements E(admin).upgrade(bundlecap, vatParameters), which queues the
request to the kernel, and awaits the response. The kernel always rejects the
request. The added unit test asserts that it gets an error of "unimplemented"
rather than some other error.

The next step is to enhance vat-warehouse to stop any current worker, replace
the source bundle, start a new worker, perform the upgrade delivery, make
sure it succeeds, and commit the results (or abort the changes).

refs #1848


packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
```diff
@@ -92,6 +94,7 @@ bundleID before submitting to the kernel), or (temporarily) a full bundle.
 export function buildDevice(tools, endowments) {
   const { hasBundle, getBundle, getNamedBundleID } = endowments;
   const { pushCreateVatBundleEvent, pushCreateVatIDEvent } = endowments;
+  const { pushUpgradeVatEvent } = endowments;
   const { terminate } = endowments;
   const { meterCreate, meterAddRemaining, meterSetThreshold, meterGet } =
     endowments;
```
FUDCo, line 100: Any particular reason for having 5 separate destructing assignments from `endowments` instead of 1?

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
```diff
@@ -214,6 +217,22 @@ export function buildDevice(tools, endowments) {
...
+          // see #4381, we really need to accept slots in vatParameters
+          // so we can pass contract bundlecaps to upgraded ZCF
+          const { body, slots } = argsCapdata;
+          // this has all the same problems described in terminateWithFailure
+          assert.equal(slots.length, 0, 'upgradeVat() cannot handle refs yet');
+          const args = JSON.parse(body);
+          assert(Array.isArray(args), 'upgradeVat() args array');
+          const [bundleID, vatParameters] = args;
+          assert.typeof(bundleID, 'string', `upgradeVat() bundleID`);
+          const vpCapdata = { body: JSON.stringify(vatParameters), slots };
```
FUDCo, line 232: Probably should assert that `args` is not only an array but specifically an array of length 2, else the reuse of `slots` on line 231 could pull in other (weird) slot entries that belong to something other than `vatParameters` once `slots` are usable there.  I don't think such extra stuff would introduce a vulnerability or other problems, but it _would_ be a little messy and why take the trouble to analyze such a case when you can just exclude it.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -89,6 +90,19 @@ export function buildRootObject(vatPowers) {
       done() {
         return doneP;
       },
+      upgrade(bundlecap, vatParameters) {
+        let bundleID;
+        try {
+          bundleID = D(bundlecap).getBundleID();
+        } catch (e) {
+          // 'bundlecap' probably wasn't a bundlecap
```
FUDCo, line 98: "Probably"?  Or should the error do something with the message in the error from `getBundleID`?

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -193,6 +207,17 @@ export function buildRootObject(vatPowers) {
     }
   }
 
+  // the kernel queues this to us when a vat upgrade completes or fails
+  function vatUpgradeCallback(upgradeID, success, error) {
+    const { resolve, reject } = pendingUpgrades.get(upgradeID);
+    pendingUpgrades.delete(upgradeID);
```
FUDCo, line 213: Thinking back to the thing we had to do with the race between the `pending` and `running` maps in the event of vat termination during `startVat`, I'm wondering if there's possibly a similar kind of thing here with `pendingUpgrades`.  Haven't thought deeply about it but calling the notion to your attention just in case.

packages/SwingSet/test/snapshots/test-state.js.md
```diff
@@ -0,0 +1,15 @@
...
+
+## crankhash - initial state and additions
+
+> initial state
+
+    'fb271936c31500468a08241699d29c850bcac0167c4b0bbd0db9d9124926eabe'
+
+> expected activityhash
+
+    '55447e103e9da37a42e37759b09e4c57014a6ccee3273d27af4addac26c28ec1'
```
FUDCo, line 15: Should this file and the corresponding (binary!) `.snap` file be put in git or is this a commit mistake?

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -0,0 +1,78 @@
...
+  const config = {
+    defaultManagerType,
+    bootstrap: 'bootstrap',
+    defaultReapInterval: 'never',
+    vats: {
+      bootstrap: { sourceSpec: bfile('bootstrap-upgrade.js') },
+    },
+    bundles: {
+      ulrik1: { sourceSpec: bfile('vat-ulrik-1.js') },
+      ulrik2: { sourceSpec: bfile('vat-ulrik-2.js') },
```
FUDCo, line 25: "ulrik"?  Inquiring minds want to know...

FUDCo (approved):
I think this is OK, but there are a few things I flagged that you might want to check on before landing it (i.e., not bad/weird enough to explicitly request changes, but probably worth a look).

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
warner: Just trying to keep `prettier` from splitting it up into a gigantic vertical sprawl, which I found less readable. Not sure if my approach is the best, and I think it got worse over time (slow creep of additional endowments) but I'm increasingly annoyed at `prettier`'s enthusiasm for increasing the line count. Next time I'm in this file I'll probably look for a better approach.

packages/SwingSet/test/snapshots/test-state.js.md
warner: Yeah, both belong in git, this is AVA's way to help with the updating of golden-master -type tests. Previously any changes to the initial kernel state (like adding the `vat.nextUpgradeID` key) required `test-state.js` to be modified, to replace the initial activityhash with a new value. Now, you run `yarn test` with an extra argument to say "yep, I know the recorded hash is wrong, please accept the current version as the right one". The `.md` and `.snap` files record the expected version. I'm surprised that `.snap` is binary, but it looks like that's just how AVA does it.

(I'm following the pattern that @turadg established with `test-xsnap-store.js`)

packages/SwingSet/test/upgrade/test-upgrade.js
warner: Needed a name that started with U, for Ulrik the Upgraded (vat), like Tammy the Terminated or Pauline the Pipelined (neither of which are examples we use but we totally should). And all the other "U" baby names I found online were in norwegian or icelandic and contained characters I don't know how to type :)

FUDCo: "Ulrik the Upgraded" sound like one of the less well known kings of Sweden.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
warner: Hm, looking at `device-vat-admin.js`, I think we've got two possible error cases. One is if we make it all the way through to `dispatch.invoke` within `device-vat-admin`, which can only happen if the `D(target)` target object is really a device node of `device-vat-admin`, and then something goes wrong internally. `D(vatAdminNode).getBundleID()` (i.e. right method, wrong device node) would throw `target[getBundleID] does not exist`. If we add some new kind of `device-vat-admin` device node in the future, `D(newNode).getbundleID()` would probably do the same thing. If something goes wrong with the device's saved state (say some kind of GC or `vatstore` bug), and it didn't recognize the device nodes it created earlier, it would throw `unknown device node ${dnid}, shouldn't happen`. None of these three errors would be provokable by the user calling `upgrade()`.

The other category of error would be if `D(bundlecap).getBundleID()` doesn't make it to `device-vat-admin`. If `bundlecap` is not a device node, liveslots and `vatPowers.D` will throw `D() must be given a device node` before it ever even makes it out of the vat. That's the most likely user-provokable error, e.g. if they did `E(adminNode).upgrade()` or `upgrade(bundle)` or `upgrade('v1')` or `upgrade(zoeInstallationHandle)` (which is a Presence). Another possibility is a device node, but from the wrong device: `E(adminNide).upgrade(timerRepeaterNode)` or something, which will probably get a "unknown method" error from that other device.

So I figured the only user-provokable error is "not a bundlecap", and the error returned by `D()` is not so helpful to the user, so catch-and-throw-different seemed best.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
warner: That's a great point, if the vat is terminated while we've got an upgrade pending, what should we do? Also, if someone creates a vat, then schedules an upgrade right away before the `startVat` has even had a chance to run (this "continuous deployment" thing taken to an extreme), and then `startVat` fails.

I'm going to make a note to address this as the upgrade process matures. The answer might be to reject all `pendingUpgrades` if/when we get `vatTerminated`, or if we get a `newVatCallback` with an error.

PR #4683, opened 2022-02-26
Enforce `home.board` prefixes

### LINKED ISSUE #2014: Board IDs should self-identify as boardIDs
**What is the Problem Being Solved?**

Board IDs are currently raw numbers. They are used to refer to registered ocaps out of band, including installations and user depositFacets. The problem is that then they cannot be distinguished with the wild from other numbers like keys. There remain situations where the IDs are shared among people.

**Description of the Design**

- [ ] Add a prefix (e.g., board:092384)
- [ ] Specify the boardID format (e.g., it currently has a CRC
- [ ] Make sure they are actually sparse

**Security Considerations**

Sparseness protects against typos when entering one. A prefix protects against using one as a different kin dof ID and being confused. (e.g., as a public key).

**Test Plan**

michaelfig:
closes: #2014

## Description

Update boardIds to be `board0${crc}${serial}`.  The starting sequence number and prefix are configurable.

Also, update the wallet UI not to add or strip `board:` prefixes.


packages/vats/src/lib-board.js
```diff
@@ -32,40 +31,50 @@ const calcCrc = num => {
 /**
  * Create a board to post things on.
  *
- * @param {number} [seed=0]
+ * @param {number | bigint} [firstSequence=0]
+ * @param {string} [prefix]
  * @returns {Board}
  */
-function makeBoard(seed = 0) {
+function makeBoard(firstSequence = 0, prefix = 'board0') {
```
turadg: making this parameterized implies that we're not guaranteeing all board ids begin the same way. That implies software can't look at a board ID and know that it's a board ID.

#2014 talks about "distinguishing in the wild" which sounds like it's mostly for human consumption and allowing new prefixes or synonyms could be socialized.

Is it explicitly not an objective that board IDs can be detected as such by software? (i.e.. matching an invariant pattern) That seems worth including to me, but I'm new to this area.

packages/vats/src/lib-board.js
```diff
@@ -32,40 +31,50 @@ const calcCrc = num => {
...
     getId: value => {
       if (!valToId.has(value)) {
-        // Retry until we have a unique id.
-        let id;
-        do {
-          const num = /** @type {number} */ (sparseInts.next().value);
-          id = `${num}${calcCrc(num)}`;
-        } while (idToVal.has(id));
+        lastSequence += 1n;
+        const seq = `${lastSequence}`;
```
turadg: This seems more like a type conversion than a string template:
```suggestion
        const seq = String(lastSequence);
```
(As a template, reader has to see whether there are any other changes going on.)

packages/vats/src/lib-board.js
```diff
@@ -32,40 +31,50 @@ const calcCrc = num => {
...
+
+        // Reverse the sequence, so that the initial part of the board ID is
+        // well-distributed.
```
turadg: Why is it important for apparent number to be well-distributed? You don't want the ordering to be so obvious?

packages/vats/src/lib-board.js
```diff
@@ -32,40 +31,50 @@ const calcCrc = num => {
...
+        const reverseSeq = seq.split('').reverse().join('');
+
+        // Append the CRC, so that the last part of the board ID is
+        // well-distributed.
```
turadg, line 56: If it's important to be well-distributed, would it suffice to have `board0${crc}${serial}`? Somewhat simpler to express and parse.

packages/vats/src/lib-board.js
```diff
@@ -9,18 +8,18 @@ import { crc6 } from './crc.js';
 import './types.js';
 
 const CRC_NUM_DIGITS = 2;
```
turadg: out of scope but curious, why 2 digits?

michaelfig: More than 1 but not too many? :p

Not for any good reason.  I think I did it earlier to try to help disambiguate in the face of typos.

packages/vats/src/lib-board.js
michaelfig: I'm following this as a convention established by other parts of our code.

But I agree, the `String` call is clearer in at least this case.

packages/vats/src/lib-board.js
michaelfig: I find that people when comparing two large numbers tend to look at the first few digits and the last few digits.  Let me address this below....

packages/vats/src/lib-board.js
michaelfig: It's a tradeoff, but I can agree with your suggestion.  Your comment addresses the constraint I was trying to express.

packages/vats/src/lib-board.js
michaelfig: You're right that it's for human consumption: UI software should *not* require a particular boardID structure (except for 'string'), though a conventional suggestion is fine for assisting humans.

The parameters allow the creation of a board with different behaviour.  But the `home.board` on Agoric will have conventional behaviour.  This is the same motivation as not requiring the use of `RUN` in our contracts since they could be repurposed.  But typically on the Agoric chain, it will be `RUN`.

packages/vats/src/lib-board.js
turadg: kk, changing this is out of scope

packages/vats/src/lib-board.js
turadg: So the motivation is to reduce effort to see that two numbers are different? Impressed with the thoughtfulness! 👍

turadg (approved):

packages/wallet/ui/src/components/ImportContact.jsx
```diff
@@ -84,7 +84,7 @@ export const ImportContactWithoutContext = ({
             variant="standard"
             InputProps={{
               startAdornment: (
-                <InputAdornment position="start">board:</InputAdornment>
+                <InputAdornment position="start">board0</InputAdornment>
```
samsiegart, line 87: We might want to move this below the input rather than in the box, like "board0..." as a help tip. When it's at the start, it's unclear that you should still include "board0" in the input

michaelfig: > We might want to move this below the input rather than in the box, like "board0..." as a help tip.

I'd gladly approve this kind of change next time you're working nearby.

>  When it's at the start, it's unclear that you should still include "board0" in the input

Agreed.

Issue #4565, opened 2022-02-16
Remove Zoe support for full-source bundles in favor of hash bundles


warner:
[updated by @kriskowal]

## What is the problem being solved

Zoe currently accepts bundles that contain all of their source, resulting in very large transactions to the chain. Per #4563, then #4974, Zoe will be able to additionally accept “hash bundles”, specifically `EndoZipBase64Sha512Bundle`. These would be bundles that just refer to the bundle by its consistent hash, instead of carrying all the source, and SwingSet will receive the bundles by `controller.validateAndInstallBundle(sourceBundle)` instead.

This issue tracks the eventual breaking change that will remove support for source bundles, making hash bundles the only supported option.

---

[original ticket by @warner]

## What is the Problem Being Solved?

After #4563 adds the ability to do `E(zoe).installBundleID(bundleID)`, and #4564 changes all our dapps to use it, we can remove support for `E(zoe).install(bundle)`. This will simplify zoe's contract-installation storage table (just `{ Bundlecap }`, not `{ Bundlecap | Bundle }`, as well as the Zoe-to-ZCF `evaluateContract()` message.

## Description of the Design

* deprecate and remove `E(zoe).install(bundle)`, leaving only `E(zoe).installBundleID(bundleID)`
* Zoe's `installationStorage.js` stores only bundlecaps
* `zcfZygote.js`'s `evaluateContract(bundleOrBundlecap)` becomes `evaluateContract(bundlecap)`

## Test Plan

The one zoe unit test that asserts `bundle` support is removed.

Tartuffo:
De-prioritized for MN-1 in discussion with @warner on Feb 22.

dckc:
This has fairly large documentation impact.

dckc:
I was a little confused by the title; it looks a lot like #3724 / #4395 at a glance.

Perhaps "obsolete zoe.install(bundle) in favor of installBundleID" would be more clear?

kriskowal:
@warner, @michaelfig, and I arrived at a different consensus, one where we keep `E(zoe).install(bundle)` and instead of changing the name, we first expand the definition of a suitable bundle, then contract it in a future breaking change. Currently, `install` accepts what we would call a `SourceBundle` or merely `Bundle`, then we introduce a `HashBundle` such that `install(Bundle | HashBundle)` and later withdraw support for source bundles, `install(HashBundle)`.

[...]

PR #4684, opened 2022-02-27
test(zoe): switch to E(zoe).installBundleID()

warner:
The new preferred way to install a contract is by its bundleID, using
`E(zoe).installBundleID(id)` instead of `E(zoe).install(bundle)`. The bundle
itself must first be installed/registered with the VatAdmin service. In a
real swingset, this happens out-of-band, via
`controller.validateAndInstallBundle`. In a non-swingset unit test, the
`fakeVatAdmin` has a new `vatAdminState.installBundle(id, bundle)`.

This changes most Zoe unit tests to perform `vatAdminState.installBundle` and
then use Zoe's `installBundleID` method. This requires some mechanical
changes to the way many tests get access to their fakeVatAdmin. I took the
opportunity to switch all tests to importing `makeFakeVatAdmin` and building
their own local copy, rather than relying upon the shared singleton created
and exported by `fakeVatAdmin.js` itself.

This leaves one test in `unitTests/` using the old `E(zoe).install(bundle)`,
to ensure it still works until we decide to remove it entirely. All tests in
`swingsetTests/` still use `install(bundle)`, those will be changed later.

It also updates all swingset-based zoe tests to install contracts by BundleID
instead of a full bundle. In all cases, we add the bundles by name to
`config.bundles`, from which you can obtain the BundleID with:

```js
const bcap = await E(vatAdminService).getNamedBundleCap(name);
const id = D(bcap).getBundleID();
```

In a real chain, the bundles would be installed into the kernel at runtime,
and zoe would be given a BundleID by the (off-chain) user doing the
install (or their deploy script).

refs #4565

### DIFF
--- a/packages/zoe/test/swingsetTests/zoe/bootstrap.js
+++ b/packages/zoe/test/swingsetTests/zoe/bootstrap.js
@@ -81,24 +81,24 @@ const makeVats = (log, vats, zoe, installations, startingValues) => {
 };
 
 export function buildRootObject(vatPowers, vatParameters) {
-  const { argv, contractBundles: cb } = vatParameters;
+  const { D } = vatPowers;
+  const { argv } = vatParameters;
   return Far('root', {
     async bootstrap(vats, devices) {
       const vatAdminSvc = await E(vats.vatAdmin).createVatAdminService(
         devices.vatAdmin,
       );
       const zoe = await E(vats.zoe).buildZoe(vatAdminSvc);
-      const installations = {
-        automaticRefund: await E(zoe).install(cb.automaticRefund),
-        coveredCall: await E(zoe).install(cb.coveredCall),
-        secondPriceAuction: await E(zoe).install(cb.secondPriceAuction),
-        atomicSwap: await E(zoe).install(cb.atomicSwap),
-        simpleExchange: await E(zoe).install(cb.simpleExchange),
-        autoswap: await E(zoe).install(cb.autoswap),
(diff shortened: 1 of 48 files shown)


packages/zoe/test/swingsetTests/zoe/bootstrap.js
```diff
@@ -81,24 +81,24 @@ const makeVats = (log, vats, zoe, installations, startingValues) => {
...
+      const installations = {};
+      const installedPs = vatParameters.contractNames.map(name =>
+        E(vatAdminSvc)
+          .getNamedBundleCap(name)
+          .then(bcap => E(zoe).installBundleID(D(bcap).getBundleID()))
+          .then(installation => {
+            installations[name] = installation;
+          }),
+      );
+      await Promise.all(installedPs);
```
erights, line 101: OMG was the old code doing a ton of unnecessary `await`s. Good to see these reduced to one? Just curious: did you notice a speedup?

erights (approved):
LGTM

packages/zoe/test/swingsetTests/zoe/bootstrap.js
warner: Huh I didn't look. But also a handful of intra-kernel roundtrips would be completely swamped by the overhead of creating all the contract bundles, which takes on the order of 50 seconds, so I think I'd have to build a special test to be able to measure a difference.

PR #4686, opened 2022-02-27
refactor(run-protocol)!: compound interest from assetNofifier instead of vault

### LINKED ISSUE #4540: Refactor vault notifications to not cascade interest charge events

#4341 required moving interest accrual state from the vaults to the vault manager. https://github.com/Agoric/agoric-sdk/pull/4527 makes it so that when vaults are charged interest, it only affects the manager's state. However the downstream consumers of the vault's state still assume that interest charges come through there so it echoed the manager's notifications. This won't scale to millions of vaults.

## Description of the Design

The Vault’s dapp can have close coupling with the vault manager. No need to be consistent with other dapps, so do whatever’s simplest.
Simpler to have two notifiers for this case.
Notifiers should always include units. 

When we do this is a good time to also rename `getDebtAmount` to `getActualDebtAmount`.

## Security Considerations
TBD
## Test Plan

TBD

turadg:
closes: #4540

## Description

See #4540 for motivation. Recommended to review commit-by-commit. Fixup commits will be autosquashed before merging to master.

### Security Considerations

This introduces new requirements on the consuming dapp for determining what the current debt is. The vault itself will only report its snapshot of the debt and interest. The consumer has to:
- maintain the association of the  `assetNotifier` with the `vaultNotifier` and `vault` (received together from the vault kit)
- listen to the assetNotifier for changes to compounded interest (daily per current parameters)
- use new contractSupport functions to put those two values together to get the current debt with accrued interest

These introduces risks into calculating the debt _apparent to the dapp_ but not the real debt calculations in the system.

There are two breaking changes contained in this PR, denoted in the commit titles:
[refactor(run-protocol)!: rename uiNotifier to vaultNotifier](https://github.com/Agoric/agoric-sdk/pull/4686/commits/fad520a32b04c30046bd32345e33af7835779022)
[refactor(run-protocol)!: return assetNotifier instead of notifying on…](https://github.com/Agoric/agoric-sdk/pull/4686/commits/290eaf0e41f3ff798b52fa86d89f3e1000d85097)

### Testing Considerations

Covered in integration test updates and new unit tests.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -37,8 +37,8 @@
 
 /**
  * @typedef  {Object} VaultFactoryPublicFacet - the public facet
- * @property {() => Promise<Invitation>} makeLoanInvitation
- * @property {() => Promise<Invitation>} makeVaultInvitation
+ * @property {() => Promise<Invitation<VaultKit>>} makeLoanInvitation
+ * @property {() => Promise<Invitation<VaultKit>>} makeVaultInvitation
  * @property {() => Promise<Array<Collateral>>} getCollaterals
  * @property {() => Issuer} getRunIssuer
  * @property {(paramDescription: ParamDescription) => bigint} getNatParamState
@@ -105,7 +105,7 @@
  * @typedef {BaseVault & VaultMixin} Vault
  * @typedef {Object} VaultMixin
  * @property {() => Promise<Invitation>} makeAdjustBalancesInvitation
- * @property {() => Promise<Invitation>} makeCloseInvitation
+ * @property {() => Promise<Invitation<string>>} makeCloseInvitation
  * @property {() => Promise<Invitation>} makeTransferInvitation
  * @property {() => ERef<UserSeat>} getLiquidationSeat
  * @property {() => Notifier<VaultUIState>} getNotifier
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -110,7 +110,7 @@
 /**
  * @typedef {Object} BaseVault
  * @property {() => Amount<NatValue>} getCollateralAmount
- * @property {() => Amount<NatValue>} getDebtAmount
+ * @property {() => Amount<NatValue>} getCurrentDebt
  * @property {() => Amount<NatValue>} getNormalizedDebt
  *
  * @typedef {BaseVault & VaultMixin} Vault
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -36,19 +36,6 @@
  */
 
 /**
- * @typedef  {Object} VaultFactoryPublicFacet - the public facet
- * @property {() => Promise<Invitation<VaultKit>>} makeLoanInvitation
- * @property {() => Promise<Invitation<VaultKit>>} makeVaultInvitation
- * @property {() => Promise<Array<Collateral>>} getCollaterals
- * @property {() => Issuer} getRunIssuer
- * @property {(paramDescription: ParamDescription) => bigint} getNatParamState
- * @property {(paramDescription: ParamDescription) => Ratio} getRatioParamState
- * @property {() => Record<Keyword, ParamShortDescription>} getGovernedParams
- * @property {() => Promise<GovernorPublic>} getContractGovernor
- * @property {(name: string) => Amount} getInvitationAmount
- */
-
-/**
  * @typedef  {Object} VaultFactory - the creator facet
  * @property {AddVaultType} addVaultType
  * @property {() => Promise<Array<Collateral>>} getCollaterals
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -1,5 +1,8 @@
 // @ts-check
 
+/** @typedef {import('./vault').VaultUIState} VaultUIState */
+/** @typedef {import('./vault').VaultKit} VaultKit */
+
 /**
  * @typedef  {Object} AutoswapLocal
  * @property {(amount: Amount, brand: Brand) => Amount} getInputPrice
@@ -55,20 +58,6 @@
  */
 
 /**
- * @typedef {Object} BaseUIState
- * @property {Amount<NatValue>} locked Amount of Collateral locked
- * @property {Amount<NatValue>} debt Amount of Loan (including accrued interest)
- */
-
-/**
- * @typedef {BaseUIState & LiquidationUIMixin} VaultUIState
- * @typedef {Object} LiquidationUIMixin
- * @property {Ratio} interestRate Annual interest rate charge
- * @property {Ratio} liquidationRatio
- * @property {'active' | 'liquidating' | 'liquidated' | 'closed' | 'transfer'} vaultState
- */
-
-/**
  * @callback ReallocateReward
  *
  * Transfer the indicated amount to the vaultFactory's reward
@@ -120,22 +109,6 @@
  * @property {() => Promise<Invitation>} makeTransferInvitation
  * @property {() => ERef<UserSeat>} getLiquidationSeat
  * @property {() => Notifier<VaultUIState>} getNotifier
- */
-
-/**
- * @typedef {Object} LineOfCreditKit
- * @property {Notifier<BaseUIState>} uiNotifier
- * @property {BaseVault} vault
- * @property {{
- *    AdjustBalances: () => Promise<Invitation>,
- *    CloseVault: () => Promise<Invitation>,
- *  }} invitationMakers
- */
-
-/**
- * @typedef {Object} VaultKit
- * @property {Vault} vault
- * @property {Notifier<VaultUIState>} vaultNotifier
  */
 
 /**
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -135,7 +135,7 @@
 /**
  * @typedef {Object} VaultKit
  * @property {Vault} vault
- * @property {Notifier<VaultUIState>} uiNotifier
+ * @property {Notifier<VaultUIState>} vaultNotifier
  */
 
 /**
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -640,7 +640,7 @@ export const makeInnerVault = (
     outerUpdater = updater;
     updateUiState();
     return harden({
-      uiNotifier: vault.getNotifier(),
+      vaultNotifier: vault.getNotifier(),
       invitationMakers: Far('invitation makers', {
         AdjustBalances: vault.makeAdjustBalancesInvitation,
         CloseVault: vault.makeCloseInvitation,
@@ -650,7 +650,10 @@ export const makeInnerVault = (
     });
   };
 
-  /** @type {OfferHandler} */
+  /**
+   * @param {ZCFSeat} seat
+   * @param {InnerVault} innerVault
+   */
   const initVaultKit = async (seat, innerVault) => {
     assert(
       AmountMath.isEmpty(debtSnapshot.run),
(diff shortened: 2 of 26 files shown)


Chris-Hibbert:
Seeing blank *Security Considerations*, *Documentation Considerations*, and *Testing Considerations* makes me squeamish. If there aren't any implications for security, say why. This is a change in behavior that affects dapps and the UI. It's okay to say that something needs to be written in a section of the docs that doesn't exist yet. It's also okay to put off writing the docs, but at least say who will need to know about the change or who will be interested in the behavior when we have users and developers.

It's exceedingly rare that actual code changes don't need to be tested. I sometimes write a cursory description of the testing, but I like to at least think about which behavior should have regression tests, and how extensive the tests for ill-behaved inputs should be.

It also concerns me when code is sent for review without passing tests. If the test  failure can be safely ignored, please explain why.  I sometimes wait for tests to be green before doing a thorough review, presuming the code isn't final yet.

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -34,8 +37,8 @@
 
 /**
  * @typedef  {Object} VaultFactoryPublicFacet - the public facet
- * @property {() => Promise<Invitation>} makeLoanInvitation
- * @property {() => Promise<Invitation>} makeVaultInvitation
+ * @property {() => Promise<Invitation<VaultKit>>} makeLoanInvitation
+ * @property {() => Promise<Invitation<VaultKit>>} makeVaultInvitation
```
Chris-Hibbert: Should one of these be deprecated?   Ah, I see it in `vaultFactory.js`. Shouldn't it also be annotated in the type declaration?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -721,14 +719,13 @@ export const makeInnerVault = (
...
   });
 
   return innerVault;
 };
 
-/**
- * @typedef {ReturnType<typeof makeInnerVault>} InnerVault
- */
+/** @typedef {ReturnType<typeof makeInnerVault>} InnerVault */
+/** @typedef {Unpromise<ReturnType<InnerVault['initVaultKit']>>} VaultKit */
```
Chris-Hibbert, line 731: I haven't seen Unpromise before. What does it mean?

and how does ReturnType work? These two lines seem to be using it in different ways. Why doesn't the return type of makeInnerVault just go on its declaration?

packages/zoe/src/contractFacet/types.js
```diff
@@ -82,6 +82,7 @@
  */
 
 /**
+ * @template {object} [OR=any]
```
Chris-Hibbert: I think templatized declarations should say what their template parameters represent. I had to look around to find that `OR` is `offerResults`. Would you mind saying so here?

packages/zoe/src/contractSupport/interest.js
```diff
@@ -0,0 +1,55 @@
...
+  multiplyRatios,
+} from './ratio.js';
+
+/**
+ *
+ * @param {Ratio} currentCompoundedInterest
+ * @param {Ratio} previousCompoundedInterest
+ * @returns {Ratio} additional compounding since the previous
+ */
+export const calculateRelativeCompounding = (
```
Chris-Hibbert: does this need to be exported?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -864,12 +878,18 @@ test('interest on multiple vaults', async t => {
...
 
+  const assetUpdate = await E(assetNotifier).getUpdateSince();
   const aliceUpdate = await E(aliceNotifier).getUpdateSince();
   const bobUpdate = await E(bobNotifier).getUpdateSince();
+
   // 160 is initial fee. interest is 3n/week. compounding is in the noise.
   const bobAddedDebt = 160n + 3n;
   t.deepEqual(
-    bobUpdate.value.debt,
+    calculateCurrentDebt(
```
Chris-Hibbert, line 890: I think this verifies that `calculateCurrentDebt` was used in calculating Bob's debt. We should also have unit tests showing that `calculateCurrentDebt` produces the correct number.

packages/zoe/src/contractSupport/interest.js
```diff
@@ -0,0 +1,55 @@
...
+  return floorMultiplyBy(debtSnapshot, interestSinceSnapshot);
+};
+
+/**
+ *
+ * @param {Amount<NatValue>} debt
+ * @param {Ratio} interestApplied
+ * @returns {Amount<NatValue>}
+ */
+export const reverseInterest = (debt, interestApplied) => {
```
Chris-Hibbert, line 58: I'd like to see unit tests.

packages/zoe/src/contractSupport/interest.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+/**
+ *
+ * @param {Amount<NatValue>} debtSnapshot
+ * @param {Ratio} interestSnapshot
+ * @param {Ratio} currentCompoundedInterest
+ * @returns {Amount<NatValue>}
+ */
+export const calculateCurrentDebt = (
```
Chris-Hibbert, line 35: I'd like to see unit tests.

packages/run-protocol/src/vaultFactory/types.js
turadg: Ideally yes, but unfortunately JSDoc can only mark a whole symbol as deprecated https://jsdoc.app/tags-deprecated.html

I tried moving converting this embedded TS syntax, but it's still embedded in a JSDoc so the `/** @deprecated */` comment can't be used.

turadg: Ah, I found a solution with using the implementation to define the type. The TS and the IDE can see what property is deprecated:
Screen Shot 2022-02-28 at 7 41 27 AM

I also think it's a code quality and velocity improvement to define types as they are returned by maker functions (like class constructors).

packages/zoe/src/contractFacet/types.js
turadg: Good call. I didn't realize I could have more comments on this line without confusing tsc.

packages/zoe/src/contractSupport/interest.js
turadg: not strictly. I thought it might be useful but I'll unexport and leave that to when someone needs it.

packages/run-protocol/src/vaultFactory/types.js
Chris-Hibbert: > I also think it's a code quality and velocity improvement to define types as they are returned by maker functions (like class constructors).

Sorry, I didn't get this.  Which part is an improvement? How does it help? Are we using something like class constructors?

packages/zoe/src/contractSupport/interest.js
Chris-Hibbert: My thought was more that if it's exported, then it needs to be tested separately. If it's only used internally, then testing the external method that uses it should be sufficient.

If you expect it to be separately useful, it would be helpful to add tests for it.

turadg:
> Seeing blank Security Considerations, Documentation Considerations, and Testing Considerations makes me squeamish.

Good :)  I submitted this as non-draft in error. Thanks for the advice on how to fill them out.

> It also concerns me when code is sent for review without passing tests. If the test failure can be safely ignored, please explain why. I sometimes wait for tests to be green before doing a thorough review, presuming the code isn't final yet.

Also an error. The tests were passing locally, but due to a stale build for the swingSet tests. 

I've made the requested changes in fixup commits so you can see them. After approval I'll do an [autosquash](https://thoughtbot.com/blog/autosquashing-git-commits) to clean them up for merging to master.

packages/zoe/src/contractSupport/interest.js
```diff
@@ -0,0 +1,60 @@
...
+  floorDivideBy,
+  floorMultiplyBy,
+  invertRatio,
+  multiplyRatios,
+  ratiosSame,
+} from './ratio.js';
+
+/**
+ *
+ * @param {Ratio} currentCompoundedInterest as coefficient
```
turadg, line 13: @Chris-Hibbert WDYT of having a type specifically for "interest rate" so that this qualification isn't necessary? It could also make it impossible for the numerator and denominator to have different brands.

packages/run-protocol/src/vaultFactory/types.js
turadg: I'll try to give more context:

`makeFoo` is like a `Foo` class constructor. If we weren't programming in SES we'd be using classes. When you have types for classes, you rely on the class definition as the type.

We have been using a typedef to say what `Foo` is and then saying that `makeFoo` returns that. That's kind of like defining an interface that only ever gets implemented by one class. Instead, we can use `ReturnType` to say "type `Foo` is whatever `makeFoo` returns".

That's an improvement because:
- when you see `Foo` you can command-clicked once or twice to get to the actual implementation
- changes to what `makeFoo` returns don't require also revising the interface; what `makeFoo` returns is authoritative for what type `Foo` is

packages/run-protocol/src/vaultFactory/vault.js
turadg: > I haven't seen `Unpromise` before. What does it mean?

It's a handy type @michaelfig defined for us in https://github.com/Agoric/agoric-sdk/commit/20941e675302ee5905e4825638e661065ad5d3f9#diff-113f01ea0b5f91ff269bcc4b0a0b5f8391d8efc186eeb13197086dd9ad39aabcR8 . It unwraps a promise type to what it promises to return.

> and how does `ReturnType` work?

It evaluates to the return type of the argument (which must be a function type).

> These two lines seem to be using it in different ways.

Is the difference you're referring to the use of `typeof`? That's because `makeInnerVault` is a value, not a type. It needs the `typeof` to extract the type from the value. `InnerVault` is already a type.

> Why doesn't the return type of `makeInnerVault` just go on its declaration?

It would be nice to have a way to jump there. I figure it doesn't because the type just happens to be a return type. The normal behavior for jumping to a type is its declaration, which is here. Though this is better than ambient types in that we can continue to click to get to the implementation.

I tried defining the type at the site of its implementation, but defining it within a function (`start`) limits the scope of the typedef to within that function.

Chris-Hibbert: > Is the difference you're referring to the use of typeof? That's because makeInnerVault is a value, not a type. It needs the typeof to extract the type from the value. InnerVault is already a type.

As I read these, they're different forms of saying 'infer the return type from the code.'  Neither points to an actual declaration, AFAICT.

`@typedef {ReturnType<typeof makeInnerVault>} InnerVault` says that `InnerVault` is whatever is returned by `makeInnerVault`.  `makeInnerVault` has jsDoc that declares its parameters, but no `@returns`, so I presume typescript infers it from the code.  Then `ReturnType<InnerVault['initVaultKit']>`  is a reference to the `initVaultKit` method in InnerVault, which also has no declaration, so tsc infers this from the code as well.

I hadn't previously seen either `typeof` or a reference to a method of an object written in these ways. And both bottom out in type inference rather than type declaration. I guess that saves maintenance work.

Chris-Hibbert:
There are two changes mentioned in **Documentation Considerations**.  Where should those get written up? Are there changes needed to the doc site? Should we announce them on Discord?

If changes are needed on the Doc site, I'd appreciate an issue if it won't be done by the time this PR is ready for approval.

packages/zoe/src/contractSupport/interest.js
Chris-Hibbert: [after much out-of-band discussion]

I'm still not clear on what "as coefficient" is intended to clarify. But I think it would be reasonable to have a sub-type of Ratio that only had a single brand, and required that its operands be of that brand.
Making all the functions in `ratio.js` work with both types looks like it would be a chore.

Chris-Hibbert (approved):
Thanks for the additional tests, and for bearing with me through the questions.

LGTM.

turadg:
> Where should those get written up?

I believe that by adhering to conventional commits they'll automatically be noted in CHANGELOG.md upon release.

> Are there changes needed to the doc site?

Yes, `uiNotifier` appears in https://github.com/Agoric/documentation/blob/c73a3c5aa80fdc6c6571658461d6f30a313cc49a/main/zoe/guide/contracts/treasury.md?plain=1#L75-L81 

- need to rename `uiNotifier` to `vaultNotifier`
- need to add `assetNotifier`
- need to add `TransferVault` to the invitation set (done in a previous PR)

I've filed in https://github.com/Agoric/documentation/issues/630

> Should we announce them on Discord?

I don't see a reason to but I'm new here and open to being educated on this point.

PR #4696, opened 2022-02-28
Remove possible spurious loss of slotToVal entries on export retirement

### LINKED ISSUE #3978: retireOneExport should not delete valToSlot
As @FUDCo and I were reviewing #3732 , in particular https://github.com/Agoric/agoric-sdk/pull/3968#discussion_r729187128 , we concluded that the `retireOneExport()` function in liveslots is currently doing the wrong thing. When the kernel retires a Remotable, liveslots should simple remove that vref from `kernelRecognizableRemotables`, and *not* touch `valToSlot` or `slotToVal`.

Previously (before virtual objects), I was using the presence or absence of a vref in `slotToVal` as an indicator of the kernel's interest in an object. Back then, if the kernel couldn't recognize it, we didn't need to maintain the vref-to-Remotable mapping: re-exporting the object could give it a new vref and the kernel couldn't tell the difference (because it had already forgotten the old vref, when it did the `retire`).

But now, the following sequence might happen:
* a Remotable is created, and exported to the kernel
  * serializing the new Remotable causes a vref to be assigned and `valToSlot`/`slotToVal` to be populated
* the same Remotable gets stored in some virtualized data
  * this adds the vref to the DB, and increments a refcount for it
* the kernel drops and retires the exported vref
* someone stores the Remotable in some other bit of virtualized data (serializing it), and the serialized data should use the same vref as before
* or, someone reads the old virtualized data, deserializes the old vref, and expects to get back the same Remotable as before

If `retireOneExport` were to delete the `valToSlot` entry, that last `serialize` would allocate a new vref, and that last `deserialize` would crash because the stored vref was no longer in `slotToVal`.

The fix will be to replace that entire "if (isRemotable)" clause of `retireOneExport` with a simple `kernelRecognizableRemotables.delete(vref)`, and leave slotToVal/valToSlot alone. And the only purpose of `kernelRecognizableRemoteables` is to permit/inhibit the `syscall.retireExports` during `processDeadSet`.

The more interesting part will be to write the unit test for this. It ought to follow the sequence above: first attempting to deserialize the old VO (and asserting that it doesn't crash), then serialize the old Remotable into a new VO and compare the backing store / DB entries to assert that the vrefs are the same.

FUDCo:
Removed `slotToVal` and `valToSlot` manipulations from `retireOneExport`, as they are redundant with what finalization does and caused inappropriate loss of vref mappings in some circumstances.

This also cleans up some unfortunate naming that contributed to confusion in diagnosing what was going on.

Closes #3978

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -84,7 +84,7 @@ function build(
    * and use a FinalizationRegistry to learn when the vat has dropped it, so
    * we can notify the kernel. We retain strong references to unresolved
    * Promises. When an import is added, the finalizer is added to
-   * `droppedImportRegistry`.
+   * `vreffedObjectRegistry`.
    *
    * slotToVal is a Map whose keys are slots (strings) and the values are
    * WeakRefs. If the entry is present but wr.deref()===undefined (the
@@ -174,7 +174,7 @@ function build(
 
   */
 
-  function finalizeDroppedImport(vref) {
+  function finalizeDroppedObject(vref) {
     // TODO: Ideally this function should assert that it is not metered.  This
     // appears to be fine in practice, but it breaks a number of unit tests in
     // ways that are not obvious how to fix.
@@ -196,7 +196,7 @@ function build(
       slotToVal.delete(vref);
     }
   }
-  const droppedImportRegistry = new FinalizationRegistry(finalizeDroppedImport);
+  const vreffedObjectRegistry = new FinalizationRegistry(finalizeDroppedObject);
 
   async function scanForDeadObjects() {
     // `possiblyDeadSet` accumulates vrefs which have lost a supporting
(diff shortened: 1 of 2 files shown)


warner (changes requested):
Let's simplify it further by removing the rest of that `retireOneExport` clause: no `meterControl.assertNotMetered()`, no `const wr = slotToVal.get(vref)`, no check of `wr.deref()`, no check of `exportedRemotables.has(val)`. Just
```js
} else {
  // Remotable
  kernelRecognizableRemotables.delete(vref);
}
```

I think the early return if `exportedRemotables.has(val)` would be wrong. It's in a clause where the kernel has already done something wrong (`dispatch.retireExports` without `dispatch.dropExports`, or maybe the vat re-exported the object after `dispatch.dropExports` and the kernel didn't squelch the `retireExports` like it's supposed to), but still the least-wrong thing to do is to still remove it from `kernelRecognizableRemotables`.

FUDCo:
Radical.  I like it.  Ready for re-review.

warner (approved):
Great!

PR #4697, opened 2022-02-28
experiment: refactor vault to API of virtual objects

turadg:
_no ticket_

## Description

For consideration of doing before Restival until we can fully virtualized the run-protocol objects. This introduces a `nonvirtualStore` that has a `defineKind` that matches the same API but without the virtualization abiltiies.

PR #4705, opened 2022-03-01
fix: emulate multi-facet support

erights:
Adapted @turadg 's emulated `defineKind` from #4697 into the `defineHeapKind` seen here, to emulate upcoming multi-facet support.

Use it to virtualize (or pretend to virtualize) purses, payments, ledgers

Compare to #4618 

Suggestion to reviewers: review hiding whitespace differences.

### DIFF
--- a/packages/ERTP/src/purse.js
+++ b/packages/ERTP/src/purse.js
@@ -1,43 +1,60 @@
 import { makeNotifierKit } from '@agoric/notifier';
-import { Far } from '@endo/marshal';
+import { defineHeapKind } from '@agoric/store';
 import { AmountMath } from './amountMath.js';
 
-export const makePurse = (allegedName, assetKind, brand, purseMethods) => {
-  let currentBalance = AmountMath.makeEmpty(brand, assetKind);
+export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
+  const makePurseKit = defineHeapKind(
+    allegedName,
+    () => {
+      const currentBalance = AmountMath.makeEmpty(brand, assetKind);
 
-  /** @type {NotifierRecord<Amount>} */
-  const { notifier: balanceNotifier, updater: balanceUpdater } =
-    makeNotifierKit(currentBalance);
+      /** @type {NotifierRecord<Amount>} */
+      const { notifier: balanceNotifier, updater: balanceUpdater } =
+        makeNotifierKit(currentBalance);
 
-  const updatePurseBalance = newPurseBalance => {
-    currentBalance = newPurseBalance;
-    balanceUpdater.updateState(currentBalance);
-  };
-
-  /** @type {Purse} */
-  const purse = Far(`${allegedName} purse`, {
-    deposit: (srcPayment, optAmountShape = undefined) => {
-      // Note COMMIT POINT within deposit.
-      return purseMethods.deposit(
+      return {
         currentBalance,
-        updatePurseBalance,
-        srcPayment,
-        optAmountShape,
-      );
+        balanceNotifier,
+        balanceUpdater,
+      };
     },
-    withdraw: amount =>
-      // Note COMMIT POINT within withdraw.
-      purseMethods.withdraw(currentBalance, updatePurseBalance, amount),
-    getCurrentAmount: () => currentBalance,
-    getCurrentAmountNotifier: () => balanceNotifier,
-    getAllegedBrand: () => brand,
-    // eslint-disable-next-line no-use-before-define
-    getDepositFacet: () => depositFacet,
-  });
+    state => {
+      const { balanceNotifier, balanceUpdater } = state;
+      const updatePurseBalance = newPurseBalance => {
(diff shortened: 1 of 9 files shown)


packages/SwingSet/src/storeModule.js
```diff
@@ -11,6 +11,7 @@ export const {
 
 export {
   M,
+  defineHeapKind,
```
warner, line 14: Where does this `defineHeapKind` come from? Would we need to change swingset to import it from `store` and add it to the global `VatData`?

I'm hoping we can clean up the relationship between `store`, swingset, and downstream clients. Having ERTP depend upon swingset seems like a bad idea, but I'm not sure to to split the definition of the extended stores (their types, really), the implementation of the basic forms (which don't require swingset), the mock implementations of the more complex types (for unit tests), and the module that exposes the real implementation of the complex types (from swingset).

erights: I just cargo culted what we did for heap stores, which I like.

`@agoric/store` should not depend on swingset. The heap-only abstractions should not depend on swingset.

packages/ERTP/src/purse.js
```diff
@@ -1,43 +1,60 @@
...
+      const updatePurseBalance = newPurseBalance => {
+        state.currentBalance = newPurseBalance;
+        balanceUpdater.updateState(state.currentBalance);
+      };
 
-  return purse;
+      /** @type {Purse} */
+      const purse = {
+        deposit: (srcPayment, optAmountShape = undefined) => {
+          // Note COMMIT POINT within deposit.
```
warner, line 31: That comment isn't clear to my inexperienced eyes. Does it mean "please take note of the (surprising) commit point that exists within `deposit`? Maybe phrase that "Note: deposit() has an internal commit point".

packages/store/src/defineHeapKind.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    harden(wrappedData);
+
+    const representative = actualize(wrappedData);
+    return [representative, wrappedData];
+  }
+
+  function makeNewInstance(...args) {
+    const objID = `nonvirtual/${nextInstanceID}`;
+    nextInstanceID += 1;
```
FUDCo, line 52: If you're totally faking it, then I think you can get rid of `objID` and `nextInstance` entirely.

packages/store/src/defineHeapKind.js
```diff
@@ -0,0 +1,72 @@
...
+      Object.defineProperty(wrappedData, prop, {
+        get: () => {
+          return innerSelf.rawData[prop];
+        },
+        set: value => {
+          innerSelf.rawData[prop] = value;
+        },
+      });
+    }
+    harden(wrappedData);
```
FUDCo, line 44: Note that even though you're wrapping the data with getters and setters, you're not passing them through the marshaller and so you're not getting any of the arms-length handling that gives you.  In particular, you can set the value of a property and then mutate it in place, which you can't do with the other kinds of virtual objects.  If the goal is to emulate the constraints of the real thing so that code can be implemented to the real API, then you should probably at least be doing something to emulate the immutability of the values stored in virtual object properties.  I'm not sure how you'd do that, but if you don't then somebody is going to be surprised when they write something like `state.foo.push(47)` and have it work using `defineHeapKind` only to have it blow up in their faces when they switch to `defineKind`.

FUDCo: It strikes me that what is implemented by SwingSet's test tooling (`fakeVirtualSupport.js` et al) is an always up to date full faith reproduction except in RAM, because it uses the actual virtual object manager only backed by a JavaScript `Map` rather than by the database.  You can't quite get away with using it directly since it also makes its own copy of LiveSlots and you'd want to use the real LiveSlots, but it's possible it could be modified to separate the storage so it can be backed by a `Map` even when the rest of your vat is saving stuff to disk.  Or at least it might be a starting point for a less brittle emulation.

FUDCo: On the other hand, if you don't mind having mutable state properties, then you can probably simplify things by ditching the getter/setter wrapper.

packages/store/src/defineHeapKind.js
```diff
@@ -0,0 +1,72 @@
+// @ts-check
+
+import { Far } from '@endo/marshal';
+
+const { ownKeys } = Reflect;
+
+const becomePassable = (tag, thing) => {
```
Chris-Hibbert, line 7: Even for internal use, this deserves at least a short comment. Its arguments can be of two different kinds, which are distinguished by sniffing, and handled differently.

Chris-Hibbert:
When implementing a new persistent storage mechanism, and applying it to new objects, I'd expect to see a test that shows storing objects and retrieving them from the store. (When I write a persistent object I expect to need that to be sure I haven't missed anything.)  Is that here, and I just missed it?

erights:
The thing to emulate has changed enough that, when we do write a heap version, we should start over from scratch. Closing this.

Issue #4382, opened 2022-01-25
design better vat-facing upgrade protocol (buildRootObject()? upgrade()?)


warner:
## What is the Problem Being Solved?

We need to define how a new version of a vat is launched.

We expect the initial version of a vat to export a function named `buildRootObject`, which is called exactly once during that version's lifetime, and is supposed to return a single remotable object. The calling vat will get this object as `root` in the result of `vatAdminService~.createVat()`, so it can start interacting with its creation.

However we haven't yet defined what the second (and subsequent) versions of a vat must do, when triggered by `vatAdminFacet~.upgrade(newBundleCap, newVatParameters)`. We know there will be some well-defined calling pattern, with either one function being exported, or one per version upgrade. This function will get `vatPowers` just like `buildRootObject`, and it will get a `vatParameters` from the upgrading caller. The function must reassociate all the virtual-object Kinds that were created by the earlier version. I expect it must also reassociate vrefs for any non-virtual "precious" Remotables that were exported by its predecessor (but we might dodge that by instructing vat authors to use virtual objects even for singletons).

#1848 is about what you (as a calling vat, e.g. Zoe) would use to tell the kernel about some other vat that you want to upgrade. This ticket is about what happens within the upgraded vat.

The questions we need to answer include:
* What function (or functions) should each version export?
  * maybe each version exports `buildRootObject`
  * maybe only version-1 uses `buildRootObject`, and version-2++ provides `upgrade` or `upgradeRootObject`
  * maybe each version must export a delta-specific upgrade function, e.g. `upgradeV1ToV2` and `upgradeV2ToV3`
  * or a target-version -specific function: `upgradeToV2`, `upgradeToV3`
  * this has deep consequences for how vat authors organize their source code repository, and how they think about code edits/commits vs deployment/upgrade
* Can vats reassociate arbitrary exported Remotables, or just the root object?
  * What combination of authorities are necessary to perform this reassociation? Per-object/obligation handles in baggage, plus a reassociation facility in `vatPowers`?
  * Ideally this should parallel the way `makeKind` reassociation is performed
  * If there are remaining exports (those not reassociated by the end of the initial crank), should that flunk the upgrade process? Or should they be disowned by the vat? Or should liveslots "ground" them by wiring them (locally) to error-producing dead-ends?

## Description of the Design

[...]

PR #4706, opened 2022-03-01
fix: virtualize pool

erights:
Virtualize pool. Unlike #4707, this one does not virtualize vpool (the single pool) because it seems that the vpool need not be made persistent.

See https://github.com/Agoric/agoric-sdk/pull/4707#discussion_r831615803

Note to reviewers: I may still be confused. I suspect that @Chris-Hibbert and I need to pair program this before it is good.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/pool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/pool.js
@@ -1,9 +1,9 @@
 // @ts-check
 
 import { E } from '@endo/eventual-send';
-import { Far } from '@endo/marshal';
 import { AssetKind, AmountMath, isNatValue } from '@agoric/ertp';
 import { makeNotifierKit } from '@agoric/notifier';
+import { defineKind } from '@agoric/swingset-vat/src/storeModule.js';
 
 import {
   calcLiqValueToMint,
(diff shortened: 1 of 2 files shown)


Chris-Hibbert:
I'm expecting to see a test demonstrating that objects can be defined, saved, flushed from memory, and restored with behavior identical to the original. Is there a reason that's not appropriate here?

erights:
At https://github.com/Agoric/agoric-sdk/pull/4706#issuecomment-1055681522 @Chris-Hibbert asks:

> I'm expecting to see a test demonstrating that objects can be defined, saved, flushed from memory, and restored with behavior identical to the original. Is there a reason that's not appropriate here?

@FUDCo , should this PR test that? If so, how? Thanks.

erights:
Converting back to draft because it is only a step towards #4707 . I'll let you know when #4707 is ready for review.

erights:
Given @Chris-Hibbert 's comments at https://github.com/Agoric/agoric-sdk/pull/4707#discussion_r831615803 I put #4707  into draft status. Instead, it seems this PR #4706 is closer to what we want, so I'm declaring #4706  Ready to Review. Everyone, please review this one instead.

erights:
Changed my mind again. Back to Draft until @Chris-Hibbert and I pair on this and it becomes ready. No need for others to review before then.

warner:
> At [#4706 (comment)](https://github.com/Agoric/agoric-sdk/pull/4706#issuecomment-1055681522) @Chris-Hibbert asks:
> 
> > I'm expecting to see a test demonstrating that objects can be defined, saved, flushed from memory, and restored with behavior identical to the original. Is there a reason that's not appropriate here?
> 
> @FUDCo , should this PR test that? If so, how? Thanks.

We don't have a way to exercise that until #1848 / #4382 is done and we can upgrade vats. We've got test coverage on durable objects/collections in general via `packages/SwingSet/` .. that will have to be enough until we get upgrade ready for use.

Although.. I suppose if you had a swingset-based test in which the virtual/durable collections were used to create some objects in vat A, then you carefully drop all references to them from RAM (leaving only exports and/or refs from virtualized data), then you wait a crank or two, then send a message back into vat A that references those objects.. that would exercise the code that rebuilds the objects from saved data.

The tricky part is knowing that the original objects really got GCed, which userspace doesn't get to see. You could add some temporary debug messages to liveslots.js while testing the test, to convince yourself that GC is really happening.

Running the workers under XS (`config.defaultManagerType = 'xs-worker'`) is a good idea, that makes GC far more predictable.

Chris-Hibbert:
> We don't have a way to exercise that until https://github.com/Agoric/agoric-sdk/issues/1848 / https://github.com/Agoric/agoric-sdk/issues/4382 is done and we can upgrade vats. We've got test coverage on durable objects/collections in general via packages/SwingSet/ .. that will have to be enough until we get upgrade ready for use.

I was more interested in virtualization than durability. Can we page objects out and back in at this point?

packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -43,210 +43,257 @@ export const makeAddPool = (
...
 
-        const tokenKeywordAmountOut = AmountMath.make(
-          secondaryBrand,
-          calcValueToRemove(
-            liqTokenSupply,
-            pool.getSecondaryAmount().value,
-            liquidityValueIn,
-          ),
-        );
+  const makePool = defineKind(
```
Chris-Hibbert, line 104: is the best documentation on `defineKind` the comment on `defineKindInternal`  in `virtualObjectManager.js`?   The parameter types don't seem to be declared in a way that's visible to VSCode and WebStorm, so I always have to find the docs before I can tell what arguments should be provided in what order.

packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -43,210 +43,257 @@ export const makeAddPool = (
...
+    'pool',
+    (liquidityZcfMint, poolSeat, secondaryBrand) => {
```
Chris-Hibbert, line 106: I'm trying to understand `defineKind`.  How does it know what parameters to provide to the `init` function? (i.e. If the pool doesn't declare them somewhere, what rules produce the parameters?)

packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -43,210 +43,257 @@ export const makeAddPool = (
...
+        poolSeat,
+        liquidityBrand,
+        secondaryBrand,
+        updater,
+        notifier,
+      } = state;
+
+      /** @type {XYKPool} */
+      const pool = {
+        getLiquiditySupply: () => state.liqTokenSupply,
```
Chris-Hibbert, line 138: why is `liqTokenSupply` access through `state`, but `liquidityIsuser`, `poolSeat`, `updater`, etc. (line 126) are extracted and accessed as top-level values?

Chris-Hibbert (commented):
This looks fine and straightforward. I asked one question about how `makeKind` works.

Do we have the ability to demonstrate that virtualization can handle paging objects out and back in again? 

I've asked before and it seems like the responses have assumed I was asking about durability and revival of the contract rather than the pool objects.

packages/run-protocol/src/vpool-xyk-amm/pool.js
turadg: https://github.com/Agoric/agoric-sdk/pull/4603/commits/48dbf0f1fa8b86ee3b6781c4e200fc16cfa854e0#diff-4d09d78ca33d43e2584cfbf5adc00ed1caac51f1edbc10bae3d3c14320e56d30 will help

packages/run-protocol/src/vpool-xyk-amm/pool.js
FUDCo: `defineKind` doesn't "know" the parameters to `init`.  The maker function calls `init` with whatever parameters _it_ was given.

FUDCo:
> At [#4706 (comment)](https://github.com/Agoric/agoric-sdk/pull/4706#issuecomment-1055681522) @Chris-Hibbert asks:
> 
> > I'm expecting to see a test demonstrating that objects can be defined, saved, flushed from memory, and restored with behavior identical to the original. Is there a reason that's not appropriate here?
> 
> @FUDCo , should this PR test that? If so, how? Thanks.

I don't think it should have to be the responsibility of users of the virtual object system to write tests to verify that the virtual object system works as advertised.  The virtual objects implementation itself does have tests for that though.

Chris-Hibbert:
> I don't think it should have to be the responsibility of users of the virtual object system to write tests to verify that the virtual object system works as advertised. The virtual objects implementation itself does have tests for that though.

I'm willing to accept that it works as advertised. As a reviewer of a conversion, I'm trying to gain confidence that the right state has been captured. I'd feel a lot more confident if I saw that objects that had been paged out and in behaved the same as before. Without that I have to have confidence that I've identified all the state that is in use, and that I understand how the parts of makeKind interact in great detail.

FUDCo:
> > I don't think it should have to be the responsibility of users of the virtual object system to write tests to verify that the virtual object system works as advertised. The virtual objects implementation itself does have tests for that though.
> 
> I'm willing to accept that it works as advertised. As a reviewer of a conversion, I'm trying to gain confidence that the right state has been captured. I'd feel a lot more confident if I saw that objects that had been paged out and in behaved the same as before. Without that I have to have confidence that I've identified all the state that is in use, and that I understand how the parts of makeKind interact in great detail

Ah, I misunderstood the intent of what you were trying to get at.  That makes a lot of sense and seems like a thing that's genuinely important to verify.

erights:
@FUDCo 
> Ah, I misunderstood the intent of what you were trying to get at. That makes a lot of sense and seems like a thing that's genuinely important to verify.

Is there a way to verify that? Should this PR do so? How?

turadg:
Done in #5187

PR #4708, opened 2022-03-01
refactor kernel message-routing code

warner:
Small refactorings to the kernel message delivery process.. trying to pave the way for separate vat input/output queues (to support #3465) and a cleaner framework for vat-upgrade performing multiple deliveries (#1848). This is not intended to change any behavior or message ordering.

The `routeSendEvent()` function is a start at the `route(send)` action described in https://github.com/Agoric/agoric-sdk/issues/3465#issuecomment-1055670456 , and `deliverRunQueueEvent` could become the code that runs at the start of a "routing crank".

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -746,6 +746,22 @@ export default function buildKernel(
     }
   }
 
+  function decrementSendEventRefCount(message) {
+    kernelKeeper.decrementRefCount(message.target, `deq|msg|t`);
+    if (message.msg.result) {
+      kernelKeeper.decrementRefCount(message.msg.result, `deq|msg|r`);
+    }
+    let idx = 0;
+    for (const argSlot of message.msg.args.slots) {
+      kernelKeeper.decrementRefCount(argSlot, `deq|msg|s${idx}`);
+      idx += 1;
+    }
+  }
+
+  function decrementNotifyEventRefCount(message) {
+    kernelKeeper.decrementRefCount(message.kpid, `deq|notify`);
+  }
+
   const gcMessages = ['dropExports', 'retireExports', 'retireImports'];
 
   async function deliverRunQueueEvent(message) {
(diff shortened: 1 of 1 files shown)


packages/SwingSet/src/kernel/kernel.js
```diff
@@ -439,66 +439,6 @@ export default function buildKernel(
...
-        policyInput = await deliverToVat(vatID, target, msg);
-      } else if (msg.result) {
-        resolveToError(msg.result, VAT_TERMINATION_ERROR);
-      }
-    } else if (type === 'promise') {
-      const kp = kernelKeeper.getKernelPromise(target);
-      if (kp.state === 'redirected') {
-        // await deliverToTarget(kp.redirectTarget, msg); // probably correct
-        // TODO unimplemented
-        throw new Error('not implemented yet');
```
mhofman, line 459: I assume we don't need to keep this?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -746,6 +686,88 @@ export default function buildKernel(
...
+  function routeSendEvent(message) {
+    const { target, msg } = message;
+    const { type } = parseKernelSlot(target);
+    assert(
+      ['object', 'promise'].includes(type),
+      X`unable to send() to slot.type ${type}`,
+    );
+
+    function splat(error) {
+      if (msg.result) {
```
mhofman, line 711: Unrelated to the changes in this PR, but it does feel strange to drop errors on the ground if there is nowhere to send them. Have we considered remembering which vat triggered the send, and raise this as a global error there? Probably moot I suppose since live slots automatically creates a promise for the result.

mhofman (approved):
I really like this separation. Much cleaner!

packages/SwingSet/src/kernel/kernel.js
warner: yeah, it was aspirational, but we'll figure something else out if/when we ever implement redirection

packages/SwingSet/src/kernel/kernel.js
warner: Correct, `syscall.send` accepts messages without `.result`, but liveslots has no way to create them, because we have no way to sense that userspace ignored the return from `E().foo()`. @michaelfig had an idea a long while back about putting a WeakRef on the return Promise and watching to see if it got collected before the crank finished (and using that as evidence that we could do a `sendOnly`), but it's fussy enough that we never pursued it. `E.sendOnly(target).method(args)` is still a strong contender though, and I believe the HandledPromise API has a place for it.

I don't really know what to do about the global error handler thing. It's just like a global promise rejection thing. If errors are a normal part of your API (they happen, that's fine, you don't want to make a fuss about it), then it's annoying to have them be announced everywhere (/me casts a stern glance at `marshal`). If they aren't, then even though the sending code doesn't care, the *programmer* should maybe know that something suprising happened, and silently discarding them doesn't help.

For now, I think our stance is "the sending code should care", and we consider it at least a review-worthy problem if someone does `E()` and doesn't `.catch` or `await`. "No error left behind".

From the kernel's point of view, I'm inclined to say that if the vat did `sendOnly`, we'll respect that and not try to work around their decision.

BTW, the only code that does `sendOnly` right now is in devices, like when device-timer tells vat-timer about an alarm going off, and pretty much all of those messages get routed to a built-in vat. So I don't think we have a lot of cases where the kernel dropping an error is going to be provokable by user code.

PR #4711, opened 2022-03-01
refactor(run-protocol)!: vaults hold liquidation proceeds until closed

### LINKED ISSUE #3773: Treasury: merge liquidationSeat into vaultSeat

The Treasury's vault has a separate seat to hold the liquidation payout. This is unnecessary and requires extra work on the part of the borrower.

## Description of the Design

give liquidation the ability to deposit to the vaultSeat. Then the borrower doesn't need an extra seat. They do still need the ability to be notified when liquidation happens.

## Security Considerations

minimal

## Test Plan

refactoring existing tests should be sufficient.

turadg:
closes: #3773

## Description

Removes liquidationSeat from vault. The assets left over from liquidation now rest on the vault. To make this work, a liquidated vault stay open until closed and the payout of proceeds comes in the close offer result.

### Security Considerations

Reduces the seats. Reduces two terminal states to one (named CLOSED).

`liquidationSeat` is mentioned in https://github.com/Agoric/documentation/blob/c73a3c5aa80fdc6c6571658461d6f30a313cc49a/main/zoe/guide/contracts/treasury.md?plain=1#L23-L36

There are many changes coming vaults for MN-1 so it'll be most efficient to rewrite its docs at that time.

### Testing Considerations

Tests updated.


packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -438,11 +438,16 @@ test('first', async t => {
     'vault is cleared',
   );
 
-  const liquidations = await E(
-    E(vault).getLiquidationSeat(),
-  ).getCurrentAllocation();
-  t.deepEqual(liquidations.Collateral, AmountMath.make(aethBrand, 566n));
```
Chris-Hibbert, line 444: I expect this amount to show up in the vaultSeat.  `await E(vaultSeat).getPayout('Collateral')`

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -560,11 +576,16 @@ test('price drop', async t => {
     RUN: AmountMath.make(runBrand, 14n),
   });
 
-  const liquidations = await E(
-    E(vault).getLiquidationSeat(),
-  ).getCurrentAllocation();
-  t.deepEqual(liquidations.Collateral, AmountMath.make(aethBrand, 1n));
-  t.deepEqual(liquidations.RUN, AmountMath.makeEmpty(runBrand));
```
Chris-Hibbert, line 567: make sure that `await E(faucetSeat).getPayouts()` contains 1 aEth and 0 RUN.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -692,11 +713,16 @@ test('price falls precipitously', async t => {
     RUN: AmountMath.make(runBrand, 19n),
   });
 
-  const liquidations = await E(
-    E(vault).getLiquidationSeat(),
-  ).getCurrentAllocation();
-  t.deepEqual(liquidations.Collateral, AmountMath.make(aethBrand, 1n));
-  t.deepEqual(liquidations.RUN, AmountMath.makeEmpty(runBrand));
```
Chris-Hibbert, line 699: check the payouts here, too.

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -42,15 +41,14 @@ const liquidate = async (zcf, vault, burnLosses, strategy, collateralBrand) => {
...
-    liquidationSeat,
   );
   trace(` offeredTo`, collateralToSell, runDebt);
 
   // await deposited, but we don't need the value.
   await Promise.all([deposited, E(liqSeat).getOfferResult()]);
 
   // Now we need to know how much was sold so we can pay off the debt
-  const runProceedsAmount = liquidationSeat.getAmountAllocated('RUN', runBrand);
+  const runProceedsAmount = vaultSeat.getAmountAllocated('RUN', runBrand);
```
Chris-Hibbert: I would add a comment here that we can use this because only liquidation adds RUN to the vaultSeat.

Chris-Hibbert (commented):
Moving the tests was the right thing to do.

We need to verify that the liquidation proceeds are available to the vault holder. I expect them to be in `vaultSeat`.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -544,27 +544,69 @@ test('price drop', async t => {
...
+  const emptyPurse = await E(services.runKit.issuer).makeEmptyPurse();
+  /** @type {UserSeat<string>} */
+  const closeSeat = await E(zoe).offer(
+    E(vault).makeCloseInvitation(),
+    harden({
+      give: { RUN: AmountMath.makeEmpty(runBrand) },
+      want: { Collateral: AmountMath.makeEmpty(aethBrand) },
+    }),
+    harden({
+      // ??? is there a better way?
```
turadg: @Chris-Hibbert is there a simpler way to get these amounts?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -682,21 +724,65 @@ test('price falls precipitously', async t => {
...
   });
 
-  const liquidations = await E(
-    E(vault).getLiquidationSeat(),
-  ).getCurrentAllocation();
-  t.deepEqual(liquidations.Collateral, AmountMath.make(aethBrand, 1n));
-  t.deepEqual(liquidations.RUN, AmountMath.makeEmpty(runBrand));
+  t.deepEqual(
+    await E(vault).getCollateralAmount(),
+    // ??? why is there any collateral left when there's still debt?
```
turadg: @Chris-Hibbert can you help me to understand this? I thought liquidation used up collateral until debt was covered (or there was none left)

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -148,16 +142,12 @@ export const makeInnerVault = (
     phase = newPhase;
   };
 
-  const assertPhase = allegedPhase => {
+  const assertVaultIsOpen = () => {
     assert(
-      phase === allegedPhase,
-      X`vault must be ${allegedPhase}, not ${phase}`,
+      phase === VaultPhase.ACTIVE || phase === VaultPhase.LIQUIDATED,
```
Chris-Hibbert, line 150: `VaultIsOpen` seems like the wrong name for `ACTIVE || LIQUIDATED`. Can we come up with a better one?  Is this `assertVaultCanBeClosed`? (Not that that rolls off the tongue.)

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -349,43 +340,58 @@ export const makeInnerVault = (
...
+      assert(
+        AmountMath.isGTE(runReturned, currentDebt),
+        X`You must pay off the entire debt ${runReturned} > ${currentDebt}`,
+      );
 
-    // Return any overpayment
+      // Return any overpayment
+
+      const { zcfSeat: burnSeat } = zcf.makeEmptySeatKit();
+      burnSeat.incrementBy(seat.decrementBy(harden({ RUN: currentDebt })));
```
Chris-Hibbert: I don't think you need to move `currentDebt` to a new seat before burning. You can burn them from `seat` directly.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: You do have to close the vault (by exercising a closeInvitation). Once that's done, you can ask for the current allocation of that seat (which you do on line 593), but it's not much more work to actually withdraw it, and that's what actual clients would do.

It would be fine to drop 593-597. They don't prove anything beyond what's revealed by getting the payouts, and getting their amounts.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: Good catch!

liquidation wants to use up the collateral until the debt is covered. The AMM is giving the liquidator a price break here.

The liquidator wasn't able to get enough for the 400 aEth to cover the 398 RUN debt, so it said to the AMM "sell the 400 aEth for whatever we can get".  The AMM said "you can get 325 RUN for that, but you only have to pay 399 aEth to get 325 RUN", so it refunded one aEth. 

Since there's a remaining debt, the vaults should keep the extra collateral rather than returning it. We now have enough mechanism in place that it makes sense to hold onto it. In fact, when @dtribble's batched liquidation comes in, we'll be able to add the unsold collateral to what remains, and offer it with the next tranche.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Good call. Since the vault context is implied I made it simply `assertCloseable`. That highlighted that some other places really needed `assertActive` so I restored that too and renamed the existing `assertActive` which wasn't about a vault state to `assertOwned` because it's about whether the outer vault owns an innervault to transfer.

packages/run-protocol/src/vaultFactory/vault.js
turadg: The change here is only indenting, but I gave it a try removing `burnSeat` and the "close loan" test errors that,
```
The amount to be subtracted {"brand":"[Alleged: RUN brand]","value":"[5250n]"} was greater than the allocation\'s amount {"brand":"[Alleged: RUN brand]","value":"[0n]"} for the keyword "RUN"
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -46,57 +43,55 @@ export const VaultPhase = /** @type {const} */ ({
...
  *
  * @param {InnerVault | null} inner
  */
 const makeOuterKit = inner => {
+  /** @type {NotifierRecord<VaultUIState>} */
   const { updater: uiUpdater, notifier } = makeNotifierKit();
 
-  const assertActive = v => {
+  /** @returns true iff this outer vault is backed by an inner vault */
+  const assertOwned = v => {
```
Chris-Hibbert, line 76: The terminology seems reversed here. The query is about the outer vault, which either owns or doesn't own the inner vault. It's not the one that might be 'owned'.

Why isn't 'active' a good term for this?

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: I tried getting rid of `burnSeat` and leaving only
```
      runMint.burnLosses(harden({ RUN: currentDebt }), seat);
      seat.incrementBy(
        vaultSeat.decrementBy(
          harden({ Collateral: getCollateralAllocated(vaultSeat) }),
        ),
      );
      zcf.reallocate(seat, vaultSeat);
```
and the test passes.   The one drawback of this is that it doesn't have a single atomic reallocation to move the debt off the seat and return the collateral. But the current code does the burn non-atomically, so I don't think it leaves us in a more robust state.

turadg:
@Chris-Hibbert ready for another review. the two test failures are,
- flaky integration test for getting-started
- conventional commits noise (until https://github.com/Agoric/agoric-sdk/issues/4721 )

I do plan to do a merge commit and I'll autosquash upon approval.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -344,54 +338,70 @@ export const makeInnerVault = (
...
+      // Simply reallocate vault assets to the offer seat.
+      // Don't take anything from the offer, even if vault is underwater.
+      seat.incrementBy(
+        vaultSeat.decrementBy(
+          harden({
+            Collateral: getCollateralAllocated(vaultSeat),
+            RUN: getRunAllocated(vaultSeat),
+          }),
+        ),
+      );
```
```suggestion
      seat.incrementBy(vaultSeat.decrementBy(vaultSeat.getCurrentAllocation()));
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -34,6 +33,9 @@ import {
 import { startVaultFactory } from '../../src/econ-behaviors.js';
 import '../../src/vaultFactory/types.js';
 import * as Collect from '../../src/collect.js';
+import { calculateCurrentDebt } from '../../src/interest-math.js';
+
+// #region Support
```
Chris-Hibbert, line 38: I think you're doing the same thing I do this way.  (You can see more instances of `//////` in this file.
```suggestion
// Support /////////////////////////////////
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -544,27 +544,62 @@ test('price drop', async t => {
...
+  const emptyPurse = await E(services.runKit.issuer).makeEmptyPurse();
+  /** @type {UserSeat<string>} */
+  const closeSeat = await E(zoe).offer(
+    E(vault).makeCloseInvitation(),
+    harden({
+      give: { RUN: AmountMath.makeEmpty(runBrand) },
+      want: { Collateral: AmountMath.makeEmpty(aethBrand) },
+    }),
+    harden({
+      RUN: await E(emptyPurse).withdraw(AmountMath.makeEmpty(runBrand)),
```
Chris-Hibbert: You shouldn't have to include an empty payment here.  I thought there would be a way to tell Zoe, but the contract is insisting that the payment has to be present. I'd consider that a bug in the contract. Vault's `closeHook` should allow you to close the vault without sending an empty payment.

packages/zoe/src/contractFacet/zcfSeat.js
```diff
@@ -50,6 +50,7 @@ export const createSeatManager = (
    * @returns {Allocation}
    */
   const getCurrentAllocation = zcfSeat => {
+    // TODO update docs that getCurrentAllocation() fails after exit
```
Chris-Hibbert, line 53: We prefer that this kind of TODO include the issue number (which means creating the issue before merging the PR.)

Chris-Hibbert (approved):
Some questions and cleanup suggestions, but I think they're straightforward at this point.

packages/run-protocol/src/vaultFactory/vault.js
turadg: > The query is about the outer vault

It's a little confusing but the param is just for the error message. The query is about the inner vault. 

> Why isn't 'active' a good term for this?

Because "active" is a particular state a vault can be in.  I have another PR that has another idea for the name so we can pick it up again in https://github.com/Agoric/agoric-sdk/pull/4743

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Basically, though `#region` is understood by VS Code to fold: https://code.visualstudio.com/docs/editor/codebasics#_folding

Looks like WebStorm can be configured to do this too: https://www.jetbrains.com/help/webstorm/working-with-source-code.html#use-the-surround-with-action

packages/run-protocol/src/vaultFactory/vault.js
turadg: Thanks for the code. I thought the burnSeat change is the same that I made but I must have had an error.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Great. That's the "better way" I was looking for. I've simplifed this to `const closeSeat = await E(zoe).offer(E(vault).makeCloseInvitation())` by only expecting a give/want proposal for when the vault is ACTIVE (ignoring the proposal shape when LIQUIDATED).

Issue #4568, opened 2022-02-16
Minimum acceptable Liquidation Contract


rowgraus:
## What is the Problem Being Solved?
Our existing liquidation contract sells vaults on the AMM one at a time until liquidations are complete.  However, there is no protection against large vaults (relative to AMM pool size) coming up for liquidation, such as the $65m vault that was liquidated on Maker several weeks ago.  A sale of a large vault relative to the collateral in the liquidity pool would cause a large price impact and lead to significant losses for the RUN Protocol. 

For reference, the price impact of a trade on an X*Y=K curve is 2x the size of the trade relative to the size of the pool.  For example, a $10m trade on a pool with $50m in liquidity would lead to a drop in price received for the trader of 40% (2 x 20%).

This ticket is to create a new liquidation contract which implements controls on the sale to minimize price impacts.

## Description of the Design
Specific elements include:
1. Creating and defining a "cohort" of liquidated vaults, which share in leftover change after liquidation of the full cohort is complete
2. Splitting individual sales on the AMM based on X% of the total AMM pool size (specific % a governance-controllable parameter. 0.05% reasonable)
3. Delay sale actions by a controllable parameter (gov. controllable parameter. 10 seconds reasonable)
4. Stop sale activity if AMM price diverges from Oracle price by > then X% (gov. controllable parameter. 5% reasonable)

Raw notes from previous discussion:
- Aggregation of liquidated vaults (liq contract has internal seat to sell from with loop until it is empty)
- Rate limiting of liquidation:
-- Limit size of sales to limit price impact to a threshold (0.1% of pool size - controllable by governance)
-- We expect Liquidator to pay gas fees out of execution fees it got from everyone else
-- Delay between sale actions by controllable parameter (10 seconds)
- Limit of oracle vs. AMM price - stop selling if price diverges by X% controllable by governance
- Cohort model to return assets
-- Message to liquidation engine that says “ok I’m done now”
-- Cohort model to be defined as we go

## Test Plan

* sell an amount smaller than the tranche size
* sell an amount exactly equal to the tranche size
* sell larger than tranche size
* oracle diverges from amm price and so pause liquidation

ref: #3533

Tartuffo:
@turadg  For Restival, create a new ticket to clean up liquidation selection in Vault Manager.

dckc:
The problem statement isn't clear to me. Why should we protect against "large vaults (relative to AMM pool size) coming up for liquidation"? What bad thing happens if we don't?

[...]

PR #4715, opened 2022-03-01
refactor(run-protocol): store liquidation cohort before liquidating

### LINKED ISSUE #4571: Liquidation cohort is stored separately

https://github.com/Agoric/agoric-sdk/issues/4568 is the motivation. This makes incremental progress on it so that we can test this part in Restival.

## Description of the Design

Have liquidation put vaults in a new (non-virtual) store before triggering liquidations. Durability will be tackled separately.

## Security Considerations
--
## Test Plan

--

turadg:
closes: #4571

## Description

Populates a `vaultsToLiquidate` queue before beginning liquidation. 

Separated in commits for easier review. I'll squash the whole PR for merge.

### Security Considerations

Doesn't change API.

Internal implementation.

### Testing Considerations

No change in behavior. Eventually we'll want these stores to persist durably which will need testing, but that's down the road.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -102,13 +102,18 @@ export const makeVaultManager = (
 
   let vaultCounter = 0;
 
-  // A store for vaultKits prioritized by their collaterization ratio.
-  //
-  // It should be set only once but it's a `let` because it can't be set until after the
-  // definition of reschedulePriceCheck, which refers to sortedVaultKits
+  /**
+   * A store for vaultKits prioritized by their collaterization ratio.
+   *
+   * It should be set only once but it's a `let` because it can't be set until after the
+   * definition of reschedulePriceCheck, which refers to sortedVaultKits
+   *
+   * @type {ReturnType<typeof makePrioritizedVaults>=}
+   */
   // XXX misleading mutability and confusing flow control; could be refactored with a listener
-  /** @type {ReturnType<typeof makePrioritizedVaults>=} */
   let prioritizedVaults;
+
+  // Progress towards durability https://github.com/Agoric/agoric-sdk/issues/4568#issuecomment-1042346271
   /** @type {MapStore<string, InnerVault>} */
   const vaultsToLiquidate = makeScalarBigMapStore('vaultsToLiquidate');
 
@@ -119,8 +124,10 @@ export const makeVaultManager = (
   /** @type {Ratio}} */
   let compoundedInterest = makeRatio(100n, runBrand); // starts at 1.0, no interest
 
-  // timestamp of most recent update to interest
-  /** @type {bigint} */
+  /**
+   * timestamp of most recent update to interest
+   * @type {bigint}
+   */
   let latestInterestUpdate = startTimeStamp;
 
   const { updater: assetUpdater, notifier: assetNotifer } = makeNotifierKit(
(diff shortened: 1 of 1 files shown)


packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -107,7 +108,10 @@ export const makeVaultManager = (
   // definition of reschedulePriceCheck, which refers to sortedVaultKits
   // XXX misleading mutability and confusing flow control; could be refactored with a listener
   /** @type {ReturnType<typeof makePrioritizedVaults>=} */
-  let prioritizedVaults;
+  let illiquidVaults;
```
Chris-Hibbert: I think these are the same `prioritizedVaults` as before. 'illiquid' sounds to me like they have insufficient capitalization and should be liquidated.  Would you mind renaming it back to prioritizedVaults, or something representing them as loans in good standing?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -129,30 +133,43 @@ export const makeVaultManager = (
   );
 
   /**
-   *
-   * @param {[key: string, vaultKit: InnerVault]} record
+   * @param {Iterable<[key: string, vaultKit: InnerVault]>} recordEntries
```
Chris-Hibbert: `recordEntries` is confusing to me. I alternate between reading `record` as a noun and a verb. Can they be `vaultEntries` or something like that?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -107,7 +108,10 @@ export const makeVaultManager = (
   // definition of reschedulePriceCheck, which refers to sortedVaultKits
   // XXX misleading mutability and confusing flow control; could be refactored with a listener
   /** @type {ReturnType<typeof makePrioritizedVaults>=} */
-  let prioritizedVaults;
+  let illiquidVaults;
+  /** @type {MapStore<string, InnerVault>} */
+  const vaultsToLiquidate = makeScalarBigMapStore('vaultsToLiquidate');
```
Chris-Hibbert, line 118: I think it's worth mentioning that this is here to support  https://github.com/Agoric/agoric-sdk/issues/4568 .  Otherwise, the separate store doesn't seem to serve any purpose.

Chris-Hibbert (commented):
This looks right, but I'd prefer to re-read it after the renamings I've suggested. The current names are confusing to me.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Yeah, it's a poor name choice. I was looking for something that described _how_ they're prioritized because `vaultsToLiquidate` are also ordered. But I'll leave that for https://github.com/Agoric/agoric-sdk/issues/4415

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -129,30 +140,43 @@ export const makeVaultManager = (
...
+            zcf,
+            vault,
+            runMint.burnLosses,
+            liquidationStrategy,
+            collateralBrand,
+          );
+
+          vaultsToLiquidate.delete(key);
+        } catch (e) {
+          // XXX should notify interested parties
```
Chris-Hibbert, line 172: If the vault is now in the `liquidating` state, should we try again to remove it from `vaultsToLiquidate`?

Chris-Hibbert (approved):
Looks good.  I had one question/suggestion.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Good question. My thought was it still needs to be liquidated. This way it'll keep retrying. I don't know of failure modes in which you wouldn't want to retry. I figure we'll come back to that when "notify interested parties" is addressed.

PR #4716, opened 2022-03-02
refactor(swingset): improve vat deliveries

warner:
This moves some logic out of `deliverAndLogToVal` and into a higher-level
dispatch function, where it can react to all failure modes in a single place.

We pull "run-queue events" off the run-queue (currently named the "acceptance
queue"). Many of these events are aimed at a specific vat, including 'notify',
but the primary one is 'send' and is aimed at a *target*, which is either a
kernel object or a kernel promise. For these we use `routeSendEvent` to
either queue the event on a promise queue, reject it if it can't be
delivered, or deliver it to a specific vat. This is the part that will change
with #3465: in that world, each vat-input-queue exists for a specific vat, so
the routing step moves earlier (into a "routing crank" that pulls events from
a vat-output-queue and possibly adds them to a vat-input-queue).

Some kinds of deliveries care about metering (or don't), and some have
opinions about what should happen if a crank gets unwound. These are now
reported as additional properties in the DeliveryStatus object that is
returned by this delivery path. `processDeliveryMessage` looks at these
properties, plus the delivery status (if any) to decide what to do at the end
of the crank.

I think most of the behavior should be the same as before. One change is that
the `runPolicy` will probably get more information about non-'send' cranks
than before (e.g. `create-vat` might report metering).

This refactoring should make it easier to implement #1848 vat-upgrade, as
well as #3465 queueing changes.


warner:
recommend reviewing with "hide whitespace", to reduce the diff

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -307,12 +307,42 @@ export default function makeKernelKeeper(
...
+    assert(
+      [
+        'local',
+        'nodeWorker',
+        'node-subprocess',
+        'xs-worker',
+        'xs-worker-no-gc',
+      ].includes(mt),
+    );
+    return undefined;
```
FUDCo: Won't a function return `undefined` anyway if you don't have an explicit `return` statement?  Or is this to deal with some kind of TypeScript nonsense?

warner: yeah, typescript

mhofman: That would be the difference between a return type of `void` vs `undefined`.

warner: let me know if there's a better way to write that assertion function, I just copied something else that works

FUDCo (approved):
This is a very nice blood pressure reducing cleanup!  I particularly like getting rid of the output named `policyInput` which, even though I kinda sorta understood the reasoning behind the name, has always bugged me.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -286,35 +274,27 @@ export default function buildKernel(
...
-    assert.typeof(compute, 'bigint');
-    const res = kernelKeeper.deductMeter(meterID, compute);
-
-    // We record the deductMeter() in postAbortActions.meterDeductions. If
-    // the delivery is rewound for any reason (syscall error, res.underflow),
-    // then deliverAndLogToVat will repeat the deductMeter (which will repeat
-    // the notifyMeterThreshold), so their side-effects will survive the
-    // abortCrank(). But we don't record it (again) during the repeat, to
-    // make sure exactly one copy of the changes will be committed.
+  let terminationTrigger;
```
mhofman, line 285: I assume it wasn't easy to refactor this to no longer be a global?

packages/SwingSet/src/kernel/state/kernelKeeper.js
mhofman: Assertions don't require a return value in TS. This might be a case of jsdoc eslint plugin not understanding assertions ?

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -307,12 +307,42 @@ export default function makeKernelKeeper(
...
+  }
+  /**
+   *
+   * @returns { ManagerType }
+   */
```
mhofman: FYI this should be unnecessary given that `mt` should be typed correctly after the assertion.

packages/SwingSet/src/types-internal.js
```diff
@@ -0,0 +1,15 @@
+import './types-exported';
```
mhofman: This is a real ES import, so it needs an extension.

```suggestion
import './types-exported.js';
```

But that bring the question why this is needed? Is it for bringing in the ambient types from `./types.js`? In which case why not go straight for the source of ambientness in the first place?

```suggestion
import './types.js';
```

packages/SwingSet/src/types-internal.js
```diff
@@ -0,0 +1,15 @@
...
+
+export {};
+
+/**
+ * @typedef { string } VatID
```
mhofman, line 6: Nit: avoid redefining if there is already a definition somewhere else (except if you mean to have a different but "compatible" definition for external use, such as a subset)

```suggestion
 * @typedef { import('./types-exported.js'}.VatID } VatID
```

packages/SwingSet/src/types-internal.js
```diff
@@ -0,0 +1,15 @@
...
+ * @typedef { string } MeterID
+ * @typedef { { meterID?: MeterID } } HasMeterID
+ * @typedef { import('./types-exported.js').DynamicVatOptionsWithoutMeter } DynamicVatOptionsWithoutMeter
+ *
+ * vatKeeper.setSourceAndOptions(source, RecordedVatOptions)
```
mhofman: This line seem out of place.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -323,101 +303,87 @@ export default function buildKernel(
...
-    let policyInput = ['none'];
     kernelKeeper.incStat('dispatches');
     kernelKeeper.incStat('dispatchDeliver');
     // eslint-disable-next-line no-use-before-define
-    if (!vatWarehouse.lookup(vatID)) {
-      if (msg.result) {
-        resolveToError(msg.result, VAT_TERMINATION_ERROR);
-      }
-    } else {
+    if (vatWarehouse.lookup(vatID)) {
```
mhofman: Nit: I wish we were consistent in which path we handle exceptional conditions. For example `processNotify` handles errors inside the `if` block, and the normal condition outside of it.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -786,108 +754,262 @@ export default function buildKernel(
...
+   * @typedef { { type: 'startVat', vatID: VatID } } RunQueueEventStartVat
+   * @typedef { { type: 'dropExports', vatID: VatID, krefs: string[] } } RunQueueEventDropExports
+   * @typedef { { type: 'retireExports', vatID: VatID, krefs: string[] } } RunQueueEventRetireExports
+   * @typedef { { type: 'retireImports', vatID: VatID, krefs: string[] } } RunQueueEventRetireImports
+   * @typedef { { type: 'bringOutYourDead', vatID: VatID } } RunQueueEventBringOutYourDead
+   * @typedef { RunQueueEventNotify | RunQueueEventSend | RunQueueEventCreateVat |
+   *            RunQueueEventUpgradeVat | RunQueueEventStartVat |
+   *            RunQueueEventDropExports | RunQueueEventRetireExports | RunQueueEventRetireImports |
+   *            RunQueueEventBringOutYourDead
+   *          } RunQueueEvent
```
mhofman, line 786: Thanks so much for typing this!

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -786,108 +754,262 @@ export default function buildKernel(
...
+      console.error(`We're currently already running at`, processQueueRunning);
+      assert.fail(X`Kernel reentrancy is forbidden`);
+    }
+    processQueueRunning = Error('here');
+    try {
+      const policyInput = await processDeliveryMessage(message);
+      return policyInput;
     } finally {
       processQueueRunning = undefined;
     }
```
mhofman: Maybe add a note that the await is necessary for the `try` / `finally`. Not always obvious :)
Alternatively this is where a `.finally` comes in handy

```suggestion
    return processDeliveryMessage(message).finally(() => { processQueueRunning = undefined; });
```

mhofman (changes requested):
This looks great! Thanks so much for starting this refactor.

I'm not familiar enough with the termination conditions and the start vat logic to be confident enough in my review for those parts, and I trust @FUDCo on his approval there.

The requested change is for the extension in the `import` in the types file, in case that file ever ends up actually imported by ES code someday. Rest is mostly nits on types and consistency.

packages/SwingSet/src/kernel/kernel.js
warner: It needs to be set by the two functions (`requestTermination` and `vatFatalSyscall`) which are handed to other code, so yeah I didn't see a way to avoid it.

Hm, it's worth considering whether we should move the flag into the vat-warehouse or the vat-manager. Basically after you tell vat-warehouse to make a delivery, you ask vat-warehouse whether the vat manager is now in a fatal-error state. Maybe the `KernelSyscallResult` that comes back to the vat-manager with a non `['ok', results]` status could trigger the manager to remember the problem, and then report it back to the delivery handler. That would help with our future plans for parallelism, for sure. I'll add a note.

packages/SwingSet/src/types-internal.js
warner: it was a reminder of where `RecordedVatOptions` gets used.. is there a better way to do comments within JSDoc comments? I could use `* // vatKeeper.set...` maybe.

packages/SwingSet/src/kernel/state/kernelKeeper.js
warner: I can remove the `@returns { ManagerType }`, thanks. (I started from `kernel.js` that calls this, added the `@returns` so the `options.managerType = kernelKeeper.getDefaultManagerType()` didn't complain, and then added the assertion to make the `@returns { ManagerType }` shut up.. should have started with the assertion).

If I remove the `return undefined;` from `insistManagerType()`, I get this warning:

```
$ eslint '**/*.js'

/home/warner/stuff/agoric/agoric-sdk/packages/SwingSet/src/kernel/state/kernelKeeper.js
  310:3  warning  JSDoc @returns declaration present but return expression not available in function  jsdoc/require-returns-check
```

I don't know how to make that go away, so I'm inclined to leave the `return undefined;`, but if you know a way, I'll apply it.

mhofman: Yeah sounds like a limitation of the JSDoc eslint plugin, not a problem with typescript. Besides an ugly eslint disable.

An alternative might be to return `mt` instead.

packages/SwingSet/src/types-internal.js
warner: Argh, I know I typed `./types-exported.js`, but something auto-corrected the extension away. I'll add it back.

It looks like `import ./types.js` works too.

With those changes, we've got:
* exported.js : import './src/vats/timer/types.js'
* src/index.js: export * from './types-exported.js'
* src/types-exported.js: import './types.js', export {}, some typedefs
* src/types.js: only typedefs, no JS imports or exports, TS import(types-exported) and (@endo/marshal) and ./vatKeeper, ./kernelKeeper, others
* src/types-internal.js: import './types.js', export {}, some typedefs
* kernel.js: TS import(types-internal)

If I understand it right, `src/types.js` lacks a JS `import` or `export`, so everything defined there is added to the ambient type set (including things like `@typedef { import('./types-exported.js').BundleID } BundleID`). Those are necessary for now, but we want to move away from that and make everything be imported/exported properly. So the goal is to move things out of `types.js` until it's empty, and we can rename everything to be shorter.

`src/index.js` re-exports everything from `types-exported.js`. This is where we want to go: other packages should reference types with `{ import('@agoric/swingset-vat').BundleID }`, and we should move externally-visible types from `types.js` to `types-exported.js`. This includes the API surface of `controller`, but also things like `vatAdminService` and the timer service. Today's addition is the `options` bag that `E(vatAdminService).createVat(bundlecap, options)` takes.

`types-internal.js` is where I'm moving things from `types.js` that are only for internal use. For example `kernel.js` now does:

```js
   * @typedef { import('../types-internal.js').InternalDynamicVatOptions } InternalDynamicVatOptions
```

I thought about defining that in `types.js` and having `kernel.js` import from there, but TS complains that `types.js` is "not a module" (which I assume is because it lacks a JS `export {}`). I also assume that adding `export {}` to fix that would disqualify `types.js` from contributing to the ambient type definitions, which would break the entire world (since it still depends on those ambient types).

If that all sounds correct, I'll stick with adding new things to `types-internal.js` (so I'm not accidentally adding them to the ambient set, which might hide things we add without also importing). But since `types-internal.js` does a `import './types.js'`, I should expect `@typedef { import('../types-internal.js').Foo } Foo` to be able to pick up anything from `types.js`.

In the long run, once we've drained `types.js`, I'll want to rename `types-exported.js` and `types-internal.js`. I don't know which one should win the short name, but hopefully by the time swingset gets there, somebody will have figured out a good pattern in some other package and I can just copy it.

packages/SwingSet/src/types-internal.js
warner: oh in fact I should just have `types-internal.js` import everything, with `import './types-exported.js'`.. the internal types are a superset of the exported ones

Also, `VatID` wasn't supposed to be visible to the outside world, at least not as part of the current public API. Vats don't learn their own IDs.

packages/SwingSet/src/kernel/kernel.js
warner: Yeah, in general I'm trying to avoid nesting: each `if` clause ends with a `return` so you don't have to worry about what else might happen later in the function. After that, I try to make the initial `if` be a positive case: `if (foo())` rather than `if (!foo())`.

`processSend` has a fairly simple thing to do in the positive case (send to the vat), whereas the negative case has an extra conditional (return `null`, but also reject the result promise iff it has one). If it weren't for the extra typescript/eslint-disable comments, that positive clause would be shorter than the negative clause:

```js
    if (vatWarehouse.lookup(vatID)) {
      return deliverAndLogToVat(vatID, kd, vd);
    }
```

vs

```js
    if (msg.result) {
      resolveToError(msg.result, VAT_TERMINATION_ERROR);
    }
    return null;
```

But for `processNotify`, there are more non-delivery cases to peel off first (dead vat, already-resolved so not in c-list, already-retired so not in promise-table), so doing the negative cases first reduces the nesting.

But.. I like the idea of having the delivery be the last thing in each handler, that seems like the best consistency to aim for. So I think I'll switch `processSend` around:

```js
  async function processSend(vatID, target, msg) {
    insistMessage(msg);
    kernelKeeper.incStat('dispatches');
    kernelKeeper.incStat('dispatchDeliver');
    if (!vatWarehouse.lookup(vatID)) {
      // splat
      if (msg.result) {
        resolveToError(msg.result, VAT_TERMINATION_ERROR);
      }
      return null;
    }

    /** @type { KernelDeliveryMessage } */
    return deliverAndLogToVat(vatID, kd, vd);
  }
```

packages/SwingSet/src/kernel/kernel.js
warner: As long as that doesn't run afoul of our "don't trust possibly-doctored promises", I like it. In this case the promise is coming from our own code (and no user-provided Promises ever make it into the kernel), so we know it's safe.

packages/SwingSet/src/kernel/state/kernelKeeper.js
warner: `return mt;` gets me `Type 'string' is not assignable to type 'void`. I don't know how to express "returns $type and also asserts that the input value is $type".

packages/SwingSet/src/kernel/kernel.js
mhofman: I still need to properly grok the interaction of deliveries with the vat-manager / vat-wharehouse logic.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -286,35 +274,35 @@ export default function buildKernel(
...
-  function deductMeter(meterID, compute, firstTime) {
-    assert.typeof(compute, 'bigint');
-    const res = kernelKeeper.deductMeter(meterID, compute);
+  // TODO: instead of using a kernel-wide flag here, consider making each
+  // VatManager responsible for remembering if/when a KernelSyscallResult
+  // reports a non-'ok' status and therefore the vat is toast. Then the
+  // delivery handler could ask the manager (or vat-warehouse) afterwards for
+  // the sticky-fatal state. If we did that, we wouldn't need
+  // `vatFatalSyscall`. We'd still need a way for `requestTermination` to
+  // work, though.
```
mhofman, line 283: Note to self: could this be handled a similar way, something in the vat translation layer that remembers the `exit` call?

packages/SwingSet/src/types-internal.js
```diff
@@ -0,0 +1,17 @@
+import './types.js';
+import './types-exported.js';
```
mhofman: I'm surprised we still need the `exported` one.

packages/SwingSet/src/types-internal.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @typedef { string } VatID
+ * @typedef { string } MeterID
+ * @typedef { { meterID?: MeterID } } HasMeterID
+ * @typedef { import('./types-exported.js').DynamicVatOptionsWithoutMeter } DynamicVatOptionsWithoutMeter
+ *
+ * // used by vatKeeper.setSourceAndOptions(source, RecordedVatOptions)
+ *
+ * @typedef { DynamicVatOptionsWithoutMeter & HasMeterID } InternalDynamicVatOptions
+ * @typedef { StaticVatOptions | { InternalDynamicVatOptions & HasMeterID } } RecordedVatOptions
```
mhofman: I'm confused why this doesn't require a `import('./types-exported.js').StaticVatOptions`

mhofman (approved):

packages/SwingSet/src/kernel/kernel.js
warner: Oh, you know, we should consider making the `syscall.exit` handler just push an `terminate-vat` event onto the vat's own input-queue, and perform the teardown later. Sort of "please shut me down, eventually, when it's convenient". Then vat-manager or the translation layer doesn't need to remember anything. And if our scheduler's policy always drains the vat-output-queue before looking at the vat-input-queue, then this would automatically deliver all the previously-emitted work before shutdown, which might line up nicely with our plans around involuntary termination needing to not interfere with previous output, and might simplify the "vat is dead but queues are not" question.

I don't know enough about our use cases for this to know whether promptness is important to us. Maybe it'd be when a contract has finished transferring all its assets and obligations to a successor, and it self-destructs to free up chain capacity, and to prevent surprises when people try to talk to the old one by mistake? Seems like the vat would know that it's in this state, so it could maybe reject all attempts to start new work while waiting to be terminated. OTOH, it would be really annoying to have to add extra "am I active" checks in vat code to make up for the platform not being able to respond to a `syscall.exit` promptly. We might work around this by configuring the input scheduler to service `terminate-vat` events ahead of anything else (but still behind draining the output queue).

Issue #2589, opened 2021-03-06
replace JSON-based marshal with dense binary encoding


warner:
## What is the Problem Being Solved?

(I thought we had a ticket for this already, but I couldn't find it.. feel free to close this as a duplicate if you find one)

Our `marshal` library currently uses a JSON-based encoding scheme to convert capability-bearing object graphs into pure data ("capdata", which is an object with a `body` string and `slots` array of object reference strings). It works by building a tree of objects from the original object graph, then using the built-in `JSON.stringify` to convert the tree into a single string. The `deserialize` function does the same thing in reverse.

The swingset comms vat has a related encoding scheme to convert messages to/from the "comms protocol", which adds additional control messages to the mix (deliver, notify, dropImport, and sequence numbers). It basically takes the two-part capdata object and encodes it into a single string.

The resulting output formats are pretty easy to specify, and nicely neutral (it would be pretty straightforward to port them to other languages than JavaScript), and convenient for humans to read the encoded output. But they're awfully fluffy. Common JS objects like `undefined` and `BigInt(1)` expand to 23- and 33- byte strings respectively.

#1661 touches on shrinking the representation of `undefined`, but points out that it would be a distraction: the real task is to replace the encoding entirely, with some dense binary format.

Given our heritage and community, [CapnProto](https://capnproto.org/) is a likely choice. #1827 explores the idea (in a different context). We might also consider [CBOR](https://cbor.io/) (RFC 8949), which I think is the encoding of preference in the IPFS world.

This touches on the question of how we want to represent binary data in our system: currently we can use the JS `ByteArray` in vat code, but they are mutable, and `marshal` can't handle them, so we're left in the awkward position of recommending base64-encoded strings, eww. #46 ("large string device") touches on this, as the data it manages might be arbitrary bytes.

## Compatibility Issues

Every vat within a single kernel should probably use the same encoding scheme, because the messages created by liveslots in one vat are sent into the kernel, (sometimes) stored in kernel promise queues, and delivered to other vats. The kernel should not be in the business of interpreting the `body` of capdata, so I don't think I'd want it responsible for conversion. User-level vat code does not generally see the output of `marshal`, because liveslots and/or the virtual objects layer abstracts that away, although I believe userspace *is* given access to `serialize` (but not `deserialize`) for some reason I don't currently understand.

[...]

PR #4717, opened 2022-03-02
docs(swingset): Improve "delivery" documentation

gibson042:
## Description

General improvements to documentation (fixing typos, improving internal consistency, clarifying concepts, etc.). It might make sense to review commit by commit and/or in rich-text mode.

### Security Considerations

n/a

See above.

### Testing Considerations

n/a


packages/SwingSet/docs/delivery.md
```diff
@@ -146,7 +149,7 @@ enum CapSlot {
     Object(ObjectID),
 }
 struct CapData {
-    body: Vec<u8>,
+    body: String,
```
gibson042, line 153: `Vec<u8>` appeared to be a typo, especially with other page content referring to body as a string and providing examples of such.

packages/SwingSet/docs/delivery.md
```diff
@@ -589,19 +596,20 @@ resolves, or both. To reduce the noise of unwanted notifications, Vats will
...
 
 ### syscall.resolve()
 
-`syscall.resolve` is used to resolve one or more Promises (usually just one,
+`syscall.resolve()` is used to resolve one or more Promises (usually just one,
 but occasionally a batch of mutually-referencing Promises must be resolved in
 a single syscall because their identifiers are aggressively retired
-immediately after translation).
+immediately after translation). \[TODO: resolve inconsistency between this claim
+and [Syscall/Dispatch API](#syscalldispatch-api)]
```
gibson042: The inconsistency is documentation of `trait Syscall` including `fn resolve(subject: PromiseID, to: Resolution)`, which seems unable to support resolving multiple promises in a batch.

packages/SwingSet/docs/delivery.md
```diff
@@ -809,15 +815,16 @@ The `subject` argument of `dispatch.notify()` specifies which Promise is
 being resolved. It is always translated into a `PromiseID`.
 
 `dispatch.notify()` will be sent to all Vats which had subscribed to hear
-about the resolution of that Promise. The Decider vat for a Promise will not
-generally subscribe themselves, since they are the ones causing the Promise
-to become resolved, so they have no need to hear about it from the kernel.
+about the resolution of that Promise [TODO: document ordering
+relevance/constraints]. The Decider vat for a Promise will not generally
```
gibson042: This is a question from me about whether the deliver order of notifications to multiple subscribing vats is deterministic, and if not whether that is guaranteed to be safe.

packages/SwingSet/docs/delivery.md
```diff
@@ -3,38 +3,42 @@ A walkthrough of how messages are passed from one Vat to another.
...
-local or remote objects with either the `E()` wrapper `p=E(x).foo(args)` or
-(eventually) the wavy dot syntax `p=x~.foo(args)`. Other forms of Vat Code
-could exist, such as non-orthogonal (database/ORM-based) persistence, or in a
-non-SES language (e.g. a WASM box).
+Each vat contains some application-specific code (named "Vat Code"). For
+SwingSet, most Vat Code uses orthogonal peristence (i.e., invisible to the vat
+code, which effectively perceives its memory data as eternal) and is written in
+the SES subset of Javascript, employing native platform Promises and making
+eventual-send calls to local or remote objects with either the `E()` wrapper
+(`resultPromise=E(x).foo(args)`) or (eventually) the wavy dot syntax
```
warner: let's drop mentions of wavy-dot, it's not coming back any time soon

packages/SwingSet/docs/delivery.md
```diff
@@ -3,38 +3,42 @@ A walkthrough of how messages are passed from one Vat to another.
...
+[Vat-Outbound Slot Translation](#vat-outbound-slot-translation)). To ensure that
+Vats are isolated from the kernel, and to build a deterministic transcript,
+these syscalls are limited to taking pure data as arguments (i.e. everything
+could be JSON serialized into a single string, without loss of correctness). The
+primary syscall is named `syscall.send()`.
+
+Each vat is represented in the kernel as a `dispatch` object with methods that
+close over its internal state (cf.
+[Vat-Inbound Slot Translation](#vat-inbound-slot-translation)). A vat's code
+executes when the kernel initiates a **turn** by invoking one of these methods,
```
warner: We're using "crank" for this, instead of "turn". A turn is one step of the engine's microtask queue, e.g. a single Promise's `.then` callback. A crank is 1 or more turns initiated by a `dispatch` invocation and finishing when the microtask queue is empty. Since vat code doesn't get access to timers or other forms of IO, they can only hold onto control until the end of the crank, at which point they cannot regain agency until the kernel sends them a new dispatch.

In fact, we're a bit casual about the difference between what vat code experiences and what the vat as a whole (including liveslots, and the worker process supervisor) experiences, and indeed what the kernel sees. Also our terminology has shifted over time.

If we're talking about the kernel, these days we use "crank" to talk about the processing of an event from the run-queue, which may or may not perform a delivery to a vat (e.g. the crank might merely queue a message on some unresolved kernel promise). If it does, we send a delivery message to the worker, which begins the "crank" that the vat knows about. Liveslots gets control first, deserializes the arguments, then invoke vat code, which can run for multiple turns until it loses agency. Then liveslots regains control, does some cleanup, then sends the results back to the waiting kernel.

For the purposes of this paragraph, I guess we should say the kernel initiates a "delivery". If "crank" comes up elsewhere, we could differentiate between a "kernel crank" (that may or may perform a delivery) and a "vat crank" which is the multi-turn execution triggered by a delivery).

packages/SwingSet/docs/delivery.md
warner: Hm, I believe my original attempt at a Rust implementation used `Vec<u8>` (the Rust name for bytes, as opposed to unicode), because I think that's what comes out of the `serde` JSON+UTF8 serializer and it's what can be sent over a wire. But `String` (i.e. the Rust name for unicode strings) is more meaningful to a JS audience, and is equally plausible, and is closer to what we actually do now in JS, depending upon where you look (there's one point where `syscall.send` includes a string `.body`, then there's a later point where the whole syscall gets serialized into bytes to send over the wire). So `String` sounds fine.

(#2589 is a long-neglected aspiration to move away from JSON and use something denser, maybe CBOR, which would also accomodate `bytes` arguments better, and if/when we figure that out, `.body` will probably need to be `bytes` instead of `String`)

packages/SwingSet/docs/delivery.md
```diff
@@ -194,18 +197,19 @@ in other tables). Each contains some additional state-specific data:
 
 * `Unresolved`: includes an optional Decider VatID, list of subscribers
   (VatIDs), and queue of pending messages
-* `Fulfilled`: includes an ObjectID (an Export) to which it was resolved
-* `Data`: includes resolution data (body+slots)
-* `Rejected`: includes the rejection data (body+slots, maybe an Error object)
+* `Fulfilled`: includes the ObjectID with which it was fulfilled
```
warner: We no longer distinguish between `fulfilled` (fulfilled to a callable object) and `data` (fulfilled to anything else). The current states (look for `resolveKernelPromise` in `src/kernel/state/kernelKeeper.js`) are `unresolved` or `fulfilled` or `rejected`, and the last two mean there's a `.data` property on the promise record that contains capdata, plus a `.rejected` boolean. We wait until we're trying to send a message to the resolved promise before we look at `.data` to distinguish between something that accepts messages and a data record that cannot.

packages/SwingSet/docs/delivery.md
```diff
@@ -375,19 +383,18 @@ prefer to avoid deserializing the messages until their resolution is known,
 to avoid a wasteful reserialization cycle.
 
 If the deciding Vat has *not* opted into pipelining, the messages are queued
-in the kernel's Promise Table entry instead. They remain there until the
+in the kernel's Promise table entry instead. They remain there until the
```
warner: I go back and forth about how to reinforce the idea that a "kernel promise" is not a javascript `Promise` object (but does correspond to `Promise`s in vats). Sometimes I use "kernel promise" vs "promise", sometimes I use "kernel promise" vs "`Promise`" (i.e. use the font and capitalization to evoke the JS syntax for the non-kernel thing). I've used "Promise Table" (both caps) to sort of glue the two words together and avoid making it sound like the kernel has a table of `Promise`s (it has a table of kernel promises, but no actual `Promise` objects).

Eh, words, they're hard. No recommendations here.

packages/SwingSet/docs/delivery.md
```diff
@@ -375,19 +383,18 @@ prefer to avoid deserializing the messages until their resolution is known,
...
-first message). Until the first message is delivered, the result Promise has
-no Decider, so the second message cannot be delivered. But that's ok, because
-by the time the second message gets to the front of the run queue, the first
-will have been delivered, setting the Decider of the result Promise to some
-vat, providing a place to deliver the second one (or the knowledge that the
+`syscall.resolve()` [below](#syscallresolve) for details.
+
+When a `syscall.send()` is submitted, the run-queue will have two pending
+deliveries: the first is targeting an object in some Vat, and the second targets
+the `result` PromiseID. Until the first message is delivered, the result Promise
```
warner: wordsmithing nit: removing the definite article ("when *the* initial pair" -> "when *a* send is..") makes me think this paragraph has stopped talking about pipelining, and removing "of the first message" loses an additional reminder, so it's not immediately clear to me that we're still talking about the same example. Maybe find a way to connect it more tightly. Maybe reference the method names above? "When the `syscall.send()` for `getRecord` is submitted .. the second `getBalance` targets the `result` PromiseID of the `getRecord` message" or something?

packages/SwingSet/docs/delivery.md
warner: oh yeah, when we switched to batch resolution, I didn't attempt to update the pseudo-rust writeup

The new signature is more like:
```rust
struct OneResolution {
  subject: PromiseID,
  rejected: bool,
  resolution: CapData,
}

fn resolve(resolutions: Vec<OneResolution>) {}
```

and `dispatch.notify` takes a batch of resolutions too

packages/SwingSet/docs/delivery.md
```diff
@@ -615,22 +623,20 @@ whether the Promise was allocated by this Vat or a different one.
...
+* `Data(CapData)`: the Promise is "fulfilled" to data, rather than to a callable
   object. It is an error to send messages to data.
 * `Reject(CapData)`: the Promise is "rejected" to data which we call the
   "error object". Sending a message to a Rejected Promise causes the result
   of that message to be Rejected too, known as "rejection contagion".
-
-. Any `result`
-  promises in the queued messages should be rejected with the same `CapData`
-  provided as `resolution`.
+* `Forward(PromiseID)`: the Promise is now "forwarded": it has not settled to
```
warner: We haven't implemented forwarding yet, so we should probably take it out. If/when we do, it will be included in `syscall.resolve` (you just resolve the promise to capdata that names a single promise, just like `Fulfill(ObjectID)` is now expressed by `syscall.resolve`ing the promise to capdata that names a single object).

Maybe mention that we'll implement forwarding in the future, but continue to describe its properties and limitations (no cycles).

packages/SwingSet/docs/delivery.md
```diff
@@ -615,22 +623,20 @@ whether the Promise was allocated by this Vat or a different one.
 The `resolution` has several forms, and we assign a different name to each.
 
 * `Fulfill(ObjectID)`: the Promise is "fulfilled" to a callable Object
-* `Forward(PromiseID)`: the Promise is now "forwarded": it has not settled to
-  a specific object, but the original Promise is effectively replaced with
-  some other Promise
-* `Data(CapData)`: the Promise is "fulfilled" to data, rather than a callable
+* `Data(CapData)`: the Promise is "fulfilled" to data, rather than to a callable
```
warner: same as above, there's now just a single API for resolution, but the capdata you use determines what happens later when someone tries to send a message at that promise

packages/SwingSet/docs/delivery.md
```diff
@@ -668,7 +674,7 @@ includes it in an argument, the lower-level translation layer can create a
 new promptly-resolved Promise for it.
 
 (TODO: `resolve()` is a good opportunity to remove the promise from the
-resolving vat's c-list, however if we have queued messages, it must live long
+resolving vat's C-List, however if we have queued messages, it must live long
```
warner: FYI we now do actually remove the promise from the c-list upon resolution, so we can probably delete this paragraph.

packages/SwingSet/docs/delivery.md
```diff
@@ -679,9 +685,9 @@ Note: we no longer have distinct syscalls or states for the different flavors
 of resolved promises. Instead, each resolved promise is recorded with a
 boolean `isRejected` flag, and a `CapData` to hold the resolution data. The
 `Fulfill` and `Data` flavors both set `isRejected = false`, and only differ
-by the contents of the resolution data. If `!isRejected` and the data holds a
-single Object, then messages can be sent to the promise, and they will be
-passed along to the Object. Otherwise messages sent to the promise will
+by the contents of the resolution data. If `isRejected` is false and the data
```
warner: oh hey I did update the document slightly

packages/SwingSet/docs/delivery.md
warner: We don't make any particular guarantees about that. Currently we probably accidentally deliver notifies in the order of subscription, but 1: vats would be hard pressed to tell, because we don't make any guarantees about the order of vat execution, also they only talk through queues, and those are the same queues that the subscribe/notify messages travel, so unless they both have access to a device that reveals ordering differences they have no reliable way to perform a test. And 2: the ordering is about to change a bunch with #3465 .

packages/SwingSet/docs/delivery.md
```diff
@@ -826,16 +833,17 @@ A Vat might be informed that one Promise has been resolved to another Promise
 ## Sample Message Delivery
 
 This section follows a variety of messages are they are sent from Vat-1 to
-Vat-2.
+Vat-2. Syntax like `foo!bar(baz)` is used to indicate a message to Object or
```
warner: Ah, bang notation, even older than tildot/wavydot. This might be worth replacing with `E()`, except I really like e.g. `x~.foo()~.bar()` over `E(E(x).foo()).bar()`

packages/SwingSet/docs/delivery.md
```diff
@@ -966,19 +974,19 @@ receiving pipelined messages.
 
 The two `send` calls will look like:
 
-* `syscall.send(target=o-1001, msg={method: "foo", .. result=p+104})`
-* `syscall.send(target=p+104, msg={method: "bar", .. result=p+105})`
+* `syscall.send(target=o-1001, msg={method: "foo", …, result=p+104})`
```
warner, line 975: You're determined to drag me kicking and screaming into the modern unicode world, aren't you? :)

I don't have a button on my keyboard for that symbol. I'd kinda prefer to keep this document plain ASCII, but if you think there's an ambiguity caused by U+002E,U+002E (I know `..` looks like `...` which looks like a real JS operator, and `..` is an operator in other languages), then stick with U+2026 and I'll cut-and-paste this symbol if I ever need to edit this doc.

packages/SwingSet/docs/delivery.md
```diff
@@ -1289,20 +1297,20 @@ The kernel routes messages between multiple vats. The Comms Vat routes
...
-excent, the kernel is treated like just another remote machine: there is also
+extent, the kernel is treated like just another remote machine: there is also
 a kernel-facing C-List (however kernel "messages", really syscalls, are
 delivered immediately, whereas messages to remote machines are asynchronous).
 The Comms Vat does not need to manage a run-queue (`dispatch()` causes an
 immediate external message), nor does each unresolved promise have a queue of
 messages (these are pipelined immediately).
 
-The one wrinkle is that Vat-Vat connections are symmetric, which impacts the
+The one wrinkle is that Vat→Vat connections are symmetric, which impacts the
```
warner, line 1305: ditto

packages/SwingSet/docs/delivery.md
```diff
@@ -1311,20 +1319,20 @@ externally-facing side of their per-machine C-Lists. As a result, the IDs
...
+most important thing in their self-centered world, so vat exports get the
+positive number (`o+1`). Comms vats, being more worldly, are obsequiously
+polite, so they always deliver a remote message in the form that will most
+please the recipient, so when a machine's export is sent back to them, the
 exporter+recipient will receive a positive number (`ro+2`), even though the
-exporter would *send* that object as `ro-2` to please the importer).
+exporter would *send* that object as `ro-2` to please the importer.
 
-The message names are also different. In the local-machine Vat-Kernel-Vat
+The message names are also different. In the local-machine Vat→Kernel→Vat
```
warner, line 1328: ok here's a case where the unicode arrows are a tiny bit nicer than `Vat->Kernel->Vat`, I'l begrudgingly admit :)

warner (approved):
Good changes, minor update suggested. Land after considering them (of course squash before landing).

packages/SwingSet/docs/delivery.md
gibson042: Then I think it makes sense to use `String` here for now and adjust upon adopting a binary encoding, unless that will actually cause problems for anyone consuming this.

packages/SwingSet/docs/delivery.md
gibson042: Thanks. It also looks like the "forwarded" is unimplemented, and possibly stubbed out in a few files as "redirected". Should I leave the documentation for it?

packages/SwingSet/docs/delivery.md
gibson042: Is that a concern w.r.t. to nondeterminism and subsequent loss of consensus, or is every instance guaranteed to deliver in the same (unpredictable) order?

packages/SwingSet/docs/delivery.md
gibson042: I figure this note allows us to keep it here as a not-actually-executable shorthand.

packages/SwingSet/docs/delivery.md
gibson042: That was indeed the concern. Hopefully changes won't be needed too often. :P (or more properly 😜, although that one *I* had to copy).

P.S. I don't know this part of Mac yet, but I think every Linux input method has a default `<Multi_key> <period> <period>` .XCompose sequence for "…" (which is how I type it).

packages/SwingSet/docs/delivery.md
gibson042: Excellent points. I moved the vat code to immediately precede this in a new "Example" subsection.

packages/SwingSet/docs/delivery.md
gibson042: That was indeed the concern. Hopefully changes won't be needed too often. :P (or more properly 😜, although that one *I* had to copy from [Emojipedia](https://emojipedia.org/winking-face-with-tongue/)).

P.S. I don't know this part of Mac yet, but I think every Linux input method has a default `<Multi_key> <period> <period>` .XCompose sequence for "…" (which is how I type it, where my Multi_key is configured to be the right-side Ctrl).

gibson042:
@warner The followup changes were sufficiently substantial that I'd like another round of review before merging.

packages/SwingSet/docs/delivery.md
warner: All consensus instances get the same order: they're all using the same kernel, and it's that kernel that decides how to service the queues. The term we use / property we provide is "arrival order non-determinism": until the thing arrives, you don't know the ordering, but once happens, the ordering is fixed and nobody will see any other order.

At a different level, vats shouldn't be able to sense much about other vats, and if the kernel has the flexibility to do things in various orders, its choices might be influenced by its knowledge of all vats, and thus the kernel might reveal a side-channel that we'd rather keep closed. Some of this is inevitable, since vats talk to each other. And it's probably the case that if vats A and B only talk to each other, and vats C and D only talk to each other, then neither pair's event ordering will be influenced by the other pair. So I'm not worried about our current ordering.

At an even different level, within a single vat, when it sends a bunch of messages out and is holding promises for their resolution, we're also trying to avoid making claims about the order in which they resolve on that particular vat. But those resolutions arrive as `dispatch.notify` events, which go into the vat's transcript, and all consensus instances run their vats with the same transcripts. So at the point of entry to the vat (`dispatch.*`), we have arrival-order non-determinism: the order is deterministic across all consensus instances, but the vat has no idea what that order is (and shouldn't depend upon any particular one), but the vat will never observe two different orders (in fact, orthogonal persistence means the vat will never even know that it was restarted and the transcript re-executed, as it has no access to anything which could prove otherwise).

packages/SwingSet/docs/delivery.md
warner: Yeah, good point, and I do that all the time in unit test comments. Let's update it to tildot then, so at least we only have one archaic dialect instead of two.

packages/SwingSet/docs/delivery.md
```diff
@@ -494,28 +508,28 @@ There are a few restrictions on the API:
...
+  which the local Vat is the Decider. In both cases, the Vat could have
+  delivered the message directly, instead of taking the time and effort of going
+  through the kernel's run-queue. On the other hand, this may achieve certain
   ordering properties better.
 
 In some places, `dispatch.deliver()` is named `message`: we're still in the
 process of refactoring and unifying the codebase.
 
-## Outbound (Vat->Kernel) translation
+## Vat-Outbound Slot Translation
```
warner: I'd recommend against this change. Each new person I've explained this to has gotten confused about the directions of "outbound" vs "inbound" for the first few days.. we have a coherent metaphor ("vats are egocentric"), but it isn't obvious, and I'd like to reinforce the scheme whereever we can, especially in a document like this one. So for section titles, having both the word "outbound" and the "vat->kernel" direction expressed at the same time feels important to me.

Feel free to rephrase it, and "Vat-Outbound" does make more sense than just "Outbound", but I think it's important to indicate the direction in the same place as introducing "outbound". Maybe "Vat Outbound Slot Translation (from Vat into Kernel)"? "Vat Outbound Slot Translation (Vat->Kernel)"?

Also maybe I'm being paranoid but I've seen enough `<unrenderable glyph>` boxes in documents I read, that I don't trust that all readers will render a unicode arrow, and if that happens then we're losing critical information. If you do want to use a unicode arrow in the title, maybe add a few words to be beginning of the section, "As messages are sent "outbound" from the vat into the kernel, ..." for the hapless folks reading the .md in a terminal and cursing about newfangled character sets when the old one worked just fine.

packages/SwingSet/docs/delivery.md
```diff
@@ -768,12 +769,12 @@ explicitly rejected, the original message is discarded, as there is nobody to
 accept it. However if the message requested a `result`, that result Promise
 must be rejected, just as if decider Vat had called
 `syscall.resolve(Reject(error))`. The rejection error is either a
-`CannotSendToData` object (for `Data`), or a copy of the Promise's own
-rejection object (for `Rejected`).
+`CannotSendToData` object (for `Fulfill` to non-Object data) or a copy of the
+Promise's own rejection object (for `Rejected`).
 
-## Inbound (Kernel->Vat) translation
+## Vat-Inbound Slot Translation
```
warner: ditto

packages/SwingSet/docs/delivery.md
```diff
@@ -809,33 +810,35 @@ The `subject` argument of `dispatch.notify()` specifies which Promise is
...
+subscribe themselves, since they are the ones causing the Promise to become
+resolved, so they have no need to hear about it from the kernel.
 
-The `Resolution` value of `dispatch.notify()` may contain slots, which are
+The `Resolution` values of `dispatch.notify()` may contain slots, which are
 translated just like `Message.args`.
 
 TODO: After a Vat receives notice of a Promise being resolved, we might
-choose to remove that promise from the vat's c-list, and forbid the Vat from
+choose to remove that promise from the vat's C-List, and forbid the Vat from
```
warner: Let's change this paragraph: we do indeed remove the promise from the c-list ("c-list promise entries are retired upon resolution")

We also use the word "retired" to describe removing c-list *object* entries when one side no longer needs to talk about the object, which is somewhat after the "drop" that one side does when it can no longer *reach* that object, because of WeakMaps. So let's be sure to use "retire" for promises here, and not "drop".

We don't exactly forbid the vat from mentioning the `vpid` again (we don't keep a history of retired IDs), but if it did, the kernel would treat it as a brand new promise. Liveslots won't, because it uses a persistent incrementing counter to generate IDs, so it will never re-generate the old one.. oh, huh, ok that's a problem for vat-upgrade, I'll make a ticket #4730 about that.

warner (approved):
Good, a few small recommendations remaining.

packages/SwingSet/docs/delivery.md
gibson042: I agree with keeping special characters out of section headings, even including ASCII punctuation (which is actually what motivated this change—I want predictable and stable auto-generated anchor ids). But I just pushed a fix for that by adding anchor elements directly.

PR #4718, opened 2022-03-02
chore: Sync Endo packages 2022-03

kriskowal:
This change brings in the latest from Endo, including a fix for line number generation and relaxation on dependence on pre-initialization of `@agoric/babel-standalone` https://github.com/endojs/endo/pull/1096


kriskowal:
This includes a commit that merges the semver compatible versions of babel/generator. I don’t know why yarn doesn’t do this on its own. Seems like a bug.

michaelfig (approved):

Issue #4719, opened 2022-03-02
fully validate bundles during `controller.validateAndInstallBundle`


warner:
## What is the Problem Being Solved?

Endo is nearly ready (https://github.com/Agoric/agoric-sdk/pull/4718) to provide full support for bundle hashes and their validation.

Once that lands, the task is to update swingset to use these hashes and validation functions, rather than the ad-hoc code I wrote earlier.

## Description of the Design

Remove `src/lib-nodejs/validate-archive.js` in favor of Endo-provided functions. Rewrite `validateAndInstallBundle()` to use them.

## Security Considerations

The new validation will be far more complete: the ad-hoc tool I wrote only hashed the compartment map, but of course the entire archive must be validated.

## Test Plan

Small update to existing unit tests.

PR #4720, opened 2022-03-02
test(swingset): temporarily tolerate new endo bundle hashes

warner:
Endo is adding an `.endoZipBase64Sha512` property to the output of
`bundleSource`. Swingset's tests aren't quite ready for this. Temporarily
delete this property until we're ready to use it properly.

refs #4719


kriskowal (approved):
Consider hardening the attenuated bundle. The checker will require that.

Issue #4637, opened 2022-02-22
send `dispatch.startVat` to setup()-based vats too, remove comms DB initialized check


warner:
## What is the Problem Being Solved?

https://github.com/Agoric/agoric-sdk/pull/4575#discussion_r809758731 suggests sending `dispatch.startVat()` to `setup()`-based ("raw") vats too, not just liveslots-based ones. At the time, I couldn't think of a use case for it, but looking at some old tickets, I realized there's a performance improvement to be had.

Commit 683b77108dbb6afaf6b2557dfacae2298016eaca (to fix #3726) makes the comms vat do a DB read on every delivery, because it must do some one-time initialization, and we can't use an in-RAM flag to remember whether it's been done or not (because comms does not use a transcript). We could save that DB read if the kernel could promise that it would inform the comms vat about initialization, exactly once (for the lifetime of the durable state).

## Description of the Design

* change `kernel.js` `processCreateVat()` to call `processStartVat()` (or whatever refactoring #4636 suggests) unconditionally, instead of only when `enableSetup` is false
* update all unit tests that use raw vats to provide a dummy `dispatch.startVat` (except for the new unit test that asserts that it is called, exactly once, even when the kernel is restarted)
* change comms to perform it's one-time initialization in response to `dispatch.startVat`, instead of checking the DB flag on every delivery

## Security Considerations

none

## Test Plan

unit tests

Tartuffo:
Nice to have, not critical (would remove a syscall from every comms message).

PR #4725, opened 2022-03-03
fix(swingset): deliver startVat to setup() vats

warner:
For normal vats (both static and dynamic), `dispatch.startVat()` is the
signal to evaluate the userspace code bundle and invoke `buildRootObject`.
Liveslots has access to `syscall` earlier, but it must not use it until
`startVat()` indicates the kernel is ready (prepared to record the syscalls
in a transcript). However, `setup()`-based "raw vats" did not receive
`startVat`.

This changes the kernel to send `startVat` to `setup()`-based vats too. The
only real `setup()`-based vat we use is comms, which now ignores `startVat`,
but in the future will be able to use it to avoid constant vatstore
initialization queries.

Several unit tests create `setup()`-based vats: these were changed to ignore
`startVat` too. The test which counted message deliveries were updated to
accomodate the extra deliveries.

refs #4637
but doesn't close it until comms is improved too

### DIFF
--- a/packages/SwingSet/src/controller/initializeKernel.js
+++ b/packages/SwingSet/src/controller/initializeKernel.js
@@ -88,9 +88,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
       const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
       vatKeeper.setSourceAndOptions({ bundleID }, creationOptions);
       vatKeeper.initializeReapCountdown(creationOptions.reapInterval);
-      if (!creationOptions.enableSetup) {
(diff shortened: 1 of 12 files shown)


FUDCo (approved):
Det är bra.

warner:
@FUDCo I added a new commit to update the comms side to take advantage of this, please re-review. I also found another assertion that we could strip. Together, this should save us two DB reads per comms message.

warner:
Oops, now that comms is initializing during startup (even if nobody ever talks to it), the virtual-objects "golden master" -style test is failing because of the extra vatStore writes. Lemme fix that.

FUDCo (approved):
Es iz gut.

Issue #4560, opened 2022-02-16
export external types as modules


kriskowal:
At time of writing, we need to use `--maxNodeModulesJsDepth` in `tsc` commands for `lint:types` scripts in most Agoric SDK packages because most of our packages do not provide type definitions and do use ambient types expressed in JSDoc comments. 

The worst outcome of this is that tools like IDE LSP plugins for tsc can’t see the necessary command line flag, and that flag cannot be expressed in project files like `jsconfig.json`. Consequently, type checking may pass in CI but appear to be failing in IDE’s. Lesser bad outcomes include slow type checking performance. One small upside is that we don’t have to use a build step to keep exported type definitions in sync with the JavaScript code.

To eliminate this problem we will need to both migrate away from ambient types and add a build step to generate type definitions to most projects. Packages that do not have internally coherent types will need manually written type definitions.

Migrating away from ambient types involves adding `export {}` to any `types.js` or `exported.js` files that only contain JSDoc comments, and then replacing `import 'types.js'` invocations with use of `/** @type {import('./types.js').T} */` style references for individual imported types.

It’s also necessary to replace the convention of `exported.js` with an entry in the main module of each package like:

```js
// eslint-disable-next-line import/export
export * from './src/exported.js';
```

Such that `/* @type {import('dep').T */` can be used to refer to all the types exported by package `dep`.

Migrating away from ambient types is a **breaking change** and requires conventional commit ceremony:

```
fix(package)!: Migrate away from ambient types

*BREAKING CHANGE*: Package no longer carries ambient types. All types must be expressly imported in type definitions like `/** @type {import('package').Type} */` and `import 'package/src/export.js'` will no longer work as expected.
```

That covers ambient types. Then, to remove `--maxNodeModulesJsDepth`, you will need to arrange a different `tsc --build` command with `"noEmit": false` in a `jsconfig.build.json` and TODO👉 the necessary incantations to put type definition files in the right places 👈TODO and add that build step to the `"build"` script in `package.json` TODO according to conventions yet to be established TODO.

mhofman:
For reference, the `@endo/marshal` package just went through this ceremony: https://github.com/endojs/endo/pull/1025, and the corresponding changes to the `agoric-sdk` don't seem to be too painful: https://github.com/Agoric/agoric-sdk/pull/4413

Chris-Hibbert:
[...]

PR #4729, opened 2022-03-03
refactor(SwingSet): Define SwingSet types as module exported

mhofman:
Add "re-exporting" ambient file for backwards compatibility

refs: #4560 

## Description

- Moves previously ambient type (`types.js`) definitions as being declared in a module file (`types-external.js`), requiring them to be explicitly imported for usage, and combine into it the previous `types-exported.js` (fixing a typo for the type of `DynamicVatOptionsWithoutMeter.vatParameters`)
  - Even though these ambient types were not previously included in the `@agoric/swingset-vat/exported.js` file, by being ambient they effectively were made transitively external when someone imported from `@agoric/swingset-vat`.
- Add a `types-ambient.js` file which "typedef re-exports" types from `types-external.js`
  - This is for backwards compatibility of previous users of ambient types.
  - `type.js` -> `types-ambient.js` with the following modifications
    - Remove all `@property` lines
    - `s/@typedef +\{(?:[^@]|\n)*\} +(\w+)( +[0-9a-zA-Z ]+)?$/@typedef { import('./types-external.js').$1 } $1/`
    - Double check which line were not substituted (had an `@` in the type definition)

### Security Considerations

None. Typing only change

The type definitions in `types-external.js` should probably move to being defined closer to the internal API that use the types, and `@typedef`ed in `types-external.js` so they can be used by external users through `import('@agoric/swingset-vat').ExportedType`

The `types-ambient.js` should over time shrink into oblivion, with ambient usages replaced by the corresponding `import().Type`

### Testing Considerations

`yarn lint` passing successfully


warner:
Not sure what that test failure was, looks like a temporary network problem, I hit the "re-run all jobs" button

warner (approved):
sounds at least as plausible as anything else I've seen :)

I'll land into my PR branch once @turadg is ok with it.

mhofman:
> Not sure what that test failure was, looks like a temporary network problem, I hit the "re-run all jobs" button

Yeah not sure either. @michaelfig any idea?

PR #4732, opened 2022-03-03
refactor(run-protocol): extract `chargeInterest` for use outside vaults

turadg:
## Description

getRUN will be charging interest like vaults do, so this pulls a `chargeInterest` function out. It has the added benefit of generalizing interest charging to any kind of NatValue brand.

Moves interest utils up to root of run-protocol since they're for more than just vaults now. Also does some validation that the brands match up.

Will merge as squash.

### Security Considerations

No change in powers.

--

### Testing Considerations

CI sufficient.


packages/ERTP/src/issuerKit.js
```diff
@@ -11,10 +11,38 @@ import { makePaymentLedger } from './paymentLedger.js';
...
+ * The allegedName becomes part of the brand in asset descriptions. The
+ * allegedName doesn't have to be a string, but it will only be used for
+ * its value. The allegedName is useful for debugging and double-checking
+ * assumptions, but should not be trusted.
+ *
+ * The assetKind will be used to import a specific mathHelpers
+ * from the mathHelpers library. For example, natMathHelpers, the
+ * default, is used for basic fungible tokens.
+ *
+ *  `displayInfo` gives information to UI on how to display the amount.
```
Chris-Hibbert:
```suggestion
 *  `displayInfo` gives information to the UI on how to display the amount.
```

I realize you only moved these comments, but since I spotted typos, now seems the right time to fix them.

packages/ERTP/src/issuerKit.js
```diff
@@ -11,10 +11,38 @@ import { makePaymentLedger } from './paymentLedger.js';
...
+ *
+ * @param {string} allegedName
+ * @param {K} [assetKind=AssetKind.NAT]
+ * @param {AdditionalDisplayInfo} [displayInfo={}]
+ * @param {ShutdownWithFailure=} optShutdownWithFailure If this issuer fails
+ * in the middle of an atomic action (which btw should never happen), it
+ * potentially leaves its ledger in a corrupted state. If this function was
+ * provided, then it the failed atomic action will call it, so that some
```
Chris-Hibbert:
```suggestion
 * provided, then the failed atomic action will call it, so that some
```

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
```diff
@@ -10,8 +10,8 @@ import { toVaultKey } from './storeUtils.js';
 
 /**
  *
- * @param {Amount<NatValue>} debtAmount
- * @param {Amount<NatValue>} collateralAmount
+ * @param {Amount<'nat'>} debtAmount
+ * @param {Amount<'nat'>} collateralAmount
```
Chris-Hibbert: Is there any chance we can make a constant or symbol to use here rather than a bare string?  I realize this is a type declaration rather than javascript, so the answer may be "no", which would be unfortunate. If it only had to appear in our code that would be one thing, but if authors of contracts have to do this, I'm concerned about the opportunity for typos.

packages/zoe/src/contractSupport/interest.js
```diff
@@ -27,10 +27,10 @@ const calculateRelativeCompounding = (
 
 /**
  *
- * @param {Amount<NatValue>} debtSnapshot
+ * @param {Amount<'nat'>} debtSnapshot
```
Chris-Hibbert: All the methods in this file are only used within Vaults. Let's move it there for a start.

The follow-on question is whether it should be merged in with the new `interest.js`.  I guess these only apply to the way that `Vaults` aggregate interest charges at the `VaultManager` level, so that would argue that it makes sense to leave them separate.

Would you mind moving it and adding comments about how special-purpose this is?

Chris-Hibbert (commented):
I like the refactoring and relocation of interest calculation, though there's an associated piece that should be moved out of Zoe, I think.

On the changes to type declaration of Amounts, I'll have to defer to @erights 

You have my LGTM once MarkM approves the ERTP type changes.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
dckc: Yes, `tsc` should grok a `const NAT = 'nat';` thing, I'm told (and I've seen).

Whether it's smart enough for `const { NAT } = AssetKind;`, I don't know.

turadg: > concerned about the opportunity for typos

The type parameter K must be a subtype of `AssetKind` which is the union of `'nat' | 'set' | 'copySet' | 'copyBag'`, so if someone were to type any string other than one of those four it would be a type failure. ([type](https://github.com/Agoric/agoric-sdk/blob/7e1e93241a5cd88a5931240c66bddbdab86b939e/packages/ERTP/src/types.js#L55))

There may be an aesthetic argument to use `AssetKind.NAT` ([const](https://github.com/Agoric/agoric-sdk/blob/7e1e93241a5cd88a5931240c66bddbdab86b939e/packages/ERTP/src/amountMath.js#L14-L25)) but it would be more verbose without any safety benefit: `Amount<AssetKind.NAT>`.

Dan's `Amount<NAT>` idea would b no more verbose and a little cleaner but then you'd have to define or import NAT in the file you're typing that. (That import works, btw, because AssetKind object's type is a not `[index]: string` but literally each key to each value.)

Given how easy this would be to revert or change to either of the above, can we let it through and adjust as needed?

packages/zoe/src/contractSupport/interest.js
turadg: >  I guess these only apply to the way that Vaults aggregate interest charges at the VaultManager level, so that would argue that it makes sense to leave them separate.

The refactoring was prompted by working with @dckc on getRUN . It's not a vault and needs to charge interest and is in progress https://github.com/Agoric/agoric-sdk/pull/4731 

So I'll move it from Zoe's `interest.js` to Run-protocol's.

turadg:
> On the changes to type declaration of Amounts, I'll have to defer to @erights. You have my LGTM once MarkM approves the ERTP type changes.

Since that change is tangential to supporting getRUN I've pulled it off to https://github.com/Agoric/agoric-sdk/pull/4736

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
turadg: Second thought, this PR is unblocking and the types are just nice to have so I extracted to https://github.com/Agoric/agoric-sdk/pull/4736

dckc: > Given how easy this would be to revert or change to either of the above, can we let it through and adjust as needed?

I don't feel comfortable speaking for @erights on that.

turadg:
> call it keyword rather than name.

This makes sense, will do.

> I think we never use alleged names for anything other than debugging / display. I suggest making this an argument to chargeInterest.

Oh I wish this before I ran into the async lookup problem that required [27116fc](https://github.com/Agoric/agoric-sdk/pull/4732/commits/27116fc2c8ad3cebda909fa9a042cf0328941f21). Is there no way to find out from a mint what keyword it expects in its `mintGains` method? I'll look in `getIssuerRecord`

dckc:
> ... Is there no way to find out from a mint what keyword it expects in its `mintGains` method?

There's a misunderstand somewhere around here. Mints don't expect keywords. Zoe and contracts do. Keywords are part of the Zoe API, not the ERTP API.

turadg:
> Mints don't expect keywords

Maybe we're talking about different keywords? I'm referring to the `AmountKeywordRecord` type that `mintGains` of the mint expects. 
Screen Shot 2022-03-04 at 11 04 44 AM

Chris-Hibbert:
> the `AmountKeywordRecord` type that `mintGains` of the `mint` expects.

The mint/issuer doesn't constrain the type used there. That's a convention that the contract uses to coordinate between the seats. If the parts of your contract agree, you could call it `Gains:` or `Foo:`, and everyone would be happy.

dckc:
@Chris-Hibbert writes:
> You have my LGTM once MarkM approves the ERTP type changes.

and @turadg took the commit with those changes out. So I suppose we're ready to go here.

dckc (approved):

PR #4734, opened 2022-03-04
chore: Synchronize encouragementBot code between SwingSet and swingset-runner

gibson042:
## Description

packages/SwingSet/demo/encouragementBot and packages/swingset-runner/demo/encouragementBot are near-duplicates of each other, as are packages/SwingSet/demo/encouragementBotComms and packages/swingset-runner/demo/encouragementBotComms. This brings them into as much alignment as seems possible without touching [packages/swingset-runner/demo/encouragementBotComms/swingset.json](https://github.com/Agoric/agoric-sdk/blob/d97ac1dd457dec91a31184b87405592359740007/packages/swingset-runner/demo/encouragementBotComms/swingset.json).

### Security Considerations

n/a

n/a

### Testing Considerations

n/a

### DIFF
--- a/packages/SwingSet/demo/encouragementBot/bootstrap.js
+++ b/packages/SwingSet/demo/encouragementBot/bootstrap.js
@@ -1,13 +1,14 @@
 import { E } from '@agoric/eventual-send';
 import { Far } from '@endo/marshal';
 
-console.log(`=> loading bootstrap.js`);
+console.log('=> loading bootstrap.js');
 
 export function buildRootObject(vatPowers) {
   const log = vatPowers.testLog;
+  log('=> buildRootObject called');
   return Far('root', {
     bootstrap(vats) {
-      console.log('=> bootstrap() called');
+      log('=> bootstrap() called');
       E(vats.user)
         .talkToBot(vats.bot, 'encouragementBot')
         .then(
--- a/packages/swingset-runner/demo/encouragementBot/vat-user.js
+++ b/packages/swingset-runner/demo/encouragementBot/vat-user.js
@@ -2,8 +2,9 @@ import { E } from '@agoric/eventual-send';
 import { Far } from '@endo/marshal';
 
 const log = console.log;
-
 export function buildRootObject(_vatPowers) {
+  // TODO: add a controller command to get the test log
+  // const log = vatPowers.testLog;
   return Far('root', {
     talkToBot(bot, botName) {
       log(`=> user.talkToBot is called with ${botName}`);
@@ -12,7 +13,7 @@ export function buildRootObject(_vatPowers) {
         .then(myEncouragement =>
           log(`=> user receives the encouragement: ${myEncouragement}`),
         );
-      return 'Thanks for the setup. I sure hope I get some encouragement...\nuser vat is happy\n';
+      return 'Thanks for the setup. I sure hope I get some encouragement...';
     },
   });
 }
--- a/packages/SwingSet/demo/encouragementBotComms/bootstrap.js
+++ b/packages/SwingSet/demo/encouragementBotComms/bootstrap.js
@@ -7,7 +7,7 @@ export function buildRootObject(vatPowers) {
   const { D, testLog: log } = vatPowers;
   return Far('root', {
     async bootstrap(vats, devices) {
-      console.log('=> bootstrap() called');
+      log('=> bootstrap() called');
 
       const BOT = 'bot';
       const USER = 'user';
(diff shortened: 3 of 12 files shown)


packages/SwingSet/demo/encouragementBot/bootstrap.js
```diff
@@ -1,13 +1,14 @@
 import { E } from '@agoric/eventual-send';
 import { Far } from '@endo/marshal';
 
-console.log(`=> loading bootstrap.js`);
+console.log('=> loading bootstrap.js');
 
 export function buildRootObject(vatPowers) {
   const log = vatPowers.testLog;
+  log('=> buildRootObject called');
```
FUDCo, line 8: I'm confused by this delta because `vatPowers.testLog` is not referenced in the master branch version of this file, but line 7 doesn't show up as a diff -- it shows the "before" file as having it also.  So I'm not sure what this PR is a PR against.  So something is wrong here.

In any case, you can't use testLog in a swingset-runner demo.  Well, you can, but it's a bit bucket and there will be no output.

packages/swingset-runner/demo/encouragementBot/vat-user.js
```diff
@@ -12,7 +11,7 @@ export function buildRootObject(_vatPowers) {
         .then(myEncouragement =>
           log(`=> user receives the encouragement: ${myEncouragement}`),
         );
-      return 'Thanks for the setup. I sure hope I get some encouragement...\nuser vat is happy\n';
+      return 'Thanks for the setup. I sure hope I get some encouragement...';
```
FUDCo, line 16: This change will break the swingset-runner tests, which I don't see changed in this PR, and indeed the CI run seems to show them not working.

FUDCo (changes requested):
I'm unclear on the purpose of this PR, though perhaps it's just an exercise for familiarization, which is fine.  However, it breaks the thing it changes in a couple of ways, so that will have to be rectified.

I _am_ persistently annoyed by the deltas between the swingset-runner and SwingSEt/vat versions of this, but it has never risen to the top of any priority list because demo code.  A similar and far more annoying thing applies to the swingset-runner demos that want to suck in the various Zoe tests, but have to make do with copy and edit because they aren't structured to be used outside the zoe test directory.  I have an item to fix that at some point, but it's the same never-a-priority problem.

packages/SwingSet/demo/encouragementBot/bootstrap.js
gibson042: The line is present on master, but the binding is never referenced: https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/demo/encouragementBot/bootstrap.js#L7

But it _is_ used in [vat-user](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/demo/encouragementBot/vat-user.js#L8) and [vat-bot](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/demo/encouragementBot/vat-bot.js#L6)... are you saying those don't actually work?

warner:
The `packages/SwingSet/` -side changes look fine in isolation. I'll defer to Chip for the swingset-runner changes and how to best unify the two copies.. he owns swingset-runner.

packages/swingset-runner/demo/encouragementBot/vat-user.js
gibson042: Yes, I admit to leaning on the CI to discover which tests (if any) would be affected. Should be fixed now.

packages/SwingSet/demo/encouragementBot/bootstrap.js
FUDCo: I somehow got confused seeing `demo` in the path and encoded that this was the `swingset-runner` version, not the `SwingSet` version.  I feel better now!

packages/swingset-runner/demo/encouragementBot/vat-user.js
FUDCo: Yah.  I always make it a point to do a `yarn test` in any package containing a file I've changed, and only fall back on CI for finding issues that are due to more distant dependencies.

FUDCo: In any case, (1) as I said above (mistakenly thinking I was looking at a `swingset-runner` file), you can't really use `vatPowers.testLog` in `swingset-runner` demos, and (2) the "user vat is happy" and "bot vat is happy" text are part of the `swingset-runner` tests and need to go back in.

warner:
@gibson042 : Chip says this PR should be cancelled, could you remove it?

Issue #3788, opened 2021-09-02
RUNstake: borrow RUN against staked BLD


michaelfig:
## What is the Problem Being Solved?

per [RUNStake User Stories](https://docs.google.com/document/d/19xP4AjLDdHvlGYFyC_Yq3Opp7oR60rRc-9UKAyb26a4/edit#heading=h.ekia0c1wmi3q) section by @rowgraus and co 5 Apr:

As a BLD staker with no RUNstake position, I need to be able to open a position by placing a lien on my staked BLD and receiving RUN
 - Process should feel familiar to vaults
 - Subject to RUN per BLD limits
 - Subject to asset level debt limit

As a BLD staker with an existing RUNStake position, I need to be able to manage it; that is: adjust my RUN debt and BLD lien within the bounds of the contract.
 - Process should feel familiar to vaults

As a BLD staker with an existing RUNStake position, I need to be able to close it entirely
 - May be combined with/handled by the ‘manage’ use case above

## Description of the Design

Like vaultFactory, the runStake contract provides loans and charges interest and minting fees; differences include:
  1. The loan is secured by a lien on staked BLD, not by transferring collateral.
  2. There is no liquidation. Instead, the staked BLD generates a stream of RUN rewards which can be used to pay down the debt. (See #4244 for more on that part.)

Design notes in [runStake/README.md](https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/runStake/README.md) include a sequence diagram of interactions between a user/dapp, walletBridge, runStake (and attestation), and Cosmos SDK.

We use a semi-fungible `copyBag` token to represent lien attestations: the items is the address where BLD is staked and the quantity is the lien (delta).

 - [x] add `getLiened()` to [attMaker](https://github.com/Agoric/agoric-sdk/blob/dc-run-loc/packages/zoe/src/contracts/attestation/attestation.js#L97) so the wallet can get it before taking out a loan.

related work:

 - #3277
 - #4165
 - #4245
 - #4244
 - #4263
 - #3991 

## Security Considerations

 - line of credit involves use of the RUN mint
 - BLD liens merit care

## Test Plan

 - [x] [RUN LoC Test Cases \- Google Sheets](https://docs.google.com/spreadsheets/d/1XC754C9d6KyEuYkMjrSWNi5xDSvnL6-oc8DQw26vtn4/edit#gid=0)
   - [x] payback
 - [x] use governance for minting ratio
 - [x] charge interest
[...]

Issue #4291, opened 2022-01-13
Sync Zoe JSDoc with API docs


dckc:
## What is the Problem Being Solved?

Zoe JSDoc that @erights has recently focussed on overlaps with end-user doc by @Chris-Hibbert and co. Let's sync them up somehow

## Description of the Design

ref [Dec 17 discussion](https://github.com/Agoric/agoric-sdk/issues/3616#issuecomment-996857365)

Also in the ZEstival, @Chris-Hibbert point out docs on [Handling Authority](https://github.com/Agoric/agoric-sdk/blob/master/packages/zoe/docs/handling-authority.md)

## Security Considerations

_IOU_

## Test Plan

the docs repo has tooling for testing examples

https://docs.agoric.com/reference/zoe-api/user-seat.html and related are sourced from JSDoc instead of Markdown

dckc:
## typedoc?

I looked at.[typedoc](https://github.com/TypeStrong/typedoc) earlier and passed on it because it's for .ts, not JSDoc... But if we head to .d.ts for apis, it could be a match!

dckc:
@Tartuffo I would think that documentation for our smart contract platform doesn't become critical until after MN-1.

turadg:
> ## typedoc?
> I looked at.[typedoc](https://github.com/TypeStrong/typedoc) earlier and passed on it because it's for .ts, not JSDoc... But if we head to .d.ts for apis, it could be a match!

Something that understands TypeScript will be necessary. TypeDoc (and presumably other type-smart tools) look for a unified project, so I think we'll need to https://github.com/Agoric/agoric-sdk/pull/4404 to get any type-aware doc generators to work.

dckc:
just discovered earlier exploration in this area:

 - https://github.com/Agoric/documentation/issues/319

turadg:
https://github.com/Agoric/agoric-sdk/pull/8343 is a fresh attempt.

One package without deps works, but when packages reference each other it chokes because our types don't all resolve as a single project. https://github.com/Agoric/agoric-sdk/pull/4404 would solve that and may make the work worthwhile..

dckc:
I think we're not quite finished with this one. #8343 says...
> Should not be published for external consumption in its current state.

dckc:
recent work in the area:

 - @erights on ownables/revocables https://github.com/Agoric/agoric-sdk/pull/8745#issuecomment-1922700728

dckc:
closely related:

[...]

PR #4736, opened 2022-03-04
chore: improve use of AssetKind type

turadg:
no-ticket, pulled out of https://github.com/Agoric/agoric-sdk/pull/4732/ in support of https://github.com/Agoric/agoric-sdk/issues/3788

## Description

Our current types for AssetKind have these deficiencies:
- the default assetKind  wasn't being picked up so to get it you'd have to pass it in like `AmountMath.makeEmpty(runBrand, 'nat')`
- didn't type the brand, issuer, mint, etc that are linked to the Amount's kind, resulting in more casting than needed
- when an annotation was needed, it was like `Amount<NatValue>`, more verbose than necessary.

This relies on parameterizing assetKind instead of assetValue. That type parameter applies to many other types and doesn't require importing a type in order to use because it's a string literal. E.g. `nat` instead of NatValue type. There's no risk of typos because the [type](https://github.com/Agoric/agoric-sdk/blob/7e1e93241a5cd88a5931240c66bddbdab86b939e/packages/ERTP/src/types.js#L55) parameter K must be a subtype of AssetKind which is the union of 'nat' | 'set' | 'copySet' | 'copyBag', so if someone were to type any string other than one of those four it would be a type failure.  ( There may be an aesthetic argument to use AssetKind.NAT ([const](https://github.com/Agoric/agoric-sdk/blob/7e1e93241a5cd88a5931240c66bddbdab86b939e/packages/ERTP/src/amountMath.js#L14-L25)) but it would be more verbose without any safety benefit: Amount<AssetKind.NAT>.) [A suggestion ](https://github.com/Agoric/agoric-sdk/pull/4732#discussion_r819176376)to use `Amount<NAT>` would be no more verbose and a little cleaner but then you'd have to define or import NAT in the file you're typing that. 

### Security Considerations

--

--

### Testing Considerations

tsc


packages/ERTP/src/amountMath.js
```diff
@@ -86,12 +86,7 @@ const helpers = {
 
 /**
  * @template {AmountValue} V
- * @type {(value: V) =>
- *  V extends NatValue ? 'nat' :
- *  V extends SetValue ? 'set' :
- *  V extends CopySetValue ? 'copySet' :
- *  V extends CopyBagValue ? 'copyBag' :
- *  never}
+ * @type {(value: V) => AssetKindForValue<V>}
```
dckc, line 89: In general, an asset kind determines a value type, but not the other way around. We could have another assetkind that uses bigint values but uses mod p math or some such. The mapping is currently invertible, so I guess this is OK, but it gives me pause.

packages/ERTP/src/amountMath.js
```diff
@@ -148,9 +143,9 @@ const optionalBrandCheck = (allegedBrand, brand) => {
 };
 
 /**
- * @template {AmountValue} [V=AmountValue]
- * @param {Amount<V>} leftAmount
- * @param {Amount<V>} rightAmount
+ * @template {AssetKind} [K=AssetKind]
+ * @param {Amount<K>} leftAmount
+ * @param {Amount<K>} rightAmount
  * @param {Brand | undefined} brand
```
dckc:
```suggestion
 * @param {Brand<K> | undefined} brand
```

Pinning down things like the kind is the reason to supply the optional brand.

packages/ERTP/src/amountMath.js
```diff
@@ -302,9 +297,9 @@ const AmountMath = {
    * whether rectangle A is greater than rectangle B depends on whether rectangle
    * A includes rectangle B as defined by the logic in MathHelpers.
    *
-   * @template {AmountValue} [V=AmountValue]
-   * @param {Amount<V>} leftAmount
-   * @param {Amount<V>} rightAmount
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
    * @param {Brand=} brand
```
dckc:
```suggestion
   * @param {Brand<K>=} brand
```

packages/ERTP/src/amountMath.js
```diff
@@ -316,9 +311,9 @@ const AmountMath = {
    * Returns true if the leftAmount equals the rightAmount. We assume
    * that if isGTE is true in both directions, isEqual is also true
    *
-   * @template {AmountValue} [V=AmountValue]
-   * @param {Amount<V>} leftAmount
-   * @param {Amount<V>} rightAmount
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
    * @param {Brand=} brand
```
dckc:
```suggestion
   * @param {Brand<K>=} brand
```

packages/ERTP/src/amountMath.js
```diff
@@ -333,11 +328,11 @@ const AmountMath = {
    * amount, it usually means including all of the elements from both
    * left and right.
    *
-   * @template {AmountValue} [V=AmountValue]
-   * @param {Amount<V>} leftAmount
-   * @param {Amount<V>} rightAmount
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
    * @param {Brand=} brand
```
dckc:
```suggestion
   * @param {Brand<K>=} brand
```

packages/ERTP/src/amountMath.js
```diff
@@ -352,11 +347,11 @@ const AmountMath = {
    * left amount must include the right amount, this is NOT equivalent
    * to set subtraction.
    *
-   * @template {AmountValue} [V=AmountValue]
-   * @param {Amount<V>} leftAmount
-   * @param {Amount<V>} rightAmount
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
    * @param {Brand=} brand
```
dckc:
```suggestion
   * @param {Brand<K>=} brand
```

packages/ERTP/src/amountMath.js
```diff
@@ -366,21 +361,21 @@ const AmountMath = {
   /**
    * Returns the min value between x and y using isGTE
    *
-   * @template {AmountValue} [V=AmountValue]
-   * @param {Amount<V>} x
-   * @param {Amount<V>} y
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} x
+   * @param {Amount<K>} y
    * @param {Brand=} brand
```
dckc:
```suggestion
   * @param {Brand<K>=} brand
```

packages/ERTP/src/amountMath.js
```diff
@@ -366,21 +361,21 @@ const AmountMath = {
...
   /**
    * Returns the max value between x and y using isGTE
    *
-   * @template {AmountValue} [V=AmountValue]
-   * @param {Amount<V>} x
-   * @param {Amount<V>} y
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} x
+   * @param {Amount<K>} y
    * @param {Brand=} brand
```
dckc:
```suggestion
   * @param {Brand<K>=} brand
```

packages/ERTP/src/issuerKit.js
```diff
@@ -11,10 +11,38 @@ import { makePaymentLedger } from './paymentLedger.js';
...
+ * @param {ShutdownWithFailure=} optShutdownWithFailure If this issuer fails
+ * in the middle of an atomic action (which btw should never happen), it
+ * potentially leaves its ledger in a corrupted state. If this function was
+ * provided, then the failed atomic action will call it, so that some
+ * larger unit of computation, like the enclosing vat, can be shutdown
+ * before anything else is corrupted by that corrupted state.
+ * See https://github.com/Agoric/agoric-sdk/issues/3434
+ * @returns {{
+ *  mint: Mint<K>,
+ *  issuer: Issuer<K>,
```
dckc, line 38: and lo, there was much rejoicing in the land!

packages/ERTP/src/issuerKit.js
```diff
@@ -61,3 +89,5 @@ const makeIssuerKit = (
 harden(makeIssuerKit);
 
 export { makeIssuerKit };
+
+/** @typedef {ReturnType<typeof makeIssuerKit>} IssuerKit */
```
dckc, line 93: I'm fond of this idiom for internal use, but for an external API, it seems more important to spell out the type. Callers should not need to look at the implementation of `makeIssuerKit` to find it out this type. Yes, an IDE can figure it out for you, but we don't always look at code via an IDE.

somewhat related: generating documentation: #4291 . If we can use `tsc` to generate the spelled out type for documentation purposes, maybe this would be OK. But let's not presume that in this PR.

packages/ERTP/src/paymentLedger.js
```diff
@@ -16,12 +16,13 @@ import '@agoric/store/exported.js';
  * Make the paymentLedger, the source of truth for the balances of
  * payments. All minting and transfer authority originates here.
  *
+ * @template {AssetKind} [K=AssetKind]
  * @param {string} allegedName
  * @param {Brand} brand
  * @param {AssetKind} assetKind
```
dckc, line 22: Again, `brand` is how the caller picks `K`, right?

```suggestion
 * @param {Brand<K>} brand
 * @param {AssetKind} assetKind
```

The `assetKind` should match `K` too, but I don't know whether there's a straightforward way to express that.

packages/ERTP/src/paymentLedger.js
```diff
@@ -248,7 +249,12 @@ export const makePaymentLedger = (
     });
   };
 
-  /** @type {MintPayment} */
+  /**
+   * Creates a new Payment containing newly minted amount.
+   *
+   * @param {Amount<K>} newAmount
+   * @returns {Payment}
```
dckc, line 256: Oh... we don't get `Payment<K>` yet? nor `Purse<K>`? Hm... going there does seem a little complex, so I'm fine with that.

packages/ERTP/src/types.js
```diff
@@ -17,8 +17,8 @@
  * relies heavily on polymorphic MathHelpers, which manipulate the unbranded
  * portion.
  *
- * @property {Brand} brand
- * @property {V} value
+ * @property {Brand<K>} brand
+ * @property {AssetValueForKind<K>} value
```
dckc, line 21: spif!

packages/ERTP/src/types.js
```diff
@@ -58,6 +58,21 @@
  */
 
 /**
+ * @template {AssetKind} K
+ * @typedef {K extends 'nat' ? NatValue: K extends 'set' ? SetValue: K extends 'copySet' ? CopySetValue: K extends 'copyBag' ? CopyBagValue : never} AssetValueForKind
```
dckc: stick to 80 columns, please?

packages/ERTP/src/types.js
```diff
@@ -260,55 +277,17 @@
...
- * its value. The allegedName is useful for debugging and double-checking
- * assumptions, but should not be trusted.
- *
- * The assetKind will be used to import a specific mathHelpers
- * from the mathHelpers library. For example, natMathHelpers, the
- * default, is used for basic fungible tokens.
- *
- *  `displayInfo` gives information to UI on how to display the amount.
- *
- * @typedef {Object} IssuerKit
```
dckc, line 287: As noted above, please leave this `IssuerKit` type spelled out for now.

I'm all for moving the `MakeIssuerKit` stuff by the function code.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
```diff
@@ -10,8 +10,8 @@ import { toVaultKey } from './storeUtils.js';
 
 /**
  *
- * @param {Amount<NatValue>} debtAmount
- * @param {Amount<NatValue>} collateralAmount
+ * @param {Amount<'nat'>} debtAmount
+ * @param {Amount<'nat'>} collateralAmount
```
dckc, line 14:
```suggestion
 * @param {Amount<NAT>} debtAmount
 * @param {Amount<NAT>} collateralAmount
```

If there is not a straightforward way to arrange this, then I can't speak for @Chris-Hibbert and @erights in approving tihs.

packages/zoe/src/types.js
```diff
@@ -31,10 +31,11 @@
...
  
  *
+ * @template {AssetKind} [K=AssetKind]
  * @typedef {Object} IssuerRecord
- * @property {Brand} brand
- * @property {Issuer} issuer
- * @property {AssetKind} assetKind
+ * @property {Brand<K>} brand
+ * @property {Issuer<K>} issuer
+ * @property {K} assetKind
```
dckc, line 38: I think this answers one of my earlier questions about constraining `assetKind` to match `brand`.

dckc (changes requested):
Please keep thie `IssuerKit` type spelled out as discussed below.

And unless we can write `'nat'` as `NAT`, I can't speak for @erights and @Chris-Hibbert on this.

packages/ERTP/src/types.js
turadg: We discussed offline. 

The decision rests on whether `IssuerKit` is can be made by multiple functions. If so it's an interface and needs its own type declaration. If not, it's like a class constructor and should use `ReturnType`.

So far nothing else creates objects like this. https://github.com/Agoric/agoric-sdk/blob/16366f0aec5cc5c5face2739768d33a2a16a627b/packages/vats/src/virtual-purse.js#L30-L35 comes close but is not compatible. 

We defer to @erights on whether IssuerKit will be instantiated by other functions and I'm removing that change from this PR so that it doesn't need his review to land.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
turadg: This can totally be done, by defining `const NAT = 'nat'` but since `NAT` is `'nat'` then I think what the user sees in an IDE or type error will still be `'nat'`.

packages/ERTP/src/types.js
dckc: My concern here is mis-placed. `Issuer` is the thing that is interface-like that's important to keep spelled out.

dckc (approved):
let's go for it!

packages/ERTP/src/types.js
```diff
@@ -260,55 +282,17 @@
...
- * @property {DisplayInfo} displayInfo
- */
-
-/**
- * @callback MintPayment
- *
- * Creates a new Payment containing newly minted amount.
- *
- * @param {Amount} newAmount
- * @returns {Payment}
```
mhofman, line 302: I think this type disappeared in the refactor?

packages/ERTP/src/types.js
```diff
@@ -260,55 +282,17 @@
...
 /**
+ * @template {AssetKind} [K=AssetKind]
  * @typedef {Object} Mint
  * Holding a Mint carries the right to issue new digital assets. These
  * assets all have the same kind, which is called a Brand.
  *
- * @property {() => Issuer} getIssuer Gets the Issuer for this mint.
- * @property {MintPayment} mintPayment
+ * @property {() => Issuer<K>} getIssuer Gets the Issuer for this mint.
+ * @property {MintPayment<K>} mintPayment
```
mhofman, line 295: `MintPayment` used here has no declaration.

Issue #4728, opened 2022-03-03
get more RUN against slashed BLD should fail


dckc:
## Describe the bug

Slashing should affect the getRUN limit, but after we get an attestation, I don't see how to detect slashing.

## To Reproduce

This is case 9 resulting from a Dec 6 discussion with @JimLarson and @rowgraus:

9 | Extending LoC - unbonded (FAIL) |   |   |  
-- | -- | -- | -- | --
  |   | buy BLD | 1,000 |  
  |   | stake BLD | 800 |  
  |   | lien BLD | 800 |  
  |   | borrow RUN | 100 |  
  |   | slash | 300 |  
  |   | check BLD staked | 500 |  
  |   | borrow more RUN | 50 | fail
  |   | check RUN balance | 100 |  
  |   | check BLD liened | 800 |  

The "borrow more RUN" step _should_ fail, but the getRUN contract only has an immutable attestation, not a way to check the underlying cosmos state.

## Expected behavior

The "borrow more RUN" step _should_ fail.

## Additional context

This is in development of #3788.

cc @dtribble @JimLarson @rowgraus @turadg

dckc:
Based on discussion with @dtribble , I'm postponing this to the general product backlog.

ivanlei:
This is an antique plan lost to the sands of time.

Issue #4244, opened 2022-01-04
RUN rewards go first to pay off getRUN balance


dckc:
## What is the Problem Being Solved?
BLD stakers can getRUN to get liquid RUN in exchange for a lien on their BLD.

This is somewhat like a vault, but it doesn't liquidate. Instead, until the balance is paid off, any RUN rewards for the relevant account go to pay down the balance.

This ticket covers the functionality needed to intervene in the RUN rewards stream and pay down user debts.

Note: Current plan is this functionality does not launch with Mainnet 1

## Description of the Design
Mainnet 1 Design: N/A; launch with manual payback

Future Design:
After each epoch, when RUN rewards are collected, service iterates over users with outstanding getRUN balances, calculates expected RUN rewards, and reduces balance in lieu of paying out RUN.  

## Acceptance Criteria

* All user with no outstanding getRUN balance receives RUN rewards proportional to BLD staked
* All users with outstanding getRUN balance receives no RUN rewards, but getRUN debt balance is reduced by the amount they should have received based on BLD staked
* All users with getRUN balances which are paid off in full by the RUN rewards for an epoch show a getRUN debt balance of zero and have received the remainder of their RUN rewards

## Test Plan

cc @rowgraus @dtribble @Chris-Hibbert

Tartuffo:
Investigate alternative mechanism - a tool to run through everything and pay off debts.  @dtribble to find out fro Zaki what is new in the Cosmos space for this.  Then Dan can pick it up, but lowest prio for MN-1.

Tartuffo:
@JimLarson to add a ticket to do the design of doing this in Golang, using the hook Jim put in.  They we will decide between implementing this in Go or in JS.  Doing it in Go will likely be faster to impl.

JimLarson:
Wrote a rough design in #4466. It's complicated a little by the need to do some duplicate bookkeeping on the Golang side so it can synchronously know the remaining debt. The golang side could even be the record of truth for the getRUN state. But this interacts with the planned design of the JS level code.

dckc:
On 3 March, @dtribble had an idea: `claimUpTo(address, qty) => claimedQty`
or in batch:

```js
        claimUpTo: async (addresses, debts) => {
          return zip(addresses, debts).map(([a, q]) =>
            min(rewardsPending.get(a) || 0n, q),
          );
        },
```

oops... that's missing denom. Anyway...

We had a good discussion of this and #4466 last Fri (18 March). We recalled
[...]

PR #4741, opened 2022-03-04
feat(run-protocol): RUNstake contract only, without payoff from rewards

dckc:
refs: #3788

## Description

getRUN contract only, without rewards integration (#4244) and without integration with bootstrap, front end / wallet, etc.

### Security Considerations

We plan to grant this contract access to mint RUN.

 -  for governance parameters such as `MintingRatio`, document plausible range of values and risks associated with a setting that is too high or too low.
    - postponed as https://github.com/Agoric/documentation/issues/653

### Performance Considerations

 - [x] re-consider diagnostic value of `checkAccountState(address, lienDelta)` given performance implications noted by @JimLarson 

### Testing Considerations

 - [x] happy-path API usage test
 - [x] regression test for forged attestation
 - [x] tests for requirements identified Dec 6 (with exceptions such as #4728)
 - [x] tests for interest, fees


dckc:
@Chris-Hibbert @dtribble @turadg What do you think about landing this much; that is; without interest, and without integration with the front end and bootstrap?

packages/run-protocol/src/attestation/attestation.js
```diff
@@ -0,0 +1,335 @@
...
+};
+harden(validateInputs);
+
+/**
+ * @param {ContractFacet} zcf
+ * @param {ZCFMint} zcfMint
+ * @param {Amount} amountToMint
+ * @returns {Promise<Payment>}
+ */
+const mintZCFMintPayment = (zcf, zcfMint, amountToMint) => {
```
Chris-Hibbert: why does `addReturnableLien` want to return a payment directly, rather than adding an allocation to a seat so Zoe can manage offer safety and invitations and such?

packages/run-protocol/src/getRUN.js
```diff
@@ -0,0 +1,338 @@
...
+
+  /** @type { OfferHandler } */
+  const makeLineOfCreditHook = seat => {
+    return makeLineOfCreditKit(zcf, seat, creditPolicy, runMint, runBrand);
+  };
+
+  const publicFacet = wrapPublicFacet(
+    Far('getRUN Public API', {
+      getIssuer: att.publicFacet.getIssuer,
+      getBrand: () => att.publicFacet.getBrand,
```
Chris-Hibbert: Doesn't this want to set getBrand to the getBrand function, rather than setting it to be a function that returns a function?  Or am I mis-reading it?
```suggestion
      getBrand: att.publicFacet.getBrand,
```

packages/run-protocol/test/test-getRUN.js
```diff
@@ -0,0 +1,619 @@
...
+  t.deepEqual(actual, proposal.want.RUN);
+});
+
+/**
+ * @param { StartFaker['publicFacet'] } faker
+ * @param { Brand } bldBrand
+ * @returns { Promise<[Amount, Payment]> }
+ *
+ * @param { (faker: ERef<StartFaker['publicFacet']>, bldBrand: Brand)
+ *            => Promise<[Amount, Payment]> } [mockAttestation]
```
dckc: oops... there's some left-over cruft here from an earlier code organization

packages/run-protocol/test/runLoC-sheet-pivot.js
```diff
@@ -0,0 +1,64 @@
+import { ROWS } from './runLoC-test-case-sheet.js';
```
dckc: this file is probably not worth keeping; it's from an earlier design

packages/run-protocol/test/runLoC-test-case-sheet.js
```diff
@@ -0,0 +1,41 @@
+import * as CSV from './csvParse.js';
```
dckc: likewise, this file is probably not worth keeping; it's from an earlier design

packages/run-protocol/test/csvParse.js
```diff
@@ -0,0 +1,14 @@
+// @ts-check
```
dckc: this special-case CSV parser should be tucked into `getRUN-test-sheet.js`, which is the only place that uses it

packages/run-protocol/test/test-getRUN.js
```diff
@@ -0,0 +1,619 @@
...
+/**
+ * @typedef {EconomyBootstrapPowers & WellKnownSpaces & {
+ *   consume: {
+ *     client: ERef<ClientManager> // to add home.attMaker
+ *   },
+ *   produce: {
+ *     client: Producer<ClientManager>
+ *   }
+ * }} GetRunPowers
+ */
```
dckc: oops; I think this is obsolete

dckc:
What made me think of landing this much is: it basically handles the cases that @rowgraus and @JimLarson and I worked out in a spreadsheet in December:

 - [getRUN-test-steps.js](https://github.com/Agoric/agoric-sdk/blob/dc-getRUN-contract/packages/run-protocol/test/getRUN-test-steps.js)
 - [getRUN Test Cases](https://docs.google.com/spreadsheets/d/1XC754C9d6KyEuYkMjrSWNi5xDSvnL6-oc8DQw26vtn4/edit#gid=0)

I'm going to take a break and think it over...

Chris-Hibbert (commented):
Oops. I started looking at this, and noticed a few things.  Sending them on so they don't get lost.

Sorry I didn't make more progress with the review.

packages/run-protocol/src/attestation/attestation.js
dckc: I don't know. It could be left over from when there were to kinds of attestations. I got rid of the timed attestation code and tested that the remaining stuff passes its tests but I haven't gotten as far as I'd like getting down to only the essential code.

packages/run-protocol/src/getRUN.js
dckc: that looks funny to me too. I suspect it's not tested.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+  /** seat keyword for use in offers to return an attestation. */
+  Attestation: 'Attestation',
+});
+
+/**
+ * Find-or-create value in store.
+ *
+ * @type {<K, V>(store: Store<K, V>, key: K, thunk: () => V) => V}
+ */
+const getOrElse = (store, key, make) => {
```
turadg, line 20: is this something `Store` itself will support @FUDCo ?

packages/run-protocol/src/contractSupport.js
```diff
@@ -0,0 +1,55 @@
+// The proposal is not allowed to include any keys other than these,
```
turadg: looks like this was meant to be part of a doc comment for `assertOnlyKeys`

packages/run-protocol/src/contractSupport.js
```diff
@@ -0,0 +1,55 @@
...
+ * the gain and a loss on the `fromSeat`. The gain/loss are typically from the
+ * give/want respectively of a proposal. The `key` is the allocation keyword.
+ *
+ * @param {ZCFSeat} fromSeat
+ * @param {ZCFSeat} toSeat
+ * @param {Amount} fromLoses
+ * @param {Amount} fromGains
+ * @param {Keyword} key
+ */
+export const transfer = (fromSeat, toSeat, fromLoses, fromGains, key) => {
```
turadg: +1 to pulling this out in a shared module. Heads up there will be some merge conflicts after https://github.com/Agoric/agoric-sdk/pull/4877/ (e.g. `transfer` -> `stageDelta` atm)

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -442,3 +443,163 @@ export const startRewardDistributor = async ({
   );
 };
 harden(startRewardDistributor);
+
+/**
+ * @typedef { import('@endo/eventual-send/').Unpromise<T> } Unpromise<T>
```
turadg: `Unpromise` is deprecated (in agoric-sdk before those changes were globbered). Instead use `Awaited` built into TS

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,125 @@
...
+  const att = await makeAttestationFacets(
+    zcf,
+    stakeBrand,
+    lienAttestationName,
+    lienBridge,
+  );
+  // TODO: remove publicFacet wart
+  const attestBrand = await E(att.publicFacet).getBrand();
+
+  const paramManager = await makeRunStakeParamManager(
```
turadg: https://github.com/Agoric/agoric-sdk/pull/4868 will make it so you don't have to hold the full `paramManager` or check that electorate matches.

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,125 @@
...
+
+  // assertElectorateMatches(paramManager, otherGovernedTerms);
+
+  /**
+   * We provide an easy way for the policy to add rewards to
+   * the rewardPoolSeat, without directly exposing the rewardPoolSeat to them.
+   *
+   * @type {ReallocateWithFee}
+   */
+  const reallocateWithFee = (fee, fromSeat, otherSeat = undefined) => {
```
turadg: check out https://github.com/Agoric/agoric-sdk/pull/4877/files#diff-0b7d727a28b84089ac4ebe9b76f8957bb71bfee261640828bad01c4189cbe73fR89

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const publicFacet = wrapPublicFacet(
+    Far('getRUN Public API', {
+      makeLoanInvitation: () =>
+        zcf.makeInvitation(makeRUNstakeHook, 'make RUNstake', undefined),
+      makeReturnAttInvitation: att.publicFacet.makeReturnAttInvitation,
+    }),
+  );
+
+  return { publicFacet, creatorFacet: wrapCreatorFacet(att.creatorFacet) };
```
turadg: surprising to inline one wrap and not the other

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,351 @@
...
+import { calculateCurrentDebt, reverseInterest } from '../interest-math.js';
+
+const { details: X, quote: q } = assert;
+
+const trace = makeTracer('R1');
+
+const VaultPhase = {
+  ACTIVE: /** @type {'Active'} */ ('Active'),
+  CLOSED: /** @type {'Closed'} */ ('Closed'),
+};
```
turadg: not a vault. do we have a name yet for this thing that can be active or closed? it's in the state of `RUNStakeKit` but a kit shouldn't have a state. Maybe `StakingAccount`? 

Also `{const}` can save some typing
```suggestion
const AccountPhase = /** @type {const} */ {
  ACTIVE: 'Active',
  CLOSED: 'Closed',
};
```

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * Make RUNstake kit, subject to policy.
+ *
+ * @param {ZCF} zcf
+ * @param {ZCFSeat} startSeat
+ * @param {ReturnType<import('./runStakeManager.js').makeRunStakeManager>} manager
+ * @param { ZCFMint<'nat'> } mint
+ * return value follows the wallet invitationMakers pattern
+ * @throws if startSeat proposal is not consistent with governance parameters in manager
```
turadg: TIL `@throws`. Will try to use that more

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,351 @@
...
+  [VaultPhase.ACTIVE]: [VaultPhase.CLOSED],
+  [VaultPhase.CLOSED]: [],
+};
+
+/**
+ * Make RUNstake kit, subject to policy.
+ *
+ * @param {ZCF} zcf
+ * @param {ZCFSeat} startSeat
+ * @param {ReturnType<import('./runStakeManager.js').makeRunStakeManager>} manager
```
turadg: consider defining `RunStakeManager` in its file and importing that.

[this may become standard practice](https://docs.google.com/document/d/1bRIUeOokKcbIHZHt68d9kyiJOB5YEUQVpLpFAmDLkhk/edit#heading=h.6lmefd8kl5rm) for `makeFoo` functions.

packages/run-protocol/test/runStake/runStake-test-sheet.js
```diff
@@ -0,0 +1,200 @@
...
+  const splitFields = line =>
+    [...line.matchAll(fieldPattern)].map(([_all, txt]) => unquote(txt));
+
+  /** @param { string } content */
+  const parse = content => content.split('\n').map(splitFields);
+
+  return harden({ parse });
+})();
+
+// copy/paste from spreadsheet seems to give tab-separated data
```
turadg: can the spreadsheet be linked here?

packages/run-protocol/test/runStake/runStake-test-sheet.js
```diff
@@ -0,0 +1,200 @@
+// @ts-check
+/**
+ * Test steps with support for spreadsheet integration.
+ */
+import util from 'util';
+
+// ack: https://stackoverflow.com/a/18147076/7963
+const CSV = (() => {
```
turadg: since this is test code, seems safe to use a CSV library. maybe one that supports TSV

packages/run-protocol/test/runStake/runStake-test-sheet.js
```diff
@@ -0,0 +1,200 @@
...
+        }
+      }
+    }
+  }
+  if (label) {
+    yield [num, label, steps];
+  }
+}
+
+export const CASES = [...eachCase(ROWS)];
```
turadg: How is this export used? I thought it was imported into the test, but I don't see this module being imported.

If it actually is imported, then consider doing this as a script instead and check the JS file into SCM since this value only changes when someone copies in new values from the spreadsheet.

If it's being used in that way already to create a fixture, then put comments in the fixture for how to update it using this script. Though if they were to edit in place that wouldn't necessarily be a problem.

packages/run-protocol/test/runStake/test-runStake.js
```diff
@@ -0,0 +1,773 @@
...
+
+const SECONDS_PER_HOUR = 60n * 60n;
+const SECONDS_PER_DAY = 24n * SECONDS_PER_HOUR;
+
+test.before(async t => {
+  /** @param { string } ref */
+  const asset = async ref =>
+    new URL(await metaResolve(ref, import.meta.url)).pathname;
+
+  t.log('bundling...', contractRoots);
```
turadg, line 55: How I wish we could cache bundling builds.

turadg (commented):
Did a cursory review

packages/run-protocol/src/runStake/attestation.js
FUDCo: I confess I don't understand what you are asking.

turadg: This PR defines a `getOrElse` function that lets the caller say "give me the value for this key from the store, or if it's not there define it in the store". That seems generally useful so I asked if the Store API might offer that in the future. If not the `MapStore` type then perhaps a helper like this within `@agoric/store` ?

FUDCo: I don't think there are any plans for anything like this.  The authority to consult would be @erights, who I tentatively predict, on the basis of no knowledge whatsoever on my part, will have a subtle and nuanced argument why doing this is would be a bad idea.

Chris-Hibbert:
Dan asked
> What do you think about landing this much; that is; without interest, and without integration with the front end and bootstrap?

I don't mind omitting frontend integration from first merge. 

I have found integration with bootstrap to be informative, and based on conversations with you, I've attempted to make test setup consistent with it, which has made that integration valuable to do early.

Interest is straightforward. 

But parameter governance and interest charging and other library support have been evolving quickly. Perhaps merging early eases migration support enough to be worth the added pain. If it's not ready for use in the economy, the obvious step is to not activate in bootstrap.

packages/run-protocol/test/test-getRUN.js
dckc: that's gone.

packages/run-protocol/test/runLoC-sheet-pivot.js
dckc: deleted.

packages/run-protocol/test/csvParse.js
dckc: moved.

packages/run-protocol/src/attestation/attestation.js
dckc: I figured it out and added a comment:

```
 * To support `makeAttestation`, we directly mint attestation payments.
 * We still use a ZCFMint because returning attestations is done
 * through Zoe offers.
```

packages/run-protocol/src/getRUN.js
dckc: This is subsumed by using `getTerms()` to get brands and issuers.

packages/run-protocol/src/runStake/runStake.js
dckc: my eyes are getting blurry. I'm struggling to see the motivation or something.

turadg: sorry, I considered being more specific and should have been. The link shows an in-flight PR where `reallocateWithFee` has been expanded in scope to `mintAndReallocate`. I thought you might want to consider the same for parity.

packages/run-protocol/src/runStake/runStakeKit.js
dckc: Very nearby we use "account" to refer to cosmos accounts. I'm trying out `LoanPhase`...

packages/run-protocol/test/runStake/runStake-test-sheet.js
dckc: In case you're curious, it's linked from #3788 . But I no longer maintain the tests this way. (more below...)

packages/run-protocol/test/runStake/runStake-test-sheet.js
dckc: > How is this export used?

It's not. I ended up generating .js code from the TSV data instead of doing the computation at runtime (well, test time). And the TSV data has fallen behind `runStake-test-steps.js`. I suppose it's time to let it go. I'm deleting it.

packages/run-protocol/src/contractSupport.js
```diff
@@ -0,0 +1,54 @@
...
+ * be empty because gain/loss comes from the give/want for a specific asset
+ * on a proposal. We use two Amounts because an Amount cannot represent
+ * a negative number (so we use a "loss" that will be subtracted).
+ *
+ * @param {Amount} base
+ * @param {Amount} gain
+ * @param {Amount} loss
+ * @returns {Amount}
+ */
+export const applyDelta = (base, gain, loss) =>
```
turadg: https://github.com/Agoric/agoric-sdk/pull/4877 calls this `addSubtract` which seems more apt.

packages/run-protocol/src/contractSupport.js
```diff
@@ -0,0 +1,54 @@
...
+  AmountMath.subtract(AmountMath.add(base, gain), loss);
+
+// usually 'Collateral' and 'RUN'.
+export const assertOnlyKeys = (proposal, keys) => {
```
turadg, line 25: worth a type annotation

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -442,3 +443,158 @@ export const startRewardDistributor = async ({
...
+  const poserInvitationP = E(
+    economicCommitteeCreatorFacet,
+  ).getPoserInvitation();
+  const [initialPoserInvitation, electorateInvitationAmount] =
+    await Promise.all([
+      poserInvitationP,
+      E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
+    ]);
+
+  // t.log({ electorateCreatorFacet, electorateInstance });
```
turadg: remove before merge

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -442,3 +443,158 @@ export const startRewardDistributor = async ({
...
+        config.mintingRatio[1],
+        bldBrand,
+      ),
+      interestRate: makeRatio(config.interestRateBP, runBrand, BASIS_POINTS),
+      loanFee: makeRatio(config.loanFeeBP, runBrand, BASIS_POINTS),
+      electorateInvitationAmount,
+    },
+  );
+
+  /** @type {{ publicFacet: GovernorPublic, creatorFacet: GovernedContractFacetAccess<any>}} */
```
turadg: `any` -> `unknown`

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -442,3 +443,158 @@ export const startRewardDistributor = async ({
...
+        terms: runStakeTerms,
+        issuerKeywordRecord: { Stake: bldIssuer },
+        privateArgs: { feeMintAccess, initialPoserInvitation, lienBridge },
+      }),
+    },
+    harden({ economicCommitteeCreatorFacet }),
+  );
+
+  const governedInstance = await E(governorFacets.creatorFacet).getInstance();
+  const creatorFacet = E(governorFacets.creatorFacet).getCreatorFacet();
```
turadg, line 630: creatorFacet.getCreatorFacet() is quite confusing. @Chris-Hibbert maybe that's another thing for https://github.com/Agoric/agoric-sdk/issues/4818

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -442,3 +443,158 @@ export const startRewardDistributor = async ({
...
+
+  const {
+    issuers: { BldLienAtt: attIssuer },
```
turadg: The naked keyword makes sense to me but I see most code keep a manifest constant for this. Why not use one here?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+ */
+const mintZCFMintPayment = (zcf, zcfMint, amountToMint) => {
+  const { userSeat, zcfSeat } = zcf.makeEmptySeatKit();
+  zcfMint.mintGains(harden({ [KW.Attestation]: amountToMint }), zcfSeat);
+  zcfSeat.exit();
+  return E(userSeat).getPayout(KW.Attestation);
+};
+
+/**
+ * Make an ZCFMint attentuated so that minting represents
```
turadg:
```suggestion
 * Make an attentuated ZCFMint so that minting represents
```

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+ * @param {ERef<StakingAuthority>} lienBridge bridge to account state
+ */
+const makeAttestationKit = async (zcf, keyword, stakeBrand, lienBridge) => {
+  const { add, subtract, makeEmpty, isGTE, coerce } = AmountMath;
+  const empty = makeEmpty(stakeBrand);
+  const zcfMint = await zcf.makeZCFMint(keyword, AssetKind.COPY_BAG);
+  const { issuer, brand: attBrand } = zcfMint.getIssuerRecord();
+
+  /**
+   * @param {string} address
```
turadg: what kind of address? is there any validation to do of the value?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+  const wrapLienedAmount = (address, lienedAmount) =>
+    AmountMath.make(attBrand, makeCopyBag([[address, lienedAmount.value]]));
+
+  /** @param {Amount<'copyBag'>} attestationAmount */
+  const unwrapLienedAmount = attestationAmount => {
+    assert(
+      attestationAmount.brand === attBrand,
+      X`The escrowed attestation ${attestationAmount} was not of the attestation brand ${attBrand}`,
+    );
+    // Caller is responsible to split payload of >1 item.
```
turadg: something for the jsdoc?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+    fit(attestationAmount.value.payload, harden([[M.string(), M.bigint()]]));
+    /** @type {[string, bigint][]} */
+    const [[address, valueReturned]] = attestationAmount.value.payload;
```
turadg, line 84: nit
```suggestion
    const [address, valueReturned] = attestationAmount.value.payload[0];
```

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+    const lienedAmount = AmountMath.make(stakeBrand, valueReturned);
+    return { address, lienedAmount };
+  };
+
+  /** @type {(hi: Amount, lo: Amount) => Amount} */
+  const subtractOr0 = (hi, lo) => (isGTE(hi, lo) ? subtract(hi, lo) : empty);
```
turadg: hax0r codez!

jk/ but consider `subtractOrEmpty` or `floorSubtract`.

also, param names are confusing since `lo` may be greater than `hi`.  I had to look it up, but apparently there are mathematical terms: Minuend - Subtrahend = Difference.

```suggestion
  const subtractToZero = (m, s) => (isGTE(m, s) ? subtract(m, s) : empty);
```

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+  // check that x <= hi - lo
+  const checkDelta = (x, label, hi, lo) => {
+    const bound = subtractOr0(hi, lo);
+    assert(
+      isGTE(bound, coerce(stakeBrand, x)),
+      // TODO: X`...${raw(label)}...`
+      X`Only ${bound} was ${label}, but an attestation was attempted for ${x}`,
+    );
+  };
+  const checkAccountState = async (address, toLien) => {
```
turadg: do we have a convention for what `checkFoo` means? I've seen it used to throw and to return a value. I think returning a value is bad and throwing is good. But if we are going to return values from `checkFoo` then I think we should have a clear verb for throwing, like `validateFoo`.

packages/run-protocol/src/econ-behaviors.js
Chris-Hibbert: should this be dropped?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+ * @param {ZCF} zcf
+ * @param {Brand} stakeBrand
+ * @param {string} keyword for attestation issuer
+ * @param {ERef<StakingAuthority>} lienBridge
+ *
+ * NOTE: the liened amount is kept both here in JS and on the
+ * other side of the lienBridge. The mint created here
+ * is the authority on the liened amount. The liened amount
+ * on the other side of the bridge may lag. Do not create two
+ * mints that share a bridge.
```
turadg, line 216: Is there a way to automatically detect this?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+
+  const { issuer, brand, lienMint } = await makeAttestationKit(
+    zcf,
+    keyword,
+    stakeBrand,
+    lienBridge,
+  );
+
+  /** @param { string } address */
+  const makeAttMaker = address =>
```
turadg: I assume "Att" stands for "attestation" but the returned object is more than an attestation maker. is this an "attestation kit"?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+    creatorFacet: Far('creator', {
+      /**
+       * IMPORTANT: The attestation maker should only be given to the owner of the
+       * address. Only the owner can authorize creation of attestations and
+       * the resulting liens on their underlying tokens.
+       *
+       * @param { string } address
+       */
+      provideAttestationMaker: address => {
+        assert.typeof(address, 'string');
```
turadg: feeling again the need to know what kind of address and whether a valid one has any more constraints than "string"

packages/run-protocol/src/runStake/params.js
```diff
@@ -0,0 +1,89 @@
+// @ts-check
+
+import {
+  CONTRACT_ELECTORATE,
+  makeParamManager,
+  ParamTypes,
+} from '@agoric/governance';
+
+/** RUNstakeParams are the parameters subject to governance. */
+export const RUNstakeParams = /** @type { const } */ ({
```
turadg: I don't know if we have style guidelines on this already but I advocate that acronyms by treated as words when smooshed: `RunStakeParams`.

(though is RUN an acronym or just a haughty word?)

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+import { makeAttestationFacets } from './attestation.js';
+import { makeRunStakeParamManager } from './params.js';
+import { makeRunStakeKit } from './runStakeKit.js';
+import { makeRunStakeManager } from './runStakeManager';
+
+/**
+ * Provide loans on the basis of staked assets that earn rewards.
+ *
+ * In addition to brands and issuers for `Staked`, `RUN`, and attestation,
+ * terms of the contract include a periodic interest rate plus
```
turadg: would help to have the term keywords mentioned in the text.
```suggestion
 * terms of the contract include a periodic interest rate (`InterestRate`) plus
```

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * Finally, `Close` the loan, providing `{ give: RUN: debtAmount }}`
+ *
+ * To start the contract, authorize minting assets by providing `feeMintAccess`
+ * and provide access to the underlying staking infrastructure in `lienBridge`.
+ *
+ * @typedef {{
+ *   feeMintAccess: FeeMintAccess,
+ *   initialPoserInvitation: Invitation,
+ *   lienBridge: ERef<StakingAuthority>,
```
Chris-Hibbert, line 64: does anything related to the lienBridge need to be legible in order for clients to trust it?

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * The creator facet can make an `AttMaker` for each account, which
+ * authorizes placing a lien some of the staked assets in that account.
+ * @typedef {{
+ *   provideAttestationMaker: (addr: string) => AttMaker,
+ *   getLiened: (address: string, brand: Brand<'nat'>) => Amount<'nat'>,
+ * }} RunStakeCreator
+ *
+ * @type {ContractStartFn<RunStakePublic, RunStakeCreator,
+ *                        RunStakeTerms, RunStakePrivateArgs>}
```
Chris-Hibbert, line 75: Awesome documentation to go with the types!

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @typedef {{
+ *   electionManager: VoteOnParamChange,
+ *   main: {
+ *     MintingRatio: ParamRecord<'ratio'>,
+ *     InterestRate: ParamRecord<'ratio'>,
+ *     LoanFee: ParamRecord<'ratio'>,
+ *     Electorate: ParamRecord<'invitation'>,
+ *   },
+ * }} RunStakeGovernanceTerms
```
turadg: Since `Electorate` and `electionManager` are part of all governance so for documentation purposes here it may be better to separate out,
```
@typedef {{
     MintingRatio: ParamRecord<'ratio'>,
     InterestRate: ParamRecord<'ratio'>,
     LoanFee: ParamRecord<'ratio'>,
}} RunStakeParamTerms
```
then,
```
 * @typedef {{
 *   electionManager: VoteOnParamChange,
 *   main: RunStakeParamTerms & {
 *     Electorate: ParamRecord<'invitation'>,
 *   },
 * }} RunStakeGovernanceTerms
```
The simplification PR has a utility type for the latter.

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * As in vaultFactory, `timerService` provides the periodic signal to
+ * charge interest according to `chargingPeriod` and `recordingPeriod`.
+ *
+ * @typedef { RunStakeGovernanceTerms & {
```
turadg: I like this technique of interleaving the typedefs with explanation

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * The public facet provides access to invitations to get a loan
+ * or to return an attestation in order to release a lien on staked assets.
+ *
+ * @typedef {{
+ *   makeLoanInvitation: () => Promise<Invitation>,
+ *   makeReturnAttInvitation: () => Promise<Invitation>,
+ * }} RunStakePublic
+ *
+ * To take out a loan, get an `AttMaker` for your address from
```
turadg: do we really need to abbreviate `AttestationMaker` type? clearly you'll want to name an instance shorter but the type should err towards prioritize unambiguity

turadg (commented):
First pass, got about half-way down.

I'd love to see a `runStake/README.md` before my next pass. The contract has a lot about how it's used. In the README I'd like to see how it was designed and why.

Also somewhere I'd like to see Attestation defined.

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -0,0 +1,171 @@
...
+    { timerService, chargingPeriod, recordingPeriod, startTimeStamp },
+  );
+
+  const publicFacet = wrapPublicFacet(
+    Far('runStake public interface', {
+      makeLoanInvitation: () =>
+        zcf.makeInvitation(
+          seat => makeRunStakeKit(zcf, seat, manager, runMint),
+          'make RUNstake',
+          undefined,
```
Chris-Hibbert: Unnecessary. I'd omit it.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,256 @@
...
+    });
+
+  return harden({
+    publicFacet: Far('attestation publicFacet', {
+      getIssuer: () => issuer,
+      getBrand: () => brand,
+      makeReturnAttInvitation: () =>
+        zcf.makeInvitation(lienMint.returnAttestation, 'returnAttestation'),
+    }),
+    creatorFacet: Far('creator', {
```
Chris-Hibbert, line 253: I don't think I'd call these `publicFacet` and `creatorFacet`, since they're intended to be providing components of the public/creator facets to runStake which will mix them with others. Even though creatorFacet is currently used as a whole, `runStake.js` may need to add other methods later from other sources.   

What do you think of `publicMethods` and `creatorMethods`?  I'd have called them `att*Methods`, but they're accessed as `att.publicFacet` already.

The wrapCreatorFacet call would read better to me as 

```
 const creatorFacet = wrapCreatorFacet({ ...att.creatorMethods });
```

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,365 @@
...
+  CLOSED: 'Closed',
+});
+
+/**
+ * @typedef {LoanPhase[keyof LoanPhase]} LoanPhaseValue
+ * @type {{[K in LoanPhaseValue]: Array<LoanPhaseValue>}}
+ */
+const validTransitions = {
+  [LoanPhase.ACTIVE]: [LoanPhase.CLOSED],
+  [LoanPhase.CLOSED]: [],
```
Chris-Hibbert: Are you expecting this to grow into a more interesting state diagram? Otherwise representing the states at all seems like overkill.

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,365 @@
...
+    collateralBrand,
+    AssetKind.COPY_BAG,
+  );
+  const emptyDebt = AmountMath.makeEmpty(debtBrand);
+
+  const { zcfSeat: vaultSeat } = zcf.makeEmptySeatKit();
+
+  /**
+   * Calculate the fee, the amount to mint and the resulting debt.
+   * The give and the want together reflect a delta, where the
```
Chris-Hibbert: fragment

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,365 @@
...
+      want: { RUN: runWanted },
+    } = startSeat.getProposal();
+
+    const { maxDebt } = manager.maxDebtForLien(attestationGiven);
+    assert(
+      AmountMath.isGTE(maxDebt, runWanted),
+      X`wanted ${runWanted}, more than max debt (${maxDebt}) for ${attestationGiven}`,
+    );
+
+    // return checkBorrow(attAmt, runWanted);
```
Chris-Hibbert: Will this be needed?

packages/run-protocol/src/runStake/runStake.js
dckc: > do we really need to abbreviate `AttestationMaker` type?

No.

I inherited the name `AttMaker`. Since it was previously approved and merged, I was reluctant to change it. But I agree to change it now.

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,365 @@
...
+    const { phase } = state;
+    const uiState = snapshotState(phase);
+    trace('updateUiState', uiState);
+
+    switch (state.phase) {
+      case LoanPhase.ACTIVE:
+        uiUpdater.updateState(uiState);
+        break;
+      case LoanPhase.CLOSED:
+        uiUpdater.finish(uiState);
```
Chris-Hibbert: I think it's an error to call `finish` more than once. I think that means if updateUiState is called again after closing the vault, it'll generate an error.

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,365 @@
...
+    const giveColl = proposal.give.Attestation || emptyCollateral;
+    const wantColl = proposal.want.Attestation || emptyCollateral;
+
+    // new = after the transaction gets applied
+    const newCollateral = applyDelta(collateral, giveColl, wantColl);
+    // max debt supported by current Collateral as modified by proposal
+    const { amountLiened, maxDebt: newMaxDebt } =
+      manager.maxDebtForLien(newCollateral);
+
+    const giveRUN = AmountMath.min(proposal.give.RUN || emptyDebt, debt);
```
Chris-Hibbert: They can't give more than they currently owe? I don't get this. I'll look for a test case that makes it clear.   ...  okay, got it. If they try to pay back more than their debt, this will leave the excess allocated to the seat. Zoe will make that look the same as if runStake accepted it and reallocated it back.

I think the code will work fine, but I think this isn't covered by a test case. It appears that `payoffRun` doesn't put the returned overpayment back in the RUN purse.

If you write a test that attempts to pay more than the current debt, I'd expect the test to not show the overpayment being returned until `payoffRun` deposits any returned RUN back into the purses.

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -0,0 +1,365 @@
...
+      ),
+    );
+
+    zcf.reallocate(seat, vaultSeat);
+
+    mint.burnLosses(harden({ RUN: currentDebt }), vaultSeat);
+    seat.exit();
+    assignPhase(LoanPhase.CLOSED);
+    updateDebtSnapshot(emptyDebt);
+    updateUiState();
```
Chris-Hibbert, line 299: should it call `assertVaultHoldsNoRun()`?

packages/run-protocol/src/runStake/attestation.js
erights: @dtribble I remember you had a good verb to mean something like this `getOrElse`. Do you remember what it was?

@FUDCo I will try to avoid taking that as a challenge ;) .

I would be fine with a helper. I am reluctant to extend the type itself, but I do not currently have any principled objection. Just a general sense that the object's own API should remain reasonably small and any addition must pay for itself.

Whether helper or method, should the `make` function take the key as a parameter?

Chris-Hibbert (commented):
Review unfinished.   Sending findings so far.

dckc:
> I'd love to see a `runStake/README.md` before my next pass. The contract has a lot about how it's used. In the README I'd like to see how it was designed and why.

Do you have an example of that sort of thing? I don't recall seeing one.

I'm not sure I know how it was designed and why, beyond: a bunch of meetings and stuff. I suppose one bit of "why" is the motivating use case in the referenced issue, #3788: 
> Alice wants to buy a baseball card, priced at 1 RUN. She has no run, but she has BLD. She would like to borrow RUN against her BLD to buy the card.

turadg:
From our call,
[](https://mermaid.live/edit#pako:eNqVVE1v2kAQ_Ssjn4hCVNqGiw-RIKnUqmkVQaNcfNnsjs0Ke9fdj1CE-O-dXduKjUmjckCGnff2zZvnOSRcC0zSJFMWf3tUHO8kKwyrMgX0YdxpA4LVdfO7ZsZJLmumHOxYWaJbGikKHJ8y59A65qRW40Pj1dqx7RlYoUumiqzFKO0QjCw2DnQeZaTwVe_iE-RSCQtug8C9MUjgcB-ex_bFpoNfIC3URr9IgQKcjtwWmBLAtXJGU6WA5z08RQw8fmvoo4Srm5vLE-oC3YM3FpeMGuE4yZLl_V2WXPRRBDtBzWez2XnhPR9TYMXHT5-DYEbibKXpQQiD1g41DUGk6Af7M4nYVkbvPNY3rsfSBefaK7cOVg4wQ1kd4raRYX1da-MslBLbNpqKptlhE80NcVreglSw20i-gWvyAMisyIFirLThalLwXyTnYvTLILPe7OlwB1v6sv8K3VPIJoWDDILr-QxWjz9DJsIlkpqcN5eeTGEUjBU6bxR7LnHx2tZkR9xpxzqFQr40gQiMV1R48X6el6XmWwvkkbdogARQoFkZbYmFDcUg9qO5VPRC9nW1Et6JjEV3H61eVGEgkziASxiC38zC4vY70IJxG9MbwOCqsZEPbF-Ft51c6LnUDjx4EFPbkY167nZPSgQ5mbX7ULeEOSk5jOyfwmBAR-iYO6Kg8Q2JgZFwIS1hocQLV2h9eXYgV124adQyl2gs5EZXY1gyTSo0FZOCNvch_JcltAcrzJKUHgUz24w2-pHqfC3oRf4iJO3xJM1ZaXGaMO_0eq94kjrjsStqF39bdfwL32kM1w)

^ uses the "Copy to Markdown" button. @arirubinstein pointed out we could also copy the source and let Github render:

```mermaid
sequenceDiagram
    actor dapp
    participant walletBridge
    participant attestation
    participant runStake
    participant golang

    note right of dapp: How dapp finds the current state

    note right of walletBridge: walletBridge is provided to dapps and controlled by Wallet UI
    dapp ->>+ walletBridge: getPurseBalance("BLD")
    dapp -->> walletBridge: 5000

    note right of attestation: ag123 is a cosmos address
    dapp ->> attestation: getMax(ag123)
    attestation ->> golang: getAccountState(ag123)
    note right of golang: Cosmos supports lien
    golang -->> attestation: account status in which 4000 BLD liened
    attestation  -->> dapp: account status in which 4000 BLD liened
    note right of dapp: Treasury now knows

    note right of dapp: Want to get 450 RUN by liening 500 BLD
    dapp ->> walletBridge: getReturnableAttestation(want: 450 RUN, give: 500 BLD-Att)
    note right of walletBridge: Blocks on user approval in wallet
    walletBridge ->> attestation: makeAttestation(500 BLD)
    attestation ->> golang: setLienedAmount(4000 + 500 BLD)
    golang -->> attestation: ACK or throws

    attestation -->> walletBridge: Payment of 500 BLD-Att liened on ag123

    walletBridge ->> runStake: offer w/payment for {give: 500 BLD-Att, want: 450 RUN} 

    runStake --> walletBridge: Payment for 450RUN and offerResult
    walletBridge --> dapp: notifiers from offerResult
```

packages/run-protocol/src/econ-behaviors.js
dckc: Hm... I find this sort of thing pretty handy to have around for debugging. Are you sure it's not worth keeping?

packages/run-protocol/src/runStake/attestation.js
dckc: Hm. The "so that ..." goes with "attenuated". I think this change makes it less clear.

packages/run-protocol/src/econ-behaviors.js
turadg: Fair, I'm not sure. I wish we had some way to opt into logging output. I was very used to the selectors in https://www.npmjs.com/package/debug but I understand the SES prevents us from setting env vars. Just realizing now though that we can set the output config in code so at least there's a single place to modify.

Anyway, I'm now a weak yes on leaving commented out log output in code.

packages/run-protocol/src/runStake/attestation.js
turadg: Maybe I misunderstood the intent of the statement. Is the goal to provide a subset of ZCFMint powers? 

Related, an `@returns` may read better. (Though we might need to disable the eslint rule that requires every JSDoc param tag to have a type, so that it can be inferred by what's actually returned.)

packages/run-protocol/src/econ-behaviors.js
dckc: Agreed. In fact, `BldLienAtt` vs. `Attestation` is unmotivated complexity.

I changed `'BldLienAtt'` and `'Attestation'` to `KW.Attestation` throughout.

And `'RUN'` to `KW.Debt` while I'm at it.

packages/run-protocol/src/runStake/attestation.js
dckc: This design relies on `lienBridge` to validate addresses. As far as this layer is concerned, they're just strings.

packages/run-protocol/src/runStake/attestation.js
dckc: Why use destructuring assignment only half way?

packages/run-protocol/src/runStake/attestation.js
dckc: The odds that someone is going to recognize `m` and `s` as short for `minuend` and `subtrahend` seem remote. I was going to use `a` and `b` or `x` and `y`, but I thought `hi` and `lo` were a little more clear: even though `lo` could be higher than `hi`, that's the exceptional case that rounds to `empty`.

The term I found was _truncated subtraction_ ([source](https://math.stackexchange.com/a/328542)). But that introduces a line-wrap, which (to quote Dean) is evil. But `subtractOrEmpty` works with 1-letter args. OK. Done.

packages/run-protocol/src/runStake/attestation.js
dckc: There's a convention in `@agoric/store` to use `isFoo(x)` to return a boolean about whether `x` is a `Foo` and `assertFoo(x)` to throw if `x` is not. Both are implemented in terms of `checkFoo(x, checker)` where `checker(cond, detail)` is a `cond => cond` for `isFoo` and `assert` for `assertFoo`.

So I changed this to `assertAccountStateOK`.

packages/run-protocol/src/runStake/params.js
Chris-Hibbert: not an acronym.

packages/run-protocol/src/runStake/attestation.js
dckc: Sort of... the `lienBridge` could have a pointer back to the calling object and throw if it's already used. But it's a bit of whack-a-mole: we would have the same prohibition not to make two `lienBridge`s. I think the complexity adds more risk than it's worth, given that the callers are all in extremely privileged bootstrap code.

packages/run-protocol/src/runStake/attestation.js
dckc: The `makeXKit` convention is for making an `X` plus some related objects. Here we return exactly one object that has a `makeAttestation` method; that is: we return an attestation maker.

packages/run-protocol/src/runStake/params.js
dckc: Since we have `makeRunStakeParamManager` and friends, I'm demoting this to a local name (`PKey`).

The only import was a test, and the test should independently spell the keyword; the test should fail if we un-intentionally change the API.

packages/run-protocol/src/econ-behaviors.js
dckc: you too? I'm the only one who, since I needed it earlier, likes to keep it around in case I need it later?
My vague memory of working in the kernel says we used to keep them.

Oh. this one has no hope of working when it's un-commented. Never mind. gone.

packages/run-protocol/src/runStake/runStake.js
dckc: I hope not. I thought about moving it from `privateArgs` to terms, but it's too powerful: it has `setLien`.

packages/run-protocol/src/runStake/attestation.js
dckc: Well, they are facets of a (lien) attestation service; there's no particular information about being used in runStake here. They work just fine as their own contract; I only recently deleted the `start` function, mostly to avoid bundling time when testing.

I'm more inclined to clarify the separation as follows:

```js
  const creatorFacet = wrapCreatorFacet(
    Far('runStake creator', {
      provideAttestationMaker: att.creatorFacet.provideAttestationMaker,
      getLiened: att.creatorFacet.getLiened,
    }),
  );
```

packages/run-protocol/src/runStake/runStakeKit.js
dckc: Right; a boolean suffices. Done.

packages/run-protocol/src/runStake/runStakeKit.js
dckc: done.

packages/run-protocol/src/runStake/runStakeKit.js
dckc: no; pruned.

packages/run-protocol/src/runStake/runStakeKit.js
dckc: fixed by setting updater to undefined etc. as in vault.js

packages/run-protocol/src/runStake/runStakeKit.js
dckc: I beefed up `payoffRun` and added a test for paying off more than you owe.

dckc:
@Chris-Hibbert @turadg please take another look. I think I'm caught up on your comments so far.

packages/run-protocol/src/runStake/README.md
```diff
@@ -0,0 +1,62 @@
...
+  - walletBridge: with high privilege and constrained interaction
+  - attestation: a service within the runStake contract
+  - golang: the Cosmos SDK layer, including the `x/lien` module
+
+Before this interaction, an `ag123` account has been provisioned,
+which provides an `AttestationMaker` for that account to the wallet.
+
+Note that the account already has 4000 BLD liened before this interaction.
+
+```mermaid
```
turadg, line 17: 😍

packages/run-protocol/src/runStake/README.md
```diff
@@ -0,0 +1,62 @@
+# RUNstake design notes
+
+The RUNstake contract provides loans on the basis of
+staked assets that earn rewards.
+
+The following sequence diagram shows an interaction between runStake and:
+  - dapp: with rich interaction but low privilege
+  - walletBridge: with high privilege and constrained interaction
+  - attestation: a service within the runStake contract
+  - golang: the Cosmos SDK layer, including the `x/lien` module
```
turadg: should we just call this Cosmos?

packages/run-protocol/src/runStake/attestation.js
turadg: As far as the code reader is concerned, what kind of address is it? :)  I get that its runtime type is `string` and that it represents an "address". Is it a Cosmos account address? I don't think that all needs to go in the symbol names but consider,
```suggestion
   * @param {string} address Cosmos account
```

In the future we may want to make this an opaque type because it's not meant to do all the things a string can do (e.g. `concat`).

packages/run-protocol/src/runStake/attestation.js
Chris-Hibbert: > The "so that ..." goes with "attenuated". I think this change makes it less clear.

I suspect @turadg  jumped to the same conclusion I did. The best explanation I could come up with for the "an" was that it used to be "an attenuated ZCFMint" that somehow got switched. Since that doesn't seem to be the case, either "a ZCFMint attenuated so that ... " or "Attenuate a ZCFMint so that ..." would be clear.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+ * Callers are responsible to split any such payment before returning.
+ *
+ * @param {ZCF} zcf
+ * @param {Brand<'nat'>} stakeBrand brand of the staked assets
+ * @param {ERef<StakingAuthority>} lienBridge bridge to account state
+ */
+const makeAttestationKit = async (zcf, stakeBrand, lienBridge) => {
+  const { add, subtract, makeEmpty, isGTE, coerce } = AmountMath;
+  const empty = makeEmpty(stakeBrand);
+  const zcfMint = await zcf.makeZCFMint(KW.Attestation, AssetKind.COPY_BAG);
```
Chris-Hibbert: I'd prefer this was `attMint` or `lienMint` or even `nftMint`.  When I see `zcfMint` I think of a default fungible mint.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+  };
+
+  /** @param {ZCFSeat} seat */
+  const returnAttestation = async seat => {
+    assertProposalShape(seat, { give: { Attestation: null }, want: {} });
+
+    /** @type {Amount<'copyBag'>} */
+    const attestationAmount = seat.getAmountAllocated(KW.Attestation);
+    const { address, lienedAmount: amountReturned } =
+      unwrapLienedAmount(attestationAmount);
```
Chris-Hibbert: this returns the amount allocated to the seat rather than the amount in the proposal.  I think it should do the minimum of those two.

packages/run-protocol/src/runStake/attestation.js
Chris-Hibbert: I like that.

packages/run-protocol/src/runStake/runStake.js
Chris-Hibbert: With Electorate, we were able to separate invitation from invitationAmount. Since the clients of governed contracts very much want to know who can govern the contract, this was very important.  Similarly, the ContractGovernor inserts its own identity into the terms of the governed contract so clients can find and verify either from the other.

Can clients of a runStake contract be fooled into buying a pig in a poke if they can't identify the lienBridge? If clients would want to know, we could put the identity of the instance in the terms and ask zoe to verify the link between them.  Do we use a public or private facet of the lienBridge?

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -0,0 +1,168 @@
...
+
+const { details: X } = assert;
+
+const trace = makeTracer('RM'); // TODO: how to turn this off?
+
+/**
+ * @param {ZCF} zcf
+ * @param {ZCFMint<'nat'>} debtMint
+ * @param {{ debt: Brand<'nat'>, Attestation: Brand<'copyBag'>, Stake: Brand<'nat'> }} brands
+ * @param {*} paramManager
```
Chris-Hibbert: haven't I heard that declaring it `unknown` would let the types flow through?

packages/run-protocol/test/runStake/runStake-test-steps.js
```diff
@@ -0,0 +1,211 @@
+/**
+ * @typedef {[string, unknown] | [string, unknown, false]} Step
+ * @typedef {[string, string, Step[]]} TestCase
+ * @type {TestCase[]}
+ */
+export const CASES = [
```
Chris-Hibbert, line 6: This is nice. I'm going to try to emulate it at some point.

Chris-Hibbert (commented):
This is looking good.

I think I should make one more pass through it earlier in the day. I may have been flagging near the end.

packages/run-protocol/src/runStake/README.md
```diff
@@ -0,0 +1,62 @@
+# RUNstake design notes
```
turadg, line 1: tyvm

packages/run-protocol/src/runStake/attestation.js
turadg: good point!  it just took me a second to realize the double brackets were to get the first tuple. but it's a good thing for my brain to get faster at detecting.

tangentially, I'd really love to type type `fit` so that the annotation on line 80 wasn't necessary. Filed https://github.com/Agoric/agoric-sdk/issues/4917

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * @param { string } address
+   * @param { Amount<'nat'> } lienDelta
+   */
+  const mintAttestation = async (address, lienDelta) => {
+    assert.typeof(address, 'string');
+    // This account state check is primarily to provide useful diagnostics.
+    // Since things like the bonded amount can change out from under us,
+    // we shouldn't rely on it completely.
+    // TODO: consider performance implications vs. diagnostics.
```
turadg: How/when will this be done? Seems like a MN-1 blocker, in which case it needs a ticket with that release tag.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+    await assertAccountStateOK(address, lienDelta);
+
+    const current = getOrElse(amountByAddress, address, () => empty);
+    const target = add(current, lienDelta);
+
+    // Since our mint is not subject to normal supply/demand,
+    // minting here and failing in setLiened below does no harm.
```
turadg, line 150: no harm, okay, but why does this promise need to start before `setLiened`?

what would break this way?
```
    await E(lienBridge).setLiened(address, current, target);
    amountByAddress.set(address, target);

    return mintZCFMintPayment(
      zcf,
      zcfMint,
      wrapLienedAmount(address, lienDelta),
    );
```

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+  const returnAttestation = async seat => {
+    assertProposalShape(seat, { give: { Attestation: null }, want: {} });
+
+    /** @type {Amount<'copyBag'>} */
+    const attestationAmount = seat.getAmountAllocated(KW.Attestation);
+    const { address, lienedAmount: amountReturned } =
+      unwrapLienedAmount(attestationAmount);
+
+    const liened = amountByAddress.get(address); // or throw
+    const updated = subtract(liened, amountReturned); // or throw
```
turadg: took me a tick to realize "updated what" and that these were pre and post values. consider,
```suggestion
    const lienedPre = amountByAddress.get(address); // or throw
    const lienedPost = subtract(lienedPre, amountReturned); // or throw
```

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+
+    await E(lienBridge).setLiened(address, liened, updated);
+    // COMMIT. Burn the escrowed attestation payment and exit the seat
+    amountByAddress.set(address, updated);
+    zcfMint.burnLosses(harden({ Attestation: attestationAmount }), seat);
+    seat.exit();
+  };
+
+  /**
+   * Get liened amount synchronously.
```
turadg: why say "synchronously"?  isn't this authoritative?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+    return getOrElse(amountByAddress, address, () => empty);
+  };
+
+  const lienMint = harden({
+    mintAttestation,
+    returnAttestation,
+    getLiened,
+    /**
+     * @param { Amount<'copyBag'> } attAmt
+     * @throws { Error } if `attAmt` payload length is not 1
```
turadg: the type being thrown is not informative here. consider,
```suggestion
     * @throws if `attAmt` payload length is not 1
```

packages/run-protocol/src/runStake/attestation.js
turadg: Makes sense not to complicate the code. Is there a way with telemetry to see if this problem does occurr?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -0,0 +1,254 @@
...
+ * other side of the lienBridge. The mint created here
+ * is the authority on the liened amount. The liened amount
+ * on the other side of the bridge may lag. Do not create two
+ * mints that share a bridge.
+ *
+ * The provideAttestationMaker method on the returned creatorFacet
+ * provides the authorization for an address. The account holder
+ * can then use attMaker.makeAttestation(lienDelta).
+ *
+ * Note the Attestation keyword for use in returnAttestation offers.
```
turadg: is this comment for an earlier version? i'm not seeing any keywords in this function. 

I do see that `returnAttestation` itself has that keyword

packages/run-protocol/src/runStake/attestation.js
turadg: > The `makeXKit` convention is for making an `X`

Ah, there's always an X object in the kit, got it. 

> plus some related objects

Just want to confirm that Kit can includes related _functions_ as well. E.g. `makePromiseKit` returns `{ promise, reject, resolve }`.

> Here we return exactly one object that has a `makeAttestation` method

I understand that if it were `makeAttestionKit` then it would have to be `{ attestion, …relatedStuff }`

> that is: we return an attestation maker.

Isn't `makeAttestation` function the attestation maker? The object being returned here has `makeAttestion` and several related functions. Maybe `Kit` isn't the right term but there's something wrapping the actual attestation maker. 

By my read, this far object is the "attestation tools" and the current `makeAttestationTools` function is better understood as `makeAttestationFacets`. Though I like your idea below to return wrappers, in which case this would be `makeAttestionWrappers`.

Though I'd go further and adopt the pattern from https://github.com/Agoric/agoric-sdk/blob/master/packages/governance/src/contractHelper.js#L100-L106 and name this something like `handleAttestations`.

turadg (commented):
Managed an almost complete pass. Please re-request my review when this has latest changes from master.

packages/run-protocol/src/runStake/attestation.js
dckc: It's a Cosmos account iff the `lienBridge` is a bridge to Cosmos.

Actually, we do have an `Address` type. I'll use that and clarify it a bit.

packages/run-protocol/src/runStake/attestation.js
dckc: This is another decision that I inherited and so didn't feel obliged to do any better. But it's not standing up under scrutiny. So I'll reconsider.

packages/run-protocol/src/runStake/runStake.js
dckc: `lienBridge` isn't a contract, so I don't think we can use Zoe instances to identify it.

I think the reason to trust a `runStake` instance is that you found it via `agoricNames`.

packages/run-protocol/src/runStake/attestation.js
dckc: I'd like to do it during this PR review. IOU some discussion.

packages/run-protocol/src/runStake/attestation.js
dckc: I think by design nothing would break. I'm just uneasy about the amount of moving parts in a call to `mintZCFMintPayment` so I was avoiding moving them after the commit point.

packages/run-protocol/src/runStake/attestation.js
dckc: This JSDoc comment is meant to specify what clients would want to know about this contract. I figure keywords are part of that, even though they only get used in `returnAttestation`.

turadg: Agreed. Though if we use the bare string `Attestation` we're countering some of the benefits of it being defined in the code by `KW.Attestation` instead of simply `'Attestation'`. If the goal of defining the manifest constant is that we can easily change the underlying value, then I think this comment should use it. If not, then I think we should reconsider whether the manifest constants are paying for their way.

Separately, as a reader I still don't get what I'm supposed to note. Maybe "Note that `Attestation` is the keyword is in `returnAttestation` offers."

packages/run-protocol/src/runStake/attestation.js
dckc: > should the `make` function take the key as a parameter?

Interesting idea. I can see some utility. The design here is inspired by ELib, where [fetch/2](https://github.com/kpreid/e-on-java/blob/master/src/jsrc/org/erights/e/elib/tables/EMap.java#L307) took a [Thunk](https://github.com/kpreid/e-on-java/blob/master/src/jsrc/org/erights/e/elib/base/Thunk.java), which takes no args.

For reference, monte adopted the same design:
https://monte.readthedocs.io/en/latest/runtime.html#safeScope.Map.fetch
https://github.com/monte-language/typhon/blob/master/typhon/objects/collections/maps.py#L138

packages/run-protocol/src/runStake/attestation.js
dckc: How could they be different? Doesn't Zoe guarantee that the proposal is the same as what's allocated to the seat in an offer handler?

packages/run-protocol/src/runStake/runStakeManager.js
dckc: no, not correctly, anyway. To make types flow through, I think type parameters are necessary.

packages/run-protocol/src/runStake/attestation.js
dckc: we discussed it and went with diagnostics over performance for now. We added some comments.

packages/run-protocol/src/runStake/attestation.js
dckc: the `mintZCFMintPayment` call includes remote operations, which could fail, so we agreed this must go before the commit point.

packages/run-protocol/src/runStake/attestation.js
dckc: I did add the `getAccountState` that asserts that both golang and JS stores agree, but when I went to remove this `getLiened` API, I found it was used in lots of tests. So I kept it at this layer but removed it from the attestation tool (nee `attMaker`) and `runStake` creator facet APIs.

packages/run-protocol/src/runStake/attestation.js
dckc: I suppose `seat.getProposal()` is a more idiomatic way to get the amount. I went with that.

packages/run-protocol/src/runStake/types.js
```diff
@@ -27,5 +27,8 @@
  * @property {Amount<'nat'>} unbonding
  * @property {bigint} currentTime
  *
+ * Perormance note: getAccountState is expensive, as it has to iterate over all
```
turadg: s/Perormance/Performance

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -95,16 +112,13 @@ const makeAttestationIssuerKit = async (zcf, stakeBrand, lienBridge) => {
     );
   };
   const assertAccountStateOK = async (address, toLien) => {
-    const s = await E(lienBridge).getAccountState(address, stakeBrand);
+    const s = await getAccountState(address, stakeBrand);
     const unlocked = subtractOrZero(s.total, s.locked);
     checkDelta(toLien, 'unliened', s.total, s.liened);
     checkDelta(toLien, 'bonded and unliened', s.bonded, s.liened);
     checkDelta(toLien, 'unlocked and unliened', unlocked, s.liened);
   };
```
turadg: How expensive are these checks? If cheap then better to include in `getAccountState` so they're validated whenever it's accessed.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -153,12 +167,12 @@ const makeAttestationIssuerKit = async (zcf, stakeBrand, lienBridge) => {
...
     // COMMIT. Burn the escrowed attestation payment and exit the seat
     amountByAddress.set(address, lienedPost);
-    zcfMint.burnLosses(harden({ Attestation: attestationAmount }), seat);
+    attMint.burnLosses(harden({ Attestation: attestationAmount }), seat);
     seat.exit();
   };
 
   /**
-   * Get liened amount synchronously.
+   * Get liened amount from the JS store, without using the lienBridge.
```
turadg: please help reader understand the authority of this info. If the JS store and the lienBridge disagree about the amount, what should happen? (IIUC, that triggers death from confusion)

turadg (commented):
Looking good. Will have my approval when the checkboxes in the PR body are ticked or removed.

packages/run-protocol/src/runStake/attestation.js
dckc: Um... that's what I did, no?

packages/run-protocol/src/runStake/runStake.js
dckc: Ugh. I'm struggling to get this change right.

packages/run-protocol/src/runStake/attestation.js
dckc: ok, I'll add the info that this is mainly for testing to the code as well as this conversation.

turadg (approved):
Big one!

PR #4742, opened 2022-03-05
Implement multifaceted virtual objects

### LINKED ISSUE #3834: multiple facets for each virtual object

In the kernel meeting last week (8-sep-2021), @dtribble raised the idea of adding swingset support for multiple facets on virtual objects. A common case is for a single object to have both an "exchange" facet (allowing you to change ownership) and a "use" facet (allowing you to do something other-than / subordinate-to ownership). For example, you might put the exchange facet into an Offer (which gets escrowed by Zoe), but you can continue to exercise the use facet until Zoe transfers ownership upon contract completion.

This pattern can be implemented by contract code, but it might be easier and more efficient if swingset (and the virtual object manager) could help out. Basically we've got a single bundle of state, and multiple forms of behavior to wrap around it. Given an existing virtual object, it might be interesting to define a new object (with new behavior) that gets to modify the same original state. Both the old and new virtual objects (facets) would have ephemeral Representatives, and the VOM would know how to recreate either facet upon demand, so everything else about them could be removed from RAM when not in use.

It is also quite common for one facet to get revoked once the other is created: exercising the exchange facet (to transfer ownership) must revoke the previous use facet. So low-level support for revoking facets might help.

## Description of the Design

To approach this, `makeKind()` would probably need to return an extra object for use in a rights-amplification pattern. Later, if you invoke this extra object and give it a Representative (facet-1), you can create a related facet-2 Representative, which gets its own vref.

`makeKind()` might also return a third object for use in revocation. One open question is whether revocation should be implemented with an interposed revocable forwarder on the outside of the target object (so nobody can invoke its methods), or inside (so the revoked facet's behavior methods cannot reach the state), or both.

This is entangled with upgrade and #3062: one way to change/fix/improve the behavior of a class of objects would be to create new facets on the old state. 

## Security Considerations

The authority to create a new facet around old data is equivalent to that of modifying the old data, so it must be carefully managed. Merely having access to facet-1 is certainly not sufficient to create facet-2.

The ability to change the behavior of previously-defined facets (upgrade) needs to be clearly marked as well. The ability to create the class of objects in the first place is not sufficient: while the provider of the kind constructor can obviously make the objects/facets behave in any way they desire, that does not give them the ability to retroactively modify that behavior in ways that could not be deduced from a reading the original code. So out upgrade story needs to include some additional control object.

FUDCo:
These changes implement the multifaceted virtual objects API per our current design.

Closes #3834

### DIFF
--- a/packages/SwingSet/docs/virtual-objects.md
+++ b/packages/SwingSet/docs/virtual-objects.md
@@ -25,44 +25,42 @@ The `descriptionTag` parameter is a short description string for the kind.  This
 
 The `init` parameter is a function that will be called when new instances are first created. It is expected to return a simple JavaScript object that represents the initialized state for the new virtual object instance.  Any parameters passed to the maker function returned by `defineKind` are passed directly to the `init` function.
 
-The `actualize` parameter is a function that binds an in-memory instance (the "Representation) of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return a new Javascript object with methods that close over the given state.  This returned object will become the body of the new instance.  It is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.
+The `actualize` parameter is a function that binds an in-memory instance (the "Representative") of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return either:
+
+1. A new JavaScript object with methods that close over the given state.  This returned object will become the body of the new instance.  This object can be empty; in such a case it can serve as a powerless but unforgeable "marker" handle.
+
+2. A new JavaScript object populated with objects as described in (1).  These will become facets of the new instance.  The returned object will be an object mapping to the facets by name.
+
+The `actualize` function is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.  The `actualize` function for a given VDO kind must always return the same shape of result: if it returns a single instance body, it must always return a single instance body; if it returns a object full of facets, it must always return an object with the exact same facet names.
 
 The `finish` parameter will, if present (it is optional), be called exactly once as part of instance initialization.  It will be invoked immediately after the `actualize` function is called for the first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function and thus becomes available to whoever requested its creation.  `finish` is passed two parameters: the virtual object's state (exactly as passed to the `actualize` function) and the virtual object itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other virtual objects.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
 
 For example:
 
 ```javascript
-  function initCounter(name) {
-    return {
-      counter: 0,
-      name,
-    };
-  }
-  function actualizeCounter(state) {
-    return {
-      inc() {
-        state.counter += 1;
-      },
-      dec() {
-        state.counter -= 1;
-      },
-      reset() {
-        state.counter = 0;
-      },
-      rename(newName) {
-        state.name = newName;
-      },
-      getCount() {
-        return state.counter;
-      },
-      getName() {
-        return state.name;
-      },
-    };
-  }
-  function finishCounter(state, counter) {
+  const initCounter = (name) => ({ counter: 0, name });
+
+  const actualizeCounter = (state) => ({
+    inc: () => {
+      state.counter += 1;
+    },
+    dec: () => {
+      state.counter -= 1;
+    },
+    reset: () => {
+      state.counter = 0;
+    },
+    rename: (newName) => {
+      state.name = newName;
+    },
+    getCount: () => state.counter,
+    getName: () => state.name,
+  });
+
+  const finishCounter = (state, counter) => {
     addToCounterRegistry(counter, state.name);
-  }
+  };
+
   const makeCounter = defineKind('counter', initCounter, actualizeCounter, finishCounter);
 ```
 
@@ -80,11 +78,73 @@ This defines a simple counter object with two properties in its state, a count a
   console.log(`${barCounter.getName()} count is ${barCounter.getCount()`); // "new bar count is 1"
 ```
 
-Note that the `init`, `actualize`, and `finish` functions are defined explicitly here for clarity of exposition, but in practice you'd usually declare them inline as arrow functions in the parameters of the `defineKind` call.
+Suppose you instead wanted to provide the the increment and decrement capabilities as independent facets.  A simplified version of the above (without the name property, counter registry, and `reset` method) might look like:
+
+```javascript
+  const initFacetedCounter = () => ({ counter: 0 });
+
+  const actualizeFacetedCounter = (state) => {
+    const getCount = () => state.counter;
+    return {
+      incr: {
+        step: () => {
+          state.counter += 1;
+        },
+        getCount,
+      },
+      decr: {
+        step: () => {
+          state.counter -= 1;
+        },
+        getCount,
+      },
+    };
+  }
+
+  const makeFacetedCounter = defineKind('counter', initCounter, actualizeCounter);
+```
+
+Which you'd use like:
+
+```javascript
+  const { incr, decr } = makeFacetedCounter('foo');
+
+  incr.step();
+  incr.step();
+  console.log(`count is ${decr.getCount()`); // "count is 2"
+  decr.step();
+  console.log(`count is ${incr.getCount()`); // "count is 1"
+```
+
+Note that the `init`, `actualize`, and `finish` functions are defined explicitly in the above examples for clarity of exposition, but in practice you'd usually declare them inline in the parameters of the `defineKind` call:
+
+```javascript
+  const makeFacetedCounter = defineKind(
+    'counter',
+    () => ({ counter: 0 }),
+    (state) => {
+      const getCount = () => state.counter;
+      return {
+        incr: {
+          step: () => {
+            state.counter += 1;
+          },
+          getCount,
+        },
+        decr: {
+          step: () => {
+            state.counter -= 1;
+          },
+          getCount,
+        },
+      };
+    },
+  );
+```
 
 Additional important details:
 
-- The set of state properties is fully determined by the `init` function.  That is, the set of properties that exist on `state` is completely determined by the enumerable properties of the object that `init` returns.  State properties cannot thereafter be added or removed.
+- The set of state properties of an instance is fully determined by the `init` function.  That is, the set of properties that exist on in instance's `state` is completely determined by the enumerable properties of the object that `init` returns.  State properties cannot thereafter be added or removed.  Currently there is no requirement that all instances of a given kind have the same set of properties, but code authors should not rely on this as such enforcement may be added in the future.
 
 - The values a state property may take are limited to things that are serializable and which may be hardened (and, in fact, _are_ hardened and serialized the moment they are assigned).  That is, you can replace what value a state property _has_, but you cannot modify a state property's value in situ.  In other words, you can do things like:
 
--- a/packages/SwingSet/src/lib/parseVatSlots.js
+++ b/packages/SwingSet/src/lib/parseVatSlots.js
@@ -12,34 +12,100 @@ import { assert, details as X } from '@agoric/assert';
 // as badly.  Also, "slot" is a short, single syllable word, which is nice.
 // But still "slot" implies containership which I think is wrong.)
 
-// Object/promise references (in vats) contain a three-tuple of (type,
-// allocator flag, index). The 'ownership' flag is expressed as a sign: "-"
-// means the index was allocated by the kernel, and thus the actual object
-// lives in some other vat (and any messages sent to it must go into the
-// kernel). "+" means the index was allocated by this vat, hence the object
-// lives in this vat, and we can expect to receive messages for it.
-
 /**
- * Parse a vat slot reference string into a vat slot object:
+ * Parse a vref string into its component parts:
  *   {
  *      type: STRING, // 'object', 'device', 'promise'
  *      allocatedByVat: BOOL, // true=>allocated by vat, false=>by the kernel
- *      id: Nat
+ *      id: Nat,
+ *      subid: Nat,
+ *      baseRef: STRING,
+ *      facet: Nat
  *   }
  *
- * @param {string} s  The string to be parsed, as described above.
+ * A vref string can take one of the forms:
  *
- * @returns {*} a vat slot object corresponding to the parameter.
+ *   T-N
+ *   T+N
+ *   T+N/I
+ *   T+N/I:F
  *
- * @throws if the given string is syntactically incorrect.
+ * Where:
+ *
+ *   T is a single character encoding the type of entity being referenced: 'd'
+ *      for 'device', 'o' for 'object', or 'p' for 'promise'.  One of the string
+ *      values 'device', 'object', or 'promise' is returned as the `type`
+ *      property of the result.
+ *
+ *   '+' or '-' encodes who allocated the reference: '-' for the kernel
+ *      (typically an import) or '+ for the vat (typically an export).  This is
+ *      returned in the `allocatedByVat` property of the result as a boolean.
+ *
+ *   N is a decimal integer representing the identity of the referenced entity.
+ *      This is returned in the `id` property of the result as a BigInt.
+ *
+ *   I if present (only allowed if T is 'o') is a decimal integer representing
+ *      the instance id of the referenced object.  In this case N denotes a
+ *      category of objects that share a common shape, either one of the store
+ *      types or a virtual object kind, and I indicates which instance of that
+ *      category is being referred to.  If present this is returned as the
+ *      `subid` property of the result as a BigInt.
+ *
+ *   F if present (only allowed if I is also present) is a decimal integer
+ *      referencing a facet of the referenced object.  In this case N/I denotes
+ *      a particular object instance and F indicates which of several possible
+ *      facets of that instance is being addressed.  If present this is returned
+ *      in the `facet` property of the result as a BigInt.
+ *
+ * The `baseRef` property of the result is `vref` stripped of any facet indicator.
+ *
+ * A "vref" identifies an entity visible to vat code to which messages may be
+ * sent and which may be compared for equality to other such entities.  Let's
+ * call such an entity an "addressable object".
+ *
+ * A "baseRef" designates an entity that is managed by LiveSlots, both as a unit
+ * of garbage collection specifically and as a unit of memory management more
+ * generally.  Such an entity may be a promise or remotable object or imported
+ * presence, all of which will always be JavaScript objects in memory, or it may
+ * be a virtual object or collection, which can be in memory or on disk or both.
+ * Let's call such an entity a "base object".  In most cases this is one and the
+ * same with the addressable object that the vref designates, but in the case of
+ * a faceted object it is the faceted object as a whole (represented in memory,
+ * though not on disk, as the cohort array) rather than any particular
+ * individual facet (the faceted object per se is never exposed directly to code
+ * running within the vat; only its facets are).
+ *
+ * XXX TODO: The previous comment suggests some renaming is warranted:
+ *
+ * `slotToVal` maps a baseRef to a base object (actually to a weakRef that
+ *    points to a base object)
+ * `getValForSlot` maps a baseRef to a base object, or to undefined if it is not
+ *    resident in memory
+ * `convertSlotToVal` maps a vref to to an addressable object, loading it from
+ *    disk if necessary
+ *
+ * `valToSlot` maps an addressable object to a vref
+ * `getSlotForVal` maps an addressable object to a vref
+ * `convertValToSlot` maps an addressable object to a vref, generating a new
+ *    vref if necessary
+ *
+ * @param {string} vref  The string to be parsed, as described above.
+ *
+ * @returns {*} a vref components descriptor corresponding to the vref string
+ *    parameter, assuming it is syntactically well formed.
+ *
+ * @throws if the given vref string is syntactically incorrect.
  */
-export function parseVatSlot(s) {
-  assert.typeof(s, 'string');
+export function parseVatSlot(vref) {
+  assert.typeof(vref, 'string');
+  const parts = vref.split(':');
+  assert(parts.length === 1 || parts.length === 2, X`invalid vref ${vref}`);
+  const [baseRef, facetStr] = parts;
   let type;
   let allocatedByVat;
-  const typechar = s[0];
-  const allocchar = s[1];
-  const idSuffix = s.slice(2);
+  const typechar = baseRef[0];
+  const allocchar = baseRef[1];
+  const idSuffix = baseRef.slice(2);
 
   if (typechar === 'o') {
     type = 'object';
@@ -48,7 +114,7 @@ export function parseVatSlot(s) {
   } else if (typechar === 'p') {
     type = 'promise';
   } else {
-    assert.fail(X`invalid vatSlot ${s}`);
+    assert.fail(X`invalid vref ${vref}`);
   }
 
   if (allocchar === '+') {
@@ -56,23 +122,27 @@ export function parseVatSlot(s) {
   } else if (allocchar === '-') {
     allocatedByVat = false;
   } else {
-    assert.fail(X`invalid vatSlot ${s}`);
+    assert.fail(X`invalid vref ${vref}`);
   }
 
   const delim = idSuffix.indexOf('/');
   let id;
   let subid;
+  let facet;
   let virtual = false;
   if (delim > 0) {
-    assert(type === 'object' && allocatedByVat, X`invalid vatSlot ${s}`);
+    assert(type === 'object' && allocatedByVat, X`invalid vref ${vref}`);
     virtual = true;
     id = Nat(BigInt(idSuffix.substr(0, delim)));
     subid = Nat(BigInt(idSuffix.slice(delim + 1)));
   } else {
     id = Nat(BigInt(idSuffix));
   }
+  if (subid !== undefined && facetStr !== undefined) {
+    facet = Nat(BigInt(facetStr));
+  }
 
-  return { type, allocatedByVat, virtual, id, subid };
+  return { type, allocatedByVat, virtual, id, subid, baseRef, facet };
 }
 
 /**
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -101,12 +101,12 @@ function build(
    */
 
   const valToSlot = new WeakMap(); // object -> vref
-  const slotToVal = new Map(); // vref -> WeakRef(object)
+  const slotToVal = new Map(); // baseRef -> WeakRef(object)
   const exportedRemotables = new Set(); // objects
   const kernelRecognizableRemotables = new Set(); // vrefs
   const pendingPromises = new Set(); // Promises
   const importedDevices = new Set(); // device nodes
-  const possiblyDeadSet = new Set(); // vrefs that need to be checked for being dead
+  const possiblyDeadSet = new Set(); // baseRefs that need to be checked for being dead
   const possiblyRetiredSet = new Set(); // vrefs that might need to be rechecked for being retired
 
   function retainExportedVref(vref) {
@@ -157,41 +157,41 @@ function build(
 
     Each state thus has a set of perhaps-measurable properties:
 
-    * UNKNOWN: slotToVal[vref] is missing, vref not in deadSet
+    * UNKNOWN: slotToVal[baseRef] is missing, baseRef not in deadSet
     * REACHABLE: slotToVal has live weakref, userspace can reach
     * UNREACHABLE: slotToVal has live weakref, userspace cannot reach
-    * COLLECTED: slotToVal[vref] has dead weakref
-    * FINALIZED: slotToVal[vref] is missing, vref is in deadSet
+    * COLLECTED: slotToVal[baseRef] has dead weakref
+    * FINALIZED: slotToVal[baseRef] is missing, baseRef is in deadSet
 
     Our finalizer callback is queued by the engine's transition from
-    UNREACHABLE to COLLECTED, but the vref might be re-introduced before the
+    UNREACHABLE to COLLECTED, but the baseRef might be re-introduced before the
     callback has a chance to run. There might even be multiple copies of the
     finalizer callback queued. So the callback must deduce the current state
     and only perform cleanup (i.e. delete the slotToVal entry and add the
-    vref to the deadSet) in the COLLECTED state.
+    baseRef to the deadSet) in the COLLECTED state.
 
   */
 
-  function finalizeDroppedObject(vref) {
+  function finalizeDroppedObject(baseRef) {
     // TODO: Ideally this function should assert that it is not metered.  This
     // appears to be fine in practice, but it breaks a number of unit tests in
     // ways that are not obvious how to fix.
     // meterControl.assertNotMetered();
-    const wr = slotToVal.get(vref);
+    const wr = slotToVal.get(baseRef);
     // The finalizer for a given Presence might run in any state:
     // * COLLECTED: most common. Action: move to FINALIZED
     // * REACHABLE/UNREACHABLE: after re-introduction. Action: ignore
     // * FINALIZED: after re-introduction and subsequent finalizer invocation
-    //   (second finalizer executed for the same vref). Action: be idempotent
+    //   (second finalizer executed for the same baseRef). Action: be idempotent
     // * UNKNOWN: after re-introduction, multiple finalizer invocation,
-    //   and post-crank cleanup does dropImports and deletes vref from
+    //   and post-crank cleanup does dropImports and deletes baseRef from
     //   deadSet. Action: ignore
 
     if (wr && !wr.deref()) {
       // we're in the COLLECTED state, or FINALIZED after a re-introduction
       // eslint-disable-next-line no-use-before-define
-      addToPossiblyDeadSet(vref);
-      slotToVal.delete(vref);
+      addToPossiblyDeadSet(baseRef);
+      slotToVal.delete(baseRef);
     }
   }
   const vreffedObjectRegistry = new FinalizationRegistry(finalizeDroppedObject);
@@ -218,10 +218,10 @@ function build(
       // manifestation *right now* (i.e. the non-resurrected ones), for which
       // we must check the remaining pillars.
       const deadSet = new Set();
-      for (const vref of possiblyDeadSet) {
+      for (const baseRef of possiblyDeadSet) {
         // eslint-disable-next-line no-use-before-define
-        if (!slotToVal.has(vref)) {
-          deadSet.add(vref);
+        if (!slotToVal.has(baseRef)) {
+          deadSet.add(baseRef);
         }
       }
       possiblyDeadSet.clear();
@@ -239,33 +239,35 @@ function build(
       }
       possiblyRetiredSet.clear();
 
-      const deadVrefs = Array.from(deadSet);
-      deadVrefs.sort();
-      for (const vref of deadVrefs) {
-        const { virtual, allocatedByVat, type } = parseVatSlot(vref);
+      const deadBaseRefs = Array.from(deadSet);
+      deadBaseRefs.sort();
+      for (const baseRef of deadBaseRefs) {
+        const { virtual, allocatedByVat, type } = parseVatSlot(baseRef);
         assert(type === 'object', `unprepared to track ${type}`);
         if (virtual) {
           // Representative: send nothing, but perform refcount checking
           // eslint-disable-next-line no-use-before-define
-          const [gcAgain, doRetire] = vrm.possibleVirtualObjectDeath(vref);
-          if (doRetire) {
-            exportsToRetire.push(vref);
+          const [gcAgain, retirees] = vrm.possibleVirtualObjectDeath(baseRef);
+          if (retirees) {
+            retirees.map(retiree => exportsToRetire.push(retiree));
           }
           doMore = doMore || gcAgain;
         } else if (allocatedByVat) {
           // Remotable: send retireExport
-          if (kernelRecognizableRemotables.has(vref)) {
-            kernelRecognizableRemotables.delete(vref);
-            exportsToRetire.push(vref);
+          // for remotables, vref === baseRef
+          if (kernelRecognizableRemotables.has(baseRef)) {
+            kernelRecognizableRemotables.delete(baseRef);
+            exportsToRetire.push(baseRef);
           }
         } else {
           // Presence: send dropImport unless reachable by VOM
           // eslint-disable-next-line no-lonely-if, no-use-before-define
-          if (!vrm.isPresenceReachable(vref)) {
-            importsToDrop.push(vref);
+          if (!vrm.isPresenceReachable(baseRef)) {
+            importsToDrop.push(baseRef);
             // eslint-disable-next-line no-use-before-define
-            if (!vrm.isVrefRecognizable(vref)) {
-              importsToRetire.push(vref);
+            if (!vrm.isVrefRecognizable(baseRef)) {
+              // for presences, baseRef === vref
+              importsToRetire.push(baseRef);
             }
           }
         }
@@ -491,21 +493,21 @@ function build(
     return valToSlot.get(val);
   }
 
-  function getValForSlot(slot) {
+  function getValForSlot(baseRef) {
     meterControl.assertNotMetered();
-    const wr = slotToVal.get(slot);
+    const wr = slotToVal.get(baseRef);
     return wr && wr.deref();
   }
 
-  function requiredValForSlot(slot) {
-    const wr = slotToVal.get(slot);
+  function requiredValForSlot(baseRef) {
+    const wr = slotToVal.get(baseRef);
     const result = wr && wr.deref();
-    assert(result, X`no value for ${slot}`);
+    assert(result, X`no value for ${baseRef}`);
     return result;
   }
 
-  function addToPossiblyDeadSet(vref) {
-    possiblyDeadSet.add(vref);
+  function addToPossiblyDeadSet(baseRef) {
+    possiblyDeadSet.add(baseRef);
   }
 
   function addToPossiblyRetiredSet(vref) {
@@ -574,15 +576,15 @@ function build(
         assert.equal(passStyleOf(val), 'remotable');
         slot = exportPassByPresence();
       }
-      const { type } = parseVatSlot(slot); // also used as assertion
+      const { type, baseRef } = parseVatSlot(slot); // also used as assertion
       valToSlot.set(val, slot);
-      slotToVal.set(slot, new WeakRef(val));
+      slotToVal.set(baseRef, new WeakRef(val));
       if (type === 'object') {
         // Set.delete() metering seems unaffected by presence/absence, but it
         // doesn't matter anyway because deadSet.add only happens when
         // finializers run, and we wrote xsnap.c to ensure they only run
         // deterministically (during gcAndFinalize)
-        vreffedObjectRegistry.register(val, slot, val);
+        vreffedObjectRegistry.register(val, baseRef, val);
       }
     }
     return valToSlot.get(val);
@@ -598,13 +600,23 @@ function build(
     return result;
   }
 
-  function registerValue(slot, val) {
-    const { type } = parseVatSlot(slot);
-    slotToVal.set(slot, new WeakRef(val));
-    valToSlot.set(val, slot);
+  function registerValue(baseRef, val, valIsCohort) {
+    const { type, facet } = parseVatSlot(baseRef);
+    assert(
+      !facet,
+      `registerValue(${baseRef} should not receive individual facets`,
+    );
+    slotToVal.set(baseRef, new WeakRef(val));
+    if (valIsCohort) {
+      for (let i = 0; i < val.length; i += 1) {
+        valToSlot.set(val[i], `${baseRef}:${i}`);
+      }
+    } else {
+      valToSlot.set(val, baseRef);
+    }
     // we don't dropImports on promises, to avoid interaction with retire
     if (type === 'object') {
-      vreffedObjectRegistry.register(val, slot, val);
+      vreffedObjectRegistry.register(val, baseRef, val);
     }
   }
 
@@ -616,8 +628,9 @@ function build(
   // m.unserialize) must be wrapped by unmetered().
   function convertSlotToVal(slot, iface = undefined) {
     meterControl.assertNotMetered();
-    const { type, allocatedByVat, virtual } = parseVatSlot(slot);
-    let val = getValForSlot(slot);
+    const { type, allocatedByVat, virtual, facet, baseRef } =
+      parseVatSlot(slot);
+    let val = getValForSlot(baseRef);
     if (val) {
       if (virtual) {
         // If it's a virtual object for which we already have a representative,
@@ -626,16 +639,23 @@ function build(
         // code to make a new representative anyway (which we'll discard) so
         // that as far as the user code is concerned we are making a new
         // representative with each act of deserialization.  This way they can't
-        // detect reanimation by playing games inside their instanceKitMaker to
+        // detect reanimation by playing games inside their kind definition to
         // try to observe when new representatives are created (e.g., by
         // counting calls or squirreling things away in hidden WeakMaps).
-        vrm.reanimate(slot, true); // N.b.: throwing away the result
+        vrm.reanimate(baseRef, true); // N.b.: throwing away the result
+        if (facet !== undefined) {
+          return val[facet];
+        }
       }
       return val;
     }
+    let result;
     if (virtual) {
       assert.equal(type, 'object');
-      val = vrm.reanimate(slot, false);
+      val = vrm.reanimate(baseRef, false);
+      if (facet !== undefined) {
+        result = val[facet];
+      }
     } else {
       assert(!allocatedByVat, X`I don't remember allocating ${slot}`);
       if (type === 'object') {
@@ -665,8 +685,11 @@ function build(
         assert.fail(X`unrecognized slot type '${type}'`);
       }
     }
-    registerValue(slot, val);
-    return val;
+    registerValue(baseRef, val, facet !== undefined);
+    if (!result) {
+      result = val;
+    }
+    return result;
   }
 
   function resolutionCollector() {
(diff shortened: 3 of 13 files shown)


FUDCo:
Probably worth calling out for the record one small change to the code as implemented from the implementation design in #3834: this implementation uses a different separator character (in this case, comma) between the main portion of the vref and the facet identifier than the separator character (slash) between the base kind vref and the instance identifier.  The design proposed the new multi-facet virtual objects would get `o+NN/AA/BB`, where `NN` is the kind ID, `AA` is the instance (cohort) ID, and `BB` is the facet ID.  I found it highly advantageous to use `o+NN/AA,BB` instead, as there are numerous cases where I needed to strip off the facetID while leaving the rest intact, and having a separate separator meant not having to count separators.  While I don't think this bit of minor pragmatism is likely to be terribly controversial, it warrants mention.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -80,7 +86,74 @@ This defines a simple counter object with two properties in its state, a count a
...
+        },
+        getCount,
+      },
+      decrementer: {
+        dec() {
+          state.counter -= 1;
+        }
+      },
+    };
+  }
```
erights:
```suggestion
  const actualizeFacetedCounter = state => {
    const getCount = () => state.counter;
    return {
      incrementer: {
        inc: () => {
          state.counter += 1;
        },
        getCount,
      },
      decrementer: {
        dec: () => {
          state.counter -= 1;
        },
      },
    };
  }
```
or
```suggestion
  const actualizeFacetedCounter = state => ({
      incrementer: {
        inc: () => {
          state.counter += 1;
        },
        getCount: () => state.counter,
      },
      decrementer: {
        dec: () => {
          state.counter -= 1;
        },
      },
    });
```

erights: Please start using arrow functions for all new code everywhere, including in this PR. This is even more important for documentation than it is for internal code. Thanks.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -80,7 +86,74 @@ This defines a simple counter object with two properties in its state, a count a
...
 
-Note that the `init`, `actualize`, and `finish` functions are defined explicitly here for clarity of exposition, but in practice you'd usually declare them inline as arrow functions in the parameters of the `defineKind` call.
+Suppose you instead wanted to provide the the increment and decrement capabilities as independent facets.  A simplified version of the above (without the name property and `reset` method) might look like:
+
+```javascript
+  function initFacetedCounter() {
+    return {
+      counter: 0,
+    };
+  }
```
erights:
```suggestion
  const initFacetedCounter = () => {
    return {
      counter: 0,
    };
  }
```
or better
```suggestion
  const initFacetedCounter = () => ({
      counter: 0,
    });
```

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -80,7 +86,74 @@ This defines a simple counter object with two properties in its state, a count a
...
+Which you'd use like:
+
+```javascript
+  const { inc, dec } = makeFacetedCounter('foo');
+
+  incrementer.inc();
+  incrementer.inc();
+  console.log(`count is ${decrementer.getCount()`); // "count is 2"
+  decrementer.dec();
+  console.log(`count is ${incrementer.getCount()`); // "count is 1"
```
erights: This code only works if the facets are named `inc` and `dec`. By themselves, they're not clear enough. But `increment` and `decrement` are too long to be pleasant to use. I always use `incr` and `decr`.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -352,6 +357,54 @@ export function makeVirtualObjectManager(
     configurable: true,
   });
 
+  /**
+   * Assess the facetiousness of a value.  If the value is an object containing
```
erights, line 362: OMG facetiousness. I should have seen that coming ;)

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -352,6 +357,54 @@ export function makeVirtualObjectManager(
...
+   * @returns {Facetiousness} an assessment of the facetiousness of `obj`
+   */
+  function assessFacetiousness(obj) {
+    if (typeof obj !== 'object') {
+      return 'not';
+    }
+    if (Object.getOwnPropertySymbols(obj).length !== 0) {
+      return 'not';
+    }
+    // XXX should there be some kind of prototype check here for further paranoia?
```
erights: Once you've determined that it should be a far object, if it is valid at all, then let the `Far` constructor take care of all further validation. This will change with far classes, and I'd like to keep it in one place.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -352,6 +357,54 @@ export function makeVirtualObjectManager(
...
+        return 'not';
+      }
+      if (!established) {
+        established = current;
+      } else if (established !== current) {
+        return 'not';
+      }
+    }
+    if (!established) {
+      return 'one';
```
erights, line 406: So the empty object is interpreted as a single far object. Good. (No change suggested)

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +540,40 @@ export function makeVirtualObjectManager(
...
+        case 'one': {
+          representative = Far(tag, self);
+          vrm.checkOrAcquireFacetNames(kindID, null);
+          exposedRepresentative = representative;
+          break;
+        }
+        case 'many': {
+          exposedRepresentative = {};
+          representative = [];
+          const facetNames = Object.getOwnPropertyNames(self).sort();
```
erights, line 561: Since you've already validated that these are only strings, a simple `.sort()` works. Good. No change suggested.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +540,40 @@ export function makeVirtualObjectManager(
...
+      switch (facetiousness) {
+        case 'one': {
+          representative = Far(tag, self);
+          vrm.checkOrAcquireFacetNames(kindID, null);
+          exposedRepresentative = representative;
+          break;
+        }
+        case 'many': {
+          exposedRepresentative = {};
+          representative = [];
```
erights: I'm surprised that `representative` is an array. Given the contrasting names, may I assume this array is never exposed?

On the exposed representative, curious why you don't use `self` itself as you do for the single facet case above. It doesn't matter so no objection. But if it's otherwise indifferent, it would reduce the cost of making the multi-facetted representative.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +540,40 @@ export function makeVirtualObjectManager(
...
+            exposedRepresentative[facetName] = facet;
+            representative.push(facet);
+          }
+          harden(exposedRepresentative);
+          break;
+        }
+        case 'not':
+          assert.fail(`invalid self actualization for ${tag}`);
+        default:
+          assert.fail(`this can't happen`);
```
erights:
```suggestion
          assert.fail(X`internal: unexpected facetiousness: ${q(facetiousness)}`);
```
We do need better handling in general for cannot-happen internal errors. In liveslots, it should kill the vat. In the kernel, it should panic. However we haven't made that convenient enough to start practicing, so no change suggested. For now, I sometimes remember to begin the error message with "internal: " to assure the user it isn't their fault and to encourage them to report it.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +540,40 @@ export function makeVirtualObjectManager(
...
+          for (const facetName of facetNames) {
+            const facet = Far(`${tag} ${facetName}`, self[facetName]);
+            exposedRepresentative[facetName] = facet;
+            representative.push(facet);
+          }
+          harden(exposedRepresentative);
+          break;
+        }
+        case 'not':
+          assert.fail(`invalid self actualization for ${tag}`);
```
erights:
```suggestion
          assert.fail(X`invalid self actualization for ${q(tag)}`);
```
This is a user provided string, so better to quote it into the error message so that it's harder for it to create a misleading one.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -160,6 +165,46 @@ export function makeVirtualReferenceManager(
...
+   * @param {T[]} a2
+   * @returns {boolean}
+   */
+  function arrayEquals(a1, a2) {
+    assert(Array.isArray(a1));
+    assert(Array.isArray(a2));
+    if (a1.length !== a2.length) {
+      return false;
+    }
+    return a1.every((elem, idx) => a2[idx] === elem);
```
erights: I see that you're only using it for an array of strings, for which it is correct, and you're not exporting it, so no immediate dangers of this being used for anything else. But you're typing it as if it could be used for non-strings, in which case it should be correct for non-strings. Your code will fail on `arrayEquals([NaN], [NaN])`.
```suggestion
    return a1.every((elem, idx) => Object.is(a2[idx], elem));
```

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -160,6 +165,46 @@ export function makeVirtualReferenceManager(
...
 
+  /**
+   * Compare two arrays (shallowly) for equality.
+   *
+   * @template T
+   * @param {T[]} a1
+   * @param {T[]} a2
+   * @returns {boolean}
+   */
+  function arrayEquals(a1, a2) {
```
erights: Hey, it is a new function.

```suggestion
  const arrayEquals = (a1, a2) => {
```

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -160,6 +165,46 @@ export function makeVirtualReferenceManager(
...
+    if (a1.length !== a2.length) {
+      return false;
+    }
+    return a1.every((elem, idx) => a2[idx] === elem);
+  }
+
+  /**
+   * Check a list of facet names against what's already been established for a
+   * kind.  If they don't match, it's an error.  If nothing has been established
+   * yet, establish it now.
```
erights, line 241: Do you already have something like this for the field names from an `init`, to ensure they are uniform across a kind? Or do you allow these to differ per instance?

erights (dismissed):
I reviewed what I understood. But even after it looks good to me, you should get @warner's approval before merging. There too much I simply cannot evaluate.

packages/SwingSet/docs/virtual-objects.md
FUDCo: I was actually thinking the opposite way -- we have our weird coding style that we use for reasons of our own, but to explain a strange API to outsiders I thought it made sense not to multiply the number of unfamiliar elements they have to absorb at once.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: Correct, this array is never exposed.  We use an array to make vref lookup more direct; otherwise we'd also have to store a name table.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: That's actually an interesting question.  We really do have to enforce the invariance of the set of facets so that the object's API can't be mutated into different shapes depending on when it's deserialized.  However, the virtual object manager allows different instances to be initialized with different field names, and since any given instance is only initialized once it'll work fine so long as the relevant methods are expecting a variant structure.  However, we've never contemplated making any use of this "feature" and it's arguably a bad coding practice.  Enforcing field name uniformity might be a good idea (certainly it would be easy to do), but we haven't made a decision and we should.

packages/SwingSet/docs/virtual-objects.md
FUDCo: One bit of ugliness with this example is confusion between facet names vs. method names.  I'm thinking that maybe the facet name should denote the purpose of the facet (so let's go with `incr` and `decr`, since you seem to like those), and then have a common method name for bumping the counter (say, `step`) that both use.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -25,44 +25,42 @@ The `descriptionTag` parameter is a short description string for the kind.  This
...
 
-The `actualize` parameter is a function that binds an in-memory instance (the "Representation) of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return a new Javascript object with methods that close over the given state.  This returned object will become the body of the new instance.  It is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.
+The `actualize` parameter is a function that binds an in-memory instance (the "Representative") of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return either:
+
+1. A new JavaScript object with methods that close over the given state.  This returned object will become the body of the new instance.
```
warner: Let's mention "this object can be empty, to serve as a powerless "marker" handle".

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -25,44 +25,42 @@ The `descriptionTag` parameter is a short description string for the kind.  This
...
-      },
-      getName() {
-        return state.name;
-      },
-    };
-  }
-  function finishCounter(state, counter) {
+  const initCounter (name) => ({ counter: 0, name });
+
+  const actualizeCounter (state) => ({
```
warner: missing `=`

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -80,11 +78,73 @@ This defines a simple counter object with two properties in its state, a count a
...
+        },
+      };
+    },
+  );
+```
 
 Additional important details:
 
-- The set of state properties is fully determined by the `init` function.  That is, the set of properties that exist on `state` is completely determined by the enumerable properties of the object that `init` returns.  State properties cannot thereafter be added or removed.
+- The set of state properties of an instance is fully determined by the `init` function.  That is, the set of properties that exist on `state` is completely determined by the enumerable properties of the object that `init` returns.  State properties cannot thereafter be added or removed.
```
warner: Should we mention (maybe with a TODO) that this set might vary between instances? Or that `init()` "SHOULD" return the same set of properties for all instances, even though maybe we don't catch this yet, and that a future implementation will be strict about enforcing it?

packages/SwingSet/src/lib/parseVatSlots.js
```diff
@@ -35,11 +37,14 @@ import { assert, details as X } from '@agoric/assert';
  */
 export function parseVatSlot(s) {
   assert.typeof(s, 'string');
+  const parts = s.split(',');
```
warner: Hm, if we're adding another delimiter here, I'd bikeshed towards `.` instead of `,`, because then facets are like `.a`/`.b` `.1/.2` suffixes that I've seen on other clusters of related identifiers (maybe I'm thinking of the grouping of 802.3 and 802.11 -type standards). Or nested numbered bullet-points in word-processor documents.

warner: Also, our format is now big enough that we should add a comment to the top here with a list of examples. `o+1`, `o+2/1` and `o+2/2`, `o+3/1.0` and `o+3/1.1` and `o+3/2.0` and `o+3/2.1`.

Also, would it make anything cleaner if we declared that *all* virtual objects might be multi-facet, but the ones that only define a single facet get `.0`? I think that might remove a category of vref formats in exchange for adding an extra `.0` to all VO vrefs (or maybe the `o+NN/MM` case would still be needed by virtual collections? Or maybe those could use `o+NN/MM.0` too?)

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -600,10 +600,16 @@ function build(
     return result;
   }
 
-  function registerValue(slot, val) {
+  function registerValue(slot, val, faceted) {
```
warner: Could we make a separate function for this, like `registerFacetedValue`? I'd like to put a type definition on these (eventually), and making `val` either be `Remotable` or `Array<Remotable>` would be kinda gross.

OTOH, we're already going to have that weird type for `slotToVal`.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -352,6 +357,53 @@ export function makeVirtualObjectManager(
...
+   * facets and 'many' is returned.  In all other cases `obj` does not represent
+   * any kind of facet abstraction and 'not' is returned.
+   *
+   * @typedef {'one'|'many'|'not'} Facetiousness
+   *
+   * @param {*} obj  The (alleged) object to be assessed
+   *
+   * @returns {Facetiousness} an assessment of the facetiousness of `obj`
+   */
+  function assessFacetiousness(obj) {
```
warner: Please add a few unit tests for this function to show that it handles a variety of different kinds of inputs.

What happens if it's given a cycle? "Don't do that" might be a reasonable answer, but if it's convenient to have it return `'not'` or throw an error instead of going into an infinite loop, that'd be nice. I don't know that I'd want to spend a wrapper function and WeakSet to acheive that, though.

Should we be worried about malice if `obj` is not frozen and changes its properties as you look at it? @erights I know `Far` is not idempotent, but would it be a good idea to do one of these three before calling `assertFacetiousness()`:?
* require the user's actualize function to `harden()` its return value
* have `defineKind` do `harden()` on the actualize return value
* have `defineKind` to `Object.freeze()` on the actualize return value

The first feels like a burden on userspace. The last feels halfhearted.

Is there any way that `passStyleOf` could take over some of this work? I'm guessing no, because of the `Far` registration thing, but it feels like there's a lot of overlap between the kind of analysis this is doing, with the work that `marshal` does.

packages/SwingSet/src/liveslots/virtualObjectManager.js
warner: The fact that this line is what notices empty objects is worth a brief `// empty objects are Far, not 0 facets`

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +539,40 @@ export function makeVirtualObjectManager(
...
+          exposedRepresentative = representative;
+          break;
+        }
+        case 'many': {
+          exposedRepresentative = {};
+          representative = [];
+          const facetNames = Object.getOwnPropertyNames(self).sort();
+          vrm.checkOrAcquireFacetNames(kindID, facetNames);
+          for (const facetName of facetNames) {
+            const facet = Far(`${tag} ${facetName}`, self[facetName]);
```
warner, line 568: @erights Any opinions on the synthesized `iface` names here? Should we join with a period instead of space to mirror the way you might dereference facets out of the return value of `makeFoo()`? (`Purse.depositFacet` and `Purse.purse` maybe?)

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +539,40 @@ export function makeVirtualObjectManager(
       if (initializing) {
         cache.remember(innerSelf);
       }
-      const representative = Far(tag, actualize(wrappedData));
+      const self = actualize(wrappedData);
+      let representative;
+      let exposedRepresentative;
```
warner: The aspecific arity of these two variables bugs me, having a (singular) "representative" name used to sometimes hold an Array feels confusing. I kind of want to name it "storage value" and "return value" or something, but that's not quite right either. My suggestion about storing non-faceted VOs using facet-id `0` overlaps with this: if we did something like that, then this code would always use `storedRepresentatives` (but `facetiousness === 'one'` would mean `storedRepresentatives.length === 1`), and instead of `exposedRepresentative` we could have like `returnedRepresentation` that was either an array of Representatives or a single Representative.

If we don't go that way, maybe just renamed these to `storedRepresentation` and `returnedRepresentation`? Or something with "cohort" in the name, maybe.

warner: Maybe just `stored` and `returned`. The only difference is that one is an array (or singleton) and the other is a record (or singleton), right? And the order matters.. what if this loop builds up an array of `[name, representative]` pairs, then the caller can switch on `.length` and then `.map()` it into the stored cohort or `Object.something` it into the returned form..

hm, not sure I can think of an approach that would make me happy :)

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -113,9 +118,9 @@ export function makeVirtualReferenceManager(
       // we must ensure addToPossiblyDeadSet only happens once).
       case 'recognizable': {
         syscall.vatstoreSet(key, '0');
-        const refCount = getRefCount(vobjID);
+        const refCount = getRefCount(slot);
```
warner: This is probably correct, but the types are now muddled. `getRefCount()` takes `vobjID` and parses out `slot` before touching vatstore, but in this case `setExportStatus` is doing the same thing, so it's kinda being transformed twice, and that smells wrong.

I'm trying to think of better names for the arguments in all these functions, but nothing is coming to mind. Multi-facet virtual objects live or die as a "cohort". It'd be easier to reason about if single-facet virtual objects lived in a cohort of size one. Then vrefs point to specific Representatives (facets), there's a function to extract the "cohort ID" from a vref (possibly a no-op), reference counts are on cohort IDs.

Oh, something which does feel more specifically wrong: why is `getExportStatus` looking at the cohort as a whole? The kernel doesn't know that the different facets are different objects, so it's going to do `dispatch.dropExports` on them individually (and we're going to re-export the facets at different times). That feels like it means we need to track the exported/recognizable/neither status for each vref (facet) separately.

Let's see, that's not the whole story. If one facet of the cohort is kept alive by the kernel, but all the other ones have been dropped by everybody, then we still need to keep the state alive. That facet might have the ability to reach the other facets during actualization, which means the refcounts on the other facets is immaterial: we must be able to recreate them if/when the one exported facet becomes real again.

Ugh, I'm going to have to think about this, maybe we can talk about it tomorrow.

warner (commented):
I'm submitting a partial review to start the discussion, I'll re-request myself for review and continue the process tomorrow. I've only gotten as far as `virtualReferences.js`.

packages/SwingSet/docs/virtual-objects.md
FUDCo: oops...

packages/SwingSet/docs/virtual-objects.md
FUDCo: I think we need to make a decision whether we are even going to allow `init` functions to return different object shapes within a kind.

packages/SwingSet/src/lib/parseVatSlots.js
FUDCo: When I first saw this was indifferent and so my reaction was "if you like `.` then fine", but then it struck me that we already use `.` informally as a kind of path separator in key strings, so I'm thinking that maybe `.` is not a good choice.  If somebody has a key string and does a split on `.` they'll end up separating the facet number from the rest of the vref, possibly without intending to do so.  However, I have no particular attachment to `,` if there's some other character you like better besides `.`.

In any case I will add a comment documenting the format.

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: I'm disinclined to do this, both for the reason you mention and because it would result in needless code duplication just to placate TypeScript.  TypeScript is supposed to be our servant, not the other way around.  We should discuss.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: Protecting against `obj` being a cyclic structure is probably warranted paranoia.  That will be quite simple because the recursion is not allowed to go more than 1 layer deep.

`obj` can't change its shape while we're looking at it because we have control.

`defineKind` calls `Far` on the facets so the rest of the stuff you're concerned about here is not an issue.

FUDCo: Unit testing here seems sensible but raises an issue that has always bugged me a bit about unit tests, namely that implementing them requires exposing APIs that are not supposed to be exposed.

packages/SwingSet/src/lib/parseVatSlots.js
FUDCo: In discussion, @warner and I decided on `:`

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: In discussion, @warner persuaded me (or, at least, caused me to persuade myself) that separate function is the way to go.  Tentatively calling these `registerValue` and `registerValueCohort`.

FUDCo: As I went to implement these changes, reality intruded and pushed to revert the previous decision -- it turns out that in the places that the `faceted` flag is used, it really is a boolean that could take either value, so each call site would have to be translated into an `if`.  Furthermore, we'd now have to pass around two functions everywhere that we currently pass around the one `registerValue` function.  Making the parameter mandatory everywhere rather than defaulting it when it is `false` also helps clarify things.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: `stored` and `returned` don't quite work because they are both returned here (the returning that the proposed name refers to happens a layer or two above in the call stack).  I tried it with `storedRepresentative` and `exposedRepresentative` and I like how that reads.  I think I'm going to go with that for now, at least until the next review pass.

erights:
Hi @FUDCo I asked myself for another review. Please don't squash any further changes yet.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: The only existing test cases where I faced this, space was better. It mirrored the choices we were already doing manually when we had to choose the string per facet. For a brand `'Foo'`, we ended up with `'Foo issuer'`, `'Foo mint'` etc. The tag was an identification as adjective and the facetName was the noun naming what kind of thing the facet is, among the group of things identified by that tag.

But this was a terribly tiny example. Let's revisit this if it doesn't work as well on other examples.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +539,40 @@ export function makeVirtualObjectManager(
...
+          exposedRepresentative = {};
+          representative = [];
+          const facetNames = Object.getOwnPropertyNames(self).sort();
+          vrm.checkOrAcquireFacetNames(kindID, facetNames);
+          for (const facetName of facetNames) {
+            const facet = Far(`${tag} ${facetName}`, self[facetName]);
+            exposedRepresentative[facetName] = facet;
+            representative.push(facet);
+          }
+          harden(exposedRepresentative);
```
erights: Before `exposedRepresentative` is hardened, `assessFacetiousness` cannot be reliable, because unhardened objects are too slippery. Don't try to validate much about unhardened objects --- that's rather hopeless. `assessFacetiousness` need only make a correct decision if `exposedRepresentative` is well behaved. Otherwise, `assessFacetiousness` need only make ***some*** decision. Whatever decision that is, after `exposedRepresentative` is hardened, that decision must either be enforced or the attempt at enforcement must throw. 

For the `'one'` case, the `exposedRepresentative` is `Far`ed, after which you know that it is a hardened passable. For the `'many'` case, if you were hardening in place as my shim did, then just hardening it doesn't ensure that it is a valid copyRecord, so I suggest adding at least the following check. This will also that it is passable, and that it has no cycles within the pass-by-copy structure.

```suggestion
          harden(exposedRepresentative);
          assertRecord(exposedRepresentative);
```

However, you'd still need to go back over it to ensure that the contents are still the objects that you `Far`ed. Now that I've talked this through, I like what you're doing better than my suggestion. By constructing a new `exposedRepresentative` from scratch yourself starting from a well behaved empty object, by asking `self` each question only once, and by not giving `self` and premature access to `exposedRepresentative`, no further mischief is possible.

I think both the non-defensive `assessFacetiousness` and the defensive code above as written, are fine. Could you didn't take my suggestion:

> On the exposed representative, curious why you don't use self itself as you do for the single facet case above. It doesn't matter so no objection. But if it's otherwise indifferent, it would reduce the cost of making the multi-facetted representative.

Bad suggestion. Bad!

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: See my reasoning below for why it is fine for `assertFacetiousness` does not need to be defensive or validating. It operates on an unfrozen object. Trying to validate that there's anything stable about an unfrozen object is close enough to impossible. Instead, whatever decision it makes then needs to be enforced *after* the relevant objects are hardened, so that they'll hold still while we're examining them. The worst a TOCTTOU can do is cause the making of the representative to fail.

@FUDCo @warner please check my reasoning here. If you buy it, it is certainly worth a significant inline comment so they next reader doesn't have to recapitulate this reasoning.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Even with your code, adding the `assertRecord(exposedRepresentative);` is belt and suspenders. I leave it to you whether it is worth it.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: As we discussed, it turns out we do need to track export status independently for each facet.  The latest rev now does that.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: I think you should revisit this in the latest rev.  I did something else before I saw this comment and I think what I did works better than either the way it worked before or what you propose here.

FUDCo: BTW, I completely don't follow what you say w.r.t `assessFacetiousness` and lack of hardening.  The job of `assessFacetiousness` is to analyze the object it is given; that analysis remains completely reliable as long as we have not returned control to user code, which is how it is used.  If it determines that the object returned by `actualize` has the right shape, that means it's in a suitable state to extract the parts needed and wrap them with `Far`, which is what the code does.

packages/SwingSet/docs/virtual-objects.md
warner: Right, but I think we have opinions (we're in favor of fixed-shape, because give us more flexibility to make a real DB schema, right?), so we should make it clear that authors ought to use a fixed shape, even if we aren't enforcing it yet. "All instances of a given kind ought to have the same set of properties" might be enough.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -487,11 +543,40 @@ export function makeVirtualObjectManager(
...
+      const facetiousness = assessFacetiousness(self);
+      switch (facetiousness) {
+        case 'one': {
+          storedRepresentative = Far(tag, self);
+          vrm.checkOrAcquireFacetNames(kindID, null);
+          exposedRepresentative = storedRepresentative;
+          break;
+        }
+        case 'many': {
+          exposedRepresentative = {};
```
warner: I'd really like to call this a `cohort`: this object is *not* a single Representative, it has no identity (you can't send it into the kernel). There are three lines at the end of the file in the common case between `one` and `many` that stand in the way of using `storedCohort` and `exposedCohort` in the `many` case. I'll keep thinking about how to deal with that. I think it's another argument for "mono-faceted virtual objects are a cohort of size one".

packages/SwingSet/src/liveslots/virtualObjectManager.js
warner: > as long as we have not returned control to user code

I think the (or at least one) threat is when `actualize` returns an object with getters, or a `Proxy`, and as you dereference `self[facetName]`, you're surprisingly giving control back to user code, which might then change the shape of that object (removing a property, changing the value of a property). So step one is to decide `one` vs `many`, step two is to extract the single/multiple facets claimed by the first step, step three is to `Far` them, step four is to assemble a new non-sneaky object out of the pieces, all just like what you're doing.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -62,48 +62,69 @@ export function makeVirtualReferenceManager(
...
     }
   }
 
-  function getExportStatus(vobjID) {
-    const raw = syscall.vatstoreGet(`vom.es.${vobjID}`);
-    switch (raw) {
-      case '0':
-        return 'recognizable';
-      case '1':
+  function getExportStatus(slot) {
```
warner: getExportStatus() now takes a cohort ID (I think you're generally using `slot` for this type), but `possibleVirtualObjectDeath()` takes a full vref (named "`vobjID`") and passes it to `getExportStatus()` verbatim. That sounds like a confused pair of types.

warner: I spent an hour writing code to exercise the problem that I expected with this confusion, and it turned out to be ok. `possibleVirtualObjectDeath` is not called with full vrefs: it's only called with the strings that are passed into `vreffedObjectRegistry.register()`, which (in this case) comes from `registerValue()`, and (for multi-facet virtual objects) that is only given the cohort ID, not the full vref-with-facet.

I really want to be able to check the correctness of this by eye, so I'm very concerned about the imprecise identifiers. I think this might be correct, and obviously your tests pass, but I'm going to remain worried about the behavior of this code in real-world deployments until we perform a throrough refactoring, starting with the difference between "slot" and "vref" and "vobjID" and any of the other variable/type names we now have flying around.

I'm glad we talked through possible naming conventions, I look forward to making some progress towards that cleanup. I like the idea of using `cohortID` in the APIs of functions that use their argument to index `vom.rc.${xx}` and `vom.es.${xx}`.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -600,10 +600,16 @@ function build(
     return result;
   }
 
-  function registerValue(slot, val) {
+  function registerValue(slot, val, valIsCohort) {
```
warner: Let's put an assert here:
```js
const { type, facet } = parseVatSlot(slot);
assert(!facet, `registerValue(${slot}) should not receive individual facets`);
```

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -540,17 +625,14 @@ export function makeVirtualObjectManager(
         propertyNames.add(prop);
       }
       const innerSelf = { vobjID, rawData, repCount: 0 };
-      const [initialRepresentative, wrappedData] = makeRepresentative(
-        innerSelf,
-        true,
-        false,
-      );
-      registerEntry(vobjID, initialRepresentative);
+      const [representative, exposedRepresentative, wrappedData] =
```
warner: So `representative` here is the "cohort record": it's the thing that gets registered in the FinalizationRegistry, so it's what GC will be watching for. What keeps it alive?

I see the `stateToRepresentative.set(wrappedData, storedRepresentative)`, and `wrappedData` is really the `state` object that is passed to the user's `actualize` and `finish` functions, and `storedRepresentative` is returned by `makeRepresentative()` and becomes this `representative` value. (complaint: having the same value take on different names in different places makes it really hard to trace this through).

There's nothing else here which keeps `wrappedData`/`state` alive. So if any of the user's facets don't happen to capture `state`, then `state` might get dropped even though one of the facets is still alive, which means the cohort record will get dropped, which means `bringOutYourDead` will think the entire cohort has died, when in fact the in-memory pillar should still be standing.

I'm bothered that this isn't caught by a unit test. I started to write one, was stymied by the VOM cache (even setting the size to 1, I had to allocate a bunch of dummy objects to get `state` to fall out of the cache and see the GC I was expecting), and stopped when my extra `console.log`s showed me the problem was real. In case it helps, the setup looks like this:

```js
const { defineKind } = VatData;
export function buildRootObject(vatPowers) {
  const makeFoo = defineKind(
    'foo',
    () => ({ prop: 'value' }),
    state => ({
      facet1: { foo: () => 0 },
      facet2: { bar: () => 0 },
    }),
  );
  let facet1;
  return Far('root', {
    bootstrap(_vats) {},
    start() {
      facet1 = makeFoo().facet1;
      makeFoo();
      makeFoo();
      makeFoo();
    },
    drop() {
      facet1 = null;
    },
  });
}
```

and to drive it:
```js
// eslint-disable-next-line import/order
import { test } from '../tools/prepare-test-env-ava.js';

// eslint-disable-next-line import/order
import { buildVatController } from '../src/index.js';
import { capargs } from './util.js';

test('drop', async t => {
  const config = {
    defaultManagerType: 'xs-worker',
    bootstrap: 'vat1',
    vats: {
      vat1: {
        sourceSpec: new URL('vat-4742.js', import.meta.url).pathname,
        creationOptions: { virtualObjectCacheSize: 1 },
      },
    },
  };
  const c = await buildVatController(config);
  c.pinVatRoot('vat1');
  await c.run();
  console.log(`-- READY`);

  c.queueToVatRoot('vat1', 'start', capargs([]));
  await c.run();

  console.log(`-- dropping`);
  c.queueToVatRoot('vat1', 'drop', capargs([]));
  await c.run();
});
```

I think the fix needs a WeakMap from each Representative (facet) to the cohort record (named `representative` here). So every facet is involved in a reference cycle with every other facet, as well as the cohort record (which is what actually matters, since it's what the FinalizationRegistry is watching).

packages/SwingSet/src/liveslots/liveslots.js
warner: Here are the notes we made on vref terminology:
```
 * A slot string (aka "vref") can take one of the forms:
 *
 *   T-N
 *   T+N
 *   o+N/I
 *   o+N/I:F

 * vrefs for exported virtual objects
 *   single-facet: vref = o+N/I
 *    (kind handle: o+9/N)
 *    //cohort ID: o+N/I
 *    slotToVal.get(vref) -> representative
 *    valToSlot.get(representative) -> vref
 *    vom.rc.${vref}
 *    vom.es.${vref} = 'n/s/r'

 *   multi-facet:  vref = o+N/I:F
 *    (kind handle: o+9/N)
 *    cohortID: o+N/I
 *    slotToVal.get(cohortID) -> cohort record
 *    valToSlot.get(representative) -> vref
 *    vom.rc.${cohortID}
 *    vom.es.${cohortID} = /[nsr]{numFacets}/

 * "cohortID" vs "cohortRef" 
 * slotToVal = Map< (CohortID | RemotableExportID | PromiseID | DeviceID) ,
                    WeakRef< CohortRecord | Representative | Far | Presence | Promise | DeviceNode > >
 * convertSlotToVal(vref => Representative | Far | Presence | Promise | DeviceNode)
 * valToSlot = WeakMap<Representative | Far | Presence | Promise | DeviceNode, vref>
 * convertValToSlot(... => vref)

 * or CohortID includes o-N , p+N, d-N, etc, just not exactly virtual object vrefs
```

We decided that `slot` is the most generic term, used in the definition of CapData and things like a Message (with `target`, `method`, `args: CapData = { body, slots }`, and `result`). Within liveslots.js, `vref` is the most generic term. `convertSlotToVal` takes a vref, but the `slotToVal` table (which yes wants to be renamed) is actually indexed by something slightly different, since when the vref is really an exported virtual object (or maybe an exported multi-facet virtual object), the table is indexed by the cohortID rather than the vref.

As discussed, please add a comment describing our grand naming plan, maybe at the top of liveslots, or the top of `parseVatSlots`, to remind us about incremental improvements we can make over time.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -62,48 +62,69 @@ export function makeVirtualReferenceManager(
...
+    const es = syscall.vatstoreGet(`vom.es.${slot}`);
+    if (es) {
+      if (es.indexOf('r') >= 0) {
```
warner: This "max(facetFlags)" approach feels right. I need to think about what actually happens as the parts of a cohort are getting cleaned out, though, if there's anything funny that might cause problems.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -62,48 +62,69 @@ export function makeVirtualReferenceManager(
...
 
   function setExportStatus(vobjID, exportStatus) {
-    const key = `vom.es.${vobjID}`;
+    const { slot, facet } = parseVatSlot(vobjID);
+    const key = `vom.es.${slot}`;
+    const esRaw = syscall.vatstoreGet(key);
+    const es = Array.from(esRaw || 'n'.repeat(getFacetCount(slot)));
+    const facetIdx = facet === undefined ? 0 : facet;
+    // The export status of each facet is encoded as:
+    // 's' -> 'recogizable' ('s' for "see"), 'r' -> 'reachable', 'n' -> 'none'
```
warner, line 157: As discussed, I'll make a note to change these flags from `[rsn]` to something with a more obvious ranking, maybe `[Rs_]` (where the vertical height of the letter indicates how "strong" the reference is), or maybe `[Rc_]` because there's a `c` in `reCognizable`. Using `[210]` would be the most obvious, but you made an excellent point that it would be horribly confused with a multi-digit integer reference count.

I won't implement that change until I have some free time, which means possibly never.

packages/SwingSet/src/liveslots/virtualReferences.js
warner: I'd like to at least announce/document that Kinds should produce the same facets (and the same state property names) on all instances, so if/when we manage to enforce it later, users will have been warned.

Fixed facet/state names gives us more to work with, and we could imagine converting that data to SQL table schema.

packages/SwingSet/test/virtualObjects/test-virtualObjectManager.js
```diff
@@ -77,6 +84,63 @@ function zotVal(arbitrary, name, tag, count) {
   });
 }
 
+test('multifaceted virtual objects', t => {
```
warner, line 87: Let's make a second test here which ignores `state`, just a pair of marker facets, to exercise the problems with the cohort record not being retained. I'd have to think a while to figure out how to capture the problem in a unit test, but let's at least have *some* code which exercises that case, to give any assertions or whatever we've got a chance to fire.

warner (changes requested):
* cohort record needs retention
* type naming comments

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: I tried renaming these in terms of the roles the values were playing rather than trying to pin down the noun, calling them `toHold` (meaning: what the vom will hold in memory) and `toExpose` (meaning: what the vom will reveal to user code).  It reads nicely.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: I'm thinking of changing `parseVatSlot` to rename the `slot` property of the return value as `cohortID`, meaning the given slot string stripped of any facet tag.  We then simply consider any referenceable object other than a facet to be the sole member of a single-member cohort.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: I could have sworn that the enforced requirement that facets be invariant was documented, but I just looked and it wasn't (it is now, or anyway will be with the next commit).  The open question is whether we should enforce invariance on properties as well.  (It also strikes me that we should ask the same question about the methods of the facets, but then we'd be getting into whether the method functions were the same functions each time, which is basically impossible to check).

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: On reconsideration, the more I think about this the less I like `cohortID`.  The problem is that that label only really has meaning in one narrow context, so using it to differentiate the general case from the much rarer specific case where the differentiation matters is confusing.  How does `baseRef` grab you?

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: In the course of writing the requested comment in `parseVatSlots.js` I think I have clarified the terminology, and in a way that addresses the concern I expressed in my comment above.   For now I'm introducing with the term "baseRef", which we can replace with something better if we decide we don't like it, but which captures the distinction I'd like to express.

 A "vref" identifies an entity visible to vat code to which messages may be sent and which may be compared for equality to other such entities.  Let's call such an entity an "addressable object".
 
A "baseRef" designates an entity that is managed by LiveSlots, both as a unit of garbage collection specifically and as a unit of memory management more generally.  Such an entity may be a promise or remotable object or imported presence, all of which will always be JavaScript objects in memory, or it may be a virtual object or collection, which can be in memory or on disk or both.  Let's call such an entity a "base object".  In most cases this is one and the same with the addressable object that the vref designates, but in the case of a faceted object it is the faceted object as a whole (represented in memory, though not on disk, as the cohort array) rather than any particular individual facet (the faceted object per se is never exposed directly to code running within the vat; only its facets are).

Given these definitions, we can describe our various translation functions/tables:

`slotToVal` maps a baseRef to a base object (actually, to a weakRef that points to the base object)
`getValForSlot` maps a baseRef to a base object, or to `undefined` if it is not resident in memory
`convertSlotToVal` maps a vref to to an addressable object, loading it from disk if necessary
 
`valToSlot` maps an addressable object to a vref
`getSlotForVal` maps an addressable object to a vref
`convertValToSlot` maps an addressable object to a vref, generating a new vref if necessary

If we accept these as correct, some renaming of functions seems called for, probably best done in a separate PR.  In the meantime, I will rename the parameters and variables within them to reflect the asymmetric typing that the above definitions reveal.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -477,57 +534,91 @@ export function makeVirtualObjectManager(
...
+          vrm.checkOrAcquireFacetNames(kindID, null);
+          toExpose = toHold;
+          break;
+        }
+        case 'many': {
+          toExpose = {};
+          toHold = [];
+          const facetNames = Object.getOwnPropertyNames(self).sort();
+          assert(
+            facetNames.length > 1,
```
warner, line 563: I suspect we'll want to relax this rule, it'll be annoying for someone commenting out one facet of a two-facet cohort during development/debugging (they'd have to comment out a pair of braces too), but I agree it will probably avoid some confusion here and there.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -51,59 +52,109 @@ export function makeVirtualReferenceManager(
...
+   *     indicated virtual object is reachable and an array of vrefs (to facets
+   *     of the object) that should now be regarded as unrecognizable
+   */
+  function getExportStatus(baseRef) {
+    const es = syscall.vatstoreGet(`vom.es.${baseRef}`);
+    if (es) {
+      const reachable = es.indexOf('r') >= 0;
+      const retirees = [];
+      if (!reachable) {
+        if (es === 's') {
```
warner, line 90: Ah, ok, this is why you needed to prohibit single-facet cohorts, because otherwise we'd retire the baseref instead of `${baseref}:0`. I consider this another argument in favor of "all virtual objects are members of a `length >= 1` cohort", but I don't think I'm going to convince you of that this week, so I'll put it aside until someone else comes along and trips over this subtlety :).

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -160,6 +218,46 @@ export function makeVirtualReferenceManager(
...
+   *    checked or acquired, or null if the kind is unfaceted
+   */
+  function checkOrAcquireFacetNames(kindID, facetNames) {
+    const kindInfo = kindInfoTable.get(`${kindID}`);
+    assert(kindInfo, `no kind info for ${kindID}`);
+    if (kindInfo.facetNames !== undefined) {
+      if (facetNames === null) {
+        assert(kindInfo.facetNames === null);
+      } else {
+        assert(arrayEquals(facetNames, kindInfo.facetNames));
```
warner: This `assert` should have a descriptive string with something like `all virtual objects of the same Kind must have the same facet names`, otherwise userspace will get a pretty opaque error. If there's an easy way to include the name/tag of the Kind, and/or the list of expected/received facet names, that will make their life even easier.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -440,8 +538,28 @@ export function makeVirtualReferenceManager(
    * @returns {boolean} true if this possibly creates a GC opportunity
    */
   function ceaseRecognition(vref) {
-    const recognizerSet = vrefRecognizers.get(vref);
     let doMoreGC = false;
+    const { id, facet } = parseVatSlot(vref);
+    if (facet === undefined) {
```
warner, line 546: `ceaseRecognition()` is only called by `dispatch.retireImports` in liveslots, so it always gets a vref (not a baseref). Moreover, `retireImports` only deals with Presences: there's an assert on `!parseVatSlot(vref).allocatedByVat`. So I don't see how this facet/`facetNames` loop could be correct.

packages/SwingSet/test/virtualObjects/test-representatives.js
```diff
@@ -376,10 +384,40 @@ test('virtual object gc', async t => {
...
+    'v1.vs.vom.es.o+1/3': 'r',
     'v1.vs.vom.o+1/2': '{"label":{"body":"\\"thing #2\\"","slots":[]}}',
     'v1.vs.vom.o+1/3': '{"label":{"body":"\\"thing #3\\"","slots":[]}}',
     'v1.vs.vom.o+1/8': '{"label":{"body":"\\"thing #8\\"","slots":[]}}',
     'v1.vs.vom.o+1/9': '{"label":{"body":"\\"thing #9\\"","slots":[]}}',
   });
 });
+
+test('empty facets are not orphaned', async t => {
+  const config = {
```
warner, line 397: let's add a comment, "facets which don't reference their `state` should still keep their cohort alive", that might not be obvious to us six months from now

warner (changes requested):
`ceaseRecognition` needs discussion, the rest is minor, we can land once we figure out that one.

packages/SwingSet/src/liveslots/virtualReferences.js
warner: @FUDCo explained this to me.. I'd missed a call to `ceaseRecognition()` from `possibleVirtualObjectDeath`, to accomodate the case where an entire cohort is dropped, in which case `ceaseRecognition()` can indeed be called with a baseref instead of a vref. And we confirmed that this clause gets exercised all the time in the new unit tests.

So I'll replace my review with a request to add a comment to `ceaseRecognition()` saying that it could be called with either a vref (in the case of `dispatch.retireImports`) or a baseref (in the case of `possibleVirtualObjectDeath`).

And I'll add a comment (in a different PR) to `dispatch.retireImports` to make it more clear how/when/why it's being called, which should help avoid the sort of confusion we had while discussing this non-problem.

warner (approved):

warner:
My remaining requested changes are cosmetic: comments and assert description strings. Feel free to land after you've added those (and rebased/squashed/etc).

@erights said he was satisfied but his review says "Changes requested", so I'm going to dismiss his review so it won't prevent you from landing this (and so we don't have to bother him with a re-review request).

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: It's not so much about avoiding confusion as that a single-facet faceted object introduces various kinds of weird internal implementation complications that I'd prefer to duck dealing with right now.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: Ah, you figured it out!  Yes, I'm glad I didn't spend time explaining in my previous comment :-)

PR #4527, opened 2022-02-10
feat(run-protocol): interest charging O(1) for all vaults in a manager

turadg:
closes: #4341 
refs: #4345 

## Description

Vault need to be able to persist to disk (https://github.com/Agoric/agoric-sdk/issues/4511). On the path to that is making them store as virtual objects in the new Collections API. While doing this refactor though we design for the durable case and take as a requirement that when interest is charged each day that it does not require disk IO for each vault (O(n)).

To make charging interest O(1) we virtualize the debt that a vault owes to be a function of stable vault attributes and values that change in the vault manager when it charges interest. Specifically,
- a `compoundedInterest` value on the manager that keeps track of interest accrual since its launch
- a `debtSnapshot` on the vault by which one can calculate the actual debt

To maintain that the keys of vaults to liquidate are stable requires that its keys are also time-independent so they're recorded as a "normalized collateralization ratio", with the actual collateral divided by the normalized debt.

Before merging I plan to,
- [x] fix or remove each skipped test
- [x] clean up the tracing (either remove or figure out a way to configure what gets output, e.g. like https://www.npmjs.com/package/debug ) 
- [x] turn each new TODO line into a ticket or do it
- [x] remove each ??? (hopefully by replacing with an answer)
- [x] convert the forEach method with callback to a generator
- [x] assess maximum key length limits with Collections API
- [ ] ~eliminate Eslint-disable lines by updating eslint config (https://github.com/Agoric/agoric-sdk/pull/4530)~

### Security Considerations

Now the actual debts of all vaults are manipulated by a single value in the vault manager, making it higher stakes. Of course if the vault manager is compromised there are bigger problems and, on the other hand, this moves that out of the vaults which are more distributed.

An implication of this change is that we'll have to https://github.com/Agoric/agoric-sdk/issues/4540 for which there may be some misleading UI could deceive users. Let's consider that when we come to that design.

### Documentation Considerations

Reviewers, how much of the design should I document in the README.md?

Anything to do in https://github.com/Agoric/documentation ? 

### Testing Considerations

There are new tests for the functionality. I think it maintained the APi that the UI was depending on. @samsiegart anything extra to do?

Issue #4482, opened 2022-02-07
Track each vault's total accrued interest in addition to total debt


turadg:
## What is the Problem Being Solved?
https://github.com/Agoric/agoric-sdk/issues/4341 will make actual accrued interest be a function of what's stored in the vault and a running tally of compounded interest.

## Description of the Design

This will require wallets to have both the vault and that tally state and do the math in order to show accrued interest.

## Security Considerations

TBD

## Test Plan

TBD

Tartuffo:
@turadg Is something you'll be doing?

turadg:
I don't know yet is why I didn't assign it. It's a consequence of #4341 which I'm doing, so it may make sense for me to but I don't know what priorities and availability will be when it comes time for this.

turadg:
Obsoleted with the sunsetting of Inter Protocol

PR #4743, opened 2022-03-05
refactor(run-protocol): extract a vaultKit module

turadg:
## Description

Several refactorings and cleanups. I split across multiple commits to see the flow and I'll autosquash upon approval.

- extract a vaultKit module
- make naming more consistent
- define more types with their constructor functions

### Security Considerations

None.

No, implementation details.

### Testing Considerations

No changes.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -10,20 +10,24 @@ import {
   floorMultiplyBy,
   floorDivideBy,
 } from '@agoric/zoe/src/contractSupport/index.js';
-import { makeNotifierKit } from '@agoric/notifier';
 
 import { assert } from '@agoric/assert';
 import { AmountMath } from '@agoric/ertp';
 import { Far } from '@endo/marshal';
 import { makeTracer } from '../makeTracer.js';
 import { calculateCurrentDebt, reverseInterest } from '../interest-math.js';
+import { makeVaultKit } from './vaultKit.js';
 
 const { details: X, quote: q } = assert;
 
-const trace = makeTracer('Vault');
+const trace = makeTracer('IV');
 
-// a Vault is an individual loan, using some collateralType as the
-// collateral, and lending RUN to the borrower
+/**
+ * @file This has most of the logic for a Vault, to borrow RUN against collateral.
+ *
+ * The logic here is for InnerVault which is the majority of logic of vaults but
+ * the user view is the `vault` value contained in VaultKit.
+ */
 
 /**
  * Constants for vault phase.
@@ -65,38 +69,6 @@ const validTransitions = {
  */
 
 /**
- *
- * @param {InnerVault | null} inner
- */
-const makeOuterKit = inner => {
-  /** @type {NotifierRecord<VaultUIState>} */
-  const { updater: uiUpdater, notifier } = makeNotifierKit();
-
-  /** @returns true iff this outer vault is backed by an inner vault */
-  const assertOwned = v => {
-    // console.log('OUTER', v, 'INNER', inner);
-    assert(inner, X`Using ${v} after transfer`);
-    return inner;
-  };
-  const vault = Far('vault', {
-    getNotifier: () => notifier,
-    makeAdjustBalancesInvitation: () =>
-      assertOwned(vault).makeAdjustBalancesInvitation(),
-    makeCloseInvitation: () => assertOwned(vault).makeCloseInvitation(),
-    makeTransferInvitation: () => {
-      const tmpInner = assertOwned(vault);
-      inner = null;
-      return tmpInner.makeTransferInvitation();
-    },
-    // for status/debugging
-    getCollateralAmount: () => assertOwned(vault).getCollateralAmount(),
-    getCurrentDebt: () => assertOwned(vault).getCurrentDebt(),
-    getNormalizedDebt: () => assertOwned(vault).getNormalizedDebt(),
-  });
-  return { vault, uiUpdater };
-};
-
-/**
  * @typedef {Object} InnerVaultManagerBase
  * @property {(oldDebt: Amount, newDebt: Amount) => void} applyDebtDelta
  * @property {() => Brand} getCollateralBrand
@@ -136,8 +108,10 @@ export const makeInnerVault = (
    */
   const assignPhase = newPhase => {
     const validNewPhases = validTransitions[phase];
-    if (!validNewPhases.includes(newPhase))
-      throw new Error(`Vault cannot transition from ${phase} to ${newPhase}`);
+    assert(
+      validNewPhases.includes(newPhase),
+      `Vault cannot transition from ${phase} to ${newPhase}`,
+    );
     phase = newPhase;
   };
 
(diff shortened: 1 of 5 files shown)


packages/run-protocol/src/vaultFactory/vaultKit.js
```diff
@@ -0,0 +1,73 @@
...
+    assert(inner, X`Using ${v} after transfer`);
+    return inner;
+  };
+
+  const vault = Far('vault', {
+    getNotifier: () => notifier,
+    makeAdjustBalancesInvitation: () =>
+      owned(vault).makeAdjustBalancesInvitation(),
+    makeCloseInvitation: () => owned(vault).makeCloseInvitation(),
+    makeTransferInvitation: () => {
```
dtribble, line 43: Adding a comment here would be good: "Starting a transfer revokes the outer vault. The associated updater will get a special notification the the vault is being transferred."

packages/run-protocol/src/vaultFactory/vaultKit.js
```diff
@@ -0,0 +1,73 @@
...
+    // for status/debugging
+    getCollateralAmount: () => owned(vault).getCollateralAmount(),
+    getCurrentDebt: () => owned(vault).getCurrentDebt(),
+    getNormalizedDebt: () => owned(vault).getNormalizedDebt(),
+  });
+  return { vault, vaultUpdater: updater };
+};
+
+/**
+ *
```
dtribble, line 57: Return the kit for a vault in the structure that wallets are expecting

packages/run-protocol/src/vaultFactory/vaultKit.js
```diff
@@ -0,0 +1,73 @@
...
+ * @param {InnerVault} inner
+ * @param {Notifier<import('./vaultManager').AssetState>} assetNotifier
+ */
+export const makeVaultKit = (inner, assetNotifier) => {
+  const { vault, vaultUpdater } = wrapVault(inner);
+  const vaultKit = harden({
+    assetNotifier,
```
dtribble, line 64: The structure here is not what wallets are expecting.  They need `uiNotifier`.  This got renamed, but it means that the wallet doesn't know to treat it specially (in lib-wallet getUINotifier).

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -10,20 +10,24 @@ import {
...
 
 const { details: X, quote: q } = assert;
 
-const trace = makeTracer('Vault');
+const trace = makeTracer('IV');
 
-// a Vault is an individual loan, using some collateralType as the
-// collateral, and lending RUN to the borrower
+/**
+ * @file This has most of the logic for a Vault, a to borrow RUN against collateral.
```
dtribble: "a to borrow" => "to borrow"

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -105,6 +77,28 @@ const makeOuterKit = inner => {
...
 
+/**
+ * @typedef {Readonly<{
+ * assetNotifier: Notifier<import('./vaultManager').AssetState>,
+ * idInManager: VaultId,
+ * priceAuthority: ERef<PriceAuthority>,
+ * runMint: ZCFMint,
+ * vaultSeat: ZCFSeat,
+ * zcf: ContractFacet,
+ * }>} ImmutableState
```
dtribble, line 88: I wonder if we should have `props` a la React for immutable and have separate storage for those!

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -117,61 +111,67 @@ export const makeInnerVault = (
...
+  };
 
+  // #region Phase logic
   /**
    * @param {InnerPhase} newPhase
    */
   const assignPhase = newPhase => {
+    const { phase } = state;
     const validNewPhases = validTransitions[phase];
     if (!validNewPhases.includes(newPhase))
```
dtribble: MUST add braces

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -105,6 +77,28 @@ const makeOuterKit = inner => {
  * @property {(oldDebt: Amount, oldCollateral: Amount, vaultId: VaultId) => void} updateVaultPriority
  */
 
+/**
+ * @typedef {Readonly<{
```
turadg, line 81: I put these in here but stopped short of using them because they're available in closure scope and it doesn't make much sense to change them before factoring out `makeMethods` or whatever we call it. I may have time for that before Monday if it's worthwhile, though I'd prefer to land the non-state parts of this PR first.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: How did this get through review in the first place? We require braces for control constructs.

dtribble: This should just be an `assert`

dtribble (changes requested):
Generally looks great. I few key minor but important issues.

packages/run-protocol/src/vaultFactory/vaultKit.js
turadg: Let me know if you're talking about more than `uiNotifier` discussed in https://github.com/Agoric/agoric-sdk/pull/4743#discussion_r820119347 .

packages/run-protocol/src/vaultFactory/vault.js
turadg: Neat idea. We'd need the virtual objects API to send them in. Could be useful for distinguishing between things the contract gets to change vs things it just needs to be provided consistently.

packages/run-protocol/src/vaultFactory/vault.js
turadg: > How did this get through review in the first place? We require braces for control constructs.

I suppose because "require braces for control constructs" is not in https://github.com/Agoric/agoric-sdk/wiki/Coding-Style or the lint configuration. IMO if something is a rule and can be reliably checked by the machine, then humans shouldn't pay attention to looking for it. https://github.com/Agoric/agoric-sdk/issues/4747

In this case I'll change to `assert`

packages/run-protocol/src/vaultFactory/vaultKit.js
turadg: The vault notifications were broken by https://github.com/Agoric/agoric-sdk/pull/4527 because they can no longer rely on one notifier. We discussed multiple designs in the RUN protocol meeting and agreed this was the best option. https://github.com/Agoric/agoric-sdk/commit/554d41ed9f9b35cd59133818e428d3055006e1ca renamed uiNotifier to make it clear to consumers that there was no longer a single notifier for UI.

I think that's the right thing for now for clarity of how the contract works. ("UI updater" is a misnomer.) Updating the UI code is queued up in https://github.com/Agoric/agoric-sdk/issues/4482 which can happen after Restival because UI is not in scope.

dtribble (approved):

PR #4745, opened 2022-03-05
feat(vats): move BLD mint, issuer to its own vat (DRAFT)

### LINKED ISSUE #4578: Isolate BLD issuer / mint in its own contract vat

The BLD issuer kit is created in the bootstrap vat, which means that the mint, issuer, and all the purses are there too. The bootstrap vat gets access to extremely high authority objects; it should _only_ partition access to those object among other components. Running the BLD mint in the same vat is awkward w.r.t. POLA.

## Description of the Design

Create a new contract whose only job is to create an IssuerKit. In fact, the `mint` could be the `creatorFacet` and the issuer could be the `publicFacet`.

## Security Considerations

See problem description above.

## Test Plan

It's not clear that this should result in any observable differences, so basic unit tests for the new contract, plus all the existing tests, seem to suffice.

cc @dtribble @Chris-Hibbert

dckc:
closes: #4578

## Description

adds a `mintHolder.js` contract that basically just wraps `makeIssuerKit`.

### Security Considerations

isolates BLD mint from bootstrap vat

TODO:

 - [x] check that the bootstrap visualization reflects this change (update manifest if not)

### Testing Considerations

very lightly tested. I should kick the tires a bit more.


packages/run-protocol/src/mintHolder.js
```diff
@@ -0,0 +1,29 @@
...
+ * This contract holds one mint; it basically wraps
+ * makeIssuerKit in its own contract, and hence in
+ * its own vat.
+ *
+ * @param {ContractFacet} zcf
+ * @returns {{ publicFacet: Issuer, creatorFacet: Mint }}
+ *
+ * @typedef {{
+ *   assetKind: AssetKind,
+ *   displayInfo: DisplayInfo,
```
michaelfig, line 17: Should the type contain `keyword` too?

```suggestion
 *   displayInfo: DisplayInfo,
 *   keyword: string,
```

packages/vats/src/core/chain-behaviors.js
```diff
@@ -275,11 +275,17 @@ export const connectChainFaucet = async ({ consume: { client } }) => {
 };
 harden(connectChainFaucet);
 
+// XXX: move shareBootContractBundles belongs in basic-behaviors.js
```
michaelfig, line 278: IIUC, we should be lazier with the contract bundles.  Reference their hashes in a governance action, and install the bundle out-of-band.

michaelfig (approved):
SARTM!

(Seems About Right To Me, given that changing AMM types etc. is probably out of scope.)

packages/vats/src/core/chain-behaviors.js
dckc: We're not quite there yet though, IIUC.

PR #4751, opened 2022-03-06
feat(swingset-vat): Integrate Endo checkBundle

### LINKED ISSUE #4719: fully validate bundles during `controller.validateAndInstallBundle`

kriskowal:
- feat(swingset-vat): Integrate Endo checkBundle
- chore: Update yarn.lock

closes: #4719

## Description

This change replaces the place-holder code for integrity checks with new features
for this purpose from Endo.

1. `bundleSource` now provides a hash for Endo Zip Base64 style bundles (default).
2. `checkBundle` is a new function that verifies the internal integrity of a bundle.

This leaves open further work to do an additional check to ensure that bundles
contain only executable code.

### Security Considerations

1. Bundles must be frozen before an integrity check can be trusted because
   accessors or proxy traps could invalidate the internalc consistency between
   the hash and content properties.

1. We will eventually need a guide for verifying the integrity of bundles on-chain.
   So far, this seems to involve obtaining the bundle hash from the chain,
   extracting the hash component, finding the allegèd sources, running project
   initialization programs in a safe container to get an inert copy of the
   repository and its dependency graph, running the bundler on the inert
   contract code, and verifying that the hashes match.

### Testing Considerations

I have used the existing tests against which the place-holder previously ran.


kriskowal:
@warner pinging you for an early review. I’m still thinking of ways to make this appear to be a straight line to design, and I believe I ought to factor out “computeBundleTaggedHash” or dispense with the notion entirely.

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -392,7 +391,12 @@ export async function initializeSwingset(
       // during config, we believe bundle.id, but not at runtime!
       return bundle;
     }
-    return computeBundleID(bundle).then(id => ({ ...bundle, id }));
+    const { endoZipBase64Sha512 } = bundle;
+    assert.typeof(endoZipBase64Sha512, 'string');
+    return {
+      ...bundle,
+      id: `b1-${endoZipBase64Sha512}`,
+    };
```
kriskowal, line 399: I would like to eliminate this since the bundle ID is already implied in the bundle and I don’t think we should be getting here unless the bundle was vetted, so I’m looking into how to not need `addBundleID`.

kriskowal: But `bundle.id` annotation turns out to be pervasive, so leaving this where I found it.

packages/SwingSet/src/controller/controller.js
```diff
@@ -303,22 +311,14 @@ export async function makeSwingsetController(
...
-    bundle = harden({
-      moduleFormat: bundle.moduleFormat,
-      endoZipBase64: bundle.endoZipBase64,
-    });
-
-    // this only computes the hash of the compartment map, it does not check
-    // that the rest of the bundle matches
-    const bundleID = await computeBundleID(bundle);
-    if (allegedBundleID) {
+    await checkBundle(bundle, computeSha512);
```
warner, line 327: This is meant to be the main defensive entry point, so we must assume `bundle` can be anything. We don't need to `harden` the object first, because `checkBundle` asserts `frozen`. But.. what if `bundle` has a getter for `endoZipBase64Sha512` that returns the bundle byte's hash the first time, then the caller's expected hash the second time? Could that trick it?

(this would be a `check-bundle` issue that I failed to think of when reviewing that other package, not a bug in this callsite, but we shouldn't switch to `checkBundle` until we know it's fully defensive)

packages/SwingSet/src/controller/controller.js
```diff
@@ -303,22 +311,14 @@ export async function makeSwingsetController(
...
+    const { endoZipBase64Sha512 } = bundle;
+    assert.typeof(endoZipBase64Sha512, 'string');
```
warner, line 329: This assert is necessary because the bundle could be internally self-consistent, but be of a format that doesn't provide `endoZipBase64Sha512`, right? Just making sure I understand that that's the only surprise that could make it past `checkBundle`.

packages/SwingSet/src/controller/initializeSwingset.js
warner: Adding `.id` was quick and convenient, feel free to change the pattern. In general, I've waffled between passing around `{ id, bundle }` vs passing around `{ id, ...bundle }`. The latter allowed me to change less code, but that doesn't mean it's the cleanest way to do it.

warner (approved):
Looks good, but please think about my getter concern before landing.

packages/SwingSet/src/controller/controller.js
kriskowal: https://github.com/endojs/endo/issues/1106

packages/SwingSet/src/controller/controller.js
kriskowal: I should also verify `moduleFormat`. At the moment, `checkBundle` will only bless a bundle if `moduleFormat` is `"endoZipBase64"`, `endoZipBase64` is a string, and `endoZipBase64Sha512` is a string. So this is fine for now, but in the face of evolution, `checkBundle` should eventually allow other formats.

I’m less concerned here about surprises coming out of `checkBundle` and more about other ways bundles might arrive here that might not necessarily get checked. The presence of this property is necessary to compute an ID for now.

PR #4753, opened 2022-03-06
fix: freeze globalThis in coreEval to enforce OCap discipline

dckc:
refs: #4642, #4352

I pointed out to @erights the other day that in `coreEval`, rather than make powers ambient globals in the compartment, we evaluate a function in a powerless environment and then pass powers to it, in order to allow the evaluated code to take advantage of OCap discipline.

@erights pointed out that `globalThis` isn't frozen in that compartment. Oops.


erights (approved):
LGTM Thanks!

PR #4754, opened 2022-03-06
test(zoe): good error message if payment not wrapped with keywords

### LINKED ISSUE #3765: Need to improve error message for a payment being sent instead of paymentKeywordRecord

If a payment is sent in `zoe.offer` instead of a payment keyword record, the error message is not helpful:

```
  Error {
    message: 'keyword "getAllegedBrand" must be ascii and must start with a capital letter.',
  }

  › makeError (node_modules/ses/dist/ses.cjs:2572:17)
  › fail (node_modules/ses/dist/ses.cjs:2700:20)
  › baseAssert (node_modules/ses/dist/ses.cjs:2718:13)
  › assertKeywordName (node_modules/@agoric/zoe/src/cleanProposal.js:31:3)
  › Array.forEach (<anonymous>)
  › cleanKeywords (node_modules/@agoric/zoe/src/cleanProposal.js:103:12)
  › depositPayments (node_modules/@agoric/zoe/src/zoeService/escrowStorage.js:68:29)
  › Alleged: zoeService.offer (node_modules/@agoric/zoe/src/zoeService/offer/offer.js:39:37)
``` 
We should improve the error message through asserting that the paymentKeywordRecord is a copyRecord before proceeding.

dckc:
closes: #3765

### DIFF
--- a/packages/zoe/test/unitTests/test-zoe.js
+++ b/packages/zoe/test/unitTests/test-zoe.js
@@ -4,7 +4,7 @@ import { test } from '@agoric/zoe/tools/prepare-test-env-ava.js';
 
 import path from 'path';
 
-import { AmountMath, AssetKind } from '@agoric/ertp';
+import { AmountMath, AssetKind, makeIssuerKit } from '@agoric/ertp';
 import { E } from '@agoric/eventual-send';
 import { makePromiseKit } from '@endo/promise-kit';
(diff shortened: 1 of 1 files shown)


erights (approved):
Agreed that the current error message is good enough to close the bug and then not worry about (without evidence to the contrary).

LGTM.

PR #4562, opened 2022-02-16
Fix typos

rex4539:
_p.s.: see #4755 by @warner_

PR #4755, opened 2022-03-07
Fix typos

warner:
Copy of @rex4539 's PR #4562 , updated to current trunk. Still needs @erights 's recommended fix applied and review.

I'm happy to do the work to review this, but since I'm submitting this branch, Github won't let me assign myself as reviewer. 

My plan is to review rex's changes and then announce my satisfaction, and then maybe somebody else can rely upon my claim and do a cursory-but-Github-satisfying review.


warner:
@samsiegart could you look at the change to `wallet/ui/` about "palette"? I *think* it's correct, because it matches the other spellings in that package, but it's the most behavior-affecting change in the PR and I don't know that code at all.

@dckc take a look at the test-xsnap changes I made to make sure the error we're catching is really about `print` or `gc` and not, say, `evaluate` being misspelled. There are ways to make it more refined (and less dependent upon xsnap's error-message internals) but that involved more deeply nested code than I wanted to write today.

@dckc I've scanned all changes to make sure they make sense, but if you could also do a pass, I'd appreciate it.

samsiegart (approved):
Palette change LGTM. It was just a mock value for a test, but looks like it's not needed, so not affecting anything anyway. It's spelled correctly elsewhere in the dapp

dckc:
> @dckc take a look at the test-xsnap changes I made to make sure the error we're catching is really about `print` or `gc` and not, say, `evaluate` being misspelled.

looks good.

> There are ways to make it more refined (and less dependent upon xsnap's error-message internals) but that involved more deeply nested code than I wanted to write today.

Counting exact error messages as part of our API seems to be the norm in agoric-sdk in any case.

> @dckc I've scanned all changes to make sure they make sense, but if you could also do a pass, I'd appreciate it.

Have not done that yet; here's hoping...

packages/xsnap/test/test-xsnap.js
```diff
@@ -134,12 +134,15 @@ test('print - start compartment only', async t => {
     print('print:', 123);
     try {
       (new Compartment()).evaluate('print("456")');
-    } catch (_err) {
-      send('no print in Compartment');
+    } catch (err) {
+      send('err:' + err);
```
erights, line 140: Did you intend no space? If so
```suggestion
      send(`err:${err}`);
```
otherwise
```suggestion
      send(`err: ${err}`);
```

packages/xsnap/test/test-xsnap.js
```diff
@@ -151,12 +154,15 @@ test('gc - start compartment only', async t => {
     gc();
     try {
       (new Compartment()).evaluate('gc()');
-    } catch (_err) {
-      send('no gc in Compartment');
+    } catch (err) {
+      send('err:' + err);
```
erights, line 160: here too

packages/xsnap/test/test-xsnap.js
warner: eh, space vs no space didn't feel important

the whole clause is already inside a quasiliteral, so I couldn't use the normal backticks

erights: > the whole clause is already inside a quasiliteral, so I couldn't use the normal backticks

Oops. I didn't notice that. Nevermind.

dckc (approved):
i skimmed the changes; they seem right.

erights:
@warner , is this PR stuck in CI? Is that why it's not merged?

warner:
yeah, merge conflicts with current trunk, I'll address those

PR #4757, opened 2022-03-08
chore: clean up threat model docs

turadg:
## Description

There was some stale info in the VaultFactory model, which this removes.

I also made the gen script work with Hombrew plantuml, write to a README.md so it appears in Github directory view, and use SVGs to better resolution independence.

### Security Considerations

Less wrong docs helps security investigations.

This doesn't perfectly match the code but VaultFactory is at least a valid subset now.

### Testing Considerations

--


jessysaurusrex (approved):
LGTM

PR #4758, opened 2022-03-08
refactor: Relieve dependency on babel-standalone

kriskowal:
refs: https://github.com/endojs/endo/issues/768

## Description

Since time immemorial, SES depended on a fork of Babel-Standalone to facilitate its Module->to->Program transform in the `StaticModuleRecord` constructor. This bloated the weight of SES up to 3MB because the Standalone contains every Babel plugin and shim. We factored it out of SES to keep light, only bringing in the dependency on Babel if you needed to create a bundle. Then, we learned that Babel Standalone shims couldn’t initialize after Lockdown, so we had to tread carefully to ensure that Babel Standalone was initialized exactly once in an application, before Lockdown. So, even Dapps had to take a dependency on Babel Standalone and initialize it before Lockdown in any test that used Bundle Source.

That era comes to an end with this change. With the guidance of @nicolo-ribaudo and the labor of @michaelfig, we have replaced Babel Standalone with the three internal modules of Babel that we need, no shims and no plugins. Consequently, we don’t need to care about Babel Standalone anywhere anymore.

### Security Considerations

This change greatly reduces the weight of the Agoric SDK supply chain, eliminating many shims that previously had to be initialized in vetted shim territory. It may now even be possible to create bundles from within a Compartment.

The same relaxation on dependence on Babel Standalone is now possible in our documentation.

### Testing Considerations

Existing tests adequately cover bundling.


packages/agoric-cli/src/entrypoint.js
```diff
@@ -2,9 +2,6 @@
 
 /* global process */
 
-import '@endo/init/pre-bundle-source.js';
-// TODO Remove esm preinitialization
-// https://github.com/endojs/endo/issues/768
```
michaelfig, line 7: Replace with:

```suggestion
import '@endo/init/pre.js';
// TODO Remove esm preinitialization
// https://github.com/endojs/endo/issues/768
```

packages/agoric-cli/test/test-main.js
```diff
@@ -1,7 +1,4 @@
 /* global globalThis */
-// TODO Remove esm preinitialization
-// https://github.com/endojs/endo/issues/768
-import '@endo/init/pre-bundle-source.js';
```
michaelfig, line 4: Replace with:

```suggestion
import '@endo/init/pre.js';
// TODO Remove esm preinitialization
// https://github.com/endojs/endo/issues/768
```

packages/swingset-runner/src/runner-debug-entrypoint.js
```diff
@@ -5,10 +5,6 @@
  * as yet not-entirely-ESM-supporting version of NodeJS.
  */
 
-// TODO Remove babel-standalone preinitialization
-// https://github.com/endojs/endo/issues/768
-import '@endo/init/pre-bundle-source.js';
-
```
michaelfig: Replace with:

```suggestion
import '@endo/init/pre.js';
```

michaelfig (approved):
Just a few places where `@endo/init/pre.js` should be loaded so that `ses` is loaded before the vetted shims (but before `lockdown`), which should help improve confinement a bit.

PR #4764, opened 2022-03-08
chore(run-protocol): improve comments and naming

turadg:
## Description

- Fixes a state error
  - not in service yet but https://github.com/Agoric/agoric-sdk/pull/4736 removed a `ts-expect-error` which revealed an error in https://github.com/Agoric/agoric-sdk/pull/4743
  - Apparently a TS suppression on a key in the object suppresses errors on the whole object. (Maybe just for .js?) This changes that key to `any` so it can cast to the type within `state` without suppression. It then fixes type errors that were detected.
- replace 'run' with 'debt' where safe to do
- add some comments

### Security Considerations

--

Gotcha documented in https://github.com/Agoric/agoric-sdk/wiki/Typescript-using-JSDoc/

### Testing Considerations

--


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -142,8 +144,8 @@ export const makeInnerVault = (
 
     // Two values from the same moment
     interestSnapshot: manager.getCompoundedInterest(),
-    // @ts-expect-error until https://github.com/Agoric/agoric-sdk/pull/4736
-    runSnapshot: AmountMath.makeEmpty(runBrand),
+    /** @type {any} cast */
```
dckc, line 147: This contract only handles fungible debt, right? Have you tried declaring the `debtBrand` about (around line 120) to `Brand<'nat'>` to see if this cast becomes unnecessary? I wonder if `getIssuerRecord` gives you the assetKind so that you could do a runtime check to be sure.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -285,12 +287,13 @@ export const makeInnerVault = (
...
   const snapshotState = newPhase => {
-    const { runSnapshot: run, interestSnapshot: interest } = state;
+    const { debtSnapshot: run, interestSnapshot: interest } = state;
     /** @type {VaultUIState} */
     return harden({
       // TODO move manager state to a separate notifer https://github.com/Agoric/agoric-sdk/issues/4540
       interestRate: manager.getInterestRate(),
       liquidationRatio: manager.getLiquidationMargin(),
+      // XXX 'run' is implied by the brand in the amount
       debtSnapshot: { run, interest },
```
dckc, line 297: perhaps rename things to `{ amount, interest }`?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -618,15 +621,15 @@ export const makeInnerVault = (
     // vaultSeat. Would using a separate seat clarify the accounting?
     // TODO what if there isn't anything to mint?
     const { vaultSeat } = state;
-    runMint.mintGains(harden({ RUN: toMint }), vaultSeat);
+    mint.mintGains(harden({ RUN: toMint }), vaultSeat);
```
dckc, line 624: not ready to change the `RUN` keyword yet?

packages/run-protocol/src/vaultFactory/vaultKit.js
```diff
@@ -31,6 +31,11 @@ const wrapVault = innerVault => {
     return inner;
   };
 
+  /**
+   * Public API of the vault.
+   *
+   * @see {InnerVault} for the internal API it wraps.
```
dckc, line 37: That's perhaps expedient, but for actual clients of this contract, referring them to the inner guts doesn't seem like the way to go.

_mostly thinking out loud..._ I suppose it's more relevant for agoric.com/documentation , but I have hopes of generating that from this.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -160,7 +160,7 @@ export const makeVaultManager = (
           await liquidate(
             zcf,
             vault,
-            runMint.burnLosses,
+            debtMint.burnLosses,
```
dckc, line 163: `debtMint` is a bit nicer than bare `mint`.

dckc (approved):
looks good. just a few minor comments to consider.

packages/run-protocol/src/vaultFactory/vault.js
turadg: naming and organization to be cleaned up in https://github.com/Agoric/agoric-sdk/issues/4482

packages/run-protocol/src/vaultFactory/vault.js
turadg: No that's queued up in https://github.com/Agoric/agoric-sdk/issues/4800 and blocked on a design decision for what it will be.

packages/run-protocol/src/vaultFactory/vaultKit.js
turadg: I agree and see that as out of scope. We need a bigger discussion around naming of contract facets.

packages/run-protocol/src/vaultFactory/vault.js
turadg: `Brand` is not parameterized yet in master. https://github.com/Agoric/agoric-sdk/pull/4736/ will fix this and the expect-error will error to be cleaned up.

packages/run-protocol/src/vaultFactory/vaultKit.js
dckc: I'm not talking about the name of the facets. I'm talking about information hiding, or lack thereof. Clients of this API should not need to be aware of the inner API.

turadg: Ohhh. Just this jsdoc `@see`. 

I'm not sure. Developers of agoric-sdk should know about it. End users of a GUI or REPL shouldn't (and won't see the jsdoc). What other case is there? Developers building other contracts that compose the vaultFactory contract? In that case I think it behooves them to understand the inner/outer relationship (e.g. to understand how transfers work).

PR #4767, opened 2022-03-08
Rework `xsnap/src/object-inspect.js`

### LINKED ISSUE #2146: console API on XS: compatibility, security impact?
**What is the Problem Being Solved?**

On XS, to date, we have stubbed out the `console` API; for example, in https://github.com/Agoric/agoric-sdk/pull/2145). This API gets exposed to user code, which is an incompatibility risk and such incompatibility risks notoriously lead to vulnerabilities.

See also UTF-8 in XS https://github.com/Agoric/agoric-sdk/issues/2118

cc @kriskowal @erights @warner 

**Description of the Design**

goals... somewhat conflicting...

 - [ ] never throws an error
 - [ ] when on chain: deterministic: don't let caller detect when you did toString() -> don't mess with args... so noop
 - [ ] slog
 - [ ] post-hoc debugging

also:

 1. [console-shim.js](https://github.com/Agoric/agoric-sdk/blob/master/packages/xsnap/lib/console-shim.js) as of 79e81b0 
     - provides a `console` that has at least `noop` methods for everything in SES [VirtualConsole](https://github.com/Agoric/SES-shim/blob/master/packages/ses/src/error/types.js#L259) (52e6830).
 2. [vatConsole](https://github.com/Agoric/agoric-sdk/blob/79e81b014ea6e3df1a98ef0d35e7cad2d1d966a6/packages/SwingSet/src/kernel/slogger.js#L34) seems to make a facet with `['debug', 'log', 'info', 'warn', 'error']` methods for use in local vat managers. Other vat managers seem to do likewise.

**Security Considerations**

Incompatibility risks notoriously lead to vulnerabilities

Are contract writers prepared for `console.clear()` to throw? Do we teach them about commit points?

**Test Plan**

?

### LINKED ISSUE #4814: review or replace `object-inspect.js` used by xsnap console

In #4364, we changed the `xsnap` console renderer. Previously, it used the `assert.quote` function from SES, which grounds out in [`bestEffortStringify()`](https://github.com/endojs/endo/blob/13eb59d6d76aa12c82be69f28d1c2fe2c00acbc6/packages/ses/src/error/stringify-utils.js#L84), which is basically a JSON encoder with a replacer that handles non-JSON things like `undefined`. After #4364, it used a vendored copy of `object-inspect.js` (taken from the [NPM object-inspect package](https://www.npmjs.com/package/object-inspect)), which should be pretty close to what Node.js uses for its own `console` global.

During the resulting discussion (Agoric folks can look at `#eng_ses_internal` on 08-feb-2022), @erights and @dckc expressed concern about `object-internal.js`. We didn't come to a consensus before we had to move on, but I think our general opinion was that we must either do a line-by-line review of `object-internal.js`, or write our own version of it (and use that version to replace `assert.quote`, making it more readable in the process).

The concerns are:
* `object-internal.js` runs in the start compartment as a vetted shim, so obviously has a lot of power
* the behavior must be a deterministic function of the inputs to a single `console.*` call
  * there must be no information or determinism leakage between multiple calls
    * (the implementation needs a WeakSet to tolerate cycles, but this should be scoped to a single call)
  * callers will receive a hardened `console` object and will not have access to a clock or weakrefs
  * however callers might supply a `Proxy`, or objects with getters, in an attempt to confuse the console
* the console should bottom out in a function that delivers one string per `console.log()` invocation
  * we might route these strings to the kernel process for inclusion in debug logs, and/or write them to stdout/stderr, and/or drop them
  * the renderer's behavior and metering should not depend upon what happens after it emits the string
  * it must emit strings even if the caller attempts to trick it with Proxies or getters
* the inputs to `console.*` might not be hardened, and `console.*` must not harden them

michaelfig:
closes: #2146
closes: #4814

## Description

This is the next step toward improving the code used by xsnap to provide a working `console.log`.

### Security Considerations

Improves the security of our xsnap process by providing a vetted shim that really has been vetted.


packages/xsnap/test/test-promise-reject.js
```diff
@@ -0,0 +1,49 @@
...
+      Promise.reject(new Error('lose'));
+      gc();
+      report.result = ArrayBuffer.fromString(JSON.stringify(true));
+      return report;
+    };
+  `);
+
+  return {
+    /** @param { string } src */
+    async run(src) {
```
dckc: concise method syntax is against house style now, no?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+const bigIntValueOf = BigInt.prototype.valueOf;
+const getOwnPropertySymbols = Object.getOwnPropertySymbols;
+const symToString = Symbol.prototype.toString;
+// ie, `has-tostringtag/shams
+const toStringTag = Symbol.toStringTag;
+const isEnumerable = Object.prototype.propertyIsEnumerable;
+const dateToISOString = Date.prototype.toISOString;
+
+const gPO = Reflect.getPrototypeOf;
+const hasOwn = Object.prototype.hasOwnProperty;
```
dckc: Why alias all this stuff? I have some vague ideas, but I don't understand well enough to maintain this list. In particular, the `mapSize` and `setSize` contortions.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
+/* eslint-disable no-empty */
+/* eslint-disable no-nested-ternary,no-use-before-define */
```
dckc: The justification for `no-use-before-define` is to reduce the diffs from the code we started from?

I'm not getting much benefit. github thinks this is a complete new file. :-/

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+  }
+  if (typeof obj === 'number') {
+    if (obj === 0) {
+      return Infinity / obj > 0 ? '0' : '-0';
+    }
+    return String(obj);
+  }
+  if (typeof obj === 'bigint') {
+    return `${obj}n`;
+  }
```
dckc, line 99: bookmark... I read to here.

dckc (commented):
this is how far I got before the dinner bell rang...

packages/xsnap/test/test-promise-reject.js
erights: Correct. Please use arrow functions unless there's a reason not to.

michaelfig: > Correct. Please use arrow functions unless there's a reason not to.

This file was included accidentally.  Removed.

erights:
@FUDCo , @michaelfig is here doing a permissive unprivileged object-quotation system much like I originally wanted to do for `q` (aka `assert.quote`). I didn't do that because, *as far as I remember*, you really wanted the output of `q` to be within JSON and eventually convinced me to do that. IIUC, you do not recall lobbying for that and are in fact indifferent to the JSON-ability of the `q` output. I'm just giving background. I don't care what really happened historically.

With that as background, could you confirm that, once this PR is merged, you would not object to `q` using its form of quotation?

dckc:
How does test coverage for this version compare with what it replaces? Help me figure it out?

https://agoric-sdk-coverage.netlify.app/packages/xsnap/lib/object-inspect.js.html ?
70% Statements 378/540 49.75% Branches 103/207 77.14% Functions 27/35 70% Lines 378/540

michaelfig:
@erights, let's discuss how to review this:

- is this okay as a vetted shim?  Are we confident it doesn't reach for powerful things?
- what level of coverage do we need? https://agoric-sdk-coverage.netlify.app/packages/xsnap/lib/object-inspect.js.html

erights:
> is this okay as a vetted shim? Are we confident it doesn't reach for powerful things?

@michaelfig , Why would it need to be a vetted shim, rather than unprivileged code running under normal post-lockdown SES restrictions?

michaelfig:
> > is this okay as a vetted shim? Are we confident it doesn't reach for powerful things?
> 
> @michaelfig , Why would it need to be a vetted shim, rather than unprivileged code running under normal post-lockdown SES restrictions?

Sorry, my confusion.  Yes, it's unprivileged code in the post-lockdown start compartment.  I'd like ideally to run it in a compartment, but am unsure about how.

erights:
> Yes, it's unprivileged code in the post-lockdown start compartment. I'd like ideally to run it in a compartment, but am unsure about how.

Excellent. Yes, let's find time to discuss. Thanks!

dckc:
> Sorry, my confusion. Yes, it's unprivileged code in the post-lockdown start compartment.

Really? It looks to me like we run this to build the console that we give to SES:

https://github.com/Agoric/agoric-sdk/blob/17e161e696d7d8538ce1a825960811ebdfa67fa4/packages/xsnap/lib/ses-boot.js#L1-L2

https://github.com/Agoric/agoric-sdk/blob/17e161e696d7d8538ce1a825960811ebdfa67fa4/packages/xsnap/lib/console-shim.js#L3

michaelfig:
> It looks to me like we run this to build the console that we give to SES:

Ahh, yes. The reason it's a vetted shim is because of initialisation ordering in SES, not because we actually need any of that power.

Instinctively, we should be able to relax that constraint and install the object inspector from a post-lockdown compartment.

dckc:
> ... initialisation ordering in SES ...
> 
> Instinctively, we should be able to relax that constraint and install the object inspector from a post-lockdown compartment.

Yes, we used to have a lo-fi console arg quoter that we used for SES initialization, and then we replaced the lo-fi console quoter with `assert.quote`. I guess we don't want to go back to exactly that, but something analagous seems workable...

https://github.com/Agoric/agoric-sdk/blob/a3306d01d8e87c4bc7483a61e42cc30b006feb81/packages/xsnap/lib/console-shim.js
https://github.com/Agoric/agoric-sdk/blob/a3306d01d8e87c4bc7483a61e42cc30b006feb81/packages/xsnap/lib/ses-boot.js

erights:
> Yes, we used to have a lo-fi console arg quoter that we used for SES initialization, and then we replaced the lo-fi console quoter with `assert.quote`. I guess we don't want to go back to exactly that, but something analagous seems workable...
> 
> https://github.com/Agoric/agoric-sdk/blob/a3306d01d8e87c4bc7483a61e42cc30b006feb81/packages/xsnap/lib/console-shim.js https://github.com/Agoric/agoric-sdk/blob/a3306d01d8e87c4bc7483a61e42cc30b006feb81/packages/xsnap/lib/ses-boot.js

TIL `setQuote`. I had no idea. Neat!

packages/xsnap/src/object-inspect.js
warner: Is this meant to guard against some other vetted shim mutating one of the standard objects? I like defensive, but is that still in the threat model?

michaelfig: > Is this meant to guard against some other vetted shim mutating one of the standard objects? I like defensive, but is that still in the threat model?

This is a light refactoring of the original `object-inspect.js` code.  I didn't rewrite it further because I didn't see the need to.

packages/xsnap/src/object-inspect.js
michaelfig: > The justification for `no-use-before-define` is to reduce the diffs from the code we started from?

Not to reduce the diffs, to reduce the porting effort.

FUDCo (approved):
ya shoor okay

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
+/* eslint-disable no-empty */
+/* eslint-disable no-nested-ternary,no-use-before-define */
+// Adapted from object-inspect@1.12.0
```
erights: Please include a link to these original sources.

packages/xsnap/src/object-inspect.js
```diff
@@ -31,7 +31,7 @@ const setSize = Object.getOwnPropertyDescriptor(Set.prototype, 'size').get;
 const setForEach = Set.prototype.forEach;
 const weakMapHas = WeakMap.prototype.has;
 const weakSetHas = WeakSet.prototype.has;
-const weakRefDeref = WeakRef.prototype.deref;
+const weakRefDeref = WeakRef && WeakRef.prototype.deref;
```
erights: Why does this work? If there's no `WeakRef` in scope, this should throw a ReferenceError. I would have expected either
```js
... = globalThis.WeakRef && ...
```
or
```js
... = typeof WeakRef !== 'undefined' && ...
```
I am more curious about why it works now than I am suggesting that you change it. Once I understand, I may ask for that change, but I'd like to understand first.

packages/xsnap/test/test-inspect.js
```diff
@@ -1,47 +1,116 @@
...
-    '[Function (anonymous)]',
+import * as fs from 'fs';
+import * as proc from 'child_process';
+import * as os from 'os';
+import { xsnap } from '../src/xsnap.js';
+import { options } from './message-tools.js';
+
+import unconfinedInspect from '../src/object-inspect.js';
+
+const io = { spawn: proc.spawn, os: os.type() }; // WARNING: ambient
```
erights, line 12: I was momentarily alarmed ;)

But it is in a test and it is commented with a screaming WARNING, so ok.

However, I notice file doesn't start with 
```js
/* global os */
```
and there's no diagnostic about the `os`. Why not?

I also checked in vscode, and there's no red squiggle. Unless there's a reason not to, please add the global comment at the top of the file.

packages/xsnap/test/test-inspect.js
```diff
@@ -1,47 +1,116 @@
...
-  t.is(
-    inspect(() => 'bab'),
-    '[Function (anonymous)]',
+import * as fs from 'fs';
+import * as proc from 'child_process';
+import * as os from 'os';
+import { xsnap } from '../src/xsnap.js';
+import { options } from './message-tools.js';
+
+import unconfinedInspect from '../src/object-inspect.js';
```
erights, line 10: The unconfined one should have `WeakRef` in scope. Given the weirdness of using `deref`, please add a test case for that. I'm curious to see how it renders.

packages/xsnap/test/test-inspect.js
```diff
@@ -1,47 +1,116 @@
...
+async function makeWorker() {
+  const opts = options(io);
+  const vat = xsnap(opts);
+
+  const boot = fs.readFileSync(
+    new URL('../dist/bundle-ses-boot.umd.js', import.meta.url).pathname,
+    'utf8',
   );
 
-  // These are deficiencies in user space introspection of Proxies.
```
erights, line 35: I think this comment as stated is no longer appropriate. Unlike Node's inspect, it is a design goal of our inspect that it has no special privilege. Hmm. I think. Is that a goal of our inspect? If so, please revise the comment so it's not misleading.

packages/xsnap/test/test-inspect.js
erights: Oh. I see that you're using the same testCases for confined and unconfined. So nevermind. I'm not that curious.

erights: Could you also add a promise to testCases? That way, if we get more ambitious later on how promises print, we'll have a before and after.

packages/xsnap/src/object-inspect.js
erights: OH NO. The fact that it works is an artifact of the inability of our shim to implement the proper variable lookup semantics. In the correct semantics, `WeakRef` would throw `ReferenceError` but `typedef WeakRef` would not. But the 8 magic lines of code cannot tell the difference, so we have `WeakRef` return `undefined`, in order to implement the correct `typedef` behavior, in order not to break correct working programs. The too lenient variable lookup behavior would not break (and has not broken) correct working programs. However, it also doesn't catch us when we write bad programs, as here, that will break on the non-shim implementation.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+  }
+  if (depth >= maxDepth && maxDepth > 0 && typeof obj === 'object') {
+    return isArray(obj) ? '[Array]' : '[Object]';
+  }
+
+  const indent = getIndent(opts, depth);
+
+  if (seen.has(obj)) {
+    return '[Circular]';
+  }
```
erights: I don't see how this can be correct. I see only additions to the shared `seen` set, but no deletions. This should also trigger on DAGs, not just cycles. The `bestEffortsStringify` does likewise, but prints something like `[Seen before]`, which may be less useful than cycle detection, but is not misleading.

Please make consistent. I prefer that you do so by fixing the algorithm so that it actually detects cycles, rather than revising the string literal.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+      return '[]';
+    }
+    const xs = arrObjKeys(obj, inspect);
+    if (indent && !singleLineValues(xs)) {
+      return `[${indentedJoin(xs, indent)}]`;
+    }
+    return `[ ${$join.call(xs, ', ')} ]`;
+  }
+  if (isError(obj)) {
+    const parts = arrObjKeys(obj, inspect);
```
erights, line 160: Does this print the stack when there's a `stack` property?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+  if (typeof obj === 'function') {
+    const name = nameOf(obj);
+    const keys = arrObjKeys(obj, inspect);
+    return `[Function${name ? `: ${name}` : ' (anonymous)'}]${
+      keys.length > 0 ? ` { ${$join.call(keys, ', ')} }` : ''
+    }`;
+  }
+  if (isSymbol(obj)) {
+    const symString = symToString.call(obj);
+    return typeof obj === 'object' ? markBoxed(symString) : symString;
```
erights: `typeof obj === 'object'` cannot possibly be true here, because isSymbol tests that `typeof obj === 'symbol'`.

I'd prefer that registered and unregistered symbols render differently, unless there's a reason not to.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+  }
+  if (isArray(obj)) {
+    if (obj.length === 0) {
+      return '[]';
+    }
+    const xs = arrObjKeys(obj, inspect);
```
erights: Please rename the variable `xs`. Given the context, it is confusing.

packages/xsnap/src/object-inspect.js
mhofman: In a locked down environment this shouldn't be necessary. If we do want to adopt this style for this file, maybe we could enable the eslint rule to disable polymorphic calls like the one endo uses: https://github.com/endojs/endo/pull/827

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+// ie, `has-tostringtag/shams
+const toStringTag = Symbol.toStringTag;
+const isEnumerable = Object.prototype.propertyIsEnumerable;
+const dateToISOString = Date.prototype.toISOString;
+
+const gPO = Reflect.getPrototypeOf;
+const hasOwn = Object.prototype.hasOwnProperty;
+
+export default function inspect0(obj, opts = {}, depth = 0, seen = new Set()) {
+  if (typeof obj === 'undefined') {
```
mhofman: Small nit: any reason not to remember `const type = typeof obj` ?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+    return `\\${x}`;
+  }
+  return `\\x${n < 0x10 ? '0' : ''}${$toUpperCase.call(n.toString(16))}`;
+}
+
+function markBoxed(str) {
+  return `Object(${str})`;
+}
+
+function weakCollectionOf(type) {
```
erights: Since you're also using this for `WeakRef`, please rename to `weakContainerOf` or anything that can correctly correspond to `WeakMap`, `WeakSet`, and `WeakRef`.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+const symToString = Symbol.prototype.toString;
+// ie, `has-tostringtag/shams
+const toStringTag = Symbol.toStringTag;
+const isEnumerable = Object.prototype.propertyIsEnumerable;
+const dateToISOString = Date.prototype.toISOString;
+
+const gPO = Reflect.getPrototypeOf;
+const hasOwn = Object.prototype.hasOwnProperty;
+
+export default function inspect0(obj, opts = {}, depth = 0, seen = new Set()) {
```
mhofman: Nit: `s/obj/val`

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+    return String(obj);
+  }
+  if (typeof obj === 'bigint') {
+    return `${obj}n`;
+  }
+
+  const maxDepth = typeof opts.depth === 'undefined' ? 5 : opts.depth;
+  if (typeof depth === 'undefined') {
+    depth = 0;
+  }
```
mhofman: We're already setting a default above, aren't we?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+  if (isBoolean(obj)) {
+    return markBoxed(booleanValueOf.call(obj));
+  }
+  if (isString(obj)) {
+    return markBoxed(inspect(String(obj)));
+  }
+  if (isDate(obj)) {
+    return dateToISOString.call(obj);
+  }
+  if (!isDate(obj) && !isRegExp(obj)) {
```
erights: Cannot possibly be a Date here.

erights: Why not do 
```js
if (isRegExp(obj)) {
  return String(obj);
}
```
? 

It would be perfectly consistent with the rest of their code. If you want to minimize gratuitous changes, you can leave it alone. But does it seem as weird to you as it does to me?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+    depth = 0;
+  }
+  if (depth >= maxDepth && maxDepth > 0 && typeof obj === 'object') {
+    return isArray(obj) ? '[Array]' : '[Object]';
+  }
+
+  const indent = getIndent(opts, depth);
+
+  if (seen.has(obj)) {
+    return '[Circular]';
```
mhofman, line 109: Can't duplicated values can also trigger this. E.g.

```js
const obj = {};

inspect0({foo: obj, bar: obj});
```

packages/xsnap/src/object-inspect.js
erights: Nevermind. There's also a `ys`. Maybe just a comment?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,491 @@
...
+      inspectString($slice.call(str, 0, opts.maxStringLength), opts) + trailer
+    );
+  }
+  // eslint-disable-next-line no-control-regex
+  const s = $replace.call(
+    $replace.call(str, /(['\\])/g, '\\$1'),
+    // eslint-disable-next-line no-control-regex
+    /[\x00-\x1f]/g,
+    lowbyte,
+  );
```
erights: I could not puzzle it out. Please add a brief explanatory comment.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+  if (typeof obj === 'function') {
+    const name = nameOf(obj);
+    const keys = arrObjKeys(obj, inspect);
+    return `[Function${name ? `: ${name}` : ' (anonymous)'}]${
+      keys.length > 0 ? ` { ${$join.call(keys, ', ')} }` : ''
+    }`;
+  }
+  if (isSymbol(obj)) {
+    const symString = symToString.call(obj);
+    return typeof obj === 'object' ? markBoxed(symString) : symString;
```
mhofman: I don't see how `typeof obj === 'object'` is ever possible given the definition of `isSymbol`

erights (changes requested):
Lots of minor nits but one genuine bug -- where an implementation limitation of the SES shim led you astray.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+      return `[${indentedJoin(xs, indent)}]`;
+    }
+    return `[ ${$join.call(xs, ', ')} ]`;
+  }
+  if (isError(obj)) {
+    const parts = arrObjKeys(obj, inspect);
+    if ('cause' in obj && !isEnumerable.call(obj, 'cause')) {
+      return `{ [${String(obj)}] ${$join.call(
+        $concat.call(`[cause]: ${inspect(obj.cause)}`, parts),
+        ', ',
```
mhofman: why not ```parts.unshift(`[cause]: ${inspect(obj.cause)}`)``` and let it fall through?
Also `$concat.call(string, arr)` seems wrong.

packages/xsnap/src/object-inspect.js
mhofman: Seeing the `isNumber` and co below, it seems the intent is to detect boxed symbols, but apparently there is no good way to do this: https://github.com/inspect-js/is-symbol/blob/main/index.js

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+  if (isString(obj)) {
+    return markBoxed(inspect(String(obj)));
+  }
+  if (isDate(obj)) {
+    return dateToISOString.call(obj);
+  }
+  if (!isDate(obj) && !isRegExp(obj)) {
+    const ys = arrObjKeys(obj, inspect);
+    const isPlainObject = gPO
+      ? gPO(obj) === Object.prototype
```
mhofman: Most definitions consider null protos to be plain objects

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+      : obj instanceof Object || obj.constructor === Object;
+    const protoTag = obj instanceof Object ? '' : 'null prototype';
+    const stringTag =
+      !isPlainObject && toStringTag && Object(obj) === obj && toStringTag in obj
```
mhofman: I'm confused by this late `Object(obj) === obj` check, especially we code above already does things considering `obj` is in fact an object (or at least can implicitly be converted to one)

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+        ? $slice.call(toStr(obj), 8, -1)
+        : protoTag
+        ? 'Object'
+        : '';
+    const constructorTag =
+      isPlainObject || typeof obj.constructor !== 'function'
+        ? ''
+        : obj.constructor.name
+        ? `${obj.constructor.name} `
+        : '';
```
mhofman: This would use own `constructor` prop, including for `null` proto objects, which I find quite weird.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+}
+
+// Symbol and BigInt do have Symbol.toStringTag by spec, so that can't be used to eliminate false positives
+function isSymbol(obj) {
+  return typeof obj === 'symbol';
+}
+
+function isBigInt(obj) {
+  return typeof obj === 'bigint';
+}
```
mhofman, line 295: The above `isFoo` seem targeted at finding the boxed foo primitives, but somehow this is only looking for the primitives themselves? I'd consider removing these helpers and rolling the primitives detection directly along the other ones (`bigint` is already there, but `symbol` isn't)

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+  } catch (e) {}
+  return false;
+}
+
+function isWeakRef(x) {
+  if (!weakRefDeref || !x || typeof x !== 'object') {
+    return false;
+  }
+  try {
+    weakRefDeref.call(x);
```
mhofman, line 365: This is sad as it forces the target to stay alive another turn

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+}
+
+function inspectString(str, opts) {
+  if (str.length > opts.maxStringLength) {
+    const remaining = str.length - opts.maxStringLength;
+    const trailer = `... ${remaining} more character${
+      remaining > 1 ? 's' : ''
+    }`;
+    return (
+      inspectString($slice.call(str, 0, opts.maxStringLength), opts) + trailer
```
mhofman, line 410: technically this can cut in the middle of a surrogate pair.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+  const lineJoiner = `\n${indent.prev}${indent.base}`;
+  return `${lineJoiner + $join.call(xs, `,${lineJoiner}`)}\n${indent.prev}`;
+}
+
+function arrObjKeys(obj, inspect) {
+  const isArr = isArray(obj);
+  const xs = [];
+  if (isArr) {
+    xs.length = obj.length;
+    for (let i = 0; i < obj.length; i += 1) {
```
mhofman, line 492: is there an iteration max? worried about large sparse arrays blowing up the output.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+      xs[i] = has(obj, i) ? inspect(obj[i], obj) : '';
+    }
+  }
+  const syms = getOwnPropertySymbols(obj);
+  for (const key in obj) {
+    if (!has(obj, key)) {
+      // eslint-disable-next-line no-continue
+      continue;
+    }
```
mhofman: Why not do a `Object.getOwnPropertyNames` instead ?

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,489 @@
...
+    if (isArr && String(Number(key)) === key && key < obj.length) {
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+    if ($test.call(/[^\w$]/, key)) {
+      xs.push(`${inspect(key, obj)}: ${inspect(obj[key], obj)}`);
```
mhofman: Confused why `obj` is included in `${inspect(key, obj)}` but is not in the symbol version below: `${inspect(syms[j])}`

packages/xsnap/test/test-inspect.js
dckc: `os` is not a global; it's imported using `import * as os from 'os';`.

erights: Gotcha. Nevermind.

packages/xsnap/src/object-inspect.js
erights: Good point!

packages/xsnap/src/object-inspect.js
erights: The `for/in` has another problem: it does not have a deterministic standard semantics, in the case where the object and/or its prototype chain are mutated during the iteration. This case is known to differ between engines, and is so highly optimized on each that it has been impossible to get the engines to agree on any deterministic spec.

That's not a killer for us. We live with several of these static spec indeterminacies. In the short term, by requiring all validators to run the same engine. In the long term, by writing down a "Deterministic JS" spec that's a refinement[1] of the JS spec. Or better, by skipping that and going directly to writing a Deterministic Hardened JS spec that's a refinement of the hardened JS spec. (We still must keep the distinction. Hardened JS will continue to run on Node. Deterministic Hardened JS never will.)

[1] "refinement" meaning that anything that conforms to the refined spec also conforms to the original spec, but not vice versa. Everything the refined spec requires, the original spec allows.

Nevertheless, all that said, it is worth some trouble to get rid of all `for/in`s in our sources. This one enumerates inherited and own enumerable string-based properties, and then filters that down to own properties.

Note that `Object.getOwnPropertyNames` is not a semantics preserving replacement, in that it will include non-enumerable string-named own properties as well.

warner:
y'all are reviewing this better than I could possibly do, I'll defer to you

packages/xsnap/src/object-inspect.js
erights: See https://github.com/endojs/endo/issues/1136

michaelfig: Yeah, I was numbed by the code transformation.  Normally I wouldn't make this mistake, so thanks for pointing it out.

packages/xsnap/src/object-inspect.js
michaelfig: Definitely not.  The stack printing happens only in the causal console.

packages/xsnap/src/object-inspect.js
michaelfig: I don't know enough about this code to comment.  I'll leave it as-is unless you have better advice.

packages/xsnap/src/object-inspect.js
michaelfig: I've extracted the prototype's constructor property instead.

packages/xsnap/src/object-inspect.js
michaelfig: I don't want to remove them quite yet.  That may be a useful cleanup, but it's more than I want to take on in this PR.

packages/xsnap/src/object-inspect.js
michaelfig: Anything I should try to do about that?

packages/xsnap/src/object-inspect.js
michaelfig: The symbol cannot possibly be the same as the object, so no need to check it for circularity.

packages/xsnap/src/object-inspect.js
michaelfig: I used `getOwnPropertyNames` and `isEnumerable` to check.

packages/xsnap/src/object-inspect.js
michaelfig: Large anything will blow up the output, and cause a metering violation.  No problem there.

packages/xsnap/src/object-inspect.js
michaelfig: Not Node.js `console.log`, though.

packages/xsnap/src/object-inspect.js
michaelfig: Anything I should do about that?

packages/xsnap/src/object-inspect.js
mhofman: I don't see another way. `deref` is the only thing that brand checks the `[[WeakRefTarget]]` slot.

packages/xsnap/src/object-inspect.js
mhofman: But the symbol's value can refer to the object, can't it?

packages/xsnap/src/object-inspect.js
mhofman: We add a trailer which starts with a non surrogate char, so it may just technically be an invalid string, but all implementations handle malformed strings like that.

packages/xsnap/src/object-inspect.js
michaelfig: > But the symbol's value can refer to the object, can't it?

As far as I read it, the `inspect(syms[j])` inspects the symbol itself, not the symbol property's value.

packages/xsnap/lib/confined-object-inspect.js
```diff
@@ -0,0 +1,15 @@
+import objectInspectSources from '../dist/src-object-inspect.js';
+
+// Ensure the object inspector is confined.
+const c = new Compartment();
+harden(c.globalThis);
+
+const src = objectInspectSources.replace(
+  /(^|\s)(export\s+default)(\s+)/g,
+  '$1/* $2 */$3',
+);
```
erights, line 19: OMG. Definitely worth a comment to explain why.

packages/xsnap/src/object-inspect.js
erights: @mhofman I don't see how that trailer helps. In the case in question, it would already be invalid as a unicode string.

packages/xsnap/test/test-inspect.js
erights: yes, deletion is fine.

packages/xsnap/src/object-inspect.js
```diff
@@ -0,0 +1,520 @@
...
+    const keys = arrObjKeys(obj, inspect);
+    return `[Function${name ? `: ${name}` : ' (anonymous)'}]${
+      keys.length > 0 ? ` { ${$join.call(keys, ', ')} }` : ''
+    }`;
+  }
+  if (isSymbol(obj)) {
+    const registered = symKeyFor(obj);
+    if (registered !== undefined) {
+      // Registered symbol.
+      return `Symbol.for(${registered})`;
```
erights, line 145: Thanks!

packages/xsnap/src/object-inspect.js
erights: Yes, that helped. Thanks.

erights (approved):
LGTM, thanks!

packages/xsnap/src/object-inspect.js
mhofman: I think what I meant is that this won't be concatenated with another string which may start with the second half of another surrogate pair, so if we split in the middle, the first half will just be detected as a broken pair, and processed as such by every algorithms which handles strings.

erights: Thanks. I understand now, and it is a valid concern.

packages/xsnap/lib/confined-object-inspect.js
michaelfig: Definitely!

Issue #4663, opened 2022-02-24
Parity Stability Module to allow direct trading of stable currencies with RUN


dckc:
## What is the Problem Being Solved?

We don't currently have support for trading stable currencies in our AMM. An interim approach would be to allow them to be traded using the x*y=k curves we currently have, but that means we should think about the upgrade path to move them to something better.

## Description of the Design

Various possibilities were discussed (@Chris-Hibbert @dtribble, folks from RMIT) in a meeting on Feb. 23.

Progress is being made on a stable swap mechanism that would allow various stable coins to trade against one another with low slippage.

There's a plausible design idea for integrating this with the AMM. When a trade is proposed between two coins participating in the stable curve, it's treated as a single pool by the current virtualPool framework. If a trade is proposed between a stable coin and something else supported in the AMM, it's treated as a doublePool trade (from X to RUN and RUN to Y). I think pool fees are charged inside each pool and protocol fees are charged outside, so this should continue to work.

It would be nice if we could declare that some stable coins (like USDC) are stable enough that we can issue unlimited RUN against them. In this case, we need another contract that accepts USDC (perhaps up to some limit) and mints new RUN against it. It would also be able to trade in the reverse direction, up to the amount of USDC it holds or some lower limit we may impose. To the limit of its balances, this would ensure parity between RUN and ultra stable tokens.

Another idea we considered was to use the stable curve implementation to support trading among various mutually stable currencies, like WETH and other bridged ETH equivalents. It's not  obvious how to integrate this into the AMM, which uses RUN to bridge between other currency pairs, but this might be a useful additional institution.

## Security Considerations

The implications are more relevant to economic stability than to software security.

## Test Plan

the current state isn't clear enough to speculate about testing yet.

Chris-Hibbert:
btulloh wrote on Keybase: 
> These are really two separate things (one with two applications) that should be tracked separately. The first and third elements are having a stableswap curve available in the AMM for exchanging like (correlated) assets. The second element is not related to the AMM, but is what MakerDAO calls the Peg Stability Module. This would be a separate contract from the AMM and be managed by the RUN protocol. We need both for different reasons.

I agree. The paragraphs starting "There's a plausible design" and "Another idea" are talking about applying the "stable swap mechanism" discussed earlier in two different applications. The paragraph starting "It would be nice" concerns the idea of a PSM.

Chris-Hibbert:
Renamed to just cover the middle idea. 

[...]

PR #4786, opened 2022-03-09
Create Initial Parity Stability Module

dtribble:
refs: #4663

## Description

Initial cut at a Parity Stability Module: 
* provide an anchoring token (e.g., like USDC) to mint a local stable token (e.g., RUN), and vice versa
* charge different fees for trading into and out of the stable token
* enforce an external minting limit, by limiting the size of the pool of anchor

### Security Considerations

It must enforce a minting limit for stable.

### Testing Considerations

Key test cases:
* give `anchor`, get `stable`
  * fees are charged
  * the pool balance is updated
* give `stable`, get `anchor`
  * fees are charged
  * the pool balance is updated
* from a non-zero pool, give `anchor` to overflow the pool
  * offer is rejected
  * the pool is not changed


packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+
+  const giveStable = (seat, given, wanted = emptyAnchor) => {
+    const fee = ceilMultiplyBy(given, gov.getGiveStableRate());
+    const afterFee = AmountMath.subtract(given, fee);
+    const maxAnchor = AmountMath.make(anchorBrand, afterFee.value);
+    // TODO this prevents the reallocate from failing. Can this be tested otherwise?
+    assert(
+      AmountMath.isGTE(maxAnchor, wanted),
+      X`wanted ${wanted} is more then ${given} minus fees ${fee}`,
+    );
```
Chris-Hibbert, line 119: Why is an assert better than letting the reallocate fail? I think the assert is extraneous.

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+      given,
+    );
+    assertUnderLimit(anchorAfterTrade);
+    const asStable = AmountMath.make(stableBrand, given.value);
+    const fee = ceilMultiplyBy(asStable, gov.getWantStableRate());
+    const afterFee = AmountMath.subtract(asStable, fee);
+    assert(
+      AmountMath.isGTE(afterFee, wanted),
+      X`wanted ${wanted} is more then ${given} minus fees ${fee}`,
+    );
```
Chris-Hibbert, line 135: reallocate will verify this.

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+    stableMint.burnLosses({ Stable: afterFee }, stage);
+  };
+
+  const wantStable = (seat, given, wanted = emptyStable) => {
+    const anchorAfterTrade = AmountMath.add(
+      anchorPool.getAmountAllocated('Anchor', anchorBrand),
+      given,
+    );
+    assertUnderLimit(anchorAfterTrade);
+    const asStable = AmountMath.make(stableBrand, given.value);
```
Chris-Hibbert:
```suggestion
    const asStable = AmountMath.coerce(stableBrand, given);
```

If `given` isn't trustworthy (it is, Zoe has checked it) then `anchorAfterTrade`  should use asStable rather than given.  If `given` has the wrong brand, the arithmetic and reallocation will fail.

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+    } else {
+      throw Error(`unexpected brand ${given.brand}`);
+    }
+    seat.exit();
+  };
+  const makeSwapInvitation = () => zcf.makeInvitation(swapHook, 'swap');
+
+  // Eventually the reward pool will live elsewhere. For now it's here for
+  // bookkeeping. It's needed in tests.
+  const getRewardAllocation = () => feePool.getCurrentAllocation();
```
Chris-Hibbert: `vaultFactory/collectRewardFees` shows how it is made accessible to the fee collector.  Any reason not to incorporate that code now?

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+  const { brand: stableBrand } = stableMint.getIssuerRecord();
+  zcf.setTestJig(() => ({
+    stableIssuerRecord: stableMint.getIssuerRecord(),
+  }));
+  const emptyStable = AmountMath.makeEmpty(stableBrand);
+  const emptyAnchor = AmountMath.makeEmpty(anchorBrand);
+
+  // HACK for simple governance API
+  const getGovernance = _ => {
+    const {
```
Chris-Hibbert: This is a fine temporary stand-in for governance, but I think it ought to use governance before merging.

would you like someone to add governance?

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * @param {Amount<'nat'>} given
+ * @param {Brand} stableBrand
+ * @returns {Amount<'nat'>}
+ */
+const minusFee = (given, stableBrand) => {
+  const feeBP = given.brand === stableBrand ? GiveStableFeeBP : WantStableFeeBP;
+  return AmountMath.make(
+    stableBrand,
+    BigInt((given.value * (BASIS_POINTS - feeBP)) / BASIS_POINTS),
```
Chris-Hibbert: If this used NatSafeMath and Ratios, I wouldn't worry that future tests might pass in `given`s that were small enough that this wouldn't be reliable.

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -0,0 +1,190 @@
...
+const setJig = _jig => {
+  // testJig = jig;
+};
+
+/**
+ * @param {ExecuteContract} t
+ */
+test.before(async t => {
+  // makeBundle is slow, so we bundle each contract once and reuse in all tests.
+  const [psmBundle] = await Promise.all([makeBundle(psmRoot)]);
```
Chris-Hibbert: does Promise.all add anything here?
```suggestion
  const psmBundle = await makeBundle(psmRoot);
```

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+    seat.exit();
+  };
+  const makeSwapInvitation = () => zcf.makeInvitation(swapHook, 'swap');
+
+  // Eventually the reward pool will live elsewhere. For now it's here for
+  // bookkeeping. It's needed in tests.
+  const getRewardAllocation = () => feePool.getCurrentAllocation();
+
+  const publicFacet = Far('Parity Stability Module', {
+    getCurrentLiquidity: () =>
```
Chris-Hibbert: I don't think I'd call this 'Liquidity'. I think 'collateral', 'balance', or 'reserves' would make more sense to me.

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -0,0 +1,190 @@
...
+    harden({ give: { In: give } }),
+    harden({ In: anchorMint.mintPayment(give) }),
+  );
+  const runPayout = await E(seat1).getPayout('Out');
+  const expectedRun = minusFee(give, runBrand);
+  const actualRun = await E(runIssuer).getAmountOf(runPayout);
+  t.deepEqual(actualRun, expectedRun);
+  const liq = await E(publicFacet).getCurrentLiquidity();
+  t.true(AmountMath.isEqual(liq, give));
+  trace('get stable', { give, expectedRun, actualRun, liq });
```
Chris-Hibbert: I think once the test is working, this should be dropped.  (Unless we're going to give `trace` a toggle.)

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+import { Far } from '@endo/marshal';
+// import { CONTRACT_ELECTORATE } from '@agoric/governance';
+
+import { AmountMath } from '@agoric/ertp';
+import { makeMakeCollectFeesInvitation } from '../vaultFactory/collectRewardFees.js';
+
+const { details: X } = assert;
+
+const BASIS_POINTS = 10000n;
+
```
Chris-Hibbert, line 30: Please add a top comment explaining the point of the contract. We usually also document the keywords in use and the main invitations. It will be much easier fro someone to add contract docs to our documentation site if we have at least a rudimentary description here.

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,170 @@
...
+  to.incrementBy(txTo);
+}
+
+/**
+ * @param {ZCF<{
+ *    anchorBrand: Brand,
+ *    main: {
+ *      WantStableFeeBP: bigint,
+ *      GiveStableFeeBP: bigint,
+ *      MintLimit: Amount } }>
```
Chris-Hibbert, line 53: I think we also need a multiplication factor here. Even if we're trading at "par", the different currencies may have a different number of decimal places.

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -0,0 +1,190 @@
...
+  const give = mintLimit;
+  const seat1 = await E(zoe).offer(
+    E(publicFacet).makeSwapInvitation(),
+    harden({ give: { In: give } }),
+    harden({ In: anchorMint.mintPayment(give) }),
+  );
+  trace('gone over limit');
+
+  const paymentPs = await E(seat1).getPayouts();
+  // const refundAmount = await E(localIssuerP).getAmountOf(paymentPs.Transfer);
```
Chris-Hibbert: use it or delete it.

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -0,0 +1,190 @@
...
+
+  trace('PAYOUTS', paymentPs);
+  // We should get 0 RUN but all our anchor back
+  // TODO should this be expecteed to be an empty Out?
+  t.falsy(t.Out);
+  // const actualRun = await E(runIssuer).getAmountOf(runPayout);
+  // t.deepEqual(actualRun, AmountMath.makeEmpty(runBrand));
```
Chris-Hibbert, line 217: please uncomment.

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -0,0 +1,190 @@
...
+    minusFee(runGive, runBrand).value,
+  );
+  t.deepEqual(actualAnchor, expectedAnchor);
+  const liq2 = await E(publicFacet).getCurrentLiquidity();
+  t.deepEqual(AmountMath.subtract(give, expectedAnchor), liq2);
+  trace('get anchor', { runGive, expectedRun, actualAnchor, liq2 });
+});
+
+test('limit', async t => {
+  const {
```
Chris-Hibbert, line 180: This tests trying to extract more than has been added. I also want to see trying to get more than the `mintLimit`

packages/run-protocol/src/psm/psm.js
dtribble: I wanted to check before minting to avoid trivially causing mint/burn efforts.

packages/run-protocol/src/psm/psm.js
dtribble: I like 'balance' and 'reserve'. It might make sense to change all the "anchor" names to "reserve" as well, since that's the reserve asset for this contract.

packages/run-protocol/test/psm/test-psm.js
dtribble: It's to keep the pattern from other tests that use multiple bundles. But can be deleted.

packages/run-protocol/test/psm/test-psm.js
dtribble: It's contributing more than `mintLimit` in `anchor` (line 168), so that is what it's testing.

packages/run-protocol/test/psm/test-psm.js
dtribble: This is related to the TODO.  I would love to uncomment, but then it fails because `getAmountOf` fails internally. That's because `t.Out` is undefined (as in line 178). I'm leaving the comment until you and Mark see that and agree that it's expected. Because it surprised me!

packages/run-protocol/src/psm/psm.js
dtribble: This is just because it was paralleling the other branch, which used the assert to avoid a spurious minting event.

turadg: Either way, WDYT of a `@throws` comment on the function jsdoc to describe?

packages/run-protocol/src/psm/psm.js
dtribble: This is doing a 1:1 conversion from USDC to RUN. Your point earlier that this needs to use a price even if it is 1:1 because they might have different decimals is spot on though.  So this needs to change to a multiply.

packages/run-protocol/src/psm/psm.js
dtribble: Will do.

packages/run-protocol/src/psm/psm.js
dtribble: Governance APIs are still moving. I don't think it should be blocked on that. It will need that before it could be turned on, but there's multiple concurrent streams of work that could continue on this (governance and bootstrap and making it durable).

packages/run-protocol/src/psm/psm.js
dtribble: I shall avoid @throws both because of Java PTSD (pervasive throw system disorders) and because we are still debating exactly how to abruptly terminate.

packages/run-protocol/src/collectFees.js
```diff
@@ -0,0 +1,31 @@
+// @ts-check
+/**
+ * Provide shared support for providing access to fees from a service contract.
+ *
+ * @param {ZCF} zcf
+ * @param {ZCFSeat} feeSeat
+ * @param {Brand} feeBrand
+ * @param {string} keyword
+ * @returns {{makeCollectFeesInvitation: (function(): *)}}
```
Chris-Hibbert:
```suggestion
 * @returns {{makeCollectFeesInvitation: () => Invitation}}
```

packages/run-protocol/src/collectFees.js
```diff
@@ -0,0 +1,31 @@
...
+export const makeMakeCollectFeesInvitation = (
+  zcf,
+  feeSeat,
+  feeBrand,
+  keyword,
+) => {
+  const collectFees = seat => {
+    const amount = feeSeat.getAmountAllocated(keyword, feeBrand);
+    feeSeat.decrementBy(harden({ [keyword]: amount }));
+    seat.incrementBy(harden({ RUN: amount }));
```
Chris-Hibbert, line 19: Didn't quite get rid of all the `RUN`. Should this be `[keyword]` as well?

packages/run-protocol/src/makeTracer.js
```diff
@@ -1,13 +1,28 @@
...
+      const logDisabled = (..._args) => {};
+      return logDisabled;
+    }
+    case 'info': {
+      const infoTick = (...args) => {
+        console.info(key, (debugCount += 1), ...args);
+      };
+      return infoTick;
+    }
+    default: {
```
Chris-Hibbert, line 19: If you don't expect info and default to diverge, these should use the same code. If you do expect them to diverge, how would they differ?

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,178 @@
...
+      wantStable(seat, given, wanted);
+    } else {
+      throw Error(`unexpected brand ${given.brand}`);
+    }
+    seat.exit();
+  };
+  const makeSwapInvitation = () => zcf.makeInvitation(swapHook, 'swap');
+
+  // Eventually the reward pool will live elsewhere. For now it's here for
+  // bookkeeping. It's needed in tests.
```
Chris-Hibbert: I think this comment is now extraneous.

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -188,3 +207,51 @@ test('limit', async t => {
...
+    harden({ In: anchorMint.mintPayment(giveAnchor) }),
+  );
+  const runPayout = await E(seat1).getPayout('Out');
+  const expectedRun = minusAnchorFee(giveAnchor, anchorPerStable);
+  const actualRun = await E(runIssuer).getAmountOf(runPayout);
+  t.deepEqual(actualRun, expectedRun);
+  const liq = await E(publicFacet).getCurrentLiquidity();
+  t.true(AmountMath.isEqual(liq, giveAnchor));
+  const giveRun = AmountMath.make(runBrand, 100n * 1_000_000n);
+  trace('get stable ratio', { giveRun, expectedRun, actualRun, liq });
```
Chris-Hibbert, line 258: I don't think it makes sense to leave `trace` statements in tests. If they're important enough to leave around, they should be converted to assertions.

packages/run-protocol/src/psm/psm.js
dtribble: I added it to terms rather than governance

packages/run-protocol/src/collectFees.js
dtribble: No I was deliberately taking from an application-specified keyword and dumping them into the 'RUN' keyword on the rewards seat so that all rewards end up on the same keyword (regardless of the internal keyword used).  Just solving one problem at a time.

packages/run-protocol/src/makeTracer.js
dtribble: `info` uses `console.info`; `default` uses `console.log`. So Info I believe is normally not on.

packages/run-protocol/test/psm/test-psm.js
dtribble: They should be using the logging that only turns on if the test fails. They aren't asserts. They are informational.

packages/run-protocol/src/makeTracer.js
turadg: > So Info I believe is normally not on

It's often disabled but Chrome, for example, defaults `info` to on:
Screen Shot 2022-03-24 at 4 15 49 PM

> info uses console.info; default uses console.log

In Node, console.info is an alias for console.log.  ([ref](https://nodejs.org/api/console.html#consoleinfodata-args))

Maybe this will all get sorted out in the emerging [Console Standard](https://console.spec.whatwg.org/#loglevel-severity).

packages/run-protocol/src/psm/psm.js
dtribble: I believe that it should not finish governance before merging. Governance raises additional questions, tests cases, bootstrap integration issues, etc. Similarly, it should not wait for integrating durability. Those are orthogonal features that can be integrated independently.

packages/run-protocol/src/psm/psm.js
Chris-Hibbert: The comment is no longer needed true, since we're using collectFeesInvitation.

packages/run-protocol/src/collectFees.js
Chris-Hibbert: updating this type will be useful to callers.

packages/run-protocol/src/psm/psm.js
Chris-Hibbert: Were you going to change the names, or were you waiting for more discussioni?

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+      getMintLimit: () => MintLimit,
+    };
+  };
+  const gov = getGovernance(zcf);
+
+  const { zcfSeat: anchorPool } = zcf.makeEmptySeatKit();
+  const { zcfSeat: feePool } = zcf.makeEmptySeatKit();
+  const { zcfSeat: stage } = zcf.makeEmptySeatKit();
+
+  const assertUnderLimit = anchorAmount => {
```
Chris-Hibbert: Is this intended to prevent minting that adds up to more than mint limit, or just individual transactions that are larger than the limit? It seems to be doing the latter, and I'm not sure why that's useful. As a client, I can make multiple requests to stay under the limit.

packages/run-protocol/src/collectFees.js
turadg: We could also just drop the `@returns`. It's implied by what's actually returned.

packages/run-protocol/src/makeTracer.js
```diff
@@ -1,13 +1,28 @@
 let debugInstance = 1;
 
-const makeTracer = name => {
+const makeTracer = (name, enable = true) => {
```
turadg, line 3: if we're going to have this option, `enabled = false` makes more sense to me. Bigger change but this change is already out of scope ;-)

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * @file The Parity Stability Module supports efficiently minting/burning a
+ * stable token at a specified fixed ratio to a reference stable token, which
+ * thereby acts as an anchor to provide additional stability. For flexible
+ * economic policies, the fee percentage for trading into and out of the stable
+ * token are specified separately.
+ *
+ */
+
+const BASIS_POINTS = 10000n;
```
turadg: this is defined in 13 places. now may be a good time to pick a place to keep it.

also `10_000n` for readibility.

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Stage a transfer of a singe asset from one seat to another, with an optional
```
turadg: s/singe/single

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+ * remapping of the Keywords. Check that the remapping is for the same amount.
+ *
+ * @param {ZCFSeat} from
+ * @param {ZCFSeat} to
+ * @param {AmountKeywordRecord} txFrom
+ * @param {AmountKeywordRecord} txTo
```
turadg, line 38: `AmountKeywordRecord` can have any number of keywords. Consider a tuple: `[Keyword, Amount]`

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+ */
+function stageTransfer(from, to, txFrom, txTo = txFrom) {
+  assert(AmountMath.isEqual(Object.values(txFrom)[0], Object.values(txTo)[0]));
```
turadg, line 41: If you are going to stick with an `AmountKeywordRecord` parameter, check that it has exactly one keyword/amount.

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+ *    anchorPerStable: Ratio,
+ *    main: {
+ *      WantStableFeeBP: bigint,
+ *      GiveStableFeeBP: bigint,
+ *      MintLimit: Amount } }>
+ * } zcf
+ * @param {{feeMintAccess: FeeMintAccess}} privateArgs
+ */
+export const start = async (zcf, privateArgs) => {
+  const { anchorBrand, anchorPerStable } = zcf.getTerms();
```
turadg, line 58: Isn't `anchorBrand` term redundant with `anchorPerStable`? 
```suggestion
  const { anchorPerStable } = zcf.getTerms();
  const anchorBrand = anchorPerStable.numerator.brand;
```

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+
+  const { feeMintAccess } = privateArgs;
+  // TODO should this know that the name is 'Stable'
+  // TODO get the RUN magic out of here so the contract is more reusable
+  const stableMint = await zcf.registerFeeMint('Stable', feeMintAccess);
+  const { brand: stableBrand } = stableMint.getIssuerRecord();
```
turadg, line 64: worth checking,
```
assert(keyEq(stableBrand, anchorPerStable.denominator.brand));
```

packages/run-protocol/src/psm/psm.js
turadg: Fair. Though "HACK for simple governance API" sounds like it's got governance. Suggestion "XXX mock governance API for terms that will be governed"

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+  const emptyAnchor = AmountMath.makeEmpty(anchorBrand);
+
+  // HACK for simple governance API
+  const getGovernance = _ => {
+    const {
+      main: { WantStableFeeBP, GiveStableFeeBP, MintLimit },
+    } = zcf.getTerms();
+    assert(
+      AmountMath.isGTE(MintLimit, emptyAnchor),
+      X`MintLimit ${MintLimit} must specificy a limit in ${anchorBrand}`,
```
turadg: too icy 🥶

s/specificy/specify

packages/run-protocol/src/psm/psm.js
turadg: What's the status of resolving that debate?

When it's resolved, how will we know to come back and make this operate that way?

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+
+  // Eventually the reward pool will live elsewhere. For now it's here for
+  // bookkeeping. It's needed in tests.
+  const getRewardAllocation = () => feePool.getCurrentAllocation();
+
+  const publicFacet = Far('Parity Stability Module', {
+    getCurrentLiquidity: () =>
+      anchorPool.getAmountAllocated('Anchor', anchorBrand),
+    makeSwapInvitation,
+    // getContractGovernor: () => electionManager,
```
turadg: stray

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,187 @@
...
+  const { makeCollectFeesInvitation } = makeMakeCollectFeesInvitation(
+    zcf,
+    feePool,
+    stableBrand,
+    'Stable',
+  );
+  const creatorFacet = Far('Parity Stability Module', {
+    getRewardAllocation,
+    makeCollectFeesInvitation,
+    // getContractGovernor: () => electionManager,
```
turadg: stray. if this is meant to be left for future work, leave a comment. (though I think it's better to remove)

packages/run-protocol/src/collectFees.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Provide shared support for providing access to fees from a service contract.
+ *
+ * @param {ZCF} zcf
+ * @param {ZCFSeat} feeSeat
+ * @param {Brand} feeBrand
+ * @param {string} keyword
+ * @returns {{makeCollectFeesInvitation: (function(): *)}}
+ */
+export const makeMakeCollectFeesInvitation = (
```
turadg, line 10: `makeMake` 🙄

turadg (commented):
Took a pass at the code level, not economics.

packages/run-protocol/test/psm/test-psm.js
dtribble: I switched to natSafeMath, but it seems on the edge of the obscurity being worse that not having the checking: 
```
    NatMath.floorDivide(
      NatMath.multiply(stable.value, NatMath.subtract(BASIS_POINTS, feeBP)),
      BASIS_POINTS,
    ),
```
vs
```
 BigInt((stable.value * (BASIS_POINTS - feeBP)) / BASIS_POINTS),
```

packages/run-protocol/src/psm/psm.js
dtribble: The git UI had several items visually collapsed so I missed them.  Fixed.

packages/run-protocol/test/psm/test-psm.js
dtribble: disabled them

packages/run-protocol/test/psm/test-psm.js
dtribble: @erights ^^

packages/run-protocol/test/psm/test-psm.js
dtribble: `makeTracer` now supports disabling, so I did that.

packages/run-protocol/src/psm/psm.js
dtribble: Now it's `getPoolBalance`

packages/run-protocol/src/psm/psm.js
dtribble: It prevents minting a total that gets to be more than the mint limit. It adds the incoming anchor amount to the total that's currently on the anchor pool, and asserts if that is over the limit. At least it's supposed to, which matches the first criteria.  I moved the addition into `assertUnderLimit` so it should be more clear now.

packages/run-protocol/src/makeTracer.js
dtribble: That should be a separate work item, and only after letting people know at scrum.

packages/run-protocol/src/psm/psm.js
dtribble: It's for internal usage as a sanity check, so I wasn't trying to be too rigorous. It's more useful to have the same format as the arguments to decrementBy et al.  Since those APIs will be changing, I thought it not worth further effort yet.

packages/run-protocol/src/psm/psm.js
dtribble: The `anchorBrand` should be in `anchorPerStable`, but one is a ratio and the other a brand. From a terms point of view, I think it is more clear to have the brand be front and center.

dtribble: Added an assert instead.

packages/run-protocol/src/psm/psm.js
dtribble:
```javascript
  assert(
    anchorPerStable.numerator.brand === anchorBrand &&
      anchorPerStable.denominator.brand === stableBrand,
    X`Ratio ${anchorPerStable} is not consistent with brands ${anchorBrand} and ${stableBrand}`,
  );
```

packages/run-protocol/src/collectFees.js
dtribble: Not my name. I just moved it. :)

packages/run-protocol/src/psm/psm.js
dtribble: There's a separate discussion about 1. using `assert` vs. 2. use `assert` here instead of relying on reallocate. 
WRT 1: Vaults use asserts, so any change to that pattern will go through all of RUN protocol and change it.
WRT 2: I think that the details that includes fees will be informative in the easy mistake case that someone says "give 2 USDC want 2 RUN"

So I concluded that I should leave it.

packages/run-protocol/src/psm/psm.js
dtribble: An as above, I think that the details that includes fees will be informative in the easy mistake case that someone says "give 2 USDC want 2 RUN"

packages/run-protocol/src/psm/psm.js
dtribble: It's not a ratio multiply. Much cleaner!

packages/run-protocol/src/makeTracer.js
turadg: Fair point about backwards compatibility. Re: scrum, this is a `run-protocol` function so I think we have quorum on this review to make a decision.

I assume you're going to squash this PR into a single commit: `feat(run-protocol): initial Parity Stability Module`. Having this change in that commit is subpar hygiene but not a blocker.

packages/run-protocol/src/collectFees.js
turadg: sure, but if there's a better name this is a good time to change it.

It's common to have chained makers. is prepending "make" the pattern to follow? (I thought there must be something better but I can't think of one. I've asked in chat.)

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,186 @@
...
+import { Far } from '@endo/marshal';
+// import { CONTRACT_ELECTORATE } from '@agoric/governance';
+
+import { AmountMath } from '@agoric/ertp';
+import { makeMakeCollectFeesInvitation } from '../collectFees.js';
+
+const { details: X } = assert;
+
+/**
+ * @file The Parity Stability Module supports efficiently minting/burning a
```
turadg, line 21: happy to see a @file comment. Seems like it would be even better at the top (before imports)

packages/run-protocol/src/psm/psm.js
```diff
@@ -0,0 +1,186 @@
...
+import '@agoric/governance/src/exported';
+import {
+  assertProposalShape,
+  ceilMultiplyBy,
+  floorDivideBy,
+  floorMultiplyBy,
+  makeRatio,
+} from '@agoric/zoe/src/contractSupport/index.js';
+import { Far } from '@endo/marshal';
+// import { CONTRACT_ELECTORATE } from '@agoric/governance';
```
turadg, line 13: stray

packages/run-protocol/src/psm/psm.js
turadg: Not sure if you saw this. I'll accept not changing the args because the APIs may be changing, but I think the code in master should have this check for what the API is now. Also conveys the expectations to the code reader.

```suggestion
  const fromValues = Object.values(txFrom);
  const toValues = Object.values(txTo);
  assert(fromValues.length === toValues.length === 1);
  assert(AmountMath.isEqual(fromValues[0], toValues[0]));
```

turadg (approved):
Some requests but none blocking.

turadg (commented):
Just noticed this has auto-merge enabled so I'm changing to Comment review so you can see my requests before it's merged.

turadg:
> Just noticed this has auto-merge enabled so I'm changing to Comment review so you can see my requests before it's merged.

TIL that adding a Comment review does not dismiss an Approve. So the auto-merge merged.

@dtribble please consider the comments added this morning.

Issue #4381, opened 2022-01-25
allow bundlecaps in vatParameters


warner:
## What is the Problem Being Solved?

To enable contract upgrade, specifically to allow the version-2 vat code to perform #3062 Kind-behavior replacements during the very first crank of the new version's worker lifetime (before any inbound deliveries might reference those virtual objects), we need to change ZCF to:
* realize when it's running in `version >= 2`, rather than `version 1`
  * (the version-1 vat will receive a delivery instructing it what contract bundle to load, so we can make a #2268 zygote from the not-yet-specialized vat before that point)
* read `vatParameters` to learn the new contract bundlecap that is to be loaded
* use the bundlecap to retrieve the actual bundle
* `importBundle()` to load the contract code
* inform the contract code that an upgrade is in progress, and let it complete
  * the contract code can then reattach the virtual-object Kinds created by its predecessor

Currently, `vatParameters` is stored as a string in the kvStore, delivered to vat workers as they are started, and unpacked with a simple `JSON.parse` by liveslots.

To make this work, we need `vatParameters` to be able to hold caps, so the vat-upgrade code can pass a bundlecap for the new contract version to the upgraded ZCF vat. It isn't sufficient to use a bundleID, because one needs access to `devices.bundle` to synchronously convert that into a real bundle, and then we'd need to store `devices.bundle` in vatParameters. The need to fetch the bundle in the initial startup crank precludes a scheme which sends a message to `vatAdmin` to retrieve the bundle, since the response (`dispatch.notify`) could not be delivered until some future crank, which is too late.

## Description of the Design

Here's what I've sketched out so far:
* the kvStore holds `vNN.vatParameters` as `JSON.stringify(capdata)`, using krefs instead of vrefs
* `vatKeeper.setVatParameters(capdata)` is responsible for updating refcounts
  *  so if version-1 references bundlecapA and version-2 references bundlecapB instead, `setVatParameters(ver2capdata)` will increfB and decref A (except that bundlecaps, like all device nodes, are currently immortal, and don't have refcounts; but all other objects should be tracked properly)
* the vat upgrade process uses `vatKeeper.setVatParameters` before launching the new vat

[...]

PR #4787, opened 2022-03-09
move vatParameters to startVat() delivery

### LINKED ISSUE #4766: comms: provide `sendExplicitSeqNums` in vatParameters

The comms vat has two initialization parameters:
* `identifierBase`: this initializes the counter used to allocate object/promise identifiers like `roNN` and `lpNN`. We use it during a few unit tests that involve multiple comms vats, so we can tell their messages apart in the debug logs (#1432 , #2529).
* `sendExplicitSeqNums`: If `true`, transmitted messages include an explicit sequence number. If `false`, the sequence number is implicit, and messages are slightly smaller (maybe 10 bytes). The receiver can handle either, and it will validate any explicit seqnums it gets. The idea was that we leave it turned on until we're confident in the implicit tracking and/or we stop wanting to debug or interpret comms messages, then turn it off. It currently defaults to `true`. When `true`, we do one additional DB read per `transmit()` (to fetch the current seqnum). We could combine `remote.nextSendSeqNum()` and `remote.advanceSendSeqNum()` to avoid the extra read.

`identifierBase` is only used during initialization of the persistent state, and obviously must be the same for all members of a consensus machine.

`sendExplicitSeqNums` would tolerate being changed over time, but for any given call, it must be the same for all members of a consensus machine.

To support #4381, I've been moving `vatParameters` out of `setBundle` (and therefore `setup()`) and into the `dispatch.startVat` message. This delays the availability of both parameters until the `startVat` message is delivered. `identifierBase` was straightforward (we don't need to do `state.initialize()` until `startVat`), but I ran into a problem with `sendExplicitSeqNums`. When `vatParameters` arrived as an argument to `setup()`, we could hold `sendExplicitSeqNums` in RAM for the lifetime of the (local) worker, knowing that we'd get the same value after a reboot.

But when they only arrive in `startVat`, we don't get a copy during the second and subsequent incarnations of the worker. To maintain consistent behavior across reboots, we must either move the parameter back into the `setup()` args, or store it into the DB during `startVat` (and read it out of the DB later).

I see the following options:

* 1: create `setupParameters` (basically a renaming of the old `setup()`-delivered vatParameters), only it for setup vats, and only for manager-local. Remove vat-selected option for exporting default `setup` or named `buildRootObject`, change `managerOptions.enableSetup` into `.useSetup`, pass `setupParameters.sendExplicitSeqNums`. All incarnations get the same value in `setup()`
* 2: pass it through `vatParameters`, write into DB, read from DB on every `transmit`
* 3: pass through `vatParameters`, write into DB, only read from DB during the first `transmit` of a new incarnation
* 4: remove the parameter entirely, always send explicit seqnums
* 5: remove the parameter entirely, never send explicit seqnums

I've been looking forward to deleting a lot of `vatParameter` -handline code, by removing it from `managerOptions`. Option 1 would only let me do half of that cleanup, and would add a new argument to DynamicVatOptions (we'd have both `setupParameters` and `vatParameters`). There would additional code to restrict use of `setupParameters` to local workers (because we don't currently support `setup()` anywhere else, and don't see much of a need for it, since the only `setup()` vat is comms, and maybe vattp some day.

Option 2 would add an extra DB read to every `transmit`. We could compensate for this by consolidating:
```js
    const seqNum = db.getSendExplicitSeqNums() ? remote.nextSendSeqNum() : ''; // one DB read, maybe one DB read
    remote.advanceSendSeqNum(); // one DB read, one DB write
```

into something like:
```js
    const sn = remote.advanceSendSeqNum(); // returns old seqnum, one DB read, one DB write
    const seqNum = db.getSendExplicitSeqNums() ? sn : ''; // one DB read
```

but of course if we stopped making it configurable (options 4 or 5) we could remove both DB reads.

It also causes this really-for-debugging value to show up in the DB, which feels wrong. But, the behavior of the comms vat is changed by this value (the `syscall.send`s it makes to vattp will contain different strings), so we should accept that this value is a legitimate part of the configuration, and treat it as such.

Option 3 would cause variable read syscalls depending upon how recently the worker had been restarted. @FUDCo points out that this isn't technically a problem: the comms vat doesn't use a transcript, so nobody is tracking the syscalls that it makes, and a DB *read* won't modify the crank-hash, and the comms vat isn't metered. But I'm still really reluctant to introduce this sort of variation, since we have an incentivized testnet killed by just this sort of thing (#3726, also maybe #3021, #3471 in the mailbox device).

I'm also somewhat reluctant to remove the explicit sequence numbers. I'm no longer worried that the code might misbehave, but It makes studying the comms protocol messages a lot easier, and the space savings doesn't seem to be significant (although the cost does grow, logarithmically, over time, just like all our identifiers get longer over time).

So I'm going to go with option 2 for now: store it in the DB during `startVat()`, read from the DB during every `transmit()`. The comms vat runs co-resident with the kernel, so DB reads are much cheaper than they would be in an xsnap process, but if we find we want to squeeze out those reads for performance in the future, I'll lean towards option 4 and always send the seqnum.

## Description of the Design

```js
function doStartVat(vatParameters={}) {
  const { sendExplicitSeqNums = true } = vatParameters;
  state.setSendExplicitSeqNums(sendExplicitSeqNums);
}

function transmit() {
  const sendExplicitSeqNums = state.getExplicitSeqNums();
  ..
}
```

(this also opens up the possibility of adding a message that turns it on or off later)

## Security Considerations

consensus considerations, but no security differences

## Test Plan

Maybe unit tests, this is used so infrequently that I may be satisfied with coverage provided by the existing tests.

warner:
fix(swingset): deliver vatParameters in startVat()

Previously, `vatParameters` arrived at a vat worker in the `setBundle`
message to the supervisor (the same one that provided the vat's source
bundle). This changes the code to deliver them inside the
`dispatch.startVat()` invocation instead.

Liveslots-based vats see no change: both cases hand vatParameters in via the
`buildRootObject()` call. `setup()`-based vats must change: the `setup()`
call no longer contains a `vatParameters` argument, and vatParameters are
available temporally later than before.

As a result, the comms vat must wait for `startVat()` to learn its two
configuration settings: `identifierBase` and `sendExplicitSeqNums`. Comms now
uses both to initialize its state DB, and reads `sendExplicitSeqNums` from
the DB for each `transit()` call.

This paves the way for `vatParameters` to contain slots, not just inert data.
The VatTranslator that converts `startVat` KernelDeliveryObjects into
VatDeliveryObjects does not yet perform slot translation, nor does liveslots
treat `vatParameters` as capdata, but they are now in the right place to
perform those tasks.

refs #4381
closes #4766

### DIFF
--- a/packages/SwingSet/src/controller/initializeKernel.js
+++ b/packages/SwingSet/src/controller/initializeKernel.js
@@ -70,7 +70,6 @@ export function initializeKernel(config, hostStorage, verbose = false) {
         'useTranscript',
         'reapInterval',
       ]);
-      creationOptions.vatParameters = vatParameters;
       creationOptions.description = `static name=${name}`;
       creationOptions.name = name;
       if (creationOptions.useTranscript === undefined) {
@@ -88,7 +87,9 @@ export function initializeKernel(config, hostStorage, verbose = false) {
       const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
       vatKeeper.setSourceAndOptions({ bundleID }, creationOptions);
       vatKeeper.initializeReapCountdown(creationOptions.reapInterval);
-      kernelKeeper.addToAcceptanceQueue(harden({ type: 'startVat', vatID }));
+      kernelKeeper.addToAcceptanceQueue(
+        harden({ type: 'startVat', vatID, vatParameters }),
+      );
       if (name === 'vatAdmin') {
         // Create a kref for the vatAdmin root, so the kernel can tell it
         // about creation/termination of dynamic vats, and the installation
--- a/packages/SwingSet/src/controller/initializeKernel.js
+++ b/packages/SwingSet/src/controller/initializeKernel.js
@@ -157,7 +157,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
       // non-empty object as vatObj0s, since an empty object would be
       // serialized as pass-by-presence. It wouldn't make much sense for the
       // bootstrap object to call itself, though.
-      const vref = Far('vref', {});
+      const vref = Far('root', {});
       vatObj0s[name] = vref;
       const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
       const kernelSlot = vatKeeper.mapVatSlotToKernelSlot(vatSlot);
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1217,6 +1217,13 @@ export default function buildKernel(
     vatKeeper.initializeReapCountdown(creationOptions.reapInterval);
 
     await vatWarehouse.loadTestVat(vatID, setup, creationOptions);
+
+    /** @type { RunQueueEventStartVat } */
+    const startVatMessage = { type: 'startVat', vatID };
+    // eslint-disable-next-line no-unused-vars
+    const ds = await processStartVat(startVatMessage);
+    // TODO: do something with DeliveryStatus, maybe just assert it's ok
+
     return vatID;
   }
 
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -526,13 +526,13 @@ export default function buildKernel(
    * @returns { Promise<DeliveryStatus> }
    */
   async function processStartVat(message) {
-    const { type, vatID } = message;
+    const { vatID, vatParameters } = message;
     // console.log(`-- processStartVat(${vatID})`);
     insistVatID(vatID);
     // eslint-disable-next-line no-use-before-define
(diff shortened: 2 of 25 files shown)


warner:
@FUDCo I recommend reviewing one commit at a time, they're pretty independent, but the first two seemed too small for their own PRs.

FUDCo (approved):
Looks fine, though it seems like a lot of mechanism just to add one parameter to `startVat`.

warner:
> Looks fine, though it seems like a lot of mechanism just to add one parameter to `startVat`.

True, but the important thing is that `startVat` (as a delivery) can have its arguments translated through the c-list (and populates the c-list, and increments the refcounts), which means `vatParameters` will be able to hold slots, which means they'll be able to hold bundlecaps, which is important for our upgrade story.

Issue #4722, opened 2022-03-02
Ability to use debug AVA tests with a launch configuration


turadg:
## What is the Problem Being Solved?

When running tests with a launch configuration ([as here](https://github.com/avajs/ava/blob/main/docs/recipes/debugging-with-vscode.md#creating-a-launch-configuration)) I get this error,
```
SyntaxError: The requested module '[file:///opt/agoric/agoric-sdk/node_modules/@babel/types/lib/index.js]()' does not provide an export named 'default'
```

That's caused by `import '@agoric/babel-standalone' in `@endo/init/pre-bundle-source.js`. (That particular error goes away when I comment out that line.)

## Description of the Design

This might be solved by https://github.com/Agoric/agoric-sdk/pull/4718 . I'll check after that is merged.

## Security Considerations
--
## Test Plan

--

kriskowal:
#4718 has merged. We may also be able to land a change that generally closes https://github.com/endojs/endo/issues/768

turadg:
@kriskowal progress had been made! Error now,
```
[/opt/agoric/agoric-sdk/node_modules/]()@agoric[/babel-generator/lib/generators/modules.js:182]()
    if (specifiers?.length) {
                   ^

  SyntaxError: Invalid or unexpected token
```

kriskowal:
> @kriskowal progress had been made! Error now,
> 
> ```
> [/opt/agoric/agoric-sdk/node_modules/]()@agoric[/babel-generator/lib/generators/modules.js:182]()
>     if (specifiers?.length) {
>                    ^
> 
>   SyntaxError: Invalid or unexpected token
> ```

cc @michaelfig This babel-generator version might be too new.

kriskowal:
@turadg Can you confirm your Node.js version?

turadg:
```
> node --version
v17.6.0
```
This works
```
cd packages/run-protocol
yarn test --serial test/vaultFactory/test-vaultFactory.js
yarn ava --serial test/vaultFactory/test-vaultFactory.js
```
But these gives the syntax error:
```
yarn ava packages/run-protocol/test/vaultFactory/test-vaultFactory.js
./node_modules/.bin/ava packages/run-protocol/test/vaultFactory/test-vaultFactory.js

```
The latter is how my launch configuration is set up, so it can be run anywhere in the repo.

kriskowal:
[...]

PR #4789, opened 2022-03-09
build: 'module' type for root package

### LINKED ISSUE #4788: Remove remaining 'esm' module loader dependencies

Tech debt and might block #4727. (A smaller change unblocked #4722)

It's not maintained and it has resulted in time lost debugging things like optional chaining support in its transformation.

## Description of the Design

Use Node.js ESM (NESM). Process: remove 'esm' from the project and burn down errors. 

## Security Considerations
--
## Test Plan

CI

turadg:
enough progress on https://github.com/Agoric/agoric-sdk/issues/4788 to unblock https://github.com/Agoric/agoric-sdk/issues/4722

## Description

Uses [Node.js's `type` field](https://nodejs.org/api/packages.html#type) to make the root package be loaded as ES modules (ESM) instead of CommonJS (CJS).

For modules that need to be CJS, change the extension to `.cjs`.

### Security Considerations

--

There is something about how to transition packages to ESM which I think still applies: https://github.com/Agoric/agoric-sdk/blob/45f9ada36e859a42cff529eaf77fad9a464678ed/COVERAGE.md#nodejs-esm-support

### Testing Considerations

Simplifies Ava config.


golang/cosmos/package.json
```diff
@@ -5,8 +5,7 @@
   "parsers": {
     "js": "mjs"
   },
-  "main": "src/index.js",
-  "module": "src/index.js",
+  "main": "src/index.cjs",
```
kriskowal, line 5: The `"parsers"` above can also be removed (anywhere you encounter them going forward). They were necessary for Endo to briefly (and badly) straddle compatibility with both RESM and NESM.

kriskowal (approved):

PR #4812, opened 2022-03-10
refactor(run-protocol): simplify centralSupply.js contract

### LINKED ISSUE #4799: centralSupply expectedAmount arg: unnecessary complexity
The logic in centralSupply that checks whether expectedAmount is satisfied is no longer needed. In fact, that whole arg is not needed.

I think it is tested, though, so this (taking it out) is a breaking change of sorts.

dckc:
- no need for expectedAmount
 - runMint is near; no need for eventual getIssuerRecord()
 - simplify memoization

closes: #4799

### Security Considerations

Complexity is an enemy of security.


packages/run-protocol/src/centralSupply.js
```diff
@@ -18,39 +16,21 @@ export const start = async (zcf, { feeMintAccess }) => {
...
-        assert(
-          AmountMath.isGTE(bootstrapAmount, expectedAmount),
-          X`${bootstrapAmount} is not at least ${expectedAmount}`,
-        );
-      }
-      return bootstrapPayment;
-    };
-    return getBootstrapPayment;
-  };
+  const { brand: runBrand } = runMint.getIssuerRecord();
```
turadg: minor: since there's only one brand being mentioned, we don't need to specify `run`. 

e.g.
```
const mint = await zcf.registerFeeMint('RUN', feeMintAccess);
const { brand } = runMint.getIssuerRecord();
```

turadg (approved):
Code LGTM and appears to carry out the change in #4799.

packages/run-protocol/src/centralSupply.js
dckc: I considered that... at the time, I landed on the side of fewer changes.

Now my hesitation is the cost of switching branches and running ci again... but I don't suppose there is any reason to rush.

Chris-Hibbert (approved):

Issue #4770, opened 2022-03-08
Lint rules for contracts


turadg:
## What is the Problem Being Solved?

Our contracts need tighter linting than our general eslint-config. `@jessie-check` provides some of that additional constraint. 

**TBD what else we need to constrain** (ticket created with the idea that jessie-check wasn't appropriate but it is in part)

## Description of the Design

Define file path patterns for what needs to be contract-linted so that files don't have to include a pragma.

Define rules for that set 

Consider using rules that require type information from TypeScript (using https://typescript-eslint.io/ )

## Security Considerations
--
## Test Plan

--

turadg:
https://github.com/Agoric/agoric-sdk/issues/3787 will introduce some constraints on use of promises in contracts. 
A promise in a contract that needs to survive an upgrade will not be able to use .then. All of RAM goes away during upgrade. `await` is an implicit `then` with a postponed callback. 

Because upgrades are voluntary we can wait to quiesce non-durable promises. But we also need to think through unexpected crashing.

"All contracts should be of the form that awaits inside of a contract are not awaiting someone else's action." - @dtribble

Chris-Hibbert:
How hard would it to detect files with a zoe-compliant `start()` function?

turadg:
> How hard would it to detect files with a zoe-compliant start() function?

I think the rule would have a precondition that it searches the file for whether to try testing it. Like https://github.com/Agoric/agoric-sdk/blob/fd1c24a84584f6b5f7b7d5e8b21d756464db05b6/packages/eslint-plugin/lib/processors/use-jessie.js  (btw @michaelfig where does that code live now? I can't find a repo for https://www.npmjs.com/package/@jessie.js/eslint-plugin )

I imagine you can get an AST for that first pass and skip if there's no `start` function exported. I don't think you'd be able to check much more of "zoe-compliant" with some Typescript annotations.

Though if the inclusion criterion was being zoe-compliant then it wouldn't be feasible to lint for that.

I take it you're asking because you'd prefer not to have a contract signifier in the filename. If so, what are some reasons?

Chris-Hibbert:
just long names.  I don't think it's a huge deal, I just thought it might not be hard to find the files without a naming convention.

dckc:
> Because `@jessie-check` errors on more than we want ...

Is that so? Please elaborate.

erights:
> > Because `@jessie-check` errors on more than we want ...
> 
> Is that so? Please elaborate.

[...]

PR #4816, opened 2022-03-10
Restival vault improvements

dtribble:
closes: #XXXX
refs: #XXXX

## Description

* make vault interest computation non async
* fix misplaced `await` usages
* fix `jessie` errors (e.g., delete `new` from Errors)
* minor naming improvements

### Security Considerations

Interleavings were possible with async interest computation. No exploit determined but eliminated the vulnerability and removed the complexity.

N/A

### Testing Considerations

Doc or pure refactor, so covered by existing tests.

### DIFF
--- a/packages/run-protocol/src/collect.js
+++ b/packages/run-protocol/src/collect.js
@@ -7,8 +7,10 @@ export const mapValues = (obj, f) =>
   harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v)])));
 
 /** @type { <X, Y>(xs: X[], ys: Y[]) => [X, Y][]} */
-export const zip = (xs, ys) => harden(xs.map((x, i) => [x, ys[i]]));
+export const zip = (xs, ys) => harden(xs.map((x, i) => [x, ys[+i]]));
 
 /** @type { <V>(obj: Record<string, ERef<V>>) => Promise<Record<string, V>> } */
-export const allValues = async obj =>
-  harden(fromEntries(zip(keys(obj), await Promise.all(values(obj)))));
+export const allValues = async obj => {
+  const resolved = await Promise.all(values(obj));
+  return harden(fromEntries(zip(keys(obj), resolved)));
+};
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -143,7 +143,7 @@ export const setupAmm = async ({
 
   if (!ammPublicFacet) {
     // ISSUE: what is this test for?!
-    throw new Error(`ammPublicFacet broken  ${ammPublicFacet}`);
+    throw Error(`ammPublicFacet broken  ${ammPublicFacet}`);
   }
 
   ammInstanceProducer.resolve(instance);
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -20,6 +20,7 @@ import { PROTOCOL_FEE_KEY, POOL_FEE_KEY } from './vpool-xyk-amm/params.js';
 
 import * as Collect from './collect.js';
 
+const { details: X } = assert;
 const { entries } = Object;
 
 const SECONDS_PER_HOUR = 60n * 60n;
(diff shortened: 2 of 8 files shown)


packages/run-protocol/src/econ-behaviors.js
```diff
@@ -143,7 +143,7 @@ export const setupAmm = async ({
 
   if (!ammPublicFacet) {
     // ISSUE: what is this test for?!
-    throw new Error(`ammPublicFacet broken  ${ammPublicFacet}`);
+    throw Error(`ammPublicFacet broken  ${ammPublicFacet}`);
```
turadg: could be a good time to answer "what is this test for?!" seems like it's a safety assert. should use `assert` instead?

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
```diff
@@ -168,6 +171,7 @@ export const makePrioritizedVaults = reschedulePriceCheck => {
     addVault(vaultId, vault);
   };
 
+  // TODO make Far
```
turadg: why not now?

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
```diff
@@ -1,4 +1,5 @@
 // @ts-check
+// @jessie-check
```
turadg, line 2: This is more strict that it needs to be. https://github.com/Agoric/agoric-sdk/issues/4770

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -146,27 +146,28 @@ export const start = async (zcf, privateArgs) => {
     return vm;
   };
 
+  const makeLoanHook = async seat => {
```
turadg: for my edification, why was this moved?

requested change: the type annotation should move with it,
```
/** @param {ZCFSeat} seat */
```

packages/run-protocol/src/econ-behaviors.js
erights: Agree that it should use `assert`, as well as `X` (aka `assert.details`). As written, it reveals a stringification to `ammPublicFacet` to callers, which is likely ok but a bad habit to fall into, as well as denying the logging system access to the actual `ammPublicFacet` object.

packages/run-protocol/src/collect.js
```diff
@@ -7,8 +7,10 @@ export const mapValues = (obj, f) =>
   harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v)])));
 
 /** @type { <X, Y>(xs: X[], ys: Y[]) => [X, Y][]} */
-export const zip = (xs, ys) => harden(xs.map((x, i) => [x, ys[i]]));
+export const zip = (xs, ys) => harden(xs.map((x, i) => [x, ys[+i]]));
```
erights, line 10: Is there a note or issue somewhere that this file should be moved to a more generic place so its exports can be reused?

packages/run-protocol/src/vaultFactory/vaultFactory.js
Chris-Hibbert: @dtribble  said a couple of times that he liked to move code to higher levels to reduce indentation. Since this isn't using anything from inside `makeVaultInvitation`, this is a safe transform.

Chris-Hibbert (approved):
Looks good, modulo Turadg's suggestions.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: Code taking more vertical space means less on the screen at a time, so less context. Our narrow lines means that each level of indentation has a chance of leading to more lines wrapping and thus les vertical content.

packages/run-protocol/src/econ-behaviors.js
dtribble: I think ti should just be removed,.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
dtribble: but it passes so I thought I'd leave it.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
dtribble: done

packages/run-protocol/src/collect.js
dtribble: I added a note in the restival doc

packages/run-protocol/src/econ-behaviors.js
dtribble: The test in order to confirm that the primary goal of that method worked, which is that it made an amm.

Issue #4730, opened 2022-03-03
exported object ID counter vs upgrade


warner:
## What is the Problem Being Solved?

While reviewing @gibson042 's #4717, I realized that liveslots keeps the "next exported object ID counter" in RAM (`nextExportID` and `nextPromiseID`, via `allocateExportID()` and `allocatePromiseID()`). That's fine for a single vat-version lifetime, but when the vat is upgraded, those RAM counters will be reset, and the vat will start exporting new objects (and promises) from `o+1` and `p+1`.

That means the new vat-version will export objects that collide with their earlier exports. We might terminate all non-durable object exports during upgrade (I'm still undecided whether vats should reconnect their "precious" Far/Remotable exports, or if they should be terminated), which might reduce the scope of the problem, but I think we still use that counter to allocate the base vref for new virtual/durable Kinds, and we need that to not collide.

Promises are troublesome too. Half of them aren't a problem: we intend to reject all outstanding promises during upgrade, so we'll remove them from the upgraded vat's c-list (the kernel rejects them, so vat-version2 doesn't need to, so it doesn't even need to reference them, and there are not references to them in durable data so vat-version2 doesn't need to be told about them either).

But that's determined by the *decider*, not the `vpid` allocator. If vat-version1 sent a message and attached a result promise vref of `p+1`, then that's going to be decided by some other vat, and so we won't reject it when the sending vat is upgraded. What we can do is remove the upgraded vat from the subscribers list, so it won't receive notifications of resolution, so it won't be a problem that the kpid is missing from the c-list. That sounds expensive, though: we have an index from kpid to subscribing VatIDs, but nothing mapping VatID to a list of subscribed promises. So maybe we should leave the upgraded vat in the subscribing list, but delete all c-list entries involving promises. When the notification arrives, `handleNotify` will see that the kpid is missing from the c-list, and conclude that the vat had already received the message somehow (batched resolutions can cause this).

Ok, so `nextPromiseID` isn't a problem: the c-list seen by vat-version2 will be entirely devoid of promise references.

But `nextExportID` might. The fix is to keep it in the vatStore, durably, but that sounds expensive too.

It's starting to sound like a pre-upgrade "cache flush" step would be convenient. We already plan to do a `bringOutYourDead` before upgrade, to write out the virtual object cache. Maybe we should write `nextExportID` to the vatStore at the same time? Kinda gross but maybe ok.

cc @FUDCo

warner:
[...]

PR #4824, opened 2022-03-11
Implement durable kind API

### LINKED ISSUE #4495: decide+implement API for durable virtual objects/collections: makeDurableKind?

Our plan for vat upgrade requires user code to use "Durable" to mark their precious-for-upgrade data. Instead of plain "virtual objects", they must use "durable objects" (aka "durable virtual objects", we're still seeking out a comfortable name). Instead of #2004 "virtual collections", they must use "durable collections".

For durable collections, the API is the same as for non-durable collections, so we only need to pick new names. I think @FUDCo is going with something like `makeScalarDurableMapStore` vs `makeScalarBigMapStore`, etc, for the four kinds of collections we currently have (`makeScalarBig{Weak,}{Map,Set}Store`). These collection objects can be serialized to the #4325 "baggage" to be retrieved by the new version during upgrade.

For non-durable virtual objects, we define `makeInstance = makeKind(behavior)`, which creates the instance maker from a behavior record. For *durable* virtual objects, we need a way for the new version to establish replacement behavior, which requires some sort of handle that is distinct from the mere instance-making authority. @FUDCo 's current proposal is:

```js
const handle = allocateDurableKind();
const makeInstance = makeDurableKind(handle, behavior);
baggage.set('name', handle);
```

The `handle` is an unforgeable serializable durable object. If the kind creates instances with vrefs like `o+12/NN`, then the handle would serialize to just `o+12`. It is technically exportable, although it's unlikely you would ever send it off-vat. Since it's durable, it can be stored in the baggage and retrieved by the next version. That next version can call `makeDurableKind` with the new behavior:

```js
const handle = baggage.get('name');
const makeInstance = makeDurableKind(handle, newBehavior);
```

## Description of the Design

Once we agree upon the API, the implementation needs are:

* define the durable versions of the collection creation functions, add to VatData
  * this is mostly trivial, the main work is to forbid non-durable data as keys/values
* define `makeDurableKind`, make the handle serializable
* docs to explain to userspace authors how to use the durability

We need to sketch out enough of the upgrade process to see what a vat's version2 would look like, and decide that we're happy with the approach. We don't need to implement baggage or upgrade for this ticket, but upgrade needs to follow smoothly from the API choices we make here.

## Security Considerations

The "handle" should be closely held, to make sure only the right portion of the new vat's code gets the ability to commandeer the durable objects.

## Storage Considerations

We need to decide how and when the vat's non-durable data gets deleted during upgrade, which may influence the implementation of the durable stores. I see two constraints:
* it should happen before the new-version userspace gets to run, or at least the new userspace should not be able to reach the non-durable data
* if the new version fails (i.e. a crash during `upgrade()` (or whatever #4382 decides), the old data should be retained, so the old version of the vat can continue operation

One option is that liveslots and the kernel have an agreement, e.g. "all vatstore keys starting with `vsd.` are durable and retained, all keys starting with `vs.` are non-durable and deleted). Then the kernel begins the `upgrade-vat` event by deleting all the non-durable keys. These deletions are stored in the crank buffer, and unwound if the upgrade crank fails.

Another option is that liveslots is responsible for enumerating and deleting all the non-durable keys itself. This reduces coupling between liveslots and kernel, however it would involve O(N) syscalls that'll be pretty expensive.

@FUDCo pointed out that each kind has an ID (`o+NN`), and each kind is either durable or not, and the metadata to indicate durability is stored in the vatstore. So if the kernel knows about the algorithm liveslots uses, the DB has enough information to figure out which keys should be kept. I'm concerned about reducing the coupling, though, so I'd be more comfortable with a protocol like `vsd.*`.

## Test Plan

unit tests

FUDCo:
This implements the durable kind API (though not the rest of the baggage machinery) per #4495 

In particular, we provide the new `VatData` function `makeKindHandle` to produce durable (and storable) handles that may be used to associate durable kinds with their implementations.  At least, that is what it is called until we change the name again.

Closes #4495

### DIFF
--- a/packages/SwingSet/docs/virtual-objects.md
+++ b/packages/SwingSet/docs/virtual-objects.md
@@ -1,39 +1,49 @@
-# TODO (update to include new stores and durable objects stuff)
+# TODO (update to include new stores stuff)
 
 # Vat Secondary Storage
 
 The kernel currently provides two secondary storage mechanisms for the use of (application) code running in a vat:
 
-- Virtual objects
+- Virtual and durable objects
 - Persistent stores
 
-Each of these is accessed via a global made available to vat code.
+These are accessed via properties of the `VatData` global made available to vat code, or more stylishly by importing from `@agoric/swingset-vat/src/storeModule.js` (that's a working placeholder module which we will be replacing it with a more ergonomic package name once we figure out what that should be).
 
-## Virtual Objects
+The APIs described here all have to do with storing data on disk.  However, you should understand an important distinction made in these APIs between the labels "virtual" and "durable".  In our usage here, things that are "virtual" will automatically swap their state to disk storage and thus don't eat up RAM space in the running vat process even if they grow large in number.  In contrast, things that are "durable" are not only stored on disk but survive the lifetime of the vat process holding them and may be retrieved later in a future version of the vat.
 
-A virtual object is an object with a durable identity whose state is automatically and transparently backed up in secondary storage.  This means that when a given virtual object is not in active use, it need not occupy any memory within the executing vat.  Thus a vat can manage an arbitrarily large number of such objects without concern about running out of memory.
+## Virtual and Durable Objects
 
-A virtual object has a "kind", which defines what sort of behavior and state it will possess.  A kind is not exactly a data type, since it comes with a concrete implementation, but it indicates a family of objects that share a set of common behaviors and a common state template.
+A virtual or durable object (which we'll abbreviate VDO to save breath) is an object whose state is automatically and transparently backed up in secondary storage.  This means that when a given VDO is not in active use, it need not occupy any memory within the executing vat.  Thus a vat can manage an arbitrarily large number of such objects without concern about running out of memory.
 
-A vat can define new kinds of virtual object by calling the `defineKind` function provided as a vat global:
+A VDO has a "kind", which defines what sort of behavior and state it will possess.  A kind is not exactly a data type, since it comes with a concrete implementation, but it indicates a family of objects that share a set of common behaviors and a common state template.
 
-  `defineKind(descriptionTag, init, actualize, finish)`
+A vat can define new kinds of VDOs by calling the `defineKind` or `defineDurableKind` functions:
 
-The return value from `defineKind` is a maker function which the vat can use to create instances of the newly defined object kind.
+  `maker = defineKind(descriptionTag, init, actualize, finish)`
+or
+  `maker = defineDurableKind(kindHandle, init, actual, finish)`
 
-The `descriptionTag` parameter is a short description string for the kind.  This is the same kind of tag string you would use in a call to `Far`.  It will appear on `.toString()` representations of corresponding `Presence` objects that are exported to remote vats.
+The return value from `defineKind` or `defineDurableKind` is a maker function which the vat can use to create instances of the newly defined VDO kind.
 
-The `init` parameter is a function that will be called when new instances are first created. It is expected to return a simple JavaScript object that represents the initialized state for the new virtual object instance.  Any parameters passed to the maker function returned by `defineKind` are passed directly to the `init` function.
+The `descriptionTag` parameter is a short description string for the kind.  This is the same kind of tag string you would use in a call to `Far`.  It will appear on `.toString()` representations of corresponding `Presence` objects that are exported to remote vats.  Note that this string should only be used for diagnostics and debugging, as it is not safe from substitution by adversarial intermediaries.
 
-The `actualize` parameter is a function that binds an in-memory instance (the "Representative") of the virtual object with the virtual object's state, associating such instances with the virtual object's behavior.  It is passed the virtual object's state as a parameter and is expected to return either:
+A `kindHandle` is a type of durable object that can be used to identify the kind in a later incarnation of the vat.  The usage of a kind handle rather than a simple tag string is the main thing that distinguished the two kind definition functions.  You obtain a kind handle for use in `defineDurableKind` by calling
+
+  `kindHandle = makeKindHandle(descriptionTag)`
+
+where `descriptionTag` is exactly the same as the same named parameter of `defineKind`.  The difference is that a kind handle is itself that a durable object that may be stored for later retrieval, and used in a future call to `defineDurableKind` to associate new behavior with the kind in question.
+
+The `init` parameter is a function that will be called when new instances are first created. It is expected to return a simple JavaScript object that represents the initialized state for the new VDO instance.  Any parameters passed to the maker function returned by `defineKind`/`defineDurableKind` are passed directly to the `init` function.
+
+The `actualize` parameter is a function that binds an in-memory instance (the "Representative") of the VDO with the VDO's state, associating such instances with the VDO's behavior.  It is passed the VDO's state as a parameter and is expected to return either:
 
 1. A new JavaScript object with methods that close over the given state.  This returned object will become the body of the new instance.  This object can be empty; in such a case it can serve as a powerless but unforgeable "marker" handle.
 
 2. A new JavaScript object populated with objects as described in (1).  These will become facets of the new instance.  The returned object will be an object mapping to the facets by name.
 
-The `actualize` function is called whenever a new virtual object instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a virtual object is received as a parameter of a message and deserialized.  The `actualize` function for a given VDO kind must always return the same shape of result: if it returns a single instance body, it must always return a single instance body; if it returns a object full of facets, it must always return an object with the exact same facet names.
+The `actualize` function is called whenever a new VDO instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a VDO is received as a parameter of a message and deserialized.  Note that for any given VDO kind, the shape of the value returned by the `actualize` function may not vary over successive calls.  That is, if it's a single facet, it must always be a single facet, and if it's multiple facets it must always be the same set of multiple facets.
 
-The `finish` parameter will, if present (it is optional), be called exactly once as part of instance initialization.  It will be invoked immediately after the `actualize` function is called for the first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function and thus becomes available to whoever requested its creation.  `finish` is passed two parameters: the virtual object's state (exactly as passed to the `actualize` function) and the virtual object itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other virtual objects.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
+The `finish` parameter is optional. It is a function that, if present, will be called exactly once as part of instance initialization.  It will be invoked _immediately_ after the `actualize` function for that instance is called for the very first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function to whoever requested its creation.  `finish` is passed two parameters: the VDO's state (exactly as passed to the `actualize` function) and the VDO itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other VDOs.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
 
 For example:
 
@@ -64,7 +74,7 @@ For example:
   const makeCounter = defineKind('counter', initCounter, actualizeCounter, finishCounter);
 ```
 
-This defines a simple counter object with two properties in its state, a count and a name.  You'd use it like this:
+This defines a simple virtual counter object with two properties in its state, a count and a name.  You'd use it like this:
 
 ```javascript
   const fooCounter = makeCounter('foo');
@@ -78,7 +88,7 @@ This defines a simple counter object with two properties in its state, a count a
   console.log(`${barCounter.getName()} count is ${barCounter.getCount()`); // "new bar count is 1"
 ```
 
-Suppose you instead wanted to provide the the increment and decrement capabilities as independent facets.  A simplified version of the above (without the name property, counter registry, and `reset` method) might look like:
+Suppose you instead wanted to provide a version with the increment and decrement capabilities made available as independent facets.  A simplified version of the above (without the name property, counter registry, and `reset` method) might look like:
 
 ```javascript
   const initFacetedCounter = () => ({ counter: 0 });
@@ -104,7 +114,15 @@ Suppose you instead wanted to provide the the increment and decrement capabiliti
   const makeFacetedCounter = defineKind('counter', initCounter, actualizeCounter);
 ```
 
-Which you'd use like:
+If you wanted to also make this durable, instead of the last line you'd generate
+the kind with something more like:
+
+```javascript
+  const facetedCounterKind = makeKindHandle('durable counter');
+  const makeFacetedCounter = defineDurableKind(facetedCounterKind, initCounter, actualizeCounter);
+```
+
+In either case you'd use it like:
 
 ```javascript
   const { incr, decr } = makeFacetedCounter('foo');
@@ -154,6 +172,6 @@ Additional important details:
 
   `state.zot.push(4);`
 
-- A virtual object can be passed as a parameter in messages to other vats.  It will be passed by presence, just like any other non-data object you might send in a message parameter.
+- A VDO can be passed as a parameter in messages to other vats.  It will be passed by presence, just like any other non-data object you might send in a message parameter.
 
-- A virtual object's state may include references to other virtual objects. The latter objects will be persisted separately and only deserialized as needed, so "swapping in" a virtual object that references other virtual objects does not entail swapping in the entire associated object graph.
+- A VDO's state may include references to other VDOs. The latter objects will be persisted separately and only deserialized as needed, so "swapping in" a VDO that references other VDOs does not entail swapping in the entire associated object graph.
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -3,6 +3,8 @@
 
 import { assert, details as X, q } from '@agoric/assert';
 import { Far } from '@endo/marshal';
+import { parseVatSlot } from '../lib/parseVatSlots.js';
+
 // import { kdebug } from './kdebug.js';
 
 /**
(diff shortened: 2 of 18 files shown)


packages/pegasus/globals.d.ts
```diff
@@ -1,6 +1,7 @@
 interface VatData {
```
erights, line 1: Are we really repeating this everywhere? Please no.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
+    syscall.vatstoreSet(
+      `vom.kind.${kindID}`,
+      JSON.stringify(durableKindDescriptor),
+    );
+    return kindHandle;
   }
 
-  function defineDurableKind(tag, init, actualize, finish) {
-    return defineKindInternal(tag, init, actualize, finish, true);
+  function defineDurableKind(kindToken, init, actualize, finish) {
```
erights:
```suggestion
  function defineDurableKind(kindHandle, init, actualize, finish) {
```
Is this right? If not, then I don't yet understand this code.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
+  const kindDescriptors = new WeakMap();
+
+  function reanimateDurableKindID(vobjID, _proforma) {
+    const { subid: kindID } = parseVatSlot(vobjID);
+    const raw = syscall.vatstoreGet(`vom.kind.${kindID}`);
+    assert(raw, X`unknown kind ID ${kindID}`);
+    const durableKindDescriptor = JSON.parse(raw);
+    const kindToken = Far('kind', {});
+    kindDescriptors.set(kindToken, durableKindDescriptor);
+    return kindToken;
```
erights:
```suggestion
    const kindHandle = Far('kind', {});
    kindDescriptors.set(kindHandle, durableKindDescriptor);
    return kindHandle;
```
Is this right? If not, then I don't yet understand this code.

Or better
```suggestion
    const kindHandle = makeHandle('Kind');
    kindDescriptors.set(kindHandle, durableKindDescriptor);
    return kindHandle;
```
though feel free to ignore this until we put `makeHandle` in a more reusable place.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
+  function makeKindHandle(tag) {
+    if (!kindIDID) {
+      kindIDID = `${allocateExportID()}`;
+      syscall.vatstoreSet('kindIDID', kindIDID);
+      vrm.registerKind(kindIDID, reanimateDurableKindID, () => null, true);
+    }
+    const kindID = `${allocateExportID()}`;
+    const kindIDvref = `o+${kindIDID}/${kindID}`;
+    const durableKindDescriptor = { kindID, tag };
+    const kindHandle = Far('kind', {});
```
erights, line 680:
```suggestion
    const kindHandle = makeHandle('Kind');
```
though feel free to ignore this until we put `makeHandle` in a more reusable place.

erights (changes requested):

erights (changes requested):
New functions must be arrow functions.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -1,39 +1,49 @@
...
+A vat can define new kinds of VDOs by calling the `defineKind` or `defineDurableKind` functions:
 
-The return value from `defineKind` is a maker function which the vat can use to create instances of the newly defined object kind.
+  `maker = defineKind(descriptionTag, init, actualize, finish)`
+or
+  `maker = defineDurableKind(kindHandle, init, actual, finish)`
+
+The return value from `defineKind` or `defineDurableKind` is a maker function which the vat can use to create instances of the newly defined VDO kind.
 
 The `descriptionTag` parameter is a short description string for the kind.  This is the same kind of tag string you would use in a call to `Far`.  It will appear on `.toString()` representations of corresponding `Presence` objects that are exported to remote vats.
```
erights:
```suggestion
The `descriptionTag` parameter is a short description string for the kind.  This is the same kind of tag string you would use in a call to `Far`.  It will appear on `.toString()` representations of corresponding `Presence` objects that are exported to remote vats. ***Use this string only for diagnostics and debugging.*** It is not safe from substitution by adversaries.
```

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: or, should the handle's farName include the tag?

```suggestion
    const kindHandle = Far(`${tag} kind`, {});
```
or
```suggestion
    const kindHandle = makeHandle(`${tag} kind`);
```

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -1,39 +1,49 @@
...
+The `finish` parameter is optional. It is a function that, if present, be called exactly once as part of instance initialization.  It will be invoked _immediately_ after the `actualize` function for that instance is called for the very first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function to whoever requested its creation.  `finish` is passed two parameters: the VDO's state (exactly as passed to the `actualize` function) and the VDO itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other VDOs.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
```
erights:
```suggestion
The `finish` parameter is optional. It is a function that, if present, will be called exactly once as part of instance initialization.  It will be invoked _immediately_ after the `actualize` function for that instance is called for the very first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function to whoever requested its creation.  `finish` is passed two parameters: the VDO's state (exactly as passed to the `actualize` function) and the VDO itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other VDOs.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
```

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -78,7 +88,7 @@ This defines a simple counter object with two properties in its state, a count a
   console.log(`${barCounter.getName()} count is ${barCounter.getCount()`); // "new bar count is 1"
 ```
 
-Suppose you instead wanted to provide the the increment and decrement capabilities as independent facets.  A simplified version of the above (without the name property, counter registry, and `reset` method) might look like:
+Suppose you instead wanted to provide a durable version with the increment and decrement capabilities made available as independent facets.  A simplified version of the above (without the name property, counter registry, and `reset` method) might look like:
```
erights: I think it is confusing to make both of these changes in one transition. How about you first make it multi-faceted and then make it durable? Or the other way around? Or as two separate changes to the original? I know it seems more verbose but I think it's worth it.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -154,6 +165,6 @@ Additional important details:
 
   `state.zot.push(4);`
 
-- A virtual object can be passed as a parameter in messages to other vats.  It will be passed by presence, just like any other non-data object you might send in a message parameter.
+- A VDO can be passed as a parameter in messages to other vats.  It will be passed by presence, just like any other non-data object you might send in a message parameter.
 
-- A virtual object's state may include references to other virtual objects. The latter objects will be persisted separately and only deserialized as needed, so "swapping in" a virtual object that references other virtual objects does not entail swapping in the entire associated object graph.
+- A VDO's state may include references to other VDOs. The latter objects will be persisted separately and only deserialized as needed, so "swapping in" a VDO that references other VDOs does not entail swapping in the entire associated object graph.
```
erights, line 177: Wow. Somehow I didn't catch that earlier and I didn't know that. It took me a moment to even realize how that could be possible. Cool!

packages/pegasus/globals.d.ts
FUDCo: Just keeping up with the status quo.  It's deeply, deeply irritating.  I will file a ticket on this.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: Dang, I thought I'd caught all those but apparently a few sneaked through the pre-checkin copy edit pass.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: I'm going to ignore all this `makeHandle` stuff until `makeHandle` is a thing.

FUDCo: > or, should the handle's farName include the tag?

I don't think so.  That would be conflating value and type in a way I find confusing.

packages/SwingSet/docs/virtual-objects.md
FUDCo: Yeah, I had the same misgiving but was concerned about making an already long and extended example even longer.  But it does conflate two unrelated things so probably going with a longer exposition is better.

FUDCo: Actually, as I went to edit this, it became clear that making the example durable only changes the last two lines of the example, so the durable case becomes an easy addendum rather than a whole 'nother example.

packages/pegasus/globals.d.ts
erights: ?

erights (changes requested):
All new functions must be arrow functions.

Aside from that, looks good to me!

packages/pegasus/globals.d.ts
FUDCo: Ticket filed.

erights (approved):
LGTM, thanks!

packages/pegasus/globals.d.ts
warner: #4830 for anyone watching

packages/SwingSet/test/virtualObjects/test-virtualObjectManager.js
```diff
@@ -390,6 +390,69 @@ test('virtual object cycles using the finish function', t => {
...
+  const fetchedKindID = placeToPutIt.get('savedKindID');
+  t.is(log.shift(), `get vc.1.ssavedKindID => ${kindSer}`);
+  t.is(log.shift(), 'get vom.kind.10 => {"kindID":"10","tag":"testkind"}');
+  t.deepEqual(log, []);
+
+  // Use it now, to define a durable kind
+  const makeThing = defineDurableKind(fetchedKindID, initThing, actualizeThing);
+  t.deepEqual(log, []);
+
+  // Make in instance of the new kind, just to be sure it's there
```
warner: typo s/in/an/

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
+    kindDescriptors.set(kindHandle, durableKindDescriptor);
+    return kindHandle;
   }
 
-  function defineDurableKind(tag, init, actualize, finish) {
-    return defineKindInternal(tag, init, actualize, finish, true);
+  const makeKindHandle = tag => {
+    if (!kindIDID) {
+      kindIDID = `${allocateExportID()}`;
```
warner, line 673: I'm a bit bothered by the idea that this will differ from one vat to another, depending upon when that vat happens to make it's first handle.. I think that's going to make debugging harder. It overlaps with a similar concern about the IDs used for the standard virtual collections.

Is there any good way to make these values be the same for all vats?

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
   }
 
   function defineKind(tag, init, actualize, finish) {
-    return defineKindInternal(tag, init, actualize, finish, false);
+    const kindID = `${allocateExportID()}`;
+    return defineKindInternal(kindID, tag, init, actualize, finish, false);
+  }
+
+  let kindIDID;
```
warner, line 658: I think there's going to be a problem with this and upgrade, overlapping with #4730. If we run this code a second time in version-2 of a vat, where the value of `kindIDID` was already allocated by version-1 (and stored on disk), this implementation will allocate a new value, and overwrite the one on disk. And I think a version-2 which starts by defining a new durable kind, then unserializes something containing a Kind Handle, might not recognize those earlier ones on disk, because by the time `unserialize` happens, the meta kind is registered under a non-matching ID.

I'm thinking that during `startVat` we need to check the DB for the stored `kindIDID`, and used the stored value if present.

This won't cause us a problem today, but it'll cause problems when we implement upgrade, and I'd like to avoid as many roadblocks as we can anticipate.

Basically, we're going to have trouble with any counter we define only in RAM, but which is used for identifiers that can survive a vat upgrade, which includes all exported vrefs, and anything used in durable data. We need a general pattern of:
* long-term stable DB key name to hold the allocated value
* conditional initialization during `vatStart`
* we can get away with holding a copy in RAM during a single version's lifetime, but we need the latest value to be written to the DB before any upgrade
  * which might mean writing it to DB after every increment, or maybe we rely upon some sort of `bringOutYourDead` (but that feels fragile to me)

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
+  const kindDescriptors = new WeakMap();
+
+  function reanimateDurableKindID(vobjID, _proforma) {
+    const { subid: kindID } = parseVatSlot(vobjID);
+    const raw = syscall.vatstoreGet(`vom.kind.${kindID}`);
+    assert(raw, X`unknown kind ID ${kindID}`);
+    const durableKindDescriptor = JSON.parse(raw);
```
warner: let's `harden` this, just as a matter of principle

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
-    return defineKindInternal(tag, init, actualize, finish, true);
+  const makeKindHandle = tag => {
+    if (!kindIDID) {
+      kindIDID = `${allocateExportID()}`;
+      syscall.vatstoreSet('kindIDID', kindIDID);
+      vrm.registerKind(kindIDID, reanimateDurableKindID, () => null, true);
+    }
+    const kindID = `${allocateExportID()}`;
+    const kindIDvref = `o+${kindIDID}/${kindID}`;
+    const durableKindDescriptor = { kindID, tag };
```
warner: harden here too

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -639,11 +642,47 @@ export function makeVirtualObjectManager(
...
+    const kindHandle = Far('kind', {});
+    kindDescriptors.set(kindHandle, durableKindDescriptor);
+    registerValue(kindIDvref, kindHandle, false);
```
warner, line 682: I'm a little bit worried that `registerValue` won't be sufficient; that we should really use some form of `makeKindHandle = defineDurableKind(..)` to define these things, so that e.g. serialize/unserialize and GC and export-status and everything will behave exactly the same (well-tested) way that normal kinds do. This would require something sneaky to break the circular reference, of course. `defineDurableKind` only needs the `kindHandle` to get a descriptor, so what if we broke `defineDurableKind` into an internal form (which takes the descriptor directly), and an external form (which takes a `kindHandle`).

Then we could build `durableKindHandleDescriptor = { kindID: kindIDID, tag: 'kind' }` and then `makeKindHandle = defineDurableKindInternal(durableKindHandleDescriptor, ( kindHandle behavior ...))`.

warner: I guess the lack of state and multiple facets on a `kindHandle` means that we wouldn't be using a lot of the functionality of `define(Durable)Kind`. I'm just worried about us making some change to e.g. the way `unserialize` works, applying that change to `defineKindInternal` or `makeRepresentative`, but forgetting to apply it to `reanimateDurableKindID`.

warner (approved):
This can probably land with just the recommended `harden` calls.

Please think about the `kindIDID`-on-DB issue, it may be that we can't address it properly until #4730 has a solution, in which case we could fix both at the same time. I guess at least add a note to the `let kindIDID` line to remind us that in-RAM counters are going to be trouble.

Consider the "use 'defineDurableKind` to define the `KindHandle` kind" thing, but I wouldn't consider it mandatory.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: I don't think there's a _good_ way, though it's certainly quite doable.  The reason I did it the way it works now is that many vats don't need these things at all and so they only get created/allocated the first time somebody asks for one.  We could pre-allocate these for every vat, which would yield uniformity, but then we'd be carrying the overhead (mostly in the form of junk database entries, which I've found to be its own source of confusion when debugging) in all the places it's not used.  In practice I haven't found the variance of the IDs from one vat to another to be a debugging issue, since if you are looking at things at that low a level you see the record of the ID allocations along with their use.  And of course you typically debug in the context of a single vat and in that context the IDs stay the same from one debug run to another.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: On first read-thru I didn't understand this comment at all, until I looked and saw that the problem is actually `allocateExportID` itself (i.e., that it's a RAM-only counter) and not so much the way it's used here.  It's also used in a bunch of other places and all of those are going to have the same kind of problem with restart-after-upgrade.  I wrote the above code on the (mistaken) belief that `allocateExportID` kept a persistent counter (we have a lot of those; I had misremembered and thought it was one also).  `allocateExportID` definitely should keep its allocation counter in the database.  We already have quite a few DB-based counters that write on increment, so I don't think one more will have a substantial performance impact.

FUDCo: Given that the vulnerability extends beyond durable kinds, I'm going to treat this as a separable issue and give it its own ticket rather than block landing this PR. #4855

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: > so what if we broke defineDurableKind into an internal form (which takes the descriptor directly), and an external form (which takes a kindHandle).

I'm confused about what you're proposing here, since what you describe is how it works now.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: And it looks like issue #4730 that @warner cited is already that issue.  I'll close the new one as a dup.

PR #4825, opened 2022-03-11
chore: configure VSCode to autofix eslint

turadg:
no ticket, came up in chat

## Description

With lint-while-type enabled (the default) there are lots of _problems_ displayed that aren't important to the developer because they can be automatically fixed by the tooling. This configures VSCode to automatically fix such problems on save and to downgrade their severity while coding. It also downgrades "no-unused-vars" because, while it can't be fixed automatically, it's a common interim state while typing and doesn't need attention.

This uses the new script from https://github.com/Agoric/agoric-sdk/pull/4418/ to make adopting these settings a pull operation and developers can see what would be changed. 

The ESLint configuration settings are described in https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/scripts/configure-vscode.sh
+++ b/scripts/configure-vscode.sh
@@ -92,7 +92,7 @@ for file in .vscode/launch.json .vscode/settings.json; do
 		else
 			printf "The file %s.new has these changes:\n\n" $file
 			git --no-pager diff --no-index $file $file.new
-			printf "\n\nTo overwrite yours:\n\n  mv %s.new %s\n\n" $file
+			printf "\n\nTo overwrite yours:\n\n  mv %s.new %s\n\n" $file $file
 		fi
 	else
 		mv $file.new $file
--- a/scripts/configure-vscode.sh
+++ b/scripts/configure-vscode.sh
@@ -18,9 +18,27 @@ mkdir -p .vscode ||
 
 cat >.vscode/settings.json.new <<\EOF ||
 {
+  // Automatically format with Prettier on save
   "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


scripts/configure-vscode.sh
```diff
@@ -60,22 +78,21 @@ cat > .vscode/launch.json.new <<\EOF ||
...
 	if test -f $file; then
 		if git diff --no-index --quiet --exit-code $file $file.new; then
 			echo "Your existing configuration matches the recommendations."
 			rm $file.new
 		else
-			printf "The file $file.new has these changes:\n\n"
+			printf "The file %s.new has these changes:\n\n" $file
 			git --no-pager diff --no-index $file $file.new
-			printf "\n\nTo overwrite yours:\n\n  mv $file.new $file\n\n"
+			printf "\n\nTo overwrite yours:\n\n  mv %s.new %s\n\n" $file
```
michaelfig:
```suggestion
			printf "\n\nTo overwrite yours:\n\n  mv %s.new %s\n\n" $file $file
```

michaelfig: I recommend installing Timon Wong's ShellCheck extension. That would have caught this error.

michaelfig (approved):

scripts/configure-vscode.sh
turadg: Thanks! I've got that and it runs in my IDE. Version 0.8.0 on my machine is not detecting this error.

PR #4827, opened 2022-03-11
docs(swingset): Improve SwingSet networking documentation

gibson042:
## Description

Update SwingSet networking documentation for better internal consistency and linking.

### Security Considerations

n/a

See above.

### Testing Considerations

n/a

### DIFF
--- a/packages/SwingSet/docs/networking.md
+++ b/packages/SwingSet/docs/networking.md
@@ -9,7 +9,7 @@ For now, our IBC implementation can only use pre-established hops, and provides 
 
 A channel via these IBC hops will terminate in IBC-aware code on either end. These endpoints might be traditional (static) IBC handlers (such as an ICS-20 token transfer module), or dynamic IBC handlers (e.g. running in a SwingSet vat). SwingSet vat code that wants to speak to vat code in a different SwingSet machine would not use the IBC connection directly: instead it would simply perform normal eventual-send operations (`E(target).foo(args)`) and let the "CapTP" promise-pipelining layer handle the details. But vat code which wants to speak to an ICS-20 handler in some other chain would need to use this layer.
 
-Vats which live inside a solo machine are able to use traditional networking layers, like TCP, HTTP, and WebSockets. This enables them to communicate with e.g. browser-side UI frontends that use WebSockets to query the vat for order status. These connections do not have to follow normal ocap rules, so vat code which accept them must provide their own authentication and access control protections.
+Vats which live inside a solo machine are able to use traditional networking layers, like TCP, HTTP, and WebSockets. This enables them to communicate with e.g. browser-side UI frontends that use WebSockets to query the vat for order status. These connections do not have to follow normal ocap rules, so vat code which accepts them must provide its own authentication and access control protections.
 
 Solo machines may be able to talk to chains and vice versa using specialized protocols. This will be used by CapTP to provide ocap messaging between these heterogeneous machine types.
 
@@ -40,11 +40,11 @@ To get a listening port, you need a `NetworkInterface` object (such as the one o
 
 ```js
 // ask for a random allocation - ends with a slash
-E(home.network).bind('/ibc-port/`)
+E(home.network).bind('/ibc-port/')
   .then(port => usePort(port));
-//
+
 // or ask for a specific port name
-E(home.network).bind('/ibc-port/my-cool-port-name`)
+E(home.network).bind('/ibc-port/my-cool-port-name')
   .then(port => usePort(port));
 ```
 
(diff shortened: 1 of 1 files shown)


packages/SwingSet/docs/networking.md
```diff
@@ -56,53 +57,49 @@ You can ask the `Port` object this returns for its local address, which is espec
...
-* `onClose(connection, reason, handler)`: this is called when the connection is closed, either because one side wanted it to close, or because an error occurred
-
-The `reason` in `onclose` is optional, as in it may be `undefined`.
+* `onReceive(connection, packetBytes, handler)`: this is called each time the remote end sends a packet of data
+* `onClose(connection, reason, handler)`: this is called when the connection is closed, either because one side wanted it to close, or because an error occurred. `reason` may be `undefined`.
 
-`onReceive` is the most important method. Each time the remote end sends a packet, your `onReceive` method will be called with the data inside that packet (currently as a String, but ideally as an ArrayBuffer with a custom `.toString()` method with an optional `encoding` argument (default `'latin1'`), so that it can contain arbitrary bytes).
+`onReceive()` is the most important method. Each time the remote end sends a packet, your `onReceive()` method will be called with the data inside that packet (currently as a String, but ideally as an ArrayBuffer with a custom `toString(encoding='latin1')` method so that it can contain arbitrary bytes).
```
gibson042, line 100: I'm curious, why is this "latin1" rather than "utf8"?

warner:
Repointing at @michaelfig who wrote the network device.

packages/SwingSet/docs/networking.md
michaelfig: > why is this "latin1" rather than "utf8"?

All combinations of octet values from `\x00` to `\xff` are accommodated in `latin1` encoding and round trip in the low part of a UTF16 code point.  Some of those combinations are invalid `utf8` and IIUC don't round-trip in UTF16 correctly.  I learned this long ago enough that it may have just been a quirk of the system I was working with.  I'd be happy to be reeducated if you know better.

packages/SwingSet/docs/networking.md
```diff
@@ -128,12 +125,12 @@ Note that if you want to listen on this port again, you can just call `port.addL
...
+Removing a listener doesn't release the port address to make it available for other `bind()` requests.  You can call:
 
 ```js
 port.revoke();
 ```
 
 to completely deallocate the port, remove all listeners, close all pending connections, and release its address.
 
-**CAUTION:** Be aware that if you call `E(home.ibcport[0]).revoke()`, it will be useless for new `.connect` or `.addListener` attempts.  You will need to provision a new Agoric client via https://testnet.agoric.com/ to obtain a new setup with a functioning `home.ibcport`.
+**CAUTION:** Be aware that if you call `E(home.ibcport[0]).revoke()`, it will be useless for new `connect()` or `addListener()` attempts.  You will need to provision a new Agoric client via https://testnet.agoric.com/ to obtain a new setup with a functioning `home.ibcport`.
```
michaelfig: Let's not point to obsolete websites.

```suggestion
**CAUTION:** Be aware that if you call `E(home.ibcport[0]).revoke()`, it will be useless for new `connect()` or `addListener()` attempts.  You will need to provision a new Agoric client to obtain a new setup with a functioning `home.ibcport[0]`.
```

packages/SwingSet/docs/networking.md
```diff
@@ -56,53 +57,49 @@ You can ask the `Port` object this returns for its local address, which is espec
...
+`onReceive()` is the most important method. Each time the remote end sends a packet, your `onReceive()` method will be called with the data inside that packet (currently as a String, but ideally as an ArrayBuffer with a custom `toString(encoding='latin1')` method so that it can contain arbitrary bytes).
 
-The return value of `onReceive` is nominally a Promise for the ACK data of the message (and should thus appear as the eventual resolution of the Promise returned by `connection.send()` on the other side). If the promise does not resolve to an ACK or resolves to an empty ACK, the implementation will automatically send a trivial `'\1'` ACK, since empty (`''`) ACKs are not supported by Cosmos ibc-go.
+The return value of `onReceive()` is nominally a Promise for the ACK data of the message (and should thus appear as the eventual resolution of the Promise returned by `connection.send()` on the other side). If the promise does not resolve to an ACK or resolves to an empty ACK, the implementation will automatically send a trivial `'\1'` ACK, since empty (`''`) ACKs are not supported by Cosmos ibc-go.
```
michaelfig:
```suggestion
The return value of `onReceive()` is nominally a Promise for the ACK data of the message (and should thus appear as the eventual resolution of the Promise returned by `connection.send()` on the other side). If the promise does not resolve to an ACK or resolves to an empty ACK, the implementation will automatically send a trivial `'\x01'` ACK, since empty (`''`) ACKs are not supported by Cosmos ibc-go.
```

michaelfig (approved):
LGTM!  Just a few minor notes to incorporate, and please lmk about the `latin1` vs `utf8` encoding question.

packages/SwingSet/docs/networking.md
gibson042: That is true, but losslessly converting an ArrayBuffer of arbitrary octets into a string of [U+0000, U+00FF] code points doesn't seem particularly useful (binary treatment wouldn't need such conversion, and a sequence of _non_-arbitrary octets could be use a more relevant conversion). Regardless, I'm happy to leave this for now.

gibson042 (commented):
Thanks, suggestions applied.

Issue #4784, opened 2022-03-09
[Governance]Using enum values as keys is incompatible with Jessie


Chris-Hibbert:
## What is the Problem Being Solved?

We'd like our contracts to be written in Jessie. Jessie doesn't support using arbitrary computed property names. Parameter governance documentation encourages this with ` getLoanParams()[INTEREST_RATE_KEY]`.

## Description of the Design

The value of the enum is a constant string. If we turned them into keywords, Jessie could know enough to accept them as "not arbitrary".

We could maybe get TypeScript to carry the fact that they're keywords around, and this might help.

There are efforts to change the standard way we declare enums. If these turned the ENUM values into something that satisfied Jessie, that would also be a win. @turadg 

Governance could change the way that parameters are made available to sidestep the issue. I think having a method that took the enum value and returned a single governed param value would take care of this. I haven't investigated whether there were other considerations that drove the design here.

## Security Considerations

Maintain governance guarantees.

## Test Plan

Yes, test it thoroughly.

turadg:
For governed params specifically, returning each value instead of the map would work.

For the more general case we could have an accessor function but that would introduce runtime overhead. @erights will Jessie fail if an object is indexed with a non-keyword key?  (i.e. a string where the leading character isn't uppercase). If so, we could rely on that and use a linter to give early feedback to the developer of what will happen at runtime.

dckc:
This is fixed in #4806 , yes?

Chris-Hibbert:
Yes.

Issue #4818, opened 2022-03-10
Review/rename 'main' keyword in parameter governance


Chris-Hibbert:
## What is the Problem Being Solved?

Parameter Governance uses `main` as the keyword in `terms` that contains the descriptions of parameters. 

## Description of the Design

Is the main concern usability?

## Security Considerations

As long as the keyword is documented and used consistently, there aren't any security consequences.

## Test Plan

Just a renaming. No affect on test requirements.

turadg:
Recap from our discussion:
- Rename ‘main’ to “governed”. 

Other improvements we discussed that could be part of this ticket or another:
- In handleParamGovernance, `augmentPublicFacet` and `makeGovernorFacet`
-  "limited creator facet" --> "governed facet"
- Some naming changes to distinguish between the facet for the "agent" which can't govern and the fully powerful facet for the "governor"
- type interfaces for the above that can be used in contracts not using the helper
- move `makeParamManager` within the helper function so the contract doesn't hold it

Chris-Hibbert:
blocked-by: #4868

Issue #4740, opened 2022-03-04
ParamManager shouldn't be exposed in governed contracts


Chris-Hibbert:
## What is the Problem Being Solved?

```
const paramMgr = await makeParamManager(zoeService, numberParam);

 const { wrapPublicFacet, wrapCreatorFacet, getAmount } = 
    handleParamGovernance(zcf, paramMgr); // paramMgr must be closely held
```
requires that the paramManager be visible in the contract, but it should only be passed to `handleParamGovernance`

## Description of the Design

The paramMgr creation should be hidden within `handleParamGovernance`.
```
 const { wrapPublicFacet, wrapCreatorFacet, getAmount } = 
    handleParamGovernance(zcf, numberParam);
```

## Security Considerations

Reduced security exposure of valuable OCaps.

## Test Plan

simplify existing tests.

Chris-Hibbert:
Closed by #4868

PR #4831, opened 2022-03-11
refactor(governance): statically typed param manager

### LINKED ISSUE #4806: Simplify governance terms API

Governance values aren't always fetched in their current state. We need to make that easier.

## Description of the Design

- Auto-generate mapping of keys to accessor functions
- Use type system instead of typed accessors ([demo](https://www.typescriptlang.org/play?#code/C4TwDgpgBA4hzAgJwM4B4AqAaKBVKEAHogHYAmKUA1hCAPYBmUGAfFALxQDeAsAFBRBUANoAFJHUhJQUAJYk8UAIaUABgHN4AEi4BhJWFnAlAG1kAvCGhTAk89VABkUcZOSgWAX1UBdAFxQABQAlBxsGGISUqA+-J4A3Pz8DACuJADGwLJ0CprAMHQAbsgkSqUZVhgExBDklHQARgBWEJk4+ESkFNS0jMwsgebpDAHYPSAoAQBKEEpkOSYgAIJISEogaLgswQFwCMjoY1vc-ILpOTbK6ekc3J7KlGUgiQJQDHRIQecklzQgUH0-ihQrxXmcLsAoKUALbQTgabRcP7eF5CK7pYQwiA+W4hMJQIYMYR-HyoqCeU5QJDwFJIBRKa4vCl8fjfS5kFLQhoALWGty4chIiGpNimSkQAQAjDhqeckGR7KJkNkyAEAOQoOiwgDKtnsapwaRSKAgqqgaq1RkQZDV5JebMh6iKtzyBWKdLKZXSEECHK5vIYOGEavkwogovFEAN5tlHwVJHUSrsdBtPmC9oudBMEAAdCY6OpAk7Cjm8gBJIXIcPAMWIEI54B0ABiskIppC6dZmezeYLRaKpfgMzl8cTypT9cbuDAUn0Jo7LwA9IuoAABYAoAC0REgmW3qw+XZ+Wdz+cLxcHwFwJGN7eCnb4QA
))
- clean up the redundancy between poolFeeBP and POOL_FEE_KEY (we don’t need to keep the initial value)

## Security Considerations
--
## Test Plan

--

turadg:
Progress on #4806 (there may be more simplifications to do, but I think the new param manager abstraction is worth landing first). Also supports #4784 though I think we should keep that open until we remove `getParams()`.

## Description

The existing param manager is a builder pattern like this,
```
  return makeParamManagerBuilder()
    .addBrandedRatio(LIQUIDATION_MARGIN_KEY, rates.liquidationMargin)
    .addBrandedRatio(INTEREST_RATE_KEY, rates.interestRate)
    .addBrandedRatio(LOAN_FEE_KEY, rates.loanFee)
    .build();
```
Or in the specific case of invitations,
```
  const paramManagerBuilder = makeParamManagerBuilder(zoe)
    .addBrand('Currency', drachmaBrand)
    .addAmount('Amt', drachmaAmount);
  // addInvitation is async, so it can't be part of the cascade.
  await paramManagerBuilder.addInvitation('Invite', invitation);
  const paramManager = paramManagerBuilder.build();
```

We want the resulting builder to have typed getters for each param (needed by https://github.com/Agoric/agoric-sdk/issues/4784) and that's overly complicated to express with static types. Much more straightforward is to pass a spec in one param and transform that in the type system to the resulting getters. E.g.,
```
  makeParamManagerSync({
    [LIQUIDATION_MARGIN_KEY]: ['brandedRatio', rates.liquidationMargin],
    [INTEREST_RATE_KEY]: ['brandedRatio', rates.interestRate],
    [LOAN_FEE_KEY]: ['brandedRatio', rates.loanFee],
  });
```

We can't special case async params but we can have a different maker. An advantage of this approach is that it expresses in the type system when `zoe` is required (previously was optional and a runtime error when absent during `addInvitation`).
```
  const paramManager = await makeParamManager(
    {
      [MALLEABLE_NUMBER]: ['nat', numberParam.value],
      [CONTRACT_ELECTORATE]: ['invitation', initialPoserInvitation],
    },
    zcf.getZoeService(),
  );
```

The resulting object has getters for each param and the type system provides the type safety that the typed getters (e.g. `getRatio`) were before. To support passing the readers to other objects, there's a new `asGetters()`. I expect it will replace all need for `getParams()` but I'll leave that for #4784.

### Security Considerations

Gets closer to Jessie compliance for contracts.

More self-documenting code with type system.

### Testing Considerations

New unit tests.


packages/governance/src/paramGovernance/paramManager.js
```diff
@@ -290,13 +288,15 @@ const makeParamManagerBuilder = zoe => {
...
       getUnknown: name => getTypedParam(ParamType.UNKNOWN, name),
       getVisibleValue,
-      getParamList,
       getInternalParamValue,
-      ...updateFunctions,
+      // Getters and setters for each param value
+      ...getters,
+      ...setters,
+      // Collection of all getters for passing to read-only contexts
+      asGetters: () => getters,
```
turadg: to support https://github.com/Agoric/agoric-sdk/issues/4784, in another PR

packages/run-protocol/src/vpool-xyk-amm/params.js
```diff
@@ -43,17 +50,17 @@ const makeAmmParams = (
...
   main: makeAmmParams(poserInvitationAmount, protocolFeeBP, poolFeeBP),
 });
 
-harden(makeParamManager);
+harden(makeAmmParamManager);
 harden(makeAmmTerms);
 harden(makeAmmParams);
 
-export { makeParamManager, makeAmmTerms, makeAmmParams };
+export { makeAmmParamManager, makeAmmTerms, makeAmmParams };
```
Chris-Hibbert, line 68: I think I prefer the style where this returns simple names like `makeTerms`, and `makeParams`. Then clients who import just one contract can use the simple names. When something needs to import more than one similarly-named values, they can rename on import.  WDYT?

packages/zoe/src/contracts/callSpread/fundedCallSpread.js
```diff
@@ -55,7 +55,9 @@ import { Position } from './position.js';
  *   denominator.value=10000)
  */
 
-/** @type {ContractStartFn} */
+/**
+ * @param {ContractFacet<Record<string, any>>} zcf
```
Chris-Hibbert: Can there be a default declaration for the case where we aren't actually saying anything about the contract's API? This is something I'd have to cut-and-paste to reuse.  
How hard is it to declare the types of the objects extracted from terms?

packages/zoe/src/types.js
```diff
@@ -23,12 +23,12 @@
  * @property {BrandKeywordRecord} brands - record with keywords
  * keys, brand values
  *
- * @typedef {StandardTerms & Record<string, any>} Terms
+ * @typedef {StandardTerms & Record<string, any>} AnyTerms
```
Chris-Hibbert, line 26: I don't think Zoe should rely on standard governance terms. I'd prefer an `AnyTerms` that omitted those, so simple contracts had a better name for `Record<string, any>`.

packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js
```diff
@@ -21,7 +21,13 @@ const BASIS_POINTS = 10000n;
...
 const DAY = SECONDS_PER_HOUR * 24n;
 
-/** @type {ContractStartFn} */
+/**
+ * @type {ContractStartFn<
+ * undefined,
+ * { makeAdjustBalancesInvitation: unknown, mintRun: unknown },
+ * {},
+ * {feeMintAccess: FeeMintAccess}>}
+ */
```
Chris-Hibbert: This seems like a good example of filling in all the details once.

packages/vats/src/core/basic-behaviors.js
```diff
@@ -233,9 +233,10 @@ export const addBankAssets = async ({
 
   const bundle = await mintHolderBundle;
 
-  /** @type {{ creatorFacet: ERef<Mint>, publicFacet: ERef<Issuer> }} */
+  /** @type {StartInstance<Issuer, Mint>} cast, XXX E() loses the method type */
+  const startInstance = E(zoe).startInstance;
```
Chris-Hibbert: I'm not a fan of pulling the `E()` out of the call site.  @dtribble  shared an example yesterday where the type info passed through the `E()` call.

Chris-Hibbert (commented):
Here are some reactions. I've been mostly looking at the consequence on contracts. I haven't looked at governance yet. I'll look there next, I think.

I think most of the changes to declarations for simple contracts are an improvement. I pointed out a few places where I think we can help more.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -18,7 +18,11 @@ const trace = makeTracer('LM');
  * uses the AMM's swapIn instead.
  */
 
-/** @type {ContractStartFn} */
+/**
+ * @type ContractStartFn<undefined, {makeDebtorInvitation: unknown}, {
```
Chris-Hibbert:
```suggestion
 * @type ContractStartFn<undefined, {makeDebtorInvitation: (debtAmount:Amount} => Invitation, {
```

packages/run-protocol/src/vpool-xyk-amm/params.js
turadg: Yeah, I tend to prefer that style myself. The module provides a namespace so no need to put that name into each export. In this particular case I think `makeAmmParamManager` is worth keeping because,
1. the other exports also have `Amm`
2. this file _imports_ `makeParamManager` from governance so we'd have to dance around the name conflict

packages/zoe/src/contracts/callSpread/fundedCallSpread.js
turadg: Yes, and there is one that treats each custom term access as `unknown`:
```
 * @template {object} [CT=Record<string, unknown>] Contract's custom terms
 * @typedef {Object} ContractFacet
```
The `Record<string, any>` here is only to a more permissive type than the default. I don't think we should put `any` into the default because then you wouldn't get any feedback that things could be better if you defined a type.

packages/zoe/src/types.js
turadg: > I don't think Zoe should rely on standard governance terms. 

Agreed. Are the `issuers` and `brands` of `StandardTerms` for governance? The typedef was added by Kate 19 months ago so I figured it's a Zoe concern.

packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js
turadg: It is and I think we can do better, but I'm going to save that for another PR. The problem I see is that the contract's `start` function is kind of like a constructor for a contract instance and what is returns is the definition of what it should return. IOW, we should rely on the type of what it actually returns instead of declaring a type and checking that it returned that.

We do want some constraints on what it can return, which I think we can accomplish with a utility type. See this [demo validating a CopyRecord](https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=2&pc=1#code/C4TwDgpgBACghgZwXARgG2gXigOwK5ppQA+ueAtihAE4lQLDUCWOA5gNxQD0X9weKALAAoEaEhQAwgHswIAEoQAxtOoATADwAVTIpXqNDZmwA0sRMnQQAfNajYA3iKhRqEOGuk40IKAG0YKBYoAGsIEGkAMygtAF0ALigACi0A2KgIAA9gCBw1BHMkVAwoAH4YtKhEnAgANxoASnYRAF9m0WEVHAZXeygnYRc4RIBySOlpEZNnKBREgCZpwaglRIAROByltpERagA6OH2upU2kpr39lH3gaQAxJkyINXP2g6V9gFUtSXaRHl6pxwOGkwFm0EY7hyaigiFhUlkCmUqjUIi6PSU1ESMjkehRGnEECirjs2Gof06ByOJzOF0pVxu90ez1e-14mOOQQKhKgIxq9WoIzR7y+PwpAIA7gALXJQcgQSg0ApwNw4EZgkIgiU4NFeHp4THYxF49R9fCEWHKnAgdoGqkINSRTgA+BFKwiIA). We could have a `Contract` type that has fields for each of the relevant types, and create that from the `start` function. PR forthcoming.

packages/vats/src/core/basic-behaviors.js
turadg: Yeah I'd really prefer not to. I just haven't found a way yet to not need the cast.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -139,19 +154,11 @@ const start = async (zcf, privateArgs) => {
     )}`,
   );
 
-  const { wrapPublicFacet, wrapCreatorFacet, getNat, getInvitationAmount } =
-    handleParamGovernance(zcf, paramManager);
-
-  const electorateInvAmt = getInvitationAmount(CONTRACT_ELECTORATE);
-  assert(
-    keyEQ(electorateInvAmt, electorateParam.value),
-    X`electorate amount (${electorateParam.value} didn't match ${electorateInvAmt}`,
```
Chris-Hibbert, line 150: Does anything validate that the `electorateParam` in the terms (is it still there?) matches the `electorateInvitationAmount`? Clients can only see the electorateParam. The `electorateInvitationAmount` is a private param set by contract governance.

packages/governance/src/contractGovernor.js
```diff
@@ -73,7 +80,18 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
...
- * @type {ContractStartFn}
+ * @template {object} PF Public facet of governed
+ * @template {ContractPowerfulCreatorFacet} CF Creator facet of governed
+ * @type {ContractStartFn<GovernorPublic, GovernedContractFacetAccess<PF>,{
+ *   timer: Timer,
+ *   electorateInstance: Instance,
+ *   governedContractInstallation: Installation<CF>,
+ *   governed: {
+ *     issuerKeywordRecord: IssuerKeywordRecord,
+ *     terms: {main: {[CONTRACT_ELECTORATE]: unknown}},
```
Chris-Hibbert: I think this must be an Amount
```suggestion
 *     terms: {main: {[CONTRACT_ELECTORATE]: Amount}},
```

packages/governance/src/contractGovernor.js
```diff
@@ -97,11 +115,13 @@ const start = async zcf => {
     electionManager: zcf.getInstance(),
   });
 
+  /** @type {StartInstance<GovernedPublicFacet<PF>, CF>} cast */
+  const startInstance = E(zoe).startInstance;
```
Chris-Hibbert: This is a problem.  In the IDEs, the type often get through. Do you know why it isn't working here?

I'd prefer a `ts-ignore` on `E(zoe).startInstance()` over pulling out the `E()` call like this.

packages/governance/src/contractHelper.js
```diff
@@ -17,13 +17,12 @@ const { details: X, quote: q } = assert;
  *  parameter values, and the governance guarantees only hold if they're not
  *  used directly by the governed contract.
  *
- * @param {ContractFacet} zcf
- * @param {ParamManagerFull} paramManager
- * @returns {ParamGovernorBundle}
+ * @template T
+ * @param {ContractFacet<{electionManager: VoteOnParamChange, main: Record<string, ParamRecord>}>} zcf
```
Chris-Hibbert: @turadg  I think you were one of the people who looked askance at `main`. Do you have an alternative to propose? If we can decide soon, it could be part of this refactoring.  My suggestions would be `govern`, `params`, or `contract`.  The last seems the weakest, but I don't have preference between the other two.

packages/governance/src/contractHelper.js
```diff
@@ -17,13 +17,12 @@ const { details: X, quote: q } = assert;
  *  parameter values, and the governance guarantees only hold if they're not
  *  used directly by the governed contract.
  *
- * @param {ContractFacet} zcf
- * @param {ParamManagerFull} paramManager
- * @returns {ParamGovernorBundle}
+ * @template T
```
Chris-Hibbert, line 23: Does this need `template T`?

packages/governance/src/contractHelper.js
```diff
@@ -47,11 +46,11 @@ const handleParamGovernance = (zcf, paramManager) => {
   };
 
   /**
-   * @param {T} originalPublicFacet
-   * @returns {T & GovernedPublicFacet}
-   * @template T
+   * @template PF
```
Chris-Hibbert, line 56:
```suggestion
   * @template PF public facet
```

packages/governance/src/contractHelper.js
```diff
@@ -62,9 +61,9 @@ const handleParamGovernance = (zcf, paramManager) => {
   };
 
   /**
-   * @param {T} originalCreatorFacet
-   * @returns {T & LimitedCreatorFacet}
-   * @template T
+   * @template CF
```
Chris-Hibbert, line 71:
```suggestion
   * @template CF creator facet
```

packages/governance/src/contractHelper.js
```diff
@@ -74,12 +73,12 @@ const handleParamGovernance = (zcf, paramManager) => {
   };
 
   /**
-   * @param {T} originalCreatorFacet
-   * @returns { GovernedCreatorFacet<T> }
-   * @template T
+   * @template CF
```
Chris-Hibbert, line 83:
```suggestion
   * @template CF creator facet
```

packages/governance/src/paramGovernance/paramMakers.js
```diff
@@ -54,6 +54,7 @@ harden(makeGovernedRatio);
 harden(makeGovernedString);
 harden(makeGovernedUnknown);
 
+// ??? do we still need these?
```
Chris-Hibbert: it doesn't seem that we do.

packages/governance/src/paramGovernance/paramManager.js
```diff
@@ -73,29 +75,30 @@ const makeParamManagerBuilder = zoe => {
...
-    updateFns[`update${name}`] = setParamValue;
+    setters[`update${name}`] = setParamValue;
     namesToParams.init(name, publicMethods);
   };
 
   // HANDLERS FOR EACH PARAMETER TYPE /////////////////////////////////////////
 
   /** @type {(name: string, amount: Amount, builder: ParamManagerBuilder) => ParamManagerBuilder} */
-  const addAmount = (name, amount, builder) => {
+  const addAmount = (name, value, builder) => {
```
Chris-Hibbert: Please update the type.

I almost argued for changing `value` back to `amount` for the obvious reason, but decided you were right to change it to be consistent with the other functions in this file.

packages/governance/src/paramGovernance/paramManager.js
```diff
@@ -175,7 +178,12 @@ const makeParamManagerBuilder = zoe => {
   };
 
   // Invitations are closely held, so we should only reveal the amount publicly.
+  // ??? how is this guaranteed: ?
```
Chris-Hibbert: The approach here makes it possible for contracts to get the actual invitation privately, and legibly assure clients that it matches the publicly visible invitation amount. Contract reviewers still have to manually verify that the actual invitation is handled carefully.

packages/governance/src/paramGovernance/paramManager.js
```diff
@@ -254,26 +256,22 @@ const makeParamManagerBuilder = zoe => {
...
     return namesToParams.get(name).getInternalValue();
   };
 
   const getParams = () => {
-    /** @type {Record<Keyword,ParamDescription>} */
+    /** @type {Record<Keyword,ParamRecord>} */
     const descriptions = {};
     for (const [name, param] of namesToParams.entries()) {
-      descriptions[name] = param.makeShortDescription();
+      console.log(name, param);
```
Chris-Hibbert: use `trace()` or remove.

packages/governance/src/paramGovernance/typedParamManager.js
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * @template {Record<Keyword, ST<ParamType>>} T
+ * @typedef {ParamManagerBase & Getters<T> & Updaters<T> & {asGetters: () => Getters<T>}} TypedParamManager
+ */
+
+/**
+ * @param {ParamType} type
+ */
+const builderMethodName = type =>
+  `add${type[0].toUpperCase() + type.substring(1)}`;
```
Chris-Hibbert, line 28: I worry that allowing callers to use names that start with lower case and converting will run into some of the issues that drove us to insist Zoe keywords have initial capital letters. I don't understand those issues well enough to say whether we're immune to them.

I concede that it's more friendly to correct mistakes than to turn them into errors. I think the cost of the errors is only at initial contract writing time, and not an ongoing burden.

packages/governance/src/paramGovernance/typedParamManager.js
```diff
@@ -0,0 +1,105 @@
...
+ * @see makeParamManager
+ * @template {Record<Keyword, SyncSpecTuple>} T
+ * @param {T} spec
+ * @returns {TypedParamManager<T>}
+ */
+const makeParamManagerSync = spec => {
+  const builder = makeParamManagerBuilder();
+
+  for (const [name, [type, value]] of Object.entries(spec)) {
+    const add = builder[builderMethodName(type)];
```
Chris-Hibbert, line 96: Does anything here complain if an invitation is provided?  I think there should be a test for that case.

I found the test!  Thanks for staying ahead of me.

packages/governance/src/types.js
```diff
@@ -20,8 +20,8 @@
  */
 
 /**
- * @typedef { 'amount' | 'brand' | 'instance' | 'installation' | 'invitation' |
- *   'nat' | 'ratio' | 'string' | 'unknown' } ParamType
+ * @typedef { 'amount' | 'brand' | 'brandedAmount'| 'brandedRatio'  | 'instance' | 'installation' | 'invitation' |
+ *   'nat' | 'ratio' | 'relativeTime' | 'string' | 'unknown' } ParamType
```
Chris-Hibbert: `relativeTime` is a good idea, but it immediately makes me ask for `timeStamp`.

packages/governance/src/types.js
```diff
@@ -549,6 +539,11 @@
  * @property {(name: string) => any} getUnknown
  */
 
+/**
+ * @template {object} PF Public facet
+ * @typedef {PF & GovernedPublicFacetMethods} GovernedPublicFacet
```
Chris-Hibbert, line 541: Yay!

packages/governance/test/swingsetTests/contractGovernor/governedContract.js
```diff
@@ -38,17 +44,20 @@ const start = async (zcf, privateArgs) => {
   const { initialPoserInvitation } = privateArgs;
 
   const paramManager = await makeParamManager(
+    {
+      [MALLEABLE_NUMBER]: ['nat', numberParam.value],
```
Chris-Hibbert, line 44: I still don't like using strings here.  And WebStorm doesn't provide the assistance that VSCode does.  It has no understanding of the constraints. Is it possible to use type names instead of strings here?

NoHelp
 
Do you know what info VSCode is keying off of here? WebStorm has access to the same language server that VSCode uses.

packages/governance/test/unitTests/test-typedParamManager.js
```diff
@@ -0,0 +1,336 @@
...
+  t.is(paramManager.getOurWeapons(), 'fear,surprise');
+  // @ts-expect-error
+  t.throws(() => paramManager.updateOurWeapons(300000000), {
+    message: '300000000 must be a string',
+  });
+});
+
+test('Unknown', async t => {
+  const paramManager = makeParamManagerSync({
+    Label: ['string', 'birtday'],
```
Chris-Hibbert:
```suggestion
    Label: ['string', 'birthday'],
```
... unless this was intentional?

Chris-Hibbert (commented):
On the whole, this looks very good. 

My one qualm is that only VSCode seems to catch typos in.
```
      [MALLEABLE_NUMBER]: ['nat', numberParam.value],
```
I may be the only one using WebStorm, but I'm not the only one who doesn't use VSCode.

packages/governance/test/swingsetTests/contractGovernor/governedContract.js
turadg: I figured that VSCode is using the tsc's [LSP](https://microsoft.github.io/language-server-protocol/), though I don't know why WebStorm wouldn't also be. 

I'll make string exports from typedParamManager so these can be manifest

Chris-Hibbert: Oh, I wonder if I just have to kick the language server. I'll take another look at that...

That didn't work, but telling WebStorm that `.js` means `typescript` worked. I'll expect more type checking help from WebStorm from here out.

turadg: Glad we got that sorted out. I'll leave the strings since the type checker will verify they're legit and both our IDEs will autocomplete them.

packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js
turadg: Contract type work: https://github.com/Agoric/agoric-sdk/issues/4845

packages/governance/src/contractHelper.js
turadg: I do think we should improve the name but I'm not confident enough just yet what it should be. I'm sure we'll come back to it.

packages/governance/test/unitTests/test-typedParamManager.js
turadg: just a typo, fixed

packages/governance/src/contractHelper.js
turadg: yes for the `TypedParamManager<T>` a couple lines down

packages/governance/src/contractHelper.js
turadg: I find that a little confusing actually, because there are two public facets being described here. The original and the governed.

packages/governance/src/contractHelper.js
turadg: Ditto concern above about the two facets

packages/governance/src/contractHelper.js
turadg: ditto

packages/governance/src/paramGovernance/paramMakers.js
turadg: I deleted a bunch 5ebe1464012c3f859aec7d38ec4d62b8bde4ba97 and will try to get the rest before merging.

packages/governance/src/paramGovernance/paramManager.js
turadg: 04f2f368f

packages/governance/src/paramGovernance/paramManager.js
turadg: 11b7d32ba

packages/governance/src/paramGovernance/typedParamManager.js
turadg: In this case we have a guarantee of the domain of `type[0]` and thus all the strings. If somehow a bad string got in, it would just call a non-existent `add` method on the builder and error.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: I took it out because I thought it was tautological. Maybe I misread the code. It looked like it came into this function, was passed into `handleParamGovernance` and then this check was that what came back out matched what was passed in. ??

packages/vats/src/core/basic-behaviors.js
turadg: f1e8ffc19faa5f6dee506e8bf7500adbfe700d32 for now

packages/governance/src/paramGovernance/paramManager.js
turadg: @Chris-Hibbert WDYT of `selectGetters()` ? since it returns a subset of the methods of the param manager

packages/governance/src/types.js
turadg: I had just copied that from an earlier union. After 5c7e4d2a39cb568bda4408dc7fd8c2aa16b24294 I saw that it wouldn't work so I removed in cdb34071444888377d13123d85d642668e246af6

packages/zoe/src/types.js
Chris-Hibbert: those are Zoe things.

packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js
Chris-Hibbert: > IOW, we should rely on the type of what it actually returns instead of declaring a type and checking that it returned that.

strongly agree.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: I think the assert verifies that the invitation amount known to parameter governance (which was initialized based on `privateArgs`) matches the one that was specified in `terms`. Those are required to be the same, but not tautalogically.

packages/governance/test/swingsetTests/contractGovernor/governedContract.js
Chris-Hibbert: Dean added another reason to prefer non-strings. All the refactoring support can work with names, and distinguish the same string as being from separate namespaces.  So if 'nat' here is `GOVERNED_PARAM.NAT`, then refactoring can distinguish it from `NAT` in some other context.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: Got it, thanks. Made re-usable in d45aa114334b161778fb30cc337ed55da97a5308

packages/zoe/src/types.js
turadg: Right. The only constraint of `AnyTerms` is that it include `StandardTerms` which "are Zoe things", so no need to omit them.

IOW, the code already satisfies "I don't think Zoe should rely on standard governance terms".

Chris-Hibbert: Hmm. I don't see it any more, maybe I mis-read something. I made my original comment because I thought I saw that this change was pulling a reliance on the minimum terms for governance into Zoe. At this point, I have to say this looks right.  mea culpa?

packages/governance/src/paramGovernance/paramManager.js
Chris-Hibbert: maybe `paramGetters` or `allGetters`?

packages/governance/src/paramGovernance/typedParamManager.js
Chris-Hibbert: I think you're right.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: > d45aa114334b161778fb30cc337ed55da97a5308

coming soon? I don't see it in the list of commits in this PR. Is there a different way to look for a commit?

turadg: Github was impaired and my branch didn't push. I think it's in now. Though my full length SHA doesn't get hyperlinked, your quoted comment of it does. 🤷

packages/governance/src/paramGovernance/paramManager.js
turadg: I want to keep that it's a subset interface on the full manager. "param" is implied by the paramManager and "all" seems uninformative. So IMO,
`asGetters` > `selectGetters` > `paramGetters` > `allGetters`

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: Oh, that looks good! Thanks for adding it in all the right places.

Except I'm hoping to [hide paramManager from the contract](https://github.com/Agoric/agoric-sdk/issues/4740).

I think that won't be too bad. When I refactor that, the call to `assertElectorateMatches()` can go inside `handleParamGovernance()`.

packages/governance/src/contractGovernor.js
```diff
@@ -80,12 +80,18 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
...
- * @param {ZCF<{
- * timer: Timer,
- * electorateInstance: Instance,
- * governedContractInstallation: Installation,
- * governed: { issuerKeywordRecord: IssuerKeywordRecord, terms: any, privateArgs: any},
- * }>} zcf
+ * @template {object} PF Public facet of governed
+ * @template {ContractPowerfulCreatorFacet} CF Creator facet of governed
+ * @type {ContractStartFn<GovernorPublic, GovernedContractFacetAccess<PF>,{
+ *   timer: Timer,
```
Chris-Hibbert:
```suggestion
 *   timer: TimerService,
```

packages/governance/src/contractHelper.js
Chris-Hibbert: After discussion, I think we agreed that the two are the original and the augmented facets. this is the original, and `wrapPublicFacet` (or whatever we rename it to) will turn it into the augmented version.

packages/governance/src/contractHelper.js
Chris-Hibbert: This one is the limited creator facet, which will be turned into a `goverorFacet`.

packages/governance/src/constants.js
```diff
@@ -0,0 +1,22 @@
...
+  Brand: 'brand',
+  BrandedAmount: 'brandedAmount',
+  BrandedRatio: 'brandedRatio',
+  Instance: 'instance',
+  Installation: 'installation',
+  Invitation: 'invitation',
+  Nat: 'nat',
+  RatioValue: 'ratioValue',
+  String: 'string',
+  Unknown: 'unknown',
```
Chris-Hibbert: I thought javascript had the convention that the values of an enum were upper case while the enum itself was camel case. This is gratuitously incompatible with the rest of our enums.

packages/governance/src/paramGovernance/paramManager.js
```diff
@@ -137,34 +157,34 @@ const makeParamManagerBuilder = zoe => {
...
     };
-    buildCopyParam(name, value, assertNat, ParamType.NAT);
+    buildCopyParam(name, value, assertNat, ParamTypes.Nat);
     return builder;
   };
 
   /** @type {(name: string, value: Ratio, builder: ParamManagerBuilder) => ParamManagerBuilder} */
-  const addRatio = (name, value, builder) => {
-    buildCopyParam(name, value, assertIsRatio, ParamType.RATIO);
+  const addRatioValue = (name, value, builder) => {
```
Chris-Hibbert: why include `value` in some names? Most of these `addFoo` methods don't have it.  `addAmountValue` particularly grates.

packages/governance/src/paramGovernance/paramManager.js
Chris-Hibbert: I don't understand `asGetters`. `asFoo` is a method that takes things that are not `Foo` and wraps them so they can act like `Foo`. 

`selectGetters` makes sense here, since it's choosing a subset of methods to do something with. But it doesn't work as the name the callers would see when they have  a `paramManager`: `pm.selectGetters().setFoo(bar)`.   The contract should say something more like `pm.allGetters().setFoo(bar)` or 
```
  const { setFoo, setGoo, setHoo } = pm.allGetters();
  setFoo(bar);
```

packages/governance/test/unitTests/test-paramGovernance.js
```diff
@@ -95,8 +91,14 @@ const setUpGovernedContract = async (zoe, electorateTerms, timer) => {
 
   const governedTerms = {
     main: {
-      [MALLEABLE_NUMBER]: makeGovernedNat(602214090000000000000000n),
-      [CONTRACT_ELECTORATE]: makeGovernedInvitation(invitationAmount),
+      [MALLEABLE_NUMBER]: {
+        type: ParamTypes.Nat,
+        value: 602214090000000000000000n,
```
Chris-Hibbert: Yeah, one of the reasons I liked `makeGovernedNat`, etc. was because clients of a contract have to build the terms, and it's a lot less messy than making them recreate the structure.

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -39,12 +39,26 @@ import { makeVaultParamManager, makeElectorateParamManager } from './params.js';
 const { details: X } = assert;
 
 /**
- * @param {ZCF<Record<string, any> &
+ * @typedef {{
```
Chris-Hibbert: It would be nicer (lower maintenance, etc.) if this inferred the type.

Chris-Hibbert (commented):
This is looking really good.  A few more suggestions.   And I think you have action items from our discussion.

packages/governance/src/contractHelper.js
turadg: Right. I captured that here https://github.com/Agoric/agoric-sdk/issues/4818#issuecomment-1071277879

I think those renamings are out of scope for this "statically typed param manager" PR but I'm happy to take on that issue. I'll assign myself.

packages/governance/src/paramGovernance/paramManager.js
turadg: > it doesn't work as the name the callers would see when they have a paramManager: pm.selectGetters().setFoo(bar)

Right now the getters are on both:
```
type TypedParamManager = ParamManagerBase & Getters<T> & Updaters<T> & {selectGetters: () => Getters<T>}}
```
 I expected `pm.getFoo()` is what someone holding `pm` would use and the `asGetters()` would be only for passing along in another object.

For now I'll make that `selectGetters` but another improvement we could consider is having all the "reads" be on a sub-object that can be passed. Then there's a clear distinction in the powers of the full object and its sub.
```
type TypedParamManager = Updaters<T> & {read: Getters<T>}}
```
To do that we'll need to kill the get[typename] style methods of `ParamManagerBase`. I don't want to saddle this PR with that but if you think it's worth doing I'll open a ticket.

turadg: Hey, I kind of like `readonly` for what this is doing. Though it has to be a method: 207bc70d3

packages/governance/src/constants.js
turadg: > javascript had the convention that the values of an enum were upper case while the enum itself was camel case

Like this?
```
  RatioValue: 'RATIO_VALUE'
```

That style is common, yeah. Used in https://www.typescriptlang.org/docs/handbook/enums.html

> This is gratuitously incompatible with the rest of our enums.

I don't know an official style but I see other enums like this:
```
  ACTIVE: 'active',
  LIQUIDATING: 'liquidating',
  CLOSED: 'closed',
  LIQUIDATED: 'liquidated',
  TRANSFER: 'transfer',
```
or
```
      NOT_REVOKED: 0,
      REVOKING: 1,
      REVOKED: 2,
```

I think that's strictly less readable than title case, but I'll change it to get this PR through. The more I work with this constant the more I think its' a mistake to manifest _values_ for literals of _types_ space. Will advance that idea in https://docs.google.com/document/d/1JbZhNtfgZ7Q4O6rnpkN3fj9-Melc_SCyDLqOpY15Hp0/edit

packages/governance/src/paramGovernance/paramManager.js
turadg: It qualifies what part can be changed. Before it was `BrandedRatio` but _all ratios_ are branded. (same goes for`Amount`) So that was confusing (I can point you to chats).

If you change an `Amount` param, it should change the `Amount` which includes the brand. Are you on board with that?

Then we need a name for the param type in which you can only change the `value` field of the Amount. Thus, `AmountValue` (and `RatioValue`). Is there a better name you'd request?

packages/governance/test/unitTests/test-paramGovernance.js
turadg: It was useful. Not sure this way is so bad though. I thought you [said to remove it](https://github.com/Agoric/agoric-sdk/pull/4831#discussion_r826364937). Are you requesting it back?

packages/governance/src/constants.js
Chris-Hibbert: I meant what you ended up on.

Maybe you'd call those (what I meant by "values") the _Names_ in the enum? (ACTIVE, LIQUIDATING)  The strings or numbers seem like noise to me. They're not an interesting part of the enum to me. They should appear only once to satisfy the language, and every other reference should be by name. Otherwise the enum isn't saving anything.  I guess that's part of why I dislike using the strings in the code.

packages/governance/test/unitTests/test-paramGovernance.js
Chris-Hibbert: I think I acquiesced, because I thought it wasn't being used any more but here we see that some clients still have to build the structure. The contract writer knows the answer better, and should help when he can. It's not necessary for a test case, but I'll probably add those helpers when I write contracts.

packages/run-protocol/src/vaultFactory/vaultFactory.js
turadg: okey doke 8fdfcea

packages/governance/test/unitTests/test-paramChangePositions.js
```diff
@@ -17,7 +17,10 @@ const negative = name => {
 };
 
 test('positions amount', t => {
-  const amountSpec = { parameterName: 'amount', key: 'something' };
+  const amountSpec = {
+    parameterName: ParamTypes.AMOUNT,
```
Chris-Hibbert, line 21: You were right to replace names with 'pName' in a few places. This is not a type, it's a name.

Chris-Hibbert (approved):
Thanks, LGTM.

packages/governance/src/constants.js
turadg: Yeah I'd call those name or keys. I totally agree that the point of an enum is to hide away the underlying literals (which I'd call values, à la key-values)

PR #4836, opened 2022-03-12
refactor(vault): simplify `adjustBalances`

### LINKED ISSUE #4835: Refactor `adjustBalances`

`Vault.adjustBalances` is complex. Simplify it so that it is also more reusable for other purposes.

## Description of the Design

* Remove the state dependencies from helper methods.
* Use a pair of a gain and a loss as a delta for computation
* Use the same pair for transfers

## Security Considerations

Simpler code provides higher assurance.

## Test Plan

As a refactor, the unit tests have good coverage of these changes.

dtribble:
Closes: #4835

## Description

refactor(vault): simplify `adjustBalances`

* refactor to use a gain and loss as a "delta"
* simplify `loanFee`; e.g., using  fee * 0 === 0
* use `loanFee` in `initVault`

### Security Considerations

Simplification is good here.

None

### Testing Considerations

The pure refactor means that existing tests provide good coverage.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -27,6 +27,15 @@ const trace = makeTracer('IV');
  *
  * The logic here is for InnerVault which is the majority of logic of vaults but
  * the user view is the `vault` value contained in VaultKit.
+ *
+ * A note on naming convention:
+ * - `Pre` is used as a postfix for any mutable value retrieved *before* an
+ *    `await`, to flag values that must used very carefully after the `await`
+ * - `new` is a prefix for values that describe the result of executing a
+ *   transaction; e.g., `debt` is the value before the txn, and `newDebt`
+ *   will be value if the txn completes.
+ * - the absence of one of these implies the opposite, so `newDebt` is the
+ *   future value fo `debt`, as computed based on values after any `await`
  */
 
 /**
@@ -423,15 +432,9 @@ export const makeInnerVault = (
     );
   };
 
-  // Was this simple, but you can't subtract `empty` from a missing allocation.
-  // const transfer = (fromSeat, toSeat, fromLoses, fromGains, key) => {
-  //   toSeat.incrementBy(fromSeat.decrementBy(harden({ [key]: fromLoses })));
-  //   fromSeat.incrementBy(toSeat.decrementBy(harden({ [key]: fromGains })));
-  // };
-
   /**
    * Stage a transfer between `fromSeat` and `toSeat`, specified as the delta between
-   * the gain and a loss on teh `fromSeat`. The gain/loss are typically from the
+   * the gain and a loss on the `fromSeat`. The gain/loss are typically from the
    * give/want respectively of a proposal. The `key` is the allocation keyword.
    *
    * @param {ZCFSeat} fromSeat
@@ -441,6 +444,7 @@ export const makeInnerVault = (
    * @param {Keyword} key
    */
   const transfer = (fromSeat, toSeat, fromLoses, fromGains, key) => {
+    // Must check `isEmpty`; can't subtract `empty` from a missing allocation.
     if (!AmountMath.isEmpty(fromLoses)) {
       toSeat.incrementBy(fromSeat.decrementBy(harden({ [key]: fromLoses })));
     }
@@ -451,10 +455,10 @@ export const makeInnerVault = (
 
   /**
    * Apply a delta to the `base` Amount, where the delta is represented as
-   * and amount to gain and and amount to lose. Typically one of those will
+   * an amount to gain and an amount to lose. Typically one of those will
    * be empty because gain/loss comes from the give/want for a specific asset
-   * on a proposal. We have to use two amount because we cannot otherwise
-   * represent a negative number.
+   * on a proposal. We use two Amounts because an Amount cannot represent
+   * a negative number (so we use a "loss" that will be subtracted).
    *
    * @param {Amount} base
    * @param {Amount} gain
@@ -466,9 +470,10 @@ export const makeInnerVault = (
 
   /**
    * Calculate the fee, the amount to mint and the resulting debt.
-   * The give and the want together reflect a delta, where typpically
-   * one is zero. In that case, the `fee` will also be zero, so the
-   * simple math works
+   * The give and the want together reflect a delta, where typically
+   * one is zero because they come from the gave/want of an offer
+   * proposal. If the `want` is zero, the `fee` will also be zero,
+   * so the simple math works.
    *
    * @param {Amount} currentDebt
    * @param {Amount} giveAmount
@@ -500,13 +505,13 @@ export const makeInnerVault = (
       // The collateral did not go up. If the collateral decreased, we pro-rate maxDebt.
       // We can pro-rate maxDebt because the quote is either linear (price is
       // unchanging) or super-linear (also called "convex"). Super-linear is from
-      // AMMs: selling less collateral would mean an even smalle price impact, so
-      // this is a conservative choice. `floorMultiply` because the debt ceiling
-      // should constrain more.
+      // AMMs: selling less collateral would mean an even smaller price impact, so
+      // this is a conservative choice.
       const runPerCollateral = makeRatioFromAmounts(
         maxDebtPre,
         newCollateralPre,
       );
+      // `floorMultiply` because the debt ceiling should be tight
       const maxDebtAfter = floorMultiplyBy(newCollateral, runPerCollateral);
       assert(
         AmountMath.isGTE(maxDebtAfter, newDebt),
@@ -622,7 +627,6 @@ export const makeInnerVault = (
       want: { RUN: wantRUN },
     } = seat.getProposal();
 
-    // todo trigger process() check right away, in case the price dropped while we ran
     const {
       newDebt: newDebtPre,
       fee,
@@ -632,7 +636,7 @@ export const makeInnerVault = (
       !AmountMath.isEmpty(fee),
       X`loan requested (${wantRUN}) is too small; cannot accrue interest`,
     );
-    assert(AmountMath.isEqual(newDebtPre, toMint), X`loan fee mismatch`);
+    assert(AmountMath.isEqual(newDebtPre, toMint), X`fee mismatch for vault`);
     trace(
       idInManager,
       'initVault',
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -121,6 +121,9 @@ export const makeInnerVault = (
   /** @type {{brand: Brand<'nat'>}} */
   const { brand: debtBrand } = mint.getIssuerRecord();
 
+  const emptyCollateral = AmountMath.makeEmpty(collateralBrand);
+  const emptyRUN = AmountMath.makeEmpty(debtBrand);
+
   /**
    * State object to support virtualization when available
    *
@@ -410,7 +413,6 @@ export const makeInnerVault = (
   const assertOnlyKeys = (proposal, keys) => {
     const onlyKeys = clause =>
       Object.getOwnPropertyNames(clause).every(c => keys.includes(c));
-
     assert(
       onlyKeys(proposal.give),
       X`extraneous terms in give: ${proposal.give}`,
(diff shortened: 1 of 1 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -360,16 +363,16 @@ export const makeInnerVault = (
       // you're paying off the debt, you get everything back. If you were
       // underwater, we should have liquidated some collateral earlier: we
       // missed our chance.
-      const currentDebt = getCurrentDebt();
+      const debt = getCurrentDebt();
```
turadg, line 375: 👍, in the context of closeHook 'current' is irrelevant

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
-        ),
-      };
-    } else {
-      return {
-        vault: startVaultAmount,
-        client: startClientAmount,
-      };
-    }
-  };
+  // Was this simple, but you can't subtract `empty` from a missing allocation.
```
turadg: This is explaining the implementation so let's put it inside the function. Also, it never "was" so,
```suggestion
  // If you could subtract `empty` from a missing allocation it would be simply:
  // toSeat.incrementBy(fromSeat.decrementBy(harden({ [key]: fromLoses })));
  // fromSeat.incrementBy(toSeat.decrementBy(harden({ [key]: fromGains })));
```
Though I think the whole code example is superfluous. Just remove this block and comment why the `isEmpty()` check is needed. I'll make that suggestion below.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+   * give/want respectively of a proposal. The `key` is the allocation keyword.
+   *
+   * @param {ZCFSeat} fromSeat
+   * @param {ZCFSeat} toSeat
+   * @param {Amount} fromLoses
+   * @param {Amount} fromGains
+   * @param {Keyword} key
+   */
+  const transfer = (fromSeat, toSeat, fromLoses, fromGains, key) => {
+    if (!AmountMath.isEmpty(fromLoses)) {
```
turadg, line 448:
```suggestion
    // Must check because you can't subtract `empty` from a missing allocation.
    if (!AmountMath.isEmpty(fromLoses)) {
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
-      seat.incrementBy(
-        vaultSeat.decrementBy(harden({ Collateral: proposal.want.Collateral })),
-      );
-    } else if (proposal.give.Collateral) {
-      vaultSeat.incrementBy(
-        seat.decrementBy(harden({ Collateral: proposal.give.Collateral })),
-      );
+  /**
+   * Stage a transfer between `fromSeat` and `toSeat`, specified as the delta between
+   * the gain and a loss on teh `fromSeat`. The gain/loss are typically from the
```
turadg:
```suggestion
   * the gain and a loss on the `fromSeat`. The gain/loss are typically from the
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+   * give/want respectively of a proposal. The `key` is the allocation keyword.
+   *
+   * @param {ZCFSeat} fromSeat
+   * @param {ZCFSeat} toSeat
+   * @param {Amount} fromLoses
+   * @param {Amount} fromGains
+   * @param {Keyword} key
```
turadg, line 444: consider putting `key` first because it's what's being transferred. "transfer Collateral from to"

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
     }
   };
 
   /**
-   * Calculate the target RUN level for the vaultSeat and clientSeat implied
-   * by the proposal. If the proposal wants collateral, transfer that amount
-   * from vault to client. If the proposal gives collateral, transfer the
-   * opposite direction. Otherwise, return the current level.
+   * Apply a delta to the `base` Amount, where the delta is represented as
+   * and amount to gain and and amount to lose. Typically one of those will
```
turadg:
```suggestion
   * an amount to gain and an amount to lose. Typically one of those will
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+   * be empty because gain/loss comes from the give/want for a specific asset
+   * on a proposal. We have to use two amount because we cannot otherwise
+   * represent a negative number.
```
turadg:
```suggestion
   * on a proposal. We have to use two amounts because Amount cannot represent
   * a negative number.
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
-      );
-    } else if (proposal.give.RUN) {
-      // We don't allow debt to be negative, so we'll refund overpayments
-      const currentDebt = getCurrentDebt();
-      const acceptedRun = AmountMath.min(proposal.give.RUN, currentDebt);
-      vaultSeat.incrementBy(seat.decrementBy(harden({ RUN: acceptedRun })));
-    }
+  /**
+   * Calculate the fee, the amount to mint and the resulting debt.
+   * The give and the want together reflect a delta, where typpically
```
turadg:
```suggestion
   * The give and the want together reflect a delta, where typically
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+   * one is zero. In that case, the `fee` will also be zero, so the
+   * simple math works
```
turadg:
```suggestion
   * simple math works.
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
-    } else if (proposal.give.RUN) {
-      // We don't allow debt to be negative, so we'll refund overpayments
-      const currentDebt = getCurrentDebt();
-      const acceptedRun = AmountMath.min(proposal.give.RUN, currentDebt);
-      vaultSeat.incrementBy(seat.decrementBy(harden({ RUN: acceptedRun })));
-    }
+  /**
+   * Calculate the fee, the amount to mint and the resulting debt.
+   * The give and the want together reflect a delta, where typpically
+   * one is zero. In that case, the `fee` will also be zero, so the
```
turadg:
```suggestion
   * either is zero. In that case, the `fee` will also be zero, so the
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
   };
 
   /**
-   * Calculate the fee, the amount to mint and the resulting debt
+   * Check whether we can proceed with an `adjustBalances`.
    *
-   * @param {ProposalRecord} proposal
-   * @param {{vault: Amount, client: Amount}} debtAfter
+   * @param {Amount} newCollateralPre
+   * @param {Amount} maxDebtPre
```
turadg, line 493: the "Pre" convention is unfamiliar and should be explained in JSDoc

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+    newCollateralPre,
+    maxDebtPre,
+    newCollateral,
+    newDebt,
+  ) => {
+    if (AmountMath.isGTE(newCollateralPre, newCollateral)) {
+      // The collateral did not go up. If the collateral decreased, we pro-rate maxDebt.
+      // We can pro-rate maxDebt because the quote is either linear (price is
+      // unchanging) or super-linear (also called "convex"). Super-linear is from
+      // AMMs: selling less collateral would mean an even smalle price impact, so
```
turadg: Good comments!
```suggestion
      // AMMs: selling less collateral would mean an even smaller price impact, so
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+      // this is a conservative choice. `floorMultiply` because the debt ceiling
```
turadg: this comment goes with line 510

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+      // should constrain more.
+      const runPerCollateral = makeRatioFromAmounts(
+        maxDebtPre,
+        newCollateralPre,
+      );
+      const maxDebtAfter = floorMultiplyBy(newCollateral, runPerCollateral);
+      assert(
```
turadg, line 516: this is a runtime error so was the caller to `checkRestart` supposed to ensure this wouldn't happen?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
-    let fee = AmountMath.makeEmpty(debtBrand);
-    if (proposal.want.RUN) {
-      fee = ceilMultiplyBy(proposal.want.RUN, manager.getLoanFee());
-      toMint = AmountMath.add(proposal.want.RUN, fee);
-      newDebt = AmountMath.add(currentDebt, toMint);
-    } else if (proposal.give.RUN) {
-      newDebt = AmountMath.subtract(currentDebt, debtAfter.vault);
-    } else {
-      newDebt = currentDebt;
+  const checkRestart = (
```
turadg, line 498: "check" doesn't imply what should happen and "restart" is not clear what's being restarted.
```suggestion
  const shouldRestartAdjustBalances = (
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -653,40 +610,48 @@ export const makeInnerVault = (
       AmountMath.isEmpty(state.debtSnapshot),
       X`vault must be empty initially`,
     );
-    const oldDebt = getCurrentDebt();
-    const oldCollateral = getCollateralAmount();
-    trace('initVaultKit start: collateral', { oldDebt, oldCollateral });
+    // TODO should this be simplified to know that the oldDebt mut be empty?
```
turadg, line 618: yeah, good idea. if `debtSnapshot` is empty then "current debt" must be empty.

for `collateralPre`, why would `vaultSeat` have collateral at this point?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -653,40 +610,48 @@ export const makeInnerVault = (
...
     // get the payout to provide access to the collateral if the
     // contract abandons
     const {
-      give: { Collateral: collateralAmount },
-      want: { RUN: wantedRun },
+      give: { Collateral: giveCollateral },
+      want: { RUN: wantRUN },
     } = seat.getProposal();
 
     // todo trigger process() check right away, in case the price dropped while we ran
```
turadg: if this todo should be done by MN-1 let's have an issue link here.

if not, let's downgrade this comment to something we _could_ do

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -653,40 +610,48 @@ export const makeInnerVault = (
...
+    const {
+      newDebt: newDebtPre,
+      fee,
+      toMint,
+    } = loanFee(debtPre, emptyDebt, wantRUN);
+    assert(
+      !AmountMath.isEmpty(fee),
+      X`loan requested (${wantRUN}) is too small; cannot accrue interest`,
+    );
+    assert(AmountMath.isEqual(newDebtPre, toMint), X`loan fee mismatch`);
```
turadg: took me a second to see why these values must be equal.
```suggestion
    assert(AmountMath.isEqual(newDebtPre, toMint), X`fee mismatch for new vault`);
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -549,87 +528,65 @@ export const makeInnerVault = (
...
-    const oldUpdater = state.outerUpdater;
+    const { vaultSeat, outerUpdater: updaterPre } = state;
     const proposal = clientSeat.getProposal();
-    const oldDebt = getCurrentDebt();
-    const oldCollateral = getCollateralAmount();
-
     assertOnlyKeys(proposal, ['Collateral', 'RUN']);
 
-    const targetCollateralAmount = targetCollateralLevels(clientSeat).vault;
+    const debtPre = getCurrentDebt();
```
Chris-Hibbert, line 540: We need a comment clarifying that 'Pre' suffixes indicates "before the price check", while `new` as a prefix means the intended outcome of the trade.  (would `out` be an improvement?)

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -421,126 +423,103 @@ export const makeInnerVault = (
...
+      // We can pro-rate maxDebt because the quote is either linear (price is
+      // unchanging) or super-linear (also called "convex"). Super-linear is from
+      // AMMs: selling less collateral would mean an even smalle price impact, so
+      // this is a conservative choice. `floorMultiply` because the debt ceiling
+      // should constrain more.
+      const runPerCollateral = makeRatioFromAmounts(
+        maxDebtPre,
+        newCollateralPre,
+      );
+      const maxDebtAfter = floorMultiplyBy(newCollateral, runPerCollateral);
```
```suggestion
      // this is a conservative choice.
      const runPerCollateral = makeRatioFromAmounts(
        maxDebtPre,
        newCollateralPre,
      );
      // `floorMultiply` because the debt ceiling should be tight.
      const maxDebtAfter = floorMultiplyBy(newCollateral, runPerCollateral);
```

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: You're right.  On first read, my response was "oh, this will abort the transaction".   But we should abort the transaction by exiting the seat.

Chris-Hibbert (approved):
@turadg  found most of my issues, and explained them better.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: I wrote it differently, but did make a clarifying change.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: I added a general comment at the top

packages/run-protocol/src/vaultFactory/vault.js
dtribble: the whole `if` block is about pro-rating, so I don't understand this.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: Check doesn't care what should happen. It's just saying your info is stale. It's up to the caller what should happen. That's why I pulled that out of the name of the function (as well as to prevent spurious line wraps :D).

packages/run-protocol/src/vaultFactory/vault.js
dtribble: We have many rejections that are from assertions (e.g., offer format rejections or brand rejections). This is just using that mechanism. This will exit the seat but also makes sure that nothing else gets run.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: It wouldn't. You added this I thought :D.  I think this should be fixed as part of changing to initialize before it's put into the prioritizedVaults collections.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: i'ts unrelated to my changes, but I think the right thing is to delete this.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: I added naming conventions to the file comment at the tope.

packages/run-protocol/src/vaultFactory/vault.js
turadg: > Check doesn't care what should happen.

A callee doesn't necessarily care what will happen, but since this includes "Restart" that makes it concerned with how this value will be used. The end of that function says,
```
    // The collateral went up. Restart if the debt *also* went up because
    // the price quote might not apply at the higher numbers.
```

So the function body is prescribing what should happen. My request is that you make that clear in the function name. 

> It's just saying your info is stale. It's up to the caller what should happen.

That would be fine too. Then I'd request the name be something like "isNewDebtGreater`. 

Anything would be better than "check" which doesn't convey how to interpret the boolean result.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Indeed, https://github.com/Agoric/agoric-sdk/pull/4527/files#diff-89c04745aa456eb87a1a04d60cdde489f4a0c6bbb0127b385845969ab259e985R613 .

PR #4837, opened 2022-03-14
ci: disable eslint rules that are covered by tsc

turadg:
## Problem being solved

ESlint rules that that don't TypeScript can give erroneous errors:
Screen Shot 2022-03-17 at 6 39 53 AM

Many errors are also redundant with what TypeScript reports, duplicating the error reports in the IDE. When suppressions are required, they have to be done twice in the code.

Lastly there's some compute cost to checking the errors that can be saved by disabling them.

## Description

Disable eslint rules that are caught by TypeScript: 
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended.ts
```
        'constructor-super': 'off', // ts(2335) & ts(2377)
        'getter-return': 'off', // ts(2378)
        'no-const-assign': 'off', // ts(2588)
        'no-dupe-args': 'off', // ts(2300)
        'no-dupe-class-members': 'off', // ts(2393) & ts(2300)
        'no-dupe-keys': 'off', // ts(1117)
        'no-func-assign': 'off', // ts(2539)
        'no-import-assign': 'off', // ts(2539) & ts(2540)
        'no-new-symbol': 'off', // ts(2588)
        'no-obj-calls': 'off', // ts(2349)
        'no-redeclare': 'off', // ts(2451)
        'no-setter-return': 'off', // ts(2408)
        'no-this-before-super': 'off', // ts(2376)
        'no-undef': 'off', // ts(2304)
        'no-unreachable': 'off', // ts(7027)
        'no-unsafe-negation': 'off', // ts(2365) & ts(2360) & ts(2358)
        'no-var': 'error', // ts transpiles let/const to var, so no need for vars any more
        'prefer-const': 'error', // ts provides better types with const
        'prefer-rest-params': 'error', // ts provides better types with rest args over arguments
        'prefer-spread': 'error', // ts transpiles spread to apply, so no need for manual apply
        'valid-typeof': 'off', // ts(2367)
```

For files that aren't `ts-check`, TypeScript 4.6 (which we use) will still [check syntax and binding errors](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#syntax-errors-js).

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@endo/eslint-config": "^0.4.6",
     "@jessie.js/eslint-plugin": "^0.1.3",
     "@types/node": "^16.7.10",
-    "@typescript-eslint/parser": "^5.14.0",
+    "@typescript-eslint/parser": "^5.15.0",
     "ava": "^3.15.0",
     "c8": "^7.7.2",
     "conventional-changelog-conventionalcommits": "^4.6.0",
--- a/packages/eslint-config/eslint-config.json
+++ b/packages/eslint-config/eslint-config.json
@@ -1,7 +1,8 @@
 {
   "extends": [
     "@endo",
-    "plugin:@jessie.js/recommended"
+    "plugin:@jessie.js/recommended",
+    "plugin:@typescript-eslint/eslint-recommended"
   ],
   "rules": {
     "jsdoc/no-multi-asterisks": "off",
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -27,7 +27,7 @@
   "devDependencies": {
     "@endo/eslint-config": "^0.4.6",
     "@jessie.js/eslint-plugin": "^0.1.3",
-    "@typescript-eslint/parser": "^5.14.0",
+    "@typescript-eslint/parser": "^5.15.0",
     "eslint": "^7.32.0",
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-config-jessie": "^0.0.6",
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -51,7 +51,7 @@
     "@babel/core": "^7.12.13",
     "@babel/plugin-syntax-jsx": "^7.12.1",
     "@material-ui/core": "4.11.3",
-    "@typescript-eslint/eslint-plugin": "^5.14.0",
+    "@typescript-eslint/eslint-plugin": "^5.15.0",
     "ava": "^3.15.0",
     "babel-eslint": "^10.1.0",
     "babel-plugin-named-asset-import": "^0.3.7",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4113,6 +4113,21 @@
     semver "^7.3.5"
     tsutils "^3.21.0"
 
+"@typescript-eslint/eslint-plugin@^5.15.0":
+  version "5.15.0"
+  resolved "https://registry.yarnpkg.com/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.15.0.tgz#c28ef7f2e688066db0b6a9d95fb74185c114fb9a"
+  integrity sha512-u6Db5JfF0Esn3tiAKELvoU5TpXVSkOpZ78cEGn/wXtT2RVqs2vkt4ge6N8cRCyw7YVKhmmLDbwI2pg92mlv7cA==
+  dependencies:
+    "@typescript-eslint/scope-manager" "5.15.0"
+    "@typescript-eslint/type-utils" "5.15.0"
+    "@typescript-eslint/utils" "5.15.0"
+    debug "^4.3.2"
+    functional-red-black-tree "^1.0.1"
+    ignore "^5.1.8"
+    regexpp "^3.2.0"
+    semver "^7.3.5"
+    tsutils "^3.21.0"
+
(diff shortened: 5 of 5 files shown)


mhofman:
I have 2 concerns:
- Some files are not checked by `tsc`, which would leave them unchecked for these rules
- I've been trying to avoid duplication of eslint plugin & co deps, which has required me to do some `yarn.lock` surgery when bumping versions. Happy to do this for you here.

turadg:
> Some files are not checked by tsc, which would leave them unchecked for these rules

Good point. Time to flip the switch from opt-in to opt-out?

> I've been trying to avoid duplication of eslint plugin & co deps

Yeah, I would have preferred not to have to update this dep in multiple places. I thought of solving that as out of scope for this PR but feel free.

mhofman:
> Time to flip the switch from opt-in to opt-out?

I'm still concerned about opted out files that contain these errors.

What is the goal of removing these eslint check? Is it to deduplicate errors in IDEs?

mhofman:
> Yeah, I would have preferred not to have to update this dep in multiple places. I thought of solving that as out of scope for this PR but feel free.

See https://github.com/Agoric/agoric-sdk/pull/4857

turadg:
> I'm still concerned about opted out files that contain these errors.

If a file opts out of type checking, that's some signal of opting out of linting. They can always re-enable eslint rules in the file.

> What is the goal of removing these eslint check? Is it to deduplicate errors in IDEs?

I've updated description to include motivation.

turadg:
I'd also like to disable `imported/named` because TS will find it and ESlint can report erroneously (happens often when refactoring):
eslint slower than tsc, giving erroneous errors

Chris-Hibbert:
> I'd also like to disable `imported/named` because TS will find it and ESlint can report erroneously (happens often when refactoring):

What does 'imported/named` refer to?  the image shows an export.  I assume you mean something other than our usual way of doing imports: `import { foo } from './bar.js'`, but I can't guess what. Do you want to disable everywhere, or throughout some files, or at single instances?

Sorry if I'm being dense.

turadg:
> What does 'imported/named' refer to?

This linter rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/named.md
"Verifies that all named imports are part of the set of named exports in the referenced module."

> the image shows an export.

That's to demonstrate that the `makeParamTerms` export that `import/named` claims isn't there actually is. The ESlint IDE plugin has a stale state resulting in phantom problems. My assertion is that we should remove it because it is both redundant and faulty. TS gives us strictly superior coverage for this type of error.

I think that [since 4.6](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#syntax-errors-js) it even works on files that don't have ts-check enable.

mhofman:
> The ESlint IDE plugin has a stale state resulting in phantom problems.

The plugin's problems are unrelated to all these rules. I encounter it all the time on a variety of rules, including some that are not in the PR. The fix is usually simple once you recognize the symptoms: restart the plugin.

turadg:
> The plugin's problems are unrelated to all these rules. I encounter it all the time on a variety of rules, including some that are not in the PR.

My claim is not that any specific rules are the cause of the stale ESLint problems. My point is that ESlint tends to get into a stale state, which you confirm.

> The fix is usually simple once you recognize the symptoms: restart the plugin.

…reminds me of the [Henny Youngman joke](https://www.goodreads.com/quotes/7570350-the-patient-says-doctor-it-hurts-when-i-do-this). If that works for you, okay, you aren't motivated to make this change. However I don't want to have to restart my plugin and I want to save others from having to do that.

We have a type checker. It works well. We don't need to repeat the same checks in ESlint. Moreover we're _better _off_ without them because they:
- cost extra compute
- are sometimes stale (as in screenshot in comment)
- in some cases consistently _wrong_ (as in PR description)

turadg:
Thanks @mhofman for taking the time to chat on a call. Conclusions:
- the error in the PR description is better solved by importing the type of the value instead of the value itself (however klunky the syntax is for jsdoc type imports)
- Typescript's non opt-in checks don't cover `no-undef`
- we don't have many ts-nocheck now (just one, which I think can go away) but we want to [flip the opt-in to opt-out](https://github.com/Agoric/agoric-sdk/issues/4885) and when we do there may be many more, which if this went through would have no checks for what this PR drops from eslint config

Therefore I'm closing this until the above is complete.

turadg:
Another one came up today,
```ts
/**
 * @param {*} maybeString
 * @returns {asserts maybeString is string}
 */
function assertString(maybeString) {
  if (typeof maybeString !== 'string') throw Error('not a string');
}
```
The rule https://github.com/gajus/eslint-plugin-jsdoc/blob/master/.README/rules/require-returns-check.md errors because it doesn't understand.

mhofman:
Right I've actually encountered this one before. It's admittedly a bit of an edge case, but I agree it's annoying. Also given that we often use return inference, I may be in favor of disabling this specific rule altogether.

turadg:
Another one today with erroneous `consistent-return`: https://github.com/Agoric/agoric-sdk/pull/5129/files#r854447445

mhofman:
I would be open to `consistent-return` being disabled. I personally don't think it brings much value for non-typed code anyway.

PR #4844, opened 2022-03-15
fix(vault): make vault transfer invitation legible

### LINKED ISSUE #4820: `vault.makeTransferInvitation`

The transfer invitation must include details of the vault being transferred in order for it to have a legible value.

## Description of the Design

Add (at least) the current debt and collateral in the invitation custom properties,

## Security Considerations

For an invitation to be a usefully transferrable eright, the invitation's amount must contain enough reliable information for someone else to determine whether they would want an invitation with that amount. 

This is *invitation legibility* (#4822), also known as "The Fred problem" and "The (invitation) pig-in-a-poke problem". It is this legibility that enables [higher-order smart contract composition](https://www.youtube.com/watch?v=iyuo0ymTt4g&list=PLzDw4TTug5O0ywHrOz4VevVTYr6Kj_KtW), where derived rights created by contracts to be treated like any other eright by other contracts. This is indeed the purpose of converting active vaults into their exclusively transferrable invitation form.

In general, for an eright to be meaningfully tradable, they must be assayable. For the trading of erights to be protected by Zoe offer safety, the amount used in a proposal's `want` must be adequate to describe the right desired such that any right that satisfies that `want` actually satisfied the desire it is supposed to represent.

The current invitation form of a vault has none of the information needed to evaluate whether one would want the right it represents. The call to `zcf.makeInvitation` should include `customProperties` to be included in the invitation's amount that will reliably provide the needed information.

## Test Plan

Make a test that wants more than the invitation specifies and therefore rejects the reallocate.

dtribble:
closes: #4820

## Description

make vault transfer invitation legible by adding custom properties to the requested invitation for the vault state at the time of the `makeTransferInvitation`

* add debt, interestSnapshot, collateral, and state to invitation custom properties
* add test to check for the custom properties
* bring the debt snapshot to current before transferring

### Security Considerations

Support the ability of a third party to automatically confirm that they received a transfer invitation for a vault in a specific state.

Properties should be described in `makeTransferInvitation` documentation.

### Testing Considerations

Added a test to verify the expected values.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -690,12 +690,28 @@ export const makeInnerVault = (
     makeAdjustBalancesInvitation,
     makeCloseInvitation,
     makeTransferInvitation: () => {
-      const { outerUpdater } = state;
+      // Bring the debt snapshot current for the final report before transfer
+      updateDebtSnapshot(getCurrentDebt());
+      const {
+        outerUpdater,
+        debtSnapshot: debt,
+        interestSnapshot: interest,
+        phase,
+      } = state;
       if (outerUpdater) {
         outerUpdater.finish(snapshotState(VaultPhase.TRANSFER));
         state.outerUpdater = null;
       }
-      return zcf.makeInvitation(makeTransferInvitationHook, 'TransferVault');
+      const transferState = {
+        debtSnapshot: { debt, interest },
+        locked: getCollateralAmount(),
+        vaultState: phase,
+      };
+      return zcf.makeInvitation(
+        makeTransferInvitationHook,
+        'TransferVault',
+        transferState,
+      );
     },
 
     // for status/debugging
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -1367,7 +1367,7 @@ test('transfer vault', async t => {
   ).getOfferResult();
   t.throwsAsync(() => E(aliceVault).getCurrentDebt());
   const debtAfter = await E(transferVault).getCurrentDebt();
-  t.deepEqual(debtAmount, debtAfter, 'vault lent 5000 RUN + fees');
+  t.deepEqual(debtAfter, debtAmount, 'vault lent 5000 RUN + fees');
   const collateralAfter = await E(transferVault).getCollateralAmount();
   t.deepEqual(collateralAmount, collateralAfter, 'vault has 1000n aEth');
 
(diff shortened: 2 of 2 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -690,12 +690,28 @@ export const makeInnerVault = (
...
+        debtSnapshot: debt,
+        interestSnapshot: interest,
+        phase,
+      } = state;
       if (outerUpdater) {
         outerUpdater.finish(snapshotState(VaultPhase.TRANSFER));
         state.outerUpdater = null;
       }
-      return zcf.makeInvitation(makeTransferInvitationHook, 'TransferVault');
+      const transferState = {
```
Chris-Hibbert, line 705: I looked more closely at `state` to see if there was anything else that should be here. The `priceAuthority` isn't necessary because it's in terms. Why does it need to be in `state` at all?  Why is `mint` included?  Are they there because they'll need to be part of what's durable?
Should the `notifier` be included in `transferState`?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1370,6 +1378,21 @@ test('transfer vault', async t => {
...
+        debt: debtAmount,
+        interest: aliceFinish.value.debtSnapshot.interest,
+      },
+      description: 'TransferVault',
+      locked: collateralAmount,
+      vaultState: 'active',
+    },
+    inviteProps,
+  );
+
```
turadg: The actual value goes first. Also [Ava has a `like`](https://github.com/avajs/ava/blob/main/docs/03-assertions.md#likeactual-selector-message):
```suggestion
  t.like(
    inviteProps,
    {
      debtSnapshot: {
        debt: debtAmount,
        interest: aliceFinish.value.debtSnapshot.interest,
      },
      description: 'TransferVault',
      locked: collateralAmount,
      vaultState: 'active',
    },
  );

```

turadg (approved):

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1370,6 +1378,21 @@ test('transfer vault', async t => {
     'transfer closed old notifier',
   );
 
+  t.deepEqual(
```
Chris-Hibbert: I think the contract should publish a validation function that clients can use to check a purported transfer invitation. The contract would know that the recipient should check the instance, possibly governance, provide a minimum for locked collateral and a maximum debt, and know the allowed states. 
I don't think ad hoc checking of some aspects of the state is enough to qualify as legibility.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: The `priceAuthority` and `mint` are in there for durability. They should likely be all folded into a single manager reference. 

The notifier should not be in the transfer state because the previous one is closed and the new one won't be created until the transfer is done.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: Can you point me at an example of what you are thinking? It seems like the real thing should include patterns (e.g., Fred wants a vault worth at least X), which shouldn't be in this PR.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1370,6 +1378,17 @@ test('transfer vault', async t => {
     'transfer closed old notifier',
   );
 
+  t.like(inviteProps, {
+    // simplify test by including any properties not mentioned below
```
turadg: this line no longer applicable

turadg (approved):
Code LGTM, except for stale comment that you might want to remove before merge.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: Chris and I discussed. This should turn into support for creating a relevant `want` pattern once that's available.

Issue #4849, opened 2022-03-15
Test for type declarations


turadg:
## What is the Problem Being Solved?

Many of our type are "best effort" and inaccurate. `any` abounds. Having tests of our types per se would help in two ways.

1. Prevent regressions. While we _improve_ the types in some places, they may regress in others if we don't have checks.
2. Document known problems. Bad type defs lead to yak shaving, where fixing one little thing branches into many others. If we had a list of known issues (documented in tests) then when those errors were encountered there could be some documentation as to why they're in that state, saving yaks.

## Description of the Design

Use https://github.com/SamVerschueren/tsd

Since types spread throughout the repo I think this should be a repo test instead of a package test.

Good to match whatever's done in https://github.com/endojs/endo/issues/1118

## Security Considerations
n/a, static analysis

## Test Plan

New part of CI linting.

PR #4850, opened 2022-03-15
chore: improve types and docs of EProxy

turadg:
## Description

- incorporate https://github.com/endojs/endo/pull/1102
- document with comments from https://github.com/endojs/endo/pull/1095
- fix loss of generic type through `ESingleCall`
- improve naming
- use function overloading in EProxy instead of conditional type

### Security Considerations

n/a, static analysis

Just works™.

### Testing Considerations

https://github.com/Agoric/agoric-sdk/issues/4849


michaelfig (approved):
LGTM!

kriskowal:
This will need to be reprised in Endo. I’m deleting the package here, now. https://github.com/Agoric/agoric-sdk/pull/4851

PR #4851, opened 2022-03-15
chore(eventual-send): Complete migration to Endo

kriskowal:
I evidently missed this in the bulk of https://github.com/endojs/endo/issues/506. The package has migrated. This removes the legacy version.

Refs: https://github.com/endojs/endo/issues/506


michaelfig (approved):
Mechanical, the way I like it!

PR #4854, opened 2022-03-16
chore: types for contract that flow through startInstance

### LINKED ISSUE #4845: Type the contract kit that startInstance returns

It's typed `any`. The lack of types makes it hard to reason about types in governance, acutely in https://github.com/Agoric/agoric-sdk/pull/4831

## Description of the Design

Type the installation of the contract and use that in `startInstance` to type the result

turadg:
closes: #4845

## Description

`startInstance` returns `publicFacet`, `creatorFacet` without any type information which makes it harder to reason about how the result can be used.

This makes `startInstance` type inspect the `installation` parameter and use that to determine the return type. To accomplish this feat, it makes the `Installation` type include an [opaque tag](https://github.com/sindresorhus/type-fest/blob/main/source/opaque.d.ts) with the start function that indicates what the installation does.

This will also help with typing contract composition. For example in @Chris-Hibbert 's work on governance.

### Security Considerations

Potential risk that people believe the types when they shouldn't. Given that the types have to be annotated on the installation, that's the decision point on what to believe and the rest flows from that.

Gains from types as documentation.

### Testing Considerations

CI


packages/governance/src/committee.js
```diff
@@ -97,7 +100,7 @@ const start = zcf => {
     getQuestion: handleP => getQuestion(handleP, allQuestions),
   });
 
-  /** @type {CommitteeElectorateCreatorFacet} */
+  /** stale def: {CommitteeElectorateCreatorFacet} */
```
Chris-Hibbert: Should this be dropped, or does `stale def` mean something to typescript?

packages/governance/src/contractGovernor.js
```diff
@@ -38,7 +38,14 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
   return validateQuestionDetails(zoe, electorate, questionDetails);
 };
 
-/*
+/**
+ * @typedef {StandardTerms} ContractGovernorTerms
+ * @property {Timer} timer
```
Chris-Hibbert:
```suggestion
 * @property {TimerService} timer
```

and elsewhere.

packages/governance/src/contractHelper.js
```diff
@@ -17,7 +17,7 @@ const { details: X, quote: q } = assert;
  *  parameter values, and the governance guarantees only hold if they're not
  *  used directly by the governed contract.
  *
- * @param {ContractFacet} zcf
+ * @param {ZoeCF} zcf
```
Chris-Hibbert: I'm not sure `ZoeCF` is a better name.  Adding "Zoe" is helpful, but we use `CF` in a lot of places here and in governance to mean `CreatorFacet`.

Chris-Hibbert (approved):
I didn't understand the stuff in `packages/deploy-script-support/src/startInstance.js`

The rest seems to be an improvement.  I asked for a couple of minor improvements.

packages/governance/src/contractHelper.js
turadg: Good point that it could be read as "creator facet". I suppose this comes down to the general confusion around "facet" terminology.

ZCF is such an essential thing to know when working with contracts, WYDT of `ZCF`?

packages/governance/src/committee.js
turadg: just wanted to highlight for posterity how the type coverage can now detect this.

Would remove but it's used in multiple places. 25b0946d7

packages/governance/src/contractGovernor.js
turadg: 8ad1d5a56

packages/governance/src/contractHelper.js
Chris-Hibbert: I think that works quite well.

turadg: 4a7bcc

packages/governance/src/contractGovernor.js
```diff
@@ -73,7 +80,12 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
  * contract (as electionManager) so clients will be able to look up the state
  * of the governed parameters.
  *
- * @type {ContractStartFn}
+ * @param {ZCF<{
+ * timer: Timer,
```
Chris-Hibbert, line 84:
```suggestion
 * timer: TimerService,
```

Chris-Hibbert (approved):
LGTM.   

Thanks, Turadg.

PR #4857, opened 2022-03-16
chore: bump ts eslint to 5.15

mhofman:
Extracted from https://github.com/Agoric/agoric-sdk/pull/4837 with cleaning up of some duplicate packages that were recently introduced

## Description

I'm trying to keep babel and eslint packages duplication to a minimum, which requires some `yarn.lock` surgery when updating versions.


turadg (approved):
LGTM. Though do we need to source it in all three package.json?

mhofman:
I believe `eslint-config` yes, not sure why `ui-components` does however.

PR #4858, opened 2022-03-17
refactor: seat.decrementBy() before incrementBy

### LINKED ISSUE #4798: decrementBy should precede incrementBy as a general safety habit
We should search all of our code, and whenever incrementBy and DecrementBy are being used together, that it be in the correct order: xyz.incrementBy(seat.decrementBy(...)); 

MarkM can explain why this is important for offer safety.

What if instead we replace them with a transferBetweenSeats() ?

Chris-Hibbert:
closes: #4798

## Description

These are not the linear case that was the top-line of the issue, but
they do follow a pattern where there are multiple increments and
decrements together, and decrement calls might error due to NSF, so
even though nothing happens till reallocate, it's still esthetically
better to do the decrements first.

### Security Considerations

An inspection pursuant to #4798 found no violations of the principal that `decrementBy` should precede `incrementBy` in the linear transfer case. These clean-ups are pursuant to the milder "general hygiene" case made there.

No change in visible behavior.

### Testing Considerations

No change in testable behavior.  No that's not true. We could add tests showing early exits if the decrementBy calls failed, but it doesn't seem worth the trouble.


dtribble (approved):
LGTM.  Elsewhere, we try to have a pattern like "decrement X and then increment X". I note that since the amounts are calculated elsewhere, there's not really applicable here, so this change seems reasonable.

PR #4859, opened 2022-03-17
ci: eslint warn import/no-default-export

turadg:
## Description

As yet undocumented coding style is to not use default exports. This documents it in the linter.

Not erroring yet because there are so many. Sample:
https://github.com/Agoric/agoric-sdk/blob/master/packages/deployment/src/main.js
https://github.com/Agoric/agoric-sdk/blob/master/packages/agoric-cli/src/start.js#L47
https://github.com/Agoric/agoric-sdk/blob/master/packages/cosmic-swingset/src/json-stable-stringify.js#L29
https://github.com/Agoric/agoric-sdk/blob/master/packages/solo/src/add-chain.js#L87
https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/src/kernel/deviceManager.js#L29

very common pattern in React components:
https://github.com/Agoric/agoric-sdk/blob/master/packages/wallet/ui/src/App.jsx#L62
(and 25 other components)

recommendation in our documentation:
https://github.com/Agoric/agoric-sdk/blob/f3744012338f7362516098d72bfb7191acb2f3a7/packages/SwingSet/docs/static-vats.md?plain=1#L50-L55

### Security Considerations

--

We could document this in the wiki, but it's better to have one source of truth. 

### Testing Considerations

--

### DIFF
--- a/packages/eslint-config/eslint-config.json
+++ b/packages/eslint-config/eslint-config.json
@@ -1,3 +1,4 @@
+// If you have questions about the settings of any rule, discussion may be found in the PR that made the change.
 {
   "extends": [
     "@endo",
@@ -12,6 +13,7 @@
     // "jsdoc/valid-types": 1,
     // "jsdoc/no-undefined-types": [1, {"definedTypes": ["never", "unknown"]}],
     "jsdoc/tag-lines": "off",
+    "import/no-default-export": "warn",
     "import/no-extraneous-dependencies": [
       "error",
       {
--- a/packages/wallet/ui/package.json
+++ b/packages/wallet/ui/package.json
@@ -78,6 +78,7 @@
     "rules": {
       "react/prop-types": "off",
       "react/react-in-jsx-scope": "off",
+      "import/no-default-export": "off",
       "import/no-extraneous-dependencies": "off",
       "react/display-name": "off"
     },


dckc:
> We could document this in the wiki, but it's better to have one source of truth.

I could have sworn some version of [Coding Style](https://github.com/Agoric/agoric-sdk/wiki/Coding-Style) said something like "we use lint tools for most stuff; the following are best practices that we have not yet automated." But now I can't find it.

Maybe I'll add it...

turadg:
This is breaking dapp building (yay integration tests!) for the same reason as [545becf](https://github.com/Agoric/agoric-sdk/pull/4859/commits/545becfb516648fec060cd79ca5dc1a1dc084fe3), `react-scripts` build running in CI treats eslint warnings as errors. Well, it treats [build warnings](https://github.com/facebook/create-react-app/blob/main/packages/react-error-overlay/build.js#L45) as errors which in my experience with Webpack seems worthwhile. What I don't know is why eslint errors are showing up in Webpack. Do you know @samsiegart ?

samsiegart:
> This is breaking dapp building (yay integration tests!) for the same reason as [545becf](https://github.com/Agoric/agoric-sdk/pull/4859/commits/545becfb516648fec060cd79ca5dc1a1dc084fe3), `react-scripts` build running in CI treats eslint warnings as errors. Well, it treats [build warnings](https://github.com/facebook/create-react-app/blob/main/packages/react-error-overlay/build.js#L45) as errors which in my experience with Webpack seems worthwhile. What I don't know is why eslint errors are showing up in Webpack. Do you know @samsiegart ?

I guess create-react-app is defaulting to treating all warnings as errors. I see the same in dev mode as well. We can override `eslintConfig` in the dapps' `package.json`s, maybe something like https://stackoverflow.com/a/66156990 would work.

turadg:
@samsiegart my question was why _eslint_ warnings are showing up in _Webpack_'s warnings. I think if we stop that then we won't have to worry about Eslint config changes breaking the build step.

samsiegart:
> @samsiegart my question was why _eslint_ warnings are showing up in _Webpack_'s warnings. I think if we stop that then we won't have to worry about Eslint config changes breaking the build step.

The webpack config is built into create-react-app https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/webpack.config.js#L764. The `failOnError` property seems relevant.

To change the _Webpack_ config, we'd have to eject the app (probably not worth it) or could use something like https://github.com/gsoft-inc/craco.

turadg:
Thanks for the research @samsiegart . So to allow dapps to continue to use default exports their eslint config can't warn on that. The wallet package here is configured to override that new `@agoric` config rule, but the other repos that use the lint config would need to be be updated too.

Almost all the uses of default exports are in UI code (GUIs in dapps and also TUIs in agoric-cli and deployment scripts). So instead of making the `@agoric` config more restrictive, I think we're better off with a variant like "lib". Because it's tighter in scope it can error instead of merely warn.

That's a bigger can of worms so deferring to https://github.com/Agoric/agoric-sdk/issues/4893 and closing this PR for now.

PR #4862, opened 2022-03-18
Use possibly separate brands for oracle prices; convert units within the vault factory

### LINKED ISSUE #4714: Vaults should price collateral in Dollars

The stability of RUN relies on Vaults pricing collateral in dollars and lending RUN and liquidating Vaults based on the dollar price of the assets.
```
dtribble(keybase): Dollars. Or USDC potentially
btulloh(keybase): We must use USD prices of the collateral as reported by the oracle. Not USDC prices.
```

## Description of the Design

The current code in `vaults.js` asks for prices in RUN.
```
  const quoteAmount = await E(priceAuthority).quoteGiven(
      collateralAmount,
      runBrand,
    );
```

```
michaelfix(keybase): I know we must have USD price oracles.  My question is one of API: should we access those price authority pairs using xBrand-to-runBrand, or should we somehow use a fake brand (no mint or issuer) specifically for USD and have xBrand-to-fakeUsdBrand?
dckc(keybase): The latter. It's not so much "fakeUsd" as "USD as defined by the oracle network"
(oops... I see you used "fake" to mean no mint)
```

## Security Considerations

This is about economic security of the platform, and doesn't impact platform security.

## Test Plan

We'll need tests for the dollar-based oracles, and we'll need to update tests to rely on dollar-based oracles.

michaelfig:
closes: #4714

## Description

As part of ongoing oracle work, this PR decouples the units quoted by the priceAuthority (USD) from the debt units (RUN).    It does the same for the priceAuthority asset vs. the vault collateral brand.

It also makes it possible to bootstrap the oracle network using price quotes in terms of `E(home.agoricNames).lookup('oracleBrand', 'USD')` even before RUN is enabled.  That well-known USD brand is inert; there is no corresponding mint or issuer.  Even if someday we were able to print money on chain, all we'd need to do is update the agoricNames entry.  @dckc can you please review the bootstrap stuff?

The main work to plumb this through the vaultFactory was to create `debtToPrice`, which is something like 1 RUN (debt) : 1 USD (price).  Each of the brands in this ratio has independent decimalPlaces, and the ratio is used to scale values when translating between the two brands.  I would argue that the ratio should not be configurable for the life of the vault factory.  I'll tag @turadg for review of the run-protocol stuff, and leave it up to you to involve other folks if you need.

Note that this also introduces `collateralToPrice` as an optional argument to `addVaultType`.  This is so that the `priceAuthority` could be initialised with a different brand for the asset, and the collateral amounts are scaled accordingly.

### Security Considerations

The inertness of the new USD brand was done in order to avoid increasing the attack surface when just a brand with a few methods was needed.

Economic security, allowing drawing the lines between USD and RUN in a different, better, way.

The main observable change is that now, `getCollateralQuote` values will be in terms of USD.  This is a good thing, but may need UI work to put `$` on any displayed price quotes.  @samsiegart, can you confirm if this needs doing?

Another change (which is less observable) is that the vault's priceAuthorities need to quote in terms of USD.  This affects oracles, but not many other folk.

### DIFF
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -2,7 +2,10 @@
 
 import { E, Far } from '@endo/far';
 import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
-import { CENTRAL_ISSUER_NAME } from '@agoric/vats/src/core/utils.js';
+import {
+  CENTRAL_ISSUER_NAME,
+  ORACLE_PRICE_BRAND_NAME,
+} from '@agoric/vats/src/core/utils.js';
 import '@agoric/governance/exported.js';
 import '@agoric/vats/exported.js';
 import '@agoric/vats/src/core/types.js';
@@ -236,6 +239,9 @@ export const startVaultFactory = async (
     produce, // {  vaultFactoryCreator }
     brand: {
       consume: { [CENTRAL_ISSUER_NAME]: centralBrandP },
+    },
+    oracleBrand: {
+      consume: { [ORACLE_PRICE_BRAND_NAME]: oraclePriceBrandP },
     },
     instance,
     installation,
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -1,10 +1,13 @@
 // @ts-check
 
 import { E, Far } from '@endo/far';
-import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
+import {
+  makeRatio,
+  makeUnitConversionRatio,
+} from '@agoric/zoe/src/contractSupport/index.js';
 import {
   CENTRAL_ISSUER_NAME,
-  ORACLE_PRICE_BRAND_NAME,
+  ORACLE_BRAND_PRICE_NAME,
 } from '@agoric/vats/src/core/utils.js';
 import '@agoric/governance/exported.js';
 import '@agoric/vats/exported.js';
@@ -241,7 +244,7 @@ export const startVaultFactory = async (
       consume: { [CENTRAL_ISSUER_NAME]: centralBrandP },
     },
     oracleBrand: {
-      consume: { [ORACLE_PRICE_BRAND_NAME]: oraclePriceBrandP },
+      consume: { [ORACLE_BRAND_PRICE_NAME]: oraclePriceBrandP },
     },
     instance,
     installation,
@@ -265,7 +268,15 @@ export const startVaultFactory = async (
     E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
   ]);
 
-  const centralBrand = await centralBrandP;
+  const [centralBrand, oraclePriceBrand] = await Promise.all([
+    centralBrandP,
+    oraclePriceBrandP,
+  ]);
+
+  const debtToPrice = await makeUnitConversionRatio(
+    centralBrand,
+    oraclePriceBrand,
+  );
 
   /**
    * Types for the governed params for the vaultFactory; addVaultType() sets actual values
@@ -292,6 +303,7 @@ export const startVaultFactory = async (
 
   const vaultFactoryTerms = makeGovernedTerms(
     priceAuthority,
+    debtToPrice,
     loanParams,
     installations.liquidate,
     chainTimerService,
--- a/packages/run-protocol/src/vaultFactory/params.js
+++ b/packages/run-protocol/src/vaultFactory/params.js
@@ -58,6 +58,7 @@ const makeElectorateParamManager = async (zoe, electorateInvitation) => {
 
 /**
  * @param {ERef<PriceAuthority>} priceAuthority
+ * @param {Ratio} debtToPrice
  * @param {LoanTiming} loanTiming
  * @param {Installation} liquidationInstall
  * @param {ERef<TimerService>} timerService
@@ -68,6 +69,7 @@ const makeElectorateParamManager = async (zoe, electorateInvitation) => {
  */
 const makeGovernedTerms = (
   priceAuthority,
+  debtToPrice,
   loanTiming,
   liquidationInstall,
   timerService,
@@ -86,6 +88,7 @@ const makeGovernedTerms = (
   return harden({
     ammPublicFacet,
     priceAuthority,
+    debtToPrice,
     loanParams: vmParamMgr.getParams(),
     loanTimingParams: timingParamMgr.getParams(),
     timerService,
--- a/packages/run-protocol/src/vaultFactory/vaultFactory.js
+++ b/packages/run-protocol/src/vaultFactory/vaultFactory.js
@@ -26,8 +26,12 @@ import {
   assertProposalShape,
   getAmountOut,
   getAmountIn,
+  makeUnitConversionRatio,
 } from '@agoric/zoe/src/contractSupport/index.js';
-import { makeRatioFromAmounts } from '@agoric/zoe/src/contractSupport/ratio.js';
+import {
+  makeRatio,
+  makeRatioFromAmounts,
+} from '@agoric/zoe/src/contractSupport/ratio.js';
 import { Far } from '@endo/marshal';
 import { assertElectorateMatches } from '@agoric/governance';
 
@@ -45,10 +49,14 @@ const { details: X } = assert;
  *   liquidationInstall: Installation<import('./liquidateMinimum.js').start>,
  *   loanTimingParams: {ChargingPeriod: ParamRecord<'nat'>, RecordingPeriod: ParamRecord<'nat'>},
  *   timerService: TimerService,
- *   priceAuthority: ERef<PriceAuthority>}>} zcf
+ *   priceAuthority: ERef<PriceAuthority>,
+ *   debtToPrice?: Ratio }>} zcf
  * @param {{feeMintAccess: FeeMintAccess, initialPoserInvitation: Invitation}} privateArgs
  */
-export const start = async (zcf, privateArgs) => {
+export const start = async (zcf, { feeMintAccess, initialPoserInvitation }) => {
+  const debtMint = await zcf.registerFeeMint('RUN', feeMintAccess);
+  const { issuer: debtIssuer, brand: debtBrand } = debtMint.getIssuerRecord();
+
   const {
     ammPublicFacet,
     priceAuthority,
@@ -57,14 +65,12 @@ export const start = async (zcf, privateArgs) => {
     electionManager,
     governedParams: governedTerms,
     loanTimingParams,
+    debtToPrice = makeRatio(1n, debtBrand, 1n, debtBrand),
   } = zcf.getTerms();
 
   /** @type {Promise<GovernorPublic>} */
   const governorPublic = E(zcf.getZoeService()).getPublicFacet(electionManager);
 
-  const { feeMintAccess, initialPoserInvitation } = privateArgs;
-  const debtMint = await zcf.registerFeeMint('RUN', feeMintAccess);
-  const { issuer: debtIssuer, brand: debtBrand } = debtMint.getIssuerRecord();
   zcf.setTestJig(() => ({
     runIssuerRecord: debtMint.getIssuerRecord(),
   }));
@@ -131,6 +137,7 @@ export const start = async (zcf, privateArgs) => {
     collateralIssuer,
     collateralKeyword,
     initialParamValues,
+    collateralPriceBrand,
   ) => {
     await zcf.saveIssuer(collateralIssuer, collateralKeyword);
     const collateralBrand = zcf.getBrandForIssuer(collateralIssuer);
@@ -152,13 +159,18 @@ export const start = async (zcf, privateArgs) => {
     );
     const liquidationStrategy = makeLiquidationStrategy(liquidationFacet);
 
-    const startTimeStamp = await E(timerService).getCurrentTimestamp();
+    const collateralToPrice = await makeUnitConversionRatio(
+      collateralBrand,
+      collateralPriceBrand || collateralBrand,
+    );
 
+    const startTimeStamp = await E(timerService).getCurrentTimestamp();
     const vm = makeVaultManager(
       zcf,
       debtMint,
-      collateralBrand,
       priceAuthority,
+      collateralToPrice,
+      debtToPrice,
       loanTimingParams,
       vaultParamManager.readonly(),
       mintAndReallocate,
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -3,7 +3,6 @@
 import '@agoric/zoe/exported.js';
 
 import { E } from '@endo/eventual-send';
-import { Nat } from '@agoric/nat';
 import {
   assertProposalShape,
   makeRatioFromAmounts,
@@ -13,6 +12,7 @@ import {
   ceilDivideBy,
   makeRatio,
   floorDivideBy,
+  floorMultiplyBy,
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { makeNotifierKit, observeNotifier } from '@agoric/notifier';
 import { AmountMath } from '@agoric/ertp';
@@ -45,8 +45,9 @@ const trace = makeTracer('VM');
  *
  * @param {ZCF} zcf
  * @param {ZCFMint<'nat'>} debtMint
- * @param {Brand} collateralBrand
  * @param {ERef<PriceAuthority>} priceAuthority
+ * @param {Ratio} collateralToPrice not a tunable parameter; determined by priceAuthority
+ * @param {Ratio} debtToPrice not a tunable parameter; determined by priceAuthority
  * @param {{
  *  ChargingPeriod: ParamRecord<'nat'>
  *  RecordingPeriod: ParamRecord<'nat'>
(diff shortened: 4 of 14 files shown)


packages/vats/src/core/utils.js
```diff
@@ -63,6 +66,52 @@ export const feeIssuerConfig = {
...
 
+/**
+ * @template {AssetKind} K
+ *
+ * Create a brand that cannot ever produce mints, issuers, payments or purses.
+ *
+ * The allegedName becomes part of the brand in asset descriptions. The
+ * allegedName doesn't have to be a string, but it will only be used for
+ * its value. The allegedName is useful for debugging and double-checking
+ * assumptions, but should not be trusted.
```
dckc: "trusted" is kinda fuzzy... I tend to phrase this as: no code should depend on an allegedName for correctness. MarkM tends to emphasize that outside of debug consoles, an allegedName is never shown to a user.

OTOH... there's nothing novel about the use of allegedName here vs. any other ERTP usage. Perhaps it's more straightforward to leave this comment out?

packages/vats/src/core/utils.js
```diff
@@ -63,6 +66,52 @@ export const feeIssuerConfig = {
...
+ * @param {AdditionalDisplayInfo} [displayInfo={}]
+ * @returns {Brand<K>}
+ */
+export const makeInertBrand = (
+  allegedName,
+  assetKind = /** @type {K} */ (AssetKind.NAT),
+  displayInfo = harden({}),
+) => {
+  const cleanDisplayInfo = coerceDisplayInfo(harden(displayInfo), assetKind);
+  return Far(`${allegedName} brand`, {
```
dckc: is there some reason not to use `makeIssuerKit` and return only the brand?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -53,6 +53,7 @@ const trace = makeTracer('VM');
  * @param {ZCFMint<'nat'>} debtMint
  * @param {Brand} collateralBrand
  * @param {ERef<PriceAuthority>} priceAuthority
+ * @param {Ratio} debtPriceRatio not a tunable parameter; determined by priceAuthority
```
dckc: `xPriceRatio` is redundant. A price is a ratio somewhat inherently. `debtPrice` is more consistent with most code I have read.

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -1,13 +1,20 @@
...
 import {
   makeGovernedInvitation,
   CONTRACT_ELECTORATE,
   makeGovernedNat,
 } from '@agoric/governance';
-import { CENTRAL_ISSUER_NAME } from '@agoric/vats/src/core/utils.js';
+import {
+  CENTRAL_ISSUER_NAME,
+  ORACLE_PRICE_BRAND_NAME,
+  makeUnitAmount,
```
dckc: I wonder if `makeUnitAmount` belongs closer to ERTP... it looks like it belongs in the list with `makeRatioFromAmounts` above from `zoe/src/contractSupport`. Or... if it's not used outside of `run-protocol`, perhaps somewhere in `run-protocol`? I have started a `run-protocol/src/contractSupport.js` file in my runStake PR -- stuff that I'm factoring out of vaultFactory.

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -233,6 +250,7 @@ export const startVaultFactory = async (
 
   const vaultFactoryTerms = makeGovernedTerms(
     priceAuthority,
+    oraclePriceRatio,
```
dckc: Making `oraclePriceRatio` governed seems out of scope here. That's a much bigger policy decision.

Please let's just make it an ordinary immutable contract term.

Otherwise, the stakeholders that need to approve this are more than just @turadg and I, I think.

packages/run-protocol/src/vaultFactory/params.js
```diff
@@ -97,6 +98,7 @@ const makeElectorateParamManager = async (zoe, electorateInvitation) => {
  */
 const makeGovernedTerms = (
   priceAuthority,
+  debtPriceRatio,
```
dckc: Let's not add `debtPriceRatio` here. See above.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -97,6 +97,21 @@ export const bridgeCoreEval = async allPowers => {
...
 
+/**
+ * @param {BootstrapPowers} powers
+ */
+export const makeOffChainBrands = async ({
+  brand: {
+    produce: { USD: usdBrandProducer },
+  },
+}) => {
+  // Create the USD brand referred to by price oracles.
```
dckc: I suggest moving this comment to the function JSDoc comment. It's not just an implementation detail.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -97,6 +97,21 @@ export const bridgeCoreEval = async allPowers => {
 };
 harden(bridgeCoreEval);
 
+/**
+ * @param {BootstrapPowers} powers
+ */
+export const makeOffChainBrands = async ({
```
dckc: Since this is referenced in `SHARED_BOOTSTRAP_MANIFEST`, it should be in `basic-behaviors.js`, no? Please move it.

packages/vats/src/core/types.js
```diff
@@ -145,8 +147,8 @@
  *     consume: Record<WellKnownName['issuer'], Promise<Issuer>>,
  *   },
  *   brand: {
- *     produce: Record<WellKnownName['issuer'], Producer<Brand>>,
- *     consume: Record<WellKnownName['issuer'], Promise<Brand>>,
+ *     produce: Record<WellKnownName['brand'], Producer<Brand>>,
+ *     consume: Record<WellKnownName['brand'], Promise<Brand>>,
```
dckc: yikes! good catch. I wonder why that worked previously.

dckc (dismissed):
Please move `makeOffChainBrands` to `basic-behaviors.js`.

And make `debitPriceRatio` a normal immutable contract term. (I didn't review all of the vaultFactory stuff, but I couldn't help noticing this, even though it's not a bootstrap issue.)

My other comments are only suggestions to consider.

samsiegart:
>The main observable change is that now, getCollateralQuote values will be in terms of USD. This is a good thing, but may need UI work to put $ on any displayed price quotes

I think showing $ prices in the UI would be a design change, and we'd probably also need to show explicit RUN:USDC prices somewhere too. If `E(priceAuthority).quoteGiven(inputAmount, debt.brand)` changes to return a ratio of USDC:[debt.brand] then we'll need to do some change in the UI though. We could also just request the RUN price from the UI and convert prices to RUN:[debt.brand] before showing them.

packages/run-protocol/src/econ-behaviors.js
michaelfig: I don't see how to transmit it to the vaultFactory without it being in vaultFactoryTerms.  I'm hoping somebody else can advise on how to refactor this if necessary.

packages/vats/src/core/utils.js
michaelfig: That's probably the best way.  Removing this one.

packages/run-protocol/src/econ-behaviors.js
michaelfig: I put it in `zoe/src/contractSupport`.

packages/run-protocol/src/vaultFactory/vaultManager.js
michaelfig: As discussed, I'd like to open the bikeshedding for the appropriate name for this parameter.  My best suggestion so far is along the lines of `debtToPriceRatio`.

Basically, given a debt brand (`RUN`), and a price authority output brand (`USD`), have `debtToPriceRatio` of N RUN : M USD (N and M are different if the decimalPlaces don't match) so that debt amounts can be converted back and forth to price amounts.

michaelfig:
> > The main observable change is that now, getCollateralQuote values will be in terms of USD. This is a good thing, but may need UI work to put $ on any displayed price quotes

> I think showing $ prices in the UI would be a design change, and we'd probably also need to show explicit RUN:USDC prices somewhere too.

BTW, the prices are not in terms of USDC (stablecoin), just actual USD (fiat).  IIUC, the economics are that for every vault we make the operational assumption that 1.0 RUN is always exactly 1.0 USD.

> If `E(priceAuthority).quoteGiven(inputAmount, debt.brand)` changes to return a ratio of USDC:[debt.brand] then we'll need to do some change in the UI though.

It won't return a ratio.  But you'll need to do: `E(priceAuthority).quoteGiven(inputAmount, usd.brand)` to get a USD price quote.  That `usd.brand` will have its own decimalPlaces, etc, and needs to be handled differently than `debt.brand`.

> We could also just request the RUN price from the UI and convert prices to RUN:[debt.brand] before showing them.

I don't know enough about this to comment.

packages/run-protocol/src/econ-behaviors.js
dckc: sorry; my concern here is misplaced. `makeGovernedTerms` doesn't mean "make a collection of governed terms". It means "make the terms for this contract, including some that are governed." So it doesn't turn oraclePriceRatio into something that the electorate can change.

dckc (commented):
my substantive comments are addressed.

dckc (commented):
bootstrap stuff looks ok

packages/vats/src/core/utils.js
```diff
@@ -12,8 +12,8 @@ const mapEntries = (obj, f) =>
   // @ts-ignore entries() loses key type
   fromEntries(entries(obj).map(([p, v]) => f(p, v)));
 
-// TODO: phase out ./issuers.js
 export const CENTRAL_ISSUER_NAME = 'RUN';
+export const ORACLE_PRICE_BRAND_NAME = 'USD';
```
turadg: It could be used in brand or elsewhere. 
```suggestion
export const ORACLE_PRICE_NAME = 'USD';
```

packages/vats/src/core/utils.js
```diff
@@ -25,6 +25,7 @@ export const agoricNamesReserved = harden({
   brand: {
     BLD: 'Agoric staking token',
     RUN: 'Agoric RUN currency',
+    [ORACLE_PRICE_BRAND_NAME]: 'US Dollar',
```
turadg: if this should be not be the literal `USD` then the others shouldn't be either. I don't have a request on which way to go, just that they're consistent. 

My _slight_ preference is to use the string literal because it's a value that can't really change.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -264,13 +268,15 @@ export const makeInnerVault = (
   const maxDebtFor = async collateralAmount => {
     const quoteAmount = await E(priceAuthority).quoteGiven(
       collateralAmount,
-      debtBrand,
+      debtPriceRatio.denominator.brand,
```
turadg: why change this? `debtBrand` is the same value.

the fact that it's the same value should be asserted when it's constructed up above,
```
  const { brand: debtBrand } = mint.getIssuerRecord();
  assert(keyEq(debtBrand, debtPriceRatio.denominator.brand);
```

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -43,10 +46,15 @@ const { details: X } = assert;
  *  {electionManager: Instance,
  *   main: {Electorate: ParamRecord<'invitation'>},
  *   timerService: TimerService,
- *   priceAuthority: ERef<PriceAuthority>}>} zcf
+ *   priceAuthority: ERef<PriceAuthority>,
+ *   debtPriceRatio?: Ratio }>} zcf
  * @param {{feeMintAccess: FeeMintAccess, initialPoserInvitation: Invitation}} privateArgs
  */
 export const start = async (zcf, privateArgs) => {
+  const { feeMintAccess, initialPoserInvitation } = privateArgs;
```
turadg: if it's going to be at the top, might as well destructure in the params,
```suggestion
export const start = async (zcf, { feeMintAccess, initialPoserInvitation }) => {
```

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -55,14 +63,12 @@ export const start = async (zcf, privateArgs) => {
     electionManager,
     main: governedTerms,
     loanTimingParams,
+    debtPriceRatio = makeRatio(1n, runBrand, 1n, runBrand),
```
turadg: why does this need a default value?

seems like it should either be passed in always or never.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Reasons I've heard to omit `Ratio` is that's [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) which this codebase does not use. Let the type be held in the type system and not in the symbol name.

`debtToPrice` or `debtOverPrice` make sense to me to make clear what the numerator and denominator are.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -84,11 +86,10 @@ export const makeVaultManager = (
...
     async getCollateralQuote() {
       // get a quote for one unit of the collateral
-      const displayInfo = await E(collateralBrand).getDisplayInfo();
-      const decimalPlaces = displayInfo?.decimalPlaces || 0n;
+      const collateralUnit = await makeUnitAmount(collateralBrand);
       return E(priceAuthority).quoteGiven(
-        AmountMath.make(collateralBrand, 10n ** Nat(decimalPlaces)),
-        debtBrand,
+        collateralUnit,
+        debtPriceRatio.denominator.brand,
```
turadg:
```suggestion
        debtBrand,
```

packages/zoe/src/contractSupport/priceAuthority.js
```diff
@@ -5,10 +5,22 @@ import { Far } from '@endo/marshal';
...
 import { makeNotifier } from '@agoric/notifier';
 
 import '../../exported.js';
 
+/**
+ * Create a "unit" amount, which is 10**decimalPlaces for a fungible brand.
+ *
+ * @param {Brand<'nat'>} brand
+ */
+export const makeUnitAmount = async brand => {
```
turadg: This helper seems good, but are (and were) we abusing `getDisplayInfo` ? 
https://github.com/Agoric/agoric-sdk/blob/9a43d45aa6421c43a10acd34a1b898dd24b2dfb5/packages/ERTP/src/types.js#L89-L91

packages/zoe/tools/fakePriceAuthority.js
```diff
@@ -14,12 +14,14 @@ import { natSafeMath } from '../src/contractSupport/index.js';
 import './types.js';
 import '../exported.js';
 
+/** @typedef {number | bigint} Bigish */
```
turadg: this seems like a recipe for sloppiness. please consider removing.

If it kept, there must be a better name. When read out of context `Bigish` conveys nothing about what is big (there's no mention of number or int).

turadg (commented):
Took a pass. Please re-request once it's up to date with master. (Sorry if it got stale waiting for me to review; I wasn't sure what my role was on this one.)

packages/zoe/src/contractSupport/priceAuthority.js
dckc: I wondered about this too. I think it's a justifiable exception, but it does merit an explicit justification comment.

I think I'd feel better if it were only defined locally where it's being used rather than being exported as if it were encouraged for general use.

packages/run-protocol/src/vaultFactory/vault.js
michaelfig: No, `debtBrand` is a unit with a `RUN` brand.  `debtToPrice.denominator.brand` is a unit with a `USD` brand.

packages/run-protocol/src/vaultFactory/vaultFactory.js
michaelfig: The default value is for the existing users, who exactly match the debt and priceAuthority prices.  I did not want to complicate the simplest uses of the vault contracts by requiring this parameter to be set every time.

packages/run-protocol/src/vaultFactory/vaultManager.js
michaelfig: I will use `debtToPrice` and `collateralToPrice`.

packages/run-protocol/src/vaultFactory/vaultManager.js
michaelfig: No, `debtToPrice.denominator.brand`.  This is the substance of the changes I'm making here.

packages/vats/src/core/utils.js
michaelfig: No, it can only be used as a brand.  I'll rename to `ORACLE_BRAND_PRICE_NAME`.

packages/zoe/src/contractSupport/priceAuthority.js
michaelfig: I tried defining locally, but it was repulsive to me.  I will add the justification comment.

packages/zoe/tools/fakePriceAuthority.js
michaelfig: `Numeric`?

dtribble: Nat will coerce either, I think. Any reason not to require the caller to do that conversion?

dckc:
obsolete in favor of #5190, I'm pretty sure.

PR #4863, opened 2022-03-18
Create CODEOWNERS

### LINKED ISSUE #3382: Add code owners

We need to tell github about codeowners so that bugs and reviews are automatically assigned

Tartuffo:
Added the CODEOWNERS file that we iterated on as a team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,47 +1,47 @@
-# Default code owners when not specified.  
+# Default code owners when not specified.
 # We prefer to explicitly specify owners for all subdirectories.
-*                                warner, dckc, Chris-Hibbert, FUDCo, michaelfig
+*                                @warner, @dckc, @Chris-Hibbert, @FUDCo, @michaelfig
 
-/* @Agoric/codeowner-tooling    # Toplevel files only
-/.github/                        kriskowal, michaelfig, erights
-/docker/                         kriskowal, michaelfig, erights
-/patches/                        kriskowal, michaelfig, erights
-/scripts/                        kriskowal, michaelfig, erights
-/yarn.lock                       mhofman
+/*                               @kriskowal, @michaelfig, @erights
+/.github/                        @kriskowal, @michaelfig, @erights
+/docker/                         @kriskowal, @michaelfig, @erights
+/patches/                        @kriskowal, @michaelfig, @erights
+/scripts/                        @kriskowal, @michaelfig, @erights
+/yarn.lock                       @mhofman
 
-# Golang parts of the repo.
-/go.*                            JimLarson
-/golang/                         JimLarson
+# Golang @parts @of @the @repo.
+/go.*                            @JimLarson
+/golang/                         @JimLarson
 
-# Explicit owners for our packages
-/packages/ERTP/                  erights, hibbert
-/packages/SwingSet/              warner, FUDCo
-/packages/access-token/          kriskowal, michaelfig
-/packages/agoric-cli/            kriskowal, michaelfig
-/packages/assert/                kriskowal, turadg, michaelfig, erights
-/packages/cosmic-swingset/       michaelfig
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,47 @@
+# Default code owners when not specified.  
+# We prefer to explicitly specify owners for all subdirectories.
+*                                warner, dckc, Chris-Hibbert, FUDCo, michaelfig
```
dckc: @warner would you please refine this list or clarify expectations?

I'm uneasy about being in this list. I don't understand the expectations. I should have noted this earlier, I suppose, but I supposed that I'm obliged a much as anyone else, and I expected clarity to arise in due course. Surely this doesn't mean that all 5 us need to review every change to every misc. file?

I think it's important to have @warner here. And there should be at least other person for the usual "redundancy in all critical systems" reasons. But 5 seems like a recipe for anybody/somebody/nobody dysfunction.

dckc (changes requested):
The robot doesn't like the syntax. It looks like an `@` is missing from most of the names. And the `codeowner-tooling` team doesn't exist.

.github/CODEOWNERS
Tartuffo: My understanding is that this sets the _default_ reviewer for the various directories, and one goal here is to prevent PRs being created by external people which never get reviewed because the external person did not now who to add as a reviewer.  I don't think there is any other implicit duty that comes to you as an "owner" on this.

The * clause on line 3 should almost never apply, as the explicit settings lower in the file take precedence.  But if you want to be taken off the list, I'm happy to do that.

FUDCo (commented):
Substantively this seems fine.  I presume the syntax issues will be taken care of.

Tartuffo:
Yes, I fixed the syntax in my local checkout, waiting to see if Dan wants to be removed from the * clause before committing.

.github/CODEOWNERS
```diff
@@ -0,0 +1,47 @@
...
+/packages/notifier/              kriskowal, turadg, michaelfig, erights
+/packages/pegasus/               michaelfig
+/packages/run-protocol           hibbert, turadg
+/packages/same-structure/        kriskowal, turadg, michaelfig, erights
+/packages/sharing-service/       michaelfig, hibbert
+/packages/solo/                  kriskowal, michaelfig
+/packages/sparse-ints/           kriskowal, turadg, michaelfig
+/packages/spawner/               michaelfig
+/packages/stat-logger/           warner, FUDCo
+/packages/store/                 warner, FUDCo, erights
```
warner: If there's any suggestion of priority here, I'd put `erights` first, then `FUDCo`, then me.. I've barely touched that module.

.github/CODEOWNERS
warner: I'd say the expectation is that whoever the robot picks as a reviewer should either do the review or pick a more appropriate reviewer. Anything is better than ignoring it, and any of us have more information about who the best reviewer should be than an outsider.

warner (approved):
You'll have to fix the syntax first.. I've got no idea how that works. But the owner assignments look good to me.

.github/CODEOWNERS
dckc: The robot will pick just one? I thought it would pick all five, leading to anybody/somebody/nobody syndrome.

I guess I don't feel that strongly. Let's try it and see.

Tartuffo: I fixed the syntax problem in the file.  Please review again!

.github/CODEOWNERS
Tartuffo: I don't think there is a priority, I think if no reviewer explicitly assigned, it'll ask all to review.

dckc (approved):
The robot likes the syntax now. Let's give it a whirl.

PR #4865, opened 2022-03-18
feat(swingset): add "device hooks"

### LINKED ISSUE #4726: implement device "hooks" to invoke kernel endowments with dref-kref translation

To facilitate #4381, I need a way for `device-vat-admin` to pass capdata into the kernel endowment through a path that will translate the device-side slots (drefs) into kernel-side ones (krefs). The current endowment mechanism just gives the device a kernel function to call (like `getBundle` or `pushCreateVatIDEvent`), but the kernel function does not know which device is calling it, and does not have access to the `deviceTranslator.js` objects to perform the translation.

This looks a lot like a syscall. Devices can use `syscall.sendOnly`, which translates dref-to-kref and then pushes a kref-based message onto the run-queue. So I'm thinking I should add a syscall that can invoke pre-arranged "kernel hooks".

## Description of the Design

The kernel would configure these for each device by providing a record of named functions. Each function will be called with kref-based capdata. On the device side, a new `syscall.callKernelHook` is used: it takes a hook name and an argument. `deviceManager.js` arranges for the syscall to translate the argument into kref space, invoke the given kernel function, then translate any return value back into dref space:

```js
// device code:
import { buildSerializationTools } from '.../deviceTools.js';
const { serialize, unserialize, deviceNodeForSlot, presenceForSlot } = buildSerializationTools(syscall, deviceName);

const argCapdata = serialize([deviceNodeForSlot('d+1'), presenceForSlot('o-2')]);
const resultCapdata = syscall.callKernelHook('add', argCapdata);
const result = unserialize(resultCapData);
```

This is the moral equivalent of a vat doing something like `result = D(hooks).add([devNode, presence])`.

Within the kernel.js `start` function, the hook would be configured with:
```js
deviceHooks[deviceName] = {
  add(argCapData) { ... ; return resultCapData; },
};
```

where the kernel-side hook receives and returns (kref-space) capdata. Translation occurs in both directions.

All arguments passed through `callKernelHook()` must be serializable: devices have access to external "magic" endowment functions, but they should not send these through `callKernelHook`. The serializer provided by `buildSerializationTools` will not know what to do with such endowments (it is much more limited than the one used by liveslots).

## Security Considerations

Devices are pretty powerful. They're run in their own compartment, but they share an engine with the kernel, so an infinite loop will prevent the kernel from making progress. They are subject to ocap protections (the `syscall` they receive is hardened and defensive, etc), but if the host application chooses to provide a device with powerful endowments, the device becomes as powerful as those endowments.

These hooks provide a way to limit that power somewhat, at least they provide a data-only interface to things outside the device. It might be interesting to pursue this model for externally-provided endowments (like the Mailbox device getting access to the host-provided storage buffer), and also for input endowments (q.v. #720). I'd set aside the name `syscall.callExternalHook` or `callEndowedHook` or something for that path.

## Test Plan

Testing this might be tricky. These hooks only make sense for the kernel to configure itself: kref space isn't meaningful outside the kernel (although test functions like `controller.queueToKref` suggest otherwise). For regular devices, a unit test can configure the device and its endowments in the same way normal host applications would: `config.devices` and the `deviceEndowments` that is passed as the second argument to `makeSwingsetController()`. But if these are only meaningful when provided by the kernel, it would be awkward to introduce a user-visible way to inject them.

The built-in kernel endowments would be limited to vat-admin (indeed that's the primary, and thus far only use case for this feature). And I wouldn't want test functionality to be hanging off a device or endowment that's used in production.

One possibility is to put an `addTestDeviceHook` property on `runtimeOptions`. If present, it would be a function that would be used by a hook made available to device-vat-admin.

Another is to add `kernel.addDeviceHook(deviceName, hook)`. Calling this would require access to the `kernel` object, not just the `controller`, which matches the pattern used in a handful of unit tests (`test-kernel.js` among them), and would be hidden from normal userspace.

warner:
feat(swingset): add "device hooks"

Raw devices can use this feature to call kernel functions through a "hook"
that translates device-side slots ("drefs") into kernel-side slots ("krefs").
This will make it possible for devices to submit capdata to kernel endowments
like `pushCreateVatIDEvent`, so vatParameters can carry slots.

closes #4726

### DIFF
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -240,11 +240,17 @@ export function swingsetIsInitialized(hostStorage) {
   return !!hostStorage.kvStore.get('initialized');
 }
 
+/** @typedef {{ kernelBundles?: Record<string, string>, verbose?: boolean,
+ *              addVatAdmin?: boolean, addComms?: boolean, addVattp?: boolean,
+ *              addTimer?: boolean,
+ *            }} InitializationOptions
+ */
+
 /**
  * @param {SwingSetConfig} config
  * @param {string[]} argv
  * @param {HostStore} hostStorage
- * @param {{ kernelBundles?: Record<string, string>, verbose?: boolean }} initializationOptions
+ * @param {InitializationOptions} initializationOptions
  * @param {{ env?: Record<string, string | undefined > }} runtimeOptions
  */
 export async function initializeSwingset(
@@ -300,6 +306,10 @@ export async function initializeSwingset(
       bundleFormat: config.bundleFormat,
     }),
     verbose,
+    addVatAdmin = true,
+    addComms = true,
+    addVattp = true,
+    addTimer = true,
   } = initializationOptions;
 
   kvStore.set('kernelBundle', JSON.stringify(kernelBundles.kernel));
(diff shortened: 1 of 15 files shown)


packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -244,7 +244,10 @@ export function swingsetIsInitialized(hostStorage) {
  * @param {SwingSetConfig} config
  * @param {string[]} argv
  * @param {HostStore} hostStorage
- * @param {{ kernelBundles?: Record<string, string>, verbose?: boolean }} initializationOptions
+ * @param {{ kernelBundles?: Record<string, string>, verbose?: boolean,
+ *           addVatAdmin?: boolean, addComms?: boolean, addVattp?: boolean,
+ *           addTimer?: boolean,
+ *         }} initializationOptions
```
FUDCo: This type seems brittle.  Maybe at least define it by name somewhere?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -126,6 +126,16 @@ export default function buildKernel(
     log: [],
   };
 
+  /**
+   * @typedef { (args: SwingSetCapData) => SwingSetCapData } DeviceHook
+   * @typedef { string } HookName
+   * @typedef { Record<HookName, DeviceHook> } OneDeviceHooks
```
FUDCo: `OneDeviceHooks` is not a good name.

FUDCo (approved):
Looks good aside from a couple of minor aesthetic nits.

packages/SwingSet/src/controller/initializeSwingset.js
warner: yeah, I suspect we have multiple definitions for this particular thing, that need to be consolidated, but I'll poke around and see if defining `InitializationOptions` is an option

packages/SwingSet/src/kernel/kernel.js
warner: changed to `HooksForOneDevice`

PR #4868, opened 2022-03-18
refactor(governance): less boilerplate in param handling

### LINKED ISSUE #4806: Simplify governance terms API

turadg:
closes: #4806 

## Description

**Before**
```js
  const {
    main: { [MALLEABLE_NUMBER]: numberParam, ...otherOovernedTerms },
  } = zcf.getTerms();
  const { initialPoserInvitation } = privateArgs;

  const paramManager = await makeParamManager(
    {
      [MALLEABLE_NUMBER]: ['nat', numberParam.value],
      [CONTRACT_ELECTORATE]: ['invitation', initialPoserInvitation],
    },
    zcf.getZoeService(),
  );

  const { wrapPublicFacet, wrapCreatorFacet } = handleParamGovernance(
    zcf,
    paramManager,
  );

  assertElectorateMatches(paramManager, otherOovernedTerms);
```

**After**
```js
  const { augmentPublicFacet, makeGovernorFacet } = await handleParamGovernance(
    zcf,
    privateArgs,
    {
      [MALLEABLE_NUMBER]: ParamTypes.NAT,
    },
  );
```

The contract above just makes parameters governable. An example of the contract needs to access the values can be seen in [multipoolMarketMaker changes](https://github.com/Agoric/agoric-sdk/pull/4868/files#diff-9dd2074a6f56145811acdde4b6f0e960f1d5079e59989b7c1dd8bf8b0107a2f9R165).

Basically, instead of accessing the `paramManager` directly you call getters on a new `params` object returned by `handleParamGovernance`. It's simply the subset of `paramManager` that are getters.
```
const { augmentPublicFacet, makeGovernorFacet, params } = await handleParamGovernance(
```


dckc:
The old made a `paramManager`; I don't see that in the new.

Passing all of `privateArgs` is, in general, excess authority. Is the callee actually only expecting `{ initialPoserInvitation }`?

turadg:
> The old made a paramManager; I don't see that in the new.

Right, it's best not to expose the whole param manager. Only the facets need it. The contract `start` may need readers, which are available in the response:
```
const { augmentPublicFacet, makeGovernorFacet, params } = handleParamGovernance(…);
```

Example of using that: https://github.com/Agoric/agoric-sdk/pull/4868/files#diff-9dd2074a6f56145811acdde4b6f0e960f1d5079e59989b7c1dd8bf8b0107a2f9R165 

> Passing all of privateArgs is, in general, excess authority. Is the callee actually only expecting { initialPoserInvitation }?

Yeah, I noted that with `XXX`. It only needs `initialPoserInvitation`. But I figured `handleParamGovernance` already has to be fully trusted and passing `privateArgs` ensures that the `Invitation` being passed in is the one expected by contract governance start requirements. I'm not clear that there's any security to be gained by this but totally open to it:
```
  const { augmentPublicFacet, makeGovernorFacet } = await handleParamGovernance(
    zcf,
    privateArgs.initialPoserInvitation,
    {
      [MALLEABLE_NUMBER]: ParamTypes.NAT,
    },
  );
```

dtribble:
How is the API used to actually get parameter values?  Ideally, add a usage to the before and after

dckc:
> ... It only needs `initialPoserInvitation`. But I figured `handleParamGovernance` already has to be fully trusted and passing `privateArgs` ensures that the `Invitation` being passed in is the one expected by contract governance start requirements. I'm not clear that there's any security to be gained...

`privateArgs` may include things like `feeMintAccess`, which is clearly more authority than `handleParamGovernance` needs. There's a sense in which the platform on which the code runs (swingset, XS, ...) is fully trusted, but for code running within vats, I'm not sure what "fully trusted" would mean.

Chris-Hibbert:
**Security Considerations**

This affects contract legibility. We need to ensure that the consequences for the governed contract and the governance code are compatible. We also need to ensure that contracts that don't use `contractHelper.js` (i.e. `vaultFactory`) will continue to function and be legible.

To the extent that this reduces the accessibility of the unfettered paramManager within the governed contract, it's a win.

**Documentation Considerations**

This is a change in governance.  The docs (currently only internal in `README.md` and `.../docs/*.puml`) need to be updated to reflect the changes.

**Testing Considerations**

Testing can't help much with the assurances of legibility or security, so the testing requirements are just that continuing tests continue to work.

packages/governance/src/index.js
```diff
@@ -17,7 +17,7 @@ export {
   validateQuestionFromCounter,
 } from './contractGovernor.js';
 
-export { handleParamGovernance } from './contractHelper.js';
+export { facetHelpers, handleParamGovernance } from './contractHelper.js';
```
Chris-Hibbert: Why did facetHelpers get added if it's not used?

packages/governance/src/paramGovernance/paramManager.js
```diff
@@ -94,6 +94,7 @@ const makeParamManagerBuilder = zoe => {
       getType: () => type,
     });
 
+    // names are keywords so they will necessarily be TitleCase
```
Chris-Hibbert:
```suggestion
    // names will be used as keywords so they must be converted to TitleCase
```

packages/governance/src/contractHelper.js
```diff
@@ -5,31 +5,30 @@ import { keyEQ } from '@agoric/store';
...
+import { makeParamManagerFromTerms } from './paramGovernance/typedParamManager.js';
 
 const { details: X, quote: q } = assert;
 
 /**
  * Helper for the 90% of contracts that will have only a single set of
  * parameters. In order to support managed parameters, a contract only has to
  *   - define the parameter template, which includes name, type and value
- *   - call handleParamGovernance() to get wrapPublicFacet and wrapCreatorFacet
+ *   - call facetHelpers() to get augmentPublicFacet and makeGovernorFacet
```
Chris-Hibbert: If I understand the point here, this comment needs to be updated to distinguish it from the (currently identical) comment on the new `handleParamGovernance`.  It appears that that one is for the 90% case, and this is for something else, though I'm not clear what.   I think `facetHelpers` also returns `params`, but it requires that the `paramManager` be provided explicitly. 

What's the model for when facetHelpers would be useful?

packages/governance/src/contractHelper.js
```diff
@@ -98,11 +97,45 @@ const handleParamGovernance = (zcf, paramManager) => {
...
+    makeGovernorFacet,
+    params: paramManager.readonly(),
   });
 };
+
+/**
+ * Helper for the 90% of contracts that will have only a single set of
+ * parameters. In order to support managed parameters, a contract only has to
+ *   - define the parameter template, which includes name, type and value
+ *   - call facetHelpers() to get augmentPublicFacet and makeGovernorFacet
```
Chris-Hibbert:
```suggestion
 *   - call `handleParamGovernance()` to get `augmentPublicFacet` and `makeGovernorFacet`
```

Chris-Hibbert (commented):
I didn't finish a thorough review because I'm confused about the relationship between `handleParamGovernance` and `facetHelpers`. Do they return the same things? When should each be used?

packages/governance/src/index.js
turadg: Discussed offline but explaining here for anyone else following along:
- this makes a more complete `handleParamGovernance`
- the old one was kept (because it's needed by ^) and renamed to `facetHelpers` 
- everything outside this module is using the more complete helper (`handleParamGovernance`) but this is expected so consumers have to option of getting the lighter help

packages/governance/src/paramGovernance/paramManager.js
turadg: If the keyword isn't TitleCase an error is thrown in Zoe (ERTP?). What's a better way to say that? (there's no conversion)

Chris-Hibbert: I thought [this did a conversion](https://github.com/Agoric/agoric-sdk/blob/45723b48823ed59c647d4b1f4b4d7ae16c421209/packages/governance/src/paramGovernance/typedParamManager.js#L28).

turadg: That does case convert the param _type_ and here we're talking about the param _name_. E.g. `Electorate` is a valid param name but `electorate` isn't. The type is `invitation` which to use the builder must call `addInvitation`.

Chris-Hibbert: Thanks.  I get it now.

Chris-Hibbert:
this is expected to address #4740

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -71,7 +64,7 @@ const { quote: q, details: X } = assert;
  * getLiquidityIssuer, and getLiquiditySupply.
  *
  * This contract has two parameters (poolFee and protocolFee) that are
- * managed by governance. The contract calls `handleParamGovernance()` at
+ * managed by governance. The contract calls `facetHelpers()` at
```
Chris-Hibbert: Isn't it the case that the contract still calls handleParamGovernance?  These three changes to the comments may have been what confused me about the goal of this PR.

packages/governance/src/contractHelper.js
Chris-Hibbert: I can't think of a reason why a contract would want to use `facetHelpers`.  The two cases I can think of are 1) satisfied by the new `handleParamGovenance`, and 2) doing things manually.  I would certainly not document `facetHelpers` as "for the 90% case, and I think it would be better to not export it until/unless we find a reason to want to use it.

turadg: Ok cool, I've removed it from the exports and comments 31a7d7d63940f62747c02767e7be08a2e5883183

dckc (commented):
I defer to @Chris-Hibbert on this

turadg:
@Chris-Hibbert PTAAL. While rebasing on your `governedApis` work I noticed this asymmetry:

> Governed methods and parameters must be included in terms.
> 
> ```
>   terms: {
>     governed: {
>       [MALLEABLE_NUMBER]: { type: ParamTypes.NAT, value: number },
>       [CONTRACT_ELECTORATE]: {
>         type: ParamTypes.INVITATION,
>         value: invitationAmount,
>       },
>     },
>     governedApis: ['makeItSo'],
>   },
> ```

`governedApis` are also governed terms. So instead of changing `main` to `governed` this now changes it to `governedParams`.

Separately, it got me thinking about how `makeItSo` is a method, not a a whole API. The contract has a single application-programming-interface which includes all these governable params and methods. WDYT of `governedMethods`?  (probably worth its own PR. I'm happy to take that on if it has your support.)

Chris-Hibbert:
> governedApis are also governed terms. So instead of changing main to governed this now changes it to governedParams.

That seems right to me.

> Separately, it got me thinking about how makeItSo is a method, not a a whole API. The contract has a single application-programming-interface which includes all these governable params and methods. WDYT of governedMethods? (probably worth its own PR. I'm happy to take that on if it has your support.)

I'm fine with that. I agree that governedAPIs is clumsy. I'm an adherent of OO programming, but we don't have classes, and it's often hard to figure out which part of our design is the objects, so I'm uncomfortable calling them "methods", but I don't have a suggestion I like better. "governedFunctions" is worse as a name.

packages/governance/test/swingsetTests/contractGovernor/governedContract.js
```diff
@@ -26,45 +22,25 @@ const makeTerms = (number, invitationAmount) => {
...
  * GovernedCreatorFacet<any>,
- * {
- *   electionManager: VoteOnParamChange,
- *   main: {
- *     MalleableNumber: ParamRecord<'nat'>,
- *     Electorate: ParamRecord<'invitation'>,
- *   },
- * },
+ * GovernanceTerms<{
+ *   MalleableNumber: ParamRecord<'nat'>,
```
Chris-Hibbert, line 26: Why did Electorate get dropped?   ....  Aah, now I see it. It's already in `GovernanceTerms`, good!

packages/governance/test/unitTests/test-typedParamManager.js
```diff
@@ -40,6 +41,28 @@ test('types', t => {
   );
 });
 
+test.only('makeParamManagerFromTerms', async t => {
```
Chris-Hibbert: I have a pre-commit check that warns me when I try to check in a `test.only`.   I guess this is still draft, so just a pointer.

packages/run-protocol/src/test.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Execute one step of iteration over a range of vatstore keys.
```
Chris-Hibbert: Is this supposed to be in this PR?

Chris-Hibbert (commented):
Looks good!

(Approval after it's not a draft.)

packages/governance/test/unitTests/test-typedParamManager.js
turadg: good catch! let's make CI do that: https://github.com/Agoric/agoric-sdk/issues/4941

packages/run-protocol/src/test.js
turadg: nope, thanks

Chris-Hibbert (approved):
Thanks for fixing CollateralReserve!

LGTM

PR #4869, opened 2022-03-19
feat!: add the ability to invoke an API to contract governance

Chris-Hibbert:
refs: #4188

## Description

Add to contract governance the ability to invoke particular declared methods under governance..

This is parallel to parameter governance, and uses similar mechanisms and legibility. The set of APIs that are eligible to be invoked this way are declared by the contract by listing them in an `getGovernedApis` method in the creator facet, and listing them in terms.

### Security Considerations

For legibility, clients of the contract have to be able to tell which methods can be called under governance, and they have to be able to examine the election manager to see how they can be invoked.

This needs to be documented along with the rest of governance.

### Testing Considerations

The existing test of contract governance should show that governance can invoke this.


Chris-Hibbert:
@turadg   I'm hoping you'll be willing to review the governance aspects of this PR. Otherwise I may have to call in @erights.  At a minimum, I'd like a review (and suggestions) for the TypeScript declarations.

packages/governance/src/question.js
```diff
@@ -80,6 +83,14 @@ const looksLikeParamChangeIssue = issue => {
   assertInstance(issue.contract);
 };
 
+const looksLikeApiInvocation = issue => {
```
turadg: The type annotations in the switch aren't doing what I suspect they were meant to. I think you're after  [Assertion Functions](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#assertion-functions) .

```suggestion
/**
 * @param {*} issue
 * @returns {asserts issue is ApiInvocationIssue}
 */
const looksLikeApiInvocation = issue => {
```

packages/governance/src/types.js
```diff
@@ -351,6 +351,13 @@
  * @property {ParamValue} proposedValue
  */
 
+/**
+ * @typedef {Object} ApiInvocationIssue
+ * @property {string} methodName
+ * @property {Instance} contract
+ * @property {any} methodParams
```
turadg: Is this accurate?
```suggestion
 * @property {Record<string, Callable>`} methodParams
```
(`Callable` is defined in `@agoric/eventual-send`)

packages/governance/src/types.js
```diff
@@ -490,11 +497,18 @@
...
+ */
+
+/**
+ * @template {GovernableCreatorFacet} [CF=any]
+ * @typedef {CF} LimitedCreatorFacet
  *
  * The creatorFacet for the governed contract that will be passed to the
  * responsible party. It does not have access to the paramManager.
- * @property {() => VoteOnParamChange} getContractGovernor
+ * @property {() => Instance} getContractGovernor
```
turadg, line 483: Oh good, I was confused about this in another PR.

packages/governance/src/types.js
```diff
@@ -490,11 +497,18 @@
...
 
 /**
- * @typedef {Object} LimitedCreatorFacet
+ * @typedef {{
+ *   getGovernedApis: () => unknown
+ * }} GovernableCreatorFacet
+ */
+
+/**
+ * @template {GovernableCreatorFacet} [CF=any]
```
turadg: If this isn't too onerous:
```suggestion
 * @template {GovernableCreatorFacet} [CF=unknown]
```

packages/governance/src/types.js
```diff
@@ -542,15 +557,15 @@
...
 
 /**
- * @typedef {Object} GovernedCreatorFacet
+ * @template {object} CF creator facet
+ * @typedef {CF} GovernedCreatorFacet
  * @property {() => ParamManagerRetriever} getParamMgrRetriever - allows accessing
  *   and updating governed parameters. Should only be directly accessible to the
  *   contractGovernor
- * @property {() => T} getLimitedCreatorFacet - the creator
+ * @property {() => CF & LimitedCreatorFacet} getLimitedCreatorFacet - the creator
```
turadg: `LimitedCreatorFacet` is defined as a generic type with original CF.
```suggestion
 * @property {() => LimitedCreatorFacet<CF>} getLimitedCreatorFacet - the creator
```
The union would make sense if we made the type be "additions to a governed creator facet". We've talked about the problems with that name but it's probably better to do in another PR.

packages/governance/src/types.js
```diff
@@ -593,12 +617,24 @@
...
 
+/**
+ * @typedef {Object} ApiGovernor
+ * @property {VoteOnApiInvocation} voteOnApiInvocation
+ * @property {CreatedQuestion} createdQuestion
+ */
+
+/**
+ * @callback GetUpdatedPoserFacet
+ * @returns {Promise<PoserFacet>}
```
turadg: this isn't really a callback. just inline where this is used,
```
* @param {() => Promise<PoserFacet>} getUpdatedPoserFacet
```

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
```diff
@@ -163,6 +163,27 @@ const setupElectorateChange = async (
...
+    'governanceApi',
+    [], // empty params
+    installations.binaryVoteCounter,
+    2n,
+  );
+
+  E(E(zoe).getPublicFacet(instance))
+    .getOutcome()
+    .then(outcome => log(`vote outcome: ${q(outcome)}`))
+    .catch(e => log(`vote failed ${e}`));
```
turadg, line 179: tangential, can/should our log distinguish between error messages and others? (likd stdout stderr)

turadg (commented):
> At a minimum, I'd like a review (and suggestions) for the TypeScript declarations.

☑️

> I'm hoping you'll be willing to review the governance aspects of this PR.

Will take a complete pass later today.

packages/governance/src/question.js
Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: I think I really meant Array<any>. What did you intend to convey with `Callable`?

 (I presume the single quote is just a typo, and not some syntax I haven't learned yet.)  

[after discussion]

I think we ended up with something that should mean an Array whose type will come from the contract. I put in `{[unknown]}`, hoping that that's how one writes a TBD tuple.

packages/governance/src/types.js
Chris-Hibbert: When I do that, I get *Type 'unknown' does not satisfy the constraint 'GovernableCreatorFacet'*.

packages/governance/src/types.js
Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: done

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
Chris-Hibbert: This is the log that produces golden output. Since the usage of that is just to compare all output against a correct output, I don't think it helps to split the output, or label errors distinctly.

packages/governance/src/types.js
```diff
@@ -205,6 +205,12 @@
  * @returns {VoteCounterFacets}
  */
 
+/**
+ * @callback LooksLikeApiInvocation
```
turadg: why name this as a type? the type looks good (bonus points for `unknown`) but it's making one more named type that doesn't seem to add any value over typing the `looksLikeApiInvocation` function.

If you want a generic interface to re-use, I'd rename to `AssertsApiInvocationIssue`, but even then fi we want that more generic functionality I'd say to pull it out into a utility module and have the type on the function.

packages/zoe/test/unitTests/zcf/setupZcfTest.js
```diff
@@ -48,7 +48,7 @@ export const setupZCFTest = async (issuerKeywordRecord, terms) => {
   ).startInstance(installation, issuerKeywordRecord, terms);
   const { vatAdminState } = fakeVatAdmin;
   // @ts-ignore fix types to understand that zcf is always defined
-  assert(zcf !== undefined);
+  assert(zcf !== undefined, 'zcf is required');
```
turadg: We can help the tester a little more here. Also it's idiomatic to check for presence with a truthiness.

```suggestion
  // @ts-expect-error setZCF may not have been called yet
  assert(zcf, 'zcf is required; did you forget to setZCF?');
```

Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: thanks for your (out-of-band) help fixing the delcarations.

Tartuffo:
@Chris-Hibbert is this strictly needed for MN-1 launch?

Chris-Hibbert:
> is this strictly needed for MN-1 launch?

Yes, according to @rowgraus, @dtribble, and @btulloh

packages/governance/src/types.js
erights: > thanks for your (out-of-band) help fixing the delcarations.

Does the current PR reflect this out-of-band effort yet?

Chris-Hibbert: Yep. The typescript declarations reflect the current state of the code and make a lot of the API visible in an IDE.

packages/governance/README.md
```diff
@@ -70,32 +70,37 @@ voterHandle.
...
-
-The governor starts up the Contract and can see what parameters are subject to
-governance. It provides a private facet that carries the ability to request
-votes on changing particular parameters. Some day we may figure out how to make
-the process and schedule of selecting parameter changes to vote on also subject
-to governance, but that's too many meta-levels at this point.
+We want some contracts to be able to make the manner of their governance
+legible, that is that changes to their parameters, and the ability to invoke
+some management methods are controlled by a public process. Observers (clients
+and voters) must be able to see who has control, and how those changes can
```
turadg: Glad to see this definition of this important jargon. Consider for "governed":
> Contracts can provide aspects that are _governed_ by a public process.
> Specifically,
> - changes to a defined subset of parameters
> - invocation of a defined subsets of methods

(Should we name those two aspects of governance?)

Then "legibility":

> The governance package enforces _legibility_, meaning that observers 
> (clients and voters) are able to see who has control and how those changes
> can happen. To make control of parameters legible…

packages/governance/README.md
```diff
@@ -70,32 +70,37 @@ voterHandle.
...
+request votes on changing parameters and invoking the controlled APIs. At some
+point, we may add governance control over the process and schedule of calling
+those votes, but that hasn't been planned in detail yet.
 
 The party that has the question-creating facet of the ContractGovernor can
-create a question that asks about changing a particular parameter on the
-contract instance. The electorate object creates new questions, and makes a new
-instance of a VoteCounter so everyone can see how questions will be counted.
-
+create questions on parameters or APIs for that contract instance. The
```
turadg, line 101: Out of scope and just thinking out loud: "Questions" sounds like a query for information. Questions can be answered in many ways, but these questions get resolved with exactly one answer. I suppose that's a fact of how the question is used rather than the question itself. "Resolution" would be too specific because some questions actually are queries for information (e.g. SURVEY)

packages/governance/README.md
```diff
@@ -118,8 +123,9 @@ provided by the governed contract and doesn't include any powerful governance
 capabilities. ContractGovernor starts the governed contract, so it gets the
 powerful creatorFacet. ContractGovernor needs access to the paramManager, but
 shouldn't share it. So the contractGovernor's creatorFacet provides access to
-the governed contract's publicFacet, creatorFacet, instance, and
-`voteOnParamChange()`, which the contract's owner should treat as powerful. 
+the governed contract's publicFacet, creatorFacet, instance,
```
turadg: If `publicFacet` is meant to be indicate the part of the response type then it should be backticked. If not, it should be the words "public facet". I figure it's the former since the `voteOnApiInvocation()` is backticked. Though I'm not sure it needs ()` until it's being invoked :)

packages/governance/README.md
```diff
@@ -118,8 +123,9 @@ provided by the governed contract and doesn't include any powerful governance
...
+`voteOnApiInvocation()`, and `voteOnParamChange()`, which the contract's owner
```
turadg: I can't tell which items the "which" applies to.

packages/governance/README.md
```diff
@@ -198,6 +212,21 @@ uses the values passed in `terms` to create the paramManager, reviewers of the
...
+```
+  terms: {
+    main: {
+      [MALLEABLE_NUMBER]: { type: ParamTypes.NAT, value: number },
+      [CONTRACT_ELECTORATE]: {
+        type: ParamTypes.INVITATION,
+        value: invitationAmount,
+      },
+    },
+    governedApis: ['governanceApi'],
```
turadg: `governanceApi` sounds like a term of art. consider "doThingChosenByElectorate"

packages/governance/README.md
```diff
@@ -198,6 +212,21 @@ uses the values passed in `terms` to create the paramManager, reviewers of the
 contract can verify that all and only the declared parameters are under the
 control of the paramManager and made visible to the contract's clients.
 
+Governed methods and parameters must be included in terms.
+
+```
+  terms: {
+    main: {
+      [MALLEABLE_NUMBER]: { type: ParamTypes.NAT, value: number },
```
turadg, line 225: out of scope: WDYT of changing to a tuple?
```
[MALLEABLE_NUMBER]: [ParamTypes.NAT, number],
```
Do we anticipate other elements?

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,132 @@
...
+  ElectionType,
+  coerceQuestionSpec,
+} from '../question.js';
+
+const { details: X } = assert;
+
+/**
+ * Make a pair of positions for a question about whether to invoke an API. If
+ * the vote passes, the method will be called on the governedApis facet with the
+ * parameters that were provided.
```
turadg: isn't this the _arguments_ provided? These are the actual values to be used in the call (which will be bound in the method body to its parameters)

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,132 @@
...
+    // * If the vote was negative, resolve to the outcome
+    // * If the API invocation succeeds, say so
+    // * If the update fails, reject the promise
+    // * if the vote outcome failed, reject the promise.
+    E(counterPublicFacet)
+      .getOutcome()
+      .then(outcome => {
+        if (keyEQ(positive, outcome)) {
+          E(governedApis)
+            [apiMethodName](...methodParams)
```
turadg: What is to be invoked is defined in multiple places:
- `positive`
- `outcome`
- `issue`
- local scope from parameters of `voteOnApiInvocation`

I see a check that `outcome` and `positive` match, because that's a signal as to what the outcome was.

As an extra security check, and to make clear to the reader, please check here that the outcome matches the parameters. E.g.,
```suggestion
        if (keyEQ(positive, outcome)) {
          keyEQ(outcome, { apiMethodName, methodParams });
          E(governedApis)
            [apiMethodName](...methodParams)
```

Though as noted above, I think this must be `methodArgs`.

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,132 @@
...
+  const negative = harden({ dontInvoke: apiMethodName });
+  return { positive, negative };
+};
+
+/**
+ * manage contracts that allow governance to invoke functions.
+ *
+ * @param {ERef<ZoeService>} zoe
+ * @param {Instance} governedInstance
+ * @param {any} governedApis
```
turadg:
```suggestion
 * @param {ERef<{ [methodName: string]: (...args: any) => unknown }>} governedApis
```

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,132 @@
...
+
+    const { positive, negative } = makeApiInvocationPositions(
+      apiMethodName,
+      methodParams,
+    );
+
+    /** @type {ApiInvocationIssue} */
+    const issue = harden({ apiMethodName, methodParams });
+    const questionSpec = coerceQuestionSpec({
+      method: ChoiceMethod.UNRANKED,
```
turadg, line 72: Looking up what `UNRANKED` means,
>  * "unranked" is more formally known as "approval" voting, but this is hard for
>  * people to intuit when there are only two alternatives.

Out of scope but I question this conclusion. "unranked" suggest to me "not ranked voting" which suggests the most common type of voting: pick one. I contend `ChoiceMethod.APPROVE` would be more clear. "Pick what you approve"

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,132 @@
...
+    const { publicFacet: counterPublicFacet, instance: voteCounter } = await E(
+      getUpdatedPoserFacet(),
+    ).addQuestion(voteCounterInstallation, questionSpec);
+
+    voteCounters.add(voteCounter);
+
+    // CRUCIAL: Here we wait for the voteCounter to declare an outcome, and then
+    // attempt to invoke the API if that's what the vote called for. We need to
+    // make sure that outcomeOfUpdateP is updated whatever happens.
+    //
```
turadg, line 91: what if it gets closed by exceeding the deadline? resolved negative or rejected?

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,132 @@
...
+    // * If the update fails, reject the promise
+    // * if the vote outcome failed, reject the promise.
+    E(counterPublicFacet)
+      .getOutcome()
+      .then(outcome => {
+        if (keyEQ(positive, outcome)) {
+          E(governedApis)
+            [apiMethodName](...methodParams)
+            .then(returnValue => outcomeOfUpdateP.resolve(returnValue))
+            .catch(e => {
```
turadg: Four cases defined here. If we use regular promises instead of promisekit, a throw will automatically reject. So think think this can be simplified to:
```
    const outcomeOfUpdate = E(counterPublicFacet)
      .getOutcome()
      .then(outcome => {
        if (keyEQ(positive, outcome)) {
          return E(governedApis)[apiMethodName](...methodParams);
        } else {
          return negative;
        }
      });
```
Now it's much easier for me to see that it resolves to either the return value of the method call or an object representing the position of a negative outcome. This is confusing. The term "outcome" is used here to mean "positive or negative" but "outcomeOfUpdate" returns "negative or the return value from executing the positive outcome".

What are the use cases for getting a result back from the method call? Since it's something people vote on executing, it seems like you would only execute "void" methods. In which case all the caller needs to know is a) whether the vote went positive or negative and b) that when they get this answer that if it was positive then that call is completed. IIUC, alll the other states are exceptional (e.g. failure to make remote call)

packages/governance/src/contractHelper.js
```diff
@@ -68,9 +68,9 @@ const handleParamGovernance = (zcf, paramManager) => {
   };
 
   /**
-   * @template CF
-   * @param {CF} originalCreatorFacet
-   * @returns {CF & LimitedCreatorFacet}
+   * @template {GovernableCreatorFacet} CF creator facet
```
turadg: 👍

packages/governance/src/contractHelper.js
```diff
@@ -80,20 +80,24 @@ const handleParamGovernance = (zcf, paramManager) => {
...
     // exclusively for contractGovernor, which only reveals limitedCreatorFacet
     return Far('creatorFacet', {
+      // @ts-expect-error special case
       getParamMgrRetriever: () => {
         return Far('paramRetriever', { get: () => paramManager });
       },
       getInvitation: name => paramManager.getInternalParamValue(name),
       getLimitedCreatorFacet: () => limitedCreatorFacet,
+      getGovernedApis: () => Far('governedAPIs', governedApis),
+      getGovernedApiNames: () => Object.keys(governedApis),
```
turadg, line 105: We talked about getting the API names from the governed API remote call. IIUC we can't because the whole object is Far and there's no way to make it an copyable object of method names mapped to Far methods. Is that right?

I request that be documented as explanation for the surprise that the API requires a `getGovernedApiNames` that is seemingly redundant with getting the keys of what's returned by `getGovernedApis`.

packages/governance/src/question.js
```diff
@@ -33,6 +33,7 @@ const ChoiceMethod = {
  *   PARAM_CHANGE: 'param_change',
  *   ELECTION: 'election',
  *   SURVEY: 'survey',
+ *   API_INVOCATION: 'api_invocation',
```
turadg: we can DRY this out by replacing this type annotation with,
```
const ElectionType = /** @type {const} */ {
```

packages/governance/src/question.js
```diff
@@ -80,6 +83,15 @@ const looksLikeParamChangeIssue = issue => {
   assertInstance(issue.contract);
 };
 
+/** @type {LooksLikeApiInvocation} */
```
turadg: please inline this type since it's only purpose is to annotate this function.

packages/governance/src/question.js
```diff
@@ -80,6 +83,15 @@ const looksLikeParamChangeIssue = issue => {
...
+const looksLikeApiInvocation = issue => {
```
turadg: We already have a word for "looks like":
```suggestion
const assertApiInvocation = issue => {
```

I think it would be good to revise the untouched `looksLike` functions too.

packages/governance/src/types.js
```diff
@@ -583,8 +606,17 @@
...
- * @param {bigint} deadline
- * @returns {ParamChangeVoteResult}
+ * @param {Timestamp} deadline
+ * @returns {ContractGovernanceVoteResult}
+ */
+
+/**
+ * @callback VoteOnApiInvocation
+ * @param {string} apiMethodName
+ * @param {[unknown]} methodParams
```
turadg: brackets with contents indicate a tuple. I think this is an array (single type for all elements and unknown length).
```suggestion
 * @param {unknown[]} methodArgs
```

packages/governance/README.md
Chris-Hibbert: To the extent the elections field has a term that encompasses both referendums and elections to office I think it's "ballot question". "ballot measures" usually seek to enact a law. I see "question" more often for initiatives, recalls, advisory measures than for elections, but people understand in the latter case that they are answering the question "who shall be selected to fill this office?"

packages/governance/README.md
Chris-Hibbert: done

packages/governance/README.md
Chris-Hibbert: clarified.

packages/governance/README.md
Chris-Hibbert: I wouldn't have a problem with that. Would it make it harder to validate (because of indeterminate order?)  main: is for governed params. If we add more kinds, then, like governedApis, they'd be placed in parallel sections.

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: fixed

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: thanks

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: I would contend that approval voting is one kind of unranked voting. The normal case for approval voting is not to limit the number of votes any voter can cast. I've seen elections for school board or other multiple-seat elections where voters are instructed to "vote for up to N", where N is usually the number of seats to be filled. I'm calling this "unranked", since it's not classical approval voting.

In approval voting, you can vote for three candidates in a one-winner race. The idea is to encourage more voters to indicate all the candidates they would find acceptable so we can find the intersection of interest groups.

You're probably right that voters wouldn't be confused by calling it APPROVE, but it doesn't seem to me to reflect voting terminology.

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: We don't yet have elections that can close before the deadline. (I think we want them for emergency actions.) 

This code doesn't attempt to decide whether the ballot counter missed its deadline.

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: Ahh!   I didn't catch that the comment on documentation also applied to the variable name in the code. I've changed the name and added the assertion.

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: Thanks for the suggestion. I simplified in both ways (dropping explicit promises, and returning the positive position on success.)

packages/governance/src/question.js
Chris-Hibbert: done.

packages/governance/src/question.js
Chris-Hibbert: done after additional consultation. It ended up needing
```
const ElectionType = /** @type {const} */ ({
```
with parens around the object.

packages/governance/README.md
Chris-Hibbert: done

packages/governance/src/contractHelper.js
Chris-Hibbert: added.

packages/governance/src/question.js
Chris-Hibbert: I think we use "looksLike" when the verification is leaky.  (`looksLikeBrand`)  But I won't say that any of these are leaky or unreliable in the same way, so I've changed them.

packages/governance/README.md
```diff
@@ -198,6 +217,21 @@ uses the values passed in `terms` to create the paramManager, reviewers of the
...
+      [CONTRACT_ELECTORATE]: {
+        type: ParamTypes.INVITATION,
+        value: invitationAmount,
+      },
+    },
+    governedApis: ['makeItSo'],
```
turadg, line 231: 🖖

packages/governance/docs/contractGovernance.puml
```diff
@@ -10,7 +10,7 @@ package "GovernedContract Vat" <<Rectangle>>  {
     }
 
     class GovernedContract {
-        <i>verifiable</i>: Governor, params
+        <i>verifiable</i>: Governor, params, governedApis
```
turadg, line 13: I looked into replacing this with Mermaid.js so it wouldn't need a build step. Unfortunately it [doesn't support package in class digrams](https://github.com/mermaid-js/mermaid/issues/1052).

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -0,0 +1,129 @@
...
+    // * If the vote passed, invoke the API, and return the positive position
+    // * If the vote was negative, return the negative position
+    // * If we can't do either, (the vote failed or the API invocation failed)
+    //   return a broken promise.
+    const outcomeOfUpdate = E(counterPublicFacet)
+      .getOutcome()
+      // @ts-expect-error return types don't appear to match
+      .then(outcome => {
+        if (keyEQ(positive, outcome)) {
+          keyEQ(outcome, harden({ apiMethodName, methodArgs }));
```
turadg: necessary change:
```suggestion
          assert(keyEQ(outcome, harden({ apiMethodName, methodArgs })));
```

packages/governance/src/contractGovernor.js
```diff
@@ -135,18 +136,73 @@ const start = async zcf => {
...
+        governedNames,
+        timer,
+        getUpdatedPoserFacet,
+      );
+    }
+
+    // if we aren't governing APIs, voteOnApiInvocation shouldn't be called
+    return {
+      voteOnApiInvocation: () => {
+        throw Error('api governance not configured');
```
turadg, line 191: I'm not certain when to use this instead but I've seen:
```suggestion
        assert.fail('api governance not configured');
```

packages/governance/src/contractGovernor.js
```diff
@@ -135,18 +136,73 @@ const start = async zcf => {
...
+      },
+      createdQuestion: () => false,
+    };
+  };
+
+  const { voteOnApiInvocation, createdQuestion } = await initApiGovernance();
```
turadg: Above there's a helpful renaming destructure: `createdQuestion: createdParamQuestion`
```suggestion
  const { voteOnApiInvocation, createdQuestion: createdApiQuestion } = await initApiGovernance();
```

packages/governance/src/question.js
```diff
@@ -78,10 +69,28 @@ const looksLikeParamChangeIssue = issue => {
   assert(issue && issue.proposedValue);
   const assertInstance = makeAssertInstance('contract');
   assertInstance(issue.contract);
+  return undefined;
```
turadg: stray, remove

packages/governance/src/question.js
```diff
@@ -90,23 +99,31 @@ const looksLikeIssueForType = (electionType, issue) => {
...
     case ElectionType.ELECTION:
-      looksLikeSimpleIssue(/** @type {SimpleIssue} */ (issue));
+      assertSimpleIssue(/** @type {SimpleIssue} */ (issue));
       break;
     case ElectionType.PARAM_CHANGE:
-      looksLikeParamChangeIssue(/** @type {ParamChangeIssue} */ (issue));
+      assertParamChangeIssue(/** @type {ParamChangeIssue} */ (issue));
+      break;
+    case ElectionType.API_INVOCATION:
+      assertApiInvocation(/** @type {ApiInvocationIssue} */ (issue));
```
turadg: remove casts. we don't know what they really are at this point

packages/governance/src/question.js
```diff
@@ -90,23 +99,31 @@ const looksLikeIssueForType = (electionType, issue) => {
...
       break;
     default:
       throw Error(`Election type unrecognized`);
   }
+  return undefined;
```
turadg: ditto as above.

if this is to make eslint happy, eslint has a bug in a rule which should be suppressed

packages/governance/src/types.js
```diff
@@ -351,6 +319,12 @@
  * @property {ParamValue} proposedValue
  */
 
+/**
+ * @typedef {Object} ApiInvocationIssue
+ * @property {string} apiMethodName
+ * @property {[unknown]} methodArgs
```
turadg: array, not 1-tuple
```suggestion
 * @property {unknown[]} methodArgs
```

packages/governance/src/types.js
```diff
@@ -450,6 +424,17 @@
  * @property {ParamValue} proposedValue
  */
 
+/**
+ * @typedef {Object} InvokeApiPosition
+ * @property {string} apiMethodName
+ * @property {any} methodArgs
```
turadg:
```suggestion
 * @property {unknown[]} methodArgs
```
or `any[]` if that makes a lot of errors. but we at least know that it's an array

packages/governance/src/types.js
```diff
@@ -490,11 +475,17 @@
  */
 
 /**
- * @typedef {Object} LimitedCreatorFacet
+ * @typedef {{
+ * }} GovernableCreatorFacet
```
turadg: I forget, is there really no requirement on a creator facet to be governed?

if so, why name it?

packages/governance/src/types.js
```diff
@@ -583,8 +575,17 @@
...
- * @param {bigint} deadline
- * @returns {ParamChangeVoteResult}
+ * @param {Timestamp} deadline
+ * @returns {ContractGovernanceVoteResult}
+ */
+
+/**
+ * @callback VoteOnApiInvocation
+ * @param {string} apiMethodName
+ * @param {[unknown]} methodArgs
```
turadg:
```suggestion
 * @param {unknown[]} methodArgs
```

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -240,6 +240,8 @@ export const start = async (zcf, privateArgs) => {
     getParamMgrRetriever,
     getInvitation: electorateParamManager.getInternalParamValue,
     getLimitedCreatorFacet: () => vaultFactory,
+    getGovernedApis: () => harden({}),
+    getGovernedApiNames: () => harden({}),
```
turadg, line 278: I thought we said that undefined getters would be treated the same as getters returning empty. This is busywork for the contract author.

turadg (commented):
Very close!

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: good catch.

packages/governance/src/contractGovernor.js
Chris-Hibbert: `throw Error` is much more common in contracts.  I'm not sure that `assert.fail` doesn't work as well, but I know that the former works well with our distributed stack trace utilities.

packages/governance/src/contractGovernor.js
Chris-Hibbert: good call.

packages/governance/src/question.js
Chris-Hibbert: It makes the `returns ... assert` not need an `@typescript-ignore`.

packages/governance/src/question.js
Chris-Hibbert: That's what it's for. I thought the return was less ugly than the `ignore`.  If your claim is that the broken rule should be called out, I'll change it back.

packages/governance/src/question.js
Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: Everything required is provided. I'll drop it.  I don't think we'll be adding requirements back.

packages/governance/src/types.js
Chris-Hibbert: done

packages/run-protocol/src/vaultFactory/vaultFactory.js
turadg: This was clarified in chat: the auto-providing is done by `handleParamGovernance` which vaultFactory eschews (because it's an uncommonly complex governance)

turadg (approved):
👍

Issue #4864, opened 2022-03-18
replace device-vat-admin with a kernel-owned device node


warner:
In a discussion with @FUDCo about how I'm actually using this to achieve #4381, it occurred to me that an alternative approach would be to implement "kernel-provided device nodes", in which the kernel would set `dNN.owner = null` to indicate that the kernel owns a device node. So when `vat-vat-admin` does `syscall.callNow(deviceVatAdmin, 'createVat', args)`, it would get routed to the kernel instead of `device-vat-admin` (and `device-vat-admin` would go away).

This would have a lot of benefits over the PR I'm working on:
* with no `device-vat-admin`, userspace is no longer obligated to help the kernel wire up its own internal components
* the bootstrap vat no longer has a way to forge `newVatCallback` or vat-termination messages
* the lack of GC in devices no longer causes vatParameter objects to get retained forever

This would not require changes to the code in `vat-vat-admin` (it will keep making `D()` calls), but `device-vat-admin` goes away, userspace `bootstrap()` functions must change to replace their `vatAdminService = await E(vats.vatAdmin).createVatAdminService(devices.vatAdmin)` with `vatAdminService = vats.vatAdmin`.

It would require `initializeSwingSet` to allocate the new kernel-owned device node, and include it in the `vatParameters` arguments passed to vat-vat-admin, effectively moving the `createVatAdminService` call into `buildRootObject`.

_Originally posted by @warner in https://github.com/Agoric/agoric-sdk/issues/4726#issuecomment-1072675872_

PR #4872, opened 2022-03-19
feat(swingset): allow object refs in create/upgrade/terminate vat

### LINKED ISSUE #4381: allow bundlecaps in vatParameters

### LINKED ISSUE #4588: slots in vatAdmin terminate/createVat are not translated correctly
While working on #4566 I realized that we're probably mishandling object references (slots) in the "terminate vat from the outside" case:
```js
// this is vat 1
  const { admin } = await E(vatAdminService).createVat(bundlecap);
  const obj = Far('iface', {});
  const reason = [ 'data', obj ]; // usually an Error, but anything is legal
  E(admin).terminateWithFailure(reason);
```

The object `obj` is assigned a vat-1 vref as `terminateWithFailure` gets serialized for `syscall.send`, then assigned a kref as the syscall is translated. When the message is delivered to `vat-admin`, the c-list allocates a vat-admin vref, which is what appears at the `vatAdminWrapper.js` `terminateWithFailure()` call:
```js
    const adminNode = Far('adminNode', {
      terminateWithFailure(reason) {
        D(vatAdminNode).terminateWithFailure(vatID, reason);
      },
```

The `D(vatAdminNode)` does a `syscall.callNow`, which translates the vref through the kref into a `devices.vatAdmin` dref, which is what arrives at the device code.

Previously (on current trunk), the device code was written with `deviceSlots`, which provided a copy of its `m.serialize` to the device code, so `vatAdmin-src.js` could re-serialize the `reason` object into capdata. `vatAdmin-src.js` then passed the `reason` capdata to its kernel endowment (`terminate(vatID, serialize(reason))`).

The problem is that `serialize(reason)` is *device-side* capdata, with "drefs" (the device equivalent of "vrefs" in a vat). But the kernel `terminate` endowment is going to use that `reason` capdata to reject the `admin.done()` promise: the kernel enqueues a `vatTerminated` message to vat-admin, and `queueToKref` assumes that the arguments you give it use krefs.

The consequence is that when we attempt to deliver the `vatTerminated` to vat-admin, the translator will encounter invalid slot types. This will either cause a kernel panic, or cause the failure of that message (and `admin.done()` will never fire), probably the former.

In my #4566 rewrite, I encountered the same problem in a raw-device form of `vatAdmin-src.js`. I get dref capdata from the `dispatch.invoke`, and I want to provide kref capdata to the kernel endowment.

The same problem exists in `createVat` with the `options` argument. This one is more pressing because #4381 will want to include bundlecaps in `options.vatParameters`.

For now, I'm adding an assertion that the `.slots` are empty, so we catch this in the calling vat. For most of these methods that's going to be `vatAdminWrapper.js`, which will probably cause the `terminateWithFailure()` or `createVat()` to fail, which is a perfectly reasonable way to handle it. But to implement #4381 we'll need something better.

It probably wants to be a syscall, because that provides an easy place to perform dref-to-kref translation. But after translation, we want to hit some kernel-provided API. The `terminate` case has no return value, but the `createVat` cases want to return a vatID from that API.

This might overlap with #720 (a kernel input queue). One thought is that the built-in device might be given an endowment which is a marker/handle object (one for `terminate`, another one or two for `createVat`/`createVatByBundle`). We add a `syscall.triggerKernelSomething(handle, deviceCapdata)`, the syscall translates `deviceCapdata` into `kernelCapdata`, then invokes whatever function the kernel had previously registered with `handle` (possibly through a queue that protects the kernel from reentrancy).

The way this interacts with #720 would be that external callers could invoke the device's exterior APIs any time they want, and `syscall.triggerKernelSomething` would know to queue the actual translation and invocation until the kernel was safely idle. I'm not sure that's sound.

warner:
This allows `E(vatAdminService).createVat(bundleID, { vatParameters })` to
include object references in `vatParameters`, which are then delivered
through the `dispatch.startVat()` delivery and made available to
`buildRootObject(vatPowers, vatParameters)`.

In vat-vat-admin, the CreateVatOptions validation was rewritten to
ensure that any existing slots are only in vatParameters.

From within a vat, `vatPowers.exitVat(completion)` and
`.exitVatWithFailure(reason)` can include object references, and they will be
delivered to the parent caller's `done` promise.

From outside the vat, `E(adminNode).terminateWithFailure(reason)` can take
object refs in `reason` and they will be delivered to the parent caller's
`done` promise. `E(adminNode).upgrade(bundleID, vatParameters)` can take
object refs in `vatParameters` just like `createVat` (although `upgrade`
itself is still non-functional).

The kernel will maintain a refcount on each object passed through this
mechanism, to keep it from being collected while in transit.

This is implemented with the new "kernel device hooks" feature, which allows
a device to call into the kernel and have its drefs translated into
krefs.

This will help with ZCF/contract vat upgrade, to pass the new contract
bundlecap into the new ZCF vat via vatParameters.

closes #4588
closes #4381
refs #1848

### DIFF
--- a/packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
+++ b/packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
@@ -93,9 +93,6 @@ bundleID before submitting to the kernel), or (temporarily) a full bundle.
 
 export function buildDevice(tools, endowments) {
   const { hasBundle, getBundle, getNamedBundleID } = endowments;
-  const { pushCreateVatBundleEvent, pushCreateVatIDEvent } = endowments;
-  const { pushUpgradeVatEvent } = endowments;
-  const { terminate } = endowments;
   const { meterCreate, meterAddRemaining, meterSetThreshold, meterGet } =
     endowments;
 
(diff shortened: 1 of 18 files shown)


packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
```diff
@@ -194,65 +191,43 @@ export function buildDevice(tools, endowments) {
...
         // D(devices.vatAdmin).createByBundle(bundle, options) -> vatID
         if (method === 'createByBundle') {
-          // see #4588
-          assert.equal(argsCapdata.slots.length, 0, 'cannot handle refs');
-          const args = unserialize(argsCapdata);
-          const [bundle, options] = args;
-          // options cannot hold caps yet, but I expect #4486 will remove
-          // this case before #4381 needs them here
-          const vatID = pushCreateVatBundleEvent(bundle, options);
+          const res = syscall.callKernelHook('createByBundle', argsCapdata);
```
FUDCo, line 194: Presumably the syscall implementation does parameter validation, but it feels weird to be passing a capdata received from the outside world directly into the kernel without looking at it.

FUDCo (approved):
I like it.  It feels simpler and cleaner than its predecessor.

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
warner: Yeah, `vat-vat-admin` does the user-visible-error -producing validation (in `convertOptions`), and the device basically just fronts for the kernel hook. (hence #4864 being about cutting out `device-vat-admin.js` entirely).

Issue #3062, opened 2021-05-10
vat upgrade by replacing virtual-object behavior record


warner:
## What is the Problem Being Solved?

@dtribble and I were brainstorming about upgrade pathways. I'm thinking of a taxonomy of upgrades, rated by depth/intricacy:

1. changing a parameter, like an interest rate, or interval duration
2. changing the behavior of some pure function in userspace vat code
3. changing the behavior of data/reference-bearing userspace vat code (retaining the old data, but changing the code that runs against it)
4. changing liveslots
5. changing the kernel

The first two can be accomplished by having some object in the vat code which knows the interest rate or function in question. Operations ask it for the rate every time, or the submit data to the function and get something back. The vat code which initially creates this will create a special facet that allows the value to be changed, and it will give this facet to a governance contract. A governance decision (using whatever voting mechanism that other contract cares for) can invoke the facet and give it a new number, or a new bundle of code to evaluate and overwrite the old function.

But the third one is interesting. As an example, think about an Issuer which has some (large) number of outstanding Purses, Payments, and DepositFacets. We want to change the implementation, but retain the validity of all those exported objects.

## Description of the Design

@dtribble 's idea was to lean on the fact that we're representing all of those things as virtual objects. Each call to `makeKind()` is effectively creating a table, and providing both a behavior record and a function to create new rows. We'll need to give a name to each table: perhaps we add a `name` argument to `makeKind` and we throw an error if it is not unique.

Now, the upgraded vat wants to connect new behavior to the old records. So imagine a `vatPowers.replaceKind(name, behavior, initFunction)` that does mostly the same thing as `makeKind` *but* instead of allocating a new table, it re-uses an old one. We might add arguments which let the upgraded code know when it's being given an old record (maybe a version number along with each `state` object), so it can react differently to the old data.

To trigger this upgrade, we might use the `vatAdmin` facet and add a method which behaves like the "spawn zygote vat" API (#2268), except that it takes a new vat code bundle. When invoked, a new vatID is created, all the virtual-object exports of the old vat are re-targeted to point at the new one, all the non-virtual-object exports are left on the old vat, and the old vat is terminated. This would retain the identity of things which we can't carry over, they'd just be dead references (like exports of a terminated vat).

[...]

Issue #4390, opened 2022-01-26
clean up null-vs-undefined in VatSyscallResults


warner:
kernel.js `vatSyscallHandler` has a comment:

```js
// Vats (liveslots) record the response in the transcript
// (which is why we use 'null' instead of 'undefined', TODO clean
// this up), but otherwise most syscalls ignore it. 
```

The task is to clean this up. Ideally we use exactly one of 'null' or 'undefined' everywhere. The major constraint is that the result is stored in the transcript, which does a `JSON.stringify`, which tends to map both `null` and `undefined` into the same thing.

This also touches the vat translators, and the return values of most of the syscalls in `kernelSyscall.js`.

erights:
> a `JSON.stringify`, which tends to map both null and undefined into the same thing.

```js
JSON.stringify({x: null, y: undefined}); // '{"x":null}'
```

Tartuffo:
Also applies to sys call params.

Tartuffo:
@erights and @warner please review and determine if we really need to do this for MN-1.

PR #4873, opened 2022-03-19
Implement baggage

### LINKED ISSUE #4325: "baggage": prepare for vat upgrade, deliver durable objects to successor

As part of #1848 and #2004, we need a way for version-1 of a vat to prepare for upgrade to version-2. The earlier version should store critical data (anything necessary for version-2 to fulfill the obligations impliedexpressed by version-1's durable exports) in durable storage, where it can be accessed by version-2. But then version-2 needs a way to reach it.

Our plan is to define a special durable `BigMapStore`, colloquially named "the baggage", with at least one string key. As a durable collection, it can only hold durable things (pass-by-copy data, imported Presences, exported Remotables that can be reattached at upgrade, exported virtual objects whose Kinds can be reattached at upgrade, and other durable collections). The handle to the baggage is provided to each version at startup, through some special TBD argument (maybe `vatPowers.baggage`).

In addition to durable data, the baggage should be able to hold whatever sort of "reattachment handle" / "identity resumption/reclamation facet" / (I haven't found a good word yet) that #3062 exposes, through which version-2 can provide new behavior records for the virtual-object Kinds that the earlier version created. This provides a user-code -legible way to connect the original Kind definition in version-1, through a named path in the baggage, to the code in version-2 that attaches a new Kind definition to that named path.

The metaphor is that version-1 of the vat is a business owner preparing for retirement, and their shop is going to be moved to a new building. Version-1 puts important documents and inventory in durable collections, and puts references to those collections in the "baggage". Upon retirement (upgrade), everything that's durable is moved to the new building, and anything that isn't durable is left behind. But these are ocaps, so merely being present in the new building is not sufficient: you need an access path. Version-2 is the new owner in the new location, who is handed the "baggage" by the estate/transfer agent. Anything not transitively reachable from the baggage is unreachable, even it if was durable. Both the old and new owners can use Store access to store/retrieve their data, e.g. version-1 does `baggage.set('pending orders', pendingOrdersCollection)`, and version-2 does `pendingOrdersCollection = baggage.get('pending orders')`.

The baggage Store can use non-string keys too, later, but the starting point *must* be a string (or Number, or other flat data), because version-2 starts with nothing but the baggage and the source code they rode in on. That source code can contain data but not objects. The baggage is the only real "object" they get from version-1.

## Description of the Design

TBD

Obviously we should find a better name than "baggage". "upgrade data"?

Maybe deliver as `vatPowers.baggage`, which limits access to just `buildRootObject` and whatever function version-2++ uses at startup (maybe also `buildRootObject`, maybe something like `upgrade` or `start`). Another option is an additional argument to `buildRootObject`.

## Security Considerations

All sorts of critical (mutable!) state is reachable from the baggage, so it must be closely held and thoughtfully guarded. It's easy to store a hierarchy of data in there, so POLA should be exercised (give `pendingOrders` to the order-handling code, rather than all baggage). Clearly the baggage must not be exposed as a global.

## Test Plan

Unit tests to show `baggage` being present on `vatPowers` and usability as a normal collection.

Unit/integration tests to populate `baggage` with some plausible use-case data (order book, some Kind handles) and then demonstrate a version-2 upgrade path which then reads from baggage and restores functionality using the data therein.

## Documentation Plan

`docs/vat-upgrade.md` should have a section on the baggage, what it offers, how to use it, when to *not* use it (intentionally abandoned non-durable data), and some examples.

FUDCo:
This implements the primary vat-side interface for upgrade (#4325, #4382).  However, it does not yet implement the set of post upgrade checks (#3062) that will be required (e.g., check upon return from `buildRootObject` to verify that all the durable kinds have had behavior associated with them or have been explicitly dismissed); the ability to test or otherwise exercise the latter portion of the upgrade machinery will have to wait until the kernel-side interfaces (#1848, #3062) are in place.

I am going to tentatively declare that this PR closes #4325, leaving the remaining work to be covered by #3062.

Note to reviewers: the actual baggage implementation per se consists of the (relatively small) changes to `collectionManager.js` and `liveslots.js`.  The bulk of the rather large number of file changes in this PR are test enhancements or fixes, though implementation testing did surface a couple of other problems that are also addressed herein:

1. It turns out that you can't use the vatstore at all from vats running with the `nodeWorker` or `node-subprocess` worker types due to how the syscall communications channel between the worker and the kernel works.  This was a known limitation of these worker types, but since the baggage mechanism requires the vatstore in all cases, this limitation renders these vat types essentially unusable.  Fortunately they _are_ basically unused except for four tests which mostly just verify their existence; these tests have been disabled until we decide to either eliminate these vat worker types or make significant changes to how they communicate with the kernel process.

2. There was a problem with parameter encoding for `vatstoreGetAfter` that introduced a mismatch in some cases between how it was encoded for kernel<->worker communication and how it was encoded for transcript replay, which caused spurious anachrophobia errors on replay.  This has been addressed by tweaks to `djson.js` and `supervisor-helper.js`.

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -1165,6 +1165,7 @@ function build(
     }
 
     const vatParameters = m.unserialize(vatParametersCapData);
+    baggage = collectionManager.provideBaggage();
 
     // Below this point, user-provided code might crash or overrun a meter, so
     // any prior-to-user-code setup that can be done without reference to the
@@ -1184,7 +1185,6 @@ function build(
     );
 
     // here we finally invoke the vat code, and get back the root object
-    baggage = collectionManager.provideBaggage();
     const rootObject = buildRootObject(
       harden(vpow),
       harden(vatParameters),
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -1097,6 +1097,7 @@ function build(
     assert(key.match(/^[-\w.+/]+$/), X`invalid vatstore key`);
   }
 
+  let baggage;
   async function startVat(vatParametersCapData) {
     insistCapData(vatParametersCapData);
     assert(!didStartVat);
@@ -1183,7 +1184,12 @@ function build(
     );
 
     // here we finally invoke the vat code, and get back the root object
-    const rootObject = buildRootObject(harden(vpow), harden(vatParameters));
+    baggage = collectionManager.provideBaggage();
+    const rootObject = buildRootObject(
+      harden(vpow),
+      harden(vatParameters),
+      baggage,
+    );
     assert.equal(
       passStyleOf(rootObject),
       'remotable',
--- a/packages/SwingSet/src/supervisors/supervisor-helper.js
+++ b/packages/SwingSet/src/supervisors/supervisor-helper.js
@@ -120,12 +120,11 @@ function makeSupervisorSyscall(syscallToManager, workerCanBlock) {
     },
     vatstoreSet: (key, value) => doSyscall(['vatstoreSet', key, value]),
     vatstoreGetAfter: (priorKey, lowerBound, upperBound) => {
-      const result = doSyscall([
-        'vatstoreGetAfter',
-        priorKey,
-        lowerBound,
-        upperBound,
-      ]);
+      const syscall = ['vatstoreGetAfter', priorKey, lowerBound];
+      if (upperBound) {
+        syscall.push(upperBound);
+      }
+      const result = doSyscall(syscall);
       return result === null ? [undefined, undefined] : result;
     },
     vatstoreDelete: key => doSyscall(['vatstoreDelete', key]),
--- a/packages/SwingSet/test/workers/test-worker.js
+++ b/packages/SwingSet/test/workers/test-worker.js
@@ -65,7 +65,7 @@ test('xs vat manager', async t => {
   t.deepEqual(JSON.parse(c.kpResolution(c.bootstrapResult).body), expected);
 });
 
-test('nodeWorker vat manager', async t => {
+test.skip('nodeWorker vat manager', async t => {
   const c = await makeController('nodeWorker');
   t.teardown(c.shutdown);
 
@@ -75,7 +75,7 @@ test('nodeWorker vat manager', async t => {
   t.deepEqual(JSON.parse(c.kpResolution(c.bootstrapResult).body), expected);
 });
 
-test('node-subprocess vat manager', async t => {
+test.skip('node-subprocess vat manager', async t => {
   const c = await makeController('node-subprocess');
   t.teardown(c.shutdown);
 
--- a/packages/SwingSet/src/lib/djson.js
+++ b/packages/SwingSet/src/lib/djson.js
@@ -13,7 +13,7 @@
  * @param {any} val
  */
 function replacer(_, val) {
-  if (typeof val === 'object') {
+  if (val && typeof val === 'object') {
     const sortedObject = {};
     const names = Array.from(Object.getOwnPropertyNames(val));
     names.sort();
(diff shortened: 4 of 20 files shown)


FUDCo:
Something is very messed up here git-wise.  Somehow this PR has sucked in a bunch of commits that should already be part of the main branch and there are a lot of spurious changes in the list of changed files.  I think I'll need some help sorting this out before this is actually ready for review.

FUDCo:
Git madness appears to have been resolved.  Review away!

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1183,7 +1184,12 @@ function build(
     );
 
     // here we finally invoke the vat code, and get back the root object
-    const rootObject = buildRootObject(harden(vpow), harden(vatParameters));
+    baggage = collectionManager.provideBaggage();
```
warner: I'm thinking we should call `provideBaggage` earlier, just after the `m.unserialize(vatParametersCapData)`, for two reasons:
* it won't be metered, just in case that somehow makes a difference
* the baggage `vref` is allocated before userspace gets control, so it should be pretty stable, which I think is going to make debugging a lot easier (we don't need to check `baggageID` before interpreting DB dumps)

packages/SwingSet/src/supervisors/supervisor-helper.js
```diff
@@ -120,12 +120,11 @@ function makeSupervisorSyscall(syscallToManager, workerCanBlock) {
...
     vatstoreGetAfter: (priorKey, lowerBound, upperBound) => {
-      const result = doSyscall([
-        'vatstoreGetAfter',
-        priorKey,
-        lowerBound,
-        upperBound,
-      ]);
+      const syscall = ['vatstoreGetAfter', priorKey, lowerBound];
+      if (upperBound) {
+        syscall.push(upperBound);
```
warner, line 125: hm, I can imagine what prompted this, but I'd prefer to avoid variable-length syscall objects (e.g. `VatSyscallVatstoreGetAfter` is always a length-4 array, with the last property being of variable type, rather than sometimes it's length-3 and sometimes it's length-4). Was the problem resolved by the `djson` fix you made? Or could we resolve it by doing something like `if (!upperBound) { syscall[3] = null; }` or something?

I know we have a general problem (#4390) with null-vs-undefined as we convert through both the kernel/worker pipe and to/from the transcript on the vat-manager side, and this local issue is a consequence of that technical debt.

packages/SwingSet/test/workers/test-worker.js
```diff
@@ -65,7 +65,7 @@ test('xs vat manager', async t => {
   t.deepEqual(JSON.parse(c.kpResolution(c.bootstrapResult).body), expected);
 });
 
-test('nodeWorker vat manager', async t => {
+test.skip('nodeWorker vat manager', async t => {
```
warner, line 68: #4884 filed to remind us to decide the fate of these two worker types, some day

warner (approved):
Looks good. Let me know what you think about the two recommended changes: if they're impossible to pull off, I'm ok with landing without them, but I'd like to see if we could do it.

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: I deliberately put the creation of the baggage immediately before the invocation of `buildRootObject` so that I could be sure, without having to think too deeply about it, that the environment was properly set up (as it would have to be since you need to be able to create collections inside `buildRootObject`).  This was mostly a labor saving move on my part, since the choice of where to do this seemed otherwise relatively arbitrary.  A quick check says moving it to where you suggest shouldn't be a problem, aside from the inevitable rejiggering of some of the tests due to the resulting change in order of DB accesses.

packages/SwingSet/src/supervisors/supervisor-helper.js
FUDCo: The problem with `dsjon` was something I tripped over but wasn't ultimately _the_ problem (though it was real).  Requiring the syscall array to be fixed length is very doable, but the practical effect would simply be to move this bit of conditional logic from the place the syscall array is constructed to the place where it is used, so I'm not sure that would be a win.

PR #4874, opened 2022-03-20
Implement and display zoe offer arguments in the wallet

### LINKED ISSUE #4860: Support offerArgs in wallet.addOffer

Zoe supports an offerArgs param in `zoe.offer`, but the wallet API does not have any logic to include this arg. This should be added so a dapp can propose an offer in the wallet with specified args.

## Description of the Design

Add `offerArgs` as a param to `wallet.addOffer`, and in `executeOffer` the wallet should include the arg.

## Test Plan

A unit test should ensure the arg is being passed through.

## Additional Considerations

How should the wallet UI show these args in the proposal? Perhaps there should be a way to expand a proposal to show the stringified args. The wallet can hydrate this string using known petnames and add it to `proposalForDisplay`.

cc @michaelfig

michaelfig:
closes: #4860

## Description

`proposalTemplate.arguments` can be arbitrary passable data (including ocaps), and it will be propagated to the `zoe.offer` call.

image

### Security Considerations

The rendering of arguments in the wallet needs to be clear and unambiguous.  Currently, it's just a rehydrated marshal value.

### Testing Considerations

Needs tests.

### DIFF
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -380,7 +380,12 @@ export function makeWallet({
   };
 
   const displayProposal = proposalTemplate => {
-    const { want, give, exit = { onDemand: null } } = proposalTemplate;
+    const {
+      want,
+      give,
+      exit = { onDemand: null },
+      arguments: args,
+    } = proposalTemplate;
     const displayRecord = pursePetnameValueKeywordRecord => {
       if (pursePetnameValueKeywordRecord === undefined) {
         return undefined;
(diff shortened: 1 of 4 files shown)


michaelfig:
@samsiegart I'd love it if you could take this PR on, write tests, and improve the display (if feasible).  I needed this feature for the oracle bootstrap, so this is for MN-1.

samsiegart:
> @samsiegart I'd love it if you could take this PR on, write tests, and improve the display (if feasible). I needed this feature for the oracle bootstrap, so this is for MN-1.

Sounds good will do, thanks for getting this started!

samsiegart (approved):
I pushed a small commit for testing, if that looks good then 🚀

PR #4876, opened 2022-03-20
fix(wallet-api): don't drop unused invitations on the floor

### LINKED ISSUE #4875: Wallet backend loses track of unspent invitations
## Describe the bug

If `wallet.declineOffer` is called, the unspent invitation is not reclaimed.

## To Reproduce
1. Go to any dapp
2. Propose an offer via the wallet
3. Scroll down to note the Zoe invite purse balance
4. Click the offer "decline" button
5. no unspent invitation ever shows up in the Zoe invite purse.

## Expected behavior

After declining, the invitation should be redeposited in the Zoe invite purse, and visible in its balance.

## Platform Environment
 - what OS are you using? what version of Node.js? doesn't matter
 - is there anything special/unusual about your platform? not applicable
 - what version of the Agoric-SDK are you using? (run `git describe --ta-gs --always`) agoricdev-8-212-gaa5355cd9

## Additional context
I was testing the wallet with valuable single-use invitations.  Though most dapps hand out invitations willy-nilly, they should be treated as precious and reclaimed to be used again in a future offer if desired.

## Screenshots

michaelfig:
closes: #4875

## Description

A small but important tweak to `lib-wallet.js`.

### Security Considerations

Improved security for users of valuable invitations.

Fully compatible.

### Testing Considerations

None additional needed.

### DIFF
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -1064,11 +1064,12 @@ export function makeWallet({
     return false;
   }
 
-  function declineOffer(id) {
+  async function declineOffer(id) {
     const offer = idToOffer.get(id);
     if (consummated(offer)) {
       return;
     }
+
     // Update status, drop the proposal
     const declinedOffer = addMeta({
       ...offer,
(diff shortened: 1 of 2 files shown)


samsiegart (approved):
This is a bit confusing to me because I think there's cases where a user simply wants to discard an offer and doesn't want the invitation showing up in their purse forever. However, it's probably safer to just hold onto it.

I guess there's a separate issue in how to show invitations to the user, whether we want to even display the purse in the UI, and how to meaningfully render it in an offer if the user is purchasing an invitation.

PR #4877, opened 2022-03-20
feat(vault): move minting out of vault

### LINKED ISSUE #4838: Use a stage for minting RUN in vault

`Vault.adjustBalances` may allow users to withdraw RUN without losing associated collateral because the RUN is minted onto the `vaultSeat`.

## Description of the Design

All RUN should be minted onto a `stage` seat so that it is only ever provided to a client seat as part of a reallocate.

## Security Considerations

Operations don't check whether offer `want` can be satisfied, so clients can cause a reallocate to fail. Rather than  rigorously preventing that, it's simpler to design such that that is acceptable.

dtribble:
close: #4838 

## Description

This does several refactoring cleanups. The core and most important is moving direct mint access out of the vaults. This makes it much easier to do per-collateral limit checking and accounting.

* Expose a per-collateral facet to support collateral-specific notifiers (like compounded interest) and enable future segregation into per-collateral contracts
* Consolidate mint access into base contract for POLA in `vault`
* Setup to further refactor so that mint/burn is combined with limit checking and bookkeeping

TODO
* [x] move `burn` into manager
* [x] move `mint` function into manager so it can track per-collateral limits
* [x] integrate bookkeeping with mint/burn
* [x] fix types

### Security Considerations

These changes support better mint POLA, collateral and vault segregation, and simplify bookkeeping 

The addition of a per-collateral facet should appear in the documentation

### Testing Considerations

This is a refactoring, so all the paths fall under existing tests. With the abstraction of functionality, testing could generally be simplified.

TODO future: the new internal operation should get unit test
* [ ] Add unit tests for new internal operations
* [ ] Add unit test variant to create vaults via the collateral manager

### DIFF
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -46,16 +46,29 @@
  */
 
 /**
- * @callback ReallocateWithFee
+ * @callback MintAndReallocate
  *
- * Transfer the indicated amount to the vaultFactory's reward
- * pool, taken from the `fromSeat`. Then reallocate over all the seat
- * arguments and the rewardPoolSeat.
+ * Mint new debt, and transfer a `fee` part of that to the vaultFactory's reward
+ * pool. Then reallocate over all the seat arguments and the rewardPoolSeat. Update
+ * the `totalDebt` if the reallocate succeeds.
  *
+ * TODO check limits.
+ *
+ * @param {Amount} toMint
  * @param {Amount} fee
- * @param {Amount} wanted
  * @param {ZCFSeat} fromSeat
  * @param {...ZCFSeat} otherSeats
+ * @returns {void}
+ */
+
+/**
+ * @callback BurnDebt
+ *
+ * Burn debt tokens off a seat and update
+ * the `totalDebt` if the reallocate succeeds.
+ *
+ * @param {Amount} toBurn
+ * @param {ZCFSeat} fromSeat
  * @returns {void}
  */
 
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -52,9 +52,10 @@
  * pool, taken from the `fromSeat`. Then reallocate over all the seat
  * arguments and the rewardPoolSeat.
  *
- * @param {Amount} amount
+ * @param {Amount} fee
+ * @param {Amount} wanted
  * @param {ZCFSeat} fromSeat
- * @param {ZCFSeat=} otherSeat
+ * @param {...ZCFSeat} otherSeats
  * @returns {void}
  */
 
@@ -78,6 +79,7 @@
  * @typedef {Object} VaultManagerBase
  * @property {(seat: ZCFSeat) => Promise<VaultKit>}  makeVaultKit
  * @property {() => void} liquidateAll
+ * @property {() => CollateralManager} getPublicFacet
  */
 
 /**
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -3,6 +3,8 @@
 /** @typedef {import('./vault').VaultUIState} VaultUIState */
 /** @typedef {import('./vaultKit').VaultKit} VaultKit */
 /** @typedef {VaultKit['vault']} Vault */
+/** @typedef {import('./vaultManager').VaultManager} VaultManager */
+/** @typedef {import('./vaultManager').CollateralManager} CollateralManager */
 
 /**
  * @typedef  {Object} AutoswapLocal
@@ -86,17 +88,6 @@
 
 /**
  * @typedef {string} VaultId
- */
-
-/**
- * @typedef {Object} VaultManagerBase
- * @property {(seat: ZCFSeat) => Promise<VaultKit>}  makeVaultKit
- * @property {() => void} liquidateAll
- * @property {() => CollateralManager} getPublicFacet
- */
-
-/**
- * @typedef {VaultManagerBase & GetVaultParams} VaultManager
  */
 
 /**
--- a/packages/run-protocol/src/vaultFactory/types.js
+++ b/packages/run-protocol/src/vaultFactory/types.js
@@ -48,7 +48,7 @@
 /**
  * @callback MintAndReallocate
  *
- * Mint new debt, and transfer a `fee` part of that to the vaultFactory's reward
+ * Mint new debt `toMint` and transfer the `fee` portion to the vaultFactory's reward
  * pool. Then reallocate over all the seat arguments and the rewardPoolSeat. Update
  * the `totalDebt` if the reallocate succeeds.
  *
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -343,14 +343,14 @@ const constructFromState = state => {
       // you're paying off the debt, you get everything back.
       const debt = getCurrentDebt();
       const {
-        give: { RUN: runOffered },
+        give: { RUN: given },
       } = seat.getProposal();
 
       // you must pay off the entire remainder but if you offer too much, we won't
       // take more than you owe
       assert(
-        AmountMath.isGTE(runOffered, debt),
-        X`Offer ${runOffered} is not sufficient to pay off debt ${debt}`,
+        AmountMath.isGTE(given, debt),
+        X`Offer ${given} is not sufficient to pay off debt ${debt}`,
       );
 
       // Return any overpayment
@@ -473,12 +473,12 @@ const constructFromState = state => {
       // unchanging) or super-linear (also called "convex"). Super-linear is from
       // AMMs: selling less collateral would mean an even smaller price impact, so
       // this is a conservative choice.
-      const runPerCollateral = makeRatioFromAmounts(
+      const debtPerCollateral = makeRatioFromAmounts(
         maxDebtPre,
         newCollateralPre,
       );
       // `floorMultiply` because the debt ceiling should be tight
-      const maxDebtAfter = floorMultiplyBy(newCollateral, runPerCollateral);
+      const maxDebtAfter = floorMultiplyBy(newCollateral, debtPerCollateral);
       assert(
         AmountMath.isGTE(maxDebtAfter, newDebt),
         X`The requested debt ${q(
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -446,7 +446,7 @@ const constructFromState = state => {
    * @param {Amount} fromGains
    * @param {Keyword} key
    */
-  const transfer = (fromSeat, toSeat, fromLoses, fromGains, key) => {
+  const stageDelta = (fromSeat, toSeat, fromLoses, fromGains, key) => {
     // Must check `isEmpty`; can't subtract `empty` from a missing allocation.
     if (!AmountMath.isEmpty(fromLoses)) {
       toSeat.incrementBy(fromSeat.decrementBy(harden({ [key]: fromLoses })));
@@ -468,7 +468,7 @@ const constructFromState = state => {
    * @param {Amount} loss
    * @returns {Amount}
    */
-  const applyDelta = (base, gain, loss) =>
+  const addSubtract = (base, gain, loss) =>
     AmountMath.subtract(AmountMath.add(base, gain), loss);
 
   /**
@@ -485,7 +485,7 @@ const constructFromState = state => {
   const loanFee = (currentDebt, giveAmount, wantAmount) => {
     const fee = ceilMultiplyBy(wantAmount, manager.getLoanFee());
     const toMint = AmountMath.add(wantAmount, fee);
-    const newDebt = applyDelta(currentDebt, toMint, giveAmount);
+    const newDebt = addSubtract(currentDebt, toMint, giveAmount);
     return { newDebt, toMint, fee };
   };
 
@@ -546,7 +546,7 @@ const constructFromState = state => {
     const giveColl = proposal.give.Collateral || emptyCollateral;
     const wantColl = proposal.want.Collateral || emptyCollateral;
 
-    const newCollateralPre = applyDelta(collateralPre, giveColl, wantColl);
+    const newCollateralPre = addSubtract(collateralPre, giveColl, wantColl);
     // max debt supported by current Collateral as modified by proposal
     const maxDebtPre = await maxDebtFor(newCollateralPre);
     assert(
@@ -559,7 +559,7 @@ const constructFromState = state => {
     // Get new balances after calling the priceAuthority, so we can compare
     // to the debt limit based on the new values.
     const collateral = getCollateralAllocated(vaultSeat);
-    const newCollateral = applyDelta(collateral, giveColl, wantColl);
+    const newCollateral = addSubtract(collateral, giveColl, wantColl);
 
     const debt = getCurrentDebt();
     const giveRUN = AmountMath.min(proposal.give.RUN || emptyDebt, debt);
@@ -585,11 +585,11 @@ const constructFromState = state => {
     // mint to vaultSeat, then reallocate to reward and client, then burn from
     // vaultSeat. Would using a separate seat clarify the accounting?
     // TODO what if there isn't anything to mint?
-    mint.mintGains(harden({ RUN: toMint }), vaultSeat);
 
-    transfer(clientSeat, vaultSeat, giveColl, wantColl, 'Collateral');
-    transfer(clientSeat, vaultSeat, giveRUN, wantRUN, 'RUN');
-    manager.reallocateWithFee(fee, vaultSeat, clientSeat);
+    stageDelta(clientSeat, vaultSeat, giveColl, wantColl, 'Collateral');
+    // `wantRUN is allocated in the reallocate and mint operation, and so not here
+    stageDelta(clientSeat, vaultSeat, giveRUN, emptyDebt, 'RUN');
+    manager.reallocateWithFee(fee, wantRUN, clientSeat, vaultSeat);
 
     // parent needs to know about the change in debt
     updateDebtAccounting(debtPre, collateralPre, newDebt);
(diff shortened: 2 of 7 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -307,7 +274,8 @@ export const makeInnerVault = (
       // TODO move manager state to a separate notifer https://github.com/Agoric/agoric-sdk/issues/4540
       interestRate: manager.getInterestRate(),
       liquidationRatio: manager.getLiquidationMargin(),
-      // XXX 'run' is implied by the brand in the amount
+      // XXX 'run' is implied by the brand in the amount\
+      // TODO rename the snapshot value to `debt`
```
turadg: this is safe to do in this PR as nothing downstream is relying on `debtSnapshot` yet

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -112,24 +108,12 @@ const validTransitions = {
...
  */
-export const makeInnerVault = (
-  zcf,
-  manager,
-  assetNotifier,
-  idInManager,
-  mint,
-  priceAuthority,
-) => {
+export const makeInnerVault = (zcf, manager, idInManager) => {
```
turadg: Excellent. POLA and will also reduce redundant storage in virtualization.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -364,14 +332,13 @@ export const makeInnerVault = (
     assertCloseable();
     const { phase, vaultSeat } = state;
     if (phase === VaultPhase.ACTIVE) {
+      // TODO it seems to me that `Collateral` shouldn't be required here.
```
turadg: I think that's right. When closing the vault it's likely that the agent seeks collateral back but it's not strictly necessary. We should allow them to close without specifying and simply get whatever collateral is available. Of course they can still specify Collateral if they choose.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -543,20 +507,19 @@ export const makeInnerVault = (
...
     assert(
       updaterPre === state.outerUpdater,
       X`Transfer during vault adjustment`,
     );
     assertActive();
 
-    // After the AWAIT, we retrieve the vault's allocations again.
-    // Get new balances after calling the priceAuthority, so we can compare
-    // to the debt limit based on the new values.
+    // After the AWAIT, we retrieve the vault's allocations again,
```
turadg: Why is `AWAIT` all caps? Would this suffice?
```suggestion
    // After the `await`, we retrieve the vault's allocations again,
```

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -75,27 +76,40 @@ export const start = async (zcf, privateArgs) => {
 
   assertElectorateMatches(electorateParamManager, governedTerms);
 
+  /** For temporary staging of newly minted tokens */
+  const { zcfSeat: stage } = zcf.makeEmptySeatKit();
```
turadg: Why is this is here instead of within `mintAndReallocate` ?

Also "stage" is a confusing name for a ZCFSeat since it has a staging concern within it. Consider `unallocated` which will be unambiguous within `mintAndReallocate`.

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -75,27 +76,40 @@ export const start = async (zcf, privateArgs) => {
...
+      zcf.reallocate(rewardPoolSeat, stage, seat, ...otherSeats);
+    } catch (e) {
+      stage.clear();
+      rewardPoolSeat.clear();
+      runMint.burnLosses(harden({ RUN: toMint }), stage);
+      throw e;
+    } finally {
+      assert(
+        AmountMath.isEmpty(stage.getAmountAllocated('RUN', runBrand)),
+        X`Stage should be empty of RUN`,
```
turadg: since it exists to hold newly minted RUN, shouldn't it be empty entirely at this point?

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -75,27 +76,40 @@ export const start = async (zcf, privateArgs) => {
...
+      );
     }
+    // TODO add aggregate debt tracking
```
turadg: that's this ticket? https://github.com/Agoric/agoric-sdk/issues/4482

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -199,9 +214,20 @@ export const start = async (zcf, privateArgs) => {
...
+    );
+    /** @type {VaultManager} */
+    return collateralTypes.get(brandIn).getPublicFacet();
+  };
+
   const publicFacet = Far('vaultFactory public facet', {
+    getCollateralManager,
     /** @deprecated use makeVaultInvitation instead */
     makeLoanInvitation: makeVaultInvitation,
+    /** @deprecated use getCollateralManager and then makeVaultInvitation instead */
```
turadg, line 238: Why is this better than taking a collateral type on `makeVaultInvitation`?
```
  // TODO add a `collateralBrand` argument to makeVaultInvitation`
  /** Make a loan in the vaultManager based on the collateral type. */
  const makeVaultInvitation = () => {
    return zcf.makeInvitation(makeVaultHook, 'MakeVault');
  };
```

If that is deprecated then it should have `@deprecated` (which I think will flow here but no harm in repeating)

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -342,12 +337,27 @@ export const makeVaultManager = (
 
   observeNotifier(periodNotifier, timeObserver);
 
+  const mintAndReallocate = (toMint, fee, seat, ...otherSeats) => {
+    mintAndReallocateWithFee(toMint, fee, seat, ...otherSeats);
+    totalDebt = AmountMath.add(totalDebt, toMint);
+    // TODO signal updater?
```
turadg: Good question. We do "track" total debt but I suppose there are two levels of tracking: available or notified. I don't know of use cases to be notified of increases in total debt of a vault manager. What are some?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -389,10 +392,19 @@ export const makeVaultManager = (
...
   /** @type {VaultManager} */
   return Far('vault manager', {
     ...shared,
     makeVaultKit,
     liquidateAll,
+    getPublicFacet: () => publicFacet,
   });
 };
+
+/** @typedef {Awaited<ReturnType<typeof makeVaultManager>>['getPublicFacet']} CollateralManager */
```
turadg: do we need this new name?

turadg (commented):
Great improvement. I can take on fixing the types if you like.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: We have named a thing VaultManager which is worth defining,
```
/** @typedef {Awaited<ReturnType<typeof makeVaultManager>>} VaultManager */
```
Once we have that, it's not much to say `VaultManager['getPublicFacet']` where you were going to use `CollateralManager`.

Note that because it's in this file it won't be available ambiently. I think that's a good thing and we should move in that direction, but we haven't decided on that as an org. If you want it to be ambient, you can import it into `types.js` with `/** @typedef {import('./vaultManager').VaultManager} VaultManager */`. I would suggest instead using that import whenever you need the type.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: It should be used by the UI?

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: We reuse the same seat across multiple different requests, rather than allocating a new one each time. It's for staging mint/burn.  `mintHelper`? `tmpSeat`?

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: Yes probably. I'll need to figure out a straightforward way to ask that.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: yes

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: I think it's better because we can then add `getNotifier` and other operations on a per-collateral basis. It also enables collateral-specific extensions (if e.g., for stable token collateral, liquidity token collateral, etc.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: The CollateralManager is the public face, so I would think it should have a name and work backwards from there to VaultManager, which is internal.

packages/run-protocol/src/vaultFactory/vault.js
turadg: It isn't yet. We had a ticket to support dynamic interest calculation but I can't find it now; I think it got renamed.

 https://github.com/Agoric/agoric-sdk/issues/4882 may address.

packages/run-protocol/src/vaultFactory/vaultFactory.js
turadg: Re-use for resource efficiency? (Should be fully empty between requests?)

"seat" > "helper" so maybe `tmpSeat`. Could be more specific though, `mintSeat`.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Ok so the intent is to rename VaultManager to CollateralManager? That sounds good to me and I request that either take place in this PR or be punted entirely to a different PR.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: Its because we used to write a `// AWAIT` comment before we came up with our current rule.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: I try to avoid bad hungarian notation (tmpSeat is largely like indexBigInt).  For some reason, that seems better for Seat than other usages (e.g. always prefer price to priceRatio, debt to debtAmount).

turadg: Ok, but isn't `indexHelper` even worse? :)  If there are multiple indexes in the same scope, you need to qualify them. That's why `tmp` isn't sufficient.

Another reason for superfluous type is to prevent collision. If you dropped the type from `mintSeat` it would be `mint` which is in this code often means the "the mint that mints" or the verb "to mint". 

My concerns with "stage" are a) ZCFSeat has its own staging, b) in code "stage" reads first as a verb,  c) reading the value somewhere without the comment it's not clear what it's to be used for.

Thinking through the above and re-reading the code, the best concise description I have of what this object does is `gainsHolder`.

Regardless, these concerns won't block my approval when this PR is ready for full review.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: Hmm. There's multiple facets:
- the private facet innerVaults use
- the private facet the vaultFactory uses -- this should go away once the deprecated invitations go away
- the public facet that is exposed to clients via `getCollateralManager`

I should note that I don't particularly like the name CollateralManager :) But it doesn't make sense to rename the private thing that will go away.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: This would push the updated total out to the external DB for apps to query. Not urgent

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: I also removed the TODO since we instead have them ask for the collateral manager.

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -46,15 +46,29 @@
  */
 
 /**
- * @callback ReallocateWithFee
+ * @callback MintAndReallocate
  *
- * Transfer the indicated amount to the vaultFactory's reward
- * pool, taken from the `fromSeat`. Then reallocate over all the seat
- * arguments and the rewardPoolSeat.
+ * Mint new debt, and transfer a `fee` part of that to the vaultFactory's reward
```
turadg:
```suggestion
 * Mint new debt `toMint` and transfer the `fee` portion to the vaultFactory's reward
```

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -46,15 +46,29 @@
...
+ * @param {ZCFSeat} fromSeat
+ * @param {...ZCFSeat} otherSeats
+ * @returns {void}
+ */
+
+/**
+ * @callback BurnDebt
+ *
+ * Burn debt tokens off a seat and update
+ * the `totalDebt` if the reallocate succeeds.
```
turadg, line 70: weird wrapping:
```suggestion
 * Burn debt tokens off a seat and update the `totalDebt` if the reallocate
 * succeeds.
```
We needn't spend time on whitespace in code review so I filed https://github.com/Agoric/agoric-sdk/issues/4900

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -366,12 +332,9 @@ export const makeInnerVault = (
     if (phase === VaultPhase.ACTIVE) {
       assertProposalShape(seat, {
         give: { RUN: null },
-        want: { Collateral: null },
       });
 
-      // you're paying off the debt, you get everything back. If you were
-      // underwater, we should have liquidated some collateral earlier: we
-      // missed our chance.
```
turadg, line 377: I've found this clarification helpful before. principle for removal?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -465,7 +425,7 @@ export const makeInnerVault = (
    * @param {Amount} loss
    * @returns {Amount}
    */
-  const applyDelta = (base, gain, loss) =>
+  const addSubtract = (base, gain, loss) =>
```
turadg, line 434: seems generally useful on `AmountMath` but I don't think we should put it there until there are more instances

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -75,27 +76,42 @@ export const start = async (zcf, privateArgs) => {
...
-          RUN: fee,
-        }),
-      ),
-    );
-    if (otherSeat !== undefined) {
-      zcf.reallocate(rewardPoolSeat, fromSeat, otherSeat);
-    } else {
-      zcf.reallocate(rewardPoolSeat, fromSeat);
+  const mintAndReallocate = (toMint, fee, seat, ...otherSeats) => {
+    const wanted = AmountMath.subtract(toMint, fee);
```
turadg: "want" is a proposal concept which is not the concern of this function. consider `proceeds`

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -75,27 +76,42 @@ export const start = async (zcf, privateArgs) => {
...
+    runMint.mintGains(harden({ RUN: toMint }), mintSeat);
+    try {
+      rewardPoolSeat.incrementBy(mintSeat.decrementBy(harden({ RUN: fee })));
+      seat.incrementBy(mintSeat.decrementBy(harden({ RUN: wanted })));
+      zcf.reallocate(rewardPoolSeat, mintSeat, seat, ...otherSeats);
+    } catch (e) {
+      mintSeat.clear();
+      rewardPoolSeat.clear();
+      runMint.burnLosses(harden({ RUN: toMint }), mintSeat);
```
turadg: what happens if `burnLosses` fails here? please document why this is safe.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: should have a ticket link for resolving the TODO (either by doing it or deciding it need not be done)

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -389,10 +392,19 @@ export const makeVaultManager = (
     }
   };
 
+  const publicFacet = Far('collateral manager', {
+    makeVaultInvitation: () => zcf.makeInvitation(makeVaultKit, 'MakeVault'),
+    getNotifier: () => assetNotifer,
```
turadg, line 395: wonder if we can collapse "collateral" and "asset" terms into one. "assetNotifier" was chosen because it's about the collateral and also the debt. Which implies "collateral manager" is doing more than managing collateral per se.

turadg (commented):
Noted some concerns for approval but nothing that would block another approval.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: It presumes that we should have liquidated you when we had the chance, which is not the point of liquidation. In this case, if the user can pay off the debt, the vault is solvent and the contract should be happy.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: In all cases, we are minting because an offer wanted it, so aligns well when reading its usage. But I shall consider your alternative word.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: Not well, since these are all assets, but what distinguishes here is that it is specific to a collateral type.

turadg (approved):

PR #4880, opened 2022-03-21
chore: fix mintPayment type

turadg:
## Description

The removal of `MintPayment` type wasn't complete. (h/t @mhofman https://github.com/Agoric/agoric-sdk/pull/4736/files#r830781495) I wonder what it would take for the type system to have noted this problem. I suppose tsc is permissive with type annotations in JSdoc because they could be for another type specification system. 

I didn't restore the callback because it's not a callback/interface. It's a function that exists only on `Mint` objects so I corrected the type inline. 

(While iterating on approaches I defined the `PaymentLedger` type and it's left in here)

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -253,7 +253,7 @@ export const makePaymentLedger = (
    * Creates a new Payment containing newly minted amount.
    *
    * @param {Amount<K>} newAmount
-   * @returns {Payment}
+   * @returns {Payment<K>}
    */
   const mintPayment = newAmount => {
     newAmount = coerce(newAmount);
@@ -365,3 +365,5 @@ export const makePaymentLedger = (
     mint,
   });
 };
+
+/** @typedef {ReturnType<makePaymentLedger>} PaymentLedger */
--- a/packages/ERTP/src/types.js
+++ b/packages/ERTP/src/types.js
@@ -292,7 +292,7 @@
  * assets all have the same kind, which is called a Brand.
  *
  * @property {() => Issuer<K>} getIssuer Gets the Issuer for this mint.
- * @property {MintPayment<K>} mintPayment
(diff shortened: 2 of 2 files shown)


mhofman:
> I wonder what it would take for the type system to have noted this problem.

`noImplicitAny` is what it would take

packages/ERTP/src/paymentLedger.js
```diff
@@ -365,3 +365,5 @@ export const makePaymentLedger = (
     mint,
   });
 };
+
+/** @typedef {ReturnType<makePaymentLedger>} PaymentLedger */
```
mhofman, line 369: Drive-by changes?

mhofman (approved):
LGTM, but I can't find any about the paymentLedger changes in the description

packages/ERTP/src/paymentLedger.js
turadg: yes. I'll add to description

turadg:
>> I wonder what it would take for the type system to have noted this problem.

> `noImplicitAny` is what it would take

I'm not sure of that. Here's a [playground with all typechecking options disabled](https://www.typescriptlang.org/play?strict=false&noImplicitAny=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&target=99&jsx=0&filetype=js#code/PQKhCgAIUgBAXAngBwKYBNUDNIG8D2ARgFaoDG8AvpAGL75QyzIBO+aLSeAKiqgHL54AEWwBLAHYZqhAIYtGwcOFBMkaPHXzUQSsvgkBneJCz1IAXjxyWALkgBGSgG5wQA) and it still can report that a type is being used that is not defined.
Screen Shot 2022-03-21 at 9 13 36 AM

So something is different about our setup that `MintPayment` didn't error and it's not `noImplicitAny`.

Issue #4882, opened 2022-03-21
UiNotifier in Vaults Contract Incompatible with Dapp


samsiegart:
## Describe the bug

The create-vault offer result in the vaults contract was changed to return `{ vaultNotifier, assetNotifier, ...}` instead of `{ uiNotifier, ... }`. The wallet depends on the key `uiNotifier` specifically for dapps to be able to access it from an offer result, while the rest of the offer result should not be accessible by the dapp because it has too many powers.

## To Reproduce
1. Follow the steps to set up `dapp-treasury`
2. Open a vault with any collateral
3. Navigate to the "Vaults" view
4. See error about UiNotifier

## Expected behavior
The dapp is able to access the UI notifiers to see the vault and asset data.

samsiegart:
Proposed solution:

- Modify the contract to return an offer result of `{ uiNotifiers: { vaultNotifier, assetNotifier }, ...}`
- Add method to the wallet bridge, `getUINotifiers(offerId)` which works similarly to `getUINotifier(offerId)`
- Use `getUINotifiers` in the vaults dapp to retrieve the correct data

Other considerations:

Should we deprecate `getUINotifier` and migrate all other contracts and dapps to use `getUINotifiers` instead?

turadg:
I thought this was covered in another issue but I can't find it and I suspect it was rescoped. Agree this is a "bug" right now.

Having the offer result return a set of notifiers instead of one is good and necessary. While we're making that change, I think we should remove the "ui" as there's no such constraint on how they'll be used. I.e. `{ notifiers: { vaultNotifier, assetNotifier }}`. Or potentially `{ notifiers: { vault, asset } } `.

No opinion on the bridge aspects.

PR #4886, opened 2022-03-21
fix(run-protocol): restructure vault notifiers to work with wallet

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/4882

Named it "publicNotifiers" after @michaelfig 's suggestion in https://github.com/Agoric/agoric-sdk/pull/4889


turadg (approved):
LGTM

I wonder if wallet can make use of the `VaultKit` type to get static checking that this object shape matches up.

Chris-Hibbert (approved):

PR #4888, opened 2022-03-21
fix(run-protocol): addVaultType with invalid args modified state

### LINKED ISSUE #4284: Method `home.vaultFactoryCreator.addVaultType` modifies state of the service even when being called with invalid arguments
## Describe the bug

The method `home.vaultFactoryCreator.addVaultType` modifies state of the service even when being called with invalid arguments.  

## To Reproduce
1. in simulator
1. tap the faucet `history[0] = E(home.faucet).tapFaucet()`
2. Add vault type with already registered issuer, which will fail since the issuer is already added to the vault
```js
E(home.vaultFactoryCreator).addVaultType(history[0][0]["issuer"], "AAAAAAAAAAAAAAAAA", null)
```
2. attempt to add the vault type again, and observe that the keyword is saved and is no longer unique, despite the .addVaultType call above failing
```js
E(home.vaultFactoryCreator).addVaultType(history[0][0]["issuer"], "AAAAAAAAAAAAAAAAA", null)
```

## Expected behavior
With invalid arguments, the addVaultType method should not modify any service state (leak)

## Platform Environment
 macOS 12.1, node v16.13.1
 I only have 16gb of ram

ref #4264

dckc:
closes: #4284

## Description

Check args against a pattern before doing anything that might modify state.

### Security Considerations

fixes a leak

### Testing Considerations

includes a regression test

cc @arirubinstein

### DIFF
--- a/packages/run-protocol/src/contractSupport.js
+++ b/packages/run-protocol/src/contractSupport.js
@@ -1,7 +1,14 @@
 // @ts-check
 import { AmountMath } from '@agoric/ertp';
+import { M } from '@agoric/store';
 
 const { details: X } = assert;
+
+export const amountPattern = harden({ brand: M.remotable(), value: M.any() });
+export const ratioPattern = harden({
+  numerator: amountPattern,
+  denominator: amountPattern,
+});
 
 /**
  * Apply a delta to the `base` Amount, where the delta is represented as
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -15,6 +15,7 @@ import {
 import { makeScriptedPriceAuthority } from '@agoric/zoe/tools/scriptedPriceAuthority.js';
 import { makeManualPriceAuthority } from '@agoric/zoe/tools/manualPriceAuthority.js';
 import { assertAmountsEqual } from '@agoric/zoe/test/zoeTestHelpers.js';
+import { M, matches } from '@agoric/store';
 import { makeParamManagerBuilder } from '@agoric/governance';
 
 import { makeTracer } from '../../src/makeTracer.js';
@@ -2189,4 +2190,80 @@ test('mutable liquidity sensitivity of triggers and interest', async t => {
   // No change for Alice
   aliceUpdate = await E(aliceNotifier).getUpdateSince(); // can't use updateCount because there's no newer update
   t.is(aliceUpdate.value.vaultState, Phase.ACTIVE);
+});
+
+test('addVaultType: invalid args do not modify state', async t => {
+  const {
+    aethKit: { brand: aethBrand },
+  } = t.context;
+  const kw = 'Chit';
+  const chit = makeIssuerKit(kw);
+  const params = defaultParamValues(chit.brand);
+
+  const services = await setupServices(
+    t,
+    [500n, 15n],
+    AmountMath.make(aethBrand, 900n),
+    undefined,
+    undefined,
+    500n,
+  );
+
+  const { vaultFactory } = services.vaultFactory;
+
+  const failsForSameReason = async p =>
+    p
+      .then(oops => t.fail(`${oops}`))
+      .catch(reason1 => t.throwsAsync(p, { message: reason1.message }));
+  await failsForSameReason(
+    E(vaultFactory)
+      // @ts-ignore bad args on purpose for test
+      .addVaultType(chit.issuer, kw, null),
+  );
+  await failsForSameReason(
+    E(vaultFactory)
+      // @ts-ignore bad args on purpose for test
+      .addVaultType(chit.issuer, 'bogus kw', params),
+  );
+
+  // The keyword in the vault manager is not "stuck"; it's still available:
+  const actual = await E(vaultFactory).addVaultType(chit.issuer, kw, params);
+  t.true(matches(actual, M.remotable()));
+});
+
+test('addVaultType: extra, unexpected params', async t => {
+  const {
+    aethKit: { brand: aethBrand },
+  } = t.context;
+  const chit = makeIssuerKit('chit');
+
+  const services = await setupServices(
+    t,
+    [500n, 15n],
+    AmountMath.make(aethBrand, 900n),
+    undefined,
+    undefined,
+    500n,
+  );
+
+  const { vaultFactory } = services.vaultFactory;
+
+  const params = { ...defaultParamValues(aethBrand), shoeSize: 10 };
+  const extraParams = { ...params, shoeSize: 10 };
+  const { interestRate: _1, ...missingParams } = {
+    ...defaultParamValues(aethBrand),
+    shoeSize: 10,
+  };
+
+  await t.throwsAsync(
+    E(vaultFactory).addVaultType(chit.issuer, 'Chit', missingParams),
+    { message: /Must have same property names/ },
+  );
+
+  const actual = await E(vaultFactory).addVaultType(
+    chit.issuer,
+    'Chit',
+    extraParams,
+  );
+  t.true(matches(actual, M.remotable()), 'unexpected params are ignored');
 });
--- a/packages/run-protocol/src/vaultFactory/params.js
+++ b/packages/run-protocol/src/vaultFactory/params.js
@@ -8,6 +8,8 @@ import {
   makeParamManager,
   ParamTypes,
 } from '@agoric/governance';
+import { M } from '@agoric/store';
+import { amountPattern, ratioPattern } from '../contractSupport.js';
 
 export const CHARGING_PERIOD_KEY = 'ChargingPeriod';
 export const RECORDING_PERIOD_KEY = 'RecordingPeriod';
@@ -42,6 +44,17 @@ const makeVaultParamManager = initial =>
     [LOAN_FEE_KEY]: [ParamTypes.RATIO, initial.loanFee],
   });
 /** @typedef {ReturnType<typeof makeVaultParamManager>} VaultParamManager */
+
+export const vaultParamPattern = M.split(
+  {
+    liquidationMargin: ratioPattern,
+    liquidationPenalty: ratioPattern,
+    interestRate: ratioPattern,
+    loanFee: ratioPattern,
+    debtLimit: amountPattern,
+  },
+  M.any(),
+);
 
 /**
  * @param {ERef<ZoeService>} zoe
--- a/packages/run-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/run-protocol/src/vaultFactory/vaultDirector.js
@@ -6,7 +6,7 @@ import '@agoric/zoe/src/contracts/exported.js';
 import { E } from '@endo/eventual-send';
 import '@agoric/governance/src/exported.js';
 
-import { makeScalarMap } from '@agoric/store';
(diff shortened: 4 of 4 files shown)


Chris-Hibbert:
I think the test failures may be due to a github outage issue. I'm seeing similar failures in an unrelated PR.

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -108,6 +108,24 @@ export const start = async (zcf, privateArgs) => {
 
   /** @type {AddVaultType} */
   const addVaultType = async (collateralIssuer, collateralKeyword, rates) => {
+    const amountPattern = { brand: M.remotable(), value: M.any() };
+    const ratioPattern = {
+      numerator: amountPattern,
+      denominator: amountPattern,
+    };
+    fit(
```
Chris-Hibbert: Why does this help?  (`fit` is new for me.)  As I read the doc for `fit`, it looks like it returns a boolean.  What causes it to act as a guard here?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2296,3 +2296,52 @@ test('mutable liquidity triggers and interest sensitivity', async t => {
...
+  const kw = 'AAAAAAAAAAAAAAAAA';
+  await E(vaultFactory)
+    // @ts-ignore bad args on purpose for test
+    .addVaultType(aethIssuer, kw, null)
+    .then(oops => t.fail(`${oops}`))
+    .catch(reason1 =>
+      t.throwsAsync(
+        // @ts-ignore bad args on purpose for test
+        E(vaultFactory).addVaultType(aethIssuer, kw, null),
+        { message: reason1.message },
```
Chris-Hibbert: This shows that it returns the same error message on the second call, but it doesn't show that it could succeed on the second call if the other args were no longer erroneous.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dckc: The doc for `fit` might be a bit spread out. It's defined as:

```
  const fit = (specimen, patt) => {
    checkMatches(specimen, patt, assertChecker);
  };
```

... where [assertChecker](https://github.com/endojs/endo/blob/master/packages/marshal/src/helpers/passStyle-helpers.js#L50) comes from `@endo/marshal`. The doc for its type, [Checker](https://github.com/endojs/endo/blob/master/packages/marshal/src/types.js#L228-L240) should help:

```
 * @callback Checker
 * Internal to a useful pattern for writing checking logic
 * (a "checkFoo" function) that can be used to implement a predicate
 * (an "isFoo" function) or a validator (an "assertFoo" function).
 *
 *    * A predicate ideally only returns `true` or `false` and rarely throws.
 *    * A validator throws an informative diagnostic when the predicate
 *      would have returned `false`, ...
```

cc @erights

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dckc: Good thing you prompted me to finish the test; it showed that the call to `fit()` didn't work; it was missing a `harden()`.

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -129,6 +129,24 @@ export const start = async (zcf, privateArgs) => {
 
   /** @type {AddVaultType} */
   const addVaultType = async (collateralIssuer, collateralKeyword, rates) => {
+    const amountPattern = { brand: M.remotable(), value: M.any() };
+    const ratioPattern = {
+      numerator: amountPattern,
+      denominator: amountPattern,
+    };
```
Chris-Hibbert: Should these be in some standard place?  Seems like we'll use them often.

Chris-Hibbert (approved):
Looks good.

I recommend finding a common place for soon-to-be common code.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dckc: how about `packages/run-protocol/src/contractSupport.js`?

Chris-Hibbert: That seems right.

packages/run-protocol/src/contractSupport.js
```diff
@@ -1,8 +1,15 @@
 // @ts-check
 import { AmountMath } from '@agoric/ertp';
+import { M } from '@agoric/store';
 
 const { details: X } = assert;
 
+export const amountPattern = { brand: M.remotable(), value: M.any() };
+export const ratioPattern = {
```
Chris-Hibbert: They'll be part of a hardened structure, but it would still be a mistake if anyone modified them before using them. Wouldn't it be helpful to `harden()` them?

Chris-Hibbert (approved):

packages/run-protocol/src/contractSupport.js
dckc: eek! yes, of course; lack of `harden()` is a pretty bad hazzard here.

turadg: I had a comment here but I think it got lost. (Update: it was a review comment and I hadn't submitted the review.)

+1 to hardening here. But why do we need to harden the arguments to `fit`. That function must already be trusted.

Chris-Hibbert: I dunno.  `fit` is not only trusted, it's completely local, right?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -34,6 +34,7 @@ import { startVaultFactory } from '../../src/econ-behaviors.js';
 import '../../src/vaultFactory/types.js';
 import * as Collect from '../../src/collect.js';
 import { calculateCurrentDebt } from '../../src/interest-math.js';
+import { M, matches } from '@agoric/store';
```
turadg: @dckc Heads up on this lint error.

We really should automate import sorting.

packages/run-protocol/src/contractSupport.js
dckc: `harden()` is designed to be used on all newly created objects before they escape the static scope in which they are created:

> `harden()` should be called on all objects that will be transferred across a trust boundary. It's important to `harden()` an object before exposing the object by returning it or passing it to some other function.
 -- https://agoric.com/documentation/guides/js-programming/hardened-js.html#defensive-objects-with-harden

> To enable sound static reasoning, in Jessie all objects made by literal expressions (object literals, array literals, the many forms of function literals) must be tamper-proofed with harden before they can be aliased or escape from their static context of origin.
 -- https://github.com/endojs/Jessie

This supports defensive correctness at a granularity finer than vats -- all the way down to individual objects.

In the interest of programmer convenience, in some of our APIs, the callee hardens for the caller. I think @erights is uneasy in all such cases, since (1) it risks masking bugs / confusion (as discussed in https://agoric.com/documentation/guides/js-programming/far.html#pass-styles-and-harden ) , and (2) some callers might be surprised by having their objects hardened as a consequence of passing them to functions such as `fit()`.

dckc:
oops; I squashed locally. I should have let github do that...

dckc:
@Chris-Hibbert This use of pattern matching refuses extra fields. For example, `demoIssuers.js` was providing `initialPrice` and `initialMargin` along with `interestRate` and such. That was throwing. I just changed `demoIssuers.js` to not do that, but... is that the way we want to go?

mhofman:
> @Chris-Hibbert This use of pattern matching refuses extra fields. For example, `demoIssuers.js` was providing `initialPrice` and `initialMargin` along with `interestRate` and such. That was throwing. I just changed `demoIssuers.js` to not do that, but... is that the way we want to go?

I'm triggering the loadgen CI on this as I believe it will break.

In general I'm against unrecognized props in option bags not being simply ignored.

mhofman (dismissed):
Requesting change until I figure out the result of the loadgen test

mhofman:
> I'm triggering the loadgen CI on this as I believe it will break.
> 
> In general I'm against unrecognized props in option bags not being simply ignored.

Test ran without errors, but I'm still concerned extra properties would cause errors, and that this will bite us in the future

Chris-Hibbert:
> In general I'm against unrecognized props in option bags not being simply ignored.

@mhofman 
Sorry too many negatives collided in my head there. Can you say something simple about what you favor?

Chris-Hibbert (approved):
Thanks for the changes. Once CI and @mhofman are satisfied, this can be merged.

LGTM.

mhofman:
> Sorry too many negatives collided in my head there. Can you say something simple about what you favor?

I disagree with having a runtime error if an options bag has extra unexpected properties, which this PR introduces. While it doesn't break the loadgen today, if any changes like https://github.com/Agoric/testnet-load-generator/pull/72 and https://github.com/Agoric/testnet-load-generator/pull/75 are needed in the future, it will cause the loadgen to lose backwards compatibility.

mhofman (dismissed):
Given this PR is on automerge, I want to make sure the extra property question is resolved before this lands.

dckc:
TIL: `M.split()`

@mhofman I added tests for both missing and extra params.

@Chris-Hibbert I hoisted `vaultParamPattern` to `params.js` since it took so much work to get right... maybe a client would want to use it?
And rather than `fit(harden([...args]))` I made separate calls to `fit()` for each arg... except `collateralKeyword`, where I used `assertKeywordName(collateralKeyword)` instead. And I added a test for bad keyword.

Chris-Hibbert:
> TIL: M.split()

Did you find documentation for it somewhere? I perused `.../packages/store` and didn't come up with anything.

Chris-Hibbert:
> I hoisted vaultParamPattern to params.js since it took so much work to get right... maybe a client would want to use it?

Seems plausible

> And rather than fit(harden([...args])) I made separate calls to fit() for each arg... except collateralKeyword, where I used 
assertKeywordName(collateralKeyword) instead. And I added a test for bad keyword.

That works.

Chris-Hibbert (approved):

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2190,3 +2191,79 @@ test('mutable liquidity sensitivity of triggers and interest', async t => {
...
+    AmountMath.make(aethBrand, 900n),
+    undefined,
+    undefined,
+    500n,
+  );
+
+  const { vaultFactory } = services.vaultFactory;
+
+  const params = { ...defaultParamValues(aethBrand), shoeSize: 10 };
+  const extraParams = { ...params, shoeSize: 10 };
```
mhofman, line 2252: Thanks!

mhofman (approved):
Approving for the handling of extra params

dckc:
> > TIL: M.split()
> 
> Did you find documentation for it somewhere? I perused `.../packages/store` and didn't come up with anything.

You too? Now I feel less dumb. :)

With help from @erights , I found the following; he and @mhofman and I agreed that it could be made easier to find:

https://github.com/Agoric/agoric-sdk/blob/4a46d9e37fc3946edb99e8ce41f52872e7816094/packages/store/src/types.js#L530-L536

PR #4889, opened 2022-03-22
feat(wallet): support reading notifiers from offer result

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/4882

We may want to deprecate `getUINotifier` at some point, but for now just add `getNotifiers`.

This will be used alongside https://github.com/Agoric/agoric-sdk/pull/4886

cc @turadg


packages/wallet/api/src/lib-wallet.js
```diff
@@ -1528,6 +1528,23 @@ export function makeWallet({
     return offerResult.uiNotifier;
   }
 
+  /**
+   * Gets the notifiers from an offer's result.
+   *
+   * @param {string} rawId - The offer's raw id.
+   * @param {string} dappOrigin - The origin of the dapp the offer came from.
+   */
+  async function getNotifiers(rawId, dappOrigin = 'unknown') {
```
michaelfig: Can you please name this `getPublicNotifiers`, and have it expose `offerResult.publicNotifiers`?  The reason is that we want to prevent the case where the contract doesn't intend to expose the notifiers to the (untrusted) dapp UI.

`getUINotifiers` and `offerResult.uiNotifiers` would also be an option, if we want to be consistent with the original decision for disambiguation.

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1528,6 +1528,23 @@ export function makeWallet({
...
+    const id = makeId(dappOrigin, rawId);
+    const offerResult = await idToOfferResultPromiseKit.get(id).promise;
+    assert(
+      passStyleOf(offerResult) === 'copyRecord',
+      `offerResult must be a record to have notifiers`,
+    );
+    assert(offerResult.notifiers, X`offerResult does not have notifiers`);
+    return offerResult.notifiers;
```
michaelfig: It's important to ensure the notifiers are a record as well.

As a general practice, please only ask for `offerResult.notifiers` once (via destructuring).  It's not as important here, but we want to encourage programmers to "avoid conflicting answers by only asking orthogonal questions".

```suggestion
    const { notifiers } = offerResult;
    assert(notifiers, X`offerResult does not have notifiers`);
    assert(
      passStyleOf(notifiers) === 'copyRecord',
      X`notifiers ${notifiers} must be a record`,
    );
    return notifiers;
```

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1528,6 +1528,23 @@ export function makeWallet({
...
+   *
+   * @param {string} rawId - The offer's raw id.
+   * @param {string} dappOrigin - The origin of the dapp the offer came from.
+   */
+  async function getNotifiers(rawId, dappOrigin = 'unknown') {
+    const id = makeId(dappOrigin, rawId);
+    const offerResult = await idToOfferResultPromiseKit.get(id).promise;
+    assert(
+      passStyleOf(offerResult) === 'copyRecord',
+      `offerResult must be a record to have notifiers`,
```
michaelfig:
```suggestion
      X`offerResult ${offerResult} must be a record to have notifiers`,
```

michaelfig (commented):
The general idea is good.

Please ping me again when you make a decision around the naming of the `getNotifiers` method and `.notifiers` property.

packages/wallet/api/src/lib-wallet.js
samsiegart: Changed to `getPublicNotifiers`/`offerResult.publicNotifiers`.

@turadg pointed out that `uiNotifiers` might not necessarily just be used by a UI, so I think `public` works best.

packages/wallet/api/src/lib-wallet.js
samsiegart: Done

packages/wallet/api/src/lib-wallet.js
samsiegart: Done

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1528,6 +1528,34 @@ export function makeWallet({
...
+   * @param {string} dappOrigin - The origin of the dapp the offer came from.
+   */
+  async function getPublicNotifiers(rawId, dappOrigin = 'unknown') {
+    const id = makeId(dappOrigin, rawId);
+
+    const offerResult = await idToOfferResultPromiseKit.get(id).promise;
+    assert(
+      passStyleOf(offerResult) === 'copyRecord',
+      `offerResult ${offerResult} must be a record to have publicNotifiers`,
+    );
```
turadg, line 1545: Is this exceptional? Reading this, I take it an offerResult can be other things if it doesn't have publicNotifiers and doesn't necessarily have publicNotifiers.

The asserts in this function treat it as an error to call `getPublicNotifiers` for an offer that doesn't have publicNotifiers. That would be surprising to me as a user of the function as I expect generally calling getThings when there are no things to just return a value indicating that: null, undefined or an empty collection. In this case, it may help to distinguish no key from a value that is set so I'd suggest returning `undefined` instead of these first two throws. (The third looks like a necessary type guard.)

If the throws behavior of this function is better for how dapps are programmed, I'd like to understand that better and also request that `@throws` be used in the jsdoc to make it explicit.

packages/wallet/api/test/continuingInvitationExample.js
```diff
@@ -22,6 +22,7 @@ export const start = zcf => {
     updater.updateState('first offer made');
     return harden({
       uiNotifier: notifier,
+      publicNotifiers: { notifier },
```
turadg: Part of the move away from `uiNotifier` is to name the notifiers for what they provide rather than the expected subscriber. We could demonstrate that more here with:
```suggestion
      publicNotifiers: { offers: notifier },
```

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1358,13 +1358,20 @@ test('addOffer makeContinuingInvitation', async t => {
...
   assert(accepted);
 
   const update = await E(uiNotifierP).getUpdateSince();
   t.is(update.value, 'first offer made');
 
+  // Test getNotifiers as well, even though it's redundant in this example.
+  const publicNotifiersUpdate = await E(
+    E.get(publicNotifiersP).notifier,
+  ).getUpdateSince();
+  t.is(publicNotifiersUpdate.value, 'first offer made');
```
turadg, line 1381: WDYT of testing that they're the same?
```suggestion
  t.deepEqual(publicNotifiersUpdate, update);
```

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1606,6 +1634,7 @@ export function makeWallet({
       return paymentsNotifier;
     },
     getUINotifier,
+    getPublicNotifiers,
```
turadg, line 1719: > We may want to deprecate getUINotifier at some point

Why not now? For reader of the code it helps explain why there are two methods.
```
    /** @deprecated use getPublicNotifiers instead */
    getUINotifier,
    getPublicNotifiers,
```

turadg (commented):
+1 to "public" as the term for these.

One concern about exceptions gating my approval but I'm reviewing with Comment in case I'm missing something and Michael wants to Approve.

packages/wallet/api/src/lib-wallet.js
samsiegart: I'm okay with that as long as @michaelfig agrees. I was on the fence of whether we'd also want to support the single notifier for simpler contracts, but it's probably better to just standardize.

packages/wallet/api/test/test-lib-wallet.js
samsiegart: I prefer this way because if/when we remove the other notifier, they're not as intertwined

packages/wallet/api/src/lib-wallet.js
samsiegart: I think for contracts, a successful offer result should either always have a notifier or never have one. If the existence of a notifier is being used as a signal for something in the dapp, there's probably better ways to infer whatever that signal is.

That assumption would boil this case down to just programming error--you either implemented your contract incorrectly or tried to access a notifier on the wrong type of offer result. In that case, I think seeing these errors from a capTP call in the browser is more helpful for debugging.

turadg: > I think for contracts, a successful offer result should either always have a notifier or never have one.

Agreed.

> That assumption would boil this case down to just programming error--you either implemented your contract incorrectly or tried to access a notifier on the wrong type of offer result.

It's that latter case I'm asking about. It sounds like you're saying that before you call `getPublicNotifiers` you should know whether the offer result has public notifiers. If that's so I think the jsdoc should indicate that somehow. E.g. `@throws if the offer result doesn't have notifiers`.

packages/wallet/api/src/lib-wallet.js
michaelfig: > I'm okay with that as long as @michaelfig agrees. I was on the fence of whether we'd also want to support the single notifier for simpler contracts, but it's probably better to just standardize.

I'm fine with deprecating, but not removing quite yet.

samsiegart: Added `@deprecated` doc comment

packages/wallet/api/src/lib-wallet.js
samsiegart: Added `@throws ...` to the jsdoc

packages/wallet/api/test/continuingInvitationExample.js
samsiegart: Done

turadg (approved):
👍

michaelfig (approved):
LGTM!

Issue #3595, opened 2021-08-04
Complete the end to end of ATOM/USD price feed


rowgraus:
Taken and edited from: https://github.com/Agoric/agoric-sdk/issues/3595#issuecomment-1072706254

We do not presume economy (vaults / AMM) but `startPriceAuthority` in bootstrap populates home.priceAuthority with a price authority registry.

User bob wants to know the price of ATOM in our system.

NodeCoordinator submits a proposal to BLDer DAO, who approves:
- instantiating aggregator contract with ATOM-USD oracle brands (`E(home.agoricNames).lookup('oracleBrand', 'ATOM')` and `E(home.agoricNames).lookup('oracleBrand', 'USD')`)
- adding aggregator priceAuthority to the price authority registry for ATOM-USD pair
- sending oracle invitations to NodeOp1's and 2's deposit facet

NodeOp1 registers an oracle node with the aggregator:
- NodeOp1 deploys oracle handler plugin
- NodeOp1 runs flux-notifier script to create fluxNotifier and use oracle invitation to register it with the aggregator

So bob can get a price:
- Bob submits a price query to home.priceAuthority
- NodeOp1's oracle plugin polls the price, reports it to the aggregator
- Likewise NodeOp2..N.
- Aggregator computes price and publishes via home.priceAuthority

dckc:
@michaelfig tells me this is pending availability of oracles.

p.s. The Pegasus contract has been imported into agoric-sdk.  `dapp-pegasus` is just a placeholder for when we have a Pegasus UI.

p.p.s. search keyword: IBC

Tartuffo:
Need to get this into good enough state to be handed over to Oracle providers.

dckc:
interviewed @michaelfig to get a story:

do not presume economy (vaults / AMM)
but `startPriceAuthority` in bootstrap populates home.priceAuthority

User bob wants to know the price of Atom in our system.

NodeCoordinator who installs aggregator contract, oracle contract. (proposes to BLDer DAO, who approves...)
Proposal includes adding price authority for Atom/USD from aggregator to the price authority registry.

NodeCoordinator uses creator facet to get invitations; hands out to NodeOp1 and 2 via deposit facet.

NodeOp1 deploys script that uses invitation to register themselves with the aggregator.

atom/USD aggregator is registered with home.priceAuthority
So bob can get a price:
NodeOp1 puts finger in wind, chooses 23 USD/Atom as a price, uses registered widget to submit price.
[...]

PR #4891, opened 2022-03-22
Updates for oracle bootstrap

michaelfig:
refs: #3595

## Description

Several changes needed for bootstrapping the oracle price feeds.  Aside from the price aggregator mechanics, here are some of the additional tweaks:

- more AmountKind type propagation
- scale the maximum mempool transaction size based on our changes to the maximum rpc message size
- move `startPriceAuthority` into pre-economy bootstrap
- make preapproved wallet bridge more useful (allow it to specify all the offer arguments)

### Testing Considerations

Being tested with third parties as part of the oracle bootstrap process.

### DIFF
--- a/packages/ERTP/src/types.js
+++ b/packages/ERTP/src/types.js
@@ -161,6 +161,7 @@
  */
 
 /**
+ * @template {AssetKind} [K=AssetKind]
  * @callback IssuerGetAmountOf
  *
  * Get the amount of digital assets in the payment. Because the
@@ -171,8 +172,7 @@
  * resolution.
  *
  * @param {ERef<Payment>} payment
- * @returns {Promise<Amount>}
- *
+ * @returns {Promise<Amount<K>>}
  */
 
 /**
@@ -244,7 +244,7 @@
  * @property {() => Purse} makeEmptyPurse Make an empty purse of this
  * brand.
  * @property {IssuerIsLive} isLive
- * @property {IssuerGetAmountOf} getAmountOf
+ * @property {IssuerGetAmountOf<K>} getAmountOf
  * @property {IssuerBurn} burn
  * @property {IssuerClaim} claim
  * @property {IssuerCombine} combine
--- a/packages/wallet/api/src/wallet.js
+++ b/packages/wallet/api/src/wallet.js
@@ -238,8 +238,9 @@ export function buildRootObject(vatPowers) {
    * @type {WalletBridge}
    */
   const preapprovedBridge = Far('preapprovedBridge', {
-    addOffer(offer) {
-      return walletAdmin.addOffer(offer);
+    /** @type {WalletAdminFacet['addOffer']} */
+    addOffer(offer, meta = undefined) {
+      return walletAdmin.addOffer(offer, meta);
     },
     getDepositFacetId(brandBoardId) {
       return walletAdmin.getDepositFacetId(brandBoardId);
--- a/packages/wallet/api/src/wallet.js
+++ b/packages/wallet/api/src/wallet.js
@@ -238,8 +238,11 @@ export function buildRootObject(vatPowers) {
    * @type {WalletBridge}
    */
   const preapprovedBridge = Far('preapprovedBridge', {
-    /** @type {WalletAdminFacet['addOffer']} */
-    addOffer(offer, meta = undefined) {
+    /**
+     * @param {unknown} offer
+     * @param {{}} [meta]
+     */
+    addOffer(offer, meta) {
       return walletAdmin.addOffer(offer, meta);
     },
     getDepositFacetId(brandBoardId) {
--- a/packages/zoe/src/contracts/priceAggregator.js
+++ b/packages/zoe/src/contracts/priceAggregator.js
@@ -24,6 +24,8 @@ const { add, multiply, floorDivide, ceilDivide, isGTE } = natSafeMath;
  * @param {ZCF<{
  * timer: TimerService,
  * POLL_INTERVAL: bigint,
+ * brandIn: Brand,
+ * brandOut: Brand,
  * unitAmountIn: Amount,
  * }>} zcf
  */
@@ -31,7 +33,8 @@ const start = async zcf => {
   const {
     timer,
     POLL_INTERVAL,
-    brands: { In: brandIn, Out: brandOut },
+    brandIn,
+    brandOut,
     unitAmountIn = AmountMath.make(brandIn, 1n),
   } = zcf.getTerms();
 
--- a/packages/zoe/src/contracts/priceAggregator.js
+++ b/packages/zoe/src/contracts/priceAggregator.js
@@ -84,12 +84,20 @@ const start = async zcf => {
     async wake(timestamp) {
       // Run all the queriers.
       const querierPs = [];
-      oracleRecords.forEach(({ querier }) => {
+      const samples = [];
+      oracleRecords.forEach(({ querier, lastSample }) => {
         if (querier) {
           querierPs.push(querier(timestamp));
         }
+        // Push result.
+        samples.push(lastSample);
       });
-      await Promise.all(querierPs);
+      if (!querierPs.length) {
+        // Only have push results, so publish them.
+        // eslint-disable-next-line no-use-before-define
+        querierPs.push(updateQuote(samples, timestamp));
+      }
+      await Promise.all(querierPs).catch(console.error);
     },
   });
   E(repeaterP).schedule(waker);
@@ -174,7 +182,6 @@ const start = async zcf => {
       { add, divide: floorDivide, isGTE },
     );
 
-    // console.error('found median', median, 'of', samples);
     if (median === undefined) {
       return;
     }
@@ -260,7 +267,7 @@ const start = async zcf => {
       assert(records.has(record), X`Oracle record is already deleted`);
 
       /** @type {OracleAdmin} */
-      const oracleAdmin = {
+      const oracleAdmin = Far('OracleAdmin', {
         async delete() {
           assert(records.has(record), X`Oracle record is already deleted`);
 
@@ -289,11 +296,11 @@ const start = async zcf => {
           // the median calculation.
           pushResult(result);
         },
-      };
+      });
 
       if (query === undefined) {
         // They don't want to be polled.
-        return harden(oracleAdmin);
+        return oracleAdmin;
       }
 
       let lastWakeTimestamp = 0n;
@@ -320,7 +327,7 @@ const start = async zcf => {
       await record.querier(now);
 
       // Return the oracle admin object.
-      return harden(oracleAdmin);
+      return oracleAdmin;
     },
   });
 
--- a/packages/zoe/src/contracts/priceAggregator.js
+++ b/packages/zoe/src/contracts/priceAggregator.js
@@ -335,6 +335,9 @@ const start = async zcf => {
     getPriceAuthority() {
       return priceAuthority;
     },
+    async getRoundStartNotifier() {
+      return undefined;
+    },
   });
 
   return harden({ creatorFacet, publicFacet });
--- a/packages/zoe/src/contracts/priceAggregator.js
+++ b/packages/zoe/src/contracts/priceAggregator.js
@@ -74,7 +74,7 @@ const start = async zcf => {
   /** @type {Set<OracleRecord>} */
   const oracleRecords = new Set();
 
-  /** @type {LegacyMap<Instance, Set<OracleRecord>>} */
+  /** @type {LegacyMap<OracleKey, Set<OracleRecord>>} */
   // Legacy because we're storing a raw JS Set
   const instanceToRecords = makeLegacyMap('oracleInstance');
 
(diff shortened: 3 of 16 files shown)


michaelfig:
@dckc, can you please review?  You were closer to many of these changes than the code owners.

@turadg, please review the AmountKind change and the contract updates.

packages/ERTP/src/types.js
```diff
@@ -161,6 +161,7 @@
  */
 
 /**
+ * @template {AssetKind} [K=AssetKind]
```
turadg, line 164: 👍

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -434,7 +434,7 @@ const start = async zcf => {
     const rrId = reportingRoundId;
 
     let canSupersede = true;
-    if (_roundId !== 1n) {
+    if (_roundId > 1n) {
```
dckc, line 447: Why change this?

Is it a huge hassle to add a unit test for whatever is going on here?

packages/wallet/api/src/wallet.js
```diff
@@ -238,8 +238,8 @@ export function buildRootObject(vatPowers) {
    * @type {WalletBridge}
    */
   const preapprovedBridge = Far('preapprovedBridge', {
-    addOffer(offer) {
-      return walletAdmin.addOffer(offer);
+    addOffer(offer, meta = undefined) {
```
turadg: jsdoc better for optional.

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -684,7 +686,7 @@ const start = async zcf => {
           pushResult({
             roundId: _roundIdRaw,
             data: _submissionRaw,
-          });
+          }).catch(console.error);
```
dckc, line 693: I'm not confident that I understand why this is OK. Add a comment, perhaps?

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -251,6 +253,98 @@ test('median aggregator', /** @param {ExecutionContext} t */ async t => {
...
+    issuers: { Quote: rawQuoteIssuer },
+    unitAmountIn = AmountMath.make(brandIn, 1n),
+  } = await E(zoe).getTerms(aggregator.instance);
+  /** @type {Issuer<'set'>} */
+  const quoteIssuer = rawQuoteIssuer;
+
+  const pricePush = await makeFakePriceOracle(t);
+  const pa = E(aggregator.publicFacet).getPriceAuthority();
+
+  // TODO: Port this to makeQuoteNotifier(amountIn, brandOut)
```
dckc: That TODO looks done. What am I missing?

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -251,6 +253,98 @@ test('median aggregator', /** @param {ExecutionContext} t */ async t => {
...
+  // @ts-ignore fix needed
```
dckc: consider `@ts-expect-error`

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -231,19 +241,75 @@ const start = async zcf => {
...
+          if (!oracleNotifier || !updateCount) {
+            // Interrupt the cycle because we either are deleted or the notifier
+            // finished.
+            return;
+          }
+          // Queue the next update.
+          E(oracleNotifier).getUpdateSince(updateCount).then(recurse);
+
+          // See if we have associated parameters or just a raw value.
+          const data = value.data || value;
```
turadg: why would it be one or the other?

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -231,19 +241,75 @@ const start = async zcf => {
...
+      ) => {
+        const admin = await creatorFacet.initOracle();
+        seat.exit();
+        if (!oracleNotifier) {
+          // No notifier to track, just let them have the direct admin.
+          return admin;
+        }
+
+        // Adapt the notifier to push results.
+        const recurse = ({ value, updateCount }) => {
```
turadg, line 281: I'm not clear on what `value` can be. A type annotation would help. Is `value` readonly?

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -24,14 +24,17 @@ const { add, multiply, floorDivide, ceilDivide, isGTE } = natSafeMath;
...
  * unitAmountIn: Amount,
  * }>} zcf
  */
 const start = async zcf => {
   const {
     timer,
     POLL_INTERVAL,
-    brands: { In: brandIn, Out: brandOut },
+    brandIn,
+    brandOut,
```
dckc, line 37: any particular reason to move away from a `brands` keyword record?

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -67,8 +67,9 @@ test.before(
       'b1-aggregator',
     );
 
-    const link = makeIssuerKit('$LINK', AssetKind.NAT);
-    const usd = makeIssuerKit('$USD', AssetKind.NAT);
+    const link = makeIssuerKit('$ATOM', AssetKind.NAT);
+    const { brand: atomBrand } = makeIssuerKit('$ATOM', AssetKind.NAT);
+    const { brand: usdBrand } = makeIssuerKit('$USD', AssetKind.NAT);
```
dckc: `AssetKind.NAT` is the default, yes? Maybe get rid of it here while we're at it?

packages/zoe/src/contracts/priceAggregator.js
dckc: oic. This is moving from the brands part of the terms to the custom terms.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -231,19 +241,75 @@ const start = async zcf => {
...
+          E(oracleNotifier).getUpdateSince(updateCount).then(recurse);
+
+          // See if we have associated parameters or just a raw value.
+          const data = value.data || value;
+
+          // Push the current scaled result.
+          const scaledData = Math.floor(parseInt(data, 10) * scaleValueOut);
+          const newData = BigInt(scaledData);
+
+          if (value.data) {
```
turadg: line 268 sort of also has this condition. consider making one conditional:
```
const scaleData = (data) => BigInt(Math.floor(parseInt(data, 10) * scaleValueOut));

let result;
if (value.data) {
  // We have some associated parameters to push.
  result = { ...value, data: scaleData(value.data) };
} else {
  result = scaleData(value);
}
admin.pushResult(result).catch(console.error);
```

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -231,19 +241,75 @@ const start = async zcf => {
...
+            // Stop tracking the notifier on delete.
+            oracleNotifier = undefined;
+            return admin.delete();
+          },
+        });
+      };
+
+      return zcf.makeInvitation(offerHandler, 'oracle invitation');
+    },
+    async initOracle(oracleInstance = undefined, query = undefined) {
```
turadg: ditto on jsdoc as way to expression optional param

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -241,19 +241,75 @@ const start = async zcf => {
       });
       ({ priceAuthority, adminFacet: priceAuthorityAdmin } = paKit);
     },
-    async initOracle(oracleInstance, query = undefined) {
+    makeOracleInvitation: async () => {
```
dckc, line 252: My understanding of this is a little fuzzy. There a unit test that tells a little more about how it works?

_wishing for that widget that adds test coverage details to each PR_

Or do you feel like adding a paragraph of JSDoc?

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -241,19 +241,75 @@ const start = async zcf => {
...
+      ) => {
+        const admin = await creatorFacet.initOracle();
+        seat.exit();
+        if (!oracleNotifier) {
+          // No notifier to track, just let them have the direct admin.
+          return admin;
+        }
+
+        // Adapt the notifier to push results.
+        const recurse = ({ value, updateCount }) => {
```
dckc, line 281: I usually call this `recur`, but I just looked it up and the dictionary definition of `recurse` seems to match better.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -231,19 +241,75 @@ const start = async zcf => {
...
+          },
+        });
+      };
+
+      return zcf.makeInvitation(offerHandler, 'oracle invitation');
+    },
+    async initOracle(oracleInstance = undefined, query = undefined) {
       assert(quoteKit, X`Must initializeQuoteMint before adding an oracle`);
 
+      const oracleKey = oracleInstance || Far('fresh key');
```
turadg: I don't understand this "fresh key". Will be be fresh each time it's constructed?

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -151,17 +156,21 @@ const start = async zcf => {
...
   /**
    * @typedef {Object} OracleRecord
    * @property {(timestamp: Timestamp) => Promise<void>=} querier
    * @property {number} lastSample
    */
 
-  /** @type {LegacyMap<Instance, Set<OracleRecord>>} */
-  const instanceToRecords = makeLegacyMap('oracleInstance');
+  /**
+   * @typedef {{}} OracleKey
```
turadg, line 169: because TS is structurally typed, this make the type just documentation. Reading above it looks like
```suggestion
   * @typedef {Instance | Remotable<'fresh key'> } OracleKey
```

turadg (commented):
> @turadg, please review the AmountKind change and the contract updates.

Done. Thanks for the instructions and clean commits.

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -584,26 +588,29 @@ const start = async zcf => {
     async initOracle(oracleInstance) {
       assert(quoteKit, X`Must initializeQuoteMint before adding an oracle`);
 
+      /** @type {OracleKey} */
+      const oracleKey = oracleInstance || Far('oracleKey', {});
```
dckc, line 592: This seems to be the heart of the matter of this commit. But it's not clear to me why we would lack an `oracleInstance` here.

dckc (commented):
I looked at each commit. I'm not quite confident enough to approve. I'd like to see responses to some of my questions or maybe talk about it with you.

packages/zoe/src/contracts/priceAggregator.js
michaelfig: > any particular reason to move away from a `brands` keyword record?

Yes, the `brands` term is only populated from issuers given to the contract.  In this case, I am deliberately avoiding requiring issuers.

packages/zoe/src/contracts/priceAggregator.js
michaelfig: Three sentences of JSDoc it is!

packages/zoe/src/contracts/priceAggregator.js
michaelfig: Yes.  Every call to `Far` creates a new object identity.  I was leaning on default arguments to the `Far` function, but have made it more explicit now.

packages/zoe/src/contracts/priceAggregatorChainlink.js
michaelfig: I'm trying not to change the semantics of the chainlink contract (unused at this time).  Before it just dropped uncaught rejections on the floor (which really needs to be fixed in our `xsnap` module), but in the course of debugging I wanted to log them.

packages/zoe/src/contracts/priceAggregatorChainlink.js
michaelfig: It's an optional argument.

packages/zoe/src/contracts/priceAggregatorChainlink.js
michaelfig: IOU unit tests when we actually start using this contract.  As it stands, you're seeing my attempts to keep the types working while trying to debug what is going on.

packages/zoe/src/contracts/priceAggregatorChainlink.js
michaelfig: It is the type of an object without properties, which is what I want to indicate.  Maybe this is where JSDoc types diverge from TS interfaces?

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
michaelfig: No longer needed.

packages/wallet/api/src/wallet.js
michaelfig: Okay, I added a jsdoc to it.  How do I indicate to the reader of this code (who isn't using Typescript) that the meta argument is optional without overriding the underlying default?

turadg: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

I don't know the types of those params so improvising:
```
/**
 * @param {Offer} offer
 * @param {Meta} [meta]
 */
addOffer(offer, meta) {
```
(default values can be specified with `=` but `[meta=undefined]` I think is redundant because `undefined` is what any param without an argument resolves to)

packages/zoe/src/contracts/priceAggregatorChainlink.js
dckc: ok... I see this refs #3595 but doesn't claim to close it.

packages/zoe/src/contracts/priceAggregatorChainlink.js
dckc: ah... and even when it's omitted, you need a unique identity.

dckc (approved):
This looks like forward progress, to me.

I trust you'll consider @turadg 's comments as well.

packages/zoe/src/contracts/priceAggregatorChainlink.js
turadg: > It is the type of an object without properties, which is what I want to indicate.

In that case, 👍 . I assumed an `OracleKey` was something more specific than an object without properties. 

> Maybe this is where JSDoc types diverge from TS interfaces?

I'm not aware of any divergences.

PR #4470, opened 2022-02-05
fix: Encode Passables, not just keys

erights:
Generalize encodeKey to encodePassable, so we can apply this encoding to any passable.  We've talked about replacing marshal's JSON-based encoding with a compact binary one, such as syrup. The advantage of this encoding is it preserves sort order. Aside from that, seems comparable to things like syrup.

PR #4896, opened 2022-03-23
fix: renamings #4470 missed

erights:
See https://github.com/Agoric/agoric-sdk/pull/4470#pullrequestreview-917799775


mhofman (approved):
Thanks for adding the clarification, and for the rename.

PR #4897, opened 2022-03-23
add no-op dispatch.stopVat

warner:
This adds a no-op `dispatch.stopVat()` call to liveslots, and calls it as the first part of vat upgrade (the rest of which is still unimplemented). `stopVat` will be used to delete the vat's non-durable state, and decrement all refcounts properly.

refs #1848

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -61,6 +61,7 @@ function build(
   }
 
   let didStartVat = false;
+  let didStopVat = false;
 
   const outstandingProxies = new WeakSet();
 
@@ -893,6 +894,7 @@ function build(
 
   function deliver(target, method, argsdata, result) {
     assert(didStartVat);
+    assert(!didStopVat);
     insistCapData(argsdata);
     lsdebug(
       `ls[${forVatID}].dispatch.deliver ${target}.${method} -> ${result}`,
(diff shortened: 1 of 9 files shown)


packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1255,6 +1259,16 @@ function build(
     // we do not use vreffedObjectRegistry for root objects
   }
 
+  /**
+   * @returns { Promise<void> }
+   */
+  async function stopVat() {
+    assert(didStartVat);
+    assert(!didStopVat);
+    didStopVat = true;
```
FUDCo, line 1268: This assignment doesn't seem entirely no-op'y, since it will cause `notify` and `deliver` to throw.

FUDCo (approved):
I'll leave it to you to decide if this is really as much of a no-op as you intended, but it seems like it's not entirely.  But since this will only manifest if you actually do a `stopVat` and that's what you're working on I suspect it's probably not an actual problem other than perhaps a slightly confusing PR title.

packages/SwingSet/src/liveslots/liveslots.js
warner: yeah, I'm ok with that, it's less "no-op" and more "don't do nearly enough to qualify for the `stopVat` name but still let's establish exactly when and how this should be called and we'll fill in the rest of the work later" :)

PR #4906, opened 2022-03-23
autoformat jsdoc of some packages

turadg:
## Description

Enable https://github.com/Agoric/agoric-sdk/pull/4903 just for these packages,
- `ertp`
- `inter-protocol`
- `vats`

There are more formatting options we could consider eventually. This adopts the minimum set to reduce the friction to getting the main value: auto-formatting type annotations.

### Security Considerations

--

Until #4900 is complete we'll have to selectively enable paths with the glob.

### Testing Considerations

- [X] Verify that re-reunning `yarn format` is stable and only affects opt-in packages.
- [X] verify that `yarn prettier` in a opt-in path is formatted and others aren't
- [X] verify that `formatOnSave` in a opt-in path is formatted and others aren't

### DIFF
--- a/packages/inter-protocol/scripts/deploy-contracts.js
+++ b/packages/inter-protocol/scripts/deploy-contracts.js
@@ -43,14 +43,24 @@ export default async (homeP, endowments) => {
   });
 
   console.log('getting installCache...');
-  /** @type {CopyMap<string, {installation: Installation, boardId: string, path?: string}>} */
+  /**
+   * @type {CopyMap<
+   *   string,
+   *   { installation: Installation; boardId: string; path?: string }
+   * >}
+   */
   const initial = await provideWhen(scratch, 'installCache', () =>
     makeCopyMap([]),
   );
   console.log('initially:', initial.payload.keys.length, 'entries');
 
   // ISSUE: getCopyMapEntries of CopyMap<K, V> loses K, V.
-  /** @type {Map<string, {installation: Installation, boardId: string, path?: string}>} */
+  /**
+   * @type {Map<
+   *   string,
+   *   { installation: Installation; boardId: string; path?: string }
+   * >}
+   */
   const working = new Map(getCopyMapEntries(initial));
 
   let added = 0;
--- a/packages/inter-protocol/scripts/manual-price-feed.js
+++ b/packages/inter-protocol/scripts/manual-price-feed.js
@@ -8,24 +8,25 @@ import process from 'process';
  * After extracting the oracleAdmins to entries in home.scratch, you can use one
  * from the REPL like:
  *
- * lookup('agoricNames', 'oracleBrand', 'ATOM').then(brand => atom = brand)
- * -> [Object Alleged: ATOM brand]{}
- * lookup('agoricNames', 'oracleBrand', 'USD').then(brand => usd = brand)
- * -> [Object Alleged: USD brand]{}
- * pa = home.priceAuthority
- *  -> [Object Alleged: PriceAuthority]{}
- * qn = E(pa).makeQuoteNotifier({ value: 1n * 10n ** 6n, brand: atom }, usd)
- * -> [Object Alleged: QuoteNotifier]{}
- * oa = E(home.scratch).get("offerResult unknown#1652669688625")
- * -> [Object Alleged: OracleAdmin]{}
- * E(oa).pushResult('19.37')
- * E(qn).getUpdateSince()
+ * lookup('agoricNames', 'oracleBrand', 'ATOM').then(brand => atom = brand) ->
+ * [Object Alleged: ATOM brand]{} lookup('agoricNames', 'oracleBrand',
+ * 'USD').then(brand => usd = brand) -> [Object Alleged: USD brand]{} pa =
+ * home.priceAuthority -> [Object Alleged: PriceAuthority]{} qn =
+ * E(pa).makeQuoteNotifier({ value: 1n * 10n ** 6n, brand: atom }, usd) ->
+ * [Object Alleged: QuoteNotifier]{} oa = E(home.scratch).get("offerResult
+ * unknown#1652669688625") -> [Object Alleged: OracleAdmin]{}
+ * E(oa).pushResult('19.37') E(qn).getUpdateSince()
  *
- * @typedef {{ board: import('@agoric/vats').Board, chainTimerService, scratch, zoe }} Home
+ * @typedef {{
+ *   board: import('@agoric/vats').Board;
+ *   chainTimerService;
+ *   scratch;
+ *   zoe;
+ * }} Home
  * @param {Promise<Home>} homePromise
  * @param {object} root0
- * @param {(...path: string[]) => Promise<any>} root0.lookup
- * A promise for the references available from REPL home
+ * @param {(...path: string[]) => Promise<any>} root0.lookup A promise for the
+ *   references available from REPL home
  */
 export default async function priceAuthorityFromNotifier(
   homePromise,
--- a/packages/inter-protocol/scripts/manual-price-feed.js
+++ b/packages/inter-protocol/scripts/manual-price-feed.js
@@ -8,14 +8,18 @@ import process from 'process';
  * After extracting the oracleAdmins to entries in home.scratch, you can use one
  * from the REPL like:
  *
- * lookup('agoricNames', 'oracleBrand', 'ATOM').then(brand => atom = brand) ->
- * [Object Alleged: ATOM brand]{} lookup('agoricNames', 'oracleBrand',
- * 'USD').then(brand => usd = brand) -> [Object Alleged: USD brand]{} pa =
- * home.priceAuthority -> [Object Alleged: PriceAuthority]{} qn =
- * E(pa).makeQuoteNotifier({ value: 1n * 10n ** 6n, brand: atom }, usd) ->
- * [Object Alleged: QuoteNotifier]{} oa = E(home.scratch).get("offerResult
- * unknown#1652669688625") -> [Object Alleged: OracleAdmin]{}
- * E(oa).pushResult('19.37') E(qn).getUpdateSince()
+ *     lookup('agoricNames', 'oracleBrand', 'ATOM').then(brand => atom = brand)
+ *     -> [Object Alleged: ATOM brand]{}
+ *     lookup('agoricNames', 'oracleBrand', 'USD').then(brand => usd = brand)
+ *     -> [Object Alleged: USD brand]{}
+ *     pa = home.priceAuthority
+ *      -> [Object Alleged: PriceAuthority]{}
+ *     qn = E(pa).makeQuoteNotifier({ value: 1n * 10n ** 6n, brand: atom }, usd)
+ *     -> [Object Alleged: QuoteNotifier]{}
+ *     oa = E(home.scratch).get("offerResult unknown#1652669688625")
+ *     -> [Object Alleged: OracleAdmin]{}
+ *     E(oa).pushResult('19.37')
+ *     E(qn).getUpdateSince()
  *
  * @typedef {{
  *   board: import('@agoric/vats').Board;
--- a/packages/inter-protocol/src/auction/auctionBook.js
+++ b/packages/inter-protocol/src/auction/auctionBook.js
@@ -34,34 +34,38 @@ const DEFAULT_DECIMALS = 9;
 
 /**
  * @file The book represents the collateral-specific state of an ongoing
- * auction. It holds the book, the capturedPrice, and the collateralSeat that has
- * the allocation of assets for sale.
+ *   auction. It holds the book, the capturedPrice, and the collateralSeat that
+ *   has the allocation of assets for sale.
  *
- * The book contains orders for the collateral. It holds two kinds of
- * orders:
+ *   The book contains orders for the collateral. It holds two kinds of orders:
+ *
  *   - Prices express the bidding offer in terms of a Bid amount
- *   - ScaledBids express the offer in terms of a discount (or markup) from the
- *     most recent oracle price.
+ *   - ScaledBids express the offer in terms of a discount (or markup) from the most
+ *       recent oracle price.
  *
- * Offers can be added in three ways. 1) When the auction is not active, prices
- * are automatically added to the appropriate collection. When the auction is
- * active, 2) if a new offer is at or above the current price, it will be
- * settled immediately; 2) If the offer is below the current price, it will be
- * added in the appropriate place and settled when the price reaches that level.
+ *   Offers can be added in three ways. 1) When the auction is not active, prices
+ *   are automatically added to the appropriate collection. When the auction is
+ *   active, 2) if a new offer is at or above the current price, it will be
+ *   settled immediately; 2) If the offer is below the current price, it will be
+ *   added in the appropriate place and settled when the price reaches that
+ *   level.
  */
 
 const trace = makeTracer('AucBook', true);
 
 /**
  * @typedef {{
- *   maxBuy: Amount<'nat'>
+ *   maxBuy: Amount<'nat'>;
  * } & {
- *   exitAfterBuy?: boolean,
- * } & ({
- *   offerPrice: Ratio,
- * } | {
- *    offerBidScaling: Ratio,
- * })} OfferSpec
+ *   exitAfterBuy?: boolean;
+ * } & (
+ *     | {
+ *         offerPrice: Ratio;
+ *       }
+ *     | {
+ *         offerBidScaling: Ratio;
+ *       }
+ *   )} OfferSpec
  */
 /**
  * @param {Brand<'nat'>} bidBrand
@@ -88,17 +92,19 @@ export const makeOfferSpecShape = (bidBrand, collateralBrand) => {
 
 /**
  * @typedef {object} BookDataNotification
- *
- * @property {Ratio         | null}      startPrice identifies the priceAuthority and price
- * @property {Ratio         | null}      currentPriceLevel the price at the current auction tier
- * @property {Amount<'nat'> | null}      startProceedsGoal The proceeds the sellers were targeting to raise
- * @property {Amount<'nat'> | null}      remainingProceedsGoal The remainder of
- *     the proceeds the sellers were targeting to raise
- * @property {Amount<'nat'> | undefined} proceedsRaised The proceeds raised so far in the auction
- * @property {Amount<'nat'>}             startCollateral How much collateral was
- *    available for sale at the start. (If more is deposited later, it'll be
- *    added in.)
- * @property {Amount<'nat'> | null}      collateralAvailable The amount of collateral remaining
+ * @property {Ratio | null} startPrice identifies the priceAuthority and price
+ * @property {Ratio | null} currentPriceLevel the price at the current auction
+ *   tier
+ * @property {Amount<'nat'> | null} startProceedsGoal The proceeds the sellers
+ *   were targeting to raise
+ * @property {Amount<'nat'> | null} remainingProceedsGoal The remainder of the
+ *   proceeds the sellers were targeting to raise
+ * @property {Amount<'nat'> | undefined} proceedsRaised The proceeds raised so
+ *   far in the auction
+ * @property {Amount<'nat'>} startCollateral How much collateral was available
+ *   for sale at the start. (If more is deposited later, it'll be added in.)
+ * @property {Amount<'nat'> | null} collateralAvailable The amount of collateral
+ *   remaining
  */
 
 /**
@@ -231,8 +237,8 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
     {
       helper: {
         /**
-         * remove the key from the appropriate book, indicated by whether the price
-         * is defined.
+         * remove the key from the appropriate book, indicated by whether the
+         * price is defined.
          *
          * @param {string} key
          * @param {Ratio | undefined} price
@@ -247,8 +253,8 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
         },
 
         /**
-         * Update the entry in the appropriate book, indicated by whether the price
-         * is defined.
+         * Update the entry in the appropriate book, indicated by whether the
+         * price is defined.
          *
          * @param {string} key
          * @param {Amount} collateralSold
@@ -264,7 +270,8 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
         },
 
         /**
-         * Settle with seat. The caller is responsible for updating the book, if any.
+         * Settle with seat. The caller is responsible for updating the book, if
+         * any.
          *
          * @param {ZCFSeat} seat
          * @param {Amount<'nat'>} collateralWanted
@@ -341,15 +348,16 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
         },
 
         /**
-         *  Accept an offer expressed as a price. If the auction is active, attempt to
-         *  buy collateral. If any of the offer remains add it to the book.
+         * Accept an offer expressed as a price. If the auction is active,
+         * attempt to buy collateral. If any of the offer remains add it to the
+         * book.
          *
-         *  @param {ZCFSeat} seat
-         *  @param {Ratio} price
-         *  @param {Amount<'nat'>} maxBuy
-         *  @param {object} opts
-         *  @param {boolean} opts.trySettle
-         *  @param {boolean} [opts.exitAfterBuy]
+         * @param {ZCFSeat} seat
+         * @param {Ratio} price
+         * @param {Amount<'nat'>} maxBuy
+         * @param {object} opts
+         * @param {boolean} opts.trySettle
+         * @param {boolean} [opts.exitAfterBuy]
          */
         acceptPriceOffer(
           seat,
@@ -388,16 +396,16 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
         },
 
         /**
-         *  Accept an offer expressed as a discount (or markup). If the auction is
-         *  active, attempt to buy collateral. If any of the offer remains add it to
-         *  the book.
+         * Accept an offer expressed as a discount (or markup). If the auction
+         * is active, attempt to buy collateral. If any of the offer remains add
+         * it to the book.
          *
-         *  @param {ZCFSeat} seat
-         *  @param {Ratio} bidScaling
-         *  @param {Amount<'nat'>} maxBuy
-         *  @param {object} opts
-         *  @param {boolean} opts.trySettle
-         *  @param {boolean} [opts.exitAfterBuy]
+         * @param {ZCFSeat} seat
+         * @param {Ratio} bidScaling
+         * @param {Amount<'nat'>} maxBuy
+         * @param {object} opts
+         * @param {boolean} opts.trySettle
+         * @param {boolean} [opts.exitAfterBuy]
          */
         acceptScaledBidOffer(
           seat,
@@ -466,10 +474,10 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
          * @param {Amount<'nat'>} assetAmount
          * @param {ZCFSeat} sourceSeat
          * @param {Amount<'nat'>} [proceedsGoal] an amount that the depositor
-         *    would like to raise. The auction is requested to not sell more
-         *    collateral than required to raise that much. The auctioneer might
-         *    sell more if there is more than one supplier of collateral, and
-         *    they request inconsistent limits.
+         *   would like to raise. The auction is requested to not sell more
+         *   collateral than required to raise that much. The auctioneer might
+         *   sell more if there is more than one supplier of collateral, and
+         *   they request inconsistent limits.
          */
         addAssets(assetAmount, sourceSeat, proceedsGoal) {
           const { state, facets } = this;
@@ -754,7 +762,11 @@ export const prepareAuctionBook = (baggage, zcf, makeRecorderKit) => {
     },
   );
 
-  /** @type {(...args: Parameters<typeof makeAuctionBookKit>) => ReturnType<typeof makeAuctionBookKit>['self']} */
+  /**
+   * @type {(
+   *   ...args: Parameters<typeof makeAuctionBookKit>
+   * ) => ReturnType<typeof makeAuctionBookKit>['self']}
+   */
   const makeAuctionBook = (...args) => makeAuctionBookKit(...args).self;
   return makeAuctionBook;
 };
--- a/packages/inter-protocol/src/auction/auctioneer.js
+++ b/packages/inter-protocol/src/auction/auctioneer.js
@@ -43,11 +43,10 @@ const { add, multiply } = natSafeMath;
 const trace = makeTracer('Auction', true);
 
 /**
- * @file
- * In this file, 'Bid' is the name of the ERTP issuer used to purchase
- * collateral from various issuers. It's too confusing to also use Bid as a verb
- * or a description of amounts offered, so we've tried to find alternatives in
- * all those cases.
+ * @file In this file, 'Bid' is the name of the ERTP issuer used to purchase
+ *   collateral from various issuers. It's too confusing to also use Bid as a
+ *   verb or a description of amounts offered, so we've tried to find
+ *   alternatives in all those cases.
  */
 
 const MINIMUM_BID_GIVE = 1n;
@@ -66,18 +65,19 @@ const makeBPRatio = (rate, bidBrand, collateralBrand = bidBrand) =>
 /**
  * The auction sold some amount of collateral, and raised a certain amount of
  * Bid. The excess collateral was returned as `unsoldCollateral`. The Bid amount
- * collected from the auction participants is `proceeds`.
- * Return a set of transfers for atomicRearrange() that distribute
- * `unsoldCollateral` and `proceeds` proportionally to each seat's deposited
- * amount. Any uneven split should be allocated to the reserve.
+ * collected from the auction participants is `proceeds`. Return a set of
+ * transfers for atomicRearrange() that distribute `unsoldCollateral` and
+ * `proceeds` proportionally to each seat's deposited amount. Any uneven split
+ * should be allocated to the reserve.
  *
  * @param {Amount} unsoldCollateral
  * @param {Amount} proceeds
- * @param {{seat: ZCFSeat, amount: Amount<'nat'>, goal: Amount<'nat'>}[]} deposits
+ * @param {{ seat: ZCFSeat; amount: Amount<'nat'>; goal: Amount<'nat'> }[]} deposits
  * @param {ZCFSeat} collateralSeat
- * @param {ZCFSeat} bidHoldingSeat seat with the Bid allocation to be distributed
+ * @param {ZCFSeat} bidHoldingSeat seat with the Bid allocation to be
+ *   distributed
  * @param {string} collateralKeyword The Reserve will hold multiple collaterals,
- *      so they need distinct keywords
+ *   so they need distinct keywords
  * @param {ZCFSeat} reserveSeat
  * @param {Brand} brand
  */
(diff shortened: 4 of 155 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -25,27 +25,30 @@ const trace = makeTracer('IV');
...
- * TRANSFER     - vault is able to be transferred (payments and debits frozen until it has a new owner)
- * CLOSED       - vault was closed by the user and all assets have been paid out
- * LIQUIDATED   - vault was closed by the manager, with remaining assets paid to owner
+ * - ACTIVE - vault is in use and can be changed
+ * - LIQUIDATING - vault is being liquidated by the vault manager, and cannot be
+ *   changed by the user
+ * - TRANSFER - vault is able to be transferred (payments and debits frozen until
+ *   it has a new owner)
+ * - CLOSED - vault was closed by the user and all assets have been paid out
+ * - LIQUIDATED - vault was closed by the manager, with remaining assets paid to owner
```
dckc: This is harder to read; is there a convention for vertical alignment in JSDoc?

turadg: Good call. Yeah, instead of list items I should have just indented (c43b982cba82928eadfa5d18219d2284f29382cd):
Screen Shot 2022-03-24 at 7 05 08 AM

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -37,24 +37,21 @@
...
  */
 
 /**
- * @callback ReallocateWithFee
- *
- * Transfer the indicated amount to the vaultFactory's reward
- * pool, taken from the `fromSeat`. Then reallocate over all the seat
- * arguments and the rewardPoolSeat.
- *
+ * @callback ReallocateWithFee Transfer the indicated amount to the
```
Chris-Hibbert: This change makes it harder to distinguish the comment from the name when reading the source. Is there punctuation that can be inserted (' - ' for instance) that makes it more visible?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -25,27 +25,28 @@ const trace = makeTracer('IV');
...
 /**
  * Constants for vault phase.
  *
- * ACTIVE       - vault is in use and can be changed
- * LIQUIDATING  - vault is being liquidated by the vault manager, and cannot be changed by the user
- * TRANSFER     - vault is able to be transferred (payments and debits frozen until it has a new owner)
- * CLOSED       - vault was closed by the user and all assets have been paid out
- * LIQUIDATED   - vault was closed by the manager, with remaining assets paid to owner
+ *     ACTIVE       - vault is in use and can be changed
+ *     LIQUIDATING  - vault is being liquidated by the vault manager, and cannot be changed by the user
```
Chris-Hibbert: Is it the presence of the dashes here that told it not to reflow?   Did it notice that the dashes lined up, or did it accidentally preserve that?

Chris-Hibbert (commented):
just getting started.  Here are some reactions.

On the whole, I like the changes. I don't understand when it switches commas to semicolons in lists, though.

packages/run-protocol/src/vaultFactory/vault.js
turadg: The dashes were lined up before. The indentation told it not to reformat. Though there is a bug to watch out for https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/155

packages/run-protocol/src/vaultFactory/types.js
turadg: Agree for code reviews. In an IDE it's still quite readable:
Screen Shot 2022-03-24 at 9 26 08 AM
 
 I don't see an option to change this. I could file one but I'm not sure what to ask for… that an exception be made for callbacks?

I think this would be most in line with JSDoc and our other function comments:
```
/**
 * Transfer the indicated amount to the vaultFactory's reward pool, taken from
 * the `fromSeat`. Then reallocate over all the seat arguments and the rewardPoolSeat.
 *
 * @callback ReallocateWithFee
 * @param {Amount} amount
 * @param {ZCFSeat} fromSeat
 * @param {ZCFSeat} [otherSeat]
 * @returns {void}
 */
```
which renders on hover as,
Screen Shot 2022-03-24 at 9 30 07 AM

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -21,10 +21,10 @@
 
 /**
  * @typedef {Object} Rates
- * @property {Ratio} liquidationMargin - Margin below which collateral will be
+ * @property {Ratio} liquidationMargin Margin below which collateral will be
  *   liquidated to satisfy the debt.
- * @property {Ratio} interestRate - Annual interest rate charged on loans
- * @property {Ratio} loanFee - The fee (in BasisPoints) charged when opening or
+ * @property {Ratio} interestRate Annual interest rate charged on loans
+ * @property {Ratio} loanFee The fee (in BasisPoints) charged when opening or
```
Chris-Hibbert: Dropping the dashes leads to worse readability in the source. Are the dashes not understood as separators by the jsdoc interpreter?

packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
```diff
@@ -53,33 +53,35 @@ export { imbalancedRequest };
...
- *   so targetK equals startK because we square targetY
- * targetK = targetX * targetY = desiredRatio * (startK / desiredRatio)
+ *     endX = poolX + giveX;   endY = poolY + giveY
+ *     desiredRatio = endX / endY
+ *     targetY = sqrt(startK / desiredRatio)
+ *     targetX = desiredRatio * targetY
+ *       so targetK equals startK because we square targetY
+ *     targetK = targetX * targetY = desiredRatio * (startK / desiredRatio)
+ *
+ * Since startK/endK is less than one, and we have to worry about early loss of
```
Chris-Hibbert: Repeated line?

packages/run-protocol/src/vpool-xyk-amm/constantProduct/calcFees.js
```diff
@@ -69,8 +69,7 @@ const calcFee = ({ amountIn, amountOut }, feeRatio) => {
 
 /**
  * Estimate the swap values, then calculate fees. The swapFn provided by the
- * caller will be swapInNoFees or swapOutNoFees.
- * SwapOut.
+ * caller will be swapInNoFees or swapOutNoFees. SwapOut.
```
Chris-Hibbert: I think *SwapOut* is a mistake.  drop it.

packages/run-protocol/src/vpool-xyk-amm/constantProduct/core.js
```diff
@@ -58,13 +58,13 @@ const ceilMultiplyKeepBrand = (amount, ratio) => {
...
+ *     Equivalently: (deltaX / (deltaX + x)) * y;
  *
- * @param {Amount} x - the amount of Brand X in pool
- * @param {Amount} y - the amount of Brand Y the pool
- * @param {Amount} deltaX - the amount of Brand X to be added
- * @returns {Amount} deltaY - the amount of Brand Y to be taken out
+ * @param {Amount} x The amount of Brand X in pool
+ * @param {Amount} y The amount of Brand Y the pool
+ * @param {Amount} deltaX The amount of Brand X to be added
+ * @returns {Amount} The amount of Brand Y to be taken out
```
Chris-Hibbert: dropping the dashes reduces readability in the source.  Is there an alternative?

packages/run-protocol/src/vpool-xyk-amm/constantProduct/invariants.js
```diff
@@ -6,12 +6,12 @@ import { AmountMath } from '@agoric/ertp';
 import { natSafeMath } from '@agoric/zoe/src/contractSupport/index.js';
 
 /**
- * xy <= (x + deltaX)(y - deltaY)
+ * `Xy <= (x + deltaX)(y - deltaY)`
```
Chris-Hibbert: changing from lower case is a mistake.

packages/run-protocol/src/vpool-xyk-amm/constantProduct/invariants.js
```diff
@@ -22,12 +22,12 @@ export const checkKInvariantSellingX = (x, y, deltaX, deltaY) => {
 };
 
 /**
- * xy <= (x + deltaX)(y - deltaY)
+ * `Xy <= (x + deltaX)(y - deltaY)`
```
Chris-Hibbert: `xy` should still be lower case

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -95,32 +93,30 @@ const { quote: q, details: X } = assert;
  * The creatorFacet has one method (makeCollectFeesInvitation, which returns
  * collected fees to the creator). `handleParamGovernance()` adds internal
  * methods used by the contractGovernor. The contractGovernor then has access to
- * those internal methods, and reveals the original AMM creatorFacet to its own
- * creator.
+ * those internal methods, and reveals the original AMM creatorFacet to its own creator.
```
Chris-Hibbert: The line breaking algorithm is surprising.  If the last line can be this long, why didn't it re-flow the previous lines?

```
 * The creatorFacet has one method (makeCollectFeesInvitation, which returns collected
 * fees to the creator). `handleParamGovernance()` adds internal methods used by the
 * contractGovernor. The contractGovernor then has access to those internal methods, and
 * reveals the original AMM creatorFacet to its own creator.
 ```

I've seen other examples above where the last line is visibly longer than those that precede it.  

It's a benefit to not have to think about line breaking, but this is a bizarre approach.

File: packages/run-protocol/src/vpool-xyk-amm/singlePool.js [hunk 11]
```diff
@@ -10,8 +10,8 @@ import {
 /**
  * @param {ZCF} zcf
  * @param {XYKPool} pool
- * @param {() => bigint} getProtocolFeeBP - retrieve governed protocol fee value
- * @param {() => bigint} getPoolFeeBP - retrieve governed pool fee value
+ * @param {() => bigint} getProtocolFeeBP - Retrieve governed protocol fee value
+ * @param {() => bigint} getPoolFeeBP - Retrieve governed pool fee value
```
Chris-Hibbert: See!  The dashes can be kept.

packages/run-protocol/src/vaultFactory/types.js
Chris-Hibbert: It's nice that VSCode does that, but WebStorm doesn't color code the types.js file.  

As a matter of fact, I now see that  at the call site for reallocateWithFee, WebStorm tells me that the type is `ReallocateWithFee`, and doesn't give any help on the parameters.  (This is at line 589 in vault.js.  WS shows no parameter help for `manager.reallocateWithFee(`, but the adjacent `transfer` shows a popup of `fromSeat, toSeat, fromLoses, toGains, key`.)  I've been noticing missing parameter help recently, but didn't notice that it was simultaneously present on other calls.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -23,7 +23,7 @@ const { details: X, quote: q } = assert;
 const trace = makeTracer('IV');
 
 /**
- * @file This has most of the logic for a Vault, to borrow RUN against collateral.
+ * @file This Has most of the logic for a Vault, to borrow RUN against collateral.
```
turadg: I don't think this `@file` take a keyword before the prose

packages/run-protocol/src/vaultFactory/types.js
Chris-Hibbert: > It's nice that VSCode does that, but WebStorm doesn't color code the types.js file.

That was just the style I was using. It did distinguish and allowed me to make them more distinct.

> at the call site for reallocateWithFee, WebStorm tells me that the type is ReallocateWithFee, and doesn't give any help on the parameters.

After some experimentation with @turadg, we found that `@callback` didn't pass the parameters through, but a direct declaration of the arguments did.  I will experiment so we don't have to describe all the parameters at each declaration.  (The particular example I was experimenting with was passed through several layers of API to get to where it was being used.)

packages/run-protocol/src/vaultFactory/types.js
Chris-Hibbert: I withdraw this. The IDE does a decent job of distinguishing defined names from commentary.

packages/run-protocol/src/vpool-xyk-amm/constantProduct/invariants.js
Chris-Hibbert: Interesting, it both added backticks and changed the case. I don't think it should have done either.

packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
turadg: oops yeah, bad copy paste in the manual fix commit

packages/run-protocol/src/vpool-xyk-amm/constantProduct/invariants.js
turadg: Human error: I added the backticks so it wouldn't change the case, but I forgot to undo its case change. Fixed.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: My guess is it has a width tolerance factor to prevent runts (had to [look that up](https://opusdesign.us/wordcount/typographic-widows-orphans/)).

I'm happy that code formatters mean I don't have to think about line wrapping. =)

packages/run-protocol/src/vpool-xyk-amm/singlePool.js
turadg: Yes, it doesn't do any dash removal. That was me manually to be consistent with the places we don't use dashes and because the new capitalization provides a distinguishing feature when syntax highlighting isn't available.

packages/run-protocol/src/vaultFactory/vault.js
turadg: this was fixed 8 hours ago before my comment! https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/143#issuecomment-1077458920

018d23930f953f0e8bdfb282a4ac334edcb145a2

turadg:
> On the whole, I like the changes.

Me too. I'm going to mark this ready for review.

UPDATE: I noticed in VS Code that when the plugin is installed by Yarn that my `saveOnFormat` is formatting everywhere because it's not getting the `--no-plugin-search` option in package.json. I think this is a blocker. I'll try to solve before marking for review again.

> I don't understand when it switches commas to semicolons in lists, though.

That's for TypeScript fragments within a `@typedef`, which is more in line with TypeScript conventions. It's also what [Prettier would do for plain `.ts`](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAYhCbAXm2AB0pttIBbGgQwEYlsBnGAJwEsoBzAGgpVaDAEwsoAVxoAjOB0GU2cGlwnS5HANwUAvhRD8QEDDC7RWyUPQ4cIAdwAKNhJZT0ANvfppLRmRz0YADWcDAAyvQ0cAAyPHDIAGaerHD+gSFh4RhBPLzInJJpICpyACZlcGUx9HyS9Lxw+BwMMGZ8yCD0kjAQhiAAFjA0HgDqA1zwrDlgcOGuk1wAbpNonWCsfiA8qRwwjoG8DEkpxQBWrAAe4XkecACKkhDwJx6pRjkcu53oWKxg3FM-Qw3Fgoy4ZRgA2QAA4AAwfOypUaBDCdEFwXZLBJGACOT3gBxMbi6rAAtFA4FUqv0OHB8Vw6QcGsckMk3sVUqoChwikZWLcHgSEmzTkYYPQZODIdCkKJxYEuB48gBhCB0eidTEAVn6klSABVJW52e8QEsigBJKCVWDhAFcUwAQRt4XQd1eqV0uiAA).

packages/run-protocol/src/interest.js
```diff
@@ -92,7 +89,7 @@ export const makeInterestCalculator = (
 };
 
 /**
- * compoundedInterest *= (new debt) / (prior total debt)
+ * CompoundedInterest *= (new debt) / (prior total debt)
```
erights: The upper casing here is a bit unfortunate. But since this is not code, it is not worth fixing.

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -37,24 +37,22 @@
...
 /**
- * @callback ReallocateWithFee
- *
- * Transfer the indicated amount to the vaultFactory's reward
- * pool, taken from the `fromSeat`. Then reallocate over all the seat
- * arguments and the rewardPoolSeat.
+ * Transfer the indicated amount to the vaultFactory's reward pool, taken from
+ * the `fromSeat`. Then reallocate over all the seat arguments and the rewardPoolSeat.
  *
+ * @callback ReallocateWithFee
```
erights: Did it swap the declaration and prose automatically, or did manually intervene? Do they both mean the same thing in jsdoc?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -25,27 +25,28 @@ const trace = makeTracer('IV');
...
- * A note on naming convention:
- * - `Pre` is used as a postfix for any mutable value retrieved *before* an
- *    `await`, to flag values that must used very carefully after the `await`
- * - `new` is a prefix for values that describe the result of executing a
- *   transaction; e.g., `debt` is the value before the txn, and `newDebt`
- *   will be value if the txn completes.
- * - the absence of one of these implies the opposite, so `newDebt` is the
- *   future value fo `debt`, as computed based on values after any `await`
+ *   A note on naming convention:
+ *
```
erights: Did it insert a blank line here? Why?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -242,8 +247,9 @@ export const makeInnerVault = (
    * by their debt-to-collateral ratios without having to mutate the debts as
    * the interest accrues.
    *
+   * @returns {Amount<'nat'>} As if the vault was open at the launch of this
+   *   manager, before any interest accrued
    * @see getActualDebAmount
```
erights: Swapped automatically?

packages/run-protocol/src/vpool-xyk-amm/constantProduct/calcFees.js
```diff
@@ -32,7 +32,7 @@ const amountGT = (left, right) =>
  * Apply the feeRatio to the amount that has a matching brand. This used to
  * calculate fees in the single pool case.
  *
- * @param {{ amountIn: Amount, amountOut: Amount}} amounts - a record with two
+ * @param {{ amountIn: Amount; amountOut: Amount }} amounts - A record with two
```
erights: Why didn't it remove this hyphen?

packages/run-protocol/src/vpool-xyk-amm/constantProduct/core.js
```diff
@@ -80,13 +80,12 @@ export const calcDeltaYSellingX = (x, y, deltaX) => {
  * pool for that asset. swapInReduced calls this with the calculated amountOut
  * to find out if less than the offeredAmountIn would be sufficient.
  *
- * deltaX = (deltaYOverY/(1 - deltaYOverY))*x
- * Equivalently: (deltaY / (Y - deltaY )) * x
+ * DeltaX = (deltaYOverY/(1 deltaYOverY))*x Equivalently: (deltaY / (Y - deltaY )) * x
```
erights: This lost linebreak expressed intended meaning. Needs to be restored somehow.

packages/run-protocol/src/vpool-xyk-amm/constantProduct/invariants.js
```diff
@@ -6,12 +6,12 @@ import { AmountMath } from '@agoric/ertp';
 import { natSafeMath } from '@agoric/zoe/src/contractSupport/index.js';
 
 /**
- * xy <= (x + deltaX)(y - deltaY)
+ * `xy <= (x + deltaX)(y - deltaY)`
```
erights: Ok, this was manual. But still curious: Did you do the manual intervention on the input or the output?

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
erights: TIL runt

erights (approved):
LGTM, thanks!

packages/run-protocol/src/vaultFactory/types.js
turadg: This was manual. When the prose is below a tag it is treated as part of the tag and put onto the same line. I.e. `@callback ReallocateWithFee Transfer the indicated amount…`. IDE renders it the same either way way.

I thought of filing that as a bug but on second thought I figured it was a reasonable thing to keep tags grouped together at the bottom like in other functions. That's the ordering in JSDoc [examples of callback](https://jsdoc.app/tags-callback.html).

Thinking through it more, it's better for the tool not to move `@typedef` or `@callback` descriptions. I've filed https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/156

packages/run-protocol/src/interest.js
turadg: Agreed. It was a kind of hybrid code. As code it should have backticks. As not code, it would be better without camelCase.

turadg: I'm going to fix as it's pseudo-code

packages/run-protocol/src/vaultFactory/vault.js
turadg: Yes. I think because list items followed.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Yes, it puts `@see` at the end

packages/run-protocol/src/vpool-xyk-amm/constantProduct/calcFees.js
turadg: It doesn't do anything with hyphens. I removed some manually for consistency in this package. Rationale being that the capitalization is enough to set it apart from the param name.

packages/run-protocol/src/vpool-xyk-amm/constantProduct/invariants.js
turadg: If the question is whether the formatter would modify it without backticks, yes it would capitalize because `jsdocCapitalizeDescription` is left at its default of `true`.

Alternately I can disable that and we can leave capitalization alone. IMO it's better to enable it to force us to notice what is interpreted as prose, because any rendering of the docstring will render it that way.

Though I can also accept that even if we do agree to that, as a matter of migration we tackle that in a separate stage so that just getting all onto this tool happens faster.

turadg:
This is blocked on getting the selective opt-in to work. Reported in `https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/157`

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -12,67 +12,65 @@ const { values } = Object;
...
+ *   The public facet provides access to invitations to get a loan or to return an
+ *   attestation in order to release a lien on staked assets.
  *
  * @typedef {{
- *   makeLoanInvitation: () => Promise<Invitation>,
- *   makeReturnAttInvitation: () => Promise<Invitation>,
+ *   makeLoanInvitation: () => Promise<Invitation>;
+ *   makeReturnAttInvitation: () => Promise<Invitation>;
  * }} RunStakePublic
+ *   To take out a loan, get an `AttestationMaker` for your address from the
```
turadg: @dckc the formatter reads the paragraph following a typedef as being about that typedef and so indents it.

I think this will be fixed by `https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/156`

packages/run-protocol/test/reserve/setup.js
```diff
@@ -61,27 +61,27 @@ const setupReserveBootstrap = async (
...
- *     zoe: ZoeService,
- *     feeMintAccess: FeeMintAccess,
- *     installation: Installation,
- *     committeeCreator: ElectorateCreatorFacet,
- *     electorateInstance: Instance,
- *     governor: any,
- *     reserve: ReserveKit,
- *     invitationAmount: Amount,
- *     space: any,
+ *   zoe: ZoeService;
```
kriskowal: semicolons. That’s interesting. Is that consistent with TS?

kriskowal (commented):
Good idea as far as I’m concerned.

packages/run-protocol/test/reserve/setup.js
mhofman: I believe TSC prefers semicolons

packages/inter-protocol/scripts/manual-price-feed.js
```diff
@@ -8,24 +8,25 @@ import process from 'process';
...
- * E(oa).pushResult('19.37')
- * E(qn).getUpdateSince()
+ * lookup('agoricNames', 'oracleBrand', 'ATOM').then(brand => atom = brand) ->
+ * [Object Alleged: ATOM brand]{} lookup('agoricNames', 'oracleBrand',
+ * 'USD').then(brand => usd = brand) -> [Object Alleged: USD brand]{} pa =
+ * home.priceAuthority -> [Object Alleged: PriceAuthority]{} qn =
+ * E(pa).makeQuoteNotifier({ value: 1n * 10n ** 6n, brand: atom }, usd) ->
+ * [Object Alleged: QuoteNotifier]{} oa = E(home.scratch).get("offerResult
+ * unknown#1652669688625") -> [Object Alleged: OracleAdmin]{}
+ * E(oa).pushResult('19.37') E(qn).getUpdateSince()
```
dckc: this is mangling a REPL session.

I'm not sure how to fix.

packages/inter-protocol/scripts/deploy-contracts.js
```diff
@@ -43,14 +43,24 @@ export default async (homeP, endowments) => {
   });
 
   console.log('getting installCache...');
-  /** @type {CopyMap<string, {installation: Installation, boardId: string, path?: string}>} */
+  /**
+   * @type {CopyMap<
```
dckc, line 47: I think this file is dead code. I guess it doesn't hurt to tweak it, though...

packages/inter-protocol/src/auction/auctionBook.js
```diff
@@ -34,34 +34,38 @@ const DEFAULT_DECIMALS = 9;
...
- *    offerBidScaling: Ratio,
- * })} OfferSpec
+ *   exitAfterBuy?: boolean;
+ * } & (
+ *     | {
+ *         offerPrice: Ratio;
+ *       }
+ *     | {
+ *         offerBidScaling: Ratio;
+ *       }
```
dckc, line 67: if we write each of these as 1 line instead of 3, will the tool leave them alone?

packages/inter-protocol/src/auction/auctioneer.js
```diff
@@ -66,18 +65,19 @@ const makeBPRatio = (rate, bidBrand, collateralBrand = bidBrand) =>
 /**
  * The auction sold some amount of collateral, and raised a certain amount of
  * Bid. The excess collateral was returned as `unsoldCollateral`. The Bid amount
- * collected from the auction participants is `proceeds`.
- * Return a set of transfers for atomicRearrange() that distribute
- * `unsoldCollateral` and `proceeds` proportionally to each seat's deposited
- * amount. Any uneven split should be allocated to the reserve.
+ * collected from the auction participants is `proceeds`. Return a set of
```
dckc: does `@returns` make sense here?

packages/inter-protocol/src/auction/auctioneer.js
```diff
@@ -131,41 +131,42 @@ const distributeProportionalShares = (
...
```
dckc: again: `@returns`?

packages/inter-protocol/src/auction/offerBook.js
```diff
@@ -25,12 +25,14 @@ const nextSequenceNumber = () => {
...
- * received: Amount<'nat'>,
- * } &  {exitAfterBuy: boolean} & ({ bidScaling: Pattern, price: undefined } | { bidScaling: undefined, price: Ratio})
- * } BidderRecord
+ *   seat: ZCFSeat;
+ *   wanted: Amount<'nat'>;
+ *   seqNum: NatValue;
+ *   received: Amount<'nat'>;
+ * } & { exitAfterBuy: boolean } & (
+ *     | { bidScaling: Pattern; price: undefined }
+ *     | { bidScaling: undefined; price: Ratio }
```
dckc, line 34: this answers my earlier question: yes, the tool leaves it alone if we write these as 1 line each instead of 3

packages/inter-protocol/src/auction/offerBook.js
```diff
@@ -51,7 +54,6 @@ export const prepareScaledBidBook = baggage =>
     'scaledBidBook',
     undefined,
     /**
-     *
```
dckc, line 54: Removing that blank line when it's not used is nice.

packages/inter-protocol/test/reserve/test-reserve.js
```diff
@@ -12,8 +12,8 @@ import { setupReserveServices } from './setup.js';
 /**
  * @param {ERef<ZoeService>} zoe
  * @param {ERef<FeeMintAccess>} feeMintAccessP
- * @param {*} faucetInstallation
- * @param {*} stableInitialLiquidity
+ * @param {any} faucetInstallation
+ * @param {any} stableInitialLiquidity
```
dckc, line 16: rewriting `*` as `any`... interesting.

packages/inter-protocol/scripts/manual-price-feed.js
dckc: ah. fixed in next commit

packages/vats/test/bootstrapTests/test-vaults-upgrade.js
```diff
@@ -21,8 +21,8 @@ const collateralBrandKey = 'ATOM';
 /**
  * @param {import('ava').ExecutionContext} t
  * @param {object} [options]
- * @param {number} [options.incarnation=1]
- * @param {boolean} [options.logTiming=true]
+ * @param {number} [options.incarnation=1] Default is `1`
+ * @param {boolean} [options.logTiming=true] Default is `true`
```
dckc: it writes out defaults? interesting

dckc (commented):
The style is generally nicer.

My eyes glazed over about half way thru, so I can't vouch for all the details.

I have a vague unease about changing our bundles for non-functional reasons, but I suppose I can go with "the future is longer than the past".

packages/inter-protocol/src/auction/auctionBook.js
```diff
@@ -88,17 +92,19 @@ export const makeOfferSpecShape = (bidBrand, collateralBrand) => {
...
+ * @property {Amount<'nat'> | null} startProceedsGoal The proceeds the sellers
+ *   were targeting to raise
+ * @property {Amount<'nat'> | null} remainingProceedsGoal The remainder of the
+ *   proceeds the sellers were targeting to raise
+ * @property {Amount<'nat'> | undefined} proceedsRaised The proceeds raised so
+ *   far in the auction
+ * @property {Amount<'nat'>} startCollateral How much collateral was available
+ *   for sale at the start. (If more is deposited later, it'll be added in.)
+ * @property {Amount<'nat'> | null} collateralAvailable The amount of collateral
+ *   remaining
```
Chris-Hibbert, line 107: Is there another way to get whitespace back here?  It very effectively pointed out the parallels in the types, which is now harder to see. (And harder to tell that it was considered salient.)

packages/inter-protocol/src/vaultFactory/vault.js
```diff
@@ -22,31 +22,34 @@ const trace = makeTracer('Vault', true);
...
- * TRANSFER     - vault is able to be transferred (payments and debits frozen until it has a new owner)
- * CLOSED       - vault was closed by the user and all assets have been paid out
- * LIQUIDATED   - vault was closed by the manager, with remaining assets paid to owner
+ * ACTIVE - vault is in use and can be changed LIQUIDATING - vault is being
+ * liquidated by the vault manager, and cannot be changed by the user. If
+ * liquidation fails, vaults may remain in this state. An upgrade to the
+ * contract might be able to recover them. TRANSFER - vault is able to be
+ * transferred (payments and debits frozen until it has a new owner) CLOSED -
+ * vault was closed by the user and all assets have been paid out LIQUIDATED -
+ * vault was closed by the manager, with remaining assets paid to owner
```
Chris-Hibbert: Formatting destroyed here. Does this need leading bullets to maintain the list nature?

packages/vats/test/bootstrapTests/test-vaults-upgrade.js
turadg: I hadn't noticed that. Seems intrusive and redundant to me. I'd expect the doc generator to do that, using the actual default value.

This may be moot with the `no-default-params` lint rule we have. Either way I'll see if I can stop this.

packages/inter-protocol/src/vaultFactory/vault.js
turadg: yeah leading bullets would do it.

packages/inter-protocol/src/auction/auctionBook.js
turadg: the tool doesn't do vertical alignment. what parallel is lost? 

Note that when this is rendered into HTML it will look like the following. (using default theme)
Screenshot 2023-07-18 at 11 36 25 AM

Chris-Hibbert: When I was looking through the changes, it became clear to me that the reformatter thinks that the only version of the comments that matters (not the **primary** version, the **only** version) is what gets rendered for readers. I find the comments themselves useful at times to the maintainer (me).

The original formatting (which is what I look at to maintain it) directs my attention to the fact that there is a semantic difference between values which are `| null`, `| undefined`, and unadorned. The laid-out version makes that hard to see.

```
 * @typedef {object} BookDataNotification
 *
 * @property {Ratio         | null}      startPrice identifies the priceAuthority and price
 * @property {Ratio         | null}      currentPriceLevel the price at the current auction tier
 * @property {Amount<'nat'> | null}      startProceedsGoal The proceeds the sellers were targeting to raise
 * @property {Amount<'nat'> | null}      remainingProceedsGoal The remainder of
 *     the proceeds the sellers were targeting to raise
 * @property {Amount<'nat'> | undefined} proceedsRaised The proceeds raised so far in the auction
 * @property {Amount<'nat'>}             startCollateral How much collateral was
 *    available for sale at the start. (If more is deposited later, it'll be
 *    added in.)
 * @property {Amount<'nat'> | null}      collateralAvailable The amount of collateral remaining
 * 
 ```

turadg: ok I understand. There no option to retain the whitespace and I don't expect there will be since one key value of the tool is normalizing whitespace. It doesn't know when it's intentional or accidental. The way to maintain whitespace is to make it a codeblock, but that can't happen within the typedef.

If this is a show-stopped for you, let's discuss higher bandwidth

packages/run-protocol/src/vaultFactory/vault.js
turadg: I lost this change. I could restore it if necessary

Chris-Hibbert: Looks fine now.

packages/inter-protocol/src/auction/auctionBook.js
Chris-Hibbert: > If this is a show-stopped for you, let's discuss higher bandwidth

Not a show-stopper. On the whole, this is an improvement.

IMO, it may not be enough of an improvement to outweigh the control that it gives up, but I seem to be a minority in that viewpoint. I'm okay with that.

dckc (approved):
> On the whole, this is an improvement.

concur.

PR #4909, opened 2022-03-24
WIP: Associate each payment with a home purse

erights:
As an experiment, and as a record of where we would need to look, this PR renames the four payment-to-payment linear operations in zoe, adding a first purse argument to each. The idea is that each live payment could be tied to a home purse, to which it should return in an emergency if it is still live.

***Not to be reviewed***. It is an artifact of code-owners that reviewers were added to a Draft PR.


erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/5008

PR #4912, opened 2022-03-24
Standardized on one LICENSE file, added LICENSE where it was missing,…

### LINKED ISSUE #4910: Correctly apply the Apache 2.0 license to our code
The Apache Foundations documents how to apply their license [here](https://www.apache.org/foundation/license-faq.html#Apply-My-Software).

Apache suggests we have their stock license in a LICENSE file, as well as a NOTICE file which declares our copyright and acknowledges the copyright of any contained software developed by other people or organizations.

I did a survey of how Apache and other licenses are applied in some popular open source projects, results here:

- Apache Foundation 
  - Their example NOTICE file https://www.apache.org/licenses/example-NOTICE.txt includes notes
  - that some of the code was developed at the NCSA, and some code derived from RSA.  We could model this.

- Apache Lucene
  - Apache 2.0 license
  - Has LICENSE.txt and NOTICE.txt at top level only.
  - https://github.com/apache/lucene/blob/main/NOTICE.txt specifies a copyright, and lists other software it depends on.

- Cosmos SDK
  - Apache 2.0 license
  - Has the stock LICENSE file: https://github.com/cosmos/cosmos-sdk/blob/master/LICENSE
  - LICENSE is only at the top level.
  - No copyright notice, at top level or in code.

- Hyperledger / Fabric
  - Apache 2.0 license
  - Has a a stock LICENSE file: https://github.com/hyperledger/fabric/blob/main/LICENSE
  - Short NOTICE file with coyright: https://github.com/hyperledger/fabric/blob/main/NOTICE
  - Those files are only at the top level.
  - Code files contain a copyright notice (Greg Haskin and IBM)

- Uniswap / interface
  - GPL 3
  - https://github.com/Uniswap/interface/blob/main/LICENSE
  - At top level only.  No copyright notice, not even in code.

- React
  - MIT license
  - https://github.com/facebook/react/blob/main/LICENSE specifies FB copyright
  - License is at the top leve only.
  - No NOTICE file.
  - Code contains FB copyright notice.

- TensorFLow
  - Apache 2.0 license
  - Customized LICENSE file https://github.com/tensorflow/tensorflow/blob/master/LICENSE includes a COPYRIGHT section that gives credit to UCB for TF being derived from Caffe.
  - No NOTICE file, however, or other top level copyright notice.
  - Code however, does have copyright block comments ("Copyright 2015 The TensorFlow Authors. All Rights Reserved.").

- Django
  - Not a standard license.
  - LICENSE file specifies copyright by the foundation and redistribution rights.
  - https://github.com/django/django/blob/main/LICENSE

My plan is to get rid of the LICENSE files in our sub directories, and add LICENSE and NOTICE files at the top level.  The NOTICE file will include our copyright, as well as calling out the copyright on the few pieces of code in this repo that come from other sources.

Tartuffo:
closes: #4910


golang/cosmos/NOTICE
```diff
@@ -0,0 +1,5 @@
+Agoric SDK
+Copyright 2018-2022 Agoric Systems Operating Company
+
+This product includes software developed at
+Agoric Systems Operating Company (https://agoric.com/).
```
dtribble: discuss

packages/xsnap/NOTICE
```diff
@@ -0,0 +1,5 @@
+Agoric SDK
+Copyright 2018-2022 Agoric Systems Operating Company
+
+This product includes software developed at
+Agoric Systems Operating Company (https://agoric.com/).
```
dtribble: Discuss

FUDCo:
I'm presuming that this license file is the same in its N replicated copies, and that we don't actually have to review to ensure that?

dckc (commented):
I defer to others on this.

golang/cosmos/NOTICE
turadg: Why do we need a `NOTICE` file? [Apache guidance](https://infra.apache.org/licensing-howto.html) says,
> The `NOTICE` file … is required by [ASF policy](https://www.apache.org/legal/src-headers.html#notice).

We're using APL not part of ASF.

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
turadg: this should have specify the year and copyright owner.

turadg (dismissed):
Proposing that we remove `NOTICE` files, not required by Apache License.

Requesting that there be a script uses to automatically put `LICENSE` files where they need to be. For example https://github.com/babel/babel/blob/main/scripts/clone-license.sh

Blocking on specifying the year and copyright holder.

kriskowal:
In the Endo repository, I created a `skel` directory and some `scripts` that use that to make and update packages.

LICENSE
kriskowal: Also, when we bulk update licenses, we should be keeping the oldest copyright year, not the current year.

Tartuffo: Many projects (Cosmos SDK, Hyperledger / Fabric, Apache Camel) use exactly byte-for-byte the LICENSE file found here: https://www.apache.org/licenses/LICENSE-2.0.txt including the "APPENDIX: How to apply the Apache License to your work." which has that "to be filled in" Copyright notice.  The point is that APPENDIX comes after "END OF TERMS AND CONDITIONS", and your actual copyright notice(s) are supposed to be in the NOTICES file.

I'm inclined to leave it byte-for-byte exactly what is on the apache.org website.

Tartuffo: @turadg re: "Proposing that we remove NOTICE files, not required by Apache License." the way the license is structured, I don't see anything on the Apache website that says that the copyright should go into the LICENSE file.  That APPENDIX certainly does not.  

One *could* put it there in the LICENSE file, and some projects do.  However, my gut is to leave LICENSE file as written and vetted by their lawyers.

The NOTICE file is seems like a pretty logical place to put the copyright, as Apache puts it there for their projects, and other projects have followed suit.

We could also assert the Copyright and license in our README.md.

Tartuffo: @kriskowal In my survey documented in #4910, I did not see any projects that had LICENSE (or NOTICE) below the top level, so I don't see copying it down the tree and then having to keep it up to date.

golang/cosmos/NOTICE
Tartuffo: AFAICT, 99% of code in `golang/` is our code that uses the Cosmos SDK, not code copied in from Cosmos SDK.  (@michaelfig and @JimLarson please confirm)

The exception is `golang/cosmos/third_party/proto` which I'll give credit to "The GoGo Authors" and "Google Inc." in the NOTICE file.

LICENSE
turadg: Oh, I see. This part of LICENSE is just instructions on how to use it. 👍 removing by block.

The actual copyright holder is specified in NOTICE. This is unlike MIT License in which both are in one file:
https://github.com/facebook/react/blob/main/LICENSE

Fwiw here's an example of a LICENSE that roles up the copyright portion redundant with its NOTICE: https://github.com/moby/moby/blob/master/LICENSE
https://github.com/moby/moby/blob/master/NOTICE

Tartuffo:
I got rid of all of the extraneous sub-directory LICENSE and NOTICE files, and in our NOTICE file I added copyright and license assertions for all applicable code I could find in this repo.

NOTICE
```diff
@@ -0,0 +1,21 @@
+Agoric SDK
+Copyright 2018-2022 Agoric Systems Operating Company.
+
+This product includes software developed at
+Agoric Systems Operating Company (https://agoric.com/).
+
+The code in golang/cosmos/third_party/proto/google/protobuf is licensed
+under a BSD style license, Copyright 2008 Google Inc.
```
dtribble, line 8: "style" seems a strange word here. "based" perhaps?

NOTICE
```diff
@@ -0,0 +1,21 @@
+Agoric SDK
+Copyright 2018-2022 Agoric Systems Operating Company.
+
+This product includes software developed at
```
dtribble, line 4: Does this need to mention our license?  or is that in the LICENSE file?

dtribble (approved):
I think t's fine. I had a few questions/clarification, but feel free to proceed!

NOTICE
Tartuffo: That is the recommendation from Apache on how to apply their license, and born out by examples sited on the ticket.

NOTICE
Tartuffo: I've seen the term "style" used in other BSD covered license notices.  See also the use of style in https://linuxreviews.org/BSD_licenses.  I don't know why.

PR #4913, opened 2022-03-24
refactor(run-protocol): innerVault as virtual object

### LINKED ISSUE #4345: Use Virtual Objects in Vaults

## Description of the Design

## Security Considerations

## Test Plan

Presumably extensive.

turadg:
Part of #4345 

## Description

Converts the `makeInnerVault` function to use the `defineKind` function of the Collections API so that vaults can be marshaled and eventually made durable.

### Security Considerations

This separates what was one function,`args => instance` into two with an intermediate marshallable state, `args => state` and `state => instance`. This makes the instance very dependent on the assumption that the state passed into the latter is what was returned by the former.

--

### Testing Considerations

The virtualObjectManager does to marshalling and unmarshalling. The two functions are both at module scope so share no closing state. I _think_ this is sufficient to ensure they can be unmarshalled elsewhere but reviewers please advise.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -13,7 +13,7 @@ import {
 
 import { assert } from '@agoric/assert';
 import { AmountMath } from '@agoric/ertp';
-import { Far } from '@endo/marshal';
+import { defineKind } from '@agoric/vat-data';
 import { makeTracer } from '../makeTracer.js';
 import { calculateCurrentDebt, reverseInterest } from '../interest-math.js';
 import { makeVaultKit } from './vaultKit.js';
@@ -117,7 +117,7 @@ const validTransitions = {
  * @param {ZCFMint} mint
  * @param {ERef<PriceAuthority>} priceAuthority
  */
-export const makeInnerVault = (
+const initState = (
   zcf,
   manager,
   assetNotifier,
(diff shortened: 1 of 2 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -157,8 +149,24 @@ export const makeInnerVault = (
 
     // Two values from the same moment
     interestSnapshot: manager.getCompoundedInterest(),
-    debtSnapshot: emptyDebt,
+    debtSnapshot: AmountMath.makeEmpty(mint.getIssuerRecord().brand),
   };
+  return harden(state);
```
erights: How about changing like 133 to
```js
return harden({
```
or
```js
const state = harden({
  ...
});
return state;
```
?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -157,8 +149,24 @@ export const makeInnerVault = (
...
+};
+
+/** @param {ImmutableState & MutableState} state */
+const constructFromState = state => {
+  /** @type {ImmutableState} */
+  const { idInManager, manager, mint, priceAuthority, zcf } = state;
```
erights, line 155: It is a shame that in the upcoming gc-tolerant API we'll loose the ability to destructure this once for all the methods that need it. I like what your doing with the split type and wish we could keep it.

erights (approved):
LGTM, thanks!

PR #4919, opened 2022-03-25
Endow a `lookup(...path)` function to REPL and deploy scripts

### LINKED ISSUE #4918: Companion `lookup(...path)` function for deploy scripts and REPL

Our deploy scripts and REPL need to find objects by naming paths.  To make this easier, I'd like to add a `lookup(...path)` function to have an integrated namespace for traversal.

In doing so, deployment script parameters can specify `lookup` naming paths to find objects, without changes to code.

## Description of the Design

As one of the "deploymentPowers" arguments and also in the REPL as a `lookup` endowment, we would have:

`lookup('agoricNames', 'brand', 'RUN')` - lookup by agoricNames
`lookup('board', 'board0...')` - lookup by board ID
`lookup('namesByAddress', 'agoric1...', 'depositFacet')` - lookup via published names by address
`lookup('wallet', 'brand', 'LINK')` - lookup brand via wallet
`lookup('wallet', 'offerResult', 345)` - lookup offer result identified in wallet

## Security Considerations

This is a powerful endowment and should not be shared.  However, it's no more powerful than the existing capabilities granted to deploy scripts and the REPL.

## Test Plan

Manual testing for now.

michaelfig:
closes: #4918

## Description

This feature exposes several different naming paths via a powerful `lookup(...)` function.   See the attached issue for architecture.

### Security Considerations

No more powers granted besides what was already given to the REPL and deploy scripts.

Backward compatible.

Intended to be first used in the oracle deploy scripts.  I expect this will be documented when more deploy scripts use the mechanism.

### DIFF
--- a/packages/vats/src/types.js
+++ b/packages/vats/src/types.js
@@ -8,9 +8,10 @@
 /**
  * @typedef {Object} Board
  * @property {(id: string) => any} getValue
- * @property {(value: any) => string} getId
- * @property {(value: any) => boolean} has
+ * @property {(value: unknown) => string} getId
+ * @property {(value: unknown) => boolean} has
  * @property {() => string[]} ids
+ * @property {(...path: string[]) => Promise<unknown>} lookup
  */
 
 /**
--- a/packages/agoric-cli/src/deploy.js
+++ b/packages/agoric-cli/src/deploy.js
@@ -265,6 +265,38 @@ export { bootPlugin } from ${JSON.stringify(absPath)};
                 bundleSource(pathResolve(file), options),
               pathResolve,
               installUnsafePlugin,
+              /**
+               * Recursively look up names in the context of the bootstrap
+               * promise, such as:
+               *
+               * ['agoricNames', 'oracleBrand', 'USD']
+               * ['namesByAddress']
+               * ['namesByAddress', 'agoric1...']
+               * ['namesByAddress', 'agoric1...', 'depositFacet']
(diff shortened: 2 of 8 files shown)


packages/vats/src/types.js
```diff
@@ -11,6 +11,7 @@
  * @property {(value: any) => string} getId
  * @property {(value: any) => boolean} has
  * @property {() => string[]} ids
+ * @property {(...path: string[]) => Promise<any>} lookup
```
dckc: Add a justification for `any`, please. Or change to `unknown`.

dckc (commented):
This looks mostly harmless, aside from some documentation burden.

I don't understand the motivation well enough to approve it, though.

dckc (approved):
found the motivation:

> Intended to be first used in the oracle deploy scripts.

packages/vats/src/types.js
dckc: by way of example: https://github.com/Agoric/agoric-sdk/blob/master/packages/vats/src/core/types.js#L109-L113

```
 * @property {() => ERef<Record<string, any>>} getChainBundle Required for ag-solo, but deprecated in favour of getConfiguration
 *   NOTE: we use `any` rather than `unknown` because each client that wants to call a method such as
 *   `E(userBundle.bank).deposit(payment)` has to cast userBundle.bank;
 *   ideally, the cast is to some useful type. But unknown can't be cast directly to some other type;
 *   it has to be cast to any first.
```

Issue #1691, opened 2020-09-04
vat upgrade: the retroactive time-travelling manchurian candidate sleeper agent protocol


warner:
**What is the Problem Being Solved?**

One blocker for mainnet is confidence that we can upgrade today's vats to tomorrow's code. This requirement is made more exciting by:

* chains: the code must behave consistently and deterministically on all replicas
* schema-less orthogonal persistence: when state is kept in JavaScript Maps/WeakMaps instead of a database, we can't just write a new program and give it the old data

If we can plan ahead far enough to keep all necessary data in a database-like table, with a well-defined schema, then we've got more options (#3062), but we might not pull that off (we might have to upgrade vats which don't do that).

So the question is: given a vat which didn't plan ahead for upgrade, which has several months of history and accumulated state on the heap (intricately entwined with the objects that provide its behavior), how do we replace the code with something that behaves better, while preserving the state?

## Cutover

Since our chain's history is fixed, whatever changes we make to upgrade version 1 to version 2 cannot take visible effect until some activating event. This cutover will be triggered by some new message (probably to the vat's root object) which has never been seen before.

The history up until the cutover point is consistent with the behavior of version 1. Any vat running version 1 and receiving the recorded sequence of input messages will produce the recorded sequence of output messages.

The history up to that point is *also* consistent with the behavior of a version 2 vat that has not yet seen the trigger message. From an outside observer's point of view (seeing only the output messages), the version 2 code behaves just like version 1 did. It might be doing something different internally, like building up data structures that will ease the cutover, but none of the external messages will reveal that fact. Then, when the trigger is received, the vat is free to behave in some new way.

So our upgrade tool is to replace the original vat's code with version 2, which has been carefully designed to mimic version 1 until the trigger message is received. Then we replay the entire transcript to bring the vat back up to the current block. Eventually (we can wait as long as we want) we send in the trigger event, and cutover happens.

## Shadow State

The version 2 vat code might construct a parallel set of objects to prepare for the cutover. These objects weren't in the version 1 code because they weren't necessary for its main functionality.

[...]

PR #4927, opened 2022-03-26
feat(swingset): get a small upgrade to work

warner:
This enables a partial vat upgrade test to work. It makes a lot of
assumptions and is missing a lot of functionality, but it's a concrete
step forward. Limitations include:

* if the stopVat or startVat fail somehow, the vat will be left in an
unusable state, and the caller will not be notified
* stopVat does not clean up the non-durable objects
* stopVat does not reject the lingering promises
* startVat does not assert that userspace reconnects all Kinds
* nothing about reconnecting Kinds is tested at all
* nothing about remaining durable objects is tested at all
* the kernel does not unsubscribe the vat from remaining promises
* transcript pointers are cleared, but the contents are not deleted
* the snapshot decref/deletion is not tested
* stopVat is metered, but probably shouldn't be
* startVat metering needs more thought

refs #1848

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -640,33 +640,68 @@ export default function buildKernel(
    */
   async function processUpgradeVat(message) {
     assert(vatAdminRootKref, `initializeKernel did not set vatAdminRootKref`);
-    // const { bundleID } = message;
-    const { vatID, upgradeID, vatParameters } = message;
+    const { vatID, upgradeID, bundleID, vatParameters } = message;
     insistCapData(vatParameters);
 
     // eslint-disable-next-line no-use-before-define
     assert(vatWarehouse.lookup(vatID));
+    const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
     /** @type { import('../types-external.js').KernelDeliveryStopVat } */
-    const kd = harden(['stopVat']);
+    const kd1 = harden(['stopVat']);
     // eslint-disable-next-line no-use-before-define
-    const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
-    const status = await deliverAndLogToVat(vatID, kd, vd);
+    const vd1 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd1);
+    const status1 = await deliverAndLogToVat(vatID, kd1, vd1);
+    if (status1.terminate) {
+      // TODO: if stopVat fails, stop now, arrange for everything to
+      // be unwound. TODO: we need to notify caller about the failure
+      console.log(`-- upgrade-vat stopVat failed: ${status1.terminate}`);
+    }
 
-    // TODO: if status.terminate then abort the crank, discard the
-    // upgrade event, and arrange to use vatUpgradeCallback to inform
-    // the caller
+    // stop the worker, delete the transcript and any snapshot
+    // eslint-disable-next-line no-use-before-define
+    await vatWarehouse.destroyWorker(vatID);
+    const source = { bundleID };
+    const { options } = vatKeeper.getSourceAndOptions();
+    vatKeeper.setSourceAndOptions(source, options);
+    // TODO: decref the bundleID once setSourceAndOptions increfs it
 
-    // for now, all attempts to upgrade will fail
+    // pause, take a deep breath, appreciate this moment of silence
+    // between the old and the new. this moment will never come again.
 
-    // TODO: decref the bundleID and vatParameters.slots
-    const args = {
-      body: JSON.stringify([upgradeID, false, { error: `not implemented` }]),
-      slots: [],
-    };
-    queueToKref(vatAdminRootKref, 'vatUpgradeCallback', args, 'logFailure');
-    // if stopVat fails, we want everything to be unwound. TODO: we
-    // need to notify caller about the failure
-    return { ...status, discardFailedDelivery: true };
+    // deliver a startVat with the new vatParameters
+    /** @type { import('../types-external.js').KernelDeliveryStartVat } */
+    const kd2 = harden(['startVat', vatParameters]);
+    // eslint-disable-next-line no-use-before-define
+    const vd2 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd2);
+    // decref vatParameters now that translation did incref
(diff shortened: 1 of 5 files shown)


warner:
@FUDCo there are a lot of TODOs in these changes, but they should all be in code that's only exercised when a vat is upgraded. I'm thinking that it's better to land incremental functionality than to wait until everything is polished, so we can start playing with it sooner.

The next step will be to have `stopVat` reject any lingering promises, and add tests of the same.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -640,33 +640,68 @@ export default function buildKernel(
...
+    // deliver a startVat with the new vatParameters
+    /** @type { import('../types-external.js').KernelDeliveryStartVat } */
+    const kd2 = harden(['startVat', vatParameters]);
+    // eslint-disable-next-line no-use-before-define
+    const vd2 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd2);
+    // decref vatParameters now that translation did incref
+    for (const kref of vatParameters.slots) {
+      kernelKeeper.decrementRefCount(kref, 'upgrade-vat-event');
+    }
+    const status2 = await deliverAndLogToVat(vatID, kd2, vd2);
```
FUDCo, line 680: I'm a little nervous about doing two `deliverAndLogToVat` calls within a single crank.  While there's nothing specifically I can point at that seems wrong per se, up until now cranks and deliveries have been in 1:1 correspondence. In particular, we should be extra careful that there's nothing in `deliverAndLogToVat` that bakes in assumptions that it is at the start of a crank on the way in or at the end of a crank on the way out.

packages/SwingSet/src/kernel/state/vatKeeper.js
```diff
@@ -567,6 +567,27 @@ export function makeVatKeeper(
...
+          // TODO: if we roll back (because the upgrade failed), we must
+          // not really delete the snapshot
+          snapStore.prepareToDelete(snapshotID);
+        }
+        kvStore.delete(skey);
+      }
+    }
+    // TODO: same rollback concern
+    // TODO: streamStore.deleteStream(transcriptStream);
+    const newStart = streamStore.STREAM_START;
```
FUDCo, line 587: Do we really want to reset the stream pointer to the start position or merely remove the prior transcript entries?  Seems like for debugging purposes it would be nicer to maintain the continuity of transcript position across the upgrade, so that, for example, if one was looking at a historical record of a transcript there would be no ambiguity about which incarnation of the vat code it was associated with.

FUDCo (approved):
Approving because I don't want to block your progress, but a couple of concerns in the comments below.

packages/SwingSet/src/kernel/kernel.js
warner: Yep, this is the first time we're taking advantage of the difference between "crank" and "delivery". I think `vatWarehouse.deliverToVat` and the slogger is entirely ready for it, but the part that I know needs some polish is on the delivery results. The only sort of errors we've dealt with so far are all the kinds that terminate the vat. We're now entering into territory where e.g. `startVat` will soon be able to notice that userspace failed to reconnect all of the durable Kinds, and if that happens we want to roll back the entire upgrade. We'll need a response that says "hey caller, the worker is wedged and no longer viable, but if you `abortCrank` and jettison the worker, you could resume from the earlier snapshot without problems". My plan is to get roll-back-failed-upgrade working after getting the successful upgrade paths done, at which point I'll be looking more closely at these delivery results and their error modes.

packages/SwingSet/src/kernel/state/vatKeeper.js
warner: Yeah, I was wondering about that too. The conclusion I came to was that the transcript is intimately tied to the bundleID being executed: it doesn't make sense to retain a transcript without also retaining the bundle. It's not like you can replay the concatenated transcripts on top of only the original source bundle and get to the current state.

To rebuild the vat from nothing requires a data structure that looks like "load bundle1, execute these N1 transcript entries, then shutdown and load bundle2, then execute these other N2 entries, etc". To do that properly we'd need to put the bundleID in the `startVat` delivery, and record the `stopVat` deliveries too. And you'd need to retain the source bundles for all the bundleIDs in all the `startVat`s in the extended transcript.

That's not a bad design, but I haven't stopped to think about how long it would take to make the necessary changes to achieve it. I'll spend a bit evaluating that before I land this PR. At the very least I'll sketch out what the complete design would be so I can compare.

Performance-wise, I was hoping to delete as much data as possible, and being able to delete the transcript and the old bundles is appealing. A new validator who's trying to catch up doesn't need to replay the transcripts from bundle-1: it's sufficient for them to replay from just the most recent `startVat`. That helps their catchup performance a lot, especially if we manage periodic "null upgrades" whose main job is to discard the transcripts. But of course we don't have to delete the historical transcripts to achieve this, as long as new validators can merkle-root-validate the most recent transcript segment without needing to fetch all of the earlier ones too (i.e. we need to be judicious with our merkle tree layout).

FUDCo: Deleting the old transcript sounds fine to me.  I was just thinking that having continuity of the crank numbers in the historical timeline would be helpful when debugging timelines that cross an update boundary.

warner: It looks like it's non-trivial to make this change now. We could change `startVat` to include the bundleID, but the manager would have to snoop that and rewrite the delivery to include the whole bundle (or, better, send a `setBundle` command with the bundle contents before sending the `deliver(startVat)` command). But given the way the supervisor is currently written, that's too late, because the supervisor needs the full bundle to define `buildVatNamespace` before passing it into `makeLiveSlots`. We'd have to change `makeLiveSlots` to *return* a `setBuildVatNamespace`, sort of like how it used to return a `setBuildRootObject`.

I'll open a new ticket to see if we can come back to this later. There's an open policy question as to whether we should retain the historical transcripts from the very beginning of time, or just enough to enable #1691 -style replay (which, to be clear, cannot benefit from replay beyond the most recent version, which is a significant downside of the null-upgrade step).

warner: I created #4940 to track this idea.

PR #4929, opened 2022-03-26
chroe(deps): upgrade lerna to 4.0.0

arirubinstein:
closes: #XXXX
refs: #XXXX


michaelfig:
We've upgraded on current master, so this is now stale.

PR #4931, opened 2022-03-26
feat(run-protocol): liquidate serially

### LINKED ISSUE #4750: Simplify liquidation queuing

> after resolving design issues in the rate limited liquidation, it seems like we are better off if the liquidation process just liquidates one vault at a time rather than a whole cohort. So that means we don't need to have a separate collection vaults that are liquidating as opposed to simply "the one I'm liquidating first"

## Description of the Design

Revert https://github.com/Agoric/agoric-sdk/pull/4715 to restore `liquidateAndRemove` promise. (No need to persist anything because if the vat crashes then when the vaultManager is restored, the first vault to be liquidated may be different.

## Security Considerations
Reduced power of the cohort state.

## Test Plan

No change in tests

turadg:
closes: #4750
refs: #XXXX

## Description

Restores `liquidateAndRemove` promise from #4715 but changes the high-water-mark case to liquidate serially instead of starting them all at once in parallel.

For the `liquidateAll` case it continues start them all at once in parallel, assuming that that would maximize the throughput and efficiency is not a concern for that case.

### Security Considerations

Reduces dependence on durable state.

--

### Testing Considerations

Existing unit tests.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -175,6 +175,7 @@ export const makeVaultManager = (
     const highestDebtRatio = prioritizedVaults.highestRatio();
     if (!highestDebtRatio) {
       // if there aren't any open vaults, we don't need an outstanding RFQ.
+      trace('no open vaults');
       return;
     }
 
@@ -198,6 +199,7 @@ export const makeVaultManager = (
         highestDebtRatio.denominator, // collateral
         triggerPoint,
       );
+      trace('updating level for outstandingQuote');
       return;
     }
 
@@ -205,15 +207,15 @@ export const makeVaultManager = (
       return;
     }
 
+    // There are two awaits in a row here. The first gets a mutableQuote object
+    // relatively quickly from the PriceAuthority. The second schedules a
+    // callback that may not fire until much later.
+    // Callers shouldn't expect a response from this function.
     outstandingQuote = await E(priceAuthority).mutableQuoteWhenLT(
       highestDebtRatio.denominator, // collateral
       triggerPoint,
     );
 
-    // There are two awaits in a row here. The first gets a mutableQuote object
-    // relatively quickly from the PriceAuthority. The second schedules a
-    // callback that may not fire until much later.
-    // Callers shouldn't expect a response from this function.
     const quote = await E(outstandingQuote).getPromise();
     // When we receive a quote, we liquidate all the vaults that don't have
     // sufficient collateral, (even if the trigger was set for a different
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -154,14 +154,18 @@ export const makeVaultManager = (
       });
   };
 
-  // When any Vault's debt ratio is higher than the current high-water level,
-  // call reschedulePriceCheck() to request a fresh notification from the
-  // priceAuthority. There will be extra outstanding requests since we can't
-  // cancel them. (https://github.com/Agoric/agoric-sdk/issues/2713). When the
-  // vault with the current highest debt ratio is removed or reduces its ratio,
-  // we won't reschedule the priceAuthority requests to reduce churn. Instead,
-  // when a priceQuote is received, we'll only reschedule if the high-water
-  // level when the request was made matches the current high-water level.
+  /**
+   * When any Vault's debt ratio is higher than the current high-water level,
+   * call `reschedulePriceCheck()` to request a fresh notification from the
+   * priceAuthority. There will be extra outstanding requests since we can't
+   * cancel them. (https://github.com/Agoric/agoric-sdk/issues/2713).
+   *
+   * When the vault with the current highest debt ratio is removed or reduces
+   * its ratio, we won't reschedule the priceAuthority requests to reduce churn.
+   * Instead, when a priceQuote is received, we'll only reschedule if the
+   * high-water level when the request was made matches the current high-water
+   * level.
+   */
   const reschedulePriceCheck = async () => {
     assert(prioritizedVaults);
     const highestDebtRatio = prioritizedVaults.highestRatio();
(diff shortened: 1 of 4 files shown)


packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -229,23 +216,30 @@ export const makeVaultManager = (
...
     outstandingQuote = undefined;
-    // Ensure all vaults complete
-    await executeLiquidation();
+
+    // Liquidate serially
+    for (const [key, vault] of prioritizedVaults.entriesPrioritizedGTE(
+      quoteRatioPlusMargin,
+    )) {
+      // eslint-disable-next-line no-await-in-loop -- We want each liquidation to happen in a separate turn
+      await liquidateAndRemove([key, vault]);
```
dtribble: This is actually exactly the scenario that the `await` rule is designed to prevent, so I don't think this is an allowable exception. If `entriesPrioritizedGTE` returns an empty iteration, then line 229 will execute in the same turn, whereas if it does not, it would execute in a different turn.

dtribble (changes requested):
It's great except for a critical `await` error

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2294,15 +2294,13 @@ test('mutable liquidity triggers and interest sensitivity', async t => {
...
-  aliceUpdate = await E(aliceNotifier).getUpdateSince();
+  aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
   t.deepEqual(aliceUpdate.value.debtSnapshot.debt, aliceRunDebtLevel);
+  t.is(aliceUpdate.value.vaultState, Phase.ACTIVE);
 
   await manualTimer.tick();
   // price levels changed and interest was charged.
 
-  // expect Alice to be liquidated because her collateral is too low.
-  aliceUpdate = await E(aliceNotifier).getUpdateSince();
```
turadg, line 2367: @Chris-Hibbert can you confirm these test changes are legit?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -136,54 +131,51 @@ export const makeVaultManager = (
...
+    return liquidate(
+      zcf,
+      vault,
+      debtMint.burnLosses,
+      liquidationStrategy,
+      collateralBrand,
+    )
+      .then(() => {
+        assert(prioritizedVaults);
+        prioritizedVaults?.removeVault(key);
```
Chris-Hibbert, line 152: Why does this need optional chaining when it has just asserted that `prioritizedVaults` is non-null?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: This update isn't used. It seems fine to drop it.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2311,8 +2309,10 @@ test('mutable liquidity triggers and interest sensitivity', async t => {
...
   }
   await waitForPromisesToSettle();
-  aliceUpdate = await E(aliceNotifier).getUpdateSince();
-  bobUpdate = await E(bobNotifier).getUpdateSince();
-  t.is(aliceUpdate.value.vaultState, Phase.ACTIVE);
+  bobUpdate = await E(bobNotifier).getUpdateSince(bobUpdate.updateCount);
   t.is(bobUpdate.value.vaultState, Phase.LIQUIDATED);
+
+  // No change for Alice
+  aliceUpdate = await E(aliceNotifier).getUpdateSince(); // can't use updateCount because there's no newer update
```
Chris-Hibbert, line 2379: `updateCount` is optional. You don't need to justify not including it. Not including it means "I want the current state, even if it hasn't changed."

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1787,14 +1787,14 @@ test('mutable liquidity triggers and interest', async t => {
...
-  aliceUpdate = await E(aliceNotifier).getUpdateSince();
+  aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
   t.deepEqual(aliceUpdate.value.debtSnapshot.debt, aliceRunDebtLevel);
 
   await manualTimer.tick();
   // price levels changed and interest was charged.
 
   // expect Alice to be liquidated because her collateral is too low.
-  aliceUpdate = await E(aliceNotifier).getUpdateSince();
+  aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
```
Chris-Hibbert, line 1798: I think adding `updateCount` to these calls is acceptable, though I'm not sure it's an improvement.  

Using `getUpdateSince` without an argument can be more robust in a test because the test won't wait in vain when something changes. It'll return the existing state, which should be tested.

Chris-Hibbert (commented):
> can you confirm these test changes are legit?

They shouldn't break the test. They change the test's behavior if the code being tested stops working, but they should reveal the same faults.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: I had to include it in the other places or the update would come back before the liquidation processor settled. There are more promise resolutions now with "liquidate one at a time".

I actually think we should make a habit of including the `updateCount` so that you have a guarantee that you're getting a new state. If you don't expect a new state then don't ask for it and if you do then the test should fail if it's not available.

I even think it might be worth a helper wrapper around this notifier/response pair so you could have,
```
const aliceAfterLiquidation = await alicesUpdates.next();
```
(it would do the `updateCount` stuff itself to ensure you're getting a new update)

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: I spoke to this above but for review hygiene: it's better to be more explicit about whether a new state is expected, especially due to the interleaving of async calls that are common in our code.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: probably a result of refactoring. corrected.

dtribble (approved):

PR #4933, opened 2022-03-27
fix(swingset): stopVat rejects lingering promises

warner:
Any promise that is decided by a vat but not resolved by the time
upgrade happens should be rejected by `stopVat()`. This tracks decided
promises (both exported promises and the `result` vpid of inbound
deliveries) in a new `deciderVPIDs` Set. We remove vpids from the Set
when we send `syscall.resolve`, and we send rejections for everything
remaining during `stopVat`.

`dispatch.notify` (specifically `notifyOnePromise`) was changed to
ignore the kernel when it notifies us about a promise that we don't
recognize. This can now happen when version-1 subscribed to a Promise
that didn't get resolved before the vat got upgraded to
version-2. Rather than figure out a way to unsubscribe from these
promises (there shouldn't be very many of them), we just tolerate the
late resolution notification.

Note that this introduces a strong requirement that the kernel does
not re-use imported vpids, which was previously merely a
recommendation. Vat upgrade does not reset `v$NN.p.nextID`, so this
should be ok.

refs #1848

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -294,6 +294,16 @@ function build(
   const disavowedPresences = new WeakSet();
   const disavowalError = harden(Error(`this Presence has been disavowed`));
 
+  // This tracks all promises that we decide (i.e. we are obligated to
+  // call syscall.resolve on them before the end of stopVat). We add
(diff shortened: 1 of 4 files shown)


FUDCo (approved):
は、私にはよく見えますよ。

Issue #4737, opened 2022-03-04
add negative test of nameless/non-method/remote-function call


warner:
## What is the Problem Being Solved?

https://github.com/Agoric/agoric-sdk/issues/61#issuecomment-1059459378 points out that swingset does not currently support cross-vat "nameless calls", like `E(obj)(args)` (intended to be used with `obj = Far('iface', args => blah)`). The exact failure mode is a little obscure: it rejects the result promise, but the error is triggered by `insistMessage` when it sees `message.method` is not a string (it is `undefined`).

We have a plan to support this, and it would probably be useful, but we haven't gotten around to it yet. We don't have a test that demonstrates that it fails (or what the failure looks like), nor do we have any documentation recommending for or against using the pattern.

The tricky part is that `HandledPromise` *does* have support for this, so it works locally, just not when crossing a vat boundary. That's not good: code which works in some configurations would fail in others.

We don't have time to implement #61 now, but it might be good to:
* move the assertion up into liveslot's `applyMethod` handler, and include the phrase "`E(obj)(args)`" in the error so users know what to look for
  * (the handler is called in a later turn, so we can't give them a stack trace to where they strayed beyond the implementation's abilities, and there's no method name to share with them, but describing the troublesome pattern might still help)
* write a unit test that attempts the call and asserts that the result promise is rejected, with the expected error message
* add a `console.log()` complaint when this happens too
  * in my test, where I didn't attach an error handler to the `p = E(obj)(args)`, the error seems to have been silently discarded
* and add a pointer to #61 so users who hit this can learn more

@michaelfig @erights do you think this limitation is serious enough to consider looking for a way to disable nameless *local* calls? I don't know how we would, but maybe we should look?

michaelfig:
> @michaelfig @erights do you think this limitation is serious enough to consider looking for a way to disable nameless _local_ calls? I don't know how we would, but maybe we should look?

No.  The semantics of `E(x)...` and `E.get(x)...` are that you need to handle rejected promises, or chain these calls together so that the rejected promise is propagated.  Dropping them on the floor is the caller's fault, because they can fail for arbitrary reasons.

The `unhandledPromiseRejection` callback should have caught this error.  I'm confused why it didn't.

michaelfig:
> The `unhandledPromiseRejection` callback should have caught this error. I'm confused why it didn't.

[...]

PR #4939, opened 2022-03-28
fix(SwingSet): report but don't crash on unhandled rejections

### LINKED ISSUE #4938: XS doesn't report unhandled rejections
> OIC, when running in XS we don't get such a callback.

Here is the code in `packages/xsnap/xsnap-native/xsnap/sources/xsnapPlatform.c`:

```c
	case XS_UNHANDLED_EXCEPTION_EXIT:
		fxReport(the, "unhandled exception: %s\n", fxToString(the, &mxException));
		mxException = mxUndefined;
		break;
	case XS_UNHANDLED_REJECTION_EXIT:
		fxReport(the, "unhandled rejection: %s\n", fxToString(the, &mxException));
		mxException = mxUndefined;
		break;
```

Both unhandled rejections and unhandled exceptions are swallowed up (`fxReport` is a no-op).  We should probably ensure they are reflected to the host as a console message, but that requires some rewiring we should do at an opportune time.

Also, I found unhandled rejections have an `mxException` of `undefined` before this switch statement, so they're hard to diagnose.  I reported that to the XS folks.

_Originally posted by @michaelfig in https://github.com/Agoric/agoric-sdk/issues/4737#issuecomment-1059881368_

michaelfig:
refs: agoric-labs/xsnap-pub#21
closes: #4938

## Description

After this patch and the referenced xsnap-pub PR, unhandled rejections are logged consistently by the kernel and don't crash AVA, for example.

Here is output from the local (Node.js) worker running `cd packages/SwingSet && yarn test test/workers/test-worker.js`:

```
UnhandledPromiseRejectionWarning: (Error#3)
Error#3: one unhandled rejection
  at Alleged: root.one (.../swingset-vat/test/workers/vat-target.js:74:20)
```

And the corresponding output from the XS worker:

```
UnhandledPromiseRejectionWarning: (Error#2)
Error#2: one unhandled rejection
Error: one unhandled rejection
 at apply ()
 at Error (/Users/michael/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:7593)
 at one (.../swingset-vat/test/workers/vat-target.js:74)
 at apply ()
 at apply ()
 at dispatchToHandler (/Users/michael/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:10951)
 at (/Users/michael/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:10589)
 at win (/Users/michael/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:11294)
 at ()
```

The XS worker still has some way to go before it's as legible as the local worker, but this PR at least makes sure the errors don't disappear.

### Security Considerations

Unhandled rejections are reported at the correct level: the kernel output and not under control of the vat.

### DIFF
--- a/packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
+++ b/packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
@@ -342,4 +342,10 @@ function makeWorker(port) {
 // xsnap provides issueCommand global
 const port = managerPort(globalThis.issueCommand);
 const worker = makeWorker(port);
+
+// Send unexpected console messages to the manager port.
+globalThis.print = (...args) => {
+  port.send(['console', 'error', ...args]);
+};
+
 globalThis.handleCommand = port.handlerFrom(worker.handleItem);
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -45,8 +45,13 @@ function makeConsole(tag) {
   return harden(cons);
 }
 
-/** @param {Error} e */
-function unhandledRejectionHandler(e) {
+/**
+ * @param {unknown} e
+ * @param {Promise} pr
+ */
+function unhandledRejectionHandler(e, pr) {
+  // Don't trigger sensitive hosts (like AVA).
+  pr.catch(() => {});
   console.error('UnhandledPromiseRejectionWarning:', e);
 }
 
--- a/packages/SwingSet/test/workers/vat-target.js
+++ b/packages/SwingSet/test/workers/vat-target.js
@@ -71,6 +71,7 @@ export function buildRootObject(vatPowers, vatParameters) {
   //   syscall.reject(pC, Error('oops'))
   //   syscall.fulfillToData(rp3, 'rp3 good')
   function one() {
+    Promise.reject(Error('one unhandled rejection'));
     precB.resolve(callbackObj); // syscall.fulfillToPresence
     precC.reject(Error('oops')); // syscall.reject
     return 'rp3 good';


packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
```diff
@@ -342,4 +342,10 @@ function makeWorker(port) {
 // xsnap provides issueCommand global
 const port = managerPort(globalThis.issueCommand);
 const worker = makeWorker(port);
+
+// Send unexpected console messages to the manager port.
+globalThis.print = (...args) => {
```
dckc, line 347: This makes my head hurt a little. I defer to @warner

michaelfig: > This makes my head hurt a little. I defer to @warner

Do you mean just these four lines need @warner's perusal, and otherwise you approve?  Or would you like @warner to review the whole PR?

dckc: I prefer that somebody else reviews the whole PR.

I'm not confident that I'm in sync with the discussion since your [March 5 comment](https://github.com/Agoric/agoric-sdk/issues/4737#issuecomment-1059834032):
> ISTR @erights or @kriskowal had a sketch for what ideal unhandled promise rejection tracking would look like. Can we suggest that to the XS folk?

Sending `globalThis.print` to the manager port specifically seems like something where @warner is quite a bit more swapped in than I am.

I'm open to a 15-ish minute call to get swapped in on all of the above.

FUDCo (commented):
This looks reasonable to me, but I'll also defer to @warner, as the interaction between the kernel and the workers is his bailiwick and even though this is a small change the subtleties here are subtle.

packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
kriskowal: I assume my sketch is in general irrelevant here, but for sake of completeness, my hope is to one day have a platform that provides hooks, possibly just to a debugger or orchestrator, to surface events “deferred” (promise creation), “resolved” (progress), “fulfilled”, and “rejected” such that a debugger can reveal, at any point in time, what promises are in an unhandled rejected state (possibly forever) or in a pending (possibly forever) state. The trouble with Node.js’s interpretation of an unhandled rejection is that it doesn’t account for the possibility that the promises is in a queue and will have a handler added in the future, but without a causal debugger with the feature I’m describing, rejected promises disappear too easily without a “trace”. Such a debugger would let you investigate both unterminated promise chains and live lock.

kriskowal: I’ll also point out that `p.catch(() => {})` defeats Node.js’s trick for revealing most stray promise chains, leaving you more in need of a causal debugger.

warner: Ok so before this, the start compartment's `globalThis.print` is defined by XS, and from what I remember it is the *only* form of IO made available to the start compartment. Our `packages/xsnap/lib/console-shim.js` uses it to build a workalike `console` object. The supervisor (in `makeLogMaker`) uses that `console` to build a `makeLog` that temporarily overwrites `globalThis.print` with a form (`portSendingPrinter`) that sends things to the parent process, so the functions you get through `makeLogMaker` will send to the parent process instead of stdout. We use `makeLogMaker` to create a `console` for vat code (including liveslots) to use.

I think that means the status quo is for all `console.log` messages within liveslots or the vat code to get sent to the kernel process, but any `console.log` messages within the supervisor (or `printf()` within xsnap.c) is written to the worker process' stdout.

Is that right?

And then this PR replaces the start compartment's `globalThis.print` with a form that sends to the parent process. I think that means any `console.log` messages within liveslots, vat code, or the supervisor (and probably anything else JS in the entire worker process) will send to the parent process, leaving at most `printf()` within xsnap.c going to stdout.

Is that right?

I'm guessing that the `console.log` statements I'm currently enjoying using in liveslots are actually getting written to the kernel process' stdout (by some helpful code on the kernel side), and not just being written by the worker process to the inherited (same) stdout FD?

If all that is true, then this PR shifts an extra chunk of logging from the worker's stdout to (the kernel port and then to both the kernel process' stdout and any other special destination), but most of the previous logging was already following that path, and this is just to grab the low-level unhandled-rejection messages coming out of the engine. That sounds fine to me.

If we were previously relying upon inherited stdout, then I'd want to think about this some more.

Separate question: does this interact correctly with the `makeLogMaker` code that temporarily swaps out `globalThis.print`? It feels like it might be redundant with that.

michaelfig: > If all that is true, then this PR shifts an extra chunk of logging from the worker's stdout to (the kernel port and then to both the kernel process' stdout and any other special destination), but most of the previous logging was already following that path, and this is just to grab the low-level unhandled-rejection messages coming out of the engine. 

That's all correct.

> That sounds fine to me.

Great, thanks!  Would you be able to approve accordingly?  I'm marking as automerge.

dckc (approved):
the comments by @warner address my concerns.

packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
warner: it landed already, but yep, given that answer, I'm r+

PR #4942, opened 2022-03-28
chore(swingset): improve virtual-object code and tests

warner:
Make some small non-functionality-changing improvements to the
virtual-object code and its tests.

collectionManager.js and virtualReferesnces.js add some internal
asserts, to catch things I stumbled across during development.

test-storeGC.js acquired a new test case which checks our assumption
that the `scalarMapStore` collection type will be alloced a specific
ID (`o+1`). This assumption is baked into the rest of test-storeGC.js,
and will change when liveslots is modified to perform more allocations
at startup. By testing our test's assumption, it should be easier to
safely update the test. collectionManager.js now exports
testHooks.obtainStoreKindID to support this test.

In test-virtualObjectManager.js I factored out the kindHandle
IDs. This should not change the behavior of the current test, it just
consolidates a number of hard-coded identifiers into a single
place. We'll be changing the way kindHandle allocation happens soon,
and it will be easier to review the corresponding test changes if they
only modify the one variable.

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -135,7 +135,10 @@ export function makeCollectionManager(
     keySchema = M.any(),
     valueSchema,
   ) {
-    const { hasWeakKeys, durable } = storeKindInfo[kindName];
+    assert.typeof(kindName, 'string');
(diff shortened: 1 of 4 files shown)


FUDCo (approved):
Bonum mihi videtur.

PR #4945, opened 2022-03-28
chore(swingset): update vatstore-usage docs with new example IDs

warner:
This updates the object IDs used as examples to match the values they
ought to have for real, now that initialization is done early.

It also discusses "baggage" with more detail, and emphasizes the
surprising difference between the way virtual kinds and collections
use the second component of the vref.

### DIFF
--- a/packages/SwingSet/src/liveslots/vatstore-usage.md
+++ b/packages/SwingSet/src/liveslots/vatstore-usage.md
@@ -15,6 +15,23 @@ The rest of the vat's keyspace is used by liveslots:
 
 This file describes the layout of the vatstore keyspace.
 
+
+# Counters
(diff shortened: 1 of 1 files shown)


packages/SwingSet/src/liveslots/vatstore-usage.md
```diff
@@ -84,15 +107,39 @@ Liveslots provides a handful of "virtual collection" types to vats, to store hig
 
 Each function accepts a `isDurable` argument, so there are currently 8 collection types.
```
FUDCo: a -> an

FUDCo (approved):
Found one old typo, otherwise G2G.

Issue #4861, opened 2022-03-18
Add a dedicated contractHelper to VaultFactory


Chris-Hibbert:
## What is the Problem Being Solved?

`contractHelper` does a bunch of work for the vast majority of contracts that have a single set of parameters and in the process hides some powerful capabilities from the body of the contract. `VaultFactory` can't use `contractHelper` since it has parameters that are declared once for `vaultFactory`, but then governed separated for each VaultManager (viz, each token type). 

## Description of the Design

We should add a special contractHelper just for the `VaultFactory`, even if it doesn't generalize for any other contract, just because it would take all those details and those powerful OCaps out of the body of `VaultFactory`.

## Security Considerations

improved clarity and cleanliness, even though there might not be any additional sharing.

## Test Plan

The existing tests are thorough.

PR #4948, opened 2022-03-29
feat(run-protocol): debtLimit governed param

### LINKED ISSUE #3501: Governance-controllable RUN debt-limit per collateral type in Vault Factory & getRUN contract
Allow governance to set and modify a cap to the amount of RUN that may be issued against a *specific* collateral type (vaults AND getRUN).

Attempts to mint additional RUN beyond the debt limit cap should fail
UIs should be able to show a user before attempting to make a transaction if it would fail for this reason

## Description of the Design
Parameter that gives a total limit - per collateral type - of the RUN that is available to be minted.  Any attempts to mint RUN which would go above the limit for that collateral type should fail.  

However, other functions on a vault which *do not* increase RUN should succeed (i.e., even if we have surpassed the debt limit on the collateral for my vault, I should be able to perform vault actions like closing vault or adding collateral, which do not increase the debt)

Limit needs to be adjustable by governance

turadg:
advances: #3501 (might close, depending on what's left for UI work)

## Description

Add a param to vault-manager governance: `debtLimit`.

If minting triggered by a vault operation would exceed the debt limit, fail. Minting for charging interest is always allowed.

### Security Considerations

New governed param. Malicious manipulation could halt minting.

Other packages calling `addVaultType` will need to include an initial value for `debtLimit`.

### Testing Considerations

The param and check are tested. I'm not sure how to test this requirement of the work item,
> UIs should be able to show a user before attempting to make a transaction if it would fail for this reason

Any advice @samsiegart ?

### DIFF
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -7,6 +7,7 @@ import '@agoric/governance/exported.js';
 import '@agoric/vats/exported.js';
 import '@agoric/vats/src/core/types.js';
 
+import { AmountMath } from '@agoric/ertp';
 import { makeGovernedTerms } from './vaultFactory/params.js';
 import { makeAmmTerms } from './vpool-xyk-amm/params.js';
 import { makeReserveTerms } from './reserve/params.js';
@@ -251,11 +252,17 @@ export const startVaultFactory = async (
 
   const centralBrand = await centralBrandP;
 
-  // declare governed params for the vaultFactory; addVaultType() sets actual rates
-  const rates = {
-    liquidationMargin: makeRatio(105n, centralBrand),
-    interestRate: makeRatio(250n, centralBrand, BASIS_POINTS),
-    loanFee: makeRatio(200n, centralBrand, BASIS_POINTS),
+  /**
+   * Types for the governed params for the vaultFactory; addVaultType() sets actual values
+   *
+   * @type {VaultManagerParamValues}
+   */
+  const vaultManagerParams = {
+    // XXX the values aren't used. May be addressed by https://github.com/Agoric/agoric-sdk/issues/4861
+    debtLimit: AmountMath.make(centralBrand, 0n),
+    liquidationMargin: makeRatio(0n, centralBrand),
+    interestRate: makeRatio(0n, centralBrand, BASIS_POINTS),
+    loanFee: makeRatio(0n, centralBrand, BASIS_POINTS),
   };
 
   const [ammInstance, electorateInstance, contractGovernorInstall] =
@@ -273,7 +280,7 @@ export const startVaultFactory = async (
     installations.liquidate,
     chainTimerService,
     invitationAmount,
-    rates,
+    vaultManagerParams,
     ammPublicFacet,
   );
   const governorTerms = harden({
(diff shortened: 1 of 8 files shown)


packages/run-protocol/src/econ-behaviors.js
```diff
@@ -250,8 +251,13 @@ export const startVaultFactory = async (
 
   const centralBrand = await centralBrandP;
 
+  // ??? why are there values here?
```
Chris-Hibbert: * vaultFactory is a weird case
* Other contracts require that the list of governed params be specified in `terms`, including initial values, since that provides a legible set of values to initialize from.
* We could build a vaultFactoryContractHelper (#4861), which could provide helper functions specific to vaultFactory, but so far that isn't a priority.
* Various helpers validate parameter governance and expect the format of parameters in `terms` to include values that may no longer be current.

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -22,19 +22,20 @@
...
  */
 
 /**
  * @callback AddVaultType
  * @param {Issuer} collateralIssuer
  * @param {Keyword} collateralKeyword
- * @param {Rates} rates
+ * @param {VaultManagerParamValues} params
  * @returns {Promise<VaultManager>}
  */
```
Chris-Hibbert, line 41: There's a `* TODO check limits.` on line 58 below that I think this PR resolves.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2130,6 +2131,69 @@ test('excessive loan', async t => {
   });
 });
 
+/**
+ * Each vaultManager manages one collateral type and has a governed parameter, `debtLimit`,
+ * that specifies a cap on the amount of debt the manager will allow.
+ *
+ * Attempts to adjust balances on vaultsbeyond the debt limit fail.
```
```suggestion
 * Attempts to adjust balances on vaults beyond the debt limit fail.
```

Chris-Hibbert (approved):
LGTM.   nice and simple

packages/run-protocol/src/econ-behaviors.js
turadg: Thanks. I updated the comment and zeroed out the values. Requesting another review.

Chris-Hibbert:
> Thanks. I updated the comment and zeroed out the values. Requesting another review.

There are two unaddressed comments remaining. (a typo and a possibly resolved TODO.)
Would you take another look?

turadg:
> There are two unaddressed comments remaining. (a typo and a possibly resolved TODO.)

Ah, missed the typo, thanks. The TODO was in [ce9011d](https://github.com/Agoric/agoric-sdk/pull/4948/commits/ce9011df6164a25932a1cf24665e256c273648c9) but I forgot to mention it.

I see some CI errors too so I'll try to fix those and force push again.

Chris-Hibbert (approved):

mhofman:
This PR broken the loadgen with error: `assertAmount: cannot coerce undefined to object`

mhofman:
It looks like the signature of `addVaultType` changed. This is a breaking change. Can we revert and find an alternative approach?

PR #4946, opened 2022-03-29
chore: drop an extraneous reference to `main`.

Chris-Hibbert:
refs: #4868
closes: #4818

## Description

Drop an extraneous reference to `main`.
It was in an ignored position.

### Security Considerations

type checking didn't notice it.

### Documentation Considerations

None

### Testing Considerations

The test didn't notice it, either, since it was an ignored parameter.

PR #4950, opened 2022-03-29
refactor(run-protocol): throw in getParamMgrRetriever.get for unsupported paramDesc

turadg:
## Description

Follow up to https://github.com/Agoric/agoric-sdk/pull/4946 so that it would have failed tests.


Chris-Hibbert (approved):
Thanks.

PR #4952, opened 2022-03-29
Eliminate the propertyNames Set

### LINKED ISSUE #4935: durable Kinds forget state property names across upgrade, are inconsistent when init() adds more
I'm working on getting #1848 upgrade working, and I've got a test that creates a durable Kind (and handle) in v1, creates an exports an instance, gets upgraded, the v2 code re-defines that kind (using the handle from baggage), then tries to invoke a method on a durable object. The `defineDurableKind` part works ok, but when I try to invoke the method, it fails because the `state` object is empty: no property getters/setters.

I think there are some significant problems in `defineKindInternal`. There's a `propertyNames` Set that lives in RAM, scoped to `defineKindInternal` (so one per Kind). The `Set` is populated by `makeNewInstance` as it walks the properties of the initial state (as returned by the user-provided `init` function). It looks like this `add` happens for every call to `init`. That means the properties we track could grow over time, if the user's `init` function doesn't return a consistent set of properties.

When we make a Representative, `makeRepresentative` uses this `propertyNames` Set to create the `wrappedState` object: it makes one getter/setter pair per property name in `propertyNames`. That means the wrapper will grow new properties over time based upon the last `init` that took place, unrelated to the properties that were stored for that particular instance. That's inconsistent at best, probably crashy if some behavior function tries to `get()` a property that some other instance invented but which has no state on *this* instance (it'll try to `m.unserialize(null)`, which isn't capdata), and feels like it might enable an ambient communication channel between unrelated instances (but one that probably needs the Kind's behavior functions to help, which doesn't count).

The problem that affects Durables is that when we start the vat back up in a new version, we need to be able to thaw out these instances in a heap that has never seen an `init`. We can't rely upon userspace to tell us what the property names are. Instead, we must either store the property names in the DB at some point, or we need to build getters/setters for each Representative based upon the state stored on that instance (which means we need to fetch the state during `makeRepresentative`, to at least read the `.keys()` on it).

Storing the property names in the DB makes the most sense, but it's a bit awkward that we don't learn what they are until the first `init` is called. But not impossibly awkward. The DB will have most of the description of the Kind as soon as `defineKind` is called, but the `.statePropertyNames` key will be empty until the first instance is created. Then the key will be populated, and all subsequent `init` calls are obligated to use the same names. We can hold the property names in RAM too, for performance during the remainder of that version.

When `defineKindInternal` is called, it needs to use the `kindID` to look up the stored list of property names. If present, it should be loaded into RAM, used to define the getters/setters, and imposed upon all `init` calls. If missing, the next `init` call should be used to populate it (and save to DB). In the missing case, we don't need the list for getters/setters because we know there aren't any instances yet.

One other option (I think we discussed this a while ago) might be to require userspace to declare the state property names they're going to use, when calling `makeKindHandle`. Then we could store the property names in the `kindDescriptor` DB key. That would be simple and elegant for the implementation, but maybe a bit clunky for userspace, which would see two lists of property names: one in `makeKindHandle`, the other in their `init` function.

OTOH, storing the list in the `kindDescriptor` would both force all instances to have the same schema, *and* preclude version-2 from augmenting the schema (which feels like a more significant restriction). I know @FUDCo has thought a lot about lazy/immediate DB upgrades (the usual "read v1, write v2" upgrade-when-touched approach), but we haven't really talked as a group about schema changes.

@FUDCo thoughts? @erights could we get away with requiring `makeKindHandle(tag, [ ...statePropertyNames ])`? I think I prefer the simplicity of it, and if we don't have an obvious way to handle new property names anyways, then we might has well be more strict now, and find a way to relax the rules later on.

FUDCo:
Eliminate the `propertyNames` Set used by the virtual object manager to track what properties are "live".  Instead, believe each instance with respect to which properties it actually has.

Closes #4935

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -498,7 +498,6 @@ export function makeVirtualObjectManager(
    */
   function defineKindInternal(kindID, tag, init, actualize, finish, durable) {
     let nextInstanceID = 1;
-    const propertyNames = new Set();
 
     function makeRepresentative(innerSelf, initializing, proForma) {
       if (!proForma) {
@@ -518,7 +517,10 @@ export function makeVirtualObjectManager(
       }
 
       const wrappedState = {};
-      for (const prop of propertyNames) {
+      if (!initializing) {
+        ensureState();
+      }
+      for (const prop of Object.getOwnPropertyNames(innerSelf.rawState)) {
         Object.defineProperty(wrappedState, prop, {
           get: () => {
             ensureState();
@@ -631,7 +633,6 @@ export function makeVirtualObjectManager(
         }
         data.slots.map(vrm.addReachableVref);
         rawState[prop] = data;
-        propertyNames.add(prop);
       }
       const innerSelf = { baseRef, rawState, repCount: 0 };
       const [toHold, toExpose, state] = makeRepresentative(
--- a/packages/SwingSet/test/virtualObjects/test-representatives.js
+++ b/packages/SwingSet/test/virtualObjects/test-representatives.js
@@ -295,6 +295,7 @@ test.serial('exercise cache', async t => {
   done();
 
   await hold(T8); // cache unchanged - [t1 t2 t3 t4]
+  ck('get', dataKey(8), thingVal('thing8'));
   ck('get', rcKey(4), undefined);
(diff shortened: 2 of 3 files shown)


warner (approved):
Looks good, thanks for the fast turnaround, I'll update my branches to use this.

warner:
Oh, please put a `refs #4935` in the commit comments

PR #4953, opened 2022-03-30
chore(run-protocol): toward PSM bootstrap integration

### LINKED ISSUE #5010: Integrate PSM with bootstrap

dckc:
goal:
closes #5010 

@dtribble This experiment in PSM bootstrap raises an issue:

 - [ ] Where do we get the anchor issuer?

see also:
 - #4786
 - #4663

### DIFF
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -8,6 +8,7 @@ import '@agoric/vats/exported.js';
 import '@agoric/vats/src/core/types.js';
 
 import { AmountMath } from '@agoric/ertp';
+import { CONTRACT_ELECTORATE, ParamTypes } from '@agoric/governance';
 import { makeGovernedTerms } from './vaultFactory/params.js';
 import { makeAmmTerms } from './vpool-xyk-amm/params.js';
 import { makeReserveTerms } from './reserve/params.js';
@@ -39,6 +40,8 @@ const CENTRAL_DENOM_NAME = 'urun';
  *   vaultFactoryCreator: VaultFactory,
  *   vaultFactoryGovernorCreator: GovernedContractFacetAccess<unknown>,
  *   vaultFactoryVoteCreator: unknown,
+ *   psmCreatorFacet: unknown,
+ *   psmGovernorCreatorFacet: GovernedContractFacetAccess<unknown>,
  * }>} EconomyBootstrapSpace
  */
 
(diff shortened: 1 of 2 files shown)


Chris-Hibbert:
The PSM only has [interim fake governance](https://github.com/Agoric/agoric-sdk/blob/52f60eb192217ff3e4cf84a5a2ff8ada19fb5dcc/packages/run-protocol/src/psm/psm.js#L77) at this point. I don't think it's worth trying to set up bootstrap before giving it actual governance. 

I also think the churn that got in the way of adding governance is done, and governance is stable enough that it makes sense to add governance to PSM now.

dckc:
> ... I don't think it's worth trying to set up bootstrap before giving it actual governance.

Perhaps some of the technical details aren't wired up quite right. But I found the exercise of wiring it up made me think through some policy issues.

Is it clear to you whence comes the `AUSD` issuer, @Chris-Hibbert ?

dckc:
Oops! I meant to make this a **draft** PR. Now I understand your comment better, @Chris-Hibbert .

Chris-Hibbert:
Misunderstandings all around. I also see that you manually asked for review from Dean, and the rest of us were added auto-mistakenly. I'll wait for an actual invitation to weigh in before reviewing further.

> Is it clear to you whence comes the AUSD issuer, @Chris-Hibbert ?

No.

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -649,3 +651,110 @@ export const startRunStake = async (
...
+    brand: {
+      consume: { AUSD: anchorBrandP, RUN: runBrandP },
+    },
+    issuer: {
+      consume: { AUSD: anchorIssuerP },
+    },
+  },
+  {
+    WantStableFeeBP = 1n,
+    GiveStableFeeBP = 3n,
```
dtribble, line 697: @btulloh @rowgraus what should our initial values be?

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -649,3 +651,110 @@ export const startRunStake = async (
...
+    feeMintAccessP,
+    runBrandP,
+    anchorBrandP,
+    anchorIssuerP,
+    contractGovernor,
+    E(zoe).install(bundle),
+    chainTimerService,
+  ]);
+
+  const poserInvitationP = E(
```
dtribble, line 719: This seems like it could be included in the Promise.all above

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -649,3 +651,110 @@ export const startRunStake = async (
...
+      consume: { AUSD: anchorBrandP, RUN: runBrandP },
+    },
+    issuer: {
+      consume: { AUSD: anchorIssuerP },
+    },
+  },
+  {
+    WantStableFeeBP = 1n,
+    GiveStableFeeBP = 3n,
+    MINT_LIMIT = 20_000_000n * 1_000_000n,
```
dtribble, line 698:
```
// The maximum number of tokens of the anchor brand allowed into the pool
```
This should remove the 1_000_000 and multiply by the `10**decimals` below when we get the brand.
This should be a higher number.  40M?

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -649,3 +651,110 @@ export const startRunStake = async (
...
+    },
+    [CONTRACT_ELECTORATE]: {
+      type: ParamTypes.INVITATION,
+      value: electorateInvitationAmount,
+    },
+  };
+
+  const governorFacets = await E(zoe).startInstance(
+    governor,
+    {},
```
dtribble, line 745: should the anchor issuer just be added directly to the issuers here?

packages/run-protocol/src/econ-behaviors.js
dckc: That would be relying on displayInfo for contract correctness and not just display. Is this an exception to a rule? Or is there just no rule?

packages/run-protocol/src/econ-behaviors.js
dckc: yes, that seems right

I should look more closely, though.

packages/run-protocol/src/econ-behaviors.js
dtribble: I now understand your offline comment. Yes, this should just be a number that comes from governance, and should have no use of "decimals

dckc:
obsolete in favor of #5165

PR #4957, opened 2022-03-30
feat(swingset): add syscall.abandonExports

### LINKED ISSUE #4951: add syscall.abandonExport()

When a vat is being upgraded, we preserve its durable exports (instances of a kind defined with `defineDurableKind()`), but we do *not* preserve the merely-virtual exports (defined with `defineKind()`).

By "not preserve", we mean that other (importing) vats will see the object become broken, just as if the vat had been terminated. If they send a message to this object, their result promise will be rejected.

Note that we're preserving the identity of the dead object. The kernel object table will continue to have an entry for it (however the `owner` field will be set to `null`). The importing vat c-lists will retain their entries. Only the exporting vat's c-list will be modifed, removing both the `koNN -> o+NN/etc` and the `o+NN/etc -> koNN` mappings.

We've talked in the past about making dead objects "taboo", revoking their identity entirely. We're not doing that now. If were, we'd also need a `dispatch.abandonImport()` call, to notify importing vats that their import has been removed from the c-list (therefore they must not reference it in outbound messages ever again).

The syscall name is a slight deviation from our existing naming scheme, but only in emphasis. From the vat's point of view, we have three messages that deal with imports:

* `syscall.dropImport()` : "Hey kernel, you know that thing I was importing? Well, I can no longer reach it, so don't keep it alive for my sake. I'll never mention it in a message again, but it might still be used as the key of a WeakMap/WeakSet, so let's keep it in the c-list for now."
* `syscall.retireImport()`: "Hey kernel, you know that thing I dropped? I no longer have any WeakMaps/WeakSets that are keyed by it (because I deleted the last one, or never had any in the first place), so let's delete it from our shared c-list.
* `dispatch.retireImport()`: "Hey vat, you know that import that you dropped (with syscall.dropImport)? Everybody else dropped it too, even the exporting vat stopped its internal references, so the object is completely gone. Please delete any WeakMap/WeakSet entries that were keyed by it. I'm deleting it from our shared c-list now. You'll never hear about it again (until/unless it gets re-introduced in some new message).

And we have three messages (now four) that deal with exports:

* `dispatch.dropExport()`: "Hey vat, you know that object you exported earlier? Everybody else dropped it. Don't keep it alive for our sake. I'll never mention it in an inbound `dispatch.deliver` or `dispatch.notify` again. However those other vats might still be using it as the key of a WeakMap/WeakSet, so please let me know if/when you delete it. I'm leaving it in our shared c-list for now.
* `dispatch.retireExport()`: "Hey vat, you know that object we dropped earlier? Well, no other vat is using it as a WeakMap/WeakSet key, so we don't need the identity any more. I'm deleting it from our shared c-list. Don't tell us if if/when you delete it, we no longer care."
* `syscall.retireExport()`: "Hey kernel, you know that object you dropped earlier? I deleted it. Tell all other vats (with `dispatch.retireImport`) to delete any WeakMap/WeakSet entries that were keyed by it, and delete it from our shared c-list. I'll never mention it again (until it gets re-exported in some new message).
* NEW `syscall.abandonExport()` (would be `syscall.dropExport` but with more emphasis): "Hey kernel, you know that object I exported earlier? I know everybody is still referencing it, but I don't love it anymore, so I'm going to forget about it. Delete it from our shared c-list. If anybody else tries to talk to it, give them an error. We will never mention it again (and it won't ever be re-introduced because I'm forgetting about it entirely, even from my virtualized data, and my allocation counters won't ever re-use the ID either).

If/when we destroy the object's identity, we'd use `dispatch.abandonImport` to signal the importing vats about the c-list deletion. But we're not doing that now.

Because we don't currently have any way to send importing vats a notification that their import has stopped working, they'll only find out if/when they try to send a message to it.

## Description of the Design

Add `syscall.abandonExports(vrefs)`.

The kernel will react by:
* mapping all vrefs to krefs (vat-fatal error if the vref was not in the c-list, or if it was not an object export `o+...`)
* clear the `owner` field in the kernel object table (set it to `null`)
* delete both entries from the calling vat's c-list

We need to confirm that the kernel object entry remains present until all other references are dropped. I *think* we already do this w.r.t. terminated vats, but we should double-check. The `gcAction` that wants to send `dispatch.dropExport` to the owning vat must tolerate that `owner` being empty.

## Security Considerations

None I can think of.

Note that userspace will not get access to this. It will only be used by liveslots during `dispatch.stopVat()`.

Some day, we might define a revocation authority at the time the object is created, in which case the userspace code that created an object might also be allowed to destroy it.

## Test Plan

Unit tests must confirm that the c-list entries are deleted, the kernel object table `owner` is cleared, and any external refcounts remain unchanged. It should also confirm that GC can still take place when the last importing vat retires their import, and that we don't have any GC actions sent to the exporting vat.

We should also test that an object in the merely-recognizable state can be abandoned properly. In this case, the exporting vat will see `dispatch.dropExport`, then `dispatch.stopVat`, and will emit `syscall.abandonExport` during the stopVat delivery.

refs #1848 (needed to delete leftover non-durable exports)

warner:
This allows liveslots to abandon a previously-exported object. The
kernel marks the object as orphaned (just as if the exporting vat was
terminated), and deletes the exporter's c-list entry. All importing
vats continue to have the same access as before, and the refcounts are
unchanged.

Liveslots will use this during `stopVat()` to revoke all the
non-durable objects that it had exported, since these objects won't
survive the upgrade. The vat version being stopped may still have a
Remotable or a virtual form of the export, so userspace must not be
allowed to execute after this syscall is used, otherwise it might try
to mention the export again, which would allocate a new mismatched
kref, causing confusion and storage leaks.

Our naming scheme would normally call this `syscall.dropExports`
rather than `syscall.abandonExports`, but I figured this is
sufficiently unusual that it deserved a more emphatic name. Vat
exports are an obligation, and this syscall allows a vat to shirk that
obligation.

closes #4951
refs #1848

### DIFF
--- a/packages/SwingSet/docs/garbage-collection.md
+++ b/packages/SwingSet/docs/garbage-collection.md
@@ -358,6 +358,14 @@ At this point, the kref is only referenced in the queued `retireImport` action a
 
 If the last importing vat had previously called `syscall.retireImport`, there will be no subscribers, but the kernel object data will still be present (a general invariant is that the exporting clist entry and the kernel object table entry are either both present or both missing). A `dispatch.retireExport` will be on the queue for the exporting vat, but it has not yet arrived (otherwise it would be illegal for the exporting vat to call `syscall.retireExport`). That `dispatch.retireExport` GC action will be nullified during `processOneGCAction` because its work was already performed by the exporter's `syscall.retireExport`.
 
+## syscall.abandonExport processing
+
+During vat upgrade, the last delivery made to the old version is a special `dispatch.stopVat()`. This instructs the soon-to-be-deleted worker to destroy anything that will not survive the upgrade. All Remotables will be abandoned here, because they live solely in RAM, and the RAM heap does not survive. All non-durable virtual objects will also be abandoned, since only durable ones survive. We may also drop otherwise-durable objects which were only referenced by abandoned non-durable objects.
+
+During this phase, the vat performs a `syscall.abandonExports()` with all the abandoned vrefs. The kernel reacts to this in roughly the same way it reacts to the entire vat being terminated (which it is, sort of, at least the heap is being terminated). Each vref is deleted from the exporting vat's c-list, because *that* vat isn't going to be referencing it any more. The kernel object table is updated to clear the `owner` field: while the kernel object retains its identity, it is now orphaned, and any messages sent to it will be rejected (by the kernel) with a "vat terminated" error.
+
+No other work needs to be done. Any importing vats will continue to hold their reference as before. They can only tell that the object has been abandoned if they try to send it a message. Eventually, if all the importing vats drop their reference, and nothing else in the kernel is holding one, the kernel object entry will be deleted. In this case, no `dispatch.retireExports` is sent to the old exporting vat, since it's already been removed from their c-list.
+
 ## Post-Decref Processing
 
 Those three syscalls may cause some krefs to become eligible for release. The "kernelKeeper" tracks these krefs in an ephemeral `Set` named `maybeFreeKrefs`. Every time a decrement causes the reachable count to transition from 1 to 0, or the recognizable count to transition from 1 to 0, the kref is added to this set.
--- a/packages/SwingSet/src/kernel/kernelSyscall.js
+++ b/packages/SwingSet/src/kernel/kernelSyscall.js
@@ -299,6 +299,14 @@ export function makeKernelSyscallHandler(tools) {
     return OKNULL;
   }
 
+  function abandonExports(vatID, koids) {
+    assert(Array.isArray(koids), X`abandonExports given non-Array ${koids}`);
+    for (const koid of koids) {
+      kernelKeeper.orphanKernelObject(koid, vatID);
+    }
+    return OKNULL;
+  }
+
   // callKernelHook is only available to devices
 
   function callKernelHook(deviceID, hookName, args) {
@@ -368,6 +376,10 @@ export function makeKernelSyscallHandler(tools) {
       case 'retireExports': {
         const [_, ...args] = ksc;
         return retireExports(...args);
+      }
+      case 'abandonExports': {
+        const [_, ...args] = ksc;
+        return abandonExports(...args);
       }
       case 'callKernelHook': {
         const [_, ...args] = ksc;
--- a/packages/SwingSet/src/kernel/state/kernelKeeper.js
+++ b/packages/SwingSet/src/kernel/state/kernelKeeper.js
@@ -527,6 +527,13 @@ export default function makeKernelKeeper(
     return owner;
   }
 
+  function orphanKernelObject(kref, oldVat) {
+    const ownerKey = `${kref}.owner`;
+    const ownerVat = kvStore.get(ownerKey);
+    assert.equal(ownerVat, oldVat, `export ${kref} not owned by old vat`);
+    kvStore.delete(ownerKey);
+  }
+
   function deleteKernelObject(koid) {
     kvStore.delete(`${koid}.owner`);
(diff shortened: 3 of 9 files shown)


packages/SwingSet/docs/garbage-collection.md
```diff
@@ -358,6 +358,14 @@ At this point, the kref is only referenced in the queued `retireImport` action a
...
 
+## syscall.abandonExport processing
+
+During vat upgrade, the last delivery made to the old version is a special `dispatch.stopVat()`. This instructs the soon-to-be-deleted worker to destroy anything that will not survive the upgrade. All Remotables will be abandoned here, because they live solely in RAM, and the RAM heap does not survive. All non-durable virtual objects will also be abandoned, since only durable ones survive. We may also drop otherwise-durable objects which were only referenced by abandoned non-durable objects.
+
+During this phase, the vat performs a `syscall.abandonExports()` with all the abandoned vrefs. The kernel reacts to this in roughly the same way it reacts to the entire vat being terminated (which it is, sort of, at least the heap is being terminated). Each vref is deleted from the exporting vat's c-list, because *that* vat isn't going to be referencing it any more. The kernel object table is updated to clear the `owner` field: while the kernel object retains its identity, it is now orphaned, and any messages sent to it will be rejected (by the kernel) with a "vat terminated" error.
```
FUDCo, line 365: Re: "a `syscall.abandonExports()` with all the abandoned vrefs": presumably Liveslots could issue multiple such calls?  My read of the code suggests there's no reason from the kernel's perspective it has to be all at once.  I can imagine batching if the number of severed exports is very large, or possibly different cleanup passes that discover different things (e.g., stuff in non-durable stores, then stuff in non-durable VOs, then stuff in unreferenced durable VOs, etc.).

Also, is `syscall.abandonExports` limited strictly to use during vat stoppage?  It seems like it is closely related to the `vatPowers.disavow` operation (indeed, it's not clear to me how they could be any different).

FUDCo (approved):
Aside from some rambling questions that the documentation raised in my mind, this looks good.

packages/SwingSet/docs/garbage-collection.md
warner: It could make multiple calls, yeah. It's more efficient to do larger batches, and I'll probably make `stopVat` do a single large call, because I'm more likely to get it right that way (I'm already having problems with `stopVat` not being aware of what is dropped by the `bringOutYourDead` call that it makes, and then `abandonExports` does something bad).

It *could* be called at other times, but I think `stopVat` is the only appropriate time because we're basically yanking the rug out from userspace, we're breaking their promise to make that previously-exported object available, and if we ever allow userspace to run after doing `stopVat`, they're likely to become confused or cause significant confusion.

`disavow` is for imports; it says "I have this Presence here, and I know that means I'm holding a reference, but pretend that I just dropped it and drop/retire the import". It's a different sort of rug-pull than the exports that `syscall.abandonExport` does. Abandoning an export is a tiny bit more serious than disavowing an import: the import is depended upon by userspace code (if it tries to do something with the Presence, things will break), whereas the export is depended upon by the kernel and other vats, so `abandonExport` lets you surprise more than just yourself. That said, neither is really meant for userspace control, and `vatPowers.disavow` is mainly just for unit tests.

FUDCo: Ah, I hadn't twigged to "disavow import" vs. "disavow export".  Naturally the latter is more radical, but vat code always has the option to act as if an export has been disavowed by simply throwing unconditionally in response to everything send to it (which means exposing `abandonExports` doesn't really do anything useful, so, yeah, let's not).

PR #4959, opened 2022-03-30
test(swingset): test that baggage and durables survive vat upgrade

warner:
We create a durable Kind, and reattach behavior to it in v2. The
handle must travel through baggage, demonstrating that baggage works.

I'm still looking for the right way to use VatData these from with
swingset tests.. other packages should import @agoric/vat-data, but
that might be circular from here

refs #1848


FUDCo (approved):
Very nice.  The flow for upgrade looks more or less as I expected, which is good calibration-of-understanding benchmark (for me, anyway).

PR #4960, opened 2022-03-30
fix: check that all durable kinds are reconnected in new version

warner:
This iterates through all previously-defined durable Kinds and asserts
that they have been reconnected by the time buildRootObject()
completes.

still needs better error delivery path

refs #1848

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -658,6 +658,7 @@ export function makeVirtualObjectManager(
 
   let kindIDID;
   const kindDescriptors = new WeakMap();
+  const definedDurableKinds = new Set(); // kindID
 
   function initializeKindHandleKind() {
     kindIDID = syscall.vatstoreGet('kindIDID');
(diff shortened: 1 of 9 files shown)


warner:
Oops, this changes the vatstore behavior in a lot of tests, and I didn't run enough of the suite to catch them. I'll keep it as a draft until I get those fixes. I don't expect the main code to change, though, only the tests.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -697,7 +698,39 @@ export function makeVirtualObjectManager(
...
+
+  function insistAllDurableKindsReconnected() {
+    // identify all user-defined durable kinds by iterating `vom.kind.*`
+    const start = `vom.kind.`;
+    const end = `vom.kind/`; // '.' and '/' are sequential in ASCII/UTF-8
+    let key = '';
+    let value;
+    const missing = [];
+    for (;;) {
+      [key, value] = syscall.vatstoreGetAfter(key, start, end);
```
FUDCo: Don't manage `start` and `end` yourself.  The API is designed to handle prefix encoding so you can just do
```
      [key, value] = syscall.vatstoreGetAfter(key, 'vom.kind.');
```
Let the vatstore worry about how ASCII works, 'cause it's engineered to do that.

A better idiom for the loop as a whole is actually:
```
let [key, value] = syscall.vatstoreGetAfter('', prefix);
while (key) {
  <<do stuff>>
  [key, value] = syscall.vatstoreGetAfter(key, prefix);
}
```

FUDCo (approved):
One quibble with the coding of the iteration over kinds, which you're free to ignore though you shouldn't.

Other than that, looks good.

packages/SwingSet/src/liveslots/virtualObjectManager.js
warner: ooh, cool

PR #4961, opened 2022-03-30
chore(swingset): force GC on more intermittently-failing tests

warner:
The previous PR didn't go far enough, there are other tests in the
same file that are still failing intermittently. This turns on the
forceGC flag on *all* tests in test-virtualObjectGC.js

refs #4936


FUDCo (approved):
The things we do for love.

PR #4962, opened 2022-03-30
Implement the latest version of the `defineKind` API

### LINKED ISSUE #4905: new defineKind API with { state, facets } argument

To prevent the GC sensitivity of #4892, in today's kernel meeting we came up with a new API for `defineKind`. The current API looks like:
```js
function init(args) {
  const initialState = {}; // use 'args'
  return initialState;
}
function actualize(state) {
  const oneFacet = { meth1(args) { .. }, meth2(args) { .. } };
  // or
  const facets = {
     facetA: { meth1(args) { .. }, meth2(args) { .. } },
     facetB: { meth1(args) { .. }, meth2(args) { .. } },
  };
  return oneFacet; // or facets
}
function finish(state, cohort) {
  const { facetA } = cohort;
  // do some final wiring
}
const makeFoo = defineKind('foo iface', init, actualize, finish);
```

In the new API, the third argument to `defineKind` will be a record of unbound functions, each of which takes an initial argument `{ state, self }` or `{ state, facets }`:
```js
// one facet:
const behavior = {
  meth1: ({ state, self }, ...args) => { .. },
  meth2: ({ state, self }, ...args) => { .. },
};
// multiple facets
const behavior = {
  facetA: {
    meth1: ({ state, facets }, ...args) => { .. },
    meth2: ({ state, facets }) => { .. },
  },
  facetB: {
    meth1: ({ state, facets }) => { .. },
    meth2: ({ state, facets }, ...args) => { .. },
  },
};
```

In addition, the fourth argument will be an options bag, with the only currently-accepted option being `finish`.

```js
const makeFoo = defineKind('foo iface', init, behavior, { finish });
```

The `behavior` record is examined during `defineKind`, and a copy is made of the functions it provides (to ensure that userspace does not get control later, during actualization). Later, when `makeFoo(args)` is invoked and we need to make a new virtual object cohort, the sequence is:
* call `init(args)`, get back the initial state record
* create a `state` object with getters and setters for each property in the initial state record
* create a "context" object with initial contents `{ state, facets: {} }` (or just `{ state }` for single-facet kinds)
* for each facet in `behavior` (or just the one facet):
  * create a new object (this will become the facet)
  * for each function in the behavior record:
    * create a new function (with `Function.prototype.bind`) that curries the context object as the first argument
    * add that function as a property of the new facet object
  * add the facet to `context.facets` (or assign the one facet as `context.self`)
* build the cohort object, harden everything
* call `finish` if defined
* return the cohort to the user (return value of `makeFoo()`)

The goal is to prevent userspace from getting control during actualization, giving it a way to create per-instance objects that could be used to sense GC. The bound methods (and `state`, and the context object, and the facet objects within) are all per-instance, but they are all visible to the Kind machinery during actualization, so we can establish WeakMap links from them to the cohort record, which means they'll have the same lifetime, which prevents their use as GC sensors. By having userspace provide a table of functions at `makeKind` time, rather than executing a userspace `actualize` function at actualize time, we don't give them any opportunity to make new per-instance objects (or to even sense when actualization is taking place).

I think this will let us safely remove the `proForma` "call `actualize` anyways and throw out the result" code in the VOM, which would be lovely.

One small surprise for Kind authors is that their methods will have a different signature than the code which invokes them. The first argument is provided by the Kind machinery.

Kinds which simply want to refer to their state should get it by destructuring the first argument:
```js
const behavior = {
  incr: ({ state }, delta) => { state.count += delta; }
};
// counter.incr(delta)
```

When a kind needs to refer to other facets, it should grab that from `facets`:
```js
const behavior = {
  depositFacet: {
    deposit: ({ facets }, payment) => facets.purse.deposit(payment),
  },
  purse: {
    withdraw: ({ state }, amount) => { state.balance -= amount; ... },
    deposit: ({ state }, payment) => { ... },
    getDepositFacet: ({ facets }) => facets.depositFacet,
  },
};
```

If a single-facet virtual object needs to refer to itself, it should use `self`:
```js
const behavior = {
  subscribe: ({ state, self }) => state.publisher.subscribe(self),
};
```

## Rejected Variants

We considered `method: ({ state, ...facets }) => ...`, which would shrink some things like `getDepositFacet: ({ depositFacet }) => depositFacet` by the length of `facets.`. This would preclude the use of `state` as a facet name, and would complicate the addition of new context properties in the future. We decided that referencing other facets was infrequent enough that it wasn't worth the preclusion.

There may be value to internally representing all Kinds as multi-faceted, even those with a single facet. Independent of that, we considered having `makeKind` always be multi-facet, so authors of single-facet Kinds would need to pick a facet name. This would remove `self` from the context argument (it would always have `facets`, even if there was only one property on it). The Kind defintion would then look like:
```js
const behavior = {
  dummyName: {
    method1: ({ state, facets }) => { stuff(state, facets.dummyFacet ...} },
  },
};
```

and userspace would look like:
```js
const foo = makeFoo(args).dummyName;
```

either of which were sufficient to disqualify the idea. Instead, `defineKind` will look at the shape of `behavior` to decide whether it creates single-facet objects or multi-facet cohorts. The internals might or might not manage both in the same way, but the `behavior` record and the return value of `makeFoo` will be different for the two cases.

We considered using JavaScript's special `this` variable to hold the `{ state, facets }` context object. This would be closer to the usual prototype-based inheritance (possibly identical), but would require `behavior` be defined with concise method syntax, precluding the use of `=>` arrow functions (which have other benefits). We hope that "Far Classes" will be implemented eventually, which should avoid the surprising extra context argument and also be stored more efficiently.

We decided that `finish` is used infrequently enough that a reader of a `defineKind` definition would benefit from seeing its name appear in the options bag, and that this benefit (plus the general benefit of options bags if/when new arguments are added) justified changing the fourth `defineKind` argument from an optional `finish` function to an optional `{ finish }` bag.

## Security Considerations

Userspace must not get control during actualization, even if the behavior record they provide is a `Proxy` or has getters. We must enumerate the contents once, during `defineKind`, and assert that we get plain `Function` objects from it. We should store those Functions in the Kind record for use during actualization.

We must double-check that the act of binding a Function is not visible to that Function, perhaps through some `Proxy` trickery that I'm not aware of (does `.toString` get called for any reason?).

## Test Plan

Unit tests, ideally ones that don't pull in all of swingset, but maybe that's the easiest way to do it.

FUDCo:
This implements the newest version of the `defineKind` API, as laid out in #4905.

Closes #4905

### DIFF
--- a/packages/ERTP/src/payment.js
+++ b/packages/ERTP/src/payment.js
@@ -9,13 +9,11 @@ import { defineKind } from '@agoric/vat-data';
  * @returns {() => Payment<K>}
  */
 export const makePaymentMaker = (allegedName, brand) => {
-  const makePayment = defineKind(
-    `${allegedName} payment`,
-    () => ({}),
-    () => ({
-      getAllegedBrand: () => brand,
-    }),
-  );
-  return makePayment;
+  const makePayment = defineKind(`${allegedName} payment`, () => ({}), {
+    getAllegedBrand: () => brand,
+  });
+  // XXX the following type cast is meatball surgery to make tsc shut up
+  // somebody who understands this should do it properly
+  return /** @type {() => Payment<K>} */ (makePayment);
 };
 harden(makePaymentMaker);
--- a/packages/ERTP/src/purse.js
+++ b/packages/ERTP/src/purse.js
@@ -3,6 +3,11 @@ import { defineKind } from '@agoric/vat-data';
 import { AmountMath } from './amountMath.js';
 
 export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
+  const updatePurseBalance = (state, newPurseBalance) => {
+    state.currentBalance = newPurseBalance;
+    state.balanceUpdater.updateState(state.currentBalance);
+  };
+
   // - This kind is a pair of purse and depositFacet that have a 1:1
   //   correspondence.
   // - They are virtualized together to share a single state record.
@@ -24,41 +29,33 @@ export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
         balanceUpdater,
       };
     },
-    state => {
-      const { balanceNotifier, balanceUpdater } = state;
-      const updatePurseBalance = newPurseBalance => {
-        state.currentBalance = newPurseBalance;
-        balanceUpdater.updateState(state.currentBalance);
-      };
-
-      /** @type {Purse} */
-      const purse = {
-        deposit: (srcPayment, optAmountShape = undefined) => {
+    {
+      purse: {
+        deposit: ({ state }, srcPayment, optAmountShape = undefined) => {
           // Note COMMIT POINT within deposit.
           return purseMethods.deposit(
             state.currentBalance,
-            updatePurseBalance,
+            newPurseBalance => updatePurseBalance(state, newPurseBalance),
             srcPayment,
             optAmountShape,
           );
         },
-        withdraw: amount =>
+        withdraw: ({ state }, amount) =>
           // Note COMMIT POINT within withdraw.
           purseMethods.withdraw(
             state.currentBalance,
-            updatePurseBalance,
+            newPurseBalance => updatePurseBalance(state, newPurseBalance),
             amount,
           ),
-        getCurrentAmount: () => state.currentBalance,
-        getCurrentAmountNotifier: () => balanceNotifier,
+        getCurrentAmount: ({ state }) => state.currentBalance,
+        getCurrentAmountNotifier: ({ state }) => state.balanceNotifier,
         getAllegedBrand: () => brand,
         // eslint-disable-next-line no-use-before-define
-        getDepositFacet: () => depositFacet,
-      };
-      const depositFacet = {
-        receive: purse.deposit,
-      };
-      return { purse, depositFacet };
+        getDepositFacet: ({ facets }) => facets.depositFacet,
+      },
+      depositFacet: {
+        receive: ({ facets }, ...args) => facets.purse.deposit(...args),
+      },
     },
   );
   return () => makePurseKit().purse;
--- a/packages/SwingSet/docs/virtual-objects.md
+++ b/packages/SwingSet/docs/virtual-objects.md
@@ -19,9 +19,9 @@ A VDO has a "kind", which defines what sort of behavior and state it will posses
 
 A vat can define new kinds of VDOs by calling the `defineKind` or `defineDurableKind` functions:
 
-  `maker = defineKind(descriptionTag, init, actualize, finish)`
+  `maker = defineKind(descriptionTag, init, behavior, options)`
 or
-  `maker = defineDurableKind(kindHandle, init, actual, finish)`
+  `maker = defineDurableKind(kindHandle, init, behavior, options)`
 
 The return value from `defineKind` or `defineDurableKind` is a maker function which the vat can use to create instances of the newly defined VDO kind.
 
@@ -31,50 +31,56 @@ A `kindHandle` is a type of durable object that can be used to identify the kind
 
   `kindHandle = makeKindHandle(descriptionTag)`
 
-where `descriptionTag` is exactly the same as the same named parameter of `defineKind`.  The difference is that a kind handle is itself that a durable object that may be stored for later retrieval, and used in a future call to `defineDurableKind` to associate new behavior with the kind in question.
+where `descriptionTag` is exactly the same as the same named parameter of `defineKind`.  The difference is that a kind handle is itself a durable object that may be stored for later retrieval, and used in a future call to `defineDurableKind` to associate new behavior with the kind in question.
 
 The `init` parameter is a function that will be called when new instances are first created. It is expected to return a simple JavaScript object that represents the initialized state for the new VDO instance.  Any parameters passed to the maker function returned by `defineKind`/`defineDurableKind` are passed directly to the `init` function.
 
-The `actualize` parameter is a function that binds an in-memory instance (the "Representative") of the VDO with the VDO's state, associating such instances with the VDO's behavior.  It is passed the VDO's state as a parameter and is expected to return either:
+The `behavior` parameter is an object that describes the VDO's behavior.  It must take one of two forms:
 
-1. A new JavaScript object with methods that close over the given state.  This returned object will become the body of the new instance.  This object can be empty; in such a case it can serve as a powerless but unforgeable "marker" handle.
+1. An object whose named properties are all functions that will become methods of the virtual objects returned by the maker function.  The behavior object can be empty; in such a case the resulting VDO can serve as a powerless but unforgeable "marker" handle.
 
-2. A new JavaScript object populated with objects as described in (1).  These will become facets of the new instance.  The returned object will be an object mapping to the facets by name.
+2. An object whose named properties are objects as described in (1).  These will become facets of new instances of the VDO.  The return value from the maker function object will be an object mapping to the facets by name.
 
-The `actualize` function is called whenever a new VDO instance is created, whenever such an instance is swapped in from secondary storage, and whenever a reference to a VDO is received as a parameter of a message and deserialized.  Note that for any given VDO kind, the shape of the value returned by the `actualize` function may not vary over successive calls.  That is, if it's a single facet, it must always be a single facet, and if it's multiple facets it must always be the same set of multiple facets.
+In either case, the individual behavior functions must have the signature:
 
-The `finish` parameter is optional. It is a function that, if present, will be called exactly once as part of instance initialization.  It will be invoked _immediately_ after the `actualize` function for that instance is called for the very first time.  In other words, it will be called after the instance per se exists but before that instance is returned from the maker function to whoever requested its creation.  `finish` is passed two parameters: the VDO's state (exactly as passed to the `actualize` function) and the VDO itself. The `finish` function can modify the object's state in the context of knowing the object's identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other VDOs.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.  In particular, if one or more of the object's methods need to refer to the object itself (for example, so it can pass itself to other objects), the `finish` function provides a way to capture that identity as part of the object's state.
+  `methodname(context, ...args) { ...`
+
+where `context` describes the invocation context of the method and `...args` are whatever arguments were passed to the method when it was invoked.  In the case of a single facet VDO, `context` will take the form `{ state, self }`, where `state` is the VDO state and `self` is a reference to the VDO itself.  In the case of a multi-facet VDO, `context` will instead be `{ state, facets }`, where `facets` is an object whose named properties are the facets of the VDO.
+
+The `options` parameter is optional. It provides additional parameters to characterize the VDO.  Currently there is only one supported option, `finish`, though we anticipate more options may be added in future versions of the API.
+
+The `finish` option is a function that, if present, will be called at the end of instance initialization.  It will be invoked after the VDO is created but before it is returned from the maker function.  `finish` is passed one parameters, a `context` object identical to that passed to method behaviors.  The `finish` function can modify the object's state at a time when the object's identity is known (or its facets' identies are known), and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other VDOs.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.
 
 For example:
 
 ```javascript
   const initCounter = (name) => ({ counter: 0, name });
 
-  const actualizeCounter = (state) => ({
-    inc: () => {
+  const counterBehavior = {
+    inc: ({state}) => {
       state.counter += 1;
     },
-    dec: () => {
+    dec: ({state}) => {
       state.counter -= 1;
     },
-    reset: () => {
+    reset: ({state}) => {
       state.counter = 0;
     },
-    rename: (newName) => {
+    rename: ({state}, newName) => {
       state.name = newName;
     },
-    getCount: () => state.counter,
-    getName: () => state.name,
-  });
+    getCount: ({state}) => state.counter,
+    getName: ({state}) => state.name,
+  };
 
   const finishCounter = (state, counter) => {
     addToCounterRegistry(counter, state.name);
   };
 
-  const makeCounter = defineKind('counter', initCounter, actualizeCounter, finishCounter);
+  const makeCounter = defineKind('counter', initCounter, counterBehavior, { finish: finishCounter });
 ```
 
-This defines a simple virtual counter object with two properties in its state, a count and a name.  You'd use it like this:
+This defines a simple virtual counter object with two properties in its state: a count and a name.  Note that none of the methods bother to declare the `self` context parameter because none of them need to refer to it.  You'd use it like this:
 
 ```javascript
   const fooCounter = makeCounter('foo');
@@ -93,33 +99,33 @@ Suppose you instead wanted to provide a version with the increment and decrement
 ```javascript
   const initFacetedCounter = () => ({ counter: 0 });
 
-  const actualizeFacetedCounter = (state) => {
-    const getCount = () => state.counter;
-    return {
-      incr: {
-        step: () => {
-          state.counter += 1;
-        },
-        getCount,
-      },
-      decr: {
-        step: () => {
-          state.counter -= 1;
-        },
-        getCount,
-      },
-    };
-  }
+  const getCount = ({state}) => state.counter,
 
-  const makeFacetedCounter = defineKind('counter', initCounter, actualizeCounter);
+  const facetedCounterBehavior = {
+    incr: {
+      step: ({ state }) => {
+        state.counter += 1;
+      },
+      getCount,
+    },
+    decr: {
+      step: ({ state }) => {
+        state.counter -= 1;
+      },
+      getCount,
+    },
+  };
+
+  const makeFacetedCounter = defineKind('counter', initFacetedCounter, facetedCounterBehavior);
 ```
 
-If you wanted to also make this durable, instead of the last line you'd generate
-the kind with something more like:
+Note how the `getCount` method is declared once and then used in two different facets.
+
+If you wanted to also make this durable, instead of the last line you'd generate the kind with something more like:
 
 ```javascript
   const facetedCounterKind = makeKindHandle('durable counter');
-  const makeFacetedCounter = defineDurableKind(facetedCounterKind, initCounter, actualizeCounter);
+  const makeFacetedCounter = defineDurableKind(facetedCounterKind, initCounter, facetedCounterBehavior);
 ```
 
 In either case you'd use it like:
@@ -134,23 +140,23 @@ In either case you'd use it like:
   console.log(`count is ${incr.getCount()`); // "count is 1"
 ```
 
-Note that the `init`, `actualize`, and `finish` functions are defined explicitly in the above examples for clarity of exposition, but in practice you'd usually declare them inline in the parameters of the `defineKind` call:
+Note that the `init` and `finish` functions, as well as the behavior, are defined explicitly in the above examples for clarity of exposition, but in practice you'd usually declare them inline in the parameters of the `defineKind` call:
 
 ```javascript
+  const getCount = ({state}) => state.counter;
   const makeFacetedCounter = defineKind(
     'counter',
     () => ({ counter: 0 }),
     (state) => {
-      const getCount = () => state.counter;
       return {
         incr: {
-          step: () => {
+          step: ({state}) => {
             state.counter += 1;
           },
           getCount,
         },
         decr: {
-          step: () => {
+          step: ({state}) => {
             state.counter -= 1;
           },
           getCount,
@@ -162,7 +168,7 @@ Note that the `init`, `actualize`, and `finish` functions are defined explicitly
 
 Additional important details:
 
-- The set of state properties of an instance is fully determined by the `init` function.  That is, the set of properties that exist on in instance's `state` is completely determined by the enumerable properties of the object that `init` returns.  State properties cannot thereafter be added or removed.  Currently there is no requirement that all instances of a given kind have the same set of properties, but code authors should not rely on this as such enforcement may be added in the future.
+- The set of state properties of an instance is fully determined by the `init` function.  That is, the set of properties present in instance's `state` is completely determined by the named enumerable properties of the object that `init` returns.  State properties cannot thereafter be added or removed from the instance.  Currently there is no requirement that all instances of a given kind have the same set of properties, but code authors should not rely on this as such enforcement may be added in the future.
 
 - The values a state property may take are limited to things that are serializable and which may be hardened (and, in fact, _are_ hardened and serialized the moment they are assigned).  That is, you can replace what value a state property _has_, but you cannot modify a state property's value in situ.  In other words, you can do things like:
 
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -7,6 +7,14 @@ import { parseVatSlot } from '../lib/parseVatSlots.js';
 
 // import { kdebug } from './kdebug.js';
 
+// Marker associated to flag objects that should be held onto strongly if
+// somebody attempts to use them as keys in a VirtualObjectAwareWeakSet or
+// VirtualObjectAwareWeakMap, despite the fact that keys in such collections are
+// nominally held onto weakly.  This to thwart attempts to observe GC by
+// squirreling away a piece of a VO while the rest of the VO gets GC'd and then
+// later regenerated.
+const unweakable = new WeakSet();
+
 /**
  * Make a simple LRU cache of virtual object inner selves.
  *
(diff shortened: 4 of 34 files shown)


FUDCo:
Note: I've marked this as DRAFT because it will not pass CI, because there are a couple of uses of `defineKind` elsewhere in the source tree that need to be updated to use the new API, and what those uses are doing is complicated and weird and hard to understand.  MarkM and I will pair on fixing these on Thursday afternoon, but I wanted to get this PR up so the main substance can be reviewed.

warner:
Does this also add the crosslinks and `unweakables`?

FUDCo:
Yes.  Two separate commits, one for the API overhaul and one for the anti-GC-observability stuff.

dckc (commented):
I defer to other reviewers.

packages/ERTP/src/purse.js
```diff
@@ -24,41 +29,34 @@ export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
...
-        state.currentBalance = newPurseBalance;
-        balanceUpdater.updateState(state.currentBalance);
-      };
-
+    {
       /** @type {Purse} */
-      const purse = {
-        deposit: (srcPayment, optAmountShape = undefined) => {
+      purse: {
+        deposit: ({ state }, srcPayment, optAmountShape = undefined) => {
```
turadg, line 34: the type of `purse` is no longer `Purse`, which is the use API.

we might want a helper like `WithContext` so that we can still define an external API and then use it here to constrain the implementation. e.g. `/** @type {WithContext<Purse>} */`. Though I'm sure there's a better name.

packages/SwingSet/docs/virtual-objects.md
```diff
@@ -31,50 +31,56 @@ A `kindHandle` is a type of durable object that can be used to identify the kind
...
+
+The `options` parameter is optional. It provides additional parameters to characterize the VDO.  Currently there is only one supported option, `finish`, though we anticipate more options may be added in future versions of the API.
+
+The `finish` option is a function that, if present, will be called at the end of instance initialization.  It will be invoked after the VDO is created but before it is returned from the maker function.  `finish` is passed two parameters: (1) the VDO's state and (2) the VDO itself or an object containing its facets.  This is essentially the same two elements that are in the `context` parameter that is passed to method behaviors.  The `finish` function can modify the object's state in the context of knowing the object's (or its facets) identity, and thus can be used in cases where a validly initialized instance requires it to participate in some kind of cyclical object graph with other VDOs.  It can also be used, for example, to register the object with outside tracking data structures, or do whatever other post-creation setup is needed for the object to do its job.
```
warner: Should we maybe give it the same `context` object as the `behavior` methods get? I don't think we explored this when talking about `finish`, but now I'm wondering if it would be more uniform. Not a big deal either way, for sure.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -411,6 +449,27 @@ export function makeVirtualObjectManager(
     }
   }
 
+  function copyMethods(behaviorTemplate) {
```
warner: `copyMethods` is given the user's `behavior` object and *returns* a `behaviorTemplate`, right? (or at least one facet's portion of a `behaviorTemplate`). Can we rename the argument and `obj` to reflect that? We need to defend against the user's `behavior` object doing something tricky (like changing its entries between two calls to `Object.entries`), but I think it can't do that because we only run `Object.entries()` against it once (in `copyMethods`), but these argument names make me think otherwise.

Also, (and maybe this is an @erights question) should we consider `harden`ing the user's `behavior` object when it is first submitted to `defineKind`? We're no longer using their records to build our facets (we build entirely new records), so having them pre-`harden`ed shouldn't interfere with the new objects being made `Far`.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -411,6 +449,27 @@ export function makeVirtualObjectManager(
...
+      obj[name] = func;
+    }
+    return obj;
+  }
+
+  function bindMethods(context, behaviorTemplate) {
+    const obj = {};
+    for (const [name, func] of Object.entries(behaviorTemplate)) {
+      assert.typeof(func, 'function');
+      const method = (...args) => Reflect.apply(func, null, [context, ...args]);
```
warner, line 465: I'll definitely defer to @erights here to weigh in on the correctness. I think I assumed that there was a `Reflect` method that curried `this` and the first N arguments into a new function that could be called with `(...args)`. This looks like we're calling `Reflect.apply` every time someone invokes the method.

But now that I skim through MDN on `Reflect.*`, it seems like this is the only tool it offers, in which case this seems correct to me.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -411,6 +449,27 @@ export function makeVirtualObjectManager(
...
+      unweakable.add(method);
+      unweakable.add(context);
```
warner: I'd suggest adding `context` to the unweakables down where it's created, rather than here (multiple times).

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -411,6 +449,27 @@ export function makeVirtualObjectManager(
     }
   }
 
+  function copyMethods(behaviorTemplate) {
+    const obj = {};
+    for (const [name, func] of Object.entries(behaviorTemplate)) {
```
warner: This enables Symbol-named methods, yeah? I think we want that, at least in the long run, and anyways `Far` should be the judge of what keys are allowed, so I think `Object.entries` is exactly right.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -544,61 +633,55 @@ export function makeVirtualObjectManager(
...
-          }
-          harden(toExpose);
-          break;
+      unweakable.add(state);
+      if (facetNames === null) {
+        const context = { state };
+        context.self = bindMethods(context, behaviorTemplate);
+        toHold = Far(tag, context.self);
+        // eslint-disable-next-line no-proto
+        linkToCohort.set(toHold.__proto__, toHold);
```
warner: Is there any benefit to using `Object.getPrototypeOf` instead of `__proto__`? What happens if someone defines a method named `__proto__` in their `behavior`?

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -544,61 +633,55 @@ export function makeVirtualObjectManager(
...
+          linkToCohort.set(facet, toHold);
         }
-        case 'not':
-          assert.fail(X`invalid self actualization for ${q(tag)}`);
-        default:
-          assert.fail(X`unexepected facetiousness: ${q(facetiousness)}`);
+        unweakable.add(facets);
+        harden(context);
+        harden(facets);
+        harden(toExpose);
```
warner, line 670: think we should `harden(toHold)` for completeness' sake?

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -669,12 +746,18 @@ export function makeVirtualObjectManager(
...
-  function reanimateDurableKindID(vobjID, _proforma) {
+  function reanimateDurableKindID(vobjID) {
     const { subid: kindID } = parseVatSlot(vobjID);
     const raw = syscall.vatstoreGet(`vom.kind.${kindID}`);
     assert(raw, X`unknown kind ID ${kindID}`);
     const durableKindDescriptor = harden(JSON.parse(raw));
     const kindHandle = Far('kind', {});
+    // @ts-expect-error it would be great if TypeScript knew JavaScript
+    // eslint-disable-next-line no-proto
+    linkToCohort.set(kindHandle.__proto__, kindHandle);
```
warner: Maybe we should have a helper function `protect(obj, target)`, which walks the prototype chain, adding each prototype to `unweakable` and establishing the `linkToCohort`, until it reaches `Object.prototype`. I had a sudden fear that someone might change `Far` down the road to have *two* hoisted prototypes, and we'd never notice the security problem.

warner: Also it gets called like 4 times and it's currently 4-6 lines each, so it might be more concise too.

packages/SwingSet/test/virtualObjects/test-representatives.js
```diff
@@ -420,13 +420,61 @@ test('empty facets are not orphaned', async t => {
...
+  const c = await buildVatController(config, [mode], { hostStorage });
   c.pinVatRoot('bootstrap');
 
   await c.run();
   t.deepEqual(
     c.kpResolution(c.bootstrapResult),
     capargs({ '@qclass': 'undefined' }),
   );
-  t.deepEqual(c.dump().log, ['compare originalFacet === thing : true']);
+  t.deepEqual(c.dump().log, ['compare old === new : true']);
```
warner, line 431: nice! thanks for testing this so thoroughly

warner:
The `xsnap-native` change looks like a merge problem? I'm guessing it should not be in this branch.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: I wish. Two factors combine to make this not work well:
   * `Object.entries` omits non-enumerable properties
   * Object literals with symbol-named properties initialize those properties to be non-enumerable.

```js
> x = {[Symbol.asyncIterator]: () => null}
{ [Symbol(Symbol.asyncIterator)]: [Function: [Symbol.asyncIterator]] }
> Object.entries(x)
[]
> Object.getOwnPropertyDescriptors(x)
{
  [Symbol(Symbol.asyncIterator)]: {
    value: [Function: [Symbol.asyncIterator]],
    writable: true,
    enumerable: true,
    configurable: true
  }
}
```

erights: No, forget what I just said. The session I captured above shows that my second bullet is false. The new property is enumerable. But `Object.entries` omitted it anyway. Which corrected my dim memory: `Object.entries` only includes string-named properties.

In any case, I agree that `Far` should know what it accepts and others should make minimal assumptions about what that is. Use `Reflect.ownKeys` or `Object.getOwnPropertyDescriptors` in order to minimize such assumptions.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Oh god yes! Please avoid `__proto__`. Not only can `Object.prototype.__proto__` be overridden, Caridy convinced us and tc39 that secure environments should probably remove it on initialization. SES does not yet do so, but probably will and should (attn @kriskowal @mhofman ). tc39 made it normative optional, so implementations are allowed to not provide it in the first place (as a future SES/XS may elect to do).

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: I expect `FarClasses` will need something like this anyway.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: > should we consider `harden`ing the user's `behavior` object when it is first submitted to `defineKind`? We're no longer using their records to build our facets (we build entirely new records), so having them pre-`harden`ed shouldn't interfere with the new objects being made `Far`.

That's a good point. Yes, we should preemptively `harden`.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: We created our own `uncurryThis`, but it just calls `Reflect.apply` to solve the underlying problem.

packages/SwingSet/docs/virtual-objects.md
FUDCo: The same thought occurred to me as I was writing it, but ultimately I decided that the gratuitous destructuring didn't really pay for itself.  However, that's not a strongly held belief on my part and I'm happy to change it to work the other way if that's the consensus.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: I'm plucking the methods out of the `behavior` object that's passed in to guard against other crap that the user might have hung off the object.  After extracting the methods there's really no benefit to `harden`ing (or otherwise fiddling with) the original object.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: This is the incantation that @erights gave me.  I have no investment in it.  If a better recipe comes along, I'm happy to use it.  In the meantime I think this is the best we can do.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: This was a case of my own ignorance coming into play.  I had never before, in years of JS development, ever had occasion to want to get to the prototype of an object.  Fortunately, this is a very easy fix.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: In practical terms that would result in neither harm nor benefit, but I like the aesthetic of positivism w.r.t immutability.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: So are we going to explicitly allow symbol-named methods here and leave it to `Far` to reject them?
What about symbol-named facets?

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: I started with something like that and it was a fail, because it turns out that things that need to make `unweakable` and things that need `linkToCohort` are not as correlated as one might imagine.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: Thinking about this further, symbol-named facets aren't straightforward, because I need to put them in order and sorting Symbols is a pain.  In any event, I don't think there's a use case for this aside from obsessive generality, so let's take the idea of symbol-named facets off the table for now.

FUDCo: On deeper analysis, symbol-named methods are kind of PITA too, though if we think it's worth the effort I can do it.  But I really don't think it's worth the effort.

erights: Definitely no symbol-named facets. The record naming facets is a copyRecord, and the properties of a copyRecord must be string-named own enumerable properties, so they're exempt from all these painful differences between ways to enumerate own properties. So this restriction is architectural, not expedient.

> symbol-named methods here and leave it to `Far` to reject them?

`Far` does not reject symbol-named methods. Nor does it reject methods named by non-enumerable properties.

But we currently have only one case we need, `@@asyncIterator`, which we could continue to special case for now. If postponing general support for symbol-named methods gets us to MN1 faster, I'm all for postponing. But I think our distributed computational model should eventually support symbol-named methods in general.

packages/SwingSet/docs/virtual-objects.md
erights: My inclination is the other way, to use the same context object. Not because the destructuring is compelling, but so there's one less thing to learn.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: I agree there is no strong reason to `harden` this one. As long as the extraction is careful not to be confusable by a self-mutating behavior object. For example, doing the extraction all at once with an `Object.getOwnPropertyDescriptors` and then fully ignoring the original object would be safe.

FUDCo: That's an interesting wrinkle I hadn't thought about: is `Object.getOwnPropertyDescriptors` atomic in a way that iterating over the array returned by `Object.entries` is not?  Thinking about it, that seems right, since if you use `Object.entries`, each pass of the iteration gives the object being iterated over the opportunity to do something funky mid-iteration if one or more of the properties is a getter.  (Of course in order to use the value returned by `Object.getOwnPropertyDescriptors` you have to iterate over _it_ using `Object.entries` or one of its brethren, which is kind of a pain and feels at first glance like you're just recursing on the problem -- but presumably you are now in exclusive possession of _that_ object in a way that you are not in exclusive possession of the original object).  On the other hand, it's unclear to me what mischief somebody could cause by having the behavior object modify itself during iteration, since `defineKind` is only going to iterate over the behavior object once, so any modification is as if that was what it was supposed to do all along.  All `defineKind` cares about is obtaining a set of method functions, so if the object squirmed around somehow while being iterated I don't see why `defineKind` would care. It got its bag of methods from the thing its caller gave it, so where's the harm? Seems to me the worst case is they wrote some weird code that they might not understand later, but anyone can confuse themselves if they work at it.

packages/SwingSet/docs/virtual-objects.md
FUDCo: OK, done.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: If I can put in symbol-named methods in less than, say, an hour's work I'll do it.  I suspect I probably can.

FUDCo: A weird thought occurred to me as I was pondering implementation: as long as we do not support symbol-named methods in our over-the-wire message encoding, someone can use symbol-named methods as a kind of private-to-the-local-vat scoping technique.  I'm not saying this is a good idea, but I can imagine people doing it.  Then, when we turn that feature on, we break their imagined security perimeter.  (Even once we support symbols in the method selector portion of a message, I presume they would still be able to use non-registered, non-well-known symbols for this purpose.)  This makes me wonder if we should hold off implementing symbol-named methods until we are also implementing symbols as method selectors in messages.

FUDCo:
I think this is now ready for final review.

Note to @michaelfig @Chris-Hibbert @turadg: would you please take a look at `packages/run-protocol/src/vaultFactory/vault.js`?  I had to make fairly invasive changes to update it for the new `defineKind` API,, which I don't believe changed any of the semantics (it passes all the tests, anyway), but given the radical transformation I prefer somebody take a look who actually understands the substance of the underlying code.

turadg:
> would you please take a look at `packages/run-protocol/src/vaultFactory/vault.js`

Thanks for the ping. I don't see any change in semantics, which is good. I see a separation of `selfBehavior` and `helperBehavior`, which is good. I do have some concerns.

1) Git blame churn. Many lines in this file are being touched. I suggest we have a codeowner of run-protocol be the author of the commit for `vault.js` so that git blame tracks back. I'm happy to take that on.

2) Loss of JSDoc. JSDoc (at least in my IDE) is on the value of the object and when a function is composed into a new object (e.g. `selfBehavior`) it doesn't come along. I think each self behavior function should be nested in the `selfBehavior` object. Without that the function documentation will be lost in the calling context.

3) Untyped `facets`. That can be fixed with,
```js
/**
 * @typedef {{
 *   state: ImmutableState & MutableState,
 *   facets: {
 *     self: import('@agoric/vat-data/src/types').FunctionsMinusContext<typeof selfBehavior>,
 *     helper: import('@agoric/vat-data/src/types').FunctionsMinusContext<typeof helperBehavior>,
 *   },
 * }} MethodContext
 */
```

4) Loss of argument names in types. With the type transformation to remove the first parameter, the names of the other parameters become `args_0` etc. I don't know what to do about this. Maybe there's a trick to do in `FunctionsMinusContext`. I'll research this.

5) Loss of type information within behavior functions. Previously the type of `state` was declared once and shared in each behavior function because the value was shared. Now the context passed in can be _anything_ unless there's an annotation constraining it. I don't know how to solve this without annotating each function.

6) Loss of calback types like `OfferHandler`. This could be solved with a `PlusContext` partner for `MinusContext`. E.g. `{WithContext<OfferHandler>}` below.
https://github.com/Agoric/agoric-sdk/blob/df924b3fbdf1e63b586dba4aaec2f3b988588b44/packages/run-protocol/src/vaultFactory/vault.js#L360

Not all the above has to be solved in this PR. My blocking requests are to:
- inline the function definitions to the behavior objects
- type `facets` in MethodContext

packages/ERTP/src/payment.js
```diff
@@ -9,13 +9,11 @@ import { defineKind } from '@agoric/vat-data';
...
-    () => ({
-      getAllegedBrand: () => brand,
-    }),
-  );
-  return makePayment;
+  const makePayment = defineKind(`${allegedName} payment`, () => ({}), {
+    getAllegedBrand: () => brand,
+  });
+  // XXX the following type cast is meatball surgery to make tsc shut up
+  // somebody who understands this should do it property
```
turadg: mind filing a ticket for this?

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -106,6 +106,13 @@ const validTransitions = {
  * }} MutableState
  */
 
+/**
+ * @typedef {{
+ *   state: ImmutableState & MutableState,
+ *   facets: *,
```
turadg, line 112:
```suggestion
 *   facets: {
 *     self: import('@agoric/vat-data/src/types').FunctionsMinusContext<typeof selfBehavior>,
 *     helper: import('@agoric/vat-data/src/types').FunctionsMinusContext<typeof helperBehavior>,
 *   },
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -135,504 +142,553 @@ const initState = (zcf, manager, idInManager) => {
...
-/** @param {ImmutableState & MutableState} state */
-const constructFromState = state => {
-  /** @type {ImmutableState} */
-  const { idInManager, manager, zcf } = state;
-
-  // #region Computed constants
-  const collateralBrand = manager.getCollateralBrand();
-  const debtBrand = manager.getDebtBrand();
+// #region Computed constants
+const collateralBrand = ({ state }) => state.manager.getCollateralBrand();
```
turadg: please move the behavior functions within the object that will hold them

turadg (changes requested):

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -135,504 +142,553 @@ const initState = (zcf, manager, idInManager) => {
...
-    // After the `await`, we retrieve the vault's allocations again,
-    // so we can compare to the debt limit based on the new values.
-    const collateral = getCollateralAllocated(vaultSeat);
-    const newCollateral = addSubtract(collateral, giveColl, wantColl);
+/**
+ * Adjust principal and collateral (atomically for offer safety)
+ *
+ * @param {MethodContext} context
+ * @param {ZCFSeat} clientSeat
+ */
```
turadg: Needs a return type to prevent circular inference.
```suggestion
  * @returns {string} success message
  */
```

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -411,6 +449,26 @@ export function makeVirtualObjectManager(
     }
   }
 
+  function copyMethods(behaviorTemplate) {
```
warner: Can we rename `behaviorTemplate` here to `behavior`, to match the caller's signature? It's confusing to me that this claims to *accept* `behaviorTemplate` rather than returning it.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -544,61 +632,53 @@ export function makeVirtualObjectManager(
...
-            toExpose[facetName] = facet;
-            toHold.push(facet);
-            facetToCohort.set(facet, toHold);
-          }
-          harden(toExpose);
-          break;
+      unweakable.add(state);
+      if (facetNames === null) {
+        const context = { state };
+        unweakable.add(context);
```
warner, line 646: because I just wondered about it again, and it took me a minute to convince myself it was ok, let's add a one-line comment here
```
# 'context' does not need a linkToCohort because it's held by the bound methods

warner (changes requested):
Two small changes to make, plus the `xsnap-native` submodule change needs to not be in this PR. Rebasing onto trunk might clear that out, but if not grab me and we can figure it out.

packages/run-protocol/src/vaultFactory/vault.js
FUDCo: Turns out tsc wanted `Promise<string>` rather than `string`.  I _think_ this is right...

packages/run-protocol/src/vaultFactory/vault.js
FUDCo: I'm unclear what you mean by "the behavior functions".   Did you mean inline _all_ the behavior functions or just the two that you highlighted in this comment?  Because they weren't all inlined before, by any means.

packages/ERTP/src/payment.js
FUDCo: Done.  #5018

packages/run-protocol/src/vaultFactory/vault.js
FUDCo: I thought this would be a straightforward edit, but it turned out to yield a bunch of follow-on type errors that I have not the first clue how to resolve.  I'm going to put it back the way it was and ask you to toss it in with all the other TS backfilling you've already flagged here.  Possibly also this could be added to the work in #5018?

FUDCo:
How do I expunge the `xsnap-native` nonsense that somehow got sucked into this PR?  Rebasing didn't seem to do the trick.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Ok

packages/run-protocol/src/vaultFactory/vault.js
turadg: I did mean all of them. They weren't all inlined in object creation but they were all nested and that's a property that I want to maintain. They also all had JSDoc before and the new object construction breaks that unless they're inlined.

FUDCo: Done!  Please give it a look.

packages/SwingSet/src/liveslots/virtualObjectManager.js
warner: oh, crud, I got the comment backwards, and it's going to confuse me in its current state

please replace with:

`'context' does not need a linkToCohort because it holds the facets (which hold the cohort)`

sorry

warner (approved):
r+ module that one extra comment change

FUDCo:
> r+ module that one extra comment change

OK, but I think your prior attempt at the comment was also correct.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Perfect, thank you.

turadg (approved):

PR #4963, opened 2022-03-30
chore: remove dckc from * "everything" owner list

dckc:
I'm assigned as an owner of #4962 , but I don't see much reason for that. I think the `*` line at the top was intended to match things that don't match other lines, but I don't think it works that way. I think it tags me as an owner of every PR.

refs: #3382

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,6 @@
 # Default code owners when not specified.
 # We prefer to explicitly specify owners for all subdirectories.
-*                                @warner @dckc @Chris-Hibbert @FUDCo @michaelfig
+*                                @warner @Chris-Hibbert @FUDCo @michaelfig
 
 /*                               @kriskowal @michaelfig @erights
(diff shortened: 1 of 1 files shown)


michaelfig (approved):

michaelfig:
> I'm assigned as an owner of #4962 , but I don't see much reason for that.

That PR changes `packages/vat-data`, which has no code owners by default, so you were tagged.

> I think the `*` line at the top was intended to match things that don't match other lines, but I don't think it works that way. I think it tags me as an owner of every PR.

I can vouch that it doesn't tag you for everything.  Is the right thing to do instead to add a line for `/packages/vat-data` ?

dckc:
> I can vouch that it doesn't tag you for everything.

The fact that you knew that and I didn't tells me it's best that we merge this PR. Thanks for approving.

> Is the right thing to do instead to add a line for `/packages/vat-data` ?

Not instead but perhaps in addition. I leave that to you / others.

PR #4964, opened 2022-03-30
updates to codeowners

turadg:
## Description

- Adds an owner for /vat-data I had added (per https://github.com/Agoric/agoric-sdk/pull/4963#issuecomment-1083296820)
- Adds an owner for /web-components (don't know why this didn't have one)
- Remove the redundant catch-all wildcard (all paths are under `/`)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,11 +9,11 @@
 /scripts/                        @kriskowal @michaelfig @erights
 /yarn.lock                       @mhofman
 
-# Golang @parts @of @the @repo.
+# Golang parts of the repo.
 /go.*                            @JimLarson
 /golang/                         @JimLarson
 
-# Explicit @owners @for @our @packages
+# Explicit owners for our packages
 /packages/ERTP/                  @erights @Chris-Hibbert
 /packages/SwingSet/              @warner @FUDCo
 /packages/access-token/          @kriskowal @michaelfig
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -2,7 +2,6 @@
 # We prefer to explicitly specify owners for all subdirectories.
 *                                @warner @Chris-Hibbert @FUDCo @michaelfig
 
-/*                               @kriskowal @michaelfig @erights
```
michaelfig, line 5: My intent in proposing this line was to catch only the files in the root of the repository.  If it does more than that, I'm happy to remove it, otherwise, let's leave it in.

michaelfig (approved):
Approved modulo a check for the semantics of `/*`.

.github/CODEOWNERS
turadg: Ah. My IDE says that even a file like `/packages/who.txt` is owned by this list.

How about `/*.*`? (EDIT: done)

michaelfig: I think your IDE may be incorrect.  Both [Gitlab](https://docs.gitlab.com/ee/user/project/code_owners.html#example-codeowners-file) and [Github CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file) say something to this effect:

```
# End a path in `/*` to specify Code Owners for every file in
# a directory, but not nested deeper. This code matches
# `docs/index.md` but not `docs/projects/index.md`:
/docs/* @root-docs
```

turadg: Aha, yes. Also corroborated by https://github.com/hmarr/codeowners  

https://github.com/Agoric/agoric-sdk/pull/4966 and uninstalled the buggy extension =)

PR #4969, opened 2022-03-30
ci(mergify): run `master` CI on `mergify/*` instead to fail early

michaelfig:
## Description

Run tests on `mergify/*` temporary branch so that failures prevent merging to master.

### DIFF
--- a/.github/workflows/after-merge.yml
+++ b/.github/workflows/after-merge.yml
@@ -3,7 +3,7 @@
 #  Require branches to be up to date before merging
 on:
   push:
-    branches: [master] # $default-branch
+    branches: [mergify/*]
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


.github/workflows/after-merge.yml
```diff
@@ -3,7 +3,7 @@
 #  Require branches to be up to date before merging
 on:
   push:
-    branches: [master] # $default-branch
+    branches: [merge/*]
```
mhofman:
```suggestion
    branches: [mergify/*]
```

mhofman (changes requested):
Should we keep `master` as well for the cases where mergify is bypassed? I guess the problem is if mergify doesn't do a fast forward?

michaelfig:
> Should we keep `master` as well for the cases where mergify is bypassed? I guess the problem is if mergify doesn't do a fast forward?

If we do that, the CI will run twice.

mhofman (approved):

mhofman:
LGTM once the formatting is fixed

mhofman:
@michaelfig I can't see that any of the jobs ran on this PR before merging. Any idea why?

PR #4971, opened 2022-03-30
feat(run-protocol): debt limit for RUNstake

### LINKED ISSUE #3501: Governance-controllable RUN debt-limit per collateral type in Vault Factory & getRUN contract

turadg:
closes: #3501

## Description

Repeat https://github.com/Agoric/agoric-sdk/pull/4948 for RUNstake

### Security Considerations

New governed param. Malicious manipulation could halt minting.

Downstream has to now provide a debtLimit initial value. I don't think loadgen has RUNstake yet but cc @mhofman  in case.

### Testing Considerations

Test using the RUNstake driver

### DIFF
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -92,8 +92,9 @@ async function waitForPromisesToSettle() {
 }
 
 /**
+ * dL: 1M, lM: 105, iR: 100, lF: 500
+ *
  * @param {Brand} debtBrand
- * @returns dL: 1M, lM: 105, iR: 100, lF: 500
  */
 function defaultParamValues(debtBrand) {
   return harden({
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -72,7 +72,7 @@ const facetHelpers = (zcf, paramManager) => {
    * @template {{}} CF creator facet
    * @param {ERef<CF>} originalCreatorFacet
    * @param {Record<string, (...args: any[]) => any>} governedApis
-   * @returns { ERef<GovernedCreatorFacet<CF>> }
+   * @returns {ERef<GovernedCreatorFacet<CF>>}
    */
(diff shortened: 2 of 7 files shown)


mhofman:
> I don't think loadgen has RUNstake yet but cc @mhofman in case.

I need some context. Loadgen provisions some RUN at the cosmos level. As long as that continues to work, we're fine.

turadg:
> I need some context. 

The change in my PR that broke loadgen was requiring a `debtLimit` value in `addVaultType` as part of setting up contracts. This makes starting a RUNstake contract require that value too. If loadgen is making RUNstake contracts then it'll break after this is in master unless it passes that in.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -92,8 +92,9 @@ async function waitForPromisesToSettle() {
 }
 
 /**
+ * dL: 1M, lM: 105, iR: 100, lF: 500
+ *
  * @param {Brand} debtBrand
- * @returns dL: 1M, lM: 105, iR: 100, lF: 500
```
turadg, line 96: this made eslint warn

Chris-Hibbert (approved):
LGTM
(Please fix breaking DApp, even if CI doesn't require it.)

turadg:
> (Please fix breaking DApp, even if CI doesn't require it.)

Can't, It's a transient network failure in yarn install. I'll rebuild though.

Issue #4974, opened 2022-03-31
Add hash bundle support to Zoe `install`


kriskowal:
## What is the Problem Being Solved?

Zoe currently accepts bundles that contain all of their source, resulting in very large transactions to the chain. Per #4563, we added a new installation method to Zoe, `installBundleID` that allows a bundle to be published to the SwingSet controller without a chain transaction and then install it in Zoe using its “identifier” (tagged consistent hash).

@michaelfig observes that `install` is already sufficiently generic to support an evolving bundle payload type. @dckc observes that this change also avalanches through our documentation https://github.com/Agoric/agoric-sdk/issues/4565#issuecomment-1068459453.

## Description of the Design

So, instead, the plan is to consolidate the behavior of `installBundleID` into `install` and instead introduce a new “hash bundle” type that would be accepted by `install` in addition to “source bundles”, and eventually instead of source bundles https://github.com/Agoric/agoric-sdk/issues/4565.

kriskowal:
Solved in https://github.com/Agoric/agoric-sdk/pull/4975

Issue #4564, opened 2022-02-16
change all dapp deploy scripts to install by hash bundle, not source bundle


warner:
## What is the Problem Being Solved?

After #4563 gives Zoe the ability to `E(zoe).install(bundleID)` (instead of a bundle), and after #4396 gives cosmic-swingset the ability to accept "install bundle" transactions, the next step of the process is to change the deploy scripts in all dapps to use them. Currently, these deploy scripts use `bundleSource()` on the local contract code, then `E(zoe).install(bundle)`, which creates a large (swingset/mailbox) message to deliver the bundle to the chain and install it. That should change to:
* use `bundleSource()` to get both a bundle and it's bundleID (hash)
* use a new client-side facility to construct and send the #4396 install-bundle transaction to the chain
* send `E(zoe).install(bundleID)`

## Description of the Design

We need the client-side facility to create the new transaction types.

## Security Considerations

On the real chain, for the Mainnet-1 timeframe, there will be a governance process to limit the installation of new contracts. The bundleID must be in an approved list before the install-bundle transaction will be accepted. Test chains won't have such a limitation. I *think* the real chain process will look like:
* author develops the contract code, test, finalize, publish source code for review
* author bundles and computes the bundleID
* author submits governance proposal to allow the bundleID to be installed
  * voters will look at the alleged source code (a git tag in some repo) for correctness/safety
  * voters will rebuild the bundle, compute the ID, compare against proposal
* voters approve the proposal
  * ID goes onto the approved-IDs list
  * `E(zoe).install(ID)` is called by the proposal code, waits for bundle to be installed
* author runs deploy script
  * deploy script constructs install-bundle txn, signs, submits
  * txn goes into block, accepted by approved-IDs list, causes `controller.validateAndInstallBundle()`
  * E(zoe).install() wakes up because the #4521 "bundle is now installed" promise fires
  * proposal code adds installation handle to.. the Board?
  * installation complete, available for instantiation

## Test Plan

cc @michaelfig

Tartuffo:
@kriskowal Is this something you could take on?

kriskowal:
I can. I would bump the estimate to 5. I’ll need to get up-to-speed on @warner’s bundle caps API and some kernel layers I haven’t touched yet. The extra time is probably a good investment.

Tartuffo:
Estimate bumped.

Tartuffo:
[...]

Issue #4396, opened 2022-01-26
install-bundle transaction message


warner:
## What is the Problem Being Solved?

#3269 has swingset expose a `controller.installBundle()` API. This ticket is to have cosmic-swingset take advantage of that API, and make the functionality available to external users through a new signed cosmos transaction message type.

Currently, contracts are installed into the chain by sending an object message to Zoe. These messages are by far the largest in our system: most operational messages (make offer, accept invitation, deposit/withdraw) serialize into a few hundred bytes, but contract bundles tend to be 1-2 MB in size. We'd like to move these large bodies out of object messages. Once complete, we'll be able to impose limits or fees on object messages to keep them to a reasonable size.

Cosmic-swingset defines a number of "modules": Go code to which messages can be dispatched. Each signed transaction contains one or more messages, each message has a type, each type is dispatched to a particular module. Currently the swingset module handles a message for mailbox additions, and another for changing the bean price table.

The task of this ticket is to add a third message type to the swingset module. This should be called `installBundle`, and should accept the 1-2MB bundle object as a payload. When executed, it should call `controller.installBundle()` with the payload. This needs to happen in the context of a block, so the kernel state is committed afterwards (which should happen naturally if it gets executed in DeliverTx as usual).

Contract installation needs to be limited in some fashion to protect the storage space of the validators, and to discourage the use of the bundlestore for spam. It might also be a good place to impose the MN-1-vs-2-vs-3 limits on who can install contracts, to guard the not-yet-hardened runtime against malicious contracts (and/or we should build a lock on `zoe~.install`). Even after everything is permissionless, bundle installation should probably change some large (but configurable) fee. A more sophisticated approach might include a deposit and a governance vote, something to require that contract installers are serious and will provide a net positive value to the chain as a whole. #4395 addresses this question.

## Description of the Design

I'll leave the details to @michaelfig , but I think the swingset Go module wants to acquire a new message type, and a handler which delivers something to the JS side, which should then do a `controller.installBundle()`. This should look a lot like the existing "add message to mailbox" pathway.

The fees to be charged need to be configurable.

If #4395 says that access to this message should be restricted, the module will need to implement that restriction.

warner:
[...]

PR #4975, opened 2022-03-31
feat: Support publishing source bundles to chain and installing hash bundles to Zoe

kriskowal:
This stack of changes-in-progress will address:

- #4974
- #4396
- #4564
- #4565

For manual verification, clone `dapp-fungible-faucet`, check out `publish-bundle`, and run these commands in separate sessions:

```
agoric start local-solo
agoric start local-chain
DEBUG=agoric:deploy agoric deploy ./contract/deploy.js
```


packages/agoric-cli/src/deploy.js
```diff
@@ -231,11 +232,12 @@ export { bootPlugin } from ${JSON.stringify(absPath)};
...
           // Use Node.js ESM support if package.json of template says "type":
           // "module".
-          const read = async url => fs.readFile(new URL(url).pathname);
+          const read = async location =>
+            fs.readFile(url.fileURLToPath(location));
           const { packageDescriptorText } =
             await readContainingPackageDescriptor(
               read,
-              `file://${moduleFile}`,
+              url.pathToFileURL(moduleFile),
```
dckc, line 369: something tells me "portable" also means powerful / ambient

packages/publish-bundle/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
dckc: How did you scaffold this new project? Would you please leave some clues in the commit message?

packages/agoric-cli/src/deploy.js
kriskowal: It certainly at least means “sensitive to platform”. The big difference is the treatment of drive letters as absolute paths instead of concatenating them like `c:/users/kris/endo/d:/users/kris/` etc.

packages/publish-bundle/LICENSE
kriskowal: I’ll leave a note. It’s only semi automatic.

packages/agoric-cli/src/deploy.js
```diff
@@ -1,17 +1,23 @@
-/* global process setTimeout setInterval clearInterval */
+/* eslint-env node */
```
kriskowal: Spread the word.

packages/agoric-cli/src/deploy.js
```diff
@@ -89,9 +103,60 @@ export default async function deployMain(progname, rawArgs, powers, opts) {
...
+    const { connectionNumber } = await inquirer.prompt({
+      name: 'connectionNumber',
+      message: 'Please choose a connection to publish to',
+      choices: lookup.keys(),
+    });
+    const connection = lookup.get(connectionNumber);
+    console.log({ connection });
+    return connection;
+  };
+
```
kriskowal: - This is interactive
- Perhaps it should not prompt if there is 1 option
- Perhaps it should err if there are 0 options
- Perhaps interactive mode should be overridden by a command line option, if present, but describing connections requires one of various combinations of flags

Please make your preference known :-) @michaelfig @warner

packages/agoric-cli/src/deploy.js
```diff
@@ -89,9 +103,60 @@ export default async function deployMain(progname, rawArgs, powers, opts) {
...
+    cosmosHelper,
+    pathResolve: path.resolve,
+    writeFile: fs.writeFile,
+  });
+  const publishBundle = makeBundlePublisher({
+    getAccessToken,
+    jsonHttpCall,
+    publishBundleCosmos,
+    getDefaultConnection,
+  });
```
kriskowal: I’m a little dissatisfied that the HTTP guts are inside `@agoric/publish-bundle` but the Cosmos SDK guts are in `@agoric/agoric`. Putting it all here makes it harder to test in `@agoric/solo`, which doesn’t work in `solo/test/test-home.js` yet anyway. Putting all the guts in `@agoric/publish-bundle` would require some duplication or refactoring of the utilities above that are needed to make the bundle publisher.

packages/agoric-cli/src/deploy.js
```diff
@@ -242,11 +307,12 @@ export { bootPlugin } from ${JSON.stringify(absPath)};
 
           // Use Node.js ESM support if package.json of template says "type":
           // "module".
-          const read = async url => fs.readFile(new URL(url).pathname);
+          const read = async location =>
+            fs.readFile(url.fileURLToPath(location));
```
kriskowal, line 365: This is a drive-by portability fix for Windows.

packages/agoric-cli/src/publish.js
```diff
@@ -0,0 +1,65 @@
...
+      const tempFilePath = pathResolve(tempDirPath, 'bundle.json');
+      await writeFile(tempFilePath, `${JSON.stringify(bundle)}\n`);
+      const args = [
+        'tx',
+        'swingset',
+        'install-bundle',
+        '--gas',
+        'auto',
+        '--gas-adjustment',
+        '1.2',
```
kriskowal, line 185: I definitely want to put off parameterizing these.

packages/agoric-cli/src/start.js
```diff
@@ -73,6 +73,7 @@ export default async function startMain(progname, rawArgs, powers, opts) {
 
   const sdkPrefixes = {};
   if (!opts.sdk) {
+    // TODO importMetaResolve should be more reliable.
```
kriskowal: Note to self: remove this comment and track elsewhere.

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -79,6 +80,17 @@ export default function makeBlockManager({
         break;
       }
 
+      case ActionType.INSTALL_BUNDLE: {
+        p = (async () => {
+          const bundle = JSON.parse(action.bundle);
+          harden(bundle);
+          return validateAndInstallBundle(bundle);
+        })().catch(error => {
+          console.error(error);
```
kriskowal, line 89: I imagine this error message would be opaque without some more context. Not sure what would be useful to the end user.

packages/publish-bundle/node-powers.js
```diff
@@ -0,0 +1,70 @@
...
+/* eslint-env node */
+
+const textEncoder = new TextEncoder();
+const textDecoder = new TextDecoder();
+
+/**
+ * @param {Object} powers
+ * @param {typeof import('http')} powers.http
+ */
+export const makeJsonHttpClient = ({ http }) => {
```
kriskowal: I wrote this so it could be faked in tests and to avoid coupling to the powerful `http`.

packages/solo/src/chain-cosmos-sdk.js
```diff
@@ -21,10 +22,9 @@ const console = anylogger('chain-cosmos-sdk');
 // the `agd` tool in our repo is built by 'cd golang/cosmos &&
 // make'. It lives in the build tree along with `bin/ag-solo`, in case there are
 // multiple checkouts of `agoric-sdk`.
-export const HELPER = new URL(
-  '../../../golang/cosmos/build/agd',
-  import.meta.url,
-).pathname;
+export const HELPER = url.fileURLToPath(
+  new URL('../../../golang/cosmos/build/agd', import.meta.url),
+);
```
kriskowal, line 27: Again with a drive-by fix for Windows portability.

packages/solo/src/web.js
```diff
@@ -10,6 +10,10 @@ import morgan from 'morgan';
 
 import { getAccessToken } from '@agoric/access-token';
 
+const textDecoder = new TextDecoder();
+
+const maximumBundleSize = 1024 * 1024 * 128; // 128MB
```
kriskowal, line 13: Arbitrary. Open to suggestions for a less arbitrary mechanism or a less arbitrary magic number.

packages/solo/src/web.js
```diff
@@ -194,6 +200,66 @@ export async function makeHTTPListener(
...
+      return;
+    }
+
+    res.json({
+      ok: true,
+      connections,
+    });
+  });
+
+  app.post('/publish-bundle', async (req, res) => {
```
kriskowal, line 213: This is not tested. It would be tested by `solo/test/test-home.js` if that worked. There’s probably something wrong here with request or response size limits, despite the limit set above. However, this is just for the `http` connection type and isn’t employed when submitting to a Cosmos chain, fake or otherwise.

packages/solo/test/test-home.js
```diff
@@ -1,16 +1,19 @@
 /* global process */
 
-import '@endo/init/pre-bundle-source.js';
-
 // `test.after.always` does not yet seem compatible with ses-ava
 // See https://github.com/endojs/endo/issues/647
 // TODO restore
-// import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava';
```
kriskowal, line 8: Drive-by fix for the `.js` extension, left over before the RESM-to-NESM migration.

packages/solo/test/test-home.js
```diff
@@ -1,16 +1,19 @@
 /* global process */
 
-import '@endo/init/pre-bundle-source.js';
```
kriskowal, line 3: We don’t need this anymore, since Endo isn’t allergic to its own Babel anymore.

packages/solo/test/test-home.js
```diff
@@ -1,16 +1,19 @@
...
-
 // `test.after.always` does not yet seem compatible with ses-ava
 // See https://github.com/endojs/endo/issues/647
 // TODO restore
-// import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava';
+// import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
 import '@agoric/swingset-vat/tools/prepare-test-env.js';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import test from 'ava';
+// TODO
```
kriskowal: Uncomment all the TODOs in this file to reproduce the publishBundle HTTP failure. Otherwise, this should be sufficient to test the HTTP connection type in `@agoric/publish-bundle`, but not sufficient to test for regressions in publishing to a Cosmos chain. I’m not sure where to hook up such a test.

packages/zoe/src/contractFacet/internal-types.js
```diff
@@ -20,7 +20,7 @@
 
 /**
  * @typedef ZCFZygote
- * @property {(bundleOrBundleCap: SourceBundle | BundleCap) => void} evaluateContract
+ * @property {(bundleOrBundleCap: Record<string, string>) => void} evaluateContract
```
kriskowal: `SourceBundle` wasn’t defined any more specifically than this. It probably should be `unknown` since bundles get trafficked as JSON and probably should be verified anywhere they’re bought or sold. I could easily be swayed to go the other way, making these type unions hyper explicit, but that seemed out of scope for this change.

packages/zoe/src/zoeService/zoe.js
```diff
@@ -62,7 +62,8 @@ const makeZoeKit = (
     shutdownZoeVat,
   );
 
-  const getBundleCapFromID = bundleID => E(vatAdminSvc).getBundleCap(bundleID);
+  const getBundleCapFromID = bundleID =>
+    E(vatAdminSvc).waitForBundleCap(bundleID);
```
kriskowal, line 66: @warner This was the most surgical change I could make to ensure that `E(zoe).install(hashBundle)` would wait for the chain to receive the bundle separately. This might be giving us a hint that `getBundleCap` should behave like `waitForBundleCap`, rather than having separate methods, or it might be a hint that `getBundleCapFromID` defined here should be renamed `waitForBundleCapWithID` or some such.

packages/zoe/tools/fakeVatAdmin.js
```diff
@@ -56,6 +56,12 @@ function makeFakeVatAdmin(testContextSetter = undefined, makeRemote = x => x) {
       }
       return Promise.resolve(idToBundleCap.get(bundleID));
     },
+    waitForBundleCap: bundleID => {
+      if (!idToBundleCap.has(bundleID)) {
+        idToBundleCap.init(bundleID, fakeBundleCap());
+      }
+      return Promise.resolve(idToBundleCap.get(bundleID));
+    },
```
kriskowal, line 64: @warner For the fake, I duplicated `getBundleCap`. This might be another hint that I should just change the behavior of the real `getBundleCap`.

packages/agoric-cli/src/publish.js
```diff
@@ -0,0 +1,65 @@
...
+        chainID,
+        '--yes',
+        `@${tempFilePath}`,
+      ];
+      await pspawn(cosmosHelper, args);
+    } finally {
+      removeCallback();
+    }
+  };
+
```
kriskowal, line 202: @michaelfig This would be a good place to wait for an event from the chain telling us whether the INSTALL_BUNDLE action succeeded or failed, probably by laying a trap somewhere keyed on the hash of the bundle we attempted to install. I think I’ll need you looking over my shoulder for that project.

Otherwise, this hangs indefinitely if publishing fails.

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -79,6 +80,17 @@ export default function makeBlockManager({
         break;
       }
 
+      case ActionType.INSTALL_BUNDLE: {
+        p = (async () => {
+          const bundle = JSON.parse(action.bundle);
+          harden(bundle);
+          return validateAndInstallBundle(bundle);
```
kriskowal, line 87: After this would be a good time to emit an “event” as to whether the installation worked or not. My understanding from @michaelfig is that this could arrive in the Agoric deploy script via its WebSocket connection, so `publishBundle` would need a mechanism to connect a listener for a particular bundle hash’s installBundle resolution on chain.

packages/agoric-cli/src/deploy.js
dckc: hm. I kinda like the explicit list.

kriskowal:
Commits are individually reviewable.

packages/agoric-cli/src/deploy.js
kriskowal: Oh, indeed. I must have been cross-eyed when I wrote this because I thought that the environment might provide a better hint than `any` types for all these globals to TypeScript, but this is an _eslint_ pragma so it probably does not.

dckc (commented):
I browsed it; I don't see any problems.

I hope to try it out soonish.

turadg (approved):
Read commit by commit. I don't know much about bundling mechanics but the code LGTM.

One request: the new package.json has older deps than the rest of agoric-sdk; update those so yarn.lock doesn't take on new versions unnecessarily.

dckc:
What would it cost to measure code coverage?

packages/agoric-cli/src/deploy.js
michaelfig: Thanks for asking!  I'd like a flag describing the desired installation target:

* `--target=local` for the local solo's own kernel (targeting spawner)
* `--target=cosmos` for the cosmos-based chain
* `--target=sim` for the sim-chain
* and an alias `--target=agoric` for the union of `cosmos` and `sim` connections

I think the default should be `--target=agoric`.

Then, choose the connection automatically if `--target` is specified and a single connection matches.    Otherwise, if stdin is a TTY, give a prompt for all connections, moving any ones matching `--target` to be first in the list.  The final fallback if stdin is not a TTY is to print an error message and exit.

packages/zoe/src/contractFacet/internal-types.js
turadg: >  I could easily be swayed to go the other way, making these type unions hyper explicit, but that seemed out of scope for this change.

I'd like to help sway you. If it's out of scope, please leave the existing union as this change is a loss of information. (I take it there are still types, declared or not, `SourceBundle` and `BundleCap` and this `bundleOrBundleCap` is still their union.) While `SourceBundle` is not well typed, at least it's named and there's a ticket to improve it:
https://github.com/Agoric/agoric-sdk/blob/454afd34274017175db5ac0513d6cee5b7375a3a/packages/zoe/src/zoeService/utils.d.ts#L4-L6

`BundleCap` needs typing too of course
https://github.com/Agoric/agoric-sdk/blob/454afd34274017175db5ac0513d6cee5b7375a3a/packages/SwingSet/src/types-external.js#L329

golang/cosmos/x/swingset/keeper/msg_server.go
```diff
@@ -155,3 +155,29 @@ func (keeper msgServer) Provision(goCtx context.Context, msg *types.MsgProvision
 
 	return &types.MsgProvisionResponse{}, nil
 }
+
+type installBundleAction struct {
+	*types.MsgInstallBundle
+	Type        string `json:"type"` // INSTALL_BUND
```
michaelfig:
```suggestion
	Type        string `json:"type"` // INSTALL_BUNDLE
```

packages/agoric-cli/src/deploy.js
michaelfig:
```suggestion
/* global process setTimeout setInterval clearInterval */
```

packages/agoric-cli/src/deploy.js
```diff
@@ -89,9 +103,60 @@ export default async function deployMain(progname, rawArgs, powers, opts) {
     1000,
   );
 
-  const retryWebsocket = async () => {
-    const accessToken = await getAccessToken(opts.hostport);
+  const accessToken = await getAccessToken(opts.hostport);
```
michaelfig: For robustness, the `accessToken` should be refetched on every retry of the websocket.

packages/agoric-cli/src/publish.js
michaelfig: The correct way to do that is via the WebSocket, as done for `swingset deliver` transactions in `chain-cosmos-sdk.js`.

packages/cosmic-swingset/src/block-manager.js
michaelfig: Actually, after succeeding in JS, the event should be emitted from Golang, and tracked in deploy-bundle via a Tendermint RPC WebSocket subscription (not the ag-solo WebSocket).

packages/solo/src/web.js
```diff
@@ -1,4 +1,4 @@
-/* global setTimeout clearTimeout setInterval clearInterval */
+/* eslint-env node */
```
michaelfig:
```suggestion
/* global setTimeout clearTimeout setInterval clearInterval */
```

yarn.lock
```diff
@@ -14727,6 +14753,11 @@ prettier-linter-helpers@^1.0.0:
   dependencies:
     fast-diff "^1.1.2"
 
+prettier@^1.19.1:
```
michaelfig: Something in your changes is loading the older prettier.  Please correct.

michaelfig (changes requested):
A few comments, and as @turadg suggested, please don't remove the `SourceBundle` type.

packages/solo/package.json
```diff
@@ -33,7 +33,6 @@
     "@agoric/telemetry": "^0.2.0",
     "@agoric/vats": "^0.9.0",
     "@agoric/wallet": "^0.13.0",
-    "@agoric/xsnap": "^0.12.0",
```
michaelfig, line 36: Why?

michaelfig (commented):
Looks great!  Please fix the lint.

packages/solo/package.json
kriskowal: This is a drive-by. There is no motivation except that xsnap is not uttered anywhere in the solo package, so it appears to be extraneous.

michaelfig (approved):
My phone is not cooperative.  Lgtm.

packages/solo/src/web.js
```diff
@@ -194,6 +198,33 @@ export async function makeHTTPListener(
...
+  });
+
+  app.post('/publish-bundle', async (req, res) => {
+    if (!(await validateAccessToken(req))) {
+      res.json({ ok: false, rej: 'Unauthorized' });
+      return;
+    }
+    try {
+      const bundle = harden(req.body);
+      await validateAndInstallBundle(bundle);
```
warner, line 220: I'm not sure how it should work / already works in the solo machine, but after doing `validateAndInstallBundle`, the host app is supposed to commit the DB changes. Did you run into anything that might do a swingstore commit?

packages/solo/package.json
warner: yeah I think swingset has that dependency for you

golang/cosmos/x/swingset/types/msgs.go
```diff
@@ -183,3 +184,32 @@ func (msg MsgProvision) GetSignBytes() []byte {
...
+
+// Type should return the action
+func (msg MsgInstallBundle) Type() string { return "installBundle" }
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgInstallBundle) ValidateBasic() error {
+	if msg.Submitter.Empty() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "Submitter address cannot be empty")
+	}
+	if len(msg.Bundle) == 0 {
```
warner, line 206: I'm guessing it's not sufficiently convenient+useful to check e.g. JSON-shape, expected keys, base64-ness, or deeper?

I suppose the only benefit of additional checks would be that I might pay for an installation of something bogus, and get it into a block, only to get rejected after that block was executed by validators.

Eventually I think we want to do more extensive checks here to prevent abuse of blocks as a storage mechanism, catch the misshapen junk before it can get added to a block or flooded to all nodes.

packages/cosmic-swingset/src/block-manager.js
warner: yeah, it might be worthwhile logging the bundle ID that failed to install, so a validator operator could correlated it with the installation txn

and/or the "event" we emit with success/failure should include a copy of this `error` message

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -79,6 +80,17 @@ export default function makeBlockManager({
         break;
       }
 
+      case ActionType.INSTALL_BUNDLE: {
```
warner, line 83: BTW my note about "commit swingstore state" on the `solo` side is not a concern here, the block-manager is already in a context where the state will be committed once block processing is complete

packages/zoe/src/zoeService/zoe.js
warner: I think this is fine. I vaguely remember thinking that Zoe's `E(zoe).install()` should wait for the bundle to be installed out-of-band, so that you can only get an installation handle for installations that are actually functional, and aren't waiting for a bundle install that might never happen. And since `E(zoe).install()` is interacting with a wrapper function (which sometimes comes from the fake vatadmin service), that wrapper function should do the waiting, which means that when it calls the real vatadmin service, it should be using `waitForBundleCap` instead of `getBundleCap`.

I've assumed that other (non-zoe, non-agoric) uses/users of swingset and `vatAdminService` may have legitimate reasons to query for a bundleID and want a negative answer (instead of an unbounded pause), which means keeping `getBundleCap` as the form that doesn't wait, and Zoe can use `waitForBundleCap` (which does).

And yeah, at some point renaming this zoe thing `waitForBundleCapFromID` might make sense.

Incidentally, while investigating this function, I found it is named `getBundleCapFromID` here and in `installationStorage.js`, but `getBundleCapForID` in `zoe.js`, which is a bit confusing for code spelunkers.

warner (approved):
Looks good! Only minor comments.

packages/cosmic-swingset/src/block-manager.js
kriskowal: I think this isn’t the right place to annotate the error with the bundle ID, though it already is annotated through many paths, and it wouldn’t hurt to make sure it’s on every pertinent one. I looked into this and ran into a chicken-and-egg with JSON.parse being a necessary step to see the bundle ID here.

Emitting an event here would be nice too, but I’m going to call it a future enhancement.

packages/zoe/src/contractFacet/internal-types.js
kriskowal: I was swayed and the types have been restored.

golang/cosmos/x/swingset/types/msgs.go
kriskowal: I believe keeping the bundle format opaque at this layer keeps with the spirit of Michael’s intention for the type to evolve freely.

PR #4970, opened 2022-03-30
ci(mergify): ensure that all the required checks pass

michaelfig:
## Description

Wait in the Mergify queue until CI passes.

PR #4977, opened 2022-03-31
ci(Mergify): configuration update

arirubinstein:
This change has been made by @arirubinstein from https://mergify.com config editor.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,16 +3,6 @@ queue_rules:
   - name: main
     conditions:
       - base=master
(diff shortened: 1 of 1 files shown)


warner (approved):
I'm jumping in here, but it looks good to me, and might unblock everyone else (including me).

mhofman:
Can we mark this as a revert of https://github.com/Agoric/agoric-sdk/pull/4970 ?

Also, we need to figure out how to by-pass these now existing mergify checks for this PR specifically.

PR #4979, opened 2022-03-31
fix liveslots double-retire bug by using a Set for importsToDrop/Retire

### LINKED ISSUE #4978: liveslots double-retires an import, causing vat-fatal error
While working on vat-upgrade, I discovered a bug in liveslots that can cause a vat-fatal error.

If a target vat has a method which takes an import and wraps it in a virtual object like this:
```js
const makeVir = defineKind('vir', arg => ({ arg }), ({state}) => {})); // holds state but no behavior
function acceptImport(presence) { // import is o-50
  const vir = makeVir(presence); // gets o+10/1
  // drop vir and presence, returning nothing
}
```

What happens is that we drop both the Presence (for the import) and the Representative (for the virtual object) at the same time. The two finalizers run and present both `o-50` and `o+10/1` to `deadSet` in the same `scanForDeadObjects` loop. The virtual object `o+10/1` is processed first, we find that the export and vdata pillars are gone too, and it is deleted. While deleting its data, we decref the import, causing `o-50` to be pushed onto `possiblyDeadSet`, and setting `gcAgain` (because we might also have released some in-RAM objects). Then the loop moves on to `o-50`, finds that it is not referenced, and schedules it for drop+retire by adding it to `importsToDrop`/`importsToRetire`.

We reach the end of the loop and find `doMore` is set (because of `gcAgain`), so we start the loop again. This time `deadSet` has `o-50` (again), which is still just as dead as before, and it gets re-pushed onto `importsToDrop`/`importsToRetire`.

Because these are Arrays and not Sets, we try to do `syscall.dropImports(['o-50', 'o-50'])` (which works for some reason), and then `syscall.retireImports(['o-50', 'o-50'])`, where the first copy deletes the c-list entry and the second copy fails translation. The translation failure is a vat-fatal error, and the vat dies.

In practice, I have to create two virtual objects, each with their own import. I found that if I make N virtual objects, the last Representative created sticks around, and only the first N-1 of them get finalized. There must be some other bug causing the last one to stick around.

The fix is probably to make `importsToDrop` and `importsToRetire` into a Set, and then linearize and sort them before making the syscalls.

I've got a failing test, I'll make a PR to see if that fix actually works.

cc @FUDCo

warner:
Make importsToDrop and importsToRetire into a Set, and then linearize and sort them before making the syscalls, to avoid a vat-fatal error when both an imported Presence and a Representative for a virtual object that contains the import are finalized at the same time.

closes #4978

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -204,7 +204,9 @@ function build(
     // by a remaining pillar, or the pillar which was dropped might be back
     // (e.g., given a new in-memory manifestation).
 
-    const [importsToDrop, importsToRetire, exportsToRetire] = [[], [], []];
+    const importsToDrop = new Set();
+    const importsToRetire = new Set();
+    const exportsToRetire = new Set();
     let doMore;
     do {
       doMore = false;
@@ -234,7 +236,7 @@ function build(
           // i.e., always drop before retiring
           // eslint-disable-next-line no-use-before-define
           if (!vrm.isVrefRecognizable(vref)) {
-            importsToRetire.push(vref);
+            importsToRetire.add(vref);
           }
         }
       }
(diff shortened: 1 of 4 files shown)


FUDCo (approved):
Nice catch!

PR #4980, opened 2022-03-31
refactor(governance): rename "api" to "method"

turadg:
Follow up to #4868 

## Description

What are called `governedApis` are a set of methods that contribute to the API of the contract, not each a new API in itself. This renames those to `governedMethods` and the `voteOnApiInvocation` to `voteOnMethodInvocation`.

You can see from commit history I tried the more concise `voteOnInvocation` but it lost the symmetry with `voteOnParamChange` (vote on noun action).

### Security Considerations

--

Docs updated in global replace. Docs outside this repo not written yet.

### Testing Considerations

--

### DIFF
--- a/packages/governance/src/contractGovernor.js
+++ b/packages/governance/src/contractGovernor.js
@@ -188,16 +188,16 @@ const start = async zcf => {
       );
     }
 
-    // if we aren't governing APIs, voteOnApiInvocation shouldn't be called
+    // if we aren't governing APIs, voteOnInvocation shouldn't be called
     return {
-      voteOnApiInvocation: () => {
+      voteOnInvocation: () => {
         throw Error('api governance not configured');
       },
       createdQuestion: () => false,
     };
   };
 
-  const { voteOnApiInvocation, createdQuestion: createdApiQuestion } =
+  const { voteOnInvocation, createdQuestion: createdApiQuestion } =
     await initApiGovernance();
 
   const validateVoteCounter = async voteCounter => {
@@ -231,7 +231,7 @@ const start = async zcf => {
 
   const creatorFacet = Far('governor creatorFacet', {
     voteOnParamChange,
-    voteOnApiInvocation,
+    voteOnInvocation,
     getCreatorFacet: () => limitedCreatorFacet,
     getInstance: () => governedInstance,
     getPublicFacet: () => governedPF,
--- a/packages/governance/src/contractGovernor.js
+++ b/packages/governance/src/contractGovernor.js
@@ -188,16 +188,16 @@ const start = async zcf => {
       );
     }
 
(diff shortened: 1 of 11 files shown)


Chris-Hibbert:
I'm not completely happy with `voteOnMethodInvocation`, not that I have a better proposal right now.

Part of the design is that the methods to be invoked are segregated into a separate facet, to emphasize that the mechanism can't invoke arbitrary methods of the contract. In this sense, the methods accessible to governance form an API or a particular collection of methods.

packages/governance/src/contractGovernor.js
```diff
@@ -171,33 +171,33 @@ const start = async zcf => {
   // this conditional was extracted so both sides are equally asynchronous
   /** @type {() => Promise<ApiGovernor>} */
   const initApiGovernance = async () => {
-    // @ts-ignore `governedApis` is present on contracts wiht API invocation.
+    // @ts-ignore `governedMethods` is present on contracts wiht API invocation.
```
Chris-Hibbert, line 174:
```suggestion
    // @ts-ignore `governedMethods` is present on contracts that support method invocation.
```

packages/governance/src/contractGovernor.js
```diff
@@ -171,33 +171,33 @@ const start = async zcf => {
...
-        governedApis,
+        governedMethods,
         governedNames,
         timer,
         getUpdatedPoserFacet,
       );
     }
 
-    // if we aren't governing APIs, voteOnApiInvocation shouldn't be called
+    // if we aren't governing APIs, voteOnMethodInvocation shouldn't be called
```
Chris-Hibbert, line 191:
```suggestion
    // if we don't support method invocation, voteOnMethodInvocation shouldn't be called
```

packages/governance/src/contractHelper.js
```diff
@@ -87,11 +87,11 @@ const facetHelpers = (zcf, paramManager) => {
       getLimitedCreatorFacet: () => limitedCreatorFacet,
       // The contract provides a facet with the APIs that can be invoked by
```
Chris-Hibbert, line 88:
```suggestion
      // The contract provides a facet with the methods that can be invoked by
```

Chris-Hibbert (changes requested):
The usages that have been corrected to "method" don't grate on me as much as I expected, but unless we have a rule of thumb about what aspect of the mechanism it makes sense to refer to as "API", I think more renaming is in order. We still refer to API governance, and API questions, etc.


Find and replace many more usages of API for consistency.

rename `governApi.js`.
also search for "ApiGovernance", "invokeApi", "governApi", "governedApi", and "ApiQuestion".
Many things in `question.js` have inconsistent terminology, including the enum `API_INVOCATION`, and `assertApiInvocation`.

turadg:
Contract governance is mothballed

PR #4981, opened 2022-03-31
test: lint Ava tests

### LINKED ISSUE #4941: Lint for Ava tests

Things like `test.only` should fail CI (example https://github.com/Agoric/agoric-sdk/pull/4868#discussion_r836802722  )

## Description of the Design

Adopt https://github.com/avajs/eslint-plugin-ava

## Security Considerations
n/a

## Test Plan
n/a

turadg:
closes: #4941

## Description

Motivation in #4941.

- Adds https://github.com/avajs/eslint-plugin-ava
- Disabled [`use-test`](https://github.com/avajs/eslint-plugin-ava/blob/v13.2.0/docs/rules/use-test.md) due to https://github.com/avajs/eslint-plugin-ava/pull/294#issuecomment-1084648317 
- Selectively disabled [`no-lone-blocks`](https://eslint.org/docs/rules/no-lone-blocks) that started erroring, I suppose because a dep that was updated caught more cases. Alternately we could disable the rule.
- A test that imported from another test required refactoring

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/packages/swing-store/test/exports.js
+++ b/packages/swing-store/test/exports.js
@@ -1,0 +1,99 @@
+import { Buffer } from 'buffer';
+import { makeB0ID } from './util.js';
+
+export const snapshotData = 'snapshot data';
+// this snapHash was computed manually
+export const snapHash =
+  'e7dee7266896538616b630a5da40a90e007726a383e005a9c9c5dd0c2daf9329';
+
+/** @type {import('../src/bundleStore.js').Bundle} */
+export const bundle0 = { moduleFormat: 'nestedEvaluate', source: '1+1' };
+export const bundle0ID = makeB0ID(bundle0);
+
+export function buildData() {
+  // build an export manually
+  const exportData = new Map();
+  const artifacts = new Map();
+
+  // shadow kvStore
+  exportData.set('kv.key1', 'value1');
+
+  // now add artifacts and metadata in pairs
+
+  artifacts.set(`bundle.${bundle0ID}`, JSON.stringify(bundle0));
+  exportData.set(`bundle.${bundle0ID}`, bundle0ID);
+
+  const sbase = { vatID: 'v1', hash: snapHash, inUse: 0 };
+  const tbase = { vatID: 'v1', startPos: 0, isCurrent: 0, incarnation: 1 };
+  const addTS = (key, obj) =>
+    exportData.set(key, JSON.stringify({ ...tbase, ...obj }));
+  const t0hash =
+    '5bee0f44eca02f23eab03703e84ed2647d5d117fed99e1c30a3b424b7f082ab9';
+  const t2hash =
+    '57152efdd7fdf75c03371d2b4f1088d5bf3eae7fe643babce527ff81df38998c';
+  const t5hash =
+    '1947001e78e01bd1e773feb22b4ffc530447373b9de9274d5d5fbda3f23dbf2b';
+  const t8hash =
+    'e6b42c6a3fb94285a93162f25a9fc0145fd4c5bb144917dc572c50ae2d02ee69';
+
+  addTS(`transcript.v1.0`, { incarnation: 0, endPos: 2, hash: t0hash });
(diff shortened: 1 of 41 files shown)


turadg:
Wrinkle with `https://github.com/eslint/eslint/issues/3458` because this makes packages that import `@agoric` eslint config now have to also import `eslint-plugin-ava`.

If we're going to have [an eslint config]((https://www.npmjs.com/package/@agoric/eslint-config)) that gets imported by packages outside this repo then I suppose it should be minimal and very general. Then more specific ones that can be opted into. For example, sdk and dapp. The ava linting would only be on the 'sdk' one. 

So I'll change this PR to
- define that "sdk" config
- use it in all the SDK packages
- make that config lint Ava

packages/cosmic-swingset/test/test-provision-smartwallet.js
```diff
@@ -59,13 +59,7 @@ test.before(async t => {
   t.context = { scenario2, walletTool, pspawnAgd, bootstrapAddr, soloAddr };
 });
 
-// SKIP: struggling with timing issues resulting in one of...
-// Error: cannot grab 250000uist coins: 0uist is smaller than 250000uist: insufficient funds
-// error: code = NotFound desc = account agoric1mhu... not found
-// Sometimes I can get this test to work alone, but not
-// if run with the test above.
-// TODO: https://github.com/Agoric/agoric-sdk/issues/6766
```
Chris-Hibbert: Should this PR 
```
closes: #6766
```
?

packages/solo/src/start.js
```diff
@@ -1,4 +1,5 @@
 // @ts-check
+/* eslint-disable no-lone-blocks */
```
Chris-Hibbert: Is this required because of the switch/case? If so, I think the rule is being too aggressive. I wouldn't want to have to add this caveat every time I use swtch/case. Doesn't our style recommend using braces with switch/case?  (It should, IMHO.)

packages/swing-store/test/exports.js
```diff
@@ -0,0 +1,99 @@
+import { Buffer } from 'buffer';
+import { makeB0ID } from './util.js';
```
Chris-Hibbert, line 2: remark: Good extraction!

Chris-Hibbert (approved):
with caveats, as discussed.

packages/solo/src/start.js
turadg: Remnant of an earlier revision. I'll remove

packages/cosmic-swingset/test/test-provision-smartwallet.js
turadg: sadly, no. It's still failing https://github.com/Agoric/agoric-sdk/actions/runs/5867547993/job/15908602617?pr=4981

I'll revert this change

FUDCo:
I think `no-lone-blocks` is a dumb rule that serves no useful purpose that I can discern.  We should disable it universally.

FUDCo (approved):
This all looks quite reasonable to me.

Chris-Hibbert:
> I think no-lone-blocks is a dumb rule that serves no useful purpose that I can discern. We should disable it universally.

I don't know why it's a good rule to enforce, but I've found that my tastes don't seem to prevail unless I can show why a proposed rule would get in the way. I asked enough questions to convince myself that it doesn't prevent anything that I'm aware of doing.   @FUDCo is there something that you occasionally do that this would grouse about?

FUDCo:
I think the example you already cited of switch cases is compelling by itself, but also: sometimes lone blocks are useful for expository purposes to highlight functionally distinct parts of a function, much as blank lines are used.  Also, I've occasionally found them useful to control what gets captured by a closure -- sometimes optimizations inside the JS engine can result in more stuff getting captured by a closure than you want or than is lexically apparent from reading the code; arguably this latter case is more about working around JS engine issues, but real nonetheless.

PR #4986, opened 2022-03-31
refactor(run-protocol): `const` for `prioritizedVaults`

### LINKED ISSUE #4916: `prioritizedVaults` should be `const`

The `vaultManager` module has a lot of code because `prioritizedVaults` should never be null, but starts that way because of a circularity.

## Description of the Design

With the change to how liquidation works, it should now be possible to refactor to have the property be `const`.

## Security Considerations

Improvement should allow `refreshVaultPriority` to just be passed through to vaults so that there's not access to the collection from vaults. 

## Test Plan

Purerefactor.

dtribble:
closes: #4916

## Description

Change `prioritizedVaults` to be const to eliminate all the asserts that it is assigned, when it should always be set.

### Security Considerations

Having mutable cells for things that should be constant is an unnecessary weakness.

N/A

### Testing Considerations

This is a pure refactor and so existing tests are sufficient.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/prioritizedVaults.js
+++ b/packages/run-protocol/src/vaultFactory/prioritizedVaults.js
@@ -46,8 +46,17 @@ export const currentDebtToCollateral = vault =>
  * @param {() => void} reschedulePriceCheck called when there is a new
  * least-collateralized vault
  */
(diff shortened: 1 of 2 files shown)


turadg (approved):
🧹

mhofman:
In the future, please use `mergify` to merge PRs

PR #4987, opened 2022-03-31
Clear up some `NameAdmin` power access

michaelfig:
## Description

Give holders of `NameAdmin` objects access to the corresponding read-only facet (`NameAdmin.readonly()` returns a `NameHub`).

Also, keep `agoricNamesAdmin` in the core boot powers so that arbitrary names can be added by big-hammer governance.

As a drive-by, use PromiseSpaceOf to DRY out produce/consume types (this was a good change by @dckc that didn't make it into another PR).

### Security Considerations

`NameAdmin`s already have unconditional power to mutate the set of names they control, so this isn't really an increase of scope.

### DIFF
--- a/packages/vats/src/nameHub.js
+++ b/packages/vats/src/nameHub.js
@@ -60,39 +60,71 @@ export const makeNameHubKit = () => {
     },
   });
 
+  /** @type {LegacyMap<string, NameRecord>} */
+  // Legacy because a promiseKit is not a passable
+  const keyToAdminRecord = makeLegacyMap('nameKey');
+
   /** @type {NameAdmin} */
   const nameAdmin = Far('nameAdmin', {
     reserve(key) {
       assert.typeof(key, 'string');
-      if (!keyToRecord.has(key)) {
-        keyToRecord.init(key, makePromiseKit());
+      for (const map of [keyToAdminRecord, keyToRecord]) {
+        if (!map.has(key)) {
+          map.init(key, makePromiseKit());
+        }
       }
     },
-    update(key, newValue) {
+    update(key, newValue, adminValue) {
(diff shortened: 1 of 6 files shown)


packages/vats/src/nameHub.js
```diff
@@ -60,40 +60,73 @@ export const makeNameHubKit = () => {
...
+          map.set(key, record);
+        } else {
+          map.init(key, record);
         }
-        keyToRecord.set(key, record);
-      } else {
-        keyToRecord.init(key, record);
       }
     },
+    async lookupAdmin(...path) {
```
dckc, line 97: A separate `lookupAdmin` method? hm. This structure reminds me of the [Emily](https://www.hpl.hp.com/techreports/2006/HPL-2006-116.pdf) file API (which I have ported to python umpteen times, none of which is handy just now).

Ah... actually, it had `subEdFiles` and `subRdFiles` just like we have `lookup` and `lookupAdmin` here. Fair enough.


```ocaml
type readable = {
isDir : unit -> bool;
exists : unit -> bool;
subRdFiles : unit -> readable list;
subRdFile : string -> readable;
inChannel : unit -> in_channel;
getBytes : unit -> string;
fullPath : unit -> string;
}
type editable = {
ro : readable;
subEdFiles : unit -> editable list;
subEdFile : string -> editable;
outChannel : unit -> out_channel;
setBytes : string -> unit;
mkDir : unit -> unit;
createNewFile : unit -> unit;
delete : unit -> unit;
}
```

dckc (approved):
It's unsettling that we didn't notice dropping `agoricNamesAdmin` from the bootstrap space before.

Thanks for fixing that.

PR #4988, opened 2022-03-31
ci(mergify): The queue name is main

mhofman:
refs: #4984 

## Description

While our default branch name is still `master`, the mergify queue name is `main`.

### Testing Considerations

Still checking if these jobs will trigger,


kriskowal (approved):
That makes sense and I’ll take your word for it.

PR #4852, opened 2022-03-16
IBC-V3 Upgrade and Interchain Account Support

schnetzlerjoe:
## Description

This pull request adds interchain account support (ICS-27) into Agoric as well as upgrades IBC-Go to V3. The main files would be golang/cosmos/app/app.go (adding intertx module for ICS-27 among other things) as well as packages/agoric-cli/src/start.js edits to allow for automatic generation of rly accounts/funding for relaying via IBC using "agoric start local-chain" command.

If not merged can be used as a basis for the team to add the support for it. I am building a protocol on Agoric that requires use of it (via bounty btw!). Thanks!

### Security Considerations

The biggest changes are dependencies mainly IBC-Go V3. That would be the biggest security issue.

### Testing Considerations

There are no tests as it was mainly dependency upgrades but I could add some tests for ICS-27/intertx in /golang/cosmos if needed?

PR #4989, opened 2022-04-01
chore: gitignore output (build, bundles, compiled, dist, _agstate)

turadg:
## Description

`_agstate` is a path installed by agoric-cli that shouldn't get into SCM. It did in https://github.com/Agoric/agoric-sdk/pull/4852/files#diff-55542844d6fa2276f4f3d97f96ecd742b6f20b843aafae1c3dc64120ce0d36c9 so this adds it to gitignore.

It also adds some other build output ignores from the `.prettierignore`. We might want to define a set of wildcards that are shared between ignore files (`.dockerignore` has others).

### Security Considerations

--

--

### Testing Considerations

- [x] check that the matches for the new ignores aren't in SCM


mhofman (commented):
Have we double checked that no existing files would be targeted by these new ignores? I think a lot of the `build` stuff is already targeted by `.gitignore` files in subfolders.

mhofman:
I'm all in favor of aligning our ignore files, but I'd prefer being more surgical (aka have the other ignore files follow the current gitignore files).

Also `_agstate` getting added would mean someone used `agoric` cli from inside the SDK folder. Maybe we could prevent that instead / in addition?

turadg:
> Have we double checked that no existing files would be targeted by these new ignores? I think a lot of the build stuff is already targeted by .gitignore files in subfolders.

What's the harm in repeated ignores?

> I'm all in favor of aligning our ignore files, but I'd prefer being more surgical (aka have the other ignore files follow the current gitignore files).

I'm not following whether there is a requested change. 

> Also _agstate getting added would mean someone used agoric cli from inside the SDK folder. Maybe we could prevent that instead / in addition?

Yes, that happened in the linked ticket. I have no ideas for how to prevent the `_agstate` files being generated.

mhofman:
I think my concern is some folder being legitimately named `build` or `dist` that shouldn't be ignored. Even worse is if there is an existing one part of the checked in tree named like that, which is why I asked if we checked for existing files which would now be considered ignored.

mhofman:
Looks like my concern was founded:

```
$ find . -not -path './.git/*' -not -path './packages/xsnap/moddable/*' -not -path './packages/xsnap/xsnap-native/*' | git check-ignore --stdin -v --no-index
.gitignore:54:bundles   ./packages/SwingSet/test/bundles
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/vat-named.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/bootstrap-bundles.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/vat-disk.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/.gitignore
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/test-bundles-controller.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/test-bundles-kernel.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/vat-install.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/test-bundles.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/nonvat-importable.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/nonvat-install.js
```

.gitignore
```diff
@@ -50,16 +47,19 @@ typings/
...
 .env
 
-# next.js build output
+# build output
+_agstate
 .next
-
-# bundles
+build
+bundles
```
mhofman, line 54: This conflicts with existing files in `packages/SwingSet/test/bundles`

mhofman (dismissed):

turadg:
>  I think my concern is some folder being legitimately named build or dist that shouldn't be ignored. Even worse is if there is an existing one part of the checked in tree named like that,

Agree 100%. That's why I had put that in the test plan. Thanks for finding the `packages/SwingSet/test/bundles` path that I missed and sharing the snippet.

> which is why I asked if we checked for existing files which would now be considered ignored.

The reason I misunderstood this request is I thought the sentence after it, " I think a lot of the build stuff is already targeted by .gitignore files in subfolders"  was meant to be clarify what the request was.

As for what to do about `packages/SwingSet/test/bundles`, fortunately `bundles` is unnecessary because in the other 5 instances of `bundles` are build output their contents are all matched by `bundle-*`.
```
find . -not -path './node_modules/*' -name 'bundles' |git check-ignore --stdin -v --non-matching
.gitignore:54:bundles	./packages/run-protocol/bundles
.gitignore:54:bundles	./packages/zoe/bundles
.gitignore:54:bundles	./packages/spawner/bundles
.gitignore:54:bundles	./packages/pegasus/bundles
::	./packages/SwingSet/test/bundles
.gitignore:54:bundles	./packages/wallet/api/bundles
```

mhofman:
> The reason I misunderstood this request is I thought the sentence after it, " I think a lot of the build stuff is already targeted by .gitignore files in subfolders" was meant to be clarify what the request was.

I think I was indirectly asking for the motivation of moving to less specific ignore patterns, and that I would have preferred making the `.prettierignore` and co more like our collection of `.gitignore` instead of the other way around. I won't die on this hill, and it sure would be easier if there was a way to "import" one ignore file from another to reduce duplication.

mhofman (approved):
I'll approve as-is but would welcome less broad patterns.
I've removed the `automerge` label to give another chance to update. Feel free to add back.

turadg:
After the new build caching it's common to get these untracked file errors:
```
Warning: 7 uncommitted changes
```

Screen Shot 2022-04-18 at 8 07 52 AM

I suppose that's due to the new `bundle-noActionElectorate-meta.json` file, but it does support the idea that people are using directories named `bundles` for build output.

I've updated the PR to ignore `bundles`, with an exception for the extant `packages/SwingSet/test/bundles`.

.gitignore
turadg: I've added an exception to the existing `packages/SwingSet/test/bundles/.gitignore`

warner (approved):
Looks good to me

packages/SwingSet/test/bundles/.gitignore
```diff
@@ -1 +1,2 @@
 disk-bundle
+!.
```
mhofman, line 2: I don't believe that works. The exclusion needs to be collocated with the ignore pattern

turadg: I bet you're thinking of [this in the gitignore docs](https://git-scm.com/docs/gitignore):
> It is not possible to re-include a file if a parent directory of that file is excluded

In this case, it's the directory itself that's being re-included.

I can confirm it works on my machine. If it doesn't work on yours I'll do a fast fix by moving it into the root file.

mhofman: I verified and it doesn't work on my machine with the command I used above

mhofman:
```
node ➜ /workspaces/agoric/agoric-sdk (ta/gitignore_agstate) $ find . -not -path './.git/*' -not -path './packages/xsnap/moddable/*' -not -path './packages/xsnap/xsnap-native/*' | git check-ignore --stdin -v --no-index
.gitignore:54:bundles   ./packages/SwingSet/test/bundles
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/vat-named.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/bootstrap-bundles.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/vat-disk.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/.gitignore
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/test-bundles-controller.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/test-bundles-kernel.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/vat-install.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/test-bundles.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/nonvat-importable.js
.gitignore:54:bundles   ./packages/SwingSet/test/bundles/nonvat-install.js
node ➜ /workspaces/agoric/agoric-sdk (ta/gitignore_agstate) $ cat ./packages/SwingSet/test/bundles/.gitignore 
disk-bundle
!.
```

mhofman:
```
node ➜ /workspaces/agoric/agoric-sdk (ta/gitignore_agstate) $ touch ./packages/SwingSet/test/bundles/foo.js
node ➜ /workspaces/agoric/agoric-sdk (ta/gitignore_agstate) $ git status
On branch ta/gitignore_agstate
Your branch is behind 'origin/ta/gitignore_agstate' by 3 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)

nothing to commit, working tree clean
```

turadg: Ah, I was testing existing files this way:
- `git ls-files packages/SwingSet/test/` showed the files
- edited a file and git status showed it dirty

But I didn't test adding a new file to the directory,
- `touch packages/SwingSet/test/bundles/newfile` doesn't find anything

So I see you're right that the exclusion was a no-op. I'll PR a fix.

PR #4990, opened 2022-04-01
ci(mergify): experiment with merge queue

michaelfig:
## Description

Another experiment to see if enabling both merge queue conditions and Github Actions `push: [ mergify/queue/** ]` at the same time will work.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,6 +3,7 @@ queue_rules:
   - name: main
     conditions:
       - base=master
+      - check-success=mergify-merge-queue-test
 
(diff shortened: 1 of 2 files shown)


.mergify.yml
```diff
@@ -3,6 +3,7 @@ queue_rules:
   - name: main
     conditions:
       - base=master
+      - check-success=dev-canary
```
mhofman: if this doesn't run we'll get everything stuck again, won't we?

mhofman (approved):

PR #4996, opened 2022-04-01
feat(run-protocol): charge penalty for liquidation

### LINKED ISSUE #4775: [Vaults] When Vaults are liquidated, charge a fee

When vaults are liquidated, a fee should be charged and accrued to the reward pool

## Description of the Design

New governance parameter `liquidationPenalty`, defaulting to 10%. Govern per VaultManager.

Proceeds from the penalty will go to the reserve pool. For now hold them in a `penaltyPoolSeat`.

Eventually make a timer for batching emptying the penalty seat into the reserve. The reserve doesn’t benefit from getting it sooner.

## Security Considerations

It's governed, but otherwise not a security issue

## Test Plan

add verification in the existing liquidation tests.

turadg:
closes: #4775

## Description

- governance param LiquidationPenalty on the vault manager (so it’s per collateral type)
- defaults to 10%
- charging logic is in the Vault Factory’s liquidation helper so that the liquidation contract is clean and if governance changes the contract or strategy it can’t change the penalty behavior
- penalty proceeds stored in a penaltyPoolSeat on the VM
- not actually transferred until #5012 so for tests use a getPenaltyAllocation akin to getRewardAllocation

### Security Considerations

Creates a new holder of assets, the `penaltyPoolSeat`.

Backwards compatible, no breaking changes.

### Testing Considerations

New tests

### DIFF
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -28,6 +28,16 @@ const BASIS_POINTS = 10_000n;
 const CENTRAL_DENOM_NAME = 'urun';
 
 /**
+ * @file A collection of productions, each of which declares inputs and outputs.
+ * Each function is passed a set of powers for reading from and writing to the vat config.
+ *
+ * Each of the things they produce they're responsible for resolving or setting.
+ *
+ * In production called by @agoric/vats to bootstrap.
+ * In testing called by differenty.
+ */
+
+/**
  * @param {EconomyBootstrapPowers} powers
  * @param {{ committeeName: string, committeeSize: number }} electorateTerms
  */
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -34,7 +34,6 @@ const CENTRAL_DENOM_NAME = 'urun';
  * Each of the things they produce they're responsible for resolving or setting.
  *
  * In production called by @agoric/vats to bootstrap.
- * In testing called by differenty.
  */
 
 /**
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -271,6 +271,7 @@ export const startVaultFactory = async (
     // XXX the values aren't used. May be addressed by https://github.com/Agoric/agoric-sdk/issues/4861
     debtLimit: AmountMath.make(centralBrand, 0n),
     liquidationMargin: makeRatio(0n, centralBrand),
+    liquidationPenalty: makeRatio(10n, centralBrand, 100n),
     interestRate: makeRatio(0n, centralBrand, BASIS_POINTS),
     loanFee: makeRatio(0n, centralBrand, BASIS_POINTS),
   };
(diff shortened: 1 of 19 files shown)


packages/run-protocol/src/econ-behaviors.js
```diff
@@ -27,6 +27,16 @@ const BASIS_POINTS = 10_000n;
...
 const CENTRAL_DENOM_NAME = 'urun';
 
+/**
+ * @file A collection of productions, each of which declares inputs and outputs.
+ * Each function is passed a set of powers for reading from and writing to the vat config.
+ *
+ * Each of the things they produce they're responsible for resolving or setting.
+ *
+ * In production called by @agoric/vats to bootstrap.
+ * In testing called by differenty.
```
Chris-Hibbert: maybe drop this line?

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -3,11 +3,30 @@
...
 const trace = makeTracer('LIQ');
 
+/**
+ * @param {Amount<'nat'>} proceeds
+ * @param {Amount<'nat'>} debt - after incurring penalty
+ * @param {Amount<'nat'>} penaltyPortion
+ */
+const partitionProceeds = (proceeds, debt, penaltyPortion) => {
+  const isShortfall = !AmountMath.isGTE(proceeds, debt);
+  const debtPaid = isShortfall ? proceeds : debt;
```
Chris-Hibbert:
```suggestion
  const debtPaid = AmountMath.min(proceeds, debt);
```

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -31,10 +52,16 @@ const liquidate = async (
...
   collateralBrand,
+  penaltyPoolSeat,
+  penaltyRate,
 ) => {
   innerVault.liquidating();
-  const debt = innerVault.getCurrentDebt();
-  const { brand: runBrand } = debt;
+
+  const debtBeforePenalty = innerVault.getCurrentDebt();
+  const penalty = floorMultiplyBy(debtBeforePenalty, penaltyRate);
```
Chris-Hibbert: The contract usually gets the benefit of rounding:
```suggestion
  const penalty = ceilMultiplyBy(debtBeforePenalty, penaltyRate);
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -11,7 +11,7 @@ import { AMM_INSTANCE } from './params.js';
 const makeLiquidityKeyword = keyword => `${keyword}Liquidity`;
 
 /**
- * Collateral Reserve holds onto Collateral for the RUN protocol, and can
+ * Asset Reserve holds onto assets for the RUN protocol, and can
```
Chris-Hibbert: renaming the collateralReserve seems unrelated to the point of this PR.

Chris-Hibbert (approved):
Looks good.

packages/run-protocol/src/reserve/assetReserve.js
turadg: moved to https://github.com/Agoric/agoric-sdk/pull/5028

packages/run-protocol/src/vaultFactory/liquidation.js
turadg: 👍 I suppose `isShortfall` was defined for tracing but I already took it out of the trace because `debtPaid` implies it.

PR #5001, opened 2022-04-03
test(runStake): use test driver rather than data file

dtribble:
test(runStake): use test driver rather than 

## Description

* convert the test cases in the data file to use test driver explicitly
* step converts to a call on the driver.

### Testing Considerations

This enables normal developer tool support (operation completion and Intellisense) and stack traces test failure; e.g., 
```
  Difference:

    {
      brand: Object @Alleged: RUN brand {},
  -   value: 194967256n,
  +   value: 196000000n,
    }

  › approxEqual (packages/run-protocol/test/runStake/test-runStake.js:477:7)
  › Object.checkRUNDebt (packages/run-protocol/test/runStake/test-runStake.js:713:7)
  › async packages/run-protocol/test/runStake/test-runStake.js:942:3

  ─
  ```


dckc (commented):
Looks consistent with a suggestion you made earlier.

At a glance, I don't see any critical problems. I can take a closer look if we decide this addresses a MN-1 issue.

Commenting out code rather than deleting it is a little odd, but consistent with marking this as DRAFT.

dckc:
This seems appropriate to include in #3788

dckc (approved):
I checked it out and verified that messing with the tests breaks in the expected way.

Good stuff.

PR #5007, opened 2022-04-05
fix: implement store snapshot

erights:
For the virtual and durable stores, `snapshot` was not yet implemented.


FUDCo (approved):
Credible.

PR #5008, opened 2022-04-05
fix: Make payments recoverable

erights:
Associate with each purse a recover set of live payments that can be recovered in emergencies, such as when a payment seems lost or stuck. Recovering a payment is like cancelling an outstanding check.

Withdrawing from a purse puts the resulting payment in that purse's recovery set.

### DIFF
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -50,10 +50,34 @@ export const makePaymentLedger = (
 
   /** @type {WeakMapStore<Payment, Amount>} */
   const paymentLedger = makeScalarBigWeakMapStore('payment');
-  /** @type {WeakMapStore<Payment, SetStore<Payment>>} */
+
+  /**
+   * A withdrawn live payment is associated with the recovery set of
+   * the purse it was withdrawn from. Let's call these "recoverable"
+   * payments. All recoverable payments are live, but not all live
+   * payments are recoverable. We do the bookkeeping for payment recovery
+   * with this weakmap from recoverable payments to the recovery set they are
+   * in.
+   * A bunch of interesting invariants here:
+   *    * Every payment that is a key in the outer `paymentRecoverySets`
+   *      weakMap is also in the recovery set indexed by that payment.
+   *    * Implied by the above but worth stating: the payment is only
+   *      in at most one recovery set.
+   *    * A recovery set only contains such payments.
+   *    * Every purse is associated with exactly one recovery set unique to
+   *      it.
+   *    * A purse's recovery set only contains payments withdrawn from
+   *      that purse and not yet consumed.
+   *
+   * @type {WeakMapStore<Payment, SetStore<Payment>>}
+   */
   const paymentRecoverySets = makeScalarBigWeakMapStore('payment-recovery');
 
   /**
+   * To maintain the invariants listed in the `paymentRecoverySets` comment,
+   * `initPayment` should contain the only
+   * call to `paymentLedger.init`.
+   *
    * @param {Payment} payment
    * @param {Amount} amount
    * @param {SetStore<Payment>} [optRecoverySet]
@@ -66,6 +90,13 @@ export const makePaymentLedger = (
     paymentLedger.init(payment, amount);
   };
 
+  /**
+   * To maintain the invariants listed in the `paymentRecoverySets` comment,
+   * `deletePayment` should contain the only
+   * call to `paymentLedger.delete`.
+   *
+   * @param {Payment} payment
+   */
   const deletePayment = payment => {
     paymentLedger.delete(payment);
     if (paymentRecoverySets.has(payment)) {
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -48,8 +48,36 @@ export const makePaymentLedger = (
     throw reason;
   };
 
-  /** @type {WeakStore<Payment, Amount>} */
+  /** @type {WeakMapStore<Payment, Amount>} */
   const paymentLedger = makeScalarBigWeakMapStore('payment');
+  /** @type {WeakMapStore<Payment, SetStore<Payment>>} */
+  const paymentRecoverySets = makeScalarBigWeakMapStore('payment-recovery');
+
+  /**
+   * @param {Payment} payment
+   * @param {Amount} amount
+   * @param {SetStore<Payment>} [optRecoverySet]
+   */
+  const initPayment = (payment, amount, optRecoverySet = undefined) => {
+    if (optRecoverySet !== undefined) {
+      if (!optRecoverySet.has(payment)) {
+        optRecoverySet.add(payment);
+      }
+      paymentRecoverySets.init(payment, optRecoverySet);
+    }
+    paymentLedger.init(payment, amount);
+  };
+
+  const deletePayment = payment => {
+    paymentLedger.delete(payment);
+    if (paymentRecoverySets.has(payment)) {
+      const recoverySet = paymentRecoverySets.get(payment);
+      paymentRecoverySets.delete(payment);
+      if (recoverySet.has(payment)) {
+        recoverySet.delete(payment);
+      }
+    }
+  };
 
   /** @type {(left: Amount, right: Amount) => Amount } */
   const add = (left, right) => AmountMath.add(left, right, brand);
@@ -103,9 +131,14 @@ export const makePaymentLedger = (
    *
    * @param {Payment[]} payments
    * @param {Amount[]} newPaymentBalances
+   * @param {SetStore<Payment>} [optRecoverySet]
    * @returns {Payment[]}
    */
-  const moveAssets = (payments, newPaymentBalances) => {
+  const moveAssets = (
+    payments,
+    newPaymentBalances,
+    optRecoverySet = undefined,
+  ) => {
     assertCopyArray(payments, 'payments');
     assertCopyArray(newPaymentBalances, 'newPaymentBalances');
 
@@ -141,11 +174,11 @@ export const makePaymentLedger = (
     let newPayments;
     try {
       // COMMIT POINT
-      payments.forEach(payment => paymentLedger.delete(payment));
+      payments.forEach(payment => deletePayment(payment));
 
       newPayments = newPaymentBalances.map(balance => {
         const newPayment = makePayment();
-        paymentLedger.init(newPayment, balance);
+        initPayment(newPayment, balance, optRecoverySet);
         return newPayment;
       });
     } catch (err) {
@@ -178,7 +211,7 @@ export const makePaymentLedger = (
       assertAmountConsistent(paymentBalance, optAmountShape);
       try {
         // COMMIT POINT.
-        paymentLedger.delete(payment);
+        deletePayment(payment);
       } catch (err) {
         shutdownLedgerWithFailure(err);
         throw err;
@@ -187,8 +220,11 @@ export const makePaymentLedger = (
     });
   };
 
-  /** @type {IssuerClaim} */
-  const claim = (paymentP, optAmountShape = undefined) => {
+  const claimInternal = (
+    paymentP,
+    optAmountShape = undefined,
+    optRecoverSet = undefined,
+  ) => {
     return E.when(paymentP, srcPayment => {
       assertLivePayment(srcPayment);
       const srcPaymentBalance = paymentLedger.get(srcPayment);
@@ -197,10 +233,15 @@ export const makePaymentLedger = (
       const [payment] = moveAssets(
         harden([srcPayment]),
         harden([srcPaymentBalance]),
+        optRecoverSet,
       );
       return payment;
     });
   };
+
+  /** @type {IssuerClaim} */
+  const claim = (paymentP, optAmountShape = undefined) =>
+    claimInternal(paymentP, optAmountShape, undefined);
 
   /** @type {IssuerCombine} */
   const combine = (fromPaymentsPArray, optTotalAmount = undefined) => {
@@ -217,6 +258,7 @@ export const makePaymentLedger = (
       const [payment] = moveAssets(
         harden(fromPaymentsArray),
         harden([totalPaymentsBalance]),
+        undefined,
       );
       return payment;
     });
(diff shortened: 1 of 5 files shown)


packages/ERTP/src/paymentLedger.js
```diff
@@ -48,8 +48,36 @@ export const makePaymentLedger = (
...
+  /**
+   * @param {Payment} payment
+   * @param {Amount} amount
+   * @param {SetStore<Payment>} [optRecoverySet]
+   */
+  const initPayment = (payment, amount, optRecoverySet = undefined) => {
+    if (optRecoverySet !== undefined) {
+      if (!optRecoverySet.has(payment)) {
+        optRecoverySet.add(payment);
+      }
```
michaelfig: Is protecting this `.add` with an if statement useful?  It looks like just a distraction.

packages/ERTP/src/paymentLedger.js
```diff
@@ -48,8 +48,36 @@ export const makePaymentLedger = (
...
+  };
+
+  const deletePayment = payment => {
+    paymentLedger.delete(payment);
+    if (paymentRecoverySets.has(payment)) {
+      const recoverySet = paymentRecoverySets.get(payment);
+      paymentRecoverySets.delete(payment);
+      if (recoverySet.has(payment)) {
+        recoverySet.delete(payment);
+      }
```
michaelfig: Same here?

packages/ERTP/src/types.js
```diff
@@ -350,6 +350,24 @@
...
+ * @property {IssuerClaim} claim
+ * Like `issuer.claim`, but also associates the new payment with this purse's
+ * recovery set, so the payment can be recovered in emergencies.
+ *
+ * @property {() => CopySet<Payment>} getRecoverySet
+ * The set of payments associated with this purse that are still live. These
+ * are the payments that can still be recovered in emergencies by, for example,
+ * depositing into this purse. Such a deposit action is like canceling an
+ * outstanding check because you're tired of waiting for it. Once your
+ * cancellation is acknowledged, ypu can spend the assets at stake on other
```
michaelfig:
```suggestion
 * cancellation is acknowledged, you can spend the assets at stake on other
```

packages/ERTP/test/unitTests/test-recovery.js
```diff
@@ -0,0 +1,42 @@
...
+  const precious = num => AmountMath.make(brand, num);
+  const payment1 = mint.mintPayment(precious(37n));
+  const payment2 = mint.mintPayment(precious(41n));
+  const alicePurse = issuer.makeEmptyPurse();
+  const bobPurse = issuer.makeEmptyPurse();
+
+  t.assert(keyEQ(alicePurse.getRecoverySet(), emptySet));
+  alicePurse.deposit(payment1);
+  bobPurse.deposit(payment2);
+  t.assert(keyEQ(alicePurse.getRecoverySet(), emptySet));
```
michaelfig: Do you also want to assert emptySet on `bobPurse` before and after depositing `payment2`?

michaelfig (commented):
Looking good so far.  Let me know when you're ready for a final review (especially if you plan on doing something about `virtual-purse.js`, which notably the `vbank` relies on).

erights:
> Looking good so far. Let me know when you're ready for a final review (especially if you plan on doing something about virtual-purse.js, which notably the vbank relies on).

I took a look at vbank again and decided not to do it in this PR. Would you like to do vbank after this PR?

With vbank postponed, this is now ready for review. Thanks.

michaelfig:
> With vbank postponed, this is now ready for review. Thanks.

It seems I was only tagged as a reviewer because of my codeowner status for `virtual-purse.js`.  I would prefer if somebody else took responsibility for this review, as I may be lacking some necessary context.

erights:
With @Chris-Hibbert OOO, @dtribble can you review? Thanks.

packages/ERTP/src/types.js
```diff
@@ -350,6 +350,24 @@
  *
  * @property {(amount: Amount) => Payment} withdraw
  * Withdraw amount from this purse into a new Payment.
+ *
+ * @property {IssuerClaim} claim
+ * Like `issuer.claim`, but also associates the new payment with this purse's
+ * recovery set, so the payment can be recovered in emergencies.
+ *
+ * @property {() => CopySet<Payment>} getRecoverySet
```
dtribble, line 354: this does make purses non-robust servers in that any holder of a purse can impair the use by any other holder of a purse. I think that's OK , but it is a change.

packages/vats/src/virtual-purse.js
```diff
@@ -143,6 +145,18 @@ function makeVirtualPurse(vpc, kit) {
       });
       return pmt;
     },
+    claim: (_paymentP, _optAmountShape) => {
+      // TODO implement
+      assert.fail(X`virtual purses do not yet implement claim`);
```
dtribble: We shouldn't commit with this.  At the very least there would need to be an issue, but what are the design issues for this?

packages/ERTP/src/paymentLedger.js
```diff
@@ -48,8 +48,32 @@ export const makePaymentLedger = (
     throw reason;
   };
 
-  /** @type {WeakStore<Payment, Amount>} */
+  /** @type {WeakMapStore<Payment, Amount>} */
   const paymentLedger = makeScalarBigWeakMapStore('payment');
+  /** @type {WeakMapStore<Payment, SetStore<Payment>>} */
```
dtribble: This needs documentation here. It's too clever to leave without explanation.

packages/ERTP/src/paymentLedger.js
```diff
@@ -197,11 +229,16 @@ export const makePaymentLedger = (
...
         harden([srcPayment]),
         harden([srcPaymentBalance]),
+        optRecoverSet,
       );
       return payment;
     });
   };
 
+  /** @type {IssuerClaim} */
+  const claim = (paymentP, optAmountShape = undefined) =>
```
dtribble: Keep this code somewhere, but all the claim operations in our code are in tests. Because of Zoe, all real usage ends up depositing into Zoe escrow. So all the `claim` and related operations can be removed, leaving just the recovery logic around the `withdraw` operation.

packages/ERTP/src/purse.js
```diff
@@ -23,35 +26,59 @@ export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
...
             amount,
+            state.recoverySet,
           ),
         getCurrentAmount: ({ state }) => state.currentBalance,
         getCurrentAmountNotifier: ({ state }) => state.balanceNotifier,
         getAllegedBrand: () => brand,
         // eslint-disable-next-line no-use-before-define
         getDepositFacet: ({ facets }) => facets.depositFacet,
+
+        claim: ({ state }, paymentP, optAmountShape) =>
```
dtribble: As above, delete

packages/ERTP/src/types.js
```diff
@@ -350,6 +350,24 @@
  *
  * @property {(amount: Amount) => Payment} withdraw
  * Withdraw amount from this purse into a new Payment.
+ *
+ * @property {IssuerClaim} claim
```
dtribble: Delete.

packages/vats/src/virtual-purse.js
```diff
@@ -143,6 +145,18 @@ function makeVirtualPurse(vpc, kit) {
       });
       return pmt;
     },
+    claim: (_paymentP, _optAmountShape) => {
```
dtribble: delete

packages/vats/src/virtual-purse.js
```diff
@@ -143,6 +145,18 @@ function makeVirtualPurse(vpc, kit) {
...
+      // TODO implement
+      assert.fail(X`virtual purses do not yet implement claim`);
+    },
+    getRecoverySet: () => {
```
dtribble: This should be implemented. It looks to me like it's exactly the same code; all the recovery is on the JS side.

dtribble (changes requested):
Key things to complete:
- remove the `claim operations`
- add an implementation comment
- provide the implementation for virtual purses too

packages/ERTP/src/types.js
erights: (we discussed and agreed it is ok. Because anyone holding a purse can withdraw everything from it, we encourage a programming pattern where a single purse is never help my multiple mutually suspicious parties, or even by multiple non-coordinated parties. Within that framing, we are not introducing any new vulnerabilities.)

packages/ERTP/src/paymentLedger.js
erights: Git keeps the code. I really doubt we will need it again. It is sufficiently easy to reproduce anyway that I'm just "deleting" it.

packages/vats/src/virtual-purse.js
erights: Since you have a separate review comment on virtual-purse `getRecoverySet` below and I've already deleted `claim`, I'm resolving this.

packages/vats/src/virtual-purse.js
erights: What I'm confused about here is whether a vpurse
   * a vpurse should forward to an underlying ERTP purse which implements this logic, or 
   * should reimplement that on its own.
   * one case of vpurse should forward and the other case should reimplement.

I suspect the third bullet is the right answer, but I'm not well enough oriented to be confident.

@michaelfig can you do the vpurse adjustment and I'll review?

@dtribble if yes, would you rather @michaelfig do that in a separate PR or as an enhancement to this PR?

Either way, when this PR is ready, someone else should merge it if I don't seem to be available.

erights: Thanks

michaelfig: > - should reimplement that on its own

In our uses of this `virtual-purse.js`, each vpurse should implement its own recoverySet, so that recoverAll calls vpurse.deposit (which is where the vpurse magic happens).  I think the right thing is to make it part of the current PR.  I can attempt this if you don't have bandwidth for now.

Note that vpurses are not yet Virtual nor Durable objects, and even without that restriction, are somewhat simpler than `@agoric/ertp` purses.  That makes things easier.

erights: >  I can attempt this if you don't have bandwidth for now.

Yes please! Much appreciated!

michaelfig: Oh, when implementing I found that this is actually a case of "forwarding to an ERTP purse which implements this logic."  But it's still pretty simple.

packages/vats/src/virtual-purse.js
```diff
@@ -46,22 +46,47 @@ import '@agoric/notifier/exported.js';
...
+
+  /**
+   * Claim a payment for recovery via our recoveryPurse.
+   *
+   * @param {ERef<Payment>} payment
+   * @param {Amount} [optAmountShape]
+   */
+  const recoverableClaim = async (payment, optAmountShape) => {
+    const pmt = await payment;
+    const amt = await E(recoveryPurse).deposit(pmt, optAmountShape);
```
michaelfig, line 61: Ironically, I could have used `E(recoveryPurse).claim` here.

erights:
@michaelfig , please review my changes to your changes at https://github.com/Agoric/agoric-sdk/pull/5008/commits/5d71a1cdea4bc8bf4c4e0de07488f520429eaea0

erights (commented):
@michaelfig FWIW, your changes at https://github.com/Agoric/agoric-sdk/pull/5008/commits/b337db7d050faed61311a329bf45361e11505eb9 LGTM modulo my refinement in the next commit.

packages/ERTP/src/purse.js
```diff
@@ -23,35 +26,57 @@ export const makePurseMaker = (allegedName, assetKind, brand, purseMethods) => {
...
         getAllegedBrand: () => brand,
         // eslint-disable-next-line no-use-before-define
         getDepositFacet: ({ facets }) => facets.depositFacet,
+
+        getRecoverySet: ({ state }) => state.recoverySet.snapshot(),
+        recoverAll: ({ state, facets }) => {
+          let amount = AmountMath.makeEmpty(brand, assetKind);
+          for (const payment of state.recoverySet.keys()) {
+            // This does cause deletions from the set while iterating,
+            // but this special case is allowed.
```
dtribble, line 70: ha!  Glad to see that show up. It's also noteworthy that the deletion is indirect because of the operation invoked. It would be more complicated if the iteration had to decide whether to delete.

packages/ERTP/src/paymentLedger.js
```diff
@@ -48,9 +48,64 @@ export const makePaymentLedger = (
...
   const paymentLedger = makeScalarBigWeakMapStore('payment');
 
+  /**
+   * A withdrawn live payment is associated with the recovery set of
+   * the purse it was withdrawn from. Let's call these "recoverable"
+   * payments. All recoverable payments are live, but not all live
+   * payments are recoverable. We do the bookkeeping for payment recovery
+   * with this weakmap from recoverable payments to the recovery set they are
+   * in.
+   * A bunch of interesting invariants here:
```
dtribble, line 61: love it.

packages/vats/src/virtual-purse.js
```diff
@@ -46,22 +46,47 @@ import '@agoric/notifier/exported.js';
...
+  /**
+   * @returns {{
+   *   retain: (pmt: Payment, optAmountShape?: Pattern) => Promise<Amount>,
+   *   redeem: (amt: Amount) => Promise<Payment>,
+   * }}
+   */
+  const makeRetainRedeem = () => {
+    if (mint) {
+      const retain = (payment, optAmountShape) =>
+        E(issuer).burn(recoverableClaim(payment), optAmountShape);
```
dtribble: If you are just going to burn, is there any reason to have a transition to a recoverable claim?

packages/vats/src/virtual-purse.js
```diff
@@ -46,22 +46,47 @@ import '@agoric/notifier/exported.js';
...
+      const redeem = amount => recoverableClaim(E(mint).mintPayment(amount));
```
dtribble, line 75: This is an interesting reason to have a `claim` method on purse:  minting!

dtribble: (could be added separately.

dtribble (approved):
A few comments and questions on the virtual-purse, but it all looks correct.

packages/vats/src/virtual-purse.js
michaelfig: Yeah, I was thinking about this last night.  No need for this recoverableClaim, nor for the other retain (in which we just deposit to the escrow purse).

PR #5009, opened 2022-04-05
refactor(run-protocol)!: remove `interestRate` and `liquidationRatio` from vault notifier

### LINKED ISSUE #4685: remove vault manager notifications from vault's notifications

https://github.com/Agoric/agoric-sdk/issues/4540 removed the cascading but `interestRate` and `liquidationRatio` that vaultNotifier returns still come from the vault manager.

## Description of the Design

Remove those too.

Ensure tests still verify the values, but from governance subscription.

turadg:
closes: #4685

## Description

Neither were used anywhere.

When needed, `interestRate` is now on `AssetNotifier`.

### Security Considerations

--

Breaking change, denoted in conventionalcommit syntax.

### Testing Considerations

- [x] test for asset notifier on a new Collateral Manager ([exists](0f11c9c87c130277c439d2fa7cf4aa3bb844561d))
- [ ] test improvements so dapp-treasury integration test fails on this breaking change (probably https://github.com/Agoric/dapp-treasury/issues/97 )

### DIFF
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -979,7 +979,6 @@ test('interest on multiple vaults', async t => {
     ),
     run.make(3200n + bobAddedDebt),
   );
-  t.deepEqual(bobUpdate.value.interestRate, rates.interestRate);
 
   // 236 is the initial fee. Interest is ~4n/week
   const aliceAddedDebt = 236n + 4n;
@@ -997,7 +996,6 @@ test('interest on multiple vaults', async t => {
     debt: run.make(4935n),
     interest: makeRatio(100n, run.brand, 100n),
   });
-  t.deepEqual(aliceUpdate.value.interestRate, rates.interestRate);
 
   const rewardAllocation = await E(vaultFactory).getRewardAllocation();
   const rewardRunCount = aliceAddedDebt + bobAddedDebt;
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -980,10 +980,6 @@ test('interest on multiple vaults', async t => {
     run.make(3200n + bobAddedDebt),
   );
   t.deepEqual(bobUpdate.value.interestRate, rates.interestRate);
-  t.deepEqual(
-    bobUpdate.value.liquidationRatio,
-    makeRatio(105n, run.brand, 100n),
-  );
 
   // 236 is the initial fee. Interest is ~4n/week
   const aliceAddedDebt = 236n + 4n;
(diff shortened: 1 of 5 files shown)


packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -969,11 +969,6 @@ test('interest on multiple vaults', async t => {
     ),
     AmountMath.make(runBrand, 3200n + bobAddedDebt),
   );
-  t.deepEqual(bobUpdate.value.interestRate, rates.interestRate);
```
dtribble, line 982: Are these getting tested in the vaultManager's notifier?

samsiegart:
The UI uses both of these currently.

>When needed, interestRate is now on AssetNotifier.

Sounds good.

How do we show the liquidation ratio now? I suppose we can get that from `getCollaterals()` and check from the vault's collateral. Can the liquidation ratio and interest rate change after you've opened a vault/could they be different between new vaults and old vaults of the same collateral type?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: No, nothing in agoric-sdk is reading these removed values. I'll start some new tests.

Chris-Hibbert: I think Dean is asking a different question. These values are now superfluous in `vault.js` because all the relevant work is done in `vaultManager.js`. I checked, and the values are visible to the UI from the versions that would remain in `vaultManager.js`.  The question is whether there are tests corresponding to the one removed here that includes verifying the new state of the values in `vaultManager`'s updates.

Chris-Hibbert (approved):
LGTM

The question about whether the corresponding update is verified in vaultManager should be answered here, but if the answer is no, the fix can come later. (so I guess I'm asking for either a fix here or an issue #.)

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: > The question is whether there are tests corresponding to the one removed here that includes verifying the new state of the values in vaultManager's updates.

Yep, that was my understanding. The answer to that question is "no" and that's why I'm going to "start some new tests".

turadg: > The question is whether there are tests corresponding to the one removed here that includes verifying the new state of the values in vaultManager's updates.

I said "no" but that was wrong. Here the assetNotifier from vaultManager is read: https://github.com/Agoric/agoric-sdk/blame/540978c997e30c25905ea1fb837ae701db6f3536/packages/run-protocol/test/swingsetTests/governance/vat-alice.js#L47-L63

turadg:
Next steps:
- drop the need for integration tests, not happening for MN-1
- don't move `interestRate` and `liquidationRatio` to the asset notifier since they're already exposed as governed params. consumers can merge subscriptions if they need a single stream.
- move `liquidatorInstance` to governance

PR #5011, opened 2022-04-05
test(run-protocol): inline golden log

turadg:
_no ticket_

## Description

Ava's snapshot testing creates two files. One is human readable but the source of truth is binary and effectively uneditable. This prevents a workflow in which the snapshot is manually updated. While refactoring that workflow is valuable so that you can update parts that you expect to change 
but leave alone the other parts that should continue to fail.

So for now we approximate Jest's inline snapshots with a manual comparison to string.

In the future, I hope Ava will support such a workflow: https://github.com/avajs/ava/issues/1806#issuecomment-1088953027B

### Security Considerations

--

If anything, clearer logic for these tests.

### Testing Considerations

--


Chris-Hibbert (approved):
Awesome!  thanks.

PR #5033, opened 2022-04-06
chore: bump deps

turadg:
## Description

Bump these deps reported by `yarn outdated`:
- better-sqlite3
- c8
- node-lmdb
- prettier
- typescript
- inquirer

Also run `yarn-deduplicate` to bump transitive deps.


packages/SwingSet/package.json
```diff
@@ -83,4 +83,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
+}
```
FUDCo: The missing terminal newline seems to be a problem in every file in this PR.

FUDCo (dismissed):
Every edited `package.json` file lost its terminal newline and that needs to be corrected.

Other than that this seems fine.

packages/SwingSet/package.json
turadg: I didn't do this so I think it was `yarn`.

If this is whitespace policy a requirement I ask that CI fail. We shouldn't have to discuss whitespace on PRs.

I'm happy to make a PR for that, applying Prettier to .json files.

dckc: > We shouldn't have to discuss whitespace on PRs.

Well, maybe ideally not. But here we are.

I concur; please restore the newlines.

turadg: I was wrong about `yarn`. If it was yarn doing this I would wait for https://github.com/Agoric/agoric-sdk/pull/5035/ to land before manually changing here because if the tool is going to add newlines then it shouldn't be a manual step to remove them. However it was my IDE settings doing that so 329eed8a4d53f3aa19fe364806add85e5da24980

kriskowal (commented):
Thank you for fixing the UNIX-style text file endings. There appears to be one more newline missing at the top level.

Please consider applying the corresponding change to Endo. An easy way to do this would be to sync everything `~/endo/scripts/all-versions.js ~/agoric-sdk | ~/endo/scripts/set-versions.sh`. It would be interesting to see if anything else changes. The `all-versions` script gets the highest version predicate for every transitive dependency in the SDK, and `set-versions.sh` applies it to every same-named shallow dependency. It’s not very good at `"*"`.

yarn.lock
```diff
@@ -6148,6 +6148,14 @@ chalk@^4.0.0, chalk@^4.1.0:
...
     supports-color "^7.1.0"
 
+chalk@^4.1.1:
+  version "4.1.2"
+  resolved "https://registry.yarnpkg.com/chalk/-/chalk-4.1.2.tgz#aac4e2b7734a740867aeb16bf02aad556a1e7a01"
+  integrity sha512-oKnbhFyRIXpUuez8iBMmyEa4nbj4IOQyuhc/wy9kY7/WVPcwIO9VA668Pu8RkO7+0G76SLROeyw9CpQ061i4mA==
+  dependencies:
+    ansi-styles "^4.1.0"
+    supports-color "^7.1.0"
+
```
mhofman: I wish yarn was more aggressive at bumping minor of existing deps (see version `4.1.0` being already in use above)

mhofman (commented):
This is fine by me.

> Revert "chore: bump node-lmdb" (bad typedefs)

Could you say more? What was the symptom, how was it detected, should we pin the explicit version instead of using a `^` ?

turadg:
> Could you say more? What was the symptom, how was it detected

Good question. Calls like this are valid but not in the typedef,
https://github.com/Agoric/agoric-sdk/blob/32f57f2/packages/swing-store/src/swingStore.js#L149

> should we pin the explicit version instead of using a ^ ?

I'd say no because the next patch version may fix it. It's easy enough to detect if not because the type checker fails.

mhofman:
> > Could you say more? What was the symptom, how was it detected
> 
> Good question. Calls like this are valid but not in the typedef, https://github.com/Agoric/agoric-sdk/blob/32f57f2/packages/swing-store/src/swingStore.js#L149

Oh that probing is yucky. I'm wondering if it was a WSL1 symptom, in which case I'd be in favor of removing altogether. 

> 
> > should we pin the explicit version instead of using a ^ ?
> 
> I'd say no because the next patch version may fix it. It's easy enough to detect if not because the type checker fails.

I misunderstood. If it'd just types, then yeah no problem.

yarn.lock
turadg: Lots of dupes, yeah. I added a `yarn-deduplicate` commit 9bf8751cbedd52c12326480d0bf700e66a4dab3a

dckc (approved):
Based on comments from @mhofman and @kriskowal , it looks like the team is ready for this go in, so I'll hit the relevant button.

yarn.lock
mhofman: Sweet!

PR #5035, opened 2022-04-06
chore: format package.json and document why not others

turadg:
## Description

Attempt to reduce PR discussion of whitespace. Automatically format `package.json` wit Prettier so it has trailing newline.

I tried to do all `.json` but some weren't improved by Prettier. I documented those cases.

### Security Considerations

--

--

### Testing Considerations

--


.prettierignore
```diff
@@ -15,6 +15,11 @@ public
 *.json
 *.md
 
+# selectively include instead of all .json because
+# - many (such as jsconfig.json) allow trailing commas want to preserve but Prettier will remove per JSON
+# - some have careful manual formatting such as a metrix and there's no safe way to have an ignore comment in JSON
```
kriskowal: What’s Metrix?

kriskowal (approved):
As long as this change does not sort properties in `package.json`, this should be fine.

.prettierignore
turadg: "matrix" as in https://github.com/Agoric/agoric-sdk/blob/master/packages/swingset-runner/demo/zoeTests/swingset.json

I'll fix before merge

kriskowal: Oh, very good. Matrixes vs. matrices: fight! (I favor matrixes since @romancortes and @aadsm convinced me that, when two words are equivalent and accepted, favoring the regular form is easier on an international audience.)

mhofman:
As long as tools which automatically update `package.json` do not conflict with the prettier format, I'm ok with this

Issue #5037, opened 2022-04-07
Version tagging for dapps


samsiegart:
## What is the Problem Being Solved?

Dapps are constantly broken due to `agoric-sdk` changes. This is because there's an implicit assumption that dapps will be run alongside the latest `master` branch of `agoric-sdk`.

## Description of the Design

The dapps should all have branches that correspond to specific `agoric-sdk` releases, and their `README`s should document which branch of the dapp and sdk to run. Whenever an `agoric-sdk` **devnet** release is cut, e.g. https://github.com/Agoric/agoric-sdk/tree/agoricdev-8, we should go through all the dapps, cut a new release branch (e.g. https://github.com/Agoric/dapp-treasury/tree/beta), make sure it works, and update the `README`.

## Test Plan

The following dapps should initially be fixed up to work with `https://github.com/Agoric/agoric-sdk/tree/agoricdev-8`, have a new branch called `dev-8` cut, and have their `README`s updated to instruct users to checkout `agoricdev-8` and `dev-8` respectively:

- [x] https://github.com/Agoric/dapp-treasury
- [x] https://github.com/Agoric/dapp-amm
- [x] https://github.com/Agoric/dapp-fungible-faucet
- [x] https://github.com/Agoric/dapp-card-store

michaelfig:
Please make the branch `agoricdev-8`, not `dev-8`.  We're referring to a specific chain ID, not just a short tag.

michaelfig:
@samsiegart, I think I see a problem that likely misguided you: `git checkout beta` in the `agoric-sdk` should always point to the latest devnet, but it was pointed at `agoricdev-7`.  I just now updated it to `agoricdev-8`; that would probably make your instructions a lot clearer on all the repos (rather than referring to an `agoricdev-*` tag by name, just use `beta`).

Sorry about that!

samsiegart:
> @samsiegart, I think I see a problem that likely misguided you: `git checkout beta` in the `agoric-sdk` should always point to the latest devnet, but it was pointed at `agoricdev-7`. I just now updated it to `agoricdev-8`; that would probably make your instructions a lot clearer on all the repos (rather than referring to an `agoricdev-*` tag by name, just use `beta`).

Updated all the linked PRs to point to `beta` instead of `agoricdev-*`. PTAL.

PR #5041, opened 2022-04-08
chore: use beta dapp branches for agoric init

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/5037


michaelfig (commented):
Putting `-b beta` in the published `agoric-cli` will live longer than most things if people download it from NPM.

I'm not sure if this would be better served by a documentation update.  Please let me know if you've considered that, and if so, I can approve this PR.

samsiegart:
Closed after discussion. We're going to just document in the dapps to use the `--dappBranch` option

PR #5042, opened 2022-04-08
Core-eval governance support

michaelfig:
## Description

New features and tests to help make core-eval governance more reproducible.


packages/deploy-script-support/src/getBundlerMaker.js
```diff
@@ -0,0 +1,38 @@
...
+      log('Instantiating bundle maker...');
+      const { publicFacet } = await E(zoe).startInstance(
+        E(zoe).install(bundle),
+      );
+      bundlerMaker = publicFacet;
+      await E(scratch).set('bundlerMaker', bundlerMaker);
+    }
+
+    if (board) {
+      const bundlerMakerId = await E(board).getId(bundlerMaker);
```
dckc: this `await` is not top level. worth a comment to say that this is not production code? either here or at the top of the file or both?

packages/deploy-script-support/src/getBundlerMaker.js
```diff
@@ -0,0 +1,38 @@
+// @ts-check
+import { E } from '@endo/far';
+import url from 'url';
+
+export const makeGetBundlerMaker =
+  ({ board, zoe, scratch }, { lookup, bundleSource }) =>
```
dckc: lack of types obscures the fact that board is optional. hm.

packages/deploy-script-support/src/getBundlerMaker.js
```diff
@@ -0,0 +1,38 @@
+// @ts-check
```
dckc, line 1: Add a note that this should go away after we have bundlecaps, with a pointer to and from a relevant issue?

dckc (approved):
presuming you have tested that this is useful in your oracle work, looks good.

my comments are not critical

JimLarson (approved):
The Go file LGTM.

packages/cosmic-swingset/scripts/clean-core-eval.js
```diff
@@ -0,0 +1,71 @@
...
+  const clean = defangEvaluableCode(text);
+  const withURL = `${clean}\n//# sourceURL=${fn}\n`;
+
+  // end-of-line whitespace disrupts YAML formatting
+  const trimmed = withURL.replace(/[\r\t ]+$/gm, '');
+  compartmentEvaluate(trimmed);
+  stdout.write(trimmed);
+};
+
+if (isEntrypoint(import.meta.url)) {
```
dckc, line 62: interesting approach. `isEntrypoint` is powerful, but I suppose so was `require.main`.

A [Stackoverflow post](https://stackoverflow.com/a/68848622/7963) has a succinct version:

```
if (import.meta.url === `file://${process.argv[1]}`) {
  // module was not imported but called directly
}
```

but that's less robust; it's followed by one that uses `url.pathToFileURL` as in `isEntrypoint`.

PR #5043, opened 2022-04-08
fix(swingset): disable metering on both initial and from-snapshot workers

### LINKED ISSUE #5040: xsnap launches with different metering limit from snapshot vs from empty
@arirubinstein and @mhofman and I were investigating a devnet follower node restart failure in which bouncing a follower caused an immediate "historical inaccuracy" error. This indicates that swingset was trying to bring a new worker online (which starts from a heap snapshot and then replays a recent subset of the transcript on top of it), but during the transcript replay, one of the deliveries made a different syscall than the one recorded in the transcript.

In the samples we looked at, the replayed delivery was an `install` to a special helper contract vat which was introduced in #5002 . This contract was created to overcome a cosmos/cosmic-swingset transaction size limit: the contract bundles we use have grown large enough that the whole bundle cannot fit in a single txn. While we wait for side-loading of contract bundles (#4564), this dodges the txn limit by performing a boatload of little txns instead, one per module. Each module is sent to the helper contract in its own txn, and then the last step is to ask the helper contract to synthesize the full bundle out of all those modules. The process is sort of like building a ship in a bottle through the narrow opening: it's a lot more work than inserting a pre-built ship, but it meets the constraint of the neck.

The `install` message which performs the bundle assembly is somewhat computationally intense. A successful run reported 3,648,285,911 computrons. This is 36x the per-crank limit that `xsnap` applies (`xsnap.js` `DEFAULT_CRANK_METERING_LIMIT = 1e8`).

We launch `xsnap` with a function in `controller.js` named `startXSnap`:

```js
  async function startXSnap(
    name,
    handleCommand,
    metered,
    snapshotHash = undefined,
  ) {
    if (snapStore && snapshotHash) {
      // console.log('startXSnap from', { snapshotHash });
      return snapStore.load(snapshotHash, async snapshot => {
        const xs = doXSnap({ snapshot, name, handleCommand, ...xsnapOpts });
        await xs.isReady();
        return xs;
      });
    }
    // console.log('fresh xsnap', { snapStore: snapStore });
    const meterOpts = metered ? {} : { meteringLimit: 0 };
    const worker = doXSnap({ handleCommand, name, ...meterOpts, ...xsnapOpts });

    for (const bundle of bundles) {
      assert(
        bundle.moduleFormat === 'getExport',
        X`unexpected: ${bundle.moduleFormat}`,
      );
      // eslint-disable-next-line no-await-in-loop
      await worker.evaluate(`(${bundle.source}\n)()`.trim());
    }
    return worker;
  }
```

The `doXSnap` function this uses is normally set equal to the `xsnap` exported by `packages/xsnap/src/xsnap.js`:

```js
export const DEFAULT_CRANK_METERING_LIMIT = 1e8;

export function xsnap(options) {
  const {
    snapshot = undefined,
    meteringLimit = DEFAULT_CRANK_METERING_LIMIT,
    ..
  } = options;
..
  let args = [name];
  if (snapshot) {
    args.push('-r', snapshot);
  }
  if (meteringLimit) {
    args.push('-l', `${meteringLimit}`);
  }
..
  const xsnapProcess = spawn(bin, args, {
    stdio: ['ignore', stdout, stderr, 'pipe', 'pipe'],
  });
```

We don't intend to enable metering for static vats, or indeed for any vats right now. `startXSnap` is always called with a falsy `metered` option.

However, it turns out that this function applies different metering arguments when launching from a snapshot versus launching the original (empty) worker. The `...meterOpts` field is added to the second call (`const worker =`), but not to the branch that adds the `snapshot` option. This allows `xsnap` function use its default `meteringLimit` in the first (snapshot) call. In the second call, our `meteringLimit: 0` option overrides `xsnap`'s default.

So the add `-l 100000000` when using a snapshot, but 

As a result, the original launch (in an empty vat) of this helper vat gets `-l 0`, allowing the expensive `install()` message to succeed. But if/when that node is restarted, the helper vat is loaded into a new worker that's started from a snapshot, which means it get `-l 100000000`, and the next `install()` delivery it executes will exceed the per-crank meter. The worker is terminated.

If this delivery happens during a transcript replay (which is likely, as the idle state of the helper vat is just after an `install()` message), swingset will notice the "historical inaccuracy": the transcript recorded a successful delivery, while the replay experienced a delivery failure.

If the node manages to survive transcript replay, the very next (new) message that does an `install` is likely to suffer a similar failure. Here, the node doesn't have any transcript to compare against, so it will think vat performed an in-consensus operation (it expects that *all* other validators/followers have observed the same metering fault). When the next block proposal arrives, it will notice an AppHash discrepancy, and will report a consensus failure instead.

The fix is to change `startXSnap` to apply the `...meterOpts` in both cases, disabling metering uniformly for both old and new workers.

## Deployment

The `controller.js` fix is trivial, but getting that change into an existing chain (e.g. the devnet, where we first saw this problem) is not. We currently package the kernel sources into a single bundle during initialization, then store that bundle in the database. Each time the node is launched, it reads the bundle from the DB and then evaluates it (`installBundle`) into a new compartment. Changing the source code does not cause a rebundling, so the node would continue to run the old (buggy) kernel each time.

To overcome this, we have roughly three options.

The first is a `bash` hack that Ari put together. Basically we replace the original `xsnap` binary with a small shell script that deletes any `-l NNN` option, adds a `-l 0` option, and then `exec`s the original `xsnap`. With this hack we were able to get our devnet node working and syncing again, without making any kernel source changes.

The second would be to dust off `packages/SwingSet/misc-tools/rekernelize.js`, apply the `controller.js` fix described above (to the devnet branch), make a new kernel bundle, then use `rekernelize` to install this bundle into your node's DB. The next time you restart, it will get the modified kernel bundle, which will include the fix.

The third would be to change `controller.js` to rebundle the kernel on each launch, ignoring the DB bundle entirely (discussed as one of the options in #4375). That would increase restart time by 5-10 seconds, but would enable kernel changes to be applied in much the same way that golang changes are delivered in the cosmos side: you fetch new source code, rebuild, and restart.

warner:
When vat options instruct swingset to disable metering in a vat, a bug
in `makeStartXSnap()` caused metering to be enabled anyways for the
second and subsequent worker used to host that vat. The first worker
starts from an empty XS engine, but after a few deliveries the
vat-manager will write out a heap snapshot, after which any new worker
will start from the snapshot. The `xsnap()` options that would disable
metering were not applied to the call that loaded a worker from
snapshot.

The consequence was that the first launch of a kernel would allow vats
to perform CPU-intense deliveries (as intended), but once the kernel
was restarted, those same delieries would terminate the vat (`compute
meter exceeded`) during transcript replay. If the snapshot was taken
recently enough that the remaining transcript did not include the
CPU-intense delivery, the new kernel would start, but the vat would
still be killed by any new delivery that exceeded the
unintentionally-imposed meter.

closes #5040

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -100,16 +100,22 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
     metered,
     snapshotHash = undefined,
   ) {
+    const meterOpts = metered ? {} : { meteringLimit: 0 };
     if (snapStore && snapshotHash) {
       // console.log('startXSnap from', { snapshotHash });
       return snapStore.load(snapshotHash, async snapshot => {
-        const xs = doXSnap({ snapshot, name, handleCommand, ...xsnapOpts });
+        const xs = doXSnap({
+          snapshot,
+          name,
+          handleCommand,
+          ...meterOpts,
+          ...xsnapOpts,
+        });
         await xs.isReady();
         return xs;
       });
     }
     // console.log('fresh xsnap', { snapStore: snapStore });
-    const meterOpts = metered ? {} : { meteringLimit: 0 };
     const worker = doXSnap({ handleCommand, name, ...meterOpts, ...xsnapOpts });
 
     for (const bundle of bundles) {
--- a/packages/SwingSet/test/test-xsnap-metering.js
+++ b/packages/SwingSet/test/test-xsnap-metering.js
@@ -1,0 +1,79 @@
+// eslint-disable-next-line import/order
+import { test } from '../tools/prepare-test-env-ava.js';
+
+import { spawn } from 'child_process';
+import fs from 'fs';
+import tmp from 'tmp';
+import { makePromiseKit } from '@endo/promise-kit';
+import { makeSnapStore, makeSnapStoreIO } from '@agoric/swing-store';
+
+import { makeStartXSnap } from '../src/controller/controller.js';
+
+// controller.js had a bug (#5040) wherein 'meterOpts' were applied
+// inconsistently to fresh workers vs ones from snapshot, allowing
+// fresh workers more runtime than ones reloaded from snapshot
+
+function handleCommand() {}
+
+function make(snapStore) {
+  const pk = makePromiseKit();
+  const startXSnap = makeStartXSnap([], {
+    snapStore,
+    env: {},
+    spawn: (command, args, opts) => {
+      pk.resolve(args);
+      return spawn(command, args, opts);
+    },
+  });
+  return { p: pk.promise, startXSnap };
+}
+
+function checkMetered(t, args, metered) {
+  const a = args.join(' ');
+  if (metered) {
+    // metered workers either have '-l NN' (with NN>0)
+    t.notRegex(a, /-l 0/);
+  } else {
+    // unmetered workers have '-l 0', or omit -l entirely
+    const elli = args.indexOf('-l');
+    if (elli === -1) {
+      t.pass();
+    } else {
+      t.is(args[elli + 1], '0');
+    }
+  }
+}
+
+async function doTest(t, metered) {
+  const pool = tmp.dirSync({ unsafeCleanup: true });
+  t.teardown(() => pool.removeCallback());
+  await fs.promises.mkdir(pool.name, { recursive: true });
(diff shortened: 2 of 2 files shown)


FUDCo (approved):
Oy.

dckc (approved):

PR #5050, opened 2022-04-08
fix(ui-components): dont throw on negative nat input

### LINKED ISSUE #92: Typing in negative value breaks the UI in autoswap
Typing in a negative value (-1) in the autoswap ui yields an error and removes the component from the page.

on commit: e070dc2f858db387dedb252137ae190dd7f48235

Reproduction steps:
1. launch vaultfactory
2. click autoswap
3. type `-1` in the first input field
4. component is removed with stack

https://user-images.githubusercontent.com/22369/158084318-56d341b9-e5a4-4ffe-9dc5-372ecce0feee.mov

samsiegart:
fixes: https://github.com/Agoric/dapp-treasury/issues/92


packages/ui-components/src/components/NatAmountInput.jsx
```diff
@@ -39,13 +39,6 @@ const makeNatAmountInput =
       value === null ? '0' : valueString,
     );
 
-    const preventSubtractChar = e => {
```
samsiegart, line 42: This wasn't being invoked on keypress, and it's also not sufficient for if someone pastes a string like "-5" instead of typing from scratch

packages/ui-components/src/components/NatAmountInput.jsx
```diff
@@ -55,8 +48,11 @@ const makeNatAmountInput =
     // We want to delay the input validation so that the user can type
     // freely, and then it gets formatted appropriately after the user stops.
     const handleOnChange = ev => {
-      const str = ev.target.value;
-      // Show the user exactly what they are typing
+      /** @type {string} */
+      let str = ev.target.value;
+      if (str.includes('-')) {
+        str = '';
+      }
```
michaelfig: I approve of your approach.  Just wondering if you've considered something like:
```suggestion
      let str = ev.target.value;
      str = str.replace(/-/g, '');
```

michaelfig (approved):
LGTM!  Just one question, but not a blocker.

packages/ui-components/src/components/NatAmountInput.jsx
samsiegart: Done. I'm just afraid of regex, this seems like a slightly better experience though

PR #5051, opened 2022-04-08
fix(actions): slack webhook url

arirubinstein:
## Description
Adds the slack webhook url to the notify stanzas


michaelfig (approved):
Ah, good catch!  I didn't look closely enough to see it.

PR #5052, opened 2022-04-08
refactor(run-protocol): virtual kind for vault manager

turadg:
Part of #4345

## Description

Make vault manager a virtual object using `defineKind`. Also virtualize whatever else must be to achieve that.

### Security Considerations

Object state is now managed by SwingSet. 

Moves governance to a `factoryPowers` object until governance can be made durable generally.

Refactor, shouldn't affect usage.

### Testing Considerations

Refactor, existing tests shouldn't be affected.


packages/vat-data/package.json
```diff
@@ -23,8 +23,9 @@
     "@agoric/store": "^0.6.10"
   },
   "devDependencies": {
+    "@agoric/swingset-vat": "*",
     "ava": "^3.12.1",
-    "@agoric/swingset-vat": "*"
+    "tsd": "^0.20.0"
```
turadg: all the `vat-data` changes (including yarn.lock) are being reviewed in a separate PR.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -38,127 +37,207 @@ const trace = makeTracer('VM');
...
+      {
+        interestRate,
+        chargingPeriod: state.factoryPowers
+          .getGovernedParams()
+          .getChargingPeriod(),
+        recordingPeriod: state.factoryPowers
+          .getGovernedParams()
+          .getRecordingPeriod(),
+      },
+      // TODO make something like _.pick
```
Chris-Hibbert: This doesn't mean anything to me.  Is this something to be done soon, or an idea that might be viable later?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -171,16 +250,22 @@ export const makeVaultManager = (
    * Instead, when a priceQuote is received, we'll only reschedule if the
    * high-water level when the request was made matches the current high-water
    * level.
+   *
+   * @param {{ state: ImmutableState & MutableState, facets: * }} context FIXME
```
Chris-Hibbert: Is the intent to merge this with the `FIXME` still present?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -171,16 +250,22 @@ export const makeVaultManager = (
...
    */
-  const reschedulePriceCheck = async () => {
+  reschedulePriceCheck: async ({ state, facets: { helper } }) => {
+    const { outstandingQuote, priceAuthority } = state;
+    const { liquidationInProgress, prioritizedVaults } = state;
```
Chris-Hibbert: Why is this best as two statements?  I appreciate fewer lines of code, but the grouping and location doesn't have clear significance.

Would it be better to extract each of them closer to where they are first used?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -227,164 +312,169 @@ export const makeVaultManager = (
       getAmountIn(quote),
     );
 
-    outstandingQuote = undefined;
+    state.outstandingQuote = null;
```
Chris-Hibbert, line 310: (For my education)  When is it better to use `null` rather than `undefined`?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -227,164 +312,169 @@ export const makeVaultManager = (
...
+      zcf,
+      vault,
+      debtMint.burnLosses,
+      state.liquidationStrategy,
+      state.collateralBrand,
+      penaltyPoolSeat,
+      factoryPowers.getGovernedParams().getLiquidationPenalty(),
+    )
+      .then(() => {
+        prioritizedVaults?.removeVault(key);
```
Chris-Hibbert: I'm sure I've asked this before, but why is optional chaining helpful here, but not on any of the other reads of `prioritizedVaults`?  

If we don't have a theory or a test that breaks, I'd prefer uniformity over keeping it because someone must have had a reason.

I also realize that it isn't new.

This use of optional chaining in `getCollateralQuote` makes sense to me. `displayInfo` is allowed to be null, and there's a fallback in case it is.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -227,164 +312,169 @@ export const makeVaultManager = (
...
+    ...otherSeats
+  ) => {
+    helper.checkDebtLimit(toMint);
+    state.factoryPowers.mintAndReallocate(toMint, fee, seat, ...otherSeats);
+    state.totalDebt = AmountMath.add(state.totalDebt, toMint);
+  },
   /**
+   * @param {MethodContext} context
+   * @param {*} toBurn
+   * @param {*} seat
```
Chris-Hibbert: Why leave these as `*`?

```suggestion
   * @param {Amount} toBurn
   * @param {ZCFSeat} seat
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -227,164 +312,169 @@ export const makeVaultManager = (
...
+   */
+  burnAndRecord: ({ state }, toBurn, seat) => {
+    const { burnDebt } = state.factoryPowers;
+    burnDebt(toBurn, seat);
+    state.totalDebt = AmountMath.subtract(state.totalDebt, toBurn);
+    // TODO signal updater?
```
Chris-Hibbert: I don't think so. I'm pretty sure it's always part of an operation that's recorded appropriately.

Chris-Hibbert (approved):
LGTM, with very minor questions.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Referring to https://lodash.com/docs/4.17.15#pick

Do you think that would be useful? If you do I'll make a ticket and link it here. Otherwise I'll just remove the comment.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: oops nope

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: I'm not sure how I got to this, wasn't intended. I'll merge or move them closer to use.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: It's debatable and I don't think we have coding standards on this. My reasoning was that `defineKind` requires all keys that will ever be on `state` to be on the object that `initState` returns. They could be `foo: undefined` but that is about the same as not being there because missing keys also resolve to `undefined`. So I think that since we do have to explicitly define them we're better off using `null` as a habit to convey that.

Another tack is to appeal to language: the outstanding quote _is_ defined, to be null.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: This is a vestige. Shouldn't be necessary now that `prioritizedVaults` is always defined. Will fix.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: 👍

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: What I went with: pull out of state what is _immutable_ and when it's mutable access it on the `state` object.

turadg: What I went with: pull out of state what is _immutable_ and when it's mutable access it on the `state` object.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Ok. @dtribble asked this in https://github.com/Agoric/agoric-sdk/blame/512db545c4e88fa4126c44a29f3a8775c330264f/packages/run-protocol/src/vaultFactory/vaultManager.js#L361

I'll remove the comment in this PR and if there's work to do I'll [make a ticket](https://github.com/Agoric/agoric-sdk/pull/4877).

packages/run-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: I agree that would sometimes be useful.

PR #5055, opened 2022-04-09
fix(swingset): track doMoreGC correctly

### LINKED ISSUE #5044: two virtual-object GC problems
@FUDCo and I noticed two problems in the virtual-object GC code today:

## Collections vs Remotables in values

`collectionManager.js` has a function named `deleteInternal()` used to delete an entry (key plus value) from a map-like collection. The value may contain some arbitrary number of vref-indexed objects. When the entry goes away, we need to decref each of those slots, by calling `vrm.removeReachableVref(vref)` on them. There is a loop that looks like:
```js
  value.slots.map(vrm.removeReachableVref);
```

The problem is that `vrm.removeReachableVref(vref)` has a boolean return value `doMoreGC`, which is `true` when the vref in question pointed to a Remotable (in RAM), and that Remotable's refcount just dropped to 0. In that case, we're going to be freeing some RAM, and that RAM might have captured/closed-over some other RAM presences/representatives/remotables. So if that ever happens, for *any* slot in this `value`, we need to inform our caller, so they can perform a new `gcAndFinalize` to let those new finalizers get noticed.

The loop was ignoring the return value. Instead, we need something like
```js
value.slots.forEach(vref => { doMoreGC = vrm.removeReachableVref(vref) || doMoreGC; } );
```

To test this, we'll need a collection with a string-keyed entry whose value contains a `[remotable, presence1]`. The Remotable's definition should closes over a `presence2`. Nothing else should keep either Presence alive. Allow everything to settle, then delete the collection entry (without causing anything to be deleted from RAM: no Presences or Remotables or Representatives should go out of scope on the delivery that does `collection.delete('stringkey')`).

On the next `bringOutYourDead`, the Remotable should be collected along with the vref for `presence2`, leaving `presence1` unreferenced. However the bug will inhibit a second call to `gcAndFinalize`, so that BOYD will *not* observe the presence1 finalizer run. The test program should be able to observe the presence-2 vref being missing from the c-list, but the presence1-vref still being present. A second BOYD (e.g. any random message, assuming the `reapInterval` is set to 1) will flush out presence-1 vref.

Once the bug is fixed, both presence-1 and presence-2 vrefs should be removed from the c-list during the first message, and it should not require a second message to get both.

## Virtual-Object Properties vs doMoreGC

In `virtualObjectManager.js`, the `deleteStoredVO()` function does a similar call to `vrm.removeReachableVref()` which *does* preserve the "do more" flag:

```js
        for (const propValue of Object.values(rawState)) {
          propValue.slots.map(
            vref => (doMoreGC = doMoreGC || vrm.removeReachableVref(vref)),
          );
        }
```

The problem is that if one call sets `doMoreGC`, subsequent calls will be bypassed by the short-circuit operator.

The fix is to swap the arguments:
```js
            vref => (doMoreGC = vrm.removeReachableVref(vref) || doMoreGC),
```

To test this one, we need a virtual object with one property. The value of this property should have two slots, like `state.x = [remotable, presence]`. The first slot should be a Remotable, so `doMoreGC` is set to `true`. The bug will cause the second slot to not be removed. By using a presence, the test program can inspect the vat's c-list to see whether it was dropped or not.

```js
// target vat
export function buildRootObject() {
  let foo;
  const makeFoo = defineKind('foo', (rem, pres) => { x: [rem, pres] }, {});
  return Far('root', {
    create(pres) { foo = makeFoo(Far('rem', {}), pres); },
    forget() { foo = 0; },
  });
}
```

```js
// bootstrap vat
export functionbuildRootObject() {
  const sensor = Far('sensor', {});
  let target;
  return Far('root', {
    bootstrap(vats) {
      target = vats.target;
    },
    start() {
      E(target).create(sensor);
      return sensor;
    },
    drop() {
      return E(target).forget();
    },
  });
}
```

The test program should create these two vats, use `queueToKref` to send `start` to bootstrap, and wait for `c.run()` to settle, then look at the result and extract the kref of `sensor`, then look up that kref in the target vat's c-list to make sure it is still imported. Then it should send `drop` to bootstrap and wait for it to settle, then re-examine the c-list. At this point, if the bug is fixed, the `sensor` kref should be missing, because the target vat should have dropped it along with the rest of `foo`. (Note that you may need to set `virtualObjectCacheSize: 0` and create a dummy object to make sure `foo` is really gone). Of course, run the test before fixing the bug, to make sure it's working.

warner:
Virtual collections `deleteInternal()` was ignoring the `doMoreGC`
result when dropping the value of a collection entry. This could fail
to re-run `gcAndFinalize` when necessary, allowing a Remotable (and
any in-memory objects it referenced) to survive longer than
appropriate before getting GCed.

Also, `deleteStoredVO()` allowed `doMoreGC` to short-circuit a call to
`removeReachableVref`, so if a Remotable appeared in the slots of a
virtual object's state, any slots after that point would not be
collected.

fixes #5044

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -333,18 +333,18 @@ export function makeCollectionManager(
       const rawValue = syscall.vatstoreGet(dbKey);
       assert(rawValue, X`key ${key} not found in collection ${q(label)}`);
       const value = JSON.parse(rawValue);
-      value.slots.map(vrm.removeReachableVref);
+      const doMoreGC1 = value.slots.map(vrm.removeReachableVref).some(b => b);
       syscall.vatstoreDelete(dbKey);
-      let doMoreGC = false;
+      let doMoreGC2 = false;
       if (passStyleOf(key) === 'remotable') {
         deleteOrdinal(key);
         if (hasWeakKeys) {
           vrm.removeRecognizableValue(key, `${collectionID}`, true);
         } else {
-          doMoreGC = vrm.removeReachableVref(convertValToSlot(key));
+          doMoreGC2 = vrm.removeReachableVref(convertValToSlot(key));
         }
       }
-      return doMoreGC;
+      return doMoreGC1 || doMoreGC2;
     }
 
     function del(key) {
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -687,9 +687,9 @@ export function makeVirtualObjectManager(
       const rawState = fetch(baseRef);
       if (rawState) {
         for (const propValue of Object.values(rawState)) {
-          propValue.slots.map(
-            vref => (doMoreGC = doMoreGC || vrm.removeReachableVref(vref)),
-          );
+          propValue.slots.forEach(vref => {
+            doMoreGC = vrm.removeReachableVref(vref) || doMoreGC;
+          });
         }
       }
       syscall.vatstoreDelete(`vom.${baseRef}`);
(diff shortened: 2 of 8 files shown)


warner:
Leaving this in draft state until I can build some tests that fail without it.

warner:
Took longer than I thought, but I found a good way to test the fixes.

FUDCo (approved):
This may be the biggest ratio of test size / fix size that I've seen, but looks good.

PR #5059, opened 2022-04-10
Virtualize weak store GC tables

### LINKED ISSUE #4834: vrefRecognizers uses RAM for purely virtual objects
@FUDCo and I identified a performance (RAM usage) problem in the current virtualReferences.js code. There's a `Map` named `vrefRecognizers` which maps vrefs to a Set of things that might be able to recognize the corresponding object. This includes vref-aware WeakMaps/WeakSets, and all weak-keyed virtual collections. It tracks Presence-style keys and Representative-style keys, not Remotable-style keys.

The problem is that `vrefRecognizers` will have an entry for every vref that is used as a key, even if there is currently no Presence or Representative for that vref. This violates our intention that we don't consume RAM for objects in this state. We wrote this code before we had virtual collections, and it was ok for the vref-aware WeakMaps/WeakSets of the time (which are not intended to be virtual, so they're allowed to consume RAM for every entry). But when we implemented virtual collections, we should have built an on-disk version of this table.

So the work item is to figure out how to keep that table on disk. We probably want two separate tables: use the existing approach for vref-aware WeakMap/WeakSets, but build a new thing for virtual collections.

The on-disk table will probably map from the vref of the thing being used as a weak key, to the vref of the collection enclosing it. Actually it will probably map to a stringified array of vrefs, since a single object can be used as a key in multiple collections.

Until we implement this, virtual collections will continue to consume memory O(N) in the number of entries.

FUDCo:
Weak map stores and weak set stores previously kept critical inverse-index information (needed for GC) in RAM, a bit of repurposed machinery from the implementation of virtual object aware WeakMap and WeakSet, which we belatedly realized was highly inappropriate.  Now this information is instead kept in the vat store.

Along the way I found a bug in the GC of weak map store entries which is now fixed (along with a test to verify that).

Closes #4834


packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -129,6 +129,21 @@ export function makeCollectionManager(
...
+  function deleteCollectionEntry(collectionID, vobjID) {
+    const ordinalKey = prefixc(collectionID, `|${vobjID}`);
+    const ordinalString = syscall.vatstoreGet(ordinalKey);
+    syscall.vatstoreDelete(ordinalKey);
+    const ordinalTag = zeroPad(ordinalString, BIGINT_TAG_LEN);
+    const recordKey = prefixc(collectionID, `r${ordinalTag}:${vobjID}`);
+    const rawValue = syscall.vatstoreGet(recordKey);
+    if (rawValue !== undefined) {
+      const value = JSON.parse(rawValue);
+      value.slots.map(vrm.removeReachableVref);
```
warner: I think this introduces a third instance of the two bugs in #5044 (fixed by #5055).. any time we call `removeReachableVref` we need to pay attention to the `doMoreGC` return value.

At this point, it probably makes more sense to have liveslots export a `doMoreGC` *function*, and it set a `needMoreGC` flag, and have the `scanDeadSlots` loop look for that flag at the end. Then all the places that currently try to return a `doMoreGC` flag can instead call `doMoreGC()`. That should be simpler than threading the return value back through all the right places (and would change the #5055 fix too).

I'm ok with this landing as-is if we expand #5044 to include it, and delay landing #5055 until it fixes this too. I'd rather rely upon your new test to exercise this instance of the fix, rather than writing a third test for #5055.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -466,43 +466,56 @@ export function makeVirtualReferenceManager(
...
-  function removeRecognizableValue(value, recognizer) {
+  function removeRecognizableValue(value, recognizer, recognizerIsVirtual) {
     const vref = getSlotForVal(value);
     if (vref) {
-      removeRecognizableVref(vref, recognizer);
+      removeRecognizableVref(vref, recognizer, recognizerIsVirtual);
     }
   }
 
+  let deleteCollectionEntry;
```
warner, line 513: This bit of circularity is unfortunate, but I imagine this is the best available way to deal with it.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -466,43 +466,56 @@ export function makeVirtualReferenceManager(
...
     if (type === 'object' && (!allocatedByVat || virtual)) {
-      const recognizerSet = vrefRecognizers.get(vref);
-      assert(recognizerSet && recognizerSet.has(recognizer));
-      recognizerSet.delete(recognizer);
-      if (recognizerSet.size === 0) {
-        vrefRecognizers.delete(vref);
-        if (!allocatedByVat) {
-          addToPossiblyRetiredSet(vref);
+      if (recognizerIsVirtual) {
+        syscall.vatstoreDelete(`vom.ir.${vref}|${recognizer}`);
```
warner, line 491: This introduces a constraint that vrefs cannot have `|` in them, right? We don't currently do that, but we should make a note of it, so we don't accidentally allow them in the future.

* add a brief section to your excellent comment in `parseVatSlot()` about characters which are prohibited from vrefs because other tables need them for delimiters, and give it a one-entry list naming `|` and a pointer to `removeRecognizableVref` and the `vom.ir.${vref}|${recognizer}` keys it uses
* Also update `src/liveslots/vatstore-usage.md` with an explanation of the new `vom.ir.*` keys

packages/SwingSet/src/liveslots/collectionManager.js
warner: Or, since it looks like the function that calls this is prepared to accept a `doMore` return value, maybe just improve the `map` to track the `removeReachableVref` return value in the same way that #5055 does it, and return that here. Then we can put off the cleanup for later, knowing that this does the right thing.

warner: Also.. this is probably a place where the typechecker would have helped, since the caller is expecting a boolean and this is returning void.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -546,16 +559,30 @@ export function makeVirtualReferenceManager(
...
+    }
     return doMoreGC;
   }
 
   function isVrefRecognizable(vref) {
-    return vrefRecognizers.has(vref);
+    if (vrefRecognizers.has(vref)) {
+      return true;
+    } else {
+      const [key] = syscall.vatstoreGetAfter('', `vom.ir.${vref}|`);
```
warner, line 580: ooh, this makes me nervous about determinism, we should think it through

`isVrefRecognizable()` might be called by `scanForDeadObjects()` (depending upon `possiblyRetiredSet`, and checks of `getValForSlot`, which is sensitive to GC). It's also called in there for actually-dead vrefs that referred to Presences (imports).

My worry is that doing the vatstore syscall from a GC-sensitive function means we're sometimes making the syscall and sometimes we're not, and that's a recipe for a consensus failure.

I'm also a tiny bit worried that allowing `vrefRecognizers.has` to short-circuit the syscall might make it worse, although I'm not sure `vrefRecognizers.has` is itself GC-sensitive.

warner (changes requested):
One batch of comments to add, one `doMoreGC` return value to track, and one brainstorming session on GC-determinism in `isVrefRecognizable` to have.

packages/SwingSet/src/liveslots/virtualReferences.js
warner: Oh, I think we're ok here, that's what `dispatch.bringOutYourDead` is for. The transcript will observe a `vatstoreGetAfter` syscall as a consequence of finalizers firing, but only during the consensus delivery of a BOYD. And I think `vrefRecognizers` is only updated by consensus actions.

warner:
Oh one other thing I noticed: the `@typedef { .. } Recognizer` just before the `const vrefRecognizers =` should be updated to remove the bare function option.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: I believe we already had the exclusion of `|` due to its use as a delimiter elsewhere, but even if that recollection is faulty you're right that this exclusion should be documented.  I have updated `parseVatSlot` and the docs accordingly.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: Yes. I muttered under my breath about this at length when I was unable to think of a better way.

packages/SwingSet/src/liveslots/collectionManager.js
FUDCo: Good catch!

I added JSDoc comments to the function, though this is not the complete solution since the type info also needs to be propagated through `setDeleteCollectionEntry` et al, which is far more TS pain than I care to handle right now.

warner (approved):
Updates look good.. thanks!

PR #5063, opened 2022-04-11
refactor!: atomically update multiple parameters via governance

### LINKED ISSUE #4881: Governance change multiple parameters simultaneously

Under governance, we can change parameters in a contract, but if multiple parameters need to be coordinated, they'll be changed in separate calls.

## Description of the Design

The solution needs to ensure that when governance seeks to change multiple parameters together, that the question reflects that they're together and it needs to do the update in a single call to the contract, or ensure that no other calls to the contract are allowed between the changes.

## Security Considerations

Changing parameters one at a time could allow external requests (or internal actions like charging interest to see an inconsistent set of parameters.

## Test Plan

validate that multiple parameters are changed. If the design is to send entrained messages that don't allow other messages to interleave, point to a swingset test that shows they can't be separated.

See also #4783.

Chris-Hibbert:
closes: #4881

## Description

The description of the change in the issue becomes  a set of
parameter changes so that if the vote passes, a single call from
governParam to the paramManager will make all the changes
atomically. If the update fails, all the changes will be unwound.

type declarations were updated.

Invitations require a two-phase update in paramManager, so all setters
had to be changed similarly for compatibility.

`setters` are no longer shared in the paramManager, though
`updateParams` still is, so the powerful facet of paramManager
continues to need to be closely held.

Tests for not updating partially are included

### Security Considerations

Changes parameter governance to take a list of parameters to change rather than a single change.

We don't yet have user docs for governance, so this just gets added to the list of things to write up. The format for declaring parameters didn't change, so that doc is fine.

### Testing Considerations

Tested that if an update fails, none of the proposed changes takes effect.


packages/governance/src/contractGovernance/governParam.js
```diff
@@ -21,10 +20,15 @@ const { details: X, quote: q } = assert;
 const CONTRACT_ELECTORATE = 'Electorate';
 
 /** @type {MakeParamChangePositions} */
-const makeParamChangePositions = (paramSpec, proposedValue) => {
-  const positive = harden({ changeParam: paramSpec, proposedValue });
-  const negative = harden({ noChange: paramSpec });
-  return { positive, negative };
+const makeParamChangePositions = changes => {
+  const dense = changes.map(c => {
```
turadg: `dense` and `sparse` aren't clear to me. Also, why is `dense` an array ob objects with a single name:value ? 

Consider `paramValues` as a mapping of all names to their values.

packages/governance/src/contractGovernance/governParam.js
```diff
@@ -21,10 +20,15 @@ const { details: X, quote: q } = assert;
...
+    return { [c.parameterName]: c.proposedValue };
+  });
+  const positive = { changes: dense };
+  const sparse = changes.map(c => c.parameterName);
```
turadg:
```suggestion
  const paramNames = changes.map(c => c.parameterName);
```

packages/governance/src/contractGovernance/governParam.js
```diff
@@ -51,13 +55,22 @@ const validateParamChangeQuestion = details => {
...
-const assertBallotConcernsQuestion = (paramName, questionDetails) => {
+/** @type {AssertBallotConcernsParam} */
+const assertBallotConcernsParam = (paramSpec, questionDetails) => {
+  const { key, parameterName } = paramSpec;
+  /** @type {ParamChangeIssue} */
+  // @ts-ignore cast
+  const issue = questionDetails.issue;
+  assert(issue, X`must be a param change issue`);
+
+  const foundName = issue.changes.find(c => c.parameterName === parameterName);
```
turadg: if `changes` were a map this would be O(1) lookup

packages/governance/src/contractGovernance/governParam.js
```diff
@@ -51,13 +55,22 @@ const validateParamChangeQuestion = details => {
...
 
-/** @type {AssertBallotConcernsQuestion} */
-const assertBallotConcernsQuestion = (paramName, questionDetails) => {
+/** @type {AssertBallotConcernsParam} */
+const assertBallotConcernsParam = (paramSpec, questionDetails) => {
+  const { key, parameterName } = paramSpec;
+  /** @type {ParamChangeIssue} */
+  // @ts-ignore cast
+  const issue = questionDetails.issue;
+  assert(issue, X`must be a param change issue`);
```
turadg: no "details"
```suggestion
  assert(issue, 'must be a param change issue');
```

packages/governance/src/contractGovernance/governParam.js
```diff
@@ -72,29 +85,32 @@ const setupParamGovernance = async (
   /** @type {WeakSet<Instance>} */
   const voteCounters = new WeakSet();
 
-  /** @type {VoteOnParamChange} */
-  const voteOnParamChange = async (
-    paramSpec,
-    proposedValue,
+  /** @type {VoteOnParamChanges} */
+  const voteOnParamChanges = async (
```
turadg, line 90: "vote on changes" doesn't convey that it's atomic. consider "changeset" as the term for a collection of param changes that live or die to together.

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -102,6 +101,7 @@ const makeParamManagerBuilder = zoe => {
     // to our caller. They must handle them carefully to ensure that they end up
     // in appropriate hands.
     setters[`update${name}`] = setParamValue;
+    setters[`prepareToUpdate${name}`] = () => {};
```
turadg: void fn:
```suggestion
    setters[`prepareToUpdate${name}`] = () => undefined;
```
do we really need this no-op though? It looks like only the async types need it.

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -273,6 +281,33 @@ const makeParamManagerBuilder = zoe => {
     return harden(descriptions);
   };
 
+  /** @type {UpdateParams} */
+  const updateParams = async paramChanges => {
+    const asynchResults = [];
```
turadg:
```suggestion
    const asyncResults = [];
```

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -273,6 +281,33 @@ const makeParamManagerBuilder = zoe => {
...
+    for (const changeSpec of paramChanges) {
```
turadg: to avoid mutating,
```
const asyncResults = paramChanges.map(s => setters[`prepareToUpdate${s.parameterName}`](s.proposedValue)));
```
or if that's too dense,
```
const asyncResults = paramChanges.map(s => {
  const prepare = setters[`prepareToUpdate${s.parameterName}`];
  return prepare(s.proposedValue);
});
```

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -273,6 +281,33 @@ const makeParamManagerBuilder = zoe => {
...
+      asynchResults.push(
+        setters[`prepareToUpdate${changeSpec.parameterName}`](
+          changeSpec.proposedValue,
+        ),
+      );
+    }
+    // if all updates didn't succeed, fail the request
```
turadg:
```suggestion
    // if any update doesn't succeed, fail the request
```

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -273,6 +281,33 @@ const makeParamManagerBuilder = zoe => {
...
+    const newValues = [];
+    for (let i = 0; i < paramChanges.length; i += 1) {
+      const changeSpec = paramChanges[i];
+      newValues.push(
+        setters[`update${changeSpec.parameterName}`](
+          changeSpec.proposedValue,
+          results[i],
+        ),
+      );
+    }
```
turadg:
```suggestion
  const newValues = paramChanges.map((s, i) => {
    const set = setters[`update${changeSpec.parameterName}`];
    return set(s.proposedValue, results[i]);
  }
```

packages/governance/src/contractGovernance/governParam.js
Chris-Hibbert: > Why is dense an array ob objects with a single name:value ?

Good point.   fixed.   
I changed the names from 'dense' and 'sparse' to 'details' and 'namesOnly'.

packages/governance/src/contractGovernance/governParam.js
Chris-Hibbert: done

packages/governance/src/contractGovernance/paramManager.js
Chris-Hibbert: thanks

packages/governance/src/contractGovernance/paramManager.js
Chris-Hibbert: > do we really need this no-op though? It looks like only the async types need it.

In updateParams, we're treating the async and sync types polymorphically, so they both need `prepareToUpdate()`. Is there a better way to give the sync types a no-op function?

packages/governance/src/contractGovernance/paramManager.js

packages/governance/src/contractGovernance/paramManager.js
Chris-Hibbert: Thanks.  I changed `set()` to `setFn()`;

packages/governance/src/contractGovernance/governParam.js
Chris-Hibbert: I don't think the name needs to say it. I think most users would assume they're updated atomically.

packages/governance/src/contractGovernance/governParam.js
Chris-Hibbert: I refactored from a list of changes to a record.

packages/governance/src/contractGovernance/governParam.js
Chris-Hibbert: fixed.

packages/governance/src/contractGovernance/paramManager.js
Chris-Hibbert: Let me try this one again. I think it works with the refactoring I've done.

Chris-Hibbert: done.

packages/governance/test/swingsetTests/contractGovernor/test-governor.js
```diff
@@ -110,40 +110,65 @@ test.serial('contract governance', async t => {
...
+  'Electorate was {"brand":"[Alleged: Zoe Invitation brand]","value":[{"description":"questionPoser","handle":"[Alleged: InvitationHandle]","installation":"[Alleged: Installation]","instance":"[Alleged: InstanceHandle]"}]} after the vote.',
+  'MalleableNumber was "[299792458n]" after the vote.',
 ];
 
 test.serial('change electorate', async t => {
   const dump = await main(t, ['changeElectorateStart']);
   t.deepEqual(dump.log, expectedChangeElectorateLog);
 });
 
+const expectedBrokenUpdateLog = [
```
turadg, line 147: consider putting the log literals within their test.

packages/governance/src/types.js
```diff
@@ -321,9 +321,9 @@
 
 /**
  * @typedef {Object} ParamChangeIssue
- * @property {ParamSpecification} paramSpec
+ * @property {string} key
```
turadg, line 325: `ParamKey` typedef says "identifier for a paramManager within a contract". Is this one of those?

turadg (approved):
Overall LGTM.

I noticed some type fixes to do so per earlier conversations I committed them. I also notice a couple other code quality cleanups and pushed those too. Feedback on that liberty welcomed.

packages/governance/src/types.js
Chris-Hibbert: Yes.

Chris-Hibbert: Turns out it's the same purpose, but here it only needs the string, and not the object with a `key` field.  I created an issue (#5114) to remove ParamKey entirely.

PR #5088, opened 2022-04-11
chore: weaken swingset dependency on vat-data to dev-dependency

warner:
Swingset *tests* use `@agoric/vat-data`, but not `src/`, so this
changes swingset/package.json to list it in `devDependencies`, not
`dependencies`.

Meanwhile, `@agoric/vat-data` lists swingset in its `devDependencies`
for its own tests. This gives us a cycle of `devDependencies` but not
normal dependencies.

Code which imports vat-data must (generally) be running inside a vat,
which means the application that decided to put that code into a vat
probably depends upon swingset: it had to import
`makeSwingsetController` to create the housing. But the vat-data
-importing code doesn't know that, and doesn't need a direct
dependency on swingset, because it never imports anything from
swingset itself.

### DIFF
--- a/packages/SwingSet/package.json
+++ b/packages/SwingSet/package.json
@@ -21,6 +21,7 @@
     "lint:eslint": "eslint '**/*.js'"
   },
   "devDependencies": {
+    "@agoric/vat-data": "^0.1.0",
     "@endo/ses-ava": "^0.2.21",
     "@types/tmp": "^0.2.0",
     "ava": "^3.12.1",
@@ -34,7 +35,6 @@
     "@agoric/store": "^0.6.10",
     "@agoric/swing-store": "^0.6.5",
     "@agoric/xsnap": "^0.11.2",
-    "@agoric/vat-data": "^0.1.0",
     "@endo/base64": "^0.2.21",
     "@endo/bundle-source": "^2.1.1",
     "@endo/captp": "^2.0.3",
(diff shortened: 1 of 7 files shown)


turadg (approved):

warner:
@turadg do you have any idea why that [test failure](https://github.com/Agoric/agoric-sdk/runs/5980653719?check_suite_focus=true#step:4:681) is happening?:

```
Error {
    message: 'config.vats.bob: Failed to load module "./test/virtualObjects/vat-orphan-bob.js" in package "file:///home/runner/work/agoric-sdk/agoric-sdk/packages/SwingSet/" (1 underlying failures: Cannot find external module "@agoric/vat-data" in package file:///home/runner/work/agoric-sdk/agoric-sdk/packages/SwingSet/',
  }

  › packages/SwingSet/src/controller/initializeSwingset.js:444:15
  › async processGroup (packages/SwingSet/src/controller/initializeSwingset.js:447:27)
  › async initializeSwingset (packages/SwingSet/src/controller/initializeSwingset.js:466:31)
  › async buildVatController (packages/SwingSet/src/controller/controller.js:493:23)
  › async orphanTest (packages/SwingSet/test/virtualObjects/test-representatives.js:423:13)
  › async packages/SwingSet/test/virtualObjects/test-representatives.js:467:3
```

It sounds like `node_modules/` is not getting a copy of `vat-data`, but I thought that having it in `devDependencies` should have been sufficient to get it installed.

warner:
Ah, TIL about `config.includeDevDependencies`, which must be set to `true` for tests that use vats which import code from outside swingset's `dependencies`. A lot of our tests use packages that swingset uses internally, or to build liveslots (`marshal`, `eventual-send`, `promise-kit`), but when we're using beyond that set (and therefore only add it to `devDependencies`), we'll need this flag to tell `bundle-source` to look for them too.

PR #5089, opened 2022-04-11
feat(vat-data) pickFacet

turadg:
## Description

It's common when defining a kind to export a function that makes the kind and returns one of its facets. That can get unwieldy: https://github.com/Agoric/agoric-sdk/blob/d390e21c1e5fde116f2ba00913e3e3bde294707d/packages/run-protocol/src/vaultFactory/vaultManager.js#L563-L584

So this adds a simplifying helper,
```
export const makeVaultManager = pickFacet(makeVaultManagerKit, 'self');
```

### Security Considerations

None, simple helper.

Optional, apparent from uses.

### Testing Considerations

Tested by uses of it.

### DIFF
--- a/packages/vat-data/src/index.js
+++ b/packages/vat-data/src/index.js
@@ -48,6 +48,14 @@ export const {
 } = VatDataGlobal;
 
 /**
+ * When making a multi-facet kind, it's common to pick one facet to expose. E.g.,
+ *
+ *     const makeFoo = (a, b, c, d) => makeFooBase(a, b, c, d).self;
+ *
+ * This helper reduces the duplication:
+ *
+ *     const makeFoo = pickFacet(makeFooBase, 'self');
+ *
  * @type {import('./types').PickFacet}
  */
 export const pickFacet =
--- a/packages/vat-data/src/index.js
+++ b/packages/vat-data/src/index.js
@@ -46,3 +46,11 @@ export const {
   makeScalarBigSetStore,
   makeScalarBigWeakSetStore,
 } = VatDataGlobal;
+
+/**
+ * @type {import('./types').PickFacet}
+ */
+export const pickFacet =
+  (maker, facetName) =>
+  (...args) =>
+    maker(...args)[facetName];
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -9,7 +9,7 @@ import {
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { assert } from '@agoric/assert';
 import { AmountMath } from '@agoric/ertp';
-import { defineKind } from '@agoric/vat-data';
+import { defineKind, pickFacet } from '@agoric/vat-data';
 import { makeTracer } from '../makeTracer.js';
 import { calculateCurrentDebt, reverseInterest } from '../interest-math.js';
 import { makeVaultKit } from './vaultKit.js';
@@ -705,7 +705,6 @@ const makeInnerVaultBase = defineKind('InnerVault', initState, {
  * @param {InnerVaultManagerBase & GetVaultParams} manager
  * @param {VaultId} idInManager
  */
-export const makeInnerVault = (zcf, manager, idInManager) =>
-  makeInnerVaultBase(zcf, manager, idInManager).self;
+export const makeInnerVault = pickFacet(makeInnerVaultBase, 'self');
 
 /** @typedef {ReturnType<typeof makeInnerVault>} InnerVault */
--- a/packages/vat-data/src/types.d.ts
+++ b/packages/vat-data/src/types.d.ts
@@ -59,7 +59,18 @@ export type VatData = {
   ) => WeakSetStore<K>;
 };
 
+// The JSDoc is repeated here and at the function definition so it appears
+// in IDEs where it's used, regardless of type resolution.
 interface PickFacet {
+  /**
+   * When making a multi-facet kind, it's common to pick one facet to expose. E.g.,
+   *
+   *     const makeFoo = (a, b, c, d) => makeFooBase(a, b, c, d).self;
+   *
+   * This helper reduces the duplication:
+   *
+   *     const makeFoo = pickFacet(makeFooBase, 'self');
+   */
   <M extends (...args: any[]) => any, F extends keyof ReturnType<M>>(
     maker: M,
     facetName: F,
--- a/packages/vat-data/src/types.d.ts
+++ b/packages/vat-data/src/types.d.ts
@@ -58,3 +58,10 @@ export type VatData = {
     options?: StoreOptions,
   ) => WeakSetStore<K>;
 };
+
+interface PickFacet {
+  <M extends (...args: any[]) => any, F extends keyof ReturnType<M>>(
+    maker: M,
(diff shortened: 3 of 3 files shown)


warner:
If you didn't have this helper, you'd have to write something like:
```js
const makeFoo = defineKind(..);
export const makeFooFacet = harden((...args) => makeFoo(...args).facet);
```

right?

Could we avoid the need for a helper function if the `finish` option on `defineKind` could provide the return value from `defineKind`?

cc @FUDCo

turadg:
```
/**
 * @param {ZCF} zcf
 * @param {InnerVaultManagerBase & GetVaultParams} manager
 * @param {VaultId} idInManager
 */
export const makeInnerVault = (...args) => makeInnerVaultBase(...args).self;
```
That causes lint errors:
```
Missing JSDoc @param "args" declaration.
Expected @param names to be "[object Object]". Got "zcf, manager, idInManager".
```

If we dropped the JSDoc then callsites wouldn't have any documentation about how to use the function.

> Could we avoid the need for a helper function if the finish option on defineKind could provide the return value from defineKind?

You mean have an option to `defineKind` to specify what should be returned from it? I think we should minimize conditioning the return type on the parameters. I thought you had the same opinion.

warner:
> ```
> /**
>  * @param {ZCF} zcf
>  * @param {InnerVaultManagerBase & GetVaultParams} manager
>  * @param {VaultId} idInManager
>  */
> export const makeInnerVault = (...args) => makeInnerVaultBase(...args).self;
> ```
> 
> In that case the JSDoc wouldn't match any parameters. An IDE wouldn't be able to provide any documentation about how to use the function.

Oh, I assumed the types could be inferred. I'm guessing that's also why your original example had to spell out all the parameters twice? A `...args` would lose the type information? That's a shame.

> > Could we avoid the need for a helper function if the finish option on defineKind could provide the return value from defineKind?
> 
> You mean have an option to `defineKind` to specify what should be returned from it? I think we should minimize conditioning the return type on the parameters. I thought you had the same opinion.

We've already got a `finish` option which gives the user a chance to e.g. register facets with external tables, or do other work, after the facets are created but before they're returned from `makeFoo`. We could enhance that to allow it to override the return value, to strip out certain facets or return just one.

You're right that I'm against having function signatures depend upon the value of their arguments. I guess this would be at the edge of what I'd be happy with.. I'm more concerned about the input arguments than the return value, plus the return value would depend upon the inputs to `defineKind` not `makeFoo`, which feels better. Also the return value of `makeFoo` would be the same as the return value of the `finish` option.

But.. that last statement is only true if we made `finish` mandatory and didn't provide the convenience of `return undefined` meaning "return all facets", without which `finish` would be kind of annoying and error-prone. And I suspect it'd be really hard to get the types figured out right in either case.

Ok, so given the clear need for usable type definitions, I guess `finish` isn't the right place to provide this. I'm sad that we can't use the two-line form I suggested, JS provides such nice `...` convenience for building simple things. I wonder if the callers could just do `const facet = makeFoo().facet` themselves, but I don't know much about the real-world uses and I'm totally willing to believe that it's better to export the more-restricted function (for safety or clarity or whatever).

turadg:
> Oh, I assumed the types could be inferred. I'm guessing that's also why your original example had to spell out all the parameters twice?

Correct.

>  A ...args would lose the type information? That's a shame.

More specifically it would never have the type information. This helper does use `...args` but it expresses that the function signature of the outer call should be defined in terms of the inner call. I don't think you'd want the type system to assume that without a declaration.

> I wonder if the callers could just do const facet = makeFoo().facet themselves, but I don't know much about the real-world uses

Indeed they can. This is for cases when the caller shouldn't know about the base "kind" and only the exported facet.

> Ok, so given the clear need for usable type definitions,

Due to the antipathy to TypeScript I want to be clear that I'm appealing to usable JSDoc. The requirements motivating this PR are: 
- use cases in which we want to export only one facet of a multi-facet kind, not all of them
- that an IDE showing the signature of that wrapper function show the arguments it takes and their types

FUDCo:
> Could we avoid the need for a helper function if the `finish` option on `defineKind` could provide the return value from `defineKind`?

I don't care for that.  It would be more invisible magic voodoo.  The helper function at least has the virtue of being clear about its intentions at the call site, whereas if you put the logic in `finish` it would only be visible to somebody inspecting the code of the kind definition.

It _really_ bothers me that in
```
export const makeInnerVault = (...args) => makeInnerVaultBase(...args).self;
```
TS can't infer that the parameters of `makeInnerVault` are the same as the parameters of `makeInnerVaultBase`.  I think that's just a straight up bug in the type inference rules.

turadg:
> I think that's just a straight up bug in the type inference rules.

I disagree. I don't expect or want the type system to assume that that outer function call has the same signature as a call within it.

Does anyone have requested changes for this PR? I can't tell what's not yet approved:
- motivation is legitimate
- having a helper function is a good solution
- implementation details (e.g. function name)

FUDCo:
> > I think that's just a straight up bug in the type inference rules.
> 
> I disagree. I don't expect or want the type system to assume that that outer function call has the same signature as a call within it.

If I instead wrote:
```
export const makeInnerVault = (x, y, z) => makeInnerVaultBase(x, y, z).self;
```
It would infer the types of `x`, `y`, and `z`.  I don't see how using `...args` changes the logic -- "this over here matches that over there, therefor they are the same".  The virtue of using spread is that it's clearer about the intent of the outer function to delegate to the inner one, plus under maintenance it automatically propagates changes made to the inner signature to the outer signature.

packages/vat-data/src/index.js
```diff
@@ -46,3 +46,11 @@ export const {
   makeScalarBigSetStore,
   makeScalarBigWeakSetStore,
 } = VatDataGlobal;
+
+/**
+ * @type {import('./types').PickFacet}
+ */
+export const pickFacet =
```
dtribble, line 61: Please add a descriptive comment of what this is for.

dtribble (approved):
LGTM, though I woudl like to see a descriptive comment on the use of `pickFacet`

packages/vat-data/src/index.js
turadg: Screen Shot 2022-04-12 at 2 08 38 PM

mhofman:
> Also the return value of `makeFoo` would be the same as the return value of the `finish` option.

I think that makes it acceptable as it can easily be typed.

> But.. that last statement is only true if we made `finish` mandatory and didn't provide the convenience of `return undefined` meaning "return all facets", without which `finish` would be kind of annoying and error-prone. And I suspect it'd be really hard to get the types figured out right in either case.

Correct, I believe we should force a return value in `finish` in that case, but finish could still be optional and return the original value.


> I don't care for that. It would be more invisible magic voodoo. The helper function at least has the virtue of being clear about its intentions at the call site, whereas if you put the logic in `finish` it would only be visible to somebody inspecting the code of the kind definition.

But someone would have to inspect the behavior to know what the maker is returning in the first place. Behavior, init and finish should all be collocated where `defineKind` is called. I don't see that as particularly more cumbersome.

> It would infer the types of `x`, `y`, and `z`.

No it wouldn't. The outer function would be documented with all its params as `any`. We have a lot of bad patterns like that in our codebase.

PR #5091, opened 2022-04-12
fix!: fix a bug in addLiquidity

### LINKED ISSUE #8: "Add liquidity" gives me all the liquidity for the pool
I wanted just 10 LP tokens, but got millions. When I "remove liquidity" and remove all of them, it works and I become very rich. For some reason, the UI says I own a "50%" share of the pool but I seem to own 100%.

alltheliquidity
50percent

Chris-Hibbert:
closes: https://github.com/Agoric/dapp-amm/issues/8

## Description

The AMM was passing the new central token amount where the central pool amount was required in calculating the amount of liquidity tokens to mint.

### Security Considerations

Economic and contract considerations, but no impact on security, _per se_.

None.

### Testing Considerations

The AMM wasn't testing removal of liquidity at all. The existing tests of adding liquidity were using `addLiquidityAtRate`, which uses `addLiquidityActual()`, which sidesteps the location of the problem.


Chris-Hibbert:
@turadg  This is ready for review, though I'm waiting for word from @samsiegart as to whether it fixes the bug he encountered.

turadg (approved):
Approving to get in the tests and this fix. If it doesn't solve the problem observed in UI that may be an additional bug and PR fix.

> whether it fixes the bug he encountered

I assume the new tests fail on master.

samsiegart:
This seems to fix it. I cannot repro the bug when running off this branch

Chris-Hibbert:
> I assume the new tests fail on master.

Thanks for asking. I hadn't explicitly verified that, so I re-checked. Yes, they do fail with pool.js un-fixed.

PR #5092, opened 2022-04-12
Run test improvements

dtribble:
closes: #XXXX
refs: #XXXX

## Description

OfferTo
* offerTo support for offerArgs
* pass `debt` to liquidation as an offerArg

Testing
* add test/support.js for test helpers
* consolidate `makeBundle` and `setupZoeForTest`
* make `setupZoeForTest` sync and return promises
* add bootstrap `space` setup helper

Refactor vault tests:
* add `test.before` for shared setup
* use `t.context` pervasively for shared and default config
* change methods to override `t.context` before invoking `setupServices`
* use `econ-behaviors` directly
* move `addVaultType` into shared setup
* eliminate governance terms

Tracing
* turn off unneeded tracing
* add trace info in liquidation
* enable passing the Ava `t` as the first argument to trace to only show trace on failure

### Security Considerations
N/A

### Testing Considerations

Lots of testing  maintenance improvements

### DIFF
--- a/packages/run-protocol/src/makeTracer.js
+++ b/packages/run-protocol/src/makeTracer.js
@@ -17,8 +17,12 @@ const makeTracer = (name, enable = true) => {
       return infoTick;
     }
     default: {
-      const debugTick = (...args) => {
-        console.log(key, (debugCount += 1), ...args);
+      const debugTick = (optLog, ...args) => {
+        if (optLog.log) {
+          optLog.log(key, (debugCount += 1), ...args);
+        } else {
+          console.info(key, (debugCount += 1), optLog, ...args);
+        }
       };
       return debugTick;
     }
--- a/packages/run-protocol/src/vaultFactory/liquidateMinimum.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateMinimum.js
@@ -8,7 +8,7 @@ import { Far } from '@endo/marshal';
 import { makeDefaultLiquidationStrategy } from './liquidation.js';
 import { makeTracer } from '../makeTracer.js';
 
-const trace = makeTracer('LM');
+const trace = makeTracer('LiqMin');
 
 /**
  * This contract liquidates the minimum amount of vault's collateral necessary
@@ -24,15 +24,18 @@ const trace = makeTracer('LM');
 const start = async zcf => {
   const { amm } = zcf.getTerms();
 
-  /**
-   * @param {Amount<'nat'>} runDebt
-   */
-  const makeDebtorHook = runDebt => {
-    const runBrand = runDebt.brand;
-    return async debtorSeat => {
+  const makeDebtorHook = _runDebt => {
+    /**
+     * @param {ZCFSeat} debtorSeat
+     * @param {object} root0
+     * @param {Amount<'nat'>} root0.debt
+     */
+    return async (debtorSeat, { debt: runDebt }) => {
+      const runBrand = runDebt.brand;
       const {
         give: { In: amountIn },
       } = debtorSeat.getProposal();
+      // TODO XXX isn't `inBefore === aboutIn`?
       const inBefore = debtorSeat.getAmountAllocated('In');
 
       const swapInvitation = E(amm).makeSwapOutInvitation();
@@ -40,8 +43,7 @@ const start = async zcf => {
         give: { In: amountIn },
         want: { Out: runDebt },
       });
-      trace(`OFFER TO DEBT: `, runDebt.value);
-
+      trace(`OFFER TO DEBT: `, { runDebt, amountIn });
       const { deposited, userSeatPromise: liqSeat } = await offerTo(
         zcf,
         swapInvitation,
@@ -55,14 +57,20 @@ const start = async zcf => {
       // check whether swapIn liquidated assets until getOfferResult() returns.
       // Of course, we also can't exit the seat until one or the other
       // liquidation takes place.
-      await deposited;
-      await E(liqSeat).getOfferResult();
+      const [offerResult, amounts, _] = await Promise.all([
+        E(liqSeat).getOfferResult(),
+        E(liqSeat).getCurrentAllocation(),
+        deposited,
+      ]);
+      trace('offerResult', { offerResult, amounts });
 
       // if swapOut failed to make the trade, we'll sell it all
       const sellAllIfUnsold = async () => {
-        if (
-          !AmountMath.isEqual(inBefore, debtorSeat.getAmountAllocated('In'))
-        ) {
+        const unsold = debtorSeat.getAmountAllocated('In');
+        // TODO XXX This could be true if it gave me more collateral. This
+        // should check that we actually got the debt amount
+        if (!AmountMath.isEqual(inBefore, unsold)) {
+          trace('Changed', { inBefore, unsold });
           return;
         }
 
@@ -85,7 +93,11 @@ const start = async zcf => {
         });
       };
       await sellAllIfUnsold();
-
+      const proceeds = debtorSeat.getAmountAllocated('RUN', runBrand);
+      trace(`Liq results`, {
+        runDebt,
+        proceeds,
+      });
       debtorSeat.exit();
     };
   };
--- a/packages/run-protocol/src/vaultFactory/liquidation.js
+++ b/packages/run-protocol/src/vaultFactory/liquidation.js
@@ -75,6 +75,8 @@ const liquidate = async (
     strategy.keywordMapping(),
     strategy.makeProposal(collateralToSell, debt),
     vaultZcfSeat,
+    undefined,
+    harden({ debt }),
   );
   trace(` offeredTo`, collateralToSell, debt);
 
@@ -128,7 +130,7 @@ const makeDefaultLiquidationStrategy = amm => {
   trace(`return from makeDefault`);
 
   return {
-    makeInvitation: () => E(amm).makeSwapInInvitation(),
+    makeInvitation: _runDebt => E(amm).makeSwapInInvitation(),
     keywordMapping,
     makeProposal,
   };
--- a/packages/run-protocol/src/vaultFactory/vaultFactory.js
+++ b/packages/run-protocol/src/vaultFactory/vaultFactory.js
@@ -148,7 +148,12 @@ export const start = async (zcf, privateArgs) => {
     const { creatorFacet: liquidationFacet } = await E(zoe).startInstance(
       liquidationInstall,
       harden({ RUN: debtIssuer, Collateral: collateralIssuer }),
-      harden({ amm: ammPublicFacet }),
+      harden({
+        amm: ammPublicFacet,
+        priceAuthority,
+        timerService,
+        debtBrand,
+      }),
     );
     const liquidationStrategy = makeLiquidationStrategy(liquidationFacet);
 
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -27,7 +27,7 @@ import { chargeInterest } from '../interest.js';
 
 const { details: X, quote: q } = assert;
 
-const trace = makeTracer('VM');
+const trace = makeTracer('VM', false);
 
 /**
  * @typedef {{
@@ -157,6 +157,7 @@ export const makeVaultManager = (
         liquidationInProgress = false;
         // XXX should notify interested parties
         console.error('liquidateAndRemove failed with', e);
+        throw e;
       });
   };
 
@@ -297,7 +298,9 @@ export const makeVaultManager = (
     const debtPost = AmountMath.add(totalDebt, toMint);
     const limit = loanParamGetters.getDebtLimit();
     if (AmountMath.isGTE(debtPost, limit)) {
-      assert.fail(X`Minting would exceed total debt limit ${q(limit)}`);
+      assert.fail(
+        X`Minting ${toMint} would exceed total debt limit ${q(limit)}`,
+      );
     }
   };
 
--- a/packages/run-protocol/test/psm/test-psm.js
+++ b/packages/run-protocol/test/psm/test-psm.js
@@ -7,12 +7,7 @@ import '../../src/vaultFactory/types.js';
 
 import path from 'path';
 import { E } from '@endo/eventual-send';
-import bundleSource from '@endo/bundle-source';
-import { makeFakeVatAdmin } from '@agoric/zoe/tools/fakeVatAdmin.js';
-import { makeZoeKit } from '@agoric/zoe';
 import { AmountMath, makeIssuerKit } from '@agoric/ertp';
-import { resolve as importMetaResolve } from 'import-meta-resolve';
-import { makeLoopback } from '@endo/captp';
 import {
   makeRatio,
   floorDivideBy,
@@ -20,19 +15,13 @@ import {
   natSafeMath as NatMath,
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { makeTracer } from '../../src/makeTracer.js';
+import { makeBundle, setUpZoeForTest } from '../supports.js';
 
 const pathname = new URL(import.meta.url).pathname;
 const dirname = path.dirname(pathname);
 
 const psmRoot = `${dirname}/../../src/psm/psm.js`;
 const trace = makeTracer('TestPSM', false);
-
-const makeBundle = async sourceRoot => {
-  const url = await importMetaResolve(sourceRoot, import.meta.url);
-  const contractBundle = await bundleSource(new URL(url).pathname);
-  console.log(`makeBundle ${sourceRoot}`);
-  return contractBundle;
-};
 
 const BASIS_POINTS = 10000n;
 const WantStableFeeBP = 1n;
@@ -88,16 +77,9 @@ test.before(async t => {
   // makeBundle is slow, so we bundle each contract once and reuse in all tests.
   const psmBundle = await makeBundle(psmRoot);
   t.context.bundles = { psmBundle };
-  const { makeFar, makeNear: makeRemote } = makeLoopback('zoeTest');
-  const { zoeService, feeMintAccess: nonFarFeeMintAccess } = makeZoeKit(
-    makeFakeVatAdmin(setJig, makeRemote).admin,
-  );
-  /** @type {ERef<ZoeService>} */
-  const zoe = makeFar(zoeService);
-  t.context.zoe = zoe;
-  trace('makeZoe');
-  const feeMintAccess = await makeFar(nonFarFeeMintAccess);
-  t.context.feeMintAccess = feeMintAccess;
+  const { zoe, feeMintAccess } = setUpZoeForTest(setJig);
+  t.context.zoe = await zoe;
+  t.context.feeMintAccess = await feeMintAccess;
 
   const runIssuer = await E(zoe).getFeeIssuer();
   const runBrand = await E(runIssuer).getBrand();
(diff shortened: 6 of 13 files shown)


packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -317,6 +320,7 @@ export const offerTo = async (
   // @ts-expect-error A required parameter cannot follow an optional parameter.
   fromSeat,
   toSeat,
+  offerArgs = undefined,
```
turadg, line 323: ` = undefined` has no runtime effect because it's what JS already does when a parameter isn't passed in.

so it's purely for documentation, which we should leave to JSDoc.

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -304,6 +304,9 @@ const reverse = (keywordRecord = {}) => {
  *   The seat in contractA to deposit the payout of the offer to.
  *   If `toSeat` is not provided, this defaults to the `fromSeat`.
  *
+ * @param {Object=} offerArgs
```
turadg, line 307: [JSDoc params syntax](https://jsdoc.app/tags-param.html) allows Google Closure syntax but it's native is like this:
```
 * @param {Object} [offerArgs]
```
I think we should standardize on that, but just FYI since that's out of scope of this PR.

Also FYI, `Object` type is an interface for what's common in _all_ JS objects (including Number, etc). When the offer args aren't known this would be more accurate: `Record<string, unknown>`.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -8,7 +8,7 @@ import { Far } from '@endo/marshal';
 import { makeDefaultLiquidationStrategy } from './liquidation.js';
 import { makeTracer } from '../makeTracer.js';
 
-const trace = makeTracer('LM');
+const trace = makeTracer('LiqMin');
```
turadg, line 11: +1 to this verbosity :)

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -24,24 +24,26 @@ const trace = makeTracer('LM');
 const start = async zcf => {
   const { amm } = zcf.getTerms();
 
-  /**
-   * @param {Amount<'nat'>} runDebt
-   */
-  const makeDebtorHook = runDebt => {
-    const runBrand = runDebt.brand;
-    return async debtorSeat => {
+  const makeDebtorHook = _runDebt => {
```
turadg, line 27: the unused params should be removed

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -128,7 +130,7 @@ const makeDefaultLiquidationStrategy = amm => {
   trace(`return from makeDefault`);
 
   return {
-    makeInvitation: () => E(amm).makeSwapInInvitation(),
+    makeInvitation: _runDebt => E(amm).makeSwapInInvitation(),
```
turadg, line 133: what motivated this?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -297,7 +298,9 @@ export const makeVaultManager = (
     const debtPost = AmountMath.add(totalDebt, toMint);
     const limit = loanParamGetters.getDebtLimit();
     if (AmountMath.isGTE(debtPost, limit)) {
-      assert.fail(X`Minting would exceed total debt limit ${q(limit)}`);
+      assert.fail(
+        X`Minting ${toMint} would exceed total debt limit ${q(limit)}`,
```
turadg, line 302: intentional to expose `toMint` and not `limit`?

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -24,24 +24,26 @@ const trace = makeTracer('LM');
...
+    /**
+     * @param {ZCFSeat} debtorSeat
+     * @param {object} root0
+     * @param {Amount<'nat'>} root0.debt
```
turadg, line 31: I think this is the prevalent style and less verbose:
```suggestion
     * @param {{ debt: Amount<'nat'> }} options
```

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -24,24 +24,26 @@ const trace = makeTracer('LM');
...
+     */
+    return async (debtorSeat, { debt: runDebt }) => {
+      const runBrand = runDebt.brand;
```
turadg, line 34: why rename `debt`? especially when 'run' is going away
```suggestion
    return async (debtorSeat, { debt }) => {
      const debtBrand = debt.brand;
```

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -24,24 +24,26 @@ const trace = makeTracer('LM');
...
       const {
         give: { In: amountIn },
       } = debtorSeat.getProposal();
+      // TODO XXX isn't `inBefore === aboutIn`?
```
turadg, line 38: I don't know this logic well enough to help answer that. Let me know if you explicitly want review on that question.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -55,14 +57,20 @@ const start = async zcf => {
       // check whether swapIn liquidated assets until getOfferResult() returns.
       // Of course, we also can't exit the seat until one or the other
       // liquidation takes place.
-      await deposited;
-      await E(liqSeat).getOfferResult();
+      const [offerResult, amounts, _] = await Promise.all([
```
turadg, line 60: No use destructuring a value that won't be used.
```suggestion
      const [offerResult, amounts] = await Promise.all([
```

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -55,14 +57,20 @@ const start = async zcf => {
...
+      ]);
+      trace('offerResult', { offerResult, amounts });
 
       // if swapOut failed to make the trade, we'll sell it all
       const sellAllIfUnsold = async () => {
-        if (
-          !AmountMath.isEqual(inBefore, debtorSeat.getAmountAllocated('In'))
-        ) {
+        const unsold = debtorSeat.getAmountAllocated('In');
+        // TODO XXX This could be true if it gave me more collateral. This
```
turadg, line 70: sounds like a bug. needs a ticket? alternately a [failing test](https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#failing-tests) in this PR would be enough to ensure we come back and solve it.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -85,7 +93,11 @@ const start = async zcf => {
         });
       };
       await sellAllIfUnsold();
-
+      const proceeds = debtorSeat.getAmountAllocated('RUN', runBrand);
+      trace(`Liq results`, {
+        runDebt,
+        proceeds,
+      });
```
turadg, line 100: the `const proceeds` reads at first like it might be part of the logic. to isolate consider,
```suggestion
      const proceeds = debtorSeat.getAmountAllocated('RUN', runBrand);
      trace(`Liq results`, {
        runDebt,
        debtorSeat.getAmountAllocated('RUN', runBrand),
      });
```

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -75,6 +75,8 @@ const liquidate = async (
     strategy.keywordMapping(),
     strategy.makeProposal(collateralToSell, debt),
     vaultZcfSeat,
+    undefined,
```
turadg, line 78: WDYT of specifying?
```suggestion
    vaultZcfSeat,
```

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -148,7 +148,12 @@ export const start = async (zcf, privateArgs) => {
     const { creatorFacet: liquidationFacet } = await E(zoe).startInstance(
       liquidationInstall,
       harden({ RUN: debtIssuer, Collateral: collateralIssuer }),
-      harden({ amm: ammPublicFacet }),
+      harden({
+        amm: ammPublicFacet,
+        priceAuthority,
+        timerService,
+        debtBrand,
```
turadg, line 155: I don't see these being used.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -157,6 +157,7 @@ export const makeVaultManager = (
         liquidationInProgress = false;
         // XXX should notify interested parties
         console.error('liquidateAndRemove failed with', e);
+        throw e;
```
turadg, line 160: was this necessary or hygiene?

packages/run-protocol/src/makeTracer.js
```diff
@@ -17,8 +17,12 @@ const makeTracer = (name, enable = true) => {
...
     }
     default: {
-      const debugTick = (...args) => {
-        console.log(key, (debugCount += 1), ...args);
+      const debugTick = (optLog, ...args) => {
+        if (optLog.log) {
+          optLog.log(key, (debugCount += 1), ...args);
+        } else {
+          console.info(key, (debugCount += 1), optLog, ...args);
+        }
```
turadg, line 25: I don't see this used

packages/run-protocol/test/psm/test-psm.js
```diff
@@ -7,33 +7,22 @@ import '../../src/vaultFactory/types.js';
...
 import { makeTracer } from '../../src/makeTracer.js';
+import { makeBundle, setUpZoeForTest } from '../supports.js';
 
 const pathname = new URL(import.meta.url).pathname;
 const dirname = path.dirname(pathname);
 
 const psmRoot = `${dirname}/../../src/psm/psm.js`;
 const trace = makeTracer('TestPSM', false);
 
-const makeBundle = async sourceRoot => {
```
turadg, line 30: 👍 🧹

packages/run-protocol/test/supports.js
```diff
@@ -26,4 +43,77 @@ export function makeFakeInnerVault(
...
+};
+harden(makeBundle);
+
+// Some notifier updates aren't propagating sufficiently quickly for the tests.
+// This invocation (thanks to Warner) waits for all promises that can fire to
+// have all their callbacks run
+export const waitForPromisesToSettle = async () => {
+  const pk = makePromiseKit();
+  setImmediate(pk.resolve);
+  return pk.promise;
```
turadg, line 68: we don't need Agoric-specific code to do this.
```suggestion
  return new Promise(resolve => setImmediate(resolve));
```

Simple but still worthwhile to have as a function because it's such a common need in tests. Though it does speak to the value of a _general_ testing support package. I'd like to tackle that in the MN-1.1 timeframe.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1,54 +1,60 @@
-// @ts-check
-/* global setImmediate */
+// no @ts-check
```
turadg, line 1: Once we flip from "opt in" to "opt out" defaults then the above absence of pragma would be checked. 
The way to opt out:
```suggestion
// @ts-nocheck
```

turadg: My guess is you disabled it because `t.context` isn't typed. I'd suggest disabling just those errors so we don't lose type assistance on the rest of the file.

You could also type the context like https://github.com/Agoric/agoric-sdk/blob/d5d1b565f5803e8209425d04b9c0b9b244e42e36/packages/run-protocol/test/swingsetTests/vaultFactory/test-vaultFactory.js#L20-L23

turadg (commented):
Took a pass because it's draft. Before final review I'd ask that the test refactoring and the runtime changes be presented in separate PRs. The runtime changes seem intended not to change functionality so I'd have more confidence if no tests were touched while doing that.

packages/run-protocol/test/runStake/test-runStake.js
```diff
@@ -68,29 +69,13 @@ test.before(async t => {
...
-harden(setUpZoeForTest);
-
 export const setupBootstrap = async (
   bundles,
   timer = buildManualTimer(console.log),
   zoe,
 ) => {
   if (!zoe) {
-    ({ zoe } = await setUpZoeForTest());
+    zoe = await setUpZoeForTest().zoe;
```
dckc, line 78: does that mean the same thing? the former awaits the result of `setUpZoeForTest()` and then gets the `.zoe` property; the latter gets the `.zoe` property and then awaits it, no?

packages/run-protocol/test/supports.js
```diff
@@ -26,4 +43,77 @@ export function makeFakeInnerVault(
...
-}
+};
+
+/**
+ *
+ * @param {string} sourceRoot
+ * @returns {Promise<SourceBundle>}
+ */
+export const makeBundle = async sourceRoot => {
+  const url = await importMetaResolve(sourceRoot, import.meta.url);
```
dckc, line 54: this is a little awkward in that the `sourceRoot` has to be relative to this file, not relative to the file in which it occurs.

packages/run-protocol/test/runStake/test-runStake.js
turadg: Good catch. `produce.zoe.resolve(zoe)` now has a fully resolved `zoe` instead of an eref. Reading the types I think `EconomyBootstrapSpace` takes Erefs:
```
    ({ zoe } = setUpZoeForTest());
```
It should work either way but giving an Eref may catch bugs in code that assumes a resolved ref.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
Chris-Hibbert: Yes, `inBefore` will be identical to `amountIn`.

packages/zoe/src/contractSupport/zoeHelpers.js
dtribble: In this cse, I'm copying the type from the underlying type of `zoe.offer`

packages/run-protocol/src/vaultFactory/liquidation.js
dtribble: Accurately representing the arguments as called. But I've now changed the only caller so I should indeed remove it.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: Nope. Will fix.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: I wanted to separate out the renaming through the rest of the code of `runDebt` to `debt` but it's minor.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: My theory is that it will help prevent a future edit that adds a 4th argument from incorrectly binding the 3rd argument rather than the 4th.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: There's no way currently to make that happen. It was more an issue when we were considering incrementally adding more in-progress liquidation assets.

However it's worth noting that this scenario exists because I cannot get the "satisfied/rejected" state from making an offer. We need that API (@erights). That's especially opaque in the AMM and with `offerTo`

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: DRY!  Also, what I want to trace is the value that would be used, so I never want to do additional retrieval work in a `trace` statement.

packages/run-protocol/src/vaultFactory/vaultFactory.js
dtribble: They are for the new liquidation service. So this is prepatory.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: I think this was to stop the infinite attempt to re-liquidate something already in progress of liquidation. Since it's state has changed, it will always fail.

packages/run-protocol/src/makeTracer.js
dtribble: It's used in all the `trace` statements. the version that takes a log source as the first argument is used in test-vaultFactory.

packages/run-protocol/test/runStake/test-runStake.js
dtribble: I *changed* `setUpZoeForTest` to synchronously return a pair of promises. So the change is deliberate. I'm up for alternative naming, etc.  This is to reduce the awaits necessary in various places.

packages/run-protocol/test/supports.js
dtribble: That surprised me, but turned out well because then everything in test was relative to the test directory. Since various projects were reaching into other projects for files, this seemed good. We could have a shared helper that you then wrap inside a per-component test directory.  Which piece is the magic?  `importMetaResolve` of `import.meta.url`?

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: I changed the comment.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: oh yes. This also gives it a name in the trace result.

turadg: I mistyped. What I'm suggesting is:
```suggestion
      trace(`Liq results`, {
        runDebt,
        proceeds: debtorSeat.getAmountAllocated('RUN', runBrand),
      });
```
That's assuming `proceeds` will only be used for diagnostics.

packages/run-protocol/test/runStake/test-runStake.js
dckc: ah. ok.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
turadg: I buy that. But I expect the type checker and unit tests to catch if someone makes that erroneous edit so I personally wouldn't encumber the actual code with it. Not requesting it change.

packages/run-protocol/test/supports.js
dckc: I have a mild preference for using absolute paths for anything that goes between modules. I tend to open-code `new URL('../xyz.js', import.meta.url).pathname`.

But I don't feel strongly about it. I can go with...

> everything in test was relative to the test directory

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: There's too many. It's too confused about the context, and I don't think it needs to be. I'll wait for some expert to straighten that out more generally.

dtribble:
Split into 3 other PRs.

packages/run-protocol/src/makeTracer.js
turadg: Ah, thanks. I was looking for `log` and the arg is `t`.  https://github.com/Agoric/agoric-sdk/pull/5092/files#diff-d7aa780281e2f8cb56ac8c6ac94cfaf827df62ef48478949275d7fe9b204725cR258

Issue #4263, opened 2022-01-09
Make the wallet support getRUN / BLD Boost


samsiegart:
## What is the Problem Being Solved?

For the `getRUN` protocol, dapps (e.g. the treasury) need to be able to propose offers for the wallet to lock and unlock various amounts of BLD, and borrow/repay various amounts of RUN. The problem is that with `getRUN`, the wallet needs to get an attestation for the `give` amount in the offers ([ref](https://github.com/Agoric/agoric-sdk/blob/dc-run-loc/packages/treasury/test/test-runLoC.js#L258)). Also, the BLD purse cannot be selected since it always must be the virtual BLD purse that can contain locked tokens. Due to these differences from a regular offer proposal, the wallet needs a way to know when to handle this separate case.

## Description of the Design

Add a new method to the [wallet bridge API](https://agoric.com/documentation/guides/wallet/api.html#walletbridge-api-commands) called `addGetRUNProposal` that takes a give/want amount of BLD/RUN. It should show an offer in the UI similar to regular offers, but when accepted it should get an attestation and do all of the required steps for `getRUN` (see ref above).

Note this doesn't include the other backend work required to support `makeAttestations` from the wallet, just the API that the dapps will use.

## Security Considerations

The UI should make it clear that tokens are being locked rather than "given" to the contract. It should also reflect the locked amount in the purse.

## Test Plan

Unit testing and e2e testing with the treasury dapp

samsiegart:
I'm trying to decide if it's too clunky to bake this into the wallet API so explicitly, but I'm leaning towards not because it is after all a unique concept that's built into the chain.

samsiegart:
cc @michaelfig @dckc

dckc:
> Add a new method to the wallet bridge API called addGetRUNProposal that takes a give/want amount of BLD/RUN.

I suggest rather than a new method, the wallet uses the existing add API to add a proposal and uses this special code path in case the issuer is the attestation issuer.

samsiegart:
> I suggest rather than a new method, the wallet uses the existing add API to add a proposal and uses this special code path in case the issuer is the attestation issuer.

How would the dapp get the attestation issuer/amount? I wasn't sure if the attestation was a step we'd want to abstract away from the dapps, since it seemed like it was subject to change.

dckc:
[...]

PR #5094, opened 2022-04-12
feat(run-protocol): expose wrapLienedAmount on attestationTool

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/4263


dckc (approved):
good stuff; I'm not sure why I neglected to do this in the first place.

PR #5095, opened 2022-04-12
fix(centralSupply): shutdown after providing bootstrap payment

### LINKED ISSUE #4795: centralSupply contract vat continues to run after bootstrap is done with it
It has access to `feeMintAccess`, i.e. the right to print RUN.

dckc:
closes: #4795

### DIFF
--- a/packages/vats/src/centralSupply.js
+++ b/packages/vats/src/centralSupply.js
@@ -31,7 +31,10 @@ export const start = async (zcf, { feeMintAccess }) => {
 
   return {
     creatorFacet: Far('creator', {
-      getBootstrapPayment: () => bootstrapPayment,
+      getBootstrapPayment: () => {
+        zcf.shutdown('payment retrieved');
+        return bootstrapPayment;
+      },
     }),
   };
 };
--- a/packages/vats/test/test-bootstrapPayment.js
+++ b/packages/vats/test/test-bootstrapPayment.js
@@ -152,3 +152,13 @@ test('bootstrap payment - default value is 0n', async (/** @type {CentralSupplyT
 
   t.true(AmountMath.isEqual(bootstrapAmount, AmountMath.make(runBrand, 0n)));
 });
+
+test('bootstrap payment - contract exits', async (/** @type {CentralSupplyTestContext} */ t) => {
+  const { adminFacet, creatorFacet } = await startContract(t, 0n);
+
+  const pmt = await E(creatorFacet).getBootstrapPayment();
+  t.log({ pmt });
+  const completion = await E(adminFacet).getVatShutdownPromise();
+  t.log({ completion });
+  t.pass();
+});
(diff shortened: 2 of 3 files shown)


dckc:
@warner @Chris-Hibbert any idea why the shutdown promise doesn't resolve?

```
~/projects/agoric/agoric-sdk/packages/run-protocol
14:18 connolly@jambox$ yarn test test/vaultFactory/test-bootstrapPayment.js -m '*exits'
yarn run v1.22.17
$ ava test/vaultFactory/test-bootstrapPayment.js -m '*exits'

  bootstrap payment - contract exits

    ℹ start...
    ℹ get pmt ...
    ℹ {
        pmt: Object @Alleged: RUN payment {},
      }

  Error: Promise returned by test never resolved
    at process.emit (node:events:520:28)
```

dckc:
@warner thanks for the clue about `fakeVatAdmin`. 00ce476 seems to address it; what do you think?

packages/run-protocol/src/centralSupply.js
```diff
@@ -31,7 +31,10 @@ export const start = async (zcf, { feeMintAccess }) => {
 
   return {
     creatorFacet: Far('creator', {
-      getBootstrapPayment: () => bootstrapPayment,
+      getBootstrapPayment: () => {
+        zcf.shutdown('payment retrieved');
+        return bootstrapPayment;
```
Chris-Hibbert: shutdown is synchronous and seems to kill the containing vat. Can the function return anything after that? Does this only work in tests?

It's also counter to our approach for contracts to return payments from API methods rather than as payouts. Why are we doing it this way?

Chris-Hibbert (changes requested):

packages/run-protocol/src/centralSupply.js
dckc: > It's also counter to our approach for contracts to return payments from API methods rather than as payouts. Why are we doing it this way?

I don't know. It predates this PR.

> shutdown is synchronous and seems to kill the containing vat. Can the function return anything after that?

If not, our other contracts don't work either; for example, `atomicSwap.js`:

```js
    const matchingSeatOfferHandler = matchingSeat => {
      const swapResult = swap(zcf, firstSeat, matchingSeat);
      zcf.shutdown('Swap completed.');
      return swapResult;
    };
```

`sellItems.js`:
```js
    if (AmountMath.isEmpty(getAvailableItems())) {
      zcf.shutdown('All items sold.');
    }
    return defaultAcceptanceMsg;
```

warner:
> @warner thanks for the clue about `fakeVatAdmin`. [00ce476](https://github.com/Agoric/agoric-sdk/commit/00ce476e922033cf11f1c7ae18cec8240115bb2f) seems to address it; what do you think?

That looks like a good implementation of `vatPowers.exitVat` and `adminNode.done`.

packages/run-protocol/src/centralSupply.js
Chris-Hibbert: I'd be surprised if anyone is currently looking for those results. 

OTOH, someone certainly wants the bootstrapPayment. If that's accessible outside of tests, then it works.

dckc: > Why are we doing it this way?

@dtribble reminds us: offer handling used to have fees etc. that complicate things, and the plan is that they will again in due course.

Chris-Hibbert (approved):
LGTM.

Thanks for answering my questions

PR #5097, opened 2022-04-12
Split single- and multi-faceted VO definitions into their own functions

### LINKED ISSUE #5093: Split single- and multi-facet virtual object definition into separate API calls

Currently, `defineKind` and `defineDurableKind` each take a behavior parameter that specifies the shape of the VDO being define.  This object may be either a bag of method functions or a bag of facets, each of which is in turn a bag of method functions.  Several people have voiced concerns about the overloading of the behavior parameter makes type declarations very tricky and make the whole API harder to document and less clear in usage.

## Description of the Design

Instead, we have decided to break out the single- vs. multi-facet kind definitions into their own API entry points, so we will split `defineKind` into `defineKind` and `defineKindMulti`, and likewise split `defineDurableKind` into `defineDurableKind` and `defineDurableKindMulti`, where the first of each pair produces a single-facet VDO and expects `behavior` to _always_ be a bag of method functions, and the second of each pair produces a multi-facet VDO and expects `behavior` to _always_ be a bag of facets.

## Security Considerations

This is purely an ergonomic change and introduces no new security issues.

## Test Plan

Existing tests should be mostly sufficient, once updated to incorporate the single- vs. multi- distinction in the relevant places.  Possibly it might also be good to add a test of behavior parameter validation, e.g., to catch cases where older code in need of update passes a facet bag to `defineKind`.

FUDCo:
Closes #5093

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -156,8 +156,9 @@ export function makeCache(size, fetch, store) {
  * not need to occupy memory when they are not in use.  It provides five
  * functions:
  *
- * - `defineKind` and `defineDurableKind` enable users to define new types of
- *    virtual object by providing an implementation of the new kind of object's
+ * - `defineKind`, `defineKindMulti`, `defineDurableKind`, and
+ *    `defineDurableKindMulti` enable users to define new types of virtual
+ *    object by providing an implementation of the new kind of object's
  *    behavior.  The result is a maker function that will produce new
  *    virtualized instances of the defined object type on demand.
  *
@@ -175,8 +176,8 @@ export function makeCache(size, fetch, store) {
  *    actually is prior to a controlled shutdown; normal code should not use
  *    this.
  *
- * `defineKind` and `defineDurableKind` are made available to user vat code in
- * the `VatData` global (along with various other storage functions defined
+ * The `defineKind` functions are made available to user vat code in the
+ * `VatData` global (along with various other storage functions defined
  * elsewhere).
  */
 export function makeVirtualObjectManager(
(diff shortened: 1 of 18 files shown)


packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -156,8 +156,9 @@ export function makeCache(size, fetch, store) {
  * not need to occupy memory when they are not in use.  It provides five
  * functions:
  *
- * - `defineKind` and `defineDurableKind` enable users to define new types of
- *    virtual object by providing an implementation of the new kind of object's
+ * - `defineKind`, `defineKindMult`, `defineDurableKind`, and
```
turadg:
```suggestion
 * - `defineKind`, `defineKindMulti`, `defineDurableKind`, and
```

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -564,11 +576,13 @@ export function makeVirtualObjectManager(
     const facetiousness = assessFacetiousness(behavior);
     switch (facetiousness) {
       case 'one': {
+        assert(!multifaceted);
```
turadg, line 579: nice

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -74,11 +74,7 @@ async function testUpgrade(t, defaultManagerType) {
   const hostStorage = provideHostStorage();
   const { kvStore } = hostStorage;
   await initializeSwingset(config, [], hostStorage);
-  const c = await makeSwingsetController(
-    hostStorage,
```
turadg, line 78: assuming this is misc cleanup

turadg (approved):
LGTM!

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: Yes.  Debugging debris, presumably left over from @warner working on upgrade tests.

PR #5104, opened 2022-04-13
chore: Sync Endo versions

kriskowal:
- chore: Sync Endo versions
- chore: Update yarn.lock
- chore(swingset-vat): Update xsnap snapshots


turadg (approved):
Changes look good.

That the commit messages are "conventional" makes me think this may be a merge commit. If so consider cleaning them up. E.g. "fix: Manual canonicalization of yarn.lock" isn't a fix for anything in master.

kriskowal:
My reasoning for separating the manual fix is to clarify that `chore: Update yarn.lock` is purely mechanical, as a signal to reviewers. I could squash it, but perhaps it continues to be a useful hint for review-after-land. I don’t have a strong feeling. I do intend to merge commit this.

PR #5109, opened 2022-04-13
refactor(run-protocol): extract vaultDirector kind from vaultFactory contract

turadg:
## Description

#4345 requires that all the state be in virtual objects. For vault factory, a big step towards that is having the object extracted from the contract. This does that.

It names that kind "director" as the "manager of vault managers". The contract function becomes much lighter, setting up governance, instantiating this director kind and then returning its facets.

The steps of the refactoring are separated by commit so it's probably easiest to review commit-by-commit. If I make changes to them I'll try to use fixup so they're apparent.

### Security Considerations

No change in powers.

New concept of a "director" object but I don't think there were docs on this part of the contract anyway.

### Testing Considerations

Will simplify testing by letting us test the director independent of the whole contract bundle. Future work.


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -0,0 +1,333 @@
...
+            ),
+          };
+        }),
+      ),
+    );
+  };
+
+  /** @type {VaultFactory} */
+  const machineBehavior = {
+    // TODO move this under governance #3972
```
Chris-Hibbert: 3924 is a  better link.

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -0,0 +1,333 @@
...
+    /** @type {AddVaultType} */
+    addVaultType: async (
```
Chris-Hibbert, line 152: `addVaultType` is too big to be inline in `machineBehavior`. It really ought to be top-level.  Since it has internal `await`s, it has to be top-level anyway.

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -300,7 +329,10 @@ export const start = async (zcf, privateArgs) => {
       getGovernedParams: ({ collateralBrand }) =>
         // TODO use named getters of TypedParamManager
         vaultParamManagers.get(collateralBrand).getParams(),
-      getContractGovernor: () => governorPublic,
+      /** @returns {Promise<GovernorPublic>} */
+      getContractGovernor: () =>
+        // ??? does this need to be cached?
```
Chris-Hibbert: It doesn't need to be cached, but it can be.

Chris-Hibbert (approved):
This looks good.

I really want the ability to test our virtual object conversions. `grumble`

dtribble:
My basic reaction is one of horror. I'm sure I'll appreciate it after the shock wears off.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: This is en route to `defineKindMulti` where it will have one level less of nesting. This just made it easier to move.

> Since it has internal awaits, it has to be top-level anyway.

Pretty sure that nesting rule is only about `awaits` not the function definition location. https://github.com/Agoric/agoric-sdk/wiki/No-Nested-Await

PR #5111, opened 2022-04-14
fix: avoid race by requesting lien change as a delta

### LINKED ISSUE #4887: Add compare-and-swap fields in SetLien request

We need to keep lien state in sync between JS and golang data structures. The `setLien` API with just the `target` amount seems to have lost-update risks.

## Description of the Design

Add another denom / amount pair of args to the current message structure:

```
{
        type: XLien.LIEN_GET_ACCOUNT_STATE,
        address,
        denom: XLien.denom,
        amount: '0',
      }
```

## Security Considerations

n/a

## Test Plan

usual unit testing

JimLarson:
Closes: #4887

## Description

Avoid a race in changes to the lien amount by requesting a delta rather than a new balance.

### Security Considerations

Fixes a race that could have allowed too many attestations to be minted.

None.

### Testing Considerations

Needs integration testing with JS side.


JimLarson:
@dckc We need to decide whether to land this first and then update your JS-side marshaling, or whether to land your end-to-end integration first and then have this PR update both sides in a single PR.

JimLarson:
In the meeting today @dckc suggested that I extract the necessary bits from the dc-getRUN-integration branch. However, the changes between that branch and master are .. extensive, and I don't have the familiarity to do such extraction in a timely fashion. Therefore I think we should commit this PR as a golang-only change and integrate later.

dckc (approved):
I only see happy-path tests. A test for a failure mode would be nice. But it's not critical.

JimLarson:
The sad-path tests for setting the liened state are mostly in `keeper_test.go`. We can't test the actual race here, but we might be able to in a JS-side integration test.

michaelfig (approved):
LGTM!  Sorry I didn't get to this for so long.

JimLarson:
Implemented the JS-side changes. My first JS PR, so please examine extra carefully. PTAL @michaelfig @dckc @turadg

packages/run-protocol/src/my-lien.js
```diff
@@ -30,9 +30,32 @@ export const makeStakeReporter = (bridgeManager, stake, denom = 'ubld') => {
...
+   * @param {bigint} delta
+   * @returns {Promise<Amount<`nat`>>}
+   */
+  const changeLiened = async (address, delta) => {
+    assert.typeof(address, 'string');
+    const newAmount = await E(bridgeManager).toBridge(XLien.name, {
+      type: XLien.LIEN_CHANGE_LIENED,
+      address,
+      denom,
+      delta: `${delta}`,
```
turadg, line 44: what happens if this isn't stringified?

if `toBridge` requires the second argument to be `Record<string, string>` let's annotate that.

packages/run-protocol/src/my-lien.js
```diff
@@ -30,9 +30,32 @@ export const makeStakeReporter = (bridgeManager, stake, denom = 'ubld') => {
...
+    });
+    return harden(toStake(newAmount));
+  };
 
   /** @type {StakingAuthority} */
   const stakeReporter = Far('stakeReporter', {
+    increaseLiened: async (address, increase) => {
+      const delta = AmountMath.getValue(stake, increase);
```
turadg: precedes this PR but `stake` reads like an amount, whereas it's the brand that is to be staked.

I think it's the only brand in scope so imo a (non-blocking) improvement would be to rename it `brand`.

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -95,11 +95,15 @@ const makeAttestationIssuerKit = async (zcf, stakeBrand, lienBridge) => {
     return { address, lienedAmount };
   };
 
-  /** @type {Store<Address, Amount<'nat'>>} */
+  /**
+   * Non-authoritative store of nonzero lien amounts.
```
turadg, line 85: 🙏  documenting

packages/run-protocol/src/runStake/README.md
```diff
@@ -40,8 +40,8 @@ sequenceDiagram
     dapp ->> walletBridge: getReturnableAttestation(want: 450 RUN, give: 500 BLD-Att)
     note right of walletBridge: Blocks on user approval in wallet
     walletBridge ->> attestation: makeAttestation(500 BLD)
-    attestation ->> Cosmos_SDK: setLienedAmount(4000 + 500 BLD)
-    Cosmos_SDK -->> attestation: ACK or throws
+    attestation ->> Cosmos_SDK: changeLienedAmount(+500 BLD)
```
turadg: `changeLiened` isn't exported
```suggestion
    attestation ->> Cosmos_SDK: increaseLiened(500 BLD)
```

packages/run-protocol/test/attestation/test-attestation.js
```diff
@@ -26,23 +24,29 @@ export const makeMockLienBridge = (uBrand, _t) => {
...
+    const currentAmount = liened.get(address) || empty;
+    const current = AmountMath.getValue(uBrand, currentAmount);
+    const changed = current + delta;
+    const changedAmount = AmountMath.make(uBrand, changed);
+    // t.log('changeLiened:', { address, changedAmount });
+    liened.set(address, changedAmount);
+    return changedAmount;
+  };
   return Far('stakeReporter', {
+    increaseLiened: (address, increase) => {
```
turadg, line 37: this and the other mock `stakeReporter` are repeating a lot of the application logic. We've been burned by stale or inaccurate mocks. Would it be feasible to just mock `bridgeManager` and use the real `makeStakeReporter`?  cc @dckc

turadg (commented):
Reviewed for clarity and maintainability. Ignored Go and the bridging logic.

michaelfig (approved):
Golang and bridge still LGTM!

packages/run-protocol/test/attestation/test-attestation.js
dckc: I can see the appeal. It's probably feasible but more than 20 minutes of work. I don't feel strongly about whether it's worthwhile.

dckc (approved):
looks good.

https://github.com/Agoric/agoric-sdk/issues/4887#issuecomment-1165100248 agrees with my memory of why we picked this approach.

packages/run-protocol/test/attestation/test-attestation.js
turadg: Yeah, I gave it a try and it's more work than I thought. Withdrawing the suggestion.

packages/run-protocol/src/my-lien.js
JimLarson: Done.

packages/run-protocol/src/runStake/README.md
JimLarson: Done.

packages/run-protocol/src/runStake/attestation.js
JimLarson: De nada!

packages/run-protocol/src/my-lien.js
JimLarson: I *think* the bridge device just JSON-ifies whatever data is passed. I started down the rabbit hole to see where this goes and got lost. In any case, I'm confident that I'm following the established pattern from the old version, and I delegate further questions to the real experts.

JimLarson:
@turadg PTAL

turadg (approved):

JimLarson:
Thanks, @turadg ! I'm going to try to do an end-to-end test before merging.

JimLarson:
Thanks, @dckc and @samsiegart for helping me do the end-to-end test!

PR #5116, opened 2022-04-14
refactor(run-protocol): vaultDirector to multifacted vobj

### LINKED ISSUE #4345: Use Virtual Objects in Vaults

turadg:
closes: #4345

## Description

Turns the new vaultDirector into a virtual object.

### Security Considerations

State storage to to disk.

--

### Testing Considerations
Refactor so it maintains existing tests AMAP, though I did run into a test bug I fixed in 75e3e7

It took me too long to figure out [1c87554](https://github.com/Agoric/agoric-sdk/pull/5116/commits/1c8755466b163cb9d83e9ca22c45e1e63101013f) so I think we need more test coverage or some static type validation of what is returned by facets. The problem wasn't detected in unit tests, only integration tests, and was much harder to debug there.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/run-protocol/src/vaultFactory/vaultDirector.js
@@ -16,7 +16,8 @@ import { makeRatioFromAmounts } from '@agoric/zoe/src/contractSupport/ratio.js';
 import { Far } from '@endo/marshal';
 
 import { AmountMath } from '@agoric/ertp';
-import { assertKeywordName } from '@agoric/zoe/src/cleanProposal';
+import { assertKeywordName } from '@agoric/zoe/src/cleanProposal.js';
+import { defineKindMulti } from '@agoric/vat-data';
 import { makeVaultManager } from './vaultManager.js';
 import { makeLiquidationStrategy } from './liquidateMinimum.js';
 import { makeMakeCollectFeesInvitation } from '../collectFees.js';
@@ -34,6 +35,7 @@ const { details: X } = assert;
  * debtMint: ZCFMint<'nat'>,
  * collateralTypes: Store<Brand,VaultManager>,
  * electionManager: Instance,
+ * electorateParamManager: import('@agoric/governance/src/contractGovernance/typedParamManager').TypedParamManager<{Electorate: "invitation"}>,
  * mintSeat: ZCFSeat,
  * penaltyPoolSeat: ZCFSeat,
  * rewardPoolSeat: ZCFSeat,
@@ -46,6 +48,11 @@ const { details: X } = assert;
  *  getGovernedParams: () => import('./vaultManager.js').GovernedParamGetters,
  *  mintAndReallocate: MintAndReallocate,
  * }} FactoryPowersFacet
+ *
+ * @typedef {{
+ *   state: ImmutableState;
+ *   facets: import('@agoric/vat-data/src/types').KindFacets<typeof behavior>;
+ * }} MethodContext
  */
 
 /**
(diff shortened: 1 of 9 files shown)


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -74,260 +81,296 @@ const initState = (zcf, electorateParamManager, debtMint) => {
...
-      initialParamValues,
-    ) => {
-      await zcf.saveIssuer(collateralIssuer, collateralKeyword);
-      const collateralBrand = zcf.getBrandForIssuer(collateralIssuer);
-      // We create only one vault per collateralType.
-      assert(
-        !collateralTypes.has(collateralBrand),
-        `Collateral brand ${collateralBrand} has already been added`,
-      );
+// TODO put on machineFacet for use again in publicFacet
```
turadg, line 115: what I went with is defining this at top scope so it can be shared.

thoughts on that vs giving it a primary facet and letting another reexport?

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -74,260 +81,296 @@ const initState = (zcf, electorateParamManager, debtMint) => {
...
+            getAmountIn(priceQuote),
+          ),
+        };
+      }),
+    ),
+  );
+};
+
+/** @type {import('@agoric/vat-data/src/types').FunctionsPlusContext<VaultFactory>} */
+const machineBehavior = {
```
turadg, line 145: the functions within this behavior could be at module-scope. That's how @FUDCo had in the initial defineKind-ization of vault.js. I propose a style of defining behavior objects with the methods inlined so it's easier to navigate.

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -74,260 +81,296 @@ const initState = (zcf, electorateParamManager, debtMint) => {
...
-        getGovernedParams: () => ({
-          ...vaultParamManager.readonly(),
-          getChargingPeriod: () => loanTimingParams[CHARGING_PERIOD_KEY].value,
-          getRecordingPeriod: () =>
-            loanTimingParams[RECORDING_PERIOD_KEY].value,
-        }),
-        mintAndReallocate,
-        burnDebt,
-      });
+    const factoryPowers = Far('vault factory powers', {
```
turadg, line 245: will defining a far per vault manager pose any problems for durability @FUDCo ?

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -74,260 +81,296 @@ const initState = (zcf, electorateParamManager, debtMint) => {
...
-    getRewardAllocation: rewardPoolSeat.getCurrentAllocation,
-    getPenaltyAllocation: penaltyPoolSeat.getCurrentAllocation,
-  };
+    ).makeCollectFeesInvitation();
+  },
+  /** @param {MethodContext} context */
+  getContractGovernor: ({ state }) => state.zcf.getTerms().electionManager,
 
-  const machineFacet = Far('vaultFactory machine', machineBehavior);
+  // XXX accessors for tests
```
turadg, line 283: once we have a way to instantiate a vobj from a state fixture, we don't need these methods. we can just inspect the state after the operations.

packages/run-protocol/src/vaultFactory/vaultDirector.js
FUDCo: I had originally had the behavior functions in `vault.js` at module scope simply because it was the most similar to what had been there before.  @turadg asked me to rewrite them to be inline and I think I like the inline style better in that case.  I don't know how it translates to this file however.

packages/run-protocol/src/vaultFactory/vaultDirector.js
FUDCo: On general principles it should not be a problem, but the key question is: will you be able to reconstruct the Far from the durable state after resurrection?  (I can't discern the answer to that myself that without digging into this code considerably beyond my feeble understanding of what it's doing.)  If you can, then it should be no problem; you just need to make sure your startup flow actually does it.  Otherwise you'll either need to (a) capture more info in the durable part so that you _can_ reconstruct it or (b) make it durable itself.  I don't have an opinion which is more right, as it would depend on the particulars, other than the general advice to take whichever course causes you the least pain.

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: both approaches seem fine where appropriate. This seems sufficient for this use. I can think of other places where we might want a facet.  (I'm considering that for the AMM's singlePool, but that has more justification to be a facet.)

Chris-Hibbert (approved):

PR #5118, opened 2022-04-15
chore: Sync Endo versions

kriskowal:
This brings in a `bundle-source` bin with cache support.


dckc (approved):
I'm curious about the manual edit to yarn.lock. point me to the history?

kriskowal:
@dckc I was curious too, so I upgraded `yarn` after this change and it seems more consistent now.

Issue #5106, opened 2022-04-13
fix vdata promise handling (was: Loadgen triggers `I don't remember allocating`)


mhofman:
## Describe the bug
The loadgen is able to trigger a liveslot `I don't remember allocating` error, which should never happen. This seems to only happen on a very specific revision of agoric-sdk: cbfdcc99

## To Reproduce
1. `git clone https://github.com/Agoric/testnet-load-generator.git`
2. `cd testnet-load-generator`
3. `git checkout e4269620ea768cc976289e5cb9f77e28dde81a3f` (current main at the time of this issue)
4. `SDK_REVISION=cbfdcc99 ./start.sh --stages=1 --stage.0.duration=0` (No need to run any load, it's the setup that fails)

## Expected behavior
No liveslots error

## Platform Environment
 - what OS are you using? what version of Node.js? N/A
 - is there anything special/unusual about your platform? Debian
 - what version of the Agoric-SDK are you using? `cbfdcc99c3ac14fe5153e716408aa436cd9ded7b`

## Additional context
The exact same PR stacked on top of a few other commits does not trigger this issue: See 1c354ca7e12343f61f4a0cbd93cfd9b4636c23a1

The failed run artifacts, including slogs are available here: https://github.com/Agoric/agoric-sdk/actions/runs/2157824567

Running with the above command runs a local chain which produces more details however.

## Screenshots
```
2022-04-13T18:11:23.145Z SwingSet: ls: v22: Error#1: I don't remember allocating p+35
2022-04-13T18:11:23.145Z SwingSet: ls: v22: Error: I don't remember allocating (a string)
 at construct ()
 at Error (/workspaces/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:7062)
 at makeError (/workspaces/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:1207)
 at fail (/workspaces/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:1335)
 at baseAssert (/workspaces/agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:1353)
[...]

PR #5122, opened 2022-04-16
fix(run-protocol): store Presences, not Promises, in VaultManager

### LINKED ISSUE #5121: change vaultFactory (and run-protocol?) to require settled `priceAuthority` and `timerService`, not promises

#5106 is a swingset bug that's triggered by userspace code storing Promises in virtual data (`state` properties of virtual objects, or virtual collections). I'm working on fixing swingset, but I don't know that I can finish it in time for next week's deadline.

Also, you can't store Promises in *durable* objects/collections, so any contract code doing that will probably need to change before it can be made upgradable.

The one case where we do this (vaultFactory) looks like it'll be pretty easy to convert now, which means I can deprioritize fixing the swingset bug, and also it brings us closer to durable-izing / upgrade-able-izing the contract.

## Description of the Design

The main thing is that the `state` of `makeVaultManagerKit` needs to have arguments which are `PriceAuthority` and `TimerService`, not `ERef<PriceAuthority>`/etc. I'm looking to see how far upstream that can/should be pushed. Something before the call to `makeVaultManager()` needs to `await`  both arguments, to turn them into a Presence rather than a Promise.

cc @dtribble @turadg

warner:
To avoid triggering a swingset bug (#5106), and to prepare for turning
our virtual Kinds into durable Kinds, we need to not store Promises in
virtualized data. In addition, storing a Presence instead of a Promise
can provide a small performance improvement.

Currently, the only case where we store Promises in vdata is in
VaultManager, which stores Promises for `priceAuthority` and
`timerService` in its `state` object.

This changes the setup code to `await` both objects before handing
them to `makeVaultManager`, to ensure that we store Presences, not
Promises.

closes #5121
refs #5106

### DIFF
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -283,18 +283,19 @@ export const startVaultFactory = async (
     ]);
   const ammPublicFacet = await E(zoe).getPublicFacet(ammInstance);
   const feeMintAccess = await feeMintAccessP;
-
+  const pa = await priceAuthority;
+  const timer = await chainTimerService;
   const vaultFactoryTerms = makeGovernedTerms(
-    priceAuthority,
+    pa,
     loanParams,
     installations.liquidate,
-    chainTimerService,
+    timer,
     invitationAmount,
     vaultManagerParams,
     ammPublicFacet,
   );
   const governorTerms = harden({
-    timer: chainTimerService,
+    timer,
     electorateInstance,
     governedContractInstallation: installations.VaultFactory,
     governed: {


packages/run-protocol/src/econ-behaviors.js
```diff
@@ -281,18 +281,19 @@ export const startVaultFactory = async (
     ]);
   const ammPublicFacet = await E(zoe).getPublicFacet(ammInstance);
   const feeMintAccess = await feeMintAccessP;
-
+  const pa = await priceAuthority;
+  const timer = await chainTimerService;
```
turadg, line 287: we could be awaiting all those in the `Promise.all` to start in parallel but that does obscure readability.

I figure this whole file will get some cleanup soon.

turadg (approved):

Issue #5049, opened 2022-04-08
Install latest chain platform software in devnet on April 18


Tartuffo:
## What is the Problem Being Solved?

In preparation for the RUN Protocol preview release on April 22 (#5062), we plan to install the latest chain platform software in devnet on April 18.

## Description of the Design

Address the following blocker issues:

 - #4564 
 - #4911
 - #5128 
 - #5094 
 - #5029 

Then release as usual following [internal testnet / devnet release process](https://github.com/Agoric/agoric-private/wiki/Testnet-Release-Process).

## Security Considerations

No impact on production.

## Test Plan

TBD

dckc:
"latest" now? Or "latest" with some in-progress features? If so, which features? What's in the critical path?

In particular, I gather hash bundles are planned; I'm not sure which issue captures that feature; #4974 is at least part of it.

Also, at least one of the solutions we are considering for #5029 involves a platform change (bootstrap vat), so unless / until we choose another design, that one seems critical path.

dckc:
It's alive!

https://devnet.explorer.agoric.net/blocks/2
4F1529001AD114859965C6B2A8B71770646137C167C84F83C7319AD1402D5F52
19 Apr 2022, 12:13:52am UTC (12 minutes ago)

https://github.com/Agoric/agoric-sdk/pull/5142

Issue #5029, opened 2022-04-06
RUN bootstrap packaging does not support post-genesis deployment


dckc:
## Describe the bug

The bootstrap code for runStake is in econ-behaviors.js now, but that code post-dates agoricdev-8, so to deploy runStake to the current devnet (#4348) I had to re-package the relevant code (including stuff like `AmountMath.make`) as an evaluatable script. The result was 100 or so lines of code with no automated tests.

Deploying contracts that weren't available at genesis is more likely the rule than the exception, so we should figure out a maintainable way to do it.

## To Reproduce
n/a?

## Expected behavior

governance proposals to start run protocol contracts should consist of no more than a few lines of JS that hasn't been through automated tests.

## Platform Environment

agoricdev8
more details are in #4348

## Additional context

@michaelfig and I started brainstorming ideas involving the endo bundler, endoCAS, integration tests similar to agoric-cli, etc.

dckc:
An earlier (Feb 27) version of the runStake governance proposals shows a bit of flailing: https://github.com/Agoric/agoric-sdk/commit/9173b19da7d349366a7005eae528c23b5fd4d3df

dckc:
I had to use the `getRUN` slot in agoricNames because the shape of `agoricNames` is decided at chain bootstrap.

dckc:
@michaelfig has an idea to re-use installation infrastructure to address this.

dckc:
discussion with @michaelfig :
I'll add `evaluateInstallation` (getBundle, importBundle - but might need a device later) available in the bootstrap args

Issue #3609, opened 2021-08-06
prototype caching/construction of bundles on filesystem


warner:
## What is the Problem Being Solved?

We have a lot of generated bundles in our system. Within swingset itself, we've got the main kernel bundle and a pair (SES/lockdown and a supervisor) for `xsnap` instances. Swingset has a number of "system vats" (timer, vat-admin, comms) which are used in most every application. Zoe has ZCF. All of our main contracts (treasury, etc) have a contract bundle.

All of these start as an entry-point file (with some default export), which then references plenty of others via `import` statements. They get converted into a serializable "bundle" artifact via the `bundleSource()` tool. They then take various pathways into the code, but all wind up as the argument of an `importBundle()` call.

During unit tests, we generally run `bundleSource()` as late as possible, during the individual test program, so that any code changes are incorporated into the test's execution. We mostly avoid requiring a distinct "build" step because of the expensive confusion incurred by forgetting to run it ("I'm changing my code but it doesn't have any effect!").

But bundling the source takes a long time, nearly 3 seconds for the full set of kernel bundles (it took significantly longer when we were applying the metering transform too). I've changed a lot of kernel tests to make these bundles in a `test.before` block (amortized among all tests in the same file). The Zoe tests resorted to a separate `yarn build` step (included in `yarn test`) to make it faster: this is one of the few cases where the bundle is written to disk, and imported/read by each test which needs it, rather than being built in only in RAM.

After doing some measurements to see whether it's currently justified or not, what I'd like to consider is a `make`-like tool to generate the bundles on-demand but keeps them on disk, with intelligent dependency checking to avoid rebuilds unless necessary. The hope would be to construct each bundle exactly once per code change, removing the overhead when iterating on a test that doesn't change the files that make up the bundle.

It's possible that the new Endo zipfile-based bundler will be faster than `rollup`, making this unnecessary. On the other hand it might be slower.

cc @kriskowal 

## Description of the Design

I'm thinking about something like:

* `bundleSource` exports a function that takes an entry point filename and an output bundle filename, and returns a Promise for a bundle object
  * despite the shared written-bundle location, it is safe to call this function simultaneously from multiple (parallel) AVA tests
* the function will use the output bundle file (and a couple of related+nearby files) to coordinate simultaneous callers
  * if instance 1 is first on the scene, it will leave a marker to claim the construction process
  * all other instances will recognize the marker and wait for instance 1 to finish building the bundle
[...]

PR #5123, opened 2022-04-16
feat: separate genesis from RUN protocol installation

dckc:
refs: #5029

## Description

The bootstrap behaviors in `@agoric/run-protocol` assumed that deployment and genesis were the same time.

This PR clearly separates genesis bootstrap from later governance actions where RUN protocol contracts are given access to bootstrap powers such as the RUN mint.

It should also facilitate using modules in swingset-core-eval proposals by making more use of installations and less bundles.

Meanwhile, I reached my tolerance for waiting for bundling over and over, so this PR also includes a prototype of bundle caching (#3609).

### Security Considerations

This change _should_ continue to separate production mode from demo, but as there are high-power authorities in play, careful review is in order.

It's not clear that the sim chain survives this change in tact. We may need to notify devs to stand by until we get that part put back together.

_The plan is to make a separate bootstrap vat, further separating demo / simulation mode from production mode. The RUN protocol contract bundles may be present in the sim bootstrap vat, but not in production._

IOU issues for:
 - [x] refactoring the overlap between `bundleTool.js` here and `tool.js` in `@endo/bundle-source` (noted in https://github.com/Agoric/agoric-sdk/issues/3609#issuecomment-1100762465 )
 - [x] restoring the sim chain to working order #5128 

The convention for paths for contract bundle roots changes to **package-relative**.

### Testing Considerations

Restoring previous tests to working order is the goal of this PR.

Some other kicking the tires is in order before the upcoming devnet launch (#5049). Ideally we'll develop some integration tests for launching RUN protocol by cosmos governance.


turadg (approved):
`/run-protocol` changes LGTM

Issue #4318, opened 2022-01-18
High priority economic elements (e.g., oracles) can have priority execution


rowgraus:
We need to identify all the messages involved in the core economy and make sure they get priority. This includes:
* cosmos transactions containing price updates from oracles
* device (possibly including vattp/comms) messages that deliver those updates to a vat handling economic activity
* sale orders to liquidate vaults that are below their collateral threshold
* reactions such as promise resolutions from all of those

The "cosmos transactions" part happens before swingset sees anything. @dtribble suggested that the price oracles' addresses be treated specially by cosmic-swingset: we want any txns signed by those addresses to get pulled from the mempool and into a block first, then look for txns by other senders. I don't know quite where this would need to get implemented, @michaelfig has mentioned that the priority of mempool txn selection isn't easy to control.

It should be possible to funnel these messages into SwingSet through dedicated devices, with their handling vat running at high priority, thus bypassing contention with other non-priority messages going through mailboxes/vattp/comms. As long as the reaction from that point all target high priority vats, it should grant the ability to complete processing of high priority economic messages before regular / low priority messages.

The following steps are part of this goal:
- https://github.com/Agoric/agoric-sdk/issues/5023
- https://github.com/Agoric/agoric-sdk/issues/4542
- https://github.com/Agoric/agoric-sdk/issues/5024
- https://github.com/Agoric/agoric-sdk/issues/5025
- https://github.com/Agoric/agoric-sdk/issues/3465
- https://github.com/Agoric/agoric-sdk/issues/5026
- https://github.com/Agoric/agoric-sdk/issues/5027
- https://github.com/Agoric/agoric-sdk/issues/4317

Tartuffo:
@warner For proper project planning and tracking, this needs an area label covered by one of our weekly planning meetings.  Please pick the appropriate one from: agd, agoric-cli, agoric-cosmos, amm, core economy, cosmic-swingset, endo, ertp, getrun, governance, installation-bundling, metering, oracle, pegasus, run-protocol, ses, staking, swingset, swingset-runner, tc39, token economy, tooling, ui, wallet, xsnap, zoe, zoe contract

mhofman:
If cosmic-swingset is able to extract a priority from a message when grabbing it from the mempool, we may be able to have a different mailbox device queue per priority on the swingset side.

[...]

PR #5124, opened 2022-04-16
Restrict promise redirect

### LINKED ISSUE #5023: Constraint decider changes to simplify promise enqueuing logic
Technically the kernel allows a decider to got from the exporting vat, to the kernel, to another vat, back to the kernel, then to another vat, etc. The decider can change one of 2 ways:
- An existing promise is used as the result of a send
- A promise is forwarded to another promise

In practice neither can currently happen:
- liveslots and comms vat always create a new promise
- Forwarding is not implemented

## Description of the Design
If we assume that the decider will never change more than vat exporting new promise -> kernel -> receiver of message tied to result promise, then messages sent to the promise can only ever be placed on the promise queue (while the kernel is decider, or if the decider vat doesn't support pipelining), or the decider pipelining vat (aka comms only for now), and will only ever move from the first to the second.

This ties in with #4542 so that messages can be enqueued to the correct queue right away.

In the future, when forwarding is supported, simply placing a forwarding on the vat outbound queue will need to synchronously pluck messages sent to the promise that were sitting on the pipelining vat's inbound queue, and move them back to the kernel's promise's queue until the forwarding is later processed from the outbound queue.

## Security Considerations
Audit kernel and liveslot code which deals with decider changes, and add assertions as necessary.

## Test Plan
Update existing unit tests and add new negative tests.

mhofman:
closes: #5023
refs: #4318

Best reviewed commit-by-commit

## Description

This changes the kernel and comms promise handling logic to enforce 2 basically existing invariants:
- A promise cannot be resolved to another promise. That would be a redirect, which isn't currently implemented
- A non-pipelining decider vat should not be able to use an imported promise as result of a send

The goal is to limit the situations where messages sent to a promise need to change queues. The only case should now be when a message is sent to a result promise decided by a pipelining vat. Such a message can change either
- from the promise queue to the vat's queue because the send for which the promise is the result has been queued to the vat, or
- from the vat's queue to either the resolved target's vat, or back to the promise queue (if the send for which the promise is the result was just forwarded back).

### Security Considerations

There shouldn't be any. All now explicitly disallowed behaviors were not currently happening in legitimate flows.

TODO

### Testing Considerations

Both positive and negative tests added for promise resolve as well as re-using of imported result promises.


mhofman:
@warner I have a little bit of doc to write, and one test to add that I just thought about, but I would already appreciate a preliminary review.

mhofman:
This is actually ready for review. I still need to see if any doc needs updating. I also looked into a more thorough pipeline test over the weekend, but realized that because of the current kernel runqueue, it wouldn't test anything useful, but I'll it for the next step.

packages/SwingSet/src/kernel/vatTranslator.js
```diff
@@ -255,12 +258,20 @@ function makeTranslateVatSyscallToKernelSyscall(vatID, kernelKeeper) {
...
     if (resultSlot) {
       insistVatType('promise', resultSlot);
-      result = mapVatSlotToKernelSlot(resultSlot);
+      // Require that the promise used by the vat to receive the result of this
+      // send is allocated by the vat if it does not support pipelining.
+      // When a promise previously received as the result of a delivery is used
+      // as the result of a send call, it's in effect doing a redirect, which
+      // is currently only supported for pipelining vats.
+      result = mapVatSlotToKernelSlot(resultSlot, {
+        requireVatAllocated: !enablePipelining,
```
warner: I think this also allows the case where the vat is re-using a vpid that it previously exported, and I'm not sure you want that (but I might be wrong, I need to think about it more carefully). The `allocatedByVat` flag is not a direct indicator of the history of the promise, although possible cases are probably limited (and limited even further by this new assertion).

We might be able to use `allocatedByVat` to achieve the goal, but maybe the new option should be more like "this must be a brand new promise"? (Note that I just filed #5189 and in that case we might see the vpid in `args` before we see it as `result`, which could cause a false negative for a "must be new" check.. we might salvage that by processing `result` before we translate `args`).

warner: Yeah, thinking about it more, I think the check should be `requireNew`.

When one of us fixes #5189, the new tests (where userspace includes the result promise in the args of the same message) will fail this `requireNew` check, and we'll fix it (both here and in liveslots) by translating the result before translating the arguments.

```js
const args = [];
const p = E(target).foo(args);
args.push(p);
```

warner (changes requested):

packages/SwingSet/src/kernel/vatTranslator.js
mhofman: Happy to revert to the initial "fresh" / `requireNew` approach. After realizing that pipelining vats needed the flexibility, I no longer saw a reason to prevent regular vats from pre-allocating a promise (it is currently marked as supported in the docs).

> The `allocatedByVat` flag is not a direct indicator of the history of the promise

Just to make sure I understand this flag correctly, is there any way for `allocatedByVat` to be `true` if the promise was in fact imported and not exported? That's the looser restriction I had settled on after moving away from "fresh", which seemed to satisfy all documented usages (and tests).

packages/SwingSet/test/test-vpid-kernel.js
```diff
@@ -381,7 +381,8 @@ async function doTest123(t, which, mode) {
 
 for (const caseNum of [1, 2, 3]) {
   for (const mode of modes) {
-    test(`kernel vpid handling case${caseNum} ${mode}`, async t => {
+    const doTest = caseNum === 3 ? test.failing : test;
+    doTest(`kernel vpid handling case${caseNum} ${mode}`, async t => {
```
mhofman: @warner, I'm not quite sure what to do about test 3 & 7.

mhofman:
@warner PTAL

I changed `requireVatAllocated` into `requireNew` and updated tests accordingly, as well as added a test which represents what #5189 would end up as in syscalls: 50ec896

I will extract 5c98f5b and 4bd30d9, as it's the trigger for the GC weirdness I experience described in #5202, into a separate PR

packages/SwingSet/src/kernel/vatTranslator.js
warner: You're correct, if kernel is the one introducing the promise into the vat, then `allocatedByVat` will always be false. It's just that I avoid using "import" and "export" for Promises, since that's not nearly as meaningful as which side is the decider. For *objects*, import-vs-export is a big deal, but not so much for Promises.

packages/SwingSet/docs/delivery.md
```diff
@@ -621,8 +638,9 @@ immediately after translation).
 The subject of a `syscall.resolve()` must either be a new Promise ID, or a
 pre-existing one for which the calling Vat is the Decider. The
 `KernelPromise` must be in the `Unresolved` state. If any of these conditions
-are not met, the Vat calling `resolve` will be terminated. It doesn't matter
-whether the Promise was allocated by this Vat or a different one.
+are not met, the Vat calling `resolve` will be terminated. Furthermore, the
+promise can only have been allocated by another vat unless the resolving vat is
```
warner: I can't parse that sentence. s/unless the/if the/ ?

Maybe "Furthermore, only pipelining-enabled vats are allowed to resolve promise IDs that they did not allocate." ?

packages/SwingSet/test/snapshots/test-xsnap-store.js.md
```diff
@@ -12,8 +12,8 @@ Generated by [AVA](https://avajs.dev).
 
 > after SES boot - sensitive to SES-shim, XS, and supervisor
 
-    'dd40da432d48e6e6b9530c8ae3712eb8d4e6c9f98b8e96c6d3638d1c89f2b8b9'
+    'e48339a392bbcf34fe316dc20fe8bb218f6ab39b9022e462c125ab1e12ebd18d'
```
warner, line 15: oh right, because the supervisor imports `message.js`, which imports `capdata.js`, which was changed

warner (approved):
One comment nit, rest looks good, land after fixing.

packages/SwingSet/docs/delivery.md
mhofman: Done.

mhofman:
Will squash now, then rebase once #5204 has landed.

Edit: [eeb79ec..e5f505b](https://github.com/Agoric/agoric-sdk/compare/eeb79ec14db5f6b1052ae40ed51d750e64d600ad..e5f505b965f3c835e11802f29a36f7dca998eaf7)

Issue #4856, opened 2022-03-16
Simplify AMM Swap API


dtribble:
## What is the Problem Being Solved?

Supporting both "swapIn" and "swapOut" in the contract API increases the complexity of the contract while narrowing its use cases. This is because the `swapOut` cases uses the `want` in the proposal as both the limit and as the sale target. Thus for example, clients cannot express the requirement from liquidation of "purchase _up to_ a limit but no more", because a want expresses "at least" not "at most". 

## Description of the Design

Replace `swapIn` and `swapOut` with a single `swap` operation that takes an additional `maxOut` offer argument that is *not* the `want` amount. The `maxOut` argument, if present, specifies that the most that the client needs back. 

* The `swapIn` operation would simply be a deprecated alias for `swap`. 
* The `swapOut` operation would invoke `swap` with `maxOut` as the `Out` amount from the `want`.

The pseudo code for the logic of the swap operation is (with conveniently sloppy fee compute):

```
swap({ give: { In } , want: { Out }, { maxOut = undefined })
  if In.brand === central
    getPool(Out.brand).giveCentral(In, maxOut)
  else if Out.brand === central
    getPool(In.brand).giveSecondary(In, maxOut)
  else 
    virtualPool(In.brand, Out.brand).swapSecondaries(In, maxOut)
    
pool.giveCentral(X, maxY = undefined) => Y:secondary
  fee = computeFee(X)
  afterFee = X - fee
  ySimple = increaseX(afterFee, xPool, yPool)
  Y = maxY && maxY < ySimple ? maxY : ySimple
  X = decreaseY(Y, xPool, yPool)
 
pool.giveSecondary(X, maxY = undefined) => Y:central
  yRaw = increaseX(X, xPool, yPool)
  ySimple = yRaw - computeFee(yRaw)
  Y = maxY && maxY < ySimple ? maxY : ySimple
  fee = computeFee(Y)
  X = decreaseY(Y, xPool, yPool)
 
virtualPool(xPool, yPool).swapSecondaries(X:secondary, maxY = undefined) => Y:secondary
  { Y: central } = xPool.giveSecondary(X)
  { X: maxCentral, Y, fee: fee2 } = yPool.giveCentral(central, maxY)
  { X, fee: fee1} = xPool.giveSecondary(X, maxCentral)
  fee = fee1 + fee2

increaseX: Return the Y to be returned for a contribution of 
decreaseY: Return the minimum X required to release Y
```

[...]

PR #5125, opened 2022-04-16
feat(amm): support `maxOut` optional arg to `swap`

dtribble:
refs: #4856

## Description

Implements the new API in terms of the existing underlying implementation.

* Move swap implementation into shared `swap.js`
* Implement `maxOut` using existing pricing operations
* implement `swapOut` in terms of `swapIn` with `maxOut`
* remove now-unneeded code in pool implementations
* make `allocateGainsAndLosses` API the same for single and double pools

### Security Considerations

N/A

Need to deprecate the old API and add the new argument to Swap.

### Testing Considerations

SwapOut is defined in terms of the new API, so it gets lot of testing now. Additional specific cases should be added for which the `want` is not the same as the `maxOut`.

[Test cases from the liquidation branch](https://github.com/Agoric/agoric-sdk/blob/c76e38190713ae908fdec6bbcc381c9c86fdf4fb/packages/run-protocol/test/vaultFactory/test-liquidator.js#L584)

### DIFF
--- a/packages/run-protocol/src/vaultFactory/liquidateMinimum.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateMinimum.js
@@ -6,7 +6,6 @@ import { AmountMath } from '@agoric/ertp';
 import { Far } from '@endo/marshal';
 
 import { defineKind } from '@agoric/vat-data';
-import { makeDefaultLiquidationStrategy } from './liquidation.js';
 import { makeTracer } from '../makeTracer.js';
 
 const trace = makeTracer('LiqMin');
@@ -34,59 +33,35 @@ const start = async zcf => {
     const {
       give: { In: amountIn },
     } = debtorSeat.getProposal();
-    const swapInvitation = E(amm).makeSwapOutInvitation();
+
+    const swapInvitation = E(amm).makeSwapInvitation();
     const liqProposal = harden({
       give: { In: amountIn },
-      want: { Out: debt },
+      want: { Out: AmountMath.makeEmpty(debtBrand) },
     });
     trace(`OFFER TO DEBT: `, debt, amountIn);
-    const { deposited, userSeatPromise: liqSeat } = await offerTo(
+    const { deposited } = await offerTo(
       zcf,
       swapInvitation,
       undefined, // The keywords were mapped already
       liqProposal,
       debtorSeat,
+      debtorSeat,
+      { maxOut: debt },
     );
 
-    // Three (!) awaits coming here. We can't use Promise.all because
+    // Multiple awaits coming here. We can't use Promise.all because
     // offerTo() can return before getOfferResult() is valid, and we can't
     // check whether swapIn liquidated assets until getOfferResult() returns.
     // Of course, we also can't exit the seat until one or the other
     // liquidation takes place.
     const amounts = await deposited;
-    await E(liqSeat).getOfferResult();
-    trace('exact sell result', amounts);
-
-    // if swapOut failed to make the trade, we'll sell it all
-    const sellAllIfUnsold = async () => {
-      const unsold = debtorSeat.getAmountAllocated('In');
-      // We cannot easily directly check that the `offerTo` succeeded, so we
-      // check that amotun unsold is not what we started with.
-      if (!AmountMath.isEqual(amountIn, unsold)) {
-        trace('Changed', { inBefore: amountIn, unsold });
-        return;
-      }
-
-      trace('liquidating all collateral because swapIn did not succeed');
-      const strategy = makeDefaultLiquidationStrategy(amm);
-      const { deposited: sellAllDeposited, userSeatPromise: sellAllSeat } =
-        await offerTo(
-          zcf,
-          strategy.makeInvitation(debt),
-          undefined, // The keywords were mapped already
-          strategy.makeProposal(amountIn, AmountMath.makeEmpty(debtBrand)),
-          debtorSeat,
-        );
-      // await sellAllDeposited, but don't need the value
-      await Promise.all([
-        E(sellAllSeat).getOfferResult(),
-        sellAllDeposited,
-      ]).catch(sellAllError => {
-        throw Error(`Unable to liquidate ${sellAllError}`);
-      });
-    };
-    await sellAllIfUnsold();
-    trace(`Liq results`, debt, debtorSeat.getAmountAllocated('RUN', debtBrand));
+    trace(
+      `Liq results`,
+      debt,
+      debtorSeat.getAmountAllocated('RUN', debtBrand),
+      amounts,
+    );
     debtorSeat.exit();
   };
 
--- a/packages/run-protocol/src/vaultFactory/liquidateMinimum.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateMinimum.js
@@ -49,19 +49,13 @@ const start = async zcf => {
       debtorSeat,
       { maxOut: debt },
     );
-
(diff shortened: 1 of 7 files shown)


packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -34,59 +33,35 @@ const start = async zcf => {
...
       swapInvitation,
       undefined, // The keywords were mapped already
       liqProposal,
       debtorSeat,
+      debtorSeat,
+      { maxOut: debt },
     );
 
-    // Three (!) awaits coming here. We can't use Promise.all because
+    // Multiple awaits coming here. We can't use Promise.all because
```
turadg: just one now, right?

packages/run-protocol/src/vpool-xyk-amm/swap.js
```diff
@@ -1,16 +1,24 @@
 // @ts-check
 
+// @ts-ignore
```
turadg: ignores solved in cad92bb15

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -193,6 +193,7 @@ const start = async (zcf, privateArgs) => {
     }
 
     const pool = isSecondary(brandOut) ? getPool(brandOut) : getPool(brandIn);
+    // @ts-expect-error cast
```
turadg: @Chris-Hibbert is it safe to cast to `DoublePoolInternalFacet` here?

turadg (commented):
I think the test failure is legit. Debt is empty in this block:
```
  // An emergency liquidation got less than full value
  const debtAfterLiquidation = await E(vault).getCurrentDebt();
```

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: Lines 184-193 are the doublePool case. This is handling a single pool, so the renaming of `VPoolInternalFacet` to `DoublePoolInternalFacet` looks like a mistake.  This correctly returns `VpoolWrapper`.

packages/run-protocol/src/vpool-xyk-amm/swap.js
```diff
@@ -19,23 +25,28 @@ export const makeMakeSwapInvitation = (zcf, provideVPool) => {
...
-    const pool = provideVPool(amountIn.brand, amountOut.brand).externalFacet;
-    return pool.swapIn(seat, amountIn, amountOut);
+    const pool = provideVPool(amountIn.brand, amountOut.brand).internalFacet;
+    let prices;
+    const maxOut = args.maxOut;
+    if (maxOut) {
+      AmountMath.coerce(amountOut.brand, maxOut);
+      prices = pool.getPriceForOutput(amountIn, maxOut);
+    }
+    if (!prices || !AmountMath.isGTE(prices.swapperGets, maxOut)) {
```
Chris-Hibbert: I think there are scenarios where current calls to `swapOut()` would fail to trade, and this code would fall through to `getPriceForInput()` and sell anyway. One I found is if there is insufficient liquidity to satisfy the `swapOut`. In that case, I think the failure to trade is the correct outcome, even though that's not what you want in the liquidation case.

I would also like to see reasonably thorough testing of the `swapIn(amountIn, { maxOut })` cases.

Chris-Hibbert (changes requested):

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: > renaming of `VPoolInternalFacet` to `DoublePoolInternalFacet` looks like a mistake

All properties of `VPoolInternalFacet` specified double,
```
/**
 * @typedef {Object} VPoolInternalFacet - virtual pool for price quotes and trading
 * @property {GetDoublePoolSwapQuote} getPriceForInput
 * @property {GetDoublePoolSwapQuote} getPriceForOutput
 * @property {(ZCFSeat, DoublePoolSwapResult) => string} allocateGainsAndLosses
 */
```

Is there a type for `VPoolInternalFacet` that doesn't presume pool type?

dtribble: As far as I can tell, no. So the types are not generic here, even though the code kinda is.

packages/run-protocol/src/vpool-xyk-amm/swap.js
dtribble: Since it also passes the `want.Out` as a `want.Out` for the `swapIn`, I believe in _all_ those cases it should properly fail to reallocate because the want is not satisfied.

dtribble:
[Test cases from the liquidation branch](https://github.com/Agoric/agoric-sdk/blob/c76e38190713ae908fdec6bbcc381c9c86fdf4fb/packages/run-protocol/test/vaultFactory/test-liquidator.js#L584)

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
dtribble: fixed

Chris-Hibbert:
switching between `getPriceForInput` and `getPriceForOutput` when the user calls either `swapIn` or `swapOut` breaks the assumptions [laid out in the doc](https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/vpool-xyk-amm/constantProduct/README.md).

The first consequence I can see is that the way fees are calculated is supposed to depend on whether they called `swapIn` or `swapOut` ("The pool fee is charged against the side not specified by the user" assumes that `swapIn` specifies the input price as base, and `swapOut` specifies the output price as the base.)  In some cases this means the pool fee will be charged in the opposite currency.  

That's probably defensible, but we haven't done so yet.

I'm not sure those are the only consequences. I am convinced that offer safety constraints are not broken by this approach. When the user specifies `amountOut`, there won't be a transaction unless they get at least that much. If they specify `stopAfter` instead, they don't get offer safety.

You suggested that we could change to an approach where we charge a flat fee on top of the transaction and allocate it to pool and protocol fees separately from calculating prices. That would also be defensible, but that's not what this change does.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
dtribble: It looks like someone made the type a union type, so it's now actually accurate.

packages/run-protocol/src/vpool-xyk-amm/swap.js
```diff
@@ -3,14 +3,20 @@
 import { assertProposalShape } from '@agoric/zoe/src/contractSupport/index.js';
 
 import '@agoric/zoe/exported.js';
+import { AmountMath } from '@agoric/ertp';
 
 /**
  * @param {ZCF} zcf
- * @param {(brandIn: Brand, brandOut: Brand) => VPoolWrapper<unknown>} provideVPool
+ * @param {(brandIn: Brand, brandOut: Brand) => VPoolWrapper<DoublePoolInternalFacet>} provideVPool
```
Chris-Hibbert, line 10: provideVPool can provide either singlePool or doublePool, though no one should care. I'd make it `VPoolWrapper<unknown>`, which allows future extensions and declares that we don't care.

Chris-Hibbert (approved):
Thanks for listening to my concerns.

This is good.

PR #5129, opened 2022-04-17
feat(vault): Liquidate incrementally

dtribble:
refs: #4568

## Description

Liquidate in tranches the limit the price impact within a given block.

### Security Considerations

Doesn't have access to the mint, and so only gets assets granted to it. Thus minimal risk, except failure would impair liquidation. That's mitigated by having the liquidation contract be replacable.

Specification is needed for security reviewers

### Testing Considerations

Testing multi-block liquidation is challenging.

### DIFF
--- a/packages/run-protocol/src/collect.js
+++ b/packages/run-protocol/src/collect.js
@@ -5,7 +5,7 @@ const { entries, fromEntries, keys, values } = Object;
 /** @type { <K extends string, T, U>(obj: Record<K, T>, f: (t: T) => U) => Record<K, U>} */
 export const mapValues = (obj, f) =>
   // @ts-ignore entries() loses the K type
-  harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v)])));
+  harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v, p)])));
 
 /** @type { <X, Y>(xs: X[], ys: Y[]) => [X, Y][]} */
 export const zip = (xs, ys) => harden(xs.map((x, i) => [x, ys[+i]]));
--- a/packages/run-protocol/src/collect.js
+++ b/packages/run-protocol/src/collect.js
@@ -2,7 +2,7 @@
 
 const { entries, fromEntries, keys, values } = Object;
 
-/** @type { <K extends string, T, U>(obj: Record<K, T>, f: (t: T) => U) => Record<K, U>} */
+/** @type { <K extends string, T, U>(obj: Record<K, T>, f: (t: T, k: K) => U) => Record<K, U>} */
 export const mapValues = (obj, f) =>
   // @ts-ignore entries() loses the K type
   harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v, p)])));
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -55,7 +55,7 @@ const start = async zcf => {
     AMMMaxSlippage,
   } = /** @type {LiquidationContractTerms} */ zcf.getTerms();
 
-  trace('terms', zcf.getTerms());
+  // trace('terms', zcf.getTerms());
   const nextBlock = async () => E(timerService).delay(1n);
 
   const SCALE = 1_000_000n;
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -229,7 +229,7 @@ const start = async zcf => {
       proposal,
       debtorSeat,
       debtorSeat,
-      { maxOut: debt },
+      { stopAfter: debt },
     );
     await Promise.all([E(liqSeat).getOfferResult(), deposited]);
     const amounts = await E(liqSeat).getCurrentAllocation();
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -4,7 +4,6 @@
 import { E } from '@endo/eventual-send';
 import {
   getAmountOut,
-  makeRatio,
   assertProposalShape,
   offerTo,
   natSafeMath as NatMath,
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -1,0 +1,322 @@
+/* eslint-disable no-await-in-loop */
+// no @ts-check
+
+import { E } from '@endo/eventual-send';
+import {
+  getAmountOut,
+  makeRatio,
+  assertProposalShape,
+  offerTo,
+  natSafeMath as NatMath,
+} from '@agoric/zoe/src/contractSupport/index.js';
+import { AmountMath } from '@agoric/ertp';
+import { Far } from '@endo/marshal';
+
+import { ceilMultiplyBy } from '@agoric/zoe/src/contractSupport/ratio';
+import { makeTracer } from '../makeTracer.js';
+// import { quote } from '@agoric/assert';
+
+const { details: X } = assert;
+const trace = makeTracer('LiqI');
+
+/**
+ * This contract liquidates a payment in multiple tranches to limit the
+ * price impact an the AMM of any one sale. Each block it will compute
+ * a tranche of collateral to sell, where the size is a function of
+ * the amount of that collateral in the AMM pool and the desired price impact.
+ * It then gets 3 prices for the current tranche:
+ * - AMM quote - compute xYK locally based on the pool sizes
+ * - Reserve quote - based on a low price at which the Reserve will purchase
+ * - Oracle quote - provide by the oracle that triggered liquidation
+ * - Oracle limit - a limit that is DIVERGENCE_TOLERANCE below the
+ *   Oracle quote.
+ * These sell based on the following:
+ * - If the AMM quote is *below* the oracle limit, skip this block
+ *   to wait for oracle and AMM prices to converge.
+ *    - should it still sell to teh Reserve here?
+ * - If the Reserve quote is higher than the AMM quote, sell to the
+ *   Reserve with a `want` of the Oracle limit
+ * - Otherwise, sell to the AMM with a `want` of the Oracle limit
+ * If selling a tranche for the minimum would pay off the debt (`isLast`),
+ * then use sell using`swapOut` with a minimum of the debt. If that fails
+ * this is not the last tranche, tehn sell with the `swapIn` operation,
+ * using the oracle price reference as the minimum.
+ *
+ * TODO integrate the reserve
+ */
+
+/** @type {ContractStartFn} */
+const start = async zcf => {
+  const { amm, priceAuthority, timerService, debtBrand } =
+    /** @type {LiquidationContractTerms} */ zcf.getTerms();
+  const nextBlock = async () => E(timerService).delay(1n);
+
+  // TODO make subject to governance
+  const SCALE = 1_000_000n;
+  const BASIS_POINTS = 10_000n * SCALE;
+  const BP2 = BASIS_POINTS * BASIS_POINTS;
+
+  const AMM_FEE_BP = 30n * SCALE;
+  const MAX_IMPACT_BP = 50n * SCALE;
+  // sqrt(impact + 1) - 1
+  // Since we take SQRT but want to be BP, multiply in BP
+  const IMPACT_FACTOR_BP =
+    BigInt(
+      Math.ceil(
+        Math.sqrt(Number((MAX_IMPACT_BP + BASIS_POINTS) * BASIS_POINTS)),
+      ),
+    ) - BASIS_POINTS;
+  const IMPACT_FACTOR_BP2 = IMPACT_FACTOR_BP * (BASIS_POINTS - AMM_FEE_BP);
+
+  const asFloat = (numerator, denominator) =>
+    Number(numerator) / Number(denominator);
+  trace(
+    'FACTORS',
+    asFloat(AMM_FEE_BP, BASIS_POINTS),
+    asFloat(MAX_IMPACT_BP, BASIS_POINTS),
+    Math.sqrt(Number(MAX_IMPACT_BP + BASIS_POINTS)),
+    asFloat(IMPACT_FACTOR_BP, BASIS_POINTS),
+    asFloat(IMPACT_FACTOR_BP2, BP2),
+  );
+
+  // How close the sale price on the AMM must be to the expected price from the oracle
+  // quote must be at least 70% 0f what the oracle says
+  const ORACLE_TOLERANCE = 3000n * SCALE;
+  const ORACLE_LIMIT_BP = makeRatio(
+    BASIS_POINTS - ORACLE_TOLERANCE,
+    debtBrand,
+    BASIS_POINTS,
+  );
+
+  // 3% max slippage = 97% min
+  const MAX_SLIPPED = makeRatio(9700n * SCALE, debtBrand, BASIS_POINTS);
+
+  trace('CONSTANTS', {
+    IMPACT_FACTOR_BP,
+    BASIS_POINTS,
+    IMPACT_FACTOR_BP2,
+    BP2,
+  });
+
+  // the maximum number fo failure sina  row before we stop liquidating.
+  // TODO make this larger andmake it reset on success
+  const MAX_FAILURES = 4;
+  let failures = 0;
+
+  const computeAMMProceeds = (poolCentral, poolCollateral, tranche, debt) => {
+    const k = NatMath.multiply(poolCentral.value, poolCollateral.value);
+    const postSaleCollateral = AmountMath.add(tranche, poolCollateral);
+    const quoteValue = NatMath.subtract(
+      poolCentral.value,
+      NatMath.floorDivide(k, postSaleCollateral.value),
+    );
+    const ammQuote = AmountMath.make(debt.brand, quoteValue);
+    const minAmmProceeds = ceilMultiplyBy(ammQuote, MAX_SLIPPED);
+    trace('AMM estimate', {
+      tranche,
+      minAmmProceeds,
+      poolCentral,
+      poolCollateral,
+    });
+    return minAmmProceeds;
+  };
+
+  /**
+   * Compute the tranche size whose sale on the AMM would have
+   * a price impact of MAX_IMPACT_BP.
+   * This doesn't use ratios so that it is usable for any brand
+   *
+   * @param {Amount} poolSize
+   * @returns {Amount}
+   */
+  const maxTrancheWithFees = poolSize =>
+    AmountMath.make(
+      poolSize.brand,
+      NatMath.ceilDivide(
+        NatMath.multiply(poolSize.value, IMPACT_FACTOR_BP2),
+        BP2,
+      ),
+    );
+
+  /**
+   *
+   * @param {ZCFSeat} seat
+   * @param {Amount} originalDebt
+   * @yields {LiquidationStep}
+   */
+  async function* processTranches(seat, originalDebt) {
+    let proceedsSoFar = seat.getAmountAllocated('Out');
+    let toSell = seat.getAmountAllocated('In');
+    while (
+      !AmountMath.isGTE(proceedsSoFar, originalDebt) &&
+      !AmountMath.isEmpty(toSell)
+    ) {
+      const debt = AmountMath.subtract(originalDebt, proceedsSoFar);
+      const { Secondary: poolCollateral, Central: poolCentral } = await E(
+        amm,
+      ).getPoolAllocation(toSell.brand);
+      const maxAllowedTranche = maxTrancheWithFees(poolCollateral);
+      trace('TRANCHE', asFloat(maxAllowedTranche.value, 100000n));
+      const tranche = AmountMath.min(maxAllowedTranche, toSell);
+      // compute the expected proceeds from the AMM for tranche
+      const minAmmProceeds = computeAMMProceeds(
+        poolCentral,
+        poolCollateral,
+        tranche,
+        debt,
+      );
+      // true if the minimum proceeds will pay off the debt
+      const isLast = AmountMath.isGTE(minAmmProceeds, debt);
+
+      // this could use a unit rather than the tranche size to run concurrently
+      /** @type PriceQuote */
+      const oracleQuote = await E(priceAuthority).quoteGiven(
+        tranche,
+        debtBrand,
+      );
+      // TODO Make a governance parameter
+      const oracleLimit = ceilMultiplyBy(
+        getAmountOut(oracleQuote),
+        ORACLE_LIMIT_BP,
+      );
+
+      trace('TRANCHE', {
+        debt,
+        tranche,
+        minAmmProceeds,
+        oracleLimit,
+        isLast,
+        poolCentral,
+        poolCollateral,
+      });
+      if (AmountMath.isGTE(minAmmProceeds, oracleLimit)) {
+        yield {
+          collateral: tranche,
+          ammProceeds: minAmmProceeds,
+          debt,
+          oracleLimit,
+          isLast,
+        };
+      } else {
+        trace('SKIP');
+      }
+      proceedsSoFar = seat.getAmountAllocated('Out');
+      toSell = seat.getAmountAllocated('In');
+      if (
+        AmountMath.isGTE(proceedsSoFar, originalDebt) ||
+        AmountMath.isEmpty(toSell)
+      ) {
+        trace('exiting async loop');
+        return;
+      }
+      await nextBlock();
+    }
+  }
+
+  async function sellTranche(debtorSeat, collateral, debt, oracleLimit) {
+    const swapInvitation = E(amm).makeSwapInvitation();
+    const want = AmountMath.min(debt, oracleLimit);
+    // TODO if the debt shouldn't require all the collatearal to cover it,
+    // this may not provide relevant offer safety
+    const proposal = {
+      give: { In: collateral },
+      want: { Out: want },
+    };
+    const { deposited, userSeatPromise: liqSeat } = await offerTo(
+      zcf,
+      swapInvitation,
+      undefined, // The keywords were mapped already
+      proposal,
+      debtorSeat,
+      debtorSeat,
+      { maxOut: debt },
+    );
+    await Promise.all([E(liqSeat).getOfferResult(), deposited]);
+    const amounts = await E(liqSeat).getCurrentAllocation();
+    trace('offerResult', { amounts });
+  }
+
+  const debtorHook = async (debtorSeat, { debt: originalDebt }) => {
+    trace('LIQ', originalDebt);
+    assertProposalShape(debtorSeat, {
+      give: { In: null },
+    });
+    assert(
+      originalDebt.brand === debtBrand,
+      X`Cannot liquidate to ${originalDebt.brand}`,
+    );
+
+    // Just do the incremental liquidation for the debtorSeat
+    // Three outcomes of total liquidation:
+    // 1. the assets sell for more than the debt - covered
+    // 2. the assets sell for less than the debt - shortfall
+    // 3. we stop selling assets so we have both assets and remaining debt (reserve purchase)
+
+    for await (const t of processTranches(debtorSeat, originalDebt)) {
+      const { collateral, oracleLimit, ammProceeds, debt, isLast } = t;
+      trace(`OFFER TO DEBT: `, {
+        collateral,
+        ammProceeds,
+        oracleLimit,
+        debt,
+        isLast,
+      });
+      const collateralBefore = debtorSeat.getAmountAllocated('In');
+      const proceedsBefore = debtorSeat.getAmountAllocated('Out');
+
+      await sellTranche(debtorSeat, collateral, debt, oracleLimit);
+
+      const proceedsAfter = debtorSeat.getAmountAllocated('Out');
+      if (AmountMath.isEqual(proceedsBefore, proceedsAfter)) {
+        const collateralAfter = debtorSeat.getAmountAllocated('In');
+        trace('NOSELL', {
+          proceedsBefore,
+          proceedsAfter,
+          collateralBefore,
+          collateralAfter,
+        });
+        // nothing got sold
+        failures += 1;
+        if (failures > MAX_FAILURES) {
+          throw Error(`Too many sequential failures ${failures}`);
+        }
+      }
+    }
+    debtorSeat.exit();
+    trace('exit seat');
+  };
+
+  const creatorFacet = Far('debtorInvitationCreator', {
+    makeDebtorInvitation: () => zcf.makeInvitation(debtorHook, 'Liquidate'),
+  });
+
+  return harden({ creatorFacet });
+};
+
+/** @type {MakeLiquidationStrategy} */
+const makeLiquidationStrategy = creatorFacet => {
+  const makeInvitation = () => E(creatorFacet).makeDebtorInvitation();
+
+  const keywordMapping = () =>
+    harden({
+      Collateral: 'In',
+      RUN: 'Out',
+    });
+
+  const makeProposal = (collateral, run) =>
+    harden({
+      give: { In: collateral },
+      want: { Out: AmountMath.makeEmptyFromAmount(run) },
+    });
+
+  return {
+    makeInvitation,
+    keywordMapping,
+    makeProposal,
+  };
+};
+
+harden(start);
+harden(makeLiquidationStrategy);
+
+export { start, makeLiquidationStrategy };
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-await-in-loop */
-// no @ts-check
+// @ts-check
 
 import { E } from '@endo/eventual-send';
 import {
@@ -12,7 +12,7 @@ import {
 import { AmountMath } from '@agoric/ertp';
 import { Far } from '@endo/marshal';
 
-import { ceilMultiplyBy } from '@agoric/zoe/src/contractSupport/ratio';
+import { ceilMultiplyBy } from '@agoric/zoe/src/contractSupport/ratio.js';
 import { makeTracer } from '../makeTracer.js';
 // import { quote } from '@agoric/assert';
 
@@ -20,29 +20,29 @@ const { details: X } = assert;
 const trace = makeTracer('LiqI');
 
 /**
+ * @file
  * This contract liquidates a payment in multiple tranches to limit the
- * price impact an the AMM of any one sale. Each block it will compute
+ * price impact on the AMM of any one sale. Each block it will compute
  * a tranche of collateral to sell, where the size is a function of
  * the amount of that collateral in the AMM pool and the desired price impact.
  * It then gets 3 prices for the current tranche:
- * - AMM quote - compute xYK locally based on the pool sizes
+ * - AMM quote - compute XYK locally based on the pool sizes
  * - Reserve quote - based on a low price at which the Reserve will purchase
- * - Oracle quote - provide by the oracle that triggered liquidation
+ * - Oracle quote - provided by the oracle that triggered liquidation
  * - Oracle limit - a limit that is DIVERGENCE_TOLERANCE below the
  *   Oracle quote.
- * These sell based on the following:
+ * Then sell based on the following:
  * - If the AMM quote is *below* the oracle limit, skip this block
  *   to wait for oracle and AMM prices to converge.
- *    - should it still sell to teh Reserve here?
  * - If the Reserve quote is higher than the AMM quote, sell to the
  *   Reserve with a `want` of the Oracle limit
  * - Otherwise, sell to the AMM with a `want` of the Oracle limit
- * If selling a tranche for the minimum would pay off the debt (`isLast`),
- * then use sell using`swapOut` with a minimum of the debt. If that fails
- * this is not the last tranche, tehn sell with the `swapIn` operation,
- * using the oracle price reference as the minimum.
  *
- * TODO integrate the reserve
+ * Selling uses the `oracleLimit as the `want` as the limit to allowed
+ * slippage, and provides the remaining `debt` as the `stopAfter` so
+ * that we sell no more than is needed to pay off the debt.
+ *
+ * TODO integrate the reserve, including the above Reserve strategies.
  */
 
 /** @type {ContractStartFn} */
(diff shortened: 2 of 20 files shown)


packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -26,7 +26,7 @@ import { chargeInterest } from '../interest.js';
 
 const { details: X, quote: q } = assert;
 
-const trace = makeTracer('VM', false);
+const trace = makeTracer('VM', true);
```
Chris-Hibbert, line 30: back to `false` before merge?

And does it make sense to drop some of the added `traces` now?

packages/run-protocol/src/vpool-xyk-amm/swap.js
```diff
@@ -19,23 +27,36 @@ export const makeMakeSwapInvitation = (zcf, provideVPool) => {
...
   const swapOut = seat => {
-    assertProposalShape(seat, {
-      give: { In: null },
-      want: { Out: null },
-    });
     // The offer's amountOut is a minimum; the offeredAmountIn is a max.
     const {
-      give: { In: amountIn },
-      want: { Out: amountOut },
+      want: { Out: maxOut },
```
Chris-Hibbert: The code the comment used to refer to is changed. Please update comment.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1614,21 +1603,23 @@ test('mutable liquidity triggers and interest', async t => {
...
+      aliceLoanAmount,
     ),
   );
 
   let aliceUpdate = await E(aliceNotifier).getUpdateSince();
   t.deepEqual(aliceUpdate.value.debtSnapshot.debt, aliceRunDebtLevel);
 
-  // Create a loan for Bob for 740 RUN with 100 Aeth collateral
+  // Create a loan for Bob for 650 RUN with 100 Aeth collateral
+  // 7.4
```
Chris-Hibbert: 7.4 is no longer correct

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1695,12 +1686,22 @@ test('mutable liquidity triggers and interest', async t => {
 
   aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
   t.deepEqual(aliceUpdate.value.debtSnapshot.debt, aliceRunDebtLevel);
+  trace(t, 'alice reduce collateral');
 
-  await manualTimer.tick();
+  await E(priceAuthority).setPrice(makeRatio(7n, runBrand, 1n, aethBrand));
+  trace(t, 'changed price to 7');
+
+  // await manualTimer.tick();
```
Chris-Hibbert: drop  // tick

packages/run-protocol/src/vpool-xyk-amm/swap.js
```diff
@@ -19,23 +27,36 @@ export const makeMakeSwapInvitation = (zcf, provideVPool) => {
...
-    const pool = provideVPool(amountIn.brand, amountOut.brand).externalFacet;
-    return pool.swapIn(seat, amountIn, amountOut);
+    if (args) {
+      AmountMath.coerce(amountOut.brand, args.maxOut);
+      // TODO check that there are no other keys
+    }
+
+    const pool = provideVPool(amountIn.brand, amountOut.brand).internalFacet;
+    let prices;
+    const maxOut = args && args.maxOut;
```
Chris-Hibbert: turadg and I have started using
```suggestion
    const maxOut = args?.maxOut;
```

packages/run-protocol/src/collect.js
```diff
@@ -5,7 +5,7 @@ const { entries, fromEntries, keys, values } = Object;
 /** @type { <K extends string, T, U>(obj: Record<K, T>, f: (t: T) => U) => Record<K, U>} */
 export const mapValues = (obj, f) =>
   // @ts-ignore entries() loses the K type
-  harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v)])));
+  harden(fromEntries(entries(obj).map(([p, v]) => [p, f(v, p)])));
```
turadg, line 8: this change doesn't match the type signature and doesn't appear to be used.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
```diff
@@ -106,15 +106,7 @@ export const makePrioritizedVaults = (reschedulePriceCheck = () => {}) => {
    */
   const removeVault = key => {
     const vault = vaults.removeByKey(key);
-    const debtToCollateral = currentDebtToCollateral(vault);
-    if (
-      !oracleQueryThreshold ||
-      ratioGTE(debtToCollateral, oracleQueryThreshold)
-    ) {
-      // don't call reschedulePriceCheck, but do reset the highest.
-      // This could be expensive if we delete individual entries in order. Will know once we have perf data.
```
turadg, line 115: not sure if:
1. you're now confident this will be a perf issue
2. looking at perf of this will be tracked in some other way
3. you don't think it needs noting because if there are perf problems with this they'll be easily diagnosed
4. the comment was accidentally removed and will be replaced

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -325,20 +325,20 @@ const helperBehavior = {
...
     }
     const { phase } = state;
-    const uiState = facets.helper.snapshotState(phase);
+    const uiState = facets.helper.getStateSnapshot(phase);
     trace('updateUiState', state.idInManager, uiState);
 
     switch (phase) {
       case VaultPhase.ACTIVE:
       case VaultPhase.LIQUIDATING:
+      case VaultPhase.LIQUIDATED:
```
turadg, line 338: 👍

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg, line 30: do you want this enabled in master?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -294,11 +294,15 @@ const helperBehavior = {
     // callback that may not fire until much later.
     // Callers shouldn't expect a response from this function.
     const { priceAuthority } = state;
-    state.outstandingQuote = await E(priceAuthority).mutableQuoteWhenLT(
+    trace('posting quote request', triggerPoint);
+
+    state.outstandingQuote = E(priceAuthority).mutableQuoteWhenLT(
```
turadg, line 294: the comment above is no longer accurate. maybe:
> There are two promises here to get the quote. The first gets a mutableQuote object from the PriceAuthority but it's not awaited. …

Alternately, this would be the second time a change in an important design comment was hidden from code review diff. I think that's because it's not obvious from the code flow that there's some block that's being documented. Consider encapsulating this logic into a helper function like:
```
const quote = await updatedQuote(priceAuthority, highestDebtRatio.denominator, triggerPoint);
```
And the jsdoc on that function can explain the design.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -316,7 +321,9 @@ const helperBehavior = {
       prioritizedVaults.entriesPrioritizedGTE(quoteRatioPlusMargin);
     await (next ? facets.helper.liquidateAndRemove(next) : null);
 
-    facets.helper.reschedulePriceCheck();
+    trace('price check liq', next && next[0]);
```
turadg, line 319:
```suggestion
    trace('price check liq', next?.[0]);
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -316,7 +321,9 @@ const helperBehavior = {
...
+
+    await facets.helper.reschedulePriceCheck();
```
turadg: would this get more execution into the same turn?
```suggestion
    return facets.helper.reschedulePriceCheck();
```

packages/run-protocol/src/vpool-xyk-amm/swap.js
```diff
@@ -1,8 +1,10 @@
 // @ts-check
 
+// @ts-ignore
```
turadg: Looks like changes to this file crept in through rebasing. Please remove.

packages/run-protocol/test/bundleTool.js
```diff
@@ -90,7 +90,7 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
     return meta;
   };
 
-  const validate = async targetName => {
+  const validate = async (targetName, rootOpt) => {
```
turadg: prefer that bundleTool changes get into master outside this PR. wouldn't be so much an issue if this will have a merge commit but looking at the commit history (e.g. "[chore(vault): fix ups after rebase](https://github.com/Agoric/agoric-sdk/pull/5129/commits/b586fe8c13d9d180ddcc436116f46dcabe6b351b)" i think a squash will be better. So request is either:
1. plan to squash, move out the bundleTool change, and update the PR title to be a conventional commit
2. plan to merge and clean up the commit list

packages/run-protocol/test/vaultFactory/test-liquidation-helpers.js
```diff
@@ -10,29 +10,11 @@ export const mockBrand = Far('brand');
 
 const amount = n => AmountMath.make(mockBrand, BigInt(n));
 
-for (const [
```
turadg, line 13: 👍

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -0,0 +1,642 @@
...
+  };
+}
+// #endregion
+
+// sell in a single tranche
+//   - no shortfall
+//   - with shortfall
+// sell in multiple tranches
+// stall then oracle correct
+// stall then trade the amm back to the price
```
turadg: notes to self?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -0,0 +1,642 @@
...
+
+// #region driver
+const makeDriver = async (t, initialPrice, priceBase) => {
```
turadg, line 315: I like this driver thing a lot. Might be worth pulling out of the `test-` file for easier re-use and to make it easier to work with the test specs.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -272,36 +265,30 @@ async function setupServices(
...
     ? makeScriptedPriceAuthority({
-        actualBrandIn: aethBrand,
-        actualBrandOut: runBrand,
-        priceList: priceOrList,
-        timer,
-        quoteMint,
-        unitAmountIn,
-        quoteInterval,
-      })
+      actualBrandIn: aethBrand,
```
turadg: needs fixing. you can avoid CI errors of formatting with [format-on-save](https://github.com/Agoric/agoric-sdk/blob/b586fe8c13d9d180ddcc436116f46dcabe6b351b/scripts/configure-vscode.sh#L22-L23)

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: what's 7.4? looks stale

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1614,21 +1603,23 @@ test('mutable liquidity triggers and interest', async t => {
...
   const aliceLoanProceeds = await E(aliceLoanSeat).getPayouts();
   t.deepEqual(aliceLentAmount, aliceLoanAmount, 'received 5000 RUN');
+  trace(t, 'alice vault');
 
   const aliceRunLent = await aliceLoanProceeds.RUN;
   t.truthy(
     AmountMath.isEqual(
       await E(runIssuer).getAmountOf(aliceRunLent),
-      AmountMath.make(runBrand, 5000n),
+      aliceLoanAmount,
```
turadg, line 1610: 👍

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1695,12 +1686,22 @@ test('mutable liquidity triggers and interest', async t => {
...
+
+  // await manualTimer.tick();
   // price levels changed and interest was charged.
 
   // expect Alice to be liquidated because her collateral is too low.
   aliceUpdate = await E(aliceNotifier).getUpdateSince(aliceUpdate.updateCount);
+  trace(t, 'alice liquidating?', aliceUpdate.value.vaultState);
+  t.is(aliceUpdate.value.vaultState, Phase.LIQUIDATING);
+
+  bobUpdate = await E(bobNotifier).getUpdateSince();
```
turadg, line 1700: safer way to iterate updates:
```suggestion
  bobUpdate = await E(bobNotifier).getUpdateSince(bobUpdate.count);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1709,18 +1710,44 @@ test('mutable liquidity triggers and interest', async t => {
...
-  bobUpdate = await E(bobNotifier).getUpdateSince(bobUpdate.updateCount);
   t.is(aliceUpdate.value.vaultState, Phase.LIQUIDATED);
+  trace(t, 'alice liquidated');
 
-  for (let i = 0; i < 5; i += 1) {
-    manualTimer.tick();
-  }
-  await waitForPromisesToSettle();
-  bobUpdate = await E(bobNotifier).getUpdateSince(bobUpdate.updateCount);
+  bobUpdate = await E(bobNotifier).getUpdateSince();
```
turadg, line 1723:
```suggestion
  bobUpdate = await E(bobNotifier).getUpdateSince(bobUpdate.count);
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1709,18 +1710,44 @@ test('mutable liquidity triggers and interest', async t => {
...
+  trace(
+    t,
+    'bob state?',
+    bobUpdate.value.vaultState,
+    await E(bobVault).getCurrentDebt(),
+  );
+  // 5 days pass
```
turadg, line 1730: thinking aloud: a test util for this would be handy

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1709,18 +1710,44 @@ test('mutable liquidity triggers and interest', async t => {
...
+  manualTimer.tick();
+  manualTimer.tick();
+  manualTimer.tick();
+  manualTimer.tick();
+  await manualTimer.tick();
 
+  bobUpdate = await E(bobNotifier).getUpdateSince();
```
turadg, line 1738: updateCount

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1709,18 +1710,44 @@ test('mutable liquidity triggers and interest', async t => {
...
+  trace(
+    t,
+    'bob 2 state?',
+    bobUpdate.value.vaultState,
+    await E(bobVault).getCurrentDebt(),
+  );
+
+  await waitForPromisesToSettle();
+  bobUpdate = await E(bobNotifier).getUpdateSince();
```
turadg, line 1747: I suspect that if you want for a fresh update you won't have to mess with settling promises.
```suggestion
  bobUpdate = await E(bobNotifier).getUpdateSince(bobUpdate.updateCount);
```

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+import { ceilMultiplyBy } from '@agoric/zoe/src/contractSupport/ratio';
+import { makeTracer } from '../makeTracer.js';
+// import { quote } from '@agoric/assert';
+
+const { details: X } = assert;
+const trace = makeTracer('LiqI');
+
+/**
+ * This contract liquidates a payment in multiple tranches to limit the
+ * price impact an the AMM of any one sale. Each block it will compute
```
turadg:
```suggestion
 * price impact on the AMM of any one sale. Each block it will compute
```

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+ * a tranche of collateral to sell, where the size is a function of
+ * the amount of that collateral in the AMM pool and the desired price impact.
+ * It then gets 3 prices for the current tranche:
+ * - AMM quote - compute xYK locally based on the pool sizes
+ * - Reserve quote - based on a low price at which the Reserve will purchase
+ * - Oracle quote - provide by the oracle that triggered liquidation
```
turadg:
```suggestion
 * - Oracle quote - provided by the oracle that triggered liquidation
```

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+import { Far } from '@endo/marshal';
+
+import { ceilMultiplyBy } from '@agoric/zoe/src/contractSupport/ratio';
+import { makeTracer } from '../makeTracer.js';
+// import { quote } from '@agoric/assert';
+
+const { details: X } = assert;
+const trace = makeTracer('LiqI');
+
+/**
```
turadg, line 20: consider putting this at the top as a `/** @file` comment.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+ * It then gets 3 prices for the current tranche:
+ * - AMM quote - compute xYK locally based on the pool sizes
+ * - Reserve quote - based on a low price at which the Reserve will purchase
+ * - Oracle quote - provide by the oracle that triggered liquidation
+ * - Oracle limit - a limit that is DIVERGENCE_TOLERANCE below the
+ *   Oracle quote.
+ * These sell based on the following:
+ * - If the AMM quote is *below* the oracle limit, skip this block
+ *   to wait for oracle and AMM prices to converge.
+ *    - should it still sell to teh Reserve here?
```
turadg: likely to be missed later. better to put with the `TODO integrate the reserve`

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+ * - If the Reserve quote is higher than the AMM quote, sell to the
+ *   Reserve with a `want` of the Oracle limit
+ * - Otherwise, sell to the AMM with a `want` of the Oracle limit
+ * If selling a tranche for the minimum would pay off the debt (`isLast`),
+ * then use sell using`swapOut` with a minimum of the debt. If that fails
+ * this is not the last tranche, tehn sell with the `swapIn` operation,
```
turadg:
```suggestion
 * this is not the last tranche, then sell with the `swapIn` operation,
```

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+  const MAX_IMPACT_BP = 50n * SCALE;
+  // sqrt(impact + 1) - 1
+  // Since we take SQRT but want to be BP, multiply in BP
+  const IMPACT_FACTOR_BP =
+    BigInt(
+      Math.ceil(
+        Math.sqrt(Number((MAX_IMPACT_BP + BASIS_POINTS) * BASIS_POINTS)),
+      ),
+    ) - BASIS_POINTS;
+  const IMPACT_FACTOR_BP2 = IMPACT_FACTOR_BP * (BASIS_POINTS - AMM_FEE_BP);
```
turadg: since these will be subject to governance, they should be coded as functions of the values that will be governable. (we can't make `SCALE` subject to governance yet but we know that it won't be constant)

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+  const MAX_SLIPPED = makeRatio(9700n * SCALE, debtBrand, BASIS_POINTS);
+
+  trace('CONSTANTS', {
+    IMPACT_FACTOR_BP,
+    BASIS_POINTS,
+    IMPACT_FACTOR_BP2,
+    BP2,
+  });
+
+  // the maximum number fo failure sina  row before we stop liquidating.
```
turadg:
```suggestion
  // the maximum number of failures in a row before we stop liquidating.
```

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+  // TODO make this larger andmake it reset on success
```
turadg: TODO when? seems important

also typos

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+      ),
+    );
+
+  /**
+   *
+   * @param {ZCFSeat} seat
+   * @param {Amount} originalDebt
+   * @yields {LiquidationStep}
+   */
+  async function* processTranches(seat, originalDebt) {
```
turadg, line 141: not a blocker but generators won't work in Jessie or TinySES.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+      // true if the minimum proceeds will pay off the debt
+      const isLast = AmountMath.isGTE(minAmmProceeds, debt);
+
+      // this could use a unit rather than the tranche size to run concurrently
+      /** @type PriceQuote */
+      const oracleQuote = await E(priceAuthority).quoteGiven(
+        tranche,
+        debtBrand,
+      );
+      // TODO Make a governance parameter
```
turadg: when?

turadg (commented):
I took a pass at code but haven't inspected the economics logic.

Chris-Hibbert (commented):
It looks like I have some leftover comments I never sent from an earlier review. I'll send these, and then take another look.

packages/run-protocol/src/collect.js
dtribble: It's used in tests. It matches the typical iteration apis where the second arg is the key if you need it. Any callee that doesn't care just ignores it.

dtribble: Is this the right type: 
```
/** @type { <K extends string, T, U>(obj: Record<K, T>, f: (t: T, k: K) => U) => Record<K, U>} */
```

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
dtribble: #4 when I integrated some set of fixes here.

I think the right implementation is to remember the key of the highest and only update then. I want that to be a separate checkin.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: nope.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: that woudl have the same bug as the `await` bug this is fixing.  I'll add comments here.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+ * - If the AMM quote is *below* the oracle limit, skip this block
+ *   to wait for oracle and AMM prices to converge.
+ *    - should it still sell to teh Reserve here?
+ * - If the Reserve quote is higher than the AMM quote, sell to the
+ *   Reserve with a `want` of the Oracle limit
+ * - Otherwise, sell to the AMM with a `want` of the Oracle limit
+ * If selling a tranche for the minimum would pay off the debt (`isLast`),
+ * then use sell using`swapOut` with a minimum of the debt. If that fails
+ * this is not the last tranche, tehn sell with the `swapIn` operation,
+ * using the oracle price reference as the minimum.
```
```suggestion
 * This contract liquidates a payment in multiple tranches to limit the
 * price impact on the AMM of any one sale. Each block it will compute
 * a tranche of collateral to sell, where the size is a function of
 * the amount of that collateral in the AMM pool and the desired price impact.
 * It then gets 3 prices for the current tranche:
 * - AMM quote - compute XYK locally based on the pool sizes
 * - Reserve quote - based on a low price at which the Reserve will purchase
 * - Oracle quote - provided by the oracle that triggered liquidation
 * - Oracle limit - a limit that is DIVERGENCE_TOLERANCE below the
 *   Oracle quote.
 * Then sell based on the following:
 * - If the AMM quote is *below* the oracle limit, skip this block
 *   to wait for oracle and AMM prices to converge.
 *    - should it still sell to the Reserve here?
 * - If the Reserve quote is higher than the AMM quote, sell to the
 *   Reserve with a `want` of the Oracle limit
 * - Otherwise, sell to the AMM with a `want` of the Oracle limit
 * If selling a tranche for the minimum would pay off the debt (`isLast`),
 * then sell using `swapOut` with a minimum of the debt. If that fails
 * this is not the last tranche, so sell with the `swapIn` operation,
 * using the oracle price reference as the minimum.
```

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Then I don't understand the bug. How will 24ea7c83f fail?

That change doesn't fail any tests. Does the bug have a regression test?

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+/** @type {ContractStartFn} */
+const start = async zcf => {
+  const { amm, priceAuthority, timerService, debtBrand } =
+    /** @type {LiquidationContractTerms} */ zcf.getTerms();
+  const nextBlock = async () => E(timerService).delay(1n);
+
+  // TODO make subject to governance
+  const SCALE = 1_000_000n;
+  const BASIS_POINTS = 10_000n * SCALE;
+  const BP2 = BASIS_POINTS * BASIS_POINTS;
```
Chris-Hibbert, line 63: BP2 is 10**20. that's surprising.

I think if you want to use a denominator with higher resolution than Basis Points, it doesn't make sense to use BP_squared. SCALE (=10^6) would be fine, but this code is confusing me when numerators and denominators both have _BP appended to them.

When computing ratios, 30 BP is  30 / 10_000.   If you want something at higher resolution, say a half basis point, it would make sense to think of that as 50 / SCALE.  But multiplying the numerator by SCALE and the denominator by SCALE^2 is a surprising representation.

And mixing SCALE and BP2 is hard to follow.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+
+  const AMM_FEE_BP = 30n * SCALE;
+  const MAX_IMPACT_BP = 50n * SCALE;
```
Chris-Hibbert: Why does this multiply by 10**6 to get a number it calls `_BP`?

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+  const { amm, priceAuthority, timerService, debtBrand } =
+    /** @type {LiquidationContractTerms} */ zcf.getTerms();
+  const nextBlock = async () => E(timerService).delay(1n);
+
+  // TODO make subject to governance
+  const SCALE = 1_000_000n;
+  const BASIS_POINTS = 10_000n * SCALE;
+  const BP2 = BASIS_POINTS * BASIS_POINTS;
+
+  const AMM_FEE_BP = 30n * SCALE;
```
Chris-Hibbert: I would expect AMM_FEE_BP to be `makeRatio(30n, something, 10_000n).

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+  const MAX_IMPACT_BP = 50n * SCALE;
+  // sqrt(impact + 1) - 1
+  // Since we take SQRT but want to be BP, multiply in BP
+  const IMPACT_FACTOR_BP =
+    BigInt(
+      Math.ceil(
+        Math.sqrt(Number((MAX_IMPACT_BP + BASIS_POINTS) * BASIS_POINTS)),
+      ),
+    ) - BASIS_POINTS;
```
Chris-Hibbert: ImpactFactor is getting swamped by BP. This is computing (with some rounding)
```
IMPACT_FACTOR_BP = sqrt((50 * 10^6 + 10^10) * 10^10) - 10^10);
```

If the eventual goal is to use it in  `IMPACT_FACTOR * poolSize /  SCALE` (line 132) to get a tranche size that doesn't move the price more than e.g. .5 BP, I think you want something more like `IMPACT_FACTOR = sqrt(50 + 1) - 1`.   i.e.  7/10^6 change on each side of the pool.

Chris-Hibbert (commented):
Not done reading yet, but the math on max impact was confusing enough that I'm sending these comments now.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+      want: { Out: want },
+    };
+    const { deposited, userSeatPromise: liqSeat } = await offerTo(
+      zcf,
+      swapInvitation,
+      undefined, // The keywords were mapped already
+      proposal,
+      debtorSeat,
+      debtorSeat,
+      { stopAfter: debt },
```
Chris-Hibbert, line 223: The fact that adding a parameter made you fill in a previously-optional parameter argues that `offerTo` should take a bag of parameters.   Certainly not something to address here.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+
+  // the maximum number fo failure sina  row before we stop liquidating.
+  // TODO make this larger andmake it reset on success
+  const MAX_FAILURES = 4;
+  let failures = 0;
+
+  const computeAMMProceeds = (poolCentral, poolCollateral, tranche, debt) => {
+    const k = NatMath.multiply(poolCentral.value, poolCollateral.value);
+    const postSaleCollateral = AmountMath.add(tranche, poolCollateral);
+    const quoteValue = NatMath.subtract(
```
Chris-Hibbert: I'd call this poolCentralEstimate, I think. calling it quote made it seem like it was modelling what the price authority would say or something.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,322 @@
...
+      poolCentral.value,
+      NatMath.floorDivide(k, postSaleCollateral.value),
+    );
+    const ammQuote = AmountMath.make(debt.brand, quoteValue);
```
Chris-Hibbert: this is also an estimate, not a quote.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: That's further obscuring the interlock around `outstandingQuote`: 
```
if (!outstandingQuote) {
  ...return...
}
outstandingQuote = somethingNonNull;
```
The key is that MUST be an atomic block without an `await`. Putting the `await` before the `somethingNonNull` opens a race that hid a test failure. I'm considering refactoring it a different way so that the atomicity can't be compromised.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: Hah: ```
Property 'next' does not exist on type '[string, KindFacet<{ getVaultSeat: ({ state }: MethodContext) => ZCFSeat; initVaultKit: ({ state, facets }: MethodContext, seat: ZCFSeat) => Promise<{ publicNotifiers: { ...; }; invitationMakers: { ...; } & RemotableBra```

Nope!

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: It complains if you try to return something because of "inconsistent return". Since they are in theory equivalent, it's not worth working up.

packages/run-protocol/test/bundleTool.js
dtribble: it should be a separate commit here directly from the source branch, so won't this get dropped out in rebase once that's in `master`?

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: yep. tests to be added at the bottom. WIP.  Moving out

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: agreed. separate PR for that?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: I have format on save. No idea how this was incorrectly indented in the first place.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: These are actually different: with no argument, it asks for the current state. By passing the argument, it says "block until the state actually changes".  Since I'm trying to verify that the state does NOT change, it needs to not provide the argument. NOTE that there was a test bug that it was not verifying that things did NOT change. Thus when the interlock bug allowed multiple outstanding changes, bob's vault liquidated too early but this test didn't notice.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: I'm trying a different approach in test-liquidator. Not going to try to fix it here, but I'l take every idea for the new tests.

packages/run-protocol/test/bundleTool.js
turadg: I don't follow. Are you planning to use automerge's `squash`, `merge` or `rebase`? 

Either way, I think what you want to do is before your next push fixup an earlier commit with https://github.com/Agoric/agoric-sdk/pull/5129/commits/b586fe8c13d9d180ddcc436116f46dcabe6b351b so it's not in the commit list.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: It is at the top already. I'll add `@file`

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: I expanded the TODO.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: This makes me note that `isLast is no longer required and shoudl be removed.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: These need to be ripped out and replaced with governance. That's the separate ticket for governance.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: This needs to be removed. That was for dealing with runaways during development. There's no number that's right for production.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: Correct. This is outside the scope of those.

packages/run-protocol/src/vpool-xyk-amm/swap.js
dtribble: resolved elsewhere.

packages/run-protocol/src/vpool-xyk-amm/swap.js
dtribble: They are equivalent, and the current code is clearer to me. Also, the type verifier doesn't treat them the sameyet and some sometimes that complains.

dtribble: So feel free to change, but out of scope for this PR :)

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: renamed with "estimate" in most of the names

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
dtribble: after this PR :)

packages/run-protocol/src/collect.js
turadg: > It's used in tests.

The only mapValues with key I see  this commented out line,
```
  // Collect.mapValues(contractRoots, (root, k) => loader.load(root, k)),
```

> Is this the right type:

Yep.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
turadg: I meant tippy top, above imports, but under imports is fine

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Hmm, did you have exactly this line?
```
    trace('price check liq', next?.[0]);
```

When I check out your branch and make that edit it's golden.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: > they are in theory equivalent

I thought the `await` would force something to pause until the next turn.

Also, won't awaiting a recursive call cause a monotonically growing call stack? 

>It complains if you try to return something because of "inconsistent return"

Ah, that's eslint `consistent-return` complaining because it doesn't know the real type. In both cases the return should be `Promise<void>` so the eslint erroneous error is fair to suppress.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: > These are actually different: with no argument, it asks for the current state. By passing the argument, it says "block until the state actually changes".

Yeah, that's what I consider an improvement.

> Since I'm trying to verify that the state does NOT change

But that's brittle because you don't know whether it _will_ change. 

> it needs to not provide the argument

Not necessarily; it could provide `updateCount - 1`. Then that would help express the intent to the reader.

packages/run-protocol/src/vaultFactory/vaultManager.js
mhofman: Ok that is a fair case of eslint being too limited not knowing the return type.

That said, the preferable approach here should depend on how you want the stack to look like. If it's really recursive, then you indeed want `return` for doing a kind of "async tail call optimization". If you want to preserve the recursion in a stack trace, you need to `await`.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: New comment looks great. thanks.

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -0,0 +1,302 @@
...
+  };
+
+  function computeOracleLimit(oracleQuote, oracleTolerance) {
+    return ceilMultiplyBy(getAmountOut(oracleQuote), oneMinus(oracleTolerance));
+  }
+
+  const getAMMFeeBP = async () => {
+    const zoe = zcf.getZoeService();
+    const instance = await E(zoe).getInstance(E(amm).makeSwapInvitation());
+    const terms = await E(zoe).getTerms(instance);
```
Chris-Hibbert, line 103: I don't think you need two awaits.
```suggestion
    const instanceP = E(zoe).getInstance(E(amm).makeSwapInvitation());
    const terms = await E(zoe).getTerms(instanceP);
```

Chris-Hibbert (approved):
I don't have remaining concerns. This is acceptable to me.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
dtribble: That doesn't address it either. We need to include settling.

PR #5130, opened 2022-04-17
Implement durable promise watchers

### LINKED ISSUE #5006: register handlers for durable promises

At today's meeting we figured out a smaller approach for #3787 . The problem statement is:

* a vat subscribes to some large number of imported promises
* the vat gets upgraded
* the new version still needs to react to those promises becoming resolved

The "large number" means that we don't want to consume RAM for each promise while we're waiting for it to be resolved. We aren't sure we can manage that in the short-term, but changes to `HandledPromise` (in particular being able to sense `.then` being called) might enable it.

For now we're going to focus on allowing the callbacks to survive upgrade.

## Description of the Design

Rather than implementing something generic like (#4932), we realized we can do something specifically for promise handlers. We'll define a new kind of durable collection. This collection is like a Map: it is enumerable, and stores at most one value per key. The keys are Promises, specifically *imported* promises (decided by the kernel, not the vat holding this collection), which either arrived in some inbound message or resolution, or created as the result promise of an outbound message. The values are any durable data, which we name `context`.

Like `defineDurableKind`, each instance of this collection has a long-lived durable Handle that must be created exactly once, and then stored into (or retrieved from) the vat-upgrade "baggage".

Unlike Kinds (where each Kind has a distinct behavior), each *instance* of this collection is configured with a distinct behavior. The author provides this as a pair of functions named `fulfilledFunc` and `rejectedFunc`. These correspond to the two arguments to `Promise.prototype.then()`. But unlike those, these functions receive *two* arguments:
```js
function fulfilledFunc(fulfillment, context) { ... }
function rejectedFunc(reason, context) { ... }
```

Users can add promise/context pairs to this table by calling `init`, just like a Map-like Store.

If/when the Promise becomes resolved, the appropriate fulfillment/rejection function will be invoked, with both the per-Promise `context` and the resolution value. The Promise will also be removed from the table at that time, so the table will only ever hold unresolved Promises.

Like a WeakStore, the table is not enumerable. However the entries are retained until the Promise is resolved, even if everybody else drops the Promise object.

The entries of this table survive the vat being upgraded. Liveslots will convert each promise into its `vpid` and store that in a table, associated with the serialized (durable) `context` data. When a promise resolution notification (`dispatch.notify`) arrives, liveslots will deserialize the context object and invoke the user-provided function.

We couldn't decide on a name for it, so I'm going to call it EnthusiasticWeasel to avoid biasing the eventual bikeshedding.

```js
const handle = makeEnthusiasticWeaselHandle('table name');

function fulfilledFunc(fulfillment, context) {
  E(context.listener).report(context.other, fulfillment);
}
functiln rejectedFunc(reason, context) {
  E(errorHandler).abort(context.other, reason);
}
const table = makeEnthusiasticWeaselTable(handle, fulfilledFunc, rejectedFunc);

root = Far('root', {
  receivePromise(p, listener) { table.init(p, { listener, other }); },
  doThing(target) { table.init(E(target).foo(), { other }); }
});
```

The table should probably include `.delete` functionality. It's conceivable that it should include `.get` and `.set`, but maybe we can get away without it.

## Future Syntax

@dtribble suggested it would be interesting to integrate this into our eventual-send operation, by adding options that name a table and a context. Something like:
```js
E(target, { watcher: [ table, { other } ] }).foo();
```
which would automatically do `table.init(p, { other })`. This could add the `vpid` without needing to reify the result `Promise` object (although we'd need to do that anyways if we couldn't tell if someone paid attention to the return value of `E(..).foo()`).

Independently of that, if we find a way to instrument the HandledPromise returned by `E(..).foo()` to sense the use-vs-neglect of `.then`, then we could avoid reifying the Promise entirely, which means the memory footprint of these result promises *and* their callbacks goes away, making this suitable for high-cardinality subscribers.

## Security Considerations

Each EnthusiasticWeasel table effectively provides access to the values captured by the handler functions, so different components of the system should have their own instances. This lines up well with our planned usage of Handles for Kind registration, e.g. Zoe gets access to one portion of "baggage", while the contract gets a separate portion, and each can stash their own instance of these tables in their own portion.

Allowing `.get` or `.set` would allow someone to amplify access to both the Promise and the table into access to the handler, which would be surprising (generally, having access to a Promise means you can add *new* callback behavior, but not invoke or read or interfere with anybody else's callbacks). That's probably a good argument to omit `.get` and `.set`, and maybe even `.delete`. And it means the Store's use of `.init` instead of `.set` is great, no possibility of overwriting someone else's behavior.

## Test Plan

Unit tests.

## Alternate Designs

We originally considered making the table enumerable but not associating behavior. Instead, at the beginning of each new vat version, user code would be able to enumerate the promises in the table, and could add their own `.then` handlers to each. This requires reifying all Promises (which costs RAM) when in fact the behavior can probably be factored out and provided just once (using a context object to behave differently for each Promise).

Instead of providing the two behavior functions at table definition time, we considered an additional call which sets (or re-sets) the behavior later. We've observed that a lot of our durable behavior tools wind up involved in circular definitions, where we need to access the table from some behavior functions, but we need those functions before we can defined the table. By providing behavior to `makeEnthusiasticWeaselTable()`, we avoid the questions of "what if someone forgets to call `setBehavior`?" and "what if someone calls it twice?", but we push the mutable-binding complexity onto the user.

The promise provided to `table.init(p, context)` must be decided by the kernel, not the vat. It will throw an error if given a promise that is (or would be) decided by the local vat, e.g. a Promise that liveslots does not yet know about (which would be exported to the kernel and thus decided by the vat). We considered allowing these promises to be used, and invoking the handler functions if the vat subsequently resolved them, but that leaves open a question about what should happen in version-2. Only the old version-1 RAM image had access to the `resolve`/`reject` functions, so after upgrade, these promises won't ever resolve (the Promises might be durable, but the resolvers/rejectors are not). So we'd have to arrange for liveslots to reject them all (e.g. call all their `rejectionF` behaviors) during upgrade, shortly after the new behaviors were reassociated. That seemed unhelpful, so we decided to refuse to accept non-imported Promises outright.

## Names

Some words that might go into the name:
* register
* define
* unsettled
* promise
* when
* then
* table
* store
* durable
* handler
* behavior
* callback
* callbacks

Note that the callbacks/behavior themselves are not durable (each version of the vat must supply new ephemeral functions). The context object *is* durable (`table.init` will throw an error if you attempt to save non-durables in `context`, just like a durable MapStore). The Promises must be imported, so it might be fair to call them durable promises.

## CC

@FUDCo @erights @michaelfig

FUDCo:
Closes #5006

This provides two new functions on `VatData`, per the design hashed out in #5006 as modified by the experience of implementing it:

```
providePromiseWatcher(kindHandle, fulfillHandler, rejectHandler)
watchPromise(p, watcher, ...args)
```
where:
`kindHandle` is a kind handle such as is returned by `makeKindHandle`
`fulfillHandler` is a function of the form `(value, ...args) => void`
`rejectHandler` is a function of the form `(err, ...args) => void`
`p` is a promise
`watcher` is a watcher object (explained below)
`...args` are arbitrary arguments that will be passed to the watcher when the promise is fulfilled or rejected

`providePromiseWatcher` returns a promise watcher object for use in `watchPromise`.  The two handler arguments are functions that will handle promise fulfillments and rejections respectively.

The `watcher` that is passed to `watchPromise` can be either a watcher object as returned by `providePromiseWatcher` or any durable object (durable object facet) that implements one or both of the methods:
```
onFulfilled(context, value, ...args)
onRejected(context, err, ...args)
```

`watchPromise` can watch any promise and this watching will survive shutdown and restart (upgrade) of its vat, though of course when the vat restarts, promises for which the vat itself was the decider will all have been rejected by the act of shutdown.


FUDCo:
Turns out this needs TS type defs in the `vat-data` package.  I have no clue how to do this, so help would appreciated on that score.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -719,7 +719,10 @@ export default function buildKernel(
       return `notify(vatID: ${message.vatID}, kpid: @${message.kpid})`;
     } else if (message.type === 'create-vat') {
       // prettier-ignore
-      return `create-vat ${message.vatID} opts: ${JSON.stringify(message.dynamicOptions)}`;
+      return `create-vat ${message.vatID} opts: ${JSON.stringify(message.dynamicOptions)} vatParameters: ${JSON.stringify(message.vatParameters)}`;
+    } else if (message.type === 'upgrade-vat') {
+      // prettier-ignore
+      return `upgrade-vat ${message.vatID} upgradeID: ${message.upgradeID} vatParameters: ${JSON.stringify(message.vatParameters)}`;
```
mhofman, line 725: Drive by change?

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+    } else {
+      watchedPromiseTable.init(vpid, harden([[watcher, ...args]]));
+      pseudoThen(p, vpid);
+    }
+  }
+
+  function providePromiseWatcher(
+    kindHandle,
+    fulfillHandler = x => x,
+    rejectHandler = x => x,
```
mhofman: Default reject should throw
```suggestion
    rejectHandler = x => { throw x; },
```

I know we don't yet have a good mechanism to report unhandled rejections, but at least it would be the correct default behavior.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+
+    p.then(
+      res => settle(res, true),
+      rej => settle(rej, false),
+    );
+  }
+
+  function loadWatchedPromiseTable() {
+    const deadPromisesRaw = syscall.vatstoreGet('deadPromises');
+    if (deadPromisesRaw) {
```
mhofman: From what I gather, `deadPromisesRaw` is only missing the first time a vat is created, and will currently always exist on upgrades, albeit possibly empty?

I'm concerned that the restoration of in-RAM reactions to the watched promises is conditioned on the existence of a `deadPromisesRaw`. If there is ever a refactor omitting an empty `deadPromises` in the vat store, reactions wouldn't be restored.

I'd prefer if the `for...of` loop was unconditional and the dead condition was updated to `deadPromises && deadPromises.has(vpid)`

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+        if (handler) {
+          handler(value, ...args);
+        }
+      }
+    }
+
+    p.then(
+      res => settle(res, true),
+      rej => settle(rej, false),
+    );
```
mhofman, line 85: Doesn't this force the presence of an imported promise to stay in RAM?

I was hoping that in the case of an imported promise (e.g. as the result of an `E().foo()`, or as an argument), the  liveslots created instance could be recognized and the "listen" wiring would be done without touching the promise itself, allowing it to be garbage collected.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+      const deadPromises = new Set(deadPromisesRaw.split(','));
+      syscall.vatstoreDelete('deadPromises');
+      for (const [vpid, watches] of watchedPromiseTable.entries()) {
+        if (deadPromises.has(vpid)) {
+          watchedPromiseTable.delete(vpid);
+          for (const watch of watches) {
+            const [watcher, ...args] = watch;
+            if (watcher.onRejected) {
+              watcher.onRejected('vat upgraded', ...args);
+            }
```
mhofman: `onRejected` is user provided and can throw. Furthermore, user-code should not be executed synchronously in the middle of liveslots init.

```suggestion
            Promise.resolve().then(() => { watcher.onRejected('vat upgraded', ...args); });
```

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+  function pseudoThen(p, vpid) {
+    function settle(value, wasFulfilled) {
+      const watches = watchedPromiseTable.get(vpid);
+      watchedPromiseTable.delete(vpid);
+      for (const watch of watches) {
+        const [watcher, ...args] = watch;
+        const handler = wasFulfilled ? watcher.onFulfilled : watcher.onRejected;
+        if (handler) {
+          handler(value, ...args);
+        }
```
mhofman: handler is user provided and can throw, which shouldn't impact other watchers.

```suggestion
        try {
          handler(value, ...args);
        } catch(e) {}
```

But I would actually prefer if each handler was called on a clean stack:

```suggestion
        Promise.resolve().then(() => { handler(value, ...args); });
```

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+    const watcherVref = convertValToSlot(watcher);
+    assert(watcherVref, 'invalid watcher');
+    const { virtual } = parseVatSlot(watcherVref);
+    assert(virtual, 'promise watcher must be a virtual object');
+    if (watcher.onFulfilled) {
+      assert.typeof(watcher.onFulfilled, 'function');
+    }
+    if (watcher.onRejected) {
+      assert.typeof(watcher.onRejected, 'function');
+    }
```
mhofman: I'm not sure we should care whether those methods exist on the object or not. Maybe just triggering an unhandled rejection at resolution time is enough?

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+      for (const watch of watches) {
+        const [watcher, ...args] = watch;
+        const handler = wasFulfilled ? watcher.onFulfilled : watcher.onRejected;
+        if (handler) {
+          handler(value, ...args);
+        }
+      }
+    }
+
+    p.then(
```
mhofman, line 82: FYI, since this can be a user provided promise, this can technically throw or synchronously re-enter into liveslots, but from what I gather, this is safe in this case. If we could add a note however, that would be great.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+    const { virtual } = parseVatSlot(watcherVref);
+    assert(virtual, 'promise watcher must be a virtual object');
+    if (watcher.onFulfilled) {
+      assert.typeof(watcher.onFulfilled, 'function');
+    }
+    if (watcher.onRejected) {
+      assert.typeof(watcher.onRejected, 'function');
+    }
+
+    const vpid = convertValToSlot(p);
```
mhofman: `convertValToSlot` will brand test the promise, which in the case of an evil promise can cause re-entrancy (the Promise internals will however absorb any thrown exception). Given the position of the call, it should be safe, but maybe add a note?

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,162 @@
...
+    assert(vpid, 'invalid promise');
+    const { type } = parseVatSlot(vpid);
+    assert(type === 'promise', 'watchPromise only watches promises');
+    if (watchedPromiseTable.has(vpid)) {
+      const watches = watchedPromiseTable.get(vpid);
+      watchedPromiseTable.set(vpid, harden([...watches, [watcher, ...args]]));
```
mhofman: Out of curiosity, why not nest a durable scalar set instead of rewriting an array value?

mhofman (dismissed):
Preliminary review of the implementation.

The blocker right now is the mishandling of throwing handlers, and the synchronous invocation of these during init for disconnected promises.

I am also concerned that this approach has an unnecessary RAM cost for imported promises.

Rest is mostly nits and clarifications.

> two new functions on `VatData`
> `providePromiseWatcher(kindHandle, fulfillHandler, rejectHandler)`

I was hoping the `providePromiseWatcher` could live in user-land (e.g. in `@agoric/vat-data`), but I suppose we need a mechanism to retrieve the singleton instance on upgrade? I kinda wish `vat-data` could keep `promiseWatcherByKindTable` in the baggage, but I suppose we still don't have a good mechanism to initialize such contract side helpers besides putting stuff on the global object when creating the compartment.

> any durable object (durable object facet) that implements one or both of the methods

My reading of the code is that both are not required. However I'm not sure if we should test the object. We could imagine a future where the watcher just needs to be durable compatible, and the actual watcher is in fact in a remote vat, in which case testing can only happen for local watchers.

> Turns out this needs TS type defs in the `vat-data` package. I have no clue how to do this, so help would appreciated on that score.

I can give it a shot tomorrow.

packages/SwingSet/src/kernel/kernel.js
FUDCo: Missing diagnostic output that I needed in the course of debugging this.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: In our long discussion we talked about this and concluded that throwing would be Bad.  I don't recall the reasoning however.  I don't have a particular opinion on this myself, other than a nagging worry about who would be catching this if it threw.

mhofman: From what I remember we mentioned we were not going to try and chain watchers for now (it's very much possible, but we need a proper virtual promise abstraction then), which meant that the user's watcher should handle errors within the watcher.

However I firmly believe we shouldn't explicitly swallow errors if we can help. I think we should root all watcher callbacks in a new call stack / turn, and let the errors bubble to whatever future unhandled rejection logic may exist.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Since the `'deadPromises'` (not `deadPromisesRaw`, that's just a variable name) entry in the vatstore is created by the shutdown operation, you're correct that it will be missing at vat creation time and should always exist at startup time thereafter.  Consequently the existence of the `'deadPromises'` key in the vatstore is a sort of proxy for "yes, we are doing a restart".  Its absence on restart would be symptom that something terrible had gone wrong with the shutdown, and I suspect in that case restoring the promise watchers would be the least of our concerns.

That said, I like the idea of decoupling the fetching of the list of dead promises from the restoration of the promise watchers.  I think the approach I'd take would, rather than changing the conditional on the loop, is to simply initialize the `deadPromises` variable to an empty Set in the case where the `'deadPromises'` key is absent.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Yes and no.  It does rely on the promise being in RAM but if there's an imported promise it's going to be in RAM anyway, so doing this costs us nothing extra.  Keep in mind that this is not a mechanism to virtualize promises, it's more a way to virtualize `then` calls.  In this case virtualization is not for the purpose of reducing the RAM footprint but for the purpose of gaining durability.  In other words, if instead of using a `promiseWatcher` someone had done a regular `then` call they'd also be holding it in RAM (albeit their attention to the promise would be volatile) -- which is what you'd want because GC of the promise would otherwise cause it to be released by the vat and the import dropped, and so the kernel would not send a notification to the vat when the promise is resolved.

Once we have actual virtual promises then we'll be in a position to keep an on-disk refcount for a promise, keeping the promise alive from the kernel's perspective but allowing the JavaScript promise object to disappear from RAM inside the vat, and then having a mechanism whereby the delivery of the promise resolution notification from the kernel would trigger the flow into whatever handlers had been registered.  But this is not that.

mhofman: > if there's an imported promise it's going to be in RAM anyway

I didn't realize that the promise instance created by an imported promise was held strongly by anything in liveslots.

> In this case virtualization is not for the purpose of reducing the RAM footprint but for the purpose of gaining durability.

Ack, was just hoping this could put us on the path to both.

> having a mechanism whereby the delivery of the promise resolution notification from the kernel would trigger the flow into whatever handlers had been registered

Right, that's indeed what would be needed.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Brian and I just discussed this and persuaded ourselves that you are right about how the default reject handler should be written.  I think there's also an analogous tweak to do in the case where the user provides their own watcher facet (instead of using the thing handed back by `providePromiseWatcher`) and their watcher facet has no `onRejected` method.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: If you have multiple promise watchers on the same promise, it's possible this is due to a division of labor (presumably having something to do with how the user code is architected) in which, say, one watcher is dealing with the success case and a different one is dealing with the failure case.  So it seemed reasonable to allow a handler to simply be absent.  In particular, it's only important that _somebody_ handle the rejection as long as somebody actually does.  However, given the consensus that we need to be extra careful about unhandled rejections, I think substituting the default rejection handler (i.e., throw) when `onRejected` is absent is probably the right thing to do, so that swallowing the rejection without action must be a deliberate user choice.  On the other hand, I think treating an absent `onFulfilled` as a no-op seems fine.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: I'm not sure adding disclaimers everywhere the code touches a promise is helpful.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: I thought about that, but since we expect the typical cardinality of this array to be small (in fact, usually 1), I think the overhead of a durable set would almost always dwarf the cost of rewriting the array.  Plus, the entry is the watcher plus its args, so it would end up being a set of copy arrays (even more expensive).  Either that or a map from a nonce or counter to the watcher plus args, and then we'd also have to keep track of the counter.  Anyway, a tarbaby of multiplying complications.

packages/SwingSet/src/liveslots/watchedPromises.js
mhofman: IMO watchers being an object differ enough from promise reactions as independent callbacks, that requiring a custom watchers (i.e. the ones not using `providePromiseWatcher`) to implement both isn't too much to ask. However I think my point here was that I'm not sure we should perform much validation of the watcher object beside it's an object that can be stored durably.

packages/SwingSet/src/liveslots/watchedPromises.js
mhofman: I think it's fair to assume `convertValToSlot` implies the object will be probed and could perform re-entrancy. I'm ok with no comment in this case.

packages/SwingSet/src/liveslots/watchedPromises.js
mhofman: I figured that was the case.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+
+  function loadWatchedPromiseTable() {
+    const deadPromisesRaw = syscall.vatstoreGet('deadPromises');
+    let deadPromises;
+    if (deadPromisesRaw) {
+      deadPromises = new Set(deadPromisesRaw.split(','));
+      syscall.vatstoreDelete('deadPromises');
+    } else {
+      deadPromises = new Set();
+    }
```
mhofman: Silly question, why not write it as:
```suggestion
    const deadPromisesRaw = syscall.vatstoreGet('deadPromises') || '';
    syscall.vatstoreDelete('deadPromises');
    const deadPromises = new Set(deadPromisesRaw.split(','));
```

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+    for (const [vpid, watches] of watchedPromiseTable.entries()) {
+      if (deadPromises.has(vpid)) {
+        watchedPromiseTable.delete(vpid);
+        for (const watch of watches) {
+          const [watcher, ...args] = watch;
+          if (watcher.onRejected) {
+            Promise.resolve().then(() => {
+              watcher.onRejected('vat upgraded', ...args);
+            });
+          }
```
mhofman: This was surprisingly inconsistent with the normal rejection behavior which throws the error if `onRejected` doesn't exist, but I suppose we wouldn't win much from throwing a `'vat upgraded'` for very single of these now dead promises.

In a future where `watchPromise` creates a chained virtual promise, we may be able to do better, but maybe not.

mhofman (commented):
Thanks for addressing my initial concerns. Will review the tests later this afternoon.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+      }
+    }
+
+    p.then(
+      res => settle(res, true),
+      rej => settle(rej, false),
+    );
+  }
+
+  function loadWatchedPromiseTable() {
```
warner, line 90: Let's shuffle the functions in this file to match their temporal invocation order:
* first the two that liveslots `startVat` does:
  * `preparePromiseWatcherTables`
  * `loadWatchedPromiseTable`
* then the two that appear in user code:
  * `providePromiseWatcher`
  * `watchPromise` (which isn't generally called until after using `providePromiseWatcher` to create the watcher)

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+      // artificially increment the table's refcount so it never gets GC'd
+      vrm.addReachableVref(watcherTableID);
+    }
+
+    let watchedPromiseTableID = syscall.vatstoreGet('watchedPromiseTableID');
+    if (watchedPromiseTableID) {
+      watchedPromiseTable = convertSlotToVal(watchedPromiseTableID);
+    } else {
+      watchedPromiseTable = makeScalarBigMapStore('watchedPromises', {
+        keySchema: M.string(),
```
warner: let's add a `// vpid` comment to the schema line, a cheap way to help orient future maintainers

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  function prepareShutdownRejections(deciderVPIDs) {
+    const deadPromises = [];
+    for (const vpid of deciderVPIDs) {
+      if (watchedPromiseTable.has(vpid)) {
+        deadPromises.push(vpid);
+      }
+    }
+    syscall.vatstoreSet('deadPromises', deadPromises.join(','));
```
warner, line 187: the use of `,` as a delimiter means we should probably update the comment in `parseVatSlot.js` to prohibit comma as a component in vrefs, with a pointer to this file and the `deadPromises` record

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+
+    p.then(
+      res => settle(res, true),
+      rej => settle(rej, false),
+    );
+  }
+
+  function loadWatchedPromiseTable() {
+    const deadPromisesRaw = syscall.vatstoreGet('deadPromises');
+    let deadPromises;
```
warner: we could shave a few lines by using `let deadPromises = new Set()` instead of the `else` below, but I don't know how that sits with your aesthetics

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+        for (const watch of watches) {
+          const [watcher, ...args] = watch;
+          if (watcher.onRejected) {
+            Promise.resolve().then(() => {
+              watcher.onRejected('vat upgraded', ...args);
+            });
+          }
+        }
+      } else {
+        const p = convertSlotToVal(vpid);
```
warner: This `convertSlotToVal` is an interesting case: when version-2 starts, the kernel will see a duplicate `syscall.subscribe()` for each not-dead yes-watched promise `vpid`, because this is normally the pathway by which the vat imports promises that arrive in message arguments. If the kernel-side promise is unresolved, the `syscall.subscribe` is a no-op (`kernelKeeper.addSubscriberToPromise` uses a `Set` for deduplication). If the promise already was resolved (perhaps it was resolved just before the vat is upgraded, but the `notify` is still sitting in a queue), the kernel will enqueue a duplicate `notify` event, however that will be discarded when it arrives at the vat and we see the `vpid` missing from the c-list.

So, not a problem, but something to keep in mind.

packages/SwingSet/test/promise-watcher/bootstrap-promise-watcher.js
```diff
@@ -0,0 +1,45 @@
...
+
+export function buildRootObject() {
+  let vatAdmin;
+  let uptonRoot;
+  let uptonAdmin;
+  const pk1 = makePromiseKit();
+  const pk2 = makePromiseKit();
+  const pk3 = makePromiseKit();
+  const pk4 = makePromiseKit();
+
```
warner, line 14: this test is great! I think we should exercise a few additional things:
* result promises: have vat-bootstrap create another two PromiseKits, and create a pair of `Far` objects which return the `.promise` (one each). Then inside `upgradeV2`, resolve one and reject the other. Create a second pair of `Far` objects, one of which returns `val5` and the other does `throw('val6')` or something. The vat under test should `VatData.watch(E(far1).foo())` for all four: two should resolve/reject promptly, the other two will fire after upgrade, and the watcher should see it in version-2.
* let's exercise the multiple argument ability: `VatData.watch(p, watcher, 'arg1', 'arg2')`, where we're currently only passing a single argument
* the test that exercises the `makeDK` option should exercise the omission of `onFulfilled` and/or `onRejected`, to demonstrate that it doesn't crash anything (unless that runs afoul of an aggressive unhandled-rejection checker)

warner (changes requested):

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,186 @@
...
+        Promise.resolve().then(() => {
+          if (wasFulfilled) {
+            if (watcher.onFulfilled) {
+              watcher.onFulfilled(value, ...args);
+            }
+          } else {
+            if (watcher.onRejected) {
+              watcher.onRejected(value, ...args);
+            } else {
+              throw value;
```
warner: oh and let's add a comment here, something like `// allow unhandled rejection handler to fire`

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Not silly.  It literally didn't occur to me that you could split an empty string, though of course you can.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: I think this was just an oversight on my part.  I'll tweak.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: @mhofman already came up with a more elegant formulation which I have incorporated.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: For the record: per our conversations we've replaced this use of `convertSlotToVal` with a new liveslots function, `revivePromise`, which is expected to be used in this context and avoids the various landmines.

packages/SwingSet/src/liveslots/stop-vat.js
```diff
@@ -41,7 +41,7 @@ function rejectAllPromises({ m, deciderVPIDs, syscall }) {
   // and attached a .then to it), and stopVat() must not allow
   // userspace to gain agency.
 
-  const rejectCapData = m.serialize('vat upgraded');
+  const rejectCapData = m.serialize(Error('vat upgraded'));
```
mhofman: I vaguely remember a discussion about the upgrade message being differentiable by subscribers because it was a string instead of an Error like when a vat is terminated for failure. @warner

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -52,94 +86,30 @@ export function makeWatchedPromiseManager(
...
-          if (watcher.onRejected) {
-            Promise.resolve().then(() => {
-              watcher.onRejected('vat upgraded', ...args);
-            });
-          }
+          Promise.resolve().then(() => {
+            if (watcher.onRejected) {
+              watcher.onRejected(Error('vat upgraded'), ...args);
+            } else {
+              assert.fail('vat upgraded');
```
mhofman: Maybe we could add which `watcher` wasn't notified in the details? Probably not worth it tho.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -166,6 +136,44 @@ export function makeWatchedPromiseManager(
...
 
+  function watchPromise(p, watcher, ...args) {
+    // The following wrapping defers setting up the promise watcher itself to a
+    // later turn so that if the promise to be watched was the return value from
+    // a preceding eventual message send, then the assignment of a vpid to that
+    // promise, which happens in a turn after the initiation of the send, will
+    // have happened by the time the code below executes, and thus when we call
+    // `convertValToSlot` on the promise here we'll get back the vpid that was
+    // assigned rather than generating a new one that nobody knows about.
+    Promise.resolve().then(() => {
```
mhofman, line 150: Ooof that feels brittle. Any way to pass an option to `convertValToSlot` that there must already be a slot allocated?

packages/SwingSet/test/promise-watcher/unhandledRejectionDetector.js
```diff
@@ -0,0 +1,9 @@
+import process from 'process';
+
+let unhandledRejectionHandler;
+
+export function handleUnhandledRejections(handler) {
+  unhandledRejectionHandler = handler;
+}
+
+process.on('unhandledRejection', rej => unhandledRejectionHandler(rej));
```
mhofman, line 9: If `handleUnhandledRejections` is never called, this will throw in the event emitter, the consequences of which I'm unsure.

mhofman (approved):
A few more nits, but looking good overall!

packages/SwingSet/src/liveslots/stop-vat.js
FUDCo: I was getting complaints from eslint, so assumed it was a mistake.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Yeah, we're a little nervous too.  The problem is that there _isn't_ already a slot allocated so `convertValToSlot` allocates one, but then the message send code allocates another one (in a way that's entirely different).  There's magic voodoo promise plumbing in the middle that one interferes with at one's peril.

packages/SwingSet/test/promise-watcher/unhandledRejectionDetector.js
FUDCo: Don't care.  It's test scaffolding and it's only there when it's used.

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: There's no graceful way to identify the watcher instance in the details, so no.

packages/SwingSet/src/liveslots/watchedPromises.js
mhofman: But by the time `Promise.resolve()` reaction runs and calls convertValToSlot, the goal is that it would be allocated then, or did I misunderstood? If that's the case, there is no way to assert it?

FUDCo: Ah, I misunderstood what you were saying.  Yes, by the time `Promise.resolve()` runs everything is OK -- so I gather you're suggesting an option to assert that everything actually _is_ OK at that point?

mhofman: Aye, just in case something (e.g. the promise implementation) changes from under our feet in some conditions, and our assumption no longer holds

FUDCo: Well, if there _is_ a `slotToVal` entry then `convertValToSlot` will just use it, so in that case there's not actually extra machinery in `convertValToSlot` that we're leaning on here.  And in that case we might just call `requiredValForSlot`.  But I'm pretty sure this would not do the right thing in the case where the vpid is not being generated by a message send, in which case there might well not be an existing `slotToVal` entry and we really do need `convertValToSlot` to do all the work.

FUDCo: And never mind that last comment.  I'm always getting `convertSlotToVal` and `convertValToSlot` mixed up, because we also have `getValForSlot` and `getSlotForVal` where the pattern of the names is reversed.  This confusion has been on our hitlist for a while now.

FUDCo: So let me try again: in the case where the vpid is being generated by a message send, we need the deferred execution that `Promise.resolve()...` provides so that `convertValToSlot` will see the vpid that the message send produced.  However, in the case where it's not a message send, we really do need `convertSlotToVal` to generate the vpid.  In the latter case we don't need the deferral, but at the point where `watchPromise` is called we don't know which case we're in, and the deferral is harmless in the case where we don't actually need it.

mhofman: That makes sense. I guess the only approach would be to somehow synchronously mark the promise returned by the send, and check for it here, but that seems to involved. I'm a little surprised that the promise used as result of the send is processed by `convertValToSlot` with a delay. Isn't there other cases where this would trip things, like using such a promise as a send argument immediately:

```js
const rp = E(alice).foo();
E(bob).bar(rp);
```

packages/SwingSet/src/liveslots/stop-vat.js
warner: We should definitely be consistent (among all places that announce vat-upgrade). I think I've been slightly conditioned to not use Errors because the `marshal` serialization is so.. enthusiastic.. about announcing them to the console (multiple times), and `stop-vat.js` is new so I might have used a string while writing (and running) the tests.

We do need Notifier subscribers to be able to distinguish between vat-upgrade (meaning they should try again) and vat-failure (meaning they should stop). We could either use string-vs-Error for that, or have clients distinguish between `err.message` (which makes us more dependent upon the actual strings, and reduces our wiggle room for changing them in the future).

@erights as keeper of Notifier, do you think we should make vat-upgrade reject promises with a string, so you can change the Notifier client-side pattern to re-try unless it gets rejected with an Error? Or is there some different/better pattern we should use?

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -0,0 +1,194 @@
...
+    for (const [vpid, watches] of watchedPromiseTable.entries()) {
+      if (deadPromises.has(vpid)) {
+        watchedPromiseTable.delete(vpid);
+        for (const watch of watches) {
+          const [watcher, ...args] = watch;
+          Promise.resolve().then(() => {
+            if (watcher.onRejected) {
+              watcher.onRejected(Error('vat upgraded'), ...args);
+            } else {
+              assert.fail('vat upgraded');
```
warner: Let's build a shared `Error('vat upgraded')` at like line 99, and then change this `assert.fail` into `throw err`. Using an `assert` makes it read like this is a "shouldn't happen" case, where actually what we're doing is allowing the unhandled rejection handler to observe the unhandled rejection.

warner (changes requested):
One minor change, plus we need an answer from @erights about whether `vat upgraded` should be an Error or a string. After that we're good to land.

packages/SwingSet/src/liveslots/watchedPromises.js
erights: > One minor change, plus we need an answer from @erights about whether `vat upgraded` should be an Error or a string. After that we're good to land.

Let's try a string. I bet that non-error rejections are only lightly tested at this point, so it wouldn't surprise me if it revealed some such bugs. Which would be a good thing.

Unless there are surprises, I think I like the string idea. Or at least something that is obviously not an Error.

FUDCo: Well OK then.

mhofman: @erights I assume you mean to comment on https://github.com/Agoric/agoric-sdk/pull/5130/files#r855445250 ?

For here, an error instance is appropriate. However, we probably should use a shared frozen error built once outside the loop

packages/SwingSet/src/liveslots/stop-vat.js
mhofman: I'll open an issue. I feel uncomfortable discriminating based on string vs error instance like this for notifiers

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Two additional thoughts: the original reason I switched it from a string to an Error was that eslint whines if you throw a string.  Obviously this can be suppressed with a disabling comment, bit it seems kinda hinky.  Also, I think I'm unclear what the benefit of using a string in this particular case is supposed to be.  I understand flushing out bugs, but is this the right place to be doing that?

mhofman: String is useless here, I think the comment was about the place you changed from string to error. You change shows that the current reliance on string vs error is a little brittle. I'm opening a separate issue for that.

FUDCo: What reliance on string vs. error?  I don't understand what you're referring to here.

packages/SwingSet/src/liveslots/stop-vat.js
mhofman: @erights @warner @gibson042 can you help me triage https://github.com/Agoric/agoric-sdk/issues/5185
Not sure which label it should fit under.

packages/SwingSet/src/liveslots/watchedPromises.js
mhofman: Please see https://github.com/Agoric/agoric-sdk/pull/5130/files#r855510573

FUDCo: Oy.  My preference would be to use Error here (and everywhere else there's a throw, but that would be a separate task).  And file a ticket to get rid of the annoying behavior of marshal spewing to the log any time it serializes an Error, which i hate, hate, hate.

FUDCo: @erights I guess I'm asking you to reconsider your call above.

erights: Sure. Let's have a conversation where we jointly reconsider. Schedule something?

FUDCo: @erights and I discussed this and concluded that signaling vat upgrade with a string is preferred to using Error (even better would be to have an exceptional condition indicator object analogous to Error but which is not Error, but we don't have the time to prioritize that right now).  Whatever entity is attending to the vat upgrade event should not be discriminating on the basis of whether the thing thrown was string or an Error however, but should specifically be looking for the particular string (in this case, `'vat upgraded'`) that indicates the condition of interest.

On that basis I'm going to revise this PR to revert to using a string.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -710,6 +723,7 @@ function build(
         // this is a new import value
         val = makeImportedPresence(slot, iface);
       } else if (type === 'promise') {
+        // XXX note: this assert is the same as the one 5 lines above and therefor pointless
```
warner, line 726: agreed, that's vestigal, we can remove it the next time we swing through this neighborhood

packages/SwingSet/src/liveslots/watchedPromises.js
mhofman: Can we please make it a `const` at the top level somewhere, and document this string must not change? Then you can throw the const everywhere, and shouldn't have to disable eslint.

packages/SwingSet/test/promise-watcher/test-promise-watcher.js
```diff
@@ -0,0 +1,115 @@
...
+    'p3-dk.res resolved val3 version v2 via VDO (res)',
+    'p4-pw rejected err4 version v2 via watcher [d,e,f]',
+    'p4-dk.full rejected err4 version v2 via VDO',
+    'p4-dk.rej rejected err4 version v2 via VDO (rej)',
+    'rp3-pw resolved rvalafter version v2 via watcher []',
+    'rp3-dk resolved rvalafter version v2 via VDO',
+    'rp4-pw rejected rerrafter version v2 via watcher []',
+    'rp4-dk rejected rerrafter version v2 via VDO',
+  ]);
+  if (os.platform() === 'darwin') {
```
warner, line 108: FWIW, I was able to remove this check on my pseudo-linux box (macOS running a linux kernel under Docker using Node v16.14.0) and also a real linux box (also with Node v16.14.0). What version of Node are you running under macOS locally?

I'm not concerned deeply about testing the unhandled rejection thing, but the `os.platform` check is sorta awkward, so after this lands, let's see if we can get rid of it. Maybe we just need to be running a newer version of Node in CI.

warner (approved):
great!

warner:
Oh see if you can add that `const vatUpgradedMessage = 'vat upgraded'; // part of the swingset API` thing @mhofman recommended.

PR #5133, opened 2022-04-18
fix: interface guards

erights:
Stacked on https://github.com/Agoric/agoric-sdk/pull/5947

Wraps raw methods with a pattern-based defensive wrapper. Like dynamic type enforcement.

Uses an efficient class-like representation, where all methods of a kind are inherited from a per-kind common object.

Endo marshal patch mirrors https://github.com/endojs/endo/pull/1251

### DIFF
--- a/packages/ERTP/src/interfaces.js
+++ b/packages/ERTP/src/interfaces.js
@@ -1,6 +1,5 @@
 /* eslint-disable no-use-before-define */
-/* global foo */
-const M = foo();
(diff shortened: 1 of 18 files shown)


erights:
I hope to replace this PR with https://github.com/Agoric/agoric-sdk/pull/5960 , so I'm moving all the reviewers to that one.

erights:
Subsumed by https://github.com/Agoric/agoric-sdk/pull/5960 . Closing

Issue #4534, opened 2022-02-11
runStake: use virtual objects for high-cardinality state


Tartuffo:
part of #4511
There's a documentation task in #4518 .

Tartuffo:
@dckc Please put an estimate on this.

Tartuffo:
@dckc Only the virtual objects part needs to be done for RUN Protocol RC0 in devnet.  Durable objects can come after.

dckc:
So this should be split into 2 issues?

Tartuffo:
> So this should be split into 2 issues?

Yes, for tracking completion of RUN Protocol Preview release, that makes sense.

Issue #4343, opened 2022-01-21
Use Virtual Objects in Governance


Chris-Hibbert:
## What is the Problem Being Solved?

For MainNet 1, we must support durable objects for restart and upgrade. This is part way to https://github.com/Agoric/agoric-sdk/issues/5200

## Description of the Design

Collect a catalogue of all the persistent data types in Governance, and convert them to Virtual Objects.

## Security Considerations

Trust paging of defineKind APIs.

## Test Plan

Presumably extensive.

Tartuffo:
Does need collections, just virtual objects for durability.

Chris-Hibbert:
A quick survey of current storage:

**BinaryVoteCounter**

```
  /** @type {Store<Handle<'Voter'>,RecordedBallot> } */
  const allBallots = makeStore('voterHandle');
```

  The counter itself is an object with state
    question, isOpen, questionSpec, outcomePromise, tallyPromise

**Committee  (and shareHolders)**

```
  /** @type {Store<Handle<'Question'>, QuestionRecord>} */
  const allQuestions = makeStore('Question');
```

  invitations is an array

   The committee itself is an object with state:
       subscription, publication, committeeName, publicFacet

**contractGovernor** has no stores, but it has state itself

**governParams**

```
  const voteCounters = new WeakSet();

  let poserFacet;
  let currentInvitation;
```

**paramManager**
```
  const namesToParams = makeStore('Parameter Name');
  const { publication, subscription } = makeSubscriptionKit();
  const updateFns = {};
```

turadg:
With https://github.com/Agoric/agoric-sdk/pull/7657 it's mostly durable. Any work now will be on making it fully durable, not virtual,
- #5200

PR #5135, opened 2022-04-18
feat(run-protocol): convert RunStakeManager to vobj

turadg:
part of #4534

## Description

Converts `RunStakeManager` to be a virtual object. RunStakeKit also needs to be converted (https://github.com/Agoric/agoric-sdk/pull/5168 ) . It's more important since it has higher cardinality, but ordering wise it's better to have this one first. 

It defers the governance param durability to https://github.com/Agoric/agoric-sdk/issues/4343

### Security Considerations

Relies on disk IO for state.

--

### Testing Considerations

Refactor, existing tests suffice. Durability is not in scope.

### DIFF
--- a/packages/run-protocol/src/runStake/attestation.js
+++ b/packages/run-protocol/src/runStake/attestation.js
@@ -211,8 +211,8 @@ const makeAttestationIssuerKit = async (zcf, stakeBrand, lienBridge) => {
  * Authorize each account holder to lien some of their staked assets
  * and get an attestation that the lien is in place.
  *
- * @param {ZCF} zcf
- * @param {Brand} stakeBrand
+ * @param {ZCF<import('./runStake').RunStakeTerms>} zcf
+ * @param {Brand<'nat'>} stakeBrand
  * @param {ERef<StakingAuthority>} lienBridge
  *
  * NOTE: the liened amount is kept both here in JS and on the
--- a/packages/run-protocol/src/runStake/attestation.js
+++ b/packages/run-protocol/src/runStake/attestation.js
@@ -211,7 +211,7 @@ const makeAttestationIssuerKit = async (zcf, stakeBrand, lienBridge) => {
  * Authorize each account holder to lien some of their staked assets
  * and get an attestation that the lien is in place.
  *
- * @param {ZCF<import('./runStake').RunStakeTerms>} zcf
+ * @param {ZCF} zcf
  * @param {Brand<'nat'>} stakeBrand
  * @param {ERef<StakingAuthority>} lienBridge
  *
--- a/packages/run-protocol/src/runStake/runStakeManager.js
+++ b/packages/run-protocol/src/runStake/runStakeManager.js
@@ -16,6 +16,86 @@ const { details: X } = assert;
 const trace = makeTracer('RM'); // TODO: how to turn this off?
 
 /**
+ * @typedef {{
+ * compoundedInterest: Ratio,
+ * latestInterestUpdate: NatValue,
+ * totalDebt: Amount<'nat'>,
+ * }} AssetState
+ * @typedef {AssetState & {
+ * assetNotifier: Notifier<AssetState>,
+ * assetUpdater: IterationObserver<AssetState>,
+ * chargingPeriod: bigint,
+ * periodNotifier: Promise<Notifier<bigint>>,
+ * poolIncrementSeat: ZCFSeat,
+ * recordingPeriod: bigint,
+ * startTimeStamp: bigint,
+ * }} ImmutableState
+ * @typedef {{
+ * }} MutableState
+ * @typedef {ImmutableState & MutableState} State
+ */
+
+/**
+ * @param {ZCF} zcf
+ * @param {ZCFMint<'nat'>} debtMint
+ * @param {{ debt: Brand<'nat'>, Attestation: Brand<'copyBag'>, Stake: Brand<'nat'> }} brands
+ * @param {{
+ *  getDebtLimit: () => Amount<'nat'>,
+ *  getInterestRate: () => Ratio,
+ *  getMintingRatio: () => Ratio,
+ *  getLoanFee: () => Ratio,
+ * }} paramManager
+ * @param {MintAndReallocate} mintAndReallocateWithFee
+ * @param {BurnDebt} burnDebt
+ * @param {Object} timing
+ * @param {ERef<TimerService>} timing.timerService
+ * @param {bigint} timing.chargingPeriod
+ * @param {bigint} timing.recordingPeriod
+ * @param {bigint} timing.startTimeStamp
+ *
+ * @returns {State}
+ */
+const initState = (
+  zcf,
+  debtMint,
+  brands,
+  paramManager,
+  mintAndReallocateWithFee,
+  burnDebt,
+  { chargingPeriod, recordingPeriod, startTimeStamp, timerService },
+) => {
+  const totalDebt = AmountMath.makeEmpty(brands.debt, 'nat');
+  const compoundedInterest = makeRatio(100n, brands.debt); // starts at 1.0, no interest
+  const latestInterestUpdate = startTimeStamp;
+
+  const { updater: assetUpdater, notifier: assetNotifier } = makeNotifierKit(
+    harden({
+      compoundedInterest,
+      interestRate: paramManager.getInterestRate(),
+      latestInterestUpdate,
+      totalDebt,
+    }),
+  );
+
+  // ??? does this promise need to be a presence?
+  const periodNotifier = E(timerService).makeNotifier(0n, recordingPeriod);
+  const { zcfSeat: poolIncrementSeat } = zcf.makeEmptySeatKit();
+
+  return {
+    assetNotifier,
+    assetUpdater,
+    chargingPeriod,
+    compoundedInterest,
+    latestInterestUpdate,
+    periodNotifier,
+    poolIncrementSeat,
+    recordingPeriod,
+    startTimeStamp,
+    totalDebt,
+  };
+};
+
+/**
  * @param {ZCF} zcf
  * @param {ZCFMint<'nat'>} debtMint
  * @param {{ debt: Brand<'nat'>, Attestation: Brand<'copyBag'>, Stake: Brand<'nat'> }} brands
@@ -42,7 +122,7 @@ export const makeRunStakeManager = (
   paramManager,
   mintAndReallocateWithFee,
   burnDebt,
-  { timerService, chargingPeriod, recordingPeriod, startTimeStamp },
+  timing,
 ) => {
   /** @param { Amount<'copyBag'>} attestationGiven */
   const maxDebtForLien = attestationGiven => {
@@ -69,16 +149,14 @@ export const makeRunStakeManager = (
     return { maxDebt, amountLiened };
   };
 
-  let totalDebt = AmountMath.makeEmpty(brands.debt, 'nat');
-  let compoundedInterest = makeRatio(100n, brands.debt); // starts at 1.0, no interest
-  let latestInterestUpdate = startTimeStamp;
-  const { updater: assetUpdater, notifier: assetNotifer } = makeNotifierKit(
-    harden({
-      compoundedInterest,
-      interestRate: paramManager.getInterestRate(),
-      latestInterestUpdate,
-      totalDebt,
-    }),
+  const state = initState(
+    zcf,
+    debtMint,
+    brands,
+    paramManager,
+    mintAndReallocateWithFee,
+    burnDebt,
+    timing,
   );
 
   /**
@@ -90,8 +168,7 @@ export const makeRunStakeManager = (
     trace('chargeAllVaults', { updateTime });
     const interestRate = paramManager.getInterestRate();
 
-    // Update local state with the results of charging interest
-    ({ compoundedInterest, latestInterestUpdate, totalDebt } = chargeInterest(
+    const changes = chargeInterest(
       {
         mint: debtMint,
         mintAndReallocateWithFee,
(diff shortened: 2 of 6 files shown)


dckc:
> This [`RunStakeManager`] is the only object I could find for #4534 that will need to be durable.

The individual runStake vaults won't need to be durable? I must be missing something. Where will the balance snapshots be restored from on upgrade?

dckc:
> Refactor, existing tests suffice.

I don't think this is just a refactor. It _should_ have the observable effect of letting the contract support lots of accounts without growing the heap size of the vat, right? Ideally, we would test that; whether the cost is worthwhile in this PR is something I don't feel strongly about. I suppose it's reasonable to rely on the SwingSet layer to do that correctly.

packages/run-protocol/src/contractSupport.js
```diff
@@ -1,4 +1,5 @@
 // @ts-check
+import { q } from '@agoric/assert';
```
dckc: I think `@agoric/assert` is deprecated and the preferred idiom is `const { quote: q } = assert;`.

packages/run-protocol/src/contractSupport.js
```diff
@@ -63,3 +64,20 @@ export const stageDelta = (fromSeat, toSeat, fromLoses, fromGains, key) => {
...
+/**
+ * @param {Amount<'nat'>} debtLimit
+ * @param {Amount<'nat'>} totalDebt
+ * @param {Amount<'nat'>} toMint
+ * @throws if minting would exceed total debt
+ */
+export const checkDebtLimit = (debtLimit, totalDebt, toMint) => {
+  const debtPost = AmountMath.add(totalDebt, toMint);
+  if (AmountMath.isGTE(debtPost, debtLimit)) {
+    assert.fail(
```
dckc: Why a separate `if` and `assert`? Why not just `assert()` what we need?

Also: the code checks for >= but the diagnostic says "exceed", i.e. strictly greater. Which is intended?

packages/run-protocol/src/runStake/attestation.js
```diff
@@ -206,8 +206,8 @@ const makeAttestationIssuerKit = async (zcf, stakeBrand, lienBridge) => {
  * Authorize each account holder to lien some of their staked assets
  * and get an attestation that the lien is in place.
  *
- * @param {ZCF} zcf
- * @param {Brand} stakeBrand
+ * @param {ZCF<import('./runStake').RunStakeTerms>} zcf
```
dckc: `attestation.js` isn't supposed to know about / rely on `runStake.js`. Only the other way around. If this type was somehow necessary, something went wrong.

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -5,123 +5,156 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
  *  getLoanFee: () => Ratio,
- * }} paramManager
- * @param {MintAndReallocate} mintAndReallocateWithFee
- * @param {BurnDebt} burnDebt
+ * }} ParamManager
+ * @typedef {{
+ * compoundedInterest: Ratio,
+ * latestInterestUpdate: NatValue,
+ * totalDebt: Amount<'nat'>,
+ * }} AssetState
```
dckc: reading these without indentation is awkward. Please let's indent, even though we don't yet have tool support.

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -5,123 +5,156 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
+ * chargingPeriod: bigint,
+ * debtMint: ZCFMint<'nat'>,
+ * periodNotifier: Promise<Notifier<bigint>>,
+ * poolIncrementSeat: ZCFSeat,
+ * recordingPeriod: bigint,
+ * startTimeStamp: bigint,
+ * zcf: ZCF,
+ * }} ImmutableState
+ * @typedef {{
+ * }} MutableState
```
dckc: What's this for?

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -5,123 +5,156 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
+    totalDebt,
+    zcf,
+  };
+};
+
+/**
+ *
+ * @param {MethodContext} context
+ */
+const finish = ({ state }) => {
```
dckc: finish what?

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -5,123 +5,156 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
+      {
+        latestInterestUpdate: state.latestInterestUpdate,
+        compoundedInterest: state.compoundedInterest,
+        totalDebt: state.totalDebt,
       },
-      { latestInterestUpdate, compoundedInterest, totalDebt },
       updateTime,
-    ));
+    );
+    Object.assign(state, changes);
```
dckc: Hm. Is this `Object.assign(...)` a recommended pattern? It throws type safety right out the window, no?

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -5,123 +5,156 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
+const finish = ({ state }) => {
+  const { debtMint, mintPowers, periodNotifier, poolIncrementSeat, zcf } =
+    state;
+
   /**
    *
    * @param {bigint} updateTime
-   * @param {ZCFSeat} poolIncrementSeat
    */
-  const chargeAllVaults = async (updateTime, poolIncrementSeat) => {
```
dckc, line 89: Where did `poolIncrementSeat` go? I feel like I missed a memo.

packages/run-protocol/src/runStake/runStakeManager.js
dckc: hm... part of the `defineKind` pattern?

packages/run-protocol/src/runStake/runStakeManager.js
dckc: oh. another part of the `defineKind` pattern.

packages/run-protocol/test/attestation/test-attestation.js
```diff
@@ -60,6 +60,7 @@ const makeContext = async t => {
   const uBrand = stakeKit.brand;
 
   const { publicFacet, creatorFacet } = await makeAttestationFacets(
+    // @ts-expect-error cast
```
dckc: I don't think this cast is necessary. `makeAttestationFacets` should accept any zcf.

packages/run-protocol/test/attestation/test-userFlow.js
```diff
@@ -51,6 +51,7 @@ test(`typical flow`, async t => {
   const address = 'myaddress';
 
   const { publicFacet, creatorFacet } = await makeAttestationFacets(
+    // @ts-expect-error cast
```
dckc: likewise

packages/vat-data/src/types.d.ts
```diff
@@ -23,22 +23,41 @@ type ActualBehavior<B> = {
   [Facet in keyof B]: FunctionsMinusContext<B[Facet]>;
 };
 
-type KindContext<S, B> = { state: S; facets: ActualBehavior<B> };
-
 interface KindDefiner {
+  <P, S, F>(
+    tag: string,
```
dckc: the changes in this file make sense to me in a vague sort of way, but I don't know the defineKind API well enough to approve it. paging @warner ? @FUDCo ?

dckc (changes requested):
I'm mostly still getting up to speed, but the one thing I'm pretty sure of is that this PR shouldn't include a change to the xsnap moddable version.

packages/vat-data/src/types.d.ts
turadg: Pulled out to https://github.com/Agoric/agoric-sdk/pull/5150

packages/run-protocol/src/contractSupport.js
turadg: > I think @agoric/assert is deprecated 

How so? I don't see anything to that effect in the package.

> the preferred idiom is const { quote: q } = assert;.

Preferred why and by whom? The advantage of this one is it auto-imports.

packages/run-protocol/src/contractSupport.js
turadg: > Why not just assert() what we need?

It was to distinguish between “must be true or the system is confused” and “if this isn’t true throw something” (`assert` for the former and `if`/`fail` for the latter). But this came up in the Zoe meeting and it was agreed we won't distinguish between those cases.

I'll simplify this.

> Why a separate if and assert? Why not just assert() what we need?

> the code checks for >= but the diagnostic says "exceed", i.e. strictly greater. Which is intended?

Good question. It should be `isGT` but we don't have that. I'll add it.

packages/run-protocol/src/contractSupport.js
dckc: I'll try to ask around about this bit of lore.

packages/run-protocol/src/contractSupport.js
turadg: https://github.com/Agoric/agoric-sdk/pull/5161

dckc: I'd be a little surprised if @erights prefers to add `isGT` rather than just changing the order of args to `isGTE`.

packages/run-protocol/test/attestation/test-attestation.js
turadg: Got it.

packages/run-protocol/src/runStake/runStakeManager.js
turadg: Good point! WDYT of a [helper for type safety](https://www.typescriptlang.org/play?filetype=js#code/LAKA9AVBoAQTABALgUwLYAcA2BDVMAVGAMwFcstZ4EMcAnHNGAbwIF8TzKQ5FaGmzAAr0kASxxYAPAQB8HfuMlUwoAMYB7AHYBnJDEUSsAQR06xAcy0wAvDAAUZCgBoDoowEpbslrBgB5ACMAKxQ1JAA6HDNLLUcuV0NJDwBuUDY0kFB1bT0YPTwUWxYtRhQALhgAclQAEyrXTSwNOkqquhR6jOyQJJMYq3sC1FdmUrQK6s6qtlSevtNzQeGUUYBrFABPNrqZuaze90lF2KGkQtHxyYBmWcygA)?

turadg: 6019c23a4

packages/run-protocol/src/runStake/runStakeManager.js
turadg: Now that `chargeAllVaults` is defined within `finish()` it's in scope on line 111.

dckc: Reading diffs is evidently not sufficient here. I should check out the branch and look locally.

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -6,123 +6,165 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
       compoundedInterest,
-      interestRate: paramManager.getInterestRate(),
+      interestRate: mintPowers.getGovernedParams().getInterestRate(),
       latestInterestUpdate,
       totalDebt,
     }),
   );
 
+  // ??? does this promise need to be a presence?
+  const periodNotifier = E(timerService).makeNotifier(0n, recordingPeriod);
```
warner: Yes, if you store the `periodNotifier` Promise in `state`, then 1: it will trigger the #5106 bug that I haven't fixed yet, and 2: it won't be durable (once you change `defineKind` to `defineDurableKind`, your `init` calls will start failing with a complaint about a non-durable value (a Promise) being stored in `state`).

Also, it doesn't look like this notifier needs to be available to the main methods.. it's only used during `finish` to wire up a responder. See my comment below on another approach for that.

packages/run-protocol/src/runStake/runStakeManager.js
```diff
@@ -6,123 +6,165 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
...
+    zcf,
+  };
+};
+
+/**
+ *
+ * @param {MethodContext} context
+ */
+const finish = ({ state }) => {
+  const { debtMint, mintPowers, periodNotifier, poolIncrementSeat, zcf } =
```
warner: This use of `finish` (and accessing `state` within it) is beyond our intentions/expectations for virtual objects. Using `state` from outside behavior functions feels like a violation of the intended encapsulation.

I think a better (or less surprising, at least to me) way to approach this would be use `defineKindMulti` to create two facets: one with all the main behavior, and a second with just this `chargeAllVaults` behavior. Then `finish` could wire up the `observeNotifier` to the second facet (ideally just give it behavior methods like `updateState` and `fail` and `finish`).

I don't know how `observeNotifier` is currently implemented, but if we can eventually change it to use a DurableMapStore, then it could store the vref of the second facet, and then it wouldn't even need to keep a Representative in RAM. That doesn't matter here because the runStakeManager is probably a singleton object, but it's probably not the worst pattern to stick to.

Using multiple facets would require an extra wrapper function at the end, so the external caller gets the main facet but never the second one (which is only used for the `observeNotifier`).

With that, the `periodNotifier` wouldn't need to be created during `init` or stored in `state`. It would be created by the wrapper function, which would do the wiring. The `timerService` wouldn't be passed into `init` either. `init` would only get the things that must go into `state`, and `state` would only hold the things needed by behavior methods.

We added `finish` to enable registration of facets to happen before a caller got to see them, but since it doesn't let you influence the return value of the generated construction function (`makeRunStakeManager`), specifically allowing you to remove internal facets from the return value, it only really provides half the job.

I kind of want to get rid of `finish`, or at least remove `state` from its arguments (only passing the facets). I think it was more important in an earlier design for virtual objects (@FUDCo might remember more here), where there was a certain critical moment when object identities had been established but certain code couldn't do anything with them yet. If its current utility is a per-instance post-facet-construction registration/cross-wiring function, but you need an external wrapper anyways (to limit the facets being returned), then it might be just a tempting nuisance, and maybe we should remove it. Or, go all-in and make it capable of doing both jobs (but that would either make it mandatory, or would do something like make the return value mandatory, which I think would be error prone).

@mhofman has some ideas for a different API that would allow this notifier-registration stuff to happen during `init`, which would also remove `finish`, but that feels like a signficant API change and I'm afraid of those at this stage :).

turadg: Thanks for explaining all that about the API design intent, better practices now, and how things might change.

Given the fluctuations in the API, my bar right now is _what works_. I commit to fixing anything that breaks as the API changes (including tighter constraints on how it's used). I expect there will be new hurdles to make the vobjs durable, but I prefer to jump those when I get to them. These big refactors tend to have high churn in merge conflicts so I want this branch to live as briefly as possible.

With [3828638](https://github.com/Agoric/agoric-sdk/pull/5135/commits/3828638d233c88e9882823278369ffef7c3c1e97) and [bcfa3f2](https://github.com/Agoric/agoric-sdk/pull/5135/commits/bcfa3f2d4d6c75f1a9446d2d8c1f0901c7e22dd6) does this PR seem an acceptable use of the vobj API?

turadg:
>> This [RunStakeManager] is the only object I could find for #4534 that will need to be durable.

> The individual runStake vaults won't need to be durable? I must be missing something. Where will the balance snapshots be restored from on upgrade?

I neglected to include RunStakeKit, which indeed must be made durable. Clarified in PR description.

However this PR is only about virtualization. Durability comes after.

>> Refactor, existing tests suffice.
> I don't think this is just a refactor. It should have the observable effect of letting the contract support lots of accounts without growing the heap size of the vat, right? Ideally, we would test that; whether the cost is worthwhile in this PR is something I don't feel strongly about. I suppose it's reasonable to rely on the SwingSet layer to do that correctly.

Ideally, yes. I'm relying on the guarantees of the SwingSet layer. There aren't many guards yet in that layer so this PR could be abusing something, but I'd like to get this progress in ASAP because:
- it moves so much that continual integration with master is laborious
- any abuses will be discovered soon enough in the durability work

dckc (approved):
looks good. thanks for explaining stuff.

PR #5136, opened 2022-04-18
fix(run-protocol): use wallet-friendly offer result notifiers

samsiegart:
refs: https://github.com/Agoric/dapp-treasury/issues/72


turadg (commented):
Surprised no test is failing with this change. I think static typing is the right kind of check for this and we don't have that yet for dapps so I think it can wait until we do.

turadg (approved):

dckc (approved):
thanks, @samsiegart !

PR #5137, opened 2022-04-18
chore: rename 'bundles' dir that wasn't build output

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/4989 had an error that @mhofman found: https://github.com/Agoric/agoric-sdk/pull/4989#discussion_r852306510

This removes the special case and a deeply nested `.gitignore`.

### Security Considerations

--

--

### Testing Considerations

--


packages/SwingSet/test/bundling/test-bundles.js
```diff
@@ -47,7 +47,7 @@ test('bundles', async t => {
   // We save this vat bundle (with 'disk()') to disk, to exercise
   // config.bundles.NAME.bundleSpec
   const diskBundle = await bundleSource(bfile('vat-disk.js'));
-  const diskBundleFilename = bfile('disk-bundle'); // in .gitignore
+  const diskBundleFilename = bfile('bundle-disk.js');
```
mhofman, line 50: Still in `.gitignore`, right? But top level now?

mhofman (approved):

packages/SwingSet/test/bundling/test-bundles.js
turadg: Well, still gitignored but now by a glob. There's no longer an entry for the particular file, which is what I took the comment to mean. I expect our globs to be conventional enough that we don't have to indicate with comments what files will match.

PR #5146, opened 2022-04-19
fix(agoric-cli): Fully validate agoric start portNum

### LINKED ISSUE #5145: `agoric start` with invalid portNum fails slowly and messily
## Describe the bug
`agoric start local-chain` and `agoric start local-solo` both take an optional portNum argument and superficially validate it, but an invalid numeric port still results in filesystem changes and isn't even detected until up to a minute later when some downstream Go code tries to use it.

## To Reproduce
* `agoric start --reset local-chain 65536` (too high; ultimately ag-chain-cosmos fails with «_Error: failed to listen on 0.0.0.0:65536: listen tcp: address 65536: invalid port_»)
* `agoric start --reset local-chain 12345678` (way too high; ultimately ag-chain-cosmos fails with «_Error: invalid address (12345677): strconv.ParseUint: parsing "12345677": value out of range_»)
* `agoric start --reset local-chain 0` (too low; ultimately ag-chain-cosmos fails with «_Error: invalid address (-1): strconv.ParseUint: parsing "-1": invalid syntax_»)
* `agoric start --reset local-chain 17566` (apparently too low; ultimately ag-chain-cosmos fails with «_Error: listen tcp: address -1: invalid port_» [presumably because of [`finishCosmosApp`](https://github.com/Agoric/agoric-sdk/blob/3d0d4773bbb0824ffe6f06b5f32162e66911bf23/packages/agoric-cli/src/chain-config.js#L92)])
* `agoric start --reset local-chain 1234.5678` (non-integer; ultimately ag-chain-cosmos fails with «_Error: invalid address (1233.5678): strconv.ParseUint: parsing "1233.5678": invalid syntax_»)

## Expected behavior
Invalid input is rejected quickly, clearly, and without making changes to the filesystem.

gibson042:
closes: #5145

## Description

Make `agoric start` fail quickly and cleanly when a nonsensical port is specified.

### Security Considerations

None AFAICT.

We should probably define a valid port range. This PR rejects port 0 but not port 1, even though ports up to 17566 seem to result in failure of ag-chain-cosmos, at least without other configuration (presumably because of [`finishCosmosApp`](https://github.com/Agoric/agoric-sdk/blob/3d0d4773bbb0824ffe6f06b5f32162e66911bf23/packages/agoric-cli/src/chain-config.js#L92)).

### Testing Considerations

I don't see any existing tests that could be extended to cover this, and am inclined to leave things in that state but would be willing to add them with a little guidance (e.g., references to analogs).


packages/agoric-cli/src/chain-config.js
```diff
@@ -98,7 +98,8 @@ export function finishCosmosApp({
   app['pruning-keep-every'] = '50000';
   app['pruning-interval'] = '1000';
 
-  const apiPort = DEFAULT_API_PORT + (rpcPort - DEFAULT_RPC_PORT) / 100;
+  const apiPort =
+    DEFAULT_API_PORT + Math.ceil((rpcPort - DEFAULT_RPC_PORT) / 100);
```
gibson042, line 102: I'm open to other ideas here, but it's important to avoid generating a non-integer port number.

michaelfig: This is good.  I have no other suggestions.

michaelfig (approved):
LGTM!  Thanks.

PR #5148, opened 2022-04-19
style(run-protocol): @jessie-check for runStake

dckc:
refs: #3788

Zoe API design question: The only part of this code that wasn't already in Jessie was an attempt to have a single-source of truth for keywords in the contract: `proposal.give[KW.Attestation]`. `proposal.give.Attestation` looks nicer, but it's a use that is not checked against any defining occurrence.

### DIFF
--- a/packages/run-protocol/src/runStake/runStakeKit.js
+++ b/packages/run-protocol/src/runStake/runStakeKit.js
@@ -1,4 +1,5 @@
 // @ts-check
+// @jessie-check
 import { Far } from '@endo/far';
 import { AmountMath, AssetKind } from '@agoric/ertp';
 import { assertProposalShape } from '@agoric/zoe/src/contractSupport/index.js';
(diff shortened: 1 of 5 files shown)


dckc:
I notice the reserve contract has `collateralSeat.getCurrentAllocation()[collateralKeyword]`.

cc @Chris-Hibbert

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -212,17 +213,17 @@ export const makeRunStakeKit = (zcf, startSeat, manager) => {
...
 
     // new = after the transaction gets applied
     const newCollateral = addSubtract(collateral, giveColl, wantColl);
     // max debt supported by current Collateral as modified by proposal
     const { amountLiened, maxDebt: newMaxDebt } =
       manager.maxDebtForLien(newCollateral);
 
-    const giveRUN = AmountMath.min(proposal.give[KW.Debt] || emptyDebt, debt);
-    const wantRUN = proposal.want[KW.Debt] || emptyDebt;
+    const giveRUN = AmountMath.min(proposal.give.Debt || emptyDebt, debt);
```
turadg, line 225: LGTM.

> proposal.give.Attestation looks nicer, but it's a use that is not checked against any defining occurrence.

So now someone could type `proposal.give.Debt` and get no error, whereas `proposal.give[KW.Debte]` would have said `KW` doesn't have that key. But in both cases `give` doesn't constrain the keys.

I think where we want to get to is that you can validate the proposal with pattern matching so that the type system has a guarantee from some runtime validation as to what's in `give`. So imo let's work towards that and not invest in the stop-gap.

turadg (approved):

PR #5149, opened 2022-04-19
fix(run-protocol): vault debt ratio ordering after interest charges

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/4527 made it so that `orderedVaultStore` would maintain ordering by actual debt-to-collateral even as interest was charged. That required ordering by the normalized debt that was stable as interest charges accrurred.

There was a bug in `orderedVaultStore`'s `addVault` that keyed on the debt-to-collateral at the time of addition. This fixes that bug and also the tests that should have caught it.

### Security Considerations

--

--

### Testing Considerations

New tests with higher fidelity fake vaults.


dtribble (approved):
LGTM

PR #5150, opened 2022-04-19
chore(vat-data): improve type definitions

turadg:
## Description

- Adds a typedef for the single-faceted `defineKind`
- Adds an opaque type for durable handles
- Correctly types the durable kind
- Cleans up some dead code

### Security Considerations

--

--

### Testing Considerations

New types tests.


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -258,7 +258,6 @@ const makeVaultDirector = (zcf, electorateParamManager, debtMint) => {
         zcf.getTerms().priceAuthority,
         factoryPowers,
         timerService,
-        // @ts-expect-error
```
turadg, line 261: this edit in `run-protocol` triggered the mandatory code review by @dtribble

dtribble (commented):
The RUN protocol part looks fine.

packages/vat-data/src/types.d.ts
```diff
@@ -5,7 +5,6 @@ import type {
   WeakMapStore,
   WeakSetStore,
 } from '@agoric/store';
-import { Context } from 'vm';
```
mhofman, line 8: How had that snuck in?

packages/vat-data/src/types.d.ts
```diff
@@ -23,24 +22,44 @@ type ActualBehavior<B> = {
   [Facet in keyof B]: FunctionsMinusContext<B[Facet]>;
 };
 
-type KindContext<S, B> = { state: S; facets: ActualBehavior<B> };
+type MultiKindContext<S, B> = { state: S; facets: ActualBehavior<B> };
 
-interface KindDefiner {
-  <P, S, B>(
+declare const kindHandleMarker: unique symbol;
+export type DurableKindHandle = { readonly [kindHandleMarker]: string };
```
mhofman: Since the handle is a Far object, can we just make this a class with private?

```suggestion
declare class DefineKindHandleClass {
    private name: string;
}
export type DefineKindHandle = DefineKindHandleClass;
```

packages/vat-data/src/types.d.ts
```diff
@@ -23,24 +22,44 @@ type ActualBehavior<B> = {
...
+
+export type VatData = {
+  // virtual kinds
+  defineKind: <P, S, F>(
+    tag: string,
+    init: (...args: P) => S,
+    facet: F,
+    options?: {
+      finish?: (context: { state: S }, kind: FunctionsMinusContext<F>) => void;
```
mhofman: Not a type review, but a question on the API itself for @FUDCo. Any reason the `kind` is in is own arg, and we're not using the `self` of the `context`?

packages/vat-data/src/types.d.ts
```diff
@@ -23,24 +22,44 @@ type ActualBehavior<B> = {
   [Facet in keyof B]: FunctionsMinusContext<B[Facet]>;
 };
 
-type KindContext<S, B> = { state: S; facets: ActualBehavior<B> };
+type MultiKindContext<S, B> = { state: S; facets: ActualBehavior<B> };
```
mhofman: I find it weird that for single facet we name the helper `FunctionsMinusContext` but for multi, it's called `ActualBehavior`. In my mind it's `FacetFromBehavior` and `MultiFacetFromBehaviors` or something like that.

mhofman (dismissed):
Could we try a class with private approach instead of the unique symbol approach for kindHandle's type?

Besides that, all nits

packages/vat-data/src/types.d.ts
turadg: Yes, but why?

packages/vat-data/src/types.d.ts
turadg: That's out of scope of this change. Are you blocking this change on me renaming the extant types?

packages/vat-data/src/types.d.ts
turadg: https://github.com/Agoric/agoric-sdk/pull/4962/files#diff-4d09d78ca33d43e2584cfbf5adc00ed1caac51f1edbc10bae3d3c14320e56d30R8

mhofman (approved):

packages/vat-data/src/types.d.ts
mhofman: As we discussed off-line, class with privates are real nominal types in TS, where a unique symbol, even not directly exported, leaks in the types through `keyof DefineKindHandle`

PR #5126, opened 2022-04-16
test(governance): fix some async validation

turadg:
I ran into some mysteries debugging https://github.com/Agoric/agoric-sdk/pull/5116 and here's the result.

PR #5151, opened 2022-04-19
fix: reconcile use of path to paramManager vaults with others

Chris-Hibbert:
## Description

Vaults specify the path to the paramManagerRetriever slightly
differently from "standard" contracts. This makes parameter governance
support both approaches polymorphically.

Parameter Governance shouldn't know the specifics of what VaultManager
is doing, but it must accomodate those differences.

### Security Considerations

None

When we write up governance, this is an extra detail to be included in what contracts that don't follow the standard conventions need to do.

### Testing Considerations

Existing tests were sufficient.


packages/governance/src/contractGovernance/governParam.js
```diff
@@ -58,20 +58,28 @@ const validateParamChangeQuestion = details => {
   );
 };
 
-/** @type {AssertBallotConcernsParam} */
+/**
+ * assert that the parameter described by paramSpec is proposed to be changed in
+ * the question described by questionSpec.
+ *
+ * @param {any} paramSpec
+ * @param {QuestionSpec<ParamChangeIssue<any>>} questionSpec
```
turadg: all `ParamChangeIssue<any>` should be `ParamChangeIssue<unknown>`

packages/governance/src/question.js
```diff
@@ -91,7 +91,7 @@ const assertApiInvocation = issue => {
 // eslint-disable-next-line jsdoc/require-returns-check
 /**
  * @param {ElectionType} electionType
- * @param {unknown} issue
+ * @param {any} issue
```
turadg: why this change?

packages/governance/src/types.js
```diff
@@ -321,9 +322,9 @@
  */
 
 /**
+ * @template P path for a paramManagerRetriever
```
turadg: you can avoid edits to all the existing uses of `ParamChangeIssue` by giving the template a default value.
```suggestion
 * @template [P=unknown] path for a paramManagerRetriever
```

turadg (approved):
LGTM except for the `any` type erasure. Please address before merging.

Feel free to incorporate https://github.com/Agoric/agoric-sdk/pull/5126 or leave it for me to rebase onto master after this is in.

I'll rebase https://github.com/Agoric/agoric-sdk/pull/5116 too.

PR #5159, opened 2022-04-19
fix: repair types left as <any>

Chris-Hibbert:
refs: #5151

## Description

5151 inadvertently declared some types as any. This replaces those with better choices.

### Security Considerations

None

None

### Testing Considerations

None


turadg (approved):
🙏

PR #5160, opened 2022-04-19
feat(ertp): more AmountMath helpers

turadg:
## Description

I needed `isGT` and I thought it would be useful to have in `AmountMath`. I wondered with `priceAuthority.js` had defined them there instead of AmountMath. By trying to implement, I learned a little more about non-Nat amounts. Is it feasible to have generic definitions for all amount kinds?

If not, would it be okay to have a NatMath utility that works with `Amount<'nat'>`?

### DIFF
--- a/packages/ERTP/src/amountMath.js
+++ b/packages/ERTP/src/amountMath.js
@@ -346,6 +346,24 @@ const AmountMath = {
     return !AmountMath.isGTE(leftAmount, rightAmount, brand);
   },
   /**
+   * Returns true if the leftAmount is less than or equal to the
+   * rightAmount.
+   *
+   * For non-scalars, "greater than or equal to" depends on the kind of amount,
+   * as defined by the MathHelpers. For example, whether rectangle A is greater
+   * than rectangle B depends on whether rectangle A includes rectangle B as
+   * defined by the logic in MathHelpers.
+   *
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
+   * @param {Brand<K>} [brand]
+   * @returns {boolean}
+   */
+  isLTE: (leftAmount, rightAmount, brand) => {
+    return AmountMath.isGTE(rightAmount, leftAmount, brand);
+  },
+  /**
    * Returns true if the leftAmount equals the rightAmount. We assume
    * that if isGTE is true in both directions, isEqual is also true
    *
--- a/packages/ERTP/src/amountMath.js
+++ b/packages/ERTP/src/amountMath.js
@@ -291,11 +291,31 @@ const AmountMath = {
     return h.doIsEmpty(h.doCoerce(value));
   },
   /**
+   * Returns true if the leftAmount is greater than (and not equal to) the
+   * rightAmount.
+   *
+   * For non-scalars, "greater than or equal to" depends on the kind of amount,
+   * as defined by the MathHelpers. For example, whether rectangle A is greater
+   * than rectangle B depends on whether rectangle A includes rectangle B as
+   * defined by the logic in MathHelpers.
+   *
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
+   * @param {Brand<K>=} brand
+   * @returns {boolean}
+   */
+  isGT: (leftAmount, rightAmount, brand = undefined) => {
+    return !AmountMath.isGTE(rightAmount, leftAmount, brand);
+  },
+  /**
    * Returns true if the leftAmount is greater than or equal to the
-   * rightAmount. For non-scalars, "greater than or equal to" depends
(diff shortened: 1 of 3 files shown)


packages/ERTP/src/amountMath.js
```diff
@@ -149,7 +149,7 @@ const optionalBrandCheck = (allegedBrand, brand) => {
  * @param {Brand<K> | undefined} brand
  * @returns {MathHelpers<*>}
  */
-const checkLRAndGetHelpers = (leftAmount, rightAmount, brand = undefined) => {
+const checkLRAndGetHelpers = (leftAmount, rightAmount, brand) => {
```
erights, line 152: Why remove the `= undefined` ? I purposely write those because I find it makes the code clearer.

(elsewhere too)

packages/ERTP/src/amountMath.js
```diff
@@ -282,42 +282,98 @@ const AmountMath = {
    * @param {Brand=} brand
```
erights, line 282: Some places (like here) you're leaving the `=} name` in place to indicate optional. Other places you're replacing it with `} [name]`. Did we ever decide which we prefer? Assuming the two are semantically identical, I'm happy either way, but we should be consistent.

packages/ERTP/src/amountMath.js
```diff
@@ -282,42 +282,98 @@ const AmountMath = {
...
+   *
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
+   * @param {Brand<K>} [brand]
+   * @returns {boolean}
+   */
+  isGT: (leftAmount, rightAmount, brand) => {
+    return !AmountMath.isGTE(rightAmount, leftAmount, brand);
+  },
```
erights, line 310: This definition is wrong for partial orders. Instead
```suggestion
  isGT: (leftAmount, rightAmount, brand = undefined) => {
    return AmountMath.isGTE(leftAmount, rightAmount, brand) && !AmountMath.isEqual(leftAmount, rightAmount);
  },
```
which directly mirrors your prose definition above.

packages/ERTP/src/amountMath.js
```diff
@@ -282,42 +282,98 @@ const AmountMath = {
...
+   * For non-scalars, "greater than or equal to" depends on the kind of amount,
+   * as defined by the MathHelpers. For example, whether rectangle A is greater
+   * than rectangle B depends on whether rectangle A includes rectangle B as
+   * defined by the logic in MathHelpers.
    *
    * @template {AssetKind} [K=AssetKind]
    * @param {Amount<K>} leftAmount
    * @param {Amount<K>} rightAmount
-   * @param {Brand<K>=} brand
+   * @param {Brand<K>} [brand]
```
erights, line 323: This is the inconsistent change I refer to above.

packages/ERTP/src/amountMath.js
```diff
@@ -282,42 +282,98 @@ const AmountMath = {
...
+   *
+   * @template {AssetKind} [K=AssetKind]
+   * @param {Amount<K>} leftAmount
+   * @param {Amount<K>} rightAmount
+   * @param {Brand<K>} [brand]
+   * @returns {boolean}
+   */
+  isLT: (leftAmount, rightAmount, brand) => {
+    return !AmountMath.isGTE(leftAmount, rightAmount, brand);
+  },
```
erights, line 347:
```suggestion
  isLT: (leftAmount, rightAmount, brand = undefined) => {
    return AmountMath.isGT(rightAmount, leftAmount, brand);
  },
```

packages/ERTP/test/unitTests/test-amountProperties.js
```diff
@@ -2,6 +2,7 @@ import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
 import { makeCopyBag } from '@agoric/store';
 import fc from 'fast-check';
 
+import { q } from '@agoric/assert';
```
erights, line 5:
```suggestion
const { quote: q } = assert;
```

erights (changes requested):
The general issue is that amounts are partially ordered and your generic code assumes that they are fully ordered.

packages/ERTP/src/amountMath.js
turadg: I understand it to be documentation of the expected params because it has no runtime effect. For documentation of expected params, we have JSDoc so I think we should use it consistently.

packages/ERTP/src/amountMath.js
turadg: Good call-out. I don't think we resolved that thread. I'll ping you on it. My position is to not use `=}` which is the Google Closure syntax, to reduce the number of syntaxes to know.

erights: Were there any outstanding counter-arguments?

erights: Where is that thread?

PR #5161, opened 2022-04-19
refactor(run-protocol): DRY checkDebtLimit

turadg:
Extracted from https://github.com/Agoric/agoric-sdk/pull/5135

## Description

- factor out `checkDebtLimit` that was copy-pasted
- more idiomatic assert
- more accurate error message

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/packages/run-protocol/src/contractSupport.js
+++ b/packages/run-protocol/src/contractSupport.js
@@ -2,7 +2,7 @@
 import { AmountMath } from '@agoric/ertp';
 import { M } from '@agoric/store';
 
-const { details: X } = assert;
+const { details: X, quote: q } = assert;
 
 export const amountPattern = harden({ brand: M.remotable(), value: M.any() });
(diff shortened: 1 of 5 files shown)


packages/run-protocol/src/contractSupport.js
```diff
@@ -1,4 +1,5 @@
 // @ts-check
+import { q } from '@agoric/assert';
```
dckc: I don't think I should approve code with this idiom. Your point that the deprecation isn't documented is well made, but I still think it's the position of several in the team, and as a reviewer, I'm obliged to represent them.

turadg: Deferring resolution, moved to chat

dtribble (approved):
LGTM

dckc (approved):
good stuff.

I gather @dtribble is fine with this too.

Issue #4339, opened 2022-01-21
Remove Prettier from eslint


turadg:
## What is the Problem Being Solved?
https://github.com/Agoric/agoric-sdk/pull/4330#discussion_r789387409

## Description of the Design

* Remove prettier concern from eslint
 * anyone relying on the eslint autofix can enable formatOnSave with https://github.com/Agoric/agoric-sdk/blob/master/scripts/configure-vscode.sh
* Add prettier check to CI
* `yarn lint-fix` in each package should still fix formatting (cc @kriskowal @warner )

## Security Considerations
n/a

## Test Plan

- [ ] prettier problems don't appear in IDE
- [ ] pretty-check failures fail CI

dckc:
Note we rely on eslint for [@jessie-check](https://agoric.com/documentation/guides/js-programming/hardened-js.html#tool-support-eslint-config). Does this put that at risk? @michaelfig ?

turadg:
@michaelfig pointed out that some people have `eslint --fix` set up to run on save, and taking Prettier out of that would stop their auto-formatting. [Prettier has editor integrations](https://prettier.io/docs/en/editors.html#visual-studio-code), but the change would impact developers.

So https://github.com/Agoric/agoric-sdk/pull/4342/ just downgrades the error to a warning, which is an improvement:
<img width="642" alt="Screen Shot 2022-01-21 at 3 28 24 PM" src="https://user-images.githubusercontent.com/21505/150616117-cdb7c6dc-ecef-4b64-a9af-c8c3ac671109.png">

TBD whether we want to remove it entirely.

michaelfig:
> Note we rely on eslint for [@jessie-check](https://agoric.com/documentation/guides/js-programming/hardened-js.html#tool-support-eslint-config). Does this put that at risk?

No, we're good.  Eslint still runs from VSCode and CI, it's just making `eslint --fix` optional that we're talking about.

erights:
This seems a relevant enough place to mention the anomalies seen at https://github.com/endojs/endo/pull/973#discussion_r790068936

I do not understand what I'm seeing there.

mhofman:
Any reason people can't update their IDEs to run both eslint and prettier fix on save? VSCode should do prettier by default. I actually don't like having eslint fix on by default as that's usually more invasive changes, but I understand some may want to have it. The code hint overload can be real between all these tools.

Btw, do we know if the eslint-prettier plugin respects the `// prettier-ignore` instruction?

[...]

PR #5162, opened 2022-04-20
ci: eslint and fix .ts files

turadg:
## Description

Each package has its own lint glob but formatting (with Prettier) is configured repo-wide. When a developer runs `yarn lint` in a package, it may not lint the same files that the formatting check covers.

Eventually I think we should separate auto-formatting concerns from linting (https://github.com/Agoric/agoric-sdk/issues/4339 ) but until then the `yarn lint` command should cover the same files that CI reports in the "lint" job.

This changes the per-package config to include .ts files, using `ext` instead of a glob.

It also makes `lint-fix` re-use the `lint:eslint` command to ensure it always fixes the same set of files.

I tried to make `.ts` inclusion be at the repo level but can't due to https://github.com/eslint/eslint/issues/11223

### Security Considerations

--

--

### Testing Considerations

- [x] introduce a prettier deviation into a .ts file and confirm that `lint-fix` corrects it

### DIFF
--- a/golang/cosmos/package.json
+++ b/golang/cosmos/package.json
@@ -12,7 +12,7 @@
     "build:gyp-debug": "make compile-gyp GYP_DEBUG=--debug",
     "test:xs": "exit 0",
     "build": "exit 0",
-    "lint-fix": "eslint --fix '**/*.{cjs,js}'",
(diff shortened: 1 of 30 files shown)


FUDCo (approved):
Both astonishing and gratifying.

PR #5163, opened 2022-04-20
fix: hotfix for @endo/init#1166

mhofman:
Integrate https://github.com/endojs/endo/pull/1166 as a patch so that we don't need to wait for a full release crank.


kriskowal (approved):
Could use a reference in code to the issue and what to change here after the release is integrated.

mhofman:
> what to change here after the release is integrated

It should just be to regenerate patch files for endo packages, which would result in removal of this patch.
Maybe automatic regeneration of patches should be part of the version sync process?

PR #5164, opened 2022-04-20
feat(swingset-vat): Include alleged bundleID in checkBundle diagnostics

kriskowal:
## Description

Recent bundle caching introduced a class of errors that is difficult to inspect.
This change threads the bundle ID through to checkBundle so it can be included in error messages.
Further work in checkBundle finishes threading the diagnostic bundle name through to the error message in question.

See: https://github.com/endojs/endo/pull/1167

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -318,7 +318,7 @@ export async function makeSwingsetController(
    * Validate and install a code bundle.
    *
    * @param { EndoZipBase64Bundle } bundle
-   * @param { BundleID? } allegedBundleID
+   * @param { BundleID } [allegedBundleID]
    * @returns { Promise<BundleID> }
    */
   async function validateAndInstallBundle(bundle, allegedBundleID) {
@@ -335,7 +335,7 @@ export async function makeSwingsetController(
       ),
       `Bundle with alleged ID ${allegedBundleID} must be a frozen object with only string value properties, no accessors`,
     );
-    await checkBundle(bundle, computeSha512);
+    await checkBundle(bundle, computeSha512, allegedBundleID);
     const { endoZipBase64Sha512 } = bundle;
     assert.typeof(endoZipBase64Sha512, 'string');
     const bundleID = `b1-${endoZipBase64Sha512}`;


packages/SwingSet/src/controller/controller.js
```diff
@@ -318,7 +318,7 @@ export async function makeSwingsetController(
    * Validate and install a code bundle.
    *
    * @param { EndoZipBase64Bundle } bundle
-   * @param { BundleID? } allegedBundleID
+   * @param { BundleID= } allegedBundleID
```
warner: What does `=` mean?

If it means "must always be provided" (which I assume it doesn't, because then we'd just remove the punctuation entirely), then this function becomes slightly harder to use, because a small host app that wants to install+reference some dynamic vat bundles must also figure out their bundle IDs ahead of time, instead of learning them from the return value of `validateAndInstallBundle`. Not a big deal, and probably less important than the diagnostic value I believe this supports.

kriskowal: It’s equivalent to `{BundleID | undefined} allegedBundleID` or `{BundleID} [allegedBundleID]` whereas `{BundleID?} allegedBundleID` is equivalent to `{BundleID | null} allegedBundleID`. We may or may not have resolved to use brackets like `{BundleID} [allegedBundleID]` for this instead cc @turadg @dtribble. I saw a weakly held majority opinion but not unanimous consensus.

turadg: Pretty sure we landed on eschewing Google Closure syntax:
```suggestion
   * @param { BundleID } [allegedBundleID]
```

Incidentally, https://github.com/hosseinmd/prettier-plugin-jsdoc we may adopt (https://github.com/Agoric/agoric-sdk/pull/4903) makes the braces tight around types `{BundleID}` which helps readability because it maintains spaces for objects. e.g.,
```
   * @param {FooType} foo
   * @param {{ name: string }} nameRecord
   * @param {FooType & { name: string }} weirdThing
```

kriskowal: Sounds good to me.

warner (approved):
aye, looks good

PR #5168, opened 2022-04-20
refactor(run-protocol): RunStakeKit facets for vobj

turadg:
part of https://github.com/Agoric/agoric-sdk/issues/4534

## Description

Refactors `runStakeKit.js` to support virtual objects. Principally:
- initial state is constructed in a single `initState`
- facets to be exposed on the vobj are named (`helper` and `pot` for what holds the stake, fka `vault` which has another meaning)
- offer result concern is moved to the contract

Probably easiest to review by commit. I'll try avoid rebasing and instead make add fixup commits.

### Security Considerations

--

"Pot" term, which may not last long. I thought it better than "vault" even for now.

### Testing Considerations

Existing tests since purely refactor.

### DIFF
--- a/packages/run-protocol/src/runStake/runStakeKit.js
+++ b/packages/run-protocol/src/runStake/runStakeKit.js
@@ -71,45 +71,12 @@ const initState = (zcf, startSeat, manager) => {
     collateralBrand,
     AssetKind.COPY_BAG,
   );
-  const emptyDebt = AmountMath.makeEmpty(debtBrand);
 
   const { zcfSeat: vaultSeat } = zcf.makeEmptySeatKit();
-  const { notifier, updater } = makeNotifierKit();
 
-  /** @type {ImmutableState} */
-  const fixed = {
-    collateralBrand,
-    debtBrand,
-    emptyCollateral,
-    emptyDebt,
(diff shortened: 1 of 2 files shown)


packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -21,16 +21,32 @@ export const KW = /** @type { const } */ ({
 });
 
 /**
- * Make RUNstake kit, subject to runStake terms.
+ * @typedef {Readonly<{
+ *   collateralBrand: Brand,
+ *   debtBrand: Brand,
+ *   emptyCollateral: Amount<'copyBag'>,
```
dckc: probably redundant. not much harm done.

dckc (approved):
I like it.

the use-before-define suppression should be local

packages/run-protocol/src/runStake/runStakeKit.js
turadg: reminded me we don't need to keep the empties in state https://github.com/Agoric/agoric-sdk/pull/5168/commits/f4e60de2b7321bf86ca8c309e9731f9782761fda

PR #5171, opened 2022-04-21
refactor(run-protocol): convert runStakeKit to vobj

### LINKED ISSUE #4534: runStake: use virtual objects for high-cardinality state

turadg:
closes: #4534

## Description

RunStakeKit is the high cardinality object of RUNstake. This converts it to being a virtual object, completing #4534.

It also fixes part of the typedef for PlusContext. I'll do a merge commit to separate them in master.

### Security Considerations

State can be paged out and in. Relies on security of vobj system.

--

### Testing Considerations

Refactor. Existing tests suffice. Durability out of scope.

### DIFF
--- a/packages/run-protocol/src/runStake/params.js
+++ b/packages/run-protocol/src/runStake/params.js
@@ -2,19 +2,7 @@
 // @jessie-check
 
 import { CONTRACT_ELECTORATE, ParamTypes } from '@agoric/governance';
-import { KW as AttKW } from './attestation.js';
-
-export const KW = /** @type { const } */ ({
-  [AttKW.Attestation]: AttKW.Attestation,
-  Debt: 'Debt',
-});
-
-const PKey = /** @type { const } */ ({
-  DebtLimit: 'DebtLimit',
-  MintingRatio: 'MintingRatio',
-  InterestRate: 'InterestRate',
-  LoanFee: 'LoanFee',
-});
+import { ParamKW as PKey } from './constants.js';
 
 export const makeRunStakeTerms = (
   { timerService, chargingPeriod, recordingPeriod },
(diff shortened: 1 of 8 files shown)


packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -128,6 +136,292 @@ const initState = (zcf, startSeat, manager) => {
...
+  /**
+   * @param {MethodContext} context
+   * @param {ZCFSeat} seat
+   */
+  getRunAllocated: ({ state }, seat) =>
+    seat.getAmountAllocated(KW.Debt, state.debtBrand),
+  /**
+   * @param {MethodContext} context
+   */
+  getCollateralAmount: ({ state, facets }) => {
```
dckc, line 150: What about:

```suggestion
  getCollateralAmount: ({ state, { helper }) => {
```

so we can leave `helper.getCollateralAllocated(...)` alone?

dckc: or `const { helper } = facets`?

dckc (approved):
I reviewed this somewhat lightly. Looks OK.
One minor suggestion; no need to act on it.

packages/run-protocol/src/runStake/runStakeKit.js
turadg: I went with the latter. The double destructuring leads causes a circular type definition.

packages/run-protocol/src/runStake/params.js
```diff
@@ -2,10 +2,12 @@
 // @jessie-check
 
 import { CONTRACT_ELECTORATE, ParamTypes } from '@agoric/governance';
+import { KW as AttKW } from './attestation.js';
```
dckc: won't this cause the same problem?

turadg: yeah would in next PR so fixed with [c73f6f2](https://github.com/Agoric/agoric-sdk/pull/5171/commits/c73f6f2d7e0e83c5ab94db47412df74f8e4a5e6c)

PR #5173, opened 2022-04-21
refactor(run-protocol): vault title vobj

### LINKED ISSUE #4345: Use Virtual Objects in Vaults

turadg:
closes: #4345 

## Description

One last stateful object in vaultFactory.

This is the "outer vault" that end users hold, allowing the "inner vault" that has the assets to be transferred. In the interests of clarity this renames the "outer" entity to "title", since it's a record of ownership of the vault. "Vault" can then be referred to consistently without a qualifier.

In the UI there need not be any distinction. Users will think of a vault and transferring the vault.

### Security Considerations

--

Outer/inner wasn't documented so nothing to change. Not user facing.

### Testing Considerations

Refactor so existing tests.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultKit.js
+++ b/packages/run-protocol/src/vaultFactory/vaultKit.js
@@ -1,62 +1,7 @@
 // @ts-check
-import { makeNotifierKit } from '@agoric/notifier';
 import '@agoric/zoe/exported.js';
 import { Far } from '@endo/marshal';
-
-const { details: X } = assert;
-
-/**
- * @typedef {{
- * inner: InnerVault | null,
- * }} State
- */
-/**
- *
- * @param {InnerVault} innerVault
- */
-const wrapVault = innerVault => {
-  /** @type {NotifierRecord<VaultUIState>} */
-  const { updater, notifier } = makeNotifierKit();
-
-  /** @type {State} */
-  const state = {
-    inner: innerVault,
-  };
-
-  // Throw if this wrapper no longer owns the inner vault
-  const owned = v => {
-    const { inner } = state;
-    // console.log('OUTER', v, 'INNER', inner);
-    assert(inner, X`Using ${v} after transfer`);
-    return inner;
-  };
-
-  /**
-   * Public API of the vault.
-   *
-   * @see {InnerVault} for the internal API it wraps.
-   */
-  const vault = Far('vault', {
-    getNotifier: () => notifier,
-    makeAdjustBalancesInvitation: () =>
-      owned(vault).makeAdjustBalancesInvitation(),
-    makeCloseInvitation: () => owned(vault).makeCloseInvitation(),
-    /**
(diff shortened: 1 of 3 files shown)


packages/run-protocol/src/vaultFactory/vaultTitle.js
```diff
@@ -0,0 +1,88 @@
...
+  getNotifier: ({ state }) => state.notifier,
+  /** @param {MethodContext} context */
+  makeAdjustBalancesInvitation: ({ facets }) =>
+    facets.helper.owned().makeAdjustBalancesInvitation(),
+  /** @param {MethodContext} context */
+  makeCloseInvitation: ({ facets }) =>
+    facets.helper.owned().makeCloseInvitation(),
+  /**
+   * Starting a transfer revokes the outer vault. The associated updater will
+   * get a special notification the the vault is being transferred.
```
Chris-Hibbert, line 63:
```suggestion
   * Starting a transfer revokes the vaultTitle. The associated updater will
   * get a special notification that the vault is being transferred.
```

packages/run-protocol/src/vaultFactory/vaultTitle.js
```diff
@@ -0,0 +1,88 @@
...
+ *   facets: {
+ *     helper: import('@agoric/vat-data/src/types').KindFacet<typeof helper>,
+ *     self: import('@agoric/vat-data/src/types').KindFacet<typeof title>,
+ *   },
+ * }>} MethodContext
+ */
+
+/**
+ *
+ * @param {InnerVault} vault
```
Chris-Hibbert, line 28: I thought part of the point was that `InnerVault` could become just `Vault`.

packages/run-protocol/src/vaultFactory/vaultKit.js
```diff
@@ -1,62 +1,7 @@
...
-    getCollateralAmount: () => owned(vault).getCollateralAmount(),
-    getCurrentDebt: () => owned(vault).getCurrentDebt(),
-    getNormalizedDebt: () => owned(vault).getNormalizedDebt(),
-  });
-  return { vault, vaultUpdater: updater };
-};
+import { makeVaultTitle } from './vaultTitle';
 
 /**
  * Create a kit of utilities for use of the (inner) vault.
```
Chris-Hibbert, line 7:
```suggestion
 * Create a kit of utilities for use of the vault.
 *
 * @param {Vault} vault
```

Chris-Hibbert (approved):
Seems fine. I thought part of the point (" 'Vault' can then be referred to consistently without a qualifier.") was to shorten the name of the innerVault.

turadg:
> I thought part of the point (" 'Vault' can then be referred to consistently without a qualifier.") was to shorten the name of the innerVault.

Yes. This enables that. It's separate work from the vobj scope of this PR so I'll do a fast follow.

PR #5177, opened 2022-04-21
chore: clean up eslintignores

turadg:
## Description

The `packages/SwingSet/test/bundling/bundle-disk.js` left by a test is erroring in linter locally. This makes that file be stored in the `os.tmpdir()` so it's not inspected by any repo tooling.

The per-package `.eslintignore` is a maintenance cost and extra work to keep in sync with `.gitignore` and `.prettierignore`. There are [tools to help](https://github.com/mysticatea/eslint-gitignore) but if we can eliminate the complexity the better. It's also not strictly a match because some files are meant to be kept in SCM but not linted.

So this also removes empty `.eslintignore` and moves file-specific exceptions to a pragma in the file.

The only `.eslintignore` that's left is in cosmic-swingset, which is different enough from the rest that I wasn't confident modifying.

### Security Considerations

--

One less thing to keep track of.

### Testing Considerations

- [x] for each touched package, run `yarn build && yarn test && yarn lint:eslint` to make sure it passes

### DIFF
--- a/packages/SwingSet/test/bundling/test-bundles.js
+++ b/packages/SwingSet/test/bundling/test-bundles.js
@@ -3,8 +3,6 @@ import { test } from '../../tools/prepare-test-env-ava.js';
 
 // eslint-disable-next-line import/order
 import fs from 'fs';
-import path from 'path';
-import os from 'os';
 import bundleSource from '@endo/bundle-source';
 import { assert } from '@agoric/assert';
 import { parse } from '@endo/marshal';
@@ -14,10 +12,6 @@ import { capargs } from '../util.js';
 
 function bfile(name) {
   return new URL(name, import.meta.url).pathname;
-}
-
-function tmpfile(name) {
-  return path.join(os.tmpdir(), name);
 }
 
 // bundle IDs are hex SHA512, so this must be 128 characters long (plus the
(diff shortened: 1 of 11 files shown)


mhofman (commented):
Main concern here is that we'll just put the junk in a place where it'll accumulate. We should try to cleanup after ourselves as much as possible.

This is one of those cases where I think the rename uncovered a flow in the original test setup. Like @FUDCo said, the junk should never have stayed around in the first place.

I'm happy to go fix the test if you want.

turadg:
> the junk should never have stayed around in the first place.

yeah, maybe so. I figured it wasn't deleted by the test to make it easier to debug, but that would be easy enough to do by stepping through with a debugger or just commenting out the deletion line. So I think it's safe to delete and that commit would be welcomed.

packages/SwingSet/test/bundling/test-bundles.js
```diff
@@ -14,6 +16,10 @@ function bfile(name) {
   return new URL(name, import.meta.url).pathname;
 }
 
+function tmpfile(name) {
```
warner: Writing to a shared tmpdir enables a symlink-based confused-deputy attack (not like we're using shared machines much, but still best to avoid). Could you take a look at my abandoned PR #5180 and maybe use that approach? Keep the file in the tree, but use `t.teardown` to delete it afterwards, and also name it something not lintable like `bundle-disk.bundle`.

warner (commented):
rest looks good to me

warner (approved):
I added one commit to clean up the file with `t.teardown`. Looks good.

PR #5178, opened 2022-04-21
ci(docker): parallel multiarch manifests

michaelfig:
## Description

Properly implement Docker CI builds with my two favourite features:
- fast path; 11 minute build for amd64 setup images
- merge separate architectures under the same tagged manifest as they become available

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,7 +8,21 @@ on:
     workflow_dispatch:
 
 jobs:
+  snapshot:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: '${{ steps.snapshot-tag.outputs.tag }}'
+    steps:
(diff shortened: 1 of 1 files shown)


arirubinstein (approved):
bye `awk` 👋

.github/workflows/docker.yml
```diff
@@ -8,7 +8,21 @@ on:
...
 jobs:
+  snapshot:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: '${{ steps.snapshot-tag.outputs.tag }}'
+    steps:
+      - name: Generate Snapshot Tag
+        id: snapshot-tag
+        run: |
+          TIMESTAMP=`date +%Y%m%d%H%M%S`
```
mhofman, line 19: Should we use the commit date instead?

PR #5179, opened 2022-04-21
refactor(run-protocol): consistence invitation patterns

turadg:
## Description

We have a lot of different ways to name the parts of an invitation. More consistency would reduce effort in code review and navigation.

### DIFF
--- a/packages/run-protocol/src/psm/psm.js
+++ b/packages/run-protocol/src/psm/psm.js
@@ -146,7 +146,8 @@ export const start = async (zcf, privateArgs) => {
     }
   };
 
-  const swapHook = seat => {
+  /** @param {ZCFSeat} seat */
+  const handleSwapOffer = seat => {
     assertProposalShape(seat, {
       give: { In: null },
     });
@@ -163,7 +164,7 @@ export const start = async (zcf, privateArgs) => {
     }
     seat.exit();
   };
-  const makeSwapInvitation = () => zcf.makeInvitation(swapHook, 'swap');
+  const makeSwapInvitation = () => zcf.makeInvitation(handleSwapOffer, 'swap');
 
   const publicFacet = Far('Parity Stability Module', {
     getPoolBalance: () => anchorPool.getAmountAllocated('Anchor', anchorBrand),
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -74,7 +74,7 @@ const start = async (zcf, privateArgs) => {
   };
 
   // Anyone can deposit collateral to the reserve.
-  const addCollateralHook = async seat => {
+  const handleAddCollateralOffer = async seat => {
     const {
       give: { Collateral: amountIn },
     } = seat.getProposal();
@@ -90,7 +90,7 @@ const start = async (zcf, privateArgs) => {
   };
 
   const makeAddCollateralInvitation = () =>
-    zcf.makeInvitation(addCollateralHook, 'Add Collateal');
+    zcf.makeInvitation(handleAddCollateralOffer, 'Add Collateal');
 
   /** @type {ZCFMint} */
   const runMint = await zcf.registerFeeMint('RUN', feeMintAccess);
--- a/packages/run-protocol/src/runStake/runStake.js
+++ b/packages/run-protocol/src/runStake/runStake.js
@@ -171,7 +171,7 @@ export const start = async (
   /**
    * @param {ZCFSeat} seat
    */
-  const offerHandler = seat => {
+  const handleLoanOffer = seat => {
     const { helper, pot } = makeRunStakeKit(zcf, seat, manager);
 
     return harden({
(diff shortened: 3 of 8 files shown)


packages/run-protocol/src/psm/psm.js
```diff
@@ -163,7 +164,7 @@ export const start = async (zcf, privateArgs) => {
     }
     seat.exit();
   };
-  const makeSwapInvitation = () => zcf.makeInvitation(swapHook, 'swap');
+  const makeSwapInvitation = () => zcf.makeInvitation(handleSwapOffer, 'swap');
```
erights, line 167: What's your suggested casing rule? I would have expected either
   * `'CollectFees'` and `'Swap`', or
   * `'collectFees'` and `'swap'`

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -90,7 +90,7 @@ const start = async (zcf, privateArgs) => {
   };
 
   const makeAddCollateralInvitation = () =>
-    zcf.makeInvitation(addCollateralHook, 'Add Collateal');
+    zcf.makeInvitation(handleAddCollateralOffer, 'Add Collateal');
```
erights, line 93: I would have expected either
   * `'Collect Fees`' and `'Add Collateral'`, or
   * `'CollectFees'`, and `'AddCollateral'`

packages/run-protocol/src/runStake/runStake.js
```diff
@@ -191,7 +195,7 @@ export const start = async (
   const publicFacet = augmentPublicFacet(
     Far('runStake public', {
       makeLoanInvitation: () =>
-        zcf.makeInvitation(offerHandler, 'make RUNstake'),
+        zcf.makeInvitation(handleLoanOffer, 'make RUNstake'),
```
erights, line 198: Purposeful violation? If intentional, I'm not objecting. But worth at least a comment to explain why.

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -347,11 +347,14 @@ export const makeRunStakeKit = (zcf, startSeat, manager) => {
...
-      return zcf.makeInvitation(helper.adjustBalancesHook, 'AdjustBalances');
+      return zcf.makeInvitation(
+        helper.handleAdjustBalancesOffer,
+        'AdjustBalances',
+      );
     },
     makeCloseInvitation: () => {
       assert(state.open);
-      return zcf.makeInvitation(helper.closeHook, 'CloseVault');
+      return zcf.makeInvitation(helper.handleCloseOffer, 'CloseVault');
```
erights, line 357: Why not `handleCloseVaultOffer` for consistency?

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -24,43 +33,47 @@ const trace = makeTracer('LiqMin');
...
-   * @param {{ debt: Amount<'nat'> }} options
-   */
-  const debtorHook = async (debtorSeat, { debt }) => {
-    const debtBrand = debt.brand;
-    const {
-      give: { In: amountIn },
-    } = debtorSeat.getProposal();
+  const makeDebtorInvitation = curryInvitation(
+    zcf,
+    'Liquidate',
```
erights, line 38: ?

packages/run-protocol/src/psm/psm.js
turadg: I just neglected to update the string. I _think_ we're better off making string be exactly the chars in `make` … `Invitation` but I don't feel as strongly about it. I could see people wanting to put more descriptive names there and such uniqueness could actually help in a string search of the repo.

packages/run-protocol/src/runStake/runStake.js
turadg: Purposeful violation of the string? yeah, for comments above. 

I believe the method names are consistent. (`Loan` as key)

packages/run-protocol/src/runStake/runStakeKit.js
turadg: This is trying to be consistent with `makeCloseInvitation`. The key is `Close` so the two methods are:
- `makeCloseInvitation`
- `handleCloseOffer`.

Ah, you're probably referring to the string. Yeah I see that I was very inconsistent in that! Working through this, I think that can be more free-form, especially since it's loosely user-facing. We may want to refactor methods names and not have to change that string people are used ti finding in logs.

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
turadg: This is was in ce264a4 to show an option for not defining a method name at all. I think it reads well and doesn't require more indentation.

Chris-Hibbert:
I don't have strong opinions. I'm happy to let @erights discuss.

PR #5180, opened 2022-04-21
chore(swingset): delete temporary bundle after test

warner:


warner:
nevermind, I like your approach in #5177 better

Issue #5182, opened 2022-04-21
Consistent deployment test CI job


mhofman:
## What is the Problem Being Solved?

It seems like the `deployment-test` integration job can run inconsistently. I believe one reason my be because it pulls the `agoric/deployment:latest` docker image. Instead it should use the revision specific image uploaded by the `docker` job.

## Description of the Design

For that we may need to:
- update the CI jobs for `deployment-test` to depend on `docker`
- Update the `docker` job to generate deterministic tags (right now it uses the wall clock for a timestamp, it should use the commit time).

## Test Plan

We should make sure that the deployment's `intergation-test.sh` can still run locally. E.g. make the image configurable by env, and default to `latest`.

I'd also like if we could more visible show which reversion of the chain we end up running in the integration test.

mhofman:
I was mistaken and didn't understand that `agoric/deployment` image is stateless.

I did however changed the image tag generation to be deterministic, since that has other potential uses.

PR #5188, opened 2022-04-21
fix(ci): use a stable docker snapshot-tag

mhofman:
refs: #5182 

## Description

While it might not solve the issue, this should allow us to generate a well-known docker tag given a revision in case we ever need it later for any purpose.

### Security Considerations

None

### Testing Considerations

The manual trigger is busted (bad yaml indentation), but I manually tried this with a tag


michaelfig (approved):
LGTM!  Thanks for this.

PR #5190, opened 2022-04-21
Price feed fixes

michaelfig:
## Description

Some fixes for price feeds as well as drive-bys.  Reviewing commit-by-commit is probably the best way.


packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -40,14 +45,14 @@ const start = async zcf => {
 
   const unitIn = AmountMath.getValue(brandIn, unitAmountIn);
 
-  /** @type {IssuerRecord & { mint: ERef<Mint> }} */
-  let quoteKit;
```
dckc, line 44: good riddance!

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -13,6 +13,8 @@ import { launch } from './launch-chain.js';
 import makeBlockManager from './block-manager.js';
 import { getTelemetryProviders } from './kernel-stats.js';
 
+let whenHellFreezesOver = null;
```
dckc, line 16: lint gripes about unused variable; I presume you'll suppress it (and remove the setInterval global comment above)

dckc (approved):
good stuff.

packages/zoe/src/contracts/scaledPriceAuthority.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * A contract that scales a source price authority to a target price authority
+ * via ratios.
+ *
+ * @param {ZCF<{
+ *   sourcePriceAuthority: ERef<PriceAuthority>,
+ *   scaleIn: Ratio,
+ *   scaleOut: Ratio,
+ * }>} zcf
```
dckc: it's useful to name the type of the terms so that clients can use it

dckc (approved):
scaled price authority looks reasonable.

PR #5191, opened 2022-04-21
real governance for PSM

### LINKED ISSUE #5153: Add actual governance to PSM module

The PSM module [currently doesn't have](https://github.com/Agoric/agoric-sdk/blob/ba8b3073672f6b80d98b5b72c6538b9962196fd5/packages/run-protocol/src/psm/psm.js#L77) its parameters  managed under governance.

## Description of the Design

Add ParamGovernance.

## Security Considerations

Standard governance legibility concerns.

## Test Plan

Verify that the params can be changed by governance.
Notice that starting up the PSM will become more complex, since a Governor will be required.

turadg:
closes: #5153

## Description

Replaces fake governance on PSM with real.

Some other cleanup along the way. Will do a merge commit.

### Security Considerations

Governable parameters.

--

### Testing Considerations

Made tests set up governance.

### DIFF
--- a/packages/run-protocol/src/psm/psm.js
+++ b/packages/run-protocol/src/psm/psm.js
@@ -1,16 +1,15 @@
 // @ts-check
 import '@agoric/zoe/exported.js';
 import '@agoric/zoe/src/contracts/exported.js';
-import '@agoric/governance/src/exported';
+import '@agoric/governance/src/exported.js';
 import {
   assertProposalShape,
   ceilMultiplyBy,
   floorDivideBy,
   floorMultiplyBy,
-  makeRatio,
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { Far } from '@endo/marshal';
-// import { CONTRACT_ELECTORATE } from '@agoric/governance';
+import { handleParamGovernance, ParamTypes } from '@agoric/governance';
 
 import { AmountMath } from '@agoric/ertp';
 import { makeMakeCollectFeesInvitation } from '../collectFees.js';
@@ -25,8 +24,6 @@ const { details: X } = assert;
  * token are specified separately.
  *
  */
-
-const BASIS_POINTS = 10_000n;
 
 /**
  * Stage a transfer of a single asset from one seat to another, with an optional
@@ -44,20 +41,19 @@ function stageTransfer(from, to, txFrom, txTo = txFrom) {
 }
 
 /**
- * @param {ZCF<{
+ * @param {ZCF<GovernanceTerms<{
+ *    GiveStableFee: ParamRecord<'ratio'>,
+ *    WantStableFee: ParamRecord<'ratio'>,
+ *    MintLimit: ParamRecord<'amount'>,
+ *   }> & {
  *    anchorBrand: Brand,
  *    anchorPerStable: Ratio,
- *    governedParams: {
- *      WantStableFeeBP: bigint,
- *      GiveStableFeeBP: bigint,
- *      MintLimit: Amount } }>
- * } zcf
- * @param {{feeMintAccess: FeeMintAccess}} privateArgs
+ * }>} zcf
+ * @param {{feeMintAccess: FeeMintAccess, initialPoserInvitation: Invitation}} privateArgs
  */
-export const start = async (zcf, privateArgs) => {
+export const start = async (zcf, { feeMintAccess, initialPoserInvitation }) => {
   const { anchorBrand, anchorPerStable } = zcf.getTerms();
 
-  const { feeMintAccess } = privateArgs;
   // TODO should this know that the name is 'Stable'
   // TODO get the RUN magic out of here so the contract is more reusable
   const stableMint = await zcf.registerFeeMint('Stable', feeMintAccess);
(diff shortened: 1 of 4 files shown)


packages/run-protocol/test/psm/test-psm.js
```diff
@@ -18,6 +18,10 @@ import { makeTracer } from '../../src/makeTracer.js';
 import { unsafeMakeBundleCache } from '../bundleTool.js';
 import { setUpZoeForTest } from '../supports.js';
 
+/** @type {import('ava').TestInterface<Awaited<ReturnType<makeTestContext>>>} */
+// @ts-expect-error cast
+const test = anyTest;
```
dckc, line 25: nice :)

packages/run-protocol/src/psm/psm.js
```diff
@@ -67,31 +66,20 @@ export const start = async (zcf, privateArgs) => {
       anchorPerStable.denominator.brand === stableBrand,
     X`Ratio ${anchorPerStable} is not consistent with brands ${anchorBrand} and ${stableBrand}`,
   );
+  const bpRatio = bp => makeRatio(bp, stableBrand, BASIS_POINTS);
```
dckc: oops; I forgot to tell you: @dtribble said basis points are too coarse; he wants the governance parameters to be ratios.

packages/run-protocol/src/psm/psm.js
```diff
@@ -103,13 +91,13 @@ export const start = async (zcf, privateArgs) => {
...
     assert(
-      AmountMath.isGTE(gov.getMintLimit(), anchorAfterTrade),
+      AmountMath.isGTE(params.getMintLimit(), anchorAfterTrade),
       X`Request would exceed mint limit`,
     );
   };
 
   const giveStable = (seat, given, wanted = emptyAnchor) => {
-    const fee = ceilMultiplyBy(given, gov.getGiveStableRate());
+    const fee = ceilMultiplyBy(given, bpRatio(params.getGiveStableFeeBP()));
```
dckc: likewise...

packages/run-protocol/src/psm/psm.js
```diff
@@ -127,7 +115,7 @@ export const start = async (zcf, privateArgs) => {
   const wantStable = (seat, given, wanted = emptyStable) => {
     assertUnderLimit(given);
     const asStable = floorDivideBy(given, anchorPerStable);
-    const fee = ceilMultiplyBy(asStable, gov.getWantStableRate());
+    const fee = ceilMultiplyBy(asStable, bpRatio(params.getWantStableFeeBP()));
```
dckc: and again

dckc (changes requested):
please change the NAT basis points parameters to ratios; sorry I forgot to tell you about the additional requirement.

Strictly speaking it could be a separate PR, but it seems expedient to include it here.

packages/run-protocol/src/psm/psm.js
turadg: We don't have a brand-less ratio. We could do a pair of nats to make one.

Is there a grain size that would be sufficient, like 10^-12?

dckc: I don't see a need for brand-less ratios. The caller would have to supply the right brands. Am I confused?

turadg: Cool. I was confusing this with another conversation in RUN meeting.

Done.

dckc (approved):
bingo! thanks.

dckc:
oops; this is going to break the psm bootstrap code in ways that I'm not sure our tests cover. Oh well. that bootstrap code is kinda iffy anyway.

packages/run-protocol/src/psm/psm.js
```diff
@@ -74,42 +70,38 @@ export const start = async (zcf, privateArgs) => {
...
-      AmountMath.isGTE(gov.getMintLimit(), anchorAfterTrade),
+      AmountMath.isGTE(params.getMintLimit(), anchorAfterTrade),
       X`Request would exceed mint limit`,
     );
   };
 
+  /**
+   * @param {ZCFSeat} seat
+   * @param {Amount<'nat'>} given
+   * @param {Amount<'nat'>} wanted
```
Chris-Hibbert: It looks like `wanted` is optional. Is this the right syntax for that?
```suggestion
   * @param {Amount<'nat'>} [wanted]
```

packages/run-protocol/src/psm/psm.js
```diff
@@ -124,10 +116,15 @@ export const start = async (zcf, privateArgs) => {
     stableMint.burnLosses({ Stable: afterFee }, stage);
   };
 
+  /**
+   * @param {ZCFSeat} seat
+   * @param {Amount<'nat'>} given
+   * @param {Amount<'nat'>} wanted
```
Chris-Hibbert: optional

Chris-Hibbert (approved):

dckc:
@turadg did you see @Chris-Hibbert 's optional comments? turning off automerge for a sec...

packages/run-protocol/src/psm/psm.js
turadg: Good catch. Turns out [Typescript infers optional from a default value](https://www.typescriptlang.org/play?filetype=js#code/PQKhFgCgAIWgBADgQwE7ILbQN4GcAuqAlgHYDmAvtANoGoC6UswUAxgPYkHTuL5GdkAG2gBeaAAo6ASjEA+HE2h0AdBxKtk+CdIDcUCvshRe-QUJ1GooCDDhI0mHHVKVlhJiBaR13ACYApgBmyACuQvgBfmKSdKIA5AAWRPGyogrYSqrqmtp6BlaQgSHhkX6WUEA), but this is correct JSDoc.

PR #5192, opened 2022-04-22
Add swing store trace option

mhofman:
## Description

We noticed that the app hash diverges when trying to diagnose why the deployment-test sometimes results in consensus failures, but nothing else seemed different in the slog. This change adds an option to trace modifying calls to the swing store, which is enabled for the integration test

### Security Considerations

None

This tracing can be enabled through a new environment variable `SWING_STORE_TRACE` as well as by sniffing for the presence of the `--trace-store` argument that triggers cosmos to write its own trace (though ignoring the path value and using a hardcoded one)

Since it looks like we may need this on for a bit, I decided to wire it properly

### Testing Considerations

[Tried this based on 3 different revisions so far](https://github.com/Agoric/agoric-sdk/actions/workflows/deployment-test.yml?query=event%3Aworkflow_dispatch+actor%3Amhofman), including the last known failure, and haven't been able to reproduce, which really means something either non-deterministic or environmental is at play here.

### DIFF
--- a/packages/deployment/scripts/capture-integration-results.sh
+++ b/packages/deployment/scripts/capture-integration-results.sh
@@ -15,5 +15,12 @@ for node in validator{0,1}; do
   "$thisdir/setup.sh" ssh "$node" cat "$home/config/genesis.json" > "$RESULTSDIR/$node-genesis.json" || true
   "$thisdir/setup.sh" ssh "$node" cat "$home/data/chain.slog" > "$RESULTSDIR/$node.slog" || true
   "$thisdir/setup.sh" ssh "$node" cat "$home/data/ag-cosmos-chain-state/flight-recorder.bin" > "$RESULTSDIR/$node-flight-recorder.bin" || true
+  "$thisdir/setup.sh" ssh "$node" cat "$home/data/ag-cosmos-chain-state/store-trace.log" > "$RESULTSDIR/$node-store-trace.log" || true
   "$thisdir/setup.sh" ssh "$node" cat "$home/data/kvstore.trace" > "$RESULTSDIR/$node-kvstore.trace" || true
 done
+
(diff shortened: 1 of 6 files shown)


michaelfig (approved):
LGTM.

packages/deployment/scripts/capture-integration-results.sh
```diff
@@ -15,5 +15,12 @@ for node in validator{0,1}; do
...
   "$thisdir/setup.sh" ssh "$node" cat "$home/data/ag-cosmos-chain-state/flight-recorder.bin" > "$RESULTSDIR/$node-flight-recorder.bin" || true
+  "$thisdir/setup.sh" ssh "$node" cat "$home/data/ag-cosmos-chain-state/store-trace.log" > "$RESULTSDIR/$node-store-trace.log" || true
   "$thisdir/setup.sh" ssh "$node" cat "$home/data/kvstore.trace" > "$RESULTSDIR/$node-kvstore.trace" || true
 done
+
+if [ -d /usr/src/testnet-load-generator ]
+then
+  cp ~/.ag-chain-cosmos/data/ag-cosmos-chain-state/flight-recorder.bin "$RESULTSDIR/flight-recorder.bin" || true
+  cp ~/.ag-chain-cosmos/data/ag-cosmos-chain-state/store-trace.log "$RESULTSDIR/store-trace.log" || true
+fi
```
warner: maybe add the missing newline

warner (approved):
Looks good. Maybe expand on the commit comment that removes the `kernelStateDBDir` form of `provideHostStorage`: "`This `provideHostStorage` is only for debugging or one-off ephemeral applications that don't need persistence. Real applications should import `@agoric/swing-store` and create their own hostStorage object so they can control the timing of database commits and synchronize with outgoing IO.".

PR #5193, opened 2022-04-22
ci: Prototype of run integration tests before merge

mhofman:
This PR add the ability to run integration style tests as part of the PR checks, but only under the following combined conditions:
- If the PR is not a draft
- If the PR targets `master`
- If the PR is labeled as `automerge` or `manualmerge` and not `bypass:integration`
- A check has not already run for the same files state

Furthermore it should attempt to run the test on master only if it hadn't previously run on the same state of the git tree, avoiding unnecessary duplicate checks.

The ultimate goal with the `manualmerge` label is that the GH checks would require one of the "merge" labels to enable merging, but that's for a future improvement.

I have setup the branch protections to require the always passing `mergify-merge-queue-test` which I've updated to use this new shared integration workflow.

### DIFF
--- a/.github/workflows/mergify-test.yml
+++ b/.github/workflows/mergify-test.yml
@@ -1,12 +1,36 @@
+name: Proto merge integration
+
 on:
+  workflow_dispatch:
   push:
+    branches:
+      - master
+    tags:
+      - '@agoric/sdk@*'
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+      - labeled
+      - auto_merge_enabled
 
 jobs:
+  pre_check:
+    uses: ./.github/workflows/pre-check-integration.yml
+
   mergify-merge-queue-test:
+    needs: pre_check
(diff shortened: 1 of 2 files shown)


michaelfig (dismissed):
Rubber-stamping!

.github/workflows/mergify-test.yml
```diff
@@ -1,12 +1,36 @@
+name: Proto merge integration
```
michaelfig, line 1: Can you please rename `mergify-test.yml` to something better?  It is just debris left behind from a prior test that should be cleaned up.

michaelfig (approved):

.github/workflows/mergify-test.yml
mhofman: I was thinking of dropping the `mergify-test.yml` file completely once I've confirmed it does what I intend on regular PRs and master, and upgrade at least `deployment-test` to use this new logic (but possibly `integration-test` as well)

PR #5198, opened 2022-04-22
fix(cosmic-swingset): Fully drain actionQueue when recovering an interrupted block

### LINKED ISSUE #5196: Interrupted block recovery appears to be broken
## Describe the bug
I'm not sure how to verify or test this, but it looks like cosmic-swingset code for handling "after a restart when Tendermint is reevaluating the block that was interrupted and not committed." The code at https://github.com/Agoric/agoric-sdk/blob/4aaa7095c005a2423f2e9b20b945800382484053/packages/cosmic-swingset/src/block-manager.js#L197-L207 interrupts a for-of iteration of `actionQueue.consumeAll()` with `break`, which results in the invocation of the [iterator's `return` method](https://github.com/Agoric/agoric-sdk/blob/4aaa7095c005a2423f2e9b20b945800382484053/packages/cosmic-swingset/src/chain-main.js#L141-L144) after receiving the first result (i.e., after one invocation of `next()`). But that `return` method seems to have two problems:
1. It commits to "head" a value that has been incremented only once, potentially leaving the queue nonempty.
2. It returns undefined rather than an object, which per [IteratorClose](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-iteratorclose) results in a TypeError exception and thus suppression of later code intended to handle the restart.

## Expected behavior
I _think_ the for-of loop is intended to silently drain `actionQueue`.

gibson042:
Fixes #5196

## Description

Make `chainQueue` iterators interface-conforming and close across all methods, and update `blockingSend` functions returned from `makeBlockManager` to fully drain such queues when recovering an interrupted block.

### Security Considerations

I'm not aware of any, but this is unfamiliar code.

n/a

### Testing Considerations

Based on the fact that the existing `return` code results in a TypeError, there don't seem to be any tests covering this. I would appreciate help in writing them.


michaelfig (approved):
LGTM!  How did you find this problem?

gibson042:
Reviewing code that runs at startup to improve my comprehension for building out the new vat correctly. Do you have any thoughts on how to add a test covering this?

michaelfig:
> Do you have any thoughts on how to add a test covering this?

Short of extracting the code to a helper module and scaffolding it from there, no I don't have any thoughts.

PR #5199, opened 2022-04-22
fix(bundleTool): don't ignore module path in metadata check

Chris-Hibbert:
## Description

commit, originally by @dckc extracted from #5129

[See also](https://github.com/Agoric/agoric-sdk/pull/5129/commits/00d32a7e84d8d4fd9309a1e115145d5164c93b19#r854207585)

### Security Considerations

This is updating bundleTool.  I dunno what the security implications are?

I think there are none. This is an internal tool.

### Testing Considerations

None


dckc (approved):
thanks!

PR #5201, opened 2022-04-22
chore(deps): upgrade Ava to 4.2

### LINKED ISSUE #4895: Upgrade Ava to version 4

We're on an older version which lacks some useful functionality (e.g. macro improvements, shared workers) and causes some confusion when reading documentation.

## Description of the Design

https://github.com/Agoric/agoric-sdk/pull/4727 took a stab. We'll need to solve that Ava dropped support for React.

## Security Considerations
--
## Test Plan

--

turadg:
closes: #4895

Upgrade Ava to latest (4.2) and make it work.

Changes:
- https://github.com/avajs/ava/releases/tag/v4.0.0
- https://github.com/avajs/ava/releases/tag/v4.1.0
- https://github.com/avajs/ava/releases/tag/v4.2.0

TODO
- [ ] update to 4.3 and note https://github.com/avajs/ava/pull/3013
- [ ] get integration tests passing
- [ ] update [VScode config](https://github.com/avajs/ava/issues/2946#issuecomment-1144185486)?


turadg:
Last hurdle here seems to be this error in integration tests with dapps,
```
  Uncaught exception in test/test-contract.js

  AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: The expression evaluated to a falsy value:

    assert(refs.runnerChain)
```

which seems to be https://github.com/avajs/ava/issues/2946#issuecomment-1073268723

@kriskowal any ideas?

kriskowal:
This does indeed look like the same problem others are having on the new versions of AVA. It doesn’t seem to be interference with hardened JavaScript and it’s hard to imagine it being related to the other things we’ve seen where VS Code debugger tampering with the Node.js environment when a debugger is attached. I’m at a loss.

turadg:
My current hypothesis for the failures are that it's conflicting versions of Ava due to this line in the failing tests:
```
import { test } from '@agoric/zoe/tools/prepare-test-env-ava.js';
```

That package (in this PR) is using Ava 4 and the dapps are on Ava 3. I think we need to decouple the Ava versions. cc @michaelfig  https://github.com/Agoric/agoric-sdk/issues/5244 may help.

turadg:
> That package (in this PR) is using Ava 4 and the dapps are on Ava 3. I think we need to decouple the Ava versions

One solution may be to make a new `@endo/ses-ava-4` wrap that can coexist with the earlier one (for Ava 3). cc @kriskowal

Issue #5202, opened 2022-04-22
Node 16 GC issues bollixing tests that depend on GC


FUDCo:
## Describe the bug
The exact underlying problem is unclear at this point, but it appears that forcing the garbage collector to run in Node 16 may, in some cases, result in incomplete garbage collection.  This unpredictably and disruptively interferes with the correct execution of the SwingSet tests that exercise garbage collection of virtual objects and stores.

## To Reproduce
1. Be sure you're running Node 16.  The problem does not appear in Node 14.  Its status in other Node versions is as yet unexamined.
2. From the `packages/SwingSet` directory run `yarn test test/stores/test-storeGC.js`
3. As of master branch commit 2d0ce6b2e7cffcdd191490289991a3d80cf484f9 the `store refcount management 3` test will fail.
If you run the failing test specifically, i.e., with `yarn test test/stores/test-storeGC.js -m 'store refcount management 3'` the test will succeed.
The failure will also occur with some subsets of the tests in this file but not other subsets.
It can be reproduced in the debugger as long as you don't try to get too close to the failure point.  In particular, if you single step through the `prepareStoreLinked` function (the specific piece of code which manifests the misbehavior in this case) the problem will go away.

## Expected behavior
All tests should succeed.

## Platform Environment
 - specific failure conditions were found in MacOS 12.3.1 Monterey, though I do not believe the OS is implicated.  The exact versions of Node where this was studied were 16.13.2 where it failed and 14.18.3 where it did not.
 - this was studied using the code on the `master` branch as of the time this report was being written, commit 2d0ce6b2e7cffcdd191490289991a3d80cf484f9

## Additional context
This is a remanifestation of a problem I thought we had earlier defeated with some hackery in the GC test infrastructure to force host GC at the appropriate time, but that hackery appears to have been insufficient.

My current working hypothesis is that the forced GC does not perform a complete garbage collection and thus fails to identify all the garbage that should have been collected.  Experimentally, at the point where the call to force GC happens, issuing the call twice in a row appears to make the problem go away, at least in the instance of the particular problem that was examined here.  

The proposed fix is to make that double GC call permanent.  This is a very speculative and unsatisfying solution, but since the problem manifests in weird special test code rather than any production systems, I think it is acceptable at least until we know more.

FUDCo:
@mhofman @warner

mhofman:
[...]

PR #5204, opened 2022-04-23
Split test storeGC

mhofman:
## Description

I've been hitting the issue in #5202 pretty consistently with a seemingly unrelated commit from #5124: https://github.com/Agoric/agoric-sdk/commit/4bd30d994fb29c07d2f203b3d5ee9f23687dc92b.

The only solution I found was to split up the `test-storeGC.js` into multiple files, which I've extracted into this PR.

### Testing Considerations

Making future tests pass again!


FUDCo:
I'm confused about the role of the first commit here.  Should this PR consist solely of the second commit?

mhofman:
The first commit is the motivation for the second one. Since it's self contained as well, I figured  putting them in the same PR didn't hurt.

FUDCo:
I think these are completely decoupled.  You need the test split to run the tests at all, but that's independent of the other changes -- those changes didn't cause the need to do the test split, it's just that the test failures are blocking you.  But they'll block anybody else trying to make forward progress against `master` with more or less the same runtime setup you've got (i.e., Node 16 et al).  When I reproduced the test failures in my environment I didn't have your first set of changes at all.  Since these are two completely unrelated bits of work, I think they should be separate PRs.

mhofman:
Well at least in my environment, and apparently in CI, the store GC tests work fine without the first commit, but fail with the first commit. So I wouldn't say they're independent, and that commit was indeed my motivation for the split

mhofman:
@FUDCo I've removed the first commit, PTAL

FUDCo (approved):
👍

PR #5186, opened 2022-04-21
refactor(run-protocol): AttestationTool vobj

turadg:
closes: #4534

## Description

`makeAttestationFacets` holds a `Store<Address, AttestationMaker>` so AttestionMaker needs to be a vobj.

This makes it so and cleans up types. (`AttestationMaker` --> `AttestationTool`)

### Security Considerations

State paged out, relies on VatData.

### Documentation Considerations

--

### Testing Considerations

Refactor, existing tests suffice. Durability out of scope.

PR #5205, opened 2022-04-23
fix(run-protocol): restore wrapLienedAmount on AttestationTool

turadg:
## Description

Solving https://github.com/Agoric/agoric-sdk/pull/5186#issuecomment-1107069676

That had been added in https://github.com/Agoric/agoric-sdk/pull/5094

### Security Considerations

No new.

--

### Testing Considerations

There should be a test for the feature that was added in https://github.com/Agoric/agoric-sdk/pull/5094 . I'd like to hold off on merging this until we have a regression test but if it's urgent @samsiegart okay with me to merge.

### DIFF
--- a/packages/run-protocol/src/runStake/attestationTool.js
+++ b/packages/run-protocol/src/runStake/attestationTool.js
@@ -48,8 +48,12 @@ const behavior = {
   wrapLienedAmount: ({ state }, lienedAmount) =>
     state.lienMint.wrapLienedAmount(state.address, lienedAmount),
 
-  /** @param {MethodContext} context */
-  unwrapLienedAmount: ({ state }) => state.lienMint.unwrapLienedAmount,
+  /**
+   * @param {MethodContext} context
+   * @param {Amount<'copyBag'>} attAmount
+   */
+  unwrapLienedAmount: ({ state }, attAmount) =>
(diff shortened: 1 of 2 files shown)


dckc:
the dapp can do its own wrapping / unwrapping

this feature should probably go in something like `params.js` that clients can use without grabbing the whole contract

it was just convenient to have them close over the relevant brands

dckc:
@turadg you assigned @samsiegart but didn't make him a reviewer.

(several folks around here tend to assign themselves on PRs they author)

turadg:
> @turadg you assigned @samsiegart but didn't make him a reviewer.

A user input error. I’ve corrected it, though I assume anyone who wants to review would review, including Sam since he reported the problem.

> (several folks around here tend to assign themselves on PRs they author)

That makes sense to me if reassigning the PR to someone else,. The 95% case is that the author is the assignee so clicking that seems like busywork. We don’t use it for reporting.

> the dapp can do its own wrapping / unwrapping
> this feature should probably go in something like params.js that clients can use without grabbing the whole contract

SGTM.l @samsiegart will that work for you?

dckc (approved):
I also prefer to have a regression test, but if this becomes urgent, feel free to merge it.

samsiegart (approved):

PR #5206, opened 2022-04-23
feat(wallet): support attestations in offers

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/4263

cc @turadg


packages/wallet/api/src/lib-wallet.js
```diff
@@ -556,6 +556,46 @@ export function makeWallet({
...
+      return pk.promise;
+    };
+  };
+
+  const getAttIssuer = makeMemoizedGetter(async () =>
+    E(agoricNames)?.lookup('issuer', 'Attestation'),
+  );
+
+  const getAttBrand = makeMemoizedGetter(async () => {
+    const issuer = await getAttIssuer();
```
dckc: No need for this await, is yhere?

packages/wallet/api/src/lib-wallet.js
```diff
@@ -556,6 +556,46 @@ export function makeWallet({
...
+    return () => {
+      if (pk === undefined) {
+        pk = makePromiseKit();
+        doGet();
+      }
+      return pk.promise;
+    };
+  };
+
+  const getAttIssuer = makeMemoizedGetter(async () =>
```
turadg: Most `async` functions here can be synchronous functions returning `E()`

packages/wallet/api/src/lib-wallet.js
```diff
@@ -556,6 +556,46 @@ export function makeWallet({
...
+    const issuer = await getAttIssuer();
+    return E(issuer).getBrand();
+  });
+
+  const getRunStakePublicFacet = makeMemoizedGetter(async () => {
+    const instance = await E(agoricNames)?.lookup('instance', 'runStake');
+    return E(zoe).getPublicFacet(instance);
+  });
+
+  const attMakerPK = makePromiseKit();
```
turadg, line 584: Worth having a type annotation here

packages/wallet/api/src/lib-wallet.js
```diff
@@ -572,21 +612,62 @@ export function makeWallet({
...
-      async ([keyword, amount]) => {
+      async ([keyword, { type, ...amount }]) => {
         const purse = purseKeywordRecord[keyword];
         assert(
           purse !== undefined,
           X`purse was not found for keyword ${q(keyword)}`,
         );
-        const payment = await E(purse).withdraw(amount);
-        paymentToPurse.set(payment, purse);
+        let payment;
```
turadg: No need for a mutable var. Can just return in each condition. Type system will ensure consistency.

packages/wallet/api/src/lib-wallet.js
samsiegart: Good catch, removed

packages/wallet/api/src/lib-wallet.js
samsiegart: Done

packages/wallet/api/src/lib-wallet.js
samsiegart: Done

packages/wallet/api/src/lib-wallet.js
samsiegart: Done

samsiegart:
Addressed feedback, added unit tests, and tested manually against WIP UI and [local chain](https://gist.github.com/samsiegart/92f32c0bc8a93be75944473ac32f0763). PTAL.

packages/wallet/api/src/lib-wallet.js
```diff
@@ -33,6 +33,7 @@ import { makePaymentActions } from './actions.js';
 
 import '@agoric/store/exported.js';
 import '@agoric/zoe/exported.js';
+import '@agoric/run-protocol/exported.js';
```
turadg, line 36: 👍

packages/wallet/api/src/lib-wallet.js
```diff
@@ -556,11 +557,42 @@ export function makeWallet({
       });
   }
 
+  /**
+   * @param { () => ( Promise | undefined ) } get - The function whose return value
+   * to memoize.
+   */
+  const makeMemoizedGetter = get => {
+    let pk;
```
turadg, line 565: Couldn't this be a constant? There's one assignment and I don't see the benefit to deferring it.

packages/wallet/api/src/lib-wallet.js
```diff
@@ -556,11 +557,42 @@ export function makeWallet({
...
+      }
+      return pk.promise;
+    };
+  };
+
+  const getAttIssuer = makeMemoizedGetter(() =>
+    E(agoricNames)?.lookup('issuer', 'Attestation'),
+  );
+
+  const getAttBrand = makeMemoizedGetter(() => E(getAttIssuer()).getBrand());
```
turadg, line 581: This doesn't need a function def here does it? Can't we pass `getBrand` itself?

packages/wallet/api/src/lib-wallet.js
```diff
@@ -556,11 +557,42 @@ export function makeWallet({
...
+
+  /** @type import('@endo/promise-kit').PromiseKit<AttestationTool> */
+  const attMakerPK = makePromiseKit();
+
+  const getAccountState = async () => E(attMakerPK.promise).getAccountState();
```
turadg: This isn't an async function. It's a sync function that returns a promise. I think it doesn't even need a function definition:
```
const { getAccountState } = E(…
```

packages/wallet/api/test/attestationExample.js
```diff
@@ -0,0 +1,64 @@
...
+  const mintAttestation = ({ value }) => {
+    const { userSeat, zcfSeat } = zcf.makeEmptySeatKit();
+    mint.mintGains({ Attestation: AmountMath.make(brand, value) }, zcfSeat);
+    zcfSeat.exit();
+    return E(userSeat).getPayout('Attestation');
+  };
+
+  const wantAttHandler = seat => {
+    const proposal = seat.getProposal();
+    // @ts-expect-error
```
turadg: This wouldn't be necessary if the array declaration were typed.

packages/wallet/api/test/attestationExample.js
```diff
@@ -0,0 +1,64 @@
...
+    callHistory.returnProposals.push(proposal);
+    // @ts-expect-error
+    callHistory.returnAllocations.push(seat.getCurrentAllocation());
+    seat.exit();
+    return 'thank you for that';
+  };
+
+  const makeWantAttInvitation = () =>
+    zcf.makeInvitation(wantAttHandler, 'WantAtt');
```
turadg: Nascent consistency idea: if we call it `handleWantAtt` then it'll be verb first like 'make' and have the same case sensitive string for the noun.

turadg (commented):
Reviewed as a code reader which is all I can manage on my phone on my day off. Leaving attestation logic review to tomorrow or to @dckc

packages/run-protocol/src/runStake/attestationTool.js
```diff
@@ -40,6 +40,20 @@ const behavior = {
...
+   */
+  wrapLienedAmount: ({ state }, lienedAmount) =>
+    state.lienMint.wrapLienedAmount(state.address, lienedAmount),
+
+  /**
+   * @param {MethodContext} context
+   * @param {Amount<'copyBag'>} attAmount
+   */
+  unwrapLienedAmount: ({ state }, attAmount) =>
+    state.lienMint.unwrapLienedAmount(attAmount),
```
dckc: This would be a contract change. Didn't we already make this change?

dckc: The contract already provides those; I just checked...

```
command[114] E(home.attMaker).help()
history[114] Promise.reject("TypeError: target has no method \"help\", has [\"getAccountState\",\"makeAttestation\",\"makeReturnAttInvitation\",\"unwrapLienedAmount\",\"wrapLienedAmount\"]")
```

packages/wallet/api/src/lib-wallet.js
samsiegart: I was using `pk === undefined` as a way to signal that `get()` was not in flight already. Maybe there's a better way to do so?

packages/wallet/api/src/lib-wallet.js
samsiegart: We could, that would just mean it does the request when the wallet is initialized. Maybe that's okay, I was just thinking we didn't want to fetch these until they're used.

packages/wallet/api/src/lib-wallet.js
samsiegart: Nice, fixed

packages/wallet/api/test/attestationExample.js
samsiegart: Fixed

packages/wallet/api/test/attestationExample.js
samsiegart: Good idea, done

packages/run-protocol/src/runStake/attestationTool.js
samsiegart: Yes sorry, I had patched it while working and it must've gotten mixed up. I removed the file from the commit, but I did just have to fix the type definition for `AttestationTool`.

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -85,15 +85,51 @@ async function setupTest() {
   const addLiquidityInvite =
     await autoswapPublicFacet.makeAddLiquidityInvitation();
 
+  // Create attestationExample instance
```
dckc: This adds a `bundleSource`, `install`, and `startInstance` to every test, right? 9 of them, it seems. How about putting the `bundleSource` and `install` in `test.before()`?

I would measure the performance impact, but I'm not sure how:

```
agoric-sdk/packages/wallet$ yarn test api/test/test-lib-wallet.js 
yarn run v1.22.17
$ exit 0 api/test/test-lib-wallet.js
Done in 0.05s.

agoric-sdk/packages/wallet$ yarn test
yarn run v1.22.17
$ exit 0
Done in 0.05s.
```

```
agoric-sdk/packages/wallet$ cd api

agoric-sdk/packages/wallet/api$ yarn test
yarn run v1.22.17
$ ava

  ✖ No tests found in test/test-getPursesNotifier.js
  ✖ No tests found in test/test-lib-wallet.js

  ─

  Uncaught exception in test/test-lib-wallet.js

  SyntaxError: Unexpected token '==='



  Uncaught exception in test/test-getPursesNotifier.js

  SyntaxError: Unexpected token '==='

  ─

  1 test passed
  2 uncaught exceptions
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

dckc (commented):
I'm struggling to review this because `yarn test` doesn't do anything; the package doesn't follow our [unit testing norms](https://github.com/Agoric/agoric-sdk/wiki/agoric-sdk-unit-testing)

packages/wallet/api/test/test-lib-wallet.js
dckc: @turadg who set up the ci for this package? git blame shows 2f0073a for these lines...

```yaml
      - name: yarn test (wallet/api)
        run: cd packages/wallet/api && yarn test
```

that commit has you as author, but it's only a formatting change.

michaelfig: "this package" should be `packages/wallet/api` to run the test, which is what CI does.  `packages/wallet` is an empty stub node.  Were it enabled, it would run the tests twice.

michaelfig: The tests work in current master's `cd packages/wallet/api && yarn test`.

samsiegart: I just pushed a commit to optimize the tests. It was taking 13 seconds on my machine before the changes, 26 seconds after the changes, and now 6 seconds after optimizing the contract installation/instantiation

dckc: > SyntaxError: Unexpected token '==='
I messed up checking out the branch; I used `git pull` at one point and got merge conflicts.

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1072,15 +1155,161 @@ test('lib-wallet addOffer for autoswap swap', async t => {
...
+    attestationIssuer,
+  } = await setupTest({ attestation: true });
+
+  const issuerManager = wallet.getIssuerManager();
+  await issuerManager.add('Attestation', attestationIssuer);
+  await wallet.makeEmptyPurse('Attestation', 'staking purse');
+
+  const invitation = await attestationPublicFacet.makeReturnAttInvitation();
+
+  const rawId = '1593482020370';
```
dckc: Is that an arbitrary Id? or something carefully calculated to get expected results?
I changed it and the test still works, so it seems to be arbitrary. I suggest:

```suggestion
  const rawId = '123-arbitrary';
```

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1072,15 +1155,161 @@ test('lib-wallet addOffer for autoswap swap', async t => {
...
+          type: 'Attestation',
+        },
+      },
+      exit: {
+        onDemand: null,
+      },
+    },
+  };
+
+  await wallet.addOffer(offer);
```
dckc, line 1228: odd... I get a type when I hover over `wallet`, but the type of the arg to `addOffer` is `any`. Bummer.

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1072,15 +1155,161 @@ test('lib-wallet addOffer for autoswap swap', async t => {
...
+  const accepted = await wallet.acceptOffer(id);
+  assert(accepted);
```
dckc, line 1230: Is this supposed to be a test assertion? How about...

```suggestion
  t.true(accepted);
```

packages/wallet/api/src/lib-wallet.js
dckc: This looks like the interim kludge we came up with, not the notifier based solution from Apr 13: https://github.com/Agoric/agoric-sdk/issues/4263#issuecomment-1098477427

On the other hand... did that train leave the station? Did we need to add the `walletNotifier` property in the governance proposal that just passed? @michaelfig ?

dckc (commented):
This doesn't look like the notifier based solution we discussed Apr 13 https://github.com/Agoric/agoric-sdk/issues/4263#issuecomment-1098477427

I defer to @michaelfig as to whether we want this interim solution merged.

If so, please create an issue to represent the notifier based solution.

dckc (approved):
> If so, please create an issue to represent the notifier based solution.

never mind; I see this refs: #4263 , but doesn't claim to close it.

As discussed, let's go ahead.

I prefer to have the module-level `let`s converted to properties on `t.context` but I can live with it as is.

packages/wallet/api/test/test-lib-wallet.js
samsiegart: It's actually to make typescript happy, on the next line it tries to access `{ depositedP } = accepted`, but accepted might be `undefined`.

samsiegart:
> I prefer to have the module-level `let`s converted to properties on `t.context` but I can live with it as is.

Fixed

packages/wallet/api/test/test-lib-wallet.js
samsiegart: Done

packages/wallet/api/test/test-lib-wallet.js
turadg: fwiw, that won't be necessary in Ava 4 which adds type guards https://github.com/avajs/ava/releases/tag/v4.0.0

packages/wallet/api/src/lib-wallet.js
turadg: Hmm, maybe I don't understand the intent of this function. Is is to defer calling the getter? Does this also do what you want?
```js
  const makeMemoizedGetter = get => {
    return () => {
      const pk = makePromiseKit();
      pk.resolve(get());
      return pk.promise;
    };
  };
```

If so, that can be simplified,
```js
  const makeMemoizedGetter = get => {
    return () => Promise.resolve(get());
  };
```
Though I wouldn't call those "memoizing" because there's no cache. It's just a deferrred promise.

packages/wallet/api/src/lib-wallet.js
turadg: Why wouldn't we want to get them at wallet initialization time? Will they change? Is it a perf optimization? Does wallet init block some UI?

turadg (commented):
I think the memoization needs some work but not blocking.

packages/wallet/api/test/test-lib-wallet.js
dckc: ah. type safety. fair enough.

packages/wallet/api/src/lib-wallet.js
samsiegart: They shouldn't change, it's to keep the wallet initialization slim. Maybe it's an unnecessary optimization, but wallet init blocks `agoric start --reset` and eventually whatever UI we have for deploying a new wallet on-chain.

packages/wallet/api/src/lib-wallet.js
samsiegart: Yes it's to:
1) Defer calling the getter until its result needs to be read.
2) Only call the getter once no matter how many subsequent/concurrent reads are made for the getter's return value.

The second part works as currently written, but not with those examples.

turadg: ha yeah, I lost that along the way. Thanks for spelling it out. A _deferred and memoized_  getter.

This reads clearer to me as a memoization idiom and doesn't require Agoric's custom promise kit:
```
const makeMemoizedGetter = get => {
  let valP;
  return () => {
    if (!result) {
      valP = Promise.resolve(get());
    }
    return valP;
  };
};
```
With types:
```
/**
 * @template T
 * @param {() => T} get
 */
const makeMemoizedGetter = get => {
  /** @type {Promise<T>?} */
  let result;
  return () => {
    if (!result) {
      result = Promise.resolve(get());
    }
    return result;
  };
};
```
not blocking

turadg (approved):

michaelfig (approved):
Sorry for the delay.  I consider the other two review approvals to be sufficient, and I took a look through the code for any obvious gotchas.

LGTM!

PR #5207, opened 2022-04-23
First pass at VO and DVO test harnesses

### LINKED ISSUE #4997: need test harness for virtual/durable object restoration

@Chris-Hibbert pointed out that we need a good way to test that a userspace (contract) which uses virtual/durable Kinds will correctly reload those objects after getting paged out and paged back in (or after the entire vat gets upgraded).

Instances of a virtual/durable Kind are presented to JavaScript code by a "Representative", much like how a "Presence" object is the way JS interacts with a remote object in some other vat. Each Representative may or may not have its state data in RAM: there is an LRU cache which limits how many objects' data will be loaded at a time. When a method is invoked on a Representative, its state data is loaded (or moved to the front of the cache), and this might push some other object's data off the cache, which is what writes it to the DB. The dirty state records are also written to DB when a special "bringOutYourDead" delivery is made, which happens on a configurable schedule, somewhere between once after every normal delivery, once every N deliveries, and once at the end of each block.

The Representative itself is held in memory either by the userspace code that holds it, or by the LRU cache. Once it is not held by either, the next time GC happens (either "organically" because XS needs a new `struct slot` and the free list is empty, or when `bringOutYourDead` forces a GC sweep).

By design, normal vat userspace cannot reliably force an object out of memory, nor can it sense when one has been forced out. If you drop your RAM references to the Representative, then allocate or interact with a whole bunch of other objects, you might be able to both flood the LRU *and* trigger organic GC. The next time you try to get a Representative (e.g. reading it from a property of some other virtual object, fetching it from a collection, or receiving it in a message from a different vat), the manager will build a new one for you. This is the point at which the Kind's registered `behavior` record is used, to create a `state` object and a new batch of methods to wrap it, which collectively make up the Representative(s).

So the low-level requirement is for some sort of test harness to allow this "rebuild a Representative" process to be exercised.

I'm not yet sure how to do this. If virtual objects work correctly, there should be no way for userspace to mess up (if it could, userspace could probably cause a consensus failure or state divergence). So in once sense, userspace authors don't need to test this; swingset authors do.

There is a related higher-level requirement here: contract upgrade means vat upgrade, and the version-1 code must be designed to store enough data in durable objects/collections to enable version-2 to fulfill its obligations to the existing clients. *That* sort of "restore" is very much user-visible. We need a test harness for this too.

FUDCo:
Closes #4997

This is the initial version of test harnesses for virtual objects.  This been partially exercised by @Chris-Hibbert, though the tooling that deals with durable objects will have to wait for people to start doing durable objects stuff before it can be vetted thoroughly.  However, it is as ready to go now as it will be for a while.

### DIFF
--- a/packages/SwingSet/test/liveslots-helpers.js
+++ b/packages/SwingSet/test/liveslots-helpers.js
@@ -14,6 +14,9 @@ import {
   makeBringOutYourDead,
 } from './util.js';
 
+/**
+ * @param {boolean} [skipLogging = false]
+ */
 export function buildSyscall(skipLogging) {
   const log = [];
   const fakestore = new Map();
@@ -114,7 +117,7 @@ export function buildSyscall(skipLogging) {
     },
   };
 
-  return { log, syscall };
+  return skipLogging ? { syscall } : { log, syscall };
 }
 
 export async function makeDispatch(
--- a/packages/SwingSet/test/liveslots-helpers.js
+++ b/packages/SwingSet/test/liveslots-helpers.js
@@ -14,12 +14,18 @@ import {
   makeBringOutYourDead,
 } from './util.js';
 
-export function buildSyscall() {
+export function buildSyscall(skipLogging) {
   const log = [];
   const fakestore = new Map();
   let sortedKeys;
   let priorKeyReturned;
   let priorKeyIndex;
+
+  function appendLog(s) {
+    if (!skipLogging) {
+      log.push(s);
+    }
+  }
 
   function ensureSorted() {
     if (!sortedKeys) {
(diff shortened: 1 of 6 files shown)


Chris-Hibbert:
This works for the simplest case that I tried first (AMM pools), but thinking about other contracts, I wonder whether there are scenarios that are easiest to test with multiple objects created together, either all of the same kind, or of different kinds. I have noticed that the vault virtualization has separate virtual objects for `vaultTitle`, `vaultDirector`, `innerVault`, and `vaultManagerKit`. I presume some of these embed the others.

FUDCo:
That is a good point.  If the thing returned by your `makeTestObject` function is actually the root of an object graph, you should be able to write tests that exercise the whole graph, not just the single object that was returned.  (This might argue for revising some of the parameter names and documentation to reflect this possibility).  In an extreme case one might create a special virtual object kind to be used strictly within the test itself (i.e., not part of the implementation one is trying to validate) to function as a kind of ersatz graph root, to organize or hold onto a more complicated constellation of objects that are not otherwise intrinsically structured as a natural graph on their own.  It's plausible to me that there might be more complex situations that can't readily be fitted into this mold, though at the moment I'm unable to think of any such use cases -- but that may just be a failure of imagination on my part.

packages/SwingSet/test/vo-test-harness/test-vo-test-harness.js
```diff
@@ -0,0 +1,52 @@
...
+      `thing test failed in ${phase} phase`,
+    );
+  }
+
+  await runVOTest(t, prepare, makeTestThing, testTestThing);
+}
+
+test.failing('fail during "before" phase', async t => {
+  await voTestTest(t, 'before');
+});
```
turadg, line 40: Copied from https://github.com/Agoric/agoric-sdk/pull/5224#discussion_r858149318

>> I think you want a passing test that expects a throw.

> Not relevant in this case.

Why?

packages/SwingSet/test/liveslots-helpers.js
```diff
@@ -14,13 +14,19 @@ import {
   makeBringOutYourDead,
 } from './util.js';
 
-export function buildSyscall() {
+export function buildSyscall(skipLogging) {
```
turadg, line 20:
```suggestion
/**
  * @param {boolean} [skipLogging=false]
  */
export function buildSyscall(skipLogging) {
```

I'm surprised that this still returns `log` (even empty) when `skipLogging` is true. No request re: that.

packages/SwingSet/test/vo-test-harness/test-vo-test-harness.js
FUDCo: Because this is a test testing another test. The event being detected is not an exception being thrown, the event is a test failing as expected because it (successfully!) found the failure condition it was checking for.

packages/SwingSet/test/liveslots-helpers.js
FUDCo: I was just being lazy, since this is not production code.  The point was simply to avoid accumulating a bunch of crap in a buffer that's going to be ignored anyway.  But in the interest of hygiene I can change the return to:
```
return skipLogging ? { syscall } : { log, syscall };
```
(I'll also add the param type comment)

warner (approved):
Cool, looks good.

PR #5208, opened 2022-04-23
refactor: change notification of param changes to be once per set

Chris-Hibbert:
spin-off of #5129

## Description

When converting parameter changes under governance to be done atomically as a set of changes, I didn't update the notification. This changes it to only notify once when a collection of changes is processed.

### Security Considerations

It no longer gives the new values, so the recipient of the notification has to look it up if they care.

None.

### Testing Considerations

Added a new test to show successful update of multiple params

### DIFF
--- a/packages/governance/src/contractGovernance/paramManager.js
+++ b/packages/governance/src/contractGovernance/paramManager.js
@@ -76,7 +76,6 @@ const makeParamManagerBuilder = zoe => {
     const setParamValue = newValue => {
       assertion(newValue);
       current = newValue;
-      publication.updateState({ name, type, value: current });
       return harden({ [name]: newValue });
     };
     setParamValue(value);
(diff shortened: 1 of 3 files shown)


packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -312,6 +306,7 @@ const makeParamManagerBuilder = zoe => {
       return [name, setFn(results[i])];
     });
     const newValues = Object.fromEntries(tuples);
+    publication.updateState({ paramNames });
```
dtribble, line 309: Just to confirm, a multi-parameter update can update just a subset of the params?

dtribble (approved):
LGTM

packages/governance/src/contractGovernance/paramManager.js
Chris-Hibbert: absolutely!

PR #5211, opened 2022-04-24
feat(vault): governance upgrade of liquidation

### LINKED ISSUE #5015: Governance determine liquidation strategy for Vaults

The liquidation strategy is a parameter for Vaults. It's currently specified as a `Strategy` object, which is an API with `{ makeInvitation, keywordMapping, makeProposal }`.  

Our assumption was that the liquidation strategy could be specified separately from the contract, but no mechanism was provided.

## Description of the Design

The choice of liquidation approach should be controlled by governance. It's obviously a parameter. It would be possible to add a type for the strategy, or treat is as `unknown`, but I think would be perfectly sensible to use an `instance`. The liquidation contract (`liquidateMinimum.js`) is parameterized with the AMM, and probably the liquidation penalty, so the `installation` wouldn't be sufficient.

Vaults should be refactored to take the liquidation `instance` as a governed parameter.

We don't yet have a solid proposal for a second liquidation approach, so the need for this is not imminent. Various theories are being bandied about. 

## Security Considerations

Standard cross-contract dependency issues. The liquidation contract collects fees and holds them for the Reserve (#5012). Governance would control which contract the Vaults make use of.

## Test Plan

Verify that liquidation approach changes after the vote.

dtribble:
closes: #5015

## Description

* move liquidator spawning into vaultManager
* react to governance by setting up a new liq contract
*   only if the install or parameters have changed
* remove liquidation strategy

### Security Considerations

The old liquidation contract remains; it just has no customers.

N/A

### Testing Considerations

Added "update liquidator" to vote in a change to the liquidation terms and observe that the liquidation instance was updated

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/run-protocol/src/vaultFactory/vaultDirector.js
@@ -153,7 +153,7 @@ const watchGovernance = (govParams, vaultManager, oldInstall, oldTerms) => {
   observeIteration(subscription, {
     updateState(_paramUpdate) {
       const { install, terms } = getLiquidationConfig(govParams);
-      if (install === oldInstall || keyEQ(terms, oldTerms)) {
+      if (install === oldInstall && keyEQ(terms, oldTerms)) {
         return;
       }
       oldInstall = install;
@@ -274,7 +274,7 @@ const machineBehavior = {
     );
     collateralTypes.init(collateralBrand, vm);
     const { install, terms } = getLiquidationConfig(directorParamManager);
-    console.log('PARAM UPDATE', install, terms);
+    // console.log('PARAM UPDATE', install, terms);
     await vm.setupLiquidator(install, terms);
     watchGovernance(directorParamManager, vm, install, terms);
     return vm;
--- a/packages/run-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/run-protocol/src/vaultFactory/vaultDirector.js
@@ -6,7 +6,7 @@ import '@agoric/zoe/src/contracts/exported.js';
 import { E } from '@endo/eventual-send';
 import '@agoric/governance/src/exported.js';
 
-import { fit, M, makeScalarMap } from '@agoric/store';
+import { fit, keyEQ, M, makeScalarMap } from '@agoric/store';
 import {
   assertProposalShape,
   getAmountOut,
@@ -18,8 +18,8 @@ import { Far } from '@endo/marshal';
 import { AmountMath } from '@agoric/ertp';
 import { assertKeywordName } from '@agoric/zoe/src/cleanProposal.js';
 import { defineKindMulti } from '@agoric/vat-data';
+import { observeIteration } from '@agoric/notifier';
 import { makeVaultManager } from './vaultManager.js';
-import { makeLiquidationStrategy } from './liquidateMinimum.js';
 import { makeMakeCollectFeesInvitation } from '../collectFees.js';
 import {
   makeVaultParamManager,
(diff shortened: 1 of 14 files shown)


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -143,6 +143,28 @@ const getCollaterals = async ({ state }) => {
...
+  install: directorParamManager.getInstallation(LIQUIDATION_INSTALL_KEY),
+  terms: directorParamManager.getUnknown(LIQUIDATION_TERMS_KEY),
+});
+
+const watchGovernance = (govParams, vaultManager, oldInstall, oldTerms) => {
+  const subscription = govParams.getSubscription();
+  observeIteration(subscription, {
+    updateState(_paramUpdate) {
+      const { install, terms } = getLiquidationConfig(govParams);
+      if (install === oldInstall || keyEQ(terms, oldTerms)) {
```
Chris-Hibbert: I think you don't want to block updates when the installation is the same. The only way to change the terms is to start a new instance with different terms.
```suggestion
      if (install === oldInstall && keyEQ(terms, oldTerms)) {
```

Chris-Hibbert:
> We could switch between the two liquidation contracts we have. How do we cause the governance to happen in the unit tests with the current governance setup?

We could also switch to the same liquidateIncrementally contract, but with different params.  Oh, that won't work because it tests for `if (install === oldInstall || keyEQ(terms, oldTerms)) {`.  If we change that to `&&`, then it'll restart if either the installation or the terms change. 

If the change to liquidation were simpler and setup of the contract were simpler, we could set up  the vaultFactory with fake governance, without an AMM and switch from incremental to mock liquidation. The fake governance could  directly change the param values without requiring a vote. The mock liquidation would report that it had been instantiated and called.

But I think we want to actually see the liquidation validating that it got appropriate params. And we already have support for setting up governance. So the test just needs a single-member committee to vote to change the params.

Chris-Hibbert (commented):
The changes look reasonable, but I think we do need a test of changing the liquidation. If you make the update I suggest below, we can test by just changing liquidation terms.

Chris-Hibbert:
I think the test failures will be fixed by integrating with master so you get #5208

packages/run-protocol/src/vaultFactory/vaultDirector.js
dtribble: Yep. That was a bug. It should have been &&

dtribble:
> I think the test failures will be fixed by integrating with master so you get #5208

It does have this integrated. But it also has my bugfix to publish the initial bindings on the subscription. I must not haev re-run the governance tests before submitting. I'm delighted if that's the problem :).

dckc (commented):
I defer to @Chris-Hibbert 
LMK if you need me to act on this.

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -143,6 +143,28 @@ const getCollaterals = async ({ state }) => {
   );
 };
 
+const getLiquidationConfig = directorParamManager => ({
+  install: directorParamManager.getInstallation(LIQUIDATION_INSTALL_KEY),
+  terms: directorParamManager.getUnknown(LIQUIDATION_TERMS_KEY),
```
Chris-Hibbert, line 148: This should work as well, and type-check equivalently.  Your preference.
```suggestion
  install: directorParamManager.getLiquidationInstall(),
  terms: directorParamManager.getLiquidationTerms(),
```

Chris-Hibbert (approved):
This looks good.

In order for voters to be able to assure themselves that the proposed liquidation terms will perform as advertised, I think it's helpful for the subject contract (`liquidateIncrementally.js`) to export a helper function that validates that it would be able to parse them, and be unlikely to hiccup when restarting.  Simple things like ensuring the terms parse, and the ratios have plausible known brands. Otherwise voters have to judge based on human eyes reading the textual form of the json.

This doesn't need to be part of this PR, but it should follow soon after.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: The getters are preferred as string args won't pass jessie-check.

turadg:
> I think it's helpful for the subject contract (liquidateIncrementally.js) to export a helper function

Contracts go on chain. Does the helper function need to be on chain? If not then better to factor that out into another module.

Chris-Hibbert:
> Does the helper function need to be on chain? 

It doesn't need to be on chain, but it needs to be versioned with the contract itself.

dtribble:
> It doesn't need to be on chain, but it needs to be versioned with the contract itself.

https://github.com/Agoric/agoric-sdk/issues/5216

PR #5214, opened 2022-04-25
refactor(vault): make liquidation recursion a loop

dtribble:
refs: #5211

## Description

The liquidation recursion was unbounded, and so could consume a lot of space, depending on implementation details. This should remove that risk.

### Security Considerations

Users could tickle the vulnerability by changing their collateralization ration.

N/A

### Testing Considerations

The existing tests confirm that liquidation still works.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -55,9 +55,6 @@ const start = async zcf => {
     AMMMaxSlippage,
   } = /** @type {LiquidationContractTerms} */ zcf.getTerms();
 
-  // trace('terms', zcf.getTerms());
-  const nextBlock = async () => E(timerService).delay(1n);
-
   const SCALE = 1_000_000n;
   const BASIS_POINTS = 10_000n * SCALE;
   const BP2 = BASIS_POINTS * BASIS_POINTS;
@@ -133,19 +130,38 @@ const start = async zcf => {
   };
 
   /**
+   * A generator that yields once per block, starting in the current block.
+   *
+   * @yields {void}
+   */
+  async function* oncePerBlock() {
+    yield;
+    while (true) {
+      yield E(timerService).delay(1n);
+    }
+  }
+
+  /**
+   * Generate tranches to sell until the debt is paid off.
    *
    * @param {ZCFSeat} seat
    * @param {Amount} originalDebt
    * @yields {LiquidationStep}
    */
   async function* processTranches(seat, originalDebt) {
-    let proceedsSoFar = seat.getAmountAllocated('Out');
-    let toSell = seat.getAmountAllocated('In');
-    while (
-      !AmountMath.isGTE(proceedsSoFar, originalDebt) &&
-      !AmountMath.isEmpty(toSell)
-    ) {
-      const debt = AmountMath.subtract(originalDebt, proceedsSoFar);
+    while (true) {
+      const proceedsSoFar = seat.getAmountAllocated('Out');
+      const toSell = seat.getAmountAllocated('In');
+      if (
+        AmountMath.isGTE(proceedsSoFar, originalDebt) ||
+        AmountMath.isEmpty(toSell)
+      ) {
+        trace('exiting async loop');
+        return;
+      }
+      await oncePerBlock();
+
+      // Determine the max allowed tranche size
       const { Secondary: poolCollateral, Central: poolCentral } = await E(
         amm,
       ).getPoolAllocation(toSell.brand);
@@ -154,9 +170,12 @@ const start = async zcf => {
         MaxImpactBP,
         AMMFeeBP,
       );
-      trace('TRANCHE', asFloat(maxAllowedTranche.value, 100000n));
+      trace('Pool', { poolCentral, poolCollateral });
+      trace('Max tranche', asFloat(maxAllowedTranche.value, 100000n));
       const tranche = AmountMath.min(maxAllowedTranche, toSell);
+
       // compute the expected proceeds from the AMM for tranche
+      const debt = AmountMath.subtract(originalDebt, proceedsSoFar);
       const minAmmProceeds = estimateAMMProceeds(
         poolCentral,
         poolCollateral,
@@ -173,15 +192,9 @@ const start = async zcf => {
       );
       const oracleLimit = computeOracleLimit(oracleQuote, OracleTolerance);
 
-      trace('TRANCHE', {
-        debt,
-        tranche,
-        minAmmProceeds,
-        oracleLimit,
-        poolCentral,
-        poolCollateral,
-      });
+      trace('Tranche', { debt, tranche, minAmmProceeds, oracleLimit });
       if (AmountMath.isGTE(minAmmProceeds, oracleLimit)) {
+        // our prices are within the same range; go ahead and try to sell
         yield {
           collateral: tranche,
           ammProceeds: minAmmProceeds,
@@ -189,18 +202,9 @@ const start = async zcf => {
           oracleLimit,
         };
       } else {
+        // prices are too far apart; don't sell this block
         trace('SKIP');
       }
-      proceedsSoFar = seat.getAmountAllocated('Out');
-      toSell = seat.getAmountAllocated('In');
-      if (
-        AmountMath.isGTE(proceedsSoFar, originalDebt) ||
-        AmountMath.isEmpty(toSell)
-      ) {
-        trace('exiting async loop');
-        return;
-      }
-      await nextBlock();
     }
   }
 
--- a/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -156,7 +156,7 @@ const start = async zcf => {
         AmountMath.isGTE(proceedsSoFar, originalDebt) ||
         AmountMath.isEmpty(toSell)
       ) {
-        trace('exiting async loop');
+        trace('exiting processTranches loop');
         return;
       }
       await oncePerBlock();
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -170,6 +170,20 @@ const initState = (
   return state;
 };
 
+/**
+ * Threshold to alert when the price level falls enough that the vault
+ * with the highest debt to collateral ratio will no longer be valued at the
+ * liquidationMargin above its debt.
+ *
+ * @param {*} highestDebtRatio
+ * @param {*} liquidationMargin
+ */
+const liquidationThreshold = (highestDebtRatio, liquidationMargin) =>
+  ceilMultiplyBy(
+    highestDebtRatio.numerator, // debt
+    liquidationMargin,
+  );
+
 // Some of these could go in closures but are kept on a facet anticipating future durability options.
 const helperBehavior = {
   /**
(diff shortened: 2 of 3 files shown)


packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +254,94 @@ const helperBehavior = {
...
+        );
+        trace('update quote', triggerPoint, highestDebtRatio.denominator);
+      }
+    } else {
+      state.liquidationInProgress = true;
+
+      // eslint-disable-next-line consistent-return
+      return facets.helper
+        .processLiquidations()
+        .catch(e => console.log('Liquidator failed', e))
```
turadg, line 278: is this sufficient logging? something we'd want to instrument?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +254,94 @@ const helperBehavior = {
...
+        liquidationMargin,
       );
-      return;
-    }
 
-    if (state.liquidationInProgress) {
-      return;
+      // Update the current in-progress quote if there is one. Otherwise,
+      // the new threshold will be picked up by the next quote request.
+      if (state.outstandingQuote) {
```
turadg: why calculate `triggerPoint` if there's no `oustandingQuote`?

I'd find this "return early" flow a clearer to follow because there are two reasons _not_ to "reschedule price check":
- if there's no `liquidationInProgress` return promise to start liquidation
- if there's no `outstandingQuote` return 
- (else) actually reschedule the price check

Will approve after aligning on this

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +254,94 @@ const helperBehavior = {
...
-    await (next ? facets.helper.liquidateAndRemove(next) : null);
-    trace('price check liq', next && next[0]);
-
-    // eslint-disable-next-line consistent-return
-    return facets.helper.reschedulePriceCheck();
+  processLiquidations: async ({ state, facets }) => {
+    const { prioritizedVaults, priceAuthority } = state;
+    const govParams = state.factoryPowers.getGovernedParams();
+
+    async function* liquidations() {
```
turadg: this won't jessie-check. not a blocker.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +254,94 @@ const helperBehavior = {
...
+        }
+        const liquidationMargin = govParams.getLiquidationMargin();
+
+        // ask to be alerted when the price level falls enough that the vault
+        // with the highest debt to collateral ratio will no longer be valued at the
+        // liquidationMargin above its debt.
+        const triggerPoint = ceilMultiplyBy(
+          highestDebtRatio.numerator, // debt
+          liquidationMargin,
+        );
```
turadg: duplicated above. worth a little helper that has the doc comment once.
```
/**
 * Threshold to alert when the price level falls enough that the vault
 * with the highest debt to collateral ratio will no longer be valued at the
 * liquidationMargin above its debt.
 */
const liquidationThreshold = (highestDebtRatio, liquidationMargin) => {
    return ceilMultiplyBy(
          highestDebtRatio.numerator, // debt
          liquidationMargin,
        );
}
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +254,94 @@ const helperBehavior = {
...
+        state.outstandingQuote = E(priceAuthority).mutableQuoteWhenLT(
+          highestDebtRatio.denominator, // collateral
+          triggerPoint,
+        );
+        trace('posted quote request', triggerPoint);
+
+        // The rest of this method will not happen until after a quote is received.
+        // This may not happen until much later, when themarket changes.
```
turadg:
```suggestion
        // This may not happen until much later, when the market changes.
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +254,94 @@ const helperBehavior = {
...
+          prioritizedVaults.entriesPrioritizedGTE(quoteRatioPlusMargin);
+        if (next) {
+          yield next;
+        }
+      }
+    }
+    for await (const next of liquidations()) {
+      await facets.helper.liquidateAndRemove(next);
+      trace('price check liq', next && next[0]);
+    }
```
turadg: seeing just this part it looks like once a liquidation is triggered by a drop, all vaults under water will be liquidated regardless of further price movement. I suggest a jsdoc comment on `liquidations()` fn def that it yields one vault per quote resolution. (basically repeating the "will not happen until after a quote is received" aspect )

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: I don't think so, but it's equivalent to the logging that it had. It's also not clear what more it should do.

turadg: How about a StatsD-style counter that we watch?

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: rename `liquidations` to `eventualLiquidations`

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -170,6 +170,20 @@ const initState = (
   return state;
 };
 
+/**
+ * Threshold to alert when the price level falls enough that the vault
+ * with the highest debt to collateral ratio will no longer be valued at the
+ * liquidationMargin above its debt.
+ *
+ * @param {*} highestDebtRatio
+ * @param {*} liquidationMargin
```
turadg: I suppose your editor auto-added these. I'm allergic to `*`/`any` so I'd prefer typing them or omitting them. But we don't have any consensus on that so just expressing a personal opinion.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +268,83 @@ const helperBehavior = {
...
       );
-      trace(
-        'updating level for outstandingQuote',
-        triggerPoint,
-        highestDebtRatio.denominator,
-      );
-      return;
+      trace('update quote', highestDebtRatio);
     }
+  },
```
turadg, line 300: if not doing the above, could be clearer that doing nothing in the else is intentional
```suggestion
  // do nothing; the new threshold will be picked up by the next quote request
  },
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +268,83 @@ const helperBehavior = {
...
-    // ask to be alerted when the price level falls enough that the vault
-    // with the highest debt to collateral ratio will no longer be valued at the
-    // liquidationMargin above its debt.
-    const triggerPoint = ceilMultiplyBy(
-      highestDebtRatio.numerator, // debt
-      liquidationMargin,
-    );
+    // INTERLOCK: the first time through, start the process to process
+    // liquidations over time.
+    if (!state.liquidationInProgress) {
```
turadg: this [return early style](https://medium.com/swlh/return-early-pattern-3d18a41bba8) reads very well here. for below I think the style also makes sense and you don't even have to use a negative condition.

```
if (state.outstandingQuote === null) {
  // the new threshold will be picked up by the next quote request
  return;
}

// There is already an activity processing liquidations. It may be
// waiting for the oracle price to cross a threshold.
// Update the current in-progress quote.
const govParams = state.factoryPowers.getGovernedParams();
const liquidationMargin = govParams.getLiquidationMargin();
```

turadg (approved):

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -133,19 +130,38 @@ const start = async zcf => {
   };
 
   /**
+   * A generator that yields once per block, starting in the current block.
+   *
+   * @yields {void}
+   */
+  async function* oncePerBlock() {
```
turadg, line 137: neat!

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -133,19 +130,38 @@ const start = async zcf => {
...
-    ) {
-      const debt = AmountMath.subtract(originalDebt, proceedsSoFar);
+    while (true) {
+      const proceedsSoFar = seat.getAmountAllocated('Out');
+      const toSell = seat.getAmountAllocated('In');
+      if (
+        AmountMath.isGTE(proceedsSoFar, originalDebt) ||
+        AmountMath.isEmpty(toSell)
+      ) {
+        trace('exiting async loop');
```
turadg: minor:
```suggestion
        trace('exiting processTranches loop');
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -92,6 +90,10 @@ const trace = makeTracer('VM', true);
  * }>} MethodContext
  */
 
+// EPHERMERAL STATE
```
turadg:
```suggestion
// EPHEMERAL STATE
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +268,83 @@ const helperBehavior = {
...
-    // the `outstandingQuote` is assigned. So, if there's an outstanding quote,
-    // reset the level. If there's no current quote (because this is the first
-    // vault, or because a quote just resolved), then make a new request to the
-    // priceAuthority, and when it resolves, liquidate the first vault.
-    if (state.outstandingQuote) {
+    // There is already an activity processing liquidations. It may be
+    // waiting for the oracle price to cross a threshold.
+    // Update the current in-progress quote if there is one. Otherwise,
+    // the new threshold will be picked up by the next quote request.
+    if (outstandingQuote) {
```
turadg: don't know if you saw my last comment here (not being passive aggressive ;-), just thought it might be lost in the rebasing )

This would let the "actually reschedule price check" be unnested and be consistent with the `if (!liquidationInProgress) {` early return style:

```
if (outstandingQuote === null) {
  // Don't schedule. The new threshold will be picked up by the next quote request.
  return;
}
```

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -254,79 +268,83 @@ const helperBehavior = {
...
+        // liquidationMargin above its debt.
+        outstandingQuote = E(priceAuthority).mutableQuoteWhenLT(
+          highestDebtRatio.denominator, // collateral
+          liquidationThreshold(highestDebtRatio, liquidationMargin),
+        );
+        trace('posted quote request', highestDebtRatio);
+
+        // The rest of this method will not happen until after a quote is received.
+        // This may not happen until much later, when the market changes.
+        // eslint-disable-next-line no-await-in-loop
```
turadg, line 325: thoughts on disabling this lint rule?

turadg (approved):

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: Some of the key reasons in the pattern don't apply here (updating the quote is not the happy path), but we have several motivations for eliminating the indent level, so I'll try it.

I note however the `=== null` is a negative test, and is more fragile than testing for falsy.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: I didn't see the last comment previously, but saw it when looking at these comments. Addressed.

packages/run-protocol/src/vaultFactory/vaultManager.js
dtribble: Nope. The `await` here carefully doesn't violate our driving rule, but typical usages would. Given that at least one of the `reschedulePriceCheck` bugs was a bogus `await` violation, it's important to internalize the underlying rule and enforce it pretty rigorously.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: > I note however the === null is a negative test

I thought your concern with "negative test" is the cognitive burden of flipping the logic. This doesn't, right? Or did I misunderstand your concern?

> and is more fragile than testing for falsy.

I think that's a virtue here. If the value is ever a non-null falsy that's a big surprise.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: As part of that rigorous enforcement I suggest that each suppression of the rule be accompanied by an explicit justification. That way the reader won't take away that suppressions are commonplace or that rules are lax.

PR #5223, opened 2022-04-25
test(vault): multiple liquidations

dtribble:
refs: #4568

## Description

Add tests for liquidation

### Security Considerations

N/A

N/A

### Testing Considerations

We don't have enough tests that liquidate multiple vaults in various ways. This adds them.

### DIFF
--- a/packages/run-protocol/test/vaultFactory/test-liquidator.js
+++ b/packages/run-protocol/test/vaultFactory/test-liquidator.js
@@ -325,6 +325,8 @@ async function setupServices(
 // #endregion
 
 // #region driver
+const AT_NEXT = {};
+
 const makeDriver = async (t, initialPrice, priceBase) => {
   const timer = buildManualTimer(t.log);
   const services = await setupServices(t, initialPrice, priceBase, timer);
@@ -343,10 +345,10 @@ const makeDriver = async (t, initialPrice, priceBase) => {
   ).getNotifier();
   let managerNotification = await E(managerNotifier).getUpdateSince();
 
-  /** @type {UserSeat<VaultKit>} */
-  let lastSeat;
+  /** @type {UserSeat} */
+  let currentSeat;
   let notification = {};
-  let lastOfferResult;
+  let currentOfferResult;
   const makeVaultDriver = async (collateral, debt) => {
     /** @type {UserSeat<VaultKit>} */
     const vaultSeat = await E(zoe).offer(
@@ -363,33 +365,37 @@ const makeDriver = async (t, initialPrice, priceBase) => {
       vault,
       publicNotifiers: { vault: notifier },
     } = await E(vaultSeat).getOfferResult();
-    t.truthy(await E(vaultSeat).hasExited());
+    t.true(await E(vaultSeat).hasExited());
     return {
       vault: () => vault,
       vaultSeat: () => vaultSeat,
       notification: () => notification,
       close: async () => {
-        lastSeat = await E(zoe).offer(E(vault).makeCloseInvitation());
-        lastOfferResult = await E(lastSeat).getOfferResult();
+        currentSeat = await E(zoe).offer(E(vault).makeCloseInvitation());
+        currentOfferResult = await E(currentSeat).getOfferResult();
         t.is(
-          lastOfferResult,
+          currentOfferResult,
           'your loan is closed, thank you for your business',
         );
         t.truthy(await E(vaultSeat).hasExited());
       },
-      checkNotify: async (phase, expected) => {
-        notification = await E(notifier).getUpdateSince();
-        trace(t, 'notify', notification);
-        t.is(notification.value.vaultState, phase);
-        expected && t.like(notification.value, expected);
-      },
-      awaitNotify: async (phase, expected) => {
+      /**
+       *
+       * @param {Vault.Phase} phase
+       * @param {Object} likeExpected
+       * @param {undefined|AT_NEXT|number} optSince
+       */
+      notified: async (phase, likeExpected, optSince) => {
+        // optSince can be AT_NEXT in order to wait for the
+        // next update.
         notification = await E(notifier).getUpdateSince(
-          notification.updateCount,
+          optSince === AT_NEXT ? notification.updateCount : optSince,
         );
-        trace(t, 'notify', notification);
         t.is(notification.value.vaultState, phase);
-        expected && t.like(notification.value, expected);
+        if (likeExpected) {
+          t.like(notification.value, likeExpected);
+        }
+        return notification;
       },
       checkBorrowed: async (loanAmount, loanFee) => {
         const debtAmount = await E(vault).getCurrentDebt();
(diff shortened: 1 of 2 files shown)


packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -340,19 +340,73 @@ const makeDriver = async (t, initialPrice, priceBase) => {
...
+      vault,
+      publicNotifiers: { vault: notifier },
+    } = await E(vaultSeat).getOfferResult();
+    t.truthy(await E(vaultSeat).hasExited());
+    return {
+      vault: () => vault,
+      vaultSeat: () => vaultSeat,
+      notification: () => notification,
+      close: async () => {
+        lastSeat = await E(zoe).offer(E(vault).makeCloseInvitation());
```
Chris-Hibbert: I prefer `finalSeat`.  `lastSeat` often means *previous*.

Chris-Hibbert (approved):

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: Ironically, this field was indeed intended to be the seat for the most recent offer, so if there were follow-up assertion tests, it would be available. Thus the next operations might be `checkPayouts(...)` which would go against the `lastSeat`

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -342,19 +342,73 @@ const makeDriver = async (t, initialPrice, priceBase) => {
...
+      }),
+      harden({
+        Collateral: aethMint.mintPayment(collateral),
+      }),
+    );
+    const {
+      vault,
+      publicNotifiers: { vault: notifier },
+    } = await E(vaultSeat).getOfferResult();
+    t.truthy(await E(vaultSeat).hasExited());
```
turadg: it should return a boolean
```suggestion
    t.true(await E(vaultSeat).hasExited());
```

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -342,19 +342,73 @@ const makeDriver = async (t, initialPrice, priceBase) => {
...
+          lastOfferResult,
+          'your loan is closed, thank you for your business',
+        );
+        t.truthy(await E(vaultSeat).hasExited());
+      },
+      checkNotify: async (phase, expected) => {
+        notification = await E(notifier).getUpdateSince();
+        trace(t, 'notify', notification);
+        t.is(notification.value.vaultState, phase);
+        expected && t.like(notification.value, expected);
```
turadg: this expression style is common within larger declarations but I've rarely seen it in imperative. I thought this is preferred here,
```
if (expected) {
  t.like(notification.value, expected);
}
```
Is this an exception for test code or am I just wrong? :)  if so, fyi @samsiegart who I asked recently to remove a use of `&&`.

Either way, I'd argue that this would be better so as to allow testing for falsy expectations,
```
// note that notifiers can send `undefined` as the actual state but this driver doesn't handle that case
if (expected !== undefined) {
  t.like(notification.value, expected);
}
```

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -342,19 +342,73 @@ const makeDriver = async (t, initialPrice, priceBase) => {
...
+      },
+      awaitNotify: async (phase, expected) => {
```
turadg: I like the check/await pair here but it's not obvious from names how they're different. They're both awaited internally so the `await` is a bit misnomer. The different is about waiting for the "next" in a sequence of updates so I'd suggest "next". Also would help tp have the `t.like` conveyed as what comes into `expected`.

So consider, `notifiedLike` and `notifiedNextLike`.

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -364,55 +418,14 @@ const makeDriver = async (t, initialPrice, priceBase) => {
...
-      t.is(notification.value.vaultState, phase);
-      expected && t.like(notification.value, expected);
-    },
-    awaitNotify: async (phase, expected) => {
-      notification = await E(notifier).getUpdateSince(notification.updateCount);
-      trace(t, 'notify', notification);
-      t.is(notification.value.vaultState, phase);
-      expected && t.like(notification.value, expected);
-    },
+    makeVault: makeVaultDriver,
```
turadg: `makeVaultDriver` is more accurate. let's keep that name throughout.

otherwise reading below,
```
const v = await d.makeVault(collateralAmount, loanAmount);
```
looks like `v` is a vault made by the driver whereas it's another driver spawned by the main one to wrap a new vault.

consider also keeping `d` in driver objects. that would help in reading tests in code review where most lines (including the assignment) are hidden
```
const vd = await d.makeVaultDriver(collateralAmount, loanAmount);
```
or perhaps a pathname style for when there's more than one driver of vaults:
```
dv1.checkNotify();
dv2.close();
```

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -342,19 +342,73 @@ const makeDriver = async (t, initialPrice, priceBase) => {
...
+      },
+      awaitNotify: async (phase, expected) => {
+        notification = await E(notifier).getUpdateSince(
+          notification.updateCount,
+        );
+        trace(t, 'notify', notification);
+        t.is(notification.value.vaultState, phase);
+        expected && t.like(notification.value, expected);
+      },
+      checkBorrowed: async (loanAmount, loanFee) => {
```
turadg, line 400: the other "check" functions are void but this one gets a value. consider `getCheckedDebt` or simply `getDebt` as it's expected that the driver will be asserting all over the place.

turadg: oh I see `checkManagerNotifier` also returns a value. same opinion on that, though that might also benefit from a signature change to remove `optUpdateSince` and have the check/await pair style you had for vault driver.

packages/run-protocol/test/vaultFactory/test-liquidator.js
Chris-Hibbert: sorry, my point is that "last" is ambiguous between "final" and "previous". Please say one or the other.

Chris-Hibbert (approved):

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: OK I think that `currentSeat` is actually the most appropriate

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: In all these cases, the notifications are records. So I'm going to keep it focused on "the details of the record". They can always look at the value explicitly, which seems good enough for now.

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: In production code, I generally avoid having a boolean to give alternative behavior. But in this case, I think it would be simpler to have them in a single operation with a simple flag (or eventually optional sequence number). e.g.,
```
      vaultNotified: async (phase, expected, since) => {
        notification = await E(notifier).getUpdateSince(since === true ? notification.updateCount : since);
        t.is(notification.value.vaultState, phase);
        if (expected) {
          t.like(notification.value, expected);
        }
        return notification;
      },
```

packages/run-protocol/test/vaultFactory/test-liquidator.js
dtribble: I'm going the other way with the check/await pattern to have a single operation with all the knobs for the particular thing to check.

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: makes sense for test code. I don't like that there will be two independently optional arguments but again, test code. I would be explicit about that though with a jsdoc since the driver is an API (albeit an internal one).

some ideas for more clarity,
```
vaultNotified: async (phase, expectedLike, awaitNewNotification = false) => {
  const since = awaitNewNotification ? notification.updateCount : undefined;
```

also, falsy values can be expected so,
```
  // `undefined` can be a notifier state but it's rare enough that we use it here to signal an unspecified parameter
  if (expectedLike !== undefined)
```
(edit: recalled I mentioned this separately above)

PR #5224, opened 2022-04-25
test: a virtual object test for AMM pools.

Chris-Hibbert:
refs: #5207

## Description

Pretty rudimentary.  It just sees that the objects come back. The test
fails when I manually remove (tested) stuff from state.

There's one test that changed state is persisted: I update the
notifier, and verify that on revival, the count has increased.

A plausible next test would be to add liquidity, but that's a bunch
more work.

### Security Considerations

Not related to security

Virtualization doesn't need to be exposed to users, though we'll want to explain it to contract developers at some point.

### Testing Considerations

This is a test.


packages/SwingSet/test/liveslots-helpers.js
```diff
@@ -14,13 +14,19 @@ import {
   makeBringOutYourDead,
 } from './util.js';
 
-export function buildSyscall() {
+export function buildSyscall(skipLogging) {
```
turadg: Will take a closer look at the test on Wednesday. Just at a quick glance: why this new option?

Also why change all the `log.push` instead of just changing the thing that reads the array?

packages/SwingSet/test/vo-test-harness/test-dvo-test-harness.js
```diff
@@ -0,0 +1,24 @@
...
+}
+
+async function dvoTestTest(t, mode) {
+  function testLogCheck(_t, phase, log) {
+    t.deepEqual(log, ['start test', phase, 'test thing', { mode }, 'end test']);
+  }
+  await runDVOTest(t, testLogCheck, bfile('vat-dvo-test-test.js'), { mode });
+}
+
+test.failing('fail during "before" phase', async t => {
```
turadg: I think you want a passing test that expects a throw.

packages/SwingSet/test/vo-test-harness/test-vo-test-harness.js
```diff
@@ -0,0 +1,52 @@
+import { test, runVOTest } from '../../tools/vo-test-harness.js';
+
+async function voTestTest(t, mode) {
+  let makeThing;
+
+  // eslint thinks `retainer` is unused, but eslint is very wrong
```
turadg: Should we disable the rule? Typescript gets it right.

packages/SwingSet/tools/vo-test-harness.js
```diff
@@ -0,0 +1,132 @@
...
+  await dispatchMessage('dropHeld');
+  await dispatchMessage('fetchAndHold');
+  await dispatchMessage('testHeld', capargs(['after']));
+}
+
+function bfile(name) {
+  return new URL(name, import.meta.url).pathname;
+}
+
+const stupidM = makeMarshal(
```
turadg: "Fake"?

packages/SwingSet/test/vo-test-harness/test-dvo-test-harness.js
FUDCo: Not relevant in this case.

packages/SwingSet/test/liveslots-helpers.js
FUDCo: You should make these comments on #5207, which is the PR containing the commits you're comment on here.

It's so the array can be absent when it's not being used, which could be significant for a new class of tests that aren't ever going to look at the log contents (which record _all_ syscall activity), especially for tests that are large or long running.

packages/SwingSet/test/vo-test-harness/test-vo-test-harness.js
FUDCo: The value is actually unused in the sense that it's never actually made use of, but the assignment pins the reference in memory for GC purposes which is why it's there.  How would TypeScript know that's what's going on here?

packages/SwingSet/tools/vo-test-harness.js
FUDCo: Possibly, but I prefer "stupid" because it respects what's being done here less.

Chris-Hibbert:
Hmm.  I changed the base PR so this would be relative to [Chip's PR](https://github.com/Agoric/agoric-sdk/pull/5207). I don't know why github is showing both sets of changes. The goal is that after Chip's PR adds the test harness, this PR would add one test for the AMM.

I'm going to [try to] mark this draft and if I can't figure out how to repair it, I'll close it and start over.

Chris-Hibbert:
Okay, that seems to have worked.  @turadg please move your comments over to #5207.  This PR no longer includes any of those changes.

packages/run-protocol/test/test-voPool.js
```diff
@@ -0,0 +1,89 @@
...
+  const quoteIssuerKit = makeIssuerKit('Quotes');
+  const liquidityZcfMint = await zcf.makeZCFMint('Liquidity');
+  const { userSeat: poolSeatP } = zcf.makeEmptySeatKit();
+  const { zcfSeat: protocolSeatP } = zcf.makeEmptySeatKit();
+  const [poolSeat, protocolSeat] = await Promise.all([
+    poolSeatP,
+    protocolSeatP,
+  ]);
+
+  const makeVirtualPoolKind = () => {
```
turadg:
```suggestion
  const defineVirtualPoolKind = () => {
```

packages/run-protocol/test/test-voPool.js
```diff
@@ -0,0 +1,89 @@
...
+
+test.serial('one update', async t => {
+  await voPoolTest(
+    t,
+    context => {
+      return context.pool.updateState();
+    },
+    async context => {
+      const notification = await context.pool.getNotifier().getUpdateSince();
+      t.is(2, notification.updateCount);
```
turadg: Ava assertions take _actual_ first and _expected_ second.

packages/run-protocol/test/test-voPool.js
```diff
@@ -0,0 +1,89 @@
...
+      state.toCentralPA = pool.getToCentralPriceAuthority();
+      state.singlePool = pool.getVPool();
+      state.liquidityIssuer = pool.getLiquidityIssuer();
+      mutation(context);
+    } else if (phase === 'after') {
+      const newNotifier = pool.getNotifier();
+      t.is(newNotifier, state.notifier);
+      t.is(pool.getToCentralPriceAuthority(), state.toCentralPA);
+      t.is(singlePool, state.singlePool);
+      t.is(pool.getLiquidityIssuer(), state.liquidityIssuer);
```
turadg: Ava assertions take actual first and expected second. `state.*` are the ones being tested for actuality.

turadg (approved):

packages/run-protocol/test/test-voPool.js
Chris-Hibbert: done

packages/run-protocol/test/test-voPool.js
Chris-Hibbert: thanks. Done.

packages/run-protocol/test/test-voPool.js
Chris-Hibbert: done

turadg (approved):

Issue #5036, opened 2022-04-07
Zoe supports Snoozing specific incoming offers for invitations


Tartuffo:
When RUN Protocol is launched, we may face a situation where there are bugs or attacks on contracts that pose economic danger to individual users or the entire chain economy.

We considered the following options for handling these kinds of situations:

- "Hammer": for a contract instance, Zoe rejects offers for invitations whose descriptionString is in the snooze list
- "Wrench": for a contract instance, Zoe rejects all reallocates that use a brand that is in the snooze list
- "Screwdriver": for a contract instance, Zoe rejects all offers for a particular internal “partition” (e.g., pool or vault or account)

We've decided to implement the "Hammer" solution for MN-1.

The appropriate Governance committee can vote to "snooze" incoming offers for invitations from making it to the target contract.   During the snooze period, those offers for invitations will be rejected, with an appropriate error message.

On the Zoe side, we could leverage or extend the stopAcceptingOffers capability granted to Zoe in zoeService/startInstance.js.

This issue covers the code implementation to support this feature.  Another ticket should be created to cover the governance workflow and documentation.

Tartuffo:
@rowgraus Please create a ticket for the Governance side of this.

Tartuffo:
Design here: https://drive.google.com/drive/u/0/folders/13MifhaEn0slybFA-_bD_xMDOFJc2kJLV

Chris-Hibbert:
Discussed at Zoe meeting 2022/4/27.  The accepted approach is via contract governance and zcf.

1. add a method to ZCF to allow the contract to specify a list of strings. Attempts to exercise invitations whose description match a string on the list will error.
2. Add support for governance to be able to update that list.
3. (#4142) Also add a question posing contract to guide and constrain the Econ committee in creating governance questions for the contracts they control.

implications: 

A. Zoe remains ignorant of governance
B. The contract's electorate can make these changes, which also means that the economic committee has this power for the core economic contracts.

Chris-Hibbert:
Fullfilled by #5749

PR #5225, opened 2022-04-26
fix: fix types of stopAcceptingOffers

erights:
refs: #5036 (Found while exploring it)

zcf.stopAcceptingOffers was declared as returning `void` but it actually returns a `Promise<void>` (or should given the other fix explained below). The test in test-zcf.js `await`s it, so is clearly relying on it to return a promise.

After all the delegation, the actual implementation is declared as returning `void`, which seems correct to me. However, the implementation was returning `false`. In general, when something is declared as returning `void` we should be careful to return only `undefined`, as anything else could be a capability leak. Though it is not in this case. Since it returns `false` unconditionally, it is not even an information leak.

@turadg @michaelfig @mhofman Shouldn't TS have caught both of these problems? Do you understand why it did not?

### DIFF
--- a/packages/zoe/src/contractFacet/types.js
+++ b/packages/zoe/src/contractFacet/types.js
@@ -43,7 +43,7 @@
  * @property {ZCFRegisterFeeMint} registerFeeMint
  * @property {ZCFMakeEmptySeatKit} makeEmptySeatKit
  * @property {SetTestJig} setTestJig
- * @property {() => void} stopAcceptingOffers
+ * @property {() => Promise<void>} stopAcceptingOffers
  * @property {() => Instance} getInstance
  */
 
--- a/packages/zoe/src/zoeService/startInstance.js
+++ b/packages/zoe/src/zoeService/startInstance.js
@@ -97,7 +97,9 @@ export const makeStartInstance = (
           acceptingOffers = false;
           zoeSeatAdmins.forEach(zoeSeatAdmin => zoeSeatAdmin.fail(reason));
         },
-        stopAcceptingOffers: () => (acceptingOffers = false),
+        stopAcceptingOffers: () => {
+          acceptingOffers = false;
+        },
         makeUserSeat: (
           invitationHandle,
           initialAllocation,


mhofman:
> Shouldn't TS have caught both of these problems? Do you understand why it did not?

Speaking only regarding the `false` return value for a return type of `void`, I believe it depends on how the type is checked.

A function returning something is always assignable to a variable with a function returning nothing `(...args: any[]) => void`. It's one of the [FAQ entries](https://github.com/microsoft/TypeScript/wiki/FAQ#why-are-functions-returning-non-void-assignable-to-function-returning-void). Basically `void` in this case means "I don't care what the return value is".

Now I believe there is a way to trigger a type error in some case. I think when an object/class has methods which differ in return types from what is expected, I've seen TS in some cases report an error.

Chris-Hibbert (approved):
lgtm

turadg:
> A function returning something is always assignable to a variable with a function returning nothing (...args: any[]) => void.

Well not exactly.  [Testing in the playground](https://www.typescriptlang.org/play?filetype=js#code/LAKA9AVBoAQTABAxgQwDZoEYqQaxgGoD2AlgCYDC6WOus8CATgKYAuArowHYDOMA3gDdSZAL70woUEiK9WMAGa8YAXgGwYkBiw7c+QkaLiSQMGMwAeABzQkkJVgC4YACgCUqgHyL0PZgBoNLURWAE8rZgFicioMbDwjCBMzVDjaABVw5md3Lx80P0CQcRBQLXoQrLJmBQEAeUwAK2YkViMAZSIAW2YGxoqEK0YiCMYwgVyVb2FyI1SaPAAFYdGwiSlwKAGwiIFOnr6S42lZHnkebt6m1XVTGHn43GWR5jHQnI8p-MLQEtAlHgAOgetGeq1C7kBgnQ7F6CncAG5QEA) shows that's true for `@property` fn definitions but not for named `@callback` types or explicit types on implementation functions.

I think this further supports keeping type definitions at the implementation sites and defining separate types only when needed as interfaces that can be implemented in multiple places.

erights:
How should we type return types so that implicit return (which actually returns `undefined`) passes, but explicitly returning a non-`undefined` fails?

turadg:
> How should we type return types so that implicit return (which actually returns undefined) passes, but explicitly returning a non-undefined fails?

By my test in the playground link, I think the over-permissiveness is only with `@property` on a typedef. ([regardless of whether it's a named callback type](https://www.typescriptlang.org/play?filetype=js#code/LAKA9AVBoAQTABAxgQwDZoEYqQaxgGoD2AlgCYDC6WOus8CATgKYAuArowHYDOMA3gDdSZAL70woUEiK9WMAGa8YAXgGwYkBiw7c+QkaLiSQMGMwAeABzQkkJVgC4YACgCUqgHyL0PZgBoNLURWAE8rZgFicioMbDwjCBMzVDjaABVw5md3Lx80P0CQcRBQLXoQrLJmBQEAeUwAK2YkViMAZSIAW2YGxoqEK0YiCMYwgVyVb2FyIxQuWViaPAAFYdGwgaGR5jHQqJEl+NwjLhQeympjtZ29iSlwKAGwiIFOnr6S42lZHnkebq9JqqdSmGDzRZXWg3DahHIeKb5QoaM4XI7Q9a7MLwvIKXwBUAlUAAj5NAB0EK46NWmL27jJgnQ7F6CncAG5iYC+mTUcxLmkabcwvTGWhmXVWW4OSAgA))

So "keeping type definitions at the implementation sites" will suffice to get what you're asking for. I.e. in this PR,
```js
        /** @type {() => void} */
        stopAcceptingOffers: () => (acceptingOffers = false),
```
That would have errored. And then the `InstanceAdmin` type could be defined using the `const instanceAdmin` value. I notice there's a `makeInstanceAdmin` function so I'd recommend,
```
/** @typedef {ReturnType<typeof makeInstanceAdmin>} InstanceAdmin */
```

I think we should be following that pattern generally. For every `makeSomeType` we have a type `SomeType` which is the return value of the maker. This is documented in our (nascent unvetted) [TypeScript Best Practices](https://docs.google.com/document/d/1bRIUeOokKcbIHZHt68d9kyiJOB5YEUQVpLpFAmDLkhk/edit#heading=h.6lmefd8kl5rm).

Issue #5226, opened 2022-04-26
Non-deterministic bootstrap, swingset sensitivity to config.bundles insertion order


mhofman:
## Describe the bug
It seems that the cause of transient failures in the [deployment integration CI job](https://github.com/Agoric/agoric-sdk/actions/workflows/deployment-test.yml) is due to bundles sometimes being added in a different order.

If I had a guess, it'd be that multiple bundles are loaded in parallel during bootstrap, which results in a filesystem race with a different resolution order.

## To Reproduce
No consistent reproduction. I do have a full trace however.

## Expected behavior
Deterministic bootstrap

## Platform Environment
 - what OS are you using? Debian on WSL2

## Screenshots
```diff
--- validator0-swingstore-trace 2022-04-25 19:22:38.390978962 -0700
+++ chain-stage-0-swingstore-trace      2022-04-25 19:22:30.351551207 -0700
@@ -34,6 +34,4 @@
 set bundle.b1-fbbb4cca1bafdacb2d62b8c758687e26e6a26e7b2a8e52668bea7950e1914d5db53e38873dc70b15f7a94fa3258c843164400ec244ce7694add59372389648ef {"moduleFormat":"endoZipBase64","endoZipBase64":"UEsDBAo
 set bundle.b1-4d1ecba749269094203571b57b2c29623c9e3538d8ff1f81619936f949933f52bcd1b8e6a5048a521807c36e7290f1bd61ce689573c505b9306efc903202e779 {"moduleFormat":"endoZipBase64","endoZipBase64":"UEsDBAo
-set bundle.b1-b7c184daa30a846c4d37828a8538692e5f9f516d4955276ff09323d2b213229e611266e5acd35753ab6f87c0baa076546ce3aaf97be8d738c7746a3cf7afeccc {"moduleFormat":"endoZipBase64","endoZipBase64":"UEsDBAo
-set bundle.b1-0e514260bc8b97f9285128d043c25f7efe2d599495bbc885e0bf10ca55c0b4604a3b6cae3323d044b1c426cde2bfd0f2bdcb0a379f1d924ae67e5eb3b3a35c90 {"moduleFormat":"endoZipBase64","endoZipBase64":"UEsDBAo
[...]

PR #5231, opened 2022-04-26
fix(extract-proposal): make results deterministic

michaelfig:
refs: #5226

## Description

Make `extractCoreProposalBundles` results deterministic.  Fortunately, this was not used in production, but it did result in a test flake.

This code is very tight.  The ordering of assignment of `bundleID`s and the hardening of `bundleHandles` for `deeplyFulfilled` to work was tricky to get right.

### Testing Considerations

Tested locally on sim-chain.  That will also run in CI, and loadgen tests should verify this as well.


packages/cosmic-swingset/src/extract-proposal.js
```diff
@@ -97,25 +96,60 @@ export const extractCoreProposalBundles = async (
...
+          if (a > b) {
+            return 1;
+          }
+          return 0;
+        })
+        .forEach((handle, j) => {
+          handle.bundleID = `coreProposal${i}_${j}`;
+          harden(handle);
+          bundleSpecs.push([handle.bundleID, { sourceSpec: handle.source }]);
+        });
```
mhofman: Why not do a `.map` ?

```suggestion
      // Add the proposal bundle handles in sorted order.
      const bundleSpecs = [...bundleHandles.values()]
        .sort(({ source: a }, { source: b }) => {
          if (a < b) {
            return -1;
          }
          if (a > b) {
            return 1;
          }
          return 0;
        })
        .map((handle, j) => {
          handle.bundleID = `coreProposal${i}_${j}`;
          harden(handle);
          return [handle.bundleID, { sourceSpec: handle.source }];
        });
```

packages/cosmic-swingset/src/extract-proposal.js
```diff
@@ -58,20 +58,18 @@ export const extractCoreProposalBundles = async (
...
     .then(stbuf => (stbuf.isDirectory() ? dirname : path.dirname(dirname)));
 
-  const bundleToSource = new Map();
-  const bundleHandles = new Set();
-  const makeCPArgs = await Promise.all(
+  const extracted = await Promise.all(
     coreProposals.map(async (initCore, i) => {
+      const bundleToSource = new Map();
+      const bundleHandles = new Set();
+
```
mhofman: Is it possible that different proposals may share a bundle?
From what I gather, before `bundleToSource` was shared, so if the same absolute bundle path came up in multiple proposals, the source had to match.

Since `bundleToSource` doesn't seem to impact output and is just there to ensure there are no collisions, is there a reason to not keep it shared across proposals?

packages/cosmic-swingset/src/extract-proposal.js
```diff
@@ -97,25 +96,60 @@ export const extractCoreProposalBundles = async (
...
+          handle.bundleID = `coreProposal${i}_${j}`;
+          harden(handle);
+          bundleSpecs.push([handle.bundleID, { sourceSpec: handle.source }]);
+        });
 
+      // Now that we've assigned all the bundleIDs and hardened the handles, we
+      // can extract the behavior bundle.
+      const { sourceSpec, getManifestCall } = await deeplyFulfilled(
+        harden(proposal),
+      );
```
mhofman, line 122: I find it unfortunate we need to jump through hoops because `deeplyFulfilled` automatically `harden`s.

packages/cosmic-swingset/src/extract-proposal.js
```diff
@@ -97,25 +96,60 @@ export const extractCoreProposalBundles = async (
...
       const absSrc = pathResolve(initDir, sourceSpec);
-      const bundleHandle = harden({
+      const behaviorBundleHandle = harden({
         bundleID: `coreProposal${i}_behaviors`,
         source: absSrc,
       });
-      bundleHandles.add(bundleHandle);
+      bundleHandles.add(behaviorBundleHandle);
```
mhofman: Why do we care about adding this to `bundleHandles`?

mhofman (approved):
I think I've managed to follow the flow of the different pieces of information to gain confidence that this indeed will result in a deterministic order.

I do find the different transformations of the data into spec + ref&call merged together then split back up somewhat confusing. I can't help wondering if there may be a clearer approach, but it's probably not worth spending more time on this.

packages/cosmic-swingset/src/extract-proposal.js
michaelfig: It's not that `deeplyFulfilled` hardens, it's that it requires hardened input.  But yes, it's unfortunate.

PR #5233, opened 2022-04-26
bankers rounding

### LINKED ISSUE #4573: Adopt banker's rounding for interest calculations

As yet we have two rounding methods for integer division (ceil and floor). Because of the repeated division operations in how https://github.com/Agoric/agoric-sdk/pull/4527 tracks compounded interest, those strategies produce measurable skew.

## Description of the Design

A rounding strategy without bias toward/away from zero is [Round half to even](https://en.wikipedia.org/wiki/Rounding#Round_half_to_even). (It is also free of positivity/negativity bias, which isn't relevant to interest calculation but would have other applications.)

This strategy is also known as *bankers' rounding* and is the default rounding mode used in IEEE 754 operations for results in binary floating-point formats.

## Security Considerations
---

## Test Plan

---

turadg:
closes: #4573

## Description

Adds helpers for half-to-even rounding strategy that is the default in IEEE 754. I named them simply `divideBy` and `multiplyBy` since they are the default strategy.

Changes `quantize` to use this half-to-even strategy.

### Security Considerations

--

--

### Testing Considerations

Tests updated.


packages/zoe/test/unitTests/contractSupport/test-ratio.js
```diff
@@ -351,12 +369,12 @@ test('ratio - quantize', t => {
     [12345n, 12345n, 10n ** 15n, 10n ** 15n],
 
     [12345n, 123n, 100365854n, 10n ** 6n],
-    [12345n, 123n, 10036586n, 10n ** 5n],
+    [12345n, 123n, 10036585n, 10n ** 5n],
```
turadg, line 381: now each value here is the simple rounding of the line above divided by 10. (`100365854 / 10` should round to `10036585` not `100365856`)

packages/zoe/src/contractSupport/safeMath.js
```diff
@@ -19,5 +19,26 @@ export const natSafeMath = harden({
...
+    // if remainder > half divisor, should have rounded up instead of down, so add 1
+    if (rem * 2n > b) {
+      return div + 1n;
+    } else if (rem * 2n === b) {
+      // Add 1 if result is odd to get an even return value
+      if (div % 2n === 1n) return div + 1n;
+    }
+    return div;
+  },
+  isGTE: (x, y) => Nat(x) >= Nat(y),
```
michaelfig, line 43: Care to comment on the drive-by `isGTE` change?  It seems obvious, but I'd like to know the rationale.

michaelfig (commented):
LGTM, but I will defer to @erights for this change to contractSupport.

packages/zoe/src/contractSupport/safeMath.js
turadg: To fulfill the description of `natSafeMath` that it validates inputs.
https://github.com/Agoric/agoric-sdk/blob/067ea32b069596202d7f8e7c5e09d5ea7821f6b2/packages/zoe/src/contractSupport/safeMath.js#L9-L10

packages/zoe/src/contractSupport/ratio.js
```diff
@@ -8,7 +8,8 @@ import { isNat } from '@agoric/nat';
 
 import { natSafeMath } from './safeMath.js';
 
-const { multiply, floorDivide, ceilDivide, add, subtract } = natSafeMath;
+const { multiply, floorDivide, ceilDivide, bankersDivide, add, subtract } =
+  natSafeMath;
```
Chris-Hibbert, line 12: Please update the last paragraph of the comment below.

packages/zoe/test/unitTests/contractSupport/test-ratio.js
```diff
@@ -313,18 +321,28 @@ test('add ratios', t => {
   );
 });
 
-test('ratio - complement', t => {
```
Chris-Hibbert, line 316: Why drop the tests for oneMinus?

Chris-Hibbert (commented):
Getting distracted by other things. Here's what I have so far.

packages/zoe/test/unitTests/contractSupport/test-ratio.js
turadg: I misread that as a test of two thirds. Restored.

Chris-Hibbert (approved):

PR #5234, opened 2022-04-26
refactor(ertp): makePaymentMaker → definePaymentKind

### LINKED ISSUE #5018: ERTP payment will need proper TS type def for new VO API

PR #4962, when it lands, introduces a revised API for virtual objects.
This required updating the existing VOs to match the API changes, which that PR does.
However, it does not properly update all of the corresponding JSDoc TS type annotations, on account of non-obviousness.

In particular, `packages/ERTP/src/payment.js` contains the comment:
```
  // XXX the following type cast is meatball surgery to make tsc shut up
  // somebody who understands this should do it properly
```
Somebody who understands this should do it properly.

turadg:
closes: #5018

## Description

Most of the work for #5018 happened already so the particular makePayment fix was trivial. ce35937453692be9a3830c608cea8bdfe24b32e1

While looking at the code it occurred to me that `make…Maker` could be more specific since the `makePaymentMaker` is really a `defineKind` helper. So f60b3dabf18581519d7744be2857a7a54a05b1e6 renames it to `definePaymentKind`.

### Security Considerations

--

Reviewer, is the `defineFooKind` naming for helpers good? Shall I document that practice?

### Testing Considerations

Refactor, existing tests.


packages/ERTP/src/payment.js
```diff
@@ -8,12 +8,10 @@ import { defineKind } from '@agoric/vat-data';
  * @param {Brand<K>} brand
  * @returns {() => Payment<K>}
  */
-export const makePaymentMaker = (allegedName, brand) => {
+export const definePaymentKind = (allegedName, brand) => {
```
turadg, line 11: @FUDCo why is it safe for the kind's behavior to reference this parameter of its enclosing scope? To make this kind durable would `brand` have to go on `state`?

Chris-Hibbert (approved):
The renaming is fine. 

It seems fine to declare that `defineFooKind` is the canonical way to name these functions.

I don't have an opinion on what it takes to make durable objects, or whether this code needs to get that far.

packages/ERTP/src/payment.js
FUDCo: Yes, for durability, `brand` would have to go on `state`. Or, at the very least, there'd need to be some persistent way to recover `brand` from the payment.  Depending on the context, a reasonable alternative might be a durable map store that mapped from kind handle to the brand, which could save you from lots of redundant stores of `brand` at the expense of a little bit of extra complication to extract it.

This does raise an interesting API design question, though, which is whether we should provide a way to go from a durable virtual object to its kind handle.  I can imagine that being useful but I could also imagine there might be ocap issues.

FUDCo: It also strikes me as possible that if the invoking context that sets up the kind for a given payment type always passes in the same brand for that type with each incarnation of the vat, then the above code would actually be OK as is for durable kinds.  The key is that each incarnation of the vat already needs code to reassociate each of its existing kinds with their behavior when the vat is restarted, and that code might very well be in a position to perform the kind handle to brand mapping without any special API magic

PR #5235, opened 2022-04-27
feature!: impose a minimum loan size when opening a vault

### LINKED ISSUE #4967: Vault creation should have a minimum collateral
Per discussion in wallet meeting, creating a vault should have a governance-controlled minimum collateral parameter at the asset level, which would be enforced only on creation.  

Ideal scenario would be ~$5-$10 range as initial value

Chris-Hibbert:
closes: #4967

## Description

specified by a new governed parameter of the vaultDirector, meaning
the amount is common across all collateral types.

### Security Considerations

Nothing notable. Governance.

It returns a coherent error message, so no additional documentation necessary.

### Testing Considerations

Had to update many tests, so I could tell the constraint was binding. I added one additional test that fails because of the limit.

### DIFF
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -44,6 +44,7 @@ const CENTRAL_DENOM_NAME = 'urun';
  *   vaultFactoryCreator: VaultFactory,
  *   vaultFactoryGovernorCreator: GovernedContractFacetAccess<unknown>,
  *   vaultFactoryVoteCreator: unknown,
+ *   minInitialDebt: NatValue,
  * }>} EconomyBootstrapSpace
  */
 
@@ -226,6 +227,7 @@ export const setupReserve = async ({
  * @param { EconomyBootstrapPowers } powers
  * @param {object} config
  * @param { LoanTiming } [config.loanParams]
+ * @param {Amount} minInitialLoan
  */
 export const startVaultFactory = async (
   {
@@ -251,6 +253,7 @@ export const startVaultFactory = async (
       recordingPeriod: SECONDS_PER_DAY,
     },
   } = {},
+  minInitialLoan,
 ) => {
   const installations = await Collect.allValues({
     VaultFactory,
@@ -298,7 +301,9 @@ export const startVaultFactory = async (
     vaultManagerParams,
     ammPublicFacet,
     liquidationDetailTerms(centralBrand),
+    minInitialLoan,
   );
+
   const governorTerms = harden({
     timer,
     electorateInstance,
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -227,7 +227,7 @@ export const setupReserve = async ({
  * @param { EconomyBootstrapPowers } powers
  * @param {object} config
  * @param { LoanTiming } [config.loanParams]
- * @param {bigint} minInitialLoan
+ * @param {bigint} minInitialDebt
  */
 export const startVaultFactory = async (
   {
@@ -253,7 +253,7 @@ export const startVaultFactory = async (
       recordingPeriod: SECONDS_PER_DAY,
     },
   } = {},
-  minInitialLoan = 5_000_000n,
+  minInitialDebt = 5_000_000n,
 ) => {
   const installations = await Collect.allValues({
     VaultFactory,
@@ -301,7 +301,7 @@ export const startVaultFactory = async (
     vaultManagerParams,
     ammPublicFacet,
     liquidationDetailTerms(centralBrand),
-    AmountMath.make(centralBrand, minInitialLoan),
+    AmountMath.make(centralBrand, minInitialDebt),
   );
 
   const governorTerms = harden({
--- a/packages/run-protocol/src/econ-behaviors.js
+++ b/packages/run-protocol/src/econ-behaviors.js
@@ -227,7 +227,7 @@ export const setupReserve = async ({
  * @param { EconomyBootstrapPowers } powers
  * @param {object} config
  * @param { LoanTiming } [config.loanParams]
- * @param {Amount} minInitialLoan
+ * @param {bigint} minInitialLoan
  */
 export const startVaultFactory = async (
   {
@@ -253,7 +253,7 @@ export const startVaultFactory = async (
       recordingPeriod: SECONDS_PER_DAY,
     },
   } = {},
-  minInitialLoan,
+  minInitialLoan = 5_000_000n,
 ) => {
   const installations = await Collect.allValues({
     VaultFactory,
@@ -301,7 +301,7 @@ export const startVaultFactory = async (
     vaultManagerParams,
     ammPublicFacet,
     liquidationDetailTerms(centralBrand),
-    minInitialLoan,
+    AmountMath.make(centralBrand, minInitialLoan),
   );
 
   const governorTerms = harden({
--- a/packages/run-protocol/src/vaultFactory/params.js
+++ b/packages/run-protocol/src/vaultFactory/params.js
@@ -21,16 +21,23 @@ export const INTEREST_RATE_KEY = 'InterestRate';
 export const LOAN_FEE_KEY = 'LoanFee';
 export const LIQUIDATION_INSTALL_KEY = 'LiquidationInstall';
 export const LIQUIDATION_TERMS_KEY = 'LiquidationTerms';
+export const MIN_INITIAL_LOAN_KEY = 'MinInitialLoan';
 
 /**
  * @param {Amount} electorateInvitationAmount
  * @param {Installation} liquidationInstall
+<<<<<<< HEAD
  * @param {object} liquidationTerms
+=======
+ * @param {Object} liquidationTerms
+ * @param {Amount} minInitialLoan
+>>>>>>> 620ac5802 (feature!: impose a minimum loan size when opening a vault)
  */
 const makeVaultDirectorParams = (
   electorateInvitationAmount,
   liquidationInstall,
   liquidationTerms,
+  minInitialLoan,
 ) => {
   return harden({
     [CONTRACT_ELECTORATE]: {
@@ -44,6 +51,10 @@ const makeVaultDirectorParams = (
     [LIQUIDATION_TERMS_KEY]: {
       type: ParamTypes.UNKNOWN,
       value: liquidationTerms,
+    },
+    [MIN_INITIAL_LOAN_KEY]: {
+      type: ParamTypes.AMOUNT,
+      value: minInitialLoan,
     },
   });
 };
@@ -77,18 +88,21 @@ export const vaultParamPattern = M.split(
  * @param {Invitation} electorateInvitation
  * @param {Installation} liquidationInstall
  * @param {object} liquidationTerms
+ * @param {Amount} minInitialLoan
  */
 const makeVaultDirectorParamManager = async (
   zoe,
   electorateInvitation,
   liquidationInstall,
   liquidationTerms,
+  minInitialLoan,
 ) => {
   return makeParamManager(
     {
       [CONTRACT_ELECTORATE]: [ParamTypes.INVITATION, electorateInvitation],
       [LIQUIDATION_INSTALL_KEY]: [ParamTypes.INSTALLATION, liquidationInstall],
       [LIQUIDATION_TERMS_KEY]: [ParamTypes.UNKNOWN, liquidationTerms],
+      [MIN_INITIAL_LOAN_KEY]: [ParamTypes.AMOUNT, minInitialLoan],
     },
     zoe,
   );
@@ -103,6 +117,7 @@ const makeVaultDirectorParamManager = async (
  * @param {VaultManagerParamValues} vaultManagerParams
  * @param {XYKAMMPublicFacet} ammPublicFacet
  * @param {object} liquidationTerms
+ * @param {Amount} minInitialLoan
  * @param {bigint=} bootstrapPaymentValue
  */
 const makeGovernedTerms = (
@@ -114,25 +129,27 @@ const makeGovernedTerms = (
   vaultManagerParams,
   ammPublicFacet,
   liquidationTerms,
+  minInitialLoan,
   bootstrapPaymentValue = 0n,
 ) => {
-  const timingParamMgr = makeParamManagerSync({
+  const loanTimingParams = makeParamManagerSync({
     [CHARGING_PERIOD_KEY]: ['nat', loanTiming.chargingPeriod],
     [RECORDING_PERIOD_KEY]: ['nat', loanTiming.recordingPeriod],
-  });
+  }).getParams();
 
-  const vmParamMgr = makeVaultParamManager(vaultManagerParams);
+  const loanParams = makeVaultParamManager(vaultManagerParams).getParams();
 
   return harden({
     ammPublicFacet,
     priceAuthority,
-    loanParams: vmParamMgr.getParams(),
-    loanTimingParams: timingParamMgr.getParams(),
+    loanParams,
+    loanTimingParams,
     timerService,
     governedParams: makeVaultDirectorParams(
       invitationAmount,
       liquidationInstall,
       liquidationTerms,
+      minInitialLoan,
     ),
     bootstrapPaymentValue,
   });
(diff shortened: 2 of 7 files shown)


packages/run-protocol/src/econ-behaviors.js
```diff
@@ -44,6 +44,7 @@ const CENTRAL_DENOM_NAME = 'urun';
  *   vaultFactoryCreator: VaultFactory,
  *   vaultFactoryGovernorCreator: GovernedContractFacetAccess<unknown>,
  *   vaultFactoryVoteCreator: unknown,
+ *   minInitialLoan: bigint,
```
turadg: Terminology?

packages/run-protocol/src/vaultFactory/params.js
```diff
@@ -45,6 +48,10 @@ const makeVaultDirectorParams = (
       type: ParamTypes.UNKNOWN,
       value: liquidationTerms,
     },
+    [MIN_INITIAL_LOAN_KEY]: {
+      type: ParamTypes.AMOUNT,
```
turadg, line 52: Is there any risk in this being a NAT? The brand is implied.

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -101,12 +101,24 @@ const makeVaultInvitation = ({ state }) => {
...
     const { brand: brandIn } = collateralAmount;
     assert(
       collateralTypes.has(brandIn),
       X`Not a supported collateral type ${brandIn}`,
     );
+
+    assert(
+      AmountMath.isGTE(
+        borrowAmount,
+        // @ts-ignore VaultDirectoryParams isn't right
```
turadg: ts-expect-error

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -48,17 +56,17 @@ export const start = async (zcf, privateArgs) => {
   }));
 
   const {
-    // @ts-expect-error
     [LIQUIDATION_INSTALL_KEY]: { value: liqInstall },
```
turadg, line 59: 😃

Chris-Hibbert:
@turadg  I improved the typescript in `vaultFactory.js`, getting rid of some `@ts-expect-error` markers, but wasn't able to make `vaultDirector.js` happy, so there are a couple of them there. If you see how to fix it, I'd appreciate an update.

turadg (approved):
Approving assuming you’ll address comments before merge.

I can fix the types tomorrow.

packages/run-protocol/src/vaultFactory/params.js
Chris-Hibbert: It could be a NAT, but I like the clarity of specifying it in RUN. There was discussion in the issue of whether the constraint was on the collateral amount or the loan amount, and specifying it in RUN makes it clear that it's the latter.

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: done in parallel to your review.

packages/run-protocol/src/vaultFactory/params.js
Chris-Hibbert: I ended up changing it in econ-behaviors so it could be defaulted. The brand isn't available during initialization.  This meant the tests had to change to bigints, but the contract still takes an Amount.

packages/run-protocol/src/econ-behaviors.js
dtribble: Seems like this should at least be a `nat` is not an Amount.

packages/run-protocol/src/vaultFactory/params.js
dtribble: The debt brand is available here, so I think having this be an Amount is the right thing.

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -59,7 +59,7 @@ const { details: X } = assert;
 
 /**
  * @param {import('./vaultFactory.js').VaultFactoryZCF} zcf
- * @param {import('@agoric/governance/src/contractGovernance/typedParamManager').TypedParamManager<{Electorate: "invitation"}>} directorParamManager
+ * @param {import('@agoric/governance/src/contractGovernance/typedParamManager').TypedParamManager<{Electorate: "invitation", LiquidationInstall: "installation", LiquidationTerms: "unknown", MinInitialLoan: "amount"}>} directorParamManager
```
dtribble: This seems like it should also be `VaultDirectorParams` above.

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -101,12 +101,24 @@ const makeVaultInvitation = ({ state }) => {
     });
     const {
       give: { Collateral: collateralAmount },
+      want: { RUN: borrowAmount },
```
dtribble: How about just `requestedAmount`?

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -101,12 +101,24 @@ const makeVaultInvitation = ({ state }) => {
...
     );
+
+    assert(
+      AmountMath.isGTE(
+        borrowAmount,
+        // @ts-expect-error VaultDirectoryParams isn't right
+        state.directorParamManager.getMinInitialLoan(),
+      ),
+      // @ts-expect-error VaultDirectoryParams isn't right
+      X`Minimum loan amount is ${state.directorParamManager.getMinInitialLoan()}, won't lend ${borrowAmount}`,
```
dtribble: `The requested amount (${requestedAmount}) must be at least ${state.directorParamManager.getMinInitialLoan()}`

packages/run-protocol/src/vaultFactory/vaultFactory.js
```diff
@@ -24,15 +24,23 @@ import {
...
+  MIN_INITIAL_LOAN_KEY,
 } from './params.js';
 import { makeVaultDirector } from './vaultDirector.js';
 
 /**
- * @typedef {ZCF<GovernanceTerms<{}> & {
+ * @typedef {{
+ *   LiquidationTerms: ParamRecord<'unknown'>,
+ *   LiquidationInstall: ParamRecord<'installation'>,
+ *   MinInitialLoan: ParamRecord<'amount'>,
```
dtribble: Don't use "loan" for this.

dtribble: An alternative to renaming it is to structure it differently; e.g., "MinCreationFee" and then either always charge at least that, or reject any loan that doesn't have at least that much in fees.

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -106,6 +106,7 @@ test.before(async t => {
       chargingPeriod: 2n,
       recordingPeriod: 6n,
     },
+    minInitialLoan: 5_000_000n,
```
dtribble: Please use an actual Amount here (like with the initial liquidity below). That was clarifying when I redid the test setup for `test-liquidator`

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -621,6 +626,7 @@ test('update liquidator', async t => {
   } = t.context;
   t.context.runInitialLiquidity = AmountMath.make(debtBrand, 500_000_000n);
   t.context.aethInitialLiquidity = AmountMath.make(aethBrand, 100_000_000n);
+  t.context.minInitialLoan = 270n;
```
dtribble: Please use an actual Amount.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1551,6 +1561,8 @@ test('mutable liquidity triggers and interest', async t => {
     chargingPeriod: SECONDS_PER_WEEK,
     recordingPeriod: SECONDS_PER_WEEK,
   };
+  t.context.minInitialLoan = 500n;
```
dtribble: Why do all the tests have this added. Most should just use the defaults.

dtribble (changes requested):

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: The production default is on the order of 5_000_000n. I could set the default in the tests to 200n, and it wouldn't effect tests. I could set it to 500n and it would only require a few tests to be updated. Do you have a preference?

packages/run-protocol/test/vaultFactory/test-liquidator.js
Chris-Hibbert: I used to have an actual Amount. Check the final commit to see.   That didn't work well with econ-behaviors where I needed a default value for some tests in `.../packages/solo/` that didn't set anything.  The default there can't be an amount because the brand isn't available during initialization.

packages/run-protocol/src/vaultFactory/vaultFactory.js
Chris-Hibbert: In the issue (#4967) I asked about how to structure the restriction. I think the limit that makes the most sense to users and is straightforward for us to implement is a limit on how much the user borrows. If we need to discuss alternatives further, the context is in that issue.

packages/run-protocol/src/econ-behaviors.js
Chris-Hibbert: fixed

Chris-Hibbert: You're right that it should be positive. TypeScript doesn't seem to have `nat`.  I respond about Amounts elsewhere.

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: tried that.  I'll leave it to @turadg to fix.

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: good idea.  Done

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: I went with `The request must be for at least ${getMinInitialDebt}. ${requestedAmount} is too small`

packages/run-protocol/test/vaultFactory/test-liquidator.js
Chris-Hibbert: responded above.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: https://github.com/Agoric/agoric-sdk/pull/5243 has the fix. Feel free to click merge on that if you want it in this PR or merge this PR and I'll land the TS fix on master.

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -226,6 +227,7 @@ export const setupReserve = async ({
  * @param { EconomyBootstrapPowers } powers
  * @param { Object } config
  * @param { LoanTiming } [config.loanParams]
+ * @param {bigint} minInitialDebt
```
dtribble, line 230: I'm not sure whether the braces should be closer in all the other lines or whether this needs spaces around braces. Why did the formatting accept this inconsistency?

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -251,6 +253,7 @@ export const startVaultFactory = async (
       recordingPeriod: SECONDS_PER_DAY,
     },
   } = {},
+  minInitialDebt = 5_000_000n,
```
dtribble, line 256: Why is there a hard-coded value here? I thought the default value should come in from governance in the first place.

dtribble, line 256: The centralBrandP is available here so this could just be a real amount?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -106,6 +106,7 @@ test.before(async t => {
       chargingPeriod: 2n,
       recordingPeriod: 6n,
     },
+    minInitialDebt: 5_000_000n,
```
dtribble: Note that we have `runBrand` here as well.

dtribble: This should set the minimum very low for all the tests that don't care, and then the one that cares should set it to an appropriate value.  `price drop` has nothign to do with this limit and so should not have to mention it.

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -478,6 +480,7 @@ test('price drop', async t => {
     chargingPeriod: 2n,
     recordingPeriod: 10n,
   };
+  t.context.minInitialDebt = 270n;
```
dtribble: This test should not have to change this value. Default it to 0n in tests except for the test of the limit.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1551,6 +1561,8 @@ test('mutable liquidity triggers and interest', async t => {
     chargingPeriod: SECONDS_PER_WEEK,
     recordingPeriod: SECONDS_PER_WEEK,
   };
+  t.context.minInitialDebt = 500n;
```
dtribble: As above, this should be 0n in the default test config.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2006,6 +2022,43 @@ test('excessive loan', async t => {
   });
 });
 
+test('loan too small', async t => {
+  const {
+    zoe,
+    aethKit: { mint: aethMint, brand: aethBrand },
+    runKit: { brand: runBrand },
+  } = t.context;
+
```
dtribble, line 2021: this should set the min value appropriate for this test.

dtribble (approved):
The changes in the tests should be fixed, but I don't require a re-review of that. The change between bigint and Amount in the param definition seems like it could still be an amount. But the code looks like it will work fine.

packages/run-protocol/src/econ-behaviors.js
turadg: We don't have formatting of JSDoc yet. Far back burner https://github.com/Agoric/agoric-sdk/pull/4903

fwiw, the tighter is what the (only available jsdoc) autoformatter does. helps distinguish from object braces. e.g. `@param {{ aKey: number, bKey: string }} options`

packages/run-protocol/src/econ-behaviors.js
Chris-Hibbert: I was forced into this because some test in `packages/vats` depends on econ-behavior, and doesn't specify any parameters, so the defaults have to work.

Chris-Hibbert: I tried to do that, but centralBrandP is coming in as a separate parameter, so AFAICT it can't be used to construct a default value for this parameter.

packages/run-protocol/test/vaultFactory/test-liquidator.js
Chris-Hibbert: done

packages/run-protocol/test/vaultFactory/test-liquidator.js
Chris-Hibbert: done

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: done

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: done

packages/run-protocol/src/econ-behaviors.js
turadg: it's `NatValue`. that's really `bigint` but it serves as documentation,
https://github.com/Agoric/agoric-sdk/blob/454afd34274017175db5ac0513d6cee5b7375a3a/packages/ERTP/src/types.js#L435

and eventually we could use that information to automate guards.

Chris-Hibbert: thanks.  done.

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: done

PR #5239, opened 2022-04-28
feat(governance): add the ability to govern json parameters

### LINKED ISSUE #5184: Support JSON as a governance param type

Liquidation needs a dynamic set of arguments provided via governance, depending on the specific liquidation installation used

## Description of the Design

Add `ParamTypes.JSON` to the set of supported param types

## Security Considerations

N/A

## Test Plan

Extend current param tests for the new type

Chris-Hibbert:
closes: #5184

## Description

Adds the ability to govern json parametes.

### Security Considerations

A new type for governance. 

Nothing special required. we still need to document governance.

### Testing Considerations

Added positive, negative, and regression tests for json blobs.

### DIFF
--- a/packages/governance/src/contractGovernance/paramManager.js
+++ b/packages/governance/src/contractGovernance/paramManager.js
@@ -45,6 +45,7 @@ const assertElectorateMatches = (paramManager, governedParams) => {
  * @property {(name: string, value: bigint) => ParamManagerBuilder} addNat
  * @property {(name: string, value: Ratio) => ParamManagerBuilder} addRatio
  * @property {(name: string, value: string) => ParamManagerBuilder} addString
+ * @property {(name: string, value: string) => ParamManagerBuilder} addJson
  * @property {(name: string, value: any) => ParamManagerBuilder} addUnknown
  * @property {() => AnyParamManager} build
  */
@@ -160,6 +161,21 @@ const makeParamManagerBuilder = zoe => {
   const addString = (name, value, builder) => {
     const assertString = v => assert.typeof(v, 'string');
     buildCopyParam(name, value, assertString, ParamTypes.STRING);
+    return builder;
+  };
+
+  /** @type {(name: string, value: string, builder: ParamManagerBuilder) => ParamManagerBuilder} */
+  const addJson = (name, value, builder) => {
+    const assertJson = v => {
+      try {
+        assert.typeof(v, 'string');
+        JSON.parse(v);
+      } catch (e) {
+        throw Error(`Expected a json string. "${v}" is not valid json.`);
+      }
+      return true;
+    };
+    buildCopyParam(name, value, assertJson, ParamTypes.JSON);
     return builder;
   };
 
@@ -328,6 +344,7 @@ const makeParamManagerBuilder = zoe => {
       getNat: name => getTypedParam(ParamTypes.NAT, name),
       getRatio: name => getTypedParam(ParamTypes.RATIO, name),
       getString: name => getTypedParam(ParamTypes.STRING, name),
+      getJson: name => getTypedParam(ParamTypes.JSON, name),
       getUnknown: name => getTypedParam(ParamTypes.UNKNOWN, name),
       getVisibleValue,
       getInternalParamValue,
@@ -350,6 +367,7 @@ const makeParamManagerBuilder = zoe => {
     addNat: (n, v) => addNat(n, v, builder),
     addRatio: (n, v) => addRatio(n, v, builder),
     addString: (n, v) => addString(n, v, builder),
+    addJson: (n, v) => addJson(n, v, builder),
     build: () => makeParamManager(),
   };
   return builder;
--- a/packages/governance/src/contractGovernance/paramManager.js
+++ b/packages/governance/src/contractGovernance/paramManager.js
@@ -112,7 +112,7 @@ const makeParamManagerBuilder = zoe => {
   const addAmount = (name, value, builder) => {
     const assertAmount = a => {
       assert(a.brand, `Expected an Amount for ${name}, got "${a}"`);
-      return AmountMath.coerce(value.brand, a);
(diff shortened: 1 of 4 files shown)


packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -163,6 +164,21 @@ const makeParamManagerBuilder = zoe => {
...
+  /** @type {(name: string, value: string, builder: ParamManagerBuilder) => ParamManagerBuilder} */
+  const addJson = (name, value, builder) => {
+    const assertJson = v => {
+      try {
+        assert.typeof(v, 'string');
+        JSON.parse(v);
+      } catch (e) {
+        throw Error(`Expected a json string. "${v}" is not valid json.`);
+      }
+      return true;
```
turadg: this is consistent with the others, but why do any return `true`? IMO `assert` should always return void and sometimes throw.

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -320,6 +320,41 @@ test('Strings', async t => {
...
+  const paramManager = makeParamManagerBuilder()
+    .addNat('Acres', 50n)
+    .addJson('Blob', '{ "author": "Crockford" }')
+    .build();
+  t.is(paramManager.getJson('Blob'), '{ "author": "Crockford" }');
+
+  await paramManager.updateParams({
+    Blob: '{ "authors": "Crockford,Morningstar" }',
+  });
+  t.is(paramManager.getBlob(), '{ "authors": "Crockford,Morningstar" }');
```
turadg, line 333: nice :)

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -320,6 +320,41 @@ test('Strings', async t => {
...
+        Blob: 300000000,
+      }),
+    {
+      message: 'Expected a json string. "300000000" is not valid json.',
+    },
+  );
+  await t.throwsAsync(
+    () =>
+      paramManager.updateParams({
+        Blob: 'author: bob',
```
turadg, line 353: this seems quite similar to the test above.

some tests for common manual editing:
- valid JSON with trailing whitespace passes
- valid JSON with trailing semicolon fails

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -320,6 +320,41 @@ test('Strings', async t => {
   });
 });
 
+test('JSON blobs', async t => {
+  const paramManager = makeParamManagerBuilder()
+    .addNat('Acres', 50n)
+    .addJson('Blob', '{ "author": "Crockford" }')
+    .build();
+  t.is(paramManager.getJson('Blob'), '{ "author": "Crockford" }');
```
turadg, line 328: shouldn't the result be the JS value instead of the string?
```suggestion
  t.is(paramManager.getJson('Blob'), { "author": "Crockford" });
```

Requesting that change or some documentation on why the `string` type is not sufficient.

turadg (changes requested):

packages/governance/src/contractGovernance/paramManager.js

packages/governance/test/unitTests/test-buildParamManager.js
Chris-Hibbert: It's a good question. My strong suspicion is that for governance purposes, we're talking about json values encoded as strings. I think people want to see the stringified version when they're voting, and I think governed values that are specified as json would be desired by the contract as json strings.

But I could be wrong about that. Particularly since the contract that I thought was using json actually uses the value of its parameter as a record. (I'm adding that in a separate PR.)

What do you think? Would it make sense to specify it as a string and extract the value as an object?

packages/governance/test/unitTests/test-buildParamManager.js
Chris-Hibbert: done

erights (commented):
Why JSON rather than one of
   * any passable value, 
   * any passable key,
   * any passable pure data

?

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -331,6 +331,13 @@ test('JSON blobs', async t => {
     Blob: '{ "authors": "Crockford,Morningstar" }',
   });
   t.is(paramManager.getBlob(), '{ "authors": "Crockford,Morningstar" }');
+
+  await paramManager.updateParams({
+    // trailing whitespace is preserved
+    Blob: '{ "authors": "Crockford,Morningstar" }    ',
+  });
+  t.is(paramManager.getBlob(), '{ "authors": "Crockford,Morningstar" }    ');
```
turadg, line 339: nit: better to make this part of the test output.
```suggestion
    Blob: '{ "authors": "Crockford,Morningstar" }    ',
  });
  t.is(paramManager.getBlob(), '{ "authors": "Crockford,Morningstar" }    ', 'extraneous whitespace must be preserved);
```

packages/governance/test/unitTests/test-buildParamManager.js
turadg: >  I think people want to see the stringified version when they're voting

Not sure about this. I think they want to see the data. If it were a tree that'd probably be easier to read. 

> and I think governed values that are specified as json would be desired by the contract as json strings.
> But I could be wrong about that. Particularly since the contract that I thought was using json actually uses the value of its parameter as a record. (I'm adding that in a separate PR.)

Yeah, I think if it's JSON the consumer would want the data itself.

> Would it make sense to specify it as a string and extract the value as an object?

👍 to extracting the value as an object. Not sure about the _string_ encoding, in light of @erights question about supporting any passable.

Chris-Hibbert:
@erights points out that JSON can't represent bigints (and a smattering of other things), so it's not actually useful for these purposes.

Issue #4142, opened 2021-12-02
A Contract for posing Governance Questions


Chris-Hibbert:
## What is the Problem Being Solved?

The facet that allows one to pose new governance questions is returned when an Electorate is started. For contract governance, its returned when the governed contract is started. Currently, that happens in `bootstrap.js` for the VaultMaker and the AMM, but `bootstrap.js` doesn't make any use of it.

## Description of the Design

Allow stakeholders to put up some stake in order to get the ability to pose a question. `bootstrap.js` could then hand its poser facets to this contract, and the question creation contract could have a publicly accessible facet for creating new questions.

## Security Considerations

Resolves the issue of how the fundamental economic contracts can make changes to their parameters.  This contract would be part of governance, so it would need to be included in that security review.

## Test Plan

nothing special

Chris-Hibbert:
At the Zoe meeting on 2022/4/27, we talked about how this should be managed for the contracts the Economic committee controls. A proposal was raised to create a contract that would hold the poser facet for the core contracts of our economy. It would allow proposal that match one of a few templates to be sent to a vote. We already have a start at a list of governed params that they might consider changing (*link?*). The ability to invoke the imminent ZCF API that allows some invitations to be blocked should also be include.

dckc:
We deployed such a contract (from a pending PR, https://github.com/Agoric/agoric-sdk/pull/5257 ) to devnet yesterday.
https://devnet.explorer.agoric.net/proposals/5
End Voting Time 2 May 2022, 11:28:20pm UTC

rowgraus:
@Tartuffo per conversation with @Chris-Hibbert we believe the existing contract referenced above is sufficient for Mainnet 1.  The result is that one member of the committee will have the ability to pose arbitrary governance questions (which may be malicious or dumb) to the econ committee, but the committee must approve any question for it to be implemented.

This issue, which would constrain the questions allowed, is a nice to have.  Moving to Mainnet 1.1 tag/release

Chris-Hibbert:
We have two examples now:

 #6110 adds a charter for the PSM which plays this role for the PSM. 
#6089 added a charter for the broader econ committee.

PR #5245, opened 2022-04-28
feature: add a contract that supports voting on economic contracts

Chris-Hibbert:
refs #4142

## Description

add a contract that supports voting on economic contracts

### Security Considerations

...lots to say here...

TBD

### Testing Considerations

TBD

### DIFF
--- a/packages/run-protocol/src/voting.js
+++ b/packages/run-protocol/src/voting.js
@@ -38,7 +38,7 @@ export const start = async (zcf, privateArgs) => {
   };
 
   const voteOnApiInvocation = (contract, method, amounts, deadline) => {
-    return E(reserve).voteOnApiInvocation(method, amounts, counter, deadline);
+    return E(contract).voteOnApiInvocation(method, amounts, counter, deadline);
   };
 
   const makeNullInvitation = () => {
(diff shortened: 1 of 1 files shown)


packages/run-protocol/src/voting.js
```diff
@@ -0,0 +1,64 @@
...
+    deadline,
+  ) => {
+    return E(target).voteOnParamChanges(counter, deadline, {
+      paramPath: path,
+      changes: params,
+    });
+  };
+
+  const voteOnApiInvocation = (contract, method, amounts, deadline) => {
+    return E(reserve).voteOnApiInvocation(method, amounts, counter, deadline);
```
dckc:
```suggestion
    return E(contract).voteOnApiInvocation(method, amounts, counter, deadline);
```

Chris-Hibbert: That was intentional, since reserve is the only one with an ApiInvocation, but we can make it general.

dckc: or take out the `contract` arg

Chris-Hibbert: Aah.  you're right. I was part way through a conversion. I think making it general was the right direction to go.

dckc:
We incorporated this into #5257 ... I'm not sure whether it's worth keeping this open.

michaelfig:
> I'm not sure whether it's worth keeping this open.

I'll express my opinion via the Close button.

PR #5250, opened 2022-04-29
feat(swingset): allow buildRootObject to return a Promise

### LINKED ISSUE #5246: async buildRootObject

#3272 turns out to have a surprising requirement: we need to do `importBundle` (which is async, for deep reasons) from inside `buildRootObject` (which is sync). This comes from ZCF needing to import the new version of the contract, then invoke the contract method that re-establishes behavior for the durable Kinds, all of which needs to happen before the end of the `buildRootObject` delivery (so it can execute the normal deliveries that will immediately follow).

`D(bundlecap).getBundle() -> bundle` is synchronous, for exactly this reason, but I forgot that `importBundle` is async.

## Description of the Design

It looks like making liveslots tolerate an async `buildRootObject` is the best way to deal with this. The tricky part is the error handling, in particular what to do if the user-provided `buildRootObject` function returns a Promise that never resolves. We must prevent userspace from hanging the supervisor.

The tail end of `startVat` currently does:
```js
    const rootObject = buildRootObject(harden(vpow), harden(vatParameters), baggage);
    assert.equal(passStyleOf(rootObject), 'remotable', X`buildRootObject() for vat ${forVatID} returned ${rootObject}, which is not Far`);
    assert(getInterfaceOf(rootObject) !== undefined, X`buildRootObject() for vat ${forVatID} returned ${rootObject} with no interface`);
    // Need to load watched promises *after* buildRootObject() so that handler kindIDs
    // have a chance to be reassociated with their handlers.
    watchedPromiseManager.loadWatchedPromiseTable();

    const rootSlot = makeVatSlot('object', true, BigInt(0));
    valToSlot.set(rootObject, rootSlot);
    slotToVal.set(rootSlot, new WeakRef(rootObject));
    retainExportedVref(rootSlot);
    // we do not use vreffedObjectRegistry for root objects

    vom.insistAllDurableKindsReconnected();
  }
```

and the code which calls it is basically:
```js
async function dispatch(delivery) {
  const p = Promise.resolve(delivery).then(startVat);
  return waitUntilQuiescent().then(() => p);
}
```

I'm thinking the trick will be to connect the `buildRootObject` promise to a function that finishes the `startVat` process, then sets a flag. We return a separate promise from `startVat`. I need a way for that second promise to reject if the flag isn't set at the time `waitUntilQuiescent` fires. That probably means introducing a callback into the `() => p` portion, to query the flag.

## Security Considerations

Main requirement is to not allow an unresolved userspace promise to hang the worker (and kernel).

## Test Plan

unit tests

warner:
New (or upgraded) vats are defined by a user-provided
`buildRootObject` function. This function has two responsibilities:

* provide a root object
* use `defineDurableKind` to bind behavior to any pre-existing durable
  kinds

Previously, `buildRootObject` was given exactly one "turn" to fulfill
these responsibilities: it was required to return the root object
synchronously. If it returned a Promise, that was rejected as an
error.

This requirement is at odds with the needs for updated ZCF vats to
evaluate their contract code and interact with it (giving the contract
a chance to bind their own Kind behavior). The `importBundle` function
we use to load contracts is async. However, it is still "prompt": it
needs no IO, nor any messages leaving the vat. So it will still
complete within the *crank*, just not synchronously within the turn.

So this commit changes liveslots to allow `buildRootObject` to return
a Promise for the root object. The requirement, however, is that this
Promise must resolve by the end of the initial crank (before
`waitUntilQuiescent` / `setImmediate` determines that the promise
queue is empty). If the Promise is not fulfilled by then, the vat
creation/update is considered broken.

closes #5246

### DIFF
--- a/packages/SwingSet/src/kernel/vat-loader/manager-subprocess-xsnap.js
+++ b/packages/SwingSet/src/kernel/vat-loader/manager-subprocess-xsnap.js
@@ -131,6 +131,9 @@ export function makeXsSubprocessFactory({
     async function issueTagged(item) {
       parentLog(item[0], '...', item.length - 1);
       const result = await worker.issueStringCommand(JSON.stringify(item));
+      // note: if 'result.reply' is empty, that probably means the
+      // worker reached end-of-crank (idle) without seeing the
+      // `dispatch()` promise resolve
       const reply = JSON.parse(result.reply);
       assert(Array.isArray(reply));
       const [tag, ...rest] = reply;
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -1275,7 +1275,7 @@ function build(
     );
 
     // here we finally invoke the vat code, and get back the root object
-    const rootObject = buildRootObject(
+    const rootObject = await buildRootObject(
       harden(vpow),
       harden(vatParameters),
       baggage,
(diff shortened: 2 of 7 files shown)


warner:
oops, this is a recreation of #5249 , which I accidentally closed by deleting the branch (`git push :5246-async-buildrootobject`) instead of updating it (`git push +5246-async-buildrootobject`)

@mhofman could you re-review to satisfy the merge criteria? I added a comment to address your note.

mhofman (approved):
Carrying over previous approval

Issue #5024, opened 2022-04-06
Place "async" vat operations onto outbound queue


mhofman:
## What is the Problem Being Solved?

In the latest [two run queue](https://github.com/Agoric/agoric-sdk/issues/3465) design, syscalls such as `send`, `resolve` and `subscribe` will be placed on a vat specific outbound queue as-is. Only when picked form the outbound queue, the syscalls will be fully processed, with corresponding reactions placed on other vat's inbound queues or kernel promise queues.

## Description of the Design

As a first step #4638 added a single acceptance queue. Once `send` can be enqueue directly without going through the delivery queue (#4542), the `send`, `resolve` and `subscribe` syscall should be updated to not perform synchronous processing during the syscall, and instead delay that until plucked from the acceptance queue.

`exit` should be able to stay synchronous, so that a vat is immediately marked as terminated. In the future, the vat outbound queue will be drained even if the vat is marked as terminated.

`dropImports` and `retireImports` will stay no-ops (after translation), and the kernel ref counts should be sufficient to keep the kernel as a "holder" if needed, including for references still in the outbound queue.

`retireExports` and `abandonExports` can both keep updating kernel structures synchronously. In the case of `retireExports`, if the vat is well behaved, it should not have mentioned the retired object in a previous syscall. `abandonExports` simply does not make any `send`, `resolve` or `subscribe` calls.

## Security Considerations

Ref counting must stay consistent. A message in the current acceptance and future outbound queues must keep incrementing ref counts. Moving from one queue to another should not change ref counting.

## Test Plan
Add and update unit tests. Run loadgen as simple integration test.

mhofman:
I just thought of a new comms / kernel test to add:
- vatA sends foo message to object on remote vat with result r1
- vatA sends bar message to r1 promise with result r2
- vatA sends baz message to r2 promise
- remote resolves r1 to object imported from vatA
  - comms reflects bar message back to kernel and vatA, reusing r2, this redirecting it
- kernel must move message baz from comms queue back to promise queue

There are also variations, like sending the initial message to a promise that vatA resolves after step 1 2 or 3, resulting in any some of those messages having to move from promise to decider queue, or straight to decider queue.

Some of these may not currently pipeline correctly because once in the promise queue, the message doesn't move back to the run-queue when the decider is updated.

Issue #2870, opened 2021-04-13
comms/kernel GC cycle collection tricks: back-pointers


warner:
## What is the Problem Being Solved?

In both the kernel and the comms vat, we've got object- and promise- tables whose entries need to be deleted when they're no longer referenced. This will mirror the JS engine's GC approach, except all the objects involved are logical ones, referenced by data (krefs), so we can't benefit from the engine's support for collection actual `Object` objects.

Simple refcounting is pretty straightforward and effective unless it encounters cycles. Currently (until we implement #2069 auxdata), cycles will only happen when `Promise` p1 is resolved to data which references p2, and p2 is resolved to data referencing p1 (or a larger cycle). The comms vat does not have a run-queue or messages queued up on a target promise (all messages are pipelined immediately to the target's decider), so the only inbound references will be clist entries: other vats (or the kernel) who might still be able to reference the promise. Once resolved, we retire the promise-id as quickly as possible: the kernel can retire it immediately, but for other vats we must wait for our outbound resolution message to be acknowleged, which will take a round-trip time, so the clist's reference won't be deleted right away.

While alive, the cycle of mutually-referencing promises will have at least one inbound reference from a clist. When the last clist entry goes away, we'd like to detect and delete the cycle.

The standard way of noticing cycles is to perform a full mark-and-sweep operation, starting from the roots (all clist entries). I'm nervous about the performance consequences of walking through the entire comms clist and promise tables (as a side-effect of receiving a retirement ack, no less, so: jank). In addition, so far we've managed to avoid adding a secondary-storage API to enumerate DB keys, and both clists and the promise table are stored in secondary storage. So we're also disinclined to support such enumeration, just to keep the API simple.

Given such a constraint, @FUDCo and I were noodling about tricks we might still be able to use. I've been pondering back-pointers as a tool, and he suggested tracking inbound references from clist entries (as a simple counter) separately from the cross-promise references (recorded as a list of other promise-ids). The algorithm to apply when deleting a c-list entry would then look like:

* decrement the c-list refcount of the target promise
* if the resulting c-list refcount is nonzero, end
* else: examine the cross-promise reference list. If zero, drop the target promise, recurse
* else: perform a search:
  * crawl the transitive closure of inbound cross-promise pointers, starting at the target
  * for each node found, if the c-list refcount is non-zero, end: the cycle is still reachable
  * if the crawl terminates without finding an inbound c-list reference, the cycle is non-reachable
[...]

PR #5252, opened 2022-04-29
feat(swingset): queue to promise

### LINKED ISSUE #4542: syscall.send to resolved promise should run-queue to object, not to promise

mhofman:
closes: #4542
refs: #4318 #5024

Best reviewed commit-by-commit.

## Description

This PR updates the processing of message sends onto promises so that:
- The target is checked when plucking the send from the acceptance queue and either queued onto the run-queue or to the corresponding promise queue if the promise is neither resolved nor pipelined
- Sends queued onto promises are moved to the acceptance-queue when their decider changes to a pipelining vat, or when they are resolved (without a target update)

Ultimately the sends should not transit back through the acceptance queue, and should have their target updated right away to release the resolved promise. This will be done in the follow-up PR addressing #5024 (along with immediate queueing of notifies on the run-queue).

### Security Considerations

While this change may update the order of message sends between objects and promises, this is valid per our ordering guarantees. In consideration, this is a small ordering update compared to the relaxed ordering we want to introduce.

IOU an update to docs once the queue processing churn has settled

### Testing Considerations

Added a test for a preexisting limitation of message sends onto pipelined promises. Update existing tests to account for new message order.

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -390,8 +390,10 @@ export default function buildKernel(
     insistMessage(msg);
     kernelKeeper.incStat('dispatches');
     kernelKeeper.incStat('dispatchDeliver');
+
     // eslint-disable-next-line no-use-before-define
-    if (!vatWarehouse.lookup(vatID)) {
+    const vatInfo = vatWarehouse.lookup(vatID);
+    if (!vatInfo) {
       // splat
       if (msg.result) {
         resolveToError(msg.result, VAT_TERMINATION_ERROR);
@@ -403,6 +405,11 @@ export default function buildKernel(
     const kd = harden(['message', target, msg]);
     // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
+
+    if (vatInfo.enablePipelining && msg.result) {
+      kernelKeeper.requeueKernelPromise(msg.result);
+    }
+
     return deliverAndLogToVat(vatID, kd, vd);
   }
 
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1071,13 +1071,13 @@ export default function buildKernel(
   }
 
   let processQueueRunning;
-  async function tryProcessDeliveryMessage(message) {
+  async function tryProcessMessage(processor, message) {
     if (processQueueRunning) {
       console.error(`We're currently already running at`, processQueueRunning);
       assert.fail(X`Kernel reentrancy is forbidden`);
     }
     processQueueRunning = Error('here');
-    return processDeliveryMessage(message).finally(() => {
+    return processor(message).finally(() => {
       processQueueRunning = undefined;
     });
   }
@@ -1085,32 +1085,24 @@ export default function buildKernel(
   async function processAcceptanceMessage(message) {
     kdebug(`processAcceptanceQ ${JSON.stringify(message)}`);
     kdebug(legibilizeMessage(message));
-    if (processQueueRunning) {
-      console.error(`We're currently already running at`, processQueueRunning);
-      assert.fail(X`Kernel reentrancy is forbidden`);
-    }
     kernelSlog.write({ type: 'crank-start', message });
     /** @type { PolicyInput } */
     const policyInput = ['none'];
-    try {
-      processQueueRunning = Error('here');
 
-      kernelKeeper.addToRunQueue(message);
+    kernelKeeper.addToRunQueue(message);
 
-      kernelKeeper.processRefcounts();
-      kernelKeeper.saveStats();
-      const crankNum = kernelKeeper.getCrankNumber();
-      kernelKeeper.incrementCrankNumber();
-      const { crankhash, activityhash } = kernelKeeper.commitCrank();
-      kernelSlog.write({
-        type: 'crank-finish',
-        crankNum,
-        crankhash,
-        activityhash,
-      });
-    } finally {
-      processQueueRunning = undefined;
-    }
+    kernelKeeper.processRefcounts();
+    kernelKeeper.saveStats();
+    const crankNum = kernelKeeper.getCrankNumber();
+    kernelKeeper.incrementCrankNumber();
+    const { crankhash, activityhash } = kernelKeeper.commitCrank();
+    kernelSlog.write({
+      type: 'crank-finish',
+      crankNum,
+      crankhash,
+      activityhash,
+    });
+
     return harden(policyInput);
   }
 
@@ -1426,19 +1418,16 @@ export default function buildKernel(
     return undefined;
   }
 
-  function startProcessingNextMessageIfAny() {
-    /** @type {Promise<PolicyInput> | undefined} */
-    let resultPromise;
+  function getNextMessageAndProcessor() {
     let message = getNextAcceptanceMessage();
-    if (message) {
-      resultPromise = processAcceptanceMessage(message);
-    } else {
+    /** @type {(message:any) => Promise<PolicyInput>} */
+    let processor = processAcceptanceMessage;
+    if (!message) {
       message = getNextDeliveryMessage();
-      if (message) {
-        resultPromise = tryProcessDeliveryMessage(message);
-      }
+      processor = processDeliveryMessage;
     }
-    return { resultPromise };
+
+    return { message, processor };
   }
 
   async function step() {
@@ -1448,10 +1437,10 @@ export default function buildKernel(
     if (!started) {
       throw new Error('must do kernel.start() before step()');
     }
-    const { resultPromise } = startProcessingNextMessageIfAny();
+    const { processor, message } = getNextMessageAndProcessor();
     // process a single message
-    if (resultPromise) {
-      await resultPromise;
+    if (message) {
+      await tryProcessMessage(processor, message);
       if (kernelPanic) {
         throw kernelPanic;
       }
@@ -1480,14 +1469,14 @@ export default function buildKernel(
     }
     let count = 0;
     for (;;) {
-      const { resultPromise } = startProcessingNextMessageIfAny();
-      if (!resultPromise) {
+      const { processor, message } = getNextMessageAndProcessor();
+      if (!message) {
         break;
       }
       count += 1;
       /** @type { PolicyInput } */
       // eslint-disable-next-line no-await-in-loop
-      const policyInput = await resultPromise;
+      const policyInput = await tryProcessMessage(processor, message);
       if (kernelPanic) {
         throw kernelPanic;
       }
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -737,7 +737,7 @@ export default function buildKernel(
    * This does not decrement any refcounts. The caller should do that.
    *
    * @param { RunQueueEventSend } message
-   * @returns { { vatID: VatID, targetObject: string } | null }
+   * @returns { { vatID: VatID | null, target: string } | null }
    */
   function routeSendEvent(message) {
     const { target, msg } = message;
@@ -759,12 +759,12 @@ export default function buildKernel(
       if (!vatID) {
         return splat(VAT_TERMINATION_ERROR);
       }
-      return { vatID, targetObject };
+      return { vatID, target: targetObject };
     }
 
-    function enqueue() {
-      kernelKeeper.addMessageToPromiseQueue(target, msg);
-      return null; // message is queued, not sent to a vat right now
+    function requeue() {
+      // message will be requeued, not sent to a vat right now
+      return { vatID: null, target };
     }
 
     if (type === 'object') {
@@ -788,7 +788,7 @@ export default function buildKernel(
       }
       case 'unresolved': {
         if (!kp.decider) {
-          return enqueue();
+          return requeue();
         } else {
           insistVatID(kp.decider);
           // eslint-disable-next-line no-use-before-define
@@ -798,9 +798,9 @@ export default function buildKernel(
             return splat(VAT_TERMINATION_ERROR);
           }
           if (deciderVat.enablePipelining) {
-            return { vatID: kp.decider, targetObject: target };
+            return { vatID: kp.decider, target };
           }
-          return enqueue();
+          return requeue();
         }
       }
       default:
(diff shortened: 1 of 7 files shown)


packages/SwingSet/src/kernel/kernel.js
```diff
@@ -403,6 +405,11 @@ export default function buildKernel(
     const kd = harden(['message', target, msg]);
     // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
+
+    if (vatInfo.enablePipelining && msg.result) {
```
warner, line 409: This is probably the only good place to perform the requeuing, but it'd be nice if there were a way to do it in the translator, at the same time we do the `setDecider`. But `vatInfo.enablePipelining` isn't very available there, which suggests we have a funny split between where we keep track of per-vat runtime-ish things like `enablePipelining`, and per-message / per-clist-entry state. It would certainly be wrong to make `kernelKeeper.setDecider()` do the re-queueing, that's too much for that method, but maybe we should plan a future refactoring that creates a "changePromiseDeciderAndRequeue" method somewhere.

No changes recommended now.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -755,9 +762,9 @@ export default function buildKernel(
       return { vatID, targetObject };
     }
 
-    function enqueue() {
-      kernelKeeper.addMessageToPromiseQueue(target, msg);
-      return null; // message is queued, not sent to a vat right now
+    function requeue() {
+      // message will be requeued, not sent to a vat right now
+      return { vatID: null, targetObject: target };
```
warner: If the `targetObject` return value can actually be a Promise (kpid), not just an Object (koid), we should rename it.

We might be able to just use `target` for the output, since in many cases it is unchanged. It depends upon the caller, but if they hold the result in `route` (instead of immediately unpacking it into `{ vatID, target } = routeSendEvent()`), then the difference between the input `target` and the output `route.target` is pretty clear.

If not, maybe `routeTarget` or `routedTarget`.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1064,46 +1081,60 @@ export default function buildKernel(
   }
 
   let processQueueRunning;
-  async function tryProcessDeliveryMessage(message) {
+  async function tryProcessMessage(processMessage, message) {
```
warner: Since these arguments come from a function named `getNextMessageAndProcessor`, I think they should be named `processor` and `message`, instead of `processMessage`. (Also, `processMessage` bumps up against our incomplete taxonomy of message types, where I've used all of ("send", "deliver", "delivery", "message") to mean the kind of delivery that has a method name and arguments instead of being a promise resolution, and `processor` doesn't)

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1064,46 +1081,60 @@ export default function buildKernel(
...
-    return processDeliveryMessage(message).finally(() => {
+    return processMessage(message).finally(() => {
       processQueueRunning = undefined;
     });
   }
 
   async function processAcceptanceMessage(message) {
     kdebug(`processAcceptanceQ ${JSON.stringify(message)}`);
     kdebug(legibilizeMessage(message));
-    if (processQueueRunning) {
```
warner, line 1081: I'm guessing you determined that this extra check is unnecessary, and/or breaks with the new arrangement?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1419,19 +1450,16 @@ export default function buildKernel(
...
-  function startProcessingNextMessageIfAny() {
-    /** @type {Promise<PolicyInput> | undefined} */
-    let resultPromise;
+  function getNextMessageAndProcessor() {
     let message = getNextAcceptanceMessage();
-    if (message) {
-      resultPromise = processAcceptanceMessage(message);
-    } else {
+    /** @type {(message:any) => Promise<PolicyInput>} */
+    let processMessage = processAcceptanceMessage;
```
warner: yeah I think this could read better as `let processor =`

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -821,27 +832,6 @@ export default function makeKernelKeeper(
     insistKernelType('promise', kernelSlot);
     insistMessage(msg);
 
-    // Each message on a promise's queue maintains a refcount to the promise
```
warner: Ok, so messages queued to an object retain a refcount to their target object, but messages queued *on* a promise do not. Seems reasonable, given the container-ish relationship between promise, queue, and message.

packages/governance/test/swingsetTests/contractGovernor/test-governor.js
```diff
@@ -155,10 +155,10 @@ const expectedBrokenUpdateLog = [
   '&& running a task scheduled for 2. &&',
   'vote outcome: {"changes":{"Electorate":{"brand":"[Alleged: Zoe Invitation brand]","value":[{"description":"questionPoser","handle":"[Alleged: InvitationHandle]","installation":"[Alleged: Installation]","instance":"[Alleged: InstanceHandle]"}]},"MalleableNumber":"[42n]"}}',
   'Validation complete: true',
-  'vote rejected outcome: Error: (an object) was not a live payment for brand (an object). It could be a used-up payment, a payment for another brand, or it might not be a payment at all.',
-  'update failed: Error: (an object) was not a live payment for brand (an object). It could be a used-up payment, a payment for another brand, or it might not be a payment at all.',
-  'Electorate,MalleableNumber changed in a vote.',
+  'vote rejected outcome: Error: A Zoe invitation is required, not (an object)',
```
warner, line 166: yeah why did that test change that way?

warner (approved):
Looks pretty good, some cosmetic changes requested, but feel free to land after those.

packages/SwingSet/src/kernel/state/kernelKeeper.js
mhofman: Ugh? No they do definitely keep a refcount to the target promise. What makes you say that?

packages/governance/test/swingsetTests/contractGovernor/test-governor.js
mhofman: I am not landing this PR until I figure it out. This is just weird.

packages/SwingSet/src/kernel/kernel.js
mhofman: No I refactored the re-entrancy check to share it with `processDeliveryMessage` in https://github.com/Agoric/agoric-sdk/pull/5252/commits/f57b3400a771062b2d83798f232b770c2c36fb5e

packages/governance/test/swingsetTests/contractGovernor/test-governor.js
Chris-Hibbert: in  paramManager.js, we see
```
    const prepareToSetInvitation = async invite => {
      const [preparedAmount] = await Promise.all([
        E(E(zoe).getInvitationIssuer()).getAmountOf(invite),
        assertInvitation(invite),
      ]);

      return [invite, preparedAmount];
    };
```
`E(E(zoe).getInvitationIssuer()).getAmountOf(invite)` and `assertInvitation(invite)` do slightly different checks. `getAmountOf()` calls `assertLivePayment`, which rejects with "was not a live payment for ...".

`assertInvitation` calls `E(zoe).getInvitationDetails`, which also calls `getAmountOf(invite)`, but catches errors and returns "A Zoe invitation is required, ...".

So changing the order or precedence of `getAmountOf` and `assertInvitation` in the two calls in the `Promise.all` could have that effect.

mhofman: Ah so this is a case that `Promise.all` shorts circuits on errors, and we effectively ended up with a race. That's one reason I hate the `Promise.all` error short-circuiting and wrote a [`PromiseAllOrErrors`](https://github.com/Agoric/testnet-load-generator/blob/b9ffa47ddfe8a509b250532f449c180b020a1c22/runner/lib/helpers/async.js#L34-L54) for that (main reason is so that I don't unexpectedly swallow errors).

mhofman: I have added a comment in the test as to the reason of the change, and that it may change again in the future.

In general we'll need to find a way to make these "log" based tests more flexible, as there will definitely be more order changes in the future.

packages/SwingSet/src/kernel/state/kernelKeeper.js
warner: you deleted the portion of `addMessageToPromiseQueue` that increments the refcounts on everything in the message

(this comment is supposed to be associated with the deleted lines 825-845 of `kernelKeeper.js`.. I'm not sure that GitHub renders comments correctly when attached to deleted lines, vs added lines)

yeah now I'm not sure why I said "seems reasonable".. are these refcounts being incremented somewhere else instead now?

mhofman: Ha now I understand where the comment comes from.

So with the new structure, a ref count is added when messages are first added to the acceptance queue right after the syscall is received, and decremented when removed from the run-queue just before the delivery. When moved between queues the refcount doesn't change, except when retargeting from a promise to an object, which currently can only happen when [moving from the acceptance queue to the run queue](https://github.com/Agoric/agoric-sdk/pull/5252/files#diff-93cec0b19f5deab7c2117e72ee0a19eec0957fc7f6ea0445a8c8f6d3e5a7684fR1107-R1108). In the next PR retargeting will also possibly happen when moving from the promise queue to the run-queue (currently it goes back to the acceptance queue) when resolving / setting the decider.

I probably should find a place in the code to write that down.

mhofman: @warner PTAL with https://github.com/Agoric/agoric-sdk/pull/5252/commits/b5d29084ab7c3c7ecee762390f846e9aec6fc56e

warner: Ah, now I see. So previously, while the message was on the run-queue (or the acceptance-queue), it held a refcount on its target+arguments+result (and those refcounts were labeled `enq|msg|*`). In `routeSendEvent`, if we determined the message was aimed at an unresolved non-pipelined-vat-decided promise, we used `kernelKeeper.addMessageToPromiseQueue()` to enqueue it right away, which added refcounts (labelled `pq|${kpid}|*`). `deliverRunQueueEvent` called `routeSendEvent`, then removed the `enq|msg|*` refcounts. This happens to be "make-before-break", which has a tiny performance improvement (the refcount doesn't drop to zero, so we don't mark it as needing investigation later).

In the branch, `routeSendEvent` does not do the queue-to-promise itself, it just figures out the target. Upon return, `deliverRunQueueEvent` has the three-way dispatch (splat, deliver, enqueue-to-promise), and it does the decref for splat and deliver, but doesn't touch the refcounts for enqueue-to-promise.

That reduces churn further (no DB refcount changes for enqueue-to-promise), which is nice. The one downside is that the debug information associated with the refcounts remains labelled as `enq|msg|*` (i.e. "this refcount comes from the run-queue") rather than naming the particular promise that's holding it. We don't record this information, it just shows up in asserts (and if you enable some debugging messages), but we've considered using it for real (#2870) to help deal with cycles better. If we did that, we'd need to revisit the labelling and make sure it's accurate.

I'm on the fence about that tradeoff: two fewer DB sets during the queue-to-promise step, vs accurate debug tags. I guess the performance argument is stronger. But maybe we should add a note (maybe to `incrementRefCount`? Or in `deliverRunQueueEvent` just before the call to `addMessageToPromiseQueue`?) to remind ourselves that these labels are inaccurate for the sake of performance and reducing DB churn.

warner: Oh, and this will show up again if/when we implement redirect/forwarding: if the message is moved to a different queue, the refcount label would get stale. We might need a functional change too: the message will hold a refcount on its original target, but if that was `kpid1`, and then `kpid1` got forwarded to `kpid2`, we either want to decref `kpid1` and incref `kpid2`, or do some bulk thing (that abandons the idea of tags entirely) and incref `kpid2` by the entire refcount of `kpid1` (before deleting `kpid1` without looking at the refcount).

warner (approved):
Thanks for the updates.. those help a bunch. Please add one more comment about the refcount debug label becoming stale, and then this is good to land.

packages/SwingSet/src/kernel/state/kernelKeeper.js
mhofman: Yeah I'd rather find a way that doesn't require updating tags when moving from queue to queue. I see `enq` as a generic "on some kernel queue". I need to understand the use cases for an actionable queue descriptors, but processing all target, return and slots for every queue switch doesn't seem right.

PR #5255, opened 2022-04-29
refactor(notifier): clearer signature for makeNotifierKit()

turadg:
## Description

Changes the `args` name to `initialStateArr` to be a little more clear about what the parameter is for.

**Before**
Screen Shot 2022-04-29 at 2 23 16 PM

**After**
Screen Shot 2022-05-01 at 7 30 13 AM

Also errors if more than one argument is passed. This isn't the norm for passing in extra arguments but since we're typing the rest params and it is either `[]` or `[T]` I think it makes the runtime align with the documentation.

### Security Considerations

--

Improves documentation.

### Testing Considerations

Added test for the explicit `undefined` which an earlier iteration of this PR failed.


erights (changes requested):
Please test this with a call `makeNotifierKit(undefined)`. The original distinguishes it from `makeNotifierKit()`, but I don't believe this version does. The problem is that, for purposes of optional parameters, an `undefined` arg is treated the same as an absent arg. I don't know of any way for the callee to distinguish these other than looking at the arity of the arguments array.

Of course, if your test works, I'll revise my theory and we can continue with this PR.

packages/notifier/src/notifier.js
```diff
@@ -30,24 +30,25 @@ export const makeNotifier = sharableInternalsP => {
   });
 };
 
+const deferInitialState = Symbol('defer initial NotifierKit state');
```
erights: Although it doesn't work in this case, I generally like this technique of introducing a unique value and using it in an encapsulated manner to distinguish it from any value that users of this abstraction could have passed in. However, it is important that any value used in this way cannot leak outside this module. Otherwise, a caller that happens to handle such a value (whether maliciously or accidentally) can confuse the abstraction by passing it back in. This encapsulation needs to be robustly preserved under maintenance. 
   * The value declaration itself should be commented to make it clear that the value must not leak. When I use this technique I try to always remember to add this comment.
   * The code that actually uses the dangerous value should be written to *obviously* not leak it. There should be little code between where it is introduced and where it is tested and erased. It should be trivial for a reviewer to verify that it does not leak.
   * CI reveals that it actually did leak, which genuinely surprised me! Because of the amount of code between its introduction and its testing/erasure, I have not yet even spotted how it leaked.

erights (changes requested):

erights:
Even though the code change doesn't work (unless your tests say otherwise), I would still welcome a PR that gave the existing code better types. Feel free to reuse this PR for that purpose.

packages/notifier/src/notifier.js
```diff
@@ -138,8 +139,8 @@ export const makeNotifierKit = (...args) => {
     },
   });
 
-  if (args.length >= 1) {
-    updater.updateState(args[0]);
+  if (initialState !== deferInitialState) {
+    updater.updateState(deferInitialState);
```
erights: Found it. 

```suggestion
    updater.updateState(initialState);
```

This should have damaged almost any use of notifiers. I'm alarmed that only two CI tests failed.

turadg: I should have run tests before requesting review. When I fixed this error, all tests passed. So I added a test for what should have failed with the default param change: [121e248](https://github.com/Agoric/agoric-sdk/pull/5255/commits/121e24813703f203fc8ae76d579709cf0e123055)

turadg:
> Even though the code change doesn't work (unless your tests say otherwise)

One failure was more of a typo but the larger one I didn't think it through enough.

> I would still welcome a PR that gave the existing code better types. Feel free to reuse this PR for that purpose.

Done and updated description. Not as _much_ more clear but probably best that can be done right now. If we had `.ts` syntax I think we could decouple the typing info from the JSDoc. I.e. we could have the rest types and the JSDoc could say that there's one optional argument called `initialState`.

packages/notifier/src/notifier.js
```diff
@@ -37,17 +37,17 @@ export const makeNotifier = sharableInternalsP => {
...
- * syntax. Rather we use the arity of the `args` array.
+ * syntax. Rather we use the arity of the `initialStateArr` array.
  *
  * If no initial state is provided to `makeNotifierKit`, then it starts without
  * an initial state. Its initial state will instead be the state of the first
  * update.
  *
  * @template T
- * @param {[] | [T]} args the first state to be returned
+ * @param {[] | [T]} initialStateArr the first state to be returned (typed as rest array to permit 'undefined')
```
erights:
```suggestion
 * @param {[] | [T]} initialStateArr the first state to be returned (typed as rest array to permit `undefined`)
```

packages/notifier/src/notifier.js
```diff
@@ -138,8 +138,9 @@ export const makeNotifierKit = (...args) => {
     },
   });
 
-  if (args.length >= 1) {
-    updater.updateState(args[0]);
+  assert(initialStateArr.length <= 1, 'too many arguments');
+  if (initialStateArr.length === 1) {
```
erights, line 142: Better, thanks!

erights (approved):
LGTM, thanks

PR #5256, opened 2022-04-29
ci: improve test parallelism

mhofman:
## Description

I've seen some of the XS tests take long times (over 20 minutes). This updates the matrix strategy of the "Test all packages" job to include `xs` as one of the strategies alongside both node versions. It also enables the `concurrency` actions option to cancel jobs that were still running on the same PR. Finally it fixes the `deployment-test` job to not reuse node_modules and built output if it was previously cached.

### Security Considerations

None

None

### Testing Considerations

Need to update required tests in branch protections

### DIFF
--- a/.github/actions/build-node/action.yml
+++ b/.github/actions/build-node/action.yml
@@ -5,6 +5,11 @@ inputs:
   node-version:
     description: 'The version of Node.js to use'
     required: true
+  cache-built:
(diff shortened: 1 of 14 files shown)


michaelfig (approved):
Straightforward and mostly mechanical.  LGTM!

Issue #5169, opened 2022-04-20
Econ committee for Devnet


rowgraus:
### Description

Arrange for a committee of a few people to be able to vote on proposals such as changing interest rates.

 - [x] send voter invitations to wallets
 - [x] figure out how to pose questions

Need addresses for folks on Devnet

dckc:
Devnet Econ members, from today's meeting:
Rowland: agoric1qed57ae8k5cqr30u5mmd46jdxfr0juyggxv6ad
Bill: agoric1xgw4cknedau6xhrlyn6c8e40d02mejee8gwnef
DCKC: agoric1yumvyl7f5nkalss7w59gs6n3jtqv5gmarudx55

dckc:
Today we agreed this need not be deployed at the same time as the RUN protocol contracts (#5062); it can follow by some days.

dckc:
related work: #4561

dckc:
drafted code to distribute the invitations with @turadg :
https://gist.github.com/dckc/1bd0a160ccf0873cef36f3088cbc24c4

I'm not sure about posing questions yet...

dckc:
### Committee of 1 seated on agoricdev-10

Good news: working with @Chris-Hibbert and @michaelfig , I got my invitation to propose questions and vote on them.

It's possible to replace this committee with a committee of 3; we'll see whether this is cost-effective just now.

```
command[0] E(home.myAddressNameAdmin).getMyAddress()
history[0] "agoric14s56vwftwx8p5kz23fercqptnwex5uw2yj903l"

command[22] history[18].at(-2)[1]
history[22] [Object Alleged: Zoe Invitation purse]{}
command[23] E(history[22]).getCurrentAmount()
history[23] {
  "brand":[Object Alleged: Zoe Invitation brand]{},
  "value":[
   {
    "description":"identifies the voting contract instance",
    "handle":[Object Alleged: InvitationHandle]{},
    "installation":[Object Alleged: Installation]{},
    "instance":[Object Alleged: InstanceHandle]{}
   },
   {
    "description":"Voter0","handle":[Object Alleged: InvitationHandle]{},
    "installation":[Object Alleged: Installation]{},
    "instance":[Object Alleged: InstanceHandle]{}
   }
  ]
}
```

```
$ make submit-proposal
git describe --tags --always
@agoric/sdk@16-205-g268355a7e
git diff --stat
 packages/run-protocol/Makefile                        | 10 ++++++----
[...]

Issue #3924, opened 2021-10-04
Enable Governance to add Collateral type to Treasury


Chris-Hibbert:
## What is the Problem Being Solved?

Adding a Collateral type to the Treasury should be under the control of governance.

## Description of the Design

I think this requires a special purpose function, and some general purpose controls to manage and make visible the control over when the function gets invoked. 

The special purpose function includes creating a vaultManager, endowing/finding an appropriate AMM, and connecting the two up as well as setting up the parameters for the new pool & vaultMgr.

## Security Considerations

This involves governance, so the standard legibility targets apply. 

## Test Plan

Probably requires SwingSet tests including multiple voters, and a demonstration of someone validating the chain of responsibility.

as part of RUN Protocol preview:
 - [x] #5169
 - [x] #3595
 - [x] #5084
 - [x] #4714

Chris-Hibbert:
Because
*  this impacts the interactions between the Treasury and the AMM
* it may require some updates on the AMM side to support automated changes
* setting up parameters for the AMM is currently tangled in the Treasury and will eventually be separate.

I think the following order will reduce cross-dependencies most quickly:
* add parameter governance to the constantProduct AMM
* start up constant product AMM stand-alone
* remove AMM creation from Treasury; make it rely on a specified AMM
* add governance for new collateral types to Treasury, depending on the new AMM. The Treasury will assume it doesn't need to create a new pool

The plan for governance control of adding new collateral types is
* a generic mechanism in the governance package for queuing up a question that will invoke contract-provide code if it passes, and 
* code in the treasury to plug into that mechanism and add a new collateral type. I expect the code in the treasury to 
    * check preconditions
    * format a question spec
    * carry out the vote if successful.
 * The new mechanism will expect to rely on the same electorate used for parameter governance.

I think this approach leaves paramManager untouched. It may add functionality to contractHelper.

dckc:
@Chris-Hibbert for the demo (#3913) the thing I need most urgently is a test case - or at least a somewhat detailed sketch of it.

Would you please draft something?

_p.s. we talked thru enough details for now._

Tartuffo:
This is a bootstrap value that Governance can change, consumed by code in the top of the Vault Factory.

dckc:
> The special purpose function includes creating a vaultManager, endowing/finding an appropriate AMM, and connecting the two up as well as setting up the parameters for the new pool & vaultMgr.

I don't see why it's any more than just calling `addVaultType`, as in:

[...]

Issue #5218, opened 2022-04-25
clean up PSM bootstrap


dckc:
## What is the Problem Being Solved?

`writePsmScript.js` from #5215 is not a maintainable approach.

## Test Plan

cc @michaelfig @turadg

dckc:
@turadg writes in https://github.com/Agoric/agoric-sdk/pull/5257#discussion_r867239559 :
> consider having writePsmScript.js format its output.

I expect writePsmScript.js to go away altogether.

PR #5257, opened 2022-04-29
test: governance for vault collateral, oracles

dckc:
refs: #3924
refs: #4142

## Description

toward unit test for the plan to have the BLDer DAO
  - [x] elect the gov committee (#5169)
  - [x] add an ibc denom as an "ATOM" ERTP asset
  - [x] register a scaled price authority to connect IBC ATOM to Oracle ATOM and RUN to oracle USD
  - [x] add "ATOM" as collateral with 0 debt limit

Then add funds to the AMM via the reserve by gov ctte action.

I think we have code for all the pieces but only the 1st step has a working test, so far.

https://github.com/Agoric/agoric-sdk/issues/3924#issuecomment-1112699027

### Security Considerations

demonstrates secure governance

?

### Testing Considerations

This may be an odd mix of one-off code.


packages/deployment/Dockerfile.sdk
```diff
@@ -66,7 +66,7 @@ RUN \
 # Remove dev dependencies.
 RUN rm -rf packages/xsnap/moddable
 # FIXME: This causes bundling differences.  https://github.com/endojs/endo/issues/919
-# RUN yarn install --frozen-lockfile --production --network-timeout 100000
+RUN yarn install --frozen-lockfile --production --network-timeout 100000
```
dckc: was this on purpose?

packages/deploy-script-support/src/coreProposalBehavior.js
```diff
@@ -76,9 +76,13 @@ export const makeCoreProposalBehavior = ({
     );
 
     // Publish the installations for behavior dependencies.
-    entries(installations || {}).forEach(([key, value]) => {
-      produceInstallations[key].resolve(value);
-    });
+    const installAdmin = E(agoricNamesAdmin).lookupAdmin('installation');
+    await Promise.all(
+      entries(installations || {}).map(([key, value]) => {
+        produceInstallations[key].resolve(value);
```
dckc, line 91: I wonder about the timing of this. I see `scaledPriceAuthority` was moved out of the argument destructuring. Was that because of types or because this mechanism didn't work in time?

packages/run-protocol/src/core-proposal.js
```diff
@@ -141,9 +141,6 @@ const REWARD_MANIFEST = harden({
       runStakeCreatorFacet: true,
       zoe: true,
     },
-    produce: {
-      distributor: 'distributeFees',
-    },
```
dckc, line 146: this was just noise / dead code?

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -130,14 +130,15 @@ export const addAssetToVault = async ({
...
+export const getManifestForAddAssetToVault = (
   { restoreRef },
-  { denom, scaledPriceAuthority },
+  { denom, scaledPriceAuthorityRef },
 ) => {
   return {
     manifest: {
       addInterchainAsset: {
         consume: { zoe: true, bankManager: true, agoricNamesAdmin: true },
+        produce: { ibcAtomMintForTesting: true },
```
dckc: Passing this mint out for testing is an expedient hack. hm.

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -214,6 +243,8 @@ const makeScenario = async t => {
       t.context.installation.voting,
     ); // kludge
 
+    process.env.INTERCHAIN_DENOM = 'ibc/abc123';
```
dckc: I'm still struggling to make my peace with passing things thru `process.env`.

packages/run-protocol/src/core-proposal.js
```diff
@@ -270,6 +267,10 @@ export const SIM_CHAIN_POST_BOOT_MANIFEST = harden({
 const roleToGovernanceActions = harden({
   chain: CHAIN_POST_BOOT_MANIFEST,
   'sim-chain': SIM_CHAIN_POST_BOOT_MANIFEST,
+  'run-preview': {
+    ...ECON_COMMITTEE_MANIFEST,
+    ...CHAIN_POST_BOOT_MANIFEST,
```
dckc, line 272: The name `CHAIN_POST_BOOT_MANIFEST` is getting worse and worse.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -169,14 +169,12 @@ const start = async (zcf, privateArgs) => {
     { addLiquidityToAmmPool },
   );
 
-  /** @typedef {typeof creatorFacet} ReserveCreatorFacet */
```
dckc, line 172: Did this not work?

packages/run-protocol/test/supports.js
```diff
@@ -84,7 +85,7 @@ export const setupBootstrap = (t, optTimer = undefined) => {
   brand.produce.RUN.resolve(runBrand);
   issuer.produce.RUN.resolve(runIssuer);
 
-  return { produce, consume, ...spaces };
+  return { produce, consume, modules: { utils: { ...utils } }, ...spaces };
```
dckc, line 88: I wonder about adding ERTP and contractSupport (for `makeRatio`) to our list of available modules.

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -74,15 +57,26 @@ const makeTestContext = async () => {
   const install = (src, dest) =>
     bundleCache.load(src, dest).then(b => E(zoe).install(b));
   const installation = {
-    liquidate: install(contractRoots.liquidate, 'liquidateMinimum'),
-    VaultFactory: install(contractRoots.VaultFactory, 'VaultFactory'),
-    amm: install(contractRoots.amm, 'amm'),
-    reserve: install(contractRoots.reserve, 'reserve'),
     mintHolder: install(contractRoots.mintHolder, 'mintHolder'),
     voting: install(contractRoots.voting, 'voting'),
```
dckc: It looks like `voting` should go away from here for similar reasons.

Maybe `mintHolder` stays because it gets installed at chain bootstrap?

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -182,29 +167,73 @@ const makeScenario = async t => {
...
+    };
 
-    const USD = makeIssuerKit('USD');
-    const ATOM = makeIssuerKit('ATOM');
-    space.oracleBrand.produce.USD.resolve(USD.brand);
+  /**
+   * @param {string[]} coreProposals
+   */
+  const evalProposals = async coreProposals => {
+    const { code } = await extractCoreProposalBundles(
```
dckc, line 199: `extractCoreProposalBundles` closes over ambient `fs` and such in a way that's making this hard for me to follow.
Let's please
  1. make those overrideable defaults a la `writeFile = fs.promises.writeFile` in `makeWriteCoreProposal` and
  2. override them when we use them here in order to be explicit about authority

packages/deployment/Dockerfile.sdk
michaelfig: No, accident!  (See the comma.  I mean "No, it was an accident!")

packages/run-protocol/src/core-proposal.js
michaelfig: Yup.

packages/run-protocol/test/test-gov-collateral.js
michaelfig: We should fix that as technical debt that needs attention.  But only when we're bored.

packages/run-protocol/src/reserve/assetReserve.js
michaelfig: No, it didn't.  The typedefs were scoped to the function they appeared in, so they weren't reachable from other modules.

packages/run-protocol/test/test-gov-collateral.js
michaelfig: Yes, that's right.  I didn't get back to removing `voting` after I added the `committee-invite-core.js`.

packages/run-protocol/test/test-gov-collateral.js
michaelfig: Agreed.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
michaelfig: I kept the mints around even not when testing because I don't want to drop anything that valuable on the floor.

packages/run-protocol/test/test-gov-collateral.js
dckc: perhaps in a follow-on PR

packages/run-protocol/package.json
```diff
@@ -10,10 +10,12 @@
   "scripts": {
     "build": "yarn build:bundles",
     "build:bundles": "node ./scripts/build-bundles.js",
-    "build:eval:main": "agoric deploy ./scripts/init-core.js",
+    "prep:eval:main": "agoric deploy ./scripts/init-core.js",
```
turadg, line 13: 👍 prep != build

packages/run-protocol/package.json
```diff
@@ -10,10 +10,12 @@
...
+    "prep:eval:collateral": "agoric deploy ./scripts/add-collateral-core.js",
+    "prep:eval:invite-committee": "agoric deploy ./scripts/invite-committee-core.js",
     "deploy:contracts": "agoric deploy scripts/deploy-contracts.js",
     "build:eval-permit:psm": "node src/psm/writePsmScript.js --permit >scripts/gov-startPSM-permit.json",
-    "build:eval-code:psm": "node src/psm/writePsmScript.js >scripts/gov-startPSM.js; yarn prettier --write scripts/gov-startPSM.js",
+    "prep:eval-code:psm": "node src/psm/writePsmScript.js >scripts/gov-startPSM.js; yarn prettier --write scripts/gov-startPSM.js",
```
turadg, line 18: out of scope but consider having `writePsmScript.js` format its output. https://prettier.io/docs/en/api.html

packages/run-protocol/src/committee-proposal.js
```diff
@@ -0,0 +1,105 @@
+// @ts-check
+import { E } from '@endo/far';
+
+// must match packages/wallet/api/src/lib-wallet.js
+const DEPOSIT_FACET = 'depositFacet';
+
+const { values } = Object;
+
+/** @type { <X, Y>(xs: X[], ys: Y[]) => [X, Y][]} */
```
turadg, line 9: TIL you can inline generics in `@type`

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const brand = await E(issuer).getBrand();
+  const kit = { mint, issuer, brand };
+
+  // Create the mint list if it doesn't exist and wasn't already rejected.
+  produceInterchainMints.resolve([]);
+  Promise.resolve(interchainMints).then(
+    mints => mints.push(mint),
+    () => {}, // If the interchainMints list was rejected, ignore the error.
+  );
```
turadg, line 53: isn't `interchainMints` a promise?
and the error handling is already optional so no point in sending a noop.
```suggestion
  interchainMints.then(mints => mints.push(mint));
```

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -0,0 +1,452 @@
...
+  inviteCommittee: '../scripts/invite-committee-core.js',
+};
+
+const voterAddresses = {
+  Rowland: `agoric1qed57ae8k5cqr30u5mmd46jdxfr0juyggxv6ad`,
+  Bill: `agoric1xgw4cknedau6xhrlyn6c8e40d02mejee8gwnef`,
+  Dan: `agoric1yumvyl7f5nkalss7w59gs6n3jtqv5gmarudx55`,
+};
+
+const makeTestContext = async () => {
```
turadg, line 48: this is only used within `test.before`. consider inlining it to avoid footgun of calling it within a test.

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -0,0 +1,452 @@
...
+    runKit: { brand: runBrand, issuer: runIssuer },
+    installation,
+  };
+};
+
+test.before(async t => {
+  t.context = await makeTestContext();
+});
+
+const makeScenario = async t => {
```
turadg, line 92: each test begins with `const s = await makeScenario(t);` consider in `test.before`,
```
t.context.scenario = await makeScenario(t);
```
then within test,
```
const { s } = t.context.scenario;
```

michaelfig (approved):
Yaaay!  Onward and upward!

packages/vats/src/core/types.js
```diff
@@ -78,7 +78,10 @@
  */
 
 /**
- * @typedef {{ resolve: (v: ERef<T>) => void, reset: () => void}} Producer<T>
+ * @typedef {Object} Producer<T>
+ * @property {(v: ERef<T>) => void} resolve
+ * @property {(r: any) => void} reject
```
turadg: nit
```suggestion
 * @property {(r: unknown) => void} reject
```

packages/vats/src/distributeFees.js
```diff
@@ -62,5 +62,7 @@ export async function buildDistributor(
   };
 
   const epochNotifier = E(epochTimer).makeNotifier(0n, epochInterval);
-  return observeNotifier(epochNotifier, timeObserver);
+  observeNotifier(epochNotifier, timeObserver).catch(e => {
+    console.error('fee distributor failed with', e);
```
turadg, line 66: loud enough?

turadg (commented):
I don't know the bootstrap system well enough to approve. I left some generic JS observations.

packages/run-protocol/package.json
michaelfig: We'll likely merge writePsmScript.js with the rest of the proposal generation mechanism.  But yes, that's a good suggestion for that refactor.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
michaelfig: Without the noop and if interchainMints is resolved to a rejection, this will produce a fresh unhandled rejection which we don't want.  Leaving as-is.

packages/vats/src/distributeFees.js
michaelfig: Yep!

packages/run-protocol/package.json
dckc: > We'll likely merge writePsmScript.js with the rest of the proposal generation mechanism.

The IOU is #5218

packages/run-protocol/test/test-gov-collateral.js
dckc: its type is used everywhere...

```js
/** @type {import('ava').TestInterface<Awaited<ReturnType<makeTestContext>>>} */
// @ts-expect-error cast
const test = anyTest;
```

packages/run-protocol/test/test-gov-collateral.js
dckc: I think I put as much stuff as is shareable in `t.context`. `makeScenario()` makes mutable stuff; if we shared that between tests, it would be bad, I think.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
turadg: > produce a fresh unhandled rejection which we don't want

To make sure I understand, this is to be explicit about ignoring the rejection? Is there a functional difference from noop handling or merely never handling? (in this case in which there's no await)

packages/vats/src/distributeFees.js
dckc: more to the point... @michaelfig tells me he has plans to systematically address unhandled rejections, but it depend on work in endo...

When that's in place, we can get rid of this sort of `catch`.

packages/run-protocol/test/test-gov-collateral.js
turadg: Ah right. It would have to be in `test.beforeEach`. https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#test-context

packages/run-protocol/test/supports.js
turadg: FYI @Chris-Hibbert I looked into removing `modules:` and it's needed for bundling. For example, `lmbd` dep.

PR #5258, opened 2022-04-29
chore: Improve StorageAPI-related comments and code consistency

gibson042:
## Description

Mostly adds more explanation to comments, and also some TODOs about potential improvements.

### Security Considerations

n/a

Nothing specific

### Testing Considerations

n/a


packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -1,7 +1,7 @@
 import { assert } from '@agoric/assert';
 import { insistStorageAPI } from '../../lib/storageAPI.js';
 
-// We manage a host-realm Storage object with a has/getKeys/get/set/del API.
+// We manage a host-realm Storage object with a has/getKeys/get/set/delete API.
```
warner: We no longer use multiple Realms, so let's say "we wrap a host-provided Storage object with ..." and remove the bit about cross-realm contamination.

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -14,8 +14,10 @@ import { insistStorageAPI } from '../../lib/storageAPI.js';
 // xenophobia.
 
 /**
- * Given two iterators over ordered sequences, produce a new iterator that will
- * iterate in order over the merged output of the two iterators.
+ * Given two iterators over sequences of comparable unique elements
```
warner: wrap to 80-ish cols

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -36,6 +38,7 @@ function* mergeSortedIterators(it1, it2) {
       v2 = it2.next();
       yield result;
     } else {
+      // TODO? assert(v1.value > v2.value);
```
warner: That could only fail if there's a NaN involved, yeah? Or if `.value` is an unstable getter, or something with an unstable `.toString`? Maybe just update the comment to indicate this is only meant for strings.

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -124,6 +128,7 @@ export function buildCrankBuffer(
         added.values(),
         kvStore.getKeys(start, end),
       )) {
+        // TODO: Remove this redundant range verification?
```
warner: maybe, yeah, but it's cheap and confidence-inspiring

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -193,7 +198,9 @@ export function buildCrankBuffer(
     if (oldActivityhash === undefined) {
       oldActivityhash = '';
     }
-    const hasher = createSHA256('activityhash\n');
+    const hasher = createSHA256();
```
warner, line 197: ok, if you think it's easier to read this way

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -288,6 +288,12 @@ export default async function main(progname, args, { env, homedir, agcc }) {
   // instance, and we update the 'portNums.storage' value each time toSwingSet is called
   async function launchAndInitializeSwingSet(bootMsg) {
     // this object is used to store the mailbox state.
+    // TODO: Generalize StorageAPI caching for both makeChainStorage and buildCrankBuffer?
```
warner: The two caching layers have different lifetimes and ownership:
* the crankbuffer is owned by the kernel, invisible to the host application, and closed/flushed/written-back when the kernel decides to commit a crank, and abandoned when the kernel decides to abandon a crank
* the outer cache (the "block buffer") is owned by the host app, invisible to the kernel, commited when the host commits the rest of its state, and only abandoned if/when the host app is terminated

It'd be interesting to think about using nested DB transactions to hold both buffers directly in the DB, but 1: we'd need a different DB (LMDB only offers a single transaction), and 2: we'd have to rethink how we build the crankhash.

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -288,6 +288,12 @@ export default async function main(progname, args, { env, homedir, agcc }) {
...
+    // const sendToSwingset = msg => chainSend(portNums.storage, msg);
```
warner: do these comments contain proposed code additions, or are they reminders of where the values come from, or what? If they're proposals, we should probably put them in an issue where we can have more context about what they're intended to do.

warner (changes requested):

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -744,7 +745,7 @@ export default function makeKernelKeeper(
       assert(k.startsWith(exportPrefix), k);
       // The void for an object exported by a vat will always be of the form
       // `o+NN`.  The '+' means that the vat exported the object (rather than
-      // importing it) and therefor the object is owned by (i.e., within) the
+      // importing it) and therefore the object is owned by (i.e., within) the
```
FUDCo, line 748: Very nice grammar nerd move there.  Respect.

FUDCo (commented):
Good pending Brian's issues.

packages/cosmic-swingset/src/chain-main.js
gibson042: They were conversation starters for proposed changes, which seem worth at least discussion. I have removed them and will followup in a separate PR.

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -53,14 +54,12 @@ const makeChainStorage = (call, prefix = '', imp = x => x, exp = x => x) => {
       changedKeys.add(key);
     },
     get(key) {
-      if (cache.has(key)) {
-        // Our cache has the value.
-        return cache.get(key);
-      }
+      if (cache.has(key)) return cache.get(key);
```
warner, line 56: wow, you got that past eslint? more power to ya :)

warner (approved):
looks good to me

PR #5260, opened 2022-04-29
feat(vault): econ metrics notifiers

### LINKED ISSUE #4649: Expose key Vault Factory economic data via subscription

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose from the Vault Factory:

## Per Contract (Director object)
- ~Governance parameters, contract terms, fees~ [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- Allowed collateral types
- Current reward pool (fees collected so far)
- Number and amount of shortfall

## Per Collateral (Manager object)
- ~Governance parameters, including interest rate, liquidation ratio, liquidation penalty per collateral, current debt limit~  [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- ~Current debt limit~ [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- Number of vaults
- Total outstanding debt (vaultKit `publicNotifiers.asset`)
- Total collateral held
- Compounded interest (vaultKit `publicNotifiers.asset`)
- ~Liquidation strategy~ [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- Liquidation state (how much is currently being liquidated)
         1. Collateral being sold
         2. Amount sold so far
         3. Proceeds so far

## Per Vault
- Debt (`debtSnapshot`.debt)
- Collateral (`locked`)
- Interest snapshot (`debtSnapshot`.interest)
- Status (open, closed, liquidating, liquidated) `vaultState`

Use existing notifiers except for values that have high change frequency.

turadg:
closes: #4649

## Per Contract (Director object)
- [x] ~Governance parameters, contract terms, fees~ [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- [x] Allowed collateral types
- [x] Current reward pool (fees collected so far)
- [X] ~Number and amount of shortfall~ (DEFERRED https://github.com/Agoric/agoric-sdk/issues/5300)

## Per Collateral (Manager object)
- [x] ~Governance parameters, including interest rate, liquidation ratio, liquidation penalty per collateral, current debt limit~  [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- [x] ~Current debt limit~ [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- [x] Number of vaults
- [x] Total outstanding debt (vaultKit `publicNotifiers.asset`)
- [x] Total collateral held
- [x] Compounded interest (vaultKit `publicNotifiers.asset`)
- [x] ~Liquidation strategy~ [governed](https://github.com/Agoric/agoric-sdk/issues/5268)
- [x] ~Liquidation state~ (DEFERRED https://github.com/Agoric/agoric-sdk/issues/5300)

## Per Vault
- [x] Debt (`debtSnapshot`.debt)
- [x] Collateral (`locked`)
- [x] Interest snapshot (`debtSnapshot`.interest)
- [x] Status (open, closed, liquidating, liquidated) `vaultState`

Use existing notifiers except for values that have high change frequency.

### Security Considerations

Everything being exposed here is intentional.

We will need to update these docs. Waiting on technical writer.

### Testing Considerations

New tests for notifier output.


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -33,6 +33,13 @@ import {
 const { details: X } = assert;
 
 /**
+ * @typedef {{
+ * governedValues: {}, // ??? what type? ??? aren't there already subscriptions?
```
turadg: Since governance is already on-chain, do we need to republish all the values or can we point to them somehow?

Also, @Chris-Hibbert WDYT of ParamManager having methods to support this?

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -72,13 +80,17 @@ const initState = (zcf, directorParamManager, debtMint) => {
 
   const vaultParamManagers = makeScalarMap('brand');
 
+  // XXX need a way to register this data stream for consumption
+  const { updater: econMetrics } = makeNotifierKit();
```
turadg: @gibson042 I punted on mocking this because I wasn't sure where to put it. seems like @dtribble may have some designs in mind so I don't want to proceed too far.

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: It absolutely makes sense for ParamManager to provide support. I'm waiting on more detailed requirements. 

I'm hoping to hear
 * what publishing/sharing mechanisms are needed
 * whether we want to emit events or statistics. At ${JOB-1} the s/w mostly said "x happened" and the monitoring systems did all the counting, averaging, etc. I can imagine we want more to happen in the contracts, so we'd emit counts, averages, maximums, etc.

packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -37,7 +38,7 @@
  * @param {Issuer} collateralIssuer
  * @param {Keyword} collateralKeyword
  * @param {VaultManagerParamValues} params
- * @returns {Promise<VaultManager>}
+ * @returns {Promise<VaultManagerObject>}
```
dtribble: "Object" is a terrible thing to stick in a name. It's completely vacuous.

packages/run-protocol/src/vaultFactory/prioritizedVaults.js
```diff
@@ -174,6 +174,7 @@ export const makePrioritizedVaults = (reschedulePriceCheck = () => {}) => {
     addVault,
     entries: vaults.entries,
     entriesPrioritizedGTE,
+    getSize: vaults.getSize,
```
dtribble: "count" would be much better "size". "Size" could mean any number of things (total market cap?  AUM? disk space used?)

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -66,14 +66,15 @@ const validTransitions = {
 /**
  * @typedef {Phase[keyof typeof Phase]} TitlePhase
  *
- * @typedef {object} VaultUIState
+ * @typedef {object} VaultTitleState
```
dtribble: "Title" has the wrong connotations:
- it's a thing looked up in a registry
- it feels ACL-based
- it's primarily the meta-data, not the authority. 

Thus the "title" woudl be more the invitation made in a transfer request (I sign over the title to you and then you get the keys to the property)

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -66,14 +66,15 @@ const validTransitions = {
 /**
  * @typedef {Phase[keyof typeof Phase]} TitlePhase
  *
- * @typedef {object} VaultUIState
```
dtribble, line 69: We need a general term for the state snapshot from a notifier. We could remove "UI" from that with "vaultSnapshot"? but "vaultState" isn't terrible.

packages/run-protocol/src/vaultFactory/vault.js
dtribble: This is not just being pedantic. Using a term "near but wrong" would be a larger cause of deep confusion than using a new or clumsy term. Hence, "outer" would still be better. 
- ref
- outer
- handle (but used elsewhere ina. consistent way without operations)
- proxy (but used in JS already)
- front
- facet
- facade (implies fake though)
- aspect (but "aspect oriented programming polluted that)

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -254,6 +275,7 @@ const machineBehavior = {
       }
       // TODO add aggregate debt tracking at the vaultFactory level #4482
       // totalDebt = AmountMath.add(totalDebt, toMint);
+      facets.machine.notifyEcon();
```
dtribble: what is "machine"?

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -33,15 +33,23 @@ import {
...
  * debtMint: ZCFMint<'nat'>,
- * collateralTypes: Store<Brand,VaultManager>,
+ * collateralTypes: Store<Brand,VaultManagerObject>,
  * electionManager: Instance,
  * directorParamManager: import('@agoric/governance/src/contractGovernance/typedParamManager').TypedParamManager<{VaultDirectorParams}>,
  * mintSeat: ZCFSeat,
  * penaltyPoolSeat: ZCFSeat,
  * rewardPoolSeat: ZCFSeat,
  * vaultParamManagers: Store<Brand, import('./params.js').VaultParamManager>,
+ * econNotifier: Notifier<EconState>
```
dtribble: "econ" isn't great but it may be OK.  We should use the same naming across all the contracts.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -237,12 +255,23 @@ const helperBehavior = {
       compoundedInterest: state.compoundedInterest,
       interestRate,
       latestInterestUpdate: state.latestInterestUpdate,
-      totalDebt: state.totalDebt,
+      // XXX move to EconState and type as present with null
       liquidatorInstance: state.liquidatorInstance,
```
dtribble, line 263: That should be in governance notifier I would think

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -293,32 +293,36 @@ async function setupServices(
...
-    vaultFactoryCreatorFacet,
+    vaultFactoryCreatorFacetP,
     E(governorCreatorFacet).getPublicFacet(),
     aethVaultManagerP,
     pa,
   ]);
-  trace(t, 'pa', { governorInstance, vaultFactory, lender, priceAuthority });
+  trace(t, 'pa', {
+    governorInstance,
+    vaultFactory: vaultFactoryCreator,
```
dtribble: remove `vaultFactory:`

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -364,7 +368,11 @@ test('first', async t => {
     undefined,
     500n,
   );
-  const { vaultFactory, lender, aethVaultManager } = services.vaultFactory;
+  const {
+    vaultFactoryCreator: vaultFactory,
```
dtribble: it would be good to not have this property renaming.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -293,32 +293,36 @@ async function setupServices(
...
     aethIssuer,
     'AEth',
     rates,
   );
-  /** @type {[any, VaultFactory, VFC['publicFacet']]} */
-  // @ts-expect-error cast
   const [
     governorInstance,
-    vaultFactory,
+    vaultFactoryCreator,
```
dtribble: this renaming causes a lot of churn. If you really want it, the uses below need to be updated as well so we dont' have a bunch of `vaultFactory: vaultFactoryCreator`. Given that the type system clarifies that it's the creator facet, this seems unnecessary.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -784,7 +792,7 @@ test('vaultFactory display collateral', async t => {
     500n,
   );
 
-  const { vaultFactory } = services.vaultFactory;
+  const { vaultFactoryCreator: vaultFactory } = services.vaultFactory;
```
dtribble: As above

dtribble (dismissed):
Generally great, but some namings and renamings are problematic. Note that those could mostly be not entangled with the core of this change.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: goal was to not have `vaultFactory.vaultFactory` as before. granted the type system indicates they're different, but just eyeballing it's not clear.

In general I think we would benefit from more consistency in how we name facets, contracts, and virtual/durable objects in contracts. That's a bigger endeavor that is out of scope so I'll revert this change.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: That term started here https://github.com/Agoric/agoric-sdk/commit/15322a1f65bb67e1865f8276600a76171696dc74#diff-0b7d727a28b84089ac4ebe9b76f8957bb71bfee261640828bad01c4189cbe73fR283

I kept it because it distinguishes between the "creator facet" of the contract and the "limited creator facet" that it can return. I think we should move from "this thing is for what consumer" to "what this thing does or how". (For the same reasons that "public notifiers" is better than "UI notifiers")

packages/run-protocol/src/vaultFactory/vault.js
turadg: > This is not just being pedantic. Using a term "near but wrong" would be a larger cause of deep confusion than using a new or clumsy term.

Agree naming is important and "near but wrong' can be worse that obviously wrong.

> Thus the "title" woudl be more the invitation made in a transfer request

I don't agree with the bullets supporting it but I agree with this conclusion.

Note that this just a typename. In the vault offer result it's,
```
    publicNotifiers: {
      vault: title.getNotifier(),
      asset: assetNotifier,
    },
```

So how about `VaultNotifierState`? `VaultState` could also work but given how easy it is to change type names I'd err towards more explicit and if we find over time the disambiguation is unnecessary we can shorten it.

EDIT: I went with `VaultNotification`.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Came up above and proposed `Notification`. I could see that being used for the wrapping object (with `value` and `updateCount`) but that's so generic as to not need a name.

packages/run-protocol/src/vaultFactory/types.js
turadg: Agreed. `VaultManager` was used already by `ReturnType<typeof makeVaultManagerKit>['manager']` so I was trying to make a small change but I'll make the better change.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -237,12 +256,23 @@ const helperBehavior = {
...
       latestInterestUpdate: state.latestInterestUpdate,
-      totalDebt: state.totalDebt,
+      // XXX move to governance and type as present with null
       liquidatorInstance: state.liquidatorInstance,
     });
     state.assetUpdater.updateState(payload);
   },
 
+  /** @param {MethodContext} context */
+  metricsNotify: ({ state }) => {
```
Chris-Hibbert: notify would conventionally be a fine name, but with notifier/updater the notifier half is about getting notified and the updater half is about sending notifications. I think this needs to be `updateMetrics`.

Chris-Hibbert: I think `metricNotify` should be called on `liquidateAll()`, but I don't see how it will be.

Do individual liquidations invoke it?

Chris-Hibbert:
> /home/runner/work/agoric-sdk/agoric-sdk/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Error:   43:1  error  `@agoric/notifier` import should occur before import of `../../src/makeTracer.js`  import/order

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: d1b23440f7a9184ea3f3873b70f68f5c508cbea1

turadg: They should. 41fa373c0f7660122cbce86b5f607c8b526a5009

though I ran into an accounting issue that isn't obvious to solve and will be changed by https://github.com/Agoric/agoric-sdk/pull/5343 so I'll come back to this after that.

Chris-Hibbert (approved):

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -34,11 +38,15 @@ const trace = makeTracer('VM', false);
  *  compoundedInterest: Ratio,
  *  interestRate: Ratio,
  *  latestInterestUpdate: bigint,
- *  totalDebt: Amount<'nat'>,
```
samsiegart, line 37: This caused a null reference error in the UI https://github.com/Agoric/dapp-treasury/blob/main/ui/src/components/vault/VaultConfigure/VaultConfigure.jsx#L200. I guess typescript/CI didn't catch this somehow...

turadg: Ah, I didn't realize that was being used. Yeah, the dapp-treasury repo currently has no type checking https://github.com/Agoric/dapp-treasury/issues/97

PR #5264, opened 2022-04-30
Ratios for `priceAggregator`

michaelfig:
## Description

More improvements for Agoric/dapp-oracle#33.  This can be reviewed commit-by-commit

- accept Ratios as input to the priceAggregator and use them internally
- more tolerant (and accurate) brand rules for several Ratio operations, based on working out the algebra
- `parseRatio` and `subtractRatios` to fill some API gaps
- propagate VatData into solo-spawned vats, so that ERTP doesn't crash due to its absence

### Security Considerations

Should make the priceAggregator more reviewable and easy to verify that it calculates prices correctly.

### Testing Considerations

Tested end-to-end against dapp-oracle.

### DIFF
--- a/packages/zoe/src/contractSupport/ratio.js
+++ b/packages/zoe/src/contractSupport/ratio.js
@@ -195,20 +195,60 @@ export const invertRatio = ratio => {
 export const addRatios = (left, right) => {
   assertIsRatio(right);
   assertIsRatio(left);
-  assert(
-    left.numerator.brand === left.denominator.brand &&
-      left.numerator.brand === right.numerator.brand &&
-      left.numerator.brand === right.denominator.brand,
-    X`all brands must match:  ${q(left)} ${q(right)}`,
+  assert.equal(
+    left.numerator.brand,
+    right.numerator.brand,
+    X`numerator brands must match: ${q(left)} ${q(right)}`,
+  );
+  assert.equal(
+    left.denominator.brand,
+    right.denominator.brand,
+    X`denominator brands must match: ${q(left)} ${q(right)}`,
+  );
+
+  // Simplifying the expression:
+  //  (and + bnd) / y d**2
+  //  (a + b) nd / y d**2
+  //  ((a + b) n / y d) * (d / d)
+  //  (a + b) n / yd
+  return makeRatio(
+    add(
+      multiply(left.numerator.value, right.denominator.value), // a nd
+      multiply(left.denominator.value, right.numerator.value), // b nd
+    ), // (a + b) nd
+    left.numerator.brand,
+    multiply(left.denominator.value, right.denominator.value), // y d**2
+    left.denominator.brand,
+  );
+};
+
+/**
+ * @param {Ratio} left
+ * @param {Ratio} right
+ * @returns {Ratio}
+ */
+export const subtractRatios = (left, right) => {
+  assertIsRatio(right);
+  assertIsRatio(left);
+  assert.equal(
+    left.numerator.brand,
+    right.numerator.brand,
+    X`numerator brands must match: ${q(left)} ${q(right)}`,
+  );
+  assert.equal(
+    left.denominator.brand,
+    right.denominator.brand,
+    X`denominator brands must match: ${q(left)} ${q(right)}`,
   );
 
   return makeRatio(
-    add(
-      multiply(left.numerator.value, right.denominator.value),
-      multiply(left.denominator.value, right.numerator.value),
-    ),
+    subtract(
+      multiply(left.numerator.value, right.denominator.value), // a nd
+      multiply(left.denominator.value, right.numerator.value), // b nd
+    ), // (a - b) nd
     left.numerator.brand,
-    multiply(left.denominator.value, right.denominator.value),
+    multiply(left.denominator.value, right.denominator.value), // y d**2
+    left.denominator.brand,
   );
 };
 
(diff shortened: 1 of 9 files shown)


packages/zoe/test/unitTests/contractSupport/test-ratio.js
```diff
@@ -396,3 +519,49 @@ test('ratio - quantize', t => {
     );
   }
 });
+
+test('ratio - parse', t => {
+  const { brand: moeBrand } = makeIssuerKit('moe');
+  const { brand: larryBrand } = makeIssuerKit('larry');
+
+  t.deepEqual(
+    parseRatio(1024.93803, moeBrand),
```
dckc, line 528: Add a case for a float such as 0.1 + 0.2 that isn't directly representable?

packages/zoe/src/contractSupport/ratio.js
```diff
@@ -195,20 +195,30 @@ export const invertRatio = ratio => {
 export const addRatios = (left, right) => {
   assertIsRatio(right);
   assertIsRatio(left);
-  assert(
-    left.numerator.brand === left.denominator.brand &&
-      left.numerator.brand === right.numerator.brand &&
-      left.numerator.brand === right.denominator.brand,
-    X`all brands must match:  ${q(left)} ${q(right)}`,
```
dckc, line 202: Code could be relying on this check. So this is a breaking change, in a way.

packages/zoe/src/contractSupport/ratio.js
```diff
@@ -195,20 +195,30 @@ export const invertRatio = ratio => {
...
+    right.numerator.brand,
+    X`numerator brands must match: ${q(left)} ${q(right)}`,
+  );
+  assert.equal(
+    left.denominator.brand,
+    right.denominator.brand,
+    X`denominator brands must match: ${q(left)} ${q(right)}`,
   );
 
+  // Simplifying the expression:
```
dckc, line 209: this makes me wish for property testing.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -35,15 +39,17 @@ const start = async (
...
-  } = zcf.getTerms();
+  const { timer, POLL_INTERVAL, brandIn, brandOut } = zcf.getTerms();
 
-  const unitIn = AmountMath.getValue(brandIn, unitAmountIn);
+  /** @param {ERef<Brand>} brand */
+  const getDecimalP = async brand => {
+    const displayInfo = E(brand).getDisplayInfo();
+    return E.get(displayInfo).decimalPlaces;
+  };
+  const [decimalPlacesIn = 0, decimalPlacesOut = 0] = await Promise.all([
```
dckc: We don't normally rely on displayInfo for correctness (only for display). Please add a comment justifying this exception.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -112,21 +118,21 @@ const start = async (
...
-      const valueOutForUnitIn =
-        overrideValueOut === undefined
-          ? lastValueOutForUnitIn // Use the latest value.
-          : overrideValueOut; // Override the value.
-      if (valueOutForUnitIn === undefined) {
+      // Use the current price.
+      const price =
+        overridePrice === undefined
+          ? lastPrice // Use the latest value.
+          : overridePrice; // Override the value.
```
dckc: More clear as `const price = overridePrice ||  lastPrice;`?

packages/zoe/src/types.js
```diff
@@ -43,5 +43,6 @@
  */
 
 /**
+ * @template [OfferResult]
```
dckc, line 46: where is this used?

Is it straightforward to demonstrate how it's used in a test?

dckc (commented):
I'm not quite sure about approving the change to parameterize Invitation without understanding it better.

The use of displayInfo.decimalPlaces for more than display should have explicit documentation.

In sum, I'd like to discuss this a bit more before approving.

packages/zoe/src/types.js
michaelfig: It's used in `packages/zoe/src/zoeService/types.js#Offer` as the type for `zoe.offer`'s invitation, and before introducing this optional template parameter, the type of the first arg to `zoe.offer` failed and was silently replaced by `any`.  The producing code is in `packages/zoe/src/contractFacet/types.js#MakeInvitation`, which will propagate the offer result types.

The reason this didn't cause a type error is that Zoe extensively uses the `@callback` tag, which TypeScript currently doesn't report `@param` type syntax or semantic problems.  It's a known issue in TS, but I don't have the link handy.

dckc: Ok. I'm familiar with the ts bug.

packages/zoe/src/contractSupport/ratio.js
michaelfig: I would be surprised to find that code expects an exception even if the operation can produce correct ratios.  That, to me, would indicate a bug in the reliant code.

dckc: I'd be surprised too. Since we're pre-1.0, I don't feel much obligation to support code outside our own. So this is mostly a somewhat theoretical / forward-looking concern.

packages/zoe/src/contractSupport/ratio.js
michaelfig: > this makes me wish for property testing.

Agreed!  I would gladly throw property testing against the ratio library, if I knew how to use it.

dckc: it's reasonably straightforward: https://github.com/Agoric/agoric-sdk/blob/master/packages/ERTP/test/unitTests/test-amountProperties.js

dckc (approved):
This looks ready to go, with or without property testing.

PR #5265, opened 2022-04-30
TypeScript 4.7

turadg:
_no tickets_

## Description

[Release notes for TypeScript 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-rc//) describe many improvements useful to this codebase.

For example a bug fix that found this problem 03446a1c4ae3d22fb10b063510e179662462d043

One that may prove useful:
> a new preview editor command for [Go to Source Definition](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-rc/#go-to-source-definition). This can be helpful in cases where an ordinary Go to Definition comand would take you to a declaration file instead of the actual JavaScript or TypeScript source. [We’re still looking for feedback on this feature](https://github.com/microsoft/TypeScript/issues/49003)!

### Security Considerations

--

--

### Testing Considerations

- [x] wait for TS 4.7 release and verify CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lerna": "^3.22.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.6.2",
-    "typescript": "~4.6.3"
+    "typescript": "~4.7.2"
   },
   "engines": {
     "node": ">=14.15.0"
--- a/yarn.lock
+++ b/yarn.lock
@@ -16770,10 +16770,10 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@^4.3.2, typescript@~4.6.3:
-  version "4.6.3"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.6.3.tgz#eefeafa6afdd31d725584c67a0eaba80f6fc6c6c"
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "lerna": "^3.22.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.6.2",
-    "typescript": "~4.6.3"
+    "typescript": "beta"
```
mhofman: I'd prefer a specific version (`4.7.0-beta`) instead of a generic `beta` label

mhofman (commented):
Looking forward to the control flow and inference improvements.

package.json
turadg: This is a draft PR and in the description I said the proposed change would be 4.7, i.e. `~4.7.0` here.

Are you saying you think `4.7.0-beta` is okay to merge to master?

package.json
```diff
@@ -24,7 +24,7 @@
     "lerna": "^3.22.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.6.2",
-    "typescript": "~4.6.3"
+    "typescript": "^4.7.1-rc"
```
mhofman: I know this is still draft, but given that `typescript` doesn't follow semver, we need to use `~` to only pick up patch updates and explicitly bump "minor" which are roughly majors for TS.

yarn.lock
```diff
@@ -16770,11 +16770,16 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@^4.3.2, typescript@~4.6.3:
+typescript@^4.3.2:
```
mhofman: Can you run a `yarn-deduplicate`?

mhofman (changes requested):

mhofman (approved):

mhofman:
Removed automerge label to leave priority to 5448.

PR #5269, opened 2022-05-02
bump eslint-plugin-jsdoc

turadg:
## Description

The version we have would error when doing `yarn install` under Node 18. (not in its list of supported engines)

The latest version changed https://github.com/gajus/eslint-plugin-jsdoc/pull/855 to align more with TypeScript. This fixes those warnings too (mostly with `yarn lint-fix` and then manually a few stragglers)

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -20,12 +20,12 @@ const unweakable = new WeakSet();
  *
  * @param {number} size  Maximum number of entries to keep in the cache before
  *    starting to throw them away.
- * @param {(baseRef: string) => Object} fetch  Function to retrieve an
+ * @param {(baseRef: string) => object} fetch  Function to retrieve an
  *    object's raw state from the store by its baseRef
- * @param {(baseRef: string, rawState: Object) => void} store  Function to
+ * @param {(baseRef: string, rawState: object) => void} store  Function to
  *   store raw object state by its baseRef
  *
- * @returns {Object}  An LRU cache of (up to) the given size
+ * @returns {object}  An LRU cache of (up to) the given size
  *
  * This cache is part of the virtual object manager and is not intended to be
  * used independently; it is exported only for the benefit of test code.
@@ -140,7 +140,7 @@ export function makeCache(size, fetch, store) {
  *   of virtual references.
  * @param {() => number} allocateExportID  Function to allocate the next object
  *   export ID for the enclosing vat.
- * @param { (val: Object) => string} getSlotForVal  A function that returns the
+ * @param {(val: object) => string} getSlotForVal  A function that returns the
  *   object ID (vref) for a given object, if any.  their corresponding export
  *   IDs
  * @param {*} registerValue  Function to register a new slot+value in liveSlot's
@@ -150,7 +150,7 @@ export function makeCache(size, fetch, store) {
  * @param {number} cacheSize  How many virtual objects this manager should cache
  *   in memory.
  *
- * @returns {Object} a new virtual object manager.
+ * @returns {object} a new virtual object manager.
  *
  * The virtual object manager allows the creation of persistent objects that do
  * not need to occupy memory when they are not in use.  It provides five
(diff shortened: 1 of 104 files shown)


mhofman:
I remember some problem with using `object` as the base type for typedefs, mostly that typescript in non strict mode considers it like `any`.

See https://github.com/microsoft/TypeScript/issues/47694#issuecomment-1031755370

turadg:
> I remember some problem with using object as the base type for typedefs, mostly that typescript in non strict mode considers it like any.

Oh my. Indeed.

Though it appears to do the same for `Object` so not relevant to this change?

mhofman:
> Though it appears to do the same for `Object` so not relevant to this change?

Right, `Object` as the base was wrong in the first place. As long as we've confirmed that there is no side-effect on `tsc` checks from using `object`, I'm ok with this change

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -20,12 +20,12 @@ const unweakable = new WeakSet();
  *
  * @param {number} size  Maximum number of entries to keep in the cache before
  *    starting to throw them away.
- * @param {(baseRef: string) => Object} fetch  Function to retrieve an
+ * @param {(baseRef: string) => object} fetch  Function to retrieve an
  *    object's raw state from the store by its baseRef
- * @param {(baseRef: string, rawState: Object) => void} store  Function to
+ * @param {(baseRef: string, rawState: object) => void} store  Function to
```
mhofman, line 25: Here is potentially a place where it'd revert back to `any`

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -140,7 +140,7 @@ export function makeCache(size, fetch, store) {
  *   of virtual references.
  * @param {() => number} allocateExportID  Function to allocate the next object
  *   export ID for the enclosing vat.
- * @param { (val: Object) => string} getSlotForVal  A function that returns the
+ * @param {(val: object) => string} getSlotForVal  A function that returns the
```
mhofman, line 143: I believe it's the same here, `val` would now become an `any` type.

mhofman (commented):
I need to pay closer attention to all the changes, and do some experiments in the TS Playground.

While `Object` -> `object` should be safe for any `@typedef {object}` or `@param {object}` used as the base of other `@param`, I'm not certain when used as return values or in TS style functions types it'd be safe.

I agree it would be great to fix all our `Object` usages, as they're all most likely wrong, but I'd rather get it right than change it into another wrong. Maybe we can file an issue and pick this up after MN-1, or there was a motivation for this change now besides declared support for Node 18?

packages/SwingSet/src/liveslots/virtualObjectManager.js
turadg: It was already `any`. That's my point in the statement,
> Though it appears to do the same for Object so not relevant to this change?

`Object`
Screen Shot 2022-05-03 at 12 19 02 PM


`object`
Screen Shot 2022-05-03 at 12 18 52 PM

packages/SwingSet/src/liveslots/virtualObjectManager.js
turadg: > now become an `any` type

Again, already was.

turadg:
> I need to pay closer attention to all the changes, and do some experiments in the TS Playground. …  I'd rather get it right than change it into another wrong.

What are your concerns? From what I see, all the `Object` are `any` before and after.

> Maybe we can file an issue and pick this up after MN-1, or there was a motivation for this change now besides declared support for Node 18?

Yes, motivation was to update the dep to support Node 18. I thought fixing the warnings would be good to include now and safe. If you don't think so I can remove that and suppress the warnings instead. But based on…

> As long as we've confirmed that there is no side-effect on tsc checks from using object, I'm ok with this change

Will you Approve so I can merge?

mhofman (approved):
> What are your concerns? From what I see, all the `Object` are `any` before and after.

I had missed that all usages of  `Object` also resulted in `any`, making it strictly equivalent to `object` in JSDoc.

> > As long as we've confirmed that there is no side-effect on tsc checks from using object, I'm ok with this change
> 
> Will you Approve so I can merge?

Given that `Object` and `object` are as bad as each other in all usages, approved.

PR #5270, opened 2022-05-02
test(swingset): improve tests on result promises that appear in args

### LINKED ISSUE #5189: a result promise being included in the args of the same message might cause problems

While pondering #5106, an unusual sequence occurred to me that is worth a unit test.

```js
const payload = [];
const p = E(target).foo(payload);
payload.push(p);
```

The `HandledPromise` handler (provided by liveslots) doesn't run on the same turn as the call to `E` or the Proxy that it returns. So the caller has a whole turn to mutate the arguments they provided before anything gets serialized. That's not a big deal by itself, but it does mean that they might mention the result promise they got back from E.

Liveslots' `queueMessage` assumes that the `returnedP` it gets has not been observed before. If it was also included in `args`, it will be serialized earlier than expected, and assigned a `vpid` just as if it was an exported promise. But that will be a different vpid than the one it allocates for the result, so when it populates `valToSlot`/`slotToVal` at the end of the function, it will be overwriting the other assignment.

One symptom might be that sending the result promise to someone else might not give them a promise that resolves at the right time. It might also interfere with pipelining. Also some promises might not round-trip properly. All these cases are worth exercising with a unit test.

## Description of the Design

I think the fix is to allocate the vpid and populate `valToSlot`/etc earlier, before we serialize the arguments. That way the serialization should re-use the vpid we allocated, instead of allocating a new one.

I don't know how this will/should interact with the `const p = makePipelinablePromise(resultVPID)`, though. I *think* that this line should be unchanged: we create a new (local) Promise that's distinct from the result promise / `returnedP` that userspace sees, as before. We pass `resultVPID` into `makePipelinablePromise` so it can prepare to handle `dispatch.notify` later (`notify` causes `p` to get resolved, we return `p` from the handler, so `returnedP` will get resolved when HandledPromise sees `p` resolve).

## Security Considerations

I *think* this could only cause grief to the code that decided to exercise it. It might enable a memory leak of some sort, but that grief is also self-directed. So I'm ok with deferring it to MN-2 or MN-3.

## Test Plan

Unit tests that look at the syscalls being made.

warner:
This doesn't change any behavior, but asserts that we're more tolerant
of messages whose arguments reference their own result
promise. Liveslots cannot generate these, and the kernel refuses to
accept them from non-pipelining vats, but remote systems might
construct such messages. So comms accept them, the kernel must not
kill comms when they appear, and they must be translatable and
deliverable to other vats. And liveslots must not choke when they
arrive.

The original concern of #5189 is a non-issue for liveslots-based vats,
because HandledPromise and E automatically harden the arguments before
the caller can mutate them. But this provides test coverage beyond
that point.

closes #5189

### DIFF
--- a/packages/SwingSet/test/test-comms.js
+++ b/packages/SwingSet/test/test-comms.js
@@ -9,6 +9,7 @@ import { debugState } from '../src/vats/comms/dispatch.js';
 import {
   capargs,
   makeMessage,
+  makeResolve,
   makeDropExports,
   makeRetireExports,
   makeRetireImports,
@@ -67,10 +68,15 @@ function mockSyscall() {
   const sends = [];
   const resolves = [];
   const gcs = [];
+  const subscribes = [];
   const fakestore = new Map();
   const syscall = harden({
-    send(targetSlot, method, args, _result) {
-      sends.push([targetSlot, method, args]);
+    send(targetSlot, method, args, result) {
+      if (result) {
+        sends.push([targetSlot, method, args, result]);
+      } else {
+        sends.push([targetSlot, method, args]);
+      }
       // return 'r-1';
     },
     resolve(resolutions) {
@@ -78,7 +84,9 @@ function mockSyscall() {
         resolves.push(resolution);
       }
     },
-    subscribe(_targetSlot) {},
+    subscribe(targetSlot) {
+      subscribes.push(targetSlot);
+    },
     vatstoreGet(key) {
       return fakestore.get(key);
     },
@@ -98,7 +106,7 @@ function mockSyscall() {
       gcs.push(['retireExports', vrefs]);
     },
   });
-  return { syscall, sends, resolves, gcs, fakestore };
+  return { syscall, sends, resolves, gcs, subscribes, fakestore };
 }
 
 /*
@@ -126,7 +134,7 @@ function encodeArgs(body) {
 test('transmit', t => {
   // look at machine A, on which some local vat is sending messages to a
   // remote 'bob' on machine B
-  const { syscall, sends } = mockSyscall();
+  const { syscall, sends, subscribes, resolves } = mockSyscall();
   const dispatch = buildCommsDispatch(syscall, 'fakestate', 'fakehelpers');
   dispatch(['startVat', capargs()]);
   const { state, clistKit } = debugState.get(dispatch);
@@ -140,7 +148,7 @@ test('transmit', t => {
   const transmitterID = 'o-1';
   const aliceKernel = 'o-10';
   const aliceLocal = provideLocalForKernel(aliceKernel);
-  const { remoteID } = state.addRemote('remote1', transmitterID);
+  const { remoteID, receiverID } = state.addRemote('remote1', transmitterID);
   const bobLocal = provideLocalForRemote(remoteID, 'ro-23');
   const bobKernel = provideKernelForLocal(bobLocal);
 
@@ -196,12 +204,29 @@ test('transmit', t => {
     'transmit',
     encodeArgs('4:0:deliver:ro+23:cat::ro-20:ro+23:ro-21;argsbytes'),
   ]);
+
+  // bob!ref(res): a message that references its own result promise
+  const vpid = 'p-1';
+  dispatch(makeMessage(bobKernel, 'ref', capargs('args', [vpid]), vpid));
+  t.deepEqual(sends.shift(), [
+    transmitterID,
+    'transmit',
+    encodeArgs('5:0:deliver:ro+23:ref:rp-40:rp-40;"args"'),
+  ]);
+  // comms is decider, but the promise was briefly in the "decided by
+  // kernel" state, so we should see a syscall.subscribe . In an ideal
+  // world we'd avoid this.
+  t.deepEqual(subscribes, [vpid]);
+  // when comms receives a notify, it should do syscall.resolve
+  const res1 = `1:0:resolve:fulfill:rp+40;"resolution"`;
+  dispatch(makeMessage(receiverID, 'receive', encodeArgs(res1)));
+  t.deepEqual(resolves, [[vpid, false, capargs('resolution')]]);
 });
 
 test('receive', t => {
   // look at machine B, which is receiving remote messages aimed at a local
   // vat's object 'bob'
-  const { syscall, sends, gcs } = mockSyscall();
+  const { syscall, sends, gcs, subscribes } = mockSyscall();
   const dispatch = buildCommsDispatch(syscall, 'fakestate', 'fakehelpers');
   dispatch(['startVat', capargs()]);
   const { state, clistKit } = debugState.get(dispatch);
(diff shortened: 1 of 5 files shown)


packages/SwingSet/test/test-comms.js
```diff
@@ -67,18 +68,25 @@ function mockSyscall() {
...
-    send(targetSlot, method, args, _result) {
-      sends.push([targetSlot, method, args]);
+    send(targetSlot, method, args, result) {
+      if (result) {
+        sends.push([targetSlot, method, args, result]);
+      } else {
+        sends.push([targetSlot, method, args]);
+      }
       // return 'r-1';
     },
```
mhofman, line 81: Out of curiosity, why not do 
```suggestion
    send(...sendArgs) {
      sends.push(sendArgs);
      // return 'r-1';
    },
```

packages/SwingSet/test/test-kernel.js
```diff
@@ -1072,6 +1072,89 @@ test('promise reject', async t => {
   t.deepEqual(kernel.dump().acceptanceQueue, []);
 });
 
+async function doResultInArgs(t, enablePipelining) {
+  // #5189
```
mhofman: Nit: any reason not to use full URLs for issues?

packages/SwingSet/test/test-kernel.js
```diff
@@ -1072,6 +1072,89 @@ test('promise reject', async t => {
...
+    });
+    return;
+  }
+  syscallA.send(bobForA, 'one', capargs(args, [resP]), resP);
+
+  const q1 = kernel.dump().acceptanceQueue[0];
+  const m1 = q1.msg;
+  const kpid = m1.result; // probably kp40
+  const p1 = kernel.dump().promises[0];
+  t.is(p1.id, kpid);
```
mhofman, line 1134: Should we ensure there is only a single entry in the promise table? Why not do a `deepEqual` of `kernel.dump().promises` ?

packages/SwingSet/test/test-kernel.js
```diff
@@ -1072,6 +1072,89 @@ test('promise reject', async t => {
...
+    t.throws(() => syscallA.send(bobForA, 'one', capargs(args, [resP]), resP), {
+      message: /syscall translation error: prepare to die/,
+    });
+    return;
+  }
+  syscallA.send(bobForA, 'one', capargs(args, [resP]), resP);
+
+  const q1 = kernel.dump().acceptanceQueue[0];
+  const m1 = q1.msg;
+  const kpid = m1.result; // probably kp40
```
mhofman, line 1131: More for my understanding of the test infra here, but what would be wrong with doing `kernel.addExport(vatA, resP)`? It seems to be what other tests do (at least the ones I've had to updated)

packages/SwingSet/test/test-kernel.js
```diff
@@ -1072,6 +1072,89 @@ test('promise reject', async t => {
...
+
+  await kernel.run();
+  const b1 = logB.shift();
+  t.is(b1[0], 'message');
+  t.is(b1[2].result, b1[2].args.slots[0]);
+
+  // vatB should now hold decision-making authority
+  const p2 = kernel.dump().promises[0];
+  t.is(p2.id, kpid);
+  t.is(p2.decider, vatB);
```
mhofman, line 1149: Similarly, why not do a `deepEqual` of `kernel.dump().promises`

packages/SwingSet/test/test-kernel.js
```diff
@@ -1072,6 +1072,89 @@ test('promise reject', async t => {
...
+}
+
+test('result promise in args (pipelining)', async t => {
+  return doResultInArgs(t, true);
+});
+
+test('result promise in args (non-pipelining)', async t => {
+  return doResultInArgs(t, false);
+});
```
mhofman:
```suggestion
test('result promise in args (pipelining)', doResultInArgs, true);

test('result promise in args (non-pipelining)', doResultInArgs, false);
```

packages/SwingSet/test/test-liveslots.js
```diff
@@ -1495,3 +1495,103 @@ test('unserializable promise rejection', async t => {
...
+  //   besides `thenHandler` did the necessary resolve/reject already.
+
+  function build(_vatPowers) {
+    return Far('root', {
+      one(p, target) {
+        // the promise we receive should have the same identity as our
+        // result promise
+        E(target).two(p);
+        // we should be able to pipeline messages to it
+        E(p).three();
```
mhofman, line 1556: Knowing the upcoming resolution it feels weird to send a message to the eventual resolution `"four"` ?

packages/SwingSet/test/test-liveslots.js
```diff
@@ -1495,3 +1495,103 @@ test('unserializable promise rejection', async t => {
...
+  const s2 = log.shift();
+  t.is(s2.type, 'send');
+  t.is(s2.targetSlot, target);
+  t.is(s2.method, 'two');
+  t.deepEqual(s2.args.slots, [resP]);
+  t.is(log.shift().type, 'subscribe'); // result of two()
+
+  const s3 = log.shift();
+  t.is(s3.type, 'send');
+  t.is(s3.targetSlot, resP); // huh, this will be queued in the kernel
```
mhofman, line 1588: I assume the `huh` is because the vat is the decider of the unresolved promise, right?

packages/SwingSet/test/test-liveslots.js
```diff
@@ -1495,3 +1495,103 @@ test('unserializable promise rejection', async t => {
...
+  //   that happens, `thenHandler` does both a syscall.resolve() and
+  //   checks/resolves `importedPromisesByPromiseID`.
+
+  // * So if we wind up with both a real "imported" Promise and a
+  //   result vpid, any messages our userspace sends to the Promise
+  //   will be queued into the kernel, even though we're going to be
+  //   calling syscall.resolve . If we resolve it to something local,
+  //   those messages will take a round-trip through the kernel. It
+  //   might be nice to forward `res` to the generated Promise, to
+  //   avoid that trip in some cases.
```
mhofman, line 1539: To clarify my understanding, if we have the mirror flow, where the vat receives as argument the result promise of a previously delivered send, does liveslot realize this and locally queues messages sent to the promise, only forwarding them to the kernel if we resolve to a remote object? I assume the day we add promise redirects, we'll need to forward on redirection as well?

mhofman (approved):
Thanks for adding these tests!

LGTM, mostly nits and clarification questions.

packages/SwingSet/test/test-comms.js
warner: That might work. I know we've had problems with `undefined` vs `null` in the past, so I've learned to be very explicit with the lengths of arrays that we send over the syscall wire, but this (test) case probably isn't vulerable to those problems.

packages/SwingSet/test/test-kernel.js
warner: probably a good idea, I do want to keep this concise and not distract from the test itself

packages/SwingSet/test/test-kernel.js
warner: I'll check that `.length` is 1, but I don't want to bother checking the other fields (`.refCount`, `.policy`, `.subscribers`, `.queue`, `.state`), which `deepEqual` would obligate me to do

packages/SwingSet/test/test-kernel.js
warner: The test is exercising the vatTranslator for the outbound `syscall.send`, so I want the c-list entry to be created by the translator, not manually.

packages/SwingSet/test/test-kernel.js
warner: same as above: too much work to check the fields that I don't care about

packages/SwingSet/test/test-kernel.js
warner: I keep forgetting about that AVA feature, thanks

packages/SwingSet/test/test-liveslots.js
warner: true, if I allowed `three()` to be delivered (i.e. if I pretended to be the kernel more accurately, and actually honored all those `syscall.sends`) then that'd cause a rejection (fulfillToData), but the "four" makes the order of checks more clear

packages/SwingSet/test/test-liveslots.js
warner: right, ideally the message would remain within the vat, instead of roundtripping through the kernel.. ties into the comment above

packages/SwingSet/test/test-liveslots.js
warner: It would be nice if liveslots could recognize that case and send fewer messages into the kernel.

What happens now is that the first message (`dispatch.deliver(method='one', result=vpid1)`) creates a Promise and registers it with slotToVal/valToSlot. The Promise actually comes from `HandledPromise.applyMethod()`, as the `returnedP` argument to the handler. The second mesage (`dispatch.deliver(method='two', args=[vpid1])`) deserializes the arguments, finds `vpid1` in `slotToVal`, and thus converts it into that earlier Promise. So if userspace in `two(argp)` does an eventual-send to the `argp` it received, that will be queued in the JS engine on `returnedP`. If `one` resolves it to a local object, those messages will be delivered locally, without kernel involvement.

If/when we implement redirects/forwards, and `one` forwards it to something imported from the kernel, then we'll need to arrange for a `syscall.send` to happen.

packages/SwingSet/test/test-comms.js
warner: Oh, so I applied that patch and now the test is failing https://github.com/Agoric/agoric-sdk/runs/6278370102?check_suite_focus=true#step:5:1390 . It looks like `sends` is being given a four-element Array (with an `undefined` at the end). So maybe `send()` is being called with an `undefined` at the end, but `...sendArgs` is paying attention to the strict arity of the call site, rather than the defined-ness of the values.

I'll undo that patch.

PR #5272, opened 2022-05-03
ci: only notify if integration job fails on master

mhofman:
Avoid pinging slack and emails if the deployment job fails in PRs, as the author should be directly notified through Github. In case the PR is merged regardless of the failed job (bypassed, or not required), the jobs should re-run on the merge commit in the "push" branches or tags, and will then trigger the email/slack notifications.

Add notifications to the integration-test job.


michaelfig (approved):
LGTM!

PR #5273, opened 2022-05-03
feat: add governance for RECORDS as a parameter type

### LINKED ISSUE #5216: Add helper to verify liquidation parameter upgrades
## Description

From review of #5211:
> In order for voters to be able to assure themselves that the proposed liquidation terms will perform as advertised, I think it's helpful for the subject contract (`liquidateIncrementally.js`) to export a helper function that validates that it would be able to parse them, and be unlikely to hiccup when restarting. Simple things like ensuring the terms parse, and the ratios have plausible known brands. Otherwise voters have to judge based on human eyes reading the textual form of the json.

## Security Considerations

Fast-fail of upgrades, esp. before they are committed, is a good thing.

Chris-Hibbert:
closes: #5216
see-also: #5239 

## Description

Add the ability for governance to manage passable records.

### Security Considerations

Allows more distinction in parameter types.

Nothing special

### Testing Considerations

Includes tests parallel to other parameter types.

### DIFF
--- a/packages/governance/src/constants.js
+++ b/packages/governance/src/constants.js
@@ -14,10 +14,9 @@ export const ParamTypes = /** @type {const} */ ({
   NAT: 'nat',
   RATIO: 'ratio',
   STRING: 'string',
-
-  // passable record
-  RECORD: 'record',
+  PASSABLE_RECORD: 'record',
   UNKNOWN: 'unknown',
 });
+
 harden(ParamTypes);
 /** @typedef {typeof ParamTypes[keyof typeof ParamTypes]} ParamType */
--- a/packages/governance/src/constants.js
+++ b/packages/governance/src/constants.js
@@ -14,6 +14,9 @@ export const ParamTypes = /** @type {const} */ ({
   NAT: 'nat',
   RATIO: 'ratio',
   STRING: 'string',
+
+  // passable record
+  RECORD: 'record',
   UNKNOWN: 'unknown',
 });
 harden(ParamTypes);
--- a/packages/governance/src/contractGovernance/paramManager.js
+++ b/packages/governance/src/contractGovernance/paramManager.js
@@ -1,6 +1,6 @@
 // @ts-check
 
-import { Far } from '@endo/marshal';
+import { Far, passStyleOf } from '@endo/marshal';
 import { AmountMath } from '@agoric/ertp';
 import { assertKeywordName } from '@agoric/zoe/src/cleanProposal.js';
 import { Nat } from '@agoric/nat';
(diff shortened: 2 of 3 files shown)


packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -296,6 +296,65 @@ test('Ratio', async t => {
...
+    A2: 'Your Mother Should Know',
+    B: 'I Am the Walrus',
+    C1: 'The Fool on the Hill',
+    C2: 'Flying',
+    D: 'Blue Jay Way',
+  });
+  const paramManager = makeParamManagerBuilder()
+    .addRecord('BestEP', epRecord)
+    .build();
+  // @ts-ignore intentional error
```
turadg:
```suggestion
  // @ts-expect-error
```

Chris-Hibbert: changing it made the error go away, which is the point of `expect-error`.

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -44,6 +44,7 @@ const assertElectorateMatches = (paramManager, governedParams) => {
  * @property {(name: string, value: Invitation) => Promise<ParamManagerBuilder>} addInvitation
  * @property {(name: string, value: bigint) => ParamManagerBuilder} addNat
  * @property {(name: string, value: Ratio) => ParamManagerBuilder} addRatio
+ * @property {(name: string, value: Object) => ParamManagerBuilder} addRecord
```
turadg: Nit 
```suggestion
 * @property {(name: string, value: object) => ParamManagerBuilder} addRecord
```

Though don't we have CopyRecord? I don't think it's well typed by at least it would be documentation.

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -296,6 +296,64 @@ test('Ratio', async t => {
...
+  const replacement = harden({
+    A1: "Wouldn't It Be Nice",
+    A2: "Don't Talk",
+    B1: 'God Only Knows',
+    B2: "I Know There's an Answer",
+  });
+  await paramManager.updateParams({ BestEP: replacement });
+  t.is(paramManager.getRecord('BestEP'), replacement);
+
+  const brokenRecord = {
```
turadg, line 322: 🔁

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -296,6 +296,64 @@ test('Ratio', async t => {
   );
 });
 
+test('Record', async t => {
+  const epRecord = harden({
```
turadg, line 300: Lol

packages/governance/src/constants.js
```diff
@@ -14,6 +14,9 @@ export const ParamTypes = /** @type {const} */ ({
   NAT: 'nat',
   RATIO: 'ratio',
   STRING: 'string',
+
+  // passable record
```
turadg: WDYT of disambiguating in the key?

turadg (approved):

packages/governance/src/contractGovernance/paramManager.js
Chris-Hibbert: got it from endo. Seems to fit.

packages/governance/src/constants.js
Chris-Hibbert: Probably a good idea. I resisted because of the length, but that's probably not an important consideration.

PR #5275, opened 2022-05-03
Store metadata for each virtual object kind

### LINKED ISSUE #5262: Revise storage scheme for DVOs to make durable objects more easily distinguishable

Currently virtual object kinds and durable virtual object kinds share a common identifier space.  This means that when code iterates through the storage during `stopVat`, it cannot easily distinguish the two, which is annoying because we want to retain the durable DVOs but purge the volatile non-durable VOs.  It would be better if the code didn't have to jump through a lot of hoops to sort this out, and the easiest thing is to given then separate identifier schemes so that one could iterate through the portion of the vatstore keyspace that only pertains to VOs and just throw out everything that's there.

## Description of the Design

Currently, we only store metadata in the vatstore for durable kinds.  A key of the form `vom.kind.${kindID}` holds a record `{ kindID, tag }`; this is used to repopulate the in-memory `kindDescriptors` table on restart after an upgrade.  The design proposed here is to replace the key with `vom.dkind.${kindID}` for durable kinds and add `vom.vkind.${kindID}` for non-durable virtual kinds.  The record contents will be the same as before.  Note that, strictly speaking, we don't actually need anything besides a placeholder for the record value for non-durable kinds, but this makes the design symmetrical and this information could be of potential use in diagnostics when scanning VOs for deletion following an upgrade.

## Security Considerations

Nothing new

## Test Plan

-- existing VO/DVO tests should continue to work
-- some test should exercise distinguishing VO data from DVO data

FUDCo:
Previously we only stored metadata for durable virtual object kinds, as this metadata is only strictly required for restart after an upgrade.  This PR changes things so we also store this metadata for non-durable kinds (in a separately prefixed portion of the store) so that the non-durable virtual objects can efficiently be scanned for deletion during either `stop-vat` or as part of an incremental cleanup that runs after upgrade.

Closes #5262

### DIFF
--- a/packages/SwingSet/src/liveslots/vatstore-usage.md
+++ b/packages/SwingSet/src/liveslots/vatstore-usage.md
@@ -94,6 +94,11 @@ Durable Kinds are defined just like virtual Kinds, but they use a different cons
 The KindHandle is a durable virtual object of a special internal Kind. This is the first Kind allocated, so it generally gets Kind ID "1", and the handles will get vrefs of `o+1/N`.
 
 
+# Kind Metadata
+
+For each virtual object kind that is defined we store a metadata record for purposes of scanning directly through the defined kinds when a vat is stopped or upgraded.  For durable kinds this record is stored in `vom.dkind.${kindID}`; for non-durable kinds it is stored in `vom.vkind.${kindID}`.  Currently this metadata takes the form of a JSON-serialized record `{ kindID, tag }`, where the `kindID` property is the kind ID (redundantly) and `tag` is the tag string as provided in the `defineKind` or `makeKindHandle` call.
+
+
 # Virtual/Durable Collections (aka Stores)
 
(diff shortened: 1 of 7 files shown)


warner (approved):
Nice! So lazy stale virtual data deletion can work by:
* in `startVat`, update one key, maybe named `currentVersionKindIDBoundary`, to the current value of `exportID` (to remember that all non-durable KindIDs less than this number are stale)
* in `bringOutYourDead`, find the first key with a `vom.vkind.` prefix
  * if none, we're done
* given that as $KindID, find the first key with a `vom.${KindID}/` prefix
  * if none, move to the next `vom.vkind.` item
  * if present, delete it, also examine `vom.rc.${baseref}` and `vom.es.${baseref}`, process refcounts for imports / durables, then delete the keys
* loop until N items have been processed, or we run out of items
* maybe remember the fact that we're done in RAM, to avoid the extra DB query on each BOYD

PR #5254, opened 2022-04-29
ci: integration tests before merge

mhofman:
refs: #5193 

## Description

Follow-up to https://github.com/Agoric/agoric-sdk/pull/5193 enabling the new integration trigger for the `deployment-test` and `integration` jobs.

### Security Considerations

None

### Documentation Considerations

Added a paragraph to `CONTRIBUTING.md`

### Testing Considerations

This flow has been in [successful trial](https://github.com/Agoric/agoric-sdk/actions/workflows/mergify-test.yml) over the last couple days:

PR #5276, opened 2022-05-03
ci: require integration tests only before merging

mhofman:
refs: #5254

## Description

Instead of adding integration tests to branch protection, add them as required for the mergify merge queue. This should allow the PR to be embarked on the merge train before these long tests complete, so that if a merge or rebase is necessary, we don't wait for nothing.

### Security Considerations

None, these tests will ultimately run

### Testing Considerations

We'll see how this goes and I'll fast follow in case of issues.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,7 +3,22 @@ queue_rules:
   - name: main
     conditions:
       - base=master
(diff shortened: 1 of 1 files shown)


michaelfig (approved):
LGTM!

PR #5279, opened 2022-05-04
chore: Better document use of StorageAPI

gibson042:
## Description

Consistently use and reference documentation regarding **StorageAPI** (typed as `KVStore`) and "enhanced storage API" (typed as `KVStorePlus`).

### Security Considerations

n/a

It would be nice to eliminate one concept from each pair, but I'm not taking that step in this PR.

### Testing Considerations

n/a

### DIFF
--- a/packages/SwingSet/src/kernel/state/storageWrapper.js
+++ b/packages/SwingSet/src/kernel/state/storageWrapper.js
@@ -1,8 +1,9 @@
 import { assert } from '@agoric/assert';
 import { insistStorageAPI } from '../../lib/storageAPI.js';
 
-// We wrap a provided object implementing the Storage API (has/get/set/delete/getKeys)
-// and add some convenience methods.
+// We wrap a provided object implementing StorageAPI methods { has, getKeys,
+// get, set, delete } (cf. packages/SwingSet/docs/state.md#transactions) and
+// add some convenience methods.
 
 // NOTE: There's a lot of suspenders-and-belt paranoia here because we have to
 // be vewy, vewy careful with host-realm objects.  This raises a question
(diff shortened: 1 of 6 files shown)


packages/SwingSet/src/lib/storageAPI.js
```diff
@@ -19,10 +19,9 @@ export function insistStorageAPI(kvStore) {
 }
 
 /**
- * Assert function to ensure that something expected to be an enhanced storage
- * object actually implements the enhanced storage API.  It should be a storage
- * object that additionally has the methods `enumeratePrefixedKeys`,
- * `getPrefixedValues`, and `deletePrefixedKeys`.
+ * Assert function to ensure that an object implements the enhanced storage API.
+ * (StorageAPI plus methods { enumeratePrefixedKeys, getPrefixedValues,
+ * deletePrefixedKeys }), also known as "KVStorePlus".
```
gibson042, line 24: Given that the enhancements all relate to number-suffixed keys, should "enhanced storage API" be renamed to something like "numericalKeyStorageAPI" or "StorageAPIPlusNumericalKeys"?

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -1,8 +1,9 @@
 import { assert } from '@agoric/assert';
 import { insistStorageAPI } from '../../lib/storageAPI.js';
 
-// We wrap a provided object implementing the Storage API (has/get/set/delete/getKeys)
-// and add some convenience methods.
+// We wrap a provided object implementing StorageAPI methods { has, getKeys,
```
warner, line 4: "host-provided object" would probably be the best way to describe it

packages/SwingSet/src/lib/storageAPI.js
warner: Eh, I think that'd put too much detail into the name, it's not like we have a second enhancement layer that needs to be distinguished from this one.

warner (approved):
looks good

PR #5280, opened 2022-05-04
fix: unbork mergify yaml

mhofman:
refs: #5276 

YAML whitespaces are hard...


mhofman:
The mergify warning is actually what this is fixing. The check is reading from `master` and not from this branch.

Chris-Hibbert (approved):
I'll take your word for it for an emergency fix. 

Would someone who understands this please give a retroactive review?

Issue #5176, opened 2022-04-21
Tendermint admission controls based on cosmic-swingset state


mhofman:
## What is the Problem Being Solved?

To support #5334 ~~and more specifically #5027~~, we need the ability to only allow some messages into the mempool and into blocks based on the system's state:
- reject messages from getting into the mempool/blocks if Swingset is overloaded.
- accept messages from ~~high priority~~ economy critical senders if ~~low-priority~~ non-economic message are currently rejected
- accept cosmos governance messages that update the admission policy itself

## Description of the Design

The JavaScript side of cosmic-swingset will store the admission policy and expose it and the current state of Swingset (#5350). This info will likely be returned as the result of the init and commit actions and should be cached at the cosmos level for subsequent usage.

Policy updates would come in the form of cosmos level governance messages and would be routed to cosmic-swingset in a dedicated action. The admission checks should probably hard code these type of governance transactions as always being allowed regardless of policy so that the chain cannot permanently DOS itself. See https://github.com/Agoric/agoric-sdk/issues/5335

The info returned by cosmic-swingset is guaranteed to be deterministic and can be used for consensus block checks. Of course it can also be used for admissibility checks into the mempool.

Some potential pointers:
[`CheckAdmissibility`](https://github.com/Agoric/agoric-sdk/blob/adfdf3638f56e8f37d34661f033f42843354db26/golang/cosmos/x/swingset/types/msgs.go#L31) may be capable of this, while having access to current state.

There may be some further information the [Cosmos docs for transaction lifecycle](https://github.com/cosmos/cosmos-sdk/blob/58597139fa0fb9e9be60deebee3df1663aa2cfaf/docs/basics/tx-lifecycle.md).

If tendermint doesn't provide context for the mempool admissibility checks, the admission policy state may need to be store globally...

## Security Considerations
This mechanism can in effect cause a pause of all swingset ingress, including economic governance. At this point it cannot pause actions that have already been delivered into swingset from progressing internally.

## Test Plan
TBD

mhofman:
#5281 will add a `controller.getQueuesLength()` which returns 2 numbers: an `activeQueues` and an `inactiveQueues`. The former represent queues that can be processed next without any further state change. The latter represents messages pending on some state change (in particular a promise resolution, or pipeline), which could happen after some other delivery is made. The sum of both is probably the number we want to consider as the "total queue depth" but if there is nothing in the activeQueues, messages need to be delivered into swingset to nudge it.

[...]

PR #3625, opened 2021-08-09
expose hash of kernel state changes as "activityhash"

warner:
The multiple members of a consensus machine are supposed to perform
identical exection of every crank. To detect any possible divergence as
quickly as possible, the kernel maintains the "crankhash": a
constantly-updated string which incorporates (by SHA256 hash) a copy of every
DB write and delete.

`controller.getCrankHash()` can be run after one or more cranks have
finished, and the resulting hex string can be e.g. stored in a host
application consensus state vector. If two members diverge in a way that
causes their swingset state to differ, they will have different crankhashes,
and the consensus state vectors will diverge. This should cause at least one
of them to fall out of consensus.

Some keys are excluded from consensus: currently just those involving vat
snapshots and the truncation (non-initial starting point) of the transcript.

refs #3442

PR #5281, opened 2022-05-04
feat(swingset): get queues length

### LINKED ISSUE #5238: add controller.getRunQueueSize()

To support #5176, in which cosmos will stop accepting messages into the mempool and into blocks once the kernel announces it is too full, we need to give the kernel a voice. @michaelfig says it'd be sufficient to let cosmic-swingset poll the kernel for it's queue depth.

## Description of the Design

Add `controller.getRunQueueSize()` (feel free to bikeshed the name). Returns a set of Nat (non-negative integer), one for active queues, one for inactive queues (such as promise queues). Only safe/consistent to call when the kernel is idle (i.e. outside of a `controller.run()`).

Caller is responsible for deciding how to interpret the result. 0 means empty, but the "full" threshold is up to the caller (and should be configurable). Once priority queues are introduced (#4318), this would need to be updated to expose different numbers for low and high priority queues, and the caller may pay attention to two thresholds.

## Security Considerations

None I can think of.

## Consensus Considerations

The kernel-reported value will grow in response to kernel device IO calls (e.g. adding messages to the Mailbox device will push work onto the kernel queue, increasing the number). Calling `controller.run()`, with some particular `runPolicy`, on a non-empty queue could either increase or decrease the number (but in the long run should decrease, assuming no runaway computation).

So this should be safe to use for consensus operations as long as all replicas use the same sequence of device writes and `controller.run()` calls, and use the same `runPolicy`, and refrain from calling it concurrently with a `controller.run()` (which might sample the queue at random points during execution, which would give inconsistent results).

~~However we don't currently plan to use this for consensus operations: accepting messages into the mempool is done separately by each validator and is outside of consensus.~~

We plan to use this mechanism for both consensus (validating a proposed block) and non consensus (accepting into the mempool)

## Test Plan

Unit tests, probably in `test-kernel.js`.

mhofman:
closes: #5238
refs: #5176

## Description

This PR tracks the number of messages pending on promise queues as a new kernel stat, and exposes the run-queue + acceptance queue length, as well as the promise queue length in a new controller method, which can be used by cosmic-swingset to decide if messages should be delivered to mailbox or possibly bridge devices.

To make sure the stats are consistent with the rest of the kernel state, this PR also:
- Split the kernel stats into consensus and local sections, saved separately (partially undoes https://github.com/Agoric/agoric-sdk/pull/3625/commits/df8359eca80e28736d294a558ed6c5e3b8b14127)
- Commits / rolls-back the `kernelStats` along kernel state, but avoiding intermediary DB writes

### Security Considerations

The `kernelStats` were previously out of consensus and not thoroughly vetted to be deterministic, or consistent with the kernel state. I believe I have audited all changes to the stats, especially the ones that are non in consensus, and made sure any crank abort or reload would reflect in the stats kept in RAM.

Description added, stats don't seem to be documented anywhere else than in `metrics.js`.

I have now fully typed the metrics allowing calls like `incStats` to type check that the stat name is valid.

### Testing Considerations

I added unit tests for the stats logic, however nothing tests the content of stats against expectations regarding the rest of the kernel state.

One example are the up/down stats. The base value should always be a pure function of the known state of the kernel, but nothing enforces or verifies that. I've manually reviewed all paths where existing run-queue and acceptance queue, as well as the new promise queue, are modified to ensure stats are tracked appropriately

Putting some of these stats under consensus should quickly highlight non deterministic stats (if any) by the deployment integration test. Swingstore traces in integration should also highlight divergences in non-consensus stats however such consistency is not yet enforced in the integration test due to snapshot hash divergences.


warner:
API seems good.. having it depend upon `stats` feels sligtly weird, but maybe that's just because of the not-intended-for-consensus property that you pointed out. OTOH it'd be a shame to duplicate code to get this data.

I'm not 100% sure the stats are managed accurately across crank unwinding and/or replay, I seem to remember them being held in RAM for some period of time, and then getting written to the DB only at the end of a crank, and I remember being nervous about that but decided that since they weren't being used for consensus, it wouldn't matter.

So maybe it *would* be a good idea to track them separately. The run-queue and acceptance-queue are still single keys, we could just JSON.parse and `.length` them when requested (which won't be very often, assuming cosmic-swingset only asks once per block, but OT3H maybe that's a bad assumption to make).

Dunno.

mhofman:
I did double check the write logic, and it seemed sane. I will double check the abort crank logic however, as now I'm not sure if we reload the saved object at that point.

That said, I think it should be safe to use for our purpose:
- Accepting messages into the mem-pool is non-consensus in the first place
- Accepting messages into a block is done by the block proposer, which I think means it's also out of consensus in the sense other nodes would accept the proposer's decision?

I'm happy to move a subset of stats explicitly into consensus. Not sure the best approach for that is. I already have reservations about all the transient serializations and deserializations happening during a crank, and moving stats under that mechanism doesn't feel quite right.

mhofman:
I did find issues related to crank abort, and when stats are loaded or first initialized. I have now aligned them with other store entries, and moved them in consensus, including participating in the crankhash. They are however still only written down at the end of the crank to avoid DB churn.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1645,6 +1643,27 @@ export default function buildKernel(
       return kernelKeeper.getActivityhash();
     },
 
+    // While stats are local and technically not part of concensus state, these
```
warner: typo s/concensus/consensus/

also let's say ".. and not enforced as part of consensus state, we would expect all replicas to have the same value", to make it clear that this equality is an expectation, not a demand

packages/SwingSet/test/test-state.js
```diff
@@ -701,6 +703,22 @@ test('meters', async t => {
   t.deepEqual(k.getMeter(m3), { remaining: 'unlimited', threshold: 5n });
 });
 
+const makeTestCrankHasher = k => {
```
warner: This is ok, but a big purpose of this test is to ensure we really understand what's going into the activityhash, which is why it was doing `h.update()` with literal or quasiliteral strings: minimal indirection, maximum visibility of the hash inputs.

Having kernelStats be part of the consensus state makes that approach pretty annoying, so I completely get why this layer of indirection is useful, and it's fine, just keep in mind the reason for the original structure.

warner:
Yeah, I'm dubious about enforcing consensus of kernelStats. It would prevent us from adding new stats in minor kernel updates (those updates would have to be deployed at exactly the same time across all replicas), as well as stats on things that really are replica-local (like the number of times the kernel has been rebooted, or heap snapshot activity, or some sort of cache instrumentation).

Until now (where we want to use some stats for backpressure), these stats have been entirely for telemetry/debugging purposes. Our current plan for backpressure is to use it in a non-consensus fashion (mempool inlet control).

OTOH (and @michaelfig can probably clarify), it might be easy to accidentally (or intentionally) implement the cosmic-swingset -side backpressure queue-depth check in a way that *is* part of consensus (at the beginning of DeliverTx, causing a txn error if the block proposer allows a swingset message in despite the queues being too full).

Part of me wants two separate kernelStats, one local (with most of the stats), and the other in consensus (only containing the queue depths). Then we don't need to worry about cosmic-swingset accidentally relying upon something that swingset doesn't guarantee consensus about. But having two sets of stats feels like bloat, especially if they both need `initializeStats` and updates after each crank.

Maybe there's a deeper problem with thinking of SwingStore as an acceptable place to hold non-consensus data at all, and we should provide the kernel with a secondary non-consensus storage tool instead of using `local.*` to distinguish them. If we ever wanted to try (again) to store swingset data in the cosmos-sdk IAVL tree, the `local.` trick wouldn't work, and we'd definitely need a secondary DB. But that sounds like far more work than we can do for MN-1.

Hrmph. So the short-term options are:
* 1: include kernelStats in consensus
  * risks a MN-1 consensus failure if one of our stats is less stable than we thought (I'm looking suspiciously at `kernelDevices`, probably ok but hmm)
  * adding new non-consensus stats would require more work in the future (but maybe YAGNI)
* 2: split into `kernelStats` and `local.kernelStats`
  * somewhat less risk
  * more work now, kinda annoying bloat

@FUDCo any opinions here?

mhofman:
> Part of me wants two separate kernelStats, one local (with most of the stats), and the other in consensus (only containing the queue depths). Then we don't need to worry about cosmic-swingset accidentally relying upon something that swingset doesn't guarantee consensus about. But having two sets of stats feels like bloat, especially if they both need `initializeStats` and updates after each crank.

I think 2 stats is right when we need them, and when I audited the current stats, they all seemed to be based on consensus state already, so I didn't see the harm in moving them all into consensus. I'm not sure 2 different keys, one prefixed by `local.` is that much bloat.

> * I'm looking suspiciously at `kernelDevices`

If we believe this may be different between validators, I'm happy to review, but the devices are already stored under the `device.` and `d$NN` keys in the store, so they're already part of consensus.

FUDCo:
> @FUDCo any opinions here?

I'd also go with option 2.  I think it's proper to distinguish between stats that are kept for performance monitoring, tuning, and debugging vs. stats that are part of the operational logic of the chain.

mhofman:
@warner @FUDCo PTAL

packages/SwingSet/src/controller/controller.js
```diff
@@ -368,7 +368,7 @@ export async function makeSwingsetController(
         `alleged bundleID ${allegedBundleID} does not match actual ${bundleID}`,
       );
     }
-    kernel.installBundle(bundleID, bundle);
+    await kernel.installBundle(bundleID, bundle);
```
warner, line 371: good catch, I think the old version was sync

packages/SwingSet/test/test-state.js
```diff
@@ -701,6 +707,27 @@ test('meters', async t => {
...
+const { consensusStats: testCrankHasherConsensusStats } =
+  testCrankHasherStats.getSerializedStats();
+
+const makeTestCrankHasher = () => {
+  const h = createHash('sha256');
+  const update = (...args) => {
+    h.update(args.map(val => `${val}\n`).join(''));
+  };
+  const digest = () => {
+    update('add', 'kernelStats', testCrankHasherConsensusStats);
```
warner: yeah ok, this is a decent compromise, but maybe add a note to remind ourselves that this is hard-wired to include only the initial value of the kernel stats, so any new test that calls a `kernelKeeper` function that increments a state will get the wrong crankhash

warner (approved):
Looks good, minor comment suggestion

packages/SwingSet/src/kernel/kernel.js
mhofman: That comment has been removed already.

packages/SwingSet/test/test-state.js
mhofman: I'd be happy to revert the string serialization part.

mhofman:
@warner, I addressed your feedback in e16a9a3972f419ad7b9a43a10d278f6b4b60aeb4 then squashed the PR in a way that hid all the back and forth in this PR for a clean merge, while preserving individual commits that make sense: [`e16a9a3..e10564d`](https://github.com/Agoric/agoric-sdk/compare/e16a9a3972f419ad7b9a43a10d278f6b4b60aeb4..e10564d2e06a571e6d4f2bfd9a4c1ae9f8e622ee)

packages/SwingSet/test/test-state.js
mhofman: Reverted

PR #5282, opened 2022-05-04
fix: provide provide, a get-or-make for mapStores

erights:
Call `provide(mapStore, key, makeValue)` to get or make the value associated with the key.
If there already is one, return that. Otherwise,
call `makeValue(key)`, remember it as the value for
that key, and return it.

This is related to the upsert or emplace of https://github.com/tc39/proposal-upsert . 

As shown at https://gist.github.com/warner/95594e60b194673420bd515ab8d3662c#file-util-js , durability patterns make heavy use of such a `provide` notion. 

Note to reviewers: perhaps we should instead make this an instance method of all mapStores. But this involves changing all the mapStore implementations and types, which I wish to avoid for the moment. Please comment on this, but separately from other comments.

### DIFF
--- a/packages/store/src/index.js
+++ b/packages/store/src/index.js
@@ -75,6 +75,8 @@ export {
   makeScalarMapStore as makeStore, // Deprecated legacy
 } from './stores/scalarMapStore.js';
 
+export { provide } from './stores/store-utils.js';
+
 // /////////////////////// Deprecated Legacy ///////////////////////////////////
 
 // export default as well as makeLegacy* only for compatibility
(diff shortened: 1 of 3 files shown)


FUDCo:
I think providing this as a library function rather than as a store method makes for better composability.

FUDCo (approved):
Upsert will always appear.  This looks great.

turadg:
> I think providing this as a library function rather than as a store method

Another option is to have both. Not necessarily recommending that but I do think having it on the stores is valuable and that's compatible with having it as a library function. The following is a reason given for having it as a library function:

> …makes for better composability.

What are some examples of how it would be used composed?

erights:
Two of these tests have been stuck in CI since last night, preventing merging. What do we do to unstick these?

PR #5283, opened 2022-05-04
feat(ertp): durability

erights:
Make ERTP Durable


packages/ERTP/src/brand.js
```diff
@@ -1,22 +1,32 @@
...
-  /** @type {Brand} */
-  const brand = Far(`${allegedName} brand`, {
-    isMyIssuer: allegedIssuerP => E.when(allegedIssuerP, isMyIssuerNow),
+export const makeDurableBrand = (
+  issuerBaggage,
+  allegedName,
+  isMyIssuerNow,
+  displayInfo,
+) => {
+  const brandKindHandle = provide(issuerBaggage, 'brandKindHandle', () =>
```
turadg: important or incidental that each brand have its own kind handle?

packages/ERTP/src/brand.js
```diff
@@ -1,22 +1,32 @@
...
  * @param {string} allegedName
  * @param {(allegedIssuer: Issuer) => boolean} isMyIssuerNow
  * @param {DisplayInfo} displayInfo
  * @returns {Brand}
  */
-export const makeBrand = (allegedName, isMyIssuerNow, displayInfo) => {
-  /** @type {Brand} */
-  const brand = Far(`${allegedName} brand`, {
-    isMyIssuer: allegedIssuerP => E.when(allegedIssuerP, isMyIssuerNow),
+export const makeDurableBrand = (
```
turadg: is this name interim for refactoring or meant for master? generally, do you advise including `Durable` in the name or a durable kind maker even when there is no non-durable kind to disambiguate?

reading further I see the `makeBrand` is used within this method, but that conflict isn't the concern of the API. The `makeDurableBrand` name seems more appropriate to me on line 22 as the result of `defineDurableKind`.

packages/ERTP/src/issuerKit.js
```diff
@@ -85,9 +94,52 @@ const makeIssuerKit = (
...
+ * before anything else is corrupted by that corrupted state.
+ * See https://github.com/Agoric/agoric-sdk/issues/3434
+ * @returns {{
+ *  mint: Mint<K>,
+ *  issuer: Issuer<K>,
+ *  brand: Brand<K>,
+ *  displayInfo: DisplayInfo,
+ * }}
+ */
+export const makeIssuerKit = (
```
turadg, line 142: why retain the non-durable version?

packages/store/src/stores/store-utils.js
```diff
@@ -86,3 +86,23 @@ export const makeCurrentKeysKit = (
...
+ * call `makeValue(key)`, remember it as the value for
+ * that key, and return it.
+ *
+ * @template K,V
+ * @param {MapStore<K,V>} mapStore
+ * @param {K} key
+ * @param {(key: K) => V} makeValue
+ * @returns {V}
+ */
+export const provide = (mapStore, key, makeValue) => {
```
turadg: @dckc this would obviate https://github.com/Agoric/agoric-sdk/blob/7dab31c6e19bb4e1345ce8e6fd8a8d5e133b685f/packages/run-protocol/src/runStake/attestation.js#L24-L36

turadg: Out of scope for this PR but I'd like to understand why we don't include this on the store next to `get` and `init`.

packages/ERTP/src/paymentLedger.js
```diff
@@ -400,39 +416,44 @@ export const makePaymentLedger = (
...
-    getDisplayInfo: () => displayInfo,
-    makeEmptyPurse,
-  });
-
-  /** @type {Mint<K>} */
-  const mint = Far(`${allegedName} mint`, {
-    getIssuer: () => issuer,
-    mintPayment,
-  });
+  const dropContext =
```
turadg: why drop the context? doesn't `defineDurableKind`'s `behavior` param require functions with `context` param just as `defineKind` does?

packages/store/src/stores/store-utils.js
dckc: looks like we're headed in that direction: https://github.com/Agoric/agoric-sdk/pull/5282

packages/ERTP/src/issuerKit.js
Chris-Hibbert: This calls `makeDurableIssuerKit()` on line 138, so it's not the original non-durable function.

I'm guessing here, but to check my understanding, I see that the two functions have different parameter lists. I presume `makeIssuerKit()` would be called by ordinary clients as they have been doing. `makeDurableIssuerKit()` must be intended to be called by the durability layer to revive objects when it already has a handle for them.

packages/ERTP/src/brand.js
erights: Which way to go is important, and is central to the nature of this experiment. Here, I have all the elements of a new issuer that are per-issuer --- issuer, mint, and brand --- be singletons with their own kind per instance. As a result, they use state closed over by their behavior where we would normally use durable instance state. Curious to explore the tradeoffs.

packages/ERTP/src/brand.js
erights: Interim. I changed names as I changed APIs, to keep track of where I was in refactoring. Now that I'm on the other side, all the unnecessarily complex internal names can be simplified again.

Will look again at makeBrand vs makeDurableBrand. Thanks.

packages/ERTP/src/issuerKit.js
erights: Tons of existing callers of the existing APIs that I am unprepared to break or fix.

packages/store/src/stores/store-utils.js
erights: Ah. I didn't know about `getOrElse`. Yes, we don't need both. Seem identical but for 
   * naming
   * providing the key as argument to `make`.

Perhaps it should be an instance method. Asked for feedback on this in the PR comment above. @FUDCo as you saw prefers the static function. I am undecided.

packages/ERTP/src/paymentLedger.js
erights: I only use `dropContext` for those methods that do not need the context. 

In the singleton pattern as supported by `provideDurableSingleton*`, we use lexical state rather than a state record. Although we do not yet, we could consistently use lexical state instead of the `self` or `facets` portion of the context as well, following a suggestion of @dtribble 's, we could. Thus, it is even possible we should fold `dropContext` into `provideDurableKind*` so that it is implicit and applies to all behavior methods. I'll leave this conversation unresolved until we decide to either do that or not do that.

packages/ERTP/src/brand.js
FUDCo: If brands are distinct and brands per se all have the same API (which I presume they sort of have to), I see little, if any, benefit to each having its own kind handle, while there is a non-negligible cost (storage space, of course, but more significantly the added complexity of the bookkeeping needed to keep track).

FUDCo: Actually, I was not understanding the closed-over state part of the story.  I think I tentatively withdraw that last comment.

packages/ERTP/src/issuerKit.js
turadg: >  so it's not the original non-durable function

Yeah, I meant retaining the _availability_ of a non-durable version.

> unprepared to break or fix

what would break if there was one `makeIssuerKit()` which returned a durable one?

packages/store/src/stores/store-utils.js
dckc: `getOrElse` was a one-off in a specific contract. #5282 puts it in the store package, which is where it belongs. In due course, `attestation.js` should be refactored.

packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -63,7 +63,7 @@ test('issuer.getAssetKind', t => {
   t.is(issuer.getAssetKind(), AssetKind.NAT);
 });
 
-test('issuer.makeEmptyPurse', async t => {
+test.skip('issuer.makeEmptyPurse', async t => {
```
Chris-Hibbert: These are breaking because we're using fakeDurableNotifiers that don't notify.

erights:
Not green only because of a genuine bug in the dapp-otc repository, which should be fixed by https://github.com/Agoric/dapp-otc/pull/38

Since we are otherwise green, testing both with and without `relaxDurabilityRules`, and actually testing a vat-upgrade (ertpService), I think we're now ***Ready For Review***.

packages/ERTP/src/issuerKit.js
erights: There are 350 calls to `makeIssuerKit` in agoric-sdk, spread across 107 files. Most of those are tests that don't care about durability, so we don't want to require that those be refactored to pass in a baggage argument.

When we do want to make an *upgradable* issuerKit, we need to call a maker to which we do pass a baggage argument. But either way, there is only one underlying implementation of the issuerKit, and that one implementation makes a durable issuer kit.

Because `makeIssuerKit` is not given a baggage, the issuerKit it gives you back is durable but not upgradable.

erights:
But see https://github.com/Agoric/agoric-sdk/issues/5628

Chris-Hibbert (commented):
Submitting this because there's an outstanding comment that I can only see by doing so. I don't know how relevant it still is.

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -0,0 +1,107 @@
...
+let VatDataGlobal;
+if ('VatData' in globalThis) {
+  assert(globalThis.VatData, 'VatData defined in global as null or undefined');
+  VatDataGlobal = globalThis.VatData;
+} else {
+  // XXX this module has been known to get imported (transitively) in cases that
+  // never use it so we make a version that will satisfy module resolution but
+  // fail at runtime.
+  const unvailable = () => assert.fail('VatData unavailable');
+  VatDataGlobal = {
```
Chris-Hibbert, line 31: I don't get what's going on here.   Why do these names have to appear in this list?  Why are there no others?

turadg: This ensures that if `global.VatData` isn't defined, the keys still resolve to functions that can be called. They'll just throw an exception when that happens.

packages/ERTP/src/brand.js
erights: At this point, I am satisfied with the singleton pattern for such low cardinality things, especially those naturally coded with lots of lexical state such as an issuerKit. Thus I'm resolving this conversation. If you think this should remain controversial, please feel free to unresolve and continue the conversation.

packages/ERTP/src/brand.js
erights: Now that this PR has otherwise settled, we still need to revisit these renamings, so leaving this conversation open until then.

packages/store/src/stores/store-utils.js
erights: `provide` is still static, at least for now. But I just replaced the one cal to `getOrMake` with `provide`, which I think is adequate to resolve this conversation. Resolving.

packages/vat-data/src/vat-data-bindings.js
erights: @Chris-Hibbert , if @turadg 's explanation is an adequate answer, please resolve this conversation. Thanks.

erights: Note that the code in question is not new in this PR. It seems so only because it was moved from the corresponding `index.js`

erights:
See also https://github.com/Agoric/agoric-sdk/pull/5674

I will probably merge that one into this before proceeding with this. I'll PTAL at that time.

packages/ERTP/src/paymentLedger.js
erights: Merging https://github.com/Agoric/agoric-sdk/pull/5674 into this PR will resolve this conversation.

packages/ERTP/src/issuerKit.js
```diff
@@ -24,32 +24,29 @@ import './types.js';
  *
  *  `displayInfo` gives information to the UI on how to display the amount.
  *
- * @param {string} allegedName
- * @param {K} [assetKind=AssetKind.NAT]
- * @param {AdditionalDisplayInfo} [displayInfo={}]
+ * @param {MapStore<string,any>} issuerBaggage
```
turadg: if we can't make an explicit type, it's safer to use `unknown` so the uses of it don't think they're type checking when they're not
```suggestion
 * @param {MapStore<string, unknown>} issuerBaggage
```

given that this MapStore must have particular keys and non-homogenous values, I'd like to consider having a type such as `Baggage<T>` that encodes the mapping. It could still be accessed by `baggage.get(key)` instead of `baggage[key]` so no runtime change needed. That's out of scope of this PR but I'm happy to supply it now or later.

Though as I read down, I think it may be better to change this signature to take an object. Will describe closer.

packages/ERTP/src/issuerKit.js
```diff
@@ -24,32 +24,29 @@ import './types.js';
...
  *  brand: Brand<K>,
  *  displayInfo: DisplayInfo,
  * }}
  */
-const makeIssuerKit = (
-  allegedName,
-  // @ts-expect-error K could be instantiated with a different subtype of AssetKind
-  assetKind = AssetKind.NAT,
-  displayInfo = harden({}),
+export const provideDurableIssuerKit = (
```
turadg: IIUC we need this function that takes a MapStore because we need the thing that imbues behavior to be agnostic to whether the data is coming from storage or being constructed anew. It's a new and critical pattern I think would be worth documenting in a Markdown file.

Regarding naming, this "provide" is overloaded and may be confused with other uses. Since the pattern is about taking data that may be old and making it live again with behavior, some alternatives:
- rehydrate (common in React, but also means make interactive)
- revive
- restore

packages/ERTP/test/swingsetTests/ertpService/vat-ertp-service.js
```diff
@@ -0,0 +1,64 @@
...
+      issuerBaggageSet.add(issuerBaggage);
+
+      return issuerKit;
+    },
+  });
+
+  if (didStart) {
+    console.log(`DID START`);
+    for (const issuerBaggage of issuerBaggageSet.values()) {
+      provideDurableIssuerKit(issuerBaggage);
```
turadg: it looks like `provideDurableIssuerKit` takes a MapStore instead of an object to keep this loop simple. 

An alternative is to make it take a simple object and in this loop read what's coming out of `issuerBaggageSet` to create an object to pass to `provideDurableIssuerKit`. An advantage of this approach is that there can be data transformations for each upgrade. `provideDurableIssuerKit` shouldn't have to be concerned with compatibility with multiple versions of the data schema, right?

packages/ERTP/src/issuerKit.js
```diff
@@ -24,32 +24,29 @@ import './types.js';
...
+  issuerBaggage,
   optShutdownWithFailure = undefined,
-  { elementSchema = undefined } = {},
 ) => {
+  const allegedName = issuerBaggage.get('allegedName');
+  const assetKind = issuerBaggage.get('assetKind');
+  const displayInfo = issuerBaggage.get('displayInfo');
+  const elementSchema = issuerBaggage.get('elementSchema');
```
turadg: If we make this a single call to create an object we can get autocomplete of the names:
```suggestion
  const { allegedName, assetKind, displayInfo, elementSchema }  = issuerBaggage.getAll('allegedName', 'assetKind', 'displayInfo', 'elementSchema');
```

packages/ERTP/src/issuerKit.js
```diff
@@ -99,9 +77,104 @@ const makeIssuerKit = (
...
+ * its value. The allegedName is useful for debugging and double-checking
+ * assumptions, but should not be trusted.
+ *
+ * The assetKind will be used to import a specific mathHelpers
+ * from the mathHelpers library. For example, natMathHelpers, the
+ * default, is used for basic fungible tokens.
+ *
+ *  `displayInfo` gives information to the UI on how to display the amount.
+ *
+ * @param {MapStore<string,any>} issuerBaggage
```
turadg:
```suggestion
 * @param {MapStore<string, unknown>} issuerBaggage
```

packages/ERTP/src/issuerKit.js
```diff
@@ -99,9 +77,104 @@ const makeIssuerKit = (
...
+ * larger unit of computation, like the enclosing vat, can be shutdown
+ * before anything else is corrupted by that corrupted state.
+ * See https://github.com/Agoric/agoric-sdk/issues/3434
+ * @param {Partial<{elementSchema: Pattern}>} [options]
+ * @returns {{
+ *  mint: Mint<K>,
+ *  issuer: Issuer<K>,
+ *  brand: Brand<K>,
+ *  displayInfo: DisplayInfo,
+ * }}
```
turadg: This is repeated 3 times. I'd suggest using the `IssuerKit` type. 
This line will to be updated by the return type of `provideDurableIssuerKit` as that's now what produces the actual object:

 https://github.com/Agoric/agoric-sdk/blob/7a9a7a301d67cc681ed4753114f0aba8e9339581/packages/ERTP/src/issuerKit.js#L107

packages/vat-data/src/kind-utils.js
```diff
@@ -0,0 +1,59 @@
...
+  return /** @type {Record<K, U>} */ (harden(Object.fromEntries(mapEnts)));
+};
+
+/**
+ * @template T
+ * @param {import('@agoric/store').MapStore<string, unknown>} baggage
+ * @param {string} kindName
+ * @param {T} methods
+ * @param {import('./types.js').DefineKindOptions<unknown>} [options]
+ * @returns {T}
```
turadg: should this return what `Far` returns?

packages/vat-data/src/kind-utils.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+export const ProvideFar = (baggage, kindName, methods, options = undefined) => {
```
turadg: ditto on "Provide" term feedback

packages/vat-data/src/types.d.ts
```diff
@@ -39,21 +40,25 @@ declare class DurableKindHandleClass {
 }
 export type DurableKindHandle = DurableKindHandleClass;
 
+type DefineKindOptions<C> = {
```
turadg, line 43: 👍

packages/zoe/src/makeHandle.js
```diff
@@ -1,8 +1,29 @@
 // @ts-check
 
 import { assert } from '@agoric/assert';
+import { provide } from '@agoric/store';
+import { defineDurableKind, makeKindHandle } from '@agoric/vat-data';
 import { Far } from '@endo/marshal';
 
+/**
+ * @template {string} H
+ * // ??? Purposeful use of `any`. Should we use `unknown` instead?
```
turadg: yes, because the types are not known.

packages/zoe/src/makeHandle.js
```diff
@@ -1,8 +1,29 @@
...
+ */
+export const defineDurableHandle = (baggage, handleType) => {
+  assert.typeof(handleType, 'string', 'handleType must be a string');
+  const durableHandleKindHandle = provide(
+    baggage,
+    `${handleType}KindHandle`,
+    () => makeKindHandle(`${handleType}Handle`),
+  );
+  const makeHandle = defineDurableKind(durableHandleKindHandle, () => ({}), {});
+  return /** @type {() => Handle<H>} */ (makeHandle);
```
turadg, line 24: this doesn't work?
```suggestion
  return makeHandle;
```

packages/ERTP/src/issuerKit.js
Chris-Hibbert: All of the Baggages are currently `MapStore<string,any>`.  having a `Baggage` type would probably be better, but the point is that as far as baggage is concerned, keys are strings, and the value types are only known at the point of access. The Baggages are passed through many intermediaries, who don't and can't know what's being stored at higher and lower levels.

packages/ERTP/src/issuerKit.js
Chris-Hibbert: The Provide pattern we're using here is to supply a constructor for the first time the `provide` is called, and all other times the value will be pulled from the store. The caller wants the code to be agnostic about whether this is the first time or later.  So 'revive', 'restore', 'rehydrate' ignore this aspect of "first time or later".  I think a standard name would mean something more like "build or retrieve", and I don't have a good candidate for that.

As an alternative to "create", "make", or "build", we've agreed that "provide" means something like "find it if it already exists, or build it if you can't find it", which matches this usage reasonably closely. The prior uses of "provide" presumed that the callee knew how to build, and now we're providing the thunk that will do the building.

packages/ERTP/src/issuerKit.js
Chris-Hibbert: This doesn't seem like a big win to me. I typed the names once each and pasted them into the string rather than typing them again. Nothing outside this code should know that these names are being defined or retrieved.

packages/ERTP/src/issuerKit.js
Chris-Hibbert: Good idea.

packages/vat-data/src/kind-utils.js
Chris-Hibbert: Yes, I think so. Where is that declaration?

packages/ERTP/src/issuerKit.js
turadg: Ah, right. It's not any particular baggage until it's what came from storage is validated to match a pattern. Would it make sense to validate the baggage data before passing it to the likes of `provideDurableIssuerKit`? If so I would propose `type Baggage<T=Record<string, unknown>> ` that we use for now and then when we have some data validation we can narrow to a type that has specific keys and value types.

packages/ERTP/src/issuerKit.js
turadg: > we've agreed that "provide" means something like "find it if it already exists, or build it if you can't find it",

Agree completely with this.

> which matches this usage reasonably closely

This part doesn't make sense to me. The _use_ of `provideDurableIssuerKit` matches that pattern in that one use is building it and one is finding it, but the function itself is taking a store and making a behavior.

> So 'revive', 'restore', 'rehydrate' ignore this aspect of "first time or later".

Agreed. But the same goes for `provideDurableIssuerKit`. The pattern I see is that 10 there's a maker and a store reviver —the first time or later, yes, but: the maker operates by constructing a store to send into the reviver.

I'll bring up in Zoe meeting today.

packages/vat-data/src/kind-utils.js
turadg: https://github.com/endojs/endo/blob/c4f23d754d92883b5f470bc74a0b029e152e0e08/packages/marshal/src/make-far.js#L137-L149

packages/ERTP/src/issuerKit.js
turadg: From our chat, cases in use of `provide`:

1. as in `provide()` in vat-data that takes a baggage, a name and a thunk for making it if it's not there. (`getOrMake`)
2. as in `provideDurableSetStore` that takes the name but already knows how to make the thing and the caller isn't concerned with whether it's the first or nth call
3. as in `provideErtpService` that doesn't take any name and is just responible for adding behavior to data from baggage (not necessariliy re-viving, but "viving". Vivify?)
4. as in `ProvideFar` which is give me something that has this behavior I'm passing in, and if there already was a thing then use this new behavior in all existing references. In short: vivify singleton objects.

Some options for (3) and (4):
- [actuate](https://www.merriam-webster.com/dictionary/actuate) 
- [animate](https://www.merriam-webster.com/dictionary/animate)
- init
- setup
- [vitalize](https://www.merriam-webster.com/dictionary/vitalize)
- [vivify](https://www.merriam-webster.com/dictionary/vivify)

packages/ERTP/src/paymentLedger.js
erights: Having merged #5674 , I'm resolving this conversation. As always, please unresolve it if there's more to say or I missed something.

packages/ERTP/src/issuerKit.js
erights: For now, I will proceed with `provide` for (1) and (2), and `vivify` for (3) and (4). But leaving this conversation open until I've done so.

packages/vat-data/src/kind-utils.js
erights: Now case (4) of the previous conversation, so resolving this one.

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -256,15 +256,15 @@ export function makeVirtualReferenceManager(
    */
   function isDurable(vref) {
     const { type, id, virtual, allocatedByVat } = parseVatSlot(vref);
-    if (type !== 'object') {
+    if (relaxDurabilityRules) {
+      // we'll pretend an object is durable if running with relaxed rules
+      return true;
```
erights, line 261: Open question to reviewers (esp @FUDCo ): With this change, promises are allowed in faux durable storage, just like they are in virtual storage. But this is more tolerant than the development issues we introduced `relaxDurabilityRules` to address. This further relaxation stops more bugs from being caught early, but is not representative of vat boundary erasure. Vat boundary erasure may turn remote presences into far objects, which motivated these erasure rules. But it may only turn remote promises into local promises, where the unsuppressed case still fails to be durable.

erights:
@FUDCo please review at least the SwingSet portion of this PR. Thanks.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1350,11 +1352,7 @@ function build(
     );
 
     // here we finally invoke the vat code, and get back the root object
-    const rootObject = await buildRootObject(
-      harden(vpow),
-      harden(vatParameters),
-      baggage,
-    );
+    const rootObject = await buildRootObject(vpow, vatParameters, baggage);
```
FUDCo, line 1355: Where does `vatParameters` get hardened?  When you moved the hardening of `vpow` to where `vpow` is created I think the hardening of `vatParameters` got lost.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -198,6 +198,11 @@ export function makeVirtualObjectManager(
   unserialize,
   cacheSize,
 ) {
+  const isDurableObject = specimen => {
+    const capData = serialize(specimen);
+    return capData.slots.every(vrm.isDurable);
+  };
+
```
FUDCo: Not entirely sure yet how I feel about exposing this bit of functionality, but I'm leaning towards "I don't like it".  What is this actually being used for?

In any case, it's not `isDurableObject` as much as it's `couldBeDurableObjectIfIWantedItToBe`.  At the very least, the name should reflect that it's not querying the ontological status of the thing, but rather that it's pre-flighting the thing for possible use as a durable object.  But now that I think about it, it's not even doing that, as there is no requirement that `specimen` even be an object; `[47, 'hello']` would pass. What it's really asking is whether `specimen` could be used as a value in a situation where serializing it to durable object state is a possibility.  But if the answer is "no" then what is the action that actual code would take as a consequence?

FUDCo (commented):
W.r.t the changes to SwingSet: Much of it is tweaks dealing with error message censorship, which is always a pain, and my take on these is pretty much "yeah, whatever".  I'm not convinced that the error censorship stuff is adding more value than it subtracts but it's not time to litigate that now.

I do have some serious misgivings about `isDurableObject` (see comment).

packages/zoe/src/makeHandle.js
erights: I get
```
$ yarn lint
yarn run v1.22.19
$ run-s --continue-on-error lint:*
$ eslint .
$ tsc --maxNodeModuleJsDepth 4 -p jsconfig.json
src/makeHandle.js:24:3 - error TS2322: Type '() => KindFacet<{}>' is not assignable to type '() => Handle<H>'.
  Type 'KindFacet<{}>' is not assignable to type 'Handle<H>'.
    Type 'KindFacet<{}>' is not assignable to type 'H'.
      'H' could be instantiated with an arbitrary type which could be unrelated to 'KindFacet<{}>'.

24   return makeHandle;
     ~~~~~~~~~~~~~~~~~~


Found 1 error in src/makeHandle.js:24
```

I suspect this has a simple solution, but I do not know what that is.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: We make use of it at https://github.com/Agoric/agoric-sdk/blob/9b8f3f21a2c5a81dacd71c0233efe80ecdb50a84/packages/zoe/src/contractFacet/offerHandlerStorage.js#L24 to distinguish whether the contract author intended the offerHandler to be upgradable. I think this is a sensible use.

In case I convince you that this predicate should remain visible, what name do you suggest? I am not attached to the current name.

packages/SwingSet/src/liveslots/liveslots.js
erights: https://github.com/endojs/endo/blob/c5582ca7473e0a5d94ef4753ff54e0626cdb1d0a/packages/marshal/src/marshal.js#L507-L510

erights: https://github.com/endojs/endo/blob/c5582ca7473e0a5d94ef4753ff54e0626cdb1d0a/packages/marshal/src/marshal.js#L501

FUDCo: Ah, so anything that comes back from `unserialize` lands in our hands in an already hardened state?  That makes sense.  (But wait: I thought you couldn't harden something twice, so why didn't the previous call to harden (the call you took out that I was asking about) cause an error?)

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: How about `canBeDurable`?  This name makes no commitment about what type of thing it is that's being tested nor about whether its durability is currently enforced or merely a quality of its current state.

packages/vat-data/src/kind-utils.js
erights: @turadg did you mean https://github.com/endojs/endo/blob/c4f23d754d92883b5f470bc74a0b029e152e0e08/packages/marshal/src/make-far.js#L133 ?

packages/ERTP/src/brand.js
erights: Done. Resolving. PTAL and unresolve if more names need to be changed (or for any other reason of course)

packages/ERTP/src/issuerKit.js
erights: @turadg I don't understand the role of `Record` in the last suggestion, and thus I don't understand the last suggestion. We defined a `Baggage` type as `MapStore<string, unknown>` and then used it everywhere we declare baggage. So done enough? Resolving.

packages/ERTP/src/issuerKit.js
erights: Done! Resolving.

packages/ERTP/test/swingsetTests/ertpService/vat-ertp-service.js
erights: This seems related to the previous `Record` suggestion I did not understand. In the hope that this further refinement can wait until after this PR, resolving. But please unresolve if we need to do something here in this PR.

packages/ERTP/src/issuerKit.js
erights: Replaced with `Baggage` which has the `unknown` in it. Done. Resolving.

packages/ERTP/src/issuerKit.js
erights: Done.

packages/vat-data/src/kind-utils.js
erights: I went ahead and used the type at that other location. Done.

packages/zoe/src/makeHandle.js
erights: Gone.

packages/zoe/src/makeHandle.js
erights: @turadg I'm leaving this conversation open until I hear back.

packages/SwingSet/src/liveslots/virtualReferences.js
erights: @FUDCo from your overall review comment, I take it that you are ok with this change as is. It is more than a tweak to the reporting of diagnostic info.

Resolving. But please unresolve if you are not comfortable with this change in behavior.

packages/SwingSet/src/liveslots/liveslots.js
erights: You can always `harden` multiple times. It is fully idempotent. Redundant applications are completely harmless.

I think you're thinking of `Far`. It can only be applied one in order to prevent an attack, among other things.

Resolving.

packages/SwingSet/src/liveslots/virtualReferences.js
FUDCo: I'm OK with it, in the sense that the main person who is likely to stub their toe is you, so if you're comfortable with it then so am I.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Nice! Done. Resolving.

erights:
@turadg PTAL, including the typing conversation I left open above.

@FUDCo Did I resolve all your concerns adequately to move forward with this PR?

packages/SwingSet/src/liveslots/virtualReferences.js
erights: Thanks! Reresolving.

packages/zoe/src/makeHandle.js
turadg: The only simpler solution that comes to mind is a new function like `defineDurableHandle`. The `defineDurableKind` function assumes that the facets object has entries with behaviors and its `KindFacet` wraps those with a context remover.

Alternately we could have the type be conditioned to not use `KindFacet` when the object is empty, but you'd still be getting an empty object and not the `Handle` you're after.

I think the code in the PR is acceptable for now so I'm resolving the comment. If we see this casting repeatedly for making handles it may be worth a new function designed for this use case.

packages/ERTP/src/issuerKit.js
```diff
@@ -3,54 +3,37 @@
...
- *  displayInfo: DisplayInfo,
- * }}
+ * @returns {IssuerKit<K>}
  */
-const makeIssuerKit = (
-  allegedName,
-  // @ts-expect-error K could be instantiated with a different subtype of AssetKind
-  assetKind = AssetKind.NAT,
-  displayInfo = harden({}),
+export const vivifyIssuerKit = (
```
turadg, line 28: ⚡

packages/vat-data/src/kind-utils.js
```diff
@@ -0,0 +1,69 @@
...
+ * https://github.com/Agoric/agoric-sdk/pull/5674#discussion_r908883510
+ * TODO need to move this to a common place.
+ *
+ * @template T, U
+ * @template {keyof T} K
+ * @param {{ [K2 in keyof T]: T[K2] }} original
+ * @param {(pair: [K, T[K]]) => [K, U]} mapPairFn
+ * @returns {Record<K, U>}
+ */
+export const objectMap = (original, mapPairFn) => {
```
turadg: this abstraction doesn't seem right to me. Like `map` of Array, I would expect it to map the value to another value. This leaks the `Object.entries` tuple concept.

the one place this is used would be better served by a cleaner abstraction:
```
const behavior = objectMap(methods, dropContext);
```

packages/vat-data/src/types.d.ts
```diff
@@ -93,6 +96,7 @@ export type VatData = {
     label: string,
     options?: StoreOptions,
   ) => WeakSetStore<K>;
+  canBeDurable: (specimen: any) => boolean;
```
turadg:
```suggestion
  canBeDurable: (specimen: unknown) => boolean;
```

turadg (approved):
Approved with some non-blocking suggestions.

packages/ERTP/src/issuerKit.js
erights: What's the name for that cool icon?

turadg: `:zap:` :smile:

packages/vat-data/src/kind-utils.js
erights: Done. Also consolidated the two implementations of `objectMap` into one and corrected the comment. Much improved!

packages/vat-data/src/types.d.ts
erights: Done

turadg (approved):

PR #5289, opened 2022-05-04
ci: cancel other jobs on pull-requests only

mhofman:
## Description

#5254 added the ability to cancel previously running integration jobs, with the intent to cancel outdated jobs
#5256 did the same for pull request jobs

However the integration jobs were cancelled regardless of the branch on which the job was running. Only integration jobs on pull requests should be canceled and it's perfectly valid for multiple master and beta branches integration jobs to be running concurrently.

### Security Considerations

Make sure we have integration running fully on all master commits.

### Testing Considerations

Hard to test but I have 100% confidence this will work

### DIFF
--- a/.github/workflows/pre-check-integration.yml
+++ b/.github/workflows/pre-check-integration.yml
@@ -30,7 +30,7 @@ jobs:
       - id: step1
         uses: fkirc/skip-duplicate-actions@master
         with:
-          cancel_others: 'true'
+          cancel_others: "${{ github.event_name == 'pull_request' }}"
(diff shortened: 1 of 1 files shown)


mhofman:
https://github.com/Agoric/agoric-sdk/runs/6296525831?check_suite_focus=true is an example of a job on master cancelling https://github.com/Agoric/agoric-sdk/actions/runs/2271944605

mhofman:
I will bypass integration tests on this PR to try and trip the issue if someone merges soon after this lands

michaelfig (approved):
Lgtm!

Issue #5266, opened 2022-05-02
intermittent test failure in vat replay


warner:
https://github.com/Agoric/agoric-sdk/runs/6242581254?check_suite_focus=true#step:5:702

has an intermittent test failure:

```
anachrophobia strikes vat v2
delivery completed with 2 expected syscalls remaining
expected: {"0":"vatstoreGet","1":"vom.rc.o-50","length":2}
expected: {"0":"vatstoreGetAfter","1":"","2":"vom.ir.o-50|","length":3}
REJECTED from ava test: (Error#14)
Error#14: historical inaccuracy in replay of v2
  at Object.finishReplayDelivery (.../swingset-vat/src/kernel/vat-loader/transcript.js:91:13)
  at replayOneDelivery (.../swingset-vat/src/kernel/vat-loader/manager-helper.js:189:19)
  at async Object.replayTranscript (.../swingset-vat/src/kernel/vat-loader/manager-helper.js:220:1)
  at async ensureVatOnline (.../swingset-vat/src/kernel/vat-warehouse.js:134:1)
  at async Object.start (.../swingset-vat/src/kernel/vat-warehouse.js:177:1)
  at async Object.start (.../swingset-vat/src/kernel/kernel.js:1314:1)
  at async makeSwingsetController (packages/SwingSet/src/controller/controller.js:310:3)
  at async buildVatController (packages/SwingSet/src/controller/controller.js:503:22)
  at async packages/SwingSet/test/vat-admin/terminate/test-terminate.js:187:16

  ✖ vat-admin › terminate › terminate › replay does not resurrect dead vat Rejected promise returned by test
```

in a PR based on current trunk (fce10348004383ddb5db79ae6ce7cd4caac0ba1e).

warner:
This appears to be `vat-admin` (v2) on `deliveryNum: 5` receiving an import of the root object (of the soon-to-be-terminated dynamic vat exercised by this test), sending it to the parent vat, and dropping it. Then the `dispatch.bringOutYourDead` on `deliveryNum: 6` performs GC, which should notice the now-finalized Presence object, check that nothing else is referencing the vref, and then informs the kernel.

[...]

PR #5291, opened 2022-05-04
chore(swingset): split test out to a separate file

warner:
We think GC under Node.js is a bit squirrly in the face of AVA's
parallelism, and moving the intermittently-failing test into its own
file might help

refs #5266


FUDCo (approved):
Meh.

PR #5292, opened 2022-05-04
refactor(cosmic-swingset): Improve makeChainStorage parameters

gibson042:
## Description

Replace the confusing `imp` and `exp` parameters with named `fromChainShape` and `toChainShape` options, and also rename `obj` in get and commit in anticipation of non-object values for #4558.

### Security Considerations

n/a

### Documentation Considerations

n/a

### Testing Considerations

n/a

Issue #4558, opened 2022-02-16
merkle-subtree-writing objects


warner:
(not sure what label to apply here, it overlaps swingset and cosmic-swingset, plus using it for notification purposes touches a bunch of core vats)

## What is the Problem Being Solved?

#3756 wants a cheap way for a contract vat to write to some portion of the chain's externally-queriable (and provable) state space. We know we want the Notifier's `updater` API to drive this, and it would be nice to integrate in with the `notifier` half. We know there probably needs to be a device node involved, to execute the write. To make this as cheap as possible, we'd like the device write to happen from the same vat that hosts the `updater` (rather than having a separate "chain state writer service" vat which follows the `notifier` and reacts to updates by invoking the device.

In today's meeting, we discussed having a device nodes that represent specific paths within the chain's state vector. One device node might manage `/swingset/state/amm/pool1`. Another might manage the parent node: `/swingset/state/amm`. Each node gives the holder four authorities:
* write data to that portion of the state vector
* delete data from that portion of the state vector
* create a child device node with a path extension
* report the path of the device node

cosmic-swingset would create the device and give it the ability to write into the state vector (and simultaneously add an Event to the current block, with the same update). Then `bootstrap` would get this device node, and create child nodes to hand off to each vat. The AMM vat would receive `/swingset/state/amm`, and could dole out subsets as it sees fit.

For efficiency, we might use fewer device nodes and provide an API which allows writes to arbitrary child paths beneath the device node's starting point. Then plain JS wrapper objects within the vat would provide the appropriate division of authority. You'd only make a new device node when sending it to a new vat.

In this approach, the kernel is not deeply involved. The device would probably be defined in `packages/cosmic-swingset/`, and the chain-side `bootstrap` would perform the subdivision. The sim-chain bootstrap would need to create some simulated equivalent.

I'll write up an alternate approach in #4559, in which the kernel is deeply aware of this new IO channel, and a device is not used.

cc @gibson042 @michaelfig @dtribble @erights 

## Description of the Design

* cosmic-swingset defines a device
* it receives three endowments:
  * `getRootPath() -> string`
  * `write(path: string, data: string)`
  * `delete(path: string)`
* each device node encapsulates a path
* the root device node encapsulates the root path
* each device node has an API:
  * `write(data: string)`
  * `delete()`
  * `getPath() -> string`
  * `getChild(name: string) -> deviceNode`
  * (maybe) `writeSubpath(subpath: string, data: string)`
[...]

PR #5294, opened 2022-05-04
refactor: Add a buffered StorageAPI helper

gibson042:
As suggested at https://github.com/Agoric/agoric-sdk/pull/5258#discussion_r863276862

## Description

This PR extends #5292, and lays some groundwork for #4558. c67df20001ff1a034a50d3807e0ce1e687492c7f is the most relevant commit.

### Security Considerations

None that I can see.

As commented, I wonder if StorageAPI functions should be extracted into their own module.

### Testing Considerations

`makeBufferedStorage` really should be tested. Any pointers on a good analog?

### DIFF
--- a/packages/SwingSet/src/kernel/state/storageWrapper.js
+++ b/packages/SwingSet/src/kernel/state/storageWrapper.js
@@ -1,7 +1,7 @@
 // @ts-check
 
-import { assert, details as X } from '@agoric/assert';
-import { insistStorageAPI } from '../../lib/storageAPI.js';
+import { assert } from '@agoric/assert';
+import { insistStorageAPI, makeBufferedStorage } from '../../lib/storageAPI.js';
 
 // We wrap a provided object implementing StorageAPI methods { has, getKeys,
 // get, set, delete } (cf. packages/SwingSet/docs/state.md#transactions) and
@@ -14,88 +14,6 @@ import { insistStorageAPI } from '../../lib/storageAPI.js';
 // nature of that parameter, perhaps we should establish some naming convention
 // that signals that the object could be foreign and thus deserving of
 // xenophobia.
-
-/**
- * Given two iterators over sequences of unique strings sorted in ascending
- * order lexicographically by UTF-16 code unit, produce a new iterator that will
- * output the ascending sequence of unique strings from their merged output.
- *
- * @param { Iterator } it1
- * @param { Iterator } it2
- *
- * @yields any
- */
-function* mergeUtf16SortedIterators(it1, it2) {
-  /** @type {IteratorResult<any> | null} */
-  let v1 = null;
-  /** @type {IteratorResult<any> | null} */
-  let v2 = null;
-  /** @type {IteratorResult<any> | null} */
-  let vrest = null;
-  /** @type {Iterator<any> | null} */
-  let itrest = null;
-
-  try {
-    v1 = it1.next();
-    v2 = it2.next();
-    while (!v1.done && !v2.done) {
-      if (v1.value < v2.value) {
-        const result = v1.value;
-        v1 = it1.next();
-        yield result;
-      } else if (v1.value === v2.value) {
-        const result = v1.value;
-        v1 = it1.next();
-        v2 = it2.next();
-        yield result;
-      } else {
-        const result = v2.value;
-        v2 = it2.next();
-        yield result;
-      }
-    }
-
-    itrest = v1.done ? it2 : it1;
-    vrest = v1.done ? v2 : v1;
-    v1 = null;
-    v2 = null;
-
-    while (!vrest.done) {
-      const result = vrest.value;
-      vrest = itrest.next();
-      yield result;
-    }
-  } finally {
-    const errors = [];
-    try {
-      if (vrest && !vrest.done && itrest && itrest.return) {
-        itrest.return();
-      }
-    } catch (e) {
-      errors.push(e);
-    }
-    try {
-      if (v1 && !v1.done && it1.return) {
-        it1.return();
-      }
-    } catch (e) {
-      errors.push(e);
-    }
-    try {
-      if (v2 && !v2.done && it2.return) {
-        it2.return();
-      }
-    } catch (e) {
-      errors.push(e);
-    }
-    if (errors.length) {
-      const err = assert.error(X`Merged iterator failed to close cleanly`);
-      for (const e of errors) {
-        assert.note(err, X`Caused by ${e}`);
-      }
-    }
-  }
-}
 
 /**
  * Create and return a crank buffer, which wraps a storage object with logic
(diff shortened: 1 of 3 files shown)


packages/cosmic-swingset/src/chain-main.js
```diff
@@ -29,65 +30,80 @@ const toNumber = specimen => {
...
+      // Deletion in chain storage manifests as set-to-undefined.
+      storage.set(key, undefined);
     },
-    abort() {
-      // Just reset our state.
-      cache = new Map();
-      changedKeys = new Set();
+    // eslint-disable-next-line require-yield
+    *getKeys(_start, _end) {
+      throw new Error('not implemented');
```
gibson042, line 89: I'm expecting to introduce this for #4558.

FUDCo (approved):
Seems legit.

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -202,22 +90,15 @@ export function buildCrankBuffer(
     hasher.add('\n');
     hasher.add(crankhash);
     hasher.add('\n');
+
+    // Store the new activityhash.
```
warner: By using the buffered kvStore for this following `kvStore.set('activityhash')`, the hash gets put into the pending buffer, not the actual store. To match the previous behavior, I think this needs an additional commit() call afterwards, or perhaps it'd be safe to move the one `commit()` from the start of this method to the end.

Also, we haven't been including the `set('activityhash')` in the activityhash itself. I don't know how to best achieve that with this new layering.. making `onPendingSet` ignore the one key feels weird.

warner (changes requested):
The way swingset manages the kvStore is pretty old and creaky, so refactoring is welcome. That said, I'm slightly skeptical that swingset is a good source for a generic "buffered KV store" library (as your "should this live in a different package?" comment notes). And I don't really understand how cosmic-swingset uses its own buffering layer, so I can't effectively argue for or against sharing an implementation. I'm not excited about deep imports (it imposes hidden constraints on future internal refactoring). But most of those concerns might go away if this moved into a separate package some day.

So I guess I'm -0 on the change.. it's a good refactoring, it feels slightly weird, but it's probably a net win overall.

The activityhash write sequencing definitely needs to change to match the previous commit boundaries. Our tests on activityhash should be observing the writes made to the backing store.. if they aren't, we should improve them, and maybe they should also test two cranks in a row (which might have caught the inclusion of activityhash in the hash).

packages/SwingSet/src/lib/storageAPI.js
```diff
@@ -40,3 +42,166 @@ export function insistEnhancedStorageAPI(kvStore) {
...
+      if (additions.has(key)) return true;
+      if (deletions.has(key)) return false;
+      return kvStore.has(key);
+    },
+    get(key) {
+      assert.typeof(key, 'string');
+      if (additions.has(key)) return additions.get(key);
+      if (deletions.has(key)) return undefined;
+      const value = kvStore.get(key);
+      if (onGet !== undefined) onGet(key, value);
```
warner, line 166: is there any particular reason to not use `if (onGet) onGet(key, value)` ?

packages/SwingSet/src/kernel/state/storageWrapper.js
gibson042: `kvStore` in this scope is the backing StorageAPI instance, not the `crankBuffer` wrapping it. These bindings are identical to those currently in `buildCrankBuffer`, and so is the call sequencing AFAICT.

**Current**: https://github.com/Agoric/agoric-sdk/blob/6a5f377c5ba115576a0e813cc6266039452732c0/packages/SwingSet/src/kernel/state/storageWrapper.js
* [`crankBuffer.set()`](https://github.com/Agoric/agoric-sdk/blob/6a5f377c5ba115576a0e813cc6266039452732c0/packages/SwingSet/src/kernel/state/storageWrapper.js#L148) and [`crankBuffer.delete()`](https://github.com/Agoric/agoric-sdk/blob/6a5f377c5ba115576a0e813cc6266039452732c0/packages/SwingSet/src/kernel/state/storageWrapper.js#L163) update `additions` and `deletions` and feed consensus-relevant data to `crankhasher`.
* [`commitCrank()`](https://github.com/Agoric/agoric-sdk/blob/6a5f377c5ba115576a0e813cc6266039452732c0/packages/SwingSet/src/kernel/state/storageWrapper.js#L182) sends a sequence of calls to the backing `kvStore`:
  1. `kvStore.set(key, value)` for each addition
  2. `kvStore.delete(key)` for each deletion
  3. `oldActivityhash = kvStore.get('activityhash')`
  4. `kvStore.set('activityhash', activityhash)`, where `activityhash` is a digest of data including `oldActivityhash` and `crankhasher.digest()` (before the replacement of `crankhasher` between steps 2 and 3).

**Proposed**: https://github.com/Agoric/agoric-sdk/blob/45406c68018b36f94cfd5cd380be888ca31ed3ac/packages/SwingSet/src/kernel/state/storageWrapper.js
* `crankBuffer.set()` and `crankBuffer.delete()` update `additions` and `deletions` [_[buffered storage `set()` and `delete()`](https://github.com/Agoric/agoric-sdk/blob/45406c68018b36f94cfd5cd380be888ca31ed3ac/packages/SwingSet/src/lib/storageAPI.js#L125-L136)_] and feed consensus-relevant data to `crankhasher` [_[`buildCrankBuffer` behavior via buffered storage `onPendingSet` and `onPendingDelete`](https://github.com/Agoric/agoric-sdk/blob/45406c68018b36f94cfd5cd380be888ca31ed3ac/packages/SwingSet/src/kernel/state/storageWrapper.js#L46-L63)_].
* [`commitCrank()`](https://github.com/Agoric/agoric-sdk/blob/45406c68018b36f94cfd5cd380be888ca31ed3ac/packages/SwingSet/src/kernel/state/storageWrapper.js#L73) sends a sequence of calls to the backing `kvStore`:
  1. `kvStore.set(key, value)` for each addition and then `kvStore.delete(key)` for each deletion [_in [buffered storage `commit()`](https://github.com/Agoric/agoric-sdk/blob/45406c68018b36f94cfd5cd380be888ca31ed3ac/packages/SwingSet/src/lib/storageAPI.js#L190)_]
  2. `oldActivityhash = kvStore.get('activityhash')`
  3. `kvStore.set('activityhash', activityhash)`, where `activityhash` is a digest of data including `oldActivityhash` and `crankhasher.digest()` (before the replacement of `crankhasher` after step 1).


I've added clarifying comments: https://github.com/Agoric/agoric-sdk/blob/80afe8640d8e683e729eed3bc809bb04c5551280/packages/SwingSet/src/kernel/state/storageWrapper.js#L81-L98

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -66,35 +61,49 @@ const makeChainStorage = (call, prefix = '', options = {}) => {
...
-  return storage;
+  const {
+    kvStore: buffered,
+    commit,
+    abort,
+  } = makeBufferedStorage(storage, {
+    // Enqueue a write of any retrieved value, to handle callers like mailbox.js
+    // that expect local mutations to be automatically written back.
+    onGet(key, value) {
+      buffered.set(key, value);
```
warner, line 100: I can't quite trace what's happening here. There's a weird property that I think we need to maintain: if the caller does `x = s.get('foo'); x.prop = 'changed'; s.commit()`, then we need to write back the modified `.prop` to the backing store. (I'm embarrassed that this seems to be necessary, and I'm sure that it's because I made a really weird API for the original mailbox structure, and my only defense is that this was some of the earliest Swingset code ever written).

Does this code maintain that property? I can tell there's a cache here at this level (`storage`), and a buffer of pending writes at the `makeBufferedStorage` level, but I'd like the expert to tell me if this `buffered.set` is sufficient to get the same value into both places.

I like `makeBufferedStorage` on the swingset side, it seems like a nice refactoring. I'm a little dubious about the utility using it here if maintaining that funky mutable-read-data behavior requires something weird. But if it works, and it supports the next phase of your work, I'm ok with it.

I *would* feel better if there were tests on this code, and I believe there are not. `makeBufferedStorage` should be testable within swingset, and this `storage` definition should be testable in isolation (from swingset and from the chain's backing store).

warner (approved):
Tentatively approving. Please get @michaelfig 's opinion on the chain-main.js changes to make sure I didn't miss something, and please find a way to write a test for it. `packages/SwingSet/test/test-state.js` has tests for `buildCrankBuffer`, maybe you could crib something from there to build a new `packages/cosmic-swingset/test/test-chain-storage.js`.

packages/SwingSet/src/kernel/state/storageWrapper.js
warner: Ah, gotcha, I was confused by the overlapping names (and I never remember which way around JS's destructuring assignment works). The extra comments helped a lot.

michaelfig:
I'd like to test this PR against actual chain+ag-solo, but I think I have to wait for you, @gibson042, to resolve the conflicts and bring up-to-date with current master.

packages/SwingSet/src/lib/storageAPI.js
gibson042: No, the behavior would be equivalent either way. Is the coercing form preferred?

packages/cosmic-swingset/src/chain-main.js
gibson042: > Does this code maintain that property?

The short answer is "yes". This `onGet` configuration for the `makeBufferedStorage()` call that wraps the `chainSend()`-backed storage interface results in every `get(key)` call on the returned storage interface being immediately followed with a `set(key, fetchedValue)` that adds a reference to its `additions` map (where `commit()` will find it and translate to a `set(key, valueWithAnyMutations)` against the `chainSend()`-backed storage—which performs its own translation to a stringified {method: 'set', key, value} message). All of which is intended to match current behavior, complete with idiosyncrasies.

michaelfig (approved):
Great stuff!

PR #5297, opened 2022-05-05
Persistently track transcript start position

### LINKED ISSUE #3293: Track transcript SwingStore start position and have upgrade reset to end rather than beginning
(Note: this task has been redefined, see https://github.com/Agoric/agoric-sdk/issues/3293#issuecomment-1113885212 below.  We are retaining the issue with its original summary for continuity, but the work to be done here is rather different from what was originally outlined.)

The SwingStore's stream API provides read and write operations but not delete.  Delete is not currently needed, but we _will_ need it if we wish to retire a vat completely and reclaim its resources.

## Description of the Design

The proposal here is to augment the `StreamStore` API with a `deleteStream(streamName)` method, which will operate in a manner similar to the existing `closeStream` call but which will also delete the associated stream file.

## Security Considerations

I do not believe there are any particular security considerations here above and beyond those posed by access to the `StreamStore` API in the first place.  Obviously there will remain policy questions about whether or not deleting a stream in any particular case is a good idea or not, in particular when we think keeping old streams around may be useful as a diagnostic aid.  However, that is a pragmatic consideration rather than a security consideration per se.

## Test Plan

Augment the existing `StreamStore` API tests to exercise this operation and verify that the associated stream file is, in fact, gone after its use.

FUDCo:
Keep track of transcript start position in the kernel DB so that transcript continuity is preserved across upgrade boundaries.

Closes #3293

### DIFF
--- a/packages/SwingSet/src/kernel/state/kernelKeeper.js
+++ b/packages/SwingSet/src/kernel/state/kernelKeeper.js
@@ -71,7 +71,8 @@ const enableKernelGC = true;
 //   $vatSlot is one of: o+$NN/o-$NN/p+$NN/p-$NN/d+$NN/d-$NN
 // v$NN.c.$vatSlot = $kernelSlot = ko$NN/kp$NN/kd$NN
 // v$NN.nextDeliveryNum = $NN
-// v$NN.t.endPosition = $NN
+// v$NN.t.startPosition = $NN // inclusive
+// v$NN.t.endPosition = $NN  // exclusive
 // v$NN.vs.$key = string
 // v$NN.meter = m$NN // XXX does this exist?
 // v$NN.reapInterval = $NN or 'never'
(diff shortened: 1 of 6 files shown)


packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -71,6 +71,7 @@ const enableKernelGC = true;
 //   $vatSlot is one of: o+$NN/o-$NN/p+$NN/p-$NN/d+$NN/d-$NN
 // v$NN.c.$vatSlot = $kernelSlot = ko$NN/kp$NN/kd$NN
 // v$NN.nextDeliveryNum = $NN
+// v$NN.t.startPosition = $NN
```
warner: `startPosition` is inclusive and `endPosition` is exclusive, right? So when they're equal, that means the transcript is empty? Might be worth a brief comment to remind us.

warner (approved):
Looks great, and the test is perfect.

packages/SwingSet/src/kernel/state/kernelKeeper.js
FUDCo: Yes, that's right.  I will add a comment.

PR #5302, opened 2022-05-06
Defend against LMDB key size overflow

### LINKED ISSUE #5277: LMDB key length limits can crash the kernel

Our friends at Atredis pointed out that LMDB has some surprisingly small key-length limits, and there are ways for userspace to trigger these, causing an LMDB error as the block buffer is committed (which would crash the host application).

With the framing we put around Store keys, the largest key that can be stored in a low-numbered collection (`collectionID = '2')` for a low-numbered vat (`vatID = 'v2'`) is 242 characters long.

```js
c1 = VatData.makeScalarBigMapStore('myData');
c1.init('A'.repeat(243), 'B'.repeat(32));
```

That `Store.init` key expands to a `syscall.vatstoreSet` key of `vc.2.sAAAAAA..`, which expands to an LMDB key of `v2.vs.vc.2.sAAAA..` . That sounds like a key-size limit of 254-ish bytes.

Since Stores are frequently used to hold remotely-provided data (e.g. names), this could probably be triggered remotely.

## Description of the Design

The long-term fix is to move away from LMDB to something like SQLite, where key lengths are much less restricted.

The short-term fix is to have `Store.init` throw an error if given a key that is too long. We can choose a safe limit like 128 characters, giving us plenty of headroom to guard against long vatIDs and collectionIDs (whose identifiers are included in the prefix, and thus count against the limit). The other methods should treat oversized keys as not existing in the table (so `has` returns `false`, `get` and `delete` throw a "no such key" error), without encoding the key or submitting it to a `vatstore` operation.

We can impose the same limit on userspace `vatstore.get()/set()/etc` calls, if those are enabled by `options.enableVatstore` (which defaults to `false`). Both limitations are imposed by liveslots. By throwing a catchable error, userspace can deny specific requests by abusive clients without losing the ability to provide good service to well-behaved clients.

At the `syscall.vatstoreSet` layer, we can impose a larger limit (maybe 200 bytes, giving some headroom for the vatID), and make it a vat-fatal error to exceed the limit. This protects the kernel (really the host application) against a renegade liveslots or compromised worker process, which gets full syscall access.

The kernel should start `initializeSwingset` by writing a DB record with a 250-byte key. If the kernel is given a SwingStore that cannot handle keys of this size, this check will let us find out early. It'd be nice to delete this key after the check is complete, but given the single initial DB commit, I'm not sure there's a good time to do it. (and we can't delete it during `initializeSwingset`, because then it won't actually test the DB's abilities).

Later, if/when we move to a more capable database, we can enlarge or remove the syscall limit, and release a new version of liveslots (used only by new vats) which has a corresondingly larger internal limit. (Moving to SQLite is likely to change the syscall.vatstore API significantly anyways).

## Alternate Designs

I briefly considered a more detailed approach, in which the SwingStore's `kvStore` has an additional method named `getMaxKeySize()`, and the kernel reads that to decide what limit to impose upon `syscall.vatstore` operations (where long vatIDs would get a slightly shorter limit), and it tells liveslots about the per-vat limit, so liveslots can compute the per-collection limit (again, slightly shorter for long `collectionID` prefix strings), and impose it upon Store operations (which know their own key encoding overhead).

That's far too much work for the flexibility it buys.

## Security Considerations

For MN-1, it's critical that we do not allow remote callers to crash the entire kernel by passing a surprisingly large name string. For MN-2/3, it will be equally critical that we protect the kernel against individual vats deliberately using a large key.

The limit must be enforced identically on all validators.

## Test Plan

Unit tests

FUDCo:
Closes #5277

Probably easiest to review each of the two commits separately.


packages/SwingSet/test/stores/test-collections.js
```diff
@@ -408,6 +408,35 @@ test('set fail on concurrent modification', t => {
   }, m(`keys in store cannot be added to during iteration`));
 });
 
+test('fail on oversized keys', t => {
```
warner, line 411: Let's add at least one knife-edge test: assert that exactly 219 chars passes, but exactly 220 chars fails. If we're making this part of the observable/consensus behavior, we should make it clear that it can't be changed without carefully considering the compatibility consequences.

warner (approved):
Deleting code is a joyous thing.

One small test tweak, then land with great rejoicing.

PR #5305, opened 2022-05-06
fix(run-protocol): add diagrams for reserve, and fix a typo in the reserve

### LINKED ISSUE #5303: Sequence Diagram: Reserve
@arirubinstein commented on [Wed May 04 2022](https://github.com/Agoric/agoric-private/issues/14)

https://github.com/Agoric/agoric-sdk/tree/master/packages/run-protocol/src/reserve

arirubinstein:
closes: #5303

## Description

Adds a sequence diagram for methods in the reserve, and corrects a typo in the invitation descriptor

### DIFF
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -90,7 +90,7 @@ const start = async (zcf, privateArgs) => {
   };
 
   const makeAddCollateralInvitation = () =>
-    zcf.makeInvitation(addCollateralHook, 'Add Collateal');
+    zcf.makeInvitation(addCollateralHook, 'Add Collateral');
 
   /** @type {ZCFMint} */
   const runMint = await zcf.registerFeeMint('RUN', feeMintAccess);
--- a/packages/run-protocol/src/reserve/doc.md
+++ b/packages/run-protocol/src/reserve/doc.md
@@ -1,0 +1,129 @@
+# Reserve 
+
+### addIssuer
+
+```mermaid
+sequenceDiagram
+
+participant caller
+participant reserveCreatorFacet
+participant ammPublicFacet
+participant brandForKeyword
+participant keywordForBrand
+participant zcf
+
+
+caller ->> reserveCreatorFacet: addIssuer(inIssuer, inKeyword)
+reserveCreatorFacet ->> reserveCreatorFacet: get brand from inIssuer
+reserveCreatorFacet ->> ammPublicFacet: get liquidity issuer for inIssuer
+reserveCreatorFacet ->> reserveCreatorFacet: get liquidity brand from liquidity issuer
+reserveCreatorFacet ->> brandForKeyword: store inKeyword => [brand, inKeyword]
+reserveCreatorFacet ->> keywordForBrand: store liqudityIssuer => inKeyword
+par do
+reserveCreatorFacet ->> zcf: Save Issuer: inIssuer, inKeyword
+reserveCreatorFacet ->> zcf: Save Issuer: liquidityIssuer,  keyword + "Liquidity"
+end
+```
+
+### getKeywordForBrand
+
+```mermaid
+sequenceDiagram
+
+participant caller
+participant reserveContract
+participant keywordForBrand
+
+
+caller ->> reserveContract: getKeywordForBrand(inBrand)
+reserveContract ->> keywordForBrand: check for inBrand membership
+reserveContract ->> keywordForBrand: get for inBrand object
+reserveContract ->> caller: return keyword
+ 
+```
+
+
+### collateralSeat
+
+```mermaid
+sequenceDiagram
+participant caller 
+participant reserveCreatorFacet
+participant collateralSeat
+
+caller ->> reserveCreatorFacet: getAllocations()
+reserveCreatorFacet ->> collateralSeat: getCurrentAllocation()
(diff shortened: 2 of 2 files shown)


packages/run-protocol/src/reserve/doc.md
```diff
@@ -0,0 +1,129 @@
+# Reserve 
+
+### addIssuer
+
+```mermaid
+sequenceDiagram
+
+participant caller
+participant addIssuer
```
turadg: `addIssuer` is [a method of the reserve's creator facet](https://github.com/Agoric/agoric-sdk/blob/2b5e79c8671d9f3cd891e28b0b20de8df4e3bca9/packages/run-protocol/src/reserve/assetReserve.js#L167).
```suggestion
participant reserveCreatorFacet
```

packages/run-protocol/src/reserve/doc.md
```diff
@@ -0,0 +1,129 @@
...
+participant ammPublicFacet
+participant brandForKeyword
+participant keywordForBrand
+participant zcf
+
+
+caller ->> addIssuer: addIssuer(inIssuer, inKeyword)
```
turadg: not sure what `in` means here. `addIssuer` has no constraint on the what the issuer is used for.
```suggestion
caller ->> addIssuer: addIssuer(issuer, keyword)
```

packages/run-protocol/src/reserve/doc.md
```diff
@@ -0,0 +1,129 @@
...
+addIssuer ->> addIssuer: get brand from inIssuer
+addIssuer ->> ammPublicFacet: get liquidity issuer for inIssuer
```
turadg:
```suggestion
addIssuer ->> ammPublicFacet: get liquidity issuer for issuer's brand
```
Though I'm not clear on what a "liquidity issuer" is. @Chris-Hibbert would know.

packages/run-protocol/src/reserve/doc.md
```diff
@@ -0,0 +1,129 @@
...
+participant caller
+participant addIssuer
+participant ammPublicFacet
+participant brandForKeyword
+participant keywordForBrand
+participant zcf
+
+
+caller ->> addIssuer: addIssuer(inIssuer, inKeyword)
+addIssuer ->> addIssuer: get brand from inIssuer
```
turadg: relevant?

packages/run-protocol/src/reserve/doc.md
```diff
@@ -0,0 +1,129 @@
...
+addIssuer ->> ammPublicFacet: get liquidity issuer for inIssuer
+addIssuer ->> addIssuer: get liquidity brand from liquidity issuer
+addIssuer ->> brandForKeyword: store inKeyword => [brand, inKeyword]
+addIssuer ->> keywordForBrand: store liqudityIssuer => inKeyword
+par do
```
turadg, line 22: TIL `par`

packages/run-protocol/src/reserve/doc.md
```diff
@@ -0,0 +1,129 @@
...
+addIssuer ->> zcf: Save Issuer: inIssuer, inKeyword
+addIssuer ->> zcf: Save Issuer: liquidityIssuer,  keyword + "Liquidity"
+end
+```
+
+### getKeywordForBrand
```
turadg, line 28: what's the value of diagramming this? how will it be used?

turadg (commented):
Took a pass and stopped when I realized I'm not sure how these diagrams will be used. Not sure how to assess whether they do the job.

If you're looking for a more narrow review, lemme know.

(also, this is not a `fix` commit that should bump the version number. I think it's `docs:`)

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -90,7 +90,7 @@ const start = async (zcf, privateArgs) => {
   };
 
   const makeAddCollateralInvitation = () =>
-    zcf.makeInvitation(addCollateralHook, 'Add Collateal');
+    zcf.makeInvitation(addCollateralHook, 'Add Collateral');
```
Chris-Hibbert, line 93: This is probably my most common typo these days. I should add a check for it.

packages/run-protocol/src/reserve/doc.md
Chris-Hibbert: Providers of liquidity to a pool in the AMM get Liquidity tokens (represented by a unique ERTP issuer) corresponding to the secondary currency in that pool. The Reserve needs to know about that because it's providing liquidity, and will get the corresponding liquidity tokens. The diagram looks correct to me.

packages/run-protocol/src/reserve/doc.md
arirubinstein: these are intended for the next audit per our rules of engagement/scope of work.  This is the first to show parties and flow, however I intend to produce one of these for each contract in run protocol

packages/run-protocol/src/reserve/doc.md
arirubinstein: in was to designate initial input to the method - aka "attacker" / caller-controlled

packages/run-protocol/src/reserve/doc.md
arirubinstein: I reworked the diagram to call the facet with the methods

turadg (approved):

PR #5301, opened 2022-05-06
chore(xsnap): Upgrade moddable submodule

mhofman:
refs: #3769

## Description

Update the Moddable SDK to the latest version including fixes for snapshoting determinism. Now with the latest improvements, a worker loaded from the same state (new or same snapshot), with the same trace, will have the same snapshot hash.

Update the deployment integration test to check that the swingstore traces of the 2 validator nodes, and thus their snapshots, are identical.

The next frontier is consistent snapshot when loaded from different points (e.g. new + replay vs loaded from snapshot).

### Security Considerations

The new Moddable SDK comes with some fuzzing related fixes.

### Documentation Considerations

N/A

### Testing Considerations

Ran a full deployment integration test locally to verify snapshot consistency.

PR #5322, opened 2022-05-06
fix(deployment): Update CI script

mhofman:
refs: #5301

## Description

Print details on swingstore mismatch failures
Ensure that teardown doesn't cause job failure
`tar` xsnap-trace folders to avoid large number of files slowing CI

### Testing Considerations

Tested extensively locally, will force integrations on this PR


michaelfig (approved):
LGTM!

For future style reference, I prefer:

```
if COND; then
  consequent
fi
```

It reads a little more compactly.

PR #5324, opened 2022-05-08
Prevent bootstrap coreProposal path nondeterminism

michaelfig:
## Description

Validators on the `agoricdev-11` network were getting AppHash errors for block 1.  As it turned out, the diverging validators had their `agoric-sdk` living in a different directory (such as /root/agoric-sdk) than the instagoric validators (/usr/src/agoric-sdk).

It turns out that the bootstrap `coreProposal` code, specifically `extractCoreProposalBundles`, was putting the absolute source and bundle paths into the proposal that was being evaluated on chain in block 1.  This resulted in divergent behaviour when those paths were different, because the core proposal code was attached to the first bootstrap message delivery, and hence recorded in the bootstrap vat's transcript.

The solution is to only record the `bundleID` in the on-chain proposal.  The local paths should only be added to the SwingSet config, since that configuration is outside of consensus.

### Security Considerations

Improves resilience by removing local path considerations from the kernel db.  This also prevents a potential privacy leak if validators share their kernel db with other parties.

No doc concerns.

### Testing Considerations

Manual testing was performed.  Will be tested in a stage network as well by adding a validator with a different path to the coreProposals.

### DIFF
--- a/packages/deploy-script-support/src/coreProposalBehavior.js
+++ b/packages/deploy-script-support/src/coreProposalBehavior.js
@@ -19,7 +19,7 @@ export const permits = {
  * @param {[string, ...unknown[]]} opts.getManifestCall
  * @param {typeof import('@endo/far').E} opts.E
  * @param {(...args: unknown[]) => void} [opts.log]
- * @param {(ref: string) => Promise<unknown>} [opts.restoreRef]
+ * @param {(ref: unknown) => Promise<unknown>} [opts.restoreRef]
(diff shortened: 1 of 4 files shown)


dckc:
My spidey sense [noticed](https://github.com/Agoric/agoric-sdk/pull/5257#pullrequestreview-959291542) the lack of ocap discipline... We were wrong to treat it as just a maintainability issue that could be postponed. Here's hoping we learned the lesson...

dckc (approved):

PR #5331, opened 2022-05-10
fix: array.map that should be array.forEach

erights:
## Description

Calls to `array.map` that do not use the resulting array should call `array.forEach` instead.

### Security Considerations

None

None

### Testing Considerations

None

### DIFF
--- a/packages/SwingSet/src/kernel/state/kernelKeeper.js
+++ b/packages/SwingSet/src/kernel/state/kernelKeeper.js
@@ -611,7 +611,7 @@ export default function makeKernelKeeper(
           body: kvStore.get(`${kernelSlot}.data.body`),
           slots: commaSplit(kvStore.get(`${kernelSlot}.data.slots`)),
         };
-        p.data.slots.map(parseKernelSlot);
+        p.data.slots.forEach(parseKernelSlot);
         break;
       default:
         assert.fail(X`unknown state for ${kernelSlot}: ${p.state}`);
--- a/packages/SwingSet/src/kernel/vatTranslator.js
+++ b/packages/SwingSet/src/kernel/vatTranslator.js
@@ -117,7 +117,7 @@ function makeTranslateKernelDeliveryToVatDelivery(vatID, kernelKeeper) {
     const vrefs = krefs.map(kref =>
       mapKernelSlotToVatSlot(kref, gcDeliveryMapOpts),
     );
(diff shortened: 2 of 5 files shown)


FUDCo (changes requested):
You need to remove xsnap from the commit.

FUDCo:
BTW, this is the kind of hygiene thing that's very easy to do and can be dropped in at pretty much any time, so I'd rather have it filed as a bug so that we can slot it in between other discreet units of kernel work, rather than having to deal with the inevitable merge conflicts that will arise if it interleaves with kernel tasks that are in progress.

FUDCo:
Also, there should be bug # anyway so we can track it.

FUDCo (approved):
Oh alright.

turadg:
Does our style have an opinion on `forEach` vs `for...of` ? Some reasons for the latter: https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md

erights:
If does not. I use them approx synonymously, with a mild preference for `for..of` when there is no other reason to choose.

Good to see someone has gathered arguments on one side. I certainly find it plausible that this article will convince us to give up on `forEach`.

mhofman:
FYI the [airbnb style guide](https://github.com/airbnb/javascript#iterators--nope) makes the other recommendation. I admittedly have the same approach as Mark here, use both depending on "feeling", with a strong preference for `for..of`. I've always found the AirBnb recommendation misplaced in this case.

Btw, a major reason to go for `for..of` not mentioned anywhere in these stylistic guidelines is typing inference for `tsc`, which is incapable of knowing if a closure is immediately invoked or will be called some other time, which prevents it from doing control flow analysis of the closed over variables.

turadg:
>  I've always found the AirBnb recommendation misplaced in this case.

Yeah, something seems to have gotten lost in translation to the example. I mean, the rationale is good, "Dealing with pure functions that return values is easier to reason about than side effects." and then lists things that are good fp style (no mention of `forEach`) but then the example demonstrates with `forEach`. Maybe to give an intermediate "good" that is similar to a loop to contrast with "best" . 🤷🏻‍♂️

erights:
@Chris-Hibbert @FUDCo and I just ran across a minor irritation with `forEach` as compared with `for..of`. `for..of` has a better debugging experience. But not by a lot.

PR #5337, opened 2022-05-10
fix(run-protocol): raise demo issuer debt limit

samsiegart:
The minimum initial debt is 5 RUN on sim chain, so the debt limit has to be higher than that


dckc (approved):
To save the next person some trouble, that 1M arbitrary constant buried in the implementation should be hoisted to `AMMDemoState` by way of `defaultConfig`.

But my attention is elsewhere just now, so let's not let the perfect be the enemy of the good...

Issue #5227, opened 2022-04-26
XS snapshot hash determinism


mhofman:
## Describe the bug
A XS worker started from the same state (either from scratch or from the same snapshot), and processing the exact same commands should result in identical snapshots taken at the same point.

## To Reproduce
Check the content of `validator-swingstore-trace.diff` in the "deployment" integration test results, shown in the "capture results" step of the CI job.

## Expected behavior
Identical swing store trace

## Additional context
While snapshots are not currently part of consensus state, there should currently be no reason for them to diverge. There are known differences in snapshots when starting from a different "state" (e.g. scratch vs snapshot, or different snapshot points), which are tracked separately (#5330).

To generate xsnap trace diff (from an integration test result artifact):

```sh
(mkdir validator0-xsnap-trace \
  && cd $_ && tar -xzf ../$_.tgz \
  && for v in *; do ln -s $v $(jq -r '.name | split(":") | .[0]' $v/00000-options.json) ; done; )
(mkdir validator1-xsnap-trace \
  && cd $_ && tar -xzf ../$_.tgz \
  && for v in *; do ln -s $v $(jq -r '.name | split(":") | .[0]' $v/00000-options.json) ; done; )
for v in validator0-xsnap-trace/v*; do for file in $v/*; do 
  diff -U0 $file validator1${file#validator0};
done; done | grep -v "No newline at end of file" > xsnap-trace.diff
```

## Screenshots

<details>

<summary>xsnap-trace.diff</summary>

```diff
--- v1/00158-snapshot.dat	2022-04-25 19:19:42.000000000 -0700
+++ ../validator1-xsnap-trace/v1/00158-snapshot.dat	2022-04-25 19:19:42.000000000 -0700
@@ -1 +1 @@
-/home/ag-chain-cosmos/.ag-chain-cosmos/data/ag-cosmos-chain-state/xs-snapshots/save-raw-Tkhy7E.xss
+/home/ag-chain-cosmos/.ag-chain-cosmos/data/ag-cosmos-chain-state/xs-snapshots/save-raw-a10vbL.xss
--- v1/01525-snapshot.dat	2022-04-25 19:21:00.000000000 -0700
+++ ../validator1-xsnap-trace/v1/01525-snapshot.dat	2022-04-25 19:21:01.000000000 -0700
@@ -1 +1 @@
[...]

PR #5338, opened 2022-05-10
feat: improve snapshot consistency

mhofman:
refs: #5227 

## Description

Pull in the latest Moddable SDK changes to fix another occurrence of snapshot content non-determinism.
Enable random snapshot memory initialization in the deployment integration test, allowing to force detection of some non-determinism occurrences.
Add an env controlled option to keep the XS snapshots on disk once no longer in use.
Add a script automating generating minimal traces and diffs when snapshots content differs.

### Security Considerations

None

New (undocumented) xnsap environment variable `XSNAP_RANDOM_INIT` impacting XS compilation. Does not impact behavior.
New (undocumented) swingset host (chain + solo) environment variable `XSNAP_KEEP_SNAPSHOTS` instructing the snap store to no delete unused snapshots from disk (but keep deleting them from tracked snapshots)

### Testing Considerations

Tested locally, will force integration test to run.


.github/workflows/deployment-test.yml
```diff
@@ -39,6 +40,8 @@ jobs:
         with:
           node-version: 14.x
           cache-built: false
+        env:
+          XSNAP_RANDOM_INIT: '1'
```
warner: since XS and xsnap don't document this, it might be worth a short comment here about what this does, somebody grepping widely enough might find it

warner (approved):
Looks good!

mhofman:
@warner @michaelfig PTAL

I extended the scope to capture "deleted" snapshots and add a script to generate an artifact that can be submitted to moddable for them to investigate the divergences.

I have not automated the generation of this artifact in CI but I probably could figure something out if we think it'd be valuable (right now I just plan on running this locally)

mhofman:
@kriskowal this also fixes the merge commit issue I raised in https://github.com/Agoric/agoric-sdk/pull/5380#issuecomment-1131153010

image

kriskowal:
@mhofman For my edification, how does having a submodule pinned to the right parent of a merge commit manifest as a problem?

mhofman:
@kriskowal it wouldn't have been a problem if that branch was up to date, but it wasn't, as shown in the screenshot, which means there actually is a [diff](https://github.com/agoric-labs/xsnap-pub/compare/7e0acce423171a5494cff25bb0c8df043520ff35...d708a9daa89f536bfda22d726e1c45aa716e788c) when updating to the merge commit itself.

kriskowal:
> @kriskowal it wouldn't have been a problem if that branch was up to date, but it wasn't, as shown in the screenshot, which means there actually is a [diff](https://github.com/agoric-labs/xsnap-pub/compare/7e0acce423171a5494cff25bb0c8df043520ff35...d708a9daa89f536bfda22d726e1c45aa716e788c) when updating to the merge commit itself.

Ah, I see, so the problem was simply that HEAD had advanced ahead of my base before I merged without my noticing. I’ll go fiddle with that project’s settings so that is less likely in the future.

packages/swing-store/src/snapStore.js
```diff
@@ -161,7 +164,9 @@ export function makeSnapStore(
     for (const hash of toDelete) {
       const fullPath = hashPath(hash);
       try {
-        unlinkSync(fullPath);
+        if (keepSnapshots !== true) {
```
warner, line 238: that seems reasonable, it's not something that could be turned on and off over time (we don't persist the list of deletable files, so once `commitDeletes` is called, we forget entirely about the deletability of each file), but I don't think that's a problem for this use case

warner (approved):
looks good

packages/swing-store/src/snapStore.js
mhofman: Yeah I double checked what happens if the snap-store tries to write a file which already exists.

There is first a `existsSync` check, so if the file exists then, it simply doesn't try to write the file to disk.
https://github.com/Agoric/agoric-sdk/blob/6747ffc6f520990cec223302a65b1551cb44cdc2/packages/swing-store/src/snapStore.js#L127-L129

Then no matter what, `atomicWrite` uses [`fs.rename`](https://nodejs.org/docs/latest-v16.x/api/fs.html#fsrenameoldpath-newpath-callback), which overwrites the destination in case it already existed. So there is no way this could fail.
https://github.com/Agoric/agoric-sdk/blob/6747ffc6f520990cec223302a65b1551cb44cdc2/packages/swing-store/src/snapStore.js#L79

michaelfig (approved):
Plumbing looks good.  I'm excited about further progress towards deterministic snapshots.

mhofman:
@warner I've updated the Moddable SDK to the latest version. I'm double checking to see if anything is worth our attention, and wanted to give you a heads up since it's no longer what you had reviewed (the rest hasn't changed besides rebasing on top of #5543)

mhofman:
Here is the full diff for reference: https://github.com/agoric-labs/moddable/compare/2fdfb4cd9d1a1a8ae72dda28669b44788c9b700c...2e30065710fed92484efa854e9646abcc54b927a

mhofman:
I see a bunch of tweaks to the JS engine, but nothing that should impact us. `Object.hasOwn` and `Error#cause` are now supported, and there seem to be some tweaks to dynamic import and their lockdown, which we don't use.

warner:
Oh I should have mentioned, since this updates the moddable SDK, we need to bump the `METER_TYPE` in `packages/xsnap/api.js`.

I've got another PR in the works to change xsnap, I'll include the update in that one.

mhofman:
> we need to bump the `METER_TYPE` in `packages/xsnap/api.js`.

When does the meter type need to be bumped? Should this be automatically derived from the XS version / revision somehow ?

PR #5339, opened 2022-05-11
Fix bug in fake virtual stuff due to recent liveslots additions

FUDCo:
In support of #5283.

This deals with some version drift between liveslots and the fake virtual object mock.  The fake virtual object stuff is able to achieve great verisimilitude because it actually uses the real virtual object stuff to do all the virtual object stuff.  But it has to fake some stuff from liveslots and that was what was faulty.

@erights and @Chris-Hibbert encountered this in the course of testing their
durabilization of ERTP.  Note that this bug only implicates tests and not the
actual underlying virtual objects implementation.

### DIFF
--- a/packages/SwingSet/tools/fakeVirtualSupport.js
+++ b/packages/SwingSet/tools/fakeVirtualSupport.js
@@ -179,15 +179,23 @@ export function makeFakeLiveSlotsStuff(options = {}) {
   }
 
   function convertSlotToVal(slot) {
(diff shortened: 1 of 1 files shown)


FUDCo:
@erights @Chris-Hibbert

erights (approved):
As far as I understand it, LGTM

FUDCo:
This appears to have surfaced another issue that I'm now chasing...

PR #5340, opened 2022-05-11
Scaled unit ratios and `overrideManifest`

michaelfig:
## Description

Fixes a hardcoded price scaling ratio to be calculated for the actual brands.  Also, to prepare for installing just this change on devnet, provide an `overrideManifest` in the governance proposals to make the manifest explicit in the core proposal.

### Security Considerations

Improved legibility of core proposals generated by writeCoreProposal and extractCoreProposalBundles.

None.

### Testing Considerations

Tested by run-protocol governance unit tests and chain integration testing, as well as manually inspecting a generated core-proposal.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/addAssetToVault.js
+++ b/packages/run-protocol/src/vaultFactory/addAssetToVault.js
@@ -87,10 +87,37 @@ export const registerScaledPriceAuthority = async ({
     ]),
   ]);
 
-  // we used 4 decimals for ibcAtom above...
-  // 1_000_000 atomBrand = 10_000 ibcAtomBrand
-  const scaleIn = makeRatio(100n, atomBrand, 1n, ibcAtomBrand);
-  const scaleOut = makeRatio(1n, usdBrand, 1n, runBrand);
+  // We need "unit amounts" of each brand in order to get the ratios right.  You
+  // can ignore decimalPlaces when adding and subtracting a brand with itself,
+  // but not when creating ratios.
+  const getDecimalP = async brand => {
+    const displayInfo = E(brand).getDisplayInfo();
+    return E.get(displayInfo).decimalPlaces;
+  };
+  const [
+    decimalPlacesAtom = 0,
+    decimalPlacesIbcAtom = 0,
+    decimalPlacesUsd = 0,
+    decimalPlacesRun = 0,
+  ] = await Promise.all([
+    getDecimalP(atomBrand),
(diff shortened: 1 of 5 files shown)


dckc (approved):
Yay! manifest in the proposal!

```js
const manifestInstallRef = "board05669";
const getManifestCall = harden([
  "getManifestForInviteCommittee",
  {
    econCommitteeCharterRef: "board02568",
    voterAddresses: {
      Dan: "agoric1j5sm4kr2mdae6sfey6qjm9vaamddmp0yuxs6kz",
      Michael: "agoric1y5gpzqd876gqmkyl2c5q66vp6e30se4vwn644v",
      Rowland: "agoric14nt7n99jstnk7zzgmw4hzhc05yqjplerrv9uqd",
    },
  },
]);
const overrideManifest = {
  inviteCommitteeMembers: {
    consume: {
      agoricNames: true,
...
```

Calling it an override rather than the normal case doesn't seem ideal, but I can see how it's expedient. Works for me.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -87,10 +87,37 @@ export const registerScaledPriceAuthority = async ({
...
 
-  // we used 4 decimals for ibcAtom above...
-  // 1_000_000 atomBrand = 10_000 ibcAtomBrand
-  const scaleIn = makeRatio(100n, atomBrand, 1n, ibcAtomBrand);
-  const scaleOut = makeRatio(1n, usdBrand, 1n, runBrand);
+  // We need "unit amounts" of each brand in order to get the ratios right.  You
+  // can ignore decimalPlaces when adding and subtracting a brand with itself,
+  // but not when creating ratios.
+  const getDecimalP = async brand => {
+    const displayInfo = E(brand).getDisplayInfo();
```
turadg, line 94: we're veering more from the design that these are merely display info.

if they are truly display info, then couldn't we get this info from somewhere other than the brand? 

if they aren't merely display info, we should revise the design. Not in this PR but if you agree I can make a ticket to follow up.

dckc: I agree that it's worth getting wider agreement on whether this justification is sound. A ticket seems reasonable; perhaps we can talk about it this afternoon with Zoe stuff.

turadg: https://github.com/Agoric/agoric-sdk/issues/5348

PR #5343, opened 2022-05-11
feat(vault): liquidation penalty handled by liquidation contracts

### LINKED ISSUE #5167: Liquidation penalty should be handled in the liquidation contract (not vaults)
Different liquidation strategies may require different handling of the liquidation penalty (e.g., to pass it off to third party liquidators in a different model).  Per Dean, this is currently handled in vaults.

Why: so liquidation strategy can be governed, including how penalties are charged and where they go.

**Acceptance criteria**
- liquidation contract encodes penalty charging and where it goes
- reserve public facet gets passed to liquidation contracts
- no vault test numbers are changed (liquidateMinimum gets it too)
- no more penaltyPool seat
- PenaltyRate remains per collateral type (leave where it is for now and pass in upon liquidation)
- Now or later: liquidateIncrementally pushes penalties into the reserve

turadg:
closes: #5167 

## Description

Moves the penalty _charging_ concern from Vault Factory contract to the liquidation contracts. It's up to them what to do with the `penaltyRate` input.

This also removes the `penaltyPoolSeat` since it's raison d'être n'est plus.

Next up is https://github.com/agoric/agoric-sdk/issues/5367 

### Security Considerations

Empowers the liquidation contracts to charge _or not charge_ the liquidation penalty specified by `penaltyRate`.

Provides them a public facet of the asset reserve.

--

### Testing Considerations

Updated test setup to accommodate wiring of the `reserve` instance. The test of liquidation penalty gathering (removed here) will be part of #5367.


packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -1,14 +1,18 @@
...
+  offerTo,
+} from '@agoric/zoe/src/contractSupport/index.js';
 import { AmountMath } from '@agoric/ertp';
 import { Far } from '@endo/marshal';
 
 import { makeTracer } from '../makeTracer.js';
 
 const trace = makeTracer('LiqMin', false);
 
+// ??? is this contract used only for tests and demo purposes?
```
Chris-Hibbert: currently only used that way, but also available to governance to fall back to.

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -81,33 +59,24 @@ const liquidate = async (
...
-
+  trace(` offeredTo`, { collateralToSell, debt });
+
+  // await deposited and offer result, but ignore the latter
+  const [proceeds] = await Promise.all([
+    deposited,
+    E(liqSeat).getOfferResult(),
+  ]);
+  // all the proceeds from AMM sale are on the vault seat instead of a staging seat
+  // ??? ^ okay?
```
Chris-Hibbert: Why drop `penaltyPoolSeat`? That seemed like a good place from which to stage funds going to the reserve.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -460,9 +460,9 @@ test('first', async t => {
     'unused collateral remains after liquidation',
   );
 
-  t.deepEqual(await E(vaultFactory).getPenaltyAllocation(), {
-    RUN: AmountMath.make(runBrand, 30n),
-  });
+  // ??? remove penalty pool from vault factory
+  // used to be funded from liquidation penalty
+  t.deepEqual(await E(vaultFactory).getPenaltyAllocation(), {});
```
Chris-Hibbert: I think it's correct to take the penalty funds out of the vault, but then they have to be accessible (even if we aren't yet retrieving them) from the liquidation contracts.

Chris-Hibbert (changes requested):

packages/run-protocol/src/vaultFactory/liquidation.js
turadg: The requirements of the ticket are that the contract manages where funds go.

They each have their own `penaltyPoolSeat` now. I still have to wire up moving the assets out to the reserve.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Yes, but I'm not sure that `test-vaultFactory` needs to be concerned with that. It's a concern of the liquidation contracts what they do with the `liquidationPenalty` argument.

turadg: I see I misread your comment. Agree completely and working on it.

packages/run-protocol/src/econ-behaviors.js
```diff
@@ -241,7 +241,8 @@ export const startVaultFactory = async (
   {
     consume: {
       chainTimerService,
-      priceAuthority,
+      priceAuthority: priceAuthorityP,
+      reservePublicFacet: reservePublicFacetP,
```
dckc: odd... usually we don't bother putting public facets in the produce / consume space, since we can get the instances via agoricNames and then the publicFacet via zoe

But if that's outside the scope of this PR...

to permit more stuff to be consumed, add it to https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/core-proposal.js#L40-L46

turadg: thanks! [842b068](https://github.com/Agoric/agoric-sdk/pull/5343/commits/842b0687dc20107814831991d9ca76b6b59dfc64)

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -39,31 +51,44 @@ const start = async zcf => {
...
     debtorSeat.exit();
   };
 
   /**
    * @type {ERef<Liquidator>}
    */
-  const creatorFacet = Far('debtorInvitationCreator', {
-    makeLiquidateInvitation: () => zcf.makeInvitation(debtorHook, 'Liquidate'),
+  const creatorFacet = Far('debtorInvitationCreator (minimum)', {
+    makeLiquidateInvitation: () =>
```
turadg, line 90: renamed à la https://github.com/Agoric/agoric-sdk/pull/5179

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -3,30 +3,11 @@
...
 import { makeTracer } from '../makeTracer.js';
 
 const trace = makeTracer('LIQ');
 
-/**
- * @param {Amount<'nat'>} proceeds
- * @param {Amount<'nat'>} debt - after incurring penalty
- * @param {Amount<'nat'>} penaltyPortion
- */
-const partitionProceeds = (proceeds, debt, penaltyPortion) => {
```
turadg, line 20: now that the math is split between two contracts, not worth having this combo fn

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -242,16 +252,28 @@ const start = async zcf => {
...
     assert(
       originalDebt.brand === debtBrand,
       X`Cannot liquidate to ${originalDebt.brand}`,
     );
-    for await (const t of processTranches(debtorSeat, originalDebt)) {
+    const penalty = ceilMultiplyBy(originalDebt, penaltyRate);
+    const debtWithPenalty = AmountMath.add(originalDebt, penalty);
+    trace('LIQ', { originalDebt, debtWithPenalty });
+
+    for await (const t of processTranches(debtorSeat, debtWithPenalty)) {
```
Chris-Hibbert, line 276: Good catch!  Did changing to `debtWithPenalty` break any tests?

I don't see any related changes, so if not, would you mind adding one that would have detected the difference?

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -276,15 +298,28 @@ const start = async zcf => {
         });
       }
     }
+
+    // Now we need to know how much was sold so we can pay off the debt.
+    // We can use this because only liquidation adds debt brand to the seat.
+    const debtPaid = debtorSeat.getAmountAllocated('Out', debtBrand);
+    const penaltyPaid = AmountMath.min(penalty, debtPaid);
+
+    // Allocate penalty portion of proceeds to a seat that will be transferred to reserve
```
Chris-Hibbert: It's not the seat that will be transferred...
```suggestion
    // Allocate penalty portion of proceeds to a seat that will hold it for transfer to reserve
```

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -276,15 +298,28 @@ const start = async zcf => {
         });
       }
     }
+
+    // Now we need to know how much was sold so we can pay off the debt.
+    // We can use this because only liquidation adds debt brand to the seat.
```
```suggestion
    // We can use this seat because only liquidation adds debt brand to it
```

packages/run-protocol/src/vaultFactory/liquidateMinimum.js
```diff
@@ -39,31 +51,44 @@ const start = async zcf => {
...
     trace(`Liq results`, {
-      debt,
+      debtWithPenalty,
       amountIn,
       paid: debtorSeat.getCurrentAllocation(),
       amounts,
     });
+
+    // Now we need to know how much was sold so we can pay off the debt.
+    // We can use this because only liquidation adds debt brand to the seat.
```
Chris-Hibbert:
```suggestion
    // We can use this seat because only liquidation adds debt brand to it
```

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -81,33 +57,24 @@ const liquidate = async (
...
-
+  trace(` offeredTo`, { collateralToSell, debt });
+
+  // await deposited and offer result, but ignore the latter
+  const [proceeds] = await Promise.all([
+    deposited,
+    E(liqSeat).getOfferResult(),
+  ]);
+  // all the proceeds from AMM sale are on the vault seat instead of a staging seat
+  // ??? ^ okay?
```
Chris-Hibbert: Yes, it's fine that all the proceeds end up on the VaultSeat. You can drop the `???` comment.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -460,9 +473,6 @@ test('first', async t => {
     'unused collateral remains after liquidation',
   );
 
-  t.deepEqual(await E(vaultFactory).getPenaltyAllocation(), {
-    RUN: AmountMath.make(runBrand, 30n),
```
Chris-Hibbert, line 464: I expected the new PenaltySeats to have accessors like this.  Why don't they?

Chris-Hibbert (changes requested):
This is basically right, but I have a few questions and suggestions.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: The plan is not to expose the penalty seat. @dtribble said not to use the fee collector pattern here and instead have the liquidation contract move the assets directly to the reserve. To give input on that: https://github.com/Agoric/agoric-sdk/issues/5367

packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
Chris-Hibbert: As you mentioned, the behavior doesn't change here, the debt calculation just moved to the callee rather than the caller.

Chris-Hibbert (approved):

PR #5357, opened 2022-05-12
fix: deployment job should capture results on failure

mhofman:
## Description
 
#5301 introduced a bug where the results of failing deployment tests are no longer processed. This fixes that.


michaelfig (approved):
LGTM!

PR #5360, opened 2022-05-13
fix(deployment): Do not override explicit faucet SOLO_COINS

mhofman:
## Description

Apparently the vault tasks of the loadgen have been failing in CI since https://github.com/Agoric/agoric-sdk/pull/5235 landed with the following error: `The request must be for at least (a bigint). (a bigint) is too small` It doesn't fail when running a loadgen locally, which indicates that the issue is specific to the setup of a testnet setup vs a local chain. Given the name of the PR, I figured this had something to do with the amount of RUN given to a solo through the faucet, which has been drastically reduced as a side effect of #5190 

### Security Considerations

None

Added a comment indicating we need to use existing value if any

### Testing Considerations

I will look into requiring that all loadgen tests succeed, for the CI test to succeed to avoid similar issues in the future.


michaelfig (approved):
Ship it!

PR #5362, opened 2022-05-13
Decentral economy config

michaelfig:
## Description

This creates `agoric-sdk/packages/vats/decentral-economy-config.json` which boots up a lot more of the economy than its predecessors.  It takes advantage of arguments to the `defaultCoreProposal` function.

I am requesting review and intend to merge, even though there are still some manual governance pieces to perform, namely:

- The oracle governance currently needs to execute in an ag-solo environment.  Instagoric should run it from the ag-solo to produce the oracleAdmin facet that can be used to push prices.  It would make sense to have the (single member) econ committee be just that privileged ag-solo instance.
- Once the oracle is sorted out, the economic committee will have to vote to raise the vault debt ceiling for that collateral.

### Testing Considerations

Tested with local-chain as well as sim-chain.

### DIFF
--- a/golang/cosmos/x/vbank/vbank.go
+++ b/golang/cosmos/x/vbank/vbank.go
@@ -143,6 +143,9 @@ func (ch portHandler) Receive(ctx *vm.ControllerContext, str string) (ret string
 		if err != nil {
 			return "", fmt.Errorf("cannot convert %s to address: %w", msg.Address, err)
 		}
+		if err = sdk.ValidateDenom(msg.Denom); err != nil {
+			return "", fmt.Errorf("invalid denom %s: %w", msg.Denom, err)
+		}
 		coin := keeper.GetBalance(ctx.Context, addr, msg.Denom)
 		packet := coin.Amount.String()
 		if err == nil {
@@ -156,6 +159,9 @@ func (ch portHandler) Receive(ctx *vm.ControllerContext, str string) (ret string
 		addr, err := sdk.AccAddressFromBech32(msg.Sender)
 		if err != nil {
 			return "", fmt.Errorf("cannot convert %s to address: %w", msg.Sender, err)
+		}
+		if err = sdk.ValidateDenom(msg.Denom); err != nil {
+			return "", fmt.Errorf("invalid denom %s: %w", msg.Denom, err)
 		}
 		value, ok := sdk.NewIntFromString(msg.Amount)
 		if !ok {
@@ -181,6 +187,9 @@ func (ch portHandler) Receive(ctx *vm.ControllerContext, str string) (ret string
 		addr, err := sdk.AccAddressFromBech32(msg.Recipient)
 		if err != nil {
 			return "", fmt.Errorf("cannot convert %s to address: %w", msg.Recipient, err)
+		}
+		if err = sdk.ValidateDenom(msg.Denom); err != nil {
+			return "", fmt.Errorf("invalid denom %s: %w", msg.Denom, err)
 		}
 		value, ok := sdk.NewIntFromString(msg.Amount)
 		if !ok {
--- a/packages/deploy-script-support/src/coreProposalBehavior.js
+++ b/packages/deploy-script-support/src/coreProposalBehavior.js
@@ -113,7 +113,7 @@ export const makeCoreProposalBehavior = ({
 
 export const makeEnactCoreProposalsFromBundleCap =
   ({ makeCoreProposalArgs, E }) =>
-  allPowers => {
+  async allPowers => {
     const {
       vatPowers: { D },
       devices: { vatAdmin },
@@ -127,7 +127,7 @@ export const makeEnactCoreProposalsFromBundleCap =
       return install;
     };
 
-    return Promise.all(
+    await Promise.all(
       makeCoreProposalArgs.map(async ({ ref, call, overrideManifest }) => {
         const subBehavior = makeCoreProposalBehavior({
           manifestInstallRef: ref,
--- a/packages/deploy-script-support/src/extract-proposal.js
+++ b/packages/deploy-script-support/src/extract-proposal.js
@@ -68,12 +68,43 @@ export const extractCoreProposalBundles = async (
 
   const bundleToSource = new Map();
   const extracted = await Promise.all(
-    coreProposals.map(async (initCore, i) => {
-      console.log(`Parsing core proposal:`, initCore);
+    coreProposals.map(async (coreProposal, i) => {
+      console.log(`Parsing core proposal:`, coreProposal);
+
+      /** @type {string} */
+      let entrypoint;
+      /** @type {unknown[]} */
+      let args;
+      /** @type {string} */
+      let module;
+      if (typeof coreProposal === 'string') {
+        module = coreProposal;
+        entrypoint = 'defaultProposalBuilder';
+        args = [];
+      } else {
+        ({
+          module,
+          entrypoint = 'defaultProposalBuilder',
+          args = [],
+        } = coreProposal);
+      }
+
+      assert.typeof(
+        module,
+        'string',
+        X`coreProposal module ${module} must be string`,
+      );
+      assert.typeof(
+        entrypoint,
+        'string',
+        X`coreProposal entrypoint ${entrypoint} must be string`,
+      );
+      assert(Array.isArray(args), X`coreProposal args ${args} must be array`);
+
       const thisProposalBundleHandles = new Set();
       assert(getSequenceForProposal);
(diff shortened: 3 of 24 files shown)


golang/cosmos/x/vbank/vbank.go
```diff
@@ -143,6 +143,9 @@ func (ch portHandler) Receive(ctx *vm.ControllerContext, str string) (ret string
 		if err != nil {
 			return "", fmt.Errorf("cannot convert %s to address: %w", msg.Address, err)
 		}
+		if err = sdk.ValidateDenom(msg.Denom); err != nil {
+			return "", fmt.Errorf("invalid denom %s: %w", msg.Denom, err)
```
dckc, line 147: How did you discover this bug? Did some other part of this PR trip it?

packages/deploy-script-support/src/coreProposalBehavior.js
```diff
@@ -127,16 +127,15 @@ export const makeEnactCoreProposalsFromBundleCap =
...
-          manifestInstallRef: ref,
-          getManifestCall: call,
-          overrideManifest,
-          E,
-          restoreRef,
-        });
-        return subBehavior(allPowers);
-      }),
-    );
+    // Run linearly to prevent governance interdependencies.
```
dckc: Did you run into an example of interference? What was it?

Fixes should come with tests. Is writing a test so costly that it's not worthwhile in this case?

dckc: Or perhaps the changes to `core-proposal.js` are tested (in the solo test-home.js?) and depend on this fix, so while we don't have a unit test for just this fix, we do run this code each ci?

packages/deploy-script-support/src/extract-proposal.js
```diff
@@ -68,12 +68,43 @@ export const extractCoreProposalBundles = async (
 
   const bundleToSource = new Map();
   const extracted = await Promise.all(
-    coreProposals.map(async (initCore, i) => {
-      console.log(`Parsing core proposal:`, initCore);
+    coreProposals.map(async (coreProposal, i) => {
```
dckc, line 71: features should have tests too... but again, I suppose the tests for the changes to `core-proposal.js` indirectly test this.

packages/run-protocol/scripts/add-collateral-core.js
```diff
@@ -5,17 +5,30 @@ import { makeHelpers } from '@agoric/deploy-script-support';
...
+export const defaultProposalBuilder = async (
+  { publishRef, install },
+  options = {},
+) => {
+  const {
+    interchainOracle = 'ATOM',
+    interchainDenom = process.env.INTERCHAIN_DENOM,
+    interchainDecimals = 6,
+    interchainKeyword = 'IbcATOM',
+    interchainProposedName = interchainOracle,
```
dckc: whoa... the scoping in default arguments is wild.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -13,7 +13,13 @@ const reserveThenGetNames = async (nameAdmin, names) => {
 
 /**
  * @param { EconomyBootstrapPowers } powers
- * @param {*} config
+ * @param {object} config
```
dckc, line 25: tangentially... is it a feature or a bug that this governance script now lives in the same directory as the smart contract? It was originally developed in hack-and-slash mode. Putting it here makes it more likely to be considered "in scope" for security reviews. I suppose PRs such as this one are polishing it in due course.

cc @arirubinstein @jessysaurusrex @turadg @dtribble @Chris-Hibbert

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -25,18 +31,22 @@ export const addInterchainAsset = async (
   },
   {
     options: {
-      denom,
-      keyword = 'IbcATOM',
-      proposedName = 'ATOM',
-      decimalPlaces = 4,
-    },
+      interchainOracle = 'ATOM',
```
dckc: managing the scope of these option names seems a bit tricky. Do we document the things they need to not collide with anywhere?

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -124,36 +146,52 @@ export const registerScaledPriceAuthority = async ({
...
-export const addAssetToVault = async ({
-  consume: { vaultFactoryCreator, reserveCreatorFacet, agoricNamesAdmin, zoe },
-  brand: {
-    consume: { RUN: runP },
+/**
+ * @param {EconomyBootstrapPowers} powers
+ * @param {object} config
+ * @param {object} [config.options]
+ * @param {string} [config.options.interchainKeyword]
+ * @param {string} [config.options.interchainOracle]
```
dckc: add some prose docs for these?

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -124,36 +146,52 @@ export const registerScaledPriceAuthority = async ({
...
+    },
+    brand: {
+      consume: { RUN: runP },
+    },
+    instance: {
+      consume: { amm },
+    },
   },
-  instance: {
-    consume: { amm },
```
dckc: yikes! how did that ever work before?!

packages/run-protocol/src/vaultFactory/addAssetToVault.js
```diff
@@ -165,11 +203,17 @@ export const addAssetToVault = async ({
...
+    interchainDecimals,
+    interchainKeyword,
+    interchainProposedName,
+    scaledPriceAuthorityRef,
+  },
 ) => {
   return {
     manifest: {
-      addInterchainAsset: {
+      [addInterchainAsset.name]: {
```
dckc, line 211: So you have made your peace with this idiom after all?

dckc (commented):
I don't think I found any critical problems, but I'd like to discuss the level of documentation and perhaps test coverage before I approve.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
turadg: Agree with the principle that `run-protocol/src` contains logic for the contracts as they run, not how they're deployed. The value in keeping them separated is both for security review and for developers looking to this source for writing contracts and re-using these.

Thankfully this is used only for `scripts/add-collateral-core.js` so it would be easy as moving it over there. Longer term would it make sense to have another package (`@agoric/inter-protocol-vats`?) for the vat setup concern?

packages/run-protocol/src/vaultFactory/addAssetToVault.js
michaelfig: It's better than a string, even if it's still a little fragile.

golang/cosmos/x/vbank/vbank.go
michaelfig: Yeah, my crazy `@INTERCHAIN_DENOM@` template variable caused a Golang panic.  Drive-by fix for all the similar places.

packages/deploy-script-support/src/coreProposalBehavior.js
michaelfig: I may have fixed the issue now by reserving the econ committee namesByAddress.  I'll try reverting this code to be sure.

packages/deploy-script-support/src/extract-proposal.js
michaelfig: I've been testing manually, but will automate if you have a suggestion how.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
michaelfig: I've bundled them in `options.interchainAssetOptions` instead.  That seems to be the right kind of namespacing.

packages/run-protocol/src/vaultFactory/addAssetToVault.js
michaelfig: I created a separate type that should make this self-documenting.

packages/deploy-script-support/src/extract-proposal.js
dckc: I was thinking that `test-home.js` tested `core-proposal.js` already.

dckc (approved):
Ideally we'd have more automated / regression tests, but I reproduced your results, and `test-gov-collateral.js` exercises much of this, and the cost of more automated tests looks pretty high.

So let's go for it.

PR #5363, opened 2022-05-13
fix(swingset): move upgrade() vatParameters into options bag

warner:
Previously, the parent vat did `E(adminFacet).upgrade(vatParameters)`

Now it does `E(adminFacet).upgrade(bundleCap, { vatParameters })`

This matches the way `vatParameters` are passed to the original
`createVat()` call.

All other options are rejected.

This changes the `adminFacet` API, but internally continues to send
`vatParameters` in their own argument to the underlying device
nodes. If/when we send actual options, we'll need to decide whether to
change the shape of the device node API by converting `vatParameters`
into `options`, or just add additional arguments.

closes #5435

### DIFF
--- a/packages/SwingSet/docs/vat-upgrade.md
+++ b/packages/SwingSet/docs/vat-upgrade.md
@@ -1,5 +1,3 @@
-(NOTE: none of this is implemented yet)
-
 # Vat Upgrade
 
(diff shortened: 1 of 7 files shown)


warner:
oops, wrong branch name

PR #5365, opened 2022-05-13
fix(swingset): move upgrade() vatParameters into options bag

### LINKED ISSUE #5345: vat upgrade should put vatParameters in an option bag

Vats are created with `vatAdminService~.createVat(bundleCap, options)`, where `options` includes `{ vatParameters }` and others.

Vats are upgraded with `adminFacet~.upgrade(newBundleCap, newVatParameters)`.

We should change this to be `upgrade(newBundleCap, { vatParameters })`, for parallelism with the original creation API, and to start an options bag for later.

Note that we don't currently have any other options. Many of the options available during initial creation are not changable by upgrade (e.g. the `managerType` is fixed at creation time). I don't currently think any of those other options are worth changing during upgrade, or it would be difficult to allow them to be changed. But `vatParameters` is how we'll give a new contract bundlecap to ZCF, so it's important to be replaceable.

## Description of the Design

Change the function signature.

## Security Considerations

None

## Test Plan

Update the existing unit tests

warner:
Previously, the parent vat did `E(adminFacet).upgrade(vatParameters)`

Now it does `E(adminFacet).upgrade(bundleCap, { vatParameters })`

This matches the way `vatParameters` are passed to the original
`createVat()` call.

All other options are rejected.

This changes the `adminFacet` API, but internally continues to send
`vatParameters` in their own argument to the underlying device
nodes. If/when we send actual options, we'll need to decide whether to
change the shape of the device node API by converting `vatParameters`
into `options`, or just add additional arguments.

closes #5345

### DIFF
--- a/packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
+++ b/packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
@@ -98,7 +98,13 @@ export function buildRootObject(vatPowers) {
       done() {
         return doneP;
       },
-      upgrade(bundlecap, vatParameters) {
+      upgrade(bundlecap, options = {}) {
+        const { vatParameters, ...rest } = options;
+        const leftovers = Object.keys(rest);
+        if (leftovers.length) {
+          const bad = leftovers.join(',');
+          assert.fail(`upgrade() received unknown options: ${bad}`);
+        }
         let bundleID;
         try {
           bundleID = D(bundlecap).getBundleID();
--- a/packages/SwingSet/docs/vat-upgrade.md
+++ b/packages/SwingSet/docs/vat-upgrade.md
@@ -1,5 +1,3 @@
-(NOTE: none of this is implemented yet)
-
 # Vat Upgrade
 
 Dynamic vats can be upgraded to use a new code bundle. This might be used to fix a bug, to add new functionality, or merely to delete accumulated state and reduce memory usage.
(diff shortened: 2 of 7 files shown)


packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -98,7 +98,13 @@ export function buildRootObject(vatPowers) {
       done() {
         return doneP;
       },
-      upgrade(bundlecap, vatParameters) {
+      upgrade(bundlecap, options) {
```
FUDCo: `options` should default to `{}`, so existing code that isn't using it don't need to change.

FUDCo (changes requested):
One minor thing I really think should be tweak, but other than that it looks good to me.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: Hmm.  On second thought I'm not sure how downstream code will handle a missing `vatParameters` parameter.  I had been thinking you could get away with leaving it off if you didn't have any parameters, and maybe you could have.  If so, you should still be able to do that.  If not, maybe you should.

warner: Good catch!

I guess omitting `vatParameters` in the upgrade should get you the same value as you'd get omitting it in the original `createVat`. That *ought* to be `undefined`, but now that I'm looking at the code, I'm not sure that's what happens. I'll add a test.

FUDCo (approved):
は、私にはよく見えますよ

Issue #4374, opened 2022-01-25
handle named bundles in kernel config and core-bootstrap


warner:
## What is the Problem Being Solved?

Once we have bundlecaps (#4372), we'd like the new #4165 / #4247 "core bootstrap" scheme to use them too. Currently this relies upon `vatAdminService~.createVatByName(name)`, rather than a bundlecap or bundleID.

That means changing the kernel config format to somehow handle named bundles properly. We want vats to uniformly defined by a bundleID, so that the future upgrade protocol has a well-defined initial value to upgrade from.

The old code processes a `config.bundles` section which maps bundle names to either a filename (from which `bundleSource` can create a bundle) or an actual bundle (used only by tests). It then builds a table of named bundles, which `vatAdminService~.createVatByName()` can reference. @dckc 's 

The new approach should process `config.bundles` and install each bundle, then build a table mapping name to bundleID. The `bootstrap()` message should be enhanced to provide this table to the bootstrap vat (perhaps through a third argument, so `bootstrap(vats, devices, bundles)`), which can use `devices.bundle` to get bundlecaps, and send them to `vatAdminService~.createVat(bundlecap)`.

When we're done, we can remove `createVatByName`, and `createVat()` (which exclusively take a bundlecap) will be the only remaining API.

warner:
#4372 takes `config.bundles` and installs each one, so bootstrap can do `bundlecap = D(devices.bundle).getNamedBundleCap(name)` and pass that through to whatever does `createVat()`. I decided to avoid the third `bootstrap()` argument.

For now, `vatAdminService~.createVatByName()` still works, so I think I can wait for the new bootstrap code to land before thinking about how to change it to use bundlecaps.

warner:
The new bootstrap code landed a while ago. It [currently](https://github.com/Agoric/agoric-sdk/blob/70fa3908dde52a79658b5ef8f58c17406a030ed3/packages/vats/src/core/basic-behaviors.js#L230-L232) grabs the whole bundle and passes it to the old `E(zoe).install(bundle)` API:

```js
 const bundleCap = D(vatAdmin).getNamedBundleCap(name); 
 const bundle = D(bundleCap).getBundle(); 
 producer.resolve(E(zoe).install(bundle)); 
```

To avoid holding the whole bundle, we need to give bootstrap a way to get from a name to a bundleID, so it can use `E(zoe).installBundleID(id)` instead of `install(bundle)`.

[...]

PR #5368, opened 2022-05-13
feat(swingset): add vatAdminService.getBundleIDByName()

warner:
This API takes the name of a statically-configured bundle (i.e. a
property name of `config.bundles`) and returns its BundleID (a hash
string). This can be used later in `E(zoe).installBundleID(id)`.

refs #4374 , specifically this API will allow bootstrap to deal with
bundleIDs and bundleCaps rather than bundles.

### DIFF
--- a/packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
+++ b/packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
@@ -257,6 +257,20 @@ export function buildDevice(tools, endowments) {
           }
           return returnCapForBundleID(bundleID);
         }
+        // D(devices.bundle).getBundleIDByName(name) -> id
+        if (method === 'getBundleIDByName') {
+          const args = unserialize(argsCapdata);
+          const [name] = args;
+          assert.typeof(name, 'string', `getBundleIDByName() name`);
+          let bundleID;
+          try {
+            // this throws on a bad name, so make a better error
+            bundleID = getNamedBundleID(name);
+          } catch (e) {
+            throw Error(`unregistered bundle name '${name}'`);
+          }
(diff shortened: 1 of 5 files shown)


packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
```diff
@@ -257,6 +257,20 @@ export function buildDevice(tools, endowments) {
           }
           return returnCapForBundleID(bundleID);
         }
+        // D(devices.bundle).getBundleIDByName(name) -> id
+        if (method === 'getBundleIDByName') {
+          const args = unserialize(argsCapdata);
+          const [name] = args;
+          assert.typeof(name, 'string', `getBundleIDByName() name`);
```
dckc, line 264: I take it kernel code doesn't use `` `X`...` `` for error message details?

dckc (commented):
Looks plausible... fine for my needs.

I leave it to someone else to approve; I don't think I'm sufficiently swapped in on kernel stuff.

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
warner: I generally find `X`/`details` to impede debugging, and I value that more than the possible information leakage when caller A calls B, B does `getBundleIDByName(secretName)`, and A catches the error to learn the name (none of which matters on a transparent chain anyways).

At some point we should probably have a collective discussion on the utility of that argument censorship (and more generally to find a coding style that we're all comfortable with and make the kernel use it, since we in kernel land have not been following all parts of the general agoric-sdk style, i.e. named `=>` functions).

FUDCo (approved):
Credible.

PR #5370, opened 2022-05-14
refactor(swingset): Reserve host prefix in the store

### LINKED ISSUE #5349: Reserve a section of the swingstore for the host

Currently host cosmic-swingset uses the key `cosmos/meta` of the swingstore to save it's own state just before committing. We'd like to expand what is currently stored by the host, and keep it more structured.

supports #5350

## Description of the Design

Reserve the prefix `host.` for host operations. Swingset should not attempt to read or write in those values.

## Security Considerations

It's probably sufficient to have this simply a documented contract between the host and Swingset, without explicit enforcement. The writes performed by the host are already not part of the crank buffer or activityhash computation performed by swingset, but we could add a check there for good measure.

## Test Plan
TBD

mhofman:
closes: #5349

## Description

Designate a section of the swingstore prefixed by `host.` for `cosmic-swingset`'s usage.

### Security Considerations

None

Should there be?

### Testing Considerations

Added assertion so Swingset will never write to these keys.
Added a unit test verifying that assertion.

### DIFF
--- a/packages/cosmic-swingset/src/launch-chain.js
+++ b/packages/cosmic-swingset/src/launch-chain.js
@@ -31,7 +31,12 @@ import {
 
 const console = anylogger('launch-chain');
 
-const SWING_STORE_HOST_PREFIX = 'host';
+/**
+ * Return the key in the reserved "host.*" section of the swing-store
+ *
+ * @param {string} path
+ */
+const getHostKey = path => `host.${path}`;
 
 async function buildSwingset(
   mailboxStorage,
@@ -266,12 +271,9 @@ export async function launch({
     savedBlockTime,
     savedChainSends,
   ) {
-    kvStore.set(`${SWING_STORE_HOST_PREFIX}.height`, `${savedHeight}`);
-    kvStore.set(`${SWING_STORE_HOST_PREFIX}.blockTime`, `${savedBlockTime}`);
-    kvStore.set(
-      `${SWING_STORE_HOST_PREFIX}.chainSends`,
-      JSON.stringify(savedChainSends),
-    );
+    kvStore.set(getHostKey('height'), `${savedHeight}`);
+    kvStore.set(getHostKey('blockTime'), `${savedBlockTime}`);
+    kvStore.set(getHostKey('chainSends'), JSON.stringify(savedChainSends));
     await commit();
   }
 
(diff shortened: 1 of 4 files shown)


michaelfig (commented):
LGTM!  I'd like to defer to @warner unless you've already gotten a conceptual ack from him.

mhofman:
Yeah sorry I should have clarified. I wanted to make sure the "meta" split logic was ok with @michaelfig and that the concept of reserving `host.` and the new check in the Swingset store wrapper was ok for @warner

michaelfig (approved):

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -31,7 +31,7 @@ import {
 
 const console = anylogger('launch-chain');
 
-const SWING_STORE_META_KEY = 'cosmos/meta';
+const SWING_STORE_HOST_PREFIX = 'host';
```
warner: if we do it this way, the prefix should be `host.` , and we should prohibit the kernel from writing to that prefix

mhofman:
@warner ignoring the queue length policy we discussed for the next step, does this PR stand on its own as making explicit a behavior that was already occurring?

warner:
Yep, this is OK on its own, I was just looking out for confusion about keys like `hosting` that overlap with `SWING_STORE_HOST_PREFIX` but are really in a different key space. Maybe just add a comment to `SWING_STORE_HOST_PREFIX = 'host'` to say that keys like `host.*` are reserved for the cosmic-swingset host app. The check on the kernel side is correct (`host.` not `host`). 

(I'd originally wanted `SWING_STORE_HOST_PREFIX = 'host.'`, and to change the other uses there to be like `${SWING_STORE_HOST_PREFIX}foo` instead of `.foo`, but that looks too much like a typo)

warner (approved):

mhofman:
> (I'd originally wanted `SWING_STORE_HOST_PREFIX = 'host.'`, and to change the other uses there to be like `${SWING_STORE_HOST_PREFIX}foo` instead of `.foo`, but that looks too much like a typo)

Yeah I originally had it that way but it looked wrong, and too prone to errors. The alternative would be to replace the `SWING_STORE_HOST_PREFIX` const with a helper which generates the full key:

```js
const getHostKey = suffix => `host.${suffix}`;
```

mhofman:
> The alternative would be to replace the `SWING_STORE_HOST_PREFIX` const with a helper which generates the full key:

@warner I actually did just that, and prefer it.

PR #5371, opened 2022-05-14
fix(solo): introduce runStake attMaker to wallet admin facet

### LINKED ISSUE #4263: Make the wallet support getRUN / BLD Boost

dckc:
closes #4263

## Description

This addresses a gap in the runStake user flows where a
user would have to use the REPL to do:

 E(E(home.wallet).getAdminFacet()).resolveAttMaker(home.attMaker)

We use setPresences to detect the arrival of attMaker on the chain
bundle (decentralObjects) and setWallet to detect the arrival of the
wallet. When both are available, we introduce them with
resolveAttMaker.

### Security Considerations

It's by design that a user's wallet has the power to place liens on their stake.

reduces the need for special "now do this in the REPL" step for runStake.

### Testing Considerations

@samsiegart please verify that this works with run.agoric.net.

I had some trouble getting the synchronization right. @michaelfig and I found that by adding an await ... somewhere... (in `registerWallet()`?) it would work. Then I stashed it and came back to it, and that await was gone, but this didn't fail like it seemed to earlier.


packages/solo/src/vat-http.js
```diff
@@ -97,6 +104,14 @@ export function buildRootObject(vatPowers) {
     urlToHandler.set(url, commandHandler);
   }
 
+  const [resolveWallet, resolveAttMaker] = introduceWith(
+    async (wallet, attMaker) => {
+      const walletAdmin = await E(wallet).getAdminFacet();
+      // console.debug('introduce', { wallet, walletAdmin, attMaker });
+      E(walletAdmin).resolveAttMaker(attMaker);
```
samsiegart: Ah didn't expect this method to be useful still, cool!

packages/solo/src/vat-http.js
```diff
@@ -3,8 +3,15 @@ import { E } from '@endo/eventual-send';
 import { Far } from '@endo/marshal';
 import { assert, details as X } from '@agoric/assert';
 import { getReplHandler } from '@agoric/vats/src/repl.js';
+import { makePromiseKit } from '@endo/promise-kit';
 import { getCapTPHandler } from './captp.js';
 
+const introduceWith = f => {
```
samsiegart: Nit: This function seems a little mysterious on its own, maybe name it something more specific to attMaker or just do it inline.

samsiegart (approved):

packages/solo/src/vat-http.js
dckc: I was hoping / planning to not use it, but I couldn't figure out an alternative.

samsiegart:
I just saw that I need to test this manually, will do so then re-review

samsiegart (approved):
I tested this with a local chain and it seemed to work

PR #5372, opened 2022-05-14
fix(vats): store results of createVatByName for zoe etc.

### LINKED ISSUE #5013: don't drop adminNode needed to upgrade Zoe

We're definiing a bunch of mechanisms to allow individual contracts to be upgraded, most of which is driven by Zoe and the per-contract governance code.. We need to nail down the process by which Zoe itself is upgraded.

At the swingset level, this will use the same tools as contract upgrade: something with the admin facet for the Zoe vat (probably bootstrap) will use it to tell swingset to shut down the old vat and replace it with a new code bundle. Zoe will need to keep all its retained state in durable objects/collections, and deliver the important parts to its successor via `baggage`. The new version will need to re-establish behavior for all durable kinds.

Tentatively assigning to MarkM as keeper of Zoe.

## Description of the Design

TBD

## Security Considerations

all of them

dckc:
closes #5013

## Description

In order to not lose the adminNode generated in E(loadVat)('zoe'),
add a store of the results of createVatByName.

This also means multiple calls to `loadVat()` on the same name
return the same vat root.

### Security Considerations

adminNodes are high privilege; keeping them around carries a risk of leaks. This risk is mitigated by the permit mechanism, which makes it statically evident from `manifest.js` that `makeVatsFromBundles` is the only function that produces it, and currently, no functions consume it.

Plus, losing them altogether leaves no means to upgrade zoe or any of the vats created using `loadVat`.

cleaned up a TODO, actually.

### Testing Considerations

I tried to test this mechanism in `test-boot.js` but the `consume` object is not in scope, so it's not straightforward.

I manually tested it by logging the keys of the store each time one was added. (Then I took that debug logging out.)

I trust we will eventually come up with tests for actually upgrading Zoe, though I don't see an issue that says so in so many words.

### DIFF
--- a/packages/vats/src/core/basic-behaviors.js
+++ b/packages/vats/src/core/basic-behaviors.js
@@ -3,6 +3,8 @@ import { E, Far } from '@endo/far';
 import { AssetKind, makeIssuerKit } from '@agoric/ertp';
 
 import { Nat } from '@agoric/nat';
+import { makeScalarMapStore } from '@agoric/store';
+import { provide } from '@agoric/store/src/stores/store-utils.js';
 import { makeNameHubKit } from '../nameHub.js';
 
 import { feeIssuerConfig } from './utils.js';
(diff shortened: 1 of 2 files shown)


michaelfig (approved):
LGTM!  Glad to see these are no longer dropped.

packages/vats/src/core/basic-behaviors.js
```diff
@@ -53,22 +55,33 @@ const bootMsgEx = {
...
+  const store = makeScalarMapStore();
+  vatStore.resolve(store);
+
   loadVat.resolve(bundleName => {
-    console.info(`createVatByName(${bundleName})`);
-    const root = E(svc)
-      .createVatByName(bundleName)
-      .then(r => r.root);
-    return root;
+    return provide(store, bundleName, _k => {
```
warner, line 79: I think this'll work, but I'm not excited about storing Promises long-term in Stores when we know it will be resolved reasonably soon (every time you send that Promise off-vat, swingset has to allocate a new `vpid` and then `syscall.resolve` it right away, which is a tiny bit ugly compared to just sending the Presence).  Also, if/when we make `bootstrap` Durable, this will have to change, since Durable Stores can't hold Promises.

I don't know of a decent way to benefit from the `provide` helper and also store the Presence. And open-coding `provide` would require extra work to get the handy memoization around `(bundleName)`. Maybe we could use an async form of `provide`, except it would have to return a Promise before resolution, and then a Presence afterwards, which would be a different kind of ugly.

Not suggesting you change it, just wanted to share my internal mental image of how Promises are used.

warner (approved):
that works!

packages/vats/src/core/basic-behaviors.js
dckc: > Maybe we could use an async form of `provide`,

I think I saw one of those around somewhere... I thought briefly about trying it...

> except it would have to return a Promise before resolution, and then a Presence afterwards, which would be a different kind of ugly.

I hadn't even thought of that. Yeah; I think I'll leave this as is.

PR #5373, opened 2022-05-15
Reconnect wallet more robustly

michaelfig:
## Description

Much more robust AgoricWalletConnection resets in the case of a broken or missing connection.


samsiegart (approved):
I have noticed that if I reset my ag-solo I have to refresh the wallet UI. Does this fix that?

michaelfig:
> I have noticed that if I reset my ag-solo I have to refresh the wallet UI. Does this fix that?

Yes, it does!

PR #5374, opened 2022-05-15
Bootstrap with oracle invitations

### LINKED ISSUE #5242: Create fake Oracle for economic testing
## Problem to Solve

For economic stress test and other test deployments, we must enable directly specify the price, and have it propagated through.

## Design

Clone `ManualPriceAuthority` and move the `setPrice` operation onto the creator facet rather than the priceAuthority. Note that ManualPriceAuthority` is made to simplify unit tests. Most code is reused from elsewhere, so I don't think it's worth trying to make a single notifier that covers both use cases.

michaelfig:
closes: #5242 

## Description

Enough mechanism to get an oracle invite in the econ committee member's wallet on boot with `decentral-economy-config.json`.

image


dckc:
> closes: https://github.com/Agoric/agoric-sdk/issues/5242

I'd like us to be more explicit about that. How do we manipulate the prices? Is an integration test feasible?

packages/run-protocol/src/proposals/committee-proposal.js
```diff
@@ -1,46 +1,7 @@
...
-  };
-
-  return Promise.all(
-    paths.map(async path => {
-      assert(Array.isArray(path), X`path ${path} is not an array`);
-      return nextPath(nameAdmin, path);
-    }),
-  );
-};
+import { reserveThenDeposit } from './utils.js';
```
dckc, line 4: how about `nameHubAux.js` or `proposalUtils.js`? `utils.js` says very little about what's inside.

packages/vats/src/nameHub.js
```diff
@@ -74,6 +74,28 @@ export const makeNameHubKit = () => {
         }
       }
     },
+    default(key, newValue, adminValue) {
```
dckc: unit test, please?

packages/vats/src/nameHub.js
```diff
@@ -74,6 +74,28 @@ export const makeNameHubKit = () => {
...
+      if (keyToRecord.has(key)) {
+        const record = keyToRecord.get(key);
+        if (!record.promise) {
+          // Already initalized.
+          return /** @type {any} */ (record.value);
+        }
+      }
+      return nameAdmin.update(key, newValue, adminValue);
+    },
+    set(key, newValue, adminValue) {
```
dckc: unit test, please?

packages/run-protocol/scripts/price-feed-core.js
```diff
@@ -0,0 +1,103 @@
...
+  const proposalBuilder = powers =>
+    defaultProposalBuilder(powers, {
+      AGORIC_INSTANCE_NAME,
+      IN_BRAND_DECIMALS: parseInt(IN_BRAND_DECIMALS, 10),
+      OUT_BRAND_DECIMALS: parseInt(OUT_BRAND_DECIMALS, 10),
+      oracleAddresses,
+      brandIn: lookup(inLookup).catch(() => undefined),
+      brandOut: lookup(outLookup).catch(() => undefined),
+    });
+  await writeCoreProposal('gov-price-feed', proposalBuilder); // gov-price-feed.js gov-price-feed-permit.json
```
dckc, line 100: Managing names such as `'gov-price-feed'` seems awkward. That is: the fact that running this script creates a file with this prefix in the current directory, and we have to be sure they're distinct for all these scripts. hm.

dckc (commented):
got called away; sharing what I've got so far...

packages/run-protocol/src/proposals/committee-proposal.js
michaelfig: It's already `proposals/utils.js`.  Is that enough, or would you prefer `proposals/proposalUtils.js`?

packages/vats/src/nameHub.js
michaelfig: Gladly!  I was getting to that today.

packages/vats/src/nameHub.js
michaelfig: Also gladly!

packages/run-protocol/scripts/price-feed-core.js
michaelfig: If we want better coordination, that seems like something for `agoric deploy ...` to endow.

packages/run-protocol/src/proposals/committee-proposal.js
dckc: > Is that enough

ok

michaelfig:
> I'd like us to be more explicit about that. How do we manipulate the prices? Is an integration test feasible?

I put some usage notes at the top of `manual-price-feed.js`.  I will think about how to do an integration test.

packages/vats/src/nameHub.js
```diff
@@ -35,7 +37,7 @@ export const makeNameHubKit = () => {
       }
       return E(firstValue).lookup(...remaining);
     },
-    entries() {
+    entries: () => {
```
turadg, line 40: we could fix this in one PR over the codebase, which might help the git blame history be more clear.

turadg, line 40: we could fix this in one PR over the codebase, which might help the git blame history be more clear.

packages/vats/src/types.js
```diff
@@ -38,11 +38,17 @@
...
  * deleted).
+ * @property {( key: string, newValue: unknown, newAdmin?: unknown) =>
+ *   Promise<any> } default Update if not already updated.  Return
+ *   existing value, or newValue if not existing.
  * @property {(
- *   key: string, newValue: unknown, newAdmin?: unknown) => void
+ *   key: string, newValue: unknown, newAdmin?: unknown) => Promise<any>
+ * } set Update only if already initialized. Reject if not.
+ * @property {(
+ *   key: string, newValue: unknown, newAdmin?: unknown) => Promise<any>
```
turadg: Why return `newValue`?

As above:
```suggestion
 * @property {<T>(
 *   key: string, newValue: T, newAdmin?: unknown) => Promise<T>
```

packages/vats/src/types.js
```diff
@@ -38,11 +38,17 @@
  * @property {(key: string) => void} reserve Mark a key as reserved; will
  * return a promise that is fulfilled when the key is updated (or rejected when
  * deleted).
+ * @property {( key: string, newValue: unknown, newAdmin?: unknown) =>
+ *   Promise<any> } default Update if not already updated.  Return
```
turadg, line 42: Can we assume the new value is the same type as the prior value?
```suggestion
 * @property {<T>( key: string, newValue: T, newAdmin?: unknown) =>
 *   Promise<T> } default Update if not already updated.  Return
```

packages/vats/src/types.js
```diff
@@ -38,11 +38,17 @@
...
+ *   existing value, or newValue if not existing.
  * @property {(
- *   key: string, newValue: unknown, newAdmin?: unknown) => void
+ *   key: string, newValue: unknown, newAdmin?: unknown) => Promise<any>
+ * } set Update only if already initialized. Reject if not.
```
turadg, line 46: I don't see this used anywhere. Are we going to need it?

dckc (approved):
LGTM.

packages/vats/src/nameHub.js
michaelfig: I prefer to do it incrementally, but I understand your point.

packages/vats/src/types.js
michaelfig: No, we can't make that assumption.

packages/vats/src/types.js
michaelfig: I put it in for symmetry with the stores.

packages/vats/src/types.js
michaelfig: Okay, I won't return newValue.

turadg: Looks like you made `set` and `update` return void but didn't change the runtime. 

https://github.com/Agoric/agoric-sdk/commit/ac12696a10372d66cff048afd78d6750d9771011#diff-42f4291b619264addc7a7e12a8902cf3c60e5f797e97b5e4fa2a1e0c01d111ffR101

https://github.com/Agoric/agoric-sdk/commit/ac12696a10372d66cff048afd78d6750d9771011#diff-42f4291b619264addc7a7e12a8902cf3c60e5f797e97b5e4fa2a1e0c01d111ffR122

michaelfig: Oops.  I'll fix that right away!

PR #5377, opened 2022-05-16
feat(AMM)!: make amm.addPool() require minimum collateral

Chris-Hibbert:
closes: #4643

## Description

When creating a new pool for the AMM, there will be a minimum amount
of IST. The caller is expected to provide an equivalent value of the
collateral to balance the AMM pool. The liquidity tokens corresponding
to that minimum will be added to the AssetReserve. If the caller adds
more liquidity than the minimum, they get the extra liquidity tokens.

This required updating a number of tests to initialize their AMMs with
a minimal amount.

### Security Considerations

This adds to the economic security of the INTER protocol, but doesn't have much to do with system security.

### Documentation Considerations

`MinInitialPoolLiquidity` needs to be added to the doc for governance parameters, and the contract description.

### Testing Considerations

updated existing tests.

Issue #4800, opened 2022-03-09
Remove “BLD” and “RUN” keys from run-protocol code


Tartuffo:
# RUN/IST role cases

## vaultFactory
Token that is minted, secured by collateral (per vault manager)
keyword: `Minted`
Proposal shape will be {give: Collateral, want: Debt}

## runStake / bldBoost
Token that is minted, secured by collateral.
keyword: 'Minted'

## feeDistributor
Fee.
keyword: `Fee`
The CollectFees offer result will by a seat with `getPayout('Fee')`.

## assetReserve
Fee mint to hold penalties
keyword: `Fee`

## psm
Stable
keyword: Stable (already)

dtribble:
This includes removing it as the keyword for the stable token in 
```javascript
  const runMint = await zcf.registerFeeMint('RUN', privateArgs.feeMintAccess);
```

dckc:
@turadg to propose role-based keywords

Chris-Hibbert:
I'd pretty strongly prefer 'loan' to 'debt' for the vault factory. It's only 'debt' on one side of the balance sheet.  Similarly with bldBoost.

Chris-Hibbert:
'collateral' doesn't seem ideal for BLD. There are lots of assets that are collateral. BLD is uniquely the one that represents equity in the chain. Maybe 'Stake' would work.

PR #5387, opened 2022-05-17
liquidation penalty proceeds to reserve

### LINKED ISSUE #5367: Distribute liquidation penalties to the reserve

Liquidation penalties don't yet transfer to the reserve. https://github.com/Agoric/agoric-sdk/issues/5167 moved liquidation penalty handling from VaultManager into the liquidation contracts. It didn't include transferring assets into the reserve since that wasn't a feature yet.

## Description of the Design

Update `liquidateIncrementally` but not `liquidateMinimum` as it won't be the contract at launch.

When a liquidation occurs, charge the penalty to a `penaltyPool` seat private to the contract.

Push fees directly into the reserve. No need for the collector à la `makeCollectFeesInvitation`.

Still separate the fee pushing from the fee collecting. ??? on a timer or triggered after each liquidation completion?

## Security Considerations
TBD
## Test Plan

New tests on `liquidateIncrementally` to ensure fees are collected in the proper amounts and moved to the reserve.

turadg:
closes: #5367

## Description

Upon completion of liquidation, push penalty proceeds to the reserve.

### Security Considerations

Reserve can now accept RUN anytime from anywhere.

--

### Testing Considerations

New test

### DIFF
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -7,8 +7,13 @@ import { handleParamGovernance, ParamTypes } from '@agoric/governance';
 import { offerTo } from '@agoric/zoe/src/contractSupport/index.js';
 
 import { AMM_INSTANCE } from './params.js';
+import { makeTracer } from '../makeTracer.js';
+
+const trace = makeTracer('Reserve', true);
 
 const makeLiquidityKeyword = keyword => `${keyword}Liquidity`;
+
+const RunKW = 'RUN';
 
 /**
  * Asset Reserve holds onto assets for the RUN protocol, and can
@@ -28,9 +33,17 @@ const makeLiquidityKeyword = keyword => `${keyword}Liquidity`;
  * @param {{feeMintAccess: FeeMintAccess, initialPoserInvitation: Payment}} privateArgs
  */
 const start = async (zcf, privateArgs) => {
-  /** @type {MapStore<Brand, Keyword>} */
+  /**
+   * Used to look up the unique keyword for each brand, including RUN.
+   *
+   * @type {MapStore<Brand, Keyword>}
+   */
   const keywordForBrand = makeStore('keywords');
-  /** @type {MapStore<Keyword, [Brand, Brand]>} */
+  /**
+   * Used to look up the brands for keywords, excluding RUN because it's a special case.
+   *
+   * @type {MapStore<Keyword, [brand: Brand, liquidityBrand: Brand]>}
+   */
   const brandForKeyword = makeStore('brands');
 
   const { augmentPublicFacet, makeGovernorFacet, params } =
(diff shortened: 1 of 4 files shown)


packages/run-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -80,9 +80,27 @@ const start = async zcf => {
...
+    });
+    const { deposited, userSeatPromise } = await offerTo(
+      zcf,
+      addCollateral,
+      harden({ Out: 'Collateral' }),
+      proposal,
+      penaltyPoolSeat,
+    );
+    await userSeatPromise;
+    const deposits = await deposited;
```
```suggestion
    const [deposits] = await Promise.all([userSeatPromise, deposited]);
```

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -45,6 +45,7 @@ const contractRoots = {
   liquidate: './src/vaultFactory/liquidateIncrementally.js',
   VaultFactory: './src/vaultFactory/vaultFactory.js',
   amm: './src/vpool-xyk-amm/multipoolMarketMaker.js',
+  reserve: './src/reserve/assetReserve.js',
```
Chris-Hibbert, line 54: Please add a test that notices that the penalties have been drained to the reserve.

Chris-Hibbert (dismissed):

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: Did you see https://github.com/Agoric/agoric-sdk/pull/5387/commits/513f944aaf416157b5f66be917fe4f080b8a4b4f? (it was 16min before the review posted but I'm guessing the browser window hadn't updated)

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -102,6 +102,10 @@ const start = async (zcf, privateArgs) => {
 
   /** @type {ZCFMint} */
   const runMint = await zcf.registerFeeMint('RUN', feeMintAccess);
+  const centralBrand = runMint.getIssuerRecord().brand;
+  // FIXME second term of tuple must be the liquidity brand
+  brandForKeyword.init('Central', [centralBrand, centralBrand]);
```
Chris-Hibbert: I don't think it would be constructive to add it this way. The liquidity keyword can't be replaced once this runs.  Should it call `addIssuer` in the test?

turadg: Agree there's got to be a better way. `addIssuer` fails however. I'll wait to see what options might work after #5377

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -7,9 +7,14 @@ import { handleParamGovernance, ParamTypes } from '@agoric/governance';
 import { offerTo } from '@agoric/zoe/src/contractSupport/index.js';
 
 import { AMM_INSTANCE } from './params.js';
+import { makeTracer } from '../makeTracer.js';
+
+const trace = makeTracer('Reserve', true);
 
 const makeLiquidityKeyword = keyword => `${keyword}Liquidity`;
 
+const RunKW = 'RUN';
```
dckc, line 16: Interesting... I have put my toe in the water with constants for keywords... I'm not entirely happy with the results. let's see how it goes here...

dckc: Did you consider something like `StableKW` where only the value would change when we go to IST?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -818,3 +826,30 @@ test('amm stopAfter - want too much', async t => {
...
+  const collateralAmount = AmountMath.make(aethBrand, 400n);
+  const loanAmount = AmountMath.make(runBrand, 270n);
+  await d.makeVaultDriver(collateralAmount, loanAmount);
+
+  // liquidate
+  d.setPrice(AmountMath.make(runBrand, 636n));
+  await waitForPromisesToSettle();
+
+  const { reserveCreatorFacet } = t.context;
+  const reserveAllocations = await E(reserveCreatorFacet).getAllocations();
```
dckc, line 851: `getAllocations` is on the _creator_ facet? odd. I wonder why that is.

dckc (approved):
LGTM

packages/run-protocol/src/reserve/assetReserve.js
turadg: I like that idea and it didn't occur to me. I wouldn't want to do it in just this PR but I'll think about it for https://github.com/Agoric/agoric-sdk/issues/4800

IST seems so central to the package that my intuition is abstracting it would just add cognitive load and provide no benefit. I see it could be beneficial in making the transition easier, but that's a one-shot change that shouldn't burden regular development.

PR #5388, opened 2022-05-18
Generate economy and loadgen configs from the core config

michaelfig:
## Description

- clear up some nameHub implementation that didn't match the types
- generate economy and loadgen configs from the core config to reduce redundancy
- silence an annoying `unrecognized syscall` error when converting slogs to OpenTelemetry traces
- allow `agoric deploy --hostport=wss://NETWORK.agsolo.agoric.net`


packages/cosmic-swingset/economy-template.json
```diff
@@ -0,0 +1,74 @@
...
+        "econCommitteeOptions": {
+          "committeeSize": 1
+        }
+      }
+    ]
+  },
+  "@agoric/pegasus/scripts/init-core.js",
+  {
+    "module": "@agoric/run-protocol/scripts/add-collateral-core.js",
+    "args": [
```
dckc: You were planning to make `entrypoint` explicit. Not yet, though?

dckc (approved):
looks good, though...

"Instagoric fixes" seems kinda un-inspired as a summary. If the PR title ends up in the git history, "generate economy and loadgen configs from the core config to reduce redundancy" seems like the high-order bit.

packages/cosmic-swingset/economy-template.json
michaelfig: > You were planning to make `entrypoint` explicit. Not yet, though?

Ah yes.  Will do now.

PR #5389, opened 2022-05-18
style: disable space-comment eslint rule

turadg:
_no ticket_

## Description

Some machine-readable comments require no spaces, e.g. `///` for TS. We had a suppression for this thought it looks like [the imported style config has since added an exception](https://github.com/airbnb/javascript/blob/d8cb404da74c302506f91e5928f30cc75109e74d/packages/eslint-config-airbnb-base/rules/style.js#L502).

Another example is `#region` folding markers (supported by VS Code built in and WebStorm with config). These cause lint errors when used as `//#region`. A workaround is `// #region` but it draws more attention to the term instead of making it appear as marking token.

We could add an [exception for this in local config](https://stackoverflow.com/questions/62469552/eslint-how-to-spaced-comment-exceptions-on-vscode-folding-regions-comments) or PR to the Airbnb Styleguide project to add it. Instead I propose here removing it.

Open to doing either of the other options if reviewers prefer.


michaelfig:
> Another example is `#region` folding markers[...]
> We could add an [exception for this in local config](https://stackoverflow.com/questions/62469552/eslint-how-to-spaced-comment-exceptions-on-vscode-folding-regions-comments) or PR to the Airbnb Styleguide project to add it.

I would prefer this option.  `spaced-comments` does help make our codebase more consistent, so I'd prefer not disabling it overall unless there is team agreement around doing so.

kriskowal:
I have no strong feelings one way or the other.

image

turadg:
> > Another example is `#region` folding markers[...]
> > We could add an [exception for this in local config](https://stackoverflow.com/questions/62469552/eslint-how-to-spaced-comment-exceptions-on-vscode-folding-regions-comments) or PR to the Airbnb Styleguide project to add it.
> 
> I would prefer this option. `spaced-comments` does help make our codebase more consistent, so I'd prefer not disabling it overall unless there is team agreement around doing so.

Done @michaelfig

michaelfig (approved):

PR #5391, opened 2022-05-18
test(zoe): offerTo handling of broken invitation

turadg:
## Description

While working on https://github.com/Agoric/agoric-sdk/pull/5387 I had an error in my invitation handled but didn't get any errors back. I thought this might be that the error was swallowed but it just doesn't come out until `getOfferResult()`.

This adds a test documenting that. 

### Security Considerations

--

Because this behavior is surprising, should it be documented more conspicuously? Perhaps the `offerTo` docs should recommend always calling `getOfferResult` to check whether the handler succeeded? (Or is there a way to detect earlier?)

### Testing Considerations

This is a test.

### DIFF
--- a/packages/zoe/src/contractSupport/zoeHelpers.js
+++ b/packages/zoe/src/contractSupport/zoeHelpers.js
@@ -283,9 +283,6 @@ const reverse = (keywordRecord = {}) => {
  * contract instance (contractA) and depositing the payouts in another
  * seat in the current contract instance (contractA).
  *
- * @template {object=} A Offer args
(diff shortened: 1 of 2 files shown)


Chris-Hibbert (approved):
> Because this behavior is surprising, should it be documented more conspicuously? Perhaps the offerTo docs should recommend always calling getOfferResult to check whether the handler succeeded? (Or is there a way to detect earlier?)


I think from the calling contract, `getOfferResult` is the best way to find out, so yes, it should be documented in `offerTo`.

PR #5392, opened 2022-05-18
Serialize method names in message encoding

### LINKED ISSUE #2481: move method name into serialized message data, rather than separate field
Blocking #2092

Blocking our use of `subscription`s. We're using `notifier's where we should be using descriptors because of this.

@warner explains possible solutions at 
https://github.com/Agoric/agoric-sdk/issues/2092#issuecomment-748232036
and
https://github.com/Agoric/agoric-sdk/issues/2092#issuecomment-748241538

FUDCo:
Changes message encoding to eliminate the `method` field and move its contents into the `args` capdata object (now renamed `methargs`) so that the method name will be serialized, allowing non-string values (notably symbols) to be used as method selectors (and also allows remote function call, i.e., no method selector at all, to be supported).

Fixes #2481

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -15,6 +15,7 @@ import { assert, details as X } from '@agoric/assert';
 import { importBundle } from '@endo/import-bundle';
 import { xsnap, recordXSnap } from '@agoric/xsnap';
 
+import { QCLASS } from '@endo/marshal';
 import { checkBundle } from '@endo/check-bundle/lite.js';
 import { createSHA256 } from '../lib-nodejs/hasher.js';
 import engineGC from '../lib-nodejs/engine-gc.js';
@@ -22,7 +23,6 @@ import { startSubprocessWorker } from '../lib-nodejs/spawnSubprocessWorker.js';
 import { waitUntilQuiescent } from '../lib-nodejs/waitUntilQuiescent.js';
 import { makeGcAndFinalize } from '../lib-nodejs/gc-and-finalize.js';
 import { insistStorageAPI } from '../lib/storageAPI.js';
-import { insistCapData } from '../lib/capdata.js';
 import { provideHostStorage } from './hostStorage.js';
 import {
   swingsetIsInitialized,
@@ -446,15 +446,35 @@ export async function makeSwingsetController(
     /**
      * @param {string} vatName
      * @param {string} method
-     * @param {import('@endo/marshal').CapData<unknown>} args
+     * @param {unknown[]} args
      * @param {ResolutionPolicy} resultPolicy
+     * @param {string[]} [slots]
      */
-    queueToVatRoot(vatName, method, args, resultPolicy = 'ignore') {
+    queueToVatRoot(
+      vatName,
+      method,
+      args = [],
+      resultPolicy = 'ignore',
+      slots = [],
+    ) {
+      function replacer(_, arg) {
+        if (typeof arg === 'bigint') {
+          return { [QCLASS]: 'bigint', digits: String(arg) };
+        }
+        if (arg === undefined) {
+          return { [QCLASS]: 'undefined' };
+        }
+        return arg;
+      }
+
       const vatID = kernel.vatNameToID(vatName);
       assert.typeof(method, 'string');
-      insistCapData(args);
       const kref = kernel.getRootObject(vatID);
-      const kpid = kernel.queueToKref(kref, method, args, resultPolicy);
+      const methargs = {
+        body: JSON.stringify([method, args], replacer),
+        slots,
+      };
+      const kpid = kernel.queueToKref(kref, methargs, resultPolicy);
       if (kpid) {
         kernel.kpRegisterInterest(kpid);
       }
--- a/packages/SwingSet/src/vats/comms/delivery.js
+++ b/packages/SwingSet/src/vats/comms/delivery.js
@@ -58,19 +58,15 @@ export function makeDeliveryKit(
 
   // dispatch.deliver from kernel lands here (with message from local vat to
   // remote machine): translate to local, join with handleSend
-  function sendFromKernel(ktarget, method, kargs, kresult) {
+  function sendFromKernel(ktarget, kmethargs, kresult) {
     const target = getLocalForKernel(ktarget);
-    const args = mapDataFromKernel(kargs, null);
+    const methargs = mapDataFromKernel(kmethargs, null);
     assert(
       state.getObject(target) || state.getPromiseStatus(target),
       X`unknown message target ${target}/${ktarget}`,
     );
-    assert(
-      method.indexOf(':') === -1 && method.indexOf(';') === -1,
-      X`illegal method name ${method}`,
-    );
     const result = provideLocalForKernelResult(kresult);
-    const localDelivery = harden({ target, method, result, args });
+    const localDelivery = harden({ target, methargs, result });
     handleSend(localDelivery);
   }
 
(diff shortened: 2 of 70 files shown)


packages/SwingSet/src/controller/controller.js
```diff
@@ -446,15 +446,29 @@ export async function makeSwingsetController(
...
-     * @param {import('@endo/marshal').CapData<unknown>} args
+     * @param {unknown[]} args
      * @param {ResolutionPolicy} resultPolicy
+     * @param {string[]} [slots]
      */
-    queueToVatRoot(vatName, method, args, resultPolicy = 'ignore') {
+    queueToVatRoot(vatName, method, args, resultPolicy = 'ignore', slots = []) {
+      function replacer(_, arg) {
+        if (typeof arg === 'bigint') {
+          return { [QCLASS]: 'bigint', digits: String(arg) };
```
warner, line 462: Should we maybe do `marshal.serialize` for these two (and peel off the `.body` component of the result), instead of peeking inside the serialization format and using QCLASS directly?

We only have one other place in `src/` that uses QCLASS directly (`makeUndeliverableError`), and I'm not sure how to get rid of that (we don't want a kernel stack to get dragged into a serialized error that's being delivered to userspace), but ideally we wouldn't have *any* `QCLASS` references from swingset.

warner: Hrm, I suppose the `legibilize` helper functions are another instance, and those aren't really ones we can delegate to `marshal`. OTOH they're debug functions, and it might be nice if the core functionality didn't make assumptions about how `marshal` works. OT3H `queueToVatRoot` is mostly used by test cases.

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -825,6 +825,23 @@ export default function makeKernelKeeper(
     // does not designate which current queue the message sits on, but that
     // there is some kernel queue holding the message and its content.
 
+    // XXX resolve before committing: the following fragment was in Brian's
```
warner: Yeah, I think this is stale, and came from one of mhofman's PRs that I was looking at around the same time. Please delete it.

packages/SwingSet/src/vats/comms/delivery.js
```diff
@@ -58,19 +58,15 @@ export function makeDeliveryKit(
 
   // dispatch.deliver from kernel lands here (with message from local vat to
   // remote machine): translate to local, join with handleSend
-  function sendFromKernel(ktarget, method, kargs, kresult) {
+  function sendFromKernel(ktarget, methargs, kresult) {
     const target = getLocalForKernel(ktarget);
-    const args = mapDataFromKernel(kargs, null);
+    const kargs = mapDataFromKernel(methargs, null);
```
warner: I think the `k` prefix is misaligned here. I think this should be:
```js
function sendFromKernel(ktarget, kmethargs, kresult) {
  const target = getLocalForKernel(ktarget);
  const methargs = mapDataFromKernel(kmethargs, null);
```

and of course update `localDelivery =` too.

packages/SwingSet/test/test-state.js
```diff
@@ -497,13 +497,13 @@ test('kernelKeeper promises', async t => {
   t.deepEqual(k.getKernelPromise(p1).subscribers, ['v3', 'v5']);
 
   const expectedAcceptanceQueue = [];
-  const m1 = { method: 'm1', args: { body: '', slots: [] } };
+  const m1 = { methargs: { body: '["m1"]', slots: [] } };
```
warner: as discussed, we shouldn't omit the arguments array, as that'll be treated as an eventual-get instead of an eventual-send

I suspect there's new way to factor our little helper functions (`capargs`/`capargsOneslot`/etc) now that we're building methargs instead of args, but I doubt it's worth putting too much time into now. Maybe for a later cleanup pass.

packages/SwingSet/test/util.js
```diff
@@ -99,8 +103,27 @@ export function capargsOneSlot(slot, iface = 'export') {
...
+  args = [],
+  slots = [],
+  result = null,
+) {
+  // Note:
+  //   args === [] --> empty arg list
+  //   args === null --> no arg list
+  // Normally these are interchangeable, but for testing purposes we sometimes
+  // need to generate one and sometimes the other, because even though the
+  // message system doesn't care, the comparison done by t.deepEquals() does.
```
warner: as discussed, this will become false, and we should provide `args = []` even if our test doesn't care about the args (unless we're explicitly testing eventual-get, in which case args=undefined or args=null or something.. I never know how that works with default arguments in function definitions)

Hm, that reminds me, the difference between `null` and `undefined` in JSON serialized situations is significant, as well as default arguments. We use JSON serialization for the worker-to-vat pipe. Now that both `method` and `args` are in a *marshal*led capdata, we'll trivially preserve `method = undefined` (indicating eventual-send-function, vs the usual eventual-send-method), as well as `args = undefined` (indicating eventual-get). Previously, with `method` being sent separately, we needed to worry about the worker-to-vat pipe converting an eventual-send-function's `method = undefined` into `method = null`. (I suppose liveslots just looks for truthyness, so the difference was irrelevant, but I know we've had problems with transcript equality here).

I guess the place to pay attention is functions like this `makeMessage` where `method` or `args` are coming from something other than liveslots and HandledPromise handle invocations. It might not be trivial for `makeMessage` to sense someone calling it with an explicit `args = undefined` (as an attempt to trigger eventual-get).

That's probably fine, I don't think we'll use `makeMessage` to construct an eventual-get operation, and if we do we might declare that `makeMessage (args=null)` is how you say "please build me the same message that liveslots would have created for `args=undefined`".

But yeah, I think this should do `body = [method, args]` unconditionally, and never `body = [method]`. That might mean additional churn, if we have test code calling `makeMessage` with `args = null` for some reason.

packages/SwingSet/src/controller/controller.js
warner: Ok, after looking at the rest of the PR and seeing just how casually we throw `qclass` literals around, I'm not going to hold up approval for this additional usage. Call it a nice-to-have.

warner (approved):
Looks pretty good.. needs a few variable-name changes, and the "args are always present, except `eventual-get` uses `args = undefined`" thing we discussed. Feel free to land after those are addressed.

warner:
@FUDCo Oh, I forgot to mention: we need to update `docs/comms.md` too, to update the wire protocol definition to remove `$method` from the `send:` lines.

packages/SwingSet/src/controller/controller.js
```diff
@@ -446,15 +446,35 @@ export async function makeSwingsetController(
...
      */
-    queueToVatRoot(vatName, method, args, resultPolicy = 'ignore') {
+    queueToVatRoot(
+      vatName,
+      method,
+      args = [],
+      resultPolicy = 'ignore',
+      slots = [],
+    ) {
+      function replacer(_, arg) {
```
erights, line 460: Why introduce a `replacer` that special cases bigints and undefined but nothing else? What is special about these two?

packages/SwingSet/src/controller/controller.js
```diff
@@ -446,15 +446,35 @@ export async function makeSwingsetController(
...
+        return arg;
+      }
+
       const vatID = kernel.vatNameToID(vatName);
       assert.typeof(method, 'string');
-      insistCapData(args);
       const kref = kernel.getRootObject(vatID);
-      const kpid = kernel.queueToKref(kref, method, args, resultPolicy);
+      const methargs = {
+        body: JSON.stringify([method, args], replacer),
```
erights, line 474: Is this an additional layer of JSON encoding, or does it replace a `JSON.stringify` removed elsewhere?

erights (commented):
I know it is already merged. Just some questions?

packages/SwingSet/src/controller/controller.js
FUDCo: Previously callers of `queueToVatRoot` had to provide a pre-cooked capdata object (and of course creating a capdata requires a `JSON.stringify` call to produce the `body` string).  `queueToVatRoot` is used by tests to inject messages into the kernel from the outside.  This particular PR required updating a lot of tests, so one of the things I did was move some of the necessary quasi-serialization work that previously each test had to do by hand into `queueToVatRoot` in order to make the API less delicate.  Ideally we would just use marshal and avoid a lot of the mess here, but unfortunately, we can't.  This entry point has to weirdly straddle the world as seen from inside a vat and world as seen from inside the kernel -- it's used to send messages into a vat as if the caller was another vat sending the message, but the caller isn't in a vat; it's a piece of unit test code that in effect exists in kernel space.  In particular, the stuff that would otherwise be done by `convertValToSlot` must instead be done by providing the slots array (containing raw kref strings) directly, because there is no clist for the sender.  Tests can get away with this because they construct their worlds to be just so and the worlds they construct tend to be very simple.

The replacer you asked about is migrated from some of the tests which sent messages with arguments containing (in this case) bigints and undefineds. and this freed the callers from having to construct `@qclass` structures manually.  This is another piece of making the API less delicate.  Conceivably other `@qclass` things might need to be added in the future if tests require them, but I don't expect that.

Issue #5366, opened 2022-05-13
Support off-chain consumption of historical chainStorage-backed subscriptions


gibson042:
## What is the Problem Being Solved?

Off-chain consumers need to read sequential values from a chain without silent loss of intermediate values.

## Description of the Design

Create a function that accepts an RPC node endpoint/connection/etc. and chainStorage key (cf. #4558) and returns a subscription object (cf. https://agoric.com/documentation/guides/js-programming/notifiers.html ) such that values written to chain storage at that key appear as async iteration results and any interruption is visible (e.g., as an exception).

## Test Plan

- [ ] spreadsheet loader works
- [ ] AMM dApp notices price changes without a transaction (without paying for the notifier)

PR #5393, opened 2022-05-18
feat(vaultManager): expose liquidation metrics

### LINKED ISSUE #5300: Expose econ metrics on vault liquidations

Visibility into the economy

## Description of the Design

Based on Feb 22 meeting (https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#) but modified to:

Vault Manager will provide notification that includes:

1. Total amount of collateral reclaimed til now
2. Total proceeds
3. Total shortfall
4. Total overage
5. Total count of all liquidations done
6. Total debt (outstanding debt remaining)
7. Total collateral of remaining Vaults.

turadg:
closes: #5300

## Description

Include liquidation metrics in VaultManager's notifications.

### Security Considerations

Exposes more data, but only that we want to be public.

The metrics and their descriptions:
```js
/**
 * @typedef {object} MetricsNotification
 *
 * @property {number}         numVaults        present count of vaults
 * @property {Amount<'nat'>}  totalCollateral  present sum of collateral across all vaults
 * @property {Amount<'nat'>}  totalDebt        present sum of debt across all vaults
 *
 * @property {Amount<'nat'>}  totalCollateralSold       running sum of collateral sold in liquidation // totalCollateralSold
 * @property {Amount<'nat'>}  totalOverageReceived      running sum of overages, central received greater than debt
 * @property {Amount<'nat'>}  totalProceedsReceived     running sum of central received from liquidation
 * @property {Amount<'nat'>}  totalShortfallReceived    running sum of shortfalls, central received less than debt
 * @property {number}         numLiquidationsCompleted  running count of liquidations
 */
```

It wasn't clear how to distinguish metrics for "now" vs "up to now". This This introduces a naming scheme: nouns are present values; past-participles are summative. It's included as a comment in `vaultManager.js`. Are there other places it should go? (assuming reviewer agrees to the scheme)

### Testing Considerations

New tests and new harness for expressing diffs from one subscription state to the next.

I want to have a test validating the arithmetic among debt, proceeds, overage, and sales. I'm waiting on https://github.com/Agoric/agoric-sdk/issues/5366 to be able to query the latest state. If that's done by the time this needs to land I'll include it; if not I'll change the FIXME to a TODO.


packages/run-protocol/src/vaultFactory/types.js
```diff
@@ -85,9 +85,9 @@
  * @property {() => Ratio} getLoanFee
  * @property {() => Promise<PriceQuote>} getCollateralQuote
  * @property {() => Ratio} getInterestRate - The annual interest rate on a loan
- * @property {() => RelativeTime} getChargingPeriod - The period (in seconds) at
+ * @property {() => NatValue} getChargingPeriod - The period (in seconds) at
```
turadg: these are specifically seconds, not arbitrary RelativeTime. they also can't be relative; they must be positive.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -429,6 +432,7 @@ const helperBehavior = {
    * @returns {Promise<string>} success message
    */
   adjustBalancesHook: async ({ state, facets }, clientSeat) => {
+    trace('adjustBalancesHook start');
```
turadg, line 435: not necessary for master. I'll remove before next push.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -31,7 +31,22 @@ import { checkDebtLimit } from '../contractSupport.js';
 
 const { details: X } = assert;
 
-const trace = makeTracer('VM', false);
+const trace = makeTracer('VM');
```
turadg, line 34: didn't mean to enable this in master. I'll remove before next push.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -549,12 +597,7 @@ const selfBehavior = {
       // TODO `await` is allowed until the above ordering is fixed
       // eslint-disable-next-line @jessie.js/no-nested-await
       const vaultKit = await vault.initVaultKit(seat);
-      state.totalCollateral = AmountMath.add(
-        state.totalCollateral,
-        vaultKit.vault.getCollateralAmount(),
-      );
```
turadg, line 555: this didn't previously account for changes in collateral after a vault was created. it's handled now in `updateVaultAccounting`

packages/run-protocol/src/vaultFactory/types.js
Chris-Hibbert: `RelativeTime` should perhaps be renamed to `Duration` and declared to be `Nat`. I don't think it can be negative.

For the moment, I think the `RelativeTime` used by the `TimerService` is actually what we want here.

Chris-Hibbert (changes requested):

packages/run-protocol/src/vaultFactory/types.js
turadg: sgtm. done.

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -6,7 +6,26 @@ import { AmountMath } from '@agoric/ertp';
...
 
-const trace = makeTracer('LIQ');
+const trace = makeTracer('LIQ', false);
+
+/**
+ *
+ * @param {Amount<'nat'>} debt
+ * @param {Amount<'nat'>} proceeds
+ */
+const calcMetrics = (debt, proceeds) => {
```
Chris-Hibbert: I think we're now consistently using 'metrics' for the data that gets published by subscriptions. As long as it's not exposed outside this helper, it doesn't matter. I think the naming is inconsequential, since the return value is `{ proceeds: proceeds.RUN, overage, shortfall }`.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -33,6 +33,21 @@ const { details: X } = assert;
 
 const trace = makeTracer('VM', false);
 
+// Metrics naming scheme: nouns are present values; past-participles are summative.
```
Chris-Hibbert: Is 'summative' an actual word? How about 'running  totals' or 'accumulations'?  If you prefer summative, I'll be fine with it. Its meaning is obvious.

packages/run-protocol/test/metrics.js
```diff
@@ -19,13 +21,14 @@ export const subscriptionTracker = async (t, subscription) => {
     const prevNotif = notif;
     notif = await metrics.getUpdateSince(notif.updateCount);
     const actualDelta = diff(prevNotif.value, notif.value);
+    console.log({ actualDelta });
```
Chris-Hibbert: shouldn't this have a label?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2398,60 +2400,264 @@ test('director notifiers', async t => {
 
   // Not testing rewardPoolAllocation contents because those are simply those values.
   // We could refactor the tests of those allocations to use the data now exposed by a notifier.
+
+  t.pass();
```
Chris-Hibbert: Isn't it a bug if no assertions get called on `t`?

packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -6,7 +6,26 @@ import { AmountMath } from '@agoric/ertp';
...
+  if (AmountMath.isGTE(debt, proceeds)) {
```
dtribble, line 17: This should be a conditional expression?

Chris-Hibbert (commented):
I think my only substantive comment is about whether `t.pass()` is necessary.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: It is indeed a word. Primed in my brain summative assessments. Accumulative is much better though because they grow monotonically. Changing to that.

packages/run-protocol/test/metrics.js
turadg: ideally. this is just a test so I don't feel strongly. I'll remove though and people can add it whip it up if they need to debug a failure.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Yeah, this was a holdover from when `assertChange` was asserting without `t`, in which case Ava needn't to be told that if we get to this line we passed. No longer necessary so I'll remove.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -178,14 +185,19 @@ const initState = (
     ...fixed,
     assetNotifier,
     assetUpdater,
+    compoundedInterest,
     debtBrand: fixed.debtBrand,
-    vaultCounter: 0,
+    latestInterestUpdate,
```
dtribble, line 190: "latest" here seems redundant.  All state is "latest"

Chris-Hibbert (approved):
Looking good!

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -243,7 +255,7 @@ const helperBehavior = {
       },
       updateTime,
     );
-    Object.assign(state, stateUpdates);
+    partialAssign(state, stateUpdates);
```
dtribble, line 258: I don't like this pattern. It makes the entire state of the vault vulnerable to the interest computation. Can we replace with explicit assign of the allowed fields?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -484,15 +522,32 @@ const managerBehavior = {
...
+      oldCollateral,
+      vaultId,
+    );
+    // totalCollateral += vault's collateral delta (post — pre)
+    state.totalCollateral = AmountMath.subtract(
+      AmountMath.add(state.totalCollateral, vault.getCollateralAmount()),
+      oldCollateral,
+    );
+    // debt accounting managed through minting and burning
+    facets.helper.updateMetrics();
```
dtribble, line 550: will this trigger two updates, one for collateral and one for debt? For an adjustBalances there should just be one.

packages/run-protocol/test/metrics.js
dtribble: comment out now?

dtribble: or should be using trace

packages/run-protocol/test/metrics.js
```diff
@@ -10,6 +10,8 @@ import { diff } from 'deep-object-diff';
 export const subscriptionTracker = async (t, subscription) => {
   const metrics = makeNotifierFromAsyncIterable(subscription);
   let notif;
+  const getLastNotif = () => notif;
```
dtribble, line 13: That's a strange name. Isn't it the last state snapshot?

dtribble: (oh it's in test. ok nvm :)

dtribble (approved):
LGTM. My only objection is in the partialAssign usage, which predates this PR. Please eitehr fix that or file a ticket for it.

packages/run-protocol/test/metrics.js
Chris-Hibbert: He did drop it.

packages/run-protocol/src/vaultFactory/liquidation.js
turadg: you mean `return AmountMath.isGTE(debt, proceeds) ?` or something else?

packages/run-protocol/test/metrics.js
turadg: was removed in last push

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: Good point. I'll let this one merge since it's on the train and follow up in the morning with another.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: will follow up

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: I think I'd see that in tests.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: @dtribble this is not about the state of "interest update", it's specifying the last in a stream of "interest update" values. What would be a better name?

this is sent as [part of the AssetNotification](https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/vaultFactory/vaultManager.js#L52-L58) so renaming would be a breaking change. If you think the new name idea is worth that lemme know.

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: d179e5

packages/run-protocol/src/vaultFactory/liquidation.js
turadg: 8b6231

PR #5395, opened 2022-05-19
Support symbol-named methods in virtual objects

### LINKED ISSUE #5359: allow Symbol-named methods in virtual/durable objects

@erights noticed that virtual objects (`defineKind` and `defineDurableKind`) only accept string-named methods in their `behavior` records. He needs to use `Symbol.asyncIterator` to build a durable Notifier (#4567).

@FUDCo says this will be pretty easy to do once #2481 is complete.

## Description of the Design

`defineKindInternal` needs to scan the `behavior` records with something like `Reflect.keys` or `Object.getOwnPropertySymbols` instead of / in addition to `Object.keys` / `Object.getOwnPropertyNames`, to get all the symbol-named properties too.

## Security Considerations

`Far`/`Remotable` already have a set of rules about what is/isn't allowed as a property, and we want to be consistent with that, but we already use `Far` to create the Representatives so worst-case we're too restrictive instead of too liberal.

## Test Plan

unit tests

FUDCo:
Fixes #5359

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -419,11 +419,9 @@ export function makeVirtualObjectManager(
     if (typeof obj !== 'object') {
       return 'not';
     }
-    if (Object.getOwnPropertySymbols(obj).length !== 0) {
-      return 'not';
-    }
     let established;
-    for (const [_name, value] of Object.entries(obj)) {
+    for (const prop of Reflect.ownKeys(obj)) {
+      const value = obj[prop];
       let current;
       if (typeof value === 'function') {
         current = 'one';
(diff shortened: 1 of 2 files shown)


packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -419,11 +419,9 @@ export function makeVirtualObjectManager(
     if (typeof obj !== 'object') {
       return 'not';
     }
-    if (Object.getOwnPropertySymbols(obj).length !== 0) {
-      return 'not';
-    }
     let established;
-    for (const [_name, value] of Object.entries(obj)) {
+    for (const prop of Reflect.ownKeys(obj)) {
```
warner, line 423: According to my crib sheet, `Reflect.ownKeys` returns non-enumerable properties, whereas `Object.entries` ignored them. But I think that's fine here. It'll cause a tiny behavior change: if someone provides a behavior object with non-enumerable properties, then previously they'd be ignored entirely, and now they'll be copied onto the Representative (and become enumerable), which doesn't seem like a problem to me.

warner (approved):
Looks good!

Issue #5075, opened 2022-04-11
As a liquidity provider, I need to be able to bring an arbitrary IBC asset over to Agoric and create a pool on the AMM


rowgraus:
- [x] permissioned - requiring a decision of the BLDer DAO
 - [ ] permissionless

samsiegart:
We don't have any UI for this currently

samsiegart:
dupe of https://github.com/agoric/agoric-sdk/issues/5078?

samsiegart:
@rowgraus Is this a dupe of #5078 ?

dckc:
Given an ERTP issuer, creating an AMM pool is permissionless. But to get an ERTP issuer from an arbitrary IBC asset, there are 2 paths I know of:

 1. `E(bankManager).addAsset(denom, keyword, proposedName, issuerKit)`
     - **not permissionless** - access to `bankManager` requires a governance decision by the BLDer DAO, because...
     - The `addAsset` call causes the IBC asset identified by `denom` to show up in everybody's wallet labelled with `keyword` and `proposedName`.
 2. Using pegasus
     - outside the scope of RUN preview and mainnet 1, yes?

So this does _not_ look like a dup, to me.

dckc:
We tested the permissioned path (USDC / AUSD).

I presume the permissionless path is out of scope for RUN preview (#5085).

Issue #4643, opened 2022-02-23
Make `amm.addPool` require initial assets


dtribble:
## What is the Problem Being Solved?

Manage production of AMM pools. There needs to be a cost to it to prevent spamming and to ensure initial collateral.

## Description of the Design

* Move `addPool` to a separate contract invitation that charges a fee and requires minimum initial collateral to create the pool
* The cost may be initial liquidity, which gets added to the Reserve.
* The initial cost and minimum liquidity need to be governance parameters.

## Security Considerations

Right now it's too easy and cheap to create pools.

dtribble:
The initial cost and minimum liquidity need to be governance paramters.

Tartuffo:
Should only happen when adding a new pool to the AMM.  addPool() should require, for example at least 1,000 RUN and some additional collateral. 

Some % (controlled by Gov) of the Liquidity Tokens should be held (by the Reserve) so the pool can't be completely drained to spam pools.  These LTs are an "minimum initial contribution"

dckc:
## catch-22? How to initialize a new interchain asset?

The status quo is:
  1. Someone sets up an IBC channel and brings over some tokens; this fixes a cosmos level denom of the form `ibc/...long-hex-hash...`.
  2. The BLDer DAO authorizes the following:
      1. use `E(bankManager).addAsset(denom, ...)` (aka VBANK) to create an ERTP issuer (and make it widely known)
      2. `addPool` in the AMM using this issuer
      3. `addIssuer` in the reserve
      4. `addVaultType` in the vault manager, with a debtLimit of 0
   3. once the AMM pool has sufficient liquidity, the econ committee raises the debtLimit

The straightforward approach in #5377 requires funds at the `addPool` step. We tried taking that step out, but then `addIssuer` in the reserve fails, because the reserve turns around and asks the AMM for the relevant liquidity issuer.

Possible solutions include:

  - Separate what the BLDer DAO authorizes into 2 parts: (a) `E(bankManager).addAsset(denom, ...)` and (b) reserve `addIssuer` and the rest. Let market participants fund the AMM pool in between
    - **drawback**: 2 BLDer DAO votes
  - Allow `addPool` without funds as a privileged operation (on the AMM creator facet)
    - **drawback**: window for abuse between the time the pool is created and when the econ committee raises the debt limit? @rowgraus how much of a problem is that?
  - Move the constraint on minimum funds from pool creation time to the 1st time someone tries to add liquidity
[...]

PR #5397, opened 2022-05-19
feat: permissionless AMM pools from Interchain assets

dckc:
refs: #5075, #4643

## Description

Given sufficient IST, create an issuer for an IBC denom and an invitation to add a pool for it to the AMM.

There are 2 steps:
  1. offer IST (Central) and provide a denom in the offer args; get back the issuer for that denom in the VBANK
  2. offer the new asset (Secondary) and get back liquidity.

Un-ties knots discussed in https://github.com/Agoric/agoric-sdk/issues/4643#issuecomment-1130411097 , https://github.com/Agoric/agoric-sdk/issues/5075#issuecomment-1115225570

TODO:
 - [x] `bankManager` should have a method like `addAsset` that does _not_ propagate to every wallet in the system.
   - postponed as #5412
 - [x] bootstrap wiring

### Security Considerations

Does not provide offer safety, due to usual propoerties of `offerTo`.

Requires `bankManager` as a private arg.

Yet Another Contract to document / review.
cc @jessysaurusrex 

### Testing Considerations

Lightly tested. Reviewers will please suggest further tests.

### DIFF
--- a/packages/run-protocol/src/interchainPool.js
+++ b/packages/run-protocol/src/interchainPool.js
@@ -1,0 +1,137 @@
+// @ts-check
+import { E, Far } from '@endo/far';
+import { AmountMath, AssetKind, makeIssuerKit } from '@agoric/ertp';
+import { offerTo } from '@agoric/zoe/src/contractSupport/index.js';
+
+const { details: X, quote: q } = assert;
+
+const COSMOS_DECIMALS = 6;
+
+/**
+ * Given sufficient IST, create an issuer for an IBC denom
+ * and an invitation to add a pool for it to the AMM.
+ *
+ * @param {ZCF<InterchainPoolTerms>} zcf
+ * @param {{
+ *   bankManager: ERef<BankManager>,
+ * }} privateArgs
+ *
+ * TODO: get minimumCentral dynamically from the AMM
+ * @typedef {{
+ *   minimumCentral: Amount<'nat'>,
+ *   amm: Instance,
+ * }} InterchainPoolTerms
+ *
+ * @typedef {{
+ *   denom: string,
+ *   decimalPlaces: number,
+ * }} AssetDetail
+ */
+export const start = (zcf, { bankManager }) => {
+  const {
+    brands: { Central: centralBrand },
+    minimumCentral,
+    amm,
+  } = zcf.getTerms();
+  AmountMath.coerce(centralBrand, minimumCentral);
+
+  const zoe = zcf.getZoeService();
+  /** @type {ERef<XYKAMMPublicFacet>} */
+  const ammPub = E(zoe).getPublicFacet(amm);
+
+  let kwNonce = 0;
+
+  /**
+   * @param {ZCFSeat} seat
+   * @param {AssetDetail} detail
+   */
+  const step1Handler = async (seat, detail) => {
+    assert.typeof(
+      detail,
+      'object',
+      'offer arg must be { denom, decimalPlaces? }',
+    );
+    const { denom, decimalPlaces = COSMOS_DECIMALS } = detail;
+    assert.typeof(denom, 'string');
+    assert.typeof(decimalPlaces, 'number');
+
+    const {
+      give: { Central: centralAmt },
+    } = seat.getProposal();
+    assert(
+      AmountMath.isGTE(centralAmt, minimumCentral),
+      X`at least ${q(minimumCentral)} required; only ${q(centralAmt)} given`,
+    );
+
+    const interAsset = makeIssuerKit(
+      denom,
+      AssetKind.NAT,
+      harden({ decimalPlaces }),
+    );
+    kwNonce += 1;
+
+    await zcf.saveIssuer(interAsset.issuer, `Interchain${kwNonce}`);
+
+    /** @type { OfferHandler } */
+    const step2Handler = async seat2 => {
+      const {
+        give: { Secondary: secondaryAmt },
+      } = seat2.getProposal();
+      AmountMath.coerce(interAsset.brand, secondaryAmt);
+
+      const liquidityIssuer = await E(ammPub).addPool(
+        interAsset.issuer,
+        `Interchain${kwNonce}`,
+      );
+      await zcf.saveIssuer(liquidityIssuer, `Liquidity${kwNonce}`);
+
+      const proposal = harden({
+        give: {
+          Central: centralAmt,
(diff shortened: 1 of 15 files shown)


packages/run-protocol/src/interchainPool.js
```diff
@@ -0,0 +1,138 @@
...
+    );
+
+    const interAsset = makeIssuerKit(
+      denom,
+      AssetKind.NAT,
+      harden({ decimalPlaces }),
+    );
+    kwNonce += 1;
+
+    const liquidityIssuer = await E(amm).addPool(
```
dckc: @Chris-Hibbert the `addPool` call would move to step 2 when we require funds (#5377). I can't move it yet because the AMM requires the liquidity brand in the **want** when adding liquidity, even though it doesn't use it.

Maybe I should fix that in this PR?

Chris-Hibbert: > @Chris-Hibbert the addPool call would move to step 2 when we require funds (https://github.com/Agoric/agoric-sdk/pull/5377). I can't move it yet because the AMM requires the liquidity brand in the want when adding liquidity, even though it doesn't use it.

> Maybe I should fix that in this PR?

Before and after #5377 the APIs for adding a pool are just different. Our choices are to reconcile in whichever PR is second, or merge this into #5377 once approved.

packages/run-protocol/src/interchainPool.js
```diff
@@ -0,0 +1,138 @@
...
+ * Given sufficient IST, create an issuer for an IBC denom
+ * and an invitation to add a pool for it to the AMM.
+ *
+ * @param {ZCF<InterchainPoolTerms>} zcf
+ * @param {{
+ *   bankManager: ERef<BankManager>,
+ * }} pa
+ *
+ * @typedef {{
+ *   minimumCentral: Amount<'nat'>,
```
Chris-Hibbert, line 21: If this should eventually be governed, let's at least add a comment to that effect here.

packages/run-protocol/src/interchainPool.js
```diff
@@ -0,0 +1,138 @@
...
+      const { userSeatPromise, deposited } = await offerTo(
+        zcf,
+        invitation,
+        undefined,
+        proposal,
+        seat2,
+      );
+      return deposited.then(_ => {
+        seat.exit();
+        seat2.exit();
```
Chris-Hibbert, line 108: I think there should be a `.catch()` clause here that returns the money from seat2 to seat1 and exits the seat.

I would test it by setting the AMM's minimum higher than the minimum enforced here.

Chris-Hibbert (approved):

packages/run-protocol/src/interchainPool.js
dckc: I added a TODO to get this value dynamically from the AMM (once #5377 lands)

packages/run-protocol/src/interchainPool.js
dckc: I plan to get the AMM's minimum on demand, so that test wouldn't last.

What is it that could throw? `seat.exit()` or `deposited`? In either case, what difference does it make which seat the money is returned in?

packages/run-protocol/src/interchainPool.js
dckc: I did fix the odd AMM constraint and move `addPool` to step 2. That way, when #5377 lands, we'll have the collateral on hand when we want to `addPool`.

dckc:
@Chris-Hibbert @turadg I wired up everything to get `test-gov-collateral.js` and such working. Please take a look.

packages/run-protocol/src/interchainPool.js
```diff
@@ -0,0 +1,137 @@
...
+
+        return E(userSeatPromise).getOfferResult();
+      });
+    };
+
+    const keyword = denom; // ISSUE: should not show up in all wallets.
```
turadg: what does `ISSUE` mean? something to do or just know about?

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -135,6 +146,20 @@ const makeScenario = async t => {
...
+        t.log('addAsset', { denom, keyword, issuer: `${kit.issuer}` });
+        t.truthy(kit.mint);
+        ibcKitP.resolve(kit);
+      },
+      getFeeCollectorDepositFacet: () =>
+        harden({
+          receive: () => {},
+        }),
+    });
+    // @ts-ignore mock doesn't have all the methods
```
turadg, line 156: `@ts-expect-error`

(not worth a push bc I plan to lint this soon)

turadg (commented):
Nothing stands out to me as a concern but given the security risks and my understanding of the interchain I don't feel comfortable approving and leave that to @Chris-Hibbert .

packages/run-protocol/src/interchainPool.js
dckc: > what does `ISSUE` mean?

um... something to do, I suppose.

In this case, it originated from discussion with @dtribble , who was OK postponing it; I didn't turn it into a separate issue, #5412 , until I got a chance to run it by @michaelfig , where we decided it's feasible in due course. Now that I have a number for it, I can add it to that comment.

packages/run-protocol/src/interchainPool.js
Chris-Hibbert: > what difference does it make which seat the money is returned in?

I thought seat2 was purely internal, but I now see that I was wrong.

It still seems surprising that the caller has to check both seats to see if there's money there.  

If someone has already saved an issuer on the AMM and called it 'Interchain5', then the `saveIssuer()` call will fail, and the allocation will be stuck on seat2.

Chris-Hibbert (approved):

packages/run-protocol/src/interchainPool.js
dckc: > It still seems surprising that the caller has to check both seats to see if there's money there.

In discussion with @dtribble and co., we didn't see an obvious improvement. (There are 2 offers, so having 2 seats to check on failure never seemed odd, to me.)

> If someone has already saved an issuer on the AMM and called it 'Interchain5', then the `saveIssuer()` call will fail

The `saveIssuer()` call here doesn't depend on issuers in the AMM, as we discussed. (There could be a collision at `addPool`, but that's before the `saveIssuer()` here.)

PR #5398, opened 2022-05-19
fix(ava-xs): report nested deepEqual differences

turadg:
## Description

Fixes the ava-xs `t.deepEqual` implementation so nested property differences are reported as assertion failures instead of being treated as equal.

The root bug was in `deepDifference()`: when comparing object properties, it returned success as soon as a nested comparison had no difference and ignored non-null nested deltas. The fix returns the first nested delta and only succeeds after all own properties compare equal.

This also adds regression coverage for the ava-xs assertion harness and an XS-compatible fixture that exercises the real `ava-xs` runner.

## Review Notes

The previous `test: count ava-xs test failures` change has been dropped. It mixed assertion failure accounting with test counting, as noted in review. Current `avaHandler.cjs` already follows the linked WIP approach by throwing on `loadScript` failure instead of emitting a synthetic `not ok` result.

## Testing

- `yarn workspace @agoric/xsnap ava test/ava-xs.test.js`
- `yarn workspace @agoric/xsnap node src/ava-xs.js test/ava-xs-fixtures/deep-equal.js`
- `yarn workspace @agoric/xsnap lint:types`
- `yarn workspace @agoric/xsnap lint:eslint`


turadg:
@dckc where should a test like this go? where it is fails because it runs in regular `Ava` which functions correctly. It's only in `ava-xs` that it fails.

dckc:
The bug should be fixed rather than documented, no?

Note also plans to move ava-xs and xsnap to endo:

 - https://github.com/endojs/endo/issues/535
 - https://github.com/endojs/endo/issues/681

turadg:
> The bug should be fixed rather than documented, no?

Sure, it _should_. Given priorities, this is what I have time for. And it's work that would have to be done anyway for a fix: a test.

So… where should a test like this go? I can't find any testing of the `avaAssertXs` module.

dckc:
I'm not used to testing test harnesses. But the filename you picked seems as good as any.

erights:
What bug in deepEqual? I could not tell from the PR.

turadg:
> What bug in deepEqual? I could not tell from the PR.

`t.deepEqual({ num: 1 }, { num: 2 })` is passing when run in `ava-xs`. That isn't run in CI because there are no tests of packages/xsnap/src/avaAssertXS.js`. 

> I'm not used to testing test harnesses.

Sure, when consuming a test harness. In this case, the package is making an assertion library. That's what needs tests, just like Ava does https://github.com/avajs/ava/blob/e387cba63ccf1a24a96b2375d61738ca38f48082/test-tap/assert.js#L368

I think the trick will be to define the `t` object separately from `avaAssertXS.js` so it can be worked with as a specimen instead of on the global object. Then regular Ava can be used to test it.

packages/xsnap/src/avaXS.js
```diff
@@ -142,6 +141,8 @@ async function runTestScript(
       assertionStatus[msg.status] += 1;
       if (msg.status === 'not ok') {
         console.warn({ ...msg, filename, label });
+        testStatus.total += 1;
```
dckc: This seems to mix up counting tests with counting assertions. I wonder why the logic below here doesn't suffice.

dckc (commented):
I'd like to talk this over.

packages/xsnap/src/avaXS.js
dckc: a thought: 3472734ac

packages/xsnap/src/avaAssertXS.js
```diff
@@ -42,8 +41,9 @@ function deepDifference(x, y) {
       Reflect.apply(hasOwnProperty, obj, [prop]);
     for (const prop of Reflect.ownKeys(x)) {
       if (hasOwnPropertyOf(y, prop)) {
-        if (!deepDifference(x[prop], y[prop])) {
-          return null;
+        const delta = deepDifference(x[prop], y[prop]);
+        if (delta) {
+          return delta;
```
dckc, line 46: I gather this is the gist of it.

dckc (approved):
looks good

Issue #5386, opened 2022-05-17
wire RPC publishing nodes into bootstrap/run-protocol code


warner:
## What is the Problem Being Solved?

@gibson042 's finished with the task of creating "RPC publishing nodes": an object which represents the authority to:

* 1: publish data to the chain's persistent+hashed storage, where it can be queried using a non-transactional RPC query at some particular hierarchical path string
  * (and updates to that data appear in the block Results log)
* 2: create a new publishing node whose path string is an extension of the original

These publishing nodes also have a method which reveals their path string, so if A chooses a child name and creates a new publishing node for B, A can hand just the publishing node to B, it doesn't also need to tell B what path was chosen.

The bootstrap code (`packages/vats/`) creates the "root publishing node", which has the shortest RPC query path possible (maybe `/swingset/publish` or something: it lives in the same namespace as all other cosmos-sdk modules).

The question is how to disseminate this publishing authority to the smart wallet and the core economy contracts.

We agreed this morning to avoid human-meaningful names if we possibly can. We also agreed that it would be good to be able to change the mapping in the future, e.g. if the smart wallet initially uses a path with `/42/`, we might need to replace that contract with a new one (which could receive a prefix of `/43/`), and clients need a way to discover the new path.

We agreed to defer having Zoe be involved in the process. Instead, whoever creates each contract should acquire a publishing node first, then pass it to the new contract instance via `privateArgs`. Maybe in the MN-2/3 timeframe we could make this more automatic, but we don't need it in MN-1.

## Description of the Design

The plan we sketched out calls for the bootstrap code to use the root publishing node to publish a table of name-to-path mappings. In particular the data it publishes (and updates via the Notifier protocol) is a JSON-encoded object which has one particular key whose value is a mapping from names to paths.

The bootstrap code has a function that manages a counter, and will create new numbered child nodes upon request. Perhaps `createChildNode(name) -> node`. The name is used to update the mapping.

The code that launches the RUN protocol needs access to this function. It will create new nodes for the contracts it creates, and pass them to `privateArgs`.

The new contracts will take those publishing nodes and use them to publish whatever they need.

The smart wallet will use its node to creates a bunch of address-keyed child nodes, one per client, under which it can publish data about the particular wallet account.

## Security Considerations

[...]

Issue #4398, opened 2022-01-26
Smart wallet middleware


michaelfig:
## What is the Problem Being Solved?

The existing Wallet UI is very close to using a model that is possible to implement via CosmJS-based middleware directly to the on-chain backend.  This issue is to track the pieces needed for an end-to-end.

## Description of the Design

On-chain, per-account state is accessible via Cosmos queries, events, and simple marshalled eventual sends can be made via transactions.

- [x] on-chain notifiers are published by middleware both as updates to queriable state, and susbcribable Cosmos events containing notifier records serialised in the context of the account.  This is possible with the `x/swingset` storage API.
- [x] UI middleware subscribes to events, then issues initialisation queries to obtain and deserialise those records.
- [ ] replace ocaps with a Presence that submits a `MsgWalletAction` Cosmos transaction to do a semantic operation (i.e. only one message needed per UI action)
- [ ] these actions are powerful, so may need confirmation from an external signer (Keplr or MetaMask)
- [ ] an "authz" designation from that signer to a password-protected, per-wallet-UI key can cut down on confirmations #4406 
- [ ] any state changes will be published via the notifiers, so that multiple client UIs can be kept consistent
- [ ] the async result of a `MsgWalletAction` is also published as a Cosmos event, but is designed only to report success or failure to the (single) calling UI initiating it (if any)

NOTE: the wallet backend implementation is specifically not included in this item.  For now, the backend will be a stub that produces:

```ts
type WalletBackend = Notifier<{
  contacts: Notifier<{
    meta: {
      id: string,
    },
    name: string,
    lastReceivedMessage: any,
    actions: ERef<{
      sendMessage: (data: any) => void,
    }>,
  }>,
}>

E(contact.actions).sendMessage('Hello, world!');
```
 
## Security Considerations

Key management is important.  Will need to be reviewed by our security experts.

## Test Plan

Integration testing, probably manual for now.

dckc:
@turadg to split out read-write as a separate issue

dckc:
I'd like to keep using the checklist here, so I'm re-opening this.

In particular, our initial choice of marshaller for the wallet state publishes purses etc. so it needs work, as discussed in https://github.com/Agoric/agoric-sdk/pull/5701#discussion_r917077689

dckc:
tldr: It looks like the `walletFactory` (aka multi-tennant wallet) should register for `BRIDGE.WALLET` events, look up the relevant wallet by address, and forward actions there.

[...]

PR #5400, opened 2022-05-19
Publish amm pool list subscription to chain storage

michaelfig:
refs: #4398
refs: #5386

## Description

Created new `makeStoredSubscription` to produce a `StoredSubscription<T>` (a subscription with Merkle tree query information).  Passed a `storageNode` and `marshaller` from bootstrap to the AMM contract for a new `StoredSubscription<{ XYK: Brand[] }>`.

Drive by:
- reenabled CI build cache
- allow synchronous throws in `iterationObserver`s to terminate the iteration
- fix `makeScaledPriceAuthority` to query its underlying authority just-in-time

### Testing Considerations

Here's a demo:

```
snow:cosmic-swingset michael$ agd query swingset keys ''
keys:
- activityhash
- published
pagination: null
snow:cosmic-swingset michael$ agd query swingset keys published
keys:
- amm
pagination: null
snow:cosmic-swingset michael$ agd query swingset keys published.amm
keys:
- metrics
pagination: null
snow:cosmic-swingset michael$ agd query swingset keys published.amm.metrics
keys: []
pagination: null
snow:cosmic-swingset michael$ agd query swingset storage published.amm.metrics
value: '{"body":"{\"XYK\":[{\"@qclass\":\"slot\",\"iface\":\"Alleged: IbcATOM brand\",\"index\":0}]}","slots":["board0371"]}'
snow:cosmic-swingset michael$
```

### DIFF
--- a/packages/notifier/src/asyncIterableAdaptor.js
+++ b/packages/notifier/src/asyncIterableAdaptor.js
@@ -99,7 +99,7 @@ export const makeAsyncIterableFromNotifier = notifierP => {
  * @returns {Promise<undefined>}
  */
 export const observeIterator = (asyncIteratorP, iterationObserver) => {
-  return new Promise(ack => {
+  return new Promise((ack, observerError) => {
     const recur = () => {
       E.when(
         E(asyncIteratorP).next(),
@@ -117,7 +117,7 @@ export const observeIterator = (asyncIteratorP, iterationObserver) => {
           iterationObserver.fail && iterationObserver.fail(reason);
           ack(undefined);
         },
-      );
+      ).catch(observerError);
     };
     recur();
   });
--- a/packages/notifier/src/storesub.js
+++ b/packages/notifier/src/storesub.js
@@ -1,0 +1,78 @@
+// @ts-check
+import { E } from '@endo/eventual-send';
+import { Far, makeMarshal } from '@endo/marshal';
+import { observeIteration } from './asyncIterableAdaptor.js';
+
+/**
+ * Begin iterating the source, storing serialized iteration values.  If the
+ * storageNode's `setValue` operation rejects, the iteration will be terminated.
+ *
+ * Returns a StoredSubscription that can be used by a client to directly follow
+ * the iteration themselves, or obtain information to subscribe to the stored
+ * data out-of-band.
+ *
+ * @template T
+ * @param {Subscription<T>} subscription
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<ReturnType<typeof makeMarshal>>} [marshaller]
+ * @returns {StoredSubscription<T>}
+ */
+export const makeStoredSubscription = (
+  subscription,
+  storageNode,
+  marshaller = makeMarshal(undefined, undefined, {
+    marshalSaveError: () => {},
+  }),
+) => {
+  /** @type {Unserializer} */
+  const unserializer = Far('unserializer', {
+    unserialize: E(marshaller).unserialize,
+  });
+
+  // Abort the iteration on the next observation if the publisher ever fails.
+  let publishFailed = false;
+  let publishException;
+
+  const fail = err => {
+    publishFailed = true;
+    publishException = err;
+  };
+
+  // Must *not* be an async function, because it sometimes must throw to abort
+  // the iteration.
+  const publishValue = obj => {
+    assert(storageNode);
+    if (publishFailed) {
+      // To properly abort the iteration, this must be a synchronous exception.
+      throw publishException;
+    }
+
+    // Publish the value, capturing any error.
+    E(marshaller)
+      .serialize(obj)
+      .then(serialized => {
+        const encoded = JSON.stringify(serialized);
+        return E(storageNode).setValue(encoded);
+      })
+      .catch(fail);
+  };
+
+  if (storageNode) {
+    // Start publishing the source.
+    observeIteration(subscription, {
+      updateState: publishValue,
+      finish: publishValue,
+    }).catch(fail);
+  }
+
+  /** @type {StoredSubscription<T>} */
+  const storesub = Far('StoredSubscription', {
+    getStoreKey: () => storageNode && E(storageNode).getStoreKey(),
+    getUnserializer: () => unserializer,
+    getSharableSubscriptionInternals:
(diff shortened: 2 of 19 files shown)


packages/notifier/src/asyncIterableAdaptor.js
```diff
@@ -99,7 +99,7 @@ export const makeAsyncIterableFromNotifier = notifierP => {
  * @returns {Promise<undefined>}
  */
 export const observeIterator = (asyncIteratorP, iterationObserver) => {
-  return new Promise(ack => {
+  return new Promise((ack, observerError) => {
```
erights, line 102: When I see `adjectiveError` I expect it is some kind of error.

Should this be `observeError` or `errorObserver`? Either would be consistent with an observer that observes and error. Since it is a function rather than an object with methods, I prefer `observeError`.

packages/notifier/src/marshal.js
```diff
@@ -0,0 +1,104 @@
...
+ *     unserialize: import('@endo/marshal').Unserialize<unknown>
+ *   }>
+ * }} SubscriberDetails information and capabilities needed to follow a
+ * publisher out-of-band
+ */
+
+/**
+ * @template T
+ * @typedef {object} Publisher
+ * @property {(obj: ERef<T>) => void} publish publishes a jsonable value
```
erights: What do you mean by "jsonable"? If you mean what I expect, why do you hate bigints ;)

michaelfig: Because the publisher publishes only marshalled values, it isn't required to accept anything but jsonable input.

packages/notifier/src/storesub.js
```diff
@@ -0,0 +1,100 @@
...
+ * @property {() => Unserializer} getUnserializer get the unserializer for the stored data
+ */
+
+/**
+ * @template T
+ * @typedef {Subscription<T> & StoredFacet} StoredSubscription
+ */
+
+/**
+ * Begin iterating the source, publishing serialized iteration values.  If the
```
dckc: s/publishing/storing/?

(I'm in by-commit view or I'd suggest...)

packages/notifier/src/asyncIterableAdaptor.js
dckc, line 102: add a test for this API change?

packages/vats/src/lib-board.js
```diff
@@ -44,8 +45,30 @@ function makeBoard(
   const idToVal = makeStore('boardId');
   const valToId = makeStore('value');
 
+  // Create a marshaller that just looks up objects, not publish them.
+  const readonlyMarshaller = Far('board readonly marshaller', {
```
dckc, line 63: test for this feature?

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -161,8 +165,18 @@ const start = async (zcf, privateArgs) => {
...
   /** @type {SubscriptionRecord<MetricsNotification>} */
-  const { publication: metricsPublication, subscription: metricsSubscription } =
-    makeSubscriptionKit();
+  const {
+    publication: metricsPublication,
+    subscription: rawMetricsSubscription,
+  } = makeSubscriptionKit();
+  const { storageNode, marshaller } = privateArgs;
+  const metricsStorageNode =
+    storageNode && E(storageNode).getChildNode('metrics'); // TODO: magic string
```
dckc, line 174: maybe change the "TODO: magic string" comment to a justification: 'metrics' is derived from `getMetrics` which is already part of the public API. I suppose it's still feasible to make a manifest constant.

packages/run-protocol/src/vpool-xyk-amm/types.js
```diff
@@ -136,7 +136,7 @@
  * Prices and notifications about changing prices.
  * @property {() => Brand[]} getAllPoolBrands
  * @property {() => Allocation} getProtocolPoolBalance
- * @property {() => Subscription<MetricsNotification>} getMetrics
+ * @property {() => EOnly<StoredSubscription<MetricsNotification>>} getMetrics
```
dckc: `EOnly` is a left-over?

dckc (approved):
fun stuff!

just a few tweaks to consider

.github/actions/build-node/action.yml
```diff
@@ -5,11 +5,6 @@ inputs:
   node-version:
     description: 'The version of Node.js to use'
     required: true
-  cache-built:
```
mhofman: This was needed for the deployment test job which couldn't use the build cache because it runs on a different linux environment, so the native modules would not match. The default value of `true` should have made it no impact on any other job. I need to understand why you claim it affected other jobs.

mhofman (dismissed):

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
michaelfig: I left this TODO intact, for a later date.

.github/actions/build-node/action.yml
michaelfig: I'll create a new PR to distinguish the cache key without an explicit toggle.

packages/notifier/test/test-notifier-adaptor.js
```diff
@@ -49,6 +49,52 @@ test('observeIteration - update from iterator fails', t => {
...
+      observeIteration(finiteStream, {
+        updateState: state => {
+          t.is(state, 1);
+          throw new Error('sync propagates');
+        },
+        finish: state => {
+          t.fail(`unexpected finish with state ${state}`);
+        },
+        fail: reason => {
+          t.is(reason.message, 'sync propagates');
```
gibson042, line 65: @michaelfig question: how does the code calling `observeIteration` differentiate between a failure in the iterator and an error thrown from its own listener?

michaelfig: Not currently possible.  The right behaviour would be: if you install a `fail` handler, it will only be called if the iterator publishes a failure.  Listener throws should only be reflected by rejecting the outermost promise, and not call the `fail` handler.

PR #5401, opened 2022-05-19
refactor: improved debugging messages for durability

Chris-Hibbert:
## Description

While working on durability (#5283), we (@erights and I) improved some error message so they explicitly told us where our problem was.

Our first version threw *undefined must be a string* from inside `decodeToJustin`, even when there was not problem in the serialization. We eventually traced that to `decodeToJustin` not understanding the current representation of `ibid`s. This version sidesteps that issue when there's no durability problem.

```js
      if (durable) {
        serializedValue.slots.forEach((vref, slotIndex) =>
          assert(
            vrm.isDurable(vref),
            X`value is not durable: ${value} at slot ${slotIndex} of ${decodeToJustin(
              {
                body: JSON.parse(serializedValue.body),
                slots: serializedValue.slots,
              },
              true,
            )}`,
          ),
        );
      }
```

### Security Considerations

Reveals some internal information, but should only appear when debugging.

No user-serviceable parts inside.

### Testing Considerations

This is useful during tests.

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -14,6 +14,7 @@ import {
   makeCopyMap,
 } from '@agoric/store';
 import { Far, passStyleOf } from '@endo/marshal';
+import { decodeToJustin } from '@endo/marshal/src/marshal-justin.js';
 import { parseVatSlot } from '../lib/parseVatSlots.js';
 
 // The maximum length of an LMDB key is 254 characters, which puts an upper
@@ -60,6 +61,18 @@ const MAX_DBKEY_LENGTH = 220;
 
 function pattEq(p1, p2) {
   return compareRank(p1, p2) === 0;
+}
+
+function throwNotDurable(value, slotIndex, serializedValue) {
+  assert.fail(
+    X`value is not durable: ${value} at slot ${slotIndex} of ${decodeToJustin(
+      {
+        body: JSON.parse(serializedValue.body),
+        slots: serializedValue.slots,
+      },
+      true,
+    )}`,
+  );
 }
 
 export function makeCollectionManager(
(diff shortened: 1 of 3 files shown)


erights (approved):
The code LGTM, thanks!

The PR comment reads like the embedded snippet may be the replacement code rather than the code being replaced.

erights (changes requested):
Doing a request changes to block submission until I make a suggestion. Coming soon...

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -62,6 +63,18 @@ function pattEq(p1, p2) {
...
+function throwNotDurable(value, slotIndex, serializedValue) {
+  assert.fail(
+    X`value is not durable: ${value} at slot ${slotIndex} of ${decodeToJustin(
+      {
+        body: JSON.parse(serializedValue.body),
+        slots: serializedValue.slots,
+      },
+      true,
+    )}`,
+  );
```
erights: It's a lot more ceremony, but I think it is worth it. @FUDCo , please weigh in. Thanks.

```suggestion
  const body = JSON.parse(serializedValue.body);
  let encodedValue;
  try {
    encodedValue = decodeToJustin(
      {
        body,
        slots: serializedValue.slots,
      },
      true,
    );
  } catch (justinError) {
    const err = assert.error(
      X`value is not durable: ${value} at slot ${slotIndex} of ${serializedValue}`,
    );
    assert.note(
      err,
      X`decodeToJustin(${serializedValue}) threw ${justinError} `,
    );
    throw err;
  }
  assert.fail(
    X`value is not durable: ${value} at slot ${slotIndex} of ${encodedValue}`,
  );
```

erights (approved):
LGTM

packages/SwingSet/src/liveslots/collectionManager.js
erights: See https://github.com/endojs/endo/issues/1185

I'll fix that. But since that's in endo, we can't assume the fix here until the next cycle of the release dance completes. An error in reporting an error is rarely more important to report than the error we were originally trying to report in the first place. Diagnostics reporting errors should try to be robust in this sense. The above change reports both errors, but emphasizes the error we were trying to report.

erights: I just pushed the change I suggest above, so resolving. Of course, unresolve and continue the conversation if there's any issue.

FUDCo (approved):
Seems legit.

Issue #4639, opened 2022-02-22
Expose important economic data as RPC-backed subscriptions


Tartuffo:
This data can be used in Blockchain Explorers and via sites like The Graph for the community to get key information about the state of the RUN Protocol components.

Values will be emitted as a stream of states (cf. [Notifiers and Subscriptions](https://agoric.com/documentation/guides/js-programming/notifiers.html)) published into cosmos chain storage (which is a map where values are strings and keys are strings representing a dot-separated hierarchical path, e.g. "animalia.arthropoda.crustacea.malacostraca.decapoda") and ideally also [events](https://docs.cosmos.network/master/core/events.html), and consumable as a "prefix-lossy" subscription reading from the associated key.

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose:

1. AMM
    1. All trading pairs, the algo for them
    2. The pool amounts (X, Y, LP)
    3. Trades as they occur
    4. Liquidity change events (add, remove)
    5. Governance parameters, contract terms, fees
2. Vault Factory
    1. Per Contract
        1. Governance parameters, contract terms, fees
        2. Allowed collateral types
        3. Current reward pool (fees collected so far)
        4. Number and amount of shortfall
    2. Per Collateral
        1. Governance parameters, interest rate, liquidation ratio, liquidation penalty per collateral
        2. Current debt limit
        3. Number of vaults
        4. Outstanding debt, collateral held
        5. Compounded interest
        6. Liquidation strategy (may be per contract)
        7. Liquidation state (how much is currently being liquidated)
    3. Per Vault
        1. Debt
        2. Collateral
        3. Interest snapshot (used to calc the compounded interest per collateral)
        4. Status (open, closed, liquidating)
3. Liquidation contract
    1. Governance parameters (price impact limits, rate limits)
    2. Collateral and debt that it is currently working on.
    3. Cohort sale event (with price and the net shortfall)
4. Reserve Contract
    1. How much of each asset type
    1. Size of the reward pool?  (Question for Chris)
    1. Governance parameters
5. Price Authourity
    1. For each asset
    2. Aggregate price
    3. Source information (per oracle node operator)
    4. Governance parameters
6. runStake
    1. Per Contract
        1. Net amount minted
        2. Total collateral
        3. Distribution of utilization of loan amount (?)
[...]

Issue #4648, opened 2022-02-23
Expose key AMM economic data via subscription


Tartuffo:
This data can be used in Blockchain Explorers and via sites like The Graph for the community to get key information about the state of the RUN Protocol components.

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose from the AMM:

1. All trading pairs, the algo for them
2. The pool amounts (X, Y, LP)
3. Trades as they occur
4. Liquidity change events (add, remove)
5. Governance parameters, contract terms, fees

Chris-Hibbert:
AIUI, our theory is to publish states rather than events. My interpretation of that is that we'll have a separate stream of updates for each pool. It will publish the central and secondary balances in the pool and the amount of liquidity outstanding. These will be updated on every trade and liquidity event. A watcher can infer from this stream whether it's a response to a trade or liquidity change. 

We could also publish the fee balances, but since that isn't mentioned above, I'll leave it out for now.

I'll add a symbol for the algorithm, which will always be XY=K for now, and may eventually include "STABLE".

dckc:
Should this stay open until the data are available via RPC?

If not, which issue represents the work to make the AMM data available via RPC?

dckc:
Is it #5386 ? or #4331? should those be on the econ stress test release?

turadg:
> Should this stay open until the data are available via RPC?

I think we agreed to decouple reading and writing, trusting in the Subscriptions abstraction. So the tickets for pushing the data to a subscription on the public facet aren't concerned with how it gets read.

> If not, which issue represents the work to make the AMM data available via RPC?
> Is it https://github.com/Agoric/agoric-sdk/issues/5386 ? or https://github.com/Agoric/agoric-sdk/issues/4331? should those be on the econ stress test release?

I don't know yet how the econ testers will get the data. However it is, yes it should be included as a release dependency. cc @Tartuffo

Tartuffo:
Dan is working on a dashboard in a spreadsheet.

dckc:
> So the tickets for pushing the data to a subscription on the public facet aren't concerned with how it gets read.

Well, the title of this issue says "via RPC" in so many words. If that's no longer the scope, let's change it.

PR #5402, opened 2022-05-19
test: metrics subscription testing support

turadg:
## Description

For the epic https://github.com/Agoric/agoric-sdk/issues/4639 we're creating many new subscription streams. We want them to all work consistently in the Inter Protocol. While working on https://github.com/Agoric/agoric-sdk/pull/5393 I found it worthwhile to make a support utility. This makes it available throughout the package.

It takes a public facet and requires that it have a method `getMetrics` to help enforce that convention.

### Security Considerations

non-prod

helps codify a nascent convention

### Testing Considerations

test utility itself

### DIFF
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -7,7 +7,6 @@ import { E } from '@endo/eventual-send';
 import { deeplyFulfilled } from '@endo/marshal';
 
 import { makeIssuerKit, AssetKind, AmountMath } from '@agoric/ertp';
-import { makeNotifierFromAsyncIterable } from '@agoric/notifier';
 import buildManualTimer from '@agoric/zoe/tools/manualTimer.js';
 import {
   makeRatio,
@@ -42,6 +41,8 @@ import {
   installGovernance,
 } from '../supports.js';
 import { unsafeMakeBundleCache } from '../bundleTool.js';
+
+import { metricsTracker } from '../metrics.js';
 
 /** @type {import('ava').TestInterface<any>} */
 const test = unknownTest;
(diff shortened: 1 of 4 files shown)


packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2386,14 +2385,14 @@ test('director notifiers', async t => {
     'Chit',
     defaultParamValues(chit.brand),
   );
-  state = await E(metrics).getUpdateSince(state.updateCount);
-  t.deepEqual(state.value, {
-    collaterals: [aethBrand, chit.brand],
-    rewardPoolAllocation: {},
+  await m.assertChange({
+    collaterals: { 1: chit.brand },
```
turadg, line 2389: since this is expresses changes, it has to reference the place in the array (`1:`)

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2386,14 +2385,14 @@ test('director notifiers', async t => {
...
   });
 
   // Not testing rewardPoolAllocation contents because those are simply those values.
   // We could refactor the tests of those allocations to use the data now exposed by a notifier.
+
+  t.pass();
```
Chris-Hibbert: I'd prefer to pass `t` into `metricsTracker()`, and have it use `t.is()` rather than `assert`.

turadg: needed because otherwise there are no `t` assertions and Ava warns

Chris-Hibbert (changes requested):
In #4648 (metrics for AMM), I have `getMetrics()` for the top-level pools, but `getPoolMetrics(brand)` to get the metrics subscriptions for each of the individual pools. Is there any chance we can make a helper that will support both?

Chris-Hibbert (approved):
Nice!  thanks.

dckc:
metrics seems like a confusing name.

metrics are usually low level stuff, like prometheus. not contract stuff

turadg:
> metrics are usually low level stuff, like prometheus. not contract stuff

To my mind, metrics are whatever you're measuring. We have sprint metrics. We have business metrics.

Open to another name if you want to suggest one.

dckc:
I guess I can deal with "metrics" meaning one thing in `packages/run-protocol` and another thing in `packages/SwingSet`, especially since I don't have another one to suggest. I'm hesitant to speak for the rest of the team, though. hm.

Issue #5346, opened 2022-05-11
Replace uses of Notifier with new prefix-lossy Subscriptions


turadg:
## What is the Problem Being Solved?

With upcoming scaling abilities of Subscriptions, using Notifier unnecessarily constrains how the data is consumed.

## Description of the Design

Replace uses of Notifier with new Subscriptions

Tartuffo:
@turadg estimate please?

PR #5403, opened 2022-05-20
feat(notifier): subscription kit history limit

michaelfig:
refs: #5346

## Description

Implements "prefix lossy subscriptions" via `makeSubscriptionKit(historyLimit)` parameter (default=Infinity).

### Security Considerations

Improves memory profile, given cooperating subscribers (an old and slow subscription may still hold references to the entire history, but new ones won't).

Documented in README.md and backwards compatible.

### Testing Considerations

The test suite tests several important values of historyLimit.

### DIFF
--- a/packages/notifier/README.md
+++ b/packages/notifier/README.md
@@ -52,7 +52,7 @@ You can use the JavaScript `AsyncIterable` API directly, but it is more convenie
 the JavaScript `for-await-of` syntax or
 the `observeIteration` adaptor.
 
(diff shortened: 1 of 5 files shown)


erights:
@michaelfig be aware that @dtribble also has a sketch of a prefix lossy subscriber at https://gist.github.com/dtribble/4e7cc65fb0fa8aab5a4069e80d915454 , that we've been meaning to talk about. I think it is very close. Your approach seems very different. We should understand these differences before proceeding with an approach.

michaelfig:
Withdrawing in favour of the new-and-improved subscriptions/notifiers @erights is working on.

PR #5406, opened 2022-05-20
feat(keeper): storage path key encoding

michaelfig:
## Description

Move to a length-prefixed path encoding for storage to reduce kvstore usage.

### Testing Considerations

Passes regression tests.

### DIFF
--- a/golang/cosmos/x/swingset/keeper/keeper.go
+++ b/golang/cosmos/x/swingset/keeper/keeper.go
@@ -1,13 +1,14 @@
 package keeper
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/tendermint/tendermint/libs/log"
+	db "github.com/tendermint/tm-db"
 
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/store/prefix"
@@ -35,19 +36,71 @@ type Keeper struct {
 
 var _ types.SwingSetKeeper = &Keeper{}
 
-// A prefix of bytes, since KVStores can't handle empty slices as keys.
-var keyPrefix = []byte{':'}
+// Keys are structured as: `${numberOfPathElements}\0${zeroSeparatedPath}`, such
+// as `0\0` for the root element, and `1\0foo` for `foo`, and `3\0foo\0bar\0baz`
+// for `foo.bar.baz`.  The reason for this is space efficiency:
+//   - we can read a node's data with just a single store operation
+//     after mapping the key
+//   - we can scan for all of `foo.bar`'s children by iterating over the prefix
+//    `3\0foo\0bar\0`.
+// We still need to iterate up the tree until we are sure the correct ancestor
+// nodes are present or absent, but we don't need to fetch all an ancestor's
+// keys to do so.
+var (
+	keySeparator      = []byte{0}
+	keySeparatorRune  = rune(0)
+	pathSeparatorRune = '.'
+	pathSeparator     = string(pathSeparatorRune)
+	dataPrefix        = ":"
+	emptyValue        = []byte{1}
+)
 
-// keyToString converts a byte slice path to a string key
-func keyToString(key []byte) string {
-	return string(key[len(keyPrefix):])
+// keyToPath converts a byte slice path to a string key
+func keyToPath(key []byte) string {
+	// Split the key into its length and path components.
+	split := bytes.SplitN(key, keySeparator, 2)
+	pathBytes := split[1]
+
+	pathStr := strings.Map(func(chr rune) rune {
+		switch chr {
+		case pathSeparatorRune:
+			panic(fmt.Sprintf("key %q cannot contain %q", key, pathSeparatorRune))
+		case keySeparatorRune:
+			return pathSeparatorRune
+		default:
+			return chr
+		}
+	}, string(pathBytes))
+
+	return pathStr
 }
 
-// stringToKey converts a string key to a byte slice path
-func stringToKey(keyStr string) []byte {
-	key := keyPrefix
-	key = append(key, []byte(keyStr)...)
-	return key
+func pathToDataKey(keyStr string) []byte {
+	return append([]byte(dataPrefix), []byte(keyStr)...)
+}
+
+// pathToKey converts a string key to a byte slice path
+func pathToKey(keyStr string) []byte {
+	return pathWithLengthToKey(keyStr, 0)
+}
+
+func pathWithLengthToKey(keyStr string, initialPathLength int) []byte {
+	pathLength := initialPathLength
+	if len(keyStr) > 0 {
+		pathLength += 1
+	}
+	pathStr := strings.Map(func(chr rune) rune {
+		switch chr {
+		case keySeparatorRune:
+			panic(fmt.Sprintf("key %q cannot contain %q", keyStr, keySeparatorRune))
+		case pathSeparatorRune:
+			pathLength += 1
+			return keySeparatorRune
+		default:
+			return chr
+		}
+	}, keyStr)
+	return []byte(fmt.Sprintf("%d%c%s", pathLength, keySeparatorRune, pathStr))
 }
 
 // NewKeeper creates a new IBC transfer Keeper instance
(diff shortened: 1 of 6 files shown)


golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -35,19 +36,67 @@ type Keeper struct {
...
+var (
+	keySeparator      = []byte{0}
+	keySeparatorRune  = rune(0)
+	pathSeparatorRune = '.'
+	pathSeparator     = string(pathSeparatorRune)
+	dataPrefix        = ":"
+	emptyData         = []byte(dataPrefix)
+)
+
+// keyToPath converts a byte slice path to a string key
```
JimLarson: Either this name or description is wrong. looks like it should be:

```
keyToPath converts a byte slice path key into a path string.
```

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -35,19 +36,67 @@ type Keeper struct {
...
+		}
+	}, string(pathBytes))
 
-// keyToString converts a byte slice path to a string key
-func keyToString(key []byte) string {
-	return string(key[len(keyPrefix):])
+	return pathStr
 }
 
 // stringToKey converts a string key to a byte slice path
```
JimLarson: Similarly here - either they name or comment is wrong.

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -35,19 +36,67 @@ type Keeper struct {
...
 func stringToKey(keyStr string) []byte {
-	key := keyPrefix
-	key = append(key, []byte(keyStr)...)
-	return key
+	return stringWithLengthToKey(keyStr, 0)
+}
+
+func stringWithLengthToKey(keyStr string, initialPathLength int) []byte {
```
JimLarson: We should call the numeric prefix the "path depth" instead of "string length", as a length-prefixed string has a long and venerable tradition as something else entirely.

Consider instead:

```
// pathToKey converts a dot-separated path to a storage key
func pathToKey(path string) []byte {
    depth := bytes.Count([]byte(path), []byte(".")) + 1
    encodedPath := bytes.ReplaceAll(([]byte(path), []byte("."), keySeparator)
    return []byte(fmt.Sprintf("%d%s%s", depth, string(keySeparator), path)
}

// pathToChildrenPrefix converts a dot-separated path to the storage key prefix of its children
func pathToChildrenPrefix(path string) []byte {
    ...as above, but use depth+1 and append a final separator...
}

// keyToPath converts a storage key to the decoded dot-separated path
func keyToPath(key []byte) string {
    encodedDepth, encodedPath, found := bytes.Cut(key, keySeparator)
    if !found { ... error ...}
    // also validate depth
    return string(bytes.ReplaceAll(encodedPath, keySeparator, []byte("."))
}

```

JimLarson (commented):
Saving some outdated comments - will start over with recent changes.

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -35,19 +36,71 @@ type Keeper struct {
...
+			panic(fmt.Sprintf("key %q cannot contain %q", key, pathSeparatorRune))
+		case keySeparatorRune:
+			return pathSeparatorRune
+		default:
+			return chr
+		}
+	}, string(pathBytes))
+
+	return pathStr
+}
```
JimLarson: Consider the following implementation:

```
_, encodedPath, found := bytes.Cut(key, keySeparator)
if !found { ... panic... }
// the zero byte is guaranteed not to be found in utf-8 except for encoding nul
return bytes.ReplaceAll(encodedPath, keySeparator, []byte("."))
````

File: golang/cosmos/x/swingset/keeper/keeper.go [hunk 5]
```diff
@@ -35,19 +36,71 @@ type Keeper struct {
...
+
+func pathToDataKey(keyStr string) []byte {
+	return append([]byte(dataPrefix), []byte(keyStr)...)
+}
 
-// keyToString converts a byte slice path to a string key
-func keyToString(key []byte) string {
-	return string(key[len(keyPrefix):])
+// pathToKey converts a string key to a byte slice path
```
JimLarson: Terminology is still confusing. Since we have "key" both at the JS and Go levels, I'd prefer something like "path" to name the dot-separated string used as a key at the JS level, vs the byte array "store key" used to access the store. And we have both "data store" and ... "child store"?

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -35,19 +36,71 @@ type Keeper struct {
...
+		case keySeparatorRune:
+			panic(fmt.Sprintf("key %q cannot contain %q", keyStr, keySeparatorRune))
+		case pathSeparatorRune:
+			pathLength += 1
+			return keySeparatorRune
+		default:
+			return chr
+		}
+	}, keyStr)
+	return []byte(fmt.Sprintf("%d%c%s", pathLength, keySeparatorRune, pathStr))
```
JimLarson: I think this can be much shorter with the `bytes.` methods `Count` and `ReplaceAll`.

golang/cosmos/x/swingset/types/key.go
```diff
@@ -9,7 +9,6 @@ const (
 )
 
 var (
-	DataPrefix   = []byte(StoreKey + "/data")
-	KeysPrefix   = []byte(StoreKey + "/keys")
-	EgressPrefix = []byte(StoreKey + "/egress")
+	DataKeyPrefix = []byte(ModuleName + "/data")
+	PathKeyPrefix = []byte(ModuleName + "/path")
```
JimLarson: Consider moving the key/path conversion functions to this file.

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -278,91 +334,90 @@ func (k Keeper) GetStorage(ctx sdk.Context, path string) string {
...
-			keyList = append(keyList, keyStr)
-		}
-
-		// Update the list of keys
-		ancestorKey := stringToKey(ancestor)
-		if len(keyList) == 0 {
-			// No keys left, delete the parent.
-			keysStore.Delete(ancestorKey)
+			pathStore.Delete(pathToKey(ancestor))
+		} else if i < len(pathComponents) && k.HasStorage(ctx, ancestor) {
```
JimLarson: This works, but looks like it unnecessarily re-creates storage-less path entries. Assuming the invariant:

```
path entries exist if and only if self or some descendant has non-empty storage
```

then any path entry can assume that all its ancestors also have path entries. So if there is some routine `k.HasPath(ctx, path)` to see if the path entry exists, you can use `HasPath` instead of `HasStorage` here.

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -35,19 +36,71 @@ type Keeper struct {
...
+// Keys are structured as: `${numberOfPathElements}\0${zeroSeparatedPath}`, such
+// as `0\0` for the root element, and `1\0foo` for `foo`, and `3\0foo\0bar\0baz`
+// for `foo.bar.baz`.  The reason for this is space efficiency:
+//   - we can read a node's data with just a single store operation
+//     after mapping the key
+//   - we can scan for all of `foo.bar`'s children by iterating over the prefix
+//    `3\0foo\0bar\0`.
+// We still need to iterate up the tree until we are sure the correct ancestor
+// nodes are present or absent, but we don't need to fetch all an ancestor's
+// keys to do so.
```
JimLarson: Need to document the two kinds of keys you're using, and the invariants between the two stores, and really ought to define a "path" here, or point to its definition somewhere else.  I suggest something like the following:

- A "path" is a sequence of zero or more dot-separated nonempty strings of 7-bit non-nul, non-dot ASCII characters. So `""`, `"foo"`, and `"foo.bar.baz"` are paths but `"."`, "foo.", and "fo\0o" are not.
- A storage key for a path is the path prefixed with `:`.
- A path key for a path is the path prefixed with its length (in ASCII digits), separated by nul, followed by the path with dots replaced with nul. So the path key for the empty path is `0\0`.
- Path store entries have just a placeholder value. Path store entries exist if and only if self or some descendant have a non-empty data entry.

JimLarson (changes requested):
Found an efficiency bug where you can stop adding ancestor path entries sooner.

I still think this would be cleaner with just one store - use path keys and prefix the values with the `emptyValue` byte.

golang/cosmos/x/swingset/keeper/keeper.go
michaelfig: `bytes.Cut` is not present in golang 1.17.

golang/cosmos/x/swingset/keeper/keeper.go
michaelfig: Great, thanks!

golang/cosmos/x/swingset/keeper/keeper.go
michaelfig: "store key" means something different in Cosmos.

I settled for "encodedKey" and "encodedStore".

michaelfig:
> I still think this would be cleaner with just one store - use path keys and prefix the values with the emptyValue byte.

It turns out I'm not relying on proofs for data yet, so I can make this change.  Thanks!

michaelfig:
Thanks for your reviews, @JimLarson... I've made (at least my version of) all the changes described above in #5465.

 I'm closing this PR in favour of that broader revamp.

PR #5410, opened 2022-05-20
lint floating promises

### LINKED ISSUE #5407: lint floating promises

We use promises extensively. The language allows floating promises (fire and forget) but that's rarely what we want. We've seen bugs with missing error handling, surprising order of execution or loss of stack trace.

## Description of the Design

Adopt the typescript-eslint rule [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md)

## Security Considerations
--

## Test Plan

Check with developers to get the right signal to noise ratio in the config

turadg:
closes: #5407

## Description

working draft of #5407

will probably land some of these changes before this PR

### Testing Considerations

The collection of type information slows down linting.  Measured with `time yarn lerna run --no-bail lint:eslint` on my machine:

- before: `84.86s user 5.35s system 454% cpu 19.827 total` (1.4min)
- after: `824.27s user 135.32s system 556% cpu 2:52.39 total` (14min)

In CI the `lint` job went from [5 min before](https://github.com/Agoric/agoric-sdk/runs/6693643820?check_suite_focus=true) to [21 min after](https://github.com/Agoric/agoric-sdk/runs/6694898124?check_suite_focus=true)

### DIFF
--- a/packages/governance/src/closingRule.js
+++ b/packages/governance/src/closingRule.js
@@ -10,7 +10,7 @@ import { Far } from '@endo/marshal';
 /** @type {CloseVoting} */
 export const scheduleClose = (closingRule, closeVoting) => {
   const { timer, deadline } = closingRule;
-  E(timer).setWakeup(
+  return E(timer).setWakeup(
     deadline,
     Far('close voting', {
       wake: closeVoting,
--- a/packages/governance/src/contractGovernance/paramManager.js
+++ b/packages/governance/src/contractGovernance/paramManager.js
@@ -231,7 +231,7 @@ const makeParamManagerBuilder = zoe => {
      *
      * @param {SetInvitationParam} param0
      */
-    const setInvitation = async ([newInvitation, amount]) => {
+    const setInvitation = ([newInvitation, amount]) => {
       currentAmount = amount;
       currentInvitation = newInvitation;
       return harden({ [name]: currentAmount });
--- a/packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
+++ b/packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
@@ -43,7 +43,7 @@ const build = async (log, zoe) => {
         },
       });
       const subscription = E(electoratePublicFacet).getQuestionSubscription();
-      observeIteration(subscription, votingObserver);
+      void observeIteration(subscription, votingObserver);
 
       return Far(`Voter ${name}`, {
         verifyBallot: (question, instances) =>
@@ -69,7 +69,7 @@ const build = async (log, zoe) => {
         },
       });
       const subscription = E(electoratePublicFacet).getQuestionSubscription();
-      observeIteration(subscription, votingObserver);
+      void observeIteration(subscription, votingObserver);
 
       return Far(`Voter ${name}`, {
         verifyBallot: (question, instances) =>
--- a/packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
+++ b/packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
@@ -144,7 +144,7 @@ const checkContractState = async (zoe, contractInstanceP, log) => {
   const paramChangeObserver = Far('param observer', {
     updateState: update => log(`${update.paramNames} changed in a vote.`),
   });
-  observeIteration(subscription, paramChangeObserver);
+  void observeIteration(subscription, paramChangeObserver);
 
   // it takes a while for the update to propagate. The second time it seems good
   paramValues = await E(contractPublic).getGovernedParams();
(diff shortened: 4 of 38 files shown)


packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
```diff
@@ -43,7 +43,7 @@ const build = async (log, zoe) => {
         },
       });
       const subscription = E(electoratePublicFacet).getQuestionSubscription();
-      observeIteration(subscription, votingObserver);
+      void observeIteration(subscription, votingObserver);
```
erights, line 46: I have rarely see the `void` unary operator used. I'm surprised to see it.

Do your linting rules understand the `void` unary operator as allowing the operand to be a floating promise without complaint?

packages/governance/src/closingRule.js
```diff
@@ -10,7 +10,7 @@ import { Far } from '@endo/marshal';
 /** @type {CloseVoting} */
 export const scheduleClose = (closingRule, closeVoting) => {
   const { timer, deadline } = closingRule;
-  E(timer).setWakeup(
+  return E(timer).setWakeup(
```
erights, line 13: Was the need for the `return` here a genuine bug noticed by these linting rules, or was the `return` just needed to satisfy the rules. It is certainly an observable semantic change! (whereas `void`, IIUC, would satisfy the same linting rule but not change the semantics)

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -231,7 +231,7 @@ const makeParamManagerBuilder = zoe => {
      *
      * @param {SetInvitationParam} param0
      */
-    const setInvitation = async ([newInvitation, amount]) => {
+    const setInvitation = ([newInvitation, amount]) => {
```
erights, line 234: Removing the `async` converts some potential rejections into potential throws. Is this consistent with existing callers?

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
```diff
@@ -390,7 +390,7 @@ const makeBootstrap = (argv, cb, vatPowers) => async (vats, devices) => {
 
       await E(timer).tick();
       await E(timer).tick();
-      E.when(
+      await E.when(
```
erights, line 393: Omitted `await` a genuine bug found by linting rule?

packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
turadg: > Do your linting rules understand the void unary operator as allowing the operand to be a floating promise without complaint?

Exactly. It's the linter's convention "to explicitly mark a promise as intentionally not awaited."

packages/governance/src/closingRule.js
turadg: To satisfy the rule either `return` or `void` was necessary. I went with `return` because I figured allowing the caller of `scheduleClose` to get rejections would be preferred. Granted it's a semantic change.

I'll try to make sure that any semantic changes are instead only _suggested_ in this PR though the use of a eslint-disable comment.

packages/governance/src/contractGovernance/paramManager.js
turadg: > Removing the async converts some potential rejections into potential throws. Is this consistent with existing callers?

I planned to rely on unit and types testing to confirm that. But now I'm thinking to [eschew any semantic changes in this PR](https://github.com/Agoric/agoric-sdk/pull/5410#discussion_r878746786).

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
turadg: This is just a test, so not a bug in what we publish. It might be a fault in how the test runs; I planned to see in the test results from CI.

But it does raise a question about the idea above of avoiding semantic changes: should they also be avoided in tests? I think not.

turadg:
@mhofman I'm responding to your [comment](https://github.com/Agoric/agoric-sdk/pull/5457#issuecomment-1143978850) here since it's no longer relevant to that PR.

> I'm a little confused by the time increase.

When [typescript-eslint rules](https://typescript-eslint.io/rules/) are enabled that require type information, that has to be collected before the linting of the file. That will of course take time, but I think it's taking extra time because of our TS config. I think also if we ran linting in a single process instead of one per package it would be able to reuse more of the type information it collected.

> If it impacts CI, does it also impact a local IDE too?

Not much because the type information will be stored in the eslint process.

> A ten fold increase does not seem acceptable.

I don't think relative increase is the measure we should consider. It could go from .1s to 10s, a hundredfold increase, and that would be fine. The cost of development is devs waiting for their PRs to check. I do think 23min for the `lint` step could slow people down, which is why I raised the question.

If it _had to_ take 23min, I would argue that's worth it to get the linting of floating promises. But we can certainly cut that down. A simple option would be two split the lint job into two. There's still 5min setup time but it would be 5+(23-5)/2 or 14min which is comparable to other jobs. I expect we could speed up that 10min by refactoring our TS or ES configs, but that would take more eng work that I don't think can be prioritized right now.

turadg:
Linting landed by another PR. Lint fixes will be done incrementally.

PR #5411, opened 2022-05-20
Make legibilizeValue behave better given weird input

### LINKED ISSUE #5409: vat terminated when doing attMaker.makeAttestation from wallet
## Describe the bug
BLD Boost (RUN Stake) does not work on local chain. When I try to open a loan the following error occurs: https://gist.github.com/samsiegart/a67434be91ad31314591f61dd70c91be

## To Reproduce
1. Follow https://gist.github.com/samsiegart/ed32bcbaeb88b36ddb584f11a2baed06
2. Try to "Make Offer" from the RUN Stake UI and accept it in your wallet
3. Observe the offer fails and the above logs occur

## Expected behavior
It liens my BLD and gives me RUN.

FUDCo:
Fixes #5409

### DIFF
--- a/packages/SwingSet/src/lib/kdebug.js
+++ b/packages/SwingSet/src/lib/kdebug.js
@@ -1,5 +1,3 @@
-import { assert, details as X } from '@agoric/assert';
-
 let enableKDebug = false;
 
 export function kdebugEnable(flag) {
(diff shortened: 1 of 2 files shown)


warner:
Let's put an extra try/catch around the outermost legibilize-foo function, and test it with a capdata `.body` that's completely unparseable. The worst case should be an kernel-generated error message (`makeUndeliverableError.js`) which must resort to `"<unknown method>"`, rather than causing surprising crashes within the kernel somewhere.

warner (approved):
Good, thanks. Please add a test that triggers each `catch`, then feel free to land.

Issue #4399, opened 2022-01-26
Reserve contract receives JS metering fees; IST fees distributed to reward pool and reserve


rowgraus:
This ticket covers inflows to the Reserve as well as how the Reserve covers shortfalls from vault liquidations

Inflows to the Reserve should include:
- [x] All IST generated from execution fees (gas) on the Agoric chain should flow to an IST purse controlled by the Reserve #5427
- [x] 50% of fees collected from Inter Protocol contracts go to the Reserve.  Best if this is a governable parameter. #5448

Reward distribution mechanics
- [x] Daily, 10% of total reward purse should be distributed to stakers/validators as rewards (simple smoothing function) #5427

Tartuffo:
@warner @michaelfig Need a vBank purse for the Cosmos layer to put fees into.
That vBank purse then needs to be available in the vault contract, to cover defaults.  Needs to be setup in bootstrap.

Tartuffo:
@dckc Dan you are the primary on this.  LMK if that is problematic.

dtribble:
The Reserve pool now exists, so this should be unblocked.

dckc:
> Need a vBank purse for the Cosmos layer to put fees into.

I'm looking at `packages/vats/src/vat-bank.js` and I don't see how to do this. What should I be looking at, @michaelfig ? @JimLarson ?

michaelfig:
We will change the vbank from `getFeeCollector()` to `getRewardsCollector()` and `getReserveCollector()`, each of which is backed by a different cosmos-level module account.  `getRewardsCollector()` is backed by the `x/distr` module account, `getReserveCollector()` is backed by a new vbank-created account.

The reserve pool will get a virtual purse via bootstrap from a closely-held vbank `getReservePurse()` method, so that it can manipulate the cosmos-level module account's balance from the JS side.

rowgraus:
Updated to include discussion on vault shortfall mechanics

rowgraus:
Discussed in RUN Protocol engineering meeting last week.  Separating out ticket for reserve to cover shortfalls from liquidations: https://github.com/Agoric/agoric-sdk/issues/5299

Tartuffo:
We need to make this as small as possible to handle fees in MN-1.  Maybe we don't even need this if we only use Cosmos-level feed.

dckc:
@dtribble suggested we get more clarity on what are all the fees collected in the system and where they go. I offered to give it a whirl.

[...]

PR #5414, opened 2022-05-21
Cosmos vbank reserve purse

michaelfig:
refs: #4399

## Description

This clarifies the separation of the reserve flow from the reward flow.

Also, implements Cosmos module accounts for vbank purses, such as the `'vbank/reserve'` module acount for reserve purses.  That module account has a fixed `agoric1ae0lmtzlgrcnla9xjkpaarq5d5dfez63h3nucl` address (hashed from `'vbank/reserve'`).

Work still needed includes the logic around diverting the `x/swingset` fees and reward stream, and also integrating the module account purses in the JS reserve contract.


dckc (approved):
looks like forward progress.

JimLarson (commented):
FYI, the Golang looks good.

PR #5415, opened 2022-05-21
chore(wallet-ui): upgrade react-scripts to v5

### LINKED ISSUE #4934: Upgrade react-scripts to 5.0.0
## Describe the bug

react-scripts in @agoric/wallet-ui needs to be updated to react-scripts ^5.0.0
This change also appears to need some additional work with webpack and polyfill changes introduced in ^5.0.0

samsiegart:
fixes: https://github.com/Agoric/agoric-sdk/issues/4934


packages/wallet/ui/config-overrides/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "wallet-ui-overrides",
```
michaelfig: Please always prefix package names to prevent confusion via npm squatters.

```suggestion
  "name": "@agoric/wallet-ui-overrides",
```

michaelfig (commented):
Looks good!  Please wait for @mhofman's approval.

packages/wallet/ui/config-overrides/package.json
samsiegart: Done

mhofman:
Alright it took me a minute, but I think I fixed most issues:
- I removed most forced resolutions that were needed for the old react-scripts, and left a couple in place to avoid regressing in duplicate eslint or babel packages
- I harmonized the babel package versions (we ended up with a spread over more than 3 minor versions, which ended up causing duplicate packages)
- I added version resolutions to force patched packages to have a single version (the "lesser used" version would result in not being patched)
- I restored the acorn patch that was too eagerly removed (I don't claim the patch is necessary, just that it was still applicable)
- Updated the depd patch to the new version.

packages/wallet/ui/config-overrides/package.json
mhofman: Do we even need a package name? I don't see this used anywhere

mhofman (approved):

packages/wallet/ui/config-overrides/package.json
samsiegart: This is a hack to let the override script be in CommonJS, since it seems to be necessary, while `@agoric/wallet-ui` is `"type": "module"`.

samsiegart:
Thanks for patching and reviewing!

packages/wallet/ui/config-overrides/package.json
mhofman: You don't need to give a name. Also why not use the `.cjs` extension?

mhofman:
Looks like a first run of the swingset tests failed, which caused mergify to abort

packages/wallet/ui/config-overrides/package.json
samsiegart: Removed `name`. Discussed offline: `.cjs` wasn't working

Issue #5413, opened 2022-05-20
notifier returned by makeNotifierFromAsyncIterable is lossless


turadg:
## Describe the bug
The design of Notifier is that if you call `getUpdateSince()` without any arguments, you get the latest state.

This is different for the alleged Notifier returned by `makeNotifierFromAsyncIterable`. It never skips elements of the iteration of the underlying subscription.

## Expected behavior

A `Notifier` skips states consistently. Either `makeNotifierFromAsyncIterable` conforms or is renamed. (Or replaced if there's something better.)

PR #5418, opened 2022-05-21
WIP fix: yet another prefix lossy subscription sketch

erights:
Based on @dtribble 's at https://gist.github.com/dtribble/4e7cc65fb0fa8aab5a4069e80d915454

### DIFF
--- a/packages/notifier/README.md
+++ b/packages/notifier/README.md
@@ -140,7 +140,7 @@ local or remote promise for an AsyncIterable. `observeIteration` only sends it e
 using `E` (equivalent to the tildot syntax `~.`), and so doesn't care about these differences.
 
 While correct, Bob’s code is sub-optimal. Its distributed systems properties are not terrible, but
-Bob does better using `getUpdatesSince()` (provided by
+Bob does better using `getIterationSince()` (provided by
 SubscriptionKit). This lets Bob make a local AsyncIterable that coordinates better with producer
 Paula's IterationObserver.
 
--- a/packages/notifier/README.md
+++ b/packages/notifier/README.md
@@ -1,16 +1,16 @@
 # NotifierKits and SubscriptionKits
 
-This package provides two similar abstractions for producing and consuming asynchronous value 
+This package provides two similar abstractions for producing and consuming asynchronous value
 sequences, the *NotifierKit* and the *SubscriptionKit*. Both let a service notify clients of state changes.
 
-In JavaScript, async iterations are manipulated by `AsyncGenerators`, `AsyncIterables`, and `AsyncIterators`. 
-For an introduction to these concepts and implementations, 
+In JavaScript, async iterations are manipulated by `AsyncGenerators`, `AsyncIterables`, and `AsyncIterators`.
+For an introduction to these concepts and implementations,
 see [here](https://javascript.info/async-iterators-generators).
 
-For NotifierKit user documentation see 
+For NotifierKit user documentation see
 [here](https://agoric.com/documentation/distributed-programming.html#notifiers).
-The following doc more precisely describes the semantics of the NotifierKit and the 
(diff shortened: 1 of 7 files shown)


packages/notifier/src/subscriber.js
```diff
@@ -51,51 +24,117 @@ const makeSubscriptionIterator = tailP => {
...
+
+  let currentUpdateCount = 1;
+  /** @type {SubscriptionUpdates<T>} */
+  let currentP = tailP;
+  const advanceCurrent = () => {
+    // If tailP has not advanced past currentP, do nothing.
+    if (currentP !== tailP) {
+      currentUpdateCount += 1;
+      currentP = tailP;
+    }
```
michaelfig: Style suggestion:
```suggestion
    if (currentP === tailP) {
      return;
    }
    currentUpdateCount += 1;
    currentP = tailP;
```

packages/notifier/README.md
```diff
@@ -102,64 +102,63 @@ observeIteration(subscription, observer);
...
+values. This reference may be a local AsyncIterable, a remote presence of an AsyncIterable, or a
+local or remote promise for an AsyncIterable. `observeIteration` only sends it eventual messages
 using `E` (equivalent to the tildot syntax `~.`), and so doesn't care about these differences.
 
-While correct, Bob’s code is sub-optimal. Its distributed systems properties are not terrible, but 
-Bob does better using `getSharableSubscriptionInternals()` (provided by
-SubscriptionKit). This lets Bob make a local AsyncIterable that coordinates better with producer 
-Paula's IterationObserver. 
+While correct, Bob’s code is sub-optimal. Its distributed systems properties are not terrible, but
+Bob does better using `getUpdatesSince()` (provided by
```
michaelfig: I'm concerned that `getUpdatesSince` is a homonym and easily confused with `getUpdateSince`.  We need a better name.  `startIterationSince`?

michaelfig (commented):
I think I see how the introduction of the `updateCount` can help satisfy the "make a client-side notifier for a subscription" problem.

The reason why the implementation in #5403 handles configurable subscription history is that the other "prefix lossy" subscriptions are essentially the same as `historyLimit=1`, but I see legitimate uses of `historyLimit=Infinity` (current subscription implementation) and `historyLimit=0` (start with the next published value).  I want to be sure we are agreed upon breaking compatibility and restricting existing flexibility before we remove support for all subscriptions except `historyLimit=1`.

Having only "prefix lossy" subscriptions available to me means that some of my uses will require more bookkeeping on the client-side to consume as intended, and also while necessarily consuming more memory on the server-side.

packages/notifier/src/subscriber.js
erights: Nice. This way of saying it also gives a more natural place to put the comment.

packages/notifier/README.md
erights: `startIterationSince` didn't sound right. Not sure why, but to my ear it seems it should either be `startIterationAfter` or `getIterationSince`. I went with the latter. Also other renamings for consistency.

packages/notifier/src/types.js
```diff
@@ -106,51 +106,49 @@
...
- * localIterable =
- *   makeSubscription(E(remoteIterable).getSharableSubscriptionInternals());
- * ```
- * The resulting `localIterable` also supports such remote use, and
- * will return access to the same representation.
+ * @property {GetIterationSince<T>} getIterationSince
+ * Internally used to get the "current" SharableSubscriptionInternals
+ * in order to make a subscription iterator that starts there.
+ * The answer is "current" in that it was accurate at some moment between
+ * when you asked and when you got the answer.
```
michaelfig, line 143: Bookmarking this code so that I can refer to it.

turadg:
issue is closed https://github.com/Agoric/agoric-sdk/issues/5413

Issue #403, opened 2020-01-10
Divisible contracts


erights:
https://github.com/Agoric/agoric-sdk/issues/399 reminds me of something I had never previously explained. There's a lack of expressiveness of our current higher-order contract notion via invites that we should eventually repair. I'll tell the story as a variation of our normal Alice-Bob-Fred coveredCall example.

In the normal example, Alice writes a call by making an atomic offer, say for 100 shares of Q for 300Ag, where Alice's 100Q are placed in escrow, overing the call. Because Alice's offer is atomic, Bob is invited only to exercise the full call or none of it. Likewise, Bob can only sell all or none to Fred, unless Bob goes to the trouble of making yet another derivative contract. (The fact that Bob can, means that we don't currently have any ultimate lack of expressiveness. But we need something less awkward.)

Say instead, that Alice's offer is expressed as a multiple -- 100 time 1Q for 3Ag each. The same 100Q are placed into escrow. The multiples representation is just a clever way to express divisibility with quantization and without roundoff hazard. Multiply rather than divide.

Bob could still be invited to participate in this contract as a whole. However, what if invites could have the kind of fungibility suggest at https://github.com/Agoric/agoric-sdk/issues/399 ? Then, the invite given to Bob could be a fungible quantity of 100 identical invites, each to a corresponding division of Alice's escrowed Q. Bob could sell a portion of this to Fred.

Given that contracts are stateful programs, it is tricky to figure out how to divide the behavior of an arbitrary stateful program into independent slices that can operate on disjoint quantities. I have some thoughts on this, but it's a long way from a working plan. Generalizing invites this way will take some research.

katelynsills:
We decided that this is not by Beta and not by Mainnet.

erights:
> We decided that this is not by Beta and not by Mainnet.

That makes sense to me.

dckc:
we seem to be tracking this as #5540 now

PR #5422, opened 2022-05-24
feat(zoe): multiples, for making divisible offers

erights:
OMG Multiples! But tested only that we didn't break anything, because everything currently uses only `multiples: 1n`.

refs: #403, #5540

@Chris-Hibbert @dtribble @mhofman @turadg , once ready for review, I will ask you to review. Until then, please ignore. (@turadg , I cannot unlist you because of codeowners. @Chris-Hibbert I cannot and should not unlist you since you already commented.)

(Attempt https://github.com/Agoric/agoric-sdk/pull/11386 to bring this up to date failed)

### DIFF
--- a/packages/zoe/src/cleanProposal.js
+++ b/packages/zoe/src/cleanProposal.js
@@ -4,9 +4,11 @@ import { assertRecord } from '@endo/marshal';
 import { assertKey, assertPattern, mustMatch, isKey } from '@agoric/store';
 import { FullProposalShape } from './typeGuards.js';
 import { arrayToObj } from './objArrayConversion.js';
+import { natSafeMath } from './contractSupport/safeMath.js';
 
 import './internal-types.js';
 
+const { values } = Object;
 const { ownKeys } = Reflect;
 
 export const MAX_KEYWORD_LENGTH = 100;
@@ -140,6 +142,7 @@ export const cleanProposal = (proposal, getAssetKindByBrand) => {
   const {
     want = harden({}),
     give = harden({}),
+    multiples = 1n,
     exit = harden({ onDemand: null }),
     ...rest
   } = proposal;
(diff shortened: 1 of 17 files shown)


Chris-Hibbert:
This is nice!   I think it will work for the most obvious cases.

I don't know how to think about the corner cases without an outline of the expected usage pattern. With that, someone other than you (@erights) could conceivably finish this.

erights:
> This is nice! I think it will work for the most obvious cases.
> 
> I don't know how to think about the corner cases without an outline of the expected usage pattern. With that, someone other than you (@erights) could conceivably finish this.

@turadg @dtribble would that be one of you?

turadg:
> @turadg @dtribble would that be one of you?

SGTM! This is quite interesting. I'll have to sequence it with MN-1 priorities. Is this aiming for that release?
 
> I don't know how to think about the corner cases without an outline of the expected usage pattern.

Same here, I would need at least some description of tests of multiples > 1.

packages/zoe/src/cleanProposal.js
```diff
@@ -162,10 +164,41 @@ export const cleanProposal = (proposal, getAssetKindByBrand) => {
...
+    return amount;
+  }
+  const { brand, value } = amount;
+  assert(value >= 1n);
+  assert.typeof(
+    value,
+    'bigint',
+    X`multiples > 1 not yet implemented for non-fungibles: ${multiples} * ${amount}`,
+  );
+  return harden({ brand, value: value * multiples });
```
Chris-Hibbert: don't we use natSafeMath even for addition and multiplication?

packages/zoe/src/contractFacet/offerSafety.js
```diff
@@ -2,25 +2,41 @@
...
     }
     const allocationAmount = allocation[keyword];
-    return AmountMath.isGTE(allocationAmount, requiredAmount);
-  };
-  return Object.entries(giveOrWant).every(isGTEByKeyword);
+    if (!AmountMath.isGTE(allocationAmount, requiredAmount)) {
+      return 0n;
+    }
+    if (typeof requiredAmount.value !== 'bigint') {
+      multiples = 1n;
```
Chris-Hibbert: too much inconsistency between immediate returns and setting `multiples`.  Here I'd `return 1n`.

erights: The `multiples` here are being set within the loop, whereas the `multiples` variable is being set outside the loop. If on one iteration this condition applies, it'll set `multiples` to `1n`. If on a later iteration the condition above it applies, then the function as a whole will return `0n`. If we returned here, the function as a whole would return `1n` instead, which would be wrong.

packages/zoe/src/cleanProposal.js
erights: Done. Also for the divide in offerSafety.js where it was even more valuable because it explicitly states it is doing a floorDivide, which is important for any reader to know.

PR #5423, opened 2022-05-24
fix(interchainPool): get MinInitialLiquidity from AMM on demand

dckc:
refs: #4643

## Description

The minimum liquidity in the terms of the `interchainPool.js` contract could get out of sync with the AMM. So rather than a contract term, get it from the AMM on demand.

### Security Considerations

Reduces a source of inconsistency.

The code is the only documentation for this contract, so far.

### Testing Considerations

straightforward.

I cleaned up the other mock `BankManager` while I was at it.

### DIFF
--- a/packages/run-protocol/src/interchainPool.js
+++ b/packages/run-protocol/src/interchainPool.js
@@ -2,6 +2,7 @@
 import { E, Far } from '@endo/far';
 import { AmountMath, AssetKind, makeIssuerKit } from '@agoric/ertp';
 import { offerTo } from '@agoric/zoe/src/contractSupport/index.js';
+import { MIN_INITIAL_POOL_LIQUIDITY_KEY } from './vpool-xyk-amm/params.js';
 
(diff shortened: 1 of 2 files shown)


packages/run-protocol/src/interchainPool.js
```diff
@@ -28,15 +27,10 @@ const COSMOS_DECIMALS = 6;
...
-    brands: { Central: centralBrand },
-    minimumCentral,
-    amm,
-  } = zcf.getTerms();
-  AmountMath.coerce(centralBrand, minimumCentral);
+  const { amm } = zcf.getTerms();
 
   const zoe = zcf.getZoeService();
-  /** @type {ERef<XYKAMMPublicFacet>} */
+  /** @type {ERef<GovernedPublicFacet<XYKAMMPublicFacet>>} */
```
turadg, line 33: 👍

turadg (approved):

PR #5425, opened 2022-05-24
fix: default missed signed blocks should be higher

### LINKED ISSUE #5424: Signed Blocks window should be higher
## Describe the bug

The current signed blocks window is very low at 100, which results in a very short jailing time.  This should be increased to 10000 to match cosmos-relative networks

arirubinstein:
closes: #5424 

## Description

Fixes the default window for validator outages


michaelfig (changes requested):
This is not the place to change it.  Please see how these values are used in the math that calculates the signed blocks window, and fix it some other way (*not* by touching the `ORIG_*` values, but by adjusting or adding another one).

arirubinstein:
Refactored

michaelfig (approved):
Perfect!

Issue #5419, opened 2022-05-23
kernel: ##### KERNEL PANIC: unable to re-create vat v4 ##### (truncated .gz file)


arirubinstein:
On a test network, a few validator nodes dropped, and were unable to recover from a reboot with the following error.

```
{"level":"info","time":"2022-05-21T19:55:22Z","message":"starting ABCI with Tendermint"}
{"level":"info","args":["ag-chain-cosmos","--home","/root/.agoric","start","--home=/state/agoriloadtest-8","--log_format=json"],"time":"2022-05-21T19:55:22Z","message":"Start chain delegating to JS executable"}
{"level":"info","time":"2022-05-21T19:55:23Z","message":"starting ABCI with Tendermint"}
{"level":"info","module":"proxy","impl":"multiAppConn","time":"2022-05-21T19:55:24Z","message":"Starting multiAppConn service"}
{"level":"info","module":"proxy","connection":"query","module":"abci-client","impl":"committingClient","time":"2022-05-21T19:55:24Z","message":"Starting committingClient service"}
{"level":"info","module":"proxy","connection":"snapshot","module":"abci-client","impl":"committingClient","time":"2022-05-21T19:55:24Z","message":"Starting committingClient service"}
{"level":"info","module":"proxy","connection":"mempool","module":"abci-client","impl":"committingClient","time":"2022-05-21T19:55:24Z","message":"Starting committingClient service"}
{"level":"info","module":"proxy","connection":"consensus","module":"abci-client","impl":"committingClient","time":"2022-05-21T19:55:24Z","message":"Starting committingClient service"}
{"level":"info","module":"events","impl":"EventBus","time":"2022-05-21T19:55:24Z","message":"Starting EventBus service"}
{"level":"info","module":"events","module":"pubsub","impl":"PubSub","time":"2022-05-21T19:55:24Z","message":"Starting PubSub service"}
{"level":"info","module":"txindex","impl":"IndexerService","time":"2022-05-21T19:55:24Z","message":"Starting IndexerService service"}
{"level":"info","module":"consensus","hash":"\u000b&m\ufffd\ufffd\ufffd\ufffddOZ\ufffd~o\ufffd\u000fR7K\ufffd\ufffd\ufffd\ufffd\u0019\ufffd[\ufffdL\ufffd(Nq\ufffd","height":10889,"protocol-version":0,"software-version":"0.29.0","time":"2022-05-21T19:55:24Z","message":"ABCI Handshake App Info"}
[...]

PR #5426, opened 2022-05-24
fix(swing-store): check fs streams are ready

mhofman:
refs: #5419 

## Description

I wondering if something is causing the fs streams to not be in a good state and pipe to fail silently in that case. I had to add a [similar helper in the loadgen](https://github.com/Agoric/testnet-load-generator/commit/7d8ef1793b7115d9d58e041f94e321a18912ab9c), and it shouldn't hurt here.

### Security Considerations

None, this is doing extra error checking

N/A

### Testing Considerations

I'd like to experimentally test this commit

### DIFF
--- a/packages/swing-store/src/snapStore.js
+++ b/packages/swing-store/src/snapStore.js
@@ -10,6 +10,43 @@ const pipe = promisify(pipeline);
 const { freeze } = Object;
 
 /**
+ *
+ * @param {import("fs").ReadStream | import("fs").WriteStream} stream
+ * @returns {Promise<void>}
+ */
+export const fsStreamReady = stream =>
+  new Promise((resolve, reject) => {
+    if (stream.destroyed) {
+      reject(new Error('Stream already destroyed'));
+      return;
+    }
+
+    if (!stream.pending) {
+      resolve();
+      return;
+    }
+
+    const onReady = () => {
+      cleanup(); // eslint-disable-line no-use-before-define
+      resolve();
+    };
+
+    /** @param {Error} err */
+    const onError = err => {
+      cleanup(); // eslint-disable-line no-use-before-define
+      reject(err);
+    };
+
+    const cleanup = () => {
+      stream.off('ready', onReady);
+      stream.off('error', onError);
+    };
+
+    stream.on('ready', onReady);
+    stream.on('error', onError);
+  });
+
+/**
  * @param {string} root
  * @param {{
  *   tmpName: typeof import('tmp').tmpName,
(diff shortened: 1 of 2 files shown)


packages/swing-store/src/snapStore.js
```diff
@@ -9,6 +9,45 @@ const pipe = promisify(pipeline);
...
+    if (!stream.pending) {
+      resolve();
+      return;
+    }
+
+    stream.destroyed;
+
+    const onReady = () => {
+      cleanup(); // eslint-disable-line no-use-before-define
+      resolve();
```
warner, line 31: is there any chance that an error would cause both `onReady` and `onError` to fire, causing a race between resolution and rejection?

packages/swing-store/src/snapStore.js
```diff
@@ -87,14 +126,16 @@ export function makeSnapStore(
   /** @type {(input: string, f: NodeJS.ReadWriteStream, output: string) => Promise<void>} */
   async function filter(input, f, output) {
     const source = createReadStream(input);
-    const destination = createWriteStream(output);
+    const destination = createWriteStream(output, { flags: 'wx' });
```
warner, line 127: Why `x` ? Setting the executable bit doesn't make sense to me.

warner (commented):
I don't know enough about the Stream API to approve this.. maybe grab @kriskowal instead of me.

packages/swing-store/src/snapStore.js
mhofman: `onError` could happen after `onReady`, but at that point pipe should handle it. With the `cleanup` here, there is no way for both to be triggered at this step (and regardless, the resolve/reject already perform and already called check.

packages/swing-store/src/snapStore.js
mhofman: `x` is for exclusive access. It will fail if the file already exists instead of truncating an existing file.

packages/swing-store/src/snapStore.js
```diff
@@ -9,6 +9,45 @@ const pipe = promisify(pipeline);
...
+      reject(new Error('Stream already destroyed'));
+      return;
+    }
+
+    if (!stream.pending) {
+      resolve();
+      return;
+    }
+
+    stream.destroyed;
```
kriskowal: Does this have a desirable side effect not achieved on line 19?

kriskowal (approved):
I abhor the Node.js stream API. This looks like it should work to me.

packages/swing-store/src/snapStore.js
mhofman: Yeah I saw that, I think it was a left over of bad copy paste, it's also in the original 😳

Issue #5375, opened 2022-05-16
InstAgoric network for Econ Stress Test (xnet)


Tartuffo:
This must include the installation of the code-complete Inter Protocol contracts, as well as the Oracles and required collaterals wired through.

 - RPC-based dash board (#5681) requires
   - [x] #5963 landed to address #5959, needed by #5354
   - [x] re-spin [xnet](https://xnet.agoric.net/) at SDK_TAG:-20220816214108 6b7b997 for short duration
   - [x] AMM working on xnet
     - [x] pool with > 0 liquidity - [add-initial-liquidity.js](https://gist.github.com/dckc/7733f85be2c25a1a47b639e4b94816ac#file-add-initial-liquidity-js)
       - [x] shared ag-solo, which has lots of tokens
     - [x] confirm that a trade works
       - [x] smart wallet provisioning: https://github.com/Agoric/instagoric/pull/2
       - [x] deploy recent wallet UI code to [wallet.agoric.app](https://wallet.agoric.app), amm.agoric.app
       - [x] deploy recent AMM code to [amm.agoric.app](https://amm.agoric.app)
   - [x] vaults working on xnet
     - [x] vote to raise debt limit >0: [vote-raise-debt-limit.js](https://gist.github.com/dckc/7733f85be2c25a1a47b639e4b94816ac#file-vote-raise-debt-limit-js)
       - [x] shared ag-solo, which has econ committee voting invitation
     - [x] Oracle price feed: [manual-price-feed.js](https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/scripts/manual-price-feed.js)
        - (https://github.com/Agoric/agoric-sdk/issues/5242 )
     - [ ] confirm that creating a vault with the UI, liquidating works
       - [x] create a vault from a script using [borrow-from-vault.js](https://gist.github.com/dckc/73205178bd44cf3f4bbcf7db0fc280a8#file-borrow-from-vault-js)
   - [x] re-spin [xnet](https://xnet.agoric.net/) for longer duration
     - [x] #5987 landed to address #5910

dckc:
[...]

PR #5428, opened 2022-05-25
Add addPool to AMM creatorFacet and reinstate economy boot

michaelfig:
refs: #5375

## Description

- Put addPool on the AMM creatorFacet, primarily to un-tie knots in automated bootstrap; not intended for normal production use.
- Fix economy bootstrap, and add to CI.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -55,38 +55,22 @@ export const makeAddIssuer = (zcf, isInSecondaries, brandToLiquidityMint) => {
 /**
  * @param {ZCF<import('./multipoolMarketMaker.js').AMMTerms>} zcf
  * @param {(brand: Brand, pool: PoolFacets) => void} initPool add new pool to store
- * @param {Brand} centralBrand
- * @param {ERef<Timer>} timer
- * @param {IssuerKit} quoteIssuerKit
  * @param {import('./multipoolMarketMaker.js').AMMParamGetters} params retrieve governed params
- * @param {ZCFSeat} protocolSeat seat that holds collected fees
  * @param {ZCFSeat} reserveLiquidityTokenSeat seat that holds liquidity tokens
  *   from adding pool liquidity. It is expected to be collected by the Reserve.
  * @param {WeakStore<Brand,ZCFMint>} brandToLiquidityMint
  * @param {() => void} updateMetrics
+ * @param {ReturnType<typeof definePoolKind>} makePool
  */
 export const makeAddPoolInvitation = (
   zcf,
   initPool,
-  centralBrand,
-  timer,
-  quoteIssuerKit,
   params,
-  protocolSeat,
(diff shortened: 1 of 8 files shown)


packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -291,10 +291,38 @@ const start = async (zcf, privateArgs) => {
     }),
   );
 
-  /** @type {GovernedCreatorFacet<*>} */
-  const creatorFacet = makeGovernorFacet(
+  // TODO: pass this into makeAddPoolInvitation
```
turadg: I think this should be done before merge to DRY out the two addPools. They've already diverged in that [the other one updates metrics](https://github.com/Agoric/agoric-sdk/blob/0f4e0b53ffea8ea608f200752d8d342c0a2ce332/packages/run-protocol/src/vpool-xyk-amm/addPool.js#L97). cc @Chris-Hibbert

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -41,8 +41,9 @@ export const makeAddIssuer = (zcf, isInSecondaries, brandToLiquidityMint) => {
         AssetKind.NAT,
         harden({ decimalPlaces: 6 }),
       ),
-      mint => {
-        zcf.saveIssuer(secondaryIssuer, keyword);
+      /** @param {ZCFMint} mint */
+      async mint => {
+        await zcf.saveIssuer(secondaryIssuer, keyword);
```
turadg, line 46: noting https://github.com/Agoric/agoric-sdk/issues/5407 to link this instance to that ticket

Chris-Hibbert (changes requested):
This makes it possible to create a new empty pool, violating the requirement for a minimum funding. It also allows/requires the pool to be funded using `addLiquidity` rather than going through `addPoolAndLiquidityHandler`, which means that no liquidity will be extracted to the reserve.

dckc:
@Chris-Hibbert , this no-fee `addPool` method is primarily intended to un-tie knots in automated bootstrap; it's not intended for normal production use. (I clarified the PR description to say as much.)

> This makes it possible to create a new empty pool, violating the requirement for a minimum funding.

The motivation for minimum funding was essentially spam prevention, wasn't it? This no-fee `addPool` method is on the creator facet, which is only available to the BLDer DAO, not to spammers. There's nothing inherently wrong with an empty AMM pool, is there?

> It also allows/requires the pool to be funded using `addLiquidity` rather than going through `addPoolAndLiquidityHandler`, which means that no liquidity will be extracted to the reserve.

Yes, this is a consideration that anyone with access to the AMM creator facet should be aware of. Is it a reason not to add the feature, though?

dckc:
discussion with @Chris-Hibbert , @dtribble and co suggests:

 - try PSM?
 - try 0 minimum required liquidity?

dckc:
obsolete in favor of #5441

PR #5429, opened 2022-05-25
feat(amm)!: push reserved assets to reserve contract

### LINKED ISSUE #5271: Reserve's addIssuer should be public

Before assets can be "contributed" to the Reserve, the Issuer must be known. Currently `addIssuer` is on the creator facet.

## Description of the Design

Move addIssuer to the public facet to make adding new collateral types permissionless.

## Security Considerations

We want to accept arbitrary assets, but someone could use this to add an unlimited number of issuers to our zcf.

## Test Plan

no big deal.

### LINKED ISSUE #5408: Reserve collects liquidity held by AMM from creating new pools

See https://github.com/Agoric/agoric-sdk/pull/5377/files#r877423297

#5377 changed the process for adding new pools to the AMM to enforce a minimum liquidity, and hold the corresponding liquidity tokens for the reserve. 

## Description of the Design

? How should the reserve receive the funds?
- PULL
  - \+ In general, pull is more reusable. Downside 
  - \- requires polling, especially wasteful given that pool creation is rare
- PUSH
  - simpler to wire up
  - matches https://github.com/Agoric/agoric-sdk/pull/5387

Decision: PUSH

## Security Considerations

legibility, bootstrap dependency issues. This needs to be done in a way that is visible and subject to governance, without allowing the process to be subverted.

## Test Plan

test deliver of funds. proving security properties is external to the testing.

turadg:
closes: #5408
closes: #5271

## Description

When the AMM makes a pool, add the collateral to the reserve.

This requires that the reserve has the issuer, so when an issuer is added to AMM, also add it to the reserve. Put this on the public facet so the AMM doesn't need special access, but use a special method so the the reserve validates it back against the AMM.

### Security Considerations

New method on AMM creator facet to set the reserve instance

Requires AMM to have `resolveReserveFacet` called before pools are created. 

### Testing Considerations

New tests. Updated bootstrap. Made some test config more consistent.


packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -630,6 +645,8 @@ test('price drop', async t => {
     shortfallBalance: AmountMath.makeEmpty(runBrand),
   });
   await m.assertChange({
+    // FIXME expects to know liquidity brand
```
turadg, line 648: these tests check the brand, but they don't know the brand of the liquidity. maybe a mock will work

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -77,20 +90,74 @@ const start = async (zcf, privateArgs) => {
...
+   * @param {Brand} secondaryBrand
+   */
+  const addIssuerFromAmm = async secondaryBrand => {
+    assert(
+      secondaryBrand !== runKit.brand,
+      `${RunKW} is a special case handled by the reserve contract`,
+    );
+
+    const secondaryIssuer = await E(ammPublicFacet).getIssuer(secondaryBrand);
+    // ??? why don't these objects have methods?
```
turadg: @Chris-Hibbert @dckc do you know a way to get objects that match the typedef for `Brand` and/or `Issuer`? Then I could use `getAllegedName()` to make the tests more readable.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -51,9 +51,22 @@ const start = async (zcf, privateArgs) => {
   /**
    * Used to look up the brands for keywords, excluding RUN because it's a special case.
    *
-   * @type {MapStore<Keyword, [brand: Brand, liquidityBrand: Brand]>}
```
turadg, line 54: @Chris-Hibbert given the requirement that `AddCollateral` invitation accept any brand, I thought it cleanest to make these maps ignore the liquidity distinction. instead that's handled through a separate lookup between brands.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: The test can look up the liquidityBrand from the AMM.  The AMM is in scope 
(`services.ammFacets.ammPublicFacet`).

packages/deploy-script-support/test/unitTests/test-startInstance.js
```diff
@@ -47,10 +49,15 @@ test('startInstance', async t => {
     },
   };
 
+  /** @type {Petname=} */
```
Chris-Hibbert: Does this drive-by fix fit in another PR?

packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -261,6 +265,9 @@ export const setupReserve = async ({
     },
   },
 }) => {
+  trace('setupReserve');
+  await reserveInstallation;
```
dckc: what's this for? why await it and then throw away the resolution?

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -77,20 +90,74 @@ const start = async (zcf, privateArgs) => {
...
+    liquidityBrandForBrand.init(baseBrand, liquidityBrand);
+
+    await zcf.saveIssuer(liquidityIssuer, liquidityKeyword);
+  };
+
+  let keywordNum = 0;
+  /**
+   * @param {Brand} secondaryBrand
+   */
+  const addIssuerFromAmm = async secondaryBrand => {
```
Chris-Hibbert: `secondaryBrand` is really the AMM's concept. I realize the Reserve knows that these brands are not the AMM's `centralBrand`, but after initialization, the Reserve doesn't treat them differently.  I'd prefer that the param name here either be `brand` or `ammSecondaryBrand`. WDYT?

packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -786,3 +797,17 @@ export const startRunStake = async (
...
+ * @param {EconomyBootstrapPowers} space
+ */
+export const linkAmmAndReserve = async space => {
+  trace('linkAmmAndReserve');
+  const reserveInstance = await space.instance.consume.reserve;
+  trace({ reserveInstance });
+  const reservePublicFacet = await E(space.consume.zoe).getPublicFacet(
+    reserveInstance,
+  );
+  const ammCreatorFacet = await space.consume.ammCreatorFacet;
```
dckc: These `await`s aren't necessary. I can imagine they were helpful for debugging... and I guess this isn't performance-sensitive stuff... so I guess we might as well leave them there to facilitate future debugging.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -77,20 +90,74 @@ const start = async (zcf, privateArgs) => {
...
+    assert(
+      secondaryBrand !== runKit.brand,
+      `${RunKW} is a special case handled by the reserve contract`,
+    );
```
Chris-Hibbert, line 161: If `addIssuerFromAMM` is public, it has to verify that the brand is a secondary brand (not a  liquidity brand) with the AMM before adding it.  Otherwise, someone could bollix up our careful plans by calling `addIssuerFromAmm(doubloonLiquidityBrand).` 
Why doesn't it work to add both issuers when someone (the AMM?) calls `addIssuerFromAmm`?

packages/run-protocol/src/reserve/assetReserve.js
dckc: are the objects remote? you should be able to use `E(brand).getAllegedName()`.

dckc (commented):
my brain is kinda polluted with other stuff; I hope @Chris-Hibbert can get to this soon

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -179,14 +244,19 @@ const start = async (zcf, privateArgs) => {
...
-    updateMetrics,
+    handlePoolAdded,
   );
   const addIssuer = makeAddIssuer(
     zcf,
     isSecondary,
     secondaryBrandToLiquidityMint,
+    () => {
+      // DISCUSSION: this gumms up all the testing setup to require the reserve to be created AND set onto the AMM
+      // before the AMM can add an issuer
```
Chris-Hibbert: If the Reserve (or some intermediary) pulled the liquidity tokens later, then the AMM wouldn't have to be sure that the Reserve was ready for the Liquidity tokens when the pool was created.

packages/run-protocol/src/proposals/econ-behaviors.js
turadg: exactly

packages/run-protocol/src/proposals/econ-behaviors.js
turadg: earlier feedback on missing producers. I'll do some cleanup since they're confusing

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -166,11 +172,70 @@ const start = async (zcf, privateArgs) => {
...
+    liquidityKeyword,
+  ) => {
+    trace('handlePoolAdded', { secondaryBrand, liquidityKeyword });
+    updateMetrics();
+
+    if (!reserveDepositFacet) {
+      console.error('🚨 no reserveDepositFacet');
+      return;
+    }
+    assert(reserveDepositFacet, 'Missing reserveDepositFacet');
```
Chris-Hibbert: This won't do anything after the conditional just above.

packages/run-protocol/test/amm/vpool-xyk-amm/setup.js
```diff
@@ -68,26 +82,38 @@ export const setupAMMBootstrap = async (
...
 ) => {
-  if (!zoe) {
-    ({ zoe } = await setUpZoeForTest());
+  trace('setupAmmServices', { farZoeKit });
+  if (!farZoeKit) {
+    farZoeKit = await setUpZoeForTest();
   }
-  const space = await setupAMMBootstrap(timer, zoe);
+  const feeMintAccess = await farZoeKit.feeMintAccess;
+  const zoe = await farZoeKit.zoe;
```
Chris-Hibbert: Is this trying to ensure that `zoe` is resolved and local? Does that matter once we call `space.produce.zoe.resolve(zoe)`?

packages/run-protocol/test/test-interchainPool.js
```diff
@@ -21,7 +21,8 @@ const contractRoots = {
 
 const makeTestContext = async () => {
   const bundleCache = await makeNodeBundleCache('bundles/', s => import(s));
-  const { zoe } = setUpZoeForTest();
+  const farZoeKit = await setUpZoeForTest();
+  const { zoe } = farZoeKit;
```
Chris-Hibbert, line 25: Is this an orthogonal fix, or is it somehow associated with the reserve changes?

Chris-Hibbert (commented):
Not a final review, just responses to questions, and initial reactions

packages/run-protocol/src/reserve/assetReserve.js
turadg: agreed it's an AMM concept. I thought being within method scoped names to AMM but I take your point. I do want to be clear that this is only for _secondary_ brands from AMM and not liquidity brands, so i'll go with `ammSecondaryBrand`.

packages/deploy-script-support/test/unitTests/test-startInstance.js
turadg: I don't know where it would go. I was burning down my problems pane.

Prefer that I take it out? would a separate commit in this PR with "merge" instead of squash suffice?

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: thanks. due to merge.

packages/run-protocol/test/test-interchainPool.js
turadg: I had to do a lot of bootstrap cleanup to get things loading in the right order. I don't remember whether this one was strictly necessary.

packages/run-protocol/test/amm/vpool-xyk-amm/setup.js
turadg: byproduct of debugging/refactoring. restored to `const { feeMintAccess, zoe } = farZoeKit;`.

in another case `feeMintAccess` had to be awaited and had been before.

packages/deploy-script-support/test/unitTests/test-startInstance.js
dckc: separate commit would be nice. don't take it out, is my preference

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: Yes. The "just-in-time" design had that. I thought you pressed for the caller having to set things up before sending liquidity to the reserve.

I think it's moot now as I've done the plumbing to make the AMM have the reserve early enough. I think it's better overall as it ensures a "valid" state of the contract earlier.

packages/run-protocol/src/reserve/assetReserve.js
turadg: >If addIssuerFromAMM is public, it has to verify that the brand is a secondary brand (not a liquidity brand) with the AMM before adding it. Otherwise, someone could bollix up our careful plans by calling addIssuerFromAmm(doubloonLiquidityBrand).

Had to read this a few times. Repeating what I got:
- this design has `addIssuerFromAMM` public which means it has to verify the brand argument is a secondary brand with the AMM before adding it
- the line `const issuer = await E(ammPublicFacet).getIssuer(ammSecondaryBrand);` can return a liquidity brand
- therefore someone could bollix up our careful plans by calling addIssuerFromAmm(doubloonLiquidityBrand).

> Why doesn't it work to add both issuers when someone (the AMM?) calls addIssuerFromAmm?

IIRC, the liquidity issuer wasn't yet defined on the AMM so the `E(ammPublicFacet).getLiquidityIssuer` call would fail. 

Let's chat about whether we can simplify this.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -295,6 +371,15 @@ const start = async (zcf, privateArgs) => {
   const creatorFacet = makeGovernorFacet(
     Far('AMM Fee Collector facet', {
       makeCollectFeesInvitation,
+      /**
+       * Must be called before adding pools. Not provided at contract start time due to cyclic dependency.
+       *
+       * @param {AssetReservePublicFacet} facet
+       */
+      setReserveDepositFacet: facet => {
+        // FIXME more than deposits
```
turadg: for now how about simply `setReservePublicFacet`?

packages/run-protocol/test/reserve/test-reserve.js
```diff
@@ -206,7 +206,8 @@ test('governance add Liquidity to the AMM', async t => {
     'should be 80K',
   );
 
-  await E(reserve.reserveCreatorFacet).addIssuer(moolaR.issuer, 'Moola');
+  // FIXME handle repeats
```
turadg: `conjureKeyword` seems to work well. I'll remove this comment in next push if you agree.

Chris-Hibbert:
I added **closes: #5271** in the top comment.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -51,9 +53,22 @@ const start = async (zcf, privateArgs) => {
...
    * Used to look up the brands for keywords, excluding RUN because it's a special case.
    *
-   * @type {MapStore<Keyword, [brand: Brand, liquidityBrand: Brand]>}
+   * @type {MapStore<Keyword, Brand>}
    */
   const brandForKeyword = makeStore('brands');
+  /**
+   * @type {MapStore<Brand, Brand>}
+   */
+  const liquidityBrandForBrand = makeStore('liquidityBrands');
```
Chris-Hibbert, line 62: The argument to `makeStore()` is supposed to describe the **key**, not the **value**.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -77,20 +92,82 @@ const start = async (zcf, privateArgs) => {
...
+      brand,
+    });
+
+    saveBrandKeyword(brand, keyword);
+    await zcf.saveIssuer(issuer, keyword);
+  };
+
+  /**
+   * @param {Issuer} baseIssuer on which the liquidity issuer is based
+   */
```
Chris-Hibbert, line 112:
```suggestion
  /** @param {Issuer} issuer on which the liquidity issuer is based */
```

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -295,6 +371,15 @@ const start = async (zcf, privateArgs) => {
...
+        reserveDepositFacet = facet;
```
Chris-Hibbert: I think this should assert `!reserveDepositFacet` before setting.

I'd prefer that this facet was visible via governance, but failing that, it should be set-once.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -35,20 +41,30 @@ export const makeAddIssuer = (zcf, isInSecondaries, brandToLiquidityMint) => {
...
+    const issuer = zcf.getIssuerForBrand(secondaryBrand);
+    console.log(
+      'Saved issuer',
+      secondaryIssuer,
+      'to keyword',
+      keyword,
+      'and got back',
+      issuer,
+    );
+    // this ensures that getSecondaryIssuer will return
```
Chris-Hibbert:
```suggestion
    // this ensures that getSecondaryIssuer(thisIssuer) will return even before the pool is created
```

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -35,20 +41,30 @@ export const makeAddIssuer = (zcf, isInSecondaries, brandToLiquidityMint) => {
...
+    brandToLiquidityMint.init(secondaryBrand, mint);
+    const { issuer: liquidityIssuer } = mint.getIssuerRecord();
+    // ??? still need this deferred lookup?
+    const addIssuerToReserve = getAddIssuerToReserve();
```
Chris-Hibbert: yes, I believe so.  It also makes it simpler to transition to a governed parameter if we decide to go that way.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -179,14 +248,20 @@ const start = async (zcf, privateArgs) => {
...
     params,
     protocolSeat,
-    reserveLiquidityTokenSeat,
     secondaryBrandToLiquidityMint,
-    updateMetrics,
+    handlePoolAdded,
   );
+  /**
+   *
+   */
```
Chris-Hibbert: ???

packages/run-protocol/test/amm/vpool-xyk-amm/setup.js
```diff
@@ -13,35 +13,49 @@ import {
...
+ */
 export const setupAMMBootstrap = async (
   timer = buildManualTimer(console.log),
-  zoe,
+  farZoeKit,
 ) => {
-  if (!zoe) {
-    ({ zoe } = await setUpZoeForTest());
+  if (!farZoeKit) {
+    farZoeKit = await setUpZoeForTest();
```
Chris-Hibbert, line 56: Can this be simplified back out, too? [I thought you said similar code was added elsewhere to assist in debugging, but was no longer useful.]

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -141,13 +142,17 @@ const setupAmmAndElectorate = async (t, aethLiquidity, runLiquidity) => {
   const space = setupBootstrap(t, timer);
   const { consume, instance } = space;
   installGovernance(zoe, space.installation.produce);
+  // TODO
```
Chris-Hibbert: TODO **what?**

turadg:
> I added **closes: #5271** in the top comment.

If that auto-closes it may look like "Reserve's addIssuer … public" is the state of master. Since we decided to take another approach, would it be better to close that with a comment on why?

Chris-Hibbert (changes requested):
Getting close. A few minor requests for changes.

packages/run-protocol/test/amm/vpool-xyk-amm/setup.js
turadg: I don't think so. In this case the need for `farZoeKit` had to be wired through.

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: hehe. I remember that was for DRYing out the next two lines. I plan to use https://github.com/Agoric/agoric-sdk/blob/358c8b7bef2c9a659344d72fd52aa80f14a93b68/packages/run-protocol/test/supports.js#L163-L167

packages/run-protocol/src/reserve/assetReserve.js
turadg: `baseIssuer` isn't part of the sentence. Here's how it looks in an IDE:
Screen Shot 2022-05-27 at 6 41 52 AM

If you're proposing to rename the parameter, I think that would make the body less clear. I could add a dash to separate, but I think that's a style decision we should perform universally, not ad hoc in PR review. I could make it `baseIssuer issuer on which` if you think that would be better.

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: I'll save that for another PR. WDYT of it @dckc ?

packages/run-protocol/src/reserve/assetReserve.js
turadg: not sure how to distinguish then between two maps with the same keys but different values.

leaving for now since the maps in master describe the value, not the keyword https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/reserve/assetReserve.js#L50-L56

packages/deploy-script-support/test/unitTests/test-startInstance.js
turadg: pulled non-run-protocol changes out to https://github.com/Agoric/agoric-sdk/pull/5440/commits/66fffd94c124a453a59ca8881981a0d47a976714 so I can squash this PR.

packages/run-protocol/test/vaultFactory/test-liquidator.js
dckc: it's a reasonable idiom... I'm not sure it carries its weight as an exported function, where a reader has to switch context to learn what it does.

Presuming `const { entries } = Object`, it's a one-liner:

```js
  entries(installations).forEach(([key, installation]) => space.installation.produce[key].resolve(installation));
```

I don't have much of a problem making the function available, but I think there's a significant risk that  folks will open-code it rather than remember that it's available as a helper function.

The helper function should take just the `space.installation` part, not the whole `space`. POLA.

Chris-Hibbert:
>> I added closes: https://github.com/Agoric/agoric-sdk/issues/5271 in the top comment.

> If that auto-closes it may look like "Reserve's addIssuer … public" is the state of master. Since we decided to take another approach, would it be better to close that with a comment on why?

I don't have strong opinions. It's clear that merging this, makes that issue moot. Feel free to remove the `closes:` marker. I justs wanted to remember to close the other.

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: stet

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
```diff
@@ -315,3 +322,63 @@ test('amm add and remove liquidity', async t => {
...
+
+  // Set up central token
+  const centralR = makeIssuerKit('central');
+  const moolaR = makeIssuerKit('moola');
+  const moola = value => AmountMath.make(moolaR.brand, value);
+  const central = value => AmountMath.make(centralR.brand, value);
+
+  const electorateTerms = { committeeName: 'EnBancPanel', committeeSize: 3 };
+  // This timer is only used to build quotes. Let's make it non-zero
+  const timer = buildManualTimer(console.log, 30n);
```
Chris-Hibbert: When the timer won't be used again in the current scope, I prefer to initialize it inline in the call (line 344) so as not to add its name to the local scope. In this case, there's a default value, so it's just as good to let it default, so it's never visible here. `electorateTerms` can also be moved into the call.

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -288,10 +291,10 @@ const setupServices = async (
...
     installation: { produce: iProduce },
   } = space;
-  // make the installation available for setupReserve
-  iProduce.reserve.resolve(t.context.installation.reserve);
-  // produce the reserve instance in the space
-  await setupReserve(space);
+  // // make the installation available for setupReserve
+  // iProduce.reserve.resolve(t.context.installation.reserve);
+  // // produce the reserve instance in the space
+  // await setupReserve(space);
```
Chris-Hibbert: ???

Chris-Hibbert (approved):
Minor comments. I trust your judgement on whether to address them or leave them as is.

packages/run-protocol/src/reserve/assetReserve.js
turadg: TIL stet!

packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -267,7 +267,6 @@ export const setupReserve = async ({
   },
 }) => {
   trace('setupReserve');
-  console.trace('setupReserve');
```
dckc: ah. that makes sense.

Perhaps the `console` provided by SwingSet doesn't have `.trace`.

turadg: ahh

Issue #5299, opened 2022-05-05
Reserve contract covers vault shortfalls


rowgraus:
## What is the Problem Being Solved?
The Inter Protocol relies on successful liquidation to ensure the stable token remains over collateralized.  In cases where individual liquidations do not cover the outstanding debt for any reason, the protocol must track the outstanding balance so that it can be made whole through later operations.  This ticket allows the Reserve contract to both track the ongoing shortfall and to burn IST to reduce it.

## Description of the Design
Shortfalls that occur during liquidation are passed to the reserve contract which manages a shortfall balance.  This logic must survive as liquidation contract is changed.

The reserve contract must burn IST to reduce shortfall.  Suggest adding method that can be called by governance which takes IST from the Reserve's IST purse to handle this.

Reserve's IST purse is funded through execution fees and excess reward flows as described in #4399, but it must also be able to accept additional ad-hoc deposits as needed.  I.e., governance could decide to sell other assets to and fund the Reserve IST purse for the purpose of burning to reduce shortfall.

## Security Considerations
The liquidation shortfall which occurs during execution of the liquidation contract within the vault manager must be passed correctly to the reserve to ensure shortfalls cannot fail to be tracked

rowgraus:
Econ committee abilities which allow it to let the reserve manage the shortfall

Governance: Burn IST from purse and reduce debt
Governance: Redeem liquidity tokens 
Governance: Sell on AMM and deposit into IST pool
Governance: Take collateral in collateral purse, mint IST, deposit

dckc:
> Governance: Take collateral in collateral purse, mint IST, deposit

sounds like #5298

PR #5431, opened 2022-05-25
feat: Reserves track liquidation shortfall

Chris-Hibbert:
refs: #5299

## Description

Add the ability for Vaults to report to the Reserve on shortfalls during liquidation.

### Security Considerations

Uses governance.

More notification streams (for the shortfall level and Reserve pools.)

### Testing Considerations

I had to disable the SwingSets in run-protocol, since I didn't see a way to initialize the Reserve compatibly with other services.


packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -95,6 +96,36 @@ const start = async (zcf, privateArgs) => {
   /** @type {ZCFMint} */
   const runMint = await zcf.registerFeeMint('RUN', feeMintAccess);
 
+  const { brand: runBrand } = await E(runMint).getIssuerRecord();
+  const { publication: metricsPublication, subscription: metricsSubscription } =
+    makeSubscriptionKit();
+
+  // shortfall in Vaults due to liquidations less than debt. This value can be
+  // reduced by various actions which burn IST.
```
turadg: until we rename, might be more confusing to refer to it in two ways
```suggestion
  // reduced by various actions which burn RUN.
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -95,6 +96,36 @@ const start = async (zcf, privateArgs) => {
...
+  const updateMetrics = () => {
+    const metrics = harden({
+      allocations: getAllocations(),
+      shortfall: vaultShortfall,
+    });
+    metricsPublication.updateState(metrics);
+  };
+  updateMetrics();
+
+  const addLiquidationShortfall = shortfall => {
```
turadg: for parity with `reduce…`
```suggestion
  const increaseLiquidationShortfall = shortfall => {
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -182,12 +213,29 @@ const start = async (zcf, privateArgs) => {
     zcf.reallocate(offerToSeat, collateralSeat);
   };
 
+  const getShortfallReportInvitation = () => {
+    const shortfallReportOffer = () => {
```
turadg: non-blocking suggestion for consistency https://github.com/Agoric/agoric-sdk/pull/5179/files#diff-46a4c9227ae419f47fed07515b2459edf285ccf8aee937434e3471b2ea6cabb6

```suggestion
  const maketShortfallReportInvitation = () => {
    const handleShortfallReportOffer = () => {
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -182,12 +213,29 @@ const start = async (zcf, privateArgs) => {
...
+      return Far('shortfallReporter', {
+        addLiquidationShortfall,
+        // currently exposed for testing. Maybe it only gets called internally?
+        reduceLiquidationShortfall,
```
turadg, line 230: if it's not part of the API, let's mark it as such so it's not a breaking change to remove.

do we have a convention for expressing that? e.g. underscore prefix

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -404,6 +404,7 @@ const helperBehavior = {
       liquidator,
       state.collateralBrand,
       factoryPowers.getGovernedParams().getLiquidationPenalty(),
+      factoryPowers.getShortfallReporter(),
```
turadg: noting from our conversation to [get the numbers out](https://github.com/Agoric/agoric-sdk/pull/5393/files#diff-6270d15d5734246fb29efbc9d2589a31d6c4feca1aa5361843a720689fd797bfR99-R100) instead of giving `liquidate` more dependencies

packages/run-protocol/test/reserve/test-reserve.js
```diff
@@ -338,3 +339,55 @@ test('request more collateral than available', async t => {
...
+    AmountMath.make(runBrand, 1000n),
+  );
+  const metricsSub = await E(reserve.reserveCreatorFacet).getMetrics();
+  const m = await subscriptionTracker(t, metricsSub);
+  await m.assertInitial({
+    allocations: {},
+    shortfall: AmountMath.makeEmpty(runBrand),
+  });
+  await m.assertChange({
+    shortfall: { value: 1000n },
```
turadg: nice and clean expressing what has changed.

please keep a running value and adjust it. e.g.,

```
  await E(reporterFacet).addLiquidationShortfall(
    AmountMath.make(runBrand, 500n),
  );
  totalShortfall += 500n;
  await m.assertChange({
    shortfall: { value: totalShortfall },
  });
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -118,6 +119,36 @@ const start = async (zcf, privateArgs) => {
   const makeAddCollateralInvitation = () =>
     zcf.makeInvitation(addCollateralHook, 'Add Collateral');
 
+  const { brand: runBrand } = await E(runMint).getIssuerRecord();
+  const { publication: metricsPublication, subscription: metricsSubscription } =
```
turadg, line 132: please type the state with documentation. [example](https://github.com/Agoric/agoric-sdk/pull/5393/files#diff-fe43c0eebf753f26efd39077f96e11f163a1da8b42fcf5409eb39ec10a2d5924R38-R49)

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -118,6 +119,36 @@ const start = async (zcf, privateArgs) => {
...
+    makeSubscriptionKit();
+
+  // shortfall in Vaults due to liquidations less than debt. This value can be
+  // reduced by various actions which burn IST.
+  let vaultShortfall = AmountMath.makeEmpty(runBrand);
+
+  const updateMetrics = () => {
+    const metrics = harden({
+      allocations: getAllocations(),
+      shortfall: vaultShortfall,
```
turadg: for naming, `shortfall` sounds like an event metric but this is a sum. consider `totalShortfall`

turadg (dismissed):

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/assetReserve.js

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: from #5299, I chose `shortfallBalance`

packages/run-protocol/src/reserve/assetReserve.js

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: I changed to `makeShortfallReportingInvitation` and `handleShortfallReportingOffer`, but that still has `make` and `handle`.

It's not so much an invitation about reports, as it's about reporting.

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: in Java, we had @visibleForTesting, which the tools supported. JavaScript doesn't seem to have an equivalent.   AFAICT there's no convention here.

I'm not positive that it shouldn't be possible to call externally. I can imagine actions that the econ committee takes that address a shortfall (by burning somewhere else) that doesn't involve the Reserve. We want them to be able to update the accounting when they do that.

packages/run-protocol/test/reserve/test-reserve.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/assetReserve.js
turadg: added 8f3af1bdd so the `getMetrics()` result has the type too

packages/run-protocol/test/swingsetTests/setup.js
```diff
@@ -202,19 +203,23 @@ const buildOwner = async (
   /** @type {AssetReservePublicFacet} */
   // @ts-expect-error cast, never used
   const reservePublicFacet = null;
+  /** @type {Amount} */
+  // @ts-expect-error cast, never used
```
turadg, line 207: the comment "tests are currently disabled" was confusing to me at first: what tests.

in 8f3af1bdd I replaced using the other comment. as I write this I suspect it's not as true in this case, but either the whole suite is going away or when it's reinstated this comment won't matter much

turadg (approved):

PR #5432, opened 2022-05-25
feat: Add a helper for publishing iteration results to chain storage

### LINKED ISSUE #5353: Support publishing to chainStorage-backed subscriptions

On-chain vats need to publish sequential values for off-chain consumers.

## Description of the Design

Create a function that accepts a chainStorage node (cf. #4558) and returns an updater/publication object (cf. https://agoric.com/documentation/guides/js-programming/notifiers.html ) such that writes end up in chain storage. Since such writes are committed only at block boundaries, It will probably be necessary to invent a container format communicating multiple results in a single chainStorage string (and possibly a way to opt out of that for publishers whose consumers are guaranteed to tolerate loss of intermediate state).

## Security Considerations

We can control the publish _capability_, but all data written to chain storage is inherently public.

gibson042:
closes: #5353
refs: #5385 ([diff](https://github.com/Agoric/agoric-sdk/compare/gibson-4558-chainStorage...gibson-5353-publish-to-chainStorage))

## Description

Adds a `publishToChainNode` helper for consuming iteration results and losslessly publishing them to chain storage.

### Security Considerations

We should be mindful about publishing too much data, but some redundancy seems necessary to avoid loss.

Nothing specific.

### Testing Considerations

TypeScript gave me some trouble with mocks. :confused:

### DIFF
--- a/packages/vats/src/lib-chainPublish.js
+++ b/packages/vats/src/lib-chainPublish.js
@@ -10,8 +10,8 @@ import { E } from '@endo/far';
  * Array items are possibly-duplicated [index, value] pairs, where index is a
  * string (ascending numeric or terminal "finish" or "fail").
  *
- * @param {AsyncIterator} iterable
- * @param {ReturnType<typeof import('./lib-chainStorage.js').makeChainStorageRoot>} chainStorageNode
+ * @param {ERef<AsyncIterator>} iterable
+ * @param {{ setValue: (val: any) => void }} chainStorageNode
  * @param {{ timerService: ERef<TimerService>, serialize?: (obj: any) => string }} powers
  */
 export async function publishToChainNode(
--- a/packages/vats/src/lib-chainPublish.js
+++ b/packages/vats/src/lib-chainPublish.js
@@ -10,12 +10,12 @@ import { E } from '@endo/far';
  * Array items are possibly-duplicated [index, value] pairs, where index is a
  * string (ascending numeric or terminal "finish" or "fail").
  *
- * @param {ERef<AsyncIterator>} iterable
+ * @param {ERef<AsyncIterable>} source
  * @param {{ setValue: (val: any) => void }} chainStorageNode
  * @param {{ timerService: ERef<TimerService>, serialize?: (obj: any) => string }} powers
  */
 export async function publishToChainNode(
-  iterable,
+  source,
   chainStorageNode,
   { timerService, serialize = JSON.stringify },
 ) {
@@ -49,7 +49,7 @@ export async function publishToChainNode(
       E(chainStorageNode).setValue(serialize(combined));
     };
   };
-  await observeIteration(iterable, {
+  await observeIteration(source, {
     updateState: makeAcceptor(),
     finish: makeAcceptor('finish'),
     fail: makeAcceptor('fail'),
--- a/packages/vats/src/lib-chainPublish.js
+++ b/packages/vats/src/lib-chainPublish.js
@@ -1,11 +1,14 @@
 // @ts-check
 
+import { observeIteration } from '@agoric/notifier';
 import { E } from '@endo/far';
 
 /**
  * Publish results of an async iterable into a chain storage node as an array
  * serialized using JSON.stringify or an optionally provided function (e.g.,
  * leveraging a serialize function from makeMarshal).
+ * Array items are possibly-duplicated [index, value] pairs, where index is a
+ * string (ascending numeric or terminal "finish" or "fail").
  *
  * @param {AsyncIterator} iterable
  * @param {ReturnType<typeof import('./lib-chainStorage.js').makeChainStorageRoot>} chainStorageNode
@@ -20,27 +23,35 @@ export async function publishToChainNode(
   let isNewBlock = true;
   let oldResults = [];
   let results = [];
-  for await (const result of iterable) {
-    if (isNewBlock) {
-      isNewBlock = false;
-      oldResults = results;
-      results = [];
-      E(timerService)
-        .delay(1n)
-        .then(() => {
-          isNewBlock = true;
-        });
-    }
-    // To avoid loss when detecting the new block *after* already consuming
-    // results produced within it, we associate each result with an index and
-    // include "old" results in the batch.
-    // Downstream consumers are expected to deduplicate by index.
-    // We represent the index as a string to maintain compatibility with
-    // JSON.stringify and to avoid overly inflating the data size (e.g. with
-    // "@qclass" objects from makeMarshal serialize functions).
-    results.push([String(nextIndex), result]);
-    nextIndex += 1n;
-    const batch = harden(oldResults.slice().concat(results));
-    E(chainStorageNode).setValue(serialize(batch));
-  }
+  const makeAcceptor = forceIndex => {
+    return value => {
+      if (isNewBlock) {
+        isNewBlock = false;
+        oldResults = results;
+        results = [];
+        E(timerService)
+          .delay(1n)
+          .then(() => {
+            isNewBlock = true;
+          });
+      }
+      // To avoid loss when detecting the new block *after* already consuming
+      // results produced within it, we associate each result with an index and
+      // include "old" results in the batch.
+      // Downstream consumers are expected to deduplicate by index.
+      // We represent the index as a string to maintain compatibility with
+      // JSON.stringify and to avoid overly inflating the data size (e.g. with
+      // "@qclass" objects from makeMarshal serialize functions).
+      const index = forceIndex || String(nextIndex);
+      nextIndex += 1n;
+      results.push([index, value]);
+      const batch = harden(oldResults.slice().concat(results));
+      E(chainStorageNode).setValue(serialize(batch));
+    };
+  };
+  await observeIteration(iterable, {
+    updateState: makeAcceptor(),
+    finish: makeAcceptor('finish'),
+    fail: makeAcceptor('fail'),
+  });
 }
--- a/packages/vats/src/lib-chainPublish.js
+++ b/packages/vats/src/lib-chainPublish.js
@@ -37,7 +37,7 @@ export async function publishToChainNode(
       }
       // To avoid loss when detecting the new block *after* already consuming
       // results produced within it, we associate each result with an index and
-      // include "old" results in the batch.
+      // include results of the previous batch.
       // Downstream consumers are expected to deduplicate by index.
       // We represent the index as a string to maintain compatibility with
       // JSON.stringify and to avoid overly inflating the data size (e.g. with
@@ -45,8 +45,8 @@ export async function publishToChainNode(
       const index = forceIndex || String(nextIndex);
       nextIndex += 1n;
       results.push([index, value]);
-      const batch = harden(oldResults.slice().concat(results));
-      E(chainStorageNode).setValue(serialize(batch));
+      const combined = harden(oldResults.slice().concat(results));
+      E(chainStorageNode).setValue(serialize(combined));
     };
   };
   await observeIteration(iterable, {
--- a/packages/vats/src/lib-chainPublish.js
+++ b/packages/vats/src/lib-chainPublish.js
@@ -1,0 +1,46 @@
+// @ts-check
+
+import { E } from '@endo/far';
+
+/**
+ * Publish results of an async iterable into a chain storage node as an array
+ * serialized using JSON.stringify or an optionally provided function (e.g.,
+ * leveraging a serialize function from makeMarshal).
+ *
+ * @param {AsyncIterator} iterable
+ * @param {ReturnType<typeof import('./lib-chainStorage.js').makeChainStorageRoot>} chainStorageNode
+ * @param {{ timerService: ERef<TimerService>, serialize?: (obj: any) => string }} powers
+ */
+export async function publishToChainNode(
+  iterable,
+  chainStorageNode,
+  { timerService, serialize = JSON.stringify },
+) {
+  let nextIndex = 0n;
+  let isNewBlock = true;
+  let oldResults = [];
+  let results = [];
+  for await (const result of iterable) {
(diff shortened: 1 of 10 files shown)


packages/vats/src/lib-chainPublish.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Publish results of an async iterable into a chain storage node as an array
+ * serialized using JSON.stringify or an optionally provided function (e.g.,
+ * leveraging a serialize function from makeMarshal).
+ *
+ * @param {AsyncIterator} iterable
+ * @param {ReturnType<typeof import('./lib-chainStorage.js').makeChainStorageRoot>} chainStorageNode
+ * @param {{ timerService: ERef<TimerService>, serialize?: (obj: any) => string }} powers
+ */
+export async function publishToChainNode(
```
gibson042, line 17: This was originally envisioned as a function that accepts a chainStorage node and returns a subscription kit, but upon digging in it seemed more useful to me for it to consume from an existing iterable and return nothing. Still another option would be to make it effectively a tee, but I think it makes more sense to have the caller compose e.g. `getKey()` and the base iterable.

packages/vats/src/lib-chainPublish.js
```diff
@@ -0,0 +1,46 @@
...
+  for await (const result of iterable) {
+    if (isNewBlock) {
+      isNewBlock = false;
+      oldResults = results;
+      results = [];
+      E(timerService)
+        .delay(1n)
+        .then(() => {
+          isNewBlock = true;
+        });
```
gibson042: Is this the right/best way to detect a new block?

packages/vats/src/lib-chainPublish.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+    // To avoid loss when detecting the new block *after* already consuming
+    // results produced within it, we associate each result with an index and
+    // include "old" results in the batch.
+    // Downstream consumers are expected to deduplicate by index.
```
gibson042: I didn't see a way around this, because there's no way to know when the timerService delay resolves (or at any other time, for that matter) which results have actually been published in a block vs. which are still pending.

gibson042 (commented):
Broad question: is packages/vats/src/lib-chainPublish.js the right place for this code, or should it be in some other package?

packages/vats/src/lib-chainPublish.js
dckc: There's precedent for it. I'm not sure that it's guaranteed to detect a new block, but I think it's pretty close in practice. Does anything depend on this happening exactly once per block?

dckc (commented):
I'm used to new **feat**ures coming with tests. Is there a reason to skip it here? Is it cost-prohibitive?

gibson042:
No, I just wanted to get this up for review ASAP. I'm adding tests now.

packages/vats/test/test-lib-chainPublish.js
```diff
@@ -0,0 +1,166 @@
...
+    });
+  };
+  const timerService = {
+    delay(n) {
+      t.is(n, 1n);
+      return advanceTimerP;
+    },
+  };
+
+  // Create an async iterable with a function for supplying results.
```
gibson042, line 314: Is there any existing code for creating test doubles, particularly timerService and/or an async iterable that gets fed values asynchronously?

gibson042 (commented):
@dckc Basic tests added, working on advanced tests now.

packages/vats/test/test-lib-chainPublish.js
```diff
@@ -0,0 +1,166 @@
+// @ts-check
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
+
+import { E } from '@endo/far';
+import { publishToChainNode } from '../src/lib-chainPublish.js';
+
+test('publishToChainNode', async t => {
+  // eslint-disable-next-line no-use-before-define
```
dckc, line 9: any particular reason to break from convention here?

packages/vats/test/test-lib-chainPublish.js
```diff
@@ -0,0 +1,166 @@
...
+    customAsyncIterable,
+    // internal facets
+    nodeValueHistory,
+    advanceTimer,
+    supplyIterationResult,
+  } = inputs;
+  publishToChainNode(customAsyncIterable, storageNode, { timerService });
+
+  supplyIterationResult('foo');
+  await E(Promise).resolve();
```
dckc: :exploding_head:

packages/vats/src/lib-chainPublish.js
michaelfig: Detecting a new block should not be done in js, rather this functionality belongs in Golang.

packages/vats/src/lib-chainPublish.js
michaelfig: I don't think this is the right abstraction.  It crosses layers and combines concerns in a muddy way.  What are your trying to improve wrt the sketch in #5400?

packages/vats/src/lib-chainPublish.js
michaelfig: We have control of the chain behaviour.  We should solve it there.

packages/vats/src/lib-chainStorage.js
```diff
@@ -0,0 +1,70 @@
...
+        );
+        return makeChainStorageNode(`${key}.${name}`);
+      },
+      setValue(value) {
+        assert.typeof(value, 'string');
+        toStorage({ key, method: 'set', value });
+      },
+      async delete() {
+        assert(key !== rootKey);
+        // A 'set' with no value deletes a key if it has no children, but
```
michaelfig, line 41: This issue has been resolved in Golang.  Please remove the race and make this a direct wrapper.

packages/vats/src/lib-chainStorage.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @param {(message: any) => any} toStorage a function for sending a storageMessage object to the storage implementation (cf. golang/cosmos/x/swingset/storage.go)
+ * @param {string} rootKey
+ */
+export function makeChainStorageRoot(toStorage, rootKey) {
+  assert.typeof(rootKey, 'string');
+
+  function makeChainStorageNode(key) {
+    const node = {
+      getKey() {
```
michaelfig, line 24: Good!  The composer in #5400 can pass this method through directly.

michaelfig (changes requested):

michaelfig:
> Broad question: is packages/vats/src/lib-chainPublish.js the right place for this code, or should it be in some other package?

The middleware belongs here, but the composer used by contracts belongs in `packages/notifier`.

packages/vats/src/lib-chainPublish.js
gibson042: What layers do you think are being crossed and what concerns are being combined? Transmitting arbitrary data into a chainStorage node seems very respectful of layers and independent concerns. As for #5400, I looked it over carefully but didn't see anything to address the specific requirements of publishing to chain storage... am I just missing something?

packages/vats/src/lib-chainPublish.js
gibson042: > Does anything depend on this happening exactly once per block?

@dckc No, but it would be a problem for this to happen _more_ than once per block because then results produced earlier in the block could be lost.

> this functionality belongs in Golang

@michaelfig How exactly would that work, given that many vats will have chain publishers that need to purge old results? I guess we could create a new vat that somehow gets end-of-block messages and is the sole source of subscriptions that publish to chain storage, but a) I'd need help with the first part, and b) I'm not sure using such a vat would actually be an improvement over using `timerService` (although I'm willing to believe it might be—it would really help to see the code that actually needs chainStorage-backed subscriptions rather than guessing from the bottom up like this).

packages/vats/src/lib-chainPublish.js
gibson042: Could you elaborate on how?

packages/vats/src/lib-chainStorage.js
gibson042: I don't think we _want_ to allow deleting a non-leaf node—IIUC, @warner and I agree that silent failure (by means of replacement with "set data to empty") seems like bad behavior that could leave orphans. But at any rate, this is a comment for #5385 rather than here.

packages/vats/test/test-lib-chainPublish.js
gibson042: I find having to slog through a test-double library before getting to the actual tests makes the file much less readable. And the point will be moot if the functionality of `getMockInputs` can be moved into a test-helpers module.

packages/vats/test/test-lib-chainPublish.js
gibson042: Yeah, I needed a sufficient delay to match the code under test and didn't like the approach of appending `.then()`s to something like `Promise.resolve()` until I observed progress.

packages/vats/test/test-lib-chainPublish.js
dckc: > I find having to slog through a test-double library before getting to the actual tests makes the file much less readable.

That sounds like you disagree with the convention, not like there is some reason to break from it in this case. I appreciate the appeal of successive elaboration (I wrote python that way for decades...) but `no-use-before-define` is not only something we inherited from airbnb style; it's a considered opinion by @erights and co that goes back at least as far as 2004: [The Power of Irrelevance](http://erights.org/data/irrelevance.html).

packages/vats/test/test-lib-chainPublish.js
dckc: There's a manual timer in `packages/zoe` somewhere. I'm not familiar with "test doubles" so I'm not sure how relevant that is.

packages/vats/test/test-lib-chainPublish.js
gibson042: Yes, I disagree with the convention in general, and here in particular strongly enough to explicitly deviate from it. But I'll change this code to conform with it if you think I should, and will read the essay regardless.

packages/vats/test/test-lib-chainPublish.js
gibson042: "Test double" is the general term encompassing mocks/spies/stubs/etc.—implementations used as stand-ins to facilitate testing: https://en.wikipedia.org/wiki/Test_double

packages/vats/src/lib-chainStorage.js
michaelfig: As I mentioned in https://github.com/Agoric/agoric-sdk/pull/5385#discussion_r876523826 the problem is solved and merged.  I think you and @warner may be misunderstanding the intended semantics (implemented by #5394):

- nodes with empty data are not stored in the tree, and unreferenced path linkage is cleared when emptying a node
- path linkage of nodes with nonempty data is always present, and when an empty node is filled any missing path linkage is built

The linkage information is just used to enumerate children.  To demonstrate the separation between linkage and data, that linkage information is kept in `swingset/path` Cosmos IAVL key prefix, unlike the `swingset/data` key prefix.

Please remove this racy workaround.  It is unnecessary, and the only reason I'm requesting changes.

packages/vats/src/lib-chainPublish.js
michaelfig: > What layers do you think are being crossed and what concerns are being combined?

I guess what jumps out at me looking at the helper is it is trying to combine the iterator observation with the block batching.  I think that the block batching belongs in the cosmic-swingset layer, and the iteration client is handled by #5400.

> Transmitting arbitrary data into a chainStorage node seems very respectful of layers and independent concerns.

You're right that it needs to be done at a lower layer than #5400, but it should be even lower than JS entirely.  I'm trying hard to "put the code where the data is".  Since it is a cosmic-swingset concern (touching block height, event log, and merkle tree), it belongs outside of vats.  I appreciate you're trying to respect peoples' time, but I would really like it if you would reach out if something about the layering doesn't make sense or seems unnecessarily weird (e.g. the chainTimer hack).

> As for #5400, I looked it over carefully but didn't see anything to address the specific requirements of publishing to chain storage... am I just missing something?

Here is an adapter that explains how #5400 accomplishes these platform concerns, presuming that the batching is solved at the cosmic-swingset layer:

```js
// adapt chainStorageNode to makeMarshalSubscriber
const publisher = Far('chainPublisherNode', {
  publish: jsonable => E(chainStorageNode).setValue(JSON.stringify(jsonable)),
  getPublisherDetails: () => { chainStorageKey: E(chainStorageNode).getKey() },
});

// Then something like:
makeMarshalSubscriber(observeEach(subscription), publisher);
```

Is that clearer now?  I'll be working on splitting `x/pubstore` Golang module from `x/swingset`.  I want to implement cosmos-specific publishing concerns there, such as batching, history, and event logging.

With some polish and moving the code around until we have minimal interfaces, I think this could be quite good.

michaelfig (changes requested):
I think `lib-chainPublish.js` should be done in Golang, but otherwise this PR looks good.

... except I am requesting changes to remove the non-robust and unnecessary child deletion test.

michaelfig (approved):
After discussing with @gibson042, I'm happy with the direction we settled on.  I'm removing my request for changes.

gibson042:
This is a draft pending completion of [makePublishKit](https://github.com/Agoric/agoric-sdk/pull/5435). Once done, it will be updated to export a function from the notifier package that accepts a publish-kit subscriber and a [chain] storage node and returns an object exposing both the subscriber interface (for getting iteration results with configurable lossiness) and the node's storage key (for supporting off-chain consumption).

michaelfig:
Hi @gibson042.  @dckc and I plumbed this feature through using #5400.

I'd appreciate it if you had a look at that PR, and if we're missing any features from `makeStoredSubscription`, to focus your efforts there.

PR #5434, opened 2022-05-25
refactor(vaults): rename VaultTitle to VaultHolder

turadg:
_no ticket_, decided in meeting and small enough to PR directly

## Description

A _title_ is metadata about ownership. E.g. a pink slip is something I can give you about ownership of the car, but it's not the car itself. Whereas the object wrapping a vault with ownership _can be used as a vault_. 

This [revokable authority with the caretaker pattern ](https://www.researchgate.net/figure/Revocable-authority-with-the-Caretaker-pattern_fig17_267401630) will come up again and we need a consistent name, but we're deferring that more general decision.

### Security Considerations

Clearer ocaps.

--

### Testing Considerations

No tests should be affected.


dtribble (approved):
Well that was easy! LGTM

PR #5436, opened 2022-05-26
feat: allow vats to be marked critical and panic the kernel if a critical vat fails

### LINKED ISSUE #4279: mark some vats as critical, panic the kernel if one dies

Currently, the only way for userspace code to panic the kernel is if the `bootstrap()` message (to the bootstrap vat) rejects its result promise. The kernel panics if a device throws an error, but devices are not really user code. The kernel also has a bunch of invariant checks that can cause a panic, but those are all internal and should not be provokable by userspace code.

However, in any given deployment, there are probably some userspace vats which are considered so critical that when they malfunction, we'd rather have the overall application crash than have the kernel kill off just the faulty vat. The comms and vat-vattp vats always fall into this category. In the Agoric chain, vat-zoe and all the vats that make up the token economy are chain-critical. It will be easier to recover a fault by halting the chain, than to have the kernel delete vat-zoe and reap all of the objects it has ever exported.

So the requirement is for some vats to be marked as "critical". If the kernel ever finds a reason to kill that vat (illegal syscall, metering fault, explicit `syscall.terminate`), the kernel should panic instead. This should cause `controller.run()` to reject its return promise instead of fulfilling it, and the host application should react by exiting immediately rather than committing the state changes.

## Description of the Design

The `vatAdminService~.createVat()` call should accept a `critical` option. If `true`, the kernel should panic instead of killing this vat. The kernel config object should also accept this flag, for static vats.

This will go into the `managerOptions`. In `kernel.js`, when a vat is about to be killed, it should check the flag, and `panic()` if true (with some details about which vat triggered the panic).

For now, we'll allow any call to `createVat` to set this flag, meaning that any code which can create vats can also panic the kernel. vat-zoe is the only chain-side vat which gets access to `vatAdminService`, and the power to panic the kernel is comparable to the authority to create unmetered vats, which `vatAdminService` also provides. Later, if we find a use case for it, we can have vatAdminService produce less-powerful facets. Or, riffing on an approach @erights suggested for Meters, the service could provide a special `criticality` object, and the only way to mark a vat as critical is to pass a `managerOptions` of `{ critical: criticalityObject }`, in a rights-amplification pattern that needs access to both `vatAdminService` and `criticality`. But for now, only zoe has access to `vatAdminService`, so we don't need to build anything too fancy.

## Security Considerations

Building a deliberately-crashing vat and marking it `critical` is a kernel-killing power, so we should be conscious how it is disseminated.

We should think carefully about potential bugs in critical vats and decide whether halting the host application really is the best approach. If multiple services can be severed from one another, then allowing some portion of a chain to keep running even though others are dead might be better. We should war-game a bug in a core service somehow.

## Test Plan

Unit tests

FUDCo:
Vats now accept a `critical` property as one of their creation options.  If a vat is marked as critical, failure of that vat causes a kernel panic.

In order to avoid allowing just anyone to set up a situation that could bring down the kernel and thus the chain, dynamic vats cannot be configured as critical simply by setting the `critical` option to true (as they can with static vats).  Instead, for the `critical` option value you must provide the "critical vat key", a capability that can only be obtained from the vatAdmin vat.  The vatAdmin root now supports an additional method, `getCriticalVatKey()` which returns this capability.  Note that this is only available on the vatAdmin root object, not on any vatAdminService object, and thus is initially only available to the bootstrap vat.

Fixes #4279

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -250,6 +250,8 @@ export default function buildKernel(
    * @param {SwingSetCapData} info
    */
   function terminateVat(vatID, shouldReject, info) {
+    const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
+    const critical = vatKeeper.getOptions().critical;
     insistCapData(info);
     // guard against somebody telling vatAdmin to kill a vat twice
     if (kernelKeeper.vatIsAlive(vatID)) {
@@ -262,6 +264,18 @@ export default function buildKernel(
       // ISSUE: terminate stuff in its own crank like creation?
       // eslint-disable-next-line no-use-before-define
       vatWarehouse.vatWasTerminated(vatID);
+    }
+    if (critical) {
+      // The following error construction is a bit awkward, but (1) it saves us
+      // from doing unmarshaling while in the kernel, while (2) it protects
+      // against the info not actually encoding an error, but (3) it still
+      // provides some diagnostic information back to the host, and (4) this
+      // should happen rarely enough that if you have to do a little bit of
+      // extra interpretive work on the receiving end to diagnose the problem,
+      // it's going to be a small cost compared to the trouble you're probably
+      // already in anyway if this happens.
+      panic(`critical vat ${vatID} failed`, Error(info.body));
+      return;
     }
     if (vatAdminRootKref) {
       // static vat termination can happen before vat admin vat exists
@@ -306,7 +320,7 @@ export default function buildKernel(
     insistCapData(info);
     // if vatFatalSyscall was here already, don't override: bad syscalls win
     if (!terminationTrigger) {
-      terminationTrigger = { vatID, abortCrank: false, reject, info };
+      terminationTrigger = { vatID, abortCrank: reject, reject, info };
     }
   }
 
--- a/packages/SwingSet/test/vat-admin/terminate/test-terminate.js
+++ b/packages/SwingSet/test/vat-admin/terminate/test-terminate.js
@@ -16,18 +16,36 @@ test.before(async t => {
   t.context.data = { kernelBundles };
 });
 
-async function doTerminate(t, mode, reference, extraMessage = []) {
+async function doTerminateNonCritical(
+  t,
+  deadVatID,
+  mode,
+  dynamic,
+  reference,
+  extraMessage = [],
+) {
   const configPath = new URL('swingset-terminate.json', import.meta.url)
     .pathname;
   const config = await loadSwingsetConfigFile(configPath);
-  const controller = await buildVatController(config, [mode], t.context.data);
+  const hostStorage = provideHostStorage();
+  const controller = await buildVatController(config, [mode, false, dynamic], {
+    ...t.context.data,
+    hostStorage,
+  });
   t.is(controller.kpStatus(controller.bootstrapResult), 'unresolved');
+  const preProbe = hostStorage.kvStore.get(`${deadVatID}.options`);
+  if (!dynamic) {
+    t.truthy(preProbe);
+  }
+
   await controller.run();
   t.is(controller.kpStatus(controller.bootstrapResult), 'fulfilled');
   t.deepEqual(
     controller.kpResolution(controller.bootstrapResult),
     capargs('bootstrap done'),
   );
+  const staticDone = dynamic ? [] : ['done'];
+  const dynamicDone = dynamic ? [reference, 'done'] : [];
   t.deepEqual(controller.dump().log, [
     'FOO 1',
     'count1 FOO SAYS 1',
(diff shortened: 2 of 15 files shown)


packages/SwingSet/src/kernel/kernel.js
```diff
@@ -263,6 +265,18 @@ export default function buildKernel(
...
+    if (critical) {
+      // The following error construction is a bit awkward, but (1) it saves us
+      // from doing unmarshaling while in the kernel, while (2) it protects
+      // against the info not actually encoding an error, but (3) it still
+      // provides some diagnostic information back to the host, and (4) this
+      // should happen rarely enough that if you have to do a little bit of
+      // extra interpretive work on the receiving end to diagnose the problem,
+      // it's going to be a small cost compared to the trouble you're probably
+      // already in anyway if this happens.
+      panic(`critical vat failure ${vatID}`, Error(info.body));
```
warner: I might reword this slightly, to `critical vat ${vatID} failed`. The original could be read to mean "a vat failed in a particularly critical way", as opposed to "a vat, which was marked critical, has suffered an (ordinary) failure".

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -966,7 +980,7 @@ export default function buildKernel(
       const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
 
       // deliveries cause garbage, garbage needs collection
-      const { decrementReapCount } = status;
+      const { decrementReapCount, terminate } = status;
```
warner: nit: I'm -0 on merging the later destructuring up here, I guess it's a minor aesthetic thing. There are three sections here (`scheduleReap`, meter deduction, check termination), each one starts with unpacking the necessary bits of `status` and then interpreting it somehow. Unpacking `terminate` here but then not using it until the third clause makes it look (to me) that `terminate` is somehow needed for this first section, but it isn't.

(an admission: the only reason I used any destructuring at all is because that made the lines short enough to keep `prettier` from newlining the code into a bunch of uselessly short lines, making the otherall function no longer fit on a single page. If it wasn't going to do that, I'd have used `if (status.decrementReapCount && ...` instead)

Maybe move all the destructuring, from all three sections, up to sit right after `const vatKeeper = ..`? Except that having more than two or three things destructured at once will *also* cause `prettier` to splat expand the statement into a gratuitious number of lines. Can't win.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -306,7 +320,7 @@ export default function buildKernel(
     insistCapData(info);
     // if vatFatalSyscall was here already, don't override: bad syscalls win
     if (!terminationTrigger) {
-      terminationTrigger = { vatID, abortCrank: false, reject, info };
+      terminationTrigger = { vatID, abortCrank: reject, reject, info };
```
warner, line 323: huh, so we were allowing messages from the crank that did `vatPowers.exitVatWithFailure()`? oops.

I wonder what I was thinking. I'd say we should update the `vatPowers` docs to make it clear that `exitVat` accepts the vat's final messages, and `exitVatWithFailure` unwinds them, but apparently `vatPowers` doesn't show up in `packages/SwingSet/docs/` anywhere, which is too bad. I'm pretty sure @erights is depending upon it working this way.

Thanks for fixing this and updating the test to catch it.

packages/SwingSet/test/vat-admin/terminate/test-terminate.js
```diff
@@ -16,83 +16,276 @@ test.before(async t => {
...
-  await doTerminate(
+test('exit happy path with ante-mortem message (dynamic, critical)', async t => {
+  await doTerminateCritical(t, 'v8', 'happyTalkFirst', true);
+});
+
+test('exit happy path with ante-mortem message (static, critical)', async t => {
+  await doTerminateCritical(t, 'v3', 'happyTalkFirst', false);
+});
+
+test('exit sad path with ante-mortem message (dynamic, non-critical)', async t => {
```
warner, line 184: Let's add a comment (on the last argument line of the `doTerminateNonCritical` call) to remind ourselves that `GOT QUERY not dead quite yet (but soon)` should *not* be observed, because `vatPowers.exitVatWithFailure` is explicitly defined to unwind any message sent from the final crank. The test does exactly the right thing (thank you!) but because it's exercising an omission, it's not obvious what property is being checked (and therefore what we must make sure we retain in the future).

warner (approved):
Looks great! Thanks for catching that exitVatWithFailure-doesn't-unwind bug.

A few nits, and we need to figure out the CI thing (but it's probably transient, maybe a timeout or the CI host got rebooted or something?).

packages/SwingSet/src/kernel/kernel.js
FUDCo: Actually, this happened as the result of another change I made that referenced `terminate` earlier than it had been, and then that change ended up getting backed out in the course of finishing up the work on this item.  So we can put it back the way it was if that's your preference.

PR #5437, opened 2022-05-26
fix: record XS snapshots and file sizes to slog and console

warner:
This logs some additional information about XS heap snapshots, to help debug #5419 

* when snapshots files are first created, print the filename and size (both raw and compressed) to console.log
* when we record a vat snapshot, write the snapshot name and vatID into the slog trace
* when we create a worker from a snapshot, write the snapshot name and vatID into the slog trace

The console.log line a bit noisy, so we should remove it once we figure out what's going on with #5419 . The slog entries can remain.

cc @arirubinstein

### DIFF
--- a/packages/SwingSet/src/kernel/state/vatKeeper.js
+++ b/packages/SwingSet/src/kernel/state/vatKeeper.js
@@ -590,6 +590,12 @@ export function makeVatKeeper(
       JSON.stringify({ snapshotID, startPos: endPosition }),
     );
     addToSnapshot(snapshotID);
+    kernelSlog.write({
+      type: 'heap-snapshot-save',
+      vatID,
+      snapshotID,
+      endPosition,
+    });
     return true;
   }
 
--- a/packages/swing-store/src/snapStore.js
+++ b/packages/swing-store/src/snapStore.js
@@ -60,6 +60,7 @@ export const fsStreamReady = stream =>
  *   open: typeof import('fs').promises.open,
  *   resolve: typeof import('path').resolve,
  *   rename: typeof import('fs').promises.rename,
+ *   stat: typeof import('fs').promises.stat,
  *   unlink: typeof import('fs').promises.unlink,
  *   unlinkSync: typeof import('fs').unlinkSync,
  * }} io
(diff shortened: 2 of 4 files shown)


packages/SwingSet/src/kernel/state/vatKeeper.js
```diff
@@ -584,6 +584,12 @@ export function makeVatKeeper(
       JSON.stringify({ snapshotID, startPos: endPosition }),
     );
     addToSnapshot(snapshotID);
+    kernelSlog.write({
+      type: 'heap-snapshot-save',
+      vatID,
+      snapshotID,
+      endPosition,
+    });
```
mhofman, line 598: I wish we could somehow flow the console data into the slog as that seems like a more reliable debugging place.

Also it would really help XS trace debugging if we could somehow record the temp file name that was used, but that's tricky since a snapshot command may collide with an existing snapshot and get re-used.

packages/swing-store/src/snapStore.js
```diff
@@ -100,17 +102,19 @@ export function makeSnapStore(
 
   /**
    * @param {string} dest basename, relative to root
-   * @param { (name: string) => Promise<T> } thunk
-   * @returns { Promise<T> }
-   * @template T
+   * @param { (name: string) => Promise } thunk
+   * @returns { Promise<{ filename, size }> }
```
dckc:
```suggestion
   * @param { (name: string) => Promise<void> } thunk
   * @returns { Promise<{ filename: string, size: number }> }
```

dckc (approved):
Looks reasonable.

packages/SwingSet/src/kernel/state/vatKeeper.js
warner: Yeah, I'm on the fence about dumping the entirely of `console.log` into the slog, but I certainly do get frustrated by the thought that there's vital data on `stderr` that's closely related to the structured things in the slog but that we miss because it isn't captured (or is captured, separately, but not in a way that we can easily correlate).

I could add the tempfile name to the `console.log` message written by `snapStore.js` .. would that be helpful?

mhofman: Console output is virtually useless for post processing, and I actually would like to stop relying on it so much for following status.

This was more a wish. I had looked at the threading of these values and didn't see a non-convoluted way to get there.

PR #5440, opened 2022-05-26
vault security improvements

turadg:
## Description

Follow ups to https://github.com/Agoric/agoric-sdk/pull/5393#discussion_r882179573

### Security Considerations

Improvements

--

### Testing Considerations

No changes.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/liquidation.js
+++ b/packages/run-protocol/src/vaultFactory/liquidation.js
@@ -9,25 +9,6 @@ import { makeTracer } from '../makeTracer.js';
 const trace = makeTracer('LIQ', false);
 
 /**
- *
- * @param {Amount<'nat'>} debt
- * @param {Amount<'nat'>} proceeds
- */
-const discrepancy = (debt, proceeds) => {
-  if (AmountMath.isGTE(debt, proceeds)) {
-    return {
-      overage: AmountMath.makeEmptyFromAmount(debt),
-      shortfall: AmountMath.subtract(debt, proceeds),
-    };
-  } else {
-    return {
-      overage: AmountMath.subtract(proceeds, debt),
-      shortfall: AmountMath.makeEmptyFromAmount(debt),
-    };
-  }
-};
-
-/**
  * Liquidates a Vault, using the strategy to parameterize the particular
  * contract being used. The strategy provides a KeywordMapping and proposal
  * suitable for `offerTo()`, and an invitation.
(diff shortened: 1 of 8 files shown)


packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -86,7 +67,15 @@ const liquidate = async (
   ]);
   // NB: all the proceeds from AMM sale are on the vault seat instead of a staging seat
 
-  const { shortfall, overage } = discrepancy(debt, proceeds.RUN);
+  const [overage, shortfall] = AmountMath.isGTE(debt, proceeds.RUN)
```
turadg, line 70: I'm not sure this is better. read better before, imho.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -255,7 +255,16 @@ const helperBehavior = {
       },
       updateTime,
     );
-    partialAssign(state, stateUpdates);
+
+    // specify keys defensively
```
turadg: another way is,
```
state.compoundedInterest = stateUpdates.compoundedInterest;
```
(x3)

opinions @dtribble ?

dckc:
The PR title here buries the lede considerably. Please change it to something like:

 - fix: whole vault state was vulnerable to interest computation

well, I suppose it's not an observable bug fix. So...

 - refactor: avoid vault state vulnerability to interest computation

turadg:
Yeah, not a fix. Does your request to change the PR title have to do with the commit history or something else?

I'll be landing this as a merge commit so it's this commit that will appear in the history: [refactor(interest): specify keys defensively](https://github.com/Agoric/agoric-sdk/pull/5440/commits/5587ade472f45d4b6a657d6d839d41a1d4ca2590) 
Do you want me to rename that commit?

EDIT: I did while amending it: 
[refactor(interest): eschew partialAssign which made state vulnerable to interest result](https://github.com/Agoric/agoric-sdk/pull/5440/commits/692773bd61d5e418e09bcc4c712e06a4b097885c)

packages/run-protocol/src/vaultFactory/vaultManager.js
dckc: While use of `partialAssign` with the 3 explicit fields addresses the risk that changes in the interest computation might change something else, assignments to the individual fields are even more clear which parts of the `state` can be affected; no need for the reader to check the `partialAssign` function.

packages/run-protocol/src/vaultFactory/liquidation.js
dckc: either way is OK to me.

dckc (approved):
looks good.

packages/run-protocol/src/vaultFactory/liquidation.js
Chris-Hibbert: I might have gone with

```
  const short = AmountMath.isGTE(debt, proceeds.RUN);
  const overage = short
    ? AmountMath.makeEmptyFromAmount(debt)
    : AmountMath.subtract(proceeds.RUN, debt);
  const shortfall = short
    ? AmountMath.subtract(debt, proceeds.RUN)
    : AmountMath.makeEmptyFromAmount(debt);
```
I'm not a fan of the trinary assigning two values

PR #5441, opened 2022-05-26
fix: reinstate economy boot using minInitialPoolLiquidity=0

dckc:
refs: #5375
obsoletes: #5428

## Description

 - add economy bootstrap to CI (by way of `make scenario2-run-chain-to-halt` in cosmic-swingset)
 - set `minInitialPoolLiquidity` to 0 in "economy" bootstrap
 - when adding an issuer from an IBC denom (`publishInterchainAssetFromBank`), add it to the AMM, assuming `minInitialPoolLiquidity` is 0
 - fix(AMM): `await zcf.saveIssuer()` (drive-by, found during testing)

### Security Considerations

Users of "economy" bootstrap must manage the risk of spam AMM pools.

can't think of any

### Testing Considerations

restoring `test-gov-collateral.js` to working order was a bit challenging; the `benefactor.makePool()` code is now dead code, but I left a TODO to test that path, so I'm inclined to keep it around.


packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -41,8 +41,9 @@ export const makeAddIssuer = (zcf, isInSecondaries, brandToLiquidityMint) => {
         AssetKind.NAT,
         harden({ decimalPlaces: 6 }),
       ),
-      mint => {
-        zcf.saveIssuer(secondaryIssuer, keyword);
+      /** @param {ZCFMint} mint */
+      async mint => {
+        await zcf.saveIssuer(secondaryIssuer, keyword);
```
turadg, line 46: linking https://github.com/Agoric/agoric-sdk/issues/5407 for motivation

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -415,7 +425,7 @@ test('Benefactor can add to reserve', async t => {
   await s.startDevNet();
   await s.provisionMembers();
   await s.startRunPreview();
-  await s.benefactor.makePool(2000n, 1000n);
+  // await s.benefactor.makePool(2000n, 1000n);
```
turadg, line 428: please delete dead code

turadg (approved):

packages/run-protocol/test/test-gov-collateral.js
dckc: justification above doesn't suffice?

```
      // This isn't used now that we make the pool from a denom
      // in publishInterchainAssetFromBank in addAssetToVault.js
      // But it should still work. TODO: Perhaps we should test both ways?
      // i.e. from a board ID as well?
```

basically: this design has flip-flopped rapidly recently. I'd like to keep it there as a hedge.

PR #5444, opened 2022-05-26
feat(Reserve):  Reserve can burn RUN to reduce shortfall

### LINKED ISSUE #5299: Reserve contract covers vault shortfalls

Chris-Hibbert:
closes: #5299

## Description

Add an ability (controlled by governance) to burn RUN from the Reserve to reduce shortfalls from liquidations.

### Security Considerations

Governance & economic

We haven't started documenting the capabilities accessible to the econ committee.  This is one.

### Testing Considerations

Add tests so show that the shortfall is reduced.

### DIFF
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -222,6 +222,19 @@ const start = async (zcf, privateArgs) => {
     zcf.reallocate(offerToSeat, collateralSeat);
   };
 
+  const burnRUNToReduceShortfall = reduction => {
+    reduction = AmountMath.coerce(runBrand, reduction);
+    const runKeyword = keywordForBrand.get(runBrand);
+    const runBalance = collateralSeat.getAmountAllocated(runKeyword);
+    const amountToBurn = AmountMath.min(reduction, runBalance);
+    if (AmountMath.isEmpty(amountToBurn)) {
+      return;
+    }
+
+    runMint.burnLosses(harden({ [runKeyword]: amountToBurn }), collateralSeat);
+    reduceLiquidationShortfall(amountToBurn);
+  };
+
   const makeShortfallReportingInvitation = () => {
     const handleShortfallReportingOffer = () => {
       return Far('shortfallReporter', {
@@ -246,7 +259,7 @@ const start = async (zcf, privateArgs) => {
       makeShortfallReportingInvitation,
       getMetrics: () => metricsSubscription,
     },
-    { addLiquidityToAmmPool },
+    { addLiquidityToAmmPool, burnRUNToReduceShortfall },
   );
 
   const publicFacet = augmentPublicFacet(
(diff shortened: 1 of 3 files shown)


packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -222,6 +222,19 @@ const start = async (zcf, privateArgs) => {
     zcf.reallocate(offerToSeat, collateralSeat);
   };
 
+  const burnRUNToReduceShortfall = reduction => {
```
turadg, line 225: RUN is all-caps even in function names?

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -222,6 +222,19 @@ const start = async (zcf, privateArgs) => {
...
+    reduction = AmountMath.coerce(runBrand, reduction);
+    const runKeyword = keywordForBrand.get(runBrand);
+    const runBalance = collateralSeat.getAmountAllocated(runKeyword);
+    const amountToBurn = AmountMath.min(reduction, runBalance);
+    if (AmountMath.isEmpty(amountToBurn)) {
+      return;
+    }
+
+    runMint.burnLosses(harden({ [runKeyword]: amountToBurn }), collateralSeat);
```
turadg, line 234: does this still allow that the assertReserve contract's keywords are private and meaningless?

Chris-Hibbert: it should.  Just above this, `runKeyword` is set using `const runKeyword = keywordForBrand.get(runBrand);`, so whatever we use for the keywords should still be in `keywordForbrand`.

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: That was the policy I understood. I wouldn't mind if that's not the rule.

turadg: can wait for https://github.com/Agoric/agoric-sdk/issues/4800

packages/run-protocol/src/reserve/assetReserve.js
turadg: cool. I wasn't sure how keywords interacted with burnLosses but I see they're ignored https://github.com/Agoric/agoric-sdk/blob/fd44fca1ae25f9b17c34404bc03a7e79bdb4982b/packages/zoe/src/contractFacet/zcfZygote.js#L174-L175

turadg (approved):

PR #5448, opened 2022-05-27
Fee distributor contract

### LINKED ISSUE #4399: Reserve contract receives JS metering fees; IST fees distributed to reward pool and reserve

michaelfig:
closes: #4399

## Description

Split fees 50% between rewards and reserve.


packages/governance/test/unitTests/test-committee.js
```diff
@@ -74,7 +74,7 @@ test('committee-open question:one', async t => {
     electionType: ElectionType.SURVEY,
     maxChoices: 1,
     closingRule: {
-      timer: buildManualTimer(console.log),
+      timer: buildManualTimer(t.log),
```
turadg, line 77: why this change?

In my experience, putting lots in the Ava log makes the actual failure scroll off screen. E.g. I just made a change in the other direction: https://github.com/Agoric/agoric-sdk/pull/5440/commits/109145e616704826168799331917082fbff3a3c0

packages/deployment/Dockerfile.sdk
```diff
@@ -15,7 +15,7 @@ RUN make GIT_COMMIT="$GIT_COMMIT" MOD_READONLY= compile-go
 # OTEL fetch
 # from https://github.com/open-telemetry/opentelemetry-collector-releases/releases
 
-FROM node:14-buster AS otel
+FROM node:16-buster AS otel
```
turadg, line 18: necessary? (ok if not, just want to know)

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * @typedef {object} DistributorParams
+ *
+ * @property {bigint} [epochInterval=1n] - parameter to the epochTimer
+ *  controlling the interval at which rewards should be sent to the bank.
+ */
+
+/**
+ * @callback BuildFeeDistributor
```
turadg: please define this on the function declaration instead a callback. that when when you navigated to a part of the type you end up at the implementation as well.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+import { makeScalarMapStore } from '@agoric/store';
+
+/**
+ * @typedef {object} FeeCollector
+ *
+ * @property {() => ERef<Payment>} collectFees
+ */
+
+/**
+ * @typedef {object} DistributorParams
```
turadg: why create a named type for this? I see you want to document `params.epochInterval`. Please do that with https://jsdoc.app/tags-param.html#parameters-with-properties on the function signature.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+  });
+}
+
+/**
+ * A distributor of fees from Inter Protocol sources to deposit facets. Each
+ * time the epochTimer signals the end of an Epoch, it will ask the contracts
+ * for fees that have been collected to date and send that payment to the
+ * depositFacet.
+ *
+ * @param {() => Promise<any>} schedulePayments - distribute to the destinations
```
turadg: fully typing this isn't worth it, but `unknown` is always safer than `any`.
```suggestion
 * @param {() => Promise<unknown>} schedulePayments - distribute to the destinations
```

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+ * @param {ERef<TimerService>} timerService - timer that is used to schedule collections
+ * @param {RelativeTime} [collectionInterval] - how often to collect fees
```
turadg: in seconds? if so, should the timer service also be specified as in seconds?

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+  );
+  observeNotifier(collectionNotifier, timeObserver).catch(e => {
+    console.error('fee distributor failed with', e);
+  });
+}
+
+/**
+ * @type {BuildFeeDistributor}
+ */
+export const buildDistributor = async (
```
turadg: only this function needs to be exported, not the `startDistributing` one.

both of them "start distributing" but over different scopes. the non-exported helper could be `startDistributingPayments` and this one `startDistributingFromCollectors`.

please stop the unnecessary export and make the exported name include "start".

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+    timerService,
+    epochInterval,
+  );
+};
+
+/**
+ * @typedef {{ pushPayment: (payment: Payment, issuer: Issuer) => Promise<Amount>}} FeeDestination
+ *
+ * @param {Record<string, ERef<FeeDestination>>} [destinations]
+ * @param {Record<string, any>} [keywordShares]
```
turadg:
```suggestion
 * @param {Record<Keyword, ERef<FeeDestination>>} [destinations]
 * @param {Record<Keyword, bigint>} [keywordShares]
```

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+      totalShares += share;
+      return {
+        share,
+        destination,
+      };
+    });
+  return { shares, totalShares };
+};
+
+/** @typedef {ReturnType<typeof makeShareConfig>} ShareConfig */
```
turadg, line 105: 👍

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+ * shareFees tries to be tolerant of single points of failure.
+ *
+ * @param {Payment} payment
+ * @param {Issuer} issuer
+ * @param {ShareConfig} shareConfig
+ */
+export const shareFees = async (payment, issuer, { shares, totalShares }) => {
+  // Divide the payment into shares.
+  /** @type {Amount<'nat'>} */
+  const totalAmount = await E(issuer).getAmountOf(payment);
```
turadg: consider s/total/payment to keep clear what the value is of

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+  if (AmountMath.isEmpty(totalAmount)) {
+    return;
+  }
+  const { value: totalValue, brand } = totalAmount;
+
+  let remainder = totalValue;
+  /** @type {[ERef<FeeDestination>, Amount][]} */
```
turadg: doesn't hurt but this is inferred from 167

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+  const destinedAmountEntries = shares
+    .map(({ share, destination }, i) => {
+      const value =
+        i === shares.length - 1
+          ? remainder
+          : (totalValue * share) / totalShares;
+      remainder -= value;
+      /** @type {[ERef<FeeDestination>, Amount]} */
+      const entry = [destination, { value, brand }];
+      return entry;
```
turadg: consider,
```suggestion
      return /** @type {const} */ ([destination, { brand, value }]);
```

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+    })
+    .filter(([_, amt]) => !AmountMath.isEmpty(amt));
+
+  // Split the payment and assert we have a conservation of the total.
```
turadg: does `splitMany` do that assertion? I didn't see that in the docs but I'll take your word :) 

if so, would help to clarify that:
```suggestion
  // Split the payment, which ensures a conservation of the total.
```

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+          await E(result).getOfferResult();
+
+          // We deliberately drop our payouts on the floor, since the ERTP purse
+          // recovery mechanism can get them back to the distributor contract.
+          return paymentAmount;
+        },
+      });
+    },
+
+    setDestinations: async destinations => {
```
turadg: please jsdoc the param

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+
+  // Send each nonempty payment to the corresponding destination.
+  await Promise.all(
+    destinedAmountEntries.map(([destination], i) =>
+      E(destination).pushPayment(sharedPayment[i], issuer),
+    ),
+  );
+};
+
+/** @type {ContractStartFn} */
```
turadg: please type just the `zcf` param so the return type can be inferred. even better to define the terms like so:
[
```suggestion
/**
 * @param {ZCF} zcf
  */
```

even better to define the terms like so ](https://github.com/Agoric/agoric-sdk/blob/9156f614b7ea8ff8a1862194317d8ebf6324e9a2/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js#L49-L62)

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+export const start = async zcf => {
+  const { timerService, collectionInterval, ...rest } = zcf.getTerms();
```
turadg: the `...rest` makes the types a little loose. please put it under a keyword `keywordShares` key or explain why that's worse.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+ * @param {Issuer} issuer
+ * @param {ShareConfig} shareConfig
+ */
+export const shareFees = async (payment, issuer, { shares, totalShares }) => {
+  // Divide the payment into shares.
+  /** @type {Amount<'nat'>} */
+  const totalAmount = await E(issuer).getAmountOf(payment);
+  if (AmountMath.isEmpty(totalAmount)) {
+    return;
+  }
```
turadg, line 128: consider also returning if there are no shares. It's not immediately obvious that this does nothing in that case.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+    );
+  };
+
+  const publicFacet = Far('feeDistributor publicFacet', {
+    distributeFees,
+  });
+
+  const zoe = zcf.getZoeService();
+  const creatorFacet = Far('feeDistributor creatorFacet', {
+    addFeeSource: async sourceP => {
```
turadg: please type facet methods

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+      const [source, collector] = await Promise.all([
+        sourceP,
+        makeFeeCollector(zoe, sourceP),
+      ]);
+      sourceToCollector.init(source, collector);
+      if (shareConfig.totalShares <= 0n) {
+        return;
+      }
+      // Run once immediately for this source.
```
turadg: why?

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+      await E(collector).collectFees().then(distributeFees);
+    },
+    deleteFeeSource: async sourceP => {
```
turadg: consider `removeFeeSource` (it will still exist, just removed from this collector)

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+        },
+      });
+    },
+    /**
+     * Create a destination that generates invitations and makes Zoe offers.
+     *
+     * @param {string} keyword
+     * @param {any} target
+     * @param {PropertyKey} makeInvitationMethod
+     * @param {any[]} [args]
```
turadg:
```suggestion
     * @param {unknown} target
     * @param {PropertyKey} makeInvitationMethod
     * @param {unknown[]} [args]
```

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,302 @@
...
+  startDistributing(schedulePayments, timerService, collectionInterval);
+
+  return harden({
+    creatorFacet,
+    publicFacet,
+  });
+};
+
+/** @typedef {ReturnType<typeof start>['creatorFacet']} FeeDistributorCreatorFacet */
+/** @typedef {ReturnType<typeof start>['publicFacet']} FeeDistributorPublicFacet */
```
turadg:
```suggestion
/** @typedef {Awaited<ReturnType<typeof start>>['creatorFacet']} FeeDistributorCreatorFacet */
/** @typedef {Awaited<ReturnType<typeof start>>['publicFacet']} FeeDistributorPublicFacet */
```
the reason this didn't error before is the `ReturnType<typeof start>` is `any` due to `ContractStartFn`

turadg (changes requested):
covered what I could before my meeting rn.

packages/deployment/Dockerfile.sdk
michaelfig: Yes, necessary for latest instagoric.  Sorry, should've marked this as a drive-by.

packages/governance/test/unitTests/test-committee.js
michaelfig: Console pollution for successful tests really irks me; it makes actual failures scroll off the screen.

packages/run-protocol/src/feeDistributor.js
michaelfig: No, it's in the timer services underlying unit, which may not be seconds (it may be epochs for example).  I'll specify it.

packages/governance/test/unitTests/test-committee.js
michaelfig: I think the solution in both of our cases is to buy a bigger scrollback buffer. :)

michaelfig: I agree with your change to gov-collateral, but I find that the manualTimers are not noisy enough to warrant sending them anywhere but `t.log`.

turadg: ah, yeah me too. didn't realize that `t.log` only prints if the test fails.

though because not all logging has access to `t.log` this change makes it so the stream is messages is separated into two different places, making it harder to follow the messages.

what I've done in the past to keep logs clean is override `console` methods with a buffer. we could do something for Ava like that which only prints the console if the test fails. WDYT?

for now humbling requesting that this change be separated to another PR where it can be considered more holistically, but won't block on it.

turadg (approved):
Some requests weren't addressed but I'll approve to save time.

packages/run-protocol/src/feeDistributor.js
michaelfig: Thanks for the advice.  I extracted the terms with `Parameters<...>`.

packages/run-protocol/src/feeDistributor.js
michaelfig: To both ensure we don't have a bad fee collector, and to catch up in case we're adding a fee collector much later than we started distributing fees.

packages/run-protocol/src/feeDistributor.js
michaelfig: I opted for `startPeriodicCollection` and a `stop` method on its return value.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,308 @@
...
+
+  const creatorFacet = Far('feeDistributor creatorFacet', {
+    makeContractFeeCollector,
+    /**
+     * Start distributing fees from this collector.
+     * 
+     * @param {string} debugName 
+     * @param {ERef<FeeCollector>} collectorP
+     */
+    startPeriodicCollection: async (debugName, collectorP) => {
```
turadg, line 201: 👌 naming

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,308 @@
...
+  });
+}
+
+/**
+ * @typedef {{ pushPayment: (payment: Payment, issuer: ERef<Issuer>) => Promise<Amount>}} FeeDestination
+ *
+ * @param {Record<Keyword, ERef<FeeDestination>>} [destinations]
+ * @param {Record<Keyword, bigint>} [keywordShares]
+ */
+export const makeShareConfig = (destinations = {}, keywordShares = {}) => {
```
turadg, line 89: not imported. consider not exporting.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,308 @@
...
+/** @typedef {ReturnType<typeof makeShareConfig>} ShareConfig */
+
+/**
+ * Split and deposit a payment given a share configuration.
+ *
+ * @param {Payment} payment
+ * @param {ERef<Issuer>} issuer
+ * @param {ShareConfig} shareConfig
+ */
+export const sharePayment = async (
```
turadg, line 114: not imported. consider not exporting.

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,308 @@
...
+ */
+
+/**
+ * wrapper to take the vaultFactory or AMM's creatorFacet, and make a function that
+ * will request an invitation and return a promise for a payment.
+ *
+ * @param {ERef<ZoeService>} zoe
+ * @param {ERef<CollectibleContractFacet>} creatorFacet
+ */
+export function makeContractFeeCollector(zoe, creatorFacet) {
```
turadg: not imported. consider not exporting. TMK, contracts are to export only `start`.

also shouldn't this be `const makeContractFeeCollector = (zoe, creatorFacet) => {` ?

packages/run-protocol/src/feeDistributor.js
```diff
@@ -0,0 +1,308 @@
...
+ * time the epochTimer signals the end of an Epoch, it will ask the contracts
+ * for fees that have been collected to date and send that payment to the
+ * depositFacet.
+ *
+ * @param {() => Promise<unknown>} schedulePayments - distribute to the destinations
+ * @param {ERef<TimerService>} timerService - timer that is used to schedule collections
+ * @param {RelativeTime} [collectionInterval] - how often to collect fees in the
+ * `timerService` unit
+ */
+export function startDistributing(
```
turadg: same as above, not imported and should be const

packages/run-protocol/test/test-feeDistributor.js
```diff
@@ -8,8 +8,8 @@ import { setup } from '@agoric/zoe/test/unitTests/setupBasicMints.js';
 
 import { makePromiseKit } from '@endo/promise-kit';
 import { assertPayoutAmount } from '@agoric/zoe/test/zoeTestHelpers.js';
-import { E } from '@endo/far';
-import { buildDistributor } from '../src/distributeFees.js';
+import { E, Far } from '@endo/far';
+import { makeFeeDistributor } from '../src/feeDistributor.js';
```
turadg, line 12: I like that you're testing the object. is anything testing the contract?

turadg (approved):
Outstanding concerns but nothing that should delay security review.

packages/run-protocol/test/test-feeDistributor.js
michaelfig: > I like that you're testing the object. is anything testing the contract?

Yes, the `run-protocol/test/test-gov-collaterals.js` is testing it.

packages/run-protocol/src/feeDistributor.js
michaelfig: I'm exporting for future tests.  I just never got to them.

> TMK, contracts are to export only `start`.

Not necessarily, as long as there is motivation.

packages/run-protocol/src/feeDistributor.js
michaelfig: Agreed on constness.  I'll leave it for future `function` keyword cleanups that may be more extensive.

PR #5449, opened 2022-05-27
increase provisioned RUN from 50 to 500 for local-chain

0xshipthecode:
## Issues referenced

I did not create an issue because the problem is simple. 50 `RUN` is not enough to re-deploy a simple contract a few times. I have increased the `RUN` to 500 to prevent us early adopters from having to restart the `local-chain`. 

## Description

Only the variable `PROVISIONED_COINS` in `packages/agoric-cli/src/start.js` was modified.  The amount of `RUN` was increased to 500.

### Security Considerations

There are no security implications. Users spam their own local chain.

Tell users the local-chain is now more generous and they can deploy many contracts happily!

### Testing Considerations

None.


dckc:
Hi! Thanks for contributing.

I suppose we haven't noticed this issue or don't run into it much because we tend to use a somewhat custom configuration to run our clients where we use cosmos transactions to send them more tokens:

https://github.com/Agoric/agoric-sdk/blob/98d35809ec28b26aa3854eedd1e1af2c9801efbe/packages/cosmic-swingset/Makefile#L214-L221

I don't see any problem with starting with 500 instead of 50 in the plain `agoric start local-chain` case. @michaelfig , how about you?

michaelfig (approved):
Yes, the approach is fine.

LGTM!

0xshipthecode:
I pushed another commit to set `RUN` budget to 500 for the `local-solo` as well.

michaelfig:
> I pushed another commit to set `RUN` budget to 500 for the `local-solo` as well.

Great!  Can you please update the branch so that we can merge it in the SDK?  (Either just click the Github button, or merge locally and push to your branch.)

michaelfig:
#5590 will implement this PR, but can be landed automatically.  No need for more updates.

0xshipthecode:
Updated per your request @michaelfig !

PR #5452, opened 2022-05-27
rename Store param 'keyName'  to 'tag'

turadg:
## Description

makeStore() [docs](https://github.com/Agoric/agoric-sdk/blob/358c8b7bef2c9a659344d72fd52aa80f14a93b68/packages/store/src/stores/scalarMapStore.js#L127) seems to be outdated. Per @FUDCo , "It’s just an arbitrary tag string for documentary and debugging purposes."

This also has some typedef annotation for VirtualObjectManager that I used while investigating code for the primary change. It's in a separate commit so it will appear separately in master history.

### Security Considerations

--

--

### Testing Considerations

Shouldn't affect any tests.

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -1,5 +1,5 @@
 // @ts-check
-/* eslint-disable no-use-before-define */
+/* eslint-disable no-use-before-define, jsdoc/require-returns-type */
 
 import { assert, details as X, q } from '@agoric/assert';
 import { Far } from '@endo/marshal';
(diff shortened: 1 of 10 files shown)


FUDCo (changes requested):
The changes to the three files in the `SwingSet` package seem entirely unrelated to the declared purpose of this PR.  They look to me like something that just got unintentionally swept up by a `git commit -a`.  Please either put these changes in a separate PR or document them.

turadg:
> seem entirely unrelated to the declared purpose of this PR.

Indeed. Sorry for the surprise. I kept them in a [separate commit](https://github.com/Agoric/agoric-sdk/pull/5452/commits/3842fe5558a2774f71065a488544af5ff3c59f67) so that master history wouldn't keep them separated but I didn't want to go through another PR and build cycle to get small typedef changes.

I'll add it to the PR description.

FUDCo (approved):
Well OK then.

PR #4404, opened 2022-01-27
refactor: unify TypeScript project

turadg:
## Description

I was trying to create a type coverage report and along the way it was easiest to unify the type checking into one TypeScript project. With this the checker can be more comprehensive and it turns out there are over a thousand type errors that were masked by the earlier config.

EDIT: having a single TypeScript project will solve OOM in eslint `https://github.com/typescript-eslint/typescript-eslint/issues/1192#issuecomment-1153418862`

PR #5457, opened 2022-05-27
lint ts-ignore

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/5410 required a lot of plumbing work to make the actual lint rule enable-able. This pulls that work out into a separate PR.

To make it a little more worthy, this also enables the `prefer-ts-expect-error` rule.

This also refactors config responsibilities to be in line with https://github.com/endojs/endo/pull/1201#issue-1256332542

Recommended to review by commit.

### Security Considerations

n/a

Reduces need for documenting ts-ignore vs ts-expect-error because the linting system will provide feedback.

### Testing Considerations

The collection of type information slowed down linting so it was postponed to #5410

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -44,6 +44,7 @@ module.exports = {
     'coverage/**',
     '**/output/**',
     'bundles/**',
+    'bundle-*',
     'dist/**',
     'examples/**',
     'test262/**',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -48,6 +48,16 @@ module.exports = {
     'dist/**',
     'examples/**',
     'test262/**',
+    '*.html',
     'ava*.config.js',
+  ],
+  overrides: [
+    {
+      // disable type-aware linting in HTML
+      files: ['*.html'],
+      parserOptions: {
+        project: false,
+      },
+    },
   ],
 };
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,0 +1,52 @@
+/* eslint-env node */
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  plugins: [
+    '@jessie.js/eslint-plugin',
+    '@typescript-eslint',
+    'eslint-plugin-import',
+    'eslint-plugin-prettier',
+  ],
+  extends: ['@agoric'],
+  rules: {
+    '@typescript-eslint/prefer-ts-expect-error': 'warn',
+    'jsdoc/no-multi-asterisks': 'off',
+    'jsdoc/multiline-blocks': 'off',
+    // Use these rules to warn about JSDoc type problems, such as after
+    // upgrading eslint-plugin-jsdoc.
+    // Bump the 1's to 2's to get errors.
+    // "jsdoc/valid-types": 1,
+    // "jsdoc/no-undefined-types": [1, {"definedTypes": ["never", "unknown"]}],
+    'jsdoc/tag-lines': 'off',
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        devDependencies: [
+          '**/*.config.js',
+          '**/*.config.*.js',
+          '**/*test*/**/*.js',
+          '**/demo*/**/*.js',
+          '**/scripts/**/*.js',
+        ],
+      },
+    ],
+    // CI has a separate format check but keep this warn to maintain that "eslint --fix" prettifies
+    // UNTIL https://github.com/Agoric/agoric-sdk/issues/4339
+    'prettier/prettier': 'warn',
+  },
+  settings: {
+    jsdoc: {
+      mode: 'typescript',
+    },
+  },
+  ignorePatterns: [
+    'coverage/**',
+    '**/output/**',
+    'bundles/**',
+    'dist/**',
+    'examples/**',
+    'test262/**',
+    'ava*.config.js',
+  ],
+};
--- a/package.json
+++ b/package.json
@@ -17,10 +17,16 @@
     "ava": "^3.15.0",
     "c8": "^7.11.0",
     "conventional-changelog-conventionalcommits": "^4.6.0",
-    "eslint": "^7.32.0",
     "eslint-config-airbnb-base": "^14.2.0",
+    "eslint-config-jessie": "^0.0.6",
+    "eslint-config-prettier": "^6.15.0",
     "eslint-plugin-import": "^2.25.3",
-    "eslint-plugin-jsx-a11y": "^6.4.0",
+    "eslint-plugin-jsdoc": "^39.2.9",
+    "eslint-plugin-jsx-a11y": "^6.5.1",
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-react-hooks": "^4.3.0",
+    "eslint-plugin-react": "^7.28.0",
+    "eslint": "^7.32.0",
     "lerna": "^3.22.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.6.2",
@@ -28,9 +34,6 @@
   },
   "engines": {
     "node": ">=14.15.0"
-  },
-  "eslintConfig": {
-    "root": true
   },
   "prettier": {
     "arrowParens": "avoid",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@endo/eslint-config": "^0.4.10",
     "@jessie.js/eslint-plugin": "^0.1.3",
     "@types/node": "^16.7.10",
-    "@typescript-eslint/parser": "^5.15.0",
+    "@typescript-eslint/parser": "^5.27.0",
     "ava": "^3.15.0",
     "c8": "^7.11.0",
     "conventional-changelog-conventionalcommits": "^4.6.0",
--- a/packages/eslint-config/eslint-config.json
+++ b/packages/eslint-config/eslint-config.json
@@ -2,44 +2,5 @@
   "extends": [
     "@endo",
     "plugin:@jessie.js/recommended"
-  ],
-  "parser": "@typescript-eslint/parser",
-  "plugins": ["@typescript-eslint"],
-  "rules": {
-    "@typescript-eslint/prefer-ts-expect-error": "warn",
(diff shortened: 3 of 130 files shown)


.eslintrc.cjs
```diff
@@ -0,0 +1,68 @@
...
+        devDependencies: [
+          '**/*.config.js',
+          '**/*.config.*.js',
+          '**/*test*/**/*.js',
+          '**/demo*/**/*.js',
+          '**/scripts/**/*.js',
+        ],
+      },
+    ],
+    'prettier/prettier': 'warn', // @jessie.js/recommended errors. This repo checks format separately but keeps warn to maintain that "eslint --fix" prettifies
```
michaelfig: Can you move the `@jessie.js` comment to a separate line?

package.json
```diff
@@ -29,9 +35,6 @@
   "engines": {
     "node": ">=14.15.0"
   },
-  "eslintConfig": {
```
michaelfig, line 32: Why remove this?

packages/eslint-config/eslint-config.json
```diff
@@ -2,41 +2,5 @@
...
-        "devDependencies": [
-          "**/*.config.js",
-          "**/*.config.*.js",
-          "**/*test*/**/*.js",
-          "**/demo*/**/*.js",
-          "**/scripts/**/*.js"
-        ]
-      }
-    ],
-    "prettier/prettier": "warn" // @jessie.js/recommended errors. This repo checks format separately but keeps warn to maintain that "eslint --fix" prettifies
```
michaelfig, line 27: Ah, I see you inherited the comment from this line.  Still nice to fix.

packages/ui-components/package.json
```diff
@@ -51,7 +51,7 @@
     "@babel/core": "^7.12.13",
     "@babel/plugin-syntax-jsx": "^7.12.1",
     "@material-ui/core": "4.11.3",
-    "@typescript-eslint/eslint-plugin": "^5.15.0",
+    "@typescript-eslint/eslint-plugin": "^5.27.0",
```
michaelfig, line 54: Does this package need to depend on `@typescript-eslint`?

packages/web-components/src/AgoricWalletConnection.js
```diff
@@ -133,7 +133,7 @@ export const makeAgoricWalletConnection = (makeCapTP = defaultMakeCapTP) =>
           return this._bridgePK.promise;
         },
         reset: () => {
-          this.reset();
+          void this.reset();
```
michaelfig: Is this explicitly ignoring a dangling promise?  This idiom seems like a good way to indicate that situation.

packages/zoe/test/unitTests/contracts/test-atomicSwap.js
```diff
@@ -322,7 +322,7 @@ test('zoe - non-fungible atomicSwap', async t => {
   // Alice makes an instance and makes her offer.
   const installation = await alice.installCode();
 
-  const vorpalSword = createRpgItem('Vorpal Sword', 38);
+  const vorpalSword = createRpgItem('Vorpal Sword', 'vorping');
```
michaelfig, line 325: 😆

packages/zoe/test/unitTests/contracts/test-coveredCall.js
```diff
@@ -923,7 +923,7 @@ test('zoe - coveredCall non-fungible', async t => {
   // Setup Bob
   const aGloriousShield = createRpgItem(
     'Glorious Shield',
-    25,
+    'blinding',
```
michaelfig, line 926: 😂

michaelfig (approved):
Whew!  I made it through, with just a few questions (inline) I'd appreciate you answering.

Also, are all the `@ts-nocheck`s when type checking a test file had diminishing returns?

Great stuff, though!  I'd love to see this land.

.eslintrc.cjs
turadg: Hoping to get this in which will obviate the "jessie" part https://github.com/Agoric/eslint-config-jessie/pull/28

There's still the "--fix" issue which I'll punt until https://github.com/Agoric/agoric-sdk/issues/4339

packages/ui-components/package.json
turadg: Good question. Yes for `react-app` this extends on line 35. https://github.com/facebook/create-react-app/blob/main/packages/eslint-config-react-app/index.js#L45

package.json
turadg: root config now in `.eslintrc.js`.

packages/eslint-config/eslint-config.json
turadg: What to fix?

packages/web-components/src/AgoricWalletConnection.js
turadg: Yeah. Though I'm saving that for https://github.com/Agoric/agoric-sdk/pull/5457  (removed)

turadg:
I got the timing test in and it's a significant cost so requested review. (I also pushed some fixups to the commit history, rebased to keep it clean for a merge)

Including types in the lint takes it from 1.4min to 14min. [CI takes 5min](https://github.com/Agoric/agoric-sdk/runs/6693643820?check_suite_focus=true) right now on lint check after 5min restore-node. So the additional 13min would make the lint job go from 10min to 23min. (To be confirmed by [the job results](https://github.com/Agoric/agoric-sdk/runs/6694898124?check_suite_focus=true) )

Acceptable?

I think it’ll be worth it to lint floating promises and other [type-based rules](https://typescript-eslint.io/rules/) but I want to check before merge.

We may also be able to speed it up by cleaning up the TS projects or moving to a single eslint command. When I try all of packages/ it OOMs but running time yarn eslint packages/*/src/**/*.js takes just 96s and time yarn eslint packages/*/test/**/*.js takes 112s. 

UPDATE: on second thought, since the slowdown is only for the type info which is only for https://github.com/Agoric/agoric-sdk/pull/5410 I'll punt that change and get this in now. It touches many files so get stale quickly.

mhofman:
I'm a little confused by the time increase. A ten fold increase does not seem acceptable. If it impacts CI, does it also impact a local IDE too?

.eslintrc.cjs
```diff
@@ -0,0 +1,62 @@
+module.exports = {
```
mhofman, line 2: Out of curiosity, why is this a separate file and no longer in the root `package.json`? Is that what allows the removal of all the dummy `"extends"` in the sub `package.json`s?

yarn.lock
```diff
@@ -7045,6 +7045,13 @@ debug@^3.1.0, debug@^3.1.1, debug@^3.2.7:
   dependencies:
     ms "^2.1.1"
 
+debug@^4.3.2, debug@^4.3.4:
```
mhofman: nit: how come this ended up duplicated away from `debug@4, debug@^4.0.1, debug@^4.1.0, debug@^4.1.1, debug@^4.2.0, debug@^4.3.1`

yarn.lock
```diff
@@ -9210,6 +9217,18 @@ globby@^11.0.1, globby@^11.0.4:
     merge2 "^1.3.0"
     slash "^3.0.0"
 
+globby@^11.1.0:
```
mhofman: nit: dedup with `globby@^11.0.1, globby@^11.0.4`

tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
mhofman, line 1: Nit: what is the relation of this root `tsconfig.json` with the addition of a root `eslintrc.cjs` ?

mhofman (commented):
I'm still parsing through some of the changes. I'm a little confused as to what causes the time increase here. Is it the new root tsconfig?

.eslintrc.cjs
turadg: it was necessary for using `dirname` which let the lint config find the ts/jsconfig. I removed that to stop culling type info in this PR but I kept the file format since it will be necessary later.

yarn.lock
turadg: dunno. I did dedup to clean up this and globby

yarn.lock
turadg: done

tsconfig.json
turadg: the `.eslintrc.js` is a JS file which was being linted, but the linter complained that it couldn't find the TS project. This creates a minimal TS project that encompasses `.eslintrc.js`. I still hope to unify TypeScript config someday: https://github.com/Agoric/agoric-sdk/pull/4404

packages/eslint-config/eslint-config.json
```diff
@@ -2,41 +2,5 @@
...
-      {
-        "devDependencies": [
-          "**/*.config.js",
-          "**/*.config.*.js",
-          "**/*test*/**/*.js",
-          "**/demo*/**/*.js",
-          "**/scripts/**/*.js"
-        ]
-      }
-    ],
```
mhofman, line 26: I thought that dapps depended on `@agoric/eslint-config`. If we remove this, won't dapps lose their default config?

michaelfig:
> running time yarn eslint packages/_/src/**/_.js takes just 96s and time yarn eslint packages/_/test/**/_.js takes 112s.

I think this indicates it would be profitable to split the lint Github action job by globs (lint1= src, lint2=test, etc), just as is done for some of the `test-all-packages` runs (such as SwingSet).

packages/eslint-config/package.json
```diff
@@ -24,10 +24,10 @@
   "files": [
     "eslint-config.json"
   ],
-  "devDependencies": {
+  "peerDependencies": {
```
mhofman, line 27: Similar issue here, I think that dapps using "link" mode would automatically find the right dev dependencies, but arguably this would already be broken in "npm" mode.

packages/eslint-config/eslint-config.json
turadg: Good observation. The dapps would no longer have their jsdoc and import rules externally specified. I think that's an improvement but I neglected to call it out. I'll reference https://github.com/endojs/endo/pull/1201#issue-1256332542 in the PR description.

PR #5461, opened 2022-05-28
fix: schema limit minting and offers

erights:
Introduce patterns in some strategic places to do earlier or better conformance checking of passables against some requirements.

This PR introduces an optional `elementSchema` pattern to `makeIssuerKit` for use with non-fungible or semi-fungible assets, where each asset description must match the pattern. Thus, for example, the invitation issuer knows the schema that all invitation amount elements need to satisfy. Minting a malformed invitation amount element could be caught earlier and more reliably.

The `brand`, `assetType`, and optional `elementSchema` together determine the `amountSchema`, which is the pattern than valid amounts for that issuer need to satisfy. For example, an invitation amount must have the invitation brand, and its value must represent a set (currently by an array) each of whose elements satisfy the invitation element schema. Anyone knowing the brand can ask it for the amount pattern and check their own amounts against it. They can do this with remote brands, since patterns are pass-by-copy, and pattern matching is pass-invariant.

This PR introduces an optional `proposalSchema` as an additional optional argument to `makeInvitation`. A contract could use this rather than the current ad-hoc `assertProposalShape`. As a result, an offer with a proposal that doesn't match would be rejected by zoe, never reaching the contract.

### DIFF
--- a/packages/ERTP/src/brand.js
+++ b/packages/ERTP/src/brand.js
@@ -1,14 +1,25 @@
 // @ts-check
+import { M } from '@agoric/store';
 import { E } from '@endo/eventual-send';
 import { Far } from '@endo/marshal';
+
+const { details: X, quote: q } = assert;
 
 /**
  * @param {string} allegedName
  * @param {(allegedIssuer: Issuer) => boolean} isMyIssuerNow
  * @param {DisplayInfo} displayInfo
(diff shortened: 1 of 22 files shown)


Chris-Hibbert (approved):
This looks good.

Issue #4723, opened 2022-03-02
Sponsoring a friend to create an Agoric account


Tartuffo:
#4654 is about bootstrapping account creation in a decentralized manner by using a lien on some BLD (or, ultimately, RUN) that a user has.

This issue is about someone who has BLD or RUN being able to "sponsor" the creation of an account for someone else, even if my BLD is locked.

This should have some built-in restrictions, so for example a single user can't create thousands of accounts.

PR #5465, opened 2022-05-29
Chain storage revamp

michaelfig:
closes: #5406

## Description

Mass exodus of storage system from `x/swingset` to `x/vstorage`.  This replaces #5406, and incorporates all of its semantic changes.

`x/vstream` (#5466) is a layer on top of `x/vstorage` with change notification events to enable future subscriptions as well as historical backfilling.


michaelfig:
Note to reviewers; the only CI failure is for protobuf backward incompatibility.  I think that's okay, since this is advertised as a breaking change.

golang/cosmos/app/app.go
```diff
@@ -704,7 +713,7 @@ func (app *GaiaApp) MustInitController(ctx sdk.Context) {
 		Type:        "AG_COSMOS_INIT",
 		ChainID:     ctx.ChainID(),
 		Params:      app.SwingSetKeeper.GetParams(ctx),
-		StoragePort: vm.GetPort("storage"),
+		StoragePort: vm.GetPort("vstorage"),
```
JimLarson: Make a `storagePort` field in `GaiaApp` for symmetry with `vibc/vbank/lienPort`, or alternatively, remove all these fields and do `vm.GetPort()` here for them too.

golang/cosmos/proto/agoric/swingset/swingset.proto
```diff
@@ -78,3 +78,23 @@ message StringBeans {
...
+    string nickname = 1 [
+        (gogoproto.jsontag)    = "nickname",
+        (gogoproto.moretags)   = "yaml:\"nickname\""
+    ];
+    bytes peer = 2 [
+        (gogoproto.casttype)   = "github.com/cosmos/cosmos-sdk/types.AccAddress",
+        (gogoproto.jsontag)    = "peer",
+        (gogoproto.moretags)   = "yaml:\"peer\""
+    ];
+    repeated string power_flags = 3 [
```
JimLarson, line 96: Add comments describing all these fields, but in particular add a reference to where the vocabulary of power flags is defined.

golang/cosmos/proto/agoric/vstorage/genesis.proto
```diff
@@ -0,0 +1,22 @@
...
+message GenesisState {
+    option (gogoproto.equal) = false;
+
+    repeated DataEntry data = 1 [
+        (gogoproto.jsontag)    = "data",
+        (gogoproto.moretags)   = "yaml:\"data\""
+    ];
+}
+
+// A vstorage entry.
```
JimLarson: Note whether placeholder entries for data-less ancestors are expected.

golang/cosmos/proto/agoric/vstorage/genesis.proto
```diff
@@ -0,0 +1,22 @@
...
+message DataEntry {
+    string path = 1;
```
JimLarson, line 23: Note the syntax of paths, e.g. `"."` is not legal.

golang/cosmos/proto/agoric/vstorage/query.proto
```diff
@@ -0,0 +1,58 @@
...
+message QueryDataRequest {
+  repeated string path = 1 [
+    (gogoproto.jsontag)    = "path",
+    (gogoproto.moretags)   = "yaml:\"path\""
+  ];
+}
+
+// QueryDataResponse is the vstorage path data response.
+message QueryDataResponse {
+  string value = 1 [
```
JimLarson, line 34: Specify the semantics vis a vis the repeated paths in the query vs the single value in the result. Does it return the first path with data? The last? Does it combine all retrieved values with some separator?

golang/cosmos/proto/agoric/vstorage/vstorage.proto
```diff
@@ -0,0 +1,26 @@
...
+message Data {
+    option (gogoproto.equal) = false;
+
+    string value = 1 [
+        (gogoproto.jsontag)    = "value",
+        (gogoproto.moretags)   = "yaml:\"value\""
+    ];
+}
+
+// Keys are the vstorage node subkeys.
```
JimLarson: Decide on "subkeys" or "child keys" and use the choice consistently. Whatever they're called, make it clear that we're returning the just the parent-relative name, not the full path for each child/sub key.

golang/cosmos/x/swingset/genesis.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func NewGenesisState() *types.GenesisState {
 	return &types.GenesisState{
-		Storage: []*types.StorageEntry{},
+		Params: types.Params{},
```
JimLarson: `Params` is not nullable, so I *think* the empty structure is there implicitly.

golang/cosmos/x/vstorage/genesis.go
```diff
@@ -0,0 +1,44 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+func NewGenesisState() *types.GenesisState {
+	return &types.GenesisState{
+		Data: []*types.DataEntry{},
+	}
+}
+
+func ValidateGenesis(data *types.GenesisState) error {
```
JimLarson, line 17: You can also validate the path, e.g. `"."` should be illegal.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (k Keeper) ImportStorage(ctx sdk.Context, entries []*types.DataEntry) {
+	for _, entry := range entries {
+		// This set does the bookkeeping for us in case the entries aren't a
+		// complete tree.
+		k.SetStorage(ctx, entry.Path, entry.Value)
+	}
+}
+
+// GetData gets generic storage
```
JimLarson: Somewhere there should be documentation that the default value (at the storage level) is the empty string.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		parts := strings.Split(types.EncodedKeyToPath(iterator.Key()), types.PathSeparator)
+		keyStr := parts[len(parts)-1]
+		keys.Keys = append(keys.Keys, keyStr)
+	}
+	return &keys
+}
+
+// HasStorage tells if a given path has data.
```
JimLarson: Add a note that there can be store entries without data.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+func (k Keeper) HasStorage(ctx sdk.Context, path string) bool {
+	return k.GetData(ctx, path) != ""
+}
+
+func (k Keeper) Exists(ctx sdk.Context, path string) bool {
```
JimLarson: Consider renaming to `HasEntry`. Otherwise, explain that it detects both storage and placeholder entries.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+			k.PathToEncodedKey(path),
+			[]byte(value),
+		),
+	)
+}
+
+// SetStorage sets the data value for a path.
+//
+// Maintains the invariant: path entries exist if and only if self or some
+// descendant has non-empty storage
```
JimLarson, line 137: Thank you! Invariants that aren't documented don't get maintained.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+		if value == "" {
+			if k.HasStorage(ctx, ancestor) || k.HasKeys(ctx, ancestor) {
+				// If the key is needed, skip out.
+				return
+			}
+			// Remove the key.
+			encodedKey = types.PathToEncodedKey(ancestor)
+			store.Delete(encodedKey)
+		} else if i < len(pathComponents) && k.Exists(ctx, ancestor) {
+			// The key is present, so we can skip out.
```
JimLarson: `// The ancestor is present, implying all further ancestors are present, so we can return.`

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+		// Decide if we need to add or remove the ancestor.
+		if value == "" {
+			if k.HasStorage(ctx, ancestor) || k.HasKeys(ctx, ancestor) {
+				// If the key is needed, skip out.
+				return
+			}
+			// Remove the key.
+			encodedKey = types.PathToEncodedKey(ancestor)
+			store.Delete(encodedKey)
+		} else if i < len(pathComponents) && k.Exists(ctx, ancestor) {
```
JimLarson: The `i < len(pathComponents` is always true and can be dropped.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -0,0 +1,175 @@
...
+		store.Delete(encodedKey)
+	} else {
+		// Update the value.
+		bz := bytes.Join([][]byte{types.EncodedDataPrefix, []byte(value)}, []byte{})
+		store.Set(encodedKey, bz)
+	}
+
+	// Update our other parent keys.
+	pathComponents := strings.Split(path, types.PathSeparator)
+	for i := len(pathComponents) - 1; i >= 0; i-- {
```
JimLarson: Consider pulling the `value == ""` test outside of the loop, whose branches are specialized loops.  Might be a little clearer:

```
if value == "" {
    // delete placeholder ancestors if they're no longer needed
    for i := len(pathComponents) - 1; i >= 0; i-- {
        ancestor := componentsToPath(pathComponents[0:i])
        if k.HasStorage(ctx, ancestor) || k.HasKeys(ctx, ancestor) {
            // this and further ancestors are needed, skip out
            break
        }
        store.Delete(types.PathToEncodedKey(ancestor))
    }
} else {
    // add placeholders as needed
    for i := len(pathComponents) - 1; i >= 0; i-- {
        ancestor := componentsToPath(pathComponents[0:i])
        if k.Exists(ctx, ancestor) {
            // The ancestor exists, implying all further ancestors exist, so we can break.
            break
        }
        store.Set(types.PathToEncodedKey(ancestor), types.EncodedDataPrefix)
    }
}
```

golang/cosmos/x/vstorage/keeper/querier.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// NewQuerier is the module level router for state queries
+func NewQuerier(keeper Keeper, legacyQuerierCdc *codec.LegacyAmino) sdk.Querier {
+	return func(ctx sdk.Context, path []string, req abci.RequestQuery) (res []byte, err error) {
+		switch path[0] {
+		case QueryData:
+			return queryData(ctx, strings.Join(path[1:], "/"), req, keeper, legacyQuerierCdc)
+		case QueryChildren:
+			return queryKeys(ctx, strings.Join(path[1:], "/"), req, keeper, legacyQuerierCdc)
```
JimLarson: I'm liking "query children" as a consistent name to use throughout.

golang/cosmos/x/vstorage/keeper/querier.go
```diff
@@ -0,0 +1,65 @@
...
+			return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownRequest, "unknown vstorage query endpoint")
+		}
+	}
+}
+
+// nolint: unparam
+func queryData(ctx sdk.Context, path string, req abci.RequestQuery, keeper Keeper, legacyQuerierCdc *codec.LegacyAmino) (res []byte, err error) {
+	value := keeper.GetData(ctx, path)
+	if value == "" {
+		return []byte{}, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "could not get vstorage %+v", path)
```
JimLarson: Double check use of empty array vs `nil` below.

golang/cosmos/x/vstorage/types/path_keys.go
```diff
@@ -0,0 +1,65 @@
...
+// PathToEncodedKey converts a path to a byte slice key
+func PathToEncodedKey(path string) []byte {
+	depth := strings.Count(path, PathSeparator)
+	encodedPath := PathSeparator + path
+	if len(path) > 0 {
+		// Increment so that only the empty path is at depth 0.
+		depth += 1
+	}
+	encoded := []byte(fmt.Sprintf("%d%s", depth, encodedPath))
+	if bytes.Contains(encoded, EncodedKeySeparator) {
```
JimLarson: Consider a `ValidatePath()` function that ensures it doesn't contain nul and has no empty strings between dots, and use it on entry here and below.

JimLarson (approved):
Looks great! Review items are mostly about comments, naming, and refactoring and I trust you to apply them judiciously, but I'd also be happy to take another look before merging.

golang/cosmos/proto/agoric/swingset/swingset.proto
michaelfig: Power flags are obsolete and have no effect.  I don't intend to remove them until we implement permissionless client provisioning #4723.

golang/cosmos/proto/agoric/vstorage/query.proto
michaelfig: Mistake!  There should only be one path in the query.

golang/cosmos/x/vstorage/keeper/keeper.go
michaelfig: Great!  I used this.

golang/cosmos/x/vstorage/keeper/querier.go
michaelfig: Okay, I'll see about making that change.

PR #5466, opened 2022-05-30
Cosmos `x/vstream` implementation

michaelfig:
refs: #5487

## Description

On-chain stream publishing for the client described in [`@agoric/streams`](https://github.com/Agoric/agoric-sdk/blob/mfig-chain-streams/packages/streams/README.md#agoric-streams).  See the [`x/vstream` spec](https://github.com/Agoric/agoric-sdk/blob/mfig-vstream/golang/cosmos/x/vstream/spec/README.md)


golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,77 @@
...
+    ];
+
+    // The store name under which the cell is stored.
+    string store_name = 2 [
+        (gogoproto.jsontag)    = "store_name",
+        (gogoproto.moretags)   = "yaml:\"store_name\""
+    ];
+
+    // The subkey under which the cell is stored.
+    bytes store_subkey = 3 [
```
JimLarson, line 63: This is the subkey/child key for vstorage? Make that explicit, and be consistent with the choice of "sub" vs "child" made for vstorage.

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,77 @@
...
+    // If state=END_STATE_FAILURE, the last value is the serialised error.
+    repeated bytes values = 1;
+
+    // The block height in which this cell was last updated.
+    int64 updated_block_height = 2 [
+        (gogoproto.jsontag)    = "updated_block_height",
+        (gogoproto.moretags)   = "yaml:\"updated_block_height\""
+    ];
+
+    // The first value's sequence number.  0 means unspecified
```
JimLarson: Note that sequence numbers are contiguous throughout the entire stream.

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,77 @@
...
+        (gogoproto.moretags)   = "yaml:\"first_sequence_number\""
+    ];
+
+    // The state as of the last value.
+    EndState end_state = 4 [
+        (gogoproto.jsontag)    = "end_state",
+        (gogoproto.moretags)   = "yaml:\"end_state\""
+    ];
+
+    // The prior position in the stream.
```
JimLarson: Add `If this is the first cell in the stream, this will be the default value. All other cells will have a reference with a nonzero block height.`

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,77 @@
...
+    ];
+
+    // The subkey under which the cell is stored.
+    bytes store_subkey = 3 [
+        (gogoproto.jsontag)    = "store_subkey",
+        (gogoproto.moretags)   = "yaml:\"store_subkey\""
+    ];
+
+    // The sequence number of the value at this position.
+    uint64 sequence_number = 4 [
```
JimLarson, line 70: I first thought this was redundant with `firstSequenceNumber` in `StreamCell`, but now I see this is necessary for a reference to a fork point within a cell.

golang/cosmos/types/stream/cell.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"encoding/json"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	agoric "github.com/Agoric/agoric-sdk/golang/cosmos/types"
+)
+
+func NewStreamCell(blockHeight int64, seq uint64) StreamCell {
```
JimLarson: Document that `seq` is the first sequence number to use in the cell.

golang/cosmos/x/lien/spec/README.md
```diff
@@ -0,0 +1,13 @@
+<!--
```
JimLarson, line 1: Unrelated change, please remove.

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,77 @@
...
+      END_STATE_FINISHED = 2;
+      // END_STATE_FAILURE means this cell has terminated with an error value.
+      END_STATE_FAILURE = 3;
+    }
+
+    // Chronologically-ordered list of values for this cell.
+    // If state=END_STATE_FAILURE, the last value is the serialised error.
+    repeated bytes values = 1;
+
+    // The block height in which this cell was last updated.
```
JimLarson, line 28: Since it sounds like a cell is updated only in a single block, change this to just `block_height`.

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,77 @@
...
+        (gogoproto.moretags)   = "yaml:\"end_state\""
+    ];
+
+    // The prior position in the stream.
+    StreamPosition prior = 5 [
+      (gogoproto.nullable) = false
+    ];
+}
+
+// StreamPosition is data that refers to a particular stream cell value.
```
JimLarson, line 48: Add `The default value of this message is the equivalent of a null pointer.`

golang/cosmos/types/stream/cell.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		if err := json.Unmarshal(data, &priorCell); err != nil {
+			return nil, err
+		}
+	}
+	priorReference := NewStreamPosition(
+		priorCell.UpdatedBlockHeight,
+		state.StoreName(),
+		state.StoreSubkey(),
+		uint64(len(priorCell.Values)),
```
JimLarson: Don't you need to add `priorCell.FirstSequenceNumber`? Add a test which would have caught this.

golang/cosmos/types/stream/cell.go
```diff
@@ -0,0 +1,52 @@
...
+		StoreSubkey:    subkey,
+		SequenceNumber: seq,
+	}
+}
+
+func NewZeroStreamPosition() StreamPosition {
+	return NewStreamPosition(0, "", nil, 0)
+}
+
+func GetLatestPosition(ctx sdk.Context, state agoric.StateRef) (*StreamPosition, error) {
```
JimLarson, line 37: Document: returns position of last value in the cell, or a zero stream position if the referenced cell does not exist.

golang/cosmos/types/stream/cell.go
```diff
@@ -0,0 +1,52 @@
...
+	var priorCell StreamCell
+	if state.Exists(ctx) {
```
JimLarson: Consider testing for `!state.Exists(ctx)` and returning `NewZeroStreamPosition()`. It looks like a non-existing cell results in the return of a zero stream position, but I think it's safer to make this explicit rather than relying on undocumented, implicit behavior.

golang/cosmos/types/state_ref.go
```diff
@@ -0,0 +1,57 @@
+package types
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// An interface to allow updating KVStore-backed state and extracting proof
```
JimLarson, line 9: An interface ... to a stored StreamCell.

JimLarson (changes requested):
Preview of comments. More to come, hopefully this weekend.

golang/cosmos/proto/agoric/types/stream.proto
michaelfig: I want to specify that the value of the cell may be queried at a height > `updated_block_height`.  Just using `block_height` does not convey that meaning as clearly.

golang/cosmos/proto/agoric/types/stream.proto
michaelfig: As you discovered, this sequence number can be calculated from the `prior` value.  Removing.

golang/cosmos/proto/agoric/types/stream.proto
michaelfig: No, this is at a lower level than vstorage: the Cosmos SDK multistore.  I've noted as such.

golang/cosmos/types/stream/cell.go
michaelfig: I changed this to take a `prior` position.  It's better not to leak the sequence number abstraction.

golang/cosmos/types/stream/cell.go
michaelfig: Turns out that was redundant information that I've removed.

golang/cosmos/types/state_ref.go
michaelfig: It's not StreamCell specific.

golang/cosmos/x/lien/spec/README.md
michaelfig: I find drive-bys acceptable if it would be onerous to put them in a separate PR.

Do you want me not to make this change?  I will remove it in that case.

golang/cosmos/proto/agoric/types/stream.proto
JimLarson: Hmm. In the comment at least, instead of "was last updated", try "was produced", and see if that inspires a different name.

golang/cosmos/proto/agoric/types/stream.proto
JimLarson: Hmm, "subkey" doesn't appear to be standard cosmos terminology - the APIs just use "key".

golang/cosmos/proto/agoric/types/stream.proto
JimLarson: I see you've removed `firstSequenceNumber` from `StreamCell`. I guess it's redundant as long as you're always accessing a cell via a `StreamPosition`, but it seems a little too bold to leave this essential information only extrinsic. Consider restoring the field.

golang/cosmos/types/stream/cell.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"encoding/json"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	agoric "github.com/Agoric/agoric-sdk/golang/cosmos/types"
+)
+
+// NewStreamCell creates a new StreamCell at blockHeight with the specified
+// prior position.
```
JimLarson, line 12: Use a zero stream position if this is the first cell in the stream.

golang/cosmos/types/stream/cell.go
```diff
@@ -0,0 +1,59 @@
...
+		SequenceNumber: seq,
+	}
+}
+
+func NewZeroStreamPosition() StreamPosition {
+	return NewStreamPosition(0, "", nil, 0)
+}
+
+// GetLatestPosition returns the position of the last value in the cell, or a
+// zero stream position if the referenced cell does not exist.
```
JimLarson, line 36: Since a StateRef could reference anything, document that it is expected to reference a StreamCell.

Add a note that it will unconditionally return the position of the last value in the cell regardless of the cell's end state. The cell end state should be checked to ensure the position is meaningful, e.g. not `END_STATE_FAILURE`.

golang/cosmos/types/stream/operation.go
```diff
@@ -0,0 +1,188 @@
...
+
+var _, _ StreamCellUpdater = AppendStreamCellUpdater{}, FailStreamCellUpdater{}
+
+type StreamOperation struct {
+	state   agoric.StateRef
+	Updater StreamCellUpdater
+	Prior   StreamPosition
+}
+
+func NewStreamOperation(ctx sdk.Context, state agoric.StateRef, updater StreamCellUpdater) StreamOperation {
```
JimLarson, line 25: Document that `state` is expected to reference a `StreamCell`.

golang/cosmos/types/stream/operation.go
```diff
@@ -0,0 +1,188 @@
...
+	return StreamOperation{
+		state:   state,
+		Updater: updater,
+		Prior:   NewZeroStreamPosition(),
+	}
+}
+
+func (so StreamOperation) GetLatestPosition(ctx sdk.Context) (*StreamPosition, error) {
```
JimLarson, line 33: Note that it returns the position of the last value regardless of the end state.

golang/cosmos/types/stream/operation.go
```diff
@@ -0,0 +1,188 @@
...
+	return GetLatestPosition(ctx, so.state)
+}
+
+// LoadAndCheckHead returns a head satisfying prior that we can mutate.
```
JimLarson, line 37: I don't understand "head satisfying prior". How about "Given the position of the last value in the stream (or a zero position if the stream is empty), returns a StreamCell to which new stream values can be appended. Does not check whether an existing cell is failed or finished. This cannot be used to fork a stream - the StreamOperation's StateRef must match the store and key of prior". (Or skip the checks and allow this to create a fork.)

golang/cosmos/types/stream/streamer.go
```diff
@@ -0,0 +1,35 @@
+package stream
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	agoric "github.com/Agoric/agoric-sdk/golang/cosmos/types"
+)
+
+// Streamer defines an interface that allows streaming to an arbitrary StateRef.
```
JimLarson, line 9: StateRef must reference a StreamCell (or empty).

golang/cosmos/types/stream/operation.go
```diff
@@ -0,0 +1,188 @@
...
+type StreamCellUpdater interface {
+	Update(cell StreamCell) error
+}
+
+var _, _ StreamCellUpdater = AppendStreamCellUpdater{}, FailStreamCellUpdater{}
+
+type StreamOperation struct {
+	state   agoric.StateRef
+	Updater StreamCellUpdater
+	Prior   StreamPosition
```
JimLarson, line 22: Is the `Prior` field used? It seems redundant with the `priorPos` argument to `StreamOperation.Commit()`.

golang/cosmos/x/lien/spec/README.md
JimLarson: Remove. I'll be updating `x/lien` soon any you can have me add it in that PR.

golang/cosmos/types/stream/cell.go
JimLarson: Not fixed yet.

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,74 @@
+syntax = "proto3";
+package agoric.types;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/Agoric/agoric-sdk/golang/cosmos/types/stream";
+
+// StreamCell represents one publish state in a vstream stream.
```
JimLarson, line 8: "published". Consider instead "StreamCell contains the stream values that are published during the execution of a single block. Each StreamCell contains a reference to the prior cell in the stream.

golang/cosmos/proto/agoric/types/stream.proto
```diff
@@ -0,0 +1,74 @@
+syntax = "proto3";
+package agoric.types;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/Agoric/agoric-sdk/golang/cosmos/types/stream";
+
```
JimLarson, line 7: Introductory comment: "A stream is a sequence of arbitrary bytes values used by publishers to communicate with consumers. Values in a stream are indexed by 1-based contiguous sequence numbers. A stream is stored as a sequence of StreamCells. StreamCells are typically written to the same key in a KVStore, relying on reading at old block heights to see older values."

golang/cosmos/types/state_ref.go
JimLarson: Okay, but see below for comments that make it explicit when we are expecting it to reference a StreamCell. When do you think the SDK might start using generics?

```
type StateRef[T Jsoner] interface {
    Read(ctx sdk.Context) (T, error)
    Write(ctx sdkContext, value T) error
    ...
}
```

golang/cosmos/types/state_ref.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type KVStoreStateRef struct {
+	storeKey sdk.StoreKey
+	subkey   []byte
+}
+
+var _ StateRef = KVStoreStateRef{}
+
+func NewKVStoreStateRef(storeKey sdk.StoreKey, subkey []byte) *KVStoreStateRef {
```
JimLarson, line 27: Consider not pointer-izing.

golang/cosmos/types/stream/operation.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type StreamCellUpdater interface {
+	Update(cell StreamCell) error
+}
+
+var _, _ StreamCellUpdater = AppendStreamCellUpdater{}, FailStreamCellUpdater{}
+
+type StreamOperation struct {
+	state   agoric.StateRef
```
JimLarson, line 20: Document that `state` is expected to reference a `StreamCell`.

JimLarson (commented):
Here's my final comments on the code. I'll address the spec documents in another PR - probably after I've seen more of how this all will be used.

michaelfig:
@gibson042 did a great job of implementing this feature mostly in JS with a light touch in Go, so this PR is superseded.  Closing.

PR #5481, opened 2022-05-31
feat: add fakeDurable option to assist durability conversion

### LINKED ISSUE #5454: need fakeDurable option
Current transitive insistence on durability inhibits incremental development. An explicit `fakeDurable` option to our big stores and our durable kinds would let us manage the horizon during development in a more incremental manner. A fakeDurable thing (instance or store) would be accepted as durable during development, but would accept non-durable things.

Until genuine transitivity is reached, things are still not upgradable. Upgrade will need to test for that.

FUDCo:
This change adds a `fakeDurable` option to the `makeScalaraBigFooStore` and `defineDurableKind`/`defineDurableKindMulti` calls.  Collections or virtual objects that result from using this option are not actually durable but will appear durable for purposes of being inserted into another durable collection or durable object.  Note that such objects will not survive upgrade -- they will be discarded on upgrade like any other non-durable object -- but can be used in code that is in the midst of conversion from using purely virtual objects to durable objects. By relaxing the requirement for transitive durability, we hope to ease the conversion effort by enabling objects that will eventually be durable to be used as if they were durable even though they're not durable yet.

Closes #5454

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -792,8 +792,17 @@ export function makeCollectionManager(
    */
   function makeScalarBigMapStore(
     label = 'map',
-    { keySchema = M.scalar(), valueSchema = undefined, durable = false } = {},
+    {
+      keySchema = M.scalar(),
+      valueSchema = undefined,
+      durable = false,
+      fakeDurable = false,
+    } = {},
   ) {
+    assert(
+      !durable || !fakeDurable,
+      'durable and fakeDurable are mutually exclusive',
+    );
     const kindName = durable ? 'scalarDurableMapStore' : 'scalarMapStore';
     const [vobjID, collection] = makeCollection(
       label,
@@ -803,6 +812,9 @@ export function makeCollectionManager(
     );
     const store = collectionToMapStore(collection);
     registerValue(vobjID, store, false);
+    if (fakeDurable) {
+      vrm.registerFakeDurable(vobjID);
+    }
     return store;
   }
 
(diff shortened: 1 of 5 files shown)


packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -890,8 +926,17 @@ export function makeCollectionManager(
...
+      keySchema = M.scalar(),
+      valueSchema = undefined,
+      durable = false,
+      fakeDurable = false,
+    } = {},
   ) {
+    assert(
+      !durable || !fakeDurable,
+      'durable and fakeDurable are mutually exclusive',
+    );
```
erights, line 935: Amount of duplicated code here is unfortunate. If there's a way to consolidate without making other things noticeably worse, it's probably worth it.

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -572,7 +572,18 @@ export function makeVirtualObjectManager(
...
   ) {
-    const finish = options ? options.finish : undefined;
+    let finish;
+    let fakeDurable;
+    if (options) {
+      ({ finish, fakeDurable } = options);
+    }
+    if (fakeDurable) {
+      assert(
+        durable,
```
erights, line 582: A bit surprising since collectionManager uses flags of exactly these two names but insists that those are mutually exclusive. I don't have a suggestion, so it you don't think of an improvement, you can leave this as is.

erights (approved):
LGTM, thanks!!!!

warner (approved):
This does what it says on the tin, so I'm ok with it landing.

But I'm not excited about this surviving in our codebase for very long: I'd like the companion ticket (about making it disabled by default) to consider removing it entirely once we've completed the durabilization transition. I'm hoping that future authors will be able to avoid the incremental conversion process.

PR #5487, opened 2022-06-01
feat(casting): `@agoric/casting` MVP

michaelfig:
refs: #5385

## Description

This is a complete, but polling-based chain casting follower.  Being primarily event-driven is a goal for a future PR.
The best description is in https://github.com/Agoric/agoric-sdk/tree/mfig-chain-streams/packages/casting/README.md

As further PRs in this area are merged, the intent is to update this package to take advantage of new features.  Notably, the #5466 `x/vcast` module will enable full chain support of the abstractions in #5418.


golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -302,6 +303,19 @@ func (k Keeper) GetKeys(ctx sdk.Context, path string) *types.Keys {
 	return &keys
 }
 
+func (k Keeper) SetStorageAndNotify(ctx sdk.Context, path, value string) {
```
dckc: do we have go test coverage tooling?

packages/chain-streams/node-fetch-shim.js
```diff
@@ -0,0 +1,4 @@
+/* global globalThis */
+import fetch from 'node-fetch';
+
+globalThis.fetch = fetch;
```
dckc: ouch. Why add a powerful global?

dckc: hm. I suppose this only words in compartments where `globalThis` isn't frozen; i.e. where OCap discipline isn't enforced.

packages/chain-streams/test/fake-rpc-server.js
```diff
@@ -0,0 +1,230 @@
...
+};
+
+export const startFakeServer = (t, fakeValues) => {
+  const { log = console.log } = t;
+  const { serialize } = makeMarshal();
+  lastPort += 1;
+  const PORT = lastPort;
+  return new Promise(resolve => {
+    log('starting http server on port', PORT);
+    const app = express();
```
dckc: exporting a `startFakeServer` that closes over `express` doesn't follow OCap discipline. How about putting `app` on `t.context`? so the caller would do `t.context.app = express()` in `test.before()`?

packages/chain-streams/test/fake-rpc-server.js
```diff
@@ -0,0 +1,230 @@
...
+    '[slot(0,"Alleged: for testing Justin")]',
+  ],
+  // Tests https://github.com/endojs/endo/issues/1185 fix
+  [
+    '[{"@qclass":"slot","iface":"Alleged: for testing Justin","index":0},{"@qclass":"slot","index":0}]',
+    '[slot(0,"Alleged: for testing Justin"),slot(0)]',
+  ],
+]);
+
+export const develop = async () => {
```
dckc: Where is this `develop()` thing used?

packages/chain-streams/src/defaults.js
```diff
@@ -0,0 +1,54 @@
...
+import { Far } from '@endo/far';
+import { makeMarshal } from '@endo/marshal';
+
+/**
+ * Resolve a Promise after a given number of milliseconds.
+ *
+ * @param {number} ms
+ * @returns {Promise<void>}
+ */
+export const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
```
dckc: > No modules should export powerful objects (for example, objects that close over fs).
 -- https://github.com/Agoric/agoric-sdk/wiki/OCap-Discipline

End's [node-powers.js](https://github.com/endojs/endo/blob/master/packages/compartment-mapper/node-powers.js) is a nice approach.

packages/chain-streams/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+This [Agoric](https://agoric.com) Chain Streams package consumes data server
+publication leaders in a flexible, future-proof way.
+
+TL;DR: You can run `yarn demo`, or to consume a mailbox stream do:
+```sh
+npx agoric stream -Bhttp://devnet.agoric.net/network-config :mailbox.agoric1foobarbaz -otext
+```
+
+An example of following an on-chain mailbox in codd (using this package) is:
```
kriskowal:
```suggestion
An example of following an on-chain mailbox in code (using this package) is:
```

packages/agoric-cli/src/stream.js
```diff
@@ -0,0 +1,120 @@
...
+    },
+    keepPolling: async () => {
+      let toSleep = sleep * 1000;
+      if (toSleep <= 0) {
+        toSleep = (5 + Math.random()) * 1000;
+      }
+      await delay(toSleep);
+      return true;
+    },
+  };
```
kriskowal: This might be one of those places where exponential back-off with full jitter is appropriate. https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

packages/chain-streams/src/defaults.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Report an error, then retry the leader operation after a second or two.
+ *
+ * @param {any} err
+ * @returns {Promise<void>}
+ */
+export const DEFAULT_RETRY_CALLBACK = err => {
+  console.warn('retrying after error', err);
+  return delay(1000 + Math.random() * 1000);
+};
```
kriskowal: Consider threading the attempt number so this can be `delay(Math.random() * Math.min(cap, base * 2 ** attempt))`

packages/chain-streams/src/leader.js
```diff
@@ -0,0 +1,65 @@
...
+  const { retryCallback = DEFAULT_RETRY_CALLBACK } = leaderOptions;
+
+  // Shuffle the RPC addresses, so that we don't always hit the same one as all
+  // our peers.
+  shuffle(endpoints);
+
+  let lastRespondingEndpointIndex = 0;
+
+  /** @type {import('./types.js').ChainLeader} */
+  const leader = Far('round robin leader', {
```
kriskowal: If you’re measuring latency and error rate, [The Power of Two Random Choices](https://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf)

kriskowal (approved):
Regarding naming, I think “mailbox”, “publication”, “rendezvous”, “drop”, or “feed” might be more appropriate. I would like to reserve “stream” to mean bidi async iterator generally, which is what you’re returning from utility methods off your stream.

michaelfig:
> Regarding naming, I think “mailbox”, “publication”, “rendezvous”, “drop”, or “feed” might be more appropriate. I would like to reserve “stream” to mean bidi async iterator generally, which is what you’re returning from utility methods off your stream.

Thanks for the feedback.  I think I've settled on `@agoric/casting`.

packages/chain-streams/src/leader.js
michaelfig: I don't want to read a paper right now.  Can you be more specific as to what you're recommending, so that I can at least capture it as a TODO?

kriskowal: Round-robin will be fine. Two random produces a slightly better load distribution when some hosts are more or less responsive than others, and does so in constant time, without maintaining a logarithmic heap.

```js
const i1 = Math.floor(Math.random() * options.length)
const i2 = Math.floor(i1 + 1 + Math.random() * (options.length - 1)) % options.length;
if (latencies[i1] < latencies[i2]) {
  return options[i1];
} else {
  return options[i2];
}
```

Latencies are initialized to Infinity and updated with the latency of the most recent request/response (before consuming the response body).

kriskowal (approved):
Reviewed the name change.

JimLarson (approved):
Last retro-review: golang code looks good to me.

dckc (commented):
Oops... I just found some pending comments. Not critical, I suppose.

PR #5488, opened 2022-06-01
fix(swingset): update lmdb

### LINKED ISSUE #5031: non-deterministic crash of single node under load
## Describe the bug
The benchmark machine has not been able to run a single full loadgen test, failing at varying times while under load. It appears the kernel process gets killed with either `SIGSEGV` or `SIGABRT`, most times printing nothing, other printing a message like `free(): invalid next size (fast)` or `double free or corruption (out)`. These actually happened on the same revision (`6fcb5462e508`) when attempting to run the loadgen again.

## To Reproduce
1. Login to [benchmark machine](https://github.com/Agoric/agoric-private/wiki/Benchmark-machine-for-loadgen-tests)
2. Check logs of recent `daily-perf` runs, including revision above

## Expected behavior
No crash

## Platform Environment
 - what OS are you using? what version of Node.js? Node 16.14.0 (same behavior on Node 16.13.1). go 1.17.7. 
 - is there anything special/unusual about your platform? Loadgen docker container
 - what version of the Agoric-SDK are you using? `6fcb5462e508` and others

mhofman:
closes: #5031
best reviewed commit-by-commit

## Description

Switches from node-lmdb to [lmbd-js](https://github.com/DoctorEvidence/lmdb-js) which seem more actively maintained. Disable `useWritemap` option which seem to be causing some issues (but not the only cause of them), as it seems no longer necessary.

This PR does the minimal amount of code changes in the swingstore by synchronous creating a transaction. This works with the new API because we already ensured that we only ever had one transaction at a time. Once inside a transaction, read after writes are guaranteed to contain the data from the transaction.

However to adapt to the more asynchronous nature of lmdb-js, this PR updates the API of the swingstore to make `commit` and `close` asynchronous. While closing a synchronous transaction is not actually asynchronous, it would be too much of a hack to make it happen synchronously. Since the committers were already all asynchronous, this was a low impact change.

`lmdb-js` also seem to have a couple differences in behavior, including not passing through the `mapSize` option or encoding strings (both keys and values) as UTF-8. The former will result in a non preallocated DB file (previously sparse) growing in size with usage. The latter has no practical impact since the LMDB key size was also increased.

Drive by fix of `getKeys` iterator interruption which were previously not closing cursors. I've decided to pass errors through, as these really should not happen.

### Security Considerations

The max key size logic and checks in collections is now no longer matching reality as it assumed a UTF-16 encoding of strings. However since lmdb-js bumps the max key size, it is impossible to generate keys that are too large, and a new test asserts this. 

The collections documentation was amended.

### Testing Considerations

Fixed some swingstore and swingset tests to handle async commit, and handle parallel execution.
Added a test for lmdb max key size.
Ran a full daily loadgen test on my machine, which used to crash reliably.


packages/SwingSet/docs/run-policy.md
```diff
@@ -88,7 +88,7 @@ while(1) {
   processInboundIO();
   const policy = make100CrankPolicy();
   await controller.run(policy);
-  commit();
+  await commit();
```
michaelfig, line 91: Please also update callers of `commit()` in `packages/solo/src` and `packages/cosmic-swingset/src`.

mhofman: They were already `await`ing!

michaelfig:
Here's what I see missing `await`:
https://github.com/Agoric/agoric-sdk/blob/01cdbc9ff3dfee1fc9dad5c4b19a0c41ff697787/packages/solo/src/start.js#L251
https://github.com/Agoric/agoric-sdk/blob/3c7e57b8f62c0b93660dd57c002ffb96c2cd4137/packages/solo/src/reset-state.js#L14

mhofman:
> Here's what I see missing `await`:
> 
> https://github.com/Agoric/agoric-sdk/blob/01cdbc9ff3dfee1fc9dad5c4b19a0c41ff697787/packages/solo/src/start.js#L251
> 
That one is a JSONStore, which is a different API. I'm open to make it apparent async as well for consistency.

> 
> https://github.com/Agoric/agoric-sdk/blob/3c7e57b8f62c0b93660dd57c002ffb96c2cd4137/packages/solo/src/reset-state.js#L14

This one has already been updated: https://github.com/Agoric/agoric-sdk/blob/35dc1da06b45113ad4f4f6183d7965b025b99259/packages/solo/src/reset-state.js#L14

mhofman:
> That one is a JSONStore, which is a different API. I'm open to make it apparent async as well for consistency.

@michaelfig I did make it consistent since it didn't really hurt. The only somewhat similar API is now the chain-store, but that one seem to rely on a synchronous `commit` right now, and I didn't feel like untangling that.

patches/ordered-binary+1.2.5.patch
```diff
@@ -0,0 +1,181 @@
+diff --git a/node_modules/ordered-binary/index.js b/node_modules/ordered-binary/index.js
```
mhofman: This package is used for the LMDB key encoding, but is incompatible with the SES shim. I'm still unsure why I needed to patch it as it should have loaded pre-lockdown. We could probably try to figure out a way to avoid it entirely since we only use string keys, but patching was the path of least resistance for me.

mhofman: It was another case of `lmdb` not being loaded early enough.

packages/cosmic-swingset/src/entrypoint.js
```diff
@@ -2,8 +2,8 @@
 
 import '@endo/init/pre-bundle-source.js';
 
-// import node-lmdb early to work around SES incompatibility
-import 'node-lmdb';
+// import lmdb early to work around SES incompatibility
+import 'lmdb';
```
warner, line 6: Do we know that `lmdb` has the same incompatibility as `node-lmdb`? We should try moving the import later and see if it has problems.

packages/solo/src/entrypoint.js
```diff
@@ -2,8 +2,8 @@
 
 import '@endo/init/pre-bundle-source.js';
 
-// import node-lmdb early to work around SES incompatibility
-import 'node-lmdb';
+// import lmdb early to work around SES incompatibility
```
warner, line 5: same

packages/solo/src/pipe-entrypoint.js
```diff
@@ -1,7 +1,7 @@
 /* global process */
 // @ts-check
 import '@endo/init/pre-bundle-source.js';
-import 'node-lmdb';
+import 'lmdb';
```
warner, line 4: same

packages/swing-store/test/test-state.js
```diff
@@ -1,10 +1,11 @@
+// @ts-check
+
 // import LMDB before SES lockdown, as workaround for
 // https://github.com/Agoric/SES-shim/issues/308
-import 'node-lmdb';
+import 'lmdb';
```
warner, line 5: same, if possible let's take advantage of this change to confine `lmdb` slightly better

packages/SwingSet/src/kernel/state/storageWrapper.js
```diff
@@ -24,30 +26,74 @@ import { insistStorageAPI } from '../../lib/storageAPI.js';
...
-      yield result;
-    } else if (v1.value === v2.value) {
-      const result = v1.value;
-      v1 = it1.next();
-      v2 = it2.next();
-      yield result;
-    } else {
-      const result = v2.value;
-      v2 = it2.next();
+  /** @type {IteratorResult<any> | null} */
```
warner, line 29: Please coordinate with @gibson042 since his #5294 moves this function around.

packages/SwingSet/test/vat-warehouse/test-warehouse.js
```diff
@@ -1,5 +1,10 @@
 // @ts-check
 
+import '@endo/init/pre-bundle-source.js';
+
+// import lmdb early to work around SES incompatibility
```
warner, line 5: same

packages/access-token/src/json-store.js
```diff
@@ -198,7 +198,7 @@ function makeJSONStore(dirPath, forceReset = false) {
   /**
    * Commit unsaved changes.
    */
-  function commit() {
+  async function commit() {
```
warner, line 201: Does this need to be async? This little "json store" doesn't overlap much with the SwingStore, but I suppose if there's any chance of both appearing in the same program, I can see a consistency argument.

packages/swing-store/src/swingStore.js
```diff
@@ -150,34 +153,39 @@ function makeSwingStore(dirPath, forceReset, options) {
...
   });
 
   function ensureTxn() {
-    if (!txn) {
+    assert(db);
+    if (!txnDone && !txnFinish) {
       trace('begin-tx');
-      txn = lmdbEnv.beginTxn();
+      txnDone = db.transactionSync(
+        () =>
```
warner, line 181: interesting API they've got there

packages/swing-store/src/swingStore.js
```diff
@@ -224,14 +231,14 @@ function makeSwingStore(dirPath, forceReset, options) {
...
-      yield key;
-      key = cursor.goToNext();
-    }
-    cursor.close();
+    /** @type {import("lmdb").RangeOptions} */
+    const rangeOptions = {};
+    if (start) rangeOptions.start = start;
+    if (end) rangeOptions.end = end;
+
+    const keys = ensureTxn().getKeys(rangeOptions);
```
warner, line 243: Is `keys` an iterator, or an array? I want to make sure we aren't holding a gigantic list of keys in RAM here.

packages/swing-store/test/test-state.js
```diff
@@ -70,27 +69,44 @@ test('in-memory kvStore read/write', t => {
...
+  // for vat store prefixes.
+  // However some unicode codepoints may end up serialized as 3 bytes with UTF-8
+  // This tests that no matter what, we can write 254 unicode characters in the
+  // 0x0800 - 0xFFFF range (single UTF-16 codepoint, but 3 byte UTF-8).
+
+  const [dbDir, cleanup] = await tmpDir('testdb');
+  t.teardown(cleanup);
+  t.is(isSwingStore(dbDir), false);
+  const store = initSwingStore(dbDir);
+  store.kvStore.set('€'.repeat(254), 'Money!');
```
warner, line 104: Let's assert both sides of the size limit (i.e. assert that 255 3-byte-UTF-8 chars fails), since the success/failure of writes needs to be part of consensus. This isn't an exact constraint (we'd need a mix of encoded sizes, plus a healthy variety of surrogate pairs, to exercise this thoroughly), but I think it's good to be in the habit of knowing what the precise rules are.

I think that means this DB change is consensus-breaking (there are non-trivial unicode strings which would pass the earlier encoding scheme's limit but not the new one, or vice versa), but I'm ok with making that now.. better than later.

packages/swingset-runner/src/kerneldump-entrypoint.js
```diff
@@ -6,7 +6,7 @@
  */
 
 // LMDB bindings need to be imported before lockdown.
-import 'node-lmdb';
+import 'lmdb';
```
warner, line 9: same

warner (approved):
Seems good to me, but please get @FUDCo 's approval too, he knows the DB layer much better than me.

packages/cosmic-swingset/src/entrypoint.js
mhofman: Not the same, but a different one. I know how to fix it (it uses an direct eval as a sort of runtime macro/codegen), it wasn't worth the effort in this PR.

packages/access-token/src/json-store.js
mhofman: Yeah I wasn't sure. Can easily revert this specific commit.

packages/swing-store/src/swingStore.js
mhofman: You're supposed to do all operations within the callback, but that obviously doesn't fit our usage pattern.

packages/swing-store/src/swingStore.js
mhofman: A proper iterator! With lots of logic regarding transactions boundaries.

packages/swing-store/test/test-state.js
mhofman: > 255 3-byte-UTF-8 chars fails

It won't fail. The upper limit is actually 1971 bytes I believe.

> I think that means this DB change is consensus-breaking

I don't believe so. Nothing is writing large keys since the only user controlled key size is the collections keys which is independently checked.

warner: Ah, ok, so the previous limits were (pardon my js/python pseudocode) `str.length < X && len(str.encode('utf-16')) < Y`, and the new limits are `str.length < X && len(str.encode('utf-8') < Z`, but for all possible strings, the `.length < X` limit (imposed by liveslots) is tighter than the encoded limit (imposed lower down). So landing this can't make any failing userspace behavior start to succeed, or succeeding userspace behavior start to fail.

As long as that criteria is met, I guess we don't need more stringent testing at this layer.

packages/access-token/src/json-store.js
warner: I guess I'm ambivalent.

packages/cosmic-swingset/src/entrypoint.js
warner: aye, then ignore my other comments

packages/swing-store/test/test-state.js
mhofman: Yes, this test is specifically about checking this assumption. That the worst case `str = threeByteUtfChar.repeat(X)` then `len(str.encode('utf-8')) < Z`. Aka `3 * X < Z`

packages/swing-store/src/swingStore.js
```diff
@@ -224,14 +231,14 @@ function makeSwingStore(dirPath, forceReset, options) {
...
-    const cursor = new lmdb.Cursor(txn, dbi);
-    let key = start === '' ? cursor.goToFirst() : cursor.goToRange(start);
-    while (key && (end === '' || key < end)) {
-      yield key;
-      key = cursor.goToNext();
-    }
-    cursor.close();
+    /** @type {import("lmdb").RangeOptions} */
+    const rangeOptions = {};
+    if (start) rangeOptions.start = start;
```
FUDCo: I'm not sure why our linter didn't pick this up, but please write this as
```
if (start) {
  rangeOptions.start = start;
}
```
-- always use braces
-- no putting the if consequence on same line as the conditional

FUDCo (approved):
One minor stylistic quibble but aside from that this looks very good.  I'm surprised (and happy) that the changes in the underlying LMDB API could be accommodated with such a light touch.

PR #5492, opened 2022-06-01
fix: addLiquidity wasn't validating brand of want

### LINKED ISSUE #5293: adding liquidity with bad want brand leads to confused pool allocation
## Describe the bug

I gave the wrong brand in the `want` of my `addLiquidity` offer, and as a result, not only did I get an error, but I struggled to add any liquidity after that... on examining the pool allocation, we found it had liquidity tokens, not the Central and Secondary tokens.

## To Reproduce

This is what I did, anyway:

```
command[78] proposal = {give: {Central: mk(RUN, 500n), Secondary: mk(IbcATOM, 500n)}, want: { Liquidity: mk(RUN, 0n) }}
history[78] {"give":{"Central":{"brand":[Object Alleged: RUN brand]{},"value":500000000n},"Secondary":{"brand":[Object Alleged: IbcATOM brand]{},"value":5000000n}},"want":{"Liquidity":{"brand":[Object Alleged: RUN brand]{},"value":0n}}}
command[79] E(zoe).offer(E(ammAPI).makeAddLiquidityInvitation(), proposal, {Central: RUN.pmt, Secondary: IbcATOM.pmt})
history[79] [Object Alleged: userSeat]{}
command[80] E(history[79]).getOfferResult()
history[80] Promise.reject("RangeError: Brands in left (an object) and right (an object) should match but do not")

...

command[104] E(ammAPI).getPoolAllocation(IbcATOM.brand)
history[104] {"Liquidity":{"brand":[Object Alleged: IbcATOMLiquidity brand]{},"value":500000000n}}
```

## Expected behavior

Bad inputs should not lead to any state changes.

## Platform Environment

devnet - agoricdev10

_escalated from https://github.com/Agoric/agoric-sdk/issues/5131#issuecomment-1117496634_

Chris-Hibbert:
closes: #5293

## Description

`addLiquidity()` wasn't validating the brand of want, which was allowing it to increment `liqTokenSupply` before failing.

I also moved the decrement to `liqTokenSupply` in `removeLiquidity()` after
`reallocate()`, since that's what will fail if those brands aren't correct.

### Security Considerations

No security impact.

No doc changes required

### Testing Considerations

added a test for `addLiquidity()`.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/pool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/pool.js
@@ -238,8 +238,6 @@ const poolBehavior = {
       ),
     );
 
-    state.liqTokenSupply -= liquidityValueIn;
-
     poolSeat.incrementBy(
       userSeat.decrementBy(harden({ Liquidity: liquidityIn })),
     );
@@ -252,6 +250,7 @@ const poolBehavior = {
       ),
     );
     state.zcf.reallocate(userSeat, poolSeat);
+    state.liqTokenSupply -= liquidityValueIn;
 
     userSeat.exit();
     updateUpdaterState(state.updater, facets.pool);
--- a/packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
@@ -22,9 +22,17 @@ const makeMakeAddLiquidityInvitation = (zcf, getPool) => {
         Secondary: null,
       },
     });
+
     // Get the brand of the secondary token so we can identify the liquidity pool.
     const secondaryBrand = seat.getProposal().give.Secondary.brand;
+
     const pool = getPool(secondaryBrand);
+    const liquidityBrand = pool.getLiquidityIssuer().getBrand();
+    assert(
+      seat.getProposal().want.Liquidity.brand === liquidityBrand,
+      `liquidity brand must be ${liquidityBrand}`,
+    );
+
     return pool.addLiquidity(seat);
   };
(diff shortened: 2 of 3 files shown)


packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -252,6 +250,7 @@ const poolBehavior = {
       ),
     );
     state.zcf.reallocate(userSeat, poolSeat);
+    state.liqTokenSupply -= liquidityValueIn;
```
turadg, line 253: would this error have been prevented if `liqTokenSupply` were an amount?

turadg (commented):
I wonder if should do more to ensure "Bad inputs should not lead to any state changes."

Want to discuss before approving.

packages/run-protocol/src/vpool-xyk-amm/pool.js
Chris-Hibbert: No, the issue is that it's an accounting number that isn't part of the reallocate. If it were included in the reallocate, then it could have been atomic with the other state changes.

Since it's tracking the outstanding liquidity (held by liquidity providers) it doesn't seem possible to have zoe manage conservation and atomicity.

Chris-Hibbert (commented):
> I wonder if should do more to ensure "Bad inputs should not lead to any state changes."

I'm looking for general approaches to help with that. This example didn't directly suggest solutions, but I have a couple of other issues to address, and I'm going to try to generalize and think about systemic tooling that would help.

turadg (approved):

PR #5495, opened 2022-06-01
eslint warn floating promises

### LINKED ISSUE #5407: lint floating promises

turadg:
closes: #5407

## Description

Enable [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md). This defers solving all the floating promises by using `warn` level instead of `error`. Eventually we'll want to `error` but I propose we get there incrementally, distributed among teams.

This [rule from @typescript-eslint](https://typescript-eslint.io/rules/) is the first we've enabled to require type information (by setting `parserOptions.project`) . That increases the time cost of linting. Measured with `time yarn lerna run --no-bail lint:eslint` on my machine:
- before: 84.86s user 5.35s system 454% cpu 19.827 total (1.4min)
- after: 824.27s user 135.32s system 556% cpu 2:52.39 total (14min)

In CI the lint job went from [5 min before](https://github.com/Agoric/agoric-sdk/runs/6693643820?check_suite_focus=true) to [21 min after](https://github.com/Agoric/agoric-sdk/runs/6694898124?check_suite_focus=true). To prevent developers having to wait longer for PRs to complete, this splits the lint job into two `lint-primary` and `lint-rest`. 

The comments in that job config note that performance could be greatly improved if the linter were able to [reuse type information across projects](https://github.com/typescript-eslint/typescript-eslint/issues/2094). Alternately we could [unify the packages into one TypeScript project](https://github.com/Agoric/agoric-sdk/pull/4404).

I also tried cutting the lint job time by shortening the setup step, replacing `restore-node` with `setup-node` and yarn install from cache. This cut [5min setup time](https://github.com/Agoric/agoric-sdk/runs/6710173722?check_suite_focus=true) down to [2min](https://github.com/Agoric/agoric-sdk/runs/6710773587?check_suite_focus=true).  (Saving 6min compute over the two jobs.) But it failed because there are build artifacts the linting depends on, such as:
- eslint `import/no-unresolved` fails when `@agoric/ui-components` isn't built
- typescript cannot find module '../bundles/bundle-contractFacet.js' in ./zoe/tools/fakeVatAdmin.js (maybe this one should use bundleCache instead?)

### Security Considerations

--

Notify developers that they'll begin seeing new warnings and that the Github PR annotations may show many more in the code review web UI.

### Testing Considerations

Verified that lint jobs are not the slowest zebra: iIn last run lint-primary (10m) and lint-rest (12m) were both less than test-cosmic-swingset (13m) and test-swingset3 (13min).

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,18 +1,23 @@
 /* eslint-env node */
+const process = require('process');
+const lintTypes = !!process.env.AGORIC_ESLINT_TYPES;
+
 module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
-  parserOptions: {
-    sourceType: 'module',
-    project: [
-      './packages/*/jsconfig.json',
-      './packages/*/tsconfig.json',
-      './packages/wallet/*/jsconfig.json',
-      './tsconfig.json',
-    ],
-    tsconfigRootDir: __dirname,
-    extraFileExtensions: ['.cjs'],
-  },
+  parserOptions: lintTypes
+    ? {
+        sourceType: 'module',
+        project: [
+          './packages/*/jsconfig.json',
+          './packages/*/tsconfig.json',
+          './packages/wallet/*/jsconfig.json',
+          './tsconfig.json',
+        ],
+        tsconfigRootDir: __dirname,
+        extraFileExtensions: ['.cjs'],
+      }
+    : undefined,
   plugins: [
     '@jessie.js/eslint-plugin',
     '@typescript-eslint',
@@ -22,7 +27,7 @@ module.exports = {
   extends: ['@agoric'],
   rules: {
     '@typescript-eslint/prefer-ts-expect-error': 'warn',
-    '@typescript-eslint/no-floating-promises': 'warn',
+    '@typescript-eslint/no-floating-promises': lintTypes ? 'warn' : 'off',
     // so that floating-promises can be explicitly permitted with void operator
     'no-void': ['error', { allowAsStatement: true }],
     'jsdoc/no-multi-asterisks': 'off',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -2,6 +2,17 @@
 module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
+  parserOptions: {
+    sourceType: 'module',
+    project: [
+      './packages/*/jsconfig.json',
+      './packages/*/tsconfig.json',
+      './packages/wallet/*/jsconfig.json',
+      './tsconfig.json',
+    ],
+    tsconfigRootDir: __dirname,
+    extraFileExtensions: ['.cjs'],
+  },
   plugins: [
     '@jessie.js/eslint-plugin',
     '@typescript-eslint',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -22,6 +22,9 @@ module.exports = {
   extends: ['@agoric'],
   rules: {
     '@typescript-eslint/prefer-ts-expect-error': 'warn',
+    '@typescript-eslint/no-floating-promises': 'warn',
+    // so that floating-promises can be explicitly permitted with void operator
+    'no-void': ['error', { allowAsStatement: true }],
     'jsdoc/no-multi-asterisks': 'off',
     'jsdoc/multiline-blocks': 'off',
     // Use these rules to warn about JSDoc type problems, such as after
--- a/.github/workflows/test-all-packages.yml
+++ b/.github/workflows/test-all-packages.yml
@@ -23,17 +23,40 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-  lint:
+  ##################
+  # Lint tests
+  # We run per package bc of https://github.com/typescript-eslint/typescript-eslint/issues/1192
+  # We split the package tests into two jobs because type linting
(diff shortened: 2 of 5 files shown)


michaelfig (approved):
Great stuff!  Thanks for shepherding this through.

mhofman:
Sorry I haven't had time to look at this yet. I'd like to review before this lands.

mhofman (changes requested):
Running `yarn lint` locally took 33 minutes and failed with an OOM in the `solo` package.

Let's find a way to reduce the impact of this change.

turadg:
> Running yarn lint locally took 33 minutes and failed with an OOM in the solo package.

Recording numbers from out chat: (in packages/zoe)
mhofman's VM: 99s before, 1221s after (~12x slowdown)
turadg's M1: 5s before, 42s after (~8x slowdown)

> Let's find a way to reduce the impact of this change.

👍  [e1625fe](https://github.com/Agoric/agoric-sdk/pull/5495/commits/e1625fec12081dbe9cc9708e50c975f9784ba26d) makes it opt in with an `AGORICSDK_ESLINT_ENABLE_TYPE_RULES` flag. CI will use this and those who want to use it locally can enable it too.

.eslintrc.cjs
```diff
@@ -1,18 +1,23 @@
 /* eslint-env node */
+const process = require('process');
+const lintTypes = !!process.env.AGORICSDK_ESLINT_ENABLE_TYPE_RULES;
```
mhofman: I'd prefer an explicit test for `=== '1' || === 'true'` as the above would be truthy with `AGORICSDK_ESLINT_ENABLE_TYPE_RULES = '0'`

mhofman (changes requested):
Thanks for making this optional. One change request for the env check.

Should we also add a new script to the root `package.json` that runs `lint:packages` with the `AGORICSDK_ESLINT_ENABLE_TYPE_RULES="true"` and `NODE_OPTIONS="--max-old-space-size=8192"` set (which apparently is needed on my machine, not just in CI). We can bike-shed the name, something like `lint:with-types` ?

.eslintrc.cjs
turadg: I hear you would prefer that and you put a block on merge because of it. Imo "I'd prefer" is not a reason to block a change.

I could add an explicit test but I don't think it's worth it given how this will be used. It's reasonable for an env flag to be based on presence. Knowing the people who are going to use this flag, I don't expect anyone will try `AGORICSDK_ESLINT_ENABLE_TYPE_RULES=0`.

I'm not claiming it won't be better if it did handle that, just that it's not worth the added code or time to write it or blocking a PR.

turadg:
> Should we also add a new script to the root package.json that runs lint:packages

All the `lint:*` packages get run by `yarn lint` so that would duplicate. We could have a name in the root doesn't start with `lint:` but in package.json we'd lose the ability to document why the env vars are necessary. We could have a separate script, but I find the existing solution sufficient. It works for CI. On my machine I'll set the env var. On yours you know what to do. On others they can follow either example.

If this were production code the standards would be higher but it's just a little option for stronger linting. I expect our linting to get refactored more soon so don't want to create more commands that have to keep working as expected. This env var has the air of something that could go away.

.eslintrc.cjs
michaelfig: I would prefer to use the presence of `$CI` to determine the linting mode.  Totally easy to set: `CI=t yarn lint`.

turadg: wfm. I considered CI but felt aversion because I wanted people to run it locally too, but pros:
1. it is a much more conventional flag
2. it often means "do more, slower stuff"
3. makes obvious when using it that it's the same settings as run in CI

mhofman: `CI` is really a conflation. I'm not really a fan of implicit behavior changes based on environment sniffing. As @turadg mentions, there is value in being able to explicitly run this test locally.

mhofman:
> All the `lint:*` packages get run by `yarn lint` so that would duplicate.

Right, I forgot about that.

> On yours you know what to do.

No I don't. Me 3 days ago learned that I also need to set `NODE_OPTIONS` or I fail with OOM. Me in a few days will have forgotten. Calling commands with explicit env is fairly unergonomic if these are not documented anywhere (I guess they're documented in the GitHub actions `yml` files, but that's kinda obscure and not where I'd look in the first place).

I suppose I would have less qualms if somehow the `.eslintrc.js` file would somehow automatically set the required `max-old-space-size` if the env to trigger type checks is set.

.eslintrc.cjs
michaelfig: I think this, combined with your other comment, is the best argument for a long-named environment variable (but please, `AGORIC_` prefix rather than `SMASHINGTHE_FIRST_TWO_WORDS_TOGETHER`), plus a script that runs with the appropriate NODE_OPTIONS and enabling environment variable.

Then CI can use that script directly as well.  No need to put it in `package.json`, I think, unless it's something like `lint-slowly` :)

mhofman (approved):

Issue #5494, opened 2022-06-01
re-examine IST contracts for user-supplied parameter validation


Chris-Hibbert:
## What is the Problem Being Solved?

#5492, #5293 were recent cases in which contracts were not sufficiently validating their user-supplied inputs. #5131 is a case where the contract wasn't checking prerequisites.

## Description of the Design

Examine IST contracts and verify that user-supplied inputs, including brands in proposals are validated before being used.

## Security Considerations

General contract hygiene.

## Test Plan

Include tests for each fix.

dckc:
Another security consideration: **availability** of the Inter protocol contracts is critical to the Agoric economy.

Our fail-stop norms and tools do a good job on _integrity_, and while a stuck AMM is better than one that leaks assets, it's no good when a vault needs to be liquidated.

Chris-Hibbert:
With the closing of #5496, I'm done looking for more issues here.

PR #5496, opened 2022-06-01
fix: addPool: move validation before first side effects

Chris-Hibbert:
refs: #5494

## Description

In the AMM's addPool, move validation before first side effects
Before the fix would have called addPool before checking central Brand

### Security Considerations

contract robustness

No changes to documented behavior

### Testing Considerations

Added a test.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -147,10 +147,6 @@ export const makeAddPoolInvitation = (
       );
     }
 
-    const {
-      poolFacets: { pool, helper },
-    } = await addPool(secondaryBrand);
-
     assert(
       AmountMath.isGTE(centralAmount, minPoolLiquidity),
       `The minimum initial liquidity is ${minPoolLiquidity}, rejecting ${centralAmount}`,
@@ -164,6 +160,13 @@ export const makeAddPoolInvitation = (
     const [liquidityKeyword] = Object.entries(zcf.getTerms().issuers).find(
       ([_, i]) => i === issuer,
     );
+    assert(liquidityKeyword, 'liquidity brand required');
+
+    // COMMIT POINT /////////////////////
+
+    const {
+      poolFacets: { pool, helper },
+    } = await addPool(secondaryBrand);
 
     // in addLiquidityInternal, funder provides centralAmount & secondaryAmount,
     // and receives liquidity tokens equal to centralAmount. Afterward, we'll
--- a/packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
+++ b/packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
@@ -608,3 +608,62 @@ test('amm remove zero liquidity', async t => {
     `No liquidity removed`,
   );
 });
+
+test('amm bad add liquidity offer', async t => {
+  const centralLiquidityValue = 1_500_000_000n;
+  const secondaryLiquidityValue = 300_000_000n;
+
+  // Set up central token
+  const centralR = makeIssuerKit('central');
+  const moolaR = makeIssuerKit('moola');
+  const sousR = makeIssuerKit('sous');
+  const moola = value => AmountMath.make(moolaR.brand, value);
+  const sous = value => AmountMath.make(sousR.brand, value);
+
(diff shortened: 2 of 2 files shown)


packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
```diff
@@ -608,3 +608,62 @@ test('amm remove zero liquidity', async t => {
     `No liquidity removed`,
   );
 });
+
+test('amm bad add liquidity offer', async t => {
+  const centralLiquidityValue = 1_500_000_000n;
+  const secondaryLiquidityValue = 300_000_000n;
+
+  // Set up central token
```
turadg: we have this boilerplate in many tests. I'd like to make a helper like `setupTokens('central', 'moola', 'sous')` that gives them consistent names. WDYT of that returning,
```
{
 central, // (value) => Amount
 centralBrand,
 centralIssuer,
  moola,
  …,
  sousIssuer,
}
```
They could of course be renamed in object destructuring assignment but the defaults would support some consistency.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -164,6 +160,13 @@ export const makeAddPoolInvitation = (
     const [liquidityKeyword] = Object.entries(zcf.getTerms().issuers).find(
       ([_, i]) => i === issuer,
     );
+    assert(liquidityKeyword, 'liquidity brand required');
+
+    // COMMIT POINT /////////////////////
```
turadg, line 165: I like this convention. Aligns with https://github.com/Agoric/agoric-sdk/wiki/Prepare-Commit that @dckc started.

That says that the commit phase "must" be atomic. Should this all get wrapped in a try/catch? 

A convention that could help us to be more consistent and which would be easy to lint: define a `commit` function (could be within the handler so it doesn't need params) and in the end of the handler function,
```
try {
  await commit();
} catch {
  // abort so partial state is not observable
}
return result;
```
I don't know if that's the solution but if you think it would be feasible and worthwhile to have some more consistent structure I'd like to include it in https://github.com/Agoric/agoric-sdk/issues/4770

That's not for MN-1 but maybe an MN-1 thing would be to come up with some rules that we manually lint before release. Worth a ticket?

turadg (approved):
LGTM. Commented with two tangential topics.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
dckc: > Aligns with Prepare-Commit that @dckc started.

well, at the request of @erights 

> Should this all get wrapped in a try/catch?

I think we avoid that as a rule. I haven't looked at the rest of the code in this case. Is there some reason to believe that the stuff after the commit-point not atomic?

> ... define a commit function

I think @erights tried that and the resulting reduction in clarity was judged to be a net loss. Here's hoping I/we can find that experiment. Oh... wait... he tried a utility function that served as the try-catch idiom... I think what you're suggesting here is a `commit()` function for each case.

We wouldn't want to `await` the `commit()`. The `commit()` should be in the same turn as when the prerequisites are checked.

Chris-Hibbert: I'm not sure that we can make it atomic. My goal was to move all the (local) checks before starting to add the pool. There's an `await` immediately following the `COMMIT POINT`, and another on the call to `onOfferHandled`, which has 6 `await`s of its own. 

After this clean-up it will no longer be possible for `assert(AmountMath.isGTE(centralAmount, minPoolLiquidity))` to cause a problem after `addPool`.  So the pool will be created, and the `reallocate` should complete successfully.  But after that, the AMM can't completely protect itself from issues in the Reserve contract.

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
Chris-Hibbert: There are already a few helpers, and they're not as general as one might like. setupBasicMints in `run-protocol/test/swingsetTests/setup.js` only makes moola.  setupBasicMints.js in zoe unitTests is the right thing, but isn't exported cleanly.  

Would you object to a deep import of '@agoric/zoe/test/unitTests/setupBasicMints.js'?

PR #5497, opened 2022-06-02
jessie-check contracts

turadg:
## Description

A discussion on https://github.com/Agoric/agoric-sdk/issues/4770 suggested we should be using `@jessie-check` in all our contracts. This PR attempts that.

TODO
- [ ] solve `new Uint8Array`
- [ ] solve `do/while statements are not allowed in Jessie`  (why is this a rule?)
- [ ] solve `generators are not allowed in Jessie`. Would it suffice to pull [this](https://github.com/Agoric/agoric-sdk/blob/34807542f0888f08e1397fc343e6ed2717dd8c60/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js#L171-L182) out of the contract? there'd still be [processTranches](https://github.com/Agoric/agoric-sdk/blob/34807542f0888f08e1397fc343e6ed2717dd8c60/packages/run-protocol/src/vaultFactory/liquidateIncrementally.js#L190)
- [ ] solve `Unexpected `await` expression (not top of function body)` without changing semantics (several in liquidateIncrementally)
- [ ] reconsider the `curly` rule. Why is this in Jessie?

### DIFF
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -247,11 +247,11 @@ const start = async (zcf, privateArgs) => {
     const collateralKeyword = getKeywordForBrand(collateralAmount.brand);
     if (
       !AmountMath.isGTE(
-        collateralSeat.getCurrentAllocation()[collateralKeyword],
+        collateralSeat.getCurrentAllocation()[+collateralKeyword],
         collateralAmount,
       )
     ) {
-      throw new Error('insufficient reserves for that transaction');
+      throw assert.fail('insufficient reserves for that transaction');
     }
 
     // create the RUN
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -1,4 +1,5 @@
 // @ts-check
+// @jessie-check
 
 import { E, Far } from '@endo/far';
 import { makeStore } from '@agoric/store';
--- a/packages/zoe/src/contracts/autoswap.js
+++ b/packages/zoe/src/contracts/autoswap.js
@@ -4,6 +4,7 @@
 import { Far } from '@endo/marshal';
 import { assert } from '@agoric/assert';
 import { AmountMath, isNatValue } from '@agoric/ertp';
+import { makeMap } from 'jessie.js';
 
 // Eventually will be importable from '@agoric/zoe-contract-support'
 import {
@@ -71,7 +72,7 @@ const start = async zcf => {
   const { brands } = zcf.getTerms();
   Object.values(brands).forEach(brand => assertNatAssetKind(zcf, brand));
   /** @type {Map<Brand,Keyword>} */
-  const brandToKeyword = new Map(
+  const brandToKeyword = makeMap(
     Object.entries(brands).map(([keyword, brand]) => [brand, keyword]),
   );
   /**
--- a/packages/zoe/src/contracts/autoswap.js
+++ b/packages/zoe/src/contracts/autoswap.js
@@ -1,4 +1,5 @@
 // @ts-check
+// @jessie-check
 
 import { Far } from '@endo/marshal';
 import { assert } from '@agoric/assert';
--- a/packages/zoe/src/contracts/priceAggregatorChainlink.js
+++ b/packages/zoe/src/contracts/priceAggregatorChainlink.js
@@ -1,4 +1,5 @@
 // @ts-check
+// @jessie-check
 
 import { AmountMath, AssetKind, makeIssuerKit } from '@agoric/ertp';
 import { E } from '@endo/eventual-send';
--- a/packages/zoe/src/contracts/priceAggregatorChainlink.js
+++ b/packages/zoe/src/contracts/priceAggregatorChainlink.js
@@ -8,6 +8,7 @@ import { makeNotifierKit } from '@agoric/notifier';
 import { makeLegacyMap } from '@agoric/store';
 import { Nat, isNat } from '@agoric/nat';
 import { assert, details as X } from '@agoric/assert';
+import { makeSet } from 'jessie.js';
 import {
   calculateMedian,
   natSafeMath,
@@ -295,7 +296,9 @@ const start = async (
     }
 
     const roundTimedOut = timedOut(_roundId, blockTimestamp);
-    if (!roundTimedOut) return;
+    if (!roundTimedOut) {
+      return;
+    }
 
     const prevId = subtract(_roundId, 1);
 
@@ -348,10 +351,13 @@ const start = async (
    * @param {bigint} blockTimestamp
    */
   const oracleInitializeNewRound = (_roundId, _oracle, blockTimestamp) => {
-    if (!newRound(_roundId)) return;
-    const lastStarted = oracleStatuses.get(_oracle).lastStartedRound; // cache storage reads
-    if (_roundId <= add(lastStarted, restartDelay) && lastStarted !== 0n)
+    if (!newRound(_roundId)) {
       return;
+    }
+    const lastStarted = oracleStatuses.get(_oracle).lastStartedRound; // cache storage reads
+    if (_roundId <= add(lastStarted, restartDelay) && lastStarted !== 0n) {
+      return;
+    }
     initializeNewRound(_roundId, blockTimestamp);
 
     oracleStatuses.get(_oracle).lastStartedRound = _roundId;
@@ -414,8 +420,9 @@ const start = async (
     if (
       details.get(_roundId).submissions.length <
       details.get(_roundId).maxSubmissions
-    )
+    ) {
       return;
+    }
     details.delete(_roundId);
   };
 
@@ -466,20 +473,28 @@ const start = async (
       canSupersede = supersedable(subtract(_roundId, 1), blockTimestamp);
     }
 
-    if (startingRound === 0n) return 'not enabled oracle';
-    if (startingRound > _roundId) return 'not yet enabled oracle';
-    if (oracleStatuses.get(_oracle).endingRound < _roundId)
+    if (startingRound === 0n) {
+      return 'not enabled oracle';
+    }
+    if (startingRound > _roundId) {
+      return 'not yet enabled oracle';
+    }
+    if (oracleStatuses.get(_oracle).endingRound < _roundId) {
       return 'no longer allowed oracle';
-    if (oracleStatuses.get(_oracle).lastReportedRound >= _roundId)
+    }
+    if (oracleStatuses.get(_oracle).lastReportedRound >= _roundId) {
       return 'cannot report on previous rounds';
+    }
     if (
       _roundId !== rrId &&
       _roundId !== add(rrId, 1) &&
       !previousAndCurrentUnanswered(_roundId, rrId)
-    )
+    ) {
       return 'invalid round to report';
-    if (_roundId !== 1n && !canSupersede)
+    }
+    if (_roundId !== 1n && !canSupersede) {
       return 'previous round not supersedable';
+    }
     return '';
   };
 
@@ -611,7 +626,7 @@ const start = async (
       if (keyToRecords.has(oracleKey)) {
         records = keyToRecords.get(oracleKey);
       } else {
-        records = new Set();
+        records = makeSet();
         keyToRecords.init(oracleKey, records);
 
         const oracleStatus = makeOracleStatus(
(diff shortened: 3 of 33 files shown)


packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -246,11 +247,11 @@ const start = async (zcf, privateArgs) => {
     const collateralKeyword = getKeywordForBrand(collateralAmount.brand);
     if (
       !AmountMath.isGTE(
-        collateralSeat.getCurrentAllocation()[collateralKeyword],
+        collateralSeat.getCurrentAllocation()[+collateralKeyword],
```
dckc, line 250: `+` coerces to a number. I don't think this can work. Does it pass tests?

I suggest an `extractFieldFromKeywordRecord(record, key)` library function imported from a module with keyword record helpers. Perhaps with a more concise name. Or perhaps there's some `Object.getProperty` standard JS function that would work?

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -246,11 +247,11 @@ const start = async (zcf, privateArgs) => {
...
         collateralAmount,
       )
     ) {
-      throw new Error('insufficient reserves for that transaction');
+      throw assert.fail('insufficient reserves for that transaction');
```
dckc, line 254: `throw` is redundant with `assert.fail`, yes? I think we have an `assert.error` constructor.

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -347,10 +351,13 @@ const start = async (
...
   const oracleInitializeNewRound = (_roundId, _oracle, blockTimestamp) => {
-    if (!newRound(_roundId)) return;
+    if (!newRound(_roundId)) {
+      return;
+    }
     const lastStarted = oracleStatuses.get(_oracle).lastStartedRound; // cache storage reads
-    if (_roundId <= add(lastStarted, restartDelay) && lastStarted !== 0n)
+    if (_roundId <= add(lastStarted, restartDelay) && lastStarted !== 0n) {
       return;
+    }
```
dckc, line 360: This fixes a hazard that motivates the `curly` rule. (The other motivation is simplifying the Jessie grammar, I think.)

packages/zoe/src/contracts/sellItems.js
```diff
@@ -90,7 +91,7 @@ const start = zcf => {
     // Check that the wanted items are still for sale.
     if (!AmountMath.isGTE(currentItemsForSale, wantedItems)) {
       const rejectMsg = `Some of the wanted items were not available for sale`;
-      throw buyerSeat.fail(new Error(rejectMsg));
+      throw buyerSeat.fail(assert.fail(rejectMsg));
```
dckc, line 94: `assert.fail` throws. It's not a replacement for `new Error`.

I suggest just taking out `new`. `Error(...)` and `new Error(...)` mean the same thing.

packages/zoe/src/contracts/simpleExchange.js
```diff
@@ -127,7 +128,7 @@ const start = zcf => {
     }
     // Eject because the offer must be invalid
     throw seat.fail(
-      new Error(`The proposal did not match either a buy or sell order.`),
+      assert.fail(`The proposal did not match either a buy or sell order.`),
```
dckc, line 131: again, this is not semantics-preserving.

I suggest just changing `new Error(` to `Error(`

dckc (changes requested):
some of these changes are not semantics-preserving:

 - Adding + in `rec[+kwd]` coerces `kwd` to a `number`.
 - `new Error(...)` -> `assert.fail(...)` changes control flow.

packages/run-protocol/src/reserve/assetReserve.js
turadg: I was (intentionally) blindly following what the rule said to do,
```
  106:9   error  arbitrary computed property names are not allowed in Jessie; use leading '+'  no-restricted-syntax
```
what should it say instead?

splitting out into a helper function would still be running the computed property name, right?

packages/zoe/src/contracts/priceAggregatorChainlink.js
turadg: I would hope that 'jessie-check' checks what's known to fail and doesn't get into style decisions.

in the absence of curlies does Jessie fail to parse?

packages/run-protocol/src/reserve/assetReserve.js
dckc: The helper function would not be in a `@jessie-check` file. Just like `makeSet` and such.

> what should it say instead?

Maybe "if you're trying to do numeric indexing, prefix the index with a +". I agree that diagnostic is misleading.

packages/zoe/src/contracts/priceAggregatorChainlink.js
dckc: yes:
> In Jessie, only blocks are accepted as arms

https://github.com/agoric-labs/jessica/blob/e8ab6f70065360e201d5230824796b1ce6557cb7/lib/quasi-jessie.js.ts#L141

packages/run-protocol/src/reserve/assetReserve.js
turadg: > The helper function would not be in a @jessie-check file. Just like makeSet and such.

Sure. But won't it still run in a Jessie environment? I thought jessie-check was for ensuring that something can safely run in a Jessie environment. Below you show that [curlies can't run at all](https://github.com/Agoric/agoric-sdk/pull/5497#discussion_r888031197). Is the arbitrarily computed property rule the same? If so, won't the imported module have the same problem?

dckc: > But won't it still run in a Jessie environment?

No. Again, just like makeSet.

Jessie has 2 purposes:
 1. Reduce hazards in writing contracts
 2. Ease language implementation burden

In our system as a whole, we don't use Jessie in our contracts for the 2nd purpose. We support the whole (hardened) JS language at runtime.

Someone else might, for example, build a Jessie  interpreter targeting WASM for use on some other platform. They could use the @jessie-check modules without modification. But they would have to do something platform-specific for makeSet and getFieldFromKeywordRecord. For the latter, perhaps makeMap(Object.entries(rec)).get(kw). We don't want to do that because it's O(n). But it's worth putting it in a library module that's reviewed with different expectations.

erights: Yes, the primary purpose of Jessie, and the only purpose for which we are currently using it, is to help us write more reliable and analyzable code. To be fully pedantic, the `// @jessie-check` enforces only Tiny-SES rather than Jessie, where the difference is the primordials the code is allowed to assume exists. Aside from the primordials the code can assume exists, Jessie and Tiny-SES are identical.

In practice that was one too many distinctions so we use Jessie to describe both. In practice, when we say we're writing most of our code in Jessie, we do not generally mean to imply that the code cannot assume the presence of, for example `Reflect`, which is part of SES but not Jessie. 

We eventually intend to support Jessie as a std for universal mobile code, where a Jessie interpreter can be implemented simply in any normal imperative language. For that purpose, this difference will matter, as that implementation can only be simple if it omits many of the SES primordials including `Reflect`.

erights: The purpose of the square bracket indexing rules is to make code that is using computed property lookup to index into arrays not be able to be tricked into looking up, for example, inherited method properties. In general, we want code that is using `[]` safely to index into a collection of some sort safely to be analyzably safe. Zoe introduced the notion of keywords for exactly the same reason, noting that at least all the primordial method names begin with a lowercase character, and so cannot be mistaken for keywords.

If we were willing to have Jessie understand that keywords are safe computed indexes, we could similarly enable syntax that enforced that the index expression evaluated to a keyword. We have not bothered to do that, and have instead just tried to use `[]` safely by feeding it only keywords, which a static or dynamic check ensuring this.

So we have choices wrt these uses of `record[keyword]` in what would otherwise be Jessie code:
   * Have Jessie drop the `[]` restrictions. Yes, we should consider this.
   * Rewrite such code to `Reflect.get(record, keyword)` which means the same thing *given* the availability of `Reflect` which is technically part of SES and Tiny-SES but not Jessie.
   * Introduce a keyword enforcing helper function, upgrade Jessie to allow it, and use it for keyword lookup in the same way we use `+` for array lookup.
   * Do we have a `// @jessie-expect-error` directive analogous to `// @ts-expect-error`? If so, you could use it to allow expressions you know to be safe -- perhaps even adding to the comment directive an explanation of why you believe it to be safe.

The path of least resistance for now is probably rewriting to `Reflect.get(record, keyword)`. That's what I'd do, but feel free to consider the other alternatives.

packages/run-protocol/src/reserve/assetReserve.js
erights: Yes, the `throw assert.fail` should normally be just `assert.fail`. Because `assert.fail` @returns `{never}` the two should also be equivalent to typescript.

I have encountered rare circumstances where the `throw` was still necessary to inform some static check (probably es-lint) that control flow would not proceed normally past that point. Hope that you don't run into that.

packages/zoe/src/contracts/autoswap.js
```diff
@@ -1,8 +1,10 @@
 // @ts-check
+// @jessie-check
 
 import { Far } from '@endo/marshal';
 import { assert } from '@agoric/assert';
 import { AmountMath, isNatValue } from '@agoric/ertp';
+import { makeMap } from 'jessie.js';
```
erights, line 7: @michaelfig is that the right way to import from the Jessie std library? No `@something/something` ?

packages/zoe/src/contracts/priceAggregatorChainlink.js
erights: The motivation is writing reliable code. Before I religiously followed this style I would occasionally screw up by adding a statement to an indented block, not noticing that the indented block was not delimited by curlies.

There are also some strange language edge cases associated with curly-less control-flow arms. For example, what if it brings a variable into scope? What is the scope of the variable? What looks like the scope of the variable? Bad for these two answers to disagree.

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -465,20 +473,28 @@ const start = async (
...
-    if (startingRound > _roundId) return 'not yet enabled oracle';
-    if (oracleStatuses.get(_oracle).endingRound < _roundId)
+    if (startingRound === 0n) {
+      return 'not enabled oracle';
+    }
+    if (startingRound > _roundId) {
+      return 'not yet enabled oracle';
+    }
+    if (oracleStatuses.get(_oracle).endingRound < _roundId) {
       return 'no longer allowed oracle';
```
erights, line 483: The code before your change here is an example of the case that scares me: where the arm is on its own line and indented.

packages/zoe/src/contracts/sellItems.js
erights: Better would be `assert.error`. Like `Error(...)` it makes and returns an error. Like the other `assert` methods, you can use an
```js
X`literal stuff ${redacted value}`
```
for better diagnostics. Also, all the `assert` methods create the error object at the same line of code that becomes a useful place to put a breakpoint.

packages/zoe/src/contracts/simpleExchange.js
erights: Or `assert.error`

erights (approved):
LGTM AWESOME! I've been looking forward to this day for years. I am astonished how close so much of our code already was to Jessie prior to any automated checking!

Thanks for doing this!!

packages/zoe/src/contracts/autoswap.js
michaelfig: Yes, absolutely!  `'jessie.js'` is the stdlib package.

erights:
What is the status of this? Is there at least an easy subset that could move forward quickly?

turadg:
> What is the status of this?

It's a draft PR that passes 5 out of 30 CI checks. It doesn't have a ticket and if it did I don't think it would be required for MN-1, so it's prioritized behind MN-1 release.

> Is there at least an easy subset that could move forward quickly?

Probably the `jessie.js` update. Any files that had `@jessie-check` added and aren't erroring.

Again, TMK it's not required for MN-1 but if it is let's prioritize it.

erights:
It's not

Issue #5500, opened 2022-06-03
solve import.meta.url usage in test:xs-unit tests


turadg:
## What is the Problem Being Solved?

While [ava-xs runner was broken](#5498) some unit tests in zoe and ertp began using `import.meta.url` which fails under `test:xs-unit`. [27901ce](https://github.com/Agoric/agoric-sdk/pull/5501/commits/27901cebd94a5b985e64c7c921063f7645ad755b) shows the ones that fail (and are thus disabled for now).

## Description of the Design

In XS tests, support `import.meta.url` or some other mechanism to resolve the bundle path running from any CWD.

## Security Considerations
--
## Test Plan

--

kriskowal:
Support for `import.meta` in SES shim may be relevant https://github.com/endojs/endo/issues/291

Tartuffo:
A (simple) solution is to go through the tests and remove reliance on import.meta.url.

PR #5501, opened 2022-06-03
restore Ava-XS runner and disable tests that broke while it was off

### LINKED ISSUE #5498: xs tests passing blindly

The `test:xs-unit` command always passes, even when tests fail. For example in [yarn test (zoe)](https://github.com/Agoric/agoric-sdk/runs/6676047549?check_suite_focus=true):
```
$ ava-xs --verbose
# test script: test/unitTests/zoe/test-instanceAdminStorage.js
{
  status: 'not ok',
  message: 'running test script: Already locked down at TypeError: Prior lockdown (SES_ALREADY_LOCKED_DOWN) (SES_ALREADY_LOCKED_DOWN)',
  filename: 'test/unitTests/zoe/test-instanceAdminStorage.js',
  label: ''
}
# test script: test/unitTests/zoe/test-makeInvitation.js
{
  status: 'not ok',
  message: 'running test script: Already locked down at TypeError: Prior lockdown (SES_ALREADY_LOCKED_DOWN) (SES_ALREADY_LOCKED_DOWN)',
  filename: 'test/unitTests/zoe/test-makeInvitation.js',
  label: ''
}
0 tests passed
Done in 113.98s.
```

Notice that `ava-xs` is not detecting any tests running, so even though 0 passed the suite still passes.

## Description of the Design

First, make `ava-xs` count the failure of a test module to load as a failure. https://github.com/Agoric/agoric-sdk/pull/5398/commits/a29034acc5fc51250db285ece2ba1b1908f3dd36 might accomplish that.

Second, solve the [SES_ALREADY_LOCKED_DOWN](https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_ALREADY_LOCKED_DOWN.md) error. Might be due library importing @endo/init or the swingset tools thing. Only the top-level test-*.js files should be doing those only-once imports.

~Third, fix any errors that have crept in while this test runner wasn't working.~ Too much to fix to accomplish this. Should have another ticket.

## Security Considerations
--
## Test Plan

♲

turadg:
closes: #5498

## Description

avaXS had stopped working because of a SES_ALREADY_LOCKED_DOWN issue that got introduced at some point. We didn't know it because the runner had a bug whereby a module failing to load would not be treated as a failure.

1. solve the runner bug with better error handling
2. solve the lockdown issue by handling new XS interop cases that have developed in the test code
3. skip testing on the remaining cases that haven't been solved (deferred to https://github.com/Agoric/agoric-sdk/issues/5500 ) 

### Security Considerations

none.

Indicate why the tests are disabled. Done.

### Testing Considerations

Disable tests that don't run.

Reviewers, do we also need a test of avaXS itself to check that this error would have been caught? It would definitely help but it would delay other work and the tests it runs are also already run in regular Ava.


packages/zoe/package.json
```diff
@@ -73,6 +73,40 @@
   ],
   "ava-xs": {
     "exclude": [
+      "# path/url/import.meta stuff",
```
dckc, line 76: cite #5500 in this comment? add TODO marker?

dckc (commented):
I approve inasmuch as this matches the approach we agreed on.

I think it's cost-effective to get a more objective review, so for now at least, I defer to @kriskowal to approve in the github sense.

packages/ERTP/package.json
```diff
@@ -57,7 +57,13 @@
     "exported.js"
   ],
   "ava-xs": {
-    "exclude": "swingsetTests"
+    "exclude": [
+      "swingsetTests",
+      "# fast-check unsupported",
```
turadg, line 62: do we need to test these in XS?

If not, maybe we have a different directory for tests that run under both Ava and XS?

dckc: It's not clear to me why fast-check is not supported. What am I missing? What were the symptoms when we tried?

turadg: They don't _all_ fail but it looked like something with fast-check,
```
. test/unitTests/test-amountProperties.js store
. test/unitTests/test-amountProperties.js reject promise keys
. test/unitTests/test-amountProperties.js passability of stores
. test/unitTests/test-amountProperties.js passability of store iters
. test/unitTests/test-amountProperties.js iteration fails with concurrent addition
. test/unitTests/test-amountProperties.js iteration succeeds with concurrent deletion
. test/unitTests/test-amountProperties.js provide for mapStores
FAIL (todo route console) (TypeError#1)
TypeError#1: ?: no function

{
  status: 'not ok',
  id: 160,
  message: 'isEqual is a (total) equivalence relation threw: ?: no function',
  filename: 'test/unitTests/test-amountProperties.js',
  label: ''
}
F test/unitTests/test-amountProperties.js isEqual is a (total) equivalence relation
FAIL (todo route console) (TypeError#2)
TypeError#2: ?: no function

{
  status: 'not ok',
  id: 161,
  message: 'isGTE is a partial order with empty as minimum threw: ?: no function',
  filename: 'test/unitTests/test-amountProperties.js',
  label: ''
}
F test/unitTests/test-amountProperties.js isGTE is a partial order with empty as minimum
FAIL (todo route console) (TypeError#3)
TypeError#3: ?: no function

{
  status: 'not ok',
  id: 162,
  message: 'add: closed, commutative, associative, monotonic, with empty identity threw: ?: no function',
  filename: 'test/unitTests/test-amountProperties.js',
  label: ''
}
F test/unitTests/test-amountProperties.js add: closed, commutative, associative, monotonic, with empty identity
FAIL (todo route console) (TypeError#4)
TypeError#4: ?: no function

{
  status: 'not ok',
  id: 163,
  message: 'subtract: (x + y) - y = x; (y - x) + x = y if y >= x threw: ?: no function',
  filename: 'test/unitTests/test-amountProperties.js',
  label: ''
}
F test/unitTests/test-amountProperties.js subtract: (x + y) - y = x; (y - x) + x = y if y >= x
FAIL (todo route console) (TypeError#5)
TypeError#5: ?: no function

{
  status: 'not ok',
  id: 164,
  message: 'minmax threw: ?: no function',
  filename: 'test/unitTests/test-amountProperties.js',
  label: ''
}
```

dckc: ok... fast-check is unsupported in that tests that don't use it don't seem to work and we're not investing in whatever it takes to get them to work yet. fair enough.

for future reference, now that `getStackString()` is working, that's where I would pick up on "todo route console".

kriskowal (approved):
Very much in favor of gradually improving ava-xs as needed. We don’t have a ground to expect equivalence even with a lot of effort.

We will Real Soon Now™ be able to subsume some of the idiosyncrasies into fully realized Endo support for CommonJS and `import.meta`. The result will share a lot of the same structure, but we might be able to increase parity a little.

PR #5502, opened 2022-06-03
handle floating promises in run-protocol

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/5495 revealed where promises weren't being handled.

This resolves all the warnings within run-protocol. Classes of changes separated by commit.

### Security Considerations

--

--

### Testing Considerations

Note that some tests changed due only to more awaits within tests: a6275e7f4c01d9be6a929ad609e56646052490b1 (no changes required for changes to promise handling within application code)

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -264,7 +264,7 @@ const helperBehavior = {
 
     facets.helper.assetNotify();
     trace('chargeAllVaults complete');
-    facets.helper.reschedulePriceCheck();
+    return facets.helper.reschedulePriceCheck();
   },
 
   /** @param {MethodContext} context */
@@ -454,9 +454,14 @@ const helperBehavior = {
         facets.helper.updateMetrics();
 
         if (!AmountMath.isEmpty(accounting.shortfall)) {
-          E(factoryPowers.getShortfallReporter()).increaseLiquidationShortfall(
-            accounting.shortfall,
-          );
+          E(factoryPowers.getShortfallReporter())
+            .increaseLiquidationShortfall(accounting.shortfall)
+            .catch(reason =>
+              console.error(
+                'liquidateAndRemove failed to increaseLiquidationShortfall',
+                reason,
+              ),
+            );
         }
       })
       .catch(e => {
@@ -693,7 +698,7 @@ const finish = ({ state, facets: { helper } }) => {
   // push initial state of metrics
   helper.updateMetrics();
 
-  observeNotifier(state.periodNotifier, {
+  void observeNotifier(state.periodNotifier, {
     updateState: updateTime =>
       helper
         .chargeAllVaults(updateTime, state.poolIncrementSeat)
--- a/packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js
+++ b/packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js
@@ -391,7 +391,7 @@ test('amm doubleSwap', async t => {
 
   const metricsSub = await E(amm.ammPublicFacet).getMetrics();
   const m = await subscriptionTracker(t, metricsSub);
-  m.assertInitial({ XYK: [] });
+  await m.assertInitial({ XYK: [] });
 
   const aliceAddLiquidityInvitation = E(
     amm.ammPublicFacet,
--- a/packages/run-protocol/test/vaultFactory/test-liquidator.js
+++ b/packages/run-protocol/test/vaultFactory/test-liquidator.js
@@ -29,7 +29,7 @@ import {
   setUpZoeForTest,
   setupBootstrap,
   installGovernance,
-  waitForPromisesToSettle,
+  eventLoopIteration,
   withAmountUtils,
 } from '../supports.js';
 import { unsafeMakeBundleCache } from '../bundleTool.js';
@@ -669,14 +669,14 @@ test('update liquidator', async t => {
       AMMMaxSlippage: debt.makeRatio(30n),
     }),
   );
-  await waitForPromisesToSettle();
+  await eventLoopIteration();
   govNotify = await d.managerNotified();
   const newLiquidator = govNotify.value.liquidatorInstance;
   t.not(oldLiquidator, newLiquidator);
 
   // trigger liquidation
   await d.setPrice(debt.make(300n));
-  await waitForPromisesToSettle();
+  await eventLoopIteration();
   await dv.notified(Phase.LIQUIDATED);
 });
 
@@ -707,7 +707,7 @@ test('liquidate many', async t => {
   const dv9 = await d.makeVaultDriver(collateral, run.make(300n));
 
   await d.setPrice(await overThreshold(dv1));
-  await waitForPromisesToSettle();
+  await eventLoopIteration();
   await dv0.notified(Phase.LIQUIDATED);
   await dv1.notified(Phase.ACTIVE);
   await dv2.notified(Phase.ACTIVE);
@@ -720,7 +720,7 @@ test('liquidate many', async t => {
   await dv9.notified(Phase.ACTIVE);
 
   await d.setPrice(await overThreshold(dv5));
-  await waitForPromisesToSettle();
+  await eventLoopIteration();
   await dv1.notified(Phase.LIQUIDATED);
   await dv2.notified(Phase.LIQUIDATED);
   await dv3.notified(Phase.LIQUIDATED);
@@ -732,7 +732,7 @@ test('liquidate many', async t => {
   await dv9.notified(Phase.ACTIVE);
 
   await d.setPrice(run.make(300n));
-  await waitForPromisesToSettle();
+  await eventLoopIteration();
   await dv5.notified(Phase.LIQUIDATED);
   await dv6.notified(Phase.LIQUIDATED);
   await dv7.notified(Phase.LIQUIDATED);
@@ -800,7 +800,7 @@ test('penalties to reserve', async t => {
 
   // liquidate
   d.setPrice(run.make(636n));
-  await waitForPromisesToSettle();
+  await eventLoopIteration();
 
   await d.checkReserveAllocation(1000n, 29n);
 });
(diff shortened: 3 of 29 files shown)


packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -452,9 +452,14 @@ const helperBehavior = {
...
 
         if (!AmountMath.isEmpty(accounting.shortfall)) {
-          E(factoryPowers.getShortfallReporter()).increaseLiquidationShortfall(
-            accounting.shortfall,
-          );
+          E(factoryPowers.getShortfallReporter())
+            .increaseLiquidationShortfall(accounting.shortfall)
+            .catch(reason =>
+              console.error(
+                'liquidateAndRemove fialed to increaseLiquidationShortfall',
```
Chris-Hibbert: typo

packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js
```diff
@@ -391,7 +391,7 @@ test('amm doubleSwap', async t => {
 
   const metricsSub = await E(amm.ammPublicFacet).getMetrics();
   const m = await subscriptionTracker(t, metricsSub);
-  m.assertInitial({ XYK: [] });
+  await m.assertInitial({ XYK: [] });
```
Chris-Hibbert, line 394: `void` here would say "don't wait, and ignore it if a broken promise isn't handled". `await` here says "don't proceed until we have the results from the call". 

I think not having a marker means that it doesn't wait, and the test will fail if a broken promise isn't handled. Is that closer to what we want in a test? Or am I misunderstanding how the alternatives work?

I think the empty case does a worse job of telling us where the issue was, but are the others better at that?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -440,9 +440,10 @@ const makeDriver = async (t, initialPrice, priceBase) => {
     currentSeat: () => currentSeat,
     lastOfferResult: () => currentOfferResult,
     timer: () => timer,
-    tick: (ticks = 1) => {
+    tick: async (ticks = 1) => {
       for (let i = 0; i < ticks; i += 1) {
-        timer.tick();
+        // eslint-disable-next-line no-await-in-loop
+        await timer.tick();
```
Chris-Hibbert: This seems like a case where `tick()` doesn't need to be `async`, as long as all the callers `await` the results.  Alternatively, this could await the last or a `Promise.all()` for the group of them.

packages/run-protocol/test/vaultFactory/test-vault-interest.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-await-in-loop */
```
Chris-Hibbert: Let's not do this unless we're adding an issue to fix it.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-await-in-loop */
```
Chris-Hibbert: Let's find a different solution to multiple ticks than calling `await` on every  one.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1829,14 +1830,14 @@ test('mutable liquidity triggers and interest', async t => {
...
-  manualTimer.tick();
+  await manualTimer.tick();
+  await manualTimer.tick();
+  await manualTimer.tick();
+  await manualTimer.tick();
   await manualTimer.tick();
   await waitForPromisesToSettle();
 
-  shortfallBalance += 42n;
+  shortfallBalance += 46n;
```
Chris-Hibbert: I want to understand what changed here. Changing the number of ticks shouldn't change what the shortfall balance is on the next state change.

packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js
turadg: > I think not having a marker means that it doesn't wait, and the test will fail if a broken promise isn't handled. Is that closer to what we want in a test? Or am I misunderstanding how the alternatives work?

I think there's some misunderstanding. `void` does not change the handling of the promise; it serves to tell the linter that the non-handling is intentional. 

This code works exactly the same without `void`,
```
test.only('unhandled', async t => {
  const makeRejector = async () => {
    assert.fail('problem');
  };
  void makeRejector();
  t.pass();
});
```
With this Ava output,
```
  Unhandled rejection in test/amm/vpool-xyk-amm/test-xyk-amm-swap.js

  Error: problem
    at makeRejector (packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js:1174:12)
    at packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js:1176:3
    at async Promise.all (index 0)

  › makeRejector (packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js:1174:12)
  › packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js:1176:3

  ─

  1 test passed [16:12:26]
  1 unhandled rejection
```

I contend this code (as above) is better,
```
test.only('unhandled', async t => {
  const makeRejector = async () => {
    assert.fail('problem');
  };
  await makeRejector();
  t.pass();
});
```
because it handles the rejection and pretty-prints the error:
```
  Rejected promise returned by test. Reason:

  Error {
    message: 'problem',
  }

  › makeRejector (packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js:1174:12)
  › packages/run-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js:1176:9

  ─

  1 test failed [16:13:22]
```

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: > This seems like a case where tick() doesn't need to be async, as long as all the callers await the results.

The callers can't await the result of `tick:` (line 443) unless it's async (as here) or returns a Promise. So yeah, this could instead be:
```
tick: (ticks = 1) => Promise.all(Array(ticks).map(() => timer.tick());
```
Do you think that'd be better?

packages/run-protocol/test/vaultFactory/test-vault-interest.js
turadg: I've been thinking the other direction, to disable the rule in tests. The rule was created because,
> performing an await as part of each operation is an indication that the program is not taking full advantage of the parallelization benefits of async/await

but that's less pertinent to test code. In test code clarity of what's happening is more important than small perf wins. Also we often use loops like this to do data-driven coding that isn't necessarily parallelizable. E.g. in the await on 115.

So I don't think there's anything to _fix_ about allowing an await in that loop. If pressed we could refactor like so,
```
  async function checkInterest(i, charge) {
    await testJig.advanceRecordingPeriod();
    interest += charge;
    t.is(
      vault.getCurrentDebt().value,
      startingDebt + interest,
      `interest charge ${i} should have been ${charge}`,
    );
    t.is(vault.getNormalizedDebt().value, startingDebt);
  };
  await checkInterest(1, 3n);
  await checkInterest(2, 4n);
  await checkInterest(3, 4n);
  await checkInterest(4, 4n);
```
but makes room for errors in the iteration number.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Ok, would `tickN` on ManualTimer be sufficient?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: I'd like to as well but am balancing that against other priorities. Are you requesting that I figure out what happened and report here before this can be approved?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: Yes.

packages/run-protocol/test/vaultFactory/test-liquidator.js
Chris-Hibbert: yes.

packages/run-protocol/test/vaultFactory/test-vault-interest.js
Chris-Hibbert: I'm not the one that turned on `no-await-in-loop`, so I don't feel like I can approve that choice. An issue to discuss whether the rule is superfluous in tests would satisfy me.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: I do consider that a blocking issue. I can probably make time to investigate in the next few days if you want to assign the PR to me.

packages/run-protocol/test/vaultFactory/test-vault-interest.js
turadg: Good idea! https://github.com/Agoric/agoric-sdk/issues/5527

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Ok, I'll assign to you and pull it back if I find time.

turadg: To get these improvements in sooner I made the tests pass without changing numbers and left comments on what changes to tick awaiting result in number changing with `FIXME test result relies on awaiting each tick`: [aa16d77](https://github.com/Agoric/agoric-sdk/pull/5502/commits/aa16d77a25d5923bd91ee16b09830b57df046899)

Chris-Hibbert (approved):
Thanks for the changes. This looks good.

mhofman:
I'm not sure why mergify landed this PR before integration tests ran successfully, but this PR is somehow ~both~ failing getting started integrations tests

PR #5504, opened 2022-06-03
fix(run-protocol): fix add liquidity check

### LINKED ISSUE #5503: Add liquidity fails
## Describe the bug
When trying to add liquidity on sim-chain `agoric start --reset` on a pool that does in fact exist, the offer fails with:

```
2022-06-03T05:08:08.268Z fake-chain: delivering to sim-chain (trips=110)
2022-06-03T05:08:08.858Z SwingSet: ls: v20: Logging sent error stack (TypeError#4)
2022-06-03T05:08:08.859Z SwingSet: ls: v20: TypeError#4: addLiquidity: no function
2022-06-03T05:08:08.859Z SwingSet: ls: v20: TypeError: addLiquidity: no function
 at addLiquidity (.../run-protocol/src/vpool-xyk-amm/addLiquidity.js:30)
 at apply ()
```

samsiegart:
fixes: https://github.com/Agoric/agoric-sdk/issues/5503


turadg:
Note this would have been caught by https://github.com/Agoric/agoric-sdk/pull/5495 :)

Chris-Hibbert (approved):

packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
```diff
@@ -25,9 +26,9 @@ const makeMakeAddLiquidityInvitation = (zcf, getPool) => {
 
     // Get the brand of the secondary token so we can identify the liquidity pool.
     const secondaryBrand = seat.getProposal().give.Secondary.brand;
-
     const pool = getPool(secondaryBrand);
-    const liquidityBrand = pool.getLiquidityIssuer().getBrand();
+    const liquidityIssuer = E(pool).getLiquidityIssuer();
+    const liquidityBrand = await E(liquidityIssuer).getBrand();
```
dtribble: This is not a good reason for making the offer hook async and have an `await`.  There should be an API that provides the brand from the ZCF, which should be synchronous.

dtribble (changes requested):

packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
Chris-Hibbert: If you have the issuer inside a contract, you can directly call `zcf.getBrandForIssuer(issuer)` to get the brand synchronously.

samsiegart: Just got that working, thanks. PTAL.

Chris-Hibbert (approved):

turadg (approved):

dtribble (approved):
LGTM

samsiegart:
Thanks for reviews!

PR #5509, opened 2022-06-04
fix: handle adjustBalances with two empty amounts

### LINKED ISSUE #4805: [Vault] If adjust balances is called with two empty amounts, the reallocate fails
## Describe the bug

During Restival, we were trying to prove that if you tried to `adustBalances` with only a change to one of `RUN` and `Collateral`, that it would fail due to not creating a staging. It turns out that both seats get stagings if either amount is positive. That led to a bug demonstration that if both are empty, the `reallocate` fails

## To Reproduce

Create a test case that tries to trade 0 RUN for 0 Collateral

## Expected behavior

It should cleanly trade nothing, or cleanly refusse.

## Platform Environment

irrelevant

Chris-Hibbert:
fixes: #4805

## Description

vault's `adjustBalances()` wasn't protecting itself from calls with two empty amounts. The issue arises because `stageDelta` is called twice, and it's expected that one of those calls won't make a change. With two empty amounts, neither call changed allocations, and so reallocation failed because no changes were staged. The fix is to return a successful empty trade in that case.

Once I was looking at `stageDelta`, I saw that runStake also calls `stageDelta` in the same way. I added some tests of empty requests to runStake, but none of them caused a problem there, since runStake creates attestations for each trade, and fails when empty attestations are requested.

### Security Considerations

Not a security issue

No change needed.

### Testing Considerations

Added tests.

### DIFF
--- a/packages/run-protocol/src/contractSupport.js
+++ b/packages/run-protocol/src/contractSupport.js
@@ -48,6 +48,12 @@ export const assertOnlyKeys = (proposal, keys) => {
  * the gain and a loss on the `fromSeat`. The gain/loss are typically from the
  * give/want respectively of a proposal. The `key` is the allocation keyword.
  *
+ * Warning! if the amounts are both empty, this won't add a staged allocation to
+ * either seat, so the seats must only be passed to reallocate() if something
+ * else does stage an allocation. It would be wrong to throw here, since the
+ * common pattern of use is to call this twice, with one of the two calls not
+ * staging anything. It's only a mistake if both do nothing.
+ *
  * @param {ZCFSeat} fromSeat
  * @param {ZCFSeat} toSeat
  * @param {Amount} fromLoses
--- a/packages/run-protocol/src/contractSupport.js
+++ b/packages/run-protocol/src/contractSupport.js
@@ -48,12 +48,6 @@ export const assertOnlyKeys = (proposal, keys) => {
  * the gain and a loss on the `fromSeat`. The gain/loss are typically from the
  * give/want respectively of a proposal. The `key` is the allocation keyword.
  *
- * Warning! if the amounts are both empty, this won't add a staged allocation to
- * either seat, so the seats must only be passed to reallocate() if something
- * else does stage an allocation. It would be wrong to throw here, since the
- * common pattern of use is to call this twice, with one of the two calls not
- * staging anything. It's only a mistake if both do nothing.
- *
  * @param {ZCFSeat} fromSeat
  * @param {ZCFSeat} toSeat
  * @param {Amount} fromLoses
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -462,6 +462,10 @@ const helperBehavior = {
       debt,
     );
     const wantRUN = proposal.want.RUN || helper.emptyDebt();
+    if ([giveColl, giveRUN, wantColl].every(a => AmountMath.isEmpty(a))) {
+      clientSeat.exit();
+      return 'no transaction, as requested';
+    }
 
     // Calculate the fee, the amount to mint and the resulting debt. We'll
     // verify that the target debt doesn't violate the collateralization ratio,
--- a/packages/run-protocol/test/runStake/test-runStake.js
+++ b/packages/run-protocol/test/runStake/test-runStake.js
@@ -799,6 +799,7 @@ test('Starting LoC', async (/** @type {RunStakeTestContext} */ t) => {
   await d.checkBLDLiened(8000n);
   await d.borrowMoreRUN(1400n);
   await d.checkRUNDebt(1632n);
+  await d.borrowMoreRUN(0n);
 });
 
 test('Extending LoC - CR increases (FAIL)', async (/** @type {RunStakeTestContext} */ t) => {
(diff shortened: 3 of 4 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -462,6 +462,10 @@ const helperBehavior = {
       debt,
     );
     const wantRUN = proposal.want.RUN || helper.emptyDebt();
+    if ([giveColl, giveRUN, wantColl].every(a => AmountMath.isEmpty(a))) {
```
turadg, line 465: is this check not needed in runstake too?
https://github.com/Agoric/agoric-sdk/blob/6b54107e38a77a63073d9d422b186ea5ab0ee4e1/packages/run-protocol/src/runStake/runStakeKit.js#L309-L310

packages/run-protocol/src/contractSupport.js
```diff
@@ -48,6 +48,12 @@ export const assertOnlyKeys = (proposal, keys) => {
  * the gain and a loss on the `fromSeat`. The gain/loss are typically from the
  * give/want respectively of a proposal. The `key` is the allocation keyword.
  *
+ * Warning! if the amounts are both empty, this won't add a staged allocation to
+ * either seat, so the seats must only be passed to reallocate() if something
+ * else does stage an allocation. It would be wrong to throw here, since the
+ * common pattern of use is to call this twice, with one of the two calls not
```
turadg: When I see a warning in a comment, I assume someone will miss it and I try to think how we could enforce it programmatically.

Searching for all uses, there are basically two instances of the same logic that happens in vault and runStake:
```
    stageDelta(clientSeat, vaultSeat, giveColl, wantColl, 'Collateral');
    // `wantRUN` is allocated in the reallocate and mint operation, and so not here
    stageDelta(clientSeat, vaultSeat, giveRUN, helper.emptyDebt(), 'RUN');
```
```
    stageDelta(clientSeat, vaultSeat, giveColl, wantColl, KW.Attestation);
    stageDelta(clientSeat, vaultSeat, giveRUN, emptyDebt, KW.Debt);
```

So I content that _all_ uses call two times and because the function is so small we don't need to support the single call case. In which case the function can check itself whether any allocation was staged.

A clean way to do that would be to pass an array of deltas, which would have the added benefit of working for the n=1 case and n>2 cases. I.e. the function signature would be instead,
```
 * @param {ZCFSeat} fromSeat
 * @param {ZCFSeat} toSeat
 * @param {Array<[losses: Amount, gains: Amount, key: Keyword>} fromDeltas
```

turadg (commented):
Suggested an alternate solution

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: I added tests to try to replicate the same situation in runStake, and wasn't able to create a problematic situation.  RunStake creates attestations for each trade, and fails when empty attestations are requested, so I don't think the same situation can arise.

packages/run-protocol/src/contractSupport.js
Chris-Hibbert: I went through the same analysis (summarized in the top comment of this PR).

I don't think that catching the problem in stageDelta or its revision improves the situation. The reallocation will fail after an incorrect call, and the transaction will be aborted. The problem I tried to address is that the error message is inscrutable. In vaultFactory, we can detect it before calling stageDelta, and give a more understandable response. After analysis there didn't seem to be a need to do the same in runStake, since it already has guards that prevent requests for attestations for empty amounts.

packages/run-protocol/src/vaultFactory/vault.js
turadg: Re-reading the PR description I see you had noted that. I regret missing it earlier.

packages/run-protocol/src/contractSupport.js
turadg: I see in the top comment the thing about `stageDelta` but not the rest. 

> The problem I tried to address is that the error message is inscrutable.

Wouldn't throwing when no allocations are staged be strutable?

> The fix is to return a successful empty trade in that case.

I don't yet understand why that's better than erroring.

Chris-Hibbert: > Wouldn't throwing when no allocations are staged be scrutable?

somewhat. The user has no concept of allocations or reallocations, or of what seats are involved internally so rephrasing in their terms seems more helpful. 

If this seems unnecessary, I'm alright with that decision. I feel like we've established that there aren't routes through this code that lead to problematic states.

turadg: Making the offer return something sensible to the user makes sense to me. Whether that's an error or a string, I don't have a strong opinion. I suppose the current approach is good which says "you asked me to do nothing, so that's what I did." 

My concern is only with the warning comment. If it's really something we need to make sure people don't do, then I think throwing an error _would be_ in order. The reason give for not doing it is that `stageDelta` will be called multiple times and only one needs to be legit. My point is that we can bundle those calls into one by taking an array. Then the function can give an informative error message if some code called a stage that didn't stage.

I ask that "warning" be removed, either by having the function give a helpful error message for the case described or by deciding that it's not such a dangerous case and adjusting the language.

Chris-Hibbert: I removed the warning.

If you call `adjustBalances` in `runStake` with empty changes, it transfers empty attestations, so the reallocate is happy.

We only need the check for zero in vaults.

turadg (approved):

PR #5510, opened 2022-06-04
feat: add API for configuring kernel & vat options on a running swingset

### LINKED ISSUE #5237: kernel parameter configuration API and chain-side controls

We have a couple of kernel parameters that may need to be tuned after launch, as we learn more about the way things perform. We expect to have more in the future.
* `reapInterval` aka BOYD aka BringOutYourDead timing: #4160 is about choosing the initial value
* virtual object cache size: #2644 
* vat-warehouse heap-snapshot interval
* message size limit: #4280 is about implementing one, but the size then needs to be configurable
* kernel run-queue size influencing cosmic-swingset message admission (maybe increasing minimum gas price, maybe blocking messages entirely)

For all of these knobs, we need a plan for how they'll be changed after launch.

Some of them are currently stored in the DB, separately, for each vat. Some are (or will be) kernel-wide. Most do not yet have APIs to change them. We could add new methods to `vatAdminService` to change them from within JS, in which case the "Big JS Hammer" (bootstrap.js) upgrade method would be the governance mechanism that do the changes. For the per-vat items, we could methods to each vat's `adminFacet` to manipulate the values, which might allow Zoe to make the changes.

Or, we could add `controller.*` methods to changes things. That would leave `cosmic-swingset` holding the controls, and would require some cosmos-side governance mechanism to control them.

Some of these parameters might be needed to pause (or unpause) Swingset entirely. Once Swingset is paused or broken somehow, we won't be able to get a message to JS, so the unpause button must be invoked by the cosmos side.

We need both kernel-provided control APIs, and code within contracts or cosmic-swingset that will use these APIs upon completion of some kind of governance action.

## Description of the Design

TBD

FUDCo:
Fixes #5237

Adds a `changeOptions` call the vat admin node API and a `changeKernelOptions` call to the controller API.


packages/SwingSet/src/controller/controller.js
```diff
@@ -403,6 +403,10 @@ export async function makeSwingsetController(
       return kernel.shutdown();
     },
 
+    changeKernelOptions(options) {
```
warner, line 414: Let's document this, specifically to enumerate the options that can be set this way and what their legal values are. How about a new major section at the bottom of `docs/configuration.md`? Please include a note that the call must be performed as part of a "block": so 1: it should only be called while the kernel is idle (not in the middle of a `controller.run()`, where the results would be unpredictable), and 2: the swingstore `commit()` must be called afterwards.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1505,13 +1563,35 @@ export default function buildKernel(
...
+      switch (option) {
+        case 'defaultReapInterval': {
+          kernelKeeper.setDefaultReapInterval(value);
+          break;
+        }
+        case 'heapSnapshotInterval': {
+          vatWarehouse.setSnapshotInterval(value);
+          break;
+        }
+        default:
```
warner, line 1580: Might almost be worth using this Error to signal problems instead of the `assertKnownOptions` call, since it's not like we're saving `options` anywhere.

Also I think we use `throw Error` instead of `throw new Error`, in general.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -121,6 +121,9 @@ export function buildRootObject(vatPowers) {
         pendingUpgrades.set(upgradeID, upgradeRR);
         return upgradeCompleteP;
       },
+      changeOptions(options) {
+        D(vatAdminNode).changeOptions(vatID, options);
```
warner, line 141: We should do as much validation here as possible, before invoking the device, so the error message will be meaningful. The set of valid options (and the types of their values) should be checked here.

Especially because this doesn't use a `changeID` to provide a hook for returning an error signalled deeper in the kernel, like `upgrade`/`upgradeID`/`pendingUpgrades`/`vatUpgradeCallback` does.

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
```diff
@@ -233,6 +233,18 @@ export function buildDevice(tools, endowments) {
...
 
+        // D(devices.vatAdmin).changeOptions(vatID, options)
+        if (method === 'changeOptions') {
+          const args = JSON.parse(argsCapdata.body);
+          assert(Array.isArray(args), 'changeOptions() args array');
+          assert.equal(args.length, 2, `changeOptions() args length`);
+          const [vatID, _reason] = args;
+          assert.typeof(vatID, 'string', `changeOptions() vatID`);
+
+          syscall.callKernelHook('changeOptions', argsCapdata);
```
warner, line 244: I see that `vat-admin-hooks.js` is building an `undefined` capdata and returning it, but this is also using `returnFromInvoke(undefined)`. We only need one of the two. It looks like the `terminateWithFailure` example was making the same mistake.

That said, I'm not immediately sure which one to use, possibly `return syscall.callKernelHook(..)` (and I seem to remember the "device hooks" API being required to return capdata all the time). I'll look at it more deeply to figure out what to recommend.

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -317,6 +317,16 @@ export default function makeKernelKeeper(
     return ri;
   }
 
+  function setDefaultReapInterval(interval) {
+    assert(
```
warner, line 343: should we `Nat` this thing to prohibit `-3` and `2.7` and stuff?

packages/SwingSet/src/kernel/state/vatKeeper.js
```diff
@@ -137,6 +137,13 @@ export function makeVatKeeper(
     kvStore.set(`${vatID}.reapCountdown`, `${count}`);
   }
 
+  function updateReapInterval(reapInterval) {
+    kvStore.set(`${vatID}.reapInterval`, `${reapInterval}`);
```
warner, line 140: does this want a type/Nat check?

packages/SwingSet/src/kernel/vat-warehouse.js
```diff
@@ -63,6 +63,8 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
     // Often a large contract evaluation is among the first few deliveries,
     // so let's do a snapshot after just a few deliveries.
     snapshotInitial = 2,
+  } = policyOptions || {};
+  let {
```
warner: This is making me think more carefully about how `warehousePolicy` is provided. Currently, `makeSwingsetController()` can provide a different `warehousePolicy` each time, as part of the third `runtimeOptions` argument. The "runtime" in the name implies that it's safe to change the value each time you launch the kernel.

That's true for `maxVatsOnline` (which might cause vats to be paged out at a different schedule, but won't change the timing of consensus-visible `writeSnapshot` operations). But both `snapshotInitial` and `snapshotInterval` *do* change that timing. (They're consensus-visible because writing a heap snapshot starts with a `forceGC`, and that changes the heap, which changes the snapshots, and we're still hoping to get consistency of snapshot contents).

So, really, it was never ok to include `snapshotInitial` or `snapshotInterval` in `runtimeOptions`. Instead, we should have accepted initial values in `config`, as passed to `initializeSwingset`, and stored them in the kernelKeeper. Then `makeVatWarehouse` should read them from `kernelKeeper` instead of taking them as arguments.

This overlaps with a likely bug in this code: if there's nothing to persist `snapshotInterval` in the DB, then a sequence of:
* launch kernel with `runtimeOptions` that omits `snapshotInterval`
  * `makeVatWarehouse()` then uses the default of 200
  * host app does `c.changeKernelOptions({snapshotInterval: 50})`
* a few blocks go by
* kernel is rebooted
  * host app is still not overriding `runtimeOptions`
  * `makeVatWarehouse` reverts to default 200

That makes the snapshot-creating schedule dependent upon when exactly the application gets rebooted, and that's a recipe for consensus failure.

So I think we should probably clean this up as part of this PR. Your addition of `setSnapshotInterval` is fine (although it needs to write the value into `kernelKeeper` too, to fix the problem above). But `makeVatWarehouse` should do two reads to get the values to hold in RAM, and `initializeSwingset.js`/`initializeKernel.js` should remove them from the acceptable `runtimeOptions` values, and `initializeKernel()` should copy them from `config` into `kernelKeeper` somewhere. These default numbers will live in `initializeKernel()`.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -634,6 +634,58 @@ export default function buildKernel(
...
+  /**
+   *
+   * @param { RunQueueEventChangeVatOptions } message
+   * @returns { Promise<DeliveryStatus | null> }
+   */
+  async function processChangeVatOptions(message) {
+    const { vatID, options } = message;
+    insistVatID(vatID);
+    // eslint-disable-next-line no-use-before-define
+    assert(vatWarehouse.lookup(vatID));
```
warner: oh, does this tolerate a vat being terminated before the `changeVatOptions` event is processed?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -634,6 +634,58 @@ export default function buildKernel(
     );
   }
 
+  function setKernelVatOption(vatID, option, value) {
+    switch (option) {
+      case 'reapInterval': {
+        if (
+          typeof value === 'number' ||
```
warner: maybe add a Nat?

sounds like a helper function to factor out the option checking between here and the early checks in `vat-vat-admin` might be warranted

packages/SwingSet/src/kernel/vat-admin-hooks.js
```diff
@@ -112,5 +112,19 @@ export function makeVatAdminHooks(tools) {
       terminateVat(vatID, true, reasonCD);
       return harden({ body: stringify(undefined), slots: [] });
     },
+
+    changeOptions(argsCapData) {
+      // marshal([vatID, options]) -> null
+      const args = JSON.parse(argsCapData.body);
```
warner, line 119: probably appropriate to assert that `slots.length === 0`, if someone provides a vref as an option value then we should catch that as early as possible

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1181,6 +1181,27 @@ function build(
     ...collectionManager.testHooks,
   });
 
+  function setVatOption(option, value) {
+    switch (option) {
+      case 'virtualObjectCacheSize': {
+        if (typeof value === 'number') {
+          vom.setCacheSize(value);
```
warner, line 1287: I'd definitely `Nat()` this

packages/SwingSet/test/change-parameters/bootstrap-change-parameters.js
```diff
@@ -0,0 +1,31 @@
...
+      const bcap = await E(vatAdmin).getNamedBundleCap('carol');
+      const res = await E(vatAdmin).createVat(bcap);
+      carolRoot = res.root;
+      carolAdmin = res.adminNode;
+      await E(carolRoot).doSomething();
+    },
+
+    change() {
+      E(carolAdmin).changeOptions({
+        foo: 47,
```
warner: with vat-vat-admin doing checking, this `foo` will trigger an error.. we should test both a successful change and a rejected one

packages/SwingSet/test/change-parameters/vat-carol.js
```diff
@@ -0,0 +1,9 @@
+import { Far } from '@endo/marshal';
+
+export function buildRootObject() {
+  return Far('root', {
+    doSomething() {
+      console.log('carol does something');
```
warner: maybe this could return the string, instead of adding bulk to our already-noisy test logs (in the successful case)

warner (changes requested):
some small, some medium changes to make

packages/SwingSet/src/controller/controller.js
FUDCo: Taking a look at it, I see that `configuration.md` is actually quite out of date (last updated in March 2021), so perhaps aligning it with current reality should be part of this.

packages/SwingSet/src/kernel/kernel.js
FUDCo: Actually, we use `assert.fail` ...

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: The tradeoff here is that baking a bunch of knowledge into the vat admin vat about what options and option values are valid means we have to encode this knowledge redundantly in two different places, which at the very least will be a recurring opportunity to mess things up under maintenance.  Possibly we _could_ write a general purpose change-option-validator that could be called from both places, but this is complicated by the fact that (a) what to do in the event of failure is different between the two, and (b) on the receiving end (i.e., where the options actually get changed) different options get further dispatched into two different handlers depending on whether they are the kernel's business or liveslots' business, and each of _those_ handlers has different things that it needs to check.  I'm not opposed in principle to pre-flighting as much as possible in a place where it's possible to deliver error feedback to the requestor, but this is not straightforward.

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
FUDCo: This would be explained by the fact that I was using `terminateWithFailure` as my template for how to do this.

packages/SwingSet/src/kernel/state/kernelKeeper.js
FUDCo: I was thinking this might be tricky, but I just looked at the `nat` package and discovered it also has an `isNat` predicate, which will be the thing to use here since we also need to allow `'never'`.

FUDCo: The test is now `isNat(interval) || interval === 'never'` which is much cleaner.  There are several other points in the code where the more complicated type check thing was being done that could also be productively changed to use this pattern.

packages/SwingSet/src/kernel/vat-warehouse.js
FUDCo: Yeah, as I was updating `configuration.md` I noticed that we had decided to make `heapSnapshotInterval` updatable but it isn't one of the kernel configuration properties you can set initially, and that definitely seemed like something was off.  I noticed that the vat warehouse keeps the interval value in RAM but I unthinkingly just thought it was fetched by some other code somewhere higher on the call stack, but I didn't immediately sniff that that was self evidently a wrong thought.

I agree with your analysis and will make the appropriate updates as part of this PR.

packages/SwingSet/src/kernel/kernel.js
FUDCo: No.  I guess these non-send deliveries don't benefit from the automatic message splatting that regular message deliveries get.

I cribbed a lot from `processUpgradeVat` and I notice that it has the same flaw.  I will fix them both.

Also: `processStartVat` kinda sorta has the same issue, but I don't think a vat can be terminated before it is started -- though as I wrote that I started to wonder...

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: Yup.  As I've been cleaning these up I've decided I like `isNat()` much more than `Nat()`

packages/SwingSet/test/change-parameters/vat-carol.js
FUDCo: Actually, the `console.log` is just there to give the method something to do as if it was a real method, and the method only exists to be sure the Carol vat is actually alive.  It was useful to see the message as I was getting the test working, but I can take the log message out now that the test works.  There's no reason to return a value -- nothing gets checked there.

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
warner: I'll figure out both calls later, feel free to leave it as-is.

packages/SwingSet/docs/configuration.md
```diff
@@ -133,17 +184,78 @@ not yet supported.
 `creationOptions` are directives to the vat controller rather than to the vat itself:
 
 - `enablePipelining` specifies whether the vat should be launched with
-  promise pipelining enabled (defaults to `false` if omitted).
+  promise pipelining enabled.  Defaults to `false`.
+
+- `description` is an optional description string displayed in log messages for
```
warner: `description` just went away in #5525, landed a few days ago

packages/SwingSet/docs/configuration.md
```diff
@@ -133,17 +184,78 @@ not yet supported.
...
+Some swingset and vat configuration options can be dynamically updated within a
+running swingset.
+
+### Dynamic kernel options
+
+Swingset-wide (kernel) options are updated by calling the
+`changeKernelOptions(options)` method on the swingset's controller object, where
+`options` is an object indicating which options to change and the values to
+change them to.
+
```
warner, line 236: Add:

`controller.changeKernelOptions(options)` should be called when the kernel is idle, i.e. not in the middle of a `controller.run()`. Modifying options will cause state changes to be written to the block buffer (i.e. the `hostStorage` object provided to `buildVatController`), and the host application is solely responsible for committing those changes to persistent storage, so most applications will need to do a `commit()` some time after calling `controller.changeKernelOptions()`.

packages/SwingSet/docs/configuration.md
```diff
@@ -133,17 +184,78 @@ not yet supported.
...
+- `snapshotInterval`
+
+### Dynamic vat options
+
+The options for a single vat may be updated by sending the `changeOptions(options)`
+message to the vat's admin node (note that because the admin node is involved,
+dynamic options update is not available for statically configured vats, i.e.,
+vats configured using the config object described above).
+
```
warner, line 259: lets make the example a bit more precise, maybe ".. updated by using `E(adminFacet).changeOptions(options)` to send a `changeOptions` message to the particular vat's admin facet. Note that only dynamic vats have admin facets, so .."

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -645,7 +696,9 @@ export default function buildKernel(
     insistCapData(vatParameters);
 
     // eslint-disable-next-line no-use-before-define
-    assert(vatWarehouse.lookup(vatID));
+    if (!vatWarehouse.lookup(vatID)) {
```
warner, line 699: Note to self: need a test for this, bootstrap vat should use the admin facet to terminate a vat and then send an update, make sure vat-admin vat/device intermediate messages queue it in the useful order, then make sure nothing explodes

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -326,11 +339,31 @@ export default function makeKernelKeeper(
...
+  function getSnapshotInitial() {
+    return getNat('kernel.snapshotInitial');
+  }
+
+  function getSnapshotInterval() {
+    return getNat('kernel.snapshotInterval');
+  }
+
+  function setSnapshotInterval(interval) {
+    assert(isNat(interval), 'invalid snapshotInterval value');
```
warner, line 365: This might be more humane if it mentioned the non-Nat value in the error message.

packages/SwingSet/src/kernel/vat-admin-hooks.js
```diff
@@ -112,5 +112,20 @@ export function makeVatAdminHooks(tools) {
       terminateVat(vatID, true, reasonCD);
       return harden({ body: stringify(undefined), slots: [] });
     },
+
+    changeOptions(argsCapData) {
+      // marshal([vatID, options]) -> null
+      assert(argsCapData.slots.length === 0);
```
warner, line 118: Let's add a message of some sort here, this "shouldn't happen" since vat-vat-admin is doing the primary checks, but if something slips through, it'd be nice to see `changeOptions() refuses slots` or something we can grep for. (I've been burned by empty assertions one too many times, they can get hard to track down when stacks are broken for one reason or another).

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -121,6 +121,25 @@ export function buildRootObject(vatPowers) {
...
         return upgradeCompleteP;
       },
+      changeOptions(options) {
+        for (const option of Object.getOwnPropertyNames(options)) {
+          const value = options[option];
+          switch (option) {
+            case 'reapInterval':
+              assert(
+                value === 'never' || isNat(value),
+                'invalid reapInterval value',
```
warner, line 131: let's include the invalid value in the `assert` message, that will help with debugging

warner (approved):
Looks good, some cosmetic changes to make, then feel free to land

packages/SwingSet/src/kernel/state/kernelKeeper.js
FUDCo: I deliberately didn't do that in case the non-Nat value was something that exploded when you try to stringify it.

warner: unfortunate but yeah

FUDCo: Actually, I can do `${JSON.stringify(value)}` which is ugly but safer the `${value}`

FUDCo: On second thought, no, `JSON.stringify` will blow up on bigints (whereas `${value}` will blow up on symbols).

PR #5512, opened 2022-06-04
ci(actions): fix up cache keys

michaelfig:
## Description

Reenable caching and harmonise+differentiate cache keys to (hopefully) speed up CI again.

There should not be a need to disable caching anymore.

### Testing Considerations

- [X] Let the CI run to completion on one commit
- [x] rerun deployment job on that same commit
- [x] push a new commit and rerun all jobs


.github/actions/restore-node/action.yml
```diff
@@ -19,12 +19,14 @@ runs:
...
         node-version: ${{ inputs.node-version }}
+    - uses: kenchan0130/actions-system-info@v1
+      id: system-info
     - name: restore built files
       id: built
       uses: actions/cache@v1
       with:
         path: .
-        key: ${{ runner.os }}-${{ inputs.node-version }}-built-${{ github.sha }}
+        key: ${{ runner.os }}-${{ runner.arch }}-${{ steps.system-info.outputs.release }}-${{ inputs.node-version }}-node-${{ github.sha }}
```
mhofman: Nit: put `node` before the node version ;)

mhofman (commented):
I would like to see a testing strategy on this PR, even if it's manual: build and run all normal tests, then trigger deployment job on that same revision after all other tests have succeeded.

.github/workflows/deployment-test.yml
```diff
@@ -67,25 +67,24 @@ jobs:
           path: testnet-load-generator
           ref: ${{steps.get-loadgen-branch.outputs.result}}
 
-      - name: Move repos under /usr/src where scripts expect them
+      - name: Put repos under /usr/src where scripts expect them
         run: |
           set -e
           sudo mv "$GITHUB_WORKSPACE/testnet-load-generator" /usr/src/testnet-load-generator
-          sudo mv "$GITHUB_WORKSPACE" /usr/src/agoric-sdk
+          sudo cp -a "$GITHUB_WORKSPACE" /usr/src/agoric-sdk
```
mhofman, line 74: Interesting, why is this needed? And will the cache still be populated?

.github/actions/restore-golang/action.yml
```diff
@@ -24,14 +24,16 @@ runs:
...
       with:
         go-version: ${{ inputs.go-version }}
+    - uses: kenchan0130/actions-system-info@master
+      id: system-info
     - name: cache Go modules
       id: cache
       uses: actions/cache@v1
       with:
         path: ${{ env.GOPATH }}/pkg/mod
-        key: ${{ runner.os }}-${{ inputs.cache-key }}-go-${{ hashFiles('go.sum') }}
```
mhofman, line 32: Nit: remove the `cache-key` from the `inputs` definition above.

.github/actions/build-node/action.yml
```diff
@@ -1,50 +0,0 @@
...
-      shell: bash
-    - name: check dependencies
-      run: yarn check-dependencies
-      shell: bash
-    # 'yarn build' loops over all workspaces
-    - name: yarn build
-      shell: bash
-      run: yarn build
-
-    - uses: ./.github/actions/save-node
```
mhofman, line 47: We can remove `save-node` altogether now, right?

michaelfig:
@mhofman, I ran all the tests listed in the description.  This code is R4R.

.github/actions/build-node/action.yml
```diff
@@ -1,50 +0,0 @@
...
-  using: composite
-  steps:
-    - uses: actions/setup-node@v1
-      with:
-        node-version: ${{ inputs.node-version }}
-    - name: cache node modules
-      uses: actions/cache@v1
-      with:
-        path: ~/.cache/yarn
-        key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
```
mhofman, line 24: Would we not benefit from integrating the `~/.cache/yarn` cache into `restore-node` to speed up the first build of the revision?

mhofman:
I'm still a little confused as to the steps order. Won't `restore-node` be run by various workflows in parallel, negating the cache benefits? I though we'd need to switch to a triggered workflow mechanism to fully take advantage of this.

.github/workflows/deployment-test.yml
michaelfig: The problem this is solving is that putting the output of the chaintest in the build cache (`"$GITHUB_WORKSPACE"`) would interfere with a rerun of the deployment job (files already existed in the results directory).

So, copying the build outputs into `/usr/src/agoric-sdk` means that the chaintest always has a clean state to start from and can modify anything without it being persisted in the build cache.

mhofman: Did not think of that! We need tests for our test infra 🙃

.github/actions/build-node/action.yml
michaelfig: > I'm still a little confused as to the steps order. Won't `restore-node` be run by various workflows in parallel, negating the cache benefits? I though we'd need to switch to a triggered workflow mechanism to fully take advantage of this.

It's alright for running workflows in parallel... there may not be many benefits to the caching in this case, but it will speed up reruns.  The speed benefits of parallelism are still useful.  The sequential workflows (like 'test-all-packages') are the ones that benefit from the caching.

It's a tradeoff, but the point is to preserve correctness regardless of when the workflows use `restore-node` or `restore-golang`.

mhofman (approved):

PR #5521, opened 2022-06-06
test(vaultFactory): withAmountUtils helper

turadg:
_no ticket_

## Description

While working on https://github.com/Agoric/agoric-sdk/pull/5515 this easier way of working with amounts occurred to me: augmenting issuer kit with the functions that make amounts and ratios from its brands.

### Security Considerations

--

--

### Testing Considerations

Test util


packages/run-protocol/test/supports.js
```diff
@@ -1,15 +1,17 @@
 // @ts-check
 /* global setImmediate */
 
+import { AmountMath } from '@agoric/ertp';
 import binaryVoteCounterBundle from '@agoric/governance/bundles/bundle-binaryVoteCounter.js';
 import committeeBundle from '@agoric/governance/bundles/bundle-committee.js';
 import contractGovernorBundle from '@agoric/governance/bundles/bundle-contractGovernor.js';
+import * as utils from '@agoric/vats/src/core/utils.js';
```
Chris-Hibbert, line 8: can we split these out explicitly?  Oh, I see. it was just moved.   
`<grumble>`

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -114,26 +115,29 @@ test.before(async t => {
...
     loanTiming: {
       chargingPeriod: 2n,
       recordingPeriod: 6n,
     },
     minInitialDebt: 50n,
-    rates: defaultParamValues(runBrand),
-    runInitialLiquidity: AmountMath.make(runBrand, 1_500_000_000n),
-    aethInitialLiquidity: AmountMath.make(aethKit.brand, 900_000_000n),
+    rates: defaultParamValues(run),
+    runInitialLiquidity: run.make(1_500_000_000n),
```
Chris-Hibbert, line 124: Thanks. I prefer the simplest form of the name being the name of the kit, and `kit.make()` being the maker. I wouldn't mind if we changed to this style pervasively.

Though mine is not the only voice on the subject.

packages/run-protocol/test/supports.js
```diff
@@ -166,3 +164,26 @@ export const produceInstallations = (space, installations) => {
...
+
+/**
+ * @param {object} kit
+ * @param {Brand<'nat'>} kit.brand
+ * @param {Issuer<'nat'>} [kit.issuer]
+ * @param {Mint<'nat'>} [kit.mint]
+ */
+export const withAmountUtils = kit => {
+  return {
+    ...kit,
```
Chris-Hibbert, line 176: It would be nice if we could have two alternative forms of this. One with and one without the mint. If we normally passed this around in production code with the mint included that would be scary.   I see that at least the .mint form is only used in a test. Maybe it would be even better if `withAmountUtils` excluded the mint, and required that the mint be handled explicitly.

turadg: > If we normally passed this around in production code with the mint included that would be scary.

Agreed. That's why I put it under `test/` and not `src/`.

For production code I was thinking something similar that took just `brand`, but I think that would need to be part of a more holistic look at how we organize code for amounts, ratios, etc.

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: I think if you and I agree, that's good enough to adopt the style throughout `run-protocol/`.

packages/run-protocol/test/supports.js
Chris-Hibbert: If we agree not to include the mint, we can use the same pattern in production and test.

turadg: True. Do you want to include that in the scope of this PR? My preference is to land this as is and tackle the production another day.

Chris-Hibbert: I would not make widespread changes here, but it would be apropos to change the helpers now to match our target, n'est ce pas?

turadg: The thing is, in tests there are legitimate needs for the `mint`. e.g.,
https://github.com/Agoric/agoric-sdk/blob/262097f7b82e2fba1a56b6f191b8ddbd4296ddaf/packages/run-protocol/test/vaultFactory/test-liquidator.js#L277
https://github.com/Agoric/agoric-sdk/blob/262097f7b82e2fba1a56b6f191b8ddbd4296ddaf/packages/run-protocol/test/vaultFactory/test-liquidator.js#L389
https://github.com/Agoric/agoric-sdk/blob/262097f7b82e2fba1a56b6f191b8ddbd4296ddaf/packages/run-protocol/test/vaultFactory/test-liquidator.js#L479
https://github.com/Agoric/agoric-sdk/blob/262097f7b82e2fba1a56b6f191b8ddbd4296ddaf/packages/run-protocol/test/vaultFactory/test-vaultFactory.js#L301
https://github.com/Agoric/agoric-sdk/blob/262097f7b82e2fba1a56b6f191b8ddbd4296ddaf/packages/run-protocol/test/vaultFactory/test-vaultFactory.js#L301

It makes sense since tests often have the full issuer kit. If in production we add some utility similar to `withAmountUtils`, I don't think it will replace this one. I expect this one to use it and still return `mint` when it's defined.

packages/run-protocol/test/supports.js
turadg: If the grumble was for `*`, the use it to put them in the return of `setupBootstrap`. 
https://github.com/Agoric/agoric-sdk/blob/262097f7b82e2fba1a56b6f191b8ddbd4296ddaf/packages/run-protocol/test/supports.js#L90
Since it has no state, I think it would be better to import those functions directly when they're needed. But out of scope of course.

If the grumble was with imports moving, we could standardize using something like https://www.npmjs.com/package/organize-imports-cli

packages/run-protocol/test/supports.js
Chris-Hibbert: I agree that the mint is needed in several places in tests. It's also needed in a few places in production. I just think the helper should consistently not include the mint. In order to make that work in production we'll need a mechanism for passing around the helper without the mint and pass around the mint separately, and only when it is necessary. I think the same approach would work in tests as well, and the consistency would be a benefit for us.

But I'm willing to put off that discussion until we decide to add such a helper to the production side.

packages/run-protocol/test/supports.js
Chris-Hibbert: It was the former. I agree that whoever is getting `*` from `setupBootstrap` should import it directly.  But `<grumble>` because I agree that we shouldn't fix it now.

Chris-Hibbert (approved):

PR #5526, opened 2022-06-07
feat: add config flag to control `fakeDurable`

### LINKED ISSUE #5489: `fakeDurable` feature should be guarded by a kernel config option

The `fakeDurable` feature recently provided by #5481 is intended only for use at development time.  In particular, no production swingset should be running with vats containing fake durable collections or DVOs, as these are guaranteed to explode horribly in the event of any upgrade attempt, including a null upgrade.

## Description of the Design

By requiring a swingset configuration option governing the availability of the `fakeDurable` option, we can ensure that any attempt to use this feature in a production context will fail.  However, the "fake durable" concept per se is deeply entangled with the virtual objects and stores abstractions, which are part of liveslots rather than swingset itself.  Having a top-level configuration option at the kernel level specifically for this feature would be an inappropriate blurring of abstraction boundaries since the kernel itself knows nothing of the durable object abstractions. On the other hand, despite the abstraction boundary issue, it does make sense to govern this feature globally rather than making it a vat creation option, since we would not wish the feature to be available to any vat at all in operational contexts where it would be inappropriate for the feature to be used.

I propose instead that the swingset configuration have a generalized development mode flag whose value would be available to liveslots (or to any other vat-specific runtime that we might someday add), which would in turn be allowed to interpret this flag however it felt appropriate.  Liveslots would use flag this to regulate availability of the `fakeDurable` option, but I can imagine other development-time features being regulated by this flag, potentially in conjunction with vat-specific options, all in the service of code development.

I propose to add a global configuration option, `devMode`, which would be a boolean flag configured at the same level as `defaultManagerType` or `defaultReapInterval` are now.  This would be passed along to liveslots as part of its initial configuration parameters.

## Security Considerations

I don't believe this feature introduces any specific new security concerns; rather, its addition would plug a potential security hole that might otherwise result if ill-behaved vats were to make use of `fakeDurable` in a production context.

FUDCo:
Swingset now forbids the use of the `fakeDurable` feature unless a global config flag `enableFakeDurable` is set to true in the kernel's config options.

`fakeDurable` is enabled without configuration in the mock virtual stuff.

Closes #5489

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -102,6 +102,7 @@ export function makeCollectionManager(
   registerValue,
   serialize,
   unserialize,
+  enableFakeDurable,
 ) {
   // TODO(#5058): we hold a list of all collections (both virtual and
   // durable) in RAM, so we can delete the virtual ones during
@@ -784,6 +785,19 @@ export function makeCollectionManager(
     return Far('weakSetStore', weakSetStore);
   }
 
+  function assertCorrectDurabilityFlags(durable, fakeDurable) {
+    if (fakeDurable) {
+      assert(
+        enableFakeDurable,
+        'fakeDurable may only be used if enableFakeDurable is true',
+      );
+    }
+    assert(
+      !durable || !fakeDurable,
+      'durable and fakeDurable are mutually exclusive',
+    );
+  }
+
   /**
    * Produce a *scalar* big map: keys can only be atomic values, primitives, or
    * remotables.
@@ -802,10 +816,7 @@ export function makeCollectionManager(
       fakeDurable = false,
     } = {},
   ) {
-    assert(
-      !durable || !fakeDurable,
-      'durable and fakeDurable are mutually exclusive',
-    );
+    assertCorrectDurabilityFlags(durable, fakeDurable);
     const kindName = durable ? 'scalarDurableMapStore' : 'scalarMapStore';
     const [vobjID, collection] = makeCollection(
       label,
@@ -856,10 +867,7 @@ export function makeCollectionManager(
       fakeDurable = false,
     } = {},
   ) {
-    assert(
-      !durable || !fakeDurable,
-      'durable and fakeDurable are mutually exclusive',
-    );
+    assertCorrectDurabilityFlags(durable, fakeDurable);
     const kindName = durable
       ? 'scalarDurableWeakMapStore'
       : 'scalarWeakMapStore';
@@ -895,10 +903,7 @@ export function makeCollectionManager(
       fakeDurable = false,
     } = {},
   ) {
-    assert(
-      !durable || !fakeDurable,
-      'durable and fakeDurable are mutually exclusive',
-    );
+    assertCorrectDurabilityFlags(durable, fakeDurable);
     const kindName = durable ? 'scalarDurableSetStore' : 'scalarSetStore';
     const [vobjID, collection] = makeCollection(
       label,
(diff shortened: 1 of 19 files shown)


packages/SwingSet/src/kernel/vat-loader/manager-helper.js
```diff
@@ -264,6 +265,13 @@ function makeManagerKit(
     return vres;
   }
 
+  /**
+   * @returns { boolean }
+   */
+  function getEnableFakeDurable() {
```
warner: I'm not sure routing this query through `manager-helper.js` is necessary (or adds much). All the manager types already have access to `kernelKeeper`.. could we just have them call `kernelKeeper.getEnableFakeDurable()` directly?

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -781,6 +782,17 @@ export function makeCollectionManager(
     return Far('weakSetStore', weakSetStore);
   }
 
+  function assertCorrectDurabilityFlags(durable, fakeDurable) {
+    assert(
```
warner: I'd be inclined to spell this
```js
if (fakeDurable) {
  assert(enableFakeDurable, ..);
}
```

like you did in `makeVOM`.. sure it's logically equivalent, but I think easier to read with an `if`.

packages/SwingSet/test/enable-fake-durable/test-enable-fake-durable.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const [objStatus, objCapdata] = await run('testObj');
+  if (enableFakeDurable) {
+    t.is(objStatus, 'fulfilled');
+    t.deepEqual(objCapdata, { body: '{"@qclass":"undefined"}', slots: [] });
+  } else {
+    t.is(objStatus, 'rejected');
+    t.deepEqual(objCapdata, {
+      body: '{"@qclass":"error","errorId":"error:liveSlots:v1#70002","message":"fakeDurable may only be used if enableFakeDurable is true","name":"Error"}',
```
warner: This form of "assert capdata is an error" test is kinda fragile (the "70002" is smelly). For the sake of those who follow, I'd recommend something like at the end of `test-upgrade.js`:

```js
import { parse } from '@endo/marshal';
..

  const e = parse(capdata.body);
  t.truthy(e instanceof Error);
  t.regex(e.message, /upgrade\(\) received unknown options: bad/);
```

packages/SwingSet/test/test-liveslots.js
```diff
@@ -1005,6 +1005,7 @@ test('dropImports', async t => {
     () => {
       return { buildRootObject: build };
     },
+    false,
```
warner, line 1008: part of me wants to suggest that the `makeLiveSlots()` argument should declare a default `allowFakeDurable = false`, and then this line goes away, but part of me thinks explicit is better

warner (approved):
I think I'd prefer we remove the extra call through `manager-helper.js`, but the others suggestions are less important to me. Feel free to land without them if you feel strongly.

packages/SwingSet/src/kernel/vat-loader/manager-helper.js
FUDCo: That was what I tried to do initially, but there's one case where `kernelKeeper` is not in scope at the point where it's needed.

warner: counterproposal: https://github.com/Agoric/agoric-sdk/pull/5529

packages/SwingSet/test/test-liveslots.js
FUDCo: Suspenders AND belt: I'll default the argument in `makeLiveSlots` but pass it explicitly here anyway.

PR #5528, opened 2022-06-07
feat(cache): new `@agoric/cache` implementation

### LINKED ISSUE #5519: Persistent cache available to dapps and deploy scripts

The AMM dapp, for example, takes a really long time to initialize.  Also, deploy scripts often have needless round trips.

## Description of the Design

Let's expose a `cache` function via the wallet connection as well as in the deploy script endowments.  That object would use a local copy of the cache actually stored over the wallet bridge in the wallet backend.  Here is the API we could use:

- the ground state for a cache key value is `undefined`.  It is impossible to distinguish a set value of `undefined` from an unset key
- `cache(key): Promise<recentValue>` - get an eventually-consistent value for `key`
- `cache(key, (oldValue) => ERef<newValue>): Promise<newValue>` -  call the updater function transactionally with the current value of `key`, and update it to `newValue`.  Rerun the updater function with a new value if the transaction is  stale.
- `cache(key, (oldValue) => ERef<newValue>, assertedValue): Promise<newValue>` - same as above, but return a rejection if the current value of the cache is not deeply equal to `assertedValue`

This would allow idioms like:

```js
const { cache } = walletConnection || deployPowers;
// Only update name=Somebody if `name` is currently `undefined`.
await cache('name', () => 'Somebody', undefined);
// Increment the counter.
const nextValue = await cache('counter', (prior = 0) => prior + 1);
console.log('counter is', nextValue);
```

In order to avoid costly (in time, or in gas) round trips within dapps, the cache object's storage will be populated via a chain stream.  Then, to implement the `update` operation, we would use optimistic locking with retries:

1. get the value and sequence number (0=initial) for the specific key
2. do async work in the updater, possibly multiple round trips
3. submit the resulting value and generation to the wallet backend
4. in the backend, check that the sequence number indicates a change to the current key's sequence
5. if the sequence check fails, go back to step 1
6. update the key's sequence number and value

## Security Considerations

The wallet backend needs to have quotas implemented to prevent DoS from a dapp that tries to store too much data.  A simple limit on the number of keys with an LRU would allow automatic expiry.  Other than that, we don't expose any more attack surface than the bridge already has.

When saving updates to the wallet backend, we need to ensure the values contain no severable promises or presences.  We can leverage the new CapTP `isLocalOnly` check to do this within the wallet backend.

## Test Plan

Convert a dapp to use the cache, and verify that it works the same as before, but faster.

michaelfig:
closes: #5519 

## Description

Simple cache package for wallet clients (like deploy scripts, dapps, and the wallet frontend).

#dapp-fungible-faucet-branch: mfig-bigint-literals

### Testing Considerations

Tested in deploy scripts and in `cd packages/web-components && yarn start`.


packages/cache/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Cache client
+
+The client-side API is normally exposed as a single function named `cache`.  You
+can create a cache client function by running `makeCache(coordinator,
+follower)`.  If not specified, the default coordinator is just a local in-memory
+map without persistence.
+
+- the ground state for a cache key value is `undefined`.  It is impossible to distinguish a set value of `undefined` from an unset key
+- `cache(key): Promise<recentValue>` - get an eventually-consistent value for `key`
+- `cache(key, (oldValue) => ERef<newValue>): Promise<newValue>` -  call the updater function transactionally with the current value of `key`, and update it to `newValue`.  Rerun the updater function with a new value if the transaction is  stale.
```
samsiegart: Super nit: extra space before "stale". Is it also convention to have two spaces after a period?

packages/cache/jsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "target": "esnext",
+    "module": "esnext",
+
+    "noEmit": true,
+/*
+    // The following flags are for creating .d.ts files:
+    "noEmit": false,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+*/
```
samsiegart, line 13: Can we remove this? My usual inclination is to not check in commented code

packages/cache/src/cache.js
```diff
@@ -0,0 +1,51 @@
...
+   * @param {object | string | bigint | null | undefined | number | ((oldValue: any) => ERef<unknown>)} [update] a
+   * function that returns a `newValue` or promise for `newValue`
+   * @param {Pattern[]} optGuardPattern don't update unless this pattern matches
+   */
+  const cache = (key, update, ...optGuardPattern) => {
+    if (!update) {
+      return E.get(E(coordinator).getRecentState(key)).value;
+    }
+    const guardPattern =
+      optGuardPattern.length > 0 ? optGuardPattern[0] : M.any();
```
samsiegart, line 28: Does this need to be an array if we're only using the first one?

packages/cache/src/store.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+export const makeStoreCoordinator = (stateStore = makeScalarMapStore()) => {
+  const getCurrentState = key => {
+    if (!stateStore.has(key)) {
+      return GROUND_STATE;
+    }
+    return stateStore.get(key);
+  };
+  return Far('store cache coordinator', {
+    getRecentState: async key => getCurrentState(key),
```
samsiegart, line 23: Are these `async`s required? Can this just be `getRecentState: getCurrentState,`?

packages/cache/src/types.js
```diff
@@ -0,0 +1,25 @@
...
+
+import '@agoric/store/exported.js';
+
+// Ensure this is a module.
+export {};
+
+/**
+ * @typedef {object} State
+ * @property {bigint} generation
+ * @property {any} value
```
samsiegart, line 11: Question: any particular reason for using `any` over `unknown` here?

packages/web-components/src/AgoricWalletConnection.js
```diff
@@ -64,11 +66,23 @@ export const makeAgoricWalletConnection = (makeCapTP = defaultMakeCapTP) =>
...
       this._bridgePK = makePromiseKit();
+      this._cache = undefined;
 
       this.service.send({ type: 'reset' });
       this.isResetting = false;
     }
 
+    get cache() {
+      if (this._cache) {
+        // The cache is cached.
```
samsiegart, line 77: Nice

samsiegart (approved):
LGTM just some nits/questions

packages/cache/src/cache.js
michaelfig: Yes, otherwise there is no way to distinguish between `cache(key, val)` and `cache(key, val, undefined)`.

packages/cache/jsconfig.json
michaelfig: I copied from another package.  I think we should leave them the same until there is a clear reason to diverge.

packages/cache/src/store.js
michaelfig: The signature calls for a promise return, so an async wrapper is the best way to ensure no synchronous throws.

packages/cache/src/types.js
michaelfig: If I use `unknown`, the type will propagate to the caller, which means they always have to inspect the value rather than just using it without inspection.

packages/cache/README.md
michaelfig: Yes, two spaces after a period is stylish.

PR #5529, opened 2022-06-07
fix: don't add manager-helper method, use kernelKeeper directly

warner:
test to see if we can remove the manager-helper.js `mk.getEnableFakeDurable()` helper method


FUDCo (approved):
Not sure where the scope issue with `kernelKeeper` was then.  Maybe I'm about to find out :-)

PR #5534, opened 2022-06-07
fix(swingset): improve timer device

### LINKED ISSUE #4297: Chain halt: `E(agoric.chainTimerService).setWakeup(0, "banana")`
## Describe the bug
Calling `E(agoric.chainTimerService).setWakeup(0, "banana")` results in a chain halt

## To Reproduce
3. wait for chain to boot
4. Issue `E(agoric.chainTimerService).setWakeup(0, "banana")`
5. The chain will stop due to the following stack

```
2022-01-14T00:22:57.821Z block-manager: BEGIN_BLOCK error: (Error#1)
Error#1: error in devicePollFunction: (Error#2)

  at Object.poll (packages/SwingSet/src/devices/timer.js:28:14)
  at beginBlock (packages/cosmic-swingset/src/launch-chain.js:291:32)
  at kernelPerformAction (packages/cosmic-swingset/src/block-manager.js:48:13)
  at blockManager (packages/cosmic-swingset/src/block-manager.js:225:19)
  at toSwingSet (packages/cosmic-swingset/src/chain-main.js:307:12)
  at portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:121:22)
  at Object.fromGo (packages/cosmic-swingset/src/chain-main.js:138:31)

Nested error under Error#1
  Error#2: SO(x) must be called on a Presence, not undefined
  
    at SO (.../swingset-vat/src/kernel/deviceSlots.js:136:5)
    at eval (.../swingset-vat/src/devices/timer-src.js:189:11)
    at poll (.../swingset-vat/src/devices/timer-src.js:178:19)
    at poll (.../swingset-vat/src/devices/timer-src.js:256:12)
    at Object.poll (packages/SwingSet/src/devices/timer.js:26:22)
    at beginBlock (packages/cosmic-swingset/src/launch-chain.js:291:32)
    at kernelPerformAction (packages/cosmic-swingset/src/block-manager.js:48:13)
    at blockManager (packages/cosmic-swingset/src/block-manager.js:225:19)
    at toSwingSet (packages/cosmic-swingset/src/chain-main.js:307:12)
    at portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:121:22)
    at Object.fromGo (packages/cosmic-swingset/src/chain-main.js:138:31)
  
portHandler threw (Error#1)

4:22PM ERR CONSENSUS FAILURE!!! 
```
## Expected behavior
Calling setWakeup with invalid arguments should not result in a chain halt

## Platform Environment
 macOS 12.1, node v16.13.1
 I only have 16gb of ram

ref #4264

warner:
Previously, calling `setWakeup()` with an invalid handler would cause
the chain to halt when the wakeup time arrived: an exception was
raised while the host application called `poll()`. Now we check the
handler earlier, during `setWakeup()`, and throw an error immediately,
which is delivered as a rejected promise to the vat that invoked
`E(timerService).setWakeup()`.

fixes #4297

While writing the test, I found and fixed another bug. If the same
handler was added multiple times, `removeWakeup()` is supposed to
remove all instances. An iteration-over-shrinking-array bug caused
some instances to remain if they appeared in immediately successive
time slots.

The docs were updated to current configuration practices.

### DIFF
--- a/packages/SwingSet/docs/timer.md
+++ b/packages/SwingSet/docs/timer.md
@@ -5,71 +5,104 @@ architecture](devices.md). In order to install a Timer device, you first build
 a timer object in order to create the timer's endowments, source code, and 
 `poll()` function.
 
+## Kernel Configuration
+
+The timer service consists of a device (`device-timer`) and a helper vat (`vat-timer`). The host application must configure the device as it builds the swingset kernel, and then the bootstrap vat must finish the job by wiring the device and vat together.
+
 ```
-import { buildTimer, getTimerWrapperSourcePath } from `@agoric/swingset-vat`;
+import { buildTimer } from `@agoric/swingset-vat`;
 const timer = buildTimer();
 ```
 
-Then when configuring devices, the timer device can be added:
+The `timer` record contains three items: a `srcPath` that points to the timer device source code, an `endowments` which must be provided to the timer device instance, and a `poll` function for use by the host application.
 
-```
+The timer vat is added automatically, but the `config` record must still define the timer device:
+
+```js
 config.devices = [
    ...
-  ['timer', timer.srcPath, timer.endowments],
+  timer: { sourceSpec: timer.srcPath },
 ];
 ```
-    
-and the Timer vat set up: 
-```
-config.vats.set('timer', { sourcepath: getTimerWrapperSourcePath() });
-```
 
-The `t.poll(now)` must be called periodically (usually at Block start time)
-with the current time. When `poll()` returns true, the kernel has work to do, 
-and we must call `controller.run()` and manage any resulting state updates.
-
-The "wrapper" vat has to provide a function to allow registering the device
-and creating the service. The vat is responsible for creating repeater objects
-to wrap the device node as a capability that can be accessed by regular vat
-code.
-
-In `bootstrap()` for a particular SwingSet, we create a timerService, and make
-it accessible to the user in `home.localTimerService` (whose timestamps are
-measured in milliseconds).
-
-```
-const timerService = E(vats.timerWrapper).createTimerService(devices.timer);
-```
-
-Then users in the REPL can use the `localTimerService` (in milliseconds since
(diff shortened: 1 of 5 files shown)


packages/SwingSet/docs/timer.md
```diff
@@ -5,71 +5,104 @@ architecture](devices.md). In order to install a Timer device, you first build
...
-This promise will resolve to the current timestamp after 60 seconds from now.
+The `timerService` Presence returned by `createTimerService` is the API handle from which all timer services are obtained.
 
-Alteratively, if you wish to be able to cancel the delay, you can use the more
-powerful `setWakeup` method:
+## Host Application Invokes poll()
 
-```js
-timestamp = E(home.localTimerService).getCurrentTimestamp();
+Once the kernel is started, the host application will call `timer.poll(now)` periodically with the current time. This is the device's only way to learn about the passing of time. It remembers the most recent value to satisfy `getCurrentTimestamp()` requests, and it compares `now` against the list of upcoming wakeups and repeating timers to decide if something needs to be scheduled.
```
Chris-Hibbert, line 55: As I read this paragraph, I wondered whether monotonicity was enforced. Having read the code, it appears that it it not, and I vaguely recall a discussion that if the kernel decides to go back to an earlier time, the device should go along with it. Is it worth writing that up here? I don't remember what kind of restart kerfluffle could lead to that happening.

Chris-Hibbert (approved):
Thanks for this, and for adding a test and updating the docs!

packages/SwingSet/docs/timer.md
warner: Good point, I think the rule should be that `poll(olderTime)` should be ignored. It's the host application that provides time, but the kernel has state (and obligations that derive from that state), so if the host app claims that time is going backwards *but doesn't also rewind the kernel state*, the kernel should not fire wakeups twice or something. 

It looks like `updateTime()` (used by `poll()`) asserts that time has not gone backwards.

I'll add a note about the host being required to provide monotonic time.

Issue #5443, opened 2022-05-26
test-vault.js depends on unstated ordering; doesn't work with ava --serial


dckc:
## To Reproduce
1. cd `packages/run-protocol`
2. `yarn test test/vaultFactory/test-vault.js` works
3. `yarn test test/vaultFactory/test-vault.js --serial` fails

```
  bad collateral

  vault should hold 50 Collateral

  Difference:

    {
      brand: Object @Alleged: Collateral brand {
        getAllegedName: Function getAllegedName {},
        getDisplayInfo: Function getDisplayInfo {},
        isMyIssuer: Function isMyIssuer {},
      },
  -   value: 69n,
  +   value: 50n,
    }

  › packages/run-protocol/test/vaultFactory/test-vault.js:185:5
```

## Platform Environment
 - what version of the Agoric-SDK are you using? (run `git describe --tags --always`) agoricdev-12-136-g2efb0e9a3

cc @Chris-Hibbert

turadg:
Obsolete

PR #5536, opened 2022-06-07
fix adjusting balances after interest charged

### LINKED ISSUE #5530: adjust balances after interest accrued failing because key not found
## Describe the bug

Observed getting `Promise.reject("Error: key (a string) not found in collection \"orderedVaultStore\"")` when a vault balance was adjusted. Code shows that while the key function requires normalized debts, the values passed in are current debts.

## Expected behavior

Key always found.

turadg:
closes: #5530

## Description

Fixes #5530

We did have a test that charged interest and then adjusted balances: https://github.com/Agoric/agoric-sdk/blob/4c9e50ac3a9420e18b789adb618c4503af097b1c/packages/run-protocol/test/vaultFactory/test-vault-interest.js#L112-L125

But the error required exercising `updateVaultAccounting` and the test ran https://github.com/Agoric/agoric-sdk/blob/4c9e50ac3a9420e18b789adb618c4503af097b1c/packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js#L143-L145
Adds to the motivation to remove that wrapper. https://github.com/Agoric/agoric-sdk/issues/5443 (h/t Hibbert)

I made a new test case in `test-vaultFactory` which reproed. Then I added a static type `NormalizedDebt` to find exactly where the wrong kind of value was being passed in.

### Security Considerations

--

Will land as merge commit.

### Testing Considerations

New test.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -242,21 +242,21 @@ const helperBehavior = {
    * maintain aggregate debt and liquidation order.
    *
    * @param {MethodContext} context
-   * @param {NormalizedDebt} oldDebt - prior principal and all accrued interest
+   * @param {NormalizedDebt} oldDebtNormalized - prior principal and all accrued interest, normalized to the launch of the vaultManager
    * @param {Amount<'nat'>} oldCollateral - actual collateral
-   * @param {Amount<'nat'>} newDebt - actual principal and all accrued interest
+   * @param {Amount<'nat'>} newDebtActual - actual principal and all accrued interest
    */
   updateDebtAccounting: (
     { state, facets },
-    oldDebt,
+    oldDebtNormalized,
     oldCollateral,
-    newDebt,
+    newDebtActual,
   ) => {
     const { helper } = facets;
-    helper.updateDebtSnapshot(newDebt);
+    helper.updateDebtSnapshot(newDebtActual);
     // update position of this vault in liquidation priority queue
     state.manager.updateVaultAccounting(
-      oldDebt,
+      oldDebtNormalized,
       oldCollateral,
       state.idInManager,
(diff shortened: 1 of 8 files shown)


Chris-Hibbert:
#5443 calls for fixing `test-vault.js`. That's the primary client of `vault-contract-wrapper.js`, and the most likely fix is to destroy `vault-contract-wrapper.js` and test against the actual contract.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -240,21 +242,21 @@ const helperBehavior = {
    * maintain aggregate debt and liquidation order.
    *
    * @param {MethodContext} context
-   * @param {Amount} oldDebt - prior principal and all accrued interest
-   * @param {Amount} oldCollateral - actual collateral
-   * @param {Amount} newDebt - actual principal and all accrued interest
+   * @param {NormalizedDebt} oldDebtNormalized - prior principal and all accrued interest, normalized to time zero
```
```suggestion
   * @param {NormalizedDebt} oldDebtNormalized - prior principal and all accrued interest, normalized to the launch of the vaultManager
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1299,6 +1299,67 @@ test('adjust balances', async t => {
   });
 });
 
+test.only('adjust balances after interest charges', async t => {
```
```suggestion
test('adjust balances after interest charges', async t => {
```

Chris-Hibbert (approved):
Looks good!  thanks for handling this quickly.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: ty. https://github.com/Agoric/agoric-sdk/issues/4941 is off the burner

PR #5538, opened 2022-06-07
only hold vaults that aren't settled

### LINKED ISSUE #5531: closing a vault doesn’t remove it from the vaultManager
## Describe the bug

Closing a vault leaves it in the vault manager. This prevents GC.

More significantly, because `liquidateAll()` iterates over all held vaults and a vault can't transition after being CLOSED, an exception is thrown if any vault has been closed:
```
Vault cannot transition from closed to liquidating
```

Related, it doesn't currently trigger the manager to `updateMetrics`.

## Expected behavior
`liquidateAll` does not depend on whether vaults have been closed.

Any vault in the CLOSED phase is not held by its old vault manager.

Moreover nothing is held in the vault manager that can't accrue interest. So LIQUIDATED is not held either. 

Design: vault manager only holds vaults that:
- can accrue interest
- can be liquidated

turadg:
closes: #5531

## Description

This introduces the concept of "settled", which means that they have no debts remaining. When vaults are liquidated, they are settled even though they may have had a shortfall, because the contract covered the shortfall using the reserve. Vaults can also be settled by closure.

Once a vault is settled, the manager has no more concern with it. Vault manager was removing vaults in the liquidation case and not the closure case. This makes it remove vaults upon settling.

### Security Considerations

`prioritizedVaults` adds a new `hasVaultByAttributes` lookup. This authority for information is subsumed by the extant `removeVaultByAttributes`.

Added a file comment for vaultManager explaining its scope and the invariant on what it retains.

### Testing Considerations

New tests. Updates to existing tests.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/run-protocol/src/vaultFactory/vaultDirector.js
@@ -283,6 +283,7 @@ const machineBehavior = {
         seat.incrementBy(mintSeat.decrementBy(harden({ RUN: kept })));
         zcf.reallocate(rewardPoolSeat, mintSeat, seat, ...otherSeats);
       } catch (e) {
+        console.error('mintAndReallocate caught', e);
         mintSeat.clear();
         rewardPoolSeat.clear();
         // Make best efforts to burn the newly minted tokens, for hygiene.
@@ -292,6 +293,9 @@ const machineBehavior = {
         debtMint.burnLosses(harden({ RUN: toMint }), mintSeat);
         throw e;
       } finally {
+        // Note that if this assertion may fail because of an error in the
+        // try{} block, but that error won't be thrown because this executes
+        // before the catch that rethrows it.
         assert(
           Object.values(mintSeat.getCurrentAllocation()).every(a =>
             AmountMath.isEmpty(a),
--- a/packages/run-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/run-protocol/src/vaultFactory/vaultDirector.js
@@ -299,8 +299,6 @@ const machineBehavior = {
           X`Stage should be empty of RUN`,
         );
       }
-      // TODO add aggregate debt tracking at the vaultFactory level #4482
-      // totalDebt = AmountMath.add(totalDebt, toMint);
       facets.machine.updateMetrics();
     };
 
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -1,5 +1,17 @@
 // @ts-check
-
+/**
+ * @file Vault Manager object manages vault-based debts for a collateral type.
+ *
+ * The responsibilities include:
+ * - opening a new vault backed by the collateral
+ * - publishing metrics on the vault economy for that collateral
+ * - charging interest on all active vaults
+ * - liquidating active vaults that have exceeded the debt ratio
+ *
+ * Once a vault is settled (liquidated or closed) it can still be used, traded,
+ * etc. but is no longer the concern of the manager. It can't be liquidated,
+ * have interest charged, or be counted in the metrics.
+ */
 import '@agoric/zoe/exported.js';
 
 import { E } from '@endo/eventual-send';
@@ -22,7 +34,7 @@ import {
 import { AmountMath } from '@agoric/ertp';
 
 import { defineKindMulti, pickFacet } from '@agoric/vat-data';
-import { makeVault } from './vault.js';
+import { makeVault, Phase } from './vault.js';
 import { makePrioritizedVaults } from './prioritizedVaults.js';
 import { liquidate } from './liquidation.js';
 import { makeTracer } from '../makeTracer.js';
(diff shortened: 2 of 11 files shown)


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -292,15 +293,16 @@ const machineBehavior = {
         debtMint.burnLosses(harden({ RUN: toMint }), mintSeat);
         throw e;
       } finally {
+        // Not that if this assertion may fail because of an error in the
```
Chris-Hibbert: If this correction isn't right, please reword.
```suggestion
        // Note that if this assertion may fail because of an error in the
```

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -33,7 +33,7 @@ import { makeTracer } from '../makeTracer.js';
 
 const { quote: q, details: X } = assert;
 
-const trace = makeTracer('XykAmm');
+const trace = makeTracer('XykAmm', false);
```
Chris-Hibbert, line 36: I'm approving these this time, but I'd prefer a separate tiny PR that just turns tracers off. I'll approve a PR that does that any time.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -47,15 +47,6 @@ export const makeAddIssuer = (
...
     await zcf.saveIssuer(secondaryIssuer, keyword);
-    const issuer = zcf.getIssuerForBrand(secondaryBrand);
-    console.log(
-      'Saved issuer',
-      secondaryIssuer,
-      'to keyword',
-      keyword,
-      'and got back',
-      issuer,
-    );
```
Chris-Hibbert, line 58: Please put this in a separate PR.

Chris-Hibbert (changes requested):

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: Is that a personal preference or a repo policy? I don't see anything about it in https://github.com/Agoric/agoric-sdk/blob/master/CONTRIBUTING.md

If the motivation is for the git history and blame attribution, I share that goal and that's why this PR has a separate tiny commit for the tracer changes. [d3a8dc9](https://github.com/Agoric/agoric-sdk/pull/5538/commits/d3a8dc9931d1551e08cf5c591a7212c48103c8ec)

If it's for a different reason, lemme know. I'll try to document it.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
turadg: Why? Same reason as https://github.com/Agoric/agoric-sdk/pull/5538#discussion_r895083300 ?

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: Yes, the motivation is for the git history and blame attribution.  How does a separate commit make a difference? Were you planning to not squish when merging? There are 6 commits currently, so would that mean squishing everything else and leaving one commit separate? that doesn't seem sustainable for reviewers or PR authors. Wouldn't it be much simpler to have a separate tiny PR?

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: Basically the same, but it's a more substantive change, and in the AMM. I don't think this is a trivial enough unrelated change to leave in a PR for vaults.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
turadg: > Were you planning to not squish when merging? 

Correct. That's why I've been careful to separate independent changes into different commits.

> Wouldn't it be much simpler to have a separate tiny PR?

It would be more work on top of what I've already done to separate the commits. It would be simpler in that we wouldn't have this debate on this PR but in the long run it means more work in general if we can't allow a PR to have multiple independent commits merged. It's worth getting aligned as an org on what is expected and allowed. (I say allowed because the request here was made as a blocking review, such that if someone else approves it the PR still can't be merged).

The CONTRIBUTING.md says,
> We favor "Squash and merge" (which requires the PR title to be a conventional commit) since it requires the least careful attention, especially when there are appended follow-up changes, but it's not the only tool in the toolbox.
> …
> We've also squashed follow-up commits manually (through an interactive rebase) after a PR had been approved and tests were passing, then landed the stack with "Rebase and merge" or "Create merge commit".

The latter is what I've done here.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -544,26 +556,54 @@ const managerBehavior = {
    * @param {NormalizedDebt} oldDebtNormalized
    * @param {Amount<'nat'>} oldCollateral
    * @param {VaultId} vaultId
+   * @param {import('./vault.js').VaultPhase} vaultPhase at the end of whatever change updatied balances
```
Chris-Hibbert:
```suggestion
   * @param {import('./vault.js').VaultPhase} vaultPhase at the end of whatever change updated balances
```

Chris-Hibbert (approved):

PR #5435, opened 2022-05-26
fix: makePublishKit

erights:
makePublishKit subsumes and eventually replaces the desired functionality of makeNotifierKit and makeSubscriptionKit.

PR #5541, opened 2022-06-08
fix(cosmic-swingset): Publish installation success and failure topic

### LINKED ISSUE #5460: Chain ACKs publish bundle
## Describe the bug
The `publishBundle` provided by `agoric-cli` in the `deployPowers` (added in https://github.com/Agoric/agoric-sdk/issues/4975) seem to not handled errors correctly:
- The publish can fail with an `account sequence mismatch` error, and there is no retry logic
- The `publishBundle` seem to be fire and forget, claiming success even in the presence of errors

## To Reproduce
1. Checkout https://github.com/Agoric/testnet-load-generator/tree/mhofman/publish-bundle
2. Run the loadgen: `SDK_SRC=/path/to/agoric-sdk ./start.sh`

## Expected behavior
Successful loadgen

## Additional context
@kriskowal mentions that we should handle both failures of sending the bundle to the chain, and failure of the bundle to be installed on the chain.

## Screenshots
```
agoric: deploy: /workspaces/agoric/agoric-sdk/golang/cosmos/build/agd tx swingset install-bundle --gas auto --gas-adjustment 1.2 --home /workspaces/agoric/testnet-load-generator/_agstate/agoric-servers/local-solo-8000/ag-cosmos-helper-statedir --keyring-backend test --from ag-solo --chain-id agoric --yes @/tmp/agoric-cli-bundle--21515-uAC3PJ5CCakJ/bundle.json
Error: rpc error: code = InvalidArgument desc = account sequence mismatch, expected 14, got 13: incorrect account sequence: invalid request
Usage:
  agd tx swingset install-bundle <JSON>/@<FILE>/- [flags]
```

kriskowal:
closes: #5460

## Description

Deploy scripts can now publish bundles directly to the chain.  Currently, they receive no feedback as to whether the message was adopted into the chain, much less dequeued and successfully processed.  This change will create an on-chain publication that the off-chain deploy script can subscribe to, to learn of the fate of their attempted installation.  The deploy script machinery will then watch for success, failure, or timeout.  If the installation times out, it will retry.

### Testing Considerations

For manual verification, in `dapp-fungible-faucet`:

```sh
# tab 1
agoric start local-chain --reset

# tab 2
agoric start local-solo --reset

# tab 3
agoric follow :published.installation

# tab 4
agoric deploy contract/deploy.js --need=agoric
```


kriskowal:
We got publishing to work, huzzah.

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -444,11 +447,24 @@ export default async function main(progname, args, { env, homedir, agcc }) {
     if (!blockingSend) {
```
gibson042, line 509:
```suggestion
    // Ensure that initialization has completed.
    if (!blockingSend) {
```

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -84,7 +128,21 @@ export default function makeBlockManager({
...
-          return validateAndInstallBundle(bundle);
+
+          const error = await validateAndInstallBundle(bundle).then(
+            () => null,
+            (/** @type {unknown} */ errorValue) => errorValue,
+          );
+
+          const { endoZipBase64Sha512 } = bundle;
+
+          publishInstallation(
```
gibson042: Is the change from `return validateAndInstallBundle(bundle)` to returning nothing intentional?

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -6,16 +6,52 @@ import { assert, details as X } from '@agoric/assert';
...
+ */
+const makeStoredPublisher = (initialValue, storageNode) => {
+  let providedPublication;
+
+  const provide = () => {
+    if (providedPublication) {
+      return providedPublication;
+    }
+    /** @type {SubscriptionRecord<T>} */
+    const { publication, subscription } = makeSubscriptionKit(initialValue);
```
gibson042: This should probably build on #5435 to avoid the cost of holding all results.
```suggestion
    const { publisher, subscriber } = makePublishKit(initialValue);
```

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -6,16 +6,52 @@ import { assert, details as X } from '@agoric/assert';
...
 
+/** @typedef {Record<string, unknown>} InstallationNotification */
+
+/**
+ * @template T
+ * @param {T} initialValue
+ * @param {StorageNode} [storageNode]
+ */
+const makeStoredPublisher = (initialValue, storageNode) => {
+  let providedPublication;
```
gibson042: Is there a reason not to initialize more fully up front? Setting up the dependencies should be inexpensive... is it instead a case of chainStorage writes happening too soon (and if so, is that fixed by switching to `makeEmptyPublishKit()`, which has no initial value)?
```suggestion
  const { publication, subscription } = makeSubscriptionKit(initialValue);
  makeStoredSubscription(subscription, storageNode);
  return { publish: publication.updateState };
```

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -6,16 +6,52 @@ import { assert, details as X } from '@agoric/assert';
...
+
+  const provide = () => {
```
gibson042: If this pattern does remain, I think the name could be more clear.
```suggestion
  const getPublication = () => {
```

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -48,6 +90,8 @@ export default function makeBlockManager({
     let p;
     switch (action.type) {
       case ActionType.BEGIN_BLOCK: {
+        provideInstallationPublisher();
+
```
gibson042: Since `publish` itself calls `provide()`, is this even necessary?
```suggestion
```

gibson042 (approved):

packages/cosmic-swingset/src/block-manager.js
kriskowal: Yes. We’re sinking the promise from `validAndInstallBundle` and converting it into a publication. We `await` instead of `return`.

kriskowal:
@gibson042 I’m re-requesting review since I’ve updated this to use the new pubsub system and that led to substantial changes.

packages/cosmic-swingset/src/block-manager.js
kriskowal: In #6103, I made casting more resilient to uninitialized storage, but at this height of the PR stack, the `{coming: 'soon'}` value substantially greases the wheels.

packages/agoric-cli/src/publish.js
```diff
@@ -146,27 +174,42 @@ export const makeHttpBundlePublisher = ({ jsonHttpCall, getAccessToken }) => {
   return publishBundleHttp;
 };
 
+/**
+ * @param {string} address - a host or URL
```
gibson042:
```suggestion
 * @param {string} address - a domain name, IPv4 address, or URL
 * @returns {string} the corresponding URL, defaulting scheme to http
```
Note that the current implementation will not work for IPv6 addresses because those must be wrapped in square brackets to be used as URL hosts (e.g., `http://[2001:db8::1]/` rather than `http://2001:db8::1/`): https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2

This also seems like the kind of thing that should really be written _once_ [per language and cross-referenced] and exposed in a library, but I don't consider that a blocking concern for this PR.

For the record, the other places I see similar code:
* [packages/casting/src/leader-netconfig.js](https://github.com/Agoric/agoric-sdk/blob/ca150ce70eff4e5bf4bf01c2f51efc12c6e01998/packages/casting/src/leader-netconfig.js#L26)
* [packages/deployment/ansible/roles/cosmos-genesis/files/faucet-helper.sh](https://github.com/Agoric/agoric-sdk/blob/ca150ce70eff4e5bf4bf01c2f51efc12c6e01998/packages/deployment/ansible/roles/cosmos-genesis/files/faucet-helper.sh#L47)
* [packages/solo/src/chain-cosmos-sdk.js](https://github.com/Agoric/agoric-sdk/blob/ca150ce70eff4e5bf4bf01c2f51efc12c6e01998/packages/solo/src/chain-cosmos-sdk.js#L126)

packages/agoric-cli/src/publish.js
```diff
@@ -42,6 +42,18 @@
 
 const { details: X, quote: q } = assert;
 
+/**
+ * @template T
+ * @param {Array<T>} array
+ * @param {number} randomNumber
```
gibson042:
```suggestion
 * @param {number} randomNumber expected to be in [0, 1)
```

packages/cosmic-swingset/src/block-manager.js
```diff
@@ -44,6 +55,17 @@ export default function makeBlockManager({
...
 
+    if (
+      installationPublisher === undefined &&
+      installationStorageNode !== undefined
+    ) {
+      const marshaller = makeMarshal();
+      const { publisher, subscriber } = makePublishKit();
+      makeStoredSubscriber(subscriber, installationStorageNode, marshaller);
+      installationPublisher = publisher;
+      publisher.publish({ coming: 'soon' });
```
gibson042:
```suggestion
      // Publish an inconsequential initial value.
      publisher.publish({ coming: 'soon' });
```

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -502,14 +505,30 @@ export default async function main(progname, args, { env, homedir, agcc }) {
...
+      };
+      const rootStorageNode = makeChainStorageRoot(
+        toStorage,
+        'swingset',
+        'published',
+      );
+      const installationStorageNode = rootStorageNode.makeChildNode(
+        'installation',
+        { sequence: true },
+      );
```
gibson042:
```suggestion
      const rootStorageNode = makeChainStorageRoot(
        toStorage,
        'swingset',
        STORAGE_PATH.CUSTOM,
      );
      const installationStorageNode = rootStorageNode.makeChildNode(
        'installation',
        { sequence: true },
      );
```

But I wonder if "installation" should be top-level rather than a child of "published", in which case it would get its own constant in packages/vats/src/chain-storage-paths.js.

gibson042 (changes requested):
Looks good after a few tweaks; let me know if you think anything needs discussion.

packages/agoric-cli/src/publish.js
kriskowal: Agreed wrt consolidating the implementations. It was not trivial to discover the prevailing idiom for this.

packages/cosmic-swingset/src/block-manager.js
kriskowal: This may not be necessary after https://github.com/Agoric/agoric-sdk/pull/6103 and I’ll make a note.

packages/cosmic-swingset/src/chain-main.js
kriskowal: What’s the intended scope of the `published` namespace as opposed to the root of the storage tree? I’m open to abbreviating this to `installed`, `bundles`, or some such.

gibson042: "published" is the catchall namespace that was necessary to prevent collisions with system-relevant paths at "activityhash"/"mailbox"/etc.—it is intended that only a single vat gets access to the "published" storage node, that it can and does create and distribute arbitrarily-named child nodes, and that it does not share the "published" node without a coordination mechanism (basically, that it has complete and singular authority over the "published" subtree, which it partitions and delegates to code that ultimately includes third-party contracts).

Having this independent "published.installation" storage node would impose a requirement that "installation" is itself reserved in that vat, and I'd prefer to avoid having such reservations at multiple levels of the chain storage tree.

kriskowal: Perfect. That very clearly indicates that this topic should be top-level.

kriskowal:
Ready for another round, pending CI. Follow-up commits not yet squashed and individually reviewable.

packages/cosmic-swingset/src/chain-main.js
kriskowal: I changed the topic name to merely `bundles` at the top-level.

packages/vats/src/chain-storage-paths.js
```diff
@@ -9,4 +9,5 @@ export const ACTIVITYHASH = 'activityhash';
 export const BEANSOWING = 'beansOwing';
 export const EGRESS = 'egress';
 export const MAILBOX = 'mailbox';
+export const BUNDLES = 'bundles';
```
gibson042, line 12: Please also add an analogous constant to golang/cosmos/x/swingset/keeper/keeper.go to preserve synchronization as noted in the introductory comment.

gibson042 (approved):
LGTM after updating golang/cosmos/x/swingset/keeper/keeper.go .

PR #5543, opened 2022-06-08
ops: automate generation of snapshot divergence traces

mhofman:
refs: #5227

## Description

I've been heavily updating the scripts I use to compare trace outputs from, including generating artifacts to report snapshot divergences. This is extracted from #5338 

Since the deployment integration test now captures an artifact regardless of success or failure, I also changed to remove heavy trace and storage files that are likely not needed.

### Security Considerations

None, only scripts used to process traces from the deployment integration test artifact.

The script implementation could use some explanations, but the usage of the script and content of the artifact should be fairly self explanatory.

The scripts produce an artifact with the following data:
- Difference of swingstore traces (minus host specific data, but including "local" data) between the 2 validator nodes
- Difference of swingstore traces (minus host specific data, but including "local" data) between the first validator node and the loadgen follower node
- The list of vats with diverging snapshots
- The list of snapshot hashes that are not common to all nodes (divergent)
- The divergent snapshots themselves stored by hash, as well as symlinks to them:
  - symlink in a structure of the shape `${node-desc}-xs-snapshots/v${vatId}/${i}-${startPos.itemCount}` where `i` is a `0` 3 digit padded 1-based index of snapshot for that execution of the node (aka 1st, 2nd, etc. snapshot taken by the worker process)
  - symlinks from the temporary file names used to load and save a snapshot (for easier trace reproductions)
- The XS worker debug traces for the divergent vats, stored under `${node-desc}-xsnap-trace/v${vatId}`
- A diff from the XS debug traces (minus snapshot command and options init) between the 2 validator nodes
- A diff from the XS debug traces (minus snapshot command and options init) between the first validator and the loadgen follow node
  - This handles offsetting between traces that start from different places (scratch or one of the snapshots). If a later restart of the node contains a trace for that vat, command executed after the last snapshot is taken are ignored
  - The offset logic only supports one of the 2 nodes restarting (loadgen monitor in this case) and relies on one of the node (validator in this case) staying up (continuous) as reference.
- Difference of the slog file (minus time field) between the 2 validator nodes
- Difference of the slog file (minus time field) between the first validator node and the stage 0 of the loadgen monitor node (I still need to automatically stitch slogs ignoring replays to include later stages)

An example of the output artifact on f513604013: https://drive.google.com/file/d/1-UyjKv0b_MlyyEegLNbvDGJUHgo1TaP-/view?usp=sharing

### Testing Considerations

Loads of manual tests against various CI artifacts I've collected here and there, or traces from local runs. It should be handling most edge cases (missing files, etc.)

### DIFF
--- a/packages/deployment/scripts/capture-integration-results.sh
+++ b/packages/deployment/scripts/capture-integration-results.sh
@@ -37,19 +37,30 @@ if [ -f "$RESULTSDIR/divergent_snapshots" ]; then
   # Snapshot divergences were found, fail the test after capturing results
   # TODO: uncomment once transient divergences are solved
   # ret=1
-elif [ $ret -eq 0 -a "$failedtest" = "false" ]; then 
-  echo "Successful test"
-  exit 0
+  # failedtest=true
 fi
 
 for node in validator{0,1}; do
   "$thisdir/setup.sh" ssh "$node" cat "$home/config/genesis.json" > "$RESULTSDIR/$node-genesis.json" || true
-  "$thisdir/setup.sh" ssh "$node" cat "$home/data/chain.slog" > "$RESULTSDIR/$node.slog" || true
-  "$thisdir/setup.sh" ssh "$node" cat "$home/data/ag-cosmos-chain-state/flight-recorder.bin" > "$RESULTSDIR/$node-flight-recorder.bin" || true
-  "$thisdir/setup.sh" ssh "$node" cat "$home/data/kvstore-trace" > "$RESULTSDIR/$node-kvstore-trace" || true
-  "$thisdir/setup.sh" ssh "$node" tar -cz -C "$home/data/xsnap-trace" . > "$RESULTSDIR/$node-xsnap-trace.tgz" || true
-  mkdir -p "$RESULTSDIR/$node-xs-snapshots" && "$thisdir/setup.sh" ssh "$node" tar -c -C "$home/data/ag-cosmos-chain-state/xs-snapshots" . | tar -x -C "$RESULTSDIR/$node-xs-snapshots" || true
+  "$thisdir/setup.sh" ssh "$node" cat "$home/data/chain.slog" > "$RESULTSDIR/$node.slog" || \
+    "$thisdir/setup.sh" ssh "$node" cat "$home/data/ag-cosmos-chain-state/flight-recorder.bin" > "$RESULTSDIR/$node-flight-recorder.bin" || true
+  if [ "$failedtest" != "false" ]; then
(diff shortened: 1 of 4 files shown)


warner (approved):
The idea and description sounds great to me. I don't have opinions on the implementation, but it doens't look like it could hurt anything.

mhofman:
I missed that @michaelfig updated the artifact to always be uploaded. I will update this PR to make sure there is a meaningful set of data in those cases

packages/deployment/scripts/capture-integration-results.sh
```diff
@@ -16,38 +16,51 @@ for node in validator{0,1}; do
...
-diff -U0 <(head -n $commonlen $RESULTSDIR/validator0-swingstore-trace) <(head -n $commonlen $RESULTSDIR/validator1-swingstore-trace) > $RESULTSDIR/validator-swingstore-trace.diff || ret=$?
+"$thisdir/../../../scripts/process-integration-swingstore-traces.sh" "$RESULTSDIR" || ret=$?
 
 failedtest=${1:-"unknown"}
 
-if [ $ret -eq 0 ]
-then
-  if [ "$failedtest" = "false" ]
+if [ -f "$RESULTSDIR/divergent_snapshots" ]; then
+  if [ $(cat "$RESULTSDIR/validator-swingstore-trace.diff" | wc -l) -gt 0 ]; then
```
michaelfig: Avoid catting the whole file if you're just testing for nonzero size:

```suggestion
  if [ -s "$RESULTSDIR/validator-swingstore-trace.diff" ]; then
```

packages/deployment/scripts/capture-integration-results.sh
```diff
@@ -16,38 +16,51 @@ for node in validator{0,1}; do
...
+    cat "$RESULTSDIR/validator-swingstore-trace.diff" | cut -c -80 || true
+    echo "Error: Swingstore trace mismatch between validators"
+  fi
+
+  if [ -f "$RESULTSDIR/monitor-vs-validator-swingstore-trace.diff" ] && \
+     [ $(cat "$RESULTSDIR/monitor-vs-validator-swingstore-trace.diff" | wc -l) -gt 0 ]
```
michaelfig: Again:
```suggestion
     [ -s "$RESULTSDIR/monitor-vs-validator-swingstore-trace.diff" ]
```

packages/deployment/scripts/capture-integration-results.sh
mhofman: I did not know about this one.

michaelfig (approved):
LGTM!  Just an opportunity to avoid `cat ... | wc -l`.

PR #5551, opened 2022-06-09
Tighten up the casting and cache APIs

michaelfig:
refs: #5519, #4398

## Description

Complete end-to-end deploy script and wallet UI casting followers.

Many new features and tightening of old ones.  Reviewing commit-by-commit is recommended, but I'm leaning on @samsiegart for the wallet changes, @JimLarson for the `vstorage` changes, and @kriskowal for the caching/casting architecture.  Any comments from @gibson042 are also appreciated, bearing in mind that this needs continuing work to integrate with his changes.

### DIFF
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -7,6 +7,7 @@ backend.  Any passable object can be a cache key or a cache value.
 
 ```js
 import { makeCache, makeStoreCoordinator } from '@agoric/cache';
+import { M } from '@agoric/store';
 import { makeScalarBigMapStore } from '@agoric/vat-data';
 
 const store = makeScalarBigMapStore('cache');
@@ -21,23 +22,23 @@ await cache('baz', 'barbosa'); // 'barbosa'
 await cache('baz', 'babaloo', undefined); // 'barbosa'
 await cache('baz', 'babaloo', 'barbosa'); // 'babaloo'
 
+// One-time initialization.
+await cache('frotz', 'default'); // 'default'
+await cache('frotz', 'ignored'); // 'default'
 
 // Update the `'foo'` entry, using its old value (initially `undefined`).
 await cache('foo'); // `undefined`
 const updater = (oldValue = 'bar') => `${oldValue}1`;
-await cache('foo', updater); // 'bar1'
-await cache('foo', updater); // 'bar11'
+await cache('foo', updater, M.any()); // 'bar1'
+await cache('foo', updater, M.any()); // 'bar11'
 await cache('foo'); // 'bar11'
 
-// You can also assert a pattern for the key.  If it doesn't match, then the cache returns a rejection.
+// You can also specify a guard pattern for the value to update.  If it
+// doesn't match the latest value, then the cache isn't updated.
 await cache('foo', updater, 'nomatch'); // 'bar11'
 await cache('foo', updater, 'bar11'); // 'bar111'
 await cache('foo', updater, 'bar11'); // 'bar111'
 await cache('foo'); // 'bar111'
-
-// Specify a pattern of `undefined` for one-time initialisation.
-await cache('frotz', 'default', undefined); // 'default'
-await cache('frotz', 'ignored', undefined); // 'ignored'
 ```
 
 ## Cache client
@@ -63,26 +64,26 @@ The cache coordinator must implement the `Coordinator` interface, which supports
 eventual consistency with optimistic updates:
 
 ```ts
-interface State {
-  // State updates must include a generation counter exactly one
-  // greater than the current state, or they are not applied.
-  generation: bigint,
-  // Any acceptable value supported by the cache.
-  value: any,
-};
-
-// The default state for a key that has not yet been updated.
-const GROUND_STATE = { generation: 0n, value: undefined };
+interface Updater {
+  /**
+   * Calculate the newValue for a given oldValue
+   */
+  update: (oldValue: Passable) => unknown
+}
 
 interface Coordinator {
   /**
-   * Read an eventually-consistent state for the specified key.
+   * Read an eventually-consistent value for the specified key.
    */
-  getRecentState: (key: unknown) => Promise<State>,
+  getRecentValue: (key: Passable) => Promise<Passable>,
   /**
-   * Attempt to update the key to the new state.  Returns the latest known
-   * state after trying to apply the desiredState (may not match).
+   * Update a cache value to newValue, but only if guardPattern matches the current value.
    */
-  tryUpdateState: (key: unknown, desiredState: State, assertedMatch: Matcher) => Promise<State>,
+  setCacheValue: (key: Passable, newValue: Passable, guardPattern: Pattern) => Promise<Passable>,
+  /**
+   * Update a cache value via an updater calculation of newValue, but only if guardPattern
+   * matches the current value.
+   */
+  updateCacheValue: (key: Passable, updater: ERef<Updater>, assertedMatch: Matcher) => Promise<Passable>,
 }
 ```
--- a/packages/cache/src/store.js
+++ b/packages/cache/src/store.js
@@ -1,55 +1,133 @@
 // @ts-check
-import { Far } from '@endo/far';
+import { E, Far } from '@endo/far';
+import { deeplyFulfilled, makeMarshal } from '@endo/marshal';
 import { matches, makeScalarMapStore } from '@agoric/store';
 
 import '@agoric/store/exported.js';
 
-import { GROUND_STATE } from './state.js';
-
-const { details: X } = assert;
+import { GROUND_STATE, makeState } from './state.js';
 
 /**
- * @param {MapStore<unknown, import('./types').State>} [stateStore]
- * @returns {import('./types').Coordinator}
+ * @param {(obj: unknown) => unknown} [sanitize]
+ * @returns {(key: Passable) => Promise<string>}
  */
-export const makeStoreCoordinator = (stateStore = makeScalarMapStore()) => {
-  const getCurrentState = key => {
-    if (!stateStore.has(key)) {
+const makeKeyToString = (sanitize = obj => obj) => {
+  let lastNonce = 0;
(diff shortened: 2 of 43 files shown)


packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -1,26 +1,29 @@
...
     harden({
       board: E(walletBridge).getBoard(),
     }),
   );
 
+  /**
+   * @template T
+   * @param {ERef<Notifier<T>>} notifier
+   */
+  const iterateNotifier = notifier => {
```
samsiegart, line 22: Is this something that we'd want to reuse across dapps?

samsiegart (commented):
Wallet changes LGTM

packages/wallet/ui/src/util/WalletBackendAdapter.js
michaelfig: Not really.  It's a hack to deal with notifiers.   @gibson042's new `makePublisherKit` will be the way forward (instead of the wallet's current `makeNotifierKit`), and that will have helpers in `@agoric/notifier` to consume them.   At that time, we will also update the `@agoric/casting` follower API to conform with them.

packages/cache/README.md
```diff
@@ -63,26 +64,26 @@ The cache coordinator must implement the `Coordinator` interface, which supports
...
-  getRecentState: (key: unknown) => Promise<State>,
+  setCacheValue: (key: unknown, newValue: any, guardPattern: Pattern) => Promise<any>,
   /**
-   * Attempt to update the key to the new state.  Returns the latest known
-   * state after trying to apply the desiredState (may not match).
+   * Update a cache value via an updater calculation of newValue, but only if guardPattern
+   * matches the current value.
    */
-  tryUpdateState: (key: unknown, desiredState: State, assertedMatch: Matcher) => Promise<State>,
+  updateCacheValue: (key: unknown, updater: ERef<Updater>, assertedMatch: Matcher) => Promise<any>,
```
kriskowal: What’s the rationale for returning `any` in `Promise<any>` as opposed to `unknown`?

packages/cache/src/store.js
```diff
@@ -1,55 +1,104 @@
...
+  };
+  return keyToString;
+};
 
 /**
- * @param {MapStore<unknown, import('./types').State>} [stateStore]
- * @returns {import('./types').Coordinator}
+ * @param {MapStore<string, import('./types').State>} [stateStore]
  */
 export const makeStoreCoordinator = (stateStore = makeScalarMapStore()) => {
```
kriskowal: Can I trouble you for a description jsdoc for what this maker is intended to do?

packages/cache/src/store.js
```diff
@@ -1,55 +1,104 @@
...
       if (nextGeneration <= 1n) {
-        stateStore.init(key, nextState);
+        stateStore.init(keyStr, nextState);
       } else {
-        stateStore.set(key, nextState);
+        stateStore.set(keyStr, nextState);
+      }
+      return newValue;
+    },
+    updateCacheValue: async (key, updater, guardPattern) => {
```
kriskowal, line 123: Riffing on this name: “updateValue” but with “cache” is a bit dissatisfying. This differs from “updateValue” because it doesn’t take a value. “generateUpdateAsync” might be more apt. The “async” bit is what necessitates the compare-and-swap loop.

“cache” is dissatisfying because this isn’t a bank of cells that arbitrarily evict a cell to make remove for a new value. This is more like a map of cells.

packages/cache/src/store.js
```diff
@@ -1,55 +1,104 @@
...
+          const nextState = makeState(newValue, latestState);
+          if (nextGeneration <= 1n) {
+            stateStore.init(keyStr, nextState);
+          } else {
+            stateStore.set(keyStr, nextState);
+          }
+          return newValue;
+        }
+
+        // Someone else updated the state while we were calculating the new value, so loop again.
```
kriskowal: I can’t reconcile line 79, aborting if the basis shifted, and here, retrying if the generation advanced (and presumably shifted the basis?).

packages/casting/src/follower.js
```diff
@@ -0,0 +1,61 @@
...
+    getLatestIterable: async () => {
+      const { notifier, subscription } = await spec;
+      let ai;
+      if (notifier) {
+        ai = makeNotifierIterable(notifier);
+      } else {
+        assert(subscription);
+        ai = makeSubscriptionIterable(subscription);
+      }
+      return mapAsyncIterable(ai, transform);
```
kriskowal, line 29: `mapReader` might be applicable https://github.com/endojs/endo/blob/master/packages/stream/index.js#L200-L216

packages/casting/src/iterable.js
```diff
@@ -1,28 +1,75 @@
...
+  // https://web.archive.org/web/20160404122250/http://wiki.ecmascript.org/doku.php?id=strawman:concurrency#infinite_queue
+  return Far('SubscriptionIterator', {
+    next: async () => {
+      const resultP = E.get(tailP).head;
+      tailP = E.get(tailP).tail;
+      Promise.resolve(tailP).catch(() => {}); // suppress unhandled rejection error
+      return resultP;
+    },
+  });
+};
```
kriskowal: Yes!

kriskowal (approved):
I’m not entirely sure what this is going for, but the pattern of “async iterator that produces a stream of values that eventually converge, such that the iterator samples the most recent update on demand” is a good code golf exercise.

This is an untested sketch of a cell that provides a lossy update iterator.

```js
const makeCell = value => {
  let changed = makePromiseKit();
  const iterate = () => ({
    next() {
      changed = makePromiseKit();
      return {done: false, value};
    },
  });
  const set = newValue => {
    value = newValue;
    // multiple updates can occur before the subscribers sample, so a fast
    // producer does not produce a backlog.
    changed.resolve();
  };
  return {iterate, set};
};
```

And an untested sketch of the form if the sequence of updates can end:

```js
const makeLossyStream = value => {
  let changed = makePromiseKit();
  let ended = makePromiseKit();
  changed.resolve();

  const iterate = () => ({
    async next() {
      changed = makePromiseKit();
      return Promise.race([ended.promise, {done: false, value}]);
    },
  });

  const generate = () => ({
    next(newValue) {
      value = newValue;
      changed.resolve();
    },
    return(value) {
      ended.resolve({done: true, value});
    },
    throw(error) {
      ended.reject(error);
    },
  });

  return {iterate, generate};
};
```

Your update generator could be generalized in terms of compare-and-swap of a {get, set} cell with a `generate` function, except I’m not sure how you’re deciding whether to abort or rebase the update.

kriskowal:
Oh, my stamp is green because the jitter stuff you called me for looks good to me. Happy to review again if it’s cost-effective to help me understand the other parts I didn’t grasp.

michaelfig:
> I’m not entirely sure what this is going for

I should have been more explicit.  This is an interim implementation until we have the infrastructure in #5435 .  There will be a new PR to use that mechanism at that point, and far less open-coding of async iterators.

packages/cache/README.md
michaelfig: This is a vestige from an earlier sketch.  You're right; it should use `unknown`.

packages/cache/src/store.js
michaelfig: This API is specifically designed to permit arbitrary eviction of cells.  It's the current implementation that doesn't, so I'm inclined to leave the name and add a comment.

packages/cache/src/store.js
michaelfig: Rewritten for clarity.

packages/cache/README.md
michaelfig: Actually, it's Passable.

packages/casting/src/follower.js
michaelfig: Problem with mapReader is that it yields an `undefined` at the end of the stream.  That violates the expectations of subscriptions/notifiers.

packages/casting/src/iterable.js
michaelfig: Totally cribbed from `@agoric/notifier`, so I can't take credit. :)

packages/web-components/web-dev-server.config.mjs
```diff
@@ -21,6 +28,22 @@ export default /** @type {import('@web/dev-server').DevServerConfig} */ ({
...
+          '/__wds-outside-root__/**/node_modules/bech32/**/*',
+          '/__wds-outside-root__/**/node_modules/libsodium-wrappers/**/*',
+          // '/__wds-outside-root__/**/node_modules/libsodium/**/*',
+          // '**/node_modules/<package-b>/**/*',
+        ],
+      }),
+    false &&
+      polyfill({
+        include: ['/__wds-outside-root__/**/node_modules/libsodium/**/*'],
+      }),
```
kriskowal, line 46: This seems ineffectual.

packages/web-components/demo/index.html
```diff
@@ -15,21 +15,24 @@
     import './install-ses-lockdown.js';
 
     import { E } from '@endo/eventual-send';
+    // import { makeFollower, iterateLatest } from '@agoric/casting';
     import { html, render } from 'lit';
     import '../agoric-wallet-connection.js';
 
+    const { makeFollower, iterateLatest } = {}; // FIXME replace
```
kriskowal: fix me?

kriskowal (approved):

packages/web-components/web-dev-server.config.mjs
michaelfig: Ineffectual, but getting closer to working.  I couldn't enable it yet because libsodium is still broken under SES.  I'll add a comment.

packages/web-components/demo/index.html
michaelfig: Will do!

JimLarson (commented):
I skimmed the golang changes and it looks okay to me.

PR #5552, opened 2022-06-09
feat: canonicalize order of vat & bundle configuration

### LINKED ISSUE #5229: swingset should be insensitive to config.bundles insertion order
I'd like to declare that the kernel depends upon the visible contents of `config` but not the latent insertion order. That means we need to process things like `config.vats`, `config.devices`, and `config.bundles` in sorted-keys order.

I'd also like to put an explicit sort after any use of `Promise.all`, to make it easier to audit that we're insensitive to such things.

So in `src/controller/initializeSwingset.js`, every place where we use `Object.keys` should be examined to determine whether the object being listed is the output of a `Promise.all`. If so, we should sort the results before merging them into the next step.

 At a quick glance, this includes at least:
 * near the start of `initializeSwingset`, when we provide empty defaults for each of `config.vats/bundles/devices`, we should sort the tables at that point
 * `normalizeConfigDescriptor`: where we walk `Object.keys(desc)` to build up a list of normalization jobs, we don't need to sort first, because we're going to sort after
   * `normalizeConfigDescriptor` is called from the start of `loadSwingsetConfigFile`, and modifies `config.vats`/etc in-place
* `processGroup` does `const names = Array.from(Object.keys(group))`, uses that to build a list of `processDesc()` promises, then uses `bundlesWithID = await Promise.all()` to await the whole batch. We do not iterate the `bundlesWithID` that comes out (we iterate `names` instead), so as long as we sorted `config.vats/etc` earlier, `names` should be consistent
* After calling `processGroup` for each group, we merge their returned `idToBundle` records into `kconfig.idToBundle`. By sorting the groups earlier, this should be stable, however it would be nice to have the insertion order be sorted by `bundleID` so that any overlap between the groups doesn't cause confusion.
* After that, we iterate `nameToBundle` to populate `kconfig.namedBundleIDs`. `nameToBundle`'s insertion order will be equal to the insertion order of `config.bundles`, followed by `vats` and `devices`. Again, it would be nice if this were sorted by name.

_Originally posted by @warner in https://github.com/Agoric/agoric-sdk/issues/5226#issuecomment-1110100695_

FUDCo:
Closes #5229

### DIFF
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -240,6 +240,22 @@ export function swingsetIsInitialized(hostStorage) {
   return !!hostStorage.kvStore.get('initialized');
 }
 
+/**
+ * @param {Record<string, any>} obj
+ * @param {(string|undefined)[]} [firsts]
+ * @returns {Record<string, any>}
+ */
+function sortObjectProperties(obj, firsts = []) {
+  const sorted = [...firsts, ...Object.keys(obj).sort()];
+  const result = {};
+  for (const prop of sorted) {
+    if (prop && result[prop] === undefined && obj[prop] !== undefined) {
+      result[prop] = obj[prop];
+    }
+  }
+  return result;
+}
+
 /** @typedef {{ kernelBundles?: Record<string, string>, verbose?: boolean,
  *              addVatAdmin?: boolean, addComms?: boolean, addVattp?: boolean,
  *              addTimer?: boolean,
@@ -438,7 +454,7 @@ export async function initializeSwingset(
 
   async function processGroup(groupName, nameToBundle) {
     const group = config[groupName] || {};
-    const names = Array.from(Object.keys(group));
+    const names = Object.keys(group).sort();
     const processP = names.map(name =>
       processDesc(group[name], nameToBundle).catch(err => {
         throw Error(`config.${groupName}.${name}: ${err.message}`);
@@ -462,6 +478,15 @@ export async function initializeSwingset(
   // (to populate config.namedBundles) and the bundleID->bundle record (to
   // install the actual bundles)
 
+  config.bundles = sortObjectProperties(config.bundles);
+  config.vats = sortObjectProperties(config.vats, [
+    config.bootstrap,
+    'vatAdmin',
+    'comms',
+    'vattp',
+    'timer',
+  ]);
+  config.devices = sortObjectProperties(config.devices);
   const [nameToBundle, idToNamedBundle] = await processGroup('bundles');
   const [_1, idToVatBundle] = await processGroup('vats', nameToBundle);
   const [_2, idToDeviceBundle] = await processGroup('devices', nameToBundle);
@@ -469,11 +494,11 @@ export async function initializeSwingset(
   const kconfig = {
     ...config,
     namedBundleIDs: {},
-    idToBundle: {
+    idToBundle: sortObjectProperties({
       ...idToNamedBundle,
       ...idToVatBundle,
       ...idToDeviceBundle,
-    },
+    }),
   };
   for (const name of Object.keys(nameToBundle)) {
     kconfig.namedBundleIDs[name] = nameToBundle[name].id;
--- a/packages/SwingSet/test/virtualObjects/test-representatives.js
+++ b/packages/SwingSet/test/virtualObjects/test-representatives.js
@@ -378,47 +378,49 @@ test('virtual object gc', async t => {
     c.kpResolution(c.bootstrapResult),
     capargs({ '@qclass': 'undefined' }),
   );
+  const v = 'v6';
   const remainingVOs = {};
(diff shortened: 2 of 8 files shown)


packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -270,13 +293,19 @@ export async function initializeSwingset(
 
   // copy config so we can safely mess with it even if it's shared or hardened
   config = JSON.parse(JSON.stringify(config));
-  if (!config.bundles) {
+  if (config.bundles) {
+    config.bundles = sortObjectProperties(config.bundles);
```
warner: minor, but I'd make this a smidge more readable by doing the `sortObjectProperties` unconditionally just after the if:
```js
if (!config.bundles) {
  config.bundles = {};
}
config.bundles = sortObjectProperties(config.bundles);
```

perhaps doing all three sorts in a single block at the end of the three if clauses. Any thing to get rid of the extra `} else {` line.

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -240,6 +240,29 @@ export function swingsetIsInitialized(hostStorage) {
   return !!hostStorage.kvStore.get('initialized');
 }
 
+/**
+ * @param {Record<string, any>} obj
+ * @param {string} [first]
+ * @returns {Record<string, any>}
+ */
+function sortObjectProperties(obj, first) {
```
warner: It'd be even nicer if the order could be `${bootstrap}, vat-admin, comms, vattp, timer, ...uservats`. What about (note: untested):

```js
function sortObjectProperties(obj, firsts) {
  const sorted = [...firsts, Object.keys(obj).sort()]; // may have duplicates
  const result = {};
  for (const prop of sorted) {
    // don't add duplicate && only add if actually present
    if (!Object.hasOwn(result, prop) && Object.hasOwn(obj, prop)) {
      result[prop] = obj[prop];
    };
  }
  return result;
}
```

where the second test allows us to later say:
```js
config.vats = sortObjectProperties(config.vats, [config.bootstrap, 'vat-admin', 'comms', 'vattp', 'timer']);
```

without knowing whether those five vat names are actually present in `config.vats` or not.

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -469,11 +498,11 @@ export async function initializeSwingset(
   const kconfig = {
     ...config,
     namedBundleIDs: {},
-    idToBundle: {
+    idToBundle: sortObjectProperties({
       ...idToNamedBundle,
       ...idToVatBundle,
       ...idToDeviceBundle,
-    },
+    }),
```
warner, line 501: For a second I thought we'd need to sort the `Object.keys(nameToBundle)` on line 507 below, because `processGroup` takes `config.vats`/etc as inputs. But then I realized that the `sort` of `config.vats`/etc all happens well before this, so the results will still be deterministic.

It might still be nice to sort those, so the DB writes appear in a sorted order, but it's strictly cosmetic.

packages/SwingSet/test/virtualObjects/test-representatives.js
```diff
@@ -379,46 +379,46 @@ test('virtual object gc', async t => {
     capargs({ '@qclass': 'undefined' }),
   );
   const remainingVOs = {};
-  for (const key of hostStorage.kvStore.getKeys('v1.vs.', 'v1.vs/')) {
+  for (const key of hostStorage.kvStore.getKeys('v2.vs.', 'v2.vs/')) {
```
warner: What do you think about factoring that string out, to make this sort of change easier in the future? I guess it would make this delta even bigger, with all the single-quotes being turned into backquotes, for eg
```js
`${v}.vs.baggageID`: 'o+6/1',
```

warner (approved):
Happy to see this land as-is, but if we could get `vat-admin` to always be `v2` without too much trouble, I'd love it.

packages/SwingSet/src/controller/initializeSwingset.js
FUDCo: The second sort is not needed at all.  It's not that the results will still be deterministic if we don't, it's that they're already sorted at that point.

packages/SwingSet/test/virtualObjects/test-representatives.js
FUDCo: I'm inclined to agree, given the tedium of these kinds of edits.  If nothing else, Murphy's Law says that if we make it easier to update if the ordering changes in the future, then we ensure that such changes never happen.  Yay for stability!

packages/SwingSet/src/controller/initializeSwingset.js
FUDCo: I did something like that, but for future reference, XS does not implement `Object.hasOwn` yet.

mhofman:
@mergifyio refresh

mhofman:
@Mergifyio refresh

PR #5554, opened 2022-06-09
feat(vaultFactory): liquidated vaults have debts cleared

### LINKED ISSUE #5286: When vault moves to liquidation, debt should become zero

turadg:
closes: #5286

## Description

[5531](https://github.com/Agoric/agoric-sdk/issues/5531) specifies:
> vault manager only holds vaults that:
> - can accrue interest
> - can be liquidated

It implicates vault accounting because right now when a vault closes we update accounting in the manager. But this has a wrinkle because the vault’s idea of its debt and the manager’s are out of sync. The manager has already recorded the shortfall but the vault still holds onto that shortfall in its state as debt. ( When a vault is liquidated for less than its debt, the difference we call shortfall. We then burn the debt minus the shortfall. After this, the vault itself holds state of its debt being equal to shortfall. )

Liquidation means the vault manager took on the debt. It later hands any shortfall to the reserve but the vault can clean be cleaned up before that. This changes the `liquidate()` function on vaults to zero out the debt.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -585,6 +585,8 @@ const selfBehavior = {
   },
 
   /**
+   * Called by manager at start of liquidation.
+   *
    * @param {MethodContext} context
    */
   liquidating: ({ facets }) => {
@@ -594,14 +596,17 @@ const selfBehavior = {
   },
 
   /**
-   * Call must check for and remember shortfall
+   * Called by manager at end of liquidation, at which point all debts have been
+   * covered.
    *
    * @param {MethodContext} context
-   * @param {Amount} newDebt
    */
-  liquidated: ({ facets }, newDebt) => {
+  liquidated: ({ facets }) => {
     const { helper } = facets;
-    helper.updateDebtSnapshot(newDebt);
+    helper.updateDebtSnapshot(
+      // liquidated vaults retain no debt
+      AmountMath.makeEmpty(helper.debtBrand()),
+    );
 
     helper.assignPhase(Phase.LIQUIDATED);
     helper.updateUiState();
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -40,6 +40,7 @@ import {
   setUpZoeForTest,
   eventLoopIteration,
   withAmountUtils,
+  produceInstallations,
 } from '../supports.js';
 import { unsafeMakeBundleCache } from '../bundleTool.js';
 import {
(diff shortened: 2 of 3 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -597,11 +597,13 @@ const selfBehavior = {
    * Call must check for and remember shortfall
    *
```
Chris-Hibbert: This is no longer true, right?

Chris-Hibbert (approved):

packages/run-protocol/src/vaultFactory/vault.js
turadg: good catch, I'll revise

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -594,14 +596,17 @@ const selfBehavior = {
...
    * @param {MethodContext} context
-   * @param {Amount} newDebt
    */
-  liquidated: ({ facets }, newDebt) => {
+  liquidated: ({ state, facets }) => {
     const { helper } = facets;
-    helper.updateDebtSnapshot(newDebt);
+    helper.updateDebtSnapshot(
+      // liquidated vaults retain no debt
+      AmountMath.makeEmptyFromAmount(state.debtSnapshot),
```
dtribble: isn't the debtBrand well known in scope so we can just refer to it?  (minor)

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -635,7 +635,8 @@ test('price drop', async t => {
   const finalNotification = await E(vaultNotifier).getUpdateSince();
   t.is(finalNotification.value.vaultState, Phase.LIQUIDATED);
   t.deepEqual(finalNotification.value.locked, aeth.make(2n));
-  t.is(debtAmountAfter.value, 30n);
```
dtribble, line 638: We should add a check that the shortfall got recorded in the total shortfall on the vaultmanager, or some such.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -779,8 +780,8 @@ test('price falls precipitously', async t => {
   const debtAfterLiquidation = await E(vault).getCurrentDebt();
   t.deepEqual(
     debtAfterLiquidation,
-    run.make(103n),
```
dtribble, line 782: As above, make sure shortfall was recorded.

dtribble (approved):
A few requests for test improvements but I don't need to re-review those.

I built test-liquidation to be the additional tests for liquidation, but it should be expanded to check for the liquidation/shortfall amounts as well. Can you file a ticket for that?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: This was happening a few lines down so I just moved it up. Same for the other one.

turadg:
> I built test-liquidation to be the additional tests for liquidation, but it should be expanded to check for the liquidation/shortfall amounts as well. Can you file a ticket for that?

Filed as https://github.com/Agoric/agoric-sdk/issues/5559

Issue #5544, opened 2022-06-08
burnLosses may leave a staged allocation when offer safety is violated


Chris-Hibbert:
## Describe the bug

I wondered if there was a way for burnLosses to violate  offer safety, since it reaches directly into the allocation and removes some value. The answer is "no", since `burnLosses()` explicitly [checks offer safety](https://github.com/Agoric/agoric-sdk/blob/f18e90d8b64a0d43471f21cfb110c4767a18b5c6/packages/zoe/src/contractFacet/zcfZygote.js#L203) before calling reallocate.

Unfortunately, I noticed while reading the code that it stages a reallocation *before* checking for offer safety.

## To Reproduce

IOU a test case.

## Expected behavior

`burnLosses` should prevent violations of offer safety, and should also not leave staged allocations behind if it fails to make progress.

## Platform Environment

irrelevant. This was identified by reading the code.

## Additional context
None

## Screenshots
N/A.

PR #5560, opened 2022-06-09
test: demonstrate a bug caused by an offerSafety violation

Chris-Hibbert:
refs: #5544

## Description

This demonstrates the situation I reported in #5544. The question now
is whether it's a bug. It only happens when there's already a staged
allocation on a seat, and an attempt to burnLosses away from that seat
fails the offer safety check. In that case, the reallocation implied
by the burnLosses will have been included in the seat's allocation,
even though the burning wasn't done.

If this is indeed a bug, the fix is simply to move the offer safety
assertion before the conditional containing the decrementBy call.

### Security Considerations

Offer Safety is paramount.

No documentation changes desired unless this is not a bug.

### Testing Considerations

Adds a `.failing` test to show the behavior. If we decide it's a bug, we can mark it not failing when we fix it.


erights (approved):
Yes, this is a genuine bug. I don't review tests as carefully as code being tested. That said, LGTM. Thanks!

erights (changes requested):
The issue title is correct, but the PR title is not. It demonstrates a bug closely related to offer safety, but it is not an offer safety violation.

erights (approved):

PR #5561, opened 2022-06-10
build(deps): pin `@cosmjs/*` to `0.28.4`

michaelfig:
## Description

This PR prevents dragging in the `@cosmjs/*@0.28.6` release, which appears to be missing some pieces from the NPM registry.


mhofman (approved):
If that fixes it, great, but I still don't understand where it found those versions in the first place as they don't exist on NPM

michaelfig:
> If that fixes it, great, but I still don't understand where it found those versions in the first place as they don't exist on NPM

Apparently, `@cosmjs/stargate@0.28.6` exists and references the other (nonexistent) packages.  It appears that the local-npm test fetched that version, which caused the errors.

yarn.lock
```diff
@@ -1185,9 +1185,9 @@
     ws "^7"
     xstream "^11.14.0"
 
-"@cosmjs/stargate@^0.28.4":
+"@cosmjs/stargate@0.28.4":
   version "0.28.4"
-  resolved "https://registry.yarnpkg.com/@cosmjs/stargate/-/stargate-0.28.4.tgz#a5acbaa3451f7c853739064f799dec21097a06df"
+  resolved "http://localhost:4873/@cosmjs%2fstargate/-/stargate-0.28.4.tgz#a5acbaa3451f7c853739064f799dec21097a06df"
```
mhofman, line 1190: Shoot I missed this!

Issue #5557, opened 2022-06-09
swingset transcript.js returns bogus VatSyscallResults during replay for GC syscalls


warner:
When I fixed #5511 to record and return the full VatSyscallResult (both for errors and successes), I forgot to change `simulateSyscall`, which short-circuits the GC syscalls (`dropImports`, `retireImports`, `retireExports`). When it sees one of these, it 1: ignores the transcript, and 2: returns `undefined`.

That second part needs to change: `undefined` is not a valid `VatSyscallResult` (it is the data portion of `['ok', undefined]`, which *is* a valid VSR, and is the standard response to all GC syscalls, which don't have any other return value to offer).

@mhofman observed apparently non-fatal errors in the logs of a loadgen run which were caused by this bug. They were probably fatal to the `dispatch.bringOutYourDead` delivery happening at that time, but maybe because they were happening during replay, they didn't cause any significant problems beyond a different in logging.

I'm wondering if we should stop stripping the GC syscalls from the transcript. We did this when we wanted to tolerate different organic GC schedules between different vats (the "formal vs local GC" era). Given our need to have metering be identical, and our work to make GC consequences only observable during `bringOutYourDead`, I think at this point we could reasonably expect all GC activity to be in consensus, and thus it might no longer be necessary to ignore GC syscalls when replaying the transcript.

If we stopped ignoring them, this bug would probably have surfaced much faster: if the error caused a second GC syscall to not happen, we'd have gotten an `AnacrophobiaError` during replay.

We should also make sure that any errors during replay (non-`ok` VatDeliveryResults) cause a replay failure (and ultimately a kernel panic).

mhofman:
> We should also make sure that any errors during replay (non-`ok` VatDeliveryResults) cause a replay failure (and ultimately a kernel panic).

I suppose we wouldn't be replaying the delivery in the first place if it had previously failed (aka the vat had gotten terminated).

This occurred during the CI deployment test integration: https://github.com/Agoric/agoric-sdk/runs/6818734359?check_suite_focus=true#step:10:8734

<details>

<summary>Relevant logs with previous context showing it's when replaying after restart:</summary>

```
stage-1: chain: 2022-06-09T18:55:26.535Z launch-chain: buildSwingset
stage-1: chain: SwingSet xs-worker tracing: {
stage-1: chain:   XSNAP_TEST_RECORD: '/usr/src/agoric-sdk/chaintest/results/chain-stage-1-xsnap-trace'
stage-1: chain: }
[...]

Issue #5511, opened 2022-06-04
Vat transcript replay does not handle failed syscalls


mhofman:
## Describe the bug
While most syscalls are successful, `callNow` / `invoke` can explicitly fail with an `'error'` result. However the vat transcript replay always simulates an `'ok'` result.

https://github.com/Agoric/agoric-sdk/blob/2a537f7bd5ae7978248f68303be149835e8e1312/packages/SwingSet/src/kernel/vat-loader/manager-helper.js#L248-L250

I am also not sure what happens in the case of non-explicit error results, for example if an error is thrown during syscall processing, e.g. a `vatstoreGet`.

## To Reproduce
1. Make `invoke` syscall that result in an `'error'`.
2. Stop vat before making a snapshot
3. Reload vat and replay transcript
4. Observe anachrophobia and "historical inaccuracy in replay" error

## Expected behavior
Successful replay

## Additional context
This occurred during the deployment test of https://github.com/Agoric/agoric-sdk/pull/5385, e.g. https://github.com/Agoric/agoric-sdk/runs/6728088019?check_suite_focus=true

## Screenshots
Problematic syscall:
```
{"time":1654270285.0884733,"type":"deliver","crankNum":67,"vatID":"v2","deliveryNum":16,"replay":false,"kd":["message","ko26",{"methargs":{"body":"[\"createVatByName\",[\"chainStorage\"]]","slots":[]},"result":"kp50"}],"vd":["message","o+10",{"methargs":{"body":"[\"createVatByName\",[\"chainStorage\"]]","slots":[]},"result":"p-67"}]}
{"time":1654270285.091672,"type":"syscall","crankNum":67,"vatID":"v2","deliveryNum":16,"syscallNum":0,"replay":false,"ksc":["invoke","kd33","getNamedBundleCap",{"body":"[\"chainStorage\"]","slots":[]}],"vsc":["callNow","d-70","getNamedBundleCap",{"body":"[\"chainStorage\"]","slots":[]}]}
{"time":1654270285.093063,"type":"syscall-result","crankNum":67,"vatID":"v2","deliveryNum":16,"syscallNum":0,"replay":false,"ksr":["error","device.invoke failed, see logs for details"],"vsr":["error","device.invoke failed, see logs for details"]}
{"time":1654270285.0946267,"type":"console","source":"liveslots","state":"delivery","crankNum":67,"vatID":"v2","deliveryNum":16,"level":"info","args":["Logging sent error stack","(Error#1)"]}
[...]

PR #5563, opened 2022-06-10
fix(swingset): simulate full response for gc syscalls

mhofman:
refs: #5557

## Description

Fixes a transcript replay bug introduced in https://github.com/Agoric/agoric-sdk/issues/5511

### Security Considerations

None

None

### Testing Considerations

We detected this somewhat luckily. Plan is in a follow up PR to move gc calls in consensus, and enforce that all delivery results during replay are successful.


warner (approved):
If that passes CI, it should be good. I'll figure out how to write a proper test for it later.

mhofman:
@mergifyio refresh

PR #5568, opened 2022-06-10
fix: allow vault owner to only withdraw RUN

### LINKED ISSUE #5548: vault adjustBalances fails when only RUN is withdrawn
## Describe the bug

Vault holder should be able to withdraw RUN if there is sufficient collateral, but cannot unless they also give or take collateral.

## To Reproduce
1. Open a vault with ample collateral
2. Adjust balances that only withdraws run (no give)

## Expected behavior
Receive the RUN

## Actual behavior
Noop with offer result, 'no transaction, as requested'

Due to https://github.com/Agoric/agoric-sdk/blob/6b60e357cd047e3841a42365a3232df9ee2f1c65/packages/run-protocol/src/vaultFactory/vault.js

The behavior changed in https://github.com/Agoric/agoric-sdk/pull/5509 but it didn't work before that either. A couple days ago it would error with `Reallocate failed because a seat had no staged allocation. Please add or subtract from the seat and then reallocate.`

Chris-Hibbert:
closes: #5548

## Description

When the only change is withdrawing RUN, there's no change to the vaultSeat, so it has to be omitted from the list of seats sent to reallocateAndMint.

### Security Considerations

None

None

### Testing Considerations

Added a test.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -438,6 +438,12 @@ const helperBehavior = {
     const proposal = clientSeat.getProposal();
     assertOnlyKeys(proposal, ['Collateral', 'RUN']);
 
+    const allEmpty = amounts => {
+      // phrased negatively allows early return. This is equal to
+      // amounts.every(a => AmountMath.isEmpty(a));
+      return !amounts.some(a => !AmountMath.isEmpty(a));
+    };
+
     const normalizedDebtPre = self.getNormalizedDebt();
     const collateralPre = helper.getCollateralAllocated(vaultSeat);
 
@@ -464,7 +470,7 @@ const helperBehavior = {
       debt,
     );
     const wantRUN = proposal.want.RUN || helper.emptyDebt();
-    if ([giveColl, giveRUN, wantColl].every(a => AmountMath.isEmpty(a))) {
+    if (allEmpty([giveColl, giveRUN, wantColl, wantRUN])) {
       clientSeat.exit();
       return 'no transaction, as requested';
     }
@@ -489,7 +495,13 @@ const helperBehavior = {
     stageDelta(clientSeat, vaultSeat, giveColl, wantColl, 'Collateral');
     // `wantRUN` is allocated in the reallocate and mint operation, and so not here
     stageDelta(clientSeat, vaultSeat, giveRUN, helper.emptyDebt(), 'RUN');
-    state.manager.mintAndReallocate(toMint, fee, clientSeat, vaultSeat);
+
+    /** @type {Array<ZCFSeat>} */
+    const vaultSeatOpt = [];
+    if (!allEmpty([giveColl, giveRUN, wantColl])) {
+      vaultSeatOpt.push(vaultSeat);
+    }
+    state.manager.mintAndReallocate(toMint, fee, clientSeat, ...vaultSeatOpt);
 
     // parent needs to know about the change in debt
     helper.updateDebtAccounting(normalizedDebtPre, collateralPre, newDebt);
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -439,9 +439,7 @@ const helperBehavior = {
     assertOnlyKeys(proposal, ['Collateral', 'RUN']);
 
     const allEmpty = amounts => {
-      // phrased negatively allows early return. This is equal to
-      // amounts.every(a => AmountMath.isEmpty(a));
-      return !amounts.some(a => !AmountMath.isEmpty(a));
+      return amounts.every(a => AmountMath.isEmpty(a));
     };
 
     const normalizedDebtPre = self.getNormalizedDebt();
@@ -497,10 +495,9 @@ const helperBehavior = {
     stageDelta(clientSeat, vaultSeat, giveRUN, helper.emptyDebt(), 'RUN');
 
     /** @type {Array<ZCFSeat>} */
-    const vaultSeatOpt = [];
-    if (!allEmpty([giveColl, giveRUN, wantColl])) {
-      vaultSeatOpt.push(vaultSeat);
-    }
+    const vaultSeatOpt = allEmpty([giveColl, giveRUN, wantColl])
+      ? []
+      : [vaultSeat];
     state.manager.mintAndReallocate(toMint, fee, clientSeat, ...vaultSeatOpt);
 
     // parent needs to know about the change in debt
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -1303,6 +1303,103 @@ test('adjust balances', async t => {
   );
 });
 
+test('adjust balances - withdraw RUN', async t => {
+  const { zoe, aeth, run, rates } = t.context;
+  t.context;
+
+  const services = await setupServices(
+    t,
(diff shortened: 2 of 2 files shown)


packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -489,7 +495,13 @@ const helperBehavior = {
...
     // `wantRUN` is allocated in the reallocate and mint operation, and so not here
     stageDelta(clientSeat, vaultSeat, giveRUN, helper.emptyDebt(), 'RUN');
-    state.manager.mintAndReallocate(toMint, fee, clientSeat, vaultSeat);
+
+    /** @type {Array<ZCFSeat>} */
+    const vaultSeatOpt = [];
+    if (!allEmpty([giveColl, giveRUN, wantColl])) {
+      vaultSeatOpt.push(vaultSeat);
+    }
+    state.manager.mintAndReallocate(toMint, fee, clientSeat, ...vaultSeatOpt);
```
Chris-Hibbert: Is there a tighter way to do this?  I could do a ternary with two calls to `mintAndReallocate`, but I don't like that as much because of the code duplication.
```
  allEmpty([giveColl, giveRUN, wantColl])
    ? state.manager.mintAndReallocate(toMint, fee, clientSeat)
    : state.manager.mintAndReallocate(toMint, fee, clientSeat, vaultSeatOpt);
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -438,6 +438,12 @@ const helperBehavior = {
     const proposal = clientSeat.getProposal();
     assertOnlyKeys(proposal, ['Collateral', 'RUN']);
 
+    const allEmpty = amounts => {
+      // phrased negatively allows early return. This is equal to
```
turadg: good thinking about short circuit. fortunately the designers of [Array.prototype.every](https://tc39.es/ecma262/#sec-array.prototype.every) thought of that too.

so please just inline the `.every`

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -489,7 +495,13 @@ const helperBehavior = {
     stageDelta(clientSeat, vaultSeat, giveColl, wantColl, 'Collateral');
     // `wantRUN` is allocated in the reallocate and mint operation, and so not here
     stageDelta(clientSeat, vaultSeat, giveRUN, helper.emptyDebt(), 'RUN');
-    state.manager.mintAndReallocate(toMint, fee, clientSeat, vaultSeat);
+
+    /** @type {Array<ZCFSeat>} */
+    const vaultSeatOpt = [];
```
turadg: since it's just one, let's not complicate with an array
```
    /** @type {ZCFSeat?} */
    let vaultSeatOpt = undefined;
```
or to maintain the const,
```
    /** @type {ZCFSeat?} */
    const vaultSeatOpt = [giveColl, giveRUN, wantColl].every(a => a.isEmpty())
      ? undefined
      : vaultSeat;
```

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1303,6 +1303,103 @@ test('adjust balances', async t => {
...
+  t.deepEqual(debtAmount, runDebtLevel);
+
+  const { RUN: lentAmount2 } = await E(
+    aliceWithdrawRunSeat,
+  ).getCurrentAllocation();
+  const loanProceeds2 = await E(aliceWithdrawRunSeat).getPayouts();
+  t.deepEqual(lentAmount2, additionalRUN, '100 RUN');
+
+  const runLent2 = await loanProceeds2.RUN;
+  t.truthy(
```
turadg: `true` better than `truthy` here but why not use `deepEqual`? That would give a more helpful error message.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -1303,6 +1303,103 @@ test('adjust balances', async t => {
   );
 });
 
+test('adjust balances - withdraw RUN', async t => {
```
turadg, line 1307: much of what's being tested here is tested elsewhere. I don't suggest deleting but I suspect we'd benefit from being more systematic in what gets tested where and how.

packages/run-protocol/src/vaultFactory/vault.js
turadg: I can't tell whether you saw https://github.com/Agoric/agoric-sdk/pull/5568#discussion_r894836554 before this comment.

agree, better not to duplicate. I think the `vaultSeatOpt` you have is good. it's either a vault seat or undefined, which is the same as not passing it.
```
state.manager.mintAndReallocate(toMint, fee, clientSeat, vaultSeatOpt);
```

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: thanks!

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: I had not. Thanks for the suggestion. I had made an array when I thought I was passing two or three seats, and didn't reevaluate when it turned out to be one or two.

Chris-Hibbert: grr.  eslint doesn't like me passing `{ZcfSeat|undefined}` to a parameter declared as `{...ZcfSeat}`. What should I change?

...

We talked and agreed that 
```
   const vaultSeatOpt = allEmpty([giveColl, giveRUN, wantColl])
      ? []
      : [vaultSeat];
```
is a better choice, though that means the call to mintAndReallocate gets `...vaultSeatOpt`

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: Hmm.  From 1306 to 1363, it's setting up an overcollateralized loan. From that point on it's testing the new functionality of withdrawing RUN without changing collateral, which was previously untested.  I'll drop the assertions in the earlier section.

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: deepEqual is clearly better. Dunno how I ended up with this inferior sol'n

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: done. see below.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -439,9 +439,7 @@ const helperBehavior = {
     assertOnlyKeys(proposal, ['Collateral', 'RUN']);
 
     const allEmpty = amounts => {
-      // phrased negatively allows early return. This is equal to
-      // amounts.every(a => AmountMath.isEmpty(a));
-      return !amounts.some(a => !AmountMath.isEmpty(a));
+      return amounts.every(a => AmountMath.isEmpty(a));
```
turadg, line 457: I had proposed inlining this but I agree the helper is worth it,
```
    const vaultSeatOpt = allEmpty([giveColl, giveRUN, wantColl])
```
better than:
```
    const vaultSeatOpt = [giveColl, giveRUN, wantColl].every(a => AmountMath.isEmpty(a))
```

turadg (approved):

PR #5569, opened 2022-06-10
fix: patch mistaken overrides

erights:
Patch mistaken overrides.

These were the overrides that prevent us from changing https://github.com/Agoric/agoric-sdk/blob/09b239d797e67dde2a12f62f12ff154a1e402ab7/packages/agoric-cli/test/test-main.js#L4
to
```js
import '@endo/init/debug.js';
```
at 
https://github.com/Agoric/agoric-sdk/pull/5567
since `debug.js` also sets `overrideTaming` to `'min'` as it should, in order to catch such problems.

### DIFF
--- a/patches/bl++readable-stream+3.6.0.patch
+++ b/patches/bl++readable-stream+3.6.0.patch
@@ -1,0 +1,20 @@
+diff --git a/node_modules/bl/node_modules/readable-stream/errors.js b/node_modules/bl/node_modules/readable-stream/errors.js
+index 8471526..c1c08cf 100644
+--- a/node_modules/bl/node_modules/readable-stream/errors.js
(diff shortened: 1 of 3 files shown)


turadg (approved):
Why does `enumerable: true` cause the problem?

Approving since I assume you have a good answer : )

erights:
> Why does `enumerable: true` cause the problem?
> 
> Approving since I assume you have a good answer : )

Clarified by adding

```
+            // enumerable: true would accurately preserve the behavior of the
+            // original assignment, but I'm guessing that was not intentional.
+            // For an actual error subclass, this property would not
+            // be enumerable.
```
to all the relevant `enumerable: false` cases.

PR #5572, opened 2022-06-10
fix vault manager totalDebt metric

### LINKED ISSUE #5550: vault manager totalDebt metric does not return to zero when all vaults are closed/liquidated
## Describe the bug

vault manager totalDebt metric does not always return to zero when all vaults are closed/liquidated.

## To Reproduce
1. Open new vaults in a manager
2. Accrue interest that results in a shortfall and triggers liquidation
4. see that totalDebt is not zero

## Expected behavior
totalDebt is a metric of the debts in the store. It should always be the sum of the held vaults' actual debts. It could be inferred from adding that for all the vaults but for performance we track it as a scalar that moves with operations that change the vaults.

turadg:
closes: #5550
stacks on: #5531

## Description

The `totalDebt` tracker wasn't accounting for shortfalls. While debugging I also encountered off-by-one rounding errors with apparently inconsistent of ceil/floor, so this also changes to use bankers' rounding in interest calculations.

### Security Considerations

No changes here but noted some opportunities to reduce powers of the `liquidate` module.

--

### Testing Considerations

The metrics tests would take less effort to read and maintain if they used a driver. Should that be done in this PR? (e.g. moving into a new file modeled after `test-liquidator.js`.

### DIFF
--- a/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
+++ b/packages/run-protocol/test/vaultFactory/test-vaultFactory.js
@@ -853,20 +853,18 @@ test('vaultFactory display collateral', async t => {
   });
 });
 
-// charging period is 1 week. Clock ticks by days
-test.only('interest on multiple vaults', async t => {
+test('interest on multiple vaults', async t => {
   const { zoe, aeth, run, rates: defaultRates } = t.context;
   const rates = {
     ...defaultRates,
     interestRate: makeRatio(5n, run.brand),
   };
   t.context.rates = rates;
+  // charging period is 1 week. Clock ticks by days
   t.context.loanTiming = {
     chargingPeriod: SECONDS_PER_WEEK,
     recordingPeriod: SECONDS_PER_WEEK,
   };
-
-  // Clock ticks by days
   const manualTimer = buildManualTimer(t.log, 0n, SECONDS_PER_DAY);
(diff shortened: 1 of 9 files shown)


packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2507,26 +2505,27 @@ test('manager notifiers', async t => {
       Collateral: t.context.aeth.mint.mintPayment(aeth.make(AMPLE)),
     }),
   );
-  const { vault: vault1 } = await E(vaultSeat).getOfferResult();
+  let { vault } = await E(vaultSeat).getOfferResult();
```
Chris-Hibbert, line 2508: I'm not a fan of re-using variables in this way, but if that's the way this test goes, please update vault2 (line 2726) to be part of it.

Chris-Hibbert (approved):

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
turadg: Not my favorite either. I think this would be better with a driver but that's more work that I can't justify. At one point I had them numbered, but that made it harder to move things around. 

(done as a fixup and rebase)

PR #5574, opened 2022-06-10
vault manager refactorings

turadg:
refs: #4415

## Description

Code improvements that came up while working on bug fixes. Left out of those PRs since they're out of scope of fixing the bug and may require additional review.

Review by commit recommended.

### Security Considerations

Take `burnLosses` power from `vault`. Gives `vaultManager` the `vault` object instead, but it already had the power to iterate over all unsettled vaults.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/liquidation.js
+++ b/packages/run-protocol/src/vaultFactory/liquidation.js
@@ -18,9 +18,6 @@ const trace = makeTracer('LIQ', false);
  *
  * @param {ZCF} zcf
  * @param {Vault} vault
- * @param {(losses: Amount,
- *             zcfSeat: ZCFSeat
- *            ) => void} burnLosses
  * @param {Liquidator}  liquidator
  * @param {Brand} collateralBrand
  * @param {Ratio} penaltyRate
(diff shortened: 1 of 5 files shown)


packages/run-protocol/src/vaultFactory/liquidation.js
```diff
@@ -80,16 +76,13 @@ const liquidate = async (
   const runToBurn = AmountMath.min(proceeds.RUN, debt);
   // debt is fully settled, with runToBurn and shortfall
   assert(AmountMath.isEqual(debt, AmountMath.add(runToBurn, shortfall)));
-  trace('before burn', { debt, proceeds, overage, shortfall, runToBurn });
-  // TODO why grant this power; we can burn it from the caller
-  burnLosses(runToBurn, vaultZcfSeat);
```
turadg, line 85: Now vault is updated with `liquidated` before the losses are burned, which updates the consumer of the vault. If `burnLosses` fails now the vault will still be "liquidated" to the user.

As I think this through I realize this could allow a state in which a vault's phase is `liquidated` and thus settled, but it's still in the vaultManager. Though that possibility true of master, albeit with one fewer case (failure of `burnLosses`/`burnDebt`) that could manifest it. We don't have checked exceptions but maybe there's something to document about what failures must be handled? cc @dtribble

Chris-Hibbert (approved):

PR #5577, opened 2022-06-11
fix: minor tweak to make `sortObjectProperties` use `hasOwn`

FUDCo:
Yesterday's PR #5552 originally used `Object.hasOwn` but the XS engine lacked the `Object.hasOwn` function and so all the XS-based tests failed, so the code was tweaked to do something lesser.  But today a new version of XS appeared that does have `hasOwn` so this small PR puts things to right.

### DIFF
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -271,7 +271,7 @@ function sortObjectProperties(obj, firsts = []) {
   const sorted = [...firsts, ...Object.keys(obj).sort()];
   const result = {};
   for (const prop of sorted) {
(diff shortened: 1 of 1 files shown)


warner:
We think that the real issue here was whether Node.js has `Object.hasOwn`, not XS, because this code runs in the controller, which runs under XS. So we *can* land this if we require a sufficiently new Node.js version.

* added in Node 16.9
* we dropped support for Node 14
* we use/require Node LTS, i.e. Node 16.x through 16.max , where "x" is 13 .

This is trivial and low-priority. If @FUDCo gets time, he'll refresh and land this.

FUDCo:
@warner I was going through old dangling stuff and came across this one.  I think this is now safe to land and we should do so.

warner (approved):
LGTM

PR #5579, opened 2022-06-12
feat: SwingSet client input limits

### LINKED ISSUE #4277: Creating a large string in the REPL crashes v2 http vat
## Describe the bug

When interacting with the repl interface, it is possible to force the http vat to crash due to the Allocate meter tripping 

## To Reproduce
1. start an agoric chain with `agoric start --reset`
2. open the repl interface `agoric open --repl`
3. wait for the chain to start
4. issue the command `"A".repeat(500000000)`
5. `2022-01-11T21:08:46.818Z SwingSet: kernel: delivery problem, terminating vat v2 Allocate meter exceeded`

## Expected behavior
An input in the repl shouldn't be able to crash a core vat

## Platform Environment
 macOS 12.1, node v16.13.1
 I only have 16gb of ram

ref #4264

### LINKED ISSUE #4283: User sending a large string breaks their Ag Solo at the cosmos layer
## Describe the bug

It appears that some of the connective tissue between the cosmos layer and the JS layer where json message length is treated differently, resulting in a valid JS message, but when parsing it on the golang layer it is not.  The reproduction steps below crash the local node, but all vats continue to run normally (including http)

## To Reproduce
3. wait for chain to boot
4. Issue `E(home.zoe).getInstallation("A".repeat(10000000))`
5. The node will loop in a crashed state

```
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'... 9990044 more characters                 
	]
] 197                                                                                                                           

2022-01-12T03:55:15.209Z chain-cosmos-sdk: Error: RPC error -32600 - Invalid Request: error reading request body: http: request body too large
Usage:                                                                                                                                                         agd tx swingset deliver [json string] [flags]
Flags:
-a, --account-number uint      The account number of the signing account (offline mode only)
-b, --broadcast-mode string    Transaction broadcasting mode (sync|async|block) (default "sync")
--dry-run                  ignore the --gas flag and perform a simulation of a transaction, but don't broadcast it
--fee-account string       Fee account pays fees for the transaction instead of deducting from the signer
--fees string              Fees to pay along with transaction; eg: 10uatom
--from string              Name or address of private key with which to sign
--gas string               gas limit to set per-transaction; set to "auto" to calculate sufficient gas automatically (default 200000)
--gas-adjustment float     adjustment factor to be multiplied against the estimate returned by the tx simulation; if the gas limit is set manually this flag is ignored  (default 1)
--gas-prices string        Gas prices in decimal format to determine the transaction fee (e.g. 0.1uatom)
--generate-only            Build an unsigned transaction and write it to STDOUT (when enabled, the local Keybase is not accessible)
-h, --help                     help for deliver         

2022-01-12T04:01:48.527Z chain-cosmos-sdk: Error sending (SyntaxError#5)
SyntaxError#5: Unexpected end of JSON input

  at JSON.parse (<anonymous>)
  at sendFromMessagePool (packages/solo/src/chain-cosmos-sdk.js:692:28)
  at async recurseEachSend (packages/solo/src/chain-cosmos-sdk.js:788:5)
                                                                       
```

## Expected behavior

All valid JS that works normally within the XS/repl layer should be also equally valid when traveling through the cosmos layer

## Platform Environment
 macOS 12.1, node v16.13.1
 I only have 16gb of ram

ref #4264

michaelfig:
closes: #4283, closes: #4277

## Description

This applies input limit management in several of the SwingSet inputs:
- marshalled messages to/from liveslots on all workers (parameters persisted during vat initialization)
- `bridge` and `command` device inputs
- chain and sim-chain action validation using consensus parameters before committing to action queue

The marshalling input and output limits require a patch to Endo, and thus this PR will not come out of draft until that one (or a replacement) lands: endojs/endo#1215.

### Demo

Here's an example of running from the REPL.  The `TypeError: target has no method...` indicate when the message got through successfully.  The other errors show when they were rejected in transit before arriving at `home.scratch`'s vat.

image

### Security Considerations

Considerably improves robustness by filtering out unusually-large inputs before they get to the SwingSet kernel.  Positions us to be able to reduce the size of the filtered data after bundle caps are in full force.

Defensive validation of inputs may still be necessary at the vat level, depending on how they are used.

Almost no need for documentation; these are limits that produce readable exceptions when they are exceeded.

### Testing Considerations

Tests still need to be written.  This is a POC spike.


packages/SwingSet/src/limits/calculate.js
```diff
@@ -0,0 +1,193 @@
...
+
+  let totalCost = 0n;
+  const getTotalCost = () => totalCost;
+
+  /**
+   * @param {unknown} value
+   */
+  const shallowCost = value => {
+    let cost = baseCost;
+    if (Object(value) === value) {
```
dckc: This technique for testing object-ness allocates. We avoid it in at least one other case.

packages/SwingSet/src/limits/calculate.js
```diff
@@ -0,0 +1,193 @@
...
+      }
+      case 'bigint': {
+        // Compute the number of approximate bytes in the bigint.
+        let remaining = value;
+        if (remaining < 0n) {
+          remaining = -remaining;
+        }
+        // Count number of bigint chunk and multiply out.
+        while (remaining > 0n) {
+          remaining /= BIGINT_WORD_VALUE;
```
dckc: Dividing is expensive. How about pre-computing the max?

dckc: Oh... This is a cost, not a threshold. Still... A table of word-sizes could be pre-computed.

packages/SwingSet/src/limits/calculate.js
michaelfig: Thanks, that's good to know!

michaelfig: I'll use `@endo/marshal`'s `isObject`, and punt to their implementation for any future optimisations.

packages/SwingSet/src/limits/calculate.js
michaelfig: I can make it way more efficient by stringifying the bigint then counting the digits.  So I will.

packages/SwingSet/src/limits/budgets.js
```diff
@@ -0,0 +1,90 @@
...
+export const SOURCE_BUNDLE_BUDGET = registerBudget('SOURCE_BUNDLE_BUDGET', {
+  description: 'source bundle',
+  maximumPropertyCost: 32,
+  maximumTotalCost: MAX_SOURCE_BUNDLE_BYTES,
+  maximumValueCost: MAX_SOURCE_BUNDLE_BYTES,
+  maximumBigIntDigits: MAX_BIGINT_DIGITS,
+});
+
+/** @type {Budget} */
+export const VAT_MESSAGE_BUDGET = registerBudget('VAT_MESSAGE_BUDGET', {
```
michaelfig, line 52: @FUDCo Here is the default-default-default VAT_MESSAGE_BUDGET.

michaelfig:
Closing, as this might be useful as a reference but will not be adopted verbatim into SwingSet.

PR #5581, opened 2022-06-13
feat(wallet-ui): prepare for Smart Wallet connection

michaelfig:
## Description

- Much more robust connection/disconnection implementation.
- Code-splitting for different wallet backends.
- Persistent connection setup page

https://user-images.githubusercontent.com/457244/173410533-b59fd66c-8b50-4e57-a272-710f36341584.mov


packages/wallet/ui/src/components/AppBar.jsx
```diff
@@ -47,9 +64,47 @@ const useStyles = makeStyles(theme => ({
...
+export const AppBarWithoutContext = ({
+  connectionComponent,
+  connectionState,
+  wantConnection,
+  setWantConnection,
+}) => {
   const theme = useTheme();
   const classes = useStyles(theme);
+  const [status, setStatus] = useState(
+    wantConnection ? 'connecting' : 'disconnected',
```
turadg: please define an enum-like object for these strings, to make explicit what the domain of `status` is

packages/wallet/ui/src/components/AppBar.jsx
```diff
@@ -47,9 +64,47 @@ const useStyles = makeStyles(theme => ({
...
+      case 'connecting': {
+        setStatus('connecting');
+        break;
+      }
+      default:
+    }
+  }, [connectionState, connectionComponent]);
+
+  const connectionTitle = status[0].toUpperCase() + status.slice(1);
+  const connectionClassName = status === 'connecting' ? classes.connecting : '';
```
turadg: no need to bind this. more idiomatic to inline,
```
status === 'connecting' && classes.connecting
```

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+    textAlign: 'center',
+  },
+  dialog: {
+    minWidth: 240,
+  },
+}));
+
+const ConnectionSettingsDialog = ({
+  onClose,
+  open,
```
turadg, line 23: I was going to ask that this be `isOpen` to make clear it's declarative and not imperative but I see MUI's `Dialog` uses `open` so agree it's better to be consistent

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const saveAndClose = () => {
+    if (connection) {
+      setWalletConnection(connection);
+      disconnect();
+      if (
+        !allWalletConnections.some(
+          c => c.label === connection.label && c.url === connection.url,
+        )
+      ) {
```
turadg: took me a bit to parse this. consider naming the expression,
```suggestion
      const connectionKnown = allWalletConnections.some(
      …
      if (!connectionKnown) {
```

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+                });
+              } else if (newValue && newValue.inputValue) {
+                setConnection({
+                  url: newValue.inputValue,
+                  label: newValue.inputValue,
+                });
+              } else {
+                setConnection(newValue);
+              }
+            }}
```
turadg: this transformation seems worth a helper.
```suggestion
            onChange={(_, newValue) => setConnection(connectionValue(newValue))}
```

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+            filterOptions={(options, params) => {
+              // eslint-disable-next-line no-unused-vars
```
turadg: stale suppression? `inputValue` is used.

packages/wallet/ui/src/components/Loading.jsx
```diff
@@ -0,0 +1,47 @@
+import Typography from '@mui/material/Typography';
+import CircularProgress from '@mui/material/CircularProgress';
+import Paper from '@mui/material/Paper';
+
+import { withApplicationContext } from '../contexts/Application';
+
+const Loading = ({ wantConnection, walletConnection, ...rest }) => {
+  let body;
```
turadg: I try to minimize mutations in React and would have a `LoadingInner` component instead of assigning `body`, but not requesting it

packages/wallet/ui/src/components/Loading.jsx
```diff
@@ -0,0 +1,47 @@
...
+  if (walletConnection) {
+    const { url } = walletConnection || {};
```
turadg: this is guaranteed truthy by line 9
```suggestion
    const { url } = walletConnection;
```

packages/wallet/ui/src/components/SmartWalletConnection.jsx
```diff
@@ -0,0 +1,12 @@
+const SmartWalletConnection = () => 'FIXME: SmartWalletConnection';
+
+export default SmartWalletConnection;
+
+export const buildWalletConnection = withApplicationContext =>
```
turadg: why pass in `withApplicationContext`? isn't it the same for the whole application?

packages/wallet/ui/src/components/WalletConnection.jsx
```diff
@@ -71,76 +62,70 @@ const getAccessToken = () => {
...
-          <Public fontSize="inherit">{connectionState}</Public>
-        </IconButton>
-      </Tooltip>
-      <AgoricWalletConnection
-        onState={onWalletState}
-        className={classes.hidden}
-      />
-    </div>
+    <AgoricWalletConnection
+      onState={onWalletState}
```
turadg, line 120: not set in this PR but `onState` is a confusing property name since `state` has a quite specific meaning in React

packages/wallet/ui/src/contexts/Application.jsx
```diff
@@ -9,8 +11,38 @@ import {
   useReducer,
 } from 'react';
 
+// TODO: Graduate to show mainnet in production.
+const SHOW_MAINNET = process.env.NODE_ENV === 'development';
+export const DEFAULT_WALLET_CONNECTIONS = [
```
turadg: I'm impressed at the non-imperative here, but it is pretty hard to read. A quick fix would be to conditionally push onto the array. Another would be to name values then filter out empties,
```
const localConnection = window && window.location.hostname !== 'localhost'
  ? { label: window.location.hostname, url: window.location.origin }
  : { label: 'Local Solo', url: 'http://localhost:8000' }]
const stageConnection = process.env.NODE_ENV === 'development' && {
          label: 'Agoric Stage',
          url: 'https://stage.agoric.net/network-config',
        };
  …
export const DEFAULT_WALLET_CONNECTIONS = [
  LOCAL_CONNECTION,
  NET_CONNECTION,
  STAGE_CONNECTION,
  ].filter(Boolean);
```

packages/wallet/ui/src/contexts/Application.jsx
```diff
@@ -9,8 +11,38 @@ import {
...
+          label: 'Agoric Devnet',
+          url: 'https://devnet.agoric.net/network-config',
+        },
+      ]
+    : []),
+];
+
 export const ApplicationContext = createContext();
 
+export const DEFAULT_WALLET_CONNECTION = DEFAULT_WALLET_CONNECTIONS[0];
```
turadg: these two very similar names seem like a smell. also is this export used anywhere?

packages/wallet/ui/src/contexts/Application.jsx
```diff
@@ -133,8 +170,97 @@ const Provider = ({ children }) => {
...
+  const [walletConnection, setWalletConnection] = useState(
+    restoredWalletConnection,
+  );
+  const [allWalletConnections, setAllWalletConnections] = useState(
+    harden([...DEFAULT_WALLET_CONNECTIONS]),
+  );
+
+  useEffect(() => {
+    if (!connectionComponent) {
+      if (connectionState === 'error') {
```
turadg: please use an enum-like object to make explicit the domain of this value

packages/wallet/ui/src/contexts/Application.jsx
```diff
@@ -133,8 +170,97 @@ const Provider = ({ children }) => {
...
+      case 'connecting': {
+        setConnectionStatus('connecting');
+        break;
+      }
+      default:
+    }
+  }, [connectionState, connectionComponent]);
+
+  useEffect(() => {
+    if (window && window.localStorage) {
```
turadg: worth a helper,
```
maybeStore('walletConnection', walletConnection);
```

packages/wallet/ui/src/util/fixed-websocket-connector.js
```diff
@@ -0,0 +1,87 @@
...
+import { E } from '@endo/eventual-send';
+
+const CONNECTION_TIMEOUT_MS = 5000;
+
+// Wait for the wallet to finish loading.
+export const waitForBootstrap = async getBootstrap => {
+  const getLoadingUpdate = (...args) =>
+    E(E.get(getBootstrap()).loadingNotifier).getUpdateSince(...args);
+  let update = await getLoadingUpdate();
+  while (update.value.includes('wallet')) {
```
turadg, line 11: what do these values look like?

packages/wallet/ui/src/util/fixed-websocket-connector.js
```diff
@@ -0,0 +1,87 @@
...
+
+    const adminFacet = E(
+      E.get(waitForBootstrap(getBootstrap)).wallet,
+    ).getAdminFacet();
+    adminFacet.then(component.service.send({ type: 'connected' }));
+    component._bridgePK.resolve(adminFacet);
+  };
+
+  return {
+    render: () => '',
```
turadg, line 48: no need to be a string
```suggestion
    render: () => null,
```

But why is this a React component at all if it never renders?

turadg (commented):
Took a pass. No formal request changes since it's draft

packages/wallet/ui/src/components/SmartWalletConnection.jsx
michaelfig: Ah, it would be a circular reference to `Application.jsx` (which dynamically imports this module).  Circular references are not allowed across WebPack code-splitting boundaries, delineated by calls to the `import(...)` expression.

turadg: maybe `withApplicationContext` shoudn't be part of `Application.jsx`?

it's really not a context itself. I'd put it in `utils/withApplicationContext.js`. would that solve the circularity?

packages/wallet/ui/src/util/fixed-websocket-connector.js
michaelfig: `['agoric', 'wallet', ...]`

packages/wallet/ui/src/components/SmartWalletConnection.jsx
michaelfig: That's a fantastic idea!  I will take that on.

michaelfig: Thanks, that's perfect!

samsiegart (approved):

PR #5584, opened 2022-06-13
handle floating promises in governance

turadg:
no ticket, busy hands during standup

## Description

Handle floating promises.

### Security Considerations

--

--

### Testing Considerations

CI

### DIFF
--- a/packages/governance/src/closingRule.js
+++ b/packages/governance/src/closingRule.js
@@ -10,7 +10,7 @@ import { Far } from '@endo/marshal';
 /** @type {CloseVoting} */
 export const scheduleClose = (closingRule, closeVoting) => {
   const { timer, deadline } = closingRule;
-  E(timer).setWakeup(
+  void E(timer).setWakeup(
     deadline,
     Far('close voting', {
       wake: closeVoting,
--- a/packages/governance/test/swingsetTests/committeeBinary/bootstrap.js
+++ b/packages/governance/test/swingsetTests/committeeBinary/bootstrap.js
@@ -228,10 +228,16 @@ const makeBootstrap = (argv, cb, vatPowers) => async (vats, devices) => {
   const [testName] = argv;
   switch (testName) {
     case 'committeeBinaryStart':
-      committeeBinaryStart(zoe, voterCreator, timer, log, installations);
+      await committeeBinaryStart(zoe, voterCreator, timer, log, installations);
       break;
     case 'committeeBinaryTwoQuestions':
-      committeeBinaryTwoQuestions(zoe, voterCreator, timer, log, installations);
+      await committeeBinaryTwoQuestions(
+        zoe,
+        voterCreator,
+        timer,
+        log,
+        installations,
+      );
       break;
     default:
       log(`didn't find test: ${argv}`);
--- a/packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
+++ b/packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
@@ -35,8 +35,8 @@ const build = async (log, zoe) => {
       const electoratePublicFacet = E(zoe).getPublicFacet(electorateInstance);
       const seat = E(zoe).offer(invitation);
       const voteFacet = E(seat).getOfferResult();
-      E.when(E(seat).getPayouts(), async () => {
-        E.when(E(seat).hasExited(), exited => {
+      void E.when(E(seat).getPayouts(), async () => {
+        void E.when(E(seat).hasExited(), exited => {
           log(`Seat ${name} ${exited ? 'has exited' : 'is open'}`);
         });
       });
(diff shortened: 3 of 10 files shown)


packages/governance/src/closingRule.js
```diff
@@ -10,7 +10,7 @@ import { Far } from '@endo/marshal';
 /** @type {CloseVoting} */
 export const scheduleClose = (closingRule, closeVoting) => {
   const { timer, deadline } = closingRule;
-  E(timer).setWakeup(
+  return E(timer).setWakeup(
```
Chris-Hibbert:
```suggestion
  void E(timer).setWakeup(
```

`CloseVoting` claims to have return type `void`

packages/governance/test/swingsetTests/committeeBinary/bootstrap.js
```diff
@@ -228,10 +228,16 @@ const makeBootstrap = (argv, cb, vatPowers) => async (vats, devices) => {
   const [testName] = argv;
   switch (testName) {
     case 'committeeBinaryStart':
-      committeeBinaryStart(zoe, voterCreator, timer, log, installations);
+      await committeeBinaryStart(zoe, voterCreator, timer, log, installations);
```
Chris-Hibbert, line 231: I don't think `await` is doing anything helpful here. Wouldn't `void` be clearer?
[same below]

packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
```diff
@@ -43,7 +43,7 @@ const build = async (log, zoe) => {
         },
       });
       const subscription = E(electoratePublicFacet).getQuestionSubscription();
-      observeIteration(subscription, votingObserver);
+      await observeIteration(subscription, votingObserver);
```
Chris-Hibbert: Similarly here. Isn't `await` misleading? The point of the code is to set the observer off on it's own, and return the facet. Why is `await` better?

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
```diff
@@ -390,7 +390,7 @@ const makeBootstrap = (argv, cb, vatPowers) => async (vats, devices) => {
 
       await E(timer).tick();
       await E(timer).tick();
-      E.when(
+      await E.when(
```
Chris-Hibbert:
```suggestion
      void E.when(
```

The test doesn't want to wait here.  It wants these logs requests to fire later.

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
```diff
@@ -443,7 +443,7 @@ const makeBootstrap = (argv, cb, vatPowers) => async (vats, devices) => {
...
```
Chris-Hibbert:
```suggestion
      void E.when(
```

packages/governance/test/swingsetTests/committeeBinary/bootstrap.js
turadg: It's preserving the stack trace if `committeeBinaryStart` fails. Without these awaits, `makeBootstrap` will complete as soon as the case functions are called. If they fail it will throw in Node as an unhandled promise rejection.

I think if they fail then `makeBootstrap` should also fail.

packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
turadg: Yeah, I was on the fence about this case. The point is to kick off the observer and not expect anything back, but what if it fails? Like above, awaiting maintains the stack trace. 

If we don't want to detect that failure, then I think we should change `observeIteration` (the called function) to be synchronous instead of dropping the promise in every call.

From https://github.com/Agoric/agoric-sdk/blob/cc522ee7bbf53c74224e4d188336233dc7225cb5/packages/notifier/src/asyncIterableAdaptor.js#L135-L140 to 
```js
 */
export const observeIteration = (asyncIterableP, iterationObserver) => {
  const iteratorP = E(asyncIterableP)[Symbol.asyncIterator]();
  void observeIterator(iteratorP, iterationObserver);
};
```

Which do you think is better?

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
turadg: thanks, I was wondering why they were failing ;-)

packages/governance/test/swingsetTests/committeeBinary/vat-voter.js
Chris-Hibbert: Attempting to channel @erights: I expect some callers to want to be able to react to the results. In this test, it seems that the test should be conditional on results of the observation, so it doesn't really need to detect a failure in the adapter.

Chris-Hibbert (approved):

PR #5585, opened 2022-06-13
Migrate entire SDK from RUN to Inter (IST)

turadg:
closes: #4800
part of: #5326

## Description

Per https://github.com/Agoric/agoric-sdk/issues/4800:
- make vaultFactory and runStake use `Minted` instead of `RUN`
- make feeDistributor and assetReserve use `Fee` instead of `RUN`

This PR has been repurposed to do a "big bang" migration from RUN to Inter.  Here are the gates we need to pass before that can land:

- [x] This base PR is approved (@Chris-Hibbert)
- [x] https://github.com/Agoric/agoric-sdk/pull/5802
- [x] Bootstrap changes from RUN to IST https://github.com/Agoric/agoric-sdk/pull/5332
  - [x] Cosmos-level denom changes from `urun` to `uist`
- [x] Load generator changes to accommodate new keywords/tokens https://github.com/Agoric/testnet-load-generator/pull/80/ (@mhofman)
- [x] make getting-started integration pass against this branch (if necessary)

After this PR:
- [ ] Any necessary manual dapp testing in #5591 has been performed (@samsiegart)
- [ ] Documentation has been updated (@dckc)
- [ ] [run-protocol package](https://www.npmjs.com/package/@agoric/run-protocol) is deprecated and points to inter
- [ ] no results for `git grep "RUN [pP]rotocol"` 

The intention is for green CI to indicate that this job was done correctly.

## References in other repositories

If you are mentioned in a given line below, see <a href="#user-content-updating-references">instructions for updating references</a>.

DO NOT remove your mention in case problems are found that need your attention.  If you have done this correctly, somebody will merge your PR when the big bang happens.

#dapp-fungible-faucet-branch main @turadg 
#dapp-card-store-branch main @turadg
#dapp-oracle-branch: main @turadg
#dapp-otc-branch: main @turadg 
#dapp-pegasus-branch: main @turadg 
#dapp-simple-exchange-branch: main @turadg 
#dapp-treasury-branch: main @turadg 
#documentation-branch: main @turadg
#loadgen-branch: main Agoric/testnet-load-generator#80 @mhofman

### <a id="updating-references">Updating References</a>

You can edit the above lines as follows:

- if you have verified that the appropriate repo does not require any changes
  - Set `TODO-BRANCH` to either `master` or `main` as applicable
  - Set `PRNUM` to `n/a`
- if you need to create a new PR in that repo to enact changes
  - Ensure your PR contains the following in its description and wait until it passes its own CI:
[...]

PR #5595, opened 2022-06-14
4800 rename run-protocol package to inter-protocol

turadg:
part of: #4800 

## Description

Single commit for review to include in the mega https://github.com/Agoric/agoric-sdk/pull/5585

### Security Considerations

no change

Do we need a package redirect notice? I figure SDK is understood to be in flux, and the `agoric` command works as well.

### Testing Considerations

--


packages/cosmic-swingset/Makefile
```diff
@@ -49,7 +49,7 @@ $(EVAL_CLEAN): $(EVAL_CODE)
 	./scripts/clean-core-eval.js $< >$@.t || { rm -f $@.t; exit 1; }
 	mv $@.t $@
 
-RUN_PROTO=../run-protocol
+RUN_PROTO=../inter-protocol
```
Chris-Hibbert: Why leave RUN_PROTO behind?

packages/cosmic-swingset/Makefile
```diff
@@ -71,8 +71,8 @@ gov-q:
 	$(AGCH) query gov proposals --output json | \
 		jq -c '.proposals[] | [.proposal_id,.voting_end_time,.status]'
 
-deploy-runStake: ../run-protocol/test/runStake/runStake-deploy.js
-	echo agoric deploy ../run-protocol/test/runStake/runStake-deploy.js
+deploy-runStake: ../inter-protocol/test/runStake/runStake-deploy.js
```
Chris-Hibbert: Is runStake renaming a separate pass?

Chris-Hibbert (commented):
search for "RUN [pP]rotocol" in .../governance and .../inter-protocol, ignoring CHANGELOG files.

packages/cosmic-swingset/Makefile
turadg: not intentional. this was a find-replace and I haven't looked over all the instances yet. I figured I would keep putting commits into #5585 but I'll get more into this one and mark it as Draft until I think it's ready to merge.

packages/cosmic-swingset/Makefile
turadg: yep, needed to get alignment on the new name

Issue #5571, opened 2022-06-10
Under load, brands do not match


arirubinstein:
## Describe the bug
```
2022-06-10T18:38:43.026Z SwingSet: ls: v28: RangeError#266: Brands in left Object [Alleged: BLDPoolLiquidity brand] {} and right Object [Alleged: BLDPoolLiquidity brand] {} should match but do not
2022-06-10T18:38:43.027Z SwingSet: ls: v28: RangeError: Brands in left (an object) and right (an object) should match but do not
 at makeError (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:2705)
 at fail (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:2832)
 at baseAssert (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:2850)
 at equal (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:2861)
 at checkLRAndGetHelpers (.../ertp/src/amountMath.js:164)
 at isGTE (.../ertp/src/amountMath.js:303)
 at handleAddPoolOffer (.../run-protocol/src/vpool-xyk-amm/addPool.js:145)
 at handleAddPoolOffer ()
 at apply ()
 at apply ()
 at dispatchToHandler (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:8729)
 at (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:8362)
 at win (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:9072)
 at ()
```

slog at: 
https://drive.google.com/file/d/1FIg_y0Tp1PsE438LtKBxhp_ki1yqfjEF/view?usp=sharing

warner:
The cursed delivery was a `handleOffer()` on deliveryNum 679, and the marshalled arguments were:

```json
[
  {
    "@qclass": "slot",
    "iface": "Alleged: InvitationHandle",
    "index": 0
  },
  {
[...]

PR #5596, opened 2022-06-15
fix: handle parallel requests to add an issuer in AMM

Chris-Hibbert:
refs: #5571

## Description

There are two bugs in #5571, and this fixes only one of them.

### Security Considerations

The bug implies that we haven't been careful enough about races in requests from users.

None

### Testing Considerations

The test exercises each of the branches in addIssuer.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -80,9 +80,14 @@ export const makeAddIssuer = (
         brandToLiquidityMint.init(secondaryBrand, mint);
         const { issuer: liquidityIssuer } = mint.getIssuerRecord();
         liquidityPromiseKit.resolve(liquidityIssuer);
+        // we only need entries in this table until brandToLiquidityIssuer knows
+        // the issuer and the promise is resolved.
+        brandToLiquidityIssuer.delete(secondaryBrand);
+
         // defer lookup until necessary. more aligned with governed
         // param we expect this to be eventually.
         const addIssuerToReserve = getAddIssuerToReserve();
+
         // tell the reserve about this brand, which it will validate by
         // calling back to AMM for the issuer
         return addIssuerToReserve(secondaryBrand).then(() => liquidityIssuer);
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -64,46 +64,40 @@ export const makeAddIssuer = (
     /** @template T @typedef {import('@endo/promise-kit').PromiseKit<T>} PromiseKit */
     /** @type {PromiseKit<Issuer<'nat'>>} */
     const promiseKit = makePromiseKit();
-
-    // If there's a race to add an issuer, one will succeed here and
-    // others will fail. The failures should get the promise.
-    try {
-      brandToLiquidityIssuer.init(secondaryBrand, promiseKit.promise);
-    } catch (e) {
-      return brandToLiquidityIssuer.get(secondaryBrand);
-    }
+    brandToLiquidityIssuer.init(secondaryBrand, promiseKit.promise);
 
     const liquidityKeyword = `${keyword}Liquidity`;
     zcf.assertUniqueKeyword(liquidityKeyword);
 
-    const { issuer } = await zcf.saveIssuer(secondaryIssuer, keyword);
-    // AWAIT ///////////////
-
-    const mint = await zcf.makeZCFMint(
-      liquidityKeyword,
-      AssetKind.NAT,
-      harden({ decimalPlaces: DECIMAL_PLACES }),
-    );
-    // AWAIT ///////////////
-
-    console.log(
-      X`Saved issuer ${secondaryIssuer} to keyword ${keyword} and got back ${issuer}`,
-    );
-    // this ensures that getSecondaryIssuer(thisIssuer) will return even
-    // before the pool is created
-    brandToLiquidityMint.init(secondaryBrand, mint);
-    const { issuer: liquidityIssuer } = mint.getIssuerRecord();
-    // defer lookup until necessary. more aligned with governed
-    // param we expect this to be eventually.
-    const addIssuerToReserve = getAddIssuerToReserve();
-    // tell the reserve about this brand, which it will validate by
-    // calling back to AMM for the issuer
-    await addIssuerToReserve(secondaryBrand);
-    // AWAIT ///////////////
-
-    promiseKit.resolve(liquidityIssuer);
-    brandToLiquidityIssuer.delete(secondaryBrand);
-    return liquidityIssuer;
+    const decimals = harden({ decimalPlaces: DECIMAL_PLACES });
+    return Promise.all([
+      zcf.saveIssuer(secondaryIssuer, keyword),
+      zcf.makeZCFMint(liquidityKeyword, AssetKind.NAT, decimals),
+    ])
+      .then(([issuer, mint]) => {
+        console.log(
+          X`Saved issuer ${secondaryIssuer} to keyword ${keyword} and got back ${issuer}`,
+        );
+        // this ensures that getSecondaryIssuer(thisIssuer) will return even
+        // before the pool is created
+        brandToLiquidityMint.init(secondaryBrand, mint);
+        const { issuer: liquidityIssuer } = mint.getIssuerRecord();
+        promiseKit.resolve(liquidityIssuer);
+        // defer lookup until necessary. more aligned with governed
+        // param we expect this to be eventually.
+        const addIssuerToReserve = getAddIssuerToReserve();
+        // tell the reserve about this brand, which it will validate by
+        // calling back to AMM for the issuer
+        return addIssuerToReserve(secondaryBrand).then(() => liquidityIssuer);
+      })
+      .catch(e => {
+        console.log(
+          X`Failure Saving issuer ${secondaryIssuer}. Not added to Reserve`,
+        );
+        promiseKit.reject(e);
+        brandToLiquidityIssuer.delete(secondaryBrand);
+        throw e;
+      });
   };
 };
 
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -63,8 +63,8 @@ export const makeAddIssuer = (
 
     /** @template T @typedef {import('@endo/promise-kit').PromiseKit<T>} PromiseKit */
     /** @type {PromiseKit<Issuer<'nat'>>} */
-    const promiseKit = makePromiseKit();
-    brandToLiquidityIssuer.init(secondaryBrand, promiseKit.promise);
+    const liquidityPromiseKit = makePromiseKit();
+    brandToLiquidityIssuer.init(secondaryBrand, liquidityPromiseKit.promise);
 
     const liquidityKeyword = `${keyword}Liquidity`;
     zcf.assertUniqueKeyword(liquidityKeyword);
@@ -82,7 +82,7 @@ export const makeAddIssuer = (
         // before the pool is created
         brandToLiquidityMint.init(secondaryBrand, mint);
         const { issuer: liquidityIssuer } = mint.getIssuerRecord();
-        promiseKit.resolve(liquidityIssuer);
+        liquidityPromiseKit.resolve(liquidityIssuer);
         // defer lookup until necessary. more aligned with governed
         // param we expect this to be eventually.
         const addIssuerToReserve = getAddIssuerToReserve();
@@ -94,8 +94,9 @@ export const makeAddIssuer = (
         console.log(
           X`Failure Saving issuer ${secondaryIssuer}. Not added to Reserve`,
         );
-        promiseKit.reject(e);
+        liquidityPromiseKit.reject(e);
         brandToLiquidityIssuer.delete(secondaryBrand);
+        brandToLiquidityMint.delete(secondaryBrand);
         throw e;
       });
   };
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -3,58 +3,99 @@
 import { E } from '@endo/eventual-send';
 import { AmountMath, AssetKind } from '@agoric/ertp';
 import { assertProposalShape } from '@agoric/zoe/src/contractSupport/index.js';
+import { makePromiseKit } from '@endo/promise-kit';
 
 import { definePoolKind } from './pool.js';
 
 const { details: X } = assert;
 
+const DISPLAY_INFO = harden({ decimalPlaces: 6 });
+
 /**
  * @param {ZCF} zcf
  * @param {(Brand) => boolean} isInSecondaries
  * @param {WeakStore<Brand,ZCFMint>} brandToLiquidityMint
+ * @param {WeakStore<Brand,ERef<Issuer>>} brandToLiquidityIssuer
  * @param {() => (secondaryBrand: Brand) => Promise<void>} getAddIssuerToReserve
  */
 export const makeAddIssuer = (
   zcf,
   isInSecondaries,
   brandToLiquidityMint,
+  brandToLiquidityIssuer,
   getAddIssuerToReserve,
 ) => {
   /**
+   * Add a new issuer. If we previously received a request for the same issuer,
+   * we'll return a promise or the issuer.
+   *
    * @param {Issuer} secondaryIssuer
    * @param {string} keyword
    */
   return async (secondaryIssuer, keyword) => {
-    const [secondaryAssetKind, secondaryBrand] = await Promise.all([
-      E(secondaryIssuer).getAssetKind(),
-      E(secondaryIssuer).getBrand(),
-    ]);
-    assert(
-      !isInSecondaries(secondaryBrand),
-      X`issuer ${secondaryIssuer} already has a pool`,
-    );
-    assert(
-      secondaryAssetKind === AssetKind.NAT,
-      X`${keyword} asset not fungible (must use NAT math)`,
-    );
-    const liquidityKeyword = `${keyword}Liquidity`;
-    zcf.assertUniqueKeyword(liquidityKeyword);
+    return E.when(
+      Promise.all([
+        E(secondaryIssuer).getAssetKind(),
+        E(secondaryIssuer).getBrand(),
+      ]),
+      ([secondaryAssetKind, secondaryBrand]) => {
+        assert(
+          secondaryAssetKind === AssetKind.NAT,
+          X`${keyword} asset not fungible (must use NAT math)`,
+        );
 
-    const mint = await zcf.makeZCFMint(
-      liquidityKeyword,
-      AssetKind.NAT,
-      harden({ decimalPlaces: 6 }),
+        if (brandToLiquidityIssuer.has(secondaryBrand)) {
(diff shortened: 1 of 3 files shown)


Chris-Hibbert:
@erights   FYI

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -12,50 +13,92 @@ const { details: X } = assert;
...
 export const makeAddIssuer = (
   zcf,
   isInSecondaries,
   brandToLiquidityMint,
+  brandToLiquidityIssuer,
   getAddIssuerToReserve,
 ) => {
   /**
+   * Add a new issuer. If we previously received a request for the same issuer,
+   * we'll return a promise or the issuer.
```
erights: did you mean
```suggestion
   * we'll return a promise for the issuer.
```
?

erights (commented):
Wonderful that you removed non-test `await`s! Should really help us think about what happens during which turn.

Review not done. Just a comment so far.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: I actually meant that we'll return either the issuer itself or a promise for the issuer, depending.  On line 58, we return the issuer, while on line 46, we return a Promise. Since the method is `async`, the caller will always get a promise, but sometimes it's already resolved, and other times it'll be resolved later.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -12,50 +13,92 @@ const { details: X } = assert;
...
+                  X`Saved issuer ${secondaryIssuer} to keyword ${keyword} and got back ${issuer}`,
+                );
+                // this ensures that getSecondaryIssuer(thisIssuer) will return even
+                // before the pool is created
+                brandToLiquidityMint.init(secondaryBrand, mint);
+                const { issuer: liquidityIssuer } = mint.getIssuerRecord();
+                // defer lookup until necessary. more aligned with governed
+                // param we expect this to be eventually.
+                const addIssuerToReserve = getAddIssuerToReserve();
+                // tell the reserve about this brand, which it will validate by
```
dtribble: why does the AMM know anything about the Reserve?

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -12,50 +13,92 @@ const { details: X } = assert;
...
-      X`${keyword} asset not fungible (must use NAT math)`,
-    );
-    const liquidityKeyword = `${keyword}Liquidity`;
-    zcf.assertUniqueKeyword(liquidityKeyword);
-
-    const mint = await zcf.makeZCFMint(
-      liquidityKeyword,
-      AssetKind.NAT,
-      harden({ decimalPlaces: 6 }),
+    return E.when(
```
dtribble, line 70: The translation to promise format seems unnecessary

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -12,50 +13,92 @@ const { details: X } = assert;
...
+        if (brandToLiquidityIssuer.has(secondaryBrand)) {
+          return brandToLiquidityIssuer.get(secondaryBrand);
+        }
+
+        assert(
+          !isInSecondaries(secondaryBrand),
+          X`issuer ${secondaryIssuer} already has a pool`,
+        );
+
+        if (brandToLiquidityMint.has(secondaryBrand)) {
```
dtribble: Use the `provide` library for this pattern?

packages/run-protocol/src/vpool-xyk-amm/addPool.js
dtribble:
```javascript
export const makeAddIssuer = (
  zcf,
  isInSecondaries,
  brandToLiquidityMint,
  brandToLiquidityIssuer,
  getAddIssuerToReserve,
) => {
  /**
   * Add a new issuer. If we previously received a request for the same issuer,
   * we'll return a promise or the issuer.
   *
   * @param {Issuer} secondaryIssuer
   * @param {string} keyword
   */
  return async (secondaryIssuer, keyword) => {
    const [secondaryAssetKind, secondaryBrand] = await Promise.all([
      E(secondaryIssuer).getAssetKind(),
      E(secondaryIssuer).getBrand(),
    ]);

    assert(
      secondaryAssetKind === AssetKind.NAT,
      X`${keyword} asset not fungible (must use NAT math)`,
    );

    if (brandToLiquidityIssuer.has(secondaryBrand)) {
      return brandToLiquidityIssuer.get(secondaryBrand);
    }

    assert(
      !isInSecondaries(secondaryBrand),
      X`issuer ${secondaryIssuer} already has a pool`,
    );

    /** @template T @typedef {import('@endo/promise-kit').PromiseKit<T>} PromiseKit */
    /** @type {PromiseKit<Issuer<'nat'>>} */
    const promiseKit = makePromiseKit();
    brandToLiquidityIssuer.init(secondaryBrand, promiseKit.promise);

    const liquidityKeyword = `${keyword}Liquidity`;
    zcf.assertUniqueKeyword(liquidityKeyword);

    const { issuer } = await zcf.saveIssuer(secondaryIssuer, keyword);
    const mint = await zcf.makeZCFMint(
      liquidityKeyword,
      AssetKind.NAT,
      harden({ decimalPlaces: 6 }),
    );
    console.log(
      X`Saved issuer ${secondaryIssuer} to keyword ${keyword} and got back ${issuer}`,
    );
    // this ensures that getSecondaryIssuer(thisIssuer) will return even before the pool is created
    brandToLiquidityMint.init(secondaryBrand, mint);
    const { issuer: liquidityIssuer } = mint.getIssuerRecord();
    // defer lookup until necessary. more aligned with governed param we expect this to be eventually.
    const addIssuerToReserve = getAddIssuerToReserve();
    // tell the reserve about this brand, which it will validate by calling back
    // to AMM for the issuer
    await addIssuerToReserve(secondaryBrand);
    promiseKit.resolve(liquidityIssuer);
    brandToLiquidityIssuer.delete(secondaryBrand);
    return liquidityIssuer;
  };
};
```

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -263,6 +268,7 @@ const start = async (zcf, privateArgs) => {
     zcf,
     isSecondary,
     secondaryBrandToLiquidityMint,
+    secondaryBrandToLiquidityIssuerPromise,
```
dtribble, line 283: it's not called out as promises in the callee. Why does this need to get provided as an argument instead of allocaed in the callee?

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
```diff
@@ -145,6 +145,33 @@ const makeAssertPayouts = (
...
+  const centralR = makeIssuerKit('central');
+  const moolaR = makeIssuerKit('moola');
+
+  const { amm } = await setupAmmServices(
+    t,
+    { committeeName: 'EnBancPanel', committeeSize: 3 },
+    centralR,
+  );
+
+  const [issuer1, issuer2, issuer3] = await Promise.all([
```
dtribble, line 159: This should also test adding the same issuer with the SAME keyword.

dtribble (changes requested):
i think the changes are straightforward.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: from #5429, 

> When the AMM makes a pool, add the collateral to the reserve.

> This requires that the reserve has the issuer, so when an issuer is added to AMM, also add it to the reserve. Put this on the public facet so the AMM doesn't need special access, but use a special method so the the reserve validates it back against the AMM.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: @erights   Your comment seems to prefer the form without `await`.   Which preference is stronger? @dtribble?

> Wonderful that you removed non-test awaits! Should really help us think about what happens during which turn.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: I changed the comment to say 'we'll return the issuer or a promise for it.'

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: I don't think it's relevant. `provide` works when you want to get or make the thing. Here it's saying, if the brand is already knows, look it up.  The or-else is the whole rest of the method. Would you recommend calling that inside the final argument to `provide`?

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: You're right that I could (and maybe should) have put it in makeAddIssuer, but we'll be moving to durability shortly, and at that point, addIssuer will need to get the collection from its caller anyway.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
dtribble: Interesting. This makes me realize that the pool creation should be a separate contract with the AMM creator facet, which is imposing conditions on the pool creation (like min liq). Does it make sense to separate out?

packages/run-protocol/src/vpool-xyk-amm/addPool.js
dtribble: The problem was that the interlock was spread across turns. To me, this makes it more obvious and straightforward to prevent that

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
dtribble: I would not expect it to expose its internal state as opposed to simply managing it around baggage, but I agree it will get sufficiently shuffled that this doesn't matter,

packages/run-protocol/src/vpool-xyk-amm/addPool.js
dtribble: BTW the "6" in there needs to be promoted to a constant

packages/run-protocol/src/vpool-xyk-amm/addPool.js
dtribble: the "orelse" would be "make and register the brand" as a separate function.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -12,50 +13,92 @@ const { details: X } = assert;
...
+          const { issuer } = brandToLiquidityMint
+            .get(secondaryBrand)
+            .getIssuerRecord();
+          return issuer;
+        }
+
+        /** @template T @typedef {import('@endo/promise-kit').PromiseKit<T>} PromiseKit */
+        /** @type {PromiseKit<Issuer<'nat'>>} */
+        const promiseKit = makePromiseKit();
+        brandToLiquidityIssuer.init(secondaryBrand, promiseKit.promise);
```
dtribble: At this point, we've taken the lock. I believe that this doesn't get correctly pulled in the error case.

dtribble (changes requested):
The failure of the unique keyword assert would leave the promise wedged in the collection forever. Not sure you can easily induce the required order, but the missing "duplicate keyword" test might induce the failure mode.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: > Which preference is stronger?

Dean and MarkM discussed it. I couldn't tell whether they came to agreement, but I found Dean's arguments more convincing.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: This rhymes with `provide`, but isn't quite the same. if `brandToLiquidityMint.has(secondaryBrand)` , return `brandToLiquidityMint.get(secondaryBrand).getIssuerRecord().issuer`, or else return `registerLiquidityAndIssuerForBrand()`.

with the new try-catch round `.init()` I left the or-else in-line.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: I added a try-catch around this. The winner of the race will be able to `.init()`, and the also-rans will fail. The also-rans get the promise created by the winner in the catch clause.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: I assume by "separate contract" you mean "separate offer". 

We've decided that creating new pools is permissionless. Given that choice, I don't see a need to separate it further.

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
Chris-Hibbert: good idea.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -29,32 +37,71 @@ export const makeAddIssuer = (
...
+
+    /** @template T @typedef {import('@endo/promise-kit').PromiseKit<T>} PromiseKit */
+    /** @type {PromiseKit<Issuer<'nat'>>} */
+    const promiseKit = makePromiseKit();
+
+    // If there's a race to add an issuer, one will succeed here and
+    // others will fail. The failures should get the promise.
+    try {
+      brandToLiquidityIssuer.init(secondaryBrand, promiseKit.promise);
+    } catch (e) {
```
dtribble: This is not the error case I was thinking of. I think this should not be possible because the `has` check is above in the same turn. The case I'm concerned about is if it fails below (e.g., in `saveIssuer` or  `addIssuerToReserve`).  Those would prevent the promise from being resolved on line 103.

dtribble (changes requested):

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: here's a fix.

Chris-Hibbert:
@dtribble  I added optional reservations for Keywords.  PTAL.

packages/zoe/src/contractFacet/zcfZygote.js
```diff
@@ -323,6 +343,16 @@ export const makeZCFZygote = (
       }
     },
     getInstance: () => getInstanceRecord().instance,
+    reserveKeyword: keyword => {
+      const key = {};
```
dtribble: is this a token?

packages/zoe/src/contractFacet/zcfZygote.js
```diff
@@ -260,7 +273,14 @@ export const makeZCFZygote = (
     // accept raw functions. assert cannot be a valid passable!
     reallocate,
     assertUniqueKeyword,
-    saveIssuer: async (issuerP, keyword) => {
+    saveIssuer: async (issuerP, keyword, token = undefined) => {
```
dtribble: DRY

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -29,33 +37,67 @@ export const makeAddIssuer = (
...
+
+    if (brandToLiquidityMint.has(secondaryBrand)) {
+      const { issuer } = brandToLiquidityMint
+        .get(secondaryBrand)
+        .getIssuerRecord();
+      return issuer;
+    }
+
+    /** @template T @typedef {import('@endo/promise-kit').PromiseKit<T>} PromiseKit */
+    /** @type {PromiseKit<Issuer<'nat'>>} */
```
turadg: I think this can be simplified to:
```suggestion
    /** @type {import('@endo/promise-kit').PromiseKit<Issuer<'nat'>>} */
```

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -29,33 +37,67 @@ export const makeAddIssuer = (
...
+        liquidityPromiseKit.resolve(liquidityIssuer);
+        // defer lookup until necessary. more aligned with governed
+        // param we expect this to be eventually.
+        const addIssuerToReserve = getAddIssuerToReserve();
+        // tell the reserve about this brand, which it will validate by
+        // calling back to AMM for the issuer
+        return addIssuerToReserve(secondaryBrand).then(() => liquidityIssuer);
+      })
+      .catch(e => {
+        console.log(
```
turadg:
```suggestion
        console.error(
```

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -29,33 +37,67 @@ export const makeAddIssuer = (
...
-    // this ensures that getSecondaryIssuer(thisIssuer) will return even before the pool is created
-    brandToLiquidityMint.init(secondaryBrand, mint);
-    const { issuer: liquidityIssuer } = mint.getIssuerRecord();
-    // defer lookup until necessary. more aligned with governed param we expect this to be eventually.
-    const addIssuerToReserve = getAddIssuerToReserve();
-    // tell the reserve about this brand, which it will validate by calling back
-    // to AMM for the issuer
-    await addIssuerToReserve(secondaryBrand);
-    return liquidityIssuer;
+    const decimals = harden({ decimalPlaces: DECIMAL_PLACES });
```
turadg:
```suggestion
    const displayInfo = harden({ decimalPlaces: DECIMAL_PLACES });
```

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: done

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: done

packages/run-protocol/src/vpool-xyk-amm/addPool.js
Chris-Hibbert: done

turadg (approved):

Chris-Hibbert:
Added ##5692 for the race in `saveIssuer`.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -3,24 +3,32 @@
...
+  brandToLiquidityIssuer,
   getAddIssuerToReserve,
 ) => {
   /**
+   * Add a new issuer. If we previously received a request for the same issuer,
+   * we'll return the issuer or a promise for it.
+   *
    * @param {Issuer} secondaryIssuer
    * @param {string} keyword
    */
```
dtribble, line 34: If it's easy to change this to declaring the function `addIssuer` and then returning it, I think that would be more readable and debuggable,

dtribble (approved):
It still seems a little heavyweight, but the interlock looks good, and robust against any race issue in zcf itself.  A few minor optional comments.

PR #5597, opened 2022-06-15
style: trivial changes to make lint happy

erights:
For some reason, when a newline is placed between a function jsdoc-type declaration and the function itself, ts type checking makes fewer bad rejections while otherwise seeming to stay as accurate. Discovered by accident. I have no idea why.

Also removing extraneous spaces within the outer curlies of jsdoc type decls, as our style is generally to omit them. By omitting them consistently, it is easier to find things with trivial greps.

See https://github.com/endojs/endo/pull/1218

An example is the `insistManagerType` function, which before the PR can be found at
https://github.com/Agoric/agoric-sdk/blob/f677509537bcfd3c7a1a2a88e7c865db1fdb60d5/packages/SwingSet/src/kernel/state/kernelKeeper.js#L300-L316

In that state, it generated no warnings. But if the unnecessary 
```js
return undefined;
```
on line 315 is removed, then we get
```
$ yarn lint
yarn run v1.22.17
$ yarn lint:types&&yarn lint:eslint
$ tsc --maxNodeModuleJsDepth 3 -p jsconfig.json
$ eslint .

.../SwingSet/src/kernel/state/kernelKeeper.js
  300:3  warning  JSDoc @returns declaration present but return expression not available in function  jsdoc/require-returns-check

✖ 1 problem (0 errors, 1 warning)
```
However, bizarrely, this PR by inserting a newline after line 304 causes `yarn lint` to pass with no warnings.

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -113,8 +113,8 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
   /**
    * @param {string} name
    * @param {(request: Uint8Array) => Promise<Uint8Array>} handleCommand
-   * @param { boolean } [metered]
-   * @param { string } [snapshotHash]
+   * @param {boolean} [metered]
+   * @param {string} [snapshotHash]
    */
   async function startXSnap(
     name,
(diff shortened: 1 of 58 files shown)


turadg (approved):
Change LGTM and matches the style we want to adopt universally. (The auto format PR is on hold)

Two requests though:
1. Please make this a "style:" commit for the benefit of git blame digging. Can be done without another push; just change pr title and do a squash merge.
2. Describe some examples of the errors you were encountering. It's very surprising to me that white space would matter.

erights:
@turadg looking at the shell trace of the failure I record above, it looks like typescript itself doesn't fail mysteriously, but eslint does, based on eslint also trying to interpret jsdoc type declarations. Does this make any sense?

turadg:
> Does this make any sense?

Yep, thanks. I think what was happening is that the jsdoc linter plugin for eslint doesn't understand TypeScript's jsdoc notation for assertion functions. (JSdoc had a types description system before TS and TS adopted it, but the function assertion expression doesn't fit so well) The reason it goes away when you add a newline after the jsdoc is, I suppose,  that eslint no longer sees it as documenting the function so ignores it.

Issue #5152, opened 2022-04-19
xsnap message timestamp logging/reporting


warner:
## What is the Problem Being Solved?

Our "slogfile" currently records kernel-side timestamps of messages going to and from vats: deliveries, syscall requests, syscall responses, and delivery results.

To support investigations like #5139, as well as to collect data for performance improvements, it would be nice if we also had worker-side timestamps for those same messages. We expect the cross-process messaging to be quick, but we don't actually have enough data to prove it. And we expect the kernel to be idle and thus react immediately to a worker's syscall request, but if the kernel were stuck for some reason (VM thrashing, waiting for some IO write), then a late timestamp on the receipt of a syscall request would look identical to the worker spending a long time before emitting the syscall.

## Description of the Design

I'm thinking that `xsnap.c` should have an array (maybe pre-allocated) of timestamps. Whenever it receives the end of a netstring frame, it should append a `gettimeofday` (ideally as 64-bit seconds-plus-microseconds-since-epoch) to the array. When it sends a string to the parent process, it should also append a timestamp. We might need to include a tiny bit (maybe just one character) of the message so we can distinguish between console messages and syscalls.

When the delivery is complete and the worker is serializing the results, it should fetch one last timestamp, then serialize the entire array and include it in the metering results. Then it should reset the array length to zero in preparation for the next delivery to begin.

On the kernel side, we already include the full metering record in the slogfile. As long as the kernel doesn't attempt to interpret the additional data, we need no additional modifications.

With this information in the slogfile, we can write post-mortem tools that correlate the sends and receives of each syscall request and response, and build a picture of how long the worker spent working, as opposed to how long it spent waiting for the kernel.

## Security Considerations

None, the additional metering data is not shared elsewhere in the kernel, and is not included in consensus.

## Test Plan

Unit tests in `packages/xsnap/` asserting the presence of the timetable and its correct length, and that the values are sorted and bounded by parent-side timestamps captured before and after the worker interaction.

Tartuffo:
@warner Is this needed for MN-1 or can it go into the Product Backlog?

warner:
This would help with the perf investigations called for by #5507 and #5361 .

warner:
Here's a (python) tool to scan a slogfile with the new `meterResults.timestamps` data and produce a report:
```python
import sys, json, gzip, re, time
from collections import defaultdict
from pprint import pprint

class Syscall:
    def __init__(self, vsc, num, rx_request):
        self.num = num
[...]

PR #5598, opened 2022-06-15
fix(xsnap): upgrade to latest xsnap, now with timestamps, and update METER_TYPE

warner:
fix(xsnap): upgrade to latest xsnap, now with timestamps

This moves to the version of `xsnap-worker.c` which includes
receipt/transmit timestamps in the metering results.

Tests are updated to expect 'timestamps' in xsnap metering results,
and to assert that they correctly interleave with the times of
upstream/downstream commands as seen by the parent process.

refs #5152

### DIFF
--- a/packages/xsnap/api.js
+++ b/packages/xsnap/api.js
@@ -3,7 +3,7 @@
 /** The version identifier for our meter type.
  * TODO Bump this whenever there's a change to metering semantics.
  */
-export const METER_TYPE = 'xs-meter-13';
+export const METER_TYPE = 'xs-meter-14';
 
 export const ExitCode = {
   E_UNKNOWN_ERROR: -1,
--- a/packages/xsnap/test/test-xs-perf.js
+++ b/packages/xsnap/test/test-xs-perf.js
@@ -1,4 +1,4 @@
-/* global performance */
+/* global globalThis performance */
 // @ts-check
 
 import '@endo/init/debug.js';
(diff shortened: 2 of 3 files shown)


packages/xsnap/test/test-xs-perf.js
```diff
@@ -59,10 +59,81 @@ test('meter details', async t => {
...
+    'kern receive syscall 2',
+    'kern send syscall-result 2',
+    '    worker rx syscall-result 2',
+    '    worker tx delivery-result',
+    'kern receive deliver-result',
+  ];
+
+  t.deepEqual(sortedNames, expected);
+
+  // on my 2022 MBP (M1 Pro), syscalls take 75-600us to get from
```
dckc, line 134: I like to `t.log()` such numbers so I can eyeball them with `ava test -v`.

dckc (approved):
looks good. one suggestion to consider...

packages/xsnap/api.js
```diff
@@ -3,7 +3,7 @@
 /** The version identifier for our meter type.
  * TODO Bump this whenever there's a change to metering semantics.
  */
-export const METER_TYPE = 'xs-meter-13';
+export const METER_TYPE = 'xs-meter-14';
```
mhofman, line 6: What does "a change to metering semantics" mean exactly? What is the purpose of this bump? Should this be automated, or at least tested so that an effective change to XS is represented here?

dckc: indeed, there is a test:

https://github.com/Agoric/agoric-sdk/blob/102a5b62f157a3f2d236979efbf80e2996d72466/packages/xsnap/test/test-xs-perf.js#L49-L53

mhofman: Interesting, I don't see what in this PR causes the need to update the meter type.

dckc: Perhaps the addition of timestamps was, conservatively, considered to be a change in the metering semantics. Note the test for the shape of the meters object just after the test for magnitudes.

PR #5599, opened 2022-06-15
fix(swingset): use gettimeofday() in the slogfile, not CLOCK_MONOTONIC

warner:
fix(swingset): use gettimeofday() in the slogfile, not CLOCK_MONOTONIC

Unix provides multiple time APIs. `gettimeofday()` reports
seconds-since-epoch with microsecond resolution, and can jump around,
speed up, or even go backwards if/when someone changes the clock on
the host computer. `clock_gettime(CLOCK_MONOTONIC)` never goes
backwards, has nanosecond resolution, and in practice reports
seconds-since-last-reboot (of the host computer). `CLOCK_MONOTONIC` is
implemented (at least on linux) by initializing an offset (to the
current time) at boot, and reporting `gettimeofday() - offset`. If
someone calls `settimeofday()`, the kernel measures the delta, and
subtracts it from the offset, so the `CLOCK_MONOTONIC` value does
change even though the `gettimeofday()` does. However both
`gettimeofday()` and `CLOCK_MONOTONIC` will reflect the gentler
frequency adjustments that NTP makes (through `adjtime()`) to keep the
clock in sync without timequakes.

Previously, our slogfile's `.time` property recorded
`performance.timeOrigin + performance.now()`, which is effectively the
sum of `gettimeofday()` at process start, plus the delta in
`CLOCK_MONOTONIC` between one sample at process start and another when
`now()` is called, plus a tiny offset because `gettimeofday()` and the
initial sample cannot happen exactly simultaneously. This has the
benefit of never decreasing, and preserves the accuracy of
intervals (timestamp subtractions) that span a `settimeofday()`
timequake. But has the downside of becoming unique per-process when
timequakes happen: if you compare this value against a similar one
built in a different process, they will be different by the sum of any
timequakes that occurred between the first process launching and the
second one launching. If a process restarts, the timestamps before and
after the restart will be offset by a similar amount.

This changes the slogfile to use `gettimeofday()`, via a small C
module named `microtime`. This matches what #5152 changes in
`xsnap-worker.c` to record delivery/syscall send+receive
timestamps. By using `gettimeofday()` on both sides, the timestamps
remain comparable independent of when the processes are started or
restarted. The drawback is that spans which cross a timequake will be
corrupted, even becoming negative if `settimeofday()` causes the
system clock to go backwards far enough. However we only really pay
attention to one span at a time (a single block, or delivery, or
syscall), so the damage is limited. I care more about comparability of
timestamps across long-running processes (potentially across multiple
hopefully-NTP-synced machines) than I'm worried about the occasional
timequake-corrupted span.

If we manage to make swingset fast enough that the microsecond
resolution of `gettimeofday()` feels insufficient, we should probably
change both the slogger and `xsnap-worker.c` to use the earlier
approach: `timeOrigin + CLOCK_MONOTONIC` (which will involve writing
new code on the `xsnap-worker.c` side).

refs #5152

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -10,6 +10,7 @@ import { spawn as ambientSpawn } from 'child_process';
 import { type as osType } from 'os';
 import { Worker } from 'worker_threads';
 import anylogger from 'anylogger';
+import microtime from 'microtime';
 
 import { assert, details as X } from '@agoric/assert';
 import { importBundle } from '@endo/import-bundle';
@@ -205,12 +206,13 @@ export async function makeSwingsetController(
       return;
     }
 
-    // Create an object with the property order that loadgen expects.
-    const timedObj = { time: undefined, type: undefined, ...obj };
+    // microtime gives POSIX gettimeofday() with microsecond resolution
+    const time = microtime.nowDouble();
+    // this is CLOCK_MONOTONIC, seconds since process start
+    const monotime = performance.now() / 1000;
 
-    // Update the timedObj timestamp.
-    const timeMS = performance.timeOrigin + performance.now();
-    timedObj.time = timeMS / 1000;
+    // rearrange the fields a bit to make it more legible to humans
+    const timedObj = { type: undefined, ...obj, time, monotime };
 
     // Serialise just once.
     const jsonObj = JSON.stringify(timedObj, (_, arg) =>
--- a/packages/SwingSet/package.json
+++ b/packages/SwingSet/package.json
@@ -46,6 +46,7 @@
     "anylogger": "^0.21.0",
     "import-meta-resolve": "^1.1.1",
     "lmdb": "^2.4.5",
+    "microtime": "^3.1.0",
     "semver": "^6.3.0"
   },
   "peerDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -12086,6 +12086,14 @@ micromatch@^4.0.2, micromatch@^4.0.4:
     braces "^3.0.1"
     picomatch "^2.2.3"
 
+microtime@^3.1.0:
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/microtime/-/microtime-3.1.0.tgz#599a71250e3116c59f0fe5271dae4cc44321869c"
+  integrity sha512-GcjhfC2y/DF2znac8IRwri7+YUIy34QRHz/iZK3bHrh74qrNNOpAJQwiOMnIG+v1J0K4eiqd+RiGzN3F1eofTQ==
+  dependencies:
+    node-addon-api "^5.0.0"
+    node-gyp-build "^4.4.0"
+
 mime-db@1.52.0, "mime-db@>= 1.43.0 < 2":
   version "1.52.0"
   resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.52.0.tgz#bbabcdc02859f4987301c856e3387ce5ec43bf70"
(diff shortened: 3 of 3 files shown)


warner:
BTW I skimmed through the source of `microtime` (https://github.com/wadey/node-microtime) and it's pretty straightforward, and has no additional dependencies, which makes me feel better about introducing a new dependency to `agoric-sdk`.

packages/SwingSet/src/controller/controller.js
```diff
@@ -208,9 +208,8 @@ export async function makeSwingsetController(
     // Create an object with the property order that loadgen expects.
     const timedObj = { time: undefined, type: undefined, ...obj };
 
-    // Update the timedObj timestamp.
-    const timeMS = performance.timeOrigin + performance.now();
-    timedObj.time = timeMS / 1000;
+    // Update the timedObj timestamp, this uses unix gettimeofday()
+    timedObj.time = microtime.nowDouble();
```
mhofman: I would love if we could still keep a monotonic source in there. Is there any reason we couldn't include both `gettimeofday()` as `time` and `performance.now()` (without adding the origin, which becomes meaningless) as a second field?

```suggestion
    timedObj.time = microtime.nowDouble();
    timedObj.stamp = performance.now() / 1000;
```

That would require me to push an update to the loadgen, but that's pretty straightforward.

warner: yeah, ok, it feels like a bit of bloat to me, but I can see the value of it

Currently, when we want to compare slogfiles between two different validators, we have to pre-filter them with `jq 'delete(.time)'`, this will make that change to `delete(.time, .stamp)`. But actually my new timestamps-in-metering-results change will require something like `delete(.time, .stamp, .dr.[2].timestamps)`, so really we should build a specific tool for that comparison (which also tolerates only-partially-overlapping slogfiles, and startup-time replays, and such).

How about `.reltime` or `.relative` for the monotonic value? "stamp" doesn't speak to me as much.

mhofman: > really we should build a specific tool for that comparison (which also tolerates only-partially-overlapping slogfiles, and startup-time replays, and such).

Yes! Would love something that can stitch slogs together too! Currently my slog diffing is very limited:
https://github.com/Agoric/agoric-sdk/blob/1505285b137e1aa6a91246dc6f4b771650ef7612/scripts/process-integration-results.sh#L175-L184

> How about `.reltime` or `.relative` for the monotonic value? "stamp" doesn't speak to me as much.

Either is fine by me (I almost didn't want to include a name because I wasn't inspired)

warner: I pushed a change, it now uses `{ type, ...rest, time, monotime }`

packages/SwingSet/src/controller/controller.js
```diff
@@ -205,12 +206,13 @@ export async function makeSwingsetController(
...
+    // microtime gives POSIX gettimeofday() with microsecond resolution
+    const time = microtime.nowDouble();
+    // this is CLOCK_MONOTONIC, seconds since process start
+    const monotime = performance.now() / 1000;
 
-    // Update the timedObj timestamp.
-    const timeMS = performance.timeOrigin + performance.now();
-    timedObj.time = timeMS / 1000;
+    // rearrange the fields a bit to make it more legible to humans
+    const timedObj = { type: undefined, ...obj, time, monotime };
```
mhofman, line 215: Change of field order addressed in https://github.com/Agoric/testnet-load-generator/pull/86

mhofman (approved):
LGTM, pending merge of the related loadgen changes (which are needed before this PR will pass `deployment-test`)

warner:
Aye, I'll rebase/tidy-up this one while we wait for the loadgen PR to land

Issue #3916, opened 2021-10-01
errors in type declarations are silently ignored (not using noImplicitAny)


dckc:
## Describe the problem

In #3914, we have a `Amount` vs `amount` typo in `@returns { amount }`. The `tsc` compiler just (silently) treats it as a text comment so that the return type of `@callback MultiplyBy` is `any`, hiding dozens of downstream typing issues.

That issue is a specific case of a general problem where we have to manually ensure that our types are not implicitly degrading to `any`.

## To Reproduce

see, for example, #3914

## Expected behavior
Syntax errors in type declarations should get flagged as lint problems.

## Additional context

TypeScript offsers a [noImplicitAny](https://www.typescriptlang.org/tsconfig#noImplicitAny) that would catch more typos, but the cost of using it is significant.

Meanwhile, this issue serves as a place to document the cost of _not_ using it.

cc @mhofman @michaelfig @kriskowal @erights

michaelfig:
This is something that I was hoping will be solved by better JSDoc support.

gajus/eslint-plugin-jsdoc#145 tracks the missing JSDoc Typescript parsing support for the types listed in jsdoc-type-pratt-parser/jsdoc-type-pratt-parser#96.

Whenever progress is made on those issues, we can try stricter validation (with warnings, not errors), by adding the following:

```
  "rules": {
    "jsdoc/valid-types": 1,
    "jsdoc/no-undefined-types": 1
  }
```

to `packages/eslint-config/eslintrc.json`, then try running `yarn lerna run lint`.  

When that stricter validation actually doesn't report a huge number of spurious errors (maybe with minimal necessary changes to the typing), then that's when we can turn it on permanently (with `2` instead of `1`, which makes failures into errors).

dckc:
@turadg FYI

michaelfig:
I found that the solution to this particular problem is JSDoc is to add:

```js
// @ts-check
```

to the top of the file.

dckc:
support for `@callback` is a little iffy: https://github.com/microsoft/TypeScript/issues/36101

michaelfig:
The following is a more specific report of the inconsistency I'm seeing: https://github.com/microsoft/TypeScript/issues/47608

PR #5607, opened 2022-06-15
fix(vat-data): utility type FunctionsPlusContext

turadg:
## Description

@erights noticed `FunctionsPlusContext` never had the Context bound. I think the reason there wasn't an error was https://github.com/Agoric/agoric-sdk/issues/3916

This adds an explicit test, takes the Context type parameter and provides in vaultDirector where needed.

### Security Considerations

--

--

### Testing Considerations

New types test


erights (approved):
LGTM, thanks!

PR #5609, opened 2022-06-16
fix: committee exit the seat when giving voter invitation

### LINKED ISSUE #5483: voter invitation offer hangs in wallet
### Symptoms

  - a deploy script ([vote-raise-debt-limit.js](https://gist.github.com/dckc/7733f85be2c25a1a47b639e4b94816ac#file-vote-raise-debt-limit-js)) that does `const voterFacet = await E(zoe).offer(voterInvitation)` seems to work, but
  - a script that uses the wallet to handle the offer ([vote-raise-debt-limit-wb.js](https://gist.github.com/dckc/7733f85be2c25a1a47b639e4b94816ac#file-vote-raise-debt-limit-wb-js)) does not; the offer hangs in "pending" state and the wallet doesn't seems to update the offers notifier until I hit **exit**

### Diagnosis (presumed)

The offer handler in committee.js doesn't exit the seat:

https://github.com/Agoric/agoric-sdk/blob/ee043dc013f3aadf63473825968f362d83ca6f1f/packages/governance/src/committee.js#L41-L43

I don't have a clear sense of what `seat.exit()` is all about. I wonder if we have clear docs anywhere.

by the way.. why is that offer handler `Far`? Anybody know?

cc @michaelfig @Chris-Hibbert

Chris-Hibbert:
fixes: #5483

## Description

The handler that distributes a voter invitation wasn't exiting the seat. This fixes that and adds a test.

### Security Considerations

Nonee

We should document that contracts are expected to exist seats when they're no longer useful.

### Testing Considerations

verified.


packages/governance/src/committee.js
```diff
@@ -38,8 +38,9 @@ const start = zcf => {
 
   const makeCommitteeVoterInvitation = index => {
     /** @type {OfferHandler} */
-    const offerHandler = Far('voter offerHandler', () => {
+    const offerHandler = seat => {
```
erights, line 41: Why remove the `Far` declaration?

erights (approved):
LGTM, thanks!

packages/governance/src/committee.js
Chris-Hibbert: Dan asked me to justify it, and I couldn't.

I don't see other offerHandlers that are Far.  The voter facet (line 43) is still Far, and that seems to be required.

dckc (approved):

Issue #5593, opened 2022-06-14
plan for durable/non-durable/soft-durable tests in ERTP/zoe/contracts


warner:
## What is the Problem Being Solved?

@Chris-Hibbert and @erights have been enhancing the ERTP code to be "durable": storing its data in durable collections, and providing enough API surface to enable clients (e.g. Zoe and contracts) to involve ERTP in the upgrade process. When a vat that includes a copy of Zoe is upgraded, Zoe will need to be told (at the right time, during `startVat`/`buildRootObject`) that an upgrade is happening, so Zoe can re-attach its internal Kinds. Zoe uses ERTP (so Zoe is a "client" of ERTP), and during this call, Zoe is responsible for invoking an ERTP-provided "we're upgrading now" interface, so ERTP can also re-attach its own internal Kinds.

By converting ERTP's internal collections to be durable (e.g. `makeScalarBigMapStore(label, { durable: true })`), ERTP tightens a requirement on its API surface. Any objects passed into ERTP that happen to be stored in one of these collections are now required to be durable-compatible. This includes vat imports (Presences) and locally-hosted durable objects ("durable Representatives": Representatives of a durable Kind), but excludes Promises, locally-hosted non-durable objects ("virtual Represtatives: Representatives of a merely-virtual Kind), and locally-hosted ephemeral objects (Remotables). For the case of ERTP, the concern is the "Amounts" that are used in Purses and Payments, in particular when the denomination is a `Set` or other non-`Nat`: an `Invitation`, rather than 35 BLD tokens. These Amounts can include object references, so the durable-compatible-or-not requirement becomes relevant.

This new API requirement is imposed on downstream client who use ERTP (e.g. Zoe). Those clients have unit tests that want to focus attention on the logic of the client, not the requirements of ERTP. Many of those tests run in a non-swingset environment (`packages/zoe/test/unitTests/`), where a special library is imported that provides mock implementations of the virtual/durable collection types. Other tests (`packages/zoe/test/swingsetTests/`) are using a real swingset and real vats, however maybe not as many vats as a zoe-using production service would configure.

[...]

PR #5610, opened 2022-06-16
fix: eliminate `fakeDurable` and change global durability rules to match

FUDCo:
This addresses the SwingSet portion of #5593.

- Virtual objects and collections can no longer be declared `fakeDurable`; they are either durable or not, as was the case prior to the introduction of the `fakeDurable` feature.

- The `enableFakeDurable` config flag is renamed to `relaxDurabilityRules`.

- This renamed flag defaults to `false` when running in a swingset (as before) and to `true` when running on top of the mock virtuals library provided by `fakeVirtualSupport.js` (in the latter case it previously was unconditionally `true`; now it is settable as an option to `makeFakeVirtualStuff` et al).

- The meaning of the flag is altered.  By setting it to `true`, rather than allowing `fakeDurable` to be used or not, it instead globally alters the rules of what values can be admitted to durable objects and collections to be the same as their non-durable counterparts.

The expectation is that production code and tests that are exercising vat upgrade scenarios will run with the flag set to `false`, whereas tests that are exercising business logic that do not care about durability and which use mocks that won't pass the durability checks anyway will run with the flag set to `true`.  This expectation is reflected in the default flag values adopted by the kernel (`false`) and the fake virtual support code (`true`) respectively.


packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1446,6 +1445,11 @@ function build(
    * @returns {Promise<void>}
    */
   async function stopVat() {
+    assert(
+      !relaxDurabilityRules,
+      'stopVat not available when relaxDurabilityRules is true',
+    );
```
erights, line 1451: Good!

(No change suggested)

erights (approved):
LGTM

I love seeing code melt away!

This feels right. But we'll see very soon ;)

packages/SwingSet/tools/fakeVirtualSupport.js
```diff
@@ -246,14 +246,18 @@ export function makeFakeLiveSlotsStuff(options = {}) {
   };
 }
 
-export function makeFakeVirtualReferenceManager(fakeStuff) {
+export function makeFakeVirtualReferenceManager(
+  fakeStuff,
+  relaxDurabilityRules = true,
```
warner, line 251: good, nice to have the option to tighten the rules again, even in a fake environment

warner (approved):
Looks good! At least from the swingset side, I'll defer to the others about whether it meets their needs.

PR #5611, opened 2022-06-16
5556 getLiquidityIssuer

### LINKED ISSUE #5556: Difficult to recover the liquidityIssuer that hasn't been added to a pool
## Describe the bug
It's possible for user A to call amm.addIssuer() like so:
```js
        const liquidityIssuer = await E(ammAPI).addIssuer(
            issuers.BLD,
            'BLDPool',
          );
```
and throw away the reference to the subsequent `liquidityIssuer`.  This results in a state where it is difficult for another user to create an AMM pool with that subsequent `liquidityBrand` as a want, unless they manually find the issuer through zoe's getIssuer() interface 
```
E(ammAPI).getLiquidityIssuer(bldIssuer)
Promise.reject("Error: \"secondaryBrand\" not found: (an object)")
```

I believe `getLiquidityIssuer` needs to use `secondaryBrandToLiquidityMint` to be able to yield the liquidityIssuer before a pool is created, so that any user can create a pool for an existing liquidityIssuer that they did not themselves create but that exists in the AMM contract instance.

## To Reproduce
1. Call addIssuer with a brand X
2. throw away the handle to the resulting liquidityIssuer
3. Attempt to create a pool with brand X

## Expected behavior
It should be possible to create a pool from existing liquidityIssuers, and also to look up existing liquidityIssuers that do not exist in a pool yet.

## Platform Environment
[fd85b8f](https://github.com/Agoric/agoric-sdk/commit/fd85b8f1dbd8e3f70859037d226c5011199360ee)

turadg:
closes: #5556

## Description

Allow `getLiquidityIssuer` after the issuer is added, before the pool is defined.

### Security Considerations

No additional privileges, just a convenience and reduction in surprise.

--
### Testing Considerations

No new tests since a corner case for which there isn't consensus that it was a bug.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -29,7 +29,6 @@ export const makeAddIssuer = (
       E(secondaryIssuer).getAssetKind(),
       E(secondaryIssuer).getBrand(),
     ]);
-
     assert(
       !isInSecondaries(secondaryBrand),
       X`issuer ${secondaryIssuer} already has a pool`,
@@ -176,7 +175,6 @@ export const makeAddPoolInvitation = (
     zcf.reallocate(reserveLiquidityTokenSeat, seat);
     seat.exit();
     pool.updateState();
-    brandToLiquidityMint.delete(secondaryBrand);
 
     await onOfferHandled(
       secondaryBrand,
--- a/packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
@@ -157,7 +157,7 @@ const start = async (zcf, privateArgs) => {
   );
 
   /** @type {Store<Brand,PoolFacets>} */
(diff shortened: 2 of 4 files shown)


packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -170,7 +169,8 @@ export const makeAddPoolInvitation = (
     zcf.reallocate(reserveLiquidityTokenSeat, seat);
     seat.exit();
     pool.updateState();
-    brandToLiquidityMint.delete(secondaryBrand);
+    // ??? WHY?
+    // brandToLiquidityMint.delete(secondaryBrand);
```
turadg: I thought this would be a simple change but I spent some time trying to figure out why the issuer wasn't in the `brandToLiquidityMint` store. It gets deleted here.

@Chris-Hibbert why?

Chris-Hibbert: I can't think of a reason to do that. It looks purely like a mistake. I can't think of what I might have intended to do there that would have been similar.

Chris-Hibbert (approved):

PR #5613, opened 2022-06-16
5386 stored metrics

turadg:
## Description

As part of #5386 , makes all metrics subscriptions use chain storage as in https://github.com/Agoric/agoric-sdk/pull/5400/

### Security Considerations

Metrics data stored on chain

Consumers will need to know where to get metrics. The establishes a pattern that it's in `metrics` 

### Testing Considerations

Existing tests pass.

New tests?

### DIFF
--- a/packages/notifier/src/storesub.js
+++ b/packages/notifier/src/storesub.js
@@ -67,10 +67,13 @@ export const makeStoredSubscription = (
 
   /** @type {StoredSubscription<T>} */
   const storesub = Far('StoredSubscription', {
-    getStoreKey: () =>
-      (storageNode ? E(storageNode).getStoreKey() : Promise.resolve({})).then(
-        storeKey => ({ ...storeKey, subscription }),
-      ),
+    getStoreKey: async () => {
+      if (!storageNode) {
+        return harden({ subscription });
+      }
+      const storeKey = await E(storageNode).getStoreKey();
+      return harden({ ...storeKey, subscription });
+    },
     getUnserializer: () => unserializer,
     getSharableSubscriptionInternals:
       subscription.getSharableSubscriptionInternals,
--- a/packages/notifier/src/storesub.js
+++ b/packages/notifier/src/storesub.js
@@ -2,6 +2,7 @@
 import { E } from '@endo/eventual-send';
 import { Far, makeMarshal } from '@endo/marshal';
 import { observeIteration } from './asyncIterableAdaptor.js';
+import { makeSubscriptionKit } from './subscriber.js';
 
 /**
  * Begin iterating the source, storing serialized iteration values.  If the
@@ -82,3 +83,37 @@ export const makeStoredSubscription = (
   return storesub;
 };
 harden(makeStoredSubscription);
+
+/**
+ * @template T
+ * @typedef {object} StoredPublisherKit
+ * @property {StoredSubscription<T>} subscriber TODO: change to StoredSubscriber<T> when available
+ * @property {IterationObserver<T>} publisher
+ */
+
+/**
+ * @template [T=unknown]
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {string} [childPath]
+ * @returns {StoredPublisherKit<T>}
+ */
+export const makeStoredPublisherKit = (storageNode, marshaller, childPath) => {
+  const { publication, subscription } = makeSubscriptionKit();
+
+  if (storageNode && childPath) {
+    storageNode = E(storageNode).getChildNode(childPath);
+  }
+
+  // wrap the subscription to tee events to storage, repeating to this `subscriber`
+  const subscriber = makeStoredSubscription(
+    subscription,
+    storageNode,
+    marshaller,
+  );
+
+  return {
+    publisher: publication,
+    subscriber,
+  };
+};
--- a/packages/run-protocol/src/contractSupport.js
+++ b/packages/run-protocol/src/contractSupport.js
@@ -1,5 +1,6 @@
 // @ts-check
 import { AmountMath } from '@agoric/ertp';
+import { makeStoredPublisherKit } from '@agoric/notifier';
 import { M } from '@agoric/store';
 
 const { details: X, quote: q } = assert;
@@ -79,3 +80,26 @@ export const checkDebtLimit = (debtLimit, totalDebt, toMint) => {
     )}`,
   );
 };
+
+/**
+ * @template T
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @returns {MetricsPublisherKit<T>}
+ */
+export const makeMetricsPublisherKit = (storageNode, marshaller) => {
+  /** @type {import('@agoric/notifier').StoredPublisherKit<T>} */
+  const kit = makeStoredPublisherKit(storageNode, marshaller, 'metrics');
+  return {
+    metricsPublication: kit.publisher,
+    metricsSubscription: kit.subscriber,
+  };
+};
+harden(makeMetricsPublisherKit);
+
+/**
+ * @template T
+ * @typedef {object} MetricsPublisherKit<T>
+ * @property {IterationObserver<T>} metricsPublication
+ * @property {StoredSubscription<T>} metricsSubscription
+ */
--- a/packages/run-protocol/src/proposals/econ-behaviors.js
+++ b/packages/run-protocol/src/proposals/econ-behaviors.js
@@ -266,7 +266,9 @@ export const setupReserve = async ({
   consume: {
     ammCreatorFacet,
     ammInstanceWithoutReserve: ammInstanceWithoutReserveP,
+    board,
     feeMintAccess: feeMintAccessP,
+    chainStorage,
     chainTimerService,
     zoe,
     economicCommitteeCreatorFacet: committeeCreator,
@@ -294,6 +296,7 @@ export const setupReserve = async ({
     },
   },
 }) => {
+  const STORAGE_PATH = 'reserve';
   trace('setupReserve');
   const poserInvitationP = E(committeeCreator).getPoserInvitation();
   const [poserInvitation, poserInvitationAmount] = await Promise.all([
@@ -310,6 +313,12 @@ export const setupReserve = async ({
   );
 
   const feeMintAccess = await feeMintAccessP;
+
+  const chainStoragePresence = await chainStorage;
+  const storageNode = await (chainStoragePresence &&
+    E(chainStoragePresence).getChildNode(STORAGE_PATH));
+  const marshaller = E(board).getReadonlyMarshaller();
+
   const reserveGovernorTerms = {
     timer,
     electorateInstance,
@@ -317,7 +326,12 @@ export const setupReserve = async ({
     governed: {
       terms: reserveTerms,
       issuerKeywordRecord: { Central: centralIssuer },
-      privateArgs: { feeMintAccess, initialPoserInvitation: poserInvitation },
+      privateArgs: {
+        feeMintAccess,
+        initialPoserInvitation: poserInvitation,
+        marshaller,
+        storageNode,
+      },
     },
   };
   /** @type {{ creatorFacet: GovernedContractFacetAccess<AssetReserveCreatorFacet>, publicFacet: GovernorPublic, instance: Instance }} */
@@ -365,6 +379,8 @@ export const setupReserve = async ({
 export const startVaultFactory = async (
   {
     consume: {
+      board,
+      chainStorage,
       chainTimerService,
       priceAuthority: priceAuthorityP,
       zoe,
@@ -389,6 +405,7 @@ export const startVaultFactory = async (
   } = {},
   minInitialDebt = 5_000_000n,
 ) => {
+  const STORAGE_PATH = 'vaultFactory';
   trace('startVaultFactory');
   const installations = await Collect.allValues({
     VaultFactory,
@@ -457,6 +474,11 @@ export const startVaultFactory = async (
     shortfallInvitationAmount,
   });
 
+  const chainStoragePresence = await chainStorage;
+  const storageNode = await (chainStoragePresence &&
+    E(chainStoragePresence).getChildNode(STORAGE_PATH));
+  const marshaller = E(board).getReadonlyMarshaller();
+
   const governorTerms = harden({
     timer,
     electorateInstance,
@@ -468,6 +490,8 @@ export const startVaultFactory = async (
         feeMintAccess,
         initialPoserInvitation,
         initialShortfallInvitation,
+        marshaller,
+        storageNode,
       }),
     },
   });
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -5,10 +5,10 @@ import { makeStore } from '@agoric/store';
 import { AmountMath } from '@agoric/ertp';
 import { handleParamGovernance, ParamTypes } from '@agoric/governance';
 import { offerTo } from '@agoric/zoe/src/contractSupport/index.js';
-import { makeSubscriptionKit } from '@agoric/notifier';
 
 import { AMM_INSTANCE } from './params.js';
 import { makeTracer } from '../makeTracer.js';
+import { makeMetricsPublisherKit } from '../contractSupport.js';
 
 const trace = makeTracer('Reserve', false);
 
@@ -41,7 +41,12 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
  *   governedApis: ['addLiquidityToAmmPool'],
  * }
  * >} zcf
- * @param {{feeMintAccess: FeeMintAccess, initialPoserInvitation: Payment}} privateArgs
+ * @param {{
+ *   feeMintAccess: FeeMintAccess,
+ *   initialPoserInvitation: Payment,
+ *   marshaller?: ERef<Marshaller>,
+ *   storageNode?: ERef<StorageNode>,
+ * }} privateArgs
  */
 const start = async (zcf, privateArgs) => {
   /**
@@ -208,13 +213,16 @@ const start = async (zcf, privateArgs) => {
     zcf.makeInvitation(addCollateralHook, 'Add Collateral');
 
   const { brand: runBrand } = await E(runMint).getIssuerRecord();
-  /** @type {SubscriptionRecord<MetricsNotification>} */
-  const { publication: metricsPublication, subscription: metricsSubscription } =
-    makeSubscriptionKit();
 
   // shortfall in Vaults due to liquidations less than debt. This value can be
   // reduced by various actions which burn RUN.
   let shortfallBalance = AmountMath.makeEmpty(runBrand);
+
+  /** @type {import('../contractSupport.js').MetricsPublisherKit<MetricsNotification>} */
+  const { metricsPublication, metricsSubscription } = makeMetricsPublisherKit(
+    privateArgs.storageNode,
+    privateArgs.marshaller,
+  );
 
   const updateMetrics = () => {
     const metrics = harden({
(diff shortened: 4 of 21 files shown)


packages/run-protocol/src/metrics.js
```diff
@@ -0,0 +1,40 @@
...
+import { makeStoredSubscription, makeSubscriptionKit } from '@agoric/notifier';
+import { E } from '@endo/eventual-send';
+
+const PATH = 'metrics';
+
+/**
+ * @template T
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {T[]} optionalInitialState
```
michaelfig: Not specifying an initial state is a behaviour @dtribble is actively trying to stamp out.  The new `makePublisherKit` will not support it.

So, would you please change this not to be optional, and have the call sites always supply it?
```suggestion
 * @param {T} initialState
```

packages/notifier/src/subscriber.js
```diff
@@ -64,6 +64,11 @@ const makeSubscriptionIterator = tailP => {
  * @returns {SubscriptionRecord<T>}
  */
 const makeSubscriptionKit = (...optionalInitialState) => {
+  assert(
+    // optionalInitialState allows just one state and has to be an array to allow specifying 'undefined'
+    optionalInitialState.length === 0 || optionalInitialState.length === 1,
+    'Too many arguments',
+  );
```
michaelfig: I appreciate the idea of not allowing extra arguments, but that's not a conventional behaviour for JS functions.  There are users of this API outside of the SDK, so I don't want to do that.

The comment is valuable, though.  Maybe add it to the JSDoc for this parameter.
```suggestion
  // optionalInitialState allows just one state and has to be an array to allow specifying 'undefined'
```

packages/notifier/src/subscriber.js
```diff
@@ -101,7 +106,7 @@ const makeSubscriptionKit = (...optionalInitialState) => {
     },
   });
 
-  if (optionalInitialState.length > 0) {
+  if (optionalInitialState.length === 1) {
```
michaelfig:
```suggestion
  if (optionalInitialState.length > 0) {
```

packages/run-protocol/src/metrics.js
```diff
@@ -0,0 +1,40 @@
+// @ts-check
+import { makeStoredSubscription, makeSubscriptionKit } from '@agoric/notifier';
+import { E } from '@endo/eventual-send';
+
+const PATH = 'metrics';
```
michaelfig: Maybe make the PATH an optional argument that defaults to undefined (no subpath off of the storageNode).

packages/run-protocol/src/metrics.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @template T
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {T[]} optionalInitialState
+ * @returns {MetricsSubscriptionKit<T>}
+ */
+export const makeMetricsSubscriptionKit = (
```
michaelfig: If we want to make it so that we never have to change the contracts using the helper again, I'd add the following to `packages/notifier/src/storesub.js`:

```js
/**
 * @template T
 * @typedef {object} StoredPublisherKit
 * @property {StoredSubscription<T>} subscriber TODO: change to StoredSubscriber<T> when available
 * @property {IterationObserver<T>} publisher
 */

/**
 * @template T
 * @param {T} initialState
 * @param {ERef<StorageNode>} [storageNode]
 * @param {ERef<Marshaller>} [marshaller]
 * @param {string} [childPath]
 * @returns {StoredPublisherKit<T>}
 */
export const makeStoredPublisherKit ...
```

then import it as follows:

```js
import { makeStoredPublisherKit } from '@agoric/notifier';
```

packages/run-protocol/src/metrics.js
```diff
@@ -0,0 +1,40 @@
...
+  storageNode,
+  marshaller,
+  ...optionalInitialState
+) => {
+  const {
+    publication: metricsPublication,
+    subscription: rawMetricsSubscription,
+  } = makeSubscriptionKit(...optionalInitialState);
+
+  const metricsStorageNode = storageNode && E(storageNode).getChildNode(PATH);
```
michaelfig:
```suggestion
  const metricsStorageNode = storageNode && childPath && E(storageNode).getChildNode(childPath);
```

packages/run-protocol/src/metrics.js
```diff
@@ -0,0 +1,40 @@
...
+  const metricsSubscription = makeStoredSubscription(
+    rawMetricsSubscription,
+    metricsStorageNode,
+    marshaller,
+  );
+
+  return { metricsPublication, metricsSubscription };
```
michaelfig: It's a great helper, and we can make it more general without baking "metrics" in.  Then it can be part of the notifier package, too.

```suggestion
  return { publication, subscription };
```

packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -457,6 +460,11 @@ export const startVaultFactory = async (
     shortfallInvitationAmount,
   });
 
+  const chainStorage = await chainStorageP;
+  const storageNode = await (chainStorage &&
+    E(chainStorage).getChildNode(STORAGE_PATH));
+  const marshaller = E(board).getPublishingMarshaller();
```
michaelfig: I'm on the fence whether you should use `getPublishingMarshaller` here, or just `getReadOnlyMarshaller`. Maybe write a comment as to why it's safe to publish all the ocaps on the board?

michaelfig (commented):
Thanks for tagging me early enough to help influence the direction.

I strongly recommend making just one `makeStoredPublisherKit` in `@agoric/notifier` for future-proofing.

packages/notifier/src/subscriber.js
turadg: Agree it's normal that JS functions allow extra arguments. But they also don't have JSDoc saying that there's a rest parameter of arbitrary length. I agree the comment is best placed at the param declaration. I also think given that comment it's fair to throw when there's a bad input, just like we do in so many other instances. 

to grab one as an example,
```
const assertParamChangeIssue = issue => {
  assert(issue, X`argument to assertParamChangeIssue cannot be null`);
```
this is about the same,
```
  assert(
    optionalInitialState.length === 0 || optionalInitialState.length === 1,
    'optionalInitialState must have zero or one elements',
  );
```
Granted it makes the function fail on extra arguments which most JS functions don't bother to do, but the rest parameter turns the end of the parameter list into a param which can be valid or invalid.

Though I suppose there's a more general principle in JS that "extra" data in calls is valid. Like an object that has extra parameters an array can have extra elements. So I'll take it out.

packages/run-protocol/src/metrics.js
turadg: yeah, that was the next step. I wasn't sure how general purpose to take on. thanks for supporting that it go in `@agoric/notifier`

packages/run-protocol/src/proposals/econ-behaviors.js
turadg: I don't see any uses of `getReadonlyMarshaller` yet. Why doesn't `setupAMM` use that?

For the metrics use case, I think the data will be copy records and not need board objects. Is that right?

dckc:
For testing, it's important to do some integration testing; at least manually, similar to the "demo" under **Testing Considerations** in #5400 .

p.s. important detail: `agd query swingset ...` has become `agd query vstorage ...`

packages/run-protocol/src/metrics.js
turadg: This constraint greatly complicates virtual object initial state because the publisher is part of that state, but would not also require that state defined (minus the publisher) in order to create the publisher.

Why is @dtribble trying to stamp it out? Is there any other reason that `makePublisherKit` will not support it?

dtribble: @erights argued against requiring an initial state. I have arguments for it. I don't recall where we ended up. Much of that is in an IBIS.

dtribble: I think that if initialState is optional, then we should not support it at all. It's better to just explicitly publish if you want

turadg: https://github.com/Agoric/agoric-sdk/pull/5621

packages/run-protocol/src/proposals/econ-behaviors.js
dtribble: AMM had no private state and has dynamic brands, so it's plausible for it to end up causing more brands to get registered. Vaults should use a read only publisher so there's no risk of exposing a new cap.

packages/notifier/src/storesub.js
```diff
@@ -68,9 +69,9 @@ export const makeStoredSubscription = (
   /** @type {StoredSubscription<T>} */
   const storesub = Far('StoredSubscription', {
     getStoreKey: () =>
-      (storageNode ? E(storageNode).getStoreKey() : Promise.resolve({})).then(
```
turadg, line 71: @michaelfig was this resolve/then intentional?

packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2482,6 +2482,25 @@ test('addVaultType: extra, unexpected params', async t => {
   t.true(matches(actual, M.remotable()), 'unexpected params are ignored');
 });
 
+test.todo('storage keys with storageNode/marshaller');
```
turadg, line 2485: @michaelfig are there any tests like this yet? I found `makeFakeStorage` but nothing for marshaller. I guess I don't need that to test the keys but I'm trying to have tests for this stored sub feature overall

packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -338,15 +340,15 @@ export const definePoolKind = (
...
 ) => {
   const poolInit = (liquidityZcfMint, poolSeat, secondaryBrand) => {
     const { brand: liquidityBrand, issuer: liquidityIssuer } =
       liquidityZcfMint.getIssuerRecord();
     const { notifier, updater } = makeNotifierKit();
-    const {
-      publication: metricsPublication,
-      subscription: metricsSubscription,
-    } = makeSubscriptionKit();
+    const { publisher: metricsPublication, subscriber: metricsSubscription } =
```
turadg, line 350: there appears to be a bug now that no childNode is made. I'll try to make a test for that and then fix it.

packages/notifier/src/storesub.js
michaelfig: Yes (though it was unclear).  Deleting this code threw the baby out with the bath water (it no longer adds the subscription to the store key).

If you want to rewrite for clarity, I suggest:
```js
getStoreKey: async () => {
  if (!storageNode) {
    return harden({ subscription });
  }
  const storeKey = await E(storageNode).getStoreKey();
  return harden({ ...storeKey, subscription });
},
```

packages/notifier/src/storesub.js
```diff
@@ -79,3 +80,41 @@ export const makeStoredSubscription = (
...
+ */
+
+/**
+ * @template [T=unknown]
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {string} [childPath]
+ * @returns {StoredPublisherKit<T>}
+ */
+export const makeStoredSubscriptionKit = (
```
michaelfig: Please change the name.

```suggestion
export const makeStoredPublisherKit = (
```

packages/run-protocol/src/contractSupport.js
```diff
@@ -79,3 +80,26 @@ export const checkDebtLimit = (debtLimit, totalDebt, toMint) => {
...
   );
 };
+
+/**
+ * @template T
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @returns {MetricsSubscriptionKit<T>}
+ */
+export const makeMetricsSubscriptionKit = (storageNode, marshaller) => {
```
michaelfig: Do we really need this wrapper?  It changes the API surface in an arbitrary way that I find distracting.

What's wrong with open-coding the call `makeStoredPublisherKit` and making the caller more explicitly request a `'metrics'` child node?

packages/run-protocol/src/contractSupport.js
```diff
@@ -79,3 +80,26 @@ export const checkDebtLimit = (debtLimit, totalDebt, toMint) => {
...
+  return {
+    metricsPublication: kit.publisher,
+    metricsSubscription: kit.subscriber,
+  };
+};
+harden(makeMetricsSubscriptionKit);
+
+/**
+ * @template T
+ * @typedef {object} MetricsSubscriptionKit<T>
```
michaelfig: This is unnecessary, IMO.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -208,14 +213,15 @@ const start = async (zcf, privateArgs) => {
...
-    makeSubscriptionKit();
 
   // shortfall in Vaults due to liquidations less than debt. This value can be
   // reduced by various actions which burn RUN.
   let shortfallBalance = AmountMath.makeEmpty(runBrand);
 
+  /** @type {import('../contractSupport.js').MetricsSubscriptionKit<MetricsNotification>} */
+  const { metricsPublication, metricsSubscription } =
+    makeMetricsSubscriptionKit(privateArgs.storageNode, privateArgs.marshaller);
+
```
michaelfig:
```suggestion
  /** @type {import('../contractSupport.js').MetricsSubscriptionKit<MetricsNotification>} */
  const { publisher: metricsPublisher, subscriber: metricsSubscriber } =
    makeStoredPublisherKit(privateArgs.storageNode, privateArgs.marshaller, 'metrics');

```

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -42,32 +43,40 @@ const { details: X } = assert;
  * collateralTypes: Store<Brand,VaultManager>,
  * directorParamManager: import('@agoric/governance/src/contractGovernance/typedParamManager').TypedParamManager<import('./params.js').VaultDirectorParams>,
  * metricsPublication: IterationObserver<MetricsNotification>
- * metricsSubscription: Subscription<MetricsNotification>
+ * metricsSubscription: StoredSubscription<MetricsNotification>
```
michaelfig, line 46: Please rename `subscription` to `subscriber` before it becomes too entrenched.  Sorry, I should have specified this more clearly in my last review.

```suggestion
 * metricsSubscriber: StoredSubscriber<MetricsNotification>
```

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -107,19 +136,24 @@ const initState = (zcf, directorParamManager, debtMint) => {
 
   const vaultParamManagers = makeScalarMap('brand');
 
-  const { publication: metricsPublication, subscription: metricsSubscription } =
-    makeSubscriptionKit();
+  const { metricsPublication, metricsSubscription } =
+    makeMetricsSubscriptionKit(storageNode, marshaller);
```
michaelfig:
```suggestion
  const { publisher: metricsPublisher, subscriber: metricsSubscriber } =
    makePublisherKit(storageNode, marshaller, 'metrics');
```

michaelfig (changes requested):

packages/notifier/src/storesub.js
turadg: oh of course. I misread it. redoubles my motivation to rewrite for clarity :)

packages/run-protocol/src/contractSupport.js
turadg: > Do we really need this wrapper?

Of course we could repeat the code everywhere instead.

> What's wrong with open-coding the call

Nothing wrong with that. This is to reify a pattern that is used throughout the run-protocol contracts so that `metricsPublication` in the result is always the `metrics` node.

It's in contractSupport which has other convenience methods.

packages/run-protocol/src/contractSupport.js
turadg: Do you mean specifically the typedef or is this comment re-asserting https://github.com/Agoric/agoric-sdk/pull/5613/files/33c4924387a82e068418621a29fad091bdb679e5#r903497158 ?

packages/run-protocol/src/reserve/assetReserve.js
turadg: I think this is the third assertion of your point in https://github.com/Agoric/agoric-sdk/pull/5613/files/33c4924387a82e068418621a29fad091bdb679e5#r903497158 . I'm confused why you're writing out the code here. It was clear enough what you're asking for. I gave my counterpoints there.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: How can it become "too entrenched"? How would it be any less work to do rename now that in a later PR?

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: reading this as fourth https://github.com/Agoric/agoric-sdk/pull/5613/files/33c4924387a82e068418621a29fad091bdb679e5#r903497158

packages/run-protocol/src/vaultFactory/vaultDirector.js
michaelfig: Oh, I shouldn't have recommended changing  `metricsSubscription` since it was already there.  In this case, the only useful part of my comment is changing `StoredSubscription` to `StoredSubscriber`.

packages/run-protocol/src/reserve/assetReserve.js
michaelfig: > I'm confused why you're writing out the code here.

To make it easy to hit `Add suggestion to batch` if you elect to make the change.

packages/run-protocol/src/contractSupport.js
michaelfig: Okay, that's a good enough argument.  I just wanted to make sure that the helper paid its way.

packages/run-protocol/src/vpool-xyk-amm/pool.js
turadg: e9e63340c4f6caa6f186e2fef36aef1fe9faf074

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -249,6 +242,9 @@ const start = async (zcf, privateArgs) => {
     );
     return zcf.getIssuerForBrand(brand);
   };
+  const poolStorageNode = await E(privateArgs.storageNode)?.getChildNode(
+    `pool${secondaryBrandToPool.getSize()}`,
```
turadg: @Chris-Hibbert any concerns with this? I think downstream needs this PR quickly so if you recommend changes here and this lands without them I'll make sure to follow up.

michaelfig (approved):

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
Chris-Hibbert: I presume you're asking about the way of generating a unique name for the child node.   If it were just once, and the names were persistent, I don't think I'd see a problem. The AMM never deletes pools, so this number grows monotonically. 

Are the poolStorageNodes themselves durable? Will we need to be able to recreate or re-find the names or pools on revival?

turadg: > the way of generating a unique name for the child node

👍 

> The AMM never deletes pools, so this number grows monotonically.

Perfect.

> Are the poolStorageNodes themselves durable? Will we need to be able to recreate or re-find the names or pools on revival?

I believe storage nodes are durable and I think we landed on them being part of durable object state. Though they're also cheap to re-create if needed.

Issue #5553, opened 2022-06-09
Agoric wallet crashed when dApp use AssetKind.COPY_BAG


vigh18:
## Describe the bug
Agoric wallet crashed when dApp use AssetKind.COPY_BAG

## To Reproduce
1. Use the main branch of agoric-sdk.
2. Clone agoric example from gateway conference https://github.com/RBFLabs/gateway-agoric-workshop-2022
3. Run agoric install and then skip to instructions for the [react-ui](https://github.com/RBFLabs/gateway-agoric-workshop-2022#react-ui)
4. Accept dApp in wallet and wait until it tries to render dApp-NFTs.Awesomez.

## Expected behavior
Wallet dApp should not crash.

## Platform Environment
OS X, updated, beta branch of agoric-sdk

## Screenshots
<img width="1491" alt="Screenshot 2022-06-09 at 11 54 17" src="https://user-images.githubusercontent.com/14995194/172820151-08d5ab8e-3635-4893-9a2a-e982a5834b75.png">

Tartuffo:
FYI @michaelfig

turadg:
As part of fixing this the UI should be made more robust (needn't crash when UI fails to parse an amount).

Scheduling for MN-1. @samsiegart please estimate.

samsiegart:
The crash should be fixed from #5617, but actual CopyBag support will be implemented in https://github.com/Agoric/agoric-sdk/issues/5383, closing this now.

PR #5617, opened 2022-06-16
fix(wallet-ui): safely render invalid amount values

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/5553

Before: (Blank white page for the entire UI)

After:

Screen Shot 2022-06-19 at 5 24 29 PM
Screen Shot 2022-06-19 at 5 25 08 PM

### DIFF
--- a/packages/wallet/ui/src/components/Offer.jsx
+++ b/packages/wallet/ui/src/components/Offer.jsx
@@ -1,15 +1,14 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import { Nat } from '@agoric/nat';
-import { stringifyPurseValue } from '@agoric/ui-components';
 import Chip from '@mui/material/Chip';
 import CheckIcon from '@mui/icons-material/Check';
 import CloseIcon from '@mui/icons-material/Close';
 import { E } from '@endo/eventual-send';
 import Request from './Request';
 import Petname from './Petname';
-import { icons, defaultIcon } from '../util/Icons.js';
 import { formatDateNow } from '../util/Date';
 import { withApplicationContext } from '../contexts/Application';
+import ErrorBoundary from './ErrorBoundary';
+import Proposal from './Proposal';
 
 import './Offer.scss';
 
@@ -33,21 +32,6 @@ const statusColors = {
   cancel: 'default',
 };
 
-const cmp = (a, b) => {
-  if (a < b) {
-    return -1;
-  }
-  if (a > b) {
-    return 1;
-  }
-  return 0;
-};
-
-const entryTypes = {
-  want: { header: 'Want', move: 'into' },
-  give: { header: 'Give', move: 'from' },
-};
-
 const OfferWithoutContext = ({
   offer,
   pendingOffers,
@@ -60,8 +44,6 @@ const OfferWithoutContext = ({
     instancePetname,
     instanceHandleBoardId,
     requestContext: { date, dappOrigin, origin = 'unknown origin' } = {},
-    proposalForDisplay: { give = {}, want = {}, arguments: args } = {},
-    invitationDetails: { fee, feePursePetname, expiry } = {},
     id,
   } = offer;
   let status = offer.status || 'proposed';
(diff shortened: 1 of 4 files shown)


packages/wallet/ui/src/components/Offer.jsx
```diff
@@ -103,27 +103,35 @@ const OfferWithoutContext = ({
...
+          ? Nat(amount.value)
+          : amount.value;
+      valueString = stringifyPurseValue({
+        value,
+        displayInfo: amount.displayInfo,
+      });
+    } catch {
+      console.warn("Couldn't parse amount value for", pursePetname, amount);
+    }
+
```
turadg: this block is for transforming an amount to a string. I think that would be useful outside this module. It also should be thoroughly tested with different amount cases, including failure. 

```suggestion
    const valueString = stringifyAmount(amount);
```

packages/wallet/ui/src/components/Offer.jsx
```diff
@@ -132,25 +140,32 @@ const OfferWithoutContext = ({
...
+  try {
+    feeValueString =
+      fee &&
+      stringifyPurseValue({
+        value: fee.value,
+        displayInfo: fee.displayInfo,
+      });
+  } catch {
+    console.warn("Couldn't parse fee value", fee);
+  }
```
turadg: isn't this also an amount? why repeat this logic?

if it's for data validation that should happen on the way into application state.

packages/wallet/ui/src/components/tests/Offer.test.jsx
```diff
@@ -128,6 +128,63 @@ test('renders the wants', () => {
   );
   expect(want.text()).toContain('into Zoe fees');
 });
+
+test('renders unknown amounts when parsing fails', () => {
```
turadg: if you had `stringifyAmount` you wouldn't need to do all this React stuff just to be testing that "unknown" is the string for an amount that fails to parse.

packages/wallet/ui/src/components/tests/Offer.test.jsx
```diff
@@ -128,6 +128,63 @@ test('renders the wants', () => {
...
+      },
+      feePursePetname: 'Zoe fees',
+      expiry: 1723014088n,
+    },
+  };
+  const component = mount(<Offer offer={invalidOffer} />);
+
+  const give = component.find('.OfferEntry').at(0);
+
+  expect(give.text()).toContain('unknown');
```
turadg: is this accurate? this appears to the user like the system doesn't know the value. the system does but the UI encountered an error rendering it. We'd want bug reports in that case? If so, something like `<error>` would be better.

packages/wallet/ui/src/components/Offer.jsx
```diff
@@ -103,27 +103,35 @@ const OfferWithoutContext = ({
...
+      const value =
+        amount.displayInfo.assetKind === 'nat'
+          ? Nat(amount.value)
+          : amount.value;
+      valueString = stringifyPurseValue({
+        value,
+        displayInfo: amount.displayInfo,
+      });
+    } catch {
+      console.warn("Couldn't parse amount value for", pursePetname, amount);
```
turadg: some part of UX will be broken (seeing the value) so this is an `error`

turadg (changes requested):

packages/wallet/ui/src/components/tests/Offer.test.jsx
samsiegart: Do you mean wrap each amount in something like https://github.com/Agoric/dapp-treasury/blob/main/ui/src/components/ErrorBoundary.jsx? Then if it fails to parse we'll see a proper error message rather than `unknown`.

packages/wallet/ui/src/components/Offer.jsx
samsiegart: Will fix this up to reuse logic.

I don't think it should be considered as data validation, we're just trying our best to render something we got from the backend, the UI doesn't create this value.

packages/wallet/ui/src/components/tests/Offer.test.jsx
turadg: I meant simply `toContain('<error>')` as the string. an ErrorBoundary is a good idea though, to make the app more robust. I think this particular parsing error made the app crash.

samsiegart:
I redid this PR to make it just use an error boundary instead of special rendering logic (see screenshots in description). PTAL.

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -0,0 +1,115 @@
...
+      return 1;
+    }
+    return 0;
+  };
+
+  const entryTypes = {
+    want: { header: 'Want', move: 'into' },
+    give: { header: 'Give', move: 'from' },
+  };
+
```
turadg: these don't need any of the values of the enclosing function so should be in module scope.

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -0,0 +1,115 @@
...
+            </div>
+            {type.move} <Petname name={pursePetname} />
+          </div>
+        </div>
+      </div>
+    );
+  };
+
+  const Give = entry => OfferEntry(entryTypes.give, entry);
+  const Want = entry => OfferEntry(entryTypes.want, entry);
```
turadg: it looks like OfferEntry can also be module scope, so this can too.

Then consider a helper to give you the values you need on 94 and 95.

```
const gives = offerEntries(offer.proposalForDisplay.give);
const wants = offerEntries(offer.proposalForDisplay.want);
```

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -0,0 +1,115 @@
...
+    give: { header: 'Give', move: 'from' },
+  };
+
+  const {
+    proposalForDisplay: { give = {}, want = {}, arguments: args } = {},
+    invitationDetails: { fee, feePursePetname, expiry } = {},
+  } = offer;
+
+  const gives = Object.entries(give).sort(([kwa], [kwb]) => cmp(kwa, kwb));
+  const wants = Object.entries(want).sort(([kwa], [kwb]) => cmp(kwa, kwb));
```
turadg: Not used until line 94. Don't define until needed, so that the reader has less to keep track of while studying the function.

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -0,0 +1,115 @@
...
+          <h6>Expiry</h6>
+          <div className="Expiry text-gray">
+            {formatDateNow(parseFloat(expiry) * 1000)}
+          </div>
+        </div>
+      )}
+      {args !== undefined && (
+        <div className="OfferEntry">
+          <h6>Arguments</h6>
+          <pre>{JSON.stringify(args, null, 2)}</pre>
```
turadg, line 111: TIL `spacer` argument

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -0,0 +1,115 @@
...
+      {feeEntry}
+      {expiry && (
+        <div className="OfferEntry">
+          <h6>Expiry</h6>
+          <div className="Expiry text-gray">
+            {formatDateNow(parseFloat(expiry) * 1000)}
+          </div>
+        </div>
+      )}
+      {args !== undefined && (
```
turadg, line 108: are other falsy values valid arguments?

turadg (approved):
Good idea to solve the crashing problem more generally with an error boundary.

Noted some non-blocking code quality improvements.

packages/wallet/ui/src/components/Proposal.jsx
samsiegart: They shouldn't be according to this https://agoric.com/documentation/zoe/api/zoe.html#offerargs

However, I think it makes sense to render them for debugging purposes either way.

samsiegart:
I just refactored some functions into the module scope as suggested. Thanks for review!

PR #5620, opened 2022-06-17
fix vaultManager state isolation

turadg:
## Description

Fixes bug introduced in https://github.com/Agoric/agoric-sdk/pull/5214 that made vault managers share some ephemeral state.

Two variables were moved out of virtual state into mere runtime state, but in doing so moved up to module scope.

### Security Considerations

Improved state isolation.

--

### Testing Considerations

Feasible to regression test?


turadg:
the solution is not so simple https://github.com/Agoric/agoric-sdk/issues/5622

PR #5621, opened 2022-06-17
BREAKING CHANGE: no optionalInitialState in makeSubscriptionKit()

turadg:
## Description

There was [some debate](https://github.com/Agoric/agoric-sdk/pull/5613#discussion_r899691493) over whether going forward we would support an optional initial state in `makePublisherKit`. The conclusion was that we should either require an initial state or remove the option.

This PR proposes (and implements) that we remove the option since all it does is call `updateState` for the caller, which they can do whenever they want to. This flexibility make it so a virtual kind can create its state with the publisher in it and then use that state to generate a publication payload.

### Security Considerations

--

Breaking change, noted in conventional commit.

### Testing Considerations

Tests updated.

### DIFF
--- a/packages/notifier/src/subscriber.js
+++ b/packages/notifier/src/subscriber.js
@@ -60,10 +60,9 @@ const makeSubscriptionIterator = tailP => {
  * distributed pub/sub.
  *
  * @template T
- * @param {T[]} optionalInitialState
  * @returns {SubscriptionRecord<T>}
  */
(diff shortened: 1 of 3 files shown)


erights (approved):
LGTM but I was always on this side of the controversy. Please don't merge until we hear @dtribble 's side.

dtribble (approved):

PR #5624, opened 2022-06-17
fix(swingset): fix unhandled rejection in test-liveslots

### LINKED ISSUE #5623: liveslots result-promise-in-args test needs updating
## Describe the bug

@kriskowal mentioned that an Endo change is triggering a failure in `test-liveslots.js` (the test labelled 'result promise in args'). It looks like when I landed the fix for #5106, I changed this test to stop expection that `three()` gets pipelined into the kernel, but I failed to consider where `three()` *does* go. It now gets executed within the vat, and by resolving the promise to data (`'four'`) instead of a callable object with a `.three()` method, that causes an error. The attempted delivery of `three()` might be stalled by a turn or two, and I think maybe the test was finishing before the message could be delivered, concealing the resulting failure.

I think the Endo change might remove a turn from that delay, allowing the error to be revealed, in the form of an unhandled rejection. Or maybe the unhandled rejection was happening all the time, but previously we were ignoring them, and now we aren't.

The best fix is probably to resolve `p` to `target` instead of the string `'four'`. Then we'll expect to see an additional syscall after the `resolve`, a syscall.send with `target: 'o-1', method: 'three'`, and a new result promise. There will also be a syscall.subscribe for that new result promise.

warner:
update a test that was causing an unhandled rejection

closes #5623

### DIFF
--- a/packages/SwingSet/test/test-liveslots.js
+++ b/packages/SwingSet/test/test-liveslots.js
@@ -1578,9 +1578,9 @@ test('result promise in args', async t => {
         // we should be able to pipeline messages to it
         E(p).three();
         // we can subscribe to it, even though we're the decider
-        p.then(res => vatlog.push(`res: ${res}`));
+        p.then(res => vatlog.push(`res: ${res === target}`));
         // and we should be able to resolve it
-        return 'four';
+        return target;
       },
     });
   }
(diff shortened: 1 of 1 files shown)


warner:
cc @kriskowal see if this fixes the problem you're seeing

FUDCo (approved):
Credible.

kriskowal (approved):
I cherry picked this into the Endo sync #5576 and the test now passes.

PR #5625, opened 2022-06-19
chore: tidying up setupBasicMints

erights:
trivial tidying ups of setupBasicMints.js

### DIFF
--- a/packages/zoe/test/unitTests/setupBasicMints.js
+++ b/packages/zoe/test/unitTests/setupBasicMints.js
@@ -5,20 +5,20 @@ import { makeStore } from '@agoric/store';
 import { makeZoeKit } from '../../src/zoeService/zoe.js';
 import { makeFakeVatAdmin } from '../../tools/fakeVatAdmin.js';
 
-const setup = () => {
-  const moolaBundle = makeIssuerKit('moola');
-  const simoleanBundle = makeIssuerKit('simoleans');
-  const bucksBundle = makeIssuerKit('bucks');
-  const allBundles = {
-    moola: moolaBundle,
-    simoleans: simoleanBundle,
-    bucks: bucksBundle,
+export const setup = () => {
+  const moolaKit = makeIssuerKit('moola');
+  const simoleanKit = makeIssuerKit('simoleans');
+  const bucksKit = makeIssuerKit('bucks');
+  const allIssuerKits = {
+    moola: moolaKit,
+    simoleans: simoleanKit,
+    bucks: bucksKit,
   };
   /** @type {Store<string, Brand>} */
   const brands = makeStore('brandName');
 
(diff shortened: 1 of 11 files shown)


packages/zoe/test/unitTests/setupBasicMints.js
```diff
@@ -5,7 +5,7 @@ import { makeStore } from '@agoric/store';
 import { makeZoeKit } from '../../src/zoeService/zoe.js';
 import { makeFakeVatAdmin } from '../../tools/fakeVatAdmin.js';
 
-const setup = () => {
+export const setup = () => {
```
turadg, line 8: I agree exporting at the declaration is better but I don't see it documented. Is there consensus to use it? I'm sure there's a lint rule we could enable.

turadg (commented):
Code changes LGTM. However for merging, the conventional commits are wrong. None of these are fixes. I.e. these changes should bump the patch release of the semver artifacts.

I think you want `chore:`. I'd also suggest landing by squash so it's one commit.

packages/zoe/test/unitTests/setupBasicMints.js
erights: How do we find the lint rule?

turadg: First look in ESilnt's built in rules:
https://eslint.org/docs/rules/
None applicable there. Then consider what plugins might have it. Here are the rules for eslint-plugin-import:
https://github.com/import-js/eslint-plugin-import/tree/main/docs/rules

There are two relevant rules:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/exports-last.md
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/group-exports.md

Unfortunately they're to enforce the opposite of what we want. Often such rules can be configured per some preference to be "always" or "never" but these were designed less flexibly. A PR to make them more general would probably be accepted upstream. Meanwhile we could take take the code to make a custom rule.

I figure that's post MN-1 but if you're confident enough we should have this as a rule then let me know and I'll make a ticket.

erights: When discussing function styles --- `function` vs arrow function vs concise method syntax --- the pattern came up with a general sense of "this is good". But I wouldn't say we walked away with a sense of consensus on a decision. We did have consensus on the big one: arrow functions. The issue of patterns for using them were considered in weighing the tradeoffs.

In any case, I prefer it but I don't feel strongly about it. At such a time as we have a lint rule that could enforce this, we probably should. But until then, for this one, I'd just leave it as a preference rather than a rule.

erights:
Squashed and changed to chore:. PTAL

turadg (approved):

turadg:
> Squashed

btw, the `automerge:squash` label would do that automatically, using the PR title as the commit summary

PR #5626, opened 2022-06-19
test: fix test-vpid-liveslots.js

warner:
The test had a bogus extra call to `becomeDecider()`, with
result=null, that wasn't in the test plan.


FUDCo (approved):
Yep.

PR #5627, opened 2022-06-20
DO NOT MERGE durability mystery

erights:
This is a tiny improvement on https://github.com/Agoric/agoric-sdk/pull/5283 that should have worked. Instead, we get some failures in the durability implementation that seem unrelated. Capturing this separately because it may indicate a bug in the durability implementation.

See https://github.com/Agoric/agoric-sdk/issues/5628


erights:
Closing. If the demonstrated bug is still present, we'd need to start from scratch anyway.

PR #5629, opened 2022-06-20
fix(vats): ensure rejections from loadVat are caught

dckc:
stacked on #5576 

context: https://github.com/Agoric/agoric-sdk/pull/5576#issuecomment-1159821965

I looked at all calls to `E(loadVat)(...)` and made sure rejections are caught by adding an `await` here and there.


michaelfig (approved):
LGTM!

kriskowal:
Rebased into #5576. Thank you 🙏

PR #5636, opened 2022-06-21
Make stopVat do less work

### LINKED ISSUE #5342: stopVat should only reject promises and abandon non-durables

A remaining part of #1848 is to reduce the amount of work we do during `stopVat` to a bare minimum. We're concerned that `stopVat` in vats with large accumulated data will take too long, effectively making it impossible to upgrade such vats (I expect to see consensus problems if it takes more than ~30s).

The bare minimum is:
* reject all pending promises (`stop-vat.js`: `rejectAllPromises()`)
* abandon all non-durable merely-virtual exports (`abandonExports()`)

This will leave data in the DB that is no longer reachable:
* merely-virtual objects
* merely-virtual collections

Those objects might maintain refcounts on durable data and imports, which would otherwise be dropped.

We think this is acceptable for MN-1, because:
* 1: we expect our early contract developers will mostly use durable data, rather than merely-virtual data
  * partly because they need it for upgrade, partly because of this stopVat cleanup limitation
* 2: we believe we have enough of an index to allow `startVat()` (in a new version of liveslots) to perform this cleanup properly, spread out over multiple calls to `bringOutYourDead()`
  * #5090 has some details on the index

This is a compromise, of course: we would prefer to delete everything, completely and precisely, during `stopVat`. We'd *really* prefer for the kernel to delete everything, but I don't think our data structures make that easy to do without some excessive coupling/mututal-understanding between liveslots and the kernel.

## Description of the Design

Change `stop-vat.js` to only call `rejectAllPromises()` and `abandonExports()`.

Eyeball the current virtual-kind index DB keys and convince ourselves they'll be enough to perform the deletion in a later version.

## Security Considerations

Shouldn't be any: abandoning the virtuals should make them inaccessible from the outside world, and the existing prohibition on durable objects/collections holding non-durable children will make them inaccessible from the new version as well.

This does allow a vat to accumulate more DB space over time, by creating a bunch of virtuals and then upgrading. But I think the threat is of the same magnitude as the existing ability to create virtuals within a single version, independent of upgrade.

## Test Plan

There's already a test (`test/upgrade/test-upgrade.js`) which creates an extensive set of data, performs an upgrade, then looks to see what was deleted and what still remains. We just need to modify that test to expect more things to remain.

FUDCo:
Closes #5342

This PR disables a bunch of machinery inside `stopVat` per the specification laid out in #5342

After much study and lengthy pondering I have persuaded myself that it _will_ be possible to perform comparable database cleaning in a post-upgrade process of some kind, details to be determined later.

The disabled code in `stopVat` has been commented out rather than removed, on the theory that later on we'll be glad of that.

### DIFF
--- a/packages/SwingSet/src/liveslots/stop-vat.js
+++ b/packages/SwingSet/src/liveslots/stop-vat.js
@@ -166,6 +166,7 @@ function finalizeEverything(tools) {
   }
 }
 
+// eslint-disable-next-line no-unused-vars
 function deleteVirtualObjectsWithoutDecref({ vrm, syscall }) {
   // delete the data of all non-durable virtual objects, without
   // attempting to decrement the refcounts of the surviving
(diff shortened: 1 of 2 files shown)


warner (approved):
Looks good. I'm sad to see all that intricate test get disabled (omg I worked a long time on that), but when we implement a new/better cleanup mechanism, it should all be ready to turn back on.

Issue #5507, opened 2022-06-03
Avoid doing too much SwingSet work in FinalizeBlock


arirubinstein:
## Describe the bug
With a light simulated load (10 ag-solos making AMM trades), many validators have issues processing the data and participating in consensus, with multiple timeouts and rounds needed to create a block.  It's possible that the default tendermint round timeouts need to be increased to accomodate for the expected load from swingset, as it appears the default limits are being reached and resulting in base+delta*round increases.  Alternatively, it may be required to do less work per block on the swingset side in order to beat the timeout.

## To Reproduce
1. go to https://devnet.explorer.agoric.net/agoric/uptime
2. see large amount of missed blocks on all validators

## Expected behavior
with nominal load, all validators should not have any issues voting on blocks.  

<img width="984" alt="Screen Shot 2022-06-03 at 4 23 51 PM" src="https://user-images.githubusercontent.com/22369/171966232-1fc48857-40c0-46eb-93ba-0f32bc6e5021.png">

warner:
We looked at a run from the last weekend, with the #5152 timestamps in the slog. In one particular block, two (of the five) validators needed a massive 1000s to complete the block. Looking into the slogfiles, we found a 1000s+ delay in two different deliveries (validator 1 had this delay in delivery N, validator 2 had this delay in a different delivery M). In one of these two deliveries, we saw the delay occurred after the slogfile recorded `deliver-result` arriving on the kernel, and before it recorded `crank-finish`:

```
{"type":"syscall-result","crankNum":1877999,"vatID":"v14","deliveryNum":79364,"syscallNum":30,"replay":false,"ksr":["ok",null],"vsr":["ok",null],"time":1655696811.177324,"monotime":184524.70294137407}
{"type":"clist","crankNum":1877999,"mode":"export","vatID":"v14","kobj":"ko53942","vobj":"o+10/12786","time":1655696811.178376,"monotime":184524.70399383403}
{"type":"clist","crankNum":1877999,"mode":"drop","vatID":"v14","kobj":"kp239132","vobj":"p-28033","time":1655696811.178857,"monotime":184524.70447407317}
[...]

PR #5578, opened 2022-06-11
fix: alter slogger so it can't throw errors

FUDCo:
Fixes #5463

Note: this currently does not have any tests.  I was able to verify that it works by introducing errors with a temporary source mod to `controller.js` and then testing manually to eyeball that the appropriate stuff appeared in the console log, but this approach does not lend itself to automation.  However, I haven't as yet figured out a way to construct a test that could be used in CI.  Given the limited scope and relatively gentle nature of the changes as well as the very marginal edge case nature of the condition being dealt with here, it's not obvious to me that this one motivates a significant engineering investment to automate its testing.  That said, if anyone is inspired with a straightforward approach that will do the job, I'm all ears.

PR #5637, opened 2022-06-21
feat(cosmic-swingset): Add memory stats to new commit block slog event

mhofman:
refs: #5507

## Description

- Cleanup various otel spans, handling events that were previously not handled, and removing data that is unactionable raw (will be extended upon in a follow up PR)
- Adds a new `cosmic-swingset-commit-block-*` slog events, with their own timing
  - `commit-block-start` and `commit-block-finish` measure the time taken by cosmic-swingset to save/flush its data.
  - `after-commit-block` is an event containing post block cleanup stats related to node/v8 memory.
- Adds a `NODE_HEAP_SNAPSHOTS` environment option to control the generation of node heap snapshots after a block commit
  - `-1` (or any negative number) to disable, `0` to only trigger on large block interval (hardcoded at 30 seconds), any positive number to generate a snapshot at an interval
  - The snapshot should be generated after node has informed the go side of the block commit completion, which means it should happen while node is idle waiting for the next block (unless the cosmos node is already late)
- Forces a GC after every commit to remove variance based on local GC schedule.

### Security Considerations

This adds an option to dump the node heap, which shouldn't contain any sensitive data as it's all based on the consensus execution.

TBD

### Testing Considerations

Manual local testing with the loadgen deployment test, and ingest slot to otel script.

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1011,7 +1011,12 @@ export default function buildKernel(
   async function processDeliveryMessage(message) {
     kdebug(`processQ ${JSON.stringify(message)}`);
     kdebug(legibilizeMessage(message));
-    kernelSlog.write({ type: 'crank-start', message });
+    kernelSlog.write({
+      type: 'crank-start',
+      crankType: 'delivery',
+      crankNum: kernelKeeper.getCrankNumber(),
+      message,
+    });
     /** @type { PolicyInput } */
     let policyInput = ['none'];
     if (message.type === 'create-vat') {
@@ -1184,7 +1189,12 @@ export default function buildKernel(
   async function processAcceptanceMessage(message) {
     // kdebug(`processAcceptanceQ ${JSON.stringify(message)}`);
     // kdebug(legibilizeMessage(message));
-    kernelSlog.write({ type: 'crank-start', message });
+    kernelSlog.write({
+      type: 'crank-start',
+      crankType: 'routing',
+      crankNum: kernelKeeper.getCrankNumber(),
+      message,
+    });
     /** @type { PolicyInput } */
     const policyInput = ['none'];
 
--- a/packages/cosmic-swingset/src/chain-main.js
+++ b/packages/cosmic-swingset/src/chain-main.js
@@ -1,5 +1,6 @@
 import path from 'path';
 import { resolve as importMetaResolve } from 'import-meta-resolve';
+import { waitUntilQuiescent } from '@agoric/swingset-vat/src/lib-nodejs/waitUntilQuiescent.js';
 import {
   importMailbox,
   exportMailbox,
@@ -401,6 +402,14 @@ export default async function main(progname, args, { env, homedir, agcc }) {
     const keepSnapshots =
       XSNAP_KEEP_SNAPSHOTS === '1' || XSNAP_KEEP_SNAPSHOTS === 'true';
 
+    const afterCommitCallback = async () => {
+      // delay until all current promise reactions are drained so we can be sure
+      // that the commit-block reply has been sent to agcc through replier.resolve
+      await waitUntilQuiescent();
+
+      return {};
+    };
+
     const s = await launch({
       actionQueue,
       kernelStateDBDir: stateDBDir,
@@ -416,6 +425,7 @@ export default async function main(progname, args, { env, homedir, agcc }) {
       mapSize,
       swingStoreTraceFile,
       keepSnapshots,
+      afterCommitCallback,
     });
     return s;
   }
--- a/packages/cosmic-swingset/src/chain-main.js
+++ b/packages/cosmic-swingset/src/chain-main.js
@@ -1,6 +1,7 @@
 import path from 'path';
 import v8 from 'node:v8';
 import process from 'node:process';
+import fs from 'node:fs';
 import { performance } from 'perf_hooks';
 import { resolve as importMetaResolve } from 'import-meta-resolve';
 import engineGC from '@agoric/swingset-vat/src/lib-nodejs/engine-gc.js';
@@ -376,6 +377,7 @@ export default async function main(progname, args, { env, homedir, agcc }) {
       LMDB_MAP_SIZE,
       SWING_STORE_TRACE,
       XSNAP_KEEP_SNAPSHOTS,
+      NODE_HEAP_SNAPSHOTS = -1,
     } = env;
     const slogSender = await makeSlogSenderFromModule(SLOGSENDER, {
       stateDir: stateDBDir,
@@ -406,10 +408,23 @@ export default async function main(progname, args, { env, homedir, agcc }) {
     const keepSnapshots =
       XSNAP_KEEP_SNAPSHOTS === '1' || XSNAP_KEEP_SNAPSHOTS === 'true';
 
+    const nodeHeapSnapshots = Number.parseInt(NODE_HEAP_SNAPSHOTS, 10);
+
+    let lastCommitTime = 0;
+    let commitCallsSinceLastSnapshot = NaN;
+    const snapshotBaseDir = path.resolve(stateDBDir, 'node-heap-snapshots');
+
+    if (nodeHeapSnapshots >= 0) {
+      fs.mkdirSync(snapshotBaseDir, { recursive: true });
+    }
+
     const afterCommitCallback = async () => {
       // delay until all current promise reactions are drained so we can be sure
       // that the commit-block reply has been sent to agcc through replier.resolve
       await waitUntilQuiescent();
+
+      let heapSnapshot;
+      let heapSnapshotTime;
 
       const t0 = performance.now();
       engineGC();
@@ -419,13 +434,29 @@ export default async function main(progname, args, { env, homedir, agcc }) {
       const heapStats = v8.getHeapStatistics();
       const t3 = performance.now();
 
+      commitCallsSinceLastSnapshot += 1;
+      if (
+        (nodeHeapSnapshots >= 0 && t0 - lastCommitTime > 30 * 1000) ||
+        (nodeHeapSnapshots > 0 &&
+          !(nodeHeapSnapshots > commitCallsSinceLastSnapshot))
+      ) {
+        commitCallsSinceLastSnapshot = 0;
+        heapSnapshot = `Heap-${process.pid}-${Date.now()}.heapsnapshot`;
+        const snapshotPath = path.resolve(snapshotBaseDir, heapSnapshot);
+        v8.writeHeapSnapshot(snapshotPath);
+        heapSnapshotTime = performance.now() - t3;
+      }
+      lastCommitTime = t0;
+
       return {
         memoryUsage,
         heapStats,
+        heapSnapshot,
         statsTime: {
           forcedGc: t1 - t0,
           memoryUsage: t2 - t1,
           heapStats: t3 - t2,
+          heapSnapshot: heapSnapshotTime,
         },
       };
     };
--- a/packages/cosmic-swingset/src/chain-main.js
+++ b/packages/cosmic-swingset/src/chain-main.js
@@ -1,4 +1,6 @@
 import path from 'path';
+import v8 from 'node:v8';
+import process from 'node:process';
 import { performance } from 'perf_hooks';
 import { resolve as importMetaResolve } from 'import-meta-resolve';
 import engineGC from '@agoric/swingset-vat/src/lib-nodejs/engine-gc.js';
@@ -412,10 +414,18 @@ export default async function main(progname, args, { env, homedir, agcc }) {
       const t0 = performance.now();
       engineGC();
       const t1 = performance.now();
+      const memoryUsage = process.memoryUsage();
+      const t2 = performance.now();
+      const heapStats = v8.getHeapStatistics();
+      const t3 = performance.now();
 
       return {
+        memoryUsage,
+        heapStats,
         statsTime: {
           forcedGc: t1 - t0,
+          memoryUsage: t2 - t1,
+          heapStats: t3 - t2,
         },
       };
     };
--- a/packages/cosmic-swingset/src/chain-main.js
+++ b/packages/cosmic-swingset/src/chain-main.js
@@ -1,5 +1,7 @@
 import path from 'path';
+import { performance } from 'perf_hooks';
 import { resolve as importMetaResolve } from 'import-meta-resolve';
+import engineGC from '@agoric/swingset-vat/src/lib-nodejs/engine-gc.js';
 import { waitUntilQuiescent } from '@agoric/swingset-vat/src/lib-nodejs/waitUntilQuiescent.js';
 import {
   importMailbox,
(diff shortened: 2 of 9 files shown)


michaelfig (approved):
LGTM!  Nice and tidy!  Please rerequest a review when you're ready for it.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -561,6 +561,11 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         spans.end(`begin-block`);
         break;
       }
+      case 'cosmic-swingset-commit-block': {
+        spans.top().setAttributes(cleanAttrs(slogAttrs));
+        spans.pop(`block ${slogAttrs.blockHeight}`);
```
mhofman: I forgot to highlight, but I believe this will not dramatically change the `block ${blockHeight}` span. Previously, the span would be ended with the time of the last/previous named pop, which is `end-block` finish. This effectively will include the time between end-block-finish and commit-block in the block time where it was previously excluded.

Furthermore the `setAttributes` call assumes `top()` is the block span, but that's only enforced in the subsequent `pop` call.  Technically if the layering of spans change in the future but not everything is updated accordingly, we may end up with adding attributes to another span, and only "throw" on the `pop` which simply results in console error output since #5578

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -401,6 +408,65 @@ export default async function main(progname, args, { env, homedir, agcc }) {
...
+      return {
+        heapStats,
+        heapSpaceStats,
+        memoryUsage,
+        heapSnapshot,
+        statsTime: {
+          heapStats: t1 - t0,
+          heapSpaceStats: t2 - t1,
+          memoryUsage: t3 - t2,
+        },
```
mhofman: This is so we can check that gathering these memory stats does not significantly delay the commit. I originally wanted to measure the commit time itself so we can compare the value, but it didn't feel right to plumb `performance` into `launch-chain`

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -401,6 +408,65 @@ export default async function main(progname, args, { env, homedir, agcc }) {
...
+      };
+    };
+
+    const afterCommitCallback = async () => {
+      await waitUntilQuiescent();
+
+      engineGC();
```
mhofman: This is technically a behavior change forcing a full GC after committing every block.

packages/telemetry/src/slog-to-otel.js
warner: As soon as I figure out `slog-to-otel.js`, I'm planning to add a bunch of spans (about three per syscall), but they should all be children of the crank/delivery span, not the block span, so I don't think that will change the layering you describe.

packages/cosmic-swingset/src/chain-main.js
warner: But it's only affecting the behavior of the kernel, not the workers, and the kernel doesn't use WeakMaps or WeakRefs or anything that would allow GC to affect visible behavior (it should only affect timing).

The `waitUntilQuiescent()` is kinda weird in a kernel context, though. It makes sense in a worker because we deny `setImmediate()`/etc to userspace, so there's a distinction between the kind of code that loses agency before this fires, and the kind of code that *could* retain agency (but doesn't, because it's all written by us). I'm not sure what it means in the kernel world.

Nothing in the kernel/controller uses `setImmediate()`, but all the vat managers use IO, and `waitUntilQuiescent` will fire before IO returns. I think I'd want to make sure this isn't run until after `controller.run()` has returned (at which point all the workers should be idle, and no V8 IO should be happening). Given that, it might manage to delay the GC until after some lingering promises are running. But my hunch is that it'd be better/simpler/no-worse to just call `engineGC()` without the `waitUntilQuiescent()`.

mhofman: > The `waitUntilQuiescent()` is kinda weird in a kernel context, though

It's to make sure that all reactions to inform the go side have been processed before we start this work. But I agree the wording is not right (only the implementation using `setImmediate`)

> But my hunch is that it'd be better/simpler/no-worse to just call `engineGC()` without the `waitUntilQuiescent()`.

Nope, there are some promise reactions that need to happen before we do `replier.resolve()` in `fromGo`. This is basically what the `setImmediate` is for here, make sure all these promise reactions (which have no dependency on I/O) are effectively over. I actually considered using `process.nextTick()` since that uses a parallel queue to the promise queue of equal priority, but that isn't really a guarantee in future versions of node.

mhofman: > But it's only affecting the behavior of the kernel, not the workers, and the kernel doesn't use WeakMaps or WeakRefs or anything that would allow GC to affect visible behavior (it should only affect timing).

Correct this is only a timing change for the kernel.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -66,11 +77,7 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
   let currentAttrs = {};
 
   const cleanAttrs = attrs => ({
-    ...Object.fromEntries(
-      Object.entries(attrs)
-        .map(([key, value]) => [`agoric.${key}`, cleanValue(value, key)])
-        .filter(([_key, value]) => value !== undefined && value !== null),
-    ),
+    ...serializeInto(attrs, 'agoric'),
```
mhofman, line 80: @michaelfig this does cause some data to be serialized deeply, such as `slots` being written as `.slots.0=ko123` etc instead of `.slots=["ko123"]`

michaelfig: That's perfectly fine, IMHO.

warner: That sounds like it makes the number of attributes per (span?) become larger and variable.. if some crank uses a thousand slots, is that going to cause problems on the telemetry provider?

While it's nice to be able to recover the list of slots, I'm hard pressed to imagine a honeycomb-side search that could use it (slogfile-processing tools are probably more useful there). Unless it's possible to do a honeycomb search like "find me all spans WHERE any `.slots.*` attributes have value `ko123`", which would be kinda cool for causality tracing (currently I do this with a lot of `jq` work and/or ad-hoc python tools).

mhofman: Right, a base `slots` which is a comma separated string of values may be more useful, and could be used with `contains` in Honeycomb queries

packages/telemetry/src/slog-to-otel.js
```diff
@@ -106,8 +113,9 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         delete attrs.msg;
         const { methargs, method = extractMethod(methargs) } = message.msg;
         name = `E(${message.target}).${method}`;
-        attrs['message.msg.args.slots'] =
-          message.msg.methargs?.slots ?? message.msg.args.slots;
+        const slots =
+          message.msg.methargs?.slots ?? message.msg.args.slots ?? [];
+        attrs['message.msg.args.slots'] = slots.join(',');
```
mhofman, line 126: @warner is there a risk that `,` would ever be used in krefs ?

packages/telemetry/src/slog-to-otel.js
mhofman: I explicitly handled `slots` below to avoid the splitting in separate fields.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -552,15 +561,31 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         break;
       }
       case 'cosmic-swingset-begin-block': {
+        if (spans.topName() === `intra-block`) {
+          spans.pop(`intra-block`);
+        }
```
mhofman, line 585: This introduces a new span to capture the duration between "commit-block" and the next "begin-block", and layers that as a child of the "previous" block, which causes the block's duration to now measure the block to block time.

Before we were not capturing this gap anywhere, causing blocks to appear shorter than they really were.

cc @mfig @warner @arirubinstein

packages/cosmic-swingset/src/chain-main.js
mhofman: We now measure the commit time by emitting both a start and a finish for the commit-block event.

packages/telemetry/src/slog-to-otel.js
mhofman: Discussed offline, will be handled in a follow-up PR.

mhofman:
@michaelfig @warner, PTAL (recommend reviewing commit-by-commit)

I've rebased and simplified the commit-block slog events. (sorry for nullifying previous reviews)

I've also included some preliminary cleanup to the slog-to-otel in advance of a more thorough rewrite of spans.

packages/telemetry/src/slog-to-otel.js
warner: Not anytime soon, I have a bunch of analysis tools that also use `,` as a joiner. *vrefs* are more interesting, but krefs have no punctuation (purely `k[opd]\d+`) and no plans to acquire any.

packages/swing-store/src/snapStore.js
```diff
@@ -230,23 +234,27 @@ export function makeSnapStore(
     toDelete.add(hash);
   }
 
-  function commitDeletes(ignoreErrors = false) {
+  async function commitDeletes(ignoreErrors = false) {
```
warner, line 245: looks sound, although I don't get why an async version is better, and why these changes are part of the PR (this doesn't change the swingstore API, right?, just some churn in the necessary IO endowments?)

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1011,7 +1011,12 @@ export default function buildKernel(
   async function processDeliveryMessage(message) {
     kdebug(`processQ ${JSON.stringify(message)}`);
     kdebug(legibilizeMessage(message));
-    kernelSlog.write({ type: 'crank-start', message });
+    kernelSlog.write({
```
warner, line 1014: I was thinking of using `delivery-crank-start` and `routing-crank-start` instead of a separate field (likewise `delivery-crank-finish`/`routing-crank-finish`). I figured that my `jq` tools would just grow a `select(.type=="delivery-crank-start")` instead of `select(.type=="crank-start" and .crankType=="delivery")`, slightly easier to type, and it's rare that I'd want to pay attention to both types at the same time.

Also, I'm kinda wavering on the idea that the routing step should be called a "crank" (might be better to find a new name, maybe), and having separate `type:` values would make that easier to change.

OTOH if they share a `crankNum` numberspace then having separate `type` values is a tiny bit weird.

Ok leave it with `crankType` for now, but let's think about it for later.

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -272,10 +273,32 @@ export async function launch({
...
+        controller.writeSlogObject({
+          type: 'cosmic-swingset-after-commit-block',
+          blockHeight: savedHeight,
+          blockTime: savedBlockTime,
+          ...afterCommitStats,
+        });
+      });
+
+    controller.writeSlogObject({
+      type: 'cosmic-swingset-commit-block-finish',
```
warner, line 298: For the `type:` name, be aware that the `commit()` this wraps is only the swingset/swing-store commit: the rest of cosmos-sdk does not start it's IAVL commit until after this returns. The name is probably ok (we can pretend that the `cosmic-swingset-` prefix implies that we're only talking about the swingset host app's commit of swingset state), but if we ever find a way to grab the golang/IAVL commit timing and write it to the slog, we might need to find a more precise name.

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -401,6 +408,59 @@ export default async function main(progname, args, { env, homedir, agcc }) {
...
+    const afterCommitCallback = async () => {
+      // delay until all current promise reactions are drained so we can be sure
+      // that the commit-block reply has been sent to agcc through replier.resolve
+      await waitUntilQuiescent();
+
+      let heapSnapshot;
+      let heapSnapshotTime;
+
+      const t0 = performance.now();
+      engineGC();
```
warner, line 430: Ok, so this is forcing a Node.js GC every block.. normally I'd be concerned about interfering with the carefully-designed V8 GC algorithm (we shouldn't presume to know better), but given that this is only happening every 5-10 seconds, and we're measuring how much overhead we're introducing, I'm ok with it.

But let's pay attention to the results, and be willing to remove it if it appears to be overhead that could be handled better through the normal incremental GC. I see that `afterCommitCallback` implies that this might happen in parallel with golang-side work, but I never believe parallelism actually happens without proof.

packages/telemetry/src/ingest-slog-entrypoint.js
```diff
@@ -78,10 +78,11 @@ async function run() {
 
   console.log(`parsing`, slogFileName);
 
+  let update = false;
   for await (const line of lines) {
     lineCount += 1;
     const obj = JSON.parse(line);
-    const update = obj.time >= progress.lastSlogTime;
+    update ||= obj.time >= progress.lastSlogTime;
```
warner, line 85: Huh, what prompted this? Did slogfiles have non-monotonically-increasing timestamps?

I see that there's no way for `update` to ever become `false` again (within a single run of this program), but that's ok, as the feature is meant to allow the program to be restarted after a crash and pick up where it left off.

packages/telemetry/src/slog-to-otel.js
warner: I guess the reason that this code is only ever seeing krefs is that it'd fed by `crank-start` events, rather than `type: 'deliver'` events? If it were the latter, something would need to extract the kernel-format delivery (`.kd`), and I don't see that here.

warner (approved):
Ok, I think that's good. I can't visualize the collection of spans and how they line up, so I'm not confident that I understand how this changes the spans, but I know you've got another change in the works which will include docs and a list/diagram of spans, so I'll re-examine this code once those docs are available as a reference.

packages/swing-store/src/snapStore.js
mhofman: It was needed in an earlier version to allow executing JS while the delete I/Os were taking place. This version is strictly superiors though since it parallelizes the deletes.

packages/SwingSet/src/kernel/kernel.js
mhofman: Yeah the shared numbering space was one motivation for the decision. The other being that some "delivery" cranks still do not deliver anything. I was also concerned about breaking existing slog tooling which targets `crank-start`.

packages/cosmic-swingset/src/launch-chain.js
mhofman: Right, I expect we'll need to overall the layering of spans if/when we ever manage to report data from the go side.

packages/cosmic-swingset/src/chain-main.js
mhofman: Correct, that's why I added the measurement compared to the previous iteration.
I couldn't 100% confirm that it happened in parallel with the go side, but from what I could tell, it did.

packages/telemetry/src/ingest-slog-entrypoint.js
mhofman: I had some weird case where some slog lines were skipped, and this was the only explanation I found. I unfortunately didn't save the slog file.

packages/telemetry/src/slog-to-otel.js
mhofman: I plan of re-thinking the crank vs delivery slog events.

packages/swing-store/src/snapStore.js
mhofman: And you're correct, at this point it became a drive by change.

michaelfig (approved):
LGTM!  Nice work, Mathieu.

PR #5641, opened 2022-06-22
Enable collection deletion without swapping in key objects

### LINKED ISSUE #5053: collection manager `clearInternal` instantiates keys during deletion

FUDCo:
If collection clear is unconditional, allow clearing out the entire contents without deserializing the keys.

Fixes #5053

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -10,6 +10,7 @@ import {
   zeroPad,
   makeEncodePassable,
   makeDecodePassable,
+  isEncodedRemotable,
   makeCopySet,
   makeCopyMap,
 } from '@agoric/store';
@@ -64,6 +65,10 @@ const MAX_DBKEY_LENGTH = 220;
 
 function pattEq(p1, p2) {
   return compareRank(p1, p2) === 0;
+}
+
+function matchAny(patt) {
+  return patt === undefined || pattEq(patt, M.any());
 }
 
 function throwNotDurable(value, slotIndex, serializedValue) {
@@ -271,8 +276,10 @@ export function makeCollectionManager(
     // the resulting function will encode only `Key` arguments.
     const encodeKey = makeEncodePassable({ encodeRemotable });
 
+    const vrefFromDBKey = dbKey => dbKey.substring(BIGINT_TAG_LEN + 2);
+
     const decodeRemotable = encodedKey =>
-      convertSlotToVal(encodedKey.substring(BIGINT_TAG_LEN + 2));
+      convertSlotToVal(vrefFromDBKey(encodedKey));
 
     // `makeDecodePassable` has three named options:
     // `decodeRemotable`, `decodeError`, and `decodePromise`.
@@ -459,8 +466,8 @@ export function makeCollectionManager(
       let priorDBKey = '';
       const start = prefix(coverStart);
       const end = prefix(coverEnd);
-      const ignoreKeys = !needKeys && pattEq(keyPatt, M.any());
-      const ignoreValues = !needValues && pattEq(valuePatt, M.any());
+      const ignoreKeys = !needKeys && matchAny(keyPatt);
+      const ignoreValues = !needValues && matchAny(valuePatt);
       /**
        * @yields {[any, any]}
        * @returns {Generator<[any, any], void, unknown>}
@@ -516,10 +523,60 @@ export function makeCollectionManager(
       return iter();
     }
 
+    /**
+     * Clear the entire contents of a collection non-selectively.  Since we are
+     * being unconditional, we don't need to inspect any of the keys to decide
+     * what to do and therefor can avoid deserializing the keys.  In particular,
+     * this avoids swapping in any virtual objects that were used as keys, which
+     * can needlessly thrash the virtual object cache when an entire collection
+     * is being deleted.
+     *
+     * @returns {boolean} true if this operation introduces a potential
+     *   opportunity to do further GC.
+     */
+    function clearInternalFull() {
+      let doMoreGC = false;
+      const [coverStart, coverEnd] = getRankCover(M.any(), encodeKey);
+      let priorDBKey = '';
+      const start = prefix(coverStart);
+      const end = prefix(coverEnd);
+      while (priorDBKey !== undefined) {
+        const [dbKey, dbValue] = syscall.vatstoreGetAfter(
+          priorDBKey,
+          start,
+          end,
+        );
+        if (!dbKey) {
+          break;
+        }
+        if (dbKey < end) {
+          priorDBKey = dbKey;
+          const value = JSON.parse(dbValue);
+          doMoreGC =
+            value.slots.map(vrm.removeReachableVref).some(b => b) || doMoreGC;
+          syscall.vatstoreDelete(dbKey);
+          if (isEncodedRemotable(dbKey)) {
+            const keyVref = vrefFromDBKey(dbKey);
+            if (hasWeakKeys) {
+              vrm.removeRecognizableVref(keyVref, `${collectionID}`, true);
+            } else {
+              doMoreGC = vrm.removeReachableVref(keyVref) || doMoreGC;
+            }
+            syscall.vatstoreDelete(prefix(`|${keyVref}`));
+          }
+        }
+      }
+      return doMoreGC;
+    }
+
     function clearInternal(isDeleting, keyPatt, valuePatt) {
       let doMoreGC = false;
-      for (const k of keys(keyPatt, valuePatt)) {
-        doMoreGC = doMoreGC || deleteInternal(k);
+      if (isDeleting || (matchAny(keyPatt) && matchAny(valuePatt))) {
+        doMoreGC = clearInternalFull();
+      } else {
+        for (const k of keys(keyPatt, valuePatt)) {
+          doMoreGC = deleteInternal(k) || doMoreGC;
+        }
       }
       if (!hasWeakKeys && !isDeleting) {
         syscall.vatstoreSet(prefix('|entryCount'), '0');
@@ -559,10 +616,7 @@ export function makeCollectionManager(
     }
 
     function getSize(keyPatt, valuePatt) {
-      if (
-        (keyPatt === undefined || pattEq(keyPatt, M.any())) &&
-        (valuePatt === undefined || pattEq(valuePatt, M.any()))
-      ) {
+      if (matchAny(keyPatt) && matchAny(valuePatt)) {
         return Number.parseInt(syscall.vatstoreGet(prefix('|entryCount')), 10);
       }
       return countEntries(keyPatt, valuePatt);
--- a/packages/SwingSet/test/gc-helpers.js
+++ b/packages/SwingSet/test/gc-helpers.js
@@ -350,10 +350,6 @@ export function validateDeleteMetadataOnly(
         refValString(contentRef, contentType),
       ]),
     );
-    validate(
-      v,
-      matchVatstoreGet(`vc.${idx}.sfoo`, refValString(contentRef, contentType)),
-    );
     if (!nonVirtual) {
       validateUpdate(v, `vom.rc.${contentRef}`, `${rc}`, `${rc - 1}`);
     }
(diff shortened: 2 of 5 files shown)


packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -509,10 +515,62 @@ export function makeCollectionManager(
...
+    }
+
     function clearInternal(isDeleting, keyPatt, valuePatt) {
+      if (isDeleting || (matchAny(keyPatt) && matchAny(valuePatt))) {
+        return clearInternalFull();
+      }
       let doMoreGC = false;
       for (const k of keys(keyPatt, valuePatt)) {
-        doMoreGC = doMoreGC || deleteInternal(k);
+        doMoreGC = deleteInternal(k) || doMoreGC;
```
warner: oh good, thank you for catching that bug, I didn't notice it myself until just now

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -509,10 +515,62 @@ export function makeCollectionManager(
...
+              vrm.removeRecognizableVref(keyVref, `${collectionID}`, true);
+            } else {
+              doMoreGC = vrm.removeReachableVref(keyVref) || doMoreGC;
+            }
+            syscall.vatstoreDelete(prefix(`|${keyVref}`));
+          }
+        }
+      }
+      if (!hasWeakKeys) {
+        syscall.vatstoreSet(prefix('|entryCount'), '0');
```
warner: This adds a vatstoreSet to the "delete the whole collection" case that wasn't there before, right? That seems like a tiny regression.

What do you think about having the caller (`clearInternal`) retain responsibility for writing out this length=0, so it can continue to use the same `if (!hasWeakKeys && !isDeleting)` logic as before? In that case, `clearInternal()` would do something like:
```js
{
  let doMoreGC = false;
  if (isDeleting || (matchAny(keyPatt) && matchAny(valuePatt))) {
    doMoreGC = clearInternalFull();
  } else {
    for (const k of keys(keyPatt, valuePatt)) {
      doMoreGC = deleteInternal(k) || doMoreGC;
    }
  }
  if (!hasWeakKeys && !isDeleting) {
...
```

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -509,10 +515,62 @@ export function makeCollectionManager(
...
+        if (!dbKey) {
+          break;
+        }
+        if (dbKey < end) {
+          priorDBKey = dbKey;
+          const value = JSON.parse(dbValue);
+          doMoreGC =
+            value.slots.map(vrm.removeReachableVref).some(b => b) || doMoreGC;
+          syscall.vatstoreDelete(dbKey);
+          if (dbKey[0] === 'r') {
```
warner: This is coupling to the behavior of.. `packages/store/src/patterns/encodePassable.js > makeEncodePassable()`? I'm a bit nervous about bitrot if `store` decided to change their encoding scheme.

I obviously don't want to call `decodeRemotable()` on `dbKey`, that would defeat the purpose of this fix. 

One option would be to create a separate instance of `makeDecodePassable()` whose only job is to report "reference or not", probably by counting whether the `decodeRemotable` passed in was called or not:
```js
function dbKeyIsRemotable(dbKey) {
  let isRemotable = false;
  function decodeRemotable(rem) { isRemotable = true; return undefined; }
  const decoder = makeDecodePassable({decodeRemotable});
  decoder(dbKey);
  return isRemotable;
}
```

Another option is to add a unit test which explicitly checks the output of `makeEncodePassable` to see that the encoded key starts with `'r'`, and to add cross-pointing comments between the test and this line (so `store` ever changes its behavior, our test will fail, and when someone goes to update the test, they'll following the pointer here and update the `=== 'r'` comparison at the same time).

A third option (the most efficient and maintainable) would be to change `store` to export an additional function like:
```js
export const encodedIsRemotable = (encoded) => encoded[0] === 'r';
harden(encodedIsRemotable);
```

plus a unit test in `store`. Not sure what @erights would think about that.

I'm undecided as to which I like better, but I know I want at least one of them.

packages/SwingSet/test/gc-helpers.js
```diff
@@ -375,11 +371,12 @@ export function validateDeleteMetadataOnly(
   }
   let priorKey = '';
   if (entries >= 0) {
+    validate(v, matchVatstoreSet(`vc.${idx}.|entryCount`, `0`));
```
warner: I think this would go away if we avoided the "tiny regression" extra length=0 set, right?

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -233,12 +233,12 @@ const testUpgrade = async (t, defaultManagerType) => {
 
   for (let i = 1; i < NUM_SENSORS + 1; i += 1) {
     const vref = durVref(i);
-    const impKref = impKrefs[i];
+    // const impKref = impKrefs[i];
```
warner, line 261: Does the test also pass if you instead just delete the chunk of `survivingDurables.push(10); ...` above? I tried to arrange the test to let us delete adjustments as we fixed the bugs that necessitated them, and commenting out all of these checks makes the tests less strict than they need to be.

(of course #5636 disables this entire swath of checks, depending upon the order in which they land, but it would be nice if they were as accurate as possible for the day in the future that we reenable them)

warner (changes requested):
probably-small changes to make, but looks good overall

packages/SwingSet/src/liveslots/collectionManager.js
FUDCo: That makes sense.

packages/SwingSet/src/liveslots/collectionManager.js
FUDCo: My first reaction was that this ship had already sailed, as we have various other code that has baked in understanding of the encoding, but closer inspection suggests that's not so (there _is_ some code like that but it's already inside the `store` package where it belongs).  Of the three options you propose, I think the first is gross and the second is icky, so I lean towards the third.  There's already all kinds of other whacky stuff exported from `store` so I don't expect @erights to object too loudly.  In the interest of moving things forward I'm inclined to follow the "ask for forgiveness rather than permission" strategy.

packages/SwingSet/test/gc-helpers.js
FUDCo: Yep.

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: No.  I tried fiddling with the test as you suggest here but either I'm not understand what you're proposing or what you're proposing doesn't work.  I don't really understand what this test is actually doing, so...

Unless you have a more concrete thing to try, I'm going to leave this bit as is.

warner: bummer, ok, when you merge this with the #5636 commenting-out, maybe add a comment to the line 266 `t.is(reachable, expI)` line just saying `disabled as part of #5641` so we can backtrack this to the changes that necessitated it

at some point I might try to walk through this test and see how much can be re-enabled

warner (approved):
looks good!

PR #5645, opened 2022-06-22
cleanup: use details template appropriately in asserts

### LINKED ISSUE #5630: minimum liquidity error is missing data
## Describe the bug

The error for adding a pool without enough liquidity has an opaque error message:

```
    message: 'The minimum initial liquidity is [object Object], rejecting [object Object]',
```

Reported by:   @bperry

## To Reproduce
Add a pool to the AMM with too little liquidity.

## Expected behavior
Report of the amount information through the normal secure error reporting path.

## Platform Environment
audit release

## Additional context
The assert does not include the `details` (`X`) template:
```
    assert(
      AmountMath.isGTE(centralAmount, minPoolLiquidity),
      `The minimum initial liquidity is ${minPoolLiquidity}, rejecting ${centralAmount}`,
    );
```

Please ensure that other asserts include the `details` template.

Chris-Hibbert:
closes: #5630

## Description

We noticed that some asserts were using backquotes with embedded expressions, but had omitted the details template (abbreviated `X`).  Here we repair omissions of `details` and unneeded uses. The commits are separated by package.

I only looked at ERTP, Zoe, Governance, and run-protocol.

### Security Considerations

implications of missing things in reviews. Error messages missing intended details and not excluding others that should have been hidden.

None

### Testing Considerations

Updated tests.

It would be nice to have a lint check that would catch these.


packages/ERTP/src/paymentLedger.js
```diff
@@ -1,6 +1,5 @@
 // @ts-check
 
-import { assert, details as X } from '@agoric/assert';
```
turadg, line 3: Using `assert` from global instead of import seems like a regression to me. If this is the recommended style, where is it documented?

packages/governance/test/unitTests/test-buildParamManager.js
```diff
@@ -99,7 +99,7 @@ test('Amount', async t => {
   await t.throwsAsync(
     () => paramManager.updateParams({ Shimmer: 'fear,loathing' }),
     {
-      message: 'Expected an Amount for Shimmer, got "fear,loathing"',
+      message: 'Expected an Amount for "Shimmer", got ""fear,loathing""',
```
turadg: why are the quotes doubled? (or quadrupled as it were)

packages/governance/src/contractGovernor.js
```diff
@@ -173,7 +173,7 @@ const start = async zcf => {
     return poserFacet;
   };
   await getUpdatedPoserFacet();
-  assert(poserFacet, X`question poser facet must be initialized`);
+  assert(poserFacet, 'question poser facet must be initialized');
```
turadg, line 176: would your lint rule require that details templates not be used when the expression is a string literal?

turadg (commented):
No objections but withholding review until I understand why `assert` is no longer imported and whether it's worth removing the superfluous `X`.

Tangentially,
> It would be nice to have a lint check that would catch these.

I'm willing to right that up. Rule being as follows?
- `assert` second param must be string literal or string template
- if it's string template it must be `X` and it must have sub-expressions

packages/governance/test/unitTests/test-buildParamManager.js
Chris-Hibbert: The amount is usually something other than a string, so the error message includes quotes to set them off for distinctness. Whoever is substituting the string from the details template thinks they should have quotes. I agree with the latter, and don't mind double-doublequotes before and after.

packages/governance/src/contractGovernor.js
Chris-Hibbert: I wouldn't require that, since sometimes the message is still being massaged, and is likely to gain or lose details. But it seemed for stable messages, dropping the X occasionally meant one fewer import, which seemed a win.  ... Although I also made it not imported, so maybe that doesn't matter at all.

Chris-Hibbert:
> Rule being as follows?
> assert second param must be string literal or string template
> if it's string template it must be X and it must have sub-expressions

Yep. If any of those are not the case, that seems like a bug to me.

erights:
> No objections but withholding review until I understand why `assert` is no longer imported 

We now have conversations in so many places that I don't know where you and I already discussed this ;)

In any case, IIRC, when we first tried to have only the global, we could not figure out how to propagate the typing and scoping knowledge to inform our tools how to regard `assert`. Since then we've fixed that. (Though, for reasons I don't understand, we have not quite fixed it for `harden`.)

For the sake of avoiding redundancy, we should not have both a global and a module. For reasons that frankly I cannot remember, we needed to make `assert` a global. In any case, we're not about to reverse that decision unless it is causing a significant problem.

The `assert` module is in the agoric-sdk repository, and so cannot be used from the endo repository. We could migrate it, but I'd rather just kill it.

> and whether it's worth removing the superfluous `X`.

Why superfluous?

> Tangentially,
> 
> > It would be nice to have a lint check that would catch these.
> 
> I'm willing to right that up. Rule being as follows?
> 
> * `assert` second param must be string literal or string template
> * if it's string template it must be `X` and it must have sub-expressions

I am uncomfortable insisting that "it must have sub-expressions" but could probably be talked into it.

The rest of this rule looks good, thanks!

packages/ERTP/src/paymentLedger.js
erights: I agree it needs to be documented. Who wants to take this?

I know you didn't suggest it, but to be explicit, I do not think we should require this PR to provide that documentation.

erights: With those answered, I'm resolving this conversation. As always, if there's more to discuss, please unresolve it.

packages/governance/src/contractGovernor.js
erights: > I wouldn't require that, since sometimes the message is still being massaged, and is likely to gain or lose details.

Good! That's a nice rationale for not including the "it must have sub-expressions" that I mentioned I'm uncomfortable with below.

Resolving.

packages/ERTP/src/paymentLedger.js
```diff
@@ -179,7 +180,7 @@ export const makePaymentLedger = (
       payments.forEach(payment => {
         assert(
           !antiAliasingStore.has(payment),
-          `same payment ${payment} seen twice`,
+          X`same payment ${payment} seen twice`,
```
erights, line 183: In your PR comment, under "Security Considerations" you say only

"implications of missing things in reviews. Error messages missing intended details"

But this example right here demonstrates the huge security that was accidentally lost when the `X` was omitted, and that you're repairing by restoring the `X`.

We assume someone who can read the log output has a similar level of privilege as someone who can operate a debugger. So revealing the actual payment to them is good. We should not skimp on the diagnostic info we provide to help programmers figure out bugs.

But the actual payment amount should generally not be revealed to callers, or to others than may come to have access to the error object being thrown, or to the remote copies of that error object as it is serialized and unserialized during communications between mutually suspicious vats or machines. Likewise with the stack. The default[1] behavior of the line above is to disclose the payment and the stack to the log if the error is logged or serialized, but to redact this information from callers or other vats or machines.

@turadg @Chris-Hibbert @dckc , this one actually is documented ;)
See https://github.com/endojs/endo/tree/master/packages/ses/src/error/README.md

Only suggested change is to the PR comment.

[1] This default is normally loosened during development. During testing by imports such as
```js
import { test } from '@agoric/zoe/tools/prepare-test-env-ava.js';
```
Otherwise by, for example, environment variable settings like 
```sh
export LOCKDOWN_ERROR_TAMING=unsafe
```
See https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md#errortaming-options

packages/governance/src/contractGovernance/governApi.js
```diff
@@ -100,7 +100,7 @@ const setupApiGovernance = async (
         if (keyEQ(positive, outcome)) {
           assert(
             keyEQ(outcome, harden({ apiMethodName, methodArgs })),
-            `The question's method name (${apiMethodName}) and args (${methodArgs}) didn't match the outcome ${outcome}`,
+            X`The question's method name (${apiMethodName}) and args (${methodArgs}) didn't match the outcome ${outcome}`,
```
erights: I'll use this as an example of the opposite principle. In general, relatively static info like `apiMethodName` does not need to be redacted. By not redacting it, it is visible with much less trouble in more locations, further easing debugging. That's what `q` is for. Since `q` also quotes it, I removed the extra parens. 

```suggestion
            X`The question's method name ${q(apiMethodName)} and args (${methodArgs}) didn't match the outcome ${outcome}`,
```

Assumes
```js
```

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -76,7 +76,7 @@ const makeParamManagerBuilder = zoe => {
   const buildCopyParam = (name, value, assertion, type) => {
     let current;
     assertKeywordName(name);
-    assert(value !== undefined, `param ${name} must be defined`);
+    assert(value !== undefined, X`param ${name} must be defined`);
```
erights:
```suggestion
    assert(value !== undefined, X`param ${q(name)} must be defined`);
```

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -115,7 +115,7 @@ const makeParamManagerBuilder = zoe => {
   /** @type {(name: string, value: Amount, builder: ParamManagerBuilder) => ParamManagerBuilder} */
   const addAmount = (name, value, builder) => {
     const assertAmount = a => {
-      assert(a.brand, `Expected an Amount for ${name}, got "${a}"`);
+      assert(a.brand, X`Expected an Amount for ${name}, got "${a}"`);
```
erights:
```suggestion
      assert(a.brand, X`Expected an Amount for ${q(name)}, got: ${a}`);
```

packages/governance/src/contractGovernance/paramManager.js
```diff
@@ -270,7 +270,7 @@ const makeParamManagerBuilder = zoe => {
   /** @type {(name: string, value: Invitation, builder: ParamManagerBuilder) => Promise<ParamManagerBuilder>} */
   const addInvitation = async (name, value, builder) => {
     assertKeywordName(name);
-    assert(value !== null, `param ${name} must be defined`);
+    assert(value !== null, X`param ${name} must be defined`);
```
erights:
```suggestion
    assert(value !== null, X`param ${q(name)} must be defined`);
```

packages/governance/src/contractGovernance/typedParamManager.js
```diff
@@ -106,7 +108,7 @@ const makeParamManagerSync = spec => {
 
   for (const [name, [type, value]] of Object.entries(spec)) {
     const add = builder[builderMethodName(type)];
-    assert(add, `No builder method for param type '${type}'`);
+    assert(add, X`No builder method for param type '${type}'`);
```
erights:
```suggestion
    assert(add, X`No builder method for param type ${q(type)}`);
```

packages/governance/test/unitTests/test-buildParamManager.js
erights: That's fine. But what I usually do for such cases at the end of a diagnostic string is to set it off with a colon rather than quotes. I do not have any strong preference though.
```js
      assert(a.brand, X`Expected an Amount for ${q(name)}, got: ${a}`);
```

Oh, one more thing. For this ava error message testing, I have come to prefer regexps rather than literal strings, so that can be more tolerant of variations extraneous to the purpose of the test. For *example*, depending on the purpose of the test, this might be

```suggestion
      message: /Expected an Amount for .*, got .*/,
```

Without knowing more about the purpose of this test, I am *not* suggesting this. Just illustrating for when it is helpful.

packages/run-protocol/src/proposals/demoIssuers.js
```diff
@@ -543,7 +543,7 @@ export const fundAMM = async ({
         );
         assert(secondaryPayment, X`no payment for ${q(issuerName)}`);
 
-        assert(kit.issuer, `No issuer for ${issuerName}`);
+        assert(kit.issuer, X`No issuer for ${issuerName}`);
```
erights:
```suggestion
        assert(kit.issuer, X`No issuer for ${q(issuerName)}`);
```
I see the assert on line 544 likewise decided to disclose this, which seems reasonable.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -95,7 +97,7 @@ const start = async (zcf, privateArgs) => {
     trace('addIssuer', { brand, keyword });
     assert(
       keyword !== RunKW && brand !== runKit.brand,
-      `${RunKW} is a special case handled by the reserve contract`,
+      X`${RunKW} is a special case handled by the reserve contract`,
```
erights:
```suggestion
      X`${q(RunKW)} is a special case handled by the reserve contract`,
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -157,7 +159,7 @@ const start = async (zcf, privateArgs) => {
   const addIssuerFromAmm = async ammSecondaryBrand => {
     assert(
       ammSecondaryBrand !== runKit.brand,
-      `${RunKW} is a special case handled by the reserve contract`,
+      X`${RunKW} is a special case handled by the reserve contract`,
```
erights:
```suggestion
      X`${q(RunKW)} is a special case handled by the reserve contract`,
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -172,7 +174,7 @@ const start = async (zcf, privateArgs) => {
   const getKeywordForBrand = brand => {
     assert(
       keywordForBrand.has(brand),
-      `Issuer not defined for brand ${brand}; first call addIssuer()`,
+      X`Issuer not defined for brand ${brand}; first call addIssuer()`,
```
erights:
```suggestion
      X`Issuer not defined for brand ${q(brand)}; first call addIssuer()`,
```
This one is less clear, but I would disclose the brand.

packages/run-protocol/src/runStake/runStakeKit.js
```diff
@@ -293,7 +293,7 @@ const helperBehavior = {
     );
     assert(
       giveRUNonly || AmountMath.isGTE(newMaxDebt, newDebt),
-      `cannot borrow ${q(newDebt)} against ${q(amountLiened)}; max is ${q(
+      X`cannot borrow ${q(newDebt)} against ${q(amountLiened)}; max is ${q(
```
erights: These disclose actual amounts, which we should generally consider sensitive.

```suggestion
      X`cannot borrow ${newDebt} against ${amountLiened}; max is ${
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -205,7 +204,7 @@ const helperBehavior = {
     const validNewPhases = validTransitions[phase];
     assert(
       validNewPhases.includes(newPhase),
-      `Vault cannot transition from ${phase} to ${newPhase}`,
+      X`Vault cannot transition from ${phase} to ${newPhase}`,
```
erights: I'm guessing a phase is a string constant from an enum. If so...

```suggestion
      X`Vault cannot transition from ${q(phase)} to ${q(newPhase)}`,
```

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -259,7 +259,7 @@ const machineBehavior = {
     // We create only one vault per collateralType.
     assert(
       !collateralTypes.has(collateralBrand),
-      `Collateral brand ${collateralBrand} has already been added`,
+      X`Collateral brand ${collateralBrand} has already been added`,
```
erights:
```suggestion
      X`Collateral brand ${q(collateralBrand)} has already been added`,
```

packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
```diff
@@ -30,7 +32,7 @@ const makeMakeAddLiquidityInvitation = (zcf, getPool) => {
     const liquidityBrand = zcf.getBrandForIssuer(liquidityIssuer);
     assert(
       seat.getProposal().want.Liquidity.brand === liquidityBrand,
-      `liquidity brand must be ${liquidityBrand}`,
+      X`liquidity brand must be ${liquidityBrand}`,
```
erights:
```suggestion
      X`liquidity brand must be ${q(liquidityBrand)}`,
```

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
```diff
@@ -471,7 +471,7 @@ test('add wrong liquidity', async t => {
   );
 
   await t.throwsAsync(() => E(addLiquiditySeatBreaking).getOfferResult(), {
-    message: /liquidity brand must be \[object Alleged: MoolaLiquidity brand]/,
+    message: /liquidity brand must be "\[Alleged: MoolaLiquidity brand]"/,
```
erights:
```suggestion
    message: /liquidity brand must be "\[Alleged: MoolaLiquidity brand\]"/,
```
I find it bizarre that the close square bracket did not need to be escaped.

packages/zoe/src/contracts/coveredCall.js
```diff
@@ -77,7 +78,7 @@ const start = zcf => {
     const sellSeatExitRule = sellSeat.getProposal().exit;
     assert(
       isAfterDeadlineExitRule(sellSeatExitRule),
-      `the seller must have an afterDeadline exitRule, but instead had ${sellSeatExitRule}`,
+      X`the seller must have an afterDeadline exitRule, but instead had ${sellSeatExitRule}`,
```
erights:
```suggestion
      X`the seller must have an afterDeadline exitRule, but instead had ${q(sellSeatExitRule)}`,
```

packages/zoe/tools/manualTimer.js
```diff
@@ -31,12 +31,12 @@ export default function buildManualTimer(log, startValue = 0n, timeStep = 1n) {
     assert.typeof(delay, 'bigint');
     assert(
       delay % timeStep === 0n,
-      `timer has a resolution of ${timeStep}; ${delay} is not divisible`,
+      X`timer has a resolution of ${timeStep}; ${delay} is not divisible`,
```
erights, line 34: Just pointing out that this one is also arguable, but I would also choose to redact these time values, as you've done here. So no change suggested.

erights (changes requested):

packages/governance/src/contractGovernance/governApi.js
turadg: > That's what `q` is for.

I had misunderstood what it's for. Where is documentation on how to use `quote`? It's not mentioned in https://github.com/endojs/endo/blob/master/packages/ses/src/error/README.md  . With some searching I found https://github.com/endojs/endo/blob/db98546be8fda763daea27ac0513953200604ace/packages/ses/src/error/types.js#L166-L171 but that JSDoc isn't coming up for documenting the `quote` function uses. I suppose due to defining it as a callback type.

turadg:
Replying to @erights 
> > No objections but withholding review until I understand why `assert` is no longer imported
> 
> We now have conversations in so many places that I don't know where you and I already discussed this ;)

Let's take it to https://github.com/Agoric/agoric-sdk/issues/5672

> > and whether it's worth removing the superfluous `X`.
> 
> Why superfluous?

In the case I was referring to there was no need for a string template.

> > > It would be nice to have a lint check that would catch these.
…
> The rest of this rule looks good, thanks!

Recorded as https://github.com/endojs/endo/issues/1226

packages/governance/src/contractGovernance/governApi.js
erights: My choices of declaring callbacks vs declaring in place were done in ignorance. Would be overjoyed to see them moved around to follow some useful principle.

packages/governance/src/contractGovernance/paramManager.js

packages/governance/src/contractGovernance/paramManager.js

packages/governance/src/contractGovernance/paramManager.js

packages/governance/src/contractGovernance/typedParamManager.js
Chris-Hibbert: done

packages/governance/src/contractGovernance/governApi.js
Chris-Hibbert: done

packages/run-protocol/src/proposals/demoIssuers.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/assetReserve.js

packages/run-protocol/src/reserve/assetReserve.js

packages/run-protocol/src/reserve/assetReserve.js

packages/run-protocol/src/runStake/runStakeKit.js
Chris-Hibbert: done

packages/run-protocol/src/vaultFactory/vault.js
Chris-Hibbert: done

packages/run-protocol/src/vaultFactory/vaultDirector.js
Chris-Hibbert: done

packages/run-protocol/src/vpool-xyk-amm/addLiquidity.js
Chris-Hibbert: done

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
Chris-Hibbert: JS R.E.s are consistent. If there's no open square bracket, close square bracket doesn't need to be escaped.

This is from experiment. I haven't found a place where the docs admit this.

packages/zoe/src/contracts/coveredCall.js
Chris-Hibbert: done

packages/ERTP/src/paymentLedger.js
Chris-Hibbert: done

turadg (approved):
🧹 👏

packages/run-protocol/test/amm/vpool-xyk-amm/test-liquidity.js
erights: I would still prefer to see the backslash before the close square bracket. It is at least as correct, and its correctness depends only on more local information.

erights (approved):
LGTM, thanks!

Chris-Hibbert:
> I would still prefer to see the backslash before the close square bracket. It is at least as correct, and its correctness depends only on more local information.

Done.

PR #5650, opened 2022-06-22
fix(vault): use singleton pattern for  vaultManager

### LINKED ISSUE #5622: Ephemeral state in vaultManager is incorrectly module-scope
## Describe the bug
Introduced in https://github.com/Agoric/agoric-sdk/pull/5214

In [vaultmanager.js](https://github.com/Agoric/agoric-sdk/blob/master/packages/run-protocol/src/vaultFactory/vaultManager.js#L132), the EPHEMERAL STATE is supposed to be per vault manager (embedded in the scope of a `makeVaultManagerfunction`), but instead it’s module scoped!

## To Reproduce
I think the consequences would be:
- with multiple collateral types
- there would only be a single liquidation process for the first collateral
- all other collaterals would update the quote for their own collateral type
- that would either fail or would confuse the quote)

## Expected behavior
Each vault manager has it's own ephemeral state.

## Additional context
The code is complicated by the durability work.

## Solutions
1. Refactor to use the "singleton durable object" pattern for vault managers, so that it is inside the `makeVaultManager` function. 
2. Use a module-level weak map for ephemeral state
3. Make it durable -- can't do this because it have promises and other non-durable objects

dtribble:
closes: #5622

## Description

Implement ephemeral state by using singleton pattern

### Security Considerations

TBD

### Testing Considerations

TBD

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -93,7 +93,7 @@ const trace = makeTracer('VM');
  * poolIncrementSeat: ZCFSeat,
  * priceAuthority: ERef<PriceAuthority>,
  * prioritizedVaults: ReturnType<typeof makePrioritizedVaults>,
- * zcf: import('./vaultFactory.js').VaultFactoryZCF,
+ * zcf: VaultFactoryZCF,
  * }>} ImmutableState
  */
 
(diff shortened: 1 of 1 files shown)


Chris-Hibbert (commented):
Hmm.   That's definitely simpler, and closer to our usual JS style, but it seems further from the common pattern for our durable objects, so I'm not sure the simplicity is an advantage.

erights:
Note to reviewers: Hide whitespace differences before reviewing the diffs. Bigger help here than usual.

Chris-Hibbert (approved):

turadg:
Closing because I understand @dtribble and @erights concluded we should not use the singleton pattern for vaultManager.

PR #5651, opened 2022-06-23
Enforce size limits on syscalls containing capdata

### LINKED ISSUE #3242: Unrestricted length of VAT Slot IDs
_Surfaced from @informalsystems audit of Agoric/agoric-sdk/SwingSet hash 9feea16732944e01749feedc7f6fb024af4fc70d_

_Notice: As agreed with Agoric, we performed the analysis under the assumption of possible bugs in the liveslots implementation. In that case, userspace vat code together with the liveslots layer operate as a whole against the kernel, which is the necessary condition for triggering this issue. Correctly functioning liveslots implementation should prevent this issue from being exploitable. We will perform assessment of liveslots correctness in the subsequent phases of our analysis._

Currently the kernel accepts VAT Slot IDs as BigInts. E.g., in the function [parseVatSlot](https://github.com/Agoric/agoric-sdk/blob/9feea16732944e01749feedc7f6fb024af4fc70d/packages/SwingSet/src/parseVatSlots.js#L36):

```js
  if (delim > 0) {
    assert(type === 'object' && allocatedByVat, X`invalid vatSlot ${s}`);
    virtual = true;
    id = Nat(BigInt(idSuffix.substr(0, delim)));
    subid = Nat(BigInt(idSuffix.slice(delim + 1)));
  } else {
    id = Nat(BigInt(idSuffix));
  }
```

 As BigInts don't have any limit, it should be possible to kill the kernel with a single syscall, by passing a very-very large number, that would eat all the memory when trying to convert from a string representation into BigInt representation.

 **Recommendation**: Restrict the size of VAT slot IDs to some sufficiently large number, e.g. to 32 bits. As an easier to enforce measure, the string representation of the number could be restricted to e.g. 10 characters.

FUDCo:
Per #3242 these changes add limits to the size of the `body` string and `slots` array of capdata parameters to the syscalls that contain capdata (`send`, `callNow`, and `resolve`).  In the case of `send` and `callNow`, overflowing either of these limits throws an exception.  In the case of `resolve` it forces an error exit on the offending vat.

Closes #3242

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -26,20 +26,6 @@ const DEFAULT_VIRTUAL_OBJECT_CACHE_SIZE = 3; // XXX ridiculously small value to 
 const SYSCALL_CAPDATA_BODY_SIZE_LIMIT = 10_000_000;
 const SYSCALL_CAPDATA_SLOTS_LENGTH_LIMIT = 10_000;
 
-let syscallCapdataBodySizeLimit = SYSCALL_CAPDATA_BODY_SIZE_LIMIT;
-let syscallCapdataSlotsLengthLimit = SYSCALL_CAPDATA_SLOTS_LENGTH_LIMIT;
-
-function assertAcceptableSyscallCapdataSize(capdata) {
-  assert(
-    capdata.body.length < syscallCapdataBodySizeLimit,
-    'syscall capdata body too large',
-  );
-  assert(
-    capdata.slots.length < syscallCapdataSlotsLengthLimit,
-    'syscall capdata slots array too long',
-  );
-}
-
 // 'makeLiveSlots' is a dispatcher which uses javascript Maps to keep track
 // of local objects which have been exported. These cannot be persisted
 // beyond the runtime of the javascript environment, so this mechanism is not
@@ -85,6 +71,31 @@ function build(
   let didStopVat = false;
 
   const outstandingProxies = new WeakSet();
+
+  let syscallCapdataBodySizeLimit = SYSCALL_CAPDATA_BODY_SIZE_LIMIT;
+  let syscallCapdataSlotsLengthLimit = SYSCALL_CAPDATA_SLOTS_LENGTH_LIMIT;
+
+  function setSyscallCapdataLimits(
+    bodySizeLimit = SYSCALL_CAPDATA_BODY_SIZE_LIMIT,
+    slotsLengthLimit = SYSCALL_CAPDATA_SLOTS_LENGTH_LIMIT,
+  ) {
+    syscallCapdataBodySizeLimit = bodySizeLimit;
+    syscallCapdataSlotsLengthLimit = slotsLengthLimit;
+  }
+
+  function assertAcceptableSyscallCapdataSize(capdatas) {
+    let bodySizeTotal = 0;
+    let slotsLengthTotal = 0;
+    for (const capdata of capdatas) {
+      bodySizeTotal += capdata.body.length;
+      slotsLengthTotal += capdata.slots.length;
+    }
+    assert(
+      bodySizeTotal <= syscallCapdataBodySizeLimit &&
+        slotsLengthTotal <= syscallCapdataSlotsLengthLimit,
+      'syscall capdata too large',
+    );
+  }
 
   /**
    * Translation and tracking tables to map in-vat object/promise references
@@ -858,7 +869,7 @@ function build(
 
     meterControl.assertIsMetered(); // else userspace getters could escape
     const serMethargs = m.serialize(harden(methargs));
-    assertAcceptableSyscallCapdataSize(serMethargs);
+    assertAcceptableSyscallCapdataSize([serMethargs]);
     serMethargs.slots.map(retainExportedVref);
 
     const resultVPID = allocatePromiseID();
@@ -924,9 +935,10 @@ function build(
     const resolutions = resolutionCollector().forSlots(serMethargs.slots);
     if (resolutions.length > 0) {
       try {
-        resolutions.forEach(([_xvpid, _isReject, resolutionCD]) => {
-          assertAcceptableSyscallCapdataSize(resolutionCD);
-        });
+        const resolutionCDs = resolutions.map(
+          ([_xvpid, _isReject, resolutionCD]) => resolutionCD,
+        );
+        assertAcceptableSyscallCapdataSize(resolutionCDs);
       } catch (e) {
         syscall.exit(true, e);
         return null;
@@ -979,7 +991,7 @@ function build(
         return (...args) => {
           meterControl.assertIsMetered(); // userspace getters shouldn't escape
           const serArgs = m.serialize(harden(args));
-          assertAcceptableSyscallCapdataSize(serArgs);
+          assertAcceptableSyscallCapdataSize([serArgs]);
           serArgs.slots.map(retainExportedVref);
           // if we didn't forbid promises, we'd need to
           // maybeExportPromise() here
@@ -1124,9 +1136,10 @@ function build(
       const rc = resolutionCollector();
       const resolutions = rc.forPromise(vpid, isReject, value);
       try {
-        resolutions.forEach(([_xvpid, _isReject, resolutionCD]) => {
-          assertAcceptableSyscallCapdataSize(resolutionCD);
-        });
+        const resolutionCDs = resolutions.map(
+          ([_xvpid, _isReject, resolutionCD]) => resolutionCD,
+        );
+        assertAcceptableSyscallCapdataSize(resolutionCDs);
       } catch (e) {
         syscall.exit(true, e);
         return;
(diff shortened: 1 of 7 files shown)


packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -863,6 +880,7 @@ function build(
     lsdebug(
       `ls.qm send(${JSON.stringify(targetSlot)}, ${legibilizeMethod(prop)}) -> ${resultVPID}`,
     );
+    assertAcceptableSyscallCapdataSize(serMethargs);
```
warner: I think we should do this check immediately after the `m.serialize()` above, because if we make it this far and then bail, we'll call `retainExportedVref` and create a new (tracked) result promise and VPID, all of which will be leaked. If we bail earlier, we avoid that leak.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -905,6 +923,14 @@ function build(
     const maybeNewVPIDs = new Set(serMethargs.slots);
     const resolutions = resolutionCollector().forSlots(serMethargs.slots);
     if (resolutions.length > 0) {
+      try {
+        resolutions.forEach(([_xvpid, _isReject, resolutionCD]) => {
+          assertAcceptableSyscallCapdataSize(resolutionCD);
+        });
+      } catch (e) {
+        syscall.exit(true, e);
```
warner: hm, I wonder if we can give them any extra hints here, a `console.log` of something, so the author can figure out why their vat just suddenly self-destructs

The VPID might be a clue for someone looking deeply enough at the slogs. Some prefix of the `resolutionCD` might be a hint. But neither one should really be revealed to userspace, so we'd have to `console.log` it rather than making the assertion function include the hint in the Error object (which is visible to the `send()` callers). And unless it's in the Error object, this `forEach` pattern would lose access to the hint by the time it reaches the `syscall.exit`.

Hrmph. Well, I guess the slogs will still show the last delivery, and any previous syscalls that were successful. And maybe a replay of the fatal delivery might let them learn more. Still, let's keep thinking about ways to make this failure mode more ergonomic.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -905,6 +923,14 @@ function build(
     const maybeNewVPIDs = new Set(serMethargs.slots);
     const resolutions = resolutionCollector().forSlots(serMethargs.slots);
     if (resolutions.length > 0) {
+      try {
```
warner, line 945: We should do the size check *before* calling `syscall.resolve`, otherwise we're making it easy for userspace to trigger the netstring size limit (which will be a kernel panic, rather than an in-consensus fatal vat error).

Also, we can simplify things a little bit because it's safe to build up `maybeNewVPIDs` before the call to `syscall.resolve` (it's just a local Set).

So move the `resolutions.forEach(.. maybeNewVPIDs.add)` to the line before `syscall.resolve()`, and change it to do both the `add` and the `try(assert..) catch {exit..}` clause inside the loop. For tidyness, we can move the second loop (the `maybeNewVPIDs.delete` one) to before the `syscall.resolve` too. Neither has any externally-visible side-effects by this point in the function.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -957,6 +983,7 @@ function build(
           // if we didn't forbid promises, we'd need to
           // maybeExportPromise() here
           forbidPromises(serArgs);
+          assertAcceptableSyscallCapdataSize(serArgs);
```
warner: As above, we should do the checks before the `retainExportedVref`. In fact, please move the `forbidPromises()` test (but not the commented-out `maybeExportPromise()` bit) up a line too, all the tests should happen before we commit to state changes like `retainExportedVref`, to minimize the potential for leaks.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1096,6 +1123,14 @@ function build(
       assert(exportedVPIDs.has(vpid), vpid);
       const rc = resolutionCollector();
       const resolutions = rc.forPromise(vpid, isReject, value);
+      try {
+        resolutions.forEach(([_xvpid, _isReject, resolutionCD]) => {
```
warner: if we wind up with multiple copies of this `forEach` pattern, maybe we should have a helper function like `checkResolutionSize(resolutions, syscall.exit)`

packages/SwingSet/test/test-liveslots.js
```diff
@@ -1394,7 +1597,7 @@ test('promise cycle', async t => {
   t.deepEqual(log, []);
 });
 
-test('unserializable promise resolution', async t => {
+test.serial('unserializable promise resolution', async t => {
```
warner: do these need to be serial?

packages/SwingSet/test/test-liveslots.js
```diff
@@ -590,6 +597,202 @@ test('remote function call', async t => {
...
+    type: 'exit',
+    isFailure: true,
+    info: Error('syscall capdata slots array too long'),
+  });
+  t.deepEqual(log, []);
+
+  const rp8 = 'p-8';
+  await dispatch(makeMessage(rootA, 'returnBodyTooBig', [], [], rp8));
+  t.deepEqual(log.shift(), {
+    type: 'exit',
```
warner, line 968: Poor thing, it's trying so hard to exit, and we just keep ignoring it over and over again..

warner (changes requested):
some implementation changes to make

I skimmed the kernel meeeting recording, but couldn't find+remember our conclusion.. @erights, you're still ok with this `syscall.exit(error: true)` upon oversized resolution, right?

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: I'm slightly less concerned about this since, realistically, I expect this error case to arise very rarely if ever.

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: I'm confused by this comment.  We _do_ do the size check before calling `syscall.resolve`.  Are you possibly mixing up the two calls to `resolutions.forEach` with each other when reading the code?

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: We have two.

packages/SwingSet/test/test-liveslots.js
FUDCo: I'm not 100% sure _why_, but yes.  If you run the tests together these tests now fail without the `.serial` modifier, but if you run them individually they succeed.

packages/SwingSet/test/test-liveslots.js
FUDCo: I tried to raise my kids to be polite to robots since someday it will be harder to tell the difference between robots and people and anyway the robots eventually will be running everything so it's best not to get in the habit of making them angry, but maybe this is a case of "do as I say, not as I do."

packages/SwingSet/test/test-liveslots.js
warner: Ah, I found it, the new test uses `setSyscallCapdataLimits()` to reduce the size limit to an artificially low value, and those limits are kept in a module-level `let` binding so they affect all callers in the same process (and it doesn't set the size back again after the test is done). That limit is low enough to cause the other tests to fail from a "capdata too big" error instead of the other errors they were trying to exercise.

As supporting evidence, running `yarn test -s test/test-liveslots.js` also fails (the `-s` effectively adds `.serial` to every `test()` in the file). I'm guessing that if you only mark a few tests as `.serial`, they get run first, so these two tests were getting to run before the size limit was reduced.

We need those size limits to be per-liveslots-instance instead of being at module level. Just move the `let`s and the checker function inside the definition of `makeLiveSlots()`.

warner (changes requested):

packages/SwingSet/test/test-liveslots.js
FUDCo: Fixed the scoping.

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: Checking the multiple capdatas of a resolve collectively undoes the benefit of the proposed refactoring, since we have to iterate the resolution collection to extract the capdatas as an array, so I'm skipping this for now.  You might want to take a look when you review the revisions to see if you spot a cleaner way that I didn't spot.

warner (approved):
I think this looks good. We aren't enforcing limits on `vatPowers.exitVat`, but I'll make another ticket for it and we can clean it up later, so we can get this landed today.

@erights when you get a chance, I'd like you to think about our handling of oversize errors during `resolve()`. I happened to look at that code more carefully, and realized that:
* if you `resolve()` with an unserializable value, like `{ foo() { ..} }` without the `Far()` wrapper, it acts as if you did `reject(Error('unserializable'))` instead (using the Error object thrown by `marshaller.serialize()`), and the vat keeps running
* but if you `resolve()` with an oversized value, the vat exits (with something like `Error('capdata too big')`)

I'm not sure what to think about that variation: whether size limits are sufficiently different than what are effectively type errors (so it's fine to have two different responses), or if they should both be replaced with rejections, or if they should both kill the vat.

PR #5655, opened 2022-06-23
Guard netstring decode with an optional max size

FUDCo:
Per #3242 this puts a length check on the netstring decoder associated with the streams from worker processes to the kernel.

For now I've set this to 12,000,000 on the theory that given the capdata limits implemented in #5651 of 10,000,000 bytes for the body string plus 10,000 slot entries, overflowing this would require a full body string plus 10,000 slots with vrefs averaging 200 bytes each.  I think this is unlikely to happen by accident.

### DIFF
--- a/packages/SwingSet/src/lib-nodejs/spawnSubprocessWorker.js
+++ b/packages/SwingSet/src/lib-nodejs/spawnSubprocessWorker.js
@@ -23,6 +23,8 @@ function parentLog(first, ...args) {
 // always be Node.
 const stdio = harden(['inherit', 'inherit', 'inherit', 'pipe', 'pipe']);
 
+const NETSTRING_MAX_CHUNK_SIZE = 12_000_000;
+
 export function startSubprocessWorker(execPath, procArgs = []) {
   const proc = spawn(execPath, procArgs, { stdio });
 
@@ -30,7 +32,7 @@ export function startSubprocessWorker(execPath, procArgs = []) {
   toChild.pipe(netstringEncoderStream()).pipe(proc.stdio[3]);
   // proc.stdio[4].setEncoding('utf-8');
   const fromChild = proc.stdio[4]
-    .pipe(netstringDecoderStream())
+    .pipe(netstringDecoderStream(NETSTRING_MAX_CHUNK_SIZE))
     .pipe(arrayDecoderStream());
 
   // fromChild.addListener('data', data => parentLog(`fd4 data`, data));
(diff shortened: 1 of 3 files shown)


packages/SwingSet/src/lib/netstring.js
```diff
@@ -54,6 +54,12 @@ export function decode(data) {
       // reject NaN, all negative numbers
       assert.fail(X`unparseable size ${sizeString}, should be integer`);
     }
+    if (optMaxChunkSize) {
+      assert(
+        size < optMaxChunkSize,
```
warner: nit: `<=` rather than `<`, if the option is really the "maximum size"

not a big deal but if someone builds an alternate implementation and follows the variable name rather than the implementation, they could wind up off-by-one

packages/SwingSet/test/test-netstring.js
```diff
@@ -102,12 +102,13 @@ test('decode', t => {
...
 
   function bad(input, message) {
-    t.throws(() => decode(Buffer.from(input)), { message });
+    t.throws(() => decode(Buffer.from(input), 100), { message });
   }
 
   // bad('a', 'non-numeric length prefix');
   bad('a:', /unparseable size .*, should be integer/);
   bad('1:ab', 'malformed netstring: not terminated by comma');
+  bad('101:x', /size .* exceeds limit of 100/);
```
warner: please test both edges of the limit, so the test is as tight as the specification, e.g. a passing test that decides a 100-byte payload

packages/SwingSet/src/lib-nodejs/spawnSubprocessWorker.js
```diff
@@ -23,14 +23,16 @@ function parentLog(first, ...args) {
 // always be Node.
 const stdio = harden(['inherit', 'inherit', 'inherit', 'pipe', 'pipe']);
 
+const NETSTRING_MAX_CHUNK_SIZE = 12_000_000;
```
warner, line 26: let's see, this limit will be imposed on all workers that use subprocesses (xsnap, node), ok

If the limit is exceeded, it will basically look like the child process had an error (a malformed netstring, or more likely the child closed the pipe because it's in the process of exiting). I don't know that we're handling that as well as we could be, but fixing that is independent of this change. Given the similarity between this error and ones that can be triggered by e.g. OS shutdown, we should remember to treat these as non-consensus errors, and panic the kernel rather than consensus-terminate the vat.

warner (approved):
Small changes, but generally looks good

mhofman:
Just stumbled on this PR. I believe this change only affects node workers and not xsnap processes which uses it's own netstring stream library from endo, to which I'm adding max message size handling as well: https://github.com/endojs/endo/pull/1277

PR #5656, opened 2022-06-23
Include GC calls in transcript

### LINKED ISSUE #5564: Consider including GC calls in vat transcript

In https://github.com/Agoric/agoric-sdk/issues/5557, @warner suggested:

> I'm wondering if we should stop stripping the GC syscalls from the transcript. We did this when we wanted to tolerate different organic GC schedules between different vats (the "formal vs local GC" era). Given our need to have metering be identical, and our work to make GC consequences only observable during `bringOutYourDead`, I think at this point we could reasonably expect all GC activity to be in consensus, and thus it might no longer be necessary to ignore GC syscalls when replaying the transcript.
> 
> If we stopped ignoring them, this bug would probably have surfaced much faster: if the error caused a second GC syscall to not happen, we'd have gotten an `AnacrophobiaError` during replay.

## Description of the Design

Include GC syscalls in the transcript and expect identical replay during BOYD.

## Test Plan

TBD

FUDCo:
Closes #5564


warner (approved):
Looks good

PR #5635, opened 2022-06-21
Durable ZCF

Chris-Hibbert:
refs: #5333

## Description

Make ZCF durable. Zoe and ZCF don't need to be upgradeable at this point, but we do want to be able to host upgradeable contracts, which means that the per-contract ZCF state has to be durable.

### Security Considerations

Durability, durable contracts.

### Documentation Considerations

We don't yet have a theory of upgradeable contracts. When we do, we need to write that up.

### Testing Considerations

existing tests pass. No tests were added.

PR #5658, opened 2022-06-23
Pass baggage in ZCF through buildRootObject to start()

Chris-Hibbert:
refs: #5593
refs: #5635

## Description

Pass the `baggage` provided to `buildRootObject()` to `startContract()`, which calls `start()`, to make it available to the zoe contract.

### Security Considerations

contract has its own sub-baggage.

When we write up durability, the Zoe contract interface needs to mention that the baggage comes in a separate argument after `privateArgs`.

### Testing Considerations

We'll need a test showing upgrade of a contract, but we already knew that.


packages/zoe/src/contractFacet/vatRoot.js
```diff
@@ -44,11 +48,15 @@ export function buildRootObject(powers) {
       testJigSetter,
     );
     zcfZygote.evaluateContract(bundleOrBundleCap);
+    const contractBaggage = provide(baggage, 'contractBaggage', () =>
+      makeScalarBigMapStore('zoeContract'),
+    );
```
erights, line 53:
```suggestion
    const contractBaggage = provideDurableMapStore(baggage, 'contractBaggage');
```

packages/zoe/src/contractFacet/internal-types.js
```diff
@@ -11,5 +11,6 @@
  *     instanceRecordFromZoe: InstanceRecord,
  *     issuerStorageFromZoe: IssuerRecords,
  *     privateArgs?: object,
+ *     baggage?: MapStore<string,any>,
```
erights, line 14: Also the `startContract` implementation in zcfZygote, right?

erights (approved):
LGTM

Chris-Hibbert:
#5635 will include what's necessary.

PR #5659, opened 2022-06-23
chore(swingset): misc-tools need LMDB pre-import now

warner:
Some tiny tools in misc-tools/ (like db-dump.js) needed to be updated:
swing-store now uses a different LMDB backend, and that backend uses a
package (for serialization of some sort) which uses `eval` as a macro,
which is not compatible with SES. The quick fix, for these tools, is
to import the `lmdb` backend before doing `@endo/init`.


FUDCo (approved):
`import 'dwim';`

PR #5661, opened 2022-06-23
Zcf upgrade demo

warner:
This adds a unit test which is really a demo of what ZCF might do to support contract upgrade.


Chris-Hibbert (approved):
This has been very helpful. It would be good to merge it.

PR #5665, opened 2022-06-24
Ensure that errors which escape a crank panic the kernel

### LINKED ISSUE #5565: Transcript replay should panic on delivery failure
## Describe the bug
As mentioned in https://github.com/Agoric/agoric-sdk/issues/5557:

> We should also make sure that any errors during replay (non-`ok` VatDeliveryResults) cause a replay failure (and ultimately a kernel panic).

This includes BYOD regardless of whether we do https://github.com/Agoric/agoric-sdk/issues/5564 or not.

## To Reproduce
See #5557

FUDCo:
Motivated by the specific case of transcript replay errors (aka Anachrophobia) but not necessarily limited to those, this change ensures that any errors which make it outside a vat delivery cause a kernel panic.

Constructing an automated test for this proves to be subtle and very difficult, if it is possible at all, and the value of such a test is marginal given current schedule constraints.  However, I've tested this intervention manually by making temporary invasive code changes to the kernel (to provoke errors in unexpected places) and then running under Swingset-runner in verbose mode and observing the effects.

Closes #5565

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1165,9 +1165,20 @@ export default function buildKernel(
       assert.fail(X`Kernel reentrancy is forbidden`);
     }
     processQueueRunning = Error('here');
-    return processor(message).finally(() => {
-      processQueueRunning = undefined;
-    });
+    try {
+      return await processor(message).finally(() => {
+        processQueueRunning = undefined;
+      });
+    } catch (err) {
+      // panic() sets the kernelPanic flag which will be checked on the way out
+      // by run() or step().
+      panic(`error during tryProcessMessage: ${err}`, err);
+      // Due to the panic() call, the following return value will be ignored, so
+      // don't read deep meaning into the specific value that is returned here
+      // (though I believe it is correct in context); the return is present to
+      // make eslint happy.
+      return ['crank-failed', {}];
+    }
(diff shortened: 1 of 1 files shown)


packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1165,9 +1165,14 @@ export default function buildKernel(
       assert.fail(X`Kernel reentrancy is forbidden`);
     }
     processQueueRunning = Error('here');
-    return processor(message).finally(() => {
-      processQueueRunning = undefined;
-    });
+    try {
```
warner, line 1168: I don't think having this `return ['crank-failed', []]` upon error will cause the caller (`step()`) to throw as we want.

How about:
```js
const p = processor(message);
p.finally(() => processQueueRunning = undefined);
p.catch(err => panic(...));
return p;
```
?

I think all the processors are async, but `p = Promise.resolve(message).then(processor)` might be an ok safeguard.

FUDCo: The `return ['crank-failed', []]` is just there to make tsc/eslint shut up about the type of the return value.

The throw from `step()` et all is caused by having the `kernelPanic` flag be set by the call to `panic`.

warner (approved):
Got it, thanks for the comment.

PR #5670, opened 2022-06-25
chore(xsnap): update to latest moddable XS version

### LINKED ISSUE #5587: replay error involving promise fulfilling when it should reject (on linux/arm)
While investigating #5571 (which we now believe to be a Zoe bug), I replayed the AMM vat, and found one divergence from the original devnet capture. The `handleOffer()` delivery resolves to a `HandleOfferResult`, which contains a `offerResultP` and an `exitObj`. The offer result is a rejection, because the Zoe bug triggered an assertion as the AMM code was comparing two Amounts (and they had distinct Brand objects). So that last delivery is supposed to look like:

* `dispatch.deliver('handleOffer', args, resultVPID='p-192')`
  * `syscall.resolve('p-192', false, capdata({ exitObj, offerResultP }))` # and `offerResultP` is allocated vpid `p+189`
  * `syscall.send(something, 'fail', capdata(RangeError), resultVPID='p+190')` # probably zcfSeat.fail()
  * `syscall.subscribe('p+190')` # because zcfSeat.fail() is not sendOnly
  * `syscall.resolve('p+189', true, capdata(RangeError))` # the `isReject=true` means `reject`, and `false` would mean `fulfill`

And that's exactly what the slogfile reported. But when I did my replay, I got an anachrophobia error, because that final `syscall.resolve` got `false` instead of `true`. As in liveslots observed the `offerResultP` Promise to *fulfill*, not reject.

I worked out a way to patch the `zcfZygote.js` and `liveslots.js` code, and repeat the replay with more debugging messages enabled. I'm increasingly puzzled: there seems to be a place within ZCF where a rejected promise gets replaced with a fulfilling version.

The original `zcfZygote.js` code is: https://github.com/Agoric/agoric-sdk/blob/9d670a211062f248534df078b012d8fe02f1f170/packages/zoe/src/contractFacet/zcfZygote.js#L324-L343

Liveslots will attach a `.then` to the returned `offerResultP` when it is serialized and sent into the kernel. The `syscall.resolve(vpid, false, capdata(RangeError))` means that liveslots is observing `offerResultP` to fulfill instead of reject (in the replay).

My hacked up version currently looks like:
```js
  const handleOfferObj = Far('handleOfferObj', {
    handleOffer: (invitationHandle, zoeSeatAdmin, seatData) => {
      const zcfSeat = makeZCFSeat(zoeSeatAdmin, seatData);
      const offerHandler = takeOfferHandler(invitationHandle);
      //const offerResultP = E(offerHandler)(zcfSeat, seatData.offerArgs).catch(
      const p1 = E(offerHandler)(zcfSeat, seatData.offerArgs);
      p1.then(v => console.log(`### ZCF.p1 resolve`, v),
              v => console.log(`### ZCF.p1 reject`, v),);

      const offerResultP = p1.catch(
        reason => {
          console.log(`### ZCF sees offerResultP rejected`, reason);
          if (reason === undefined) {
            const newErr = new Error(
              `If an offerHandler throws, it must provide a reason of type Error, but the reason was undefined. Please fix the contract code to specify a reason for throwing.`,
            );
            throw zcfSeat.fail(newErr);
          }
          throw zcfSeat.fail(reason);
        },
      );
      offerResultP.then(v => console.log(`### ZCF.oRP resolve`, v),
                        v => console.log(`### ZCF.oRP reject`, v),);
      const exitObj = makeExitObj(seatData.proposal, zcfSeat);
      /** @type {HandleOfferResult} */
      return harden({ offerResultP, exitObj });
    },
  });
```

and when I run it, I observe the following debug messages (interspersed with the serialization of the RangeError):
```
vat ### ZCF.p1 reject (RangeError#157)
vat ### ZCF sees offerResultP rejected (RangeError#157)
vat ### ZCF.oRP resolve (RangeError#157)
```

The `ZCF.oRP resolve` is the weird thing. All code paths in the `reason => { ... }` function use `throw`, and the only other way out of that function is for something else to throw internally (like `zcfSeat.fail()` or something, so there should be no way for `offerResultP` to ever fulfill.

I ran some experiments with a standalone program doing basically the same thing, and of course it works properly (everything rejects as expected).

This is all running in XS, on an M1Pro macOS laptop. The original trace came from a devnet machine (so probably some x86-64 kernel). I'll re-run this on an intel CPU just in case it makes a difference (which would be horrifying).

warner:
The biggest change is a fix for Linux/ARM, where certain Promise
rejections were doing a fulfill by mistake.

I'm not sure if this changes the metering behavior, but there are
enough changes to the C code that I think it's likely. So I've also
bumped the metering version to 'xs-meter-15'.

closes #5587

### DIFF
--- a/packages/xsnap/api.js
+++ b/packages/xsnap/api.js
@@ -3,7 +3,7 @@
 /** The version identifier for our meter type.
  * TODO Bump this whenever there's a change to metering semantics.
  */
-export const METER_TYPE = 'xs-meter-14';
+export const METER_TYPE = 'xs-meter-15';
 
 export const ExitCode = {
   E_UNKNOWN_ERROR: -1,
--- a/packages/xsnap/test/test-xs-perf.js
+++ b/packages/xsnap/test/test-xs-perf.js
@@ -63,7 +63,7 @@ test('meter details', async t => {
     },
     'evaluate returns meter details',
   );
-  t.is(meterType, 'xs-meter-14');
+  t.is(meterType, 'xs-meter-15');
 });
 
 (globalThis.performance ? test : test.skip)('meter timestamps', async t => {


warner:
I skimmed the C changes, and didn't see anything particularly concerning, although of course there's a lot of stuff changing that I don't understand. There's some new "Archive" thing, and I think some code like `xsMapSet.c` was updated to handle them (or something that can be read-only?) as keys. One change that looks like a bugfix involved NaN and WeakMap keys.

kriskowal (approved):

PR #3436, opened 2021-06-30
Stop importing assert

erights:

Issue #5605, opened 2022-06-15
remove everything @deprecated for MN-1 release


turadg:
## What is the Problem Being Solved?

MN-1 launch requires some breaking changes. Let's bundle them together so we have fewer breaks after.

## Description of the Design

Including but not limited to,
- legacyMap.js

**Acceptance criteria**
- No `@deprecated` in repo
- Loadgen not broken (e.g. [using fallback](https://github.com/Agoric/testnet-load-generator/pull/76))

abebeos:
@turadg The related milestone looks closed (sidenote: has still one dangling issue):

https://github.com/Agoric/agoric-sdk/milestone/37

Can this techdebt issue here be closed?

ivanlei:
There are still @deprecated in the codebase https://github.com/Agoric/agoric-sdk/search?p=1&q=deprecated&type=code

PR #5673, opened 2022-06-27
deprecate assert package

### LINKED ISSUE #5672: migrate to @endo/errors package

`assert` is a global in SES. However, IDEs don't know about it or its members. To allow automatic imports we have `@agoric/assert`, but it's useful outside of Agoric.

## Description of the Design

Move most of `@agoric/assert` to use `@endo/errors`. Move `NonNullish` checked cast to `@agoric/internal`.

# Acceptance

No more use of global `assert` in agoric-sdk. This will support getting assertion functions without having set up a SES environment.

dckc:
fixes #5672

see also #3436

### DIFF
--- a/packages/ERTP/test/swingsetTests/basicFunctionality/bootstrap.js
+++ b/packages/ERTP/test/swingsetTests/basicFunctionality/bootstrap.js
@@ -1,5 +1,5 @@
 import { E } from '@endo/eventual-send';
-import { assert, details as X } from '@agoric/assert';
+const { details: X } = assert;
 import { Far } from '@endo/marshal';
 import { makeIssuerKit, AmountMath } from '../../../src/index.js';
 
--- a/packages/ERTP/src/displayInfo.js
+++ b/packages/ERTP/src/displayInfo.js
@@ -1,6 +1,6 @@
 // @ts-check
 
-import { assert, details as X, q } from '@agoric/assert';
+const { details: X, quote: q } = assert;
 import { assertRecord, isObject, passStyleOf } from '@endo/marshal';
 
 // One GOOGOLth should be enough decimal places for anybody.
--- a/packages/ERTP/src/issuerKit.js
+++ b/packages/ERTP/src/issuerKit.js
@@ -1,7 +1,6 @@
 // @ts-check
 // @jessie-check
 
-import { assert } from '@agoric/assert';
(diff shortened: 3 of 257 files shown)


erights (approved):
LGTM thanks!

See also https://github.com/Agoric/agoric-sdk/issues/5605

turadg (changes requested):
This change is fine but I don't think meets the bar to close #5672 .

I think we should also,
1. Remove all imports from this package in agoric-sdk
2. [Deprecate the package in NPM](https://docs.npmjs.com/cli/v8/commands/npm-deprecate)

We could also consider using the `@deprecated` jsdoc comment on the exports of the package, but I don't think that's useful if we do the two above.

I'm willing to do that as part of https://github.com/Agoric/agoric-sdk/issues/5605 . I pushed some work towards that into this branch.

Fwiw, I think the `const` form is worse because even though `assert` is on global very common uses of it repeat declarations like `const { details: X } = assert;`. That being such a consistent and supported pattern, I'd expect `import { X } from '@agoric/assert';`. Having that export would let IDEs autocomplete when someone tries to use X as a function.

So second thought, I'd like to request that we don't deprecate this package and instead explain why it exists. I'll leave the above to show you how I got here :)

dckc:
> ... I'd like to request that we don't deprecate this package ...

That seems like a disagreement about the issue (#5672), so I withdraw this PR.

packages/ERTP/test/swingsetTests/basicFunctionality/bootstrap.js
```diff
@@ -1,5 +1,5 @@
 import { E } from '@endo/eventual-send';
-import { assert, details as X } from '@agoric/assert';
+const { details: X } = assert;
```
erights, line 2: Not sure why lint is correctly flagging some of these as needing to be moved to after the imports, but is not flagging others, such as this one. In any case, all of these need to be after the imports.

erights (commented):
Is Closing this PR the right thing? It must have been a substantial amount of manual work, and would all be relevant again if we do end up deciding to deprecate the import. Please let's not lose this.

turadg:
> It must have been a substantial amount of manual work

maybe 10min? mostly global find-replace.

> let's not lose this

I've noted this in the issue for when it gets picked up again.

erights:
Thanks!

PR #5674, opened 2022-06-27
fix: simplify singletons

erights:
We've already noticed that singletons always have empty state records, and so their behavior methods never need to use their `state`. @dtribble noticed that we can simply use JS's const scoping (temporal dead zone) to make the `self` or `facets` arguments unnecessary. At that point the context is unnecessary and we just need the "normal" methods. Without common state, there's also little reason to group these into a cohort. Thus, our support for singletons can be much closer to `Far`. To emphasize this, at this time this PR used the term `ProvideFar`, but we should bikeshed.

### DIFF
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -487,7 +487,6 @@ export const makeDurablePaymentLedger = (
     mintPayment,
   });
 
-  // @ts-expect-error Cannot figure out types.
   const brand = /** @type {Brand} */ (
     ProvideFar(issuerBaggage, `${allegedName} brand`, {
       isMyIssuer: allegedIssuerP =>
@@ -508,7 +507,6 @@ export const makeDurablePaymentLedger = (
     assetKind,
     elementSchema,
   );
-  // @ts-expect-error Just too much typing foo for now.
   return issuerKit;
 };
 harden(makeDurablePaymentLedger);
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -5,11 +5,7 @@ import { E } from '@endo/eventual-send';
 import { isPromise } from '@endo/promise-kit';
 import { assertCopyArray } from '@endo/marshal';
 import { fit, M } from '@agoric/store';
-import {
-  dropContext,
-  provideDurableSingletonKit,
-  provideDurableWeakMapStore,
-} from '@agoric/vat-data';
+import { ProvideFar, provideDurableWeakMapStore } from '@agoric/vat-data';
 import { AmountMath } from './amountMath.js';
 import { defineDurablePaymentKind } from './payment.js';
 import { defineDurablePurse } from './purse.js';
@@ -91,7 +87,6 @@ export const makeDurablePaymentLedger = (
   const makePayment = defineDurablePaymentKind(
     issuerBaggage,
     allegedName,
-    // @ts-expect-error Just too much typing foo for now.
     getBrand,
   );
 
@@ -175,16 +170,12 @@ export const makeDurablePaymentLedger = (
   };
 
   /** @type {(left: Amount, right: Amount) => Amount } */
-  // @ts-expect-error Just too much typing foo for now.
   const add = (left, right) => AmountMath.add(left, right, brand);
   /** @type {(left: Amount, right: Amount) => Amount } */
-  // @ts-expect-error Just too much typing foo for now.
   const subtract = (left, right) => AmountMath.subtract(left, right, brand);
   /** @type {(allegedAmount: Amount) => Amount} */
-  // @ts-expect-error Just too much typing foo for now.
   const coerce = allegedAmount => AmountMath.coerce(brand, allegedAmount);
   /** @type {(left: Amount, right: Amount) => boolean } */
-  // @ts-expect-error Just too much typing foo for now.
   const isEqual = (left, right) => AmountMath.isEqual(left, right, brand);
 
   /**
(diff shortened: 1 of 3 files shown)


dtribble:
For VaultManager, my [singleton conversion](https://github.com/Agoric/agoric-sdk/pull/5650) still has a state record. What happens to everything in that state record?

Chris-Hibbert:
I thought I pushed a review earlier, but it's disappeared, sorry.  This seems like it might be _too_ simple. It looks much more like our normal JS code, and not much like our durable code at this point. I think that might be a point against it.

erights:
> I thought I pushed a review earlier, but it's disappeared, sorry. This seems like it might be _too_ simple. It looks much more like our normal JS code, and not much like our durable code at this point. I think that might be a point against it.

That is exactly why I am so excited ;)
Why a point against?

Chris-Hibbert:
> Why a point against?

Just that it doesn't look like the other durable objects, It's not recognizably durable at a glance.

erights:
Github mystery: I thought I had originally listed @Chris-Hibbert as a reviewer. When I try to add him (back), Github does not add him. I have never seen this before. Any idea?

Chris-Hibbert:
You *definitely* had listed me. I remember seeing it. I don't know what's going on.

Chris-Hibbert (approved):
I misunderstood. This is a significant improvement.

packages/ERTP/src/paymentLedger.js
```diff
@@ -91,7 +87,6 @@ export const makeDurablePaymentLedger = (
   const makePayment = defineDurablePaymentKind(
     issuerBaggage,
     allegedName,
-    // @ts-expect-error Just too much typing foo for now.
```
turadg, line 94: glad to see this solved

packages/ERTP/src/paymentLedger.js
```diff
@@ -476,40 +467,41 @@ export const makeDurablePaymentLedger = (
...
+    makeEmptyPurse,
+  });
+
+  const mint = ProvideFar(issuerBaggage, `${allegedName} mint`, {
+    getIssuer: () => issuer,
+    mintPayment,
+  });
 
-      getAllegedName: _context => allegedName,
+  // @ts-expect-error Cannot figure out types.
```
turadg: if you commit the suggestion in `kind-utils.js` this suppression can be removed

packages/vat-data/src/kind-utils.js
```diff
@@ -1,22 +1,25 @@
...
 // @ts-expect-error TODO statically recognize harden
 harden(provideKindHandle);
 
-export const provideDurableSingleton = (
-  baggage,
-  kindName,
-  behavior,
-  options = undefined,
-) => {
+export const ProvideFar = (baggage, kindName, methods, options = undefined) => {
```
turadg, line 49:
```suggestion
/**
 * @template T
 * @param {import('@agoric/store').MapStore<string, unknown>} baggage
 * @param {string} kindName
 * @param {T} methods
 * @param {import('./types.js').DefineKindOptions<unknown>} [options]
 * @returns {import('@endo/far').ERef<T>}
 */
export const ProvideFar = (baggage, kindName, methods, options) => {
```

packages/vat-data/src/kind-utils.js
```diff
@@ -1,22 +1,25 @@
...
   const kindHandle = provideKindHandle(baggage, kindName);
+  const behavior = fromEntries(
+    entries(methods).map(([k, m]) => [k, dropContext(m)]),
+  );
```
turadg: mapping object values at the existing keys is a common need so in run-protocol package we have a helper 
https://github.com/Agoric/agoric-sdk/blob/68d27f32651a49021d10a50863504e3b92988419/packages/run-protocol/src/collect.js#L6-L8

should we have it in a more shared package? `@agoric/utils`?

turadg (approved):
👌

packages/vat-data/src/kind-utils.js
erights: I did something similar but different at https://github.com/Agoric/agoric-sdk/pull/5674/commits/98a96e947077a098aab3f1f7fcda0d04438edd97

I left the optional parameter syntax on the options argument. I think we should in general.

I replaced the `ERef<T>` with simply `T`. However, I see at https://github.com/endojs/endo/blob/8077e2134615d6a1cbe8c8f5350f664b56f0a17b/packages/marshal/src/make-far.js#L88 the similar `T & RemotableBrand<{}, T>`. Please let me know if I should use this instead.

Attn @michaelfig who authored `RemotableBrand`

PTAL

packages/vat-data/src/kind-utils.js
erights: Yes, I have a similar one, yes with an extensive comment, at 
https://github.com/Agoric/agoric-sdk/blob/86d58b972b123bd77ade5371fd72340da9c6c8a0/packages/zoe/src/objArrayConversion.js#L45-L48

Which of these two do you prefer?

Like `mapValues` it is not reusable only because of where it is. However, `@agoric/utils` would imply that it is in agoric-sdk and so not reusable from the endo repository. Attn @kriskowal is there a natural endo package to migrate such things to? Should there be?

In the meantime, vat-data should clearly not depend on run-protocol or zoe. But I can at least make it a copy of one of these with a note to depend on a common one once a suitable place is found. In the absence of other feedback, I'll copy `objectMap`. Thanks.

packages/ERTP/src/paymentLedger.js
erights: done

packages/vat-data/src/kind-utils.js
erights: @turadg I copied `objectMap` but let me know if I should change it to `mapValues`.

Despite the two PTALs above, I'm going to merge this PR into https://github.com/Agoric/agoric-sdk/pull/5283 . But still, please do PTAL and comment either here or (preferred) on https://github.com/Agoric/agoric-sdk/pull/5283 . Thanks.

turadg: Agree we should find a place for utilities like this and it should be somewhere in `@endo`. I'd like to see something analogous to Lodash but SES-smart. @kriskowal WDYT of that? I can make a ticket and I'd be very happy to work on it after MN-1.

As to whether `mapValues` or `objectMap` goes in the utility package, I think `mapValues` fits the common use case better: mapping values without regard to the keys. `objectMap` takes a function that allows keyword transformations so it's not doing much abstracting over `Object.fromEntries(Object.entries.map(mapPairFn))`;

packages/vat-data/src/kind-utils.js
turadg: I'm not sure. I need to better understand the remote call and marshalling systems. I do know the types are a loose approximation so far.

PR #5677, opened 2022-06-27
SPIKE: Productionise wallet frontend

michaelfig:
closes: #XXXX
refs: #XXXX


michaelfig:
This spike is outdated, and a different approach was taken for the smart wallet.  Closing.

Issue #4432, opened 2022-02-01
Plan work to improve test design and coverage of run-protocol


turadg:
## What is the Problem Being Solved?

Some important functionality isn't yet tested. For example https://github.com/Agoric/agoric-sdk/pull/4394 changed the charging period governance but no tests broke. Writing tests was deferred to when we have a more complete specification of what's in run-protocol for MN-1 and how we should factor the tests. (E.g. what to do as unit tests and what to have in functional tests integrating SwingSet.)

## Description of the Design

TBD

## Security Considerations

Doesn't implicate security but could improve testing of security considerations.

## Test Plan

n/a

turadg:
I'll comment here with observations as they come up.

`buildVatController` is an abstraction that shouldn't be used for real applications because it doesn't give them the necessary amount of control over state initialization.. if they're using buildVatController, they probably can't survive a reboot. It's only used in tests right now and we should find a way to keep it that way. `@deprecated` isn't appropriate because it's a legit helper to use in tests. Currently used by tests in these packages:
- ERTP
- governance
- run-protocol
- sharing-service
 - spawner
- zoe

Tartuffo:
@turadg For proper project planning and tracking, this needs an area label covered by one of our weekly planning meetings.  Please pick the appropriate one from: agd, agoric-cli, agoric-cosmos, amm, core economy, cosmic-swingset, endo, ertp, getrun, governance, installation-bundling, metering, oracle, pegasus, run-protocol, ses, staking, swingset, swingset-runner, tc39, token economy, tooling, ui, wallet, xsnap, zoe, zoe contract

PR #5678, opened 2022-06-27
5268 governance StoredSubscriptions

### LINKED ISSUE #5268: Expose RUN protocol governance state via subscription

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), there are various data we want to expose.

We have tickets for different scopes of the data. This one is for what cuts across all of them: governance parameter state.

# Design

Do we mix the governance states with the existing notifiers or have separate notifiers?
If we do mix should we have a combiner function?

Use `StoreSubscriber` as in https://github.com/Agoric/agoric-sdk/pull/5613 to expose them in RPC.

turadg:
closes: Agoric/agoric-sdk#5268

## Description

Makes the `getSubscription` added by `augmentPublicFacet` use `StoredSubscription` so it can be read off-chain.

Makes each contract that uses it provide the storageNode and marshaller necessary to write to off-chain store. Node presence and path are confirmed by new `storage key` tests. Nothing tests the marshaller.

### Security Considerations

Marshaller provided to each contract. Opted for readonly.

Should we maintain a list of paths somewhere? They can be collected from the 'storage keys' tests (with the proper renaming of root):
```
mockChainStorageRoot.amm.governance
mockChainStorageRoot.amm.metrics
mockChainStorageRoot.amm.pool0.metrics

mockChainStorageRoot.psm.governance

mockChainStorageRoot.reserve.governance
mockChainStorageRoot.reserve.metrics

mockChainStorageRoot.stakeFactory.governance

mockChainStorageRoot.vaultFactory.governance
mockChainStorageRoot.vaultFactory.metrics
mockChainStorageRoot.vaultFactory.manager0.governance
mockChainStorageRoot.vaultFactory.manager0.metrics
mockChainStorageRoot.vaultFactory.manager1.governance
mockChainStorageRoot.vaultFactory.manager1.metrics
```

### Testing Considerations

Note that the PSM test doesn't integrate bootstrap because I re-used what was already there, which was a direct startInstance. It looks like we have no tests of startPSM. I've added the need to https://github.com/Agoric/agoric-sdk/issues/4432

### DIFF
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -1,13 +1,14 @@
 // @ts-check
 
 import { Far } from '@endo/marshal';
+import { makeStoredPublisherKit } from '@agoric/notifier';
 import { keyEQ } from '@agoric/store';
-// eslint-disable-next-line -- https://github.com/Agoric/agoric-sdk/pull/4837
-import { CONTRACT_ELECTORATE } from './contractGovernance/governParam.js';
 import { assertElectorateMatches } from './contractGovernance/paramManager.js';
 import { makeParamManagerFromTerms } from './contractGovernance/typedParamManager.js';
 
 const { details: X, quote: q } = assert;
+
+export const GOVERNANCE_STORAGE_KEY = 'governance';
 
 /**
  * Utility function for `makeParamGovernance`.
@@ -116,17 +117,25 @@ const facetHelpers = (zcf, paramManager) => {
  *
  * @template {import('./contractGovernance/typedParamManager').ParamTypesMap} M
  *   Map of types of custom governed terms
- * @param {import('@agoric/notifier').StoredPublisherKit<GovernanceSubscriptionState>} publisherKit
  * @param {ZCF<GovernanceTerms<M>>} zcf
  * @param {Invitation} initialPoserInvitation
  * @param {M} paramTypesMap
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
  */
 const handleParamGovernance = async (
-  publisherKit,
   zcf,
   initialPoserInvitation,
   paramTypesMap,
+  storageNode,
+  marshaller,
 ) => {
+  /** @type {import('@agoric/notifier').StoredPublisherKit<GovernanceSubscriptionState>} */
+  const publisherKit = makeStoredPublisherKit(
+    storageNode,
+    marshaller,
+    GOVERNANCE_STORAGE_KEY,
+  );
   const paramManager = await makeParamManagerFromTerms(
     publisherKit,
     zcf,
--- a/packages/run-protocol/src/proposals/core-proposal.js
+++ b/packages/run-protocol/src/proposals/core-proposal.js
@@ -193,6 +193,8 @@ const RUN_STAKE_MANIFEST = harden({
   },
   [econBehaviors.startRunStake.name]: {
     consume: {
+      board: 'board',
+      chainStorage: true,
       zoe: 'zoe',
       feeMintAccess: 'zoe',
       lienBridge: true,
--- a/packages/run-protocol/src/proposals/econ-behaviors.js
+++ b/packages/run-protocol/src/proposals/econ-behaviors.js
@@ -40,14 +40,14 @@ const CENTRAL_DENOM_NAME = 'urun';
  *   ammCreatorFacet: XYKAMMCreatorFacet,
  *   ammGovernorCreatorFacet: GovernedContractFacetAccess<XYKAMMCreatorFacet>,
  *   economicCommitteeCreatorFacet: CommitteeElectorateCreatorFacet,
- *   feeDistributorCreatorFacet: FeeDistributorCreatorFacet,
- *   feeDistributorPublicFacet: FeeDistributorPublicFacet,
+ *   feeDistributorCreatorFacet: import('../feeDistributor.js').FeeDistributorCreatorFacet,
+ *   feeDistributorPublicFacet: import('../feeDistributor.js').FeeDistributorPublicFacet,
  *   periodicFeeCollectors: import('../feeDistributor.js').PeriodicFeeCollector[],
  *   bankMints: Mint[],
  *   psmCreatorFacet: unknown,
  *   psmGovernorCreatorFacet: GovernedContractFacetAccess<unknown>,
- *   reservePublicFacet: AssetReservePublicFacet,
- *   reserveCreatorFacet: AssetReserveCreatorFacet,
+ *   reservePublicFacet: import('../reserve/assetReserve.js').AssetReservePublicFacet,
+ *   reserveCreatorFacet: import('../reserve/assetReserve.js').AssetReserveCreatorFacet,
  *   reserveGovernorCreatorFacet: GovernedContractFacetAccess<unknown>,
  *   runStakeCreatorFacet: import('../runStake/runStake.js').RunStakeCreator,
  *   vaultFactoryCreator: VaultFactory,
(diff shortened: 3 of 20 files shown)


packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -219,9 +217,11 @@ export const setupAmm = async (
   );
 
   const chainStoragePresence = await chainStorage;
-  const ammChainStorage = await (chainStoragePresence &&
+  const storageNode = await (chainStoragePresence &&
     E(chainStoragePresence).getChildNode(AMM_STORAGE_PATH));
-  const marshaller = E(board).getPublishingMarshaller();
+  // FIXME why does this hang?
```
turadg: @dckc any ideas?

dckc: I think we don't resolve the `board` promise by, for example, calling `makeBoard` in any place like `setupBootstrap` (supports.js).

For an example of where we _do_ supply a board in test bootstrap, see:
https://github.com/Agoric/agoric-sdk/blob/7dbe6c0e948f0686ed77ef3439c69f6af1dc29d2/packages/run-protocol/test/test-gov-collateral.js#L175

dckc:
> Should we maintain a list of paths somewhere?

Yes, we should, in due course. It should be a [documentation snippet](https://github.com/Agoric/documentation/blob/main/CONTRIBUTING.md#importing-and-testing-code-snippets). Perhaps add a documentation issue?

Meanwhile, how about a test that iterates over all the keys and compares against one big expected array? Hm... I suppose enumerating the keys is only doable via the RPC interface and not via JS APIs. So that would have to be an integration test. Probably not cost-effective.

For now, the list above looks pretty handy.

packages/governance/src/contractHelper.js
```diff
@@ -116,17 +115,25 @@ const facetHelpers = (zcf, paramManager) => {
...
   initialPoserInvitation,
   paramTypesMap,
+  storageNode,
+  marshaller,
 ) => {
+  /** @type {import('@agoric/notifier').StoredPublisherKit<GovernanceSubscriptionState>} */
+  const publisherKit = makeStoredPublisherKit(
+    storageNode,
+    marshaller,
+    'governance',
```
dckc: hm. magic string. Perhaps add a manifest constant in `params.js` to facilitate use by clients?

packages/run-protocol/src/proposals/core-proposal.js
```diff
@@ -193,6 +193,8 @@ const RUN_STAKE_MANIFEST = harden({
   },
   [econBehaviors.startRunStake.name]: {
     consume: {
+      board: 'board',
+      chainStorage: true,
```
dckc, line 197: good... I wonder if test coverage prompted this or if you just remembered.

packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -219,9 +217,9 @@ export const setupAmm = async (
   );
 
   const chainStoragePresence = await chainStorage;
-  const ammChainStorage = await (chainStoragePresence &&
+  const storageNode = await (chainStoragePresence &&
     E(chainStoragePresence).getChildNode(AMM_STORAGE_PATH));
-  const marshaller = E(board).getPublishingMarshaller();
+  const marshaller = await E(board).getReadonlyMarshaller();
```
dckc: That's a design change... perhaps adds a documentation consideration: take care that any brands you want marshaled in an identity-preserving way are on the board before... um... before the contract is started? Before any updates you care about?

packages/run-protocol/src/proposals/econ-behaviors.js
```diff
@@ -820,6 +820,8 @@ export const startRunStake = async (
     recordingPeriod = SECONDS_PER_DAY,
   } = {},
 ) => {
+  const STORAGE_PATH = 'stakeFactory';
```
dckc, line 823: export that `STORAGE_PATH` from `runStake/params.js` to facilitate use by clients? Or... cross that bridge when we come to it?

packages/run-protocol/src/vpool-xyk-amm/params.js
```diff
@@ -15,35 +10,11 @@ const DEFAULT_POOL_FEE_BP = 24n;
 const DEFAULT_PROTOCOL_FEE_BP = 6n;
 
 /**
- * @param {ERef<ZoeService>} zoe
- * @param {bigint} poolFeeBP
+ * @param {Invitation} electorateInvitationAmount - invitation for the question poser
```
dckc: ouch. That hurts my brain. Is it an invitation or an amount? It's an invitation, right?

turadg:
> Perhaps add a documentation issue?
https://github.com/Agoric/documentation/issues/765

> enumerating the keys is only doable via the RPC interface and not via JS APIs. So that would have to be an integration test. Probably not cost-effective.

Agree. We might want to have integration tests of the actual data but I figure that will be the scope of ETL and may integrate across repositories.

packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -23,6 +23,10 @@ export const publicPrices = prices => {
   return { amountIn: prices.swapperGives, amountOut: prices.swapperGets };
 };
 
+/**
+ * @typedef {Record<string, Amount>} NotificationState
```
dckc: Is the type really a record with arbitrary string keys? not `{ Central: Amount, Secondary: Amount, Liquidity: Amount}`?

dckc (commented):
review so far

packages/run-protocol/test/reserve/test-reserve.js
```diff
@@ -455,3 +456,22 @@ test('reserve burn IST', async t => {
...
+
+test('storage keys', async t => {
+  /** @param {NatValue} value */
+  const electorateTerms = { committeeName: 'EnBancPanel', committeeSize: 3 };
+  const timer = buildManualTimer(t.log);
+
+  const { reserve } = await setupReserveServices(t, electorateTerms, timer);
+
+  t.is(
+    // @ts-expect-error problem with E() and GovernedPublicFacet<>
```
dckc, line 468: The quest to get the governance types right goes on...

packages/run-protocol/test/test-voPool.js
```diff
@@ -1,12 +1,43 @@
...
+import { mapValues } from '../src/collect.js';
+
+/**
+ * @param {ERef<ZoeService>} zoe
+ * @param {bigint} poolFeeBP
+ * @param {bigint} protocolFeeBP
+ * @param {Amount} minInitialLiquidity
+ * @param {Invitation} poserInvitation - invitation for the question poser
+ */
+const makeAmmParamManager = async (
```
dckc, line 22: you don't want to export this from `params.js` any more?

Strictly speaking, that should get flagged as a breaking change, yes?

dckc (approved):
Looks good.

My one "ouch" comment is almost critical, but I'll leave it to your judgement.

packages/run-protocol/src/proposals/econ-behaviors.js
turadg: Thanks for finding this. I think AMM does need the write capability to write brands when adding new liquidity pools.

packages/run-protocol/src/proposals/econ-behaviors.js
turadg: cross the bridge later. I'm not certain of the design.

packages/run-protocol/src/vpool-xyk-amm/params.js
turadg: this is just matching the existing code but I'll strike "Amount"

packages/run-protocol/test/test-voPool.js
turadg: strictly speaking yes but I'm confident it wasn't used anywhere else

packages/run-protocol/src/proposals/core-proposal.js
turadg: I don't recall.

packages/governance/src/contractHelper.js
turadg: pulled up to module scope,
```
export const GOVERNANCE_STORAGE_KEY = 'governance';
```

packages/run-protocol/src/vpool-xyk-amm/pool.js
turadg: it's not that value but good call typing more explicitly

Issue #5600, opened 2022-06-15
remove `gcEveryCrank` and worker type `xs-worker-no-gc`


warner:
## What is the Problem Being Solved?

A long while ago, before we introduced `dispatch.bringOutYourDead`, liveslots did a forced GC at the end of every crank. in a panic to improve performance during one of the incentivized testnet runs, we introduced a quick hack wherein `managerType: 'xs-worker-no-gc'` would use a new `gcEveryCrank:  false` flag to replace the normal `gcAndFinalize` with a dummy version, thus disabling GC entirely.

Now that we have configurable scheduling for BOYD, and GC is only forced during BOYD, we don't need this mechanism anymore.

## Description of the Design

I count 8 files that reference either `gcEveryCrank` or `xs-worker-no-gc`:

* misc-tools/replay-transcript.js
* src/controller/initializeSwingset.js
* src/kernel/state/kernelKeeper.js
* src/kernel/vat-loader/manager-factory.js
* src/kernel/vat-loader/manager-subprocess-xsnap.js
* src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
* src/types-external.js
* src/vats/vat-admin/vat-vat-admin.js

The task is to remove references to both, making `gcAndFinalize` be always the real function.

## Security Considerations

When GC is disabled in this way, it prevents BOYD from doing a forced GC, but it doesn't prevent organic GC from happening. If we allow organic GC to happen on different schedules in different validators (which used to be a goal, but #5564 discusses why I think we should disallow it), the lack of a forced GC would probably cause GC syscalls to happen later than usual, which could then turn into a consensus break.

The lack of forced GC probably wouldn't affect the risk of worker memory exhaustion very much, because organic GC keeps happening.

`managerType: xs-worker-no-gc` is currently the only way that a dynamic vat creator has to disable forced GC. BOYD scheduling can be controlled (or effectively disabled) by the host application, through `controller.changeKernelOptions()`, but the per-vat admin facet does not provide that authority.

So I'd argue that security is marginally better without `xs-worker-no-gc`.

## Test Plan

The existing unit tests are probably sufficient, because we never got around to adding a test for `xs-worker-no-gc` anyways.

PR #5688, opened 2022-06-29
fix(swingset): fix transcript extract/replay tools

### LINKED ISSUE #5602: fix extract-transcript and replay-transcript tools

While investigating #5571 and #5587, I found that my tools for transcript extraction/replay had bitrotted. The task is to update them.

* transcript extraction needs to capture the whole syscall result, even for failed syscalls
* `gcEveryCrank` must be `true`, else GC is entirely disabled
* replay now uses the whole syscall result, even for failed syscalls
* the fake kernelKeeper needs `getEnableFakeDurable`
* the fake kernelSlog needs `delivery()` and `syscall()`
* we should track `deliveryNum` and try to align it with the original slog

## Security Considerations

None, this is for a diagnostic tool

## Test Plan

None, although if I'd put more energy into writing these tools in a form that could be tested, they probably wouldn't have bitrotted. I'm currently ok with that.

warner:
These had bitrotted:
* transcript now includes whole syscall result, even failures
* handle kerneldb bundleIDs
* new LMDB in swingstore
* gcEveryCrank=false else all GC is disabled (refs #5600)
* fake kernelKeeper/slogger needed new methods

The new code also attempts to track `deliveryNum`, to align it with
the original slog, but I'm not positive it is accurate.

closes #5602

### DIFF
--- a/packages/SwingSet/misc-tools/extract-transcript-from-kerneldb.js
+++ b/packages/SwingSet/misc-tools/extract-transcript-from-kerneldb.js
@@ -1,8 +1,8 @@
-// XXX this is wrong; it needs to use the swingstore instead of opening the LMDB
-// file directly, then use stream store reads to get the transcript entries.
-import lmdb from 'lmdb';
+import 'lmdb';
+import '@endo/init';
 import process from 'process';
 import fs from 'fs';
+import { isSwingStore, openSwingStore } from '@agoric/swing-store';
 
 const argv = process.argv.splice(2);
 const dirPath = argv[0];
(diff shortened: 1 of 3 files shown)


packages/SwingSet/misc-tools/extract-transcript-from-kerneldb.js
```diff
@@ -83,28 +99,43 @@ if (!vatName) {
...
+  // host application failed to commit the kvStore).
+  //
+  // The kvStore `${vatID}.t.startPosition` tells us the index of the
+  // first entry of the most recent version (the most recent
+  // `startVat`), while `${vatID}.t.endPosition` tells us the last
+  // committed entry.
+  //
+  // We ignore the heap snapshot ID, because we're deliberately
+  // replaying everything from `startVat`, and therefore we also
+  // ignore `snapshot.startPos` (the index of the first delivery
```
mhofman, line 117: These would be the ones stored in `local.v$NN.lastSnapshot`, right ?

mhofman (approved):
I'm not super familiar with the transcript logic yet, but this all looks fine to me.

packages/SwingSet/misc-tools/extract-transcript-from-kerneldb.js
warner: yep

PR #5695, opened 2022-06-30
fix(notifier): Make makeNotifierFromAsyncIterable lossy

### LINKED ISSUE #5413: notifier returned by makeNotifierFromAsyncIterable is lossless

gibson042:
Fixes #5413

## Description

Separate iterable consumption from `getUpdateSince` peeking.

### Security Considerations

Because the consumption is now greedy, there is a risk of getting stuck in a produce–consume pattern where each step enqueues a subsequent turn and forward progress halts. But I don't think any of our code fits that pattern, and at any rate I don't know if any mitigation is even possible.

n/a

### Testing Considerations

WIP

### DIFF
--- a/packages/governance/test/unitTests/test-paramGovernance.js
+++ b/packages/governance/test/unitTests/test-paramGovernance.js
@@ -223,10 +223,9 @@ test('change multiple params', async t => {
 
   /** @type {GovernedPublicFacet<unknown>} */
   const publicFacet = await E(governorFacets.creatorFacet).getPublicFacet();
-  const notifier = makeNotifierFromAsyncIterable(
-    await E(publicFacet).getSubscription(),
-  );
-  const update1 = await notifier.getUpdateSince();
+  const governorSubscription = await E(publicFacet).getSubscription();
+  const governorUpdates = await E(governorSubscription)[Symbol.asyncIterator]();
+  const update1 = await E(governorUpdates).next();
   // constructing the fixture to deepEqual would complicate this with insufficient benefit
   t.is(
     // @ts-expect-error reaching into unknown values
@@ -279,7 +278,7 @@ test('change multiple params', async t => {
     t.fail(`expected success, got ${e}`);
   });
 
-  const update2 = await notifier.getUpdateSince(update1.updateCount);
+  const update2 = await E(governorUpdates).next();
   t.like(update2, {
     value: {
       current: {
--- a/packages/notifier/src/subscriber.js
+++ b/packages/notifier/src/subscriber.js
@@ -46,8 +46,8 @@ const makeSubscriptionIterator = tailP => {
   return Far('SubscriptionIterator', {
     subscribe: () => makeSubscription(tailP),
     [Symbol.asyncIterator]: () => makeSubscriptionIterator(tailP),
-    next: () => {
-      const resultP = E.get(tailP).head;
+    next: async () => {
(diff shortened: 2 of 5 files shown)


packages/governance/test/unitTests/test-paramGovernance.js
```diff
@@ -223,10 +223,9 @@ test('change multiple params', async t => {
 
   /** @type {GovernedPublicFacet<unknown>} */
   const publicFacet = await E(governorFacets.creatorFacet).getPublicFacet();
-  const notifier = makeNotifierFromAsyncIterable(
-    await E(publicFacet).getSubscription(),
-  );
-  const update1 = await notifier.getUpdateSince();
+  const governorSubscription = await E(publicFacet).getSubscription();
+  const governorUpdates = await E(governorSubscription)[Symbol.asyncIterator]();
```
gibson042, line 227: I'm not clear on why such use of `makeNotifierFromAsyncIterable` for testing in governance and run-protocol causes lockup (e.g., https://github.com/Agoric/agoric-sdk/runs/7138160731?check_suite_focus=true : "_unitTests › paramGovernance › change multiple params … Error: Promise returned by test never resolved_"). It could be related to use of a fake timer, or could be an actual bug. But the tests don't appear to have been expecting lossyness, so I feel mostly fine about direct use of `await … next()`.

packages/notifier/src/subscriber.js
```diff
@@ -46,8 +46,8 @@ const makeSubscriptionIterator = tailP => {
   return Far('SubscriptionIterator', {
     subscribe: () => makeSubscription(tailP),
     [Symbol.asyncIterator]: () => makeSubscriptionIterator(tailP),
-    next: () => {
-      const resultP = E.get(tailP).head;
+    next: async () => {
+      const resultP = await E.get(tailP).head;
```
gibson042, line 50: This one _might_ not be necessary, but it does seem problematic to eagerly to let an eager consumer (such as the new makeNotifierFromAsyncIterable) read an arbitrarily long sequence of results that might never materialize.

packages/governance/test/unitTests/test-paramGovernance.js
turadg: I'm concerned that we don't understand why it's failing. If it's isolated to these cases then I can agree to sweep it under the rug as something not worth getting to the bottom of.

I'm not sure it's isolated though. https://github.com/Agoric/agoric-sdk/pull/5704 attempts to use `makeNotifierFromAsyncIterable` for notifiers (necessary for https://github.com/Agoric/agoric-sdk/issues/5386 ) and it also hangs the same way.

turadg (changes requested):
Recapping from our meeting per request:
- leave `makeNotifierFromAsyncIterable` as is
- deprecate it because it's a hot notifier (to be removed in https://github.com/Agoric/agoric-sdk/issues/5709 )
- make something new that takes a subscriber to make a cold notifier

gibson042:
Superseded by #5737.

PR #5697, opened 2022-06-30
feat!(swingset): remove vatPowers.makeMarshal

### LINKED ISSUE #5639: does `vatPowers.makeMarshal` reveal any significant WeakMap authority to userspace?
While talking to folks about marshalling today, it occured to me that userspace is denied access to the raw `WeakMap`, but `vatPowers.makeMarshal` provides access to code that might make use of it internally.

Can this be used to cause a nondeterminism problem? Specifically, the raw WeakMap could be used as a recognition predicate that can distinguish between two Representatives for the same vref, which would allow userspace to sense GC, which is not allowed.

I think the answer is "no", because the only `WeakMap` in `@endo/marshal` is in `passStyleOf`. The one used by `convertValToSlot` is provided by the caller of `makeMarshal()`, and we no longer do `ibid` mapping (which would have been harder to exploit anyways, being scoped to a single act of serialization). 

But I need to think about it, and probably check with @erights

warner:
We added `vatPowers.makeMarshal` eons ago, long enough that we don't
remember why. (I know there was a time that we needed vats to share a
`Remotable` registry WeakMap with the marshal instance used by
liveslots, so objects that the vat tagged with Far()/Remotable() would
be recognized as such by liveslots. But that was sharing
`Far`/`Remotable` and not `makeMarshal`, which exists to make new
marshallers).

We're concerned that the full-powered WeakMap used internally by
marshal (specifically the `passStyleOf` memoization cache) might be
usable to sense GC of virtual-object Representatives, which would
allow userspace to violate our determinism goals.

Lots of vat code does:
```js
import { makeMarshal } from '@endo/marshal';
```

which is fine: as vat code, that copy of `@endo/marshal` will get the
`WeakMap` provided by liveslots a vat global, and that's the tamed
version which maps multiple (sequential) Representative to the same
key.

It looks like nothing in agoric-sdk is using `makeMarshal` from
`vatPowers`, so the easy fix is to remove it.

closes #5639

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -1628,16 +1628,12 @@ export function makeLiveSlots(
   buildVatNamespace,
   relaxDurabilityRules = false,
 ) {
-  const allVatPowers = {
-    ...vatPowers,
-    makeMarshal,
-  };
   const r = build(
     syscall,
     forVatID,
     cacheSize,
     enableDisavow,
-    allVatPowers,
+    vatPowers,
     gcTools,
     liveSlotsConsole,
     buildVatNamespace,


warner:
cc @erights

mhofman (approved):
I don't know much about this piece of history, but if all tests pass, I'm happy to remove powers!

PR #5698, opened 2022-06-30
ci: Constrain GitHub workflows to the agoric organization

gibson042:
## Description

This should prevent failures in forked development repositories from issues like missing credentials (cf. https://github.com/gibson042/agoric-sdk/actions ).

### Security Considerations

This conforms with https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions and https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context AFAICT, but it is important that accommodation of development does not disrupt production workflows.

n/a

### Testing Considerations

I think the only (good) way to verify this is to sync the master branch in a forked repository after merging it.


.github/workflows/integration.yml
```diff
@@ -17,6 +17,7 @@ on:
 
 jobs:
   pre_check:
+    if: ${{ github.repository_owner == 'agoric' }}
```
mhofman: I'm surprised this job requires credentials

mhofman: Ah it's for the notifications!
Let's move the check on the notify on failure step only

mhofman (changes requested):
I think the integration job should still run on forks, just not do the notify on failure

.github/workflows/integration.yml
gibson042: Sure; done.

mhofman (approved):

turadg:
The automerge was blocked on https://github.com/Agoric/agoric-sdk/runs/7143783366?check_suite_focus=true so I clicked to rerun failed jobs.

PR #5594, opened 2022-06-14
Implement smart-wallet and read-only UI

michaelfig:
closes: #5356

## Description

Here's a (rather unimpressive) working demonstration of the smart wallet UI.  Notice the zero-latency initialization of the view, and that it's currently read-only.

https://user-images.githubusercontent.com/457244/173690436-b93d6014-5ab7-4f53-966d-aa9fc72dc5e1.mov

Issue #4484, opened 2022-02-07
make Smart Wallet contract multitenant


samsiegart:
All the wallet backends need to be implemented by a single wallet manager contract, since it is inefficient to have many small vats when a large one would suffice.  Multitenant backend wallet states, which are managed by a single Zoe contract, are presented as sealed individual backends, each of the schema in #4483, and broadcasted via the Cosmos event stream.

To get there from here, we will
 - [x] upgrade the `packages/wallet/api` directory to be multitenant, indexed first by tenant (such as their Cosmos address), and produce the existing wallet as a specialisation of that code.  Then, the wallet manager is an instance of that code as well, but built to handle inbound Cosmos transactions.

 - [ ] To facilitate migration to virtual/durable objects and collection, each individual wallet backend must be supplied with a new schema manager rather than instantiating its own collections.

# Design

Look up a tenant by its address at the top level and get ocaps. One mapping from address to that object.
That tenant's wallet is like a vault, closing over their data.

Tartuffo:
@michaelfig should this be assigned to you?  (currently @turadg who does not know much about wallet.)

Tartuffo:
@turadg is multi-tenant done, should we rename this to "Make Smart Wallet contract durable?

turadg:
This was primary about multitenancy (which is why it blocked the two things it blocks). I've made https://github.com/Agoric/agoric-sdk/issues/5894 and closed this.

PR #5701, opened 2022-06-30
feat: contract for single smart wallet

### LINKED ISSUE #5356: Smart wallets can get balances with zero transactions (for small # of users)
Epic milestone to https://github.com/Agoric/agoric-sdk/issues/3995

## Acceptance criteria

- When using the wallet, Keplr is not brought up until there is a transfer of value
- UI can show wallet balances
- Subscriptions publish to Cosmos events

turadg:
closes: #5356

## Description

Smart contract for smart wallet, forked from #5594

Names this version of the contract `singleWallet` so it can co-exist with the multitenant `walletFactory` coming in https://github.com/Agoric/agoric-sdk/pull/5721 for https://github.com/Agoric/agoric-sdk/issues/4484

### Security Considerations

Wallet now works on-chain. 

Adds a basic README to document the difference between "solo" wallet, smart/chain wallet with a singleton contract and the multitenant factory.

### Testing Considerations

CI doesn't cover everything. Ideally we test the on-chain [`home.wallet` in CI ](https://github.com/Agoric/agoric-sdk/blob/2edffdbe007e5fab3a760f986d5df86314164134/packages/solo/test/test-home.js#L58) but that's a big chunk of work and for now we can do it manually.

```
# tab 1 (chain)
cd packages/cosmic-swingset/
make scenario2-setup scenario2-run-chain
# starts bare chain, don’t need AMM

# tab 2 (client server)
cd packages/cosmic-swingset/
make scenario2-run-client
# CONFIRMED no errors in logs

# tab 3 (interactive)
agoric open --repl
# CONFIRMED in browser that `home.wallet` and `home.smartWallet` exist
agd query vstorage keys 'published.wallet'
# CONFIRMED it has a key like `published.wallet.agoric1nqxg4pye30n3trct0hf7dclcwfxz8au84hr3ht`
agoric follow :published.wallet.agoric1nqxg4pye30n3trct0hf7dclcwfxz8au84hr3ht
# CONFIRMED it has JSON data

### CODE REVIEW

File: packages/wallet/contract/test/test-contract.js [hunk 1]
```diff
@@ -0,0 +1,200 @@
...
+    const { nameHub: myAddressNameHub, nameAdmin: rawMyAddressNameAdmin } =
+      makeNameHubKit();
+
+    /** @type {MyAddressNameAdmin} */
+    const myAddressNameAdmin = Far('myAddressNameAdmin', {
+      ...rawMyAddressNameAdmin,
+      getMyAddress: () => address,
+    });
+    // reserve space for deposit facet
+    myAddressNameAdmin.reserve('depositFacet');
```
turadg, line 128:
```suggestion
    myAddressNameAdmin.reserve('smartWallet');
```

packages/agoric-cli/src/start.js
```diff
@@ -386,10 +386,12 @@ export default async function startMain(progname, rawArgs, powers, opts) {
     const newConfigToml = finishTendermintConfig({
       configToml,
       portNum,
+      enableCors: true,
     });
     const newAppToml = finishCosmosApp({
       appToml,
       portNum,
+      enableCors: true,
```
dckc: I suspect this CORS stuff is left-over from some tangentially related work. Unless you're clear on why it belongs, perhaps take it out?

packages/vats/decentral-demo-config.json
```diff
@@ -49,9 +49,6 @@
     "sharing": {
       "sourceSpec": "@agoric/vats/src/vat-sharing.js"
     },
-    "walletManager": {
-      "sourceSpec": "@agoric/vats/src/vat-walletManager.js"
-    },
```
dckc: this config probably needs the same `smart-wallet.js` contract bundle as `decntral-core-config.json`.

dckc (commented):
heading to a meeting; got thru `lib-chainStorage.js`

not sure about the CORS stuff. hm.

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -15,7 +15,7 @@ import {
 import centralSupplyBundle from '@agoric/vats/bundles/bundle-centralSupply.js';
 import {
   bridgeCoreEval,
-  makeClientManager,
+  setupClientManager,
```
dckc, line 18: why rename it? the function does make a (new) client manager.

packages/run-protocol/src/proposals/utils.js
```diff
@@ -67,17 +68,6 @@ export const reserveThenDeposit = async (
   console.info('confirmed deposit for', debugName);
 };
 
-/** @type {<T>(store: any, key: string, make: () => T) => Promise<T>} */
-const provide = async (store, key, make) => {
```
dckc, line 71: the `provide` from `@agoric/store` is not `async`. I don't see how it can work here.

packages/run-protocol/test/test-gov-collateral.js
dckc: ah. it doesn't return the thing that it creates.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -187,7 +187,7 @@ export const makeClientManager = async (
   } = {},
 ) => {
   // Create a subscription of chain configurations.
-  /** @type {SubscriptionRecord<PropertyMakers>} */
+  /** @type {SubscriptionRecord<PropertyMaker[]>} */
```
dckc, line 190: odd... I wonder why that wasn't caught before.

packages/wallet/contract/README.md
```diff
@@ -0,0 +1,56 @@
...
+make scenario2-run-client
+# confirm no errors in logs
+
+# tab 3 (interactive)
+agoric open --repl
+# confirm in browser that `home.wallet` and `home.smartWallet` exist
+agd query vstorage keys 'published.wallet'
+# confirm it has a key like `published.wallet.agoric1nqxg4pye30n3trct0hf7dclcwfxz8au84hr3ht`
+agoric follow :published.wallet.agoric1nqxg4pye30n3trct0hf7dclcwfxz8au84hr3ht
+# confirm it has JSON data
```
dckc, line 55: I think the format here is actually [Justin](https://github.com/agoric-labs/jessica/blob/master/lib/quasi-justin.js.ts). minor detail.

dckc (changes requested):
good stuff, but I think changing `provide` from async to sync has to be rolled back or otherwise fixed.

packages/vats/src/core/chain-behaviors.js
turadg: `PropertyMakers` was `PropertyMaker[]`. I changed it around to have `PropertyMaker` be named.

packages/run-protocol/src/proposals/utils.js
dckc: `store: any` was lazy (on my part)

turadg: thanks. coding too fast. thankfully the typechecker caught it too.

dckc (approved):
good stuff!

packages/vats/src/core/basic-behaviors.js
```diff
@@ -225,9 +224,39 @@ export const makeAddressNameHubs = async ({
...
+    zoe,
+  },
+  installation: {
+    consume: { singleWallet },
+  },
+}) => {
+  const STORAGE_PATH = 'wallet';
+
+  const storageNode = await getChildNode(chainStorage, STORAGE_PATH);
+  const marshaller = E(board).getPublishingMarshaller();
```
dckc, line 245: post-script: wallets have state that shouldn't be published, so using the publishing marshaller was probably not a good idea. even using the read-only marshaller would be iffy: purses shouldn't be exposed to the board vat at all.

cc @warner

dckc:
For reference, https://gist.github.com/dckc/3192a5264524635bd7261277b1fb707e has the output of `agoric follow ...`.

packages/vats/src/core/basic-behaviors.js
turadg: The marshaller privileges and what is published are independent, no?

The latter is defined here https://github.com/Agoric/agoric-sdk/pull/5701/files#diff-b1c2b2b7ded05977f142adcd00946c70016bae3a5178a46af30ef444422e3332R60-R70

Which of those shouldn't be published? `purses`?

Are you saying there should be separate streams for what goes to off-chain storage and what goes to the creator?

dckc: > The marshaller privileges and what is published are independent, no?

Yes.

And indeed, purses must not be published. Nor payments. Probably not offers either. Publishing issuers and dapps is probably harmless. contacts is iffy...

But not only should we not publish purses, we shouldn't use the board's readOnlyMarshaller, since that involves exposing the purses to the board and trusting the board not to publish them.

PR #5704, opened 2022-06-30
5346 replace uses of makeNotifierKit with makeNotifierFromSubscriber

turadg:
refs: #5346

## Description

This with https://github.com/Agoric/agoric-sdk/issues/5739 will enable https://github.com/Agoric/agoric-sdk/issues/5386

This one removes `makeNotifierKit` from VaultFactory without changing its API. Next will update the external API to provide Subscriber, for off-chain reads.

Review recommendation: by commit

### Security Considerations

--

--

### Testing Considerations

Tests needed updates for refactors, but there should be no change in behavior.
Also test locally `agoric start --reset` to ensure sim-chain hasn't broken on making storageNode and marshaller no longer optional.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -101,7 +101,7 @@ const validTransitions = {
  *
  * @typedef {{
  *   interestSnapshot: Ratio,
- *   outerUpdater: IterationObserver<VaultNotification> | null,
+ *   outerUpdater: Publisher<VaultNotification> | null,
  *   phase: VaultPhase,
  *   debtSnapshot: Amount<'nat'>,
  * }} MutableState
@@ -335,7 +335,7 @@ const helperBehavior = {
       case Phase.ACTIVE:
       case Phase.LIQUIDATING:
       case Phase.LIQUIDATED:
-        outerUpdater.updateState(uiState);
+        outerUpdater.publish(uiState);
         break;
       case Phase.CLOSED:
         outerUpdater.finish(uiState);
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -76,7 +76,7 @@ const validTransitions = {
 // XXX masks typedef from types.js, but using that causes circular def problems
 /**
  * @typedef {object} VaultManager
- * @property {() => Notifier<import('./vaultManager').AssetState>} getNotifier
+ * @property {() => Promise<Notifier<import('./vaultManager').AssetState>>} getNotifier
  * @property {(collateralAmount: Amount) => ERef<Amount<'nat'>>} maxDebtFor
  * @property {() => Brand} getCollateralBrand
  * @property {() => Brand<'nat'>} getDebtBrand
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -76,7 +76,7 @@ const validTransitions = {
 // XXX masks typedef from types.js, but using that causes circular def problems
 /**
  * @typedef {object} VaultManager
- * @property {() => Promise<Notifier<import('./vaultManager').AssetState>>} getNotifier
+ * @property {() => Subscriber<import('./vaultManager').AssetState>} getAssetSubscriber
  * @property {(collateralAmount: Amount) => ERef<Amount<'nat'>>} maxDebtFor
  * @property {() => Brand} getCollateralBrand
  * @property {() => Brand<'nat'>} getDebtBrand
@@ -528,7 +528,7 @@ const helperBehavior = {
     helper.assertCloseable();
     seat.exit();
     // eslint-disable-next-line no-use-before-define
-    const vaultKit = makeVaultKit(self, state.manager.getNotifier());
+    const vaultKit = makeVaultKit(self, state.manager.getAssetSubscriber());
     state.outerUpdater = vaultKit.vaultUpdater;
     helper.updateUiState();
 
@@ -596,7 +596,7 @@ const selfBehavior = {
     state.manager.mintAndReallocate(toMint, fee, seat, vaultSeat);
     helper.updateDebtAccounting(normalizedDebtPre, collateralPre, newDebtPre);
 
-    const vaultKit = makeVaultKit(self, state.manager.getNotifier());
+    const vaultKit = makeVaultKit(self, state.manager.getAssetSubscriber());
     state.outerUpdater = vaultKit.vaultUpdater;
     helper.updateUiState();
     return vaultKit;
--- a/packages/run-protocol/src/vaultFactory/vaultKit.js
+++ b/packages/run-protocol/src/vaultFactory/vaultKit.js
@@ -1,4 +1,5 @@
 // @ts-check
+import { makeNotifierFromSubscriber } from '@agoric/notifier';
 import '@agoric/zoe/exported.js';
 import { Far } from '@endo/marshal';
 import { makeVaultHolder } from './vaultHolder.js';
@@ -7,14 +8,14 @@ import { makeVaultHolder } from './vaultHolder.js';
  * Create a kit of utilities for use of the vault.
  *
  * @param {Vault} vault
- * @param {Promise<Notifier<import('./vaultManager').AssetState>>} assetSubscriber
+ * @param {Subscriber<import('./vaultManager').AssetState>} assetSubscriber
  */
 export const makeVaultKit = (vault, assetSubscriber) => {
   const { holder, helper } = makeVaultHolder(vault);
   const vaultKit = harden({
     publicNotifiers: {
       vault: holder.getNotifier(),
-      asset: assetSubscriber,
+      asset: makeNotifierFromSubscriber(assetSubscriber),
     },
     invitationMakers: Far('invitation makers', {
       AdjustBalances: holder.makeAdjustBalancesInvitation,
--- a/packages/run-protocol/src/vaultFactory/vaultKit.js
+++ b/packages/run-protocol/src/vaultFactory/vaultKit.js
@@ -7,14 +7,14 @@ import { makeVaultHolder } from './vaultHolder.js';
  * Create a kit of utilities for use of the vault.
  *
  * @param {Vault} vault
- * @param {Notifier<import('./vaultManager').AssetState>} assetNotifier
+ * @param {Promise<Notifier<import('./vaultManager').AssetState>>} assetSubscriber
  */
-export const makeVaultKit = (vault, assetNotifier) => {
+export const makeVaultKit = (vault, assetSubscriber) => {
   const { holder, helper } = makeVaultHolder(vault);
   const vaultKit = harden({
     publicNotifiers: {
       vault: holder.getNotifier(),
-      asset: assetNotifier,
+      asset: assetSubscriber,
     },
     invitationMakers: Far('invitation makers', {
       AdjustBalances: holder.makeAdjustBalancesInvitation,
--- a/packages/run-protocol/src/vaultFactory/vaultKit.js
+++ b/packages/run-protocol/src/vaultFactory/vaultKit.js
@@ -14,7 +14,7 @@ export const makeVaultKit = (vault, assetSubscriber) => {
   const { holder, helper } = makeVaultHolder(vault);
   const vaultKit = harden({
     publicNotifiers: {
-      vault: holder.getNotifier(),
+      vault: makeNotifierFromSubscriber(holder.getSubscriber()),
       asset: makeNotifierFromSubscriber(assetSubscriber),
     },
     invitationMakers: Far('invitation makers', {
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -672,7 +672,7 @@ const collateralBehavior = {
   makeVaultInvitation: ({ state: { zcf }, facets: { self } }) =>
     zcf.makeInvitation(self.makeVaultKit, 'MakeVault'),
   /** @param {MethodContext} context */
-  getNotifier: ({ state }) => state.assetSubscriber,
+  getSubscriber: ({ state }) => state.assetSubscriber,
   /** @param {MethodContext} context */
   getMetrics: ({ state }) => state.metricsSubscription,
   /** @param {MethodContext} context */
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -169,7 +169,10 @@ const initState = (
   storageNode,
   marshaller,
 ) => {
-  assert(storageNode && marshaller, 'Missing storageNode or marshaller');
+  assert(
+    storageNode && marshaller,
+    'VaultManager missing storageNode or marshaller',
+  );
 
   const periodNotifier = E(timerService).makeNotifier(
     0n,
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -89,14 +89,14 @@ const trace = makeTracer('VM');
  * debtBrand: Brand<'nat'>,
  * debtMint: ZCFMint<'nat'>,
  * factoryPowers: import('./vaultDirector.js').FactoryPowersFacet,
- * marshaller?: ERef<Marshaller>,
+ * marshaller: ERef<Marshaller>,
  * metricsPublication: IterationObserver<MetricsNotification>,
  * metricsSubscription: StoredSubscription<MetricsNotification>,
  * periodNotifier: ERef<Notifier<bigint>>,
  * poolIncrementSeat: ZCFSeat,
  * priceAuthority: ERef<PriceAuthority>,
  * prioritizedVaults: ReturnType<typeof makePrioritizedVaults>,
- * storageNode?: ERef<StorageNode>,
+ * storageNode: ERef<StorageNode>,
  * zcf: import('./vaultFactory.js').VaultFactoryZCF,
  * }>} ImmutableState
  */
@@ -155,8 +155,8 @@ const provideEphemera = makeEphemeraProvider(() => ({
  * @param {import('./vaultDirector.js').FactoryPowersFacet} factoryPowers
  * @param {ERef<TimerService>} timerService
  * @param {Timestamp} startTimeStamp
- * @param {ERef<StorageNode>} [storageNode]
- * @param {ERef<Marshaller>} [marshaller]
+ * @param {ERef<StorageNode>} storageNode
+ * @param {ERef<Marshaller>} marshaller
  */
 const initState = (
   zcf,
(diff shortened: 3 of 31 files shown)


packages/run-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -965,8 +965,11 @@ test('interest on multiple vaults', async t => {
   await manualTimer.tickN(8);
   await eventLoopIteration();
 
+  console.log('DEBUG getUpdateSince() on assetNotifier');
```
turadg: @gibson042 the following line hangs just as in https://github.com/Agoric/agoric-sdk/pull/5695/files#r911476188

packages/notifier/src/storesub.js
```diff
@@ -2,8 +2,85 @@
...
+ * @param {ERef<ReturnType<typeof makeMarshal>>} [marshaller]
+ * @returns {StoredSubscriber<T>}
+ */
+export const makeStoredSubscriber = (
+  subscriber,
+  storageNode,
+  marshaller = makeMarshal(undefined, undefined, {
+    marshalSaveError: () => {},
+  }),
+) => {
```
turadg: mostly a copy of @michaelfig 's `makeStoredPublisherKit`

packages/notifier/src/storesub.js
```diff
@@ -2,8 +2,85 @@
...
+  }
+
+  /** @type {StoredSubscriber<T>} */
+  const storesub = Far('StoredSubscriber', {
+    ...subscriber,
+    getStoreKey: () => {
+      if (!storageNode) {
+        return {};
+      }
+      return E(storageNode).getStoreKey();
```
gibson042: I'm surprised by this divergence between `storageNode` being falsy vs. fulfilling with a falsy value, but it probably won't come up in practice.

packages/notifier/src/storesub.js
```diff
@@ -2,8 +2,85 @@
...
+  };
+
+  // Must *not* be an async function, because it sometimes must throw to abort
+  // the iteration.
+  const publishValue = obj => {
+    assert(storageNode);
+    if (publishFailed) {
+      // To properly abort the iteration, this must be a synchronous exception.
+      throw publishException;
+    }
```
gibson042: This looks racy, and probably won't work correctly in cases where an error from `storageNode` is conditional or where it is encountered early in a final synchronous batch of updates (e.g., `publish(inputResponsibleForError); publish(inputProcessedBeforePublishFailedIsSet)`). Shouldn't we instead update whatever is requiring the error to be synchronous (`observeIteration`?), and also ensuring serial processing of the writes?

packages/notifier/src/storesub.js
```diff
@@ -115,3 +194,31 @@ export const makeStoredPublisherKit = (storageNode, marshaller, childPath) => {
...
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {string} [childPath]
+ * @returns {PublishKit<T>}
+ */
+export const makeStoredPublishKit = (storageNode, marshaller, childPath) => {
+  const { publisher, subscriber } = makePublishKit();
+
+  if (storageNode && childPath) {
+    storageNode = E(storageNode).getChildNode(childPath);
```
gibson042: I think `getChildNode` should be the caller's responsibility rather than bundled in here.

packages/notifier/src/storesub.js
```diff
@@ -115,3 +194,31 @@ export const makeStoredPublisherKit = (storageNode, marshaller, childPath) => {
...
+ * @template [T=unknown]
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {string} [childPath]
+ * @returns {PublishKit<T>}
+ */
+export const makeStoredPublishKit = (storageNode, marshaller, childPath) => {
+  const { publisher, subscriber } = makePublishKit();
+
+  if (storageNode && childPath) {
```
gibson042: Under what circumstances should it be acceptable to invoke `makeStoredPublishKit` without a valid storage node? The name implies that `storageNode` should be required. And likewise for `makeStoredSubscriber`.

packages/notifier/src/storesub.js
```diff
@@ -115,3 +194,31 @@ export const makeStoredPublisherKit = (storageNode, marshaller, childPath) => {
...
+ * What's different is this tees records, writing out to storageNode.
+ *
+ * @template [T=unknown]
+ * @param {ERef<StorageNode>} [storageNode]
+ * @param {ERef<Marshaller>} [marshaller]
+ * @param {string} [childPath]
+ * @returns {PublishKit<T>}
+ */
+export const makeStoredPublishKit = (storageNode, marshaller, childPath) => {
+  const { publisher, subscriber } = makePublishKit();
```
gibson042: On master, [`makePublishKit`](https://github.com/Agoric/agoric-sdk/blob/4cca744382fc465492153a5e9d4919edefe9c637/packages/notifier/src/publish-kit.js#L203) is differentiated from `makeEmptyPublishKit` (the former requiring an initial value, the latter not). Given the presence of such variation, I think this function is a little too specialized and it would probably be better to drop it in favor of callers handling publish kit creation and wiring into chain storage on their own, e.g.
```js
const { publisher, baseSubscriber } = makeEmptyPublishKit(); // or makePublishKit(initialValue)
const storedSubscriber = makeStoredSubscriber(baseSubscriber, storageNode, marshaller);
// Protect publisher, share storedSubscriber.
```

packages/notifier/src/types.js
```diff
@@ -241,3 +241,8 @@
  * @template T
  * @typedef {ConsistentAsyncIterable<T> & StoredFacet} StoredSubscription
  */
+
+/**
+ * @template T
+ * @typedef {StoredFacet & Subscriber<T>} StoredSubscriber
+ */
```
gibson042: :+1:

packages/notifier/test/test-stored-subscription.js
```diff
@@ -80,3 +86,56 @@ test('stored subscription', async t => {
...
+  const { publication: pubStorage, subscription: subStorage } =
+    makeSubscriptionKit();
+  const storage = makeFakeStorage('publish.foo.bar', pubStorage);
+
+  const { publisher, subscriber } = makePublishKit();
+
+  publisher.publish(initialValue);
+  const storesub = makeStoredSubscriber(subscriber, storage);
+
+  t.deepEqual(await E(storesub).getStoreKey(), await E(storage).getStoreKey());
```
gibson042: This may be highlighting a gap in `makeFakeStorage`; I would expect the key to be defined as part of its initialization rather than dynamic like this.

packages/notifier/test/test-stored-subscription.js
```diff
@@ -80,3 +86,56 @@ test('stored subscription', async t => {
...
+  const unserializer = E(storesub).getUnserializer();
+
+  const notifier = makeNotifierFromSubscriber(subscriber);
+  const ait = E(notifier)[Symbol.asyncIterator]();
+  const storeAit = subStorage[Symbol.asyncIterator]();
```
gibson042: I find these declarations a bit confusing. Suggestion:
```suggestion
  const storedSubscriberIterable = subscribeEach(subscriber);
  const storedSubscriberIterator = E(storedSubscriberIterable)[Symbol.asyncIterator]();
  const storageIterator = subStorage[Symbol.asyncIterator]();
```

packages/notifier/test/test-stored-subscription.js
```diff
@@ -80,3 +86,56 @@ test('stored subscription', async t => {
...
+
+  const { unserialize } = makeMarshal();
+  const check = async (description, origValue, expectedDone) => {
+    const { done: storedDone, value: storedEncoded } = await storeAit.next();
+    t.assert(!storedDone, `check not stored done for ${description}`);
```
gibson042:
```suggestion
    t.assert(!storedDone, `storage iterator should not report done for ${description}`);
```

packages/notifier/src/storesub.js
turadg: do you mean you'd expect this to return falsy when storageNode is falsy? agreed. I tried returning null but ran into a type error.

I'll leave this conversation open to come back to it. Let me know if you're talking about somethign else

packages/notifier/src/storesub.js
turadg: https://github.com/Agoric/agoric-sdk/pull/5704/commits/060dbf257139a9c407e8e1538620cea7b82f96e8 removed `makeEmptyPublishKit` because all our publish kits are to start empty

This function is a convenience for what contracts have to do a lot of. If you request that it move out`@agoric/notifier` I'll move it to run-protocol and not export it.

packages/notifier/src/storesub.js
turadg: > Under what circumstances should it be acceptable to invoke makeStoredPublishKit without a valid storage node?

If the that term wasn't supplied to the contract. Bootstrap doesn't always have it.

> The name implies that storageNode should be required. And likewise for makeStoredSubscriber

It is returning a `StoredSubscriber` just one that doesn't write. The type `StoredSubscriber` is far with methods like `getStoreKey()`.

packages/notifier/src/storesub.js
turadg: I don't know. Most of this function is just copied from `makeStoredPublisherKit` @michaelfig wrote in https://github.com/Agoric/agoric-sdk/pull/5400/files#diff-8f699ceaf58ba86dc7d1914c2642f6826c5e944f4ce01617526ea4008a1eef1bR43-R48

packages/notifier/src/storesub.js
turadg: I assume you have the same feedback for `makeStoredPublisherKit`. I have a use for them in contracts but I respect the idea that this core package should be conservative in its API surface.

dckc:
use case: list of vaults by risk

cc @btulloh

dckc:
Where is the code that makes a `published.vault.23423423432` thingy? Or is that yet to be done?

packages/notifier/src/storesub.js
gibson042: That, and also that a falsy storageNode is accepted while a storageNode that resolves to a falsy value (e.g., `Promise.resolve(undefined)`) results in an exception. I think this also intersects my concern about allowing invocation without a valid storage node, which seems to unnecessarily complicate things (e.g., note how the output from `getStoreKey` is meaningless in such a configuration).

turadg:
> Where is the code that makes a published.vault.23423423432 thingy? Or is that yet to be done?

YTBD. Right now just trying to refactor existing behavior to use `makeNotifierFromSubscriber`. After that I can come back and put in the proper storage nodes and add tests.

packages/notifier/src/storesub.js
gibson042: Yes, that may be a more reasonable starting point.

packages/notifier/src/storesub.js
turadg: I agree with the perspective. In production, the storageNode and marshaller should always be there so it would be fine to require it. In tests, they aren't always. I think it would be better to solve that, but pragmatically I'm not sure it's worth it.

If I can't remove the kludge in the tests, I'll at least remove it from `@agoric/notifier` and make the contracts deal with it.

dckc: > In production, the storageNode and marshaller should always be there

We have the sim chain, which while perhaps not production, is supported for smart contract development work. In the sim chain, there's no chainStorage.

turadg: Thanks, that's the thing I was trying to remember. Since it's a "sim" chain, would it be reasonable to _simulate_ chainStorage? E.g. in-memory stoageNode, identity marshaller.

dckc: oh. yeah. in-memory chainStorage would be great.

identity doesn't work as a marshaller, cuz the output has to be a string and the input is not constrained that way. But I think you can always use a vanilla output from `makeMarshal()`.

turadg: Made it required and updated bootstrap to always provide it [cbaa70b](https://github.com/Agoric/agoric-sdk/pull/5704/commits/cbaa70b6a3f7498c18f26aa85e8ee4a80088d797), falling back to a no-op "NullStorageNode" when ChainStorage isn't available.

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -76,7 +76,7 @@ const validTransitions = {
 // XXX masks typedef from types.js, but using that causes circular def problems
 /**
  * @typedef {object} VaultManager
- * @property {() => Notifier<import('./vaultManager').AssetState>} getNotifier
+ * @property {() => Promise<Notifier<import('./vaultManager').AssetState>>} getNotifier
```
dckc: we're keeping this notifier API for now? Is it deprecated?

dckc: ah... right... this is where prefix-lossy comes in.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -581,7 +588,7 @@ const managerBehavior = {
     state.totalDebt = AmountMath.subtract(state.totalDebt, toBurn);
   },
   /** @param {MethodContext} context */
-  getNotifier: ({ state }) => state.assetNotifier,
+  getNotifier: ({ state }) => makeNotifierFromSubscriber(state.assetSubscriber),
```
dckc: Perhaps add a "we make a new notifier on each call that has no memory of previous states" comment?

The difference between these is subtle. I was a about to ask why we make a new one on each call when I remembered the whole prefix-lossy concept.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -665,7 +672,7 @@ const collateralBehavior = {
   makeVaultInvitation: ({ state: { zcf }, facets: { self } }) =>
     zcf.makeInvitation(self.makeVaultKit, 'MakeVault'),
   /** @param {MethodContext} context */
-  getNotifier: ({ state }) => state.assetNotifier,
+  getNotifier: ({ state }) => state.assetSubscriber,
```
dckc: is this an internal API? rename it to `getSubscriber`?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -662,7 +666,8 @@ test('update liquidator', async t => {
     }),
   );
   await eventLoopIteration();
-  govNotify = await d.managerNotified();
+  // ??? why is AT_NEXT necessary now
```
dckc: I don't know; I wonder if I'm sufficiently up-to-speed to review this...

packages/run-protocol/src/vaultFactory/vaultKit.js
```diff
@@ -7,14 +8,14 @@ import { makeVaultHolder } from './vaultHolder.js';
...
- * @param {Promise<Notifier<import('./vaultManager').AssetState>>} assetSubscriber
+ * @param {Subscriber<import('./vaultManager').AssetState>} assetSubscriber
  */
 export const makeVaultKit = (vault, assetSubscriber) => {
   const { holder, helper } = makeVaultHolder(vault);
   const vaultKit = harden({
     publicNotifiers: {
       vault: holder.getNotifier(),
-      asset: assetSubscriber,
+      asset: makeNotifierFromSubscriber(assetSubscriber),
```
dckc, line 18: we make this notifier just once? not a new prefix-lossy one on demand?

packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js
```diff
@@ -50,7 +50,7 @@ export async function start(zcf, privateArgs) {
 
   const { zcfSeat: stage } = zcf.makeEmptySeatKit();
 
-  const { notifier: managerNotifier } = makeNotifierKit();
+  const { subscriber: assetSubscriber } = makePublishKit();
```
dckc, line 53: we throw away the publisher? how does it every get updated?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -571,10 +571,15 @@ test('price drop', async t => {
...
-    locked: aeth.makeEmpty(),
-    updateCount: undefined,
-  });
+  await dv.notified(
+    Phase.CLOSED,
+    {
+      locked: aeth.makeEmpty(),
+      updateCount: undefined,
+    },
+    // ??? why is AT_NEXT necessary now
```
dckc: again, good question

packages/zoe/src/contractSupport/priceAuthority.js
```diff
@@ -31,7 +31,7 @@ const isGT = (amount, amountLimit) => !AmountMath.isGTE(amountLimit, amount);
 /**
  * @typedef {object} OnewayPriceAuthorityOptions
  * @property {Issuer} quoteIssuer
- * @property {ERef<Notifier<Timestamp>>} notifier
+ * @property {ERef<Notifier<unknown>>} notifier
```
dckc, line 34: This is a relaxation of a type constraint? Any particular reason?

packages/vats/src/core/chain-behaviors.js
```diff
@@ -295,7 +295,7 @@ export const makeChainStorage = async ({
   const bridgeManager = await bridgeManagerP;
   if (!bridgeManager) {
     console.warn('Cannot support chainStorage without an actual chain.');
-    chainStorageP.resolve(undefined);
+    chainStorageP.resolve(null);
```
dckc, line 298: why change `undefined` to `null`?

You were considering an in-memory chainStorage; that didn't work out?

packages/run-protocol/src/vaultFactory/vault.js
dckc: and it looks like it went away in a later commit

packages/run-protocol/src/vaultFactory/vaultManager.js
dckc: looks like it got renamed later

packages/run-protocol/test/vaultFactory/test-liquidator.js
dckc: looks like that went away in a later commit

dckc (approved):
Looks pretty good... at least: I don't see any serious problems.

I'm a little fuzzy on the details; I'm relying somewhat on test coverage, static typing, and discussion with @gibson042 and co.

packages/vats/src/core/chain-behaviors.js
turadg: Because when there's no chain storage available, it's _defined_ to be missing. I think that's better expressed as null.

packages/zoe/src/contractSupport/priceAuthority.js
turadg: It was wrong

packages/run-protocol/test/vaultFactory/vault-contract-wrapper.js
turadg: It doesn't, this is the fake contract

packages/run-protocol/src/vaultFactory/vaultKit.js
turadg: right, this is interim. the contract shouldn't be making Notifiers at all because that costs state tracking. Change in next PR.

PR #5705, opened 2022-06-30
fix(telemetry): do not mutate the original slog object

mhofman:
refs: #5637

## Description

I had a stupid error in #5637 which fell through the cracks since the ingest script didn't harden it's the slog object like the kernel does.

### Security Considerations

Don't mutate objects!

None

### Testing Considerations

Updated the test tools (ingest script) to harden and prevent future issues like this.


warner (approved):
LGTM

Issue #5703, opened 2022-06-30
record bundleID of lockdown/supervisor/liveslots for each vat


warner:
## What is the Problem Being Solved?

The kernel must be capable of reloading each living vat from state in the DB, and bringing it back to the same state it was in from the previous checkpoint. The new vat runtime is built out of several layers, all of which must be identical on each execution:
* the `xsnap` process (compiled from a specific version of the XS Javascript engine, plus our `xsnap` C code application)
* the Endo `lockdown` bundle we evaluate within that process, to provide SES
* the `supervisor` bundle we load, to define the code that processes incoming netstring commands and conveys syscalls back to the kernel process
* the `liveslots` code (currently incorporated into the supervisor bundle, which `import`s it)
* the vat bundle
* all the old deliveries into the vat
* the latest heap snapshot, which encapsulates all of the above
* the most recent deliveries into the vat (since the most recent snapshot)

Currently we bundle `lockdown` and `supervisor` during kernel initialization (exactly once, while the DB is still empty), and store the bundles into the DB using calls like `kvStore.set('lockdownBundle', bundle)`.

This consistency is helpful, but we're looking ahead to the time when we want an established application (our chain) to start using an upgraded set of code. Existing vats must continue to run without behavior changes, so they must keep using the same `lockdown`/`supervisor`/liveslots as before, but new vats created after the upgrade point should use (or at least be able to use) new starting points.

#4376 is about the API to tell the kernel to start using a new version of liveslots (and about separating the liveslots code from the `supervisor` bundle, and installing it separately). But we also need to keep track of the lockdown/supervisor bundles as part of each vat's state, so some vats can use one version, while newer vats use different ones.

## Description of the Design

During `initializeSwingset`, after we create the initial `lockdown` and `supervisor` bundles, we should compute their `bundleID` hashes, and store them into the bundle table (just like vat bundles and contract bundles, with an internal equivalent of `controller.validateAndInstallBundle()`). We should then store their bundleIDs as `kvStore.set('currentLockdownBundleID')`/etc.

Each time we create a new vat, we should copy the current bundle IDs into vat state, probably `vatKeeper.setSourceAndOptions`, stashing the IDs into the `options` somewhere.

[...]

Issue #4376, opened 2022-01-25
upgrading liveslots: controller.setLiveslotsBundleID()


warner:
## What is the Problem Being Solved?

As with #4375 to upgrade the kernel, it would be nice to be able to upgrade the version of liveslots that we use in all vats.

Upgrading liveslots, without also changing the vat bundle, could be used to add new features to liveslots, or to fix a bug that the vat has not yet provoked. It can only work if the vat is already prepared for upgrade (by recording its important state in Durable collections, and providing references to them in the #4325 "baggage", and by having a starting point which knows how to reconnect everything).

The idea is:
* we store the liveslots bundle into the bundle store, just like vat bundles (in #4372) and the kernel bundle (#4375)
* each vat's kvStore includes a key that says what bundleID we use to get liveslots
* each vat-worker supervisor is modified to accept this bundle from the kernel, over the kernel-to-worker control pipe, instead of having an `import { makeLiveSlots } from '.../liveslots.js'` which drags it into the supervisor bundle
  * (come to think of it, it might be interesting to manage the supervisor bundle this way too)
* the kvStore has an additional key that tells the kernel what liveslotsBundleID to use for new vats
* we add a controller API to change this bundleID to something new

So once a vat is created, it keeps using the same liveslots, even if some new version is installed in the application. Various vats can use various versions of liveslots, depending upon whatever was current when they were first created.

Later, when you want to upgrade the liveslots in a given vat, you do a vat-upgrade but leave the vat source code untouched (or, you upgrade liveslots at the same time as you upgrade the vat bundle). This requires the kind of upgrade where we delete the heap snapshot and delete the transcript, and start the vat over from a new (/old) bundle and let it recover from durable data.

I suspect that most existing vats won't really benefit from upgrading liveslots, however being able to upgrade the vat code and have it take advantage of new liveslots features would be really valuable. So I'm thinking that `vatAdminFacet~.upgrade(newVatBundleID, { liveslotsVersion: 2 })` would be great. (Or, maybe vats automatically get the latest liveslots when their vat bundles are upgraded).

## Security Considerations

We might consider allowing new vats to be created with an arbitrary bundle to use as liveslots. Liveslots enforces ocap security within a vat, so the code within the vat must not be able to tamper with the liveslots that supports it (else it could violate ocap rules).

[...]

PR #5706, opened 2022-06-30
fix(swingset): stop using a persistent kernel bundle

### LINKED ISSUE #5679: unpersist the kernel bundle (was: API to trigger kernel rebundling)

How should a chain upgrade cause a simultaneous kernel upgrade?

For POLA purposes, the swingset kernel is loaded into its own SES `Compartment`, and `importBundle()` is the simplest way to load arbitrary module graphs into SES compartments, which means the kernel source code must be collected into a Bundle first. This bundling can take several seconds, so to save time during application startup, we don't want to re-bundle the kernel on every application launch. Currently, we bundle the kernel exactly once, the first time the host application ever runs, when it calls `initializeSwingset()` (which starts from an empty database and initializes all of the empty tables and `0`-valued counters, etc). The bundle is stored into the database with `kvStore.set('kernelBundle', bundle)`, and read from the DB on every subsequent application launch (`importBundle(kvStore.get('kernelBundle'))`). (We manage the `lockdown` and `supervisor` bundles needed for the xsnap worker in the same way).

#4375 is a proposal to allow the host application to deliberately change which kernel bundle is used, through a new API on the `controller` object. Half of this proposal seems fine: storing the kernel bundle(s) in the general bundle table (next to vat bundles), and only storing the BundleID in the `kvStore`. However the external API feels at odds with the way chain operators expect to upgrade their validators.

Many of these validator operators use a version-management tool (["cosmovisor"](https://github.com/cosmos/cosmos-sdk/tree/main/cosmovisor) in conjunction with a standard "keep the process running" supervisor like `systemd`. The cosmovisor tool watches for governance votes that call for a wholesale replacement of the validator software (via upgrade plans written out by the `x/upgrade` module). Operators are responsible for paying attention to these votes and placing the new binaries in a special directory. The chain knows to halt at the approved block height if it is not running the correct (newer) software. The cosmovisor tool knows how to switch out symlinks at the right time, so that any new launches of the validator software will use the new version, rather than the old. Together, the old validator exits at the apointed time, the supervisor reacts to a stopped validator as usual (it attempts to relaunch the program), which cause the new validator version to be launched, which then passes the "am I running the right version after `blockHeightTrigger`" check, allowing it to keep running. The block height is chosen to allow enough time (perhaps a few days) for the operator to download and verify the new code before the cutover must happen.

If swingset re-bundled its kernel code on every application launch, this would work fine: the new validator version would include new kernel code (probably a whole copy of `agoric-sdk`, as installed into various `node_modules/` directories), all re-launches of the validator after that point would get the new kernel code instead of the old kernel code, and the resulting kernel behavior would change at the right moment. However 1: rebundling the kernel is uncomfortable slow to do on every app launch (and adds to the downtime caused by simple reboots), and 2: if we're at all worried about inconsistent kernel BundleIDs (does rebundling the same source code always create the same bundle/ID?), then rebundling on every reboot would make it worse (although I think we already depend upon consistent bundle IDs, because the `kvStore.set('kernelBundle')` is included in the initial crankHash/activityHash).

So swingset needs a signal that says "this is a good time to rebundle". It needs to happen at least every time the kernel code (or its dependencies) changes. And it really shouldn't happen until the validator is willing to accept a behavior change, i.e. when the chain is upgraded.

I'm thinking that the host application should supply this signal, in the form of a `controller.something(versionstring)` API call. Swingset will take the supplied string and compare it against a value stored in the DB. If they match, nothing happens. If they differ, swingset will rebundle the kernel, store the new bundle for use upon next reboot, and update the DB entry (so the next invocation won't cause a rebundling). The host application should call this at each application restart, before launching swingset (so the new bundle will be used for the subsequent kernel import). 

The host application source code should be modified to provide a different string every time the host application's packaging (`package.json`) is changed to request a different version of the swingset package.

The string could be the swingset version, or it could be a simple counter: anything that is easy to change whenever there's any chance that the swingset kernel code or its bundled dependencies might change.

### Alternatives

It is tricky for a package to introspect its own version number from it's packaging metadata (at least in the [languages](https://github.com/python-versioneer/python-versioneer) I'm most familiar with, perhaps `JSON.parse(fs.readFileSync(new URL('../../package.json', import.meta.url).pathname)).version` would work), and even that only gets you a change when the author of the package decides to make a release.

I expect we will be testing chain upgrades from one git revision of the monorepo to another, without necessarily doing a full release process (bumping version numbers) on all packages first. And I suspect other users of Swingset will have similar needs, where they want to update Swingset-using applications in a way (perhaps by applying new commits to a private fork of the swingset-bearing monorepo, or by using `patch-package` to modify the local `node_modules/` copy of `@agoric/swingset-vat`) that does not create a version bump in the swingset packaging metadata.

So relying upon swingset itself to know when the code has changed doesn't seem like a good idea.

On the other hand, it's a little weird to see a swingset version/etc in the downstream client (host application) code, at the call site which interacts with `controller`. And there's a danger of the author who updates the downstream application's packaging (to use a new swingset version) forgetting to update that call site to reflect the change.

### More general kernel upgrade tasks

We haven't really thought through the larger-scale lifetime of a swingset kernel. We expect kernel data structures to evolve, and the DB state created today may be very different than the state used six months from now. Sometimes a new swingset will be created (the DB initialized) with one version, and then later used by a different version, which needs to be capable of handling the older data.

This might take the form of discrete DB versions: we introduce a `kvStore.get('version')` key, and each time the kernel launches, it checks the value, and performs some sort of upgrade-in-place process if it differs (and then updates the key). This would not require any special care by the host application, but would introduce an unpredictable stall in launch time, depending upon when upgrades happen and how much data must be upgraded. It also requires diligence on the kernel authors, to update the expected value every time the DB schema (or semantics) changes. Thus far we've been pretty casual about kernel changes, because most instances are just running from the current repository trunk, and actual kernel upgrades are still in our future.

Instead, we might aim for fine-grained eternal upgrade: each kernel component that uses the DB should be prepared to see data from arbitrary points in the kernel's history. This requires a different sort of diligence: no discrete schema changes, but constant backwards-compatibility.

Or, we might use the "now is a good time to re-bundle the kernel" signal to double as a "now is a good time to see if your DB needs upgrading" signal. This might move the unpredictable-duration upgrade stall to at least happen at a predictable time, and it would make the host application's operator more involved in data migration activity.

## Description of the Design

With the scheme I'm currently favoring, the changes look like this:

* `initializeSwingset()` grows an option to provide an initial `kernelBundleSequence` string, which defaults to an empty string
* `initializeSwingset()` bundles the kernel sources, installs the bundle in the bundle table, stores the bundleID in `kvStore.set('kernelBundleID')`, and stores the provided/default `kernelBundleSequence` string into `kvStore.set('kernelBundleSequence')`
* we add `controller.maybeRebundleKernel(sequenceString)`, which compares `sequenceString` against the saved `kernelBundleSequence`, and if different: rebundle the kernel sources, install the new bundle, store the bundleID, decref the old bundleID, update `kernelBundleSequence`
* 
* our host application (cosmic-swingset) should add  `kernelBundleSequence: '0'` to the options it gives `initializeSwingset()`
* cosmic-swingset also adds a call to `controller.maybeRebundleKernel('0')` and a DB commit to the code in `launch-chain.js`, just before the call to `makeSwingsetController()`
* 
* we update the release process checklist/automation to increment both `'0'` strings just before we change `cosmic-swingset/package.json` to use a different version of `@agoric/swingset-vat`

## Security Considerations

No security considerations per se, but consensus considerations require very explicit transitions from one version of the source code to another. I *think* this would make the validator upgrade process pretty straightforward, but we should think through the release-process changes and decide if this seems robust enough to make sure the kernel is rebundled whenever necessary.

## Test Plan

Unit tests which invoke `maybeRebundleKernel()` can exercise the code, however since the kernel code won't change between bundles, I'm not sure how to confirm that it really worked. Perhaps we use automated tests to make sure it doesn't explode, but use a manual test to make sure the bundle ID actually changes. The manual test would deliberately modify a file in the source tree (by adding a newline to `kernel.js`, creating unstaged/uncommitted changes in `git status`) after making the first bundle and before making the second bundle. The developer running the test would need to revert those changes before continuing.

cc @michaelfig @mhofman @FUDCo for thoughts

warner:
Previously, the kernel sources were bundled during
`initializeSwingset()`, and stored in the kvStore portion of the
swingstore kernelDB. It was read from that DB each time
makeSwingsetController() launched the application process.

That ensured the kernel behavior would remain consistent (and
deterministic) even if the user's source tree changed. Using a
persistent kernel bundle also speeds up application launch: we save
about 1.8s by not re-running `bundleSource` on each launch.

Once upon a time I thought this consistency was important/useful, but
now I see that it causes more problems than it's worth. In particular,
upgrading a chain requires an awkward "rekernelize" action to replace
the kernel bundle in the database. And we believe we can reclaim about
1.0s per launch by moving to a not-yet-written `importLocation()` Endo
tool that loads module graphs from disk without serializing the
archive in the middle.

This patch changes swingset to stop using a persistent kernel
bundle. Each application launch (i.e. `makeSwingsetController()` call)
re-bundles the kernel sources just before importing that bundle, so
that updated source trees are automatically picked up without
additional API calls or rekernelize steps. The bundle is no longer
stored in the kvStore.

The other bundles (vats, devices, xsnap helpers) are still created
during `initializeSwingset` and stored in the DB, since these become
part of the deterministic history of each vat (#4376 and #5703 will
inform changes to how those bundles are managed). Only the kernel
bundle has become non-persistent.

Bundling is slow enough that many unit tests pre-bundle both the
kernel and the sources of built-in vats, devices, and xsnap helper
bundles. This provides a considerable speedup for tests that build (or
launch) multiple kernels within a single test file. These
`kernelBundles` are passed into `initializationOptions`, which then
bypassed the swingset-bundles-it-for-you code. This saves a minute or
two when running the full swingset test suite.

Unit tests can still use `initializationOptions` to amortize bundling
costs for the non-kernel bundles. `runtimeOptions` now accepts a
`kernelBundle` member to amortize bundling the kernel itself. Tests
which use this trick and also use `makeSwingsetController()` need to
be updated (else they'll run 1.8s slower). All such tests were inside
`packages/SwingSet/test/` and have been updated, which would otherwise
cause the test suite to run about 50-60s slower.

The other agoric-sdk packages that do swingset-style tests are mostly
using the deprecated `buildVatController()`, whose options continue to
accept both kinds of bundles, and do not need changes.

closes #5679

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -26,6 +26,7 @@ import { makeGcAndFinalize } from '../lib-nodejs/gc-and-finalize.js';
 import { insistStorageAPI } from '../lib/storageAPI.js';
 import { provideHostStorage } from './hostStorage.js';
 import {
+  buildKernelBundle,
   swingsetIsInitialized,
   initializeSwingset,
 } from './initializeSwingset.js';
@@ -168,7 +169,8 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
  *   warehousePolicy?: { maxVatsOnline?: number },
  *   overrideVatManagerOptions?: unknown,
  *   spawn?: typeof import('child_process').spawn,
- *   env?: Record<string, string | undefined>
+ *   env?: Record<string, string | undefined>,
+ *   kernelBundle?: Bundle
  * }} runtimeOptions
  */
 export async function makeSwingsetController(
@@ -238,6 +240,10 @@ export async function makeSwingsetController(
   // see https://github.com/Agoric/SES-shim/issues/292 for details
   harden(console);
 
+  writeSlogObject({ type: 'bundle-kernel-start' });
+  const { kernelBundle = await buildKernelBundle() } = runtimeOptions;
+  writeSlogObject({ type: 'bundle-kernel-finish' });
+
   // FIXME: Put this somewhere better.
   const handlers = process.listeners('unhandledRejection');
   let haveUnhandledRejectionHandler = false;
@@ -254,8 +260,6 @@ export async function makeSwingsetController(
   function kernelRequire(what) {
     assert.fail(X`kernelRequire unprepared to satisfy require(${what})`);
   }
-  // @ts-expect-error assume kernelBundle is set
-  const kernelBundle = JSON.parse(kvStore.get('kernelBundle'));
   writeSlogObject({ type: 'import-kernel-start' });
   const kernelNS = await importBundle(kernelBundle, {
     filePrefix: 'kernel/...',
@@ -515,7 +519,7 @@ export async function makeSwingsetController(
  * @param {SwingSetConfig} config
  * @param {string[]} argv
  * @param {{ hostStorage?: HostStore, env?: Record<string, string>, verbose?:
- *   boolean, kernelBundles?: Record<string, string>, debugPrefix?: string,
+ *   boolean, kernelBundles?: Record<string, Bundle>, debugPrefix?: string,
  *   slogCallbacks?: unknown, testTrackDecref?: unknown, warehousePolicy?: {
  *   maxVatsOnline?: number }, slogFile?: string }} runtimeOptions
  * @typedef { import('@agoric/swing-store').KVStore } KVStore
@@ -529,12 +533,15 @@ export async function buildVatController(
     hostStorage = provideHostStorage(),
     env,
     verbose,
-    kernelBundles,
+    kernelBundles: kernelAndOtherBundles = {},
     debugPrefix,
     slogCallbacks,
     warehousePolicy,
     slogFile,
   } = runtimeOptions;
+  const { kernel: kernelBundle, ...otherBundles } = kernelAndOtherBundles;
+  const kernelBundles = runtimeOptions.kernelBundles ? otherBundles : undefined;
+
   const actualRuntimeOptions = {
     env,
     verbose,
@@ -542,6 +549,7 @@ export async function buildVatController(
     slogCallbacks,
     warehousePolicy,
     slogFile,
+    kernelBundle,
   };
   const initializationOptions = { verbose, kernelBundles };
   let bootstrapResult;
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -28,43 +28,59 @@ const { keys, values, fromEntries } = Object;
 const allValues = async obj =>
   fromEntries(zip(keys(obj), await Promise.all(values(obj))));
 
+const bundleRelative = rel =>
+  bundleSource(new URL(rel, import.meta.url).pathname);
+const bundleRelativeGE = rel =>
+  bundleSource(new URL(rel, import.meta.url).pathname, { format: 'getExport' });
+
 /**
- * Build the source bundles for the kernel and xsnap vat worker.
- *
- * @param {object} [options]
- * @param {ModuleFormat} [options.bundleFormat]
+ * Build the source bundles for the kernel. makeSwingsetController()
+ * calls this on each launch, to get the
+ * current kernel sources
  */
-export async function buildKernelBundles(options = {}) {
-  // this takes 2.7s on my computer
+export async function buildKernelBundle() {
+  // this takes about 1.0s on my computer
+  const kernelBundle = await bundleRelative('../kernel/kernel.js');
+  return harden(kernelBundle);
+}
 
-  const { bundleFormat = undefined } = options;
-
-  const src = rel =>
-    bundleSource(new URL(rel, import.meta.url).pathname, {
-      format: bundleFormat,
-    });
-  const srcGE = rel =>
-    bundleSource(new URL(rel, import.meta.url).pathname, {
-      format: 'getExport',
-    });
-
+/**
+ * Build the source bundles for built-in vats and devices, and for the
+ * xsnap vat worker.
+ */
+export async function buildVatAndDeviceBundles() {
   const bundles = await allValues({
-    kernel: src('../kernel/kernel.js'),
-    adminDevice: src('../devices/vat-admin/device-vat-admin.js'),
-    adminVat: src('../vats/vat-admin/vat-vat-admin.js'),
-    comms: src('../vats/comms/index.js'),
-    vattp: src('../vats/vattp/vat-vattp.js'),
-    timer: src('../vats/timer/vat-timer.js'),
+    adminDevice: bundleRelative('../devices/vat-admin/device-vat-admin.js'),
+    adminVat: bundleRelative('../vats/vat-admin/vat-vat-admin.js'),
+    comms: bundleRelative('../vats/comms/index.js'),
+    vattp: bundleRelative('../vats/vattp/vat-vattp.js'),
+    timer: bundleRelative('../vats/timer/vat-timer.js'),
 
-    lockdown: srcGE(
+    lockdown: bundleRelativeGE(
       '../supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js',
     ),
-    supervisor: srcGE(
+    supervisor: bundleRelativeGE(
       '../supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js',
     ),
   });
 
   return harden(bundles);
+}
+
+// Unit tests can call this to amortize the bundling costs: pass the
+// result to initializeSwingset's initializationOptions.kernelBundles
+// (for the vat/device/worker bundles), and you can pass .kernelBundle
+// individually to makeSwingsetController's
+// runtimeOptions.kernelBundle
+
+// Tests can also pass the whole result to buildVatController's
+// runtimeOptions.kernelBundles, which will pass it through to both.
+
+export async function buildKernelBundles() {
+  const bp = buildVatAndDeviceBundles();
+  const kp = buildKernelBundle();
+  const [vdBundles, kernelBundle] = await Promise.all([bp, kp]);
+  return harden({ kernel: kernelBundle, ...vdBundles });
 }
 
 function byName(a, b) {
@@ -256,7 +272,7 @@ function sortObjectProperties(obj, firsts = []) {
   return result;
 }
 
-/** @typedef {{ kernelBundles?: Record<string, string>, verbose?: boolean,
+/** @typedef {{ kernelBundles?: Record<string, Bundle>, verbose?: boolean,
  *              addVatAdmin?: boolean, addComms?: boolean, addVattp?: boolean,
  *              addTimer?: boolean,
  *            }} InitializationOptions
@@ -318,9 +334,7 @@ export async function initializeSwingset(
   }
 
   const {
-    kernelBundles = await buildKernelBundles({
-      bundleFormat: config.bundleFormat,
-    }),
+    kernelBundles = await buildVatAndDeviceBundles(),
     verbose,
     addVatAdmin = true,
     addComms = true,
@@ -328,7 +342,6 @@ export async function initializeSwingset(
     addTimer = true,
   } = initializationOptions;
 
-  kvStore.set('kernelBundle', JSON.stringify(kernelBundles.kernel));
   kvStore.set('lockdownBundle', JSON.stringify(kernelBundles.lockdown));
   kvStore.set('supervisorBundle', JSON.stringify(kernelBundles.supervisor));
 
--- a/packages/SwingSet/src/kernel/state/kernelKeeper.js
+++ b/packages/SwingSet/src/kernel/state/kernelKeeper.js
@@ -52,7 +52,6 @@ const enableKernelGC = true;
 // device.nextID = $NN
 // meter.nextID = $NN // used to make m$NN
 
-// kernelBundle = JSON(bundle)
 // namedBundleID.$NAME = bundleID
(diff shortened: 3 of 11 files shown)


packages/SwingSet/src/controller/controller.js
```diff
@@ -238,6 +239,10 @@ export async function makeSwingsetController(
   // see https://github.com/Agoric/SES-shim/issues/292 for details
   harden(console);
 
+  writeSlogObject({ type: 'bundle-kernel-start' });
```
mhofman, line 243: Adding slog events should also result in adding a case in `slog-to-otel`. I'm open to suggestions on how to test this (either through types or unit tests).

packages/SwingSet/src/controller/controller.js
```diff
@@ -542,6 +545,7 @@ export async function buildVatController(
     slogCallbacks,
     warehousePolicy,
     slogFile,
+    kernelBundle: kernelBundles?.kernelBundle,
   };
   const initializationOptions = { verbose, kernelBundles };
```
mhofman, line 554: It seems that `kernelBundle` (formerly `kernel`) is not used by `initializeSwingset` ? Should `kernelBundles` be destructured first to split the pieces? And maybe the `initializationOptions` property renamed to be explicit it no longer looks at the `kernel` option?

Aka:
```js
  const {
    hostStorage = provideHostStorage(),
    env,
    verbose,
    kernelBundles: { kernel: kernelBundle, ...vatAndDeviceBundles } = {},
    debugPrefix,
    slogCallbacks,
    warehousePolicy,
    slogFile,
  } = runtimeOptions;
```

packages/SwingSet/src/controller/controller.js
```diff
@@ -542,6 +545,7 @@ export async function buildVatController(
     slogCallbacks,
     warehousePolicy,
     slogFile,
+    kernelBundle: kernelBundles?.kernelBundle,
```
mhofman: Note to self: it previously was `kernelBundles.kernel`, renamed in `buildKernelBundles`

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -29,27 +29,29 @@ const allValues = async obj =>
...
+ * calls this on each launch, to get the
+ * current kernel sources
  */
-export async function buildKernelBundles(options = {}) {
-  // this takes 2.7s on my computer
-
-  const { bundleFormat = undefined } = options;
+export async function buildKernelBundle() {
+  // this takes about 1.0s on my computer
+  const src = rel => bundleSource(new URL(rel, import.meta.url).pathname);
```
mhofman: Is there a reason not to extract this `src` helper out of this function and `buildVatAndDeviceBundles` and into top level?

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -67,6 +69,22 @@ export async function buildKernelBundles(options = {}) {
...
+// runtimeOptions.kernelBundle
+
+// Tests can also pass the whole result to buildVatController's
+// runtimeOptions.kernelBundles, which will pass it through to both.
+
+export async function buildKernelBundles() {
+  const bp = buildVatAndDeviceBundles();
+  const kp = buildKernelBundle();
+  const [vdBundles, kernelBundle] = await Promise.all([bp, kp]);
+  return harden({ kernelBundle, ...vdBundles });
```
mhofman: Any reason to rename the previously used `kernel` to `kernelBundle` ? I'm always warry of such renames with our typing info so sparse.

mhofman (approved):
LGTM, mostly nits on option naming.

I'd like to see the slog-to-otel file updated to avoid the console err spam (arguably minimal here)

packages/SwingSet/src/controller/controller.js
warner: Yeah, until/unless we have slog-to-otel tests (was that sketch I wrote at all useful?), I guess types are good enough. Adding a single span with name `bundle-kernel` or `swingset-bundle-kernel` that gets recorded when the `bundle-kernel-finish` appears might be nice, although really I don't know how helpful it'll be to get telemetry on this, it's always going to be about the same amount of time, and happens rarely.

packages/SwingSet/src/controller/controller.js
warner: Right, `kernelBundle` is no longer used by `initializeSwingset`. It accepts a `kernelBundles` (plural) that has the vat/device/lockdown/supervisor bundles (aka "bundles that the kernel needs") but not `kernelBundle` (aka "the bundle that contains the actual kernel"). But I figured it wasn't worth the work to remove it: `initializeSwingset` will just ignore the spurious property.

This `buildVatController` is a deprecated API that's mostly used by external unit tests, so I didn't want to change the signature significantly. And by having `buildVatController` accept an option object that includes both `kernelBundle` and the other bundles, none of those tests needs to change, and they all continue to benefit from the amortized bundling costs. But the consequence is that the name `kernelBundles` as used in the third argument of `buildVatController` (and also the return value of the exported `buildKernelBundles()` function) is a slightly different/larger type than the "not the kernel too" `kernelBundles` accepted by the third argument of `initializeSwingset`.

I guess your proposal would look like:
```js
  const {
    hostStorage = provideHostStorage(),
    env,
    verbose,
    kernelBundles: { kernelBundle, ...otherBundles } = {},
    debugPrefix,
    slogCallbacks,
    warehousePolicy,
    slogFile,
  } = runtimeOptions;
  const actualRuntimeOptions = {
    env,
    verbose,
    debugPrefix,
    slogCallbacks,
    warehousePolicy,
    slogFile,
    kernelBundle,
  };
  const initializationOptions = { verbose, kernelBundles: otherBundles };
```

right? I'd be ok with that if it reads better to you.

And the wart is that `initializationOptions.kernelBundles` is a different type than `actualRuntimeOptions.kernelBundles`. I could change `makeSwingsetController` to take an option with a different name, but 1: that's a lot of churn and 2: `kernelBundles` is pretty good and 3: `vatAndDeviceAndSupervisorAndLockdownBundles` is pretty bad :).

packages/SwingSet/src/controller/initializeSwingset.js
warner: sure

packages/SwingSet/src/controller/initializeSwingset.js
warner: When the value is passed into `runtimeOptions`, I wanted it to be named `kernelBundle` and not `kernel` (since at that point it's no longer living inside a container with `bundles` in the name). I want to keep using `buildKernelBundles` and `buildVatController(.. initializationOptions.kernelBundles)` names to avoid rewriting a bunch of other tests.

I guess one option would be to have `buildKernelBundles` and `buildVatController` create/accept `.kernel`, and then rename it within `buildVatController` to pass `runtimeOptions.kernelBundle` into `makeSwingsetController`. Lemme see how that looks.

packages/SwingSet/src/controller/controller.js
warner: nope, `kernelBundles: { kernelBundle, ...otherBundles } = {}` means that `otherBundles` is set to an empty object, even if the caller didn't provide `kernelBundles`, which makes `initializeSwingset()` skips the default-value call to `await buildVatAndDeviceBundles()`:

```js
  const {
    kernelBundles = await buildVatAndDeviceBundles(),
    verbose,
    addVatAdmin = true,
    addComms = true,
    addVattp = true,
    addTimer = true,
  } = initializationOptions;
```

so they wind up empty.

I think I'd have to do something like:
```js
  const {
    hostStorage = provideHostStorage(),
    env,
    verbose,
    kernelBundles: kernelAndOtherBundles = {},
    debugPrefix,
    slogCallbacks,
    warehousePolicy,
    slogFile,
  } = runtimeOptions;
  const { kernel: kernelBundle, ...otherBundles } = kernelAndOtherBundles;
  const kernelBundles = runtimeOptions.kernelBundles ? otherBundles : undefined;
```

warner:
@mhofman New version pushed, see what you think about the changes

packages/SwingSet/src/controller/controller.js
mhofman: Oh yeah that'd be a problem.
Starting to be not worth it!

packages/SwingSet/src/controller/controller.js
```diff
@@ -532,20 +532,23 @@ export async function buildVatController(
...
     verbose,
-    kernelBundles,
+    kernelBundles: kernelAndOtherBundles = {},
     debugPrefix,
     slogCallbacks,
     warehousePolicy,
     slogFile,
   } = runtimeOptions;
+  const { kernel: kernelBundle, ...otherBundles } = kernelAndOtherBundles;
+  const kernelBundles = runtimeOptions.kernelBundles ? otherBundles : undefined;
```
mhofman, line 543: Yeah sorry that's unfortunately not as clean as I originally thought!

mhofman (approved):
Thanks for addressing those nits. I think you were right thought, and excluding the `kernelBundle` from the rest may not be worth it.

I'll address the otel stuff in my upcoming PR since we already have spam from the unknown `snapshot` slot events anyway.

warner:
Oh, sorry, I missed the fact that slog-to-otel `errors` everything it doesn't recognize.. I wonder if it'd be better to ignore those events, it's not like the operator can do anything about it, and it's not fatal or damaging.

warner:
Eh, it works now, and I fixed the type problems, so I guess I'll go with the current version.

Issue #4160, opened 2021-12-07
Performance impact of BOYD


mhofman:
## Describe the bug
The bring out your dead PR (https://github.com/Agoric/agoric-sdk/pull/3801) caused a jump in swingset time. There also seem to be more time spent in swingset during a block over the life of the chain, which is usually indicative of a memory leak.

## To Reproduce
1. Follow instructions to setup loadgen
2. Start a loadgen test with the following parameters: `--stage.duration=10 --stage.loadgen.vault.interval=12 --stage.loadgen.amm.interval=12 --stage.loadgen.amm.wait=6 --stage.loadgen.vault.limit=10 --stage.loadgen.amm.limit=10`

## Expected behavior
Not such a significant increase in swingset time. No further increase over the life of the chain

## Platform Environment
 - Loadgen docker image
 - is there anything special/unusual about your platform? WSL2
 - what version of the Agoric-SDK are you using? `9a47516a349185409d3c320b585312178c8a08b8`

## Additional context
The PR doesn't change the GC schedule so the impact is likely from the different accounting logic that BYOD introduces.

## Screenshots
```csv
"sdkRevision","sdkCommitTime","success","walletDeployDuration","loadgenDeployDuration","cycleSuccessRate","cycleAvgDuration","stage1_cycleSuccessRate","stage1_cycleAvgDuration","stage1_avgSwingsetBlockTime","stage1_avgSwingsetPercentage","stage2_cycleSuccessRate","stage2_cycleAvgDuration","stage2_avgSwingsetBlockTime","stage2_avgSwingsetPercentage","stage3_cycleSuccessRate","stage3_cycleAvgDuration","stage3_avgSwingsetBlockTime","stage3_avgSwingsetPercentage","stage4_cycleSuccessRate","stage4_cycleAvgDuration","stage4_avgSwingsetBlockTime","stage4_avgSwingsetPercentage"
"da7c924aa","2021-11-01 23:00:26",true,238.067808,94.07785,100,219.902715,100,214.911984,1.254112,24.89,100,218.745497,1.391519,27.61,100,226.627465,1.402919,27.84,100,220.151802,1.425838,28.14
"9a47516a3","2021-11-03 01:10:09",false,238.950638,98.910524,100,234.284053,100,235.138661,1.713511,34.03,100,228.431566,1.821268,36.14,100,239.761833,2.357044,47.26,,,,
```

[...]

PR #5707, opened 2022-06-30
fix(swingset): remove xs-worker-no-gc and gcEveryCrank

### LINKED ISSUE #5600: remove `gcEveryCrank` and worker type `xs-worker-no-gc`

warner:
While we were debugging performance problems last year, we introduced
`managerType: 'xs-worker-no-gc'`, which is a variant of `xs-worker`
that disables the forced GC we were doing at the end of every
delivery.

Since then, we've improved performance in several ways:

* XS bugs were retaining objects, which have since been fixed
* we only do a forced GC during special `dispatch.bringOutYourDead`
cranks
* we deliver BOYD less often (according to a configurable schedule,
see #4160 for our decisions, but we're thinking once every 100 to 1000
deliveries)

The `xs-worker-no-gc` manager type controlled a flag named
`gcEveryCrank`, which replaced the GC primitive available to liveslots
with a dummy version.

This commit removes both `xs-worker-no-gc` and `gcEveryCrank`.

closes #5600

### DIFF
--- a/packages/SwingSet/docs/configuration.md
+++ b/packages/SwingSet/docs/configuration.md
@@ -104,10 +104,10 @@ bootstrap vat.  If this property is missing, there will be no
 bootstrap vat, which may be disallowed in certain use cases.
 
 The `defaultManagerType` property indicates what sort of vat worker to run vats
-in if they are not specifically configured otherwise.  Currently allowed manager
-types are `'local'`, `'nodeWorker'`, `'node-subprocess'`, `'xs-worker'`, and
-`'xs-worker-no-gc'`.  Of these, only `'local'` and `'xs-worker'` are fully
-supported; the others are experimental.  If omitted, it defaults to `'local'`.
+in if they are not specifically configured otherwise.  Currently allowed
+manager types are `'local'`, `'nodeWorker'`, `'node-subprocess'`, and
+`'xs-worker'`.  Of these, only `'local'` and `'xs-worker'` are fully supported;
+the others are experimental.  If omitted, it defaults to `'local'`.
 
 The `includeDevDependencies` property, if `true`, instructs the bundler which
 creates bundles to include the SwingSet package's dev dependencies as well as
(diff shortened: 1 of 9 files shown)


FUDCo (approved):
Hooray for crap removal!

PR #5710, opened 2022-07-01
fix: shutdown controller after tests

### LINKED ISSUE #5447: Kernel memory leak causing test SIGKILL
## Describe the bug
As of `4279-critical-vat-panic` (https://github.com/Agoric/agoric-sdk/pull/5436), the SwingSet test `test/vat-admin/terminate/test-terminate.js` fails after 27ish tests (out of 31) on an OOM failure, even when run serial, but apparently only when using an XS worker.

## To Reproduce
1. `cd packages/SwingSet`
2. `yarn test:xs -s --verbose 'test/vat-admin/terminate/test-terminate.js'`

## Expected behavior
Successful test

## Platform Environment
 - what OS are you using? what version of Node.js? Node 16 on Debian buster (in a WSL VM)
 - is there anything special/unusual about your platform? Low-ish memory with hard cap on swap (memory = 8GB RAM + 8GB swap)
 - what version of the Agoric-SDK are you using? `c99ef9eb2606eb7f0badaaecf5df710f2f03d897`

## Additional context
https://agoricopco.slack.com/archives/C038E5369FT/p1653534455289559

The process seem to grow to a VMM size of 12GB, skipping 3-4 tests, the suite completes successfully. It runs fine under Node workers but may just be a matter of not growing the overall memory as quickly (less multi process overhead, each with their own copy of the supervisor).

Re-bundling before each test (aka not sharing any explicit state between tests) does not solve the issue. If something holds onto the memory, it's somewhere in the implementation.

## Screenshots
https://github.com/Agoric/agoric-sdk/runs/6606673500

mhofman:
closes: #5447

## Description

After investigating #5447, I realized that the culprit was the xsnap child process holding onto the kernel bundles through an obscure chain of retainers.

### Security Considerations

None

Any new tests creating a controller should add a teardown hook to shutdown the controller.

### Testing Considerations

Ran previously exploding terminate test locally with v8 heap snapshots, and confirmed no leaks were occurring.


turadg (commented):
`run-protocol` changes LGTM

mhofman:
I don't know if this change is related, but it looks like Swingset tests now only take 2/3 minutes!

packages/run-protocol/test/swingsetTests/governance/test-governance.js
```diff
@@ -71,6 +71,7 @@ test.before(async t => {
 async function main(t, argv) {
   const { kernelBundles, config } = t.context.data;
   const controller = buildVatController(config, argv, { kernelBundles });
+  t.teardown(E(controller).shutdown);
```
warner, line 74: why the `E(controller)` ?

packages/run-protocol/test/swingsetTests/vaultFactory/test-vaultFactory.js
```diff
@@ -77,6 +77,7 @@ test.before(async t => {
 async function main(t, argv) {
   const { kernelBundles, config } = t.context.data;
   const controller = buildVatController(config, argv, { kernelBundles });
+  t.teardown(E(controller).shutdown);
```
warner, line 80: same

packages/xsnap/src/xsnap.js
```diff
@@ -114,7 +114,7 @@ export function xsnap(options) {
     stdio: ['ignore', stdout, stderr, 'pipe', 'pipe'],
   });
 
-  xsnapProcess.on('exit', (code, signal) => {
+  xsnapProcess.once('exit', (code, signal) => {
```
warner, line 117: is this just hygiene, or is it possible for processes to `.on('exit')` multiple times?

warner (approved):
Swingset changes look good.. one question about the `E()` in some non-swingset tests though

packages/run-protocol/test/swingsetTests/governance/test-governance.js
mhofman: That test setup used that form and doesn't `await` the result of `buildVatController`, so I kept it intact 🤷‍♂️

packages/xsnap/src/xsnap.js
mhofman: It's not possible for `'exit'` to be triggered multiple times (however the base EventEmitter logic doesn't know that), and hygiene dictates that events should be removed once no longer used.

warner:
Wow, good catch. I remember adding explicit `c.shutdown()`s to the replay tests that explicitly exercise successive launches of the same state, and to tests that deliberately launch xsnap workers, but I didn't think about the environment variables that can cause the entire test suite to use xsnap. I have a vague memory that the lack of `.shutdown` made the parent process hang (because Node is waiting for all sockets to close before allowing the process to exit?). Did it look like all none of the AVA processes were ever exiting or something?

Did the memory being held seem appropriate for a live controller to be holding? I know there's some amount of memory needed just for the kernel to do its thing, and having a zillion of them in a single AVA process could be a problem, but it sounded like you were seeing old messages being retained too? That would sound like a different problem that might still be an issue.

packages/run-protocol/test/swingsetTests/governance/test-governance.js
warner: ah, ok. I happened to ask MarkM about that pattern today, from some unrelated code, and he said it was a wart that `E()` allowed the proxy-trap -returned function to be observed that way (`E.get` would not, and the tildot syntax would probably treat `x = target~.method; x()` quite differently than `x = target~.method()`, so he discouraged me from considering using it.

If `controller` doesn't last more than a few lines, I'd be willing to rewrite the code to do `const controller = await ..`. But yeah, I'm also ok with matching the existing pattern and calling it a day.

mhofman:
xsnap processes were lingering around, and through the pipes from them the whole kernel stayed in memory, including bundles etc. After doing a shutdown, it seems like almost nothing was surviving from one test to another.

The most heap data was the bundles, but there were other/shorter retaining paths to them, which [I had to clean up](https://github.com/endojs/endo/pull/1232) (the stackString logic in SES in particular) before I realized the retaining path was really the xsnap `childProcess`, at which point I wondered why that was actually there, and realized we needed to explicitely shutdown the controllers.

warner:
I bet I believe(d) that AVA runs each test file in a separate process, and once all those test cases have finished, the process exits. I bet I also believe(d) that when the process exits, all the child `xsnap` processes would exit. But 1: I haven't actually tested either belief (I should add some `console.log(getpid())`s and see if they're distinct), and 2: I suspect AVA is doing something clever with threads that winds up running more than one test file in a single process (based on some GC weirdness we keep seeing), and 3: I can easily imagine that turning into a whole lot of dead kernels being kept in RAM until the larger/aggregate process exits.

If AVA were really enthusiastic about threads, maybe every test it runs would contribute a dead kernel and the VM footprint would grow hugely before finally exiting for real. But only in CI where we use the environment variable to switch on XS on all tests.

PR #5712, opened 2022-07-01
fix(wallet-api): dont rely on attestation brand

### LINKED ISSUE #5675: [Wallet API] tryReclaimingWithdrawnPayments hangs on sim-chain
## Describe the bug
When running a sim-chain (`agoric start --reset`), there is no attestation issuer, so `E(agoricNames).lookup('issuer', 'Attestation')` hangs indefinitely. This causes [tryReclaimingWithdrawnPayments](https://github.com/Agoric/agoric-sdk/blob/master/packages/wallet/api/src/lib-wallet.js#L666) to hang. If there is an error in an offer, it looks like the user never gets their `GIVE` back because their withdrawn payments never make it back into their purse.

## To Reproduce
1. Start a sim-chain `agoric start --reset`
2. Start a dapp such as https://github.com/RBFLabs/gateway-agoric-workshop-2022#react-ui
3. Make an offer with a `GIVE` in the proposal that will cause an error (e.g. invalid amount value)
4. Accept the offer in your wallet
5. Observe the `GIVE` withdrawn from your purse, the offer rejects due to an error, and the payment never goes back into your purse.

## Expected behavior
The payment goes back into the user's purse

samsiegart:
fixes: https://github.com/Agoric/agoric-sdk/issues/5675

Testing: Created an offer proposal that would cause an error in Zoe with an `attestation` in the give--observed the wallet create the attestation payment, fail to execute the offer, then unlien the attestation. Screenshot:
image

### DIFF
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -580,8 +580,6 @@ export function makeWallet({
     E(agoricNames)?.lookup('issuer', 'Attestation'),
   );
 
-  const getAttBrand = makeMemoizedGetter(() => E(getAttIssuer()).getBrand());
-
   /** @type import('@endo/promise-kit').PromiseKit<AttestationTool> */
   const attMakerPK = makePromiseKit();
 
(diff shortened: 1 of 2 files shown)


turadg (approved):
Are the changes to `test-lib-wallet.js` necessary?

otherwise LGTM

samsiegart:
> Are the changes to `test-lib-wallet.js` necessary?

Yes the tests fail otherwise. I think messing with the number of async calls changed the order of ids being generated.

> otherwise LGTM

Thanks for review!

PR #5715, opened 2022-07-02
chore: speed up test-upgrade.js by sharing kernel bundles

warner:
Use the same `t.before` pattern we've used in other tests to amortize the cost of `buildKernelBundles`, reduces runtime from 40s to 20s.


warner:
Also remove some log noise

mhofman (approved):

PR #5717, opened 2022-07-04
Return ERefs rather than Promises when possible

erights:
Low priority. Not needed for MN1. But nice cleanups.

Functions that incidentally return a `Promise<T>` should also be able to return a `T`, and thus should be typed `ERef<T>`. Any function that currently type checks against the `Promise<T>` return type should continue to type check. But callers that were counting on it being a promise are encouraged by the type system to be more tolerant. This reserved refactoring possibilities for the implementation. Doing this revealed places where we were directly using `.then` when we should have been using `E.when`.

Unfortunately, TS does not allow an `async` function to be declared as returning an `ERef`. It insists on a `Promise`. Thus, we are forced to type these functions in a way that allows callers to rely on it being a promise, precluding attractive refactorings.

### DIFF
--- a/packages/ERTP/test/unitTests/test-issuerObj.js
+++ b/packages/ERTP/test/unitTests/test-issuerObj.js
@@ -104,7 +104,7 @@ test('issuer.makeEmptyPurse', async t => {
   const performWithdrawal = () => purse.withdraw(fungible837);
 
   const checkWithdrawal = async newPayment => {
-    issuer.getAmountOf(newPayment).then(amount => {
+    E.when(issuer.getAmountOf(newPayment), amount => {
       t.assert(
         AmountMath.isEqual(amount, fungible837),
         `the withdrawn payment has the right balance`,
@@ -255,6 +255,8 @@ test('issuer.burn', async t => {
     AmountMath.isEqual(burntBalance, AmountMath.make(brand, 837n)),
     `entire minted payment was burnt`,
   );
+  // @ts-expect-error ERef arg should be compat to Promise or PromiseLike
+  // param
   await t.throwsAsync(() => issuer.getAmountOf(payment1), {
     message: /was not a live payment for brand/,
   });
@@ -267,7 +269,7 @@ test('issuer.claim', async t => {
   await E(issuer)
     .claim(payment1, AmountMath.make(brand, 2n))
     .then(async newPayment1 => {
-      await issuer.getAmountOf(newPayment1).then(amount => {
+      await E.when(issuer.getAmountOf(newPayment1), amount => {
         t.assert(
           AmountMath.isEqual(amount, AmountMath.make(brand, 2n)),
           `new payment has equal balance to old payment`,
@@ -279,6 +281,8 @@ test('issuer.claim', async t => {
         );
       });
 
+      // @ts-expect-error ERef arg should be compat to Promise or PromiseLike
+      // param
       return t.throwsAsync(() => issuer.getAmountOf(payment1), {
         message: /was not a live payment for brand/,
       });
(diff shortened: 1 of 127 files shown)


packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -104,7 +104,7 @@ test('issuer.makeEmptyPurse', async t => {
   const performWithdrawal = () => purse.withdraw(fungible837);
 
   const checkWithdrawal = async newPayment => {
-    issuer.getAmountOf(newPayment).then(amount => {
+    E.when(issuer.getAmountOf(newPayment), amount => {
```
mhofman, line 107: There was apparently a double error here? Shouldn't the result of `E.when()` be `await`ed?

packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -255,6 +255,8 @@ test('issuer.burn', async t => {
     AmountMath.isEqual(burntBalance, AmountMath.make(brand, 837n)),
     `entire minted payment was burnt`,
   );
+  // @ts-expect-error ERef arg should be compat to Promise or PromiseLike
```
mhofman, line 258: So this actually brings up an issue with `ERef` and the expectation that the caller should be able to handle a synchronous result. Most code will not be able to handle a synchronous exception, and I'd argue it shouldn't.

After going back and forth on what I'd expect to happen, the ava implementation in this case does what I think is right, which is to fail the test if the function throws synchronously. It also fails if the function returns a value which is not a promise. https://github.com/avajs/ava/blob/v3.15.0/lib/assert.js#L588-L606

So I'd argue that technically `ERef` is not compatible with `PromiseLike` in general, but that in this case it is, because only a rejecting `PromiseLike` result will pass the test.

packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -411,7 +421,7 @@ test('issuer.combine array of promises', async t => {
   harden(paymentsP);
 
   const checkCombinedResult = paymentP => {
-    issuer.getAmountOf(paymentP).then(pAmount => {
+    E.when(issuer.getAmountOf(paymentP), pAmount => {
```
mhofman, line 424: Should this be `await`ed as well?

packages/notifier/src/storesub.js
```diff
@@ -60,10 +60,17 @@ export const makeStoredSubscription = (
 
   if (storageNode) {
     // Start publishing the source.
-    observeIteration(subscription, {
-      updateState: publishValue,
-      finish: publishValue,
-    }).catch(fail);
+    E.when(
```
mhofman, line 63: Handle promise ?

mhofman (commented):
The replacements from `.then()` to `E.when()` look fine, but do seem to highlight some previously missing unhandled promises. I'm not sure why they weren't warned about before.

I am a little worried about changing `Promise<void>` to `ERef<void>` since the `void` type has a special meaning of no return type, and is thus contagious. In practice an `ERef<void>` is equivalent to `void` (with the Promise part of the union being dropped), which would theoretically make `await`ing the result appear unnecessary to the type system. See https://github.com/microsoft/TypeScript/issues/42709.

Similarly `ERef<any>` will result in `any` since it's contagious too, but at least TS will not complain when `await`ing `any`.

Finally, my biggest concern is with error handling for `ERef` returning functions. These functions would imply that they may throw synchronously, which would make code like `E.when(foo(), handleFooResult)` problematic.

To summarize, I'm fairly uncomfortable with switching to `ERef` for return types in such wholesale fashion.

erights:
Closing. We may eventually scavenge some useful code from the carcass of this PR.

Issue #4548, opened 2022-02-14
Bootstrap ocaps are stored durably in vat-bootstrap


Tartuffo:
### Problem to be solved

Bootstrap is not upgradable yet holds precious state in closure.

## Description of design

Either make the bootstrap vat upgradable or move its (non-copyable) precious state into upgradable vats.

We're opting for the latter so that if necessary we can make a new bootstrap vat and have it retrieve everything precious from the space. (This requires that everything precious is created in some other vat than the original bootstrap).

```[tasklist]
### Tasks
- [x] make an agoricNames vat
- [x] move the per-user `namehub` into provisionPool vat
- [x] move `namesByAddress` into agoricNames vat
- [x] move USD oracleBrand to mintHolder
- [x] audit bootstrap exports
- [x] use durable zone
- [x] unify vat stores
```

Goals (optional)
 - [ ] make bridge manager facets into a space?
 - [ ] make vats into a space (likewise bundles?)
 - [ ] move client/home stuff to new file, out of chain-behaviors?

### Testing plan

~demonstrate upgrading the bootstrap vat~ don't need to because with the precious state secured elsewhere the bootstrap vat can be blown away

If times allows, write a test that creates a new bootstrap and core-eval can provide it with everything it needs from the space.

Tartuffo:
Per @dckc: Uncertainty in estimate: (1) I conservatively assume this is the first time I'm doing virtual / durable conversion. (2) bootstrap is kinda hairy and touches lots of stuff. (3) @dtribble  talked about a significant redesign in bootstrap when we make it durable.

Tartuffo:
Don't need to impl this for Restival.  For Restival, need it to be clear how the world boots.  Starting RUN Protocol is separate from bootstrap.   How can we add things to bootstrap after launch, how do we fix them?  We don't want accumulated non-durable stuff that will complicate this.

dckc:
Also audit for objects that are allowed to go out of scope.

  - admin facets
    - [x] design idea: capture the whole return from `startInstance`, not just the `creatorFacet`
  - runStakeGovernor instance

dckc:
@dtribble and I agreed to postpone this:

He had proposed that it stored all of it state in a baggage and then go away each time so that each invcation of the big Javasript hammer starts only from baggage. This would ensure that any limitations in our current bootstrap are irrelevant because we could load a different contractor bootstrap.

[...]

PR #5721, opened 2022-07-05
4484 multitenant smart wallet

turadg:
refs: #4484

## Description

Part of #4484. This doesn't do everything described in that ticket so I'm leaving it to the ticket author to decide what's left, if anything.

### Security Considerations

Shared contract amongst all on-chain wallets. Potential DoS attack.

--

### Testing Considerations

New tests

### DIFF
--- a/packages/store/src/stores/store-utils.js
+++ b/packages/store/src/stores/store-utils.js
@@ -106,3 +106,69 @@ export const provide = (mapStore, key, makeValue) => {
   return mapStore.get(key);
 };
 harden(provide);
+
+/**
+ * Helper for use cases in which the maker function is async. For two provide
+ * calls with the same key, one may be making when the other call starts and it
+ * would make again. (Then there'd be a collision when the second tries to store
+ * the key.) This prevents that race condition by immediately storing a Promise
+ * for the maker in an ephemeral store.
+ *
+ * Upon termination, the ephemeral store of pending makes will be lost. It's
+ * possible for termination to happen after the make completes and before it
+ * reaches durable storage.
+ *
+ * @template K
+ * @template V
+ * @param {MapStore<K, V>} durableStore
+ */
+export const makeAtomicProvider = durableStore => {
+  /** @type {Map<K, Promise<V>>} */
+  const pending = new Map();
+
+  /**
+   * Call `provideAsync` to get or make the value associated with the key,
+   * when the maker is asynchronous.
+   * If there already is one, return that. Otherwise,
+   * call `makeValue(key)`, remember it as the value for
+   * that key, and return it.
+   *
+   * @param {K} key
+   * @param {(key: K) => Promise<V>} makeValue
+   * @param {(key: K, value: V) => Promise<void>} [finishValue]
+   * @returns {Promise<V>}
+   */
+  const provideAsync = (key, makeValue, finishValue) => {
+    if (durableStore.has(key)) {
+      return Promise.resolve(durableStore.get(key));
+    }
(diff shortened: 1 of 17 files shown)


packages/store/src/stores/store-utils.js
```diff
@@ -106,3 +106,24 @@ export const provide = (mapStore, key, makeValue) => {
...
+ *
+ * @template K,V
+ * @param {MapStore<K,V>} mapStore
+ * @param {K} key
+ * @param {(key: K) => Promise<V>} makeValue
+ * @returns {Promise<V>}
+ */
+export const provideAsync = async (mapStore, key, makeValue) => {
+  if (!mapStore.has(key)) {
+    mapStore.init(key, await makeValue(key));
```
dckc: this goes against [No Nested Await](https://github.com/Agoric/agoric-sdk/wiki/No-Nested-Await). Is there any particular justification?

packages/vats/src/core/basic-behaviors.js
```diff
@@ -236,22 +236,29 @@ export const makeClientBanks = async ({
...
   installation: {
-    consume: { singleWallet },
+    consume: { walletFactory },
   },
 }) => {
   const STORAGE_PATH = 'wallet';
 
   const storageNode = await getChildNode(chainStorage, STORAGE_PATH);
   const marshaller = E(board).getPublishingMarshaller();
+  const { creatorFacet } = await E(zoe).startInstance(
```
dckc, line 246: FYI, dropping the admin facet here is not great; but we have plans elsewhere to address it: https://github.com/Agoric/agoric-sdk/issues/4548#issuecomment-1111360304

packages/wallet/contract/test/test-walletFactory.js
```diff
@@ -0,0 +1,203 @@
...
+  };
+};
+harden(setUpZoeForTest);
+
+export const mockChainStorageRoot = () => {
+  const toStorage = v => v;
+  return makeChainStorageRoot(toStorage, 'swingset', 'mockChainStorageRoot');
+};
+
+const mockAddress1 = 'mockAddress1';
```
dckc, line 60: how about testing 2 or 3 addresses / wallets?

dckc (approved):
Looks good.

It would look better with tests for multiple addresses; I leave that up to you.

packages/store/src/stores/store-utils.js
turadg: Nope. Refactored with [6001a6d](https://github.com/Agoric/agoric-sdk/pull/5721/commits/6001a6d5449a2af7095ab56c82b5b86ec8885b70). 

If you're not confident in that, it'll get cleaned up in https://github.com/Agoric/agoric-sdk/pull/5728

dckc: The introduction of the `pending` Map doesn't seem equivalent to this. If it's essential, that seems to mean the nested await form was buggy. Anyway... it looks workable, and I'm content that #5728 is where we deal with it definitively.

dckc (approved):
> Part of #4484. This doesn't do everything described in that ticket ...

really? ah. indeed... virtualization / durability.

packages/store/src/stores/store-utils.js
turadg: > The introduction of the pending Map doesn't seem equivalent to this

Right, I wasn't clear. The nested await went away as a side effect of the refactoring to `makeAtomicProvider`. The need for that function was because of a race condition I encountered when testing multiple wallets. Thanks for suggesting it!

PR #5726, opened 2022-07-06
fix(swingset): don't double-allocate durable object vrefs

### LINKED ISSUE #5725: virutal Kinds create duplicate vrefs after upgrade
@Chris-Hibbert experienced object-identity problems while working on contract upgrades. Both versions do `makeFoo = defineDurableKind(handle)`. The first version does `foo1 = makeFoo()` and stores it as a key in a durable Map (`store.init(foo1, value)`). After upgrade, the second version does `foo2 = makeFoo()` and attempts to do the same thing, but `store.init(foo2, othervalue)` fails with a "key already exists" error.

It looks like virtual kinds are not restoring their "next object ID" counter properly after upgrade, or maybe they aren't persisting the counter in the first version. So the first instance created in the new version is given the same vref as the first instance created in the old version, causing a collision.

I've got a patch to `test-upgrade.js` that exercises the bug (it compares the krefs of two specific exported durable objects, each the first one created in their respective versions).

warner:
The `nextInstanceID` for durable objects, used to allocate vrefs for
new instances, was held ephemerally in RAM. As a result, the first
call to `makeFoo()` (for `makeFoo = defineDurableKind(fooHandle..)`)
in version-2 would get the same vref as the first `makeFoo()` from
version-1, when these should be entirely distinct objects. In
practice, the new object tended to displace the old one.

This changes the virtual object manager to track `nextInstanceID`
durably, by writing it into the `vom.dkind.${kindID}` metadata. This
metadata now as `{ kindID, tag, nextInstanceID }`. We write this out
after every durable object allocation (previously, the metadata was
only written once, during `makeKindHandle`). Non-durable (merely
virtual) objects do not include this counter in the metadata, nor do
they write it to DB more frequently than before.

The upgrade test was modified to look for this overlapping vref
problem.

closes #5725

### DIFF
--- a/packages/SwingSet/src/liveslots/vatstore-usage.md
+++ b/packages/SwingSet/src/liveslots/vatstore-usage.md
@@ -98,6 +98,7 @@ The KindHandle is a durable virtual object of a special internal Kind. This is t
 
 For each virtual object kind that is defined we store a metadata record for purposes of scanning directly through the defined kinds when a vat is stopped or upgraded.  For durable kinds this record is stored in `vom.dkind.${kindID}`; for non-durable kinds it is stored in `vom.vkind.${kindID}`.  Currently this metadata takes the form of a JSON-serialized record `{ kindID, tag }`, where the `kindID` property is the kind ID (redundantly) and `tag` is the tag string as provided in the `defineKind` or `makeKindHandle` call.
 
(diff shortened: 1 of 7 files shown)


warner:
cc @Chris-Hibbert @erights

FUDCo (approved):
Well OK then

PR #5728, opened 2022-07-06
use atomic store provider in AMM

turadg:
## Description

For https://github.com/Agoric/agoric-sdk/pull/5721 I had to make an atomic provider helper to avoid a race condition. @dtribble pointed out this is similar to what https://github.com/Agoric/agoric-sdk/pull/5596 did.

#5721 landed. This PR refactors the AMM's `makeAddIssuer` to use the new `makeAtomicProvider` helper. It improves the helper's types, docs and tests.

### Security Considerations

--

Post in chat about new helper

### Testing Considerations

Expanded test coverage on helper, to have confidence in `finish` option.

### DIFF
--- a/packages/run-protocol/src/vpool-xyk-amm/addPool.js
+++ b/packages/run-protocol/src/vpool-xyk-amm/addPool.js
@@ -3,7 +3,6 @@
 import { E } from '@endo/eventual-send';
 import { AmountMath, AssetKind } from '@agoric/ertp';
 import { assertProposalShape } from '@agoric/zoe/src/contractSupport/index.js';
-import { makePromiseKit } from '@endo/promise-kit';
 
 import { definePoolKind } from './pool.js';
 
@@ -14,15 +13,13 @@ const DISPLAY_INFO = harden({ decimalPlaces: 6 });
 /**
  * @param {ZCF} zcf
  * @param {(Brand) => boolean} isInSecondaries
- * @param {WeakStore<Brand,ZCFMint>} brandToLiquidityMint
- * @param {WeakStore<Brand,ERef<Issuer>>} brandToLiquidityIssuer
+ * @param {import('@agoric/store/src/stores/store-utils.js').AtomicProvider<Brand, ZCFMint>} brandToLiquidityMintProvider
  * @param {() => (secondaryBrand: Brand) => Promise<void>} getAddIssuerToReserve
  */
 export const makeAddIssuer = (
   zcf,
   isInSecondaries,
-  brandToLiquidityMint,
-  brandToLiquidityIssuer,
+  brandToLiquidityMintProvider,
   getAddIssuerToReserve,
 ) => {
   /**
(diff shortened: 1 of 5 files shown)


packages/store/src/stores/store-utils.js
```diff
@@ -106,3 +106,69 @@ export const provide = (mapStore, key, makeValue) => {
...
+ * possible for termination to happen after the make completes and before it
+ * reaches durable storage.
+ *
+ * @template K
+ * @template V
+ * @param {MapStore<K, V>} durableStore
+ */
+export const makeAtomicProvider = durableStore => {
+  /** @type {Map<K, Promise<V>>} */
+  const pending = new Map();
```
turadg, line 127: should this be a WeakMap? a WeakStore?

packages/store/src/stores/store-utils.js
```diff
@@ -106,3 +106,69 @@ export const provide = (mapStore, key, makeValue) => {
...
+   * If there already is one, return that. Otherwise,
+   * call `makeValue(key)`, remember it as the value for
+   * that key, and return it.
+   *
+   * @param {K} key
+   * @param {(key: K) => Promise<V>} makeValue
+   * @param {(key: K, value: V) => Promise<void>} [finishValue]
+   * @returns {Promise<V>}
+   */
+  const provideAsync = (key, makeValue, finishValue) => {
```
Chris-Hibbert, line 141: finishValue isn't described in the summary above, and isn't tested.   I'd like to see both.

packages/store/src/stores/store-utils.js
Chris-Hibbert: If the provider crashes, the store will go away. If the provider doesn't crash, each key will be removed as it resolves.  I don't think there's a need for a weak collection.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -44,65 +41,54 @@ export const makeAddIssuer = (
...
+      // defer lookup until necessary. more aligned with governed
+      // param we expect this to be eventually.
+      const addIssuerToReserve = getAddIssuerToReserve();
+
+      // tell the reserve about this brand, which it will validate by
+      // calling back to AMM for the issuer
+      return addIssuerToReserve(brand);
+    };
+
+    const liquidityMint = await brandToLiquidityMintProvider.provideAsync(
```
Chris-Hibbert: Be consistent about `// AWAIT /////////` markers within a function! There's one on the `await` above, so there must be one here as well.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
```diff
@@ -44,65 +41,54 @@ export const makeAddIssuer = (
...
+          return mint;
+        })
+        .catch(e => {
+          console.error(
+            X`Failure Saving issuer ${secondaryIssuer}. Not added to Reserve`,
+          );
+          throw e;
+        });
+    };
+    /** @type {(brand: Brand) => Promise<void>} */
```
Chris-Hibbert, line 73: I'd like to see a blank line between `makeLiquidityMint` and `finish`.  The break was hard for me to parse visually.

Chris-Hibbert (commented):
I think the change is clarifying. Thanks.

I'll approve once tests are added.

packages/run-protocol/src/vpool-xyk-amm/addPool.js
turadg: I can't find any documentation about when and when not to use `// AWAIT`. Does it just mean that there's an `await` token on the next line? If so, that seems like a code editing concern, not source code.

If we are going to be consistent, shouldn't we be consistent across the repo?

I'll make the change once I can read documentation on when it's supposed to be used. I'll make the documentation if necessary.

Chris-Hibbert: > I can't find any documentation about when and when not to use // AWAIT. Does it just mean that there's an await token on the next line? If so, that seems like a code editing concern, not source code.

Usually on the previous line.    We use it to raise awareness in functions with delicate timing/interaction. That's why I'm asking for consistency within a function.

> If we are going to be consistent, shouldn't we be consistent across the repo?

We've found it to be distracting when overused.

> I'll make the change once I can read documentation on when it's supposed to be used. I'll make the documentation if necessary.

I'd talk with @dtribble

turadg: For now I removed the `await`.

packages/run-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -166,15 +166,13 @@ const start = async (zcf, privateArgs) => {
   const isSecondary = secondaryBrandToPool.has;
 
   // The liquidityBrand has to exist to allow the addPool Offer to specify want
-  /** @type {WeakStore<Brand,ZCFMint<'nat'>>} */
+  /** @type {WeakMapStore<Brand,ZCFMint<'nat'>>} */
```
turadg, line 169: `WeakStore` type is deprecated, says to use `WeakMapStore`

Chris-Hibbert (approved):

PR #5730, opened 2022-07-07
rewrite delivery-result processing, rewind failed upgrades

### LINKED ISSUE #5344: failed vat upgrades should be rewound

Currently, a successful vat upgrade works pretty well.

Upgrades fail if the new version's `buildRootObject` throws an exception (or returns a Promise that rejects or fails to resolve by the end of the crank), or it fails to `defineDurableKind()` for all Kinds inherited from its predecessor.

Currently, a failed upgrade is not handled correctly. What's supposed to happen is that the upgrade gets rewound, and we're left with the original version humming along, all its data still in place, ready to receive more messages. The caller of `adminFacet~.upgrade()` gets a rejected result promise, but all other changes are unwound.

Instead, I think we currently leave the vat in a halfway state, and we don't notify the caller correctly.

## Description of the Design

The kernel's `processUpgradeVat()` function performs two deliveries: `stopVat` and `startVat`. We need to do `abortCrank` if either one fails. And we need to submit the failure message to vat-admin even though `abortCrank` has rewound the state changes. This should look a lot like when a vat is terminated for a bad syscall or meter underrun. We use a little side table to remember the state changes that need to be made after `abortCrank` unwinds everything.

## Security Considerations

I don't know how confused the kernel can currently get in the face of a failed upgrade, but clearly we must make it impossible to corrupt anything kernel-side by intentionally upgrading to a broken version.

## Test Plan

Unit tests.

warner:
This PR modifies vat-upgrade to respond correctly to failures during
the upgrade process, by rewinding the vat back to it's pre-upgrade
state.

From the outside world, it will appear as if upgrade was never
requested, except that `upgradeVat()` rejects.

It also deducts the computrons used during `dispatch.startVat`
(including the userspace `buildRootObject()` execution) from both the
vat's Meter and the host application's `runPolicy`. Metered vats
should plan to provide an additional (estimated) 8M computrons to pay
for the work done during startVat. For block-based host applications,
the blocks that include new vat creation should now be smaller and
closer to the policy-guided limits.

closes #5344

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -251,7 +251,7 @@ export default function buildKernel(
    * @param {boolean} shouldReject
    * @param {SwingSetCapData} info
    */
-  function terminateVat(vatID, shouldReject, info) {
+  async function terminateVat(vatID, shouldReject, info) {
     const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
     const critical = vatKeeper.getOptions().critical;
     insistCapData(info);
@@ -264,7 +264,7 @@ export default function buildKernel(
       // TODO: if a static vat terminates, panic the kernel?
 
       // ISSUE: terminate stuff in its own crank like creation?
-      void vatWarehouse.vatWasTerminated(vatID);
+      await vatWarehouse.vatWasTerminated(vatID);
     }
     if (critical) {
       // The following error construction is a bit awkward, but (1) it saves us
@@ -1119,7 +1119,7 @@ export default function buildKernel(
 
       // state changes reflecting the termination must also survive, so these
       // happen after a possible abortCrank()
-      terminateVat(vatID, reject, info);
+      await terminateVat(vatID, reject, info);
       kernelSlog.terminateVat(vatID, reject, info);
       kdebug(`vat terminated: ${JSON.stringify(info)}`);
     }
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -118,6 +118,7 @@ export default function buildKernel(
 
   const kernelKeeper = makeKernelKeeper(hostStorage, kernelSlog, createSHA256);
 
+  let vatWarehouse;
   let started = false;
 
   /**
@@ -263,7 +264,6 @@ export default function buildKernel(
       // TODO: if a static vat terminates, panic the kernel?
 
       // ISSUE: terminate stuff in its own crank like creation?
-      // eslint-disable-next-line no-use-before-define
       void vatWarehouse.vatWasTerminated(vatID);
     }
     if (critical) {
@@ -374,13 +374,11 @@ export default function buildKernel(
    * @returns {Promise<DeliveryStatus>}
    */
   async function deliverAndLogToVat(vatID, kd, vd) {
-    // eslint-disable-next-line no-use-before-define
     assert(vatWarehouse.lookup(vatID));
     // Ensure that the vatSlogger is available before clist translation.
     const vs = kernelSlog.provideVatSlogger(vatID).vatSlog;
     try {
       /** @type { VatDeliveryResult } */
-      // eslint-disable-next-line no-use-before-define
       const deliveryResult = await vatWarehouse.deliverToVat(vatID, kd, vd, vs);
 
       insistVatDeliveryResult(deliveryResult);
@@ -411,7 +409,6 @@ export default function buildKernel(
     kernelKeeper.incStat('dispatches');
     kernelKeeper.incStat('dispatchDeliver');
 
-    // eslint-disable-next-line no-use-before-define
     const vatInfo = vatWarehouse.lookup(vatID);
     if (!vatInfo) {
       // splat
@@ -423,7 +420,6 @@ export default function buildKernel(
 
     /** @type { KernelDeliveryMessage } */
     const kd = harden(['message', target, msg]);
-    // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
 
     if (vatInfo.enablePipelining && msg.result) {
@@ -443,7 +439,6 @@ export default function buildKernel(
     insistVatID(vatID);
     insistKernelType('promise', kpid);
     kernelKeeper.incStat('dispatches');
-    // eslint-disable-next-line no-use-before-define
     if (!vatWarehouse.lookup(vatID)) {
       kdebug(`dropping notify of ${kpid} to ${vatID} because vat is dead`);
       return null;
@@ -473,7 +468,6 @@ export default function buildKernel(
     }
     /** @type { KernelDeliveryNotify } */
     const kd = harden(['notify', resolutions]);
-    // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
     vatKeeper.deleteCListEntriesForKernelSlots(targets);
 
@@ -490,7 +484,6 @@ export default function buildKernel(
     const { type, vatID, krefs } = message;
     // console.log(`-- processGCMessage(${vatID} ${type} ${krefs.join(',')})`);
     insistVatID(vatID);
-    // eslint-disable-next-line no-use-before-define
     if (!vatWarehouse.lookup(vatID)) {
       return null; // can't collect from the dead
     }
@@ -504,7 +497,6 @@ export default function buildKernel(
         // console.log(`   deleted ${kref}`);
       }
     }
-    // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
     return deliverAndLogToVat(vatID, kd, vd);
   }
@@ -518,13 +510,11 @@ export default function buildKernel(
     const { type, vatID } = message;
     // console.log(`-- processBringOutYourDead(${vatID})`);
     insistVatID(vatID);
-    // eslint-disable-next-line no-use-before-define
     if (!vatWarehouse.lookup(vatID)) {
       return null; // can't collect from the dead
     }
     /** @type { KernelDeliveryBringOutYourDead } */
     const kd = harden([type]);
-    // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
     return deliverAndLogToVat(vatID, kd, vd);
   }
@@ -550,11 +540,9 @@ export default function buildKernel(
     // console.log(`-- processStartVat(${vatID})`);
     insistVatID(vatID);
     insistCapData(vatParameters);
-    // eslint-disable-next-line no-use-before-define
     assert(vatWarehouse.lookup(vatID));
     /** @type { KernelDeliveryStartVat } */
     const kd = harden(['startVat', vatParameters]);
-    // eslint-disable-next-line no-use-before-define
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
     // decref slots now that create-vat is off run-queue
     for (const kref of vatParameters.slots) {
@@ -616,7 +604,6 @@ export default function buildKernel(
     // TODO warner think through failure paths
 
     return (
-      // eslint-disable-next-line no-use-before-define
       vatWarehouse
         .createDynamicVat(vatID)
         // if createDynamicVat fails, go directly to makeErrorResponse
@@ -659,7 +646,6 @@ export default function buildKernel(
   async function processChangeVatOptions(message) {
     const { vatID, options } = message;
     insistVatID(vatID);
-    // eslint-disable-next-line no-use-before-define
     if (!vatWarehouse.lookup(vatID)) {
       return null; // vat is dead, splat
     }
@@ -676,7 +662,6 @@ export default function buildKernel(
     if (haveOptionsForVat) {
       /** @type { KernelDeliveryChangeVatOptions } */
       const kd = harden(['changeVatOptions', optionsForVat]);
-      // eslint-disable-next-line no-use-before-define
       const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
       const status = await deliverAndLogToVat(vatID, kd, vd);
       return { ...status, discardFailedDelivery: true };
@@ -695,14 +680,12 @@ export default function buildKernel(
     const { vatID, upgradeID, bundleID, vatParameters } = message;
     insistCapData(vatParameters);
 
-    // eslint-disable-next-line no-use-before-define
     if (!vatWarehouse.lookup(vatID)) {
       return null; // vat is dead, splat
     }
     const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
     /** @type { import('../types-external.js').KernelDeliveryStopVat } */
     const kd1 = harden(['stopVat']);
-    // eslint-disable-next-line no-use-before-define
     const vd1 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd1);
     const status1 = await deliverAndLogToVat(vatID, kd1, vd1);
     if (status1.terminate) {
@@ -712,7 +695,6 @@ export default function buildKernel(
     }
 
     // stop the worker, delete the transcript and any snapshot
-    // eslint-disable-next-line no-use-before-define
     await vatWarehouse.destroyWorker(vatID);
     const source = { bundleID };
     const { options } = vatKeeper.getSourceAndOptions();
@@ -725,7 +707,6 @@ export default function buildKernel(
     // deliver a startVat with the new vatParameters
     /** @type { import('../types-external.js').KernelDeliveryStartVat } */
     const kd2 = harden(['startVat', vatParameters]);
-    // eslint-disable-next-line no-use-before-define
     const vd2 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd2);
     // decref vatParameters now that translation did incref
     for (const kref of vatParameters.slots) {
@@ -865,7 +846,6 @@ export default function buildKernel(
           return requeue();
         } else {
           insistVatID(kp.decider);
-          // eslint-disable-next-line no-use-before-define
           const deciderVat = vatWarehouse.lookup(kp.decider);
           if (!deciderVat) {
             // decider is dead
@@ -1147,7 +1127,6 @@ export default function buildKernel(
     }
 
     if (!didAbort) {
-      // eslint-disable-next-line no-use-before-define
       await vatWarehouse.maybeSaveSnapshot();
     }
     kernelKeeper.processRefcounts();
@@ -1272,7 +1251,6 @@ export default function buildKernel(
      * @returns {VatSyscallResult}
      */
     function vatSyscallHandler(vatSyscallObject) {
-      // eslint-disable-next-line no-use-before-define
       if (!vatWarehouse.lookup(vatID)) {
         // This is a safety check -- this case should never happen unless the
         // vatManager is somehow confused.
@@ -1354,11 +1332,7 @@ export default function buildKernel(
     overrideVatManagerOptions,
   });
 
-  const vatWarehouse = makeVatWarehouse(
-    kernelKeeper,
-    vatLoader,
-    warehousePolicy,
-  );
+  vatWarehouse = makeVatWarehouse(kernelKeeper, vatLoader, warehousePolicy);
 
   /**
    * Create a dynamically generated vat for testing purposes.  Such vats are
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -648,6 +648,7 @@ export default function buildKernel(
       kdebug(`vat ${vatID} terminated before startVat delivered`);
       return NO_DELIVERY_CRANK_RESULTS;
     }
+    const { meterID } = vatInfo;
     /** @type { KernelDeliveryStartVat } */
     const kd = harden(['startVat', vatParameters]);
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
@@ -660,7 +661,7 @@ export default function buildKernel(
     // note: if deliveryCrankResults() learns to suspend vats,
     // startVat errors should still terminate them
     const results = harden({
-      ...deliveryCrankResults(vatID, status, true),
+      ...deliveryCrankResults(vatID, status, true, meterID),
       popDelivery: true,
     });
     return results;
@@ -1168,6 +1169,12 @@ export default function buildKernel(
     /** @type { PolicyInput } */
     let policyInput = ['crank', {}];
     if (message.type === 'create-vat') {
+      // TODO: create-vat now gets metering, at least for the
+      // dispatch.startVat . We should probably tell the policy about
+      // the creation too since there's extra overhead (we're
+      // launching a new child process, at least, although that
+      // sometimes happens randomly because of vat eviction policy
+      // which should not affect the in-consensus policyInput)
       policyInput = ['create-vat', {}];
     }
 
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1042,10 +1042,8 @@ export default function buildKernel(
         // and if both vat and delivery are metered, deduct from the Meter
         if (useMeter && meterID) {
           deduction = { meterID, computrons }; // in case we must rededuct
-          const { notify, underflow } = kernelKeeper.deductMeter(
-            meterID,
-            computrons,
-          );
+          const underflow = !kernelKeeper.checkMeter(meterID, computrons);
+          const notify = kernelKeeper.deductMeter(meterID, computrons);
           if (notify) {
             notifyMeterThreshold(meterID);
           }
@@ -1099,7 +1097,7 @@ export default function buildKernel(
         // but metering deductions and underflow notifications must survive
         if (deduction) {
           const { meterID, computrons } = deduction; // re-deduct metering
-          const { notify } = kernelKeeper.deductMeter(meterID, computrons);
+          const notify = kernelKeeper.deductMeter(meterID, computrons);
           if (notify) {
             notifyMeterThreshold(meterID); // re-queue notification
           }
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -368,6 +368,7 @@ export default function buildKernel(
    * @typedef { {
    *    abort?: boolean, // changes should be discarded, not committed
    *    popDelivery?: boolean, // discard the aborted delivery
+   *    didDelivery?: boolean, // we made a delivery to a vat, for run policy
    *    computrons?: BigInt, // computron count for run policy
    *    meterID?: string, // deduct those computrons from a meter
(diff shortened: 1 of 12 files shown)


warner:
best reviewed one commit at a time, the meat is in the penultimate commit

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -576,51 +687,53 @@ export default function buildKernel(
     vatKeeper.setSourceAndOptions(source, options);
     vatKeeper.initializeReapCountdown(options.reapInterval);
 
-    function sendNewVatCallback(results, slots) {
-      const methargs = {
-        body: JSON.stringify(['newVatCallback', [vatID, results]]),
+    function makeVAMessage(arg, slots) {
```
FUDCo: This function could do with a better name.

FUDCo (approved):
This looks good, albeit perhaps slightly too clever.

One minor stylistic nit, tweak or not according to your judgement.

packages/SwingSet/src/kernel/kernel.js
warner: yeah, I was trying to keep it short enough to stop prettier from going newline-mad, I'll look for a better one

warner: changed to `makeVatAdminMessage`

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -118,6 +118,7 @@ export default function buildKernel(
 
   const kernelKeeper = makeKernelKeeper(hostStorage, kernelSlog, createSHA256);
 
+  let vatWarehouse;
```
mhofman, line 122:
```suggestion
  /** @type {ReturnType<makeVatWarehouse>} */
  let vatWarehouse;
```

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1354,11 +1332,7 @@ export default function buildKernel(
     overrideVatManagerOptions,
   });
 
-  const vatWarehouse = makeVatWarehouse(
-    kernelKeeper,
-    vatLoader,
-    warehousePolicy,
-  );
+  vatWarehouse = makeVatWarehouse(kernelKeeper, vatLoader, warehousePolicy);
```
mhofman, line 1482: This is one of those eslint-rule that really frustrate me. In this case we lose the inferred typings. Thankfully it's easy enough to restore them.

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -1019,29 +1019,41 @@ export default function makeKernelKeeper(
...
     let oldRemaining = getRequired(`${meterID}.remaining`);
-    if (oldRemaining !== 'unlimited') {
+    if (oldRemaining === 'unlimited') {
+      return false;
+    } else {
       oldRemaining = BigInt(oldRemaining);
       const threshold = BigInt(getRequired(`${meterID}.threshold`));
       let remaining = oldRemaining - spent;
       if (remaining < 0n) {
-        underflow = true;
```
mhofman, line 1029: Can we add a comment that even with the separate check, the kernel will still deduct from the meter even if it will underflow?

packages/SwingSet/src/kernel/vat-warehouse.js
```diff
@@ -180,24 +180,29 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
...
    *  | undefined // if the vat is dead or never initialized
    * }
    */
   function lookup(vatID) {
     const liveInfo = ephemeral.vats.get(vatID);
     if (liveInfo) {
-      const { enablePipelining } = liveInfo;
-      return { enablePipelining };
+      const { enablePipelining, options } = liveInfo;
+      const { meterID } = options;
```
mhofman, line 197: Nit:
```suggestion
      const {
        enablePipelining,
        options: { meterID },
      } = liveInfo;
```

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -314,27 +319,24 @@ export default function buildKernel(
   // `vatFatalSyscall`. We'd still need a way for `requestTermination` to
   // work, though.
 
-  let terminationTrigger;
+  // These two flags are reset at the beginning of each
+  // deliverAndLogToVat, set by syscall implementations, and
+  // incorporated into the DeliveryStatus result of deliverAndLogToVat
+
+  let vatRequestedTermination;
+  let illegalSyscall;
```
mhofman, line 328: It would have been cool if this refactor would have removed this global state, which is currently one of the impediment to parallel vat processing.

However I still don't see an easy way to plumb delivery specific `requestTermination` and `vatFatalSyscall` through `vatWarehouse.deliverToVat()`, or if that's even the right approach.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -350,22 +352,31 @@ export default function buildKernel(
...
-   *
+   * @typedef { {
+   *    abort?: boolean, // changes should be discarded, not committed
+   *    popDelivery?: boolean, // discard the aborted delivery
+   *    computrons?: BigInt, // computron count for run policy
+   *    meterID?: string, // deduct those computrons from a meter
+   *    decrementReapCount?: { vatID: VatID }, // the reap counter should decrement
+   *    terminate?: { vatID: VatID, reject: boolean, info: SwingSetCapData }, // terminate the vat and notify vat-admin
+   *    vatAdminMethargs?: SwingSetCapData, // methargs to notify vat-admin about create/upgrade results
+   * } } CrankResults
```
mhofman: I'm not sure I like this mish mash of data coming from different places since it's really hard to tell which fields may exist together, or what method defines each field.

First, I'm wondering if it'd be worth it to split the base delivery results from the "add-ons" like `popDelivery` and `vatAdminMethargs`. I was also wondering if we should groups the ones that belong together (like `abort` and `terminate`, or `computrons` and `meterId`).

At first I was thinking that a union of separate types might work better, but I'm not so sure anymore.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -375,75 +386,164 @@ export default function buildKernel(
...
-
       insistVatDeliveryResult(deliveryResult);
       // const [ ok, problem, usage ] = deliveryResult;
-      if (deliveryResult[0] === 'ok') {
-        return { metering: deliveryResult[2] };
-      } else {
-        // probably a hard metering fault, or a bug in the vat's dispatch()
-        return { terminate: deliveryResult[1] }; // might be dead
+
+      const status = { illegalSyscall, vatRequestedTermination };
```
mhofman, line 402: Nit:
```suggestion
      /** @type {DeliveryStatus} */
      const status = { illegalSyscall, vatRequestedTermination };
```

And remove the `@returns` JSDoc from the function, to let tsc flow the type.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -350,22 +352,31 @@ export default function buildKernel(
...
    * @typedef { {
-   *    vatID?: VatID, // vat to which the delivery was made
    *    metering?: MeterConsumption | null, // delivery metering results
-   *    useMeter?: boolean, // this delivery should count against the vat's meter
-   *    decrementReapCount?: boolean, // the reap counter should decrement
-   *    discardFailedDelivery?: boolean, // crank abort should not repeat the delivery
-   *    terminate?: string | null, // vat should be terminated
+   *    deliveryError?: string, // delivery failed
+   *    illegalSyscall?: { vatID: VatID, info: SwingSetCapData }, // 'problem' indicated by an illegal syscall
+   *    vatRequestedTermination?: { reject: boolean, info: SwingSetCapData },
```
mhofman: Nit

```suggestion
   *    illegalSyscall: { vatID: VatID, info: SwingSetCapData } | undefined, // 'problem' indicated by an illegal syscall
   *    vatRequestedTermination: { reject: boolean, info: SwingSetCapData } | undefined,
```

packages/SwingSet/src/kernel/vat-warehouse.js
```diff
@@ -353,21 +358,28 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
...
+  async function killWorker(vatID) {
     try {
       await evict(vatID);
     } catch (err) {
       console.debug('vat termination was already reported; ignoring:', err);
     }
   }
 
-  async function destroyWorker(vatID) {
+  async function abandonWorker(vatID) {
```
mhofman: Slight naming nit: in my mind abandon is weaker than kill, but here it seems to be `kill` + bury it.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -375,75 +386,164 @@ export default function buildKernel(
...
+        // * hard metering fault (E_TOO_MUCH_COMPUTATION or
+        //   E_STACK_OVERFLOW or E_NOT_ENOUGH_MEMORY), and the worker is
+        //   dead
+        // * a bug in the vat's dispatch() or liveslots, and we
+        //   shouldn't trust the surviving worker
+        // * Liveslots observed a vat-fatal error, like startVat()
+        //   noticed buildRootObject() did not restore all virtual
+        //   Kinds, and we're not going to use the worker
+        //
+        // So in all cases, our caller should abandon the worker.
```
mhofman, line 420: Is it abandon or kill we should do?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -375,75 +386,164 @@ export default function buildKernel(
...
+   * finishing up the crank.
+   *
+   * Two flags influence this:
+   *  `decrementReapCount` is used for deliveries that run userspace code
+   *  `meterID` means we should check a meter
+   *
+   * @param {VatID} vatID
+   * @param {DeliveryStatus} status
+   * @param {boolean} decrementReapCount
+   * @param {MeterID=} meterID
```
mhofman: I think we switched to the [JSDoc format for optional params](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns) (and avoid the Closure syntax)

```suggestion
   * @param {MeterID} [meterID]
```

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -375,75 +386,164 @@ export default function buildKernel(
...
+   */
+  function deliveryCrankResults(vatID, status, decrementReapCount, meterID) {
+    let meterUnderrun = false;
+    let computrons;
+    if (status.metering?.compute) {
+      computrons = BigInt(status.metering.compute);
+    }
+    // TODO metering.allocate, some day
+
+    const results = { computrons };
```
mhofman: Nit (and remove explicit `@returns`)

```suggestion
    /** @type {CrankResults} */
    const results = { computrons };
```

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -375,75 +386,164 @@ export default function buildKernel(
...
+    }
+
+    if (decrementReapCount && !(results.abort || results.terminate)) {
+      results.decrementReapCount = { vatID };
+    }
+
+    // We leave results.popDelivery up to the caller.  Send failures
+    // never set results.popDelivery (we allow the delivery to be
+    // re-attempted and it splats against the now-dead vat, or doesn't
+    // get delivered until the vat is unsuspended).  But failures in
```
mhofman, line 499: The "doesn't get delivered until the vat is unsuspended" is a future thing, right?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -375,75 +386,164 @@ export default function buildKernel(
...
+    assert(vatInfo, 'routeSendEvent() should have noticed dead vat');
+    const { enablePipelining, meterID } = vatInfo;
 
     /** @type { KernelDeliveryMessage } */
     const kd = harden(['message', target, msg]);
     const vd = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd);
 
-    if (vatInfo.enablePipelining && msg.result) {
+    if (enablePipelining && msg.result) {
+      // TODO maybe move after deliverAndLogToVat??
```
mhofman, line 527: Good question. I don't think it hurts to keep here.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -999,135 +1166,98 @@ export default function buildKernel(
       message,
     });
     /** @type { PolicyInput } */
-    let policyInput = ['none'];
+    let policyInput = ['crank', {}];
```
mhofman: I'm confused, a delivery crank can still result in a no-delivery, which should be indicated as `'none'` but I actually don't see how that can happen anymore.

A no-delivery would be if `result.computrons === undefined`, right ?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -660,42 +773,108 @@ export default function buildKernel(
...
+    function makeFailure(_errorCD) {
+      insistCapData(_errorCD); // capdata(Error)
+      // const args = [upgradeID, false, JSON.parse(errorCD.body)];
+      // actually we shouldn't reveal the details, so instead we do:
+      const error = {
+        '@qclass': 'error',
+        name: 'Error',
+        message: 'vat-upgrade failure',
+      };
+      const args = [upgradeID, false, error];
```
mhofman, line 839: Wondering if we could type this to match the expected args, something along the lines of
```js
      /** @type {Parameters<ReturnType<import('../vats/vat-admin/vat-vat-admin.js').buildRootObject>['vatUpgradeCallback']>} */
```
(of course pointless if that method's params aren't themselves typed.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -660,42 +773,108 @@ export default function buildKernel(
...
+      // notify vat-admin of the failed upgrade
+      const vatAdminMethargs = makeFailure(results1.terminate.info);
+
+      // we still report computrons to the runPolicy
+      const results = harden({
+        ...results1,
+        computrons,
+        abort: true, // always unwind
+        popDelivery: true, // don't repeat the upgrade
+        terminate: undefined, // do *not* terminate the vat
```
mhofman, line 876: This is definitely a clever/unique combination.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -999,135 +1166,98 @@ export default function buildKernel(
...
+    // TODO: policyInput=['crank-failed',{}] is meant to cover
+    // deliveries which fail so badly we don't get metering data (but
+    // which might have taken up considerable time
+    // anyways). Previously, this was triggered by meter underflow
+    // (which only happens when we have metering, so the metering data
+    // is more accurate and should be reported directly), or for a
+    // delivery failure (deliveryResult[0]!=='ok') (which means hard
+    // metering fault, bug in liveslots, or startVat() Kind
+    // restoration error). We now get metering for everything but hard
+    // metering faults.
```
mhofman, line 1187: Shouldn't a hard metering fault just assume the max computron limit?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -999,135 +1166,98 @@ export default function buildKernel(
...
+    // (Reported through vatFatalSyscall() and the 'illegalSyscall'
+    // flag), those errors require the delivery consequences be
+    // unwound (and the vat either terminated or suspended
+    // somehow). syscall.exit(isfailure=true) requests an
+    // unwind. Failed upgrades also require an unwind and set .abort
+    // directly.
+
+    if (crankResults.abort) {
+      // errors unwind any changes the vat made
+      kernelKeeper.abortCrank();
```
mhofman, line 1221: I am not sure I manage to wrap my head around the implications of aborting the crank before doing all the other things like decrementing the reapCounter. It looks like meter and popping delivery were explicitly handled after abort, and I suppose reap is never happening if we abort, so it's equivalent?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -999,135 +1166,98 @@ export default function buildKernel(
...
+    // metering fault (same), or by syscall.exit()
 
-      // state changes reflecting the termination must also survive, so these
-      // happen after a possible abortCrank()
-      await terminateVat(vatID, reject, info);
-      kernelSlog.terminateVat(vatID, reject, info);
+    if (crankResults.terminate) {
+      const { vatID, reject, info } = crankResults.terminate;
       kdebug(`vat terminated: ${JSON.stringify(info)}`);
+      kernelSlog.terminateVat(vatID, reject, info);
```
mhofman, line 1266: Nit: the slogging seem to have moved before the operation

mhofman (changes requested):
The general logic seems sound, and fairly well detailed. Thankfully it reduces the scope of some of the global state for vat termination / crank abort.

I am concerned that the no-op delivery cranks now report as `'crank'` where they previously were `'none'`.

Regarding type nits, in general I'd recommend letting types flow out and avoid explicitly specifying `@returns` unless needed.

For the `vatWarehouse` type loss however this is one of those cases where I wish we had `noImplicitAny` because that regression is way too easy to miss.

The vatAdminMethargs feels very ad-hoc, but I suppose until we have better/more use cases for the kernel generating these calls, it'd be too early to abstract.

The must change is adding an explicit type to `vatWarehouse`, but I'd also like to restore previous the policyInput behavior on no-delivery.

packages/SwingSet/src/kernel/kernel.js
mhofman: Sorry re-reading I realize this is not actionable, probably because there isn't much of an action to take.

packages/SwingSet/src/kernel/kernel.js
mhofman: Bad line quote. Comment says abandon, code does kill

packages/SwingSet/src/kernel/kernel.js
warner: Yeah, when we get there (#5747) we'll need a per-vat structure to hold the state changes that we accumulate during a (potentially-parallel) delivery, which includes DB writes, queue additions, and flags like `requestTermination`. Then at the (in-consensus) crank commit time, we copy all those state changes over into the kernelDB, also processing the consequences of the flags.

packages/SwingSet/src/kernel/kernel.js
warner: Hm, I hear you. The general idea is that `CrankResults` is a uniform way for all types of run-queue event handlers to tell the shared end-of-crank handling code what to do: they represent instructions. So each field should be treated as a separate component of those instructions, and the recipient doesn't care where they came from (it's supposed to be an abstraction boundary). If we introduced additional structure, I think the recipient would need to become more aware of the details, which would leak complexity across that boundary.

The split between `DeliveryStatus` and `CrankResults` made this a lot easier to think about. `DeliveryStatus` is only created by one place, so it's a better candidate for a type like what I believe you're thinking about. I found that allowing each event handler to build `DeliveryStatus` itself (albeit with a common helper function that does most of the work) made the end-of-crank instruction-following code a lot simpler.

What I will do is add a comment to `deliveryCrankResults` to make it clear that it's a helper function that does 80% of the work, with a return value that's meant to be customized for the particular event type, especially for types that need to combine multiple deliveries (like upgrade).

packages/SwingSet/src/kernel/vat-warehouse.js
warner: yeah, I need to clean up this collection of names, they represent some fuzzy thinking on my part about what I really want the vat-warehouse to do

packages/SwingSet/src/kernel/kernel.js
warner: yep, I'm trying to leave hints about where pause/unpause will get implemented when I think of them

packages/SwingSet/src/kernel/kernel.js
warner: Yeah, this PR changes the `runPolicy` input in a number of interesting ways (getting computrons for `startVat`/`buildRootObject` being the biggest one), so I wasn't sure how much energy to put into what used to be `none`.

Everything we feed into the runPolicy is a proxy for estimating wallclock time. Processing a run-queue event that doesn't result in a delivery takes some amount of time, doing an actual delivery takes some additional amount, spending computrons on the worker side takes even more.

To report a difference between "a delivery happened" and "no delivery happened", I can think of two options:

* 1: assume that all deliveries (or all that matter) are to vats which can meter, so treat a non-zero `crankresults.computrons` as `policyInput = ['crank'..]`, and zero- or missing `.computrons` as `['none']`
* 2: add a new field, maybe `crankresults.delivered: true`, which is set for any delivery, even to workers that don't meter

Option 1 would report `none` for e.g. comms-vat deliveries, which use `managerType: local`. Option 2 probably just needs an extra assignment in `deliveryCrankResults`.

I'll go with option 2.

packages/SwingSet/src/kernel/kernel.js
warner: yeah, that'd be nice, I definitely had errors during development because I fed the wrong arguments into this callback, lemme see what I can do quickly

packages/SwingSet/src/kernel/kernel.js
warner: Yeah, needing to break those cases apart is what drove the separation of `DeliveryStatus` and `CrankResults`: normal `dispatch.deliver` sends want to terminate and re-queue (at least for now, since we have all that splat logic already), but failed upgrade wants to do exactly the opposite.

When we get to separate input queues for each vat, I think any case that terminates the vat will also walk the input queue (in reverse order) and preemptively splat each pending message, so we can delete the input queue at the same time as we delete the vat worker. (unless that represents too much work to do all at once, in which case the mostly-dead vat will exist in a state with a non-empty input queue and no worker, and each time it gets scheduled, we'll splat a few more messages, and then when the input and output queues are finally empty, we can finish deleting it)

packages/SwingSet/src/kernel/kernel.js
warner: Hm, hard metering fault means it ran up against a number that's defined in packages/xsnap (https://github.com/Agoric/agoric-sdk/blob/5d49ad79947f44a7cbe98d232ecde105223763d5/packages/xsnap/src/xsnap.js#L21), which is not immediately available to the delivery code (or rather, I haven't thought enough about it to have a clean way to report it, that doesn't make deeper assumptions about the xsnap API). I like the idea, though.. if a vat is terminated this way, we've definitely spent some significant wallclock time already.

I'll make a separate ticket about cleaning this part up: #5752

packages/SwingSet/src/kernel/kernel.js
warner: Yeah, previously we had a mix of changes in the crank buffer by the time we decided to `abort` or `commit`: some made by the vat (syscall consequences), some made by the kernel (decrementing reapCounter, incrementing crankNum, deducting meters, pushing vat-admin messages on the run-queue). So if we decided to abort, we had to re-perform any state changes that were supposed to survive (yes crankNum, yes meter deduction, yes vat-admin, but not syscalls, and not reapCounter because the vat is being terminated anyways, or at least the worker state is being unwound so the amount of garbage is not changing).

This new approach makes the `abort` decision early, and then most of the kernel changes are applied *after* the `abortCrank`, so we don't need to remember them in RAM across the abort point. `CrankResults` remembers `.terminate` (and executes it after any possible `abortCrank`) rather than calling `terminateVat` from inside `processSend` (before the abort) and then recording the details for possible re-application after the abort. Meter deduction is performed after the abort.

packages/SwingSet/src/kernel/kernel.js
warner: yeah, I think that's ok, it will record an intent to terminate, and if the actual executor `terminateVat()` causes any messages, they can appear after the intent record (although I'm pretty sure that `terminateVat()` doesn't do anything that would appear on the slog, it just deletes a bunch of state and enqueues a message to vat-admin)

packages/SwingSet/src/kernel/kernel.js
FUDCo: The primary purpose of JSDoc annotations is to document the code for a human reader.  The way TypeScript-in-JSDoc piggybacks on top of these annotations is convenient but it's a hack and this is one of the places where the seams show.  Removing the `@returns` for the sake of allowing `tsc` to infer the type is letting the tail wag the dog.  Can we have an `@returns` annotation that omits the `{...}` type info but retains the human-readable text or will `tsc` stupidly complain if we do that?

packages/SwingSet/src/kernel/vat-warehouse.js
warner: hm, that doubles the vertical space consumed (grr `prettier`), and requires the reader to recognize the nested destructuring (which I'm just now starting to internalize), so I think I prefer the original

packages/SwingSet/src/kernel/kernel.js
warner: Hrm, so `vatUpgradeCallback` runs in userspace, so e.g. it expects `error` to either be `undefined` or a real JS `Error` object. Whereas the `kernel.js` function is assembling capdata that will be deserialized before `vatUpgradeCallback` is invoked. I don't think we have a way to express a type that is the capdata serialization of a different type.

I'll add a type definition to `vatUpgradeCallback`, and I'll add a pointer to that function from the kernel.js code that builds the arguments, but I don't see a way to do more than that right now.

packages/SwingSet/src/kernel/vat-warehouse.js
warner: ok I refactored it into `stopWorker` and `resetWorker`, which I think captures the distinction better

packages/SwingSet/src/kernel/kernel.js
mhofman: I think it may be able to do what you ask, but I'm not sure. There's always the "solution" to type both the return and this declaration. My concern with having it solely in the returns was that the inferred type for the declaration wasn't correct: it only considered these 2 props, and the other props were completely undocumented even when adding them (and since we run is non-strict type mode, TS happily let you add props that aren't on the declared on the object)

packages/SwingSet/src/kernel/vat-warehouse.js
mhofman: Yeah. I'm still a little confused in the terminology between worker and vat. In my mind the vat is "durable" and contains the state (snapshot + transcript) where the worker is ephemeral and recreated on demand (the process and its ram). However I see we use the term worker to identify more durable aspects. Anyway, that's not specific to this PR, so all good here.

packages/SwingSet/src/kernel/kernel.js
mhofman: I expect the runPolicy to change more dramatically once we have multiple queue, but I think option 2 is a step in the right direction.

packages/SwingSet/src/kernel/kernel.js
mhofman: Oh we should probably have a helper type that does tuple type -> capdata, but yeah, that's another story.

packages/SwingSet/src/kernel/kernel.js
mhofman: > I think any case that terminates the vat will also walk the input queue (in reverse order) and preemptively splat each pending message, so we can delete the input queue at the same time as we delete the vat worker.

I actually don't think we'll be able to. An `exit` syscall will terminate the vat, but all syscalls done in the same crank are still valid and should be processed. And we'll have an ordering guarantee that the outbound queue is processed before the inbound queue. So that means we can't splat anything in the inbound queue until we process the outbound queue, and there is no reason a vat termination should accelerate that processing.

mhofman (approved):
LGTM!

PR #5732, opened 2022-07-07
ci: warn on nested awaits

turadg:
refs: #4770

## Description

@dtribble proposed we enable the await rule everywhere.

### Security Considerations

--

This makes it a warning. It will appear in IDEs and developers may not know what to do with them. If they don't I expect they'll ask.

### Testing Considerations

n/a

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -72,6 +72,24 @@ module.exports = {
   ],
   overrides: [
     {
+      // Tighten rules for exported code.
+      files: ['packages/*/src/**/*.js'],
+      rules: {
+        // The rule is “no nested awaits” but the architectural goal is
+        // “no possibility of ‘awaits sometimes but not always’”. That is our
+        // architectural rule. If it’s too constraining you have to fall back to
+        // promise.then or get a reviewed exception.  “sometimes awaits” is a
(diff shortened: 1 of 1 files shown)


.eslintrc.cjs
```diff
@@ -71,6 +71,13 @@ module.exports = {
     'ava*.config.js',
   ],
   overrides: [
+    {
+      // more strict for exported code
+      files: ['packages/*/src/**/*.js'],
```
dckc, line 76: I think some packages are not direcly under `packages/`; for example `packages/wallet/contract/`.

dtribble (approved):
Works for me.

.eslintrc.cjs
turadg: Good point. That seems like a problem, yeah? I'd prefer to fix that than work around it. 

@samsiegart are you okay with a PR to flatten those? https://github.com/Agoric/agoric-sdk/issues/5733

dckc: I think there's an error stack widget that breaks in cases like `packages/wallet/contract` vs. `@agoric/smart-wallet`, so yeah, these should be fixed. (the other case is `SwingSet` but we seem to be living with that for now.)

Issue #5285, opened 2022-05-04
Make VaultFactory reachably durable


turadg:
## What is the Problem Being Solved?

https://github.com/Agoric/agoric-sdk/pull/5736 made Vault Factory sort of durable, but it has some hacks around https://github.com/Agoric/agoric-sdk/issues/4567 and it's not yet _reachably durable_ .

For the next release we need it to be not just reachably durable, but upgradable: https://github.com/agoric/agoric-sdk/issues/6791

## Description of the Design

Store everything precious.

## Security Considerations

State on disk.

## Test Plan

Manual review that all precious state is stored.

dtribble:
A simpler solution in this case:  upon upgrade I would expect to restart the liquidator activity, which is the loop waiting for the quote promise to be resolved. Basically,  `liquidationInProgress` and `outstandingQuote` need to be made ephemeral.

turadg:
I think this has been done a while, actually.

PR #5736, opened 2022-07-07
4550 vault factory durable (unreachably)

### LINKED ISSUE #4550: make vault collections durable
Related to https://github.com/Agoric/agoric-sdk/issues/4341

#4345 makes vaults use virtual objects. There's additional work to make them durable.

## Description of the Design

Roughly enable the `durable` flag to `makeScalarBigMapStore` and iron out problems.

Use a pattern query to reduce disk IO.

## Test Plan

Integration tests for upgrade using https://github.com/Agoric/agoric-sdk/issues/3272

turadg:
closes: #4550
refs: #5285

## Description

Basic mechanics to enable durability for Vault Factory. Leaves #5285 open until the requirements are fleshed out.

This is shy of "reachably durable", let alone "upgradeable". That's scheduled work.

It also required some hacks, but it works and is prone to merge conflicts so I think it's better to get in and solve the rest on top of it. https://github.com/Agoric/agoric-sdk/issues/5759 may help but needn't be a release blocker.

### Security Considerations

Storing on disk.

--

### Testing Considerations

Existing tests pass, in particular `make scenario2-run-chain-to-halt`.

Doesn't need to demonstrate upgrade or reachability which are out of scope.

A bug was [detected by loadgen](https://github.com/Agoric/agoric-sdk/runs/7438623238?check_suite_focus=true) and fixed in [d04981c](https://github.com/Agoric/agoric-sdk/pull/5736/commits/d04981c33e15bdadb0b858a5d1c3652376c8ea66). I'm noting here because we debated how to add a regression test for that and decided for now to rely on loadgen as the test. https://github.com/Agoric/agoric-sdk/issues/5801 explores another option.

### DIFF
--- a/packages/SwingSet/tools/fakeVirtualSupport.js
+++ b/packages/SwingSet/tools/fakeVirtualSupport.js
@@ -274,7 +274,12 @@ export function makeFakeWatchedPromiseManager(vrm, vom, cm, fakeStuff) {
     fakeStuff.convertSlotToVal,
   );
 }
-
+/**
+ * Configure virtual stuff with relaxed durability rules and fake liveslots
+ *
+ * @param {object} [options]
+ * @param {boolean} [options.relaxDurabilityRules=true]
+ */
 export function makeFakeVirtualStuff(options = {}) {
   const fakeStuff = makeFakeLiveSlotsStuff(options);
   const { relaxDurabilityRules = true } = options;
--- a/packages/inter-protocol/src/vaultFactory/orderedVaultStore.js
+++ b/packages/inter-protocol/src/vaultFactory/orderedVaultStore.js
@@ -1,5 +1,4 @@
 // @ts-check
-import { makeScalarBigMapStore } from '@agoric/vat-data';
 import { fromVaultKey, toVaultKey } from './storeUtils.js';
 
 /**
@@ -15,10 +14,10 @@ import { fromVaultKey, toVaultKey } from './storeUtils.js';
 /** @typedef {import('./vault').Vault} Vault */
 /** @typedef {import('./storeUtils').CompositeKey} CompositeKey */
 
-export const makeOrderedVaultStore = () => {
-  /** @type {MapStore<string, Vault>} */
-  const store = makeScalarBigMapStore('orderedVaultStore', { durable: true });
-
+/**
+ * @param {MapStore<string, Vault>} store
+ */
+export const makeOrderedVaultStore = store => {
   /**
    *
    * @param {string} vaultId
--- a/packages/inter-protocol/src/vaultFactory/orderedVaultStore.js
+++ b/packages/inter-protocol/src/vaultFactory/orderedVaultStore.js
@@ -1,5 +1,4 @@
 // @ts-check
-// XXX avoid deep imports https://github.com/Agoric/agoric-sdk/issues/4255#issuecomment-1032117527
 import { makeScalarBigMapStore } from '@agoric/vat-data';
 import { fromVaultKey, toVaultKey } from './storeUtils.js';
 
@@ -17,9 +16,8 @@ import { fromVaultKey, toVaultKey } from './storeUtils.js';
 /** @typedef {import('./storeUtils').CompositeKey} CompositeKey */
 
 export const makeOrderedVaultStore = () => {
-  // TODO make it work durably https://github.com/Agoric/agoric-sdk/issues/4550
   /** @type {MapStore<string, Vault>} */
-  const store = makeScalarBigMapStore('orderedVaultStore', { durable: false });
+  const store = makeScalarBigMapStore('orderedVaultStore', { durable: true });
 
   /**
    *
--- a/packages/inter-protocol/src/vaultFactory/prioritizedVaults.js
+++ b/packages/inter-protocol/src/vaultFactory/prioritizedVaults.js
@@ -46,11 +46,12 @@ export const currentDebtToCollateral = vault =>
  * Vaults, ordered by their debt ratio so that all the vaults below a threshold
  * can be quickly found and liquidated.
  *
- * @param {(highestRatio: Ratio) => void} higherHighestCb called when a new
+ * @param {MapStore<string, Vault>} store
+ * @param {(highestRatio: Ratio) => void} [higherHighestCb] called when a new
  * vault has a higher debt ratio than the previous highest
  */
-export const makePrioritizedVaults = (higherHighestCb = () => {}) => {
-  const vaults = makeOrderedVaultStore();
+export const makePrioritizedVaults = (store, higherHighestCb = () => {}) => {
+  const vaults = makeOrderedVaultStore(store);
 
   /**
    * Called back when there's a new highestRatio and it's higher than the previous.
--- a/packages/inter-protocol/src/vaultFactory/vault.js
+++ b/packages/inter-protocol/src/vaultFactory/vault.js
@@ -101,7 +101,6 @@ const validTransitions = {
  * idInManager: VaultId,
  * manager: VaultManager,
  * vaultSeat: ZCFSeat,
- * zcf: ZCF,
  * }>} ImmutableState
  */
 
@@ -135,6 +134,7 @@ const validTransitions = {
  * @type {(key: VaultId) => {
  * storageNode: ERef<StorageNode>,
  * marshaller: ERef<Marshaller>,
+ * zcf: ZCF,
  * }} */
 // @ts-expect-error not yet defined
 const provideEphemera = makeEphemeraProvider(() => ({}));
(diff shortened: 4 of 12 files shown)


packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -550,8 +550,8 @@ const finish = async ({ state }) => {
  * @param {import('@agoric/governance/src/contractGovernance/typedParamManager').TypedParamManager<import('./params.js').VaultDirectorParams>} directorParamManager
  * @param {ZCFMint<"nat">} debtMint
  */
-const makeVaultDirector = defineKindMulti(
-  'VaultDirector',
+const makeVaultDirector = defineDurableKindMulti(
+  makeKindHandle('VaultDirector'),
```
Chris-Hibbert: This would make a new handle if the vaultDirector were ever revived. 

The up-to-date approach is there's a `baggage` parameter, and you use `provide` to retrieve the handle if it already exists.

```suggestion
const vaultDirectorKindHandle = provideKindHandle(baggage, 'vaultDirector');
const makeVaultDirector = defineDurableKindMulti(
  vaultDirectorKindHandle,
```

packages/run-protocol/src/vaultFactory/vaultHolder.js
```diff
@@ -81,8 +81,8 @@ const holder = {
 
 const behavior = { helper, holder };
 
-export const makeVaultHolder = defineKindMulti(
-  'VaultHolder',
+export const makeVaultHolder = defineDurableKindMulti(
+  makeKindHandle('VaultHolder'),
```
Chris-Hibbert: use  `provideKindHandle`

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -813,8 +816,8 @@ const behavior = {
   self: selfBehavior,
 };
 
-const makeVaultManagerKit = defineKindMulti(
-  'VaultManagerKit',
+const makeVaultManagerKit = defineDurableKindMulti(
+  makeKindHandle('VaultManagerKit'),
```
Chris-Hibbert: use `provideKindHandle`

Chris-Hibbert (approved):

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: > This would make a new handle if the vaultDirector were ever revived.

That will only happen for reasons other than upgrade? The scope of this PR durability but not upgradability in place. (i.e. that the necessary data is available in storage)

When `provideKindHandle` is in master, I could use it with empty baggage and have a TODO to provide baggage when making it upgradeable. How's that sound?

Chris-Hibbert: I'm not sure where the cut-off for durability vs. upgradeability is. I can see an argument for stopping with what you have, but I can also argue that if the goal is upgradeability that's plausible to a reviewer, though not tested, I think we should get as far as wiring up the code to reinitialize the handleKinds.

@erights, do you have an opinion?

turadg: Recapping from other discussions: MN-1 goal is "reachably durable" and "durable" is a milestone towards that. The scope of this PR is just "durable" (no baggage).

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -3,55 +3,74 @@
...
+
+/**
+ * Ephemera is the state we cannot (or merely need not) keep durably. For
+ * vaultDirector we can keep it in module scope because there is (exactly) one
+ * vaultDirector per vaultFactory contract and (exactly) one contract per vat.
+ *
+ * @type {Ephemera}
+ */
+// @ts-expect-error not actually full until after initState
+// UNTIL resolution to https://github.com/Agoric/agoric-sdk/issues/5759
```
Chris-Hibbert: should `UNTIL` be in [Coding-Style](https://github.com/Agoric/agoric-sdk/wiki/Coding-Style)?

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -443,8 +472,8 @@ const creatorBehavior = {
    * @param {MethodContext} context
    * @param {string} name
    */
-  getInvitation: ({ state }, name) =>
-    state.directorParamManager.getInternalParamValue(name),
+  getInvitation: (context, name) =>
```
Chris-Hibbert:
```suggestion
  getInvitation: (_context, name) =>
```

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -465,53 +494,48 @@ const publicBehavior = {
...
+  getRunIssuer: () => ephemera.debtMint.getIssuerRecord().issuer,
   /**
    * subscription for the paramManager for a particular vaultManager
    *
    * @param {MethodContext} context
    * @param {{ collateralBrand: Brand }} selector
    */
-  getSubscription: ({ state }, { collateralBrand }) =>
-    state.vaultParamManagers.get(collateralBrand).getSubscription(),
+  getSubscription: (context, { collateralBrand }) =>
```
Chris-Hibbert:
```suggestion
  getSubscription: (_context, { collateralBrand }) =>
```

packages/run-protocol/src/vaultFactory/vaultDirector.js
```diff
@@ -465,53 +494,48 @@ const publicBehavior = {
...
-  getElectorateSubscription: ({ state }) =>
-    state.directorParamManager.getSubscription(),
+  getElectorateSubscription: () =>
+    ephemera.directorParamManager.getSubscription(),
   /**
    * @param {MethodContext} context
    * @param {{ collateralBrand: Brand }} selector
    */
-  getGovernedParams: ({ state }, { collateralBrand }) =>
+  getGovernedParams: (context, { collateralBrand }) =>
```
Chris-Hibbert: Why isn't lint complaining about these?

```suggestion
  getGovernedParams: (_context, { collateralBrand }) =>
```

packages/run-protocol/src/vaultFactory/vaultHolder.js
Chris-Hibbert: Not needed for merely durable target.

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -672,19 +716,34 @@ const managerBehavior = {
...
   /** @param {MethodContext} context */
   getCompoundedInterest: ({ state }) => state.compoundedInterest,
 };
 
 const selfBehavior = {
   /** @param {MethodContext} context */
-  getGovernedParams: ({ state }) => state.factoryPowers.getGovernedParams(),
+  getGovernedParams: ({ state }) => {
+    const { factoryPowers } = provideEphemera(state.collateralBrand);
+    assert(factoryPowers);
```
Chris-Hibbert: I think `assert()` without a message is pretty hard to track down. Once there's a message, it becomes much easier, even when the message is duplicated within the source file. If I'm right about that, it would be worth changing all of these to e.g.
```suggestion
    assert(factoryPowers, 'factoryPowers');
```

packages/run-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: Not necessary for mere durability.

Chris-Hibbert (approved):

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: Good question. I figure it's self-evident enough and I don't want to claim that everyone should use it.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: This feels like a lint policy to me. It passes lint so I'm going to leave it alone.

For the readability aspect, IDEs often give a hint by dimming:
Screen Shot 2022-07-20 at 2 28 08 PM

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: I agree that would be better. I intend to make it so the assertion isn't necessary so I'm opting to invest the effort on that instead.

packages/run-protocol/src/vaultFactory/vaultDirector.js
turadg: Because it's a param list and the value after is is used.

PR #5737, opened 2022-07-08
feat(notifier): Add makeNotifierFromSubscriber

### LINKED ISSUE #5413: notifier returned by makeNotifierFromAsyncIterable is lossless

gibson042:
Fixes #5413

## Description

Add `makeNotifierFromSubscriber` for lossy consumption of PublishKit subscriber objects via `subscribeLatest` (each `getUpdateSince()` call returns a unique already-settled promise for the latest result, each `getUpdateSince(lastUpdateCount)` call requests the next value [which under the covers constructs a new iterator] and returns a unique promise for it).

Also deprecates `makeNotifierFromAsyncIterable`.

This PR should land with a rebase rather than a squash; I have intentionally included the changes from #5695 along with an immediate reversion for posterity.

### Security Considerations

None known.

https://github.com/Agoric/documentation/issues/684

### Testing Considerations

More tests to come before merging.


packages/notifier/src/notifier.js
```diff
@@ -152,7 +153,106 @@ export const makeNotifierKit = (...initialStateArr) => {
...
+     */
+    getSharableNotifierInternals: () => baseNotifier,
+    getStoreKey: () => harden({ notifier }),
+  });
+  return notifier;
+};
+harden(makeNotifierFromSubscriber);
+
+/**
+ * Deprecated adaptor from async iterable to notifier.
```
turadg: https://www.typescriptlang.org/play#example/jsdoc-deprecated
```suggestion
 * @deprecated adaptor from async iterable to notifier.
```

packages/notifier/test/test-notifier-adaptor.js
```diff
@@ -136,3 +140,211 @@ test('notifier adaptor - update from notifier fails', t => {
   const n = makeNotifierFromAsyncIterable(explodingStream);
   return observeNotifier(n, u);
 });
+
+// /////////////////////////////// makeNotifierFromSubscriber /////////////////////////////////
```
turadg: please separate this into its own file. that will make it easier when doing watch tests for user to focus on what they want.

packages/notifier/test/test-notifier-adaptor.js
```diff
@@ -136,3 +140,211 @@ test('notifier adaptor - update from notifier fails', t => {
...
+    [1, 2, 4, 8, 16, 32, 'done'],
+    'only the last of values published between iteration steps should be observed',
+  );
+  t.deepEqual(await notifier.getUpdateSince(), {
+    value: 'done',
+    updateCount: undefined,
+  });
+});
+
+// TODO: makeNotifierFromSubscriber(fails)
```
turadg: In general,
```suggestion
test.todo('makeNotifierFromSubscriber(fails)')
```
But this will be done before merging?

packages/notifier/test/test-notifier-adaptor.js
```diff
@@ -136,3 +140,211 @@ test('notifier adaptor - update from notifier fails', t => {
...
+test('makeNotifierFromSubscriber - getUpdateSince timing', async t => {
+  const { initialize, publishNextBatch, subscriber } = makeGeometricPublishKit({
+    conclusionMethod: 'finish',
+    conclusionValue: 'done',
+  });
+  const notifier = await makeNotifierFromSubscriber(subscriber);
+
+  const sequence = [];
+  const firstP = notifier.getUpdateSince();
+  firstP.then(_ => sequence.push('resolve firstP'));
```
turadg: please address the floating promises on 250 and 252

packages/notifier/test/test-notifier-adaptor.js
```diff
@@ -136,3 +140,211 @@ test('notifier adaptor - update from notifier fails', t => {
...
+  const firstP2 = notifier.getUpdateSince();
+  firstP2.then(_ => sequence.push('resolve firstP2'));
+
+  await E(Promise).resolve();
```
turadg: why go through E proxy?

packages/notifier/test/test-notifier-adaptor.js
```diff
@@ -136,3 +140,211 @@ test('notifier adaptor - update from notifier fails', t => {
   const n = makeNotifierFromAsyncIterable(explodingStream);
   return observeNotifier(n, u);
 });
+
+// /////////////////////////////// makeNotifierFromSubscriber /////////////////////////////////
+
+/** @param {{conclusionMethod: 'finish' | 'fail', conclusionValue: any}} config */
+const makeGeometricPublishKit = ({ conclusionMethod, conclusionValue }) => {
```
turadg: this sounds like it's making a PublicKit but it's returning something different.

```suggestion
const makeGeometricPublishHelper = ({ conclusionMethod, conclusionValue }) => {
```
Looks like calls have the same arguments. Why parameterize?
```suggestion
const makeGeometricPublishHelper = () => {
```
Also looks like all tests share this first line. Once this is in its own file, it would work well to put in a hook:
```
test.beforeEach(t => {
  Object.assign(t.context, makeGeometricPublishHelper());
}
```
Or even,
```
test.beforeEach(t => {
  Object.assign(t.context, makeGeometricPublishHelper());
  t.context.notifier = await makeNotifierFromSubscriber(t.context.subscriber);
}
```

packages/notifier/src/notifier.js
turadg: Also please explain here why it's deprecated.

> I have intentionally included the changes from https://github.com/Agoric/agoric-sdk/pull/5695 along with an immediate reversion for posterity.

I request that those two commits be omitted (or this be landed as squash). For master it's noise. If there's something important to convey about how we got to this point then just write it in HEAD. If you want there to be some record of the other work done, it's already in the referenced PR.

turadg (changes requested):

packages/notifier/src/notifier.js
```diff
@@ -152,7 +153,106 @@ export const makeNotifierKit = (...initialStateArr) => {
 };
 
 /**
- * Adaptor from async iterable to notifier.
+ * @template T
+ * @param {ERef<Subscriber<T>>} subscriberP
+ * @returns {Promise<Notifier<T>>}
+ */
+export const makeNotifierFromSubscriber = async subscriberP => {
```
turadg:
```suggestion
export const makeNotifierFromSubscriber = async subscriber => {
```
I assume `P` is for "Promise" but that's not exactly what this is.

packages/notifier/src/notifier.js
gibson042: I'm anticipating the attempt to come up again, and would like the prior record to outlive that dead branch.

packages/notifier/test/test-notifier-adaptor.js
gibson042: Yes, done now.

packages/notifier/test/test-notifier-adaptor.js
gibson042: Address in what way?

packages/notifier/test/test-notifier-adaptor.js
gibson042: Abandoned.

packages/notifier/test/test-notifier-adaptor.js
gibson042: Renamed for clarity, but kept it outside the test context.

packages/notifier/src/notifier.js
gibson042: It's an ERef, which _may_ be a promise. I think we do use this pattern to highlight that potential.

gibson042:
@turadg https://github.com/Agoric/agoric-sdk/pull/5737/commits/f40b7ed073e97d961b92f5e28c03a747c91f47e7 addresses everything except the ["floating promises" concern](https://github.com/Agoric/agoric-sdk/pull/5737#discussion_r916970622)... what kind of change are you expecting for that?

packages/notifier/test/test-notifier-adaptor.js
turadg: make the warnings go away :) 

either by awaiting or putting `void ` to indicate it need not be awaited. (the warning string explains another option is to `.catch` but that seems silly here)

packages/notifier/test/test-notifier-adaptor.js
turadg: Makes sense. I was wrong about the parameterization

packages/notifier/src/notifier.js
turadg: I've also seen `P` for Presence, kind of the opposite.

I don't think we have any consistent guidelines on this but I would make the case that when possible we rely on the type system for whether something is a promise or not. If there's no need to bind both the promise and the resolved value then use the simpler name. non-blocking request.

turadg (commented):
LG. I'll approve when the three new eslint warnings are resolved.

gibson042:
I didn't see the promise ones locally. Anyway, resolved and pushed.

turadg (approved):
👍

packages/notifier/src/notifier.js
```diff
@@ -151,9 +152,113 @@ export const makeNotifierKit = (...initialStateArr) => {
   return harden({ notifier, updater });
 };
 
+/**
+ * @template T
+ * @param {ERef<Subscriber<T>>} subscriberP
+ * @returns {Promise<Notifier<T>>}
+ */
+export const makeNotifierFromSubscriber = async subscriberP => {
```
turadg: This isn't an async function:
```suggestion
export const makeNotifierFromSubscriber = subscriberP => {
```

I don't think it needs to return `Promise` either. Can't it return the `Notifier` as `makeNotifierFromAsyncIterable` does?

turadg (changes requested):

turadg:
Working on https://github.com/Agoric/agoric-sdk/pull/5704/ I ran into some issues using this so I took another look and found a few things to address:
- whether it [needs to be async](https://github.com/Agoric/agoric-sdk/pull/5737#discussion_r917183858)
- [documentation](https://github.com/Agoric/documentation) ticket for notifiers.html
- some confidence that the the `???` in https://github.com/Agoric/agoric-sdk/pull/5704 aren't due to bugs here

packages/notifier/src/notifier.js
turadg: The branch will be dead but https://github.com/Agoric/agoric-sdk/pull/5695 will endure for anyone who wants to see it.

Non-blocking but I do think the PR suffices as a record and to the point of discover as to why this is the design, I don't think commit history suffices. If that's the goal is should be written into HEAD.

turadg (approved):
🎉 I rebased my branch onto this version and the contract tests pass!

 :shipit: 😸

Issue #3995, opened 2021-10-25
Smart wallets complete for Launch


michaelfig:
## Organization

This epic is the last in a sequence of milestone epics.

### Acceptance criteria
- On-chain wallets validated to be working well enough for MN-1 release

## Background
Why?
 - Avoid users having to manage client state (“I bricked my machine” #2629)
 - Making an electron app (cf. [exo](https://github.com/endojs/Exo)) is risky
 - Asking users to install an electron app is burdensome (though not beyond market norms)
 - reduce round-trips (#3802 )

### ledger integration

@warner to research JSON blob details etc. what can be signed? Hash? JSON blob?

_Originally posted by @dckc in https://github.com/Agoric/agoric-sdk/issues/2629#issuecomment-948970573_

# Design notes

## Normalised data model

For a given user-specific wallet, the backend's data records each have:
- a `meta` property consisting of `{ id, createdAt, modifiedAt }`, where `id` is unique for the specific wallet instance, and the `*At` stamps are fodder for `new Date(somethingAt)`
- an `actions` presence that allows the manipulation of that record

Every collection of data records is enumerable and presents a notifier that is updated with the entire collection's contents as any changes are made.

If there are any wallet entities that don't conform to this data model (i.e. currently petnames and dapps), they should be updated to conform.

## Private data

User annotations on data (e.g. the assigned petnames, and any comment fields) should be private, that is they should be encrypted in such a way that the user's HD signer or recovery phrase can decrypt them.

It may be advantageous to represent this encrypted data as a single normalised record mapping from `id` to associated data so that it can be decrypted and re-encrypted in bulk.  Because it contains timestamps in its `"meta"` property, we will know when it applies (as suggested by @JimLarson).

## Cosmos chain integration

Pluggable storage of normalised records is an important feature of the wallet.  For a strawman implementation, the storage system would use the IAVL tree, to be able to expose them via Cosmos-idiomatic queries, events, and transactions.

The wallet backend and wallet frontend need an agreed-upon wallet middleware to encode and decode messages to and from the chain, without affecting the ocap programming model of either the frontend or backend.

### Pervasive sturdyrefs

If the normalised data model is adopted throughout the user's wallet, the user's wallet admin facet can provide a way to look up a given data record (including its `actions` presence) via its `id`.  This corresponds to a sturdyref whose contents can only be mutated by signed transactions to the appropriate wallet service.

### Queries

[...]

PR #5741, opened 2022-07-08
feat: read only smart wallet

samsiegart:

Testing steps:
- `cd agoric-sdk`
- `yarn && yarn build`
- `cd agoric-sdk/packages/cosmic-swingset`
- `make`
- `make scenario2-setup`
- `make scenario2-run-chain`
- `make scenario2-run-client`
- Get address from `cat t1/8000/ag-cosmos-helper-address`
- `agoric open`

Select "Read-only Address":
image

Then all the purses, issuers, etc. should load in a read-only state.


yarn.lock
```diff
@@ -15301,7 +15301,7 @@ rollup@^2.43.1, rollup@^2.58.0:
   optionalDependencies:
     fsevents "~2.3.2"
 
-"rollup@github:endojs/endo#rollup-2.7.1-patch-1":
+rollup@endojs/endo#rollup-2.7.1-patch-1:
```
turadg: I hope this passes CI. I'd love to stop having to revert these

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -1,53 +1,50 @@
...
     }),
   );
 
-  /**
-   * @template T
-   * @param {ERef<Notifier<T>>} notifier
-   */
-  const iterateNotifier = notifier => {
-    const storeKey = E(notifier).getStoreKey();
-    const follower = makeFollower(storeKey, makeLeader);
```
turadg, line 24: what's reading the store now? oh castingSpec specifies the key

turadg (approved):
I'm not very well acquainted with `@agoric/casting` but the code makes sense to me.

If the test plan is checking out then LGTMerge.

PR #5742, opened 2022-07-09
refactor(notifier)!: no initial state for makePublishKit

turadg:
Extracted from https://github.com/Agoric/agoric-sdk/pull/5704

## Description

All publishing kits start empty: https://github.com/Agoric/agoric-sdk/pull/5621

So `makePublishKit` shouldn't take an initial value and there's no need to have a separate `makeEmptyPublishKit`.

### Security Considerations

--

--

### Testing Considerations

Updated

### DIFF
--- a/packages/notifier/test/test-publish-kit.js
+++ b/packages/notifier/test/test-publish-kit.js
@@ -2,7 +2,6 @@
 
 import { test } from './prepare-test-env-ava.js';
 import {
-  makeEmptyPublishKit,
   makePublishKit,
   subscribeEach,
   subscribeLatest,
@@ -10,8 +9,8 @@ import {
 import '../src/types.js';
 import { invertPromiseSettlement } from './iterable-testing-tools.js';
 
-test('makeEmptyPublishKit', async t => {
-  const { publisher, subscriber } = makeEmptyPublishKit();
+test('makePublishKit', async t => {
+  const { publisher, subscriber } = makePublishKit();
 
   const pubFirst = Symbol('first');
   publisher.publish(pubFirst);
@@ -136,8 +135,8 @@ test('makeEmptyPublishKit', async t => {
   );
 });
 
-test('makeEmptyPublishKit - immediate finish', async t => {
-  const { publisher, subscriber } = makeEmptyPublishKit();
+test('makePublishKit - immediate finish', async t => {
+  const { publisher, subscriber } = makePublishKit();
 
   const pubFinal = Symbol('done');
   publisher.finish(pubFinal);
@@ -189,8 +188,8 @@ test('makeEmptyPublishKit - immediate finish', async t => {
   );
 });
 
-test('makeEmptyPublishKit - fail', async t => {
-  const { publisher, subscriber } = makeEmptyPublishKit();
+test('makePublishKit - fail', async t => {
+  const { publisher, subscriber } = makePublishKit();
 
   publisher.publish(undefined);
   const subFirst = await subscriber.subscribeAfter();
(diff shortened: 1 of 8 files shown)


erights (approved):
LGTM, thanks!

packages/notifier/test/test-publish-kit.js
```diff
@@ -288,16 +283,17 @@ test('makePublishKit', async t => {
...
       subscriber.subscribeAfter(),
-      subscriber.subscribeAfter(expectedInitialCount - 1n),
+      subscriber.subscribeAfter(0n),
     ]);
     const [subFirst] = subAll;
     t.like(subFirst, {
       head: { done: false },
-      publishCount: expectedInitialCount,
+      publishCount: 1n,
     });
```
gibson042: This test is assuming too much about the implementation details of `makePublishKit` rather than its external interface. We should just drop the `subscribeAfter` argument and `publishCount` assertion.
```suggestion
    const subAll = await Promise.all([
      subscriber.subscribeAfter(),
      subscriber.subscribeAfter(),
    ]);
    const [subFirst] = subAll;
    t.like(subFirst, {
      head: { done: false },
    });
```

gibson042 (approved):

packages/notifier/test/test-publish-kit.js
turadg: > This test is assuming too much about the implementation details

You mean the `publicCount: 1n` part? (The `0n` argument is legit.)

I agree the interface doesn't guarantee that the next value will be one greater. Because there is a `publish()` it does guarantee that it will be greater though, right?

When I read this test in master I thought it was about the counts after publish but the first test in the file checks those. This seems to really be testing publishing particular values (of the `initialValues`) loop. Since there is no "initial value" anymore, and I trust there are already tests for `.publish()`, I'll just remove this test.

@gibson042 PTAAL

turadg:
Looking at the CI output I saw a number of floating promise warnings. Tacking on [c2271c9](https://github.com/Agoric/agoric-sdk/pull/5742/commits/c2271c9414f385c76c89d59bed50c2a0af329abb) to resolve those and will merge by `automerge:rebase`

packages/notifier/test/test-publish-kit.js
```diff
@@ -265,51 +264,8 @@ test('makeEmptyPublishKit - immediate fail', async t => {
   );
 });
 
-test('makePublishKit', async t => {
-  const expectedInitialCount = (
-    await makePublishKit(undefined).subscriber.subscribeAfter()
-  ).publishCount;
-  t.true(expectedInitialCount > 0n);
-  const initialValues = {
```
gibson042, line 273: Removing this eliminates coverage that arbitrary values can be published, but that coverage probably is of little value.

packages/notifier/test/test-publish-kit.js
```diff
@@ -390,15 +346,15 @@ test('subscribeAfter bounds checking', async t => {
...
-  const { publisher: pub2, subscriber: sub2 } = makeEmptyPublishKit();
+  const { publisher: pub1, subscriber: sub1 } = makePublishKit();
+  const { publisher: pub2, subscriber: sub2 } = makePublishKit();
   const sub2LIFO = [];
 
   const sub1FirstAll = [];
-  Promise.resolve(sub1.subscribeAfter()).then(result =>
+  void Promise.resolve(sub1.subscribeAfter()).then(result =>
     sub1FirstAll.push(result),
   );
```
gibson042: Might as well be consistent here.
```suggestion
  Promise.resolve(sub1.subscribeAfter())
    .then(result => sub1FirstAll.push(result))
    .catch(t.fail);
```

packages/notifier/test/test-publish-kit.js
```diff
@@ -390,15 +346,15 @@ test('subscribeAfter bounds checking', async t => {
...
-  Promise.resolve(sub1.subscribeAfter(0n)).then(result =>
+  void Promise.resolve(sub1.subscribeAfter(0n)).then(result =>
     sub1FirstAll.push(result),
   );
```
gibson042:
```suggestion
  Promise.resolve(sub1.subscribeAfter(0n))
    .then(result => sub1FirstAll.push(result))
    .catch(t.fail);
```

gibson042 (approved):

PR #5743, opened 2022-07-09
fix(walletFactory): marshal using petnames

dckc:
using the board is inappropriate as purses etc. are closely held

refs: #4398

the results of marshaling are captured in https://gist.github.com/dckc/3192a5264524635bd7261277b1fb707e


samsiegart (approved):
I don't fully understand what this does but I tested the wallet and it seems to work with this change

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1731,6 +1731,9 @@ export function makeWallet({
 
   const firstPathToLookup = createRootLookups();
 
+  /** @type {ReturnType<typeof makeMarshal>} */
+  const marshaller = harden({ unserialize: hydrate, serialize: dehydrate });
```
turadg, line 1735: nifty

turadg (approved):

packages/wallet/api/src/lib-wallet.js
dckc: Yeah; we already had code to do wallet-specific marshaling.

The next step is to refine it to use board ids for things like brands and issuers so that we can tell when they're the same as brands that we get from `agoricNames` and such.

PR #5744, opened 2022-07-10
feat(wallet-ui): connect to keplr with smart wallet

samsiegart:

This allows the (currently read-only) smart wallet to get the address and signer from Keplr.

Food for thought:
- How do we deploy a smart wallet instance from a Keplr address?
- How do we use the signer for txns once that instance exists and we can read from it?

### DIFF
--- a/packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
+++ b/packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 import { v4 as uuid } from 'uuid';
 import { makeStyles } from '@mui/styles';
 import Autocomplete from '@mui/material/Autocomplete';
@@ -17,6 +17,7 @@ import DialogActions from '@mui/material/DialogActions';
 import Select from '@mui/material/Select';
 import MenuItem from '@mui/material/MenuItem';
 import InputLabel from '@mui/material/InputLabel';
+import { deepEquals } from '../util/DeepEquals';
 
 import { withApplicationContext } from '../contexts/Application';
 import {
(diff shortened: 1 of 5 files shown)


packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -173,6 +174,17 @@ const ConnectionSettingsDialog = ({
   }
 
   const saveAndClose = () => {
+    if (
+      JSON.stringify(config) === JSON.stringify(connectionConfig) &&
```
turadg: is this to do a deep equal or is there something about JSON specifically? don't we have a deep equal util somewhere?

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -173,6 +174,17 @@ const ConnectionSettingsDialog = ({
...
+      config.smartConnectionMethod === SmartConnectionMethod.KEPLR
```
turadg, line 185: check this first since it's cheaper (`cheapCheck && moreExpensiveCheck`)

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -365,7 +377,8 @@ const ConnectionSettingsDialog = ({
           onClick={saveAndClose}
           disabled={
             errors.size > 0 ||
-            JSON.stringify(config) === JSON.stringify(connectionConfig)
+            (JSON.stringify(config) === JSON.stringify(connectionConfig) &&
+              config.smartConnectionMethod !== SmartConnectionMethod.KEPLR)
```
turadg: this checks looks coded about three times. maybe worth a small helper, if only for readability

turadg (approved):
Shallow review since I don't know the Keplr stuff well yet

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
samsiegart: Yes it just happens that in this case JSON.stringify suffices. We don't have one in the frontend (outside of the test env).

I just went ahead and dropped a util function in rather than importing a library to do it.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
samsiegart: Just pulled this into a useMemo

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
samsiegart: Done

PR #5746, opened 2022-07-11
5622 vaultManager ephemera

### LINKED ISSUE #5622: Ephemeral state in vaultManager is incorrectly module-scope

turadg:
closes: #5622

## Description

Vault Manager has some ephemeral state in module scope that is unique per VMgr instance. This makes a map that can hold such state, keyed by the VMgr instance. The singleton pattern was also sketched and [ruled out](https://github.com/Agoric/agoric-sdk/pull/5650#issuecomment-1180486256).

It also extracts `makeEphemeralProvider` to be used in other cases for state on durable kinds that need not (or cannot) be durable.

### Security Considerations

`vaultManager.js` continues to trust `contractSupport.js`.

This is a solution to a challenge common with durable kinds. It may be worth documenting more. At the point I think it's sufficient to have the example and shared utility function.

### Testing Considerations

Ideally we'd have add a regression test that breaks in master and go green now, but the only way I see to cover this case is have liquidations going on one brand while another one is starting and that requires setting up multiple concurrent priceAuthorities which we don't currently support in our test code.

### DIFF
--- a/packages/run-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/run-protocol/src/vaultFactory/vaultManager.js
@@ -127,9 +127,33 @@ const trace = makeTracer('VM');
  * }>} MethodContext
  */
 
-// FIXME https://github.com/Agoric/agoric-sdk/issues/5622
-let liquidationQueueing = false;
-let outstandingQuote = null;
+// #region Ephemeral states
+/** @typedef {{
+ * liquidationQueueing: boolean,
+ * outstandingQuote: Promise<MutableQuote>?,
+ * }} Ephemera */
+/** @type {() => Ephemera} */
+const initEphemera = () => ({
+  liquidationQueueing: false,
+  outstandingQuote: null,
+});
+/** @type {Map<VaultManager, Ephemera>} */
+const ephemeras = new Map();
+/**
+ * Provide an object to hold state that need not (or cannot) be durable.
+ *
+ * @type {(key: VaultManager) => Ephemera}
+ */
+const provideEphemera = key => {
+  if (ephemeras.has(key)) {
+    // @ts-expect-error cast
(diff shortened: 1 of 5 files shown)


packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -127,9 +131,15 @@ const trace = makeTracer('VM');
  * }>} MethodContext
  */
 
-// FIXME https://github.com/Agoric/agoric-sdk/issues/5622
-let liquidationQueueing = false;
-let outstandingQuote = null;
+/** @typedef {{
```
Chris-Hibbert: I think this is the right place for a comment saying that these values are ephemeral, and per-vaultManager, and why that's the case.

Chris-Hibbert (approved):

PR #5748, opened 2022-07-11
fix(swingstore): clarify promise handling

### LINKED ISSUE #5729: fix swing-store "unawaited promise" warnings

My CI runs are reporting two Promises that aren't being `await`ed in swing-store:

https://github.com/Agoric/agoric-sdk/blob/d07a5025ab4fe5ef9f6008fbb8150e6e5c50154b/packages/swing-store/src/swingStore.js#L273

https://github.com/Agoric/agoric-sdk/blob/d07a5025ab4fe5ef9f6008fbb8150e6e5c50154b/packages/swing-store/src/swingStore.js#L288

If I'm reading this right, our new LMDB library returns a Promise for the write operations `db.put()` and `db.remove()`, and our swing-store layer is dropping that Promise.

@mhofman You helped us make the shift to `lmdb-js`, could you take a look at this? It might be might be trivial or it might be serious: does this API mean that we're supposed to wait for the Promise to resolve for something? I'd expect the `put` to just add the change to a txn buffer (synchronously), and then it'd be `db.commit()` that returns a Promise which doesn't resolve until everything was safely flushed.

If the put/remove Promise means something else, it might be safe to slap a `void ` on it, to silence the warning.

## Security Considerations

I think it's unlikely, but if the `db.put()` Promise somehow needs to be honored beyond/in-addition-to what we do in `commit()`, this could cause DB corruption if there's a race somehow.

## Test Plan

None, just look at CI results to see that the warning goes away.

mhofman:
closes: #5729

## Description

Captured in code comment why it's safe to ignore the promise result of the `put` and `remove` LMDB operations [within a transaction](https://github.com/kriszyp/lmdb-js#dbtransactioncallback-function-promise):
> Any `put` or `remove` operations are immediately written to the transaction and can be immediately read afterwards (you can call `get()` or `getRange()` without awaiting for a returned promise) in the transaction.

Drive by fix of the nested await warning

### Security Considerations

None

This PR content serves as documentation

### Testing Considerations

CI

### DIFF
--- a/packages/swing-store/src/swingStore.js
+++ b/packages/swing-store/src/swingStore.js
@@ -270,7 +270,9 @@ function makeSwingStore(dirPath, forceReset, options) {
   function set(key, value) {
     assert.typeof(key, 'string');
     assert.typeof(value, 'string');
-    ensureTxn().put(key, value);
+    // synchronous read after write within a transaction is safe
+    // The transaction's overall success will be awaited during commit
+    void ensureTxn().put(key, value);
     trace('set', key, value);
   }
 
@@ -285,7 +287,9 @@ function makeSwingStore(dirPath, forceReset, options) {
   function del(key) {
     assert.typeof(key, 'string');
     if (has(key)) {
-      ensureTxn().remove(key);
+      // synchronous read after write within a transaction is safe
+      // The transaction's overall success will be awaited during commit
+      void ensureTxn().remove(key);
       trace('del', key);
     }
   }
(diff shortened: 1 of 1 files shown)


warner (approved):
looks good

warner (approved):

warner:
Once https://github.com/Agoric/agoric-sdk/pull/5784 lands I'll re-queue this, the unrelated CI failures are blocking the merge.

PR #5749, opened 2022-07-11
feature: give contracts the ability to pause invitations

Chris-Hibbert:
refs: #5036

## Description

Give contracts the ability to pause invitations

This is the basic capability. The remainder of the task is making it
possible for governance to manage it.

### Security Considerations

Without more and better controls, this could make it possible for contract authors to wedge their contracts. Fortunately, the next step is adding some support in contractGovernance so this power is harder to abuse or misuse.

This should be added to Zoe documentation, though it's not a compatibility issue.

### Testing Considerations

Tested matching patterns, non-matching patterns, and no patterns. Tested that a matched pattern prevented exercising an offer.


Chris-Hibbert:
@turadg   In local testing, `test-priceAggregator.js` (oracle invitation) failed with **Promise returned by test never resolved**. I'd appreciate any help you can offer on narrowing this down. 

I've done some logging and single stepping, but that hasn't made me feel closer to identifying the problem

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -14,6 +14,8 @@ import { handlePKitWarning } from '../handleWarning.js';
 
 const { details: X, quote: q } = assert;
 
+const isASCII = /^[a-zA-Z0-9_]+$/;
```
turadg: ASCII has more than these symbols. 
```suggestion
const isAlphaNumeric = /^[a-zA-Z0-9_]+$/;
```

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -61,6 +63,18 @@ export const makeStartInstance = (
 
     const instance = makeInstanceHandle();
 
+    const snoozedStrings = [];
+    const isSnoozed = string =>
+      snoozedStrings.some(s => string.indexOf(s) > -1);
+    const updateSnoozedList = strings => {
+      assert(
+        strings.every(s => typeof s === 'string' && isASCII.test(s)),
+        X`Snoozed strings (${q(strings)}) must be non-empty ascii strings.`,
```
turadg:
```suggestion
        X`Snoozed strings (${q(strings)}) must be non-empty alphanumeric strings.`,
```

packages/zoe/src/zoeService/offer/offer.js
```diff
@@ -34,15 +35,25 @@ export const makeOfferMethod = (
...
+    const { instance, description } = await query.getInvitationDetails(
       invitation,
     );
     // AWAIT ///
 
-    const instanceAdmin = getInstanceAdmin(instanceHandle);
+    const instanceAdmin = getInstanceAdmin(instance);
+    assert(
+      !instanceAdmin.isSnoozed(description),
+      `not accepting offer with description '${description}'`,
```
turadg: this reads like the particular offer is being rejected and another might be accepted.

shouldn't it use details/quote?
```suggestion
      X`contract is not currently accepting offers. ${q(description)}`,
```

packages/zoe/src/internal-types.js
```diff
@@ -81,6 +81,7 @@
  * @property {(completion: Completion) => void} exitAllSeats
  * @property {ShutdownWithFailure} failAllSeats
  * @property {() => void} stopAcceptingOffers
+ * @property {(string: string) => boolean} isSnoozed
```
turadg: "snoozed" to me evokes "pause notifications". When you snooze an alarm, the time keeps ticking but it doesn't make sounds.

This is different in that notifications still come out but no offers are _accepted_. I like metaphors but here it might be better to be perfectly explicit: `isAcceptingOffers` (the inverse). It parallels `stopAcceptingOffers`.

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -61,6 +63,18 @@ export const makeStartInstance = (
...
+      );
+
+      snoozedStrings.splice(0, snoozedStrings.length, ...strings);
```
turadg: Looks like this is to replace the contents of the array. I'm unfamiliar with this idiom. Is it common in the repo?

What's the advantage vs `let snoozedStrings` above and `snoozedStrings = strings` here? I guess mutability. But we could type the array as `Readonly<string[]>` to give confidence it won't be modified.

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -61,6 +63,18 @@ export const makeStartInstance = (
 
     const instance = makeInstanceHandle();
 
+    const snoozedStrings = [];
+    const isSnoozed = string =>
+      snoozedStrings.some(s => string.indexOf(s) > -1);
```
turadg: not clear to me why any substring match is sufficient. please explain in jsdoc comment.

packages/zoe/src/zoeService/offer/offer.js
Chris-Hibbert: It is correct that it is accepting some and rejecting others.  If the pattern list covers all invitation patterns, it might reject all of them. This code doesn't check to see if the pattern list is empty, so it doesn't know that anything is being rejected, unless the current invitation is blocked.

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: Good point

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: done.

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: Done

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: > Looks like this is to replace the contents of the array. I'm unfamiliar with this idiom. Is it common in the repo?

I haven't seen it before; I found it (and other suggestions) on stackoverflow.

The advantage is immutability. `const` is enforced, while typescript is overridable. (and in fact, we'd override it in order to set the value.)

I'm open to other suggestions for maintaining a list of suppressed values.  I don't think we need regexps. I don't think we need an API that provides a lot of support. My expectation is that most contracts will leave it to governance, and that governance will have fairly rough tools, as it's only expected to be used in emergencies.

turadg: > The advantage is immutability. const is enforced

Well sorta. The immutability of the binding is enforced, but it's an array and this line replaces the contents. So it's a full mutation. I'd say the `const` is more deceiving than helpful. It's a hoop to jump through to avoid rebinding the value.

I think it would be more direct to `let snoozedStrings`. The reader can tell at that point that it'll be reassigned. And it has the added benefit of making it possible to express that the array itself should not be mutated, with `Readonly<>`. Yes it can be overridden but at least the intent is expressed. And there's no loss to material runtime mutability.

packages/zoe/src/zoeService/offer/offer.js
turadg: ah, I misread `updateSnoozedList`. That method name could be improved.

For one, isn't it setting the list? There's no history or partial update; it's a full replacement.

For two, I maintain that "snooze" is too metaphorical for the concept of "stop accepting offers whose descriptions contain one of these strings". (Incidentally, that could make a good JSdoc for the method, if I got it right)

Three, "list" is vague. List of what? Not snoozes.  Consider "selectors". Or since it's about what to reject, "filter"

Putting it all together, maybe `setOfferFilter`. Where the filter is a list of substrings that can match.

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: done.

packages/zoe/src/internal-types.js
Chris-Hibbert: good suggestion.

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -63,19 +63,24 @@ export const makeStartInstance = (
...
+      const proposedStrings = Array.from(strings);
       assert(
-        strings.every(s => typeof s === 'string' && isAlphanumeric.test(s)),
+        proposedStrings.every(
+          s => typeof s === 'string' && isAlphanumeric.test(s),
+        ),
         X`Snoozed strings (${q(
           strings,
-        )}) must be non-empty alpha-numeric strings.`,
+        )}) must be a (possibly empty) Array of alpha-numeric strings.`,
```
turadg: suffices to say "must be an Array of alpha-numeric strings"

packages/zoe/src/zoeService/offer/offer.js
Chris-Hibbert: done

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: I want to point out that that's how you reset the list.

turadg: pointing out how to reset the list is good. I'd expect to see that in the function jsdoc.  I'd probably want to know that before I send an invalid argument :)

packages/zoe/src/internal-types.js
```diff
@@ -81,6 +81,7 @@
  * @property {(completion: Completion) => void} exitAllSeats
  * @property {ShutdownWithFailure} failAllSeats
  * @property {() => void} stopAcceptingOffers
+ * @property {(string: string) => boolean} isBlocked
```
turadg, line 83: 👍

packages/zoe/src/zoeService/offer/offer.js
```diff
@@ -34,15 +35,25 @@ export const makeOfferMethod = (
...
+    const { instance, description } = await query.getInvitationDetails(
       invitation,
     );
     // AWAIT ///
 
-    const instanceAdmin = getInstanceAdmin(instanceHandle);
+    const instanceAdmin = getInstanceAdmin(instance);
+    assert(
+      !instanceAdmin.isBlocked(description),
+      `not accepting offer with description ${q(description)}`,
```
turadg: does this need `details`?

turadg (approved):

packages/zoe/src/zoeService/offer/offer.js
Chris-Hibbert: Yes, thanks.  I added a clause to my pre-submit  pattern, and found a second case.  Doh!

Chris-Hibbert:
@erights   I'm waiting for a review from you.

packages/zoe/src/zoeService/offer/offer.js
```diff
@@ -34,15 +35,25 @@ export const makeOfferMethod = (
...
 
-    const instanceAdmin = getInstanceAdmin(instanceHandle);
+    const instanceAdmin = getInstanceAdmin(instance);
+    assert(
+      !instanceAdmin.isBlocked(description),
+      X`not accepting offer with description ${q(description)}`,
+    );
     instanceAdmin.assertAcceptingOffers();
 
+    const { invitationHandle } = await burnInvitation(
```
erights: In the case where the offer is not blocked, I would not have expected this PR to cause an observable difference. However, the old logic first burns the offer before asserting that the contract is still accepting offers. The new one asserts that the contract is still accepting offers before it burns. Arguably the new behavior is better, in which case no change suggested. But I would like to understand the implications of making this observable change before we proceed.

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -61,6 +63,28 @@ export const makeStartInstance = (
...
+      assert(
+        proposedStrings.every(
+          s => typeof s === 'string' && isAlphanumeric.test(s),
+        ),
+        X`Blocked strings (${q(
+          strings,
+        )}) must be a (possibly empty) Array of alpha-numeric strings.`,
+      );
+
+      offerFilterStrings = proposedStrings;
```
erights: The original is ok. I'll explain why I suggest each of the changes below, but take only what you agree is an improvement.

```suggestion
    let offerFilterStrings = harden([]);
    // if any string in the list matches, don't process the invitation
    const isBlocked = string =>
      offerFilterStrings.some(s => string.indexOf(s) > -1);

    const setOfferFilter = strings => {
      harden(strings);
      fit(strings, M.arrayOf(M.string()));
      assert(
        strings.every(s => isAlphanumeric.test(s)),
        X`Blocked strings (${q(
          strings,
        )}) must be a (possibly empty) Array of alpha-numeric strings.`,
      );

      offerFilterStrings = strings;
```

I see a declaration of `ReadOnly` and code consistent with the array actually being immutable. The original copies the incoming array into a new array which it then never mutates. So it seems stronger than ReadOnly anyway.

But TS is unsound. If you want to ensure that the array isn't mutated, you can harden it. Fortunately, a hardened array of strings is a CopyArray, and so a Passable, and so something we can pattern match on. The pattern match above asserts in one line what took several lines to say. Unfortunately, we cannot express the alphanumeric constraint in the pattern language so that still needs to be explicit.

A possible reason to prefer your code over my suggested pattern match: Better diagnostics. My pattern matching logic doesn't understand the purpose of the match, and so cannot give diagnostics that are as informative.

Once we know that `strings` is a CopyArray of strings, we know it is safe enough that we don't need to copy it into `proposedStrings`. OTOH, this hardens whatever `strings` argument was passed in by our caller, which we might want to avoid. In that case, we can start with
```js
const proposedStrings = harden([...strings]);
```
and then do everything else with `proposedStrings` more like your original. Note that `[...foo]` is essentially the same as `Array.from(foo)` but IMO reads nicer.

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -61,6 +63,28 @@ export const makeStartInstance = (
 
     const instance = makeInstanceHandle();
 
+    // Invitations whose descriptions contain any of the strings will be blocked
+    /** @type {Readonly<string[]>} */
+    let offerFilterStrings = [];
+    // if any string in the list matches, don't process the invitation
+    const isBlocked = string =>
+      offerFilterStrings.some(s => string.indexOf(s) > -1);
```
erights: This is the part I don't get: Why substring rather than exact match? Do you have a use case in mind?

packages/zoe/src/zoeService/startInstance.js
erights: I did not change the code for `isBlocked` above because I discuss the substring match criteria separately.

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -11,7 +11,6 @@ import { Far } from '@endo/marshal';
 import { makeIssuerKit, AmountMath } from '@agoric/ertp';
```
erights, line 11: What are the changes to this file about? They seem completely unrelated to the rest of this PR.

packages/zoe/test/unitTests/test-blockedOffers.js
```diff
@@ -0,0 +1,104 @@
...
+
+  await t.throwsAsync(
+    () =>
+      makeOffer(
+        zoe,
+        zcf,
+        harden({ want: { A: moola(3n) }, give: { B: bucks(5n) } }),
+        harden({ B: bucksMint.mintPayment(bucks(5n)) }),
+      ),
+    { message: 'not accepting offer with description "seat"' },
```
erights: That setting the offer filter to `['eat']` blocks invitations with description `'seat'` is disturbing. It is clearly accident prone. For the scenarios off the top of my head, exact match is fine. Why do you want substring? Do you have a use case in mind?

(I seem to have a vague memory that we may have discussed this and I agreed with substring. But I don't quite remember and I am confused now.)

erights (changes requested):
I see that you make `setOfferFilter` available on `zcf` and thereby callable by the contract without requiring any governance decision. This raises the invitation legibility question. Tell me if the following is the answer:

In order for the contract to call `zcf.setOfferFilter(['exerciseOption'])` it would either need to say this statically where Fred could see it, or engage in some obscure code that Fred would never normally see in a contract that prevents Fred from statically determining that this isn't called?

This does not technically violate invitation legibility, but does add to Fred's work in a way that's hard to explain. Are you sure you want to do this?

packages/zoe/src/zoeService/offer/offer.js
Chris-Hibbert: I misunderstood the semantics of `assertAcceptingOffers()`. Since Zoe does not stop accepting offers and then restart, it's legitimate to burn the offer before checking. If we're no longer accepting offers, then the invitation won't be useable later.

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: My mental model was a big complex contract. If the author can design the contract so some subset of invitations can be disabled as a group with a single string, that makes it easier to manage.

My theory was that regExps are expensive, and we wouldn't want contracts specifying patterns for Zoe to execute. 

My usual advice would be to specify full strings that match particular invitations, and to be very careful if they're thinking about using short strings. With most contracts, the author should be able to quickly look at all the invitation descriptions. 

I'm open to other approaches, but if a contract has 10 invitation descriptions (or more), it wouldn't be a great API to require them to list all of them in order to stop the contract.

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: Rather than side-effect the caller-provided string, I used your suggestion for `harden([...strings])`. 

I do prefer the better error message, so I bundled the tests together into `isAlphaNumeric()`;

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
Chris-Hibbert: I think this is a consequence of adding ` await query.getInvitationDetails()` to the top of offer(). I don't see an alternative to doing that.

If pausing weren't such a crucial feature for protecting against runaway contracts in an emergency, I would be sympathetic to an argument that the required extra `await` is too high a price to pay for this feature.

packages/zoe/test/unitTests/test-blockedOffers.js
Chris-Hibbert: short strings in the block list are a bad idea, unless the goal is to block everything.  I talked about my reasoning for substrings in `startInstance.js`.

I'll change the test to not use a short string, since I agree that it's a bad example.

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
erights: We discussed. It is just extra intra-vat turns. No extra cranks, blocks, or network round trips. Should be assumed to have negligible performance implications until there is evidence otherwise. Resolving.

packages/zoe/src/zoeService/startInstance.js
```diff
@@ -61,6 +61,28 @@ export const makeStartInstance = (
 
     const instance = makeInstanceHandle();
 
+    // Invitations whose descriptions contain any of the strings will be blocked
```
erights:
```suggestion
    // Invitations whose descriptions match any of the strings will be blocked
```

packages/zoe/src/zoeService/startInstance.js
erights: Chris and I discussed. No special constraints on these strings are needed, and in any case would need to be coordinated with similar constraints on the descriptions strings, which are equally unneeded.

Gone. Resolving.

packages/zoe/src/zoeService/startInstance.js
erights: substring match is gone. Resolving.

packages/zoe/src/zoeService/startInstance.js
erights: Restrictions on the string contents are gone. Resolving.

packages/zoe/src/zoeService/offer/offer.js
erights: I am sooo looking forward to a lint checking rule that catches these missing `X`s

packages/zoe/src/zoeService/offer/offer.js
erights: Original order restored. Resolving.

packages/zoe/src/zoeService/startInstance.js
erights: Current state looks good. Resolving.

erights (approved):
LGTM!

packages/zoe/src/zoeService/startInstance.js
Chris-Hibbert: yep. that's better.

PR #5750, opened 2022-07-12
feat(wallet-connection): Connect dapp directly to wallet UI

samsiegart:

We can now add the `useLocalStorage={true}` property to the wallet connection component to get the dapp to talk directly to the wallet UI in local storage, that way the dapp doesn't directly depend on an ag-solo connection.

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -126,10 +126,7 @@ Lifetime: until we decide not to support Prometheus for metrics export
 
 Affects: solo
 
-Until dApps are converted to connect to the smart-wallet UI directly,
-this allows them to continue to connect to `/wallet-bridge.html` and such
-on the solo and have these endpoints serviced by `/wallet/bridge.html`
-and such in a wallet UI.
+This enables a proxy so that the solo bridge interface (/wallet-bridge.html) is backed by the smart wallet (/wallet/bridge.html). Dapps designed for the solo bridge can enable this until they connect to the smart wallet directly.
 
 ```
 BRIDGE_TARGET=http://localhost:3001 make BASE_PORT=8002 scenario3-run
--- a/docs/env.md
+++ b/docs/env.md
@@ -122,6 +122,21 @@ for the Prometheus scrape endpoint to export telemetry.
 
 Lifetime: until we decide not to support Prometheus for metrics export
 
+## SOLO_BRIDGE_TARGET
+
+Affects: solo
+
+Until dApps are converted to connect to the smart-wallet UI directly,
+this allows them to continue to connect to `/wallet-bridge.html` and such
+on the solo and have these endpoints serviced by `/wallet/bridge.html`
+and such in a wallet UI.
+
+```
+BRIDGE_TARGET=http://localhost:3001 make BASE_PORT=8002 scenario3-run
+```
+
+Lifetime: smart wallet transition period
+
 ## SOLO_LMDB_MAP_SIZE
 
 Affects: solo
--- a/packages/solo/public/wallet-bridge.html
+++ b/packages/solo/public/wallet-bridge.html
@@ -154,11 +154,13 @@ command-line request to open the Agoric wallet.`);
         });
       };
 
-      // Start the flow of messages.
-      if (window.parent !== window) {
-        window.parent.postMessage({ type: 'walletBridgeLoaded' }, '*');
-      }
-      retryWebSocket();
+      window.addEventListener('load', () => {
+        // Start the flow of messages.
+        if (window.parent !== window) {
+          window.parent.postMessage({ type: 'walletBridgeLoaded' }, '*');
+        }
+        retryWebSocket();
+      });
     </script>
   </body>
 </html>
--- a/packages/solo/public/wallet-bridge.html
+++ b/packages/solo/public/wallet-bridge.html
@@ -154,6 +154,9 @@ command-line request to open the Agoric wallet.`);
         });
       };
 
+      // This ensures the message wont be posted until after the iframe's
+      // "onLoad" event fires so we can rely on the consistent ordering of
+      // events in the WalletConnection component.
       window.addEventListener('load', () => {
         // Start the flow of messages.
         if (window.parent !== window) {
--- a/packages/solo/src/web.js
+++ b/packages/solo/src/web.js
@@ -1,9 +1,11 @@
-/* global setTimeout clearTimeout setInterval clearInterval */
+/* global setTimeout clearTimeout setInterval clearInterval process */
 // Start a network service
 import path from 'path';
 import http from 'http';
 import { createConnection } from 'net';
+import { existsSync as existsSyncAmbient } from 'fs';
 import express from 'express';
+import { createProxyMiddleware } from 'http-proxy-middleware';
 import WebSocket from 'ws';
 import anylogger from 'anylogger';
 import morgan from 'morgan';
@@ -71,6 +73,7 @@ export async function makeHTTPListener(
   walletHtmlDir = '',
   validateAndInstallBundle,
   connections,
+  { env = process.env, existsSync = existsSyncAmbient } = {},
 ) {
   // Enrich the inbound command with some metadata.
   const inboundCommand = (
@@ -126,13 +129,36 @@ export async function makeHTTPListener(
   app.use(express.json({ limit: maximumBundleSize })); // parse application/json
   const server = http.createServer(app);
 
+  // Proxy to another wallet bridge
+  const { SOLO_BRIDGE_TARGET: bridgeTarget } = env;
+  if (bridgeTarget) {
+    app.use(
+      ['/wallet-bridge.html', '/wallet'],
+      createProxyMiddleware({
+        target: bridgeTarget,
+        pathRewrite: {
+          '^/wallet-bridge.html': '/wallet/bridge.html',
+        },
+        // changeOrigin: true,
+      }),
+    );
+  }
+
   // serve the static HTML for the UI
   const htmldir = path.join(basedir, 'html');
   log(`Serving static files from ${htmldir}`);
   app.use(express.static(htmldir));
   app.use(express.static(new URL('../public', import.meta.url).pathname));
 
-  if (walletHtmlDir) {
+  if (walletHtmlDir && !bridgeTarget) {
+    // Transition localStorage based bridge
+    if (existsSync(path.join(walletHtmlDir, 'bridge.html'))) {
+      console.log('redirecting wallet bridge');
+      app.get('/wallet-bridge.html', (req, res) =>
+        res.redirect('/wallet/bridge.html'),
+      );
+    }
+
     // Serve the wallet directory.
     app.use('/wallet', express.static(walletHtmlDir));
 
--- a/packages/wallet/ui/config-overrides/index.js
+++ b/packages/wallet/ui/config-overrides/index.js
@@ -1,5 +1,52 @@
+// cribbed from https://www.npmjs.com/package/react-app-rewire-multiple-entry
+/* global __dirname, require, module */
+
+const path = require('path');
+const HtmlWebPackPlugin = require('html-webpack-plugin');
+const multiEntry = require('react-app-rewire-multiple-entry');
+
+const bridgeTemplate = path.resolve(
+  path.join(__dirname, '..', 'public', 'bridge.html'),
+);
+
+const multipleEntry = multiEntry([
+  {
+    template: 'public/index.html',
+    entry: 'src/index.jsx',
+  },
+  {
+    template: 'public/bridge.html',
+    entry: 'src/bridge.jsx',
+  },
+]);
+
 module.exports = function override(config, _env) {
   config.resolve.fallback = { path: false, crypto: false };
   config.ignoreWarnings = [/Failed to parse source map/];
+
+  const htmlWebpackPlugin = config.plugins.find(
+    plugin => plugin.constructor.name === 'HtmlWebpackPlugin',
+  );
+  if (!htmlWebpackPlugin) {
+    throw new Error("Can't find HtmlWebpackPlugin");
+  }
+
+  multipleEntry.addMultiEntry(config);
+
+  const bridgeKeys = Object.keys(config.entry).filter(k =>
+    k.startsWith('bridge'),
+  );
+  const opts = {
+    ...htmlWebpackPlugin.userOptions,
+    template: bridgeTemplate,
+    filename: './bridge.html',
+    chunks: bridgeKeys,
+  };
+  htmlWebpackPlugin.userOptions = {
+    ...htmlWebpackPlugin.userOptions,
+    excludeChunks: bridgeKeys,
+  };
+  const plug2 = new HtmlWebPackPlugin(opts);
+  config.plugins.push(plug2);
   return config;
 };
--- a/packages/wallet/ui/public/bridge.html
+++ b/packages/wallet/ui/public/bridge.html
@@ -8,7 +8,7 @@
 </head>
 
 <body style="overflow-x: hidden; overflow-y: scroll">
-    <noscript>Without JavaScript enabled, this page cannot function.</noscript>
+    <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
 </body>
 
--- a/packages/wallet/ui/public/bridge.html
+++ b/packages/wallet/ui/public/bridge.html
@@ -4,34 +4,12 @@
 <head>
     <meta charset="utf-8" />
     <script src="%PUBLIC_URL%/lockdown.umd.js"></script>
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <meta name="theme-color" content="#000000" />
-    <meta name="description" content="Agoric wallet" />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
-    <link href="https://fonts.googleapis.com/css?family=Montserrat" rel="stylesheet" />
-    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
-    <!--
-      manifest.json provides metadata used when your web app is installed on a
-      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
-    -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-    <title>Agoric Wallet Bridge!@@@</title>
+    <title>Agoric Wallet Bridge</title>
 </head>
 
 <body style="overflow-x: hidden; overflow-y: scroll">
-    <noscript>This app isn't much use without JavaScript.</noscript>
+    <noscript>Without JavaScript enabled, this page cannot function.</noscript>
     <div id="root"></div>
-    <!--
-      This HTML file is a template.
-      If you open it directly in the browser, you will see an empty page.
-
-      You can add webfonts, meta tags, or analytics to this file.
-      The build step will place the bundled scripts into the <body> tag.
-
-      To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    -->
 </body>
 
 </html>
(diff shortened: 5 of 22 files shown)


packages/solo/public/wallet-bridge.html
```diff
@@ -154,11 +154,13 @@
         });
       };
 
-      // Start the flow of messages.
-      if (window.parent !== window) {
-        window.parent.postMessage({ type: 'walletBridgeLoaded' }, '*');
-      }
-      retryWebSocket();
+      window.addEventListener('load', () => {
```
samsiegart, line 160: This ensures the message wont be posted until after the iframe's "onLoad" event fires. This is the way it works with the react bridge.jsx, so we can rely on the consistent ordering of events in the WalletConnection component.

dckc (commented):
seems pretty good, but given the low test coverage for this stuff, I'd like to give @turadg a chance to look it over

docs/env.md
```diff
@@ -122,6 +122,21 @@ for the Prometheus scrape endpoint to export telemetry.
...
 Lifetime: until we decide not to support Prometheus for metrics export
 
+## SOLO_BRIDGE_TARGET
+
+Affects: solo
+
+Until dApps are converted to connect to the smart-wallet UI directly,
+this allows them to continue to connect to `/wallet-bridge.html` and such
+on the solo and have these endpoints serviced by `/wallet/bridge.html`
+and such in a wallet UI.
```
turadg: Wasn't clear to me from reading this what the purpose was (the difference between the two paths or how it worked.)

Consider:
> This enables a proxy so that the solo bridge interface (`/wallet-bridge.html`) is backed by the smart wallet `(/wallet/bridge.html)`. Dapps designed for the solo bridge can enable this until they connect to the smart wallet directly.

packages/solo/public/wallet-bridge.html
turadg, line 160: your PR comment would make a good code comment.

```suggestion
      // This ensures the message wont be posted until after the iframe's "onLoad" event fires
      // so we can rely on the consistent ordering of events in the WalletConnection component.
      window.addEventListener('load', () => {
```

packages/solo/src/web.js
```diff
@@ -1,9 +1,11 @@
-/* global setTimeout clearTimeout setInterval clearInterval */
+/* global setTimeout clearTimeout setInterval clearInterval process */
 // Start a network service
 import path from 'path';
 import http from 'http';
 import { createConnection } from 'net';
+import { existsSync as existsSyncAmbient } from 'fs';
```
turadg, line 6: ambient doesn't seem quite right since it is an import. consider `existsSyncDefault` or `fsExistsSync`

packages/wallet/ui/config-overrides/index.js
```diff
@@ -1,5 +1,52 @@
+// cribbed from https://www.npmjs.com/package/react-app-rewire-multiple-entry
```
turadg, line 1: 👍

packages/wallet/ui/public/bridge.html
```diff
@@ -0,0 +1,15 @@
...
+<html lang="en">
+
+<head>
+    <meta charset="utf-8" />
+    <script src="%PUBLIC_URL%/lockdown.umd.js"></script>
+    <title>Agoric Wallet Bridge</title>
+</head>
+
+<body style="overflow-x: hidden; overflow-y: scroll">
+    <noscript>Without JavaScript enabled, this page cannot function.</noscript>
```
turadg: that statement is true invariantly. minor but let's tell the user something about the state.
```suggestion
    <noscript>ERROR: JavaScript not available.</noscript>
```

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
+// @ts-check
+/* eslint-disable import/no-extraneous-dependencies */
+import '@endo/eventual-send/shim';
+import './lockdown.js';
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+Error.stackTraceLimit = Infinity;
```
turadg, line 9: TIL.

How do we know this is necessary?

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+/** ISSUE: where are these defined? */
```
turadg, line 11: ISSUE like file an issue?

In master they're magic strings. I see two posts and one receive that's just a validation:
https://github.com/Agoric/agoric-sdk/blob/7e16cc01cba1ad789cc3cb79ca1f76ad1199118e/packages/solo/public/wallet-bridge.html#L159
https://github.com/Agoric/agoric-sdk/blob/7e16cc01cba1ad789cc3cb79ca1f76ad1199118e/packages/solo/public/wallet-bridge.html#L115
https://github.com/Agoric/agoric-sdk/blob/7e16cc01cba1ad789cc3cb79ca1f76ad1199118e/packages/web-components/src/bridge-iframe-connector.js#L20-L27

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+ *   addEventListener: typeof window.addEventListener,
+ *   parentPost: typeof window.postMessage,
+ *   t0: number,
+ * }} io
+ */
+const installDappConnection = ({ addEventListener, parentPost, t0 }) => {
+  /** @type { string } */
+  let origin;
+
+  /** @type {(key: string, value: string) => void} */
```
turadg, line 41: I'm surprised ts-check isn't alerting that `setItem` is undefined at this point. 

Oh, there's no type checking in this package yet. https://github.com/Agoric/agoric-sdk/pull/5757

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+  let setItem;
+  /** @type {string[]} */
+  const buffer = [];
+
+  console.debug('bridge: installDappConnection');
+
+  parentPost({ type: BridgeProtocol.loaded }, '*');
+  parentPost({ type: BridgeProtocol.opened }, '*');
```
turadg, line 49: no distinction anymore?

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+  /** @param {Record<string, any>} payload */
+
+  const { stringify, parse } = JSON;
+
+  addEventListener('message', ev => {
+    // console.debug('bridge: handling message:', ev.data);
+    if (
+      !ev.data ||
+      typeof ev.data.type !== 'string' ||
+      !ev.data.type.startsWith('CTP_')
```
turadg:
```suggestion
      !(ev.data?.type?.startsWith('CTP_')
```
If `ev.data.type` is defined and not a string, that's an error to fail on.

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+    ) {
+      return;
+    }
+
+    if (origin === undefined) {
+      // First-come, first-serve.
+      // console.debug('bridge: setting origin to', origin);
```
turadg: uncomment or remove. `debug` level is not rendered by default so fine to use liberally

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+      console.debug('bridge: message from dapp->buffer', origin, ev.data);
+      buffer.push(harden(ev.data));
+    }
+  });
+
+  return harden({
+    /** @param {typeof window.localStorage} storage */
+    connectStorage: storage => {
+      addEventListener('storage', ev => {
+        // console.debug('from storage', origin, ev.key, ev.newValue);
```
turadg: uncomment or remove

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+const sameParent = () => {
+  const me = window;
+  const { parent } = window;
+  if (me === parent) {
+    // eslint-disable-next-line no-debugger
+    debugger;
+    throw Error('window.parent === parent!!!');
+  }
+  return false;
```
turadg: this return value is a little silly because it's the only value that'll ever return.

consider `checkWindow: () => void` that throws.

either way please document why it's necessary. why does it have to be checked on every post?

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+const signalBridge = () => {
+  if ('requestStorageAccess' in document) {
+    document
+      .requestStorageAccess()
+      .then(() => {
+        console.debug('bridge: storage access granted');
+        conn.connectStorage(window.localStorage);
+      })
+      .catch(whyNot =>
+        console.error('bridge: requestStorageAccess rejected with', whyNot),
```
turadg, line 122: what's the UX in this case? worth an alert?

packages/wallet/ui/src/bridge.jsx
```diff
@@ -0,0 +1,144 @@
...
+      );
+  } else {
+    console.debug('bridge: SKIP document.requestStorageAccess');
+    conn.connectStorage(window.localStorage);
```
turadg, line 126: could this fail?

packages/wallet/ui/src/components/WalletConnection.jsx
```diff
@@ -28,7 +30,17 @@ const WalletConnection = ({
   connectionConfig,
 }) => {
   const classes = useStyles();
-  const [wc, setWC] = useState(null);
+  /**
+   * ISSUE: where to get the full types for these?
```
turadg: e.g. for WalletConnection at this line to the bottom of `AgoricWalletConnection.js`,
```
/** @typedef {InstanceType<AgoricWalletConnection>['walletConnection']} FarWalletConnection */
```
I suggest "Far" to distinguish from the other "WalletConnection" strings.

packages/wallet/ui/src/components/WalletConnection.jsx
```diff
@@ -67,8 +79,70 @@ const WalletConnection = ({
       },
     }).catch(rethrowIfNotCancelled);
 
+    /** @type {Map<string,[ReturnType<typeof makeCapTP>, number]>} */
+    const dappToConn = new Map();
+    const {
+      localStorage: storage,
+      addEventListener,
+      removeEventListener,
+    } = window; // WARNING: ambient
```
turadg: consider using `window.` at each access so the ambience is apparent at the callsite

packages/wallet/ui/src/components/WalletConnection.jsx
```diff
@@ -67,8 +79,70 @@ const WalletConnection = ({
...
+
+    function handleStorageMessage(key, newValue) {
```
turadg: a lot of logic in here that's not UI concern. consider extracting some to a utility function. that could be unit tested.

packages/wallet/ui/src/lockdown.js
```diff
@@ -0,0 +1,25 @@
...
+// eslint-disable-next-line no-constant-condition
+const consoleTaming = '%NODE_ENV%' === 'production' ? 'safe' : 'unsafe';
+
+// eslint-disable-next-line no-restricted-properties
+const { pow: mathPow } = Math;
+// eslint-disable-next-line no-restricted-properties
+Math.pow = (base, exp) =>
+  typeof base === 'bigint' && typeof exp === 'bigint'
+    ? base ** exp
+    : mathPow(base, exp);
```
turadg, line 12: do we not have a shim that can be called to do this?

packages/solo/src/web.js
dckc: It's ambient in a different sense than ambient types, but it's still ambient authority:
> All primitives for interacting with the external world are embodied by primitive objects and anything globally accessible is immutable data. There is no open(filename) function in the global namespace, nor can such a function be imported -- converting a string to file access is much like converting an integer to a pointer. -- https://github.com/Agoric/agoric-sdk/wiki/OCap-Discipline

packages/wallet/ui/src/lockdown.js
dckc: Umm... This _is_ such a shim, no?

turadg: I mean one that can be imported here.

dckc: I'm pretty sure this is a one-off. It's kind of a kludge, really.

dckc: It's not a new kludge, btw. It comes from...

https://github.com/Agoric/agoric-sdk/blob/39fe09285a230a33966b23f3ab3a1f61127f0a64/packages/wallet/ui/public/index.html#L12-L15

Git says that's from @samsiegart Oct 2021 01a91181e36 but I suspect it predates that too.

packages/wallet/ui/src/components/WalletConnection.jsx
dckc: I'd rather go the other direction: pass in these parts of `window` explicitly rather than using `window` ambiently.

OTOH, `window.addEventListener` _is_ quite likely more conventional and easier to read, and since the current code doesn't follow ocap discipline, maybe making it easier to read is better.

packages/wallet/ui/public/bridge.html
dckc: Fair enough. Better to be explicit and tell them how to fix it?

I have a hard time blaming the user when the page requires JavaScript. I was trying to concisely say something like "Please excuse the requirement for you to put yourself at risk by enabling JavaScript in your browser. This page uses JavaScript for more than progressive enhancement. Providing similar functionality without JavaScript is not practical."

packages/solo/public/wallet-bridge.html
dckc: yes, please!

packages/wallet/ui/src/bridge.jsx
dckc: I don't know. I looked around for clues without much luck. It does merit a comment, at least to point to some discussion of why we do this. Help, @kriskowal ? @erights ?

packages/wallet/ui/src/bridge.jsx
dckc: Issue like: decide between us, whether to sweep it under the rug, address it, or file an issue. Yes, the magic strings pre-date this PR. Is it time to fix them?

packages/wallet/ui/src/lockdown.js
turadg: Ah. Where is it used? I can't find any instances. If there are any consider making them bigint-safe there instead of modifying `Math`.

packages/wallet/ui/src/bridge.jsx
dckc: I don't know. It feels a bit magic, to me. In `wallet-bridge.html`, `opened` seems to go later in the process... analogous to when the user clicks the button... but I tried it that way without luck. Perhaps Sam's recent work addressed that?

packages/wallet/ui/src/bridge.jsx
dckc: as to why: @michaelfig wrote Jun 28:
> it’s just to prevent infinite regress when the bridge is loaded in its own web page.

Given that, the test is probably applied at the wrong place. It probably belongs when we first wire up the top level event handlers, not on each message.

packages/wallet/ui/src/bridge.jsx
dckc: No, I don't think so. It only calls `addEventListener`; I don't see how that could fail.

packages/wallet/ui/src/lockdown.js
dckc: I don't know where it's used. Maybe some 3rd party library?

I sure wish we had made a test.

packages/wallet/ui/src/bridge.jsx
turadg: I've been bitten by `localStorage` permission problems, but I suppose if you've got the object you won't find out about any errors until mutating. And even then you could have failures later with hitting quotas, the permission being revoked, etc. So fine to handle those cases later.

packages/wallet/ui/src/components/WalletConnection.jsx
dckc: concur

packages/wallet/ui/src/bridge.jsx
turadg: looks to me that this PR solves the problem already. the question "where are these defined?" is answered and they're no longer magic.

dckc: the other occurrences don't refer to `BridgeProtocol`. And I think there are one or two other such strings. And this `BridgeProtocol` doesn't provide much of a definition: what are the valid state transitions? What do the states mean?

turadg: ah, I thought this PR replaced the occurrences I linked to.

ok, unresolved issue. can be punted imo. this whole package needs a types cleanup

packages/wallet/ui/public/bridge.html
samsiegart: Just used the same message as index.html. Do people still use the web without javascript? At the very least they're trying to use a dapp so I don't think we need to be apologetic.

packages/solo/public/wallet-bridge.html
samsiegart: Done

docs/env.md
samsiegart: Done

packages/wallet/ui/src/bridge.jsx
samsiegart: We can leave this convo open but I won't block this PR on it since it's preexisting code.

packages/wallet/ui/src/bridge.jsx
samsiegart: Hmm, walletBridgeOpened doesn't seem to be read anywhere. Maybe useful for debugging, but only walletBridgeLoaded has any logic dependent on it.

packages/wallet/ui/src/bridge.jsx
samsiegart: Done

packages/wallet/ui/src/bridge.jsx
samsiegart: Removed

packages/wallet/ui/src/bridge.jsx
samsiegart: Removed

packages/wallet/ui/src/bridge.jsx
samsiegart: Done

packages/wallet/ui/src/bridge.jsx
samsiegart: I'm actually not sure, `requestStorageAccess` isn't in chrome.

packages/wallet/ui/src/lockdown.js
samsiegart: Yes this code has been here longer than I have, will leave this convo open but won't block this PR on it.

packages/wallet/ui/src/components/WalletConnection.jsx
samsiegart: Switched to `window.`

packages/wallet/ui/src/components/WalletConnection.jsx
samsiegart: Pulled this into a separate function. I don't know how much of this code is just placeholder/PoC until after we have the smart wallet bridge fully working, so I won't bother writing tests yet.

packages/wallet/ui/src/components/WalletConnection.jsx
samsiegart: I couldn't get this working

packages/wallet/ui/public/bridge.html
dckc: > ... so I don't think we need to be apologetic.

No, we don't need to, but...

> Do people still use the web without javascript?

Yes. In fact, I bet you do it regularly, with `curl` and the like.

dckc (approved):
It's at least a step in the right direction.

turadg (approved):
:shipit:

PR #5753, opened 2022-07-12
Publish Reserve econ data: mint & burn

### LINKED ISSUE #4651: Expose key Reserve Contract economic data via subscription

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose from the Reserve Contract:

1. How much of each asset type
2. Governance parameters

We had initially thought the Reward Pool would be housed here, but @Chris-Hibbert wrote: "The collateral reserve contract holds funds from various sources. Its main goal is holding assets that governance decisions can direct to the AMM or other uses. It is unlikely to be the main repository for reward funds."

Chris-Hibbert:
closes: #4651

## Description

Publish key economic data (minting and burning) from the Reserve contract.

### Security Considerations

publishing data that should be public.  No security concerns.

None.

### Testing Considerations

Tested that the values were published.

### DIFF
--- a/packages/run-protocol/src/reserve/assetReserve.js
+++ b/packages/run-protocol/src/reserve/assetReserve.js
@@ -25,8 +25,8 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
  *
  * @property {AmountKeywordRecord} allocations
  * @property {Amount<'nat'>} shortfallBalance shortfall from liquiditation that
- * @property {Amount<'nat'>} mintedRUN total RUN minted to date
- * @property {Amount<'nat'>} burnedRUN total RUN burned to date
+ * @property {Amount<'nat'>} totalFeeMinted total RUN minted to date
+ * @property {Amount<'nat'>} totalFeeBurned total RUN burned to date
  *   has not yet been compensated.
  */
 
@@ -224,8 +224,8 @@ const start = async (zcf, privateArgs) => {
 
   const { brand: runBrand } = await E(runMint).getIssuerRecord();
 
-  let mintedRUN = AmountMath.makeEmpty(runBrand);
-  let burnedRUN = AmountMath.makeEmpty(runBrand);
+  let totalFeeMinted = AmountMath.makeEmpty(runBrand);
+  let totalFeeBurned = AmountMath.makeEmpty(runBrand);
 
   // shortfall in Vaults due to liquidations less than debt. This value can be
   // reduced by various actions which burn RUN.
@@ -241,8 +241,8 @@ const start = async (zcf, privateArgs) => {
     const metrics = harden({
       allocations: getAllocations(),
       shortfallBalance,
-      mintedRUN,
-      burnedRUN,
+      totalFeeMinted,
+      totalFeeBurned,
     });
     metricsPublication.updateState(metrics);
   };
@@ -277,7 +277,7 @@ const start = async (zcf, privateArgs) => {
 
     // create the RUN
     const offerToSeat = runMint.mintGains(harden({ RUN: runAmount }));
-    mintedRUN = AmountMath.add(mintedRUN, runAmount);
+    totalFeeMinted = AmountMath.add(totalFeeMinted, runAmount);
 
     offerToSeat.incrementBy(
       collateralSeat.decrementBy(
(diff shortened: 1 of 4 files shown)


packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -25,6 +25,8 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
  *
  * @property {AmountKeywordRecord} allocations
  * @property {Amount<'nat'>} shortfallBalance shortfall from liquiditation that
+ * @property {Amount<'nat'>} mintedRUN total RUN minted to date
+ * @property {Amount<'nat'>} burnedRUN total RUN burned to date
```
turadg: let's use `Fee` instead of `RUN` per https://github.com/Agoric/agoric-sdk/issues/4800

and per https://github.com/Agoric/agoric-sdk/blob/fb4b156bdd74e28af8fb93f2da3f43c1041c1cdf/packages/run-protocol/src/vaultFactory/vaultManager.js#L46 `minted` should be second. 

"total" helps convey that this is "minted _ever_" not any other time window.

```suggestion
 * @property {Amount<'nat'>} totalFeeMinted total Fee minted to date
 * @property {Amount<'nat'>} totalFeeBurned total Fee burned to date
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -222,6 +224,9 @@ const start = async (zcf, privateArgs) => {
 
   const { brand: runBrand } = await E(runMint).getIssuerRecord();
 
+  let mintedRUN = AmountMath.makeEmpty(runBrand);
+  let burnedRUN = AmountMath.makeEmpty(runBrand);
```
turadg: just noting the variable names should continue to match the metric names

packages/run-protocol/test/metrics.js
```diff
@@ -102,3 +102,9 @@ export const vaultManagerMetricsTracker = async (t, publicFacet) => {
     assertFullyLiquidated,
   });
 };
+export const reserveInitialState = emptyRun => ({
+  allocations: {},
+  shortfallBalance: emptyRun,
+  burnedRUN: emptyRun,
+  mintedRUN: emptyRun,
+});
```
turadg: this is valid only for `test-reserve.js` so please keep it there

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -316,6 +325,7 @@ const start = async (zcf, privateArgs) => {
       }),
     );
     zcf.reallocate(offerToSeat, collateralSeat);
+    updateMetrics();
```
turadg, line 328: why is this needed here but not after `burnedRUN` change?

turadg (changes requested):

packages/run-protocol/test/metrics.js
Chris-Hibbert: I hoisted it when the vault test needed it as well. I didn't want the vault test to have to be updated if we add more to the record.  I didn't see a better place to put it, and `metrics.js` is at least under `test`.

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: Done

packages/run-protocol/src/reserve/assetReserve.js

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: It is already called in `burnRUNToReduceShortfall`, because that calls `increaseLiquidationShortfall`. `addLiquiditytoAmmPool` mints, but doesn't adjust the shortfall. The shortfall is increased by other contracts that get a `shortfallReportingInvitation`.

packages/run-protocol/test/metrics.js
turadg: ah, my mistake that it's used only in `test-reserve.js`.

turadg (approved):

PR #5755, opened 2022-07-12
Require that durable object kind shape be conserved across upgrades

### LINKED ISSUE #5662: durable kinds must upgrade to the same number/names of facets

While walking through ZCF/contract upgrade with @erights and @Chris-Hibbert , I realized that v1 of a vat/contract might define a durable kind with facets `[aliceFacet, bobFacet]`, and then a subsequent v2 might use the same (stored) `KindHandle` to associate new behavior that used *different* facet names, or a different number of facets.

They originally thought that this should be fine, even desirable: being able to add new facets would be a convenient form of upgrade.

But after walking through how facet vrefs are managed, I think I convinced them that we cannot accomodate changes to the shape of a cohort record (a MultiKind) during upgrade:

* we map facet names to integers (using `getOwnPropertyNames().sort()` on the behavior record), and then use those integers:
  * as the `facet` component of a vref: `o+3/2:1` means "kind o+3, instance 2, facet 1"
  * as an index into the string stored in the export-status record: a three-facet cohort might have a record of `rsn`, where the first facet has status `r` (reachable by the kernel), the second facet has status `s` (recognizable aka "seen" by the kernel), and the third facet `n` is not currently exported to the kernel at all

We record the list of `facetNames` in RAM (`virtualReferences.js > rememberFacetNames()`), but we do not store it durably in the DB. As a result, the next version might call `defineKindMulti()` with a behavior record that provides a different set of facets. Bad things that would result:

* existing vrefs could get reassigned to a different facet of the same cohort
  * if exported, a client who had access to `bobFacet` might wind up talking to `aliceFacet`
* those vrefs might wind up pointing to no facet at all (if the number of facets was reduced), probably causing a crash during message delivery, or virtual-data deserialization
* the export status of existing vrefs would probably get confused

## Description of the Design

The immediate task is to make sure this does not happen. We need to record the `facetNames` durably, when we define the Kind for the first time. In later versions, when re-defining the Kind, we must assert that the new behavior record is parsed into exactly the same set of facet names.

In the future, we can think about how to relax this, by allowing the creation of new facet names. I think we must insist that the old facets remain in place (same names, same order), but if we could figure out how to allow the export status records to be open-ended (treat short strings as if they had an implicit suffix of `n`s ?), then we could probably tolerate the addition of new ones. Note that this is at odds with our current practice of sorting the facet names; we'd have to make up some rule like "sort all the new ones, then append the new ones to the stored old list".

## Security Considerations

Vat upgrade gets to change the behavior of a lot of stuff, and the new Kind record is in complete control of how the old exported vrefs behave, but that's no reason to allow the upgrader to cause this sort of chaos.

## Test Plan

Unit tests that do an upgrade with different facets (all three cases: more facets, fewer facets, same number but differently named facets).

cc @FUDCo

FUDCo:
Fixes #5662

### DIFF
--- a/packages/SwingSet/src/liveslots/vatstore-usage.md
+++ b/packages/SwingSet/src/liveslots/vatstore-usage.md
@@ -98,7 +98,9 @@ The KindHandle is a durable virtual object of a special internal Kind. This is t
 
 For each virtual object kind that is defined we store a metadata record for purposes of scanning directly through the defined kinds when a vat is stopped or upgraded.  For durable kinds this record is stored in `vom.dkind.${kindID}`; for non-durable kinds it is stored in `vom.vkind.${kindID}`.  Currently this metadata takes the form of a JSON-serialized record `{ kindID, tag }`, where the `kindID` property is the kind ID (redundantly) and `tag` is the tag string as provided in the `defineKind` or `makeKindHandle` call.
 
-Durable kinds need to store their `nextInstanceID` in the DB, so subsequent versions can begin allocating new instances from a non-overlapping starting point. For durable kinds, the metadata record is `{ kindID, tag, nextInstanceID }`, and is updated after every durable-object allocation.
+Durable kinds need to store their `nextInstanceID` in the DB, so that subsequent versions can begin allocating new instances from a non-overlapping starting point. For durable kinds, the metadata record is initially `{ kindID, tag, nextInstanceID }`, and the `nextInstance` property is updated after every durable-object allocation.
+
+In addition, durable kinds also need to keep track of whether they are single- or multi-faceted and, if the latter, what the names of the facets are. This information is required so that kind definitions in upgraded versions of the containing vat are maintained in a backwards compatible manner over time.  When a kind becomes defined as a single-faceted kind, a property `unfaceted` (with the value `true`) will be added to the descriptor record.  For a multi-faceted kind, the record will be given a property `facets` whose value is an array of the facet names, in the same order as the assignment of facet indices within the cohort record.
 
 # Virtual/Durable Collections (aka Stores)
 
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -491,6 +491,16 @@ export function makeVirtualObjectManager(
   }
 
   /**
+   * @typedef {{
+   *  kindID: string,
+   *  nextInstanceID: number,
+   *  tag: string,
+   *  unfaceted?: boolean,
+   *  facets?: string[],
+   * }} DurableKindDescriptor
+   */
+
+  /**
    * Define a new kind of virtual object.
    *
    * @param {string} kindID  The kind ID to associate with the new kind.
@@ -512,8 +522,11 @@ export function makeVirtualObjectManager(
    *    optional finisher function that can perform post-creation initialization
    *    operations, such as inserting the new object in a cyclical object graph.
    *
-   * @param {boolean} durable  A flag indicating whether or not the newly defined
+   * @param {boolean} isDurable  A flag indicating whether or not the newly defined
    *    kind should be a durable kind.
+   *
+   * @param {DurableKindDescriptor} [durableKindDescriptor]  Descriptor for the
+   *    durable kind, if it is, in fact, durable
    *
    * @returns {*} a maker function that can be called to manufacture new
    *    instances of this kind of object.  The parameters of the maker function
@@ -586,7 +599,8 @@ export function makeVirtualObjectManager(
     multifaceted,
     behavior,
     options = {},
-    durable,
+    isDurable,
+    durableKindDescriptor,
   ) {
     const { finish } = options;
     let facetNames;
@@ -596,7 +610,7 @@ export function makeVirtualObjectManager(
     switch (facetiousness) {
       case 'one': {
         assert(!multifaceted);
-        facetNames = null;
+        facetNames = undefined;
         behaviorTemplate = copyMethods(behavior);
         break;
       }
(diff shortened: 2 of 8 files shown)


packages/SwingSet/src/liveslots/vatstore-usage.md
```diff
@@ -98,7 +98,9 @@ The KindHandle is a durable virtual object of a special internal Kind. This is t
...
+Durable kinds need to store their `nextInstanceID` in the DB, so that subsequent versions can begin allocating new instances from a non-overlapping starting point. For durable kinds, the metadata record is initially `{ kindID, tag, nextInstanceID }`, and the `nextInstance` property is updated after every durable-object allocation.
+
+In addition, durable kinds also need to keep track of whether they are single- or multi-faceted and, if the latter, what the names of the facets are. This information is required so that kind definitions in upgraded versions of the containing vat are maintained in a backwards compatible manner over time.  When a kind becomes defined as a single-faceted kind, a property `unfaceted` (with the value `true`) will be added to the descriptor record.  For a multi-faceted kind, the record will be given a property `facets` whose value is an array of the facet names, in the same order as the assignment of facet indices within the cohort record.
```
warner, line 103: thanks for the docs updates!

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -894,14 +906,29 @@ export function makeVirtualObjectManager(
...
     assert(
       !definedDurableKinds.has(kindID),
       `redefinition of durable kind "${tag}"`,
     );
+    assert(
+      !unfaceted,
+      `durable kind "${tag}" originally defined as single-faceted`,
+    );
     nextInstanceIDs.set(kindID, nextInstanceID);
+    const definedFacets = Object.getOwnPropertyNames(behavior).sort();
```
warner: I'm a tiny bit nervous about having two copies of this facet-extraction/ordering line (and what happens if this one diverges from the other copy in `defineKindInternal()`).

What do you think about passing the descriptor into `defineKindInternal`, and letting `defineKindInternal` do both the comparison/assert and the subsequent `saveDurableKindDescriptor` (if `durable` and it was updated with facet info)?

In that approach, the descriptor argument is basically describing the facet names that have been committed to, so a future version of `defineKindInternal` has the option to append new facet names to that list, and it's in complete control of the ordering it picks.

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -420,6 +420,100 @@ test('failed upgrade - lost kind', async t => {
...
+    's2mFacetiousnessMismatch',
+    `durable kind "multi" originally defined as single-faceted`,
+  );
+});
+
+test('facet kind redefinition - fail on multi- to single-facet redefinition', async t => {
+  await testMultiKindUpgradeChecks(
+    t,
+    'm2sFacetiousnessMismatch',
+    `durable kind "multi" originally defined as multi-faceted`,
```
warner, line 587: that comment is surprising (I'd expect `durable kind 'single' originally defined as multi-faceted`), but I imagine it's because the `tag` is "multi" even when `mode` says to create the v1 kind as unfaceted

packages/SwingSet/test/upgrade/vat-ulrik-1.js
```diff
@@ -167,7 +168,78 @@ export const buildRootObject = (_vatPowers, vatParameters, baggage) => {
...
+              barMethod: () => 1,
+            },
+            belch: {
+              belchMethod: () => 1,
+            },
+          });
+          break;
+        }
+        case 'facetOrderMismatch': {
+          // upgrade should succeed
```
warner: maybe augment the comment, `upgrade should succeed because we sort() facet names`

packages/SwingSet/test/upgrade/vat-ulrik-1.js
```diff
@@ -167,7 +168,78 @@ export const buildRootObject = (_vatPowers, vatParameters, baggage) => {
...
+    },
+
+    makeMultiKind: mode => {
+      const mkh = makeKindHandle('multi');
+      baggage.init('mkh', mkh);
+      switch (mode) {
+        case 's2mFacetiousnessMismatch': {
+          // upgrade should fail
+          defineDurableKind(mkh, () => ({}), {
+            fooMethod: () => 1,
```
warner, line 187: nit: since we aren't looking at method names at all, this test could be a lot shorter to read if every facet had the same method, and they were all defined ahead of time in a single `const facet = { method: () => 1 };`. Also, `const emptyInit = () => ({});`. Then e.g. `facetCountMismatch` would squish into:
```js
  defineDurableKindMulti(mkh, emptyInit, {
    foo: facet,
    bar: facet,
  });
```

also nit: can we use `handle` for the local variable name instead of `mkh`?

packages/SwingSet/test/upgrade/vat-ulrik-2.js
```diff
@@ -46,6 +67,6 @@ export const buildRootObject = (_vatPowers, vatParameters, baggage) => {
     },
     getNewDurandal: () => newDur,
   });
-  // exercise async return
+  // exercise async return // XXX WHY???
```
warner: `buildRootObject()` is now allowed to return a Promise, as long as it fulfills promptly. We needed this to allow ZCF to do an `importBundle(contractBundle)` during upgrade, and `importBundle` returns a prompt-Promise. This line ensures that we can return such a Promise without errors: it was easier and faster than writing a whole new test for that one feature. But it sounds like I should add a comment to that effect here.

Could you replace my terse comment with the following?:
```
// buildRootObject() is allowed to return a Promise, as
// long as it fulfills promptly (we added this in #5246 to allow ZCF
// to use the async importBundle() during contract upgrade).
// We return a pre-fulfilled Promise here to exercise this ability.
```

warner (approved):
A couple of suggested changes, but I think I'd be ok if it landed without them.

Please coordinate with https://github.com/Agoric/agoric-sdk/pull/5730 (which is awaiting final review, hopefully later today), as it changes some code in `test-upgrade.js` around the reporting of failures during upgrade. I don't think it will require any deep integration/merge work, but the error message might change slightly, and the comment about "failure reporting not implemented" can probably go away.

packages/SwingSet/src/liveslots/virtualObjectManager.js
FUDCo: Yeah, I had the same misgiving, but I couldn't see a graceful way to deal with it that wasn't worse.  However, I've now stared at the code for a few minutes after not having looked at it for a day and I think I see a way to handle this more cleanly.

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: Yup, you pegged it.

PR #5765, opened 2022-07-14
feat(vats): bootstrap.produceItem(), consumeItem, resetItem

### LINKED ISSUE #5740: bootstrap method to stash object/device refs for use by CORE_EVAL

As @FUDCo and I were try to think through upgrade scenarios, it occurred to us that we might some day need to add a new device to the running (chain-side) kernel. Maybe a "timer v2" with new features, or some new sort of storage thing.

We can imagine adding code to the kernel to many it possible to create a new device post-initialization, and some cosmic-swingset code that triggers at a particular block height to call some `controller` APIs. This code could synthesize the kref for the new root device node, and then use `controller.queueToVatRoot()` to send it into the bootstrap vat.

But we'll need something in the bootstrap vat to receive this reference (as a vref). We already have the CORE_EVAL "Big JS Hammer" upgrade mechanism, which is a bridge-device message that supplies JS code to be evaluated in a powerful context. But CORE_EVAL only gets a string (and the bridge device can only provide a string). So we need a way for the CORE_EVAL code to access vrefs that we invent and deliver from elsewhere.

## Description of the Design

I'm thinking we add an extra method to our `vat-bootstrap` which just takes its arguments and stashes them in a local `let` binding, then make sure this binding is in scope for the CORE_EVAL evaluation. Alternatively, create a `Map` or something, have the new method populate/`set()` things in it, and include the Map in the endowments of CORE_EVAL.

Then, if/when we invent something that must be delivered as krefs, our upgrade process does:
* create new kref
* `controller.queueToVatRoot('bootstrap', 'stashThing', capdata(newKref))`
* wait for that to be delivered
* use the Big JS Hammer to evaluate code which references the stash and uses the new facilities somehow (maybe distribute it to some new vats or something)

## Security Considerations

The Big JS Hammer is already super-powerful over the authorities that it is born with, but this would let us add new authorities (object refs) that come from even more powerful code (outside the kernel, with access to `controller` and whatever new kernel tools we invent later).

## Test Plan

A unit test that grabs a kref with `pinVatRoot()` on some secondary vat, then does `controller.queueToKref()` to the bootstrap vat, then uses the Big JS Hammer to retrieve the secondary ref from the stash and make sure it looks good.

cc @dckc @FUDCo @erights  (I don't know what Label to put on this: what do we use for code in `packages/vats` ?)

dckc:
closes: #5740

## Description

Allow the bootstrap vat holders such as the kernel to manipulate the
items passed to CORE_EVAL.

### Security Considerations

`consumeItem` allows the kernel access to existing items up to and including the IST mint. Perhaps it's not strictly needed, but I struggled to come up with a test without it.

?

### Testing Considerations

Ideally we would actually test the interaction with CORE_EVAL, but I don't see a straightforward way to arrange it.

### DIFF
--- a/packages/vats/src/core/boot.js
+++ b/packages/vats/src/core/boot.js
@@ -137,6 +137,18 @@ const buildRootObject = (vatPowers, vatParameters) => {
         console.error('BOOTSTRAP FAILED:', e);
         throw e;
       }),
+    consumeItem: name => {
+      assert.typeof(name, 'string');
+      return consume[name];
+    },
+    produceItem: (name, resolution) => {
+      assert.typeof(name, 'string');
+      produce[name].resolve(resolution);
+    },
+    resetItem: name => {
+      assert.typeof(name, 'string');
+      produce[name].reset();
+    },
   });
 };
 
--- a/packages/vats/test/test-boot.js
+++ b/packages/vats/test/test-boot.js
@@ -211,3 +211,17 @@ test('evaluateInstallation is available to core eval', async t => {
   // @ts-expect-error
   t.deepEqual(typeof actual.extract, 'function');
 });
+
+test('bootstrap provides a way to pass items to CORE_EVAL', async t => {
+  const root = buildRootObject(
+    // @ts-expect-error Device<T> is a little goofy
+    { D: d => d, logger: t.log },
+    { argv: argvByRole.chain, governanceActions: false },
+  );
+
+  await E(root).produceItem('swissArmyKnife', [1, 2, 3]);
+  t.deepEqual(await E(root).consumeItem('swissArmyKnife'), [1, 2, 3]);
+  await E(root).resetItem('swissArmyKnife');
+  await E(root).produceItem('swissArmyKnife', 4);
+  t.deepEqual(await E(root).consumeItem('swissArmyKnife'), 4);
+});


warner:
I'm thinking about a `makeVatController()` -style test in `packages/vats/` , which would build a whole kernel (with all the core bootstrap vats, but stub device implementations that aren't wired to real IO). We could then `controller.queueToVatRoot()` a message to this new facility, followed by a simulated bridge message that exercises the Big JS Hammer (I'm assuming the Hammer uses the bridge device), and then maybe a `produceItem()` call to retrieve some value to make sure it worked.

I'll poke around the directory to see if there's anything already suitable for adaptation. After review, we can land this without a decent test, but I'll feel a whole lot more confident about it being an escape hatch if we actually make sure the hinges work.

dckc:
That's quite a bit more than 3 lines:)

warner:
I've only just skimmed how this produce/consume scheme works. Am I reading it correctly that `produce[name]` or `consume[name]` will cause a new PromiseKit to come into being, and stick around forever, so that nobody needs to explicitly allocate a new `name` for use as a rendezvous point? And so it doesn't matter which order the `produce[name].resolve(data)` and `consume[name].then(..)` calls happen? That's awesome, I didn't realize the Proxy which does that is compatible with `harden()`.. I'll have to learn more.

Then let's see, our hypothetical upgrade scenario requires two things to happen. One is a `controller.queueToVatRoot()` that does the equivalent of `E(bootstrap).produceItem('name', newStuff)`, where `newStuff` contains interesting authorities like a new Timer device, or some new kernel service that we invented just for the upgrade. This call must come from cosmic-swingset or some new swingset controller/kernel code, where the new kernel thing is created. And it must be made at a specific consensus point. That probably means we want code in cosmos/cosmic-swingset to notice that a specific block height has been reached, and at that one moment, invoke this new function (before calling `controller.run()`). That will enqueue the `produceItem()` call onto the run queue, where it will eventually be delivered to the bootstrap vat and make the item available.

The second thing that needs to happen is some code getting evaluated where it has access to `consume`, so it can receive this new authority and do something interesting with it. Currently the only pathway to that is the Big JS Hammer through the `bridgeManager`. That bridge pathway is reachable by a governance proposal that contains the JS code to be evaluated

If we don't add a second `eval` pathway, then this sort of upgrade will look like:
* governance proposal 1 says "use the Big JS Hammer to evaluate this code that waits for `consume.newThing` to resolve and then uses it"
* governance proposal 2 says "upgrade to new software that include the new kernel thing and has a blockheight-triggered call to `controller.queueToVatRoot()` to deliver it

The `makePromiseSpace` scheme means that the order in which these two things happen does not matter. If the Big JS Hammer arrives first, the consuming code will be evaluated but will wait until the `newThing` is available. If the kernel upgrade happens first, the `produce.newThing` call will happen first, and the new value will wait around until the hammer evaluates code that uses it.

Obviously we really only want to do one proposal, not two. I'm wondering if we should add another method here, to allow `bridgeCoreEval` to be accessed with another `controller.queueToVatRoot` call, in addition to the bridge device. If we had that, then the single governance proposal would upgrade the validator to software that has a blockheight-triggered call into some cosmic-swingset/kernel code, which does all of:
* create the new device/etc
* use `controller.queueToVatRoot(bootstrap, 'produceItem', newthings)`
* use `controller.queueToVatRoot(bootstrap, 'coreEval', bundleWhichUsesNewThings)`

This new pathway is just as powerful as the Big JS Hammer, but it's triggered by cosmic-swingset code rather than bridge device writes. So it's only possible to invoke by new code (triggered by some block height) that lives in the JS side of cosmic-swingset. The bridge-based pathway can be triggered by the Golang code, at least by whatever portions of it have access to the bridge.

warner:
> That's quite a bit more than 3 lines:)

Fair point, I'll think about such a test in a different ticket.

dckc:
> Am I reading it correctly ...? That's awesome

Yup!

Glad the "3 line" version made so much of this clear to you.

> Obviously we really only want to do one proposal, not two.

The swingset core-eval payload is a *list* of evaluations to run. So if "upgrade to new software..." fits in the coreEval structure, one proposal can cover both. I think.

> I'll think about such a test in a different ticket.

I prefer not to defer test till after a feature is landed. Let's decide what level of testing this merits and land it with this feature, please. Some timeboxed pairing might be fun.

packages/vats/src/core/boot.js
```diff
@@ -137,6 +137,18 @@ const buildRootObject = (vatPowers, vatParameters) => {
         console.error('BOOTSTRAP FAILED:', e);
         throw e;
       }),
+    consumeItem: name => {
```
warner, line 140: To make this usable past-boot, we need to keep the root object alive (otherwise it'll get GCed, because nobody else outside this vat is holding a reference). 

If liveslots and Endo `importBundle` are doing everything correctly, then the only way to guarantee that the root object is kept alive is for cosmic-swingset to do a `controller.pinVatRoot('bootstrap')` during kernel initialization. I'll make a separate ticket to accomplish that.

(adding a top-level `let root` declaration, and changing `return Far('root', ..)` to `root = Far(..); return root;`, would mean the module namespace object keeps the root object alive, but liveslots shouldn't be retaining a reference to that namespace object: it calls `root = vatNS.buildRootObject()` and then drops the namespace, and `root` only stays alive in `slotToVal` until the kernel sends a `dispatch.dropExport`)

warner (approved):
Looks good, I'll add a `cosmic-swingset` ticket to pin the vat root, and I'll think about how to write an integration test (which should fail without the pin).

Your clever produce/consume mechanism made this really elegant!

warner:
Ok, so a bigger test should load `decentral-core-config.json`, call `initializeSwingSet()` and `makeSwingsetController` to construct a whole kernel using the same vats as the chain would use, let `controller.run()` finish executing the bootstrap work, then submit a `produceItem` with some static data to `controller.queueToVatRoot()`, another `c.run()`, then do a `consumeItem` to fetch it back again. The general pattern can be cribbed from a swingset test like https://github.com/Agoric/agoric-sdk/blob/1dd584b195212705b1f74a8c89b7f3f121640e41/packages/SwingSet/test/timer/test-timer.js

PR #5766, opened 2022-07-14
feat(notifier): makeStoredPublishKit

### LINKED ISSUE #5739: add storage for `makePublishKit`

We have `makePublishKit` but need something that stores data off-chain. Once we have that then #5386 can return its`subscriber` instead of a Notifier. Consumers that need a Notifier can pass that to `makeNotifierFromSubscriber` from #5737.

## Description of the Design

The `storageNode` and `marshaller` arguments are generally optional so we can continue that in this case. Since they are optional we can add them to the params of `makePublishKit` to optionally tee to storage.

Alternately we could have `makeStoredPublishKit`.

## Security Considerations

Will tee storage to `storageNode`. `marshaller` is powerful.

## Test Plan

Tests with storage and tests integrating with `makeNotifierFromSubscriber`.

turadg:
closes: #5739

## Description

Pulled out of https://github.com/Agoric/agoric-sdk/pull/5704/ and improved test coverage.

### Security Considerations

--

--
### Testing Considerations

New tests

### DIFF
--- a/packages/notifier/src/storesub.js
+++ b/packages/notifier/src/storesub.js
@@ -2,9 +2,78 @@
 import { E } from '@endo/eventual-send';
 import { Far, makeMarshal } from '@endo/marshal';
 import { observeIteration } from './asyncIterableAdaptor.js';
+import { makePublishKit, subscribeEach } from './publish-kit.js';
 import { makeSubscriptionKit } from './subscriber.js';
 
 /**
+ * @template T
+ * @param {Subscriber<T>} subscriber
+ * @param {(v: T) => void} consumeValue
+ */
+const forEachPublicationRecord = async (subscriber, consumeValue) => {
+  const iteratorP = E(subscribeEach(subscriber))[Symbol.asyncIterator]();
+
+  let finished = false;
+  while (!finished) {
+    // Allow nested awaits (in loop) because it's safe for each to run in a turn
+    // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
+    const { value, done } = await E(iteratorP).next();
+    // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
+    await consumeValue(value);
+    finished = !!done;
+  }
+};
+
+/**
+ * Begin iterating the source, storing serialized iteration values.  If the
+ * storageNode's `setValue` operation rejects, no further writes to it will
+ * be attempted (but results will remain available from the subscriber).
+ *
+ * Returns a StoredSubscriber that can be used by a client to directly follow
+ * the iteration themselves, or obtain information to subscribe to the stored
+ * data out-of-band.
+ *
+ * @template T
+ * @param {Subscriber<T>} subscriber
+ * @param {ERef<StorageNode>} storageNode
+ * @param {ERef<ReturnType<typeof makeMarshal>>} marshaller
+ * @returns {StoredSubscriber<T>}
+ */
+export const makeStoredSubscriber = (subscriber, storageNode, marshaller) => {
+  assert(subscriber && storageNode && marshaller, 'missing argument');
+
+  const storeValue = value =>
+    E(marshaller)
+      .serialize(value)
+      .then(serialized => {
+        const encoded = JSON.stringify(serialized);
+        return E(storageNode).setValue(encoded);
+      });
+
+  // Start publishing the source.
+  forEachPublicationRecord(subscriber, storeValue).catch(err => {
+    // TODO: How should we handle and/or surface this failure?
+    // https://github.com/Agoric/agoric-sdk/pull/5766#discussion_r922498088
+    console.error('StoredSubscriber failed to iterate', err);
+  });
+
+  /** @type {Unserializer} */
+  const unserializer = Far('unserializer', {
+    unserialize: E(marshaller).unserialize,
+  });
+
+  /** @type {StoredSubscriber<T>} */
+  const storesub = Far('StoredSubscriber', {
+    ...subscriber,
+    getStoreKey: E(storageNode).getStoreKey,
+    getUnserializer: () => unserializer,
+  });
+  return storesub;
+};
+
+/**
+ * @deprecated use makeStoredSubscriber
+ *
  * Begin iterating the source, storing serialized iteration values.  If the
  * storageNode's `setValue` operation rejects, the iteration will be terminated.
  *
@@ -85,13 +154,16 @@ export const makeStoredSubscription = (
 harden(makeStoredSubscription);
 
 /**
+ * @deprecated use StoredPublishKit
  * @template T
  * @typedef {object} StoredPublisherKit
- * @property {StoredSubscription<T>} subscriber TODO: change to StoredSubscriber<T> when available
+ * @property {StoredSubscription<T>} subscriber
  * @property {IterationObserver<T>} publisher
  */
 
 /**
+ * @deprecated use makeStoredPublishKit
+ *
  * @template [T=unknown]
  * @param {ERef<StorageNode>} [storageNode]
  * @param {ERef<Marshaller>} [marshaller]
@@ -117,3 +189,26 @@ export const makeStoredPublisherKit = (storageNode, marshaller, childPath) => {
     subscriber,
   };
 };
+
+/**
+ * Like makePublishKit this makes a `{ publisher, subscriber }` pair for doing efficient
+ * distributed pub/sub supporting both "each" and "latest" iteration
+ * of published values.
+ *
+ * What's different is `subscriber` tees records, writing out to storageNode.
+ *
+ * @template [T=unknown]
+ * @param {ERef<StorageNode>} storageNode
+ * @param {ERef<Marshaller>} marshaller
+ * @returns {StoredPublishKit<T>}
+ */
+export const makeStoredPublishKit = (storageNode, marshaller) => {
+  const { publisher, subscriber } = makePublishKit();
+
+  return {
+    publisher,
+    // wrap the subscriber to tee events to storage
+    subscriber: makeStoredSubscriber(subscriber, storageNode, marshaller),
+  };
+};
+harden(makeStoredPublishKit);
--- a/packages/notifier/test/test-stored-subscription.js
+++ b/packages/notifier/test/test-stored-subscription.js
@@ -1,32 +1,60 @@
 // @ts-check
+/* global setImmediate */
 
 // eslint-disable-next-line import/order
 import { test } from './prepare-test-env-ava.js';
 
 import { E } from '@endo/eventual-send';
-import { Far, makeMarshal } from '@endo/marshal';
-import { makeSubscriptionKit, makeStoredSubscription } from '../src/index.js';
+import { makeMarshal } from '@endo/marshal';
+import {
+  makePublishKit,
+  makeStoredPublishKit,
+  makeStoredSubscriber,
+  makeStoredSubscription,
+  makeSubscriptionKit,
+  subscribeEach,
+} from '../src/index.js';
 
 import '../src/types.js';
 import { jsonPairs } from './marshal-corpus.js';
 
+export const eventLoopIteration = async () =>
+  new Promise(resolve => setImmediate(resolve));
+
+/**
+ *
+ * @param {string} path
+ * @param {IterationObserver<unknown>} [publication]
+ */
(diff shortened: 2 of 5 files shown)


dckc (commented):
I skimmed it and I don't see any problems.

But I don't understand the code well enough to maintain it, so I'm inclined to defer to @gibson042 to approve, trusting that he knows this stuff well enough to speak for other designers of the notifier package (MarkM, Tribble, Warner, ...).


IIUC, this or related work that it's based on is motivated by a performance problem involving holding on to unbounded linked lists of promises or some such. I'm interested to know whether we have tests that demonstrate that we have addressed that performance problem.

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,76 @@
...
+ * storageNode's `setValue` operation rejects, the iteration will be terminated.
+ *
+ * Returns a StoredSubscriber that can be used by a client to directly follow
+ * the iteration themselves, or obtain information to subscribe to the stored
+ * data out-of-band.
+ *
+ * @template T
+ * @param {Subscriber<T>} subscriber
+ * @param {ERef<StorageNode>} storageNode
+ * @param {ERef<ReturnType<typeof makeMarshal>>} marshaller
```
gibson042, line 39: Why are only two of the three arguments allowed to be promises?
```suggestion
 * @param {ERef<Subscriber<T>>} subscriber
 * @param {ERef<StorageNode>} storageNode
 * @param {ERef<ReturnType<typeof makeMarshal>>} marshaller
```

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,76 @@
...
+  /** @type {Unserializer} */
+  const unserializer = Far('unserializer', {
+    unserialize: E(marshaller).unserialize,
+  });
+
+  /** @type {StoredSubscriber<T>} */
+  const storesub = Far('StoredSubscriber', {
+    ...subscriber,
+    getStoreKey: E(storageNode).getStoreKey,
+    getUnserializer: () => unserializer,
```
gibson042, line 69: What advantage does this indirection provide?
```suggestion
    unserialize: E(marshaller).unserialize,
```

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,76 @@
...
+        const encoded = JSON.stringify(serialized);
+        return E(storageNode).setValue(encoded);
+      })
+      .catch(fail);
+  };
+
+  // Start publishing the source.
+  observeIteration(subscribeEach(subscriber), {
+    updateState: publishValue,
+    finish: publishValue,
```
gibson042: It's a little strange to publish a final value but ignore a failure value... I'm not certain it's wrong (and at any rate don't know how it encoding would differentiate it from a value unless we introduce a wrapper), but it at least merits some commentary.

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,76 @@
...
+    publishFailed = true;
+    publishException = err;
+  };
+
+  // Must *not* be an async function, because it sometimes must throw to abort
+  // the iteration.
+  const publishValue = obj => {
+    if (publishFailed) {
+      // To properly abort the iteration, this must be a synchronous exception.
+      throw publishException;
```
gibson042: This still strikes me as a code smell, and I'm concerned about race conditions. It would seem much better to use a pattern that enforces serial consumption rather than the somewhat sloppy `observeIterator`.
```js
const iteratorP = E(subscribeEach(subscriber))[Symbol.asyncIterator]();
const consumeResultsSerially = async (iteratorP, consumeValue) => {
  let finished = false;
  while (!finished) {
    const { value, done } = await E(iteratorP).next();
    await consumeValue(value);
    finished = done;
  }
};
consumeResultsSerially(iteratorP, publishValue)
  // TODO: How should we handle and/or surface this failure?
  .catch(() => {});
```

packages/notifier/src/storesub.js
```diff
@@ -117,3 +187,31 @@ export const makeStoredPublisherKit = (storageNode, marshaller, childPath) => {
...
+ *
+ * What's different is `subscriber` tees records, writing out to storageNode.
+ *
+ * @template [T=unknown]
+ * @param {ERef<StorageNode>} storageNode
+ * @param {ERef<Marshaller>} marshaller
+ * @param {string} [childPath]
+ * @returns {StoredPublishKit<T>}
+ */
+export const makeStoredPublishKit = (storageNode, marshaller, childPath) => {
```
gibson042: I remain opposed to this `childPath` argument; it's just out of scope for the notifier package.

gibson042 (changes requested):

packages/notifier/src/storesub.js
dckc: @gibson042 any suggestion for what to do instead?

dckc: We have a `makeChildNode` method somewhere; is that a different package? Does this function belong there?

packages/notifier/src/storesub.js
turadg: The reason `subscriber` is taken as a presence is so this can add methods to it to make the `StoredSubscriber`: https://github.com/Agoric/agoric-sdk/blob/b662e1995831973f996ccbfc823ccb4c7c896895/packages/notifier/src/storesub.js#L63-L69

Is there a way to do that on a remote ref?

In terms of value of making the change, I don't know of use cases where you'd get a remote subscriber and want to make a stored one out of it. It seems like that would be a different stream for which you'd want a repeater.

packages/notifier/src/storesub.js
turadg: An `Unserializer` is an attenuated, read-only `Marshaller`. This is copied but I think the need was to pass around "Marshaller" or "read only Marshaller" to where it's needed. They have to be far objects. Example use:
https://github.com/Agoric/agoric-sdk/blob/b662e1995831973f996ccbfc823ccb4c7c896895/packages/casting/src/follower-cosmjs.js#L344

Since this is a read-only Marshaller, I do think it would be more consistent to have the names relate. Maybe `Unmarshaller`?

I think it would help to have this attenuation be done by a function to help document the pattern. @kriskowal WDYT of that and being  next to `makeMarshall` in `@endo/marshall`?

packages/notifier/src/storesub.js
turadg: @gibson042 you get to decide what's out of scope for `notifier` package so I'll find another place.

By the same token, should `makeStoredPublishKit` and `StoredPublishKit` also go? They can be made from `makePublishKit` and `makeStoredSubscriber`. I think the answer is no unless we also move out makeStoredSubscriber`. Though I wonder if we shouldn't export `makeStoredSubscriber` and only export `makeStoredPublishKit`.

packages/notifier/src/storesub.js
turadg: Good idea. PTAAL.

packages/notifier/src/storesub.js
gibson042: I can think of some brittle approaches like `subscribeAfter: E(subscriber).subscribeAfter`, but I agree that the value probably isn't there. Perhaps we should just rename the other parameters to indicate the difference?
```suggestion
 * @param {Subscriber<T>} subscriber
 * @param {ERef<StorageNode>} storageNodeP
 * @param {ERef<ReturnType<typeof makeMarshal>>} marshallerP
```

packages/notifier/src/storesub.js
gibson042: I'm mostly wondering if there's value in the object returned from `makeStoredSubscriber` providing a dedicated getter for such an object vs. just itself being a superset of one—e.g., `const unserializerP = E(storedSubscriber).getUnserializer()` vs. `const unserializer = Far('unserializer', { unserialize: E(storedSubscriber).unserialize })` or (depending upon trust level) `const unserializer = storedSubscriber`. This is not a blocking concern, just food for thought.

packages/notifier/src/storesub.js
gibson042: I'm on the fence about "makeStored" helpers being in the notifier package, but am willing to run with them for now. However, this parameter in particular just seems too tightly coupled... I'd strongly prefer invocations like `makeStoredPublisherKit(E(storageNode).getChildNode(childNodeName), marshaller)` over `makeStoredPublisherKit(storageNode, marshaller, childNodeName)`.

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,92 @@
...
+    // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
+    const { value, done } = await E(iteratorP).next();
+    consumeValue(value);
+    finished = !!done;
+  }
+};
+
 /**
+ * Begin iterating the source, storing serialized iteration values.  If the
+ * storageNode's `setValue` operation rejects, the iteration will be terminated.
```
gibson042:
```suggestion
 * Begin iterating the source, storing serialized iteration values.  If the
 * storageNode's `setValue` operation rejects, no further writes to it will
 * be attempted (but results will remain available from the subscriber).
```

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,92 @@
...
+ */
+const forEachPublicationRecord = async (subscriber, consumeValue) => {
+  const iteratorP = E(subscribeEach(subscriber))[Symbol.asyncIterator]();
+
+  let finished = false;
+  while (!finished) {
+    // Allow nested awaits (in loop) because it's safe for each to run in a turn
+    // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
+    const { value, done } = await E(iteratorP).next();
+    consumeValue(value);
```
gibson042: Avoiding the race (and eliminating the smelly "this must be a synchronous exception") requires not attempting a `setValue` until the previous one has settled.
```suggestion
    // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
    await consumeValue(value);
```

packages/notifier/src/storesub.js
```diff
@@ -2,9 +2,92 @@
...
+      .serialize(obj)
+      .then(serialized => {
+        const encoded = JSON.stringify(serialized);
+        return E(storageNode).setValue(encoded);
+      })
+      .catch(err => {
+        storingFailed = true;
+        storingException = err;
+      });
+  };
```
gibson042:
```suggestion
  const storeValue = value =>
    E(marshaller)
      .serialize(value)
      .then(serialized => {
        const encoded = JSON.stringify(serialized);
        return E(storageNode).setValue(encoded);
      });
```

packages/notifier/test/test-stored-subscription.js
```diff
@@ -80,3 +95,80 @@ test('stored subscription', async t => {
...
+    const origValue = unserialize({ body: encoded, slots: [] });
+    await E(publisher).publish(origValue);
+    await check(encoded, origValue, false);
+  }
+
+  const terminalNull = null;
+  await E(publisher).finish(terminalNull);
+  await check('terminal null', terminalNull, true);
+});
+
```
gibson042, line 177:
```suggestion

test.todo('stored subscriber with subscriber failure');

test.todo('stored subscriber with setValue failure');

```

gibson042 (changes requested):
I think we need to avoid the racy use of `setValue`, but everything else can be a TODO or (if you feel strongly) just left as-is.

packages/notifier/src/storesub.js
turadg: My understanding of the Agoric coding style is to actively avoid Hungarian notation. The types are there in the JSDoc and checked by tsc. There's no confusion with a presence to disambiguate.

Going to resolve this to leave as-is but feel free to reopen.

packages/notifier/src/storesub.js
turadg: > value in the object returned from `makeStoredSubscriber` providing a dedicated getter for such an object

yes, I think the value is symmetry with `Marshaller`. We can refactor this later for MN-2.

turadg:
@gibson042 I took your changes and added [0041189](https://github.com/Agoric/agoric-sdk/pull/5766/commits/0041189fd6d2cd44138d0516cce6d7fad21d657d) so the `todo` would work. Then I rebased to clean up for merge.

gibson042 (approved):
Thanks, LGTM!

PR #5768, opened 2022-07-14
Distinctive rejection values for promises severed by vat upgrade

### LINKED ISSUE #5649: each vat upgrade should use a distinct value when it rejects the abandoned promises

When a vat is upgraded, clients who are subscribed to one of their decided Promises will observe those Promises get rejected during the upgrade (since we don't have any form of "durable promise" yet). The rejection value is currently a static string, `vat upgraded`, controlled by `packages/SwingSet/src/liveslots/stop-vat.js` in the `rejectAllPromises` function.

@dtribble pointed out that downstream clients need to be able to distinguish between multiple upgrades that happen in succession (perhaps as followers of a Notification/Subscription which need to decide whether to re-subscribe?), so these promises need to be rejected with a value that is different for each upgrade.

@mhofman suggested providing the rejection value as a parameter to the `E(adminFacet).upgradeVat(newBundlecap, options)` message.

It might be interesting to allow the rejection value to contain a Presence (or other arbitrary Passables, not just pass-by-copy data). This would allow for unforgeable tokens. I think @dtribble and @erights had opinions on whether this was useful or necessary. The implementation would be more complex if we allow arbitrary capdata.

## Description of the Design

Assuming a `options.promiseRejectionValue` in `upgradeVat` (and assuming it needs to be capdata, not plain pass-by-copy):
* `src/vats/vat-admin/vat-vat-admin.js` parses `options`: this needs to allow and extract `promiseRejectionValue` and pass it to the hook
* `src/kernel/vat-admin-hooks.js > upgrade()` is called with the capdata of the arguments
  * this needs to handle the split between slots being present in either `vatParameters` or `promiseRejectionValue`, which will take some surgery: previously we've assumed that any `.slots` are part of `vatParameters`, and if both options might have object references, we'd need to tease apart their two `slots` (and probably rewrite the `.body` to use different `index: NN` values, which would be the first time the kernel is parsing `capdata.body`, and I was hoping to avoid that entirely)
  * maybe `vat-vat-admin.js` could create a new unique upgrade token for each upgrade, and deliver it in some distinct way, so the capdata slots are not co-mingled
  * the refcount on all slots must be incremented as the values are put onto `kernelKeeper.addToAcceptanceQueue()`, both for `vatParameters` and `promiseRejectionValue`
* in `kernel.js > processUpgradeVat()`, we need to pass `promiseRejectionValue` (capdata) into the `stopVat` delivery, since that's currently where the old promises get rejected
  * we currently don't provide any parameters to `stopVat`, but we do pass the new `vatParameters` capdata into `startVat()`, which can be used as a model
  * the `vatTranslators.js > makeTranslateKernelDeliveryToVatDelivery()` function must enhance its `translateStopVat` method to map the slots, just like `translateStartVat` does
* `liveslots.js` must be changed to accept parameters to `dispatch.stopVat` and deliver them to the `stopVat()` function in `stop-vat.js`, deserializing them
  * this should probably prohibit Promises, just to limit the complexity
* then `stop-vat.js` can use the provided value in its `reject()` call

We've been on the fence about whether the vat should perform its shutdown tasks during `dispatch.stopVat`, or if the kernel ought to take responsibility for them. Currently, the vat's `stopVat` does promise rejection and abandonment of many non-durable merely-virtual objects.

This feature might provide motivation for having the kernel take on responsibility for at least the promise rejection part. The kernel would need to walk the c-list looking for all promise entries (`vNN.c.p-` and `vNN.c.p+` prefixes). For each promise it finds, it needs to look up the kernel promise table entry and see who the decider is. If the vat being upgraded is the decider, the kernel can pretend the vat emitted a `syscall.resolve([vpid, true, promiseRejectionValueCapdata])`, incrementing the refcount on the value's `.slots` and removing the entry from the c-list.

Having the kernel do this is marginally less efficient than having the vat do it (within the vat, liveslots has a `Map` of promises it controls, so it can just walk that map, whereas the kernel must do a DB crawl and at least two lookups on each result). But in general we want the vat to be less involved in its imminent demise/upgrade. And it would be a tiny bit simpler if the kernel did not need to translate the `stopVat` arguments.

## Security Considerations

None I can think of. This does provide an information+authority path from the upgrader to a bunch of promise-followers that the upgrader didn't originally have access to, but the upgrader is already in control of the entire post-upgrade behavior of the vat. This just extends their control slightly earlier, giving them control over the rejection value of all the pre-upgrade pending promises.

## GC Considerations

If the rejection value is allowed to hold Presences, this may exacerbate a limitation of our device model, because `device-vat-admin` (like all devices) does not currently participate in GC. So Presences included in `vatParameters` or `promiseRejectionValue` would be dropped by `vat-vat-admin` and the vat being upgraded, but would be held by `device-vat-admin` forever (or until we figure out a backwards-compatible way to shoehorn GC into `deviceSlots.js`).

## Test Plan

The unit tests in `test/upgrade/` should be augmented to include a case where a rejection value is included in the ugprade options. If we decide to allow Presences, a separate test should include one, and the test should find a way to ensure that the Presence has been GCed appropriately once everything is finished.

FUDCo:
Closes #5649

### DIFF
--- a/packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
+++ b/packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
@@ -206,14 +206,19 @@ export function buildDevice(tools, endowments) {
           return returnFromInvoke(vatID);
         }
 
-        // D(devices.vatAdmin).upgradeVat(vatID, bundleID, vatParameters) -> upgradeID
+        // D(devices.vatAdmin).upgradeVat(vatID, bundleID, vatParameters, upgradeMessage) -> upgradeID
         if (method === 'upgradeVat') {
           const args = JSON.parse(argsCapdata.body);
           assert(Array.isArray(args), 'upgradeVat() args array');
-          assert.equal(args.length, 3, `upgradeVat() args length`);
-          const [vatID, bundleID, _vatParameters] = args;
+          assert.equal(args.length, 4, `upgradeVat() args length`);
+          const [vatID, bundleID, _vatParameters, upgradeMessage] = args;
           insistVatID(vatID);
           assert.typeof(bundleID, 'string', `upgradeVat() bundleID`);
+          assert.typeof(
+            upgradeMessage,
+            'string',
+            'upgradeVat() upgradeMessage',
+          );
 
           const res = syscall.callKernelHook('upgrade', argsCapdata);
           const upgradeID = JSON.parse(res.body);
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -394,6 +394,7 @@ export default function buildKernel(
     const vs = kernelSlog.provideVatSlogger(vatID).vatSlog;
     try {
       /** @type { VatDeliveryResult } */
+      // eslint-disable-next-line @jessie.js/no-nested-await
       const deliveryResult = await vatWarehouse.deliverToVat(vatID, kd, vd, vs);
       insistVatDeliveryResult(deliveryResult);
       // const [ ok, problem, usage ] = deliveryResult;
@@ -812,8 +813,10 @@ export default function buildKernel(
    */
   async function processUpgradeVat(message) {
     assert(vatAdminRootKref, `initializeKernel did not set vatAdminRootKref`);
-    const { vatID, upgradeID, bundleID, vatParameters } = message;
+    const { vatID, upgradeID, bundleID, vatParameters, upgradeMessage } =
+      message;
     insistCapData(vatParameters);
+    assert.typeof(upgradeMessage, 'string');
 
     const vatInfo = vatWarehouse.lookup(vatID);
     if (!vatInfo) {
@@ -821,8 +824,17 @@ export default function buildKernel(
     }
     const { meterID } = vatInfo;
     const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
+    const disconnectObject = {
+      name: 'vatUpgraded',
+      upgradeMessage,
+      incarnationNumber: vatKeeper.getIncarnationNumber(),
+    };
+    const disconnectObjectCapData = {
+      body: JSON.stringify(disconnectObject),
+      slots: [],
+    };
     /** @type { import('../types-external.js').KernelDeliveryStopVat } */
-    const kd1 = harden(['stopVat']);
+    const kd1 = harden(['stopVat', disconnectObjectCapData]);
     const vd1 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd1);
     const status1 = await deliverAndLogToVat(vatID, kd1, vd1);
 
@@ -886,6 +898,7 @@ export default function buildKernel(
     const source = { bundleID };
     const { options } = vatKeeper.getSourceAndOptions();
     vatKeeper.setSourceAndOptions(source, options);
+    const incarnationNumber = vatKeeper.incIncarnationNumber();
     // TODO: decref the bundleID once setSourceAndOptions increfs it
 
     // pause, take a deep breath, appreciate this moment of silence
@@ -918,7 +931,7 @@ export default function buildKernel(
       return results;
     }
 
-    const args = [upgradeID, true, undefined];
+    const args = [upgradeID, true, undefined, incarnationNumber];
     const vatAdminMethargs = {
       body: JSON.stringify(['vatUpgradeCallback', args]),
       slots: [],
@@ -1076,7 +1089,8 @@ export default function buildKernel(
    *              dynamicOptions: InternalDynamicVatOptions }
    *          } RunQueueEventCreateVat
    * @typedef { { type: 'upgrade-vat', vatID: VatID, upgradeID: string,
-   *              bundleID: BundleID, vatParameters: SwingSetCapData } } RunQueueEventUpgradeVat
+   *              bundleID: BundleID, vatParameters: SwingSetCapData,
+   *              upgradeMessage: string } } RunQueueEventUpgradeVat
    * @typedef { { type: 'changeVatOptions', vatID: VatID, options: Record<string, unknown> } } RunQueueEventChangeVatOptions
    * @typedef { { type: 'startVat', vatID: VatID, vatParameters: SwingSetCapData } } RunQueueEventStartVat
    * @typedef { { type: 'dropExports', vatID: VatID, krefs: string[] } } RunQueueEventDropExports
@@ -1297,10 +1311,11 @@ export default function buildKernel(
     }
     processQueueRunning = Error('here');
     try {
-      return await processor(message).finally(() => {
-        processQueueRunning = undefined;
-      });
+      const result = await processor(message);
+      processQueueRunning = undefined;
+      return result;
     } catch (err) {
+      processQueueRunning = undefined;
       // panic() sets the kernelPanic flag which will be checked on the way out
       // by run() or step().
       panic(`error during tryProcessMessage: ${err}`, err);
@@ -1594,7 +1609,7 @@ export default function buildKernel(
 
       const bundle = kernelKeeper.getBundle(source.bundleID);
       assert(bundle);
-      // eslint-disable-next-line no-await-in-loop
+      // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
       const NS = await importBundle(bundle, {
         filePrefix: `dev-${name}/...`,
         endowments: harden({ ...vatEndowments, console: devConsole, assert }),
@@ -1694,6 +1709,7 @@ export default function buildKernel(
     const { processor, message } = getNextMessageAndProcessor();
     // process a single message
     if (message) {
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await tryProcessMessage(processor, message);
       if (kernelPanic) {
         throw kernelPanic;
@@ -1748,7 +1764,7 @@ export default function buildKernel(
       }
       count += 1;
       /** @type { PolicyInput } */
-      // eslint-disable-next-line no-await-in-loop
+      // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
       const policyInput = await tryProcessMessage(processor, message);
       if (kernelPanic) {
         throw kernelPanic;
--- a/packages/SwingSet/src/kernel/vat-admin-hooks.js
+++ b/packages/SwingSet/src/kernel/vat-admin-hooks.js
@@ -80,9 +80,10 @@ export function makeVatAdminHooks(tools) {
     upgrade(argsCapData) {
       // marshal([vatID, bundleID, vatParameters]) -> upgradeID
       const argsJSON = JSON.parse(argsCapData.body);
-      const [vatID, bundleID, vpJSON] = argsJSON;
+      const [vatID, bundleID, vpJSON, upgradeMessage] = argsJSON;
       insistVatID(vatID);
       assert.typeof(bundleID, 'string');
+      assert.typeof(upgradeMessage, 'string');
       const vpCD = { body: JSON.stringify(vpJSON), slots: argsCapData.slots };
       for (const kref of vpCD.slots) {
         kernelKeeper.incrementRefCount(kref, 'upgrade-vat-event');
@@ -94,6 +95,7 @@ export function makeVatAdminHooks(tools) {
         upgradeID,
         bundleID,
         vatParameters: vpCD,
+        upgradeMessage,
       };
       kernelKeeper.addToAcceptanceQueue(harden(ev));
       const upgradeIDCD = { body: JSON.stringify(upgradeID), slots: [] };
--- a/packages/SwingSet/src/kernel/vatTranslator.js
+++ b/packages/SwingSet/src/kernel/vatTranslator.js
@@ -175,11 +175,12 @@ function makeTranslateKernelDeliveryToVatDelivery(vatID, kernelKeeper) {
     return startVatMessageVatDelivery;
   }
 
-  /** @type { import('../types-external.js').VatDeliveryStopVat } */
-  const stopVatMessage = harden(['stopVat']);
-
-  function translateStopVat() {
-    return stopVatMessage;
+  /**
+   * @param { SwingSetCapData } disconnectObjectCapData
+   * @returns { import('../types-external.js').VatDeliveryStopVat }
+   */
+  function translateStopVat(disconnectObjectCapData) {
+    return harden(['stopVat', disconnectObjectCapData]);
   }
 
   /** @type { VatDeliveryBringOutYourDead } */
@@ -241,8 +242,8 @@ function makeTranslateKernelDeliveryToVatDelivery(vatID, kernelKeeper) {
     //  ['dropExports', vrefs]
     //  ['retireExports', vrefs]
     //  ['retireImports', vrefs]
-    //  ['startVat']
-    //  ['stopVat']
+    //  ['startVat', vatParameters]
+    //  ['stopVat', disconnectObject]
     //  ['bringOutYourDead']
   }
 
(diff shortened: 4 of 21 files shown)


packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -94,7 +94,11 @@ export function buildRootObject(vatPowers) {
         return doneP;
       },
       upgrade(bundlecap, options = {}) {
-        const { vatParameters, ...rest } = options;
+        const {
+          vatParameters,
+          upgradeMessage = 'vat upgraded',
+          ...rest
+        } = options;
```
warner, line 101: This needs an `assert.typeof(upgradeMessage, 'string', 'updateMessage not a string')`, to signal the user of their mistake as early as possible (and because if `upgradeMessage` contains slots, it will confuse serialization when we invoke the device in a moment)

packages/SwingSet/src/devices/vat-admin/device-vat-admin.js
```diff
@@ -206,12 +206,12 @@ export function buildDevice(tools, endowments) {
...
 
-        // D(devices.vatAdmin).upgradeVat(vatID, bundleID, vatParameters) -> upgradeID
+        // D(devices.vatAdmin).upgradeVat(vatID, bundleID, vatParameters, upgradeMessage) -> upgradeID
         if (method === 'upgradeVat') {
           const args = JSON.parse(argsCapdata.body);
           assert(Array.isArray(args), 'upgradeVat() args array');
-          assert.equal(args.length, 3, `upgradeVat() args length`);
-          const [vatID, bundleID, _vatParameters] = args;
+          assert.equal(args.length, 4, `upgradeVat() args length`);
+          const [vatID, bundleID, _vatParameters, _upgradeMessage] = args;
```
warner: This should get a (redundant) `assert.typeof` check on `upgradeMessage`, just after the one on `bundleID`. (We can't usefully do any type assertions on `_vatParameters`, since it's allowed to be anything, so that one remains underscore-prefixed).

packages/SwingSet/src/kernel/vat-admin-hooks.js
```diff
@@ -80,7 +80,7 @@ export function makeVatAdminHooks(tools) {
     upgrade(argsCapData) {
       // marshal([vatID, bundleID, vatParameters]) -> upgradeID
       const argsJSON = JSON.parse(argsCapData.body);
-      const [vatID, bundleID, vpJSON] = argsJSON;
+      const [vatID, bundleID, vpJSON, upgradeMessage] = argsJSON;
```
warner, line 83: this too should get an `assert.typeof` on `upgradeMessage`

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1297,7 +1306,7 @@ export default function buildKernel(
     }
     processQueueRunning = Error('here');
     try {
-      return await processor(message).finally(() => {
+      return processor(message).finally(() => {
```
warner: This.. something's funky here.

The intention is
* 1: if `processor()` (which is probably `async`, but I haven't checked all possibilities) rejects, we want to `panic()`
* 2: regardless of whether it fulfills or rejects, we want to clear the `processQueueRunning` anti-reentrancy semaphore

I'm guessing that it looked safe to remove the `await` because we were *actually* awaiting the return value from `finally()`, which might be a Promise that always fulfills, even if `processor(message)` were to reject?

In that case, the old version had a bug (it would fail to panic the kernel if e.g. some `assert` within `processor` failed), and this change preserves the bug, but we need to rewrite it to fix the bug.

Maybe:
```js
try {
  const result = await processor(message);
  return result;
} catch (err) {
  panic(..)
  return ['crank-failed', {}];
}
```

Or maybe
```js
const p = processor(message);
p.finally(() => processQueueRunning = undefined);
p.catch(err => { panic(); return ['crank-failed', {}] });
return p;
```

except I'm not sure that is guaranteed to set the panic flag before the caller gets to see the return promise resolve and then start doing more things when it's supposed to be panicing.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -812,7 +813,8 @@ export default function buildKernel(
    */
   async function processUpgradeVat(message) {
     assert(vatAdminRootKref, `initializeKernel did not set vatAdminRootKref`);
-    const { vatID, upgradeID, bundleID, vatParameters } = message;
+    const { vatID, upgradeID, bundleID, vatParameters, upgradeMessage } =
+      message;
```
warner, line 817: we should have one more `assert.typeof` here, which will guard against cases where `message.upgradeMessage` is missing (maybe an internal bug some day)

packages/SwingSet/src/kernel/kernel.js
warner: Ah, @FUDCo and I looked at the docs (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally), and `finally()` does the expected thing: it returns a new Promise which (if the finally handler doesn't throw) will fulfill or reject in exactly the same way as the original, or (if the handler *does* throw) will reject with the handler's thrown value.

So the original did not have a bug, and removing the `await` causes the `panic` to be bypassed. The original was not a paragon of readability, so I'm happy to see it get rewritten, as long as the new version panics at the right time.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -821,8 +823,13 @@ export default function buildKernel(
...
     const { meterID } = vatInfo;
     const vatKeeper = kernelKeeper.provideVatKeeper(vatID);
+    const disconnectObject = {
+      name: 'vatUpgraded',
+      upgradeMessage,
+      seqNum: vatKeeper.getSeqNum(),
+    };
     /** @type { import('../types-external.js').KernelDeliveryStopVat } */
-    const kd1 = harden(['stopVat']);
+    const kd1 = harden(['stopVat', disconnectObject]);
```
warner: In general, `VatDeliveryObject`s like this `kd1` are populated with capdata (i.e. `dispatch.deliver` or the individual promise resolutions of `dispatch.notify`). I'm a little on the fence about it, but if we maintain that practice, then we should do:
```js
const disconnectObjectCD = harden({ body: JSON.stringify(disconnectObject), slots: []});
const kd1 = harden(['stopVat', disconnectObjectCD]);
```

and either change liveslots to do a `disconnectObject = m.unserialize(disconnectObjectCD)` at the top of `stopVat(disconnectObjectCD)`, or remove the `m.serialize(disconnectObject)` from `rejectAllPromises()`. Also the type definition in types-external.js would need to change.

The vatTranslator.js code that translates `stopVat` could mostly stay the same, but it would want an `insistCapData` check and also an `assert.equal(disconnectObjectCD.slots.length, 0)` to guard against us allowing slots in the future and forgetting to add `convertKernelSlotToVatSlot` calls to the translator (which would leak untranslated krefs into the vat).

On one hand, consistency is nice, and this paves the way for more complicated `disconnectObject` contents. It would take more work on both sides to allow krefs/vrefs in the object (`vatTranslator.js` would need updating, but also liveslots has more bookkeeping to do if/when it sends slots into the kernel), but we'd be halfway there.

On the other hand, I don't want to allow slots in a `disconnectObject` any time soon, so most of this would feel like self-inflicted busywork.

The one nice thing about using capdata is that changes to the contents of the disconnect object would not be visible to the liveslots *types*: if we added a fourth property, or changed the name of the `seqNum` parameter, the `VatDeliveryObject` shape would remain exactly the same. In practice it wouldn't matter: liveslots just serializes whatever you give it. But in principle, if we change `DisconnectObject` within the kernel, we've just changed the API signature of the liveslots `stopVat`, and that's running in a different process. If the rule is "it's just capdata", then they're the same.

Not sure what to recommend.

packages/SwingSet/src/lib/message.js
```diff
@@ -72,7 +72,7 @@ export function insistVatDeliveryObject(vdo) {
       break;
     }
     case 'stopVat': {
-      assert(rest.length === 0);
+      assert(rest.length === 1);
```
warner, line 75: This wants extra asserts for the properties of the new argument, like how `case 'message'` and `case 'notify'` do it.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1505,9 +1505,10 @@ function build(
   }
 
   /**
+   * @param { import('../types-external.js').DisconnectObject } disconnectObject
    * @returns {Promise<void>}
    */
-  async function stopVat() {
+  async function stopVat(disconnectObject) {
```
warner: We need to give `disconnectObject` to `watchedPromiseManager.prepareShutdownRejections()` too, and change the implementation to store the serialized DO in the DB next to `deadPromises`, and change `loadWatchedPromiseTable` to use it for the rejections instead of the current hardcoded `VatUpgradedMessage = 'vat upgraded'`, so that all promises (locally-decided locally-watched ones and locally-decided remotely-followed ones) are all rejected in the same way.

The stored DO should be deleted during `loadWatchedPromiseTable` at the same time it deletes `deadPromises`.

That also means we need to update `test-promise-watcher.js` to look for the more detailed DisconnectionObject.

packages/SwingSet/src/kernel/kernel.js
warner: The necessary change to `watchedPromiseManager.prepareShutdownRejections` pushes me slightly in favor of going with capdata, because now we're introducing a second API boundary: this one between the current version of liveslots and some future version of liveslots (communicating through the vatstore DB). "Uniform capdata with empty `.slots`" is a decent type to share across versions.

packages/SwingSet/src/types-external.js
```diff
@@ -93,6 +93,7 @@ export {};
  * }} Message
  *
  * @typedef { 'none' | 'ignore' | 'logAlways' | 'logFailure' | 'panic' } ResolutionPolicy
+ * @typedef {{ name: string, upgradeMessage: string, seqNum: number }} DisconnectObject
```
warner: as discussed, `seqNum` feels a little terse and lacking context here: someone looking at a DisconnectObject in the resolution of `E(facetAdmin).upgrade(..)` will know that it has something to do with upgrade, and that it's a sequence of *something*, but the name doesn't suggest that we're talking about a sequence of vat incarnations, coupled by upgrade events. If I was willing to commit `incarnation` for this concept, then `incarnationNum` might provide enough context to make me happy, but I'm not 100% there yet.

warner (changes requested):
The kernel await/panic thing is the most important change, followed by `vat-vat-admin` doing a type check early, followed by the watchedPromises part.

Also, `docs/vat-upgrade.md` needs updating.

I'm inclined to ask for the uniform-capdata thing, but see what you feel about it, I could be talked out of it.

packages/SwingSet/src/kernel/kernel.js
FUDCo: One benefit (albeit principally just an aesthetic benefit) would be to make `startVat` and `stopVat` have parallel form.

packages/SwingSet/src/liveslots/liveslots.js
FUDCo: Oy.  Will do.

packages/SwingSet/src/kernel/vatTranslator.js
```diff
@@ -175,11 +175,12 @@ function makeTranslateKernelDeliveryToVatDelivery(vatID, kernelKeeper) {
...
-  /** @type { import('../types-external.js').VatDeliveryStopVat } */
-  const stopVatMessage = harden(['stopVat']);
-
-  function translateStopVat() {
-    return stopVatMessage;
+  /**
+   * @param { SwingSetCapData } disconnectObject
+   * @returns { import('../types-external.js').VatDeliveryStopVat }
+   */
+  function translateStopVat(disconnectObject) {
```
warner: Let's name this `disconnectObjectCD` or `disconnectObjectCapData` to match the name we use in `kernel.js` where we build the "stopVat" `VatDeliveryObject`, and distinguish it from the variables named `disconnectObject` which hold non-capdata.

packages/SwingSet/src/lib/message.js
```diff
@@ -72,7 +72,9 @@ export function insistVatDeliveryObject(vdo) {
...
+      const [disconnectObject] = rest;
```
warner: to be consistent, let's call this `disconnectObjectCapData` too

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -116,10 +121,10 @@ export function makeWatchedPromiseManager(
           const [watcher, ...args] = watch;
           void Promise.resolve().then(() => {
             if (watcher.onRejected) {
-              watcher.onRejected(VatUpgradedMessage, ...args);
+              watcher.onRejected(disconnectObjectCapData, ...args);
             } else {
               // eslint-disable-next-line no-throw-literal
-              throw VatUpgradedMessage;
+              throw disconnectObjectCapData;
```
warner: I believe we can remove the `no-throw-literal` eslint-husher now

warner: Also, throwing capdata is a little weird, I think maybe we should `throw m.unserialize(disconnectObjectCapData)` instead.

It doesn't matter much, we're only doing this to feed an unhandled-rejection handler, and the diagnostic output it emits is just as likely to be useful with dumping capdata as it is dumping a disconnection object.

packages/SwingSet/test/promise-watcher/test-promise-watcher.js
```diff
@@ -88,22 +88,24 @@ async function testPromiseWatcher(t) {
   // console.log(`-- starting upgradeV2`);
   const [v2status /* , v2capdata */] = await run('upgradeV2', []);
   t.is(v2status, 'fulfilled');
+  const encodedDO =
```
warner: This should be looking for non-capdata.

packages/SwingSet/src/liveslots/watchedPromises.js
```diff
@@ -116,10 +121,10 @@ export function makeWatchedPromiseManager(
           const [watcher, ...args] = watch;
           void Promise.resolve().then(() => {
             if (watcher.onRejected) {
-              watcher.onRejected(VatUpgradedMessage, ...args);
+              watcher.onRejected(disconnectObjectCapData, ...args);
```
warner: This definitely needs a `m.unserialize(disconnectObjectCapData)`, because we should not be passing capdata to the userspace rejection handler. I'm a little surprised that a unit test didn't catch this.. oh, I see, the unit test was updated to look for the wrong thing.

So yeah, change both this and the `test-promise-watcher.js` to look for an object literal (the disconnection object), not the capdata encoding of that object literal.

warner (approved):
Nice! The unserialize-before-rejecting in `watchedPromises.js` and test is a necessary change. The renaming of `disconnectObject` to `disconnectObjectCapData` in the capdata-bearing variables/arguments is a nice-to-have.

Feel free to land once they're done.

packages/SwingSet/src/lib/message.js
FUDCo: I was going to do that but I was following the pattern of the parallel `startVat` request, which takes a `vatParameters` arg that's also capdata.  But since we're using `disconnectObjectCapData` everywhere it makes sense to do it here too.  (And I guess we're using `vatParameters` everywhere instead of, say `vatParametersCapData`, so...)

packages/SwingSet/src/liveslots/watchedPromises.js
FUDCo: Too many layers...

FUDCo: Marginally tricky because the promise watcher wasn't passed the unserialize function, but easily taken care of...

PR #5770, opened 2022-07-15
5764 upgrade contracts

### LINKED ISSUE #5764: Zoe support upgradeable contracts using start

#5635 made it possible for Zoe contracts to be upgraded by requiring a new API. In discussion spanning multiple meetings today, we surfaced a proposal for an approach that is much more backward compatible.

## Description of the Design

Contracts define a `start()` method that takes three parameters: `zcf`, `privateArgs`, and `baggage`. They can be "Durable" by storing their precious state in durable stores. In order to be "Reachably Durable", they have to define a `VersionID` in their baggage, and store (and retrieve) all their precious state in the baggage. 

One consequence of this is there  is no longer a separation in contract startup between installation state and instance state. This means that zygotes wouldn't be able to amortize startup costs for instance-agnostic state across multiple instances. Until we have zygotes, this was an empty advantage anyway.

Contracts that don't define `VersionID` in their top-level baggage will be blocked from upgrading. Contracts can determine whether they are the first incarnation of an instance by looking for `VersionID`. It won't be present the first time. Thereafter, if it's present, the contract can be sure that this is a revived contract, and can use the `VersionID` to distinguish versions. All versions are required to return the same members of `{ creatorFacet, publicFacet, creatorInvitation }`.  ZCF will compare identity and complain if they change between incarnations.

## Security Considerations

This is how Zoe starts contracts, so all the invariants must continue to be upheld.

## Test Plan

test coveredCall upgrade

Chris-Hibbert:
closes: #5764

## Description

Refator #5635 to be more compatible with classic contracts. `start` continues to be a supported export for contracts. In order to be upgradeable, start has to be renamed to `vivify`, though the author can wait until V2 to do so. `start` now includes a baggage parameter that the contract can use to make reachably durable contracts.

### Security Considerations

Continue to maintain Zoe's invariants.

#5708 needs to be updated with the revised interface.

### Testing Considerations

There's a test that demonstrates that the coveredCall contract can be upgraded.

### DIFF
--- a/packages/vat-data/src/kind-utils.js
+++ b/packages/vat-data/src/kind-utils.js
@@ -91,3 +91,8 @@ export const vivifySingleton = (
 };
 // @ts-expect-error TODO statically recognize harden
 harden(vivifySingleton);
+
+export const vivifyKind = (baggage, name, init, behavior) =>
+  defineDurableKind(provideKindHandle(baggage, name), init, behavior);
+// @ts-expect-error TODO statically recognize harden
+harden(vivifyKind);
--- a/packages/vat-data/src/kind-utils.js
+++ b/packages/vat-data/src/kind-utils.js
@@ -2,6 +2,10 @@ import { provide } from '@agoric/store';
 import { defineDurableKind, makeKindHandle } from './vat-data-bindings.js';
 
 /** @template L,R @typedef {import('@endo/eventual-send').RemotableBrand<L, R>} RemotableBrand */
+/** @typedef {import('@agoric/store').MapStore<string,unknown>} Baggage */
+/** @template T @typedef {import('./types.js').DefineKindOptions<T>} DefineKindOptions */
+/** @template T @typedef {import('./types.js').KindFacet<T>} KindFacet */
+/** @typedef {import('./types.js').DurableKindHandle} DurableKindHandle */
 
 const { entries, fromEntries } = Object;
 
@@ -12,6 +16,11 @@ export const dropContext =
 // @ts-expect-error TODO statically recognize harden
 harden(dropContext);
 
+/**
+ * @param {Baggage} baggage
+ * @param {string} kindName
+ * @returns {DurableKindHandle}
+ */
 export const provideKindHandle = (baggage, kindName) =>
   provide(baggage, `${kindName}_kindHandle`, () => makeKindHandle(kindName));
 // @ts-expect-error TODO statically recognize harden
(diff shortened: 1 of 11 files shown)


packages/vat-data/src/kind-utils.js
```diff
@@ -91,3 +91,8 @@ export const vivifySingleton = (
 };
 // @ts-expect-error TODO statically recognize harden
 harden(vivifySingleton);
+
+export const vivifyKind = (baggage, name, init, behavior) =>
+  defineDurableKind(provideKindHandle(baggage, name), init, behavior);
```
erights: I like it! Also, a more understandable delta from `defineKind`.

But you should still pass through the optional `options`

```suggestion
export const vivifyKind = (baggage, name, init, behavior, options = undefined) =>
  defineDurableKind(provideKindHandle(baggage, name), init, behavior, options);
```

packages/zoe/src/contracts/coveredCall-durable.js
```diff
@@ -72,24 +67,16 @@ const sellSeatExpiredMsg = `The covered call option is expired.`;
  * specified in the invitation value, and want the underlying assets
  * exactly.
  *
- * @param {import('@agoric/vat-data').Baggage} installationBaggage
+ * @param _privateArgs unknown
+ * @param {import('@agoric/vat-data').Baggage} instanceBaggage
  * @param {ZCF} zcf passed in setupInstallation, but not completely initialized
  *    until setupInstance. Passed early because it will often need to be
```
erights: Stale comment

packages/zoe/src/contracts/coveredCall-durable.js
```diff
@@ -72,24 +67,16 @@ const sellSeatExpiredMsg = `The covered call option is expired.`;
...
+  const makeExerciser = vivifyKind(
     instanceBaggage,
     'makeExerciserKindHandle',
-  );
-
-  const makeExerciser = defineDurableKind(
-    makeExerciserKindHandle,
     sellSeat => ({ sellSeat }),
     {
       handle: ({ state: { sellSeat } }, buySeat) => {
```
erights, line 83: Please note a TODO that this offerHandler should be a durable function. It is a durable object only because we don't yet support durable functions.

packages/zoe/src/contracts/coveredCall-durable.js
```diff
@@ -72,24 +67,16 @@ const sellSeatExpiredMsg = `The covered call option is expired.`;
  * specified in the invitation value, and want the underlying assets
  * exactly.
  *
- * @param {import('@agoric/vat-data').Baggage} installationBaggage
+ * @param _privateArgs unknown
+ * @param {import('@agoric/vat-data').Baggage} instanceBaggage
  * @param {ZCF} zcf passed in setupInstallation, but not completely initialized
```
erights: I don't think _privateArgs declaration syntax is correct.
@param declarations must be in the same order as the params

```suggestion
 * @param {ZCF} zcf passed in setupInstallation, but not completely initialized
 * @param {unknown} _privateArgs
 * @param {import('@agoric/vat-data').Baggage} instanceBaggage
```

erights (approved):
As a delta, LGTM!

packages/zoe/test/swingsetTests/upgradeCoveredCall/coveredCall-durable-V3.js
```diff
@@ -72,22 +67,14 @@ const sellSeatExpiredMsg = `The covered call option is expired.`;
  * specified in the invitation value, and want the underlying assets
  * exactly.
  *
- * @param {import('@agoric/vat-data').Baggage} installationBaggage
+ * @param _privateArgs
+ * @param {import('@agoric/vat-data').Baggage} instanceBaggage
  * @param {ZCF} zcf
```
erights: Likewise of the params of course

packages/vat-data/src/kind-utils.js
erights: Done by #5772

packages/zoe/src/contracts/coveredCall-durable.js
erights: Deleted by #5772

packages/zoe/src/contracts/coveredCall-durable.js
erights: TODO added by #5772

packages/zoe/src/contracts/coveredCall-durable.js
erights: Fixed by #5772

PR #5772, opened 2022-07-15
Suggestions for #5770

erights:
In the interests of getting #5770 merged into #5635 , so I can better continue review there, I put together this PR that implements my suggested changes to #5770 . By doing this as yet another delta staging, I avoid potential bert-conflict problems with @Chris-Hibbert .

### DIFF
--- a/packages/wallet/api/test/attestationExample.js
+++ b/packages/wallet/api/test/attestationExample.js
@@ -2,6 +2,7 @@
 import '@agoric/zoe/exported.js';
 import { AmountMath } from '@agoric/ertp';
 import { E } from '@endo/eventual-send';
+import { Far } from '@endo/marshal';
 
 /** @param {ZCF} zcf */
 export const start = async zcf => {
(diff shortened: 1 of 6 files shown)


packages/wallet/api/test/attestationExample.js
```diff
@@ -53,12 +54,12 @@ export const start = async zcf => {
 
   const getCallHistory = () => callHistory;
 
-  const publicFacet = {
```
erights, line 56: Why did this previously work without a `Far` declaration?

packages/zoe/test/unitTests/zcf/registerFeeMintContract.js
```diff
@@ -26,7 +27,7 @@ const start = async (zcf, privateArgs) => {
     zcfSeat,
   );
 
-  const creatorFacet = harden({
```
erights, line 29: Why did this previously work without a `Far` declaration?

PR #5774, opened 2022-07-15
testing improvements

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/5766 ran into a problem with `test.todo` not being available in ses-ava. 524e7de69408414abcfed42d6468b01953f9897b adds it. (fyi @erights )

I also got fed up with this noise in test output. 295c8aadf1840e1a315ab0354c83659e161db369 removes it.
Screen Shot 2022-07-12 at 7 55 13 AM

### Security Considerations

none

--

### Testing Considerations

Better

### DIFF
--- a/packages/SwingSet/test/test-message-patterns.js
+++ b/packages/SwingSet/test/test-message-patterns.js
@@ -2,13 +2,8 @@
 /* eslint dot-notation: "off" */
 /* eslint object-shorthand: "off" */
 
-// `test.serial` does not yet seem compatible with ses-ava
-// See https://github.com/endojs/endo/issues/647
-// TODO restore
-// import { test } from '../tools/prepare-test-env-ava.js';
-import '../tools/prepare-test-env.js';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import test from 'ava';
+// eslint-disable-next-line import/order -- has side-effects AND exports
+import { test } from '../tools/prepare-test-env-ava.js';
 
 import path from 'path';
 import bundleSource from '@endo/bundle-source';
--- a/packages/run-protocol/test/test-gov-collateral.js
+++ b/packages/run-protocol/test/test-gov-collateral.js
@@ -586,4 +586,4 @@ test('Committee can raise debt limit', async t => {
   });
 });
 
-// test.todo('users can open vaults');
+test.todo('users can open vaults');
--- a/packages/solo/test/test-home.js
+++ b/packages/solo/test/test-home.js
@@ -1,12 +1,6 @@
 /* global process */
 
-// `test.after.always` does not yet seem compatible with ses-ava
-// See https://github.com/endojs/endo/issues/647
-// TODO restore
-// import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
-import '@agoric/swingset-vat/tools/prepare-test-env.js';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import test from 'ava';
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
 
 import bundleSource from '@endo/bundle-source';
 import { AmountMath } from '@agoric/ertp';
--- a/packages/ui-components/test/display/test-display.js
+++ b/packages/ui-components/test/display/test-display.js
@@ -1,7 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
-import '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import test from 'ava'; // TODO ses-ava doesn't yet have test.todo
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
 
 import { parseAsValue } from '../../src/display/display.js';
 
--- a/packages/zoe/src/contractFacet/zcfZygote.js
+++ b/packages/zoe/src/contractFacet/zcfZygote.js
@@ -319,7 +319,6 @@ export const makeZCFZygote = (
     /** @type {SetTestJig} */
     setTestJig: (testFn = () => ({})) => {
       if (testJigSetter) {
-        console.warn('TEST ONLY: capturing test data', testFn);
         testJigSetter({ ...testFn(), zcf });
       }
     },
--- a/packages/zoe/test/unitTests/contracts/loan/test-addCollateral.js
+++ b/packages/zoe/test/unitTests/contracts/loan/test-addCollateral.js
@@ -1,8 +1,7 @@
 // @ts-nocheck
 // eslint-disable-next-line import/no-extraneous-dependencies
-import '@agoric/zoe/tools/prepare-test-env-ava.js';
+import { test } from '@agoric/zoe/tools/prepare-test-env-ava.js';
 // eslint-disable-next-line import/no-extraneous-dependencies
-import test from 'ava'; // TODO ses-ava doesn't yet have test.todo
 import '../../../../exported.js';
 
 import { AmountMath } from '@agoric/ertp';
--- a/patches/@endo+ses-ava+0.2.27.patch
+++ b/patches/@endo+ses-ava+0.2.27.patch
@@ -1,0 +1,269 @@
+diff --git a/node_modules/@endo/ses-ava/exported.js b/node_modules/@endo/ses-ava/exported.js
+index f4cba01..e69de29 100644
+--- a/node_modules/@endo/ses-ava/exported.js
++++ b/node_modules/@endo/ses-ava/exported.js
+@@ -1 +0,0 @@
+-import './src/types.js';
+diff --git a/node_modules/@endo/ses-ava/src/ses-ava-test.js b/node_modules/@endo/ses-ava/src/ses-ava-test.js
+index 36038ad..de65de7 100644
+--- a/node_modules/@endo/ses-ava/src/ses-ava-test.js
++++ b/node_modules/@endo/ses-ava/src/ses-ava-test.js
+@@ -1,18 +1,8 @@
+ // @ts-check
+ import 'ses';
+-import './types.js';
+ 
+ const { apply } = Reflect;
+ 
+-/**
+- * Just forwards to global `console.error`.
+- *
+- * @type {Logger}
+- */
+-const defaultLogger = (...args) => {
+-  console.error(...args);
+-};
+-
+ /**
+  * Determine if the argument is a Promise.
+  * (Approximately copied from promiseKit.js)
+@@ -24,21 +14,48 @@ const isPromise = maybePromise =>
+   Promise.resolve(maybePromise) === maybePromise;
+ 
+ /**
+- * @type {LogCallError}
++ * @typedef {(...args: unknown[]) => void} Logger
+  */
+-const logErrorFirst = (func, args, name, logger = defaultLogger) => {
++
++/**
++ * Calls `func(...args)` passing back approximately its outcome, but first
++ * logging any erroneous outcome to the `logger`, which defaults to
++ * `console.log`.
++ *
++ *    * If `func(...args)` returns a non-promise, silently return it.
++ *    * If `func(...args)` throws, log what was thrown and then rethrow it.
++ *    * If `func(...args)` returns a promise, immediately return a new
++ *      unresolved promise.
++ *       * If the first promise fulfills, silently fulfill the returned promise
++ *         even if the fulfillment was an error.
++ *       * If the first promise rejects, log the rejection reason and then
++ *         reject the returned promise with the same reason.
++ *
++ * The delayed rejection of the returned promise is an observable difference
++ * from directly calling `func(...args)` but will be equivalent enough for most
++ * purposes.
++ *
++ * TODO This function is useful independent of Ava, so consider moving it
++ * somewhere and exporting it for general reuse.
++ *
++ * @param {(...unknown) => unknown} func
++ * @param {unknown[]} args
++ * @param {string} name
++ * @param {Logger=} logError
++ */
++const logErrorFirst = (func, args, name, logError = console.error) => {
+   let result;
+   try {
+     result = apply(func, undefined, args);
+   } catch (err) {
+-    logger(`THROWN from ${name}:`, err);
++    logError(`THROWN from ${name}:`, err);
+     throw err;
+   }
+   if (isPromise(result)) {
+     return result.then(
+       v => v,
+       reason => {
+-        logger(`REJECTED from ${name}:`, reason);
++        logError(`REJECTED from ${name}:`, reason);
+         return result;
+       },
+     );
+@@ -47,7 +64,7 @@ const logErrorFirst = (func, args, name, logger = defaultLogger) => {
+   }
+ };
+ 
+-const testerMethodsWhitelist = [
++const overrideList = [
+   'after',
+   'afterEach',
+   'before',
+@@ -56,32 +73,52 @@ const testerMethodsWhitelist = [
+   'failing',
+   'serial',
+   'only',
+-  'skip',
+ ];
+ 
+ /**
+- * @param {TesterFunc} testerFunc
+- * @param {Logger} [logger]
+- * @returns {TesterFunc} Not yet frozen!
++* @callback BaseImplFunc
++* This is the function that invariably starts `t => {`.
++* Ava's types call this `Implementation`, but that's just too confusing.
++* @param {Assertions} t
++* @returns {unknown}
++*
++* @typedef {BaseImplFunc | Object} ImplFunc
++* @property {(...unknown) => string} [title]
++*
++* @callback TesterFunc
++* @param {string} title
++* @param {ImplFunc} [implFunc]
++* @returns {void}
++*/
++
++/**
++ * @template {TesterFunc} T
++ * @param {T} testerFunc
++ * @param {Logger} [logError]
++ * @returns {T} Not yet frozen!
+  */
+-const wrapTester = (testerFunc, logger = defaultLogger) => {
++const augmentLogging = (testerFunc, logError = console.error) => {
+   /** @type {TesterFunc} */
+-  const testerWrapper = (title, implFunc, ...otherArgs) => {
+-    /** @type {ImplFunc} */
++  const augmented = (title, implFunc, ...otherArgs) => {
+     const testFuncWrapper = t => {
+       harden(t);
+-      return logErrorFirst(implFunc, [t, ...otherArgs], 'ava test', logger);
++      return logErrorFirst(implFunc, [t, ...otherArgs], 'ava test', logError);
+     };
+     if (implFunc && implFunc.title) {
+       testFuncWrapper.title = implFunc.title;
+     }
+     return testerFunc(title, testFuncWrapper, ...otherArgs);
+   };
+-  return testerWrapper;
++  // re-use other properties (e.g. `.always`)
++  // https://github.com/endojs/endo/issues/647#issuecomment-809010961
++  Object.assign(augmented, testerFunc);
++  // @ts-expect-error cast
++  return augmented;
+ };
+ 
++// TODO check whether this is still necessary in Ava 4
+ /**
+- * The ava `test` function takes a callback argument of the form
++ * The Ava 3 `test` function takes a callback argument of the form
+  * `t => {...}`. If the outcome of this function indicates an error, either
+  * by throwing or by eventually rejecting a returned promise, ava does its
+  * own peculiar console-like display of this error and its stacktrace.
+@@ -105,25 +142,18 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
+  * that eventually rejects, the error is first sent to the `console` before
+  * propagating into `rawTest`.
+  *
+- * @param {TesterInterface} avaTest
+- * @param {Logger} [logger]
+- * @returns {TesterInterface}
++ * @template {TesterFunc} T Ava `test`
++ * @param {T} avaTest
++ * @param {Logger} [logError]
++ * @returns {T}
+  */
+-const wrapTest = (avaTest, logger = defaultLogger) => {
+-  const testerWrapper = /** @type {TesterInterface} */ (wrapTester(
+-    avaTest,
+-    logger,
+-  ));
+-  for (const methodName of testerMethodsWhitelist) {
+-    if (methodName in avaTest) {
+-      /** @type {TesterFunc} */
+-      const testerMethod = (title, implFunc, ...otherArgs) =>
+-        avaTest[methodName](title, implFunc, ...otherArgs);
+-      testerWrapper[methodName] = wrapTester(testerMethod);
+-    }
++const wrapTest = (avaTest, logError = console.error) => {
++  const sesAvaTest = augmentLogging(avaTest, logError);
++  for (const methodName of overrideList) {
++    sesAvaTest[methodName] = augmentLogging(avaTest[methodName]);
+   }
+-  harden(testerWrapper);
+-  return testerWrapper;
++  harden(sesAvaTest);
++  return sesAvaTest;
+ };
+-// harden(wrapTest);
++harden(wrapTest);
+ export { wrapTest };
+diff --git a/node_modules/@endo/ses-ava/src/types.js b/node_modules/@endo/ses-ava/src/types.js
+deleted file mode 100644
+index 9bddf17..0000000
+--- a/node_modules/@endo/ses-ava/src/types.js
++++ /dev/null
+@@ -1,72 +0,0 @@
+-// @ts-check
+-
+-/**
+- * @typedef {(...args: unknown[]) => void} Logger
+- */
+-
+-/**
+- * @callback LogCallError
+- *
+- * Calls `thunk()` passing back approximately its outcome, but first
+- * logging any erroneous outcome to the `logger`, which defaults to
+- * `console.log`.
+- *
+- * If thunk returns a non-promise, silently return it.
+- * If thunk throws, log what was thrown and then rethrow it.
+- * If thunk returns a promise, immediately return a new unresolved promise.
+- * If the first promise fulfills, silently fulfill the returned promise even if
+- * the fulfillment was an error.
+- * If the first promise rejects, log the rejection reason and then reject the
+- * returned promise with the same reason.
+- * The delayed rejection of the returned promise is an observable difference
+- * from directly calling `thunk()` but will be equivalent enough for most
+- * purposes.
+- *
+- * TODO This function is useful independent of ava, so consider moving it
+- * somewhere and exporting it for general reuse.
+- *
+- * @param {(...unknown) => unknown} func
+- * @param {unknown[]} args
+- * @param {string} name
+- * @param {Logger=} logger
+- */
+-
+-/**
+- * Simplified form of ava's types.
+- * TODO perhaps just import ava's type declarations instead
+- * TODO reconcile also with types and API defined in avaAssertXS.js
+- *
+- * @typedef {Object} Assertions
+- * @property {(actual: unknown, message?: string) => void} assert
+- * // TODO is, deepEqual, truthy, falsy, etc...
+- */
+-
+-/**
+- * @callback BaseImplFunc
+- * This is the function that invariably starts `t => {`.
+- * Ava's types call this `Implementation`, but that's just too confusing.
+- *
+- * @param {Assertions} t
+- * @returns {unknown}
+- *
+- * @typedef {BaseImplFunc | Object} ImplFunc
+- * @property {(...unknown) => string} [title]
+- *
+- * @callback TesterFunc
+- * @param {string} title
+- * @param {ImplFunc} implFunc
+- * @returns {void}
+- *
+- * @typedef {Object} TesterProperties
+- * @property {TesterFunc} after
+- * @property {TesterFunc} afterEach
+- * @property {TesterFunc} before
+- * @property {TesterFunc} beforeEach
+- * @property {TesterFunc} cb
+- * @property {TesterFunc} failing
+- * @property {TesterFunc} serial
+- * @property {TesterFunc} only
+- * @property {TesterFunc} skip
+- *
+- * @typedef {TesterFunc & TesterProperties} TesterInterface
+- */
--- a/packages/notifier/test/test-publish-kit.js
+++ b/packages/notifier/test/test-publish-kit.js
@@ -124,13 +124,13 @@ test('makePublishKit', async t => {
     'failure should not be allowed after finalization',
   );
   await t.throwsAsync(
-    // @ts-ignore known to be promise version of PublicationList; expect after https://github.com/Agoric/agoric-sdk/pull/5774/
+    // @ts-expect-error known to be promise version of PublicationList
     subFinal.tail,
     undefined,
     'tail promise of final result should be rejected',
   );
   await t.throwsAsync(
-    // @ts-ignore known to be promise version of PublicationList; expect after https://github.com/Agoric/agoric-sdk/pull/5774/
+    // @ts-expect-error known to be promise version of PublicationList
     subscriber.subscribeAfter(subFinal.publishCount),
     undefined,
     'subscribeAfter(finalPublishCount) should be rejected',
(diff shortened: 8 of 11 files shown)


patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,76 @@
...
+ };
+@@ -105,9 +109,9 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
+  * that eventually rejects, the error is first sent to the `console` before
+  * propagating into `rawTest`.
+  *
+- * @param {TesterInterface} avaTest
++ * @param {import('ava').TestInterface} avaTest
+  * @param {Logger} [logger]
+- * @returns {TesterInterface}
++ * @returns {TestInterface }
```
kriskowal:
```suggestion
+ * @returns {TestInterface}
```

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,76 @@
...
++    return testerFunc(title, ...otherArgs);
+   };
+   return testerWrapper;
+ };
+@@ -105,9 +109,9 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
+  * that eventually rejects, the error is first sent to the `console` before
+  * propagating into `rawTest`.
+  *
+- * @param {TesterInterface} avaTest
++ * @param {import('ava').TestInterface} avaTest
```
kriskowal: Are we confident that our emulation satisfies the entire ava interface now? Or, is this a convenient fiction that improves the debug experience?

kriskowal (approved):

patches/@endo+ses-ava+0.2.27.patch
turadg: You're right we don't. That's why this still returns `TestInterface`. But the `avaTest` param is indeed from Ava.

Marking this as Draft because my refactor broke some case.

turadg:
@kriskowal why do we have a whitelist of what to allow through? shouldn't it be an "override" list that's passed over after copying everything? IIUC "ses-ava" package is to provide SES-happy console output, not for actual security of the test run.

kriskowal:
> @kriskowal why do we have a whitelist of what to allow through? shouldn't it be an "override" list that's passed over after copying everything? IIUC "ses-ava" package is to provide SES-happy console output, not for actual security of the test run.

I should in this case defer to @erights, but this may have been born of abundance-of-caution or force-of-security-habit and eligible for a rethink.

turadg:
I looked into the [open issues for ses-ava](https://github.com/endojs/endo/issues?q=is%3Aissue+is%3Aopen+ses-ava) and these would be solved by having the "override" list instead of an "allow" list:
https://github.com/endojs/endo/issues/645
https://github.com/endojs/endo/issues/647

But I don't understand why this fence was put here! @erights ?

I also don't know the symptoms of not using `ses-ava` at all. Do they persist in Ava 4?

turadg:
Also, since the `prepare-test-env` is mutating the env, why not mutate Ava `test` itself? Status quo is we have these `prepare-test-env-ava.js` files that have both side-effects AND exports. Because they have side-effects they have to go before other imports. But then when they use relative paths our lint rules complain that it should be lower, so we have to suppress:
```
/* eslint-disable import/order */
import { test } from '../tools/prepare-test-env-ava.js';
```
There are 36 of those.

How about this instead?
```
import '../tools/prepare-test-env-ava.js';

import test from 'ava';
```

packages/run-protocol/test/test-gov-collateral.js
```diff
@@ -586,4 +586,4 @@ test('Committee can raise debt limit', async t => {
   });
 });
 
-// test.todo('users can open vaults');
+test.todo('users can open vaults');
```
erights, line 589: What is the whole `test` vs `anyTest` dance at the beginning of this file about? I don't get it. Can we simplify this to follow the pattern in the rest of this PR?

packages/SwingSet/test/test-message-patterns.js
```diff
@@ -2,13 +2,8 @@
...
 
-// `test.serial` does not yet seem compatible with ses-ava
-// See https://github.com/endojs/endo/issues/647
-// TODO restore
-// import { test } from '../tools/prepare-test-env-ava.js';
-import '../tools/prepare-test-env.js';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import test from 'ava';
+// eslint-disable-next-line import/order -- has side-effects AND exports
+import { test } from '../tools/prepare-test-env-ava.js';
```
erights, line 6: I'm fine with this change, but confused about your intentions for this PR. At https://github.com/Agoric/agoric-sdk/pull/5774#issuecomment-1186236576 you seem to suggest moving in exactly the opposite direction. Preferring the status quo over the change you're making here.

Regarding the need for the `eslint-disable`, could we instead have all of these be `@foo/something` imports, probably from some (new?) package in endo?

packages/solo/test/test-home.js
```diff
@@ -1,12 +1,6 @@
 /* global process */
 
-// `test.after.always` does not yet seem compatible with ses-ava
-// See https://github.com/endojs/endo/issues/647
-// TODO restore
-// import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
-import '@agoric/swingset-vat/tools/prepare-test-env.js';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import test from 'ava';
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
```
erights, line 3: Again, seems to go in the opposite direction to your suggestion at https://github.com/Agoric/agoric-sdk/pull/5774#issuecomment-1186236576 . I actually prefer the new form above over your suggestion there, but I don't understand which direction you want this PR to take things.

I mildy prefer the form above simply because it is shorter and bundles the two relevant things together. As separate imports, the dependency between them is less clear, and more of a refactoring hazard.

Why are you using the `@foo/something` for here, thereby avoiding the need for `eslint-disable`, but not in other places?

packages/ui-components/test/display/test-display.js
```diff
@@ -1,7 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
```
erights, line 1: Why is this `eslint-disable` needed?

packages/zoe/src/contractFacet/zcfZygote.js
```diff
@@ -319,7 +319,6 @@ export const makeZCFZygote = (
     /** @type {SetTestJig} */
     setTestJig: (testFn = () => ({})) => {
       if (testJigSetter) {
-        console.warn('TEST ONLY: capturing test data', testFn);
```
erights, line 322: Happy to see this go. But do we know why it was there?

packages/zoe/test/unitTests/contracts/loan/test-addCollateral.js
```diff
@@ -1,8 +1,7 @@
 // @ts-nocheck
```
erights, line 1: Surprised to see this in a PR from you ;)

Likewise with several other occurrences below.

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
+diff --git a/node_modules/@endo/ses-ava/exported.js b/node_modules/@endo/ses-ava/exported.js
+index f4cba01..e69de29 100644
+--- a/node_modules/@endo/ses-ava/exported.js
++++ b/node_modules/@endo/ses-ava/exported.js
```
erights, line 4: Oh geez, an endo patch. Is there (or will there be) a corresponding endo PR that, once integrated, would enable us to remove this patch file? I would rather first review this as an endo PR before reviewing it as an endo patch to agoric-sdk.

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+@@ -24,21 +14,46 @@ const isPromise = maybePromise =>
+   Promise.resolve(maybePromise) === maybePromise;
+ 
+ /**
+- * @type {LogCallError}
++ * @typedef {(...args: unknown[]) => void} Logger
++ */
++
++/**
++ * Calls `thunk()` passing back approximately its outcome, but first
```
erights: It looks like the following terminology problem was already there. But this PR might still be a good opportunity to fix it.

The term `thunk` here seems to refer to the `func` parameter below. If true, they should both use the same name. The CS term `thunk` that I'm familiar with refers only to a function/closure with zero arguments, which `func` is not. So if they are referring to the same thing, `func` is a better choice.

Hmm. I think I see how this terminology problem happened (and I think I may have perpetrated it). The `func` together with the `args` may have previously been packaged together as one `thunk` argument that was invoked with no args. But that always created a yet deeper stack exactly in the place where we're trying to generate good diagnostic info. So by separating them in this way, we're able to make the diagnostic stacks clearer. Yeah, I think I remember doing this. Sorry I did not clean up the comment.

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+   'only',
+-  'skip',
+ ];
+ 
+ /**
+- * @param {TesterFunc} testerFunc
+- * @param {Logger} [logger]
+- * @returns {TesterFunc} Not yet frozen!
++ * @template {Function} T
++ * @param {T} testerFunc
```
erights: I don't understand replacing the specific type `TesterFunc` with a generic type parameter. The returned `augmented` func assumes that the `testerFunc`'s second parameter is an `ImplFunc`. The `augmented` function itself is supposed to be a compat replacement for `testerFunc` only under this assumption, since it has its own `implFunc` parameter, which it replaces with `testFuncWrapper`.

If `testerFunc` is anything else, then `wrapTester(testerFunc, ...)` would be wrong. The method names of the `overrideList`/`testerMethodsWhitelist` are supposed to enumerate only methods of `test` which conform to these assumptions about `TesterFunc`.

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+     if (implFunc && implFunc.title) {
+       testFuncWrapper.title = implFunc.title;
+     }
+     return testerFunc(title, testFuncWrapper, ...otherArgs);
+   };
+-  return testerWrapper;
++  // re-use other properties (e.g. `.always`)
++  // https://github.com/endojs/endo/issues/647#issuecomment-809010961
++  Object.assign(augmented, testerFunc);
++  // @ts-expect-error cast
```
erights, line 144: Is this needed because you're (inappropriately?) using a generic type parameter rather than `TesterFunc`?

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+     };
+     if (implFunc && implFunc.title) {
+       testFuncWrapper.title = implFunc.title;
+     }
+     return testerFunc(title, testFuncWrapper, ...otherArgs);
+   };
+-  return testerWrapper;
++  // re-use other properties (e.g. `.always`)
++  // https://github.com/endojs/endo/issues/647#issuecomment-809010961
++  Object.assign(augmented, testerFunc);
```
erights, line 143: An interesting choice. This copies all the old properties of `testerFunc` onto `augmented` whether or not they have been augmented. Whereas the original sought to omit methods that should have been augmented but were not, this will preserve the original even though we had failed to augment it. This is plausible, but we should discuss.

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
++  // @ts-expect-error cast
++  return augmented;
+ };
+ 
++// TODO check whether this is still necessary in Ava 4
```
erights, line 148: Do you plan to check before this PR is R4R?

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+  * by throwing or by eventually rejecting a returned promise, ava does its
+  * own peculiar console-like display of this error and its stacktrace.
+@@ -105,25 +123,17 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
+  * that eventually rejects, the error is first sent to the `console` before
+  * propagating into `rawTest`.
+  *
+- * @param {TesterInterface} avaTest
+- * @param {Logger} [logger]
+- * @returns {TesterInterface}
++ * @template {Function} T Ava `test`
```
erights: This `@template` line has more syntax than I recognize. What does it mean?

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+-    }
++const wrapTest = (avaTest, logError = console.error) => {
++  const sesAvaTest = augmentLogging(avaTest, logError);
++  for (const methodName of overrideList) {
++    sesAvaTest[methodName] = augmentLogging(avaTest[methodName]);
+   }
+-  harden(testerWrapper);
+-  return testerWrapper;
++  harden(sesAvaTest);
++  return sesAvaTest;
```
erights, line 187: IIUC, you're still returning a new augmented `sesAvaTest` rather than mutating the original `avaTest`. Yes? I don't mind. But didn't your PR comment suggest mutating the original in place?

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+ };
+-// harden(wrapTest);
+ export { wrapTest };
```
erights, line 191: If it is exported, it should be hardened. Unless we're purposely avoiding this because this code runs before `lockdown`? Are we? Does it?

patches/@endo+ses-ava+0.2.27.patch
```diff
@@ -0,0 +1,250 @@
...
+- * This is the function that invariably starts `t => {`.
+- * Ava's types call this `Implementation`, but that's just too confusing.
+- *
+- * @param {Assertions} t
+- * @returns {unknown}
+- *
+- * @typedef {BaseImplFunc | Object} ImplFunc
+- * @property {(...unknown) => string} [title]
+- *
+- * @callback TesterFunc
```
erights, line 252: Yeah, this (and its dependents) are the ones I am skeptical you can omit while still using types to describe the augmentation.

erights (changes requested):
Since I see @kriskowal already approved, I'm changing my comments to Request changes as an interlock, even though it is still not R4R

erights:
> @kriskowal why do we have a whitelist of what to allow through? shouldn't it be an "override" list that's passed over after copying everything? IIUC "ses-ava" package is to provide SES-happy console output, not for actual security of the test run.

Yeah, it isn't for security. It is to keep track of which methods we've examined to determine whether to augment them or not. Perhaps we should have two whitelists: The ones we know we want to augment, and the ones we know we don't. It is an interesting policy choice what to do with the remainder: Do we preserve their existing behavior (as your code does) until we decide, or do we omit them until we decide?

erights:
> Also, since the `prepare-test-env` is mutating the env, why not mutate Ava `test` itself? Status quo is we have these `prepare-test-env-ava.js` files that have both side-effects AND exports. Because they have side-effects they have to go before other imports. But then when they use relative paths our lint rules complain that it should be lower, so we have to suppress:
> 
> ```
> /* eslint-disable import/order */
> import { test } from '../tools/prepare-test-env-ava.js';
> ```
> 
> There are 36 of those.
> 
> How about this instead?
> 
> ```
> import '../tools/prepare-test-env-ava.js';
> 
> import test from 'ava';
> ```

Whether or not we mutate `test` in place, shouldn't we still prefer the former to the latter because it is shorter (less big of a deal) and bundles related issues together so it is less of a refactoring hazard?

Also, wouldn't the first line of the latter still need to come first? If it does and still uses the '`../tools/...'` form you show above, wouldn't it still need the `eslint-disable`?

erights:
> I also don't know the symptoms of not using ses-ava at all. Do they persist in Ava 4?

The consequence is that the seeming console output is neither from the real console nor from the replacement that `lockdown` installs instead. Thus it lacks the crucial diagnostic info that's especially important when we're iterating the development of our test.

turadg:
> Yeah, it isn't for security.

Ok good. So the purpose of the "ses-ava" package isn't for security, just to accommodate SES's console handling when testing with Ava.

> It is to keep track of which methods we've examined to determine whether to augment them or not. Perhaps we should have two whitelists: The ones we know we want to augment, and the ones we know we don't. It is an interesting policy choice what to do with the remainder: Do we preserve their existing behavior (as your code does) until we decide, or do we omit them until we decide?

I can't think of any reason to omit any part of the Ava test framework. The user shouldn't have to lose functionality because of SES's console requirements. (The lack of `.todo` was a surprise which broke my build and my focus.)  For runtime, one list suffices: modify these methods so work better in SES. The rest should pass through. If there is any value in keeping track of "we looked into this one and it doesn't need augmentation", that's a documentation concern. A list like that in a comment above the override list would be fine by me, though I don't see the value in that. I expect whenever there's a problem that it'll get reported and then belong on the override list. If something is known to need overriding and that's not working yet, that could go into the override list, commented out with an explanation (and ideally a link to a ticket).

turadg:
> > ```
> > /* eslint-disable import/order */
> > import { test } from '../tools/prepare-test-env-ava.js';
> > ```
> >   
> > How about this instead?
> > ```
> > import '../tools/prepare-test-env-ava.js';
> > 
> > import test from 'ava';
> > ```
> 
> Whether or not we mutate `test` in place, shouldn't we still prefer the former to the latter because it is shorter (less big of a deal) and bundles related issues together so it is less of a refactoring hazard?

I think those are both offset by the value of consistent Ava tests, regardless of whether SES env is prepared. What's "related" is "prepare env" and when you prepare env or not, it should not change the ergonomics of writing a test. By bundling shims which have side effects with exports of symbols to use, I argue that this is more of a refactoring hazard.

> Also, wouldn't the first line of the latter still need to come first? If it does and still uses the '`../tools/...'` form you show above, wouldn't it still need the `eslint-disable`?

Good question. The way the rule works, if you import without a symbol it's assumed to have side effects so it's exempt from ordering rules.

Note that this PR isn't requesting the change. I asked to understand why it's like this in the first place and to see if we can agree on the proposed way. (I've since seen in PR history of Endo that it used to be the way I'm proposing.) If we do, I'll file a ticket to do that after MN-1.

But my immediate goal with this PR is to make SES-Ava conform to the Ava API.

patches/@endo+ses-ava+0.2.27.patch
turadg: I don't know why it was commented out but that seems plausible. I don't have time to research further but I can restore the line if will get this approved :)

patches/@endo+ses-ava+0.2.27.patch
turadg: Correct, in this PR not mutating the original. [Clarified](https://github.com/Agoric/agoric-sdk/pull/5774#issuecomment-1186539828)

patches/@endo+ses-ava+0.2.27.patch
turadg: `@template T` you're familiar with. The `{Function}` means T has to be a subtype of `Function`. That's so TS knows that the calls below are valid. The string after `T` is a description for display.

patches/@endo+ses-ava+0.2.27.patch
turadg: Nope. Ava 4 is on hold to at least after MN-1 https://github.com/Agoric/agoric-sdk/issues/4895

packages/zoe/src/contractFacet/zcfZygote.js
turadg: Nope. Given how old it is I would guess it was added around when test jigs were new, to notify the dev about it.

packages/zoe/test/unitTests/contracts/loan/test-addCollateral.js
turadg: ha, well, a few reasons:
- out of scope for PR
- not a priority before MN-1 to add types to code we're not working in
- types in tests are only for the test writer since they don't provide any additional confidence in production

patches/@endo+ses-ava+0.2.27.patch
turadg: > Is there (or will there be) a corresponding endo PR that, once integrated, would enable us to remove this patch file?

Yes there will be. I think that goes for every patch file Agoric can push upstream. So I think the existence of this file is documentation enough.

> I would rather first review this as an endo PR before reviewing it as an endo patch to agoric-sdk.

My job right now is to ship MN-1. That ses-ava breaks the Ava API is slowing me down. Effort fixing it in Endo proper should not be spent until after shipping.

packages/ui-components/test/display/test-display.js
turadg: Because the import in the next line isn't declared in the package dependencies.

packages/solo/test/test-home.js
turadg: I made this change to be consistent with the other code.  I do think we'd be better off with `import test from 'ava'` but that's [out of scope](https://github.com/Agoric/agoric-sdk/pull/5774#issuecomment-1186539828)).

> Why are you using the @foo/something for here … but not in other places?

I didn't change any import paths so not really my decision. But I think the relative ones are because what's imported is coming from the same package.

packages/SwingSet/test/test-message-patterns.js
turadg: > Regarding the need for the eslint-disable, could we instead have all of these be @foo/something imports, probably from some (new?) package in endo?

Not sure I follow but I'm sure it's out of scope for this PR. I'm just trying to make the `test` function work consistently. While I would like to refactor it out of the `prepare-` files that's for another day.

packages/run-protocol/test/test-gov-collateral.js
turadg: It's a cast so that `t.context` is statically typed. E.g. in this file `t.context.zoe` is known to be `ZoeService`.

Screen Shot 2022-07-17 at 8 10 33 AM
.

patches/@endo+ses-ava+0.2.27.patch
turadg: Confident now that it satisfies because it passes through whatever it doesn't override. I typed the param as generic so that it wraps whatever version of Ava was passed in.

patches/@endo+ses-ava+0.2.27.patch
turadg: If you want to now feel free to submit a suggestion through Github UI or push a commit on the branch.

patches/@endo+ses-ava+0.2.27.patch
turadg: I went with any `Function` because it works. I don't see the value in the additional constraints. If this is a blocker let me know.

patches/@endo+ses-ava+0.2.27.patch
turadg: Possibly. The error the cast is covering up,
```
Type '(title: any, implFunc: any, ...otherArgs: any[]) => any' is not assignable to type 'T'.
  '(title: any, implFunc: any, ...otherArgs: any[]) => any' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'Function'.
```
But again I don't see the value in being precise in this particular code. Let me know if it's a requirement to merge into agoric-sdk.

patches/@endo+ses-ava+0.2.27.patch
turadg: I assert this is the ideal behavior. The job of the wrapper is to fix console output where we know it's a problem. It shouldn't break anything it doesn't know about.

patches/@endo+ses-ava+0.2.27.patch
turadg: The important types here are what `wrapTest` returns and it's as simple as "whatever it got" because it shouldn't be changing the API of Ava.

packages/run-protocol/test/test-gov-collateral.js
erights: Wow, that's good to know. Resolving.

packages/SwingSet/test/test-message-patterns.js
erights: Gotcha. Resolving.

packages/solo/test/test-home.js
erights: ok. Resolving.

packages/ui-components/test/display/test-display.js
erights: Should it be?

packages/zoe/src/contractFacet/zcfZygote.js
erights: ok. Resolving.

packages/zoe/test/unitTests/contracts/loan/test-addCollateral.js
erights: It is interesting to see how our coding standards for test code increasingly differs from our coding standards for production code. Another is that we should not impose our top-level-await rule on test code. Will we maintain divergent lint settings?

In any case, for this PR no change suggested. Resolving.

patches/@endo+ses-ava+0.2.27.patch
erights: Sounds good to me. How do we remember to get back to this after MN-1? Is there an issue?

Resolving.

patches/@endo+ses-ava+0.2.27.patch
erights: It is not a blocker. I recognize that you understand our types much better than I. So when I'm puzzled by your use of types, I'd at least like to learn from you. 

Here, I am genuinely confused about why you think this change is an improvement. The code being described by these types relies on the assumption that the second parameter of the testerFunc is an ImplFunc, in order to replace this ImplFunc by wrapping. If we didn't have a type language to describe it, I'd feel compelled to write prose describing that. But we do have a type language that is good at saying things like this. Why not use it?

I suppose the blocker is that if we do not explain this constraint using the type language, then we do need explicit prose describing it well. Why wouldn't we?

patches/@endo+ses-ava+0.2.27.patch
erights: I don't care about the type error. I do care about clarity to programmers reading and maintaining this code.

patches/@endo+ses-ava+0.2.27.patch
erights: I am ok with this rationale. Resolving.

patches/@endo+ses-ava+0.2.27.patch
erights: thanks. Resolving.

patches/@endo+ses-ava+0.2.27.patch
erights: Ah. Although I knew that extra tokens after jsdoc/ts declarations are assumed to be prose descriptions, somehow that did not occur to me here. It seemed like more type notation that I didn't know. Thanks for clarifying. Resolving.

patches/@endo+ses-ava+0.2.27.patch
erights: Please restore and uncomment out the harden call.

patches/@endo+ses-ava+0.2.27.patch
erights: I do not understand your reply.

erights:
> If there is any value in keeping track of "we looked into this one and it doesn't need augmentation", that's a documentation concern. 

I am ok with this, and with your other points above that I did not comment on.

patches/@endo+ses-ava+0.2.27.patch
erights: Manually editing a patch seems a bit much, so https://github.com/endojs/endo/pull/1234 instead ;)

Resolving here.

packages/ui-components/test/display/test-display.js
turadg: Out of scope for this PR but I don't think necessarily so. That would be most expedient.

It's also worth considering that whoever added the dependence on `@agoric/swingset-vat/tools/prepare-test-env-ava.js` didn't want `ui-components` to depend on `swingset-vat` even for dev. The alternative is to put that file somewhere else. It's dependency tree can be satisfied by Endo and `vat-data`.

packages/zoe/test/unitTests/contracts/loan/test-addCollateral.js
turadg: > Will we maintain divergent lint settings?

It's managed by a rule tightening for production code:
https://github.com/Agoric/agoric-sdk/blob/9410a8500f6a30906c9e3359ece0e5d8dfd965ad/.eslintrc.cjs#L73-L91

> our coding standards for test code increasingly differs from our coding standards for production code

Put more exactly, some requirements for production are relaxed for non-production. I think this is appropriate.

patches/@endo+ses-ava+0.2.27.patch
turadg: I would think the presence of the patchfile would be a sufficient reminder but I've https://github.com/endojs/endo/issues/1235

patches/@endo+ses-ava+0.2.27.patch
turadg: > confused about why you think this change is an improvement

I don't necessarily think every line is an improvement. I think the overall PR is an improvement.

Is it a regression? It's a loss of type information but it's now _less wrong_ because not every tester function takes an impl. E.g. `test.todo` https://github.com/endojs/endo/issues/645

> we do have a type language that is good at saying things like this. Why not use it?

Because of the cost/benefit of making it accurate and not misleading. With this PR the costs of not doing it have been going up so I've gone ahead and re-added TesterFunc and made `implFunc` optional.

patches/@endo+ses-ava+0.2.27.patch
turadg: > care about clarity to programmers reading and maintaining this code.

Here here! :)

patches/@endo+ses-ava+0.2.27.patch
turadg: Basically https://github.com/Agoric/agoric-sdk/pull/5774/#discussion_r923405025

packages/ui-components/test/display/test-display.js
erights: Good. Resolving.

patches/@endo+ses-ava+0.2.27.patch
erights: Thx. Resolving

patches/@endo+ses-ava+0.2.27.patch
erights: Aha! I think I understand our misunderstanding. You say 
> not every tester function takes an impl. E.g. `test.todo`

But I omitted `todo` from the whitelist, and I see you still omit it from the `overrideList`. Then it occurred to me: I am using `TesterFunc` only to describe the methods on this list. You (and possibly TS itself?) are interpreting it as a description of all methods on `test`, which it clearly is not. If `todo` isn't described by `TesterFunc` then it should not be on this list anyway.

Making `implFunc` optional seems like a solution only because of our misunderstanding.

How should this be expressed in the type system? I thought I had done that successfully by enumerating in `TesterProperties` exactly the methods I put into the list. `todo` is absent both from the list and from the `TesterProperties` type. 

(Having the `TesterProperties` type track the list is a redundancy that creates a refactoring hazards, which I would also like to avoid if possible. But let's take that question separately.)

patches/@endo+ses-ava+0.2.27.patch
erights: Good. Answered there.

Btw, a github weirdness that seems new to me: When I click or even hover over the link above I see that github rendered the markdown without the url fragment, thereby linking only to the PR as a whole, which confused me. So I selected "Edit" from the triple-dots of this comment to see the source, then copied the url source for this link into the url bar of a separate browser tab, and viola, I saw the comment you refer to.

turadg: ah, that is weird. seems like a bug

patches/@endo+ses-ava+0.2.27.patch
turadg: Yeah, that's it. Do statically type this we'd have to annotate that `todo` is different from the others. I don't think this is worth it. I'm just trying to make `test` work so I can use it to develop. This PR makes it work. I greatly value code maintainability and I think this is maintainable enough. The type system doesn't have to cover every expression.

I don't totally know the refactoring hazards you're trying to avoid but I suspect that whatever time we're saving on that end we're simply spending now, when time is more expensive.

erights: I hear the frustration. If this were a PR against genuine source code I would feel uncomfortable rushing approval on this basis. However, here I'll take advantage of the fact that the changes I object to are on a patch. When it is time to make a corresponding change against the original sources on endo, please make me a reviewer and let's resolve this misunderstanding then.

erights (approved):
Approving in a rush as explained at https://github.com/Agoric/agoric-sdk/pull/5774#discussion_r923832098

patches/@endo+ses-ava+0.2.27.patch
turadg: Sounds like a good plan. =)

PR #5777, opened 2022-07-17
Refactor liveslots options handling

### LINKED ISSUE #5619: Growing collection of liveslots options should be passed as an options bag

The number of parameters to `makeLiveSlots`, `setBundle`, and various other functions involved in the creation or startup of vats keeps increasing as we identify additional configurations options that need to be passed in.  Each time this involves changes to many different points in the code that don't actually care about the options but which need to plumb them through.  These parameters should be replaced with an options bag object so that all the intermediaries can be blind to churn in the options set, requiring that the addition, removal, or modification of a liveslots option need only touch the point of injection and the point of use.

## Description of the Design

Currently there are parameters for virtual object cache size, disavow enablement, relaxed durability handling, and soon message budget.  These should be replaced with a singular options bag object throughout.  Because of the requirement to pass these parameters through the signal path between the kernel process and the vat process, this options bag will be subject to a requirement that everything in it be JSON serializable, but we already have that effectively because the individual options are already subject to such a requirement.

## Security Considerations

None to speak of.  This is a purely internal refactoring.

## Test Plan

This should not introduce any new behaviors, so it should be sufficient that all the existing tests continue to pass.

FUDCo:
This will enable future changes and additions to liveslots options to be made while touching far fewer places in the code.

Closes #5619

### DIFF
--- a/packages/SwingSet/misc-tools/replay-transcript.js
+++ b/packages/SwingSet/misc-tools/replay-transcript.js
@@ -156,6 +156,7 @@ async function replay(transcriptFile) {
         vatID,
         vatSourceBundle,
         managerOptions,
+        {},
         vatSyscallHandler,
       );
       console.log(`manager created`);
--- a/packages/SwingSet/src/kernel/vat-loader/manager-factory.js
+++ b/packages/SwingSet/src/kernel/vat-loader/manager-factory.js
@@ -74,7 +74,12 @@ export function makeVatManagerFactory({
   }
 
   // returns promise for new vatManager
-  function vatManagerFactory(vatID, managerOptions, vatSyscallHandler) {
+  function vatManagerFactory(
+    vatID,
+    managerOptions,
+    liveSlotsOptions,
+    vatSyscallHandler,
+  ) {
     validateManagerOptions(managerOptions);
     const {
       managerType = defaultManagerType,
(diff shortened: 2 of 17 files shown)


warner (approved):
That looks a lot better, thanks.

My one question mark is whether we should put `liveSlotsOptions` inside `managerOptions`, or have it be a peer as you did here. The way we handle options in `vat-loader.js` is due for an overhaul, in particular the way `allowedDynamicOptions`/`allowedStaticOptions` coming in from the parent vat/vat-admin (or `config`) gets translated into `managerOptions` (and now `liveSlotsOptions`) before being sent to the specific manager. We're currently allowing the parent to control things that they probably shouldn't, and we have two/three separate data structures all claiming the name "Manager Options".

Probably not anything to change now, but we should re-read `vat-loader.js` some time soon and think about making it a bit more deliberate.

FUDCo:
I pondered that very question as I was doing this.  My take was that the various options initially flow from the `config` object, so embedding `liveSlotsOptions` inside `managerOptions` would either mean giving the `config` options a hierarchy that the configuration author would need to understand, which seemed like an unnecessary imposition, or having the code that processes the `config` options rewrite them into a new shape, which seemed weird and uncouth.  As coded now there's a relatively simply partitioning into "these things go over here, those things go over there" pattern, which seems reasonable for the time being.

Issue #5759, opened 2022-07-13
support virtual objects that manage ephemeral state


turadg:
### Changing requirements should be efficient

When developing contracts, requirements may evolve and it helps developers to minimize the effort necessary to implement changes. One such requirement is the durability of data. It can be:
- *H*eap – stored in RAM and paid for
- *V*irtual - in RAM while needed but can be safely paged out and back
- *D*urable - virtual but persists through on-chain upgrade of contracts on chain

In the Virtual Object Manager (VOM) now transitioning a value from V→D or D→V is pretty declarative. Change a function name or add an option flag. But changing →H or H→ is laborious. See https://github.com/Agoric/agoric-sdk/pull/5736/ commits example.

### The resulting code should also be clean.

The movements above resulted in the simple
https://github.com/Agoric/agoric-sdk/blob/39fe09285a230a33966b23f3ab3a1f61127f0a64/packages/run-protocol/src/vaultFactory/vaultManager.js#L529

turning into https://github.com/Agoric/agoric-sdk/blob/e4c837a47ab31c31628ff8a1a736aa3a139f0044/packages/run-protocol/src/vaultFactory/vaultManager.js#L557-L561

### Representatives cannot reveal GC

https://github.com/Agoric/agoric-sdk/pull/5758 tried to solve the above problems by letting some heap data onto the `state` object. The problem with this is if that userspace could tell when GC happens because one Representative will have those properties, while later Representatives will not. 

## Description of the Design

A design that seems to satisfy the above requirements is to continue to use the WeakMap pattern as in https://github.com/Agoric/agoric-sdk/blob/39fe09285a230a33966b23f3ab3a1f61127f0a64/packages/run-protocol/src/vaultFactory/vaultManager.js#L140-L147

But instead of making the module responsible for the map and each method responsible for pulling the `ephemera` object out of it, have the VOM provide `ephemera` in the context next to `state`. So the ugly refactor above could be back to one line,

```js
  getGovernedParams: ({ ephemera }) => ephemera.factoryPowers.getGovernedParams();
```

TBD how the `ephemera` object gets initialized. Some requirements: 
- It needs access to the kind constructor args. 
-`finished`, like other context consumers needs `ephemera`.
[...]

Issue #4653, opened 2022-02-23
Expose key Build Boost economic data via subscription


Tartuffo:
This data can be used in Blockchain Explorers and via sites like The Graph for the community to get key information about the state of the RUN Protocol components.

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose from getRUN:

1. [ ] Per Contract (`assetNotifier`)
    1. [x] Net amount minted (`totalDebt`)
    2. [ ] Total collateral
    3. Governance parameters (see #5268)
        1. debt limit
        2. allowed ratio
        3. interest rate
4. [x] Per Account
    1. [x] Locked BLD amount (`locked`)
    2. [x] Debt (`debtSnapshot`)
    5. [x] Interest rate (?) `interestRate` (but see #4540)

 - [ ] publicFacet method to get `assetNotifier`

dckc:
The current code (52f60eb) has an `assetNotifier` on each offer result:

```ts
const assetNotifer: Notifier<{
    compoundedInterest: Ratio;
    interestRate: Ratio;
    latestInterestUpdate: bigint;
    totalDebt: Amount<"nat">;
}>
```

There's no publicFacet method to get that assetNotifier; the only way to get it is as a result of opening an account. That seems like a bug.

The Per Account notifier is updated as follows:

```js
    /** @type {VaultUIState} */
    const result = harden({
      // TODO move manager state to a separate notifer https://github.com/Agoric/agoric-sdk/issues/4540
      interestRate: manager.getInterestRate(),
      liquidationRatio: manager.getMintingRatio(),
      debtSnapshot: { debt, interest },
      locked: getCollateralAmount(),
      // newPhase param is so that makeTransferInvitation can finish without setting the vault's phase
      // TODO refactor https://github.com/Agoric/agoric-sdk/issues/4415
      vaultState: newActive ? 'active' : 'closed',
    });
```

The `VaultUIState` type is not declared, so it defaults to `any`. :-/ (more interest on technical debt:  #3916 )

dckc:
@Tartuffo @rowgraus what is an "interest rate curve"?

dckc:
And what does "Distribution of utilization of loan amount" mean, @rowgraus ? @Tartuffo ?

dckc:
[...]

PR #5780, opened 2022-07-18
5386 off chain reads

turadg:
refs: #5386 (open until `publicNotifiers` is removed, which won't be this PR)

## Description

Provides vault-level data off-chain. Following on https://github.com/Agoric/agoric-sdk/pull/5704. This exposes the new StoredSubscription. It also adds tests, documentation and type annotations for consuming off-chain.

### Security Considerations

All vault published info gets published to RPC nodes. It's keyed by serial vaultId within a manager, so anyone can read all vault publications. This includes the vault holder even after they're transferred the vault and their own subscription terminates. (The new one tees to the same storage node as before.)

Added some docs to the package README

### Testing Considerations

- [x] start up an ag-solo, `agoric follow` a vault, and make sure the publications come

 Wrote up and followed steps in README.md. I had to comment out some checks that depended on pricing info that wasn't resolving. Otherwise it's working.
```
❯ agoric follow :published.vaultFactory.manager0.vaults.vault1
{
  debtSnapshot: {
    debt: {
      brand: slot(0,"Alleged: RUN brand"),
      value: 2n,
    },
    interest: {
      denominator: {
        brand: slot(0),
        value: 100n,
      },
      numerator: {
        brand: slot(0),
        value: 100n,
      },
    },
  },
  locked: {
    brand: slot(1,"Alleged: IbcATOM brand"),
    value: 1000n,
  },
  vaultState: "active",
}
```

### DIFF
--- a/packages/notifier/package.json
+++ b/packages/notifier/package.json
@@ -44,6 +44,7 @@
   },
   "files": [
     "src/",
+    "tools",
     "exported.js",
     "NEWS.md"
   ],
--- a/packages/run-protocol/README.md
+++ b/packages/run-protocol/README.md
@@ -25,3 +25,99 @@ Each interest charging period (say daily) the actual debts in all vaults are aff
 - a debtSnapshot on the vault by which one can calculate the actual debt
 
 To maintain that the keys of vaults to liquidate are stable requires that its keys are also time-independent so they're recorded as a "normalized collateralization ratio", with the actual collateral divided by the normalized debt.
+
+## Reading data off-chain
+
+VaultFactory publishes data using StoredPublishKit which tees writes to off-chain storage. These can then be followed off-chain like so,
+```js
+  const key = `published.vaultFactory.metrics`; // or whatever the stream of interest is
+  const leader = makeDefaultLeader();
+  const follower = makeFollower(storeKey, leader);
+  for await (const { value } of iterateLatest(follower)) {
+    console.log(`here's a value`, value);
+  }
+```
+
+The canonical keys (under `published`) are as follows. Non-terminal nodes could have data but don't yet. A `0` indicates the index of that child in added order. To get the actual key look it up in parent. High cardinality types get a parent key for enumeration (e.g. `vaults`.)
+- `published`
+    - `vaultFactory`
+        - `governance`
+        - `metrics`
+        - `manager0`
+            - `metrics`
+            - `governance`
+            - `vaults`
+              - `vault0`
+    - `amm`
+        - `metrics`
+        - `governance`
+        - `pool0`
+            - `metrics`
+    - `psm`
+        - `governance`
+        - ~~`metrics`~~ TODO
+    - `stakeFactory`
+        - `governance`
+
+### Demo
+
+Start the chain in one terminal:
+```sh
+cd packages/cosmic-swingset
+make scenario2-setup scenario2-run-chain-economy
+```
+Once you see a string like `block 17 commit` then the chain is available. In another terminal,
+```sh
+# shows keys of the vaultFactory
+agd query vstorage keys 'published.vaultFactory'
+# lists vaults
+agd query vstorage keys 'published.vaultFactory.manager0.vaults'
+# follow metrics
+agoric follow :published.vaultFactory.manager0.metrics
+```
+
+Start a new terminal to get a prompt.
+```sh
+cd packages/cosmic-swingset/
+make scenario2-run-client
+```
+
+In yet another,
+```
+cd packages/cosmic-swingset/t1
+agoric open --repl
+```
+
+Connect the wallet and use its REPL as follows. Comments are for explanations and can't be parsed by REPL. The history numbers may be different for your shell. `history[-1]` means whatever number the last history output was.
+```
+# get an instance of the VaultFactory
+vaultFactoryInstance = E(home.agoricNames).lookup('instance', 'VaultFactory')
+# get its public facet
+vaultFactoryPublicFacet = E(home.zoe).getPublicFacet(vaultFactoryInstance)
+# get a reference to the minted brand (soon to be IST)
+E(home.agoricNames).lookup('brand', 'RUN');
+runBrand=history[-1]
+# get a reference to the collateral brand
+E(home.agoricNames).lookup('brand', 'IbcATOM')
+atomBrand=history[-1]
+# get a reference to the collateral manager, using history because the brand must be the same object
+collateralManager = E(vaultFactoryPublicFacet).getCollateralManager(atomBrand)
+# proposal
+proposal = ({
+  give: { Collateral: { brand: atomBrand, value: 1000n } },
+  want: { RUN: { brand: runBrand, value: 1n } },
+})
+# get the ATOM purse
+E(home.wallet).getPurses()
+# get the "ATOM" purse
+atomPurse = history[-1][0][1]
+# prepare funds
+E(atomPurse).withdraw(proposal.give.Collateral)
+collateral = history[-1]
+# make a vault invitation to the collateral manager
+invitation = E(collateralManager).makeVaultInvitation()
+# make the offer with invitation, proposal, payments
+seat = E(home.zoe).offer(invitation, proposal, { Collateral: collateral})
+# get the offer result
+E(seat).getOfferResult()
+```
--- a/packages/run-protocol/src/vaultFactory/vault.js
+++ b/packages/run-protocol/src/vaultFactory/vault.js
@@ -12,7 +12,12 @@ import { defineKindMulti, pickFacet } from '@agoric/vat-data';
 import { makeTracer } from '../makeTracer.js';
 import { calculateCurrentDebt, reverseInterest } from '../interest-math.js';
 import { makeVaultKit } from './vaultKit.js';
-import { addSubtract, assertOnlyKeys, stageDelta } from '../contractSupport.js';
+import {
+  addSubtract,
+  assertOnlyKeys,
+  makeEphemeraProvider,
+  stageDelta,
+} from '../contractSupport.js';
 
 const { details: X, quote: q } = assert;
 
@@ -118,11 +123,30 @@ const validTransitions = {
  */
 
 /**
+ * Ephemera are the elements of state that cannot (or need not) be durable.
+ * When there's a single instance it can be held in a closure, but there are
+ * many vault manaager objects. So we hold their ephemera keyed by the durable
+ * vault manager object reference.
+ *
+ * @type {(key: VaultId) => {
+ * storageNode: ERef<StorageNode>,
+ * marshaller: ERef<Marshaller>,
+ * }} */
+// @ts-expect-error not yet defined
+const provideEphemera = makeEphemeraProvider(() => ({}));
+
+/**
  * @param {ZCF} zcf
  * @param {VaultManager} manager
  * @param {VaultId} idInManager
+ * @param {ERef<StorageNode>} storageNode
+ * @param {ERef<Marshaller>} marshaller
  */
-const initState = (zcf, manager, idInManager) => {
+const initState = (zcf, manager, idInManager, storageNode, marshaller) => {
+  const ephemera = provideEphemera(idInManager);
+  ephemera.storageNode = storageNode;
+  ephemera.marshaller = marshaller;
+
   /**
    * @type {ImmutableState & MutableState}
    */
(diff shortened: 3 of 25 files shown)


packages/run-protocol/README.md
```diff
@@ -25,3 +25,34 @@ Each interest charging period (say daily) the actual debts in all vaults are aff
...
+    console.log(`here's a value`, value);
+  }
+```
+
+The canonical keys (under `published`) are as follows. Non-terminal nodes could have data but don't yet. A `0` indicates the index of that child in added order. To get the actual key look it up in parent.
+- `published`
+    - `vaultFactory`
+        - `governance`
+        - `metrics`
+        - `manager0`
```
dckc, line 46: does `vault${id}` go here?

packages/run-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -714,12 +714,24 @@ const selfBehavior = {
     state.vaultCounter += 1;
     const vaultId = String(state.vaultCounter);
 
-    const vault = makeVault(zcf, manager, vaultId);
+    const storageNode = E(state.storageNode).getChildNode(`vault${vaultId}`);
```
dckc: I was expecting something like `published.vaultFactory.manager6.vaults.23423`. Is there a `.vaults.` node?
Something like `published.vaultFactory.manager6.vaults.vault23423` is fine, but I wouldn't want `vaule232` to be a peer of `metrics` or `governance`.

samsiegart:
> remove publicNotifiers in this PR

We need to update the UI to use RPC for watching the vaults and collaterals. If leaving it in doesn't hurt, that would be better since we could avoid breaking the UI in the meantime.

For UI implementation, I think these are the main considerations:
- Instead of [E(treasuryAPI).getCollaterals()](https://github.com/Agoric/dapp-treasury/blob/main/ui/src/contexts/Application.jsx#L219), we need to follow `:published.vaultFactory.manager0.metrics` <-- is this path going to be well-known?
- Instead of `E(walletP).getPublicNotifiers(vaultOfferId)`, we probably need to implement something in the wallet bridge like `E(walletP).getFollowers(offerId)`. On the **contract**, the return value of the vault offer should include the vault's follower path instead of the `publicNotifiers`/`publicSubscriptions`... i.e. `publicFollowers`.
- Looks like the vaults' `publicNotifiers.asset` is removed entirely, which the UI currently references. Is all this data contained in `:published.vaultFactory.manager0.metrics` instead?

packages/run-protocol/src/vaultFactory/vaultManager.js
turadg: >  I wouldn't want vaule232 to be a peer of metrics or governance

Ah, good point. If you ask for keys of a manager you don't want every vault ever. :)

turadg:
> We need to update the UI to use RPC for watching the vaults and collaterals. If leaving it in doesn't hurt, that would be better since we could avoid breaking the UI in the meantime.

Doesn't hurt the code, just complicates project management. Will the necessary UI changes be in this repo or the dapp? If it's here I agree it's better to do in another PR that does the provider and consumer change. If it's in dapp I'd rather bite the bullet now but will defer to you.

> * Instead of [E(treasuryAPI).getCollaterals()](https://github.com/Agoric/dapp-treasury/blob/main/ui/src/contexts/Application.jsx#L219), we need to follow `:published.vaultFactory.manager0.metrics` <-- is this path going to be well-known?

It will be consistent for a given manager. I think there will also be a way to look up that string without a transaction but that's not in here. 

> * Instead of `E(walletP).getPublicNotifiers(vaultOfferId)`, we probably need to implement something in the wallet bridge like `E(walletP).getFollowers(offerId)`. On the **contract**, the return value of the vault offer should include the vault's follower path instead of the `publicNotifiers`/`publicSubscriptions`... i.e. `publicFollowers`.

That makes sense to me.

> * Looks like the vaults' `publicNotifiers.asset` is removed entirely, which the UI currently references. Is all this data contained in `:published.vaultFactory.manager0.metrics` instead?

`publicNotifiers.assets` is still there. I was asking about removing `publicNotifiers` key entirely but that's not in this PR. What's not here is `publicSubscribers.assets` because `publicSubscribers` is new and should carry forward only what we need.

samsiegart:
> Doesn't hurt the code, just complicates project management. Will the necessary UI changes be in this repo or the dapp? If it's here I agree it's better to do in another PR that does the provider and consumer change. If it's in dapp I'd rather bite the bullet now but will defer to you.

Sorry, yea I'd prefer to not take the notifiers out yet. Seems like some changes will be in this repo like the `getPublicFollowers` wallet method. Mainly, it gives us flexibility to do other work on/use the dapp in the meantime if the migration isn't straightforward, it's not fun having a broken UI hanging over my head if it can be avoided. I created https://github.com/Agoric/dapp-treasury/issues/134 to track.

> What's not here is `publicSubscribers.assets` because `publicSubscribers` is new and should carry forward only what we need.

Gotcha, so the asset data is already in the metrics subscription right?

packages/run-protocol/test/vaultFactory/test-storage.js
```diff
@@ -67,4 +67,17 @@ test('storage keys', async t => {
...
+  t.is(
+    await subscriptionKey(vda1.getVaultSubscriber()),
+    'mockChainStorageRoot.vaultFactory.manager0.vaults.vault1',
+  );
+
+  // Second aeth vault
+  const vda2 = await d.makeVaultDriver(aeth.make(1000n), run.make(50n));
+  t.is(
+    await subscriptionKey(vda2.getVaultSubscriber()),
+    'mockChainStorageRoot.vaultFactory.manager0.vaults.vault2',
```
dckc, line 81: :tada:

packages/run-protocol/README.md
```diff
@@ -25,3 +25,99 @@ Each interest charging period (say daily) the actual debts in all vaults are aff
...
+              - `vault0`
+    - `amm`
+        - `metrics`
+        - `governance`
+        - `pool0`
+            - `metrics`
+    - `psm`
+        - `governance`
+        - ~~`metrics`~~ TODO
+    - `stakeFactory`
```
dckc, line 59: FWIW, metrics and individual BLD boost vaults are TODO too: https://github.com/agoric/agoric-sdk/issues/4653
(not critical / important to add them here now)

packages/run-protocol/README.md
```diff
@@ -25,3 +25,99 @@ Each interest charging period (say daily) the actual debts in all vaults are aff
 - a debtSnapshot on the vault by which one can calculate the actual debt
 
 To maintain that the keys of vaults to liquidate are stable requires that its keys are also time-independent so they're recorded as a "normalized collateralization ratio", with the actual collateral divided by the normalized debt.
+
+## Reading data off-chain
```
dckc, line 29: These docs look _very_ handy. Thanks!

packages/notifier/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "files": [
     "src/",
+    "tools",
```
dckc, line 47: Is this an innovation I should be evaluating or following an established precedent? I suppose I see precedent:

```
agoric-sdk/packages$ echo */tools
agoric-cli/tools SwingSet/tools zoe/tools
```

packages/run-protocol/src/vaultFactory/vault.js
```diff
@@ -117,12 +122,31 @@ const validTransitions = {
...
+ * When there's a single instance it can be held in a closure, but there are
+ * many vault manaager objects. So we hold their ephemera keyed by the durable
+ * vault manager object reference.
+ *
+ * @type {(key: VaultId) => {
+ * storageNode: ERef<StorageNode>,
+ * marshaller: ERef<Marshaller>,
+ * }} */
+// @ts-expect-error not yet defined
+const provideEphemera = makeEphemeraProvider(() => ({}));
```
dckc, line 136: ugh. looks like a form of global mutable state. But I gather it's the least evil of the options we know of.

packages/run-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -301,15 +301,13 @@ const finish = context => {
...
-  const notifier = makeNotifierFromSubscriber(subscriber);
-
   const toCentralPriceAuthority = makePriceAuthority(
     getInputPriceForPA,
     getOutputPriceForPA,
     secondaryBrand,
     centralBrand,
     timer,
-    notifier,
+    subscriber,
```
dckc, line 310: I'm not certain I understand this change. It doesn't seem like a pure refactor; it seems to have observable changes. Are they testested? Or perhaps the change isn't visible from any exported API?

packages/run-protocol/src/vpool-xyk-amm/priceAuthority.js
```diff
@@ -66,24 +70,25 @@ export const makePriceAuthority = (
...
     priceAuthority,
     adminFacet: { fireTriggers },
   } = makeOnewayPriceAuthorityKit(priceAuthorityOptions);
 
-  // An observer on a notifier that gets updated every time the pool's balances
+  // An observer on a notifier that gets updated when the pool's balances
   // change via swap or liquidity operations. It checks the comparison predicate
-  // for each active trigger.
+  // for each active trigger. Due to `subscribeLatest` below it may drop
+  // intermediate values.
```
dckc, line 84: We need to be able to get every pool balance change via RPC. Is that still available after this change?

packages/run-protocol/test/vaultFactory/test-liquidator.js
```diff
@@ -3,526 +3,31 @@
...
-
-/** @typedef {Record<string, any> & {
- *   aeth: IssuerKit & import('../supports.js').AmountUtils,
- *   reserveCreatorFacet: AssetReserveCreatorFacet,
- *   run: IssuerKit & import('../supports.js').AmountUtils,
+  AT_NEXT,
+  makeManagerDriver,
+  makeDriverContext,
+  Phase,
+} from './driver.js';
```
dckc, line 16: yay for concise test code!

packages/vats/src/lib-chainStorage.js
```diff
@@ -34,7 +34,7 @@ export function makeChainStorageRoot(toStorage, storeName, rootPath) {
         assert.typeof(name, 'string');
         assert(
           pathSegmentPattern.test(name),
-          X`Path segment must be a short ASCII identifier: ${name}`,
+          X`Path segment must be a short snake_case alphanumeric identifier: ${name}`,
```
dckc, line 37: snake_case? I must have missed that; help me find docs for that?

dckc (approved):
excellent work!

I have a few questions / comments; none critical.

packages/run-protocol/README.md
turadg: Thanks for https://gist.github.com/dckc/73205178bd44cf3f4bbcf7db0fc280a8#file-borrow-from-vault-js it's cribbed from :)

packages/run-protocol/src/vaultFactory/vault.js
turadg: yeah, the recommendation is userland until https://github.com/Agoric/agoric-sdk/issues/5759 which is low priority.

The `provideEphemera` fn is global state but not mutable. And the objects it provides are mutable, but not global. They're keyed to each virtual object. That requires some discipline but I think we have it. :)

packages/run-protocol/src/vpool-xyk-amm/pool.js
turadg: This one is a pure refactor. We shouldn't need any Notifier within the contract code now that we have prefix-lossy Subscription.

packages/run-protocol/src/vpool-xyk-amm/priceAuthority.js
turadg: The publishing is a `subscribeEach` but the trigger is a `subscribeLatest` because it doesn't need intermediate values if there's a newer one already. That's just extra price quoting work.

packages/run-protocol/test/vaultFactory/test-liquidator.js
turadg: Thanks for leading the way on that

packages/vats/src/lib-chainStorage.js
turadg: https://github.com/Agoric/agoric-sdk/blob/3ce33c1c08ae4fcbff2fa0b087381a8d4be369a7/packages/vats/src/lib-chainStorage.js#L9-L12

PR #5781, opened 2022-07-19
fix(SwingSet): Add durability to vat-vattp objects

### LINKED ISSUE #5667: make vat-vattp upgradable

vat-vattp needs to be upgradable, as part of #5666 . 

vattp has two halves. I'm more familiar with the comms routing half than the networking half (and we may need @michaelfig 's help on that side).

The comms routing half is responsible for dispatching messages to/from named remotes. On the input side, the mailbox device sends all inbound messages to a single object. Each message includes both the string "remote ID" of the sender (e.g. an `agoric1234` public-key -based address), the list of pending messages, and the highest ACK sequence number that the sender has seen (of our outgoing messages). It looks the remoteID up in a table to find the vat-comms `receiver` object, filters out duplicate messages (by seqnum), then sends the new ones in separate `syscall.send`s to the comms receiver.

On the output side, each remote has a distinct `transmitter` object within vat-vattp. Comms sends individual messages to the transmitter, vat-vattp looks up the remoteID for than transmitter, then vat-vattp invokes the mailbox device to push the message onto the named remote's outbox.

vat-vattp also has some wiring APIs: one to teach it about the mailbox device (and vice versa), and another named `addRemote` which causes it to create a new `transmitter`, and to be connected to the vat-comms receiver. There's a short-lived `receiverSetter` object used to finish this handoff.

I don't know what the networking half does, but it appears to share the remoteID-keyed table.

## Description of the Design

I'm thinking that the remoteID-keyed table should be replaced by a durable Kind, with one instance per `transmitter`. The transmitter objects will have a seqnum or two in their `state` properties, as well as a `remoteID`.

The wiring API will need to become a singleton durable Kind. The `receiverSetter` can probably be ephemeral, if we feel safe in assuming that an `addRemote` won't span an upgrade.

## Security Considerations

All mailbox-based messages are routed through vat-vattp, and it has full authority to route them to whatever transmitter/receiver it wishes, so it has considerable power to subvert (or entirely replace) arbitrary user communication. So the authority to upgrade it must be closely held. (this concern is not specific to vat-vattp, it also applies to comms, and to a lesser extent to vat-vat-admin).

## Test Plan

At least careful audit, but ideally (if #5666 also provide a trigger mechanism) we should have unit tests of the actual upgrade process (using a null upgrade).

gibson042:
Fixes #5667

## Description

Makes a durable Mailbox kind and stores instances in a collection.

### Security Considerations

See #5667

Is there already a good reference for making a vat upgradable? If not, would this make a good example?

### Testing Considerations

TBD

### DIFF
--- a/packages/SwingSet/package.json
+++ b/packages/SwingSet/package.json
@@ -21,7 +21,6 @@
     "lint:eslint": "eslint ."
   },
   "devDependencies": {
-    "@agoric/vat-data": "^0.3.1",
     "@types/tmp": "^0.2.0",
     "tmp": "^0.2.1"
   },
@@ -31,6 +30,7 @@
     "@agoric/notifier": "^0.4.0",
     "@agoric/store": "^0.7.2",
     "@agoric/swing-store": "^0.7.0",
+    "@agoric/vat-data": "^0.3.1",
     "@agoric/xsnap": "^0.13.0",
     "@endo/base64": "^0.2.27",
     "@endo/bundle-source": "^2.2.2",
--- a/packages/SwingSet/src/liveslots/virtualReferences.js
+++ b/packages/SwingSet/src/liveslots/virtualReferences.js
@@ -259,8 +259,11 @@ export function makeVirtualReferenceManager(
     if (relaxDurabilityRules) {
       // we'll pretend an object is durable if running with relaxed rules
       return true;
+    } else if (type === 'device') {
+      // devices are durable
+      return true;
     } else if (type !== 'object') {
-      // promises and devices are not durable
+      // promises are not durable
       return false;
     } else if (!allocatedByVat) {
       // imports are durable
--- a/packages/SwingSet/src/vats/vattp/vat-vattp.js
+++ b/packages/SwingSet/src/vats/vattp/vat-vattp.js
@@ -1,8 +1,10 @@
 import { assert, details as X } from '@agoric/assert';
+import { provide } from '@agoric/store';
 import {
   makeScalarBigMapStore,
   defineDurableKindMulti,
   provideKindHandle,
+  vivifySingleton,
 } from '@agoric/vat-data';
 import { E } from '@endo/eventual-send';
 import { makePromiseKit } from '@endo/promise-kit';
@@ -36,10 +38,18 @@ function makeCounter(render, initialValue = 7) {
 
 export function buildRootObject(vatPowers, _vatParams, baggage) {
   const { D } = vatPowers;
-  let mailbox; // mailbox device
 
-  const mailboxKindHandle = provideKindHandle(baggage, 'Mailbox');
-  const initMailboxState = () => ({
+  let mailboxDevice;
+  // TODO: How can we persist a mailbox device reference across upgrades?
+  // const mailboxDeviceBaggageKey = 'mailboxDevice';
+  // if (baggage.has(mailboxDeviceBaggageKey)) {
+  //   mailboxDevice = baggage.get(mailboxDeviceBaggageKey);
+  // }
+
+  // QUESTION: Is there a convention around capitalization here?
+  const mailboxKindHandle = provideKindHandle(baggage, 'mailbox');
+  const initMailboxState = name => ({
+    name,
     outboundHighestAdded: 0,
     outboundHighestRemoved: 0,
     inboundHighestDelivered: 0,
@@ -49,43 +59,55 @@ export function buildRootObject(vatPowers, _vatParams, baggage) {
     mailboxKindHandle,
     initMailboxState,
     {
-      outbound: {
-        getHighestAdded: ({ state }) => state.outboundHighestAdded,
-        setHighestAdded: ({ state }, highestAdded) => {
-          assert(highestAdded >= state.outboundHighestAdded);
-          state.outboundHighestAdded = highestAdded;
+      transmitter: {
+        transmit: ({ state }, msg) => {
+          const num = state.outboundHighestAdded + 1;
+          D(mailboxDevice).add(state.name, num, msg);
+          state.outboundHighestAdded = num;
         },
-        getHighestRemoved: ({ state }) => state.outboundHighestRemoved,
-        setHighestRemoved: ({ state }, highestRemoved) => {
-          assert(highestRemoved >= state.outboundHighestRemoved);
-          state.outboundHighestRemoved = highestRemoved;
+      },
+      setReceiver: {
+        setReceiver: ({ state }, newReceiver) => {
+          assert(!state.inboundReceiver, X`setReceiver is call-once`);
+          state.inboundReceiver = newReceiver;
         },
       },
       inbound: {
-        getReceiver: ({ state }) => state.inboundReceiver,
-        initReceiver: ({ state }, receiver) => {
-          assert.equal(state.inboundReceiver, null);
-          state.inboundReceiver = receiver;
+        deliverMessages: ({ state }, newMessages) => {
+          // TODO: Minimize state interactions. `try..finally`?
+          newMessages.forEach(m => {
+            const [num, body] = m;
+            if (num > state.inboundHighestDelivered) {
+              // TODO: SO() / sendOnly()
+              E(state.inboundReceiver).receive(body);
+              state.inboundHighestDelivered = num;
+              D(mailboxDevice).ackInbound(state.name, num);
+            }
+          });
         },
-        getHighestDelivered: ({ state }) => state.inboundHighestDelivered,
-        setHighestDelivered: ({ state }, highestDelivered) => {
-          assert(highestDelivered >= state.inboundHighestDelivered);
-          state.inboundHighestDelivered = highestDelivered;
+        deliverAck: ({ state }, ack) => {
+          // TODO: Minimize state interactions. `try..finally`?
+          let num = state.outboundHighestRemoved + 1;
+          while (num <= state.outboundHighestAdded && num <= ack) {
+            D(mailboxDevice).remove(state.name, num);
+            state.outboundHighestRemoved = num;
+            num += 1;
+          }
         },
       },
     },
   );
 
-  if (!baggage.has('mailboxes')) {
-    baggage.init(
-      'mailboxes',
-      makeScalarBigMapStore('mailboxes', { durable: true }),
-    );
-  }
-  const mailboxes = baggage.get('mailboxes');
+  const mailboxCollectionBaggageKey = 'mailboxes';
+  /** @type {MapStore<string, ReturnType<typeof makeMailbox>>} */
+  const mailboxes = provide(baggage, mailboxCollectionBaggageKey, () =>
+    makeScalarBigMapStore(mailboxCollectionBaggageKey, { durable: true }),
+  );
   function provideMailbox(name) {
     if (!mailboxes.has(name)) {
-      mailboxes.init(name, makeMailbox());
+      // TODO: Is there a way to avoid this duplication of `name`
+      // as both a map key and internal mailbox state?
+      mailboxes.init(name, makeMailbox(name));
     }
     return mailboxes.get(name);
   }
@@ -179,9 +201,15 @@ export function buildRootObject(vatPowers, _vatParams, baggage) {
     return harden({ host, handler });
   }
 
-  const handler = Far('vat-tp handler', {
-    registerMailboxDevice(mailboxDevnode) {
-      mailbox = mailboxDevnode;
+  const handlerMethods = {
+    registerMailboxDevice(newMailboxDevice) {
+      // TODO: How can we persist a mailbox device reference across upgrades?
+      // if (baggage.has(mailboxDeviceBaggageKey)) {
+      //   baggage.set(mailboxDeviceBaggageKey, newMailboxDevice);
+      // } else {
+      //   baggage.init(mailboxDeviceBaggageKey, newMailboxDevice);
+      // }
+      mailboxDevice = newMailboxDevice;
     },
 
     /**
@@ -189,55 +217,25 @@ export function buildRootObject(vatPowers, _vatParams, baggage) {
      */
     addRemote(name) {
       assert(!mailboxes.has(name), X`already have remote ${name}`);
-      const r = provideMailbox(name);
-      const transmitter = Far('transmitter', {
-        transmit(msg) {
-          // Durability question: are these references to `r` and `name` safe?
-          // If not, then what is the proper strategy?
-          // Do we need to be durable all the way up, with "vat-tp handler" being
-          // a durable singleton with state referencing `mailboxes` from baggage,
-          // an addRemote function reaching into that state and returning a
-          // durable object of its own, and deliverInboundMessages/deliverInboundAck
-          // functions that also leverage that state?
-          const num = r.outbound.getHighestAdded() + 1;
-          D(mailbox).add(name, num, msg);
-          r.outbound.setHighestAdded(num);
-        },
-      });
-      const setReceiver = Far('receiver setter', {
-        setReceiver(newReceiver) {
-          assert(!r.inbound.getReceiver(), X`setReceiver is call-once`);
-          r.inbound.initReceiver(newReceiver);
-        },
-      });
+      const { transmitter, setReceiver } = provideMailbox(name);
       return harden({ transmitter, setReceiver });
     },
 
     deliverInboundMessages(name, newMessages) {
-      const i = provideMailbox(name).inbound;
-      newMessages.forEach(m => {
-        const [num, body] = m;
-        if (num > i.getHighestDelivered()) {
-          // TODO: SO() / sendOnly()
-          E(i.getReceiver()).receive(body);
-          i.setHighestDelivered(num);
-          D(mailbox).ackInbound(name, num);
-        }
-      });
+      // TODO: Stop silently creating mailboxes.
+      const { inbound } = provideMailbox(name);
+      inbound.deliverMessages(newMessages);
     },
 
     deliverInboundAck(name, ack) {
-      const o = provideMailbox(name).outbound;
-      let num = o.getHighestRemoved() + 1;
-      while (num <= o.getHighestAdded() && num <= ack) {
-        D(mailbox).remove(name, num);
-        o.setHighestRemoved(num);
-        num += 1;
-      }
+      // TODO: Stop silently creating mailboxes.
+      const { inbound } = provideMailbox(name);
+      inbound.deliverAck(ack);
     },
 
     makeNetworkHost,
-  });
+  };
 
+  const handler = vivifySingleton(baggage, 'vat-tp handler', handlerMethods);
   return handler;
 }
--- a/packages/SwingSet/src/vats/vattp/vat-vattp.js
+++ b/packages/SwingSet/src/vats/vattp/vat-vattp.js
@@ -1,4 +1,9 @@
 import { assert, details as X } from '@agoric/assert';
+import {
+  makeScalarBigMapStore,
+  defineDurableKindMulti,
+  provideKindHandle,
+} from '@agoric/vat-data';
 import { E } from '@endo/eventual-send';
 import { makePromiseKit } from '@endo/promise-kit';
 import { Far } from '@endo/marshal';
@@ -29,21 +34,60 @@ function makeCounter(render, initialValue = 7) {
   return harden(get);
 }
 
-export function buildRootObject(vatPowers) {
+export function buildRootObject(vatPowers, _vatParams, baggage) {
   const { D } = vatPowers;
   let mailbox; // mailbox device
-  const remotes = new Map();
-  // { outbound: { highestRemoved, highestAdded },
-  //   inbound: { highestDelivered, receiver } }
 
-  function getRemote(name) {
-    if (!remotes.has(name)) {
-      remotes.set(name, {
-        outbound: { highestRemoved: 0, highestAdded: 0 },
-        inbound: { highestDelivered: 0, receiver: null },
-      });
+  const mailboxKindHandle = provideKindHandle(baggage, 'Mailbox');
+  const initMailboxState = () => ({
+    outboundHighestAdded: 0,
+    outboundHighestRemoved: 0,
+    inboundHighestDelivered: 0,
+    inboundReceiver: null,
+  });
+  const makeMailbox = defineDurableKindMulti(
+    mailboxKindHandle,
+    initMailboxState,
+    {
+      outbound: {
+        getHighestAdded: ({ state }) => state.outboundHighestAdded,
+        setHighestAdded: ({ state }, highestAdded) => {
+          assert(highestAdded >= state.outboundHighestAdded);
+          state.outboundHighestAdded = highestAdded;
+        },
+        getHighestRemoved: ({ state }) => state.outboundHighestRemoved,
+        setHighestRemoved: ({ state }, highestRemoved) => {
+          assert(highestRemoved >= state.outboundHighestRemoved);
+          state.outboundHighestRemoved = highestRemoved;
+        },
+      },
+      inbound: {
+        getReceiver: ({ state }) => state.inboundReceiver,
+        initReceiver: ({ state }, receiver) => {
+          assert.equal(state.inboundReceiver, null);
+          state.inboundReceiver = receiver;
+        },
+        getHighestDelivered: ({ state }) => state.inboundHighestDelivered,
+        setHighestDelivered: ({ state }, highestDelivered) => {
+          assert(highestDelivered >= state.inboundHighestDelivered);
+          state.inboundHighestDelivered = highestDelivered;
+        },
+      },
+    },
+  );
+
+  if (!baggage.has('mailboxes')) {
+    baggage.init(
+      'mailboxes',
+      makeScalarBigMapStore('mailboxes', { durable: true }),
+    );
+  }
+  const mailboxes = baggage.get('mailboxes');
+  function provideMailbox(name) {
+    if (!mailboxes.has(name)) {
+      mailboxes.init(name, makeMailbox());
     }
-    return remotes.get(name);
+    return mailboxes.get(name);
   }
 
   const receivers = new WeakMap(); // connection -> receiver
@@ -140,51 +184,54 @@ export function buildRootObject(vatPowers) {
       mailbox = mailboxDevnode;
     },
 
-    /*
-     * 'name' is a string, and must match the name you pass to
-     * deliverInboundMessages/deliverInboundAck
+    /**
+     * @param {string} name  Unique name identifying the remote for "deliverInbound" functions
      */
     addRemote(name) {
-      assert(!remotes.has(name), X`already have remote ${name}`);
-      const r = getRemote(name);
+      assert(!mailboxes.has(name), X`already have remote ${name}`);
+      const r = provideMailbox(name);
       const transmitter = Far('transmitter', {
         transmit(msg) {
-          const o = r.outbound;
-          const num = o.highestAdded + 1;
-          // console.debug(`transmit to ${name}[${num}]: ${msg}`);
+          // Durability question: are these references to `r` and `name` safe?
+          // If not, then what is the proper strategy?
+          // Do we need to be durable all the way up, with "vat-tp handler" being
+          // a durable singleton with state referencing `mailboxes` from baggage,
+          // an addRemote function reaching into that state and returning a
+          // durable object of its own, and deliverInboundMessages/deliverInboundAck
+          // functions that also leverage that state?
+          const num = r.outbound.getHighestAdded() + 1;
           D(mailbox).add(name, num, msg);
-          o.highestAdded = num;
+          r.outbound.setHighestAdded(num);
         },
       });
       const setReceiver = Far('receiver setter', {
         setReceiver(newReceiver) {
-          assert(!r.inbound.receiver, X`setReceiver is call-once`);
-          r.inbound.receiver = newReceiver;
+          assert(!r.inbound.getReceiver(), X`setReceiver is call-once`);
+          r.inbound.initReceiver(newReceiver);
         },
       });
       return harden({ transmitter, setReceiver });
     },
 
     deliverInboundMessages(name, newMessages) {
-      const i = getRemote(name).inbound;
+      const i = provideMailbox(name).inbound;
       newMessages.forEach(m => {
         const [num, body] = m;
-        if (num > i.highestDelivered) {
+        if (num > i.getHighestDelivered()) {
           // TODO: SO() / sendOnly()
-          // console.debug(`receive from ${name}[${num}]: ${body}`);
-          E(i.receiver).receive(body);
-          i.highestDelivered = num;
+          E(i.getReceiver()).receive(body);
+          i.setHighestDelivered(num);
           D(mailbox).ackInbound(name, num);
         }
       });
     },
 
     deliverInboundAck(name, ack) {
-      const o = getRemote(name).outbound;
-      let num = o.highestRemoved + 1;
-      while (num <= o.highestAdded && num <= ack) {
+      const o = provideMailbox(name).outbound;
+      let num = o.getHighestRemoved() + 1;
+      while (num <= o.getHighestAdded() && num <= ack) {
         D(mailbox).remove(name, num);
-        o.highestRemoved = num;
+        o.setHighestRemoved(num);
         num += 1;
       }
     },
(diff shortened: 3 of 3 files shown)


packages/SwingSet/package.json
```diff
@@ -31,6 +30,7 @@
     "@agoric/notifier": "^0.4.0",
     "@agoric/store": "^0.7.2",
     "@agoric/swing-store": "^0.7.0",
+    "@agoric/vat-data": "^0.3.1",
```
warner, line 33: I ran into the need to do this too, while trying to make vat-timer similarly durable. `@agoric/vat-data` has a `devDependency` on swingset (although not a `dependency`.. it fetches `provideKindHandle`/etc from `globalThis.VatData`, not an import). Does that represent a circular import problem of any kind?

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -29,21 +34,60 @@ function makeCounter(render, initialValue = 7) {
...
+  );
+
+  if (!baggage.has('mailboxes')) {
+    baggage.init(
+      'mailboxes',
+      makeScalarBigMapStore('mailboxes', { durable: true }),
+    );
+  }
+  const mailboxes = baggage.get('mailboxes');
+  function provideMailbox(name) {
```
warner, line 113: Instead of `provideMailbox`, our `addRemote` should use `mailboxes.init(name, makeMailbox())`, and our `deliverInboundMessages/Ack` should use `mailboxes.get`. Every remote name should either be added or not, and `deliverInboundMessages('unknown remote name')` should throw an error rather than silently create a new Mailbox.

warner: Well, except that maybe we won't use those names, see below

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -140,51 +184,54 @@ export function buildRootObject(vatPowers) {
...
-          const o = r.outbound;
-          const num = o.highestAdded + 1;
-          // console.debug(`transmit to ${name}[${num}]: ${msg}`);
+          // Durability question: are these references to `r` and `name` safe?
+          // If not, then what is the proper strategy?
+          // Do we need to be durable all the way up, with "vat-tp handler" being
+          // a durable singleton with state referencing `mailboxes` from baggage,
+          // an addRemote function reaching into that state and returning a
+          // durable object of its own, and deliverInboundMessages/deliverInboundAck
+          // functions that also leverage that state?
```
warner: yeah, basically that. When we're done, the only `Far` that should appear in this file is the one that builds the root object. Everything else needs to be a durable Kind, or in a durable collection.

`r` can be used safely from durable data, because it's an instance of a durable kind (the `Mailbox` kind). And `name` is durable-safe, since it's just a string.

The approach I'd take is to use `defineKindMulti()` to create a cohort of `{ transmitter, receiverSetter, inbound }` objects for each remote, which would subsume the `Mailbox` kind you defined. The `state` of these objects will include the counters and the `receiver` object reference, and also the `name` of the remote. The  transmitter facet has the `transmit(msg)` method (which does stuff with the counters, then uses the `name` to make the call to `D(mailbox).add`. The receiverSetter facet will update `state.receiver`. The `inbound` object will implement `deliverMessages(newMessages)` and `deliverAck(ack)` for each name, so the shared `deliverInboundAck(name, ack)` will do `inbound.get(name).deliverAck(ack)`, and `deliverAck()` will do a bunch of `state.highestRemoved + 1` -type stuff. `deliverMessages()` will use `E(state.receiver).receive(body)` and `state.highestDelivered = num`.


You'll need a `makeScalarBigMapStore` to map from remote `name` to the `inbound` facet (but feel free to call it a mailbox or a mailbox handler or whatever), since that's the one place where we'll see messages for lots of different remotes all get sent to the same object. (in the outbound direction, each remote uses a distinct `transmitter` object, but since those will share counter state with the inbound side, having `transmitter` be one facet, and the inbound side be a different facet, both of the same cohort and sharing state, seems like a nice design). You won't need a map to remember transmitters: vat-comms will hold onto those object references from afar. And you won't need a map to remember the receiverSetters: we'll return those to vat-bootstrap, which will use it briefly and then drop it. Only the inbound side (which receives messages from the mailbox device) needs to lookup anything using the remote name.

warner (changes requested):
Nice start! I think the next step is to update the `addRemote` side to use a cohort of `{ transmitter, receiverSetter, inbound }` as mentioned. Then we should tackle the networking side of the file.

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -135,62 +215,41 @@ export function buildRootObject(vatPowers) {
...
-    registerMailboxDevice(mailboxDevnode) {
-      mailbox = mailboxDevnode;
+  const handlerMethods = {
+    registerMailboxDevice(newMailboxDevice) {
+      // TODO: How can we persist a mailbox device reference across upgrades?
+      // if (baggage.has(mailboxDeviceBaggageKey)) {
+      //   baggage.set(mailboxDeviceBaggageKey, newMailboxDevice);
+      // } else {
+      //   baggage.init(mailboxDeviceBaggageKey, newMailboxDevice);
+      // }
```
gibson042: @FUDCo Here is the error encountered when uncommenting this:
```
Logging sent error stack (Error#1)
Error#1: value is not durable: Object [Alleged: sender] {} at slot 0 of {
  body: slot(0,"Alleged: sender"),
  slots: [
    "d-70",
  ],
}
  at throwNotDurable (.../swingset-vat/src/liveslots/collectionManager.js:97:8)
  at eval (.../swingset-vat/src/liveslots/collectionManager.js:376:1)
  at Alleged: mapStore.init (.../swingset-vat/src/liveslots/collectionManager.js:374:23)
  at registerMailboxDevice (.../swingset-vat/src/vats/vattp/vat-vattp.js:226:9)
  at eval (.../vat-data/src/kind-utils.js:11:1)
  at Alleged: vat-tp handler.method [as registerMailboxDevice] (.../swingset-vat/src/liveslots/virtualObjectManager.js:486:33)

Error#1 ERROR_NOTE: Thrown from: (Error#2) : 3 . 0
Error#1 ERROR_NOTE: Sent as error:liveSlots:v11#70001
Nested error under Error#1
  Error#2: Event: 2.1
    at deliver (.../swingset-vat/src/liveslots/liveslots.js:1077:20)
    at dispatchToUserspace (.../swingset-vat/src/liveslots/liveslots.js:1454:1)
    at eval (.../swingset-vat/src/kernel/dummyMeterControl.js:40:31)
    at runWithoutMetering (.../swingset-vat/src/kernel/dummyMeterControl.js:22:8)
    at wrapped (.../swingset-vat/src/kernel/dummyMeterControl.js:40:8)
  
Logging sent error stack (RemoteError(error:liveSlots:v11#70001)#3)
RemoteError(error:liveSlots:v11#70001)#3: value is not durable: "[Alleged: sender]" at slot 0 of "{\n  body: slot(0,\"Alleged: sender\"),\n  slots: [\n    \"d-70\",\n  ],\n}"
  at fullRevive (.../marshal/src/marshal.js:418:20)
  at Object.unserialize (.../marshal/src/marshal.js:501:21)
  at notifyOnePromise (.../swingset-vat/src/liveslots/liveslots.js:1191:13)
  at notify (.../swingset-vat/src/liveslots/liveslots.js:1215:15)
  at dispatchToUserspace (.../swingset-vat/src/liveslots/liveslots.js:1459:1)
  at eval (.../swingset-vat/src/kernel/dummyMeterControl.js:40:31)
  at runWithoutMetering (.../swingset-vat/src/kernel/dummyMeterControl.js:22:8)
  at wrapped (.../swingset-vat/src/kernel/dummyMeterControl.js:40:8)

RemoteError(error:liveSlots:v11#70001)#3 ERROR_NOTE: Rejection from: (Error#4) : 2 . 0
RemoteError(error:liveSlots:v11#70001)#3 ERROR_NOTE: Rejection from: (Error#5) : 1 . 1
RemoteError(error:liveSlots:v11#70001)#3 ERROR_NOTE: Sent as error:liveSlots:v1#70001
Nested 2 errors under RemoteError(error:liveSlots:v11#70001)#3
  Error#4: Event: 1.1
    at Proxy.eval (.../eventual-send/src/E.js:33:41)
    at Alleged: root.bootstrap (.../swingset-vat/demo/encouragementBotComms/bootstrap.js:18:25)
  
  Error#4 ERROR_NOTE: Caused by: (Error#5)
  Nested error under Error#4
    Error#5: Event: 0.1
      at deliver (.../swingset-vat/src/liveslots/liveslots.js:1077:20)
      at dispatchToUserspace (.../swingset-vat/src/liveslots/liveslots.js:1454:1)
      at eval (.../swingset-vat/src/kernel/dummyMeterControl.js:40:31)
      at runWithoutMetering (.../swingset-vat/src/kernel/dummyMeterControl.js:22:8)
      at wrapped (.../swingset-vat/src/kernel/dummyMeterControl.js:40:8)
    
##### KERNEL PANIC: kp40.policy panic: rejected {"body":"{\"@qclass\":\"error\",\"errorId\":\"error:liveSlots:v1#70001\",\"message\":\"value is not durable: \\\"[Alleged: sender]\\\" at slot 0 of \\\"{\\\\n  body: slot(0,\\\\\\\"Alleged: sender\\\\\\\"),\\\\n  slots: [\\\\n    \\\\\\\"d-70\\\\\\\",\\\\n  ],\\\\n}\\\"\",\"name\":\"Error\"}","slots":[]} #####
REJECTED from ava test: (Error#6)
Error#6: kernel panic kp40.policy panic: rejected {"body":"{\"@qclass\":\"error\",\"errorId\":\"error:liveSlots:v1#70001\",\"message\":\"value is not durable: \\\"[Alleged: sender]\\\" at slot 0 of \\\"{\\\\n  body: slot(0,\\\\\\\"Alleged: sender\\\\\\\"),\\\\n  slots: [\\\\n    \\\\\\\"d-70\\\\\\\",\\\\n  ],\\\\n}\\\"\",\"name\":\"Error\"}","slots":[]}
  at panic (.../swingset-vat/src/kernel/kernel.js:240:18)
  at doResolve (.../swingset-vat/src/kernel/kernelQueue.js:59:1)
  at resolve (.../swingset-vat/src/kernel/kernelSyscall.js:286:1)
  at kernelSyscallHandler (.../swingset-vat/src/kernel/kernelSyscall.js:361:8)
  at vatSyscallHandler (.../swingset-vat/src/kernel/kernel.js:1438:6)
  at syscallFromWorker (.../swingset-vat/src/kernel/vat-loader/manager-helper.js:253:12)
  at doSyscall (.../swingset-vat/src/supervisors/supervisor-helper.js:71:3)
  at Object.resolve (.../swingset-vat/src/supervisors/supervisor-helper.js:104:24)
  at handle (.../swingset-vat/src/liveslots/liveslots.js:1155:9)
  at eval (.../swingset-vat/src/liveslots/liveslots.js:1174:10)
```

packages/SwingSet/src/liveslots/virtualReferences.js
```diff
@@ -259,8 +259,11 @@ export function makeVirtualReferenceManager(
     if (relaxDurabilityRules) {
       // we'll pretend an object is durable if running with relaxed rules
       return true;
+    } else if (type === 'device') {
+      // devices are durable
+      return true;
     } else if (type !== 'object') {
-      // promises and devices are not durable
+      // promises are not durable
```
gibson042, line 266: Changes pulled forward from @FUDCo.

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,253 @@ import { Far } from '@endo/marshal';
...
-        await E(comms).addRemote(name, transmitter, receiverReceiver);
-        await E(comms).addEgress(name, 0, locatorUnum);
-        gotTheirLocatorUnum(E(comms).addIngress(name, 0));
-      },
-      onReceive(connection, msg) {
-        // setReceiver ought to be called before there's any chance of
-        // onReceive being called
-        E(receivers.get(connection)).receive(msg);
+  // QUESTION: Is there a convention around capitalization here?
+  const mailboxKindHandle = provideKindHandle(baggage, 'mailbox');
```
gibson042: e.g., "mailbox" vs. "Mailbox"

packages/SwingSet/src/vats/vattp/vat-vattp.js
gibson042: https://github.com/Agoric/agoric-sdk/issues/5824

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,253 @@ import { Far } from '@endo/marshal';
...
+        transmit: ({ state }, msg) => {
+          // 'msg' will be a string (vats/comms/outbound.js deliverToRemote)
+          E(state.connection).send(msg);
+        },
+      },
+      localLocatorUnum: {
+        lookup: ({ state }, address) => {
+          return state.exportedObjects.get(address);
+        },
+      },
```
gibson042, line 247: Objects like these are needed for the comms vat handshaking in `onOpen`, and making them internal-use facets of the network host multi-kind is the best durability approach I could think of. But I'm open to other options.

gibson042 (commented):
This is now ready for full review.

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: I've been using `mailboxHandle`. My vat-timer work has a batch of handle creations at the beginning:
```js
  // first, create the KindHandles for all our durable objects
  const serviceHandle = provideKindHandle(baggage, 'timerServiceHandle');
  const oneShotEventHandle = provideKindHandle(baggage, 'oneShotEventHandle');
  const promiseEventHandle = provideKindHandle(baggage, 'promiseEventHandle');
  const repeaterEventHandle = provideKindHandle(baggage, 'repeaterEventHandle');
  const repeaterHandle = provideKindHandle(baggage, 'repeater');
  const notifierHandle = provideKindHandle(baggage, 'notifier');
```

huh, which isn't consistent

It doesn't matter too much, but now I'm thinking I want e.g. `mailboxKindHandle`, so the name of the baggage key matches the name of the variable that's holding the value. I'll change mine to folow that rule.

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: that sounds good to me

packages/SwingSet/src/liveslots/virtualReferences.js
warner: yeah, all three of us have this change in our branches, it should fold out when the last two rebase our branches onto the first that lands

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: hm, no, `repeaterEventKindHandle` makes the lines of my code too long:
```js
  // first, create the KindHandles for all our durable objects
  const timerServiceKindHandle = provideKindHandle(
    baggage,
    'timerServiceKindHandle',
  );
  const oneShotEventKindHandle = provideKindHandle(
    baggage,
    'oneShotEventKindHandle',
  );
  const promiseEventKindHandle = provideKindHandle(
    baggage,
    'promiseEventKindHandle',
  );
  const repeaterEventKindHandle = provideKindHandle(
    baggage,
    'repeaterEventKindHandle',
  );
  const repeaterKindHandle = provideKindHandle(baggage, 'repeaterKindHandle');
  const notifierKindHandle = provideKindHandle(baggage, 'notifierKindHandle');
```

so I'm going to go with `fooHandle` instead of `fooKindHandle`.

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+          }
+        },
       },
-    });
+    },
+  );
 
-    return harden({ host, handler });
+  /** @type {MapStore<string, ReturnType<typeof makeMailbox>>} */
+  const mailboxes = provideDurableMapStore(baggage, 'mailboxes');
```
warner: I've been gravitating towards putting all definitions of function-side state up towards the top of the function, in a cluster, to make it easy to see everything at once, but I'm fine with it either way.

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+  function provideMailbox(name) {
+    if (!mailboxes.has(name)) {
+      // TODO: Is there a way to avoid this duplication of `name`
```
warner: Not really, comms sends outbound messages to the `transmitter` object (so comms doesn't know the `name`), and `transmitter` needs to know the name so it can be included in the `D(mailboxdevice).add()` arguments. We have a name->facet mapping in `mailboxes`, and we need a facet->name mapping. We can either distribute that among the `state` of all the transmitters, or we could have a separate table `durableMapStore` keyed by `transmitter` and holding names in its values, and since we already have four state variables per remote, it sounds easier to add a fifth one than to add a whole separate table.

At some point it might be interesting to consider a bidirectional mapping Store type, which enforces a 1:1 mapping, but I don't think it would save any storage space (it'd be implemented in terms of two normal Maps).

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+  //     },
+  //   });
+  //
+  // B:
+  //
+  //   const { host, handler } = await E(agoric.vattp).makeNetworkHost('ag-chain-A', comms);
+  //   E(agoric.ibcport[1]).connect('/ibc-port/portADDR/ordered/vattp-1', handler);
+  //   E(E(host).lookup(helloAddress)).hello();
+
+  const nextConnectionNumberBaggageKey = 'networkHostNextConnectionNumber';
```
warner: nit: feel free to make the variable name smaller so the next clause is denser and more readable (I'm glaring at `prettier`). If all of thes uses are within a few lines (e.g. especially if you added a `allocateConnectionNumber` function immediately after the `provide` and used that below), then the name doesn't need to carry so much context

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+          E(state.connection).send(msg);
+        },
+      },
+      localLocatorUnum: {
+        lookup: ({ state }, address) => {
+          return state.exportedObjects.get(address);
+        },
+      },
+      handler: {
+        onOpen: async ({ state, facets: self }, connection, ..._args) => {
```
warner: I'd leave the parameter named `facets`, and use e.g. `addRemote(name, facets.commsTransmiter, facets.commsSetReceiver)` below, so that the use of `self` doesn't trick someone into thinking this is a single-facet Kind instead of a MultiKind

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
-  return handler;
+          const comms = state.comms;
+          // TODO: Can these calls be made in parallel?
+          await E(comms).addRemote(
+            name,
+            self.commsTransmitter,
+            self.commsSetReceiver,
+          );
+          await E(comms).addEgress(name, 0, self.localLocatorUnum);
+          state.remoteLocatorUnum = await E(comms).addIngress(name, 0);
```
warner, line 262: One tiny difference between this and the original: `onOpen()` will stall until `comms.addIngress` finishes, whereas the original would proceed along. I think that's fine, `addIngress` is prompt (it doesn't wait for anything interesting) so there's not a risk of deadlock. And I do like the simplification that the `await` provides.. easier to read this way.

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+          console.warn(`deleting connection is not fully supported in comms`);
+        },
+        infoMessage: ({ state }, ...args) => {
+          void E(state.console).log('VatTP connection info:', ...args);
+        },
+      },
+    },
+  );
+
+  const vatFacets = vivifySingleton(baggage, 'vat-tp handler', {
```
warner: This value is an instance of a single-facet Kind, not a multi-Kind. So I'd use `vattpHandler` as the variable name instead of `vatFacets`, to avoid confusion.

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+  //   E(E(host).lookup(helloAddress)).hello();
+
+  const nextConnectionNumberBaggageKey = 'networkHostNextConnectionNumber';
+  let nextConnectionNumber = provide(
+    baggage,
+    nextConnectionNumberBaggageKey,
+    () => 7n,
+  );
+  // TODO: Should this collection be Weak (non-iterable)?
+  const networkHostNames = provideDurableSetStore(baggage, 'networkHostNames');
```
warner: It's indexed by string, so I think it can't be a WeakStore (my reading of the Store docs say that only Remotables can be keys, which aligns with a normal `new WeakMap()` only accepting `Object`s as keys).

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,255 @@ import { Far } from '@endo/marshal';
...
+          // assuming state won't change underneath us
+          // and remote methods won't synchronously throw.
+          const { name, inboundReceiver, inboundHighestDelivered } = state;
+          let newHighestDelivered = inboundHighestDelivered;
+          for (const [num, body] of newMessages) {
+            if (num > newHighestDelivered) {
+              // TODO: SO() / sendOnly()
+              E(inboundReceiver).receive(body);
+              newHighestDelivered = num;
+              D(mailboxDevice).ackInbound(name, num);
```
warner: Technically this could throw.. there's no good reason for the mailbox device to do so, but we should consider what would happen if it did.

Ok, so say we get three `newMessages` here, 1/2/3. And say the device call throws during message 2, so we've already called `E(inboundReceiver).receive(1)` and `(2)`. Now the device might or might not have implemented its ack update (setting the mailbox inbound `ack` value to 2) before it throws. Once it does throw, this loop would be exited, and the function would exit, so we wouldn't set `state.inboundHighestDelivered=` at the end leaving it at `0`.

The mailbox device that pushed the `deliverMessages` call is not waiting on a result promise, so it won't notice the error. It thinks the `deliverMessages` was reliable, so (assuming the host keeps running) the next call will get message 4. We'll deliver that to the receiver (because the `num` is higher than our remembered `0`), so the receiver will miss out on 3 entirely. We currently have the comms protocol include explicit sequence numbers, which means comms will notice the gap, and `receive()` will throw. Comms will throw on every received message that isn't `3`, effectively breaking that particular connection. If/when we turn off explicit seqnums, comms would just get very confused, which is probably worse.

In some contexts (more likely an ag-solo than a chain), if the host happened to shut down and get restarted after this error, the mailbox device is within its rights to re-deliver all the un-acked inbound messages, especially when it receives an even newer message (4) later. So it might make a new `deliverMessages` call with 2/3 or 2/3/4. Assuming `ackInbound` doesn't throw again, we'd receive and process the `2` (causing a duplicate comms message, which would cause comms-with-explicit-seqnums to throw). vattp doesn't pay attention to the comms results, so vattp will charge ahead. It does `D().ackInbound()` with `2`, which is fine regardless of whether the earlier (throwing) call managed to record or not, because all `ackInbound` does is set a piece of mailbox state (it doesn't insist that the new value is one-larger or at-all-larger than the old value). So then vattp can continue ahead to `3`, which *will* be accepted by comms, and then acked, and everything can get back on track.

In the old version, the device call was the last thing we did before looping (we finished updated our local state just before calling into the device). That design was resistant to making duplicate deliveries to comms (the recorded seqnum would be accurate, so they'd be skipped by `deliverInboundMessages`), but could still experience gaps. So, not rock-solid, but maybe it has one fewer failure mode.

So, on the whole, I think I'd prefer to do the `state.inboundHighestDelivered` assignment within the loop, just before making the device call. The cost is an extra virtual-object property write, but remember that `VIRTUAL_OBJECT_CACHE_SIZE` means that we won't write this back to the DB until either a `delivery.bringOutYourDead` or the object gets pushed out of the LRU cache, so I don't think it will cause extra syscalls or DB traffic, making it effectively free.

warner (approved):
That looks great!

A couple of nits, and one requested change: `deliverInbound` should update the state *before* calling the device `ackInbound`. In general, let's do the device calls after making state changes, so just in case the device throws for some reason, we might get lucky and have less corruption/confusion within vattp.

packages/SwingSet/src/vats/vattp/vat-vattp.js
gibson042: `mailboxHandle = provideKindHandle(baggage, 'mailboxHandle')` reads strangely to me... if I understand correctly, "mailboxHandle" maps to a key in baggage whose associated value is meaningful only for uniquely identifying the mailbox object kind which has its behavior defined (in each version) by `defineDurableKindMulti(mailboxHandle, …)`.

So would something like `mailboxKind = provideKindHandle(baggage, 'mailboxKind')` be a more accurate representation of the underlying mechanics?

packages/SwingSet/src/vats/vattp/vat-vattp.js
gibson042: I'm sympathetic to that pattern, but it seems a bit at odds with `vivifySingleton` (which ensures initialization of a key in baggage) being forced to come late because of the guidance to define all variables before their references.

My current compromise is an early "_Define all durable baggage keys and kind handles_" block and use of its constants to define kinds, collections, and durable values later on with preceding "_define kind_"/"_retain collection_" comments, at the cost of long `…BaggageKey` variable names.

packages/SwingSet/src/vats/vattp/vat-vattp.js
gibson042: According to [A Taxonomy of Stores](https://github.com/Agoric/agoric-sdk/blob/0e8644a34f996242935d4a525c631ae32e1da462/packages/store/docs/store-taxonomy.md#user-content-weak-vs-strong), the relevant characteristic of weak collections is that they are not iterable—but they "_may_" allow forgeable keys (with the understanding that such entries can never be GC'd). So it seems like a WeakSetStore of strings is possible here, but it wouldn't support any automatic reduction of stored data.

packages/SwingSet/src/vats/vattp/vat-vattp.js
gibson042: > Technically this could throw.. there's no good reason for the mailbox device to do so, but we should consider what would happen if it did.
> 
> Ok, so say we get three `newMessages` here, 1/2/3. And say the device call throws during message 2, so we've already called `E(inboundReceiver).receive(1)` and `(2)`. Now the device might or might not have implemented its ack update (setting the mailbox inbound `ack` value to 2) before it throws. Once it does throw, this loop would be exited, and the function would exit, so we wouldn't set `state.inboundHighestDelivered=` at the end leaving it at `0`.

Isn't `D(validDevice).method(…)` guaranteed not to synchronously throw? That was my understanding as commented above[^1], and it seems to align with https://github.com/Agoric/agoric-sdk/blob/0e8644a34f996242935d4a525c631ae32e1da462/packages/SwingSet/src/liveslots/liveslots.js#L1020-L1033 (assuming that is the right definition).

> So, on the whole, I think I'd prefer to do the `state.inboundHighestDelivered` assignment within the loop, just before making the device call. The cost is an extra virtual-object property write, but remember that `VIRTUAL_OBJECT_CACHE_SIZE` means that we won't write this back to the DB until either a `delivery.bringOutYourDead` or the object gets pushed out of the LRU cache, so I don't think it will cause extra syscalls or DB traffic, making it effectively free.

Oh, I didn't realize we already had that. I observed a testing speedup from this change (and in fact that's what motivated it), but I can change it back if you don't find the above analysis of `D` sufficiently compelling.

[^1]: "_Minimize interactions with durable storage, assuming state won't change underneath us and remote methods won't synchronously throw._"

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: Yeah, I'm torn too. @erights added a `/* eslint-disable no-use-before-define */` to the top of vat-timer.js for me, so I consider that permission to ignore that bit of guidance :). But I'm also torn between putting all the durable things and Kind definitions in one place (where a future maintainer can easily see what they're obligated to maintain), and using convenient tools like `vivifySingleton` that don't lend themselves to that approach. See https://github.com/Agoric/agoric-sdk/pull/5862#issuecomment-1201955257 for some discussion.

Any approach you choose is good by me.

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: yeah, I think our current implementation doesn't:

```js
import { makeScalarWeakMapStore } from '@agoric/store';
test('foo', async t => {
  const s = makeScalarWeakMapStore();
  s.init('s', 'str');
  t.pass();
});
```

gets me:
```
REJECTED from ava test: (Error#1)
Error#1: Only remotables can be keys of scalar WeakMapStores: s
  at assertKVOkToAdd (packages/store/src/stores/scalarWeakMapStore.js:117:5)
  at Alleged: scalar WeakMapStore of "key".init (packages/store/src/stores/scalarWeakMapStore.js:45:7)
  at packages/SwingSet/test/test-vat-timer.js:763:5
  at async Promise.all (index 1)
```

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: Yeah, the vat-vattp code is written with the expectation that the device call won't throw, I'm just thinking deviously this week.

The real code to look at is the `DeviceHandler` definition just above that chunk:
https://github.com/Agoric/agoric-sdk/blob/0e8644a34f996242935d4a525c631ae32e1da462/packages/SwingSet/src/liveslots/liveslots.js#L997-L1015

`D(mailboxDevice)` on a real device node won't throw, but the `.ackInbound(..)` could throw if the args are unserializable, are too big, hold a promise, if the underlying device invocation throws, or if it returns a promise. None of those are things well-behaved code would do, though.

But if you saw a speedup, then yeah, go with the faster version. I don't have any good reason to believe that my ordering would be meaningfully more robust in practice (if the device throws, who knows what else will go wrong).

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -163,24 +171,27 @@ export function buildRootObject(vatPowers, _vatParams, baggage) {
...
+  const networkHostNames = provideDurableSetStore(
     baggage,
-    nextConnectionNumberBaggageKey,
-    () => 7n,
+    networkHostNamesBaggageKey,
+  );
+  let networkHostCounter = provide(
+    baggage,
+    networkHostCounterBaggageKey,
+    () => 0n,
```
warner, line 185: that changes the initial value from 7n to 0n, but I'm cool with that, I don't think the non-zero starting point was buying us anything

packages/SwingSet/src/vats/vattp/vat-vattp.js
```diff
@@ -16,181 +23,267 @@ import { Far } from '@endo/marshal';
...
+          void E(state.console).log('VatTP connection info:', ...args);
+        },
+      },
+    },
+  );
+
+  // Expose a durable service singleton.
+  return vivifySingleton(baggage, serviceSingletonBaggageKey, {
+    ...serviceMailboxFunctions,
+    makeNetworkHost: (allegedName, comms, cons = console) => {
```
warner, line 282: tiny nit, this combination of a `...` splat and a single name initially made me think that `makeNetworkHost` was somehow overriding one portion of `serviceMailboxFunctions`, and it took me a minute to realize this was composing two separate bits of functionality into the return value

what do you think about defining `networkFunctions = { makeNetworkHost }` and then having this be `{ ...serviceMailboxFunctions, ...networkFunctions }` ?

but it's fine either way

warner (approved):
nice job! some minor comments but feel free to land without them

packages/SwingSet/src/vats/vattp/vat-vattp.js
warner: (retroactive review note for future historians: it turns out that returning a virtual/durable object from `buildRootObject` causes problems, and I missed this during review. #7240 has details, and our PR to prohibit vats from doing this also changes this line to use `Far` instead of `vivifySingleton`)

PR #5784, opened 2022-07-19
fix: serialize all tests in test-terminate.js to avoid fatal memory-usage spike

warner:
Use test.serial to disable parallelism in SwingSet's test-terminate.js

This test file has 32 individual test cases, all of which launch a kernel, and under XS mode each kernel launches 5-6 `xsnap` worker subprocesses. When run in parallel, the combined memory usage of all these workers (about 170) overwhelms an 8GB host computer, causing the OOM Killer to kill `test-terminate.js`, causing GitHub CI failures.

The fix is to use `test.serial()` to prohibit that parallelism, at least for high-fan-out test files like test-terminate.js. This slows down the test somewhat (68s vs 48s on my laptop), but reduces the memory usage spike.

### DIFF
--- a/packages/SwingSet/test/vat-admin/terminate/test-terminate.js
+++ b/packages/SwingSet/test/vat-admin/terminate/test-terminate.js
@@ -16,6 +16,18 @@ test.before(async t => {
   t.context.data = { kernelBundles };
 });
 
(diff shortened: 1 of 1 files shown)


FUDCo (approved):
I hate it when technology deviates from perfection.

PR #5785, opened 2022-07-19
make assetReserve reachably durable

### LINKED ISSUE #5647: Reserve contract is reachably durable

The Reserve contract must be able to survive upgrade. Currently it holds state that is durable. Some is not virtualized.

## Description of the Design
1. Virtualize
2. Make durable
3. Make robust to upgrade

Chris-Hibbert:
closes: #5647

## Description

Add tools to contractHelper to support durable contracts
add vivifyKindMulti\
improve types for governedContracts

### Security Considerations

Not significant

None

### Testing Considerations

This qualifies as Reachably Durable, which means that the contract stores its state to baggage, but hasn't been tested to survive upgrade.

### DIFF
--- a/packages/governance/src/contractGovernor.js
+++ b/packages/governance/src/contractGovernor.js
@@ -98,11 +98,11 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
  * contract (as electionManager) so clients will be able to look up the state
  * of the governed parameters.
  *
- * @template {object} PF Public facet of governed
+ * @template {{}} PF Public facet of governed
  * @template {ContractPowerfulCreatorFacet} CF Creator facet of governed
  * @type {ContractStartFn<
  * GovernorPublic,
- * GovernedContractFacetAccess<PF>,
+ * GovernedContractFacetAccess<PF,CF>,
  * {
  *   timer: TimerService,
  *   electorateInstance: Instance,
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -89,22 +89,6 @@ const facetHelpers = (zcf, paramManager) => {
     });
   };
 
-  function governorFacet(limitedCreatorFacet, governedApis) {
-    return Far('governorFacet', {
-      getParamMgrRetriever: () =>
-        Far('paramRetriever', { get: () => paramManager }),
-      getInvitation: name => paramManager.getInternalParamValue(name),
-      getLimitedCreatorFacet: () => limitedCreatorFacet,
-      // The contract provides a facet with the APIs that can be invoked by
-      // governance
-      getGovernedApis: () => Far('governedAPIs', governedApis),
-      // The facet returned by getGovernedApis is Far, so we can't see what
-      // methods it has. There's no clean way to have contracts specify the APIs
-      // without also separately providing their names.
-      getGovernedApiNames: () => Object.keys(governedApis),
-    });
-  }
-
   /**
    * Add required methods to a creatorFacet
    *
@@ -115,27 +99,50 @@ const facetHelpers = (zcf, paramManager) => {
    */
   const makeGovernorFacet = (originalCreatorFacet, governedApis = {}) => {
     const limitedCreatorFacet = makeLimitedCreatorFacet(originalCreatorFacet);
+    const governorFacet = Far('governorFacet', {
+      getParamMgrRetriever: () =>
+        Far('paramRetriever', { get: () => paramManager }),
+      getInvitation: (_context, name) =>
+        paramManager.getInternalParamValue(name),
+      getLimitedCreatorFacet: () => limitedCreatorFacet,
+      // The contract provides a facet with the APIs that can be invoked by
+      // governance
+      getGovernedApis: () => Far('governedAPIs', governedApis),
+      // The facet returned by getGovernedApis is Far, so we can't see what
+      // methods it has. There's no clean way to have contracts specify the APIs
+      // without also separately providing their names.
+      getGovernedApiNames: () => Object.keys(governedApis),
+    });
 
     // exclusively for contractGovernor, which only reveals limitedCreatorFacet
     // @ts-expect-error type confusion
-    return governorFacet(limitedCreatorFacet, governedApis);
+    return governorFacet;
   };
 
   /**
    * Add required methods to a creatorFacet for a virtual/durable contract.
    *
    * @param {{ [methodName: string]: (context?: unknown, ...rest: unknown[]) => unknown}} originalCreatorFacet
-   * @param {Record<string, (...args: any[]) => any>} governedApis
    */
-  const makeVirtualGovernorFacet = (
-    originalCreatorFacet,
-    governedApis = {},
-  ) => {
+  const makeVirtualGovernorFacet = originalCreatorFacet => {
     const limitedCreatorFacet = makeLimitedCreatorFacet(originalCreatorFacet);
 
-    // exclusively for contractGovernor, which only reveals limitedCreatorFacet
-    const facet = governorFacet(limitedCreatorFacet, governedApis);
-    return objectMap(facet, dropContext);
+    const governorFacet = Far('governorFacet', {
+      getParamMgrRetriever: () =>
+        Far('paramRetriever', { get: () => paramManager }),
+      getInvitation: (_context, name) =>
+        paramManager.getInternalParamValue(name),
+      getLimitedCreatorFacet: ({ facets }) => facets.limitedCreatorFacet,
+      // The contract provides a facet with the APIs that can be invoked by
+      // governance
+      getGovernedApis: ({ facets }) => facets.governedApis,
+      // The facet returned by getGovernedApis is Far, so we can't see what
+      // methods it has. There's no clean way to have contracts specify the APIs
+      // without also separately providing their names.
+      getGovernedApiNames: ({ facets }) => Object.keys(facets.governedApis),
+    });
+
+    return { governorFacet, limitedCreatorFacet };
   };
 
   return harden({
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -67,7 +67,8 @@ const facetHelpers = (zcf, paramManager) => {
   /**
    * Add required methods to publicFacet, for a virtual/durable contract
    *
-   * @param {{}} originalPublicFacet
+   * @param {OPF} originalPublicFacet
+   * @template {{}} OPF
    */
   const augmentVirtualPublicFacet = originalPublicFacet => {
     return Far('publicFacet', {
@@ -115,7 +116,6 @@ const facetHelpers = (zcf, paramManager) => {
     });
 
     // exclusively for contractGovernor, which only reveals limitedCreatorFacet
-    // @ts-expect-error type confusion
     return governorFacet;
   };
 
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -3,6 +3,7 @@
 import { Far } from '@endo/marshal';
 import { makeStoredPublisherKit } from '@agoric/notifier';
 import { keyEQ } from '@agoric/store';
+import { objectMap, dropContext } from '@agoric/vat-data';
 import { assertElectorateMatches } from './contractGovernance/paramManager.js';
 import { makeParamManagerFromTerms } from './contractGovernance/typedParamManager.js';
 
@@ -43,6 +44,8 @@ const facetHelpers = (zcf, paramManager) => {
   const { electionManager } = terms;
 
   /**
+   * Add required methods to publicFacet
+   *
    * @template {{}} PF public facet
    * @param {PF} originalPublicFacet
    * @returns {GovernedPublicFacet<PF>}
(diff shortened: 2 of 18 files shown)


packages/governance/src/contractGovernor.js
```diff
@@ -98,11 +98,11 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
  * contract (as electionManager) so clients will be able to look up the state
  * of the governed parameters.
  *
- * @template {object} PF Public facet of governed
+ * @template {{}} PF Public facet of governed
```
erights, line 101: What's the difference between `{object}` and `{{}}`? Does the latter claim that the object is empty, while the first makes no such claim? Are PFs empty?

There are many such `{{}}` introduced by this PR. Whatever the answer is, does it apply to all of them?

packages/governance/src/types.js
```diff
@@ -497,11 +497,12 @@
...
 
 /**
- * @template {object} PF Public facet of governed contract
+ * @template {{}} PF Public facet of governed contract
+ * @template {{}} CF Creator facet of governed contract
  * @typedef {object} GovernedContractFacetAccess
  * @property {VoteOnParamChanges} voteOnParamChanges
  * @property {VoteOnApiInvocation} voteOnApiInvocation
- * @property {() => Promise<LimitedCreatorFacet<any>>} getCreatorFacet - creator
+ * @property {() => Promise<LimitedCreatorFacet<CF>>} getCreatorFacet - creator
```
turadg, line 505: _smiling_

packages/governance/src/contractHelper.js
```diff
@@ -42,25 +43,44 @@ const facetHelpers = (zcf, paramManager) => {
...
+  /**
+   * Add required methods to publicFacet, for a virtual/durable contract
+   *
+   * @param {{}} originalPublicFacet
+   */
+  const augmentVirtualPublicFacet = originalPublicFacet => {
+    return Far('publicFacet', {
+      ...originalPublicFacet,
+      ...commonPublicMethods,
+      ...objectMap(typedAccessors, dropContext),
```
turadg, line 77: `dropContext` name confused me. I read it as the opposite, taking a function with context and making it one without. I see the logic now in that it wraps in functions that drop context before passing it on.

@erights I think as a verb it should describe what it does to the underlying object to make it into the return type. What it's returning is a function that drops context, but it isn't itself dropping context. So perhaps `ignoreContext`.

Not a blocker and maybe out of scope.

packages/governance/src/contractHelper.js
```diff
@@ -94,11 +103,46 @@ const facetHelpers = (zcf, paramManager) => {
...
+    // exclusively for contractGovernor, which only reveals limitedCreatorFacet
+    // @ts-expect-error type confusion
+    return governorFacet(limitedCreatorFacet, governedApis);
+  };
+
+  /**
+   * Add required methods to a creatorFacet for a virtual/durable contract.
+   *
+   * @param {{ [methodName: string]: (context?: unknown, ...rest: unknown[]) => unknown}} originalCreatorFacet
+   * @param {Record<string, (...args: any[]) => any>} governedApis
```
turadg:
```suggestion
   * @param {Record<string, (...args: unknown[]) => unknown>} governedApis
```

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -51,24 +61,70 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
...
  */
-const start = async (zcf, privateArgs) => {
+const start = async (zcf, privateArgs, baggage) => {
   /**
    * Used to look up the unique keyword for each brand, including RUN.
    *
    * @type {MapStore<Brand, Keyword>}
    */
-  const keywordForBrand = makeStore('keywords');
+  const keywordForBrand = provideDurableMapStore(baggage, 'keywordForBrand');
```
turadg: 👍 naming

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -51,24 +61,70 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
...
+    baggage,
+    'liquidityBrandForBrand',
+  );
+
+  const takeFeeMint = async () => {
+    if (baggage.has('runMint')) {
+      return baggage.get('runMint');
+    }
+
+    const runMintTemp = await zcf.registerFeeMint(
```
turadg: any hazard that this function sometimes returns sync and sometimes awaits? no, I don't think so in this case.

packages/run-protocol/src/reserve/assetReserve.js
```diff
@@ -51,24 +61,70 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
...
+      privateArgs.feeMintAccess,
+    );
+    baggage.init('runMint', runMintTemp);
+    return runMintTemp;
+  };
+  /** @type {ZCFMint} */
+  const runMint = await takeFeeMint();
+  const runKit = runMint.getIssuerRecord();
+
+  const getTotalFeeMinted = () =>
```
turadg: this works, but all these baggage operations make me think it's worth having a singleton durable kind for the reserve contract (à la VaultDirector for VaultFactory). That's a significant refactoring so non-blocking. More of a discussion item on how we do these durability transformations.

turadg (commented):
No blocking requests but I'll hold Approve until it's passing tests so. I can help with the TS.

packages/governance/src/contractHelper.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: AIUI, the fact that the function is itself async takes care of that. Even when the conditional is true, the caller gets a promise for the result of `baggage.get()`.

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: I'm convinced that it would be an improvement to hold these values in `state` rather than lexically. @Turadg and I agreed that it makes sense to hold ephemeral state lexically in variables to be accessed directly, but durable state is simpler in `state`: you have to access it via the context, but you don't have to remember to update the baggage, which is what caused me to make all these helper methods.

Chris-Hibbert: I moved the mutable variables into state, and left the stores outside of state.  This ended up causing me to add helper functions that will be needed by other governed contracts that use state.

packages/governance/src/contractHelper.js
Chris-Hibbert: done

packages/run-protocol/src/reserve/assetReserve.js
Chris-Hibbert: I think I responded previously, but I don't see it here, so I'll re-post.

AIUI, no, there's no hazard. If the conditional is true, the caller will quickly get a resolved Promise for the mint retrieved from the baggage. Otherwise they'll have to wait for the round-trip before getting a resolved promise. If they don't need to resolve the promise, they don't have to wait. If they do, they'll add a turn whichever path we took here.

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
```diff
@@ -14,7 +14,7 @@ const { quote: q } = assert;
  * @param {ERef<ZoeService>} zoe
  * @param {(string:string) => undefined} log
  * @param {Record<string,Installation>} installations
- * @param {ERef<GovernedContractFacetAccess<unknown>>} contractFacetAccess
+ * @param {ERef<GovernedContractFacetAccess<{},{}>>} contractFacetAccess
```
turadg, line 17: I think `unknown` would be more accurate than the empty object, but if that opens a can of worms nevermind :)

packages/inter-protocol/src/reserve/assetReserve.js
```diff
@@ -30,6 +30,15 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
  *   has not yet been compensated.
  */
 
+/**
+ * @typedef {{
+ *   state: {},
+ *   facets: { AssetReservePublicFacet, AssetReserveCreatorFacet },
```
turadg: `facets` describes an object. I think you want something like,
```suggestion
 *   facets: {
 {
    publicFacet: AssetReservePublicFacet,
    creatorFacet: AssetReserveCreatorFacet,
    shortfallReportingFacet: unknown,
    limitedCreatorFacet: AssetReserveLimitedCreatorFacet,
    governedApis: unknown,
  }
```
I put `unknown` for what I didn' tknow

packages/inter-protocol/src/reserve/assetReserve.js
```diff
@@ -51,24 +60,50 @@ const nonalphanumeric = /[^A-Za-z0-9]/g;
  *   marshaller: ERef<Marshaller>,
  *   storageNode: ERef<StorageNode>,
  * }} privateArgs
+ * @param {Baggage} baggage
  */
-const start = async (zcf, privateArgs) => {
+const start = async (zcf, privateArgs, baggage) => {
+  // This contract mixes two styles of access to durable state. durableStores
```
turadg, line 79: wfm. thanks for explaining.

packages/inter-protocol/src/reserve/assetReserve.js
```diff
@@ -95,17 +133,12 @@ const start = async (zcf, privateArgs) => {
...
-  saveBrandKeyword(runKit.brand, RunKW);
-  // no need to saveIssuer() b/c registerFeeMint does it
-
   /**
+   * @param {MethodContext} _context
    * @param {Issuer} issuer
    * @param {string} keyword
    */
-  const addIssuer = async (issuer, keyword) => {
+  const addIssuer = async (_context, issuer, keyword) => {
```
turadg, line 154: we don't have a style for this yet but my 2¢ would be to leave it `context` so it git blame doesn't churn when you want to start or stop reading from it in the body. There's a reason the lint rules don't complain about it.

turadg (approved):
My only comments are on types and style. I see you don't have an `automerge` label so I'll mark Approve and trust you consider the feedback before merging.

packages/inter-protocol/src/reserve/assetReserve.js
erights: I thought our lint rules do complain about unused params that don't begin with underbar. ??

packages/run-protocol/src/reserve/assetReserve.js
erights: Yes, that's an interesting observation. @dtribble and I had a conversation with the same conclusion: The singleton pattern as enshrined in the singleton helpers work fine when all the per-singleton durable state is lexical `const`, i.e., never updated by assignment. But if any per-singleton state variables are updated by the singleton object, then it is better to put it into the state record, which currently requires manually writing a variation of the singleton pattern without help from the helper.

packages/governance/src/contractHelper.js
erights: I like `ignoreContext`.

I do think it is out of scope for this PR. But please let's do this renaming soon.

erights (commented):
Just a few more comments so far. We will need a discussion before I can do a full review. But I did not spot anything alarming.

NLBTM --- Nothing Looked Bad To Me ;)

Chris-Hibbert:
[found this uncompleted commented waiting to be finished. Sorry I didn't send it earlier.  --Chris]

@erights   I used a kind of mixed style in AssetReserve.  Mutable variables are in `state`, while stores are accessed lexically.  (AssetReserve is effectively a singleton, but I agree with @turadg's observation that the access via state seems more consistent with other durable services.  I don't find the mix of styles grating, but I'd appreciate another viewpoint.

packages/governance/src/contractGovernor.js
turadg: `object` is a JSDoc convention which is treated by TypeScript as `any`.

`{}` means an empty object. PF's extend that.

packages/inter-protocol/src/reserve/assetReserve.js
turadg: Not in param lists.  I figure because you necessarily have to bind earlier params to receive the latter ones.

IDE does give you a hint about it by dimming but it's not an error or warning.
Screen Shot 2022-07-25 at 10 12 43 AM

packages/governance/src/contractHelper.js
turadg: I'll take that on after this lands.

Chris-Hibbert: Would you mind linking an issue here?

packages/inter-protocol/src/reserve/assetReserve.js
erights: I think I prefer to keep the underbar for unused variables in general.

packages/governance/src/contractGovernor.js
erights: I see. Should many of our existing `{object}`s be `{{}}`s instead?

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
Chris-Hibbert: I get `unknown does not satisfy the constraint '{}'`.

packages/inter-protocol/src/reserve/assetReserve.js
Chris-Hibbert: I like the underscore here as well.

packages/inter-protocol/src/reserve/assetReserve.js
Chris-Hibbert: done

packages/governance/src/contractGovernor.js
Chris-Hibbert: many, yes.  Unfortunately, some type declarations work better with `object`, and I don't know enough to say why.

`{}` means that the type requires at least an empty object. It doesn't require that it be actually empty. AIUI, typescript can infer more sometimes about what methods the actual type will contain.

For instance,`CF` represents a CreatorFacet, and they're often not empty. This declaration lets us say that something that starts with a `CF`, and produces a `LimitedCreatorFacet<CF>` will produce a `LimitedCreatorFacet` in a particular contract that has all the methods of the `CF`, plus `getContractGovernor`.

```
/**
 * @template {{}} CF
 * @typedef {CF} LimitedCreatorFacet
 *
 * The creatorFacet for the governed contract that will be passed to the
 * responsible party. It does not have access to the paramManager.
 * @property {() => Instance} getContractGovernor
 */
```

erights (approved):
LGTM

Issue #5575, opened 2022-06-10
Intermittent swingset test error, possibly due to gc non-determinism


erights:
Captured at https://github.com/Agoric/agoric-sdk/runs/6838366644?check_suite_focus=true 

Went away when I reran jobs.

Relevant part seems to be

```
virtualObjects › virtualObjectGC › VO refcount management 3 unfaceted
  Difference:
  - undefined
  + {
  +   key: 'vom.rc.o+12/3',
  +   result: '1',
  +   type: 'vatstoreGet',
  + }
  › validate (packages/SwingSet/test/liveslots-helpers.js:278:7)
  › voRefcountManagementTest3 (packages/SwingSet/test/virtualObjects/test-virtualObjectGC.js:1341:3)
  › async packages/SwingSet/test/virtualObjects/test-virtualObjectGC.js:1373:3
```

erights:
Full trace

```
[Skip to content](https://github.com/Agoric/agoric-sdk/runs/6838366644?check_suite_focus=true#start-of-content)
Search or jump to…
[Pull requests](https://github.com/pulls)
[Issues](https://github.com/issues)
[Marketplace](https://github.com/marketplace)
[Explore](https://github.com/explore)
 
@erights 
https://github.com/Agoric
/
[agoric-sdk](https://github.com/Agoric/agoric-sdk) 
Public
[Code](https://github.com/Agoric/agoric-sdk)
Issues
861
Pull requests
76
[ZenHub](https://github.com/Agoric/agoric-sdk/runs/6838366644?check_suite_focus=true#zenhub)
[Discussions](https://github.com/Agoric/agoric-sdk/discussions)
[Actions](https://github.com/Agoric/agoric-sdk/actions)
Projects
7
[Wiki](https://github.com/Agoric/agoric-sdk/wiki)
Security
16
[Insights](https://github.com/Agoric/agoric-sdk/pulse)
[Settings](https://github.com/Agoric/agoric-sdk/settings)
fix: patch mistaken overrides Test all Packages #14818
[Summary](https://github.com/Agoric/agoric-sdk/actions/runs/2477610360)
Jobs
[build (14.x)](https://github.com/Agoric/agoric-sdk/runs/6838550177?check_suite_focus=true)
[build (16.x)](https://github.com/Agoric/agoric-sdk/runs/6838550305?check_suite_focus=true)
[lint-primary](https://github.com/Agoric/agoric-sdk/runs/6838550639?check_suite_focus=true)
[...]

PR #5793, opened 2022-07-20
upgrade to Ava 4.3

### LINKED ISSUE #4895: Upgrade Ava to version 4

turadg:
closes: #4895
supplants: #5201

## Description

Upgrade Ava to v4 for the reasons in #4895 and to perhaps help debug https://github.com/Agoric/agoric-sdk/issues/5575 . 

### Security Considerations

--

Ava version will now match the docs on the Ava repo.

### Testing Considerations

A few repo integration tests are breaking due to conflicting versions. Only non-required ones are left and I'll burn those down soon.
- https://github.com/Agoric/documentation/pull/688
- https://github.com/Agoric/dapp-card-store/pull/55


packages/SwingSet/package.json
```diff
@@ -77,7 +77,8 @@
     "files": [
       "test/**/test-*.js"
     ],
-    "timeout": "20m"
+    "timeout": "20m",
+    "workerThreads": false
```
warner, line 81: After this lands, I'd be curious to remove this and see if it affects anything. If "workerThreads" share a JS heap, then we'll probably need to keep it disabled. But maybe they only share a unix process memoryspace, and still have separate instances of the V8 engine (and thus separate JS heaps, so their GC doesn't affect each other), in which case we could use them.

warner (approved):
If all the tests are still passing, this sounds good to me.

packages/SwingSet/package.json
turadg: Yeah I would like to remove this config wherever possible. In my testing these packages failed without this, but maybe we can solve those cases.

PR #5804, opened 2022-07-22
feat: store dapps in localstorage not contract

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/4398

This establishes the pattern for moving certain backend data into the browser's localstorage, starting with dapp connections. A dapp can now connect directly to the wallet UI, go through the approval flow, and view the user's purses without going through an ag-solo.

After this: Implement addOffer in the smart wallet bridge and store pending offers in local storage until they're approved.

We should eventually have a "backup on chain" feature that saves the state of the local storage in the smart wallet, but not in the scope of this PR.


packages/wallet/ui/src/components/SmartWalletConnection.jsx
```diff
@@ -102,6 +108,8 @@ const SmartWalletConnection = ({
     return () => {
       cancelIterator && cancelIterator();
       cancelIterator = undefined;
+      cleanupStorageBridge = cleanupStorageBridge();
```
turadg: typo?
```suggestion
      cleanupStorageBridge && cleanupStorageBridge();
```

packages/wallet/ui/src/service/Dapps.js
```diff
@@ -0,0 +1,74 @@
+import { makeNotifierKit } from '@agoric/notifier';
+import {
+  loadDapps as load,
+  removeDapp as remove,
+  upsertDapp as upsert,
+} from '../store/Dapps.js';
+
+export const getDappService = publicAddress => {
+  const dapps = new Map();
+  const { notifier, updater } = makeNotifierKit();
```
turadg, line 10: we've been moving to `makePublishKit`. Because in the UI you want lossiness and the state retention isn't a burden, Notifier still works as well here but you might want to move eventually for consistency.

packages/wallet/ui/src/service/Dapps.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const deleteDapp = origin => {
+    dapps.delete(origin);
+    remove(publicAddress, origin);
+    broadcastUpdates();
+  };
+
+  const setDappPetname = (origin, petname) => {
+    const dapp = dapps.get(origin);
+    if (!dapp) return;
```
turadg: why fail silently?

packages/wallet/ui/src/service/Dapps.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const newValue = { ...dapp, petname };
+    upsertDapp(newValue);
+  };
+
+  const enableDapp = origin => {
+    const dapp = dapps.get(origin);
+    if (!dapp) return;
```
turadg: ditto failling silently

packages/wallet/ui/src/service/Dapps.js
```diff
@@ -0,0 +1,74 @@
...
+    remove(publicAddress, origin);
+    broadcastUpdates();
+  };
+
+  const setDappPetname = (origin, petname) => {
+    const dapp = dapps.get(origin);
+    if (!dapp) return;
+
+    const newValue = { ...dapp, petname };
+    upsertDapp(newValue);
```
turadg: consider inlining
```suggestion
    upsertDapp({ ...dapp, petname });
```

packages/wallet/ui/src/service/Dapps.js
```diff
@@ -0,0 +1,74 @@
...
+  storedDapps.forEach(d => {
+    let enableAction;
+    const approvedP = new Promise(res => {
+      enableAction = () => {
+        enableDapp(d.origin);
+        res();
+      };
+    });
+    if (d.enable) {
+      enableAction();
```
turadg, line 47: I have a hard time following the promise flow here. It looks like you want calling `enableAction` to `enableDapp` and then resolve `approvedP`. If `d.enable` is set then it should be called now and if it's not then let the dapp call it in `actions.enable`. I can't tell whether the dapp needs to call `enableAction` again if it's already succeeded (like does it still need to take `enableAction` if it's already enabled).

Though I tried to come up with something cleaner and I don't have anything better. Not requesting code changes but some comments.

packages/wallet/ui/src/components/SmartWalletConnection.jsx
samsiegart: Yes good catch! Fixed.

packages/wallet/ui/src/service/Dapps.js
samsiegart: I suppose that's not so good is it? Just added an assert here.

packages/wallet/ui/src/service/Dapps.js
samsiegart: Done

packages/wallet/ui/src/service/Dapps.js
samsiegart: ditto

packages/wallet/ui/src/service/Dapps.js
samsiegart: Yea sounds like you understand correctly. I think the hairiest part is how we can't store the actual promise, so we have to reconstruct it like this when loading from storage.

`enableAction` doesn't need to be called more than once (unless you refresh and load the dapp again from local storage). It doesn't prohibit you from calling it more than once though, I just didn't think it was necessary to restrict.

packages/wallet/ui/src/service/Dapps.js
samsiegart: Yea I wasn't sure exactly how `makePublishKit` worked, I was thinking notifier was okay in the client still, but maybe we're going to deprecate it?

turadg: that may happen. I guess we can punt until/if it does

samsiegart:
Addressed some feedback, PTAL. I'm especially interested to know if you find this general code design/organization pattern to be sound.

turadg (approved):
:shipit:

PR #3778, opened 2021-08-30
WIP: restart labeled time PR

erights:
Try to shed possible binary corruption in #3749

PR #5809, opened 2022-07-23
WIP branded time

### LINKED ISSUE #5798: add and fully adopt "timer brand": identify a source of time without granting wakeup/repeater authority

Today we discussed an attack in which off-chain clients use the published "terms" of a contract to access the TimerService it is using, and then instruct that service to make a bunch of wakeup calls, or a high-frequency repeater, as a DoS vector.

"Time" in distributed software is a fuzzy thing at best, and we don't want to enshrine one particular notion of time as the "one true time". Also, since clocks (at least high-resolution ones) enable code to read from a timing-based covert channel, we certainly don't want confined code to get ambient access to a clock. I like our current approach in which time sources are represented as ocaps, with an object that grants "what time is it now?" queries as well as "wake me up in 10 minutes" commands. This also matches well with having multiple sources of time: you just get multiple timer service objects.

Contracts publish which notion of time they're using so that clients can correctly compare things like option expiration dates. If the contract says the option expires at time 1234 according to clock A, and I want to compose that with some other contract which makes a derivative right that expires at time 1230, I want to make sure the second contract is also using clock A, otherwise bad things will ensue. So we need clocks to be comparable, independent of the authority we're sharing.

So we need contracts to publish a *description* of the kind of TimerService they're using, but we don't want this to enable an outsider to actually use that service.

After reading https://github.com/Agoric/agoric-sdk/issues/3612#issuecomment-932358730  and https://github.com/Agoric/agoric-sdk/issues/3612#issuecomment-939534241 , I'm thinking that this is similar to the way a Brand object describes a token type, without (directly?) granting the authority to e.g. make new Purses of that type. (I'm not sure whether Brands are the right notion to use, or Issuers, but the vernacular usage of "Brand" feels appropriate).

If each TimerService had a related (powerless) TimerBrand or TimerIdentity object, then the contract terms could safely cite the brand. Or, better yet, whenever they needed to reference a time value, they could instead emit a `{ time, timerID }` pair, in much the same way our Amount pattern combines a number (for "Nat" -type Amounts) and a Brand. Two Amounts with the same Brand are comparable (X > Y, X >= Y, X === Y), while Amounts with different Brands are not comparable. A client who asks contract A when the option expires, and wants to ensure that contract B's option lasts at least as long, will want to do `expiryA <= expiryB`, and that fits very nicely into the AmountMath utilities (which would throw an error if the compared values have different Brands).

## Description of the Design

So I think the swingset-side task is:
* add `timerService.getIdentity() -> TimerIdentifier`
  * the identity is a powerless object: the only method I can imagine putting on it is a string-returning `getDescription`, but I'm not sure what value I'd use on the chain. `agoric-chain-timer` ??

Elsewhere within our chain (I'll make a separate ticket for this), we need to make sure the full-powered TimerService is unavailable outside the contracts that need it. That means:
* don't publish it to the Board
  * instead, publish the TimerIdentity, which is powerless and safe to widely distribute
* remove the TimerService from the contract's publically-visible terms
  * instead, include the TimerIdentity
* remove the chain TimerService from the batch of `chainServices` supplied to newly-provisioned ag-solo clients, replacing it with the identity
* when complete, only the core-bootstrap code should hang on to the TimerService in a way that could be provided to new code (via the Big JS Hammer `eval()` operation)
  * our initial contracts are given access to the TimerService when they are started
  * but no other contracts can get it unless instantiated via the Big JS Hammer

A deeper change, which I think is interesting (but I don't know how much work would be involved) would be to define an Amount-like `{ time, timerID }` structure, maybe using `AmountMath.Nat` to manipulate it, and change all timer-sensitive contracts (at least the Vault, which needs to compute interest on a particular schedule) to use them. That would remove the TimerID from the contract's terms, and instead include it in each time value that crosses its API.

## Security Considerations

Preventing access to the TimerService is important to prevent DoS attacks that leverage the timer.

## Test Plan

Simple unit test which fetches a TimerID twice and asserts they are identical, plus one that demonstrates the lack of `setWakeup` or other API methods on the ID object.

erights:
Maybe fixes https://github.com/Agoric/agoric-sdk/issues/5798

Adapted from https://github.com/Agoric/agoric-sdk/pull/3778

Draft --- does not work yet. Not ready for review. But @warner you should look at it and let me know if you think this would be helpful.

### DIFF
--- a/packages/SwingSet/src/index.js
+++ b/packages/SwingSet/src/index.js
@@ -21,3 +21,7 @@ export { buildPlugin } from './devices/plugin/plugin.js';
 
 // eslint-disable-next-line import/export
 export * from './types-external.js';
+
+export { TimeMath } from './vats/timer/timeMath.js';
+
+export * from './vats/timer/typeGuards.js';
--- a/packages/SwingSet/src/vats/timer/timeMath.js
+++ b/packages/SwingSet/src/vats/timer/timeMath.js
@@ -15,13 +15,13 @@ const sharedTimeAuthority = (left, right) => {
     if (typeof right === 'bigint') {
       return undefined;
     } else {
-      return right.timerBrand;
+      return right.timeAuthority;
     }
   } else if (typeof right === 'bigint') {
-    return left.timerBrand;
+    return left.timeAuthority;
   } else {
-    const result = left.timerBrand;
-    assert.equal(result, right.timerBrand);
+    const result = left.timeAuthority;
+    assert.equal(result, right.timeAuthority);
     return result;
   }
 };
@@ -34,10 +34,10 @@ const sharedTimeAuthority = (left, right) => {
  */
 const absLike = (left, right, absoluteTimeValue) => {
   Nat(absoluteTimeValue);
-  const timerBrand = sharedTimeAuthority(left, right);
-  if (timerBrand) {
+  const timeAuthority = sharedTimeAuthority(left, right);
+  if (timeAuthority) {
     return harden({
-      timerBrand,
+      timeAuthority,
       absoluteTimeValue,
     });
   } else {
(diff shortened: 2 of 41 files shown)


erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/5821

PR #5803, opened 2022-07-21
refactor(run-protocol): rename paths to inter-protocol

turadg:
refs: #5802 5585

## Description

Rename the _file paths_ from `run-protocol` to `inter-protocol`. This leaves the package name alone so shouldn't break anything.

I figure we want to get this in sooner than later to have fewer merge conflicts.

Related,
- #5802 

### Security Considerations

--
### Documentation Considerations

--

### Testing Considerations

--

PR #5812, opened 2022-07-25
last references to run-protocol

turadg:
Missed a few in #5803

### DIFF
--- a/packages/cosmic-swingset/Makefile
+++ b/packages/cosmic-swingset/Makefile
@@ -49,30 +49,27 @@ $(EVAL_CLEAN): $(EVAL_CODE)
 	./scripts/clean-core-eval.js $< >$@.t || { rm -f $@.t; exit 1; }
 	mv $@.t $@
 
-RUN_PROTO=../run-protocol
+INTER_PROTO=../inter-protocol
 
-start-runStake: $(RUN_PROTO)/gov-runStake.js
-	make EVAL_CODE=$(RUN_PROTO)/gov-runStake.js \
-		EVAL_PERMIT=$(RUN_PROTO)/gov-runStake-permit.json \
+start-runStake: $(INTER_PROTO)/gov-runStake.js
+	make EVAL_CODE=$(INTER_PROTO)/gov-runStake.js \
+		EVAL_PERMIT=$(INTER_PROTO)/gov-runStake-permit.json \
 		VOTE_PROPOSAL=$(VOTE_PROPOSAL) scenario2-core-eval scenario2-vote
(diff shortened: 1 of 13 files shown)


packages/cosmic-swingset/Makefile
```diff
@@ -71,8 +71,8 @@ gov-q:
 	$(AGCH) query gov proposals --output json | \
 		jq -c '.proposals[] | [.proposal_id,.voting_end_time,.status]'
 
-deploy-runStake: ../run-protocol/test/runStake/runStake-deploy.js
-	echo agoric deploy ../run-protocol/test/runStake/runStake-deploy.js
+deploy-runStake: ../inter-protocol/test/runStake/runStake-deploy.js
+	echo agoric deploy ../inter-protocol/test/runStake/runStake-deploy.js
```
dckc: Please delete this obsolete Makefile rule.

There is no `runStake-deploy.js` file in the repo. I suspect I had something by that name locally at the time, but I don't even have that. I suppose it's obsolete in favor of stuff like https://github.com/Agoric/agoric-sdk/blob/master/packages/inter-protocol/src/proposals/core-proposal.js

dckc (approved):
runStake-deploy.js is obsolete; I prefer that we remove the reference to it.

Not critical, though.

packages/cosmic-swingset/Makefile
turadg: Oh good. `measure.js` may be similarly dead so I'll wait for an answer on that.

Chris-Hibbert (approved):
I don't have an opinion on measure.js, but if it stays, this change is necessary.

warner:
`measure.js` (and indeed most of `packages/SwingSet/misc-tools/`) is an ad-hoc tool we put together a while ago to get a sense of how big a computron is; how much meter is consumed by various operations. For some reason it reminds me of the little "how to think in metric" booklet I got as a child, with pictures of different items and how much they weighed/etc in an unfamiliar set of units.

I'd like to retain those tools, even if they're somewhat bit-rotted, as it will make it that much easier for us to dust off the next time we want to ask those questions. I'm ok if they don't run right now, but e.g. updating the paths to match changes elsewhere in the SDK will keep them fresh-*er*, and is exactly why I keep them checked in.

warner (approved):
`measure.js` change looks good to me, thanks

turadg:
This PR [failed integration test](https://github.com/Agoric/agoric-sdk/runs/7506282693?check_suite_focus=true) but I don't see why.

PR #5813, opened 2022-07-25
5346 public subscribers

### LINKED ISSUE #5346: Replace uses of Notifier with new prefix-lossy Subscriptions

turadg:
closes: #5346

## Description

Finishes removing notifiers from Inter protocol. Punts https://github.com/Agoric/agoric-sdk/issues/5814

### Security Considerations

--

Documented as breaking change.

### Testing Considerations

Made a `legacyOfferResult` wrapper to defer major refactoring of test-vaultFactory to after MN-1. It should use the new `driver.js` when that happens.

### DIFF
--- a/packages/inter-protocol/src/runStake/runStake.js
+++ b/packages/inter-protocol/src/runStake/runStake.js
@@ -192,7 +192,8 @@ export const start = async (
     const { helper, pot } = makeRunStakeKit(zcf, seat, manager);
 
     return harden({
-      publicNotifiers: {
+      publicSubscribers: {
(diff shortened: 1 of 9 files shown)


samsiegart (approved):
Looks good, thanks! We probably want `publicSubscribers` to just map to the casting spec path strings eventually for smart wallet, but that doesn't seem like the scope of this PR.

Issue #5801, opened 2022-07-21
(maybe) make ZCF sub-baggage durability be configurable


warner:
## What is the Problem Being Solved?

@turadg ran into a problem while working on durablizing the Vault contract: he tried to store the `zcf` reference in the contract's durable state, when in fact the `zcf` object that ZCF gives to the contract is ephemeral (a new one created on each upgrade). That was a bug, and it triggers the expected "cannot store non-durable data in a durable object/collection" error. But the problem was that `yarn test` did not catch the problem. The first automation that spotted the problem was the "loadgen still works" test that doesn't run until the PR is marked as `automerge` (@mhofman configured Mergify to not run this until after we start the merge process, because it's so slow and CPU-intensive, taking 30-45 minutes sometimes).

@turadg set out to add a test that would catch this sort of regression much earlier, because the debug-test-fix-repeat loop is really painful when it involves 30 minutes of CI and a PR approval process each time, not to mention running the risk of actually merging the results if the loadgen step passes. (We later learned about the `force:integration` label that causes this loadgen test to be run right away, which at least removes the PR approval and `automerge`-label and merge risk from the loop).

The problem, though, is that we made a [conscious decision](https://github.com/Agoric/agoric-sdk/blob/3bc3005586cd97e66d3de81ebd4756519f208ea5/packages/SwingSet/tools/fakeVirtualSupport.js#L280) to set `relaxDurabilityRules: true` in the `tools/prepare-test-env.js` module that we use on all "unit" unit tests (e.g. `packages/zoe/test/unitTests/`). We *do* enforce durability checks on the "swingset" unit tests (e.g. `packages/zoe/test/swingsetTests/`), which build a full swingset kernel, but these take more work to set up.

The `relaxDurabilityRules` flag was created to support the incremental conversion of code along the ephemeral -> virtual -> durable process, especially the interaction between code that is fully durable and code that has not yet gotten that far. In addition, it supports the testing of code that "normally" (in the production environment) expects a vat boundary between the different components, but does not have one in the mere "unit" unit test environment (since there are no vats without a full swingset kernel).

[...]

PR #5820, opened 2022-07-25
fix(vault): move outerUpdater out of durable state

turadg:
## Description

Error introduced by https://github.com/Agoric/agoric-sdk/pull/5736 . Not caught in CI due to https://github.com/Agoric/testnet-load-generator/issues/89

### Security Considerations

--

--

### Testing Considerations

We made a choice not to make durability testable before loadgen. This bug in master is relevant to https://github.com/Agoric/agoric-sdk/issues/5801


Chris-Hibbert (approved):
LGTM

PR #5821, opened 2022-07-25
fix: time as branded value

erights:
Supersedes https://github.com/Agoric/agoric-sdk/pull/5809

Prepares the ground for https://github.com/Agoric/agoric-sdk/issues/5798

Adapted from https://github.com/Agoric/agoric-sdk/pull/3778

### DIFF
--- a/packages/SwingSet/src/vats/timer/timeMath.js
+++ b/packages/SwingSet/src/vats/timer/timeMath.js
@@ -1,0 +1,264 @@
+// @ts-check
+
+// TODO Move this module somewhere more pleasantly reusable
+
+import { Nat } from '@agoric/nat';
+import { fit } from '@agoric/store';
+
+import { RelativeTimeShape, TimestampShape } from './typeGuards.js';
+import './types.js';
+
+const { details: X, quote: q } = assert;
+
+/**
+ * `agreedTimerBrand` is internal to this module.
+ *
+ * @param {TimerBrand | undefined} leftBrand
+ * @param {TimerBrand | undefined} rightBrand
+ * @returns {TimerBrand | undefined}
+ */
+const agreedTimerBrand = (leftBrand, rightBrand) => {
+  if (leftBrand === undefined) {
+    if (rightBrand === undefined) {
+      return undefined;
+    } else {
+      return rightBrand;
+    }
+  } else if (rightBrand === undefined) {
+    return leftBrand;
+  } else {
+    assert.equal(
+      leftBrand,
+      rightBrand,
+      X`TimerBrands must match: ${q(leftBrand)} vs ${q(rightBrand)}`,
+    );
+    return leftBrand;
+  }
+};
+
+/**
+ * `sharedTimerBrand` is internal to this module, and implements the
+ * transitional brand checking and contaigion logic explained in the `TimeMath`
+ * comment. It is used to define the binary operators that should follow
+ * this logic. It does the error checking between the operands, and returns
+ * the brand, if any, that should label the resulting time value.
+ *
+ * @param {Timestamp | RelativeTime} left
+ * @param {Timestamp | RelativeTime} right
+ * @returns {TimerBrand | undefined}
+ */
+const sharedTimerBrand = (left, right) => {
+  const leftBrand = typeof left === 'bigint' ? undefined : left.timerBrand;
+  const rightBrand = typeof right === 'bigint' ? undefined : right.timerBrand;
+  return agreedTimerBrand(leftBrand, rightBrand);
+};
+
+/**
+ * `absLike` is internal to this module, and used to implement the binary
+ * operators in the case where the returned time should be a `Timestamp`
+ * rather than a `RelativeTime`.
+ *
+ * @param {Timestamp | RelativeTime} left
+ * @param {Timestamp | RelativeTime} right
+ * @param {TimestampValue} absValue
+ * @returns {Timestamp}
+ */
+const absLike = (left, right, absValue) => {
+  Nat(absValue);
+  const timerBrand = sharedTimerBrand(left, right);
+  if (timerBrand) {
+    return harden({
+      timerBrand,
+      absValue,
+    });
+  } else {
+    return absValue;
+  }
+};
+
+/**
+ * `relLike` is internal to this module, and used to implement the binary
+ * operators in the case where the returned time should be a `RelativeTime`
+ * rather than a `Timestamp`.
+ *
+ * @param {Timestamp | RelativeTime} left
+ * @param {Timestamp | RelativeTime} right
+ * @param {RelativeTimeValue} relValue
+ * @returns {RelativeTime}
+ */
+const relLike = (left, right, relValue) => {
+  Nat(relValue);
+  const timerBrand = sharedTimerBrand(left, right);
+  if (timerBrand) {
+    return harden({
+      timerBrand,
+      relValue,
+    });
+  } else {
+    return relValue;
+  }
+};
+
+// For all the following time operators, their documentation is in
+// the `TimeMathType`, since that is the documentation that shows up
+// in the IDE. Well, at least the vscode IDE.
+
+const absValue = abs =>
+  typeof abs === 'bigint' ? Nat(abs) : Nat(abs.absValue);
+
+const relValue = rel =>
+  typeof rel === 'bigint' ? Nat(rel) : Nat(rel.relValue);
+
+const toAbs = (ts, brand = undefined) => {
+  if (typeof ts === 'number') {
+    ts = Nat(ts);
+  }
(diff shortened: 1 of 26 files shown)


erights:
Wow, all green. @warner thanks for your help.

This is now ready for review!

Chris-Hibbert:
I removed `Draft --- does not work yet. Not ready for review.` from the top comment.

Should it no longer be **draft**?

turadg:
I marked it Draft when I read that PR description. @Chris-Hibbert thanks for pointing out comment that it's no longer draft. I changed back to R4R.

packages/SwingSet/src/vats/timer/timeMath.js
```diff
@@ -0,0 +1,136 @@
...
+    return left.timerBrand;
+  } else {
+    const result = left.timerBrand;
+    assert.equal(result, right.timerBrand);
+    return result;
+  }
+};
+
+/**
+ * @param {Timestamp | RelativeTime} left
```
Chris-Hibbert, line 61: Pls add a description of what this does. It looks like the semantics allows unbranded times to be mixed with branded ones. Why is that sensible?

--  okay, I see that it's transitional. Please say so here if it's not intended to survive

packages/SwingSet/src/vats/timer/timed-iteration.js
```diff
@@ -36,12 +37,19 @@ export const makeTimedIterator = (
       const now = getCurrentTime();
 
       // Calculate where we are relative to our base time.
-      const currentOffset = now > baseTime ? now - baseTime : 0n;
-      const intervalRemaining = interval - (currentOffset % interval);
+      // TODO the `0n` should be branded.
```
Chris-Hibbert: Doesn't `relLike` mean that it will be as branded as the inputs were?  Isn't that all one can ask for?  Is the TODO already obsolete?

packages/SwingSet/src/vats/timer/timeMath.js
```diff
@@ -0,0 +1,136 @@
...
+  return relLike(abs1, abs2, val1 > val2 ? val1 - val2 : 0n);
+};
+
+const subtractAbsRel = (abs, rel) =>
+  absLike(abs, rel, absValue(abs) - relValue(rel));
+
+const subtractRelRel = (rel1, rel2) =>
+  relLike(rel1, rel2, relValue(rel1) - relValue(rel2));
+
+const modAbsRel = (abs, step) => {
```
Chris-Hibbert: The mod operations should return `relLike`. The result is a timespan that can be added or subtracted from another time.

packages/SwingSet/src/vats/timer/types.js
```diff
@@ -1,5 +1,53 @@
...
+// @template T @typedef {import('@endo/far').ERef<T>} ERef */
+// I get similar symptoms for Notifier
+
+/**
+ * @typedef {object} TimerBrand
+ * @property {(timer: TimerService) => ERef<boolean>} isMyTimer
+ */
+
+/**
+ * @typedef {bigint} TimestampValue
```
Chris-Hibbert, line 18: Would it help migration to make `TimestampValue` and `RelativeTimeValue` be @deprecated?

packages/inter-protocol/src/interest.js
```diff
@@ -8,6 +8,7 @@ import {
   quantize,
 } from '@agoric/zoe/src/contractSupport/ratio.js';
 import { assert, details as X } from '@agoric/assert';
+import { TimeMath } from '@agoric/swingset-vat/src/vats/timer/timeMath.js';
```
Chris-Hibbert, line 11: I'd like to see some tests of the interest library that use the new types. It's interesting that so much of `makeInterestCalculator` was untouched.

`SECONDS_PER_YEAR` really is unitless. I hope that doesn't make it hard to sunset the unitless values in the library.

packages/zoe/tools/manualTimer.js
```diff
@@ -186,7 +192,11 @@ export default function buildManualTimer(log, startValue = 0n, timeStep = 1n) {
           promiseKit.resolve(timestamp);
         },
       });
-      timer.setWakeup(delay, delayWaker);
+      // TODO Is this mismatch of absolute and relative a bug before this PR?
+      // https://github.com/Agoric/agoric-sdk/issues/5810
+      // I am temporarily extracing the bigint directly to avoid the type error,
+      // but it is clearly wrong.
+      timer.setWakeup(TimeMath.relValue(delay), delayWaker);
```
Chris-Hibbert: It's a bug.  The code in vat-timer.js adds `delay` to `now`.   This gets through tests because many of them start the clock at zero.  

If tests pass, it would be useful to just fix it here.  
```suggestion
      timer.setWakeup(TimeMath.addAbsRel(ticks, delay), delayWaker);
```

packages/SwingSet/src/vats/timer/timeMath.js
erights: Added an extended comment to `TimeMath` explaining the transitional plan and representation. Resolving.

packages/SwingSet/src/vats/timer/timed-iteration.js
erights: Good catch. Gone.

packages/SwingSet/src/vats/timer/timeMath.js
erights: Ah! I was not thinking clearly about that. Added an extended comment so other can avoid my confusion. Thanks!

packages/SwingSet/src/vats/timer/types.js
erights: Not really. We still need a name for these types, i.e., the bigint that `relValue` and `absValue` returns. The thing that we need to "deprecate" is the scalar arm of `Timestamp` and `RelativeTime`, leaving only the contents of what's currently called `TimestampRecord` and `RelativeTimeRecord`. But even names these are sensible to use during the transition, to record that a particular bit of code is only the other side. But once the transition is done those uses will get renamed to the simpler names.

So I don't know how to deprecate what we'd like to deprecate. @turadg , care to advise?

In any case, for this PR `TimeMath` now has an adequate comment explaining the transition plan and representation issues.

packages/zoe/tools/manualTimer.js
erights: Done. Letting it run in CI to see if fixing the breaks anything. Resolving. Will unresolve if anything breaks.

packages/inter-protocol/src/interest.js
erights: > SECONDS_PER_YEAR really is unitless. I hope that doesn't make it hard to sunset the unitless values in the library.

Aha! That might be why I was confused about modulus. So, if I understand:

rel % rel -> rel

rel / rel -> bigint // dimensionless

erights: > I'd like to see some tests of the interest library that use the new types. It's interesting that so much of makeInterestCalculator was untouched.

Good call. I adapted the first few tests in test-interest.js into test-interest-labeled.js . As a result, I did catch something in interest.js that needed to be fixed. PTAL

Chris-Hibbert (approved):
Fantastic!  LGTM

packages/SwingSet/src/vats/timer/types.js
turadg: > The thing that we need to "deprecate" is the scalar arm of Timestamp and RelativeTime,

Agree those should be deprecated with `@deprecate` comment.

> We still need a name for these types, i.e., the bigint that `relValue` and `absValue` returns

I'm not sure of this. Since TypeScript is structurally typed, these `TimestampValue` and `RelativeTimeValue` are only useful for documentation. Branded times is a better solution and reduces the value of these types such that they're not worth the confusion they introduce. Confusion being that they are actually the same type and passing them around like this may give a false sense of type safety.

If you want to brand the bigint like this then I think they're worth keeping around,
https://github.com/Agoric/agoric-sdk/blob/6b0deac341424da9b1e8badfbf307c37f16b4618/packages/zoe/src/zoeService/utils.d.ts#L18-L23
But I expect that when a time value brand matters it will pass around the branded object and only extract the bigint when the brand is unambiguous.

erights: Verbally discussed with @turadg and @Chris-Hibbert . Conclusion was not to address the deprecation issue in this PR. Experimenting, we did not seem to be able to deprecate a type. Resolving this conversation and moving forward with this PR.

turadg (approved):
Fine to defer the deprecation strategy. I can manage that if you like.

erights:
> Fine to defer the deprecation strategy. I can manage that if you like.

Yes, please!

PR #5826, opened 2022-07-26
fix: userSeat allocation only for testing

erights:
Renames the UserSeat allocation methods to end in "Jig" to flag that they should only be used for testing.

Insert comments to flag that we need to reexamine all uses of these for non-test production purposes.

By doing this PR, we identified two such problematic uses #5833 and #5834 . This PR adds todo comments at each of those locations pointing to their respective bugs.

### DIFF
--- a/packages/zoe/src/contractSupport/zoeHelpers.js
+++ b/packages/zoe/src/contractSupport/zoeHelpers.js
@@ -354,7 +354,10 @@ export const offerTo = async (
   const depositedPromiseKit = makePromiseKit();
 
   const doDeposit = async payoutPayments => {
-    const amounts = await E(userSeatPromise).getCurrentAllocation();
+    // TODO This uses getCurrentAllocationJig for production, and so
+    // is likely wrong
+    // https://github.com/Agoric/agoric-sdk/issues/5833
+    const amounts = await E(userSeatPromise).getCurrentAllocationJig();
 
     // Map back to the original contract's keywords
     const mappedAmounts = mapKeywords(amounts, mappingReversed);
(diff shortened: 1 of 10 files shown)


packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -354,7 +354,9 @@ export const offerTo = async (
   const depositedPromiseKit = makePromiseKit();
 
   const doDeposit = async payoutPayments => {
-    const amounts = await E(userSeatPromise).getCurrentAllocation();
+    // TODO This uses getCurrentAllocationJig for production, and so
+    // is likely wrong
+    const amounts = await E(userSeatPromise).getCurrentAllocationJig();
```
Chris-Hibbert, line 360: I think it's using `getCurrentAllocation()` as a quick hack to get the payout amounts. The alternative is to iterate over payoutPayments, mapping from keywords to issuers and retrieving amounts. Zoe is already storing amounts, so this is usually much faster.

There doesn't seem to be a possible race, since doDeposit is only called after `userSeatPromise.getPayouts()` resolves, which means the seat is closed.

Can we continue to expose `E(userSeat).getFinalAllocation()`?

Chris-Hibbert (changes requested):

packages/zoe/src/contractSupport/zoeHelpers.js
erights: I would like to stop using it in production. Figuring out how to do so can wait until after this PR. But this PR should clearly identify that we need to revisit this.

I filed https://github.com/Agoric/agoric-sdk/issues/5833 linking to this conversation, and added a link to that issue in the cautionary comment. Is that enough for this PR?

erights: I filed #5834 for the other one.

Chris-Hibbert: >  Is that enough for this PR?

Yes, it is.

Are you willing to agree that `getFinalAllocation()` wouldn't be a hazard?  I would have it return a promise that doesn't resolve until the seat exits.

Chris-Hibbert (approved):

PR #5831, opened 2022-07-26
chore: let virtual-object state be enumerable, for debugging

warner:
This helps dispel confusion when debugging, and `console.log(state)`
within a virtual-object method shows what looks like an empty object,
when in fact there are getters and setters for specific names.


FUDCo (approved):
Yeah, that makes sense.

PR #5835, opened 2022-07-27
WIP more hardens

erights:
DO NOT MERGE: Based on an ancient master

Just an experiment. Do not review

PR #5836, opened 2022-07-27
WIP yet more hardens

erights:
DO NOT MERGE: Transitively based on an ancient master

Multiline hardens, which https://github.com/Agoric/agoric-sdk/pull/5835 does not include.

Just an experiment. Do not yet review.


erights:
If we try to do this again, we should start a new PR. Closing this one as fatally stale.

PR #5837, opened 2022-07-27
chore(vats): Improve description of chain storage segment name constraints

gibson042:
## Description

Aligns the error message with the actual constraint. Also fixes a lint issue.


kriskowal (approved):
Thanks. It seems the actual limitation for using a SHA-512 is the length!

PR #5839, opened 2022-07-27
chore: test-collections.js: exercise deletion during iteration

warner:
The virtual-collections tests were exercising that new keys cannot be
added during iteration, but they did not demonstrate that keys *can*
be *deleted* during iteration.

The new test does that, and also refines when exactly the iterator
gets clobbered (the next `iter.next()` that happens after any
insertion). It also tests insertions both before and after the
iterator's position (both clobber the iterator).


FUDCo (approved):
Well OK then

Issue #5330, opened 2022-05-09
XS snapshot content deterministic after reload


mhofman:
## Describe the bug
Reloading from a snapshot and replaying doesn't yield the same snapshot content as replaying all commands from scratch.

While we're getting close to snapshot consistency without reloads thanks to #5227, there is also no reason that starting from a snapshot should result in later differences in snapshot content.

## To Reproduce
In the "deployment" integration test results, first check that `validator-swingstore-trace.diff` is empty.

Then execute:
```sh
diff -U0 -I 'set cosmos/meta.*' \
  <(head -n $(( $( cat chain-stage-0-swingstore-trace | wc -l) + $(cat chain-stage-1-swingstore-trace | wc -l) )) validator0-swingstore-trace) \
  <(cat chain-stage-0-swingstore-trace chain-stage-1-swingstore-trace) \
> monitor-vs-validator-swingstore-trace.diff
```

## Expected behavior
Only difference to the swingstore traces should be due to `cosmos/meta` key (which is used by cosmic-swingset to track which transactions have been processed), and length of the trace.

## Additional context
Snapshot consistency is needed for the ability to share validated snapshots for quick sync (#3769)

## Screenshots

<details>

<summary>monitor-vs-validator-swingstore-trace.diff</summary>

```diff
--- /dev/fd/63  2022-05-09 14:46:56.977648045 -0700
+++ /dev/fd/62  2022-05-09 14:46:56.977648045 -0700
@@ -81763,2 +81763,2 @@
-set local.v7.lastSnapshot {"snapshotID":"81461a41791141177714a6efa67ba40d95ea7cb4f0135bde13d98732a95301ff","startPos":{"itemCount":1002}}
-set local.snapshot.81461a41791141177714a6efa67ba40d95ea7cb4f0135bde13d98732a95301ff ["v7"]
+set local.v7.lastSnapshot {"snapshotID":"99cfedbb9ee19e6fb1ec96f500c1d0aa031d0fcac385dbedd60347374f334111","startPos":{"itemCount":1002}}
+set local.snapshot.99cfedbb9ee19e6fb1ec96f500c1d0aa031d0fcac385dbedd60347374f334111 ["v7"]
@@ -82052,2 +82052,2 @@
-set local.v4.lastSnapshot {"snapshotID":"bfbadf8b91d99463dfc2870bb3174031606fd9672558e3b8be8270e62d9e639a","startPos":{"itemCount":802}}
[...]

PR #5845, opened 2022-07-27
chore: update moddable SDK

mhofman:
refs: #5330

## Description

Update Moddable SDK with changes to address snapshot divergences after reload

### Security Considerations

None

None

### Testing Considerations

Locally tested that there are no longer any snapshot divergences in the loadgen deployment test (inspected logs). We don't want to enforce yet, so extracting hash divergences from swingstore traces is the next best check.


mhofman:
The other source changes seem minimal, mostly around the parser (and a plethora of device code we don't use)

warner (approved):
Looks good. Before landing, please grep `packages/xsnap/` for `xs-meter-15` and update both instances to `xs-meter-16`. I don't see any obvious XS changes that would impact metering calculations, but our policy is to bump the label each time we change XS.

Issue #5811, opened 2022-07-24
add tentative API to upgrade static vats


warner:
## What is the Problem Being Solved?

For dynamic vats, the per-vat `adminFacet` can be used to upgrade or otherwise restart that vat, with different code and/or `vatParameters` than the previous incarnation. Whoever asked for the vat to be created is given this facet, which is a very tidy way to represent the authority to influence the behavior of the vat after that initial creation event.

We do not have a corresponding control mechanism to upgrade a static vat. #5666 is about making sure all of swingset's internal vats (vat-timer, vat-vat-admin, etc) are upgradable, meaning they use durable objects, and baggage, and are prepared to behave correctly after a restart. However we don't yet have a way to trigger this restart.

We know that kernel changes are easier to deploy than vat changes, so I've not included this as a requirement for MN-1: as long as the vats themselves are prepared for upgrade, we could wait to implement the upgrade button until after MN-1 is deployed. However that does make it harder to *test* the #5666 accomplishments.

@FUDCo was looking at the dynamic vat upgrade implementation and decided that it's nearly everything we need for upgrading static vats too. The big question is where the authority lies. The two main options are 1: something new on vat-vat-admin, or 2: something on the `controller` object.

One consideration is POLA. Being able to upgrade a static vat is really powerful, because that lets you commandeer the bootstrap vat (which may still have access to every authority created in every other vat), as well as vat-vat-admin itself (which therefore lets you upgrade all dynamic vats). Currently, if you have access to the `vatAdminService`, you can create new vats, but you can't affect any existing ones that you didn't create. If we added `E(vatAdminService).upgradeStaticVat(vatID, { bundleCap, vatParameters })`, that would represent a huge increase in authority.

A second consideration is the mechanics of the chain-upgrade scenario where we anticipate needing to upgrade the static vats. If we're upgrading one of the built-in vats, there's a decent chance we'll be changing the associated device at the same time, and/or changing the kernel code (especially if we're upgrading vat-vat-admin, since it does all its work through device-vat-admin and its kernel hooks). These sorts of upgrades cannot be driven by messages to `vatAdminService`: they need to be coordinated with code that sits outside the kernel and interacts with the new device/kernel code being launched.

[...]

PR #5846, opened 2022-07-28
Make the vatAdmin vat upgradeable

### LINKED ISSUE #5669: make vat-admin upgradable

As part of #5666, we should make vat-vat-admin upgradable.

It currently exposes a singleton `VatAdminService` object (built by sending a message to the root object, connecting it to device-vat-admin). From this service you can create a dynamic vat and get back an "admin facet", with methods like `upgrade()`. You can also create Meter objects, which have an API for filling and setting notification thresholds. Meters can be passed into `createVat`. The service also provides some methods for retrieving and interrogating BundleCaps, although those are device nodes (hosted by device-vat-admin), rather than regular objects.

## Description of the Design

We'll need durable Kinds for:
* the singleton VatAdminService (which should store device-vat-admin in baggage or in `state`)
* Meters
* admin facets

## Security Considerations

None should arise because of this transformation.

## Test Plan

As with the other #5666 changes, either full unit tests of vat-vat-admin being upgraded, or the existing unit tests plus a careful audit of the new implementation for lingering non-durable state.

FUDCo:
Closes #5669 and possibly also #5811

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -147,7 +147,7 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
         bundle.moduleFormat === 'getExport',
         X`unexpected: ${bundle.moduleFormat}`,
       );
-      // eslint-disable-next-line no-await-in-loop
+      // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
       await worker.evaluate(`(${bundle.source}\n)()`.trim());
     }
     return worker;
@@ -200,6 +200,7 @@ export async function makeSwingsetController(
   }
 
   const slogF =
+    // eslint-disable-next-line @jessie.js/no-nested-await
     slogFile && (await fs.createWriteStream(slogFile, { flags: 'a' })); // append
 
   function writeSlogObject(obj) {
@@ -241,6 +242,7 @@ export async function makeSwingsetController(
   harden(console);
 
   writeSlogObject({ type: 'bundle-kernel-start' });
+  // eslint-disable-next-line @jessie.js/no-nested-await
   const { kernelBundle = await buildKernelBundle() } = runtimeOptions;
   writeSlogObject({ type: 'bundle-kernel-finish' });
 
@@ -498,6 +500,26 @@ export async function makeSwingsetController(
       }
       return kpid;
     },
+
+    upgradeStaticVat(vatName, shouldPauseFirst, bundleID, options = {}) {
+      const vatID = kernel.vatNameToID(vatName);
+      let pauseTargetBody = null;
+      let pauseTargetSlots = [];
+      if (shouldPauseFirst) {
+        pauseTargetBody = { [QCLASS]: 'slot', index: 0 };
+        pauseTargetSlots = [kernel.getRootObject(vatID)];
+      }
+      if (!options.upgradeMessage) {
+        options.upgradeMessage = `vat ${vatName} upgraded`;
+      }
+      return controller.queueToVatRoot(
+        'vatAdmin',
+        'upgradeStaticVat',
+        [vatID, pauseTargetBody, bundleID, options],
+        'ignore',
+        pauseTargetSlots,
+      );
+    },
   });
 
   return controller;
@@ -554,6 +576,7 @@ export async function buildVatController(
   const initializationOptions = { verbose, kernelBundles };
   let bootstrapResult;
   if (!swingsetIsInitialized(hostStorage)) {
+    // eslint-disable-next-line @jessie.js/no-nested-await
     bootstrapResult = await initializeSwingset(
       config,
       argv,
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -419,6 +419,7 @@ export default function buildKernel(
         //   Kinds, and we're not going to use the worker
         //
         // So in all cases, our caller should abandon the worker.
+        // eslint-disable-next-line @jessie.js/no-nested-await
         await vatWarehouse.stopWorker(vatID);
         // TODO: does stopWorker work if the worker process just died?
         status.deliveryError = deliveryResult[1];
@@ -706,6 +707,7 @@ export default function buildKernel(
     // supervisor bundles are bad.
 
     try {
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await vatWarehouse.createDynamicVat(vatID);
     } catch (err) {
       const info = makeError(`${err}`);
@@ -874,6 +876,7 @@ export default function buildKernel(
     if (results1.terminate) {
       // get rid of the worker, so the next delivery to this vat will
       // re-create one from the previous state
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await vatWarehouse.stopWorker(vatID);
 
       // notify vat-admin of the failed upgrade
@@ -918,6 +921,7 @@ export default function buildKernel(
 
     if (results2.terminate) {
       // unwind just like above
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await vatWarehouse.stopWorker(vatID);
       const vatAdminMethargs = makeFailure(results2.terminate.info);
       const results = harden({
@@ -1178,6 +1182,7 @@ export default function buildKernel(
   }
 
   async function processDeliveryMessage(message) {
+    kdebug('');
     kdebug(`processQ ${JSON.stringify(message)}`);
     kdebug(legibilizeMessage(message));
     kernelSlog.write({
@@ -1248,6 +1253,7 @@ export default function buildKernel(
       // other deliveries should be re-attempted on the next crank, so they
       // get the right error: we leave those on the queue
     } else {
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await vatWarehouse.maybeSaveSnapshot();
     }
     const { computrons, meterID } = crankResults;
@@ -1279,6 +1285,7 @@ export default function buildKernel(
       kdebug(`vat terminated: ${JSON.stringify(info)}`);
       kernelSlog.terminateVat(vatID, reject, info);
       // this deletes state, rejects promises, notifies vat-admin
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await terminateVat(vatID, reject, info);
     }
 
@@ -1311,6 +1318,7 @@ export default function buildKernel(
     }
     processQueueRunning = Error('here');
     try {
+      // eslint-disable-next-line @jessie.js/no-nested-await
       const result = await processor(message);
       processQueueRunning = undefined;
       return result;
--- a/packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
+++ b/packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
@@ -8,9 +8,17 @@
 import { makePromiseKit } from '@endo/promise-kit';
 import { makeNotifierKit } from '@agoric/notifier';
 import { Far, passStyleOf } from '@endo/marshal';
+import { E } from '@endo/eventual-send';
 import { Nat, isNat } from '@agoric/nat';
+import { provide } from '@agoric/store';
+import {
+  makeScalarBigMapStore,
+  makeScalarBigSetStore,
+  vivifyKind,
+  vivifySingleton,
+} from '@agoric/vat-data';
 
-const { details: X } = assert;
+const { details: X, quote: q } = assert;
 
 const managerTypes = ['local', 'nodeWorker', 'node-subprocess', 'xs-worker'];
 
@@ -19,20 +27,184 @@ function producePRR() {
   return [promise, { resolve, reject }];
 }
 
-export function buildRootObject(vatPowers) {
+export function buildRootObject(vatPowers, _vatParameters, baggage) {
+  const criticalVatKey = vivifySingleton(baggage, 'criticalVatKey', {});
+
   const { D } = vatPowers;
-  const pending = new Map(); // vatID -> { resolve, reject } for promise
-  const running = new Map(); // vatID -> { resolve, reject } for doneP
-  const meterByID = new Map(); // meterID -> { meter, updater }
-  const meterIDByMeter = new WeakMap(); // meter -> meterID
-  let vatAdminNode;
+  const pendingVatCreations = new Map(); // vatID -> { resolve, reject } for promise
   const pendingBundles = new Map(); // bundleID -> Promise<BundleCap>
   const pendingUpgrades = new Map(); // upgradeID -> Promise<UpgradeResults>
 
+  let vatAdminDev;
+
+  const runningVats = new Map(); // vatID -> [doneP, { resolve, reject }]
+  const exitingVats = new Set();
+  function noteRunningVat(vatID) {
+    const prr = producePRR();
+    const [doneP] = prr;
+    doneP.catch(() => {}); // shut up false whine about unhandled rejection
+    runningVats.set(vatID, prr);
+  }
+
+  const managedVats = provide(baggage, 'managedVats', () =>
+    makeScalarBigSetStore('managedVats', { durable: true }),
+  );
+  for (const vatID of managedVats.keys()) {
+    noteRunningVat(vatID);
+  }
+
+  // XXX The plan is for meters to hold their notifiers, but notifiers are not
+  // yet durable, which blocks making meters themselves durable unless the
+  // (currently) ephemeral notifiers are put someplace else.  So we're putting
+  // them in a WeakMap keyed by the meter itself.  The following has bits of
+  // code commented with either XXX TEMP or XXX RESTORE.  The former flags
+  // temporary code that should be removed once notifiers are durable.  The
+  // latter flags (commented out) code that should be made active once notifiers
+  // are durable.
+  //
+  // Note that this means that holders of notifiers (the ones obtained by
+  // calling `getNotifier` on the meter) will be left holding broken references
+  // after an upgrade.  The practical consequences of this are not yet clear to
+  // me, but hopefully this concern will soon be mooted by making notifiers
+  // actually durable.
+
+  const meterNotifiersTemp = new WeakMap(); // meter -> { notifier, updater } // XXX TEMP
+
+  const makeMeter = vivifyKind(
+    baggage,
+    'meter',
+    (remaining, threshold) => {
+      Nat(remaining);
+      Nat(threshold);
+      const meterID = D(vatAdminDev).createMeter(remaining, threshold);
+      // const { updater, notifier } = makeNotifierKit(); // XXX RESTORE
+      // return { meterID, updater, notifier }; // XXX RESTORE
+      return { meterID }; // XXX TEMP
+    },
+    {
+      addRemaining({ state }, delta) {
+        D(vatAdminDev).addMeterRemaining(state.meterID, Nat(delta));
+      },
+      setThreshold({ state }, newThreshold) {
+        D(vatAdminDev).setMeterThreshold(state.meterID, Nat(newThreshold));
+      },
+      get: ({ state }) => D(vatAdminDev).getMeter(state.meterID), // returns BigInts
+      // getNotifier: ({ state }) => state.notifier, // XXX RESTORE
+      getNotifier: ({ self }) => meterNotifiersTemp.get(self).notifier, // XXX TEMP
+    },
+    // eslint-disable-next-line no-use-before-define
+    { finish: finishMeter },
+  );
+
+  const makeUnlimitedMeter = vivifyKind(
+    baggage,
+    'unlimitedMeter',
+    () => {
+      const meterID = D(vatAdminDev).createUnlimitedMeter();
+      // const { updater, notifier } = makeNotifierKit(); // XXX RESTORE
+      // return { meterID, updater, notifier }; // XXX RESTORE
+      return { meterID }; // XXX TEMP
+    },
+    {
+      addRemaining(_context, _delta) {},
+      setThreshold(_context, _newThreshold) {},
+      get: () => harden({ remaining: 'unlimited', threshold: 0 }),
+      // getNotifier: ({ state }) => state.notifier, // will never fire // XXX RESTORE
+      getNotifier: ({ self }) => meterNotifiersTemp.get(self).notifier, // XXX TEMP
+    },
+    // eslint-disable-next-line no-use-before-define
+    { finish: finishMeter },
+  );
+
+  function finishMeter({ state, self }) {
+    // eslint-disable-next-line no-use-before-define
+    meterByID.init(
+      state.meterID,
+      // harden({ meter: self, updater: state.updater }), // XXX RESTORE
+      harden({ meter: self }), // XXX TEMP
+    );
+    // eslint-disable-next-line no-use-before-define
+    meterIDByMeter.set(self, state.meterID);
+    meterNotifiersTemp.set(self, makeNotifierKit()); // XXX TEMP
+  }
+
+  // meterID -> { meter, updater }
+  // XXX currently actually meterID -> { meter } due to notifiers not yet being durable
+  const meterByID = provide(baggage, 'meterByID', () =>
+    makeScalarBigMapStore('meterByID', { durable: true }),
+  );
+  const meterIDByMeter = new WeakMap(); // meter -> meterID
+  for (const [meterID, meterEntry] of meterByID.entries()) {
+    meterIDByMeter.set(meterEntry.meter, meterID);
+    meterNotifiersTemp.set(meterEntry.meter, makeNotifierKit()); // XXX TEMP
+  }
+
+  if (baggage.has('vatAdminDev')) {
+    vatAdminDev = baggage.get('vatAdminDev');
+  }
+
+  // Some notes on the vat pause logic: The variable `quiescentP` is set to an
+  // unresolved promise when service is paused (by sending the vat a
+  // `pauseService` message, for which this promise is the result value); the
+  // presence of a non-nullish value in this variable indicates that no service
+  // requests should be accepted (unless they can be processed in the same turn
+  // they were received, in which case the vat may, at its option, respond to
+  // them normally). The variable is cleared when service is resumed, either
+  // explicitly by sending the vat a `resumeService` message or implicitly by
+  // the vat restarting (the vat is born in an available-for-service state).
+  // The promise is fulfilled (to a boolean `true` value) when the vat quiesces,
+  // meaning that all pending non-prompt service requests have completed; the
+  // fulfillment signals to anyone watching the promise that the vat may be
+  // safely shut down.  The promise is rejected (with a 'shutdown cancelled'
+  // Error) if service is resumed prior to quiescence.  The
+  // `hasReachedQuiescence` flag is used to keep track of quiescence has been
+  // reached after pausing so that we don't try to reject a promise that has
+  // already been fulfilled.
+  let quiescentP;
+  let quiescentRR;
+  let hasReachedQuiesence = false;
+
+  function insistVatAdminServiceNotPaused() {
+    assert(!quiescentP, 'service not available, try again later');
+  }
+
+  function checkForQuiescence() {
+    if (quiescentP) {
+      if (
+        pendingBundles.size === 0 &&
+        pendingVatCreations.size === 0 &&
+        pendingUpgrades.size === 0 &&
+        exitingVats.size === 0
+      ) {
+        hasReachedQuiesence = true;
+        quiescentRR.resolve(true);
+      }
+    }
+  }
+
+  function pauseService() {
+    if (!quiescentP) {
+      [quiescentP, quiescentRR] = producePRR();
+    }
+    checkForQuiescence();
+    return quiescentP;
+  }
+
+  function resumeService() {
+    if (quiescentP) {
+      if (!hasReachedQuiesence) {
+        quiescentRR.reject(Error('shutdown cancelled'));
+      }
+      hasReachedQuiesence = false;
+      quiescentP = null;
+      quiescentRR = null;
+    }
+  }
+
   // this message is queued to us by kernel.installBundle()
   function bundleInstalled(bundleID) {
-    if (vatAdminNode && pendingBundles.has(bundleID)) {
-      const bundlecap = D(vatAdminNode).getBundleCap(bundleID);
+    if (vatAdminDev && pendingBundles.has(bundleID)) {
+      const bundlecap = D(vatAdminDev).getBundleCap(bundleID);
       if (!bundlecap) {
         // if the bundle got uninstalled by the time we got the message, keep
         // waiting, maybe it will get reinstalled some day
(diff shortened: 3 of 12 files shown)


packages/SwingSet/src/controller/controller.js
```diff
@@ -498,6 +500,27 @@ export async function makeSwingsetController(
...
+      if (shouldPauseFirst) {
+        const readyKpid = controller.queueToVatRoot(vatName, 'pauseService');
+        readyKpidBody = { [QCLASS]: 'slot', index: 0 };
+        readyKpidSlots = [readyKpid];
+      }
+      if (!options.upgradeMessage) {
+        options.upgradeMessage = `vat ${vatName} upgraded`;
+      }
+      return controller.queueToVatRoot(
+        vatName,
```
warner: Does this assume that `vatName === 'vat-admin'`? Ideally we want a way to always send the `upgradeVat` message to vat-admin, even when we're upgrading some other vat. The controller doesn't know which vat is which, but the kernel has `vatAdminRootKref`, so maybe we should have this `upgradeStaticVat` pass through to a kernel method that can talk to vat-admin more directly.

warner: talking just now, we figured this should be changed to a static string `vatAdmin`, to match the code in `initializeSwingset.js` that normally defines the vat:

https://github.com/Agoric/agoric-sdk/blob/8ed0d701b9073ff039618c0a2e3c6b04acd60177/packages/SwingSet/src/controller/initializeSwingset.js#L359

It'd be nice to know that we're using the same name in both places (i.e. define a `VAT_ADMIN_NAME = 'vatAdmin'` somewhere, use it both here and in `config.vats[VAT_ADMIN_NAME] =`, and also https://github.com/Agoric/agoric-sdk/blob/8ed0d701b9073ff039618c0a2e3c6b04acd60177/packages/SwingSet/src/controller/initializeKernel.js#L104), but I'm probably ok with our unit tests catching any potential mismatch-under-maintenance of these two.

Also, my idea about using `vatAdminRootKref` wouldn't have worked: the new `upgradeVat` method was added to the vat-vat-admin *root*, not the `vatAdminService` (which should not allow clients to upgrade static vats anyways).

packages/SwingSet/src/controller/controller.js
```diff
@@ -498,6 +500,27 @@ export async function makeSwingsetController(
...
+        'upgradeVat',
```
warner: I'd like to change this method name to `upgradeStaticVat`, to distinguish it from the `adminFacet.upgrade()` (for dynamic vats) defined in the same file.

Or, what about `upgradeNamedVat` (/`upgradeNamedStaticVat`), and pass the vat *name* through to vat-vat-admin, instead of the ID?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1178,7 +1182,7 @@ export default function buildKernel(
   }
 
   async function processDeliveryMessage(message) {
-    kdebug(`processQ ${JSON.stringify(message)}`);
+    kdebug(`\nprocessQ ${JSON.stringify(message)}`);
```
warner: nit: did you mean to leave this in? maybe use a `kdebug('')` instead of an extra newline prefix?

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -19,20 +26,141 @@ function producePRR() {
   return [promise, { resolve, reject }];
 }
 
-export function buildRootObject(vatPowers) {
+export function buildRootObject(vatPowers, _vatParameters, baggage) {
+  const markerKindHandle = provide(baggage, 'marker', () =>
```
warner: vat-data has a `provideKindHandle` utility for this pattern

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -19,20 +26,141 @@ function producePRR() {
...
+    makeKindHandle('marker'),
+  );
+  const makeMarker = defineDurableKind(markerKindHandle, () => ({}), {});
+  const criticalVatKey = provide(baggage, 'criticalVatKey', () => makeMarker());
```
warner: vat-data has a `vivifySingleton` utility that combines both this pattern and `provideKindHandle`: https://github.com/Agoric/agoric-sdk/blob/8ed0d701b9073ff039618c0a2e3c6b04acd60177/packages/vat-data/src/kind-utils.js#L122

packages/SwingSet/src/controller/controller.js
warner: Ok, now that I've read through the full implementation, I see where the factoring join point is, and why this takes `vatID` and not a name. I think I'd still like it have a distinct name, but that will vat-vat-admin to have the exported/returned root-object method (`upgradeStaticVat`) be nearly empty, and just delegate to the same shared `upgradeVat` which `adminFacet.upgrade()` does for dynamic vats. And I'm not sure that's worth it.. but let's see what you think.

packages/SwingSet/src/controller/initializeKernel.js
```diff
@@ -109,6 +109,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
         // Pin, to prevent it being GC'd when only the kvStore points to it
         kernelKeeper.pinObject(kref);
         kernelKeeper.kvStore.set('vatAdminRootKref', kref);
+        kernelKeeper.kvStore.set('vatAdminVatID', vatID);
```
warner: I see a bunch of changes to support teaching vat-vat-admin about it's own vatID.. I'm wondering if they're worth it. More details in vat-vat-admin.js.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -19,20 +26,141 @@ function producePRR() {
...
+    runningVats.set(vatID, prr);
+  }
+
+  let quiescentP;
+  let quiescentRR;
+
+  const managedVats = provide(baggage, 'managedVats', () =>
+    makeScalarBigSetStore('managedVats', { durable: true }),
+  );
+  for (const vatID of managedVats.keys()) {
```
warner, line 52: Consider `managedVats.keys().forEach(noteRunningVat)` ? I'm channeling other reviewers here, I'm only recently becoming comfortable with the more functional-programming styles myself.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -19,20 +26,141 @@ function producePRR() {
...
+    noteRunningVat(vatID);
+  }
+
+  // XXX The plan is for meters to hold their notifiers, but notifiers are not
```
warner, line 56: This temporary plan seems sound.. I tried to think of some ways to e.g. not create the Notifier until someone tried to interact with it, but the creation trigger would either be users doing `getNotifier()` or the kernel doing `meterCrossedThreshold`, and the complexity wouldn't be worth it.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
+      quiescentRR.reject(Error('shutdown cancelled'));
+      quiescentP = null;
+      quiescentRR = null;
+    }
   }
 
-  function finishVatCreation(vatID) {
-    const [promise, pendingRR] = producePRR();
-    pending.set(vatID, pendingRR);
+  function finishMeter({ state, self }) {
```
warner: code layout nit: let's move all of the `const meterByID=` and `meterIDByMeter` definitions, the  `for (const [meterID, meterEntry] of meterByID.entries()` loop, and the definition of `finishMeter` all the way up to before the definition of `makeMeter`, which would remove the forward reference for `finishMeter` and also make all the pieces easier to find

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
-    const meter = Far('meter', {
-      addRemaining(_delta) {},
-      setThreshold(_newThreshold) {},
-      get: () => harden({ remaining: 'unlimited', threshold: 0 }),
-      getNotifier: () => notifier, // will never fire
-    });
-    meterByID.set(meterID, harden({ meter, updater }));
-    meterIDByMeter.set(meter, meterID);
-    return meter;
+  function resumeService() {
```
warner, line 193: I'm not seeing anything call `resumeService` (I see it's a method on the root object, but I don't see the string anywhere else in the codebase)

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
-      setThreshold(newThreshold) {
-        D(vatAdminNode).setMeterThreshold(meterID, Nat(newThreshold));
-      },
-      get: () => D(vatAdminNode).getMeter(meterID), // returns BigInts
-      getNotifier: () => notifier,
-    });
-    meterByID.set(meterID, harden({ meter, updater }));
-    meterIDByMeter.set(meter, meterID);
-    return meter;
+  function pauseService() {
```
warner, line 185: On my first pass through, I didn't realize that `insistServiceAvailable` was using `quiescentP` as a "service is paused" flag, so I thought there was a race condition left exposed. Could you add a brief `// also used as a disable flag for insistServiceAvailable` on the next line? Or maybe name it `disabledP` or `pausedP` instead of `quiescentP` (ok I see how that would be a worse name, hm).

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -311,18 +482,33 @@ export function buildRootObject(vatPowers) {
...
+
+  function createVatAdminService(vaDevice) {
+    if (baggage.has('vatAdminDev')) {
+      baggage.set('vatAdminDev', vaDevice);
+    } else {
+      baggage.init('vatAdminDev', vaDevice);
+    }
+    vatAdminDev = vaDevice;
+    vatAdminVatID = D(vatAdminDev).getVatAdminVatID();
+    if (!vatAdminService) {
```
warner: I'd suggest moving the `makeVatAdminService` call out of this `createVatAdminService` method, doing it as a `provide` or actually a `vivifySingleton` up where you define the Kind:

```js
const vatAdminService = vivifySingleton(...);
```

That's the pattern I'm using in vat-timer: the now-somewhat-misnamed `createVatAdminService` is really just stashing the `vatAdminDev` object and returning the previously-defined service object.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -339,8 +525,23 @@ export function buildRootObject(vatPowers) {
    * @param {number} incarnationNumber
    */
   function vatUpgradeCallback(upgradeID, success, error, incarnationNumber) {
-    const { resolve, reject } = pendingUpgrades.get(upgradeID);
+    const pending = pendingUpgrades.get(upgradeID);
+    const doingSelfUpgrade = baggage.has('doingSelfUpgrade');
+    if (doingSelfUpgrade) {
+      baggage.delete('doingSelfUpgrade');
+    }
+    if (pending === undefined) {
```
warner, line 541: ok so there's a bunch of "tell vat-vat-admin about its own vatID" changes just to enable this clause.. is it worth it? What if we just said `if (!pending.has(upgradeID)) { delete, check, resolve/reject }` ?

I wondered if maybe the `resumeService` was meant to be called from here, like `if (doingSelfUpgrade)`, but that'd be wrong in two ways: `resumeService` needs to be called if the upgrade fails, not if it succeeds (because if it succeeds, the new vat code is born with the disable flag cleared), and also all vats that use the pause/resume protocol would need to be resumed, not just the local vat-vat-admin one.

I have to think more about how pause/resume ought to work. I think this code, as-is, would fail to re-enable vat-admin if the vat-vat-admin upgrade failed. We'd rewind to the old version, but the disable flag would still be set, so we wouldn't accept dynamic vat upgrade requests (or others that know they return a promise..

.. hm, but with the notable exception of `upgradeVat` itself (it looks like you can still call `vatAdminRoot.upgradeVat` even when `insistServiceAvailable` would throw). On one hand that's good: otherwise a failed vat-vat-admin upgrade would leave it bricked. On the other, it means you could e.g. `controller.upgradeStaticVat('timer')` while vat-vat-admin is supposed to be paused, so a race between the self-upgrade and the timer-upgrade could leave the timer-upgrade's promise rejected (disconnected). OT3H maybe that's both rare and not such a big deal: the kpid returned by `controller.upgradeStaticVat` is not particularly easy to wait for anyways, so I'm not sure that our upgrade-the-chain code is going to necessarily pay attention to it or doing anything in response to it fulfilling/rejecting.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
warner: Also, the pattern that @gibson042 and I were settling upon was to use e.g. `markerHandle` as the name for the baggage key. But if you use `vivifySingleton`, then `marker` would probably be the right string to use, since it will derive key names for both the handle and the instance.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
+    );
+    meterIDByMeter.set(self, state.meterID);
+    meterNotifiersTemp.set(self, makeNotifierKit()); // XXX TEMP
+  }
 
-    const [doneP, doneRR] = producePRR();
-    running.set(vatID, doneRR);
-    doneP.catch(() => {}); // shut up false whine about unhandled rejection
+  function insistServiceAvailable() {
+    assert(!quiescentP, 'service not available, try again later');
```
warner: s/service/vatAdminService/, so clients have a chance to figure out which service is unavailable.

Also, this isn't a programmer error, so I don't think it should be an `assert`. Maybe `throw Error('vatAdminService not available during self-upgrade, try again later')`.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
+    );
+    const [upgradeCompleteP, upgradeRR] = producePRR();
+    pendingUpgrades.set(upgradeID, upgradeRR);
+    return upgradeCompleteP;
+  }
+
+  function assertRunningVat(vatID, operation, exitingOK) {
+    assert(
+      runningVats.has(vatID) && (exitingOK || !exitingVats.has(vatID)),
+      `attempt to perform ${operation}() on non-running vat ${vatID}`,
```
warner: Hm, this reveals the vatID to callers, which we haven't done before.. maybe just take out that last expression? `attempt to perform ${operation}() on stale vat` ? The main way this could happen is if someone holds on to an `adminFacet` for too long, but also there are lots of race conditions that could trigger it (like if a vat dies unexpectedly and you weren't following the `done` closely enough).

Or.. this *might* be an appropriate place for the `details` guard.

warner: It might be good to put `vatAdminService` in the string, like `vatAdminService rejecting attempt to perform..`, so callers can tell who the error message is coming from. This mostly helps with `done()`, so they can tell it's not coming from the vat being watched.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
       },
-      done() {
+      done({ state }) {
+        // In some world it might make sense to allow this operation on
+        // terminated vats (since they *do* have a 'done' status after all,
+        // i.e., non-running) but we expect the promise that's returned here to
+        // reject if the vat fails, and if the vat has already exited then its
+        // exit status has been lost.  However, by rejecting the 'done'
+        // operation itself we can signal the non-running condition without
+        // promoting the illusion that we know whether it exited cleanly or not.
```
warner: Hrm, but it's not like we're boxing the `doneP` return promise. So when a client does `E(adminFacet).done().catch(err => 'I guess the vat was terminated')`, that rejection could either come from this `assertRunningVat`, or from a vat-vat-admin upgrade disconnecting its promises, or from someone doing `E(adminFacet).terminateWithFailure()`, or the vat running down its Meter, or somehow making an illegal syscall. They all show up the same way. The disconnect-because-of-vat-vat-admin upgrade should have a distinctive `.name`, but I'm not sure how you'd distinguish the others.

I'm not sure what to do here. I'm sure you considered `if (!runningVats.has(vatID)) { throw Error('done() on stale vat') }`.. well, ok, I guess that's basically what `assertRunningVat` does. Maybe just acknowledge the unavoidable ambiguity in the comment?

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -41,88 +169,114 @@ export function buildRootObject(vatPowers) {
...
+    makeKindHandle('adminNode'),
+  );
+  const makeAdminNode = defineDurableKind(
+    adminNodeKindHandle,
+    vatID => ({ vatID }),
+    {
+      terminateWithFailure({ state }, reason) {
+        insistServiceAvailable();
+        assertRunningVat(state.vatID, 'terminateWithFailure');
+        exitingVats.add(state.vatID);
```
warner: I'm trying to analyze this bit. So when someone calls `adminFacet~.terminateWithFailure()` for some random dynamic vat, we'll populate `exitingVats`, which will inhibit vat-vat-admin upgrade until the dynamic vat actually terminates (and we find out when the `vatAdminRoot.vatTerminated` callback gets run, and we do `exitingVats.delete` and then re-check for quiescence). The `assetRunningVat` protects against the `terminateWithFailure` arriving *after* the vat terminated (perhaps spontaneously, perhaps because of a rival+winning `terminateWithFailure` call): without it, we might populate `exitingVats` but have no way to un-populate it, which would inhibit future self-upgrades (although we could hack our way around that, if we needed to, by doing a chain upgrade with `controller.upgradeStaticVat('vatAdmin', shouldPauseFirst=false)`).

There's a small behavior change here relative to the old version, which seems to believe that it tolerates multiple (or just very very late) `terminateWithFailure` calls. I don't know what `D(vatAdminDev).terminateWithFailure` does in that case.. I suspect it wasn't as tolerant as the code made it appear.

So let's see, if someone fetches a vat's `done` promise with `adminFacet~.done()` before the vat gets terminated and before vat-vat-admin starts a self-upgrade, that promise will inhibit vat-vat-admin self-upgrade until the vat terminates. Which means that `done` promise *won't* reject because of vat-vat-admin self-upgrade. It will either fulfill (vat terminate-happy), or reject (vat terminate-sad). So they can rely upon rejection being a real vat-terminate-sad event.

If their `done()` request arrives after the vat is terminated, it rejects from the `assertRunningVat` message.

If it arrives after vat-vat-admin has been paused, it rejects from the `insistServiceAvailable()` check.

So if the client calls `done()` and gets a rejection, it could be any of:
* vat-terminate-sad
* vat was already terminated, we don't know its state
* vat-vat-admin is paused

and they can't easily tell the difference.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
warner: Oh, you know, maybe we should store the exit status in the admin facet `state`, so it's available for folks calling `done` much later. In the normal case, callers will drop the admin facet once they know the vat is dead (or unneeded), which would allow the exit status data to be dropped too.

This is how Linux does it: the process table entry must stick around (in the "D" dead or "Z" zombie state) until the parent process reaps the child (and learns the exit code). The kernel retains responsibility for remembering the exit code until the parent asks about it.

Let's see, that would need a mapping from `vatID` to a separate per-adminNode facet (so `makeAdminNode` becomes a defineDurableKindMulti) that's used to save the exit status. Then `done` can reliably report vat-terminate-happy by checking `state.exited` instead of `assertRunningVat`. The vatID-to-reportExitStatusNode mapping would be annoying, though, it'd be a durable Map that needs to be populated during `newVatCallback` and deleted during `vatTerminated`. Maybe that wouldn't be too bad.

warner (changes requested):
I haven't gotten to look at the test cases yet, but wanted to at least commit what I've written so far.

The `resume` question is the biggest one, I don't know where/how it should be called, but I know it needs to be called eventually. Maybe the `vatUpgrade` call should include a vref of an object to which `resume` should be sent if the upgrade fails? Or in either case (succeed *or* failure)? That would allow services to do a durable pause, although maybe that's not really useful.

Take a look at `vivifySingleton`, I need to use it myself in vat-timer, I think it would remove some boilerplate.

packages/SwingSet/src/controller/controller.js
FUDCo: Exposing an `upgradeStaticVat` method on the vat admin root (to be implemented in terms of the existing `upgradeVat` function that would be made internal) would enable expressing the `upgradeStaticVat` method in terms of the vat name rather than the vat ID, which I think would be an improvement as it would move the name lookup into the vat admin vat.  It also strikes me that the `await` for the optional pause could be more elegantly handled that way because the `await` could go before the invocation of `upgradeVat` rather than having to be awkwardly embedded inside it.

packages/SwingSet/src/kernel/kernel.js
FUDCo: Oops.  That was just a little bit of debug hackery that I missed removing (I was finding myself editing the verbose log after every run to put a line break between cranks because it was easier to read that way).  But making that format change official has much to recommend it, so if you'd prefer an extra `kdebug` call over an explicit newline char I'm OK with that.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: I added a descriptive comment, and also made things a bit clearer by grouping all the pause/resume/quiescence code together in one place.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: I changed the name to `insistVatAdminServiceNotPaused`.  Suspenders _and_ belt.

I'm not sure that `throw Error(...)` goes very far in making the case that it's not an error.
I did angst a bit about using `assert` here when I wrote it, but our coding convention lately has been to eschew `throw` in favor of `assert`, so I bit the bullet and did that.  But it does seems like we should have a better convention for "abort processing due to exceptional circumstances that are not really erroneous", since such exceptional cases do seem to come up from time to time.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: Perhaps in a different version of JavaScript.  `forEach` is a method on `Array` but set store `.keys()` returns an iterator (and you don't want it to return an array because it might be a store because it contains a jillion elements).

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: I've moved the placement of `finishMeter` to a spot right where it's used, but arranging the rest of it the way you suggest is not entirely possible due to the need to define the kind before any use is made of it.  In particular, `meterByID` contains meters, so the iteration of `meterByID` after it is defined (to reload the ephemeral state after a restart) MUST execute after the definitions of the meter kinds else the iteration will blow up with an "unknown kind" error.  However, all the meter stuff is now at least clustered together in the file.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: > maybe we should store the exit status in the admin facet `state`

That's an attractive notion, but there's no graceful place to stand to do that consistently -- the exit status is signaled by resolving or rejecting the `done` promise, but that vatAdmin vat can't hold onto that across upgrades.  Upon upgrade the `done` promise is severed and a new one created, so the actual act of resolution/rejection is handled external to admin node of the vat that whose `done` promise it is.  In principle we could make the admin node object multifaceted, with one facet that gets exported to the client and one that is retained internally (by storing it in the `runningVats` table, say), and then have a method on the internal facet to inject a done state into it.  But (1) this is a lot of gymnastics to code for a relatively minor value add, and (2) it still doesn't free any external party who is really interested in watching for termination from the obligation to look for the severed `done` promise and re-request a new one when that happens, in which case they already have everything they need to know what's going on.

warner: Hm, I guess those interested parties would react to the first `done()` promise getting upgrade-disconnected by going back to the admin facet and requesting a new `done()` promise, and if the vat had terminated-happy, that second promise could be fulfilled as expected (thanks to the stashed `state.exited`).

The original `done` promise is lost, of course, but a new one can still be created and fulfilled correctly.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -138,11 +141,19 @@ export function buildRootObject(vatPowers, _vatParameters, baggage) {
...
 
-  const pendingBundles = new Map(); // bundleID -> Promise<BundleCap>
-  const pendingUpgrades = new Map(); // upgradeID -> Promise<UpgradeResults>
+  // This promise is set when service is paused (in which state no service
+  // requests should be accepted unless they can be processed in the same turn)
+  // and cleared when service is resumed.  It is resolved when the vat quiesces,
+  // at which point all pending non-prompt service requests have completed.
+  let quiescentP;
+  let quiescentRR;
+  let amQuiescent = false;
```
warner: Hm, why does this start out as `false`? If vat-vat-admin starts up, nobody asks it to do anything, and then someone comes along and pauses it in preparation for upgrade, the "are you quiescent now" question should be answered `true` right away, right?

warner: Oh, I see, it's not really a "this is my current state" flag: it's not polled by `pauseService`. It only gets used by `resumeService`.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -183,27 +179,32 @@ export function buildRootObject(vatPowers, _vatParameters, baggage) {
 
   function resumeService() {
     if (quiescentP) {
-      quiescentRR.reject(Error('shutdown cancelled'));
+      if (!amQuiescent) {
```
warner: I'm trying to think through the possibilities, and was thinking (wrongly, as it turns out) that maybe `quiescentP` should be resolved whether `amQuiescent` is true or false. See if you agree with my logic:

When the vat starts for the first time, `amQuiescent = false` and `quiescentP = undefined`, and the various `pendingBundles`/etc `.size` are zero.

Then work happens (or doesn't), creating pending things and maybe even resolving them.

Then an upgrade request shows up, and we start with `pauseService()`, which sees 1: there's no `quiescentP` (so it creates one, which doubles as a signal to the API handlers to refuse new long-running requests), 2: `checkQuiescence()` is called (and doesn't return immedlately, because `quiescentP` exists) to check the pendings. Now we have two possible cases.

* A: the vat was idle, so `amQuiescent = true` and the promise is fulfilled
* B: the vat was doing something, so `amQuiescent` is left as `false` and the promise is left unresolved

In case B, we can imagine someone changing their mind about the upgrade before actually doing `upgradeVat` (there's no support for that, but it's not implausible, and we might do it in some emergency scenario). If they call `resumeService` (not exactly as it's meant to be called, but if you invoke "`pause`" then invoking "`resume`" is the responsible thing to do), that will see `quiescentP` is set and `amQuiescent` is still `false`, so it will reject the promise and clear all the flags. The vat should now be open for business again.

If they don't change their mind, then we'll exit case B if/when the last pending long-running action finishes. Each place that can remove an entry from one of the four `pending..` sets will also perform a `checkQuiescence`. Eventually, the last one is removed, `amQuiescent = true`, and the promise is fulfilled. Now we're in the same world as case A.

Now `upgradeVat` happens, which sends the `D(vatAdminDev).upgradeVat()` into the kernel, wires up `pendingUpgrades` to watch for the result, and returns a Promise. Liveslots finishes the delivery. Other deliveries into `vat-vat-admin` might occur before the kernel processes the upgrade request, but `quiescentP` is still set, so we'll refuse/reject any new long-running work.

Then the kernel does `processUpgradeVat`, during which the old version gets a `dispatch.stopVat`, and if that works, the new version is loaded and gets a `dispatch.startVat` (which does `buildRootObject` in the new version). If either fails, the whole event is thrown out and the vat worker is reset back to before the `stopVat`. Then the kernel pushes the `vatUpgradeCallback` message onto the queue for delivery into vat-vat-admin (which, again, might happen well after other messages already in the queue).

Those messages (and the `vatUpgradeCallback`) might find the vat in one of two states:

* C: the upgrade succeeded, so everything is new: `amQuiescent = false` again, `quiescentP = undefined` again
* D: the upgrade failed, the vat was rolled back as if it never happened, `amQuiescent = true` (as it was set by `checkQuiescence()`  just before the upgrade), `quiescentP` is a fulfilled promise (also because of `checkQuiescence`)

The non-callback messages that arrive at state C will all be processed normally, even long-running ones (because `quiescentP = undefined`, so the new incarnation is "born ready"). When the `vatUpgradeCallback` arrives, the new incarnation won't recognize the `upgradeID`, it'll whine about it in the logs, then it will return: it won't call `checkQuiescence`, and there is no `pendingUpgrades` promise to resolve, all is good. There was a result promise decided by the original `upgradeStaticVat` call, but it was rejected (disconnected) by liveslots as part of `stopVat`. The new incarnation will receive a `dispatch.notify` about that, which it should ignore because the VPID isn't present in the (now empty) liveslots tables.

If those messages arrive at state D, they'll find `quiescentP` is a (resolved) promise, so long-running ops will be rejected (the vat is still paused). When the `vatUpgradeCallback` arrives (to indicate the error), the `pendingUpgrades` promise will be rejected. That causes the `await upgradeVat` block to throw, which is caught, then it sees `pauseTarget` is true, which sends (itself) a `resumeService`, and since that's a local object, `resumeService` gets to run within the same crank (no round-trip through the kernel). `resumeService` sees that `quiescentP` is set, so it doesn't return immediately. It sees `amQuiescent = true`, so it doesn't reject the promise (which is good, because the promise was already in a fulfilled state). Then it clears `amQuiescent` and `quiescentP`, so now the vat is unpaused again, and all operations will be accepted.

Ok, so I think that convinces me the flag will be set to the right state. I was worried that `quiescentP` would be left unresolved, or that some error pathway could leave the vat permanently paused (`quiescentP` set, but also `amQuiescent = false`, so `resumeService` causes a duplicate `reject` call, which means it throws an error before it can clear `quiescentP`). The only way out of that state would be a self-upgrade (which hopefully would work because `quiescentP` is in fact fulfilled).

I'm a little nervous of the way `amQuiescent` is being used.. but the different approaches that occur to me are wrong (I want to ask for the `if (!amQuiescent)` clause to be just `if (quiescentP)`, but that would trigger the double resolution I'm worried about). So I guess it's the way it needs to be.

Let see, if we're upgrading some *other* static vat (and use the `pauseTarget` scheme), then we'll send a `pauseService` first, but we'll only send a `resumeService` if the kernel tells us (via `vatUpgradeCallback`) that the upgrade failed. So we require that all other pause/resume-capable static vats are also "born ready". I guess if we ever change our mind on that, we'll start by upgrading vat-vat-admin to a version that sends `resumeService` in more of a `finally` block.

The general weirdness of this reflexive brain surgery probably warrants a comment or two. In `upgradeStaticVat` let's note that during self-upgrade, failure routes to the `catch`, but the code will never observe a success because that means it's replaced entirely. Call it anthropically inevitable failure. Mention that the caller's result promise will be rejected in both cases (disconnected in case of success, rejected after `resumeService` unpauses everything in case of failure).

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -634,3 +656,52 @@ test('failed upgrade - unknown options', async t => {
...
+    const status = c.kpStatus(kpid);
+    const capdata = c.kpResolution(kpid);
+    return [status, capdata];
+  };
+
+  const badVABundle = await bundleSource(
+    new URL('./vat-junk.js', import.meta.url).pathname,
+  );
+  const bundleID = await c.validateAndInstallBundle(badVABundle);
+  const kpid = c.upgradeStaticVat('vatAdmin', false, bundleID, {});
```
warner: Let's also use this test to confirm that a failed upgrade (with the same `pause` enabled as we expect to use this in real life) doesn't somehow brick the vat and prevent us from making a successful attempt later. I'm thinking:

* change that `false` to `true`, to trigger pausing
* after the `buildV1` succeeds below:
  * do a `restartVatAdminVat(c)`
  * do another does-it-still-work operation (maybe `buildV2`)

packages/SwingSet/test/vat-admin/terminate/bootstrap-terminate.js
```diff
@@ -1,38 +1,46 @@
...
+  let vatAdminService;
+  let staticVats;
+  let criticalVatKey;
+  let root;
+  let adminNode;
 
   const self = Far('root', {
-    async bootstrap(vats, devices) {
+    bootstrap(vats, devices) {
+      vatAdminService = E(vats.vatAdmin).createVatAdminService(
```
warner, line 14: Huh, interseting, I didn't realize the original test wasn't `await`ing, so it holds onto `vatAdminServiceP` rather than the actual presence. Eh, it's different than our other tests, but probably good to have some diversity.

packages/SwingSet/test/vat-admin/terminate/test-terminate.js
```diff
@@ -118,39 +157,55 @@ test.serial('terminate (dynamic, non-critical)', async t => {
     'v8',
     'kill',
     true,
-    'done exception kill (Error=false)',
+    [
+      'termP.catch Error: vatAdminService rejecting attempt to perform "terminateWithFailure"() on non-running vat "v8"',
```
warner, line 161: if we put these two long strings into short-named local variables, would `prettier` let us fit the `doTerminateNonCritical` call back onto a single line?

packages/SwingSet/test/vat-admin/test-create-vat.js
```diff
@@ -85,46 +85,81 @@ async function doTestSetup(t, enableSlog = false) {
...
-test('createVatByBundle', async t => {
-  const { c, vat13Bundle } = await doTestSetup(t);
+async function testCreateVatByBundle(t, doVatAdminRestart) {
+  const { c, vat13Bundle } = await doTestSetup(t, doVatAdminRestart);
   const kpid = c.queueToVatRoot('bootstrap', 'byBundle', [vat13Bundle]);
   await c.run();
   t.deepEqual(JSON.parse(c.kpResolution(kpid).body), 13);
+}
+
+test('createVatByBundle', async t => {
```
warner, line 101: with 8 new test cases in this  file, which all launch kernels and a batch of vat workers, I think we might need the same `test.serial` as we use in test-terminate

warner (approved):
Looks good.. I was on the fence about how `vivifySingleton` would turn out, but I like it.

A few small test improvements to make, and tell me if you agree with my analysis about `quiescentP` states, then I think it's ready to land. Nice job!

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: Yes.  Arguably this could have a better name.  It really means "have reached the quiescent state after being asked to pause" or more precisely "`quiescentP` has been resolved", but neither of those are really things that lend themselves to a tidy name.  I put it in specifically because you can't synchronously poll a promise to see if it's been resolved, but since I was the one resolving it I could take note that I had when I did.  If you have an idea for a better name I'm all ears.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
FUDCo: That's a very convoluted analysis to come to the same conclusion that I did.  I can't figure out if that means you are more thorough or I was more insightful :-).  But anyway I will add some commentary to make what's going on clearer.

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: Oh, actually putting `false` in the call to `upgradeStaticVat` there was just a straight up mistake on my part.  It should be `true`.  The reason it worked is because there was no other activity going on in the vatAdmin vat at the time (on account of this test being minimalistic as tests generally should be) so we benefitted from a de facto paused state even though we hadn't asked for it.  As a rule it's an error to upgrade a vatAdmin vat without pausing it first; it's just that there's no place to stand to detect this condition, since user code never sees the `stopVat` request -- it just gets shot in the head.  In a more ideal world the `shouldPauseFirst` flag should not be a discretionary parameter for whoever calls `upgradeStaticVat`; rather, the vatAdmin's `upgradeStaticVat` method should be able to find out if the target vat supports `pauseService` and send the message if it does.  Hmm.  I wonder if we should _always_ send `pauseService` and just catch and ignore the "no such method" error if one is thrown.  Can we even do that?  That seems a little sketchy to me, because it makes the method name magic.

I think the `buildV1` after the failed upgrade is sufficient, because we already have lots of tests that restarting vatAdmin between various operations leaves it in a correct state.

packages/SwingSet/test/vat-admin/test-create-vat.js
FUDCo: I would say let's not assume there's a problem until there's a problem.  But if we see this exhibiting test flakage in CI we'll immediately know what the first move to make is.

packages/SwingSet/test/vat-admin/terminate/test-terminate.js
FUDCo: Meh.  For now I'm happy just to tell `prettier` to shut up.  `prettier` is wrong so often I've become very blasé about that.

Issue #4286, opened 2022-01-12
TimerService device stores repeaters in an ever-growing array


Chris-Hibbert:
## Describe the bug

The [TimerService stores repeaters](https://github.com/Agoric/agoric-sdk/blob/f5995854229921788d94caa97107360130085e61/packages/SwingSet/src/devices/timer-src.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L211) in an array that grows without bound. When repeaters are deleted, they are replaced in the array with null, so that indexing is preserved.

## Vulnerability

Clients can create repeaters at low cost (particularly if they delete them quickly), but this causes the timer to grow the repeaters array to an arbitrary size.

This an issue for Zoe and other core services that rely on timers.

## Expected behavior

Memory use shouldn't grow like that.

## Additional context

Identified during the Zoe Purple team exercise

zarutian:
/me thinks about if and how #3612 would fare in this context. Perhaps it would only be the same issues that Notifiers and Subscriptions have?

I think this is mainly an implementation spefic issue.

warner:
With the closing of #5668, the timer *device* still suffers from this problem, but users no longer get access to it. The new timer *vat* [stores Repeaters as virtual objects](https://github.com/Agoric/agoric-sdk/blob/130feae4b0341bd8ea17569837376481dfa82f1b/packages/SwingSet/src/vats/timer/vat-timer.js#L699-L703) so they no longer consume RAM, only DB, and they should be deleted automatically when unreferenced and unscheduled.

I'll close this ticket now, because the core problem is resolved, and our next step will be to replace the timer device with a stripped down version that has no Repeaters at all, only a single wakeup that vat-timer uses to implement everything else.

Issue #4567, opened 2022-02-16
build durable notifier


warner:
## What is the Problem Being Solved?

As part of #4550 and #4383, we're faced with a challenge. Promises can be referenced by virtual objects (#2485). The `Promise` object remains in RAM, which doesn't meet our memory-consumption goals, but it remains sound to reference them. However they are not "durable", because they cannot be entirely ejected from RAM and then resurrected on-demand later. So durable data cannot reference virtual data, and only virtual data can reference promises.

We know that we need to virtualize Notifiers (#4513) to meet our RAM goals. We know that Zoe and contracts need to hold their long-term state in durable data, and any externally-visible objects must be Durable (e.g. contract facets must survive upgrade). So the question is how clients should get reconnected to their notifiers after an upgrade.

One option is to make durable notifiers. Then the notifier can be part of the `state` of the durable facet/etc. However each notifier needs to hold onto a promise (the one it will resolve upon the next update), and promises cannot be durable. If we can create virtual promises (#3787) then we aren't violating our RAM budget, but we still can't reference the merely-virtual promise from the durable notifer's `state`.

Another is to leave notifiers as merely virtual. But the durable facet cannot reference such an object from its `state`.

A trick I've been thinking about that might be useful in either approach would be to have the durable facet Kind constructor close over a merely-virtual Notifier constructor. They could share an integer index, and use it to reference a virtual table of notifiers that were created within this version of the vat. These notifiers would be lost upon upgrade, but regenerated when clients notice their `getUpdateSince()` promises rejecting and then ask for a new notifier.

## Description of the Design

Use an ephemeral table of promises, which get rejected upon upgrade: see below.

warner:
That trick looks like the way to go:

```js
// create these handles only in version-1, otherwise get from baggage
const counterHandle = makeKindHandle('counter');
const notifierHandle = makeKindHandle('notifier');

// in all versions:

//  define a durable counter
const makeCounter = defineDurableKind(counterHandle, () => { c: 0 }, {
  next: ({state}) => { state.c += state.c; return state.c; },
});

const counter = makeCounter();

// define an ephemeral table, indexed by the counter:
const ephemeralPromises = new Map();
function getEphemeralPromiseKit(index) {
  if (!ephemeralPromises.has(index)) {
    ephemeralPromises.set(index, makePromiseKit());
  }
  return ephemeralPromises.get(index);
}

// each new Notifier needs a distinct index
function initNotifierState() {
[...]

PR #5847, opened 2022-07-28
overhaul vat-timer: virtualized, durable, upgradable, Clock, TimerBrand

### LINKED ISSUE #4282: Timer .schedule() panics when called with no arguments, halting the chain
## Describe the bug
A node (local-solo) can stop the chain (consensus failure) by issuing the following command in repl:
```js
E(E(home.chainTimerService).makeRepeater(1n,1n)).schedule()
```

This results in no handler being called by the `.wake` function, and panics the timer service and resulting chain


## To Reproduce
3. wait for chain to boot
4. Issue `E(E(home.chainTimerService).makeRepeater(1n,1n)).schedule()`
5. The chain will stop due to the following stack
```
2022-01-12T03:24:15.994Z block-manager: BEGIN_BLOCK error: (Error#1)                                                                                         
Error#1: error in devicePollFunction: (Error#2)                                                                                                              
                                                                                                                                                             
  at Object.poll (packages/SwingSet/src/devices/timer.js:28:14)                                                                                              
  at beginBlock (packages/cosmic-swingset/src/launch-chain.js:291:32)                                                                                        
  at kernelPerformAction (packages/cosmic-swingset/src/block-manager.js:48:13)                                                                               
  at blockManager (packages/cosmic-swingset/src/block-manager.js:225:19)                                                                                     
  at toSwingSet (packages/cosmic-swingset/src/chain-main.js:307:12)                                                                                          
  at portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:121:22)                                                                             
  at Object.fromGo (packages/cosmic-swingset/src/chain-main.js:138:31)                                                                                       
                                                                                                                                                             
Nested error under Error#1                                                                                                                                   
  Error#2: SO(x) must be called on a Presence, not undefined                                                                                                 
   
    at SO (.../swingset-vat/src/kernel/deviceSlots.js:136:5)
    at eval (.../swingset-vat/src/devices/timer-src.js:183:11)
    at poll (.../swingset-vat/src/devices/timer-src.js:178:19)
    at poll (.../swingset-vat/src/devices/timer-src.js:256:12)
    at Object.poll (packages/SwingSet/src/devices/timer.js:26:22)
    at beginBlock (packages/cosmic-swingset/src/launch-chain.js:291:32)
    at kernelPerformAction (packages/cosmic-swingset/src/block-manager.js:48:13)
    at blockManager (packages/cosmic-swingset/src/block-manager.js:225:19)
    at toSwingSet (packages/cosmic-swingset/src/chain-main.js:307:12)
    at portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:121:22)
    at Object.fromGo (packages/cosmic-swingset/src/chain-main.js:138:31)
   
portHandler threw (Error#1)
7:24PM ERR CONSENSUS FAILURE!!! err="Error: error in devicePollFunction: (an Error)" module=consensus stack="goroutine 74 [running]:\nruntime/debug.Stack()\n
```

## Expected behavior
A malformed timer schedule should not schedule the undefined handler, and error gracefully

## Additional notes
This behavior is also valid for localTimerService

```
2022-01-12T03:30:29.666Z start: timer-provoked kernel crank failed at 1641958229665: (Error#1)
Error#1: error in devicePollFunction: (Error#2)

  at Object.poll (packages/SwingSet/src/devices/timer.js:28:14)
  at moveTimeForward (packages/solo/src/start.js:246:19)
  at thunk (packages/vats/src/queue.js:34:26)

Nested error under Error#1
  Error#2: SO(x) must be called on a Presence, not undefined
  
    at SO (.../swingset-vat/src/kernel/deviceSlots.js:136:5)
    at eval (.../swingset-vat/src/devices/timer-src.js:183:11)
    at poll (.../swingset-vat/src/devices/timer-src.js:178:19)
    at poll (.../swingset-vat/src/devices/timer-src.js:256:12)
    at Object.poll (packages/SwingSet/src/devices/timer.js:26:22)
    at moveTimeForward (packages/solo/src/start.js:246:19)
    at thunk (packages/vats/src/queue.js:34:26)
  
```

## Platform Environment
 macOS 12.1, node v16.13.1
 I only have 16gb of ram

ref #4264

### LINKED ISSUE #4296: Kernel Panic: `E(agoric.chainTimerService).removeWakeup(0)`
## Describe the bug
Calling `E(agoric.chainTimerService).removeWakeup(0)` results in immediate kernel panic

## To Reproduce
3. wait for chain to boot
4. Issue `E(agoric.chainTimerService).removeWakeup(0)`
5. The chain will stop due to the following stack

```
2022-01-14T00:19:14.461Z SwingSet: kernel: ##### KERNEL PANIC: error during syscall/device.invoke: Error: Cannot pass non-frozen objects like (an object). Use harden() #####
Logging sent error stack (Error#1)
Error#1: syscall.callNow failed, prepare to die: you killed my kernel. prepare to die
Error: syscall.callNow failed, prepare to die: you killed my kernel. prepare to die
 at apply ()
 at Error (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:7001)
 at doSyscall (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/supervisor-subprocess-xsnap.js:9562)
 at (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/supervisor-subprocess-xsnap.js:8913)
 at apply ()
 at apply ()
 at dispatchToHandler (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:10195)
 at (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:9834)
 at win (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:10534)
 at ()

Error#1 ERROR_NOTE: Thrown from: (Error#2) : 65 . 0
Error#1 ERROR_NOTE: Sent as error:liveSlots:v16#70001
Error#2: Event: 64.1
Error: Event: 64.1
 at construct ()
 at Error (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:7003)
 at trackTurns (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:9818)
 at handle (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/lockdown-subprocess-xsnap.js:10525)
 at deliver (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/supervisor-subprocess-xsnap.js:8979)
 at dispatchToUserspace (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/supervisor-subprocess-xsnap.js:9296)
 at runWithoutMetering (/Users/b4d2/agoric-sdk/packages/SwingSet/src/kernel/vatManager/supervisor-subprocess-xsnap.js:9683)
 at ()

2022-01-14T00:19:14.466Z block-manager: END_BLOCK error: (Error#1)
Error#1: Cannot pass non-frozen objects like [] . Use harden()

  at passStyleOfInternal (.../marshal/src/passStyleOf.js:131:13)
  at passStyleOfRecur (.../marshal/src/passStyleOf.js:102:25)
  at passStyleOf (.../marshal/src/passStyleOf.js:180:12)
  at encode (.../marshal/src/marshal.js:159:25)
  at Object.serialize (.../marshal/src/marshal.js:266:21)
  at Object.invoke (.../swingset-vat/src/kernel/deviceSlots.js:209:34)
  at Object.invoke (.../swingset-vat/src/kernel/deviceManager.js:80:36)
  at invoke (.../swingset-vat/src/kernel/kernelSyscall.js:228:28)
  at Object.doKernelSyscall (.../swingset-vat/src/kernel/kernelSyscall.js:291:16)
  at vatSyscallHandler (.../swingset-vat/src/kernel/kernel.js:911:39)
  at Object.syscallFromWorker (.../swingset-vat/src/kernel/vatManager/manager-helper.js:251:18)
  at handleUpstream (.../swingset-vat/src/kernel/vatManager/manager-subprocess-xsnap.js:88:23)
  at handleCommand (.../swingset-vat/src/kernel/vatManager/manager-subprocess-xsnap.js:110:22)
  at runToIdle (packages/xsnap/src/xsnap.js:197:42)

portHandler threw (Error#1)
4:19PM ERR CONSENSUS FAILURE!!!
```
## Expected behavior
Calling removeWakup with invalid arguments should not result in a kernel panic

## Platform Environment
 macOS 12.1, node v16.13.1
 I only have 16gb of ram

ref #4264

### LINKED ISSUE #5616: consider adding getCurrentTimestamp() -only timer device node

@michaelfig said that the smart wallet would benefit from a way to synchronously retrieve the current timestamp.

Currently, the only way to get the current timestamp synchronously is through `D(devices.timer).getLastPolled()`. This is exposed *asynchronously* through `E(timerService).getCurrentTimestamp()` (hosted in vat-timer).

Both `devices.timer` and `timerService` have a lot of additional power. `timerService` has `setWakeup` and `makeRepeater`. And the timer *device* has a slew of non-ocap-safe APIs (to overcome limitations of the swingset "deviceslots" library), like deleting repeaters by their integer index.

So it might be interesting to create an object and/or a device node with a narrower API, just `getCurrentTimestamp()`. If hosted in vat-timer, the object would provide async access to the current time, but not any wakeup services. If hosted in device-timer, the device node would provide synchronous access to the current time, and nothing else.

### LINKED ISSUE #5668: make vat-timer upgradable

To satisfy #5666 and make vat-timer upgradable, it needs to be rewritten in terms of durable Kinds.

## Description of the Design

The timer vat exposes four kinds of objects (search for `Far()` to find their definitions):

* root: this doesn't need changes, and the only method it exposes is `createTimerService`
* timer service: this needs to become a singleton durable Kind, with methods `getCurrentTimestamp`, `setWakeup`, `removeWakeup`, `makeRepeater` (unless #5555 allows us to remove it, which would be great), `makeNotifier`, and `delay`
  * the timer service receives the TimerDevice in `createTimerService()`, and it must either stash it in the singleton's `state`, or in `baggage`
* a `vatRepeater` will be an instance of a durable Kind, with `schedule` and `disable` methods
* a `delayHandler` will be the same, with a single `wake` method. This object is used internally.

Note that vat-timer communicates with device-timer, mostly using integer indices rather than object references. We shouldn't need to modify the device to complete this task, however we might consider rewriting it as well, especially if we found a good way to address problems like #4286.

## Security Considerations

Now that I'm looking at vat-timer after several years (!), I notice that the authority to do a `removeWakeup(handler)` comes from holding the same handler object that was originally passed into `setWakeup(baseTime, handler)`. This isn't the best pattern: you can imagine somebody incorporating the handler methods (just `wake()`) into some larger object, then sharing that object with other parties. This would inadvertently give those parties the authority to cancel the timer.

A better pattern would be for `setWakeup` to create and return a "remover" object, and for the caller to hold onto that object if they wanted to ever call `E(remover).remove()` in the future. That would require client changes, though, and I'm not sure how widespread `setWakeup` currently is within our codebase (and external contracts/dapps).

I'm also reminded of the following warts:
* the root object (available to bootstrap as `vats.timer`) is not the actual timer service
* userspace `bootstrap` is obligated to wire `vats.timer` and `devices.timer` together (see #4523 or #4492)
* `E(vats.timer).createTimerService()` could be called multiple times, with (maybe?) bad results

I don't know that the upgradable/durable changes would affect/improve any of these, but it'd be nice.

## Test Plan

As with #5667, either unit tests of actual upgrade, or a careful audit for lingering non-durable state.

### LINKED ISSUE #5709: remove makeNotifierFromAsyncIterable from vat-timer

It's a "hot notifier" in the updater is driving the changes instead of the client. In a _cold_ notifier the client asks for the next update and the notifier figures out what to do. Is doesn't any activity at all until the client calls next() again.

## Description of the Design

Most of the uses of `makeNotifierFromAsyncIterable` take subscriptions so they'll be converted in https://github.com/Agoric/agoric-sdk/pull/5695 to use the new function for those cases.

These two don't:

https://github.com/Agoric/agoric-sdk/blob/cafa284c7d13d6ad520e34f11c9e53e0aa55f9f7/packages/SwingSet/src/vats/timer/vat-timer.js#L66-L73

https://github.com/Agoric/agoric-sdk/blob/cafa284c7d13d6ad520e34f11c9e53e0aa55f9f7/packages/zoe/tools/fakePriceAuthority.js#L233-L256

@michaelfig suspects the latter may solve a leak so tagging this with 'performance'.

## Security Considerations
--
## Test Plan

--

warner:
vat-timer is now fully virtualized, durablized, and upgradeable. RAM
usage should be O(N) in the number of:

* pending Promise wakeups (`wakeAt`, `delay`)
* active Notifier promises (`makeNotifier`)
* active Iterator promises (`makeNotifier()[Symbol.asyncIterator]`)

Pending promises will be disconnected (rejected) during upgrade, as
usual.

All handlers and Promises will fire with the most recent timestamp
available, which (under load) may be somewhat later than the scheduled
wakeup time.

Until cancellation, Notifiers will always report a scheduled time
(i.e. `start` plus some multiple of the interval). The opaque
`updateCount` used in Notifier updates is a counter starting from 1n.
When a Notifier is cancelled, the final/"finish" value is the
timestamp of cancellation, which may or may not be a multiple of the
interval (and might be a duplicate of the last non-final value). Once
in the cancelled state, `getUpdateSince(anything)` yields `{ value:
cancellationTimestamp, updateCount: undefined }`, and the
corresponding `iterator.next()` resolves to `{ value:
cancellationTimestamp, done: true }`. Neither will ever reject their
Promises (except due to upgrade).

Asking for a wakeup in the past or present will fire immediately.

Most API calls will accept an arbitrary Far object as a CancelToken,
which can be used to cancel the wakeup/repeater. `makeRepeater` is the
exception.

This does not change the device-timer API or implementation, however
vat-timer now only uses a single device-side wakeup, and only exposes
a single handler object, to minimize the memory usage and object
retention by the device (since devices do not participate in GC).

This introduces a `Clock` which can return time values without also
providing scheduling authority, and a `TimerBrand` which can validate
time values without providing clock or scheduling
authority. Timestamps are not yet Branded, but the scaffolding is in
place.

`packages/SwingSet/tools/manual-timer.js` offers a manually-driven
timer service, which can help with unit tests.

closes #4282
refs #4286
closes #4296
closes #5616
closes #5668
closes #5709
refs #5798


packages/SwingSet/docs/timer.md
```diff
@@ -67,42 +67,82 @@ A single application might have multiple sources of time, which would require th
 The `timerService` object can be distributed to other vats as necessary.
 
 ```js
-  // for this example, assume poll() provides seconds-since-epoch as a BigInt
+  // for this example, assume poll() provides seconds-since-epoch
```
erights, line 70: The triple backshash on line 12 above should be triple backshash-`js`, like this and others.

packages/SwingSet/docs/timer.md
```diff
@@ -67,42 +67,82 @@ A single application might have multiple sources of time, which would require th
...
+  await p1;
 
+  // same, but cancellable
+  const cancel2 = Far('cancel', {}); // any pass-by-reference object
+  // the cancelToken is always optional
+  const p2 = E(timerService).delay(30n, cancel2);
+  // E(timerService).cancel(cancel2) will cancel that
+
+  // same, but absolute instead of relative-to-now
+  const monday = 1660000000;
```
erights:
```suggestion
  const monday = 1_660_000_000;
```

packages/SwingSet/docs/timer.md
```diff
@@ -67,42 +67,82 @@ A single application might have multiple sources of time, which would require th
...
+  const notifier = E(timerService).makeNotifier(delay, interval, cancel2);
+
+  for await (const scheduled of notifier) {
+    console.log(`woken up, scheduled for ${scheduled}`);
+    // note: runs forever, once per 'interval'
+    break; // unless you escape early
+  }
+
+  // the old `makeRepeater` API creates a "repeater object" with .schedule
+  // and .disable methods to turn it on and off.
```
erights: From the example, it looks like we are both supporting these methods and explaining them. I'd drop talk about "the old ... API" unless we're deprecating them. If we are, we should either say that explicitly here or not bother to explain them at all.

packages/SwingSet/docs/timer.md
```diff
@@ -67,42 +67,82 @@ A single application might have multiple sources of time, which would require th
...
   const interval = 60n;
+  E(timerService).repeatAfter(delay, interval, handler, cancel2);
+
+  // repeating wakeup service, Notifier-style . This supports both
+  // the native 'notifier.getUpdateSince()' Notifier protocol, and
+  // and an asyncIterator
+
+  const notifier = E(timerService).makeNotifier(delay, interval, cancel2);
+
+  for await (const scheduled of notifier) {
```
erights: Bug. This won't work unless `notifier` is a local asyncIterable. From the code on line 121 it will always be a promise. This code should never work. An `await` or similar will generally only turn it into a presence for a remote timer.

This won't work it `notifier` is remote, which it generally will be from the code on line 121.
```suggestion
  const notifierP = E(timerService).makeNotifier(delay, interval, cancel2);
  const notifier = makeNotifier(notifierP);
  
  for await (const scheduled of notifier) {
```

packages/SwingSet/package.json
```diff
@@ -31,6 +31,7 @@
     "@agoric/notifier": "^0.4.0",
     "@agoric/store": "^0.7.2",
     "@agoric/swing-store": "^0.7.0",
+    "@agoric/vat-data": "^0.3.1",
```
erights: This creates a cyclic dependency, right? We're on with package-level cycles but try to avoid module level cycles, yes?

packages/SwingSet/src/vats/timer/types.d.ts
```diff
@@ -88,12 +88,6 @@ declare global {
      * Return value is the time at which the call is scheduled to take place
      */
```
erights, line 95: The TODO on line 12 above will be done by this PR, so this PR should remove it.

Altogether, it looks like this file is mostly stale wrt this PR so I'll hold further review until you tell me otherwise.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
-export function buildRootObject(vatPowers) {
+// RAM usage: O(number of outstanding delay() promises) +
+// O(number of notifiers??). setWakeup() only consumes DB.
+
+/**
+ * // Handler is a Far object with .wake(time) used for callbacks
+ *
+ * @typedef Handler
+ * @property {(scheduled: Timestamp) => unknown} wake
+ * @typedef {unknown} CancelToken
```
erights, line 29:
```suggestion
 *
 * @typedef {unknown} CancelToken
```

Please separate distinct type definitions by at least a newline.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
 import { TimeMath } from './timeMath.js';
 
-export function buildRootObject(vatPowers) {
+// RAM usage: O(number of outstanding delay() promises) +
+// O(number of notifiers??). setWakeup() only consumes DB.
+
+/**
+ * // Handler is a Far object with .wake(time) used for callbacks
+ *
+ * @typedef Handler
```
erights: Whether this is an improvement depends on what you intend. By putting it immediately after the `@typedef Handler`, when some elsewhere hovers over the `Handler` type, they'll see your comment. OTOH, if you intend this comment only for those reading the implementation, no change suggested.

```suggestion
 * @typedef Handler
 * Handler is a Far object with .wake(time) used for callbacks
```

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+ * @property {(scheduled: Timestamp) => unknown} wake
+ * @typedef {unknown} CancelToken
+ * // CancelToken must be pass-by-reference and durable, either local or remote
```
erights: Why both comment markers?

```suggestion
 * CancelToken must be pass-by-reference and durable, either local or remote
```

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+ * @property {() => void} scheduleYourself
+ * @property {() => void} fired
+ * @property {() => void} cancel
+ * @typedef {MapStore<TimestampValue, Event[]>} Schedule
+ * @typedef {WeakMapStore<CancelToken, Event[]>} CancelTable
+ * @typedef {unknown} PromiseEvent
+ * @typedef WakeupPromiseControls
+ * @property {(scheduled: Timestamp) => void} resolve
+ * @property {(err: unknown) => void} reject
+ * @typedef {LegacyWeakMap<PromiseEvent, WakeupPromiseControls>} WakeupPromiseTable
```
erights, line 51: Using a LegacyWeakMap is usually a sign of trouble. In any case, we're trying to phase them out, so I'm a bit alarmed to see a new usage.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+ */
+
+/* Repeaters have an Event with both 'start' and 'interval'. One-shot
+ * wakeups are Events with .interval = undefined. Each Event is
+ * tri-state: "scheduled" (.scheduled is non-undefined), "cancelled"
```
erights: This sounds like you're describing a data record, but above you define `Event` to be an object with methods.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+// these internal functions are exported for unit tests
+
+/**
+ * Insert an event into the schedule at its given time.
+ *
+ * @param {Schedule} schedule
+ * @param {TimestampValue} when
+ * @param {Event} event
+ */
+function addEvent(schedule, when, event) {
```
erights:
```suggestion
const addEvent = (schedule, when, event) => {
```
etc

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+/* Repeaters have an Event with both 'start' and 'interval'. One-shot
+ * wakeups are Events with .interval = undefined. Each Event is
+ * tri-state: "scheduled" (.scheduled is non-undefined), "cancelled"
+ * (.cancelled=true), or "executing" (only for Repeaters, indicated by
+ * .scheduled=undefined and .cancelled = false). Events are kept alive
+ * by the schedule while "scheduled", by handler.wake result promise
+ * callbacks while "executing", and by mostly nothing when "cancelled".
+ */
+
+// these internal functions are exported for unit tests
```
erights: This is fine if they cannot be used to attack (as in some `*Jig` methods I recently introduced). But if they can be used to attack, we need to at least wave a big red flag to note the vulnerability so that we can fix it before MN3.

In this case I believe they cannot be used for attack, because all the state they manipulate is provided by arguments. If so, then no change suggested, except to note why this is safe.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+ * @param {Schedule} schedule
+ * @param {TimestampValue} when
+ * @param {Event} event
+ */
+function removeEvent(schedule, when, event) {
+  if (schedule.has(when)) {
+    /** @typedef { Event[] } */
+    const oldEvents = schedule.get(when);
+    /** @typedef { Event[] } */
+    const newEvents = oldEvents.filter(ev => ev !== event);
```
erights: In this context, terms like `old*` and `new*` seem like they refer to the temporal order this very code is managing. Indeed, I had to read it twice to get past that misreading.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+  // resolve/reject functions for delay/wakeAt. We can't hold these
+  // bare functions in the (durable) PromiseEvent, but we *can* use
+  // the PromiseEvent as a key to fetch them when the event
+  // fires. It's ok for these to be ephemeral: all promises get
+  // rejected (with { name: 'vatUpgraded' }) during an upgrade, so if
+  // the timer fires *after* an upgrade, we no longer need to reject
+  // it ourselves.
+
+  /** @type {WakeupPromiseTable} */
+  const wakeupPromises = makeLegacyWeakMap('promises');
```
erights: Why not
```suggestion
  const wakeupPromises = makeScalarWeakMapStore('promises');
```
?

It seems the reason is that the values are PromiseControls, and PromiseControls are not declared `Far`. I'd rather have them declared `Far` than use a legacyMapStore.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+
+  /**
+   * @param {TimestampValue} when
+   * @param {CancelToken} cancelToken
+   * @returns {Promise<Timestamp>}
+   */
+  function wakeAtInternal(when, cancelToken) {
+    const event = makePromiseEvent(when, cancelToken);
+    const { resolve, reject, promise } = makePromiseKit();
+    const controls = harden({ resolve, reject });
```
erights: See comment on line 241

```suggestion
    const controls = Far('promise control', { resolve, reject });
```

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+
+  /**
+   * convert external (branded) RelativeTime to internal bigint
+   *
+   * @param {RelativeTime} delta
+   * @returns {RelativeTimeValue}
+   */
+  function fromRelativeTime(delta) {
+    return TimeMath.relValue(delta);
+  }
```
erights: Initially I was confused why these helpers are needed, and why they are not at the top level of the file. Then I realized that the intention is for them to capture the `brand` in question. No change requested.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,854 @@
...
+  // instance (one TimerService per vat), and both the TimerDevice and
+  // the supporting collections are closed over by the singleton.
+
+  function initService() {
+    return {};
+  }
+
+  function noContext(f) {
+    return (context, ...args) => f(...args);
+  }
```
erights: Import `dropContext` from `@agoric/vat-data` instead. It does exactly the same thing.

erights (commented):
Just some comments so far. #5862 is also a review comment.

I have not yet dug into the meat of this. Looking forward to doing so.

packages/SwingSet/src/vats/timer/vat-timer.js
Chris-Hibbert: Actually, if you're going to leave it as implementation comment, please drop the extraneous `//` within the `/** ... */` comment

packages/SwingSet/package.json
warner: Oh, yeah, Chip and Richard and I talked about that.

vat-data has a `devDependencies` on swingset, to support its unit tests, but not a `dependencies`. vat-data looks for names on `globalThis`, which are usually there because swingset added it to the vat environment, but vat-data doesn't actually import anything from swingset.

So I think we don't have a real package-level cycle: we have a package-level `devDependency` cycle, which I think is ok.

packages/SwingSet/src/vats/timer/types.d.ts
warner: yeah, I haven't tried to improve `types.d.ts` at all, I just tried to make sure the typechecker wasn't complaining. I'll make a pass through this file and try to remove the stale parts.

packages/SwingSet/src/vats/timer/vat-timer.js
warner: Oh, it looks like I could store the resolve/reject functions in a Far (even though they're never revealed off-vat), and that would let us use a `WeakMapStore` instead of the `LegacyWeakMap`. I'll do that.

The reason we can'ut use a (virtual) `BigWeakMapStore` is that we don't have virtual/durable Promises. The #4567 "durable notifier" plan will need this pattern too, to hold both the Notifier's `Promise` and it's resolve/reject control functions.

packages/SwingSet/src/vats/timer/vat-timer.js
warner: I'll add the "this is why this is safe" comment. They're pure functions. Also, they're exported under a separate name, so when this file is used as a vat entry point, we can leverage the fact that liveslots and the `subprocess-xsnap` supervisor don't touch anything other than `buildRootObject`.

For the benefit of the `manualTimer` unit-test support tool (used by zoe's tests, and others), I might need to add a non-pure portion which lets the unit test driver get a callback with every `wake()` result promise, so it can `await` them as it sees fit. I might also be able to get away without that.. still trying to look for solutions. If I do add that, we'll be relying upon the "only `buildRootObject`" aspect for security.

packages/SwingSet/src/vats/timer/vat-timer.js
warner: hm, it's "before removing" and "after removing".. maybe I'll try a `originalEvents` and `remainingEvents`, if it doesn't make the line too long

packages/SwingSet/package.json
erights: I'd forgotten about the global thing. Weird way to depend de facto but not de jure ;). devDependencies are indeed ok, so all is good here. Resolving.

packages/SwingSet/src/vats/timer/vat-timer.js
erights: I see it's done. Resolving.

packages/SwingSet/src/vats/timer/vat-timer.js
erights: I see it's done. Resolving.

packages/SwingSet/src/vats/timer/vat-timer.js
erights: I see it's done. Resolving.

packages/SwingSet/src/vats/timer/vat-timer.js
erights: Gone. Resolving.

packages/SwingSet/tools/manual-timer.js
```diff
@@ -0,0 +1,123 @@
+/* global setImmediate */
```
erights: Even though it is not in a `test` directory, since the purpose of `manual-timer` is only to support testing, I gave it a much lighter review as I generally do for test code. Just for the record. No change suggested.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,830 @@
...
+  // then the two durable tables
+
+  // The 'schedule' maps upcoming timestamps to the Event that should
+  // be fired. We rely upon the earlier-vs-later sortability of BigInt
+  // keys, and upon our Stores performing efficient iteration.
+
+  /** @type {Schedule} */
+  const schedule = provide(baggage, 'schedule', () =>
+    makeScalarBigMapStore('schedule', { durable: true }),
+  );
```
erights:
```suggestion
  const schedule = provideDurableMapStore(baggage, 'schedule');
```

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,830 @@
...
+
+  // 'cancels' maps cancel handles to Events that will be
+  // removed. Each event knows its own .scheduled time, so we can find
+  // and remove it from 'schedule'
+
+  /** @type {CancelTable} */
+  const cancels = provide(baggage, 'cancels', () =>
+    makeScalarBigWeakMapStore('cancels', { durable: true }),
+  );
```
erights:
```suggestion
  const cancels = provideDurableWeakMapStore(baggage, 'cancels');
```

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,830 @@
...
+  // First, create the KindHandles for all our durable objects. These
+  // are ongoing obligations of the vat: all future versions must
+  // define behaviors for these. We are also obligated to support the
+  // vivifySingleton calls at the bottom: wakeupHandler, timerService,
+  // timerClock, and timerBrand.
+  const oneShotEventHandle = provideKindHandle(baggage, 'oneShotEventHandle');
+  const promiseEventHandle = provideKindHandle(baggage, 'promiseEventHandle');
+  const repeaterEventHandle = provideKindHandle(baggage, 'repeaterEventHandle');
+  const repeaterHandle = provideKindHandle(baggage, 'repeaterHandle');
+  const notifierHandle = provideKindHandle(baggage, 'notifierHandle');
```
erights: For every case where we have a call to `provideKindHandle` and a corresponding call to `defineDurableKind` or `defineDurableKindMulti`, I think the code would be clearer to replace these with a single call to `vivifyKind` or `vivifyKindMulti`. Typically, programs will have no other uses of kind handles, e.g., no direct calls to `makeKindHandle`. By making this change, those programs hide the entire concept of a kindHandle, making the durability API easier to learn, and the code using it easier to understand.

For me this was one of the punchlines of seeing our own experimental use of the durability/upgrade API change over time. We gradually climbed abstractions until we got to `provide*` and `vivify*`, therefore initially thinking of them as helpers for saying the other things more compactly. But we now hope most upgradable code can rely exclusively on `provide*` and `vivify*`, typically hiding the need to think about the concepts they encapsulate.

erights (commented):
Just more comments so far. Not done reviewing yet.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,830 @@
...
+   *
+   * @param {RelativeTime} delta
+   * @returns {RelativeTimeValue}
+   */
+  function fromRelativeTime(delta) {
+    return TimeMath.relValue(delta);
+  }
+
+  function reschedule() {
+    assert(wakeupHandler, 'reschedule() without wakeupHandler');
```
erights:
```suggestion
```

This `assert` is no longer meaningful, now that `wakeupHandler` is a const.

erights (approved):
LGTM!

There are still things I'd like to understand better. But we can discuss them after this is merged (and once I get back on Wednesday).

packages/SwingSet/src/vats/timer/vat-timer.js
warner: I wasn't sure what the type parser was sensitive to.. I'll remove the `//`

packages/SwingSet/src/vats/timer/vat-timer.js
warner: I will pretend that I was thinking that far ahead :)

warner:
Questions for reviewers:
* Somebody said that the new `timerBrand` did not need a `isMyClock()` method, but now I'm not so sure. If you hold a branded Timestamp and someone hands you a Clock, would you ever want to verify that the Clock produces the right kind of timestamps? It seems to me that if you aren't going to be handed a TimerService too, there's no way to detect a deceptive Clock without this extra method.
* when Promise-flavored timer/repeater/notifier is cancelled, this rejects any Promise with `Error('TimerCancelled')`
  * should I use a non-Error rejection object instead, like `{ name: 'TimerCancelled' }` ? We do that with Promises that are rejected because of a vat upgrade, so clients can distinguish those cancellations from actual rejections, but I'm not sure that's appropriate here
  * that Error causes a `vat-timer.js` stack trace to be included in the rejection (userspace does not see it, because of our censorship, but it's noisy in unit tests, and implies something went wrong within vat-timer, when cancellation is entirely controlled by clients). Would it be appropriate for me to create that `Error` object once, at top-level module scope, and re-use the error singleton whenever a rejection is needed? OTOH I suppose that would capture the stack of liveslots and the supervisor when they first import the timer vat code, which is even less relevant.
* All APIs (except for Notifiers and their iterators) now fire with the most recent timestamp, not the time they were originally scheduled for (these would only diverge if the timer device and timer vat suffered delays in communication).
  * Notifiers (and their iterators) fire with the most recent nominal event prior to the query. They simulate an "eager" Notifier that is being triggered every `interval` ticks, from which the client is fetching the most recent update. If start a Notifier with `delay=0` and manage to submit your `getUpdateSince(undefined)` before any time has passed, you'll get a Promise that fulfills promptly to the current time.
  * I *think* this is mostly the same as the previous behavior, but the Zoe manual timer behaved slightly differently, so I wanted to make sure this specified behavior seems correct
* I replaced Zoe's `manualTimer.js` with a logging/awaiting-augmented wrapper around a new SwingSet-provided `manual-timer.js` (which only has `advanceTo(newTime)`)
  * some of the "&& scheduling event for.." log messages were too hard to support, so I removed them from the test cases that expected them
  * the delivery order of some nominally-simultaneous events seems to have changed, so a few "golden transcript" -style tests were modified to match
  * the `fakePriceAuthority` needed to start its Repeater one tick later, to get the same results as before. I'm not entirely sure how this works, so I'm hoping someone else can look at it and make sure I didn't break anything

There is a general problem of how to know when "background" async work has finished, before it is safe to either check for consequences, or to initiate further async activity that effectively polls mutable state that the background work might or might not have finished modifying. The old Zoe `manualTimer.js` offered some synchronization tools, but the new (real) timer API is bigger and doesn't lend itself to as complete a solution. I've added enough promise-tracking `await`s into the Zoe wrapper to support the existing tests (with some small changes), but in general, new tests should probably use `await eventLoopIteration()` (or some other `setImmediate` -based wrapper, *not* just a few `await Promise.resolve()`s) after calling `tick()`, if they want to ensure that all the wakeups thus triggered have finally rolled to a halt.

This is not something the `manualTimer` can do internally, because several zoe/etc `test/swingsetTests/` use the manual timer *inside a vat*, where `setImmediate` is not available. I changed `tickN()` to accept `eventLoopIteration` as an argument, which it will call and `await` if present, so the "unit" unit-tests can do that, while the "swingset" unit-tests do not.

I have not yet changed the TimerService API to use Brand-bearing timestamps, but the cutover will be just a few lines in `vat-timer.js`, and I'm guessing we should do that very soon. The swingset unit tests will require more work to adapt, but it's mostly mechanical.

packages/SwingSet/docs/timer.md
```diff
@@ -67,42 +67,86 @@ A single application might have multiple sources of time, which would require th
...
 
-  // makeRepeater() creates a repeating wakeup service: the handler will
+  // repeatAfter() creates a repeating wakeup service: the handler will
   // fire somewhat after 80 seconds from now (delay+interval), and again
-  // every 60 seconds thereafter. Individual wakeups might be delayed,
-  // but the repeater will not accumulate drift.
+  // every 60 seconds thereafter. The next wakeup will not be scheduled
+  // until the handler message is acknowledged (when its return promise is
+  // fulfilled), so wakeups might be skipped, but they will always be
+  // scheduled for the next 'start + k * repeat', so they will not
```
Chris-Hibbert:
```suggestion
  // scheduled for the next 'delay + k * interval', so they will not
```

packages/SwingSet/docs/timer.md
```diff
@@ -67,42 +67,86 @@ A single application might have multiple sources of time, which would require th
...
 
   const delay = 20n;
   const interval = 60n;
+  E(timerService).repeatAfter(delay, interval, handler, cancel2);
+
+  // repeating wakeup service, Notifier-style . This supports both the
+  // native 'E(notifierP).getUpdateSince()' Notifier protocol, and an
+  // asyncIterator. To it in a for/await loop, which does not know how
+  // to make `E()`-style eventual sends to the remote notifier, you must
+  // wrap it in a local "front-end" Notifier by calling `makeNotifier()`.
```
Chris-Hibbert:
```suggestion
  // asyncIterator. To use it in a for/await loop (which does not know how
  // to make `E()`-style eventual sends to the remote notifier) you must
  // wrap it in a local "front-end" Notifier by calling `makeNotifier()`.
```

Chris-Hibbert (commented):
some minor proofreading comments.

I'm done with vat-timer and its primary tests, but have a bunch of other code to review still.

erights:
> Questions for reviewers:
> 
> * Somebody said that the new `timerBrand` did not need a `isMyClock()` method, but now I'm not so sure. If you hold a branded Timestamp and someone hands you a Clock, would you ever want to verify that the Clock produces the right kind of timestamps? It seems to me that if you aren't going to be handed a TimerService too, there's no way to detect a deceptive Clock without this extra method.

That reasoning sounds right to me.

> * when Promise-flavored timer/repeater/notifier is cancelled, this rejects any Promise with `Error('TimerCancelled')`
>   
>   * should I use a non-Error rejection object instead, like `{ name: 'TimerCancelled' }` ? We do that with Promises that are rejected because of a vat upgrade, so clients can distinguish those cancellations from actual rejections, but I'm not sure that's appropriate here
>   * that Error causes a `vat-timer.js` stack trace to be included in the rejection (userspace does not see it, because of our censorship, but it's noisy in unit tests, and implies something went wrong within vat-timer, when cancellation is entirely controlled by clients). Would it be appropriate for me to create that `Error` object once, at top-level module scope, and re-use the error singleton whenever a rejection is needed? OTOH I suppose that would capture the stack of liveslots and the supervisor when they first import the timer vat code, which is even less relevant.

When the throw or rejection is not indicating that something bad happened, not indicating that a possible bug symptom may have just occurred, then I prefer that the thrown thing or the rejection reason not be an error, for all the reasons you state above. I like your suggestion that it should be strongly analogous, while testably distinct, from the rejection reasons with which we indicate upgrade. `harden({ name: 'TimerCancelled' })` seems like a fine choice to me.

> * All APIs (except for Notifiers and their iterators) now fire with the most recent timestamp, not the time they were originally scheduled for (these would only diverge if the timer device and timer vat suffered delays in communication).

This seems right.

>   * Notifiers (and their iterators) fire with the most recent nominal event prior to the query. They simulate an "eager" Notifier that is being triggered every `interval` ticks, from which the client is fetching the most recent update. If start a Notifier with `delay=0` and manage to submit your `getUpdateSince(undefined)` before any time has passed, you'll get a Promise that fulfills promptly to the current time.
>   * I _think_ this is mostly the same as the previous behavior, but the Zoe manual timer behaved slightly differently, so I wanted to make sure this specified behavior seems correct

In the scenario you mention, "...`delay=0`...before any time has passed...", would the old notifier wait a tick first? Neither behavior seems incorrect to me, but the new behavior seems better.

> * I replaced Zoe's `manualTimer.js` with a logging/awaiting-augmented wrapper around a new SwingSet-provided `manual-timer.js` (which only has `advanceTo(newTime)`)
>   * some of the "&& scheduling event for.." log messages were too hard to support, so I removed them from the test cases that expected them
>   * the delivery order of some nominally-simultaneous events seems to have changed, so a few "golden transcript" -style tests were modified to match

Sure. Goldens are fragile, and the notifier spec is inherently non-deterministic.

>   * the `fakePriceAuthority` needed to start its Repeater one tick later, to get the same results as before. I'm not entirely sure how this works, so I'm hoping someone else can look at it and make sure I didn't break anything
> 
> There is a general problem of how to know when "background" async work has finished, before it is safe to either check for consequences, or to initiate further async activity that effectively polls mutable state that the background work might or might not have finished modifying. The old Zoe `manualTimer.js` offered some synchronization tools, but the new (real) timer API is bigger and doesn't lend itself to as complete a solution. I've added enough promise-tracking `await`s into the Zoe wrapper to support the existing tests (with some small changes), but in general, new tests should probably use `await eventLoopIteration()` (or some other `setImmediate` -based wrapper, _not_ just a few `await Promise.resolve()`s) after calling `tick()`, if they want to ensure that all the wakeups thus triggered have finally rolled to a halt.
> 
> This is not something the `manualTimer` can do internally, because several zoe/etc `test/swingsetTests/` use the manual timer _inside a vat_, where `setImmediate` is not available. I changed `tickN()` to accept `eventLoopIteration` as an argument, which it will call and `await` if present, so the "unit" unit-tests can do that, while the "swingset" unit-tests do not.

I'm not following this yet.

> I have not yet changed the TimerService API to use Brand-bearing timestamps, but the cutover will be just a few lines in `vat-timer.js`, and I'm guessing we should do that very soon. The swingset unit tests will require more work to adapt, but it's mostly mechanical.

Good. There's no need to do that in this same PR.

packages/SwingSet/docs/timer.md
warner: aye, but maybe `now + delay + k * interval`

warner:
It looks like 1: zoe's `buildManualTimer` is being used by dapps too, so any API changes must do argument-sensing and be backwards compatible, and 2: the "`fakePriceAuthority` has to start at 1 instead of 0" problem that I saw in zoe will also occur in the dapp tests that have copied that code. So I think I have to dig into the dapps and understand that fencepost problem before I can land this without disruption, sigh.

packages/SwingSet/test/vat-timer-upgrade/test-vat-timer-upgrade.js
```diff
@@ -0,0 +1,173 @@
...
+import { test } from '../../tools/prepare-test-env-ava.js';
+
+// eslint-disable-next-line import/order
+import bundleSource from '@endo/bundle-source';
+import { parse } from '@endo/marshal';
+import { provideHostStorage } from '../../src/controller/hostStorage.js';
+import { initializeSwingset, makeSwingsetController } from '../../src/index.js';
+import { buildTimer } from '../../src/devices/timer/timer.js';
+
+const bfile = name => new URL(name, import.meta.url).pathname;
```
Chris-Hibbert, line 11: is the b short for `bundle`?

packages/zoe/test/unitTests/test-manualTimer.js
```diff
@@ -46,3 +56,142 @@ test('manualTimer makeRepeater', async t => {
...
+  const pk1 = makePromiseKit();
+  const handler = Far('handler', {
+    wake: scheduled => {
+      handlerFired = scheduled;
+      return pk1.promise;
+    },
+  });
+  manualTimer.setWakeup(1n, handler);
+  t.is(handlerFired, false);
+  let tickFired = false
```
Chris-Hibbert:
```suggestion
  let tickFired = false;
```

warner:
Today's discussion pointed out some changes that I need to make:

* change TimerNotifier to use an `updateCount` that starts at `1n` and increments sequentially from there, by using something like `BigInt(Math.floor((now - delay) / interval)+1)`
  * @erights was weirded out by `updateCount` not being sequential
  * this would also simplify the conditionals in `getUpdateSince`
  * also remove the `Error('invalid updateCount for timer Notifier')` clause
  * if the supplied `updateCount` is not `undefined`, then we check whether it matches the current value, or is different: don't do more than that
  * the `+1` is to ensure that a live `updateCount` is not falsy, to guard against clients who accidentally use `if (!updateCount)` instead of `if (updateCount === undefined)` to sense the finish condition
* change TimerNotifier's canceller to put the notifier into a "finish" state, rather than an erroring state
  * cancellation should record the current time, to compute a final update value
  * this makes all outstanding and all future `getUpdateSince` promises fire with `{ value: lastTime, updateCount: lastUpdateCount }`
* zoe's `buildManualTimer` will change from the old `buildManualTimer(log, startTime=0n, timeStep=1n)` to `buildManualTimer(log=dummy, startTime=0n, options={ timeStep=1n, eventLoopIteration }`
  * if `eventLoopIteration` is provided, `tick()` will finish with `await eventLoopIteration()`
  * this removes the `{ doWait: true }` I was using
  * unit tests will need to change their `buildManualTimer` calls to provide `eventLoopIteration`, but then they won't need to do their own call after tick()
* zoe's `fakePriceAuthority` contains a race because it has multiple uncoordinated calls into the timer, specifically `startTicker` vs `quoteAtTime`
  * the plan is to make `quoteAtTime` follow the `ticker` (the notifier driven by `startTicker`), rather than using `E(timer).setWakeup` itself
  * so `quoteAtTime` looks like:
```js
if (timeStamp <= latestTick) {
  return resolve(priceInQuote(amountIn, brandOut, latestTick));
}
// else push `resolve` into a map that is checked within the `startTicker` handler
```

We walked through the simplest `test-fakePriceAuthority.js` test (`priceAuthority quoteAtTime`), and we think the fourth call to `tick()` was a workaround for the lack of complete `eventLoopIteration` synchronication: the `time = 3n` update (which is odd, not even, so the price should be 55) should be fired by the third call to `tick()`. The failures I've seen in my branches are because the race between the `quoteAtTime` call to `E(timer).setWakeup` and the `startTicker` use of `E(timer).setRepeater`. And `setRepeater(0)` now fires earlier, which doesn't change latestTick but *does* start incrementing `currentPriceIndex` earlier.

I see calls to `buildManualTimer` in three external repositories: `dapp-card-store`, `dapp-oracle`, and `dapp-otc`. None of them use the `timeStep` argument, so it should be relatively safe to convert that argument slot into an options bag.

I see `fakePriceAuthority` mentioned in two external repos: `dapp-oracle` and `dapp-treasury`. The `dapp-oracle` copy is significantly different than the one in zoe: I see things like `makeLinearPriceAuthority` and `makeSinglePriceAuthority`. I don't know if that makes it newer or older than the zoe one, but it will probably require its own analysis. I haven't looked at the `dapp-treasury` copy yet, but it looks like it's in an example, not the main code.

packages/SwingSet/test/vat-timer-upgrade/test-vat-timer-upgrade.js
warner: I think so? I copied this from something I wrote a long time ago, and I was so terse I forget what it was an abbreviation for.

warner:
That commit (ce2347b49) implements the Notifier/Iterator changes. For clarity: the "finish state" of a cancelled TimerNotifier has a `value` of the cancellation time, not the latest update (which wouldn't exist anyways, if the Notifier was cancelled before the first event happened).

I'm still working on the `manualTimer` and `priceAuthority` changes, so zoe tests are probably borked.

warner:
I think I understand the fencepost error. In the old version (on current trunk):

* Zoe's `manualTimer.makeRepeater(delay=0)` does not wake right away (while `time = 0`). Instead, the first `wake()` occurs after the test calls `tick()` for the first time, and the handlers get a `wake(time=1)`.
* `fakePriceAuthority` has a `handler` that is subscribed to that `makeRepeater(delay=0)`
  * the first time `handler.wake` is called, it leaves `currentPriceIndex` alone (it remains as 0)
  * all subsequent times, it increments `currentPriceIndex`
  * it records the current timestamp in `latestTick`, then updates a Notifier named `ticker` with the timestamp
    * `ticker` is consumed by `generateQuotes`
* some of the API is pretty immediate: `quoteGiven`/`quoteWanted` grabs the current timestamp and uses `currentPriceIndex` to look up the current price, then returns a quote
* `quoteWhenGTE`/etc push a resolver onto `comparisonQueue`, and each time `handler.wake` fires, it resolves anything in the queue that is matched by the new price
* `quoteAtTime` uses its own `E(timer).setWakeup(timeStamp)` to wake a handler at the requested time. It then takes the handler's `time` argument and uses `priceInQuote()` to build a quote from the current `currentPriceIndex` and `time`

The main problem was that `quoteAtTime` is scheduling a wakeup for the same time as the `makeRepeater`, and the invocation order of two such "simultaneous" alarms is not a well-specified part of the TimerService API. The `test-fakePriceAuthority.js` test named `priceAuthority quoteAtTime` was accidentally dependent upon a specific ordering of events:

* the test calls `E(priceAuthority).quoteAtTime(when=3n)`, which schedules a `wakeAt(3n)`
* `tick()` is called several times, all of which fire the `makeRepeater` handler in `startTicker`
  * the `tick()` that advances time from `0n` to `1n` leaves `currentPriceIndex` at 0, and sets `latestTick = 1n`
  * the `tick()` that advances time from `1n` to `2n` updates `currentPriceIndex` to 1, and sets `latestTick = 2n`
* the `wakeAt(3n)` handler fires first, is handed `time=3n`, and `priceInQuote()` is called with `time=3n`
  * it polls `currentPriceIndex` and gets 1, which tells it to use the second time in the `[20, 55]` repeating list
  * the `quoteAtTime()` return promise resolves to a quote with `time=3n` and a price based on 55
* then the `makeRepeater()` handler fires with `time=3n`, which updates `currentPriceIndex` to 2, and sets `latestTick = 3n`

So the test was asking for the price as of time 3n, but that was a bit ambiguous as to whether it wanted the price before or after the `startTicker` handler had a chance to increase the `currentPriceIndex`. The test behavior depended upon getting the price *before* the increment.

In the new version:
* I changed the `makeRepeater` to start at time=1, not 0, to avoid the extra `wake(time=0)` call from causing the `currentPriceIndex` to be incremented one time too many
  * I also changed the handler to return immediately if `time===0n`, as a backup
* to avoid an ambiguity, I changed `quoteAtTime` to remove the uncoordinated `setWakeup()` call
  * instead, I added a `timeClients` list
  * `quoteAtTime` pushes a `[when, resolver]` pair onto that list
  * the `startTicker` handler updates `currentPriceIndex` and `latestTick`, then walks `timeClients` to find entries where `when >= latestTick`, and calls the resolver
  * this ensures that the `quoteAtTime` action happens consistently *after* the `currentPriceIndex` has been updated

As a result, when the `priceAuthority quoteAtTime` test wants a price of 55, it really needs to ask for the price at time 2n, not 3n. I updated the test to use `quoteAtTime(2n)`, and to expect to see `2n` in the quote's timestamp.

I hit a similar problem in `zoe/test/unitTests/contracts/test-callSpread.js`, which exercises timed options. I had to change the `expiration: 3n` to `2n` so they would sample the expected price.

Most of the other tests are working now that I've completed the mechanical changes of the `buildManualTimer()` signature, and passing in `eventLoopIteration` on the tests that need it.

warner:
I walked through `test-vaultFactory.js` with @Chris-Hibbert just now. It looks like the old version was doing more `tick()`s than it should have needed, because the old `tick()` was unable to completely wait for the promise queue to drain. So the test had evolved to meet the bugs in the infrastructure. The new `tick()` (in that test, and several others) is configured to use `eventLoopIteration()` to fully drain the queue, and so it needs one fewer tick. I'll put a note in the commit comment with the details.

warner:
Ok, everything should be working now. I suggest looking at each commit separately: the first one is just swingset, the second is the less-magical stuff in zoe/governance/inter-protocol that needed updating, and the last two are the more-magical stuff that needed updating. The commit comments contain a deep description of the magic.

The dapp-oracle tests are broken by the changed behaviors. I have a PR to fix it (https://github.com/Agoric/dapp-oracle/pull/40) which I'll land as soon as this one lands.

I still owe changes to https://github.com/Agoric/documentation to update our public docs on this API.

This does not yet enable branded timestamps, nor does it seek to POLA de-power clients from using the full TimerService into using just Clock or TimerBrand. Also the types might be wonky: I updated everything I could figure out, but the main `vat-timer.js` is not yet marked as returning a `{TimerService}` (I tried, but got stuck with something around iterators).

This is marked as merely referring to #4286 (instead of closing it) because `device-timer.js` is left untouched, so the "ever-growing Repeater array" is still there, however now vat-timer doesn't use that feature of the device, so it doesn't matter. I'll re-milestone that ticket once this lands, and a future cleanup pass can strip out everything in the device except the one `setWakeup` that we still need.

Likewise, this refs #5798 instead of closing it because we aren't yet using branded time.

warner:
I re-rebased because some zoe durability changes just landed.. wanted to make sure they didn't conflict. Sorry for the churn.

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,977 @@
...
+  // versions must define behaviors for these. Search for calls to
+  // 'vivifyKind', 'vivifySingleton', or 'defineDurableKindMulti'.
+  // * oneShotEvent
+  // * promiseEvent
+  // * repeaterEvent
+  // * timerRepeater
+  // * timerNotifier
+  // * timerIterator
+  // * wakeupHandler
+  // * timerservice
```
Chris-Hibbert: s/timerservice/timerService/

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,977 @@
...
+    when = fromTimestamp(when);
+    const now = getNow();
+    if (when <= now) {
+      return Promise.resolve(toTimestamp(now));
+    }
+    return wakeAtInternal(when, cancelToken);
+  }
+
+  /**
+   * delay(delay): return a Promise that fires (with the scheduled
```
Chris-Hibbert: addDelay(delay):

packages/SwingSet/src/vats/timer/vat-timer.js
```diff
@@ -1,98 +1,977 @@
...
+    delay = fromRelativeTime(delay);
+    assert(delay >= 0n, 'delay must be non-negative');
+    interval = fromRelativeTime(interval);
+    assert(interval > 0n, 'interval must be nonzero');
+    const start = getNow() + delay;
+    const active = false;
+    return { start, interval, active };
+  }
+  const repeaterFacets = {
+    cancel: {}, // internal CancelToken
```
Chris-Hibbert: Please put the comment before the code it refers to. Here, the comment is after the comma, and all the tools think that means it applies to `repeater` rather than `cancel`. (If I hover over `repeater`, I see the comment.)

Chris-Hibbert:
@warner, this new code should be converted to arrow function style.   WebStorm has good support for converting. Would you like me to push a commit that fixes all the `functions` here?

packages/SwingSet/test/timer/test-timer.js
```diff
@@ -72,4 +72,48 @@ test('timer vat', async t => {
 
   const cd9 = await run('banana', [10n]);
   t.deepEqual(parse(cd9.body), 'bad setWakeup() handler');
+
+  // start a repeater that should first fire at now+delay+interval, so
+  // 7+20+10=27,37,47,57,..  TODO: poll at 25,35,40
```
Chris-Hibbert: Why `TODO`?   The test does poll() 25,35,40.

packages/SwingSet/test/test-vat-timer.js
```diff
@@ -0,0 +1,1166 @@
...
+  t.false(brand.isMyTimerService(clock));
+
+  t.true(brand.isMyClock(clock));
+  t.false(brand.isMyClock({}));
+  t.false(brand.isMyClock(brand));
+  t.false(brand.isMyClock(ts));
+});
+
+test('clock', async t => {
+  // clock.getCurrentTimestamp() (and nothing else)
```
Chris-Hibbert: and `clock.getTimerBrand()`?

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -782,16 +784,14 @@ test('price falls precipitously', async t => {
...
+  await manualTimer.tick(); // t 0->1, p 2200->19180
   await assertDebtIs(debtAmount.value);
 
-  await manualTimer.tick();
+  await manualTimer.tick(); // t 1->2, p 19180->1650
   await assertDebtIs(debtAmount.value);
 
-  await manualTimer.tick();
+  await manualTimer.tick(); // t 2->3, p 1650->150, liquidates
+  await eventLoopIteration();
```
Chris-Hibbert: Isn't this `eventLoopIteration();` unnecessary?

Chris-Hibbert (commented):
This looks good.

The main thing it needs is conversion of the `function`s to arrow-style (which I'll happily help with).

I have a couple of question/suggestions; nearly all on comments, though one is on the `eventLoopIteration()` added back in test-vaultFactory.js.

warner:
> @warner, this new code should be converted to arrow function style. WebStorm has good support for converting. Would you like me to push a commit that fixes all the `functions` here?

Yeah, I guess so. I find the arrow style less readable, so I haven't been using it in kernel code, but I'd like to see how it looks, and maybe that will get me over it. Let's convert everything in `vat-timer.js`, `manual-timer.js`, `test-vat-timer.js`, and `test-manual-timer.js`.. I think that would avoid single files having a mix of styles, while not increasing the scope of the PR.

packages/SwingSet/src/vats/timer/vat-timer.js
warner: Oh, hm, so here the function is defined as `addDelay`, but when exported on `TimerService` it's named `delay`. I changed the local function name to avoid confusion with the `delay` parameter that it takes, and so the reader can correlate it with the external API. Which way do you think I should spell it in the comment? And, would it be a good idea to move this to `const delay = (delay, cancelToken) => ..` instead of `const addDelay = ..` ? The visual confusion would still exist, and I think eslint would still complain about the scope shadowing...

packages/SwingSet/src/vats/timer/vat-timer.js
warner: Aye. Do you know of any way to maintain the visual conciseness of:

```js
  const timerService = vivifySingleton(baggage, 'timerService', {
    getCurrentTimestamp,
    setWakeup, // one-shot with handler (absolute)
    wakeAt, // one-shot with Promise (absolute)
    delay: addDelay, // one-shot with Promise (relative)
    makeRepeater, // repeater with Repeater control object (old)
    repeatAfter, // repeater without control object
    makeNotifier, // Notifier
    cancel, // cancel setWakeup/wakeAt/delay/repeat
    getClock: () => timerClock,
    getTimerBrand: () => timerBrand,
  });
```

given that IDE interpretation constraint? I'd find the following less readable, although maybe that's the only way to avoid the problem:
```js
  const timerService = vivifySingleton(baggage, 'timerService', {
    getCurrentTimestamp,
    // one-shot with handler (absolute)
    setWakeup,
    // one-shot with Promise (absolute)
    wakeAt,
    // one-shot with Promise (relative)
    delay: addDelay,
    // repeater with Repeater control object (old)
    makeRepeater,
    // repeater without control object
    repeatAfter,
    // Notifier
    makeNotifier,
    // cancel setWakeup/wakeAt/delay/repeat
    cancel,
    getClock: () => timerClock,
    getTimerBrand: () => timerBrand,
  });
```

packages/SwingSet/test/timer/test-timer.js
warner: you're right, that's stale, I'll remove it

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
warner: correct, I'll remove it

packages/SwingSet/src/vats/timer/vat-timer.js
Chris-Hibbert: The comment should reflect the local name.  If you want, it would be fine to add a parenthetical remark that it will be exported as `delay(delay, cancelToken)`.

I don't think `const delay = (delay, ...)` would be better.

packages/SwingSet/src/vats/timer/vat-timer.js
Chris-Hibbert: This is as concise. The use of `/* */` comments means that typescript doesn't pick them up, but perhaps that's what you want here.

``` js
  const timerService = vivifySingleton(baggage, 'timerService', {
    getCurrentTimestamp,
    setWakeup /* one-shot with handler (absolute) */,
    wakeAt /* one-shot with Promise (absolute) */,
    delay: addDelay /* one-shot with Promise (relative) */,
    makeRepeater /* repeater with Repeater control object (old) */,
    repeatAfter /* repeater without control object */,
    makeNotifier /* Notifier */,
    cancel /* cancel setWakeup/wakeAt/delay/repeat */,
    getClock: () => timerClock,
    getTimerBrand: () => timerBrand,
  });
```

Chris-Hibbert:
>> @warner, this new code should be converted to arrow function style. WebStorm has good support for converting. Would you like me to push a commit that fixes all the functions here?

> Yeah, I guess so. ... Let's convert everything in vat-timer.js, manual-timer.js, test-vat-timer.js, and test-manual-timer.js.. I think that would avoid single files having a mix of styles, while not increasing the scope of the PR.

I pushed [b6d8ab6](https://github.com/Agoric/agoric-sdk/pull/5847/commits/b6d8ab60eb3686a8231869817a7c8184e127d591)

> I find the arrow style less readable, so I haven't been using it in kernel code, but I'd like to see how it looks, and maybe that will get me over it. 

I agree about readability, but there were other considerations that we decided mattered even more than readability.

packages/SwingSet/src/vats/timer/vat-timer.js
warner: ah! I hadn't even thought about `/**/` being treated differently. Applied, thanks.

Chris-Hibbert (approved):

warner:
@FUDCo and I walked through this code today, so I think I have a virtual r+ from him. We identified a couple of things to be cleaned up, and I did some refactoring to save a few duplicate `getNow()` device calls. @erights said to not wait for him. And I think @turadg has seen everything he needs to know about the test changes. So I think I'm ready to land. I'll rebase (to fix that `test-priceAggregator.js` merge conflict) and squash back down into a small number of commits.

FUDCo (approved):
Go fer it.

packages/zoe/tools/manualTimer.js
```diff
@@ -1,197 +1,97 @@
...
-  startValue = TimeMath.toAbs(startValue);
-  timeStep = TimeMath.toRel(timeStep);
-  let ticks = startValue;
 
-  /** @type {MapStore<Timestamp, ERef<TimerWaker>[]>} */
-  const schedule = makeScalarMapStore('Timestamp');
+const buildManualTimer = (log = nolog, startValue = 0n, options = {}) => {
+  const {
+    timeStep = 1n,
+    eventLoopIteration = () => 0,
```
turadg, line 63: when I looked over the timer PR test change I didn’t see anything surprising except that eventLoopIteration has to be passed in as an arg. I remember skimming it and thinking “okay” but I just took another look and I’m confused why it has to be passed in as a function instead of a flag. E.g. `iterateEventLoop: boolean = false`.

@warner what motivates receiving an arbitrary function?

warner: There are a handful of "swingset" unit tests (as opposed to "unit" unit tests) which use the manual timer too, which means they import `manualTimer.js` into code that runs inside a real vat. And `setImmediate` is not available to vat code (the kernel doesn't populate it in the Compartments that houses the vat code) to keep vats from retaining agency after claiming their crank/delivery has finished. So `manualTimer.js` can't provide that functionality by itself.

We experimented with a version that used `if (globalThis.setImmediate)` to try and work with both, but it ended up being unsatisfying, and it wasn't clear that all tests would want (or could even tolerate) the automatic wait-for-flush behavior all the time, so it seemed better to give tests a choice.

That might be worth revisiting, now that we figured out the other test failures (which turned out to be the result of different assumptions about timers that are born ready, and races between "simultaneous" events).

turadg: Ah. I'll humbly ask that we make the common case simpler (without real vat) even if it makes the utility more complex. Consider:
- add default option `iterateEventLoop = false`
- make `eventLoopIteration` option default to undefined
- when `iterateEventLoop` is true, manualTimer does that automatically if it can
- if it can't, it uses `eventLoopIteration` option and throws an error if it's not set, informing the dev that with a real vat it must be passed in

PR #5849, opened 2022-07-28
feat(wallet): shared unserializer for dapps and wallet

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/4398

- Adds a shared unserializer to the wallet package
- Changes the wallet API to provide board ids in its slots so the shared unserializer can map brands together
- Adds a `makeFollower` method to the dapp UI bridge so dapps can use the low privilege unserializer from the wallet

### DIFF
--- a/packages/wallet/api/src/lib-dehydrate.js
+++ b/packages/wallet/api/src/lib-dehydrate.js
@@ -337,7 +337,7 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
         return harden({
           kind,
           petname: valToPetname.get(val),
-          boardId,
+          ...(boardId ? { boardId } : {}),
         });
       }
     }
--- a/packages/wallet/api/src/lib-dehydrate.js
+++ b/packages/wallet/api/src/lib-dehydrate.js
@@ -1,7 +1,7 @@
 // @ts-check
 
 import { makeMarshal, mapIterable } from '@endo/marshal';
-import { makeLegacyMap, makeScalarMap } from '@agoric/store';
+import { makeLegacyMap, makeScalarMap, makeScalarWeakMap } from '@agoric/store';
 import { assert, details as X, q } from '@agoric/assert';
 
 /**
@@ -159,6 +159,8 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
       },
     };
 
+    const valToBoardId = makeScalarWeakMap('boardId');
+
     /**
      * @param {Path} path
      * @param {any} val
@@ -217,6 +219,11 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
       if (isPath(petname)) {
         addPath(petname, val);
       }
+    };
+
+    const addBoardId = (boardId, val) => {
+      assert(!valToBoardId.has(val), X`val ${val} already has a board id`);
+      valToBoardId.init(val, boardId);
     };
 
     const renamePetname = (petname, val) => {
@@ -289,12 +296,14 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
       valToPetname,
       valToPaths,
       petnameToVal,
+      valToBoardId,
       addPetname,
       addPath,
       renamePetname,
       deletePetname,
       suggestPetname,
       kind,
+      addBoardId,
     });
     petnameKindToMapping.init(kind, mapping);
     return mapping;
@@ -319,11 +328,16 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
   const convertValToName = val => {
     for (let i = 0; i < searchOrder.length; i += 1) {
       const kind = searchOrder[i];
-      const { valToPetname } = petnameKindToMapping.get(kind);
+      const { valToPetname, valToBoardId } = petnameKindToMapping.get(kind);
       if (valToPetname.has(val)) {
+        let boardId;
+        if (valToBoardId.has(val)) {
+          boardId = valToBoardId.get(val);
+        }
         return harden({
           kind,
           petname: valToPetname.get(val),
+          boardId,
         });
       }
     }
@@ -340,6 +354,10 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
     return placeholderName;
   };
 
+  /**
+   * @param {{ kind: string, petname: Petname, boardId?: string}} slot
+   * @param {string} [_iface]
+   */
   const convertNameToVal = ({ kind, petname }, _iface = undefined) => {
     const { petnameToVal } = petnameKindToMapping.get(kind);
     return petnameToVal.get(petname);
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -131,6 +131,22 @@ export function makeWallet({
   const installationMapping = makeMapping('installation');
 
   const brandTable = makeIssuerTable();
+
+  const initBrandToBoardId = async brand => {
+    const { valToBoardId: brandToBoardId } = brandMapping;
+    if (brandToBoardId.has(brand)) {
+      return brandToBoardId.get(brand);
+    }
+
+    const brandBoardId = await E(board).getId(brand);
+    if (brandToBoardId.has(brand)) {
+      return brandToBoardId.get(brand);
+    }
+
+    brandMapping.addBoardId(brandBoardId, brand);
+    return brandBoardId;
+  };
+
   /** @type {WeakStore<Issuer, string>} */
   const issuerToBoardId = makeScalarWeakMap('issuer');
 
@@ -296,7 +312,7 @@ export function makeWallet({
     const { value, brand } = currentAmount;
     const brandPetname = brandMapping.valToPetname.get(brand);
     const dehydratedCurrentAmount = dehydrate(currentAmount);
-    const brandBoardId = await E(board).getId(brand);
+    const brandBoardId = brandMapping.valToBoardId.get(brand);
 
     let depositBoardId;
     if (
@@ -786,6 +802,7 @@ export function makeWallet({
       : brandTable.initIssuer(issuer, addMeta);
     const { brand } = await recP;
     await initIssuerToBoardId(issuer);
+    await initBrandToBoardId(brand);
     const addBrandPetname = () => {
       let p;
       const already = brandMapping.valToPetname.has(brand);
@@ -807,6 +824,7 @@ export function makeWallet({
   const publishIssuer = async brand => {
     const { issuer } = brandTable.getByBrand(brand);
     const issuerBoardId = await initIssuerToBoardId(issuer);
+    await initBrandToBoardId(brand);
     updateAllIssuersState();
     return issuerBoardId;
   };
@@ -1602,6 +1620,7 @@ export function makeWallet({
     add: async (petname, issuerP) => {
       const { brand, issuer } = await brandTable.initIssuer(issuerP, addMeta);
       await initIssuerToBoardId(issuer);
+      await initBrandToBoardId(brand);
       brandMapping.suggestPetname(petname, brand);
       await updateAllIssuersState();
     },
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -132,33 +132,21 @@ export function makeWallet({
 
   const brandTable = makeIssuerTable();
 
-  const initBrandToBoardId = async brand => {
-    const { valToBoardId: brandToBoardId } = brandMapping;
-    if (brandToBoardId.has(brand)) {
-      return brandToBoardId.get(brand);
-    }
-
-    const brandBoardId = await E(board).getId(brand);
-    if (brandToBoardId.has(brand)) {
-      return brandToBoardId.get(brand);
-    }
-
-    brandMapping.addBoardId(brandBoardId, brand);
-    return brandBoardId;
-  };
-
   /** @type {WeakStore<Issuer, string>} */
   const issuerToBoardId = makeScalarWeakMap('issuer');
 
-  // Idempotently initialize the issuer's synchronous boardId mapping.
-  const initIssuerToBoardId = async issuer => {
+  // Idempotently initialize the issuer and brand's synchronous boardId mapping.
+  const initIssuerToBoardId = async (issuer, brand) => {
     if (issuerToBoardId.has(issuer)) {
       // We already have a mapping for this issuer.
       return issuerToBoardId.get(issuer);
     }
 
     // This is an interleaving point.
-    const issuerBoardId = await E(board).getId(issuer);
+    const [issuerBoardId, brandBoardId] = await Promise.all([
+      E(board).getId(issuer),
+      E(board).getId(brand),
+    ]);
     if (issuerToBoardId.has(issuer)) {
       // Somebody else won the race to .init.
       return issuerToBoardId.get(issuer);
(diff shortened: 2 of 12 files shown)


dckc:
see also

https://github.com/Agoric/agoric-sdk/blob/df7b625bae9b0dfe073bf3a29fef12ef551befeb/packages/wallet/api/test/test-sharing-unserializer.js
https://github.com/Agoric/agoric-sdk/blob/df7b625bae9b0dfe073bf3a29fef12ef551befeb/packages/wallet/api/src/sharingUnserializer.js

packages/inter-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -180,12 +180,16 @@ const start = async (zcf, privateArgs) => {
...
-  const updateMetrics = () => {
+
+  const updateMetrics = () =>
     metricsPublication.updateState(
-      harden({ XYK: Array.from(secondaryBrandToPool.keys()) }),
+      harden({
+        XYK: Array.from(secondaryBrandToPool.keys()),
+        central: centralBrand,
+        poolFee: params.getPoolFee(),
+        protocolFee: params.getProtocolFee(),
```
dckc: These are in `published.amm.governance` I'm pretty sure.

samsiegart: Confirmed. I removed this file from the PR now anyway though.

packages/wallet/api/src/lib-wallet.js
```diff
@@ -131,6 +131,22 @@ export function makeWallet({
   const installationMapping = makeMapping('installation');
 
   const brandTable = makeIssuerTable();
+
+  const initBrandToBoardId = async brand => {
+    const { valToBoardId: brandToBoardId } = brandMapping;
```
turadg: within this function name `initBrandToBoardId`, that the mapping is `brandToBoardId` is implied.

consider 
```suggestion
    const { valToBoardId: map } = brandMapping;
```

packages/wallet/api/src/lib-wallet.js
```diff
@@ -131,6 +131,22 @@ export function makeWallet({
...
+    if (brandToBoardId.has(brand)) {
+      return brandToBoardId.get(brand);
+    }
+
+    const brandBoardId = await E(board).getId(brand);
```
turadg: why could this call have the side-effect of changing `brandToBoardId.has(brand)` from false to true?

packages/wallet/api/src/lib-wallet.js
```diff
@@ -786,6 +802,7 @@ export function makeWallet({
       : brandTable.initIssuer(issuer, addMeta);
     const { brand } = await recP;
     await initIssuerToBoardId(issuer);
+    await initBrandToBoardId(brand);
```
turadg: do these always get init-ed together? That would be a helpful constraint. You wouldn't have to check each map before inserting, just either one.

Consider adding `brand` as an arg to `initIssuerToBoardId(issuer, brand)`. Or if there are cases that issuer has to be inited without brand making a new function.

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,72 @@
+// @ts-check
+import { Far, makeMarshal } from '@endo/marshal';
+
+/**
+ * Make a pair of unserializers that share low-privilege values.
+ *
+ * @param {string} [lowProperty]
+ */
+export const makeSharingUnserializer = (lowProperty = 'boardId') => {
```
turadg: while this may be the common use case it doesn't make sense to me as a _default_.
```suggestion
export const makeSharingUnserializer = (lowProperty) => {
```

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,72 @@
+// @ts-check
+import { Far, makeMarshal } from '@endo/marshal';
+
+/**
+ * Make a pair of unserializers that share low-privilege values.
+ *
+ * @param {string} [lowProperty]
```
turadg: why is there exactly one `lowProperty`?

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,72 @@
+// @ts-check
+import { Far, makeMarshal } from '@endo/marshal';
+
+/**
+ * Make a pair of unserializers that share low-privilege values.
```
turadg, line 5: please document what is functionally different between the two marshallers returned and how the argument conditions that.

turadg (commented):
Took a pass but not ready to approve until I understand it better. Left some questions.

packages/wallet/api/src/lib-dehydrate.js
```diff
@@ -319,11 +328,16 @@ export const makeDehydrator = (initialUnnamedCount = 0) => {
...
+      const { valToPetname, valToBoardId } = petnameKindToMapping.get(kind);
       if (valToPetname.has(val)) {
+        let boardId;
+        if (valToBoardId.has(val)) {
+          boardId = valToBoardId.get(val);
+        }
         return harden({
           kind,
           petname: valToPetname.get(val),
+          boardId,
```
dckc: any particular reason to include the `boardId` property when it's undefined?

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -318,16 +318,18 @@ test('lib-wallet issuer and purse methods', async (/** @type {LibWalletTestConte
...
         currentAmount: {
-          brand: { kind: 'brand', petname: 'zoe invite' },
+          brand: { boardId: 'board0223', kind: 'brand', petname: 'zoe invite' },
           value: [],
         },
       },
       {
         brand: purseLog[1].brand,
-        brandBoardId: 'board0566',
+        brandBoardId: 'board0425',
```
dckc, line 332: Why did this boardId change?

dckc (commented):
I'd like @warner to take a pretty close look at this.

@samsiegart it looks like the scope is a little larger than just the unserializer. Go ahead and say what else is in there in the PR description? Or narrow the scope of the PR.

packages/wallet/api/src/lib-dehydrate.js
samsiegart: Nope, removed now.

packages/wallet/api/src/lib-wallet.js
samsiegart: It wouldn't, but `brandToBoardId` might be called again with the same brand before this resolves. Tbh I was just following the pattern in `initIssuerToBoardId`.

packages/wallet/api/src/lib-wallet.js
samsiegart: Yea I think that's fine, done.

packages/wallet/api/src/sharingUnserializer.js
samsiegart: Added some documentation

packages/wallet/api/src/sharingUnserializer.js
samsiegart: Documented, hopefully that clarifies

packages/wallet/api/src/sharingUnserializer.js
samsiegart: Sure, removed the default

packages/wallet/api/test/test-lib-wallet.js
samsiegart: Maybe because the order of the `getId` invocations changed? It seems correct though, as long as the board id is the same as that in the amount, and it's not the same as another brand.

samsiegart:
> Go ahead and say what else is in there in the PR description? Or narrow the scope of the PR.

Added some more details

packages/wallet/api/src/lib-wallet.js
samsiegart: Removed this method

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * The "high" privilege unserializer can understand objects as slots, such as
+ * those from the wallet serializer. The "low" privilege unserializer can only
+ * understand strings as slots, such as those from the board serializer.
+ *
+ * This provides a way, for instance, to get identical brand references from
+ * both a public contract and a user's purses.
+ *
+ * @param {string} [lowProperty] - The key name to check on a high privilege
```
turadg, line 14: no longer optional

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,81 @@
...
+ * slot object. If the string value at this key is equivalent to the value of
+ * another low privilege slot, the high privilege unserializer will return the
+ * same object for both slots.
+ */
+export const makeSharingUnserializer = lowProperty => {
+  assert(lowProperty, 'lowProperty must be specified for sharing unserializer');
+  const seen = {
+    // high: new WeakMap(),  // ISSUE: WeakMap conflicts with string keys
```
turadg, line 22: is that ISSUE one that can be resolved? if not, I'd expect to see a comment like "Ideally this would be a weak map for `<reasons>` but that doesn't handle string keys"

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,81 @@
...
+          return seen.high.get(key);
+        }
+        obj = Far(iface, {});
+        seen.high.set(key, obj);
+      }
+      return obj;
+    },
+
+    /**
+     * @param {Record<string, any>} slot
```
turadg, line 55: IIUC,
> The "low" privilege unserializer can only understand strings as slots
```suggestion
     * @param {string} slot
```

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,81 @@
...
+  const slotToVal = {
+    /**
+     * @param {Record<string, any>} slot
+     * @param {string} iface
+     */
+    high: (slot, iface) => {
+      let obj;
+      let key;
+
+      if (lowProperty in slot) {
```
turadg, line 36: this condition seems more like delegating to a fallback, breaking out of the primary logic of the function. Refactoring that way would make it more clear and also allow immutable bindings.

```js
if (lowProperty in slot) {
  return slotToVal.low(slot[lowProperty]);
}

const key = JSON.stringify(slot);
…
const obj = Far(iface, {});
```

turadg (commented):
Very helpful documentation.

I don't have blockers but I'll withhold Approve so that it requires a @warner approval.

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,81 @@
...
+        key = slot[lowProperty];
+        if (seen.low.has(key)) {
+          return seen.low.get(key);
+        }
+        obj = Far(iface, {});
+        seen.low.set(key, obj);
+      } else {
+        key = JSON.stringify(slot);
```
warner, line 44: `JSON.stringify` is sensitive to the insertion order of the keys, so if a caller sends `{ foo, bar }` and then later refers to the same object with `{ bar, foo }`, this will incorrectly return a new `Far` presence, even though those should be semantically identical. (I've used `djson` in Swingset for deterministic serialization, it basically just sorts the keys alphabetically).

packages/wallet/api/src/sharingUnserializer.js
```diff
@@ -0,0 +1,81 @@
...
+      }
+      return obj;
+    },
+
+    /**
+     * @param {Record<string, any>} slot
+     * @param {string} iface
+     */
+    low: (slot, iface) => {
+      if (seen.low.has(slot)) {
```
warner, line 59: `Map` (and our `scalarMapStore`) keys off `Object` identity. If `slot` is an `Object` and not a string, a caller who sends `{ foo, bar }` first and then (a distinct) `{ foo, bar}` second, will get two different presences, when they should get the same. The supplier of the serialized data may or may not supply those semantically-identical slot values as identical objects, it depends upon how they're getting them. If they're starting with strings, and doing a `JSON.parse`, then they're almost certainly going to be distinct values, so you'll get unique `Far` presences for each act of deserialization.

I'd highly recommend using string keys if you can find a way to do it, to avoid both of these problems.

And you should definitely have a test that deserializes the same data twice and asserts that you're getting identical (`test.is(old, new)`) presences.

packages/wallet/api/src/sharingUnserializer.js
warner, line 55: If the `low` serializer is meant for data coming from the AMM (and serialized by the Board's marshaller, whose emitted slots are boardID strings), then maybe this signature should be `@param {string} slot` instead of a Record, which would address the concern below.

packages/wallet/api/test/test-sharing-unserializer.js
```diff
@@ -0,0 +1,102 @@
...
+      {
+        kind: 'purse',
+        petname: 'ATOM',
+      },
+    ],
+  });
+
+  const walletState = fromWallet.unserialize(walletCapData);
+  t.is(walletState[0].brand, b1, 'unserialization across sources');
+
```
warner, line 95: to exercise the bug I'm worried about, this should also do:
```js
const walletCapData2 = myWallet.publish()
const walletState2 = fromWallet.unserialize(walletCapData2);
t.is(walletState[0].purse, walletState2[0].purse);
```

The `Brand` will be looked up by the string `boardId` in both cases, but the purse will be looked up by an `Object`-keyed entry in the "high" table.

warner (commented):
The basic core is sound, but confusing. I'm worried about the "slots are objects with various properties", because of the difficulty of using them as keys properly (deterministic serialization, or accidentally relying upon the `Object` identity). And I'm worried about mixing that mode with the more traditional "slots are just strings" in the same deserializer.

The `Object`-identity-as-key definitely needs to be fixed, that will cause spurious duplication of received presences. You might be able to get away without deterministic serialization of the "high" slot objects into string keys, but it's kind of leaving a land mine around for some future maintainer (e.g. if the publishing side gets upgraded and starts creating the `slot` records in a different order, and now the wallet doesn't recognize them as the same objects as before).

If those are fixed, I'm ok with this landing, but I'd want @erights to pay attention to the mixing of slot types, and I'd want some clear documentation on what shape the serialized data is expected to have.

packages/wallet/api/test/test-sharing-unserializer.js
warner: oh, this goes through the `JSON.stringify` path, so my proposed test won't help

if the "low" path only gets string IDs, then the object-identity bug won't get tripped

warner:
@dckc and @michaelfig and I just walked through this, and I think the plan is to stop using `lib-dehydrate.js` in the wallet and use a new+different serializer. This new one will use `${kind}-${distinguisher}` strings as slot values, and will *not* automatically create distinguishers during serialization. Instead, the wallet code must explicitly register all values that could be serialized (each in a separate table). Brands can be serialized as `brand-${boardID}`, inserted into the "brandIDs" table when the Issuer is registered. Purses are assigned a sequence number when registered, and are serialized as `purse-${seqnum}`, as are most other objects. The `valToSlot()` function should walk through all of these tables, return the first match it finds, and throw if it doesn't find any. The `slotToVal()` function should parse out `[ kind, distinguisher ]`, use `kind` to find the right table to consult, and then `table.get(distinguisher)` (so it throws if given an unrecognized slot, since the off-chain wallet frontend/client/UI is never exporting new objects).

dckc:
next effort is available in draft form: #5874

PR #5850, opened 2022-07-28
4652 public priceAggregator medians off-chain

### LINKED ISSUE #4652: Expose key Price Aggregator economic data via subscription

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose from the Oracle:

1. For each asset pair
    1. Aggregate price https://github.com/Agoric/agoric-sdk/pull/5850
    2. Source information (per oracle node operator) Deferred to https://github.com/Agoric/agoric-sdk/issues/5854
3. ~Governance parameters~ there are none

Output stream of priceAggregator should be the same as its priceAuthority.

turadg:
closes: #4652

## Description

For each priceAggregator instance, create an off-chain storage stream.

Whenever a quote is produced by its priceAuthority, publish it on that strream.

### Security Considerations

Publishes prices, already public. For marshalling uses readonly.

zoe README updated.

### Testing Considerations

Updated the primary `priceAggregator` test to verify the publications along with other quotes.

To test off-chain I needed a publication, but it doesn't publish on boot. So to test that I hacked in a `.publish()` and verified the value came out at the desired key. Documented in README. Advice and requests welcomed for any other testing needed.

### DIFF
--- a/packages/vats/src/lib-chainStorage.js
+++ b/packages/vats/src/lib-chainStorage.js
@@ -72,6 +72,7 @@ export function makeChainStorageRoot(toStorage, storeName, rootPath) {
   const rootNode = makeChainStorageNode(rootPath);
   return rootNode;
 }
+/** @typedef {ReturnType<typeof makeChainStorageRoot>} ChainStorageNode */
 
 /**
  * @returns {StorageNode} an object that confirms to StorageNode API but does not store anywhere.
--- a/packages/vats/src/lib-chainStorage.js
+++ b/packages/vats/src/lib-chainStorage.js
@@ -11,6 +11,19 @@ const { details: X } = assert;
 const pathSegmentPattern = /^[a-zA-Z0-9_-]{1,100}$/;
 
 /**
+ * @param {string} name
+ */
+export const sanitizePathSegment = name => {
+  const candidate = name.replace(/[- ]/g, '_');
+  assert(
+    pathSegmentPattern.test(candidate),
+    `Path sanitization failed for ${name}`,
+  );
+  return candidate;
+};
+harden(sanitizePathSegment);
+
+/**
  * Create a root storage node for a given backing function and root path.
  *
  * @param {(message: any) => any} toStorage a function for sending a storageMessage object to the storage implementation (cf. golang/cosmos/x/swingset/storage.go)
--- a/packages/zoe/src/contracts/priceAggregator.js
+++ b/packages/zoe/src/contracts/priceAggregator.js
@@ -3,7 +3,11 @@
 import { makeIssuerKit, AssetKind, AmountMath } from '@agoric/ertp';
 import { E } from '@endo/eventual-send';
 import { Far } from '@endo/marshal';
-import { makeNotifierKit } from '@agoric/notifier';
+import {
+  makeNotifierKit,
+  makeStoredPublishKit,
+  observeNotifier,
+} from '@agoric/notifier';
 import { makeLegacyMap } from '@agoric/store';
 import {
   calculateMedian,
@@ -18,6 +22,7 @@ import {
   floorMultiplyBy,
   ceilDivideBy,
   multiplyRatios,
+  ratiosSame,
 } from '../contractSupport/ratio.js';
 
 import '../../tools/types.js';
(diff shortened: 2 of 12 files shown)


packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -239,35 +256,49 @@ test('median aggregator', async t => {
...
     increment: 17n,
   });
+  await publications.nextMatches({ amountOut: 2374n, timestamp: 4n });
+  await publications.nextMatches({ amountOut: 1050n, timestamp: 4n });
 
   const quote4 = await tickAndQuote();
   t.deepEqual(quote4, { amountOut: 1060n, timestamp: 5n });
+  await publications.nextMatches(quote4);
+  await publications.nextMatches({ amountOut: 1060n, timestamp: 5n });
+  await publications.nextMatches({ amountOut: 1060n, timestamp: 5n });
```
turadg: should the aggregator skip publishing repeats?

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -239,7 +254,11 @@ const start = async (
     // Publish a new authenticated quote.
     publishedTimestamp = timestamp;
     lastPrice = median;
-    updater.updateState(authenticatedQuote);
+    console.debug(
+      'DEBUG publishing authenticatedQuote amount',
+      authenticatedQuote.quoteAmount.value,
+    );
```
Chris-Hibbert: to be dropped before merge.

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -208,57 +255,118 @@ test('median aggregator', /** @param {ExecutionContext} t */ async t => {
...
   await E(price1300Admin).delete();
+  await publications.nextMatches({
+    amountIn: 2n,
+    amountOut: 1975n,
+    timestamp: 8n,
+  });
 
   const quote8 = await tickAndQuote();
   t.deepEqual(quote8, { amountOut: 1001n, timestamp: 9n });
+  await publications.nextMatches({
```
Chris-Hibbert: 1001n (just above)  seems to be wrong.  It gets that result by combining the 9th tick from `price1000` with the 10th tick from `price800`. I don't yet see why that happens.

The result your iterator gets for tick 9 (1985/2)  is correct from my calculations. The 9th tick for `price1000` is 1100, and the 9th tick for `price800` is 885.

Chris-Hibbert: Okay, it's not a bug you created or made worse.  Would you mind adding a comment on line 361 referencing #5851?  And you should expect the publication's result to continue to be ` 1985n / 2n `.

Chris-Hibbert (approved):

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -63,9 +74,22 @@ const start = async (
     return harden({ quoteAmount, quotePayment });
   };
 
-  const { notifier, updater } = makeNotifierKit();
+  // To notify the price authority of a new median
+  const {
+    notifier: authenticatedQuoteNotifier,
+    updater: authenticatedQuoteUpdater,
```
turadg: maybe instead,
```suggestion
    notifier: medianNotifier,
    updater: medianUpdater,
```

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -168,13 +192,27 @@ const start = async (
...
       createQuote: makeCreateQuote(),
-      notifier,
+      notifier: authenticatedQuoteNotifier,
       quoteIssuer: quoteKit.issuer,
       timer,
       actualBrandIn: brandIn,
       actualBrandOut: brandOut,
     });
 
+  // for each new quote from the priceAuthority, publish it to off-chain storage
```
turadg, line 200: reference tech debt https://github.com/Agoric/agoric-sdk/issues/5853

packages/vats/src/lib-chainStorage.js
```diff
@@ -10,6 +10,19 @@ const { details: X } = assert;
 // ASCII alphanumeric plus underscore.
 const pathSegmentPattern = /^[a-zA-Z0-9_-]{1,100}$/;
 
+/**
+ * @param {string} name
+ */
+export const sanitizePathSegment = name => {
+  const candidate = name.replace(/[- ]/g, '_');
```
gibson042, line 17: @turadg I missed this at the time, but two questions:
1. What is the motivation for replacing `-` (which is a valid path segment character)?
2. Shouldn't this rather specific function be defined in a higher layer?

turadg: Let's discuss in https://github.com/Agoric/agoric-sdk/pull/5926

PR #5852, opened 2022-07-29
WIP Experiment with an oo API for reachable durability

erights:
Hi @Chris-Hibbert I'd like us to look at this when we discuss the durability masterclass tomorrow. Would this be sufficiently more teachable to be worth switching to it? I think not for MN1, but it is worth discussing.

### DIFF
--- a/packages/ERTP/src/issuerKit.js
+++ b/packages/ERTP/src/issuerKit.js
@@ -3,7 +3,7 @@
 
 import { assert } from '@agoric/assert';
 import { assertPattern } from '@agoric/store';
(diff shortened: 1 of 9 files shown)


Chris-Hibbert:
The biggest cleanups in this PR were where it switched to `vivify`, which we should do regardless. 

The rest of it didn't seem to simplify much.  There were a few places where it needed to call `getBaggage()` because we stored state directly there. If that got folded in, it might be an improvement.

I'm usually a strong proponent of OO, but I don't see much gain here.

erights:
Closing as failed experiment.

PR #5855, opened 2022-07-29
deprecate unbranded time

turadg:
## Description

Follows on https://github.com/Agoric/agoric-sdk/pull/5821 per this discussion https://github.com/Agoric/agoric-sdk/pull/5821#discussion_r930201009

Marks as `@depcreated` the value types we want to move away from.

To accomplish this I had to pull the timer types out to `.ts` syntax. Separate commits.

### Security Considerations

--

If we think this types definition pattern is worth using in other packages I'll document it as best practice. cc @mhofman 

### Testing Considerations

--


packages/SwingSet/src/vats/timer/timeMath.js
```diff
@@ -6,7 +6,6 @@ import { Nat } from '@agoric/nat';
 import { fit } from '@agoric/store';
 
 import { RelativeTimeShape, TimestampShape } from './typeGuards.js';
-import './types.js';
```
mhofman, line 9: I'm confused, how can files that transitively import `timeMath.js` now get the ambient (now global) types defined in `types.d.ts` ?

packages/SwingSet/src/vats/timer/types.js
```diff
@@ -1,198 +1 @@
...
- * array of absolute times. The result is -1, 0, or 1 indicating whether
- * the first argument is less than, equal, or greater than the second.
- *
- * @property {(rel1: RelativeTime, rel2: RelativeTime) => RankComparison} compareRel
- * Compares two relative times. This comparison function is compatible
- * with JavaScript's `Array.prototype.sort` and so can be used to sort an
- * array of relative times. The result is -1, 0, or 1 indicating whether
- * the first argument is less than, equal, or greater than the second.
- */
+// JS module for types.d.ts
```
mhofman, line 1: Don't we need a single `export {}` in this file to be backwards compatible?

mhofman (commented):
I don't suppose there was an easy way to make this type file non-ambient? Also wondering how much the deprecated not sticking was due to ambient-ness as I'm sure there are deprecated declarations in `.js` files that actually report as deprecated.

packages/SwingSet/src/vats/timer/types.js
mhofman: If the file is never actually imported, let's drop it completely?

erights (approved):
LGTM, thanks!

mhofman:
> If we think this types definition pattern is worth using in other packages I'll document it as best practice.

I'm entirely happy having some explicit type definitions in `.d.ts` files, as long as we avoid this `declare global` to simulate ambient-ness. From experience, I think we should also avoid manual `.d.ts` files for `.js` files that have behavior and are imported for their runtime.

turadg:
> I don't suppose there was an easy way to make this type file non-ambient?

Not in my exploration.

> Also wondering how much the deprecated not sticking was due to ambient-ness as I'm sure there are deprecated declarations in .js files that actually report as deprecated.

Can you point me to some? When I tried in TS sandbox I concluded that [types can be deprecated with jsdoc](https://www.typescriptlang.org/play?#code/PTAEEkDtQFgOgAygC4AsCmoBSBlAIgPYDGKAhgOagACAJugA4BO6Rpy6NoAlgM6ik06nZARQYAsACgQKAJ71MPWT3YBbOKACaBAK6hW0HT0y0GzVu06lIsgO4ZmUmbN37roI5lyESRHY2ZIZAAbWTgpKS4g9EYAM1IiTABBIiJdIKhY0QBvKVB80EhSVXQALlAVRijyAG48gvJ0SDpGcsrqusl6-OAAKl7qOiYWNg4PY1BG5pjQWK50YM4olXQBUF7gbor0AA9ygCJVUmD0fdAAH1B92PQjk-3OgF8IyToiYNJmfQJIFXGYzIEcopNI6DKQLKdTyMQFwYw7TpOMAAFXk6BwRCq9GQoFsXGCwVABFiN0Y-EKPwAtAAjYLEADW1Vxn0gTPsTX4SNAQ3Mo04TAICkYyFk3D4CUSPGMNAANPxmqAOFwRIw+MEuPT0FyAG48NJ0XH4wmeCqoAi2MSYHkjSzcCGiKKgegfSVc9Wau3sAm8JrGOW6EJRdDihWOlz+b50cJdGMyABCACUkgA5PAAUTwoGR4AAsmmpCKFFmuCUVMV6AmWARGJwALygXKSAqgamMaw0NrIKqQcgyrba446MotrjkKLIPuSZ6x-qDMw2sYbAto0BJVTU0c6XQ8ZEl4PIcugeu70sH1QVqs1i4jsdBRGvFgfL5pX444rU4Hrzfbk-78udIA) but [not with plain JS](https://www.typescriptlang.org/play?filetype=js#code/PTAECECUEEDkBECi9QBUCSBZRBYAUMAFSH6iGgACALgJ4AOApgCYMBmoA3gPYBGAVgwDGVAL5oAlgFsGAZyoBDSXUhCuAJyalyFOmq6M1tTnLXiAdgHMxVKQzXg18s5rxlKu-XaMce4i+dFQeR4ZADV5ABsAVwYtYHx8IhJXbRZdIXkqZi1KWkYWdg5UWzlFZVUNUAAfUF9-M0DoSTqoriiZYulSpTiEvETiXPoGTiaWto6ShSUAfjFCeLxBLjM5IObQAF5QMyiIiIBufCA).

> I'm entirely happy having some explicit type definitions in .d.ts files, as long as we avoid this declare global to simulate ambient-ness. 

I can't tell whether this is a request for change. What is the downside to you of using `declare global`? I can't think up any ways it's worse than the status quo.

> we should also avoid manual .d.ts files for .js files that have behavior and are imported for their runtime.

No objection to that in principle but imo it's out of scope for this PR. If you want me to include some documentation at the top of the file about what to do instead of emulating this, please suggest some.

mhofman:
> When I tried in TS sandbox I concluded that [types can be deprecated with jsdoc](https://www.typescriptlang.org/play?#code/PTAEEkDtQFgOgAygC4AsCmoBSBlAIgPYDGKAhgOagACAJugA4BO6Rpy6NoAlgM6ik06nZARQYAsACgQKAJ71MPWT3YBbOKACaBAK6hW0HT0y0GzVu06lIsgO4ZmUmbN37roI5lyESRHY2ZIZAAbWTgpKS4g9EYAM1IiTABBIiJdIKhY0QBvKVB80EhSVXQALlAVRijyAG48gvJ0SDpGcsrqusl6-OAAKl7qOiYWNg4PY1BG5pjQWK50YM4olXQBUF7gbor0AA9ygCJVUmD0fdAAH1B92PQjk-3OgF8IyToiYNJmfQJIFXGYzIEcopNI6DKQLKdTyMQFwYw7TpOMAAFXk6BwRCq9GQoFsXGCwVABFiN0Y-EKPwAtAAjYLEADW1Vxn0gTPsTX4SNAQ3Mo04TAICkYyFk3D4CUSPGMNAANPxmqAOFwRIw+MEuPT0FyAG48NJ0XH4wmeCqoAi2MSYHkjSzcCGiKKgegfSVc9Wau3sAm8JrGOW6EJRdDihWOlz+b50cJdGMyABCACUkgA5PAAUTwoGR4AAsmmpCKFFmuCUVMV6AmWARGJwALygXKSAqgamMaw0NrIKqQcgyrba446MotrjkKLIPuSZ6x-qDMw2sYbAto0BJVTU0c6XQ8ZEl4PIcugeu70sH1QVqs1i4jsdBRGvFgfL5pX444rU4Hrzfbk-78udIA) but [not with plain JS](https://www.typescriptlang.org/play?filetype=js#code/PTAECECUEEDkBECi9QBUCSBZRBYAUMAFSH6iGgACALgJ4AOApgCYMBmoA3gPYBGAVgwDGVAL5oAlgFsGAZyoBDSXUhCuAJyalyFOmq6M1tTnLXiAdgHMxVKQzXg18s5rxlKu-XaMce4i+dFQeR4ZADV5ABsAVwYtYHx8IhJXbRZdIXkqZi1KWkYWdg5UWzlFZVUNUAAfUF9-M0DoSTqoriiZYulSpTiEvETiXPoGTiaWto6ShSUAfjFCeLxBLjM5IObQAF5QMyiIiIBufCA).

I'm having a hard time matching the examples to this sentence, but I do see that deprecating a `@typedef` doesn't seem to work. That seems odd. I guess one more of the difference that `type =` and `@typedef` are not strictly equivalent.

> I can't tell whether this is a request for change.

Not a request for a change, but not something I think we should endorse unless actually needed like in this case. I believe that `declare global` usage should be very limited (in theory only to cases where importing a module actually has a side effect of creating a runtime global)

packages/SwingSet/src/vats/timer/types.js
mhofman: Please remove this file if unused (not in any other `import './types.js'`) or add an `export {};` as content.

mhofman (changes requested):

packages/SwingSet/src/vats/timer/types.js
turadg: Without this the types aren't available outside the module.

packages/SwingSet/src/vats/timer/timeMath.js
turadg: Within this package the `jsconfig.json` is picking up the global types in `types.d.ts`.

Outside this package they come along with `exported.js`. That's why https://github.com/Agoric/agoric-sdk/pull/5855#discussion_r933590277

mhofman: I just checked and apparently tsc is correctly resolving the types declared in `types.d.ts` even in packages other than `Swingset`, at least in VSCode (which may be running in a different mode than a per package command line `tsc`). I am still concerned that other packages which don't have a `*.ts` in their include will simply have these type silently default to `any`, but I don't have time to validate this one way or another. I believe a triple-slash directive would have been safer in this case.

packages/SwingSet/src/vats/timer/types.js
mhofman: I do not understand how the presence of this file helps for packages outside `SwingSet`. This feels magical to me. And I'm concerned about the file not respecting the module semantics (presence of an `import` or `export` statement) since the `.d.ts` has module semantics.

packages/SwingSet/src/vats/timer/timeMath.js
turadg: > I am still concerned that other packages which don't have a *.ts in their include will simply have these type silently default to any, but I don't have time to validate this one way or another.

Does this assuage your concern?
Screen Shot 2022-07-29 at 1 29 37 PM

> I believe a triple-slash directive would have been safer in this case.

I tried that and couldn't get it to work. Maybe could be done with more time but I struck upon this solution and I don't yet understand what's deficient in it.

PR #5858, opened 2022-07-29
simplify timer global types

turadg:
## Description

For follow up to https://github.com/Agoric/agoric-sdk/pull/5855 (merged before discussion concluded)

### DIFF
--- a/packages/SwingSet/exported.js
+++ b/packages/SwingSet/exported.js
@@ -1,1 +1,1 @@
-import './src/vats/timer/types.js';
+/// <reference types="./src/vats/timer/types"/>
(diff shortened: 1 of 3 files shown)


packages/SwingSet/exported.js
```diff
@@ -1 +1 @@
-import './src/vats/timer/types.js';
```
mhofman, line 1: Ha I had missed this `import`, now I understand why the types were pulled in.

mhofman (approved):
Sorry I had totally missed the `exported.js` file, that explains how the types were pulled in other packages. I really wouldn't worry as much if we had `noImplicitAny` turned on.

We could keep the `type.js` around with a content of `export {};` and the explicit `import './src/vats/timer/types.js';` in `exported.js` instead of the `/// <reference`, but I don't really have a preference either way.

Also the removal of the `export` inside the `declare global` is a nice cleanup,

Issue #4654, opened 2022-02-23
Account provisioning: charge BLD for smart wallet


dckc:
## What is the Problem Being Solved?

In #4402 we discuss fee aspects of ~~RUN~~ IST onboarding, but there is also a question of provisioning accounts; that is: for creating an on-chain c-list and associating it with a cosmos account (key pair). How do we appropriately recover the cost of this operation?

Our devnet account provisioning process depends on a [discord faucet bot](https://github.com/Agoric/faucet) on our discord service to mitigate griefing and sybil attacks.

Additional open questions:
 - [ ] should we "walk before we run" w.r.t. the number of `SMART_WALLET` accounts, while we gather confidence in system performance?
 - [ ] provisioning ag-solo (`REMOTE_WALLET`) accounts is higher risk. how to manage that risk?

## Description of the Design

?

note [Feb 24 proposal](https://github.com/Agoric/agoric-sdk/issues/4654#issuecomment-1050386681)

## Security Considerations

_see problem statement above_

## Test Plan

_IOU_

cc @Tartuffo @michaelfig @dtribble @warner

dckc:
@michaelfig and I think we have a design:

Scenario: User Una wants to accept (make) an offer (for example, to getRUN or to trade on the AMM or to take out a Vault). How does Una communicate with Zoe or any other object in the Agoric blockchain VM? @michaelfig and I concluded that _some_ introduction mechanism is necessary to join the Agoric society where members enjoy the benefits of offer safety etc. Una must have 1 BLD. Una uses a wallet web UI to send a "please provision" message that (1) liens gp=1 BLD and (2) creates a c-list and on-chain wallet.

Una can then proceed using `WalletAction` messages as in #4406 .

or RUN?

gp is a governable parameter

p.s. Alice...

dckc:
cost?
JS "find or create" - small. 3.
cosmos level - more expensive.

option: free if they already have BLD?
easier to game

nearby issue: how to get my friend in Agoric when my BLD is locked? how much does it take to earn 1 BLD in 1 month from staking rewards?
"sponsoring" other "users"?

Tartuffo:
@dckc I created #4723 to capture the "how to help a friend create an account" use case.  Assigned to @rowgraus , and put it in the general Product Backlog (not MN-1).

dckc:
@samsiegart writes in https://github.com/Agoric/agoric-sdk/pull/5860#pullrequestreview-1056212401 :
[...]

PR #5860, opened 2022-07-29
chore: route WalletAction, WalletSpendAction to walletFactory contract

dckc:
refs: #4398 

## Description

route WalletAction, WalletSpendAction through the `bridgeManager` to walletFactory contract

 - cosmic-swingset integration test helpers:: fund-acct, wallet-action, provision-acct Makefile rules

### Security Considerations

relies on golang to authenticate the `.owner` address before we use it to look up the relevant wallet.

?

### Testing Considerations

Tested manually, so far.

Automated testing golang+JS stuff is beyond the current state-of-the-art.

### DIFF
--- a/packages/vats/src/bridge.js
+++ b/packages/vats/src/bridge.js
@@ -18,7 +18,7 @@ import { Far } from '@endo/far';
 
 /**
  * @typedef {object} BridgeHandler An object that can receive messages from the bridge device
- * @property {(srcId: string, obj: any) => Promise<any>} fromBridge Handle an inbound message
+ * @property {(srcId: string, obj: any) => Promise<undefined>} fromBridge Handle an inbound message
  *
  * @typedef {object} BridgeManager The object to manage this bridge
  * @property {(dstID: string, obj: any) => any} toBridge
@@ -46,7 +46,7 @@ export function makeBridgeManager(E, D, bridgeDevice) {
     // );
 
     // Notify the specific handler, if there was one.
-    E(srcHandlers.get(srcID)).fromBridge(srcID, obj);
+    void E(srcHandlers.get(srcID)).fromBridge(srcID, obj);
 
     // No return value.
   }
--- a/packages/vats/src/bridge.js
+++ b/packages/vats/src/bridge.js
@@ -18,7 +18,7 @@ import { Far } from '@endo/far';
 
 /**
  * @typedef {object} BridgeHandler An object that can receive messages from the bridge device
- * @property {(srcId: string, obj: any) => Promise<undefined>} fromBridge Handle an inbound message
+ * @property {(srcId: string, obj: any) => Promise<void>} fromBridge Handle an inbound message
  *
  * @typedef {object} BridgeManager The object to manage this bridge
  * @property {(dstID: string, obj: any) => any} toBridge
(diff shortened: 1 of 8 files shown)


packages/vats/src/bridge.js
```diff
@@ -46,7 +46,7 @@ export function makeBridgeManager(E, D, bridgeDevice) {
     // );
 
     // Notify the specific handler, if there was one.
-    E(srcHandlers.get(srcID)).fromBridge(srcID, obj);
+    void E(srcHandlers.get(srcID)).fromBridge(srcID, obj);
```
turadg, line 49: 👍

packages/wallet/contract/src/walletFactory.js
```diff
@@ -10,22 +10,32 @@ import '@agoric/zoe/exported.js';
...
 
 /**
  * @typedef {{
  *   agoricNames: NameHub,
  *   board: Board,
  *   namesByAddress: NameHub,
  * }} SmartWalletContractTerms
+ *
+ * @typedef {{
+ * 	 type: string, // WALLET_ACTION
```
turadg: is the type `ActionType`? 
```suggestion
 * type: keyof ActionType
```

packages/wallet/contract/src/walletFactory.js
```diff
@@ -34,12 +44,32 @@ export const start = async (zcf, privateArgs) => {
...
+     * @param {WalletAction} obj
+     */
+    fromBridge: async (srcID, obj) => {
+      console.log('walletFactory.fromBridge:', srcID, obj);
+      assert(obj, 'missing wallet action');
+      assert.typeof(obj, 'object');
+      assert.typeof(obj.owner, 'string');
+      const wallet = walletsByAddress.get(obj.owner); // or throw
+      console.log('walletFactory:', { wallet });
+      return E(wallet).performAction(obj); // TODO: add performAction to lib-wallet.js
```
turadg: I'm not sure what value `fromBridge` has when `performAction` doesn't exist but I'll trust you do :)

turadg (approved):
No blocking concerns.

packages/wallet/contract/src/walletFactory.js
```diff
@@ -34,12 +44,32 @@ export const start = async (zcf, privateArgs) => {
...
   const walletsByAddress = makeScalarBigMapStore('walletsByAddress');
   const provider = makeAtomicProvider(walletsByAddress);
 
+  const handleWalletAction = Far('walletActionHandler', {
+    /**
+     *
+     * @param {string} srcID
+     * @param {WalletAction} obj
+     */
+    fromBridge: async (srcID, obj) => {
```
samsiegart, line 66: Just thinking out loud, we will at some point want `provideSmartWallet` to be callable in this same way so a user can create their wallet from the browser.

samsiegart (approved):

dckc:
converting back to draft... I think this only does `WALLET_ACTION` and not `WALLET_SPEND_ACTION`.

dckc:
another reason to convert back to draft: we have plans to migrate from `WalletAction` to something else: https://github.com/Agoric/agoric-sdk/issues/4398#issuecomment-1200531537

packages/wallet/contract/src/walletFactory.js
dckc: yes, but `ActionType` is written as a module namespace, not a record, so `keyof` doesn't work.

I have a refinement pending.

packages/wallet/contract/src/walletFactory.js
dckc: This didn't get past `yarn lint` so made the whole like a TODO comment.

packages/wallet/contract/src/walletFactory.js
dckc: For follow-up, see notes in #4654

Issue #5791, opened 2022-07-20
[story] Oracle operators can send updates to PriceAuthority running on chain


dtribble:
Chainlink operators provide the Agoric price feeds. We want to make that as easy as possible. (Idiomatic for Chainlink and for Agoric.) We need a way to wire them together.

## What is the Problem Being Solved?

From @jacquesvcritien of the oracle team:
> The old setup consisted of an ag-solo, an [external initiator](https://github.com/smartcontractkit/external-initiator), a Chainlink node an [external adapter](https://github.com/thodges-gh/CL-EA-NodeJS-Template). The ag-solo would have the logic running on it, holds state and it would initiate new jobs by calling the external initiator which triggers a job on the Chainlink node. The result of this job is sent to the external adapter which replies back to the ag-solo to push on chain. The issue with the ag-solo was that its state was not recoverable and we saw it breaking down too often, causing node operators to constantly having to reset their setups and recreate oracles. 
>
> Due to this, Michael had suggested that I build a middleware (Node.JS script which runs with agoric run) which replaces the logic on the ag-solo, the external initiator and the external adapter. This would keep the state of the last runs and prices, submit jobs every minute to the chainlink node, receives back the result from the chainlink node and submits the result on chain if there was a change in price which exceeds a certain threshold. This middleware is ready, however, Michael had also told me that in order for this middleware to communicate with the chain directly, there was some additional software which was being built by himself/Agoric team. I do not know the exact details about the status and the specific details of this missing piece. However, if I am not mistaken, I remember Michael mentioning that it had to do with `cosmJS` if that helps. Therefore, the final missing piece is a piece of code which allows this middleware to be able to submit the price to a price authority.

This is critical for future oracles joining the network.

## Description of the Design

TBD how to submit a price to the price authority.

Depends on Agoric run scripts having access to the cache in the smart wallet. We need the caching layer to behave properly with off-chain data. Currently it's a transaction to read the cache for a given value. We need the state of the cache to be exposed as a chain storage node. (Cache needs to expose its data off-chain.)  https://github.com/Agoric/agoric-sdk/issues/5868

Design requirements:
- price info transactions are different from other Cosmos transactions (like on-chain wallets, other routes into the chain than the ag-solo)
- fit the smart wallet transactions model (where we already have solutions for signing and legibility)
  - contain collection of arbitrary objects (ocaps) you can send messages to

1. Get a reference to the smart wallet for the given casting client
[...]

PR #5871, opened 2022-08-01
5868 ChainStorageCoordinator

### LINKED ISSUE #5868: make smart wallet's cache available off-chain
Blocker for https://github.com/Agoric/agoric-sdk/issues/5791 . FYI @jacquesvcritien

We need the caching layer to behave properly with off-chain data. Currently it's a transaction to read the cache for a given value. 

## Description of the Design

Expose the state of the cache to a chain storage node so it can be read off-chain.

The cache has pluggable backends to update its underlying data so provide the chain storage node and logic there.

Interfaces
- Smart Wallet UI
  - use Caching/Casting mechanism to store what it cares about
  - should be incidental in solving the harder case of run scripts:
- `agoric run` scripts
  - particularly interesting because the oracle is implemented by a a run script (step towards giving a run script full access to an on-chain wallet, so the oracle can write price changes to the cache and they can be read anywhere and not depend on an ag-solo) 

Intended design (depends on https://github.com/Agoric/agoric-sdk/issues/5912):
- Client says to coordinator “update cache with this value”.
- Coordinator checks its local copy for the value.
- If the generation looks good just return the value.
- If not, update the cache store and ask the leader to publish it.
- Leader says “okay got it” and the data goes out in the streams.

**Acceptance criteria**
- a ChainStorageCoordinator that takes storageNode and marshaller
- it writes to the storageNode such that it can be read by a casting client
- client methods that can read and write to the cache backed by storageNode
  - read by asking storageNode for current state (quick b/c rpc)
  - write by submitting a chain transaction
- a CastingClientCoordinator that takes a [leader, casting spec, marshaller], and makes it look like a cache (for use in deploy scripts that can't get an ocap for wallet admin)
- casting spec TBD for how specify wallet and key material… but that general solution can come after E2E.
- object graphs returned by the deploy script match those of the cache (so they need the same marshaller)
- casting client has fast reads and slow writes, constructed with a smart wallet address

## Security Considerations

Cache is already publicly readable as part of the on-chain wallet.

Have to be careful of marshalling the objects that are mixed public/private. Shouldn't publish ocaps to the board. We'll need the wallet marshaller (https://github.com/Agoric/agoric-sdk/pull/5849/, https://github.com/Agoric/agoric-sdk/pull/5874, https://github.com/Agoric/agoric-sdk/pull/5883 )

turadg:
advances: #5868

## Description

Feature: `ChainStorageCoordinator` as described in #5868

### Security Considerations

Data meant to be read off-chain.

--

### Testing Considerations

New tests

### DIFF
--- a/packages/cache/src/store.js
+++ b/packages/cache/src/store.js
@@ -5,7 +5,7 @@ import { matches, makeScalarMapStore } from '@agoric/store';
 
 import '@agoric/store/exported.js';
 
-import { GROUND_STATE, makeState } from './state.js';
+import { withGroundState, makeState } from './state.js';
 
 /**
  * @param {(obj: unknown) => unknown} [sanitize]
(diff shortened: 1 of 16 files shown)


packages/wallet/api/src/lib-wallet.js
```diff
@@ -957,7 +962,7 @@ export function makeWallet({
           origin === null ||
           (offer.requestContext && offer.requestContext.dappOrigin === origin),
       )
-      .sort(([id1], [id2]) => cmp(id1, id2))
+      .sort(([id1], [id2]) => id1.localeCompare(id2))
```
dckc: Is `localeCompare` sufficiently deterministic? Doe it depend on an ambient locale setting? Looks like it's OK...

## Details

It took me a while to confirm, so I might as well save my notes:

> Much of the `Intl` package, and some locale-specific portion of other objects (e.g. Number.prototype.toLocaleString) has results which depend upon which locale is configured, and that will vary from one process to another. The way we handle this is still in development. Either these functions will be disabled, or they will act as if they run on a host with a single fixed locale defined by the SES specification.
-- https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/docs/vat-environment.md

I also did a quick experiment to confirm that XS supports it by adding this to `packages/xsnap/test/test-xs-js.js`.

```js
test('String.prototype.localeCompare', async t => {
  const opts = options(io);
  const vat = xsnap(opts);
  await vat.evaluate(`
  const encoder = new TextEncoder();
  const send = it => issueCommand(encoder.encode(JSON.stringify(it)).buffer);
  send('abc'.localeCompare('def') < 0)
  `);
  t.deepEqual(opts.messages, ['true']);
  await vat.terminate();
});
```

turadg: Thanks. Added in https://github.com/Agoric/agoric-sdk/pull/5873

dckc (commented):
I browsed around a bit, but I'm struggling to figure out which end is up.

IIUC, this is needed for Oracle stuff. A unit test that sketched how it fits would be nifty. (Or maybe it's there and I just didn't read carefully enough?)

turadg:
> I browsed around a bit, but I'm struggling to figure out which end is up.

These are the essential lines in #5868 this advances:
> * Acceptance criteria* 
> - a ChainStorageCoordinator that takes storageNode and marshaller
> - it writes to the storageNode such that it can be read by a casting client

The rest of #5868 is in progress, stacked on this.

> IIUC, this is needed for Oracle stuff. 

Yes, though this is a couple degrees removed from https://github.com/Agoric/agoric-sdk/issues/5791. 

> A unit test that sketched how it fits would be nifty. (Or maybe it's there and I just didn't read carefully enough?)

Dunno if you saw [test-storage.js](https://github.com/Agoric/agoric-sdk/pull/5871/files#diff-4ae517b94e9392d11f6b8b73bf489a5b15c782db19c28f0eedbba4e95cd346c3). That demonstrates how this new coordinator is supposed to work. Supporting oracles is days away.

packages/cache/src/store.js
```diff
@@ -33,100 +34,202 @@ const makeKeyToString = (sanitize = obj => obj) => {
...
+
+/**
+ * @param {MapStore<string, import('./types').State>} stateStore
+ * @param {Marshaller} marshaller
+ * @returns {Promise<string>}
+ */
+const stringifyStateStore = async (stateStore, marshaller) => {
+  const obj = {};
+  for (const [key, value] of stateStore.entries()) {
+    obj[key] = E(marshaller).serialize(harden(value));
```
gibson042: `harden` mutates its argument if not already hardened; is that appropriate here?

packages/cache/test/test-storage.js
```diff
@@ -0,0 +1,175 @@
...
+
+  // Initial
+  t.is(await cache('counter', increment), 1);
+  t.deepEqual(Object.keys(storageNodeState), ['cache']);
+  t.deepEqual(JSON.parse(storageNodeState.cache), {
+    '{"body":"\\"counter\\"","slots":[]}': {
+      body: '{"generation":{"@qclass":"bigint","digits":"1"},"value":1}',
+      slots: [],
+    },
+  });
```
gibson042, line 124: I find this overloaded behavior (in which the second argument to `cache` is sometimes interpreted as a passable value to set and sometimes as an updater) to be surprising, and it does not appear to be documented in the package README. What makes it a superior pattern to more predictable alternatives?

packages/cache/test/test-storage.js
```diff
@@ -0,0 +1,175 @@
...
+
+  // Again without a pattern
+  t.is(await cache('counter', increment), 1);
+  t.deepEqual(Object.keys(storageNodeState), ['cache']);
+  t.deepEqual(JSON.parse(storageNodeState.cache), {
+    '{"body":"\\"counter\\"","slots":[]}': {
+      body: '{"generation":{"@qclass":"bigint","digits":"1"},"value":1}',
+      slots: [],
+    },
+  });
```
gibson042, line 134: So "without a pattern" is the same as "with a non-matching pattern"?

packages/cache/src/store.js
dckc: args to `E(x).m(...)` get hardened whether you like it or not.
(https://github.com/Agoric/documentation/issues/679 is my attempts to find or create authoritative sources of this lore.)

packages/cache/test/test-storage.js
turadg: > does not appear to be documented in the package README

I believe it is: https://github.com/Agoric/agoric-sdk/blob/6f72d00dddf3467fca0b70e1455e19f5d87adea8/packages/cache/README.md?plain=1#L26-L32

It's also documented in the type:
https://github.com/Agoric/agoric-sdk/blob/6f72d00dddf3467fca0b70e1455e19f5d87adea8/packages/cache/src/cache.js#L23-L27

https://github.com/Agoric/agoric-sdk/blob/6f72d00dddf3467fca0b70e1455e19f5d87adea8/packages/cache/src/cache.js#L9-L12

> What makes it a superior pattern to more predictable alternatives?

I'm not sure it is. Changing it is out of scope of this PR but if you want to a ticket would be welcomed.

packages/cache/test/test-storage.js
turadg: Practically, yes. More precisely if you don't set the pattern then it's `undefined`, which will only match if the extant value is `undefined`. That will be the state if a key was set to `undefined` or not yet set.

packages/cache/src/store.js
gibson042:
```suggestion
    obj[key] = E(marshaller).serialize(value);
```

packages/cache/test/test-storage.js
gibson042: > > does not appear to be documented in the package README
> 
> I believe it is:
> 
> https://github.com/Agoric/agoric-sdk/blob/6f72d00dddf3467fca0b70e1455e19f5d87adea8/packages/cache/README.md?plain=1#L26-L32

Thank you. I think I was looking specifically at https://github.com/Agoric/agoric-sdk/tree/6f72d00dddf3467fca0b70e1455e19f5d87adea8/packages/cache#cache-client , which documents only `cache(key)`, `cache(key, (oldValue) => ERef<newValue>)`, and `cache(key, (oldValue) => ERef<newValue>, guardPattern)`, with no mention of `cache(key, value [, guardPattern])`.

> > What makes it a superior pattern to more predictable alternatives?
> 
> I'm not sure it is. Changing it is out of scope of this PR but if you want to a ticket would be welcomed.

Excellent point; created #5923.

packages/cache/test/test-storage.js
gibson042: Ah, so is that another gap in the [README](https://github.com/Agoric/agoric-sdk/tree/6f72d00dddf3467fca0b70e1455e19f5d87adea8/packages/cache#cache-client), or is it a misimplementation of intent? [`23b94b1ebd176a7f7ae90ca64d2252ed348e8836`](https://github.com/Agoric/agoric-sdk/commit/23b94b1ebd176a7f7ae90ca64d2252ed348e8836#diff-9875c3e44f152d364645383f413b44436bf5ac1fe0c5c23e211035a6fe114eb8) inverted the behavior from its previous `guardPattern = optGuardPattern.length > 0 ? optGuardPattern[0] : M.any()`, and that looks accidental to me. Maybe @michaelfig can shed some light on this?

packages/cache/src/store.js
turadg: done (in fixup and rebase)

packages/cache/test/test-storage.js
michaelfig: > inverted the behavior... and that looks accidental to me

The original behaviour was the accident.  It makes more sense for the guard pattern to be a match on the existing value, and for the default to actually be the pattern `undefined` to match the value `undefined`, since most of the time callers want to read the current value from the cache if possible without submitting an update.

gibson042: Great, let's update the README in this PR. Suggestions:
```patch
 - the ground state for a cache key value is `undefined`.  It is impossible to distinguish a set value of `undefined` from an unset key
 - `cache(key): Promise<recentValue>` - get an eventually-consistent value for `key`
-- `cache(key, (oldValue) => ERef<newValue>): Promise<newValue>` -  call the
-  updater function transactionally with the current value of `key`, and update
-  it to `newValue`.  Rerun the updater function with a new value if the
-  transaction is stale.
-- `cache(key, (oldValue) => ERef<newValue>, guardPattern): Promise<newValue>` -
-  same as above, but only update the cache if guardPattern matches the current
-  value.  Return the updated value, or the value that matches guardPattern.
+- `cache(key, (oldValue) => ERef<newValue>, guardPattern?): Promise<newValue>` -
+  transactionally updates the value of `key` from a value matched by `guardPattern`
+  to the results of sanitizing the result of calling the updater function with
+  that value. Retries all three steps (read current value, attempt match, call updater),
+  if the transaction is stale. `guardPattern` defaults to matching only `undefined`.
+  Returns the current value after any resulting update.
+- `cache(key, passable, guardPattern?): Promise<newValue>` -
+  same as above with an updater function that ignores arguments and returns
+  `passable` (e.g., `() => passable`).
```

And noticed in passing:
```patch
 ```js
-import { makeCache, makeStoreCoordinator } from '@agoric/cache';
+import { makeCache, makeScalarStoreCoordinator } from '@agoric/cache';
 import { M } from '@agoric/store';
 import { makeScalarBigMapStore } from '@agoric/vat-data';
 
 const store = makeScalarBigMapStore('cache');
-const coordinator = makeStoreCoordinator(store);
+const coordinator = makeScalarStoreCoordinator(store);
 const cache = makeCache(coordinator);
```

turadg: Added in d03f2df145e5b32b5aae6e77ef74c38c2ad7d7fd

dckc (approved):
I have no critical objections.

I haven't reached the "I'm happy to maintain this code" level of review, but it pretty clearly involves no regressions, so let's see how it goes as we get more of the Oracle stuff in place.

PR #5873, opened 2022-08-01
smart wallet refactorings

turadg:
## Description

While working on https://github.com/Agoric/agoric-sdk/pull/5871 these refactorings seemed helpful. Whatever I can land here can reduce the review load of the other.

### Security Considerations

--

--

### Testing Considerations

--

### DIFF
--- a/packages/wallet/contract/src/smartWallet.js
+++ b/packages/wallet/contract/src/smartWallet.js
@@ -1,11 +1,11 @@
 // @ts-check
-import { E, Far } from '@endo/far';
 import {
-  makeStoredSubscription,
-  makeSubscriptionKit,
+  makePublishKit,
+  makeStoredSubscriber,
   observeIteration,
 } from '@agoric/notifier';
 import spawn from '@agoric/wallet-backend/src/wallet.js';
+import { E, Far } from '@endo/far';
 
 const { assign, entries, keys, fromEntries } = Object;
 
@@ -19,7 +19,7 @@ const { assign, entries, keys, fromEntries } = Object;
  * agoricNames: NameHub,
  * board: Board
  * namesByAddress: NameHub,
- * storageNode?: ERef<StorageNode>,
+ * storageNode: ERef<StorageNode>,
  * zoe: ERef<ZoeService>,
  * }} shared
  */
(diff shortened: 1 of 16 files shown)


dckc (approved):
good stuff

turadg:
@dckc I had to make some more edits to pass CI. PTAAL. 

I rebased to keep the commits clean so I'll point out the changes:
- added your test from https://github.com/Agoric/agoric-sdk/pull/5871#discussion_r935017871 [47d2f1d](https://github.com/Agoric/agoric-sdk/pull/5873/commits/47d2f1dd7e5f17ee8c5378316e4a38d0c6166a03)
  - I made you coauthor of this and the rebase applied that to more. doesn't seem worth fixing but wanted to point out why.
- made storageNode requiredin [d59016a](https://github.com/Agoric/agoric-sdk/pull/5873/commits/d59016a521b835a7fbefc489ca41a26367d761c0)
- fixed the cmp refactoring with a new `cmpProperty` function in [36d265e](https://github.com/Agoric/agoric-sdk/pull/5873/commits/36d265e2a364192298b529cb49b1c0a465ed8028)
- [bumped maxNodeModuleJsDepth to pass tsc](https://github.com/Agoric/agoric-sdk/pull/5873/commits/34c32fc46256483477b51696849338c4a70ec9f8)

packages/wallet/contract/src/smartWallet.js
```diff
@@ -19,7 +19,7 @@ const { assign, entries, keys, fromEntries } = Object;
  * agoricNames: NameHub,
  * board: Board
  * namesByAddress: NameHub,
- * storageNode?: ERef<StorageNode>,
+ * storageNode: ERef<StorageNode>,
```
dckc, line 22: remind me how this works in the sim chain? I don't see how it can work if the root `chainStorage` is resolved to `null`:

https://github.com/Agoric/agoric-sdk/blob/16ef8776e28ea334fe037c928b531e4be5b54ab8/packages/vats/src/core/chain-behaviors.js#L296-L305

dckc: maybe it just doesn't?

turadg: Yeah, `chainStorage` resolves to null . Then this helper accommodates that: https://github.com/Agoric/agoric-sdk/blob/15720474802c5630210184602bcc29811c0a0a21/packages/vats/src/lib-chainStorage.js#L85-L97

dckc: very well.

PR #5874, opened 2022-08-02
feat(wallet/api): marshal contexts for wallet, board

dckc:
refs: #4398

cc @samsiegart @turadg @warner @michaelfig 

## Description

marshal contexts for:
 - on-chain smart wallet
 - off-chain wallet ui
   - from AMM etc. (board)
   - from private wallet state

not yet integrated with either walletFactory contract nor wallet ui. Should that be in this PR or a separate one?

### Security Considerations

It's essential that the AMM can't refer to private objects such as purses.

?

### Testing Considerations

Unit tests, so far.

### DIFF
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -40,6 +40,7 @@ import '@agoric/inter-protocol/exported.js';
 
 import './internal-types.js';
 import './types.js';
+import { makeExportContext } from './marshal-contexts.js';
 
 // does nothing
 const noActionStateChangeHandler = _newState => {};
@@ -113,9 +114,10 @@ export function makeWallet({
     return { ...record, meta };
   };
 
+  const context = makeExportContext();
   // Create the petname maps so we can dehydrate information sent to
   // the frontend.
-  const { makeMapping, hydrate, dehydrate, edgeMapping } = makeDehydrator();
+  const { makeMapping, dehydrate, edgeMapping } = makeDehydrator();
   /** @type {Mapping<Purse>} */
   const purseMapping = makeMapping('purse');
   /** @type {Mapping<Brand>} */
@@ -135,18 +137,26 @@ export function makeWallet({
   const issuerToBoardId = makeScalarWeakMap('issuer');
 
   // Idempotently initialize the issuer's synchronous boardId mapping.
-  const initIssuerToBoardId = async issuer => {
+  const initIssuerToBoardId = async (issuer, brand) => {
     if (issuerToBoardId.has(issuer)) {
       // We already have a mapping for this issuer.
       return issuerToBoardId.get(issuer);
     }
 
     // This is an interleaving point.
-    const issuerBoardId = await E(board).getId(issuer);
+    const [issuerBoardId, brandBoardId] = await Promise.all([
+      E(board).getId(issuer),
+      E(board).getId(brand),
+    ]);
     if (issuerToBoardId.has(issuer)) {
       // Somebody else won the race to .init.
       return issuerToBoardId.get(issuer);
     }
+
+    // The board id would already be registered if the issuer was added through
+    // suggestIssuer.
+    context.ensureBoardId(issuerBoardId, issuer);
+    context.initBoardId(brandBoardId, brand);
 
     // We won the race, so .init ourselves.
     issuerToBoardId.init(issuer, issuerBoardId);
@@ -785,7 +795,7 @@ export function makeWallet({
       ? brandTable.getByIssuer(issuer)
       : brandTable.initIssuer(issuer, addMeta);
     const { brand } = await recP;
-    await initIssuerToBoardId(issuer);
+    await initIssuerToBoardId(issuer, brand);
     const addBrandPetname = () => {
       let p;
       const already = brandMapping.valToPetname.has(brand);
@@ -806,7 +816,7 @@ export function makeWallet({
 
   const publishIssuer = async brand => {
     const { issuer } = brandTable.getByBrand(brand);
-    const issuerBoardId = await initIssuerToBoardId(issuer);
+    const issuerBoardId = await initIssuerToBoardId(issuer, brand);
     updateAllIssuersState();
     return issuerBoardId;
   };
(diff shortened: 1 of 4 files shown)


samsiegart:
@dckc There's still some TODOs in the code, are those blocking?

dckc:
> There's still some TODOs in the code, are those blocking?

Yes, I think we need to understand whether we really need 6 different tables on each side. We're currently getting away with fewer; why don't we see failures? Could we do with fewer? Perhaps just 2?

Also... even the 6 tables don't cover installations nor instances.

samsiegart:
Okay I just removed them for now. If there's a reason for not using `sharedData` for everything that's on the board we can change it later, but I'd prefer not to block progress towards the demo any more at the moment.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,151 @@
...
+  const severed = `SEVERED: ${iface.replace(/^Alleged: /, '')}`;
+  const thing = /** @type {any} */ (Far(severed, {}));
+  return thing;
+};
+
+/**
+ * Make context for exporting wallet data where brands etc. can be recognized by boardId.
+ *
+ * All purses, brands, etc. must be registered before serialization / unserialization.
+ * Neither the serializer nor the unserializer creates new presences.
```
dckc, line 17: This is no longer quite the case. The serializer creates `unknown:${nonce}` slots because we don't pre-register things like the instances in zoe invitation amounts.

packages/wallet/api/src/lib-wallet.js
```diff
@@ -135,19 +137,27 @@ export function makeWallet({
...
+      E(board).getId(issuer),
+      E(board).getId(brand),
+    ]);
     if (issuerToBoardId.has(issuer)) {
       // Somebody else won the race to .init.
       return issuerToBoardId.get(issuer);
     }
 
+    // The board id would already be registered if the issuer was added through
+    // suggestIssuer.
```
turadg, line 157: something like,
```suggestion
    // Don't init in case the board was already registered in suggestIssuer
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,151 @@
...
+  };
+  /** @type {MapStore<unknown, string>} */
+  const sharedData = makeScalarMap();
+
+  /**
+   * @param {string} slot
+   * @param {string} _iface
+   */
+  const slotToVal = (slot, _iface) => {
+    const kind = Object.keys(toVal).find(k => slot.startsWith(k));
```
turadg, line 36: no need to bind a delegating function
```suggestion
    const kind = Object.keys(toVal).find(slot.startsWith);
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,151 @@
...
+
+      sharedData.init(slot, it);
+      return it;
+    },
+
+    /**
+     * @param {string} slot
+     * @param {string} iface
+     */
+    fromPart: (slot, iface) => {
```
turadg, line 117: I get `fromBoard` but what's a "Part"?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,151 @@
...
+  // constraint: none of these keys has another as a prefix.
+  // and none overlaps with board
+  const myData = {
+    /** @type {MapStore<string, Purse>} */
+    purse: makeScalarMap(),
+  };
+  /** @type {MapStore<string, unknown>} */
+  const sharedData = makeScalarMap();
+
+  const kindOf = slot => Object.keys(myData).find(k => slot.startsWith(k));
```
turadg, line 92: ditto

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,151 @@
...
+    ...makeMarshal(valToSlot, slotToVal),
+  });
+};
+
+/**
+ * Make context for unserializing wallet or board data.
+ */
+export const makeImportContext = () => {
+  // constraint: none of these keys has another as a prefix.
+  // and none overlaps with board
```
turadg, line 84: how to enforce this or detect when it's violated?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,151 @@
...
+    }),
+  };
+
+  return harden({
+    fromWallet: Far('wallet unserializer', {
+      unserialize: marshal.fromPart.unserialize,
+    }),
+    fromBoard: Far('unserializer from board', {
+      /**
+       * @throws on an attempt to refer to un-published wallet objects.
```
turadg, line 146: good comment

turadg (commented):
Some questions and concerns but nothing blocking.

I won't approve until @warner does.

packages/wallet/api/src/marshal-contexts.js
dckc: I don't think that works.  `startsWith` expects `this` to be bound to the receiver (`slot`):

```js
> ["purse", "brand"].find("brand:1".startsWith)
VM400:1 Uncaught TypeError: String.prototype.startsWith called on null or undefined
```

It works with our objects-as-closures where we pretend `this` isn't part of the language :)

packages/wallet/api/src/marshal-contexts.js
dckc: lack of inspiration.

at one point, I called it a partitioning marshaler cuz of the separation into purses, brands, contacts, etc.

maybe `fromPrivate` or `fromMyWallet` or...?

packages/wallet/api/src/marshal-contexts.js
dckc: by inspection. A unit test is probably cost-effective.

packages/wallet/api/src/marshal-contexts.js
turadg: ah, thanks for explaining

packages/wallet/api/src/marshal-contexts.js
turadg: `fromMyWallet` seems most informative. it has to come from "my wallet" or "the board"?

dckc:
re-thinking... see #5883

PR #5875, opened 2022-08-02
fix(vat-data)!: diagnose provide collisions

erights:
@Chris-Hibbert , if this is green under CI, what do you think of the invariant that this check enforces? By doing it here in `provide` is it inherited by all our other `provide*` and `vivify*` abstractions? (I think so) Should we make that invariant part of the spec of all the `provide*` and `vivify*` abstractions? (aside from `provideAsync` I think)

@turadg @samsiegart @Chris-Hibbert @FUDCo @warner @mhofman , once this is ready, I will ask you to review. Until then, please ignore. Though I'm keeping @turadg and @samsiegart listed since they've already commented.

### DIFF
--- a/packages/vat-data/src/environment-options.js
+++ b/packages/vat-data/src/environment-options.js
@@ -1,0 +1,135 @@
+// @ts-check
+
+// NOTE This module is essentially a duplicate of a module by the same name
+// in the endo repository. It is copied here *temporarily* until the one in
+// endo exports bindings we can import.
+// The prelude below --- the import and initial const declarations, are
+// stated differently that the original because of the change of context,
+// but the rest should remain identical.
+// TODO make the environment-options.js in endo one that can be imported from,
+// and then delete this duplicate.
+
+import { assert } from '@agoric/assert';
+
+const { details: X, quote: q } = assert;
+const { freeze } = Object;
+const arrayPushMethod = Array.prototype.push;
+const arrayPush = (array, thing) =>
+  Reflect.apply(arrayPushMethod, array, [thing]);
+
+/**
+ * JavaScript module semantics resists attempts to parameterize a module's
+ * initialization behavior. A module initializes in order according to
+ * the path by which it is first imported, and then the initialized module
+ * is reused by all the other times it is imported. Compartments give us
+ * the opportunity to bind the same import name to different imported
+ * modules, depending on the package/compartment doing the import. Compartments
+ * also address the difficulty of parameterizing a module's initialization
+ * logic, but not in a pleasant manner.
+ *
+ * A pleasant parameterization would be for a static module to be function-like
+ * with explicit parameters, and for the parameterization to be like
+ * calling the static module with parameters in order to derive from it a
+ * module instance. Compartments instead lets us parameterize the meaning
+ * of a module instance derived from a static module according to the
+ * three namespaces provided by the JavaScript semantics, affecting the
+ * meaning of a module instance.
+ *    * The global variable namespaces.
+ *       * The global scope, aliased to properties of the global object.
+ *         This is necessarily compartment-wide, and therefore in our
+ *         recommened usage pattern, package-wide.
+ *       * The global lexical scope. The SES-shim compartments support
+ *         these both compartment-wide as well as per-module. But it is
+ *         not yet clear what we will propose in the Compartment proposal.
+ *    * The import namespace.
+ *    * The host hooks.
+ *
+ * This `environment-options.js` module looks for a setting of an
+ * `optionName` parameter rooted in the global scope. If follows the Node
+ * precedent for finding Unix environment variable settings, looking for a
+ * global `process` object holding an `env` object,
+ * optionally holding a property named for the `optionName` whose value is the
+ * configuration setting of that option. For example, for the optionName
+ * `FOO_BAR` it would look in
+ * `globalThis.process.env.FOO_BAR`.
+ *
+ * If setting is either absent or `undefined`, that indicates that
+ * this configuration option should have its default behavior, whatever that is.
+ * Otherwise, reflecting Unix environment variables, the setting must be a
+ * string. This also helps ensure that this channel is used only to pass data,
+ * not authority beyond the ability to read this global state.
+ */
+
+/**
+ * makeEnvironmentCaptor provides a mechanism for getting environment
+ * variables, if they are needed, and a way to catalog the names of all
+ * the environment variables that were captured.
+ *
+ * @param {object} aGlobal
+ */
+export const makeEnvironmentCaptor = aGlobal => {
+  const capturedEnvironmentOptionNames = [];
+
+  /**
+   * Gets an environment option by name and returns the option value or the
+   * given default.
+   *
+   * @param {string} optionName
+   * @param {string} defaultSetting
+   * @returns {string}
+   */
+  const getEnvironmentOption = (optionName, defaultSetting) => {
+    // eslint-disable-next-line @endo/no-polymorphic-call
+    assert.typeof(
+      optionName,
+      'string',
+      X`Environment option name ${q(optionName)} must be a string.`,
+    );
+    // eslint-disable-next-line @endo/no-polymorphic-call
+    assert.typeof(
+      defaultSetting,
+      'string',
+      X`Environment option default setting ${q(
+        defaultSetting,
+      )} must be a string.`,
+    );
+
+    /** @type {string} */
+    let setting = defaultSetting;
+    const globalProcess = aGlobal.process;
+    if (globalProcess && typeof globalProcess === 'object') {
+      const globalEnv = globalProcess.env;
+      if (globalEnv && typeof globalEnv === 'object') {
+        if (optionName in globalEnv) {
+          arrayPush(capturedEnvironmentOptionNames, optionName);
+          const optionValue = globalEnv[optionName];
+          // eslint-disable-next-line @endo/no-polymorphic-call
+          assert.typeof(
+            optionValue,
+            'string',
+            X`Environment option named ${q(
+              optionName,
+            )}, if present, must have a corresponding string value, got ${q(
+              optionValue,
+            )}`,
+          );
+          setting = optionValue;
+        }
+      }
+    }
+    assert(
+      setting === undefined || typeof setting === 'string',
+      X`Environment option value ${q(setting)}, if present, must be a string.`,
+    );
+    return setting;
+  };
+  freeze(getEnvironmentOption);
+
+  const getCapturedEnvironmentOptionNames = () => {
+    return freeze([...capturedEnvironmentOptionNames]);
+  };
+  freeze(getCapturedEnvironmentOptionNames);
+
+  return freeze({ getEnvironmentOption, getCapturedEnvironmentOptionNames });
+};
+freeze(makeEnvironmentCaptor);
--- a/packages/vat-data/src/vat-data-bindings.js
+++ b/packages/vat-data/src/vat-data-bindings.js
@@ -24,8 +24,13 @@ export {
 };
 
 const { getEnvironmentOption } = makeEnvironmentCaptor(globalThis);
+// TODO DANGER HAZARD DO NOT MERGE UNTIL FIXED
+// Since this is not on master yet or even ready-for-review, and we don't know
+// when it will be, by temporarily flipping the default here to `'true'`,
+// I can get CI to run our entire system under this constraint to see what
+// breaks.
 const DETECT_PROVIDE_COLLISIONS =
-  getEnvironmentOption('DETECT_PROVIDE_COLLISIONS', 'false') === 'true';
+  getEnvironmentOption('DETECT_PROVIDE_COLLISIONS', 'true') === 'true';
 
 const { details: X, quote: q } = assert;
 
--- a/packages/vat-data/src/vat-data-bindings.js
+++ b/packages/vat-data/src/vat-data-bindings.js
@@ -1,7 +1,33 @@
 /* global globalThis */
 
-import { Fail } from '@agoric/assert';
-import { provideLazy } from '@agoric/store';
+import { Fail, assert } from '@agoric/assert';
+import {
+  M,
+  makeScalarMapStore,
+  makeScalarWeakMapStore,
+  makeScalarSetStore,
+  makeScalarWeakSetStore,
+  provideLazy,
+} from '@agoric/store';
+
+// TODO import from proper place. See note in that module.
+import { makeEnvironmentCaptor } from './environment-options.js';
+
+/** @typedef {import('./types.js').Baggage} Baggage */
+
+export {
+  M,
+  makeScalarMapStore,
+  makeScalarWeakMapStore,
+  makeScalarSetStore,
+  makeScalarWeakSetStore,
+};
+
+const { getEnvironmentOption } = makeEnvironmentCaptor(globalThis);
+const DETECT_PROVIDE_COLLISIONS =
+  getEnvironmentOption('DETECT_PROVIDE_COLLISIONS', 'false') === 'true';
+
+const { details: X, quote: q } = assert;
 
 /** @type {import('./types').VatData} */
 let VatDataGlobal;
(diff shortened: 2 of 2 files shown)


erights:
Well, we do get a collision at https://github.com/Agoric/agoric-sdk/runs/7623466761?check_suite_focus=true , so that hypothesis is shakier than I thought.

erights:
Yeah, the hypothesis is wrong. The use in question is just lazy initialization, which is perfectly legitimate.

erights:
It is interesting that in our entire system, we see only that one violation. Nevertheless, closing as a failed experiment.

erights:
Reopening per discussion today. Will separate the once-only abstraction from the legitimately multi-use abstraction.

erights:
A possible hazard for reviewers to worry about. I'm leaving as I wrote it when I believed it was a fatal problem:

> This PR currently has a fatal problem that must prevent it from being merged. The bookkeeping for detecting collisions currently uses `Map` and `Set`, which limits the number of pairs to fit in memory, which will often defeat the purpose. Instead, it could use virtual (not durable) BigMapStores and BigSetStores, to avoid imposing a cardinality limit. But this seems like a steep price to pay in production just to detect a bug that should have been detected during development. So we also need to find a way to make it switchable, so the check can be turned on during development but not otherwise.

However, I now believe that populating *baggage*, specifically, using `provide*` and `vivify*` abstractions should be low cardinality. After all, we must have a low cardinality of kinds. It is plausible to me that we have a low cardinality of baggages, and of durable stores names at the leaves of the baggage tree. What's high cardinality are the instances of the kinds, and the contents of the non-baggage stores.

Thus, after making a few more changes, I am putting this back in review. Reviewers, please do worry about my original argument and whether you find my counter-argument compelling.

FUDCo:
My quick take is that the basic idea here seems reasonable.

One thought this spurred for me is that the kernel's `startVat` operation already contains a check to make sure that all the kinds that were defined in the vat's previous incarnation are also defined in the new incarnation, which means that in principle there's a place to stand to throw away the short term tracking tables once we're done with them, though it's not immediately obvious whether there's a graceful way to exploit this.  Possibly we could add a "this map store is being used as baggage" flag as one of the creation options (or maybe you could have a `makeBaggage` function).  Having this on a per-baggage basis would also mean that the tracking data wouldn't need to include the identity of the baggage, so it could just be a Set, which would simplify some of the logic.  However, this would involve integrating the "provide" functionality directly into the store somehow, which is an abstraction boundary crossing that I'm not entirely sure how I feel about.

One very minor stylistic thing: `assertUniqueProvide` is stateful.  It feels wrong to me for an assert function to be stateful.  It also feels weird using "provide" as a noun, but English is like that.

erights:
@FUDCo thanks for your comments. None of them seem like blockers. I'll respond in more detail when I'm back in town.

This is now ready for review!

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -91,19 +93,115 @@ export const partialAssign = (target, source) => {
...
+
+/**
+ * HAZARD: `assertUniqueProvide` uses top-level mutable state, but only for
+ * diagnostic purposes.
+ * Aside from being able to provoke it to throw an error, it is not
+ * otherwise usable as a top-level communications channel. The throwing-or-not
+ * of the error is the only way in which this top level state is observable.
+ *
+ * TODO since this still does violate our security assumptions, we should
+ * consider making it development-time only, and turn it off in production.
```
turadg, line 167: A signal of development-time would be useful generally. Do we have a mechanism for that yet? I'm used to using environment variables for that but SES precludes that.

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -91,19 +93,115 @@ export const partialAssign = (target, source) => {
...
+    const submap = /** @type {Map<string, (k?:string) => any>} */ (
+      new Map([[key, makeValue]])
+    );
+    collisions.set(baggage, submap);
+  }
+};
+
+/**
+ * Unlike `provide`, `provideOnce` can be called at most once
+ * within any vat incarnation with a given `baggage`,`key` pair.
```
turadg: > It is interesting that in our entire system, we see only that one violation

Since the "once" is the most common case, I suggest we use `provide` for this functionality and `provideLazy` for the non-asserting version.

It's a better default for someone who doesn't know the difference because when you do re-use `provide` it will give an error and tell you what to do. With the lazy version as the default you never get feedback that there's something better.

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -91,19 +93,115 @@ export const partialAssign = (target, source) => {
...
+      collisions.get(baggage)
+    );
+    if (submap.has(key)) {
+      if (submap.get(key) === makeValue) {
+        // The error message also tells us if the colliding calls happened
+        // to use the same `makeValue` only as a clue about whether
+        // these are two separate evaluations at of the same call site or
+        // two distinct call sites. However, this clue is unreliable in
+        // both directions.
+        assert.fail(X`provide collision on ${q(key)} with same makeValue`);
```
turadg: We can help the developer solve this problem:
```suggestion
        assert.fail(X`provide collision on ${q(key)} with same makeValue. Consider provideLazy()`);
```

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -91,19 +93,115 @@ export const partialAssign = (target, source) => {
 // @ts-expect-error TODO statically recognize harden
 harden(partialAssign);
 
+/**
+ * @type {WeakMap<import('./types.js').Baggage, Map<string, (k?:string) => any>>}
+ */
+const collisions = new WeakMap();
```
turadg: they're not collisions until a collision, right?

please name to something more accurate. e.g. `submapsSeen`

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -91,19 +93,115 @@ export const partialAssign = (target, source) => {
...
+      } else {
+        assert.fail(X`provide collision on ${q(key)} with different makeValue`);
+      }
+    } else {
+      submap.set(key, makeValue);
+    }
+  } else {
+    const submap = /** @type {Map<string, (k?:string) => any>} */ (
+      new Map([[key, makeValue]])
+    );
```
turadg: I believe the type is implied by the init value.
```suggestion
    const submap = new Map([[key, makeValue]]);
```

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -91,19 +93,115 @@ export const partialAssign = (target, source) => {
...
+};
+
+/**
+ * Unlike `provide`, `provideOnce` can be called at most once
+ * within any vat incarnation with a given `baggage`,`key` pair.
+ *
+ * `provideOnce` should only to be used to populate baggage,
+ * where the total number of calls to `provideOnce` must be
+ * low cardinality, since we keep the bookkeeping to detect collisions
+ * in normal language-heap memory. All the other baggage-oriented
```
turadg: When this is configured to be inert in production, this warning will no longer apply, correct?

What would it take to make this track only in dev mode? I think the plan for that should be sketched before merging, so we know what to design this API towards.

turadg (changes requested):

packages/vat-data/src/vat-data-bindings.js
erights: Nitpicky terminology note first: For most purposes, "Hardened JS" is the new name for "SES". However, Hardened JS is about the collection of features we expect to propose to standards committees, such as tc39 and tc53. The system we've built we have been calling "the SES-shim". But what does it shim. It includes Hardened JS but also some further abstractions we do not expect to bring to committee, at least not as part of the Hardened JS set of proposals. At the same time, I'm still talking about something narrower than Endo as a whole. Since the SES-shim shims it, I'll just call it "SES" below.

SES doesn't preclude the use of environment variables for paraterization, and in fact is moving to support it better. The Node mechanism is that environment variables show up in `process.env`. Thus, `export FOO=bar` turns into `process.env.FOO === 'bar'`.

SES promotes that into a general convention, supported by https://github.com/endojs/endo/blob/master/packages/ses/src/environment-options.js . It is worth reading the main doc-comment in that module. Right now, this module is only used within SES by `lockdown`, so that lockdown options not provided explicitly in the call to `lockdown` can instead be conditioned on environment variables. For example, my `~/.bashrc` file contains

```sh
export LOCKDOWN_ERROR_TAMING=unsafe
```

so that every call to `lockdown` that does not explicitly provide a setting for `errorTaming` acts like `errorTaming: 'unsafe'` *during development*. The name `'unsafe'` of course was chosen to indicate that this option should never have this setting in production.

The `ses` package currently exports nothing, and @kriskowal would like to keep it that way. But I would like to import from `environment-options.js` for use by our code post lockdown. @kriskowal and I have talked about moving it to a distinct package which both ses and others could import. We both find that attractive.

In the meantime, there is plenty of manual direct use of `process.env` in our own code. A bit of grepping shows at least 70 uses that bear further examination. Frankly, it is a bit out of control right now. We need to at least gather an explanation of all of them into one place.

`lockdown` does leave `process.env` intact on the global of the start compartment, which is why the above manual uses are possible. Compartments do not by default pass them through to created compartments. Neither does Swingset pass them through to created vats. Thus they are currently unavailable in vat code running under SwingSet. This is a policy choice. The SwingSet `config` could be enhanced to express other policies about passing environment settings through to the `process.env` of vats running under SwingSet.

Attn @warner @FUDCo @kriskowal 

So with regard to your overall question: On the one hard, we've all worked hard to make our programming environment as deterministic and hermetic as possible, minimizing the observable differences between underlying platforms. OTOH, I've looked for something like a development-time signal myself. We've succeeded so well at the first that I don't currently know of any good way we support the second. This is why I was in this `environment-options.js` code recently, and why @kriskowal and I talked about moving it to a separate package we could all import from. We would still need some kind of SwingSet mechanism to make some difference visible from inside SwingSet vats. We need to be careful how we introduce that.

Separately, we need to systematize and catalog our many manual uses of `process.env` throughout our system and the plethora of invented environment variables they represent.

erights: Since the global state does violate our security assumptions, as the comment states, not being able to turn it off for production is a blocker for this PR. I am reluctantly returning this PR to Draft status.

erights:
Returned to Draft, blocked on being able to make the security-violating global-bookkeeping on *only* during development and off in production. See https://github.com/Agoric/agoric-sdk/pull/5875#discussion_r939728572

packages/vat-data/src/vat-data-bindings.js
erights: Ah. How about I do condition it on a development time switch that can be sensed when running in unit testing, but cannot be sensed even in development from vats under swingset? This would 
   * give us some benefit,
   * would enable us to test the mechanism itself,
   * would fail-safe under current SwingSet, failing to catch errors during some development, rather than possibly violating a security property in production.

With this plan, I am putting the PR back into Ready for Review, but will not close this conversation until it is safe by that plan.

erights:
Returning back to Ready to Review.
See https://github.com/Agoric/agoric-sdk/pull/5875#discussion_r939728572

packages/vat-data/src/vat-data-bindings.js
erights: Done, according to the plan in the conversation below. Done that is, except it needs a test. I'll track that in the other conversation. Resolving.

packages/vat-data/src/vat-data-bindings.js
erights: Done

packages/vat-data/src/vat-data-bindings.js
erights: Done

packages/vat-data/src/vat-data-bindings.js
erights: Used `provisionTracker`. Done

packages/vat-data/src/vat-data-bindings.js
erights: Huh. Without the comment, it works now. Don't know why it did not. Done.

packages/vat-data/src/vat-data-bindings.js
erights: Leaving this conversation open until I add a tests of the mechanism itself, conditional on the presence or absence of `process.env.DETECT_PROVIDE_COLLISIONS` (the same option setting that an environment variable would produce).

Chris-Hibbert:
While I agree that provide-once is a fine rule, and it's usually a good idea to enforce style rules when possible, I'd have to say that the cost-benefit ratio here doesn't justify it for me. Rather than building a development-time only mechanism so we don't use global state in production, I'd prefer to omit the test.

packages/vat-data/src/environment-options.js
```diff
@@ -0,0 +1,135 @@
+// @ts-check
+
+// NOTE This module is essentially a duplicate of a module by the same name
+// in the endo repository. It is copied here *temporarily* until the one in
+// endo exports bindings we can import.
```
turadg, line 5: Would using `patch-package` also have worked?

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -18,6 +21,12 @@ export {
   makeScalarWeakSetStore,
 };
 
+const { getEnvironmentOption } = makeEnvironmentCaptor(globalThis);
+const DETECT_PROVIDE_COLLISIONS =
+  getEnvironmentOption('DETECT_PROVIDE_COLLISIONS', 'false') === 'true';
```
turadg: what ever sets this to true? if the goal is to help people become aware of a problem then they shouldn't need to know to enable the detection.

do we have any general ENV var for adopting development-time strictness? if not, time for one?

turadg (commented):
I don't share @Chris-Hibbert 's reservations about the development-time only mechanism but before approving I want to understand the workflow for it. It looks right now like the dev has to find this setting on their own and opt into it. Can we make it default on in some cases?

packages/vat-data/src/vat-data-bindings.js
erights: @samsiegart , what do each of the following lines do?

https://github.com/Agoric/agoric-sdk/blob/995beed49c308dc80b92b29a077192b037ca053f/packages/wallet/ui/src/lockdown.js#L4
```js
const consoleTaming = '%NODE_ENV%' === 'production' ? 'safe' : 'unsafe';
```

https://github.com/Agoric/dapp-card-store/blob/67d2c946130ae1510507f93c18768c9497c5e5eb/ui/src/install-ses-lockdown.js#L9
```js
if (process.env.NODE_ENV !== 'production') {
```

For the latter, I understand the notation, but not the comment above the line. But where does this setting of `NODE_ENV` come from?

For both, since it is checking whether it is set to `'production'`, does that mean that in the absence of this unexplained flag we default to unsafe settings? In general, the default way should be the safe way. An explicit override should be needed to sacrifice safety for better diagnostics.

samsiegart: `NODE_ENV` is provided by `react-scripts`. 

We can do `const consoleTaming = '%NODE_ENV%' === 'development' ? 'unsafe' : 'safe';`

In development mode, react devtools tries to extend `console` to add debugging features iirc

erights: > We can do `const consoleTaming = '%NODE_ENV%' === 'development' ? 'unsafe' : 'safe';`

Yes, please do. That way, if it is absent we'll default to safe. Implicit defaults should always be safe.

erights: @samsiegart , see https://github.com/Agoric/agoric-sdk/pull/6079

packages/vat-data/src/environment-options.js
erights: In theory, but we don't yet have such a separate package at the endo level so it's not clear what the patch would say.

In any case, recent experience makes me shy away from introducing new agoric-sdk dependencies on a patch to endo.

turadg (commented):
Changing my review to _Comment_ since most of the changes I had requested are addressed. Not _Approve_ yet since it's in draft and I haven't taken another long look.

EDIT: hmm, that didn't work to clear the Requested Changes flag. And I can't find a UI action to do that explicitly either; I suppose because it's in draft. Meanwhile I'll unsubscribe because I get notified on every push right now.

erights:
Should switch to use `@endo/env-options`

erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/7891 , even though it only detects zone based collisions. That will become good enough as we migrate more provide-based APIs to zones. Attn @michaelfig

PR #5876, opened 2022-08-02
fix: correct input validation of displayInfo

erights:
The input validation of displayInfo was using its own copy of `assertPure`, copied from @endo/marshal, which was dangerously wrong. See https://github.com/endojs/endo/pull/1242

Much of the input validation can be expressed more declaratively, compactly, and reliably using our `M` pattern language. Doing so took care of many issues implicitly, including the previous use of `assertPure`.

Note to reviewers:

In replacing as much of the previous ad hoc input validation with the pattern language as I could, I sacrificed the quality of the error messages. Please let me know whether you think I went too far and should restore some of the previous logic for better error messages. The differences are well represented by differences in the test files.

### DIFF
--- a/packages/ERTP/test/unitTests/test-inputValidation.js
+++ b/packages/ERTP/test/unitTests/test-inputValidation.js
@@ -32,7 +32,7 @@ test('makeIssuerKit bad displayInfo.decimalPlaces', async t => {
         // @ts-expect-error Intentional wrong type for testing
         harden({ decimalPlaces: 'hello' }),
       ),
-    { message: /"hello" - Must be >= -100/ },
+    { message: /^displayInfo: "hello" - Must be >= -100$/ },
   );
 
   t.throws(
@@ -58,13 +58,13 @@ test('makeIssuerKit bad displayInfo.decimalPlaces', async t => {
   t.throws(
     () =>
       makeIssuerKit('myTokens', AssetKind.NAT, harden({ decimalPlaces: 101 })),
-    { message: /101 - Must be <= 100/ },
+    { message: /^displayInfo: 101 - Must be <= 100$/ },
   );
 
   t.throws(
     () =>
       makeIssuerKit('myTokens', AssetKind.NAT, harden({ decimalPlaces: -101 })),
-    { message: /-101 - Must be >= -100/ },
+    { message: /^displayInfo: -101 - Must be >= -100$/ },
   );
 });
 
@@ -81,7 +81,7 @@ test('makeIssuerKit bad displayInfo.assetKind', async t => {
       ),
     {
       message:
-        /"something" - Must match one of \["nat","set","copySet","copyBag"\]/,
+        /^displayInfo: "something" - Must match one of \["nat","set","copySet","copyBag"\]$/,
     },
   );
 });
@@ -98,7 +98,8 @@ test('makeIssuerKit bad displayInfo.whatever', async t => {
         }),
       ),
     {
-      message: /Remainder \{"whatever":"something"\} - Must match \{\}/,
+      message:
+        /^displayInfo: Remainder \{"whatever":"something"\} - Must match \{\}$/,
     },
   );
 });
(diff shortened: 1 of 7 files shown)


packages/ERTP/test/unitTests/test-inputValidation.js
```diff
@@ -98,8 +98,7 @@ test('makeIssuerKit bad displayInfo.whatever', async t => {
         }),
       ),
     {
-      message:
-        'key "whatever" was not one of the expected keys ["decimalPlaces","assetKind"]',
+      message: /Remainder \{"whatever":"something"\} - Must match \{\}/,
```
Chris-Hibbert: That's a pretty unhelpful message. Anything we can do to improve it?

packages/ERTP/test/unitTests/test-inputValidation.js
```diff
@@ -114,8 +113,7 @@ test('makeIssuerKit malicious displayInfo', async t => {
         'bad displayInfo',
       ),
     {
-      message:
-        '"displayInfo" "bad displayInfo" must be a pass-by-copy record, not "string"',
+      message: /"bad displayInfo" - Must have shape of base: "copyRecord"/,
```
Chris-Hibbert: This only mentions 'displayInfo' because it was in the payload. Without it, this message would be hard to track down.

packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -44,8 +44,7 @@ test('bad display info', t => {
   const displayInfo = harden({ somethingUnexpected: 3 });
   // @ts-expect-error deliberate invalid arguments for testing
   t.throws(() => makeIssuerKit('fungible', AssetKind.NAT, displayInfo), {
-    message:
-      /key "somethingUnexpected" was not one of the expected keys \["decimalPlaces","assetKind"\]/,
+    message: /Remainder \{"somethingUnexpected":3\} - Must match \{\}/,
```
Chris-Hibbert: How hard would it be to add more user-supplied error messages to pattern matches?

erights: I see some ways to make them marginally better. But I don't know how to make them much better --- to the point that they could compete with bespoke manual checking logic. It is an interesting tradeoff. Good error messages are a huge deal. But so is reliably, and reliably reviewable, input validation. Pattern matching makes much of input validation declarative.

The particular one your commenting on here, regarding unallowed properties, I can improve. Do you want that in this PR? Or (my preference) I file an issue and improve this case and perhaps some others over time?

packages/ERTP/test/unitTests/test-inputValidation.js
erights: Yeah, this one falls into the same disallowed property bucket as the one I comment on below. So my preference is to file an issue and improve the pattern matching errors incrementally over time.

packages/ERTP/test/unitTests/test-inputValidation.js
erights: This one does suggest an easy enhancement I'd be happy to make in this PR, even though it would require a coordinated change to the pattern matcher: extend `fit` to take a third argument, which is a label string, much like the first argument to `Far` or `makeScalarMapStore`, that can be included in the thrown error as an additional clue about what is being tested or what the point is of testing it. In this case, `fit(..., ..., 'displayInfo')` would then, perhaps, result in prepending `displayInfo:` to the diagnostic string.

But, since it does involve a coordinated change to the pattern matchers, I'd still rather file a ticket and then do these improvements incrementally over time.

erights: Nah. I just gave it a shot and the easy way I had in mind wasn't so easy.

packages/ERTP/test/unitTests/test-issuerObj.js
erights: See https://github.com/Agoric/agoric-sdk/pull/5898

Chris-Hibbert (approved):
with [`markm-better-mismatch-errors`](https://github.com/Agoric/agoric-sdk/tree/markm-better-mismatch-errors) merged, this is an improvement.  

LGTM!

Issue #5825, opened 2022-07-26
x/authz and x/feegrant don't support ledger signing


dckc:
## Describe the bug

Our design to delegate to a non-interactive key (#4406) depends on using authz and feegrant. For ledger (#3628 ), this is only supported in yet-to-be-released versions of cosmos-sdk:

 - https://github.com/cosmos/cosmos-sdk/issues/11190 closed Feb 22
 - https://github.com/cosmos/cosmos-sdk/pull/11224

## To Reproduce
1. `agd tx authz grant ...`
2. result: error

_IOU more precise steps._

## Expected behavior
grant works so that we can, for example, send a `MsgWalletAction` using an ephemeral key.

## Platform Environment

https://github.com/Agoric/agoric-sdk/blob/dbdd77861de4f8602c796277f243bdb9fa02a057/go.mod#L7

## Additional context

 - #5761

Tartuffo:
Gaia 8 will include cosmos-sdk 4.6 which fixes this.  Otherwise, we have to backport.

Tartuffo:
May be solved by Gaia 8 (Rho)

ivanlei:
This is no longer an issue. Resolved upstream and verified in Dec `23 on Agoric chain as part of ReturnGrant efforts.

PR #5761, opened 2022-07-13
SPIKE: signing wallet actions with keplr, ledger

dckc:
refs: #4406, ##3628

## Description

_IOU_

to run it:

```
cd agoric-sdk/packages/wallet/ui/demo-signing
yarn start
```

Issue #5354, opened 2022-05-12
Smart wallet can trade with small # of users


turadg:
Epic/milestone for smart wallets.

## Acceptance criteria

- If you walk up to a new browser with your 24 words you can make a transaction
  - actually, that requires #4654 , which is beyond the scope here
- (stretch goal): If two browser tabs open it continues to work

## plan

 - [x] (stretch) #3908 
 - [x] #5386 
    - [x] PR #4648
    - [x] PR #5400
 - [x] #4406 
 - [x] #5356 
    - [x] #3756 
        - [x] #4558 
    - [x] #5353 
    - [x] #4331 
    - [x] (dup?) #3756

dckc:
I was thinking the main point here is that the only cosmos level transaction we have to sign to make an AMM trade is to accept the offer.

@samsiegart pointed out that this might not be feasible; we might need another one to get an invitation. #3908 is about getting an offer from a dapp to the wallet without going thru the chain.

In any case, the "two browser tabs" angle is only tangentially relevant.

dckc:
@samsiegart where are we now? Would you please add

 - a quick-n-dirty screen recording of whatever works
   - _p.s. a swap works! We got a 1.5min recording_
 - a few text bullet points about what works, and the first thing that does not
 - a sketch of how I could reproduce those results
   - _p.s. https://gist.github.com/samsiegart/9a81face0a8683fdb2ebb72c600a6563_

dckc:
yay! we landed all the pieces. Let's use #5073 to track end-to-end testing.

Issue #3628, opened 2021-08-09
Ensure Ledger users can sign Agoric transactions


rowgraus:
As an Agoric user who manages my assets with a hardware wallet (Ledger), I need to be able to sign transactions with Agoric offers (e.g., create a vault, trade on AMM, add liquidity on AMM) using my hardware wallet.

Our on-chain wallet design may handle this natively through Keplr.  This should be explicitly tested.

Must have:
* User can sign a transaction hash
* Ledger support

Nice/prefer to have:
* Offer give and want is clearly displayed to user
* Ledger support

Tartuffo:
@rowgraus What is the open question here?

rowgraus:
Story updated.  Assigned to Michael.  Should be estimated as non-zero now @Tartuffo @michaelfig

michaelfig:
Using a Ledger to sign wallet spend transactions (including offers) should require only minimal testing, as Keplr makes that process transparent to our software.  This issue entails working through the main wallet flows using a hardware wallet supported by Keplr.

I consider the nice-to-haves to be out of scope for this issue.  Trezor is not supported by Keplr, and for now, Keplr is our only supported signer.  It would be great if there are other signers available (MetaMask snaps and Cosmostation come to mind), but integrating and testing each one of them should be a separate issue, ideally not waiting on the Agoric team.

Offer legibility on hardware wallets will require coordination with Keplr around `SIGN_MODE_TEXTUAL` (#4529).  There will likely be broader Cosmos ecosystem support for Ledger+`SIGN_MODE_TEXTUAL`, and integrating or driving that will require significant resources, so I'm considering it out of scope for Mainnet 1.

If any of these nice-to-haves become necessary for Mainnet 1, let's create separate issues.

michaelfig:
Fallback plan is initially to use `SIGN_MODE_AMINO_JSON` for the wallet transactions so that Ledger can sign them.  `SIGN_MODE_TEXTUAL` (#4529) can wait if necessary.

dckc:
I'm struggling with `amino-json` too.

Note: to get the `wallet-action` CLI working, I made a tweak or two: ed0b3bf

cc @JimLarson 

## agd wallet-action cli

To diagnose issues with sending from js (#5761) I tried golang cli support. No joy, so far:

```
connolly@dcpc:~/projects/agoric-sdk/packages/cosmic-swingset$ make FROM_KEY=lx wallet-action
[...]

PR #5877, opened 2022-08-02
feat(wallet/ui): interactive, background signing

### LINKED ISSUE #4406: delegate from interactive signer to background signer for low-privilege actions

refs: #4398 

The on-chain wallet backend needs explicit approvals for operations that withdraw from purses, but we don't want to require explicit approvals for most user gestures.

## Description of the Design

Our CosmJS integration is intended to be used as follows:

- start with a "warm" or "cold" keypair in the Signer (Keplr/MetaMask)
- UI guides user through an optional preapproval process.  If not preapproved, then the wallet uses the "warm" keypair for every chain-modifying operation tx (with the full Signer approval process)
  - generate a "hot" Agoric CosmJS keypair, store encrypted with user password in browser
  - submit a batch of authorisation messages as one Signer transaction:
    - `feegrant` from the cold account to the hot account
    - `authz` to allow `MsgWalletAction` from hot account on behalf of cold account
    - optional - `authz` to allow `MsgWalletSpendAction` from hot account on behalf of cold account
- if user has a "hot" keypair, they must enter a password once on load to use it instead of the Signer
- in order to withdraw anything from a purse, the eventual send needs to be done via "MsgWalletSpendAction", where the backend basically remaps the marshalled input slot references as:
  - `refs.map(ref => (spendActionAmplifier.get(ref) || ref))`
  - and then handles the eventual send in the same way as `MsgWalletAction`
- categories of restricted "spend" behaviour:
  - sending from a purse to Zoe escrow as part of offer approval
  - sending from a bank purse (Cosmos-level account, `urun`, `ubld`) to anywhere else
  - NOT sending between purses otherwile

## Security Considerations

Needs review by security folks.

dckc:
closes: #4406
refs: #3628; obsoletes #5761

## Description

`makeInteractiveSigner` supports signing offers as in the initial smart wallet swap demo. (#5354)

It also supports authorizing a background signer which we plan to use in other user stories.
In particular, it uses `authz`. (It should use `feegrant` too, but see #5825 as well as https://github.com/cosmos/cosmjs/issues/1155 ) 

### Security Considerations

The secret key to the background signer is stored in `localStorage` in the clear, protected only by the same-origin policy; this risk is moderate, given:
  - the `authz` grant is limited to low-privilege `MsgWalletAction` messages
  - the grant is limited to 4hrs
  - the grant is limited to spending 0.25IST

End-user documentation for this feature is in order, not to mention completed UI support.

### Testing Considerations

Testing so far is manual. (_integration test built using a temporary button removed on request_)


dckc:
naming idea: `makeNonspendingKey`

brainstorming with @turadg

packages/wallet/ui/src/components/Requests.jsx
```diff
@@ -59,12 +60,24 @@ const RequestsInternal = ({
...
+      .submitSpendAction('{"give": 1, "want": 2}')
+      .catch(err => console.error('TODO: handle sign/broadcast errors', err));
+  };
   return (
     <div className="Requests">
       {requests.length ? (
         requests.map(Item)
       ) : (
         <div className="Empty">
+          <button type="button" onClick={signOffer}>
```
samsiegart: I think it's okay to not include this button in the merge, we have the code in other branches to reference

samsiegart (commented):
I don't have a grasp of all the stuff in keyManagement.js but the rest looks good to me assuming we're getting keplr to actually pop up with an approval window. Will defer approval to @michaelfig for the rest.

packages/wallet/ui/src/components/Requests.jsx
dckc: ok; it's gone

dckc:
chatting with @michaelfig 
> the grant is limited to 4hrs

he suggests aligning with a browser localStorage norm.

Also, maybe @rowgraus  should weigh in? should that be a week?

packages/wallet/ui/package.json
```diff
@@ -8,6 +8,7 @@
   "type": "module",
   "devDependencies": {
     "@agoric/assert": "^0.4.0",
+    "@agoric/cosmic-proto": "0.1.0",
```
michaelfig, line 11: Yaay!  Thanks @kriskowal!

packages/wallet/ui/src/contexts/Provider.jsx
```diff
@@ -2,6 +2,8 @@
 import { observeIterator } from '@agoric/notifier';
 import { E } from '@endo/far';
 import { useEffect, useState, useReducer } from 'react';
+import { SigningStargateClient } from '@cosmjs/stargate';
```
michaelfig, line 5: Let's (later) introduce a powerful module that closes over these dependencies and is imported by `Provider.jsx` to keep `Provider.jsx` simpler.

It's fine for now.

packages/wallet/ui/src/util/chainInfo.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
+
+/** @type {import('@keplr-wallet/types').Currency} */
```
michaelfig, line 3: Noice toipes.

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+    });
+    const decoded = result.grants.map(
+      g =>
+        g.authorization && GenericAuthorization.decode(g.authorization.value),
+    );
+
+    return decoded;
+  };
+
+  return freeze({
```
michaelfig:
```suggestion
  return harden({
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+  };
+  const signingClient = await connectWithSigner(chainInfo.rpc, offlineSigner, {
+    aminoTypes: new AminoTypes(converters),
+    registry: SwingsetRegistry,
+  });
+  console.debug('OfferSigner', { signingClient });
+
+  const fee = zeroFee();
+
+  return freeze({
```
michaelfig:
```suggestion
  return harden({
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+
+import {
+  MsgWalletAction,
+  MsgWalletSpendAction,
+} from '@agoric/cosmic-proto/swingset/msgs.js';
+
+import { stableCurrency, bech32Config } from './chainInfo.js';
+
+const { freeze } = Object;
+
```
michaelfig:
```suggestion
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+    queryGrants,
+  });
+};
+
+/**
+ * @param {string} granter bech32 address
+ * @param {string} grantee bech32 address
+ * @param {number} seconds expiration as seconds (Date.now() / 1000)
+ */
+const makeGrantWalletActionMessage = (granter, grantee, seconds) => {
```
michaelfig, line 224: I think CosmJS exports some message-specific facilities to make this kind of process simpler.  This is a fine approach until we sync better with the CosmJS team on our use cases.

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+ * }} WalletAction
+ * @typedef {{
+ *   owner: string, // base64 of raw bech32 data
+ *   spendAction: string,
+ * }} WalletSpendAction
+ */
+
+export const SwingsetRegistry = new Registry([
+  ...defaultRegistryTypes,
+  [SwingsetMsgs.MsgWalletAction.typeUrl, MsgWalletAction],
```
michaelfig, line 98: Would be nice to make this portion of the map part of `@agoric/cosmic-proto`.

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+    }),
+  },
+};
+
+/**
+ * key for use in localStorage
+ *
+ * arbitrary; suffix chosen randomly to avoid collisions
+ */
+export const STORAGE_KEY = 'agoric.eis0Aigi';
```
michaelfig:
```suggestion
export const BROWSER_LOCAL_STORAGE_KEY = 'agoric.eis0Aigi';
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+  const seed = provideLocalKey();
+  const wallet = await DirectSecp256k1Wallet.fromKey(
+    seed,
+    bech32Config.bech32PrefixAccAddr,
+  );
+
+  const accounts = await wallet.getAccounts();
+  console.debug('device account(s):', accounts);
+
+  const [{ address }] = accounts;
```
michaelfig, line 180: Use the same `?.` expression here as elsewhere and throw a nicer error?

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+const makeFeeGrantMessage = (granter, grantee, allowance, seconds) => {
+  return {
+    typeUrl: CosmosMessages.feegrant.MsgGrantAllowance.typeUrl,
+    value: {
+      granter,
+      grantee,
+      allowance: {
+        typeUrl: CosmosMessages.feegrant.BasicAllowance.typeUrl,
+        value: {
+          spendLimit: [{ denom: 'urun', amount: allowance }],
```
michaelfig:
```suggestion
          spendLimit: [{ denom: 'uist', amount: allowance }],
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+  chainInfo,
+  keplr,
+  connectWithSigner,
+) => {
+  const { chainId } = chainInfo;
+
+  const key = await keplr.getKey(chainId);
+
+  const offlineSigner = await keplr.getOfflineSignerAuto(chainId);
+  console.log('OfferSigner', { offlineSigner });
```
michaelfig:
```suggestion
  console.log('InteractiveSigner', { offlineSigner });
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+  const converters = {
+    ...SwingsetConverters,
+    ...createBankAminoConverters(),
+    ...createAuthzAminoConverters(),
+  };
+  const signingClient = await connectWithSigner(chainInfo.rpc, offlineSigner, {
+    aminoTypes: new AminoTypes(converters),
+    registry: SwingsetRegistry,
+  });
+  console.debug('OfferSigner', { signingClient });
```
michaelfig:
```suggestion
  console.debug('InteractiveSigner', { signingClient });
```

michaelfig (approved):
Given fixes, this looks good to merge!

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+      spendAction,
+      owner: toBase64(toAccAddress(owner)),
+    }),
+  },
+};
+
+/**
+ * key for use in localStorage
+ *
+ * arbitrary; suffix chosen randomly to avoid collisions
```
turadg: I don't get this random string. Wouldn't it be better to be human readable as to what's kept inside? e.g. `agoric.wallet.backgroundSignerKey`. 

It could be versioned to signal breaking changes, but I think if backwards compatibility breaks you'd just want to empty it out and start over.

packages/wallet/ui/src/contexts/Provider.jsx
```diff
@@ -148,11 +154,36 @@ const Provider = ({ children }) => {
   );
   const [keplrConnection, setKeplrConnection] = useState(null);
 
+  /**
+   * NOTE: relies on ambient window.fetch, window.keplr, Random.getBytes
```
turadg, line 158: better to document this with code, which will provide feedback when the assumptions are violated. e.g. down below,
```
assert(fetch, 'Missing window.fetch');
assert(keplr, 'Missing window.keplr');
assert(getBytes, 'Missing Random.getBytes');
```

packages/wallet/ui/src/contexts/Provider.jsx
```diff
@@ -202,7 +233,9 @@ const Provider = ({ children }) => {
     if (
       connectionConfig?.smartConnectionMethod === SmartConnectionMethod.KEPLR
     ) {
-      tryKeplrConnect();
+      tryKeplrConnect().catch(reason =>
+        console.error('tryKeplrConnect failed', reason),
```
turadg, line 240: we should have an error toast for failures like this. if we don't have a toast component yet, adding one is out of scope but should be scheduled work before launch. cc @samsiegart

packages/wallet/ui/src/util/SuggestChain.js
```diff
@@ -1,41 +1,28 @@
-import { SigningStargateClient } from '@cosmjs/stargate';
+/* global globalThis */
+
+import { stakeCurrency, stableCurrency, bech32Config } from './chainInfo.js';
 
 export const AGORIC_COIN_TYPE = 564;
 export const COSMOS_COIN_TYPE = 118;
 
-export async function suggestChain(networkConfig, caption = undefined) {
+const makeChainInfo = (networkConfig, rpcAddr, chainId, caption) => {
```
turadg, line 14: please type

packages/wallet/ui/src/util/SuggestChain.js
```diff
@@ -45,51 +32,36 @@ export async function suggestChain(networkConfig, caption = undefined) {
...
-  const offlineSigner = window.getOfflineSigner(chainId);
-  const cosmJS = await SigningStargateClient.connectWithSigner(
-    rpc,
-    offlineSigner,
-  );
+export async function suggestChain(
+  networkConfig,
+  /** @type {string} */ caption = undefined,
+  io = {},
+) {
```
turadg: if `caption` can be undefined then it's not `string`. let's type the function signature instead of inline.

```suggestion
/** @type {{
  * networkConfig: NetworkConfig
  * caption?: string
  * io?: Partial<IoPowers>
  * }}
  */
export async function suggestChain(networkConfig,, caption, io = {}) {
```

turadg: Though second thought, why make `io` optional? Please make them required so this module does not depend anything ambient. Then you can remove `/* global globalThis */`.

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+/**
+ * Maintain a key for signing non-spending messages in localStorage.
+ *
+ * See also `delegateWalletAction()` below.
+ *
+ * @param {object} io
+ * @param {typeof window.localStorage} io.localStorage
+ * @param {typeof import('@cosmjs/crypto').Random.getBytes} io.getBytes for key generation
+ */
+export const makeBackgroundSigner = async ({ localStorage, getBytes }) => {
```
turadg: `getBytes` without the `Random` namespace reads overly generic. please rename to something that conveys that it's generating random bytes. consider `makeRandomBytes`.

that could mean having to map the name at the callsite but I'd also suggest changing this from an option map to positional parameters. then the caller doesn't have to be concerned with the name in this context.

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+  const provideLocalKey = () => {
+    const stored = localStorage.getItem(STORAGE_KEY);
+    if (stored) {
+      return fromBase64(stored);
+    }
+    console.log(
```
turadg: consider lowering the log level or removing.
```suggestion
    console.debug(
```

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+
+  const [{ address }] = accounts;
+
+  /**
+   * Query grants for granter / grantee pair
+   *
+   * For example, to check whether `delegateWalletAction()` is necessary.
+   *
+   * @param {string} granter address
+   * @param {import('@cosmjs/tendermint-rpc').Tendermint34Client} rpcClient
```
turadg, line 188: yay types!

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,425 @@
...
+    const queryService = new QueryClientImpl(rpc);
+    console.log('query Grants', { granter, grantee: address });
+    const result = await queryService.Grants({
+      granter,
+      grantee: address,
+      msgTypeUrl: '', // wildcard
+    });
+    const decoded = result.grants.map(
+      g =>
+        g.authorization && GenericAuthorization.decode(g.authorization.value),
```
turadg: should `decoded` contain falsy? if not, filter before returning.

also if it wasn't supposed to, consider defining the return type of `queryGrants` as that would have highlighted the problem.

turadg (changes requested):
Some small changes requested.

packages/wallet/ui/src/util/keyManagement.js
dckc: it's not terribly well motivated... sort of a vague fear of colliding with... I don't know what.

Giving it some thought, I suppose it's scoped to our origin. So I just need to be sure I don't collide with any other code we write. But again... `genpass -1` is kind of a quick-and-dirty way to be 99.99999% sure of that :)

The alternative seems to be some sort of central registry agreed by all (potential) contributors.

Or maybe `agoric.wallet.backgroundSignerKey` has as effectively as much entropy as `genpass -1` and I shouldn't worry about collisions?

turadg: > Or maybe agoric.wallet.backgroundSignerKey has as effectively as much entropy as genpass -1 and I shouldn't worry about collisions?

Yes, I believe that to be the case.

packages/wallet/ui/src/contexts/Provider.jsx
samsiegart: Yea we have toasts but I think I need to do some work to make the toast service more robust and reusable across the app

packages/wallet/ui/src/util/keyManagement.js
dckc: I'm pretty fond of named parameters for IO powers and have spread the convention pretty far by now :) _IOU docs to institutionalize it, perhaps._

It's also my habit to inherit the names from whatever API they come from, but I'll grant that `getBytes` is pretty worthless in this context. Changing to `io.csprng`.

packages/wallet/ui/src/util/keyManagement.js
dckc: On further consideration, elsewhere we were dealing with the keplr API, which is doing IO to come up with an address and hence may fail. In this case, we're getting the address from `DirectSecp256k1Wallet`, which is doing only integer calculations that don't have that sort of failure mode. Any failures would show up at the `await wallet.getAccounts()` above.

packages/wallet/ui/src/contexts/Provider.jsx
dckc: `assert`s for fetch and keplr make sense, but `Missing Random.getBytes` would be a static error.
`Random.getBytes` is still noteworthy because it is ambient authority, contrary to [OCap Discipline](https://github.com/Agoric/agoric-sdk/wiki/OCap-Discipline).

packages/wallet/ui/src/contexts/Provider.jsx
dckc: I'm adding a TODO.

packages/wallet/ui/src/util/SuggestChain.js
dckc: Why indeed. It looks like we addressed the only call site in this regard.

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -0,0 +1,429 @@
...
+      owner: toBech32(bech32Config.bech32PrefixAccAddr, fromBase64(owner)),
+    }),
+    fromAmino: ({ spendAction, owner }) => ({
+      spendAction,
+      owner: toBase64(toAccAddress(owner)),
+    }),
+  },
+};
+
+// ack: justerest Sep 2019 https://stackoverflow.com/a/58110124/7963
```
turadg: a little ugly but I see it's necessary. might want to reference `UNTIL https://github.com/microsoft/TypeScript/issues/16655` so we know when we can remove this

turadg (approved):

packages/wallet/ui/src/util/keyManagement.js
dckc: oh. JS Arrays seem to have `flatMap` now... i.e. map and filter in one...

Issue #4383, opened 2022-01-25
store Zoe state in durable/virtual objects/collections


warner:
## What is the Problem Being Solved?

To allow Zoe to maintain low RAM usage despite being involved in every contract, as well as to enable Zoe to be upgraded, we need it to store all long-term state using `makeDurableKind` and virtual/durable collections.

I don't know exactly what state Zoe uses. I know there are a lot of Offers and Seats and Invitations and stuff. As a first step, I'm guessing that each of these needs to be converted into a virtual-object Kind, and any collections of them need to be stored in a virtual-collection using something like `makeBigWeakMap`. Performing this translation/porting is probably the bulk of the work.

A small step is to then switch to Durable objects/collections instead of merely Virtual ones, for the state that needs to survive an upgrade.

Then a medium step is to record all the collection objects (and handles to the Kinds that must survive) in the #4325 "baggage", so the new Zoe version can retrieve it.

Another large design question is what state ought to survive the upgrade (kept in Durable storage), and what can be discarded (kept in Virtual storage). And what clients will experience during the process. The simplest approach is to just store everything in Durable, but @erights points out that vats and contracts should be more deliberate about their upgrade process.

Tartuffo:
@Chris-Hibbert Assigning to you per the conversation in today's Eng Sync, please put an estimate on it once you have scoped out the task.

Chris-Hibbert:
Here's a [Catalog of Zoe's storage](https://github.com/Agoric/agoric-sdk/blob/master/packages/zoe/docs/zoe-catalog-of-storage.md). It lists 1 WeakSet, 9 WeakStores, and 2 Native Sets.  I see at least 3 objects (`instanceAdmin`, `ZoeSeatAdmin`, `IssuerRecord`)  that must be converted to virtual objects. The other values to be stored may be handles or presences.

`sumsByBrand` is ephemeral and doesn't need to be persisted.

Tartuffo:
Per conversation with @dtribble Chris does not need to prioritize this for Restival.

Tartuffo:
Likely people to pick this up if @erights doesnt have the bandwidth to do this in the MN-1 timeframe: @Chris-Hibbert and @turadg

PR #5879, opened 2022-08-02
4383 durable zoe

Chris-Hibbert:
refs: #4383

## Description

First step of making Zoe reachably durable. This covers all of Zoe's global state. Remaining tasks are 
 * instance state and instanceAdmin
 * FeeMint
 * zoeMint

**I recommend reviewing this by commits**. The commits separately update InvitationIssuer, escrow storage, installationStorage, and proposalSchema storage.

### Security Considerations

Maintain Zoe invariants

None.

### Testing Considerations

Ensure existing tests continue to pass.

### DIFF
--- a/packages/zoe/src/zoeService/makeInvitation.js
+++ b/packages/zoe/src/zoeService/makeInvitation.js
@@ -1,13 +1,15 @@
 // @ts-check
 import { assert, details as X } from '@agoric/assert';
-import { AmountMath, makeIssuerKit, AssetKind } from '@agoric/ertp';
+import { AmountMath, makeDurableIssuerKit, AssetKind } from '@agoric/ertp';
 import { InvitationElementShape } from '../typeGuards.js';
 
 /**
+ * @param {import('@agoric/vat-data').Baggage} baggage
  * @param {ShutdownWithFailure | undefined} shutdownZoeVat
  */
-export const createInvitationKit = (shutdownZoeVat = undefined) => {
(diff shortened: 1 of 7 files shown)


packages/zoe/src/zoeService/makeInvitation.js
```diff
@@ -1,13 +1,15 @@
...
+import { AmountMath, makeDurableIssuerKit, AssetKind } from '@agoric/ertp';
 import { InvitationElementShape } from '../typeGuards.js';
 
 /**
+ * @param {import('@agoric/vat-data').Baggage} baggage
  * @param {ShutdownWithFailure | undefined} shutdownZoeVat
  */
-export const createInvitationKit = (shutdownZoeVat = undefined) => {
-  const invitationKit = makeIssuerKit(
+export const createInvitationKit = (baggage, shutdownZoeVat = undefined) => {
```
erights: What do you think of a naming convention where functions that 
   * take a baggage first argument,
   * call a `vivifySomething`
   * therefore need to be called during the first crank, if they might be revivifying

should themselves be called `vivifySomething`?

```suggestion
export const vivifyInvitationKit = (baggage, shutdownZoeVat = undefined) => {
```

erights (approved):
LGTM!

packages/zoe/src/zoeService/makeInvitation.js
Chris-Hibbert: sounds like a good rule.

Issue #5681, opened 2022-06-28
Consume metrics into DB/Prometheus for Econ Stress Test


dckc:
## What is the Problem Being Solved?

Econ Stress test needs a bunch of data in a convenient form.

## Design

_up to @arirubinstein and co_

## Security Considerations

gets a whole bunch of moving parts, including a private signing key, out of the equasion

## Test Plan

manual, with, Nalin, @btulloh and co.

dckc:
## amm, vaults w/governance working via RPC

 - [ist-monitor2.mjs](https://github.com/dckc/ag-admin/blob/rpc/scripts/ist-monitor2.mjs) 2b8203f
 - [spreadsheet](https://docs.google.com/spreadsheets/d/1R94HzdrKYjFlr0PSg04XGvZXWDUjLHdZXOgnyUOuYi8/edit#gid=8916276)

## TODO: PSM, Reserve, BLD Boost

Among the other targets from #5421:

- reserve https://github.com/Agoric/agoric-sdk/issues/4651
- PSM https://github.com/Agoric/agoric-sdk/issues/4968
- BLD Boost https://github.com/Agoric/agoric-sdk/issues/4653
- Oracle / price authority https://github.com/Agoric/agoric-sdk/issues/4652

We have all but Oracle available, to at least some extent:

```
$ agd query vstorage keys published
children:
- amm
- psm
- reserve
- stakeFactory
- vaultFactory
```

## Oracle???

I'm not sure if we have clear plans for this one. _p.s. yes, we do: #4652_

That reminds me...

### `chainStorage` for unprivileged contracts?

How do ordinary contracts publish their data for read-only RPC? `chainStorage` is currently privileged.

_for follow-up, see #5687_

dckc:
6cd8054 is working pretty well. no individual vaults yet. nor reserve etc.

dckc:
I'm not sure what to make of this:

```
PROOF VERIFICATION FAILURE; crashing follower (Error#4)
Error#4: {"code":-32603,"message":"Internal error","data":"min height 76703 can't be greater than max height 76702"}
```

for now, I set `proof: 'none'`

[ist-monitor2.mjs](https://github.com/dckc/ag-admin/blob/rpc/scripts/ist-monitor2.mjs) [09d5a8d](https://github.com/dckc/ag-admin/commit/09d5a8d2f5d0bb3e611cc6e4992ad32eb156063c)

[...]

Issue #5508, opened 2022-06-03
include block height, timestamp in chain-streams


dckc:
## What is the Problem Being Solved?

Consumers of tick data require block_number,  timestamp as part of borrow and lend positions

```
{account_address, token_symbol, supply_balance, borrow_balance, block_number,  timestamp}
```

## Description of the Design

_I leave that to later._

## Test Plan

cc @michaelfig

dckc:
I learned from @michaelfig that cosmjs doesn't currently expose this, so we may have some upstream work to do.

dckc:
I was going to make an upstream feature request, but I see the question has already been asked and answered:

 - https://github.com/cosmos/cosmjs/issues/1181 Jun 17

Tartuffo:
Time consuming because we have to tweak the Go code and not break existing consumers in the process.

dckc:
> we have to tweak the Go code

Really? how so? The go code already produces the block number and timestamp, no?
The only problem I see is that the cosmjs library doesn't expose it.

dckc:
looks like [get-flattened-publication.sh](https://github.com/Agoric/agoric-sdk/blob/master/scripts/get-flattened-publication.sh) from #5878 includes a work-around.

gibson042:
Yes, to some extent. It reports the block height of the RPC response, but not of the data itself.

dckc:
Ah. thanks for clarifying.

dckc:
@gibson042 I'm struggling to find a timestamp. Help?

```
"value": "{\"blockHeight\":\"17\",\"values\":[\"{\\\"body\\...
```

gibson042:
We are not currently duplicating block timestamp in published data since `time` is available in the header of each block.
```sh
$ curl -sS https://xnet.rpc.agoric.net/ -X POST -H 'Content-Type: application/json' \
  --data '{ "jsonrpc": "2.0", "id": 1, "method": "block", "params": {} }' | \
  jq '.result.block.header'
{
  "version": {
    "block": "11"
  },
  "chain_id": "agoricxnet-13",
  "height": "77063",
  "time": "2022-08-23T16:35:47.681142362Z",
  "last_block_id": {
    "hash": "06E0AB8D1794FBA24D2709CFD31AB1289607E9CA618EB20BF238A9D3609233E4",
    "parts": {
      "total": 1,
[...]

PR #5880, opened 2022-08-02
fix: Add a script for fetching and flattening published capdata

### LINKED ISSUE #5878: Translate capdata into flat JSON

External consumers should not be expected to work with complex structured data, especially when that structure is `@qclass`-heavy capdata.

## Description of the Design

Create a utility for converting capdata into flat JSON, e.g. from
```
{
  centralAmount: {
    brand: { '@qclass': 'slot', iface: 'Alleged: RUN brand', index: 0 },
    value: { '@qclass': 'bigint', digits: '1859152806' }
  },
  liquidityTokens: {
    brand: {
      '@qclass': 'slot',
      iface: 'Alleged: BLDPoolLiquidity brand',
      index: 1
    },
    value: { '@qclass': 'bigint', digits: '4455322155' }
  },
  secondaryAmount: {
    brand: { '@qclass': 'slot', iface: 'Alleged: BLD brand', index: 2 },
    value: { '@qclass': 'bigint', digits: '2159517880' }
  }
}
```
to
```json
{ "central": "1859152806", "centralBrandBoardId": "board0425", "centralBrandAllegedName": "RUN", "liquidity": "4455322155", "liquidityBrandBoardId": "board03125", "liquidityBrandAllegedName": "BLDPoolLiquidity", "secondary": "2159517880", "secondaryBrandBoardId": "board03523", "secondaryBrandAllegedName": "BLD" }
```

## Security Considerations

The utility should be written such that adopting it does not risk a significant increase to potentially vulnerable surface area (e.g. by installing a complex executable).

## Test Plan

TBD

gibson042:
...with minimal dependencies
Fixes #5878

## Description

By liberal use of `awk` (which is required in any POSIX implementation), dependencies have been reduced to hopefully-noncontroversial `curl` and `base64`.

```sh
$ scripts/get-flattened-publication.sh https://xnet.rpc.agoric.net/ published.amm.pool0.metrics | \
  jq .
{
  "blockHeight": "513450",
  "centralAmount-brand-id": "board0425",
  "centralAmount-brand-allegedName": "RUN",
  "centralAmount-value": "202019830953",
  "liquidityTokens-brand-id": "board0188",
  "liquidityTokens-brand-allegedName": "IbcATOMLiquidity",
  "liquidityTokens-value": "202503724998",
  "secondaryAmount-brand-id": "board0639",
  "secondaryAmount-brand-allegedName": "IbcATOM",
  "secondaryAmount-value": "101498052"
}
```

It's also possible to use with e.g. `agd`, at the cost of block height information.

```sh
$ agd --node=https://xnet.rpc.agoric.net:443 query vstorage data published.amm.pool0.metrics -o json | \
  scripts/flatten_json_sequence.awk -v unwrap=value -v capdata=true | \
  jq .
{
  "centralAmount-brand-id": "board0425",
  "centralAmount-brand-allegedName": "RUN",
  "centralAmount-value": "202019830953",
  "liquidityTokens-brand-id": "board0188",
  "liquidityTokens-brand-allegedName": "IbcATOMLiquidity",
  "liquidityTokens-value": "202503724998",
  "secondaryAmount-brand-id": "board0639",
  "secondaryAmount-brand-allegedName": "IbcATOM",
  "secondaryAmount-value": "101498052"
}
```

### Security Considerations

No one should object to _using_ the code, since it is so narrowly-scoped in both power and effects. However, _auditing_ it is likely more difficult than would be the case if it were written in a more common language such as JS or Go (which could be done).

I'm also not sure if this code belongs in [agoric-sdk](https://github.com/Agoric/agoric-sdk) or another repository such as [dckc/ag-admin](https://github.com/dckc/ag-admin).

I have tried to include sufficient examples in source code comments.

### Testing Considerations

The input data is by nature a moving target, so I think it makes sense to skip tests at least until we have adopted another publication approach (cf. #5681, #5508, #5366).


dckc (commented):
> minimal dependencies

no kidding! An implementation of JSON in AWK. That's... hard core.

OTOH, as you say, reviewing it is kind of a pain. At least one of our customers is happy with python. I'd be OK reviewing python.

As is, I'll just say I have no objections but leave approval to @arirubinstein .

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,80 @@
...
+#         "key": null,
+#         "value": "<base64-encoded JSON text>",
+#         "proofOps": null,
+#         "height": "<decimal digits>",
+#         "codespace": ""
+#       }
+#     }
+#   }
+vars="$(
+  curl -sS "${URL_PREFIX%/}/abci_query?path=%22/custom/vstorage/data/$STORAGE_KEY%22" | \
```
gibson042: Incidentally, this is probably a bug somewhere in Cosmos/Tendermint. Their [documentation](https://docs.tendermint.com/master/rpc/#/ABCI/abci_query) gives example values of `path` like `/a/b/c`, but making a query with `?path=/a/b/c` yields a response like
```
{
  "jsonrpc": "2.0",
  "id": -1,
  "error": {
    "code": -32602,
    "message": "Invalid params",
    "data": "error converting http params to arguments: invalid character '/' looking for beginning of value"
  }
}
```
Something is apparently JSON-decoding the parameter value rather than using it directly.

gibson042: Specifically https://github.com/tendermint/tendermint/blob/75d51e18f740c7cbfb7d8b4d49182ee6c7f41982/rpc/jsonrpc/server/http_uri_handler.go#L42-L51 , but I can't really get to the bottom of the reflection-heavy `nonJSONStringToArg` in `httpParamsToArgs`.

turadg:
> By liberal use of awk (which is required in any POSIX implementation), dependencies have been reduced to hopefully-noncontroversial curl and base64.

Was the use of https://stedolan.github.io/jq/ controversial? 

I see the ticket requirement,
> The utility should be written such that adopting it does not risk a significant increase to potentially vulnerable surface area (e.g. by installing a complex executable).

I take that to mean that the executable we're giving them isn't complex. `jq` is included in most distributions so what we'd be providing would be much simpler using that than this awk script.

gibson042:
I don't think `jq` was _inherently_ controversial, but there was definitely consternation about the complexity of requiring expressions like the following as seen on Slack from @dckc (and which would only get more complicated to incorporate block height and/or data from `slots`):
```
.value | fromjson | .body | fromjson | {
  centralAmount: .centralAmount | { brand: .brand.iface, value: .value.digits | tonumber | (. / 1000000 ) },
  secondaryAmount: .secondaryAmount | { brand: .brand.iface, value: .value.digits | tonumber | (. / 1000000 ) },
  liquidityTokens: .liquidityTokens | { brand: .brand.iface, value: .value.digits | tonumber | (. / 1000000 ) }
}
```

turadg:
> complexity of requiring expressions like the following

Maybe we're talking about different complexity. If that jq syntax were put into a `flatten_json_sequence.jq` then the end user complexity would be the same (modulo having to perhaps `apt-get jq`).

The stark difference is within `flatten_json_sequence` file. The the AWK is _much_ harder to follow than any JQ and it could have errors in its JSON parsing.

scripts/get-flattened-publication.sh
gibson042: Opened https://github.com/tendermint/tendermint/issues/9164

arirubinstein: urlencode?

gibson042: No, I get the same results with `%22` vs. `"`. I even replaced an `a` with `\u0061` (which is a JSON escape sequence) in the quote-wrapped version and it still worked.

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+  '
+)"
+eval "$vars"
+
+# Decode the value into flat JSON and insert a member for the block height.
+printf '%s' "${value_base64:-}" | \
+  base64 -d | \
+  awk -f "$FLATTENER" -v unwrap=value -v capdata=true | \
```
gibson042: For the record, if `jq` is acceptable then this most complex use of `awk` (unwrap, expand capdata, and then flatten) can be replaced with the following:
```suggestion
  jq '[
    # Decode JSON in `value`, capture `slots`, then decode JSON in `body`.
    .value | fromjson | .slots as $slots | .body | fromjson |
    walk(
      if type=="object" and .["@qclass"]=="bigint" then
        # Replace bigint capdata with the sequence of digits.
        .digits
      elif type=="object" and .["@qclass"]=="slot" then
        # Replace slot reference capdata with {
        #   id: <value from global `slots`>,
        #   allegedName: <extracted from local `iface`>,
        # }.
        {
          id: $slots[.index],
          allegedName: .iface | sub("^Alleged: (?<name>.*) brand$"; "\(.name)"; "m")
        }
      else
        .
      end
    ) |
    paths(scalars) as $path |
    # Join deep object member names with "-".
    { key: $path | join("-"), value: getpath($path) } 
  ] | from_entries'
```

turadg: I'd approve that ;-)

> if jq is acceptable

I believe it is but I defer to @arirubinstein who I think represents the customer of this script

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,92 @@
...
+    "params": { "path": "/custom/vstorage/data/%s" }
+  }' "$STORAGE_KEY"
+)" | \
+# Decode, simplify, flatten, and compact the output.
+jq -c '
+  # Capture block height.
+  .result.response.height? as $height |
+
+  # Decode `value` as base64, then decode that as JSON.
+  .result.response.value | @base64d | fromjson |
```
gibson042, line 67: Today I learned that `jq` can encode/decode base64: https://stedolan.github.io/jq/manual/#Formatstringsandescaping

gibson042 (commented):
@arirubinstein @dckc @turadg Replaced the awk with jq. PTAAL.

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,92 @@
...
+       e.g. \"https://xnet.rpc.agoric.net/\".
+  STORAGE_KEY is a chain storage key name, e.g. \"published.amm.pool0.metrics\".
+";
+
+URL_PREFIX="${1:-}";
+STORAGE_KEY="${2:-}";
+
+if [ ":${1:-}" = '--help' -o ":$URL_PREFIX" = : -o ":$STORAGE_KEY" = : -o $# -gt 2 ]; then
+  printf '%s' "$USAGE"
+  exit 64
```
turadg, line 17: TIL!

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,92 @@
...
+fi
+
+# Verify jq version.
+jq --version | awk '
```
turadg, line 21: I'm more familiar with version checking by grep and having the condition in shell code. nbd.

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,92 @@
...
+  /^jq-1[.]([6-9]|[1-9][0-9])/ {
+    exit;
+  }
+  {
+    print "jq version out of range (must be >=1.6 <2.0): " $0 > "/dev/stderr";
```
turadg: just curious, what's required in 1.6?

scripts/get-flattened-publication.sh
```diff
@@ -0,0 +1,92 @@
...
+#       "response": {
+#         "value": "<base64-encoded JSON text>",
+#         "height": "<decimal digits>",
+#         ...
+#       }
+#     }
+#   }
+
+# Avoid the GET interface in case interpretation of `path` as JSON is ever fixed.
+# https://github.com/tendermint/tendermint/issues/9164
```
turadg, line 51: helpful 👍

turadg (approved):
> The input data is by nature a moving target, so I think it makes sense to skip tests at least until we have adopted another publication approach

Fair. I trust you've tested it with the status quo and that you'll remember to retest when the source data changes.

> not sure if this code belongs in [agoric-sdk](https://github.com/Agoric/agoric-sdk) or another repository such as [dckc/ag-admin](https://github.com/dckc/ag-admin).

Good question. I think it's fine to include here in this misc script directory. It will also help keep the version in sync with producer logic.

arirubinstein (approved):

scripts/get-flattened-publication.sh
gibson042: `@base64d`, `walk`, and `getpath`.

PR #5882, opened 2022-08-02
Cosmic SwingSet Protobuf IDL stubs for JS and TS

### LINKED ISSUE #5881: Create Cosmic Swingset IDL generated stubs npm package

Clients interacting with the Agoric chain over RPC in JavaScript need generated IDL stubs.

## Description of the Design

The prevailing idiom is to publish an npm package with the generated stubs from the IDL, rather than compelling clients do this themselves. We’ve elected to use `protoc`, `ts-proto`, and `tsc` in concert so we have both code and TypeScript types. There is no solution that generates JavaScript with TypeScript style JSDoc.

## Security Considerations

None known by this author.

## Test Plan

It should be sufficient to verify that the generated stubs can be imported and we expect any bugs discovered to either amount to configuration or implementation of the protoc and ts-proto tooling. We will shortly be using these stubs in `agoric-cli` which should give us adequate coverage in practice.

kriskowal:
closes: #5881

## Description

Creates a new package for shipping Cosmic SwingSet IDL stubs for JavaScript and TypeScript.

### Security Considerations

None known to author.

This package will be useful to anyone implementing an Agoric Chain RPC client in JavaScript or TypeScript.

### Testing Considerations

The test ensures the generated code can be imported but does not do an integration test with a  fake chain. I’ll shortly be adding this as a dependency of Agoric CLI which will exercise this IDL while deploying a contract. An integration test will follow.


kriskowal:
Adding a shallow dependency on `protobufjs` caused it to get deduplicated. There’s a commit that consolidates the patches.

Michael advised checking in generated artifacts over requiring `protoc` on all developer workstations.

Commits are individually reviewable.

packages/cosmic-swingset-proto/proto/gogoproto
```diff
@@ -0,0 +1 @@
+/Users/kris/agoric-sdk/golang/cosmos/third_party/proto/gogoproto
```
dckc: this symlink should be relative, yes?

packages/cosmic-swingset-proto/proto/swingset
```diff
@@ -0,0 +1 @@
+/Users/kris/agoric-sdk/golang/cosmos/proto/agoric/swingset
```
dckc: relative again

dckc (commented):
a couple symlinks should be relative.

Also, I think I should try using this for the wallet signing stuff to check that niceties like .ts types work.

kriskowal:
Fixed symlinks and the matching how-to on generating `proto` and added a CI workflow for the new package.

kriskowal:
Yay. All checks are green right now. I’ll rename the package to `@agoric/cosmic-proto` tomorrow, per @michaelfig’s tip.

dckc:
re security considerations: this code is between what users think they're signing (cuz our "wallet" UI said so) and what they're actually signing. So one possible attack would be to get something in this code that says "send 1000 tokens to  ATTACKER-ADDRESS".

This is mitigated by the fact that their signer (e.g. keplr) would show "send 1000 tokens to ATTACKER". But still, it would be a big pain.

dckc:
I can't seem to import stuff from `@agoric/cosmic-swingset-proto/msgs.js`.

I found this makes it happy:

```js
import { MsgWalletAction } from '@agoric/cosmic-swingset-proto/gen/swingset/msgs.js';
```

kriskowal:
> I can't seem to import stuff from `@agoric/cosmic-swingset-proto/msgs.js`.
> 
> I found this makes it happy:
> 
> ```js
> import { MsgWalletAction } from '@agoric/cosmic-swingset-proto/gen/swingset/msgs.js';
> ```

@dckc What version of Node.js and is there any tooling involved that might not support `package.json` `"exports"`?

dckc:
> @dckc What version of Node.js ...

v16.16.0 . other relevant details below.

> is there any tooling involved that might not support `package.json` `"exports"`?

I don't know.

### Details
<details>
 <summary>how to reproduce, tool versions</summary>

```
~/projects/agoric-sdk/packages/wallet/ui$ yarn eslint src/util/keyManagement.js 
yarn run v1.22.19
$ /home/connolly/projects/agoric-sdk/node_modules/.bin/eslint src/util/keyManagement.js

/home/connolly/projects/agoric-sdk/packages/wallet/ui/src/util/keyManagement.js
  20:8  error  Unable to resolve path to module '@agoric/cosmic-swingset-proto/msgs.js'  import/no-unresolved

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

context:

```
~/projects/agoric-sdk/packages/wallet/ui$ yarn eslint --version
v7.32.0

~/projects/agoric-sdk/packages/wallet/ui$ yarn node --version
v16.16.0

~/projects/agoric-sdk/packages/wallet/ui$ git log -n 1
commit 2189f62f1ed3454ed0a1756c17f47100c2ebc5e6 (HEAD -> dc-wallet-signing-prod, origin/dc-wallet-signing-prod)
Author: Dan Connolly <connolly@agoric.com>
Date:   Wed Aug 3 18:43:30 2022 -0500

    chore(wallet/ui): prune proto stubs
```
</details>

dckc (approved):
It works! Both eslint and actual `signAndBroacast()` usage.

The version I grabbed was 256f5d5 but it looks like only cosmetic changes since then.

kriskowal:
I’ve renamed the package `@agoric/cosmic-proto`. It currently only provides `@agoric/cosmic-proto/swingset/msgs.{js,ts}` and the intention is to use this package for any other Agoric/Cosmos IDL we need in the future.

This required more use of symlinks since the implementation is in the internal generated `gen/agoric/swingset/msgs.{js,ts}` modules which `exports` hides and other implementations do not recognize `exports` for the purpose of aliasing. Generating code outside of `gen` seemed like a bad idea and manually curating symlinks and `exports` is tedious but at least encourages deliberate control of the public API.

PR #5883, opened 2022-08-03
feat(wallet/api): marshal contexts for wallet, board

dckc:
refs: #4398
obsoletes #5874

## Description

When serializing wallet state for chainStorage, there's a tension between
  - keeping purses etc. closely held
  - recognizing identity of brands also referenced in the state of contracts such as the AMM

`makeMarshal` is parameterized by the type of slots. Here we use a disjoint union of
   - board ids for widely shared objects
   - `kind:seq` ids for closely held objects; for example `purse:123`

@samsiegart note that unlike #5874 , this PR doesn't include any changes to `lib-wallet.js`; I hope you can layer those changes on top of this as a form of review.

This PR includes `makeLoggingPresence`, which is some advanced work on smart wallet middleware. While the design is still in progress, test coverage of this version is good enough that I'd like to land it.

### Security Considerations

As @warner noted, there's a risk that chainStorage data from the AMM could try to refer to a purse using a slots such as `purse:123`; such a reference must fail.

not much; these are internal APIs.

### Testing Considerations

Test coverage is 100%.

```
agoric-sdk/packages/wallet/api$ yarn c8 ava test/test-wallet-marshal.js test/test-middleware.js
yarn run v1.22.19
$ .../.bin/c8 ava test/test-wallet-marshal.js test/test-middleware.js

  ✔ middleware › makeImportContext in wallet UI can unserialize messages
  ✔ middleware › makeLoggingPresence logs calls on purse/payment actions
  ✔ wallet-marshal › makeImportContext preserves identity across AMM and wallet
  ✔ wallet-marshal › ensureBoardId allows re-registration; initBoardId does not
  ✔ wallet-marshal › makeExportContext.serialize handles unregistered identites
  ─

  5 tests passed
---------------------|---------|----------|---------|---------|-------------------
File                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------------|---------|----------|---------|---------|-------------------
All files            |     100 |      100 |     100 |     100 |                   
 marshal-contexts.js |     100 |      100 |     100 |     100 |                   
---------------------|---------|----------|---------|---------|-------------------
Done in 0.71s.
```

### DIFF
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -105,12 +105,14 @@ const initSlotVal = (table, slot, val) => {
 /**
  * Make context for exporting wallet data where brands etc. can be recognized by boardId.
  *
- * Objects should be registered before serialization:
- *   - for closely-held object, use initPurseId etc.
- *   - for published objects such as brands and issuers, use initBoardId / ensureBoardId
+ * When serializing wallet state for, there's a tension between
  *
- * Unregistered objects (such as instances in amounts) get serialized
- * with a fresh slot.
+ *  - keeping purses etc. closely held
+ *  - recognizing identity of brands also referenced in the state of contracts such as the AMM
+ *
+ * `makeMarshal()` is parameterized by the type of slots. Here we use a disjoint union of
+ *   - board ids for widely shared objects
+ *   - kind:seq ids for closely held objects; for example purse:123
  */
 export const makeExportContext = () => {
   const walletObjects = {
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -16,78 +16,103 @@ const { details: X, quote: q } = assert;
  * with a fresh slot.
  */
 export const makeExportContext = () => {
-  const toVal = {
-    /** @type {MapStore<string, Purse>} */
-    purse: makeScalarMap(),
+  const myData = {
+    purse: {
+      /** @type {MapStore<string, Purse>} */
+      bySlot: makeScalarMap(),
+      /** @type {MapStore<Purse, string>} */
+      byVal: makeScalarMap(),
+    },
+    payment: {
+      /** @type {MapStore<string, Payment>} */
+      bySlot: makeScalarMap(),
+      /** @type {MapStore<Payment, string>} */
+      byVal: makeScalarMap(),
+    },
     // TODO: offer, contact, dapp
   };
-  const fromVal = {
-    /** @type {MapStore<Purse, string>} */
-    purse: makeScalarMap(),
-    // TODO: offer, contact, dapp
+  const sharedData = {
+    /** @type {MapStore<string, unknown>} */
+    bySlot: makeScalarMap(),
+    /** @type {MapStore<unknown, string>} */
+    byVal: makeScalarMap(),
   };
-  /** @type {MapStore<unknown, string>} */
-  const sharedData = makeScalarMap();
 
   /**
    * @param {string} slot
    * @param {string} _iface
    */
   const slotToVal = (slot, _iface) => {
-    const kind = Object.keys(toVal).find(k => slot.startsWith(k));
+    if (sharedData.bySlot.has(slot)) {
+      return sharedData.bySlot.get(slot);
+    }
+    const kind = Object.keys(myData).find(k => slot.startsWith(k));
     assert(kind, X`bad slot kind: ${slot}`);
     const id = slot.slice(kind.length + 1);
-    const val = toVal[kind].get(id); // or throw
+    const val = myData[kind].bySlot.get(id); // or throw
     return val;
   };
 
-  let nonce = 0;
+  let unknownNonce = 0;
 
   const valToSlot = val => {
-    if (sharedData.has(val)) {
-      return sharedData.get(val);
+    if (sharedData.byVal.has(val)) {
+      return sharedData.byVal.get(val);
     }
-    for (const kind of Object.keys(fromVal)) {
-      if (fromVal[kind].has(val)) {
-        const id = fromVal[kind].get(val);
+    for (const kind of Object.keys(myData)) {
+      if (myData[kind].byVal.has(val)) {
+        const id = myData[kind].byVal.get(val);
         return `${kind}:${id}`;
       }
     }
-    nonce += 1;
-    const slot = `unknown:${nonce}`;
-    sharedData.init(val, slot);
+    unknownNonce += 1;
+    const slot = `unknown:${unknownNonce}`;
+    sharedData.byVal.init(val, slot);
+    sharedData.bySlot.init(slot, val);
     return slot;
   };
 
+  const makeSaver = (kind, tables) => {
+    let nonce = 0;
+    return val => {
+      const slot = `${(nonce += 1)}`;
+      tables.bySlot.init(slot, val);
+      tables.byVal.init(val, slot);
+    };
+  };
+
   return harden({
+    savePurseActions: makeSaver('purse', myData.purse),
+    savePaymentActions: makeSaver('payment', myData.payment),
     /**
      * @param {string} id
      * @param {Purse} purse
      */
     initPurseId: (id, purse) => {
-      toVal.purse.init(id, purse);
-      fromVal.purse.init(purse, id);
+      myData.purse.bySlot.init(id, purse);
+      myData.purse.byVal.init(purse, id);
     },
-    purseEntries: toVal.purse.entries,
+    purseEntries: myData.purse.bySlot.entries,
     /**
      * @param {string} id
      * @param {unknown} val
      */
     initBoardId: (id, val) => {
-      sharedData.init(val, id);
+      sharedData.bySlot.init(id, val);
+      sharedData.byVal.init(val, id);
     },
     /**
      * @param {string} id
      * @param {unknown} val
      */
     ensureBoardId: (id, val) => {
-      if (sharedData.has(val)) {
-        assert.equal(sharedData.get(val), id);
+      if (sharedData.byVal.has(val)) {
+        assert.equal(sharedData.byVal.get(val), id);
         return;
       }
-      sharedData.init(val, id);
+      sharedData.bySlot.init(id, val);
+      sharedData.byVal.init(val, id);
     },
-    issuerEntries: toVal.purse.entries,
     ...makeMarshal(valToSlot, slotToVal),
   });
 };
@@ -182,18 +207,7 @@ export const makeImportContext = () => {
     }),
   };
 
-  const makeSaver = (kind, tables) => {
-    let nonce = 0;
-    return val => {
-      const slot = `${(nonce += 1)}`;
-      tables.bySlot.init(slot, val);
-      tables.byVal.init(val, slot);
-    };
-  };
-
   return harden({
-    savePurseActions: makeSaver('purse', myData.purse),
-    savePaymentActions: makeSaver('payment', myData.payment),
     fromMyWallet: Far('wallet marshaller', { ...marshal.fromMyWallet }),
     fromBoard: Far('board marshaller', { ...marshal.fromBoard }),
   });
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -8,8 +8,12 @@ const { details: X, quote: q } = assert;
 /**
  * Make context for exporting wallet data where brands etc. can be recognized by boardId.
  *
- * All purses, brands, etc. must be registered before serialization / unserialization.
- * Neither the serializer nor the unserializer creates new presences.
+ * Objects should be registered before serialization:
+ *   - for closely-held object, use initPurseId etc.
+ *   - for published objects such as brands and issuers, use initBoardId / ensureBoardId
+ *
+ * Unregistered objects (such as instances in amounts) get serialized
+ * with a fresh slot.
  */
 export const makeExportContext = () => {
   const toVal = {
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -48,6 +48,8 @@ export const makeExportContext = () => {
     return val;
   };
 
+  let nonce = 0;
+
   const valToSlot = val => {
     if (sharedData.has(val)) {
       return sharedData.get(val);
@@ -58,24 +60,43 @@ export const makeExportContext = () => {
         return `${kind}:${id}`;
       }
     }
-    assert.fail(X`cannot serialize ${val}`);
+    nonce += 1;
+    const slot = `unknown:${nonce}`;
+    sharedData.init(val, slot);
+    return slot;
   };
 
   return harden({
+    /**
+     * @param {string} id
+     * @param {Purse} purse
+     */
     initPurseId: (id, purse) => {
       toVal.purse.init(id, purse);
       fromVal.purse.init(purse, id);
     },
     purseEntries: toVal.purse.entries,
+    /**
+     * @param {string} id
+     * @param {Brand} brand
+     */
     initBrandId: (id, brand) => {
       toVal.brand.init(id, brand);
       fromVal.brand.init(brand, id);
     },
     brandEntries: toVal.purse.entries,
+    /**
+     * @param {string} id
+     * @param {Issuer} issuer
+     */
     initIssuerId: (id, issuer) => {
       toVal.issuer.init(id, issuer);
       fromVal.issuer.init(issuer, id);
     },
+    /**
+     * @param {string} id
+     * @param {unknown} val
+     */
     initBoardId: (id, val) => {
       sharedData.init(val, id);
     },
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -117,10 +117,18 @@ export const makeExportContext = () => {
   });
 };
 
+const defaultMakePresence = iface => {
+  const severed = `SEVERED: ${iface.replace(/^Alleged: /, '')}`;
+  const thing = /** @type {any} */ (Far(severed, {}));
+  return thing;
+};
+
 /**
  * Make context for unserializing wallet or board data.
+ *
+ * @param {(iface: string) => unknown} [makePresence]
  */
-export const makeImportContext = () => {
+export const makeImportContext = (makePresence = defaultMakePresence) => {
   // constraint: none of these keys has another as a prefix.
   // and none overlaps with board
   const myData = {
@@ -138,15 +146,14 @@ export const makeImportContext = () => {
     },
     // TODO: 6 in total, right?
   };
-  /** @type {MapStore<string, unknown>} */
-  const sharedData = makeScalarMap();
+  const sharedData = {
+    /** @type {MapStore<string, unknown>} */
+    bySlot: makeScalarMap(),
+    /** @type {MapStore<unknown, string>} */
+    byVal: makeScalarMap(),
+  };
 
   const kindOf = slot => Object.keys(myData).find(k => slot.startsWith(k));
-  const makePresence = (slot, iface) => {
-    const severed = `SEVERED: ${iface.replace(/^Alleged: /, '')}`;
-    const thing = /** @type {any} */ (Far(severed, {}));
-    return thing;
-  };
 
   const slotToVal = {
     /**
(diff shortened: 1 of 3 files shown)


dckc:
good progress:

```
agoric-sdk/packages/wallet/api$ yarn c8 ava test/test-wallet-marshal.js test/test-middleware.js
yarn run v1.22.19
$ /home/connolly/projects/agoric-sdk/node_modules/.bin/c8 ava test/test-wallet-marshal.js test/test-middleware.js

  ✔ middleware › makeImportContext in wallet UI can unserialize messages
  ✔ middleware › makeLoggingPresence logs calls on purse/payment actions
  ✔ wallet-marshal › makeImportContext preserves identity across AMM and wallet
  ✔ wallet-marshal › ensureBoardId allows re-registration; initBoardId does not
  ✔ wallet-marshal › makeExportContext.serialize handles unregistered identites
  ─

  5 tests passed
---------------------|---------|----------|---------|---------|-------------------
File                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------------|---------|----------|---------|---------|-------------------
All files            |   99.61 |    97.29 |     100 |   99.61 |                   
 marshal-contexts.js |   99.61 |    97.29 |     100 |   99.61 | 206               
---------------------|---------|----------|---------|---------|-------------------
Done in 0.79s.
```

samsiegart:
So you'd like me to do something like https://github.com/Agoric/agoric-sdk/commit/b516b66a2b11731f3a51469f26b4115b13bd3315 to integrate this? Can we make a checklist of things to do, like purses, payments, issuers, brands, etc

packages/wallet/api/test/test-middleware.js
```diff
@@ -0,0 +1,76 @@
...
+  };
+  const myPayment = Far('payment', {});
+  await E(purse.actions).deposit(myPayment); // promise resolves???
+  await E(purse.actions)(1, 'thing');
+  t.deepEqual(msgs, [
+    ['applyMethod', purse.actions, 'deposit', [myPayment]],
+    ['applyFunction', purse.actions, [1, 'thing']],
+  ]);
+
+  const ctx = makeExportContext();
```
samsiegart, line 40: I'm confused about this test case, because `ctx` I thought would be inside `lib-wallet`, while the logging presence would be used outside `lib-wallet`

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+import { makeScalarMap } from '@agoric/store';
+import { Far, makeMarshal, Remotable } from '@endo/marshal';
+import { HandledPromise } from '@endo/eventual-send'; // TODO: convince tsc this isn't needed
+
+const { details: X, quote: q } = assert;
+
+/**
+ * Make context for exporting wallet data where brands etc. can be recognized by boardId.
+ *
+ * Objects should be registered before serialization:
```
turadg: "should" doesn't convey the stakes here. Is the following accurate?

By default, objects get serialized with a fresh slot. There are two cases where this is bad:
1. The object is closely held, so the default behavior would expose it 
2. The object is published, so the default behavior would create a new object identity that can't be matched to the public one

Therefore to prevent these problems you must:
1. Use `initPurseId` for purse objects
2. Use `initBoardId` / `ensureBoardId` for published board objects (e.g. brands and issuers)

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+    },
+    payment: {
+      /** @type {MapStore<string, Payment>} */
+      bySlot: makeScalarMap(),
+      /** @type {MapStore<Payment, string>} */
+      byVal: makeScalarMap(),
+    },
+    // TODO: offer, contact, dapp
+  };
+  const sharedData = {
```
turadg: shared with whom? I think this is the case #2 in comment above so I suggest matching the language. e.g. `publishedData`.

is it all `boardData`?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * Objects should be registered before serialization:
+ *   - for closely-held object, use initPurseId etc.
+ *   - for published objects such as brands and issuers, use initBoardId / ensureBoardId
+ *
+ * Unregistered objects (such as instances in amounts) get serialized
+ * with a fresh slot.
+ */
+export const makeExportContext = () => {
+  const myData = {
```
turadg: `my` doesn't imply privacy. maybe `closeData`? a comment on the design of the disjoint sets could help here.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+  };
+  const sharedData = {
+    /** @type {MapStore<string, unknown>} */
+    bySlot: makeScalarMap(),
+    /** @type {MapStore<unknown, string>} */
+    byVal: makeScalarMap(),
+  };
+
+  /**
+   * @param {string} slot
```
turadg:
```suggestion
  /**
   * Look up in published data and fall back to closely-held data.
   *  
   * @param {string} slot
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+   * @param {string} _iface
+   */
+  const slotToVal = (slot, _iface) => {
+    if (sharedData.bySlot.has(slot)) {
+      return sharedData.bySlot.get(slot);
+    }
+    const kind = Object.keys(myData).find(k => slot.startsWith(k));
+    assert(kind, X`bad slot kind: ${slot}`);
+    const id = slot.slice(kind.length + 1);
```
turadg: is a slot string meant to be parseable? I don't know that as part of the commitment of the slot identifier.

Either way I'd suggest that the parsing function come from the package that composes the slot strings. For something like,

```suggestion
    const kind = Object.keys(myData).find(k => slotMatchesKind(slot, k));
    assert(kind, X`bad slot kind: ${slot}`);
    const id = idForSlotKind(slot, kind);
```

I see below that slot strings are also being composed here. If being composable and parseable by any code is a commitment of the slot design, then I'd request that the type `Slot` (LiveSlot?) be defined and its format documented on the type. I'll accept "out of scope" for addressing generally but would ask here for a simple typedef at the top with a ticket to solve generally.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+      bySlot: makeScalarMap(),
+      /** @type {MapStore<Purse, string>} */
+      byVal: makeScalarMap(),
+    },
+    payment: {
+      /** @type {MapStore<string, Payment>} */
+      bySlot: makeScalarMap(),
+      /** @type {MapStore<Payment, string>} */
+      byVal: makeScalarMap(),
+    },
```
turadg: I see this called a Table below. Worth a type, which would simplify here:
```suggestion
    /** @type {ContextTable<Payment>} */
    payment: {
      bySlot: makeScalarMap(),
      byVal: makeScalarMap(),
    },
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+    sharedData.bySlot.init(slot, val);
+    return slot;
+  };
+
+  const makeSaver = (kind, tables) => {
+    let nonce = 0;
+    return val => {
+      const slot = `${(nonce += 1)}`;
+      tables.bySlot.init(slot, val);
+      tables.byVal.init(val, slot);
```
turadg: the `tables` param reads like something with arbitrary size. Please make it a singular of some type (per comment above)

I see below there is `table.bySlot`

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+      sharedData.byVal.init(val, id);
+    },
+    ...makeMarshal(valToSlot, slotToVal),
+  });
+};
+
+const defaultMakePresence = iface => {
+  const severed = `SEVERED: ${iface.replace(/^Alleged: /, '')}`;
+  const thing = /** @type {any} */ (Far(severed, {}));
+  return thing;
```
turadg: could this be `unknown`? could be typed on the function

also, `thing` is distracting.
```suggestion
  return Far(severed, {});
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+};
+
+/**
+ * Make context for unserializing wallet or board data.
+ *
+ * @param {(iface: string) => unknown} [makePresence]
+ */
+export const makeImportContext = (makePresence = defaultMakePresence) => {
+  // constraint: none of these keys has another as a prefix.
+  // and none overlaps with board
```
turadg: reads like this is an unenforced requirement. can it be enforced as an invariant?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+    // TODO: 6 in total, right?
+  };
+  const sharedData = {
+    /** @type {MapStore<string, unknown>} */
+    bySlot: makeScalarMap(),
+    /** @type {MapStore<unknown, string>} */
+    byVal: makeScalarMap(),
+  };
+
+  const kindOf = slot => Object.keys(myData).find(k => slot.startsWith(k));
```
turadg: useful helper above too. can by parameterized with `myData` (Map<Kind, Table>)

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+     * @param {string} iface
+     */
+    fromBoard: (slot, iface) => {
+      if (sharedData.bySlot.has(slot)) {
+        return sharedData.bySlot.get(slot);
+      }
+      const kind = kindOf(slot);
+      if (kind) {
+        assert.fail(X`bad shared slot ${q(slot)}`);
+      }
```
turadg: I would expect:
```suggestion
      assert(!kind, X`bad shared slot ${q(slot)}`);
```
But does the `.fail` have a different signal?

I'm not clear on how it could get to this state. Is this one of those asserts that should impossible by design and is a safeguard here? If so, the message should say so.

If not, could the error help the developer to solve?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+    },
+
+    /**
+     * @param {string} slot
+     * @param {string} iface
+     */
+    fromMyWallet: (slot, iface) => {
+      const kind = kindOf(slot);
+      const key = kind ? slot.slice(kind.length + 1) : slot;
+      const table = kind ? myData[kind] : sharedData;
```
turadg: why does "MyWallet" include the board?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * @param {string} iface
+ * @param {{
+ *   applyMethod: (target: unknown, method: string | symbol, args: unknown[]) => void,
+ *   applyFunction: (target: unknown, args: unknown[]) => void,
+ * }} handler
+ */
+const makePresence = (iface, handler) => {
+  let it;
```
turadg: what's an `it`? okay with the code being terse if the return type is defined

packages/wallet/api/src/marshal-contexts.js
dckc: The slot type is generic. I can make this code more clear about the type we're using here, but we're not importing it.

turadg: Ah, gotcha. It's a generic (template) parameter of `CapData`, defined by whatever the marshaller is doing. So here's a suggestion,
```
/**
  * @typedef {string} WalletSlot
  *  
  * Slot defined as `${kind}:${id}`
  */
```

Then whenever the parameter is `slot` (typed `WalletSlot`) the reader can command click to learn what the commitment is. Even better if the `valToSlot` functions have it as the return type.

packages/wallet/api/src/marshal-contexts.js
dckc: yes; it's all on the board. `Data` seems off too; you can't put a `CopyRecord` nor `CopyList` on the board; I think they have to be keys in the `@agoric/store` sense. Let's try `boardObjects`.

packages/wallet/api/src/marshal-contexts.js
dckc: so that we can recognize brands in the wallet as being identical to brands from the AMM (that's how we got started with this whole mess :)

packages/wallet/api/src/marshal-contexts.js
dckc: this went away when refactoring for type-safety

packages/wallet/api/src/marshal-contexts.js
dckc: Good idea; along these lines, I refactored the whole thing for type safety.

packages/wallet/api/src/marshal-contexts.js
dckc: Close, but the default behavior would not expose it.

I think the description of this PR makes a pretty good docstring here.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+const isBoardId = specimen => {
+  return typeof specimen === 'string' && !!specimen.match(/board[^:]/);
+};
+
+/**
+ * When marshaling a purse, payment, etc. we partition the slots
+ * using prefixes.
+ *
+ * @template T
+ * @typedef {`${string & keyof T}:${Digits}`} WalletSlot<T>
```
turadg: bravo on the string template type. non-blocking suggestion to help document how this works,
```suggestion
 * @template {Record<string, IdTable<*,*>} T
 * @typedef {`${keyof T}:${Digits}`} WalletSlot<T>
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+
+/**
+ * @template T
+ * @param {T} _tables
+ * @param {string & keyof T} kind
+ * @param {number} id
+ * @returns {WalletSlot<T>}
+ */
+const makeWalletSlot = (_tables, kind, id) => {
+  const digits = /** @type {Digits} */ (`${id}`);
```
turadg, line 43: consider asserting `id >= 1` for the digits

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+ * @template T
+ * @typedef {`${string & keyof T}:${Digits}`} WalletSlot<T>
+ */
+/**
+ * @template {string} K
+ * @typedef {`${K}:${Digits}`} KindSlot<K>
+ */
+
+/**
+ * @template T
```
turadg: for documentation purposes,
```suggestion
 * @template {Record<string, IdTable<*,*>} T
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+ * @param {T} record
+ * @param {(value: string, index: number, obj: string[]) => boolean} predicate
+ * @returns {string & keyof T | undefined}
+ */
+const findKey = (record, predicate) => {
+  const key = Object.keys(record).find(predicate);
+  // @ts-expect-error clearly keys(r).find() is keyof T
```
turadg: well, `keyof T | undefined`. the jsdoc is right but this comment is loose

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+ */
+/**
+ * @template {string} K
+ * @typedef {`${K}:${Digits}`} KindSlot<K>
+ */
+
+/**
+ * @template T
+ * @param {T} _tables
+ * @param {string & keyof T} kind
```
turadg, line 38: I think the above would also simplify this,
```suggestion
 * @param {keyof T} kind
```

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+const makeWalletSlot = (_tables, kind, id) => {
+  const digits = /** @type {Digits} */ (`${id}`);
+  return `${kind}:${digits}`;
+};
+
+/**
+ * @template T
+ * @param {T} record
+ * @param {(value: string, index: number, obj: string[]) => boolean} predicate
+ * @returns {string & keyof T | undefined}
```
turadg, line 51: ditto on above suggestions about typing the template

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -0,0 +1,362 @@
...
+ * @returns {{ kind: undefined | string & keyof T, id: number }}
+ */
+const parseWalletSlot = (tables, slot) => {
+  const kind = findKey(tables, k => slot.startsWith(`${k}:`));
+  const id = kind ? Number(slot.slice(kind.length + 1)) : NaN;
+  return { kind, id };
+};
+
+/**
+ * Since KindSlots always include a colon and BoardIds never do,
```
turadg, line 71: 👏

packages/wallet/api/test/test-wallet-marshal.js
```diff
@@ -0,0 +1,166 @@
...
+
+  t.deepEqual(context.unserialize(actual), invitationAmount);
+
+  const myPayment = /** @type {Payment} */ (
+    Far('payment', { getAllegedBrand: () => assert.fail('no impl') })
+  );
+  context.savePaymentActions(myPayment);
+  const cap2 = context.serialize(myPayment);
+  t.deepEqual(cap2, {
+    body: JSON.stringify({
```
turadg, line 158: when these don't match, it'll be a string diff.  consider comparing the object with `JSON.parse(cap2)`.

If not comparing objects then it should be `t.is`

packages/wallet/api/test/test-middleware.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const capData = ctx.fromMyWallet.serialize(harden([...msgs]));
+
+  t.deepEqual(capData, {
+    body: '[["applyMethod",{"@qclass":"slot","iface":"Alleged: purse.actions","index":0},"transfer",[1]]]',
+    slots: ['purse:1'],
+  });
+});
+
+//   {
```
turadg, line 72: dead code

turadg (commented):
Looking good.

I don't know how to assess this:
> there's a risk that chainStorage data from the AMM could try to refer to a purse using a slots such as purse:123; such a reference must fail.

Can a test be added for that?

packages/wallet/api/test/test-wallet-marshal.js
dckc: The test is supposed to be at the CapData level, where `body` is a string. Writing the expected `body` using `JSON.stringify()` makes it easier to read. I suppose I could add a comment...

Meanwhile, your comment prompts me to think about this a bit more carefully... In the API, `slots` are compared by identity, so `t.deepEqual()` is only correct because we happen to be using primitives (strings). In nearby designs, the slots are `{ kind: 'purse', id: 123}` objects where it's really easy to have things that pass `t.deepEqual()` but not `t.is()`.

dckc:
> I don't know how to assess this:
> 
> > there's a risk that chainStorage data from the AMM could try to refer to a purse using a slots such as purse:123; such a reference must fail.
> 
> Can a test be added for that?

It's in there:

```
test('makeImportContext preserves identity across AMM and wallet', t => {
...
  t.throws(
    () => context.fromBoard.unserialize(walletCapData),
    { message: /bad board slot/ },
    'AMM cannot refer to purses',
  );
```

_p.s. "AMM cannot forge references to purses" would be a better phrasing._

packages/wallet/api/test/test-middleware.js
dckc: Good point; I guess I'll think this over some more.

packages/wallet/api/src/marshal-contexts.js
dckc: Thanks; I'll give that a try.

Earlier I explored similar things without luck and punted.

turadg (approved):
Approving now. (No automerge label so you'll be able to try the TS tweak before it merges.)

packages/wallet/api/test/test-middleware.js
dckc: I'd like to come back to this in a later PR.

packages/wallet/api/src/marshal-contexts.js
dckc: I integrated most of that suggestion. Constraining `T` that way helps not only documentation but also got rid of one or two `@ts-expect-error`s.

But getting rid of `string &` doesn't seem to fly.

PR #5887, opened 2022-08-03
5605 deprecate makeVaultInvitation

turadg:
part of: #5605 

## Description

The `makeLoanInvitation` and `makeVaultInvitation` methods on vaultDirector are replaced by `makeVaultInvitation` on vaultManager. This removes the old ones.

It also fixes a bug discovered in doing so, which is that the replacement method didn't check for `minInitialDebt`.

### Security Considerations

Improved, by parameter checking and reduced API surface.

Document breaking change.

### Testing Considerations

Updated tests

### DIFF
--- a/packages/inter-protocol/src/vaultFactory/vaultDirector.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultDirector.js
@@ -8,7 +8,6 @@ import { E } from '@endo/eventual-send';
 
 import { fit, keyEQ, M, makeScalarMap } from '@agoric/store';
 import {
-  assertProposalShape,
   getAmountIn,
   getAmountOut,
 } from '@agoric/zoe/src/contractSupport/index.js';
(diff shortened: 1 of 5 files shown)


samsiegart (commented):
I know we discussed using the ephemeral key to get the vault manager on the dapp side, but I didn't realize we were already using this top-level public facet method to make the invitations.

If we kept this method it would save an extra approval flow and round trip (maybe gas as well?) for the user. I'm not sure if that outweighs the technical reasons for removing this though.

I'd vote to keep this around until we've explored what the UX around ephemeral keys is like and it might be easier to make a product decision. But if @dtribble wants to veto this suggestion that's okay with me.

turadg:
> If we kept this method it would save an extra approval flow and round trip (maybe gas as well?) for the user.

IIUC message pipelining will make it so these two operations happen in one transaction. No?

> I'm not sure if that outweighs the technical reasons for removing this though.

> keep this around until we've explored what the UX around ephemeral keys is like and it might be easier to make a product decision

Okay. We need a decision before e2e but let's get aligned on the product questions and priorities before merging.

samsiegart:
> IIUC message pipelining will make it so these two operations happen in one transaction. No?

I think at minimum there will be an extra approval to create the ephemeral key, and then an approval to do the actual spend action for the offer once you have the invitation.

turadg:
Closing because very stale, will basically need to be redone.

The fix for checking minInitialDebt is already in master. `makeVaultInvitation` is not `@deprecated` and there are no outstanding deprecations in inter-protocol

PR #5889, opened 2022-08-03
5605 remove getCollaterals from vaultFactory

turadg:
part of: https://github.com/Agoric/agoric-sdk/issues/5605

## Description

`getCollaterals` was designed before our streams for governance and metrics. Users of that data should use those stream because they can be cached and pushed. This removes the obsolete way to get information about collaterals.

### Security Considerations

--

Marked as BREAKING CHANGE.

### Testing Considerations

CI

- [ ] loadgen passing without `getCollaterals`


samsiegart (approved):
🔥 🔥 🔥  :shipit:

turadg:
obsolete

PR #5890, opened 2022-08-04
Implement swingset-runner test to validate vat process failure behavior

### LINKED ISSUE #2958: test our policy to respond to a vat worker subprocess crashing

When a vat worker is running in a child subprocess, such as with XS, we introduce a new failure mode: the child process might die. This could happen because of an engine fault (a crash is the best case outcome for a buffer overflow or memory corruption bug, because if the process survives it could probably be compromised in some sneaky way). It might also happen because of resource exhaustion, like too much RAM being allocated, if our metering tools aren't able to catch that first. It might also happen because of something outside the scope of swingset entirely: the host computer is being rebooted, so it's killing off all processes (usually with a polite+catchable SIGINT, followed a few seconds later by an uncatchable SIGKILL). If the kernel happens to politely SIGINT the worker process first, the kernel might observe the worker's termination fast enough to react and record any state changes before being killed itself.

We need to decide what the kernel does when the worker process exits unexpectedly. Our options are:

* terminate the kernel itself (without committing state)
* treat it as equivalent to a vat fault (like a c-list violation, or a metering fault not saved by a Keeper), and terminate the vat
* retry the crank with a new worker

We also need a way to react to a worker process which dies outside of the context of a delivery. This could happen if the host computer is low on RAM, and the OOM (Out-Of-Memory) Killer sets its sights on the worker process. It could also happen if an incautious sysadmin kills the wrong process.

## Description of the Design

We need to enhance the vat-manager API to include worker-process crash as part of the `VatDeliveryResult` structure. It should indicate the exit code of the process, including a unix signal number (if running on unix), perhaps pre-processed to indicate what category of crash has happened. We also need some kernel options to select the policy (if we decide to support more than one).

As for the policy selection itself: the choice to terminate the vat is (highly) visible to consensus, so if we're operating in a consensus mode, we need to make sure that all validators make the same choice. We can't afford to let random host-computer reboots make one validator think the vat has died, while all the other validators think it's just fine.

### Consensus Mode

So if we're in consensus mode, I think the best option is to immediately terminate the kernel. The rule will be: if the worker reports a metering fault, terminate the vat, but if the worker dies and we don't know why, terminate the kernel.

If the worker dies because the host computer is being rebooted, this will play out the same way as if the kernel process had been killed first: when the reboot finishes, the kernel will come back up, and it will pick up where it left off, launching a new worker process for that vat (as well as all other active vats).

If the worker died because of an uncaught resource exhaustion problem, then the outcome depends upon how serious it is. If it's only affecting one validator, and it's somehow random (maybe swap space ran low on that computer), then rebooting the kernel might help, and the validator is only out a couple of blocks. If it persists, that validator will be unable to produce more blocks, and they'll get jailed, which is exactly the right way to evict a validator that isn't powerful enough to support the chain.

If the problem affects lots of validators, then it's much more likely that a fix is required in our software. If it affects more than a third of the voting power, the remainder will not be able to produce new blocks, and the chain will halt. The speed with which the chain comes to a halt will depend upon what fraction of the validators are affected. The worst case is for slightly fewer than one third of the voting power to go offline, in that it causes the most churn in the validator set. Ideally all faults are common-mode: they affect everybody or (nearly) nobody. Halting the chain is probably better than slashing/jailing a third of the validators.

It might be interesting to apply a `ulimit` size limit to the worker process, set to the same value across all validators. This might improve the odds that an uncaught memory-exhaustion attack affects all/most validators the same way.

### Non-Consensus Mode

If we're not running in a consensus mode, the kernel can do whatever it likes. The only party it must answer to is its future self.

I don't have strong feelings about what the ideal policy would be. I like the idea of using `ulimit` as a sensor for resource exhaustion attacks, but I also don't want host an unfortunately-sequenced computer reboot to make us think a vat should be terminated permanently. Fortunately I think we can use the process exit code (signal number) to distinguish between the two: when automatic stack expansion fails the process gets a SIGSEGV, if the engine tries and fails to allocate memory then it probably self-terminates with SIGABRT (need to check this), but if system shutdown kills the process we'd see SIGINT or SIGKILL or SIGTERM or something. The `setrlimit(3)` man page says CPU exhaustion is signalled (on linux) with SIGXCPU followed by SIGKILL.

Assuming we can distinguish between them reliably, it might make sense for the kernel to retry a crashed worker once (per kernel process), maybe after a multi-second timeout, after which we declare the vat terminated. That should be proof against reboot-time shutdown sequencing, and would also tolerate sysadmins fumble-fingering a `kill` command that hits the wrong process.

But I think it'd also be safe (and easier to implement) to match what we do in consensus mode: any unexplained worker death causes the kernel process to halt immediately (after logging what happened, of course, so the operator knows why).

## Consensus Considerations

As described above, in consensus mode, we must not terminate vats without consensus-based evidence of fault. We don't wat to let an attacker cause a small number of validators to drop out of consensus. Metering is supposed to catch this in a deterministic fashion, but if that fails, we'd like to improve the odds of keeping the chain intact (most validators proceeding, or most validators halting).

## Security Considerations

Several memory-corruption are somewhat probabilistic: the attacker must submit a number of probes to learn some secret value (e.g. ASLR secrets). These attacks will appear as crashes of the first N-1 probes, followed by a silent compromise on the subsequent N probes. In some cases, the secret they're trying to learn is reset by the process being reset, or the computer being rebooted. The attacker's mean time to compromise is multiplied by the time between probe attempts.

If we re-try a crashed worker immediately, and did not limit the number of times we retried, we're giving an attacker (who is trying to probe the worker's secrets with their vat delivery) the best odds of success. If we slow down the retries (randomized exponential backoff is a popular approach), or put a hard limit on retries, then we reduce their chances considerably.

If we use something like `seccomp(2)` (#2386) then the same considerations apply even more directly.

## Test Plan

The most reliable test would be to launch a real worker process, then kill it in various ways, and assert that the kernel reacts in the selected way.

We should also have direct tests of the kernel policy, where we exercise the code with a vat-manager delivery result that has been extended to describe worker crashes.

cc @michaelfig @dckc @erights

### LINKED ISSUE #5480: unexpected worker exit doesn't cause proper kernel panic
## Describe the bug

@michaelfig did an experiment where he used SIGABRT to kill a worker (`v2:http`, in particular). He observed at least one unhelpful error message (the slog code complaining that the vat did a syscall while in the IDLE state), and then saw that the kernel did not panic as expected.

One message he saw was:
```
UnhandledPromiseRejectionWarning: vat (a string) in (a string), not (a string): (a string)
(node:54) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 21
4)
2022-05-31T20:50:08.842Z SwingSet: kernel: error in kernel.deliver: (RangeError#137)
RangeError#137: vat v2 in delivery , not idle : reentrant delivery?

  at assertOldState (.../swingset-vat/src/kernel/slogger.js:123:8)
  at Object.delivery (.../swingset-vat/src/kernel/slogger.js:154:1)
  at Object.deliverToVat (.../swingset-vat/src/kernel/vat-warehouse.js:275:21)
  at async deliverAndLogToVat (.../swingset-vat/src/kernel/kernel.js:383:22)
  at async deliverRunQueueEvent (.../swingset-vat/src/kernel/kernel.js:938:12)
  at async processDeliveryMessage (.../swingset-vat/src/kernel/kernel.js:975:14)
  at async Object.run (.../swingset-vat/src/kernel/kernel.js:1532:19)
  at async crankScheduler (packages/cosmic-swingset/src/kernel-stats.js:315:5)
  at async processKernel (packages/solo/src/start.js:261:5)
  at async deliverInboundCommand (packages/solo/src/start.js:293:7)
```

another was:
```
2022-05-31T20:48:34.309Z SwingSet: kernel: error in kernel.deliver: (ExitSignal#1)
ExitSignal#1: v2:http exited due to signal SIGABRT

  at new ErrorSignal (packages/xsnap/api.js:37:5)
  at ChildProcess.<anonymous> (packages/xsnap/src/xsnap.js:116:22)
  at ChildProcess.emit (node:events:527:28)
  at ChildProcess.emit (node:domain:475:12)
```

I think we're handling unexpected worker death during a delivery, but not while outside of a delivery.

The slogger tracks the state of each vat, and asserts that e.g. syscalls only happen while the vat is in the DELIVERY state, not the IDLE state. (This might be a bit aggressive, especially because the slogger should not be able to crash the kernel, but it *is* a useful diagnostic to know that the vat is doing something out-of-turn).

I suspect the first error message is happens because the code reading from the worker pipe gets confused: when the worker exits, the pipe becomes "readable", but the next read will be EOF or an empty string or will throw an error, or something. I suspect that the empty string is somehow parsed enough to be treated as a syscall, at least enough to trigger the slogging code, which then complains that we aren't in the right state.

The `manager-subprocess-xsnap.js` code has some sort of async iterator loop that waits for data from the worker subprocess. That loop has some way to signal an error, like "child exited". We need to give that manager code a way to panic the kernel when it sees the child exit (as long as we weren't trying to deliberately exit the child at that time, i.e. we were paging the worker out).

Then we want to make sure we don't trigger the slog syscall pathway if/when the "data is available from the child" code gets triggered.

# Testing

I think we want a unit test that can start a small kernel, with a single `xsnap` worker, figure out the PID of the worker, kill it with a SIGABRT, then attempt to do a `controller.run()` and confirm that it rejects. We should manually look at the output of this test and confirm that it doesn't emit any spurious slog errors.

FUDCo:
This PR adds a `swingset-runner` demo for testing vat process failure behavior.  This requires manual testing since it is prohibitively difficult (or, at least, more difficult than we currently care to expend resources on) to automate.

The issue to which this PR is responsive contains a detailed writeup of the use of this manual test and the conclusions drawn from it.

Closes #2958
Closes #5480

### DIFF
--- a/packages/swingset-runner/demo/workerDeathTest/bootstrap.js
+++ b/packages/swingset-runner/demo/workerDeathTest/bootstrap.js
@@ -1,0 +1,14 @@
+import { E } from '@endo/eventual-send';
+import { Far } from '@endo/marshal';
+
+export function buildRootObject() {
+  console.log(`bootstrap vat initializing`);
+  return Far('root', {
+    async bootstrap(vats) {
+      const howMuch = 100_000_000;
+      console.log(`bootstrap vat invokes busy vat for ${howMuch} steps`);
+      await E(vats.busy).doStuff(howMuch);
+      await E(vats.idle).doNothing();
+    },
+  });
+}
--- a/packages/swingset-runner/demo/workerDeathTest/swingset.json
+++ b/packages/swingset-runner/demo/workerDeathTest/swingset.json
@@ -1,0 +1,15 @@
+{
+  "bootstrap": "bootstrap",
+  "defaultManagerType": "xs-worker",
+  "vats": {
+    "bootstrap": {
+      "sourceSpec": "bootstrap.js"
+    },
+    "busy": {
+      "sourceSpec": "vat-busy.js"
+    },
+    "idle": {
+      "sourceSpec": "vat-idle.js"
+    }
+  }
+}
--- a/packages/swingset-runner/demo/workerDeathTest/vat-busy.js
+++ b/packages/swingset-runner/demo/workerDeathTest/vat-busy.js
@@ -1,0 +1,22 @@
+import { Far } from '@endo/marshal';
+
+export function buildRootObject(_vatPowers, _vatParameters, baggage) {
+  console.log(`busy vat initializing`);
+  baggage.init('someKey', 'hello');
+  return Far('root', {
+    doStuff(howMuch) {
+      let n = 2.0;
+      for (let i = 0; i <= howMuch; i += 1) {
+        n = Math.sqrt(n) * Math.sqrt(n); // introduce some computational drag
+        if (i % 10_000_000 === 0) {
+          let bv = 'no baggage';
+          if (baggage) {
+            bv = baggage.get('someKey');
+          }
+          console.log(`busy vat doing step ${i} of ${howMuch}, bv=${bv}`);
+        }
+      }
+      console.log(`busy vat finished ${howMuch} steps`);
+    },
+  });
+}
--- a/packages/swingset-runner/demo/workerDeathTest/vat-idle.js
+++ b/packages/swingset-runner/demo/workerDeathTest/vat-idle.js
@@ -1,0 +1,8 @@
+import { Far } from '@endo/marshal';
+
+export function buildRootObject() {
+  console.log(`idle vat initializing`);
+  return Far('root', {
+    doNothing() {},
+  });
+}
--- a/packages/swingset-runner/src/main.js
+++ b/packages/swingset-runner/src/main.js
@@ -339,10 +339,12 @@ export async function main() {
     delete config.loopboxSenders;
     deviceEndowments.loopbox = { ...loopboxEndowments };
   }
-  if (useXS) {
-    config.defaultManagerType = 'xs-worker';
-  } else {
-    config.defaultManagerType = 'local';
+  if (!config.defaultManagerType) {
+    if (useXS) {
+      config.defaultManagerType = 'xs-worker';
+    } else {
+      config.defaultManagerType = 'local';
+    }
   }
   if (launchIndirectly) {
     config = generateIndirectConfig(config);


warner (approved):
Looks good. The one thing I might suggest is to add a comment with a link to your writeup (https://github.com/Agoric/agoric-sdk/issues/2958#issuecomment-1204614797) , so future archaeologists can know what this is for and roughly how to use it.

PR #5891, opened 2022-08-04
make Zoe's instanceRecord durable

Chris-Hibbert:
refs: #4383

## Description

On the path to durable Zoe.

See #5879 

### Security Considerations

Maintain Zoe invariants

None.

### Testing Considerations

Ensure existing tests continue to pass.


packages/zoe/src/instanceRecordStorage.js
```diff
@@ -39,7 +42,8 @@ export const makeInstanceRecordStorage = () => {
     );
 
     assertInstantiated();
-    instanceRecord = {
+    const instanceRecord = baggage.get('instanceRecord');
+    const nextInstanceRecord = {
```
erights:
```suggestion
    const nextInstanceRecord = harden({
```

This should always have had a `harden` on it. Not sure why it didn't. All our stores will only store passables. Unfortunately, to ease the transition, they harden their arguments before checking if they're passable, thereby masking bugs like the missing harden here. Ideally, the stores would only check that things were passable, rejecting non-hardened things, and objects would be hardened as early a possible, usually when first created, as here.

packages/zoe/src/instanceRecordStorage.js
```diff
@@ -53,46 +57,49 @@ export const makeInstanceRecordStorage = () => {
...
       },
     };
+    baggage.set('instanceRecord', nextInstanceRecord);
   };
 
   /** @type {GetInstanceRecord} */
   const getInstanceRecord = () => {
     assertInstantiated();
-    return harden(instanceRecord);
+    return harden(baggage.get('instanceRecord'));
```
erights:
```suggestion
    return baggage.get('instanceRecord'));
```

If it was in baggage, then it was already passable.

packages/zoe/src/zoeService/zoeStorageManager.js
```diff
@@ -120,6 +120,7 @@ export const makeZoeStorageManager = (
     uncleanIssuerKeywordRecord,
     instance,
     contractBundleCap,
+    instanceBaggage,
```
erights: When the baggage parameter is not optional, all call sites need to change anyway, so better to make the baggage parameter be first.

erights (approved):
LGTM

erights:
All three CI failures are stores of non-durables into durables. Do you understand each of these?

Chris-Hibbert:
> All three CI failures are stores of non-durables into durables. Do you understand each of these?

One of them is in bootstrap. parameters coming from `consume` clauses are always promises and should be awaited. They weren't being waited for, and some of the promises were ending up in `terms`, which was forgiving of that when it didn't have to be durable. Now that `instanceRecord` wants to  be durable, `terms` has to be durable.

One of the others seems to be `terms`, but I haven't found the source yet. I dunno so far on the third.

packages/zoe/src/instanceRecordStorage.js
Chris-Hibbert: done

packages/zoe/src/instanceRecordStorage.js
Chris-Hibbert: done

packages/zoe/src/zoeService/zoeStorageManager.js
Chris-Hibbert: done

turadg (commented):
Why all the new awaiting? My understanding from @dckc is that the proposal system is designed to be eventual.

Chris-Hibbert:
> One of the others seems to be terms, but I haven't found the source yet. I dunno so far on the third.

`FeeMintAccess` and `feeMint` needed to be made durable for Zoe.
Moved `baggage` to be the first param of `makeZoeInstanceStorageManager`

> Why all the new awaiting? 

Everything supplied in `consumes` in bootstrap arrives as a Promise. Anything that ends up in terms has to be resolved since terms must be durable so Zoe can support recovering ZCF.  This should not be a problem, since terms already had to be comparable.  (Promises are not comparable.)

packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -36,7 +36,7 @@ export const startPSM = async (
       consume: { contractGovernor, psm: psmInstallP },
     },
     instance: {
-      consume: { economicCommittee },
+      consume: { economicCommittee: economicCommitteeP },
```
turadg: the readme calls out `startPSM` as exemplary, to be emulated. I think this new deeplyFulfilled technique is the state of the art so please consider updating `startPSM` to use it. If not maybe I will later. cc @dckc

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -29,6 +30,9 @@ const SECONDS_PER_DAY = 24n * SECONDS_PER_HOUR;
 const BASIS_POINTS = 10_000n;
 const MILLI = 1_000_000n;
 
+const hardenDeeplyFullfilled = async expression =>
+  deeplyFulfilled(harden(expression));
```
turadg: This barely saves any typing and the other files use the two functions. My best guess why this is defined here is so the contets of `ammGovernorTerms` don't need to be reindented. But if so then packages/inter-protocol/src/vaultFactory/params.js wouldn't be re-indented.

Please consider using this helper there or removing it entirely

packages/pegasus/src/proposals/core-proposal.js
```diff
@@ -40,6 +40,8 @@ export const startPegasus = async ({
     produce: { [CONTRACT_NAME]: produceInstance },
   },
 }) => {
+  [board, namesByAddress] = await Promise.all([board, namesByAddress]);
```
turadg: non-blocking but I thought you wanted to avoid type changes on the same symbol

packages/vats/src/core/basic-behaviors.js
```diff
@@ -235,6 +235,11 @@ export const makeClientBanks = async ({
   },
 }) => {
   const STORAGE_PATH = 'wallet';
+  [agoricNames, board, namesByAddress] = await Promise.all([
```
turadg: ditto type change

turadg (commented):
No blockers on the proposals and InterP stuff. Not reviewing all the Zoe parts but will if you want.

packages/vats/src/core/basic-behaviors.js
Chris-Hibbert: Yes, I shoud call`hardenDeeplyFullfilled` instead.

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: I agree it barely saves typing, but it does save a layer of indentation, and for `terms` we should always use both together. I've moved it to Zoe/contractSupport, and renamed it to `deeplyFulfillTerms`.

packages/vats/src/core/basic-behaviors.js
```diff
@@ -240,10 +243,14 @@ export const makeClientBanks = async ({
     makeStorageNode(chainStorage, STORAGE_PATH),
     bridgeManagerP,
   ]);
+
+  const terms = await deeplyFulfillTerms(
+    harden({ agoricNames, namesByAddress, board }),
```
turadg: this harden is superfluous

turadg (commented):
The `deeplyFulfillTerms` is a definite improvement. I see several expanded Promise.all where it would help but that needn't block this PR.

I see this is Draft. If it may be a while, maybe we can get the proposal improvement in with another PR. Or start one in parallel scoped to adopting it comprehensively.

erights (changes requested):
Reverse introduction of `deeplyFulfillTerms`. See https://github.com/Agoric/agoric-sdk/pull/5930

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -853,26 +881,28 @@ export const startStakeFactory = async (
...
+    governedContractInstallation: installations.stakeFactory,
+    governed: harden({
+      terms: stakeFactoryTerms,
+      issuerKeywordRecord: { Stake: bldIssuer },
+      privateArgs: {
+        feeMintAccess,
+        initialPoserInvitation,
+        lienBridge,
+        storageNode,
+        marshaller,
```
erights: Re line 891 above

Terms are not private. Perhaps I am missing context, but it seems wrong that something called somethingTerms contains something called privateArgs. Doesn't this cause the allegedly private args to be exposed to anyone with access to these terms?

erights (changes requested):

packages/inter-protocol/src/proposals/startPSM.js
Chris-Hibbert: done.

packages/pegasus/src/proposals/core-proposal.js
Chris-Hibbert: vestigial.  Thanks for pointing it out.

Oh, this seems to be gone  now.  ?!?

packages/vats/src/core/basic-behaviors.js
Chris-Hibbert: thanks.

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: agreed.   This is a bigger issue than this PR.    I'll have to think about whether this is somehow safe because it's governance. I suspect the answer is 'no'.

packages/governance/test/swingsetTests/contractGovernor/test-governor.js
```diff
@@ -167,8 +167,8 @@ test.serial('brokenUpdateStart', async t => {
     // TODO: allow either message
     // 'vote rejected outcome: Error: (an object) was not a live payment for brand (an object). It could be a used-up payment, a payment for another brand, or it might not be a payment at all.',
     // 'update failed: Error: (an object) was not a live payment for brand (an object). It could be a used-up payment, a payment for another brand, or it might not be a payment at all.',
-    'vote rejected outcome: Error: A Zoe invitation is required, not "[Alleged: Zoe Invitation payment]"',
-    'update failed: Error: A Zoe invitation is required, not "[Alleged: Zoe Invitation payment]"',
+    'vote rejected outcome: Error: A Zoe invitation is required, not (an object)',
+    'update failed: Error: A Zoe invitation is required, not (an object)',
```
erights: If you're running with a `prepare-...` setup import suitable for testing, then it should already be setting `errorTaming: 'unsafe'` in which case you should not see these redactions. Is the vat these are coming from run in a non-testing-setup vat?

Chris-Hibbert: It has 
```
import test from 'ava';
```
I've updated it.

Chris-Hibbert: Hmm.   I changed that to
``` javascript
import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
```

expecting that to be what you meant, but that didn't fix it.

erights: Yes, `import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';` is exactly what I had in mind. Separately, we need to understand why that did not fix it, and where the code that generated that error message is running.

But no more on this needs to be done in this PR. Resolving.

packages/pegasus/src/proposals/core-proposal.js
turadg: still there

turadg (approved):

packages/pegasus/src/proposals/core-proposal.js
Chris-Hibbert: I kept looking for it in the other core-proposal.  Thanks for pointing it out again.

erights (approved):

Issue #5886, opened 2022-08-03
does HandledPromise's deep-stack support cause memory problems when used in iterator/generators?


warner:
## What is the Problem Being Solved?

While investigating #5671, @erights identified something to look into. HandledPromise does some clever tricks to retain enough information for a "deep stack" to be displayed in case of an error: when one `E()` call triggers another, an Error object is created (to capture a stack trace) and held internally by the new Promise.

There's a portion of vat-bank which wraps a Notifier in a generator to provide an iterator of new balances:

https://github.com/Agoric/agoric-sdk/blob/15234fa723c1e1a14843eff154b38439b1cdb1f9/packages/vats/src/vat-bank.js#L42-L52

This is wrapped by the virtual purse to eagerly update a local `lastBalance` variable each time the balance changes:

https://github.com/Agoric/agoric-sdk/blob/15234fa723c1e1a14843eff154b38439b1cdb1f9/packages/vats/src/virtual-purse.js#L105-L116

`observeIteration` is defined by our `notifier` package:

https://github.com/Agoric/agoric-sdk/blob/15234fa723c1e1a14843eff154b38439b1cdb1f9/packages/notifier/src/asyncIterableAdaptor.js#L137-L139

and uses `observeIterator`:

https://github.com/Agoric/agoric-sdk/blob/15234fa723c1e1a14843eff154b38439b1cdb1f9/packages/notifier/src/asyncIterableAdaptor.js#L101-L124

This composition probably creates a series of promises over time (one per balance update), but might do so in a fashion that makes each promise look like a causal descendant of the previous one. If so, the deep-stack code might be retaining an unbounded series of Error objects, causing memory usage to grow linearly with the number of updates, when it ought to be flat.

The first step is to write a small test harness that emulates the vat-bank loop, and run it a few million times, while watching the memory usage of the process.

warner:
I'm indeed observing a memory leak when `observeIteration` is following a generator that follows a Notifier. The following setup code:

```js
  const { updater, notifier } = makeNotifierKit(); // Y1

  async function *getBalances() { // Y2
    let uprec = await notifier.getUpdateSince(); // read Y1
    while (uprec.updateCount) {
      yield uprec.value;
      uprec = await notifier.getUpdateSince(uprec.updateCount); // read Y1
    }
  }

  observeIteration(getBalances(), {}); // reads Y2
```

[...]

PR #5892, opened 2022-08-04
fix: patch to suppress turn tracking diagnostics

erights:
refs: #5886

Fixes the above issue for immediate purposes, to unblock everything else. But leaving the issue open until the change, done properly, is done at the endo level and agoric-sdk is updated to use that.

## Description

The track-turns.js file in `@endo/eventual-send` operates behind the scenes, unobservably to code that cannot sense console output, to the hardened JS console to output so-called "deep stacks". When this console shows an error, it shown not only the stack trace of the error itself, but of any errors attached to it through the `assert.note` mechanism. The track-turns mechanism tracks some inter-turn causality, which it reports using this same `assert.note` mechanism. However, when we have a long multi-turn loop, where each turn in the loop causes the next turn in a way that `track-turns` tracks, then we can have unbounded memory usage.

These diagnostics are only needed during development anyway. They should be turned off in production. This PR just turns them off without any option to turn them back on. A later endo release will replace the need for this patch with a mechanism that defaults to having these diagnostics off, but enables them to be turned back on.

### Security Considerations

The memory leak solved by this patch PR was a threat to availability. The deep stacks showing on the console did not violate our threat model for confidentiality, so suppressing it does not address any confidentiality concern. Since the change should be unobservable to any code that cannot sense the console output, it should have no effect on integrity.

AKAIK we have never explained or documented the deep-stack diagnostic information for our users. The closest we come is https://github.com/endojs/endo/blob/master/packages/ses/src/error/README.md#logging-errors . But the deep stacks it talks about are only the inter-vat deep stacks, not the intra-vat inter-turn deep stacks that this PR changes.

Thus, no changes to docs are currently needed. Eventually we do, of course, need to explain and document for our users how to use these deep stacks to help find bugs.

### Testing Considerations

Since it is not supposed to have any observable effect, it is not clear to me how to create an automated test for this. For purposes of this PR, given the evidence we already have at #5886 I consider an adequate test that CI is green, i.e., that we have not broken anything else. I'll leave any further need to test this to the upcoming endo PR that does this "for real".

***Note for reviewers:*** Do you find my justification of lack of specific testing acceptable for this patch PR? If not, what would be a good test?


erights:
A CI failure reveals that suppressing track-turns is not as unobservable as we thought. It now also `harden`s the function results. Thus, even our patch PR must do likewise.

erights:
> AKAIK we have never explained or documented the deep-stack diagnostic information for our users. 

I'm wrong. Deep-stacks are explained, and examples are shown, at
https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md#stackfiltering-options

But since that's on the endo side, fixing it would need to be part of the endo PR anyway.

packages/SwingSet/test/snapshots/test-xsnap-store.js.md
```diff
@@ -12,8 +12,8 @@ Generated by [AVA](https://avajs.dev).
 
 > after SES boot - sensitive to SES-shim, XS, and supervisor
 
-    'f338e2ab900ed621f79cc39bbcd8929d64020f116fccd537fe8d77fc8f53c0f6'
+    'bab8eb405615d1a174c97b4a1443d699e6b983193797e9b727132a5466d1ee13'
 
 > after use of harden() - sensitive to SES-shim, XS, and supervisor
 
-    'da32888101e6f155aa9b55623f338cc51e704b1e455c2e12c975e1525c754899'
+    '179aed359604c84f412d737d10ccccea22a9db586025011097c2127cae591a4b'
```
erights: Since this PR patches part of endo, the change of hash is not unexpected.

packages/SwingSet/test/vat-admin/test-create-vat.js
```diff
@@ -258,8 +258,31 @@ function findRefs(kvStore, koid) {
   const refcount = refcountString.split(',').map(Number);
   return { refs, refcount };
 }
-
-test('createVat holds refcount', async t => {
+/**
+ * TODO Figure out why #5892 causes a refcount difference, and
+ * unskip this.
```
erights, line 263: @FUDCo @warner Please see scary refcount change. Thanks.

packages/vats/test/test-boot.js
```diff
@@ -37,45 +37,47 @@ const vatRoots = {
...
-        addRemote: () => ({}),
-      })
-    ),
-    comms: Far('comms', {
-      addRemote: noop,
-      addEgress: noop,
-      addIngress: async () => ({
-        getConfiguration: () => ({ _: 'client configuration' }),
+const makeMock = t =>
+  harden({
```
erights, line 41: With the inappropriate `harden` in track-turns added back in at https://github.com/Agoric/agoric-sdk/pull/5892/commits/346a60a6c8cda47962e8b81b428e02bd45d09f31 , the `harden`s added to `test-boot` are no longer strictly needed into for the test to run successfully. But their absence is a bug that the `harden` in trackTurns was masking. Going through a phase in which that `harden` was removed revealed these bugs. I want to include them in this PR because this PR also puts that `harden` back in, for compatibility, but at the cost of continuing to mask bugs like this.

Curiously, these missing hardens was the *only* bug revealed during the phase in which the patched trackTurns did not harden. With only these changes, we were green under CI prior to https://github.com/Agoric/agoric-sdk/pull/5892/commits/346a60a6c8cda47962e8b81b428e02bd45d09f31

packages/SwingSet/test/vat-admin/test-create-vat.js
warner: (This test exercises the refcount maintained, by the kernel, on any objects in the `vatParameters` that are passed through a create-vat request. We call `E(vatAdminService).createVat(bundleID, { vatParameters: [ object, devicenode] })`, and look at the refcount on that exported `object`)

vat-vat-admin receives the following deliveries:
* `getBundleCap()`
* `bringOutYourDead` (1)
* `createVat()`
* `bringOutYourDead` (2)
* `newVatCallback()`
* `bringOutYourDead` (3)
* `dropExports`, then `retireExports`

The `object` reference is imported into vat-vat-admin as an argument of the `createVat()` call.

My first pass at analysis shows that the `object` reference was previously dropped during the (3) BOYD call, just after `newVatCallback`, whereas in this PR's branch, it is being dropped during the (2) BOYD call, just after `createVat`. This implies that vat-vat-admin was retaining a reference to the `object` Presence longer in the previous version.

I'm still investigating what might have changed here.

warner: I don't see any reason why vat-vat-admin would retain a reference to `vatParameters` after the initial `createVat()` delivery: they are passed into a device call during `createVat()` and then not retained anywhere. So it seems like the new behavior is the correct one, and the previous retention was the surprise. I'll keep digging.

warner: I found that changing `track-turns.js` to hoist the "add rejection node" closure to its own top-level function (thus removing `funcs` and `args` from its lexical scope) caused the same test-create-vat.js failure as did this PR (disabling track-turns entirely). Given how `trackTurns` is used by `handled-promise.js`, I believe that holding `funcs` will cause swingset to hold onto unserialized delivery arguments (i.e. Presences) longer than expected, and I think that's what was causing this test to change behavior.

I'm writing up an Endo ticket to address this problem. In the meantime, I think we should land this PR, and then fix the swingset test to expect the (now-understood-to-be-correct) ordering of events. I'll take care of that (and unskip this test case) after this PR lands.

warner (approved):
As mentioned, I think we can fix the swingset issue after landing this PR.

packages/SwingSet/test/vat-admin/test-create-vat.js
warner: https://github.com/endojs/endo/issues/1245 is the Endo ticket on the object retention

PR #5893, opened 2022-08-04
test(swingset): exercise vat-vattp upgrade

warner:
After sending mailbox messages in both directions, we
restart (null-upgrade) vat-vattp, and then send another pair of
messages through it. This exercises the mailbox side of vattp:

* the previously-exported transmitter object still works
* the outbound seqnum counter was persistent
* inbound messages are routed to the previously-imported receiver
* the inbound ack number was persistent

refs #5667


FUDCo (approved):
Marvy.

gibson042 (approved):

PR #5895, opened 2022-08-04
ci(restore-node): verify no `yarn.lock` change after `yarn install`

michaelfig:
## Description

Ensure the `yarn.lock` doesn't change while `yarn install`ing the SDK.  This prevents committing a `yarn.lock` that is out-of-date wrt the actual dependencies.


kriskowal (approved):

warner (approved):
awesome, thanks.. we need to land a fix to `yarn.lock` before landing this, right?

michaelfig:
> we need to land a fix to `yarn.lock` before landing this, right?

No fix necessary, I think it was already integrated.

PR #5897, opened 2022-08-05
fix(swingset): re-enable refcount test

warner:
The 'trackTurns()' debugging feature of eventual-send (which records
deep-stack information) had an unexpected side-effect: HandledPromise
invocation arguments were retained until the target's result promise
was resolved. Details in https://github.com/endojs/endo/issues/1245 .

This appears to be the cause of the unexplained createVat() retention
of the vatParameters presences until newVatCallback() allowed the
result promise to resolve. I observed this while writing the
test-create-vat.js GC tests, but couldn't figure out the cause, so I
included a commented-out `expectedRefcount -= 1` in anticipation of
this problem being resolved someday.

When PR #5892 disabled that debugging feature (by patching our copy of
Endo's eventual-send package), the undesired retention went away,
causing the test to fail, so it was disabled.

Now that the presences are dropped properly, this patch re-enables the
test and uncomments the refcount decrement.

refs https://github.com/endojs/endo/issues/1245
refs #5892

### DIFF
--- a/packages/SwingSet/test/vat-admin/test-create-vat.js
+++ b/packages/SwingSet/test/vat-admin/test-create-vat.js
@@ -258,31 +258,8 @@ function findRefs(kvStore, koid) {
   const refcount = refcountString.split(',').map(Number);
   return { refs, refcount };
 }
-/**
- * TODO Figure out why #5892 causes a refcount difference, and
- * unskip this.
- *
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM

Congrats that the original already suspected what the right answer should be, even though that's not what it was seeing. Nice!

PR #5898, opened 2022-08-05
WIP: better mismatch errors

erights:
Stacked on https://github.com/Agoric/agoric-sdk/pull/5876

At https://github.com/Agoric/agoric-sdk/pull/5876#discussion_r935866850 @Chris-Hibbert observes that #5876 would make error messages worse. I would like to keep the declarative pattern-based input validation, so this PR is a first step at improving the error messages to regain some of the ground lost in #5876


Chris-Hibbert (approved):
This is good enough to resolve my concerns about the underlying PR. Please merge it.

PR #5904, opened 2022-08-06
fix(swingset): don't delete heap snapshot if it didn't change

### LINKED ISSUE #5901: chain node crashed trying to delete missing XS heap snapshot
@arirubinstein spun up a test network this morning, and one of the validators halted when it attempted to delete an XS heap snapshot that was already missing.

```
2022-08-05T16:28:36.603Z block-manager: block 242 commit
portHandler threw (Error#1)
Error#1: [{"errno":-2,"code":"ENOENT","syscall":"unlink","path":"/state/agoricloadtest-18/data/ag-cosmos-chain-state/xs-snapshots/132a8428be21e653646f484a8e4ea19370e3eaaac57715b1f61294747df3177c.gz"}]

  at Object.commitDeletes (packages/swing-store/src/snapStore.js:267:13)
  at async commit (packages/swing-store/src/swingStore.js:340:5)
  at async saveOutsideState (packages/cosmic-swingset/src/launch-chain.js:285:5)
  at async blockingSend (packages/cosmic-swingset/src/block-manager.js:170:9)
  at async portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:232:16)
```

I'm guessing that the refcounting mechanism (the one that tracks which snapshots are in use and which are not) got confused, and it tried to delete a snapshot twice. I'm looking at the logs now.

warner:
Deliveries to a vat might not actually change the state of the vat, if
the action is read-only and GC is good at cleaning up any temporary
objects. When the vat runs inside an `xsnap` worker, this can result
in the following heap snapshot being identical to the previous one.

`vatKeeper.saveSnapshot()` had a bug in which it always marked the
previous snapshotID as unused, even if the new snapshotID was the same
as the old one. The snapStore handled this correctly, but vatKeeper
was telling it to `prepareToDelete` by mistake.

Snapshots and the kvStore are (unfortunately) in separate databases,
with separate commit points. To avoid losing data if the application
halts after kvStore commits but before snapStore commits, we defer
deletion of unused snapshots until the `snapStore.commit()`. The
snapStore maintains an in-RAM `Set` of unreferenced snapshot IDs to
remember what needs deletion.

The vatKeeper bug manifested as a crash after the following sequence
of events:

* 1: a delivery is made that does not change the vat state (snapshotA)
  * the bug tells snapStore to add snapshotA to the deletion list
* 2: `snapStore.commit()` runs
  * snapshotA is deleted, oops
* 3: a second delivery arrives, resulting in snapshotB
  * vatKeeper tells snapStore to delete snapshotA again
* 4: `snapStore.commit()` runs again
  * snapStore attempts to re-delete the snapshotA file
  * that results in ENOENT and a crash

We only create a snapshot for every 200 cranks, so there could be a
considerable delay between events 2 and 3, allowing the crash to
happen well after the actual problem. And until recently, xsnap was
not creating deterministic snapshots, masking the problem.

The fix is to change `vatKeeper` to not assume the new snapshot is
different than the old one: if they are the same, don't mark the old
one as unused.

closes #5901

### DIFF
--- a/packages/SwingSet/src/kernel/state/vatKeeper.js
+++ b/packages/SwingSet/src/kernel/state/vatKeeper.js
@@ -590,7 +590,7 @@ export function makeVatKeeper(
 
     const snapshotID = await manager.makeSnapshot(snapStore);
     const old = getLastSnapshot();
-    if (old) {
+    if (old && old.snapshotID !== snapshotID) {
       if (removeFromSnapshot(old.snapshotID) === 0) {
         snapStore.prepareToDelete(old.snapshotID);
       }
--- a/packages/SwingSet/src/kernel/vat-warehouse.js
+++ b/packages/SwingSet/src/kernel/vat-warehouse.js
@@ -304,7 +304,10 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
       vatKeeper.transcriptSnapshotStats();
     if (snapshotInitial === totalEntries) {
       reason = { snapshotInitial };
-    } else if (totalEntries - snapshottedEntries >= snapshotInterval) {
+    } else if (
+      totalEntries > snapshotInitial &&
+      totalEntries - snapshottedEntries >= snapshotInterval
+    ) {
       reason = { snapshotInterval };
     }
     // console.log('maybeSaveSnapshot: reason:', reason);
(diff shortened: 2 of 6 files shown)


FUDCo (approved):
A classic case of easier to fix than test...

Issue #4360, opened 2022-01-23
Show net results of exited offers


samsiegart:
## What is the Problem Being Solved?

If an offer is "exited" by the user, the assets may or may not have already been exchanged by the contract. Additionally, completed offers may not have actually taken all of the `give`, or may have given more than the `want`. It would be helpful for the users to see the net results of each offer after they are exited.

Current Screenshot: 
![image](https://user-images.githubusercontent.com/8848650/150660308-123ed9e9-1982-4737-999d-5a872dcd9fe5.png)

## Description of the Design

Idea 1: Show a section at the bottom that displays the amount of each asset actually transferred by the contract.
Idea 2: Show the amount transferred next to the give/want amounts after the offer is exited.

Alternative consideration: Show a "refunded"/"exited" state for some offers instead of a completed state in all cases.

Initial discussion:

@michaelfig: "This is a good idea, maybe a "refunded" state. Zoe currently doesn't do anything to distinguish this result, so it would take some work on the Zoe service to be able to report it. We could instead hack together something in the wallet backend to heuristically determine that it was a refund (not my preference)."

@erights "'Cancelled' seems a terrible word if the proposed trade could have already happened... 'I cancelled my offer, but my offered assets are still gone'"

Details TODO

michaelfig:
I presume that this issue also includes labelling a button "exit offer" instead of "cancel"?

erights:
I like option 2. Something like a transition from

&nbsp;
--------|
Give Bid 
3.00 Run
 &nbsp;   
Want Assets 
1 cardStore.Card

to

&nbsp; | &nbsp;
---------|--------
Gave Bid | Partial Refund
3.00 Run | 2.50 Run
 &nbsp;   | &nbsp;
Wanted Assets | Assets Received
1 cardStore.Card | 1 cardStore.Card

michaelfig:
> I like option 2. Something like a transition from

We probably need some more design work around how receiving the following is represented:

1. wanted assets
2. wanted assets + additional assets
3. full refund
4. full refund + additional assets

Tartuffo:
Sam to re-write this, especially to highlight any possible MN-1 issues this might address.

Tartuffo:
@samsiegart and @rowgraus Please mock this up.  Lowish prio for MN-1.

erights:
See #3056

michaelfig:
Tentative design:

[...]

PR #5905, opened 2022-08-06
feat(zoe): add E(userSeat).wasWantSatisfied()

### LINKED ISSUE #3056: AMM contract needs to give Wallet UI feedback for rejected offers
Background:
Current behavior: a user performs a trade on the AMM.  The resulting offer rejects due to failed offer safety - perhaps the AMM price moved too much in between.  The failure shows in logs but to the user the trade simply never resolves.  No indication is given of the failure.

The wallet needs to provide some indication of transaction success/failure to the user.  Perhaps a history of offers?

### LINKED ISSUE #5818: Zoe reveal whether (to what extent) an Offer's want was satisfied

Zoe's guarantee is that at least one of the user's give and want were satisfied. (With multiples #5422, we will have to ask how many times the want was satisfied). Contracts shouldn't be trusted to answer this question. So Zoe should at least provide a helper that gives a simple answer to the question "Did my trade go through?"

## Description of the Design

In addition to `userSeat.hasExited()`, we could add `userSeat.wantSatisfied()`. For now, it would be a Promise that resolves to 0 or 1. With Multiples, the number might be greater than 1.  The holder of the userSeat would still have to check whether any of the give was returned, as that is orthogonal.

## Security Considerations

Doesn't reveal anything that should be hidden.

## Test Plan

straightforward.  also ensure that the result is reasonable with various forms of exit.

Chris-Hibbert:
closes: #5818
closes: #3056
refs: #4360

## Description

Add a method on Zoe's userSeats that reveals whether the seat exited with its wants satisfied.

### Security Considerations

This doesn't reveal interim state of a seat. In fact, it doesn't reveal anything the userSeat didn't already have access to, since it only needed the payouts and the proposal. But this is a more reliable way for dApps to find out than doing the calculation themselves.

Needs to be added to the Zoe docs.

### Testing Considerations

Tests simple cases of satisfying and not satisfying the wants. Verifies that the promise is available before the seat exits.

### DIFF
--- a/packages/zoe/src/contractFacet/offerSafety.js
+++ b/packages/zoe/src/contractFacet/offerSafety.js
@@ -7,20 +7,25 @@ import { AmountMath } from '@agoric/ertp';
  * allocationAmount greater than or equal to requiredAmount for every
  * keyword of giveOrWant?
  *
+ * To prepare for multiples, satisfiesWant and satisfiesGive return 0 or 1.
+ * isOfferSafe will still be boolean. When we have Multiples, satisfiesWant and
+ * satisfiesGive will tell how many times the offer was matched.
+ *
  * @param {AmountKeywordRecord} giveOrWant
  * @param {AmountKeywordRecord} allocation
+ * @returns {0|1}
  */
 const satisfiesInternal = (giveOrWant = {}, allocation) => {
   const isGTEByKeyword = ([keyword, requiredAmount]) => {
     // If there is no allocation for a keyword, we know the giveOrWant
     // is not satisfied without checking further.
     if (allocation[keyword] === undefined) {
-      return false;
+      return 0;
     }
     const allocationAmount = allocation[keyword];
-    return AmountMath.isGTE(allocationAmount, requiredAmount);
+    return AmountMath.isGTE(allocationAmount, requiredAmount) ? 1 : 0;
   };
-  return Object.entries(giveOrWant).every(isGTEByKeyword);
+  return Object.entries(giveOrWant).every(isGTEByKeyword) ? 1 : 0;
 };
 
 /**
@@ -77,8 +82,11 @@ const satisfiesGive = (proposal, allocation) =>
  */
 function isOfferSafe(proposal, allocation) {
   return (
-    satisfiesGive(proposal, allocation) || satisfiesWant(proposal, allocation)
+    satisfiesGive(proposal, allocation) > 0 ||
+    satisfiesWant(proposal, allocation) > 0
   );
 }
 
+harden(isOfferSafe);
+harden(satisfiesWant);
 export { isOfferSafe, satisfiesWant };
--- a/packages/zoe/src/zoeService/types.js
+++ b/packages/zoe/src/zoeService/types.js
@@ -182,7 +182,7 @@
  * @property {() => Promise<OR>} getOfferResult
  * @property {() => void=} tryExit
  * @property {() => Promise<boolean>} hasExited
- * @property {() => Promise<0|1>} wasWantSatisfied
+ * @property {() => Promise<0|1>} numWantsSatisfied
  *
  * @property {() => Promise<Allocation>} getCurrentAllocationJig
  * Labelled "Jig" because it *should* only be used for tests, though
--- a/packages/zoe/src/zoeService/types.js
+++ b/packages/zoe/src/zoeService/types.js
@@ -182,6 +182,7 @@
  * @property {() => Promise<OR>} getOfferResult
  * @property {() => void=} tryExit
  * @property {() => Promise<boolean>} hasExited
+ * @property {() => Promise<0|1>} wasWantSatisfied
  *
  * @property {() => Promise<Allocation>} getCurrentAllocationJig
  * Labelled "Jig" because it *should* only be used for tests, though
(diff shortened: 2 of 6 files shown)


packages/zoe/src/zoeService/zoeSeat.js
```diff
@@ -95,6 +96,14 @@ export const makeZoeSeatAdminKit = (
...
     getCurrentAllocationJig: async () => currentAllocation,
     getAllocationNotifierJig: async () => notifier,
+
+    wasWantSatisfied: async () => {
+      return E.when(payoutPromiseKit.promise, () =>
+        Object.keys(proposal.want).every(kwd =>
+          AmountMath.isGTE(currentAllocation[kwd], proposal.want[kwd]),
+        ),
+      );
+    },
```
erights: Id prefer that you use `satisfiesWant`, which you'll need to export from `offerSafety.js`. That way you'll already be prepared for multiples, which does export it defined as

```js
export const satisfiesWant = (proposal, allocation) =>
  satisfiesInternal(proposal.want, allocation, 1n) >= 1n;
harden(satisfiesWant);
```

This also makes clear that this PR is necessary! If clients had tested this themselves, they also would have created an incompatibility with multiples!

erights (changes requested):
Needs to be added to the `UserSeat` type as well.

packages/zoe/src/zoeService/zoeSeat.js
erights: I see that it was also exported on master, so no worries about that.

erights: Actually, I remember we had a discussion about this, and agreed it was better for the public method to return `0` or `1` for now, rather than `false` and `true`, so that it could be more informative once we have multiples. How many times was the `want` satisfied?

Chris-Hibbert: I've updated offerSafety to return 0 or 1, and added tests for `satisfiesWant`.

Chris-Hibbert (commented):
> Needs to be added to the UserSeat type as well.

Done.

packages/zoe/src/contractFacet/offerSafety.js
```diff
@@ -7,20 +7,25 @@ import { AmountMath } from '@agoric/ertp';
  * allocationAmount greater than or equal to requiredAmount for every
  * keyword of giveOrWant?
  *
+ * To prepare for multiples, satisfiesWant and satisfiesGive return 0 or 1.
+ * isOfferSafe will still be boolean. When we have Multiples, satisfiesWant and
+ * satisfiesGive will tell how many times the offer was matched.
+ *
  * @param {AmountKeywordRecord} giveOrWant
  * @param {AmountKeywordRecord} allocation
+ * @returns {0|1}
```
erights, line 16: Sometimes Typescript amazes me ;)
No change suggested.

packages/zoe/src/contractFacet/offerSafety.js
```diff
@@ -77,8 +82,11 @@ const satisfiesGive = (proposal, allocation) =>
  */
 function isOfferSafe(proposal, allocation) {
   return (
-    satisfiesGive(proposal, allocation) || satisfiesWant(proposal, allocation)
+    satisfiesGive(proposal, allocation) > 0 ||
+    satisfiesWant(proposal, allocation) > 0
```
erights, line 86: Thank you for not relying on `0` being falsy ;)
No change suggested.

packages/zoe/src/contractFacet/offerSafety.js
```diff
@@ -77,8 +82,11 @@ const satisfiesGive = (proposal, allocation) =>
...
   );
 }
 
+harden(isOfferSafe);
+harden(satisfiesWant);
```
erights, line 91: Wow, we were missing `harden`s before. Good catch!

While not suggesting a change, I will note that my style is to export each in place, and follow each immediately by a `harden`. This is safe against the unhardened state being observable *only* if the exported declarations are not `'function'` function declarations. Converting to arrow functions needs to happen sometime, but need not happen for old functions in the PR.

packages/zoe/src/zoeService/types.js
```diff
@@ -182,6 +182,7 @@
  * @property {() => Promise<OR>} getOfferResult
  * @property {() => void=} tryExit
  * @property {() => Promise<boolean>} hasExited
+ * @property {() => Promise<0|1>} wasWantSatisfied
```
erights: The name `wasWantSatisfied` definitely suggests a yes or no answer.
What name would work better with returning a count?

erights (approved):
LGTM

Need to rename `wasWantSatisfied` to something compat with returning a count. Best I came up with is "`satisfiedWant`", but is not great. Please come up with something better ;)

packages/zoe/src/zoeService/types.js
turadg: `numWantsSatisfied` or simply `wantsSatisfied` since singular is by far more common than multi

Chris-Hibbert: I like @turadg's suggestion of `numWantsSatisfied`.  (It beats out 'wantsSatisfied' because the latter can be read as boolean.

turadg: > because the latter can be read as boolean.

fwiw, that's what I proposed as its virtue. when the caller doesn't know or care about multiples they don't have to. If they try to do anything with it that works for a boolean and not `0 | 1` the type checker will inform them.

Issue #4819, opened 2022-03-10
Need dependable mechanism for getting Vault issuers into your wallet


Tartuffo:
A pet naming the wallet maintains and imports into the DApps.

One namespace shared across all user interfaces.  Also needs to work for Pegasus imports.

michaelfig:
We intend to use ibc-go's `transfer` module (running on the `transfer` port) to send tokens to the Agoric chain.  Then, we will "bless" the relevant denom (`ibc/xxx`) to correspond to a bank purse, so that every wallet automatically gets an ERTP purse for the asset, pegged to their Cosmos `ibc/xxx` token balance.

Tartuffo:
Currently, the Vault UI suggests issuers and what to call them, which are stored in the wallet.
Wallet Bridge is where DAPP communicates with user Wallet, Wallet has objects -> paths.
Needs to be a pet name path, in the Agoric supported naming hub.  So user's pet name for the naming hub, + the path in the naming hub.  
The Vault is the pet name, Vault.ATOM would be the path for ATOM. 
Governance should install a naming hub.
The missing piece is to be able to say "sign me up for (the chain's gov-approved) standard names, or Joe's standard names)
one question is whether we show the path explicitly in the UI, or make that available via mouse-over.

michaelfig:
If the assets already need to be exposed by the bank vat (like IBC `transfer` assets), we have an automatic solution.  This ticket is for assets that aren't part of the bank.

dckc:
I talked about this with @michaelfig .

The VBANK currently provides a solution in this space that seems adequate for some near-term scenarios: issuers for assets added to the VBANK get into wallets of all users. The ATOM issuer used for vaults will be created via VBANK.

It doesn't accomodate Pegasus, decentralized lists, etc.

Tartuffo:
This issue is about getting a denom into the purse.
1. I have to know the name of the denom up front.  You need the IBC path (a hash).
2. Write about 30 lines of code, submit to Gov vote, and execute the 30 lines on the vote passing.

Making this less centralized is not needed for MN-1.

dckc:
@dtribble and @michaelfig refined the design above:

 - subscription for updates to the naming hub
 - object to name(s) lookup (#5107)
 - provide same interface from VBANK

dckc:
I suppose this recent problem report means this issue came up in production:

 - #6507

it would be nice if
[...]

Issue #2230, opened 2021-01-21
Want Patterns Implementation Plan


katelynsills:
# User Stories

- **[Find invitation in wallet for offer](https://github.com/Agoric/agoric-sdk/issues/2345)**: As a user, I can find an invitation in my wallet to use for an offer, without having to know the details (such as the invitation handle) for that particular invitation. Just knowing the instance and description should be enough to choose an invitation to use.
- **[Find all possible invitations in wallet for offer](https://github.com/Agoric/agoric-sdk/issues/2346)**: As a user, I can find all matching invitations that could potentially be used for an offer, without having to know the details (such as the invitation handle) for that particular invitation. Just being able to know the instance should be enough to get a number of invitations to choose from.
- **[Buy a ticket for an event with offer safety](https://github.com/Agoric/agoric-sdk/issues/2349)**: As a user, I want to be able to buy a ticket to an event and have it be protected by offer safety. I do not want to have to specify things that are not important to me, like a seat number.
- **[Buy an option with an `expirationDate` greater than or equal to some timestamp](https://github.com/Agoric/agoric-sdk/issues/2352)**: As a user, I want to buy an option and have my purchase be protected by offer safety. I want to be able to specify things that are important to me, such as that the option will not expire until after a specific timestamp. I don't want to have to specify things that are not important to me, such as the invitation handle for a particular option. (This use case also pertains to staking derivatives.)
- **[Withdraw an invitation from a purse, even if it is not deposited in the purse yet](https://github.com/Agoric/agoric-sdk/issues/2354)**: As a user, I would like to attempt to withdraw an invitation from a purse even if it has not been deposited yet, and receive a promise that will be resolved to a payment when the withdrawal occurs. This will allow me to set up and approve offers even if I do not yet have the invitation in hand to use. I would like to be able to eventually withdraw without knowing the details of the invitation, such as the invitation handle.
- **[Dominant Assurance Contract](https://github.com/Agoric/agoric-sdk/issues/1914)**: As a user, I want to specify that either the public good is produced, OR I get money back plus the extra amount I get for participating (note that this is different than what I gave). 

# Expected Behavior
There will be (at least) three additions to `amountMath` methods (actual names TBD):

 - matchMin(amount, pattern): get the minimal match that satisfies or return `undefined`
   - aka frugal
 - matchMax(amount, pattern) : get the maximal match that satisfies or return `undefined`.
   - aka anti-frugal
 - getPatternMaker: return an object for making patterns
[...]

PR #5913, opened 2022-08-08
more RUN remnants

turadg:
## Description

Some internal references remained to RUN. This changes those.

It also adds board ids for `STABLE` as an alternative to `RUN`. Rather than removing the latter it deprecates so dapps don't break. (those that use https://github.com/search?q=org%3AAgoric+RUN_ISSUER_BOARD_ID&type=code ) 

Once those are converted to use `STABLE` we can remove the deprecated in https://github.com/Agoric/agoric-sdk/issues/5605

### Security Considerations

--

--

### Testing Considerations

Should not affect tests.

### DIFF
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -539,10 +539,10 @@ harden(grantVaultFactoryControl);
 export const configureVaultFactoryUI = async ({
   consume: { board, zoe },
   issuer: {
-    consume: { [Stable.symbol]: centralIssuerP },
+    consume: { [Stable.symbol]: stableIssuerP },
   },
   brand: {
-    consume: { [Stable.symbol]: centralBrandP },
+    consume: { [Stable.symbol]: stableBrandP },
   },
   installation: {
     consume: {
@@ -569,9 +569,9 @@ export const configureVaultFactoryUI = async ({
     amm: ammInstance,
     vaultFactory: vaultInstance,
   });
-  const [centralIssuer, centralBrand] = await Promise.all([
-    centralIssuerP,
-    centralBrandP,
+  const [stableIssuer, stableBrand] = await Promise.all([
+    stableIssuerP,
+    stableBrandP,
   ]);
 
   const invitationIssuer = await E(zoe).getInvitationIssuer();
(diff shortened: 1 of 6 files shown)


packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -589,8 +589,12 @@ export const configureVaultFactoryUI = async ({
...
     ['INSTANCE_BOARD_ID', instances.vaultFactory],
     ['INSTALLATION_BOARD_ID', installs.vaultFactory],
-    ['RUN_ISSUER_BOARD_ID', centralIssuer],
-    ['RUN_BRAND_BOARD_ID', centralBrand],
+    // @deprecated
+    ['RUN_ISSUER_BOARD_ID', stableIssuer],
+    // @deprecated
+    ['RUN_BRAND_BOARD_ID', stableBrand],
+    ['STABLE_ISSUER_BOARD_ID', stableIssuer],
+    ['STABLE_BRAND_BOARD_ID', stableBrand],
```
dckc, line 597: things like this really should be looked up from a namehub such as `E(home).agoricNames` (see also #4819), so adding features to `configureVaultFactoryUI` seems like the wrong direction. Not critical, I suppose.

packages/inter-protocol/src/stakeFactory/stakeFactory.js
```diff
@@ -209,7 +209,7 @@ export const start = async (
   const publicFacet = augmentPublicFacet(
     Far('stakeFactory public', {
       makeLoanInvitation: () =>
-        zcf.makeInvitation(offerHandler, 'make RUNstake'),
+        zcf.makeInvitation(offerHandler, 'make stakeFactory'),
```
dckc, line 212: Good catch. This change is part of the visible contract API, so it's more important than many of the others. Please consider separating it out as a separate commit marked as a breaking change fix.

packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
```diff
@@ -358,7 +358,7 @@ const helperBehavior = {
     helper.assertVaultHoldsNoMinted();
     seat.exit();
 
-    return 'Your RUNstake is closed; thank you for your business.';
+    return 'Your stakeFactory is closed; thank you for your business.';
```
dckc, line 361: Messages like this don't play a critical security role like invitation descriptions, but strictly speaking, this is also a breaking contract API change.

dckc (approved):
good work.

Please consider conventional commits for breaking changes. (not critical)

packages/inter-protocol/src/stakeFactory/stakeFactory.js
turadg: Visible to end users but not read by machines, correct? I.e. I don't believe this change breaks any commitments of the API.
https://github.com/Agoric/agoric-sdk/blob/master/packages/zoe/src/contractFacet/types.js#L93-L97

packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
turadg: The API just says that it returns a string. Anything depending on parsing that string is making unwarranted assumptions about the API.

packages/inter-protocol/src/proposals/econ-behaviors.js
turadg: Yeah. Easy enough to remove when that work is done. Meanwhile I don't want to slow down the dapps moving off RUN_*

packages/inter-protocol/src/stakeFactory/stakeFactory.js
dckc: I think it would make sense for a machine to select among invitations from an installation based on description. This is discussed in #2230 for example:

> Just knowing the instance and description should be enough to choose an invitation to use.

packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
dckc: We have quite a few tests that depend on the contents of these strings. That seemed like an unwarranted assumption to me too, but that seems to be the current design. I guess it doesn't matter that much...

PR #5916, opened 2022-08-08
bump dev deps

turadg:
## Description

- remove @agoric/babel-parser that's appears to have been missed on earlier removal
- remove object-inspect package because xsnap brought in the source file instead
- bump typescript deps
- upgrade fast-check

### Security Considerations

--

--

### Testing Considerations

--


gibson042 (approved):
Ah ha, the Go failure is back!

PR #5917, opened 2022-08-08
feat(vaultFactory)!: split numVaults to numActiveVaults and numLiquidatingVaults

### LINKED ISSUE #5691: a single liquidation failure can deadlock liquidation process

As observed in https://github.com/Agoric/agoric-sdk/pull/5689, during a liquidation `highestRatio()` doesn't return the highest ratio. When what was the highest ratio vault loses its collateral to liquidation, it has an undefined ratio and the method fails its invariant.

If everything goes smoothly, it's fine that `liquidating` is at the head because the code won't request that again until it's `liquidated` and removed. But there could be a failure to liquidate and it would be stuck in `liquidating`, blocking the liquidation queue consumer from ever succeeding and thus preventing all liquidation.

## Description of the Design

We need to hold in storage vaults that are `liquidating` because the contract can be stopped at any time for system upgrade or just restart. Because liquidation happens one at a time, only the current one need be stored. But there's another concern which is recovering from liquidation failure. The current design makes no attempt to try again when a vault fails to liquidate, but a future revision of the contract may be able to. To make that possible it needs to hold in storage all the vaults that have failed.

This suggest that `prioritizedVaults` holds only vaults that can be liquidated. (Right now the invariant is it holds vaults that aren't _settled_ (`liquidated` or `closed`), allowing for vaults that are `liquidating`.) Vaults would move from the queue to the "in progress" set when they go from `active` to `liquidating` and then be removed completely upon `liquidated`. If they get stuck in `liquidating` they stay there until some contract upgrade (perhaps with governance method features) can retrieve them.

## Security Considerations
--
## Test Plan

--

Chris-Hibbert:
closes: #5691 

## Description

Move vaults to a new SetStore upon entering liquidation.  The goal is to not leave them in orderedVaultStore if they get stuck, and to be able to find them again (via upgrade) if repair is necessary.

### Security Considerations

The goal is to make liquidation more robust. Without this fix, it seems possible for a failure in liquidation to leave a vault at the head of the queue and unable to make progress. This change moves it out of the head of the queue, so a failure to liquidate won't block the queue.

None

### Testing Considerations

I couldn't find  a cost effective way to simulate a failure in liquidation. so the recovery path is untested.

### DIFF
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -23,6 +23,7 @@ import {
   makeKindHandle,
   makeScalarBigMapStore,
   pickFacet,
+  makeScalarBigSetStore,
 } from '@agoric/vat-data';
 import {
   assertProposalShape,
@@ -56,9 +57,10 @@ const trace = makeTracer('VM');
 /**
  * @typedef {object} MetricsNotification
  *
- * @property {number}         numVaults        present count of vaults
- * @property {Amount<'nat'>}  totalCollateral  present sum of collateral across all vaults
- * @property {Amount<'nat'>}  totalDebt        present sum of debt across all vaults
+ * @property {number}         numVaults          present count of vaults
+ * @property {number}         liquidatingVaults  present count of liquidating vaults
+ * @property {Amount<'nat'>}  totalCollateral    present sum of collateral across all vaults
+ * @property {Amount<'nat'>}  totalDebt          present sum of debt across all vaults
  *
  * @property {Amount<'nat'>}  totalCollateralSold       running sum of collateral sold in liquidation // totalCollateralSold
  * @property {Amount<'nat'>}  totalOverageReceived      running sum of overages, central received greater than debt
@@ -93,6 +95,7 @@ const trace = makeTracer('VM');
  * debtMint: ZCFMint<'nat'>,
  * poolIncrementSeat: ZCFSeat,
  * unsettledVaults: MapStore<string, Vault>,
+ * liquidatingVaults: SetStore<Vault>,
  * }>} ImmutableState
  */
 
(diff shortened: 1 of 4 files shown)


packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -202,13 +205,24 @@ const initState = (
     durable: true,
   });
 
+  /**
+   * If things are going well, the set will contain at most one Vault. Otherwise
+   * failures remain and are available to be repaired via contract upgrade.
+   *
+   * @type {SetStore<Vault>}
+   */
+  const liquidatingVaults = makeScalarBigSetStore('orderedVaultStore', {
```
turadg: copy-pasta. 
```suggestion
  const liquidatingVaults = makeScalarBigSetStore('liquidatingVaults', {
```

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -56,9 +57,10 @@ const trace = makeTracer('VM');
 /**
  * @typedef {object} MetricsNotification
  *
- * @property {number}         numVaults        present count of vaults
- * @property {Amount<'nat'>}  totalCollateral  present sum of collateral across all vaults
- * @property {Amount<'nat'>}  totalDebt        present sum of debt across all vaults
+ * @property {number}         numVaults          present count of vaults
+ * @property {number}         liquidatingVaults  present count of liquidating vaults
```
turadg: these names are inconsistent. they should both or neither have `num` prefix. I request that it be included as a cue for "present count" but open to counter arguments.

turadg: separately, the names and comment imply that one is a subset of the other but in code they're complements. so either rename to `numActiveVaults` or make `numVaults` include both the cardinality of both stores. I.e. the number of unsettled vaults.

turadg (changes requested):

turadg (approved):
LGTM, though I hope it gets squashed to one commit

PR #5919, opened 2022-08-09
feat(liquidation): If liquidation results in a shortfall, don't refund excess

### LINKED ISSUE #5558: when vault is liquidated with shortfall, take all collateral

Sometimes when a vault is liquidated and there's not enough collateral to cover the debt, there's still collateral left over after.

https://github.com/Agoric/agoric-sdk/blob/b5b878d97f4079e58dc89326882122da0065b3f9/packages/run-protocol/test/vaultFactory/test-vaultFactory.js#L794-L805

## Description of the Design

In the event that a vault is liquidated and can't cover its debt, take all available collateral.

Chris-Hibbert:
closes: #5558

## Description

If liquidation results in a shortfall, don't refund the excess. The Vaults hold onto the excess collateral. There's currently no code to dispose of it.

When the value crosses some threshold, we could sell it to the AMM or
transfer it to the Reserve

### Security Considerations

None.

This is probably what people expect, so no additions. This does add to the accounting records, if that's written up anywhere.

### Testing Considerations

update some tests.

### DIFF
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -64,7 +64,7 @@ const trace = makeTracer('VM');
  * @property {Amount<'nat'>}  totalCollateralHeld collateral held as a result of not returning excess refunds
  *                                                from AMM to owners of vaults liquidated with shortfalls
  *
- * @property {Amount<'nat'>}  totalCollateralSold       running sum of collateral sold in liquidation // totalCollateralSold
+ * @property {Amount<'nat'>}  totalCollateralSold       running sum of collateral sold in liquidation
  * @property {Amount<'nat'>}  totalOverageReceived      running sum of overages, central received greater than debt
  * @property {Amount<'nat'>}  totalProceedsReceived     running sum of central received from liquidation
  * @property {Amount<'nat'>}  totalShortfallReceived    running sum of shortfalls, central received less than debt
@@ -543,22 +543,23 @@ const helperBehavior = {
         // current values
 
         // Sometimes, the AMM will sell less than all the collateral. If there
-        // was a shortfall, the investor doesn't keep the change, so we get it
+        // was a shortfall, the investor doesn't keep the change, so we get it.
         const collateralPost = vault.getCollateralAmount();
-        if (!AmountMath.isEmpty(collateralPost)) {
-          if (!AmountMath.isEmpty(accounting.shortfall)) {
-            // XXX The borrower doesn't get any excess collateal remaining when
-            // liquidation results in a shortfall. We currently do nothing with
-            // it. We could hold it until it crosses some threshold, then sell
-            // it to the AMM, or we could transfer it to the reserve. At least
-            // it's visible in the accounting.
-            vaultSeat.decrementBy(
-              state.collateralHeldSeat.incrementBy({
-                Collateral: collateralPost,
-              }),
-            );
-            zcf.reallocate(vaultSeat, state.collateralHeldSeat);
-          }
+        if (
+          !AmountMath.isEmpty(collateralPost) &&
+          !AmountMath.isEmpty(accounting.shortfall)
+        ) {
+          // XXX The borrower doesn't get any excess collateral remaining when
+          // liquidation results in a shortfall. We currently do nothing with
+          // it. We could hold it until it crosses some threshold, then sell it
+          // to the AMM, or we could transfer it to the reserve. At least it's
+          // visible in the accounting.
+          vaultSeat.decrementBy(
+            state.collateralHeldSeat.incrementBy({
+              Collateral: collateralPost,
+            }),
+          );
+          zcf.reallocate(vaultSeat, state.collateralHeldSeat);
         }
 
         state.totalCollateral = AmountMath.subtract(
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -61,6 +61,8 @@ const trace = makeTracer('VM');
  * @property {number}         numLiquidatingVaults  present count of liquidating vaults
  * @property {Amount<'nat'>}  totalCollateral    present sum of collateral across all vaults
  * @property {Amount<'nat'>}  totalDebt          present sum of debt across all vaults
+ * @property {Amount<'nat'>}  totalCollateralHeld collateral held as a result of not returning excess refunds
+ *                                                from AMM to owners of vaults liquidated with shortfalls
  *
  * @property {Amount<'nat'>}  totalCollateralSold       running sum of collateral sold in liquidation // totalCollateralSold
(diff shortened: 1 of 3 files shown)


packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -515,6 +523,26 @@ const helperBehavior = {
         facets.manager.burnAndRecord(accounting.toBurn, vaultSeat);
 
         // current values
+
+        // Sometimes, the AMM will sell less than all the collateral. If there
+        // was a shortfall, the investor doesn't keep the change, so we get it
+        const collateralPost = vault.getCollateralAmount();
+        if (!AmountMath.isEmpty(collateralPost)) {
+          if (!AmountMath.isEmpty(accounting.shortfall)) {
```
turadg: Is the nesting intentional?

While this is maybe a little harder for the eyes to follow the intent is clearer and it requires less indentation:
```suggestion
        if (!AmountMath.isEmpty(collateralPost) && !AmountMath.isEmpty(accounting.shortfall)) {
```

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -813,16 +812,14 @@ test('price falls precipitously', async t => {
...
-  // TODO take all collateral when vault is underwater https://github.com/Agoric/agoric-sdk/issues/5558
-  // t.deepEqual(
-  //   await E(vault).getCollateralAmount(),
-  //   aeth.makeEmpty(),
-  //   'Collateral used up trying to cover debt',
-  // );
+  t.deepEqual(
+    await E(vault).getCollateralAmount(),
+    aeth.makeEmpty(),
+    'Excess collateral not  returned due to shortfall',
```
turadg:
```suggestion
    'Excess collateral not returned due to shortfall',
```

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -58,6 +58,8 @@ const trace = makeTracer('VM');
  *
  * @property {number}         numVaults        present count of vaults
  * @property {Amount<'nat'>}  totalCollateral  present sum of collateral across all vaults
+ * @property {Amount<'nat'>}  totalCollateralHeld collateral held as a result of not returning excess refunds
+ *                                                from AMM to owners of vaults liquidated with shortfalls
  * @property {Amount<'nat'>}  totalDebt        present sum of debt across all vaults
  *
  * @property {Amount<'nat'>}  totalCollateralSold       running sum of collateral sold in liquidation // totalCollateralSold
```
turadg: This could read as a complement of the new value. An expression of their relationships would help. IIUK,

total collateral ever  = `totalCollateral` (now) + `totalCollateralSold` (ever)
`totalCollateralHeld` <<`totalCollateralSold` and can't be inferred from the other numbers

> This does add to the accounting records, if that's written up anywhere

Yep, this `MetricsNotification` object type. So the elaboration could go as a description of the type (which currently only has `@property` elements)

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -515,6 +523,26 @@ const helperBehavior = {
         facets.manager.burnAndRecord(accounting.toBurn, vaultSeat);
 
         // current values
+
+        // Sometimes, the AMM will sell less than all the collateral. If there
+        // was a shortfall, the investor doesn't keep the change, so we get it
```
turadg: just FYI: this doesn't end in a period while the other comment block does.

turadg (commented):
Excellent. Will approve when docs are clear.

packages/inter-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: > Yep, this MetricsNotification object type. 

Oh, sorry. I interpret the **Documentation** section as being about external user or developer documentation. I guess internal comments do also matter. I do hope our writer will find the internal docs when generating external docs, but I don't count on it.

packages/inter-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: Not intentional. It went through a stage where all the else branches were populated. I should have recognized the opportunity to condense when I was done.

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: done

packages/inter-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: "total collateral ever" includes a lot of collateral that was withdrawn by its owner, which we don't account for.
`totalCollateralHeld` was once a part of totalCollateral, but is not included in `totalCollateralSold`, as it was not sold.
At the end of a liquidation, `totalCollateral` is reduced by `collateralPre`, since all of that collateral was either sold, returned to the vault owner, or held by the VaultManager.
`collateralPost` can either be returned to the owner or held by the vaultManager, depending on the value of `shortfall`.

Does any of this need to be added to `MetricsNotification`, or described more completely in the comments here or in `liquidate()`?

turadg: > I interpret the Documentation section as being about external user or developer documentation

Me too. The `MetricsNotification` type gets exported. Anyone consumer from the subscriber will see that `MetricsNotification` is the type of data coming out. With comments there they'll be able to understand what the properties of the object are counting.

> "total collateral ever" includes a lot of collateral that was withdrawn by its owner, which we don't account for.

Ah, there's no way to reconstruct that. Got it.

> Does any of this need to be added to MetricsNotification, or described more completely in the comments here or in liquidate()?

I think some of this should make it into MetricsNotification (either on the type itself or its properties). Though I think I've figured out what makes me uneasy: `totalCollateralHeld` uses the past-participle like the "running sums" but it's a "present value" (because it's in the seat, and it can go down). This doesn't match the scheme https://github.com/Agoric/agoric-sdk/blob/a28987519403147d505346a646839ffb3e18622e/packages/inter-protocol/src/vaultFactory/vaultManager.js#L55

Also now that I think about it "held" is ambiguous in the context of getting a metric notification because all the collateral is held.

A third issue is that it's not a "total" in the way that "totalCollateral" is; that's a total because it sums over all vaults. But this is one seat.

Requesting that we solve those three. 

I think `unsoldCollateral` would do it but not requesting that specific solution.

packages/inter-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: done

packages/inter-protocol/src/vaultFactory/vaultManager.js
Chris-Hibbert: as discussed, switched to `retainedCollateral`.

turadg (approved):

PR #5922, opened 2022-08-09
fix: remove uses of __allowUnsafeMonkeyPatching__

erights:
The `__allowUnsafeMonkeyPatching__` option, as the name indicates, is an unsafe temporary hack that we did to work around what we had hoped was a temporary problem. To see if the underlying problem went away, this PR removes those vestigial uses to see what breaks.

When tried under CI, the only thing that broke is golangci-lint 
https://github.com/Agoric/agoric-sdk/runs/7754082612?check_suite_focus=true
which seems unlikely to be related. Reviewers, what else needs to be tested to see if this PR actually breaks anything?

It would be a big relief to remove uses of this unsafe hacky option.

### DIFF
--- a/packages/wallet/ui/src/lockdown.js
+++ b/packages/wallet/ui/src/lockdown.js
@@ -12,7 +12,6 @@ Math.pow = (base, exp) =>
     : mathPow(base, exp);
 
 lockdown({
-  __allowUnsafeMonkeyPatching__: 'unsafe',
   errorTaming: 'unsafe',
   overrideTaming: 'severe',
   consoleTaming,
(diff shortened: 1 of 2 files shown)


erights:
See https://github.com/endojs/endo/pull/1249

kriskowal (approved):
Fantastic. @samsiegart how confident are you that we won’t need to put this work-around back?

samsiegart:
> Fantastic. @samsiegart how confident are you that we won’t need to put this work-around back?

I need to test it manually, there's no unit tests that run under `lockdown` in the wallet UI, so I wouldn't expect tests to fail either way. Do we know exactly what changed to make this unnecessary?

erights:
> Do we know exactly what changed to make this unnecessary?

I do not.

erights:
> I need to test it manually...

@samsiegart I will hold this PR until I get your approval. Thanks!

samsiegart (approved):
Manual testing looks good

PR #5926, opened 2022-08-10
feat(chainStorage)!: assertPathSegment (replacing sanitizePathSegment)

gibson042:
## Description

Updates a comment to match the regular expression it describes.

### DIFF
--- a/packages/inter-protocol/src/proposals/price-feed-proposal.js
+++ b/packages/inter-protocol/src/proposals/price-feed-proposal.js
@@ -2,11 +2,18 @@ import { E } from '@endo/far';
 import { makeIssuerKit } from '@agoric/ertp';
 import {
   makeStorageNodeChild,
-  sanitizePathSegment,
+  assertPathSegment,
 } from '@agoric/vats/src/lib-chainStorage.js';
 import { deeplyFulfilled } from '@endo/marshal';
 
 import { reserveThenDeposit, reserveThenGetNames } from './utils.js';
+
+/** @type {(name: string) => void} */
+const sanitizePathSegment = name => {
+  const candidate = name.replace(/ /g, '_');
+  assertPathSegment(candidate);
+  return candidate;
+};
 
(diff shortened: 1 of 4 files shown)


turadg:
Please include in this PR a correction to `sanitizePathSegment` as well https://github.com/Agoric/agoric-sdk/pull/5850#discussion_r942708103

> Shouldn't this rather specific function be defined in a higher layer?

I don't think so. Its semantics are coupled to the regex right above it, which isn't exported.

gibson042:
> Please include in this PR a correction to `sanitizePathSegment` as well [#5850 (comment)](https://github.com/Agoric/agoric-sdk/pull/5850#discussion_r942708103)

To what? Some obvious possibilities include `name.replaceAll(' ', '_')` and `name.replaceAll(' ', '-')`.

> > Shouldn't this rather specific function be defined in a higher layer?
> 
> I don't think so. Its semantics are coupled to the regex right above it, which isn't exported.

I think I'd rather export the regex than include such a narrow (e.g., failing to address excessive length or invalid characters other than space) and opinionated (correcting by replacing each invalid character with a particular valid substitute rather than e.g. removing it or using a different substitute) function. The library is responsible for input validation, but (IMO) not error recovery.

turadg:
> than include such a narrow (e.g., failing to address excessive length or invalid characters other than space) and opinionated (correcting by replacing each invalid character with a particular valid substitute rather than e.g. removing it or using a different substitute) function. The library is responsible for input validation, but (IMO) not error recovery.

Persuasive points! I've moved `sanitizePathSegment` to the one place it was needed. Instead of exporting the regexp I exported an assertion for it, so it can be re-used.

I pushed them to this PR and Approved. I'll leave it to you to merge so you can review my commits. (Please don't squash as one is a breaking change to call out.)

turadg (approved):

gibson042 (commented):
Excellent; thank you!

Issue #4660, opened 2022-02-24
remove Lerna


turadg:
## What is the Problem Being Solved?

Lerna was abandoned,
> Important note: this project [is not actively maintained](https://github.com/lerna/lerna/issues/2703). Consider adopting an alternative toolset for monorepo management.

And then bought by Nx. They've been tightening its integration with their SaaS, which we don't use.

## Description of the Design

Either adopt Lerna Lite instead,
https://github.com/Agoric/agoric-sdk/pull/5927#issuecomment-1211342562

Or drop Lerna ([some research notes](https://docs.google.com/document/d/1QBiSweknlPPJgDmPRYgBvgH2zEjRveYAb1FK8thJYmo/edit#)).

## Security Considerations
TBD

## Test Plan

TBD

turadg:
Reconsider whether to remove Lerna because it has a new owner:
> Important note: this project is changing stewardship to Nrwl! Your favorite tool will continue to live on. Stay tuned for updates and soon a project roadmap!

We might want to make the jump to the new owner, https://nx.dev/

turadg:
I tried upgrading to Lerna 5 and when I tried `lerna run --stream test` it only checked a handful of packages and spuriously passed the rest. I didn't see anything in the https://github.com/lerna/lerna/blob/main/CHANGELOG.md to explain it. Makes me think we may be better off without it.

https://github.com/Agoric/agoric-sdk/pull/new/ta/lerna-v5

abebeos:
It seems the new Lerna owner is doing well and has a major interest in maintaining Lerna. The the issue #5927 is possibly fixed, according to within https://github.com/Agoric/agoric-sdk/pull/5927#issuecomment-1211342562.

Looks to me like this issue here can be closed (or at least update the original issue text)

Possible follow-up would be:

- Consider moving to lerna-light, see https://github.com/Agoric/agoric-sdk/pull/5927#issuecomment-1211342562

turadg:
Current hurdle: Agoric patches to the publishing behavior: https://github.com/Agoric/agoric-sdk/pull/8062#discussion_r1823343529

Removing Lerna might have to wait until a broader change to versioning, using Changesets

PR #3234, opened 2021-06-01
feat: updated PricedCallSpread contract

tgrecojs:
PR for [issue #2282](https://github.com/Agoric/agoric-sdk/issues/2282)

updated PricedCallSpread contract to take 4 digit input. After seeing the exisiting input argument had to fall between 0-100, I'm not sure if there should be a check to enforce inputs that are only 2 or 4 digits but would be happy to update if needed. 🙂

PR #5927, opened 2022-08-10
lerna v5

turadg:
refs: #4660

## Description

We might remove Lerna,
- https://github.com/Agoric/agoric-sdk/issues/4660

Meanwhile it's easy to get on a less ancient version. In particular, to [get pnpm support](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#features-19). (https://github.com/Agoric/agoric-sdk/pull/8062 tried going all the way to latest, but it changes version numbering behavior.)

No bad breaking changes,
https://github.com/lerna/lerna/blob/main/CHANGELOG.md#400-2021-02-10
https://github.com/lerna/lerna/blob/main/CHANGELOG.md#500-2022-05-24

This had been blocked on these:
- https://github.com/lerna/lerna/issues/3242
- https://github.com/lerna/lerna/issues/3233

But those are resolved. See Testing section.

### Testing Considerations
- [x] `yarn lerna exec npm pkg get name` lists all packages (41)

CI should test the publishing flow.

Previously: 
```
lerna notice cli v5.4.0
lerna info versioning independent
lerna info Executing command in 39 packages: "npm pkg get name"
…
lerna success exec Executed command in 39 packages: "npm pkg get name"
```
but only 19 packages ran (per `yarn lerna exec npm pkg get name |grep ": " | wc -l`)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -13,5 +13,6 @@
     "packages/*",
     "packages/wallet/*"
   ],
+  "useWorkspaces": true,
   "version": "independent"
 }
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "eslint-plugin-import": "^2.25.3",
     "eslint-plugin-jsdoc": "^46.4.3",
     "eslint-plugin-prettier": "^5.0.0",
-    "lerna": "^3.22.1",
+    "lerna": "^5.6.2",
     "npm-run-all": "^4.1.5",
     "prettier": "^3.0.0",
     "prettier-plugin-jsdoc": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -84,12 +84,5 @@
   },
   "dependencies": {
     "patch-package": "^6.5.1"
-  },
-  "resolutions": {
-    "**/acorn-globals/acorn": "^8.7.1",
-    "**/acorn-node/acorn": "^8.7.1",
-    "**/espree/acorn": "^8.7.1",
-    "**/eslint/@babel/code-frame": "^7.12.11",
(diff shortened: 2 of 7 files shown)


ghiscoding:
@turadg 
Hello I see that you linked your PR to Lerna's issue, just to let you know that the fix proposed in Lerna [PR #3234](https://github.com/lerna/lerna/pull/3234) does work and was applied into this Lerna-Lite [PR](https://github.com/ghiscoding/lerna-lite/pull/301) and that fixed what seems to be the same problem as you and was identified by the **Jest** team in this [issue](https://github.com/facebook/jest/pull/13097) when they recently switched to [Lerna-Lite](https://github.com/ghiscoding/lerna-lite). I have applied the fix and they are back in business. Since you're contemplating about the possibility to remove Lerna, you should probably take a look at [Lerna-Lite](https://github.com/ghiscoding/lerna-lite), it's much smaller (just install what you need) and the commands are executed the same way so it's easy to migrate (just swap the dependency and you're good to go). 

Also note that Lerna v5.3.0+ is now installing Nx behind the scene (want it or not) because they added Nx as a [dependency](https://github.com/lerna/lerna/blob/main/core/lerna/package.json#L57) of Lerna, however it remains totally optional in [Lerna-Lite](https://github.com/ghiscoding/lerna-lite) (you can read more about Lerna-Lite from the main readme).

Cheers and happy coding

abebeos:
> Revisit after https://github.com/lerna/lerna/pull/3234 is in master

@turadg

michaelfig (approved):

michaelfig:
Looks like `yarn.lock` needs regeneration.

Issue #5910, opened 2022-08-07
Zoe appears to have a memory leak


arirubinstein:
while debugging the bank vat memory leak, it appears that Zoe's vat also has a leak
<img width="901" alt="Screen Shot 2022-08-06 at 11 26 39 PM" src="https://user-images.githubusercontent.com/22369/183278208-3b601ed0-ce2c-435e-a900-e399e5f6aae0.png">

I can provide full slogs, logs, etc requested for this network.  I'll let it run over the weekend to generate additional data

https://ui.honeycomb.io/agoric/datasets/instagoric-loadtest/result/G24qAncnkwR

arirubinstein:
prelim assign to warner, but please reassign

arirubinstein:
this usage may actually be not as bad as initially thought - over time it seems to decrease which could just be representative of offers-in-flight
<img width="884" alt="Screen Shot 2022-08-07 at 2 28 43 PM" src="https://user-images.githubusercontent.com/22369/183311621-51190317-1342-4a91-9112-cdcc5a0b1274.png">

dckc:
I can try making the workers restart periodically. Every time we take a snapshot is the easiest; might as well start there.

warner:
or restart the validator every hour or two, if the cosmos/golang side doesn't have any problem with that

erights:
Yeah, that second set of graphs is interesting. Looking forward to the longer trace which should tell us whether this equilibrates.

warner:
I figured out that this is unrelated to Zoe: `xsnap` is leaking memory when writing snapshots. Zoe just happens to get a lot of deliveries, so it writes snapshots more often.

When I looked at the vats whose xsnap workers were consuming the most memory, I built out the following table:

```
| vatID        | RSS at crash | VmSize | raw snapshot | deliveries | metering.allocate |
|--------------+--------------+--------+--------------+------------+-------------------|
| v7-zoe       |       4.5 GB | 5.5 GB | 30 MB        |     116402 | 84_017_184        |
| v16-bank     |          2.2 |   3.25 | 19 MB        |     125002 | 54_657_056        |
| v28-zcf      |          2.0 |   3.07 | 8.9 MB       |      77202 | 67_239_968        |
| v4-vattp     |          1.0 |   2.06 | 3.9 MB       |     500202 | 42_074_144        |
| v1-bootstrap |          0.8 |   1.81 | 23.1 MB      |      39402 | 71_434_304        |
[...]

PR #5929, opened 2022-08-10
feat(SwingSet): evictAfterSnapshot option for periodic eviction of all vats

dckc:
Add a switch in case "turn it off and turn it on again" helps manage vat worker RAM usage.

refs: #5910 

`$EVICT_AFTER_SNAPSHOT` environment variable is documented.

### Testing Considerations

unit tests updated

passes a smoke test:

```
agoric-sdk/packages/cosmic-swingset$ EVICT_AFTER_SNAPSHOT=1 make scenario2-run-chain
...
2022-08-17T02:15:44.347Z SwingSet: vat: v1: agoricNamesAdmin settled; remaining: []
XS snapshot written to /home/connolly/projects/agoric-sdk/packages/cosmic-swingset/t1/n0/data/ag-cosmos-chain-state/xs-snapshots/ba667a6f910f04ec834edfb9f71b480f94b8f2fad3fe40bd0a27a4292a9d6f5d.gz : 2664962 bytes compressed, 14694111 raw
2022-08-17T02:15:45.181Z SwingSet: kernel: periodic eviction v1
```

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -86,6 +86,18 @@ end-of-block processing for `ag-chain-cosmos` and the sim-chain.
 
 Lifetime: until we have a more consistent load testing regimen
 
+## EVICT_AFTER_SNAPSHOT
+
+Affects: cosmic-swingset
+
+Purpose: work around possible memory leaks in vat workers
+
+Description: if set to non-empty string, vats are evicted
+after each snapshot, causing them to be restarted with a
+fresh process memory allocation.
+
+Lifetime: for the life of performance issues such as https://github.com/Agoric/agoric-sdk/issues/5910
+
 ## FAKE_CURRENCIES
 
 Affects: cosmic-swingset
--- a/packages/SwingSet/src/kernel/vat-warehouse.js
+++ b/packages/SwingSet/src/kernel/vat-warehouse.js
@@ -49,6 +49,7 @@ export const makeLRU = max => {
  * @param {ReturnType<typeof import('./vat-loader/vat-loader.js').makeVatLoader>} vatLoader
  * @param {{
  *   maxVatsOnline?: number,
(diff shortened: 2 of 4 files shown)


dckc:
@arirubinstein @warner maybe we should land this after all, until we have a better idea?

warner:
> @arirubinstein @warner maybe we should land this after all, until we have a better idea?

Let's look at the slogfile timing and get a sense for how long it takes to evict the worker (and how long it takes to load it back in again later).. that should help us make a decision.

warner:
Let's add `runtimeOptions.warehousePolicy.evictAfterSnapshot`, which `launch-chain.js` will add to the third argument of its `makeSwingsetController()` call, to enable this eviction only on the chain. Swingset will have it disabled by default. Let's add a copy of `test-controller.js` 's `static vats are unmetered on XS` to exercise setting this to `true`, which should cause it to observe `t.deepEqual(c.dump().log, ['buildRootObject called', 'bootstrap called', 'buildRootObject called', 'bootstrap called'])`. We could probably strip out some other parts of the test but if it's sensitive to the thing that we're testing, maybe we don't bother. It's probably worth a note in the new test that we expect the normal actions of the test (whatever `basedir-controller-2` 's config is provoking) to result in enough deliveries to trigger exactly one snapshot writing, followed by at least one delivery, so that we start a worker twice.

Although .. hang on, the second worker startup should be happening from the snapshot, so why is it invoking `vatPowers.testLog('buildRootObject called')` more than once? We might need to investigate that, to make sure the test is doing what we expect.

dckc:
> ... hang on, ... so why ...

I don't have a good mental model; I hope you can figure that out, @warner .
@FUDCo maybe you know?

dckc:
overtaken by #5987

PR #5930, opened 2022-08-10
refactor: remove deeplyFulfillTerms

erights:
Questions/suggestions for #5891 

Why do you prefer `deeplyFulfillTerms(...)` to `deeplyFulfilled(harden(...))`?

I prefer the latter because I don't want to need to teach future Jessie checking rules 
   * which functions harden their argument
   * to do the import/scope analysis to see if the function being called is one of those
   * to worry about releasing unhardened objects in a call to something that might not immediately harden it
   * to get in the habit of not visibly and obviously hardening as early as possible.

While I do hope to make progress on some abstractions around `Far` that may eventually enable us to safely harden less, we're not there yet. This direction is speculative and we should not count on it eventually succeeding.

### DIFF
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -190,17 +190,14 @@ export const setupAmm = async (
       consume: { [Stable.symbol]: runBrandP },
     },
     issuer: {
-      consume: { [Stable.symbol]: centralIssuerP },
+      consume: { [Stable.symbol]: centralIssuer },
     },
     instance: {
-      consume: { economicCommittee: electorateInstanceP },
+      consume: { economicCommittee: electorateInstance },
       produce: { ammGovernor },
     },
     installation: {
-      consume: {
-        contractGovernor: governorInstallation,
-        amm: ammInstallationP,
-      },
+      consume: { contractGovernor: governorInstallation, amm: ammInstallation },
     },
   },
   { options = {} } = {},
@@ -208,36 +205,23 @@ export const setupAmm = async (
   trace('setupAmm');
 
   const poserInvitationP = E(committeeCreator).getPoserInvitation();
-  const [
-    poserInvitation,
-    poserInvitationAmount,
-    runBrand,
-    electorateInstance,
-    centralIssuer,
-    ammInstallation,
-  ] = await Promise.all([
+  const [poserInvitation, runBrand] = await Promise.all([
     poserInvitationP,
-    E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
     runBrandP,
-    electorateInstanceP,
-    centralIssuerP,
-    ammInstallationP,
   ]);
   const { minInitialPoolLiquidity = 1_000_000_000n } = options;
 
-  const timer = await chainTimerService; // avoid promise for legibility
-
   const ammTerms = makeAmmTerms(
-    timer,
-    poserInvitationAmount,
+    chainTimerService,
+    E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
     AmountMath.make(runBrand, minInitialPoolLiquidity),
   );
 
   const storageNode = await makeStorageNode(chainStorage, AMM_STORAGE_PATH);
   const marshaller = await E(board).getPublishingMarshaller();
 
-  const ammGovernorTerms = {
-    timer,
+  const ammGovernorTerms = await deeplyFulfillTerms({
+    timer: chainTimerService,
     electorateInstance,
     governedContractInstallation: ammInstallation,
     governed: {
@@ -249,7 +233,8 @@ export const setupAmm = async (
         marshaller,
       },
     },
-  };
+  });
+
   /** @type {{ creatorFacet: GovernedContractFacetAccess<XYKAMMPublicFacet,XYKAMMCreatorFacet>, publicFacet: GovernorPublic, instance: Instance }} */
   const g = await E(zoe).startInstance(
     governorInstallation,
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -7,12 +7,10 @@ import '@agoric/governance/exported.js';
 import '@agoric/vats/exported.js';
 import '@agoric/vats/src/core/types.js';
 import { makeStorageNode } from '@agoric/vats/src/lib-chainStorage.js';
-import {
-  deeplyFulfillTerms,
-  makeRatio,
-} from '@agoric/zoe/src/contractSupport/index.js';
+import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
 import { E, Far } from '@endo/far';
-import { Stable, Stake } from '@agoric/vats/src/tokens.js';
+import { Stake, Stable } from '@agoric/vats/src/tokens.js';
+import { deeplyFulfilled } from '@endo/marshal';
 import * as Collect from '../collect.js';
 import { makeTracer } from '../makeTracer.js';
 import { makeStakeReporter } from '../my-lien.js';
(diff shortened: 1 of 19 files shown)


erights:
Be sure to hide whitespace when looking at these diffs.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -881,22 +879,24 @@ export const startStakeFactory = async (
...
+      governedContractInstallation: installations.stakeFactory,
+      governed: harden({
+        terms: stakeFactoryTerms,
+        issuerKeywordRecord: { Stake: bldIssuer },
+        privateArgs: {
+          feeMintAccess,
+          initialPoserInvitation,
+          lienBridge,
+          storageNode,
+          marshaller,
```
erights, line 895: Re line 890 above

Terms are not private. Perhaps I am missing context, but it seems wrong that something called somethingTerms contains something called privateArgs. Doesn't this cause the allegedly private args to be exposed to anyone with access to these terms?

turadg (approved):
> Be sure to hide whitespace when looking at these diffs.

ha, well whitespace was the main reason I thought it worthwhile to have the helper. to avoid deeper indentation. in hindsight not a great reason.

I considered that `harden` shouldn't be assumed to be done by the called function, but since these are only for proposal setup I thought they could be trusted. But I take your point that we should be careful not to habituate to that.

I approve of this change but request that the commit message be corrected before landing. Sufficient to change the PR title to something like "refactor: remove deeplyFulfillTerms" and squash.

Chris-Hibbert:
I'm not a fan of this.   I think every time `deeplyFulfill` is called on `terms`, `harden` should also be called.

Given that, the helper seems a better way of packaging them than requiring two calls.

And it's not obvious to me why `deeplyFulfill(harden())` works, though it's clear that since `deeplyFulfill`  requires an `await`,  `harden(deeplyFulfill())` won't work.  I think the required format `await deeplyFulfill(harden(... terms...));` will be inobvious to  many of our users.

turadg:
> I think every time deeplyFulfill is called on terms, harden should also be called.

By my understanding, not just "should" but "must". It will fail otherwise, no?

> Given that, the helper seems a better way of packaging them than requiring two calls.

I don't follow this reasoning. Here are the reasons I've seen yet:

? How should contract starters deeply fulfill their terms?
1. : By calling `deeplyFulfill(harden(` 
1.1. - Makes more indentation
1.2 - not obvious to users that they should
1.2.1 . We're just talking about proposal starting, which is already a special knowledge. Mostly done by looking at another to see what to do
2. : With a helper named `deeplyFulfillTerms`
2.1. - hinders static analysis for Jessie 
2.2. - fosters a habit of not visibly and obviously hardening as early as possible
3. : With a helper named `hardenAndDeeplyFulfill`
3.1. + helper for common case
3.2. .  might be enough to  harden is happening

Chris-Hibbert:
> 2.1. - hinders static analysis for Jessie

I was expecting/hoping that you would come up with TS magic that would pass the types through and add an indication that the hardening had happened. If that's not possible, it does narrow our options.

turadg:
> I was expecting/hoping that you would come up with TS magic that would pass the types through and add an indication that the hardening had happened. If that's not possible, it does narrow our options.

🧙 ! You're right, we could tag the objects as hardened. Then parameters that are required to hardened could be typed as requiring that tag.

But if a function takes parameters that _may_ be hardened, then the caller should still be in the habit of hardening. Right?


Zooming out, why does `deeplyFulfilled` throw if its params aren't hardened? Is hardening really necessary in these proposal fulfillments? (It wasn't happening before with `Promise.all`).

erights:
> > I was expecting/hoping that you would come up with TS magic that would pass the types through and add an indication that the hardening had happened. If that's not possible, it does narrow our options.
> 
> 🧙 ! You're right, we could tag the objects as hardened. Then parameters that are required to hardened could be typed as requiring that tag.

The Jessie checking rules I have in mind have yet to be carefully defined. But I hope they can generally be sound, or at least that we can understand what about them is and is not sound, and that the rules around hardening can be sound. We won't know until we put significantly more effort into this. OTOH, we know that TS is unsound, and we do not currently have any good understanding of some subset of TS issues for which we have any guaranteed soundness. We should not have Jessie safety depend on TS checking, at least until something big changes.

> But if a function takes parameters that _may_ be hardened, then the caller should still be in the habit of hardening. Right?

Although I'm guilty of doing so in a few places, having the callee `harden` can be a bad practice because it masks bugs: it hides more cases where an object should have already been hardened. 

> Zooming out, why does `deeplyFulfilled` throw if its params aren't hardened? Is hardening really necessary in these proposal fulfillments? (It wasn't happening before with `Promise.all`).

The reason `deeplyFulfilled` requires its inputs are hardened is that `deeplyFulfilled` is defined only over Passable values. By contrast, `Promise.all` is defined by JS, which, obviously, has no notion of Passable or harden.

turadg:
EDIT: nevermind, Chris clarified below

> > Zooming out, why does deeplyFulfilled throw if its params aren't hardened? Is hardening really necessary in these proposal fulfillments? (It wasn't happening before with Promise.all).
> 
> The reason deeplyFulfilled requires its inputs are hardened is that deeplyFulfilled is defined only over Passable values. 

What I'm getting at is there's a fourth option in which `deeplyFulfillTerms` (or simply `fulfillTerms` because we're only needing one level deep) doesn't harden at all (by eschewing `deeplyFulfill`). I suspect that if #5891 had made that it wouldn't have provoked the harden discipline.

Chris-Hibbert:
> I suspect that if [5891](https://github.com/Agoric/agoric-sdk/pull/5891) had made that it wouldn't have provoked the harden discipline.

what 5891 needed was for everything in `terms` to be durable, which requires `deeplyFulfilled`. No promises, all the way down. I think `deeplyFulfilled` requires and returns a hardened structure.

erights:
> > I suspect that if [5891](https://github.com/Agoric/agoric-sdk/pull/5891) had made that it wouldn't have provoked the harden discipline.
> 
> what 5891 needed was for everything in `terms` to be durable, which requires `deeplyFulfilled`. No promises, all the way down. I think `deeplyFulfilled` requires and returns a hardened structure.

All durable things are passable anyway. In general, when the outer literal of a tree of syntactically nested literals is an object literal `{...}`, then it should *usually* have a recognized hardener around it: `harden({...})`, `Far(..., {...})`, `*Kind*(..., () => ({...}), {...})`. If we just start seeing these recognized hardeners as a normal part of our syntax, and stop seeing bare literals as normal, perhaps then we can stop arguing about it so much.

Admittedly, such rules should be backed by static checking. But between now and MN1 I doubt anyone will or should spend time writing such a static checker. Nevertheless, we should not write code that obviously violates such static checking rules, and we should try to catch violations during review.

erights:
>> what 5891 needed was for everything in `terms` to be durable,

> All durable things are passable anyway

Completing the thought: and so must be hardened anyway.

Chris-Hibbert:
This change has been incorporated into #5891 in [commit `a8db354`](https://github.com/Agoric/agoric-sdk/pull/5891/commits/a8db35404777d1f9e385a7ce9ba1add827e4cff6)

turadg:
Note that because of https://github.com/endojs/endo/issues/1257 this change resulted in loss of types.

Chris-Hibbert:
@turadg  Did `deeplyFulfillTerms` do a better job of transmitting types than `deeplyFulfill` does? That would surprise me.

turadg:
> Did deeplyFulfillTerms do a better job of transmitting types than deeplyFulfill does? That would surprise me.

I don't recall, but it could since it didn't need to be as general as `deeplyFulfilled`. It also was agoric-sdk code so much easier to fix. Though I suppose we could still being it back as an alias:

```js
/**
  * @template {{}} T
  * @param {T} unfulfilledTerms
  * @returns {Awaited<T>}
  */
const deeplyFulfilledTerms = deeplyFulfilled;
```

erights:
> didn't need to be as general as deeplyFulfilled.

How? IIUC, the difference was only the `harden` should shouldn't effect the typing problem much one way of the other.


```js
/**
  * @template {{}} T
  * @param {T} unfulfilledTerms
  * @returns {Awaited<T>}
  */
const deeplyFulfilledTerms = deeplyFulfilled;
```

Is `Awaited` recursive over the structure of `T`?

Issue #5914, opened 2022-08-08
Support suggestIssuer in smart wallet


samsiegart:
## What is the Problem Being Solved?

For adding liquidity (#5077), the AMM needs to be able to call suggestIssuer on the wallet bridge and have the wallet import the issuer and create a purse for it.

## Description of the Design

Implement the `suggestInstance` method in [ScopedBridge.js](https://github.com/Agoric/agoric-sdk/blob/master/packages/wallet/ui/src/service/ScopedBridge.js) which shows a request in the user's inbox to import the issuer. It can also have inputs for the issuer petname and purse petname. When the user accepts the request, it can trigger a signed wallet action to call `suggestIssuer` on the wallet contract.

## Security Considerations

The scoped bridge is accessible to the dapps only once the users have approved the dapp, and the issuer will be added to the smart wallet only after the user approves it

PR #5935, opened 2022-08-11
publication record for AMM liquidityIssuer

turadg:
unblocks: https://github.com/Agoric/agoric-sdk/issues/5914

## Description

The wallet needs a way to read the liquidityIssuer for an AMM pool without a chain transaction.

This puts it on a new storage node, `init` under each pool's own storage node. `init` is chosen to convey that its set during initialization and that it will never change. (Implemented by `.finish` and not holding the publisher ref)

### Testing Considerations

@dckc tested manually: https://github.com/Agoric/agoric-sdk/pull/5935/files#r944704134

This change still needs a CI test. That'll take a new testing ability so I'm working on that independently.

### DIFF
--- a/packages/inter-protocol/src/vpool-xyk-amm/pool.js
+++ b/packages/inter-protocol/src/vpool-xyk-amm/pool.js
@@ -11,6 +11,7 @@ import {
   calcValueToRemove,
 } from '@agoric/zoe/src/contractSupport/index.js';
 
+import { E } from '@endo/eventual-send';
 import { makeMetricsPublisherKit } from '../contractSupport.js';
 import { makePriceAuthority } from './priceAuthority.js';
 import { makeSinglePool } from './singlePool.js';
@@ -22,6 +23,8 @@ import { makeSinglePool } from './singlePool.js';
 export const publicPrices = prices => {
   return { amountIn: prices.swapperGives, amountOut: prices.swapperGets };
 };
+
+/** @typedef {{ liquidityIssuerRecord: IssuerRecord<'nat'> }} InitPublication */
 
 /** @typedef {{ central: Amount, secondary: Amount }} NotificationState */
 
@@ -87,7 +90,14 @@ export const definePoolKind = (baggage, ammPowers, storageNode, marshaller) => {
    * @returns {ImmutableState & MutableState}
    */
   const poolInit = (liquidityZcfMint, poolSeat, secondaryBrand) => {
-    /** @type {PublishKit<{secondary: Amount, central: Amount}>} */
+    /** @type {StoredPublishKit<InitPublication>} */
+    const { publisher: initPublisher } = makeStoredPublishKit(
+      E(storageNode).getChildNode('init'),
+      marshaller,
+    );
+    initPublisher.finish({
+      liquidityIssuerRecord: liquidityZcfMint.getIssuerRecord(),
+    });
 
     return {
       liqTokenSupply: 0n,
--- a/packages/inter-protocol/src/vpool-xyk-amm/pool.js
+++ b/packages/inter-protocol/src/vpool-xyk-amm/pool.js
@@ -3,14 +3,15 @@
 import '@agoric/zoe/exported.js';
 
 import { AmountMath, isNatValue } from '@agoric/ertp';
-import { makeStoredPublisherKit, makeStoredPublishKit } from '@agoric/notifier';
+import { makeStoredPublishKit } from '@agoric/notifier';
+import { vivifyKindMulti } from '@agoric/vat-data';
 import {
   calcLiqValueToMint,
   calcSecondaryRequired,
   calcValueToRemove,
 } from '@agoric/zoe/src/contractSupport/index.js';
-import { vivifyKindMulti } from '@agoric/vat-data';
 
+import { makeMetricsPublisherKit } from '../contractSupport.js';
 import { makePriceAuthority } from './priceAuthority.js';
 import { makeSinglePool } from './singlePool.js';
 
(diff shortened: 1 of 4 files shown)


packages/inter-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -92,7 +90,14 @@ export const definePoolKind = (baggage, ammPowers, storageNode, marshaller) => {
...
    */
   const poolInit = (liquidityZcfMint, poolSeat, secondaryBrand) => {
-    /** @type {PublishKit<{secondary: Amount, central: Amount}>} */
+    /** @type {StoredPublishKit<InitPublication>} */
+    const { publisher: initPublisher } = makeStoredPublishKit(
+      E(storageNode).getChildNode('init'),
+      marshaller,
+    );
+    initPublisher.finish({
+      liquidityIssuerRecord: liquidityZcfMint.getIssuerRecord(),
```
samsiegart, line 99: We need the issuer's id. So `E(board).getId(liquidityZcfMint.getIssuerRecord().issuer)` effectively I think

samsiegart (dismissed):

packages/inter-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -92,7 +90,14 @@ export const definePoolKind = (baggage, ammPowers, storageNode, marshaller) => {
    * @returns {ImmutableState & MutableState}
    */
   const poolInit = (liquidityZcfMint, poolSeat, secondaryBrand) => {
-    /** @type {PublishKit<{secondary: Amount, central: Amount}>} */
+    /** @type {StoredPublishKit<InitPublication>} */
+    const { publisher: initPublisher } = makeStoredPublishKit(
+      E(storageNode).getChildNode('init'),
```
samsiegart, line 95: Can we mock a storage node and/or makeStoredPublishKit for a test?

packages/inter-protocol/src/vpool-xyk-amm/pool.js
turadg: Hmm, we don't have a board here. With this way that gives you the whole issuer record, can the client look up the ID in the board?

samsiegart: There's no way for the client to look up the id in the board with rpc, we run into the same issue where that takes a chain transaction (and we don't even fully have a way to do such chain transactions yet).

samsiegart: I assume we can pass the board through from the contract, but perhaps that's too messy, which I was afraid would be the case.

dckc: @samsiegart the dapp gets the issuer the same way it gets brands: by deserializing using the `importContext`.

dckc: Ah... objects deserialized with `importContext` are `SEVERED`; i.e. you can't actually call any methods on them. And we need an issuer that supports `E(issuer).getEmptyPurse()`. I think you have this working with a hard-coded board-id, right @samsiegart ? You can get the appropriate boardId by manually parsing the capData published in the storage node.

@turadg do we have an example of what `agoric follow` or `agd query vstorage ...` outputs for this `.init` node?

dckc: manually parsing... I suppose the way to do that is to use yet another deserializer when creating this follower.

samsiegart: The `suggestIssuer` method requires the issuer id specifically, not the issuer itself. The user needs to be able to verify the exact id before they give it a petname for their wallet.

samsiegart: And yes I have it working with a hardcoded boardId. The wallet contract handles looking up the issuer by its id and calling `E(issuer).getEmptyPurse()`

dckc: > The suggestIssuer method requires the issuer id specifically, not the issuer itself.

got it... then a trivial `x => x` should work for `convertSlotToVal`.

dckc: @samsiegart I just checked out this branch and fired up the chain; see where the boardid for the LP issuer is?

```
$ ~/go/bin/agd query vstorage data published.amm.pool0.init -o json | jq '.value | fromjson | {body: .body | fromjson, slots}'
{
  "body": {
    "liquidityIssuerRecord": {
      "assetKind": "nat",
      "brand": {
        "@qclass": "slot",
        "iface": "Alleged: IbcATOMLiquidity brand",
        "index": 0
      },
      "displayInfo": {
        "assetKind": "nat",
        "decimalPlaces": 6
      },
      "issuer": {
        "@qclass": "slot",
        "iface": "Alleged: IbcATOMLiquidity issuer",
        "index": 1
      }
    }
  },
  "slots": [
    "board0188",
    "board02810"
  ]
}
```

dckc (approved):
I tested it manually; looks good.

packages/inter-protocol/src/vpool-xyk-amm/pool.js
turadg: https://github.com/Agoric/agoric-sdk/pull/5948

turadg:
[Chain deployment test failed](https://github.com/Agoric/agoric-sdk/runs/7812850658?check_suite_focus=true) so I pushed a small change ([791df44](https://github.com/Agoric/agoric-sdk/pull/5935/commits/791df446f471b61ef961e36b96097b982ddb6f6c)) from https://github.com/Agoric/agoric-sdk/pull/5940 to trigger another ubild

PR #5936, opened 2022-08-11
fix(vats): avoid sending large contract bundles to Zoe.install()

dckc:
refs: #5910, #4374

## Description

in `installBootContracts`, we were using `getNamedBundleCap` and `getBundle` resulting in a large bundle of source code sent to `E(zoe).install(bundle)`. Now that the infrastructure is available, let's use `getBundleIDByName` and `E(zoe).installBundleID(bundleID)`.

### Security Considerations

contributes to our ability to put a cap on message size

the type for `VatAdminSvc` was missing a method.

### Testing Considerations

unit tests needed some adjustment.

### DIFF
--- a/packages/vats/src/core/basic-behaviors.js
+++ b/packages/vats/src/core/basic-behaviors.js
@@ -285,27 +285,23 @@ harden(makeClientBanks);
 
 /** @param {BootstrapSpace & { devices: { vatAdmin: any }, vatPowers: { D: DProxy }, }} powers */
 export const installBootContracts = async ({
-  vatPowers: { D },
-  devices: { vatAdmin },
-  consume: { zoe },
+  consume: { zoe, vatAdminSvc },
   installation: {
     produce: { centralSupply, mintHolder, walletFactory },
   },
 }) => {
-  for (const [name, producer] of Object.entries({
-    centralSupply,
-    mintHolder,
-    walletFactory,
-  })) {
-    // This really wants to be E(vatAdminSvc).getBundleIDByName, but it's
-    // good enough to do D(vatAdmin).getBundleIDByName
-    const bundleCap = D(vatAdmin).getNamedBundleCap(name);
+  return Promise.all(
+    Object.entries({
+      centralSupply,
+      mintHolder,
+      walletFactory,
+    }).map(async ([name, producer]) => {
+      const bundleID = await E(vatAdminSvc).getBundleIDByName(name);
 
-    const bundle = D(bundleCap).getBundle();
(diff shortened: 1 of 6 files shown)


dckc:
FYI, @arirubinstein : @warner estimates this should shave some 6MB from the Zoe vat.
(3 contracts x ~2MB/contract)

warner:
> `closes: https://github.com/Agoric/agoric-sdk/issues/4374`

Actually there's a second half to that ticket, I just added some comments on what's necessary to implement that half. I figure we'll defer it because (unlike this change) it doesn't save any memory or storage, but we should probably change this to a `refs: #4374` and double check that the ticket isn't actually closed when this PR lands.

packages/vats/test/test-boot.js
```diff
@@ -103,13 +103,16 @@ const testRole = (ROLE, governanceActions) => {
       { argv: argvByRole[ROLE], governanceActions },
     );
 
-    const fakeVatAdmin = makeFakeVatAdmin(() => {}).admin;
+    const { admin: fakeVatAdmin } = makeFakeVatAdmin(() => {});
```
warner, line 106: you could revert this if you wanted

packages/vats/test/test-boot.js
```diff
@@ -103,13 +103,16 @@ const testRole = (ROLE, governanceActions) => {
...
     const fakeBundleCaps = new Map(); // {} -> name
     const getNamedBundleCap = name => {
       const bundleCap = harden({});
       fakeBundleCaps.set(bundleCap, name);
       return bundleCap;
     };
+    const getBundleIDByName = name => `b1-${name}`;
+    const getBundleCap = id => devices.vatAdmin.getBundleCap(id);
```
warner, line 114: Would it make sense for this to be `const getBundleCap = devices.vatAdmin.getBundleCap`? Or does that obscure the fact that we're defining a function here?

packages/vats/src/core/basic-behaviors.js
```diff
@@ -274,27 +274,23 @@ harden(makeClientBanks);
...
-    producer.resolve(installation);
-  }
+  return Promise.all(
+    Object.entries({
+      centralSupply,
+      mintHolder,
+      walletFactory,
+    }).map(async ([name, producer]) => {
+      const bundleID = await E(vatAdminSvc).getBundleIDByName(name);
+
```
warner, line 300: nit: I'd remove this blank line

warner (approved):
Looks good to me, I don't know how to help with the type complaints. Ping me if you want to pair on fixing that unit test issue.

dckc:
overtaken by #6896

PR #5939, opened 2022-08-12
feat(wallet/api): marshal brand and issuer board ids

samsiegart:


dckc (approved):
yes, this looks like what we figured out during prototyping

PR #5940, opened 2022-08-12
fix: @param decls were ordered wrong

erights:


turadg:
Thanks. I'll take care of this in another PR so it doesn't need a whole merge train for this.

Note per conventional commits "fix" will bump the patch release. This is "chore" or "docs".

This appeared as a warning in your editor? I'll look into how many of those we have. Maybe it can be made an error to be prevented from merge.

erights:
> This appeared as a warning in your editor?

I no longer reliably remember, but I think I saw this during a `yarn lint-fix | grep -v await` from the top of agoric-sdk

turadg:
done elsewhe

PR #5941, opened 2022-08-12
fix: harden the right thing

erights:
I think I had just made a trivial copy paste error. Trivial and obviously correct fix.


FUDCo (approved):
Like, deep, man.

PR #5942, opened 2022-08-12
feat!: Support publishing and consuming chain storage stream cells

### LINKED ISSUE #5366: Support off-chain consumption of historical chainStorage-backed subscriptions

### LINKED ISSUE #5508: include block height, timestamp in chain-streams

gibson042:
Fixes #5366
Fixes #5508

## Description

Adds support for publishing and opportunistically consuming { height, values } chain storage stream cells in place of naked values, and switches the "published" chain storage subtree to use them.

I'm not sure how casting is intended to behave with batches, and confused about the `committer.isValid()` guard.

### Security Considerations

There's a minor risk of stored data getting confused for a stream cell, but since everything is currently under our control, that should be fine until we can eliminate naked values (at least in the "published" subtree).

I'm not sure how to document the new structure, which I'm calling a stream cell per https://github.com/Agoric/agoric-sdk/tree/mfig-vstream/golang/cosmos/x/vstream/spec .

### Testing Considerations

It looks like there are no current tests for the Go aspects, and how to add some isn't clear.


scripts/get-flattened-publication.sh
```diff
@@ -60,17 +60,26 @@ curl -sS "$URL_PREFIX" --request POST --header 'Content-Type: application/json'
...
-  # Add block height.
-  (.blockHeight |= $height) |
+  # Upgrade a naked value to a stream cell if necessary.
+  if has("height") and has("values") then . else { values: [ . | tojson ] } end |
+
+  # Capture data block height.
+  .height as $dataHeight |
+
+  # Flatten each value independently.
+  .values[] | fromjson |
```
dckc, line 90: I wondered if this change had any impact on this script. Do we have any tests that would have failed if we didn't remember to update this script?

packages/casting/src/follower-cosmjs.js
```diff
@@ -334,18 +334,36 @@ export const makeCosmjsFollower = (
...
+      // Upgrade a naked value to a JSON stream cell if necessary.
+      if (!streamCell.height || !streamCell.values) {
+        streamCell = { values: [JSON.stringify(streamCell)] };
       }
-      const value = await E(unserializer).unserialize(data);
-      if (!committer.isValid()) {
-        return;
+      for (let i = 0; i < streamCell.values.length; i += 1) {
+        const data = JSON.parse(streamCell.values[i]);
+        const last = i + 1 === streamCell.values.length;
```
turadg: nit, `last` sounds like the last element. consider `isLast`.

packages/casting/src/follower-cosmjs.js
```diff
@@ -334,18 +334,36 @@ export const makeCosmjsFollower = (
...
+        if (!unserializer) {
+          /** @type {T} */
+          const value = data;
+          committer.commit({ value });
+          if (!last) {
+            committer = prepareUpdateInOrder();
+          }
+          // eslint-disable-next-line no-continue
+          continue;
```
turadg: I wonder whether we should globally disable this rule. If it's a worthwhile rule to follow, consider refactoring so this suppression isn't necessary. E.g,
```js
      streamCells.values.forEach((dataStr, i) => {
        const data = JSON.parse(dataStr);
        const isLast = i + 1 === streamCell.values.length;
```
so that this can be,
```suggestion
          if (!last) {
            committer = prepareUpdateInOrder();
          }
          return;
```

packages/casting/test/fake-rpc-server.js
```diff
@@ -64,10 +64,19 @@ const fakeStatusResult = {
   },
 };
 
-export const startFakeServer = (t, fakeValues, marshaller = makeMarshal()) => {
+/** @typedef {Partial<import('ava').ExecutionContext<{cleanups: Array<() => void>}>> & {context}} fakeServerTestContext */
```
turadg: 😍

packages/casting/test/test-mvp.js
```diff
@@ -12,39 +12,60 @@ import {
...
-  const lo = {
-    retryCallback: null, // fail fast, no retries
-    keepPolling: () => delay(200).then(() => true), // poll really quickly
-    jitter: null, // no jitter
-  };
-  /** @type {import('../src/types.js').FollowerOptions} */
-  const so = {
-    proof: 'none',
-  };
+// TODO: Replace with test.macro({title, exec}).
```
turadg, line 15: 👍

packages/vats/src/lib-chainStorage.js
```diff
@@ -38,11 +38,14 @@ harden(sanitizePathSegment);
  * @param {(message: StorageMessage) => any} handleStorageMessage a function for sending a storageMessage object to the storage implementation (cf. golang/cosmos/x/vstorage/vstorage.go)
  * @param {string} storeName currently limited to "swingset"
```
turadg: drive-by: type is effectively `{'swingset'}`

packages/vats/test/test-lib-chainStorage.js
```diff
@@ -3,8 +3,9 @@ import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
 
 import { makeChainStorageRoot } from '../src/lib-chainStorage.js';
 
-test('makeChainStorageRoot', async t => {
-  // Instantiate chain storage over a simple in-memory implementation.
+// makeFakeChainStorageKit creates a chainStorage root node over an in-memory map
+// and exposes both the map and the sequence of received messages.
+const makeFakeChainStorageKit = (rootPath, options) => {
```
turadg: We have more than one kind of fake storage. How about `makeTransientChainStorageKit`?  Not sure we even need "chain" https://github.com/Agoric/agoric-sdk/pull/5945

packages/vats/test/test-lib-chainStorage.js
```diff
@@ -3,8 +3,9 @@ import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
 
 import { makeChainStorageRoot } from '../src/lib-chainStorage.js';
 
-test('makeChainStorageRoot', async t => {
-  // Instantiate chain storage over a simple in-memory implementation.
+// makeFakeChainStorageKit creates a chainStorage root node over an in-memory map
+// and exposes both the map and the sequence of received messages.
```
turadg: better as jsdoc. can include param types as well

scripts/get-flattened-publication.sh
turadg: I'm pretty sure we don't have any tests of shell scripts, and I think that's fine. They're not exported from the packages and not part of the API of the package. They're just here to help organize the code.

turadg (commented):
Reviewed the chain JS. I don't know enough about the Go parts to review so not approving the PR. 

If you'd like me to dive in enough to do a full review let's discuss because I'll have to reprioritize some stuff.

packages/vats/test/test-lib-chainStorage.js
turadg: Also, WDYT of this in `lib-chainStorage` for use by other packages?

scripts/get-flattened-publication.sh
dckc: It's perhaps not exported in the npm sense, but it's a customer-visible API (customer-required, in fact). I would like to see an integration test where
  - javascript code writes capData to a file
  - this script reads the capData file and writes flat data
  - we diff the flat data against expected results

Of course, I'd like to see a lot of things... I'm not sure whether it's cost-effective any time soon. It's a question of how often we're likely to tweak this script.

dckc: Perhaps it _should_ be exported in the npm sense? that is: perhaps it should go in `.bin`?

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -11,6 +13,19 @@ import (
...
+// It is persisted to storage as a { "height": "<digits>", "values": ["...", ...] } JSON text
+// that off-chain consumers rely upon.
+type StreamCell struct {
+	Height string `json:"height"`
+	// XXX Should Values be []string or []interface{}?
+	// The latter would remove a layer of JSON encoding (e.g., `[{…}]` rather than `["{…}"]`,
+	// but would add a requirement exclusive to AppendStorageValueAndNotify that its input be JSON.
+	// On the other hand, we could always extend this format in the future to include an indication
+	// that values are subject to a different encoding, e.g. `"valueEncoding":"base64"`.
+	Values []string `json:"values"`
```
kriskowal: I think `string` also limits cells to JSON (or at least text formats) in a way that `[]byte` would not. I think we should commit to JSON. Strings might be better since interface{} has a lot more liberty to mess up 64 bit integers among other things.

scripts/get-flattened-publication.sh
gibson042: The problem with a test like you describe above is that it's mostly navel-gazing—there's no connection between the hard-coded input and the kind of output that is expected of RPC nodes in the real world. But I do still thank that having _something_ is a good idea.

golang/cosmos/x/vstorage/keeper/keeper.go
```diff
@@ -127,6 +142,31 @@ func (k Keeper) SetStorageAndNotify(ctx sdk.Context, path, value string) {
...
 }
 
+func (k Keeper) AppendStorageValueAndNotify(ctx sdk.Context, path, value string) error {
+	height := strconv.FormatInt(ctx.BlockHeight(), 10)
+
+	// Preserve correctly-formatted data within the current block,
+	// otherwise initialize a blank cell.
+	currentData := k.GetData(ctx, path)
+	var cell StreamCell
+	_ = json.Unmarshal([]byte(currentData), &cell)
```
kriskowal, line 149: Error handling

packages/casting/src/follower-cosmjs.js
```diff
@@ -334,18 +334,36 @@ export const makeCosmjsFollower = (
...
       lastBuf = buf;
-      const data = decode(buf);
-      if (!unserializer) {
-        /** @type {T} */
-        const value = data;
-        committer.commit({ value });
-        return;
+      let streamCell = decode(buf);
+      // Upgrade a naked value to a JSON stream cell if necessary.
+      if (!streamCell.height || !streamCell.values) {
```
kriskowal: Is 0 a valid height?

packages/casting/src/follower-cosmjs.js
```diff
@@ -334,18 +334,36 @@ export const makeCosmjsFollower = (
...
+          continue;
+        }
+        // eslint-disable-next-line no-await-in-loop,@jessie.js/no-nested-await
+        const value = await E(unserializer).unserialize(data);
+        if (!committer.isValid()) {
+          // QUESTION: How would we get here, and what is the proper handling?
+          // eslint-disable-next-line no-continue
+          continue;
+        }
+        committer.commit({ value });
```
kriskowal: What’s the envelope around `value` for? I only see `{ value }` getting committed, not `{ value, done }`, the one likely suspect.

packages/casting/src/follower-cosmjs.js
```diff
@@ -334,18 +334,36 @@ export const makeCosmjsFollower = (
...
+        if (!last) {
+          committer = prepareUpdateInOrder();
+        }
```
kriskowal: Why replace the committer only if final?

packages/casting/test/fake-rpc-server.js
```diff
@@ -64,10 +64,19 @@ const fakeStatusResult = {
...
+/**
+ * @param {fakeServerTestContext} t
```
kriskowal: Is there a reason for the type to have a little f?

packages/casting/test/fake-rpc-server.js
```diff
@@ -64,10 +64,19 @@ const fakeStatusResult = {
...
+ * @param {Array<{any}>} fakeValues
```
kriskowal, line 70: What does `{any}` mean?

packages/casting/test/fake-rpc-server.js
```diff
@@ -64,10 +64,19 @@ const fakeStatusResult = {
...
+ * @param {object} [options]
+ * @param {Marshaller} [options.marshaller]
+ * @param {number} [options.batchSize] count of stream-cell results per response, or 0/absent to return lone naked values
+ */
+export const startFakeServer = (t, fakeValues, options = {}) => {
   const { log = console.log } = t;
   lastPort += 1;
   const PORT = lastPort;
```
kriskowal, line 78: Aside: this usually tells me we should listen on 0 and ask the kernel for the port we were assigned.

kriskowal (approved):

scripts/get-flattened-publication.sh
gibson042: OK, I've added shell-based testing. Any ideas for how to make sure they run with the rest of our tests?

golang/cosmos/x/vstorage/keeper/keeper.go
gibson042: Intentionally suppressed here because we populate `cell` manually if `Unmarshal` fails to do so.

packages/casting/src/follower-cosmjs.js
gibson042: No, because it's not a string. And `"0"` is truthy.

packages/casting/src/follower-cosmjs.js
gibson042: I haven't actually figured that out yet.

packages/casting/src/follower-cosmjs.js
gibson042: Only if _not_ final. And I think replacing it in final position would make it look like there are gaps, although I could be wrong.

packages/casting/test/fake-rpc-server.js
gibson042: No, I'll capitalize it.

packages/casting/test/fake-rpc-server.js
gibson042: IIUC, it imposes no constraints beyond existence.

scripts/get-flattened-publication.sh
dckc: The `test` nom script could call the shell test thing, maybe? (I haven't looked at the shell thing)

arirubinstein:
@gibson042 
I ran into an issue with the jq parse on vault managers:
```
./test.sh https://xnet.rpc.agoric.net published.vaultFactory.manager0.metrics                                                       5 ↵
jq: error (at <stdin>:16): null (null) cannot be matched, as it is not a string
```

gibson042:
@arirubinstein I think that was from `iface` missing in repeated references to the same slot, which should now be fixed. PTAL.

scripts/get-flattened-publication.sh
gibson042: Added a simple runner at [scripts/test/test.sh](https://github.com/Agoric/agoric-sdk/pull/5942/files#diff-686b1343c78734cb242be3e7a388cfe4cbc4c80d6e071327c6b14446518eff2c) (and a [GitHub workflow calling it](https://github.com/Agoric/agoric-sdk/pull/5942/files#diff-efee01ea8ffefa44cbe15acfb318800ac5cceef28e30c1452cb8f84b6070e49f)).

gibson042:
@arirubinstein I'd also like to rename `blockHeight` to `responseBlockHeight`, which would be a breaking change in the script output but provide more clarity in contrast with `dataBlockHeight`. Any objections?

golang/cosmos/x/vstorage/keeper/keeper.go
gibson042: I decided to stick with stream cell values being strings, aligning with the existing interfaces. So from the outside in we have
* JS ChainStorageNode option "sequence" is `false` (default) vs. `true`.
* JS ChainStorageNode method `setValue(value: string)` sends `value` with method "set" vs. "append".
* vstorage.go invokes `keeper.SetStorageAndNotify` vs. `keeper.AppendStorageValueAndNotify`, in either case with the received value.
* AppendStorageValueAndNotify loads or initializes a stream cell `struct { Height string, Values []string }`, appends the value, and calls `SetStorageAndNotify` with the cell's JSON serialization (_this representation of a stream cell as JSON is the opinionated decision_).
* SetStorageAndNotify works down to SetStorage, which performs the actual write of value as a []byte with a static prefix.

arirubinstein (approved):

PR #5943, opened 2022-08-12
style(xsnap): excuse test code from no-nested-await

dckc:
Tidy up some of the lint warnings that show up in _every_ PR.

### DIFF
--- a/packages/xsnap/src/avaAssertXS.js
+++ b/packages/xsnap/src/avaAssertXS.js
@@ -1,5 +1,5 @@
 /* global globalThis */
-/* eslint-disable no-await-in-loop */
+/* eslint-disable no-await-in-loop, @jessie.js/no-nested-await -- test code */
 // @ts-check
 
 /** global print */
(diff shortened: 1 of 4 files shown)


packages/xsnap/src/avaAssertXS.js
```diff
@@ -1,4 +1,6 @@
 /* global globalThis */
+/* We make exceptions for test code. This is a test utility. */
+/* eslint-disable @jessie.js/no-nested-await */
 /* eslint-disable no-await-in-loop */
```
turadg: Consider putting on one line:
```suggestion
/* eslint-disable no-await-in-loop, @jessie.js/no-nested-await -- test code */
```

turadg (approved):
Good enough to land.

Though consider before merging:
1. collapsing to one line
2. doing the same for `packages/agoric-cli/src/deploy.js`

dckc:
> ... consider before merging:
> 
> 1. collapsing to one line

agreed / done

> 2. doing the same for `packages/agoric-cli/src/deploy.js`

Though it's not smart contract code, it _is_ product code, not test code. I'm leaving that alone, for now.

PR #5945, opened 2022-08-12
refactor StorageNode type

turadg:
## Description

We have ChainStorageNode and StorageNode but it seems the latter is sufficient. The only difference is the former has `clearValue`, which is the same as `setValue(undefined)`.

This removes `clearValue` and replaces `ChainStorageNode` with `StorageNode`.

It also renames the `makeStorageNode` function to distinguish its use case (making a child) and makes `childName` required.

### Security Considerations

--

--

### Testing Considerations

CI

### DIFF
--- a/packages/notifier/src/types.js
+++ b/packages/notifier/src/types.js
@@ -219,6 +219,14 @@
 /** @typedef {Pick<Marshaller, 'unserialize'>} Unserializer */
 
 /**
+ * This represents a node in an IAVL tree.
+ *
+ * The active implementation is x/vstorage, an Agoric extension of the Cosmos SDK.
+ *
+ * Vstorage is a hierarchical externally-reachable storage structure that
+ * identifies children by restricted ASCII name and is associated with arbitrary
+ * string-valued data for each node, defaulting to the empty string.
+ *
  * @typedef {object} StorageNode
  * @property {(data: string) => void} setValue publishes some data
  * @property {() => ERef<Record<string, any>>} getStoreKey get the
--- a/packages/notifier/src/types.js
+++ b/packages/notifier/src/types.js
@@ -219,6 +219,15 @@
 /** @typedef {Pick<Marshaller, 'unserialize'>} Unserializer */
 
 /**
+ * Defined by vstorageStoreKey in vstorage.go
+ *
+ * @typedef VStorageKey
+ * @property {string} storeName
+ * @property {string} storeSubkey
+ * @property {string} dataPrefixBytes
+ */
+
+/**
  * This represents a node in an IAVL tree.
  *
  * The active implementation is x/vstorage, an Agoric extension of the Cosmos SDK.
@@ -229,7 +238,7 @@
  *
  * @typedef {object} StorageNode
  * @property {(data: string) => void} setValue publishes some data
- * @property {() => ERef<Record<string, any>>} getStoreKey get the
+ * @property {() => ERef<VStorageKey>} getStoreKey get the
  * externally-reachable store key for this storage item
  * @property {(subPath: string) => StorageNode} makeChildNode
  */
--- a/packages/notifier/src/types.js
+++ b/packages/notifier/src/types.js
@@ -194,8 +194,6 @@
  * ```
  * The resulting `localIterable` also supports such remote use, and
  * will return access to the same representation.
- * @property {StorageNode['getStoreKey']} getStoreKey get the
- * externally-reachable store key for this subscription
  */
 
 /**
@@ -250,6 +248,7 @@
  */
 
 /**
+ * @deprecated
  * @template T
  * @typedef {Subscription<T> & StoredFacet} StoredSubscription
  */
--- a/packages/notifier/src/types.js
+++ b/packages/notifier/src/types.js
@@ -231,7 +231,7 @@
  * @property {(data: string) => void} setValue publishes some data
  * @property {() => ERef<Record<string, any>>} getStoreKey get the
  * externally-reachable store key for this storage item
- * @property {(subPath: string) => StorageNode} getChildNode TODO: makeChildNode
+ * @property {(subPath: string) => StorageNode} makeChildNode
  */
 
 /**
--- a/packages/vats/src/lib-chainStorage.js
+++ b/packages/vats/src/lib-chainStorage.js
@@ -84,7 +84,6 @@ export function makeChainStorageRoot(
   const rootNode = makeChainStorageNode(rootPath);
   return rootNode;
 }
-/** @typedef {ReturnType<typeof makeChainStorageRoot>} ChainStorageNode */
 
 /**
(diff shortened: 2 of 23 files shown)


packages/vats/src/lib-chainStorage.js
```diff
@@ -100,16 +96,13 @@ const makeNullStorageNode = () => {
...
  * behavior) when that is unavailable.
  *
- * @param {ERef<ChainStorageNode?>} chainStorage
- * @param {string} [childName]
+ * @param {ERef<StorageNode?>} chainStorage
+ * @param {string} childName
  * @returns {Promise<StorageNode>}
  */
-export async function makeStorageNode(chainStorage, childName) {
+export async function makeStorageNodeChild(chainStorage, childName) {
```
gibson042: I support this rename, and would recommend taking it farther:
```suggestion
export async function makeStorageNodeChild(storageNode, childName) {
```

packages/vats/src/lib-chainStorage.js
```diff
@@ -36,7 +36,7 @@ harden(sanitizePathSegment);
  * Create a root storage node for a given backing function and root path.
  *
  * @param {(message: StorageMessage) => any} handleStorageMessage a function for sending a storageMessage object to the storage implementation (cf. golang/cosmos/x/vstorage/vstorage.go)
```
gibson042: Given this relationship with vstorage.go, I'd prefer to keep the `ChainStorageNode` type and drop `StorageNode` rather than vice versa.

packages/vats/src/lib-chainStorage.js
```diff
@@ -68,9 +68,6 @@ export function makeChainStorageRoot(
         assert.typeof(value, 'string');
         handleStorageMessage({ key: path, method: 'set', value });
       },
-      clearValue() {
-        handleStorageMessage({ key: path, method: 'set' });
-      },
```
gibson042: This is a hard pill to swallow because there ought to be a conceptual difference between "set to empty" vs. "unset", but factually that is not the case in vstorage.go. So I do not object.

packages/vats/test/test-lib-chainStorage.js
```diff
@@ -68,7 +75,7 @@ test('makeChainStorageRoot', async t => {
     );
   }
 
-  rootNode.clearValue();
+  rootNode.setValue(undefined);
```
gibson042: This call is invalid.
```suggestion
  rootNode.setValue('');
```

packages/vats/test/test-lib-chainStorage.js
```diff
@@ -172,22 +173,22 @@ test('makeChainStorageRoot', async t => {
     'level-skipping setValue message',
   );
 
-  childNode.clearValue();
+  childNode.setValue(undefined);
```
gibson042:
```suggestion
  childNode.setValue('');
```

packages/vats/test/test-lib-chainStorage.js
```diff
@@ -172,22 +173,22 @@ test('makeChainStorageRoot', async t => {
...
   t.deepEqual(
     messages.slice(-1),
     [{ key: childPath, method: 'set' }],
-    'child clearValue message',
+    'child setValue message',
   );
-  deepNode.clearValue();
+  deepNode.setValue(undefined);
```
gibson042:
```suggestion
  deepNode.setValue('');
```

packages/vats/test/test-lib-chainStorage.js
```diff
@@ -172,22 +173,22 @@ test('makeChainStorageRoot', async t => {
...
   t.deepEqual(
     messages.slice(-1),
     [{ key: deepPath, method: 'set' }],
-    'granchild clearValue message',
+    'granchild setValue message',
   );
-  childNode.clearValue();
+  childNode.setValue(undefined);
```
gibson042:
```suggestion
  childNode.setValue('');
```

gibson042 (changes requested):
I like the overall direction, but would prefer to keep the `ChainStorageNode` type rather than `StorageNode` (to indicate the dependency upon chain storage semantics).

Also, since you're doing this, how about moving the type definition from packages/notifier/src/types.js to a more logical home, and also providing a definition of what instances of the type represent (e.g., "_a node in a hierarchical externally-reachable storage structure that identifies children by restricted ASCII name and is associated with arbitrary string-valued data defaulting to the empty string_")?

packages/notifier/src/types.js
```diff
@@ -219,20 +217,36 @@
...
+ * @property {string} storeSubkey
+ * @property {string} dataPrefixBytes
+ */
+
+/**
+ * This represents a node in vstorage, an Agoric extension of the Cosmos SDK.
+ *
+ * Vstorage is a hierarchical externally-reachable storage structure that
+ * identifies children by restricted ASCII name and is associated with arbitrary
+ * string-valued data defaulting to the empty string.
```
gibson042:
```suggestion
 * string-valued data for each node, defaulting to the empty string.
```

packages/vats/src/lib-chainStorage.js
```diff
@@ -53,27 +53,27 @@ export function makeChainStorageRoot(
 
   function makeChainStorageNode(path) {
     const node = {
+      /** @type {() => VStorageKey} */
       getStoreKey() {
         return handleStorageMessage({ key: path, method: 'getStoreKey' });
       },
-      getChildNode(name) {
+      /** @type {(name: string) => VStorageNode} */
+      makeChildNode(name) {
```
gibson042, line 63: In Agoric conventions that I learned in the past few weeks, I think the most appropriate name for this method is `provideChildNode` ("provide" referring to operations that return a value after creating it if necessary, acknowledging that it might have already existed).

And likewise for `makeStorageNodeChild`.

gibson042 (approved):
Thanks!

packages/vats/src/lib-chainStorage.js
turadg: I think with `provide` you get the same object on repeated requests. The two methods you suggest renaming actually call `makeChainStorageNode` which always makes a fresh node. Changing that is out of scope but I can follow up to do that. WDYT @michaelfig ?

turadg:
I just noticed https://github.com/Agoric/agoric-sdk/blob/a457f750fd7462e84e92980475f68e8f66433df7/packages/casting/src/casting-spec.js#L44-L48 and now I'm not so sure about removing `StorageNode`. 

@michaelfig are there meant to be more than one backing of `StorageNode`? How do `swingset` and `vstore` relate? Is it that `swingset` is chain and `vstore` is off-chain?

turadg:
@Mergifyio unqueue

mhofman:
@mergifyio queue

mhofman:
@Mergifyio refresh

mhofman:
@mergifyio queue master

PR #5946, opened 2022-08-12
feat(wallet): smart wallet performAction

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/5914

### DIFF
--- a/packages/wallet/api/src/findOrMakeInvitation.js
+++ b/packages/wallet/api/src/findOrMakeInvitation.js
@@ -94,9 +94,23 @@ const makeContinuingInvitation = async (
   return invitationP;
 };
 
+const makeInvitation = async (
+  invitationMaker,
+  instanceHandleBoardId,
+  board,
+  zoe,
+) => {
+  const instance = E(board).getValue(instanceHandleBoardId);
+  const publicFacet = E(zoe).getPublicFacet(instance);
+  const { description, args = [] } = invitationMaker;
+
+  return E(publicFacet)[description](...args);
+};
+
 export const findOrMakeInvitation = async (
   idToOfferResultPromiseKit,
   board,
+  zoe,
   invitationPurse,
   invitationBrand,
   offer,
@@ -124,6 +138,15 @@ export const findOrMakeInvitation = async (
       boardId: offer.invitationHandleBoardId,
     };
     return findInvitation(findByBoardId, queryParams);
+  }
+
+  if (offer.invitationMaker) {
+    return makeInvitation(
+      offer.invitationMaker,
+      offer.instanceHandleBoardId,
+      board,
+      zoe,
+    );
   }
 
   if (offer.invitationQuery) {
--- a/packages/wallet/api/src/findOrMakeInvitation.js
+++ b/packages/wallet/api/src/findOrMakeInvitation.js
@@ -94,6 +94,14 @@ const makeContinuingInvitation = async (
   return invitationP;
 };
 
+/** @typedef {{method: string, args: Array<any> }} InvitationMaker */
+/**
+ * @param {InvitationMaker} invitationMaker
+ * @param {string} instanceHandleBoardId
+ * @param {Board} board
+ * @param {ZoeService} zoe
+ * @returns {Promise<Invitation>}
+ */
 const makeInvitation = async (
   invitationMaker,
   instanceHandleBoardId,
@@ -102,9 +110,9 @@ const makeInvitation = async (
 ) => {
   const instance = E(board).getValue(instanceHandleBoardId);
   const publicFacet = E(zoe).getPublicFacet(instance);
-  const { description, args = [] } = invitationMaker;
+  const { method, args = [] } = invitationMaker;
 
-  return E(publicFacet)[description](...args);
+  return E(publicFacet)[method](...args);
 };
 
 export const findOrMakeInvitation = async (
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -1802,7 +1802,12 @@ export function makeWalletRoot({
   const handleSuggestIssuerAction = ({ petname, boardId }) =>
     suggestIssuer(petname, boardId);
 
-  const performAction = async obj => {
+  /** @typedef {{action: string}} Action */
+  /**
+   * @param {Action} obj
+   * @returns {Promise<any>}
+   */
+  const performAction = obj => {
     const { type, data } = JSON.parse(obj.action);
     switch (type) {
       case 'acceptOffer':
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -1033,6 +1033,7 @@ export function makeWalletRoot({
     const invitationP = findOrMakeInvitation(
       idToOfferResultPromiseKit,
       board,
+      zoe,
       zoeInvitePurse,
       invitationBrand,
       offer,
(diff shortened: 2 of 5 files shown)


packages/wallet/api/src/findOrMakeInvitation.js
```diff
@@ -94,9 +94,23 @@ const makeContinuingInvitation = async (
   return invitationP;
 };
 
+const makeInvitation = async (
```
turadg, line 105: please add types. `invitationMaker` is a little unconventional.

packages/wallet/api/src/findOrMakeInvitation.js
```diff
@@ -94,9 +94,23 @@ const makeContinuingInvitation = async (
...
+  invitationMaker,
+  instanceHandleBoardId,
+  board,
+  zoe,
+) => {
+  const instance = E(board).getValue(instanceHandleBoardId);
+  const publicFacet = E(zoe).getPublicFacet(instance);
+  const { description, args = [] } = invitationMaker;
```
turadg: so `description` specifies a method name? this seems like an abuse of the field. why is this safe?

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1790,6 +1791,29 @@ export function makeWalletRoot({
...
+    const dappOrigin =
+      requestContext.dappOrigin || requestContext.origin || 'unknown';
+    return acceptOffer(`${dappOrigin}#${rawId}`);
+  };
+
+  const handleSuggestIssuerAction = ({ petname, boardId }) =>
+    suggestIssuer(petname, boardId);
+
+  const performAction = async obj => {
+    const { type, data } = JSON.parse(obj.action);
```
dckc, line 1811: Coordinating around what goes in `obj.action` could be tricky, but using a `type` descriminator leaves plenty of room for evolution, so I suppose this is fine.

@michaelfig I wonder if you have any thoughts.

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1193,6 +1193,166 @@ test('lib-wallet addOffer for autoswap swap', async (/** @type {LibWalletTestCon
   );
 });
 
+test('lib-wallet performAction suggestIssuer', async (/** @type {LibWalletTestContext} */ t) => {
```
dckc: :+1:

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1193,6 +1193,166 @@ test('lib-wallet addOffer for autoswap swap', async (/** @type {LibWalletTestCon
...
+
+  assert(done);
+  t.is(
+    wallet.getIssuer('bucksIssuer'),
+    bucksIssuer,
+    `bucksIssuer is stored in wallet`,
+  );
+});
+
+test('lib-wallet performAction acceptOffer', async (/** @type {LibWalletTestContext} */ t) => {
```
dckc: :+1:

dckc (approved):
thanks for thorough unit tests

packages/wallet/api/src/lib-wallet.js
```diff
@@ -1790,6 +1791,29 @@ export function makeWalletRoot({
...
+    switch (type) {
+      case 'acceptOffer':
+        return handleAcceptOfferAction(data);
+      case 'suggestIssuer':
+        return handleSuggestIssuerAction(data);
+      default:
+        throw new Error(`Unknown wallet action ${type}`);
+    }
+  };
```
turadg: This pattern requires naming things twice. consider,
```js
const handlers = {
  async acceptOffer(offer) {
  …
  },
  suggestIssuer({ petname, boardId }) {
    return suggestIssuer(petname, boardId);
  }
};
const performAction = obj => {
  const { type, data } = JSON.parse(obj.action);
  const handle = handlers[type];
  if (!handle) {
    throw new Error(`Unknown wallet action ${type}`);
  }
  return handle(data);
```

Also note `performAction` need not by synchronous. It does need to return a promise. Giving it jsdoc can make that clear.

packages/wallet/api/test/test-lib-wallet.js
turadg: context should be typed on `test` binding so they don't have to be repeated in each call. example: https://github.com/Agoric/agoric-sdk/blob/936f9af005eab2333f6460eaeb79a696c60646ce/packages/inter-protocol/test/test-gov-collateral.js#L35-L37

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1193,6 +1193,166 @@ test('lib-wallet addOffer for autoswap swap', async (/** @type {LibWalletTestCon
...
+  const { issuer: bucksIssuer } = makeIssuerKit('bucks');
+  const bucksIssuerBoardId = await E(board).getId(bucksIssuer);
+
+  const action = JSON.stringify({
+    type: 'suggestIssuer',
+    data: { petname: 'bucksIssuer', boardId: bucksIssuerBoardId },
+  });
+  const done = await wallet.performAction({ action });
+
+  assert(done);
```
turadg: failure message would be more informative with:
```suggestion
  t.truthy(done);
```

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1193,6 +1193,166 @@ test('lib-wallet addOffer for autoswap swap', async (/** @type {LibWalletTestCon
   );
 });
 
+test('lib-wallet performAction suggestIssuer', async (/** @type {LibWalletTestContext} */ t) => {
+  t.plan(1);
```
turadg: planning is helpful when the checks run in control structures. given the checks are all at top level of this function, this seems superfluous and means another thing to change when changing the test.

packages/wallet/api/src/findOrMakeInvitation.js
samsiegart: I'm not sure what you mean by safe, is `description` misleading of a name? Maybe `descriptor` or `method` instead?

The user has to approve this, so I don't see it as being less safe than the dapp passing in an arbitrary invitation like the other way.

turadg: yeah, `method` would address my concern. `description` sounds like a general field for arbitrary text but the string in this field has powers, specifically to specify the method to call.

samsiegart:
> thanks for thorough unit tests

I didn't have much of a choice, it really is the fastest way to test changes 😄

packages/wallet/api/src/findOrMakeInvitation.js
samsiegart: Done

packages/wallet/api/src/findOrMakeInvitation.js
samsiegart: Changed to method

packages/wallet/api/src/lib-wallet.js
samsiegart: Added a jsdoc.

I think for the `handlers` thing, I'd rather flatten it out into separate functions like it is now, I think it's slightly easier to understand which actions are defined this way.

packages/wallet/api/test/test-lib-wallet.js
samsiegart: Done

packages/wallet/api/test/test-lib-wallet.js
samsiegart: Removed

packages/wallet/api/test/test-lib-wallet.js
samsiegart: Done

PR #5947, opened 2022-08-12
fix: better mismatch errors

erights:
For the pattern language to actually be used a lot, it needs to emit better errors diagnosing a pattern match failure.

TODO After this is merged, merge https://github.com/Agoric/documentation/pull/695

### DIFF
--- a/packages/store/src/patterns/patternMatchers.js
+++ b/packages/store/src/patterns/patternMatchers.js
@@ -255,15 +255,28 @@ const makePatternKit = () => {
 
   // /////////////////////// matches ///////////////////////////////////////////
 
-  /** @type {CheckMatches} */
-  const checkMatches = (specimen, patt, check = x => x) => {
+  /**
+   * @param {Passable} specimen
+   * @param {Pattern} pattern
+   * @param {Checker=} check
+   * @param {string|number} [label]
+   * @returns {boolean}
+   */
+  const checkMatches = (specimen, pattern, check = x => x, label = undefined) =>
+    // eslint-disable-next-line no-use-before-define
+    applyLabelingError(checkMatchesInternal, [specimen, pattern, check], label);
+
+  /**
+   * @param {Passable} specimen
+   * @param {Pattern} patt
+   * @param {Checker=} check
+   * @returns {boolean}
+   */
+  const checkMatchesInternal = (specimen, patt, check = x => x) => {
     if (isKey(patt)) {
       // Takes care of all patterns which are keys, so the rest of this
       // logic can assume patterns that are not key.
-      return check(
-        keyEQ(specimen, patt),
-        X`${specimen} - Must be equivalent to: ${patt}`,
-      );
+      return check(keyEQ(specimen, patt), X`${specimen} - Must be: ${patt}`);
     }
     assertPattern(patt);
     const specStyle = passStyleOf(specimen);
@@ -283,7 +296,7 @@ const makePatternKit = () => {
             X`Array ${specimen} - Must be as long as copyArray pattern: ${patt}`,
           );
         }
-        return patt.every((p, i) => checkMatches(specimen[i], p, check));
+        return patt.every((p, i) => checkMatches(specimen[i], p, check, i));
       }
       case 'copyRecord': {
         if (specStyle !== 'copyRecord') {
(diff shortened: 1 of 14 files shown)


packages/zoe/test/unitTests/zcf/test-zoeHelpersWZcf.js
```diff
@@ -299,13 +299,13 @@ test(`zoeHelper with zcf - fit proposal patterns`, async t => {
...
   t.throws(() => fit(proposal, harden([])), {
-    message: /.* - Must be equivalent to: \[\]/,
+    message: /.* - Must be: \[\]/,
   });
   t.throws(
     () => fit(proposal, M.split({ want: { C: M.any() } })),
     {
       message:
-        /Must have same property names as record pattern: {"C":"\[match:any\]"}/,
+        /want: {"A":{"brand":"\[Alleged: moola brand\]","value":"\[20n\]"}} - Must have missing properties \["C"\]/,
```
Chris-Hibbert: can you get rid of the backslashes and/or square brackets in the property name?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -295,12 +308,32 @@ const makePatternKit = () => {
...
+            return check(
+              false,
+              X`${specimen} - Must have missing properties ${q(missing)}`,
+            );
+          }
+          const passNameSet = new Set(pattNames);
+          const unexpected = specNames.filter(name => !passNameSet.has(name));
+          assert(
+            unexpected.length >= 1,
+            X`Internal: must have either missing or extra: ${q(
```
Chris-Hibbert, line 323: This doesn't show up in any tests. Should be easy to provoke, right?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -255,15 +255,28 @@ const makePatternKit = () => {
...
+  /**
+   * @param {Passable} specimen
+   * @param {Pattern} pattern
+   * @param {Checker=} check
+   * @param {string|number} [label]
+   * @returns {boolean}
+   */
+  const checkMatches = (specimen, pattern, check = x => x, label = undefined) =>
+    // eslint-disable-next-line no-use-before-define
+    applyLabelingError(checkMatchesInternal, [specimen, pattern, check], label);
```
Chris-Hibbert, line 267: I can't find any tests that follow the error path, so the label isn't detectably  checked.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -976,12 +1034,8 @@ const makePatternKit = () => {
...
       harden(newBase);
       return (
-        checkMatches(specB, newBase, check) &&
-        (rest === undefined ||
-          check(
-            matches(specR, rest),
-            X`Remainder ${specR} - Must match ${rest}`,
-          ))
+        checkMatches(specB, newBase, check, 'partial-base') &&
+        (rest === undefined || checkMatches(specR, rest, check, 'partial-rest'))
```
Chris-Hibbert: Is there something we could expand *partial-base* and *partial-rest* to that would be more meaningful in context? I'm not precisely sure what the semantics is, so I don't have a suggestion, but these don't tell me much when looking at the relevant errors.

Chris-Hibbert (changes requested):
This is an improvement. I have a few suggestions for further improvements, and some requests for additional tests.

packages/store/src/patterns/patternMatchers.js
erights: I label an error `Internal:` when it means that it should not be able to happen, so if it does, something is seriously screwed up. We should also hook these up to more severe termination (like of the vat), but there's not an easy way to do that here.

packages/store/src/patterns/patternMatchers.js
erights: Agreed. And that was my second try on terminology. Maybe brainstorm sometime soon?

packages/zoe/test/unitTests/zcf/test-zoeHelpersWZcf.js
erights: By changing many of these from regexps to literal strings, I was able to get rid of many backslashes. More readable, thanks!

Done. Resolving.

packages/store/src/patterns/patternMatchers.js
erights: I don't think there's anything to do here, so Resolving. But feel free to unresolve and discuss.

packages/store/src/patterns/patternMatchers.js
erights: For example, in test-inputValidation we see the error message
```
        'displayInfo: partial-base: decimalPlaces: "hello" - Must be >= -100',
```
Every level of `"<label>: "` before the `"hello"` was added on the error path by this mechanism. Resolving.

Chris-Hibbert (approved):
This LGTM. Ping me when you want to chat about terminology.

packages/store/src/patterns/patternMatchers.js
erights: changed to required-parts optional-parts rest-parts. Much better!

PR #5948, opened 2022-08-12
test storage nodes data

turadg:
finishing #5935

## Description

We don't yet have automated tests for the data written to storage nodes. It's increasingly a problem as we move more data flows to RPC.

https://github.com/Agoric/agoric-sdk/pull/5935 was landed with manual testing, on the condition automated tests were added soon. This PR does so, adapting a new test hook from https://github.com/Agoric/agoric-sdk/pull/5942

### Security Considerations

--

--

### Testing Considerations

Exports a testing utility for use in other packages' tests

### DIFF
--- a/packages/inter-protocol/test/supports.js
+++ b/packages/inter-protocol/test/supports.js
@@ -10,14 +10,14 @@ import {
   makePromiseSpace,
 } from '@agoric/vats/src/core/utils.js';
 import { makeBoard } from '@agoric/vats/src/lib-board.js';
-import { makeChainStorageRoot } from '@agoric/vats/src/lib-chainStorage.js';
+import { makeFakeStorageKit } from '@agoric/vats/tools/storage-test-utils.js';
 import { Stable } from '@agoric/vats/src/tokens.js';
 import { makeZoeKit } from '@agoric/zoe';
 import { makeRatio } from '@agoric/zoe/src/contractSupport/ratio.js';
 import { makeFakeVatAdmin } from '@agoric/zoe/tools/fakeVatAdmin.js';
 import buildManualTimer from '@agoric/zoe/tools/manualTimer.js';
-import { makeLoopback } from '@endo/captp';
-import { E } from '@endo/far';
+import { makeLoopback, makeMarshal } from '@endo/captp';
+import { E, Far } from '@endo/far';
 import { makeTracer } from '../src/makeTracer.js';
 
 /**
@@ -64,10 +64,34 @@ export const setUpZoeForTest = (setJig = () => {}) => {
 };
 harden(setUpZoeForTest);
 
-export const mockChainStorageRoot = () => {
-  const toStorage = v => v;
-  return makeChainStorageRoot(toStorage, 'swingset', 'mockChainStorageRoot');
+export const makeMockChainStorageRoot = () => {
+  const { rootNode, data } = makeFakeStorageKit('mockChainStorageRoot');
+
+  const defaultMarshaller = makeMarshal(undefined, (_slotId, iface) => ({
+    iface,
+  }));
+
+  return Far('mockChainStorage', {
+    ...rootNode,
+    /**
+     *  Defaults to deserializing pass-by-presence objects into { iface } representations.
+     * Note that this is **not** a null transformation; capdata `@qclass` and `index` properties
+     * are dropped and `iface` is _added_ for repeat references.
+     *
+     * @param {string} path
+     * @param {Marshaller} marshaller
+     * @returns {unknown}
+     */
+    getBody: (path, marshaller = defaultMarshaller) => {
+      const dataStr = data.get(path);
+      assert(dataStr, `no data at ${path}`);
+      assert.typeof(dataStr, 'string');
+      const datum = JSON.parse(dataStr);
+      return marshaller.unserialize(datum);
+    },
+  });
 };
+/** @typedef {ReturnType<typeof makeMockChainStorageRoot>} MockChainStorageRoot */
 
 /**
  *
@@ -84,7 +108,7 @@ export const setupBootstrap = (t, optTimer) => {
 
   const timer = optTimer || buildManualTimer(t.log);
   produce.chainTimerService.resolve(timer);
-  produce.chainStorage.resolve(mockChainStorageRoot());
+  produce.chainStorage.resolve(makeMockChainStorageRoot());
   produce.board.resolve(makeBoard());
 
   const { zoe, feeMintAccess, run } = t.context;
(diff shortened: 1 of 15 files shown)


packages/vats/src/lib-chainStorage.js
```diff
@@ -113,3 +113,46 @@ export async function makeStorageNode(chainStorage, childName) {
...
 }
 harden(makeStorageNode);
+
+/**
+ * For testing, creates a chainStorage root node over an in-memory map
+ * and exposes both the map and the sequence of received messages.
+ *
+ * @param {string} rootPath
+ */
+export const makeTransientStorageKit = rootPath => {
```
turadg: @gibson042 suggested "transient" may not sound non-production enough. could be `makeFakeStorageKit`. Before renaming I'd like to get clear on the taxonomy of storage kinds with @michaelfig

packages/inter-protocol/test/supports.js
```diff
@@ -64,10 +64,20 @@ export const setUpZoeForTest = (setJig = () => {}) => {
...
+  return Far('mockChainStorage', {
+    ...rootNode,
+    getBody: path => {
+      const dataStr = data.get(path);
+      assert(dataStr, `no data at ${path}`);
+      assert.typeof(dataStr, 'string');
+      const datum = JSON.parse(dataStr);
+      return JSON.parse(datum.body);
+    },
+  });
```
gibson042: This function currently relies on its callers to make assumptions about capdata, but I think it would be better to handle more of that for them. For example:
```suggestion
  const { rootNode, data } = makeTransientStorageKit('mockChainStorageRoot');
  
  // Default to a deserialize pass-by-presence objects into { iface } representations.
  // Note that this is **not** a null transformation; capdata `@qclass` and `index` properties
  // are dropped and `iface` is _added_ for repeat references.
  const defaultMarshaller = makeMarshal(undefined, (_slotId, iface) => ({ iface }));
  
  return Far('mockChainStorage', {
    ...rootNode,
    getBody: (path, marshaller = defaultMarshaller) => {
      const dataStr = data.get(path);
      assert(dataStr, `no data at ${path}`);
      assert.typeof(dataStr, 'string');
      const datum = JSON.parse(dataStr);
      return marshaller.unserialize(datum);
    },
  });
```

```patch
--- packages/inter-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js
+++ packages/inter-protocol/test/amm/vpool-xyk-amm/test-xyk-amm-swap.js
@@ -1242,14 +1242,14 @@
   t.like(mockChainStorage.getBody('mockChainStorageRoot.amm.pool0.metrics'), {
     centralAmount: {
       brand: { iface: 'Alleged: central brand' },
-      value: { digits: '50000' },
+      value: 50000n,
     },
     liquidityTokens: {
       brand: { iface: 'Alleged: MoolaLiquidity brand' },
-      value: { digits: '50000' },
+      value: 50000n,
     },
     secondaryAmount: {
       brand: { iface: 'Alleged: moola brand' },
-      value: { digits: '10000' },
+      value: 10000n,
     },
   });
```

packages/vats/src/lib-chainStorage.js
```diff
@@ -113,3 +113,46 @@ export async function makeStorageNode(chainStorage, childName) {
...
 }
 harden(makeStorageNode);
+
+/**
+ * For testing, creates a chainStorage root node over an in-memory map
+ * and exposes both the map and the sequence of received messages.
+ *
+ * @param {string} rootPath
+ */
+export const makeFakeStorageKit = rootPath => {
```
gibson042: I like the name, but could we move this function into another file? Something with "test" in the path would be ideal, but I'd settle for a dedicated "fake-chainStorage" if something about the former doesn't work.

packages/vats/src/lib-chainStorage.js
```diff
@@ -113,3 +113,46 @@ export async function makeStorageNode(chainStorage, childName) {
   return storageNode;
 }
 harden(makeStorageNode);
+
+/**
+ * For testing, creates a chainStorage root node over an in-memory map
+ * and exposes both the map and the sequence of received messages.
+ *
+ * @param {string} rootPath
+ */
```
gibson042: Would it be useful to document the return type, or is that pointless because it's already statically known?
```suggestion
/**
 * For testing, creates a chainStorage root node over an in-memory map
 * and exposes both the map and the sequence of received messages.
 *
 * @param {string} rootPath
 * @returns {{rootNode: ReturnType<typeof makeChainStorageRoot>, data: Map<string, string>, messages: StorageMessage[]}}
 */
```

packages/vats/src/lib-chainStorage.js
```diff
@@ -113,3 +113,46 @@ export async function makeStorageNode(chainStorage, childName) {
...
+export const makeFakeStorageKit = rootPath => {
+  /** @type {Map<string, unknown>} */
```
gibson042: Why `unknown`?
```suggestion
  /** @type {Map<string, string>} */
```

gibson042 (approved):

packages/vats/src/lib-chainStorage.js
turadg: `test/` shouldn't be exported but I'll try something like `TestUtils`

packages/inter-protocol/test/supports.js
turadg: NIce!

packages/vats/src/lib-chainStorage.js
turadg: yeah, implied by the return. I don't think it provides any useful additional safety

packages/vats/src/lib-chainStorage.js
turadg: from before I new that it was always `string` :) 

updated type and methods to match the Golang

packages/vats/src/lib-chainStorage.js
```diff
@@ -54,7 +52,11 @@ export function makeChainStorageRoot(
   function makeChainStorageNode(path) {
     const node = {
       getStoreKey() {
-        return handleStorageMessage({ key: path, method: 'getStoreKey' });
+        return handleStorageMessage({
+          key: path,
+          method: 'getStoreKey',
+          value: '',
```
gibson042, line 58: We shouldn't need to send `value` for "getStoreKey" messages.
```suggestion
```

gibson042 (approved):

packages/vats/src/lib-chainStorage.js
turadg: without it the payload doesn't satisfy the `StorageMessage` type, which is defined to match Golang. I could make that `Partial<>` but I figure it's better to reveal the JS devs that Golang will always interpret some string

gibson042: OK, that's fair. But note that Go won't do anything with the string in this case other than deserialize it: https://github.com/Agoric/agoric-sdk/blob/7d141a47b311363f099f496d4ed9b4d0f28c8fff/golang/cosmos/x/vstorage/vstorage.go#L76-L86 (note the absence of a `msg.Value` reference).

turadg:
@Mergifyio requeue

PR #5949, opened 2022-08-12
Wallet UI services

samsiegart:

### DIFF
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -127,7 +127,7 @@ export const makeExportContext = () => {
       byVal: makeScalarMap(),
     },
     // TODO: offer, contact, dapp
-    /** @type {IdTable<number, Payment>} */
+    /** @type {IdTable<number, unknown>} */
     unknown: {
       bySlot: makeScalarMap(),
       byVal: makeScalarMap(),
@@ -251,7 +251,11 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
       bySlot: makeScalarMap(),
       byVal: makeScalarMap(),
     },
-    // TODO: offer, contact, dapp
+    /** @type {IdTable<number, unknown>} */
+    unknown: {
+      bySlot: makeScalarMap(),
+      byVal: makeScalarMap(),
+    },
   };
   /** @type {IdTable<BoardId, unknown>} */
   const boardObjects = {
@@ -292,8 +296,7 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
     fromMyWallet: (slot, iface) => {
       const { kind, id } = parseWalletSlot(walletObjects, slot);
       return kind
-        ? // @ts-expect-error tsc isn't quite this clever
-          provideVal(walletObjects[kind], id, iface)
+        ? provideVal(walletObjects[kind], id, iface)
         : slotToVal.fromBoard(slot, iface);
     },
   };
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -241,12 +241,12 @@ const defaultMakePresence = iface => {
  */
 export const makeImportContext = (makePresence = defaultMakePresence) => {
   const walletObjects = {
-    /** @type {IdTable<number, Purse>} */
+    /** @type {IdTable<number, unknown>} */
     purse: {
       bySlot: makeScalarMap(),
       byVal: makeScalarMap(),
     },
-    /** @type {IdTable<number, Payment>} */
+    /** @type {IdTable<number, unknown>} */
     payment: {
       bySlot: makeScalarMap(),
       byVal: makeScalarMap(),
@@ -306,7 +306,6 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
     fromMyWallet: val => {
       const kind = findKey(walletObjects, k => walletObjects[k].byVal.has(val));
       if (kind) {
-        // @ts-expect-error has(val) above ensures val has the right type
         const id = walletObjects[kind].byVal.get(val);
         return makeWalletSlot(walletObjects, kind, id);
       }
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -241,12 +241,12 @@ const defaultMakePresence = iface => {
  */
 export const makeImportContext = (makePresence = defaultMakePresence) => {
   const walletObjects = {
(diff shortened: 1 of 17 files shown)


samsiegart:
@dckc I'd like your eyes specifically on the changes in marshal-context.js. Without those we get `bad board slot unknown:1` when unserializing a contact in the ui.

@turadg For general js/react code review

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -292,8 +296,7 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
     fromMyWallet: (slot, iface) => {
       const { kind, id } = parseWalletSlot(walletObjects, slot);
       return kind
-        ? // @ts-expect-error tsc isn't quite this clever
```
turadg, line 295: …but Sam is!

packages/wallet/ui/src/components/Offer.jsx
```diff
@@ -62,9 +62,14 @@ const OfferWithoutContext = ({
...
-  const approve = () => {
+  const approve = async () => {
     setPendingOffers({ offerId: id, isPending: true });
-    E(offer.actions).accept().catch(console.error);
+    try {
+      await E(offer.actions).accept();
+    } catch (e) {
+      setPendingOffers({ offerId: id, isPending: false });
+      console.error('Failed to accept offer', e);
+    }
```
turadg: with this `try`, returned promise will now resolve even if `accept` fails. I don't know if anything is actually paying attention to the result (there's no return type) but I think it would be safer to keep this function sync and return the failed promise.

```suggestion
    return E(offer.actions).accept().catch(e => {
      setPendingOffers({ offerId: id, isPending: false });
      console.error('Failed to accept offer', e);
    });
```

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -56,12 +96,30 @@ const cmp = (a, b) => {
...
+  if (proposalForDisplay) {
+    give = proposalForDisplay.give ?? {};
+    want = proposalForDisplay.want ?? {};
+    args = proposalForDisplay.arguments;
+    hasDisplayInfo = true;
+  } else if (proposalTemplate) {
+    give = proposalTemplate.give ?? {};
+    want = proposalTemplate.want ?? {};
+    args = proposalTemplate.arguments;
+  }
```
turadg, line 124: is it intended that `! proposalForDisplay && ! proposalTemplate` result in `give = {}` etc? Consider including an explanation.
```suggestion
  } else {
    // leave default values
  }
```

packages/wallet/ui/src/service/Issuers.js
```diff
@@ -0,0 +1,44 @@
+import { makeNotifierKit } from '@agoric/notifier';
+
+export const getIssuerService = signSpendAction => {
```
turadg, line 6: consider typing

packages/wallet/ui/src/service/Issuers.js
```diff
@@ -0,0 +1,44 @@
...
+  const suggestions = new Map();
+  const { notifier, updater } = makeNotifierKit();
```
turadg, line 9: not worth changing right now but for the future consider `makePublishKit`

turadg (approved):
No blockers. I assume you've tested manually

packages/wallet/ui/src/components/Proposal.jsx
samsiegart: Yes, added an explanation

packages/wallet/ui/src/service/Issuers.js
samsiegart: Done

packages/wallet/ui/src/service/Issuers.js
samsiegart: Added TODO

packages/wallet/ui/src/components/Offer.jsx
samsiegart: That's better, done thanks

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -127,7 +127,7 @@ export const makeExportContext = () => {
       byVal: makeScalarMap(),
     },
     // TODO: offer, contact, dapp
-    /** @type {IdTable<number, Payment>} */
+    /** @type {IdTable<number, unknown>} */
```
dckc, line 130: oops! Thanks for fixing that.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -241,17 +241,21 @@ const defaultMakePresence = iface => {
  */
 export const makeImportContext = (makePresence = defaultMakePresence) => {
   const walletObjects = {
-    /** @type {IdTable<number, PurseActions>} */
+    /** @type {IdTable<number, Purse>} */
```
dckc: yes, I was probably getting ahead of myself with `PurseActions`.

packages/wallet/api/test/test-lib-wallet.js
```diff
@@ -1205,7 +1205,7 @@ test('lib-wallet performAction suggestIssuer', async t => {
     type: 'suggestIssuer',
     data: { petname: 'bucksIssuer', boardId: bucksIssuerBoardId },
   });
-  const done = await wallet.performAction({ action });
+  const done = await wallet.performAction({ spendAction: action });
```
dckc, line 1208: suggestIssuer is a `spendAction`? I suppose that's correct but just a sort of de-optimization.

dckc (approved):
looks good.

In particular, test coverage for `marshal-contexts.js` is still 100%

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -241,17 +241,19 @@ const defaultMakePresence = iface => {
  */
 export const makeImportContext = (makePresence = defaultMakePresence) => {
   const walletObjects = {
-    /** @type {IdTable<number, PurseActions>} */
```
dckc, line 244: ouch... why did we have to take that out? Can we make it `IdTable<number, unknown>`? or even `IdTable<number, *>`? Using `*` (aka `any`) should include a justification comment.

samsiegart: Added it back with `unknown`

packages/wallet/api/test/test-lib-wallet.js
samsiegart: I haven't tried getting a non-spend action end-to-end working, I assumed that was still needing design but perhaps this can be the first one we try with. We can revisit.

PR #5952, opened 2022-08-13
fix: bug in M.setOf and M.bagOf

erights:
In the internal pattern match logic of `M.setOf(keyPatt)` and `M.bagOf(keyPatt)`, even when called by `fit`, a mismatch could cause these to return `false` rather than throwing. This masked a bug, also fixed in this PR.

The tests that would test this fix are postponed until https://github.com/Agoric/agoric-sdk/pull/5947 , which is stacked on this one, so those tests can show the improved error messages.

### DIFF
--- a/packages/inter-protocol/src/stakeFactory/stakeFactoryManager.js
+++ b/packages/inter-protocol/src/stakeFactory/stakeFactoryManager.js
@@ -204,11 +204,7 @@ const manager = {
       brands.Attestation,
       X`Invalid Attestation ${attestationGiven}. Expected brand ${brands.Attestation}`,
     );
-    fit(
-      attestationGiven.value,
-      M.bagOf([M.string(), M.bigint()]),
-      'attestationGiven',
-    );
+    fit(attestationGiven.value, M.bagOf(M.string()), 'attestationGiven');
     const [[_addr, valueLiened]] = getCopyBagEntries(attestationGiven.value);
     const amountLiened = AmountMath.make(brands.Stake, valueLiened);
     const maxDebt = floorMultiplyBy(amountLiened, mintingRatio);
--- a/packages/ERTP/src/typeGuards.js
+++ b/packages/ERTP/src/typeGuards.js
@@ -58,7 +58,7 @@ const SetValueShape = M.arrayOf(M.key());
  * will never be minted. That want will never be satisfied.
  * "You can't always get..."
  */
-const CopyBagValueShape = M.bagOf();
+const CopyBagValueShape = M.bag();
 
 const AmountValueShape = M.or(
   NatValueShape,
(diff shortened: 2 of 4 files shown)


erights:
Ready for review

packages/inter-protocol/src/stakeFactory/stakeFactoryManager.js
```diff
@@ -204,11 +204,7 @@ const manager = {
       brands.Attestation,
       X`Invalid Attestation ${attestationGiven}. Expected brand ${brands.Attestation}`,
     );
-    fit(
-      attestationGiven.value,
-      M.bagOf([M.string(), M.bigint()]),
-      'attestationGiven',
-    );
+    fit(attestationGiven.value, M.bagOf(M.string()), 'attestationGiven');
```
turadg, line 207: Isn't it still a bag of tuples of [string, bigint]?

the next line reads the entries as such. 
```js
const [[_addr, valueLiened]] = getCopyBagEntries(attestationGiven.value);
```

I'm surprised no tests are failing at this change. We should be able to get TypeScript to error. I look forward to tackling that after MN-1.

turadg (changes requested):

packages/inter-protocol/src/stakeFactory/stakeFactoryManager.js
erights: It is a set of `[T, bigint]` pairs representing a multiset of `T` elements. Equally valid to describe as a single-valued map from `T` to bigint representing a multiset of `T` elements.

Taking "bag" to be synonymous with "multiset", it is a "bag of T".

However, I can see the "entries" terminology is confusing for producing the pairs. It only makes sense in terms of either view of the underlying representation. Nevertheless, it is the right API to visit the bag in O(unique elements) rather than O(elements).

What change do you think is called for?

erights: Actually, there's an API change that works as well with the bag/multiset view, and works even better with the entries/single-valued-mapping view.

```js
M.bagOf(elementPatt?: Pattern, countPatt?: Pattern)
```
which is more consistent with the existing
```js
M.mapOf(keyPatt?: Pattern, valuePatt?: Pattern)
```

Although it is hard to imagine a scenario where the countPatt is useful, it is less surprising to allow it than to omit it. I will make this change.

erights: Done. PTAL

erights:
I added the relevant `M.bagOf` tests to https://github.com/Agoric/agoric-sdk/pull/5947

packages/inter-protocol/src/stakeFactory/stakeFactoryManager.js
turadg: > However, I can see the "entries" terminology is confusing for producing the pairs

Ah yes, that's what confused me.

> What change do you think is called for?

I'll think on it. Out of scope here, I reckon.

turadg (approved):

PR #5957, opened 2022-08-13
Add `pinBootstrapRoot` swingset configuration option

### LINKED ISSUE #5771: need cosmic-swingset to pinVatRoot('bootstrap'), to enable some post-launch upgrades

#5740 describes a change to the chain's "core bootstrap" code to support future kernel upgrade which introduce new objects (in particular new devices) after the kernel has been initialized. It adds methods to the bootstrap vat's root object, which can be invoked from outside the kernel by calling something like `controller.queueToVatRoot('produceItem', newDeviceNode)`. Once these objects are injected, we can either use the existing `bridgeCoreEval` code to evaluate JS that fetches those objects/devices with `consume`, or we can use a new method (sitting next to the new `produceItem`) that evaluates JS through a different `queueToVatRoot` call.

However for any of this to work, the root object must still be around by the time we do the upgrade. The `bootstrap()` message (enqueued during `initializeSwingset()`) keeps that object around until it gets delivered (run-queue messages hold refcounts on their targets), but after delivery, there may be nothing still referring to the vat-bootstrap root object. If so, the kernel will deliver a `dispatch.dropExport()`, and the vat-bootstrap liveslots will free it. After that point, `queueToVatRoot` will fail, because the target is gone.

To avoid this, and preserve our options for upgrade, we need to change this code:

https://github.com/Agoric/agoric-sdk/blob/1dd584b195212705b1f74a8c89b7f3f121640e41/packages/cosmic-swingset/src/launch-chain.js#L104

to do a `controller.pinVatRoot('bootstrap')` after the `initializeSwingset` call.

I'm not 100% positive that will work.. all the `pinToVatRoot` calls I do in swingset unit tests are done after we've built the controller (`makeSwingsetController`), and before the first call to `await controller.run()`. Those tests only ever launch the kernel once, but the real chain separates initialization from launch (because it will launch every time the kernel process is restarted). So I'll have to double-check that the initialized-but-not-launched state is suitable for `pinVatRoot` to work. I'm pretty sure it will.

## Description of the Design

Add `c.pinVatRoot('bootstrap')`.

## Security Considerations

This causes one extremely high-authority object to stick around longer, however if was previously being collected right away (i.e. if `pinVatRoot` was actually necessary), then nothing in a vat had a reference to it, so the only newly-empowered code is outside the kernel, which is a place with a lot of power anyways.

## Memory Considerations

This might cause objects in other vats to be retained longer than expected, if they were being kept alive by something in the root vat. I think the existing `bridgeCoreEval` callback was keeping the main `produce`/`consume` promise tables alive, and that probably keeps everything else touched during startup alive, so I don't expect this to cause a change to the object graph.

## Test Plan

I really don't know how to unit test this, partially because the only tests we have in cosmic-swingset are coarse integration tests. What I'd love to have is a way to exercise a `controller.queueToVatRoot('produceItem')` somehow in a locally-launched application.

cc @michaelfig 

I think I know how to add this myself, but to write a test I'll need help.

FUDCo:
Also apply this feature to cosmic-swingset in satisfaction of the original issue.

Fixes #5771

### DIFF
--- a/packages/SwingSet/docs/configuration.md
+++ b/packages/SwingSet/docs/configuration.md
@@ -23,9 +23,10 @@ config = {
   defaultManagerType: MANAGERTYPESTRING?
   includeDevDependencies: BOOL?
   defaultReapInterval: NUMBER|'never'?
-  enableFakeDurable: BOOL?
+  relaxDurabilityRules: BOOL?
   snapshotInitial: NUMBER?
   snapshotInterval: NUMBER?
+  pinBootstrapRoot: BOOL?
   vats: {
     VATNAME: {
       sourceSpec: PATHSTRING?
@@ -103,6 +104,17 @@ The `bootstrap` property, if present, names the vat that is to be the
 bootstrap vat.  If this property is missing, there will be no
 bootstrap vat, which may be disallowed in certain use cases.
 
+The `pinBootstrapRoot` property, if `true`, causes the bootstrap vat's root
+object's reference count to be artificially incremented by one as part of
+setting it up, so that the bootstrap root won't be garbage collected even if it
+becomes nominally unreferenced.  The bootstrap vat's root will generally be
+garbage collected after it finishes executing the `bootstrap` method unless that
+method formally exports the root object to some other vat.  However, the host
+retains the ability to deliver messages to it directly (for example, using
+`controller.queueToVatRoot()`), so if the host wants the bootstrap root to
+remain available to provide additional services after bootstrapping, this flag
+should be set.  Defaults to `false`.
+
 The `defaultManagerType` property indicates what sort of vat worker to run vats
 in if they are not specifically configured otherwise.  Currently allowed
 manager types are `'local'`, `'nodeWorker'`, `'node-subprocess'`, and
@@ -122,11 +134,11 @@ the number of deliveries that should be made to the vat before reaping, or the
 string `'never'`, indicating that such activity should never happen.  If
 defaults (yes, the default has a default) to 1.
 
-The `enableFakeDurable` property, if `true`, allows vat code running inside this
-swingset to use the `fakeDurable` option for stores and virtual objects.  It
-defaults to `false`, because this is a dangerous, development-time only feature;
-any vat using it will be left in a corrupted state after delivery of a `stopVat`
-directive.
+The `relaxDurabilityRules` property, if `true`, allows vat code running inside
+this swingset to store non-durable references inside durable objects and stores
+for testing purposes.  It defaults to `false`, because this is a dangerous,
+development-time only feature; any vat using it will be left in a corrupted
+state after delivery of a `stopVat` directive.
 
 The `snapshotInterval` property is the frequency (measured in numbers of
 deliveries) with which `'xs-worker'` vats will perform snapshots to disk of
--- a/packages/SwingSet/test/test-controller.js
+++ b/packages/SwingSet/test/test-controller.js
@@ -110,14 +110,84 @@ test('simple call', async t => {
   await simpleCall(t);
 });
 
-test('bootstrap', async t => {
+async function doTestBootstrap(t, doPin) {
   const config = await loadBasedir(
     new URL('basedir-controller-2', import.meta.url).pathname,
   );
+  if (doPin) {
+    config.pinBootstrapRoot = true;
+  }
   const c = await buildVatController(config);
   t.teardown(c.shutdown);
+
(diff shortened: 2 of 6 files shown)


packages/SwingSet/docs/configuration.md
```diff
@@ -23,9 +23,10 @@ config = {
   defaultManagerType: MANAGERTYPESTRING?
   includeDevDependencies: BOOL?
   defaultReapInterval: NUMBER|'never'?
-  enableFakeDurable: BOOL?
+  relaxDurabilityRules: BOOL?
   snapshotInitial: NUMBER?
   snapshotInterval: NUMBER?
+  pinBootstrap: BOOL?
```
warner: Let's name it `pinBootstrapRoot` to be more precise.

packages/SwingSet/docs/configuration.md
```diff
@@ -103,6 +104,16 @@ The `bootstrap` property, if present, names the vat that is to be the
...
+The `pinBootstrap` property, if `true`, causes the bootstrap vat's root object's
+reference count to be artificially incremented by one as part of setting it up,
+so that the bootstrap vat won't be garbage collected even if its root object
+becomes nominally unreferenced.  The bootstrap vat will generally be garbage
+collected after it finishes executing the `bootstrap` method unless it formally
+exports its root object.  However, the host retains the ability to deliver
+messages to it directly, so if the host wants the bootstrap root to remain
+available to provide additional services after bootstrapping, this flag should
+be set.  Defaults to `false`.
+
```
warner, line 117: We should tighten this up. It increfs the root object, yes. But the goal is to keep the root *object* from being GCed (even if if becomes unreferenced). The *vat* never gets GCed: that would involve something like watching the c-list for exports and subscribed promises, and terminating the entire vat if that list ever became empty (because the vat could no longer receive messages from the outside world, so it could no longer acquire agency, so it might as well not exist) (except that exported promises would get rejected instead of remaining in limbo forever).

The second sentence needs to say something like "the root object will generally be collected unless it is referenced by some other object created during the `bootstrap()` method, or `bootstrap()` deliberately exports the root object".

Also the "to provide additional services" should mention "(e.g. `controller.queueToVatRoot`)".

packages/SwingSet/test/test-controller.js
```diff
@@ -110,14 +110,41 @@ test('simple call', async t => {
...
+  if (doPin) {
+    t.deepEqual(c.dump().log, [
+      'buildRootObject called',
+      'bootstrap called',
+      'more stuff',
+    ]);
+  } else {
+    t.deepEqual(status, 'rejected');
+    t.deepEqual(capdata, {
+      body: '{"@qclass":"error","name":"Error","message":"vat terminated"}',
```
warner, line 179: We should be testing this by looking at the kernelDB for the bootstrap vat's c-list: in the `true` case, `o+0` should be present (and have a refcount of 1), in the `false` case it should be missing. `test/test-gc-kernel.js` has some helpers for this (like `dumpObjects()`). We want to figure out the kref, then use `t.deepEqual(dumpObjects()[kref], [bootstrapVatID, 1, 1])` for the `true` case.

The `vat terminated` message is kinda weird. Was it liveslots exploding when the kernel handed it a vref that wasn't present in `slotToVal`? That's the kernel's fault, and liveslots can't do anything better than explode (if it can't trust the kernel to maintain an invariant, it can't trust it with anything else, and without the kernel it can't do anything). But it's not really what this test wants to be asserting.

Sending a message with `c.queueToVatRoot` is a good additional check. I'd be inclined to only do it in the `true` case, but maybe I'm just trying to keep spurious errors out of the stdout/stderr stream.

warner (changes requested):
Core implementation is good, but config name, docs, and test could use updating.

packages/SwingSet/test/test-controller.js
FUDCo: I think I disagree.  The purpose of the config option is to prevent the root object from being GC'd.  The fact that it does this by means of incrementing the refcount is secondary.  The test as implemented experiences the GC or not according to the option setting, which is exactly what we want to check.  There's already so much crap in the stdout/stderr stream that I think that ship has sailed.

warner (approved):

warner:
awesome, thanks

mhofman:
@mergifyio requeue

mhofman:
@mergifyio unqueue

mhofman:
@mergifyio queue

mhofman:
@Mergifyio refresh

PR #5960, opened 2022-08-14
fix: far classes with interface guards, used by ERTP

erights:
Prior to this PR, virtual and durable far objects could be written using the various "kind" definers
   * virtual vs durable (`define*Kind` vs `define*DurableKind`)
   * single object vs cohort of facets (`define*Kind` vs `define*KindMulti`)
   * whether durables are automatically registered with baggage (`defineDurableKind*` vs `vivifyKind*`)
   * there is also a special case for durable singletons (`vivifySingleton`)

In practice we expect only three to be common (See https://github.com/Agoric/agoric-sdk/discussions/5708 )
   * `vivifyDurableKind`
   * `vivifyDurableKindMulti`
   * `vivifySingleton`

This PR introduces replacements for these, in order to address two problems:
   * Kind definers require their methods to take an additional `context` first argument for receiving the object `state` and access to `self` or cohort of `facets`. This shifts the argument types by one, which creates a painful IDE experience (See https://github.com/Agoric/agoric-sdk/pull/5960#issuecomment-1218297969 below. (Note that singletons do not have this problem.)
   * As far objects, these may be exposed to malicious messages, both intra-vat (eventually) and inter-vat. Thus they must be defensively consistent, which requires strong input validation. Experience shows that it is tricky to write adequate input validation manually and to review for adequacy. (This applies to normal non-virtual far objects as well, which will be addressed by https://github.com/Agoric/agoric-sdk/pull/6107 .)

The replacements
   * virtual vs durable (`defineVirtualFarClass` vs `defineDurableFarClass`)
   * single object vs cohort of facets (`define*FarClass` vs `define*FarClassKit`)
   * are durables auto registered in baggage (`defineDurableFarClass*` vs `vivifyFarClass*`)
   * special case for singletons (`vivifyFarInstance`)

Its worth noting that we expect conversion from Kind style to Far Class style not to effect which are common in practice:
   * `vivifyFarClass`
   * `vivifyFarClassKit`
   * `vivifyFarInstance`

It addresses the two problems mentioned above
   * The methods provided to the FarClass definers take their `context` as their `this` argument, leaving their function type unaffected. This *requires* that these methods be written as functions that receive their `this` binding dynamically. We chose to use concise method syntax, where all these methods are written inline within the call to the FarClass definer. Note that these constraints are familiar to JS `class` programmers, whose methods must also be written in concise method syntax that must appear inline within the `class` definition.
   * This PR extends our pattern sublanguage with `M.interface` and helpers `M.call`, `M.callWhen`, and `M.await` for defining protective interfaceGuards that can express and enforce type-like constraints on incoming arguments and return results. This sublanguage needs to be separately documented. The FarClass definer functions all take an `interfaceGuard` argument. The methods exposed as the API will check all their incoming arguments before calling the raw methods provided in the same call to the FarClass definer. 
 
The raw methods can thus assume that all the validation expressed by the interfaceGuard for this method has already been checked. It is still the responsibility of the raw method to validate all issues that remain after the interfaceGuard has done its part of the job. For example, for a payment argument, the interfaceGuard can only validate that it is a remotable. It remains to the raw method to manually check, for example, that it is a live payment of the correct brand containing enough assets.

This raises the other reason we need to write these methods inline: By themselves, they are fragile. Also, by themselves, they are `this`-sensitive functions. For both of those reasons, we'd like to avoid the possibility of them floating around as first-class values. By having them appear inline in the FarClass definer call, we successfully encapsulate them. Afterward, they can only be reached by the protective outer shell described by the interfaceGuard.

After this PR is merged and we have experience using it, we hope to decide to deprecate the Kind functions and style in favor of the Far Class style. But since this decision should follow experience, we leave it to the follow on PR https://github.com/Agoric/agoric-sdk/pull/6106 .


erights:
Reviewers,

Not really ready for review. Please ignore for now and pay attention onto to https://github.com/Agoric/agoric-sdk/pull/5970
I did this only to tickle CI, which didn't work anyway. 
> 6 skipped and 1 successful checks

If someone understands what I need to do to tickle CI into running all our normal tests, please let me know.

mhofman:
> If someone understands what I need to do to tickle CI into running all our normal tests, please let me know.

Draft PR, or for that matter PRs that are not ready for merge don't run integration tests. Those can be forced with the `force:integration` label that I've just applied to this PR. I doubt however that the integration tests would catch something the unit tests didn't in this case.

mhofman:
Deployment test [failed ](https://github.com/Agoric/agoric-sdk/runs/7874224780?check_suite_focus=true#step:10:14340)because of https://github.com/Agoric/agoric-sdk/pull/5970#issuecomment-1218094370. Not sure why getting-started failed

erights:
Still more typing work to be done, but recording this for now. This PR introduces a "thisful" style as an alternative, where the context is passed as `this` rather than as first argument. This has costs and benefits, as demonstrated by the contrast between the first pair of images (from #5970) vs the second pair, from the current state of this PR. 

Advantage thisful style:
In both, the types are actually correct (and with more work would be more precise). But the hover on the first shows the unshifted types inferred from the behavior methods, while also showing the `KindFacets<...>` around it to transform it into the correct shifted types of the callable methods. To understand the type shown in the first hover, you need to simulate the `KindFacets<...>` transform in your head. This isn't hard. But the verbosity of the unshifted types makes these hovers much less useful. The second hover is both easier to understand and much less verbose, and so more informative.

Advantage context-argument style, arguably:
With the first context-argument style, the `context` is the first normal argument and the behavior methods are normal Jessie code written as arrow function. With the second thisful style, the behavior methods receive their context as their `this`, and so must be written using a syntax for functions that can dynamically receive their `this` binding from their callers. The natural one to use is the concise method syntax. Using classes would force this choice -- class methods can only be expressed using concise method syntax.

image
...
image

vs
image
...
image

mhofman:
One thing I noticed is that `this` doesn't actually refer to `self` for single faceted objects (or the current facet for multi-facet objects), but the `context` first arg we currently have. While that make sense, it might end up somewhat surprising for developers, especially if we do end up with class syntax.

One mitigation might be to cheat and have the `this` arg be facet specific, and have as prototype the "self" facet. Aka, `const thisArg = harden(Object.create(selfFacet, getOwnPropertyDescriptors(context)));`. However we do paint ourselves in a corner with future extensibility of the context object, risking conflict with existing methods.

If we considered only developer experience, really the state and facets would be something like private fields (`this.#state` and `this.#facets`), while leaving identically named self behavior methods accessible without conflicts (e.g. `this.state()`). However I don't think there is a way to accomplish this without something like decorators (so we can initialize the field and prevent re-assignments):
```js
class Counter {
  @state #state;
  @facets #facets;

  inc(n = 1) {
    return this.#state.value += n;
  }
}
```

A cheap alternative might be to use symbols, but that may require making the behavior declaration a "maker":
```js
defineKind('counter', () => ({ value: 0 }), ({ stateSym, facetsSym }) => ({
  inc(n = 1) {
    return this[stateSym].value += n;
  },
}));
```

packages/ERTP/test/unitTests/test-issuerObj.js
```diff
@@ -198,7 +198,10 @@ test('purse.deposit promise', async t => {
   await t.throwsAsync(
     // @ts-expect-error deliberate invalid arguments for testing
     () => E(purse).deposit(exclusivePaymentP, fungible25),
-    { message: /deposit does not accept promises/ },
+    {
+      message:
+        'fungible Purse purse.deposit args[0]: promise "[Promise]" - Must be a remotable',
```
Chris-Hibbert: **fungible Purse purse.deposit args[0]**
This has semantically distinct parts that are all run together in this message. Can we do something to make it easier to read?  

e.g.  **In 'deposit' method of "fungible Purse", the argument (Promise) must be a remotable.**

packages/ERTP/src/types.js
```diff
@@ -115,65 +115,66 @@
...
  *
  * @param {ERef<Payment>} payment
- * @param {Pattern=} optAmountShape
- * @returns {Promise<Payment>}
+ * @returns {Promise<Amount<K>>}
  */
 
 /**
- * @callback IssuerIsLive
+ * @callback IssuerBurn
```
Chris-Hibbert: Can the order of IssuerIsLive, IssuerClaim, and IssuerBurn be swapped back so the diffs will be more legible?

Chris-Hibbert (commented):
I'd like to see tests demonstrating that the interface guards prevent or detect some things or make them visible in the IDE or something.

packages/ERTP/src/types.js
erights: Will rebase on https://github.com/Agoric/agoric-sdk/pull/6012 .

packages/ERTP/test/unitTests/test-issuerObj.js
erights: Good suggestion. I did something close --- specific to the portion of these errors controlled by the code in this PR. PTAL -- do you agree? (optimistically resolving but feel free to unresolve)

Chris-Hibbert: That's very nice.  Thanks!

packages/ERTP/src/types.js
erights: Fixed by #6012 . PTAL. Resolving.

packages/ERTP/src/paymentLedger.js
```diff
@@ -458,56 +379,115 @@ export const vivifyPaymentLedger = (
...
-    getBrand: () => brand,
-    getAllegedName: () => name,
-    getAssetKind: () => assetKind,
-    getDisplayInfo: () => displayInfo,
+  /** @type {Issuer} */
+  const issuer = vivifyFarInstance(issuerBaggage, `${name} issuer`, IssuerI, {
+    getBrand() {
+      return brand;
+    },
+    getAllegedName() {
```
Chris-Hibbert, line 396: Do we need to update the Wiki re [Arrow Function style](https://github.com/Agoric/agoric-sdk/wiki/Arrow-Function-Style)?

erights: Yes, good point. I'll do that and leave this conversation unresolved until then.

erights:
This PR is ready for review. It is staged behind only #5970 , which was approved 5 days ago and is expected to be merged once endo-vs-agoric-sdk CI problems have been cured. I'm going to extend this to impose resource limits in the next PR, not this one.

packages/ERTP/src/paymentLedger.js
```diff
@@ -84,9 +87,37 @@ export const vivifyPaymentLedger = (
   elementSchema,
   optShutdownWithFailure = undefined,
 ) => {
-  const getBrand = () => brand;
+  /** @type {Brand} */
+  const brand = vivifyFarInstance(issuerBaggage, `${name} brand`, BrandI, {
+    isMyIssuer(allegedIssuer) {
+      return allegedIssuer === issuer;
```
Chris-Hibbert, line 94: The old version used `E.when()` here. What's the reason to stop supporting Promises for issuers?  (and later for payments.)

Chris-Hibbert:
> This PR is ready for review. 

Please add some detail to the top comment, which says only **TODO describe PR**.

packages/ERTP/src/paymentLedger.js
erights: I'm not! This is part of what's so cool about interface guards. The guard for this method says
```js
  isMyIssuer: M.callWhen(M.await(IssuerShape)).returns(M.boolean()),
```

The `M.callWhen(` means that the call to the unprotected method can be delayed and arrive out of order (as opposed to `M.call(` which must be delivered to the unprotected method synchronously in order). The `M.await(pattern)` means to delay invoking the unprotected method until the possible promise in that argument position fulfills. Once it does, then check its fulfillment against the argument pattern. Only once all that passes, then invoke the unprotected method with an argument in that position that satisfies the awaited pattern.

In this case, the pattern `IssuerShape` merely enforces that the argument is a remotable. But this means that the unprotected method can now be written assuming a remotable in that position!

erights: (Note: I just fixed a typo that inverted meaning: "not" -> "now")

packages/ERTP/src/typeGuards.js
```diff
@@ -129,3 +129,95 @@ export const DisplayInfoShape = M.partial(
...
+
+  const PaymentI = M.interface('Payment', {
+    getAllegedBrand: M.call().returns(brand),
+  });
+
+  const PurseI = M.interface('Purse', {
+    getAllegedBrand: M.call().returns(brand),
+    getCurrentAmount: M.call().returns(amountSchema),
+    getCurrentAmountNotifier: M.call().returns(NotifierShape),
+    deposit: M.call(PaymentShape).optional(M.pattern()).returns(amountSchema),
```
Chris-Hibbert: This deserves a comment saying that it's intentional that it doesn't use `callWhen`.

packages/ERTP/src/paymentLedger.js
Chris-Hibbert: That seems like significant info about the semantics of the method. What do you think about moving those guards next to the methods?

packages/store/src/keys/checkKey.js
```diff
@@ -161,7 +161,7 @@ export const checkCopySet = (s, check = x => x) => {
       X`Not a copySet: ${s}`,
     ) &&
     checkElements(s.payload, check) &&
-    checkKey(s.payload);
+    checkKey(s.payload, check);
```
erights: See https://github.com/Agoric/agoric-sdk/pull/6076

packages/ERTP/src/paymentLedger.js
erights: While I agree, I don't see a way to do that while still having the method guards be part of their interface guard. Instead I added some commentary at the top of the method bodies where it seemed most relevant. Resolving, because I see nothing better to do. But please unresolve if you have a suggestion.

packages/store/src/keys/checkKey.js
erights: Done

packages/ERTP/src/typeGuards.js
erights: Done

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -623,9 +581,12 @@ export function makeVirtualObjectManager(
     isDurable,
     durableKindDescriptor,
   ) {
-    const { finish } = options;
+    const {
+      finish,
+      thisfulMethods = false,
+      interfaceGuard = undefined,
```
gibson042, line 589: Remember to document these new options in `@param`, which I guess should also be retyped to use `DefineKindOptions`.

packages/store/test/test-patterns.js
```diff
@@ -264,6 +264,71 @@ const matchTests = harden([
...
+    specimen: makeTagged('match:any', undefined),
+    yesPatterns: [M.any(), M.pattern()],
+    noPatterns: [
+      [M.key(), 'A passable tagged "match:any" is not a key: "[match:any]"'],
+    ],
+  },
+  {
+    specimen: makeTagged('match:any', 88),
+    yesPatterns: [M.any(), M.not(M.pattern())],
+    noPatterns: [[M.pattern(), 'Payload must be undefined: 88']],
```
gibson042: Note for later: this message could be better (and better aligned with the others).
```suggestion
    noPatterns: [[M.pattern(), 'match:any payload: 88 - Must be undefined']],
```
And possibly better if they're all restructured to be data-last, e.g.
* match:any payload: Must be undefined: 88
* match:remotable payload: Must be a copyRecord to match copyRecord pattern {"label":"[match:kind]"}: 88
* match:remotable payload: label: Must be a string: 88

packages/ERTP/src/payment.js
```diff
@@ -1,22 +1,29 @@
...
-  const paymentKindHandle = provideKindHandle(issuerBaggage, `${name} payment`);
-  const makePayment = defineDurableKind(paymentKindHandle, () => ({}), {
-    getAllegedBrand: getBrand,
-  });
+export const vivifyPaymentKind = (issuerBaggage, name, brand, PaymentI) => {
+  const makePayment = vivifyFarClass(
+    issuerBaggage,
+    `${name} payment`,
+    PaymentI,
+    () => ({}),
```
gibson042: Nit: it might be useful to export this trivial initialization function so it can have a name, e.g. `initEmpty`.

packages/ERTP/src/paymentLedger.js
```diff
@@ -458,56 +380,127 @@ export const vivifyPaymentLedger = (
...
+        throw err;
+      }
+      return paymentBalance;
+    },
+    claim(srcPayment, optAmountShape = undefined) {
+      // IssuerI delays calling this method until `srcPayment` is a Remotable
+      assertLivePayment(srcPayment);
+      const srcPaymentBalance = paymentLedger.get(srcPayment);
+      assertAmountConsistent(srcPaymentBalance, optAmountShape);
+      // Note COMMIT POINT within moveAssets.
```
gibson042, line 437: Is there a way to formalize this information rather than trusting it to comments?

packages/ERTP/src/typeGuards.js
```diff
@@ -129,3 +129,105 @@ export const DisplayInfoShape = M.partial(
...
+
+/**
+ * @param {Pattern} [brand]
+ * @param {Pattern} [assetKind]
+ * @param {Pattern} [amountSchema]
+ */
+export const makeIssuerInterfaces = (
+  brand = BrandShape,
+  assetKind = AssetValueShape,
+  amountSchema = AmountShape,
```
gibson042: Should these parameters be renamed to shape/schema/pattern/etc. for avoiding confusion between pattern and instance of pattern?
```suggestion
  brandShape = BrandShape,
  assetValueShape = AssetValueShape,
  amountShape = AmountShape,
```

packages/store/src/patterns/interface-tools.js
```diff
@@ -0,0 +1,227 @@
...
+  const { returnGuard } = methodGuard;
+  const { syncMethod } = {
+    // Note purposeful use of `this` and concise method syntax
+    syncMethod(...args) {
+      defendSyncArgs(harden(args), methodGuard, label);
+      const result = apply(method, this, args);
+      fit(harden(result), returnGuard, `${label}: result`);
+      return result;
+    },
+  };
```
gibson042, line 57: Should we be preserving names in `defendSyncMethod` and `defendAsyncMethod`?
```suggestion
  const { syncMethod } = {
    // Note purposeful use of `this` and concise method syntax
    syncMethod(...args) {
      defendSyncArgs(harden(args), methodGuard, label);
      const result = apply(method, this, args);
      fit(harden(result), returnGuard, `${label}: result`);
      return result;
    },
  };
  // Preserve the original method name.
  defineProperties(syncMethod, {
    name: { value: method.name, configurable: true },
  });
```

packages/store/src/patterns/interface-tools.js
```diff
@@ -0,0 +1,227 @@
...
+};
+
+const defendSyncMethod = (method, methodGuard, label) => {
+  const { returnGuard } = methodGuard;
+  const { syncMethod } = {
+    // Note purposeful use of `this` and concise method syntax
+    syncMethod(...args) {
+      defendSyncArgs(harden(args), methodGuard, label);
+      const result = apply(method, this, args);
+      fit(harden(result), returnGuard, `${label}: result`);
```
gibson042, line 54: `harden(args)` and `harden(result)` have the potential to observably alter arguments and `result` vs. the undefended method; do we consider that acceptable? Presumably yes, but it's still worth confirming.

packages/store/src/patterns/interface-tools.js
```diff
@@ -0,0 +1,227 @@
...
+      awaitIndexes.push(i);
+    }
+  }
+  return {
+    awaitIndexes,
+    rawMethodGuard: {
+      argGuards: rawArgGuards.slice(0, argGuards.length),
+      optionalArgGuards: rawArgGuards.slice(argGuards.length),
+      restArgGuard,
+    },
```
gibson042, line 86: This should either support or actively reject a `restArgGuard` that is or contains an await guard.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -307,29 +290,23 @@ const makePatternKit = () => {
         }
         const [specNames, specValues] = recordParts(specimen);
```
gibson042: Nit: "spec" as the prefix of a variable name is more suggestive to me of "specification" than "specimen".
```suggestion
        const [specimenNames, specimenValues] = recordParts(specimen);
```

packages/store/src/patterns/patternMatchers.js
```diff
@@ -307,29 +290,23 @@ const makePatternKit = () => {
...
         const [pattNames, pattValues] = recordParts(patt);
-        if (!keyEQ(specNames, pattNames)) {
-          const specNameSet = new Set(specNames);
-          const missing = pattNames.filter(name => !specNameSet.has(name));
+        {
```
gibson042: I don't think this inner block is particularly helpful; `missing` and `unexpected` are both logically at the same level as `specNames` and `pattNames` IMO.
```suggestion
```

packages/store/src/patterns/patternMatchers.js
```diff
@@ -341,37 +318,26 @@ const makePatternKit = () => {
...
-          return check(false, msg);
-        }
-        const specTag = getTag(specimen);
-        if (pattTag !== specTag) {
-          return check(false, msg);
+        if (specStyle !== 'tagged' || getTag(specimen) !== pattTag) {
+          return check(
+            false,
+            X`${specimen} - Only a ${q(pattTag)} matches a ${q(
+              pattTag,
```
gibson042: The duplication of `q(pattTag)` in the error message seems like a bug, albeit a preexisting one probably related to an earlier refactoring. Suggestion:
```js
        specTag = specStyle === 'tagged' ? getTag(specimen) : undefined;
        if (specStyle !== 'tagged' || specTag !== pattTag) {
          return check(
            false,
            X`${specimen} - Must be tagged as a ${pattTag}: ${specTag}`,
          );
        }
```

packages/vat-data/src/far-class-utils.js
```diff
@@ -0,0 +1,217 @@
...
+/** @template T @typedef {import('./types.js').KindFacets<T>} KindFacets */
+/** @typedef {import('./types.js').DurableKindHandle} DurableKindHandle */
+
+/**
+ * @template A,S,T
+ * @param {string} tag
+ * @param {any} interfaceGuard
+ * @param {(...args: A[]) => S} init
+ * @param {T} methods
+ * @param {DefineKindOptions<unknown>} [options]
```
gibson042, line 27: Should the annotations for `define{Virtual,Durable}FarClass…` options include a constraint that `thisfulMethods` is absent or true and `interfaceGuard` is absent?

packages/vat-data/src/far-class-utils.js
```diff
@@ -0,0 +1,217 @@
...
+  options,
+) =>
+  // @ts-expect-error The use of `thisfulMethods` to change
+  // the appropriate static type is the whole point of this method.
+  defineKindMulti(tag, init, facets, {
+    ...options,
+    thisfulMethods: true,
+    interfaceGuard: interfaceGuardKit,
+  });
+harden(defineVirtualFarClass);
```
gibson042: This duplicate harden looks like a typo, it is probably intended for the preceding export that defines a wrapper over `defineKindMulti`.

packages/vat-data/src/far-class-utils.js
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * @template A,S,T
+ * @param {string} tag
+ * @param {any} interfaceGuardKit
+ * @param {(...args: A[]) => S} init
+ * @param {T} facets
+ * @param {DefineKindOptions<unknown>} [options]
+ * @returns {(...args: A[]) => (T & RemotableBrand<{}, T>)}
+ */
+export const defineVirtualFarClassKit = (
```
gibson042, line 55: The helper that wraps `defineKind` is named `defineVirtualFarClass`, but this helper that wraps `defineKindMulti` is named `defineVirtualFarClassKit` rather than something like `defineVirtualFarClassMulti` (and likewise for durable variants). I think that merits explanation or reconsideration.

gibson042 (commented):
A lot of the code looks much cleaner now.

packages/store/src/patterns/interface-tools.js
erights: Was already done at the bottom of `bindMethod`.

packages/vat-data/src/far-class-utils.js
erights: "annotations"?

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Documented these options in `DefineKindOptions`. Resolving.

packages/store/test/test-patterns.js
erights: Took the immediate suggestion, but not the more ambitious restructuring. Resolving.

packages/ERTP/src/typeGuards.js
erights: Good point. Done.

packages/store/src/patterns/interface-tools.js
erights: Actively reject. Done.

packages/store/src/patterns/patternMatchers.js

packages/store/src/patterns/patternMatchers.js

packages/store/src/patterns/patternMatchers.js

packages/vat-data/src/far-class-utils.js
erights: Good catch. Done

packages/ERTP/src/paymentLedger.js
erights: Probably. I tried once, and it was too big a hammer. We should try again. But I don't think it is relevant to this PR, so resolving. If I'm missing something, please unresolve.

The only reason it shows up in this PR is because diffs don't account well for code motion. See line 309 in the left side of the comparison.

packages/ERTP/src/payment.js
erights: Done

packages/store/src/patterns/interface-tools.js
erights: I need to explain this elsewhere, but confirmed. Resolving this one since I already have one open for explaining.

packages/vat-data/src/far-class-utils.js
erights: Sorry. Reading this again it's clear what you meant. I documented these as options only in `DefineKindOptions` where I made clear that they're for internal use only, and when they have which setting. Resolving.

packages/vat-data/src/far-class-utils.js
erights: Yeah, I think I made a mistake when I first suggested the "Multi" suffix. We use it to mean the same thing as our "Kit" suffix, and we do not need two. If things work out, we'll deprecate the define/vivify*Kind/Singleton functions and replace all our uses with define/vivify*FarClass/FarInstance functions. I'm using that anticipated shift to also drop "Multi" in favor of "Kit".

Resolving.

erights:
@Chris-Hibbert @gibson042 
I need to do some writing in order to address the remaining review comments. But I think I've done everything else, so if you're ok starting to PTAL while I gear up to write, PTAL. Thanks!

packages/vat-data/src/far-class-utils.js
erights: The deprecations I have in mind: https://github.com/Agoric/agoric-sdk/pull/6106/files

Chris-Hibbert (approved):
I'm fairly comfortable with the consequences in the ERTP code. The implementation is beyond me to verify.

I doubt I can make use of the tools in any novel cases without help. There are quite a few examples in the ERTP code to copy from, but I'll bet governance (for example) does things that ERTP doesn't, and I don't see documentation on the semantics or usage of the `M.whatever()` functions.

I can intuit the meaning of
```
    M.callWhen(M.await(PaymentShape))
      .optional(M.pattern())
      .returns(amountShape),
```
from context, but I'm not confident that that will be enough for me to use the tools without a fair number of questions.

erights:
Reviewers, do you agree that the following can wait until after this PR?

TODO: Extract some of the text above to serve as documentation of Far Classes without reference to kinds. This can serve as an explanation to programmers who started with Far classes and never need to learn about kinds.

TODO: Provide FarClass definers of normal in-memory heap objects https://github.com/Agoric/agoric-sdk/pull/6107

TODO: Deprecate kinds https://github.com/Agoric/agoric-sdk/pull/6106

TODO: Describe the pattern sublanguage, including the new interface guards.

TODO: Explore flatter `this` structures, so we can avoid the annoyance of always saying `this.state.foo` to access our own instance variable. See https://github.com/Agoric/agoric-sdk/pull/5960#issuecomment-1218316157

TODO: conventional class programmers are used to doing composition by inheritance. For the objects-as-closure style currently supported by `Far`, we've been doing composition with `...`. We have neither for Far classes. Attn @Chris-Hibbert @dtribble Governance currently makes heavy use of `...` composition, so it is urgent to figure out what to do instead. (Though we have an emergency stopgap: `bindAllMethods` that we can use in the short term.)

erights:
> > This PR is ready for review.
> 
> Please add some detail to the top comment, which says only **TODO describe PR**.

Done.

Chris-Hibbert:
> Reviewers, do you agree that the following can wait until after this PR?

I agree. Some of the doc is a hurdle for me, but that's about follow-on work, not this PR.

Chris-Hibbert (commented):
Changing to `comment`  to prevent accidental submission based only on this review.

The parts I understand still LGTM.

packages/ERTP/src/paymentLedger.js
```diff
@@ -458,56 +380,127 @@ export const vivifyPaymentLedger = (
...
+      const [payment] = moveAssets(
+        harden([srcPayment]),
+        harden([srcPaymentBalance]),
+      );
+      return payment;
+    },
+    combine(fromPaymentsPArray, optTotalAmount = undefined) {
+      // IssuerI does *not* delay calling `combine`, but rather leaves it
+      // to `combine` to delay further processing until all the elements of
+      // `fromPaymentsPArray` have fulfilled.
```
gibson042, line 447: I assume this stems from inability to express the need for awaiting all elements of an array (as opposed to awaiting a single value). It will be interesting to consider the use of decorators for a similar purpose if they get added to the language.

packages/store/src/types.js
```diff
@@ -500,7 +500,10 @@
  * @property {() => Matcher} set A CopySet
  * @property {() => Matcher} bag A CopyBag
  * @property {() => Matcher} map A CopyMap
- * @property {() => Matcher} remotable A far object or its remote presence
+ * @property {(label?: string) => Matcher} remotable
+ * A far object or its remote presence. The optional `label` is purely for
+ * diagnostic purpose. It does not enforce any constraint beyond the
+ * must-be-a-remotable constraint.
```
Chris-Hibbert, line 506: I think this is about the way jsdoc works, and not just WebStorm, but what I see is that anything left-justified is attributed to all the `@propoerties` above it. Text that is indented only applies to the one entry immediately above. 

```suggestion
 * @property {(label?: string) => Matcher} remotable - A far object or its remote presence.
 *   The optional `label` is purely for diagnostic purpose. It does not enforce any constraint
 *    beyond the must-be-a-remotable constraint.
```

@turadg am I right?

erights: Here's what I see hovering over this "remotable" in vscode.
image

What do you see in WebStorm?

packages/ERTP/src/typeGuards.js
```diff
@@ -114,18 +114,120 @@ harden(isCopyBagValue);
...
+    getAllegedName: M.call().returns(M.string()),
+    getAssetKind: M.call().returns(assetKindShape),
+    getDisplayInfo: M.call().returns(DisplayInfoShape),
+    makeEmptyPurse: M.call().returns(PurseShape),
+
+    isLive: M.callWhen(M.await(PaymentShape)).returns(M.boolean()),
+    getAmountOf: M.callWhen(M.await(PaymentShape)).returns(amountShape),
+    burn: M.callWhen(M.await(PaymentShape))
+      .optional(M.pattern())
+      .returns(amountShape),
```
gibson042, line 171: If I understand correctly, the pattern for expressing the shape of a function is `M.{call,callWhen}(requiredArg1Shape, …).optional(optionalArg1Shape, …).returns(returnShape)`, where `callWhen` looks for shapes output by `M.await` as arguments that should be awaited before invoking the corresponding function. I accept it in this pull request, but there's probably room for more iteration. Some possibilities:
1. Merge `call` and `callWhen`, relying solely on `M.await` for what they currently differentiate.
2. Rename `call` to something like `callable`, better matching the general use of noun-named `M` methods. Or maybe not, depending upon number 4 below.
3. Make `M.await` infectious, such that `combine` can be updated to e.g. `M.callable(M.arrayOf(M.await(PaymentShape))).…` and lose its `Promise.all`. But this one might be tricky because `M.await` indicates call instructions rather than a shape assertion and therefore is really only appropriate for call/callWhen rather than in an arbitrary pattern—another alternative would be expanding that set such that `combine` can be updated to e.g. `M.callable(M.awaitAll(M.arrayOf(M.eref(PaymentShape)))).…`.
4. It is valuable for `M.await` to have a name indicating its distinction from the other methods that _are_ shape assertions. Noun vs. verb is at least a good start, but there are also stronger possibilities (e.g., substructure like `M.transform.{call,await,etc.}`).
5. The chainable/fluent interface appears distinct to `call`/`callWhen` (e.g., AFAIK we don't support something like `M.key().and(elementShape)`)—we should probably either expand it to the rest of `M` or drop it from these methods in favor of e.g. `M.call({ required: […], optional: […], returns: shape})`).

gibson042 (approved):
@erights LGTM! :shipit:

gibson042:
Also, I agree with the proposed TODOs.

packages/ERTP/src/paymentLedger.js
erights: > I assume this stems from inability to express the need for awaiting all elements of an array (as opposed to awaiting a single value).

Correct.

> It will be interesting to consider the use of decorators for a similar purpose if they get added to the language.

Do you think people (maybe us?) will build similar callWhen/await delaying abstractions using decorators?

Resolving.

packages/ERTP/src/typeGuards.js
erights: > If I understand correctly, the pattern for expressing the shape of a function is `M.{call,callWhen}(requiredArg1Shape, …).optional(optionalArg1Shape, …).returns(returnShape)`, where `callWhen` looks for shapes output by `M.await` as arguments that should be awaited before invoking the corresponding function.

Yes.

> I accept it in this pull request, but there's probably room for more iteration. Some possibilities:
> 
> 1. Merge `call` and `callWhen`, relying solely on `M.await` for what they currently differentiate.

`combine` would do well by this rule, since it just happens not to have anything to await, but if it did, it could become a `callWhen`/`await` without losing correctness.

But I rejected that because `call` implies a guarantee of in-order delivery. Purse `deposit` relies on that guarantee, which is why it does not tolerate a promise argument. For cases like `combine`, if you wanted to express the absence of an in-order delivery guarantee, you could just use a `callWhen` that happened not to have an `await`. However, if we expected that to be common, we should optimize that case do it doesn't pay for an extra turn it doesn't need. Currently it would.

> 2. Rename `call` to something like `callable`, better matching the general use of noun-named `M` methods. Or maybe not, depending upon number 4 below.

Indeed, that's what I had in mind -- the distinction between guards and patterns. OTOH, `M.interface` reads like a noun, which breaks my story.

> 3. Make `M.await` infectious, such that `combine` can be updated to e.g. `M.callable(M.arrayOf(M.await(PaymentShape))).…` and lose its `Promise.all`. But this one might be tricky because `M.await` indicates call instructions rather than a shape assertion and therefore is really only appropriate for call/callWhen rather than in an arbitrary pattern—another alternative would be expanding that set such that `combine` can be updated to e.g. `M.callable(M.awaitAll(M.arrayOf(M.eref(PaymentShape)))).…`.

I like the `M.awaitAll` idea. Altogether, I doubt the case is common enough to provide support for. The `combine` code shows that it is convenient enough to do it manually if it only arises rarely. If we do eventually decide that we want better support, I we should do `M.awaitAll`.

> 4. It is valuable for `M.await` to have a name indicating its distinction from the other methods that _are_ shape assertions. Noun vs. verb is at least a good start, but there are also stronger possibilities (e.g., substructure like `M.transform.{call,await,etc.}`).
>
> 5. The chainable/fluent interface appears distinct to `call`/`callWhen` (e.g., AFAIK we don't support something like `M.key().and(elementShape)`)—we should probably either expand it to the rest of `M` or drop it from these methods in favor of e.g. `M.call({ required: […], optional: […], returns: shape})`).

For both 4 and 5, I'm trying for a sweet spot on brevity and readability as well. Many more functions have required arguments than have optional or rest arguments.

IIUC, all your suggestions above are tentative, such that you're ok with me merging without having changed the code itself in light of these suggestions, given my responses. But please keep these ideas in mind for possible improvements after this PR. Resolving.

packages/ERTP/src/paymentLedger.js
erights: Done.

packages/store/src/types.js
erights: Hi @Chris-Hibbert I'm still interested in investigating this anomaly. But for this PR, I'm resolving this conversation.

Chris-Hibbert: storeTypeHints

erights: Yuck!

Aside from the process question of how to do this in bulk, I don't see any reason not to do this. It adds value to the webstorm experience and does not subtract any value anywhere else. File an issue please, citing this conversation? Thanks.

dtribble:
> TODO: conventional class programmers are used to doing composition by inheritance. For the objects-as-closure style currently supported by `Far`, we've been doing composition with `...`. We have neither for Far classes. Attn @Chris-Hibbert @dtribble Governance currently makes heavy use of `...` composition, so it is urgent to figure out what to do instead. (Though we have an emergency stopgap: `bindAllMethods` that we can use in the short term.)

Governance succeeds by using the "shared method pattern" to include governance operations when the class is defined rather than trying to stick them onto the instance at the end. As far as I'm concerned that's better anyway, and is a better match for future inheritance.

packages/store/src/types.js
Chris-Hibbert: done

PR #5963, opened 2022-08-15
feat(vats): publish agoricNames to chainStorage

### LINKED ISSUE #5959: make agoricNames available via chainStorage / RPC

The AMM UI needs to know the AMM instance's board id and should no longer have to use a signed, paid-for transaction to get it
https://github.com/Agoric/dapp-amm/pull/27/files/c72887833df707be00c424eef0d91706af0d0ff3#r945300148

## Description of the Design

Put `agoricNames` in chainStorage using the board marshaler; in particular: the `.entries()` of each `NameHub`.

## Security Considerations

none; `agoricNames` is already published.

## Test Plan

_I think @turadg is putting together testing tools for this sort of thing. IOU a pointer_

dckc:
closes: #5959
stacked on: #5948

I intend to merge as separate commits.

## Description

for `brand`, `issuer`, `instance`, and all the other top-level kinds in `agoricNames`, publish their entries:

```
$ ~/go/bin/agd query vstorage data published.agoricNames.instance -o json | jq '.value | fromjson | {body: .body | fromjson, slots}'
{
  "body": [
    [ "economicCommittee",  { "@qclass": "slot", "iface": "Alleged: InstanceHandle", "index": 0 }],
    [ "amm",  {  "@qclass": "slot",  "iface": "Alleged: InstanceHandle",  "index": 1  }  ],
...
  ],
  "slots": [
    "board00613",
    "board04542",
...
  ]
}
```

### Security Considerations

None? Publishing via the board shouldn't be any more risky than publishing via agoricNames.

I suppose the way `published.agoricNames` works should be documented somewhere?

### Testing Considerations

I added unit tests for the new `NameAdmin.onUpdate` and manually tested the chainStorage integration.

@samsiegart I hope you can do some integration testing with the AMM front end.


packages/vats/src/core/chain-behaviors.js
```diff
@@ -315,6 +317,40 @@ export const makeChainStorage = async ({
   chainStorageP.resolve(rootNodeP);
 };
 
+/** @param {BootstrapPowers} powers */
+export const publishAgoricNames = async ({
+  consume: { agoricNamesAdmin, board, chainStorage: rootP },
+}) => {
+  const root = await rootP;
+  if (!root) {
+    console.warn('cannot publish agoricNames without chainStorage');
```
turadg, line 325: why not throw?

packages/vats/src/core/chain-behaviors.js
```diff
@@ -315,6 +317,40 @@ export const makeChainStorage = async ({
   chainStorageP.resolve(rootNodeP);
 };
 
+/** @param {BootstrapPowers} powers */
+export const publishAgoricNames = async ({
```
turadg, line 320: what calls this?

packages/inter-protocol/src/proposals/addAssetToVault.js
```diff
@@ -125,12 +125,13 @@ export const publishInterchainAssetFromBank = async (
...
     },
   };
-  const { creatorFacet: mint, publicFacet: issuer } = E.get(
+  const { creatorFacet: mint, publicFacet: issuerP } = E.get(
     E(zoe).startInstance(mintHolder, {}, terms),
   );
 
-  const [brand, runBrand] = await Promise.all([
-    E(issuer).getBrand(),
+  const [issuer, brand, runBrand] = await Promise.all([
```
turadg, line 133: why does `issuer` need to be awaited? `addAsset` which takes `kit: AssetIssuerKit` which has `issuer: ERef<Issuer>`.

if it really needs to be a presence, please revise the type

packages/vats/src/core/chain-behaviors.js
```diff
@@ -315,6 +317,40 @@ export const makeChainStorage = async ({
...
+  if (!root) {
+    console.warn('cannot publish agoricNames without chainStorage');
+    return;
+  }
+  const nameStorage = E(root).getChildNode('agoricNames');
+  const marshaller = E(board).getPublishingMarshaller();
+
+  // brand, issuer, ...
+  await Promise.all(
+    keys(agoricNamesReserved).map(kind =>
```
turadg: this indentation is pretty extreme. consider defining some of these callback functions passed into `then`

packages/vats/src/nameHub.js
```diff
@@ -100,6 +117,10 @@ export const makeNameHubKit = () => {
       );
       nameAdmin.update(key, newValue, adminValue);
     },
+    onUpdate: fn => {
+      assert(!fn || !updateCallback);
```
turadg: took me a second to interpret these two cases. please include an assertion message. consider a type annotation

packages/vats/test/test-name-hub.js
```diff
@@ -143,3 +143,27 @@ test('makeNameHubKit - default and set', async t => {
     message: 'key "not set" is not already initialized',
   });
 });
+
+test('makeNameHubKit - listen for updates', t => {
```
turadg, line 147: did you want to add a test like https://github.com/Agoric/agoric-sdk/pull/5948 ?

packages/vats/src/core/chain-behaviors.js
dckc: sim chain?

samsiegart:
Functionally works with https://github.com/Agoric/dapp-amm/pull/31 👍 

Deferring code review to turadg/mfig

packages/vats/src/core/chain-behaviors.js
dckc: in `boot.js` we iterate over `manifest`, look up the functions by name, and call them (using `runModuleBehaviors`)

packages/inter-protocol/src/proposals/addAssetToVault.js
dckc: The motivation for awaiting it is the call to `E(E(agoricNamesAdmin).lookupAdmin('issuer')).update(keyword, issuer)`. What goes in `agoricNames` should be a presence rather than a promise so that it can be compared for identity.

packages/vats/test/test-name-hub.js
dckc: yes, but that's not landed yet. hm.

packages/vats/src/core/chain-behaviors.js
dckc: yeah; this is a reflection of muddled thinking about whether the inner promises were going to resolve in time.

packages/inter-protocol/src/proposals/addAssetToVault.js
dckc: ... and so that the board marshaller can serialize it

packages/vats/src/core/chain-behaviors.js
turadg: I think even sim chain gets a storage node https://github.com/Agoric/agoric-sdk/blob/1c2d4d7c8cbef70f0f13708b43b08ea4a691b15e/packages/vats/src/lib-chainStorage.js#L109-L116

It should

packages/vats/src/nameHub.js
dckc: restructured a bit and added assertion message.

the type is already annotated, though somewhat indirectly via `NameAdmin` typedef in `vats/src/types.js`.

packages/vats/test/test-name-hub.js
turadg: Just got approved so it's going in soon. But if you want to follow up after merging this that works for me.

packages/vats/src/core/chain-behaviors.js
dckc: any idea why the type includes `null`?

turadg: to handle null ;-) for example, https://github.com/Agoric/agoric-sdk/blob/73eda52b1a446286d619cbfec7afe040fedc910d/packages/vats/test/test-clientBundle.js#L83

the promise has to be resolved with something for the economy to start. if real chain storage isn't available (as in sim-chain) then it makes a "NullStorageNode" so that code that depends on storageNode can function

turadg (approved):

packages/vats/src/core/chain-behaviors.js
dckc: throwing here causes at least 1 test to fail. Let's punt for now.

boot › evaluateInstallation is available to core eval (9.5s)
https://github.com/Agoric/agoric-sdk/runs/7863366960?check_suite_focus=true#step:10:2160

dckc:
@Mergifyio unqueue

dckc:
@Mergifyio requeue

mhofman:
@mergifyio enqueue

PR #5969, opened 2022-08-16
Change kernel queue storage manipulations from O(n) to O(1)

### LINKED ISSUE #5932: swingset kernel queue needs O(1) storage format
We currently store the kernel's run-queue as a single JSON-serialized array in `kvStore.get('runQueue')`. That makes inserting or draining N records O(N^2), and we don't have any backpressure or other mechanisms that would keep the N from getting too large to ignore.

I think it's time for us to improve this data structure. I've kinda been putting it off because the two-queues or "each vat gets an input queue and an output queue" work (#3465 and maybe others) would either overlap or obsolete it. But in today's meeting we agreed on a cosmic-swingset -side prioritization approach for MN-1 that won't require kernel changes. So the kernel situation isn't going to be improved by that work, and we need to improve it in other ways.

Fortunately the changes are pretty simple:

* each queue entry goes into a separate key, `runQueue-${entryNum}`
* two additional keys remember the number of the next available entry and the oldest populated entry
* shifting the oldest item from the queue requires three reads, a delete, and a write
* adding an item to the queue requires a read and two writes
* the code that keeps track of the size of the run queue (for `kernelStats`) should be double-checked to make sure it still works, but probably doesn't need many changes

I think the biggest part of the task is updating the handful of tests that assert things about the run-queue storage. I'd recommend writing a helper function that scans the kvStore and returns the same array that used to be in the single `runQueue` key.

FUDCo:
Currently applied to `runQueue` and `acceptanceQueue`.  Could be applied to others if circumstances warrant.

Fixes #5932


packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -258,6 +260,45 @@ export default function makeKernelKeeper(
     kvStore.set('crankNumber', `${crankNumber + 1n}`);
   }
 
+  function initQueue(queue) {
+    kvStore.set(`${queue}`, JSON.stringify([1, 1]));
```
warner, line 264: part of me wants for these to be BigInts.. `2^53` cranks at (an aspirational) 1000/sec gives us about 300kyears, which is obviously plenty but also not as close to infinity as I'd like

we'd need something other than `JSON.stringify`, like `s.split(',').map(BigInt)` and a `${head},${tail}`

warner (approved):
Looks good. I'm on the fence about the BigInt thing.. I know you're probably -0 on such a change, so I'll leave it up to you.

packages/SwingSet/src/kernel/state/kernelKeeper.js
FUDCo: I think 300,000 years is long enough lead time for software changes.  It is also possible that requirements may change during that interval.

PR #5970, opened 2022-08-16
fix(SwingSet)!: Representatives inherit bound methods

erights:
We can take the first big step towards far classes https://github.com/Agoric/agoric-sdk/pull/5960 without any ***intentional*** change to the existing virtual/durable API. We simply change the internal representation of representatives to be a class-like representation. All the bound methods are only created once per kind rather than per instance. These are gathered together into a prototype object. A representative is then just an empty frozen object that inherits from its prototype.

All by itself, this should have a big performance win. Before this PR, every time a representative is created or revived, we create new bound functions specific to that representative, for all methods defined for its kind. With the PR, creating the representative should be cheap and constant time.

The unfortunate consequence of this change is that autobinding is broken. For these representatives, 
```js
const m = obj.methodName;
...
m(); // no longer works
```
would bind `m` to that method as a standalone function that is nevertheless specific to that instance. That will no longer work. Invoking `m` otherwise will, at least, reliably throw an error with a decent diagnostic.

Most of the changed files in this PR are simply the damage repair from loosing the autobinding feature for virtual and durable objects. Please have a look. Whether this PR goes forward or not should largely depend on how painful it is to give up autobinding.

(Note: apply the label `force:integration` if you want CI to run these too, as it will before merging.)


FUDCo:
I was not previously aware of this "autobinding" concept.  My first take is that it's not a feature, it's a bug*, so giving it up is not a loss.

(* because it deviates from the existing expectations of regular JS programmers)

mhofman:
Agree with @FUDCo, the fact that methods or representatives were bound was an artifact of the implementation, and wouldn't be expected to work in the model of object as class instances. That said, in the object as closure model, developers may be more used to plucking local object methods out and calling them as free functions. Wondering if that's something we could prevent for Far objects somehow.

Edit: Btw this is probably another argument in favor of avoiding using `E(foo).bar` as a free function in the name of consistency, even though its implementation does also implicitly create a bound method. Wondering if we could similarly have a check in the `E` proxy handlers that the `thisArg` at the time of the call corresponds to the expected one. Opened https://github.com/endojs/endo/issues/1253

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -478,16 +478,38 @@ export function makeVirtualObjectManager(
...
+      // Violating all Jessie rules to create representative that inherit
+      // methods from a shared prototype. The bound method therefore needs
+      // to mention `this`. We define it using concise menthod syntax
+      // so that it will be `this` sensitive but not constructable.
+      //
+      // We normally consider `this` unsafe because of the hazard of a
+      // method of one abstraction being applied to an instance of
+      // another abstraction. To prevent that attack, the bound method
+      // checks that it is in the map in which such its representatives
+      // are registered.
```
Chris-Hibbert: Garbled. I think this is an improvement.
```suggestion
      // checks that it is in the map in which its representatives
      // are registered.
```

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -728,10 +771,12 @@ export function makeVirtualObjectManager(
...
-        context.self = bindMethods(context, behaviorTemplate);
-        toHold = Far(tag, context.self);
-        linkToCohort.set(Object.getPrototypeOf(toHold), toHold);
-        unweakable.add(Object.getPrototypeOf(toHold));
+        const self = harden({ __proto__: prototypeTemplate });
+        context.self = self;
+        contextMapTemplate.set(self, context);
+        toHold = self;
+        // linkToCohort.set(Object.getPrototypeOf(toHold), toHold);
+        // unweakable.add(Object.getPrototypeOf(toHold));
```
Chris-Hibbert: If this code is to remain as a comment, it would be good to have an explanation as to why. Is it the old way? the approach to aspire to?

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -740,10 +785,13 @@ export function makeVirtualObjectManager(
...
-          const facet = Far(`${tag} ${name}`, facets[name]);
-          linkToCohort.set(Object.getPrototypeOf(facet), facet);
-          unweakable.add(Object.getPrototypeOf(facet));
+          const facet = harden({
+            __proto__: prototypeTemplate[name],
+          });
+          contextMapTemplate[name].set(facet, context);
+          facets[name] = facet;
+          // linkToCohort.set(Object.getPrototypeOf(facet), facet);
+          // unweakable.add(Object.getPrototypeOf(facet));
```
Chris-Hibbert: similarly

packages/store/src/utils.js
```diff
@@ -0,0 +1,43 @@
+// @ts-check
+
+const { getPrototypeOf, create, fromEntries } = Object;
+const { ownKeys, apply } = Reflect;
+
+const compareStringifies = (left, right) => {
```
Chris-Hibbert: `compareStringified` reflects what this is doing better. AIUI, it's comparing the stringified versions of its arguments.

packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
```diff
@@ -370,14 +370,17 @@ const potBehavior = {
     const { zcf } = state;
     const { helper } = facets;
     assert(state.open);
-    return zcf.makeInvitation(helper.adjustBalancesHook, 'AdjustBalances');
+    return zcf.makeInvitation(
+      seat => helper.adjustBalancesHook(seat),
+      'AdjustBalances',
+    );
```
Chris-Hibbert: I don't get an intuition from this about what doesn't work any more.  The method being invoked here is on `helper`, right?  In the old form, `helper.adjustBalancesHook` is just as visible, and it's just as clear what object the method is being invoked on. `seat` is being passed implicitly, but if any methods are called on it in `adjustBalancesHook`, it's still apparent that the object is available for its methods to be looked up. 

What does the change in autobinding behavior require?

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Done, thanks.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Leftover from intermediate state. Gone.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Gone.

packages/store/src/utils.js
erights: Good, thanks. Done.

packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
erights: In the following explanation, I ignore the multi-facet case, as that complexity is a distraction from the immediate question.

On representative `helper`, the expression `helper.adjustBalancesHook(seat)` calls the `adjustBalancesHook` function that the virtual object system synthesized, that wraps the `adjustBalancesHook` behavior method that the user wrote. The synthesized method is only called with the non-context arguments explicitly, so it needs to recover the per-representative `context` somehow, to pass as the additional first argument to the behavior method.

The expression `helper.adjustBalancesHook` by itself, without the open paren, simply extracts the synthesized function that would have been called as a first class function.

Before this PR, the virtual object system synthesizes as `adjustBalancesHook` function per representative, each time a representative is create or recreated. This per-representative synthesized function itself captures the relevant `context`, which is how it can provide it to the behavior method. Thus, the synthesized first-class `adjustBalancesHook` function that `helper.adjustBalancesHook` evaluates to already has all the per-representative state that it needs, and can be called later as a function.

With this PR, the virtual object system only synthesizes an `adjustBalancesHook` function on a per-kind basis and stores it on a shared per-kind prototype object that representatives inherit from. Because the new synthesized method is not per-representative, it cannot capture the `context` it needs. It instead captures a `contextMap` that maps from a representative of that kind to the needed `context`. But to use that to look up the context, it needs to know what representative it is invoked on. 

The expression `foo.adjustBalancesHook(seat)` passes the value of `foo`, whatever that is, as the `this` binding to the function looked up by `foo.adjustBalancesHook`. The synthesized function, hoping that its `this` binding is a legitimate representative, looks it up in its `contextMap` to recover the needed `context`. If its `this` binding is anything other than a legitimate representative of its brand, the lookup will fail, preventing application to an illegitimate representative. In

```js
const adjustBalancesHook = helper.adjustBalancesHook;
adjustBalancesHook(seat);
```

the synthesized function is being called with function call syntax (without the `.`) rather than method call syntax. In this case, the `this` binding is `undefined`. The problem is that the synthesized method that `helper.adjustBalancesHook` is only per kind, and has lost all the representative specific context that the method call syntax would have provided --- by binding the left "operand" of the `.` to the the `this` of the called method.

There is a way to repair this, but I would guess it is even more expensive that what we're doing now. By creating the `prototype.adjustBalancesHook` property as an accessor property rather than a data property, on the expression `helper.adjustBalancesHook` where `helper` inherits from the `adjustBalancesHook` accessor property from this `prototype`. On the expression `helper.adjustBalancesHook` the getter would see the value of `helper` as the `this` of that call to the getter. We could have getter of this accessor dynamically produce a bound method wrapping the synthesized method, where the bound method captures the `this` when the property itself was looked up. But the getter cannot distinguish whether the value it returns will be immediately called, or whether it will be held onto and called later, so it would need to create this representative-specific bound method ***on each call***, which may be even more expensive that creating synthesized methods per representative.

But we don't know. It may be cheaper because the getter would only created the bound functions that are used, whereas before this PR, we're creating synthesized methods per representative whether they are used or not. If we think we really want to preserve the current autobinding behavior, we could measure this an an alternate technique and see how expensive it really is.

erights: You just provided me with a great example of a use of autobinding that gets broken by this PR. At https://github.com/Agoric/agoric-sdk/commit/f1210a980eff36d73f2d18dfada85c6c84339041 you pass `feeMint.getFeeIssuerKit` as an argument to `makeZoeStorageManager`. After I rebased, this caused at least 7 CI breakages like https://github.com/Agoric/agoric-sdk/runs/7864694216?check_suite_focus=true , which I try to correct at https://github.com/Agoric/agoric-sdk/pull/5970/commits/3a3a6a5c5e7631c2dc8c925e8f24a0f3a47d791a , replacing

```js
feeMint.getFeeIssuerKit
```
with
```js
feeMintAccess => feeMint.getFeeIssuerKit(feeMintAccess),
```

We'll see shortly if my change makes CI happier ;)

Chris-Hibbert: Thanks. This is precisely the explanation I was looking for.

erights: Thanks. Resolving.

mhofman:
Looks like there is a missing case: https://github.com/Agoric/agoric-sdk/commit/ee3d76af7be74f1e86d87b2843adb6359fb78f65

```diff
diff --git a/packages/inter-protocol/src/vaultFactory/vaultManager.js b/packages/inter-protocol/src/vaultFactory/vaultManager.js
index 5c6d104b3..0fcfe6c2f 100644
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -752,7 +752,7 @@ const collateralBehavior = {
   makeVaultInvitation: ({ state, facets: { self } }) => {
     const { zcf } = provideEphemera(state.collateralBrand);
     assert(zcf);
-    return zcf.makeInvitation(self.makeVaultKit, 'MakeVault');
+    return zcf.makeInvitation(seat => self.makeVaultKit(seat), 'MakeVault');
   },
   /** @param {MethodContext} context */
   getSubscriber: ({ state }) => {
```

Once that one is fixed, the loadgen runs successfully locally. Might be worth figuring out how to get coverage for this path in unit test. @turadg

turadg:
> Might be worth figuring out how to get coverage for this path in unit test

What value is being passed into `makeInvitation` that needs to be tested for? When I change that argument to another function, many tests fail:
```
  return zcf.makeInvitation(() => null, 'MakeVault');
```

mhofman:
@turadg what I meant is that this missing rewrite only surfaced in integration tests, which means `makeVaultInvitation` is not covered by unit tests, or at least not in a way that calls the callback given as first argument to `zcf.makeInvitation`.

turadg:
> which means makeVaultInvitation is not covered by unit tests, or at least not in a way that calls the callback given as first argument to zcf.makeInvitation.

That was what I interpreted. But I tested it in master by changing the first argument to `makeInvitation` (which I put in the comment above) and then the Ava tests failed. Dozens IIRC. So I'm asking what would be a failing argument. Or maybe this is more productive: what evidence is there that this Ava (PR) tests were not sensitive to the argument being wrong?

mhofman:
I don't know but all unit tests passed on the [commit](https://github.com/Agoric/agoric-sdk/commit/82c721006889388538b1b05c155161268b2eb3f9) before my patch above was added. Before my patch for this PR, the callback would have been the equivalent of `async () => { throw new Error(); }` or `() => Promise.reject(new Error())`. Maybe a `null` return value break things in different ways?

turadg:
> () => Promise.reject(new Error())

Mysterious. When I change it to this and run ` yarn test test/vaultFactory/test-*` I get same 31 errors as last time:
```js
return zcf.makeInvitation(() => Promise.reject(new Error()), 'MakeVault');
```

Going back to the question of whether to develop unit test coverage of this case, I don't think it's worthwhile. The code path is well tested and whatever happened here is rare enough that our existing coverage is sufficient.

erights:
I have extracted all the repairs needed to work with this PR into prior PR #5989 that should work both before and after this PR. Both are now more reviewable. This PR, which modifies the mechanism within the virtual object system, must be reviewed by @FUDCo . I've moved the other reviewers to #5989 .

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -478,16 +478,37 @@ export function makeVirtualObjectManager(
...
+    for (const prop of Reflect.ownKeys(behaviorMethods)) {
+      const func = behaviorMethods[prop];
       assert.typeof(func, 'function');
-      const method = (...args) => Reflect.apply(func, null, [context, ...args]);
-      unweakable.add(method);
-      obj[prop] = method;
+
+      // Violating all Jessie rules to create representative that inherit
+      // methods from a shared prototype. The bound method therefore needs
+      // to mention `this`. We define it using concise menthod syntax
```
FUDCo: Typo: menthod -> method

packages/SwingSet/src/liveslots/virtualObjectManager.js
```diff
@@ -478,16 +478,37 @@ export function makeVirtualObjectManager(
...
+      // so that it will be `this` sensitive but not constructable.
+      //
+      // We normally consider `this` unsafe because of the hazard of a
+      // method of one abstraction being applied to an instance of
+      // another abstraction. To prevent that attack, the bound method
+      // checks that it's `this` is in the map in which its representatives
```
FUDCo: Typo: it's -> its

FUDCo:
What's the story with all the test failures?  Based on my read of the changes it doesn't seem like those should be implicated.

erights:
> What's the story with all the test failures? Based on my read of the changes it doesn't seem like those should be implicated.

@warner and I just worked through them. If CI is happy now, then we worked through them correctly. The problem was that the new timer code was counting on autobinding in one place, and counting on methods being enumerable by `...` in two places. All should now be fixed. We'll see shortly.

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Done

packages/SwingSet/src/liveslots/virtualObjectManager.js
erights: Done

FUDCo (approved):
Lookin' good.

mhofman:
We need to fix the getting-started npm test, likely by publishing Endo with the Marshall changes and update the versions

erights:
> We need to fix the getting-started npm test, likely by publishing Endo with the Marshall changes and update the versions

@gibson042 made some great review comments on https://github.com/endojs/endo/pull/1251 . That PR is blocked on my fixing it. I'll do that next.

Issue #5696, opened 2022-06-30
make vat-network upgradable


warner:
## What is the Problem Being Solved?

As part of #5666 we need to make the ["network vat"](https://github.com/Agoric/agoric-sdk/blob/1f2c46550f36530392eee449565c77858bbd4f18/packages/vats/src/vat-network.js) upgradable. This vat hosts a copy of 
[SwingSet/src/vats/network/router.js](https://github.com/Agoric/agoric-sdk/blob/1f2c46550f36530392eee449565c77858bbd4f18/packages/SwingSet/src/vats/network/router.js).

I'm not sure what this will entail (I'm not very familiar with this network code).

## Description of the Design

This includes not only making the network vat durable, but also providing a utility for clients of the network vat (such as Pegasus or ICA) to tolerate an upgrade (and its broken promises) without losing precious state.

## Test Plan

```[tasklist]
### Tasks
- [ ] https://github.com/Agoric/agoric-sdk/issues/8906
- [ ] https://github.com/Agoric/agoric-sdk/pull/8721
- [ ] https://github.com/Agoric/agoric-sdk/issues/8950
- [ ] https://github.com/Agoric/agoric-sdk/pull/8948
- [ ] https://github.com/Agoric/agoric-sdk/issues/8867
```

warner:
@gibson042 could you take a look at this and see how feasible it seems?

warner:
@FUDCo walked through the `packages/SwingSet/src/vats/network/` code today.. we're thinking that maybe it should be moved out of swingset, either into `packages/vats/` that uses it, or into its own package.

His preliminary assessment is that making it upgradable wouldn't be too hard.

Tartuffo:
@FUDCo making you the primary on this.

FUDCo:
I spoke with @michaelfig and he said that moving this to `packages/vats` was reasonable, since it isn't really managed like the other vats that are part of SwingSet but is more of an IBC thing.  Apparently the primary reason this got put into `packages/SwingSet` was that `packages/vats` didn't exist at the time.

warner:
break this into two pieces:
* reachably durable (#5973)
* known (tested) to be upgradable

We can defer the upgrade testing for after the `pso` milestone.

dckc:
vat-network wasn't in the [pismo vats](https://github.com/Agoric/agoric-sdk/issues/5014#issuecomment-1309365804). I think it's only needed for dyanimc IBC, which isn't a vaults release feature

[...]

PR #5973, opened 2022-08-16
Upgradable network vat

FUDCo:
This is a first cut at making the network vat upgradable.

It also includes a refactoring that moves the network vat code from being in a mix of the `SwingSet` package and the `vats` package to being 100% in the `vats` package.

Currently it passes all the tests I could find that in any way implicate the network vat (in the `vats`, `solo`, `cosmic-swingset`, and `pegasus` packages).  However, while it now keeps all of its state that I could identify in durable storage, there is as yet no test as to whether it actually achieves upgradability.  As such, this PR should be viewed as a step on the road to achieving upgradability without yet making any assertions as to whether that destination has in fact been reached.  Consequently I would leave the motivating issue (#5696) open for the time being.

I have made an unsuccessful first cut at attempting to test upgradability, but various complications with the current network vat tests (most notably that these tests don't actually run in a swingset) have meant that while the proposed testing strategy that @michaelfig pointed me at seems to be sound (basically, a network vat restart in the middle of the `test network - ibc` test), the existing tests themselves do not lend themselves to being adapted in this way.  This is work for the future but I believe we have concluded that it is not on the current critical path.


michaelfig:
Sorry, it looks like I didn't approve in time to avoid bitrot.  Chip, would you be able to pick this up again if you could update it to current master?  I'll gladly review this time.

warner:
Incidentally, it sounds like we don't have plans to include dynamic IBC in the Vaults release, therefore we won't be including a vat-network in the bulldozer upgrade. The bundle might be included anyway, by virtue of being in decentral-core-config.json, except that we probably won't use that config file (we'd use something like decentral-main-psm instead). cc @dckc

turadg:
Subsumed by https://github.com/Agoric/agoric-sdk/pull/8721

PR #5974, opened 2022-08-16
5912 casting transaction

turadg:
- wip
- wip2
- test not throwing

closes: #XXXX
refs: #XXXX


patches/protobufjs+6.11.3.patch
```diff
@@ -1,50 +0,0 @@
...
-     }
- 
--    (CustomError.prototype = Object.create(Error.prototype)).constructor = CustomError;
--
--    Object.defineProperty(CustomError.prototype, "name", { get: function() { return name; } });
--
--    CustomError.prototype.toString = function toString() {
--        return this.name + ": " + this.message;
--    };
-+    CustomError.prototype = Object.create(Error.prototype, {
```
erights, line 16: Was this protobuf patch deleted because we're now depending on a protobuf in which this bug has been fixed?

turadg: @michaelfig ^

michaelfig: With this PR, we're now depending on our agoric-labs fork of protobufjs, which has the bugfix.

PR #5980, opened 2022-08-17
fix: show more stacks

erights:
I was not seeing stacks that I expected to be useful. Tracking down the particular error I was seeing, by grepping for `during vat dispatch() of`, I saw that the error was simply being stringified and not otherwise reported, which looses both the stack info and all the redacted error.message info. I found two others by grepping for `${err`. This PR fixes these three to log the error to the console. Depending on what console that is, this may or may not be very useful, but it is the right thing.

No doubt there are other cases where we're only stringifying an error and not otherwise reporting it. But this PR fixes only the three I found.

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -710,6 +710,7 @@ export default function buildKernel(
       // eslint-disable-next-line @jessie.js/no-nested-await
       await vatWarehouse.createDynamicVat(vatID);
     } catch (err) {
+      console.log('error during creation', err);
       const info = makeError(`${err}`);
       const results = {
         didDelivery: true, // ok, it failed, but we did spend the time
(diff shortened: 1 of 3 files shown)


packages/SwingSet/src/kernel/kernel.js
```diff
@@ -710,6 +710,7 @@ export default function buildKernel(
       // eslint-disable-next-line @jessie.js/no-nested-await
       await vatWarehouse.createDynamicVat(vatID);
     } catch (err) {
+      console.log('error during creation', err);
```
warner: make this one `error during createDynamicVat`, just to add some more context

warner (approved):
with one tweak, looks good

packages/SwingSet/src/kernel/kernel.js
erights: Done

erights:
Victory! At https://github.com/Agoric/agoric-sdk/runs/7885344077?check_suite_focus=true#step:10:277 I see the stack I was missing. This stack trace is informative and useful. I also see the redacted message:

```js
Symbol.for(passStyle) property expected: Object [Alleged: criticalVatKey] {}
```
vs
```
"[Symbol(passStyle)]" property expected: (an object)
```

PR #5982, opened 2022-08-17
5979 fix priceFeed resolution

### LINKED ISSUE #5979: published.priceFeed.ATOM_USD_price_feed has very little resolution
## Describe the bug

We set a price was set using `E(history[0]).pushResult('19.37')` but the ratio we got via RPC is 19 USD to 1 ATOM. Is it using a unit price of 10^-6 ATOM rather than 1 ATOM?

originally discussed in https://github.com/Agoric/agoric-sdk/issues/5375#issuecomment-1217458386

## To Reproduce

<details>
<summary>published.priceFeed.ATOM_USD_price_feed</summary>

```console
$ ~/go/bin/agd --node=https://xnet.rpc.agoric.net:443 query vstorage data published.priceFeed.ATOM_USD_price_feed -o json | jq '.value | fromjson | {body: .body | fromjson, slots}'
```
```json
{
  "body": {
    "quoteAmount": {
      "brand": {
        "@qclass": "slot",
        "iface": "Alleged: quote brand",
        "index": 0
      },
      "value": [
        {
          "amountIn": {
            "brand": {
              "@qclass": "slot",
              "iface": "Alleged: ATOM brand",
              "index": 1
            },
            "value": {
              "@qclass": "bigint",
              "digits": "1"
            }
          },
          "amountOut": {
            "brand": {
              "@qclass": "slot",
              "iface": "Alleged: USD brand",
              "index": 2
            },
            "value": {
              "@qclass": "bigint",
              "digits": "19"
            }
          },
          "timer": {
            "@qclass": "slot",
            "iface": "Alleged: timerService",
            "index": 3
          },
          "timestamp": {
            "@qclass": "bigint",
            "digits": "1660695302"
          }
        }
      ]
    },
    "quotePayment": {
      "@qclass": "slot",
      "iface": "Alleged: quote payment",
      "index": 4
    }
  },
  "slots": [
    null,
    "board02810",
    "board0639",
    null,
    null
  ]
}
```
</details>

## Expected behavior

The reported ratio should be 19370000 uist to 1000000 uatom.

## Platform Environment
 - xnet 20220816214108 6b7b997

turadg:
closes: #5979

## Description

The price feed is relayed to RPC by an observer of a QuoteNotifier. The `makeQuoteNotifier` call used `1n` as the unit amount, ignoring decimal places.

This makes it use decimal places to calculate an "integer" unit amount. It also adds unit tests of the storage output and refactors an inter-protocol test util so it can be used here.

I'm not sure about the `unitAmount()` helper function because many functions take a `unitAmountIn` parameter. I wonder if that's how this should be handled instead, perhaps making that a required argument.

### Testing Considerations

New tests but none that would have broken before this fix. I tried setting the decimalPlaces in the `$ATOM` issuer (as it must be on chain) but that broke many other test expressions.

### DIFF
--- a/packages/zoe/src/contractSupport/priceQuote.js
+++ b/packages/zoe/src/contractSupport/priceQuote.js
@@ -1,5 +1,9 @@
 // @ts-check
 
+import { AmountMath } from '@agoric/ertp';
+import { Nat } from '@agoric/nat';
+import { E } from '@endo/eventual-send';
+
 // PriceAuthorities return quotes as a pair of an amount and a payment, both
 // with the same value. The underlying amount wraps amountIn, amountOut, timer
 // and timestamp. The payment is issued by the quoteIssuer to support veracity
@@ -23,3 +27,12 @@ export const getAmountIn = quote => getPriceDescription(quote).amountIn;
 export const getAmountOut = quote => getPriceDescription(quote).amountOut;
 /** @type {(quote: PriceQuote) => Timestamp} */
 export const getTimestamp = quote => getPriceDescription(quote).timestamp;
+
+/** @param {Brand<'nat'>} brand */
+export const unitAmount = async brand => {
+  // Brand methods are remote
+  const displayInfo = await E(brand).getDisplayInfo();
+  const decimalPlaces = displayInfo.decimalPlaces ?? 0;
+  return AmountMath.make(brand, 10n ** Nat(decimalPlaces));
+};
+harden(unitAmount);
--- a/packages/zoe/src/contracts/priceAggregator.js
+++ b/packages/zoe/src/contracts/priceAggregator.js
@@ -36,6 +36,7 @@ import '../../tools/types.js';
  * POLL_INTERVAL: bigint,
  * brandIn: Brand,
  * brandOut: Brand,
+ * unitAmountIn: Amount<'nat'>,
  * }>} zcf
  * @param {{
  * marshaller: Marshaller,
@@ -44,7 +45,9 @@ import '../../tools/types.js';
  * }} privateArgs
  */
 const start = async (zcf, privateArgs) => {
-  const { timer, POLL_INTERVAL, brandIn, brandOut } = zcf.getTerms();
+  const { timer, POLL_INTERVAL, brandIn, brandOut, unitAmountIn } =
+    zcf.getTerms();
+  assert(unitAmountIn, 'missing unitAmountIn');
   assert(privateArgs, 'Missing privateArgs in priceAggregator start');
   const { marshaller, storageNode } = privateArgs;
   assert(marshaller, 'missing marshaller');
@@ -198,18 +201,15 @@ const start = async (zcf, privateArgs) => {
     });
 
   // for each new quote from the priceAuthority, publish it to off-chain storage
-  observeNotifier(
-    priceAuthority.makeQuoteNotifier(AmountMath.make(brandIn, 1n), brandOut),
-    {
-      updateState: quote => publisher.publish(quote),
-      fail: reason => {
-        throw Error(`priceAuthority observer failed: ${reason}`);
-      },
-      finish: done => {
-        throw Error(`priceAuthority observer died: ${done}`);
-      },
+  observeNotifier(priceAuthority.makeQuoteNotifier(unitAmountIn, brandOut), {
+    updateState: quote => publisher.publish(quote),
+    fail: reason => {
+      throw Error(`priceAuthority observer failed: ${reason}`);
     },
-  );
+    finish: done => {
+      throw Error(`priceAuthority observer died: ${done}`);
+    },
+  });
 
   /**
    * @param {Ratio} r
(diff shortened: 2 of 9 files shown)


packages/zoe/src/contractSupport/priceQuote.js
```diff
@@ -23,3 +25,10 @@ export const getAmountIn = quote => getPriceDescription(quote).amountIn;
 export const getAmountOut = quote => getPriceDescription(quote).amountOut;
 /** @type {(quote: PriceQuote) => Timestamp} */
 export const getTimestamp = quote => getPriceDescription(quote).timestamp;
+
+/** @param {Brand<'nat'>} brand */
+export const unitAmount = brand => {
+  const decimalPlaces = brand.getDisplayInfo().decimalPlaces ?? 0;
```
dckc: brands are typically remote. I doubt this will work.

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -1020,6 +1020,40 @@ test('storage keys', async t => {
...
+  t.deepEqual(
+    t.context.mockStorageRoot.getBody(
+      'mockChainStorageRoot.priceAggregator.ATOM_USD_price_feed',
+    ),
+    {
+      quoteAmount: {
+        brand: { iface: 'Alleged: quote brand' },
+        value: [
+          {
+            amountIn: { brand: { iface: 'Alleged: $ATOM brand' }, value: 1n },
```
dckc, line 1042: `1n` doesn't seem right; shouldn't that be `1_000_000n`?

dckc (commented):
brands are typically remote. calls to `.getDisplayInfo()` need `E(brand)`.

packages/zoe/src/contractSupport/priceQuote.js
turadg: The param is declared `Brand`, not `ERef<Brand>`. Do you mean this function won't work as written or won't be as useful as you're inferring I meant it to be?

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
turadg: In dev chains, yes. In this test the brand has no decimal places:
https://github.com/Agoric/agoric-sdk/blob/c81ad73e616c998239610e62751b2ac036b14d26/packages/zoe/test/unitTests/contracts/test-priceAggregator.js#L107

I mentioned this in test plan,
>  I tried setting the decimalPlaces in the $ATOM issuer (as it must be on chain) but that broke many other test expressions.

What coverage are you looking for?

turadg: Took some refactoring 092d8f4209292df0ac02cdda17b18f40d0fdc55c

packages/zoe/src/contractSupport/priceQuote.js
turadg: AmountMath methods take `Brand` and not `ERef` brand, so I thought it wasn't remotable, but the first line of most of them is `assertRemotable`. I take it Brand is always remote, and can't be made present? Or are the types wrong?

6d729a7c0fa9140e1b71514ce4142487c76cccfa

dckc: Unless you created the brand, you don't know what vat it was created in, so you have to assume it's remote and can only be called with `E()`. I can't think of a good way to capture this in TypeScript, though I gather there attempts rattling around somewhere.

AmountMath use brands for their identity; they don't call methods on the brand. They rely on their caller having resolved any promises to presences. For that purpose, `Promise<Brand>` is no good, hence `ERef<Brand>` is probably not right.

dckc:
I'm struggling to confirm that this works with a running `local-chain`. I'm using ccb66847b7fd

I don't see `priceFeed` in `published`:

```
$ ~/go/bin/agd query vstorage keys published
children:
- agoricNames
- amm
- psm
- reserve
- stakeFactory
- vaultFactory
```

I started the chain with `make scenario2-run-chain-economy`. Is there some other step needed to get `priceFeed` listed there? I wonder if I need to run `packages/inter-protocol/scripts/manual-price-feed.js`.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -44,7 +45,9 @@ import '../../tools/types.js';
  * }} privateArgs
  */
 const start = async (zcf, privateArgs) => {
-  const { timer, POLL_INTERVAL, brandIn, brandOut } = zcf.getTerms();
+  const { timer, POLL_INTERVAL, brandIn, brandOut, unitAmountIn } =
```
dckc, line 48: hm... contract creator has to provide `unitAmountIn`. A little inconvenient, perhaps, but certainly on the safe side of "don't use displayInfo for anything other than display."

dckc (approved):

turadg:
> I'm struggling to confirm that this works with a running `local-chain`

I'm not sure. I'm confident enough this fixes the price being quoted ( #5979 ) to land this. If it needs more work for chain setup let's chat.

PR #5983, opened 2022-08-17
test: missing `await electorateOutcome`, so didn't detect paramChange

Chris-Hibbert:
closes: #5962

## Description

There was a missing `await electorateOutcome`, which allowed me to replace 
``` js
        // The electorate didn't change
        firstElectorateInstance,
```
with `secondElectorateInstance`, which was the intended result of the test

### Security Considerations

Where else have I allowed broken tests to slip through like this?

None.

### Testing Considerations

Test didn't find the result it was looking for,  and accepted the incorrect result rather than repairing the test.


dckc (approved):
yay for more eyes looking at our stuff!

PR #5984, opened 2022-08-17
fix(wallet/ui): style connection component better

samsiegart:
refs: https://github.com/Agoric/dapp-treasury/issues/137

image

### DIFF
--- a/packages/wallet/ui/src/AgoricLogo.jsx
+++ b/packages/wallet/ui/src/AgoricLogo.jsx
@@ -1,0 +1,27 @@
+import React from 'react';
+
+const AgoricLogo = () => {
+  const pathStyle = { fill: '#bb2d40' };
+
+  return (
+    <svg
+      version="1.1"
+      id="Layer_1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      x="0px"
+      y="0px"
+      viewBox="0 0 449 294"
+      style={{ enableBackground: 'new 0 0 449 294' }}
+      xmlSpace="preserve"
+    >
+      <path
+        style={pathStyle}
+        d="M381.4,241.1V50.5H190.8v190.6H381.4z M254.4,114.1h63.5v63.5h-63.6L254.4,114.1z"
+      />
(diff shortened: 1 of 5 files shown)


packages/wallet/ui/src/AgoricLogo.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+const AgoricLogo = () => {
+  const pathStyle = { fill: '#bb2d40' };
+
+  return (
+    <svg
```
samsiegart, line 7: For some reason loading a static resource was messing with the ability of the bridge to connect to localstorage, so just include this in the bundle

turadg: I recall there are some bad ways to use SVG in React but I don't remember whether this is one of them :)  Either way, seems fine for now. Optimize after correctness.

packages/wallet/ui/src/bridge.jsx
```diff
@@ -126,9 +129,15 @@ const signalBridge = () => {
...
 
-ReactDOM.render(
-  <button id="signalBridge" onClick={signalBridge}>
-    Signal Bridge
-  </button>,
-  document.getElementById('root'),
-);
+const Bridge = () => {
+  return (
+    <button className="button" onClick={signalBridge}>
```
turadg: `button.button` is a funny target. consider an `id` since there will be exactly one in the DOM:

```suggestion
    <button id="signalWallet" onClick={signalBridge}>
```

turadg (approved):
Some suggestions, none blocking

packages/wallet/ui/src/AgoricLogo.jsx
samsiegart: Fixed in https://github.com/Agoric/agoric-sdk/pull/5984/commits/d823ccd6a70b6d94606a55cc381b07f4edb35c97, but I'll leave this in anyway because it's small enough and will look better not loading in separately.

packages/wallet/ui/src/bridge.jsx
samsiegart: Changed to `signal-wallet-button`

dckc: That might have been me, aiming for type=button, which suppresses form submission, iirc

Issue #5975, opened 2022-08-17
xsnap leaks memory during snapshot write


warner:
I was having trouble reproducing #5910 locally, by replaying a transcript extracted from the original slogfile. I realized that one thing the original was doing, but my replay was not, was writing out the periodic heap snapshot (every 200 deliveries or so). When I added that to `packages/SwingSet/misc-tools/replay-transcript.js`, I saw an immediate growth of the `xsnap` process' `VmRSS` size, about 5MB for every snapshot it wrote.

I'll dive into `xsnap.c` next, but I'm guessing that a temporary buffer is allocated to write a snapshot, and then it is never freed.

warner:
My test program (https://github.com/Agoric/agoric-sdk/blob/614c849d61f8a2c24c82fbc48fcab29338774f25/packages/xsnap/test-snapshot-memleak.js) shows a memory leak of a trivial worker (just `lockdown` and our supervisor bundle) of about 30 kB per snapshot write. The leak is unaffected by putting large strings into the worker's heap, but adding new properties to the global object increases each leak by about 24 bytes. This suggests that the snapshot writing process is `malloc()`ing a temporary buffer based on the number of `slots` in use (not `chunks`, which hold strings), and then not freeing the buffer afterwards.

That makes the leak roughly proportional to the complexity of the heap (number of objects, properties, edges between objects, etc). Zoe is one of our more complicated vats, so the snapshots include a lot more slots, so the leak is faster.

I've let the Moddable folks know.

For now, our workaround of evicting each worker after each snapshot write is a decent one. If we really wanted to, we could claw back some performance by only evicting after every N snapshots, but I think it's better to wait for the upstream fix.

warner:
The [XS code](https://github.com/Moddable-OpenSource/moddable/blob/public/xs/sources/xsSnapshot.c) does a really good job of not allocating memory. And I don't see [anything](https://github.com/agoric-labs/xsnap-pub/blob/3e671651ed6d8a58491b500c9e7359494ff7da7d/xsnap/sources/xsnap-worker.c#L104) [obvious](https://github.com/agoric-labs/xsnap-pub/blob/3e671651ed6d8a58491b500c9e7359494ff7da7d/xsnap/sources/xsnap-worker.c#L491-L495) in the xsnap-worker.c code that might accumulate data created during the snapshot write.

[...]

PR #5987, opened 2022-08-17
fix(xsnap): use patched XS to fix heap-snapshot-writing memory leak

warner:
This switches our 'moddable' submodule to use a version that include
the patch for https://github.com/Agoric/agoric-sdk/issues/5975

The long-term fix will be for us to update to a new Moddable SDK
release, which will include this fix (among others).

refs #5975


dckc (approved):
looks good

cc @kriskowal @mhofman to advise: we're diverging from upstream temporarily

PR #5989, opened 2022-08-18
fix: prepare for inherited method representation

erights:
#5970 changes the representation of virtual and durable objects to be more class-like. This breaks autobinding (see explanation there), requiring us bind extracted methods to their receiver. Also, the methods of an object are no longer necessarily own properties. Rather, gathering all the methods requires us to ascend the inheritance chain.

This PR extracts those repairs into a prior PR that should work both before and after #5970 . Separated the PRs makes both more reviewable.

### DIFF
--- a/packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
+++ b/packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
@@ -4,7 +4,7 @@ import { AmountMath, AssetKind } from '@agoric/ertp';
 import { assertProposalShape } from '@agoric/zoe/src/contractSupport/index.js';
 import { ceilMultiplyBy } from '@agoric/zoe/src/contractSupport/ratio.js';
 import { makePublishKit } from '@agoric/notifier';
-import { M, matches } from '@agoric/store';
+import { bindAllMethods, M, matches } from '@agoric/store';
 import { defineKindMulti } from '@agoric/vat-data';
 import { makeTracer } from '../makeTracer.js';
 import { addSubtract, assertOnlyKeys, stageDelta } from '../contractSupport.js';
@@ -370,14 +370,17 @@ const potBehavior = {
     const { zcf } = state;
     const { helper } = facets;
     assert(state.open);
-    return zcf.makeInvitation(helper.adjustBalancesHook, 'AdjustBalances');
+    return zcf.makeInvitation(
+      seat => helper.adjustBalancesHook(seat),
+      'AdjustBalances',
+    );
   },
   /** @param {MethodContext} context */
   makeCloseInvitation: ({ state, facets }) => {
     const { zcf } = state;
     const { helper } = facets;
     assert(state.open);
-    return zcf.makeInvitation(helper.closeHook, 'CloseVault');
+    return zcf.makeInvitation(seat => helper.closeHook(seat), 'CloseVault');
   },
   /**
    * The actual current debt, including accrued interest.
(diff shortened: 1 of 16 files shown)


packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
```diff
@@ -425,9 +428,25 @@ const finish = ({ facets }) => {
...
+ *
+ * @param {ZCF} zcf
+ * @param {ZCFSeat} startSeat
+ * @param {import('./stakeFactoryManager.js').StakeFactoryManager} manager
+ * @throws {Error} if startSeat proposal is not consistent with governance parameters in manager
+ */
+export const makeStakeFactoryKit = (zcf, startSeat, manager) => {
+  const { helper, pot } = makeStakeFactoryKitInternal(zcf, startSeat, manager);
+  return harden({
+    helper: bindAllMethods(helper),
```
turadg, line 453: this seems like a kluge. Consider calling this out as tech debt, so it's not emulated.
```
XXX work-around for change from auto-binding
```
And a TODO on the outer function,
```
TODO refactor to remove this wrapper around the kind
```

packages/store/src/utils.js
```diff
@@ -0,0 +1,43 @@
...
+  right = String(right);
+  // eslint-disable-next-line no-nested-ternary
+  return left < right ? -1 : left > right ? 1 : 0;
+};
+
+/**
+ * @param {object} obj
+ * @returns {(string|symbol)[]}
+ */
+export const getMethodNames = obj => {
```
turadg, line 25: these utils don't seem very specific to store.

I think we'd be well served by https://github.com/Agoric/agoric-sdk/issues/5992

turadg (approved):

packages/store/src/utils.js
```diff
@@ -0,0 +1,43 @@
...
+    const mNames = ownKeys(obj).filter(name => typeof obj[name] === 'function');
+    result.push(...mNames);
+    obj = getPrototypeOf(obj);
+  }
+  result.sort(compareStringified);
+  return harden(result);
+};
+harden(getMethodNames);
+
+export const bindAllMethods = obj =>
```
Chris-Hibbert, line 65: I don't understand what bindAllMethods is achieving just from reading the code. Please add a comment.  

Why is it needed in stakeFactoryKit?  Why is it only needed there?

Chris-Hibbert (approved):

packages/inter-protocol/src/stakeFactory/stakeFactoryKit.js
erights: Done

packages/store/src/utils.js
erights: Done. You may want to have a look.

turadg (approved):

packages/store/src/utils.js
erights: Hi @turadg ,
I just read #5992 . I like it a lot! But are you suggesting I should do this on the way to merging this PR?

In the past I have been flumoxed by what's needed to set up a new package. Just too much boilerplate that I'm unfamiliar with that I nevertheless need to get right. @kriskowal at one point had an automated shell script for automating new package creation for the endo repository, but I was never able to get it to work.

May I proceed to merge this PR without taking this step? Instead, I will put in a TODO into utils.js linking to #5992

packages/store/src/utils.js
Chris-Hibbert: Thanks for the explanation. That clarifies it for me.

packages/store/src/utils.js
erights: I added that TODO comment to `utils.js`

turadg: SGTM. I don't see any reason for #5992 to block this. I'm mostly trying to drum up support for it. :)  I'd be happy to do the work if the issue is blessed. I'll assign it to myself.

erights: Thanks. I'll proceed with this PR now.

Yes, I like #5992 . Would love to see it happen -- thanks! Could you do it in such a way as to leave behind an understanding about all the boilerplate needed to start a new package?

PR #5990, opened 2022-08-18
fix(store)!: move some util where they are more reusable

erights:
Extracted from #5960 into a prior PR that should work before or after #5960 . #5960 is still draft, but this portion is ready to review.

Stacked on #5989

BREAKING CHANGE: `objectMap` must now be imported from `@agoric/store` rather than `@agoric/vat-data`. See #5992 for likely further migration.


packages/store/src/utils.js
```diff
@@ -1,8 +1,116 @@
...
+ *
+ * With these differences, even if the original object was not a CopyRecord,
+ * if all the mapped values are Passable, then the returned object will be
+ * a CopyRecord.
+ *
+ * @template {string} K
+ * @template T
+ * @template U
+ * @param {Record<K,T>} original
+ * @param {(value: T, key?: string) => U} mapFn
```
turadg, line 54: minor
```suggestion
 * @param {(value: T, key?: K) => U} mapFn
```

packages/vat-data/src/index.js
```diff
@@ -1,2 +1,7 @@
 export * from './vat-data-bindings.js';
 export * from './kind-utils.js';
+
+// Reexporting from here to minimize breakage during transition.
+// Everyone should get it directly from store rather than vat-data.
+// TODO Remove once it won't break anything.
+export { objectMap } from '@agoric/store';
```
turadg: It looks like this PR has already moved all references. When I check out this branch and do "Find all references" on this export, there are none.

I think it would be worth removing it now so autocomplete gets it from the right place.

Removing it is a breaking change to the package API so for semver it should make it into CHANGELOG.md as a breaking change. (That would require a `BREAKING CHANGE` string or `!` modifier in the commit message. ) But this package hasn't reached 1.0 yet so all bets are off and I'd be very surprised if any code outside SDK imports it.

turadg (approved):

packages/store/src/utils.js
erights: When I do this, I get
```
src/utils.js:60:53 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'K'.
  'string' is assignable to the constraint of type 'K', but 'K' could be instantiated with a different subtype of constraint 'string'.

60   const mapEnts = ents.map(([k, v]) => [k, mapFn(v, k)]);
                                                       ~
```
(this is probably how I ended writing the less natural type, but frankly I don't remember)

In any case, what do you recommend I do about this complaint?

packages/vat-data/src/index.js
erights: Done

packages/store/src/utils.js
erights: After double checking that `entries` (used inside `objectMap`) enumerates only string-named properties, I just removed the `K` type parameter, replacing it everywhere with `string`. PTAL.

erights: Since 
   * you've already approved
   * my removal of `K` seems right to both me and typescript
   * I can now get this PR on the merge train
   * the only possible controversy is static rather than dynamic behavior

I'm labeling it with automerge:squash until I hear otherwise. If this merges before you suggest a different fix, we can always to that one after merge

turadg: The way now loses the keys. E.g. `objectMap({a: 1}, v=>v).whatever` won't error.

I suggest [this](https://www.typescriptlang.org/play?target=99&filetype=js#code/PQKhCgAIUgBAXApgWwA4BsCGTIG8D2ARgFaIDG8AvpACpQyyqYBOmyeN1+zAlgOY8AdpnT04SNFhwAlSMkypIzRAGcArunhjGLNngAUANxFrEALloAaSAGtEATwD8Fu-fwAzWgEpIAXgB8kNLU8qgAYoLayvBqzIIqeLiQANoACpBCtg4etAC6FsGU9MDgiAAeqNzwkGT48dVEpBQAsgp+kPrc-EIi1qERPgF4UDV1KtWIgvAJvpAA8iTk8AB0k-C8qp28AsLoXgDcI7X1cgoAolMzkGsqy6H6+sk21oa5g4FPfQoRRtY2Xm9DpAlIgYnF5osKMt3Mx8MgLuseJtQgiVAdwJRDuBjuNII12nhMBYAIzWQgWABMgkx2LG1QAJvg1IR0IgFsR2o0lq1UJ0SC8RH5AsY9ljgMBICoABZM9D067MWHMSCENQM-CqQQAcmqAHcePApVl7BZxrxBHxwIzmaz2XceCoVEI+IdrSy2SRlkA) instead.

Screen Shot 2022-08-18 at 4 13 15 PM

Issue #5992, opened 2022-08-18
package for sdk-internal exports


turadg:
## What is the Problem Being Solved?

We have some exports that don't fit conceptually into any of our agoric-sdk packages. E.g. `objectMap`. Such utilities are so generic that in most codebases they'd be imported from NPM (e.g. Lodash). 

Sometimes we do pick an arbitrary package, which can result in awkward dependency graphs (e.g. [cycles](https://github.com/Agoric/agoric-sdk/issues/4645)). Improving those graphs requires refactoring to a new package, which causes a breaking change to the export set. It's also often another imperfect home because we don't have a home for basic utils.

We have in places resorted to importing in code without declaring the package dependency. There are 174 results for disabling `import/no-extraneous-dependencies`. Most are in test code for setting up Endo globals (including `VatData`). That may be best solved by https://github.com/Agoric/agoric-sdk/issues/5244 as a `devDependency`. This issue is for a package that's a `dependency`.

## Description of the Design

Make a new package that is obviously not part of the public API of Agoric-SDK and move non-public utilities there. To convey that it shouldn't ever be depended on not to break from one release to another, it should never reach 1.0 which in semver means no guarantee. By Hyrum's Law someone will import it anyway and depend on that. It would still be better than the status quo in which we export from packages that are declared to be part of the SDK API. (Even when they aren't at the top level, many are available by deep exports https://github.com/Agoric/agoric-sdk/issues/5808) 

Some ideas for this package name:
- `@agoric/internal`
- `@agoric/private`
- `@agoric/unsupported`
- `agoricutils` (à la https://www.npmjs.com/package/fbjs )

## Security Considerations
--
## Test Plan

--

kriskowal:
What concretely do you wish to populate this package with?

erights:
I've been accumulating some at https://github.com/Agoric/agoric-sdk/blob/markm-interface-guards/packages/store/src/utils.js as I need them. All of these are potentially useful to any hardened js programmer, though some are admittedly unusual. In any case, none of these are specific to the store package. The only reason I accumulated them there is it was low enough on the import graph that everything that needed any of these could import them. Not a great reason.

erights:
Just looked them over. The only one that is compelling to move to somewhere more principled is `objectMap`. All the other are rather obscure things to want.

However, I am confident that once such a package exists, we will find other things that genuinely deserve being moved.

kriskowal:
[...]

PR #5993, opened 2022-08-18
improve static typing of contract start terms

turadg:
## Description

Work around https://github.com/endojs/endo/issues/1257 by defining a local `deeplyFulfilledObject`.

Use it in contract starts to restore type safety (see https://github.com/Agoric/agoric-sdk/pull/5930#issuecomment-1218565035 )

### Security Considerations

--

This may not be something we want to include in the API backwards compatibility. Once Endo has better type annotations this won't be necessary.

### Testing Considerations

No change in code. Tested types manually:
Screen Shot 2022-08-18 at 7 32 33 AM
Screen Shot 2022-08-18 at 7 47 15 AM

### DIFF
--- a/packages/inter-protocol/src/proposals/addAssetToVault.js
+++ b/packages/inter-protocol/src/proposals/addAssetToVault.js
@@ -101,7 +101,7 @@ export const publishInterchainAssetFromBank = async (
       consume: { [Stable.symbol]: runIssuer },
     },
     brand: {
-      consume: { [Stable.symbol]: runBrandP },
+      consume: { [Stable.symbol]: stableBrandP },
     },
   },
   { options: { interchainAssetOptions } },
@@ -116,7 +116,6 @@ export const publishInterchainAssetFromBank = async (
   assert.typeof(decimalPlaces, 'number');
   assert.typeof(proposedName, 'string');
 
-  /** @type {import('@agoric/vats/src/mintHolder.js').AssetTerms} */
   const terms = {
     keyword,
     assetKind: AssetKind.NAT,
@@ -125,16 +124,16 @@ export const publishInterchainAssetFromBank = async (
       assetKind: AssetKind.NAT,
     },
   };
-  const { creatorFacet: mint, publicFacet: issuerP } = E.get(
+  const { creatorFacet: mintP, publicFacet: issuerP } = E.get(
     E(zoe).startInstance(mintHolder, {}, terms),
   );
 
   const [issuer, brand, runBrand] = await Promise.all([
     issuerP,
     E(issuerP).getBrand(),
-    runBrandP,
+    stableBrandP,
   ]);
-  const kit = { mint, issuer, brand };
+  const kit = { mint: mintP, issuer, brand };
 
   await addPool(zoe, amm, issuer, keyword, brand, runBrand, runIssuer);
 
@@ -142,7 +141,8 @@ export const publishInterchainAssetFromBank = async (
   produceBankMints.resolve([]);
   await Promise.all([
     Promise.resolve(bankMints).then(
-      mints => mints.push(mint),
+      // @ts-expect-error pushing a promise to a presence array
+      mints => mints.push(mintP),
       () => {}, // If the bankMints list was rejected, ignore the error.
     ),
     E(E(agoricNamesAdmin).lookupAdmin('issuer')).update(keyword, issuer),
--- a/packages/inter-protocol/src/proposals/addAssetToVault.js
+++ b/packages/inter-protocol/src/proposals/addAssetToVault.js
@@ -1,10 +1,9 @@
 // @ts-check
 import { AmountMath, AssetKind } from '@agoric/ertp';
 import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
+import { deeplyFulfilledObject } from '@agoric/internal';
 import { Stable } from '@agoric/vats/src/tokens.js';
 import { E } from '@endo/far';
-import { deeplyFulfilled } from '@endo/marshal';
-
 import { reserveThenGetNames } from './utils.js';
 
 export * from './startPSM.js';
@@ -218,7 +217,7 @@ export const registerScaledPriceAuthority = async (
     runBrand,
   );
 
-  const terms = await deeplyFulfilled(
+  const terms = await deeplyFulfilledObject(
     harden({
       sourcePriceAuthority,
       scaleIn,
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -182,6 +182,7 @@ export const startInterchainPool = async (
     { Central: centralIssuer },
     terms,
     {
+      // @ts-expect-error XXX remotable types
       bankManager,
     },
   );
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -13,7 +13,7 @@ import {
 import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
 import { E, Far } from '@endo/far';
 import { Stable, Stake } from '@agoric/vats/src/tokens.js';
-import { deeplyFulfilled } from '@endo/marshal';
+import { deeplyFulfilledObject } from '@agoric/internal';
 import * as Collect from '../collect.js';
 import { makeTracer } from '../makeTracer.js';
 import { makeStakeReporter } from '../my-lien.js';
@@ -171,7 +171,7 @@ export const startInterchainPool = async (
     mgrP,
   ]);
 
-  const terms = await deeplyFulfilled(
+  const terms = await deeplyFulfilledObject(
     harden({
       minimumCentral: AmountMath.make(centralBrand, minimumCentral),
       amm,
@@ -248,7 +248,7 @@ export const setupAmm = async (
   );
   const marshaller = await E(board).getPublishingMarshaller();
 
-  const ammGovernorTerms = await deeplyFulfilled(
+  const ammGovernorTerms = await deeplyFulfilledObject(
     harden({
       timer: chainTimerService,
       electorateInstance,
@@ -344,7 +344,7 @@ export const setupReserve = async ({
   const storageNode = await makeStorageNodeChild(chainStorage, STORAGE_PATH);
   const marshaller = await E(board).getReadonlyMarshaller();
 
-  const reserveGovernorTerms = await deeplyFulfilled(
+  const reserveGovernorTerms = await deeplyFulfilledObject(
     harden({
       timer: chainTimerService,
       electorateInstance,
@@ -487,7 +487,7 @@ export const startVaultFactory = async (
     },
   );
 
-  const governorTerms = await deeplyFulfilled(
+  const governorTerms = await deeplyFulfilledObject(
     harden({
       timer: chainTimerService,
       electorateInstance: economicCommitteeInstance,
@@ -688,7 +688,7 @@ export const startRewardDistributor = async ({
 }) => {
   trace('startRewardDistributor');
   const timerService = await chainTimerService;
-  const feeDistributorTerms = await deeplyFulfilled(
+  const feeDistributorTerms = await deeplyFulfilledObject(
     harden({
       timerService,
       collectionInterval: 60n * 60n, // 1 hour
@@ -879,7 +879,7 @@ export const startStakeFactory = async (
   const storageNode = await makeStorageNodeChild(chainStorage, STORAGE_PATH);
   const marshaller = await E(board).getReadonlyMarshaller();
 
-  const stakeTerms = await deeplyFulfilled(
+  const stakeTerms = await deeplyFulfilledObject(
     harden({
       timer: chainTimerService,
       electorateInstance,
--- a/packages/inter-protocol/src/proposals/price-feed-proposal.js
+++ b/packages/inter-protocol/src/proposals/price-feed-proposal.js
@@ -4,7 +4,7 @@ import {
   makeStorageNodeChild,
   assertPathSegment,
 } from '@agoric/vats/src/lib-chainStorage.js';
-import { deeplyFulfilled } from '@endo/marshal';
+import { deeplyFulfilledObject } from '@agoric/internal';
 
 import { unitAmount } from '@agoric/zoe/src/contractSupport/priceQuote.js';
 import { reserveThenDeposit, reserveThenGetNames } from './utils.js';
@@ -118,7 +118,7 @@ export const createPriceFeed = async (
 
   const unitAmountIn = await unitAmount(brandIn);
   /** @type {import('@agoric/zoe/src/contracts/priceAggregator.js').PriceAggregatorContract['terms']} */
-  const terms = await deeplyFulfilled(
+  const terms = await deeplyFulfilledObject(
     harden({
       ...contractTerms,
       description: AGORIC_INSTANCE_NAME,
--- a/packages/inter-protocol/src/proposals/price-feed-proposal.js
+++ b/packages/inter-protocol/src/proposals/price-feed-proposal.js
@@ -104,7 +104,7 @@ export const createPriceFeed = async (
   /**
    * Values come from economy-template.json, which at this writing had IN:ATOM, OUT:USD
    *
-   * @type {[[Brand, Brand], [Installation]]}
+   * @type {[[Brand, Brand], [Installation<import('@agoric/zoe/src/contracts/priceAggregator.js').start>]]}
    */
   const [[brandIn, brandOut], [priceAggregator]] = await Promise.all([
     reserveThenGetNames(E(agoricNamesAdmin).lookupAdmin('oracleBrand'), [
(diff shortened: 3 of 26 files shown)


packages/inter-protocol/src/proposals/price-feed-proposal.js
```diff
@@ -109,7 +109,7 @@ export const createPriceFeed = async (
   ]);
 
   /** @type {import('@agoric/zoe/src/contracts/priceAggregator.js').PriceAggregatorContract['terms']} */
-  const terms = await deeplyFulfilled(
+  const terms = await deeplyFulfilledObject(
```
Chris-Hibbert, line 121: Lint is reminding you here that the import wasn't updated in htis file.

Chris-Hibbert (approved):
I don't understand the necessity, but if it's required to support typing, I'm fine with it.

packages/inter-protocol/src/proposals/addAssetToVault.js
```diff
@@ -126,14 +124,13 @@ export const publishInterchainAssetFromBank = async (
       assetKind: AssetKind.NAT,
     },
   };
-  const { creatorFacet: mint, publicFacet: issuerP } = E.get(
-    E(zoe).startInstance(mintHolder, {}, terms),
-  );
+  const { creatorFacet: mint, publicFacet: issuer } = await E(
```
erights: If the old code didn't need to pay a round trip delay here, why does the new code need to? I this a typing issue?

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -10,7 +10,7 @@ import { makeStorageNodeChild } from '@agoric/vats/src/lib-chainStorage.js';
 import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
 import { E, Far } from '@endo/far';
 import { Stable, Stake } from '@agoric/vats/src/tokens.js';
-import { deeplyFulfilled } from '@endo/marshal';
+import { deeplyFulfilledObject } from '@agoric/zoe/src/contractSupport/zoeHelpers.js';
```
erights: Can we put it somewhere we don't need to deeply import to get it?

packages/vat-data/src/kind-utils.js
```diff
@@ -21,7 +21,6 @@ export const ignoreContext =
   fn =>
   (context, ...args) =>
     fn(...args);
-// @ts-expect-error TODO statically recognize harden
```
erights, line 25: Yay! You got rid of this annoyance! How did you do that, and can we do it elsewhere we suffer from this annoyance? (I admit I never understood the issue that was blocking this)

Hmm. Maybe this PR does fix all those places. Do you know if there are others?

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -394,3 +395,25 @@ export const checkZCF = (zcf, assertFn) => {
...
+ * Caveats:
+ * - doesn't recur within Promise results
+ * - resulting type has wrapper in its name
+ *
+ * @template T
+ * @typedef {T extends PromiseLike ? Awaited<T> : T extends {} ? DeeplyAwaitedObject<T> : Awaited<T>} DeeplyAwaited
+ */
+
+/**
+ * A more constrained version of {deeplyFulfilled} for type safety until https://github.com/endojs/endo/issues/1257
```
erights: I would expect "more constrained" to mean that it imposes constraints that the original does not. I see here that it is just a renaming of the original, which is great! It is only better typed. Its runtime behavior is identical.

Wonderful that you could get the types to reflect even some of the recursion! Do you think it is possible to recur into promise types as well? No priority to do so now. I'm just wondering what we know about what's possible in TS.

On the renaming, `deeplyFulfilledObject` sounds like it applies only to objects, not to primitive values. But this is only a renaming, and the original also applies to primitive values.

packages/inter-protocol/src/proposals/addAssetToVault.js
Chris-Hibbert: Do we have adequate type info for `E.get()`? If so, this should be unnecessary. If not, that could be the typing issue.

erights: If the cost of static typing is extra rounds trips, I'm worried.

turadg: Improvements in static typing must never be at the expense of runtime efficiency.

The types did prompt me to simplify but the reason (I believe) it doesn't cost efficiency is that these are being awaited in this function anyway. On the next line, `issuerP` was being awaited and lower down there's a `Promise.all` to resolve the `mint`. Does the change here reduce efficiency in a way I missed?

packages/inter-protocol/src/proposals/econ-behaviors.js
turadg: Out of scope, imo. There are many `zoeHelpers.js` exports already and if we want to stop that there should be a separate PR. Note that until we can get [ESM subpath exports working](https://github.com/Agoric/agoric-sdk/pull/6001#discussion_r950704872) it'll mean putting all exports in one path, and I don't think these helpers belong at `@agoric/zoe`.

packages/vat-data/src/kind-utils.js
turadg: I think this line did it https://github.com/Agoric/agoric-sdk/blob/4f51b415ed6602679c8d3c36f32ae2b13a0f061f/packages/vat-data/src/index.js#L1

packages/zoe/src/contractSupport/zoeHelpers.js
turadg: > On the renaming, deeplyFulfilledObject sounds like it applies only to objects, not to primitive values.

Correct. That's what the types denote.

> I would expect "more constrained" to mean that it imposes constraints that the original does not. 

It in the sense you describe: it accepts only objects. 

> But this is only a renaming, and the original also applies to primitive values

It's an encapsulation. That `deeplyFulfilledObjects` uses `deeplyFulfilled` is an implementation detail. Callers of `deeplyFulfilledObjects` should not expect that behavior and will get a static type error if they try to use it more generally.

> It is only better typed. … Wonderful that you could get the types to reflect even some of the recursion!

Restricting to object types is part of what achieved this. We could to more work to support all the domain of `deeplyFulfilled` but I think that's Endo work and it's scope creep for solving type safety for contract terms.

> Do you think it is possible to recur into promise types as well?

It goes one level deep into the resolving Promise types (with `Awaited`):
https://github.com/Agoric/agoric-sdk/blob/4f51b415ed6602679c8d3c36f32ae2b13a0f061f/packages/zoe/src/contractSupport/zoeHelpers.js#L410

That's all that is needed for the terms validation. It should be possible to go arbitrarily deep. Just as [ReadonlyDeep](https://github.com/sindresorhus/type-fest/blob/main/source/readonly-deep.d.ts) does, but it'll take some more supporting types like [Primitive](https://github.com/sindresorhus/type-fest/blob/main/source/primitive.d.ts). I think solving this fully should be done in Endo. Agoric-SDK is just temporarily working around Endo's poor type definitions. E.g. https://github.com/endojs/endo/issues/1257

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -394,3 +395,25 @@ export const checkZCF = (zcf, assertFn) => {
...
+ * Useful in starting contracts that need all terms to be fulfilled in order to be durable.
+ *
+ * @type {<T extends {}>(unfulfilledTerms: T) => DeeplyAwaited<T>}
+ */
+export const deeplyFulfilledObject = deeplyFulfilled;
```
erights: In the PR comment, you say

> If we had https://github.com/Agoric/agoric-sdk/issues/5992 I'd have put it there.

We do now ;)

packages/inter-protocol/src/proposals/addAssetToVault.js
erights: Yes, you missed it ;)

Before this PR, we  send the async `startInstance(...)` message and immediately holds the promise `issuerP` for the issuer we're expecting, without having heard back yet. Then, through the magic of promise pipelining, the 
```js
  const [issuer, brand, runBrand] = await Promise.all([
    issuerP,
    E(issuerP).getBrand(),
    runBrandP,
  ]);
```
first, immediately sends the `getBrand()` message on the remote promise held by `issuerP`. Now, for one round trip, we should hear both about how the `issuerP` promise settles, and about how the promise for the result of the `getBrandP()` message settles. After this PR, you do one round trip to get the issuer, and then do another round trip to get that issuer's brand, completely losing the benefits of promise pipelining for a natural promise pipelining example.

erights: Yes, you missed it ;)

Before this PR, we  send the async `startInstance(...)` message and immediately holds the promise `issuerP` for the issuer we're expecting, without having heard back yet. Then, through the magic of promise pipelining, the 
```js
  const [issuer, brand, runBrand] = await Promise.all([
    issuerP,
    E(issuerP).getBrand(),
    runBrandP,
  ]);
```
first, immediately sends the `getBrand()` message on the remote promise held by `issuerP`. Now, for one round trip, we should hear both about how the `issuerP` promise settles, and about how the promise for the result of the `getBrandP()` message settles. After this PR, you do one round trip to get the issuer, and then do another round trip to get that issuer's brand, completely losing the benefits of promise pipelining for a natural promise pipelining example.

erights (changes requested):

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Not a comment on the code of the PR, but just a note about our curious divergence in terminology I see in your comments above:

> It in the sense you describe: it accepts only objects.

But it doesn't accept only objects. The runtime behavior is unchanged. `deeplyFulfilledObjects` still accepts primitive values. It is only the types describing only the subset of its behavior in which only objects are provided. The reason I call attention to these terminology issues is that we need to maintain clarity about what the runtime behavior of the code actually enforces. Any further restrictions implied by the erased type system are a fiction that, if taken seriously, can lead to a false sense of security and to missing the possibility of an attack, merely because the attack is outside what the types describe.

> It's an encapsulation. That deeplyFulfilledObjects uses deeplyFulfilled is an implementation detail. Callers of deeplyFulfilledObjects should not expect that behavior and will get a static type error if they try to use it more generally.

That's only true if we take the specification of `deeplyFulfilledObjects` to cover only the case where the inputs are well typed. This is the underspecification sin of the "unspecified behavior" that non-secure languages practice, that is incompat with secure languages.

What would be an encapsulation:

```js
const deeplyFulfilledObjects = specimen => {
  assert(isObject(specimen));
  return deeplyFulfilled(specimen);
};
```

This successfully encapsulates the behavior of `deeplyFulfilled` on primitive values in the sense that this behavior is no longer reachable.

We need alignment on how we describe such things, because confusion between us about what is actually enforced could be fatal.

erights: Likewise in the PR comment

> Once Endo has type safety

Hard to imagine that Endo will have type safety before TypeScript does.

packages/inter-protocol/src/proposals/addAssetToVault.js
erights: I see you fixed this. Resolving.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: I'm leaving this one unresolved until I review the comments in the code itself, which may or may not have the same issue as the text in these PR conversations.

packages/inter-protocol/src/proposals/econ-behaviors.js
erights: Resolving

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Thanks. Resolving.

packages/inter-protocol/src/proposals/price-feed-proposal.js
erights: I do not see any current lint warning, so resolving.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Yeah, what I see there:

image

I did the screenshot with hover to show the consequence of putting the misleading statement within `/**...*/` and the accurate statement within `// ...`. Anyone hovering over the name at a callsite will see the statement that is still misleadingly strong --- that overpromises and underdelivers. Not good for defensive programming.

erights (changes requested):
The misleading comment text (with the accurate but disappearing qualifier) is the only remaining hard issue. Terminology about unsound types must never be misunderstood as implying enforcement.

packages/zoe/src/contractSupport/zoeHelpers.js
turadg: > Terminology about unsound types must never be misunderstood as implying enforcement.

Granted. I'll try to be more sensitive to this. FWIW, my read of the function signature is that if if you give it anything other than an object, all bets are off. In my experience as a JS developer, I don't expect the function to check all my arguments and tell me which aren't valid.

However, I agree that it is strictly better to do so and I've updated the PR with your suggestion.

> Hard to imagine that Endo will have type safety before TypeScript does.

If you're saying that TypeScript does not "have type safety" then we must be using this terminology differently. I don't see it as a binary feature and it seems evident to me that TypeScript gives more safety in types fitting in production and use of symbols. It's not perfectly sound, but I don't see that as a worthy goal. Java is relatively sound but still has NullPointerException and that was a pragmatic design decision.

What does "type safety" mean to you? Maybe a good chat thread. Meanwhile I've revised the text in PR description to say "Once Endo has better type annotations this won't be necessary."

erights (approved):
LGTM, thanks.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: > all bets are off.

"all bets are off" is exactly right. This is programming with preconditions, where a precondition violation waives all specified properties and instead allow "undefined behavior". This is the opposite of *defensive consistency* (see my thesis), and presents a golden opportunity to an attacker who can cause preconditions to be violated. Defensive consistency starts by generally replacing preconditions with input validation.

Since we're building a platform for mutually suspicious parties to interact via smart contracts, including those representing prediction markets, "all bets are off" is an elegant summary of why we cannot allow such situations ;)

> In my experience as a JS developer, I don't expect the function to check all my arguments and tell me which aren't valid.

How much of that experience is in a fine-grain adversarial context, where you have direct contact between mutually suspicious objects, each responsible for defending its invariants from its immediate clients?

Defensive consistency does *not* require a function to do full input validation. But it does require stated bounds on the misbehavior that any input can cause, and confidence that the implementation succeeds at enforcing these bounds.

> Java is relatively sound but still has NullPointerException

On a program that statically type checks, Java makes some hard soundness guarantees about what absolutely cannot happen at runtime. NullPointerException is not one of them. Generic (i.e., parameterized) types are also not sound. Java generics did originally claim that a whole program that passed the type checker with no warnings was type safe wrt generics, but that was shown to be untrue. So Java is only type safe wrt outer (non-type-parameter) types under the interpretation that all object types are nullable types. But within that narrow constraint, AFAIK, Java is absolutely type safe. I have certainly written [Joe-E](https://en.wikipedia.org/wiki/Joe-E) code whose security depends on such guarantees. (Joe-E is an ocap subset of Java, enforced by putting additional static checks in from of the Java compiler.)

AFAIK, a TS program passing its type check with no warnings does not guarantee any runtime invariant whatsoever. I would like to be wrong about this.

erights: Partially in reaction to Java generics, C# parameterized types are sound.

erights: Here's an interesting riddle (from the Joe-E work) for understanding the difference between Java and JVM soundness. Does the following Java function always return true? (I've forgotten my Java syntax, so please let me know if it not clear what syntactically valid Java program I mean)

```java
boolean isAByte(byte b) {
  return -128 <= b && b <= 255;
}
```

May a Java compiler validly optimize this function to

```java
boolean isAByte(byte b) {
  return true;
}
```
?

erights: I forgot at first that Java bytes are signed, so I changed the lower bound above. I left the overly generous upper bound since it does not effect the riddle.

erights: <details>
<summary>Spoiler alert</summary>

For a hypothetical Java language as specified and implemented independent of the JVM semantics, I imagine that it could validly optimize the body to `return true;`. But, AFAIK, such a hypothetical JVM-independent Java semantics does not exist.

For actual Java, a Java function may be called from a JVM function that could not have been written in Java. The JVM type rules recognizes integral typed words (32 bits? 64 bits? I forget) but does not recognize or validate typing rules restricting numeric ranges to smaller than a word. So the Java function above, linked with a valid JVM caller that passes the JVM bytecode verifier, can return false, and therefore cannot be optimized to `return true;`. This divergence between what's possible in Java vs JVM is known (bizarrely IMO) as a "failure of full abstraction".

</details>

PR #5995, opened 2022-08-18
chore(swingset): add/improve some debug tools

warner:
Add/edit some slogfile analysis tools in swingset/misc-tools/


warner:
closed in favor of #6043

PR #5996, opened 2022-08-18
fix(wallet/ui): fix gas price in suggest chain

samsiegart:
refs: https://github.com/Agoric/dapp-treasury/issues/137


packages/wallet/ui/src/util/SuggestChain.js
```diff
@@ -42,6 +42,11 @@ const makeChainInfo = (networkConfig, rpcAddr, chainId, caption) => {
     currencies: [stakeCurrency, stableCurrency],
     feeCurrencies: [stableCurrency],
     features: ['stargate', 'ibc-transfer'],
+    gasPriceStep: {
+      low: 0,
+      average: 0,
+      high: 0,
```
dckc, line 48: maybe 'high' should be >0... someday...

dckc (approved):
yup; I saw several people test this.

PR #4422, opened 2022-01-31
feat: NonNullish checked cast

turadg:
## Description

Sometimes you or the type checker doesn't know that a value will be defined at runtime. This adds a utility that casts to a non-nullish version of the value, with a check that the cast is valid.

### Security Considerations

This function may be used in many places.

### Documentation Considerations

No action needed.

### Testing Considerations

- [x] needs tests

PR #5998, opened 2022-08-18
fix: avoid relying on bound `E` proxy methods

mhofman:
Layered on top of https://github.com/Agoric/agoric-sdk/pull/5989.

Fixes compatibility for upcoming endo breaking change: https://github.com/endojs/endo/pull/1255


packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -497,7 +497,9 @@ export const startVaultFactory = async (
   ]);
 
   const voteCreator = Far('vaultFactory vote creator', {
-    voteOnParamChanges: E(governorCreatorFacet).voteOnParamChanges,
+    /** @type {VoteOnParamChanges} */
+    voteOnParamChanges: (...args) =>
```
turadg, line 501: this makes me think it might be worth having a helper to bind a function and keep its type signature

packages/inter-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
```diff
@@ -350,7 +350,8 @@ const start = async (zcf, privateArgs, baggage) => {
     secondaryBrandToLiquidityMintProvider,
     () => {
       assert(reserveFacet, 'Must first resolveReserveFacet');
-      return E(reserveFacet).addIssuerFromAmm;
+      const reserveFacetP = E(reserveFacet);
+      return secondaryBrand => reserveFacetP.addIssuerFromAmm(secondaryBrand);
```
turadg: why the const?
```suggestion
      return secondaryBrand => E(reserveFacet).addIssuerFromAmm(secondaryBrand);
```

packages/notifier/src/storesub.js
```diff
@@ -59,13 +59,13 @@ export const makeStoredSubscriber = (subscriber, storageNode, marshaller) => {
 
   /** @type {Unserializer} */
   const unserializer = Far('unserializer', {
-    unserialize: E(marshaller).unserialize,
+    unserialize: data => E(marshaller).unserialize(data),
```
turadg, line 62: would it be reasonable to have a utility to attenuate a far object? This is basically passing on a subset of `marshaller`.

packages/inter-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
mhofman: Because of types, in the calback TS think `reserveFacet` could have gone back to null. Since it's ok to build and cache the proxy, I'm keeping it in a const to satisfy types. A small deviation on my rule to not change runtime code to make types happy.

packages/notifier/src/storesub.js
mhofman: Sure, but I don't know if the complexity of such a generic tool would be amortized?

packages/inter-protocol/src/proposals/econ-behaviors.js
erights: What would such a helper look like?

packages/inter-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
erights: A bare `E(expression)` not followed by a dot is disturbing. In the hypothetical world where we're using tildot, it would not even be possible to write it syntactically. I don't understand the typing problem --- but I usually don't. Would the tildot equivalent of @turadg 's suggested one liner have the same typing problem?

packages/notifier/src/storesub.js
erights: This is a good suggestion. The loss of autobinding raises the notational cost of manually attenuating, thereby potentially raising the comparative benefit of such a utility. Let's keep this in mind.

erights:
One of your CI tests failed with the known intermitted gc problem, so I took it upon myself to ask CI to rerun failed jobs.

erights (approved):
LGTM

But I'd really like to avoid the bare `E(expression)` expression. I'd much rather see an `// @ts-exprect-error` with an explanation.

packages/inter-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
mhofman: I went for casting instead. As agreed, we shouldn't go through runtime hoops to satisfy the type system.
(Really wish we had non null assertions in JS or at least JSDoc)

turadg: https://github.com/Agoric/agoric-sdk/pull/4422 got stalled out. Should I push it through?

packages/inter-protocol/src/proposals/econ-behaviors.js
turadg: _waves hand_… ??
```
    voteOnParamChanges: bindHere(E(governorCreatorFacet).voteOnParamChanges))
```
or maybe something more targeted,
```
    voteOnParamChanges: E.delegate(governorCreatorFacet).voteOnParamChanges,
```

Either way we could define the types so the signature doesn't have to be repeated.

packages/inter-protocol/src/vpool-xyk-amm/multipoolMarketMaker.js
erights: Yes, thanks!

packages/inter-protocol/src/proposals/econ-behaviors.js
erights: How would `bindHere` or `E.delegate` obtain the `governorCreatorFacet`? It would need that in order to do the binding. Are you thinking of a hidden coordination between `E` and (`bindHere` or `E.delegate`)? Not out of the question, but seems like too much magic to me. I'd rather see the manual binding, I think. But perhaps examples would make a compelling case?

turadg: I haven't thought it through. If I find time to I'll back with examples

Issue #5808, opened 2022-07-22
stop exporting src paths / stop deep importing


turadg:
## What is the Problem Being Solved?

19 packages export `"src"` or some variant (`"src/"`, `"src/**/*.js").

This exposes surface that is not meant to be part of the commitment of the package. 

## Description of the Design

Choose what to export and export only that. May require some package refactoring such as https://github.com/Agoric/agoric-sdk/issues/5244

mhofman:
To clarify, we can still package the src folder but would use an `exports` field in it to control what is actually exported for public consumption?

turadg:
> To clarify, we can still package the src folder but would use an `exports` field in it to control what is actually exported for public consumption?

Good clarification.  Yes, using the [`exports` field](https://nodejs.org/docs/latest-v14.x/api/packages.html#packages_main_entry_point_export) and the "src" files would still need to be packaged.

Note when I tried [exports for a new package](https://github.com/Agoric/agoric-sdk/pull/6001#discussion_r950696910) I ran into module resolution problems with ESlint and TypeScript due to tooling limitations. Haven't dug in enough to know whether there are workarounds or we just have to wait.

mhofman:
Exports and include should be orthogonal. You still need the include.

PR #6001, opened 2022-08-19
5992 internal package

### LINKED ISSUE #5992: package for sdk-internal exports

turadg:
closes: #5992

## Description

New package `@agoric/internal`, populated with non-store utils that were in `@agoric/store/src/utils.js`.

Also improves the `objectMap` type to maintain keys.

### Security Considerations

--

Explain clearly that the package is not and never will be part of the SDK's public API.

### Testing Considerations

`objectMap` now has a test.
Other had no direct tests but they're used widely enough they are covered.


packages/internal/src/utils.js
```diff
@@ -48,16 +42,16 @@ const { details: X } = assert;
...
  *
- * @template T
- * @template U
- * @param {Record<string,T>} original
- * @param {(value: T, key?: string) => U} mapFn
- * @returns {Record<string,U>}
+ * @template {Record<string, any>} O
+ * @param {O} original
+ * @template R map result
+ * @param {(value: O[keyof O], key?: keyof O) => R} mapFn
```
erights: Is the `?` right? It seems right at first, but it is in the backwards direction. It is supposed to mean that callers may omit the argument. Here, the one caller always provides the argument. It is the function itself that often ignores it.

erights: Also, lemme see if I understand. Type newbie questions.

We are not expression as much precision as we'd like, perhaps because more precision is impossible in TS. But the imprecision is that `R` types the values of all the properties of the result. Just like you have `O` be the type of the input record as a whole, could you have `R` be the type of the output record as a whole, and use typescript operators like `R[keyof O]` to say that the output type of the value associated with each key depends on the type of the input value associated with that same key?

packages/wallet/contract/src/support.js
```diff
@@ -1,6 +1,6 @@
 // @ts-check
 
-import { buildRootObject } from '@agoric/wallet-backend/src/wallet';
+import { buildRootObject } from '@agoric/wallet-backend/src/wallet.js';
```
erights, line 3: This is a drive by fix, unrelated to this PR, right? (No change suggested)

erights (approved):
LGTM.

As I saw @kriskowal mention somewhere, it seems strange that this is agoric-sdk rather than endo. I think I understand your reasons. But as we migrate some other packages from agoric-sdk to endo, if we migrate a package that depends on this one, how should we handle that? I do not want to create any dependencies from the endo repository into the agoric-sdk repository.

Nevertheless, no reason to block this PR on this question. Even if we do eventually migrate it to endo, we're a lot better off having first separated it at all.

Separately

There is an incredible amount of boilerplate that you've carefully gotten right here. Were I to create a separate package in the future, I'd cargo cult. Would it makes sense to create a pristine copy of only the generic bits of this as a template that one could start from, without accidentally including the parts you meant to be specific to this particular package?

Obviously, no reason to block this PR on that question.

Thanks!

packages/wallet/contract/src/support.js
turadg: Right. It's in a separate cleanup commit [033d122](https://github.com/Agoric/agoric-sdk/pull/6001/commits/033d1223a1aa751b1ba08790f2c6437408e21929)

packages/internal/src/utils.js
turadg: > Is the ? right? It seems right at first, but it is in the backwards direction.

You're right, this was an error. It should be `key` as the function passed in will always be called with it. I've fixed and added tests

turadg: > We are not expression as much precision as we'd like, perhaps because more precision is impossible in TS. But the imprecision is that `R` types the values of all the properties of the result.

Correct. This is treating the object as a Record of keys to some single type, which will be a union of all the types that are present. So the mapFn has to take all those. What it returns is called R, which could also be a union.

> say that the output type of the value associated with each key depends on the type of the input value associated with that same key?

I tried that but the `mapFn` is what defines the actual mapping and it would have to define the input/output for key key. I don't think that's practical.

turadg:
> As I saw @kriskowal mention somewhere, it seems strange that this is agoric-sdk rather than endo. I think I understand your reasons.

To recap: when developing agoric-sdk, getting things into Endo should not be a concern that blocks any particular PR. Getting things into Endo is a separate concern that has different stakeholders than any arbitrary PR in this repo.

> But as we migrate some other packages from agoric-sdk to endo, if we migrate a package that depends on this one, how should we handle that? I do not want to create any dependencies from the endo repository into the agoric-sdk repository.

Agree Endo should never depend on Agoric. Anything that that moves from agoric-sdk to endo will need to have its full dependency graph fulfilled without agoric-sdk. That's an appropriate time for "how should this fit this into Endo" discussion.

> Nevertheless, no reason to block this PR on this question. Even if we do eventually migrate it to endo, we're a lot better off having first separated it at all.

Exactly.

> There is an incredible amount of boilerplate that you've carefully gotten right here. Were I to create a separate package in the future, I'd cargo cult. Would it makes sense to create a pristine copy of only the generic bits of this as a template that one could start from, without accidentally including the parts you meant to be specific to this particular package?

It would have been nice to have that when making this, but IME such templates get stale and can be more misleading than informative. Better to find a minimal thing to copy from, as I did here from `@agoric/asserts`. I expect this simple package will be informative for others.

turadg:
[This failure](https://github.com/Agoric/agoric-sdk/runs/7925594193?check_suite_focus=true) appears to be because `src/` isn't exported from `package.json`. Most other packages do that but I thought we could get away from it here.

I'll let the automerge continue since it's so slow. It's a non-blocking test so I can follow up later. I'll try to see if I can make the build step produce an artifact that doesn't require exporting `src/`, which will help with https://github.com/Agoric/agoric-sdk/issues/5808

mhofman:
The getting started check needs to be green to land as part of mergify rules. For some reason the failure is holding up the queue, i will remove from the automerge

packages/internal/src/types.js
```diff
@@ -0,0 +1,5 @@
+// @ts-check
```
mhofman: What is the purpose of this file? Just a placeholder for future exported types not defined elsewhere? In that case it should be re-exported from the index

packages/internal/exported.js
```diff
@@ -0,0 +1 @@
+import './src/types.js';
```
mhofman: As far a I remember, exported.js is an artifact of modules with ambient types and has no benefits for explicitly exported types

packages/internal/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@endo/eventual-send": "^0.15.5",
+    "@endo/promise-kit": "^0.2.43"
+  },
+  "devDependencies": {
+    "@endo/init": "^0.5.43",
+    "ava": "^4.3.1"
+  },
+  "author": "Agoric",
+  "license": "Apache-2.0",
+  "files": [
```
mhofman, line 30: Need to include the src folder for publishing.

Consider using an `exports` field if you want to restrict deep imports.

mhofman (changes requested):

packages/internal/package.json
turadg: Yeah that [was my plan](https://github.com/Agoric/agoric-sdk/pull/6001#issuecomment-1221097283) but I ran into trouble (documented now in README.md).

I see @kriskowal used exports here but I think it works because the export paths are the same as the filenames, so the tooling that just looks at the filesystem happens to succeed:
https://github.com/Agoric/agoric-sdk/blob/02f2002a41a8509ee2b59c943535dbba73f5bb61/packages/cosmic-proto/package.json#L17-L22

packages/internal/exported.js
turadg: great, removed

packages/internal/package.json
mhofman: You should be able to have a single exports for the main entrypoint for now:

```json
  "exports": {
    ".": "./src/index.js",
    "./package.json": "./package.json"
  }
```

turadg: With that there are errors like,
```
Cannot find module '@agoric/internal' or its corresponding type declarations.
```

Btw, why would we need to export package.json?

turadg: Separately, do you still have blocking request changes?

mhofman: Really? That's strange. The package.json I believe is necessary for the Endo bundler to work correctly. @kriskowal has the most experience on the exports field.

mhofman (approved):
If it passes CI let's land this.

 I think this package would be a good candidate to figure out best practices for exports and types in a follow up.

turadg:
> If it passes CI let's land this.

It [passed "Chain deployment test"](https://github.com/Agoric/agoric-sdk/runs/7932538322?check_suite_focus=true) with 5f27989 . That didn't land due to two cancelled checks, so I rebased on master and it ran again. Now it's [failing Chain deployment](https://github.com/Agoric/agoric-sdk/runs/7939642054?check_suite_focus=true) and I don't see why.

PR #6002, opened 2022-08-19
chore: ground floating promises in swingset unit tests

warner:
to hush the `@typescript-eslint/no-floating-promises` line warnings


packages/SwingSet/test/test-vat-timer.js
```diff
@@ -9,6 +9,13 @@ import { buildRootObject, debugTools } from '../src/vats/timer/vat-timer.js';
 import { TimeMath } from '../src/vats/timer/timeMath.js';
 import { waitUntilQuiescent } from '../src/lib-nodejs/waitUntilQuiescent.js';
 
+function makeGround() {
```
turadg, line 12: this seems hacky to me. looks like each promise being grounded is expected to resolve before the next line. why not simply `await` each?

if there's a good reason, then should `waitUntilQuiescent` be part of grounding?

PR #6012, opened 2022-08-19
fix: align issuer declaration order

erights:
Purely a semantics-free declaration reordering. No text was changed within the reordered pieces.

There are three places where we encounter a sequence that is per issuer methods.
   * The property declarations in the `Issuer` type
   * The preceding declarations of the method types
   * The implementation

https://github.com/Agoric/agoric-sdk/pull/5960 adds a fourth one. To do this, I had to repeatedly look at all three representations method by method. For my sanity I arranged them all to follow the `Issuer` type's order of property declarations. But this creates diff noise in an already busy PR. Hence I'm pulling out the semantics-free reordering by itself, so that once #5960 is rebased on this one, the diff noise will go await as well.

### DIFF
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -466,6 +466,12 @@ export const vivifyPaymentLedger = (
   );
 
   const issuer = vivifySingleton(issuerBaggage, `${name} issuer`, {
+    getBrand: () => brand,
+    getAllegedName: () => name,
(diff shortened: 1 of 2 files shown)


erights:
See https://github.com/Agoric/agoric-sdk/pull/5960#discussion_r950386606

erights:
ping. this one should be easy to review and makes #5960 simpler to review.

Chris-Hibbert (approved):

Issue #6009, opened 2022-08-19
PSM only bootstrap configuration


dckc:
## What is the Problem Being Solved?

We have convenient ways to launch the chain with no contracts or with all Inter Protocol contracts, but no convenient way to run just the PSM.

## Description of the Design

`decentral-pso-config.json`, tweaks to `inter-protocol/proposals` etc.

PSM contract doesn't need to be launched separately from the platform (as in #5819)

## Security Considerations

considerably smaller footprint than all of Inter Protocol

## Test Plan

`make scenario2-run-chain-pso`

arirubinstein:
```
2022-08-23T22:29:39.070Z launch-chain: polled; blockTime:1661293773, h:16; ADDED = false
2022-08-23T22:29:39.072Z block-manager: DELIVER_INBOUND error: (Error#1)
Error#1: error in inboundCallback: (Error#2)

  at Object.deliverInbound (packages/SwingSet/src/devices/mailbox/mailbox.js:188:14)
  at deliverInbound (packages/cosmic-swingset/src/launch-chain.js:312:13)
  at processAction (packages/cosmic-swingset/src/block-manager.js:57:13)
  at blockingSend (packages/cosmic-swingset/src/block-manager.js:232:19)
  at async portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:232:16)

Nested error under Error#1
  Error#2: deliverInboundMessages before registerInboundHandler
  
    at deliverInboundMessages (.../swingset-vat/src/devices/mailbox/device-mailbox.js:38:1)
    at inboundCallback (.../swingset-vat/src/devices/mailbox/device-mailbox.js:25:1)
    at Object.deliverInbound (packages/SwingSet/src/devices/mailbox/mailbox.js:186:22)
    at deliverInbound (packages/cosmic-swingset/src/launch-chain.js:312:13)
    at processAction (packages/cosmic-swingset/src/block-manager.js:57:13)
    at blockingSend (packages/cosmic-swingset/src/block-manager.js:232:19)
    at async portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:232:16)
  
portHandler threw (Error#1)
[...]

PR #6013, opened 2022-08-19
4968 psm econ data

### LINKED ISSUE #4968: Expose key PSM economic data via subscription

Users and economic committee need to know metrics like total value locked.

## Description of the Design

TBD akin to the list in https://github.com/Agoric/agoric-sdk/issues/4649

## Security Considerations
TBD
## Test Plan
TBD

turadg:
closes: #4968 

## Description

See curated commit history.

### Security Considerations

--

Updated README

### Testing Considerations

Added unit tests. To cover `startPSM` I also tested manually with cosmic-swingset chain,

```
❯ make scenario2-setup scenario2-run-chain-economy

❯ agd query vstorage keys published.psm.IST.AUSD
children:
- governance
- metrics
pagination: null

❯ agoric follow :published.psm.IST.AUSD.metrics
{
  anchorPoolBalance: {
    brand: slot(0,"Alleged: AUSD brand"),
    value: 0n,
  },
  feePoolBalance: {
    brand: slot(1,"Alleged: IST brand"),
    value: 0n,
  },
  totalAnchorProvided: {
    brand: slot(0),
    value: 0n,
  },
  totalStableProvided: {
    brand: slot(1),
    value: 0n,
  },
}

### DIFF
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -15,6 +15,7 @@ import { AmountMath } from '@agoric/ertp';
 import { provide } from '@agoric/store';
 
 import { makeMakeCollectFeesInvitation } from '../collectFees.js';
+import { makeMetricsPublishKit } from '../contractSupport.js';
 
 const { details: X } = assert;
 
@@ -25,6 +26,17 @@ const { details: X } = assert;
  * economic policies, the fee percentage for trading into and out of the stable
  * token are specified separately.
  *
+ */
+
+/**
+ * @typedef {object} MetricsNotification
+ * Metrics naming scheme is that nouns are present values and past-participles are accumulative.
+ *
+ * @property {Amount<'nat'>} anchorPoolBalance  amount of Anchor token available to be swapped
+ * @property {Amount<'nat'>} feePoolBalance     amount of Stable token fees available to be collected
+ *
+ * @property {Amount<'nat'>} totalAnchorProvided  running sum of Anchor ever given by this contract
+ * @property {Amount<'nat'>} totalStableProvided  running sum of Stable ever given by this contract
  */
 
 /**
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: packages/inter-protocol/src/psm/psm.js [hunk 1]
```diff
@@ -97,6 +98,20 @@ export const start = async (zcf, privateArgs, baggage) => {
...
 
+  // Metrics
+  const { metricsPublication, metricsSubscription } = makeMetricsPublisherKit(
+    privateArgs.storageNode,
+    privateArgs.marshaller,
+  );
+  const updateMetrics = () => {
+    metricsPublication.updateState(
+      harden({
+        poolBalance: anchorPool.getAmountAllocated('Anchor', anchorBrand),
```
Chris-Hibbert: That's all we need to recover the state, but the econ committee will also want to know about flows through the system.  We should track the amounts that passed through `giveStable` and `wantStable` as well.

Chris-Hibbert (dismissed):
I didn't look closely at the test driver. I'll submit this request while I'm looking that over.

dtribble:
Everything up to the test file lgtm

dckc (commented):
Code looks good (by-commit organization reads nicely).

Re design/requirements #4639 doesn't mention the PSM, and #6009 still says "TBD". I wonder if @dtribble was reviewing from that angle. The example output in the PR description is quite handy, meanwhile.

packages/inter-protocol/src/psm/psm.js
dckc: @rowgraus @dtribble @btulloh it would be good to get your input on this.

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -147,186 +140,289 @@ test.before(async t => {
...
-  const seat1 = await E(zoe).offer(
-    E(publicFacet).makeSwapInvitation(),
-    harden({ give: { In: giveAnchor } }),
-    harden({ In: anchorMint.mintPayment(giveAnchor) }),
+    harden({ AUSD: anchor.issuer }),
+    { ...terms, ...customTerms },
+    harden({
+      feeMintAccess,
+      initialPoserInvitation,
+      storageNode: mockChainStorage.makeChildNode('thisPsm'),
```
turadg, line 168: in `startPSM` the node path is `psm.<stable>.<anchor>`. There was no no unit test of `startPSM` and this PR doesn't add one. So the test plan includes manually testing on-chain, which is fuller coverage.

dckc (approved):
with input from @btulloh , this seems good to go

packages/inter-protocol/test/psm/test-psm.js
dckc: Yeah; that's nice...

It prompted me to wonder about other places where our code might assume _the_ PSM, as if there were only one. For example, `E(home.agoricNames).lookup('instance', 'psm')`.

PR #6016, opened 2022-08-21
fix: make nested awaits somewhat safer

erights:
for many unsafe nested await expressions like 
```js
await expr;
```
change to
```js
await (async () => (expr))();
```
making it safer against some of the hazards of `await`
   * if `expr` throws, it'll still be a turn boundary before the `await` turns it back into a throw. When the nature of the unsafety was only that it was within a try block, this change actually makes that use safe. The catch, finally, and code after try will always happen in another turn.
   * make the semantic cost visible locally, and make the semantic cost carry a syntactic cost locally
   * leave the global lint warning in place, without habituating us to ignoring it.

Note: These are *not* semantics preserving changes. They may very well break something. This PR needs to be reviewed carefully before merging. Ideally with the author of each `await` reviewing the changes I made to their code.

Note: Not all packages done yet. In fact, at the moment I've only reduced the global await warnings from 191 to 136 :(

### DIFF
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -25,8 +25,10 @@ const { keys, values, fromEntries } = Object;
  * @returns {Promise<Record<string, V>>}
  * @template V
  */
-const allValues = async obj =>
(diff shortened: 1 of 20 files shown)


erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/6219

PR #6020, opened 2022-08-22
feat: PSM-only bootstrap

### LINKED ISSUE #6009: PSM only bootstrap configuration

dckc:
closes: #6009

## Description

`boot-psm.js` is an alternative to `packages/vats/src/core/boot.js` that runs all and only the bootstrap behaviors necessary to start the PSM, along with the smart wallet.

 - [x] TODO: `decentral-psm-config.json`

### Security Considerations

In order to simplify security review, where `boot.js` does dynamic function name lookup from `manifest.js` and simulates CORE_EVAL proposals, `boot-psm.js` makes statically-evident imports and function calls.

 - [x] TODO: attenuate `allPowers` on a per-call basis
 - [x] TODO: factor startEconomicCommittee out of econ-behaviors.js

?

`decentral-psm-config.json` should provide one form of documentation.

 - [ ] TODO: manifest visualization
 - visualization of slogfile
    - [x]  from running bootstrap
    - [ ] doing a trade

### Testing Considerations

 - [x] minimal unit testing
 - [x] integration testing with `local-chain` (`make scenario2-run-chain-psm`) and smart wallet (`psm-tool.js`)
 - [ ] test that important parts of `CORE_EVAL` were not left out somehow

### DIFF
--- a/packages/inter-protocol/src/proposals/startPSM.js
+++ b/packages/inter-protocol/src/proposals/startPSM.js
@@ -7,6 +7,7 @@ import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
 import { E } from '@endo/far';
 import { Stable } from '@agoric/vats/src/tokens.js';
 import { deeplyFulfilledObject } from '@agoric/internal';
+import { assert } from '@agoric/assert';
 import { reserveThenGetNamePaths } from './utils.js';
 
 const BASIS_POINTS = 10000n;
@@ -214,6 +215,31 @@ export const makeAnchorAsset = async (
   ]);
 };
 harden(makeAnchorAsset);
+
+/** @param {BootstrapSpace & { devices: { vatAdmin: any }, vatPowers: { D: DProxy }, }} powers */
+export const installGovAndPSMContracts = async ({
+  vatPowers: { D },
+  devices: { vatAdmin },
+  consume: { zoe },
+  installation: {
+    produce: { contractGovernor, committee, binaryVoteCounter, psm },
+  },
+}) => {
+  return Promise.all(
+    Object.entries({
+      contractGovernor,
+      committee,
+      binaryVoteCounter,
+      psm,
+    }).map(async ([name, producer]) => {
+      const bundleCap = D(vatAdmin).getNamedBundleCap(name);
+      const bundle = D(bundleCap).getBundle();
+      const installation = E(zoe).install(bundle);
+
+      producer.resolve(installation);
+    }),
+  );
+};
 
 export const PSM_MANIFEST = harden({
   [makeAnchorAsset.name]: {
--- a/packages/vats/decentral-psm-config.json
+++ b/packages/vats/decentral-psm-config.json
@@ -1,0 +1,56 @@
+{
+  "bootstrap": "bootstrap",
+  "defaultReapInterval": 1000,
+  "vats": {
+    "bootstrap": {
+      "sourceSpec": "@agoric/vats/src/core/boot-psm.js",
+      "parameters": {
+        "anchorAssets": [
+          {
+            "denom": "ibc/usdc1234"
+          }
+        ],
+        "economicCommitteeAddresses": []
+      }
+    }
+  },
+  "bundles": {
+    "walletFactory": {
+      "sourceSpec": "@agoric/wallet/contract/src/walletFactory.js"
+    },
+    "committee": {
+      "sourceSpec": "@agoric/governance/src/committee.js"
+    },
+    "contractGovernor": {
+      "sourceSpec": "@agoric/governance/src/contractGovernor.js"
+    },
+    "binaryVoteCounter": {
+      "sourceSpec": "@agoric/governance/src/binaryVoteCounter.js"
+    },
+    "psm": {
+      "sourceSpec": "@agoric/inter-protocol/src/psm/psm.js"
+    },
+    "bank": {
+      "sourceSpec": "@agoric/vats/src/vat-bank.js"
+    },
+    "centralSupply": {
+      "sourceSpec": "@agoric/vats/src/centralSupply.js"
+    },
+    "mintHolder": {
+      "sourceSpec": "@agoric/vats/src/mintHolder.js"
+    },
+    "board": {
+      "sourceSpec": "@agoric/vats/src/vat-board.js"
+    },
+    "chainStorage": {
+      "sourceSpec": "@agoric/vats/src/vat-chainStorage.js"
+    },
+    "zcf": {
+      "sourceSpec": "@agoric/zoe/contractFacet.js"
+    },
+    "zoe": {
+      "sourceSpec": "@agoric/vats/src/vat-zoe.js"
+    }
+  },
+  "defaultManagerType": "xs-worker"
+}
--- a/packages/vats/src/core/basic-behaviors.js
+++ b/packages/vats/src/core/basic-behaviors.js
@@ -22,7 +22,7 @@ const RESERVE_ADDRESS = 'agoric1ae0lmtzlgrcnla9xjkpaarq5d5dfez63h3nucl';
  * non-exhaustive list of powerFlags
  * REMOTE_WALLET is currently a default.
  */
-const PowerFlags = /** @type {const} */ ({
+export const PowerFlags = /** @type {const} */ ({
   SMART_WALLET: 'SMART_WALLET',
   /** The ag-solo wallet is remote. */
   REMOTE_WALLET: 'REMOTE_WALLET',
@@ -174,6 +174,28 @@ export const makeBoard = async ({
 harden(makeBoard);
 
 /**
+ * @param {NameAdmin} namesByAddressAdmin
+ * @param {string} address
+ */
+export const makeMyAddressNameAdmin = (namesByAddressAdmin, address) => {
+  // Create a name hub for this address.
+  const { nameHub: myAddressNameHub, nameAdmin: rawMyAddressNameAdmin } =
+    makeNameHubKit();
+
+  /** @type {MyAddressNameAdmin} */
+  const myAddressNameAdmin = Far('myAddressNameAdmin', {
+    ...rawMyAddressNameAdmin,
+    getMyAddress: () => address,
+  });
+  // reserve space for deposit facet
+  myAddressNameAdmin.reserve('depositFacet');
+  // Register it with the namesByAddress hub.
+  namesByAddressAdmin.update(address, myAddressNameHub, myAddressNameAdmin);
+
+  return myAddressNameAdmin;
+};
+
+/**
  * Make the agoricNames, namesByAddress name hierarchies.
  *
  * agoricNames are well-known items such as the IST issuer,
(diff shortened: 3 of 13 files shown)


dckc:
## account provisioning with out mailbox/ag-solo?

Integration testing showed that `published.wallet.*` doesn't show up in this draft. `makeClientBanks` arranges for one part of that when it starts `walletFactory`, but the other part is in `bridgeProvisioner`. I wasn't planning to do `bridgeProvisioner`, since that's tied up with mailbox/ag-solo. Then we need a different way to to `E(bridgeManager).register(BRIDGE_ID.PROVISION, ...)`

dckc:
visualization: https://gist.github.com/dckc/7da293f2fcf10d1a4853dc680ea3f667#file-slog-svg

dckc:
@Chris-Hibbert it's finally green! Please take a look.

I rebased to get the new endo; a quick local test suggests that doesn't introduce any problems; my fingers are crossed that the ci bots concur.

dckc:
Perhaps it should be a separate PR, but this includes a fix for #6053.

dckc:
### how to get diagnostic for failure to get invitation?

The wallet state doesn't change in case getting an invitation failed, as in when I was using an out-of-date method:

```
2022-08-27T18:35:20.515Z SwingSet: ls: v7: RemoteTypeError(error:liveSlots:v14#70001)#1: target has  no method "makeSwapInvitation", has ["getAmount","getBrand","getContractGovernor","getGovernedParams","getInstallation","getInstance","getInvitationAmount","getMetrics","getNat","getPoolBalance","getRatio","getString","getSubscription","getUnknown","makeGiveStableInvitation","makeWantStableInvitation"]
```

packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -215,6 +216,31 @@ export const makeAnchorAsset = async (
 };
 harden(makeAnchorAsset);
 
+/** @param {BootstrapSpace & { devices: { vatAdmin: any }, vatPowers: { D: DProxy }, }} powers */
```
turadg, line 219: `SwingsetVats` has `vatAdmin` as `VatAdminVat` which appears to match how it's being used here.

packages/vats/decentral-psm-config.json
```diff
@@ -0,0 +1,56 @@
+{
+  "bootstrap": "bootstrap",
+  "defaultReapInterval": 1000,
+  "vats": {
+    "bootstrap": {
+      "sourceSpec": "@agoric/vats/src/core/boot-psm.js",
+      "parameters": {
+        "anchorAssets": [
+          {
+            "denom": "ibc/usdc1234"
```
turadg, line 10: what is this value? might be worth a RAEDME.md in this dir to explain.

looking at how this gets loaded, must it be JSON or could we use `.js` to get comments? https://github.com/Agoric/agoric-sdk/blob/eb5b1f5b0635680081d708e1928c8ee089b61c80/packages/cosmic-swingset/src/chain-main.js#L361-L365

packages/vats/src/core/basic-behaviors.js
```diff
@@ -22,7 +22,7 @@ const RESERVE_ADDRESS = 'agoric1ae0lmtzlgrcnla9xjkpaarq5d5dfez63h3nucl';
  * non-exhaustive list of powerFlags
  * REMOTE_WALLET is currently a default.
  */
-const PowerFlags = /** @type {const} */ ({
+export const PowerFlags = /** @type {const} */ ({
```
turadg, line 27: this is a funny name in this context. Worth a comment explaining the term comes from the Golang code. https://github.com/Agoric/agoric-sdk/blob/eb5b1f5b0635680081d708e1928c8ee089b61c80/packages/cosmic-proto/gen/agoric/swingset/msgs.ts#L50 ?

packages/vats/src/core/boot-psm.js
```diff
@@ -0,0 +1,223 @@
...
+ *     economicCommitteeAddresses: string[],
+ *     anchorAssets: { denom: string }[],
+ * }} vatParameters
+ */
+export const buildRootObject = (vatPowers, vatParameters) => {
+  // @ts-expect-error no TS defs for rickety test scaffolding
+  const log = vatPowers.logger || console.info;
+
+  const {
+    anchorAssets: [{ denom: anchorDenom }], // TODO: handle >1?
```
turadg: what does "handle >1" mean?

packages/vats/src/core/boot-psm.js
```diff
@@ -0,0 +1,223 @@
...
+          },
+        },
+      }),
+      makeAnchorAsset(powersFor('makeAnchorAsset'), {
+        options: { anchorOptions: { denom: anchorDenom } },
+      }),
+      startPSM(powersFor('startPSM'), {
+        options: { anchorOptions: { denom: anchorDenom } },
+      }),
+      // Finally, allow bootstrap powers to be granted
```
turadg: why "Finally"? don't all these execute in parallel?

packages/vats/src/core/startWalletFactory.js
```diff
@@ -0,0 +1,135 @@
...
+
+const { details: X } = assert;
+
+/**
+ * @param {ERef<ZoeService>} zoe
+ * @param {Installation<import('@agoric/smart-wallet/src/walletFactory').start>} inst
+ * @typedef {Awaited<ReturnType<typeof startFactoryInstance>>} WalletFactoryStartResult
+ */
+// eslint-disable-next-line no-unused-vars
+const startFactoryInstance = (zoe, inst) => E(zoe).startInstance(inst);
```
turadg, line 17:
```suggestion
/**
 * @typedef {Awaited<ReturnType<import('@agoric/smart-wallet/src/walletFactory').start>>} WalletFactoryStartResult
 */
```

turadg: Though why define that type? The start result only has a creator facet so you could put that in the space instead of https://github.com/Agoric/agoric-sdk/blob/eb5b1f5b0635680081d708e1928c8ee089b61c80/packages/vats/src/core/types.js#L218-L220

packages/vats/src/core/startWalletFactory.js
```diff
@@ -0,0 +1,135 @@
...
+  ]);
+
+  const terms = await deeplyFulfilled(
+    harden({
+      agoricNames,
+      namesByAddress,
+      board,
+    }),
+  );
+  /** @type {WalletFactoryStartResult} */
```
turadg, line 56: doesn't hurt but not necessary. `startInstance` now carries forward the type of the installation.
```suggestion
```

packages/vats/src/core/startWalletFactory.js
```diff
@@ -0,0 +1,135 @@
...
+        myAddressNameAdmin,
+      );
+      console.info('provisioned', address, powerFlags);
+    },
+  });
+  if (!bridgeManager) {
+    console.warn('missing bridgeManager in startWalletFactory');
+  }
+  await (bridgeManager &&
+    E(bridgeManager).register(BRIDGE_ID.PROVISION, handler));
```
turadg, line 104: can we require it?

packages/vats/src/core/boot-psm.js
dckc: Handle more than one anchor asset; that is: start more than 1 PSM contract, each with its own anchor "denom".

packages/vats/src/core/boot-psm.js
dckc: Yes, they do. Perhaps "Additionally" would be better? Perhaps I should have documented the part where this list contains *only* the things necessary to run the PSM and to do smart wallet transactions.

packages/vats/src/core/startWalletFactory.js
dckc: The result from `E(zoe).startInstance()` also has an admin facet, which is useful for upgrade. I suppose the motivation for `startFactoryInstance` was under-documented.

packages/vats/src/core/startWalletFactory.js
dckc: Not straightforwardly. `bridgeManager` has been optional for quite a while, and it would take some doing to change all the places that rely on that.

packages/vats/src/core/boot-psm.js
```diff
@@ -0,0 +1,223 @@
...
+const PSM_GOV_INSTALL_MANIFEST = {
+  [installGovAndPSMContracts.name]: {
+    vatPowers: { D: true },
+    devices: { vatAdmin: true },
+    consume: { zoe: 'zoe' },
+    installation: {
+      produce: {
+        contractGovernor: 'zoe',
+        committee: 'zoe',
+        binaryVoteCounter: 'zoe',
```
Chris-Hibbert, line 53: If the strings are only for labelling the graph, can they all be 'gov' rather than 'zoe'?

Chris-Hibbert (approved):

packages/vats/src/core/boot-psm.js
dckc: installations all live in the zoe vat, IIUC

packages/inter-protocol/src/proposals/startPSM.js
dckc: I don't think `vatAdminVat` is the same as the vatAdmin device... in fact, I tested it:

```
src/proposals/startPSM.js:235:27 - error TS2345: Argument of type 'VatAdminVat' is not assignable to parameter of type 'Device<any>'.
  Type '{ createVatAdminService: (vaDevice: any) => { waitForBundleCap(bundleID: any): any; getBundleCap(bundleID: any): any; getNamedBundleCap(name: any): any; getBundleIDByName(name: any): any; createMeter(remaining: any, threshold: any): KindFacet<...>; createUnlimitedMeter(_context: any): KindFacet<...>; createVat(bundl...' is not assignable to type 'Device<any>'.
    Type '{ createVatAdminService: (vaDevice: any) => { waitForBundleCap(bundleID: any): any; getBundleCap(bundleID: any): any; getNamedBundleCap(name: any): any; getBundleIDByName(name: any): any; createMeter(remaining: any, threshold: any): KindFacet<...>; createUnlimitedMeter(_context: any): KindFacet<...>; createVat(bundl...' is not assignable to type '"Device"'.

235       const bundleCap = D(vatAdmin).getNamedBundleCap(name);
```

packages/vats/decentral-psm-config.json
dckc: I think it has to be JSON.

It's a **denom** in the sense of Cosmos-SDK [ADR 024: Coin Metadata](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-024-coin-metadata.md) for the anchor/reference asset that a PSM trades for IST.

See also [makeAnchorAsset](https://github.com/Agoric/agoric-sdk/blob/ba1ed62062a63862e2eecb598b0bd1d2ac828e1f/packages/inter-protocol/src/proposals/startPSM.js#L159-L166) and [bankManager.addAsset](https://github.com/Agoric/agoric-sdk/blob/ba1ed62062a63862e2eecb598b0bd1d2ac828e1f/packages/vats/src/vat-bank.js#L320-L328)

packages/vats/src/core/boot-psm.js
dckc: It was just about as easy to code it up as answer this, so I did.

dckc:
let's leave the chainstorage fix to its own PR after all: #6068

PR #6029, opened 2022-08-22
fix(swingset): don't un-expose 'gc', to stop intermittent test failures

### LINKED ISSUE #6028: intermittent ReferenceError: gc is not defined during CI
## Describe the bug

@erights noticed an unusual CI failure (in https://github.com/Agoric/agoric-sdk/runs/7959549359?check_suite_focus=true#step:5:12) : 

```
$ SWINGSET_WORKER_TYPE=xs-worker ava -c 2 'test/swingsetTests/**/test-*.js'

  Uncaught exception in test/swingsetTests/brokenContracts/test-crashingContract.js

  ReferenceError: gc is not defined
    at evalmachine.<anonymous>:1:1
    at Script.runInContext (vm.js:144:12)
    at Script.runInNewContext (vm.js:149:17)
    at Object.runInNewContext (vm.js:303:38)
    at packages/SwingSet/src/lib-nodejs/engine-gc.js:9:15

  › evalmachine.<anonymous>:1:1
  › packages/SwingSet/src/lib-nodejs/engine-gc.js:9:15

  ✖ test/swingsetTests/brokenContracts/test-crashingContract.js exited with a non-zero exit code: 1
```

@michaelfig was able to reproduce it locally (occasionally), and recommends we change `SwingSet/src/lib-nodejs/engine-gc.js`, which currently does:

```js
import v8 from 'v8';
import vm from 'vm';

/* global globalThis */
let bestGC = globalThis.gc;
if (typeof bestGC !== 'function') {
  // Node.js v8 wizardry.
  v8.setFlagsFromString('--expose_gc');
  bestGC = vm.runInNewContext('gc');
  // Hide the gc global from new contexts/workers.
  v8.setFlagsFromString('--no-expose_gc');
}

// Export a const.
const engineGC = bestGC;
export default engineGC;
```

to remove the cleanup step that does `--no-expose_gc`.

I don't have a theory about why this might be causing the problem, but his recommended fix sounds good to me. He points out that SES already strips unknown properties from `globalThis`, so there isn't a danger of accidentally exposing `gc` to any SES environments.

warner:
engine-gc.js acquires the `gc()` function (to force garbage
collection) on Node.js by turning on the (global) `--expose_gc` V8
flag, then evaluating code within a new context (because the original
one might not have had `--expose_gc` turned on), stashing the `gc`
global, then uses `--no-expose_gc` to hide it again.

It seems that AVA's worker-based parallelism approach causes multiple
calls to this routine within the same process, sharing a V8
instance. So there's a race between one worker doing
enable/stash/disable and another. If they interleave, a "stash" might
happen while the symbol is disabled, causing a Reference Error and a
crash.

closes #6028

### DIFF
--- a/packages/SwingSet/src/lib-nodejs/engine-gc.js
+++ b/packages/SwingSet/src/lib-nodejs/engine-gc.js
@@ -7,8 +7,13 @@ if (typeof bestGC !== 'function') {
   // Node.js v8 wizardry.
   v8.setFlagsFromString('--expose_gc');
   bestGC = vm.runInNewContext('gc');
-  // Hide the gc global from new contexts/workers.
-  v8.setFlagsFromString('--no-expose_gc');
+  // We leave --expose_gc turned on, otherwise AVA's shared workers
+  // may race and disable it before we manage to extract the
+  // binding. This won't cause 'gc' to be visible to new Compartments
+  // because SES strips out everything it doesn't recognize.
+
+  // // Hide the gc global from new contexts/workers.
+  // v8.setFlagsFromString('--no-expose_gc');
 }
 
 // Export a const.


packages/SwingSet/src/lib-nodejs/engine-gc.js
```diff
@@ -7,8 +7,11 @@ if (typeof bestGC !== 'function') {
   // Node.js v8 wizardry.
   v8.setFlagsFromString('--expose_gc');
   bestGC = vm.runInNewContext('gc');
-  // Hide the gc global from new contexts/workers.
-  v8.setFlagsFromString('--no-expose_gc');
+  // we leave --expose_gc turned on, otherwise AVA's shared workers
+  // may race and disable it before we manage to extract the symbol
```
michaelfig:
```suggestion
  // may race and disable it before we manage to extract the binding
```

michaelfig (approved):
Just minor terminology ("symbol" -> "binding"), otherwise LGTM!

warner:
I figured out that the CI failure was another instance of #5575 (and I have an idea on how to deal with that), so I'll just re-kick the CI.

packages/SwingSet/src/lib-nodejs/engine-gc.js
```diff
@@ -7,8 +7,11 @@ if (typeof bestGC !== 'function') {
   // Node.js v8 wizardry.
   v8.setFlagsFromString('--expose_gc');
   bestGC = vm.runInNewContext('gc');
-  // Hide the gc global from new contexts/workers.
-  v8.setFlagsFromString('--no-expose_gc');
+  // we leave --expose_gc turned on, otherwise AVA's shared workers
+  // may race and disable it before we manage to extract the binding
```
mhofman: Should we leave a comment that it's safe since at worse it would just leave a global `gc` in new realms (which we don't create in non-test paths), and that would get removed in Compartments anyway.

PR #6031, opened 2022-08-23
ci(mergify): don't rely on local-npm integration to pass

michaelfig:
Refs: #5970

## Description

At least until we have made our NPM release story more coherent, tolerate failures of the `local-npm` integration test so that they don't stall Mergify.

We don't recommend using `bypass:integration` for many of these cases since that bypasses valuable deployment and `link-cli` integration testing.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,15 +12,16 @@ queue_rules:
       - or:
           - label=bypass:integration
           - check-skipped=getting-started
-          - and:
-              - or:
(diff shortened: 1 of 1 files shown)


mhofman (approved):
I'm concerned about disabling this test since it caught an issue in https://github.com/Agoric/agoric-sdk/pull/6001. I understand we don't officially support the NPM path yet though, so approving.

mhofman (approved):

PR #6035, opened 2022-08-23
chore: Sync Endo versions

kriskowal:
Per Endo release https://github.com/endojs/endo/pull/1263, this change brings the latest versions from Endo into Agoric SDK, eliminating the current generation of Endo patches.


yarn.lock
```diff
@@ -15634,7 +15634,7 @@ rollup@^2.43.1, rollup@^2.58.0:
   optionalDependencies:
     fsevents "~2.3.2"
 
-rollup@endojs/endo#rollup-2.7.1-patch-1:
+"rollup@github:endojs/endo#rollup-2.7.1-patch-1":
```
turadg: seems like some different Yarn versions in play. is yours `1.22.19`?

turadg (approved):
patches 🧹  🔥

yarn.lock
kriskowal: Indeed, I’m on `1.22.19` and Node.js `14.19.0`. 🧐

kriskowal:
Looks like I need to port forward the SES-ava patch https://github.com/Agoric/agoric-sdk/runs/7980146008?check_suite_focus=true

mhofman:
@mergifyio refresh

Issue #6007, opened 2022-08-19
pause one side of PSM as econ committee


dckc:
## What is the Problem Being Solved?

The econ committee should be able to pause giveStable and wantStable invitations to the PSM separately.

## Description of the Design

seems to need

 - [x] #6005
 - [x] #5796 

## Test Plan

 - [X] unit test?
 - [x] e2e test of a vote to pause one sort of invitation

cc @dtribble @Chris-Hibbert @rowgraus

dckc:
Leaving the end-to-end test / user story out of scope here is interfering with planning efforts. This is the motivation for #6108 .

dckc:
we did an e2e test today!

PR #6036, opened 2022-08-23
store econ committee questions off-chain

turadg:
## Description

Publish committee questions to chain storage.

### Security Considerations

Gives contracts a Marshaller from the board. It is read-only.

It's a feature of `@agoric/governance` contracts but also `@agoric/inter-protocol`. Documented in both READMEs.

### Testing Considerations

Unit and swingset tests pass, but they don't cover `startEconomicCommittee`. So I did a chain test per the "Demo" section of the README. The contract publishes 'undefined' as the first question so the key is present:
```
❯ agd query vstorage keys 'published.committees'
children:
- Initial_Economic_Committee
pagination: null
❯ agd query vstorage keys 'published.committees.Initial_Economic_Committee'
children:
- latestQuestion
pagination: null
❯ agoric follow :published.committees.Initial_Economic_Committee.latestQuestion
undefined
```

### DIFF
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -6,7 +6,10 @@ import { AmountMath } from '@agoric/ertp';
 import '@agoric/governance/exported.js';
 import '@agoric/vats/exported.js';
 import '@agoric/vats/src/core/types.js';
-import { makeStorageNodeChild } from '@agoric/vats/src/lib-chainStorage.js';
+import {
+  assertPathSegment,
+  makeStorageNodeChild,
+} from '@agoric/vats/src/lib-chainStorage.js';
 import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
 import { E, Far } from '@endo/far';
 import { Stable, Stake } from '@agoric/vats/src/tokens.js';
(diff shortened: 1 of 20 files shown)


Chris-Hibbert (commented):
I don't see any test output changes showing evidence that questions were ever recorded. Did I miss it somehow?  I'm sure there are tests in which questions are asked.

dckc:
A test showing a question to the econ committee to pause one side of the PSM would be most welcome!

Chris-Hibbert:
> A test showing a question to the econ committee to pause one side of the PSM would be most welcome!

Well, it turns out that in service of #6007, I'm working on a test showing a governed PSM setting the filter.  The existing tests work with a bare PSM that doesn't actually have governance installed AFAICT.   (I didn't know that was possible!)

turadg:
> I don't see any test output changes showing evidence that questions were ever recorded. Did I miss it somehow?

Nope. I tried to call this state out in Testing Plan.

> I'm sure there are tests in which questions are asked.

If that's sufficient: e91c30161a2a8b7ad2396c4146092a7f27b628b5

Chris-Hibbert (approved):
> > I'm sure there are tests in which questions are asked.

> If that's sufficient: https://github.com/Agoric/agoric-sdk/commit/e91c30161a2a8b7ad2396c4146092a7f27b628b5

That's what I was looking for, thanks.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -95,16 +95,25 @@ export const startEconomicCommittee = async (
   },
   { options: { econCommitteeOptions = {} } = {} },
 ) => {
+  const STORAGE_PATH = 'committee';
```
dckc: This suggests it's the only committee. That's true to some extent, but I wonder to what extent we should commit to that. What do you think about `published.committee.economicGovernance` or just `published.committee1`?

turadg: Good call. Updated to be `published.committees.Initial_Economic_Committee`, sanitizing the `committeeName` property.

turadg:
@Mergifyio requeue

PR #6037, opened 2022-08-23
5796 govern pause

### LINKED ISSUE #5796: Governance for Zcf contract pause

#5749 added the ability to pause processing of invitations. Currently that functionality is accessible via zcf. This is minimally legible, in that access to the functionality is visible, but each contract would have to individually manage updating the list.

## Description of the Design

The zcf function should be handed off to governance, and general support should be added to the governance API so the electorate that governs the contract can shut things down when necessary.

## Security Considerations

Standard legibility and POLA concerns.

## Test Plan

TBD

Chris-Hibbert:
closes: #5796
refs: #6007

## Description

Add the ability for governance to manage 'Pause's for Zoe contracts. 

### Security Considerations

This gives the ability to pause some invitations to whomever has the governance facet.
This doesn't prevent contracts from calling `zcf.setOfferFilter()` directly, but it will be legible when they can.

implications for Zoe and governance.  Zoe gets `setOfferFilter()` and `getOfferFilter()`. Governance adds `voteOnOfferFilter()`.

### Testing Considerations

Added a test. I didn't actually test setting a filter via governance, and blocking actual offers, but those two parts are tested  separately here and in the recent Zoe change to add filters.


packages/governance/src/contractGovernor.js
```diff
@@ -90,10 +91,15 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
  * up to the contract.
  *
  * The contractGovernor creatorFacet includes voteOnParamChanges(),
- * which is used to create questions that will automatically update
- * contract parameters if passed. This facet will usually be closely held. The
- * creatorFacet can also be used to retrieve the governed instance, publicFacet,
- * and its creatorFacet with voteOnParamChanges() omitted.
+ * voteOnFilter, and voteOnApiInvocation. voteOnParamChanges() is used to create
```
turadg: inconsistent use of '()'. Consider dropping the parens and using backticks.
> The contractGovernor creatorFacet includes `voteOnParamChanges`, `voteOnFilter`,
> and `voteOnApiInvocation`. `voteOnParamChanges` is used to create questions that

packages/governance/src/contractGovernor.js
```diff
@@ -90,10 +91,15 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
...
+ * questions that will automatically update contract parameters if passed.
+ * voteOnFilter can be used to crate questions that will prevent the exercise of
```
turadg: s/crate/create

turadg: the questions don't prevent the exercise, right? consider something like "create questions that when affirmed will prevent…"

packages/governance/src/contractGovernor.js
turadg: alternately, refer to them subsequently by the unique string:
inconsistent use of '()'. Consider dropping the parens and using backticks.
> The contractGovernor creatorFacet includes `voteOnParamChanges`, `voteOnFilter`,
> and `voteOnApiInvocation`. ParamChanges is used to create questions that

packages/governance/src/question.js
```diff
@@ -88,6 +90,16 @@ const assertApiInvocation = issue => {
   );
 };
 
+/**
+ * @param {unknown} issue
+ * @returns {asserts issue is OfferFilterIssue}
+ */
+
```
turadg, line 97:
```suggestion
```

packages/governance/src/types.js
```diff
@@ -445,6 +463,16 @@
  * @property {CreateQuestion} createQuestion
  */
 
+/**
+ * @typedef {object} GovernorFacet
+ * @property {() => ParamManagerRetriever} getParamManagerRetriever
```
turadg: No code implements this property.
```sh
❯ git grep getParamManagerRetriever
packages/governance/docs/contractGovernance.puml:        -getParamManagerRetriever()
packages/governance/src/types.js: * @property {() => ParamManagerRetriever} getParamManagerRetriever
```

I think you want `getParamMgrRetriever`. This would have been caught if the new type were used to annotate on object, so that might be worth doing. E.g. at https://github.com/Agoric/agoric-sdk/blob/d6bba43627c64c6a130e5fcb12bda091ae51522b/packages/governance/src/contractHelper.js#L132 

When I add this annotation,
```
    /** @type {import('@agoric/vat-data/src/types.js').FunctionsPlusContext<unknown, GovernorFacet>} */
```
I get a type error for `setOfferFilter`, which I think is correct because it's a virtual object so its first param must be the context (like `getGovernedApiNames` above it):
https://github.com/Agoric/agoric-sdk/blob/d6bba43627c64c6a130e5fcb12bda091ae51522b/packages/governance/src/contractHelper.js#L144-L146

This makes me think that `makeVirtualGovernorFacet` has a bug that no runtime tests are detected. So I'm going to request changes that the type be annotated here and that there be a runtime test that confirms that zcf got valid strings here.

packages/inter-protocol/src/psm/psm.js
```diff
@@ -196,33 +197,52 @@ export const start = async (zcf, privateArgs, baggage) => {
...
 
-  /**
-   * @param {ZCFSeat} seat
-   */
-  const swapHook = seat => {
-    assertProposalShape(seat, {
-      give: { In: null },
-    });
+  /** @param {ZCFSeat} seat */
+  const giveStableHook = seat => {
```
turadg: these changes are in master now. please rebase for a smaller review.

turadg (changes requested):

packages/governance/src/contractGovernor.js
Chris-Hibbert: done

packages/governance/src/question.js
Chris-Hibbert: When I cut the blank line, it complains that the return values don't match. (This happens a few other times in this file.)   Do you have a suggested fix for that?

packages/governance/src/types.js
Chris-Hibbert: I discovered the same thing in the new test of offerFilter for PSM (#6047). It resulted in changing the parameters to `getInvitation` in `makeGovernorFacet` as well.

I added the annotation and also added it for `makeGovernorFacet`.

packages/governance/src/contractGovernor.js
Chris-Hibbert: done

packages/inter-protocol/src/psm/psm.js
Chris-Hibbert: done

packages/governance/src/question.js
turadg: oh that's a neat trick! I've been suppressing the eslint warning but the newline works better and apparently TS still picks it up:
Screen Shot 2022-08-24 at 4 30 02 PM
 
Which is good because the bug I filed got closed apparently as wontfix https://github.com/gajus/eslint-plugin-jsdoc/issues/905

Chris-Hibbert: oh, whew!  😓   I thought I was cheating, and was outing myself.   I discovered this by accident, but exploited it throughout this file.

dckc (commented):
skimmed it; looks good at a glance. I can look more closely if it's important.

packages/governance/src/contractGovernance/governFilter.js
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Make a pair of positions for a question about whether to update the offer
+ * filter. If the vote passes, the list of blocked invitation strings will be
+ * updated.
+ *
+ * @param {string[]} strings
+ */
+const makeOfferFilterPositions = strings => {
+  const positive = harden({ strings });
+  const negative = harden({ dontUpdate: strings });
```
turadg, line 24: the value for key `dontUpdate` suggests some selection of what won't be updated. Consider `changeFilter: false`. If the reader wants to know what would happen were the question resolved positively, they should look at the positive case because that will work independent of the question schema.

packages/governance/src/contractGovernance/governFilter.js
```diff
@@ -0,0 +1,117 @@
...
+    //
+    // * If the vote passed, invoke the API, and return the positive position
+    // * If the vote was negative, return the negative position
+    // * If we can't do either, (the vote failed or the API invocation failed)
+    //   return a broken promise.
+    const outcomeOfUpdate = E(counterPublicFacet)
+      .getOutcome()
+      // @ts-expect-error return types don't appear to match
+      .then(outcome => {
+        if (keyEQ(positive, outcome)) {
```
turadg: nit: I think it reads more conventionally to compare _actual, _expected_.  Just as `t.deepEqual` does and line 88 does.  Consider,
```suggestion
        if (keyEQ(outcome, positive)) {
```

packages/governance/src/contractGovernance/governFilter.js
```diff
@@ -0,0 +1,117 @@
...
+          assert(
+            keyEQ(outcome, harden({ strings })),
```
turadg: since `positive` is `{ strings }` how could this assert ever fail?

packages/governance/src/contractGovernance/governFilter.js
```diff
@@ -0,0 +1,117 @@
...
+            X`The question (${q(strings)}) didn't match the outcome ${outcome}`,
+          );
+
+          return E(governorFacet)
+            .setOfferFilter(strings)
+            .then(() => {
+              return positive;
+            });
+        } else {
```
turadg: IIUC `keyEQ` will deeply compare the string values. So this `else` isn't necessarily a `negative` vote. Check instead for the schema of the `negative` vote with `if else` and have an `else { assert.fail` for the unknown question schema.

packages/governance/src/contractGovernance/governFilter.js
```diff
@@ -0,0 +1,117 @@
...
+    voteCounters.add(voteCounter);
+
+    // CRUCIAL: Here we wait for the voteCounter to declare an outcome, and then
+    // attempt to invoke the API if that's what the vote called for. We need to
+    // make sure that outcomeOfUpdateP is updated whatever happens.
+    //
+    // * If the vote passed, invoke the API, and return the positive position
+    // * If the vote was negative, return the negative position
+    // * If we can't do either, (the vote failed or the API invocation failed)
+    //   return a broken promise.
```
turadg, line 79: Can the caller distinguish between the two kinds of failures? If that's supported part of the API, let's document it in JSDoc. If it isn't, let's say so explicitly.

turadg (changes requested):
Looked through the first few files. Will come back for more but wanted to get these requests in.

packages/governance/src/contractGovernor.js
```diff
@@ -185,6 +191,15 @@ const start = async (zcf, privateArgs) => {
       getUpdatedPoserFacet,
     );
 
+  const { voteOnFilter, createdQuestion: createdFilterQuestion } =
```
turadg: since this a return value of `setupFilterGovernance` why not have it return with key `createdFilterQuestion`?

I see there's a `FilterGovernor` type that would need to be updated. It's only used here but I see it's made for symmetry with `ApiGovernor` which is returned by two functions.

packages/governance/src/contractHelper.js
```diff
@@ -99,13 +99,18 @@ const facetHelpers = (zcf, paramManager) => {
    * @param {Record<string, (...args: any[]) => any>} governedApis
    * @returns {GovernedCreatorFacet<CF>}
    */
+
+  /**
+   * @param {{}} originalCreatorFacet
+   * @param {{}} governedApis
```
turadg: Looks like this is optional:
```suggestion
   * @param {{}} [governedApis]
```

packages/governance/src/types.js
```diff
@@ -445,6 +463,16 @@
  * @property {CreateQuestion} createQuestion
  */
 
+/**
+ * @typedef {object} GovernorFacet
```
turadg: I was surprised we didn't have this type yet. Reading through https://github.com/Agoric/agoric-sdk/blob/a866f983a6a14c8b87ecc933658773e7afe350c5/packages/governance/README.md?plain=1#L235-L240 I think we'd benefit from defining an interface.

We can't have a formal interface with `.ts` syntax but we can define it as a typedef with a comment to explain.

E.g.,
```js
/**
 * @typedef GovernableCreatorFacet
 * @property {() => ParamManagerRetriever} getParamMgrRetriever
 * @property {(name: string) => Promise<Invitation>} getInvitation
 * @property {() => unknown} getLimitedCreatorFacet
 */
```

Then this would extend that,
```suggestion
 * @typedef {GovernableCreatorFacet} GovernorFacet
```

packages/governance/test/swingsetTests/contractGovernor/bootstrap.js
```diff
@@ -198,6 +198,31 @@ const setupParameterChanges = async (
...
+  const { details, instance, outcomeOfUpdate } = await E(
+    governor,
+  ).voteOnOfferFilter(installations.binaryVoteCounter, 2n, filterChangeSpec);
+
+  E(E(zoe).getPublicFacet(instance))
+    .getOutcome()
+    .then(outcome => log(`vote outcome: ${q(outcome)}`))
+    .catch(e => log(`vote failed ${e}`));
+
+  E.when(outcomeOfUpdate, outcome => log(`updated to (${q(outcome)})`)).catch(
```
turadg, line 220: TIL `E.when`  branches off the deferred resolution without affecting other consumers of `outcomeOfUpdate` resolution.

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -197,9 +197,9 @@ async function makePsmDriver(t, customTerms) {
     },
 
     async getFeePayout() {
-      // @ts-expect-error XXX governance types
       const limitedCreatorFacet = E(creatorFacet).getLimitedCreatorFacet();
       const collectFeesSeat = await E(zoe).offer(
+        // @ts-expect-error XXX governance types
```
turadg: progress! 😆

packages/zoe/src/contractFacet/types.js
```diff
@@ -46,6 +46,7 @@
  * @property {() => Promise<void>} stopAcceptingOffers
  * @property {(strings: Array<string>) => void} setOfferFilter
  * @property {() => Instance} getInstance
+ * @property {(strings: Array<string>) => void} setOfferFilter
```
turadg: dupe with line 47

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: My view is that the `negative` position should say what it's voting no on. When there are three options on the ballot, each should be self-describing. They shouldn't be meaningful only by reference to the other positions. 

governAPI's negative gives the method name (but no parameters) and governParams gives the parameters (but not the proposed values). Including the current value of the filters wouldn't be correct, since the voting period for two vote proposals could overlap, so voting no won't necessarily leave the filters as they were before the proposal.

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: @erights   How inspectable are broken promises?  Both kinds of failures would be produced organically by async methods that don't succeed. They aren't signified by throwing any particular error.  I know i can distinguish them from their stack traces during debugging, but I don't know how legible they'd be to code.

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: done

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: good point.

Chris-Hibbert: good point.

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: done

packages/governance/src/contractGovernor.js
Chris-Hibbert: done

packages/governance/src/types.js
Chris-Hibbert: GovernedCreatorFacet is the def'n you're looking for.

I added CF as a parameterized type to GovernorFacet.

packages/zoe/src/contractFacet/types.js
Chris-Hibbert: done

packages/inter-protocol/test/psm/test-psm.js
Chris-Hibbert: and then it jumped around again!

packages/governance/src/contractHelper.js
Chris-Hibbert: done

packages/governance/src/contractGovernance/governFilter.js
turadg: > My view is that the negative position should say what it's voting no on. When there are three options on the ballot, each should be self-describing. They shouldn't be meaningful only by reference to the other positions.

My view is that the position should say what the result of voting for it will be. The positive says "these will now be the filters". The negative can only say "the filters will not change". That "dontUpdate" has a list of strings can easily be read as some qualification on what not to update.

You're the lead on governance so if my argument is not convincing then I defer to your view. Whatever principle you use to decide, that is a de facto design principle of governance. I ask that it be documented because it's part of the human API of positions. (Apologies if it is and I have missed it.)

packages/governance/src/contractGovernance/governFilter.js
turadg: A promise is either resolved (with a value) or rejected (with a [reason](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch#syntax)). So it should be possible to provide different reasons for these two kinds of failure.

What I don't know is whether we want to commit to that as part of this API. Leaving that to y'all but my request is that the decision be documented.

packages/governance/src/contractGovernance/governFilter.js
```diff
@@ -0,0 +1,114 @@
...
+    const outcomeOfUpdate = E(counterPublicFacet)
+      .getOutcome()
+      .then(outcome => {
+        if (keyEQ(outcome, positive)) {
+          return (
+            E(governorFacet)
+              // @ts-expect-error typescript doesn't see GovernorFacet.setOfferFilter
```
turadg: fixed and pushed

turadg (approved):
My outstanding requests may be resolved by documentation so I'll approve and you can document here or follow up.

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: I added more details in the governance README.

packages/governance/src/contractGovernance/governFilter.js
erights: First a bit of terminology: There are two ways to use an error to indicate something went wrong: throw, or using the error as a rejected-promise-reason. When we don't care to distinguish these, I'll use the term "raise error" to mean both.

To grok this question in its fullness, we must first distinguish two possible purposes of the information carried by a raised error.
   * Something went wrong. To provide an out-of-band human programmer with diagnostic information to help them diagnose a possible bug in some code somewhere. 
   * To enable code to test-and-branch on something, such that subsequent non-diagnostic computation proceeds differently depending on this information carried by the error. We need to distinguish two subcases:
      * Use to express non-local non-erroneous control flow
      * To react differently to different kinds or serverities of what went wrong. Depending on what kind of thing went wrong, we may wish to take more severe action programmatically, like terminating a vat or shutting down the chain.

One unfortunate feature of JavaScript creates an additional complication, which we cope with by using another unfortunate feature of JavaScript.
   * JavaScript has no non-local non-erroneous exit operation suitable for use as non-erroneous control-flow.
   * JavaScript allows non-errors to be raised (thrown or used as promise rejections).

For non-local non-erroneous control flow, our recommended practice (which we've only recently decided on and started practicing) is that normal non-erroneous control flow, not indicating a problem that should be brought to human attention, should happen by raising a non-error. Then, the code that catches something and branches is not asking "Is this an error I expect or something else?" but rather "Is this a non-error I expect or something else?". In both cases, if it is something else the normal reaction will be to rethrow as if the intermediary were not in the way. 

Let's take the prescriptive stance that, for production code, only the ""Is this a non-error I expect or something else?" conditional is legitimate. Non-diagnostic production code should never have a conditional of the form "Is this an error I expect or something else?"

This leaves the final case, where different kinds of things going wrong should result in different severities of recovery action. For these, generally, the answer is to report the error in some way other than raising it, such as a call to `zcf.shutdownWithFailure`. An important cross-over case is what happens if an error is thrown during the execution of an atomic block after the commit point. In this case, the catcher escalates the severity of a thrown error into a shutdown. But the catcher does this escalation unconditionally --- again, without inspecting the information conveyed by the error.

So now, returning to our distinct error conditions
> (the vote failed or the API invocation failed)

I covered with @Chris-Hibbert what each means, and both indicate something went wrong, such that diagnostic info would likely be useful to help a programmer figure it out. Neither represents a non-erroneous control flow condition. Neither calls for a higher severity reaction, like sudden vat termination. Thus, non-diagnostic code should not distinguish between them, and the API spec should not imply that there is any such distinction for code to test.

This leaves only the question about whether this particular piece of code, responsible for propagating both of these errors, should bother to intercept one or both in order to enhance the diagnostic information it carries, for example, by using `assert.note`. This is always an esthetic judgement call. But my sense from talking to @Chris-Hibbert is there is nothing about the distinction between these two cases that isn't adequately conveyed by their stack traces. 

> Leaving that to y'all but my request is that the decision be documented.

@Chris-Hibbert can I leave it to you to see that this explanation be moved to a suitable place? Thanks.

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: Oops!  I saw your fixes, and **liked what you had done**.  Then I pushed more changes without having pulled your changes down. AFAICT, I erased the evidence on github of your good work. Can you find and push your changes again?
@turadg

turadg: I tried to pull rebase to push them again but somehow I clobbered them. I managed to reconstruct and go a little farther in https://github.com/Agoric/agoric-sdk/compare/5796-governPause...ta/governance-types

Feel free to include in this or I can PR that after this lands. They need a little cleaning up.

packages/governance/src/contractGovernance/governFilter.js
Chris-Hibbert: > can I leave it to you to see that this explanation be moved to a suitable place? 

Yes. There's room for it in the Wiki.

PR #6039, opened 2022-08-24
fix(agoric-cli): Reliable publishBundle client

### LINKED ISSUE #5460: Chain ACKs publish bundle

kriskowal:
closes: #5460

## Description

Change #5541 causes Cosmic Swingset to report the results of InstallBundle messages. This change improves the reliability of `publishBundle` in deploy scripts by:

1. Instead of shelling out to `agd` to send an InstallBundle message to the chain, uses CosmJs. Consequently, gets the block height the message was submitted against, so the lowest height that we can look for a response in the IAVL tree synced to RPC nodes.
2. Retries around submitting an InstallBundle message.
3. Follows the `:published.installation` vstorage topic broadcast by the Agoric chain to observe whether the InstallBundle message resulted in successful or unsuccessful installation.

This necessitated a bug fix, and so inspired a refactor, of `@agoric/casting` such that it does not retain I/O handles between yielding a value and the holder requesting the next value. The resulting async iterators should more closely conform to the intended semantics of medial-lossy and medial-lossless subscriptions without needing to poll and observe every block over the lifetime of a subscription, as is now possible with Richard’s refactor of storage cells.

### Security Considerations

Clients should not rely on the shapes of storage cells.

My intention is to follow-up on this change with improved deploy-script helpers and integration of those helpers in example dapps such that `publishBundle` is abstracted away from the attention of end-users. The fungible-faucet scaffold dapp then serves as the documentation for the proper way to publish a contract.

### Testing Considerations

For manual verification, in `dapp-fungible-faucet`:

```sh
# tab 1
agoric start local-chain --reset

# tab 2
agoric start local-solo --reset

# tab 3
agoric follow :published.installation

# tab 4
agoric deploy contract/deploy.js --need=agoric
```

Otherwise, the tests in `casting` cover the refactor, though these could no-doubt be elaborated to cover more of the new behavior.

### DIFF
--- a/packages/agoric-cli/src/publish.js
+++ b/packages/agoric-cli/src/publish.js
@@ -1,6 +1,48 @@
 // @ts-check
 /// <reference types="ses"/>
 
+import { E } from '@endo/far';
+
+import {
+  iterateEach,
+  makeFollower,
+  makeLeaderFromRpcAddresses,
+  makeCastingSpec,
+} from '@agoric/casting';
+import { DirectSecp256k1HdWallet, Registry } from '@cosmjs/proto-signing';
+import { defaultRegistryTypes } from '@cosmjs/stargate';
+import { stringToPath } from '@cosmjs/crypto';
+import { Decimal } from '@cosmjs/math';
+import { Bech32 } from '@cosmjs/encoding';
+
+import { MsgInstallBundle } from '@agoric/cosmic-proto/swingset/msgs.js';
+
+// https://github.com/Agoric/agoric-sdk/blob/master/golang/cosmos/daemon/main.go
+const Agoric = {
+  Bech32MainPrefix: 'agoric',
+  CoinType: 564,
+  proto: {
+    swingset: {
+      InstallBundle: {
+        // matches package agoric.swingset in swingset/msgs.go
+        typeUrl: '/agoric.swingset.MsgInstallBundle',
+      },
+    },
+  },
+  // arbitrary fee for installing a bundle
+  fee: { amount: [], gas: '50000000' },
+  // Agoric chain does not use cosmos gas (yet?)
+  gasPrice: { denom: 'uist', amount: Decimal.fromUserInput('50000000', 0) },
+};
+
+const hdPath = (coinType = 118, account = 0) =>
+  stringToPath(`m/44'/${coinType}'/${account}'/0/0`);
+
+const registry = new Registry([
+  ...defaultRegistryTypes,
+  [Agoric.proto.swingset.InstallBundle.typeUrl, MsgInstallBundle],
+]);
+
 /**
  * @typedef {object} JsonHttpRequest
  * @property {string} hostname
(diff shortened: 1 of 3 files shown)


packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +138,376 @@ export const makeCosmjsFollower = (
...
+    for (;;) {
+      // AWAIT
+      const data = await getDataAtHeight(blockHeight);
+      if (data.length !== 0) {
+        const streamCell = streamCellForData(blockHeight, data);
+        return { data, streamCell };
       }
-    };
+      // AWAIT
+      await E(leader).jitter(where);
```
kriskowal: This retry loop will never exit. I need to account for the case of an empty data cell differently here, and recover in the caller. We previously allowed JSON.parse to gracelessly throw if we subscribed before the first publish and I think we can do better by waiting for a block to produce some data.

packages/agoric-cli/src/publish.js
```diff
@@ -1,6 +1,48 @@
...
+  proto: {
+    swingset: {
+      InstallBundle: {
+        // matches package agoric.swingset in swingset/msgs.go
+        typeUrl: '/agoric.swingset.MsgInstallBundle',
+      },
+    },
+  },
+  // arbitrary fee for installing a bundle
+  fee: { amount: [], gas: '50000000' },
```
dckc, line 33: This shouldn't be hard-coded; we do have near to medium term plans to charge by the byte: #6097

packages/agoric-cli/src/publish.js
```diff
@@ -1,6 +1,48 @@
...
+  // Agoric chain does not use cosmos gas (yet?)
+  gasPrice: { denom: 'urun', amount: Decimal.fromUserInput('50000000', 0) },
```
dckc:
```suggestion
  gasPrice: { denom: 'uist', amount: Decimal.fromUserInput('50000000', 0) },
```

dckc (commented):
looks pretty good

I don't think I should approve it without kicking the tires a bit more, but I'm happy if somebody else does.

packages/agoric-cli/src/publish.js
dckc: not critical; seems reasonable for it to come in a follow-on PR, once requirements become more clear

kriskowal:
> looks pretty good
> 
> I don't think I should approve it without kicking the tires a bit more, but I'm happy if somebody else does.

Appreciate the caution. My intention is to provide a companion changes to deploy-script-helpers and to dapp-fungible-faucet, pinned in CI to be sure.

dckc (approved):
Yay! integration test was successful!

I watched it go on Kris's desk. I'm willing to bet it works elsewhere.

PR #6040, opened 2022-08-24
chore(swingset): update some utilities in misc-tools/

warner:
This adds some tools to process a slogfile or transcript and:
* measure the time spent doing deliveries, correlated by method name
* measure the time spent doing syscalls, correlated by syscall type
* upgrade `extract-contract-bundle.js` to handle the new bundlecap approach


dckc (approved):
Approving in the sense of: yes, these are some misc swingset tools that worked at one point. No guarantees about future maintenance.

warner:
cool, just wanted to get them off my hard drive and available to others

dckc:
quite... thanks for sharing!

PR #6043, opened 2022-08-24
chore(swingset): more misc-tools updates

warner:
This refactors timestamp correlation into a library, handles 'console'
events correctly (which are not syscalls but still insert timestamps
into the metering results), and adds a generic pair-deltas.py for
statistical measurements of time between two arbitrary event types.


warner:
Another rubber-stamp for you :)

warner:
I have some more tools to include, hold off on review for a minute

warner:
ok ready now

dckc (approved):
thanks for sharing! yes, let's archive these with the rest.

PR #6047, opened 2022-08-24
6007 test pause PSM

Chris-Hibbert:
refs: #6007

## Description

A test showing that governance can set offer filters for the PSM and prevent trading

Changes to contractHelper to get the parameters right in non- and virtual- governorFacets.

### Security Considerations

minimal

None

### Testing Considerations

This is a mainly test.

### DIFF
--- a/packages/inter-protocol/test/psm/setupPsm.js
+++ b/packages/inter-protocol/test/psm/setupPsm.js
@@ -74,10 +74,11 @@ export const setupPsmBootstrap = async (
   produce.agoricNamesAdmin.resolve(agoricNamesAdmin);
 
   installGovernance(zoe, spaces.installation.produce);
-  produce.chainStorage.resolve(makeMockChainStorageRoot());
+  const mockChainStorage = makeMockChainStorageRoot();
+  produce.chainStorage.resolve(mockChainStorage);
   produce.board.resolve(makeBoard());
 
-  return { produce, consume, ...spaces };
+  return { produce, consume, ...spaces, mockChainStorage };
 };
 
 /**
@@ -163,10 +164,6 @@ export const setupPsm = async (
     E(zoe).getInvitationIssuer(),
   ).getAmountOf(poserInvitationP);
 
-  /** @type {import('@agoric/vats/tools/storage-test-utils.js').MockChainStorageRoot} */
-  // @ts-expect-error cast
-  const mockChainStorage = await space.consume.chainStorage;
-
   return {
     zoe,
     installs,
@@ -176,7 +173,7 @@ export const setupPsm = async (
     governor: g,
     psm,
     invitationAmount: poserInvitationAmount,
-    mockChainStorage,
+    mockChainStorage: space.mockChainStorage,
     space,
     knutIssuer,
   };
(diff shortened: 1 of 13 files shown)


packages/inter-protocol/test/psm/setupPsm.js
```diff
@@ -0,0 +1,184 @@
...
+  const committeeCreator = await consume.economicCommitteeCreatorFacet;
+  const electorateInstance = await instance.consume.economicCommittee;
+
+  const poserInvitationP = E(committeeCreator).getPoserInvitation();
+  const poserInvitationAmount = await E(
+    E(zoe).getInvitationIssuer(),
+  ).getAmountOf(poserInvitationP);
+
+  /** @type {import('@agoric/vats/tools/storage-test-utils.js').MockChainStorageRoot} */
+  // @ts-expect-error cast
```
turadg: you could get this directly from `setupPsmBootstrap` where it has the type `MockChainStorageRoot`.

packages/inter-protocol/test/psm/setupPsm.js
```diff
@@ -0,0 +1,184 @@
...
+harden(setUpZoeForTest);
+/**
+ * @typedef {ReturnType<typeof setUpZoeForTest>} FarZoeKit
+ */
+
+/**
+ * @param {TimerService} timer
+ * @param {FarZoeKit} [farZoeKit]
+ */
+export const setupPsmBootstrap = async (
```
turadg, line 54: this isn't specific to PSM. I wonder if we can DRY it out with other tests

packages/inter-protocol/test/psm/test-governedPsm.js
```diff
@@ -0,0 +1,62 @@
+// @ts-check
+
+// eslint-disable-next-line import/no-extraneous-dependencies
```
turadg: is this suppression necessary? `@agoric/zoe` is declared as a dependency

turadg (approved):

packages/inter-protocol/test/psm/setupPsm.js
Chris-Hibbert: It was copied from AMM, and and apparently didn't require any changes. I don't see other contracts that do the same thing, and I don't know how to generalize what's happening. I'd prefer to do the cleanup as a separate task.

packages/inter-protocol/test/psm/setupPsm.js
Chris-Hibbert: I don't understand what you're suggesting.

is it a better source for the type info, or calling `makeMockChainStorageRoot()` here rather than in `setupPsmBootstrap`?

packages/inter-protocol/test/psm/test-governedPsm.js
Chris-Hibbert: copy-pasta

packages/inter-protocol/test/psm/setupPsm.js
turadg: in short, better source for the type info. If you exported the value from `setupPsmBootstrap` then the type would come along with it and you wouldn't need the annotation/cast here.

I think you _could_ call `makeMockChainStorageRoot()` again here, but I don't see a benefit. It's important that each test have its own chain storage, but each test is calling `setupPsm` and each of those is calling `setupPsmBootstrap`.

Chris-Hibbert: done

PR #6048, opened 2022-08-24
chore: update  to match gaia 7.0.2

### LINKED ISSUE #5870: Upgrade agd to match ag0 on gaia 7.0.2

We've upgraded ag0 to match the latest revision of gaia (7.0.2) including transitive dependencies (#5773). We need to make corresponding changes to agd so that mainnet1 doesn't backslide in any of our dependencies.

## Description of the Design

Since the agd application doesn't share a version history with gaia code, changes to get to 7.0.2 will have to be discovered and applied manually.

JimLarson:
closes: #5870 

## Description

Update dependencies (particularly ibc-go) to match the versions used in gaia 7.0.2 and ag0 upgrade 7.

Structured to minimize differences vs Agoric/ag0.

### Security Considerations

We need to pay particular attention to the IBC configuration as well as the issues we encountered in recent ag0 upgrades.

### Testing Considerations

Need to test full chain startup/shutdown and IBC usage.

#loadgen-branch: mhofman/handle-agd-update


go.mod
```diff
@@ -3,24 +3,23 @@ module github.com/Agoric/agoric-sdk
...
 
 require (
-	github.com/armon/go-metrics v0.3.9
-	github.com/cosmos/cosmos-sdk v0.45.1
-	github.com/cosmos/gaia/v6 v6.0.3
-	github.com/cosmos/ibc-go/v2 v2.0.3
+	github.com/armon/go-metrics v0.3.10
+	github.com/cosmos/cosmos-sdk v0.45.4
+	github.com/cosmos/gaia/v7 v7.0.0-00010101000000-000000000000
+	github.com/cosmos/ibc-go/v3 v3.0.0
```
JimLarson, line 9: Reviewer - do we still need to use our own fork of ibc-go?

michaelfig: Not at this time.  That's applicable only when we consider adopting Interchain Account host support.

[Our fork](https://github.com/agoric-labs/ibc-go/Agoric) implements ICA message type allowlist wildcards.  This is to more conveniently restrict Cosmos-layer messages if the product folks deem some of them a bad idea to enable on the Agoric chain.

JimLarson:
The latest commit fixes CI (proto lint breakage is expected), but we don't fully understand the reason yet. Will move from draft to regular PR once we understand.

JimLarson:
No deep understanding of the genesis order change yet, but moving out of "draft" state so the overall review can proceed in parallel.

golang/cosmos/agoric/config.go
```diff
@@ -0,0 +1,35 @@
...
+	Bech32PrefixAccPub = Bech32MainPrefix + sdk.PrefixPublic
+	// Bech32PrefixValAddr defines the Bech32 prefix of a validator's operator address
+	Bech32PrefixValAddr = Bech32MainPrefix + sdk.PrefixValidator + sdk.PrefixOperator
+	// Bech32PrefixValPub defines the Bech32 prefix of a validator's operator public key
+	Bech32PrefixValPub = Bech32MainPrefix + sdk.PrefixValidator + sdk.PrefixOperator + sdk.PrefixPublic
+	// Bech32PrefixConsAddr defines the Bech32 prefix of a consensus node address
+	Bech32PrefixConsAddr = Bech32MainPrefix + sdk.PrefixValidator + sdk.PrefixConsensus
+	// Bech32PrefixConsPub defines the Bech32 prefix of a consensus node public key
+	Bech32PrefixConsPub = Bech32MainPrefix + sdk.PrefixValidator + sdk.PrefixConsensus + sdk.PrefixPublic
+)
```
kriskowal, line 27: I’ve seen these elsewhere. We should have a reference to the source-of-truth somewhere.

kriskowal (approved):
Looks mechanical to me.

dckc:
I did a quick `make scenario2-run-chain-economy` smoke test. JS bootstrap went fine and it started making blocks. :+1:

golang/cosmos/agoric/config.go
JimLarson: Discussed with Kris and the equivalent code on the JS side lives in several places. Since we can't put a coherent reference here at the moment, we'll leave it as a TODO for the JS side to consolidate, then cross-reference.

mhofman:
`deployment-test` is failing with
```
 Error: unknown command "unsafe-reset-all" for "ag-chain-cosmos"
```

Looks like some backwards incompatible change in the new cli tool.

Issue #5956, opened 2022-08-13
liveslots/marshal option to reject oversized inbound messages


warner:
## What is the Problem Being Solved?

There's a simpler version of #5955 that might buy us some useful protection. The idea would be that liveslots checks on the size of the capdata *before* deserialization. If the `.body` is too large, or if there are too many `.slots`, then `dispatch.deliver` rejects the result promise before it even attempts to deserialize anything.

That would protect the memory footprint of the target vat. We'd incur the cost of copying the `VatDeliveryObject` around, but we'd avoid any expansion of the body (e.g. `[],[],[],` expanding into N `Objects` costing a couple of words each). If we had a way to pre-check the slots for new vpids, we could avoid the expense of all the promise-tracking code (as well as the `subscribe` calls we'd normally do during deserialization).

And refusing a message at that (early) point would let us cleanly report the failure to the sender (who might also be a victim, if an attacker submitted extra properties in some other message to them, with the intent of getting that inflation pushed into their message to us).

## Description of the Design

The liveslots `dispatch.deliver()` code would inspect the capdata before calling `unserialize`.

We'd need some configuration mechanism. It *could* be hard-coded, like the `SYSCALL_CAPDATA_BODY_SIZE_LIMIT = 10*MB` we have now, or we could add some more-visible scheme (`vatPowers.setLimit()` from within the vat, `dynamicVatOptions` passed to `E(vatAdminSvc).createVat()` from the outside.. same as #5954)

## Security Considerations

This would not protect against accumulative attacks: those will require something more like Meters and per-message charges ("stamps?" #3103 has some discussion).

This would not protect against `dispatch.notify` resolution data. We *could* apply a check there, but how should we report a failure? The resolver is at fault, but they aren't listening anymore. One option would be to replace the oversized resolution with a `sorry it was too big i ate it` rejection before delivering it to the local subscriber's callback.

## Test Plan

unit tests

warner:
Parts of this might be implemented by configuring `marshal` to apply size checks. It might do these in `unserialize`, in `serialize`, or in both. If we do this, and want to make the limits configurable, we'd probably need liveslots to wrap a mutable `m` marshaller in a pair of stable `serialize`/`unserialize` functions, because we pass the latter around to lots of different components (virtualObjectManager.js, virtualReferenceManager.js, collectionManager.js). `marshal` has no long-term internal state, so it would only be statically configurable, and when we wanted to change the limits, we'd build a replacement marshaller and throw out the old one.

[...]

Issue #5955, opened 2022-08-13
use method schemas to limit inbound argument shapes


warner:
## What is the Problem Being Solved?

#5954 seeks to reduce memory-consumption attacks by unilaterally rejecting old unresolved inbound argument promises. It assumes that most inbound argument promises are short-lived, so any that remain long enough are part of an attack. If userspace did not see the doomed promise, it will not notice the (surprising, not kernel-sanctioned) rejection.

If an attacker sends:
```js
function makeEmptyPromise() {
  return new Promise(resolve => undefined);
}
const emptyPromises = [];
for (let i = 0; i < 1000; i++) {
  emptyPromises.push(makeEmptyPromise());
}
E(victim).normalAPIMethod(normalArgument, ...emptyPromises);
```

(or, to avoid a per-message size limit constraint):
```js
for (let i = 0; i < 1000; i++) {
  E(victim).normalAPIMethod(normalArgument, makeEmptyPromise());
}
```

and the victim implements:
```js
const root = Far('root', {
  normalAPIMethod(normalArgument) {
    ..
```

then those extra promises will be ignored, and we use the #5954 technique to eventually reject them and shed the memory obligations. This wouldn't kick in until some number (perhaps 10,000) promises had been accumulated that way, to reduce (but not eliminate) the risk of rejecting an innocent promise that just took too long to resolve.

But, it would be great if we could learn that they were unused by userspace earlier, so we could remove them from `importedVPIDs` right away, during the `dispatch.deliver` that unserialized them a moment before.

@erights is working on a scheme to allow Far objects to declare a schema. This would name all of their exposed methods, and for each one it would define a pattern that limits the acceptable arguments it accepts. On one level, this acts as a standardized declaration of argument preconditions: the kind of thing we usually implement with a batch of `assert`s at the top of the function. The primary goal is safety and brevity: methods could declare their schema instead of writing a set of asserts (which might be incomplete). It might be possible to extract TypeScript data about the method and semi-automatically construct the schema record.

But if this structure is visible to liveslots (or other holders of the Far object), then it could *also* be used to test recently-unserialized arguments before invoking the target. If liveslots knew that the schema was not met, it would also know that the userspace function was never invoked. Therefore any new promises in those arguments could not possibly have had their `.then` invoked, removing liveslots' obligation to track and resolve them. The liveslots `dispatch.deliver` code would look something like:

* unserialize arguments, identify new promises (instrument `makePipelinedPromise` somehow)
[...]

Issue #5954, opened 2022-08-13
liveslots: reject too many imported promises


warner:
## What is the Problem Being Solved?

Part of the memory consumption that can be [inflicted](https://github.com/Agoric/agoric-sdk/issues/5953) upon an innocent vat is when the inbound `dispatch.deliver` or `dispatch.notify` contains promise vrefs (vpids) in the arguments or resolution data. Liveslots creates actual `Promise` objects for each of these, passes them to userspace, and [remembers](https://github.com/Agoric/agoric-sdk/blob/ab99e75cf1606f6892275f81696b9433e42fddc0/packages/SwingSet/src/liveslots/liveslots.js#L781-L782) their `resolve`/`reject` functions in the `importedVPIDs` table.

Each promise incurs memory consumption within the vat worker, which cannot be shed until it receives a `dispatch.notify` and the Promise can be settled. At that point, the resolved promise loses its identity, and liveslots is allowed to stop tracking it. If userspace does not retain a reference, the `Promise` can then be garbage-collected by the engine.

The problem is that an attacker might send messages with spurious extra arguments, which contain vpids, to provoke this memory consumption. The victim vat is not looking for extra arguments (or extra properties on normal arguments): it could, but it would be pretty awkward (walking the whole argument object graph). And even if it did that, the memory consumption is triggered when liveslots deserializes the data, so by the time userspace gets them, it is too late.

The same is true for promises that appear in resolution data (when `syscall.notify` includes capdata with new vpids in `slots`), but it's even more awkward for userspace to be on the lookout for these. It would have to subscribe to every promise it ever sees, just to learn about even more promises in the resolution.

As a general design principle, we prefer to get promises *from* a service over passing them *in*. And when we do pass a promise in, the intention is to use pipelining to save roundtrips, so we expect the promise to resolve soon, possibly even on the vat to which we're sending the message.

So in normal usage, we don't expect these argument-carried promises to remain unresolved for very long. This points to a potential defensive mechanism (suggested by @dtribble): allow liveslots to unilaterally `reject()` argument promises that have remained unresolved for too long.

Using a timeout is troublesome (unpredictable/nondeterministic, for one), but a similar effect can be had by declaring a maximum count of unresolved argument promises, building a length-limited FIFO, pushing promises onto the FIFO upon receipt, and rejecting them if they reach the far end. Dean suggested a limit of 10k promises.

[...]

PR #6057, opened 2022-08-26
fix: patterns impose resource limits

erights:
For various type-oriented patterns like `M.array()`, where the pattern itself does not naturally imply a length limit, 
   * add an optional `limits` argument to the pattern, where the limits argument is a record of optional named limits
   * provide generous default limits that applies if no explicit `limits` are provided. 

The limits are expressed as numbers, not bigints. Unlimited size can thus still be expressed with a limit value of `Infinity`.

The limits are
    * `decimalDigitsLimit`, for `M.bigint`, `M.nat`, and the cardinality of `M.bag` and `M.bagOf`
    * `stringLengthLimit`, for `M.string`
    * `symbolNameLengthLimit`, for `M.symbol`
    * `numPropertiesLimit`, for `M.record` and `M.recordOf`
    * `propertyNameLengthLimit`, for `M.record` and `M.recordOf`
    * `arrayLengthLimit`, for `M.array` and `M.arrayOf`
    * `numSetElementsLimit`, for `M.set` and `M.setOf`
    * `numUniqueBagElementsLimit:`, for `M.bag` and `M.bagOf`
    * `numMapEntriesLimit`, for `M.map` and `M.mapOf`

See 
https://github.com/Agoric/agoric-sdk/issues/5955 (fixes it?)
https://github.com/Agoric/agoric-sdk/issues/5956
https://github.com/Agoric/agoric-sdk/issues/5954
#6098 
#6060 
#6061 
#6062 
#5956


packages/store/test/test-pattern-limits.js
```diff
@@ -0,0 +1,243 @@
...
+  {
+    specimen: -379n,
+    yesPatterns: [M.bigint(), M.bigint(harden({ decimalDigitsLimit: 3 }))],
+    noPatterns: [
+      [
+        M.bigint({ decimalDigitsLimit: 2 }),
+        'bigint "[-379n]" must not have more than 2 digits',
+      ],
+      [M.nat(), '"[-379n]" - Must be non-negative'],
+      [M.nat({ decimalDigitsLimit: 2 }), '"[-379n]" - Must be non-negative'],
```
erights: fix non-uniformity of error message

packages/store/test/test-pattern-limits.js
```diff
@@ -0,0 +1,243 @@
...
+    noPatterns: [[M.array(), 'Array length 10001 must be <= limit 10000']],
+  },
+  // numSetElementsLimit
+  {
+    specimen: makeCopySet([0, 1, 2, 3, 4, 5]),
+    yesPatterns: [M.set(), M.setOf(M.number())],
+    noPatterns: [
+      [
+        M.set(harden({ numSetElementsLimit: 3 })),
+        'Set must not have more than 3 elements: 6',
```
erights: fix non-uniformity of error message

packages/store/test/test-pattern-limits.js
```diff
@@ -0,0 +1,243 @@
...
+      [0, 37n],
+      [1, 3n],
+      [2, 100000n],
+      [3, 1n],
+    ]),
+    yesPatterns: [M.bag(), M.bagOf(M.number())],
+    noPatterns: [
+      [
+        M.bag(harden({ numUniqueBagElementsLimit: 3 })),
+        'Bag must not have more than 3 unique elements: "[copyBag]"',
```
erights: fix non-uniformity of error message between 206 and 222. Make all follow the winner.

erights:
This is now ready for review!

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
 
+// /////////////////////// Match Helpers Helpers /////////////////////////////
+
+/**
+ * exported primarily for testing.
+ */
+export const defaultLimits = harden({
+  decimalDigitsLimit: 100,
+  stringLengthLimit: 100_000,
+  symbolNameLengthLimit: 40,
```
dtribble: Since these may be generated, I'd up this to 100

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
 
 /** @type WeakSet<Pattern> */
 const patternMemo = new WeakSet();
 
+// /////////////////////// Match Helpers Helpers /////////////////////////////
+
+/**
+ * exported primarily for testing.
+ */
+export const defaultLimits = harden({
```
dtribble, line 57: I don't see a bigint size limit

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
+  decimalDigitsLimit: 100,
+  stringLengthLimit: 100_000,
+  symbolNameLengthLimit: 40,
+  numPropertiesLimit: 80,
+  propertyNameLengthLimit: 100,
+  arrayLengthLimit: 10_000,
```
dtribble, line 63: These 10k elements limit seem bigger than I would have expected.

packages/store/src/patterns/patternMatchers.js
dtribble, line 57: It might be worth adding a comment for why the limit was chosen.

decimalDigits: 10**100 > 2*128

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
 /** @type WeakSet<Pattern> */
 const patternMemo = new WeakSet();
 
+// /////////////////////// Match Helpers Helpers /////////////////////////////
+
+/**
+ * exported primarily for testing.
+ */
+export const defaultLimits = harden({
+  decimalDigitsLimit: 100,
```
dtribble, line 58: Ironically, I think this is too small:  
10**100 > 2**256 BUT  NOT 2**512
so I think 500 might be the right number so that it can represent numbers that are 2**512?

dtribble: Hmm. NVM. Anyone using it for keys should define a specific shape.  So perhaps we add a Key256Shape somewhere.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
+ * Thus, the result only needs to support destructuring. The current
+ * implementation uses inheritance as a cheap hack.
+ *
+ * @param {Limits} [limits]
+ * @returns {AllLimits}
+ */
+const limit = (limits = {}) =>
+  /** @type {AllLimits} */ (harden({ __proto__: defaultLimits, ...limits }));
+
+const checkIsLimitPayload = (payload, mainPayloadShape, check, label) => {
```
dtribble: comment?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
+  // const payloadLimitShape = harden(
+  //   M.split(
+  //     mainPayloadShape,
+  //     M.partial(harden([M.recordOf(M.string(), M.number())]), harden([])),
+  //   ),
+  // );
+  // return checkMatches(payload, payloadLimitShape, check, label);
+
+  const mainLength = mainPayloadShape.length;
+  if (payload.length < mainLength || payload.length > mainLength + 1) {
```
dtribble: Why isn't this just `payload.length !== mainLength`?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -36,10 +37,80 @@ import {
...
+    return check(false, X`${q(label)} payload unexpected size: ${payload}`);
+  }
+  const limits = payload[mainLength];
+  payload = harden(payload.slice(0, mainLength));
```
dtribble, line 105: since it's entirely temporary and internal, this harden seems unnecessary?

packages/store/test/test-pattern-limits.js
```diff
@@ -0,0 +1,243 @@
...
+      ],
+      [
+        M.nat({ decimalDigitsLimit: 2 }),
+        'bigint "[379n]" must not have more than 2 digits',
+      ],
+    ],
+  },
+  {
+    specimen: -379n,
+    yesPatterns: [M.bigint(), M.bigint(harden({ decimalDigitsLimit: 3 }))],
```
dtribble: The additional limit expression is pretty clumsy. What about `Limits.decimalDigits(3)` ?

I presume this will be rare so it's not a big deal now.

dtribble: Alternatively have `limit` be almost a synonym for `harden`, and then pull it from the names: `M.bigint(limits({ decimalDigits: 3 }))`

packages/store/src/patterns/patternMatchers.js
```diff
@@ -732,6 +803,105 @@ const makePatternKit = () => {
...
+
+  /** @type {MatchHelper} */
+  const matchStringHelper = Far('match:string helper', {
+    checkMatches: (specimen, [limits = undefined], check) => {
+      const { stringLengthLimit } = limit(limits);
+      return (
+        checkKind(specimen, 'string', check) &&
+        check(
+          specimen.length <= stringLengthLimit,
+          X`string ${specimen} must not be bigger than ${stringLengthLimit}`,
```
dtribble, line 899: All the error construction should be separated into a separate method, and only invoked on failure.

packages/store/src/patterns/patternMatchers.js
dtribble: It is supposed to check whether length is `mainLength` OR `mainLength+1`. Please rewrite it to that.

packages/store/src/patterns/patternMatchers.js
dtribble: The name is also ambiguous. Perhaps `alsoCheckLimits`

dtribble: Aha. This is about whether the pattern is well-formed. Rename the payload checks to "checkIsWellFormed"?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -732,6 +803,105 @@ const makePatternKit = () => {
...
 
+  /** @type {MatchHelper} */
+  const matchBigintHelper = Far('match:bigint helper', {
+    checkMatches: (specimen, [limits = undefined], check) => {
+      const { decimalDigitsLimit } = limit(limits);
+      const absSpecimen = specimen < 0n ? -specimen : specimen;
+      return (
+        checkKind(specimen, 'bigint', check) &&
+        check(
+          `${absSpecimen}`.length <= decimalDigitsLimit,
```
dtribble: Why not change the decimalDigitsLimit to a number and just do a compare. Having to *render* the number is crazy overhead! Or since this doesn't need to be exact, something like `Math.ceil(Math.log10(Math.abs(Number(specimen)))+1 < decimalDigitsLimit`

packages/store/src/patterns/patternMatchers.js
```diff
@@ -732,6 +803,105 @@ const makePatternKit = () => {
...
+
+  /** @type {MatchHelper} */
+  const matchNatHelper = Far('match:nat helper', {
+    checkMatches: (specimen, [limits = undefined], check) => {
+      const { decimalDigitsLimit } = limit(limits);
+      return (
+        checkKind(specimen, 'bigint', check) &&
+        check(specimen >= 0n, X`${specimen} - Must be non-negative`) &&
+        check(
+          `${specimen}`.length <= decimalDigitsLimit,
```
dtribble: As above, don't print the string

packages/store/src/patterns/patternMatchers.js
```diff
@@ -891,11 +1105,27 @@ const makePatternKit = () => {
...
   const matchSetOfHelper = Far('match:setOf helper', {
-    checkMatches: (specimen, keyPatt, check) =>
-      checkKind(specimen, 'copySet', check) &&
-      specimen.payload.every((el, i) => checkMatches(el, keyPatt, check, i)),
+    checkMatches: (specimen, [keyPatt, limits = undefined], check) => {
+      const { numSetElementsLimit } = limit(limits);
+      return (
+        checkKind(specimen, 'copySet', check) &&
+        check(
+          specimen.payload.length < numSetElementsLimit,
```
dtribble, line 1161: Why does this mention `payload` where the other patchers do not at this point where they are checking limits?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -891,11 +1105,27 @@ const makePatternKit = () => {
...
+          X`Set must not have more than ${q(numSetElementsLimit)} elements: ${
+            specimen.payload.length
+          }`,
+        ) &&
+        specimen.payload.every((el, i) => checkMatches(el, keyPatt, check, i))
```
dtribble: Since this is Set, should it be checking that items are unique?

packages/store/src/patterns/patternMatchers.js
dtribble: Mark responded:  the structure of a Set is as a tagged record with a `tag` and `payload`

packages/store/src/patterns/patternMatchers.js
```diff
@@ -905,17 +1135,41 @@ const makePatternKit = () => {
...
+            numUniqueBagElementsLimit,
+          )} unique elements: ${specimen}`,
+        ) &&
+        specimen.payload.every(
+          ([key, count], i) =>
+            checkMatches(key, keyPatt, check, `keys[${i}]`) &&
+            applyLabelingError(
+              check,
+              [
+                `${count}`.length <= decimalDigitsLimit,
```
dtribble: Evil printing the key.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -929,18 +1183,32 @@ const makePatternKit = () => {
...
+    checkMatches: (
+      specimen,
+      [keyPatt, valuePatt, limits = undefined],
+      check,
+    ) => {
+      const { numMapEntriesLimit } = limit(limits);
+      return (
+        checkKind(specimen, 'copyMap', check) &&
+        check(
+          specimen.payload.keys.length <= numMapEntriesLimit,
```
dtribble, line 1238: The format for all the specimen types should be described somewhere, and reference from the top of the matchers section

packages/store/src/patterns/patternMatchers.js
```diff
@@ -1207,15 +1485,15 @@ const makePatternKit = () => {
...
-    set: () => SetShape,
-    bag: () => BagShape,
-    map: () => MapShape,
+    bigint: (limits = undefined) => makeLimitsMatcher('match:bigint', [limits]),
+    nat: (limits = undefined) => makeLimitsMatcher('match:nat', [limits]),
+    string: (limits = undefined) => makeLimitsMatcher('match:string', [limits]),
+    symbol: (limits = undefined) => makeLimitsMatcher('match:symbol', [limits]),
+    record: (limits = undefined) => M.recordOf(M.any(), M.any(), limits),
+    array: (limits = undefined) => M.arrayOf(M.any(), limits),
+    set: (limits = undefined) => M.setOf(M.any(), limits),
```
dtribble, line 1543: this means it will do lots of empty matching of "any" on each element.

dtribble: Please add a matcher for just set/bag/etc. without enumerating the elements.

packages/store/test/test-pattern-limits.js
```diff
@@ -0,0 +1,243 @@
...
+      [
+        M.map(harden({ numMapEntriesLimit: 3 })),
+        'CopyMap must have no more than 3 entries: "[copyMap]"',
+      ],
+    ],
+  },
+]);
+
+test('test pattern limits', t => {
+  for (const { specimen, yesPatterns, noPatterns } of limitTests) {
```
dtribble: Replace iterating over the list of cases with a series of calls, one per case. Then a test failure shows the line number of the failed test.

dtribble (approved):
LGTM.  Several comments about perf improvements and test improvements. I could not spot any issues.

Oh and some questions and suggestions around the initial values.

packages/store/test/test-pattern-limits.js
erights: The sneaky reason I put "Limit" in all the property names is in case we ever want to grow this optional bag of limits into a more general options bag containing other options. But in any case, by the hardening rules I would still write your expression as
```js
M.bigint(limits(harden({ decimalDigits: 3 })))
```
because I don't imagine teaching the harden static checker about `limits` as a special case.

In any case, I took these as suggestions to keep in mind as we evolve this API rather than changes needed for this PR, so resolving. If I misunderstood, please unresolve.

packages/store/src/patterns/patternMatchers.js
erights: That's what I intend `decimalDigitsLimit` to be. Should the name be changed to avoid a misunderstanding? This is the only limit name doing double duty. It applies both to bigints by themselves, and it applies to the cardinality counts of copyBags.

erights: Even > 2**332 ;)

I didn't actually have crypto in mind. Instead of explaining individual choices, I'm adding a comment to `defaultLimits` 

```js
/**
 * The actual default values here are, at the present time, fairly
 * arbitrary choices and may change before they settle down. Of course
 * at some point we'll need to stop changing them. But we should first
 * see how our system holds up with these choices. The main criteria
 * is that they be big enough that "normal" innocent programs rarely
 * encounter these limits.
```

I also agree with your observation below that these limits are not intended to cover cryptograhic values. These should have shapes of their own. So resolving.

erights: I'm resolving this. But if you think I do need to change the name, please unresolve.

packages/store/src/patterns/patternMatchers.js
erights: I agree. Crypto values should have their own shape. Resolving.

packages/store/src/patterns/patternMatchers.js

packages/store/src/patterns/patternMatchers.js
erights: I am very open to suggestions. I really did make rather arbitrary choices. But resolving because we can adjust after this PR is merged. I expect multiple adjustments based on experience.

packages/store/src/patterns/patternMatchers.js
erights: Done.

packages/store/src/patterns/patternMatchers.js
erights: What static check would be safe while allowing you to skip this harden? Resolving.

packages/store/src/patterns/patternMatchers.js
erights: Good idea! But your formula seemed off by 2, and I don't see the purpose of the `ceil` since it's just feeding a floating point inequality. I'm going with
```js
          Math.log10(Math.abs(Number(specimen))) <= decimalDigitsLimit,
```

packages/store/src/patterns/patternMatchers.js
erights: Yup. And I can drop the abs.

packages/store/src/patterns/patternMatchers.js
erights: As verbally discussed, this would be a pervasive style change that we probably should do, but not in this PR. Resolving.

packages/store/src/patterns/patternMatchers.js
erights: The earlier `checkKind` on line 1111 already checks all the invariants that this tagged record must satisfy in order to be considered a set. The items have to be unique keys and appear in reverse sort order (to get the orderings among tagged records right). So if they're not unique keys, we won't get here.

packages/store/src/patterns/patternMatchers.js
erights: Ok, that's three. Extracted a reusable function, as I was too lazy to do for the first two ;)

packages/store/src/patterns/patternMatchers.js
erights: Unresolving. I think your formula was more correct than mine.

erights: Done. your's was closer.
```js
Math.floor(Math.log10(Math.abs(Number(specimen)))) + 1 <= decimalDigitsLimit
```

packages/store/test/test-pattern-limits.js
erights: Whew, that was I lot more manual conversion work than I expected. But done. And worth it: On the way, I screwed up one case and the newly revealed line number pointed me right at it!

packages/store/src/patterns/patternMatchers.js
erights: Did the next best thing: For `M.arrayOf`, `M.setOf`, and `M.matchOf` if the patterns are all `M.any()` then I skip the enumeration.

For `M.recordOf` and `M.bagOf` I still have to enumerate to enforce the per-element size limit.

packages/store/src/patterns/patternMatchers.js

packages/store/src/patterns/patternMatchers.js
erights: Instead I wrote https://github.com/Agoric/agoric-sdk/issues/6148 as a reminder that we need this.

packages/store/test/test-pattern-limits.js
erights: Postponed past this PR https://github.com/Agoric/agoric-sdk/issues/6149

packages/store/test/test-pattern-limits.js
erights: Postponed past this PR https://github.com/Agoric/agoric-sdk/issues/6149

packages/store/test/test-pattern-limits.js
erights: Postponed past this PR https://github.com/Agoric/agoric-sdk/issues/6149

PR #6058, opened 2022-08-26
fix(swing-store): use sqlite transactions for streamStore operations

### LINKED ISSUE #6056: sqlite DB flushes happening every crank, not once per block
## Describe the bug

While looking at `strace` data, Ari and I noticed a lot of SQLite-related writes and flushes happening *during* a block, on every delivery. We suspect this is hurting performance, and that we should only be touching the disk once, at the end of the block.

For durability, databases like SQLite need to `fsync` each time they commit, and that isn't allowed to return until the data is safely written to disk (safe against a power failure). This requires the host kernel (linux) to flush its write buffers and wait for the hardware (SSD or HDD) to report a successful write. On a "spinning rust" HDD, this could take as long as 11ms (1 / 5400 RPM / 60 sec/min). SSDs are faster, but still, we're not supposed to be fsyncing all the time.

We use SQLite in the `streamStore` component of `packages/swing-store`, to hold vat transcript entries.

The host application (cosmic-swingset) calls [`openSwingStore()`](https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/cosmic-swingset/src/launch-chain.js#L192) at startup to create the swingStore. Then, each time the kernel finishes making a [delivery](https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/SwingSet/src/kernel/vat-loader/transcript.js#L37), the kernel does a [`vatKeeper.addToTranscript()`](https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/SwingSet/src/kernel/state/vatKeeper.js#L497-L505). That is supposed to be buffered into an in-RAM pending DB transaction.

Later, when the entire block is complete, the cosmic-swingset [calls](https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/cosmic-swingset/src/launch-chain.js#L286) the swing-store `commit()` function:

https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/swing-store/src/swingStore.js#L331-L341

which calls the neighboring `doCommit`:

https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/swing-store/src/swingStore.js#L313-L326

which uses `txnFinish` to close out the current *LMDB* transaction. However it does not do anything special with the SQLite (`streamStore`) state, which was [set up](https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/swing-store/src/swingStore.js#L305) during swing-store initialization. That [creates a SQLite instance](https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/swing-store/src/sqlStreamStore.js#L41-L47), but leaves it in "autocommit mode":

https://github.com/Agoric/agoric-sdk/blob/372cc51aa14b5fbfc0dc181d80a051e0af0e4d8c/packages/swing-store/src/sqlStreamStore.js#L142-L147

## The Fix

We need to replace this auto-commit mode with one that accumulates the `streamStore` changes until `commit()` is called. The most important reason is for performance, to avoid the constant stream of expensive `fsync` calls in the middle of a block.

A secondary reason is for rollback reliability. If a crank is unwound, but the vat is not terminated, we may have written the nominally-abandoned transcript entry to the streamStore already. Autocommit means this spurious entry will land on disk (and be visible to the next reboot, as it wants to replay entries to bring a worker up-to-date). We happen to be safe against this, because originally the `streamStore` was just a bunch of plain files on disk (not SQLite), so we record both the starting point and the ending point of the transcript *in the kvStore*. That way, a restart after an abandoned crank will only read a subset of the recorded transcript, and will ignore the over-optimistically-committed entry.

But, it'd be nice to do this properly. That actually requires a bit more, which I don't know how to do right now: if the crank is aborted, we should go back to `streamStore` and ask it to remove the transcript entry we just wrote (do a SQL `DELETE`). It might get overwritten by a replacement entry, or the block might end without another delivery being made to that vat, or (most likely) the vat is getting terminated so the whole transcript will get deleted too. However I think we'll be ok without the delete: the only weird case if if the block ends without another delivery being made to that vat, and that will result in the spurious entry being committed along with everything else in the block, and we'll rely upon our kvStore transcript-end counter to ignore it at next reboot. Eventually the entry will probably be overwritten.

So I think we want `swingStore` to:
* hold something in RAM that remembers 0 or 1 open transactions
* if a write/delete requests arrives and there is not an open transaction, open one
* `commit()` should commit and close the transaction, if any, always leaving us with 0 open transactions

Then we should use `strace` to watch a node and make sure the number of `fsync` calls is lower: only one per block.

cc @dckc

warner:
The 'streamStore' component of swing-store is used to hold the
append-only delivery transcript for each vat. It stores each record in
a SQLite table.

Previously, streamStore used SQLite's auto-commit API: every write was
committed immediately. This causes a significant performance hit,
especially on slower disks, because a DB commit must do 3-ish
`fsync()` calls. SSDs are fast, but we observed 12.8ms stalls on the
standard disks of a popular cloud CPU provider, for each call to
writeStreamItem().

For the SwingSet client, each vat delivery triggered this stall, and a
full chain block could have hundreds of deliveries, leading to
multiple seconds of delay.

This patch changes streamStore to create an explicit DB transaction
instead of relying on automatic ones. SQLite will not commit these
transactions automatically, so the patch also introduces internal
`streamStore.commit()` and `.close()` methods.

Clients should continue to use `swingStore.commit()` and
`swingStore.close()` as usual; they will invoke the internal
streamStore methods as appropriate. The three patterns are:

* `commit()`: commit changes, keep using the DB
* `commit(); close();`: commit changes, close the DB
* `close();`: abandon the changes, close the DB

The streamStore is committed before the kvStore (LMDB) commit. As
before, we rely upon clients recording the startPos/endPos of the
stream into the kvStore, to safely tolerate a crash just after the
streamStore.commit but before the kvStore.commit. The next restart
after the crash will contain the spurious stream item in streamStore,
but the old startPos/endPos in kvStore, so clients will effectively
ignore the leftover item.

closes #6056

### DIFF
--- a/packages/swing-store/src/sqlStreamStore.js
+++ b/packages/swing-store/src/sqlStreamStore.js
@@ -67,10 +67,8 @@ export function sqlStreamStore(dbDir, io) {
   // https://sqlite.org/lang_transaction.html , especially section 2.2
 
   function ensureTransaction() {
-    // @ts-expect-error Database really does have .inTransaction:boolean
     if (!db.inTransaction) {
       db.prepare('BEGIN IMMEDIATE TRANSACTION').run();
-      // @ts-expect-error Database really does have .inTransaction:boolean
       assert(db.inTransaction);
     }
   }
@@ -163,7 +161,6 @@ export function sqlStreamStore(dbDir, io) {
   };
 
   const commit = () => {
-    // @ts-expect-error Database really does have .inTransaction:boolean
     if (db.inTransaction) {
       db.prepare('COMMIT').run();
     }
--- a/packages/swing-store/src/sqlStreamStore.js
+++ b/packages/swing-store/src/sqlStreamStore.js
@@ -54,6 +54,27 @@ export function sqlStreamStore(dbDir, io) {
     )
   `);
 
+  // We use explicit transactions to 1: not commit writes until the
+  // host application calls commit() and 2: avoid expensive fsyncs
+  // until the appropriate commit point. All API methods should call
+  // this first, otherwise sqlite will automatically start a
+  // transaction for us, but it will commit/fsync at the end of the DB
+  // run(). We use IMMEDIATE because the kernel is supposed to be the
+  // sole writer of the DB, and if some other process is holding a
+  // write lock, I'd like to find out earlier rather than later. We do
+  // not use EXCLUSIVE because we should allow external *readers*, and
+  // we might decide to use WAL mode some day. Read all of
+  // https://sqlite.org/lang_transaction.html , especially section 2.2
+
+  function ensureTransaction() {
+    // @ts-expect-error Database really does have .inTransaction:boolean
+    if (!db.inTransaction) {
+      db.prepare('BEGIN IMMEDIATE TRANSACTION').run();
+      // @ts-expect-error Database really does have .inTransaction:boolean
+      assert(db.inTransaction);
+    }
+  }
+
   const streamStatus = new Map();
 
   /**
@@ -75,6 +96,7 @@ export function sqlStreamStore(dbDir, io) {
     );
 
     function* reader() {
+      ensureTransaction();
       const query = db.prepare(`
         SELECT item
         FROM streamItem
@@ -123,6 +145,7 @@ export function sqlStreamStore(dbDir, io) {
       X`can't write stream ${q(streamName)} because it's already in use`,
     );
 
+    ensureTransaction();
     db.prepare(
       `
       INSERT INTO streamItem (streamName, item, position)
@@ -140,10 +163,15 @@ export function sqlStreamStore(dbDir, io) {
   };
 
   const commit = () => {
-    // We use the sqlite3 auto-commit API: every command automatically starts
-    // a new transaction if one is not already in effect, and every
-    // automatically-started transaction is committed when the last SQL
-    // statement finishes. https://sqlite.org/lang_transaction.html
+    // @ts-expect-error Database really does have .inTransaction:boolean
+    if (db.inTransaction) {
+      db.prepare('COMMIT').run();
+    }
+  };
+
+  const close = () => {
+    // close without commit is abort
+    db.close();
   };
 
   return harden({
@@ -151,6 +179,7 @@ export function sqlStreamStore(dbDir, io) {
     readStream,
     closeStream,
     commit,
+    close,
     STREAM_START,
   });
 }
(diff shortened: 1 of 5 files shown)


warner:
@dckc I don't know how to address the type complaint that lint found: it's unhappy that the `Database` type doesn't claim to have a `.inTransaction` property, but both the `better-sqlite3` docs and the implementation offer it. I can't find evidence of types within `node_modules/better-sqlite3/`, and the `inTransaction` property is defined as a getter in `better-sqlite3/lib/methods/wrappers.js` (maybe the type system is smart enough to find methods in the implementation, but not getters?). Any ideas?

packages/swing-store/src/sqlStreamStore.js
```diff
@@ -54,6 +54,19 @@ export function sqlStreamStore(dbDir, io) {
     )
   `);
 
+  // We use explicit transactions to 1: not commit writes until the
+  // host application calls commit() and 2: avoid expensive fsyncs
+  // until the appropriate commit point. All API methods should call
+  // this first, otherwise sqlite will automatically start a
+  // transaction for us, but it will commit/fsync at the end of the DB
+  // run(). https://sqlite.org/lang_transaction.html
+  function maybeStartTransaction() {
```
dckc: "maybe" suggests "might fail", not "might be already done", to me. Consider "ensure"

dckc:
2 ideas:
 1. @ts-expect-error with a concise comment
 2. Find some place to use `Database & { inTransaction: boolean }`

dckc (approved):

warner:
> 2 ideas:
> 
>     1. @ts-expect-error with a concise comment
> 
>     2. Find some place to use `Database & { inTransaction: boolean }`

@mhofman figured it out for me: import the types published separately in `@types/better-sqlite3`.

Issue #6022, opened 2022-08-22
XS behavior divergence during CI


warner:
## Describe the bug

Yesterday's CI run (on https://github.com/Agoric/agoric-sdk/pull/6011 to pull in the latest XS release) experienced a failure inside the deployment-test, in which two validators behaved differently:

https://github.com/Agoric/agoric-sdk/runs/7955587167?check_suite_focus=true

The `diff` reads:
```
--- /dev/fd/63	2022-08-22 16:06:31.727630864 +0000
+++ /dev/fd/62	2022-08-22 16:06:31.723630598 +0000
@@ -48679,2 +48679,2 @@
-set ko345.refCount 4,4
-set acceptanceQueue.1159 {"type":"send","target":"ko345","msg":{"methargs":{"bo
+set ko249.refCount 4,4
+set acceptanceQueue.1159 {"type":"send","target":"ko249","msg":{"methargs":{"bo
@@ -48706,2 +48706,2 @@
-set ko249.refCount 4,4
-set acceptanceQueue.1162 {"type":"send","target":"ko249","msg":{"methargs":{"bo
+set ko213.refCount 4,4
+set acceptanceQueue.1162 {"type":"send","target":"ko213","msg":{"methargs":{"bo
@@ -48731,2 +48731,2 @@
-set ko213.refCount 4,4
-set acceptanceQueue.1165 {"type":"send","target":"ko213","msg":{"methargs":{"bo
+set ko345.refCount 4,4
+set acceptanceQueue.1165 {"type":"send","target":"ko345","msg":{"methargs":{"bo
@@ -48771 +48771 @@
-set activityhash c2336968be5f3bce3752df7e5d3df32727f8758896005123b0988ec26e233e
+set activityhash 388f6454f586010f36a755c2d86aedebc992e3681041c18e74398b03aec680
@@ -48773 +48773 @@
-set runQueue.1111 {"type":"send","target":"ko345","msg":{"methargs":{"body":"[\
+set runQueue.1111 {"type":"send","target":"ko249","msg":{"methargs":{"body":"[\
@@ -48780 +48780 @@
-set activityhash 68be83bebe01e953e65cea9299a0def4deaa465e89341f525c35e7f8fef215
+set activityhash 33fdf8b639e53a4dd92e9f70d3bfd696d6872de2e5492a29cc72486fb87099
@@ -48789 +48789 @@
[...]

PR #6011, opened 2022-08-19
fix(xsnap): update to latest XS release

warner:
This provides the upstream version of the snapshot-write memory leak
fix, along with other bugfixes.

closes #5975

PR #6063, opened 2022-08-26
Fix process-integration-results.sh

mhofman:
refs: #6022 

## Description

Updates the processing script to handle some more cases:
- Run without error if no snapshot were found to diverge
- Remove new time/stats fields from slog for diffing
- Generate an artifact when xsnap traces are found to diverge (to diagnose issues like in #6022)

### Security Considerations

None, manual debug scripts

There should be some

### Testing Considerations

Generated artifacts for these divergences in #6011 tests and on a local run of the same branch.

### DIFF
--- a/scripts/process-integration-results.sh
+++ b/scripts/process-integration-results.sh
@@ -17,7 +17,7 @@ clean_slog() {
 
 "$thisdir/process-integration-swingstore-traces.sh" "$RESULTSDIR"
 
-[ -f "$RESULTSDIR/divergent_snapshots" ] || exit 0
+# [ -f "$RESULTSDIR/divergent_snapshots" ] || exit 0
 
 cd "$RESULTSDIR"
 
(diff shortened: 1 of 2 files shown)


mhofman:
Best reviewed commit-by-commit as there is a refactor (move lines around) which clobbers the diff.

scripts/process-integration-results.sh
```diff
@@ -11,6 +11,10 @@ RESULTSDIR=${RESULTSDIR-"$NETWORK_NAME/results"}
 
 [ $# -gt 0 ] && RESULTSDIR="$1"
 
+clean_slog() {
+  jq -cr 'del(.time, .monotime, .dr[2].timestamps, .memoryUsage, .heapStats, .statsTime)'
```
warner, line 15: be on the lookout for some `kernelStats` that @FUDCo will be adding to one of the cosmic-swingset `end-block` slog records, maybe next week, I think some of them are non-consensus

warner (approved):
looks good, thanks

scripts/process-integration-results.sh
mhofman: Yeah but I believe all of them are actually identical. Regardless they are already written to the swingstore and there is no diff showing up so far ;)

PR #6068, opened 2022-08-26
fix(vats): chainStorage vat writes directly to device

### LINKED ISSUE #6053: chainStorage indirects thru bootstrap vat for bridge device
## Describe the bug

Not only does the walletFactory write its state 17 times per PSM trade (#6038 ), but each one goes thru both the `chainStorage` vat (v8) and the bootstrap vat (v1) on its way to golang and the IAVL tree:

image

## To Reproduce

1. start a local-chain with a SLOG_FILE
2. do a PSM trade
3. look at `setValue` and `toBridge` deliveries in the SLOG_FILE

## Expected behavior

The `chainStorage` vat has 1 job; we might as well give it the bridge device so that it can do that job by itself, without going thru the bootstrap vat.

## Platform Environment

ac01a326c (dc-boot-psm) chore(vats): keep smartWalletStartResult in bootstrap space

cc @warner @michaelfig

dckc:
closes: #6053

### Security Considerations

Device access is a power that the chainStorage vat didn't previously have. I believe granting it is entirely appropriate.

`manifest.js` is updated with the required bootstrap powers.

### Testing Considerations

Tested manually; resulting before and after sequence diagrams are below.

### DIFF
--- a/packages/vats/src/core/chain-behaviors.js
+++ b/packages/vats/src/core/chain-behaviors.js
@@ -290,16 +290,16 @@ export const makeBridgeManager = async ({
 harden(makeBridgeManager);
 
 /**
- * @param {BootstrapPowers & {
+ * @param {BootDevices<ChainDevices> & BootstrapSpace & {
  *   consume: { loadVat: ERef<VatLoader<ChainStorageVat>> }
  * }} powers
  */
 export const makeChainStorage = async ({
-  consume: { bridgeManager: bridgeManagerP, loadVat },
+  devices: { bridge },
+  consume: { loadVat },
   produce: { chainStorage: chainStorageP },
 }) => {
-  const bridgeManager = await bridgeManagerP;
-  if (!bridgeManager) {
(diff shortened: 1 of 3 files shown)


dckc:
@michaelfig you noted a security concern that bridge device access would let the chainStorage vat write to parts of swingset storage that it shouldn't be able to. So there's a question of whether attenuating that access is worth the extra round-trip.

dckc:
This also seems to fix a bug in that `makeChainStorageRoot` seems to assume the `handleStorageMessage` argument is a _synchronous_ function but we were passing an async function.

visualizations built with [slog-to-diagram.mjs](https://gist.github.com/dckc/7da293f2fcf10d1a4853dc680ea3f667#file-slog-to-diagram-mjs)

<details>
<summary>before and after sequence diagrams</summary>

#### before this fix

963b652c6

slog

#### after this fix

b22341916

slog
</details>

warner (approved):
That looks good to me.

PR #6070, opened 2022-08-26
refactor(store)!: move from Shape to Schema terminology

erights:
As explained in https://github.com/Agoric/agoric-sdk/blob/master/packages/store/src/types.js#L56-L74 , when talking about patterns, we need an additional suffix besides "Pattern". But we only need one. Before this PR we fell into using two: "Schema" and "Shape". Either one is a decent choice. This PR switches all those uses to "Schema". The old `assertProposalShape` was not using patterns, but rather an ad-hoc specialized pattern-like encoding that we wish to deprecate, so it is *not* renamed to `assertProposalSchema`.

### DIFF
--- a/packages/ERTP/src/displayInfo.js
+++ b/packages/ERTP/src/displayInfo.js
@@ -3,7 +3,7 @@
 import { assert, details as X } from '@agoric/assert';
 import { fit } from '@agoric/store';
 
-import { DisplayInfoShape } from './typeGuards.js';
+import { DisplayInfoSchema } from './typeGuards.js';
 
 /**
  * @param {AdditionalDisplayInfo} allegedDisplayInfo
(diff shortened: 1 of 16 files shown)


turadg:
> Either one is a decent choice. This PR switches all those uses to "Schema".

Not sure if this is part of what is open to review here, but I don't think the choices are equally good. IIUC the suffix is being used just as a word to not repeat "Pattern" in the values of that type (and avoid Hungarian notation). I agree that we shouldn't repeat the type (e.g. `NatValuePattern`. I think there is a case to be made for a more neutral suffix than either, e.g. `NatValue` and rely on the type to convey that it's a pattern.

If that doesn't fly, I'd at least suggest that `Shape` is less laden than `Schema`. Seeing the latter makes me think there's some kind of schema language to understand, distinct from the Pattern thing I might already be familiar with.

So fwiw, I'd rank the suffix options:
1. ''
2. 'Shape'
3. 'Schema'


Separately, the conventional commit here is `refactor;` not `fix:` because there's no bug being resolved. It might also need `!` if this changes public exports. (If the conventional commits are too burdensome, imo it's better to omit it than convey inaccurate changelog)

erights:
@turadg for my own sanity, I know I need one or the other. Having tried on both, I prefer "Schema" but can live with "Shape", so am tossing https://github.com/Agoric/agoric-sdk/pull/6072 into the ring as an alternative to this PR.

Reviewers, which do you prefer?

erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/6072

PR #6072, opened 2022-08-26
refactor(store)!: move from Schema to Shape terminology

erights:
As explained in https://github.com/Agoric/agoric-sdk/blob/master/packages/store/src/types.js#L56-L74 , when talking about patterns, we need an additional suffix besides "Pattern". But we only need one. Before this PR we fell into using two: "Schema" and "Shape". Either one is a decent choice. This PR switches all those uses of "Schema" to "Shape". 

The old `assertProposalShape` was not using patterns, but rather an ad-hoc specialized pattern-like encoding so we deprecate it. We needed to deprecate it anyway because we all need to switch to using the optional `proposalShape` parameter of `makeInvitation` instead. Unlike `assertProposalShape`, the `proposalShape` passed to `makeInvitation` is checked by Zoe. An offer whose proposal does not fit that shape is rejected by Zoe and the contract never even hears about it. In addition, one can ask Zoe about the `proposalShape` of a given invitation, so they can check for themselves before submitting an offer.

Note https://github.com/Agoric/agoric-sdk/pull/6070#issuecomment-1228940913 for some of the reasons we chose "Shape" over "Schema". Internal to SwingSet there were other uses of "schema" that were not describing patterns. This PR does not rename them, clarifying that they name something different. Note that sometimes (e.g. in gc-helper.js) "schema" is used for a stringified pattern, which is related but still must not be confused with an actual pattern.


packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -123,10 +123,12 @@ export const fitProposalShape = (seat, proposalShape) =>
  * null contents. If the client submits an offer which does not match
  * these expectations, the seat will be exited (and payments refunded).
  *
+ * @deprecated Use optional `proposalShape` argument to `makeInvitation` with
+ * a genuine pattern.
  * @param {ZCFSeat} seat
  * @param {ExpectedRecord} expected
  */
-export const assertProposalShape = (seat, expected) => {
+export const assertProposalShapeDeprecated = (seat, expected) => {
```
turadg: imo the `@deprecated` annotation is sufficient since TS-aware IDE will render `assertProposalShape` in a strikeout font. But not opposed to this renaming.

packages/zoe/src/zoeService/offer/offer.js
```diff
@@ -55,9 +55,9 @@ export const makeOfferMethod = (
     instanceAdmin.assertAcceptingOffers();
 
     const proposal = cleanProposal(uncleanProposal, getAssetKindByBrand);
-    const proposalSchema = getProposalSchemaForInvitation(invitationHandle);
-    if (proposalSchema !== undefined) {
-      fit(proposal, proposalSchema, 'proposal');
+    const proposalShape = getProposalSchemaForInvitation(invitationHandle);
```
turadg: should this also be `getProposalShape`?

turadg (approved):
👍

erights:
Reviewers, this an #6070 are alternatives. Which do you prefer?

packages/zoe/src/zoeService/offer/offer.js
erights: Good catch! Doing a deeper search now.

FUDCo:
> Reviewers, this an #6070 are alternatives. Which do you prefer?

I like "schema" better.

turadg:
> > Reviewers, this an #6070 are alternatives. Which do you prefer?
> 
> I like "schema" better.

I hope this isn't resolved by a tally of votes of personal preference. This open source codebase is an SDK that will be used by many people who aren't themselves reviewers. We should design the APIs by evaluating what will serve them best, at least as we can intuit without their input.

packages/zoe/src/zoeService/offer/offer.js
erights: Done. Optimistically resolving, but PTAL

erights:
> We should design the APIs by evaluating what will serve them best, at least as we can intuit without their input.

Sure, but we're still doing the evaluating. I'm interpreting our stated opinions as differing intuitions about what would serve them best. I think we need a discussion in which we air the basis for our differing opinions. Does Monday afternoon work? I'll send out a calendar invite.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Done

PR #6073, opened 2022-08-26
fix(telemetry): Use transactions for tmp telemetry DB

mhofman:
refs: #6056

## Description

Solves a performance issue where automatic transactions compounded to cause noticeable stalls.
Similar to https://github.com/Agoric/agoric-sdk/pull/6058 but for the telemetry SQLite DB.

### DIFF
--- a/packages/telemetry/src/kv-string-store.js
+++ b/packages/telemetry/src/kv-string-store.js
@@ -22,6 +22,21 @@ export const makeTempKVDatabase = io => {
   return db;
 };
 
+export const makeKVDatabaseTransactionManager = db => {
+  return {
+    begin: () => {
+      assert(!db.inTransaction);
+      db.prepare('BEGIN IMMEDIATE TRANSACTION').run();
+      assert(db.inTransaction);
+    },
+    end: () => {
+      assert(db.inTransaction);
+      db.prepare('COMMIT').run();
+      assert(db.inTransaction);
+    },
+  };
+};
+
 /**
  * @param {string} kind
  * @param {sqlite3ambient.Database} [db]
--- a/packages/telemetry/src/kv-string-store.js
+++ b/packages/telemetry/src/kv-string-store.js
@@ -32,7 +32,7 @@ export const makeKVDatabaseTransactionManager = db => {
     end: () => {
       assert(db.inTransaction);
       db.prepare('COMMIT').run();
-      assert(db.inTransaction);
+      assert(!db.inTransaction);
     },
   };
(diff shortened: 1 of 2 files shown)


mhofman:
Probably not the right level of abstraction, but I'm not really willing to refactor things more than this.

warner (approved):
Sounds good to me.. thanks for remembering that we had another SQLite DB around.

I'm a bit weirded out that our telemetry is doing this much correlation work in-line. I'd like to revisit that and see if it's really pulling its weight.

To estimate what sort of load this used to represent.. it looks like we'd get one DB commit per slog `syscall.send` event (as long as a `result` vpid was provided), right?

mhofman:
> Sounds good to me.. thanks for remembering that we had another SQLite DB around.

Credit to @arirubinstein for noticing

> I'm a bit weirded out that our telemetry is doing this much correlation work in-line. I'd like to revisit that and see if it's really pulling its weight.

Yeah I was a little surprised as well, and we probably should revisit how this is implemented, but this correlation can be beneficial when looking at traces on Honeycomb.

> 
> To estimate what sort of load this used to represent.. it looks like we'd get one DB commit per slog `syscall.send` event (as long as a `result` vpid was provided), right?

Correct. Maybe not as bad as every crank, but realistically there are quite a few send in each block.

warner:
Looks like this derailed somehow.. I'll force-push a rebased version to resolve at least one of the problems.

packages/telemetry/src/kv-string-store.js
```diff
@@ -22,6 +22,21 @@ export const makeTempKVDatabase = io => {
...
+  return {
+    begin: () => {
+      assert(!db.inTransaction);
+      db.prepare('BEGIN IMMEDIATE TRANSACTION').run();
+      assert(db.inTransaction);
+    },
+    end: () => {
+      assert(db.inTransaction);
+      db.prepare('COMMIT').run();
+      assert(db.inTransaction);
```
michaelfig: Is this wrong?

```suggestion
      assert(!db.inTransaction);
```

michaelfig (commented):
It looks like the transaction flag is being asserted when it isn't set.

packages/telemetry/src/kv-string-store.js
warner: you're absolutely right.. I pushed a patch to fix it

mhofman: Yikes 😞

PR #6076, opened 2022-08-27
refactor(store)!: fix bad check defaults

erights:
There were a few places where I was calling `checkKey` with one argument when I should have been calling it with two. This bug survived because it was being another problems, which this PR fixes:

My practice of the isFoo/assertFoo/checkFoo pattern was to have the `check` parameter of the `checkFoo` functions be optional, and to default to `x => x`. This means that my failure to call it with two arguments when I should have turned into the equivalent of calling it with a `x => x`, turning the call from acting like an assert into acting like a predicate.

Another problem noticed and fixed in the process: When an `assertFoo` function calls a `checkFoo` function with `assertChecker`, then the `checkFoo` function will either return true or throw. However, the `assertFoo` function should either return silently or throw. But some of these `assertFoo` functions forgot to put the call to `checkFoo` in curly braces.

See also https://github.com/endojs/endo/pull/1274

Although these are similar PRs against both our repos, there is no dependence between them. Whew!

---

Separately, prior to this PR `checkKey` could be mislead by a tagged record that obeyed the marshal-level invariants of Tagged but, for example for a tag of `copySet`, did not obey the store-level invariants of CopySet. This led `checkKey` to thinking that a non-CopySet was a valid key, which it is not. The hazard was the use of `keyOf` to get the indicator without the invariant check.

Removes `keyOf`. Enhances `checkKey` to call the necessary invariant checkers itself.

### DIFF
--- a/packages/store/src/keys/checkKey.js
+++ b/packages/store/src/keys/checkKey.js
@@ -26,10 +26,10 @@ const { ownKeys } = Reflect;
 
 /**
  * @param {Passable} val
- * @param {Checker=} check
+ * @param {Checker} check
  * @returns {boolean}
  */
-const checkPrimitiveKey = (val, check = x => x) => {
+const checkPrimitiveKey = (val, check) => {
   if (isObject(val)) {
     return check(false, X`A ${q(typeof val)} cannot be a primitive: ${val}`);
   }
@@ -43,7 +43,7 @@ const checkPrimitiveKey = (val, check = x => x) => {
  * @param {Passable} val
  * @returns {boolean}
  */
-export const isPrimitiveKey = val => checkPrimitiveKey(val);
+export const isPrimitiveKey = val => checkPrimitiveKey(val, x => x);
 harden(isPrimitiveKey);
 
 /**
@@ -57,10 +57,10 @@ harden(assertPrimitiveKey);
 
 /**
  * @param {Passable} val
- * @param {Checker=} check
+ * @param {Checker} check
  * @returns {boolean}
  */
-export const checkScalarKey = (val, check = x => x) => {
+export const checkScalarKey = (val, check) => {
   if (isPrimitiveKey(val)) {
     return true;
   }
@@ -75,7 +75,7 @@ export const checkScalarKey = (val, check = x => x) => {
  * @param {Passable} val
  * @returns {boolean}
  */
-export const isScalarKey = val => checkScalarKey(val);
+export const isScalarKey = val => checkScalarKey(val, x => x);
 harden(isScalarKey);
 
 /**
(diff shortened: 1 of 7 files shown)


packages/store/src/keys/checkKey.js
```diff
@@ -161,7 +161,7 @@ export const checkCopySet = (s, check = x => x) => {
       X`Not a copySet: ${s}`,
     ) &&
     checkElements(s.payload, check) &&
-    checkKey(s.payload);
+    checkKey(s.payload, check);
```
erights, line 164: The omission of the `check` argument here (and in `checkCopyBag` below) were the bugs being masked by the other problems fixed in this PR.

packages/store/src/keys/checkKey.js
```diff
@@ -247,7 +247,7 @@ export const checkCopyBag = (b, check = x => x) => {
       X`Not a copyBag: ${b}`,
     ) &&
     checkBagEntries(b.payload, check) &&
-    checkKey(b.payload);
+    checkKey(b.payload, check);
```
erights, line 250: This is the other one.

turadg: Thanks for pointing to this. Not a blocker but a clean way here to factor commits for the master history (and review) would be to separate the refactor from the fix:

- refactor(checkKey)!: require check param
- fix(checkKey): include missing check param

packages/store/src/keys/copyBag.js
```diff
@@ -15,22 +15,18 @@ const { details: X } = assert;
 /**
  * @template T
  * @param {[T,bigint][]} bagEntries
- * @param {FullCompare=} fullCompare If provided and `bagEntries` is already
+ * @param {FullCompare | undefined} fullCompare If provided and `bagEntries` is already
```
turadg, line 18: I see this parameter now must be specified, due to non-optional `check`. `undefined` in param list suggests omission. Makes me think of `null` to specify absence. Not worth it here but curious your thoughts on that.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -539,27 +543,51 @@ const makePatternKit = () => {
...
+    );
+    /** @type {Kind} */
+    let specimenKind = passStyleOf(specimen);
+    if (specimenKind !== 'tagged') {
+      if (specimenKind === kind) {
+        // passStyleOf already did all the invariant checking.
+        return true;
+      }
+    } else {
+      specimenKind = getTag(specimen);
```
turadg, line 568: this is new logic I don't see noted in the PR description

turadg (approved):

packages/store/src/keys/copyBag.js
erights: In general I find JS's two bottoms a pain and rarely use null. This one is a plausible case. Trying it though, I find that it composes less pleasantly. In copySet.js 
```js
export const assertNoDuplicates = (elements, fullCompare = undefined) => {
  checkNoDuplicates(elements, fullCompare, assertChecker);
};
```
Making this code correct again after this change is simple: `fullCompare = null` in the parameter list. But I think it is more fragile.

I'm on the fence. Proceeding though with the current state.

packages/store/src/patterns/patternMatchers.js
erights: Good catch. Done.

PR #6077, opened 2022-08-27
6044 wallet voting

### LINKED ISSUE #6044: smart wallet applyMethod (e.g. for voting)

DC new one: applyMethod for voting
Oracles exercise their powers by getting an invitation … offer result, on which they can submit price info
Committees work similarly. Get an offer result with a widget that lets them submit votes.

Similar work to https://github.com/Agoric/agoric-sdk/pull/5974 but need this part earlier to unblock contract review. Also this contract-side code needs to be production-ready whereas the client just needs to work

## Description of the Design

(Need to use the Smart Wallet for this because we’re turning off agSolo, due to security risks.)

regular action, not spendAction because we don't want it to be possible to unmarshal a reference to a purse. (no need to do a withdrawal while voting).

## Security Considerations

- reject `spendAction`, voting should never do a withdrawal
- should we restrict method names (by length or allowlist) ?  will we be prepared to upgrade when there's another use case?
- contracts should not be able to forge references to offer results (i.e. offerResults belong in a private table of an export context)

## Test Plan

Try to make a committee and vote on it through a wallet

turadg:
closes: #6044 
refs: #XXXX

### DIFF
--- a/packages/governance/src/committee.js
+++ b/packages/governance/src/committee.js
@@ -18,6 +18,12 @@ import { QuorumRule } from './question.js';
 const { ceilDivide } = natSafeMath;
 
 /**
+ * @typedef {{
+ * castBallotFor: (questionHandle: Handle<'Question'>, positions: Position[]) => Promise<void>
+ * }} CommitteeVoter
+ */
+
+/**
  * Each Committee (an Electorate) represents a particular set of voters. The
  * number of voters is visible in the terms.
  *
@@ -46,7 +52,10 @@ const start = (zcf, privateArgs) => {
     );
 
   const makeCommitteeVoterInvitation = index => {
-    /** @type {OfferHandler} */
+    /**
+     * @param {ZCFSeat} seat
+     * @returns {CommitteeVoter}
+     */
     const offerHandler = seat => {
       const voterHandle = makeHandle('Voter');
       seat.exit();
--- a/packages/governance/src/committee.js
+++ b/packages/governance/src/committee.js
@@ -51,6 +51,7 @@ const start = (zcf, privateArgs) => {
       privateArgs.marshaller,
     );
 
+  /** @param {number} index */
   const makeCommitteeVoterInvitation = index => {
     /**
      * @param {ZCFSeat} seat
--- a/packages/wallet/api/src/marshal-contexts.js
+++ b/packages/wallet/api/src/marshal-contexts.js
@@ -243,18 +243,19 @@ const defaultMakePresence = iface => {
 };
 
 /**
- * Make context for unserializing wallet or board data.
+ * Make context for marshalling wallet or board data.
+ * To be used by the client. TODO consider renaming to makeClientContext
  *
  * @param {(iface: string) => unknown} [makePresence]
  */
 export const makeImportContext = (makePresence = defaultMakePresence) => {
   const walletObjects = {
-    /** @type {IdTable<number, unknown>} */
+    /** @type {IdTable<number, Purse>} */
     purse: {
       bySlot: makeScalarMap(),
       byVal: makeScalarMap(),
     },
-    /** @type {IdTable<number, unknown>} */
+    /** @type {IdTable<number, Payment>} */
     payment: {
       bySlot: makeScalarMap(),
       byVal: makeScalarMap(),
@@ -333,6 +334,13 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
   return harden({
     fromMyWallet: Far('wallet marshaller', { ...marshal.fromMyWallet }),
     fromBoard: Far('board marshaller', { ...marshal.fromBoard }),
+    /**
+     * @param {BoardId} id
+     * @param {unknown} val
+     */
+    initBoardId: (id, val) => {
+      initSlotVal(boardObjects, id, val);
+    },
   });
 };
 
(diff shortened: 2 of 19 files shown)


turadg:
@dckc draft with current state. Hope to find time later today to add the tests with a committee vote invitation.

packages/wallet/api/test/test-wallet-marshal.js
```diff
@@ -163,4 +163,19 @@ test('makeExportContext.serialize handles unregistered identites', t => {
...
+    const cap = context.serialize(val);
+    t.deepEqual(cap, {
+      body: JSON.stringify({
+        '@qclass': 'slot',
+        iface: 'Alleged: someOfferResult',
+        index: 0,
+      }),
+      slots: ['offerResult:1'],
+    });
+    t.deepEqual(context.unserialize(cap), val);
```
dckc: it's important that `unserialize` preserves identity, so this test should be `t.is(...)`

packages/wallet/api/test/test-wallet-marshal.js
```diff
@@ -50,7 +50,7 @@ const makeOnChainWallet = board => {
   });
 };
 
-test('makeImportContext preserves identity across AMM and wallet', t => {
```
dckc: I think we still want a test that `makeImportContext` preserves identity across the AMM and wallet; that was its original motivation.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -187,6 +192,8 @@ export const makeExportContext = () => {
    * @param {IdTable<number, V>} table
    */
   const makeSaver = (kind, table) => {
+    // XXX table passed in to simplify type inference
+    assert.equal(table, walletObjects[kind]);
```
dckc, line 196: good idea.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -197,6 +204,7 @@ export const makeExportContext = () => {
   };
 
   return harden({
+    saveOfferResultActions: makeSaver('offerResult', walletObjects.offerResult),
```
dckc, line 207: The `Actions` part of this / these names is a little iffy. hm.

packages/governance/src/committee.js
```diff
@@ -17,6 +17,12 @@ import { QuorumRule } from './question.js';
 
 const { ceilDivide } = natSafeMath;
 
+/**
+ * @typedef {{
+ * castBallotFor: (questionHandle: Handle<'Question'>, positions: Position[]) => Promise<void>
```
dckc, line 22: @Chris-Hibbert I just realized that in order for committee members to be able to put questions as well, the "charter" contract should be a different shape:

Our design for raising debt limits involved using a method on a public facet, which assumed users had access to `zoe` and `E(zoe).getPublicFacet()`.
https://github.com/Agoric/inter-protocol-bootstrap/blob/56de1ed1821e19828a6f74ff10b7fce27b3f2cb7/vote-raise-debt-limit.js#L177

I think life will be simpler if the new charter contract you're building uses methods on an offer result instead of methods on a public facet.

dckc (commented):
Looks like good progress. I hope to start integration testing soonish.

packages/wallet/api/test/test-wallet-marshal.js
turadg: Happy to revert. I thought I was executing your suggestion in https://github.com/Agoric/agoric-sdk/pull/5883#issuecomment-1210060306

Or are you suggesting to keep this change and add something like `test.todo('makeImportContext preserves identity across AMM and wallet')` ?

packages/wallet/api/src/marshal-contexts.js
turadg: Yeah. I opted for consistency with the others, but I don't understand why "Actions" is part of any of them. Cool to remove from all?

packages/wallet/api/test/test-wallet-marshal.js
dckc: oops... that "better phrasing" suggestion wasn't all that clear; it was about the `t.throws()` message, not the `test()` description.

packages/wallet/api/src/marshal-contexts.js
dckc: I think so. @michaelfig ?

are the other 2 used? @samsiegart ?

packages/governance/src/committee.js
Chris-Hibbert: > ... if the new charter contract you're building ...

There isn't a task corresponding to that description on my radar. Can we make an issue?
I presume we want a variant of econCommitteeCharter that serves PSM only. If the capability to invoke votes is carried by an offer result, should the various committee members each invoke a separate invitation, or is an invitation exercised once in bootstrap, and the offer result handed out from there?

dckc: > Can we make an issue?

Yes, we can; I more or less expected that was part of "Do The Right Thing" that you said what you had what you need for. The capability to vote is taken care of by the committee contract. My concern here is for posing questions. You're right that this raises a question of how the invitations get to the committee members. I suppose we can use the same mechanism as we did for the voting invitations, but it's worth some thought / discussion.

turadg:
For now, achieving voting with continuing invitations in https://github.com/Agoric/agoric-sdk/pull/6084

PR #6078, opened 2022-08-27
fix(wallet/api): prune inactive historical offers

### LINKED ISSUE #6065: smart wallet state grows with each offer
## Describe the bug

The published state includes a list of all offers, past and present.

## Expected behavior
that the offer is removed when:
- the offerResult does not support continuing offers (all PSM offers, but not vault creation)
- or the offerResult is an error (since you cannot continue that)
- then once the offer is exited, the state update is reported via subscription and the offer is deleted from the wallet

cc @warner @samsiegart @turadg @dtribble @michaelfig

dckc:
we can get to them from chainStorage

closes: #6065
refs: #XXXX

### DIFF
--- a/packages/wallet/api/src/lib-wallet.js
+++ b/packages/wallet/api/src/lib-wallet.js
@@ -451,6 +451,40 @@ export function makeWalletRoot({
     return proposalForDisplay;
   };
 
+  /**
+   * Delete inactive offers.
+   *
+   * @param {string} id
+   */
+  const pruneOfferWhenInactive = id => {
+    assert(inboxState.has(id));
+    const offer = inboxState.get(id);
+    const { status } = offer;
+
+    switch (status) {
+      case 'decline':
+      case 'cancel':
+      case 'rejected':
+        inboxState.delete(id);
+        return;
+      case 'accept':
+        assert(idToOfferResultPromiseKit.has(id));
+        void idToOfferResultPromiseKit.get(id).promise.then(result => {
+          const style = passStyleOf(result);
+          const active =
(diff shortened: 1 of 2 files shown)


dckc:
@samsiegart I think it shouldn't be too hard to update the wallet UI to handle this; I was able to do it for the `psm-tool` CLI without too much trouble:

```
const getWalletState = async (addr, ctx, { getJSON }) => {
  const txt = await vstorage.read(`published.wallet.${addr}`, getJSON);
  /** @type {{ purses: PurseState[], offers: OfferDetail[] }[]} */
  const states = storageNode.unserialize(txt, ctx);
  const offerById = new Map();
  states.forEach(state => {
    const { offers } = state;
    offers.forEach(offer => {
      const { id } = offer;
      offerById.set(id, offer);
    });
  });
  const { purses } = states.at(-1) || dieTrying();
  return { purses, offers: [...offerById.values()] };
};
```

dckc:
@michaelfig fixing the tests wasn't such a big deal after all. Please take a look.

packages/wallet/api/src/lib-wallet.js
```diff
@@ -451,6 +451,40 @@ export function makeWalletRoot({
     return proposalForDisplay;
   };
 
+  /**
+   * Delete inactive offers.
```
turadg:
```suggestion
   * Delete offer once inactive.
```
Maybe also worth explaining that this is idempotent. It can (and will?) be called multiple times for the same id.

turadg (approved):
Not super familiar with this part of code but it looks sane

PR #6079, opened 2022-08-27
fix: ALWAYS default to safe

erights:
Configuration options should ***ALWAYS*** default to a safe setting, and require an explicit override to be unsafe.

See https://github.com/Agoric/agoric-sdk/pull/5875/files?diff=split&w=1#r944032788

This PR only fixes one of the two cases I found of `NODE_ENV`. I commented the other, to not lose it. Please advise on how best to fix the commented one. Thanks.

See https://github.com/Agoric/dapp-card-store/pull/58 and https://github.com/Agoric/dapp-fungible-faucet/pull/52

### DIFF
--- a/packages/web-components/README.md
+++ b/packages/web-components/README.md
@@ -45,7 +45,7 @@ environment that `agoric-wallet-connection` uses:
 
 ```js
 // Ensure this is imported before anything else in your project.
-import './install-ses-lockdown.js';
+import './install-demo-ses-lockdown.js';
 ```
 
 Or, in your `index.html`:
@@ -112,7 +112,7 @@ This is an example of how to use the wallet connection in plain HTML:
 <button>Connect to Wallet</button>
 
 <script type="module">
-  import './install-ses-lockdown.js';
+  import './install-demo-ses-lockdown.js';
   import '@agoric/wallet-connection/agoric-wallet-connection.js';
 
   // Set up event handlers.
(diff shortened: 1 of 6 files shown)


packages/web-components/demo/install-ses-lockdown.js
```diff
@@ -2,7 +2,7 @@
 import '@endo/init/pre-remoting.js';
 
 lockdown({
-  errorTaming: 'unsafe',
+  errorTaming: 'unsafe', // TODO Wen safe?
```
samsiegart: This is just meant to be run locally, I think it's always "dev" mode, just a demo page so unsafe is probably okay.

samsiegart (approved):

packages/web-components/demo/install-ses-lockdown.js
erights: I renamed and commented to make the implied unsafety of this file clearer. Resolving.

packages/web-components/README.md
```diff
@@ -45,7 +45,7 @@ environment that `agoric-wallet-connection` uses:
 
 ```js
 // Ensure this is imported before anything else in your project.
-import './install-ses-lockdown.js';
+import './install-demo-ses-lockdown.js';
```
samsiegart: We probably just want to remove this whole section, it's not a great example. Perhaps we should just link to https://github.com/Agoric/dapp-psm/blob/main/src/installSesLockdown.ts as an example (oh, looks like we need to update the `errorTaming` there too).

samsiegart (changes requested):

samsiegart (approved):

PR #6080, opened 2022-08-27
refactor(store)!: split `provide` into collision vs no-collision variants

erights:
#5875 would enable us to diagnose provide collisions, once we have support for turning on this instrumentation only during development. (The instrumentation provides an unsafe global communications channel, and so cannot be on in production.)

In order to do so, it must distinguish between `provide` calls on baggage that should obey the no-collision rule vs other uses on baggage or any map, for example for lazy initialization, where a no-collision rule would be inappropriate. With this PR, the lazy initialization use is supported by `provideLazy`. The stricter `provide`, being meaningful only wrt baggage, is moved from `@agoric/store` to `@agoric/vat-data`. Because we're moving it up the dependency graph, we cannot easily leave behind an adapter in `@agoric/store` that forwards to the new one, so this PR pays the compat cost instead.

Even without that instrumentation, this separation of intent is useful, so this PR extracts that portion so it can be merged first.

Completely separately, assert/src/type.js comment whitespace has been a perpetual annoyance, as `yarn lint-fix` collapses newlines between separate `@callbacks` needed for readability. This PR changes the comments to be resistant to disruption by `yarn lint-fix`.

### DIFF
--- a/packages/inter-protocol/scripts/deploy-contracts.js
+++ b/packages/inter-protocol/scripts/deploy-contracts.js
@@ -23,7 +23,7 @@ const contractRoots = contractRefs.map(ref =>
 );
 
 /** @type {<T>(store: any, key: string, make: () => T) => Promise<T>} */
-const provide = async (store, key, make) => {
+const provideWhen = async (store, key, make) => {
   const found = await E(store).get(key);
   if (found) {
     return found;
@@ -47,7 +47,9 @@ export default async (homeP, endowments) => {
 
   console.log('getting installCache...');
   /** @type {CopyMap<string, {installation: Installation, boardId: string, path?: string}>} */
-  const initial = await provide(scratch, 'installCache', () => makeCopyMap([]));
+  const initial = await provideWhen(scratch, 'installCache', () =>
+    makeCopyMap([]),
+  );
   console.log('initially:', initial.payload.keys.length, 'entries');
 
   // ISSUE: getCopyMapEntries of CopyMap<K, V> loses K, V.
--- a/packages/store/src/stores/store-utils.js
+++ b/packages/store/src/stores/store-utils.js
@@ -88,7 +88,7 @@ export const makeCurrentKeysKit = (
 harden(makeCurrentKeysKit);
 
 /**
- * Call `provide` to get or make the value associated with the key.
+ * Call `provideLazy` to get or make the value associated with the key.
  * If there already is one, return that. Otherwise,
  * call `makeValue(key)`, remember it as the value for
  * that key, and return it.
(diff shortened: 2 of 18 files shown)


packages/inter-protocol/scripts/deploy-contracts.js
```diff
@@ -23,7 +23,7 @@ const contractRoots = contractRefs.map(ref =>
 );
 
 /** @type {<T>(store: any, key: string, make: () => T) => Promise<T>} */
-const provide = async (store, key, make) => {
+const provideWhen = async (store, key, make) => {
```
turadg, line 26: thanks for distinguishing

packages/store/src/stores/store-utils.js
```diff
@@ -99,16 +99,17 @@ harden(makeCurrentKeysKit);
...
   }
   return mapStore.get(key);
 };
-harden(provide);
+harden(provideLazy);
 
 /**
- * Helper for use cases in which the maker function is async. For two provide
+ * Helper for use cases in which the maker function is async. For two
+ * provideLazy
```
turadg: awkward wrapping

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -89,6 +89,41 @@ export const partialAssign = (target, source) => {
...
+ * baggage,key pair.
+ *
+ * @param {import('./types.js').Baggage} baggage
+ * @param {string} key
+ * @param {(key?: string) => any} makeValue
+ * @returns {any}
+ */
+export const provide = (baggage, key, makeValue) => {
+  return provideLazy(baggage, key, makeValue);
+};
```
turadg: why define new function?
```suggestion
export const provide = provideLazy;
```

packages/vat-data/src/vat-data-bindings.js
```diff
@@ -89,6 +89,41 @@ export const partialAssign = (target, source) => {
...
+ * TODO https://github.com/Agoric/agoric-sdk/pull/5875 :
+ * Implement development-time instrumentation to detect when
+ * `provide` violates the above prescription, and is called more
+ * than one in the same vat incarnation with the same
+ * baggage,key pair.
+ *
+ * @param {import('./types.js').Baggage} baggage
+ * @param {string} key
+ * @param {(key?: string) => any} makeValue
+ * @returns {any}
```
turadg: this is a regression of type information. please type the return:
```suggestion
 * @param {(key?: string) => V} makeValue
 * @returns {V}
 * @template V
```

turadg (changes requested):

packages/inter-protocol/scripts/deploy-contracts.js
erights: ;)

packages/store/src/stores/store-utils.js
erights: Done

packages/vat-data/src/vat-data-bindings.js
erights: Doh! Good point. Done.

packages/vat-data/src/vat-data-bindings.js
erights: Borrowed from the original, which defined `K` too. Done. PTAL

turadg (approved):
👌

PR #1275, opened 2020-07-08
chore: add name of keys to Store

katelynsills:
Add the name of keys to `makeStore` calls to improve the default error messages. 

Starts to address #1274 

Now the error reads slightly better:

Screen Shot 2020-07-08 at 4 48 54 PM

PR #6082, opened 2022-08-28
fix: patch to emulate endo #1275

erights:
Our attempt to safely tolerate Node async_hooks was safe, but missed a case we could handle safely and which is provoked at least by the vscode debugger.

This patch emulates https://github.com/endojs/endo/pull/1275 so that, for example, our experience debugging our code under vscode stops encountering this problem.

From recent experience, we know that we can't really depend on endo patches to smooth over agoric-sdk dependence on unreleased endo features. But AFAIK, we (including our users) only encounter this condition during development, and our previous problem with patches happened only under CI, which has never been known to provoke the case handled here.


mhofman (approved):
Approving since the upstream PR was approved and merged. However a reminder that `patch-package` will not work when agoric-sdk is installed as a normal dependency of a dapp. Of course this is probably fine since besides CI, probably no-one uses that flow currently.

Issue #4489, opened 2022-02-07
walletFactory uses virtual objects


michaelfig:
## What is the Problem Being Solved?

Wallet manager and individual wallet backends need to be scalable.

## Description of the Design

Do the wholesale migration of the wallet backends to collections of virtual objects.

Tartuffo:
@michaelfig Do you want this to be under the #3995 (as it is currently) or #4511 (which is the epic for all of the other virtual objects / collection work)?

Tartuffo:
This work might get subsumed fully or partially by #4484, or at least much lest work.

Tartuffo:
Wallet Manager is deprecated, so do we still need to do this?

warner:
We need to measure the slope: RAM used per provisioned account, and also GC time (because that scales with the size of the object graph, hence RAM). Then we can compare that against the 2GiB per-vat limit (#5953) to estimate how many users can be added before the worker would be killed by the memory limit, and then since it is (supposed to be) a `criticalVat`, the kernel panics and the chain halts.

Issue #4809, opened 2022-03-10
Misleading status message when AMM offer is rejected


turadg:
## Describe the bug
If the AMM offer is processed it returns "Swap successfully completed." even if the offer was rejected. For example, due to making an offer that's not on the AMM curve.

## Expected behavior
Some response that informs the user of the new state. 

## Additional context

The REPL is part of the UI, though not GUI.

See also https://github.com/Agoric/agoric-sdk/issues/3056

Tartuffo:
Need to properly surface the outcome of the offer.
Should it throw or politely return the "give"?
Can handle entirely in the wallet, but will be seen in the REPL.

dckc:
another awkward consequence of not throwing is a notifier update with no change in state: https://github.com/Agoric/agoric-sdk/pull/5420/files#r879991457

turadg:
Either this or https://github.com/Agoric/agoric-sdk/issues/3056 needs to be done for MN-1.

Need to discuss with @erights which. When determined then update both tickets.

erights:
Yes, let's discuss. @dtribble and I discussed a method that could easily be added to UserSeat that would help. But I am not sure I understand the problem and why the wallet could not easily do the equivalent of that method for itself.

I am not that oriented in the wallet, so I may be missing something obvious. It seems the problem is that the wallet is paying looking at the offerResult when it should be looking at the payout. The question it seems to want to ask is "Is this offer's proposal's `want` satisfied?" The UserSeat method @dtribble and I discussed, that @dtribble thinks would help, is an `isSatisfied` method that returns a promise that eventually resolved only to `true` or `false`. Even with that convenience, a returned promise will not survive an upgrade, so the wallet would need to cope with the normal upgrade-invalidated promise pattern, where it asks for a new promise in that case.

dckc:
The misleading "Swap successfully completed." is shown in the dapp (amm.agoric.net), not the wallet.

Are you suggesting that every dApp developer should inspect the payouts rather than the offer result to get... um... the result of the offer?

Wouldn't it be much more straightforward for the AMM contract to throw in this case? I use the AMM API quite a bit, and remembering that "Swap successfully completed." actually means "swap may or may not have successfully completed; you better look more closely" is a royal pain.

At least we should change the message to "Swap order processed. Check payouts for results."

Tartuffo:
[...]

PR #6084, opened 2022-08-29
new smart wallet

### LINKED ISSUE #4488: Refactor wallet backend for POLA

Each aspect of the wallet backend should be kept as separate as possible from the other implementations. Right now is dangerous because it has a lot of authority and it's all comingled.

## Description of the Design

Break out the wallet backend into separate modules for each area managed by the wallet.

Time box project to pick the fruit we can in the allotted time. Target 2 days of 2 engineers, plus code review.

Break the kitchen sink API into separate objects. E.g. "pet name manager" owns the pet names. 

State is split across collections and there's a bunch at the top. So any bug in any of that code could iterate over those purses and spend your money. Instead need to encapsulate the purse state.

Dapp UIs shouldn't break. The only thing that relies on the shallow list of functions is the Wallet UI (not dapps, which rely on the wallet bridge).

turadg:
closes: #4488

## Description

A whole new Smart Wallet contract. This builds on the design and code of the agsolo wallet but the fresh start reduces the code surface to audit for PSM launch. It's also designed from the ground up for POLA and to take advantage of marshalling contexts instead of board ids.

Some features that aren't needed for PSM launch have been omitted and filed for future work:
- https://github.com/Agoric/agoric-sdk/issues/6126
- https://github.com/Agoric/agoric-sdk/issues/6132

And these are still needed for PSM launch:
- https://github.com/Agoric/agoric-sdk/issues/6127
- https://github.com/Agoric/agoric-sdk/issues/4489

### Security Considerations

This let's anyone send a message to the contract that can execute arbitrary offers on a wallet instance. It relies on the bridge handler to ensure that the only messages that reach a wallet come from its owner.

New docs including an attackers guide.

### Testing Considerations

New tests

### DIFF
--- a/packages/smart-wallet/package.json
+++ b/packages/smart-wallet/package.json
@@ -14,8 +14,12 @@
   },
   "devDependencies": {
     "ava": "^4.3.1",
+    "@agoric/cosmic-swingset": "^0.37.0",
+    "@agoric/cosmic-proto": "^0.1.0",
     "@agoric/inter-protocol": "^0.11.0",
     "@agoric/swingset-vat": "^0.28.0",
+    "@agoric/wallet-backend": "^0.12.1",
+    "@agoric/vats": "^0.10.0",
     "@agoric/zoe": "^0.24.0",
     "@endo/captp": "^2.0.13"
   },
@@ -27,7 +31,6 @@
     "@agoric/notifier": "^0.4.0",
     "@agoric/store": "^0.7.2",
     "@agoric/vat-data": "^0.3.1",
-    "@agoric/vats": "^0.10.0",
     "@endo/far": "^0.2.9"
   },
   "keywords": [],
--- a/packages/smart-wallet/src/smartWallet.js
+++ b/packages/smart-wallet/src/smartWallet.js
@@ -1,0 +1,333 @@
+// @ts-check
+import { isNat } from '@agoric/nat';
+import {
+  makeStoredPublishKit,
+  observeIteration,
+  observeNotifier,
+} from '@agoric/notifier';
+import { makeScalarMapStore } from '@agoric/store';
+import { makeScalarBigMapStore } from '@agoric/vat-data';
+import { E, Far } from '@endo/far';
+import { makeInvitationsHelper } from './invitations.js';
+import { makeOffersFacet } from './offers.js';
+
+const { details: X, quote: q } = assert;
+
+/**
+ * @file Smart wallet module
+ *
+ * @see {@link ../README.md}}
+ */
+
+// One method yet but structured to support more. For example,
+// maybe suggestIssuer for https://github.com/Agoric/agoric-sdk/issues/6132
+// setting petnames and adding brands for https://github.com/Agoric/agoric-sdk/issues/6126
+/**
+ * @typedef {{
+ *   method: 'executeOffer'
+ *   offer: import('./offers.js').OfferSpec,
+ * }} BridgeAction
+ */
+
+/**
+ * @typedef {{ updated: 'offerStatus', status: import('./offers.js').OfferStatus } |
+ * { updated: 'balance'; currentAmount: Amount } |
+ * { updated: 'brand', descriptor: BrandDescriptor }
+ * } UpdateRecord Record of an update to the state of this wallet.
+ *
+ * Client is responsible for coalescing updates into a current state. See `coalesceUpdates` utility.
+ *
+ * The reason for this burden on the client is that transferring the full state is untenable
+ * (because it would grow monotonically).
+ *
+ * `balance` update supports forward-compatibility for more than one purse per
+ * brand. An additional key will be needed to disambiguate. For now the brand in
+ * the amount suffices.
+ */
+
+/**
+ * @typedef {{
+ * brand: Brand,
+ * displayInfo: DisplayInfo,
+ * issuer: ERef<Issuer>,
+ * petname: import('./types').Petname
+ * }} BrandDescriptor
+ * For use by clients to describe brands to users. Includes `displayInfo` to save a remote call.
+ */
+
+// imports
+/** @typedef {import('./types').RemotePurse} RemotePurse */
+
+/**
+ *
+ * @param {{
+ * address: string,
+ * bank: ERef<import('@agoric/vats/src/vat-bank').Bank>,
+ * }} unique
+ * @param {{
+ * agoricNames: ERef<NameHub>,
+ * board: ERef<Board>,
+ * invitationIssuer: ERef<Issuer<'set'>>,
+ * invitationBrand: Brand<'set'>,
+ * storageNode: ERef<StorageNode>,
+ * zoe: ERef<ZoeService>,
+ * }} shared
+ */
+export const makeSmartWallet = async (
+  { address, bank },
+  { board, invitationBrand, invitationIssuer, storageNode, zoe },
+) => {
+  assert.typeof(address, 'string', 'invalid address');
+  assert(bank, 'missing bank');
+  assert(invitationIssuer, 'missing invitationIssuer');
+  assert(invitationBrand, 'missing invitationBrand');
+  assert(storageNode, 'missing storageNode');
+  // cache
+  const [invitationPurse, marshaller] = await Promise.all([
+    E(invitationIssuer).makeEmptyPurse(),
+    E(board).getReadonlyMarshaller(),
+  ]);
+
+  // #region STATE
+
+  // - brandPurses is precious and closely held. defined as late as possible to reduce its scope.
+  // - offerToInvitationMakers is precious and closely held.
+  // - lastOfferId is precious but not closely held
+  // - brandDescriptors will be precious. Currently it includes invitation brand and  what we've received from the bank manager.
+  // - purseBalances is a cache of what we've received from purses. Held so we can publish all balances on change.
+
+  /**
+   * To ensure every offer ID is unique we require that each is a number greater
+   * than has ever been used. This high water mark is sufficient to track that.
+   *
+   * @type {number}
+   */
+  let lastOfferId = 0;
+
+  /**
+   * Invitation makers yielded by offer results
+   *
+   * @type {MapStore<number, import('./types').RemoteInvitationMakers>}
+   */
+  const offerToInvitationMakers = makeScalarBigMapStore('invitation makers', {
+    durable: true,
+  });
+
+  /** @type {MapStore<Brand, BrandDescriptor>} */
+  const brandDescriptors = makeScalarMapStore();
+
+  /**
+   * What purses have reported on construction and by getCurrentAmountNotifier updates.
+   *
+   * @type {MapStore<RemotePurse, Amount>}
+   */
+  const purseBalances = makeScalarMapStore();
+
+  // #endregion
+
+  // #region publishing
+  // NB: state size must not grow monotonically
+  // This is the node that UIs subscribe to for everything they need.
+  // e.g. agoric follow :published.wallet.agoric1nqxg4pye30n3trct0hf7dclcwfxz8au84hr3ht
+  const myWalletStorageNode = E(storageNode).makeChildNode(address);
+
+  /** @type {StoredPublishKit<UpdateRecord>} */
+  const updatePublishKit = makeStoredPublishKit(
+    myWalletStorageNode,
+    marshaller,
+  );
+
+  /**
+   * @param {RemotePurse} purse
+   * @param {Amount} balance
+   * @param {'init'} [init]
+   */
+  const updateBalance = (purse, balance, init) => {
+    if (init) {
+      purseBalances.init(purse, balance);
+    } else {
+      purseBalances.set(purse, balance);
+    }
+    updatePublishKit.publisher.publish({
+      updated: 'balance',
+      currentAmount: balance,
+    });
+  };
+
+  // #endregion
+
+  // #region issuer management
+  /**
+   * Private purses. This assumes one purse per brand, which will be valid in MN-1 but not always.
+   *
+   * @type {MapStore<Brand, RemotePurse>}
+   */
+  const brandPurses = makeScalarBigMapStore('brand purses', { durable: true });
+
+  /** @type { (desc: Omit<BrandDescriptor, 'displayInfo'>, purse: RemotePurse) => Promise<void>} */
+  const addBrand = async (desc, purseRef) => {
+    // assert haven't received this issuer before.
+    const descriptorsHas = brandDescriptors.has(desc.brand);
+    const pursesHas = brandPurses.has(desc.brand);
+    assert(
+      !(descriptorsHas && pursesHas),
+      'repeated brand from bank asset subscription',
+    );
+    assert(
+      !(descriptorsHas || pursesHas),
+      'corrupted state; one store has brand already',
+    );
+
+    const [purse, displayInfo] = await Promise.all([
+      purseRef,
+      E(desc.brand).getDisplayInfo(),
+    ]);
+
+    // save all five of these in a collection (indexed by brand?) so that when
+    // it's time to take an offer description you know where to get the
+    // relevant purse. when it's time to make an offer, you know how to make
+    // payments. REMEMBER when doing that, need to handle every exception to
+    // put the money back in the purse if anything fails.
+    const descriptor = { ...desc, displayInfo };
+    brandDescriptors.init(desc.brand, descriptor);
+    brandPurses.init(desc.brand, purse);
+
+    // publish purse's balance and changes
+    E.when(
+      E(purse).getCurrentAmount(),
+      balance => updateBalance(purse, balance, 'init'),
+      err =>
+        console.error(address, 'initial purse balance publish failed', err),
+    );
+    observeNotifier(E(purse).getCurrentAmountNotifier(), {
+      updateState(balance) {
+        updateBalance(purse, balance);
+      },
+      fail(reason) {
+        console.error(address, `failed updateState observer`, reason);
+      },
+    });
+
+    updatePublishKit.publisher.publish({ updated: 'brand', descriptor });
+  };
+
+  // Ensure a purse for each issuer
+  addBrand(
+    {
+      brand: invitationBrand,
+      issuer: invitationIssuer,
+      petname: 'invitations',
+    },
+    // @ts-expect-error cast to RemotePurse
+    /** @type {RemotePurse} */ (invitationPurse),
+  );
+  // watch the bank for new issuers to make purses out of
+  observeIteration(E(bank).getAssetSubscription(), {
+    async updateState(desc) {
+      /** @type {RemotePurse} */
+      // @ts-expect-error cast to RemotePurse
+      const purse = E(bank).getPurse(desc.brand);
+      await addBrand(
+        {
+          brand: desc.brand,
+          issuer: desc.issuer,
+          petname: desc.proposedName,
+        },
+        purse,
+      );
+    },
+  });
+  // #endregion
+
+  /**
+   * Similar to {DepositFacet} but async because it has to look up the purse.
+   */
+  const depositFacet = Far('smart wallet deposit facet', {
+    // TODO(PS0) decide whether to match canonical `DepositFacet'. it would have to take a local Payment.
+    /**
+     * Put the assets from the payment into the appropriate purse
+     *
+     * @param {ERef<Payment>} paymentE
+     * @param {Brand} [paymentBrand] when provided saves remote lookup. Must match the payment's brand.
+     * @returns {Promise<Amount>}
+     * @throws if the purse doesn't exist
+     * NB: the previous smart wallet contract would try again each time there's a new issuer.
+     * This version does not: 1) for expedience, 2: to avoid resource exhaustion vulnerability.
+     */
+    receive: async (paymentE, paymentBrand) => {
+      const brand = await (paymentBrand || E(paymentE).getAllegedBrand());
+      const purse = brandPurses.get(brand);
+
+      return E.when(paymentE, payment => E(purse).deposit(payment));
+    },
+  });
+
+  const offersFacet = makeOffersFacet({
+    zoe,
+    powers: {
+      invitationFromSpec: makeInvitationsHelper(
+        zoe,
+        invitationBrand,
+        invitationPurse,
+        offerToInvitationMakers.get,
+      ),
+      purseForBrand: brandPurses.get,
+      lastOfferId: {
+        get: () => lastOfferId,
+        set(id) {
+          assert(isNat(id), 'offer id must be a positive number');
+          assert(
+            id > lastOfferId,
+            'offer id must be greater than all previous',
+          );
+          lastOfferId = id;
+        },
+      },
+    },
+    onStatusChange: offerStatus =>
+      updatePublishKit.publisher.publish({
+        updated: 'offerStatus',
+        status: offerStatus,
+      }),
+    onNewContinuingOffer: (offerId, invitationMakers) =>
+      offerToInvitationMakers.init(offerId, invitationMakers),
+  });
+
+  /**
+   *
+   * @param {import('./types').WalletCapData<BridgeAction>} actionCapData
+   * @param {boolean} [canSpend=false]
+   */
+  const handleBridgeAction = (actionCapData, canSpend = false) => {
+    assert(actionCapData.body && actionCapData.slots, 'invalid capdata');
+    assert(!canSpend, 'spending not yet supported');
+    return E.when(
+      E(marshaller).unserialize(actionCapData),
+      /** @param {BridgeAction} action */
+      action => {
+        switch (action.method) {
+          case 'executeOffer':
+            return E(offersFacet).executeOffer(action.offer);
+          default:
+            assert.fail(X`invalid handle bridge action ${q(action)}`);
+        }
+      },
+    );
+  };
+
+  /**
+   * Holders of this object:
+   * - vat (transitively from holding the wallet factory)
+   * - wallet-ui (which has key material; dapps use wallet-ui to propose actions)
+   */
+  return Far('SmartWallet', {
+    handleBridgeAction,
+    getDepositFacet: () => depositFacet,
+    getOffersFacet: () => offersFacet,
+
+    getUpdatesSubscriber: () => updatePublishKit.subscriber,
+  });
+};
+harden(makeSmartWallet);
+
+/** @typedef {Awaited<ReturnType<typeof makeSmartWallet>>} SmartWallet */
--- a/packages/smart-wallet/src/utils.js
+++ b/packages/smart-wallet/src/utils.js
@@ -1,0 +1,46 @@
+// @ts-check
+
+import { observeIteration, subscribeEach } from '@agoric/notifier';
+
+/**
+ * Coalesce updates from a wallet UpdateRecord publication feed. Note that local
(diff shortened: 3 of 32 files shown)


dckc:
Let's please be sure `numWantsSatisfied` is in offer status records:
https://github.com/Agoric/agoric-sdk/issues/4809#issuecomment-1233469962

packages/wallet/contract/package.json
```diff
@@ -18,17 +18,27 @@
...
+    "@agoric/wallet-backend": "0.12.1",
     "@endo/captp": "^2.0.13"
   },
   "dependencies": {
-    "@agoric/wallet-backend": "0.12.1",
+    "@agoric/assert": "^0.4.0",
+    "@agoric/cache": "^0.1.0",
     "@agoric/deploy-script-support": "^0.9.0",
+    "@agoric/ertp": "^0.14.2",
+    "@agoric/inter-protocol": "^0.11.0",
```
dckc: wallet depends on inter-protocol?

Is it perhaps just for types? It could be a devdependency then, yes?

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+  const pursesSubscriber = makeStoredSubscriber(
+    pursePublishKit.subscriber,
+    E(myWalletStorageNode).makeChildNode('purses'),
+    marshaller,
+  );
+
+  const publishPurses = () => {
+    const record = Array.from(brandPurses.entries()).map(([brand, purse]) => {
+      const desc = brandDescriptors.get(brand);
+      return { ...desc, purseBalance: purse.getCurrentAmount() };
```
dckc: purses are remote (not created in this vat)

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+  // #region State and accessors
+  /**
+   * What we've been told by the Bank or suggestIssuer
+   *
+   * @type {WeakMapStore<Brand, BrandDescriptor>} */
+  const brandDescriptors = makeScalarWeakMapStore();
+  /**
+   * Private purses
+   *
+   * @type {MapStore<Brand, Purse>}
```
dckc: purses are remote

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
  * @param {{
  * address: string,
  * bank: ERef<import('@agoric/vats/src/vat-bank').Bank>,
- * myAddressNameAdmin: ERef<MyAddressNameAdmin>,
  * }} unique
  * @param {{
- * agoricNames: ERef<NameHub>,
  * board: ERef<Board>,
- * namesByAddress: ERef<NameHub>,
+ * invitationIssuer: Issuer<'set'>,
```
dckc: this issuer is remote. `ERef`.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+    // put the money back in the purse if anything fails.
+    brandDescriptors.init(desc.brand, desc);
+    brandPurses.init(desc.brand, purse);
+
+    publishPurses();
+  };
+
+  // Ensure a purse for each issuer
+  addIssuer({
+    brand: invitationIssuer.getBrand(),
```
dckc: remote call. use `E()`

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+  const depositFacet = Far('smart wallet deposit facet', {
+    /**
+     * Put the assets from the payment into the appropriate purse
+     *
+     * @param {Payment} payment
+     * @throws if the purse doesn't exist
+     * ??? PRODUCT: should it instead wait for an issuer/purse to put it in?
+     */
+    receive: payment => {
+      const brand = payment.getAllegedBrand();
```
dckc: remote. `E()`

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+      const purse = brandPurses.get(brand);
+      const depositP = purse.deposit(payment);
```
dckc: remote.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+      E.when(depositP, publishPurses);
```
dckc: use `void` to indicate that we're dropping this promise, right?

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+      return depositP;
+    },
   });
 
-  const marshaller = wallet.getMarshaller();
+  const findInvitationInPurse = () => {
+    const invitationBrand = invitationIssuer.getBrand();
```
dckc: remote. `E()`

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
+    /** @type {Purse<'set'>} */
+    const purse = brandPurses.get(invitationBrand);
+    // FIXME find a matching one
+    const toWithdraw = purse.getCurrentAmount();
```
dckc: remote. `E()`. We don't create any issuers, brands, nor purses in this vat, so they're all remote.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -4,87 +4,264 @@ import {
...
-  const storedSubscriber = makeStoredSubscriber(
-    subscriber,
-    myWalletStorageNode,
-    marshaller,
-  );
+    return purse.withdraw(toWithdraw);
+  };
+  const invitationFromSpec = makeInvitationsHelper(zoe, findInvitationInPurse);
 
+  const issuerSuggester = Far('issuerSuggester', {
```
dckc: move this out of scope of the purses?

packages/wallet/contract/src/walletFactory.js
```diff
@@ -101,7 +100,10 @@ export const start = async (zcf, privateArgs) => {
 
     /** @type {() => Promise<import('./smartWallet').SmartWallet>} */
     const maker = () =>
-      makeSmartWallet({ address, bank, myAddressNameAdmin }, shared);
+      makeSmartWallet({ address, bank }, shared).then(wallet => {
+        E(myAddressNameAdmin).update('depositeFacet', wallet.getDepositFacet());
```
dckc: `void`?

packages/wallet/contract/test/test-psm.js
```diff
@@ -0,0 +1,226 @@
...
+      proposal: {
+        // empty amounts
+        give: { In: AmountMath.makeEmpty(minted) },
+        want: { Out: anchor.makeEmpty() },
+      },
+    }),
+  );
+
+  // @ts-expect-error faulty generic typedef
+  wallet.executeOffer(offerSpecCapData);
```
dckc: `await`? ah. no. We shouldn't rely on getting a promise for the result. Seems worth documenting with `void` if not a comment.

dckc (commented):
solid progress!

dckc:
@turadg here's hoping the results of this work don't have memory performance issues such as

 - #6104

packages/wallet/contract/package.json
turadg: good catch. not sure why I added this since it was already in the devDeps list

packages/wallet/contract/src/smartWallet.js
turadg: payment is remote?

packages/wallet/contract/src/smartWallet.js
turadg: good catch. will handle in some fashion

packages/wallet/contract/src/smartWallet.js
turadg: btw, to save you typing in reviews: once you correct the annotation of a type, TS notifies of all the call sites that don't conform

packages/wallet/contract/src/smartWallet.js
turadg: It's this block,
```
    // ??? do we need this for PSM launch? if not better to remove the footgun to have assets that could get blown away in upgrade
    // can be held by the dapp. needed by the way we're doing AMM.
    // should have no possibility of unmarshalling a purse
    getIssuerSuggester: () => issuerSuggester,
```
so I'll just delete it for PS0.

packages/wallet/contract/src/walletFactory.js
turadg: not void. it's returning a promise for a smart wallet. up to the consume to handle rejection.

packages/wallet/contract/test/test-psm.js
turadg: you mean document at the callsite? the docs on `executeOffer` already say it's a synchronous function:

>`      * @returns {void} when the offer has been sent to Zoe; payouts go into this wallet's purses`

packages/wallet/contract/src/smartWallet.js
dckc: Yes. Payments are created by issuers. There are no issuers created by this vat, and hence no issuers local to this vat.

Everything that a contract doesn't create is remote from that contract.

packages/wallet/contract/src/smartWallet.js
dckc: @samsiegart be advised... integration with the AMM (esp suggestIssuer) is likely to be delayed.

packages/wallet/contract/src/walletFactory.js
dckc: `E(nameAdmin).update()` is returning the smart wallet? I don't see how. Maybe I'm distracted...

turadg: the line after it is

packages/wallet/contract/src/smartWallet.js
samsiegart: Makes sense, thanks for the heads up

packages/inter-protocol/src/psm/psm.js
```diff
@@ -149,7 +149,7 @@ export const start = async (zcf, privateArgs, baggage) => {
   /**
    * @param {ZCFSeat} seat
    * @param {Amount<'nat'>} given
-   * @param {Amount<'nat'>} [wanted]
+   * @param {Amount<'nat'>} [wanted] defaults to maximum anchor (given exchange rate minus fees)
```
samsiegart: Just curious, when would you want anything less than the maximum?

packages/wallet/contract/src/invitations.js
```diff
@@ -0,0 +1,92 @@
...
+  findInPurse,
+  getInvitationContinuation,
+) => {
+  // XXX make a helper with an invitation purse query object
+  const invitationGetters = /** @type {const} */ ({
+    /** @type {(spec: ContractInvitationSpec) => Promise<Invitation>} */
+    contract(spec) {
+      const { instance, publicInvitationMaker } = spec;
+      const pf = E(zoe).getPublicFacet(instance);
+      return E(pf)[publicInvitationMaker]();
```
samsiegart: Should this support `...args`? I think the last version did.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,425 @@
...
+import { E, Far, passStyleOf } from '@endo/far';
+import { makeInvitationsHelper } from './invitations.js';
+import { paymentsFromPurses, paymentsIntoPurses } from './utils.js';
 
-const { assign, entries, keys, fromEntries } = Object;
+/**
+ * @file Smart wallet module
+ *
+ * Design notes:
+ * PRIME DIRECTIVE: DON'T LOSE MONEY.
```
samsiegart: Nit: "assets" or "tokens" instead of money

packages/wallet/contract/src/utils.js
```diff
@@ -0,0 +1,63 @@
...
+import { deeplyFulfilledObject, objectMap } from '@agoric/internal';
+import { E } from '@endo/far';
+
+/**
+ *
+ * @param {AmountKeywordRecord} amountKeywordRecord
+ * @param {(brand: Brand) => ERef<Purse>} purseForBrand
+ * @returns {PaymentPKeywordRecord}
+ */
+export const paymentsFromPurses = (amountKeywordRecord, purseForBrand) => {
```
samsiegart: Nit: Rename to `withdrawPaymentsFromPurses` so it's clear that this is performing an action

packages/wallet/contract/src/utils.js
```diff
@@ -0,0 +1,63 @@
+// @ts-check
+
+import { assert, details as X, quote as q } from '@agoric/assert';
+import { deeplyFulfilledObject, objectMap } from '@agoric/internal';
+import { E } from '@endo/far';
+
+/**
+ *
+ * @param {AmountKeywordRecord} amountKeywordRecord
+ * @param {(brand: Brand) => ERef<Purse>} purseForBrand
```
samsiegart: Nit: Rename to `getPurseForBrand` (here and below). I thought that this was just a single purse reference based on the name.

packages/wallet/contract/src/utils.js
```diff
@@ -0,0 +1,63 @@
...
+};
+harden(paymentsFromPurses);
+
+/**
+ *
+ * @param {PaymentPKeywordRecord} paymentPKeywordRecord
+ * @param {(brand: Brand) => ERef<Purse>} purseForBrand
+ * @returns {Promise<Record<string, Amount>>}
+ */
+export const paymentsIntoPurses = async (
```
samsiegart: Nit: `depositPaymentsIntoPurses`

packages/inter-protocol/src/psm/psm.js
turadg: not sure myself. A case I can imagine is that if you want a certain amount and you don't know the fees you give a surplus and rely on the contract to only take what's necessary for the want.

packages/wallet/contract/src/invitations.js
turadg: yes! missed that

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,425 @@
...
+  // To ensure every offer ID is unique we require that each is a number greater
+  // than has ever been used. This high water mark is sufficient to track that.
+  let lastOfferId = 0;
+  /**
+   * What we've been told by the Bank or suggestIssuer
+   *
+   * @type {WeakMapStore<Brand, BrandDescriptor>} */
+  const brandDescriptors = makeScalarWeakMapStore();
+  /** @type {(brand: Brand) => ERef<Purse>} */
+  const getPurse = brand => brandPurses.get(brand);
```
samsiegart: This won't always be a 1:1 mapping right?

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,425 @@
...
-    getSubscriber: () => storedSubscriber,
-    performAction: obj => E(admin).performAction(obj),
+    /**
+     * Take an offer description provided in capData, augment it with payments and call zoe.offer()
+     *
+     * @param {WalletCapData<OfferSpec>} capData
+     * @returns {void} when the offer has been sent to Zoe; payouts go into this wallet's purses
+     * @throws if any parts of the offer can be determined synchronously to be invalid
+     */
+    executeOffer: capData => {
```
samsiegart: I guess this purposely doesn't handle attestations for now right?

packages/wallet/contract/src/smartWallet.js
turadg: Why not? The design is that it makes a single purse for each brand it learns of.

packages/wallet/contract/src/smartWallet.js
turadg: yeah, not in scope for PSM launch. @dckc will supporting that affect the API?

packages/wallet/contract/src/smartWallet.js
samsiegart: For now yes, as long as we can extend it in the future to support multiple purses of the same brand

packages/wallet/contract/src/smartWallet.js
dckc: @samsiegart remind me how we represented "this offer should include an attestation" in our `acceptOffer` argument?

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,273 @@
...
+    pursePublishKit.subscriber,
+    E(myWalletStorageNode).makeChildNode('purses'),
+    marshaller,
+  );
+
+  const publishPurses = async () => {
+    const entries = Array.from(brandPurses.entries());
+    // indexed the same as brand array to merge
+    const balances = await Promise.all(
+      entries.map(([_, purse]) => E(purse).getCurrentAmount()),
```
dckc: Each purse has a balance notifier; I would expect us to use that rather than polling with `getCurrentAmount()`.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,273 @@
...
+  const offersSubscriber = makeStoredSubscriber(
+    offerPublishKit.subscriber,
+    E(myWalletStorageNode).makeChildNode('offers'),
     marshaller,
   );
+  /**
+   * @param {import('./offers.js').OfferStatus} status
+   */
+  const publishOfferStatus = status => {
+    const record = { [status.id]: status };
```
dckc: why wrap `status` in another layer?

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,273 @@
...
+
+    // save all five of these in a collection (indexed by brand?) so that when
+    // it's time to take an offer description you know where to get the
+    // relevant purse. when it's time to make an offer, you know how to make
+    // payments. REMEMBER when doing that, need to handle every exception to
+    // put the money back in the purse if anything fails.
+    brandDescriptors.init(desc.brand, desc);
+    brandPurses.init(desc.brand, purse);
+
+    publishPurses();
```
dckc: I expect we would publish an `{ issuer, brand, displayInfo, petname }` record here.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,273 @@
...
+     * @returns {Promise<Amount>}
+     * @throws if the purse doesn't exist
+     * ??? PRODUCT: should it instead wait for an issuer/purse to put it in?
+     */
+    receive: (paymentE, paymentBrand) => {
+      const purse = brandPurses.get(paymentBrand);
+
+      const depositP = E.when(paymentE, payment => E(purse).deposit(payment));
+
+      E.when(depositP, publishPurses, () =>
```
dckc: the purse's balance notifier should suffice; I don't think we need to `publishPurses` here.

packages/wallet/contract/src/smartWallet.js
```diff
@@ -1,90 +1,273 @@
...
+  const marshaller = await E(board).getReadonlyMarshaller();
 
-  const cacheStorageNode = E(storageNode).makeChildNode('cache');
+  // #region precious state
+  /**
+   * Private purses
+   *
+   * @type {MapStore<Brand, ERef<Purse<'set'>>>}
+   */
+  const brandPurses = makeScalarBigMapStore('brand purses', { durable: true });
```
dckc: It still seems like we could reduce the scope of this by, for example, moving it along with `publishPurses` later in the file.

dckc (commented):
just a few comments here and there. I haven't managed to read through the whole thing.

packages/wallet/contract/src/smartWallet.js
turadg: documenting in PR: decision made with product to not support multiple purses for same brand for MN-1.

of course an upgraded contract will be able to. at that point we'll need some way to specify in offer execution which purses to take funds from. For simplicity I think we won't require that specificity unless there are multiple purses. When there are, lack of specifier could throw or we could have a "default" purse for each brand.

Should this go into the source code somewhere? jsdoc, README.md or a new DESIGN.md ?

packages/wallet/contract/src/smartWallet.js
turadg: you mean under `id`? because the commitment of our state streams is that they're the state of "something". Pushing `status` without putting it under a key means that something can flip flop from "offer 33" to "offer 44". 

What I wrote on the `OfferRecords` type is that it's a map of all offers ever, but not stored as such. Each publication is a selection from that virtual map.

I think this is right but if you disagree let's discuss at the `OffersRecord` type.

packages/wallet/contract/src/smartWallet.js
turadg: That's basically `BrandDescriptor`, plus `displayInfo` to save a request.
https://github.com/Agoric/agoric-sdk/blob/c74df8148bdd198bef4e7fc2fec9209ddecdc89f/packages/wallet/contract/src/smartWallet.js#L28

I see now `getCurrentAmountNotifier` on purse. So we don't need `purseBalance`, but we do need to give clients a way to get that notifier. I'll include that in `PursesRecord` as `currentAmountNotifier`.

packages/wallet/contract/src/smartWallet.js
turadg: Great. I didn't know about that notifier but it helps a lot. I'll refactor using it after I get the bridge working.

packages/wallet/contract/src/smartWallet.js
turadg: Good idea. For auditability I tried to gather the state and for upgradability I tried to gather precious state. But it would be better to have a comment explicitly about state requirements and let them be defined in the most advantageous place for safety.

packages/wallet/contract/src/smartWallet.js
dckc: > Should this go into the source code somewhere?

yes, please. I prefer jsdoc, but other places are ok too.

packages/wallet/contract/src/smartWallet.js
turadg: removed

packages/wallet/contract/src/utils.js
turadg: good call!

packages/wallet/contract/src/utils.js
turadg: The function does just look up a single purse for a single brand.

packages/wallet/contract/src/smartWallet.js
turadg: added in jsdoc. I think Markdown would be easier to read but am fine with this

packages/smart-wallet/src/walletFactory.js
```diff
@@ -0,0 +1,180 @@
...
+   */
+  const provideSmartWallet = async (address, bank, myAddressNameAdmin) => {
+    assert.typeof(address, 'string', 'invalid address');
+    assert(bank, 'missing bank');
+    assert(myAddressNameAdmin, 'missing myAddressNameAdmin');
+
+    /** @type {() => Promise<import('./smartWallet').SmartWallet>} */
+    const maker = () =>
+      makeSmartWallet({ address, bank }, shared).then(wallet => {
+        E(myAddressNameAdmin).update('depositeFacet', wallet.getDepositFacet());
```
dckc, line 107:
```suggestion
        E(myAddressNameAdmin).update('depositFacet', wallet.getDepositFacet());
```

packages/smart-wallet/src/smartWallet.js
```diff
@@ -0,0 +1,307 @@
...
+    // relevant purse. when it's time to make an offer, you know how to make
+    // payments. REMEMBER when doing that, need to handle every exception to
+    // put the money back in the purse if anything fails.
+    const descriptor = { ...desc, displayInfo };
+    brandDescriptors.init(desc.brand, descriptor);
+    brandPurses.init(desc.brand, purse);
+
+    // publish purse's balance and changes
+    E.when(
+      purse.getCurrentAmount(),
```
dckc:
```suggestion
      E(purse).getCurrentAmount(),
```

packages/smart-wallet/src/smartWallet.js
```diff
@@ -0,0 +1,307 @@
...
+     * Put the assets from the payment into the appropriate purse
+     *
+     * @param {ERef<Payment>} paymentE
+     * @param {Brand} paymentBrand must match the payment's brand. passed in to save lookup.
+     * @returns {Promise<Amount>}
+     * @throws if the purse doesn't exist
+     * ??? PRODUCT: should it instead wait for an issuer/purse to put it in?
+     */
+    receive: (paymentE, paymentBrand) => {
+      const purse = brandPurses.get(paymentBrand);
```
dckc: Integration testing suggests that `paymentBrand` is optional.

```suggestion
      const brand = await (paymentBrand || E(paymentE).getAllegedBrand());
      const purse = brandPurses.get(brand);
```

packages/smart-wallet/src/smartWallet.js
```diff
@@ -0,0 +1,307 @@
...
+  // Ensure a purse for each issuer
+  addBrand({
+    brand: invitationBrand,
+    issuer: invitationIssuer,
+    petname: 'invitations',
+  });
+  // watch the bank for new issuers to make purses out of
+  observeIteration(E(bank).getAssetSubscription(), {
+    async updateState(desc) {
+      await addBrand({
```
dckc: When we learn about a new issuer by way of `bank`, we need to get the purse from the bank rather than make a new one:

```js
      const purse = await E(bank).getPurse(desc.brand);
```

See also https://github.com/Agoric/agoric-sdk/commit/fa0c6a20d6a2be5034853fcac5a4af27c908077f where I added an optional arg to `addBrand` for a pre-existing purse.

dckc (commented):
oops; forgot to submit these comments...

packages/smart-wallet/package.json
```diff
@@ -26,8 +29,8 @@
     "@agoric/nat": "^4.1.0",
     "@agoric/notifier": "^0.4.0",
     "@agoric/store": "^0.7.2",
+    "@agoric/wallet-backend": "^0.12.1",
```
dckc: I wonder why this dep is here.

packages/smart-wallet/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Coalesce updates from a wallet UpdateRecord publication feed. Note that local
+ * state may not reflect the wallet's state if the initial updates are missed.
+ *
+ * If this proves to be a problem we can add an option to this or a related
+ * utility to reset state from RPC.
+ *
+ * @param {ERef<StoredSubscriber<import('./smartWallet').UpdateRecord>>} updates
+ */
+export const coalesceUpdates = updates => {
```
dckc, line 14: This isn't used by the contract, is it?

dckc (approved):
Great work!

packages/smart-wallet/package.json
turadg: Good catch .Moved to dev deps

packages/smart-wallet/src/utils.js
turadg: removed

turadg: Nope just clients

PR #6090, opened 2022-08-30
feat: charter for the PSM, enabling governance

### LINKED ISSUE #6089: Write a charter for multi-asset PSM param governance

`econCommitteeCharter` gives the econ committee the ability to wield governance powers. For the PS0 release, we need a corresponding charter for PSM-only.

The current theory is that it should use a continuing-invitation pattern rather than putting `voteOnFoo` methods in the publicFacet.

## Description of the Design

ContinuingInvitation that gives the ability to invoke `VoteOnParamChange` and `VoteOnPauseOffers`.

## Security Considerations

nothing out of the ordinary.

## Test Plan

just the usual.

Chris-Hibbert:
closes: #6089

## Description

A `Charter` for the PSM to allow the econ committee to govern it.

### Security Considerations

This bundles the ability to vote in the same facet as the ability to call for votes. For the initial release we want both of these to be in the hands of the econ committee.

None

### Testing Considerations

No tests are included.  I don't know how this would get hooked into the bootstrap space, so I'm not sure how invoke it in a test environment.

### DIFF
--- a/packages/inter-protocol/src/psm/psmCharter.js
+++ b/packages/inter-protocol/src/psm/psmCharter.js
@@ -66,9 +66,15 @@ export const start = async (zcf, privateArgs) => {
     VoteOnPauseOffers: makeOfferFilterInvitation,
   });
 
-  const publicFacet = Far('votingAPI', {
-    invitationMakers,
+  const charterMemberHandler = seat => {
+    seat.exit();
+    return invitationMakers;
+  };
+
+  const creatorFacet = Far('psm charter creator', {
+    makeCharterMemberInvitation: () =>
+      zcf.makeInvitation(charterMemberHandler, 'PSM charter member invitation'),
   });
 
-  return harden({ publicFacet });
+  return harden({ creatorFacet });
 };
--- a/packages/inter-protocol/src/psm/psmCharter.js
+++ b/packages/inter-protocol/src/psm/psmCharter.js
@@ -1,0 +1,74 @@
+// @ts-check
+
+import '@agoric/governance/src/exported.js';
+import '@agoric/zoe/exported.js';
+import '@agoric/zoe/src/contracts/exported.js';
+
+import { E, Far } from '@endo/far';
+
+/**
+ * @file
+ *
+ * This contract makes it possible for those who govern the PSM to call for
+ * votes on changes. A more complete implementation would validate parameters,
+ * constrain deadlines and possibly split the ability to call for votes into
+ * separate capabilities for finer grain encapsulation.
+ */
+
+/**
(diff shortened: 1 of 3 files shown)


packages/inter-protocol/src/psm/psmCharter.js
```diff
@@ -0,0 +1,74 @@
...
+ * separate capabilities for finer grain encapsulation.
+ */
+
+/**
+ * @param {ZCF<{binaryVoteCounterInstallation:Installation}>} zcf
+ * @param {{psm:GovernedContractFacetAccess<{},{}>}} privateArgs
+ */
+export const start = async (zcf, privateArgs) => {
+  const { binaryVoteCounterInstallation: counter } = zcf.getTerms();
+  const { psm } = privateArgs;
```
dckc, line 24: Why private? We don't want this legible?

packages/inter-protocol/src/psm/psmCharter.js
```diff
@@ -0,0 +1,74 @@
...
+  const invitationMakers = Far('PSM Invitation Makers', {
+    VoteOnParamChange: makeParamInvitaion,
+    VoteOnPauseOffers: makeOfferFilterInvitation,
+  });
+
+  const publicFacet = Far('votingAPI', {
+    invitationMakers,
+  });
+
+  return harden({ publicFacet });
```
dckc: I think we need to manage invitations like we do for the committee: have a creatorFacet method for getting them... bootstrap then sends them out... then those invitations can be used to get continuing invitations to pose questions.

dckc (commented):
close, but I see a gap...

packages/inter-protocol/src/psm/psmCharter.js
Chris-Hibbert: This is the powerful governed facet that lets you call for votes. It can't be in terms.

AFAICT, no one has a handle to the charter, so they wouldn't be able to look at its terms or call validation methods on it. As it currently stands, only the committee members (and bootstrap) will end up with references to the charter. The committee members can verify the instance and installation they've been given, and that bootstrap only provides access to the ability to solicit votes by wrapping it in a charter. Everyone else can verify how bootstrap hands it out, and whatever we do to make the committee members identities legible.

packages/inter-protocol/src/psm/psmCharter.js
Chris-Hibbert: done.

packages/inter-protocol/src/psm/psmCharter.js
dckc: Ah. I was thinking it was the instance.

dckc (approved):
looks on target

turadg (approved):

PR #6092, opened 2022-08-30
feat: add a continuing invitation to the voter facet

### LINKED ISSUE #6091: Committee provide continuing invitation for use with SmartWallet

The easiest way for the SmartWallet to help the user create offers is to start from a continuing source of invitations. To make this possible for holders of committee seats, the committeeVoterInvitation should include a method that returns an invitation maker. It currently includes only `castBallotFor`.

## Description of the Design

```javascript
  const makeCommitteeVoterInvitation = index => {
    const offerHandler = seat => {
      const voterHandle = makeHandle('Voter');
      seat.exit();
      return Far(`voter${index}`, {
        // CRUCIAL: voteCap carries the ability to cast votes for any voter at
        // any weight. It's wrapped here and given to the voter.
        //
        // Ensure that the voter can't get access to the unwrapped voteCap, and
        // has no control over the voteHandle or weight
         castBallotFor,
        getInvitationMaker,
      });
    };
```

## Security Considerations

The committeeVoter facet will continue to maintain the invariants above, so the voter will be able to vote, but won't have access to more than  her own ballot.

## Test Plan

TBD

Chris-Hibbert:
closes: #6091

## Description

Add a continuing invitation to the voter facet to be used via the SmartWallet.

### Security Considerations

The new method on the voter facet enforces the same attenuation as the previous. While the underlying capability can specify any weight, the facet provided to the user can only vote the owner's actual share.

None. Only relevant to the wallet developer.

### Testing Considerations

Added a test that grabs the voting capability, but doesn't try to use it as a continuing invitation.

### DIFF
--- a/packages/governance/src/committee.js
+++ b/packages/governance/src/committee.js
@@ -50,28 +50,30 @@ const start = (zcf, privateArgs) => {
     const offerHandler = seat => {
       const voterHandle = makeHandle('Voter');
       seat.exit();
-      return Far(`voter${index}`, {
-        // CRUCIAL: voteCap carries the ability to cast votes for any voter at
-        // any weight. It's wrapped here and given to the voter.
-        //
-        // Ensure that the voter can't get access to the unwrapped voteCap, and
-        // has no control over the voteHandle or weight
-        castBallotFor: (questionHandle, positions) => {
-          const { voteCap } = allQuestions.get(questionHandle);
-          return E(voteCap).submitVote(voterHandle, positions, 1n);
-        },
-        getInvitationMaker: () =>
-          Far('invitation maker', {
-            makeVoteInvitation: questionHandle => {
-              const continuingVoteHandler = (_seat, { positions }) => {
-                _seat.exit();
-                const { voteCap } = allQuestions.get(questionHandle);
-                return E(voteCap).submitVote(voterHandle, positions, 1n);
-              };
 
-              return zcf.makeInvitation(continuingVoteHandler, 'vote');
-            },
-          }),
+      // CRUCIAL: voteCap carries the ability to cast votes for any voter at
+      // any weight. It's wrapped here and given to the voter.
+      //
+      // Ensure that the voter can't get access to the unwrapped voteCap, and
+      // has no control over the voteHandle or weight
+      return harden({
+        voter: Far(`voter${index}`, {
+          castBallotFor: (questionHandle, positions) => {
+            const { voteCap } = allQuestions.get(questionHandle);
+            return E(voteCap).submitVote(voterHandle, positions, 1n);
+          },
+        }),
+        invitationMakers: Far('invitation makers', {
+          makeVoteInvitation: questionHandle => {
+            const continuingVoteHandler = (_seat, { positions }) => {
+              _seat.exit();
+              const { voteCap } = allQuestions.get(questionHandle);
+              return E(voteCap).submitVote(voterHandle, positions, 1n);
+            };
+
+            return zcf.makeInvitation(continuingVoteHandler, 'vote');
+          },
+        }),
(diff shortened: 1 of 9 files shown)


packages/governance/src/committee.js
```diff
@@ -60,6 +60,18 @@ const start = (zcf, privateArgs) => {
           const { voteCap } = allQuestions.get(questionHandle);
           return E(voteCap).submitVote(voterHandle, positions, 1n);
         },
+        getInvitationMaker: () =>
```
turadg: can we match the pattern in https://github.com/Agoric/agoric-sdk/blob/f016c77e3af9b917f20b03ac4eeee1f761a53852/packages/inter-protocol/src/vaultFactory/vaultKit.js#L27-L31 ? 

I think that would mean here,
```suggestion
        invitationMakers => Far('invitation makers', {
```
And then because the offer result has to be all objects or all functions, we'd make the offer result be,
```
  voter: Far(`voter${index}`, {
  …
  invitationMakers => Far('invitation makers', {
```

Chris-Hibbert: Thanks for the hint. It was close enough that I was able to get there.

packages/governance/test/unitTests/test-paramGovernance.js
```diff
@@ -370,3 +371,68 @@ test('change multiple params used invitation', async t => {
...
+
+  const details = await detailsP;
+  const positive = details.positions[0];
+  const invitations = await E(committeeCreator).getVoterInvitations();
+
+  const seat = E(zoe).offer(invitations[0]);
+  const { invitationMakers } = E.get(E(seat).getOfferResult());
+  const voteInvitation = E(invitationMakers).makeVoteInvitation(
+    details.questionHandle,
+  );
```
dckc, line 420: yep! that looks right.

dckc (approved):
looks good!

I'm a little uneasy that there were not more tests that needed updating. I suspect that I have some scripts that depend on the former API that will break. But that's what I get for not maintaining tests for them.

packages/governance/test/swingsetTests/contractGovernor/vat-voter.js
```diff
@@ -18,7 +18,7 @@ const build = async (log, zoe) => {
   return Far('voter', {
     createVoter: async (name, invitation) => {
       const seat = E(zoe).offer(invitation);
-      const voteFacet = E(seat).getOfferResult();
+      const { voter: voteFacet } = E.get(E(seat).getOfferResult());
```
turadg: nit: cognitive load to rename `voter` to `voteFacet`. consider using the `voter` name through this function

packages/governance/test/unitTests/test-paramGovernance.js
```diff
@@ -133,7 +133,7 @@ const setUpVoterAndVote = async (committeeCreator, zoe, qHandle, choice) => {
   const invitations = await E(committeeCreator).getVoterInvitations();
 
   const seat = E(zoe).offer(invitations[0]);
-  const voteFacet = E(seat).getOfferResult();
+  const voteFacet = E.get(E(seat).getOfferResult()).voter;
```
turadg: ditto on suggestion above to use `voter`

turadg (approved):

packages/governance/test/swingsetTests/contractGovernor/vat-voter.js
Chris-Hibbert: done

packages/governance/test/unitTests/test-paramGovernance.js
Chris-Hibbert: done

PR #6094, opened 2022-08-31
feat: hold onto adminFacets in bootstrap

### LINKED ISSUE #6034: bootstrap does not save adminFacets from startInstance etc.
## Describe the bug

I was looking at some code with @Chris-Hibbert  `const { creatorFacet } = await E(zoe).startInstance(...)`

(meeting time.  IOU more)

## Platform Environment

## Screenshots

Chris-Hibbert:
closes: #6034

## Description

Save the AdminFacets from the core economy contracts during bootstrap.

This saves them in spaces, so they can be used by big-hammer scripts.

This makes all the contracts started in bootstrap use a consistent format for saving their public, private, governance and admin facets. Rather than all being top level names, each contract has a single top-level name, like `ammFacets`, which contains `{ creatorFacet, governorCreatorFacet, adminFacet, publicFacet }`, and sometimes an `instance`.

`feeDistributor` and `psmCharter` are not governed, so they don't have `adminFacet`s, and `psmCharter` has no `publicFacet`.

### Security Considerations

Wouldn't want to misplace these.

None.

### Testing Considerations

Untested.

### DIFF
--- a/packages/inter-protocol/src/proposals/addAssetToVault.js
+++ b/packages/inter-protocol/src/proposals/addAssetToVault.js
@@ -244,7 +244,7 @@ export const registerScaledPriceAuthority = async (
  */
 export const addAssetToVault = async (
   {
-    consume: { vaultFactoryFacets, agoricNamesAdmin },
+    consume: { vaultFactoryKit, agoricNamesAdmin },
     brand: {
       consume: { [Stable.symbol]: stableP },
     },
@@ -260,7 +260,7 @@ export const addAssetToVault = async (
   );
 
   const stable = await stableP;
-  const vaultFactoryCreator = E.get(vaultFactoryFacets).creatorFacet;
+  const vaultFactoryCreator = E.get(vaultFactoryKit).creatorFacet;
   await E(vaultFactoryCreator).addVaultType(interchainIssuer, oracleBrand, {
     debtLimit: AmountMath.make(stable, 0n),
     // the rest of these are arbitrary, TBD by gov cttee
@@ -320,7 +320,7 @@ export const getManifestForAddAssetToVault = (
       },
       [addAssetToVault.name]: {
         consume: {
-          vaultFactoryFacets: true,
+          vaultFactoryKit: true,
           agoricNamesAdmin: true,
         },
         brand: {
--- a/packages/inter-protocol/src/proposals/addAssetToVault.js
+++ b/packages/inter-protocol/src/proposals/addAssetToVault.js
@@ -244,7 +244,7 @@ export const registerScaledPriceAuthority = async (
  */
 export const addAssetToVault = async (
   {
-    consume: { vaultFactoryCreator, agoricNamesAdmin },
+    consume: { vaultFactoryFacets, agoricNamesAdmin },
     brand: {
       consume: { [Stable.symbol]: stableP },
     },
@@ -260,6 +260,7 @@ export const addAssetToVault = async (
   );
 
   const stable = await stableP;
+  const vaultFactoryCreator = E.get(vaultFactoryFacets).creatorFacet;
   await E(vaultFactoryCreator).addVaultType(interchainIssuer, oracleBrand, {
     debtLimit: AmountMath.make(stable, 0n),
     // the rest of these are arbitrary, TBD by gov cttee
@@ -319,9 +320,7 @@ export const getManifestForAddAssetToVault = (
       },
       [addAssetToVault.name]: {
         consume: {
-          vaultFactoryCreator: true,
-          ammCreatorFacet: true,
-          reserveCreatorFacet: true,
+          vaultFactoryFacets: true,
           agoricNamesAdmin: true,
         },
         brand: {
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -6,7 +6,7 @@ import '@agoric/vats/exported.js';
 import '@agoric/vats/src/core/types.js';
 import { makeStorageNodeChild } from '@agoric/vats/src/lib-chainStorage.js';
 import { makeRatio } from '@agoric/zoe/src/contractSupport/index.js';
-import { E, Far } from '@endo/far';
+import { E } from '@endo/far';
 import { Stable, Stake } from '@agoric/vats/src/tokens.js';
 import { deeplyFulfilledObject } from '@agoric/internal';
 import * as Collect from '../collect.js';
@@ -32,6 +32,7 @@ const MILLI = 1_000_000n;
  * @typedef {GovernedCreatorFacet<import('../stakeFactory/stakeFactory.js').StakeFactoryCreator>} StakeFactoryCreator
  * @typedef {import('../stakeFactory/stakeFactory.js').StakeFactoryPublic} StakeFactoryPublic
  * @typedef {import('../reserve/assetReserve.js').GovernedAssetReserveFacetAccess} GovernedAssetReserveFacetAccess
+ * @typedef {import('../vaultFactory/vaultFactory.js').VaultFactoryContract['publicFacet']} VaultFactoryPublicFacet
  */
 
 /**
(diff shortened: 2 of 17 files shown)


dckc:
@Chris-Hibbert let me know if getting the types updated is as tricky as I suspect it might be. They mostly live in `packages/vats/src/core/types.js`.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -46,28 +46,43 @@ const sanitizePathSegment = name => {
...
+ */
+
 /**
  * @typedef { WellKnownSpaces & ChainBootstrapSpace & EconomyBootstrapSpace
  * } EconomyBootstrapPowers
  * @typedef {PromiseSpaceOf<{
  *   ammInstanceWithoutReserve: Instance,
  *   ammCreatorFacet: XYKAMMCreatorFacet,
  *   ammGovernorCreatorFacet: GovernedContractFacetAccess<XYKAMMPublicFacet,XYKAMMCreatorFacet>,
+ *   ammGovernorAdminFacet: AdminFacet,
```
dckc: hm. I suppose this works, but what I had in mind was to capture the whole result of `E(zoe).startInstance(ammGovernorInstallation)` in one field, replacing `ammGovernorCreatorFacet`.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -46,28 +46,43 @@ const sanitizePathSegment = name => {
...
  *   economicCommitteeCreatorFacet: CommitteeElectorateCreatorFacet,
  *   feeDistributorCreatorFacet: import('../feeDistributor.js').FeeDistributorCreatorFacet,
  *   feeDistributorPublicFacet: import('../feeDistributor.js').FeeDistributorPublicFacet,
```
dckc: all 3 feeDistributor facets would fit in one neat package :)

dckc (commented):
This is good progress, but there are relevant calls to `startInstance()` in `packages/vats` too; for example, in `addBankAssets()` in `packages/vats/src/core/basic-behaviors.js`.

There's one in `mintInitialSupply` too... though that's a short-lived contract by design. Let's see... that TODO about shutting it down is done in `packages/vats/src/centralSupply.js`, I see.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -500,22 +519,25 @@ export const startVaultFactory = async (
...
-          initialShortfallInvitation,
-          marshaller,
-          storageNode,
-        },
-      }),
-    );
+  const {
+    creatorFacet: governorCreatorFacet,
+    instance: governorInstance,
+    adminFacet,
```
dtribble, line 461: Hwo will this style work with multiple PSMs?

dckc: It won't. hence #6146

packages/inter-protocol/src/proposals/addAssetToVault.js
```diff
@@ -320,9 +321,7 @@ export const getManifestForAddAssetToVault = (
       },
       [addAssetToVault.name]: {
         consume: {
-          vaultFactoryCreator: true,
-          ammCreatorFacet: true,
-          reserveCreatorFacet: true,
+          vaultFactoryFacets: true,
```
dckc: :+1:

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: Done.

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: Pervasively done.

turadg (approved):
👍 to organizing. I have some hesitation to repeating 'Facets' and wonder if we could drop that from the grouping without confusing things. E.g. that `vaultFactoryFacets.governorCreatorFacet` be simply `vaultFactory.governorCreatorFacet`.

For the git history "save" made me think these objects were going into durable storage. Upon reading the code I see they're just being held in the space. Maybe "hold" would be less at risk of confusion.

packages/vats/src/core/startWalletFactory.js
```diff
@@ -146,12 +146,7 @@ export const startWalletFactory = async (
       consume: { [Stable.symbol]: feeIssuerP },
     },
   },
-  {
-    options: {
-      perAccountInitialValue = (StableUnit * 25n) / 100n,
-      walletBridgeId = BRIDGE_ID.PROVISION_SMART_WALLET,
```
dckc: are you sure this should be removed?

dckc (changes requested):
removing the `walletBridgeId` setting looks wrong

packages/vats/src/core/startWalletFactory.js
Chris-Hibbert: lint complained that walletBridgeId is unused. How do I make it happy without removing it?

packages/vats/src/core/startWalletFactory.js
```diff
@@ -238,8 +236,8 @@ export const startWalletFactory = async (
   });
 
   await Promise.all([
-    E(bridgeManager).register(walletBridgeId, handler),
-    E(E.get(econCharterStartResult).creatorFacet).addInstance(
+    E(bridgeManager).register(BRIDGE_ID.PROVISION, handler),
```
dckc: > How do I make it happy ...?

undo this change too; put it back to `walletBridgeId`.

packages/vats/src/core/startWalletFactory.js
Chris-Hibbert: reverted

packages/vats/src/core/startWalletFactory.js
Chris-Hibbert: done.

Chris-Hibbert:
@turadg, @dckc  Please take another look. A fair amount of work was required to bring this up-to-date. I'd like to merge it before the underlying code migrates again.

packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -312,29 +312,6 @@ export const installGovAndPSMContracts = async ({
   );
 };
 
-/**
- * @deprecated the PSM charter has been merged with the econ charter
- * @param {EconomyBootstrapPowers} powers
- */
-export const startPSMCharter = async ({
```
dckc: Isn't that still used by `packages/vats/src/core/boot-psm.js`? Ah... not as of fa9c030d27839d

dckc (approved):
yes, this looks ready to go.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -47,23 +48,41 @@ const MILLI = 1_000_000n;
...
  *   psmFacets: MapStore<Brand, PSMFacets>,
- *   econCharterStartResult: EconCharterStartResult,
- *   reservePublicFacet: import('../reserve/assetReserve.js').AssetReservePublicFacet,
- *   reserveCreatorFacet: import('../reserve/assetReserve.js').AssetReserveLimitedCreatorFacet,
- *   reserveGovernorCreatorFacet: GovernedAssetReserveFacetAccess,
- *   stakeFactoryCreatorFacet: StakeFactoryCreator,
- *   vaultFactoryCreator: VaultFactoryCreatorFacet,
- *   vaultFactoryGovernorCreator: GovernedContractFacetAccess<{},{}>,
- *   vaultFactoryVoteCreator: unknown,
+ *   econCharterFacets: EconCharterStartResult,
```
turadg: `SomeContractFacets` is the same as the results of its start function. In this PR,`someContractFacets` is used for the property at `SomeContractStartResult` for the type. Would it be better to be consistent between these cases?
```suggestion
 *   econCharterFacets: EconCharterFacets,
```

Or alternately,
```suggestion
 *   econCharterStartResult: EconCharterStartResult,
```

One argument against the latter is that it might read as the result of `startInstance` which would be inaccurate as that returns `StartedInstanceKit` which has `instance` and `adminFacet` too. Though that brings up another potential confusion: does `SomeContractFacets` include `adminFacet` or not? If not, then `SomeContractFacets` would be misleading.

This all isn't a blocker. Now seems like a good time to come up with consistent terminology for these types and bindings but it could be punted out to a ticket to finalize and implement.

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -268,16 +284,17 @@ export const setupReserve = async ({
...
       poserInvitationP,
       E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
       feeMintAccessP,
+      ammFacetsP,
     ]);
 
   const reserveTerms = makeReserveTerms(
     poserInvitationAmount,
-    ammInstanceWithoutReserve,
+    ammFacets.instanceWithoutReserve,
```
turadg: this is confusing because `instance` isn't a facet. Maybe just call this `ammKit` to be less specific?

packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -647,43 +674,50 @@ export const startRewardDistributor = async ({
       return undefined;
     });
 
-  const feeDistributorFacets = await E(zoe).startInstance(
+  const facets = await E(zoe).startInstance(
```
turadg: please rename to match https://github.com/Agoric/agoric-sdk/blob/0c6bc9077a852d2a0b0bc6ee1c490274252e2cf3/packages/zoe/src/zoeService/utils.d.ts#L72-L77

```suggestion
  const instanceKit = await E(zoe).startInstance(
```

turadg (commented):
I have some concerns around naming which lead me to not Approve it as is. I understand that this feature needs to get in and it's prone to merge conflicts so I'm not requesting changes, allowing Dan's extant approval to suffice.

packages/inter-protocol/src/proposals/econ-behaviors.js
dckc: In at least some cases, the collection of facets is not exactly the result of `startInstance` but also includes, for example, a governor creator facet.

packages/inter-protocol/src/proposals/econ-behaviors.js
dckc: true. `Kit` is also shorter (both in characters and in syllables) than `Facets`.

If it were a more widely exposed API, I might feel more strongly about it, but in this case, I don't mind much.

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: done

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: Yeah, It's definitely not the same as `*StartResult`, but it's also not composed of only things that are facets. I'm comfortable with `*Kit`. That naming convention doesn't seem to clash with other variable names.

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: I'm convinced.

turadg (commented):
👍

PR #6095, opened 2022-08-31
fix(wallet/contract): bounded updates: recent offers / payments only

### LINKED ISSUE #6065: smart wallet state grows with each offer

dckc:
closes: #6065 

## Description

Rather than further work on pruning outdated state, the approach here is to notify only about the most recently updated payment or offer. We can get the rest from historical data.

There remains a case of batch-updates to offers. It wasn't a problem in a test of 50 trades, up to 15 at a time.

This also fixes a problem where every status update for a payment caused the record to get deeper and deeper.

The list of purses only grows via VBANK updates, so it's not likely to get out of hand.

### Security Considerations

none, I believe.

I updated the `smartWallet.js` notification API docs. Clients will have to know to look thru history.

[psm-tool.js](https://gist.github.com/dckc/dfc2182a59b11ec4527349afd5ca050f#file-psm-tool-js) is a compatible client.

### Testing Considerations

I did several integration test runs and updated relevant unit tests.


packages/wallet/api/src/lib-wallet.js
```diff
@@ -523,7 +524,7 @@ export function makeWalletRoot({
     inboxState.set(id, offerForDisplay);
     if (doPush) {
       // Only trigger a state change if this was a single update.
-      offersUpdater.updateState([...inboxState.values()]);
+      offersUpdater.updateState([offerForDisplay]);
```
turadg, line 527: this doesn't help with `updateAllInboxState`. Missed or deferred?

packages/wallet/contract/src/smartWallet.js
```diff
@@ -50,9 +50,13 @@ export const makeSmartWallet = async (
   const notifierParts = {
     contacts: E(admin).getContactsNotifier(),
     dapps: E(admin).getDappsNotifier(),
+    /** all current issuers */
     issuers: E(admin).getIssuersNotifier(),
+    /** recently updated offers */
```
turadg: "recently" seems unnecessarily vague. It's exactly the latest offer in some cases and it's all offers when `updateAllState` is called (whenever petnames are added or changed).

packages/wallet/contract/src/smartWallet.js
```diff
@@ -50,9 +50,13 @@ export const makeSmartWallet = async (
...
     offers: E(admin).getOffersNotifier(),
+    /** recently updated payments */
```
turadg: singular "most recently payment". can still be an array for backwards compatibility and comment might explain that.

packages/wallet/api/src/lib-wallet.js
dckc: um... deferred... as in... I'm not sure what that thing does... let's see if it's a problem in practice... moderate integration testing didn't turn up anything bad.

packages/wallet/contract/src/smartWallet.js
dckc: ah. is that how it works? thanks!

turadg (approved):
LGTM

Issue #6099, opened 2022-08-31
verify by end-to-end test a contract on-chain can be upgraded


Chris-Hibbert:
## What is the Problem Being Solved?

After the bulldozer, the only way to upgrade contracts will be through the AdminFacet's `upgradeContract()`. We don't have experience yet executing that in a live chain.

## Description of the Design

A chain integration test that does the following:
- bundle and install a buggy contract via a proposal
- use a smart wallet to demonstrate the bug
- bundle and upgrade to a contract that preserves the buggy one's state but fixes the bug
- use a smart wallet to demonstrate that the state is preserved and the bug is fixed

## Security Considerations

The relevant security here is the feeling of confidence that we can do something about it if a problem is discovered in a contract on a live chain.

## Test Plan

Demonstrate one contract being upgraded on a running chain.

dckc:
@kriskowal help me try `agoric bundle` and `agoric publish` on ollinet, please?

kriskowal:
> @kriskowal help me try `agoric bundle` and `agoric publish` on ollinet, please?

I’ll implement them right away https://github.com/Agoric/agoric-sdk/issues/6087

dckc:
needs #5460 ; currently `agd query vstorage children published` doesn't include `installation`.

kriskowal:
I’ve made a first attempt to upgrade PSM and encountered an error in upgradeContract behavior https://github.com/Agoric/agoric-sdk/issues/6207

Tartuffo:
Moving this out of the critical path for the PSM-First launch.  We can side-grade the PSM to a new contract: 

1. Turn off the offer sides of the old PSM contact
2. Stand up a new PSM contract
3. Use JS Hammer to:
 a. mint a bunch of IST
 b. buy the USDC from the old contract
 c. put the USDC in a new PSM
 d. burn the bunch of IST
4. Tear down the old contract

dckc:
we also have a couple other options:
 - upgrade the contract to a PSM with durable governance (#5200) 
 - restart the whole VM (#6644)

turadg:
@dckc going to take this on because @kriskowal is out. I'm taking on a dependency,
- https://github.com/Agoric/agoric-sdk/issues/7466

Tartuffo:
Congrats everyone!  This is a major accomplishment.

PR #6101, opened 2022-08-31
feat: save PSM adminFacet in bootstrap

### LINKED ISSUE #6100: Save PSM adminFacet to enable big-hammer upgrade

In order for a javascript program running in the bootstrap environment to be able to upgrade the PSM (see #6099), the adminFacet will need to be available there.

## Description of the Design

Export the AdminFacet from the contractGovernor, and save it in the bootstrap environment.

## Security Considerations

POLA pushes us to want fine-grained access to the caps in bootstrap.

## Test Plan

See #6099

Chris-Hibbert:
closes: #6100

## Description

Save the PSM adminFacet from bootstrap. Also, the contractGovernor will now make the adminFacet available when other contracts need it.

### Security Considerations

The admin facet is very powerful. The bootstrap environment is considered a secure context in which to hold powerful facets for the Econ Committee's use.

None.

### Testing Considerations

deferred to #6099.

### DIFF
--- a/packages/governance/src/contractGovernor.js
+++ b/packages/governance/src/contractGovernor.js
@@ -163,6 +163,7 @@ const start = async (zcf, privateArgs) => {
     creatorFacet: governedCF,
     instance: governedInstance,
     publicFacet: governedPF,
+    adminFacet,
   } = await E(zoe).startInstance(
     governedContractInstallation,
     governedIssuerKeywordRecord,
@@ -281,6 +282,7 @@ const start = async (zcf, privateArgs) => {
     voteOnApiInvocation,
     voteOnOfferFilter: voteOnFilter,
     getCreatorFacet: () => limitedCreatorFacet,
+    getAdminFacet: () => adminFacet,
     getInstance: () => governedInstance,
     getPublicFacet: () => governedPF,
   });
--- a/packages/inter-protocol/src/proposals/econ-behaviors.js
+++ b/packages/inter-protocol/src/proposals/econ-behaviors.js
@@ -41,15 +41,6 @@ const sanitizePathSegment = name => {
 };
 
 /**
- * There's an obsolete version in utils.d.ts
- *
- * @typedef {object} AdminFacet
- * @property {() => PromiseLike} getVatShutdownPromise
- * @property {(contractBundleId: string, newPrivateArgs: any) => void} upgradeContract
- * @property {(newPrivateArgs:any) => void} restartContract
- */
-
-/**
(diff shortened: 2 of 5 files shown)


packages/inter-protocol/src/proposals/econ-behaviors.js
```diff
@@ -40,6 +40,15 @@ const sanitizePathSegment = name => {
   return candidate;
 };
 
+/**
+ * There's an obsolete version in utils.d.ts
```
turadg: please put these updates there instead so there's one correct one and no incorrect ones.
```ts
type AdminFacet = {
  getVatShutdownPromise: () => Promise<any>; // Completion, which is currently any
  upgradeContract: (contractBundleId: string, newPrivateArgs: any) => void;
  restartContract: (newPrivateArgs: any) => void;
};
```

packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -138,14 +138,11 @@ export const startPSM = async (
     }),
   );
 
-  const governedInstance = await E(governorFacets.creatorFacet).getInstance();
-  const creatorFacet = E(governorFacets.creatorFacet).getCreatorFacet();
-
-  psmInstanceR.resolve(governedInstance);
+  psmInstanceR.resolve(await E(governorFacets.creatorFacet).getInstance());
```
turadg, line 141: with all these facets, consider having a single entity in the space that is the result of `zoe.startInstance`.

turadg (changes requested):

packages/inter-protocol/src/proposals/startPSM.js
dckc: yeah; we tried that... ran into problems around promises or something... the answer isn't obvious, so we agreed to defer.

packages/governance/src/contractGovernor.js
```diff
@@ -281,6 +282,7 @@ const start = async (zcf, privateArgs) => {
     voteOnApiInvocation,
     voteOnOfferFilter: voteOnFilter,
     getCreatorFacet: () => limitedCreatorFacet,
+    getAdminFacet: () => adminFacet,
```
dckc, line 285: This gave me some pause... we talked it over with @erights and agreed it's appropriate.

dckc (approved):
This looks like it does what we need for now.

packages/inter-protocol/src/proposals/econ-behaviors.js
Chris-Hibbert: done

turadg (approved):

PR #5900, opened 2022-08-05
test(vaultFactory): refactor to macro

turadg:
## Description

Refactor some inter-protocol tests to use [Ava macros](https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#reusing-test-logic-through-macros)

Note that the `test.macro` introduced in Ava 4 breaks under `prepare-test-env-ava` until https://github.com/endojs/endo/issues/1235.

Also bumps the `peerDependency` in swingset-vat.

### Security Considerations

--

### Documentation Considerations

--

### Testing Considerations

--

PR #4327, opened 2022-01-20
WIP: Use delta list rather than stagings

erights:
Fixes #3850 borrowing elements from the zoe-2 design.

Revisiting the example from #3850 

```js
brandOutPoolSeat.decrementBy({ Secondary: amountOut });
seat.incrementBy({ Out: amountOut });

brandOutPoolSeat.incrementBy(
  brandInPoolSeat.decrementBy({ Central: reducedCentralAmount }));

seat.decrementBy({ In: reducedAmountIn });
brandInPoolSeat.incrementBy({ Secondary: reducedAmountIn });

zcf.reallocate(brandOutPoolSeat, brandInPoolSeat, seat);
```
we'd like to instead say something like
```js
tranferAssets(zcf, harden([
  { seat: brandOutPoolSeat, subtract: { Secondary: amountOut } },
  { seat, add: { Out: amountOut } },

  { seat: brandInPoolSeat, subtract: { Central: reducedCentralAmount } },
  { seat: brandOutPoolSeat, add: { Central: reducedCentralAmount } },

  { seat, subtract: { In: reducedAmountIn } },
  { seat: brandInPoolSeat, add: { Secondary: reducedAmountIn } },
]));
```
Since the argument is just an array, if we don't have it all up front, we could accumulate it incrementally until we're ready to act on it:
```js
const deltas = [];
// ...
deltas.push({ seat: brandInPoolSeat, subtract: { Central: reducedCentralAmount } });
// ...
transferAssets(zcf, harden(deltas));
```

For another approach, see https://github.com/Agoric/agoric-sdk/pull/6577

PR #6102, opened 2022-09-01
fix(inter-protocol): ensure PSM availability after staging failure

### LINKED ISSUE #6096: failed PSM trade prevents further service due to inconsistent seat staging
## Describe the bug

After accidentally trying to buy AUSD whent he PSM had none, attempts to trade 20 IST for USDC result in a refund rather than a trade.

## To Reproduce

In `agoricollinet-22`, check wallet balances and then attempt to trade 20 IST for USDC:

```
$ psm-tool --net ollinet --wallet $addr
{
  "balances": [
    [null,"zoe invite"],
    [1.99747,"AUSD"],
    [0,"BLD"],
    [46.800225,"IST"]
  ],
...
}

$ psm-tool --giveStable 20 --boardId $psmInstance --feePct 0.05 >,give20.json
$ agd --from=k1 tx swingset wallet-action --allow-spend "$(cat ,give20.json)"  --node https://ollinet.rpc.agoric.net:443 --chain-id agoricollinet-22 -y
code: 0
...
txhash: 07A7B0CCCD655F0EF2F8ED566380E23D23D6D665B0308671346A250880C6F3FF
```

Note that 20 IST is withdrawn for this trade:
```
$ psm-tool --net ollinet --wallet $addr
{
  "balances": [
    [null,"zoe invite"],
    [1.99747,"AUSD"],
    [0,"BLD"],
    [26.800225,"IST"]
  ],
  "offers": [
    ["2022-08-31T14:25:13.203Z","pending","psm","giveStable",{"give":{"In":[20,"IST"]},"want":{"Out":[19.990005,"AUSD"]}}]
  ]
}
```

But then it gets refunded:
```
$ psm-tool --net ollinet --wallet $addr
{
  "balances": [
    [null,"zoe invite"],
    [1.99747,"AUSD"],
    [0,"BLD"],
    [46.800225,"IST"]
  ],
  "offers": [
    ["2022-08-31T14:25:13.203Z","accept","psm","giveStable",{"give":{"In":[20,"IST"]},"want":{"Out":[19.990005,"AUSD"]}}]
  ]
}
```

Use metrics to check PSM pool balance:
```
$ agoric follow -B https://ollinet.agoric.net/network-config :published.psm.IST.AUSD.metrics
{
  anchorPoolBalance: {
    brand: slot(0,"Alleged: AUSD brand"),
    value: 23002530n,
  },
  feePoolBalance: {
    brand: slot(1,"Alleged: IST brand"),
    value: 2305n,
  },
  totalAnchorProvided: {
    brand: slot(0),
    value: 0n,
  },
  totalStableProvided: {
    brand: slot(1),
    value: 23002530n,
  },
}
```

## Expected behavior

Take my money! And give me USDC in return :)

## Platform Environment
ollinet, which is running [1af6550](https://github.com/Agoric/agoric-sdk/commit/1af6550688343129d5a458e7220898a86dd5503f)

cc @turadg @dtribble 

blocking #6064

dckc:
closes: #6096

## Description

When an attempt to trade IST for anchor/reference token fails due to lack of anchor funds in the PSM, seat stagings were left in an inconsistent state, preventing further service. This fix clears the seat stagings on failure.

@erights @Chris-Hibbert this also includes a couple zoe tweaks that @dtribble helped me with.

### Security Considerations

availability is good :)

### Testing Considerations

added a reasonably thorough unit test

### DIFF
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -160,11 +160,24 @@ export const start = async (zcf, privateArgs, baggage) => {
       AmountMath.isGTE(maxAnchor, wanted),
       X`wanted ${wanted} is more then ${given} minus fees ${fee}`,
     );
-    stageTransfer(seat, stage, { In: afterFee }, { Stable: afterFee });
-    stageTransfer(seat, feePool, { In: fee }, { Stable: fee });
-    stageTransfer(anchorPool, seat, { Anchor: maxAnchor }, { Out: maxAnchor });
-    zcf.reallocate(seat, anchorPool, stage, feePool);
-    stableMint.burnLosses({ Stable: afterFee }, stage);
+    try {
+      stageTransfer(seat, stage, { In: afterFee }, { Stable: afterFee });
+      stageTransfer(seat, feePool, { In: fee }, { Stable: fee });
+      stageTransfer(
+        anchorPool,
+        seat,
+        { Anchor: maxAnchor },
+        { Out: maxAnchor },
+      );
+      zcf.reallocate(seat, anchorPool, stage, feePool);
+      stableMint.burnLosses({ Stable: afterFee }, stage);
+    } catch (e) {
+      stage.clear();
+      anchorPool.clear();
+      feePool.clear();
+      // NOTE someday, reallocate should guarantee that this case cannot happen
+      throw e;
+    }
     totalAnchorProvided = AmountMath.add(totalAnchorProvided, maxAnchor);
   };
 
@@ -183,12 +196,15 @@ export const start = async (zcf, privateArgs, baggage) => {
       X`wanted ${wanted} is more then ${given} minus fees ${fee}`,
     );
     stableMint.mintGains({ Stable: asStable }, stage);
-    stageTransfer(seat, anchorPool, { In: given }, { Anchor: given });
-    stageTransfer(stage, seat, { Stable: afterFee }, { Out: afterFee });
-    stageTransfer(stage, feePool, { Stable: fee });
     try {
+      stageTransfer(seat, anchorPool, { In: given }, { Anchor: given });
+      stageTransfer(stage, seat, { Stable: afterFee }, { Out: afterFee });
+      stageTransfer(stage, feePool, { Stable: fee });
       zcf.reallocate(seat, anchorPool, stage, feePool);
     } catch (e) {
+      stage.clear();
+      anchorPool.clear();
+      feePool.clear();
       // NOTE someday, reallocate should guarantee that this case cannot happen
       stableMint.burnLosses({ Stable: asStable }, stage);
       throw e;
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -155,7 +155,6 @@ export const start = async (zcf, privateArgs, baggage) => {
     const fee = ceilMultiplyBy(given, params.getGiveStableFee());
     const afterFee = AmountMath.subtract(given, fee);
     const maxAnchor = floorMultiplyBy(afterFee, anchorPerStable);
-    // TODO this prevents the reallocate from failing. Can this be tested otherwise?
     assert(
       AmountMath.isGTE(maxAnchor, wanted),
       X`wanted ${wanted} is more then ${given} minus fees ${fee}`,
@@ -175,7 +174,7 @@ export const start = async (zcf, privateArgs, baggage) => {
       stage.clear();
       anchorPool.clear();
       feePool.clear();
-      // NOTE someday, reallocate should guarantee that this case cannot happen
+      // TODO(#6116) someday, reallocate should guarantee that this case cannot happen
       throw e;
     }
     totalAnchorProvided = AmountMath.add(totalAnchorProvided, maxAnchor);
@@ -205,7 +204,7 @@ export const start = async (zcf, privateArgs, baggage) => {
       stage.clear();
       anchorPool.clear();
       feePool.clear();
-      // NOTE someday, reallocate should guarantee that this case cannot happen
+      // TODO(#6116) someday, reallocate should guarantee that this case cannot happen
       stableMint.burnLosses({ Stable: asStable }, stage);
       throw e;
     }
--- a/packages/inter-protocol/test/psm/test-psm.js
+++ b/packages/inter-protocol/test/psm/test-psm.js
@@ -16,6 +16,7 @@ import {
   makeRatio,
   natSafeMath as NatMath,
 } from '@agoric/zoe/src/contractSupport/index.js';
+import centralSupplyBundle from '@agoric/vats/bundles/bundle-centralSupply.js';
 import { E } from '@endo/eventual-send';
 import { NonNullish } from '@agoric/assert';
 import path from 'path';
@@ -23,6 +24,7 @@ import { eventLoopIteration } from '@agoric/zoe/tools/eventLoopIteration.js';
 import { makeTracer } from '../../src/makeTracer.js';
 import {
   makeMockChainStorageRoot,
+  mintRunPayment,
   setUpZoeForTest,
   subscriptionKey,
   withAmountUtils,
@@ -90,6 +92,8 @@ const makeTestContext = async () => {
 
   const committeeInstall = await E(zoe).install(committeeBundle);
   const psmInstall = await E(zoe).install(psmBundle);
+  const centralSupply = await E(zoe).install(centralSupplyBundle);
+
   const mintLimit = AmountMath.make(anchor.brand, MINT_LIMIT);
 
   const marshaller = makeBoard().getReadonlyMarshaller();
@@ -119,7 +123,7 @@ const makeTestContext = async () => {
     initialPoserInvitation,
     stable: { issuer: stableIssuer, brand: stableBrand },
     anchor,
-    installs: { committeeInstall, psmInstall },
+    installs: { committeeInstall, psmInstall, centralSupply },
     mintLimit,
     marshaller,
     terms: {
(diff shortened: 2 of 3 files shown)


packages/inter-protocol/src/psm/psm.js
```diff
@@ -160,11 +160,24 @@ export const start = async (zcf, privateArgs, baggage) => {
...
+        { Anchor: maxAnchor },
+        { Out: maxAnchor },
+      );
+      zcf.reallocate(seat, anchorPool, stage, feePool);
+      stableMint.burnLosses({ Stable: afterFee }, stage);
+    } catch (e) {
+      stage.clear();
+      anchorPool.clear();
+      feePool.clear();
+      // NOTE someday, reallocate should guarantee that this case cannot happen
```
turadg: ticket? we probably will want to reference it often and it can serve to teach why and when to clear.

packages/inter-protocol/src/psm/psm.js
```diff
@@ -183,12 +196,15 @@ export const start = async (zcf, privateArgs, baggage) => {
...
-    stageTransfer(stage, feePool, { Stable: fee });
     try {
+      stageTransfer(seat, anchorPool, { In: given }, { Anchor: given });
+      stageTransfer(stage, seat, { Stable: afterFee }, { Out: afterFee });
+      stageTransfer(stage, feePool, { Stable: fee });
       zcf.reallocate(seat, anchorPool, stage, feePool);
     } catch (e) {
+      stage.clear();
+      anchorPool.clear();
+      feePool.clear();
```
turadg, line 206: could we codify this pattern now with a contract helper? something like,
```js
atomicReallocate(zcf, [
  [seat, anchorPool, { In: given }, { Anchor: given }],
  [stage, seat, { Stable: afterFee }, { Out: afterFee }],
  [stage, feePool, { Stable: fee }],
]);
```

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -209,29 +213,49 @@ async function makePsmDriver(t, customTerms) {
       return feePayoutAmount;
     },
 
-    /** @param {Amount<'nat'>} giveAnchor */
-    async swapAnchorForStable(giveAnchor) {
+    /**
+     * @param {Amount<'nat'>} giveAnchor
+     * @param {Amount<'nat'>} [wantStable]
+     * @param {*} [out]
```
turadg: yay for typing. this param would benefit from more docs:
```
```suggestion
     * @param {{seat: UserSeat, offerResult: *}} [out] will be filled
```

turadg: But I find the need for this suspicious. This function is returning payouts to abstract away seat and offerResult… but here we are needing to punch through. Will contract users need to also? If so we should just change the return type to hide less: `{seat, offerResult, payouts}`.

Or would it be sufficient to have `{numWantsSatisfied, payouts}`?

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -305,6 +329,112 @@ test('limit', async t => {
...
+    const expectedAnchor = AmountMath.make(
+      anchor.brand,
+      minusStableFee(giveRun).value,
+    );
+    t.deepEqual(actualAnchor, expectedAnchor);
+  };
+
+  let ix = 0;
+  for (const [kind, give, want, result] of trades) {
+    switch (kind) {
```
turadg: this data driven test calls for [Ava macros](https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#reusing-test-logic-through-macros).  See https://github.com/Agoric/agoric-sdk/pull/5900

packages/zoe/src/contractFacet/zcfSeat.js
```diff
@@ -335,7 +337,7 @@ export const createSeatManager = (
         );
         return amountKeywordRecord;
       },
-      clear: ignoreContext(clear),
+      clear: ({ self }) => clear(self),
```
turadg: just parroting the linter here: gotta remove the `ignoreContext` import

packages/zoe/src/contractFacet/zcfSeat.js
```diff
@@ -114,7 +114,9 @@ export const createSeatManager = (
   };
 
   const clear = zcfSeat => {
-    zcfSeatToStagedAllocations.delete(zcfSeat);
+    if (zcfSeatToStagedAllocations.has(zcfSeat)) {
```
turadg: this is a change to the throw behavior of `clear`. was the earlier behavior documented or tested?

I'll go google it since [our API docs aren't on the APIs](https://github.com/Agoric/agoric-sdk/issues/4291) ;-) https://docs.agoric.com/zoe/api/zoe-contract-facet.html#zcfseat-clear Nope.

packages/inter-protocol/src/psm/psm.js
dckc: Dunno. That comment pre-dates this PR.

turadg: Ok. Since it's written in this contract twice and relevant to other contracts, I'm requesting a ticket to document the footgun and that it be referenced from the places like this that say "someday…"

packages/inter-protocol/test/psm/test-psm.js
dckc: Yes, I was too lazy to change the other call sites. Returning just the seat might be most straightforward.

packages/inter-protocol/test/psm/test-psm.js
dckc: Note that the items have to be used within one test. The whole point is to *not* make a new PSM between them. Do macros work that way?

turadg: usually things that need be shared between tests are configured with `before` and `beforeEach`. but I won't block on this.

packages/zoe/src/contractFacet/zcfSeat.js
Chris-Hibbert: Not only are there no tests for `clear`, there appear to be [no tests for zcfSeat](https://agoric-sdk-coverage.netlify.app/#file/desc/false/true/true/true/zoe%2Fsrc%2FcontractFacet/).  

[not suggesting this PR add anything.]

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -305,6 +329,112 @@ test('limit', async t => {
   // t.throwsAsync(() => await E(seat1).getOfferResult());
 });
 
+/** @type {[kind: 'want' | 'give', give: number, want: number, result: number | false][]} */
+const trades = [
+  ['give', 200, 190, false],
+  ['want', 101, 100, 1],
+  ['give', 50, 50, false],
+  ['give', 51, 50, 1],
```
Chris-Hibbert: Even though I've been working with JS for a  couple years now, seeing `false` and `1` as two possible values of a flag strikes me as wrong.

packages/inter-protocol/src/psm/psm.js
```diff
@@ -160,11 +160,24 @@ export const start = async (zcf, privateArgs, baggage) => {
       AmountMath.isGTE(maxAnchor, wanted),
```
Chris-Hibbert, line 159: Does the TODO on line 158 interact with the current problem at all? Can it be updated?

packages/inter-protocol/src/psm/psm.js
Chris-Hibbert: I think it needs an `onFail:` clause to invoke `burnLosses`

packages/zoe/src/contractFacet/zcfSeat.js
turadg: though, why delegate here at all? we can delete the `const clear` above.
```suggestion
      clear: ({ self }) => {
          if (zcfSeatToStagedAllocations.has(self)) {
            zcfSeatToStagedAllocations.delete(self);
          }
      };
```

packages/inter-protocol/src/psm/psm.js
dtribble: note that only one of the comments is valid (the one before the mintGains). The assert got copied to provide a better diagnostic, but the comment needs to be deleted.

packages/zoe/src/contractFacet/zcfSeat.js
Chris-Hibbert: I think MarkM hinted at the right choice here. Please inline the method, so it'll be more apparent that the parameter handling is parallel to the sibling methods.

Chris-Hibbert (changes requested):
I think there's more discussion going on, so I'm going to submit my current comments and wait to see how it settles before reviewing again.

packages/inter-protocol/test/psm/test-psm.js
dckc: it needs to be shared between these trades but no other tests

turadg: then another file would work well. this `test` function is doing a lot of work to replicate some of the `before` and `macro` semantics.

packages/inter-protocol/src/psm/psm.js
dckc: postponed to #6116

packages/inter-protocol/src/psm/psm.js
dckc: we certainly have relevant tests; I'm getting rid of that TODO in favor of #6116

packages/inter-protocol/src/psm/psm.js
dckc: added #6116 and cited it. I don't understand that issue very well, so I'd appreciate it if someone else would clarify the title / description. @michaelfig helped me a little.

packages/inter-protocol/test/psm/test-psm.js
dckc: I refactored the driver to have additional methods that expose the seat.

packages/inter-protocol/test/psm/test-psm.js
dckc: Let's punt on this for today.

turadg (approved):
LGTM given https://github.com/Agoric/agoric-sdk/issues/6116 (which I've added to PSM audit release)

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -179,6 +183,42 @@ async function makePsmDriver(t, customTerms) {
...
+  /**
+   * @param {Amount<'nat'>} giveAnchor
+   * @param {Amount<'nat'>} [wantStable]
+   */
+  const swapAnchorForStableSeat = async (giveAnchor, wantStable) => {
+    const seat = E(zoe).offer(
+      E(publicFacet).makeWantStableInvitation(),
+      harden({
+        give: { In: giveAnchor },
+        ...(wantStable ? { want: { Out: wantStable } } : {}),
```
Chris-Hibbert, line 195: I like this.  It's a definite improvement on my approach of making the proposal, and then adding the want later.

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -305,6 +340,108 @@ test('limit', async t => {
   // t.throwsAsync(() => await E(seat1).getOfferResult());
 });
 
+/** @type {[kind: 'want' | 'give', give: number, want: number, ok: boolean, wants?: number][]} */
+const trades = [
+  ['give', 200, 190, false],
+  ['want', 101, 100, true, 1],
+  ['give', 50, 50, false],
+  ['give', 51, 50, true, 1],
```
Chris-Hibbert, line 348: I get what this is doing now, but only because of side conversations about ava macros. It would be nice if it said that the first line was intended to fail because the PSM doesn't have any anchor until after a `want` completes. Does the third fail because it doesn't include fees?

Chris-Hibbert (approved):

packages/inter-protocol/src/psm/psm.js
erights: Progress on #6116 at #6577 , where `stageMgr.commit()` replaces `zcf.reallocate(...)` and does guarantee that all staging are cleared, whether the commit succeeded or failed.

@dtribble 's suggested `stageMgr.transfer` operation is much like @turadg 's rows of `atomicReallocate` above. The contrast between the list and the reified stageMgr is approx the contrast between https://github.com/Agoric/agoric-sdk/pull/6577 and https://github.com/Agoric/agoric-sdk/pull/4327

PR #6103, opened 2022-09-01
refactor(casting): Use stream cells and release I/O handles

kriskowal:
refs: #5460

## Description

This change refactors casting such that followers can efficiently traverse stream cells while preserving legacy value cells, and ensures that follower iterators do not retain I/O handles from yield to next, such that a program publishing a bundle can gracefully exit.

This should not surface to documentation.

### Testing Considerations

I’ve increased test coverage and used `agoric follow` extensively to develop an intuition for whether both latest and each iterators behave correctly, and refactored the fake RPC server to simulate a wider variety of scenarios, but these tests are not sufficient to cover all of the interesting scenarios for both each and latest iterations. The tests are now however sufficient to distinguish each and latest.


packages/casting/src/follower-cosmjs.js
```diff
@@ -102,6 +123,18 @@ export const makeCosmjsFollower = (
...
   } = options;
 
+  /**
+   * @param {any} err
+   */
+  const crash = err => {
+    if (crasher) {
+      E(crasher)
+        .crash(`PROOF VERIFICATION FAILURE; crashing follower`, err)
+        .catch(e => assert(false, X`crashing follower failed: ${e}`));
```
gibson042: I suspect this issue to be preexisting, but the `assert` here will result in an unhandled rejection. To be honest, I'm not sure about the intended behavior of `crash`, and it seems like there are many similar cases and quite possibly also race conditions.

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+    }
+    return clientP;
+  };
+
+  const getBlockHeight = async () => {
+    const values = await E(leader).mapEndpoints(where, async endpoint => {
+      const client = await provideTendermintClient(endpoint);
+      const info = await client.abciInfo();
+      const { lastBlockHeight } = info;
+      assert.typeof(lastBlockHeight, 'number');
```
gibson042, line 137: Oof, that stinks. It probably should be string or bigint... is there an upstream issue?

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+  };
+
   /** @type {Map<string, import('@cosmjs/stargate').QueryClient>} */
   const endpointToQueryClient = new Map();
 
   /**
    * @param {string} endpoint
    */
-  const getOrCreateQueryClient = async endpoint => {
+  const provideQueryClient = async endpoint => {
```
gibson042, line 149: 👍

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
   const leader =
     typeof leaderOrMaker === 'function' ? leaderOrMaker() : leaderOrMaker;
 
+  const tendermintClientPs = new Map();
+  /**
+   * @param {string} endpoint
```
gibson042, line 121: Can the return be typed? My TS-foo isn't great, but I think it would be something like this (modulo typedefs):
```suggestion
   * @param {string} endpoint
   * @returns {Promise<{abciInfo: () => {lastBlockHeight: number}, abciQuery: ({path?: string, data?: string, height?: number, prove?: boolean}) => {code: any, value: any, log: any}}>}
```

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+      return result;
+    }
+
+    // Handle the data prefix if any.
+    assert(
+      result.length >= dataPrefixBytes.length,
+      X`result too short for data prefix ${dataPrefixBytes}`,
+    );
+    assert(
+      dataPrefixBytes.every((v, i) => v === result[i]),
```
gibson042:
```suggestion
      arrayEqual(result.subarray(0, dataPrefixBytes.length), dataPrefixBytes),
```
(we might as well use the helper consistently)

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+      X`${result} doesn't start with data prefix ${dataPrefixBytes}`,
+    );
+    return result.slice(dataPrefixBytes.length);
+  };
+
+  /**
+   * @param {number} height
+   */
+  const getUnprovenDataAtHeight = async height => {
```
gibson042, line 234: I'm seeing lots of similarity here (and already some divergence); should it be consolidated?
```js
const makeChainDataGetter = callback => {
  const getter = async height => {
    const {
      storeName,
      storeSubkey,
      dataPrefixBytes = defaultDataPrefixBytes,
    } = await castingSpecP;
    assert.typeof(storeName, 'string');
    assert(storeSubkey);
    const spec = { storeName, storeSubkey, dataPrefixBytes };

    const values = await E(leader).mapEndpoints(
      where,
      endpoint => callback(endpoint, spec, height),
    );
    const { value, error } = collectSingleValue(values);
    if (error !== null) {
      throw error;
    }
    assert(value);

    if (value.length === 0) {
      // No data.
      return value;
    }

    // Verify and strip off the data prefix.
    assert(
      arrayEqual(value.subarray(0, dataPrefixBytes.length), dataPrefixBytes),
      X`${value} doesn't start with data prefix ${dataPrefixBytes}`,
    );
    return value.slice(dataPrefixBytes.length);
  };
  return getter;
};

const getProvenDataAtHeight = makeChainDataGetter(async (endpoint, spec, height) => {
  const { storeName, storeSubkey, dataPrefixBytes } = spec;
  const queryClient = await provideQueryClient(endpoint);
  return E(queryClient)
    .queryVerified(storeName, storeSubkey, height)
    .then(value => ({ value, error: null }), error => ({ value: null, error }));
});

const getUnprovenDataAtHeight = makeChainDataGetter(async (endpoint, spec, height) => {
  const { storeName, storeSubkey, dataPrefixBytes } = spec;
  const client = await provideTendermintClient(endpoint);
  const response = await client.abciQuery({
    path: `store/${storeName}/key`,
    data: storeSubkey,
    height,
    prove: false,
  });
  if (response.code !== 0) {
    return {
      value: null,
      error: new Error(`Tendermint ABCI query failed: ${response.log}`),
    };
  }
  const { value } = response;
  return { value, error: null };
});

File: packages/casting/src/follower-cosmjs.js [hunk 7]
```diff
@@ -44,7 +67,7 @@ const collectSingle = values => {
 
 /**
  * @callback QueryVerifier
- * @param {() => Promise<Uint8Array>} getProvenValue
+ * @param {() => Promise<Uint8Array>} getProvenData
  * @param {(reason?: unknown) => void} crash
  * @param {() => Promise<Uint8Array>} getAllegedValue
```
gibson042: Why introduce different nouns for `getProvenData` vs. `getAllegedValue`?

packages/casting/src/follower-cosmjs.js
```diff
@@ -102,6 +123,18 @@ export const makeCosmjsFollower = (
...
+      E(crasher)
+        .crash(`PROOF VERIFICATION FAILURE; crashing follower`, err)
+        .catch(e => assert(false, X`crashing follower failed: ${e}`));
+    }
+    throw err;
+  };
+
   /** @type {QueryVerifier} */
   const queryVerifier = proofToQueryVerifier[proof];
   assert(queryVerifier, X`unrecognized follower proof mode ${proof}`);
```
gibson042: This is a bit sloppy (e.g., it will fail if proof names a property of `Object.prototype` such as "toString"), but also the indirection seems like more of a hindrance than a benefit. I think it would be better to merge this logic (and `proofToQueryVerifier` itself) into the assignment of `tryGetDataAtHeight` later in the function.
```suggestion
  let tryGetDataAtHeight;
  switch (proof) {
    case 'strict': {
      // Crash hard if we can't get a proven value.
      tryGetDataAtHeight = height => getProvenDataAtHeight(height).catch(crash);
    }
    case 'none': {
      // Trust an unproven value.
      tryGetDataAtHeight = getUnprovenDataAtHeight;
    }
    case 'optimistic': {
      tryGetDataAtHeight = height => {
        const allegedValue = await getUnprovenDataAtHeight(height);
        // Prove later, since it may take time we say we can't afford.
        getProvenDataAtHeight(height).then(provenValue => {
          if (arrayEqual(provenValue, allegedValue)) {
            return;
          }
          const msg = X`Alleged value ${allegedValue} did not match proof ${provenValue}`;
          const err = assert.error(msg);
          crash(err);
        });
        // Speculate that we got the right value.
        return allegedValue;
      });
    }
    default: {
      assert(false, X`unrecognized follower proof mode ${proof}`);
    }
  }
```

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+   * @param {number} blockHeight
+   * @param {Uint8Array} data
+   */
+  const streamCellForData = (blockHeight, data) => {
+    const cell = decode(data);
+
+    if (
+      cell &&
+      Array.isArray(cell.values) &&
+      typeof cell.blockHeight === 'string'
```
gibson042:
```suggestion
      Array.isArray(cell.values) &&
      typeof cell.blockHeight === 'string' &&
      /^0$|^[1-9][0-9]*$/.test(cell.blockHeight)
```
Also, let's make an `isStreamCell` helper to avoid duplication and divergence.

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+    ) {
+      blockHeight = Number(cell.blockHeight);
+      if (!Number.isNaN(blockHeight)) {
+        return {
+          blockHeight: Number(cell.blockHeight),
+          values: cell.values,
+        };
+      }
```
gibson042:
```suggestion
      return {
        blockHeight: Number(cell.blockHeight),
        values: cell.values,
      };
```

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+   * @returns {Promise<FollowerElement<T>>}
+   */
+  const followerElementFromStreamCellValue = async (
+    data,
+    blockHeight,
+    currentBlockHeight,
+  ) => {
+    if (typeof data === 'string') {
+      data = JSON.parse(data);
+    }
```
gibson042: Under what circumstances will `data` not be a string?
```suggestion
    assert.typeof(data, 'string');
    data = JSON.parse(data);
```

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+   * @yields {FollowerElement<T>}
+   */
+  async function* getLatestIterable() {
+    let blockHeight;
+    let data;
+    for (;;) {
+      blockHeight = await getBlockHeight();
+      const dataAndStreamCell = await getStreamCellAtHeight(blockHeight);
+      if (dataAndStreamCell) {
+        data = dataAndStreamCell.data;
```
gibson042: Does `getStreamCellAtHeight` really carry its own weight vs. direct `getDataAtHeight` and `streamCellForData`?

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
-          .catch(e => assert(false, X`crashing follower failed: ${e}`));
-      } else {
-        console.error(`PROOF VERIFICATION FAILURE; crashing follower`, err);
+  /**
+   * @yields {FollowerElement<T>}
+   */
+  async function* getLatestIterable() {
+    let blockHeight;
+    let data;
+    for (;;) {
```
gibson042, line 394:
```suggestion

    // Get an initial value.
    for (;;) {
```

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
-              .retry(where, err, attempt)
-              .then(() => (retrying = null));
-          }
-          await retrying;
-          await E(leader).jitter(where);
-
-          // eslint-disable-next-line no-use-before-define
-          tryQueryAndUpdate(castingChange).then(success, retryOrFail);
-        };
+    for (;;) {
```
gibson042:
```suggestion
    // Get subsequent values.
    for (;;) {
```

But also, it seems like the two loops can be combined.
```js
    let blockHeight;
    let data;
    for (;;) {
      const currentBlockHeight = await getBlockHeight();
      if (currentBlockHeight === blockHeight) {
        await E(leader).jitter(where);
        continue;
      }

      const dataAndStreamCell = await getStreamCellAtHeight(currentBlockHeight);
      if (!dataAndStreamCell) {
        await E(leader).jitter(where);
        continue;
      }

      const { data: currentData, streamCell: currentStreamCell } =
        dataAndStreamCell;

      blockHeight = currentBlockHeight;

      // Ignore adjacent duplicates.
      // This can only occur for legacy cells.
      // It is possible that the data changed from and back to the last
      // sampled data, but ignoring intermediate changes is consistent with
      // the semantics of getLatestIterable.
      if (data !== undefined && arrayEqual(data, currentData)) {
        continue;
      }
      // However, streamCells that vacillate will reemit, since each iteration
      // at a unique block height is considered distinct.

      yield* lastValueFromCell(currentStreamCell, currentBlockHeight);
      data = currentData;
    }
```

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+      while (rightBlockHeight > cursorBlockHeight) {
+        // Stream cells should have a numeric block height that is between
+        // 1 (the first valid block height) and the height of the block
+        // they were published in, but we clamp the figure into the valid
+        // range just in case.
+        const leftBlockHeight = clamp(
+          1,
+          rightBlockHeight - 1,
+          rightStreamCell.blockHeight - 1,
         );
```
gibson042: I'd rather see assertions than issue-hiding clamp behavior.

gibson042 (dismissed):

packages/casting/src/follower-cosmjs.js
```diff
@@ -112,290 +145,442 @@ export const makeCosmjsFollower = (
...
+          cells.push(rightStreamCell);
+        }
+        // Prepare for next iteration by moving left to right.
+        rightData = leftData;
+        rightStreamCell = leftStreamCell;
+        rightBlockHeight = leftBlockHeight;
       }
-    };
-
-    const changeFollower = E(leader).watchCasting(castingSpecP);
```
michaelfig, line 365: FFR, we will need to use something like the `changeFollower` to ensure our RPC servers don't collapse under load.  The `jitter` mechanism is intended just to provide a little bit of randomness, not to wait for the next block to appear.

michaelfig (commented):
I'd echo @gibson042's comments, and just one note for future reference about the changeFollower.

michaelfig (commented):
Also, let's discuss exposing `getReverseEachIterable` and how to structure `getEachIterableForHeight` in terms of that.

packages/casting/src/follower-cosmjs.js
kriskowal: I chose not to change this since. I suspect that Chesterton put this fence (@mfig in this metaphor 😊) here in case the appropriate mitigation is to hard-kill the container (à la `process.abort`) instead of merely throwing. I made sure that the default is to throw. I can create an issue if the dangling promises are unintended.

packages/casting/src/follower-cosmjs.js
kriskowal: Filed https://github.com/cosmos/cosmjs/issues/1263

packages/casting/src/follower-cosmjs.js
kriskowal: It can be, but in this case, the return type is safely inferred. I’m inclined to keep internal function return types implicit. They do get checked on the caller side.

packages/casting/src/follower-cosmjs.js
kriskowal: I missed this one. Thank you!

packages/casting/src/follower-cosmjs.js
kriskowal: I think rather than going higher order (as it was previously), I’ll write a utility that deduplicates their common tail.

packages/casting/src/follower-cosmjs.js
kriskowal: In a follow-up, I will change getAllegedValue to getAllegedData. Both of these functions ended with Value previously. The introduction of stream cells made the name “value” ambiguous in this module, so I changed the layering of names:

- data (previously value): bytes from the query
- cell: either a value cell or a stream cell
- value: one or more of these in a cell

packages/casting/src/follower-cosmjs.js
kriskowal: I’ve created an issue to track this https://github.com/Agoric/agoric-sdk/issues/6154

I assume we need to use a Tendermint API that long-polls for block height changes or more specifically for changes to a specific vstorage cell. If you’ve got hints, please post them there.

packages/casting/src/follower-cosmjs.js
kriskowal: No, not any more. It was a better idea before I realized we needed to do data length checks. I’m inlining this function and it’s better.

packages/casting/src/follower-cosmjs.js
kriskowal: Seems to be an artifact of the fakes for testing. I’ll drill and find the bug in the test scaffold.

packages/casting/src/follower-cosmjs.js
kriskowal: I agree. I’m inlining the proof modes.

kriskowal:
@gibson042 Please take a look at the two new commits. I’ll continue thinking about reverse iterator refactor.

packages/casting/src/follower-cosmjs.js
kriskowal: Response from CosmJS maintainer: we weighed the pros and cons, decided to normalize the API in terms of numbers. The underlying RPC interfaces were not carefully designed, so there are inconsistencies that Tendermint and Stargate attempt to paper over.

packages/casting/src/follower-cosmjs.js
kriskowal: The fake RPC server did not JSON.stringify legacy value cells. That may have been by design, so I’m going to dig deeper. I might have to run an old chain with the new casting to verify.

packages/casting/test/fake-rpc-server.js
```diff
@@ -124,24 +131,48 @@ export const startFakeServer = (t, fakeValues, options = {}) => {
...
+            const serializedValues = values.map(val =>
+              JSON.stringify(marshaller.serialize(val)),
+            );
+            responseValue = {
+              blockHeight: String(desiredHeight - 1),
+              values: serializedValues,
+            };
+          } else {
+            // Return a single naked value.
+            responseValue = JSON.stringify(marshaller.serialize(values[0]));
```
kriskowal: I added JSON.stringify on this line and am suspicious. This is what made possible to unconditionally parse the JSON in `followerElementFromStreamCellValue`, but I suspect old value cells did _not_ double envelope.

I’ve made an attempt to use the new casting with an old chain, with old storage cells, and have failed to exercise scenario2 in a way that produces anything I can consume. So, I’ve got a blind spot here.

Perhaps @michaelfig can weigh in.

kriskowal: I’ve pushed a follow-up that resolves my concern:

I’ve reduced the domain of `decode` from bytes to strings. We will only handle UTF-8 text cells at this time. The path forward for non-text cells is a binary stream cell encoding that distinguishes itself with a prefix that is not valid UTF-8, e.g., any control character, as a tag for the binary union.

The stream cell detector no longer uses `decode`, and explicitly uses `JSON.parse`. Then, we use `decode` for all values, regardless of whether they come from a value cell or a stream cell. This will cause a redundant parse for legacy value cells, but legacy is spelled out L E G A C Y for a reason.

I restored the shape of the test scaffold to ensure we’re properly testing legacy value cells.

kriskowal:
@michaelfig The tests that add a `stride` parameter verify that getEachIterable visits every value, even if the chain advances multiple blocks between iterations. Hope that helps!

michaelfig (approved):
LGTM.

It would have been nice to preserve the `decode` semantics, but I understand the rationale for not wanting to change the stream cell value (string) implementation.

PR #6105, opened 2022-09-01
feat(agoric-cli): follow options for lossy and block height prefixes

kriskowal:
refs: #5460

## Description

Following #6103, `agoric follow` became useful for ad hoc tests of each and latest casting iterators during `agoric deploy` commands. This change adds `--lossy`, `--block-height`, and `--current-block-height` options. `--lossy` suggests use of getLatestIterator instead of getEachIterator, `--block-height` adds a prefix to each log with the self-reported height of the block that published the value, and `--current-block-height` adds a prefix to each log with the current block height at the time the log was emitted.

These flags as described above are mentioned in command-line usage. A document covering uses of the `agoric` command might also mention them, or how-to guides might make use of them in the context of solving a problem or demonstrating how storage cells work. For example, these two flags block height are useful for confirming that stream cells report the height of the block that they’re published into, not the height of the chain at the time they were published.

### Testing Considerations

Manual. I would not say that the behavior is adequately tested in `casting`, but that `casting` is the right package to test the behavior revealed by the CLI.

### DIFF
--- a/packages/agoric-cli/src/follow.js
+++ b/packages/agoric-cli/src/follow.js
@@ -6,6 +6,7 @@ import { decodeToJustin } from '@endo/marshal/src/marshal-justin.js';
 import {
   iterateLatest,
   makeCastingSpec,
+  iterateEach,
   makeFollower,
   makeLeader,
 } from '@agoric/casting';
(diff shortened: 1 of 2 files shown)


michaelfig (approved):
LGTM!

PR #6106, opened 2022-09-01
fix(vat-data)!: deprecate kinds in favor of Far Classes

erights:
I believe we've now had enough experience with FarClasses, etc, to deprecate the virtual/durable kind APIs and unconditionally recommend FarClasses, etc, instead.

### DIFF
--- a/packages/vat-data/src/kind-utils.js
+++ b/packages/vat-data/src/kind-utils.js
@@ -29,7 +29,10 @@ export const provideKindHandle = (baggage, kindName) =>
   provide(baggage, `${kindName}_kindHandle`, () => makeKindHandle(kindName));
 harden(provideKindHandle);
 
-/** @type {import('./types.js').VivifyKind} */
+/**
+ * @deprecated Use vivifyFarClass instead
+ * @type {import('./types.js').VivifyKind}
+ */
 export const vivifyKind = (
   baggage,
   kindName,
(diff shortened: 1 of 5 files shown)


turadg:
@erights you marked this Ready for Review but the PR description says, "We won't know if this is a good idea until…". Do you think it's a good idea now?

erights:
> @erights you marked this Ready for Review but the PR description says, "We won't know if this is a good idea until…". Do you think it's a good idea now?

Good question! You and @Chris-Hibbert have given these more road testing at this point than I have. What do you both think? Are we ready to deprecate kinds in favor of far classes / far instances?

Chris-Hibbert:
I don't know enough yet to think we should definitely start on a project to convert everything over, but I prefer Far classes at this point. I'm not sure whether it would be more prudent to finish making Zoe durable using Kinds, rather than mixing the two styles.

erights:
Reviewers (as well as @turadg @michaelfig ), I'd appreciate help in how to deprecate these so that the deprecations show up more reliably under vscode. Thanks.

turadg:
> help in how to deprecate these so that the deprecations show up more reliably under vscode

The `export const` is implicitly defining a new types. I don't see a way to to propagate the types but we can mark the new one as deprecated.
```

/** @deprecated */
 export const {
   defineKind,
   defineKindMulti,
   defineDurableKind,
   defineDurableKindMulti,
} = VatDataGlobal;

export const {
   makeKindHandle,
   providePromiseWatcher,
```

erights:
Re https://github.com/Agoric/agoric-sdk/pull/6106#issuecomment-1305802805 , Done.

turadg (approved):

PR #6107, opened 2022-09-01
fix: heap far classes

erights:
Prior to this PR, our menagerie of far class definer apis is

(FarClass, FarClassKit, FarInstance) x (Virtual, Durable, vivify)

The hole this leaves is the normal case of objects living in the programming language's heap. This was covered by `Far` until we introduced defensive interface guards. So this PR adds a fourth setting to the second dimension above, making it

(FarClass, FarClassKit, FarInstance) x (Heap, Virtual, Durable, vivify)

In practice I expect almost all uses to be of the Heap and vivify variants. In the Kind system, the unnamed case was the Vrtual case, when we didn't yet understand what our dominant pattern of use will be. An open question: Should we drop the "Heap" from the heap far class definers, making that the unnamed case. If so, then common cases will be
   * defineFarClass
   * defineFarClassKit
   * makeFarInstance
   * vivifyFarClass
   * vivifyFarClassKit
   * vivifyFarInstance

If https://github.com/Agoric/agoric-sdk/pull/6118 passes, then it would be
   * defineExoClass
   * defineExoClassKit
   * makeExo
   * vivifyExoClass
   * vivifyExoClassKit
   * vivifyExo

### DIFF
--- a/packages/store/src/patterns/interface-tools.js
+++ b/packages/store/src/patterns/interface-tools.js
@@ -377,7 +377,7 @@ harden(defineHeapFarClassKit);
  * @param {object} [options]
  * @returns {T & RemotableBrand<{}, T>}
  */
-export const defineHeapFarInstance = (
+export const makeHeapFarInstance = (
   tag,
   interfaceGuard,
   methods,
@@ -392,4 +392,4 @@ export const defineHeapFarInstance = (
   );
   return makeInstance();
 };
-harden(defineHeapFarInstance);
+harden(makeHeapFarInstance);
--- a/packages/store/src/patterns/interface-tools.js
+++ b/packages/store/src/patterns/interface-tools.js
@@ -1,12 +1,12 @@
 import { Far } from '@endo/marshal';
 import { E } from '@endo/eventual-send';
-import { listDifference } from '@agoric/internal';
+import { listDifference, objectMap } from '@agoric/internal';
 
 import { fit } from './patternMatchers.js';
 
 const { details: X, quote: q } = assert;
 const { apply, ownKeys } = Reflect;
-const { defineProperties } = Object;
+const { defineProperties, seal, freeze } = Object;
 
 const defendSyncArgs = (args, methodGuard, label) => {
   const { argGuards, optionalArgGuards, restArgGuard } = methodGuard;
@@ -251,3 +251,145 @@ const emptyRecord = harden({});
  * @returns {{}}
  */
 export const initEmpty = () => emptyRecord;
+
+/**
+ * @template [S = any]
+ * @template [T = any]
+ * @typedef {object} Context
+ * @property {S} state
+ * @property {T} self
+ */
+
+/**
+ * @template A,S,T
+ * @param {string} tag
+ * @param {any} interfaceGuard
+ * @param {(...args: A[]) => S} init
+ * @param {T} methods
+ * @param {object} [options]
+ * @returns {(...args: A[]) => (T & RemotableBrand<{}, T>)}
+ */
+export const defineHeapFarClass = (
+  tag,
+  interfaceGuard,
+  init,
+  methods,
+  options = undefined,
+) => {
+  /** @type {WeakMap<T,Context<S,T>} */
+  const contextMap = new WeakMap();
+  const prototype = defendPrototype(
+    tag,
+    contextMap,
+    methods,
+    true,
+    interfaceGuard,
+  );
+  const makeInstance = (...args) => {
+    // Be careful not to freeze the state record
+    const state = seal(init(...args));
+    const self = harden({ __proto__: prototype });
+    // Be careful not to freeze the state record
+    /** @type {Context<S,T>} */
+    const context = freeze({ state, self });
+    contextMap.set(self, context);
+    if (options) {
+      const { finish = undefined } = options;
+      if (finish) {
+        finish(context);
+      }
+    }
+    return self;
+  };
+  return harden(makeInstance);
+};
+harden(defineHeapFarClass);
+
+/**
+ * @template A,S,F
+ * @param {string} tag
+ * @param {any} interfaceGuardKit
+ * @param {(...args: A[]) => S} init
+ * @param {F} methodsKit
+ * @param {object} [options]
+ * @returns {(...args: A[]) => F}
+ */
+export const defineHeapFarClassKit = (
+  tag,
+  interfaceGuardKit,
+  init,
+  methodsKit,
+  options = undefined,
+) => {
+  const facetNames = ownKeys(methodsKit);
+  const interfaceNames = ownKeys(interfaceGuardKit);
+  const extraInterfaceNames = listDifference(facetNames, interfaceNames);
+  assert(
+    extraInterfaceNames.length === 0,
+    X`Interfaces ${q(extraInterfaceNames)} not implemented by ${q(tag)}`,
+  );
+  const extraFacetNames = listDifference(interfaceNames, facetNames);
+  assert(
+    extraFacetNames.length === 0,
+    X`Facets ${q(extraFacetNames)} of ${q(tag)} not guarded by interfaces`,
+  );
+
+  const contextMapKit = objectMap(methodsKit, () => new WeakMap());
+  const prototypeKit = objectMap(methodsKit, (methods, facetName) =>
+    defendPrototype(
+      `${tag} ${facetName}`,
+      contextMapKit[facetName],
+      methods,
+      true,
+      interfaceGuardKit[facetName],
+    ),
+  );
+  const makeInstanceKit = (...args) => {
+    // Be careful not to freeze the state record
+    const state = seal(init(...args));
+    // Don't freeze context until we add facets
+    const context = { state };
+    const facets = objectMap(prototypeKit, (prototype, facetName) => {
+      const self = harden({ __proto__: prototype });
+      contextMapKit[facetName].set(self, context);
+      return self;
+    });
+    context.facets = facets;
+    // Be careful not to freeze the state record
+    freeze(context);
+    if (options) {
+      const { finish = undefined } = options;
+      if (finish) {
+        finish(context);
+      }
+    }
+    return facets;
+  };
+  return harden(makeInstanceKit);
+};
+harden(defineHeapFarClassKit);
+
+/**
+ * @template T,M
+ * @param {string} tag
+ * @param {InterfaceGuard|undefined} interfaceGuard
+ * @param {M} methods
+ * @param {object} [options]
+ * @returns {T & RemotableBrand<{}, T>}
+ */
+export const defineHeapFarInstance = (
+  tag,
+  interfaceGuard,
+  methods,
+  options = undefined,
+) => {
+  const makeInstance = defineHeapFarClass(
+    tag,
+    interfaceGuard,
+    initEmpty,
+    methods,
+    options,
+  );
+  return makeInstance();
+};
+harden(defineHeapFarInstance);
--- a/packages/store/test/test-heap-classes.js
+++ b/packages/store/test/test-heap-classes.js
@@ -1,0 +1,107 @@
+// @ts-check
+
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
+import {
+  defineHeapFarClass,
+  defineHeapFarClassKit,
+  defineHeapFarInstance,
+} from '../src/patterns/interface-tools.js';
+import { M } from '../src/patterns/patternMatchers.js';
+
+const UpCounterI = M.interface('UpCounter', {
+  incr: M.call()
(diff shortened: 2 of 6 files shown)


packages/store/src/patterns/interface-tools.js
```diff
@@ -251,3 +327,182 @@ const emptyRecord = harden({});
...
+ * @param tag - a string that will be used to identify the class in error messages
+ * @param interfaceGuard - a predicate that takes an object and returns true if it is an instance of
+ * the class.
+ * @param init - a function that takes the arguments passed to the constructor and returns the initial
+ * state record.
+ * @param methods - a record of methods that will be added to the prototype of the class.
+ * @param [options] - an optional object with a finish method that is called after the instance is
+ * created.
+ * @returns A function that takes a list of arguments and returns an object.
+ */
```
Chris-Hibbert: Does it work to declare the params and returns twice?   I'd expect it to work better if the comments from the second iteration were just merged into the first iteration, which has the types.
```suggestion
/**
 * It takes a bunch of
 * arguments and returns a function that creates an object with a prototype
 * that has methods that can access the object's state
 * @template A,S,T
 * @param {string} tag - a string that will be used to identify the class in error messages
 * @param {any} interfaceGuard - a predicate that takes an object and returns true if it
 *    is an instance of the class.
 * @param {(...args: A[]) => S} init - a function that takes the arguments passed to the
 *     constructor and returns the initial state record.
 * @param {T} methods - a record of methods that will be added to the prototype of the class.
 * @param {object} [options] - an optional object with a finish method that is called after
 *    the instance is created.
 * @returns {(...args: A[]) => (T & RemotableBrand<{}, T>)} A function that takes a list of
 *   arguments and returns an object.
 */
```

packages/vat-data/test/test-durable-classes.js
```diff
@@ -0,0 +1,110 @@
...
+      'In "incr" method of (UpCounter) arg 0: string "foo" - Must be a number',
+  });
+});
+
+/* Defining a class that has two interfaces, one for up and one for down. */
+test('test defineDurableFarClassKit', t => {
+  const counterKindHandle = makeKindHandle('Counter');
+
+  /* Defining a function that will create an object that has the methods defined in the UpCounterI
+interface. */
```
Chris-Hibbert:
```suggestion
  /* Defining a function that will create an object that has the methods defined
     in the UpCounterI interface. */
```

packages/vat-data/test/test-vivify.js
```diff
@@ -0,0 +1,129 @@
...
+  /* Testing that the function throws an error when the argument is less than 0. */
+  t.throws(() => upCounter.incr(-3), {
+    message: 'In "incr" method of (upCounter) arg 0: -3 - Must be >= 0',
+  });
+  /* Testing that the function throws an error when the argument is not a number. */
+  t.throws(() => upCounter.incr('foo'), {
+    message:
+      'In "incr" method of (upCounter) arg 0: string "foo" - Must be a number',
+  });
+});
```
Chris-Hibbert, line 119:
```suggestion
});

test.todo('Demonstrate that the baggage can be used to revive objects.');
```

packages/vat-data/test/test-durable-classes.js
```diff
@@ -0,0 +1,110 @@
...
+  // @ts-expect-error the type violation is what we're testing
+  t.throws(() => downCounter.decr('foo'), {
+    message:
+      'In "decr" method of (Counter down) arg 0: string "foo" - Must be a number',
+  });
+  /* Testing that the upCounter object does not have a decr method. */
+  t.throws(() => upCounter.decr(3), {
+    message: 'upCounter.decr is not a function',
+  });
+});
```
Chris-Hibbert, line 95:
```suggestion
});

test.todo('Add tests for durability');
```

Chris-Hibbert (commented):
This looks good, but I didn't do a thorough review of the implementation in `interface-tools.js`.  Please get someone else to review that.

Chris-Hibbert:
please add this to `store/src/index.js`
```
export {
  defendPrototype,
  initEmpty,
  defineHeapFarClass,
  defineHeapFarClassKit,
} from './patterns/interface-tools.js';
```

packages/store/src/patterns/interface-tools.js
cboydstun: sure looks better with just the one comment - I'm just reluctant to delete any of the team's hard work! =)

erights:
Hi @cboydstun please stop making commits to this branch until we talk. In the meaning I'm going to split this PR into one with and one without your commits.

See slack for a zoom link for us to discuss.

packages/store/src/patterns/interface-tools.js
erights: Hi @Chris-Hibbert and @cboydstun , I moved @cboydstun 's changes to https://github.com/Agoric/agoric-sdk/pull/6123 , which I'll treat as a source of suggestions of changes to bring back here. In the meantime, for comments here on those changes, it isn't clear what still applies to this PR without them, so I'll leave those conversations resolved. If you see a comment that does apply to the new state of this PR, please add it back as a fresh comment.

packages/store/src/patterns/interface-tools.js
```diff
@@ -251,3 +251,145 @@ const emptyRecord = harden({});
...
+    true,
+    interfaceGuard,
+  );
+  const makeInstance = (...args) => {
+    // Be careful not to freeze the state record
+    const state = seal(init(...args));
+    const self = harden({ __proto__: prototype });
+    // Be careful not to freeze the state record
+    /** @type {Context<S,T>} */
+    const context = freeze({ state, self });
```
dtribble, line 294: Doesn't this freeze the `state`?

packages/store/test/test-heap-classes.js
```diff
@@ -0,0 +1,107 @@
...
+});
+
+test('test defineHeapFarClass', t => {
+  const makeUpCounter = defineHeapFarClass(
+    'UpCounter',
+    UpCounterI,
+    (x = 0) => ({ x }),
+    {
+      incr(y = 1) {
+        // @ts-expect-error TS doesn't know that `this` is a `Context`
```
dtribble, line 32: Will we be able to tell the type system?  That error is going to be a recurring pain point.

packages/store/test/test-heap-classes.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+  );
+  const { up: upCounter, down: downCounter } = makeCounterKit(3);
+  t.is(upCounter.incr(5), 8);
+  t.is(downCounter.decr(), 7);
+  t.is(upCounter.incr(3), 10);
+  t.throws(() => upCounter.incr(-3), {
+    message: 'In "incr" method of (Counter up) arg 0: -3 - Must be >= 0',
+  });
+  // @ts-expect-error the type violation is what we're testing
```
dtribble, line 80: I like this `expect-error` a lot :)

packages/store/test/test-heap-classes.js
```diff
@@ -0,0 +1,107 @@
...
+  t.throws(() => downCounter.decr('foo'), {
+    message:
+      'In "decr" method of (Counter down) arg 0: string "foo" - Must be a number',
+  });
+  t.throws(() => upCounter.decr(3), {
```
dtribble, line 85: Since you didn't suppress a type warning, I take it signature types don't propagate through?

packages/store/test/test-heap-classes.js
```diff
@@ -0,0 +1,107 @@
...
+  let x = 3;
+  const upCounter = makeHeapFarInstance('upCounter', UpCounterI, {
+    incr(y = 1) {
+      x += y;
+      return x;
+    },
+  });
+  t.is(upCounter.incr(5), 8);
+  t.is(upCounter.incr(1), 9);
+  t.throws(() => upCounter.incr(-3), {
```
dtribble, line 100: since these are essentially the same test cases as above, it would be great to capture them with macros.

dtribble: someday :)

packages/vat-data/test/test-virtual-classes.js
```diff
@@ -0,0 +1,90 @@
...
+      },
+    },
+  );
+  const upCounter = makeUpCounter(3);
+  t.is(upCounter.incr(5), 8);
+  t.is(upCounter.incr(1), 9);
+  t.throws(() => upCounter.incr(-3), {
+    message: 'In "incr" method of (UpCounter) arg 0: -3 - Must be >= 0',
+  });
+  // @ts-expect-error TS doesn't know that `this` is a `Context`
```
dtribble: There's no `this` in the next line so this is the wrong error message?

packages/vat-data/test/test-vivify.js
```diff
@@ -0,0 +1,121 @@
...
+
+test('test vivifyFarClassKit', t => {
+  const baggage = makeScalarBigMapStore('baggage', { durable: true });
+
+  const makeCounterKit = vivifyFarClassKit(
+    baggage,
+    'Counter',
+    harden({ up: UpCounterI, down: DownCounterI }),
```
dtribble: This is a case where we really should not require `harden`. We are calling a library that is essentially a part of the computational model, and the `harden` obscures the most important thing here.

packages/store/src/patterns/interface-tools.js
dtribble, line 294: Perhaps note that `freeze` is not recursive so the actual `state` record is not frozen, since we get so used to recursive freeze :)

dtribble (approved):
LGTM.  A few non-blocking comments

packages/store/src/patterns/interface-tools.js
erights: It does not. `harden` would because it applies `freeze` transitively. `freeze` itself is one level, which is why we use it here --- in order to avoid freezing the state.

packages/store/test/test-heap-classes.js
erights: I have no idea, and I agree. I did try but was helpless in the face of TypeScript's rules. 

@turadg @mhofman @michaelfig ?

packages/store/test/test-heap-classes.js
erights: ;)

packages/store/test/test-heap-classes.js
erights: I'm sorry, I'm just baffled by the types. My approach is to keep them as precise as I know how, but when I don't know how to just proceed anyway. I do not have a model of when to expect it to infer what precision of types.

@turadg @mhofman @michaelfig ?

packages/store/test/test-heap-classes.js
erights: Indeed! That thought occurred to me to. I look forward to learning about ava macros.

packages/vat-data/test/test-virtual-classes.js
erights: Good catch! There were several of these, all of which are now changed to
```js
  // @ts-expect-error the type violation is what we're testing
```

packages/vat-data/test/test-vivify.js
erights: Another good catch. Not only do I agree, but my style is not to use harden in this positions. It would indeed be appropriate for a static checker to know all of these definers as hardeners, the way it would know `harden` and `Far` are hardeners.

erights: Oh, I didn't get at first what's different. This is the interfaceKit argument, not the methodsKit argument. The methods and methodsKit are indeed not frozen.

So I still agree, but at least I had not fallen out of the style I'm trying to practice.

packages/store/src/patterns/interface-tools.js
erights: See the comment on line 292 immediately above ;)

packages/vat-data/test/test-vivify.js
erights: Done

packages/store/test/test-heap-classes.js
turadg: I've been surprised by this case too. The type of `upCounter` does have the signature:
Screen Shot 2022-09-04 at 7 46 51 AM

But the type inference is allowing access of unknown properties:
Screen Shot 2022-09-04 at 7 54 53 AM

That's because we allow implicit any. With [noImplicitAny](https://www.typescriptlang.org/tsconfig#noImplicitAny) enabled:
Screen Shot 2022-09-04 at 8 01 40 AM

That's not feasible because it would raise tons of type warnings. Declaring a type (skipping inference) solves the `decr` case, but the declaration doesn't know what `this` will be:
Screen Shot 2022-09-04 at 8 09 10 AM

Example when it does:
Screen Shot 2022-09-04 at 8 11 00 AM

PR #6114, opened 2022-09-02
refactor: method schemas in Governance contracts

### LINKED ISSUE #6060: Utilize method schemas in Governance contracts

Chris-Hibbert:
closes: #6060

## Description

Convert Governance contracts to use method schemas

Can be reviewed as separate commits:

commit 0: basic governance types
commit 1: committee as heapFarClass
commit 2: binaryVoteCounter and questionDetails as heapFarClasses

### Security Considerations

This uses method schemas to enforce types in the APIs.

No changes in behavior.

### Testing Considerations

Existing tests pass.  I didn't add tests for out-of-range parameters.

### DIFF
--- a/packages/governance/src/binaryVoteCounter.js
+++ b/packages/governance/src/binaryVoteCounter.js
@@ -2,12 +2,7 @@
 
 import { Far } from '@endo/marshal';
 import { makePromiseKit } from '@endo/promise-kit';
-import {
-  defineHeapFarClassKit,
-  initEmpty,
-  keyEQ,
-  makeStore,
-} from '@agoric/store';
+import { makeHeapFarInstance, keyEQ, makeStore } from '@agoric/store';
 
 import {
   buildUnrankedQuestion,
@@ -16,7 +11,11 @@ import {
   positionIncluded,
 } from './question.js';
 import { scheduleClose } from './closingRule.js';
-import { BinaryVoteCounterIKit } from './typeGuards.js';
+import {
+  BinaryVoteCounterAdminI,
+  BinaryVoteCounterCloseI,
+  BinaryVoteCounterPublicI,
+} from './typeGuards.js';
 
 const { details: X } = assert;
 
@@ -131,58 +130,70 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
     }
   };
 
-  const makeBinaryVoteCounterKit = defineHeapFarClassKit(
-    'Binary VoteCounter',
-    BinaryVoteCounterIKit,
-    initEmpty,
+  const closeFacet = makeHeapFarInstance(
+    'BinaryVoteCounter close',
+    BinaryVoteCounterCloseI,
     {
-      closeFacet: {
-        closeVoting() {
-          isOpen = false;
-          countVotes();
-        },
-      },
-      creatorFacet: {
-        submitVote(voterHandle, chosenPositions, shares = 1n) {
-          assert(chosenPositions.length === 1, 'only 1 position allowed');
-          const [position] = chosenPositions;
-          assert(
-            positionIncluded(positions, position),
-            X`The specified choice is not a legal position: ${position}.`,
-          );
-
-          // CRUCIAL: If the voter cast a valid ballot, we'll record it, but we need
-          // to make sure that each voter's vote is recorded only once.
-          const completedBallot = harden({ chosen: position, shares });
-          allBallots.has(voterHandle)
-            ? allBallots.set(voterHandle, completedBallot)
-            : allBallots.init(voterHandle, completedBallot);
-          return completedBallot;
-        },
-      },
-      publicFacet: {
-        getQuestion() {
-          return question;
-        },
-        isOpen() {
-          return isOpen;
-        },
-        getOutcome() {
-          return outcomePromise.promise;
-        },
-        getStats() {
-          return tallyPromise.promise;
-        },
-        getDetails() {
-          return details;
-        },
-        getInstance() {
-          return instance;
-        },
+      closeVoting() {
+        isOpen = false;
+        countVotes();
       },
     },
   );
-  return makeBinaryVoteCounterKit();
+
+  const creatorFacet = makeHeapFarInstance(
+    'BinaryVoteCounter creator',
+    BinaryVoteCounterAdminI,
+    {
+      submitVote(voterHandle, chosenPositions, shares = 1n) {
+        assert(chosenPositions.length === 1, 'only 1 position allowed');
+        const [position] = chosenPositions;
+        assert(
+          positionIncluded(positions, position),
+          X`The specified choice is not a legal position: ${position}.`,
+        );
+
+        // CRUCIAL: If the voter cast a valid ballot, we'll record it, but we need
+        // to make sure that each voter's vote is recorded only once.
+        const completedBallot = harden({ chosen: position, shares });
+        allBallots.has(voterHandle)
+          ? allBallots.set(voterHandle, completedBallot)
+          : allBallots.init(voterHandle, completedBallot);
+        return completedBallot;
+      },
+    },
+  );
+
+  const publicFacet = makeHeapFarInstance(
+    'BinaryVoteCounter public',
+    BinaryVoteCounterPublicI,
+    {
+      getQuestion() {
+        return question;
+      },
+      isOpen() {
+        return isOpen;
+      },
+      getOutcome() {
+        return outcomePromise.promise;
+      },
+      getStats() {
+        return tallyPromise.promise;
+      },
+      getDetails() {
+        return details;
+      },
+      getInstance() {
+        return instance;
+      },
+    },
+  );
+
+  return harden({
+    creatorFacet,
+    publicFacet,
+    closeFacet,
+  });
 };
 
 // The contract wrapper extracts the terms and runs makeBinaryVoteCounter().
--- a/packages/governance/src/binaryVoteCounter.js
+++ b/packages/governance/src/binaryVoteCounter.js
@@ -2,15 +2,21 @@
 
 import { Far } from '@endo/marshal';
 import { makePromiseKit } from '@endo/promise-kit';
-import { keyEQ, makeStore } from '@agoric/store';
+import {
+  defineHeapFarClassKit,
+  initEmpty,
+  keyEQ,
+  makeStore,
+} from '@agoric/store';
 
 import {
-  ChoiceMethod,
   buildUnrankedQuestion,
-  positionIncluded,
+  ChoiceMethod,
   coerceQuestionSpec,
+  positionIncluded,
 } from './question.js';
 import { scheduleClose } from './closingRule.js';
+import { BinaryVoteCounterIKit } from './typeGuards.js';
 
 const { details: X } = assert;
 
@@ -77,24 +83,6 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
   /** @type {Store<Handle<'Voter'>,RecordedBallot> } */
   const allBallots = makeStore('voterHandle');
 
-  /** @type {SubmitVote} */
-  const submitVote = (voterHandle, chosenPositions, shares = 1n) => {
-    assert(chosenPositions.length === 1, 'only 1 position allowed');
-    const [position] = chosenPositions;
-    assert(
-      positionIncluded(positions, position),
-      X`The specified choice is not a legal position: ${position}.`,
-    );
-
-    // CRUCIAL: If the voter cast a valid ballot, we'll record it, but we need
-    // to make sure that each voter's vote is recorded only once.
-    const completedBallot = harden({ chosen: position, shares });
-    allBallots.has(voterHandle)
-      ? allBallots.set(voterHandle, completedBallot)
-      : allBallots.init(voterHandle, completedBallot);
-    return completedBallot;
-  };
-
   const countVotes = () => {
     assert(!isOpen, X`can't count votes while the election is open`);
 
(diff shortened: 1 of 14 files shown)


packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,103 @@
...
+  tieOutcome: NoOfferFilterPositionShape,
+});
+
+// keys are parameter names, values are proposed values
+export const ParamChangesSpecShape = M.recordOf(M.string(), M.any());
+export const YesParamChangesPositionShape = ParamChangesSpecShape;
+export const NoParamChangesPositionShape = harden({
+  noChange: M.arrayOf(M.string()),
+});
+export const ParamChangesPositionsShape = [
```
erights:
```suggestion
export const ParamChangesPositionsShape = harden([
```
and several other places

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,103 @@
...
+export const QuorumRuleShape = M.or('majority', 'no_quorum', 'all');
+export const ElectionTypeShape = M.or(
+  'param_change',
+  'election',
+  'survey',
+  'api_invocation',
+  'offer_filter',
+);
+
+export const TimerShape = M.handle('timer');
```
erights: For a remotable, the most that the pattern system can enforce is that it is a remotable. The string argument is only for diagnostic purposes. So the following does all the enforcement you have in mind, but by itself fails to provide the diagnostic info that you're expecting a handle.
```suggestion
export const TimerShape = M.remotable('timer');
```
you could just indicate that in the name manually
```suggestion
export const TimerShape = M.remotable('timerHandle');
```
or, if you think it's worth it, create an abstraction that just does the name mangling
```suggestion
const makeHandleShape = name => M.remotable(`${name}Handle`);
//...
export const TimerShape = makeHandleShape('timer');
```

packages/governance/src/typeGuards.js
Chris-Hibbert: done

packages/governance/src/typeGuards.js
Chris-Hibbert: cool!  thanks.

packages/governance/src/binaryVoteCounter.js
```diff
@@ -142,29 +131,57 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
...
+        getDetails() {
+          return details;
+        },
+        getInstance() {
+          return instance;
+        },
+      },
+    },
+  );
+  return makeBinaryVoteCounterKit();
```
erights: It looks like this classKit is only instantiated once, to get three singleton instances. You should instead just use `makeHeapFarInstance` for each, as you do in noActionElectorate.js

packages/governance/src/binaryVoteCounter.js
```diff
@@ -142,29 +131,57 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
...
+          return question;
+        },
+        isOpen() {
+          return isOpen;
+        },
+        getOutcome() {
+          return outcomePromise.promise;
+        },
+        getStats() {
+          return tallyPromise.promise;
```
erights: From this code, I'm guessing this returns a promise. But the guard declaration in typeGuards.js says it returns a `QuestionStatsShape`, which is a CopyRecord. IIUC, this should always fail on return.

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,285 @@
...
+  quorumRule: 'majority',
+  tieOutcome: NoParamChangesPositionShape,
+});
+
+export const ParamChangesQuestionDetailsShape = harden({
+  method: 'unranked',
+  issue: ParamChangesIssueShape,
+  positions: ParamChangesPositionsShape,
+  electionType: 'param_change',
+  maxChoices: M.eq(1),
```
erights: Constants match only themselves, which you're correctly making use of for strings.
```suggestion
  maxChoices: 1,
```

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,285 @@
...
+});
+
+export const BinaryVoteCounterPublicI = M.interface(
+  'BinaryVoteCounter PublicFacet',
+  {
+    // XXX I expected M.call().returns(QuestionDetailI)
+    getQuestion: M.call().returns(M.remotable()),
+    isOpen: M.call().returns(M.boolean()),
+    getOutcome: M.call().returns(M.eref(PositionShape)),
+    getStats: M.call().returns(QuestionStatsShape),
```
erights: The implementation looks like it returns a promise.

packages/governance/src/contractGovernance/governParam.js
```diff
@@ -34,30 +35,6 @@ const makeParamChangePositions = changes => {
   return harden({ positive, negative });
 };
 
-/** @type {ValidateParamChangeQuestion} */
-const validateParamChangeQuestion = details => {
```
erights, line 38: Great to see this disappear! But the type still needs to be deleted from types.js

packages/governance/src/contractGovernor.js
```diff
@@ -27,7 +28,7 @@ const validateQuestionDetails = async (zoe, electorate, details) => {
     counterInstance,
     issue: { contract: governedInstance },
   } = details;
-  validateParamChangeQuestion(details);
+  fit(details, ParamChangesQuestionDetailsShape);
```
erights, line 31: ;)

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,285 @@
...
+  'election',
+  'survey',
+  'api_invocation',
+  'offer_filter',
+);
+
+const makeHandleShape = name => M.remotable(`${name}Handle`);
+
+export const TimerShape = makeHandleShape('timer');
+export const InstanceShape = makeHandleShape('instance');
```
erights: Btw, zoe/src/typeGuards.js exports this as `InstanceHandleShape`. However, zoe as a whole doesn't export it. I'm thinking it should. When I did this I missed the `timer: M.remotable(),` on like 33 of that file, that should have been replaced with the `TimerShape` you define above.

I think a package like Zoe should export the shapes that correspond to concepts defined by that package. But that refactor is distinct from this PR. No change suggested.

packages/governance/src/noActionElectorate.js
```diff
@@ -13,35 +15,45 @@ import { makePromiseKit } from '@endo/promise-kit';
...
   });
 
-  const creatorFacet = Far('creatorFacet', {
-    getPoserInvitation: () => {
+  const creatorFacet = makeHeapFarInstance('creatorFacet', ElectorateCreatorI, {
+    getPoserInvitation() {
       return zcf.makeInvitation(() => {},
       `noActionElectorate doesn't allow posing questions`);
     },
     addQuestion() {
```
erights: I see from the guard declaration that this method has two mandatory parameters. I prefer to have these appear in the function signature, prefixed with underbar if not otherwise used.
```suggestion
    addQuestion(_instance, _question) {
```

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,285 @@
...
+  getInstance: M.call().returns(InstanceShape),
+  getQuestion: M.call(QuestionHandleShape).returns(M.promise()),
+});
+
+export const ElectorateCreatorI = M.interface('Committee AdminFacet', {
+  getPoserInvitation: M.call().returns(M.promise()),
+  addQuestion: M.call(InstanceShape, QuestionSpecShape).returns(M.promise()),
+  getVoterInvitations: M.call().returns(M.arrayOf(M.promise())),
+  getQuestionSubscriber: M.call().returns(SubscriberShape),
+  getPublicFacet: M.call().returns(ElectoratePublicI),
```
erights: This hazard is on me, and I may eventually fix it. But at least for now an InterfaceGuard such as `ElectoratePublicI` is not itself a pattern.

```suggestion
const ElectoratePublicShape = M.remotable('ElectoratePublic');
...
  getPublicFacet: M.call().returns(ElectoratePublicShape),
```

erights (changes requested):
Review going well but not done yet.

packages/governance/src/binaryVoteCounter.js
Chris-Hibbert: Done

packages/governance/src/binaryVoteCounter.js
Chris-Hibbert: done

packages/governance/src/contractGovernance/governParam.js
Chris-Hibbert: done

packages/governance/src/noActionElectorate.js
Chris-Hibbert: done

packages/governance/src/typeGuards.js
Chris-Hibbert: I added them to Zoe, as they don't belong here.

packages/governance/src/typeGuards.js
Chris-Hibbert: done

packages/governance/src/typeGuards.js
Chris-Hibbert: Done

packages/governance/src/typeGuards.js
Chris-Hibbert: done

packages/governance/src/committee.js
```diff
@@ -90,50 +114,94 @@ const start = (zcf, privateArgs) => {
...
-  const creatorFacet = Far('adminFacet', {
-    getPoserInvitation: () => getPoserInvitation(zcf, addQuestion),
-    addQuestion,
-    getVoterInvitations: () => invitations,
-    getQuestionSubscriber: () => questionsSubscriber,
-    getPublicFacet: () => publicFacet,
-  });
-
-  return { publicFacet, creatorFacet };
+  return makeCommitteeKit();
```
erights: Another one where the kit is instantiated only once, so the objects in the kit are singletons. You should just use `makeHeapFarInstance` on each.

packages/governance/src/committee.js
```diff
@@ -90,50 +114,94 @@ const start = (zcf, privateArgs) => {
...
-    };
+  const makeCommitteeKit = defineHeapFarClassKit(
+    'Committee Facets',
+    CommitteeIKit,
+    initEmpty,
+    {
+      creatorFacet: {
+        getPoserInvitation() {
+          return getPoserInvitation(zcf, async (voteCounter, questionSpec) => {
+            const quorumThreshold = quorumRule => {
```
erights: AFAICT, the body of the argument function here is identical to the body of the `addQuestion` method below. Although the switch to inline concise methods admittedly make sharing this function harder, remember that you can send messages to self.

```suggestion
        getPoserInvitation() {
          const { facet: { creatorFacet } } = this;
          return getPoserInvitation(zcf, async (voteCounter, questionSpec) =>
            creatorFacet.addQuestion(voteCounter, questionSpec),
          );
        },
```

The suggestion above is if you stick with the `defineHeapFarClassKit`. If you switch to multiple `makeHeapFarInstance`s and call the lexical variable `creatorFacet`, then the thisful line disappears and the rest of the suggestion still applies.

packages/governance/src/question.js
```diff
@@ -205,30 +65,40 @@ const coerceQuestionSpec = ({
     quorumRule,
     tieOutcome,
   });
+
+  fit(question, QuestionSpecShape);
+
+  // XXX It would be nice to enforce this using parameterized types, but that
+  // seems to only enforce type constraints, (i.e. the tieOutcome will be the
+  // same type as any of the positions) unless you can provide the concrete
+  // value at pattern creation time.
```
erights, line 74: yes. No change suggested.

packages/governance/src/question.js
```diff
@@ -43,124 +38,9 @@ const QuorumRule = /** @type {const} */ ({
   ALL: 'all',
 });
 
-/**
```
erights, line 46: Re the deletions from here (line 46) through line 196:

OMG that's a lot of manual input validation code deleted! Awesome!

It strikes me that this might serve as one of the extremes for our benchmarking, comparing manual input validation to pattern-based. Attn @gibson042 @warner

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,272 @@
...
+  issue: SimpleIssueShape,
+  positions: SimplePositionsShape,
+  electionType: M.or('election', 'survey'),
+  maxChoices: M.gte(1),
+  closingRule: ClosingRuleShape,
+  quorumRule: QuorumRuleShape,
+  tieOutcome: NoSimplePositionShape,
+  questionHandle: makeHandleShape('Question'),
+  counterInstance: InstanceHandleShape,
+});
```
erights: AFAICT, the parts in common with `SimpleQuestionSpecShape` are identical. If this is an intentional part of the design, then

```suggestion
export const SimpleQuestionDetailsShape = harden({
  ...SimpleQuestionSpecShape,
  questionHandle: makeHandleShape('Question'),
  counterInstance: InstanceHandleShape,
});
```

Wherever interface extension is an intentional design, this is a good way to express it.

packages/governance/src/typeGuards.js
```diff
@@ -0,0 +1,272 @@
...
+
+// XXX I want to add questionHandle and counterInstance to
+// ParamChangesQuestionSpecShape. I don't see any alternative to adding the
+// methods to each member separately
+export const QuestionDetailsShape = M.or(
+  ParamChangesQuestionDetailsShape,
+  ApiInvocationQuestionDetailsShape,
+  OfferFilterQuestionDetailsShape,
+  SimpleQuestionDetailsShape,
+);
```
erights, line 179: Thanks for putting the question inline. Would be good for anything in common to be pulled out of the expensive `M.or`. If you pull these two common fields out of the branches, then you should also rename the branches to not imply that they're meaningful by themselves, and stop exporting them. For three of these, they are only used here, so no problem. But `ParamChangesQuestionDetailsShape` is used elsewhere. Assuming that this is dealt with somehow, we could rewrite the factored pattern as

```suggestion
export const QuestionDetailsShape = M.split({
    questionHandle: QuestionHandleShape, // making the handle shape should already be factored out.
    counterInstance: InstanceHandleShape,
  },
  M.or(
    ParamChangesQuestionPartialDetailsShape,
    ApiInvocationQuestionPartialDetailsShape,
    OfferFilterQuestionPartialDetailsShape,
    SimpleQuestionPartialDetailsShape,
  ),
});
```

erights (approved):
LGTM!

packages/governance/src/committee.js
Chris-Hibbert: done

packages/governance/src/committee.js
Chris-Hibbert: Thank you.  I should have noticed the duplication and looked for an appropriate way to get the sharing back.

packages/governance/src/typeGuards.js
Chris-Hibbert: Thanks. I was looking for that, and didn't find it.

packages/governance/src/typeGuards.js
Chris-Hibbert: Thanks for the explanation. That's a helpful alternative.

Now that I see how `ParamChangesQuestionDetailsShape` is used, I wouldn't be surprised to find other uses of the other `QuestionDetailsShape`s. I think I'll leave it as is.

PR #6117, opened 2022-09-02
feat(deploy-script-helpers): Publish bundle

### LINKED ISSUE #5618: Encapsulate bundleSource / publishBundle under deploy script helper

One workflow for auditing the code for a contract is:
1. to go to the source repository,
2. check out the advertised version of the contract,
3. install the dependencies pinned in the lockfile,
4. generate a contract bundle from the sources, including the consistent hash of the bundle,
5. review the contract,
6. sign and publish findings, mentioning the consistent hash of the bundle

`bundleSource` is responsible for generating a consistent hash, and the hash varies by the version of `bundleSource` and its transitive dependencies, as well as the versions of the dependencies of the contract.

Currently, contract deploy scripts must be programmed against these concerns and currently receive a floating version of `bundleSource` as a power, from the `agoric-sdk` tooling:

```js
const sourceBundle = await bundleSource(path);
const hashBundle = await publishBundle(sourceBundle);
await E(zoe).install(hashBundle);
```

This poses two problems. The Agoric tooling needs to be kept in sync with the version of the chain. That means that the version of `bundleSource` will drift from the version that was used to publish the contract. Second, this places a cognitive burden on the author of a deploy script, who only needs to know the verb `install`.

## Description of the Design

@michaelfig and I have arrived at a design that solves both problems. In `@agoric/deploy-script-helpers`, we already have a `makeInstall` function that takes the open bag of deploy powers and the home config. This captures the current version of the above incantation and can be revised to take `publishBundle` from the vat powers.

```js
import { makeInstaller } from '@agoric/deploy-script-helpers';
const install = makeInstaller(home, powers);
await install(path);
```

This ensures that the version of `@endo/bundle-source` is captured in `yarn.lock` as a dependency of the deploy script, relieves the deploy script of its dependency on `powers.bundleSource`, and takes a dependency on `powers.publishBundle`, such that deploy script powers encapsulate all of these concerns.

kriskowal:
- fix(deploy-script-support): Use versioned bundleSource
- feat(deploy-script-support): Publish bundles to chain if possible

closes: #5618

## Description

If a deploy script receives a publishBundle power from the `agoric deploy` command, it can now publish the code to a chain and submit the hash in a delivery transaction to Zoe.  This will allow us to reduce the transaction size limits for these separate lanes: publishing and delivering messages.

This changes the deploy script helpers to take advantage of publishBundle if it's available.

This change should expressly reduce the burden on documentation, as opposed to having this behavior visible in all deployment scripts.

### Testing Considerations

The Dapp Fungible Faucet (our template for new dapps) does not currently use the helpers.
https://github.com/Agoric/dapp-fungible-faucet/pull/53 recruits the dapp fungible faucet integration test to exercise publishBundle.
I've verified these changes locally with a scenario2 fake chain.

dapp-fungible-faucet-branch: 4564-helpers

### DIFF
--- a/packages/deploy-script-support/src/helpers.js
+++ b/packages/deploy-script-support/src/helpers.js
@@ -22,7 +22,7 @@ const ZOE_INVITE_PURSE_PETNAME = 'Default Zoe invite purse';
 export const makeHelpers = async (homePromise, endowments) => {
   const { zoe, wallet, scratch, board } = E.get(homePromise);
 
-  const { lookup } = endowments;
+  const { lookup, publishBundle, pathResolve } = endowments;
 
   const walletAdmin = E(wallet).getAdminFacet();
   const installationManager = E(walletAdmin).getInstallationManager();
@@ -35,7 +35,14 @@ export const makeHelpers = async (homePromise, endowments) => {
 
   // Create the methods
 
-  const install = makeInstall(bundleSource, zoe, installationManager, board);
+  const install = makeInstall(
+    bundleSource,
+    zoe,
+    installationManager,
+    board,
+    publishBundle,
+    pathResolve,
+  );
 
   const startInstance = makeStartInstance(
     issuerManager,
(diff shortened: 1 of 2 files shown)


kriskowal:
Poke for review, @michaelfig

michaelfig (approved):

PR #6119, opened 2022-09-03
feat(psm): far class and guard patterns

### LINKED ISSUE #6061: Utilize method schemas in PSM and mintHolder contracts

### LINKED ISSUE #6085: Implement Far classes in PSM

dtribble:
closes: #6085
closes: #6061

## Description

    feat(inter-protocol): far class for PSM
    
    - convert PSM to FarClass
    - enforce proposal safety via patterns
    - integrate governance into farClasses

### Security Considerations

Keeps API surface unchanged

N/a

### Testing Considerations

Added positive and negative tests for contract offers..

### DIFF
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -4,13 +4,29 @@ import { Far } from '@endo/marshal';
 import { makeStoredPublisherKit } from '@agoric/notifier';
 import { getMethodNames, objectMap } from '@agoric/internal';
 import { ignoreContext } from '@agoric/vat-data';
-import { keyEQ } from '@agoric/store';
+import { keyEQ, M } from '@agoric/store';
+import { AmountShape, BrandShape } from '@agoric/ertp';
 import { assertElectorateMatches } from './contractGovernance/paramManager.js';
 import { makeParamManagerFromTerms } from './contractGovernance/typedParamManager.js';
 
 const { details: X, quote: q } = assert;
 
 export const GOVERNANCE_STORAGE_KEY = 'governance';
+
+const extendPublicAPI = {
+  getSubscription: M.call().returns(M.remotable('StoredSubscription')),
+  getContractGovernor: M.call().returns(M.remotable('Instance')),
+  getGovernedParams: M.call().returns(M.record()),
+  getAmount: M.call().returns(AmountShape),
+  getBrand: M.call().returns(BrandShape),
+  getInstance: M.call().returns(M.remotable('Instance')),
+  getInstallation: M.call().returns(M.remotable('Installation')),
+  getInvitationAmount: M.call().returns(M.promise()),
+  getNat: M.call().returns(M.bigint()),
+  getRatio: M.call().returns(M.record()),
+  getString: M.call().returns(M.string()),
+  getUnknown: M.call().returns(M.string()),
+};
 
 /**
  * Utility function for `makeParamGovernance`.
@@ -102,12 +118,12 @@ const facetHelpers = (zcf, paramManager) => {
 
   /**
    * @template {{}} CF
-   * @param {CF} originalCreatorFacet
+   * @param {CF} limitedCreatorFacet
+   * @param limitedCreatorFacet
    * @param {{}} [governedApis]
    * @returns {GovernorFacet<CF>}
    */
-  const makeGovernorFacet = (originalCreatorFacet, governedApis = {}) => {
-    const limitedCreatorFacet = makeLimitedCreatorFacet(originalCreatorFacet);
+  const makeFarGovernorFacet = (limitedCreatorFacet, governedApis = {}) => {
     const governorFacet = Far('governorFacet', {
       getParamMgrRetriever: () =>
         Far('paramRetriever', { get: () => paramManager }),
@@ -127,6 +143,17 @@ const facetHelpers = (zcf, paramManager) => {
 
     // exclusively for contractGovernor, which only reveals limitedCreatorFacet
     return governorFacet;
+  };
+
+  /**
+   * @template {{}} CF
+   * @param {CF} originalCreatorFacet
+   * @param {{}} [governedApis]
+   * @returns {GovernorFacet<CF>}
+   */
+  const makeGovernorFacet = (originalCreatorFacet, governedApis = {}) => {
+    const limitedCreatorFacet = makeLimitedCreatorFacet(originalCreatorFacet);
+    return makeFarGovernorFacet(limitedCreatorFacet, governedApis);
   };
 
   /**
@@ -159,10 +186,19 @@ const facetHelpers = (zcf, paramManager) => {
   };
 
   return harden({
+    extendPublicAPI,
+    extendPublicFacet: {
+      ...commonPublicMethods,
+      ...typedAccessors,
+    },
+    extendCreatorFacet: {
+      getContractGovernor: () => electionManager,
+    },
     augmentPublicFacet,
     augmentVirtualPublicFacet,
     makeGovernorFacet,
     makeVirtualGovernorFacet,
+    makeFarGovernorFacet,
     params: paramManager.readonly(),
   });
 };
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -13,7 +13,7 @@ const { details: X, quote: q } = assert;
 
 export const GOVERNANCE_STORAGE_KEY = 'governance';
 
-const extendPublicAPI = {
+const publicMixinAPI = harden({
   getSubscription: M.call().returns(M.remotable('StoredSubscription')),
   getContractGovernor: M.call().returns(M.remotable('Instance')),
   getGovernedParams: M.call().returns(M.record()),
@@ -26,7 +26,7 @@ const extendPublicAPI = {
   getRatio: M.call().returns(M.record()),
   getString: M.call().returns(M.string()),
   getUnknown: M.call().returns(M.string()),
-};
+});
 
 /**
  * Utility function for `makeParamGovernance`.
@@ -119,7 +119,6 @@ const facetHelpers = (zcf, paramManager) => {
   /**
    * @template {{}} CF
    * @param {CF} limitedCreatorFacet
-   * @param limitedCreatorFacet
    * @param {{}} [governedApis]
    * @returns {GovernorFacet<CF>}
    */
@@ -186,12 +185,11 @@ const facetHelpers = (zcf, paramManager) => {
   };
 
   return harden({
-    extendPublicAPI,
-    extendPublicFacet: {
+    publicMixin: {
       ...commonPublicMethods,
       ...typedAccessors,
     },
-    extendCreatorFacet: {
+    creatorMixin: {
       getContractGovernor: () => electionManager,
     },
     augmentPublicFacet,
@@ -248,4 +246,4 @@ const handleParamGovernance = async (
 
 harden(handleParamGovernance);
 
-export { handleParamGovernance };
+export { handleParamGovernance, publicMixinAPI };
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -25,7 +25,7 @@ const publicMixinAPI = harden({
   getNat: M.call().returns(M.bigint()),
   getRatio: M.call().returns(M.record()),
   getString: M.call().returns(M.string()),
-  getUnknown: M.call().returns(M.string()),
+  getUnknown: M.call().returns(M.any()),
 });
 
 /**
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -164,7 +164,7 @@ const facetHelpers = (zcf, paramManager) => {
     const limitedCreatorFacet = makeLimitedCreatorFacet(originalCreatorFacet);
 
     /** @type {import('@agoric/vat-data/src/types.js').FunctionsPlusContext<unknown, GovernorFacet<originalCreatorFacet>>} */
-    const governorFacet = Far('governorFacet', {
+    const governorFacet = harden({
       getParamMgrRetriever: () =>
         Far('paramRetriever', { get: () => paramManager }),
       getInvitation: (_context, name) =>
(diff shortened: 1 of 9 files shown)


packages/governance/src/contractHelper.js
```diff
@@ -4,14 +4,30 @@ import { Far } from '@endo/marshal';
...
+import { keyEQ, M } from '@agoric/store';
+import { AmountShape, BrandShape } from '@agoric/ertp';
 import { assertElectorateMatches } from './contractGovernance/paramManager.js';
 import { makeParamManagerFromTerms } from './contractGovernance/typedParamManager.js';
 
 const { details: X, quote: q } = assert;
 
 export const GOVERNANCE_STORAGE_KEY = 'governance';
 
+const extendPublicAPI = {
```
dtribble: This could just be an export.

erights:
```suggestion
const extendPublicAPI = harden({
```

packages/governance/src/contractHelper.js
```diff
@@ -102,12 +118,12 @@ const facetHelpers = (zcf, paramManager) => {
 
   /**
    * @template {{}} CF
-   * @param {CF} originalCreatorFacet
+   * @param {CF} limitedCreatorFacet
+   * @param limitedCreatorFacet
```
erights:
```suggestion
   * @param {CF} limitedCreatorFacet
```

packages/inter-protocol/src/psm/psm.js
```diff
@@ -243,47 +247,73 @@ export const start = async (zcf, privateArgs, baggage) => {
...
+            give: { In: stableAmountShape },
+            want: M.or({ Out: anchorAmountShape }, {}),
+          }),
+        );
+      },
+      ...extendPublicFacet,
+    },
+  );
+
+  // TODO why does this operation return an object with a single operation?
```
turadg, line 297: are you suggesting that, https://github.com/Agoric/agoric-sdk/blob/f338d2e09087dce71df148375e990256f4301bff/packages/inter-protocol/src/collectFees.js#L29

instead be simply,
```js
return makeCollectFeesInvitation;
```

👍  to that, especially since the function is named `makeMakeCollectFeesInvitation`.

packages/governance/src/contractHelper.js
```diff
@@ -4,14 +4,30 @@ import { Far } from '@endo/marshal';
...
 
+const extendPublicAPI = {
+  getSubscription: M.call().returns(M.remotable('StoredSubscription')),
+  getContractGovernor: M.call().returns(M.remotable('Instance')),
+  getGovernedParams: M.call().returns(M.record()),
+  getAmount: M.call().returns(AmountShape),
+  getBrand: M.call().returns(BrandShape),
+  getInstance: M.call().returns(M.remotable('Instance')),
+  getInstallation: M.call().returns(M.remotable('Installation')),
+  getInvitationAmount: M.call().returns(M.promise()),
```
Chris-Hibbert, line 24: It's an Amount.  Can we embed the type inside a `Promise`?

packages/governance/src/contractHelper.js
```diff
@@ -4,14 +4,30 @@ import { Far } from '@endo/marshal';
...
+  getNat: M.call().returns(M.bigint()),
+  getRatio: M.call().returns(M.record()),
+  getString: M.call().returns(M.string()),
+  getUnknown: M.call().returns(M.string()),
```
Chris-Hibbert: `returns(M.any()),`

packages/inter-protocol/src/psm/psm.js
```diff
@@ -94,18 +93,23 @@ export const start = async (zcf, privateArgs, baggage) => {
...
-        MintLimit: ParamTypes.AMOUNT,
-        WantStableFee: ParamTypes.RATIO,
-      },
-      privateArgs.storageNode,
-      privateArgs.marshaller,
-    );
+  const {
+    extendPublicAPI,
+    extendPublicFacet,
+    extendCreatorFacet,
```
Chris-Hibbert: extend seems like a verb. I'd prefer this be a noun like 'mixin'.  'additionalBehavior' (& `additionalTypes`) would also work for me to indicate that it's behavior to be added to something.

packages/governance/src/contractHelper.js
erights: You can currently do `M.eref(AmountShape)`. But I'm thinking of removing `M.eref` because it looks like it is checking something stronger than it is. It checks that the specimen is either a promise or an amount. But if it is a promise, it does *not* check that the promise fulfills to an amount, because it cannot check that synchronously. We'd have the same problem with `M.promise(AmountShape)` except that it would never be stronger than "is it a promise". I omitted having such a parameter to avoid implying that more is being enforced than it is.

That said, we already live with this problem with `M.remotable`. Perhaps if the `M.promise` argument were just a string to be used in error diagnostics it would be clearer? I don't know. It makes me quesy in a way that `M.remotable(string)` doesn't. Opinions?

Chris-Hibbert: Yeah, after a bit I realized that the argument to `M.promise()` can't be more than a hint for reporting and that's probably why you omitted it. 

If it was a string whose name was 'typeDescription' or 'nameForErrorReporting', would that make you more comfortable? My IDE shows me the names of parameters when it's hinting at completions.

How about if it were `M.call().returns(M.promise().hint('Invitation'))`?

packages/governance/src/contractHelper.js
dtribble: It's currently only provided in a hardened object, so not actually necessary. However it should move to an exported constant so aye.

packages/inter-protocol/src/psm/psm.js
dtribble: I *think* that it's nested because of some restriction in some usage where you can't return a function...  But I'll try fixing that if it's easy.

packages/governance/src/contractHelper.js
dtribble: I would like exactly the same debugging support as for `remotable`. I'm not sure it's worth the additional `hint` usage. Particularly if it's a string, it's clear that's it's not actionable.

packages/governance/src/contractHelper.js
dtribble: I'm confused. Are you suggesting that these should have broader types rather than the specifics?

packages/inter-protocol/src/psm/psm.js
dtribble: `publicMixinAPI`, `publicMixin`, `creatorMixin`

packages/governance/src/contractHelper.js
erights: Yeah, I think the uniformity with `M.remotable` is persuasive. I will suppress my queasiness ;)

packages/governance/src/contractHelper.js
Chris-Hibbert: No, I'm saying that exactly `getUnknown` returns `M.any()` rather than a string. The whole point is that it's an escape from the type mechanism for types we don't support yet.

packages/governance/src/contractHelper.js
```diff
@@ -102,12 +118,11 @@ const facetHelpers = (zcf, paramManager) => {
```
erights, line 118: The `ignoreContext` on like 94 above cannot be right. Probably you should double check all using of ignoreContext. We should only ever use it when defining a kind, not a far class.

packages/governance/src/contractHelper.js
```diff
@@ -159,10 +185,18 @@ const facetHelpers = (zcf, paramManager) => {
   };
```
erights, line 185: The `Far` on line 167 above encloses a record of kind methods --- methods that take a context as their first argument. This cannot be correct. If this is supposed to be a record of kind methods, ultimately to be used in a define*Kind call rather than a far class, then you should use `harden` rather than `Far`. You should also flag wherever this methods get passed around, because as we shift from kinds to far classes, we'll want to change all of these.

packages/governance/src/contractHelper.js
turadg: +1 to indicating what's expected inside the promise, even if it's not verified. This at least serves as documentation for the developer writing and debugging.

It could also be interpreted by software to create the function signature for `getInvitationAmount`, which needs to say what the function is supposed to do. Eventually it could be fed into the remote call handling to provide an automated check at runtime.

packages/inter-protocol/src/psm/psm.js
turadg: If so then I'd suggest the outer maker is returning an `InvitationMakers` type (the thing a continuing invitation returns). Then I'd suggest the outer be renamed to `makeCollectFeesInvitationMakers`. It's awkward because there's just but "InvitationMakers" is the object type.

Alternately, we could come up with a more concise term for "object whose properties are all invitation makers".

erights: InvitationMakersKit

turadg: That is certainly not more concise :-P 

More importantly, doesn't FooKit mean "foo and related"? (à la IssuerKit) This is not returning any more than invitation makers.

erights: I'd say *Kit means a set of related things returned as the values of named properties in a copyRecord. I'd say *FooKit says they are related to Foo, whatever that is. For an InvitationMakersKit, I would not be surprised to find a copyRecord of InvitationMakers.

That was off the top of my head. If we have a definition somewhere, I'd be curious how it differs.

packages/governance/src/contractHelper.js
turadg: By the notion [above about hints](https://github.com/Agoric/agoric-sdk/pull/6119#discussion_r962320291), should this say that it returns `M.unknown()`?

erights: How does `M.unknown()` differ from `M.any()`? Are they operationally equivalent, but if generating TS type text from patterns, would differ in what types they generate?

turadg: Exactly. At runtime they both mean "nothing to check". As a param, any and unknown both mean "accepts without checking". At use, unknown means no guarantees (so you have to check! and any means treat it as any type.

Wondering if something like "M.defined" would be useful to guarantee that's is some value, not null or undefined. Then the caller can safely discriminate on properties.

None of these hunting ideas seem like blockers to me.

packages/inter-protocol/src/psm/psm.js
dtribble: All these seem to be building the type into the name. That is generally a bad practice (bad hungarian). Instead, what's the purpose or intent?

packages/governance/src/contractHelper.js
dtribble: It's easy enough to add a clone of `any` named `unknown` to allow the right definitions.

packages/governance/src/contractHelper.js
dtribble: I'm now using the version of those methods that does NOT use the context argument.

packages/inter-protocol/src/psm/psm.js
turadg: Purpose is to get additional invitations from the result of the settled one. `continuingInvitations`?

packages/governance/src/contractHelper.js
Chris-Hibbert: Regardless of what we'll eventually do, getUnknown can't return `M.string()`. Please choose something else for the interim.

packages/inter-protocol/src/psm/psm.js
```diff
@@ -243,47 +247,75 @@ export const start = async (zcf, privateArgs, baggage) => {
...
+            give: { In: stableAmountShape },
+            want: M.or({ Out: anchorAmountShape }, {}),
+          }),
+        );
+      },
+      ...publicMixin,
+    },
+  );
+
+  // TODO why does this operation return an object with a single operation?
```
Chris-Hibbert, line 297: It gets used in places where objects can be passed, but functions cannot.  If that's no longer true, it can be simplified, I think.

Chris-Hibbert (changes requested):

packages/inter-protocol/src/psm/psm.js
erights: Do we know yet whether this can be simplified?

packages/inter-protocol/src/psm/psm.js
```diff
@@ -243,47 +247,75 @@ export const start = async (zcf, privateArgs, baggage) => {
...
     zcf,
     feePool,
     stableBrand,
     'Stable',
   );
-  const creatorFacet = Far('Parity Stability Module', {
-    getRewardAllocation,
-    makeCollectFeesInvitation,
+
+  // These are internal so they do not yet have defensive programming
```
erights: I think this comment overstates the implied degree of unsafety. On integrity, the object below is defined using only two bespoke methods, both of which have zero arguments. One returns an allocation, which is a by-copy structure. The other returns an invitation, which we presume to already be defensive.

What the comment is about is that it isn't defined using interface guards, and so not defensive against extra arguments being provided that might take up space. Aside from space, these extra arguments are ignored, so cannot do any other damage. There's also nothing retaining the extra arguments, so they'll be gc-able immediately anyway, which is as fast as discarding them (currently) via interface guards.

erights (approved):
LGTM

I did only a normal differential review. psm.js is short enough that I had wanted to do a from-scratch review of that file, but that'll have to wait until another time. I don't need to block this PR on my desire to do a from-scratch review.

packages/inter-protocol/src/psm/psm.js
dtribble: I like that.

packages/governance/src/contractHelper.js
dtribble: I agree, but it can wait until those patterns have been enhanced with the debug info fields.

packages/governance/src/contractHelper.js
dtribble: Already changed to M.any.

packages/governance/src/contractHelper.js
dtribble: This doesn't seem to point to a line of code. I think this has been addressed.

packages/inter-protocol/src/psm/psm.js
dtribble: I agree. I'll update the comment if I have to redo anything but otherwise isn't worth a trip through CI.

Chris-Hibbert (approved):
LGTM

PR #6120, opened 2022-09-03
fix: convert upgradable covered call to far classes

erights:
Staged on #5960 

TODO explain


Chris-Hibbert (approved):
The test passes!  Woohoo!

LGTM.

PR #6121, opened 2022-09-03
deprecate existing smart-wallet package

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/6084 is a total rewrite of the contract which is not backwards compatible. To land it without having to update everything else in the same PR, we need the new and old to coexist.

The approach taken here is to fork at the package level. It could have been by filenames or directories, but this way has less churn (leaves the files alone) and provides the opportunity to make the smart-wallet package fit the `packages/*` scheme.

### Security Considerations

--

This renames the package but that shouldn't be visible externally. When we publish again the `@agoric/smart-wallet` package will be the new one (with a higher version number, 0.2.0 from 0.1.1). The `legacy-smart-wallet` is marked private to not publish.

### Testing Considerations

CI


dckc (approved):
looks consistent with what we discussed

PR #6122, opened 2022-09-03
refactorings for new smart-wallet

turadg:
## Description

Misc refactors and other code improvements to support https://github.com/Agoric/agoric-sdk/pull/6084

### Security Considerations

--

Slightly improved in some places

### Testing Considerations

--

### DIFF
--- a/packages/vats/src/core/boot-psm.js
+++ b/packages/vats/src/core/boot-psm.js
@@ -138,6 +138,7 @@ export const buildRootObject = (vatPowers, vatParameters) => {
       ...ECON_COMMITTEE_MANIFEST,
       ...PSM_MANIFEST,
     };
+    /** @param {string} name */
     const powersFor = name => {
       const permit = manifest[name];
       assert(permit, `missing permit for ${name}`);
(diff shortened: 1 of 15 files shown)


packages/vats/src/core/boot-psm.js
```diff
@@ -218,6 +219,9 @@ export const buildRootObject = (vatPowers, vatParameters) => {
       assert.typeof(name, 'string');
       return consume[name];
     },
+    // ??? any more dangerous than produceItem/consumeItem?
```
turadg, line 222: @warner WDYT of using `getPromiseSpace()` instead of `produceItem` and `consumeItem`? Okay if I remove the latter?

dckc, line 222: perhaps slightly; not remarkably, though

packages/zoe/src/zoeService/types.js
```diff
@@ -181,14 +181,40 @@
  */
 
 /**
+ * Zoe uses seats to access or manipulate offers. They let contracts and users
+ * interact with them. Zoe has two kinds of seats. ZCFSeats are used within
```
dckc, line 185: Which of They/them refers to seats vs. offers?

Where did this text come from? Ah... `@link`. cool.

dckc (approved):
whew! yes, bite size PRs like this are _much_ easier to review.

PR #6123, opened 2022-09-03
fix: separate additional commits to #6107

erights:
#6107 has a bunch of recent commits that should be reviewed separately from #6107 itself. Perhaps they were meant as comments or suggestions rather than as commits? Awaiting clarification.


packages/store/src/index.js
```diff
@@ -55,7 +55,14 @@ export {
...
   fit,
 } from './patterns/patternMatchers.js';
-export { defendPrototype, initEmpty } from './patterns/interface-tools.js';
+
+export {
+  defendPrototype,
+  initEmpty,
+  defineHeapFarClass,
+  defineHeapFarClassKit,
+} from './patterns/interface-tools.js';
```
erights, line 64: This is a good change. I'm adopting it back into #6107

packages/store/src/patterns/interface-tools.js
```diff
@@ -4,10 +4,23 @@ import { listDifference, objectMap } from '@agoric/internal';
 
 import { fit } from './patternMatchers.js';
 
+/* Destructuring the object assert into two variables, details and quote. */
 const { details: X, quote: q } = assert;
+/* Using the Reflect API to apply a function to an object. */
 const { apply, ownKeys } = Reflect;
+/* Using object destructuring to assign the properties of the object Object to the variables
+defineProperties, seal, and freeze. */
 const { defineProperties, seal, freeze } = Object;
```
erights: This kind of commentary is appropriate in a tutorial to walk through code and explain each line. It is not appropriate for inclusion in the production code itself. Once oriented in reading the code itself, the comments simply restate what the code clearly says without any additional insight, and so only add noise from the pov of the production programmer.

That said, everything these comments say is correct, so you're not misunderstanding things.

packages/store/src/patterns/interface-tools.js
```diff
@@ -4,10 +4,23 @@ import { listDifference, objectMap } from '@agoric/internal';
...
 
+/**
+ * It takes a list of
+ * arguments, a method guard, and a label, and it checks that the arguments
+ * satisfy the method guard
+ * @param args - the arguments to the method
+ * @param methodGuard - The guard for the method being called.
+ * @param label - a string that will be used in error messages
+ * @returns The function defendSyncArgs is being returned.
```
erights, line 22: These have two problems:
   * The same noise problem: The comments just restate what the code already says. Great for an initial tutorial perhaps, but it detracts from readability by production programmers.
   * The lack of declared types. We use this jsdoc syntax to embed TypeScript type declarations into our JavaScript. I don't know what the effect on our typechecking is of including these `@param` and `@returns` without the TS types within curlies.

This also applies repeatedly below. OTOH, some of those below are worth keeping. We can separate these when we discuss.

That said, once again these comments are correct, so you're not misunderstanding things.

packages/store/src/patterns/interface-tools.js
```diff
@@ -44,6 +57,14 @@ const defendSyncArgs = (args, methodGuard, label) => {
   }
 };
 
+/**
+ * It takes a method, a guard, and a label, and returns a new method that is guarded by the guard
+ * @param method - The method to be guarded.
+ * @param methodGuard - a guard object that describes the parameters and return value of the method
+ * @param label - A string that will be used in error messages.
+ * @returns A function that takes a method and a methodGuard and returns a function that takes a label
+ * and returns a function that takes args and returns a result.
```
erights, line 66: I don't understand this one.

packages/store/test/test-heap-classes.js
```diff
@@ -8,28 +8,37 @@ import {
 } from '../src/patterns/interface-tools.js';
 import { M } from '../src/patterns/patternMatchers.js';
 
+/* Defining an interface for the UpCounter class. */
```
erights, line 11: Treating some test files as tutorials is much more appropriate. These in particular are good choices. I'm inclined to keep a lot of your commentary here for this reason. This is indeed example code making use of API that is new to all of us, and so even for us can be used as tutorial examples.

packages/vat-data/test/test-vivify.js
```diff
@@ -109,11 +117,15 @@ test('test vivifyFarInstance', t => {
...
     message: 'In "incr" method of (upCounter) arg 0: -3 - Must be >= 0',
   });
+  /* Testing that the function throws an error when the argument is not a number. */
   t.throws(() => upCounter.incr('foo'), {
     message:
       'In "incr" method of (upCounter) arg 0: string "foo" - Must be a number',
   });
 });
+
+test.todo('Demonstrate that the baggage can be used to revive objects.');
```
erights, line 131: Good change. Adopting that back into #6107

packages/vat-data/test/test-durable-classes.js
```diff
@@ -84,12 +97,16 @@ test('test defineDurableFarClassKit', t => {
...
     message:
       'In "decr" method of (Counter down) arg 0: string "foo" - Must be a number',
   });
+  /* Testing that the upCounter object does not have a decr method. */
   t.throws(() => upCounter.decr(3), {
     message: 'upCounter.decr is not a function',
   });
 });
+
+test.todo('Add tests for durability');
```
erights, line 112: This is a good change. Adopting it back into #6107

PR #6124, opened 2022-09-03
feat: charge fee based on bundle size

### LINKED ISSUE #6097: Prevent DDOS via large bundles

Sending too many large bundles will blow up the kernel DB.

## Description of the Design

Charge a fee for bundles based on size.  See `CalculateMessageFee()` in `golang/cosmos/x/swingset/types/msgs.go`, analogous to the fee for `Wallet{Spend}Action`.

There is already a governance parameter specifying the fee.

JimLarson:
closes: #6097

## Description

Prevents DDOS via submission of large bundles by charging a per-byte fee.

### Security Considerations

Fee needs to be sufficient to discourage DDOS while still not being an undue burden for legitimate use.

### DIFF
--- a/golang/cosmos/x/swingset/types/msgs.go
+++ b/golang/cosmos/x/swingset/types/msgs.go
@@ -19,6 +19,10 @@ var (
 	_ sdk.Msg = &MsgWalletSpendAction{}
 
 	_ vm.ControllerAdmissionMsg = &MsgDeliverInbound{}
+	_ vm.ControllerAdmissionMsg = &MsgInstallBundle{}
+	_ vm.ControllerAdmissionMsg = &MsgWalletAction{}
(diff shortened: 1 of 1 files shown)


kriskowal (approved):
If I read this correctly, the cost (in beans) for a submitting a bundle takes into account:

- fixed overhead
- overhead per message
- length of each message in bytes, but `chargeAdmission` is called in such a way that only one bundle message is selected.

I don’t know anything about beans or how to price them, but the algorithm looks right to me.

There’s at least one map-lookups under `chargeAdmission` that result in default zeroes. That might be intentional and good, but it might be better to err out if there’s no entry.

JimLarson:
Regarding the default zeroes, I'm inclined to interpret that as "working as intended". And this PR doesn't change that behavior.

PR #6130, opened 2022-09-04
WIP experiment to distinguish method, kind methods, and objects

erights:
Staged on https://github.com/Agoric/agoric-sdk/pull/6129 which is supposed to be https://github.com/Agoric/agoric-sdk/pull/6119

THIS PR IS NOT INTENDED TO EVER BE A CANDIDATE FOR MERGING.

An old Far object, the kind made with the `Kind` constructor, can serve double duty: It can serve as an object with methods to be invoked, and it serve as a record of methods to compose with other such records using `...`.

Both kinds and far classes move to an inheritance-based implementation in which this pun no longer works. Instead, https://github.com/Agoric/agoric-sdk/pull/6119 is experimenting with a style where non-instance-specific methods are composed together into the set of methods to give to the kind/farClass definer. In trying to review https://github.com/Agoric/agoric-sdk/pull/6119 I tried to figure out when it was doing which. This is not a question I know how to ask the type system, so in this PR I'm doing a name-based (aka "Hungarian notation") manual type check, just to trace this through for my understanding. Along the way, I see cases where `Far` is used to hold a record of methods that is *only* a record of methods. These should use `harden` rather than `Far`. I'm fixing these as I see them.

I don't intend to follow this through to completion; but rather to adequate understanding to review https://github.com/Agoric/agoric-sdk/pull/6119

### DIFF
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -2,8 +2,7 @@
 
 import { Far } from '@endo/marshal';
 import { makeStoredPublisherKit } from '@agoric/notifier';
-import { getMethodNames, objectMap } from '@agoric/internal';
-import { ignoreContext } from '@agoric/vat-data';
+import { getMethodNames } from '@agoric/internal';
 import { keyEQ, M } from '@agoric/store';
 import { AmountShape, BrandShape } from '@agoric/ertp';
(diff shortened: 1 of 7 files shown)


erights:
Closing as experiment that went off the rails in too many ways to be informative yet.

Issue #6127, opened 2022-09-04
support new smart wallet in Wallet UI


turadg:
## What is the Problem Being Solved?

https://github.com/Agoric/agoric-sdk/pull/6084 changes the API for smart wallet. We need the Wallet UI to migrate to using it.

## Description of the Design

Requirements
- one purse per brand (until https://github.com/Agoric/agoric-sdk/issues/6126 )
 - UI should disable aspects for >1

Non-requirements
- supporting Smart Wallet and solo wallet at once
  - solo wallet will no longer have a UI; only smart wallet will

Ideas
- see whether only https://github.com/Agoric/agoric-sdk/blob/master/packages/wallet/ui/src/components/SmartWalletConnection.jsx needs changing
- explore using `coalesceUpdates` utility from the new wallet and a Flux-like "something changed" update pattern to propagate

## Security Considerations

--

## Test Plan

--

Tartuffo:
Closed by #6134

turadg:
@Tartuffo we left this one open until @rowgraus verifies it meets Product acceptance criteria. If you want to track that separately okay.

Tartuffo:
> @Tartuffo we left this one open until @rowgraus verifies it meets Product acceptance criteria. If you want to track that separately okay.

Let's track testing separately.  There should be a user story issue for this @rowgraus

Issue #6126, opened 2022-09-03
Support multiple purses per brand


turadg:
## What is the Problem Being Solved?

The solo wallet supported multiple purses per brand. The [new smart wallet](https://github.com/Agoric/agoric-sdk/pull/6084) does not.

This isn't needed for PSM but it's not clear yet whether it's needed for MN-1. I'm filing this under New Issues with the MN-1 release so it gets attention.

## Description of the Design

Support pet names per purse in the contract and the UI.

Allow specifying which purse to use in offers. If there's only one then don't require it to be specified.

PR #6134, opened 2022-09-04
Wallet UI working with new Smart Wallet contract

turadg:
closes (almost): #6127 (leave open until Product verifies)

Stacked on,
- https://github.com/Agoric/agoric-sdk/pull/6084

## Description

Wallet UI has a WalletBackendAdapter for how the UI queries RPC nodes for data. That messaging system has changed with the new smart wallet and at least the adapter needs to be revised to match it.

Further changes may be warranted or necessary.


packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,71 @@ export const makeWalletBridgeFromFollower = (
...
+            brandPetname: descriptor.petname,
+            issuerPetname: descriptor.petname,
+            displayInfo: descriptor.displayInfo,
+          };
+          brandToPurse.set(descriptor.brand, purseObj);
+          updatePurses();
+          break;
+        }
+        case 'balance': {
+          // FIXME: We assume just one purse per brand.
```
turadg: more of a TODO since it's not a requirement yet https://github.com/Agoric/agoric-sdk/issues/6126

packages/casting/src/iterable.js
```diff
@@ -98,3 +98,23 @@ export const iterateEach = (follower, options) =>
       });
     },
   });
+
+/**
+ * TODO: Remove this function when we have `makePublisherKit`.
```
turadg: we have `makePublishKit` now

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -295,6 +294,9 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
      * @param {string} iface
      */
     fromMyWallet: (slot, iface) => {
+      if (!slot) {
+        return makePresence('dummy');
```
samsiegart, line 298: Is this an expected case, or should we assert here or something?

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -97,6 +100,7 @@ const sortedEntries = entries =>
   Object.entries(entries).sort(([kwa], [kwb]) => cmp(kwa, kwb));
 
 const Proposal = ({ offer, purses }) => {
+  console.log('offer', offer);
```
samsiegart: Nit: remove before merging

packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -15,6 +15,9 @@ const OfferEntryFromTemplate = (
 ) => {
   const value = BigInt(stringifiedValue);
   const purse = purses.find(p => p.pursePetname === pursePetname);
+  if (!purse) {
```
samsiegart, line 18: We might want to just throw here. I think there's an error boundary around it

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,82 @@ export const makeWalletBridgeFromFollower = (
...
+  const updatePurses = () => {
+    const purses = [];
+    for (const [brand, purse] of brandToPurse.entries()) {
+      if (purse.currentAmount && purse.brandPetname) {
+        console.log(purse.currentAmount);
+        pursePetnameToBrand.set(purse.pursePetname, brand);
+        purses.push(purse);
+      }
+    }
+    // console.log(purses);
```
samsiegart: Nit: remove

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,82 @@ export const makeWalletBridgeFromFollower = (
...
+    /** @type {number} */
+    let firstHeight;
+    for await (const { blockHeight } of iterateReverse(follower)) {
+      // TODO: Only set firstHeight and break if the value contains all our state.
+      firstHeight = blockHeight;
+    }
+    for await (const { value } of iterateEach(follower, {
+      height: firstHeight,
+    })) {
+      console.log(value);
```
samsiegart: Ni: remove

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,82 @@ export const makeWalletBridgeFromFollower = (
...
+    // console.log(purses);
+    notifierKits.purses.updater.updateState(harden(purses));
+  };
+
   const followLatest = async () => {
-    for await (const { value: state } of iterateLatest(follower)) {
+    /** @type {number} */
+    let firstHeight;
+    for await (const { blockHeight } of iterateReverse(follower)) {
+      // TODO: Only set firstHeight and break if the value contains all our state.
```
samsiegart, line 153: How do we tell when we get to a snapshot block?

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,82 @@ export const makeWalletBridgeFromFollower = (
...
+      firstHeight = blockHeight;
+    }
+    for await (const { value } of iterateEach(follower, {
```
samsiegart, line 156: We probably don't want to update the UI until we "catch up" to the current block, but I'm not sure how we can determine that. Otherwise, we might see for instance the purse balances cycle through several different amounts while we iterate towards the latest block.

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -184,6 +251,57 @@ export const makeWalletBridgeFromFollower = (
...
+          kw,
+          {
+            brand: brand || pursePetnameToBrand.get(pursePetname),
+            value: BigInt(value),
+          },
+        ]),
+      );
+    const offer = {
+      id: new Date().getTime(),
+      invitationSpec: {
```
samsiegart, line 273: This is what I was thinking of earlier, the dapp should probably be sending `invitationSpec` as-is instead of having to pull out `method` and `instance` from the details separately. Only the AMM and PSM use those, they were added just for smart wallet.

packages/wallet/ui/src/util/WalletBackendAdapter.js
michaelfig: They don't exist yet, which is why we only have this comment to describe what the code should eventually do.

packages/wallet/ui/src/util/WalletBackendAdapter.js
michaelfig: Let's consider that as a future API cleanup, but I'm not confident in executing that without making deeper changces to the dapp UI logic, with which I'm not really familiar.

packages/wallet/ui/src/util/WalletBackendAdapter.js
michaelfig: I see your point, but don't see how to solve it easily right now.

packages/casting/src/iterable.js
michaelfig: Copy-pasta.

packages/wallet/api/src/marshal-contexts.js
michaelfig: Expected case when the other marshaller returns `null` slots like `E(board).getReadonlyMarshaller()`

packages/wallet/ui/src/components/Proposal.jsx
michaelfig: Then we'd get a flash of an error until the purse resolved, which looked bad in my testing.

packages/casting/src/iterable.js
```diff
@@ -98,3 +98,23 @@ export const iterateEach = (follower, options) =>
...
+ *
+ * @template T
+ * @param {ERef<import('./types.js').Follower<T>>} follower
+ * @param {import('./types.js').IterateEachOptions} [options]
+ */
+export const iterateReverse = (follower, options) =>
```
dckc, line 107: @turadg take note; this looks pretty handy

dckc (commented):
I looked at it by commit; the code looks fine, but I haven't tested that it works.

samsiegart (approved):
LGTM!

packages/cosmic-swingset/Makefile
```diff
@@ -122,6 +122,7 @@ scenario2-setup-nobuild:
 	$(AGCH) --home=t1/bootstrap keys add bootstrap --keyring-backend=test
 	$(AGCH) --home=t1/bootstrap keys show -a bootstrap --keyring-backend=test > t1/bootstrap-address
 	$(AGCH) --home=t1/n0 add-genesis-account `cat t1/bootstrap-address` $(BOOT_COINS)
+	# Create
```
turadg, line 125: ???

packages/cosmic-swingset/Makefile
```diff
@@ -211,7 +212,11 @@ t1-provision-one-with-powers: wait-for-cosmos
 	  $(AGCH) --home=t1/bootstrap tx swingset provision-one --keyring-backend=test --from=bootstrap \
 		  --gas=auto --gas-adjustment=$(GAS_ADJUSTMENT) --broadcast-mode=block --yes --chain-id=$(CHAIN_ID) \
 		  t1/$(BASE_PORT) $$addr $(AGORIC_POWERS) -ojson | tee /dev/stderr | grep -q '"code":0'; }
-	
+
+# Send some USDC to the vbank/provision module account where it can be traded for IST.
+fund-provision-pool: wait-for-cosmos
```
turadg, line 217: 👍 ty

idea: make `fund-acct` depend on this having run. E.g. make this have a side-effect of noting its completion and have `fund-acct` depend on that, so that it runs if necesssary..

packages/smart-wallet/src/utils.js
```diff
@@ -9,7 +9,7 @@ import { observeIteration, subscribeEach } from '@agoric/notifier';
  * If this proves to be a problem we can add an option to this or a related
  * utility to reset state from RPC.
  *
- * @param {ERef<StoredSubscriber<import('./smartWallet').UpdateRecord>>} updates
+ * @param {ERef<Subscriber<import('./smartWallet').UpdateRecord>>} updates
```
turadg, line 12: 👍  this doesn't care about storage

packages/smart-wallet/test/devices.js
```diff
@@ -1,6 +1,6 @@
 import bundleCentralSupply from '@agoric/vats/bundles/bundle-centralSupply.js';
 import bundleMintHolder from '@agoric/vats/bundles/bundle-mintHolder.js';
-import bundleWalletFactory from '@agoric/vats/bundles/bundle-legacy-walletFactory.js';
+import bundleWalletFactory from '@agoric/vats/bundles/bundle-walletFactory.js';
```
turadg, line 3: We have a `yarn build` for this package now that builds `../bundles/bundle-walletFactory.js`.

Consider using that instead, so while making changes to this package you don't have to build all vats.

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -236,7 +235,7 @@ const defaultMakePresence = iface => {
 };
 
 /**
- * Make context for marshalling wallet or board data. To be imported into the client, which never makes objects.
```
turadg, line 239: why lose the comment explaining its purpose? Was it incorrect?

packages/wallet/api/src/marshal-contexts.js
turadg, line 298: worth a constant with a doc explaining why this is necessary.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -62,7 +62,7 @@ const ConnectionSettingsDialog = ({
   const classes = useStyles();
   const smartConnectionHrefs = allConnectionConfigs.map(({ href }) => href);
 
-  const [config, setConfig] = useState(connectionConfig);
+  const [config, setConfig] = useState(connectionConfig || {});
```
turadg, line 65: better to express defaults in the params

packages/wallet/ui/src/components/Proposal.jsx
turadg: The capitalization makes `OfferEntryFromTemplate` look like a component, and then it's surprising for a component to decide not to render itself.

consider naming this `makeOfferEntryForTemplate`. The `make` to show that it's not a component itself and `For` to acknowledge that there may not be one for the template.

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,80 @@ export const makeWalletBridgeFromFollower = (
     ]),
   );
 
+  // We assume just one cosmos purse per brand.
```
turadg, line 133: could ref https://github.com/Agoric/agoric-sdk/issues/6126 (though I now see it is below)

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,80 @@ export const makeWalletBridgeFromFollower = (
...
-      });
+      switch (updateRecord.updated) {
+        case 'brand': {
+          const { descriptor } = updateRecord;
+          const purseObj = {
+            ...brandToPurse.get(descriptor.brand),
+            brand: descriptor.brand,
+            brandPetname: descriptor.petname,
+            pursePetname: descriptor.petname,
+            displayInfo: descriptor.displayInfo,
```
turadg, line 176: the intent might be more clear with destructring,
```suggestion
            ...descriptor
```

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -130,15 +130,80 @@ export const makeWalletBridgeFromFollower = (
...
+          break;
+        }
+        case 'balance': {
+          // TODO: Don't assume just one purse per brand.
+          // https://github.com/Agoric/agoric-sdk/issues/6126
+          const { currentAmount } = updateRecord;
+          const purseObj = {
+            ...brandToPurse.get(currentAmount.brand),
+            currentAmount,
+            value: currentAmount.value,
```
turadg, line 189: odd to have `value` right next to an amount that has the same value. please remove the redundant one.

packages/smart-wallet/test/devices.js
dckc: I lean toward: you can assume `yarn build` is up-to-date for _other_ packages, but changes within a package should get bundled at test-time within a package. So this one is a little awkward: it depends on `yarn build` in `vats` whenever a change to the walletFactory contract in _this_ package is changed.

hm. not sure I have a concrete suggestion.

turadg: by that principle shouldn't this be as suggested?
```suggestion
import bundleWalletFactory from '../bundles/bundle-walletFactory.js';
```

dckc: yes, that's good, provided the `test.before()` auto-updates that bundle.

packages/cosmic-swingset/Makefile
turadg: copied to https://github.com/Agoric/agoric-sdk/pull/6197/commits/8fa8326e783125f8c14348d4a211bfa1ce1afc31

packages/wallet/api/src/marshal-contexts.js
michaelfig: I lost this in a merge conflict.  Reinstating.

packages/wallet/ui/src/util/WalletBackendAdapter.js
dckc: I suspect it's an external design constraint; that is: removing the redundant one would involve something like upgrading all the dapps (or: the half of the dapps that use the one or the other).

packages/wallet/api/src/marshal-contexts.js
michaelfig: Doc done!

packages/wallet/ui/src/util/WalletBackendAdapter.js
michaelfig: I found both were necessary.

packages/wallet/ui/src/util/WalletBackendAdapter.js
michaelfig: It's a remapping of properties, so I found following the current style was clearest for me.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
michaelfig: The `connectionConfig` may be falsy but not `undefined` (such as when it's `null`).  I found it was necessary to handle all falsy values.

PR #6135, opened 2022-09-05
fix: record-of-methods only for static composition should not be Far

erights:
Failed exercise #6130 did identify some `Far` declarations around a record of methods that were *only* used to be composed with other methods to form an overall behavior. `Far` is appropriate for defining a behavioral object with methods that are callable on that object.

Staged on #6119 because that PR uses compositions between such collection of methods for "static" behavior composition, i.e., where the composition is shared by all the instances dynamically created from that composition.

Prior to kinds and far classes, `Far` objects supported a pun --- a Far object is both an object with callable methods and a record of methods can be be extracted and composed with other to form other objects. With kinds and Far classes, the methods are no longer instance specific, and are no longer own properties of a meaningful behavioral object anyway. Records of methods used only for such static composition should stabilized with `harden` rather than `Far`.

This PR can be taken as suggestions for #6119, to be merged into #6119 if approved early. Or it can be taken as an independent PR to be merged into master after #6119 is.

### DIFF
--- a/packages/governance/src/contractHelper.js
+++ b/packages/governance/src/contractHelper.js
@@ -101,7 +101,7 @@ const facetHelpers = (zcf, paramManager) => {
    * @returns {LimitedCreatorFacet<CF>}
    */
   const makeLimitedCreatorFacet = originalCreatorFacet => {
-    return Far('governedContract creator facet', {
+    return harden({
       ...originalCreatorFacet,
       getContractGovernor: () => electionManager,
     });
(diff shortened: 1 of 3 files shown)


erights:
The exposed methods of a far object must be callable on the far object, and must be defensive. The methods written by the programmer and fed to something that makes far objects then fall into three different cases:
   * `Far` - The methods written by the programmer are the exposed methods themselves, and so must already be callable on the far object, and defensive.
   * Kinds (such as `defineDurableKind`, `vivifyKind`, etc) - The methods written by the programmer are shifted by one argument position from the methods exposed on the far object. The methods written by the programmer have an additional first `context` argument that is not supplied by the caller. The programmer-written kind methods must still be defensive by themselves.
   * Far classes (such as `vivifyFarClass`, etc) - the methods written by the programmer are not defensive by themselves. By themselves they are fragile, and so should never be exposed directly to potential callers. Instead, the programmer-written far class methods can assume the first line of defense provided by their InterfaceGuard. They are only responsible for the residual defensiveness needed after argument have passed the InterfaceGuard's checks. The programmer written methods also get their context from their `this` binding. The defensive methods exposed on the surface of the far object expects its `this` binding to be the far object itself, i.e., an instance of that far class. It then uses that `this` binding to look up the `context` to pass as the `this` binding of the programmer-written fragile method.

These three categories cannot be mixed. Generally, any programmer-written method is only ever in one of these categories.

PR #6136, opened 2022-09-05
fix(SwingSet): liveslots and supervisor errors to console.warn

mhofman:
## Description

A recent durable error in CI was hard to diagnose because liveslots errors which are the only one containing the local vat's stack trace were being printed to `stdout` and the [test in question](https://github.com/Agoric/agoric-sdk/blob/246a0ce1ffd862ef610ceeaf6343a0de2613e204/packages/cosmic-swingset/test/test-make.js#L30) discards stdout.

This PR changes the liveslots and supervisor treatment of such console logging to use the `warn` level instead, which ends up in `stderr`.

Drive by update of a couple similar cases.

### Security Considerations

None

None

### Testing Considerations

Visually confirmed locally that `yarn test` with this changes output all the necessary stacks while keeping the overall output quiet-ish.

### DIFF
--- a/packages/SwingSet/src/liveslots/liveslots.js
+++ b/packages/SwingSet/src/liveslots/liveslots.js
@@ -587,9 +587,9 @@ function build(
     // See https://github.com/Agoric/agoric-sdk/issues/2780
     errorIdNum: 70000,
     marshalSaveError: err =>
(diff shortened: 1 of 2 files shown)


erights (approved):
LGTM, thanks!

PR #6140, opened 2022-09-06
test(store): misc tests for record with optional fields

dckc:
## Description

This resulted from a discussion of how to write a pattern for `{ foo: number, bar?: number, baz?: number}`.

I don't know how much value they contribute; I don't mind if we skip this PR.

cc @cboydstun FYI... tests are often a very good way to figure out how stuff works. I don't know whether patterns fit into your schedule any time soon, but I hope we get to go over them before too long.


erights (approved):
LGTM, thanks!

Issue #6067, opened 2022-08-26
provide starter IST to newly paid accounts from a pool


dckc:
## What is the Problem Being Solved?

Alice would like to trade on Agoric; she has BLD; she executes a "please provision an account" transaction (as in #4654) which charges her 10 BLD or so. But admission fees are charged in **IST** per byte of submitted message (#5817). Alice doesn't have any IST. Now what?

## Description of the Design

In response to "please provision", for the price of 10 BLD or so, we not only provision a smart wallet account; we also provide enough IST to get started trading -- at least enough to buy more IST. Perhaps 0.25, 0.50, or 1.00 IST.

Where does this IST come from? A cosmos module account. Benefactors (such as DCF, ...) deposit funds into the account using cosmos `tx bank transfer`. The cosmos module account is exposed to the JavaScript side via VBANK to a provisioning contract. The provisioning contract handles "please provision" requests, creates smart wallets, and pays out starter IST.

(split out from #4654)

## Security Considerations

The provisioning contract is _not_ granted the capability to mint IST.

## Test Plan

TBD.

PR #6141, opened 2022-09-06
fix(vats): avoid importing sim-behaviors into boot-psm.js

dckc:
refs: #6009 

## Description

`boot-psm.js` was importing `core-proposal.js` just to get `ECON_COMMITTEE_MANIFEST`. Unfortunately, that dragged in `sim-behaviors.js` including a couple things that are permitted to have `feeMintAccess`.

Following the precedent of `startPSM.js`, this factors out `startEconCommittee.js`.

### Security Considerations

Clarifies that `fundAMM` and `connectFaucet`, which get `feeMintAccess` in demo/sim mode, are not in the production PSM configuration.

### Testing Considerations

I did some integration testing while working on #6067 . We'll see what the CI bots say...

cc @dtribble


dckc:
subsumed in some other PR

Issue #6086, opened 2022-08-30
Document how voters can currently safely validate a governance proposal to instantiate a contract


kriskowal:
## What is the Problem Being Solved?

We currently lack documentation for the workflow for proposing and validating governance proposals that would, for example, upgrade a contract.

## Description of the Design

By way of a preliminary sketch:

To propose a new contract installation, one would need to
- construct a bundle using `@endo/bundle-source`’s `bundle-source` command
- use `agd swingset install-bundle` to send the source to the chain.
- The bundle contains a hash, which can be extracted via `jq -r .endoZipBase64Sha512 bundle.json`.
- The governance proposal must contain a bootstrap script that arranges for E(zoe).install({ endoZipBase64Sha512 })` to be called and instructions for how to find the sources for that hash and *safely* verify the integrity of the sources and its third-party dependencies and reproducing the bundle with the same hash.

To verify a governance proposal that includes a E(zoe).install({ endoZipBase64Sha512 }) command in a bootstrap script:

- Skeptically follow the instructions in the governance proposal, which should be of the form:
- In a bomb-proof Docker container from which there is no escape,
- Download the bundle from the location specified in the governance proposal,
- Use a [yet-to-be-written bundle auditing tool](https://github.com/endojs/endo/issues/1656), which would produce the hash and verify the internal integrity of the bundle, then extract the original sources into the file system for manual inspection. The extracted bundle is not executable.

and/or verify the bundle is consistent with the sources of the contract from a repository and hash:

- Download the sources and their dependencies. These should be captured somewhere in a zip file and auditors should not be instructed to use `yarn` or `npm`, since these provide an opportunity for the attacker to run arbitrary code.
- Install the Agoric SDK
- Use `@endo/bundle-source`’s `bundle-source` CLI to generate a bundle.
- Extract and compare the hashes of the generated bundle.
- Proceed to review the sources only if the hashes match.

## Security Considerations

Validating a contract currently requires obtaining the sources at a particular hash, installing their dependencies, reconstructing the bundle, matching the generated hash, then inspecting the contract proper for malfeasance. Improperly installing the dependencies of an arbitrary application can empower an attacker to run arbitrary code with the authority of the user.

dckc:
a 1-off attempt: [game1-bundles.mk](https://gist.github.com/dckc/d4d299818234903da405dce972536444#file-game1-bundles-mk)

https://devnet.agoric.explorers.guru/proposal/9 on agoricdev-20

[...]

PR #6147, opened 2022-09-06
feat(agoric-cli): Add agoric publish subcommand

### LINKED ISSUE #6087: Create `agoric publish` command

For PS0 launch, we can’t rely on `deploy` scripts since they depend on `ag-solo`. The [steps for proposing upgrading a contract](https://github.com/Agoric/agoric-sdk/issues/6086) would be considerably simplified with commands that will create and publish bundles.

## Description of the Design

- `yarn bundle-source --to-json bundle-cache/ contract/src/contract.js contract` would generate a JSON bundle that is suitable for verifying the resulting hash or publishing the resulting artifact in `bundle-cache/bundle-contract.json`.
- `agoric publish bundle.zip` would queue an SwingSet InstallBundle transaction and watch for the casting of the success or failure. That will produce a JSON hashed bundle suitable for passing to `zoe~.install` in a governance script.

kriskowal:
closes: #6087

## Description

This is an `agoric publish` command that can post directly to the chain and return a hash bundle JSON blob, suitable for passing to `zoe~.install`.

This will be useful for documenting governance https://github.com/Agoric/agoric-sdk/issues/6086

### Testing Considerations

Works on my machine. To make work on your machine:

In cosmic-swingset:

```
make scenario2-run-chain
```

In another cosmic-swingset, use an ag-solo to provision an account. You may then turn off the ag-solo: it is not needed, and later commands must not depend on it.

```
make scenario2-run-client
```

In `dapp-fungible-faucet`, prepare a bundle.json:

```
yarn bundle-source --to bundles contract/src/contract.js contract
```

By whatever string-carving you prefer, convert the generated `bundles/bundle-contract.js` to `bundles/bundle-contract.json`. A small Node script that imports the JS and writes out JSON will suffice, or just removing the `export default ` prefix and `;` suffix in an editor will work too.

Continue:

```
agoric publish bundles/bundle-contract.json --node localhost:26657 --home ~/agoric-sdk/packages/cosmic-swingset/t1/8000/ 2>/dev/null
```

You can verify the publish with agoric follow: (the flags just provide block height published, block height observed, and ensure one line per publish).

```
agoric follow :bundles -bco jsonlines
```

You can also confirm with agd:

```
watch agd query vstorage data bundles
```

### DIFF
--- a/packages/agoric-cli/src/json.js
+++ b/packages/agoric-cli/src/json.js
@@ -1,0 +1,19 @@
+// @ts-check
+
+/**
+ * Parses JSON and, if necessary, throws exceptions that include the location
+ * of the offending file.
+ *
+ * @param {string} source
+ * @param {string} location
+ */
+export const parseLocatedJson = (source, location) => {
+  try {
+    return JSON.parse(source);
+  } catch (error) {
+    if (error instanceof SyntaxError) {
+      throw new SyntaxError(`Cannot parse JSON from ${location}, ${error}`);
+    }
+    throw error;
+  }
+};
--- a/packages/agoric-cli/src/main-publish.js
+++ b/packages/agoric-cli/src/main-publish.js
@@ -1,0 +1,64 @@
+/* global process */
+// @ts-check
+
+import path from 'path';
+
+import { SigningStargateClient } from '@cosmjs/stargate';
+
+import { parseLocatedJson } from './json.js';
+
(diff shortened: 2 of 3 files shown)


Chris-Hibbert:
Ack.  I defer to other reviewers.  Glad to see progress here.

dckc (approved):
LGTM!

PR #6151, opened 2022-09-07
fix(xsnap): Fix timestamps

mhofman:
Use xsnap-pub https://github.com/agoric-labs/xsnap-pub/pull/31 with fixed and more syscall timestamps, as well as less copies of buffers.

Remove a copy on the `xsnap.js` side too.

Update description of where time is spent.


dckc (approved):
very nice.

PR #6153, opened 2022-09-07
6062 wallet patterns

### LINKED ISSUE #6062: Utilize method schemas in walletFactory

turadg:
closes: #6062

## Description

Add type guards to smart-wallet package for external inputs.

Also add a feature to MarshalContext to support the tests.

### Security Considerations

More boundary checking

--

### Testing Considerations

--

### DIFF
--- a/packages/smart-wallet/src/invitations.js
+++ b/packages/smart-wallet/src/invitations.js
@@ -1,6 +1,8 @@
 // @ts-check
 import { AmountMath } from '@agoric/ertp';
+import { fit } from '@agoric/store';
 import { E } from '@endo/far';
+import { shape } from './typeGuards.js';
 
 /**
  * Supports three cases
@@ -52,12 +54,16 @@ export const makeInvitationsHelper = (
   const invitationGetters = /** @type {const} */ ({
     /** @type {(spec: ContractInvitationSpec) => Promise<Invitation>} */
     contract(spec) {
+      fit(spec, shape.ContractInvitationSpec);
+
       const { instance, publicInvitationMaker, invitationArgs = [] } = spec;
       const pf = E(zoe).getPublicFacet(instance);
       return E(pf)[publicInvitationMaker](...invitationArgs);
     },
     /** @type {(spec: PurseInvitationSpec) => Promise<Invitation>} */
     async purse(spec) {
+      fit(spec, shape.PurseInvitationSpec);
+
       const { instance, description } = spec;
       assert(instance && description, 'missing instance or description');
       /** @type {Amount<'set'>} */
@@ -74,7 +80,8 @@ export const makeInvitationsHelper = (
     },
     /** @type {(spec: ContinuingInvitationSpec) => Promise<Invitation>} */
     continuing(spec) {
-      console.log('making continuing invitation', spec);
+      fit(spec, shape.ContinuingInvitationSpec);
+
       const { previousOffer, invitationArgs = [], invitationMakerName } = spec;
       const makers = getInvitationContinuation(previousOffer);
       assert(
--- a/packages/smart-wallet/src/smartWallet.js
+++ b/packages/smart-wallet/src/smartWallet.js
@@ -1,15 +1,17 @@
 // @ts-check
+import { PaymentShape } from '@agoric/ertp';
 import { isNat } from '@agoric/nat';
 import {
   makeStoredPublishKit,
   observeIteration,
   observeNotifier,
 } from '@agoric/notifier';
-import { makeScalarMapStore } from '@agoric/store';
+import { fit, M, makeScalarMapStore } from '@agoric/store';
 import { makeScalarBigMapStore } from '@agoric/vat-data';
 import { E, Far } from '@endo/far';
 import { makeInvitationsHelper } from './invitations.js';
 import { makeOffersFacet } from './offers.js';
+import { shape } from './typeGuards.js';
 
 const { details: X, quote: q } = assert;
 
@@ -255,6 +257,8 @@ export const makeSmartWallet = async (
      * This version does not: 1) for expedience, 2: to avoid resource exhaustion vulnerability.
      */
     receive: async (paymentE, paymentBrand) => {
+      fit(harden(paymentE), M.eref(PaymentShape));
+
       const brand = await (paymentBrand || E(paymentE).getAllegedBrand());
       const purse = brandPurses.get(brand);
 
@@ -295,12 +299,14 @@ export const makeSmartWallet = async (
 
   /**
    *
-   * @param {import('./types').WalletCapData<BridgeAction>} actionCapData
+   * @param {import('@endo/captp').CapData<string>} actionCapData of type BridgeAction
    * @param {boolean} [canSpend=false]
    */
   const handleBridgeAction = (actionCapData, canSpend = false) => {
-    assert(actionCapData.body && actionCapData.slots, 'invalid capdata');
+    fit(actionCapData, shape.StringCapData);
+
     assert(!canSpend, 'spending not yet supported');
+
     return E.when(
       E(marshaller).unserialize(actionCapData),
       /** @param {BridgeAction} action */
--- a/packages/smart-wallet/src/typeGuards.js
+++ b/packages/smart-wallet/src/typeGuards.js
@@ -1,0 +1,65 @@
+// @ts-check
+
+import { M } from '@agoric/store';
+import {
+  InstanceHandleShape,
+  ProposalShape,
+} from '@agoric/zoe/src/typeGuards.js';
+
+export const shape = {
+  // smartWallet
+  StringCapData: {
+    body: M.string(),
+    slots: M.arrayOf(M.string()),
+  },
+
+  // invitations
+  ContractInvitationSpec: M.split(
+    {
+      source: 'contract',
+      instance: InstanceHandleShape,
+      publicInvitationMaker: M.string(),
+    },
+    M.partial({
+      invitationArgs: M.array(),
+    }),
+  ),
+  ContinuingInvitationSpec: M.split(
+    {
+      source: 'continuing',
+      previousOffer: M.number(),
+      invitationMakerName: M.string(),
+    },
+    M.partial({
+      invitationArgs: M.array(),
+    }),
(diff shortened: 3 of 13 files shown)


packages/smart-wallet/src/typeGuards.js
```diff
@@ -0,0 +1,66 @@
...
+      instance: InstanceHandleShape,
+      publicInvitationMaker: M.string(),
+    },
+    M.partial({
+      invitationArgs: M.array(),
+    }),
+  ),
+  ContinuingInvitationSpec: M.split(
+    {
+      source: M.string(), // 'continuing'
```
turadg: @erights I didn't see a way in the patterns lib to express a literal. I saw some "limits" feature on some matchers like string but it's all numerical.

WDYT of `M.literal(val)` ?

erights:
Try just the literal value itself

packages/smart-wallet/src/typeGuards.js
dckc: I think `M.literal(val)` is expressed as just `val`.
_p.s. I see I'm late to the party._

packages/smart-wallet/src/typeGuards.js
```diff
@@ -0,0 +1,66 @@
+// @ts-check
+
+import { M } from '@agoric/store';
+import {
+  InstanceHandleShape,
+  ProposalShape,
+} from '@agoric/zoe/src/typeGuards.js';
+
+export const shape = {
```
dtribble, line 9: In all the other typeGuards, the shapes are at top level so that they are imported. Seems like it should be consistent. Is there a motivation fore burying them?

packages/smart-wallet/src/invitations.js
```diff
@@ -52,12 +54,16 @@ export const makeInvitationsHelper = (
...
     contract(spec) {
+      fit(harden(spec), shape.ContractInvitationSpec);
+
       const { instance, publicInvitationMaker, invitationArgs = [] } = spec;
       const pf = E(zoe).getPublicFacet(instance);
       return E(pf)[publicInvitationMaker](...invitationArgs);
     },
     /** @type {(spec: PurseInvitationSpec) => Promise<Invitation>} */
     async purse(spec) {
+      fit(harden(spec), shape.PurseInvitationSpec);
```
dtribble: Why is the harden here instead of required of the argument?  It woudl not have been passable without that, and the guard will check that.

packages/smart-wallet/src/smartWallet.js
```diff
@@ -255,6 +257,8 @@ export const makeSmartWallet = async (
      * This version does not: 1) for expedience, 2: to avoid resource exhaustion vulnerability.
      */
     receive: async (paymentE, paymentBrand) => {
+      fit(harden(paymentE), M.eref(PaymentShape));
```
dtribble, line 260: What is the interface over which these are being received?  should that just be a FarClass? Not blocking, but might be simpler.

packages/smart-wallet/src/typeGuards.js
```diff
@@ -0,0 +1,66 @@
...
+    description: M.string(),
+  },
+
+  // offers
+  OfferShape: M.split(
+    {
+      id: M.number(),
+      // TODO M.unknown() to defer validation
+      invitationSpec: M.any(),
+      // XXX ProposalShape requires all whereas TS types ProposalRecord does and Proposal is partial
```
dtribble: Can delete the XXX? This perfectly reasonable.

dtribble: Unless you want to recommend renaming the current `ProposalShape` to `FullProposalShape` and publishing the below as `ProposalShape`?

packages/smart-wallet/src/typeGuards.js
```diff
@@ -0,0 +1,66 @@
...
+      proposal: M.partial(ProposalShape),
+    },
+    M.partial({ offerArgs: M.recordOf(M.string(), M.any()) }),
```
dtribble: Only `offerArgs` is allowed in the type, so should this be `M.opt({ offerArgs: M.recordOf(M.string(), M.any()) })`?

dtribble: Otherwise this permits additional properties that are not in the offerSpec type

dtribble: can `offerArgs` be present with an `undefined` value? That might require a slightly different guard

packages/smart-wallet/src/walletFactory.js
```diff
@@ -6,10 +6,19 @@
...
 import { makeAtomicProvider } from '@agoric/store/src/stores/store-utils.js';
 import { makeScalarBigMapStore } from '@agoric/vat-data';
 import { E, Far } from '@endo/far';
 import { makeSmartWallet } from './smartWallet.js';
+import { shape } from './typeGuards.js';
+
+const PrivateArgsShape = harden(
+  M.split(
+    { storageNode: M.eref(M.any()) },
+    M.partial({ bridgeManager: M.eref(M.any()) }),
```
dtribble, line 19: Should be `M.opt`, except it may also allow the property to be present but `undefined`. As is, this allows any additional properties in the shape.

packages/smart-wallet/src/walletFactory.js
```diff
@@ -50,19 +60,17 @@ export const start = async (zcf, privateArgs) => {
      */
     fromBridge: async (srcID, obj) => {
       console.log('walletFactory.fromBridge:', srcID, obj);
-      assert(obj, 'missing wallet action');
-      assert.typeof(obj, 'object');
-      assert.typeof(obj.owner, 'string');
+      fit(harden(obj), shape.WalletBridgeMsg);
       const canSpend = 'spendAction' in obj;
       assert(
```
dtribble: this assert should be subsumed in the shape check above?

packages/wallet/api/src/marshal-contexts.js
```diff
@@ -324,6 +324,24 @@ export const makeImportContext = (makePresence = defaultMakePresence) => {
...
+     * @param {unknown} val
+     */
+    ensureBoardId: (id, val) => {
+      if (boardObjects.byVal.has(val)) {
+        assert.equal(boardObjects.byVal.get(val), id);
+        return;
+      }
+      initSlotVal(boardObjects, id, val);
+    },
     fromMyWallet: Far('wallet marshaller', { ...marshal.fromMyWallet }),
```
dtribble, line 345: This things ends up with both methods and properties?  Either that or I'm confused about what gets wrapped with `far` here.

dtribble (changes requested):
Generally looks great. There's a couple things worth changing and one part I didn't understand that I need to iterate with you on

packages/smart-wallet/src/typeGuards.js
turadg: yes, a few reasons.

1. separate the semantic part from the naming part. The `Shape` aspect is semantic.
2.  to have the same string for the TS type name for eyeballing that it's the same thing. This way if there's a type `FooThing` you validate it with `shape.FooThing`.
3.  It also means that if you want to rename that shape, you're more likely to catch it in a global replace. If it were `FooThingShape` you might only rename the shapes and miss the types. They should be in sync.

packages/smart-wallet/src/invitations.js
turadg: sgtm, will remove

packages/smart-wallet/src/smartWallet.js
turadg: it will be. that's the next PR.

packages/smart-wallet/src/typeGuards.js
turadg: right I propose a rename. that particular rename SGTM. sounds like it has your support so I'll do that in this PR.

packages/smart-wallet/src/typeGuards.js
turadg: I think you may be missing the `M.split`. The desired type is https://github.com/Agoric/agoric-sdk/blob/70a070446ed63e9fdd7b4c911f76c871c3dbee8e/packages/smart-wallet/src/offers.js#L8-L15

So the matching shape would be,
```
  OfferSpec: M.split(
    // these three are required
    {
      id: M.number(),
      // TODO M.unknown() to defer validation
      invitationSpec: M.any(),
      // XXX ProposalShape requires all whereas TS types ProposalRecord does and Proposal is partial
      proposal: M.partial(ProposalShape),
    },
    // because of the split we need a partial
    M.partial({ offerArgs: 
     // should not be optional because if there is a key then it should be a defined value
      M.recordOf(M.string(), M.any())
  }),
  ),
```

But… I think the TS was too strict and `offerArgs` is really `any`. So I'll change that in both.

packages/wallet/api/src/marshal-contexts.js
turadg: yeah this is mixing, which is okay because the object returned isn't remote

packages/smart-wallet/src/typeGuards.js
```diff
@@ -0,0 +1,65 @@
...
+
+export const shape = {
+  // smartWallet
+  StringCapData: {
+    body: M.string(),
+    slots: M.arrayOf(M.string()),
+  },
+
+  // invitations
+  ContractInvitationSpec: M.split(
```
dtribble, line 17: Separately it seems that we have "Spec" as another genericish thing in the space of class/record/interface/shape/whatever. That ought to drop out of our naming or get clear guidance of when to use it in types. (post launch?)

turadg: the distinction here is meaningful. the "Spec" is a specification for producing the thing and not the thing itself. it comes up in the wallet code because we send messages to describe things to do.

e.g. `ContractInvitationSpec` says that to get the invitation, use the `instance` and call `publicInvitationMaker` with `invitationArgs` arguments. 

I'm open to changing the name but  I don't see it as a blocker for PS0 and definitely not this PR about implementing argument validation. If you like I can make a ticket to audit the smart-wallet package coding style.

dtribble (approved):
LGTM

PR #6155, opened 2022-09-07
fix(smart-wallet): invitation issuer is remote

Chris-Hibbert:
refs: #6084

## Description

Dan and I were trying to make progress on voting based on top of #6084, and ran into two issues, but the PR merged before we could suggestion this correction.

missing E() around an invitationIssuer

(I originally included updating decentral-psm-config to use smartwallet, but that's maybe not for master).

### Security Considerations

None

None

### Testing Considerations

None

### DIFF
--- a/packages/smart-wallet/src/walletFactory.js
+++ b/packages/smart-wallet/src/walletFactory.js
@@ -78,7 +78,7 @@ export const start = async (zcf, privateArgs) => {
 
   // Each wallet has `zoe` it can use to look them up, but pass these in to save that work.
   const invitationIssuer = await E(zoe).getInvitationIssuer();
-  const invitationBrand = await invitationIssuer.getBrand();
+  const invitationBrand = await E(invitationIssuer).getBrand();
 
   const shared = {
     agoricNames,


dtribble (approved):
LGTM

turadg (commented):
Change is okay but I think before merge we should find a way to prevent this error in the future. I can do that if you prefer.

Looks like this type is wrong,
https://github.com/Agoric/agoric-sdk/blob/b120bbbe7755b13e678ebba457cd0fb2d73112ca/packages/zoe/src/zoeService/types.js#L54-L58

Should be `FarRef<Issuer>`?

Is there any way to have this fail in Ava without running a full swingset test?

Chris-Hibbert:
> Change is okay but I think before merge we should find a way to prevent this error in the future. I can do that if you prefer.

It's your PR. Feel free to drop the `automerge:` label.

Chris-Hibbert:
> Change is okay but I think before merge we should find a way to prevent this error in the future. I can do that if you prefer.

I removed `automerge:`.  would you prefer to make a new PR?

turadg:
@Chris-Hibbert fixing the type was a rabbit hole. I think this is good to merge since there's no Ava test and the TS will get fixed over time.

I added back the automerge and made it "squash" to the PR title, which I've updated.

turadg (approved):

PR #6158, opened 2022-09-08
fix: Make pattern matching faster

erights:
Except for slight differences in the text of error messages, and timing of course, this PR should have no observable effect on code outside the module.

Some comparative figures from part way through development. ops/sec, so higher is better.

On the left was our pattern matching speed before we added limits. In the middle is after we added limits, causing a slowdown we found alarming. On the right is from a slightly earlier version of this PR. My current number are now slightly better than this.

As a performance oriented PR, I should note that we are not trying to do ambitious optimization. That day will come. This one is just trying to pick up some low hanging fruit in order to be good enough for now.

image

### DIFF
--- a/packages/store/src/patterns/patternMatchers.js
+++ b/packages/store/src/patterns/patternMatchers.js
@@ -157,16 +157,18 @@ const makePatternKit = () => {
   /**
    * @typedef {string} Kind
    * It is either a PassStyle other than 'tagged', or, if the underlying
-   * PassStyle is 'tagged', then the `getTag` value.
-   *
-   * We cannot further restrict this to only possible passStyles
-   * or known tags, because we wish to allow matching of tags that we
-   * don't know ahead of time. Do we need to separate the namespaces?
-   * TODO are we asking for trouble by lumping passStyles and tags
-   * together into kinds?
+   * PassStyle is 'tagged', then the `getTag` value for tags that are
+   * recognized at the store level of abstraction. For each of those
+   * tags, a tagged record only has that kind if it satisfies the invariants
+   * that the store level associates with that kind.
    */
 
-  const kindMemo = new WeakMap();
+  /**
+   * @type {WeakMap<CopyTagged, Kind>}
+   * Only for tagged records of recognized kinds whose store-level invariants
+   * have already been checked.
+   */
+  const tagMemo = new WeakMap();
 
   /**
    * Checks only recognized tags, and only if the tagged
@@ -217,13 +219,15 @@ const makePatternKit = () => {
     if (passStyle !== 'tagged') {
       return passStyle;
     }
-    // At this point we know that specimen is a well formed tagged record
-    if (kindMemo.has(specimen)) {
-      return kindMemo.get(specimen);
+    // At this point we know that specimen is a well formed
+    // as a tagged record, which is defined at the marshal level of abstraction,
+    // since `passStyleOf` checks those invariants.
+    if (tagMemo.has(specimen)) {
+      return tagMemo.get(specimen);
     }
     const tag = getTag(specimen);
     if (checkTagged(specimen, tag, check)) {
-      kindMemo.set(specimen, tag);
+      tagMemo.set(specimen, tag);
       return tag;
     }
     if (check !== identChecker) {
@@ -253,6 +257,22 @@ const makePatternKit = () => {
       check(false, details);
     }
     return false;
+  };
+
+  /**
+   * @param {Passable} specimen
+   * @param {Key} keyAsPattern
+   * @param {Checker} check
+   * @return {boolean}
+   */
+  const checkAsKeyPatt = (specimen, keyAsPattern, check) => {
+    if (keyEQ(specimen, keyAsPattern)) {
+      return true;
+    }
+    return (
+      check !== identChecker &&
+      check(false, X`${specimen} - Must be: ${keyAsPattern}`)
+    );
   };
 
   // /////////////////////// isPattern /////////////////////////////////////////
@@ -457,16 +477,13 @@ const makePatternKit = () => {
       case 'copyBag':
       case 'remotable': {
         // These kinds are necessarily keys
-        return check(keyEQ(specimen, patt), X`${specimen} - Must be: ${patt}`);
+        return checkAsKeyPatt(specimen, patt, check);
       }
       case 'copyArray': {
         if (isKey(patt)) {
           // Takes care of patterns which are keys, so the rest of this
           // logic can assume patterns that are not keys.
-          return check(
-            keyEQ(specimen, patt),
-            X`${specimen} - Must be: ${patt}`,
-          );
+          return checkAsKeyPatt(specimen, patt, check);
         }
         if (specimenKind !== 'copyArray') {
           return check(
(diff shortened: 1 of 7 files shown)


erights:
This PR is now ready for review!

packages/store/package.json
```diff
@@ -38,6 +38,8 @@
   },
   "devDependencies": {
     "@agoric/swingset-vat": "^0.28.0",
+    "@agoric/ertp": "^0.14.2",
```
dtribble: I don't think taht having th perf trst here is worth confusing the dependency graph. I propose that this package should copy the relevant shapes from elsewhere, since they are just intended to be repersentative.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -151,6 +154,107 @@ const makePatternKit = () => {
...
+   * @param {Passable} specimen
+   * @param {Checker} [check]
+   * @returns {Kind | undefined}
+   */
+  const kindOf = (specimen, check = identChecker) => {
+    const passStyle = passStyleOf(specimen);
+    if (passStyle !== 'tagged') {
+      return passStyle;
+    }
+    // At this point we know that specimen is a well formed tagged record
```
dtribble: how do we know it's well-formed?

packages/store/package.json
dtribble: I think this module should not take a dependency on zoe or ertp. Instead, copy the shapes, since they just need to be representative rather than accurate.

packages/store/test/perf-patterns.js
```diff
@@ -0,0 +1,259 @@
...
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { makeTagged } from '@endo/marshal';
+import {
+  AmountMath,
+  makeIssuerKit,
+  AmountShape,
+  BrandShape,
+  PaymentShape,
+} from '@agoric/ertp';
```
dtribble: copies these shapes into the test os we don't complicate layering.

packages/store/test/perf-patterns.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+  const totalTime = Date.now() - startAt;
+  const perStep = totalTime / times;
+  measurements.push([title, perStep, Math.round(1000 / perStep)]);
+  console.log(
+    `test '${title}': ${times} repetitions took ${totalTime}ms;  each: ${perStep}ms`,
+  );
+};
+
+const issuerKit = makeIssuerKit('fungible');
```
dtribble: using `Far` is sufficient to make things that match the various issuer patterns for measurement purposes. So you don't need this dependency

packages/store/src/patterns/patternMatchers.js
```diff
@@ -181,8 +289,11 @@ const makePatternKit = () => {
     // essentially identical.
     const checkIt = child => checkPattern(child, check);
```
dtribble, line 310: this should be moved to lin 305 so it's only allocated if used.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -354,17 +433,42 @@ const makePatternKit = () => {
...
+      case 'boolean':
+      case 'number':
+      case 'bigint':
+      case 'string':
+      case 'symbol':
+      case 'copySet':
+      case 'copyBag':
+      case 'remotable': {
+        // These kinds are necessarily keys
+        return check(keyEQ(specimen, patt), X`${specimen} - Must be: ${patt}`);
```
dtribble: since it's all the leaeves, put this assertion in an `if` statement?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -380,7 +484,15 @@ const makePatternKit = () => {
...
-        if (specimenStyle !== 'copyRecord') {
+        if (isKey(patt)) {
+          // Takes care of patterns which are keys, so the rest of this
+          // logic can assume patterns that are not keys.
+          return check(
+            keyEQ(specimen, patt),
+            X`${specimen} - Must be: ${patt}`,
+          );
+        }
+        if (specimenKind !== 'copyRecord') {
```
dtribble, line 509: It still seems to me from a "Reduce redundant code" that this check be once at the top, assuming that `specimenStyle` and `patternStyle` must be `===`

packages/store/src/patterns/patternMatchers.js
```diff
@@ -478,7 +575,12 @@ const makePatternKit = () => {
    * @param {string|number} [label]
    */
   const fit = (specimen, patt, label = undefined) => {
+    if (checkMatches(specimen, patt, identChecker, label)) {
+      return;
+    }
+    // should only throw
     checkMatches(specimen, patt, assertChecker, label);
+    assert.fail(X`internal: ${label}: inconsistent pattern match: ${q(patt)}`);
```
dtribble, line 594: I like it.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -44,6 +43,9 @@ const patternMemo = new WeakSet();
 
 // /////////////////////// Match Helpers Helpers /////////////////////////////
 
+/** For forward references to `M` */
+let MM;
```
dtribble, line 47: There's really no better way to do this with some `const [M, matchRemotableHelper,...] = [....`

dtribble (commented):
The primary thing that I think must be addressed is the module layering issue.

packages/store/package.json
erights: Good idea. Done.

erights: Done.

packages/store/test/perf-patterns.js
erights: Done

packages/store/test/perf-patterns.js
erights: Done

packages/store/src/patterns/patternMatchers.js
erights: ;)

packages/store/src/patterns/patternMatchers.js
erights: expanded comment:
```js
    // At this point we know that specimen is a well formed
    // as a tagged record, which is defined at the marshal level of abstraction,
    // since `passStyleOf` checks those invariants.
```

packages/store/src/patterns/patternMatchers.js
erights: Can't. It's used in two cases of the `switch` so needs to happen before the switch, or redundantly in each branch of the switch. If expensive. duplicating for each branch would be worth it. But this is only allocating the closure.

Resolving.

packages/store/src/patterns/patternMatchers.js
erights: Excellent suggestion. Resulted in a small but consistently measurable speedup. Doubled the `empty` ops/sec! (Which is exactly the case we'd expect is most sensitive to this.)

Done.

packages/store/src/patterns/patternMatchers.js
erights: Can't. Only applies to three of the cases. Resolving.

packages/store/src/patterns/patternMatchers.js
erights: Ran into annoying problems trying to do this. IMO not worth delaying the merge for this. I am proceeding.

dtribble (approved):
LGTM

PR #6161, opened 2022-09-08
fix(vats): Mark bootstrap vats as critical

### LINKED ISSUE #6051: mark all core vats as "critical"

Our initial chain will consist of about 20 vats: bootstrap, a handful of static vats (some created by the kernel itself), a bunch of dynamic vats created by bootstrap, and then a bunch of contract vats created by zoe (either during bootstap or during an "enable the economy" governance event shortly afterwards).

At this stage, if any of these vats fail, we should panic the kernel and halt the chain. It would be better to stop the chain than to move forward without any of these vats.

## Description of the Design

For static vats, we have [a flag for this](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/docs/configuration.md): you just set `config.vats.NAME.creationOptions.critical = true`. So the task is to modify `packages/vats/*-config.json` to add this flag to the static vats defined therein.

For dynamic vats, the party doing the creation (either bootstrap or zoe) must provide a special `criticalVatKey` object in an option to the `createVat` call. This object can only be obtained from the vat-admin *root object* (which is distinct from the `vatAdminService` on which one calls `createVat`):

```js
// bootstrap method in vat-bootstrap
function bootstrap(vats, devices) {
  const critical = await E(vats.vatAdmin).getCriticalVatKey();
  // then, later:
  E(vatAdminService).createVat(bundlecap, { critical }).then(..);
}
```

So the tasks are:
* 1: bootstrap should fetch `getCriticalVatKey()` from the vat-admin root object and `produce` it to a slot named `criticalVatKey`
* 2: bootstrap's `makeVat` code should `consume` that slot and use it as the `critical:` option in its `createVat()` calls, so all the initial dynamic vats (like zoe) will be marked as critical
* 3: I think bootstrap is the one that builds a `createZcfVat` for Zoe, so bootstrap needs to imbue that function with `criticalVatKey`, so all contract vats that Zoe creates will be marked as critical

After our initial launch, zoe should certainly *not* be marking all contract vats as critical. For one, that would allow any third-party contract to halt the chain. But also we expect to be creating more contract instances as the chain matures, and many of them will be short-lived or will serve a smaller audience, where their termination is expected and will be more tolerated.

## Security Considerations

Without this change, certain unexpected failure modes (vats consume [more than 2 GiB of RAM](https://github.com/Agoric/agoric-sdk/issues/5953), emit [oversized](https://github.com/Agoric/agoric-sdk/issues/4280) netstring messages, or suffer some internal consistency error that manifests as an illegal syscall) will terminate the vat and then commit (within consensus) the deletion of the vat's state. This will make recovery more difficult, as we'll have to roll back the state by a block or two, instead of distributing new software that only has to fix the terminating behavior.

## Test Plan

It'd be nice to somehow trigger a vat failure and make sure the chain halts, but it might be good enough to just inspect the config files and slogfiles once (looking for the `isCritical` option in the `vatOptions`).

gibson042:
closes: #6051

## Description
* Update bootstrap vats to launch as critical.
* Define a new `loadCriticalVat` variant of `loadVat`.
* Update current dynamic vats to be launched with the new variant.

### Security Considerations

`loadCriticalVat` is inherently more powerful than `loadVat`, but that power is managed by being a distinct deniable value produced by makeVatsFromBundles.

https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/docs/dynamic-vats.md mentions `createVat` but not `loadVat`, has it gone stale?

### Testing Considerations

Help wanted!

### DIFF
--- a/packages/vats/src/core/basic-behaviors.js
+++ b/packages/vats/src/core/basic-behaviors.js
@@ -101,18 +101,18 @@ harden(makeVatsFromBundles);
 
 /**
  * @param { BootstrapPowers & {
- *   consume: { loadVat: ERef<VatLoader<ZoeVat>> }
+ *   consume: { loadCriticalVat: ERef<VatLoader<ZoeVat>> }
  * }} powers
  *
  * @typedef {ERef<ReturnType<import('../vat-zoe.js').buildRootObject>>} ZoeVat
  */
 export const buildZoe = async ({
-  consume: { vatAdminSvc, loadVat, client },
+  consume: { vatAdminSvc, loadCriticalVat, client },
   produce: { zoe, feeMintAccess },
 }) => {
   const zcfBundleName = 'zcf'; // should match config.bundles.zcf=
   const { zoeService, feeMintAccess: fma } = await E(
-    E(loadVat)('zoe'),
+    E(loadCriticalVat)('zoe'),
   ).buildZoe(vatAdminSvc, feeIssuerConfig, zcfBundleName);
 
   zoe.resolve(zoeService);
@@ -126,16 +126,16 @@ harden(buildZoe);
 
 /**
  * @param {BootstrapPowers & {
- *   consume: { loadVat: ERef<VatLoader<PriceAuthorityVat>>},
+ *   consume: { loadCriticalVat: ERef<VatLoader<PriceAuthorityVat>>},
  * }} powers
  *
  * @typedef {ERef<ReturnType<import('../vat-priceAuthority.js').buildRootObject>>} PriceAuthorityVat
  */
 export const startPriceAuthority = async ({
-  consume: { loadVat, client },
+  consume: { loadCriticalVat, client },
   produce,
 }) => {
-  const vats = { priceAuthority: E(loadVat)('priceAuthority') };
+  const vats = { priceAuthority: E(loadCriticalVat)('priceAuthority') };
   const { priceAuthority, adminFacet } = await E(
     vats.priceAuthority,
   ).makePriceAuthorityRegistry();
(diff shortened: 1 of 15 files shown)


dckc:
> [dynamic-vats.md](https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/docs/dynamic-vats.md) mentions createVat but not loadVat, has it gone stale?

no; `loadVat` is specific to the bootstrap stuff in `vats`.

packages/vats/src/core/basic-behaviors.js
```diff
@@ -65,10 +65,10 @@ const bootMsgEx = {
  * @typedef {{adminNode: any, root: unknown}} VatInfo as from createVatByName
  * @typedef {MapStore<string, Promise<VatInfo>>} VatStore
  */
-export const makeVatsFromBundles = ({
+export const makeVatsFromBundles = async ({
   vats,
   devices,
-  produce: { vatAdminSvc, loadVat, vatStore },
+  produce: { vatAdminSvc, loadVat, loadCriticalVat, vatStore },
```
dckc, line 71: I think we have no use for `loadVat` any more; I'm a little concerned that it's like a rake, waiting for somebody to step on it. I suppose an audit of manifests is straightforward... though not trivial.

Did you consider `loadVat.resolve(newFangledLoadVat)`? That is: replacing `loadVat` rather than adding a 2nd tool? Oh... I suppose that would prevent core-eval from using `loadVat` in the future.

Never mind.

dckc (approved):
LGTM!

PR #6162, opened 2022-09-08
chore(smart-wallet): take out limitation against spending

dckc:
### Security Considerations

allows smart-wallet contract to withdraw funds from user purses


turadg (approved):
You can safely `bypass:integration` to get this in faster.

nit on commit message: this isn't a chore; it changes the runtime. consider `refactor`.

dckc:
The way I learned the terms, a refactor never changes observable behavior.
A chore might be part of a feature or bug fix. That's the way I see this one: it's a chore in service to the feature of executing offers.

PR #6163, opened 2022-09-08
feat(vats): upgrade to psm-first smart-wallet

dckc:
## Description

Change the `walletFactory` bundle to be built from packages/smart-wallet; that is: use the new contract in bootstrap.

### Security Considerations

smaller surface area for the contract
hardened with method schemas / patterns

### Testing Considerations

existing unit tests are quite good. This PR proposes that the next phase of testing be shared by the team.

### DIFF
--- a/packages/vats/decentral-psm-config.json
+++ b/packages/vats/decentral-psm-config.json
@@ -19,7 +19,7 @@
   },
   "bundles": {
     "walletFactory": {
-      "sourceSpec": "@agoric/wallet/contract/src/walletFactory.js"
+      "sourceSpec": "@agoric/smart-wallet/src/walletFactory.js"
     },
     "committee": {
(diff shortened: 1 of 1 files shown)


turadg:
This is going to break the UI so I'll defer to @michaelfig or @samsiegart to approve. (I figured this swap would happen in https://github.com/Agoric/agoric-sdk/pull/6134)

I also have some concerns about swapping out the old wallet without removing it as https://github.com/Agoric/agoric-sdk/pull/6134/commits/494c540cfaccf76b574d5277ae30bbb4063d3d0e does.

dckc:
Fine with me if the swap happens in #6134; I just figured I would hoist this to prompt a few folks to perhaps give some input.

samsiegart (approved):

Issue #6160, opened 2022-09-08
Generate TypeScript types from matcher patterns


turadg:
## What is the Problem Being Solved?

With the typeGuards we're repeating type descriptions that are in TS/JSdoc. This is extra work and ambiguous source of truth.

We have `TypedPattern` for type narrowing, but we've had bugs where the static and runtime types differ. (eg https://github.com/Agoric/agoric-sdk/pull/10787 ) 

## Description of the Design
When a type is described by matchers, that becomes the source of truth. TS types can be auto-generated from it.

The solution will live in Endo. This later issue tracks implementing it there,
- https://github.com/endojs/endo/issues/2392

Related work
- [sinclairzx81/typebox: JSON Schema Type Builder with Static Type Resolution for TypeScript](https://github.com/sinclairzx81/typebox)
- [monnef / oats](https://gitlab.com/monnef/oats) - "Oats takes an API description in a subset of OpenAPI format and generates schemas, types and endpoint methods or functions."
- [Using with TypeScript \| Ajv JSON schema validator](https://ajv.js.org/guide/typescript.html)

kriskowal:
I believe we agree the ideal outcome would be to solve this problem entirely with TypeScript annotations and not resort to code generation. With the ideal outcome, a program like the following would pass a type check.

```js
const pattern = pl.or(pl.number(), pl.string());
/** @type {{ value: number | string } | null} */
const matched = pattern.match(10);
if (matched !== null) {
  /** @type {number | string} */
  const value = matched.value;
}
```

Other acceptable outcomes might involve a callback.

However, if this is not possible, we could generalize pattern definitions such that a definition can be used either to generate a passable pattern object or the equivalent TypeScript definition text.

```js
/** @type {<T>(p: Pattern<T>) => T} */
const patterner = $ => $.or($.number(), $.string());
const pattern = patterner(toPattern);
```

And alternately elsewhere:

```js
console.log(patterner(toTypeScriptNotation, 'MyNumber'));
// output: number | string
```

cc @erights

mhofman:
[...]

PR #6165, opened 2022-09-08
4489 walletFactory virtual

### LINKED ISSUE #4489: walletFactory uses virtual objects

turadg:
closes: #4489 

## Description

Updates the smart wallet contract to be virtualized.

This is my first use of the new Far Class stuff. I cut some paper cuts and tried to sand the edges with [refactor(store): types for interface-tools](https://github.com/Agoric/agoric-sdk/pull/6165/commits/6a596e82be488f624599400b1f6791b9fe49ddd8) but stopped eventually. Many types in this PR have regressed to `any` but addressing that can wait to after launch.

### Security Considerations

More guards

--

### Testing Considerations

There's one test that started failing after I rebased, I think due to https://github.com/Agoric/agoric-sdk/pull/6166

- [ ] un-disable `"govern offerFilter"` test

### DIFF
--- a/packages/smart-wallet/src/smartWallet.js
+++ b/packages/smart-wallet/src/smartWallet.js
@@ -1,12 +1,12 @@
 // @ts-check
-import { PaymentShape } from '@agoric/ertp';
+import { AmountShape, PaymentShape } from '@agoric/ertp';
 import { isNat } from '@agoric/nat';
 import {
   makeStoredPublishKit,
   observeIteration,
   observeNotifier,
 } from '@agoric/notifier';
-import { fit, M, makeScalarMapStore } from '@agoric/store';
+import { fit, M, makeHeapFarInstance, makeScalarMapStore } from '@agoric/store';
 import { makeScalarBigMapStore } from '@agoric/vat-data';
 import { E, Far } from '@endo/far';
 import { makeInvitationsHelper } from './invitations.js';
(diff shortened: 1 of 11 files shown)


packages/smart-wallet/test/test-psm-integration.js
```diff
@@ -153,7 +154,8 @@ test('want stable', async t => {
   t.is(purseBalance(computedState, stableBrand), swapSize - 1n);
 });
 
-test('govern offerFilter', async t => {
+// FIXME this broke after https://github.com/Agoric/agoric-sdk/pull/6166
```
Chris-Hibbert: please create an issue and assign to me.  (Unless you want to fix it in this PR.)

packages/smart-wallet/src/smartWallet.js
```diff
@@ -60,161 +64,323 @@ const { details: X, quote: q } = assert;
...
+    // Private purses. This assumes one purse per brand, which will be valid in MN-1 but not always.
+    brandPurses: makeScalarBigMapStore('brand purses', { durable: true }),
   };
 
-  // #endregion
+  const nonpreciousState = {
+    brandDescriptors: makeScalarMapStore(),
+    // To ensure every offer ID is unique we require that each is a number greater
+    // than has ever been used. This high water mark is sufficient to track that.
+    lastOfferId: 0,
```
Chris-Hibbert, line 137: does lastOfferId need to survive upgrade?

packages/smart-wallet/src/smartWallet.js
```diff
@@ -60,161 +64,323 @@ const { details: X, quote: q } = assert;
...
+  offers: M.interface('offers facet', {
+    executeOffer: M.call(shape.OfferSpec).returns(M.promise()),
+    getLastOfferId: M.call().returns(M.number()),
+  }),
+  self: M.interface('selfFacetI', {
+    handleBridgeAction: M.call(shape.StringCapData, M.boolean()).returns(
+      M.promise(),
+    ),
+    getDepositFacet: M.call().returns(M.eref(M.any())),
+    getOffersFacet: M.call().returns(M.eref(M.any())),
```
Chris-Hibbert, line 176: These facets should be simple to describe.

packages/smart-wallet/src/smartWallet.js
```diff
@@ -60,161 +64,323 @@ const { details: X, quote: q } = assert;
...
+     * @param {import('@endo/captp').CapData<string>} actionCapData of type BridgeAction
+     * @param {boolean} [canSpend=false]
+     */
+    handleBridgeAction(actionCapData, canSpend = false) {
+      assert(!canSpend, 'spending not yet supported');
+
+      const { publicMarshaller } = this.state;
+      const { offers } = this.facets;
+
+      console.log('DEBUG executeOffer', offers.executeOffer);
```
Chris-Hibbert: Please remove before merging.

packages/smart-wallet/src/smartWallet.js
turadg: it's a SHOULD not a MUST. if it doesn't survive it's easy to determine from the `offerStatus` update in RPC. I'll make a note of that.

packages/smart-wallet/src/smartWallet.js
turadg: what's the value? these are input guards. why runtime validate the return?

seems like busy work, especially when we have to write types twice (until https://github.com/Agoric/agoric-sdk/issues/6160)

Chris-Hibbert: I'm worried that we'll do something like #6160, but have incomplete types on the returns because they're currently not enforced, so they'll all say `.returns(M.promise())` or `.returns(M.eref(M.any())`.

Not suggesting a change here, just registering my concern.

Chris-Hibbert (approved):
I approve on the basis that the conversion to `heapFarInstance`s looks clean.

I haven't looked at the smartWallet design enough to have confidence that the local rearrangements haven't done violence to it.  Is there anyone who understands wallet/smartWallet well enough to review?

turadg:
[79f3df2](https://github.com/Agoric/agoric-sdk/pull/6165/commits/79f3df25e1c98b9fca2b6e7990c8fb77ae8edce2) passed deployment test and the only change since then is fixing a TS lint error so I'm adding **bypass:integration** to get this in faster.

PR #6157, opened 2022-09-08
provide starter IST to newly provisoned accounts

dckc:
refs: #6067

## Description

 - add `provisionPool` contract; wire it up to the walletFactory as well as each PSM
 - add module account, `vbank/provision`, to hold IST as well as assets to sell on the/a PSM

TODO:
 - [ ] make initial amount governed (I prefer to do this in a follow-on PR)

### Security Considerations

In the contract: `makeBridgeProvisionTool` is factored out of the contract `start` function to keep the bridge handling stuff mostly separate from the asset-handling stuff.

Elsewhere, the use of bootstrap powers merits careful review.

### Documentation Considerations

We should provide end users with docs on the "pay 10 BLD to the community pool; get 0.25 IST for initial fees" model.

### Testing Considerations

Trading on the PSM is unit-tested.

The flow resulting in 0.25 IST going to the user was only tested manually. (Results are documented [below](https://github.com/Agoric/agoric-sdk/pull/6157#issuecomment-1241481631).)

PR #6172, opened 2022-09-09
feat(smartWallet): defer deposits until purse available

turadg:
supporting #6157 

## Description

Some brands aren't available when a payments to the deposit facet. In this case we now hold the payments in a queue and when the purse is available we deposit all of them.

The payouts handler now uses the deposit facet to get the same feature. It also attempts each payout separately instead of the previous behavior of blocking until all payments were available.

### Security Considerations

Deposits are unpermissioned so queuing them opens an attack to grow the queue unbounded. The collections are durable which should mitigate RAM consumption. If there are more problems in practice the queue size can be capped. I considered doing that in this PR but considered a worse vulnerability is for an attacker to fill the queue of a wallet and prevent legitimate deposits.

--
### Testing Considerations

--

### DIFF
--- a/packages/smart-wallet/src/payments.js
+++ b/packages/smart-wallet/src/payments.js
@@ -7,8 +7,9 @@ import { E } from '@endo/far';
  * Used in an offer execution to manage payments state safely.
  *
  * @param {(brand: Brand) => import('./types').RemotePurse} purseForBrand
+ * @param {{ receive: (payment: *) => Promise<Amount> }} depositFacet
  */
-export const makePaymentsHelper = purseForBrand => {
+export const makePaymentsHelper = (purseForBrand, depositFacet) => {
   /** @type {PaymentPKeywordRecord | null} */
   let keywordPaymentPromises = null;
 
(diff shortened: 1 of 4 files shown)


packages/smart-wallet/src/payments.js
```diff
@@ -74,20 +75,14 @@ export const makePaymentsHelper = purseForBrand => {
...
      */
     async depositPayouts(payouts) {
-      /** @type {PaymentKeywordRecord} */
-      // @ts-expect-error ???
-      const paymentKeywordRecord = await deeplyFulfilledObject(payouts);
       /** Record<string, Promise<Amount>> */
-      const amountPKeywordRecord = objectMap(paymentKeywordRecord, payment =>
-        E(purseForBrand(payment.getAllegedBrand())).deposit(payment),
+      const amountPKeywordRecord = objectMap(payouts, paymentRef =>
+        E.when(paymentRef, payment => depositFacet.receive(payment)),
```
dckc, line 85: :+1: good idea to re-use the depositFacet

packages/smart-wallet/src/smartWallet.js
```diff
@@ -248,27 +259,46 @@ const behavior = {
...
-     * NB: the previous smart wallet contract would try again each time there's a new issuer.
-     * This version does not: 1) for expedience, 2: to avoid resource exhaustion vulnerability.
+     * @returns {Promise<Amount>} amounts for deferred deposits will be empty
      */
     async receive(payment) {
       /** @type {State} */
-      const { brandPurses } = this.state;
+      const { brandPurses, paymentQueues: queues } = this.state;
       const brand = await E(payment).getAllegedBrand();
       const purse = brandPurses.get(brand);
```
dckc: This get will throw rather than returning undefined, no?

dckc (changes requested):
`get` throws on our collections, no?

(reusing the standard name with different behavior was... an odd choice, IMO)

packages/smart-wallet/src/smartWallet.js
turadg: yes good catch! I added a unit test for regression and it caught a missing harden too.

dckc (approved):
thanks for the unit test

PR #6173, opened 2022-09-10
fix(smart-wallet): missing E() around remote payment

dckc:
closes: #XXXX
refs: #XXXX


erights (approved):
Given a PR description, LGTM. 

(I'm confident of this change even in my overall ignorance of the smart wallet).

turadg:
Had been fixed in https://github.com/Agoric/agoric-sdk/pull/6172

dckc:
I must not have been caught up when I was testing.

PR #6174, opened 2022-09-10
fix: alt syntax for positive but faster assertions

erights:
Like https://github.com/endojs/endo/pull/1280

Replace some occurrences of
```js
assert(..condExpr.., X`..detailsString..`);
```
with
```js
..condExpr.. || assert.fail(X`..detailsString..`);
```

I only made the changes that I was easily able to do via regexp search and replace in vscode
```
^\s*assert\(\n\s*(.*?),\n\s*X`(.*?)`,\n\s*\);
($1) || assert.fail(X`$2`);
```
and similar, followed by a global `yarn lint-fix` and then repair until everything tested green again. No doubt there are more cases that I missed. But from @dtribble 's measurements, these are already enough to show significant perf improvements.

### DIFF
--- a/packages/ERTP/src/amountMath.js
+++ b/packages/ERTP/src/amountMath.js
@@ -25,11 +25,12 @@ const AssetKind = harden({
 const assetKindNames = harden(Object.values(AssetKind).sort());
 
 /** @type {AssertAssetKind} */
-const assertAssetKind = allegedAK =>
-  assert(
-    assetKindNames.includes(allegedAK),
-    X`The assetKind ${allegedAK} must be one of ${q(assetKindNames)}`,
-  );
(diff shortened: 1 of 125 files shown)


erights:
See also https://github.com/endojs/endo/pull/1280

erights:
Ready for review!

turadg (approved):
LGTM. I spot checked the mechanical commit and verified the TS work-around commits. Consider squashing those two (into the latter name) before landing.

turadg:
@Mergifyio requeue

dckc:
> from @dtribble 's measurements, these are already enough to show significant perf improvements.

Somebody share those measurements, please?

Issue #5976, opened 2022-08-17
Loadgen experienced suspicious XS `fatal!` errors


mhofman:
## Describe the bug

Sometime after https://github.com/Agoric/agoric-sdk/commit/00d565d7fa87c55f849825aed5c26b43cd9a82d4, the loadgen on the benchmark machine started experiencing SIGABRT caused by `xsMemory.c:303: fatal!` and `xsType.c:546: fatal!` errors. This would happen in the first block after a node restart. Unfortunately the exact commit where this started is unknown since another issue (https://github.com/Agoric/agoric-sdk/issues/5911) prevented to get that far on restarts.

It got seemingly resolved between https://github.com/Agoric/agoric-sdk/compare/e7e652986cb5410bc09152b8974d6c60cfbb0b28...914fc4c5a31f10207f99aa86e02e0d430b198fcb, but none of these are code changes. I suppose a dependency change in https://github.com/Agoric/agoric-sdk/pull/5895 may have had an impact?

## To Reproduce
Check loadgen logs

## Expected behavior
No XS fatal errors

## Additional context
```
0 4700232 daily-perf-914fc4c5a31f 2022-08-08 22:33:27.457786812 +0000 3
2 4338484 daily-perf-e7e652986cb5 2022-08-08 01:33:10.085955968 +0000 3
2 3292188 daily-perf-cf97ba310fb5 2022-08-07 06:35:13.590091804 +0000 3
2 486812 daily-perf-a679fa561bb6 2022-08-06 09:50:07.074302811 +0000 3 // xsMemory failure related SIGABRT
2 514808 daily-perf-aff48fd1b5c0 2022-08-06 07:26:17.261245675 +0000 3 // actionQueue.tail related SIGABRT
2 514356 daily-perf-f5cf38e954ee 2022-08-06 05:02:19.948927865 +0000 3
2 514800 daily-perf-74885306b09d 2022-08-06 02:38:41.157575496 +0000 3
2 512664 daily-perf-7f294a69eeae 2022-08-05 08:28:13.577604859 +0000 3 // actionQueue.tail related SIGABRT
0 4695976 daily-perf-00d565d7fa87 2022-08-05 06:03:56.213017691 +0000 236
```

`a679fa561bb6` logs:
```
stage-2: chain: 2022-08-06T09:48:23.698Z block-manager: block 570 begin
stage-2: runner: Chain ready
[...]

PR #6177, opened 2022-09-11
fix(SwingSet): transcript tools support snapshot save and load

mhofman:
refs: #5976

## Description

Insert snapshot save entry in `.sst` files generated by slog extraction
Support snapshot save and load entries when replaying `.sst` files.
The load entry has to be manually inserted after the corresponding save.

### Security Considerations

None, tools

Added some comments around `const`s tweaking the behavior of the replay tool.
IOU a snippet on how to stitch slog files across restart to extract a correct transcript.

### Testing Considerations

Used successfully to reproduce #5976 from slog


packages/SwingSet/misc-tools/replay-transcript.js
```diff
@@ -19,14 +26,65 @@ import { makeDummyMeterControl } from '../src/kernel/dummyMeterControl.js';
...
 import engineGC from '../src/lib-nodejs/engine-gc.js';
 
+// Set the absolute path of the SDK to use for bundling
+// This can help if there are symlinks in the path that should be respected
+// to match the path of the SDK that produced the initial transcript
+// For e.g. set to '/src' if replaying a docker based loadgen transcript
+const ABSOLUTE_SDK_PATH = null;
+
+// Rebuild the bundles when starting the replay.
+// Disable if bundles were previously extracted form a Kernel DB, or
```
warner, line 36: typo `form`

packages/SwingSet/misc-tools/replay-transcript.js
```diff
@@ -151,27 +248,48 @@ async function replay(transcriptFile) {
     }
     lineNumber += 1;
     const data = JSON.parse(line);
-    if (!manager) {
+    if (data.type === 'heap-snapshot-load') {
```
warner, line 251: Do we have anything that creates a `heap-snapshot-load` transcript entry?

For most purposes, I think we'd be replaying from the most recent snapshot (so we might `save` snapshots during the course of the replay, but we'd never `load` more than one). Our extract-from-slogfile tool might grow to let the operator select which span of the vat they're exercising, and if the slogfile only goes back to the most recent kernel restart, then that span might start from a snapshot instead of from the initial vat state). The extract-from-kerneldb tool could would know the most recent snapshot and transcript since then, but it would not know about `save`s (but that'd be fine, since by definition no saves happen during the transcript-suffix: if one did, the starting point would be moved to exclude it).

warner (approved):
looks good

packages/SwingSet/misc-tools/replay-transcript.js
mhofman: Not yet, but I wanted to expand the slog extraction tool to support concatenated or multiple slog files, or in general paging out and reloading a vat. Currently it chokes on any "replay" which come with snapshot load (for lack of syscall results in the slog), but we could paper over these if we've already seen the same deliveries.

PR #6179, opened 2022-09-11
feat(patterns): improve the measurement tools

### LINKED ISSUE #6098: Determine performance impact of enforcing Method Schemas
We need a perf measure comparing calling methods with and without input validation, to demonstrate that using the method schemas will not cause significant slowdown.

Ideally we'd do this w/o relying on end-to-end testing, because there is so much machinery involved it can be difficult to tell what the impact is.

dtribble:
closes: #6098

## Description
- Add support XS
- Build a single script that works for both node and XS
- Add platform-neutral support for forced GC
- add package scripts to build and run perf tasks

### Security Considerations

N/A

Byild sript instructions are in the perf-patterns file

### Testing Considerations

Currently only run manually with 
```
yarn perf:build
yarn perf:node
yarn perf:xs
```

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -14,7 +14,10 @@
     "lint-fix": "yarn lint:eslint --fix",
     "lint": "run-s --continue-on-error lint:*",
     "lint:types": "tsc --maxNodeModuleJsDepth 3 -p jsconfig.json",
-    "lint:eslint": "eslint ."
+    "lint:eslint": "eslint .",
+    "perf:build": "rollup --config test/rollup.config.js",
+    "perf:node": "NODE_OPTIONS='--jitless' node dist/bundle-perf-patterns.mjs",
+    "perf:xs": "../xsnap/moddable/build/bin/mac/release/xst -m dist/bundle-perf-patterns.mjs"
   },
   "repository": {
     "type": "git",
@@ -39,6 +42,7 @@
   "devDependencies": {
     "@agoric/swingset-vat": "^0.28.0",
     "@fast-check/ava": "^1.0.1",
+    "@rollup/plugin-node-resolve": "^6.1.0",
     "ava": "^4.3.1"
   },
   "files": [
--- a/packages/store/test/perf-patterns.js
+++ b/packages/store/test/perf-patterns.js
@@ -1,9 +1,10 @@
+/* global globalThis */
 // @ts-check
+import { setupGC } from './perf-setup.js';
 import '@agoric/swingset-vat/tools/prepare-test-env.js';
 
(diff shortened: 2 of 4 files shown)


packages/store/package.json
```diff
@@ -14,7 +14,10 @@
     "lint-fix": "yarn lint:eslint --fix",
     "lint": "run-s --continue-on-error lint:*",
     "lint:types": "tsc --maxNodeModuleJsDepth 3 -p jsconfig.json",
-    "lint:eslint": "eslint ."
+    "lint:eslint": "eslint .",
+    "perf:build": "rollup --config test/rollup.config.js",
+    "perf:node": "NODE_OPTIONS='--jitless' node dist/bundle-perf-patterns.mjs",
+    "perf:xs": "../xsnap/moddable/build/bin/mac/release/xst -m dist/bundle-perf-patterns.mjs"
```
mhofman, line 20: That is not portable. Also depends on `xst` having been built explicitly (it's not built as part of our regular build process). I think I'd rather have it reference an `xst` from the PATH and leave it to the reader on how to make that happen.

```suggestion
    "perf:xs": "xst -m dist/bundle-perf-patterns.mjs"
```

packages/store/package.json
```diff
@@ -58,4 +62,4 @@
     ],
     "timeout": "2m"
   }
-}
+}
```
mhofman, line 65: New line at end of file missing

mhofman (changes requested):

packages/store/package.json
dtribble: I'd like it to grab the XST for the version of moddable that we are using, rather than one installed in the enviroenment. That means it should use one built in the repo?

mhofman: We'd need to start by updating the build scripts to build xst, and then find a way to make it work across both Linux and Mac. I do believe it should be a separate PR, hence why I suggested removing any path here.

I doubt anyone has xst in their PATH.

erights:
What is the status of this? The bug it claims to fix, #6098 , was closed by @Tartuffo on Sep 14. Do we otherwise have the tooling to solve the problem that this PR was meant to solve? Can this PR be closed?

erights:
@dtribble We do need such measurement tools. Is this PR still relevant?  https://github.com/Agoric/agoric-sdk/pull/6179#issuecomment-1320657763

erights:
@mhofman @dtribble Given the more recent work on measurement, is there still value in this PR, or should it be closed?

turadg:
The issue this says it closes is already closed so I'm closing this.

PR #6180, opened 2022-09-12
feat(provision): use governance in provisioning

### LINKED ISSUE #6067: provide starter IST to newly paid accounts from a pool

dtribble:
closes: #6067

## Description

- [x] move initialAmount term to governance
- [x] add subscription for
  - provisionedCount: 1n,
  - totalMintedConverted
  - totalMintedProvided
- [x] connect provisionPool to psmCharter so econ committee can govern it
- drive by: propagate error from PSM bootstrap

The difference between totalMintedConverted and totalMintedProvided is the pool balance.

## Docs

added initial amount to inter-protocol/docs/governance-params.md

## Testing

I (@dckc) tested the metrics manually end-to-end.

I have not tested the ability to change the initial amount via governance; I expect we'll do that as we validate the rest of committee governance end-to-end.

### DIFF
--- a/packages/vats/src/core/startWalletFactory.js
+++ b/packages/vats/src/core/startWalletFactory.js
@@ -129,6 +129,7 @@ export const startWalletFactory = async (
     installation: {
       consume: { walletFactory, provisionPool, contractGovernor },
     },
+    instance: { produce: instanceProduce },
     brand: {
       consume: { [Stable.symbol]: feeBrandP },
     },
@@ -189,6 +190,7 @@ export const startWalletFactory = async (
     },
   );
   walletFactoryStartResult.resolve(wfFacets);
+  instanceProduce.walletFactory.resolve(wfFacets.instance);
   const poolBank = E(bankManager).getBankForAddress(poolAddr);
 
   const ppFacets = await startGovernedInstance(
@@ -215,6 +217,7 @@ export const startWalletFactory = async (
       economicCommitteeCreatorFacet,
     },
   );
+  instanceProduce.provisionPool.resolve(ppFacets.instance);
 
   provisionPoolStartResult.resolve(ppFacets);
 
@@ -270,6 +273,12 @@ export const WALLET_FACTORY_MANIFEST = {
     },
     issuer: {
       consume: { [Stable.symbol]: 'zoe' },
+    },
+    instance: {
+      produce: {
+        provisionPool: 'provisionPool',
+        walletFactory: 'walletFactory',
+      },
     },
   },
 };
--- a/packages/vats/src/core/startWalletFactory.js
+++ b/packages/vats/src/core/startWalletFactory.js
@@ -104,6 +104,7 @@ const startGovernedInstance = async (
  *
  * @param {BootstrapPowers & PromiseSpaceOf<{
  *   economicCommitteeCreatorFacet: CommitteeElectorateCreatorFacet
+ *   psmCharterCreatorFacet: Awaited<ReturnType<import('@agoric/inter-protocol/src/psm/psmCharter.js').start>>['creatorFacet'],
  * }>} powers
  * @param {{
  *   options?: {
@@ -124,6 +125,7 @@ export const startWalletFactory = async (
       zoe,
       chainTimerService,
       economicCommitteeCreatorFacet,
+      psmCharterCreatorFacet,
     },
     produce: { client, walletFactoryStartResult, provisionPoolStartResult },
     installation: {
@@ -194,7 +196,6 @@ export const startWalletFactory = async (
   const poolBank = E(bankManager).getBankForAddress(poolAddr);
 
   const ppFacets = await startGovernedInstance(
-    // zoe.startInstance() args
     {
       zoe,
       governedContractInstallation: provisionPool,
@@ -227,7 +228,16 @@ export const startWalletFactory = async (
     walletFactory: wfFacets.creatorFacet,
   });
 
-  await E(bridgeManager).register(BRIDGE_ID.PROVISION, handler);
+  await Promise.all([
+    E(bridgeManager).register(BRIDGE_ID.PROVISION, handler),
+    // psmCharter was designed to govern PSM contracts,
+    // but voteOnParamChanges should work just as well
+    // for provisionPool perAccountInitialValue
+    E(psmCharterCreatorFacet).addInstance(
+      ppFacets.instance,
+      ppFacets.creatorFacet,
+    ),
+  ]);
 
   client.resolve(
     Far('dummy client', {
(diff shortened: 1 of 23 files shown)


packages/vats/src/provisionPool.js
```diff
@@ -62,35 +82,69 @@ const makeBridgeProvisionTool = sendInitialPayment => {
...
     });
   return makeHandler;
 };
 
 /**
- * @typedef {StandardTerms & {
+ * @typedef {StandardTerms & GovernanceTerms<{
+ *    PerAccountInitialAmount: 'amount',
+ *   }> &{
  *   perAccountInitialAmount: Amount<'nat'>,
```
Chris-Hibbert, line 96:
```suggestion
 *    PerAccountInitialAmount: Amount<'nat'>,
```

are both perAccountInitialAmount and PerAccountInitialAmount expected here?

Chris-Hibbert (commented):
Looks like the right direction.

dckc:
### metrics smoke test: OK

```
$ agoric follow :published.provisionPool.metrics
{
  provisionedCount: 0n,
  totalMintedConverted: {
    brand: slot(0,"Alleged: IST brand"),
    value: 0n,
  },
  totalMintedProvided: {
    brand: slot(0),
    value: 0n,
  },
}
```

```
$ agoric follow :published.provisionPool.metrics
{
  provisionedCount: 1n,
  totalMintedConverted: {
    brand: slot(0,"Alleged: IST brand"),
    value: 1233876600n,
  },
  totalMintedProvided: {
    brand: slot(0),
    value: 250000n,
  },
}
```

### published governance: OK

```
$ agoric follow :published.provisionPool.governance
{
  current: {
    Electorate: {
      type: "invitation",
      value: {
        brand: slot(0,"Alleged: Zoe Invitation brand"),
        value: [
          {
            description: "questionPoser",
            handle: slot(1,"Alleged: InvitationHandle"),
            installation: slot(2,"Alleged: BundleInstallation"),
            instance: slot(3,"Alleged: InstanceHandle"),
          },
        ],
      },
    },
    PerAccountInitialAmount: {
      type: "amount",
      value: {
        brand: slot(4,"Alleged: IST brand"),
        value: 250000n,
      },
    },
  },
}
```

based on 939317565, which includes #6187

packages/vats/src/provisionPool.js
```diff
@@ -15,16 +17,33 @@ const { details: X, quote: q } = assert;
...
+  storageNode: M.eref(M.remotable('storageNode')),
+  marshaller: M.eref(M.remotable('marshaller')),
 });
 
+/**
+ * @typedef {object} MetricsNotification
+ * Metrics naming scheme is that nouns are present values and past-participles
+ * are accumulative.
+ *
+ * @property {bigint} provisionedCount  count of new wallets provisioned
```
turadg: this is an accumulative value. I think it should be `walletsProvisioned`. 

requesting change for this.

packages/vats/src/provisionPool.js
```diff
@@ -158,16 +220,32 @@ export const start = (zcf, privateArgs) => {
       });
   };
 
-  const creatorFacet = Far('Provisioning Pool creator', {
-    makeHandler: makeBridgeProvisionTool(sendInitialPayment),
+  // For documentation and/or future upgrade to Far
```
turadg, line 223: pretty easy to use `makeFarHeapInstance` rn, up to you

turadg (changes requested):
Logic LGTM. Nice job improving type factoring.

One request on the metrics, before downstream depends on the name `provisionedCount`. It's an accumulator so should be past-participle. E.g. `walletsProvisioned`.

packages/vats/src/provisionPool.js
dckc: I tried once without luck. I'm interested to try again with a little help.

packages/vats/src/core/startWalletFactory.js
```diff
@@ -16,11 +17,95 @@ const startFactoryInstance = (zoe, inst) => E(zoe).startInstance(inst);
...
+    governedContractInstallation,
+    issuerKeywordRecord,
+    terms,
+    privateArgs,
+  },
+  { governedParams, timer, contractGovernor, economicCommitteeCreatorFacet },
+) => {
+  const poserInvitationP = E(
+    economicCommitteeCreatorFacet,
+  ).getPoserInvitation();
```
Chris-Hibbert, line 48: For POLA, I'd prefer to pass the poserInvitations around rather than the committeeCreatorFacet. Maybe that's a TODO rather than something to do in this PR.

packages/vats/src/provisionPool.js
dckc: ok. changed.

packages/vats/src/provisionPool.js
turadg: approving now but happy to work with you to get the change in if this ends up needing another push

turadg (approved):

Chris-Hibbert (commented):
Still reading, but this is what I have so far.

packages/vats/src/provisionPool.js
turadg: https://github.com/Agoric/agoric-sdk/compare/6067_governance...ta/6067_governance

packages/vats/src/provisionPool.js
dtribble: only `PerAccountInitialAmount`

packages/vats/src/core/startWalletFactory.js
dtribble: I agree that would be better. I was thinking this was in the test setup, not the system setup.

dckc: Let's do that in a follow-on PR. Getting this green in CI is an opportunity that I don't want to pass on.

PR #6182, opened 2022-09-12
test(wallet-factory): stop using legacy bundle

dckc:
closes: #XXXX
refs: #XXXX


packages/smart-wallet/test/devices.js
```diff
@@ -10,7 +10,6 @@ export const devices = {
           case 'centralSupply':
             return bundleCentralSupply;
           case 'mintHolder':
-            return bundleMintHolder;
```
turadg, line 13: stray edit. I think you wanted to remove the next line (which needs removing for this PR )

turadg (changes requested):

PR #6187, opened 2022-09-13
test(cosmic-swingset): pay 10 BLD for account with 0.25 IST to start

dckc:
refs: #6067

## Description

Integration test, including golang cosmos-sdk layers, to follow #6157 

```
  ✔ PSM bootstrap and provision (1m 39.3s)
    ℹ make scenario2-run-psm-halt
    ℹ Fund pool with USDC
    ℹ make ACCT_ADDR=agoric1megzytg65cyrgzs6fvzxgrcqvwwl7ugpt62346 SOLO_COINS=1234000000ibc/usdc1234 fund-acct
    ℹ Fund user account with some BLD
    ℹ make ACCT_ADDR=agoric1ctncw9sguzftds55tss72n5htdla8e3s2wqrwt SOLO_COINS=123000000ubld fund-acct
    ℹ Provision an account
    ℹ make ACCT_ADDR=agoric1ctncw9sguzftds55tss72n5htdla8e3s2wqrwt provision-acct
    ℹ verify 10BLD spent, 0.25 IST received
```

### Security Considerations

?

?

### Testing Considerations

**increases ava timeout to 25 min in packages/cosmic-swingset**

 - [x] TODO: wait until the 0.25 IST deposit should clear
 - [ ] figure out why it fails with EOF in ci

### DIFF
--- a/packages/cosmic-swingset/test/test-make.js
+++ b/packages/cosmic-swingset/test/test-make.js
@@ -1,45 +1,41 @@
 import test from 'ava';
-import { spawn } from 'child_process';
-import path from 'path';
 
-const filename = new URL(import.meta.url).pathname;
-const dirname = path.dirname(filename);
+// Use ambient authority only in test.before()
+import { spawn as ambientSpawn } from 'child_process';
+import * as ambientPath from 'path';
 
-test('make and exec', async t => {
-  await new Promise(resolve =>
-    spawn('make', ['scenario2-setup'], {
-      cwd: `${dirname}/..`,
-      stdio: ['ignore', 'ignore', 'inherit'],
-    }).addListener('exit', code => {
-      t.is(code, 0, 'make scenario2-setup exits successfully');
-      resolve();
-    }),
+import { makeScenario2, pspawn } from './scenario2.js';
+
+test.before(async t => {
+  const filename = new URL(import.meta.url).pathname;
+  const dirname = ambientPath.dirname(filename);
+  const makefileDir = ambientPath.join(dirname, '..');
+
+  const io = { spawn: ambientSpawn, cwd: makefileDir };
+  const pspawnMake = pspawn('make', io);
+  const pspawnAgd = pspawn('bin/ag-chain-cosmos', io);
+  const scenario2 = makeScenario2({ pspawnMake, pspawnAgd, log: t.log });
+  await scenario2.setup();
+
+  t.context = { scenario2, pspawnAgd };
+});
+
+test.serial('make and exec', async t => {
+  const { pspawnAgd, scenario2 } = t.context;
+  t.log('exec agd');
+  t.is(await pspawnAgd([]).exit, 0, 'exec agd exits successfully');
+  t.log('run chain to halt');
+  t.is(
+    await scenario2.runToHalt(),
+    0,
+    'make scenario2-run-chain-to-halt is successful',
   );
-  await new Promise(resolve =>
-    spawn('bin/ag-chain-cosmos', {
-      cwd: `${dirname}/..`,
-      stdio: ['ignore', 'ignore', 'inherit'],
-    }).addListener('exit', code => {
-      t.is(code, 0, 'exec exits successfully');
-      resolve();
-    }),
+  t.log('resume chain and halt');
+  t.is(
+    await scenario2.runToHalt(),
+    0,
+    'make scenario2-run-chain-to-halt succeeds again',
   );
-  await new Promise(resolve =>
-    spawn('make', ['scenario2-run-chain-to-halt'], {
-      cwd: `${dirname}/..`,
-      stdio: ['ignore', 'ignore', 'inherit'],
-    }).addListener('exit', code => {
-      t.is(code, 0, 'make scenario2-run-chain-to-halt is successful');
-      resolve();
-    }),
-  );
-  await new Promise(resolve =>
-    spawn('bin/ag-chain-cosmos', ['export', '--home=t1/n0'], {
-      cwd: `${dirname}/..`,
-      stdio: ['ignore', 'ignore', 'inherit'],
-    }).addListener('exit', code => {
-      t.is(code, 0, 'export exits successfully');
-      resolve();
-    }),
-  );
+  t.log('export');
+  t.is(await scenario2.export(), 0, 'export exits successfully');
 });
--- a/packages/cosmic-swingset/Makefile
+++ b/packages/cosmic-swingset/Makefile
@@ -19,6 +19,8 @@ VOTE_PROPOSAL = 1
 VOTE_OPTION = yes
 
 GOSRC = ../../golang/cosmos
+# For deep (cross-vat) stacks, try...
+# DEBUG ?= SwingSet:ls,SwingSet:vat,track-turns
 DEBUG ?= SwingSet:ls,SwingSet:vat
 AG_SOLO = DEBUG=$(DEBUG) $(shell cd ../solo/bin && pwd)/ag-solo
 AGC = DEBUG=$(DEBUG) PATH="$$PWD/bin:$$PATH" $(GOSRC)/build/agd
@@ -150,10 +152,12 @@ scenario2-run-chain: ../vats/decentral-devnet-config.json
 		$(AGC) --home=t1/n0 start --log_level=warn $(AGC_START_ARGS)
 
 # Run a chain with an explicit halt.
+BLOCKS_TO_RUN=3
 scenario2-run-chain-to-halt: t1/decentral-economy-config.json
 	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/t1/decentral-economy-config.json" \
-		$(AGC) --home=t1/n0 start --log_level=warn --halt-height=$$(($(INITIAL_HEIGHT) + 3)); \
+		$(AGC) --home=t1/n0 start --log_level=warn --halt-height=$$(($(INITIAL_HEIGHT) + $(BLOCKS_TO_RUN))); \
 		test "$$?" -eq 98
+	echo ran to $(INITIAL_HEIGHT) + $(BLOCKS_TO_RUN)
 
 # Blow away all client state to try again without resetting the chain.
 scenario2-reset-client:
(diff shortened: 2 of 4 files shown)


packages/cosmic-swingset/package.json
```diff
@@ -61,6 +61,6 @@
     "files": [
       "test/**/test-*.js"
     ],
-    "timeout": "20m"
+    "timeout": "25m"
```
turadg: consider a TODO or issue to see if we can get this back down after launch. or simply, "this was 20min until #6187" for when someone comes looking to optimize CI

packages/cosmic-swingset/test/test-make.js
```diff
@@ -1,26 +1,51 @@
...
       stdio: ['ignore', 'ignore', 'inherit'],
     }).addListener('exit', code => {
-      t.is(code, 0, 'make scenario2-setup exits successfully');
+      t.is(code, 0, 'exec exits successfully');
       resolve();
     }),
   );
   await new Promise(resolve =>
-    spawn('bin/ag-chain-cosmos', {
+    spawn('make', ['scenario2-run-chain-to-halt'], {
```
turadg: `runMake` looks perfect for this :)

packages/cosmic-swingset/test/test-make.js
```diff
@@ -43,3 +68,92 @@ test('make and exec', async t => {
     }),
   );
 });
+
+const query = async args => {
+  const parts = [];
+  const run = new Promise((resolve, reject) => {
```
turadg: this isn't really an async function.
```suggestion
const query = args => {
  const parts = [];
  return new Promise((resolve, reject) => {
```

packages/cosmic-swingset/test/test-make.js
```diff
@@ -43,3 +68,92 @@ test('make and exec', async t => {
...
+      { cwd: `${dirname}/..` },
+    );
+    cmd.stdout.on('data', chunk => parts.push(chunk));
+    cmd.on('exit', code => {
+      if (code !== 0) {
+        reject(Error(`exit ${code} from: query ${args}`));
+        return;
+      }
+      const txt = parts.join('');
+      if (txt.trim() === '') {
```
turadg: might as well pass the trimmed version to JSON.parse()
```suggestion
      const txt = parts.join('').trim();
      if (txt === '') {
```

turadg: might as well pass the trimmed version to JSON.parse()
```suggestion
      const txt = parts.join('').trim();
      if (txt === '') {
```

packages/cosmic-swingset/test/test-make.js
```diff
@@ -43,3 +68,92 @@ test('make and exec', async t => {
     }),
   );
 });
+
+const query = async args => {
```
turadg: a comment here would help.
```
/**
 * @param {string[]} args
 * @returns {Promise<string>} stdout of `ag-chain-cosmos query <...args>`
 */
```

packages/cosmic-swingset/test/test-make.js
```diff
@@ -43,3 +68,92 @@ test('make and exec', async t => {
...
+    } catch (reason) {
+      console.warn(reason);
+      await delay(5_000);
+    }
+  }
+};
+
+test.serial('PSM bootstrap and provision', async t => {
+  // XXX coordinate with vbank/provision from golang
+  const pool = 'agoric1megzytg65cyrgzs6fvzxgrcqvwwl7ugpt62346';
```
turadg: where does this string come from? I see the comment about coordinating, which made me think it was from that part of the codebase but I don't see it anywhere.

packages/cosmic-swingset/test/test-make.js
```diff
@@ -43,3 +68,92 @@ test('make and exec', async t => {
...
+
+  const addrLine = await fs.promises.readFile(
+    './t1/8000/ag-cosmos-helper-address',
+    'utf-8',
+  );
+  const addr = addrLine.trim();
```
turadg: since you're immediately awaiting, might as well read sync. and no need to bind addrLine.
```suggestion
  const addr = fs.readFileSync(
    './t1/8000/ag-cosmos-helper-address',
    'utf-8',
  ).trim();
```

packages/cosmic-swingset/test/test-make.js
```diff
@@ -43,3 +68,92 @@ test('make and exec', async t => {
...
+
+  const show = args => {
+    t.log('make', ...args);
+    return args;
+  };
```
turadg: this is a little surprising. I'd more expect logging `make` to be the concern of `runMake`. consider giving it an optional `log` argument that defaults to noop:
```
    await runMake(
      [`ACCT_ADDR=${addr}`, 'SOLO_COINS=123000000ubld', 'fund-acct'],
      t.log,
    );
```

turadg (approved):
Great to have this test automated!

packages/cosmic-swingset/test/test-make.js
dckc: synchronous I/O messes with turn isolation. It should only be used in low-level stuff that's building a platform for normal code.

> The net effect is that a turn is E’s unit of operation. We can faithfully account for the visible effects of concurrency without any interleaving of the steps within a turn. Any actual multi-vat computation is equivalent to some fully ordered
interleaving of turns.
> The actual E system does provide synchronous file I/O operations. When these files are local, prompt, and private to the vat accessing them, this does not violate turn isolation, but since files may be remote, non-prompt, or shared, the availability of these synchronous I/O operations does violate the E model -- [Concurrency Among Strangers](http://papers.agoric.com/assets/pdf/papers/concurrency-among-strangers.pdf)

packages/cosmic-swingset/test/test-make.js
dckc: I considered factoring `runMake` out of all the places where it could/should be used, but I thought it would be good to make it clear that much of the code isn't changing at all. I'll give it another look.

packages/cosmic-swingset/test/test-make.js
turadg: Thanks for the lesson. I fail to see how it applies here as this is a test, running in a single VM, serially.

Not opposed though

dckc:
> Great to have this test automated!

yes... well, it would be... if it worked in ci like it does for me locally. But it doesn't, and the diagnostic doesn't give me much to go on:

```
Error: post failed: Post "http://localhost:26657": EOF
Usage:
  agd tx swingset provision-one [nickname] [address] [power-flags] [flags]
```

I'll need help in order to advance this.

packages/cosmic-swingset/test/test-make.js
turadg: as a separate commit that would be clear enough imo

michaelfig:
Hi @dckc, do you want to try resurrecting this one and updating it to master?  Or has its time passed?

dckc:
Let's see... yes, this is a [documented feature](https://docs.inter.trade/user-how-to/wallet-usage/create-your-agoric-account-and-smart-wallet), so it should have an automated test.

As I mentioned, I'll need help. Do you have any clues for me, @michaelfig ?

Here's hoping for time to look at it together.

dckc:
after spending about 2.5 days on this last week without getting a reliable result, how about we keep the resulting test code, which I mostly like, but use `test.skip` to keep the new test out of ci? I'm pushing changes to that effect.

packages/cosmic-swingset/test/test-make.js
```diff
@@ -1,45 +1,249 @@
...
-      cwd: `${dirname}/..`,
-      stdio: ['ignore', 'ignore', 'inherit'],
-    }).addListener('exit', code => {
-      t.is(code, 0, 'export exits successfully');
-      resolve();
+  t.log('export');
+  t.is(await scenario2.export(), 0, 'export exits successfully');
+});
+
+// SKIP: struggling with timing issues resulging in one of...
```
turadg: s/ging/ting

packages/cosmic-swingset/test/test-make.js
```diff
@@ -1,45 +1,249 @@
...
+// Error: cannot grab 250000uist coins: 0uist is smaller than 250000uist: insufficient funds
+// error: code = NotFound desc = account agoric1mhu... not found
+// Sometimes I can get this test to work alone, but not
+// if run with the test above.
```
turadg: Seems tackling someday. If you agree, this would be a good place to reference an issue for it

turadg (commented):
> how about we keep the resulting test code, which I mostly like, but use `test.skip` to keep the new test out of ci? I'm pushing changes to that effect.

Works for me, but before merge I'd like to get clear on:
1. Whether this is worth making work in CI eventually. If so, let's have an issue an reference it here.
2. How this will be tested outside of CI. What sorts of changes should trigger someone to uncomment `.skip` and disable the other one so it's testing?

It seems like if the problem is the coordination between the two tests, we could solve that by separating them and having two chain starts. That's prohibitive for CI but maybe not for the slow tests for merging to master. Alternately we could have a script like "run this locally when you change something in the econ boot". Eventually we could have CI be smart enough to run that slow one but only when necessary.

packages/cosmic-swingset/test/test-make.js
dckc: ok: #6766

dckc:
> It seems like if the problem is the coordination between the two tests, we could solve that by separating them and having two chain starts.

That prompted me to reconsider how this is factored. I factored out `scenario2.js` and moved `test-provision-smartwallet.js` to its own test file... still with `test.skip`. This suggests 2 chain starts, but we can move the test back once we figure out the quirks.

It makes the change to `test-make.js` a bit less invasive... you don't have to wade thru as much provisionpool stuff to read that test.

> That's prohibitive for CI but maybe not for the slow tests for merging to master.

I'm not familiar with how the slow tests work, so I lean toward leaving that aside.

> 2. How this will be tested outside of CI. What sorts of changes should trigger someone to uncomment `.skip` and disable the other one so it's testing?

Not sure... at least you and I know about it... I think of #6766 representing the work to make sure it's part of the development process.

packages/vats/decentral-psm-config.json
```diff
@@ -4,6 +4,7 @@
   "snapshotInterval": 1000,
   "vats": {
     "bootstrap": {
+      "sourceSpecXXX": "is boot-psm.js still working???",
```
turadg: to remove before merge

turadg (approved):
Refactoring looks good. Thanks for the documentation and the new issue

PR #6188, opened 2022-09-13
feat: Write makeSnapshot telemetry to slog

### LINKED ISSUE #6164: Telemetry for Heap Snapshotting
We found that heap snapshots are occurring in the middle of cranks, are quite large (100 MB+), and take a long time (7+ sec).

Before we can figure out how to improve this, we need to instrument to find out exactly where the time is spent.

gibson042:
closes: #6164

## Description

Endow SnapStore instances with a now function so `save` can return duration and size telemetry all the way up to VatKeeper `saveSnapshot` where it is written into the slog.

### Security Considerations

Filesystem access is generally more powerful than a timer, so this should be acceptable.

n/a

### Testing Considerations

I'm not actually testing slog contents, is that something we do?


packages/swing-store/src/snapStore.js
```diff
@@ -5,6 +5,17 @@ import { createGzip, createGunzip } from 'zlib';
 import { assert, details as d } from '@agoric/assert';
 import { promisify } from 'util';
 
+/**
+ * @template T
+ * @typedef {{
+ *   has: (hash: string) => Promise<boolean>,
+ *   load: (hash: string, loadRaw: (filePath: string) => Promise<T>) => Promise<T>,
```
mhofman: I think it's better to templatize the load function only, not the whole SnapStore, to avoid having to be explicit for types throughout.

```suggestion
 *   load: <T>(hash: string, loadRaw: (filePath: string) => Promise<T>) => Promise<T>,
```

packages/swing-store/src/snapStore.js
```diff
@@ -51,6 +62,7 @@ export const fsStreamReady = stream =>
   });
 
 /**
+ * @template T
```
mhofman: Template arguments that cannot be inferred from params are really hard to deal with in JSDoc.

packages/swing-store/src/snapStore.js
```diff
@@ -119,20 +131,18 @@ export function makeSnapStore(
   }
 
   /**
-   * @param {string} dest basename, relative to root
+   * @param {string} dest absolute path of which root is a prefix
```
mhofman: What's the motivation to switch to absolute paths? I think there is a risk a future change will make the path not prefixed by `root` in which case the rename is not atomic.
Maybe the solution is to extract the `tmpdir` as the dirname of `dest` and ignore `root` ?

packages/swing-store/src/snapStore.js
```diff
@@ -198,35 +208,40 @@ export function makeSnapStore(
...
    */
   async function save(saveRaw) {
     return withTempName(async snapFile => {
+      const t0 = /** @type {number} */ (now());
       await saveRaw(snapFile);
-      const stats = await stat(snapFile);
-      const rawsize = stats.size;
+      const rawSaveDuration = /** @type {number} */ (now()) - t0;
+      const { size: rawsize } = await stat(snapFile);
       const h = await fileHash(snapFile);
```
mhofman: I'd like to measure the time we take to perform `fileHash`

packages/swing-store/src/swingStore.js
```diff
@@ -21,6 +21,7 @@ export function makeSnapStoreIO() {
     tmpName,
     createReadStream: fs.createReadStream,
     createWriteStream: fs.createWriteStream,
+    now: Date.now,
```
mhofman: I think we should pass `performance` instead of `Date`. That's monotonic and what we usually use to measure durations in other places of the code base.

packages/SwingSet/src/kernel/state/vatKeeper.js
```diff
@@ -606,6 +612,10 @@ export function makeVatKeeper(
       type: 'heap-snapshot-save',
       vatID,
       snapshotID,
+      rawByteCount,
+      rawSaveMillisec,
```
mhofman: We usually write all our timings in seconds, clamped to 6 decimals.

mhofman (changes requested):
Mostly stylistic nits, but I'd like to measure the hashing time since it's sequential, and we should stay consistent regarding time formatting in the slog.

gibson042:
@mhofman I've taken all your suggestions. PTAL.

packages/swing-store/src/swingStore.js
```diff
@@ -17,10 +18,20 @@ export const DEFAULT_LMDB_MAP_SIZE = 2 * 1024 * 1024 * 1024;
...
 
 export function makeSnapStoreIO() {
+  const { now } = performance;
+  // TODO: find a home for `export const makeMeasureSeconds = now => {...}`.
+  const measureSeconds = async fn => {
+    const t0 = now();
+    const result = await fn();
+    const durationMillisec = now() - t0;
+    return { result, duration: durationMillisec / 1000 };
+  };
```
gibson042: Thoughts on the best location for a function that accepts a [`performance.`]`now` and returns a function that uses it to report the time taken to invoke a provided function (in seconds to promote telemetry consistency)? Is [telemetry](https://github.com/Agoric/agoric-sdk/tree/HEAD/packages/telemetry) a good fit, or should I just go straight to [internal](https://github.com/Agoric/agoric-sdk/tree/HEAD/packages/internal)?

packages/swing-store/src/snapStore.js
```diff
@@ -231,12 +237,13 @@ export function makeSnapStore(
       if (fileStat) {
         return freeze(info);
```
mhofman: Since we `stat`ed the file, should we include the `compressedByteCount` with a `0` or `undefined` `compressSeconds` ?

packages/swing-store/src/swingStore.js
mhofman: I don't think telemetry is the right home, probably the new internal? I'd want to have it be a maker that takes a `now` function.

packages/swing-store/src/snapStore.js
```diff
@@ -106,41 +134,46 @@ export function makeSnapStore(
...
+   * @returns {Promise<void>}
    */
-  async function atomicWrite(dest, thunk) {
-    assert(!dest.includes('/'));
-    const tmp = await ptmpName({ tmpdir: root, template: 'atomic-XXXXXX' });
-    let result;
+  async function atomicWrite(destFilePath, writeContents) {
+    // Atomicity requires remaining on the same filesystem,
+    // so we perform all operations in the destination directory.
+    const dir = resolve(destFilePath, '..');
```
mhofman: Why not use `dirname`?

```suggestion
    const dir = path.dirname(destFilePath);
```

mhofman (approved):

packages/swing-store/src/snapStore.js
gibson042: Good point... I think we do want to emit fields consistently, including data types. I've opted to read `compressedByteCount` from stat info, use 0 for `compressSeconds`, and introduce a boolean `newFile` to support exclusion of those values for accurate statistics.

packages/swing-store/src/snapStore.js
gibson042: `resolve` is an endowment to `makeSnapStore`, and I didn't want to require another one or risk conflict between it and the built-in path module (however unlikely that may be). But I've restored root-relative behavior to avoid that, and this should be refactored in the performance-improving followup anyway.

packages/swing-store/src/swingStore.js
gibson042: Done.

gibson042:
@mhofman Back to you. And should I start work on the performance improvements, or did you plan to take that on?

mhofman (approved):
LGTM

Issue #5894, opened 2022-08-04
make Smart Wallet contract durable


turadg:
All the wallet backends need to be implemented by a single wallet manager contract, since it is inefficient to have many small vats when a large one would suffice.  Multitenant backend wallet states, which are managed by a single Zoe contract, are presented as sealed individual backends, each of the schema in #4483, and broadcasted via the Cosmos event stream.

To facilitate migration to virtual/durable objects and collection, each individual wallet backend must be supplied with a new schema manager rather than instantiating its own collections.

While refactoring, consider moving `address` out of smartWallet scope. E.g. pass in a wallet logger and storageNode

PR #6191, opened 2022-09-13
wallet misc

turadg:
## Description

- test(agops): smoke test
- chore(smart-wallet): more logging
- build(smart-wallet): build bundles
- fix(agoric-cli): perf watch
- fix(agoric-cli): psm giveMinted
- chore: safer offerResults publishing

### Security Considerations

Stops publishing offerResult for copyRecord. Publishes simple types.

--
### Testing Considerations

--

### DIFF
--- a/packages/smart-wallet/src/offers.js
+++ b/packages/smart-wallet/src/offers.js
@@ -31,6 +31,7 @@ export const UNPUBLISHED_RESULT = 'UNPUBLISHED';
  * @param {{ receive: (payment: *) => Promise<Amount> }} opts.depositFacet
  * @param {object} opts.powers
  * @param {import('./types').Cell<number>} opts.powers.lastOfferId
+ * @param {Pick<Console, 'info'| 'error'>} opts.powers.logger
  * @param {(spec: import('./invitations').InvitationSpec) => ERef<Invitation>} opts.powers.invitationFromSpec
  * @param {(brand: Brand) => import('./types').RemotePurse} opts.powers.purseForBrand
  * @param {(status: OfferStatus) => void} opts.onStatusChange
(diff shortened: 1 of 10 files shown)


packages/smart-wallet/src/offers.js
```diff
@@ -43,7 +44,7 @@ export const makeOfferExecutor = ({
   onStatusChange,
   onNewContinuingOffer,
 }) => {
-  const { invitationFromSpec, lastOfferId, purseForBrand } = powers;
+  const { invitationFromSpec, lastOfferId, logger, purseForBrand } = powers;
```
dckc, line 47: :+1: ocap-happy logging

packages/smart-wallet/src/smartWallet.js
```diff
@@ -343,7 +343,9 @@ const behavior = {
      */
     async executeOffer(offerSpec) {
       const { facets, state } = this;
+      /** @type {State} */
       const {
+        address,
```
dckc, line 348: hm. I missed something somewhere. A wallet knows the address it's associated with? Is that necessary? It sets off my ACL spidey-sense... it feels vagely like [caller](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/caller) or `msg.sender` in solidity.

packages/smart-wallet/src/offers.js
```diff
@@ -134,15 +143,19 @@ export const makeOfferExecutor = ({
...
+          numSatisfied => {
+            logger.info(id, 'numSatisfied', numSatisfied);
+            if (numSatisfied === 0) {
+              updateStatus({ numWantsSatisfied: 0 });
+            }
+            updateStatus({
+              numWantsSatisfied: numSatisfied,
+            });
+          },
+          handleError,
```
dckc, line 157: The commit message doesn't mention this. (not critical)

dckc (approved):
:+1:

packages/smart-wallet/src/smartWallet.js
turadg: Yeah, I had the same tingle. It's not strictly necessary. We could take it out by passing two other things in instead:
- wallet logger
- wallet storageNode

But it has to persist through init and there may be some churn with durability, so I'll punt until then: https://github.com/Agoric/agoric-sdk/issues/5894

PR #6192, opened 2022-09-13
governance cleanup

turadg:
## Description

Remove dead code

### DIFF
--- a/packages/governance/src/contractGovernor.js
+++ b/packages/governance/src/contractGovernor.js
@@ -61,7 +61,6 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
 /**
  * @typedef {StandardTerms} ContractGovernorTerms
  * @property {TimerService} timer
- * @property {Instance} electorateInstance
  * @property {Installation} governedContractInstallation
  */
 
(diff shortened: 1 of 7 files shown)


Chris-Hibbert (changes requested):
The changes to `startPSM` are clearly an improvement. I'm still trying to figure out whether we lose anything with the changes to `contractGovernor`.  I'd rather not have to think about this while struggling to get to completion on PS0. 

If this were reduced to just the elimination of economicCommittee from `startPSM`, I'd approve it immediately.

dtribble (approved):
LGTM

turadg:
> I'm still trying to figure out whether we lose anything with the changes to `contractGovernor`.

My read is that the Electorate instance is a governed param now instead of a term:
https://github.com/Agoric/agoric-sdk/blob/c615183d0da33e97045957f1580b91a370fe427d/packages/governance/src/contractGovernor.js#L171-L177

So this was just uncollected garbage in that change. It's also passing all tests, so if there's a reason it _is_ required that's something to dig into. I've added a couple more cleanup commits on which I'm waiting to see CI results.

> I'd rather not have to think about this while struggling to get to completion on PS0.

Fair enough that feature complete is higher priority. I think this should get in before security review though. I'm adding it to the release.

packages/governance/src/contractGovernor.js
```diff
@@ -61,7 +61,6 @@ const validateQuestionFromCounter = async (zoe, electorate, voteCounter) => {
 /**
  * @typedef {StandardTerms} ContractGovernorTerms
  * @property {TimerService} timer
- * @property {Instance} electorateInstance
```
Chris-Hibbert, line 64: Please remove "electorate" from line 72 below, and add the following paragraph after line 84:
```
 * The electorate that will vote on governance questions is itself a governed
 * parameter. Its value is available from the publicFacet using
 * getElectorateInstance(). When creating new questions, we use
 * getUpdatedPoserFacet() to inform the electorate about the new question.
```

Chris-Hibbert (approved):

packages/governance/src/contractGovernor.js
turadg: Thanks for the close read!

I'm force-pushing with that comment change.

PR #6193, opened 2022-09-14
refactor(cosmic-swingset): merge block-manager into launch-chain

mhofman:
refs: #6210 

## Description

This is what should hopefully be a pure refactor ahead of introducing extra queues at the cosmic-swingset layer. Since following work will require not immediately feeding all incoming actions onto swingset devices, it's a lot easier if the unnecessary level of indirections are removed. The goal is to keep the safety of the block manager regarding reentrancy and hangover inconsistency handling, while making it easier to understand the execution flow.

This PR does slightly move a few operations:
- the installation publisher is now created on the first begin block action after restart. I believe this is equivalent to what was actually happening before. cc @kriskowal 
- the chain send reset / replay logic has been split up (including I believe a fix for a double crash hangover, cc @michaelfig )
- blockingSend and related processAction now lives in launch-chain, near the actual action handlers, avoiding the need to thread them through.

### Security Considerations

None that I'm aware of

Some scripts depend on the `block-manager` output on stdout, so I've maintained backward compat for that.

### Testing Considerations

Ran the loadgen locally, including using the publishBundle branch, but I'm not sure if this exercises all possible code paths, and cosmic-swingset is notoriously lacking unit tests.


mhofman:
Best reviewed commit-by-commit, as usual.

warner (approved):
@mhofman and I walked through this together, looks good!

kriskowal (approved):
Moving `installBundle` looks like a good refactor to me.

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -277,17 +278,17 @@ export default async function main(progname, args, { env, homedir, agcc }) {
...
+    return chainSends;
+  };
 
-    if (!doReplay) {
-      return;
-    }
+  // Replay and clear the chain send queue.
+  // While replaying each send, insist it has the same return result.
+  function replayChainSends() {
+    // Remove our queue.
```
michaelfig, line 290:
```suggestion
    // Copy our queue.
```

michaelfig (approved):
LGTM!  Thanks for attending to this.

PR #6199, opened 2022-09-14
fix: increase snapshot interval from 200 to 1,000

### LINKED ISSUE #6196: Change the heap snapshot interval from 200 to 1,000

dckc:
fixes #6196

### Description

snapshots seem to take a lot of time; with this change, we do it 5x less often

### Testing Considerations

untested


dckc:
better done at a higher layer

PR #6202, opened 2022-09-14
fix: psmCharter.addInstance interface shape correction

Chris-Hibbert:
## Description

Correct the interface shape declaration for `psmCharter.addInstance()`.  The last two parameters are optional.

### Security Considerations

None

None

### Testing Considerations

Exercised by running `make scenario2-setup scenario2-run-chain-psm`.


packages/inter-protocol/src/psm/psmCharter.js
```diff
@@ -96,12 +96,9 @@ export const start = async zcf => {
...
 
   const psmCharterCreatorI = M.interface('PSM Charter creatorFacet', {
-    addInstance: M.call(
-      InstanceHandleShape,
-      M.any(),
-      BrandShape,
-      BrandShape,
-    ).returns(),
+    addInstance: M.call(InstanceHandleShape, M.any())
+      .optional(BrandShape, BrandShape)
```
turadg, line 100: TIL `.optional`. Where is it documented or implemented?

turadg (approved):
`.optional` is new to me but if it works for you locally then I believe it. 

I added the `bypass:integration` label. Since the test didn't catch the bug going it it shouldn't be affected by it being fixed ;-)

dckc (approved):

packages/inter-protocol/src/psm/psmCharter.js
Chris-Hibbert: There's [very little documentation](https://github.com/Agoric/agoric-sdk/blob/c10c36d7ccf6c85239c1dbcec9534d43b20ad00a/packages/store/src/types.js).  I dunno where the implementations are. It's tested in [test-heap-classes](https://github.com/Agoric/agoric-sdk/blob/c10c36d7ccf6c85239c1dbcec9534d43b20ad00a/packages/store/test/test-heap-classes.js#L14).

turadg: Yeah, that documentation doesn't mention `optional`. I've filed https://github.com/Agoric/agoric-sdk/issues/6206 cc @erights

erights: From https://github.com/Agoric/agoric-sdk/issues/6206#issuecomment-1247546215 :

FWIW, this is implemented at https://github.com/Agoric/agoric-sdk/blob/9b786284b4ff759dd050c701829df5fd0cf41749/packages/store/src/patterns/patternMatchers.js#L1537-L1543 and used several times in https://github.com/Agoric/agoric-sdk/blob/master/packages/ERTP/src/typeGuards.js

I'm actually surprised I didn't write any comments explaining this at all. I certainly agree it is needed.

Also `rest`

PR #6203, opened 2022-09-14
test(vats): match decentral-psm-config.json against a pattern

dckc:
refs: #6009 

## Description

catch things like `[]` vs `{}` in `economicCommitteeAddresses`.

### Security Considerations

input validation is a security best practice

perhaps the bootstrap parameters for `boot-psm.js` could use more documentation.

### Testing Considerations

yup.

### DIFF
--- a/packages/vats/test/test-boot-config.js
+++ b/packages/vats/test/test-boot-config.js
@@ -1,0 +1,61 @@
+// @ts-check
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
+
+import { promises as fsPromises } from 'fs';
+import path from 'path';
+
+import { fit, M } from '@agoric/store';
+import { ParametersShape as BootParametersShape } from '../src/core/boot-psm.js';
+
+const pathname = new URL(import.meta.url).pathname;
+const dirname = path.dirname(pathname);
+const asset = (...ps) =>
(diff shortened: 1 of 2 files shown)


packages/vats/test/test-boot-config.js
```diff
@@ -0,0 +1,61 @@
...
+ * NOTE: this pattern suffices for PSM bootstrap,
+ * but does not cover the whole SwingSet config syntax.
+ *
+ * {@link packages/SwingSet/docs/configuration.md}
+ * TODO: move this to swingset?
+ *
+ * @see SwingSetConfig
+ * in packages/SwingSet/src/types-external.js
+ */
+const shape = (() => {
```
turadg, line 25: why the function? oh, to make a scope? these should be in their own `typeGuards.js` file which can encapsulate.

I think this does make more sense in SwingSet. prolly `vats/typeGuards.js`

turadg (commented):
This is strictly better than not having checks but I'll withhold Approve until it's in a more easily maintainable place. This feels like the kind of tech debt that would sit a long time otherwise.

packages/vats/test/test-boot-config.js
dckc: ok. let's do this in #6205 in any case.

PR #6204, opened 2022-09-14
feat: publish vote results from the voteCounter

### LINKED ISSUE #6198: BinaryVoteCounter should report vote results to chain storage

When we did an end-to-end test of committee voting on PSM governance, we noticed that the outcome of the vote wasn't recorded anywhere visible.  We looked at the clock and guess when it was done.

## Description of the Design

Use the standard mechanisms for publishing to chain storage.

## Security Considerations

Just visibility.

## Test Plan

straightforward.

Chris-Hibbert:
closes: #6198

## Description

 publish vote results from the voteCounter

### Security Considerations

this is already pubic.

We should have description of the structure of the data we publish. Do we, and I'm just unaware?

### Testing Considerations

The test doesn't yet verify the published data.   I hope to add that before merging.


turadg:
> We should have description of the structure of the data we publish. Do we, and I'm just unaware?

Agree. https://github.com/Agoric/agoric-sdk/blob/master/packages/inter-protocol/README.md#reading-data-off-chain is the closest we have, just the key structure.

The data streams are basically documented by the unit tests but whatever gets onto Mainnet should have end-user documentation. Ideally formatted in a way that an storage node explorer tool can include it.

Chris-Hibbert:
This now has tests and is ready to review.

packages/governance/src/binaryVoteCounter.js
```diff
@@ -113,6 +119,12 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
 
     if (!makeQuorumCounter(threshold).check(stats)) {
       outcomePromise.reject('No quorum');
+      const voteOutcome = {
+        question: details.questionHandle,
+        outcome: 'No quorum',
+      };
+      // @ts-expect-error Expand type to allow non-position?
```
turadg: yes, we should. I pushed e93c279

packages/governance/src/binaryVoteCounter.js
```diff
@@ -123,6 +135,12 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
     } else {
       outcomePromise.resolve(questionSpec.tieOutcome);
     }
+
+    // XXX if we should distinguish ties, publish should be called in if above
```
turadg, line 140: This would be good too but I don't object to leaving this tech debt for later. If you want to do it now, e93c279 shows how.

packages/governance/src/binaryVoteCounter.js
```diff
@@ -198,7 +216,7 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
 /**
  * @type {ContractStartFn<VoteCounterPublicFacet, VoteCounterCreatorFacet, {questionSpec: QuestionSpec, quorumThreshold: bigint}>}
  */
-const start = zcf => {
+const start = (zcf, { outcomesPublisher }) => {
```
turadg: Practically every publisher publishes more than once so the "s" plurality for that isn't informative. What is informative is what _each_ thing they publish is. (That helps to distinguish from when each publication is not unary.)

In this case, `outcomePublisher`.

packages/governance/src/committee.js
```diff
@@ -158,13 +158,24 @@ const start = (zcf, privateArgs) => {
           }
         };
 
+        const outcomeNode = E(privateArgs.storageNode).makeChildNode(
+          'latestOutcome',
```
turadg, line 162: a little hesitant to have a node for each of these but feels right to error on the side of more nodes so we don't create backwards compatibility problems for any nodes with broader responsibilities.

turadg (approved):
For expedience I pushed the changes I was going to suggest.

Approving with them ;-)

packages/governance/src/binaryVoteCounter.js
```diff
@@ -113,6 +119,13 @@ const makeBinaryVoteCounter = (questionSpec, threshold, instance) => {
 
     if (!makeQuorumCounter(threshold).check(stats)) {
       outcomePromise.reject('No quorum');
+      /** @type {OutcomeRecord} */
+      const voteOutcome = {
+        question: details.questionHandle,
+        outcome: 'fail',
+        reason: 'No quorum',
```
turadg, line 126: you mentioned that "quorum" isn't the right term for this. Now may be a good time to improve the message. We can at least provide the `threshold` value for diagnostics.

e..g. `Too few voters for threshold ${threshold}`

Chris-Hibbert: I'm resisting scope creep. I filed an issue (#6209) to fix the terminology from "quorum" to "threshold".   The term "quorum" appears in too many places to want to fix it at the last minute.

PR #6205, opened 2022-09-14
fix(vats): snapshot 5x less often

### LINKED ISSUE #6196: Change the heap snapshot interval from 200 to 1,000

dckc:
closes: #6196
refs: #XXXX

### Testing Considerations

added a unit test for config file syntax


turadg (approved):
Looks like it does what https://github.com/Agoric/agoric-sdk/issues/6196 says to do. 

I also confirmed 200 is the default. https://github.com/Agoric/agoric-sdk/blob/709045d2f87ddb61770dc62423f3091f4911f311/packages/SwingSet/docs/configuration.md?plain=1#L143-L148

I asked for changes on config validation PR but I don't object to e54fa8b5d48bd936cb656efc66c551085e9da6d5 bundled into this one, since that PR can serve to clean up the location.

packages/SwingSet/src/typeGuards.js
```diff
@@ -0,0 +1,41 @@
...
+  M.string(),
+  SwingSetConfigProperties,
+);
+
+/**
+ * NOTE: this pattern suffices for PSM bootstrap,
+ * but does not cover the whole SwingSet config syntax.
+ *
+ * {@link ./docs/configuration.md}
+ * TODO: move this to swingset?
```
turadg, line 28: done :)

dckc: oops.

PR #6208, opened 2022-09-15
clean up smart wallet

turadg:
## Description

The legacy-smart-wallet is no longer needed after https://github.com/Agoric/agoric-sdk/pull/6134 . This removes it.

Corrects that `vats` had the bundle for `walletFactory` contract.

### Security Considerations

--

--
### Testing Considerations

--


dckc (approved):
yay less code!

it seems like singleWallet was needed for something... but certainly not as a contract... and I guess not at all any more.

PR #6212, opened 2022-09-15
chore(deps): update Ava 4.3.1 to 5.0.1

turadg:
## Description

https://github.com/avajs/ava/releases/tag/v5.0.1

The big number is because it breaks Node <14 and TypeScript < 4.7, which we both support.

5.0.0 had some problems but https://github.com/avajs/ava/pull/3108 fixed them and 5.0.1 passes lint

### Security Considerations

--

--

### Testing Considerations

CI!


turadg:
The TypeScript 4.7 requirement breaks lines like `import('ava').ExecutionContext`. Solving that will take more investigation. Maybe getting rid of the patch, by turning replacing `import/no-unresolved` with https://github.com/import-js/eslint-import-resolver-typescript

gibson042 (approved):
LGTM, but I do see CI failures.

PR #6217, opened 2022-09-15
feat: Wallet UI offer status

michaelfig:
closes (almost): #6127 (leave open until Product verifies)

## Description

Last set of fixes needed to properly detect when a PSM offer has been processed by the contract.


packages/cosmic-swingset/Makefile
```diff
@@ -214,7 +213,7 @@ t1-provision-one-with-powers: wait-for-cosmos
 		  t1/$(BASE_PORT) $$addr $(AGORIC_POWERS) -ojson | tee /dev/stderr | grep -q '"code":0'; }
 
 # Send some USDC to the vbank/provision module account where it can be traded for IST.
-fund-provision-pool: wait-for-cosmos
```
turadg, line 217: why not wait-for-cosmos?

turadg (approved):
LGTM! Thanks for the collateral improvements too.

packages/cosmic-swingset/Makefile
michaelfig: I'm trying to avoid waiting twice as long as `fund-acct` because the next line invokes a recursive make that already depends on `wait-for-cosmos`.

PR #6167, opened 2022-09-08
feat: ensure voting via PSMCharter works with a unit test

Chris-Hibbert:
refs: #6110

## Description

On the path to end-to-end test of PSM voting by committee members, this unit test demonstrates (and makes it be the case) that votes can be instigated and carried out through the PSM Charter.

### Security Considerations

The Charter will be accessible to committee members, who will use it to start up votes.

### Documentation Considerations

Yes

### Testing Considerations

This shows, via unit test, that the Charter can be used to start votes.

PR #6218, opened 2022-09-16
test: make it possible to label representatives during debugging

Chris-Hibbert:
## Description

I've been looking for a way to distinguish different opaque representatives during debugging. @warner showed me how. I've added a flag to make it simple to turn this on when needed.

### Security Considerations

In production, logs shouldn't contain this detail. During debugging, it's fine.

We don't tend to write this kind of thing up.

### Testing Considerations

This made it far easier for me to debug while working on #6167.  I've wanted the same ability several times before.

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -18,6 +18,9 @@ import { parseVatSlot } from '../lib/parseVatSlots.js';
 // later regenerated.
 const unweakable = new WeakSet();
 
+// label representatives to simplify identification during debugging
+const LABEL_REPRESENTATIVES = false;
+
 /**
  * Make a simple LRU cache of virtual object inner selves.
  *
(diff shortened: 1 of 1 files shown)


Chris-Hibbert:
@warner I continue to find this useful while debugging. Any chance it could be added to the virtualObjectManager.js?

Chris-Hibbert:
Actually, I have an enhancement to add. I'll ask for a review again shortly.

erights (changes requested):
Just "Requesting changes" for now as an interlock, to be sure I look again before this is merged to master.

Chris-Hibbert:
I converted this to draft in response to a request to minimize open PRs that aren't slated for merging.

This is still useful to me, but it apparently conflicts with swingset's priorities.

What I get from this code is the ability to track individual objects in logging messages across vats. Without it, remote objects are all anonymous, and it is very hard to distinguish two objects of the same type, other than by their behavior.

turadg:
> This is still useful to me, but it apparently conflicts with swingset's priorities.

This would also be very useful to me for debugging contracts. Is there a way we can have a dev or debugging mode flag for this in master?

erights:
I like the goal. I do not like the means. For one thing, once we upgrade to latest endo, an eventual send with a non-understood method name will produce the list of actual methods, as it used to before virtual objects. I don't like that there will appear a method with a magic per-instance method name.

Ok if I give it a shot? I added myself to the assignees for this PR. I hope to find the time to try, sometime after we've upgraded agoric-sdk to depend on a recent endo release.

There is also the separate matter of how configuration options get passed through to code in vats. This needs a separate discussion.

Chris-Hibbert:
> There is also the separate matter of how configuration options get passed through to code in vats. This needs a separate discussion.

Is this related to how the feature could be enabled and disabled?  It would be nice to be able to control this at run-time, but not at all necessary for debugging. What other configuration options have an impact on this feature?

erights:
Closing in favor of https://github.com/endojs/endo/pull/1554 and its upcoming agoric-sdk sibling

PR #6219, opened 2022-09-16
fix: triage await safety

### LINKED ISSUE #6220: potential await hazard in liquidateIncrementally
https://github.com/Agoric/agoric-sdk/blob/a1dedeae72908fda45afcb6038d76f8359adc8de/packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js#L200

      // This await almost certainly does not do what was intended.
      // `oncePerBlock()` returns an async generator, not a promise,
      // and does nothing observably effectful to anyone not pulling on
      // the generator, which this code drops. AFAICT, this code is
      // equivalent to `await null;`. It imposes a non-delayed turn boundary
      // and does nothing else.

(Noticed while doing https://github.com/Agoric/agoric-sdk/pull/6219 )

### LINKED ISSUE #6237: Await safety not yet fixed in swing-store/src/snapStore.js for `rename` call
https://github.com/Agoric/agoric-sdk/blob/293f88efbd01553cc5be4d29c435cd509302b2e7/packages/swing-store/src/snapStore.js#L156

The triage currently at #6219 would fix it, with the explanation and code

```js
      // This nested await is safe because "IIAFE" (immediately invoked
      // async function expression). An IIAFE cannot synchronously throw
      // and is therefore a special case of "synchronous-throw-impossible".
      //
      // The original was
      // ```js
      // await rename(tmpFilePath, target);
      // ```
      // which we could not locally verify to be safe. `rename` is a parameter
      // of an exported function, which we should therefore be defensive
      // against. If it throws, `unlink(tmp)` could be reached synchronously,
      // whereas otherwise it would be reached asynchronously. Because
      // `unlink` is also a parameter of an exported function, we cannot
      // locally verify it to be stateless. In fact, from the name, we
      // expect it to be stateful. By changing the awaited expression to an
      // IIAFE, we ensure that the `unlink` can only be reached asynchronously.
      //
      await (async () => rename(tmpFilePath, target))();
```

Most of #6219 consists of commentary or cosmetic code changes, and so can be postponed until after PS0. But this is a safety fix. Whether we should cherry pick it, to get it in before PS0 depends on what the consequences are of not fixing it.

Assigning myself since I have the fix done in #6219 and would do the cherry picking if we decide to do that.

Git blame shows @gibson042 and @dckc as the one who should probably investigate this, so I'm assigning to them as well. Feel free to reassign as appropriate of course.

Also assigning @dtribble since he commented on this one and suggested a different fix.

### LINKED ISSUE #6239: Await safety not yet fixed in swing-store/src/snapStore.js for `sync` call
https://github.com/Agoric/agoric-sdk/blob/293f88efbd01553cc5be4d29c435cd509302b2e7/packages/swing-store/src/snapStore.js#L195

The triage currently at #6219 would fix it, with the explanation and code

```js
        // This nested await has been made SAFER by "IIAFE". Note that
        // doing so does *not* postpone when the `sync()` happens,
        // which would be especially scary for something named `sync()`.
        // Rather, it potentially postpones when the `close()`s happen
        // in case `sync()` throws. With the IIAFE, if this line is
        // executed, then the `close()`s would always happen in a later turn.
        //
        // But this nested await is not actually safe because it is nested in
        // an `if`, the other branch of which proceeds synchronously. The
        // `if` is not terminal because of the stateful `close()`s afterward.
        // On the else branch, those `close()`s would happen synchronously.
        await (async () => destination.sync())();
```

Most of #6219 consists of commentary or cosmetic code changes, and so can be postponed until after PS0. But this is a safety fix. Whether we should cherry pick it, to get it in before PS0 depends on what the consequences are of not fixing it.

Assigning myself since I have the fix done in #6219 and would do the cherry picking if we decide to do that.

Git blame shows @gibson042 and @mhofman as the ones who should probably investigate this, so I'm assigning to them as well. Feel free to reassign as appropriate of course.

Also assigning @dtribble since he commented on this one and suggested a different fix.

### LINKED ISSUE #6240: Await safety not yet fixed in swing-store/src/snapStore.js for `writeContents` argument function
https://github.com/Agoric/agoric-sdk/blob/293f88efbd01553cc5be4d29c435cd509302b2e7/packages/swing-store/src/snapStore.js#L265-L267

The triage currently at #6219 would fix it, with the explanation and code

```js
        // The `await` within `atomicWriteInRoot` above is only safe
        // when the second `writeContents` argument,
        // in all calls to `atomicWriteInRoot`, is a function whose calls cannot
        // synchronously throw. Even though this arrow function
        // seems to only call `filter`, which is a local async function and
        // therefore safe by itself, this arrow function also calls
        // `createGzip()`, which we cannot locally verify does not
        // synchronously throw. Thus, we made the arrow function async,
        // so that it cannot throw regardless. This seems to be consistent
        // with its purpose, and with the likely assumptions the program
        // author made when placing it here.
        //
        atomicWriteInRoot(baseName, async tmpGzPath =>
          filter(tmpSnapPath, createGzip(), tmpGzPath, { flush: true }),
        ),
```

Most of #6219 consists of commentary or cosmetic code changes, and so can be postponed until after PS0. But this is a safety fix. Whether we should cherry pick it, to get it in before PS0 depends on what the consequences are of not fixing it.

Assigning myself since I have the fix done in #6219 and would do the cherry picking if we decide to do that.

Git blame shows @gibson042 and @dckc as the ones who should probably investigate this, so I'm assigning to them as well. Feel free to reassign as appropriate of course.

### LINKED ISSUE #6242: Await safety not yet fixed in swing-store/src/snapStore.js for `runSteps` loop
https://github.com/Agoric/agoric-sdk/blob/293f88efbd01553cc5be4d29c435cd509302b2e7/packages/xsnap/src/replay.js#L213

The triage currently at #6219 would fix it, with the explanation and code

```js
    // Changed from `for of` to `for await of` to make awaits within
    // loop body safer. A `for await of` will correctly process a synchronous
    // iterator, but it still unconditionally separates the iterations
    // with turn boundaries.
    for await (const step of steps) {
```

Most of #6219 consists of commentary or cosmetic code changes, and so can be postponed until after PS0. But this is a safety fix. Whether we should cherry pick it, to get it in before PS0 depends on what the consequences are of not fixing it.

Assigning myself since I have the fix done in #6219 and would do the cherry picking if we decide to do that.

Git blame shows @dckc as the one who should probably investigate this, so I'm assigning to them as well. Feel free to reassign as appropriate of course.

Also assigning @dtribble since he commented on this one and suggested that we not fix it, or at least not fix it this way because of performance concerns.

### LINKED ISSUE #6243: Await safety not yet fixed in swing-store/src/snapStore.js for unbalanced `if`
https://github.com/Agoric/agoric-sdk/blob/293f88efbd01553cc5be4d29c435cd509302b2e7/packages/xsnap/src/replay.js#L221-L225

The triage currently at #6219 would fix it, with the explanation and code

```js
        // This nested await is safe because "all-branches-balanced".
        //
        // This nested synchronous-throw-impossible await at the top level of
        // the then branch is safe because it is balanced by a
        // synchronous-throw-impossible await at the top level of the else
        // branch. In fact the await in the else branch was introduced for
        // that balance, in order to make this one safe.
        await running;
        running = undefined;
      } else {
        // This nested await is safe because "all-branches-balanced".
        //
        // This nested synchronous-throw-impossible await at the top level of
        // the else branch is safe because it is balanced by a
        // synchronous-throw-impossible await at the top level of the then
        // branch. In fact it was introduced for that balance, in order to
        // make the `await` in the then branch safe.
        await null;
      }
```

Most of #6219 consists of commentary or cosmetic code changes, and so can be postponed until after PS0. But this is a safety fix. Whether we should cherry pick it, to get it in before PS0 depends on what the consequences are of not fixing it.

Assigning myself since I have the fix done in #6219 and would do the cherry picking if we decide to do that.

Git blame shows @dckc as the one who should probably investigate this, so I'm assigning to them as well. Feel free to reassign as appropriate of course.

### LINKED ISSUE #6244: Await safety not yet fixed in swing-store/src/snapStore.js for unbalanced `switch`
https://github.com/Agoric/agoric-sdk/blob/293f88efbd01553cc5be4d29c435cd509302b2e7/packages/xsnap/src/replay.js#L260-L261

The triage currently at #6219 would fix it, with the explanation and code

```js
      }
      // This nested await is safe because "top-level-of-for-await".
      // We introduced it to make the unbalanced awaits in the above switch
      // statement safe.
      await null;
      done.push([folder, seq, kind]);
```

Most of #6219 consists of commentary or cosmetic code changes, and so can be postponed until after PS0. But this is a safety fix. Whether we should cherry pick it, to get it in before PS0 depends on what the consequences are of not fixing it.

Assigning myself since I have the fix done in #6219 and would do the cherry picking if we decide to do that.

Git blame shows @dckc as the one who should probably investigate this, so I'm assigning to them as well. Feel free to reassign as appropriate of course.

erights:
Jessie warns of violations of https://github.com/Agoric/agoric-sdk/wiki/No-Nested-Await . Before this PR, a `yarn lint` of agoric-sdk generated 188 of these warnings. This PR triages those warnings. 10 remain warnings and should be fixed subsequent to this PR, preferably by someone who better understands the code these occur in.

For each warning, I either
   * for the 10 that still need to be fixed, I added a TODO FIXME comment and filed one of the issues listed below.
   * turned the rule off wholesale for modules in packages that likely don't need to be assert-safe.
   * fixed the code so that it obeys the No-Nested-Await rule
   * fixed the code so it is actually safe for reasons I explain, even though it still violates No-Nested-Await.
   * realized the code was already safe, sometimes for subtle reasons I explain, even though it violates the No-Nested-Await rule.

For each explanation I inserted, I tried to state the principle I found myself using to figure out that the code is safe. For each of these, I coined a term for this principle, such as "terminal-control-flow", and explain it well the first time I encounter it. Other times, I use the name of the principle, explaining only why it applies to that individual case.

Of my fixes, if they were about correctness, I have filed one of "Fixes" issues listed below that also raises the possibility of cherry picking that fix, so that it could go in ahead of the rest of this PR.

Reviewers, after this PR goes through its first sanity check by you, based on your feedback I will then collect an explanation of the surviving principles and put them someplace more reusable, like our wiki.

Everyone, *please* look at my changes to the code you're most familiar with to see if I broke anything. I tried to be very careful not to, but... . Also, please see if my explanations of why your code was safe, or with my mods is now safe, make sense to you. Even better if you could send me less convoluted replacements than mine for your current code, for example code in which we no longer need to suppress the Jessie await warning.

---

Does not yet fix https://github.com/Agoric/agoric-sdk/issues/6220 , but I expect to include it before merging if it has not already been fixed.

Still needing analysis, but could be addressed in separate PRs:
https://github.com/Agoric/agoric-sdk/issues/6226
https://github.com/Agoric/agoric-sdk/issues/6227
https://github.com/Agoric/agoric-sdk/issues/6228
https://github.com/Agoric/agoric-sdk/issues/6229
https://github.com/Agoric/agoric-sdk/issues/6230
https://github.com/Agoric/agoric-sdk/issues/6231
https://github.com/Agoric/agoric-sdk/issues/6232
https://github.com/Agoric/agoric-sdk/issues/6233
https://github.com/Agoric/agoric-sdk/issues/6234
https://github.com/Agoric/agoric-sdk/issues/6235

This PR would fix, if not already fixed by cherry picking from this PR:
Fixes https://github.com/Agoric/agoric-sdk/issues/6237
Fixes https://github.com/Agoric/agoric-sdk/issues/6239
Fixes https://github.com/Agoric/agoric-sdk/issues/6240
Fixes https://github.com/Agoric/agoric-sdk/issues/6242
Fixes https://github.com/Agoric/agoric-sdk/issues/6243
Fixes https://github.com/Agoric/agoric-sdk/issues/6244

### DIFF
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -4,30 +4,13 @@ import path from 'path';
 
 import { resolve as resolveModuleSpecifier } from 'import-meta-resolve';
 import { assert, details as X } from '@agoric/assert';
+import { allValues } from '@agoric/internal';
 import bundleSource from '@endo/bundle-source';
 
 import '../types-ambient.js';
 import { insistStorageAPI } from '../lib/storageAPI.js';
 import { initializeKernel } from './initializeKernel.js';
 import { kdebugEnable } from '../lib/kdebug.js';
-
-/**
- * @param {X[]} xs
- * @param {Y[]} ys
- * @returns {[X, Y][]}
- * @template X, Y
- */
-const zip = (xs, ys) => xs.map((x, i) => [x, ys[i]]);
-const { keys, values, fromEntries } = Object;
-/**
- * @param {Record<string, Promise<V>>} obj
- * @returns {Promise<Record<string, V>>}
- * @template V
- */
-const allValues = async obj => {
-  const vals = await Promise.all(values(obj));
-  return fromEntries(zip(keys(obj), vals));
-};
 
 const bundleRelative = rel =>
   bundleSource(new URL(rel, import.meta.url).pathname);
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -24,8 +24,10 @@ const { keys, values, fromEntries } = Object;
  * @returns {Promise<Record<string, V>>}
  * @template V
  */
-const allValues = async obj =>
-  fromEntries(zip(keys(obj), await Promise.all(values(obj))));
+const allValues = async obj => {
+  const vals = await Promise.all(values(obj));
+  return fromEntries(zip(keys(obj), vals));
+};
 
 const bundleRelative = rel =>
   bundleSource(new URL(rel, import.meta.url).pathname);
@@ -172,7 +174,14 @@ export function loadBasedir(basedir, options = {}) {
  */
 async function resolveSpecFromConfig(referrer, specPath) {
   try {
-    return new URL(await resolveModuleSpecifier(specPath, referrer)).pathname;
+    // This nested await is safe because "terminal-control-flow".
+    //
+    // Some code does execute after this one, synchronously or asynchronously.
+    // That would normally be unsafe, but all that code is clearly not stateful.
+    // Its correctness is independent on timing.
+    // eslint-disable-next-line @jessie.js/no-nested-await
+    const base = await resolveModuleSpecifier(specPath, referrer);
+    return new URL(base).pathname;
   } catch (e) {
     if (e.code !== 'MODULE_NOT_FOUND' && e.code !== 'ERR_MODULE_NOT_FOUND') {
       throw e;
@@ -235,7 +244,21 @@ export async function loadSwingsetConfigFile(configPath) {
       configPath,
       `file:///${process.cwd()}/`,
     ).toString();
+    // This nested await is safe because "not-my-problem".
+    //
+    // It does occur nested with a control flow branch. But it is at top
+    // level within that branch, followed by another await also at
+    // top level within that branch. Therefore, this await introduces
+    // no unsafety beyond that potentially caused by the next await.
+    // eslint-disable-next-line @jessie.js/no-nested-await
     await normalizeConfigDescriptor(config.vats, referrer, true);
+    // This nested await is safe because "synchonous-throw-impossible".
+    //
+    // normalizeConfigDescriptor is an async function. The only argument
+    // expression that might in theory throw is `config.bundles`, which we
+    // are confident could not actually throw. Even if it could, we'd still
+    // be safe because "terminal-control-flow". The catch body is not stateful.
+    // eslint-disable-next-line @jessie.js/no-nested-await
     await normalizeConfigDescriptor(config.bundles, referrer, false);
     // await normalizeConfigDescriptor(config.devices, referrer, true); // TODO: represent devices
     assert(config.bootstrap, X`no designated bootstrap vat in ${configPath}`);
--- a/packages/SwingSet/src/kernel/vat-warehouse.js
+++ b/packages/SwingSet/src/kernel/vat-warehouse.js
@@ -174,7 +174,18 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
         break;
       }
       logStartup(`provideVatKeeper for vat ${name} as vat ${vatID}`);
-      // eslint-disable-next-line no-await-in-loop
+      // Nested await safe because "terminal-combined-control-flow".
+      //
+      // `ensureVatOnline` is an async function, so this turn boundary always
+      // happens at the end of each iteration. However, the control-flow is
+      // still unbalanced. If the loop iterates zero times, we proceed to
+      // the next code synchronously. Else we proceed asynchronously.
+      // However, the next loop is terminal, and each iteration of that loop
+      // resemble each iteration of this one. Considered together, in all
+      // cases the first iteration of one of these loops will happen first,
+      // if there are any. And if there are more iterations from either loop,
+      // a turn boundary separated each iteration from the next.
+      // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
       await ensureVatOnline(vatID, recreate);
       numPreloaded += 1;
     }
@@ -186,7 +197,8 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
         break;
       }
       logStartup(`provideVatKeeper for dynamic vat ${vatID}`);
-      // eslint-disable-next-line no-await-in-loop
+      // Nested await safe because "terminal-control-flow"
+      // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
       await ensureVatOnline(vatID, recreate);
       numPreloaded += 1;
     }
@@ -391,6 +403,12 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
     // worker may or may not be online
     if (ephemeral.vats.has(vatID)) {
       try {
+        // This nested await is safe because "terminal-control-flow".
+        //
+        // For these purposes, we do not consider `console.debug` to be
+        // significantly stateful. That said, no stateful code in this
+        // function executes after this line.
+        // eslint-disable-next-line @jessie.js/no-nested-await
         await evict(vatID);
       } catch (err) {
         console.debug('vat termination was already reported; ignoring:', err);
--- a/packages/agoric-cli/src/cosmos.js
+++ b/packages/agoric-cli/src/cosmos.js
@@ -85,6 +85,8 @@ export default async function cosmosMain(progname, rawArgs, powers, opts) {
   }
 
   if (popts.pull) {
+    // This await is safe because "terminal-control-flow"
+    // eslint-disable-next-line @jessie.js/no-nested-await
     const exitStatus = await pspawn('docker', ['pull', IMAGE]);
     if (exitStatus) {
       return exitStatus;
--- a/packages/agoric-cli/src/deploy.js
+++ b/packages/agoric-cli/src/deploy.js
@@ -1,3 +1,8 @@
+// IIUC the purpose of the agoric-cli package, we do not need to worry
+// about await safety for any of the code in this package.
+// TODO someone who understands this package better should verify this.
+/* eslint-disable @jessie.js/no-nested-await */
+
 /* global process setTimeout setInterval clearInterval */
 /* eslint-disable no-await-in-loop */
 
--- a/packages/deploy-script-support/src/installInPieces.js
+++ b/packages/deploy-script-support/src/installInPieces.js
@@ -37,6 +37,21 @@ export const installInPieces = async (
       log(
         `waiting for ${inFlightAdditions.length} (~${approxBytesInFlight}B) additions...`,
       );
+      // This await is safe both because "initial-control-flow"
+      // and "terminal-control-flow"
+      //
+      // "initial-control-flow" is the time reversal of
+      // "terminal-control-flow". There is always a turn boundary before each
+      // iteration of a for-await-of loop. This await occurs in one branch
+      // of an unbalanced if within the loop body. But it happens before
+      // anything stateful executes within this iteration, so all subsequent
+      // code in the loop body only executes after a turn boundary.
+      // It is also safe because "terminal-control-flow" because non of the
+      // code in the remainder of the loop body is potentially stateful.
+      // The for-await-of loop protects all code after the loop anyway,
+      // because even in the zero iteration case it will always take a
+      // turn boundary.
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await Promise.all(inFlightAdditions);
       approxBytesInFlight = 0;
       inFlightAdditions = [];
--- a/packages/inter-protocol/src/proposals/demoIssuers.js
+++ b/packages/inter-protocol/src/proposals/demoIssuers.js
@@ -286,6 +286,11 @@ export const connectFaucet = async ({
             case Stake.symbol:
               return bldIssuerKit;
             default: {
+              // This nested await is safe because "terminal-control-flow".
+              //
+              // This switch is the last statement within the enclosing
+              // function.
+              // eslint-disable-next-line @jessie.js/no-nested-await
               const { mint, issuer, brand } = await provideCoin(
                 name,
                 vats.mints,
@@ -304,6 +309,8 @@ export const connectFaucet = async ({
         // Use the bank layer for BLD, IST.
         if (issuerName === Stake.symbol || issuerName === Stable.symbol) {
           const purse = E(bank).getPurse(brand);
+          // This nested await is safe because "terminal-control-flow"
+          // eslint-disable-next-line @jessie.js/no-nested-await
           await E(purse).deposit(payment);
         } else {
           toFaucet = [
@@ -491,6 +498,8 @@ export const fundAMM = async ({
       async issuerName => {
         switch (issuerName) {
           case Stable.symbol: {
+            // This nested await is safe because "terminal-control-flow"
+            // eslint-disable-next-line @jessie.js/no-nested-await
             const [issuer, brand] = await Promise.all([
               centralIssuer,
               centralBrand,
@@ -625,6 +634,9 @@ export const fundAMM = async ({
       let fromCentral;
 
       if (brandsWithPriceAuthorities.includes(brand)) {
+        // TODO FIXME This code should be refactored to make this
+        // await checkably safe, or to remove it, or to record here
+        // why it is actually safe.
         ({ toCentral, fromCentral } = await E(ammPublicFacet)
           .getPriceAuthorities(brand)
           .catch(_e => {
--- a/packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js
+++ b/packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js
@@ -196,9 +196,22 @@ const start = async zcf => {
         trace('exiting processTranches loop');
         return;
       }
+      // This await almost certainly does not do what was intended.
+      // `oncePerBlock()` returns an async generator, not a promise,
+      // and does nothing observably effectful to anyone not pulling on
+      // the generator, which this code drops. AFAICT, this code is
+      // equivalent to `await null;`. It imposes a non-delayed turn boundary
+      // and does nothing else.
+      // See https://github.com/Agoric/agoric-sdk/issues/6220
       await oncePerBlock();
 
       // Determine the max allowed tranche size
+      //
+      // This nested await is safe because "terminal-control-flow".
+      //
+      // It occurs at the top level of the body of a terminal top level
+      // while.
+      // eslint-disable-next-line @jessie.js/no-nested-await
       const { Secondary: poolCollateral, Central: poolCentral } = await E(
         amm,
       ).getPoolAllocation(toSell.brand);
@@ -223,6 +236,11 @@ const start = async zcf => {
 
       // this could use a unit rather than the tranche size to run concurrently
       /** @type PriceQuote */
+      // This nested await is safe because "terminal-control-flow".
+      //
+      // It occurs at the top level of the body of a terminal top level
+      // while.
+      // eslint-disable-next-line @jessie.js/no-nested-await
       const oracleQuote = await E(priceAuthority).quoteGiven(
         tranche,
         debtBrand,
@@ -300,6 +318,8 @@ const start = async zcf => {
       const collateralBefore = debtorSeat.getAmountAllocated('In');
       const proceedsBefore = debtorSeat.getAmountAllocated('Out');
 
+      // This nested await is safe because "top-of-for-await"
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await sellTranche(debtorSeat, collateral, debt, oracleLimit);
 
       const proceedsAfter = debtorSeat.getAmountAllocated('Out');
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -465,7 +465,16 @@ const helperBehavior = {
 
         // The rest of this method will not happen until after a quote is received.
         // This may not happen until much later, when the market changes.
-        // eslint-disable-next-line no-await-in-loop
+        //
+        // This nested await is safe because "terminal-control-flow"
+        //
+        // It occurs at top level of the body of a top level terminal
+        // while. But do note that the enclosing function is an asynchronous
+        // generator function and that there is a `yield` within the loop.
+        // I don't think that causes any additional worries, but I have not
+        // thought about it enough.
+        //
+        // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
         const quote = await E(ephemera.outstandingQuote).getPromise();
         ephemera.outstandingQuote = null;
         // When we receive a quote, we check whether the vault with the highest
@@ -486,7 +495,10 @@ const helperBehavior = {
         }
       }
     }
+
     for await (const next of eventualLiquidations()) {
+      // This nested await is safe because "top-of-for-await"
+      // eslint-disable-next-line @jessie.js/no-nested-await
       await facets.helper.liquidateAndRemove(next);
       trace('price check liq', state.collateralBrand, next && next[0]);
     }
--- a/packages/solo/src/add-chain.js
+++ b/packages/solo/src/add-chain.js
@@ -1,3 +1,8 @@
+// IIUC the purpose of the solo package, we do not need to worry
+// about await safety for any of the code in this package.
+// TODO someone who understands this package better should verify this.
+/* eslint-disable @jessie.js/no-nested-await */
+
 /* global process */
 import fetch from 'node-fetch';
 import crypto from 'crypto';
--- a/packages/vats/src/vat-bank.js
+++ b/packages/vats/src/vat-bank.js
@@ -43,7 +43,9 @@ const makePurseController = (
       let updateRecord = await balanceNotifier.getUpdateSince();
       while (updateRecord.updateCount) {
         yield updateRecord.value;
-        // eslint-disable-next-line no-await-in-loop
+        // This nested await is safe because "terminal-control-flow".
+        //
+        // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
         updateRecord = await balanceNotifier.getUpdateSince(
           updateRecord.updateCount,
         );
--- a/packages/xsnap/src/replay.js
+++ b/packages/xsnap/src/replay.js
@@ -208,7 +208,11 @@ export async function replayXSnap(
    */
   async function runSteps(rd, steps) {
     const folder = rd.path;
-    for (const step of steps) {
+    // Changed from `for of` to `for await of` to make awaits within
+    // loop body safer. A `for await of` will correctly process a synchronous
+    // iterator, but it still unconditionally separates the iterations
+    // with turn boundaries.
+    for await (const step of steps) {
       const parts = step.match(/(\d+)-([a-zA-Z]+)\.(dat|json)$/);
       if (!parts) {
         throw Error(`expected 0001-abc.dat; got: ${step}`);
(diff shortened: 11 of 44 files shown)


packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -171,7 +173,14 @@ export function loadBasedir(basedir, options = {}) {
  */
 async function resolveSpecFromConfig(referrer, specPath) {
   try {
-    return new URL(await resolveModuleSpecifier(specPath, referrer)).pathname;
+    // This nested await is safe because "terminal-control-flow".
+    //
+    // Some code does execute after this one, synchronously or asynchronously.
+    // That would normally be unsafe, but all that code is clearly not stateful.
+    // Its correctness is independent on timing.
```
erights, line 164:
```suggestion
    // Its correctness is independent of timing.
```

packages/SwingSet/src/kernel/vat-warehouse.js
```diff
@@ -168,14 +168,26 @@ export function makeVatWarehouse(kernelKeeper, vatLoader, policyOptions) {
...
+      // `ensureVatOnline` is an async function, so this turn boundary always
+      // happens at the end of each iteration. However, the control-flow is
+      // still unbalanced. If the loop iterates zero times, we proceed to
+      // the next code synchronously. Else we proceed asynchronously.
+      // However, the next loop is terminal, and each iteration of that loop
+      // resemble each iteration of this one. Considered together, in all
+      // cases the first iteration of one of these loops will happen first,
+      // if there are any. And if there are more iterations from either loop,
+      // a turn boundary separated each iteration from the next.
+      // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
```
dckc, line 188: I concur with the justifications in this file, but yes let's do refer to the full explanations by reference rather than having them inline.

packages/deploy-script-support/src/installInPieces.js
```diff
@@ -38,6 +38,21 @@ export const installInPieces = async (
...
+      // iteration of a for-await-of loop. This await occurs in one branch
+      // of an unbalanced if within the loop body. But it happens before
+      // anything stateful executes within this iteration, so all subsequent
+      // code in the loop body only executes after a turn boundary.
+      // It is also safe because "terminal-control-flow" because non of the
+      // code in the remainder of the loop body is potentially stateful.
+      // The for-await-of loop protects all code after the loop anyway,
+      // because even in the zero iteration case it will always take a
+      // turn boundary.
+      // eslint-disable-next-line @jessie.js/no-nested-await
```
dckc, line 54: concur.

packages/inter-protocol/src/proposals/demoIssuers.js
```diff
@@ -625,6 +634,9 @@ export const fundAMM = async ({
       let fromCentral;
 
       if (brandsWithPriceAuthorities.includes(brand)) {
+        // TODO FIXME This code should be refactored to make this
+        // await checkably safe, or to remove it, or to record here
+        // why it is actually safe.
```
dckc, line 639: concur with the findings in this file.

packages/swing-store/src/snapStore.js
```diff
@@ -151,11 +191,35 @@ export function makeSnapStore(
...
+      // of an exported function, which we should therefore be defensive
+      // against. If it throws, `unlink(tmp)` could be reached synchronously,
+      // whereas otherwise it would be reached asynchronously. Because
+      // `unlink` is also a parameter of an exported function, we cannot
+      // locally verify it to be stateless. In fact, from the name, we
+      // expect it to be stateful. By changing the awaited expression to an
+      // IIAFE, we ensure that the `unlink` can only be reached asynchronously.
+      //
+      // eslint-disable-next-line @jessie.js/no-nested-await
+      await (async () => rename(tmpFilePath, target))();
```
dckc: I'm not sure what to say about this file. It may be that keeping all this commentary inline is cost-effective.

packages/vats/src/vat-bank.js
```diff
@@ -43,7 +43,9 @@ const makePurseController = (
       let updateRecord = await balanceNotifier.getUpdateSince();
       while (updateRecord.updateCount) {
         yield updateRecord.value;
-        // eslint-disable-next-line no-await-in-loop
+        // This nested await is safe because "terminal-control-flow".
```
dckc, line 46: concur with the findings in this file

packages/agoric-cli/src/cosmos.js
```diff
@@ -85,6 +85,8 @@ export default async function cosmosMain(progname, rawArgs, powers, opts) {
   }
 
   if (popts.pull) {
+    // This await is safe because "terminal-control-flow"
```
dtribble, line 88: I disagree here. Since the line below the `if` is a recursive call, if `opts.dockerTag && ! popts.pull` then this will infinite loop within a turn.

dtribble: That's not something an attacker can supply, so this is low risk.

packages/xsnap/src/xsnap.js
```diff
@@ -200,7 +206,30 @@ export function xsnap(options) {
...
       } else if (message[0] === QUERY) {
-        await messagesToXsnap.next(await handleCommand(message.subarray(1)));
+        // This nested await is safe because "not-my-problem".
+        //
+        // This nested await introduces no unsafety beyond that introduced
+        // by the next await, since it is earlier within the same control
+        // flow branch, and is top level within that branch.
+        // eslint-disable-next-line @jessie.js/no-nested-await
+        const array = await handleCommand(message.subarray(1));
+        // This nested await is ALMOST safe because "terminal-control-flow".
```
dckc, line 220: paging @kriskowal . PTAL.

dckc (commented):
I looked at a handful of files as noted.

packages/inter-protocol/src/proposals/demoIssuers.js
```diff
@@ -625,6 +634,9 @@ export const fundAMM = async ({
       let fromCentral;
 
       if (brandsWithPriceAuthorities.includes(brand)) {
+        // TODO FIXME This code should be refactored to make this
```
dtribble, line 637: FYI `demoIssuers` isn't production code

packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -197,9 +197,22 @@ const start = async zcf => {
         trace('exiting processTranches loop');
         return;
       }
+      // This await almost certainly does not do what was intended.
```
dtribble, line 199: Dang it! That seems correct. How would it get integrated so that a long-lived generator is pulled on and awaited once per block

packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js
```diff
@@ -197,9 +197,22 @@ const start = async zcf => {
...
+      // and does nothing observably effectful to anyone not pulling on
+      // the generator, which this code drops. AFAICT, this code is
+      // equivalent to `await null;`. It imposes a non-delayed turn boundary
+      // and does nothing else.
+      // See https://github.com/Agoric/agoric-sdk/issues/6220
       await oncePerBlock();
 
       // Determine the max allowed tranche size
+      //
+      // This nested await is safe because "terminal-control-flow".
```
dtribble, line 210: These are all part of the related design pattern of using for `for await` and an async generator. We should have one comment to that effect and then suppress the overall nested await check in this file

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -466,7 +466,16 @@ const helperBehavior = {
...
         // The rest of this method will not happen until after a quote is received.
         // This may not happen until much later, when the market changes.
-        // eslint-disable-next-line no-await-in-loop
+        //
+        // This nested await is safe because "terminal-control-flow"
+        //
+        // It occurs at top level of the body of a top level terminal
+        // while. But do note that the enclosing function is an asynchronous
+        // generator function and that there is a `yield` within the loop.
+        // I don't think that causes any additional worries, but I have not
```
dtribble, line 474: I believe it does not. This must be used within a `for await`, which awaits at the top of the loop (and therefore at each `yield`). So a `yield` i effectively an `await`.

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -466,7 +466,16 @@ const helperBehavior = {
 
         // The rest of this method will not happen until after a quote is received.
         // This may not happen until much later, when the market changes.
-        // eslint-disable-next-line no-await-in-loop
+        //
+        // This nested await is safe because "terminal-control-flow"
```
dtribble, line 469: This should refer to the `for/await/generator/while pattern` that is used in several places for vaults.

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -487,7 +496,10 @@ const helperBehavior = {
         }
       }
     }
+
     for await (const next of eventualLiquidations()) {
+      // This nested await is safe because "top-of-for-await"
```
dtribble, line 500: We should just add the top-level of a `for-await` as an allowed context for `awaits`

packages/inter-protocol/src/proposals/demoIssuers.js
```diff
@@ -288,6 +288,11 @@ export const connectFaucet = async ({
             case Stake.symbol:
               return bldIssuerKit;
             default: {
+              // This nested await is safe because "terminal-control-flow".
```
dtribble, line 289: Not production code in any case.

dtribble (commented):
Comments so far

packages/xsnap/src/xsnap.js
```diff
@@ -200,7 +206,30 @@ export function xsnap(options) {
...
+        //   * This branch is taken and the awaited expression throws.
+        //     In that case, the function exits would executing and further
+        //     code.
+        //   * This branch is taken and does not throw. In that case,
+        //     it causes a turn boundary before the next iteration.
+        //   * Another branch of the conditional is taken, skipping
+        //     this statement and proceeding syncronously to the next
+        //     iteration. The await at the top of the loop ensures that the
+        //     remainder of the loop only happens after a turn boundary.
+        //     The only remaining unsafety is that the awaited expression of
```
dtribble, line 232: All the other branches are terminal (return or throw) and so this unsafety does not actaully occur here.

packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js
erights: If we turn the current Jessie warning off for the whole file, then we won't notice if an await slips in that isn't safe for this reason. Until we have more static check distinctions, I think we should still suppress them individually, citing the safety principle for each one. But obviously, we can pull out all further commentary.

packages/inter-protocol/src/vaultFactory/vaultManager.js
erights: Oh yes! Once we start fiddling with the static checks, the top of a for-await body should be jessie-safe.

The terminal-control-flow reason for safety is the most commonly used, should be straightforward to automate for the common cases, and covers some of the most useful cases, like many awaits in top level terminal try blocks. (comment from our slack repeated here for the record).

packages/inter-protocol/src/proposals/demoIssuers.js
erights: Good to know. That's what I suspected from the name. But OTOH it is in a src/ directory, not test/ or demo/.
OTTH, as example code it needs to set the right precedents for how people will write production code.

packages/xsnap/src/replay.js
```diff
@@ -210,7 +210,11 @@ export async function replayXSnap(
    */
   async function runSteps(rd, steps) {
     const folder = rd.path;
-    for (const step of steps) {
+    // Changed from `for of` to `for await of` to make awaits within
+    // loop body safer. A `for await of` will correctly process a synchronous
+    // iterator, but it will unconditionally separate the iterations with
+    // with turn boundaries.
+    for await (const step of steps) {
```
dtribble, line 215: This potentially adds a a *lot* of awaits. (2 additional awaits per step). While they are cheap individually, steps could I think be quite large if e.g., you are trying to catch up on a restart.

dtribble: Loading vats on replay is a serialized process, and so I think the risk of the  current code is low. I recommend NOT including this change without a lot more testing than we can perform at this time. And so it shoudl just wait until it gets refactored to not need this kind of layering.

packages/swing-store/src/snapStore.js
dtribble: Simpler might be:
```javascript
await null;
await rename(tmpFilePath, target);

File: packages/swing-store/src/snapStore.js [hunk 18]
```diff
@@ -186,15 +250,51 @@ export function makeSnapStore(
...
+        // Rather, it potentially postpones when the `close()`s happen
+        // in case `sync()` throws. With the IIAFE, if this line is
+        // executed, then the `close()`s would always happen in a later turn.
+        //
+        // But this nested await is not actually safe because it is nested in
+        // an `if`, the other branch of which proceeds synchrounously. The
+        // `if` is not terminal because of the stateful `close()`s afterward.
+        // On the else branch, those `close()`s would happen synchronously.
+        // eslint-disable-next-line @jessie.js/no-nested-await
+        await (async () => destination.sync())();
```
dtribble: as above. This seems too mysterious. I'd live with it if it were more efficient though :)

dtribble (commented):
There is one element that I think should NOT be checked in, and a few alternate suggestions. I reviewed all changes, except for the "IIUC" examples in e.g., `solo` directories.

packages/agoric-cli/src/cosmos.js
erights: interesting!

packages/swing-store/src/snapStore.js
erights: The semantics are more different. Might be ok in this case, but...

The original code always calls rename in the same turn that was already executing. For all I know that atomicity may be important. My change preserves that. Your's puts an interleaving point before the rename.

erights: OTOH, there is an earlier await almost immediately above it. In this case, your extra await would only break up the atomicity with the `resolve` which seems unlikely to cause a problem, but would still need looking into.

packages/swing-store/src/snapStore.js
erights: As above, in this case it happens to be safe to introduce a new interleaving point before the `sync()` call, because the immediately preceding stateful code is another await. But it is not a technique that works in general, so we should be careful about making it a habit.

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -25,8 +25,10 @@ const { keys, values, fromEntries } = Object;
  * @returns {Promise<Record<string, V>>}
  * @template V
  */
-const allValues = async obj =>
-  fromEntries(zip(keys(obj), await Promise.all(values(obj))));
+const allValues = async obj => {
```
turadg: we have several of these. I pushed [667ad5c](https://github.com/Agoric/agoric-sdk/pull/6219/commits/667ad5c9cafc64e922a637702751f69b259d0f9f) to DRY them

packages/agoric-cli/src/deploy.js
```diff
@@ -1,3 +1,8 @@
+// IIUC the purpose of the agoric-cli package, we do not need to worry
+// about await safety for any of the code in this package.
+// TODO someone who understands this package better should verify this.
```
turadg, line 3: confirmed. let's disable for the whole package. [397c702](https://github.com/Agoric/agoric-sdk/pull/6219/commits/397c7026f81579bc9b69c5588d88d700fb1e57b7)

packages/solo/src/add-chain.js
```diff
@@ -1,3 +1,8 @@
+// IIUC the purpose of the solo package, we do not need to worry
```
turadg, line 1: disabled for package. [0163fee](https://github.com/Agoric/agoric-sdk/pull/6219/commits/0163fee0ecc93d935fe8560ebb107e3bad838de7)

turadg (commented):
The await safety triaging seemed to have enough review so I reviewed how it's communicated. 

>  I will then collect an explanation of the surviving principles and put them someplace more reusable, like our wiki.

+1 to having an index of the principles to reference. Ideally this would be something you can pop into within the IDE. I think something like this would accomplish that: `/** @see {SafeAwait['terminal-control-flow']} */`. Short of that, I think the URL should be stable and go directly to the explanation, which both rule out the Github wiki. E.g. `shorten.er/await-safety#terminal-control-flow`

For the "IIUC" suppressions, I think you did so I pushed package level disabling (agoric-cli, solo).

I also noticed the change to `allValues` was just one of the few instances so I pushed a change to consolidate them.

.eslintrc.cjs
```diff
@@ -78,16 +78,16 @@ module.exports = {
...
-        // promise.then or get a reviewed exception.  “sometimes awaits” is a
-        // bug farm for particularly pernicious bugs in which you can combine
-        // two correct pieces of code to have emergent incorrect behavior.
-        // It’s absolutely critical for shared service code. That means
-        // contracts, but it also means kernel components that are used by
-        // multiple clients. So we enable it throughout the repo and exceptions
-        // are code-reviewed.
+        // The rule is “no nested awaits” but the architectural goal is “no
+        // possibility of ‘awaits sometimes but not always’”. That is our
+        // architectural rule for vats which require distributed determinism. If
```
dckc: How is distributed determinism relevant? I thought the issue was correctness.

turadg: Thanks for reviewing this part I forgot to mention. I was trying to distinguish why CLI and solo didn't need this linking.

My understanding is it's only relevant when there things can happen in different turns, ergo blockchain context.

What's a better description of when the lint rule needs to be enforced?

packages/xsnap/src/xsnap.js
```diff
@@ -200,7 +206,30 @@ export function xsnap(options) {
...
+        // by the next await, since it is earlier within the same control
+        // flow branch, and is top level within that branch.
+        // eslint-disable-next-line @jessie.js/no-nested-await
+        const array = await handleCommand(message.subarray(1));
+        // This nested await is ALMOST safe because "terminal-control-flow".
+        // This one appears nested in an unbalanced conditional, so we need
+        // to consider three cases:
+        //   * This branch is taken and the awaited expression throws.
+        //     In that case, the function exits would executing and further
+        //     code.
```
kriskowal, line 225: I can’t parse this sentence. Perhaps s/would/without/, s/any/any/?

packages/xsnap/src/xsnap.js
```diff
@@ -200,7 +206,30 @@ export function xsnap(options) {
...
+        //   * This branch is taken and does not throw. In that case,
+        //     it causes a turn boundary before the next iteration.
+        //   * Another branch of the conditional is taken, skipping
+        //     this statement and proceeding syncronously to the next
+        //     iteration. The await at the top of the loop ensures that the
+        //     remainder of the loop only happens after a turn boundary.
+        //     The only remaining unsafety is that the awaited expression of
+        //     that initial await may therefore be reached synchronously or
+        //     asynchronously.
```
kriskowal, line 234: The expression awaited at the top of the cthuloop, `messagesFromXsnap.next(undefined)`, just manipulates an entangled pair of async promise queues, putting undefined on one and getting the next promise from the other. I’m not sure whether this affects your analysis.

.eslintrc.cjs
erights: What are the purposes of the `cli` and `solo` packages? What are the worst case consequences of an interleaving bug in the code of these packages?

Depending on the answers, we may want to leave these warnings on for these packages and carry through the triage for the problematic awaits is these packages.

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -171,7 +156,14 @@ export function loadBasedir(basedir, options = {}) {
  */
 async function resolveSpecFromConfig(referrer, specPath) {
   try {
-    return new URL(await resolveModuleSpecifier(specPath, referrer)).pathname;
+    // This nested await is safe because "terminal-control-flow".
+    //
+    // Some code does execute after this one, synchronously or asynchronously.
+    // That would normally be unsafe, but all that code is clearly not stateful.
+    // Its correctness is independent on timing.
```
erights, line 164:
```suggestion
    // Its correctness is independent of timing.
```

.eslintrc.cjs
dckc: Different turns happen any time code doesn't complete sequentially; every `await` or `promise.then(...)` or I/O callback introduces another turn. No blockchain context needed.

My understanding is that the rule is relevant any time we want correct code; that is: to avoid a pattern that is known to cause bugs. I could be wrong. @erights ? @dtribble ?

packages/agoric-cli/src/cosmos.js
erights: See https://github.com/Agoric/agoric-sdk/issues/6235

packages/inter-protocol/src/vaultFactory/liquidateIncrementally.js
erights: See https://github.com/Agoric/agoric-sdk/issues/6220

packages/swing-store/src/snapStore.js
erights: See https://github.com/Agoric/agoric-sdk/issues/6237

packages/swing-store/src/snapStore.js
erights: See https://github.com/Agoric/agoric-sdk/issues/6239

packages/swing-store/src/snapStore.js
```diff
@@ -262,7 +362,19 @@ export function makeSnapStore(
...
+        // therefore safe by itself, this arrow function also calls
+        // `createGzip()`, which we cannot locally verify does not
+        // synchronously throw. Thus, we made the arrow function async,
+        // so that it cannot throw regardless. This seems to be consistent
+        // with its purpose, and with the likely assumptions the program
+        // author made when placing it here.
+        //
+        atomicWriteInRoot(baseName, async tmpGzPath =>
           filter(tmpSnapPath, createGzip(), tmpGzPath, { flush: true }),
         ),
```
erights: See https://github.com/Agoric/agoric-sdk/issues/6240

packages/xsnap/src/replay.js
erights: See https://github.com/Agoric/agoric-sdk/issues/6242

.eslintrc.cjs
turadg: > What are the purposes of the cli and solo packages? What are the worst case consequences of an interleaving bug in the code of these packages?

CLI is ergonomics for an end-user (or dev) to craft and send commands. Solo is is the solo Vat runner, which tmk is for testing. I can't think of any dire consequences to bugs in either.

> Different turns happen any time code doesn't complete sequentially

Where can I read more about this? I've never encountered the term before Agoric and can't find it in web search.

If this PR needs to land for PS0, @erights feel free to remove any of my changes. I can follow up with them later.

packages/SwingSet/src/controller/initializeSwingset.js
erights: Much better, thanks!

.eslintrc.cjs
erights: > I can't think of any dire consequences to bugs in either.

I think this is the important bit. It means the warnings should stay on, but fixing those is much less urgent. Perhaps those should only be classified "bug" without "security"?

erights: > Where can I read more about this? I've never encountered the term before Agoric and can't find it in web search.

https://papers.agoric.com/papers/concurrency-among-strangers/abstract/ , coauthored by myself, @dtribble , and Jonathan Shapiro.

This paper expanded into Part III of https://papers.agoric.com/papers/robust-composition/abstract/ , though the briefer standalone paper is certainly adequate for present purposes. I recommend you read my whole thesis in any case ;)

dckc: Though they don't seem to use the term "turn", the concept is well known in the JS world, yes? A "turn" is just one turn through the event loop.

For example, https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop:

> At some point during the event loop, the runtime starts handling the messages on the queue, starting with the oldest one. To do so, the message is removed from the queue and its corresponding function is called with the message as an input parameter. As always, calling a function creates a new stack frame for that function's use.
>
> The processing of functions continues until the stack is once again empty. Then, the event loop will process the next message in the queue (if there is one).

dckc: p.s. I do agree that the whole "Robust Composition" thesis is worth reading in due course. Though it's over 200 pages, it's not dense material. Much of it is stuff you almost certainly already know.

dckc: > Solo is is the solo Vat runner, which tmk is for testing.

Solo powers `ag-solo`; it's it's intended, in due course, for dapp back-ends that may hold significant assets. @dtribble recently gave an example of a gateway to an online game world, I think.

turadg: sounds like we should just drop my two eslint/comment changing commits. I figure that's easiest to do by interactive rebase and force push. To avoid clobbering any concurrent work I'll refrain unless you request it @erights

erights: I'll do it when I get back to working on this PR. Thanks.

packages/SwingSet/src/controller/initializeSwingset.js
turadg: It looks like your force pushes since then have erased https://github.com/Agoric/agoric-sdk/commit/667ad5c9cafc64e922a637702751f69b259d0f9f

I've pushed it again. @erights mind pulling f902c5d27 your local branch?

erights:
Reviewers, this PR is in good shape. The embargo has been lifted. AFAIK, nothing significant changed since the reviews above. But, this is the PR that I'm most nervous about needing to keep up to date to follow changes on master. I'd much rather merge it before there are substantial changes on master to reduce the coordination work. Thanks!

dckc (changes requested):
I haven't reviewed the whole thing... when I saw separate issues filed for the various parts, I assumed this PR would get split up similarly. But I gather you're interested to merge this soonish.

Please let's do refer to the full explanations by reference rather than having them inline (e.g. vat-warehouse.js)

And we should have a test for each fix. Perhaps that's not cost-effective, but let's make that decision explicitly; no fair skipping the "testing considerations" part of the PR template :)

I'm not sure which reviewer is in a position to approve this whole thing. I am not. Perhaps a PR per package? hm... a dozen PRs is a lot of tedium. Maybe separate commits? Also tedious, and doesn't clarify who approved what very well.

packages/SwingSet/src/kernel/vat-warehouse.js
turadg: e.g. https://gist.github.com/turadg/e7d7645789218173a578508ae7640365#file-terminal-combined-control-flow

turadg:
> Please let's do refer to the full explanations by reference rather than having them inline (e.g. vat-warehouse.js)

Start of that https://gist.github.com/turadg/e7d7645789218173a578508ae7640365 (now at https://github.com/endojs/Jessie/pull/93)

> assumed this PR would get split up similarly

For splitting up:
@warner https://github.com/Agoric/agoric-sdk/pull/6526
- swingset
- cosmic-swingset

@turadg https://github.com/Agoric/agoric-sdk/pull/6528
- agoric-cli
- inter-protocol
- internal
- web-components

@michaelfig https://github.com/Agoric/agoric-sdk/pull/6529
- cache
- deploy-script-support
- solo

@mhofman https://github.com/Agoric/agoric-sdk/pull/6530
- telemetry

@dckc https://github.com/Agoric/agoric-sdk/pull/6531
- vats
- xsnap

@Chris-Hibbert https://github.com/Agoric/agoric-sdk/pull/6533
- zoe

erights:
Closing in favor of the unbundling at https://github.com/Agoric/agoric-sdk/pull/6219#issuecomment-1287482865

PR #6221, opened 2022-09-16
feat(cosmic-swingset): Depth first scheduling

### LINKED ISSUE #6210: Depth first processing of inbound actions

Currently all inbound actions in a block are delivered in one chunk into SwingSet devices, which results in items added to the SwingSet runqueue. Since the SwingSet runqueue is singular and FIFO, external deliveries contend with each other since every message to other vat goes to the back of the runqueue.

## Description of the Design

While we plan on better handling this future scheduling work (multiple swingset queues, LIFO-ish queue or "flow" based processing, etc.), one simple solution for the PSM use case (no internal fan out) is to simply pluck cosmos actions one by one, and run SwingSet to "completion" before moving on to the next action.

Since the existing action queue is designed to be flushed by the end of the block, this requires a new queue of actions stored in cosmic-swingset JS side, stored in the swingStore host section. It also requires some changes to the run policy logic to allow multiple runs of the SwingSet controller.

## Security Considerations

One risk of run-to-completion or LIFO queueing is that an infinite message loop would starve subsequent actions / messages. While the current contract code is trusted and will not trigger such loops, we can be somewhat defensive against this and allow new actions to be queued if the run queue is below a certain threshold. That means the action processing would not be strictly FIFO, which we don't actually care about. However this may not be a sufficient mitigation if one could trigger multiple message loops.

We also need to be careful regarding hangover inconsistency, and make sure that whatever data we use to make decision on whether to wait to deliver a new action into swingset or not is persisted appropriately.

## Test Plan

Yes

mhofman:
closes: #6210

## Description

This PR further refactors cosmic-swingset and add an `inboundQueue` which buffers the cosmos inbound actions on the JS side, processing them one by one by running swingset to quiescence after delivering each action.

### Security Considerations

This approach is not safe against infinite message loops, which would cause a Denial Of Service. However the current contracts are trusted to not behave like that.

### Testing Considerations

Tested locally with the loadgen. Load testing in progress on ollinet.

### DIFF
--- a/packages/cosmic-swingset/src/launch-chain.js
+++ b/packages/cosmic-swingset/src/launch-chain.js
@@ -238,20 +238,21 @@ export async function launch({
   // and disables commit/abort.
 
   const inboundQueuePrefix = getHostKey('inboundQueue.');
-  const inboundStorage = harden({
+  const inboundQueueStorage = harden({
     get: key => {
       const val = kvStore.get(inboundQueuePrefix + key);
-      return val && JSON.parse(val);
+      return val ? JSON.parse(val) : undefined;
     },
-    set: (key, value) =>
-      value === undefined
-        ? inboundStorage.delete(key)
-        : kvStore.set(inboundQueuePrefix + key, JSON.stringify(value)),
+    set: (key, value) => {
+      value !== undefined ||
+        assert.fail(`value in inboundQueue must be defined`);
+      kvStore.set(inboundQueuePrefix + key, JSON.stringify(value));
+    },
     delete: key => kvStore.delete(inboundQueuePrefix + key),
     commit: () => {}, // disable
     abort: () => {}, // disable
   });
-  const inboundQueue = makeQueue(inboundStorage);
+  const inboundQueue = makeQueue(inboundQueueStorage);
 
   // Not to be confused with the gas model, this meter is for OpenTelemetry.
   const metricMeter = metricsProvider.getMeter('ag-chain-cosmos');
@@ -483,8 +484,12 @@ export async function launch({
       return runPolicy.shouldRun();
     }
 
-    // Then we allow the timer to poll, which might push work onto the
-    // kernel run-queue, and gets the first cycle.
+    // We update the timer device at the start of each block, which might push
+    // work onto the end of the kernel run-queue (if any timers were ready to
+    // wake), where it will be followed by actions triggered by the block's
+    // swingset transactions.
+    // If the queue was empty, the timer work gets the first "cycle", and might
+    // run to completion before the block actions get their own cycles.
     const addedToQueue = timer.poll(blockTime);
     console.debug(
       `polled; blockTime:${blockTime}, h:${blockHeight}; ADDED =`,
@@ -584,7 +589,7 @@ export async function launch({
           blockHeight,
           runNum,
         });
-        await bootstrapBlock();
+        await processAction(action.type, bootstrapBlock);
         controller.writeSlogObject({
           type: 'cosmic-swingset-run-finish',
           blockHeight,
--- a/packages/cosmic-swingset/src/launch-chain.js
+++ b/packages/cosmic-swingset/src/launch-chain.js
@@ -128,6 +128,24 @@ async function buildSwingset(
   return { controller, mb, bridgeInbound, timer };
 }
 
+/**
+ * @typedef {import('@agoric/swingset-vat').RunPolicy & {
+ *   shouldRun(): boolean;
+ *   remainingBeans(): number;
+ * }} ChainRunPolicy
+ */
+
+/**
+ * @typedef {object} BeansPerUnit
+ * @property {bigint} blockComputeLimit
+ * @property {bigint} vatCreation
+ * @property {bigint} xsnapComputron
+ */
+
+/**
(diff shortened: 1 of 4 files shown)


mhofman:
@warner I've re-arranged in your WIP commits into singular ones (best reviewed independently), and tweaked a few things, in particular:
- Made sure the `runTime` console output stayed correct. We previously dropped measurement of `runKernel`.
- Restore console logging if bootstrap fails.
- Fixed the queue storage to do serialization

Here is a rough diff (with base shift clutter) of what changed: https://github.com/Agoric/agoric-sdk/compare/b0e606850..6210-depth-first-scheduling

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -574,23 +550,41 @@ export async function launch({
...
           // And now we actually process the queued actions down here, during
-          // END_BLOCK, but still reentrancy-protected
+          // END_BLOCK, but still reentrancy-protected.
+
+          provideInstallationPublisher();
 
-          // Process our begin, queued actions, and end.
-          await processAction(beginBlockAction); // BEGIN_BLOCK
+          // First we allow the timer to poll, which might push work onto the
+          // kernel run-queue, and gets the first cycle.
```
warner: This comment, in this commit ("disentangle launch-chain some more") and the next, is not accurate: timer work gets pushed onto the *end* of the run-queue, never getting the first cycle. Let's change it to:

> We update the timer device at the start of each block, which might push work onto the end of the kernel run-queue (if any timers were ready to wake), where it will be followed by actions triggered by the block's swingset transactions.

Then, in the commit where we switch to cycles, add:

> If the queue was empty, the timer work gets the first "cycle", and might run to completion before the block actions get their own cycles.

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -208,6 +209,28 @@ export async function launch({
...
   };
 
+  // makeQueue() thinks it should commit/abort, but the kvStore doesn't provide
+  // those ('commit' is reserved for flushing the block buffer). Furthermore
+  // the kvStore only deals with string values.
+  // We create a storage wrapper that adds a prefix to keys, serializes values,
+  // and disables commit/abort.
+
+  const inboundQueuePrefix = getHostKey('inboundQueue.');
+  const inboundStorage = harden({
```
warner: let's name this `inboundQueueStorage`

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -208,6 +209,28 @@ export async function launch({
...
+    get: key => {
+      const val = kvStore.get(inboundQueuePrefix + key);
+      return val && JSON.parse(val);
+    },
+    set: (key, value) =>
+      value === undefined
```
warner: Does the `makeQueue` code ever do `set(key, undefined)`? I thought it does explicit `delete`.

The JSON serialization surprises me, but I guess the old version was relying upon the `call` send-to-cosmos function to serialize the value along with the rest of the request?

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -508,11 +567,23 @@ export async function launch({
...
           type: 'cosmic-swingset-bootstrap-block-start',
           blockTime,
         });
-        await processAction(action.type, bootstrapBlock);
+        controller.writeSlogObject({
+          type: 'cosmic-swingset-run-start',
+          blockHeight,
+          runNum,
+        });
+        await bootstrapBlock();
```
warner: The last commit ("Depth first scheduling") changes this line from `await processAction(action.type, bootstrapBlock)` to `await bootstrapBlock()`. Should that change have happened in an earlier commit? It seems to me that this specific commit should only be adding the `cosmic-swingset-run-start` logs around a call that should have been the same as in the previous commit. (The end result might be the same, and seems correct, but the path by which the commits get there might be slightly off).

warner (approved):
Looks good, a few small comments to consider.

packages/cosmic-swingset/src/launch-chain.js
mhofman: I don't know what values are passed in, and I need to handle `undefined` somehow. I think deleting is just as correct as writing `''`.

Yeah I dug through and some layer other than the queue was doing serialization before

packages/cosmic-swingset/src/launch-chain.js
mhofman: Good catch, it should not have happened at all. I'm using `processAction` to `console.log` any error. The `runTime` measurement in this case are simply dropped.

packages/cosmic-swingset/src/launch-chain.js
warner: The `kvStore` can (well, is supposed to) be able to store empty strings just fine. `set(key, '')` is entirely different than `delete(key)`.

It *might* be the case that `set(key, undefined)` is the same as `delete(key)`. I wouldn't like that in a swingstore API, but I suspect at least one of our storage layers behaves that way.

Do you know what was trying to store `undefined`? I know the Actions are objects, and for sure that needs to be serialized into the swingstore's `key: string, value: string` API, but I'd be surprised for any of the Actions to be an `undefined`.

PR #6222, opened 2022-09-16
fix(psm): "mint" limits for IST rather than anchor

### LINKED ISSUE #6223: PSM minting limit should be wrt Minted brand
## Describe the bug
For PSM, the MintingLimit parameter was specified in terms of the anchor brand. That makes the MintingLimit incomparable with among minting sources, including other PSMs.

## Expected behavior
Minting limits are expressed in the minted currency, e.g., IST.

## Platform Environment
PSM

dtribble:
closes: #6223

## Description

The MintingLimit parameter was specified in terms of the anchor brand. That makes the MintingLimit incomparable with among minting sources, including other PSMs.

Incidentally reduce the default MINT_LIMIT to a small value for testing. Governance is required to raise it to a higher amount.

### Security Considerations

The minting limits are an important tool for managing risk in IST. While the minting limit using the anchor amounts coudl be managed, using different brands for different PSM or other minting sources would make it so that we were not minting against a standard limit.

Documenting minting limits as expressed in IST will be much clearer.

### Testing Considerations

A test was added to confirm that it mints against the minted brand rather than the anchor brand.

### DIFF
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -119,7 +119,9 @@ export const start = async (zcf, privateArgs, baggage) => {
   const feePool = provideEmptyZcfSeat('feePoolSeat');
   const stage = provideEmptyZcfSeat('stageSeat');
 
-  let mintedOutstanding = AmountMath.makeEmpty(stableBrand);
+  let mintedPoolBalance = provide(baggage, 'mintedPoolBalance', () =>
+    AmountMath.makeEmpty(stableBrand),
+  );
 
   let totalAnchorProvided = provide(baggage, 'totalAnchorProvided', () =>
     AmountMath.makeEmpty(anchorBrand),
@@ -138,6 +140,7 @@ export const start = async (zcf, privateArgs, baggage) => {
       harden({
         anchorPoolBalance: anchorPool.getAmountAllocated('Anchor', anchorBrand),
         feePoolBalance: feePool.getAmountAllocated('Minted', stableBrand),
+        mintedPoolBalance,
         totalAnchorProvided,
         totalMintedProvided,
       }),
@@ -149,19 +152,19 @@ export const start = async (zcf, privateArgs, baggage) => {
    * @param {Amount<'nat'>} toMint
    */
   const assertUnderLimit = toMint => {
-    const mintedAfter = AmountMath.add(mintedOutstanding, toMint);
+    const mintedAfter = AmountMath.add(mintedPoolBalance, toMint);
     AmountMath.isGTE(params.getMintLimit(), mintedAfter) ||
       assert.fail(X`Request would exceed mint limit`);
   };
 
   const burnMinted = toBurn => {
     stableMint.burnLosses({ Minted: toBurn }, stage);
-    mintedOutstanding = AmountMath.subtract(mintedOutstanding, toBurn);
+    mintedPoolBalance = AmountMath.subtract(mintedPoolBalance, toBurn);
   };
 
   const mintMinted = toMint => {
     stableMint.mintGains({ Minted: toMint }, stage);
-    mintedOutstanding = AmountMath.add(mintedOutstanding, toMint);
+    mintedPoolBalance = AmountMath.add(mintedPoolBalance, toMint);
   };
 
   /**
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -118,6 +118,9 @@ export const start = async (zcf, privateArgs, baggage) => {
   const anchorPool = provideEmptyZcfSeat('anchorPoolSeat');
   const feePool = provideEmptyZcfSeat('feePoolSeat');
   const stage = provideEmptyZcfSeat('stageSeat');
+
+  let mintedOutstanding = AmountMath.makeEmpty(stableBrand);
+
   let totalAnchorProvided = provide(baggage, 'totalAnchorProvided', () =>
     AmountMath.makeEmpty(anchorBrand),
   );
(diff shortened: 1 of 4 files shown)


packages/inter-protocol/src/psm/psm.js
```diff
@@ -143,17 +146,24 @@ export const start = async (zcf, privateArgs, baggage) => {
...
-  const assertUnderLimit = given => {
-    const anchorAfterTrade = AmountMath.add(
-      anchorPool.getAmountAllocated('Anchor', anchorBrand),
-      given,
-    );
-    AmountMath.isGTE(params.getMintLimit(), anchorAfterTrade) ||
+  const assertUnderLimit = toMint => {
+    const mintedAfter = AmountMath.add(mintedOutstanding, toMint);
+    AmountMath.isGTE(params.getMintLimit(), mintedAfter) ||
       assert.fail(X`Request would exceed mint limit`);
```
dckc, line 157: we don't have a `fail` template tag yet? darn.

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -338,6 +338,27 @@ test('limit', async t => {
...
+  const anchorPerMinted = makeRatio(50n, anchor.brand, 100n, minted.brand);
+  const driver = await makePsmDriver(t, { anchorPerMinted });
+
+  trace('test going over limit');
+  const giveTooMuch = anchor.make(MINT_LIMIT);
+  const seat1 = await driver.swapAnchorForMintedSeat(giveTooMuch);
+  t.throwsAsync(() => E(seat1).getOfferResult(), {
+    message: 'Request would exceed mint limit',
+  });
+  trace('limit is enforced on the Minted rather than Anchor');
```
dckc: use ava's message arg instead of trace?

```suggestion
  t.throwsAsync(() => E(seat1).getOfferResult(), {
    message: 'Request would exceed mint limit',
  }, 'limit is enforced on the Minted rather than Anchor');
```

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -338,6 +338,27 @@ test('limit', async t => {
...
+
+  trace('test right at limit');
+  const give = anchor.make(MINT_LIMIT / 2n);
+  const paymentPs = await driver.swapAnchorForMinted(give);
+  await paymentPs;
+  trace('test at minted limit works');
```
dckc:
```suggestion
  await t.notThrowsAsync(paymentPs, 'test that minted limit works');
```

at least spell out "that" :)

dckc (approved):
looks good.
a couple minor suggestions to consider. not critical.

packages/inter-protocol/src/psm/psm.js
turadg: what would that do?

Incidentally, why is this a template at all when it's a simple string? While we're here, how about adding some feedback: (`X``Request would exceed mint limit ${mintLimit.value}`)

packages/inter-protocol/test/psm/test-psm.js
turadg: +1

packages/inter-protocol/src/psm/psm.js
```diff
@@ -118,6 +118,9 @@ export const start = async (zcf, privateArgs, baggage) => {
   const anchorPool = provideEmptyZcfSeat('anchorPoolSeat');
   const feePool = provideEmptyZcfSeat('feePoolSeat');
   const stage = provideEmptyZcfSeat('stageSeat');
+
+  let mintedOutstanding = AmountMath.makeEmpty(stableBrand);
```
turadg, line 122: not clear to me what this value is for. Making `burnLosses` throw an error if `toBurn` is greater than what we've yet minted?

Shouldn't `burnMinted` take the exact same amount as `mintMinted` ? 

As is, I'm not sure this value and those two new functions are in improvement.

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -311,7 +311,7 @@ test('simple trades', async t => {
 });
 
 test('limit', async t => {
-  const { mintLimit, anchor } = t.context;
+  const { anchor } = t.context;
```
turadg, line 316: for the same reasons you stopped getting `mintLimit` from the context, it should be removed from the context maker https://github.com/Agoric/agoric-sdk/blob/a1dedeae72908fda45afcb6038d76f8359adc8de/packages/inter-protocol/test/psm/test-psm.js#L129

packages/inter-protocol/test/psm/test-psm.js
```diff
@@ -338,6 +338,27 @@ test('limit', async t => {
   // t.throwsAsync(() => await E(seat1).getOfferResult());
```
turadg, line 340: Is "TODO Offer result should be an error" something we should have for PSO?

Same question for the other TODO in this test: https://github.com/Agoric/agoric-sdk/blob/a1dedeae72908fda45afcb6038d76f8359adc8de/packages/inter-protocol/test/psm/test-psm.js#L328

packages/inter-protocol/test/psm/test-psm.js
turadg: I feel more strongly about this being an actual test statement so included it in my requested changes

turadg (changes requested):

turadg (approved):
I pushed the changes I requested so I'm approving.

I had also Requested Changes on a question about `mintedOutstanding` but I figured it out while following the code in my IDE so I deleted the comment.

packages/inter-protocol/src/psm/psm.js
dckc: The fall template tag would abbreviate assert.fail and X:

```js
x > 0 || fail`x too small: ${x}`
```

dckc: It was something I saw Dean say in chat somewhere.

turadg: nifty

packages/inter-protocol/test/psm/test-psm.js
dtribble: `trace` here is wrapped around AVA's log (in tests). This provided consistent reporting along with any `trace` statements in the stuff being tested (which are using `console`)

packages/inter-protocol/test/psm/test-psm.js
dtribble: that's pre-existing. I don't know what was intended here.

turadg (approved):
+1 changes since last review

PR #6224, opened 2022-09-16
fix(SwingSet): Apply netstring limit to xsnap workers

### LINKED ISSUE #6150: Netstring limit does not apply to xsnap worker
## Describe the bug
https://github.com/Agoric/agoric-sdk/pull/5655 added logic to restrict netstring message size, but that code only applied to node workers which are not used in production. There is an [upcoming change in the endo netstring package](https://github.com/endojs/endo/pull/1277) which is the one used by `xsnap.js` to handle the communication.

We'll need to plumb the message size there instead.

mhofman:
closes: #6150
refs: #5655

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -31,6 +31,8 @@ import {
   initializeSwingset,
 } from './initializeSwingset.js';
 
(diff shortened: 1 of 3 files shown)


kriskowal (approved):

FUDCo (approved):
Yep.

PR #6236, opened 2022-09-16
fix: hoist suspicious nested await to top level

### LINKED ISSUE #6231: Await safety uncertain in SwingSet/src/controller/initializeSwingset.js
https://github.com/Agoric/agoric-sdk/blob/a1dedeae72908fda45afcb6038d76f8359adc8de/packages/SwingSet/src/controller/initializeSwingset.js#L333

The triage at https://github.com/Agoric/agoric-sdk/pull/6219 currently classifies this as 

    // TODO FIXME This code should be refactored to make this
    // await checkably safe, or to remove it, or to record here
    // why it is actually safe.
    //
    // As the default value of a destructuring, this `await` is bizarrely
    // conditional. It'll happen if `kernelBundles` is either absent or
    // undefined in `initializationOptions`, in which case we'll take a turn
    // boundary here. Otherwise, we'll continue into the subsequent code
    // synchronously. I could make this safe by introducing an extra turn
    // boundary, but for this code I'm scared to do that without understanding
    // the code better first.

Git blame shows @warner @FUDCo and @dckc as the ones who should probably investigate this, so I'm assigning to them. Feel free to reassign as appropriate of course.

FUDCo:
Fixes #6231

I'm fairly confident this was OK as it was, but just in the interest of paranoia I'm hoisting the await anyway.

@erights please verify that I've done the await voodoo right.

### DIFF
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -329,8 +329,10 @@ export async function initializeSwingset(
       assert.fail(X`unknown manager type ${defaultManagerType}`);
   }
 
+  const obtainKernelBundles = async () =>
+    initializationOptions.kernelBundles || buildVatAndDeviceBundles();
+  const kernelBundles = await obtainKernelBundles();
   const {
-    kernelBundles = await buildVatAndDeviceBundles(),
     verbose,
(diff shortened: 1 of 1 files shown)


erights (approved):
Yes, that's perfect. LGTM

PR #6246, opened 2022-09-16
fix(cosmic-swingset): add inboundQueue metrics

### LINKED ISSUE #6245: add metrics for cosmic-swingset "inboundQueue"

#6210 adds a new `inboundQueue` to hold the cosmic-swingset events (like `bridgeInbound`) that have been included in a cosmos block, but not yet delivered into swingset. The new queueing approach seeks to drain the swingset kernel's run-queue before transferring the next inbound event from `inboundQueue` into the kernel. This improves the best-case latency for economic operations (like a PSM trade), and generally means that the time to completion for these operations is a fixed function of the number of operations ahead of you in the `inboundQueue` (previously, the 95-ish steps of each operation submitted after yours would be interleaved with the 95-ish steps of your own, so their submissions would slow down yours even though you were nominally first in line).

The analogy I've been using is the waiting line for a roller coaster at a big amusement park. With the interleaved cranks, you must go through the line 95 times, so it's more of a herd than a line. With the serialized operations, we can realistically put signs on the line that say "the waiting time from this point is: 10 minutes".

It'd be nice to publish the queue depth and execution rate for `inboundQueue`. With that information, an external dashboard could estimate how long a newly-submitted PSM operation would take.

## Description of the Design

@arirubinstein says we want to submit this as a prometheus "gauge", with a callback each time we add or remove something from the queue.

This will be built on top of @mhofman 's PR #6221 , in some combination of `launch-chain.js` and `make-queue.js`. It might be as simple as adding up/down counts to the places in `runKernel` where we do `inboundQueue.push()` and each iteration of  `for (const a of inboundQueue.consumeall())`.

## Security Considerations

None I can think of.

## Performance Considerations

This will add two metric submissions to each `bridgeInbound` event (as well as all other actions that feed the kernel).

## Test Plan

Manual examination of the metrics being emitted (since `launch-chain.js` and cosmic-swingset are not well covered by unit-scale tests).

warner:
This adds a few keys to the stats that we report to e.g. Prometheus, to report on the size/add/remove counters for the new `inboundQueue`.

closes #6245

### DIFF
--- a/packages/cosmic-swingset/src/kernel-stats.js
+++ b/packages/cosmic-swingset/src/kernel-stats.js
@@ -199,17 +199,56 @@ export function makeSlogCallbacks({ metricMeter, attributes = {} }) {
 }
 
 /**
+ * Create a metrics manager for the 'inboundQueue' structure, which
+ * can be scaped to report current length, and the number of
+ * increments and decrements. This must be created with the initial
+ * length as extracted from durable storage, but after that we assume
+ * that we're told about every up and down, so our RAM-backed shadow
+ * 'length' will remain accurate.
+ *
+ * Note that the add/remove counts will get reset at restart, but
+ * Prometheus/etc tools can tolerate that just fine.
+ *
+ * @param {number} initialLength
+ */
+export function makeInboundQueueMetrics(initialLength) {
+  let length = initialLength;
+  let add = 0;
+  let remove = 0;
+
+  return harden({
+    incStat: (delta = 1) => {
+      length += delta;
+      add += delta;
+    },
+
+    decStat: (delta = 1) => {
+      length -= delta;
+      remove += delta;
+    },
+
+    getStats: () => ({
+      cosmic_swingset_inbound_queue_length: length,
+      cosmic_swingset_inbound_queue_add: add,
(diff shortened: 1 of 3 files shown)


warner:
I've manually tested this:
```
$ make scenario2-setup scenario2-run-chain-psm  # in one shell
# wait for it to get started

$ curl -s http://localhost:9464/metrics | grep cosmic  # in second shell
cosmic_swingset_inbound_queue_add 0 1663365468693
cosmic_swingset_inbound_queue_length 0 1663365468693
cosmic_swingset_inbound_queue_remove 0 1663365468693

$ make ACCT_ADDR=$(cat ./t1/8000/ag-cosmos-helper-address) AGORIC_POWERS=SMART_WALLET provision-acct fund-acct

$ curl -s http://localhost:9464/metrics | grep cosmic
cosmic_swingset_inbound_queue_add 2 1663365516811
cosmic_swingset_inbound_queue_length 0 1663365516811
cosmic_swingset_inbound_queue_remove 2 1663365516811
```

warner:
The way this is hacked into `kernel-stats.js` is a big gross, sorry. Hopefully later we can do some cleanup.

mhofman (approved):

Issue #4585, opened 2022-02-17
Telemetry for the Kernel


Tartuffo:
Per our [Observability meeting](https://docs.google.com/document/d/1TuaCxiLPqygQsYQ86VdoWQgcBur4Pi7RR2QY0hYbLw0/edit#) on Feb 17, here are areas of the kernel to instrument:

1. Length of run queue(s)
2. Number of computrons and cranks consumed in a block (probably cosmic swingset)
3. Are we approaching hard limit of computrons / crank (could be an infinite loop)
4. Amount of time doing swingset computation per block
5. Vat manager: number of Vats extant, number of Vats created and dropped / evicted.
6. Key data structures: object table, promise table, clist (from kernel stats)
7. Data publish device - requests / second
8. We should identify all of our resource checks, and put metrics around those.
9. How frequently BOYD is called, and how long BOYD took

Tartuffo:
@warner FYI

mhofman:
Not swingset stat per se, but exposing in Honeycomb how many cosmos deliveries are made into swingset (bridge, wallet, etc.)

mhofman:
Maybe combined with how many cranks happened in a block (currently a proxy is simply to count distinct crankNum and plot over time)

arirubinstein:
Adding a few request:
- All log events are structured json
- All log events include the closest trace id, allowing us to correlate logs to a specific span
- Generalize names of spans so our logs can break out spans into services (e.g. vats, major components of the kernel)
- Traces to be pushed via the opentelemetry tracing platform

Tartuffo:
We've got enough done for PSM RC0, anything else we need we can stand up in a separate follower.

PR #6247, opened 2022-09-16
Telemetry enhancement

FUDCo:
Various telemetry related improvements in support of #4585 

- The slog-to-otel converter has been extensively cleaned up and supports the latest version of slog output we are producing
- Kernel stats are output as slog entries as part of finishing up each block
- The slogulator tool has been updated to account for the many and various changes and enhancements to slog output since the last time it was worked on
- The swingset-runner tool can now produce an Open Telemetry feed of its slog if so requested

### DIFF
--- a/packages/cosmic-swingset/src/launch-chain.js
+++ b/packages/cosmic-swingset/src/launch-chain.js
@@ -475,6 +475,10 @@ export async function launch({
       await crankScheduler(runPolicy);
       const remainingBeans = runPolicy.remainingBeans();
       controller.writeSlogObject({
+        type: 'kernel-stats',
+        stats: controller.getStats(),
+      });
+      controller.writeSlogObject({
         type: 'cosmic-swingset-run-finish',
         blockHeight,
         runNum,
--- a/packages/swingset-runner/src/main.js
+++ b/packages/swingset-runner/src/main.js
@@ -610,6 +610,11 @@ export async function main() {
     if (verbose) {
       log('==> running block');
     }
+    controller.writeSlogObject({
+      type: 'cosmic-swingset-end-block-start',
+      blockHeight: blockNumber,
+      blockTime: blockStartTime,
+    });
     while (requestedSteps > 0) {
       requestedSteps -= 1;
       try {
@@ -642,6 +647,15 @@ export async function main() {
       await swingStore.commit();
     }
     const blockEndTime = readClock();
+    controller.writeSlogObject({
+      type: 'kernel-stats',
+      stats: controller.getStats(),
+    });
+    controller.writeSlogObject({
+      type: 'cosmic-swingset-end-block-finish',
+      blockHeight: blockNumber,
+      blockTime: blockEndTime,
+    });
     if (forceGC) {
       engineGC();
     }
--- a/packages/swingset-runner/src/main.js
+++ b/packages/swingset-runner/src/main.js
@@ -15,6 +15,7 @@ import { buildLoopbox } from '@agoric/swingset-vat/src/devices/loopbox/loopbox.j
 import engineGC from '@agoric/swingset-vat/src/lib-nodejs/engine-gc.js';
 
 import { initSwingStore, openSwingStore } from '@agoric/swing-store';
+import { makeSlogSender } from '@agoric/telemetry';
 
 import { dumpStore } from './dumpstore.js';
 import { auditRefCounts } from './auditstore.js';
@@ -57,7 +58,8 @@ FLAGS may be:
   --logstats       - log kernel stats after each block
   --logall         - log kernel stats, block times, memory use, and disk space
   --logtag STR     - tag for stats log file (default "runner")
-  --slog FILE      - write swingset log to FILE
+  --slog FILE      - write swingset slog to FILE
+  --teleslog       - transmit a slog feed to the local otel collector
   --forcegc        - run garbage collector after each block
   --batchsize N    - set BATCHSIZE to N cranks (default 200)
   --verbose        - output verbose debugging messages as it runs
@@ -162,6 +164,7 @@ export async function main() {
   let logStats = false;
   let logTag = 'runner';
   let slogFile = null;
+  let teleslog = false;
   let forceGC = false;
   let verbose = false;
   let doDumps = false;
@@ -219,6 +222,9 @@ export async function main() {
         break;
       case '--slog':
         slogFile = argv.shift();
+        break;
+      case '--teleslog':
+        teleslog = true;
         break;
       case '--config':
         configPath = argv.shift();
@@ -396,6 +402,37 @@ export async function main() {
       }
     }
   }
+  let slogSender;
+  if (teleslog || slogFile) {
+    const slogEnv = {
+      ...process.env,
+      SLOGFILE: slogFile,
+    };
+    const slogOpts = {
+      stateDir: dbDir,
+      env: slogEnv,
+    };
+    if (slogFile) {
+      slogEnv.SLOGSENDER = '';
+    }
+    if (teleslog) {
+      const {
+        SLOGSENDER: envSlogSender,
+        TELEMETRY_SERVICE_NAME = 'agd-cosmos',
+        OTEL_EXPORTER_OTLP_ENDPOINT = 'http://localhost:4318',
+      } = process.env;
+      const SLOGSENDER = [
+        ...(envSlogSender ? envSlogSender.split(',') : []),
+        '@agoric/telemetry/src/otel-trace.js',
+      ].join(',');
+      slogEnv.SLOGSENDER = SLOGSENDER;
+      slogEnv.OTEL_EXPORTER_OTLP_ENDPOINT = OTEL_EXPORTER_OTLP_ENDPOINT;
+      slogOpts.serviceName = TELEMETRY_SERVICE_NAME;
+    }
+    // eslint-disable-next-line @jessie.js/no-nested-await
+    slogSender = await makeSlogSender(slogOpts);
+    runtimeOptions.slogSender = slogSender;
+  }
   let bootstrapResult;
   const hostStorage = {
     kvStore: swingStore.kvStore,
@@ -556,6 +593,10 @@ export async function main() {
   if (statLogger) {
     statLogger.close();
   }
+  if (slogSender) {
+    // eslint-disable-next-line @jessie.js/no-nested-await
+    await slogSender.forceFlush();
+  }
   controller.shutdown();
 
   function getCrankNumber() {
@@ -611,7 +652,7 @@ export async function main() {
       log('==> running block');
     }
     controller.writeSlogObject({
-      type: 'cosmic-swingset-end-block-start',
+      type: 'cosmic-swingset-run-start',
       blockHeight: blockNumber,
       blockTime: blockStartTime,
     });
@@ -652,7 +693,7 @@ export async function main() {
       stats: controller.getStats(),
     });
     controller.writeSlogObject({
-      type: 'cosmic-swingset-end-block-finish',
+      type: 'cosmic-swingset-run-finish',
       blockHeight: blockNumber,
       blockTime: blockEndTime,
     });
--- a/packages/telemetry/src/slog-to-otel.js
+++ b/packages/telemetry/src/slog-to-otel.js
@@ -13,8 +13,10 @@ import {
 // diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.VERBOSE);
 
 /** @typedef {import('@opentelemetry/api').Span} Span */
+/** @typedef {import('@opentelemetry/api').Link} SpanLink */
 /** @typedef {import('@opentelemetry/api').SpanContext} SpanContext */
 /** @typedef {import('@opentelemetry/api').SpanOptions} SpanOptions */
+/** @typedef {import('@opentelemetry/api').SpanAttributes} SpanAttributes */
 
 const replacer = (_key, value) => {
   if (typeof value === 'bigint') {
@@ -80,6 +82,8 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
   /** @type {Record<string, any>} */
   let currentAttrs = {};
 
+  let isReplaying = false;
+
   const cleanAttrs = attrs => ({
     ...serializeInto(attrs, 'agoric'),
     ...overrideAttrs,
@@ -123,15 +127,17 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         break;
       }
       case 'send': {
+        name = 'message';
         // TODO: The arguments to a method call can be pretty big.
         delete attrs.msg;
         const { methargs, method = extractMethod(methargs) } = message.msg;
-        name = `E(${message.target}).${method}`;
         const slots =
           message.msg.methargs?.slots ?? message.msg.args.slots ?? [];
-        attrs['message.msg.args.slots'] = slots.join(',');
-        attrs['message.msg.method'] = name;
-        attrs['message.msg.result'] = message.msg.result;
+        attrs.details = `E(${message.target}).${method}`;
+        attrs['msg.target'] = message.target;
+        attrs['msg.method'] = method;
+        attrs['msg.slots'] = slots.join(',');
+        attrs['msg.result'] = message.msg.result;
         break;
       }
       case 'notify': {
@@ -167,20 +173,28 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
   };
 
   const makeSpans = () => {
-    /** @type {LegacyMap<string, Span>} */
+    /**
+     * @typedef {object} SpanRecord
+     * @property {Span} span
+     * @property {string} name
+     * @property {string} kind
+     * @property {string} key
+     */
+
+    /** @type {LegacyMap<string, SpanRecord>} */
     // Legacy because spans are not passable
     const keyToSpan = makeLegacyMap('spanKey');
 
-    let lastNamedPopTime;
+    let lastAssertedKindPopTime;
 
-    /** @type {{ span: Span, name: string }[]} */
+    /** @type {SpanRecord[]} */
     const spanStack = [];
 
     const makeSpanKey = keyArray => JSON.stringify(keyArray, replacer);
 
     const sp = harden({
       endAll: endTime => {
-        for (const [_keyName, span] of keyToSpan.entries()) {
+        for (const { span } of keyToSpan.values()) {
           span.end(endTime);
         }
         keyToSpan.clear();
@@ -238,7 +252,8 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         const span = sp.create(name, parent, attrs, options);
         if (keyArray.length) {
           const spanKey = makeSpanKey(keyArray);
-          keyToSpan.init(spanKey, span);
+          const record = { span, name, kind: keyArray[0], key: spanKey };
+          keyToSpan.init(spanKey, record);
         }
         return span;
       },
@@ -269,7 +284,7 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
       end: (key, attrs = {}, errorMessage = undefined) => {
         const keyArray = Array.isArray(key) ? key : [key];
         const spanKey = makeSpanKey(keyArray);
-        const span = keyToSpan.get(spanKey);
+        const { span } = keyToSpan.get(spanKey);
         if (errorMessage) {
           span.setStatus({ code: SpanStatusCode.ERROR, message: errorMessage });
         }
@@ -287,26 +302,74 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
       get: key => {
         const keyArray = Array.isArray(key) ? key : [key];
         const spanKey = makeSpanKey(keyArray);
-        return keyToSpan.get(spanKey);
+        return keyToSpan.get(spanKey).span;
       },
 
-      push: (name, attrs = currentAttrs) => {
-        const span = sp.startNamed(name, [], sp.top(), attrs);
-        spanStack.push({ span, name });
+      /**
+       * @param {string | [string, ...any]} key
+       * @param {object} [options]
+       * @param {string} [options.kind]
+       * @param {string} [options.name]
+       * @param {Record<string, any>} [options.attributes]
+       * @param {SpanLink[]} [options.links]
+       */
+      push: (
+        key,
+        {
+          kind = undefined,
+          name = undefined,
+          attributes = currentAttrs,
+          ...opts
+        } = {},
+      ) => {
+        const keyArray = Array.isArray(key) ? key : [key];
+        kind ??= keyArray[0];
+        name ??= kind;
+        const spanKey = makeSpanKey(keyArray);
+        const span = sp.create(name, sp.top(), attributes, opts);
+        const record = { span, name, kind, key: spanKey };
+        keyToSpan.init(spanKey, record);
+        spanStack.push(record);
+        return span;
       },
 
-      /** @param {string} [assertName] */
-      pop: (assertName = undefined) => {
-        const popped = spanStack.pop();
-        if (assertName !== undefined) {
-          assert.equal(popped?.name, assertName);
-          lastNamedPopTime = now;
+      /**
+       * @param {string | [string, ...any]} [assertKindOrKey]
+       * @param {object} [options]
+       * @param {Record<string, any>} [options.attributes]
+       * @param {string} [options.errorMessage]
+       */
+      pop: (
+        assertKindOrKey = undefined,
+        { attributes = {}, errorMessage = undefined } = {},
+      ) => {
+        const top = spanStack[spanStack.length - 1];
+        if (assertKindOrKey !== undefined) {
+          assert(
+            (typeof assertKindOrKey === 'string' &&
+              top?.kind === assertKindOrKey) ||
+              top?.key === makeSpanKey(assertKindOrKey),
+          );
+          lastAssertedKindPopTime = now;
         }
-        popped?.span.end(lastNamedPopTime);
+        if (top) {
+          if (errorMessage !== undefined) {
+            top.span.setStatus({
+              code: SpanStatusCode.ERROR,
+              message: errorMessage,
+            });
+          }
+          top.span.setAttributes(cleanAttrs(attributes));
+          top.span.end(lastAssertedKindPopTime);
+          spanStack.pop();
+          keyToSpan.delete(top.key);
+        }
       },
 
+      /** @returns {Span | undefined} */
       top: () => spanStack[spanStack.length - 1]?.span,
-      topName: () => spanStack[spanStack.length - 1]?.name,
+      /** @returns {string | undefined} */
+      topKind: () => spanStack[spanStack.length - 1]?.kind,
     });
     return sp;
   };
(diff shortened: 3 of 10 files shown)


packages/telemetry/src/slog-to-otel.js
```diff
@@ -126,12 +130,13 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         // TODO: The arguments to a method call can be pretty big.
         delete attrs.msg;
         const { methargs, method = extractMethod(methargs) } = message.msg;
-        name = `E(${message.target}).${method}`;
         const slots =
           message.msg.methargs?.slots ?? message.msg.args.slots ?? [];
-        attrs['message.msg.args.slots'] = slots.join(',');
-        attrs['message.msg.method'] = name;
-        attrs['message.msg.result'] = message.msg.result;
+        attrs.details = `E(${message.target}).${method}`;
```
warner, line 136: This `E(target).method` sounds more like a `.description` or `.summary` than a `.details` ?

packages/telemetry/src/slog-to-otel.js
```diff
@@ -126,12 +130,13 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+        attrs['msg.target'] = message.target;
```
warner, line 137: I don't know OTEL conventions, but is `msg` an ok top-level name for these swingset-specific things? I don't want to make us type more when drilling down in honeycomb/datadog/etc, but @arirubinstein @michaelfig is there any reason why we should add an `agoric.` or `swingset.` prefix to these attribute names?

packages/telemetry/src/slog-to-otel.js
```diff
@@ -347,60 +406,68 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+      case 'kernel-init-start': {
+        spans.push('init');
+        break;
+      }
+      case 'kernel-init-finish': {
+        spans.pop('init');
+        break;
+      }
+      case 'bundle-kernel-start': {
+        assert.equal(spans.topKind(), 'init');
```
warner, line 425: All of these `assert`s.. this is in `slogSender()`, right? What happens if one of these throws, does the kernel panic? It looks like `writeSlogObject()` does not protect against `slogSender()` throwing, and `writeSlogObject` is called from all over.

warner (approved):
It looks plausible, but I'd like to get @mhofman 's take too before we land this. I don't understand the convertor deeply, nor can I visualize the spans that will get created, and I think he's got a better intuition than me for those.

I'm a little worried that a disagreement between the kernel-side code (emitting slog events in a particular pattern) and the slog-to-otel convertor will cause `slogSender()` to throw, which would cause an exception to happen e.g. in the middle of a syscall, which will (at best) panic the kernel. I don't know how to best address that.. I know we used to have a `catch` that would ignore exceptions happening during logging, I'm not sure where that got to (and of course even if we re-introduce that protection, if the convertor throws once, it will probably be stuck in a completely confused state until at least the next reboot). But I'd like to make sure we aren't increasing the chances of kernel panic with this change. On the good side, we don't expect most validators to turn this on, so the risk is mostly to our own follower/monitoring nodes.

The other vague concern is that the telemetry code is parsing a very specific order+nesting of slog events, which I think means it should be dependent upon a specific version of swingset. With the current monorepo (and no deployed systems) that can afford to be a bit loose, but going forward we'll need to make sure each published version of `swingset` has a well-documented slog-event ordering/nesting, and make a new version of `telemetry` that can parse it correctly. It's a pity that these two components should be so tightly coupled, but I guess that's the price for having such a wide variety of slog events, and wanting to get easy-to-deep-dive spans into diagnostic/analytic tools like honeycomb.

packages/telemetry/src/slog-to-otel.js
FUDCo: I was just following @mhofman's notes.  Any of those is fine with me.

packages/telemetry/src/slog-to-otel.js
FUDCo: All of this ends up getting prefixed with `agoric.` automagically.

packages/telemetry/src/slog-to-otel.js
mhofman: We can definitely bikeshed the name of the common attribute. This should appear on all spans, so terse is probably better.

That said, I think we should look into adding this attribute in the root, and not prefixed by `agoric.`.

packages/telemetry/src/slog-to-otel.js
mhofman: #6325 [added a `try` around senders](https://github.com/Agoric/agoric-sdk/pull/6325/files#diff-31e5a1133de785fc56b949f3acc085e2bb67841025fb012d83f104498046146eR145-R149) if there are multiple senders. Might be worth applying the same case to single slogSender, which honestly won't happen in our deployments since we'll likely always have a flight-recorder and/or a slog file sender.

mhofman (commented):
Review in progress, but first here are answers to @warner's questions.

packages/swingset-runner/src/main.js
```diff
@@ -220,6 +223,9 @@ export async function main() {
       case '--slog':
         slogFile = argv.shift();
         break;
```
mhofman, line 225: Sorry for the churn, and for mising this in #6325, but we need to update `slogFile` handling and plumb it into `process.env.SLOG_FILE` so that `makeSlogSender` can write the slog file again.

packages/swingset-runner/src/main.js
```diff
@@ -396,6 +402,21 @@ export async function main() {
...
+    const {
+      TELEMETRY_SERVICE_NAME = 'agd-cosmos',
+      OTEL_EXPORTER_OTLP_ENDPOINT = 'http://localhost:4318',
+    } = process.env;
+    process.env.OTEL_EXPORTER_OTLP_ENDPOINT = OTEL_EXPORTER_OTLP_ENDPOINT;
+    // eslint-disable-next-line @jessie.js/no-nested-await
+    slogSender = await makeSlogSender({
+      serviceName: TELEMETRY_SERVICE_NAME,
+      stateDir: dbDir,
+      env: process.env,
```
mhofman: I think you'll need to set `SLOGSENDER` in the `env` object passed to `makeSlogSender` to include `'@agoric/telemetry/src/otel-trace.js'`.

packages/swingset-runner/src/main.js
```diff
@@ -396,6 +402,21 @@ export async function main() {
...
+    });
+    runtimeOptions.slogSender = slogSender;
+  }
```
mhofman:
```suggestion
  if (teleslog || slogFile) {
    if (slogFile && forceReset) {
      try {
        fs.unlinkSync(slogFile);
      } catch (e) {
        if (e.code !== 'ENOENT') {
          fail(`${e}`);
        }
      }
    }
    const {
      SLOGSENDER: envSlogSender,
      TELEMETRY_SERVICE_NAME = 'agd-cosmos',
      OTEL_EXPORTER_OTLP_ENDPOINT = 'http://localhost:4318',
    } = process.env;
    const SLOGSENDER = [
      ...new Set([
        ...(envSlogSender ? envSlogSender.split(',') : []),
        '@agoric/telemetry/src/otel-trace.js',
      ]),
    ].join(',');
    // eslint-disable-next-line @jessie.js/no-nested-await
    slogSender = await makeSlogSender({
      serviceName: TELEMETRY_SERVICE_NAME,
      stateDir: dbDir,
      env: {
        ...process.env,
        SLOGSENDER,
        SLOGFILE: slogFile,
        OTEL_EXPORTER_OTLP_ENDPOINT,
      },
    });
    runtimeOptions.slogSender = slogSender;
  }
```

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -712,6 +712,10 @@ export async function launch({
           computedHeight = blockHeight;
           savedBlockTime = blockTime;
         }
+        controller.writeSlogObject({
+          type: 'kernel-stats',
+          stats: controller.getStats(),
```
mhofman: I was wondering, should we put this in `swingset-run` instead of the `end-block`? I'm thinking that per-block kernel stats don't automatically mean much.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -412,45 +479,61 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
           break;
         }
-        const crankSpan = spans.get(getCrankKey());
-        crankSpan.setAttributes(cleanAttrs({ delivery, ...attrs }));
+
+        const name = `deliver-${kd[0]}`;
+        const activeDeliverySpan = spans.push(
+          [name, attrs.vatID, attrs.deliveryNum],
+          { attributes: {} },
+        );
```
mhofman: While I think it's right to push a span here since we need to nest syscall spans under it, I believe we need to avoid starting a span on `crank-start`, at least on delivery cranks (I'm also thinking routing cranks could just be events instead of spans to avoid adding visual noise, and since their duration isn't really meaningful)

packages/telemetry/src/slog-to-otel.js
```diff
@@ -412,45 +479,61 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+        activeDeliverySpan.setAttributes(cleanAttrs({ delivery, ...attrs }));
```
mhofman: Why not set the attributes while `push`ing the span?

packages/telemetry/src/slog-to-otel.js
```diff
@@ -412,45 +479,61 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
             target,
             { methargs, method = extractMethod(methargs), result },
           ] = kd;
+          const details = `E(${target}).${method}`;
+          const msgAttrs = {
+            'msg.target': target,
+            'msg.method': extractMethod(methargs),
+            'msg.slots': methargs.slots.join(','),
+            'msg.result': result,
+            details,
```
mhofman, line 552: Same here, I'm thinking `details` (or whatever bikeshed name) should be a top level concept of our span abstraction along with `name` and `kind`, and for top a level attribute not nested under `agoric.` to be used.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -412,45 +479,61 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+        const activeDeliverySpan = spans.push(
+          [name, attrs.vatID, attrs.deliveryNum],
+          { attributes: {} },
+        );
+        activeDeliverySpan.setAttributes(cleanAttrs({ delivery, ...attrs }));
         if (kd[0] === 'notify') {
           // Track call graph.
           const [_type, [notification]] = kd;
           // TODO: We only track the first notified kernel promise. Perhaps we should do more?
           const [kpId, { state }] = notification;
```
mhofman: Should we not set the `details` to `` `${state === 'rejected' ? 'reject' : 'resolve'}(${kpId})` ``

packages/telemetry/src/slog-to-otel.js
```diff
@@ -412,45 +479,61 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         if (!kd) {
           break;
         }
-        const crankSpan = spans.get(getCrankKey());
-        crankSpan.setAttributes(cleanAttrs({ delivery, ...attrs }));
+
+        const name = `deliver-${kd[0]}`;
```
mhofman: Isn't there a risk that `kd[0]` and `deliver = vd[0]` might differ for some delivery kinds? I know this discrepency already existed, I'm just not sure of the implications.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -412,45 +479,61 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+          );
+          // This is where the message is delivered.
+          // Track call graph.
           let cause;
           if (kernelPromiseToSendingCause.has(result)) {
             cause = JSON.parse(kernelPromiseToSendingCause.get(result));
           } else {
             // Create a new root span.
-            const name = `E(${target}).${method}`;
+            const crankSpan = spans.get(getCrankKey());
```
mhofman, line 564: Since I want to get rid of these generic crank spans, I think we'd need to use the delivery span we just created here.

packages/telemetry/src/slog-to-otel.js
mhofman: Also I'm wondering if this might not be the right place to compute common delivery attributes such as `details` so that all of delivery spans have baseline data (and we have less chance of forgetting to handle new delivery types in the future)

packages/telemetry/src/slog-to-otel.js
```diff
@@ -501,60 +585,156 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
-            const syscall = makeSyscallSpan(
-              name,
-              {
-                target,
-                method,
-                result,
-              },
-              { links },
-            );
+            const syscall = makeSyscallSpan(name, sattrs, { links });
```
mhofman, line 653: `makeSyscallSpan` uses `crankSpan` and not the delivery span as parent, is that on purpose?

(applies to all syscall types, but I can't comment on `makeSyscallSpan` directly)

packages/telemetry/src/slog-to-otel.js
```diff
@@ -501,60 +585,156 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+            }
+            let details;
+            if (resolves.length === 0) {
+              details = `reject(${rejects.join(',')})`;
+            } else if (rejects.length === 0) {
+              details = `resolve(${resolves.join(',')})`;
+            } else {
+              // prettier-ignore
+              details = `resolve(${resolves.join(',')}),reject(${rejects.join(',')})`;
+            }
```
mhofman: Nit to DRY it up:
```suggestion
            const detailsArr = [];
            if (resolves.length) {
              detailsArr.push(`resolve(${resolves.join(',')})`);
            }
            if (rejects.length) {
              detailsArr.push(`reject(${rejects.join(',')})`);
            }
            const details = `s-${vatId}: ${detailsArr.join(';')}`;
```

packages/telemetry/src/slog-to-otel.js
mhofman: I also don't believe `name` will be a sufficient differentiator to add duration to these spans once we handle `timestamps` in `deliver-results`, but then I'm actually not sure we'll still be able to create the syscall spans here since the start time should be the time registered by the xsnap worker and not the kernel.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -501,60 +585,156 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
-        switch (ksc[0]) {
+        switch (syscallType) {
           case 'send': {
             const [
               _tag,
               target,
               { methargs, method = extractMethod(methargs), result },
             ] = ksc;
-            const name = `E(${target}).${method}`;
+            const details = `E(${target}).${method}`;
```
mhofman: Why not have the same prefixed format as the other syscalls `s-${vatID}...: `. ?

Note: I think the prefix should be computed by some common logic, maybe in `makeSyscallSpan`? Especially since we'll likely have to defer the span creation until deliver-result.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -571,52 +751,48 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
       }
-      case 'bundle-kernel-start': {
-        spans.push(`bundle-kernel`);
-        break;
-      }
-      case 'bundle-kernel-finish': {
-        spans.pop('bundle-kernel');
+      case 'kernel-stats': {
+        const stats = /** @type {SpanAttributes} */ ({});
+        Object.entries(slogAttrs.stats).forEach(
```
mhofman: Why not use `cleanAttrs` here instead, or at least `serializeInto` if the goal was to avoid the potential `overrideAttrs` ?

mhofman (commented):
Thanks so much for taking this work on.

I think the place where I got stuck last and that isn't addressed in this PR is how to best create a single "delivery" span that isn't nested under a generic "crank" span.
I also believe that simply creating named syscall spans is not going to work as soon as we want to process the timestamps for them.

Finally I'd like to see the handling of `details` being less adhoc and more core to our own span abstraction.

I'm not sure how much of this we should do right away, and what should be deferred to future PRs.

packages/cosmic-swingset/src/launch-chain.js
FUDCo: Yup.  That makes sense.

packages/telemetry/src/slog-to-otel.js
warner: good, and that's on trunk now, so I won't worry about kernel-panic risk coming from any changes here

packages/telemetry/src/slog-to-otel.js
FUDCo: I ended up having to unify the crank span with the deliver span by starting and ending the deliver span from `crank-start` and `crank-finish`, and then having `deliver` and `deliver-result` add attributes to it.  I'm not sure this is ideal but I was at a loss to do it any other way since some of the `clist` events that need to be part of this span happen outside the bounds of `deliver` and `deliver-result`.  It's a little ugly but the output it produces is actually nice.  I think this is as far as I'm prepared to take it for now.

FUDCo:
@mhofman ready for you to re-review.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -347,110 +407,167 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
           ...attrs
         } = slogAttrs;
+        if (isReplaying) {
+          spans
+            .top()
+            ?.addEvent(
+              `deliver-${delivery}`,
+              cleanAttrs({ delivery, ...attrs }),
+              now,
+            );
```
mhofman, line 488: Interesting to add those as events. Wondering if there'll be an easy way to filter events from spans if the `replay` attribute is not included.
Also, should we check that the `replay` attr is actually `true` as expected here?

packages/telemetry/src/slog-to-otel.js
```diff
@@ -662,16 +871,17 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
         break;
       }
       case 'crank-start': {
-        const { message, ...crankAttrs } = slogAttrs;
-        const [name, messageAttrs, links] = extractMessageAttrs(message);
+        const { message, crankType, ...crankAttrs } = slogAttrs;
+        if (crankType === 'routing') {
+          break;
```
mhofman, line 876: Are we not reporting routing cranks at all then? Wondering if it'd be useful as an event of the `swingset-run` span.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -662,16 +871,17 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+        }
+        const [name, _messageAttrs, links] = extractMessageAttrs(message);
         spans.startNamed(
```
mhofman, line 879: I was hoping we'd be able to use `push` instead of `startNamed` to get rid of the notion of named spans with explicit parents, and the `crankKey` logic, but I guess that can come later. One benefit is that it makes the `vat-replay` span nested under `swingset-run` span instead of under the crank span. On the other end, it keeps the `snapshot-save` event on the `swingset-run` span, which is a bit obscure.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -347,110 +407,167 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
         if (!kd) {
           break;
         }
-        const crankSpan = spans.get(getCrankKey());
-        crankSpan.setAttributes(cleanAttrs({ delivery, ...attrs }));
+
+        spans.get(getCrankKey()).setAttributes(
+          cleanAttrs({
+            delivery,
+            vatName: vatIdToAttrs.get(attrs.vatID).name,
```
mhofman, line 498: Oh another magical thing that would have broke! Thanks for catching.

packages/telemetry/src/slog-to-otel.js
```diff
@@ -501,60 +623,155 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+            makeSyscallSpan(name, sattrs);
+            break;
+          }
+          case 'vatstoreSet': {
+            const [_, vatID, key, value] = ksc;
+            const sattrs = {
+              vatID,
+              key,
+              value,
+              details: `${detailPrefix}: vatstoreSet('${key}', '${value}')`,
```
mhofman, line 736: Nit: `value` is stringified already, it'd make sense to not quote it:

```suggestion
              details: `${detailPrefix}: vatstoreSet('${key}', ${value})`,
```

packages/telemetry/src/slog-to-otel.js
```diff
@@ -501,60 +623,155 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
             },
             options,
           );
+        const syscallType = ksc[0];
+        const name = `syscall-${syscallType}`;
         // TODO: get type from packages/SwingSet/src/kernel/kernelSyscall.js
-        switch (ksc[0]) {
+        const detailPrefix = `s-${attrs.vatID}`;
```
mhofman, line 629: Wondering if the prefix should be more unique. E.g. `s-${crankNum}` (if we have the crankNum)

packages/telemetry/src/slog-to-otel.js
```diff
@@ -571,52 +788,44 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
         break;
       }
-      case 'bundle-kernel-start': {
-        spans.push(`bundle-kernel`);
-        break;
-      }
-      case 'bundle-kernel-finish': {
-        spans.pop('bundle-kernel');
+      case 'kernel-stats': {
+        spans.top()?.addEvent('kernel-stats', cleanAttrs(slogAttrs.stats), now);
```
mhofman, line 792: I had missed this originally, we're ok with the stats being flat under the `agoric.` prefix, and not nested under a e.g. `agoric.stats.` prefix?

mhofman:
Blah VS code PR plugin messed up and is sending comments one by one, sorry about that

packages/telemetry/src/slog-to-otel.js
```diff
@@ -347,110 +407,167 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
+          const details = `d-${attrs.vatID}: ${detailses.join(', ')}`;
+          spans.get(getCrankKey()).setAttributes(cleanAttrs({ details }));
+
           // Track call graph.
-          const [_type, [notification]] = kd;
           // TODO: We only track the first notified kernel promise. Perhaps we should do more?
-          const [kpId, { state }] = notification;
-          const error = state === 'rejected' ? 'rejected' : undefined;
-          if (kernelPromiseToSendingCause.has(kpId)) {
+          const [tKpid, { tState }] = notifications[0];
```
mhofman, line 526: Why the `t` prefix on these ?

packages/telemetry/src/slog-to-otel.js
```diff
@@ -347,110 +407,167 @@ export const makeSlogToOtelKit = (tracer, overrideAttrs = {}) => {
...
           let cause;
           if (kernelPromiseToSendingCause.has(result)) {
             cause = JSON.parse(kernelPromiseToSendingCause.get(result));
           } else {
             // Create a new root span.
-            const name = `E(${target}).${method}`;
+            const crankSpan = spans.get(getCrankKey());
             cause = {
-              name,
+              name: `deliver-${kd[0]}`,
```
mhofman, line 566: The re-computing of the string used as crank name makes me feel uncomfortable. If this was actually used for anything. it'd be ripe for breakage in a future refactor.

warner:

packages/telemetry/src/slog-to-otel.js
FUDCo: Empirically, I've verified that it is.  However, it would be easy enough to check.  My only concern is that it makes things a bit more complicated and adds a failure mode -- I'm a subscriber to the principle of not checking for error conditions that you're not prepared to deal with anyway.

mhofman (approved):
Besides the nits mentioned above, this is good enough for now. I'm happy to take it over from here once you land this PR.

packages/telemetry/src/slog-to-otel.js
FUDCo: Correct.  My recollection was that we decided that when we had our big discussion after I returned from vacation.  The rationale was that they have no duration and thus just add noise and complication to the analysis.

packages/telemetry/src/slog-to-otel.js
FUDCo: I tried, I really did, but there's enough inexplicable weirdness in the difference between what `startNamed` and `push` do that it defeated me.

packages/telemetry/src/slog-to-otel.js
FUDCo: No. The value of `value` is always a string, so you need the quotes in the output to see where the string starts and ends.  In particular, it's especially hard to read otherwise if the value contains spaces or is an empty string.

packages/telemetry/src/slog-to-otel.js
FUDCo: Yeah, that's how you laid it out in your notes.  However, my experience was that the crankNum wasn't always available.  It could be made available with some additional effort, but I was trying to minimize the volume of work.

packages/telemetry/src/slog-to-otel.js
FUDCo: I'm OK with it.  Frankly, I find the whole name-spacing of attribute names thing kind of dumb and useless anyway.

packages/telemetry/src/slog-to-otel.js
FUDCo: Cheap workaround for eslint rule about nested declarations of the same name.

packages/telemetry/src/slog-to-otel.js
FUDCo: Actually, if we were to rename the `message` delivery to something else (e.g., `msg` or `communique`, say), this would _prevent_ that renaming from breaking the code.

packages/telemetry/src/slog-to-otel.js
mhofman: Oh I though the value was always `JSON.stringify`ed, in which case the quotes would have been redundant.

packages/telemetry/src/slog-to-otel.js
mhofman: Yeah I'm just paranoid of conflicts/overloads in otel sinks which may assume the attribute name to have some kind of semantic meaning if they appear on differently named spans.

packages/telemetry/src/slog-to-otel.js
FUDCo: No, the value is simply a string -- in some cases it's produced by a result from `JSON.stringify` (e.g., when the value is something that's been serialized) but that's by no means always the case.

packages/telemetry/src/slog-to-otel.js
mhofman: Right from that angle. I was more concerned about a refactor in this file. It's already pretty obscure as-is ;)

FUDCo:
Since it doesn't look like there's anything urgently pressing in @mhofman's comments, I'm going to land as is modulo fixing the conflict with `make-slog-sender.js` (for which I think the resolution is simply adopting the version that's on the master branch since it's a more up-to-date version of a fix to an issue that @mhofman fixed cleanly whereas I had simply put in a stopgap workaround.)

PR #6248, opened 2022-09-16
Improve the efficiency of writing snapshots

### LINKED ISSUE #6225: Snapshot creation is inefficient

#6188 revealed some inefficiencies in snapshot creation:
* Use of a potentially-remote file system for temporary files
* Multiple serial reads of raw snapshot data

## Description of the Design

We should place temporary files in TMPDIR, and tee the snapshot data to parallel readers.

## Security Considerations

None expected

## Test Plan

Unit tests will verify functionality, but the performance changes theirselves should be evaluated on a real network.

gibson042:
closes: #6225
refs: #6188

## Description

* Use TMPDIR for temporary files.
* Parallelize reading raw snapshot data for hash computation and compression.

### Security Considerations

n/a

We might want to mention somewhere that snapshot creation puts some temporary files in the system temporary directory (as communicated by e.g. the POSIX `TMPDIR` environment variable).

### Testing Considerations

Unit tests verify stable functionality, but the performance changes theirselves should be evaluated on a real network.


packages/swing-store/src/snapStore.js
```diff
@@ -217,42 +196,125 @@ export function makeSnapStore(
...
-      );
+
+      // Perform operations that read snapshot data in parallel.
+      // We still serialize the stat and opening of tmpSnapPath
+      // and creation of tmpGzPath for readability, but we could
+      // parallelize those as well if the cost is significant.
+      const snapReader = createReadStream(tmpSnapPath);
+      cleanup.push(() => snapReader.destroy());
+      await fsStreamReady(snapReader);
+      // TODO: hoist.
```
gibson042: I'll be resolving these "hoist" TODOs while the PR goes through CI, along with updating now-redundant definitions for e.g. path vs. stream parameters. They were defined inline during development, and left in place for now to avoid delaying the PR (which can be merged with them depending upon urgency).

mhofman:
@gibson042 I just pushed the changes we paired on

packages/swing-store/src/snapStore.js
```diff
@@ -91,21 +95,26 @@ export const fsStreamReady = stream =>
...
 ) {
   /** @type {(opts: unknown) => Promise<string>} */
   const ptmpName = promisify(tmpName);
 
+  /** @type {(fd: number) => Promise<void>} */
+  const pfsync = promisify(fsync);
+
   /**
    * Returns the result of calling a function with the name
    * of a temp file that exists only for the duration of
```
gibson042, line 120: Would `withTempName` be simpler on top of `aggregateTryFinally`?
```js
  async function withTempName(fn, prefix = 'tmp') {
    const name = await ptmpName({
      tmpdir: root,
      template: `${prefix}-XXXXXX.xss`,
    });
    return aggregateTryFinally(
      () => fn(name),
      // Ignore file deletion errors.
      () => unlink(name).catch(sink),
    );
  }
```
Or maybe the point is moot, since `load` can be refactored to eliminate `withTempName`.

packages/swing-store/src/snapStore.js
```diff
@@ -208,6 +188,22 @@ export function makeSnapStore(
     return hash.digest('hex');
   }
 
+  // Create a file for the compressed snapshot in-place
+  // to be atomically renamed once we know its name
+  // from the hash of raw snapshot contents.
+  // TODO: hoist.
```
gibson042:
```suggestion
  // Manually promisify `tmpFile` to preserve its post-`error` callback arguments.
```

packages/swing-store/src/snapStore.js
```diff
@@ -217,63 +213,117 @@ export function makeSnapStore(
...
+        const { size: rawByteCount } = await stat(tmpSnapPath);
+
+        // Perform operations that read snapshot data in parallel.
+        // We still serialize the stat and opening of tmpSnapPath
+        // and creation of tmpGzPath for readability, but we could
+        // parallelize those as well if the cost is significant.
+        const snapReader = createReadStream(tmpSnapPath);
+        cleanup.push(() => {
+          snapReader.destroy();
+        });
```
gibson042, line 247: Cleanup ignores return values and the other uses of `cleanup.push` don't suppress them like this.
```suggestion
        cleanup.push(() => snapReader.destroy());
```

packages/swing-store/src/snapStore.js
```diff
@@ -217,63 +213,117 @@ export function makeSnapStore(
...
+        const {
+          path: tmpGzPath,
+          fd: tmpGzFd,
+          cleanup: tmpGzCleanup,
+        } = await ptmpFile({ tmpdir: root });
+        const gzWriter = createWriteStream(noPath, {
+          fd: tmpGzFd,
+          autoClose: false,
+        });
+        cleanup.push(tmpGzCleanup);
```
gibson042:
```suggestion
        cleanup.push(tmpGzCleanup);
        const gzWriter = createWriteStream(noPath, {
          fd: tmpGzFd,
          autoClose: false,
        });
```

packages/swing-store/src/snapStore.js
```diff
@@ -217,63 +213,117 @@ export function makeSnapStore(
...
+        cleanup.push(() => gzWriter.close());
+
+        await Promise.all([
+          fsStreamReady(snapReader),
+          fsStreamReady(tmpGzCleanup),
```
gibson042:
```suggestion
          fsStreamReady(gzWriter),
```

packages/swing-store/src/snapStore.js
```diff
@@ -12,7 +13,6 @@ import { promisify } from 'util';
  * @property {boolean} newFile true if the compressed snapshot is new, false otherwise
  * @property {number} rawByteCount size of (uncompressed) snapshot
  * @property {number} rawSaveSeconds time to save (uncompressed) snapshot
- * @property {number} hashSeconds time to calculate snapshot hash
  * @property {number} compressedByteCount size of (compressed) snapshot
  * @property {number} compressSeconds time to compress and save snapshot (0 if the file is not new)
```
gibson042:
```suggestion
 * @property {number} compressSeconds time to compress and save snapshot
```

gibson042 (commented):
@mhofman Some further cleanup.

packages/swing-store/src/snapStore.js
```diff
@@ -217,63 +213,117 @@ export function makeSnapStore(
...
+        // Perform operations that read snapshot data in parallel.
+        // We still serialize the stat and opening of tmpSnapPath
+        // and creation of tmpGzPath for readability, but we could
+        // parallelize those as well if the cost is significant.
+        const snapReader = createReadStream(tmpSnapPath);
+        cleanup.push(() => {
+          snapReader.destroy();
+        });
+
+        const {
```
gibson042, line 252:
```suggestion
        // Create a file for the compressed snapshot in-place
        // to be atomically renamed once we know its name
        // from the hash of raw snapshot contents.
        const {
```

packages/swing-store/src/snapStore.js
mhofman: The return value would be adopted in the promise result in the `finally` stage, meaning the stream object would be kept around for longer than necessary.

packages/swing-store/src/snapStore.js
mhofman: I suppose we can use the same tmpFile approach for load.

packages/swing-store/src/snapStore.js
mhofman: Ooops

packages/swing-store/src/snapStore.js
gibson042: But just in the time between running the function in `PromiseAllOrErrors` and throwing away the aggregated value, which should be inconsequential.

packages/swing-store/src/snapStore.js
```diff
@@ -217,63 +210,117 @@ export function makeSnapStore(
...
-        rawSaveSeconds,
-        hashSeconds,
-      };
-      const fileStat = await stat(filePath).catch(e => {
-        if (e.code === 'ENOENT') {
-          return undefined;
+    const cleanup = [];
+    return aggregateTryFinally(
+      async () => {
+        // TODO: Refactor to use tmpFile rather than tmpName.
```
mhofman, line 232: Thinking about this, we can't use `tmpFile` here since it's not this process that opens the file, but not worth another CI churn to change the comment.

packages/swing-store/src/snapStore.js
```diff
@@ -217,63 +210,117 @@ export function makeSnapStore(
...
-        newFile: true,
-        compressSeconds,
-        compressedByteCount,
-      });
-    }, 'save-raw');
+      },
+      async () => {
+        await PromiseAllOrErrors(
+          cleanup.reverse().map(fn => Promise.resolve().then(() => fn())),
+        );
```
mhofman, line 321: Thinking about this more, shouldn't we wait for previous steps to be done to continue with the next cleanup step? Disposal of resource is hard!

packages/swing-store/src/snapStore.js
gibson042: The refactoring I have in mind would be updating the signature of `saveRaw` to accept a file handle or stream rather than a file name, but I understand that may not even be possible.

mhofman (approved):

packages/swing-store/src/snapStore.js
mhofman: Not an actual problem since only the last cleanup (first queued) is actually async. Let's follow-up to clean this up though.

packages/swing-store/src/snapStore.js
mhofman: Oh yeah we decompress to the file, then tell xsnap the filename, so it'd work.

mhofman:
For future reference, internally the `os.tmpdir()` call is used, which currently has the [following implementation in node](https://github.com/nodejs/node/blob/90af1d689de406313cc687e54e19b0ebb0ac030f/lib/os.js#L191-L194):

```
    path = safeGetenv('TMPDIR') ||
           safeGetenv('TMP') ||
           safeGetenv('TEMP') ||
           '/tmp';
```

PR #6251, opened 2022-09-17
feat: uncurryThis performance improvements 

mhofman:
Adopt changes of https://github.com/endojs/endo/pull/1288 as a patch


mhofman:
@erights the large patch is due to the `.min.js` files. I'm happy to skip those if you want.
Also it looks like our code transform is having a hard time with comments, and adding stuff in the wrong places. Nothing wrong execution-wise, just different diff than I'd expect to see.

erights:
>  I'm happy to skip those if you want.

Yes please

> code transform is having a hard time with comments

I had not realized the patch would be post transform, but I think I see why: If we only patched endo source, then part of building agoric-sdk would be building the patched endo, which is clearly beyond what patch should be doing.

But the patch should still be based on the diff of endo without https://github.com/endojs/endo/pull/1288 post-transform vs endo with https://github.com/endojs/endo/pull/1288 post transform. Does the transform screw up other things differently depending on whether it is starting with or without https://github.com/endojs/endo/pull/1288 ? That's surprising.

erights:
I had taken the simplicity of my previous experience with https://github.com/Agoric/agoric-sdk/pull/6082/files as representative. I did not realize that this situation would be so different.

mhofman:
The patch is based on a local `yarn build` of endo, which includes both the `src/` and `dist/` folders. It has to be that way to make sure all consumers get the patch, but I'm pretty sure no-one uses the `.min.js`, at least in the kernel and vats.

Regarding the transform, it seem to be adding stuff after next line comments instead of same line, and since I added JSDoc types, it shifted some things after the comments instead of keeping them where they were.

erights:
ok. will look again once the .min.js files disappear

erights (approved):
Only read lightly because the change is mechanically derived from manual changes I already approved. Nevertheless, passed my eyeballs over everything and did not see anything alarming.

LGTM!

PR #6252, opened 2022-09-17
refactor(swing-store): Use the same stream-based approach for loading snapshots as saving them

gibson042:
## Description

Copies the `const cleanup = []; return aggregateTryFinally(…)` pattern and removes consequently-unused helpers.

We could try abstracting that pattern itself into a helper, but I don't think it would save much because the particulars are different.

### Security Considerations

n/a

n/a

### Testing Considerations

n/a

### DIFF
--- a/packages/swing-store/src/snapStore.js
+++ b/packages/swing-store/src/snapStore.js
@@ -331,9 +331,9 @@ export function makeSnapStore(
     };
 
     const results = await Promise.allSettled([...toDelete].map(deleteHash));
-    const errors = results.flatMap(({ status, reason }) =>
-      status === 'rejected' ? [reason] : [],
-    );
+    const errors = /** @type {PromiseRejectedResult[]} */ (
+      results.filter(({ status }) => status === 'rejected')
+    ).map(({ reason }) => reason);
     if (errors.length) {
       throw Error(JSON.stringify(errors));
     }
--- a/packages/swing-store/src/snapStore.js
+++ b/packages/swing-store/src/snapStore.js
@@ -320,21 +320,15 @@ export function makeSnapStore(
   }
 
   async function commitDeletes(ignoreErrors = false) {
-    const deleteHash = ignoreErrors
-      ? async hash => {
-          // Always update toDelete; sink errors from unlink.
-          toDelete.delete(hash);
-          const fullPath = fullPathFromHash(hash);
-          await (keepSnapshots !== true
-            ? unlink(fullPath).catch(sink)
-            : undefined);
-        }
-      : async hash => {
-          // Update toDelete iff there is no error from unlink.
-          const fullPath = fullPathFromHash(hash);
-          await (keepSnapshots !== true ? unlink(fullPath) : undefined);
-          toDelete.delete(hash);
-        };
+    const handleError = ignoreErrors ? p => p.catch(sink) : p => p;
+    const deleteHash = async hash => {
+      const fullPath = fullPathFromHash(hash);
+      await (keepSnapshots !== true
+        ? handleError(unlink(fullPath))
+        : undefined);
+      // Update toDelete iff the preceding await did not throw an error.
+      toDelete.delete(hash);
+    };
 
     const results = await Promise.allSettled([...toDelete].map(deleteHash));
     const errors = results.flatMap(({ status, reason }) =>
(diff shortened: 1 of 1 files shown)


packages/swing-store/src/snapStore.js
```diff
@@ -344,13 +269,46 @@ export function makeSnapStore(
...
+          cleanup: tmpCleanup,
+        } = await ptmpFile({ template: `load-${hash}-XXXXXX.xss` });
+        cleanup.push(tmpCleanup);
+        const snapWriter = createWriteStream(noPath, {
+          fd,
+          autoClose: false,
+        });
+        cleanup.push(() => snapWriter.close());
+
+        await Promise.all([gzReader, snapWriter].map(s => fsStreamReady(s)));
```
mhofman: Nit: I'm not convinced by the readability of `.map` in this case

packages/swing-store/src/snapStore.js
```diff
@@ -344,13 +269,46 @@ export function makeSnapStore(
...
+        const hashStream = createHash('sha256');
+        snapReader.pipe(hashStream);
+        snapReader.pipe(snapWriter);
+
+        await Promise.all([gzReader, snapWriter].map(s => finished(s)));
+        const h = hashStream.digest('hex');
+        h === hash || assert.fail(d`actual hash ${h} !== expected ${hash}`);
```
mhofman, line 299: I did not realize we were checking the hash of the decompressed file!

packages/swing-store/src/snapStore.js
```diff
@@ -344,13 +269,46 @@ export function makeSnapStore(
...
+        await pfsync(fd);
```
mhofman: I believe that doing an `fsync` in this case is not necessary, and probably counter productive. We don't actually care if things have been flushed to disk, and would even prefer if it stayed in memory. The OS should guarantee that a different process reading this file should get the content we just wrote without it having been flushed to the underlying device.

packages/swing-store/src/snapStore.js
```diff
@@ -99,7 +97,7 @@ export function makeSnapStore(
     createWriteStream,
     measureSeconds,
     fsync,
-    open,
+    open: _open,
```
mhofman: Can't we simply remove this?

mhofman (commented):
Nice cleanup, but I think the `fsync` should be removed.

gibson042:
@mhofman Back to you.

packages/swing-store/src/snapStore.js
```diff
@@ -318,26 +320,26 @@ export function makeSnapStore(
...
+            : undefined);
         }
-      }),
-    );
+      : async hash => {
+          // Update toDelete iff there is no error from unlink.
+          const fullPath = fullPathFromHash(hash);
+          await (keepSnapshots !== true ? unlink(fullPath) : undefined);
+          toDelete.delete(hash);
+        };
```
mhofman: Edit: yikes GitHub is messing up the diff pretty bad!

```suggestion
    const handleError = ignoreErrors ? p => p.catch(sink) : p => p;
    const deleteHash = async hash => {
        const fullPath = fullPathFromHash(hash);
        await (keepSnapshots !== true ? handleError(unlink(fullPath)) : undefined);
        // Update toDelete iff there is a non-ignored error from unlink.
        toDelete.delete(hash);
      };
```

mhofman (commented):
Nit to DRY it up

packages/swing-store/src/snapStore.js
gibson042: Nice; done.

mhofman (approved):

packages/swing-store/src/snapStore.js
```diff
@@ -331,9 +331,9 @@ export function makeSnapStore(
     };
 
     const results = await Promise.allSettled([...toDelete].map(deleteHash));
-    const errors = results.flatMap(({ status, reason }) =>
-      status === 'rejected' ? [reason] : [],
-    );
+    const errors = /** @type {PromiseRejectedResult[]} */ (
+      results.filter(({ status }) => status === 'rejected')
+    ).map(({ reason }) => reason);
```
mhofman, line 336: Hoping that future versions of TS will help with cases like this.

PR #6256, opened 2022-09-17
static types improvements

turadg:
## Description

Various commits to increase type coverage. One yielded this bug detection https://github.com/Agoric/agoric-sdk/pull/6262

One new types feature is that the Installation type carries through to offer args validation and offer results, so we no longer need to cast (or guess) those.

That's the first several commits. This also goes further and tries to make progress on ( https://github.com/Agoric/agoric-sdk/issues/4560 ) by preventing node_modules diving for a few packages. This required making common ERTP types available as explicit exports instead of ambient. I expect that's the only controversial change: [428b274](https://github.com/Agoric/agoric-sdk/pull/6256/commits/428b2740a57829c8a7e741ae54a27529ee835171). If there is an impasse, I'm open to splitting that work to a separate PR.

### Security Considerations

n/a, static types

--

### Testing Considerations

Types check

### DIFF
--- a/packages/ERTP/src/amountMath.js
+++ b/packages/ERTP/src/amountMath.js
@@ -176,12 +176,13 @@ const checkLRAndGetHelpers = (leftAmount, rightAmount, brand = undefined) => {
 
 /**
  * @template {AssetKind} K
- * @param {MathHelpers<AssetKind>} h
+ * @param {MathHelpers<AssetValueForKind<K>>} h
  * @param {Amount<K>} leftAmount
  * @param {Amount<K>} rightAmount
  * @returns {[K, K]}
  */
 const coerceLR = (h, leftAmount, rightAmount) => {
+  // @ts-expect-error could be arbitrary subtype
   return [h.doCoerce(leftAmount.value), h.doCoerce(rightAmount.value)];
 };
 
@@ -243,9 +244,9 @@ const AmountMath = {
    * Return the amount representing an empty amount. This is the
    * identity element for MathHelpers.add and MatHelpers.subtract.
    *
-   * @template {AssetKind} K
+   * @template {AssetKind} [K='nat']
    * @param {Brand<K>} brand
-   * @param {K} [assetKind]
+   * @param {K} [assetKind='nat']
    * @returns {Amount<K>}
    */
   // @ts-expect-error TS/jsdoc things 'nat' can't be assigned to K subclassing AssetKind
@@ -254,6 +255,7 @@ const AmountMath = {
     assertRemotable(brand, 'brand');
     assertAssetKind(assetKind);
     const value = helpers[assetKind].doMakeEmpty();
+    // @ts-expect-error TS/jsdoc things 'nat' can't be assigned to K subclassing AssetKind
     return harden({ brand, value });
   },
   /**
--- a/packages/ERTP/src/types-module.d.ts
+++ b/packages/ERTP/src/types-module.d.ts
@@ -1,0 +1,445 @@
+/**
+ * @file Generated from types.js, then find-replace to export all
+ *
+ * This file exists because types.js is ambient and some modules need the types
+ * to be explictly imported. Changing types.js to a module breaks too much.
+ */
+/* eslint-disable no-use-before-define */
+export type InterfaceSpec = import('@endo/marshal').InterfaceSpec;
+export type GetInterfaceOf = import('@endo/marshal').MarshalGetInterfaceOf;
+/**
+ * Amounts are descriptions of digital assets, answering the questions
+ * "how much" and "of what kind". Amounts are values labeled with a brand.
+ * AmountMath executes the logic of how amounts are changed when digital
+ * assets are merged, separated, or otherwise manipulated. For
+ * example, a deposit of 2 bucks into a purse that already has 3 bucks
+ * gives a new purse balance of 5 bucks. An empty purse has 0 bucks. AmountMath
+ * relies heavily on polymorphic MathHelpers, which manipulate the unbranded
+ * portion.
+ */
+export type Amount<K extends AssetKind = AssetKind> = {
+  brand: Brand<K>;
+  value: AssetValueForKind<K>;
+};
+/**
+ * An `AmountValue` describes a set or quantity of assets that can be owned or
+ * shared.
+ *
+ * A fungible `AmountValue` uses a non-negative bigint to represent a quantity
+ * of that many assets.
+ *
+ * A non-fungible `AmountValue` uses an array or CopySet of `Key`s to represent
+ * a set of whatever asset each key represents. A `Key` is a passable value
+ * that can be used as an element in a set (SetStore or CopySet) or as the
+ * key in a map (MapStore or CopyMap).
+ *
+ * `SetValue` is for the deprecated set representation, using an array directly
+ * to represent the array of its elements. `CopySetValue` is the proper
+ * representation using a CopySet.
+ *
+ * A semi-fungible `CopyBagValue` is represented as a
+ * `CopyBag` of `Key` objects. "Bag" is synonymous with MultiSet, where an
+ * element of a bag can be present once or more times, i.e., some positive
+ * bigint number of times, representing that quantity of the asset represented
+ * by that key.
+ */
+export type AmountValue = NatValue | any[] | CopySetValue | CopyBagValue;
+/**
+ * "Value" is a deprecated alias for "AmountValue". Please use
+ * "AmountValue" instead.
+ */
+export type Value = AmountValue;
+/**
+ * See doc-comment for `AmountValue`.
+ */
+export type AssetKind = 'nat' | 'set' | 'copySet' | 'copyBag';
+export type AssetValueForKind<K extends AssetKind> = K extends 'nat'
+  ? NatValue
+  : K extends 'set'
+  ? any[]
+  : K extends 'copySet'
+  ? CopySetValue
+  : K extends 'copyBag'
+  ? CopyBagValue
+  : never;
+export type AssetKindForValue<V extends AmountValue> = V extends NatValue
+  ? 'nat'
+  : V extends any[]
+  ? 'set'
+  : V extends CopySetValue
+  ? 'copySet'
+  : V extends CopyBagValue
+  ? 'copyBag'
+  : never;
+export type DisplayInfo<K extends AssetKind = AssetKind> = {
+  /**
+   * Tells the display software how
+   * many decimal places to move the decimal over to the left, or in
+   * other words, which position corresponds to whole numbers. We
+   * require fungible digital assets to be represented in integers, in
+   * the smallest unit (i.e. USD might be represented in mill, a
+   * thousandth of a dollar. In that case, `decimalPlaces` would be
+   * 3.) This property is optional, and for non-fungible digital
+   * assets, should not be specified. The decimalPlaces property
+   * should be used for *display purposes only*. Any other use is an
+   * anti-pattern.
+   */
+  decimalPlaces?: number | undefined;
+  /**
+   * - the kind of asset, either
+   * AssetKind.NAT (fungible) or
+   * AssetKind.SET or AssetKind.COPY_SET (non-fungible)
+   */
+  assetKind: K;
+};
+/**
+ * The brand identifies the kind of issuer, and has a function to get the
+ * alleged name for the kind of asset described. The alleged name (such
+ * as 'BTC' or 'moola') is provided by the maker of the issuer and should
+ * not be trusted as accurate.
+ *
+ * Every amount created by a particular AmountMath will share the same brand,
+ * but recipients cannot rely on the brand to verify that a purported amount
+ * represents the issuer they intended, since the same brand can be reused by
+ * a misbehaving issuer.
+ */
+export type Brand<K extends AssetKind = AssetKind> = {
+  /**
+   * Should be used with `issuer.getBrand` to ensure an issuer and brand match.
+   */
+  isMyIssuer: (allegedIssuer: ERef<Issuer>) => Promise<boolean>;
+  getAllegedName: () => string;
+  /**
+   * Give information to UI on how to display the amount.
+   */
+  getDisplayInfo: () => DisplayInfo;
+  getAmountShape: () => Pattern;
+};
+/**
+ * Return true if the payment continues to exist.
+ *
+ * If the payment is a promise, the operation will proceed upon
+ * resolution.
+ */
+export type IssuerIsLive = (payment: ERef<Payment>) => Promise<boolean>;
+/**
+ * Get the amount of digital assets in the payment. Because the
+ * payment is not trusted, we cannot call a method on it directly, and
+ * must use the issuer instead.
+ *
+ * If the payment is a promise, the operation will proceed upon
+ * resolution.
+ */
+export type IssuerGetAmountOf<K extends AssetKind = AssetKind> = (
+  payment: ERef<Payment>,
+) => Promise<Amount<K>>;
+/**
+ * Burn all of the digital assets in the
+ * payment. `optAmount` is optional. If `optAmount` is present, the
+ * code will insist that the amount of the digital assets in the
+ * payment is equal to `optAmount`, to prevent sending the wrong
+ * payment and other confusion.
+ *
+ * If the payment is a promise, the operation will proceed upon
+ * resolution.
+ */
+export type IssuerBurn = (
+  payment: ERef<Payment>,
+  optAmountShape?: Pattern | undefined,
+) => Promise<Amount>;
+/**
+ * Transfer all digital assets from the payment to a new payment and
+ * delete the original. `optAmount` is optional. If `optAmount` is
+ * present, the code will insist that the amount of digital assets in
+ * the payment is equal to `optAmount`, to prevent sending the wrong
+ * payment and other confusion.
+ *
+ * If the payment is a promise, the operation will proceed upon
+ * resolution.
+ */
+export type IssuerClaim = (
+  payment: ERef<Payment>,
+  optAmountShape?: Pattern | undefined,
+) => Promise<Payment>;
+/**
+ * Combine multiple payments into one payment.
+ *
+ * If any of the payments is a promise, the operation will proceed
+ * upon resolution.
+ */
+export type IssuerCombine = (
+  paymentsArray: ERef<Payment>[],
+  optTotalAmount?: Amount | undefined,
+) => Promise<Payment>;
+/**
+ * Split a single payment into two payments,
+ * A and B, according to the paymentAmountA passed in.
+ *
+ * If the payment is a promise, the operation will proceed upon
+ * resolution.
+ */
+export type IssuerSplit = (
+  payment: ERef<Payment>,
+  paymentAmountA: Amount,
+) => Promise<Payment[]>;
+/**
+ * Split a single payment into many payments, according to the amounts
+ * passed in.
+ *
+ * If the payment is a promise, the operation will proceed upon
+ * resolution.
+ */
+export type IssuerSplitMany = (
+  payment: ERef<Payment>,
+  amounts: Amount[],
+) => Promise<Payment[]>;
+/**
+ * The issuer cannot mint a new amount, but it can create empty purses
+ * and payments. The issuer can also transform payments (splitting
+ * payments, combining payments, burning payments, and claiming
+ * payments exclusively). The issuer should be gotten from a trusted
+ * source and then relied upon as the decider of whether an untrusted
+ * payment is valid.
+ */
+export type Issuer<K extends AssetKind = AssetKind> = {
+  /**
+   * Get the Brand for this Issuer. The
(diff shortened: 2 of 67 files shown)


packages/ERTP/src/amountMath.js
```diff
@@ -243,9 +244,9 @@ const AmountMath = {
    * Return the amount representing an empty amount. This is the
    * identity element for MathHelpers.add and MatHelpers.subtract.
    *
-   * @template {AssetKind} K
+   * @template {AssetKind} [K='nat']
    * @param {Brand<K>} brand
-   * @param {K} [assetKind]
+   * @param {K} [assetKind='nat']
```
Chris-Hibbert, line 249: Does it make sense for both line 247 and 249 to specify defaults?

packages/zoe/src/zoeService/types.js
```diff
@@ -152,12 +152,13 @@
...
  *
- * @param {ERef<Installation>}
+ * @param {ERef<Installation>} allegedInstallation
  * @returns {Promise<BundleID>}
  */
 
 /**
- * @template {object} [OR=any]
+ * @template {object} [OA=any] offer args
+ * @template {object} [OR=any] offer result
```
Chris-Hibbert: Can these be `Args` and `Result` rather than `OA` and `OR`?

packages/zoe/test/unitTests/zcf/test-zcf.js
```diff
@@ -769,7 +769,7 @@ test(`zcfSeat.isOfferSafe from zcf.makeEmptySeatKit`, async t => {
  * @param {Keyword} zcfMintKeyword
  * @param {ZCFSeat} zcfSeat
  * @param {Keyword} gainsKeyword
- * @param {AmountValue} gainsValue
+ * @param {bigint} gainsValue
```
Chris-Hibbert, line 772: [nit]   Doesn't this accept an arbitrary value?  I think that's a clearer declaration.

packages/governance/src/types.js
```diff
@@ -279,7 +279,7 @@
 
 /**
  * @typedef { ElectorateCreatorFacet & {
- *   getVoterInvitations: () => Promise<Invitation>[]
+ *   getVoterInvitations: () => Promise<Invitation<{}, {voter: import('../test/swingsetTests/contractGovernor/vat-voter.js').EVatVoter}>>[]
```
Chris-Hibbert: types shouldn't depend on tests, right?

packages/zoe/src/types.js
```diff
@@ -44,7 +44,7 @@
  */
 
 /**
- * @template {object} [A=any] Offer args
  * @template {object} [R=unknown] Offer result
+ * @template {object} [A=any] Offer args
```
Chris-Hibbert: Why is this an improvement?

Shouldn't `A` default to `void` (or whatever the TS equivalent is) when it's not specified?  That appears to be the case you're trying to simplify.

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
```diff
@@ -2859,7 +2858,7 @@ test('manager notifiers', async t => {
   // can't use 0n because of https://github.com/Agoric/agoric-sdk/issues/5548
   // but since this test is of metrics, we take the opportunity to check totalCollateral changing
   const given = aeth.make(2n);
-  vaultSeat = await E(services.zoe).offer(
+  let vaultOpSeat = await E(services.zoe).offer(
```
Chris-Hibbert, line 2861: `Op` could be short for a lot of things. What are you trying to hint at here?

packages/zoe/src/contractFacet/types.js
```diff
@@ -86,8 +86,13 @@
  */
 
 /**
- * @template {object} [OR=unknown] OR is OfferResult
- * @callback MakeInvitation
+ * @typedef {<R>(
```
Chris-Hibbert, line 89: Please indicate that `R` is short for `Result`.

packages/inter-protocol/src/psm/types.js
```diff
@@ -1,2 +1 @@
-// @ts-check
```
mhofman: backtrack previous commit?

packages/zoe/src/contractFacet/types.js
```diff
@@ -98,12 +103,11 @@
...
- * @param {OfferHandler<OR>} offerHandler - a contract specific function
+ * @param offerHandler - a contract specific function
  * that handles the offer, such as saving it or performing a trade
- * @param {string} description
- * @param {object=} customProperties
- * @param {Pattern} [proposalShape]
- * @returns {Promise<Invitation>}
+ * @param description
+ * @param [customProperties]
+ * @param [proposalShape]
```
Chris-Hibbert, line 110: If these aren't intended to be actual typescript declarations, can they be dropped?  The only detail here not covered above is the comment on `offerHandler`. If it can't be added in the declaration, it would be more helpful as a second paragraph of commentary than in something that only resembles a declaration.

packages/solo/src/chain-cosmos-sdk.js
```diff
@@ -728,7 +729,7 @@ ${chainID} chain does not yet know of address ${clientAddr}${adviseEgress(
    *
    * It then delivers the mailbox to inbound.  There are no optimisations.
    *
-   * @param {number=} lastMailboxUpdate
+   * @param {UpdateCount=} lastMailboxUpdate
```
mhofman, line 733:
```suggestion
   * @param {UpdateCount} [lastMailboxUpdate]
```

packages/solo/src/chain-cosmos-sdk.js
```diff
@@ -81,7 +81,7 @@ const makeTempFile = async (prefix, contents) => {
         if (err) {
           return reject(err);
         }
-        return resolve();
+        return resolve(undefined);
```
mhofman, line 84: Confused by the need to `undefined` here. No-one is expecting a specific result type, so `void` should be valid, no?

packages/zoe/src/zoeService/types.js
```diff
@@ -49,7 +49,7 @@
  * code breaks.
  * @property {GetConfiguration} getConfiguration
  * @property {GetBundleIDFromInstallation} getBundleIDFromInstallation
- * @property {GetProposalShapeForInvitation} getProposalShapeForInvitation
+ * @property {(invitationHandle: InvitationHandle) => Pattern | undefined} getProposalShapeForInvitation
```
Chris-Hibbert, line 52: `GetProposalShapeForInvitation` is still defined in `zoeService/internal-types.js`, and used in a few places. If it doesn't describe this method, what is it for?

packages/solo/src/chain-cosmos-sdk.js
```diff
@@ -759,6 +760,7 @@ ${chainID} chain does not yet know of address ${clientAddr}${adviseEgress(
     // Reset the backoff period.
     retryBackoff = randomizeDelay(INITIAL_SEND_RETRY_DELAY_MS);
   };
+  /** @param {Error} [e]  */
```
Chris-Hibbert, line 764: academic question:   Why is this helpful here? We don't often declare the types of errors, and there doesn't appear to be anything special going on here.

packages/zoe/test/unitTests/contracts/test-priceAggregatorChainlink.js
```diff
@@ -73,6 +77,7 @@ test.before(
       /** @type {OracleHandler} */
       const oracleHandler = Far('OracleHandler', {
         async onQuery({ increment }, _fee) {
+          // @ts-expect-error xxx
```
Chris-Hibbert, line 80: You usually say more. Oversight?

packages/ERTP/src/types-module.d.ts
```diff
@@ -0,0 +1,445 @@
+/**
+ * @file Generated from types.js, then find-replace to export all
+ *
+ * This file exists because types.js is ambient and some modules need the types
+ * to be explictly imported. Changing types.js to a module breaks too much.
+ */
```
mhofman, line 6: In https://github.com/Agoric/agoric-sdk/issues/4560#issuecomment-1055698699, I was imagining a different approach:
1) Add an `export {}` at the end of `types.js` to make it a module explicitly exporting types.
2) Add an `ambient.js` which does `/* @typedef {import('./types.js').Foo} Foo */`

The reason for 1. is to be able to do an `export * from './src/types.js` in the `index.js` entrypoint of a package, so that types can be consumed by other packages without having to go for deep imports (aka `import('@agoric/ertp').Amount`)
The reason for 2. is to not break existing ambient type usages.

That way we can also over time prune the number of types in `ambient.js` and don't need to keep 2 type definitions in sync.

Also while I am ok with TS syntax, I believe we still haven't settled the question of using it. Regardless the point above makes it a requirement that any public types declared in a `.ts` file have to be re-exported by a JSDoc `@typedef` to end up as an exported type at the root of the package.

packages/SwingSet/tools/manual-timer.js
```diff
@@ -53,7 +53,7 @@ const setup = () => {
  * kernel. You can make time pass by calling `advanceTo(when)`.
  *
  * @param {ManualTimerOptions} [options]
- * @returns {ManualTimer}
+ * @returns {TimerService & { advanceTo: (when: Timestamp) => void; }}
```
Chris-Hibbert, line 56: is `ManualTimer` too hard to import here, or not correct, or what?

packages/ui-components/src/components/NatPurseSelector.jsx
```diff
@@ -9,9 +9,9 @@ import { stringifyValue } from '../display/index.js';
  * components represent initial state as null/undefined.
  *
  * @param {Array<PursesJSONState> | null} purses Unfiltered purses. This may be null to simplify use in UIs.
- * @param {Brand} [optBrand] - optional brand to filter for
+ * @param {import('@agoric/ertp/src/types-module').Brand} [optBrand] - optional brand to filter for
```
mhofman, line 12: This is exactly what I want to avoid. We should have 
```suggestion
 * @param {import('@agoric/ertp').Brand} [optBrand] - optional brand to filter for
```

packages/ERTP/src/types-module.d.ts
```diff
@@ -0,0 +1,445 @@
+/**
+ * @file Generated from types.js, then find-replace to export all
+ *
+ * This file exists because types.js is ambient and some modules need the types
+ * to be explictly imported. Changing types.js to a module breaks too much.
```
Chris-Hibbert, line 5: Do you have a tool that automatically converted, or was in done manually?

What's the plan for maintenance?  If we will have to manually keep them in sync, that should be mentioned here and there, even if you plan to keep up with it yourself.

mhofman (changes requested):
I am not ok with further leaking the internal file structure of our packages through explicit type exports. Let's make those available through the root of the package, which requires the use of `@typedef`s

Chris-Hibbert (changes requested):

packages/solo/src/chain-cosmos-sdk.js
turadg: Without it,
> Expected 1 argument, but got 0. 'new Promise()' needs a JSDoc hint to produce a 'resolve' that can be called without arguments.

packages/ERTP/src/amountMath.js
turadg: I think so. One is JSDoc telling the consumer what the default is and the other the runtime making it so.

packages/zoe/src/zoeService/types.js
turadg: Yes of course. Do you think we should adopt that style? So far we've been using all caps to denote a generic type. Usually one letter. I'm more inclined to `A` for args and `R` for results.

If we want to change to words, I'd prefer that be done separately in a repo-wide change.

packages/zoe/test/unitTests/zcf/test-zcf.js
turadg: it doesn't actually, because the `zcf.makeZCFMint` returns a mint for a specific amount kind, in this case 'nat'

packages/governance/src/types.js
turadg: Agreed. I just couldn't find this defined in src. Maybe I found this but couldn't reach it by typedef,
https://github.com/Agoric/agoric-sdk/blob/2057268dfe44f33139a9057a9a30dd2cedffad25/packages/governance/src/committee.js#L78-L83

I'll leave this open until the import doesn't go from src to test

packages/zoe/src/types.js
Chris-Hibbert: Never mind.  You fixed it in the following commit, I think.

packages/solo/src/chain-cosmos-sdk.js
mhofman: Ugh right it defaults to `any`, which somehow isn't compatible with `void` in the return value case. This is really a time where the lack of an easy way to specify template types in JSDoc is annoying. Unfortunately [instantiation expressions](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#instantiation-expressions) wouldn't really help either, and short of a "cast" on the promise itself, there isn't a good way.

In this case I guess an explicit `undefined` is ok.

packages/zoe/src/zoeService/types.js
turadg: This was to get around an import depth problem. `GetProposalShapeForInvitation` is defined in internal-types and here we're describing `ZoeService` which is external. So the type couldn't be resolved. IMO we shouldn't complicate things with callbacks defs like that but I made the minimum necessary change

packages/solo/src/chain-cosmos-sdk.js
turadg: The special thing going on is `= (e = undefined) => {` which gives a type hint to TS so that `e` is not any, it's `undefined`. Then passing errors to `failedSend` doesn't satisfy the constraint. This annotation indicates it can be Error _or_  undefined

packages/SwingSet/tools/manual-timer.js
turadg: It's not correct, due to the incomplete Timer migration. `ManualTimer` is now supposed to have `tick` and `tickN` instead of `advanceTo`.

packages/ERTP/src/types-module.d.ts
turadg: It was `yarn lint:types` with some find replace. This file should say exactly how to reproduce it 👍

packages/zoe/src/zoeService/types.js
Chris-Hibbert: OA and OR seem strictly worse than either alternative. When a single letter is used, I much prefer to also have some part of the commentary indicate what it's mnemonic for, as is the case here, but I think I saw cases where that was dropped.  Do some of the formats for specifying typescript not allow room for comments on individual parameters or template values?  If that's the case, I'd prefer clearer names.

It's often the case that the role of the templated parameters is opaque from just the type declarations. If I have to specify them for offers, invitations, or start functions, it's really helpful to get an indication of what the missing declaration is for.

packages/ui-components/src/components/NatPurseSelector.jsx
turadg: I agree that's the goal. I just didn't want to have to remake all the ambients. But I accept that as a requirement for explicit package exports.

packages/zoe/src/zoeService/types.js
turadg: I think the reason generics are usually initialisms is to distinguish from types, which are StudlyCaps.

I take the point that it's helpful to have the parameter described, and when using the TS syntax there's no documentation string. So I agree when there are multiple type slots and there isn't a documentation string that we should use the more descriptive name, even it it might be confused with a typedef.

packages/solo/src/chain-cosmos-sdk.js
turadg: done in https://github.com/Agoric/agoric-sdk/pull/6287

packages/zoe/src/zoeService/types.js
turadg: Done in https://github.com/Agoric/agoric-sdk/pull/6287

packages/governance/src/types.js
turadg: Done in https://github.com/Agoric/agoric-sdk/pull/6287

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
turadg: operation. the reason this line was needed this seat holds a "string" result and the other hold somethign else

I went with something short because the meaning it's important and I didn't want any lines to rewrap

packages/zoe/src/contractFacet/types.js
turadg: It is, but I think doing that makes the generic type less readable because it's harder to scan where the generic is used in the expression. I didn't do this in https://github.com/Agoric/agoric-sdk/pull/6287 but if you feel strongly I will.

I get maybe someone would wonder "What's R stand for?" so I'm more inclined to make it `T`. It's just the generic slot.

packages/zoe/src/contractFacet/types.js
turadg: Good call. I thought they'd still document JSDoc for non-TS consumers, but in my IDE they're basically ignored.

packages/zoe/test/unitTests/contracts/test-priceAggregatorChainlink.js
turadg: moving fast. I think I can remove it

packages/zoe/src/contractFacet/types.js
turadg: Second thought, the readability could be solved by the source code highlighter. It isn't now but that it could be solved downstream tips the balance for me. I made the requested change in https://github.com/Agoric/agoric-sdk/pull/6287

packages/inter-protocol/test/vaultFactory/test-vaultFactory.js
Chris-Hibbert: My reaction was partly due to looking at the diffs and seeing that all the lines with the new name have lots of room for a long name.

packages/zoe/src/contractFacet/types.js
Chris-Hibbert: > the source code highlighter

Would you say more about this tool?  Is it VSCode-only?   I use webStorm, and others here are using emacs and more obscure things.

packages/zoe/src/zoeService/types.js
Chris-Hibbert: since ZoeService is external, the type should be externally visible. 

> we shouldn't complicate things with callbacks defs like that

say more, please?  Is the argument that it's better/clearer to type out all the details even when a set of parameters and return values are common across multiple calls?  Is this intended as a broad prescription?

packages/zoe/src/contractFacet/types.js
turadg: No particular tool; I meant whatever is highlighting the source code when one is looking at it. So WebStorm in your case and VS Code in mine. My point was that if distinguishing the generic slot is important to someone reading, they can (in principle) get their highlighter to do it.

packages/zoe/src/zoeService/types.js
turadg: My argument, which would apply broadly, is that the shared parameter list and return value is better to repeat so that we don't unnecessarily couple the different definitions. TypeScript types structurally, so when two things share the same structure they are as equal as when they share a typedef.

There is value in the developer mental model to things being the same, but in that case we should include the function name as well and say "this method on this object is the same as it is over there. That is best accomplished by a union with a picked subset of the other object.

e.g.,
```
type SomeService = {
  help: () => string;
  doSomething: (str: string) => void;
}
type SomeServiceInternal = Pick<SomeService, 'doSomething'> & {
  doAdminThing: () => AdminFacet;
}
```

packages/ERTP/src/amountMath.js
Chris-Hibbert: The runtime making it happen is on line 254.  As I read it (I'm admittedly not the expert here) line 247 says that the AssetKind type doesn't require that its parameter be specified, and if it's omitted from the types, it'll assume `'nat'`. Line 249 says that the call to makeEmpty doesn't require a second arg, with the same default. Don't those have the same impact?

If a call appears without the second arg, both typescript and javascript should use `'nat'`.  If the arg is provided, both should draw the same conclusion. If there's an explicit declaration that conflicts with the code, TS should complain.   Won't all the same things happen if line 247 has the default and 249 doesn't?

turadg: Sorry, I misread the discussion as being about line 254 (the runtime default).

You're right, 249 shouldn't have a default. I'll fix in this in https://github.com/Agoric/agoric-sdk/pull/6287

PR #6259, opened 2022-09-17
fix(xsnap): prevent unexpected worker exit

### LINKED ISSUE #6257: worker exit and kernel panic related to UnhandledPromiseRejection in new xs/xsnap
When @mhofman did more-extensive load testing of the `pso-rc0` candidate, he observed a chain halt after about 10 hours of work. We traced it down to the fee-distribution vat worker process exiting (possibly with `rc=0`, but we're not confident of that) in the middle of a delivery. This causes the kernel to panic, because workers spontaneously exiting could be due to memory exhaustion or the host OS shutting down (and killing the worker process before killing the kernel process), and we don't want to e.g. record an in-consensus "vat terminated" consequence for something that might be local to the one validator.

```
stage-26: chain: 2022-09-17T15:00:20.731Z block-manager: block 6714 begin
stage-26: monitor-slog: begin-block 6714
stage-26: chain: 2022-09-17T15:00:20.735Z launch-chain: polled; blockTime:1663426815, h:6714; ADDED = false
stage-26: chain: 2022-09-17T15:00:20.742Z SwingSet: xsnap: v18: UnhandledPromiseRejectionWarning: (TypeError#26)
stage-26: chain: 2022-09-17T15:00:20.742Z SwingSet: xsnap: v18: UnhandledPromiseRejectionWarning: (TypeError#25)
stage-26: chain: 2022-09-17T15:00:20.746Z SwingSet: xsnap: v18: UnhandledPromiseRejectionWarning: (TypeError#27)
stage-26: chain: 2022-09-17T15:00:20.772Z SwingSet: xsnap: v18: UnhandledPromiseRejectionWarning: (TypeError#26)
stage-26: chain: 2022-09-17T15:00:20.773Z SwingSet: xsnap: v18: UnhandledPromiseRejectionWarning: (TypeError#25)
stage-26: chain: 2022-09-17T15:00:20.773Z SwingSet: xsnap: v18: UnhandledPromiseRejectionWarning: (TypeError#27)
stage-26: chain: 2022-09-17T15:00:20.782Z SwingSet: kernel: error in kernel.deliver: (Error#1)
stage-26: chain: Error#1: v18:zcf exited
stage-26: chain:
stage-26: chain:   at packages/xsnap/src/xsnap.js:140:11
stage-26: chain:
stage-26: chain: 2022-09-17T15:00:20.784Z SwingSet: kernel: ##### KERNEL PANIC: error during tryProcessMessage: Error: v18:zcf exited #####
stage-26: chain: 2022-09-17T15:00:20.785Z block-manager: END_BLOCK error: (Error#1)
stage-26: chain: portHandler threw (Error#1)
stage-26: chain: 3:00PM ERR CONSENSUS FAILURE!!! err="Error: v18:zcf exited" module=consensus stack="goroutine 137 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:24 +0x65\ngithub.com/tendermint/tendermint/consensus.(*State).receiveRoutine.func2()\n\t/go/pkg/mod/github.com/agoric-labs/tendermint@v0.34.19-alpha.agoric.1/consensus/state.go:727 +0x4c\npanic({0x7fd17adda980, 0xc00451a590})\n\t/usr/local/go/src/runtime/panic.go:1038 +0x215\ngithub.com/Agoric/agoric-sdk/golang/cosmos/x/swingset.EndBlock({{0x7f...mint/consensus.(*State).handleMsg(0xc000f80000, {{0x7fd17b06aa80, 0xc0026a2628}, {0xc000f2e390, 0x0}})\n\t/go/pkg/mod/github.com/agoric-labs/tendermint@v0.34.19-alpha.agoric.1/consensus/state.go:856 +0x45f\ngithub.com/tendermint/tendermint/consensus.(*State).receiveRoutine(0xc000f80000, 0x0)\n\t/go/pkg/mod/github.com/agoric-labs/tendermint@v0.34.19-alpha.agoric.1/consensus/state.go:763 +0x419\ncreated by github.com/tendermint/tendermint/consensus.(*State).OnStart\n\t/go/pkg/mod/github.com/agoric-labs/tendermint@v0.34.19-alpha.agoric.1/consensus/state.go:379 +0x13e\n"
stage-26: runner: Stopping loadgen
```

We weren't able to figure out exactly why it was exiting: we added a bunch of printfs to `xsnap` and didn't manage to find the place that was calling `c_exit`. But we did find that commenting out the [recently-added](https://github.com/agoric-labs/xsnap-pub/commit/84aa01ac86596d46f98eeb5930bcdd33c66e0ef6) call to `fxCheckUnhandledRejections()` at the end of `fxRunLoop` made the process stop exiting.

https://github.com/agoric-labs/xsnap-pub/blob/84aa01ac86596d46f98eeb5930bcdd33c66e0ef6/xsnap/sources/xsnapPlatform.c#L367

We know the vat in question has had a bunch of unhandled rejections (#5745) since June, but [our copy of `fxAbort()`](https://github.com/agoric-labs/xsnap-pub/blob/84aa01ac86596d46f98eeb5930bcdd33c66e0ef6/xsnap/sources/xsnapPlatform.c#L105) merely logs such problems, and explicitly does *not* exit the process (note the lack of `the->abortStatus = status; fxExitToHost(the);` at the end of the `XS_UNHANDLED_REJECTION_EXIT` clause):

https://github.com/agoric-labs/xsnap-pub/blob/84aa01ac86596d46f98eeb5930bcdd33c66e0ef6/xsnap/sources/xsnapPlatform.c#L163-L182

@mhofman is sending a bunch of data to Moddable to see if they can understand why the process is exiting in this case.

While investigating this, we rediscovered #5745, in which some userspace code is doing something like `E('Collateral').offer(..)` (sending a message to a string, not a method-bearing object), and `E(undefined).makeAddCollateralInvitation()` (sending a message to `undefined`). To learn more, we modified the `lockdown` bundle (which includes Endo, hence eventual-send) to add some more logging. The chain does not display worker's stdout/stderr, but when we replayed the vat's transcript, the errors look like:

```
delivery 232 (L 237): ["notify",[["p+216",false,{"body":"{\"brand\":{\"@qclass\":\"slot\",\"iface\":\"Alleged: IST brand\",\"index\":0},\"value\":{\"@qclass\":\"bigint\",\"digits\":\"81671918\"}}","slots":["o-55"]}]]]
xsnap UnhandledPromiseRejectionWarning: (TypeError#26)
xsnap UnhandledPromiseRejectionWarning: (TypeError#25)
xsnap UnhandledPromiseRejectionWarning: (TypeError#27)
xsnap TypeError#27: target has no method undefined , has [ '0', '1', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '2', '20', '21', '22', '23', '24', '25', '26', '3', '4', '5', '6', '7', '8', '9', 'length' ]
xsnap TypeError: target has no method "undefined", has ["0","1","10","11","12","13","14","15","16","17","18","19","2","20","21","22","23","24","25","26","3","4","5","6","7","8","9","length"]
 at makeError (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:2722)
 at fail (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:2849)
 at localApplyMethod (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:9177)
 at apply ()
 at dispatchToHandler (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:9394)
 at win (/usr/src/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/lockdown-subprocess-xsnap.js:9734)
 at ()
```

Strangely, when we added a an extra clause to the `target has no method ..` assertion message, the process stopped exiting early. We're hoping that this might help Moddable shed some light on the surprising exit.

@mhofman 's loadgen process is exercising both Vault and AMM trades, neither of which are part of the PSO release. Given our focus in the last two months, it's entirely possible that something has broken in those vats, and/or the loadgen code which exercises them is no longer entirely compatible with them. #5745 has been appropriately scheduled out past PSO, and we don't expect them to be changed or fixed right now.

So, it's possible that there's an XS/xsnap bug/misbehavior that is triggered by non-PSO vats, and that might not be triggered by the PSO operations that are part of this release. If so, this bug might not need to be fixed for PSO (although it's weird enough that I'd be worried about letting it ride).

mhofman:
refs: #6257

## Description

We've noticed that under certain circumstances, xsnap worker exits 0, which was triggered by unhandled promise rejection checks.

After some analysis we understood the root cause was an bad interleaving problem in the xsnap worker protocol.

This is a combination of fixes and safeguards.

On the Moddable SDK side:
- Clean up the unhandled promise rejection list after processing it, so that already reported rejections are not reported multiple times. This has the effect of no longer triggering the worker protocol bug, but isn't a fix for the root cause.
- Move writing of the command response to after all XS activity is over. This is the root cause fix.
- Drive by fix of snapshot consistency issue that was noticed at the same time

On the xsnap-pub side:
- Check messages for unexpected commands, and add a new `q` command for clean exit.
- Remove new `/` prefix from query/syscall responses

On the xnsap.js side:
- Prefix query responses with a `/`
- Error if unexpected messages are received from a worker during an issued command.

### Testing Considerations

Heavy loadgen test on local machine succeeded with this branch.

### DIFF
--- a/packages/xsnap/src/xsnap.js
+++ b/packages/xsnap/src/xsnap.js
@@ -23,10 +23,9 @@ export const DEFAULT_CRANK_METERING_LIMIT = 1e8;
 const encoder = new TextEncoder();
 const decoder = new TextDecoder();
 
+const COMMAND_BUF = encoder.encode('?');
 const QUERY = '?'.charCodeAt(0);
-const QUERY_BUF = encoder.encode('?');
-const UNKNOWN_RESPONSE_BUF = encoder.encode('/');
-
+const QUERY_RESPONSE_BUF = encoder.encode('/');
 const OK = '.'.charCodeAt(0);
 const ERROR = '!'.charCodeAt(0);
 
@@ -207,7 +206,7 @@ export function xsnap(options) {
         );
       } else if (message[0] === QUERY) {
         const commandResult = await handleCommand(message.subarray(1));
-        await messagesToXsnap.next([UNKNOWN_RESPONSE_BUF, commandResult]);
+        await messagesToXsnap.next([QUERY_RESPONSE_BUF, commandResult]);
       }
     }
   }
@@ -269,7 +268,7 @@ export function xsnap(options) {
    */
   async function issueCommand(message) {
     const result = baton.then(async () => {
-      await messagesToXsnap.next([QUERY_BUF, message]);
+      await messagesToXsnap.next([COMMAND_BUF, message]);
       return runToIdle();
     });
     baton = result.then(noop, noop);
--- a/packages/xsnap/src/xsnap.js
+++ b/packages/xsnap/src/xsnap.js
@@ -20,14 +20,17 @@ import { defer } from './defer.js';
 // This will need adjustment, but seems to be fine for a start.
 export const DEFAULT_CRANK_METERING_LIMIT = 1e8;
 
-const OK = '.'.charCodeAt(0);
-const ERROR = '!'.charCodeAt(0);
-const QUERY = '?'.charCodeAt(0);
-
-const OK_SEPARATOR = 1;
-
 const encoder = new TextEncoder();
 const decoder = new TextDecoder();
+
+const QUERY = '?'.charCodeAt(0);
+const QUERY_BUF = encoder.encode('?');
+const UNKNOWN_RESPONSE_BUF = encoder.encode('/');
+
+const OK = '.'.charCodeAt(0);
+const ERROR = '!'.charCodeAt(0);
+
+const OK_SEPARATOR = 1;
 
 const { freeze } = Object;
 
@@ -203,7 +206,8 @@ export function xsnap(options) {
           )}`,
         );
       } else if (message[0] === QUERY) {
-        await messagesToXsnap.next(await handleCommand(message.subarray(1)));
+        const commandResult = await handleCommand(message.subarray(1));
+        await messagesToXsnap.next([UNKNOWN_RESPONSE_BUF, commandResult]);
       }
     }
   }
@@ -265,10 +269,7 @@ export function xsnap(options) {
    */
   async function issueCommand(message) {
     const result = baton.then(async () => {
-      const request = new Uint8Array(message.length + 1);
-      request[0] = QUERY;
-      request.set(message, 1);
-      await messagesToXsnap.next(request);
+      await messagesToXsnap.next([QUERY_BUF, message]);
       return runToIdle();
     });
     baton = result.then(noop, noop);
--- a/packages/xsnap/src/xsnap.js
+++ b/packages/xsnap/src/xsnap.js
@@ -211,7 +211,9 @@ export function xsnap(options) {
         // unrecognized responses also kill the process
         xsnapProcess.kill();
         const m = decoder.decode(message);
-        throw new Error(`xsnap protocol error: received unknown message <<${m}>>`);
+        throw new Error(
+          `xsnap protocol error: received unknown message <<${m}>>`,
+        );
       }
     }
   }
--- a/packages/xsnap/src/xsnap.js
+++ b/packages/xsnap/src/xsnap.js
@@ -207,6 +207,11 @@ export function xsnap(options) {
       } else if (message[0] === QUERY) {
         const commandResult = await handleCommand(message.subarray(1));
         await messagesToXsnap.next([QUERY_RESPONSE_BUF, commandResult]);
+      } else {
+        // unrecognized responses also kill the process
+        xsnapProcess.kill();
+        const m = decoder.decode(message);
+        throw new Error(`xsnap protocol error: received unknown message <<${m}>>`);
       }
     }
   }


packages/xsnap/src/xsnap.js
```diff
@@ -20,15 +20,18 @@ import { defer } from './defer.js';
 // This will need adjustment, but seems to be fine for a start.
 export const DEFAULT_CRANK_METERING_LIMIT = 1e8;
 
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+
+const QUERY = '?'.charCodeAt(0);
+const QUERY_BUF = encoder.encode('?');
+const UNKNOWN_RESPONSE_BUF = encoder.encode('/');
```
warner: Why is "UNKNOWN" in this name?

I know we need some stronger naming conventions around the kernel/worker protocol. We've got four messages:

* 1: kernel sends a delivery to the worker (or instructions to eval a bundle, or write a snapshot)
* within that, zero or more pairs of:
  * 2: worker sends request to kernel (a syscall, or a console log)
  * 3: kernel sends response back to worker (syscall results, or "ok" pacifier for console log)
* 4: worker sends delivery results (or eval success/failure, or "ok" snapshot-write pacifier) to kernel

I like "query" as a name for 2.

For 1, on the kernel-to-worker direction, `xsnap.js` uses QUERY_BUF (`?`) to build the delivery message. Then on the `xsnap-worker.c` side, in the command loop, we use `command` to hold it, and then dispatch to a JS function (registered by `supervisor-subprocess-xsnap.js`) named `handleCommand`.

For 2, in the worker-to-kernel direction, when the supervisor wants to do a syscall, it calls `globalThis.issueCommand()`, which is implemented by a C function named `xs_issueCommand()`, which writes a netstring with a `?` prefix from a buffer named `buf`. Then `xsnap.js` matches this against QUERY (`?`) and dispatches it to a function named `handleCommand()`.

For 3, in the kernel-to-worker direction, `xsnap.js` puts the `commandResult` into a payload with UNKNOWN_RESPONSE_BUF (`/`). This gets back to the `xsnap-worker.c` that is still waiting (blocking netstring read) inside `xs_issueCommand()` and holds it in a buffer named `buf`, which then returns it to the caller of `globalThis.issueCommand()`.

For 4, in the worker-to-kernel direction, `xsnap-worker.c` finishes the `fxRunLoop()` and `xsEndHost()` and either writes a `!`-prefixed error or `.`-prefixed success (plus metering) from a buffer named `response` and functions named `fxWriteNetString` or `fxWriteOkay`. Then `xsnap.js` receives this in the `runToIdle()` loop and compares it against OK or ERROR (in the same clause that compared 2 against QUERY), and breaks out of the loop, allowing `runToIdle()` to return.

It's a bit weird that we use "command" for both 1 (in xsnap-worker.c "handleCommand") and for 2 (in "xs_issueCommand" and xsnap.js "handleCommand"). So I'm in favor of moving our nomenclature for 2/3 to something like "query". But there are enough moving parts (xsnap.js, xsnap-worker.c, supervisor) that we aren't going to fix them all now.

So anyways, yeah, `QUERY` is good for 2, but `UNKNOWN_RESPONSE_BUF` feels weird for 3. How about we use `QUERY_RESPONSE_BUF`?

packages/xsnap/src/xsnap.js
```diff
@@ -203,7 +206,8 @@ export function xsnap(options) {
           )}`,
         );
       } else if (message[0] === QUERY) {
-        await messagesToXsnap.next(await handleCommand(message.subarray(1)));
+        const commandResult = await handleCommand(message.subarray(1));
+        await messagesToXsnap.next([UNKNOWN_RESPONSE_BUF, commandResult]);
```
warner: so rename this to `QUERY_RESPONSE_BUF`

packages/xsnap/src/xsnap.js
```diff
@@ -265,10 +269,7 @@ export function xsnap(options) {
    */
   async function issueCommand(message) {
     const result = baton.then(async () => {
-      const request = new Uint8Array(message.length + 1);
-      request[0] = QUERY;
-      request.set(message, 1);
-      await messagesToXsnap.next(request);
+      await messagesToXsnap.next([QUERY_BUF, message]);
```
warner: in line with the discussion above, what do you think about making a distinction between message 1 (kernel-to-worker delivery) and message 2 (worker-to-kernel syscall), even though both use a `?` prefix? How about we replace `const QUERY_BUF =` with `const COMMAND_BUF =  encoder.encode('?')`?

warner (changes requested):
This looks great, I'm so glad they were able to help us with the unhandled-rejection timing. I very much like your `xsnap-worker.c` changes to fix the ordering.

I can't annotate the `xsnap-worker.c` lines within this PR, but I'd love it if we could change the main command loop over there to react to unrecognized commands differently. Where you do:
```c
			case '/':
			case '.':
			case '!':
				fprintf(stderr, "Unexpected command %c\n", command);
			  c_exit(E_IO_ERROR);
				break;
```

which reacts fatally to those three specific characters, but continues to treat all others as "please exit silently", I'd replace it with

```c
			case 'q':
				done = 1;
				break;
			default:
				fprintf(stderr, "Unexpected command %c\n", command);
				 c_exit(E_IO_ERROR);
				break;
```

which reacts fatally to *all* unknown characters, but reserves `q` for "please exit silently".

(also, please check for tabs vs spaces in `xsnap-worker.c` and maintain the existing conventions, which I think are 4-char hard tabs)

Please update `xsnap-native/documentation/xsnap-worker.md` around line 47 with `q` ("causes the worker to exit with rc=0") and a note that "all other command characters cause the worker to emit an error message to stderr, and exit with rc=1 (E_IO_ERROR)".

`xsnap-worker.md` doesn't currently explain the "2"/"3" QUERY/QUERY_RESPONSE protocol. Could yo uadd a section at the end to record the `?` upstream and `/` downstream prefixes for this?

Also, that doc still says "only the first 100 such timestamps are reported", that needs to be updated with the new size (300? 500?).

And, I'd love it if we could emit more information in that error message. What about `("Unexpected command prefix '%c' in command '%s'", command, nsbuf)` ? I think the most likely places to trigger this are in the "3" query-response pathway, and those tend to be pretty small, and being able to see that we're getting a `["ok"]` instead of just a `[` would have made this bug easier to diagnose.

warner:
Here's the text to add to `xsnap-worker.md` for the query/response protocol:
```
  * if evaluation/`handleCommand()` throws an error (and the evaluated code does not catch it), the worker writes `!${toString(err)}` to fd4
  * while running, if the application calls `globalThis.issueCommand(query)`, the worker will write a netstring to file descriptor 4, whose payload is a single `?` character followed by the UTF-8 encoding of `query`
    * the application will then do a blocking read on file descriptor 3 until a complete netstring is received
    * the payload of this response must start with a single `/` character
    * the remainder of the payload will be UTF-8 decoded and provided as the return value from `issueCommand()`
  * if successful, the result should be an ArrayBuffer, or an object with a `.result` property that is an ArrayBuffer
    * anything else will yield an empty response string
```

warner (approved):

PR #6260, opened 2022-09-19
fix: sync with endo 1260 encodePassable

erights:
https://github.com/endojs/endo/pull/1260 started the migration of `encodePassable.js` and `rankOrder.js` to the endo repository. Thus, it naturally contains a copy of these file that would work there. This is a maintenance hazard until the migration is complete and we can remove these files from agoric-sdk. In the meantime, development in each place can make changes that should also be made to the other. This is the agoric-sdk side of syncing the changes thus far.


erights:
Reviewers,
Now that https://github.com/endojs/endo/pull/1260 has been merged, and this PR is synced with the merged state, ping ;)

packages/store/src/patterns/encodePassable.js
```diff
@@ -237,34 +272,50 @@ const encodeTagged = (tagged, encodePassable) =>
...
 export const makeEncodePassable = ({
-  encodeRemotable = rem => assert.fail(X`remotable unexpected: ${rem}`),
-  encodePromise = prom => assert.fail(X`promise unexpected: ${prom}`),
-  encodeError = err => assert.fail(X`error unexpected: ${err}`),
+  encodeRemotable = (rem, _) => assert.fail(X`remotable unexpected: ${rem}`),
+  encodePromise = (prom, _) => assert.fail(X`promise unexpected: ${prom}`),
+  encodeError = (err, _) => assert.fail(X`error unexpected: ${err}`),
 } = {}) => {
   const encodePassable = passable => {
+    if (ErrorHelper.canBeValid(passable, x => x)) {
```
gibson042: I think this can be simplified as of \@endo/marshal 0.7.6: https://github.com/endojs/endo/commit/053ca12110c706439a5f3611592a0b49cb829ac6
```suggestion
    if (ErrorHelper.canBeValid(passable)) {
```

packages/store/src/patterns/encodePassable.js
```diff
@@ -403,5 +466,39 @@ export const makeDecodePassable = ({
...
+// /////////////////////////////////////////////////////////////////////////////
+
+/**
+ * @type {Record<PassStyle, string>}
+ * The single prefix characters to be used for each PassStyle category.
+ * `bigint` is a two character string because each of those characters
+ * individually is a valid bigint prefix. `n` for "negative" and `p` for
+ * "positive". The ordering of these prefixes is the same as the
+ * rankOrdering of their respective PassStyles. This table is imported by
+ * randOrder.js for this purpose.
```
gibson042: Typo?
```suggestion
 * rankOrder.js for this purpose.
```

packages/store/src/patterns/patternMatchers.js
```diff
@@ -509,8 +509,8 @@ const makePatternKit = () => {
             X`${specimen} - Must be a copyRecord to match a copyRecord pattern: ${patt}`,
           );
         }
-        const [specimenNames, specimenValues] = recordParts(specimen);
-        const [pattNames, pattValues] = recordParts(patt);
+        const specimenNames = recordNames(specimen);
+        const pattNames = recordNames(patt);
         const missing = listDifference(pattNames, specimenNames);
```
gibson042, line 519: In-passing observation: copyRecord field name comparison could be about 2x to 4x faster if missing and unexpected names were identified in one iteration rather than two (leveraging the fact that both `pattNames` and `specimenNames` are reverse-sorted to simultaneously traverse each in order and classify their respective items as left-only vs. right-only vs. matching).

packages/store/src/patterns/rankOrder.js
```diff
@@ -29,38 +27,48 @@ const { ownKeys } = Reflect;
...
+  fromEntries(
+    entries(passStylePrefixes)
+      // Sort entries by ascending prefix.
+      .sort(([_leftStyle, leftPrefixes], [_rightStyle, rightPrefixes]) => {
+        return trivialComparator(leftPrefixes, rightPrefixes);
+      })
+      .map(([passStyle, prefixes], index) => {
+        // Cover all strings that start with any character in `prefixes`.
+        // `prefixes` is already sorted, so that's
+        // all s such that prefixes.at(0) ≤ s < successor(prefixes.at(-1)).
```
gibson042: Since this is one-time initialization code, would it make sense to assert that `prefixes` is sorted?
```suggestion
        // Cover all strings that start with any character in `prefixes`,
        // verifying that it is sorted so that is
        // all s such that prefixes.at(0) ≤ s < successor(prefixes.at(-1)).
        prefixes === [...prefixes].sort().join('') ||
          assert.fail(X`unsorted prefixes for passStyle ${q(passStyle)}: ${q(prefixes)}`);
```

gibson042 (approved):

packages/store/src/patterns/encodePassable.js
erights: Done

packages/store/src/patterns/encodePassable.js
erights: Done

erights: Done

packages/store/src/patterns/rankOrder.js
erights: Done

packages/store/src/patterns/patternMatchers.js
erights: Added a TODO comment suggesting an algorithm like `iterDisjointUnion`.

Resolving.

erights: Added a TODO comment suggesting an algorithm like `iterDisjointUnion`.

PR #6261, opened 2022-09-19
feat(zoe): expose keywords in error messages

turadg:
## Description

@erights by way of @dckc ,
> Amounts should not be quoted. Keywords, brands, etc, relatively static info that does not reveal actual quantities, should be quoted.

### Security Considerations

Keywords are safe to expose. They're part of the contract code.

--
### Testing Considerations

--

### DIFF
--- a/packages/zoe/src/cleanProposal.js
+++ b/packages/zoe/src/cleanProposal.js
@@ -29,7 +29,11 @@ export const assertKeywordName = keyword => {
   assert.typeof(keyword, 'string');
   keyword.length <= MAX_KEYWORD_LENGTH ||
     assert.fail(
-      X`keyword ${keyword} exceeded maximum length ${MAX_KEYWORD_LENGTH} characters; got ${keyword.length}`,
+      X`keyword ${q(
+        keyword,
+      )} exceeded maximum length ${MAX_KEYWORD_LENGTH} characters; got ${
+        keyword.length
+      }`,
     );
   assert(
     firstCapASCII.test(keyword),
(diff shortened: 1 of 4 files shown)


packages/zoe/src/cleanProposal.js
```diff
@@ -29,7 +29,11 @@ export const assertKeywordName = keyword => {
   assert.typeof(keyword, 'string');
   keyword.length <= MAX_KEYWORD_LENGTH ||
     assert.fail(
-      X`keyword ${keyword} exceeded maximum length ${MAX_KEYWORD_LENGTH} characters; got ${keyword.length}`,
+      X`keyword ${q(
+        keyword,
+      )} exceeded maximum length ${MAX_KEYWORD_LENGTH} characters; got ${
+        keyword.length
```
erights: Likely my mistake, but as long as you're in the immediate vicinity, might as well fix it too. No harm quoting both `MAX_KEYWORD_LENGTH` and `keyword.length`.

erights (approved):
LGTM, thanks!

packages/zoe/src/cleanProposal.js
turadg: I pushed a fix and neglected to quote `keyword.length`. I'm inclined to push as is because the actual keyword is there to observe. (if I push again I think I'd just remove the `; got…` part.

PTAAL at the other change in [4a21b80](https://github.com/Agoric/agoric-sdk/pull/6261/commits/4a21b8076fd08edd39c70cb1929a0b0921336aa3)

erights (approved):
New changes also LGTM

PR #6262, opened 2022-09-19
fix device timer remove()

turadg:
## Description

Working on https://github.com/Agoric/agoric-sdk/pull/6256 I noticed the condition in the loop was looking for a value that couldn't be in the array. I think this resulted in a bug in `TimerMap.remove` and thus `DeviceTimer.removeWakeup`.

### Security Considerations

Without this, wake handlers aren't removed from the map. PSMO uses timers for voting so this creates a resource exhaustion risk, proportional to the number of questions posed.

--
### Testing Considerations

No tests failed. What would a good regression test be?

### DIFF
--- a/packages/SwingSet/src/devices/timer/device-timer.js
+++ b/packages/SwingSet/src/devices/timer/device-timer.js
@@ -106,14 +106,14 @@ function makeTimerMap(state = undefined) {
    *
    * @param {bigint} time
    * @param {Waker} handler
-   * @param {number} [repeater]
+   * @param {number} [index]
    * @returns {bigint}
    */
-  function add(time, handler, repeater = undefined) {
+  function add(time, handler, index = undefined) {
     assert.typeof(time, 'bigint');
     /** @type {IndexedHandler} */
     const handlerRecord =
-      typeof repeater === 'number' ? { handler, index: repeater } : { handler };
+      typeof index === 'number' ? { handler, index } : { handler };
     const { handlers: records } = eventsFor(time);
     records.push(handlerRecord);
     schedule.sort((a, b) => Number(a.time - b.time));
@@ -147,7 +147,7 @@ function makeTimerMap(state = undefined) {
   // We don't expect this to be called often, so we don't optimize for it.
   /**
    *
-   * @param {Waker} targetHandler
+   * @param {Waker} target
    * @returns {bigint[]}
    */
   function remove(targetHandler) {
@@ -156,28 +156,17 @@ function makeTimerMap(state = undefined) {
     let i = 0;
     while (i < schedule.length) {
       const { time, handlers } = schedule[i];
-      if (handlers.length === 1) {
-        if (handlers[0].handler === targetHandler) {
-          schedule.splice(i, 1);
+      // Nothing prevents a particular handler from appearing more than once
+      for (let j = handlers.length - 1; j >= 0; j -= 1) {
+        if (handlers[j].handler === targetHandler) {
+          handlers.splice(j, 1);
           droppedTimes.push(time);
-        } else {
-          i += 1;
         }
+      }
+      if (handlers.length === 0) {
+        schedule.splice(i, 1);
       } else {
-        // Nothing prevents a particular handler from appearing more than once
-        for (const { handler } of handlers) {
-          // @ts-expect-error xxx Waker vs IndexedHandler
-          if (handler === targetHandler && handlers.indexOf(handler) !== -1) {
-            // @ts-expect-error xxx Waker vs IndexedHandler
-            handlers.splice(handlers.indexOf(handler), 1);
-            droppedTimes.push(time);
-          }
-        }
-        if (handlers.length === 0) {
-          schedule.splice(i, 1);
-        } else {
-          i += 1;
-        }
+        i += 1;
       }
     }
     return droppedTimes;
--- a/packages/SwingSet/src/devices/timer/device-timer.js
+++ b/packages/SwingSet/src/devices/timer/device-timer.js
@@ -147,8 +147,8 @@ function makeTimerMap(state = undefined) {
   // We don't expect this to be called often, so we don't optimize for it.
   /**
    *
-   * @param {Waker} target
-   * @returns {bigint[]}
+   * @param {Waker} targetHandler
+   * @returns {bigint[]} times that have been removed (may contain duplicates)
    */
   function remove(targetHandler) {
     /** @type {bigint[]} */
@@ -164,6 +164,7 @@ function makeTimerMap(state = undefined) {
         }
       }
       if (handlers.length === 0) {
+        // Splice out this element, preserving `i` so we visit any successor.
         schedule.splice(i, 1);
       } else {
         i += 1;
@@ -285,7 +286,10 @@ export function buildRootDeviceNode(tools) {
       saveState();
       return baseTime;
     },
-    /** @param {Waker} handler */
+    /**
+     * @param {Waker} handler
+     * @returns {bigint[]} times that have been removed (may contain duplicates)
+     */
     removeWakeup(handler) {
       const times = deadlines.remove(handler);
       saveState();
--- a/packages/SwingSet/src/devices/timer/device-timer.js
+++ b/packages/SwingSet/src/devices/timer/device-timer.js
@@ -86,6 +86,7 @@ function makeTimerMap(state = undefined) {
     return copyState(schedule);
   }
 
+  /** @param {bigint} time */
   function eventsFor(time) {
     assert.typeof(time, 'bigint');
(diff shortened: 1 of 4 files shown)


warner:
The new vat-timer.js only ever exposes a single handler function to device-timer.js . The device has some other bugs that needed fixing, and I wanted to touch as little code as possible, so when I rewrote vat-timer to be durable, I also constrained the way it uses the device, to avoid those bugs without actually changing the device. So while this patch is probably correct and good, I think we don't strictly need it now.

Chris-Hibbert (approved):
LGTM.  Of course, @warner is the expert.

warner:
Are we sure this change would be correct? I'm looking at `eventsFor()` and `add()`, and:
* `add()` takes a `handler` (a remote object with a `.wake` method)
* `add()` does `const handlerRecord = { handler }` or `{ handler, index: repeater }`
* `add()` does `const { handlers: records } = eventsFor(time)`
  * `eventsFor()` pushes a `newRecord = { time, handlers: [] }` onto `schedule, and returns `newRecord`
* `add()` does `records.push(handlerRecord)`, then `schedule.sort`

So `schedule[i] = { time, handlers }`, and `handlers[n] = { handler, index? }`.

Then `remove` walks `schedule[i]` to look at the `{ time, handlers }` of each. If `handlers.length === 1`, it tests `handlers[0].handler === targetHandler`, which matches the shape above. If it's longer, it walks `for (const { handler } of handlers)` and compares each against `targetHandler`, which also matches the shape above.

I think the confusion stems from a `schedule[i].handlers` that should really be named `schedule[i].handlerRecords`. Then we'd have `schedule[i].handlerRecords[j].handler` to compare against `targetHandler`. The loop being patched is really iterating over handlerRecords, each of which has a `handler` property.

I'll agree that testing here is lacking. `packages/SwingSet/test/timer-device/bootstrap.js` is where this ought to be tested, and it currently only does a single wakeup per moment (`D(devices.timer).setWakeup(3n, handler)`), where to exercise this it needs to submit two separate handlers for the same moment.

turadg:
> Are we sure this change would be correct?

Simple proof that the current behavior is wrong:
https://github.com/Agoric/agoric-sdk/blob/e73f693a3ccb83ee7e1eb0bb0c277b2164b3083b/packages/SwingSet/src/devices/timer/device-timer.js#L149-L150

The element being searched by `indexOf` was a property of an element of the array, not the element itself. Though now that I think about it, that `indexOf` check is superfluous because it just came out of the array.

I agree the confusion stems from naming. That's part of where static type checking shines, because it can't get thrown off by things appearing the same. It's job is to do all the analysis above automatically all the time for us.

Since we have `noImplicitAny: false` and a lot of `any` throughout the codebase, it does take some manual validation.

Now is the change here correct? I'm not certain without any unit tests but at least the types are. The reason I think `removeWakeup` takes `IndexHandler` is [the one call to it](https://github.com/Agoric/agoric-sdk/blob/0032e11b92abc82d5422908273310f12cf2d41c0/packages/SwingSet/src/vats/timer/vat-timer.js#L327) matches the shape, [with a `wake` property](https://github.com/Agoric/agoric-sdk/blob/0032e11b92abc82d5422908273310f12cf2d41c0/packages/SwingSet/src/vats/timer/vat-timer.js#L887-L890).

So `targetHandler` is an `IndexedHandler`, as is `handlers` from Event that is at `schedule[i]`.

packages/SwingSet/src/devices/timer/device-timer.js
```diff
@@ -132,24 +149,28 @@ function makeTimerMap(state = undefined) {
...
           i += 1;
         }
       } else {
         // Nothing prevents a particular handler from appearing more than once
-        for (const { handler } of handlers) {
-          // @ts-expect-error xxx Waker vs IndexedHandler
-          if (handler === targetHandler && handlers.indexOf(handler) !== -1) {
-            // @ts-expect-error xxx Waker vs IndexedHandler
+        for (const handler of handlers) {
+          if (handler.handler === target && handlers.indexOf(handler) !== -1) {
```
turadg: I don't know why it's checking for presence of the value it just got out of the iterator

mhofman: Well that's concerning. Previously `handler` was the `handlerRecord.handler` which could never be present in the `handlers` array, so this condition would have never matched. I am wondering if we somehow never encountered a case where more than one handler appeared in the `handlers` list?

mhofman (commented):
I need to qualify the previously broken behavior. The tests will greatly help, I just want to run them before the fix.

packages/SwingSet/src/devices/timer/device-timer.js
mhofman: Ok I see that the new test is tripping this case before the fix. As a matter of commit stacking, I would have preferred to see the commit adding the test marking it as `.failing`, and for the fix to remove the `.failing`.

mhofman: I believe this is still incorrect. splicing an array while iterating over it will move the items such that the item following the one being sliced out will be skipped.

The only safe way is iterating in reverse using the index.

mhofman (changes requested):
Checking why this hasn't been a concern, the reason is that the timer vat does the following:

```js
  const reschedule = () => {
    // the first wakeup should be in the future: the device will not
    // immediately fire when given a stale request
    const newFirstWakeup = firstWakeup(schedule);
    // idempotent and ignored if not currently registered
    D(timerDevice).removeWakeup(wakeupHandler);
    if (newFirstWakeup) {
      D(timerDevice).setWakeup(newFirstWakeup, wakeupHandler);
    }
  };
```

Which means it always removes the single handler it's using before adding a new schedule. That means we only ever have at most a single pending wake up schedule.

That said the remove logic is still wrong. I can push changes to fix.

mhofman (approved):
Removing my request for changes now that I've addressed the issue, but given my changes, I'd like another set of eyes. @gibson042 PTAL.

packages/SwingSet/src/devices/timer/device-timer.js
```diff
@@ -101,8 +106,16 @@ function makeTimerMap(state = undefined) {
   // There's some question as to whether it's important to invoke the handlers
   // in the order of their deadlines. If so, we should probably ensure that the
   // recorded deadlines don't have finer granularity than the turns.
+  /**
+   *
+   * @param {bigint} time
+   * @param {Waker} handler
+   * @param {number} [repeater]
```
gibson042: We should probably just rename this `repeater` parameter to `index`.

packages/SwingSet/src/devices/timer/device-timer.js
```diff
@@ -132,33 +149,28 @@ function makeTimerMap(state = undefined) {
   }
 
   // We don't expect this to be called often, so we don't optimize for it.
-  function remove(targetHandler) {
+  /**
+   *
+   * @param {Waker} target
```
gibson042: Please keep the old name for this parameter; it's more clear. 
```suggestion
   * @param {Waker} targetHandler
```

packages/SwingSet/src/devices/timer/device-timer.js
```diff
@@ -132,33 +149,28 @@ function makeTimerMap(state = undefined) {
...
+      for (let j = handlers.length - 1; j >= 0; j -= 1) {
+        if (handlers[j].handler === target) {
+          handlers.splice(j, 1);
           droppedTimes.push(time);
-        } else {
-          i += 1;
         }
+      }
+      if (handlers.length === 0) {
+        schedule.splice(i, 1);
```
gibson042, line 168:
```suggestion
        // Splice out this element, preserving `i` so we visit any successor.
        schedule.splice(i, 1);
```

packages/SwingSet/src/devices/timer/device-timer.js
```diff
@@ -132,33 +149,28 @@ function makeTimerMap(state = undefined) {
...
     while (i < schedule.length) {
       const { time, handlers } = schedule[i];
-      if (handlers.length === 1) {
-        if (handlers[0].handler === targetHandler) {
-          schedule.splice(i, 1);
+      // Nothing prevents a particular handler from appearing more than once
+      for (let j = handlers.length - 1; j >= 0; j -= 1) {
+        if (handlers[j].handler === target) {
+          handlers.splice(j, 1);
           droppedTimes.push(time);
```
gibson042, line 163: It's current behavior that the return value can include duplicate times, but still potentially surprising. Please document in the JSDoc header, and if possible confirm that consumers behave correctly w.r.t. the possibility.

gibson042 (approved):

gibson042 (approved):

PR #6265, opened 2022-09-19
fix wallet show latest

turadg:
## Description

`agoric wallet show` offer summaries showed the first update record, ignoring subsequent. This the records merge.

For purse balances it also let first win. This changes to last wins.

### Security Considerations

--

--
### Testing Considerations

Now that  `makeWalletStateCoalescer` is used in CLI and smart-wallet, the smart-wallet PSM integration tests provide some coverage.

### DIFF
--- a/packages/agoric-cli/src/lib/format.js
+++ b/packages/agoric-cli/src/lib/format.js
@@ -89,18 +89,20 @@ export const fmtRecordOfLines = record => {
 /**
  * Summarize the offerStatuses of the state as user-facing informative tuples
  *
- * @param {{ brands: import('./format').AssetDescriptor[], offers: Map<number, OfferStatus>}} state
+ * @param {import('@agoric/smart-wallet/src/utils.js').CoalescedWalletState} state
  * @param {Awaited<ReturnType<typeof makeAgoricNames>>} agoricNames
  */
 export const offerStatusTuples = (state, agoricNames) => {
-  const { brands, offers } = state;
-  const fmt = makeAmountFormatter(brands);
+  const { brands, offerStatuses } = state;
+  const fmt = makeAmountFormatter(
+    // @ts-expect-error xxx RpcRemote
+    [...brands.values()],
+  );
   const fmtRecord = r =>
     Object.fromEntries(
       Object.entries(r).map(([kw, amount]) => [kw, fmt(amount)]),
     );
-  return [...offers.keys()].sort().map(id => {
-    const o = offers.get(id);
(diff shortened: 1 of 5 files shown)


dckc:
> This changes to last wins.

Darn it. I missed that in review.

packages/agoric-cli/src/lib/wallet.js
```diff
@@ -21,24 +21,23 @@ export const getWalletState = async (addr, ctx, { vstorage }) => {
...
     assert(update.updated, 'missing key: updated');
     switch (update.updated) {
       case 'offerStatus': {
         const { status } = update;
-        if (!offers.has(status.id)) {
-          offers.set(status.id, status);
-        }
+        const lastStatus = offers.get(status.id);
+        // merge records
+        offers.set(status.id, { ...lastStatus, ...status });
```
dckc: I'd like a unit test for this fix. Does it look straightforward?

packages/agoric-cli/src/lib/format.js
```diff
@@ -78,3 +83,43 @@ export const fmtRecordOfLines = record => {
   );
   return `{\n${lineEntries.join(',\n')}\n}`;
 };
+
+/**
+ * @param {{ brands: import('./format').AssetDescriptor[], offers: Map<number, OfferStatus>}} state
+ * @param {Awaited<ReturnType<typeof makeAgoricNames>>} agoricNames
+ */
+export const offerStatusTuples = (state, agoricNames) => {
```
dckc, line 95: A 1 or 2 line description of what this function does would help me review it more quickly, I think.

packages/agoric-cli/src/lib/wallet.js
turadg: If I'm going to put more work into this, I think I'd instead make it use [coalesceUpdates](https://github.com/Agoric/agoric-sdk/blob/f800d07a5c8c64791778a45a8db90d27c5181ad2/packages/smart-wallet/src/utils.js#L14) which does have some test coverage. That way other UIs could use the same lib.

Will that suffice? I can include that change in this PR.

dckc: yes, that sounds good.

packages/agoric-cli/src/lib/format.js
dckc: ah. this just got moved and renamed.

dckc (approved):
looks good

I haven't verified that it works, but the risk is low.

PR #6277, opened 2022-09-19
fix(psmCharter): ParamChangesOfferArgsShape path

turadg:
## Description

The `path` option to ParamChangesOffer is optional:
https://github.com/Agoric/agoric-sdk/blob/master/packages/inter-protocol/src/psm/psmCharter.js#L53

 but the guard didn't think so:
https://github.com/Agoric/agoric-sdk/blob/master/packages/inter-protocol/src/psm/psmCharter.js#L25

### Security Considerations

--

--

### Testing Considerations

Regression test?

### DIFF
--- a/packages/inter-protocol/src/psm/psmCharter.js
+++ b/packages/inter-protocol/src/psm/psmCharter.js
@@ -18,12 +18,25 @@ import { E } from '@endo/far';
  * separate capabilities for finer grain encapsulation.
  */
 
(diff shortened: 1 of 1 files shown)


dckc (approved):
I just tested this (in #6278); it works.

Chris-Hibbert (approved):
This looks correct. 

There's still a question as to whether it's necessary for RC1, and what the threshold for adding changes to RC1 should be.

PR #6286, opened 2022-09-20
Typescript 4.8

turadg:
## Description

Upgrade TypeScript 4.7 to 4.8.

https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/

- [ ] make Endo compatible with 4.8 (or [stop using maxNodeModuleJsDepth](https://github.com/Agoric/agoric-sdk/issues/4560) )

### DIFF
--- a/packages/swing-store/src/snapStore.js
+++ b/packages/swing-store/src/snapStore.js
@@ -88,8 +88,8 @@ export const fsStreamReady = stream =>
  *   tmpName: typeof import('tmp').tmpName,
  *   unlink: typeof import('fs').promises.unlink,
  * }} io
- * @param {object} [options]
- * @param {boolean | undefined} [options.keepSnapshots]
+ * @param {object} options
+ * @param {boolean} [options.keepSnapshots]
  * @returns {SnapStore}
  */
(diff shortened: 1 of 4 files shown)


packages/swing-store/src/snapStore.js
```diff
@@ -88,8 +88,8 @@ export const fsStreamReady = stream =>
  *   tmpName: typeof import('tmp').tmpName,
  *   unlink: typeof import('fs').promises.unlink,
  * }} io
- * @param {object} [options]
- * @param {boolean | undefined} [options.keepSnapshots]
+ * @param {object} options
```
mhofman, line 91: That seems like a regression, and I've also been hit by this in my own code. There is an explicit init of the options param with `= {}` so it is actually optional. This really feels like a TS bug. I need to read more on why this is happening.

mhofman: Apparently there is already a bug filled on this: https://github.com/microsoft/TypeScript/issues/50286

turadg:
The regression won't be fixed until TS 4.9 so let's skip 4.8.

PR #6287, opened 2022-09-20
static type improvements to contracts

turadg:
Pulled out of https://github.com/Agoric/agoric-sdk/pull/6256 , to punt on ERTP package exports.

## Description

Various commits to increase type coverage. 

One new types feature is that the Installation type carries through to offer args validation and offer results, so we no longer need to cast (or guess) those.

### Security Considerations

n/a, static types

--

### Testing Considerations

Types check

### DIFF
--- a/packages/ERTP/src/amountMath.js
+++ b/packages/ERTP/src/amountMath.js
@@ -176,12 +176,13 @@ const checkLRAndGetHelpers = (leftAmount, rightAmount, brand = undefined) => {
 
 /**
  * @template {AssetKind} K
- * @param {MathHelpers<AssetKind>} h
+ * @param {MathHelpers<AssetValueForKind<K>>} h
  * @param {Amount<K>} leftAmount
  * @param {Amount<K>} rightAmount
  * @returns {[K, K]}
  */
 const coerceLR = (h, leftAmount, rightAmount) => {
+  // @ts-expect-error could be arbitrary subtype
   return [h.doCoerce(leftAmount.value), h.doCoerce(rightAmount.value)];
 };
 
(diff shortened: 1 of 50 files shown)


packages/ERTP/src/amountMath.js
```diff
@@ -254,6 +255,7 @@ const AmountMath = {
     assertRemotable(brand, 'brand');
     assertAssetKind(assetKind);
     const value = helpers[assetKind].doMakeEmpty();
+    // @ts-expect-error TS/jsdoc things 'nat' can't be assigned to K subclassing AssetKind
```
Chris-Hibbert:
```suggestion
    // @ts-expect-error TS/jsdoc thinks 'nat' can't be assigned to K subclassing AssetKind
```

packages/agoric-cli/src/types.d.ts
```diff
@@ -0,0 +1,7 @@
+// This package has maxNodeModuleJsDepth=0 to prevent ambient imports.
+// A consequence is that types that are only provided ambiently have to be defined.
+
+export type Brand = unknown;
+export type Amount = { brand: Brand; value: bigint };
+export type Instance = unknown;
+export type Proposal = any;
```
Chris-Hibbert, line 7: Why are these being left untyped?

packages/vat-data/src/far-class-utils.js
```diff
@@ -11,7 +11,8 @@ import {
 
 /** @template L,R @typedef {import('@endo/eventual-send').RemotableBrand<L, R>} RemotableBrand */
 /** @template T @typedef {import('@endo/eventual-send').ERef<T>} ERef */
-/** @typedef {import('@agoric/store').InterfaceGuard} InterfaceGuard */
+// FIXME import InterfaceGuard from @agoric/store
+/** @typedef {*} InterfaceGuard */
```
Chris-Hibbert: why not do the import now?

packages/agoric-cli/src/types.d.ts
turadg: Explained in a comment

packages/vat-data/src/far-class-utils.js
turadg: The type isn't available. Documented in 17aea73588cff36dced0e7c5c1759df30fe25021

packages/agoric-cli/src/types.d.ts
Chris-Hibbert: you mean the line above that says
> types that are only provided ambiently have to be defined.

Why is it better to declare them as `unknown` or `any` than to copy their actual types?  Does the CLI not rely on their values?

turadg: > Does the CLI not rely on their values?

Well the runtime doesn't rely on these typedefs at all. This package doesn't export anything to consumes so the types are there for dx productivity. The trade-off here is that there's more productivity to be gained by leaving them as is (vs spending more time on this).

> Why is it better to declare them as unknown or any than to copy their actual types?

The runtime types also don't match because the CLI gets its values from RPC and doesn't have CapTP to translate them into regular objects. That's what the `xxx RpcRemote` comments are about.

packages/ERTP/src/amountMath.js
turadg: I thought I fixed this but I see it's not pushed. When you approve I'll do a rebase to fixup the commit this was in and force push. (Holding off on that so I don't make it harder to review)

Chris-Hibbert (approved):

PR #6289, opened 2022-09-21
feat(wallet-ui): provision new wallet

### LINKED ISSUE #6211: UI for getting a smart wallet

UI for making a `provision-one` transaction.

## Description of the Design

?

(I can provide some references on back-end stuff)

## Test Plan

cc @samsiegart @rowgraus

samsiegart:
fixes: https://github.com/Agoric/agoric-sdk/issues/6211

### DIFF
--- a/packages/wallet/ui/src/components/ProvisionDialog.jsx
+++ b/packages/wallet/ui/src/components/ProvisionDialog.jsx
@@ -26,9 +26,22 @@ const CREATION_FEE = '10 BLD';
 
 const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
   const [currentStep, setCurrentStep] = useState(steps.INITIAL);
-  const [error, setError] = useState('');
+  const [error, setError] = useState(null);
 
-  const provisionWallet = async () => {
+  const provisionWallet = async signer => {
+    setError(null);
+    setCurrentStep(steps.AWAITING_APPROVAL);
+    try {
+      await signer.submitProvision();
+    } catch (e) {
+      setCurrentStep(steps.INITIAL);
+      setError(e.message);
+      return;
+    }
+    setCurrentStep(steps.IN_PROGRESS);
+  };
+
+  const handleCreateButtonClicked = () => {
     const {
       signers: { interactiveSigner },
     } = keplrConnection;
@@ -37,18 +50,7 @@ const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
       return;
     }
 
-    setError('');
-    setCurrentStep(steps.AWAITING_APPROVAL);
-
-    try {
-      await interactiveSigner.submitProvision();
-    } catch (e) {
-      setCurrentStep(steps.INITIAL);
-      setError(e.message);
-      return;
-    }
-
-    setCurrentStep(steps.IN_PROGRESS);
+    provisionWallet(interactiveSigner);
   };
 
   const progressIndicator = text => (
@@ -114,7 +116,7 @@ const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
           <Button color="cancel" onClick={onClose}>
             Change Connection
           </Button>
-          <Button onClick={provisionWallet}>Create</Button>
+          <Button onClick={handleCreateButtonClicked}>Create</Button>
         </DialogActions>
       )}
     </Dialog>
--- a/packages/wallet/ui/src/components/ProvisionDialog.jsx
+++ b/packages/wallet/ui/src/components/ProvisionDialog.jsx
@@ -21,6 +21,9 @@ const errors = {
   NO_SIGNER: 'Cannot sign a transaction in read only mode, connect to keplr.',
 };
 
+// TODO: Read this from the chain via rpc.
+const CREATION_FEE = '10 BLD';
+
 const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
   const [currentStep, setCurrentStep] = useState(steps.INITIAL);
   const [error, setError] = useState('');
@@ -80,7 +83,7 @@ const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
             </DialogContentText>
             <DialogContentText sx={{ pt: 2 }}>
               There is no smart wallet provisioned for this address yet. A fee
-              of <b>10 BLD</b> is required to create one.
+              of <b>{CREATION_FEE}</b> is required to create one.
             </DialogContentText>
           </div>
         );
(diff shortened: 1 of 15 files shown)


packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+  const provisionWallet = async () => {
+    const {
+      signers: { interactiveSigner },
+    } = keplrConnection;
+    if (!interactiveSigner) {
+      setError(errors.NO_SIGNER);
+      return;
+    }
+
+    setError('');
```
turadg: why is this necessary? seems an odd place for it. do you want to reset the error when the user initiates a new action?

why use empty string instead of null or undefined?

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+              <b>Network Config</b>:{' '}
+              <Link href={href} underline="none" color="rgb(0, 176, 255)">
+                {href}
+              </Link>
+            </DialogContentText>
+            <DialogContentText sx={{ pt: 2 }}>
+              <b>Wallet Address:</b> {address}
+            </DialogContentText>
+            <DialogContentText sx={{ pt: 2 }}>
+              There is no smart wallet provisioned for this address yet. A fee
```
turadg, line 87: if we're not going to read it from params, at least make it a const. should have a comment indicating the tech debt and why

packages/wallet/ui/src/components/SmartWalletConnection.jsx
```diff
@@ -45,11 +54,32 @@ const SmartWalletConnection = ({
...
+  const { href, smartConnectionMethod } = connectionConfig;
+  let publicAddress;
+  if (
+    smartConnectionMethod === SmartConnectionMethod.KEPLR &&
+    keplrConnection
+  ) {
+    publicAddress = keplrConnection.address;
+  } else if (smartConnectionMethod === SmartConnectionMethod.READ_ONLY) {
+    publicAddress = connectionConfig.publicAddress;
+  }
```
turadg: consider a helper method so this is one line here assigning to a const

packages/wallet/ui/src/contexts/Provider.jsx
```diff
@@ -153,7 +152,9 @@ const Provider = ({ children }) => {
     restoredConnectionConfig !== null,
   );
   const [connectionStatus, setConnectionStatus] = useState(
-    wantConnection ? 'connecting' : 'disconnected',
```
turadg, line 156: 👍

packages/wallet/ui/src/tests/App.test.jsx
```diff
@@ -28,7 +28,6 @@ const withApplicationContext =
         allWalletConnections={[
           'http://unit-test-net.agoric.com/network-config',
         ]}
-        connectionState="Connecting"
```
turadg, line 31: 👍 one kind of status

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -391,6 +412,43 @@ export const makeInteractiveSigner = async (
...
+
+      const b64address = toBase64(toAccAddress(address));
+
+      const act1 = {
+        typeUrl: SwingsetMsgs.MsgProvision.typeUrl,
+        /** @type {Provision} */
+        value: {
+          address: b64address,
+          nickname: 'my wallet',
+          // XXX dry with PowerFlags const
```
turadg: how about now?

turadg (commented):
Doesn't seem ready for merge. I left some comments. Also the git commits are noisy. Please clean those up for the changelog

samsiegart:
> Also the git commits are noisy. Please clean those up for the changelog

You mean just squash? I'm not sure how else.

packages/wallet/ui/src/util/keyManagement.js
samsiegart: One concern I have is when we import these packages just to use a const or type etc, we increase the bundle size by a lot. Maybe it's better to just leave it?

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Yea we want to reset the error when the user initiates a new action.

I sometimes get in the habit of using empty string because certain input elements don't play well with null, but no reason otherwise. Typically I never set things to `undefined` explicitly but maybe that's okay as well.

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Made a const and TODO

packages/wallet/ui/src/components/SmartWalletConnection.jsx
samsiegart: Done

packages/wallet/ui/src/components/ProvisionDialog.jsx
turadg: > Yea we want to reset the error when the user initiates a new action.

Then please put it where they initiate the action. Either at the top of this function or sooner.

> certain input elements don't play well with null, but no reason otherwise

Please use `null` for "no error". Empty string could be an error that just doesn't have any description. If there are elements where this causes problems I'd like to fix them

packages/wallet/ui/src/util/keyManagement.js
turadg: Doesn't the bundler have tree shaking? If not please make a ticket to get that working and include it in the comment here as to why these aren't imported yet

turadg:
> You mean just squash? 

I was thinking an interactive rebase with renaming. E.g. `git rebase --interactive master` or a GUI like Github Desktop. The latter lets you drag commits together to squash particular ones.

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Changed to use null instead of empty string

I didn't want to have two setErrors in the case of the early return above. I just split this into another function to make it more readable if that's any better?

packages/wallet/ui/src/util/keyManagement.js
samsiegart: Looks like it does have tree shaking yes. Importing this only increased the bundle size by ~4kB. Still the bundle size is getting up there, so I created a separate issue for that https://github.com/Agoric/agoric-sdk/issues/6291

samsiegart:
> I was thinking an interactive rebase with renaming. E.g. `git rebase --interactive master` or a GUI like Github Desktop. The latter lets you drag commits together to squash particular ones.

Ah okay, yea just tried to clean up the messages now

turadg (approved):
Approved assuming the PR will be squashed with a `feat` title.

PR #6292, opened 2022-09-21
test(smart-wallet): local chain launch script

turadg:
## Description

We increasingly need a working smart wallet available locally and we have three files that repeat the steps to do it.

This creates a new single file they can each reference and tries to do it in a way that it can be run as one command on the CLI. So far it only works on MacOS but it still has value on Linux as a consolidation.

### Security Considerations

local

Simplifies

### Testing Considerations

Manually

### DIFF
--- a/packages/agoric-cli/test/agops-governance-smoketest.sh
+++ b/packages/agoric-cli/test/agops-governance-smoketest.sh
@@ -10,20 +10,8 @@ if [ -z "$AGORIC_NET" ]; then
 cd agoric-sdk
 yarn install && yarn build
 
-# (new tab)
-# Start the chain
-cd packages/cosmic-swingset
-make scenario2-setup scenario2-run-chain-psm
-
-# (new tab)
-cd packages/cosmic-swingset
-# Fund the pool
-make fund-provision-pool
-# Provision your wallet
-agoric wallet provision --account <key-name>
-# verify
-agoric wallet list
-agoric wallet show --from <key-name>
+# local chain running with wallet provisioned
+packages/smart-wallet/scripts/start-local-chain.sh YOUR_ACCOUNT_KEY
 "
     exit 1
 fi
--- a/packages/agoric-cli/test/agops-perf-smoketest.sh
+++ b/packages/agoric-cli/test/agops-perf-smoketest.sh
@@ -10,26 +10,8 @@ if [ -z "$AGORIC_NET" ]; then
 cd agoric-sdk
 yarn install && yarn build
 
-# (new tab)
-# Start the chain
-cd packages/cosmic-swingset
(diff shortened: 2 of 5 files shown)


dckc (approved):
I'm always reluctant to take on maintenance of shell code, but I suppose this gets rid of as much as it adds.

samsiegart:
I tried it on MacOS from my `agoric-sdk` dir:

```
samsiegart@Samuels-MacBook-Pro agoric-sdk % packages/smart-wallet/scripts/start-local-chain.sh agoric14l44ylg9kexqafvdxwacrzg5g0fal8ud0wl586
readlink: illegal option -- f
usage: readlink [-n] [file ...]
readlink: illegal option -- f
usage: readlink [-n] [file ...]
Error: agoric14l44ylg9kexqafvdxwacrzg5g0fal8ud0wl586 is not a valid name or address: Invalid key: key with addressAFEB527D05B64C0EA58D33BB81891443D3DF9F8Dnot found: key not found [agoric-labs/cosmos-sdk@v0.45.4-alpha.agoric.1/crypto/keyring/keyring.go:465]
Usage:
  agd keys show [name_or_address [name_or_address...]] [flags]

Flags:
  -a, --address                  Output the address only (overrides --output)
      --bech string              The Bech32 prefix encoding for a key (acc|val|cons) (default "acc")
  -d, --device                   Output the address in a ledger device
  -h, --help                     help for show
      --multisig-threshold int   K out of N required signatures (default 1)
  -p, --pubkey                   Output the public key only (overrides --output)

Global Flags:
      --home string              The application home directory (default "/Users/samsiegart/.agoric")
      --keyring-backend string   Select keyring's backend (os|file|test) (default "os")
      --keyring-dir string       The client Keyring directory; if omitted, the default 'home' directory will be used
      --log_format string        The logging format (json|plain) (default "plain")
      --log_level string         The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
      --output string            Output format (text|json) (default "text")
      --trace                    print out full stack trace on errors

CHAIN_LOG /Users/samsiegart/Library/Logs/Agoric/local-chain.log
SDK 
WALLET agoric14l44ylg9kexqafvdxwacrzg5g0fal8ud0wl586
WALLET_BECH32 
Starting the chain...
packages/smart-wallet/scripts/start-local-chain.sh: line 30: cd: /packages/cosmic-swingset: No such file or directory
```

Any ideas what I'm doing wrong?

turadg:
For `readlink` error, maybe you're on older than 12.6? `readlink --help` should have spit out the options including `-f`. 

For the key problem, the script expects the key _name_ in the keychain, not the bech32. It converts to bech32 as needed.

PR #6296, opened 2022-09-21
fix: arb passable tools

erights:
Move the generation of fast-check arb passables into a separate module in a new store/tools directory. This establishes a pattern of putting arb generators into tools/ directories, where they are still only to support testing, can be exported and reused by other packages for their own testing purposes.

### DIFF
--- a/packages/store/test/test-rankOrder.js
+++ b/packages/store/test/test-rankOrder.js
@@ -1,6 +1,6 @@
-// @ts-nocheck
+// @ts-check
 import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
-import { Far, makeTagged } from '@endo/marshal';
+import { makeTagged } from '@endo/marshal';
 import { fc } from '@fast-check/ava';
 import {
   FullRankCover,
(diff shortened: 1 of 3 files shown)


packages/store/tools/arb-passable.js
```diff
@@ -0,0 +1,89 @@
...
+  fc.constant(new Promise(() => {})),
+);
+
+const { arbDag } = fc.letrec(tie => {
+  return {
+    arbDag: fc.oneof(
+      {
+        // depthFactor: 0.5, // TODO Replace with what from new API?
+        withCrossShrink: true,
+      },
```
gibson042: There are guides for `depthFactor` → `depthSize` at [Arbitraries](https://github.com/dubzzz/fast-check/blob/a6fd30cedbbcf32d2376a5eff2cfe2c4c7a6dc50/packages/fast-check/documentation/Arbitraries.md#depth-size-explained) and [Migration 2.x to 3.x](https://github.com/dubzzz/fast-check/blob/c2600091f0410b301142254cde5cc61cd0385f78/packages/fast-check/MIGRATION_2.X_TO_3.X.md#depthsize-the-new-name-for-depthfactor-%EF%B8%8F), but [`oneof` documentation](https://github.com/dubzzz/fast-check/blob/a6fd30cedbbcf32d2376a5eff2cfe2c4c7a6dc50/packages/fast-check/documentation/Arbitraries.md#:~:text=oneof) suggests that it is necessary anyway: "_probability to select the first arbitrary will increase as we go deeper in the tree_" (which I also confirmed manually).
```suggestion
      { withCrossShrink: true },
```

packages/store/tools/arb-passable.js
```diff
@@ -0,0 +1,89 @@
...
+ * by contents.
+ */
+export const exampleAlice = Far('alice', {});
+export const exampleBob = Far('bob', {});
+export const exampleCarol = Far('carol', {});
+
+export const arbString = fc.oneof(fc.string(), fc.fullUnicodeString());
+
+export const arbLeaf = fc.oneof(
+  fc.record({}),
```
gibson042: Given the increasing weight applied to the first `oneof` option with depth and its preference in shrinking, this should be something which minimizes cognitive burden. `{}` isn't bad, but would a primitive (or even a particular primitive such as `null` or `""`) be better?

packages/store/tools/arb-passable.js
```diff
@@ -0,0 +1,89 @@
...
+            payload: fc.oneof(
+              fc.record({
+                keys: fc.array(tie('arbDag')),
+                values: fc.array(tie('arbDag')),
+              }),
+              tie('arbDag'),
+            ),
+          }),
+          fc.record({ type: arbString, payload: tie('arbDag') }),
+        )
```
gibson042: This can be reorganized now that we know more about `oneof`.
```suggestion
      // A tagged value, either of arbitrary type with arbitrary payload
      // or of known type with arbitrary or explicitly valid payload.
      // Ordered by increasing complexity.
      fc
        .oneof(
          fc.record({ type: arbString, payload: tie('arbDag') }),
          fc.record({
            type: fc.constantFrom('copySet'),
            payload: fc.oneof(
              tie('arbDag'),
              // copySet valid payload is an array of unique passables.
              fc.uniqueArray(tie('arbDag')),
            ),
          }),
          fc.record({
            type: fc.constantFrom('copyBag'),
            payload: fc.oneof(
              tie('arbDag'),
              // copyBag valid payload is an array of [passable, count] tuples
              // in which each passable is unique.
              fc.uniqueArray(
                fc.tuple(tie('arbDag'), fc.bigInt()),
                { selector: entry => entry[0] },
              ),
            ),
          }),
          fc.record({
            type: fc.constantFrom('copyMap'),
            payload: fc.oneof(
              tie('arbDag'),
              // copyMap valid payload is a {keys: Array<Passable>, values: Array<Passable>} record
              // in which keys are unique and both arrays have the same length.
              fc
                .uniqueArray(
                  fc.record({ key: tie('arbDag'), value: tie('arbDag') }),
                  { selector: entry => entry.key },
                )
                .map(entries => ({
                  keys: entries.map(({key}) => key),
                  values: entries.map(({value}) => value),
                })),
            ),
          }),
        )
```

packages/store/tools/arb-passable.js
```diff
@@ -0,0 +1,89 @@
...
+        // depthFactor: 0.5, // TODO Replace with what from new API?
+        withCrossShrink: true,
+      },
+      arbLeaf,
+      tie('arbDag').map(v => Promise.resolve(v)),
+      fc.array(tie('arbDag'), { maxLength: 3 }),
+      fc.dictionary(
+        arbString.filter(s => s !== 'then'),
+        tie('arbDag'),
+        { maxKeys: 3 },
```
gibson042: The improvements for recursive arbitraries also allow us to remove these size constraints.
```suggestion
      fc.array(tie('arbDag')),
      fc.dictionary(
        arbString.filter(s => s !== 'then'),
        tie('arbDag'),
```

packages/store/test/test-rankOrder.js
```diff
@@ -106,7 +50,7 @@ test('compareRank is transitive', async t => {
     fc.property(
       // operate on a set of three passables covering at least two ranks
```
gibson042:
```suggestion
      // operate on a set of three distinct passables covering at least two ranks
```

gibson042 (changes requested):
Looks good after a few tweaks.

packages/store/tools/arb-passable.js
erights: I didn't really understand the explanation, but I took your advice. Done.

packages/store/tools/arb-passable.js
erights: I think I took this advice correctly but PTAL. I just moved the order around.

Resolving, but please unresolve if I misunderstood.

packages/store/tools/arb-passable.js
erights: Your new code is strictly better, so I took that. But copySet, copyBag, and copyMap also insist that their keys be reverse sorted, so we should eventually add some of those. Resolving with your change + added TODOs.

packages/store/tools/arb-passable.js
erights: Done.

packages/store/test/test-rankOrder.js
erights: Done.

erights:
Took all your suggestions. PTAL

erights:
ping

erights:
ping

gibson042 (approved):
LGTM!

PR #6301, opened 2022-09-21
DEBUGGING: instanceAdmin loses in reallocate

Chris-Hibbert:
closes: #XXXX
refs: #XXXX

### DIFF
--- a/packages/SwingSet/src/liveslots/virtualObjectManager.js
+++ b/packages/SwingSet/src/liveslots/virtualObjectManager.js
@@ -18,6 +18,9 @@ import { parseVatSlot } from '../lib/parseVatSlots.js';
 // later regenerated.
 const unweakable = new WeakSet();
 
+// label representatives to simplify identification during debugging
+const LABEL_REPRESENTATIVES = true;
+
 /**
  * Make a simple LRU cache of virtual object inner selves.
  *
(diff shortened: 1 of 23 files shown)


erights:
Was this PR originally intended to solve #4383 ? #4383 is still open. Is this PR still relevant to resolving it?

Chris-Hibbert:
I think #6302 gives the context.  What's the status of #5875?

erights:
> I think #6302 gives the context. What's the status of #5875?

#5875 is blocked until we have adequate configurability so we can turn this on in development but off in production. We have half-implemented support for using environment variables for this https://github.com/endojs/endo/blob/master/packages/ses/src/environment-options.js , but have never made it reusable. Probably needs to be made into a separate package first.

erights:
@Chris-Hibbert What is the status of this. Is it still relevant? Are you blocked waiting on me to review?

Chris-Hibbert:
I'm not blocked on this for anything. If it's helpful to you in diagnosing anything related to #5875, then it's worth keeping. Otherwise, it can be closed.

erights:
> I'm not blocked on this for anything. If it's helpful to you in diagnosing anything related to #5875, then it's worth keeping. Otherwise, it can be closed.

I don't think it'll help. If it would, I can still grab it after it's closed. Closing.

PR #6307, opened 2022-09-23
fix(vats): handle duplicate provision requests

### LINKED ISSUE #6295: Provisioning a Smart Wallet for an address multiple times breaks the address
If I provision a smart wallet once, everything works well.  If I provision the same address twice, the wallet seems to get bugged.
I'm using the instagoric faucet provision-one interface to reproduce this however I suspect the Pay 10 BLD for a wallet may be vulnerable to this as well.

Expected: Provisioning an address many times should not break the wallet

dckc:
closes: #6295

## Description

consistently provision depositFacet

### Testing Considerations

added a regression test, though it's not as thorough as it could be

### DIFF
--- a/packages/smart-wallet/package.json
+++ b/packages/smart-wallet/package.json
@@ -33,6 +33,7 @@
     "@agoric/notifier": "^0.5.0",
     "@agoric/store": "^0.8.0",
     "@agoric/vat-data": "^0.4.0",
+    "@agoric/vats": "^0.11.0",
     "@agoric/zoe": "^0.25.0",
     "@endo/far": "^0.2.10"
   },
--- a/packages/smart-wallet/src/walletFactory.js
+++ b/packages/smart-wallet/src/walletFactory.js
@@ -21,6 +21,13 @@ const PrivateArgsShape = harden(
   ),
 );
 
+/**
+ * Make NameHub for this address and insert depositFacet
+ *
+ * @param {string} address
+ * @param {import('./smartWallet.js').SmartWallet} wallet
+ * @param {ERef<NameAdmin>} namesByAddressAdmin
+ */
 export const publishDepositFacet = async (
   address,
   wallet,
@@ -149,7 +156,8 @@ export const start = async (zcf, privateArgs) => {
             shared,
           );
 
-          await publishDepositFacet(address, wallet, namesByAddressAdmin);
+          // An await here would deadlock with invitePSMCommitteeMembers
+          void publishDepositFacet(address, wallet, namesByAddressAdmin);
 
           makerCalled = true;
           return wallet;
--- a/packages/smart-wallet/src/walletFactory.js
+++ b/packages/smart-wallet/src/walletFactory.js
@@ -140,7 +140,7 @@ export const start = async (zcf, privateArgs) => {
        *   and creating a new one.
        */
       provideSmartWallet(address, bank, namesByAddressAdmin) {
-        let created = false;
+        let makerCalled = false;
         /** @type {() => Promise<import('./smartWallet').SmartWallet>} */
         const maker = async () => {
           const invitationPurse = await E(invitationIssuer).makeEmptyPurse();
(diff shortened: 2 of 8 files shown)


packages/smart-wallet/package.json
```diff
@@ -33,6 +33,7 @@
     "@agoric/notifier": "^0.5.0",
     "@agoric/store": "^0.8.0",
     "@agoric/vat-data": "^0.4.0",
+    "@agoric/vats": "^0.11.0",
```
turadg, line 36: vats depends on smart-wallet. Can we avoid the circular dep? Perhaps by passing a kit-maker in privateArgs? Is there any package `makeMyAddressNameAdminKit` could live in for re-use?

packages/smart-wallet/src/walletFactory.js
```diff
@@ -112,35 +124,38 @@ export const start = async (zcf, privateArgs) => {
...
-       * @param {ERef<MyAddressNameAdmin>} myAddressNameAdmin
-       * @returns {Promise<import('./smartWallet').SmartWallet>}
+       * @param {ERef<NameAdmin>} namesByAddressAdmin
+       * @returns {Promise<[import('./smartWallet').SmartWallet, boolean]>} wallet
+       *   along with a flag to distinguish between looking up an existing wallet
+       *   and creating a new one.
        */
-      provideSmartWallet(address, bank, myAddressNameAdmin) {
+      provideSmartWallet(address, bank, namesByAddressAdmin) {
+        let created = false;
```
turadg: this term is a little confusing as every smart wallet is created at some point. 

consider `providerCreated` or the inverse `providerReused`.

packages/vats/src/provisionPool.js
```diff
@@ -70,20 +67,16 @@ const makeBridgeProvisionTool = (sendInitialPayment, publishMetrics) => {
...
-          ),
-        ]);
-        provisionedCount += 1n;
+        const [_, created] = await E(walletFactory).provideSmartWallet(
+          address,
+          bank,
+          namesByAddressAdmin,
+        );
+        if (created) {
+          provisionedCount += 1n;
```
turadg, line 76: "provision" is being used for the one-time event and the recurring "provideSmartWallet" that is the provider pattern (idempotent). It may help to distinguish these terms more.

e.g. here `totalCreated`.

turadg (approved):

packages/smart-wallet/package.json
dckc: I considered moving `makeMyAddressNameAdminKit` into `nameHub.js`, but that's in `@agoric/vats` too. I can imagine moving `nameHub.js` to a new package (or to internal), but that much surgery at this point in the release cycle doesn't seem wise.

The kit-maker is powerless, so putting it in `privateArgs` doesn't seem right.

So I think accepting a bit more tech debt (#4645) is the least bad option here.

packages/smart-wallet/src/walletFactory.js
dckc: how about `makerCalled`?

packages/vats/src/provisionPool.js
dckc: That would be a contract API change, which seems out of scope of this issue/PR and fairly risky at this point in the release process.

I suppose refining the meaning of `provisionedCount` is also an API change, but it seems like a much smaller one.

dckc:
just realized this should go on the `release-psmo` branch... converted back to draft to rebase...

packages/smart-wallet/package.json
turadg: Agreed it's the least bad option. I don't think moving the package has risk for the release because the release is using built assets and if anything goes wrong in the package refactoring they just won't build. But it will take time we don't want to spend. Incidentally, `internal` won't work because that's only for modules with no sdk dependencies and it depends on `@agoric/store`. Of course that may be in Endo eventually.

packages/vats/src/provisionPool.js
turadg: I just meant for the internal value, not the external contract which would still be `walletsProvisioned`.

Consider,
```
if (makerCalled) {
  makersCalled += 1n;
}
```

turadg (approved):

PR #6310, opened 2022-09-23
feat: upgrade handler for upgrade-8

### LINKED ISSUE #6285: Cosmos Upgrade Handler

To install Pismo on the chain, we need a stub upgrade handler that calls agoric-upgrade-8.

This helps drive the chain upgrade at a certain block height, even if the upgrade handler itself is empty.

JimLarson:
closes: #6285

## Description

Cosmos upgrade handler for agoric-upgrade-8.

Reviewer - is noted section necessary if we're not doing a store transformation?

### Testing Considerations

Must do an all-up test.

### DIFF
--- a/golang/cosmos/app/app.go
+++ b/golang/cosmos/app/app.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	"github.com/cosmos/cosmos-sdk/simapp"
+	store "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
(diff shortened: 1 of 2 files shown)


golang/cosmos/app/app.go
```diff
@@ -727,7 +728,28 @@ func NewAgoricApp(
...
 
-	// Upgrade handlers will go here
+	app.UpgradeKeeper.SetUpgradeHandler(
+		upgradeName,
+		func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+			return vm, nil
+		},
+	)
+
+	// REVIEWER - is this required?
```
arirubinstein: lines 738-752 should be removed

this is to add a new store type from ica, and in this (cross) upgrade we shouldn't be doing anything

JimLarson: Done.

JimLarson:
Reviewers, PTAL.

Corrected the PR target to be the `release-pismo` branch instead of `master`.

PR #6316, opened 2022-09-23
feat(agoric-cli): support --home, --keyring-backend

### LINKED ISSUE #6315: Agoric wallet provision does not support --home and --keyring-backend
When attempting to provision a smart wallet with 
```sh
% AGORIC_NET=ollinet,ollinet-41 agoric wallet provision --account agoric109ngmpqqq88x5vntezj0cc64y43xwm8xnqypmj --spend --keyring-backend test --home st.state                                                                                                                                                    1 ↵ ✹
error: unknown option '--keyring-backend'
```

it fails. Folks will need access to configure these flags for their local cosmos installs/configuration

dckc:
for `agoric wallet provision` and `agoric wallet send`

closes: #6315
refs: #6211

### Security Considerations

Key management for `--keyring-backend=test` is considerably less secure than other options, but presumably folks who ask for it know what they're asking for.

`--help` stuff added

### Testing Considerations

manually tested

### DIFF
--- a/packages/agoric-cli/src/commands/wallet.js
+++ b/packages/agoric-cli/src/commands/wallet.js
@@ -42,7 +42,6 @@ export const makeWalletCommand = async () => {
       '--keyring-backend [os|file|test]',
       'keyring\'s backend (os|file|test) (default "os")',
     )
-    .option('--dry-run', 'spit out the command instead of running it')
     .option('--nickname [string]', 'nickname to use', 'my-wallet')
     .action(function () {
       const {
@@ -50,12 +49,11 @@ export const makeWalletCommand = async () => {
         nickname,
         spend,
         home,
-        dryRun,
         keyringBackend: backend,
       } = this.opts();
+      const tx = `provision-one ${nickname} ${account} SMART_WALLET`;
       if (spend) {
-        const tx = `provision-one ${nickname} ${account} SMART_WALLET`;
-        execSwingsetTransaction(tx, networkConfig, account, dryRun, {
+        execSwingsetTransaction(tx, networkConfig, account, false, {
           home,
           backend,
         });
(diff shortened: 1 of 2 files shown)


packages/agoric-cli/src/commands/wallet.js
```diff
@@ -70,15 +86,27 @@ export const makeWalletCommand = async () => {
       normalizeAddress,
     )
     .requiredOption('--offer [filename]', 'path to file with prepared offer')
+    .option('--home [dir]', 'agd application home directory')
+    .option(
+      '--keyring-backend [os|file|test]',
+      'keyring\'s backend (os|file|test) (default "os")',
+    )
```
dckc, line 93: not terribly DRY...

packages/agoric-cli/src/commands/wallet.js
```diff
@@ -37,12 +37,28 @@ export const makeWalletCommand = async () => {
       normalizeAddress,
     )
     .option('--spend', 'confirm you want to spend')
+    .option('--home [dir]', 'agd application home directory')
+    .option(
+      '--keyring-backend [os|file|test]',
+      'keyring\'s backend (os|file|test) (default "os")',
+    )
+    .option('--dry-run', 'spit out the command instead of running it')
```
turadg: doesn't `spend` option handle this case?

packages/agoric-cli/src/commands/wallet.js
```diff
@@ -70,15 +86,27 @@ export const makeWalletCommand = async () => {
       normalizeAddress,
     )
     .requiredOption('--offer [filename]', 'path to file with prepared offer')
+    .option('--home [dir]', 'agd application home directory')
```
turadg, line 89: I think these options can be put on the `'wallet'` command so they don't have to be repeated for each action

packages/agoric-cli/src/commands/wallet.js
dckc: no; if you omit `--spend`, it doesn't spit out the command.
Oh... I suppose it could.

turadg: yeah I think that would be better. the message could say ~"re-run this with --spend or run the following command:"

packages/agoric-cli/src/commands/wallet.js
dckc: I'm not having any luck hoisting the options like that.

dckc:
oops; I think I cleared @arirubinstein 's review when I meant to renew the request for @turadg to review. wrong button.

turadg (approved):

arirubinstein (approved):

Issue #6290, opened 2022-09-21
Governance GUI for econ committee


turadg:
## What is the Problem Being Solved?

https://github.com/Agoric/agoric-sdk/issues/6108 a the CLI (TUI) for performing all the operations needed for governance. [agops-governance-smoketest.sh](https://github.com/Agoric/agoric-sdk/blob/5c54498c3a3b5c85858309e8bce4bd793a4ae00a/packages/agoric-cli/test/agops-governance-smoketest.sh) shows how to use them but while working on https://github.com/Agoric/agoric-sdk/issues/6281 we realized the command-line tools require so much setup that we should offer a browser-based UI.

## Description of the Design

https://github.com/agoric-labs/gov-demo1

- its own signing and skips the Wallet UI

TODO
- [x] make the UI able to connect to devnet and mainnet (and for developers: ollinet, local)

Consider
- [ ] adding a Connect to Wallet button instead of the popup (or at least some documentation to click Keplr icon)
- [ ] rendering the invitation that's being accepted
- [ ] detecting when the faucet is needed and blocking UI until it is

#### Tab: Vote
- [x] Accept your invitation (until it's in the wallet, then show when it was added)
- [x] Show latest question
- [ ] Show all open questions (NOT BLOCKER)
- [x] Operation to choose among positions on open questions and submit transaction to vote
- [x] Showing outcomes of past votes (maybe in another tab that doesn't require invitations)

#### Tab: Propose Change
- [x] Accept your invitation (as above)
- [x] Choose pair (updates the rest of the tab)
- [x] Change MintLimit
 - [x] show current
 - [x] input for the new one
- [x] Change GiveMintedFee
- [x] Change WantMintedFee
- [x] Change offer filter
 - [ ] show current filter (NOT BLOCKER)
 - [ ] ~give examples of how to pause the four directions~

## Security Considerations

TBD

## Test Plan

Go through the smoketest script using the GUI and make sure all the operations perform as expected.

turadg:
necessary feature complete

PR #6317, opened 2022-09-23
fix: wallet show coalesced updates

turadg:
refs: #6290 

## Description

Main change is to make the coalesceUpdates utility track the status of invitations, in support of #6290. While in there I noticed the updater was getting a list going back in time, so the "last wins" was actually "latest wins". No bueno. This reverses the updates list before coalescing.

Also refactored the CLI to use casting and cleaned up some logging and types.

### Security Considerations
--

--

### Testing Considerations

These are UIs that don't have automated tests. Tested manually.


packages/agoric-cli/src/lib/rpc.js
```diff
@@ -236,20 +235,23 @@ harden(storageHelper);
...
       const parts = storageHelper.unserializeTxt(content, ctx).at(-1);
-
+      for (const [name, remote] of parts) {
+        reverse[remote.boardId] = name;
+      }
       return [kind, Object.fromEntries(parts)];
     }),
   );
-  return Object.fromEntries(entries);
+  return { ...Object.fromEntries(entries), reverse };
```
dckc, line 254: this puts mappings in both directions in the same object? That seems surprising.

packages/smart-wallet/src/utils.js
```diff
@@ -29,7 +28,7 @@ export const makeWalletStateCoalescer = () => {
         if (currentAmount.brand.boardId === 'board02810') {
```
dckc: The zoe invitation brand has `board02810`? give it a constant name?

packages/zoe/src/zoeService/offer/burnInvitation.js
```diff
@@ -17,7 +17,7 @@ import { E } from '@endo/eventual-send';
 export const burnInvitation = (invitationIssuer, invitation) => {
   const handleRejected = reason => {
     const err = assert.error(
-      X`A Zoe invitation is required, not ${invitation}`,
+      X`A Zoe invitation is required, not ${q(invitation)}`,
```
dckc: This seems risky - we don't know what it is we're divulging here.

packages/casting/src/follower-cosmjs.js
```diff
@@ -82,7 +82,7 @@ const proofs = ['strict', 'none', 'optimistic'];
  * @param {any} sourceP
  * @param {import('./types').LeaderOrMaker} [leaderOrMaker]
  * @param {import('./types').FollowerOptions} [options]
- * @returns {Follower<FollowerElement<T>>}
```
dckc, line 85: this didn't turn out to be quite right...

returns {ValueFollower<T>} and FollowerElement is a ValueFollowerElement

dckc (approved):
please consider tweaks as discussed

packages/zoe/src/zoeService/offer/burnInvitation.js
turadg: removed

packages/smart-wallet/src/utils.js
turadg: changed to get `allegedInvitationBrand` from the petname

PR #6323, opened 2022-09-25
BREAKING CHANGE: rename from FarClass to ExoClass, etc

### LINKED ISSUE #1193: revise terminology of Far
# Problem to solve
What we call `Far` has gone through some iterations and still has room for improvement. One source of confusion is that a Far object can be near (in the same vat). It's also been known as "primary remotable", which doesn't suffer from that confusion but is much less wieldy. 

According to @erights , 
> The term “Far object” was always misleading, and therefore unfortunate terminology. The source is the Far function, which is used to make one. We originally called them “primary remotable” … But I never found myself using that in casual conversation. Nor have I used “remote” by itself for the other side. Whenever I tried, I’d have to take a beat to explain my terminology, and it never lasted from one conversation to another. For “remote”, I have adopted “remote presence” whose meaning seems reliably remembered. But it is a remote presence of what? What is the remotable that the remote presence designates? When you E(remotePresence).foo(args), what should we call the object with the foo method that receives this message?
> the distinctions … for explaining abstractions for app programmer use, are different than those … for talking about the kernel implementation. The remotable object that a remote reference designates, that received its messages with one of its methods, may be an in-memory object, or a virtual object, or a durable object.
> Now that we have `defineKind` and `defineDurableKind`, we use the `Far` function only to make the in-memory one of these. @warner calls these in-memory objects made by Far “precious objects” … By “Far object”, I mean the category that includes all three of these, i.e., the object that is designated by a remote presence, no matter how it is implemented. Most of its client semantics is independent of these implementation choices. Since virtual and durable such objects are not tied to the function named Far, now would be a good time to choose another name for this concept.

Our [documentation on Far](https://agoric.com/documentation/guides/js-programming/far.html) isn't _wrong_ but could be made more clear:
> On the specific issue, it looks like this uses “remotable” in a way that might not technically be wrong, but would lead one to think that only far objects are remotables. IOW, I’d guess people exposed to just this document would not realize that remote presences are also remotables. Once that is realized, then “remotable” is no longer adequate to mean specifically the object that remote presences designate. This document has no noun for that. It never uses the term “far object”.

# Acceptance criteria

- terms cover all cases
- internally consistent and unambiguous
- practical to use for kernel developers
- just enough complexity for new app programmers

erights:
Fixes https://github.com/endojs/endo/issues/1193 .

@dtribble and I were brainstorming name alternatives to `FarClass` (which we both hate for somewhat different reasons). There are three salient differences from normal object programming, where `FarClass` only suggests the first and third. 
   * It is a remotable, meaning it is exposed to messages from outside the vat (hence "Far")
   * Related, it is protected by an interfaceGuard that does the first type-like phase of input validation
   * It is programmed in a class-like manner, with inline methods in the concise-method syntax, accepting their instance-specific access via their `this` binding (hence "Class")

We also need to name the instance case, and `FarInstance` is way too awkward. The instance case has the first two features but not the third.

New aha: "Exo" does double duty. 
   * As a prefix meaning "outside", it suggests the remotable meaning. That was why I suggested Exo in the first place.
   * It lets us introduce the metaphor of the [ExoSkeleton](https://en.wikipedia.org/wiki/Exoskeleton) without being confused with the exoskeleton itself. "An exo is a creature with an exoskeleton" is a memorable explanation, even if there is no RL noun "Exo".

Also
   * It is short, so for the class case "ExoClass" still works.
   * It is distinct from our current "Far" and so the "Instance" qualifier isn't needed for the protective instance case.

But this last bullet (distinct from "Far") means that it isn't an answer to our first search for a term, since it doesn't cover remotable objects without a protective exoskeleton (currently, "Far objects" and "remote presences"). Oh well. I think the rename is still worth it.

### DIFF
--- a/packages/ERTP/src/payment.js
+++ b/packages/ERTP/src/payment.js
@@ -1,5 +1,5 @@
 import { initEmpty } from '@agoric/store';
-import { prepareFarClass } from '@agoric/vat-data';
+import { prepareExoClass } from '@agoric/vat-data';
 
 /** @typedef {import('@agoric/vat-data').Baggage} Baggage */
 
@@ -12,7 +12,7 @@ import { prepareFarClass } from '@agoric/vat-data';
  * @returns {() => Payment<K>}
  */
 export const preparePaymentKind = (issuerBaggage, name, brand, PaymentI) => {
-  const makePayment = prepareFarClass(
+  const makePayment = prepareExoClass(
     issuerBaggage,
     `${name} payment`,
     PaymentI,
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -2,10 +2,7 @@
 import { isPromise } from '@endo/promise-kit';
 import { assertCopyArray } from '@endo/marshal';
 import { fit, M } from '@agoric/store';
-import {
-  provideDurableWeakMapStore,
-  prepareFarInstance,
-} from '@agoric/vat-data';
+import { provideDurableWeakMapStore, prepareExo } from '@agoric/vat-data';
 import { AmountMath } from './amountMath.js';
 import { preparePaymentKind } from './payment.js';
 import { preparePurseKind } from './purse.js';
@@ -84,7 +81,7 @@ export const preparePaymentLedger = (
   optShutdownWithFailure = undefined,
 ) => {
   /** @type {Brand<K>} */
-  const brand = prepareFarInstance(issuerBaggage, `${name} brand`, BrandI, {
+  const brand = prepareExo(issuerBaggage, `${name} brand`, BrandI, {
     isMyIssuer(allegedIssuer) {
       // BrandI delays calling this method until `allegedIssuer` is a Remotable
       return allegedIssuer === issuer;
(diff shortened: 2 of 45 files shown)


kriskowal (approved):
Looks mechanical to me. I’m warming to the idea that this name change is a net improvement and, stew as I have on this question for long enough, have not found a more positive improvement.

kriskowal:
Caveat: this appears to be a breaking change and I do not know how it affects upgrade. However, the commit message should be framed like:

```
refactor(ertp)!: Rename etc

*BREAKING CHANGE*: How to update dependencies
```

This will ensure that Lerna generates a changelog and the appropriate version bumps.

erights:
See previous discussion at https://github.com/Agoric/agoric-sdk/pull/6118
See table of options at https://docs.google.com/spreadsheets/d/1yhKgzMOfmZOTnvP3Bh4p16H9RyYq_WhOlxZ3Qicx_Oc/edit#gid=0

gibson042 (approved):
I think "Exo" is a big improvement over "Far", and even the isolated noun usage works pretty well.

erights:
Please look at https://github.com/Agoric/agoric-sdk/pull/6818 for another candidate. I think I like that one better.

turadg (changes requested):
Late in the day, I confused this and #6818 and even #6118.

turadg (approved):
This simple change looks good to me. As I wrote on https://github.com/Agoric/agoric-sdk/pull/6323#pullrequestreview-1264534582 I ask that the naming rationale be documented within the codebase. E.g. `packages/store/docs/terminology.md`.

PR #6324, opened 2022-09-26
Deterministic xsnap bundles

### LINKED ISSUE #6298: consensus failure due to build-host pathname sensitivity
In today's kernel meeting, we realized that the `lockdown` and `supervisor` bundles (both created with `bundleSource({ format: 'getExports' })` and thus using `rollup`) are likely to embed the directory name of the original source tree, which of course varies from one machine to another (it would have `/home/warner/` on my machine, but not yours). These bundles are evaluated into the `xsnap` worker process, and the strings therein become part of the JS state.

That means, at the very least, the `xsnap` heap snapshots will differ between executions whose bundles were created on different machines. If you initialize your validator on machine A, it won't produce the same heap snapshot hashes as the ones from machine B. Even if you copy the state DB from A to B and then run both on B, they'll have different pathnames in their bundles.

Worse, the pathnames might be of different lengths, which means the strings used in stack traces/etc are longer, which means more memory must be allocated to hold them, which might cause organic GC to happen slightly earlier. That will turn into a syscall divergence, eventually.

I believe we currently catch this very early, because we write the full contents of these two bundles into the `kvStore` during `initializeSwingset()`, and we record an `activityHash` that includes all writes to the kvStore.

Apparently validators discovered that they get immediate consensus errors unless they unpack their source tree into a directory that matches the docker images that our CI uses, suggesting that this sensitivity already exists but has been worked around by the community.

We'd like to not be sensitive to those pathnames. The `EndoZipBase64`-format bundler (which does *not* use `rollup`) does not suffer from this problem, but we can't easily use it for `lockdown` and `supervisor` because that format requires more than a simple `eval` to load into the new `xsnap` worker. (The supervisor bundle includes the loader for `EndoZipBase64`, which makes it possible to deliver the vat bundle in that format, later).

@michaelfig said that @kriskowal has a plan for Endo to generate `getExport` -format bundles without `rollup`, which would remove these embedded directory names, and allow lockdown/supervisor bundles made on different machines to have identical contents. If we could switch to that today, we could (hopefully) remove this validator workaround.

Further work might yield `nestedEvaluate`-format bundles too, which would improve our debugging experience (we don't always get useful line numbers out of `getExport` format bundles) when we see problems in liveslots or the supervisor code.

michaelfig:
closes: #6298
refs: endojs/endo#1294

## Description

Adopts the changes from endojs/endo#1296 as a patch to the SDK, and enables the use of the improved `nestedEvaluate` to make xsnap lockdown and supervisor bundles not leak local path information.

### Security Considerations

Reduces an unnecessary information leak of where the bundle was created.

Backwards compatible.

### Testing Considerations

We need to test as part of the loadgen or instagoric to see if xsnap snapshots are truly deterministic after this change even if the SDK is built in different locations.

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -147,6 +147,7 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
 
     for (const bundle of bundles) {
       bundle.moduleFormat === 'getExport' ||
+        bundle.moduleFormat === 'nestedEvaluate' ||
         assert.fail(X`unexpected: ${bundle.moduleFormat}`);
       // eslint-disable-next-line no-await-in-loop, @jessie.js/no-nested-await
       await worker.evaluate(`(${bundle.source}\n)()`.trim());
--- a/packages/SwingSet/src/controller/initializeSwingset.js
+++ b/packages/SwingSet/src/controller/initializeSwingset.js
@@ -30,8 +30,10 @@ const allValues = async obj =>
 
 const bundleRelative = rel =>
   bundleSource(new URL(rel, import.meta.url).pathname);
-const bundleRelativeGE = rel =>
-  bundleSource(new URL(rel, import.meta.url).pathname, { format: 'getExport' });
+const bundleRelativeCallable = rel =>
+  bundleSource(new URL(rel, import.meta.url).pathname, {
+    format: 'nestedEvaluate',
+  });
 
 /**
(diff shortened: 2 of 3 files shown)


warner (approved):
Good, I like it.

warner:
Looks like trunk acquired a new Endo since this branch forked off, and patch refers to `@endo/bundle-source` 2.3.0, but trunk is using 2.3.1 . @michaelfig could you rebase your branch to trunk and refresh the patch?

PR #6325, opened 2022-09-26
feat(telemetry): concentrate slog handling in separate process

### LINKED ISSUE #6304: Offload slog processing out of Swingset

Traces show that a significant amount of time is being spent in slog related activities in a validator node. All slog entries originate from the controller's `writeSlogObject`, which generates a JSON blob from an object, writes it to a slog file, and passes it to a "sender" function, which in practice writes it to an nmapped file (flight-recorder) and generates open-telemetry traces, all that synchronously.

## Description of the Design

Factor JSON serializing and slog file writing out of the kernel, into the telemetry "slogSender", and add a path to process these slog events in a separate worker or process, freeing the main thread for actual kernel work.

## Security Considerations

This might introduce a flushing delay into the slog file, but that isn't a security concern, more a debuggability concern.

## Test Plan

Verify that data still flows into the slog file and other telemetry senders. The loadgen and some other tests rely on following a slog file, so we can rely on those

mhofman:
closes: #6304

## Description

This change consolidates all slog event handling in the `telemetry` package, including serialization and slog file writing. It also adds a new option to perform slog event processing in a sub-process, leaving only the slog object creation and a minimal serialization (using v8 structured cloning) in the main kernel process.

Added forced flushing support to the slog file "sender", with a trigger after commiting every block to make sure the slog data is not lost in case of unclean termination.

### Security Considerations

None I can think of

Updated the `env.md` describing the new `SLOGSENDER_AGENT` option, and the change to the existing `SLOGSENDER` option.

### Testing Considerations

Updated the integration deployment test to use the new sub-process processing. Tested locally with the loadgen.

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -164,8 +164,7 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
  *   verbose?: boolean,
  *   debugPrefix?: string,
  *   slogCallbacks?: unknown,
- *   slogFile?: string,
- *   slogSender?: (obj: any, jsonObj: string) => void,
+ *   slogSender?: import('@agoric/telemetry').SlogSender,
  *   testTrackDecref?: unknown,
  *   warehousePolicy?: { maxVatsOnline?: number },
  *   overrideVatManagerOptions?: unknown,
@@ -190,7 +189,6 @@ export async function makeSwingsetController(
     verbose,
     debugPrefix = '',
     slogCallbacks,
-    slogFile,
     slogSender,
     spawn = ambientSpawn,
     warehousePolicy = {},
@@ -200,12 +198,8 @@ export async function makeSwingsetController(
     throw Error('SES must be installed before calling makeSwingsetController');
   }
 
-  const slogF =
-    // eslint-disable-next-line @jessie.js/no-nested-await
-    slogFile && (await fs.createWriteStream(slogFile, { flags: 'a' })); // append
-
   function writeSlogObject(obj) {
-    if (!slogSender && !slogF) {
+    if (!slogSender) {
       // Fast path; nothing to do.
       return;
     }
@@ -218,20 +212,8 @@ export async function makeSwingsetController(
     // rearrange the fields a bit to make it more legible to humans
     const timedObj = { type: undefined, ...obj, time, monotime };
 
-    // Serialise just once.
-    const jsonObj = JSON.stringify(timedObj, (_, arg) =>
-      typeof arg === 'bigint' ? Number(arg) : arg,
-    );
-
-    if (slogSender) {
-      // Allow the SwingSet host to do anything they want with slog messages.
-      slogSender(timedObj, jsonObj);
-    }
-    if (slogF) {
-      // Record the pristine JSON object.
-      slogF.write(jsonObj);
-      slogF.write('\n');
-    }
+    // Allow the SwingSet host to do anything they want with slog messages.
+    slogSender(timedObj);
   }
 
   // eslint-disable-next-line no-shadow
@@ -543,10 +525,17 @@ export async function makeSwingsetController(
  *
  * @param {SwingSetConfig} config
  * @param {string[]} argv
- * @param {{ hostStorage?: HostStore, env?: Record<string, string>, verbose?:
- *   boolean, kernelBundles?: Record<string, Bundle>, debugPrefix?: string,
- *   slogCallbacks?: unknown, testTrackDecref?: unknown, warehousePolicy?: {
- *   maxVatsOnline?: number }, slogFile?: string }} runtimeOptions
+ * @param {{
+ *   hostStorage?: HostStore;
+ *   env?: Record<string, string>;
+ *   verbose?: boolean;
+ *   kernelBundles?: Record<string, Bundle>;
+ *   debugPrefix?: string;
+ *   slogCallbacks?: unknown;
+ *   slogSender?: import('@agoric/telemetry').SlogSender;
+ *   testTrackDecref?: unknown;
+ *    warehousePolicy?: { maxVatsOnline?: number };
+ * }} runtimeOptions
  * @typedef { import('@agoric/swing-store').KVStore } KVStore
  */
 export async function buildVatController(
@@ -562,7 +551,7 @@ export async function buildVatController(
     debugPrefix,
     slogCallbacks,
     warehousePolicy,
-    slogFile,
+    slogSender,
   } = runtimeOptions;
   const { kernel: kernelBundle, ...otherBundles } = kernelAndOtherBundles;
   const kernelBundles = runtimeOptions.kernelBundles ? otherBundles : undefined;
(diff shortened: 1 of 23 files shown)


mhofman:
Best reviewed commit-by-commit.

I know @michaelfig is the most familiar with the slog pipeline and sub-process pipes, but @kriskowal might be able to review the latter, and @warner the former.

cc @arirubinstein for operationalizing this. I believe instagoric may want to set `SLOGSENDER_AGENT=process`

arirubinstein (approved):
This is awesome, and I can't wait to before/after the throughput tests

packages/SwingSet/test/vat-admin/test-create-vat.js
```diff
@@ -61,13 +61,9 @@ async function doTestSetup(t, doVatAdminRestart = false, enableSlog = false) {
...
   let doSlog = false;
-  function slogSender(_, s) {
+  function slogSender(slogObj) {
     if (!doSlog) return;
-    const o = JSON.parse(s);
-    delete o.time;
-    delete o.replay;
-    delete o.crankNum;
-    delete o.deliveryNum;
+    const { time: _1, replay: _2, crankNum: _3, deliveryNum: _4, o } = slogObj;
```
michaelfig: I think you mean for `o` to be the rest of the properties:
```js
    const { time: _1, replay: _2, crankNum: _3, deliveryNum: _4, ...o } = slogObj;
```

packages/telemetry/src/make-slog-sender.js
```diff
@@ -1,32 +1,121 @@
...
 
-  console.warn(`Loading makeSlogSender from ${slogSenderModule}`);
-  const { makeSlogSender: maker } = await import(slogSenderModule);
-  if (typeof maker !== 'function') {
-    throw Error(`${slogSenderModule} did not export a makeSlogSender function`);
+  console.warn(
+    `Loading slog sender${
+      slogSenderModules.length > 1 ? 's' : ''
+    } from ${slogSenderModules.join(', ')}`,
+  );
```
michaelfig: Please don't do math to construct natural language sentences.  It makes i18n more difficult, as well as being harder to parse mechanically.
```js
  console.warn('Loading slog sender modules:', slogSenderModules.join(', '));
```

packages/telemetry/src/make-slog-sender.js
```diff
@@ -1,32 +1,121 @@
...
+    potentialSenders =>
+      /** @type {SlogSender[]} */ (potentialSenders.filter(Boolean)),
+  );
+
+  if (!senders.length) {
+    return undefined;
+  } else if (senders.length === 1) {
+    return senders[0];
+  } else {
+    const hasSenderNeedingJsonObj = senders.some(sender => sender.length === 2);
```
michaelfig: Oh no!  An arity check?!!

This deserves a `FIXME` comment explaining why the dodginess is worth it, and what other (more explicit) mechanisms could replace it.  And also `sender.length >= 2`.

packages/telemetry/src/make-slog-sender.js
```diff
@@ -1,32 +1,121 @@
...
+        try {
+          sender(slogObj, jsonObj);
+        } catch (err) {
+          console.error('WARNING: slog sender error', err);
+        }
+      }
+    };
+    if (hasSenderNeedingJsonObj) {
+      Object.defineProperty(slogSender, 'length', { value: 2 });
+    }
```
michaelfig: Playing with the function arity idiom is pretty unclear.

Could you instead use `Object.assign(slogSender, { needsJsonObject: true })`, and do the same for the other slogSenders?

michaelfig (approved):
I like it overall!

My main criticism is that the `sender.length` function-arity-sniffing can be clarified by making it a more explicit property on the sender.

packages/SwingSet/test/vat-admin/test-create-vat.js
mhofman: Ooops

packages/telemetry/src/make-slog-sender.js
mhofman: Yeah I felt icky doing an arity check. I'll move to an explicit property on the sender function

mhofman:
PTAL @michaelfig

michaelfig (approved):
Thanks, LGTM!

packages/SwingSet/src/controller/controller.js
```diff
@@ -164,8 +164,7 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
  *   verbose?: boolean,
  *   debugPrefix?: string,
  *   slogCallbacks?: unknown,
- *   slogFile?: string,
- *   slogSender?: (obj: any, jsonObj: string) => void,
+ *   slogSender?: import('@agoric/telemetry').SlogSender,
```
warner, line 167: Does this make packages/swingset dependent upon packages/telemetry? That doesn't feel like the right direction to me.. and I don't want users who only want swingset to wind up with OTEL libraries or whatnot. It doesn't need to happen now, but I'm inclined to make this type live in swingset, and have telemetry import it from there. Or, as we've discussed elsewhere for TimerService-related types, make a third package, upon which both swingset and telemetry can depend.

packages/SwingSet/src/controller/controller.js
```diff
@@ -562,7 +551,7 @@ export async function buildVatController(
     debugPrefix,
     slogCallbacks,
     warehousePolicy,
-    slogFile,
+    slogSender,
```
warner, line 554: I see why `slogFile` had to be removed, but I gotta admit that's going to interfere with some debugging I frequently do within swingset unit tests, where it's awfully handy to add a one-line change to the `buildSwingsetController` call to record the test actions. I guess I'll write a helper function in `test/` to make that trick possible again, the next time I need it.

packages/SwingSet/src/controller/controller.js
mhofman: This is just a type import so it doesn't introduce a runtime dependency. We can definitely move the type definition to SwingSet or something else.

packages/SwingSet/src/controller/controller.js
mhofman: It should be pretty easy to write a `slogSender` for tests that need to record slog events. It'd be a better abstraction anyway than relying on an intermediate slog file. Probably something along the lines of `test-create-vat.js`'s `slogSender`.

warner:
for my reference, here's the code I pasted into a test to get a slogfile:
```js
import { promises as fs } from 'fs';
const slogF = (await fs.open('out.slog', 'a')).createWriteStream();
const ser = obj => JSON.stringify(obj, (_, arg) => typeof arg === 'bigint' ? Number(arg) : arg);
const slogSender = entry => slogF.write(ser(entry) + '\n');
// then add `slogSender` to `runtimeOptions`
```

it's longer than I'd like, and not as trivial to remember as `slogFile='out.slog'`, so I'd like to consider re-introducing `runtimeOptions.slogFile`.

mhofman:
What's wrong with the following?

```js
import { makeSlogSender } from '@agoric/telemetry';

const slogSender = await makeSlogSender({
  env: {SLOGFILE: 'out.log'},
});
```

PR #6327, opened 2022-09-26
fix: test-rankOrder works on ava 3 or 4

erights:
https://github.com/endojs/endo/pull/1260 would move test-rankOrder to the endo repository, which is currently on ava 3, so the test in that PR has been changed to work on ava 3 or 4. The change needed IMO was also a net improvement in code quality, so at least in this case there's no cost. This PR ports that back to test-rankOrder.js in agoric-sdk to minimize the maintenance hazard while this file exists in both places.

See also 
https://github.com/endojs/endo/pull/1291
https://github.com/Agoric/agoric-sdk/pull/6296


gibson042 (approved):
LGTM!

PR #6328, opened 2022-09-26
fix(wallet-ui): show rejected offers properly

### LINKED ISSUE #6319: Wallet UI did not provide feedback on rejected offer
Constructed an offer with a give that included more funds than I had (due to a bug that is being filed separately).  When offer was submitted, I signed in Keplr.  The offer shows rejected in the CLI, but I was never given feedback.  The offer card remained as pending in the UI.  Clicking "Exit" did not do anything.  On reload, the "Pending" offer was still there.

Offers should resolve when they are rejected and show as clearly finished in the UI.

samsiegart:
fixes https://github.com/Agoric/agoric-sdk/issues/6319

### DIFF
--- a/packages/wallet/ui/src/components/Proposal.jsx
+++ b/packages/wallet/ui/src/components/Proposal.jsx
@@ -181,7 +181,7 @@ const Proposal = ({ offer, purses }) => {
           <pre>{JSON.stringify(args, null, 2)}</pre>
         </div>
       )}
-      {error !== undefined && (
+      {error && (
         <div className="OfferEntry">
           <h6>Error</h6>
           <details style={{ whiteSpace: 'pre-wrap' }}>{error}</details>
--- a/packages/wallet/ui/src/components/Proposal.jsx
+++ b/packages/wallet/ui/src/components/Proposal.jsx
@@ -104,6 +104,7 @@ const Proposal = ({ offer, purses }) => {
     proposalForDisplay,
     proposalTemplate,
     invitationDetails: { fee, feePursePetname, expiry } = {},
+    error,
   } = offer;
 
   let give = {};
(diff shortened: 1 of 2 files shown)


packages/wallet/ui/src/components/Proposal.jsx
```diff
@@ -180,6 +181,12 @@ const Proposal = ({ offer, purses }) => {
           <pre>{JSON.stringify(args, null, 2)}</pre>
         </div>
       )}
+      {error !== undefined && (
```
turadg: why not show the error if it's ever a string or other truthy object?
```suggestion
      {error && (
```

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -202,8 +202,17 @@ export const makeWalletBridgeFromFollower = (
           const { status } = updateRecord;
           console.log('offerStatus', { status, offers });
           const oldOffer = offers[status.id];
-          if (
-            oldOffer &&
+          if (!oldOffer) {
+            break;
```
turadg, line 207: here we have an offerStatus update for an unknown offer. I think we'd benefit from `console.warn` that this happened

turadg (approved):

packages/wallet/ui/src/util/WalletBackendAdapter.js
samsiegart: Done

packages/wallet/ui/src/components/Proposal.jsx
samsiegart: Good point, just copypasta, done.

turadg (approved):

PR #6331, opened 2022-09-27
chore: Sync Endo versions

kriskowal:
This change brings Agoric SDK in sync with the most recent Endo releases, including support for smallcaps, replaces a patch for a SES performance improvement around uncurryThis, and other improvements.

In this change, I’ve added instructions for this process to `MAINTAINERS.md` reflecting what I did to generate this change.


FUDCo (approved):
Looks good to me.  Well, maybe not good, but acceptable.  Well, maybe not acceptable, but necessary.  Humans should not have to see this.

PR #6332, opened 2022-09-27
feat(wallet-ui): bridge storage without click when possible

### LINKED ISSUE #6320: Hard refresh on psm.inter.trade causes it to wait for wallet connection even though it is connected
Application hangs until user clicks 'signal wallet', which is unnatural when it's already connected and user can see the dapp in their wallet connections.

Is it possible to have the app detect or attempt to signal on its own?

samsiegart:
fixes https://github.com/Agoric/agoric-sdk/issues/6320


dckc (approved):
Looks worth a try.

PR #6348, opened 2022-09-28
fix: make it possible to set decimalPlaces when calling startPSM

### LINKED ISSUE #6337: adding USDT with 18 decimals gets 1:1 anchorPerMinted despite 6 decimals for IST
In ollinet, I added a PSM for USDT with `decimalPlaces: 18` using `startPSM()` as usual, not realizing that it sets `anchorPerMinted` 1:1 irrespective of decimalPlaces:

https://github.com/Agoric/agoric-sdk/blob/1b3cbd27bda8f4e1abe08eff8260ced40b720bb1/packages/inter-protocol/src/proposals/startPSM.js#L86

I tried to trade 0.2 USDT for IST and the trade failed with **Request would exceed mint limit**, which puzzled me, because I set the mint limit to `123456_000_000n`. The mint limit is in `uist`, which prompted me to realize that this trade was asking for a **lot** of IST (roughly 2e11 = 0.2 * 1e18/1e6).

cc @warner @rowgraus @turadg

Chris-Hibbert:
closes: #6337

## Description

startPSM wasn't setting decimals when creating a new trading pair.

Also, `gov-add-psm.js` had the wrong keywords in its config.

### Security Considerations

None.

None

### Testing Considerations

Manual


packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -79,11 +79,19 @@ export const startPSM = async (
       E(E(zoe).getInvitationIssuer()).getAmountOf(poserInvitationP),
     ]);
 
+  const [anchorInfo, stableInfo] = await Promise.all(
```
turadg: s/stable/minted

packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -79,11 +79,19 @@ export const startPSM = async (
...
+    [anchorBrand, stable].map(b => E(b).getDisplayInfo()),
```
turadg: consider a note about why "display" info is reliable for this use case

turadg (approved):
regression test pretty please

dckc:
> regression test pretty please

experience with #6187 makes that look particularly challenging. I can't find any way to diagnose why it works on my desktop and not in ci.

packages/inter-protocol/src/proposals/startPSM.js
Chris-Hibbert: I don't have an answer for that. Can you write something that would justify it?

packages/inter-protocol/src/proposals/startPSM.js
Chris-Hibbert: okay.  It starts from `STABLE.symbol`, but I'll rename it locally.

dckc (approved):
I managed to test this end-to-end, locally.

PR #6349, opened 2022-09-28
fix: avoid __proto__ accessor

erights:
Salesforce deletes `Object.prototype.__proto__` for safety reasons. We may do so as well. In any case, we should not assume it has not been overridden. So delete all uses of this accessor property that would break if it were absent or deleted.


erights:
See also https://github.com/endojs/endo/issues/1303

erights:
@FUDCo ping. These are the only cases I know of that would break if `Object.prototype.__proto__` were deleted. 

(Though I haven't tested. Testing may very well reveal others.)

FUDCo (approved):
Ya shoor, you betcha.

PR #6351, opened 2022-09-28
fix: vbank updates for new zero balance

### LINKED ISSUE #6321: a balance update to zero after PSM trade never published

In Ollinet @rowgraus made a trade and the "want" side of the transaction published a balance update (IST up) but give did not (aUSD down). 

https://gist.github.com/turadg/72f81dcf32a5b86cab74377f4a1090cd

A subsequent attempt to trade confirmed the aUSD funds were no longer available, so integrity is preserved, but the UI shows an incorrect state. According to RPC output, the wallet got IST for free, giving up no aUSD.

After making another trade (in reverse) the published balances were correct again.

## Description of the Design

Finish diagnosing the problem: where did the balance update drop?
1. Do the logs show what happened to the balance update message?
2. Did the contract call the publisher?
3. Did the RPC node miss it?
4. Is the code querying RPC missing it?

We've talked about publishing full states periodically (https://github.com/Agoric/agoric-sdk/issues/6184) but we've always assumed that every update will be published and queryable.

JimLarson:
closes: #6321

## Description

Issue #6321 arose from an impedance mismatch between the swingset world which typically works a denom at a time, and the Cosmos world which handles collections of arbitrary denoms as a unit, representing only those with a nonzero quantity. To give balance updates for denoms which have dropped to zero quantity, we need to extract from an event the set of denoms that it might have affected, then at the end of a block we send updates for the whole set of such denoms, including those for which the current balance is zero.

This PR also changes which events we monitor to notice balance changes. The "transfer" events are only generated for `MsgSend` and for the output addresses of `MsgMultiSend`. By monitoring the lower-level "coin_spent/received" events, we still handle `MsgSend`, but now handle all participants of `MsgMultiSend` as well as delegation, undelegation, minting, and burning for both user and module accounts.

### Security Considerations

Missing notifications is not a good user experience, but does not lead to any integrity issues.

None.

### Testing Considerations

None.

### DIFF
--- a/golang/cosmos/x/vbank/module.go
+++ b/golang/cosmos/x/vbank/module.go
@@ -123,49 +123,34 @@ func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.V
 
 	/* Scan for all the events matching (taken from cosmos-sdk/x/bank/spec/04_events.md):
 
-	### MsgSend
-
-	| Type     | Attribute Key | Attribute Value    |
-	| -------- | ------------- | ------------------ |
-	| transfer | recipient     | {recipientAddress} |
-	| transfer | sender        | {senderAddress}    |
-	| transfer | amount        | {amount}           |
-	| message  | module        | bank               |
-	| message  | action        | send               |
-	| message  | sender        | {senderAddress}    |
-
-	### MsgMultiSend
-
-	| Type     | Attribute Key | Attribute Value    |
-	| -------- | ------------- | ------------------ |
-	| transfer | recipient     | {recipientAddress} |
-	| transfer | sender        | {senderAddress}    |
-	| transfer | amount        | {amount}           |
-	| message  | module        | bank               |
-	| message  | action        | multisend          |
-	| message  | sender        | {senderAddress}    |
+	type: "coin_received"
+	  "receiver": {recipient address}
+	  "amount": {Coins string}
+	type: "coin_spent"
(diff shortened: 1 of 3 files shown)


golang/cosmos/x/vbank/module.go
```diff
@@ -106,55 +107,51 @@ func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.V
...
-		if err != nil {
-			return err
+		currentDenoms := sdk.NewCoins()
+		if coins, ok := addressToBalance[address]; ok {
+			currentDenoms = coins
 		}
-		coins := am.keeper.GetAllBalances(ctx, account)
-		addressToBalance[address] = coins
-		return nil
+		addressToBalance[address] = currentDenoms.Add(denoms...)
```
michaelfig: Maybe rename `addressToBalance` to `addressToDenoms`, since you're tracking the denoms and the first transferred amount to fill out the balances later.  Also `ensureBalanceIsPresent` to `ensureDenomsArePresent`.

michaelfig (approved):
LGTM.

Just some suggested renaming for clarity.  Not critical for merging.

golang/cosmos/x/vbank/module.go
JimLarson: I took a look at the usage in context, and `addressToUpdate` and `ensureAddressUpdate` felt best. Changed.

PR #6352, opened 2022-09-29
fix(telemetry): handle misbehaving slog senders

mhofman:
## Description

#6325 introduced a change that highlighted an existing bug in the slog senders: `flight-recorder.js` had a `forceFlush` which didn't return a Promise, and `chain-main.js` had un unconditional `.catch` on the return value of `forceFlush`. I'm not exactly sure how we never triggered this before given that the default slog sender should have been flight recorder, but the fix is easy enough.

Drive-by change to make sure slogSender calls never throw back into swingset, as discussed in https://github.com/Agoric/agoric-sdk/pull/6247#discussion_r981845593 (cc @warner)


warner (approved):
Yeah, looks good

PR #6354, opened 2022-09-29
fix(wallet-ui): show all significant digits in nat amounts

### LINKED ISSUE #6336: Wallet Offer doesn't show enough precision
When making a PSM trade offer, the wallet offer should match the precision of the denom.  Instead it's either truncated or rounded

Screen Shot 2022-09-27 at 1 14 54 PM

samsiegart:
fixes https://github.com/Agoric/agoric-sdk/issues/6336
Screenshot:
image

### DIFF
--- a/packages/ui-components/src/display/natValue/stringifyNat.js
+++ b/packages/ui-components/src/display/natValue/stringifyNat.js
@@ -16,7 +16,7 @@ const calcTrailingZeros = value => {
 };
 
 /**
- * @param {NatValue} natValue
+ * @param {NatValue?} [natValue]
  * @param {number} [decimalPlaces]
  * @param {number} [placesToShow]
  * @returns {string}
--- a/packages/ui-components/src/display/natValue/stringifyNat.js
+++ b/packages/ui-components/src/display/natValue/stringifyNat.js
@@ -4,6 +4,17 @@ import { roundToDecimalPlaces } from './helpers/roundToDecimalPlaces.js';
 const CONVENTIONAL_DECIMAL_PLACES = 2;
 const MAX_DECIMAL_PLACES = 100;
 
+const calcTrailingZeros = value => {
+  if (!value) return 0;
+
+  let zeroes = 0;
+  while (value > 0 && value % 10n === 0n) {
+    zeroes += 1;
+    value /= 10n;
+  }
+  return zeroes;
+};
+
 /**
  * @param {NatValue} natValue
  * @param {number} [decimalPlaces]
@@ -13,10 +24,17 @@ const MAX_DECIMAL_PLACES = 100;
 export const stringifyNat = (
   natValue = null,
   decimalPlaces = 0,
-  placesToShow = CONVENTIONAL_DECIMAL_PLACES,
+  placesToShow,
 ) => {
   if (natValue === null) {
     return '';
+  }
+
+  if (placesToShow === undefined) {
+    placesToShow = Math.max(
+      Math.min(decimalPlaces, CONVENTIONAL_DECIMAL_PLACES),
+      decimalPlaces - calcTrailingZeros(natValue),
+    );
   }
 
   if (placesToShow > MAX_DECIMAL_PLACES) {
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/display/natValue/stringifyNat.js
```diff
@@ -4,21 +4,39 @@ import { roundToDecimalPlaces } from './helpers/roundToDecimalPlaces.js';
 const CONVENTIONAL_DECIMAL_PLACES = 2;
 const MAX_DECIMAL_PLACES = 100;
 
+const calcTrailingZeros = value => {
+  if (!value) return 0;
```
turadg: why would the argument be falsy? from `stringifyNat` it looks like the `value` is guaranteed to be `NatValue`. to confirm please type the function signature. 

the top of the file will need `// @ts-check` too

packages/ui-components/src/display/natValue/stringifyNat.js
```diff
@@ -4,21 +4,39 @@ import { roundToDecimalPlaces } from './helpers/roundToDecimalPlaces.js';
...
+
+  let zeroes = 0;
+  while (value > 0 && value % 10n === 0n) {
```
turadg: I suspect there's a more efficient way to calculate this but I can't think of it :)

packages/ui-components/src/display/natValue/stringifyNat.js
```diff
@@ -4,21 +4,39 @@ import { roundToDecimalPlaces } from './helpers/roundToDecimalPlaces.js';
...
   natValue = null,
   decimalPlaces = 0,
-  placesToShow = CONVENTIONAL_DECIMAL_PLACES,
+  placesToShow,
 ) => {
   if (natValue === null) {
     return '';
   }
 
+  if (placesToShow === undefined) {
```
turadg, line 36: this would fit as a default assignment, but I agree it's better here. also lets you validate that `natValue` isn't null.

packages/ui-components/test/display/natValue/test-stringifyNat.js
```diff
@@ -39,3 +39,14 @@ test('stringifyNat limit decimal places to 100', t => {
 
   t.is(stringifyNat(tinyCoin, 105, 120), `0.${'1234'.padStart(100, '0')}`);
 });
+
+test('stringifyNat shows correct number of decimal places by default', t => {
+  t.is(stringifyNat(123n), '123');
+  t.is(stringifyNat(1230000n, 4), '123.00');
+  t.is(stringifyNat(123456700n, 6), '123.4567');
```
turadg: thanks for these tests. consider describing in words what they're testing.

turadg (approved):

packages/ui-components/src/display/natValue/stringifyNat.js
samsiegart: Actually I'm not sure there is, this is technically O(log n) where n is the value of the number. If we have the prime factorization, we can count the number of 2s and 5s which is O(1) time (the number of 2 & 5 pairs we can make from the factors is the number of trailing zeros), but looks like prime factorization takes O(sqrt n) which is slower.

Maybe stringifying and going from there is quicker by a constant factor though.

samsiegart: Actually maybe we could do a binary search across powers of 10, which would be O(log log n)

turadg: Neat ideas! Yeah, I was thinking more in the "constant factor" space since n is so small. Doing anything more here seems like premature optimization to me. I should have been more clear about that in my comment: "I suspect there's a better way but I didn't think of one in a few minutes and I don't think it will make enough difference to put more effort into it"

packages/ui-components/test/display/natValue/test-stringifyNat.js
samsiegart: Done

packages/ui-components/src/display/natValue/stringifyNat.js
samsiegart: Ah good catch, removed and added @ts-check

packages/ui-components/src/display/natValue/stringifyNat.js
samsiegart: Yea of course not worth the trouble, I was just too tempted to think about it some more

PR #6355, opened 2022-09-29
fix: cheap auxdata

### LINKED ISSUE #2069: add (immutable) auxiliary data to Presences?
# What is the Problem Being Solved?

Pass-by-presence objects are currently (almost) pure behavior. If you hold a Presence, the (almost) only thing you can do with it is:

* send it a message
* compare it for identity against some other Presence
  * e.g. use it as the key of a Map or WeakMap

The one partial exception is that we associate an "Interface name" (a sort of alleged type) with the Presence, which can be retrieved with a helper function (#1816).

On the other hand, if you send a pass-by-copy object (data but no behavior), it behaves like a normal javascript object, except you can't send it messages, and you can't compare it for identity (each act of transmission results in a new copy of the object.. if we could, we'd probably want to throw an error if you ever tried to use `===` on it, or use it as the key of a Map/WeakMap).

I've wondered if it would make development easier if we added immutable data to pass-by-presence ~~copy~~ objects. This would subsume the current `getInterfaceOf(presence)` call (you'd just do `presence.interface` instead). The vat which creates the Remotable could add whatever non-Function properties it likes, as long as the object is `harden()`ed, and the non-Function properties would be serialized as usual and attached to all copies of the Presence.

This might allow (alleged) Brands to be attached to Purses/Payments, instead of requiring a `getBrand()` roundtrip. We could add descriptive names to objects, for debugging.

This relates to:

* #804 behavior+data is one step closer to an Unum implemenation
* #2018 empty objects and non-Function-bearing objects would be pass-by-copy, empty objects marked as `Remotable` and Function-bearing objects (with or without non-Function data) would be pass-by-presence
* #28 we could add debugging-helping descriptions as plain string properties on the Remotable and it would appear on the Presence in an obvious manner
* #59 reflection information (starting with just an object ID), if it was visible to the exporting vat, could be added as regular properties

## Description of the Design

My notion is:

* we add a "data" column to the kernel object table, which contains a capdata-serialized Object (mapping property name to value)
* vats and kernels do not automatically supply the data to their recipient, instead the recipient asks for a copy if they don't already have it. Vats can use `syscall.getDataForObject` to retrieve the auxililary data the first time they receive an object, and the kernel can do `dispatch.getDataForObject` to fetch it from a vat shortly after the vat exports it for the first time
  * this is a tradeoff. I can think of three basic ways to do it: "always include data", "ask for data if you don't already have it", and "try to guess whether they already have the data". The latter would be the most efficient, but I'm not convinced we can correctly make both sides of all kernel/vat boundaries get this right.
* since the data is supposed to be immutable, it ought to be stable: if a vat is asked for data and it returns something different than it did previously, that ought to be a vat-fatal error (although if we're asking them a second time, we probably don't remember enough about the first time to notice)

Extra feature: the auxilliary data could include additional references (to other objects and/or promises). These would represent new edges in the kernel-side object graph, which influences GC. I think this would be particularly useful for a Brand.

## Security Considerations

This would be an additional way for vats to introduce reference cycles into the kernel data structures (they can already do so with mutually-referential promise resolutions), which may be hard to clean up: cycles require a full mark-and-sweep GC pass, rather than merely watching for refcounts to drop to zero.

This would make it harder to implement mostly-transparent proxy wrappers, such as a recovable forwarder, since the wrapper would need to enumerate all the auxilliary data and copy it (in wrapped form) onto the new proxy. A common bug would be to forget to copy the data properties, leading to a wrapper that *behaves* like the original but *looks* bare.

## Open Questions

Does this seem useful? @erights @katelynsills @Chris-Hibbert are there ERTP things that would be easier or more ergonomic to express if you could do property lookups on your Presence objects?

For me, the high-order question is whether this auxilliary data can contain object/promise references, or if it is restricted to purely pass-by-copy data. If so, the kernel object table will be a source of GC references. I don't think it's significantly more work for me (in #1872) to accomodate this, since the kernel *promise* table is already a source, but I do wonder how much I should plan to accomodate it.

Does the "ask if you don't already have it" protocol feel right? We could explore it more, but my hunch is that having both sides know for sure whether the other side already knows about the data would complicate the GC protocol.

erights:
Liveslots adds to all remote presences an `aux()` method.

A remote presence has an `aux()` method that either returns a promise or a non-promise. The first time it is called, it does an eventual-send of an `aux()` message to the far object it designates, remembers that promise as its auxdata, and returns that. It also watches that promise to react to its fulfillment. Until that reaction, every time its `aux()` is called, it will return that same promise. Once it reacts to the promise being fulfilled, if that ever happens, then the fulfillment becomes its new auxdata which gets returned from then on.

Initially only liveslots does this. But the mechanism is in a reusable `makeRemotePresence` function that should probably move to `@endo/marshal` so that other things that make remote presences can use it.

A good first test will be for ERTP brands to provide an auxdata object with the things that you can currently ask a brand for that are stable, and that zoe needs to cache before it can register the brand. Zoe's caching logic should then become much simpler as it need only delay registration until a local call to the brand's `aux()` returns a non-promise. After that, the remote presence's `aux()` method must always immediately return the same non-promise.

(Some brands are co-located with Zoe, in which case their `aux()` method may not be similarly stable. But that would be a misbehaving brand. The only brands co-located with the Zoe service are made by Zoe and relied on not to misbehave.)

***Back in draft pending needed redesign to play well with virtual and durable storage.***

If we believe this approach is adequate for our auxdata needs, then
Fixes: https://github.com/Agoric/agoric-sdk/issues/2069

@Chris-Hibbert @warner @FUDCo @michaelfig @mhofman when this is ready, I'll ask you to review. Until then, please ignore. @Chris-Hibbert I'm leaving you listed since you already commented.

### DIFF
--- a/packages/zoe/test/unitTests/zcf/test-zcf.js
+++ b/packages/zoe/test/unitTests/zcf/test-zcf.js
@@ -5,8 +5,9 @@ import { Far } from '@endo/marshal';
 import { AssetKind, AmountMath } from '@agoric/ertp';
 import { E } from '@endo/eventual-send';
 import { getMethodNames } from '@agoric/internal';
-import { makeOffer } from '../makeOffer.js';
+import { M } from '@agoric/store';
 
+import { makeOffer } from '../makeOffer.js';
 import { setup } from '../setupBasicMints.js';
 import buildManualTimer from '../../../tools/manualTimer.js';
 
@@ -204,6 +205,13 @@ test(`zcf.saveIssuer - bad issuer, makeEmptyPurse throws`, async t => {
     // eslint-disable-next-line no-use-before-define
     isMyIssuer: i => i === badIssuer,
     getDisplayInfo: () => ({ decimalPlaces: 6, assetKind: AssetKind.NAT }),
+    aux: () =>
+      harden({
+        name: 'bogusBrand',
+        assetKind: AssetKind.NAT,
+        displayInfo: { decimalPlaces: 6, assetKind: AssetKind.NAT },
+        amountShape: M.any(),
+      }),
   });
   const badIssuer = Far('issuer', {
     makeEmptyPurse: async () => {
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -107,6 +107,9 @@ export const preparePaymentLedger = (
     getAmountShape() {
       return amountShape;
     },
+    aux() {
+      return brandAuxData;
+    },
   });
 
   const amountShape = amountShapeFromElementShape(
@@ -114,6 +117,13 @@ export const preparePaymentLedger = (
     assetKind,
     elementShape,
   );
+
+  const brandAuxData = harden({
+    name,
+    assetKind,
+    displayInfo,
+    amountShape,
+  });
 
   const { IssuerI, MintI, PaymentI, PurseIKit } = makeIssuerInterfaces(
     brand,
--- a/packages/ERTP/src/typeGuards.js
+++ b/packages/ERTP/src/typeGuards.js
@@ -156,11 +156,19 @@ export const IssuerKitShape = harden({
 
 // //////////////////////// Interfaces /////////////////////////////////////////
 
+export const BrandAuxDataShape = harden({
+  name: M.string(),
+  assetKind: AssetKindShape,
+  displayInfo: DisplayInfoShape,
+  amountShape: M.pattern(),
+});
+
 export const BrandI = M.interface('Brand', {
   isMyIssuer: M.callWhen(M.await(IssuerShape)).returns(M.boolean()),
   getAllegedName: M.call().returns(M.string()),
   getDisplayInfo: M.call().returns(DisplayInfoShape),
   getAmountShape: M.call().returns(M.pattern()),
+  aux: M.call().returns(BrandAuxDataShape),
 });
 
 /**
--- a/packages/ERTP/src/types-ambient.js
+++ b/packages/ERTP/src/types-ambient.js
@@ -86,6 +86,14 @@
  */
 
 /**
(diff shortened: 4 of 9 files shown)


packages/zoe/test/unitTests/zcf/test-zcf.js
```diff
@@ -13,6 +13,7 @@ import buildManualTimer from '../../../tools/manualTimer.js';
 
 import { setupZCFTest } from './setupZcfTest.js';
 import { assertAmountsEqual } from '../../zoeTestHelpers.js';
+import { M } from '@agoric/store';
```
Chris-Hibbert: CI failure from lint.  Move this up above line 9.

Chris-Hibbert (commented):
The ERTP/Zoe changes look fine.  Someone else should approve the SwingSet changes.

There's a lint (import ordering) failure in CI.

erights:
Converting to draft because of unintuitive interaction with durable state and upgrade :(

(Will file bugs explaining the problems)

mhofman:
> Liveslots adds to all remote presences an `aux()` method.

This will require the program to understand, and have a way to test, whether an object is local or remote. Do we have such a predicate today?



> method that either returns a promise or a non-promise

Sounds like releasing Zalgo in the worst possible shape. I'd much prefer have a this method return a pseudo promise which is guaranteed to have a stable identity, and have an external promise result memoization helper.

Even then I still don't understand how you'll be able to escape our rules that we never await conditionally. Seems like releasing Zalgo is a design choice here?

Edit: this is based on reading the PR description. I did not read the implementation.

erights:
Closing in favor of an actual solution to this problem, which I need to explain.

erights:
Reopening, as my "better idea" does not work at all. See https://github.com/Agoric/agoric-sdk/issues/2069#issuecomment-1581786978

PR #6358, opened 2022-09-29
deterministic xsnap bundles on release-pismo

warner:
This cherry-picks the #6298 fixes (from PR #6324) onto `release-pismo`. With this change, the SES and supervisor bundles we create should no longer embed the directory name where the source tree was checked out. Before this, if one validator did a `git clone https://github.com/Agoric/agoric-sdk` into `/home/alice/agoric-sdk`, and a different one used `/home/bob/agoric-sdk`, then their bundle contents would differ, which caused a consensus failure on the very first block. Validators dealt with this by agreeing to use a specific directory (I think `/usr/src/agoric`). This change should remove that requirement.

refs #6298

### DIFF
--- a/packages/SwingSet/src/controller/controller.js
+++ b/packages/SwingSet/src/controller/controller.js
@@ -147,6 +147,7 @@ export function makeStartXSnap(bundles, { snapStore, env, spawn }) {
 
     for (const bundle of bundles) {
       bundle.moduleFormat === 'getExport' ||
+        bundle.moduleFormat === 'nestedEvaluate' ||
         assert.fail(X`unexpected: ${bundle.moduleFormat}`);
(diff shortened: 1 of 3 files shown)


warner:
I tested this by rebuilding the supervisor bundle, and it now ends with:

```
//# sourceURL=/bundled-source/.../supervisor-subprocess-xsnap.js
```

instead of:
```
//# sourceURL=/home/node/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js
```

warner:
The corresponding fixes are landing on trunk in PR #6324

mhofman (approved):

michaelfig (approved):

Issue #6184, opened 2022-09-12
wallet history: on chain storage vs. update size vs. off-chain searching


dckc:
## What is the Problem Being Solved?

The following are in tension:
 1. how much state does the wallet store?
 2. how large are chainStorage updates?
 3. how much searching does the off-chain client have to do?

The current contract does well on 1 and 2 at the expense of 3. For example, it produces an update for each new asset/isser/brand. To be sure it has all of them, the client may have to search through all wallet updates back to the first one.

## Description of the Design

ideas include:
 - off-chain indexing
 - in each offerStatus update, include the id of the oldest outstanding offer?
 - every N offerStatus changes, publish an aggregate of all balances and `brand` descriptors

### UIs use `.current`
- [x] dapp-econ-gov
- [x] agoric-cli `wallet` and `psm` commands
- [x] Wallet UI
- [x] dapp-psm (should be implied by Wallet UI support)

cc @rowgraus

dckc:
The current design conflicts with normal pruning practices for RPC nodes.

@turadg is working on econ committee governance, and in devnet, he can't get the full wallet state because nodes are pruning somewhere between blocks 20,000 and 30,000 (current block height is in the 40,000s).

cc @michaelfig @gibson042 @arirubinstein

gibson042:
Agreed. We should generally err on the side of publishing complete state information rather than deltas and only deviate upon observing an actual issue.

dckc:
The actual issues that motivated the current design were:

 - #6065
 - #6038

It's possible that the full state size is now manageable, and we might be able to reduce the 6 updates per trade by 1 or 2, but we'd have to do a bunch of performance testing again.

turadg:
> publishing complete state information rather than deltas 

As a matter of terminology, what we're publishing a third way: "partial states". We've already committed to only publishing current _states_ and never _deltas_. Delta being the difference between a previous and current state, such as "+3" for a value. Publishing deltas makes it so that no state is observable without accumulating all the deltas. With partial state, what you receive is known to be true, it's just not all the state.

> only deviate upon observing an actual issue.

Agreed. As Dan referenced above, there was an actual issue that motivated this. Reducing the updates per trade would increase latency in user feedback on the state of the transaction.

Of course, _complete states_ are preferred. But that isn't that independent of the matter of whether archival can be relied upon or not? Can the smart wallet rely on its RPC nodes to be archival? My understand has been **yes**, and if that isn't correct we should confer.

[...]

Issue #6038, opened 2022-08-23
smart-wallet PSM trade performs 17 account-state storage updates


warner:
## Describe the bug

@dckc and I were examining a slogfile containing a single smart-wallet AMM trade. In the current tree (commit afabe7ff1) this needed 171 deliveries, consisting of 84 `dispatch.deliver`s and 87 `dispatch.notify` (promise resolutions). Once we increased the per-block computron limit (by about 10x), this took a total of 1.575 seconds to complete.

Many of those deliveries are to update the chain storage data. The chain storage nodes are owned by v1-bootstrap, so we can grep the slogfile for `toBridge` to find where various vats are asking v1-bootstrap to perform a syscall to send a message through the bridge to the cosmos side. I counted one `toBridge` call to update storage:`published.psm.IST.AUSD.metrics`, three to `bank` (1: `set ibc/usdc1234 = 4000440`, 2: `set ibc/usdc1234 = 0`, 3: `set uist = 4000039`), and 17 to update storage:`published.wallet.agoricNNN` (the address performing the trade).

```
% grep toBridge block30.slog|grep '"type":"deliver"' | cut -c-220|grep published.wallet.agoric1hfcw2ryh8duca65
{"type":"deliver","crankNum":1433,"vatID":"v1","deliveryNum":153,"replay":false,"kd":["message","ko34",{"methargs":{"body":"[\"toBridge\",[\"storage\",{\"key\":\"published.wallet.agoric1hfcw2ryh8duca65gadgttvn4sdgxvafs96
{"type":"deliver","crankNum":1449,"vatID":"v1","deliveryNum":154,"replay":false,"kd":["message","ko34",{"methargs":{"body":"[\"toBridge\",[\"storage\",{\"key\":\"published.wallet.agoric1hfcw2ryh8duca65gadgttvn4sdgxvafs96
{"type":"deliver","crankNum":1481,"vatID":"v1","deliveryNum":155,"replay":false,"kd":["message","ko34",{"methargs":{"body":"[\"toBridge\",[\"storage\",{\"key\":\"published.wallet.agoric1hfcw2ryh8duca65gadgttvn4sdgxvafs96
...
```

The bodies of the `toBridge` calls (and the `setValue` calls which prompted them) grew somewhat over the course of those 17 (the 142/893/139/137 are the non-account ones described above, while the account ones grew from 8872 to 15655):

```
# [vatID, targetObjectID, methodName, methargs.body.length]
[...]

PR #6359, opened 2022-09-29
6184 wallet.current record

turadg:
refs: #6184 (partial)

## Description

Add a `.current` subnode to the wallet node for clients to have a stable HEAD that has all precious state. This is a slow growing set so won't bring on the problems of https://github.com/Agoric/agoric-sdk/issues/6038

The keys are,
- **brands** is a list of brand descriptors. Changes only when new purse is added. Needed for issuers, petnames, etc
- **purses** is a list of purse states. There's currently one per brand, so we could do a map, but we know that in the future we'll have more than one purse per brand so let's not have to change the chain storage to accommodate that. Here we have an array of purses and additional fields can be added to distinguish.
- **offerToUsedInvitation** is a map of offerId to an invitation that has been consumed. This is useful for clients to look up the offer ID for a given invitation spec, by which they can use its invitation makers.
- **lastOfferId** is a way for clients to ensure they're using a sufficiently high offer ID in their next offer. This could be read from the wallet node, but if there have been many balance updates since the last offer there may be no offerStatus updates in available history.

Also updates wallet CLI and GUI to use it.

### Security Considerations

Nothing exposed that wasn't published before in some fashion.

--
### Testing Considerations

PSM integration test covers it. It would be good to have more specific tests but not economical right now.

Manually tested `agoric wallet show` and that Wallet UI shows the purse balances.


packages/smart-wallet/test/test-psm-integration.js
```diff
@@ -220,7 +246,20 @@ test('govern offerFilter', async t => {
...
+  let currentState = await headValue(currentSub);
+  t.is(
+    // @ts-expect-error cast amount kind
+    currentPurseBalance(currentState, invitationBrand).length,
+    1,
+    'one invitation consumed, one left',
+  );
+  t.deepEqual(Object.keys(currentState.offerToUsedInvitation), ['44']);
+  t.is(
+    currentState.offerToUsedInvitation[44].value[0].description,
```
dckc, line 260: Bingo

packages/smart-wallet/scripts/start-local-chain.sh
```diff
@@ -47,7 +47,7 @@ agd query bank balances "$WALLET_BECH32" | grep ubld || exit 1
 echo "Provisioning your smart wallet..."
 agoric wallet provision --spend --account "$WALLET"
 echo "waiting for blocks"
-sleep 5
+sleep 10
```
dckc, line 50: Yup. I tripped on that too.

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -139,36 +142,74 @@ export const makeWalletBridgeFromFollower = (
...
-  const followLatest = async () => {
-    const firstHeight = await getFirstHeight(follower);
+  const fetchCurrent = async () => {
+    console.log('fetchCurrent()');
+    const latestIterable = await E(currentFollower).getLatestIterable();
+    const iterator = latestIterable[Symbol.asyncIterator]();
+    const latest = await iterator.next();
+    /** @type {import('@agoric/smart-wallet/src/smartWallet').CurrentWalletRecord} */
+    const current = latest.value.value;
+    console.log('fetcHCurrent', current);
```
dckc: Looks drafty. Did you mean to keep that?
Not critical.

dckc (approved):
This looks on target.

I'm less confident about the wallet ui, so perhaps stand by for someone else to review that part.

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -139,36 +142,74 @@ export const makeWalletBridgeFromFollower = (
...
     notifierKits.purses.updater.updateState(harden(purses));
   };
 
-  const followLatest = async () => {
-    const firstHeight = await getFirstHeight(follower);
+  const fetchCurrent = async () => {
+    console.log('fetchCurrent()');
+    const latestIterable = await E(currentFollower).getLatestIterable();
+    const iterator = latestIterable[Symbol.asyncIterator]();
+    const latest = await iterator.next();
```
michaelfig, line 186: To prevent a race between the two followers, we need to store the height at which `latest` was published, like the way we used to get from `getFirstHeight`.

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -139,36 +142,74 @@ export const makeWalletBridgeFromFollower = (
...
+        currentAmount: purse.balance,
+        brandPetname: brandDescriptor.petname,
+        pursePetname: brandDescriptor.petname,
+        displayInfo: brandDescriptor.displayInfo,
+      };
+      brandToPurse.set(purse.brand, purseInfo);
+    }
+    console.log('brandToPurse map', brandToPurse);
+    updatePurses();
+  };
```
michaelfig, line 208:
```suggestion
    return firstHeight;
  };
```

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -139,36 +142,74 @@ export const makeWalletBridgeFromFollower = (
...
 
-    for await (const { value } of iterateEach(follower, {
-      height: firstHeight,
-    })) {
+  const followLatest = async () => {
+    console.log('followLatest()');
+    for await (const { value } of iterateEach(updateFollower)) {
```
michaelfig:
```suggestion
  const followLatest = async firstHeight => {
    console.log('followLatest()');
    for await (const { value } of iterateEach(updateFollower, { height: firstHeight })) {
```

michaelfig (approved):
Looks overall good.

I would prefer more careful handoff between the current state and the `{ height: lastHeight }` parameter of the update stream, but even just a comment to that effect would be fine.

dckc:
Since this has on-chain changes, it needs to go on the release-pismo branch, one way or the other.

packages/wallet/ui/src/util/WalletBackendAdapter.js
turadg: nope, thanks, i'll fixup

packages/wallet/ui/src/util/WalletBackendAdapter.js
turadg: excellent, thanks

packages/wallet/ui/src/util/WalletBackendAdapter.js
turadg: excellent, thanks

turadg:
> Since this has on-chain changes, it needs to go on the release-pismo branch, one way or the other.

I tried to make this PR base of release-pismo but it has diverged enough from master that the rebase was somewhat cumbersome. I'll wait for this to get into master and then do a separate PR to get the commits into pismo.

Issue #5833, opened 2022-07-26
UserSeat.getCurrentAllocations should be avoided in production code


erights:
https://github.com/Agoric/agoric-sdk/blob/35b68062f69dc84041b98f670572683b572aa29e/packages/zoe/src/contractSupport/zoeHelpers.js#L357

As explained at https://github.com/Agoric/agoric-sdk/pull/5826#pullrequestreview-1051462243 , this usage is likely not incorrect *only* because we know this happens only after the seat has been exited. The general availability of this method is a hazard ("footgun") likely to lead to bugs in production code.

Indeed #5826 renames this method to `getCurrenAllocationsJig` to emphasize that it should be used only for testing purposes. We use "Jig" elsewhere for similar purposes. For testing purposes, use of this specialized method is appropriate, 
   * because tests sometimes want to check whether the expected thing has happened with internal state
   * because tests run in isolation, where we're not concerned about arbitrary races caused by any other possible activity.

When access to this internal state would violate a security property, then we would need some special arrangement so *only* testing code can use it. In this case, there is no such security violation, and so this method can be publicly exposed with the cautionary name.

Chris-Hibbert:
With #6362, there are no current uses in production. Unfortunately, there are quite a few in test, so we can't easily get rid of the method.

erights:
@Chris-Hibbert assuming this is gone now, I'm closing. But please reopen if I'm missing something.

PR #6362, opened 2022-09-30
5833 add getFinalAllocation

Chris-Hibbert:
refs: #5833

#dapp-card-store-branch: 5833-eventLoopIterationInTest

## Description

Remove a use of `seat.getCurrentAllocationJig()` in the `offerTo` helper by adding a method to ZoeSeats that returns the final allocation. If called before the seat has exited, it returns a promise for that eventual outcome. Once the seat has exited, it returns the final allocation that was included in the payout.

### Security Considerations

`getCurrentAllocationJig` is deprecated because it leaks internal state that are none of the end-user's business, but as we can see here, the `offerTo` helper needs to know. 

The switch from `getCurrentAllocations()` to `getCurrentAllocationJig()` hasn't made it to the public docs yet. The docs should include the deprecated `getCurrentAllocationJig()` as well as the new `getFinalAllocation()`

### Testing Considerations

existing tests seemed sufficient.


packages/zoe/src/contractSupport/types.js
```diff
@@ -74,8 +74,7 @@
  * from keywords used in contractA to keywords used in contractB and
  * vice versa in `offerTo`
  *
- * @param {AmountKeywordRecord | PaymentPKeywordRecord | undefined }
- * keywordRecord
+ * @param {AmountKeywordRecord | PaymentPKeywordRecord | IssuerKeywordRecord | undefined } keywordRecord
```
erights, line 77: Please revise the comment immediately above to stay accurate.

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -296,12 +295,12 @@ const reverse = (keywordRecord = {}) => {
  * @param {ERef<Invitation<Result, Args>>} invitation
  *   Invitation to contractB
  *
- * @param {KeywordKeywordRecord=} keywordMapping
+ * @param {KeywordKeywordRecord} keywordMapping
```
erights: I'm surprised that TS lets you get away with this, but I consider it incorrect. It does not describe the type that the caller needs to provide. Grepping reveals at least four callers that provide `undefined` in this argument position, which works because the parameter is indeed, at the JS level, declared optional using the JS optional parameter syntax. 

I'll make incremental suggestions that preserve what I take to be your intent without making this parameter optional at either the TS or JS levels.
```suggestion
 * @param {KeywordKeywordRecord | undefined} keywordMapping
```
accurately tells callers what they can provide in this position. With the next change, it also becomes an accurate description of what value is received at the moment the function is entered.

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -328,7 +327,6 @@ export const offerTo = async (
   invitation,
   keywordMapping = {},
```
erights:
```suggestion
  keywordMapping,
```

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -328,7 +327,6 @@ export const offerTo = async (
...
   proposal,
-  // @ts-expect-error A required parameter cannot follow an optional parameter.
   fromSeat,
   toSeat,
   offerArgs,
```
erights, line 332:
```suggestion
  offerArgs,
) => {
  if (keywordMapping === undefined) {
    keywordMapping = harden({});
  }
```

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Although I do not recommend changing it, I'll note that this many differentiated parameters is too many to be passed positionally. We should have passed a record.

erights:
Do you understand why dapp-card-store failed?

Chris-Hibbert:
> Do you understand why dapp-card-store failed?

When running locally,[ adding `await eventLoopIteration()`](https://github.com/Agoric/dapp-card-store/pull/60) fixed the problem. I'm moderately convinced that when CI runs this test, it's not pulling the appropriate branch. I've asked for help, but no  one has responded yet.

Oh,, and the reason `eventLoopIteration()` is needed is that the new code always calls out to the issuers for the amounts of the payouts, while the previous version got it directly from `zcf.getCurrentAllocation()`.  If we added `zcf.getFinalAllocation()`, we wouldn't need to wait for the issuers. I don't think `zcf.getFinalAllocation()` leaks info that it would be useful to hide, as `zcf.getCurrentAllocation()` does.

erights:
Thanks for the explanation. I like `getFinalAllocation` much better. Easy to change to that?

erights:
See also https://github.com/Agoric/agoric-sdk/pull/6370 , which fixes a problem I discovered during this review. But it is not a problem introduced by this PR.

erights (commented):
Just comments so far. Will resume review after change to `getFinalAllocation` if we're doing that.

erights:
Oops, I should have caught this when you mentioned it earlier. There is no problem with `zcf.getCurrentAllocation`, as that is only for use by the contract, for whom current allocations are appropriate. The problem is that we made the current allocation visible from outside the contract, from UserSeats and notifiers. That was the abstraction violation, and why https://github.com/Agoric/agoric-sdk/issues/5833 complains specifically about UserSeats.

erights:
Double oops. I see that this PR is getting rid of a use of `UserSeat.getCurrentAllocationJig` after all, which is great. So now I understand that your question was whether we should add a `UserSeat.getFinalAllocation`, right? Sounds plausible to me, though it'll be yet another case where we have a promise that exists for a long time in an unresolved state, and that would want to survive upgrade if it could. This PR as is doesn't exacerbate that problem. I'm continuing my review with that in mind.

Chris-Hibbert:
> though it'll be yet another case where we have a promise that exists for a long time in an unresolved state, and that would want to survive upgrade if it could.

Yes, it would be a promise on which the holder should detect vat upgrade and request a new promise.

Actually, I'd recommend calling it only after you know the payout has resolved, in which case it will always resolve to an allocation.

packages/zoe/src/contractSupport/zoeHelpers.js
Chris-Hibbert: I thought the meaning of 'optional' in javascript was that you didn't have to provide a value in the call. It's pretty clear that if the 5th positional parameter is required, then you have to provide a 4th parameter (even if it's `undefined`) in order to have a 5th parameter. 

Does this interpretation mean that it's only legal to provide `undefined` as an argument if the parameter is declared optional?

packages/zoe/src/contractSupport/types.js
Chris-Hibbert: done

packages/zoe/src/contractSupport/zoeHelpers.js
Chris-Hibbert: done.  I like this approach better. It makes more sense to me than declaring the param to be optional.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: > Does this interpretation mean that it's only legal to provide `undefined` as an argument if the parameter is declared optional?

No. But the best way to declare this for a non-optional parameter is the way you already have, with a `| undefined`. I just want to reserve syntax that suggests "optional parameter" only for truly optional parameters.

erights (approved):
Please change the PR description to align with there recent changes.

LGTM!

Issue #6364, opened 2022-09-30
proposal: merge all of trunk into release-pismo


warner:
## What is the Problem Being Solved?

Some of the changes we want to include on `release-pismo` touch many files (typescript improvements), and cherry-picking the minimal set of changes to `release-pismo` is hard work. If we can afford to just include everything currently on trunk in `release-pismo`, it would be a lot easier.

## Changes currently on trunk that are not on release-pismo

Since we forked off `release-pismo`, the `master` branch has commits for the following PRs (oldest first):

| issue/PR | status          | who        | description                                         |
| -------- | --------------- | ---------- | --------------------------------------------------- |
| #6278    |                 | dckc       | feat(agoric-cli): proposeChangeMintLimit            |
| #6265    |                 | turadg     | fix-wallet-show-num-Wants-satisfied                 |
| #6252    |                 | gibson     | 6248-followup swing-store, snapshot streams         |
| #6287    |                 | turadg     | contract-types                                      |
| #6292    |                 | turadg     | streamline-local                                    |
| #6289    |                 | sam/turadg | provision-wallet-ui                                 |
| #6309    |                 | sam        | await-keplr                                         |
| #6316    |                 | dckc       | feat(agoric-cli): support --home, --keyring-backend |
| #6328    |                 | sam        | wallet-rejected-offers                              |
| #6332    |                 | sam        | optimistic-wallet-bridge                            |
| #6325    | on pismo        | mhofman    | slog sender in subprocess                           |
| #6331    |                 | kris       | new Endo                                            |
| #6338    | on pismo        | mhofman    | run CI/mergify on release-pismo                     |
| #6333    |                 | dckc       | 6329-offer-summary                                  |
| #6317    |                 | turadg     | gov-wallet-misc                                     |
| #6348    | on pismo        | chibbert   | set decimalPlaces when calling startPSM             |
| #6352    | on pismo        | mhofman    | better-handle-misbehaving-slog-senders              |
| #6324    | on pismo        | mfig       | mfig-deterministic-bundle                           |
| #6327    |                 | markm      | test-rankOrder works on ava 3 or 4                  |
| #6354    |                 | sam        | wallet-decimal-places                               |
| #6359    | wanted on pismo | mfig/turag | 6174-wallet-current                                 |

warner:
#6327 is safe. I expect the new Endo is not risky (#6331).

warner:
@Tartuffo and @arirubinstein approve the plan

[...]

PR #6365, opened 2022-09-30
Merge branch 'release-pismo' into merge-pismo-into-trunk

warner:
Merge branch 'release-pismo' into merge-pismo-into-trunk

This is a merge from `release-pismo` into current `master`. The few
conflicts were resolved in favor of what was already on `master`. The
patch-package diff on `release-pismo` for endo/bundle-source 2.3.0 was
removed, because `master` had been upgraded to 2.3.1, and had its own
patch.

This ensures that we haven't left any changes behind (included in
release-pismo but not on trunk).

Once complete, we can fast-forward `release-pismo` to this commit, to
complete #6364 ("merge all of trunk into release-pismo"). At that
point, both branches will have this commit as their shared common
ancestor.

### DIFF
--- a/.github/workflows/after-merge.yml
+++ b/.github/workflows/after-merge.yml
@@ -3,7 +3,10 @@
 #  Require branches to be up to date before merging
 on:
   push:
-    branches: [master] # $default-branch
+    branches:
+      # $default-branch
+      - master
+      - release-pismo
 
 jobs:
   build:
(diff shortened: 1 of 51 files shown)


warner:
@gibson042 `packages/swing-store/src/snapStore.js` was one conflict I had to resolve (involving the definition of `sink`, I believe), please double-check that I didn't resolve it wrong.

kriskowal (approved):
My review and the tests appear to agree that merging `release-pismo` does not frustrate kernel consistency, so no additional work necessary to resync Endo. 👍

mhofman (approved):
Re swing-store state (and lack of changes): LGTM.

warner:
note: this PR was to make sure `master` had everything from the release branch. At about the same time, we decided that everything on trunk was also safe to put on the release branch, so after this landed, I fast-forwarded `release-pismo` to the same commit (ebbfc42). We chose to fast-forward the branch rather than make a new commit so the common ancestor would be this commit (and there would be a clear width=1 choke point on a commit-graph diagram). If instead I'd filed a PR against `release-pismo` from `master` (or rather a branch created from master without any new commits on it, to avoid auto-deleting `master`), then the diagram wouldn't have a clear choke point, and future commit-history spelunkers would have a harder time eyeballing the common ancestor commit.

PR #6369, opened 2022-09-30
no more maxNodeModuleJsDepth

turadg:
## Description

Progress on https://github.com/Agoric/agoric-sdk/issues/4560

Depends on https://github.com/endojs/endo/pull/1307

I made the Endo changes and then used `yarn link` to test them in this repo. I stopped when I ran into `bundle-source` and `ses-ava` problems, which I've added to the priority list in https://github.com/endojs/endo/issues/1254

### Security Considerations

--

--

### Testing Considerations

--


mhofman:
> I made the Endo changes and then used `yarn link` to test them in this repo.

Do we know for sure that symlinks work the same as regular `node_modules` folders for tsc? I'd feel more confident if we did `yarn pack` and maybe a temporary `package.json` `resolutions` to use the updated endo packages

mhofman (approved):

kriskowal (approved):
Squee.

turadg:
> Do we know for sure that symlinks work the same as regular node_modules folders for tsc?

Not yet. This won't pass until the changes in https://github.com/endojs/endo/pull/1307 are available in NPM. Looks like rn that will take a release crank. @kriskowal could we get `dev` or `nightly` releases for Endo to reduce the friction in this cross-repo changes?

turadg:
I think once #6493 lands I can get this to complete #4560 (by hook and a little crook)

packages/zoe/src/contractSupport/priceAuthority.js
```diff
@@ -8,6 +8,9 @@ import { AmountMath } from '@agoric/ertp';
 import { makeNotifier } from '@agoric/notifier';
 
 import '../../exported.js';
+// global types
+// eslint-disable-next-line import/no-unresolved -- d.ts file, resolvable only by TypeScript
+import '@agoric/swingset-vat/src/vats/timer/types.js';
```
mhofman: `d.ts` only file cannot be `import`ed as it'd be a runtime error!

packages/internal/src/config.js
```diff
@@ -2,8 +2,8 @@
 /** @file
  *
  * Some of this config info may make more sense in a particular package. However
- * due to the maxNodeModuleJsDepth hack and our general lax dependency graph,
- * sometimes rational placements cause type resolution errors.
+ * due to #4620 and our lax package dependency graph, sometimes rational
```
mhofman: Please use a full URL instead of issue numbers which break in case of repo migration.

packages/notifier/src/asyncIterableAdaptor.js
```diff
@@ -80,6 +80,8 @@ export const makeAsyncIterableFromNotifier = notifierP => {
                 return harden({ value, done });
               });
           }
+          // xxx hint for type checker
+          assert(myIterationResultP);
```
mhofman, line 84: Ugh that's weird. Why doesn't flow control realize that `myIterationResultP` is never `undefined` here?

mhofman (changes requested):
Blocker is the `import` of `d.ts` file. I also would like to understand why that didn't cause any test to fail.

mhofman:
> I also would like to understand why that didn't cause any test to fail.

Oh thankfully it did!

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/runner/work/agoric-sdk/agoric-sdk/agoric-sdk/node_modules/@agoric/swingset-vat/src/vats/timer/types.js' imported from /home/runner/work/agoric-sdk/agoric-sdk/agoric-sdk/packages/zoe/src/contractSupport/priceAuthority.js
```

packages/zoe/src/contractSupport/priceAuthority.js
turadg: fixed

packages/internal/src/config.js
turadg: I hope we resolve these before the code leaves the repo, but I suppose that's unlikely

packages/notifier/src/asyncIterableAdaptor.js
turadg: Yeah, it's weird. I don't now why but here's the error without it,
```
> @agoric/ui-components@0.3.3 lint:types
> tsc -p jsconfig.json

../notifier/src/asyncIterableAdaptor.js(39,3): error TS2322: Type '{ [Symbol.asyncIterator]: () => { next: () => Promise<{ value: T; done: boolean; }> | undefined; } & RemotableBrand<{}, { next: () => Promise<{ value: T; done: boolean; }> | undefined; }>; } & RemotableBrand<...>' is not assignable to type 'ConsistentAsyncIterable<T>'.
  The types returned by '[Symbol.asyncIterator]().next(...)' are incompatible between these types.
    Type 'Promise<{ value: T; done: boolean; }> | undefined' is not assignable to type 'Promise<IteratorResult<T, T>>'.
      Type 'undefined' is not assignable to type 'Promise<IteratorResult<T, T>>'.
```

mhofman: Oh I think I know. TS doesn't know that the chained calls, and in particular `.then()` doesn't execute the callbacks synchronously, so it can't know that `myIterationResultP = undefined;` hasn't happened.

mhofman (approved):
LGTM. Thanks for fixing these types! One step closer to less ambient issues.

PR #6370, opened 2022-10-01
fix: protect zoe from keyword collision

erights:
Zoe makes two uses of `fromEntries` to construct a keyword record from user-provided data, where there might be a keyword collision, causing some entries to overwrite others.

This PR protects Zoe from that by providing a `fromUniqueEntries` that throws if there is a collision and otherwise acts like the built-in one. We place that in the `@agoric/internal` package so it can be widely reused anywhere in agoric-sdk where a similar hazard arises. However, it will need to move or be copied to use outside of agoric-sdk.

Found when reviewing https://github.com/Agoric/agoric-sdk/pull/6362 , though it does not change our vulnerability to this flaw.

See also https://github.com/endojs/endo/pull/1306

### DIFF
--- a/packages/internal/src/utils.js
+++ b/packages/internal/src/utils.js
@@ -8,7 +8,33 @@ import { isPromise } from '@endo/promise-kit';
 const { getPrototypeOf, create, entries, fromEntries } = Object;
 const { ownKeys, apply } = Reflect;
 
-const { details: X } = assert;
+const { details: X, quote: q } = assert;
+
+/**
+ * Throws if multiple entries use the same property name. Otherwise acts
+ * like `Object.fromEntries`. Use it to protect from property names
+ * computed from user-provided data.
+ *
+ * @template K,V
+ * @param {Iterable<[K,V]>} allEntries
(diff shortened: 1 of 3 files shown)


packages/internal/src/utils.js
```diff
@@ -8,7 +8,33 @@ import { isPromise } from '@endo/promise-kit';
 const { getPrototypeOf, create, entries, fromEntries } = Object;
 const { ownKeys, apply } = Reflect;
 
-const { details: X } = assert;
+const { details: X, quote: q } = assert;
+
+/**
+ * Throws if multiple entries use the same property name. Otherwise acts
+ * like `Object.fromEntries`. Use it to protect from property names
+ * coputed from user-provided data.
```
Chris-Hibbert:
```suggestion
 * computed from user-provided data.
```

Chris-Hibbert (approved):
lgtm, thanks.

packages/internal/src/utils.js
erights: Done

PR #6371, opened 2022-10-01
fix: pattern based compression

erights:
When the only specimens you store are those that match a pattern, store only the information which is not literally the same as the pattern.

Added to the store module
   * `compress(specimen, pattern)` - returns bindings array, or `undefined` to indicate match failure
   * `mustCompress(specimen, pattern, label)` - indicates failure with diagnostic including `label`
   * `decompress(bindings, pattern)` - undoes the others.

Modified SwingSet's virtual mapStores so that if they have a `valueShape` option, they also use it for compression.

Modified ERTP's paymentLedger to provide the `amountShape` as the `valueShape` option for the durable ledger mapStore.

Verified in the debugger that we now only store a singleton array around the value itself, dropping the enclosing amount structure and the brand which is always the same for a given ledger mapStore. IOW, instead of storing the serialization of
```js
{
   brand: moolaBrand,
   value: 2n,
}
```
for every live payment, with this PR we store the serialization of
```js
[2n]
```
for every live payment. This is not just a savings because it is smaller. We also avoid needing to adjust the refcount of their common `moolaBrand`.

Once we've switched to smallcaps, the serialization of `[2n]` is tiny.

### DIFF
--- a/packages/SwingSet/src/liveslots/collectionManager.js
+++ b/packages/SwingSet/src/liveslots/collectionManager.js
@@ -5,6 +5,7 @@ import {
   assertKeyPattern,
   assertPattern,
   matches,
+  fit,
   compareRank,
   M,
   zeroPad,
@@ -13,6 +14,8 @@ import {
   isEncodedRemotable,
   makeCopySet,
   makeCopyMap,
+  mustCompress,
+  decompress,
 } from '@agoric/store';
 import { Far, passStyleOf } from '@endo/marshal';
 import { decodeToJustin } from '@endo/marshal/src/marshal-justin.js';
@@ -213,7 +216,7 @@ export function makeCollectionManager(
     return storeKindInfo[kindName].kindID;
   }
 
-  // Not that it's only used for this purpose, what should it be called?
+  // Now that it's only used for this purpose, what should it be called?
   // TODO Should we be using the new encodeBigInt scheme instead, anyway?
   const BIGINT_TAG_LEN = 10;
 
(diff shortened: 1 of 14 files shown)


erights:
Ready for review!

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -351,16 +370,11 @@ export function makeCollectionManager(
...
-      matches(key, keyShape) ||
-        assert.fail(X`invalid key type for collection ${q(label)}`);
+      fit(key, keyShape, keyLabel);
       !has(key) ||
         assert.fail(X`key ${key} already registered in collection ${q(label)}`);
-      if (valueShape) {
-        matches(value, valueShape) ||
-          assert.fail(X`invalid value type for collection ${q(label)}`);
-      }
       currentGenerationNumber += 1;
```
erights: @FUDCo this is the change I'm most worried about. If the pattern does not match, the old code would throw before 
```js
currentGenerationNumber += 1;
```
whereas the new code throws after. We could swap the lines, but then the old code would increment before serialization and the new code would increment afterwards. I don't know the significance of `currentGenerationNumber` so I can't judge. Please advise.

erights (commented):
Question for @FUDCo (Nevermind. Resolved)

packages/SwingSet/src/liveslots/collectionManager.js
erights: Nevermind. I'm switching the order. Serialization before increment is harmless.

erights:
Closing in favor of #6432

PR #6373, opened 2022-10-02
refactor(ertp): improve vivify factoring

dtribble:
closes: #XXXX
refs: #XXXX

## Description

Construct new issuers without storing then revivifying them.

This also changes the information abotu the issuer the baggage to being a single record rather than multiple individual fields.

### Security Considerations
This provides a cleaner factoring for sharing code between creation and revivification, primarily for singletons.

This should be an easier-to-document pattern.

### Testing Considerations

Existing tests should cover this refactor.

### DIFF
--- a/packages/ERTP/src/issuerKit.js
+++ b/packages/ERTP/src/issuerKit.js
@@ -15,6 +15,16 @@ import './types.js';
 
 /**
  * @template {AssetKind} K
+ * @typedef {object} IssuerRecord
+ * @property {string} name
+ * @property {K} assetKind
+ * @property {AdditionalDisplayInfo} displayInfo
+ * @property {Pattern} elementShape
+ */
+
+/**
+ * @template {AssetKind} K
+ * @param {IssuerRecord<K>} issuerRecord
  * @param {Baggage} issuerBaggage
  * @param {ShutdownWithFailure=} optShutdownWithFailure If this issuer fails
  * in the middle of an atomic action (which btw should never happen), it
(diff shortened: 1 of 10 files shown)


packages/ERTP/src/issuerKit.js
```diff
@@ -15,6 +15,16 @@ import './types.js';
...
+ * @property {K} assetKind
+ * @property {AdditionalDisplayInfo} displayInfo
+ * @property {Pattern} elementShape
+ */
+
+/**
+ * @template {AssetKind} K
+ * @param {IssuerRecord<K>} issuerRecord
  * @param {Baggage} issuerBaggage
  * @param {ShutdownWithFailure=} optShutdownWithFailure If this issuer fails
```
erights:
```suggestion
 * @param {ShutdownWithFailure | undefined} optShutdownWithFailure If this issuer fails
```
If this parameter here is not longer optional, in the sense that it may be omitted, but still may be `undefined`, then I prefer that you type it as ` | undefined`.

erights (approved):
Much better, thanks!

LGTM

packages/ERTP/src/issuerKit.js
```diff
@@ -25,14 +35,11 @@ import './types.js';
  * See https://github.com/Agoric/agoric-sdk/issues/3434
  * @returns {IssuerKit<K>}
  */
-export const vivifyIssuerKit = (
+const setupIssuerKit = (
```
turadg, line 38: "setup" sounds to me like something that is the beginning of work, but this is finishing. This is really "make an IssuerKit from baggage". Clunky name but I expect we'll perform a thorough renaming after https://github.com/endojs/endo/issues/1193

So please consider the explicit `makeIssuerKitFromBaggage` until we have a good shorthand for "make from baggage".

erights:
Still needed. How can we advance this?

turadg:
@dtribble i could work with you or @erights to get this to the finish line

erights:
What is the relationship between this PR and #6800 ?

dtribble:
Folded into a different PR.

PR #6377, opened 2022-10-03
Slog sender improvements

mhofman:
## Description

- Fixes the `makeSlogSender` aggregator to prevent using the same sender module multiple times.
- Fixes cosmic-swingset to correctly handle when no slogSender exists.
- Add the ability to override/set env variables specific to slog senders. This for example would allow disabling tracers for the sub-process.


FUDCo (approved):
This all seems reasonable.

Issue #6372, opened 2022-10-02
PSM trades can take long amounts of time to "finish"


arirubinstein:
When testing against ebbfc4, when tasking 20 or 100 workers with making wallet action psm trades, some of the works take an extremely long time to acknowledge the completion of the trade, via `agops perf satisfaction`.  e.g. there are no logs from the contract or vbank during this period of waiting.  This can be demonstrated by the odd max times here:

<img width="742" alt="Screen Shot 2022-10-01 at 8 59 26 PM" src="https://user-images.githubusercontent.com/22369/193437248-f6834852-b716-4ce1-ab46-1283e4a46bf9.png">

For the following samples - swingset seems to be doing no vat work, but agops perf satsifaction workers are waiting.  near the end, it appears 1 worker gets satisfied per block
Blocks:
- 1483 - 1495 
- 1543 - 1571

What's happening in these block if anything? Is there notification work waiting to be saved to the chain that isn't represented in our current telemetry?
If absolutely nothing is happening in these blocks, could the client not be querying the offer status correctly?

Slogs: https://drive.google.com/drive/folders/1LBHoDsrbESyc0WG-JPg7mp9d9LvuFoUJ?usp=sharing

This is easily reproducible with a loadtest of 100 workers performing a trade

turadg:
Current hypothesis: the lag in runs 13-15 is due to validators getting out of sync and the round-robin getting unlucky. Due to the fast retries of the perf command (100ms) it can reach exponential backoff max of 30s and then in the last outstanding offers not make any requests while the data may be available.

Testing now with 100 workers on `loadtest`. Results will be in `run-ta-1539.txt`

warner:
I sampled the slogfile from blocks 1400 to 1500.

* 1400-1410: every block pulls 4 or 5 bridge items from `inboundQueue` and completes them
  * this shows the effects of raising the bean/computron limit, previously we would only complete one or two per block
  * every block is 100% swingset work (all but the last finish with negative `.remainingBeans`)
* 1411-1459 : no swingset work
* 1460-1479 : 100% swingset work (block 1479 was the first that had any `remainingBeans`)
  * this performed 92 trades in 428 seconds, at 4.65 s/trade (= 0.215 trades/s)
* blocks 1480-1500: no swingset work

I looked at the next chunk too:
* blocks 1520-1542
  * 101 trades in 510 seconds, 5.05 s/trade (= 0.198 trades/s)
[...]

PR #6379, opened 2022-10-03
6372 client misc

turadg:
## Description

Miscellaneous improvements from investigating https://github.com/Agoric/agoric-sdk/issues/6372

- chore: leader log to console
- fix(perf command): wait for numWantsSatisfied
- chore: drop default max backoff to 6s from 30s

### Security Considerations

--

--

### Testing Considerations

Manually testing with loadtests

### DIFF
--- a/packages/agoric-cli/src/commands/perf.js
+++ b/packages/agoric-cli/src/commands/perf.js
@@ -53,15 +53,13 @@ export const makePerfCommand = async logger => {
 
       const spec = `:published.wallet.${opts.from}`;
 
-      const leaderOptions = makeLeaderOptions({
-        sleep: SLEEP_SECONDS,
-        jitter: 0,
-        log: () => undefined,
-      });
-
       const leader = makeLeaderFromRpcAddresses(
         networkConfig.rpcAddrs,
-        leaderOptions,
+        makeLeaderOptions({
+          sleep: SLEEP_SECONDS,
+          jitter: 0,
+          log: console.warn,
+        }),
       );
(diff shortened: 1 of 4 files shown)


packages/agoric-cli/src/commands/perf.js
```diff
@@ -76,9 +74,10 @@ export const makePerfCommand = async logger => {
             if (status.error) {
               console.error(status.error);
               exit(1);
-            } else if (status.numWantsSatisfied)
+            } else if (status.numWantsSatisfied) {
               process.stdout.write(`satisfied: ${status.numWantsSatisfied}\n`);
-            exit(0);
```
turadg, line 81: this would exit whenever a non-error update arrived. in practice, this happens about the same time as numWantsSatisfied.

packages/casting/src/defaults.js
```diff
@@ -14,6 +14,11 @@ export const DEFAULT_POLL_WITH_EVENTS_SECONDS = 600;
 
 export const DEFAULT_KEEP_POLLING_SECONDS = 5;
 
+/**
+ * Default backoff is the block interval.
```
turadg, line 18: @michaelfig does this make sense?

mhofman: Usually you want 50% or less of the interval so that if you happen to start right before a new block is cut, you don't have to wait the whole block time to pick it up.

warner:

turadg:
obsolete

PR #6380, opened 2022-10-04
fix(wallet-ui): parse purse.brand in fetchCurrent

samsiegart:


turadg (approved):
What does this fix?

I think soon after launch we should work on test coverage.

Issue #6314, opened 2022-09-23
ante-handler should reject x/swingset messages if the inboundQueue is too full


warner:
## What is the Problem Being Solved?

The second half of our good-enough fix for #6306, after #6313 makes the `inboundQueue` length available to go, is to change the ante-handler to reject x/swingset messages if that queue is longer than some configurable threshold.

That queue is extended/drained during END_BLOCK processing, so really:

* the stateless ante-handler invocation that gates admission into the mempool should know the queue length as of the end of the previous block
  * ideally, this handler would only admit `max - current` swingset-bound messages into the mempool
  * but I suspect it's so stateless that it can't maintain a "number added since last block" counter
  * oh well
* the in-consensus ante-handler which gets called as blocks are executed should know both the queue length as of the end of the previous block, plus it should maintain a counter
  * so if the threshold is 1000, and the `inboundQueue` had 999 entries at the end of the last block, and the new block has 500 new entries, the first should be accepted and the remainder should be rejected
  * if it didn't maintain a counter, then all 500 would be accepted, driving `inboundQueue` up to 1499 entries

The cutoff threshold must be in-consensus, and should be configurable somehow (I think this is a "chain parameter"?) via governance.

warner:
I'm thinking the initial value of this should represent about an hour's worth of PSM trades, and the last number I saw was about 1.4 trades per block, at about 2s of compute time (plus 5s of consensus delay), so 7/1.4 = 5s per trade, so 3600/5 = a limit of about 720. Maybe call it 1000 entries to make it a round number.

mhofman:
Closed by #6381

Issue #6313, opened 2022-09-23
expose inboundQueue.length to x/swingset go code


warner:
## What is the Problem Being Solved?

For #6306, the go-side "ante handler" needs to know the "current" value of `inboundQueue.length`. This queue is extended/drained during END_BLOCK processing, so the value is only updated at the end of swingset processing (`ActionType.END_BLOCK` in `blockingSend()` inside `launch-chain.js`).

The task is to send this value from the JS side into the golang side at two points:
* during BEGIN_BLOCK, so the stateful x/swingset ante-handler can reference it as it processes each of the block's txns (so it can reject some of them)
* during END_BLOCK, so the stateless ante-handler can reference it as it considers accepting txns into the mempool

## Description of the Design

Not sure, @mhofman has a plan.

## Security Considerations

none

## Test Plan

probably manual, since we don't have much test coverage in cosmic-swingset

mhofman:
Closed by #6381

PR #6381, opened 2022-10-04
feat: limit inbound queue

### LINKED ISSUE #6306: how to prevent unbounded inboundQueue growth and excessive trade latency

@arirubinstein did some testing and was able to fill the new `inboundQueue` with 15k PSM trades pretty quickly, with a moderate number of workers and fees. That amount of work could take hours or days to clear. Someone who submits a request while the queue is that full would see their cosmos txn get accepted into a block fairly quickly, but the trade wouldn't actually execute for a long time.

This is a basic queueing-theory problem: when things are slow, it should be first-come-first-served, but there will always be times when there is enough demand to cause delays. Our first question is to figure out what the desired experience is. When Alice submits a trade to a queue with N seconds/minutes/hours of work ahead of her, will her request be rejected immediately, be rejected later, be processed eventually (but maybe much later than she wants)? How much control does she get to chose between those outcomes? How much visibility does she get into her request's current status?

We're limited by existing implementation layers:
* a stateless/non-consensus "ante handler" is checked before accepting a message into the mempool
* the Tendermint layer uses arrival ordering to choose which txns get from the mempool into a proposed block (filtered, but not sorted, by max gas value, against a per-validator threshold)
* once accepted into a block, gas fees are always charged, but a given message might be executed (i.e. added to `inboundQueue`) or rejected (which ought to be detectable by clients, just as if they'd ran out of funds to pay for gas, or had some syntax error in their transaction)
  * a stateful/within-consensus form of the ante handler runs here and makes this decision
* our stateful ante handler has the ability to charge an "admission fee" when the messge is executed, based on whatever criteria we like

There are four basic categories of approaches we're thinking of taking, each with different user experiences under load.

* 1: allow the unbounded `inboundQueue`
* 2: cap `inboundQueue` at some fixed depth, perhaps one that corresponds to one hour of work. New txns continue to be accepted into blocks, but when the block is executed, they'd be rejected rather than being added to `inboundQueue`
* 3: charge an admission fee upon execution which increases as `inboundQueue.length` grows, maybe exponentially
* 4: compute that admission fee, but check the txn for a `maxAdmissionFee`, and either reject the txn or (charge fee and add to `inboundQueue`

### Unbounded Queue

Our existing unbounded `inboundQueue` has two downsides:
* the chain must store unbounded state (however we can safely assume this storage cost is basically covered by our admission fee)
* a given request's latency is unbounded: it will be proportional to the number of requests on the `inboundQueue` when their message was accepted into a block and then executed

The user experience is something like:
* look at the chain, see how long the queue currently is, decide to submit or not
* submitted txns hang out in the mempool for some amount of time before being accepted into a block
* once accepted, the message is added to the end of `inboundQueue`, which might be longer or shorter than it was when the user decided to submit it
* message then slowly works through `inboundQueue` until being processed

The user's message would always get processed eventually, but it could be days if the queue were really full, and the actual time can't be known until it gets past the mempool.

### Size-Limited Queue With Reject (second ante-handler only)

Option 2 causes block transaction messages to be dropped if `inboundQueue` is too full. The user experience is:
* look at the chain, try to guess your chances of success, decide to submit of not
* submitted txn hands out in mempool for some time, then gets accepted into block
* once accepted into a block, the block is executed:
  * some percentange of the time, the txn is rejected: gas fees are charged, but no trade happens
  * some other percentage of the time, the txn is accepted into `inboundQueue`, and is processed within an hour

### Size-Limited Queue With Reject (both ante-handlers)

@michaelfig pointed out that it isn't hard to make the first (non-consensus) ante-handler do this check too. This changes the experience to:
* submit a txn
* if the validator/node you submit it to currently has a full `inboundQueue`, the txn is rejected immediately, without being added to the mempool, and you get a fast error, and no gas fees are paid
* if it thinks it has space, the txn hangs out in the mempool for a while
* eventually it gets into a block, then gets executed (which costs you gas fees regardless of the outcome)
* a second check against `inboundQueue.length` is made
  * sometimes the txn is rejected
  * sometimes the txn is added to `inboundQueue`, then gets processed within an hour

Because it's stateless, it probably always be comparing against the `inboundQueue.length` from the end of the last block (and not keeping track of how many txns have been accepted into the mempool since then), so if multiple txns arrive between blocks, they might all be accepted into the mempool, even though the chain only has room for one of them. The first one will get into `inboundQueue`, the rest will be gas-fee-charged and rejected by the second ante handler. I'd have to do some simulations to estimate what percentage of submitted txns hit the three categories: rejected by the first handler, rejected by the second handler, or accepted and eventually processed.

### Escalating Admission Fee, No User Limit

We currently charge an admission fee in the second ante handler (independent of the gas fees), in IST, whose value depends upon the type and size of the message. We could change that to impose a congestion fee: the fee grows (probably exponentially) with the current depth of `inboundQueue`. So the fees are trivially low when the queue is empty, and prohibitively high when the queue is full. This lets us avoid having a fixed size and probabilistic accept/reject behavior when it is reached.

However the user experience then looks like:
* look at the chain, estimate the admission fee based on the current `inboundQueue.length`, decide to submit or not
* submit a txn
* txn hangs out the mempool, then gets included in block and executed
* user gets charged variable fee depending upon `inboundQueue` at the time of execution
  * if user has enough funds, message is added to `inboundQueue` and executed within an hour
  * if not, message is rejected and no admission fee is charged (although gas still applies)

That would be surprising. 

### Escalating Admission Fee, User-Defined Limit

If we add a new property to the transaction class, named `maxAdmissionFee` or something, then we can give the user a way to express their willingness to pay the fee. This gives them more control and predictablity of what they pay.

This requires some work:
* the message class (in x/swingset) needs a new property, and is serialized differently than before
* the class requires new accessors to reveal this property to the ante handler
* ante handler must compare the computed `f(inboundQueue.length)` fee against the new property, reject if exceeded
* CLI tools must add a new `--max-admission-fee=` argument when creating/sending the txn

The user experience looks like:
* look at chain, estimate admission fee, decide if that is worth it or not
* choose `max-admission-fee`, submit txn
* if the node you submit it to computes a required fee (based on current `inboundQueue`) that exceeds your limit, txn is dropped immediately, no fees paid
* else txn hangs out in mempool for a while, gets added to block, executed
* real fee is calculated from current `inboundQueue` depth, compared against limit
  * if exceeded, txn is rejected, you pay gas fees but not admission fee
  * if accepted, you pay admission fee (perhaps below your limit), txn is processed within an hour

Note that, unlike gas fees (which are paid exactly for all txns that get included in a block), you do not always pay the full admission fee amount you were willing to pay. But sometimes you might.
  
This isn't quite a priority auction: you can increase the chances that your txn will be processed by stating a higher limit, but that doesn't cause your txn to be processed any earlier. To be a priority auction, we'd need some pool of work that is sorted by bid or max-fee or something, rather than being serviced in FIFO order, and we don't have any easy candidates:

* the Tendermint mempool-to-proposed-block selector doesn't currently do this (and it is not easy for us to customize), and only knows about gas, not cosmos-sdk -level transaction properties like a `max-admission-fee`
* Cosmos executes the txns within a block in strict inclusion order
* `inboundQueue` has actions which represent multiple things: user-submitted PSM trades, IBC events, vbank balance updates, ag-solo mailbox messages. Many of these require a fixed FIFO execution order.
  * to enable priority selection, we'd probably need to isolate the user-submitted PSM trades into their own priority queue, kept sorted by some bid, and arrange for the bid to be charged when the trade was processed
  * currently swingset doesn't know about fees (all fees are paid by the time x/swingset puts the message into the action queue, well before they get added to `inboundQueue`)
  
## Current Plan

We'll implement at least 2, and maybe 4.

@mhofman is going to expose `inboundQueue.length` to Golang. @michaelfig or @JimLarson  will implement an ante handler that compares it against a configured length and rejects txns if exceeded.

@michaelfig said he or @JimLarson could look into adding the `max-admission-fee` property (and API/UI), and then change the ante handler to compute/compare/charge an admission fee (using more configured parameters: if we use exponential, we need an `A*e^(B*length)`).

mhofman:
closes: #6306 #6313 #6314

closes: #6381 (adopts stacked commit)

## Description

This updates the go->swingset BlockingSend path to include a JSON parsed result, and uses that new path to communicate during begin block a number of messages go should add to the action queue. This allowed value is currently computed from a new inbound queue max block param. It is enforced by a new ante handler by rejecting transactions that would put the action queue over the number of new allowed messages.

A separate mempool value is computed by using half the max inbound size. This provides some buffer to avoid including transactions in a block that would then get rejected.

To deal with hangover inconsistency, the allowed queue sizes are stored in the host section of the swingstore, and replayed at begin block if the computed height logic detects an inconsistency (checks factored out of end block).

### Security Considerations

Cosmos may end up queueing more than the allowed messages since only external transactions going through the ante handler are subject to the check, and internally generated messages such as vbank updates are processed after. This is fine since these are usually low overhead, and will simply be accounted for at the next block anyway.

I picked a sensible default of 1000 for the max inbound queue size, but like other params, this is configurable by governance.

### Testing Considerations

Tested locally by setting the max inbound queue size to 5, and a loadgen with 32 tasks in parallel. The solo client did cause the allowed inbound to drop to 0, and triggered "mempool full" errors when trying to send a message to the chain.


mhofman:
As usual, best reviewed commit-by-commit.

@michaelfig for primary review since both @JimLarson and I worked on this PR
@warner for the architecture of the whole thing

packages/cosmic-swingset/src/chain-main.js
```diff
@@ -177,7 +177,7 @@ export default async function main(progname, args, { env, homedir, agcc }) {
     p.then(
       res => {
         // console.error(`Replying in Node to ${str} with`, res);
-        replier.resolve(`${res}`);
+        replier.resolve(stringify(res !== undefined ? res : null));
```
warner, line 180: I'm guessing we don't need to worry about `res` holding BigInts or other non-JSON-able data?

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -559,6 +578,23 @@ export async function launch({
     return p;
   }
 
+  function checkExpectedBlockHeight(blockHeight) {
```
warner: If you currently understand this pretty well, could you add a comment explaining what `blockHeight` and `computedHeight` represent? I know the theory (and I know that I used to understand this part), but I can't currently work out whether the "normal" case (processing a new block, not replaying a previous one) means `blockHeight === computedHeight` or `blockHeight === computedHeight + 1`. I can see that this is mostly just refactoring the previous code, but I can't eyeball its correctness in my current state of confusion/ignorance.

I think `computedHeight` is the one that the kernel remembers having executed? And `blockHeight` is coming from Go. So the normal case is for `blockHeight = computedHeight + 1`, yeah?

warner: Ah, ok, so I think we enter into this `if` block in the "normal" case, in which case the comment is confusing (I thought it was talking about the "trivial case" of an exceptional case).

And I think the `computedHeight > 0` clause is a bit weird: there's a separate `ActionType.BOOTSTRAP_BLOCK`, which gets `blockHeight: 0`, and which normally observes `computedHeight: 0` (because that's what we initialize it to), and that's ok: we're supposed to execute the bootstrap block actions even when they're the same. I suspect we wouldn't correctly handle a node which crashed just after the kernel committed the bootstrap block: we'd wake up, cosmos would re-send the `BOOTSTRAP_BLOCK` command, launch-chain.js would see `computedHeight: 0` and re-execute the actions, and chaos would ensue. It's unlikely to happen, though. I'd defer to @michaelfig , but if we felt we needed to handle this case correctly, maybe we should initialize `computedHeight` to `-1`, have `ActionType.BOOTSTRAP_BLOCK` react to `computedHeight === -1` by executing the actions, `=== 0` by ignoring the actions (and somehow replaying the outputs? does that clause even have a way to do that?), and `> 0` by throwing.

Anyways, so if `END_BLOCK` is never called for the bootstrap block, then it could `assert(blockHeight > 0)` just to make that clear. Then if `blockHeight === computedHeight + 1` then we're in the "normal case" where we're executing a block for the first time, if `blockHeight === computedHeight` then we're in the replaying case (where the kernel committed but the app crashed before cosmos got to commit and now we're starting back up again), and anything else is the decoherence case which we believe should never happen (like as if cosmos allowed reorgs, which it doesn't, it has finalization instead).

And `checkExpectedBlockHeight()` returns `true` if `blockHeight === computedHeight + 1` (meaning "normal case, execute this block"), `false` if `blockHeight === computedHeight` (meaning "replay case, skip execution, return the old results), and throws otherwise.

I'm not sure what the `if (computedHeight > 0` expression is doing. Could we remove it? Let's see, when BOOTSTRAP_BLOCK commits, it will save `computedHeight = 0` (from a slogfile I see the bootstrap block causes a `cosmic-swingset-run-finish` with `blockHeight: 0`), so the first END_BLOCK of the chain (`blockHeight: 1`) will skip this check entirely. If we removed it, the first END_BLOCK should observe `blockHeight: 1` and `computedHeight: 0` (so `blockHeight === computedHeight + 1`, so "normal case, execute this block"). If the app crashes after the kernel commits the first END_BLOCK, then the restart will observe `computedHeight: 1` and `blockHeight: 1` (so "replay case").

Anyways, if that logic seems sound (and it seems worth the effort, which it may not be), we could change this function to be easier to understand and have confidence in. At the very least we need to add a note to the return value to explain what true/false mean (in terms of "execute for real" vs "replay").

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -300,6 +301,23 @@ export async function launch({
...
   }
 
+  let savedQueueAllowed = JSON.parse(
+    kvStore.get(getHostKey('queueAllowed')) || '{}',
+  );
+
+  function computeQueueAllowed(_blockHeight, _blockTime, params) {
+    assert(params.queueMax);
+
+    assert(QueueInbound in params.queueMax);
```
warner, line 310: Do we use `in`? That's the one that looks at prototypes too, right? Is there a straightforward way to express the check without implying/being-vulnerable-to inheritance?

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -300,6 +301,23 @@ export async function launch({
...
+    const inboundQueueMax = params.queueMax[QueueInbound];
+
+    const inboundQueueSize = inboundQueue.size();
+
+    const inboundQueueAllowed = Math.max(0, inboundQueueMax - inboundQueueSize);
```
warner, line 317: Wanna use `remaining` or `remaining space` in this name? `inboundQueueAllowed` reads like it's a boolean, "the inbound queue is allowed to exist" or "the inbound queue allows new items". Maybe just `remainingSpace`? Or something with `available` in it? I probably have the same concern with of the other names with an `Allowed` suffix, although it's pretty pervasive so I don't want to suggest an invasive renaming.

packages/cosmic-swingset/src/launch-chain.js
warner: Maybe the function should be named `blockNeedsExecution` or `notReplay` or something.

packages/cosmic-swingset/src/launch-chain.js
warner: Maybe this function should be named `updateRemainingSpace`, because `compute` makes me wonder why it's not returning the value it computed. The updater function name should match the name of the variable being updated, of course.

golang/cosmos/ante/inbound.go
```diff
@@ -0,0 +1,73 @@
...
+	for _, msg := range msgs {
+		switch msg.(type) {
+		case *swingtypes.MsgDeliverInbound,
+			*swingtypes.MsgInstallBundle,
+			*swingtypes.MsgProvision,
+			*swingtypes.MsgWalletAction,
+			*swingtypes.MsgWalletSpendAction:
+			if inboundsAllowed == -1 {
+				state := ia.sk.GetState(ctx)
+				allowed, found := swingtypes.QueueSizeEntry(state.QueueAllowed, "inbound")
```
warner: Does `allowed` mean the maximum (1000), or the remaining space? The `allowed - actions` below suggests that it means the maximum, but I'm not sure that's how it was used over in the JS side.

golang/cosmos/ante/inbound.go
```diff
@@ -0,0 +1,73 @@
...
+					} else {
+						inboundsAllowed = 0
+					}
+				} else {
+					// no inbound allowed size from swingset
+					inboundsAllowed = 0
+				}
+			}
+			if inboundsAllowed > 0 {
+				inboundsAllowed--
```
warner, line 76: This looks like it's local to this function/method, and it looks like it's called with a single txn (which might contain multiple messages). Do ante handlers get run for all txns before any of them are processed? Or does it do `ante handler, process txn` as a pair, before moving on to the next one? If it's the former, then a local (ephemeral) `inboundsAllowed` would reset for each txn, and allow in too many.

golang/cosmos/ante/inbound.go
```diff
@@ -0,0 +1,73 @@
...
+	msgs := tx.GetMsgs()
+	inboundsAllowed := int32(-1)
+	for _, msg := range msgs {
+		switch msg.(type) {
+		case *swingtypes.MsgDeliverInbound,
+			*swingtypes.MsgInstallBundle,
+			*swingtypes.MsgProvision,
+			*swingtypes.MsgWalletAction,
+			*swingtypes.MsgWalletSpendAction:
+			if inboundsAllowed == -1 {
```
warner, line 53: ah, ok so this queries the state at most once per txn (not once per msg), and even then only if there were any swingset messages in the txn, nice

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -117,6 +120,22 @@ func (k Keeper) actionQueueIndex(ctx sdk.Context, name string) (uint64, error) {
...
+	head, err := k.actionQueueIndex(ctx, "head")
+	if err != nil {
+		return 0, err
+	}
+	tail, err := k.actionQueueIndex(ctx, "tail")
+	if err != nil {
+		return 0, err
+	}
+	size := tail - head
+	if size > math.MaxInt32 {
```
warner, line 133: yeah let's hope our `inboundQueue` never gets this big :)

warner (approved):
Looks pretty good to me, although the names sometimes felt confusing (but if they're consistent, across both JS and Go, that's the most important thing, I'll read it again in the morning and maybe it'll speak to me differently). The only potential logic bug I'm worried about is that `allowed` seems to mean "remaining space" (`inboundQueueMax - inboundQueueSize`) in launch-chain.js, and .. oh, ok the Go side is looking at the *action* queue, not the *inbound* queue, so `allowed` is in fact the remaining space (in the inbound queue), and the golang side is doing `actions = ia.sk.ActionQueueLength` to measure the current action queue, so `allowed - actions` is the remaining space in the *action* queue, and hence what matters when `ante/inbound.go` is deciding whether to accept or reject each msg.

Ok, I'm not sure why I got confused by that, but it seems right.

packages/cosmic-swingset/src/chain-main.js
mhofman: I suppose right now `BigInt`s will just get dropped. We can update when we find a need to pass them through.

packages/cosmic-swingset/src/launch-chain.js
mhofman: > I can see that this is mostly just refactoring the previous code, but I can't eyeball its correctness in my current state of confusion/ignorance.

Yeah it might be a good opportunity to rewrite this code, I also get confused by what it actually does. I wanted to refactor it without doing changes, but happy to layer these updates on top.

packages/cosmic-swingset/src/launch-chain.js
mhofman: `params.queueMax[QueueInbound]` would similarly look up the prototype chain. `parseParams` will ensure that if there is an own property, it'll be a `number`, and that the prototype is `Object.prototype`. I could make the proto `null` but I don't think we need to protect against this kind of confusion.

packages/cosmic-swingset/src/launch-chain.js
mhofman: The naming is from a cosmos pov, where only n messages are allowed into the action queue. I'm happy to bikeshed the name, but I'd like names to be consistent throughout the stack, and I don't think "remaining" has the right connotation for an instruction JS gives to go.

That said, I agree `compute` is no longer the right prefix.

golang/cosmos/ante/inbound.go
mhofman: `actions` only contains new messages for this block. Maybe a comment to that effect would clear that up.

golang/cosmos/ante/inbound.go
mhofman: I believe it's the latter. @JimLarson?

golang/cosmos/x/swingset/keeper/keeper.go
mhofman: Yeah I'm a little paranoid with overflows, but at least it makes it clear.

And this is the actionQueue, which is currently drained at every block, not even the inboundQueue, so really it should never be anywhere close to that amount.

mhofman (commented):
I'll add some comments around the confusing bits and rewrite the computed height check.
Happy to bikeshed the "allowed" naming, but I'd like alternatives to "remaining" which is too local to the JS side.

packages/cosmic-swingset/src/launch-chain.js
mhofman: So looking into this more I think I understand the current check better, and I'm having a hard time finding a better expression of it. Here are 2 constraints missing above:
- The bootstrap block (assumed blockHeight=0) does not get a commit. The first commit is after the first block.
- The first block may not be block 1, but instead be an arbitrary number.

That means getting a `blockHeight != computedHeight - 1` is actually valid if `computedHeight === 0`, hence the first part of the check.

mhofman:
Sorry for the churn, I messed up the rebase to target master. Here is the list of commits addressing feedback since the initial push last night: https://github.com/Agoric/agoric-sdk/pull/6381/files/ea3ee50b0efa1884284d8e4647ce8b321620ac4d..4e1221767a2e01b091a8a3d198be975505ee37ca

packages/cosmic-swingset/src/launch-chain.js
warner: Oh, now I see that `BOOTSTRAP_BLOCK` doesn't receive a `blockHeight`, it just assumes it's 0.

If the chain starts at block 5, there still has to be a bootstrap block, it's just that it's numbered 5 instead of 0, right? But in that case we don't find out what the number was until the BEGIN_BLOCK for the second block, which is why we need to bypass the check if `computedHeight === 0` (meaning we accept anything for the first BEGIN_BLOCK).

Ok then if/when we get time to annotate this function, we should mention those extra constraints, and explain why the bypass-if-0 is correct.

I'm ok with this landing without the docs updates, but a few lines while this is still fresh in our heads might be nice.

michaelfig: > If the chain starts at block 5, there still has to be a bootstrap block, it's just that it's numbered 5 instead of 0

@mhofman had this right, the "bootstrap block" has no correspondence to actual Cosmos state, and is always numbered 0, since that is a blockHeight that also is never used in the Cosmos SDK (voting on the first block begins after the bootstrap block is completed by SwingSet).  The bootstrap block is never committed, either: Cosmos SDK only starts committing on the next block after bootstrap (5, in your example).

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -586,21 +586,26 @@ export async function launch({
...
-      if (restoreHeight !== computedHeight) {
-        // Keep throwing forever.
-        decohered = Error(
-          // TODO unimplemented
-          `Unimplemented reset state from ${computedHeight} to ${restoreHeight}`,
-        );
-        throw decohered;
-      }
+  function blockNeedsExecution(blockHeight) {
+    if (blockHeight > 0 && savedHeight === blockHeight) {
```
warner: wait, shouldn't that be `if (savedHeight > 0` ?

Would it be correct for this to read:

```js
if (savedHeight === 0) {
  // 0 is the default we use when the DB is empty, so either we haven't executed any blocks, or we've executed the bootstrap block but no others. The first non-bootstrap block can have an arbitrary height (the chain may not start at 1), so we can't tell if it is a replay or not, so assume the block needs to be executed
  return true;
}

if (blockHeight === savedHeight + 1) {
  // execute the next block
  return true;
}

if (blockHeight === savedHeight) {
  // we have already committed this block, so "replay" by not executing (but returning all the results from the last time)
  return false;
}

// we're being asked to rewind by more than one block, or execute something more than one block in the future, neither of which we can accomodate
decohered = ...
throw decohered;
```

warner:

warner (approved):
New changes look good, but consider the rewrite; I haven't confirmed that the existing version (`if (blockHeight > 0 && savedHeight === blockHeight) {`) is correct.

packages/cosmic-swingset/src/launch-chain.js
mhofman: > wait, shouldn't that be `if (savedHeight > 0` ?

Shoot, right!

> Would it be correct for this to read:

Yes, but I've adapted the comments a bit in https://github.com/Agoric/agoric-sdk/pull/6381/commits/06b920f6e8006fde7535aa91824e054d755350da. PTAL

golang/cosmos/ante/inbound.go
JimLarson: I'm taking a look, but in the worst case, we let too many txns in, and then in the *next* block we reject them all, giving us a little hysteresis, which isn't necessarily a bad thing.

golang/cosmos/ante/inbound.go
JimLarson: It even breaks off from iterating through all the Msgs once it discovers that it goes over the limit. The could could be a little more clear if I just summed up all the swingset Msgs in the Tx, rather than computing mid-loop, but it's not too awful as it is.

mhofman:
I will rebase and fixup once master settles to avoid re-running CI multiple times.

I'd still like to get an approval from @michaelfig, and a +1 from @JimLarson to make sure the comments I added to the ante handler are sane.

golang/cosmos/ante/inbound.go
```diff
@@ -0,0 +1,102 @@
...
+			// messages allowed for mempool admission as if its max queue length
+			// was lower (e.g. 50%).
+			if inboundsAllowed == -1 {
+				state := ia.sk.GetState(ctx)
+				entry := QueueInbound
+				if simulate {
+					entry = QueueInboundMempool
+				}
+				allowed, found := swingtypes.QueueSizeEntry(state.QueueAllowed, entry)
+				if !found && simulate {
```
JimLarson: It looks like we misunderstood the `simulate` parameter - it does not mean it's running for CheckTx - i.e. it's not the signal that we're running for mempool admission. It's not clear what the signal is for mempool context. I'm continuing to dig.

golang/cosmos/ante/inbound.go
JimLarson: Confirmed that the ante handler runs at `DeliverTx` time, so we're seeing the state mutated by previous Txs in the block. So we're going to reject the Tx if it has enough swingset Msgs to overflow the inbound queue (counting the pending action queue).

golang/cosmos/ante/inbound.go
```diff
@@ -0,0 +1,102 @@
...
+			*swingtypes.MsgWalletAction,
+			*swingtypes.MsgWalletSpendAction:
+			// Lazily compute the number of allowed messages when the transaction
+			// includes a SwingSet message. This number is the difference between
+			// the number of allowed messages initially returned by SwingSet, and
+			// the number of messages already added into the actionQueue by other
+			// transactions included in the block.
+			// We store the number of allowed messages locally since messages are
+			// added to the actionQueue after going through the ante handler (in
+			// non-simulate mode) and before the next transaction is processed.
```
JimLarson: "Simulate" isn't what we think it is. Changing to "CheckTx", which is a method on `ctx`.

golang/cosmos/ante/inbound.go
```diff
@@ -0,0 +1,102 @@
...
+				state := ia.sk.GetState(ctx)
+				entry := QueueInbound
+				if simulate {
+					entry = QueueInboundMempool
+				}
+				allowed, found := swingtypes.QueueSizeEntry(state.QueueAllowed, entry)
+				if !found && simulate {
+					entry = QueueInbound
+					allowed, found = swingtypes.QueueSizeEntry(state.QueueAllowed, entry)
+					allowed /= 2
```
JimLarson: Even with `simulate` changed to `ctx.IsCheckTx()`, this really doesn't do what you hope. For instance, if the inbound queue max is 1000, current inbound queue length is 950, then `queue_allowed[QueueInbound]` is 50, and half of this is 25. So this still lets an unlimited number of transactions into the mempool as long as they're sending less than 25 swingset messages each. What we really want is to divide the limit in half to 500, see that `queue_allowed[QueueInboundMempool]` is zero, and disallow mempool admission. Will remove.

golang/cosmos/ante/ante.go
```diff
@@ -52,6 +56,7 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 		ante.NewSigGasConsumeDecorator(opts.AccountKeeper, sigGasConsumer),
 		ante.NewSigVerificationDecorator(opts.AccountKeeper, opts.SignModeHandler),
 		NewAdmissionDecorator(opts.AdmissionData),
+		NewInboundDecorator(opts.SwingsetKeeper),
```
JimLarson: Moving this up before our fee collection decorator, so hopefully we can avoid being charged fees even for an inbound queue length rejection in DeliverTx?

JimLarson (approved):
Some flaws, but not critical, so LGTM.

Issues identified in my comments are fixed in #6392

golang/cosmos/ante/inbound.go
mhofman: Fair. Closing shut is safer

mhofman (commented):
> Issues identified in my comments are fixed in https://github.com/Agoric/agoric-sdk/pull/6392

Adopting that last stacked commit. Will now rebase everything and merge.

Fixup: https://github.com/Agoric/agoric-sdk/compare/ac89ecd7868f2dfdee13052e9a06e2dc555e7279..46c2c935c820acf5b7563defb96aef63fe0508f2

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -29,6 +31,10 @@ const (
 	StoragePathBundles      = "bundles"
 )
 
+const MaxUint53 = 9007199254740991 // Number.MAX_SAFE_INTEGER = 2**53 - 1
```
michaelfig, line 34:
```suggestion
const MaxUint53 = 2<<52 - 1 // Number.MAX_SAFE_INTEGER = 2**53 - 1
```

golang/cosmos/x/swingset/keeper/keeper.go
```diff
@@ -78,22 +84,22 @@ func NewKeeper(
...
-		if err != nil {
-			return err
-		}
+	tail, err := k.actionQueueIndex(ctx, "tail")
+	if err != nil {
+		return err
+	}
+
+	// JS uses IEEE 754 floats so avoid overflowing integers
+	if tail == MaxUint53 {
```
michaelfig, line 101: Could be more defensive.

```suggestion
	if tail >= MaxUint53 {
```

mhofman:
Forced push without changes in hopes to nudge CI

packages/cosmic-swingset/src/launch-chain.js
michaelfig: `in` is fine, IMO.

mhofman:
Here is an example of the loadgen running with a max inbound queue of 20:

```
stage-1: client: 2022-10-05T17:23:46.565Z chain-cosmos-sdk: /workspaces/agoric/agoric-sdk/golang/cosmos/build/agd tx swingset deliver @/tmp/ag-solo-cosmos-deliver.202295-31116-mgr2nw.aikns --keyring-backend=test --gas=auto --gas-adjustment=1.2 --from=ag-solo --yes -ojson --sequence=95 --chain-id=agoriclocal --node=http://localhost:26657 --home=/workspaces/agoric/testnet-load-generator/_agstate/agoric-servers/local-solo-8000/ag-cosmos-helper-statedir
stage-1: chain: 2022-10-05T17:23:46.573Z block-manager: wrote SwingSet checkpoint [run=605ms, chainSave=1ms, kernelSave=9ms]
stage-1: chain: 5:23PM INF commit synced commit=436F6D6D697449447B5B31313720313931203437203233392033372036312032313920313034203134302038312031383120323234203133362032333320353920323437203134332031393720313434203136392039322032333920313533203837203135322034312034342032353320353120313831203538203136325D3A38447D
stage-1: chain: 5:23PM INF committed state app_hash=75BF2FEF253DDB688C51B5E088E93BF78FC590A95CEF995798292CFD33B53AA2 height=141 module=state num_txs=0
stage-1: chain: 5:23PM INF indexed block height=141 module=txindex
stage-1: client: 2022-10-05T17:23:46.711Z chain-cosmos-sdk: Error: rpc error: code = InvalidArgument desc = mempool is full: invalid request
```

Relevant begin block slogs:
```
{"type":"cosmic-swingset-begin-block","blockHeight":136,"blockTime":1664990595,"queueAllowed":{"inbound":9,"inbound_mempool":0},"time":1664990600.885193,"monotime":399.13648965400455}
{"type":"cosmic-swingset-begin-block","blockHeight":137,"blockTime":1664990600,"queueAllowed":{"inbound":11,"inbound_mempool":1},"time":1664990605.897754,"monotime":404.14905161100626}
{"type":"cosmic-swingset-begin-block","blockHeight":138,"blockTime":1664990605,"queueAllowed":{"inbound":11,"inbound_mempool":1},"time":1664990610.912435,"monotime":409.1637312540114}
{"type":"cosmic-swingset-begin-block","blockHeight":139,"blockTime":1664990610,"queueAllowed":{"inbound":9,"inbound_mempool":0},"time":1664990615.927071,"monotime":414.1783681540191}
{"type":"cosmic-swingset-begin-block","blockHeight":140,"blockTime":1664990615,"queueAllowed":{"inbound":7,"inbound_mempool":0},"time":1664990620.941607,"monotime":419.19290420302747}
{"type":"cosmic-swingset-begin-block","blockHeight":141,"blockTime":1664990620,"queueAllowed":{"inbound":7,"inbound_mempool":0},"time":1664990625.954639,"monotime":424.20593574100735}
{"type":"cosmic-swingset-begin-block","blockHeight":142,"blockTime":1664990625,"queueAllowed":{"inbound":7,"inbound_mempool":0},"time":1664990630.968132,"monotime":429.2196179060042}
{"type":"cosmic-swingset-begin-block","blockHeight":143,"blockTime":1664990630,"queueAllowed":{"inbound":11,"inbound_mempool":1},"time":1664990635.983167,"monotime":434.2346523840129}
```

michaelfig (approved):
LGTM!  Approach is good, and I have just a few minor comments.

golang/cosmos/x/swingset/keeper/keeper.go
mhofman: Oh I had tried something like that, and couldn't figure it out in Go

golang/cosmos/x/swingset/keeper/keeper.go
mhofman: Arguably we're already busted then ;)

golang/cosmos/x/swingset/keeper/keeper.go
JimLarson: Should fix either the code or the comment to settle on 52 vs 53.

JimLarson: And the right answer: can exactly store `2**53` in a float, `2**53 + 1` is the first integer that can't be represented in a double float.

mhofman: The code and the comment are both correct: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

mhofman: To clarify we can rewrite it as `1<<53 - 1` next time we touch this.

PR #6383, opened 2022-10-04
wallet ui misc

turadg:
## Description

- [feat: show git sha of build source](https://github.com/Agoric/agoric-sdk/commit/950b64cb41a36e447d0be8e2534e0ce47361f11d)
- [refactor: gasPriceStep to new location](https://github.com/Agoric/agoric-sdk/commit/482c2563bbc5df223cf0c4244ba0bf5f965dfd6f) to stop deprecation notice in console
- [chore: bump cosmjs and kepl-wallet deps](https://github.com/Agoric/agoric-sdk/commit/084a5e1edef21b52250a4f2fda8e9131d6fd0b28) to get types to match Keplr extension version and misc bug fixes from cosmjs (https://github.com/cosmos/cosmjs/blob/main/CHANGELOG.md#0290---2022-09-15 ) 

### Security Considerations

Exposes the git commit history into the build. Should be read as "alleged" since a build can put there whatever it wants.

--

### Testing Considerations

We're about to do a QA party on the wallet, so we'll cover this.


samsiegart (approved):
It might be nice to show the version in the UI somewhere, like in small grey text at the bottom of the nav menu, but not a blocker.

PR #6390, opened 2022-10-04
fix(ci): handle breaking change in skip duplicate action

mhofman:
Update `runId` -> `id`, which is a breaking change in the action we use to skip duplicate tests (see https://github.com/fkirc/skip-duplicate-actions/pull/269)
Pin version to avoid future breaking issues


warner (approved):
LGTM

PR #6391, opened 2022-10-04
fix(casting): Pin endpoint between getBlockHeight and getData

### LINKED ISSUE #6384: get latest RPC data in a single HTTP request

```
Something in Stargate?
```

## Description of the Design

## Security Considerations
--
## Test Plan

--

kriskowal:
- refactor(casting): Loosen block height and endpoint
- fix(casting): Pin endpoint between getBlockHeight and getData

refs: #6384

## Description

This change includes a refactoring of `casting` such that:
- We can optionally omit the block height for a query and get data for the latest block.
- We can optionally track the `endpoint` that provided the response for a query.
- We can optionally send a query to a specific `endpoint`.

Then, we thread the `endpoint` from each `getBlockHeight` request into the next `getData` request, ensuring the block exists on the node.

This does not obviate the need for `getBlockHeight` yet.
With https://github.com/cosmos/cosmjs/pull/1278, we would be able to also thread the `blockHeight` through the response path of `getData`, elliminating a round trip for `getBlockHeight`.

We could possibly close #6384, but only if we tracked that further optimization in a fresh ticket, so I've chosen just to reference it.

### Security Considerations

No.

No.

### Testing Considerations

Integration testing for this new behavior requires standing up a cluster of fake RPC nodes and teasing out a race where `getBlockHeight` and `getData` requests were likely to occur on separate RPC nodes. That might be ensured by having just a pair of RPC nodes and a fake leader that doesn't attempt to favor previously used nodes.

### DIFF
--- a/packages/casting/src/follower-cosmjs.js
+++ b/packages/casting/src/follower-cosmjs.js
@@ -74,7 +74,7 @@ const collectSingle = values => {
   return head[0];
 };
 
-// Coordinate with switch/case of tryGetDataAtHeight.
+// Coordinate with switch/case of tryGetData.
 const proofs = ['strict', 'none', 'optimistic'];
 
 /**
@@ -136,7 +136,7 @@ export const makeCosmjsFollower = (
       const info = await client.abciInfo();
       const { lastBlockHeight } = info;
       assert.typeof(lastBlockHeight, 'number');
-      return lastBlockHeight;
+      return { blockHeight: lastBlockHeight, endpoint };
     });
     return collectSingle(values);
   };
@@ -163,8 +163,13 @@ export const makeCosmjsFollower = (
 
   /**
    * @param {(endpoint: string, storeName: string, storeSubkey: Uint8Array) => Promise<Uint8Array>} tryGetPrefixedData
+   * @param {string} [specificEndpoint]
+   * @returns {Promise<{data: Uint8Array, endpoint: string}>}
    */
-  const retryGetDataAndStripPrefix = async tryGetPrefixedData => {
+  const retryGetDataAndStripPrefix = async (
+    tryGetPrefixedData,
+    specificEndpoint,
+  ) => {
     const {
       storeName,
       storeSubkey,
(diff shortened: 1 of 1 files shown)


packages/casting/src/follower-cosmjs.js
```diff
@@ -181,65 +186,93 @@ export const makeCosmjsFollower = (
...
-        },
-      ),
-    );
+    let data;
+    let endpoint;
+    if (specificEndpoint !== undefined) {
+      endpoint = specificEndpoint;
+      data = await tryGetPrefixedData(endpoint, storeName, storeSubkey);
+    } else {
+      // mapEndpoints is our retry loop.
```
turadg, line 195: Weird name huh? `map` makes me think I get back a collection with cardinality of endpoints. WDYT of `tryEndpoints`?

packages/casting/src/follower-cosmjs.js
```diff
@@ -409,15 +447,22 @@ export const makeCosmjsFollower = (
...
       if (currentBlockHeight === blockHeight) {
         // TODO Long-poll for next block
         // https://github.com/Agoric/agoric-sdk/issues/6154
         await E(leader).jitter(where);
         continue;
       }
 
-      const currentData = await getDataAtHeight(currentBlockHeight);
+      // Carry endpoint that responded with the current block height forward to
+      // getData so we are sure that the endpoint has data for that height.
```
turadg, line 460: 👌

packages/casting/src/follower-cosmjs.js
```diff
@@ -595,17 +658,17 @@ export const makeCosmjsFollower = (
     async getLatestIterable() {
       return getLatestIterable();
     },
-    async getEachIterable({ height = undefined } = {}) {
-      if (height === undefined) {
-        height = await getBlockHeight();
+    async getEachIterable({ height: blockHeight = undefined } = {}) {
+      if (blockHeight === undefined) {
+        ({ blockHeight } = await getBlockHeight());
```
turadg, line 663: `height` is unambiguous in this context. consider doing the rename in line 663 instead, keeping `height` as the local binding.

even better perhaps would be to have `getBlockHeight` return have `{ height }`. It's obviously _block_ height.

packages/casting/src/follower-cosmjs.js
```diff
@@ -136,7 +136,7 @@ export const makeCosmjsFollower = (
       const info = await client.abciInfo();
       const { lastBlockHeight } = info;
       assert.typeof(lastBlockHeight, 'number');
-      return lastBlockHeight;
+      return { blockHeight: lastBlockHeight, endpoint };
```
turadg, line 139: a little awkward to be passing around the `endpoint`. that should be encapsulated in the leader, which I think we could do by refactoring to have a single callback to `mapEndpoints` that does the right thing.

but given this will be refactored again with the cosmjs change, I think this is okay. consider a `xxx` comment about the factoring.

turadg (approved):
> Integration testing for this new behavior requires standing up a cluster

We have such a cluster! Load testing on ollinet produces this error pretty reliably. @arirubinstein has scripts and a long running tmux on GCP machine. I'll give it a whirl.

turadg:
We've diagnosed the problem as round-robin on the server side. Closing this so it doesn't appear the solving it is in progress.

PR #6392, opened 2022-10-05
feat: limit inbound queue (v2)

### LINKED ISSUE #6306: how to prevent unbounded inboundQueue growth and excessive trade latency

JimLarson:
closes: #6306 #6313 #6314

## Description

Extension of #6381 with a final fix for correctly detecting CheckTx mode in the ante handler.

### Security Considerations

See #6381

See #6381

### Testing Considerations

See #6381


mhofman:
Superseded by #6381

PR #6395, opened 2022-10-05
feat(wallet-ui): require 100IST in pool to provision smart wallet

### LINKED ISSUE #6376: Prevent buying a smart wallet when the provision pool is below 100IST

As noted in #6375, provisioning a smart wallet can fail due to the provision pool being empty, and we don't currently have a mechanism to refund the BLD.

## Description of the Design

When the provision pool is below 100IST, disable the web UI for buying a smart wallet.

The amount in the pool is `totalMintedConverted` - `totalMintedProvided`.

```
$ agoric follow :published.provisionPool.metrics
{
  totalMintedConverted: {
    brand: slot(0,"Alleged: IST brand"),
    value: 999900000n,
  },
  totalMintedProvided: {
    brand: slot(0),
    value: 250000n,
  },
  walletsProvisioned: 1n,
}
```

## Security Considerations

This is a Web UI work-around for #6375 ; clients that make provision requests by other means (e.g. `agd tx swingset provision ...`) are not protected.

## Test Plan

_TBD_

samsiegart:
fixes: https://github.com/Agoric/agoric-sdk/issues/6376

### DIFF
--- a/packages/wallet/ui/src/components/ProvisionDialog.jsx
+++ b/packages/wallet/ui/src/components/ProvisionDialog.jsx
@@ -1,4 +1,4 @@
-import { makeFollower, iterateEach } from '@agoric/casting';
+import { makeFollower, iterateLatest } from '@agoric/casting';
 import CircularProgress from '@mui/material/CircularProgress';
 import Dialog from '@mui/material/Dialog';
 import DialogTitle from '@mui/material/DialogTitle';
@@ -27,9 +27,17 @@ const errors = {
 const CREATION_FEE = '10 BLD';
 
 // 100 IST
-const MINIMUM_PROVISION_POOL_BALANCE = 100_000_000n;
+const MINIMUM_PROVISION_POOL_BALANCE = 100n * 1_000_000n;
 
-/** @typedef {import('@agoric/vats/src/provisionPool.js').MetricsNotification} ProvisionPoolMetrics */
+// XXX import from the contract
+
+/**
+ * @typedef {object} ProvisionPoolMetrics
+ * @property {bigint} walletsProvisioned  count of new wallets provisioned
+ * @property {Amount<'nat'>} totalMintedProvided  running sum of Minted provided to new wallets
+ * @property {Amount<'nat'>} totalMintedConverted  running sum of Minted
+ * ever received by the contract from PSM
+ */
 
 export const useProvisionPoolMetrics = (unserializer, leader) => {
   const [data, setData] = useState(/** @type {ProvisionPoolMetrics?} */ (null));
@@ -44,14 +52,17 @@ export const useProvisionPoolMetrics = (unserializer, leader) => {
           unserializer,
         },
       );
-      for await (const { value } of iterateEach(follower)) {
+      for await (const { value } of iterateLatest(follower)) {
         if (cancelled) {
           break;
         }
+        console.log('provisionPoolData', value);
         setData(value);
       }
     };
-    fetchData().catch(e => console.error('useEffect error', e));
+    fetchData().catch(e =>
+      console.error('useProvisionPoolMetrics fetchData error', e),
+    );
     return () => {
       cancelled = true;
     };
@@ -59,6 +70,18 @@ export const useProvisionPoolMetrics = (unserializer, leader) => {
 
   return data;
 };
+
+/**
+ *
+ * @param {ProvisionPoolMetrics} provisionPoolData
+ * @returns {boolean}
+ */
+const isProvisionPoolLow = provisionPoolData =>
+  provisionPoolData &&
+  AmountMath.subtract(
+    provisionPoolData.totalMintedConverted,
+    provisionPoolData.totalMintedProvided,
+  ).value < MINIMUM_PROVISION_POOL_BALANCE;
 
 const ProvisionDialog = ({
   onClose,
@@ -143,16 +166,7 @@ const ProvisionDialog = ({
     }
   }, [currentStep, href, address]);
 
-  console.log('provisionPoolData', provisionPoolData);
-  const provisionPoolLow = useMemo(
-    () =>
-      provisionPoolData &&
-      AmountMath.subtract(
-        provisionPoolData.totalMintedConverted,
-        provisionPoolData.totalMintedProvided,
-      ).value < MINIMUM_PROVISION_POOL_BALANCE,
-    [provisionPoolData],
-  );
+  const provisionPoolLow = isProvisionPoolLow(provisionPoolData);
 
   return (
     <Dialog open={open}>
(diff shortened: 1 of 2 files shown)


samsiegart:
@turadg Agh, any ideas how to fix the lint:types in CI? I can always just remove the type...

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -24,9 +26,52 @@ const errors = {
 // TODO: Read this from the chain via rpc.
 const CREATION_FEE = '10 BLD';
 
-const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
+// 100 IST
+const MINIMUM_PROVISION_POOL_BALANCE = 100_000_000n;
```
turadg: consider,
```suggestion
const MINIMUM_PROVISION_POOL_BALANCE = 100 * 1_000_000n;
```
to call separate the "100"

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -24,9 +26,52 @@ const errors = {
...
+
+/** @typedef {import('@agoric/vats/src/provisionPool.js').MetricsNotification} ProvisionPoolMetrics */
```
turadg: the lint:types thing will require shaving many yaks. for now:
```suggestion
// XXX import from the contract
/**
 * @typedef {object} ProvisionPoolMetrics
 * @property {bigint} walletsProvisioned  count of new wallets provisioned
 * @property {Amount<'nat'>} totalMintedProvided  running sum of Minted provided to new wallets
 * @property {Amount<'nat'>} totalMintedConverted  running sum of Minted
 * ever received by the contract from PSM
 */
```

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -24,9 +26,52 @@ const errors = {
...
+    let cancelled = false;
+    const fetchData = async () => {
+      const follower = await makeFollower(
+        `:published.provisionPool.metrics`,
+        leader,
+        {
+          unserializer,
+        },
+      );
+      for await (const { value } of iterateEach(follower)) {
```
turadg: I think we want `iterateLatest`. The intermediate states don't matter

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -24,9 +26,52 @@ const errors = {
...
+        if (cancelled) {
+          break;
+        }
+        setData(value);
+      }
+    };
+    fetchData().catch(e => console.error('useEffect error', e));
```
turadg:
```suggestion
    fetchData().catch(e => console.error('useProvisionPoolMetrics fetchData error', e));
```

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -96,7 +141,18 @@ const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
...
+  console.log('provisionPoolData', provisionPoolData);
+  const provisionPoolLow = useMemo(
+    () =>
+      provisionPoolData &&
+      AmountMath.subtract(
+        provisionPoolData.totalMintedConverted,
+        provisionPoolData.totalMintedProvided,
+      ).value < MINIMUM_PROVISION_POOL_BALANCE,
+    [provisionPoolData],
+  );
```
turadg: this is a very cheap calculation that's probably not worth memoizing.

also it's outside the rendering domain so consider pulling out into a non-React function. maybe even a different module that has the constant too. that could be unit tested.

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -105,8 +161,14 @@ const ProvisionDialog = ({ onClose, open, address, href, keplrConnection }) => {
       </DialogTitle>
       <DialogContent>
         {content}
+        {provisionPoolLow && (
+          <DialogContentText sx={{ pt: 2 }} color="error">
+            Unable to create a Smart Wallet because the provision pool balance
+            is too low.
```
turadg: I bet users are going to have a hard time making sense of this.
```suggestion
            The pool of funds to provision smart wallets is too small at this time.
```

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Pulled into a different function but not a module for now. The wallet UI still uses jest so we can't use ses and hence can't test any amountmath stuff without mocking a lot.

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Done, thanks for catching!

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Done, thanks

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Done

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Done

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Done. This is still probably something they'll have to ask about in our social channels regardless, but hopefully no one runs into this at all in any regular circumstance.

packages/wallet/ui/src/components/ProvisionDialog.jsx
```diff
@@ -60,6 +71,18 @@ export const useProvisionPoolMetrics = (unserializer, leader) => {
   return data;
 };
 
+/**
+ *
+ * @param {ProvisionPoolMetrics} provisionPoolData
```
turadg, line 77: non-blocking: the current factoring is really:
```suggestion
 * @param {ProvisionPoolMetrics=} provisionPoolData
```
It would be tighter to make this take a defined type and put the lax condition at the call site,
```
const provisionPoolLow = provisionPoolData && isProvisionPoolLow(provisionPoolData);
```

turadg (approved):

packages/wallet/ui/src/components/ProvisionDialog.jsx
samsiegart: Thanks, just added `ts-check` and fixed

PR #6397, opened 2022-10-05
fix(provisionPool): update metrics on trades

### LINKED ISSUE #6396: provisionPool metrics are not updated on swap
While working on #6376, @samsiegart noted that after the provisionPool is funded, metrics for `totalMintedConverted` still shows `0n`.

Looks like we don't publish updated metrics after that swap:

https://github.com/Agoric/agoric-sdk/blob/f9b201c03f0a02c76280b70e207324b98561e4b3/packages/vats/src/provisionPool.js#L143-L152

dckc:
closes: #6396

### DIFF
--- a/packages/vats/src/provisionPool.js
+++ b/packages/vats/src/provisionPool.js
@@ -35,12 +35,10 @@ const privateArgsShape = harden({
  * handle requests to provision smart wallets.
  *
  * @param {(address: string, depositBank: ERef<Bank>) => Promise<void>} sendInitialPayment
- * @param {(count: bigint) => void} publishMetrics
+ * @param {() => void} onProvisioned
  * @typedef {import('./vat-bank.js').Bank} Bank
  */
-export const makeBridgeProvisionTool = (sendInitialPayment, publishMetrics) => {
-  let provisionedCount = 0n;
-
+export const makeBridgeProvisionTool = (sendInitialPayment, onProvisioned) => {
   /**
    * @param {{
    *   bankManager: ERef<BankManager>,
@@ -73,9 +71,8 @@ export const makeBridgeProvisionTool = (sendInitialPayment, publishMetrics) => {
           namesByAddressAdmin,
         );
         if (created) {
-          provisionedCount += 1n;
+          onProvisioned();
         }
-        publishMetrics(provisionedCount);
         console.info(created ? 'provisioned' : 're-provisioned', address);
       },
     });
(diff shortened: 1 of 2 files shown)


packages/vats/test/test-provisionPool.js
```diff
@@ -197,7 +198,11 @@ test('provisionPool trades provided assets for IST', async t => {
     initialPoserInvitation,
   );
 
+  // mock gov terms
+  /** @type {Instance} */
+  const electionManager = /** @type {any} */ (null);
   const govTerms = {
+    electionManager,
```
turadg:
```suggestion
    electionManager: /** @type {any} */ (null),
```

packages/vats/src/provisionPool.js
```diff
@@ -120,24 +115,47 @@ export const start = async (zcf, privateArgs) => {
...
-    privateArgs.marshaller,
-  );
-  const updateMetrics = walletsProvisioned => {
-    metricsPublisher.publish(
-      harden({
-        walletsProvisioned,
-        totalMintedProvided,
-        totalMintedConverted,
-      }),
+  const makeMetrics = () => {
```
turadg, line 118: neat encapsulation. We may want to move it back into main scope for durability but we can evaluate that then.

packages/vats/src/provisionPool.js
```diff
@@ -120,24 +115,47 @@ export const start = async (zcf, privateArgs) => {
...
+    const { metricsPublisher, metricsSubscriber } = makeMetricsPublishKit(
+      privateArgs.storageNode,
+      privateArgs.marshaller,
     );
+
+    let walletsProvisioned = 0n;
+    let totalMintedProvided = AmountMath.makeEmpty(poolBrand);
+    let totalMintedConverted = AmountMath.makeEmpty(poolBrand);
+
+    const updateMetrics = () => {
```
turadg: all these onFoo handlers are updating the (local) metrics. let's use "update" to mean revisions to the truth and "publish" to notify subscribers
```suggestion
    const publishMetrics = () => {
```

turadg (approved):

turadg:
The fix looks good. Some code quality suggestions but more important is to get this into the next rc.

samsiegart:
Verified I can no longer repro the original issue on this patch

samsiegart (approved):

PR #6398, opened 2022-10-05
simplify Wallet UI network chooser

turadg:
## Description

The Wallet UI connection settings dialog lets users enter arbitrary URLs for a network config, but usually it's one of a few well-known URLs. (Not quite a [Well-known URI](https://en.wikipedia.org/wiki/Well-known_URI) but it could be.)

This **puts an element above the URL chooser to pick from a simpler list**: Main, Testnet, Devnet, Localhost. Picking any of these shows the URL in the box but not editable. To edit, pick "Custom URL" from the first menu. 

This also **removes the connection method picker** so that it's always Keplr. That reduces complexity for dapps that are bridging to this wallet, that they don't have to handle and notify of Read-only mode. For read-only use cases, the CLI works well.

It also makes **main the default network**, such that it's the setting before a user picks a network. 

### Security Considerations

Hides the path on the host that the network config is being read from. It was a canonical path before so this is hardly a difference. If we're concerned about malicious contents of the `/network-config` path we should display the actual config.

--

### Testing Considerations

Tested manually, picking different hosts.

- [x] ability to paste a custom URL
- [x] when user opens for the first time, automatically connect to Mainnet


dckc (commented):
I'm pretty strongly against this. Hope to find time to make a coherent argument to request changes.

samsiegart (commented):
I'm in favor of c2a6f5fd6c909ff9c8e4da0ee4412b7ce3ea0ea6, though perhaps we should make it explicit somewhere now that you need Keplr, since you don't have to select it anymore.

I don't think we should remove the network-config part of the URL though.

dckc:
Does this take away the read-only address feature? That's a pretty common wallet feature, and I found it handy a few times. But I won't argue that we need it, or even that my preferences are more important than simplicity for a larger user base.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -59,11 +70,20 @@ const ConnectionSettingsDialog = ({
...
-  const smartConnectionHrefs = allConnectionConfigs.map(({ href }) => href);
+  /** @type {string[]} */
+  const smartConnectionHrefs = allConnectionConfigs.map(c => c.href);
 
-  const [config, setConfig] = useState(connectionConfig || {});
+  const [configSource, setConfigSource] = useState(
+    /** @type {keyof KnownNetworkConfigUrls | 'custom'} */ (
+      connectionConfig
+        ? networkConfigUrl.toSource(connectionConfig.href)
+        : 'main'
```
samsiegart: Nit: Can this be a constant somewhere?

samsiegart (approved):

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
turadg: Good idea. Actually since mainnet is default config now, there's no need to have a default here. I've pushed an amendment

PR #6405, opened 2022-10-05
chore(wallet-ui): update token icons

### LINKED ISSUE #6318: Update token logos in smart wallet UI
Need:
New BLD Logo
New IST logo
Placeholder logo for anything we don't know

samsiegart:
fixes https://github.com/Agoric/agoric-sdk/issues/6318

image


packages/wallet/ui/src/components/BrandIcon.jsx
```diff
@@ -0,0 +1,16 @@
+import Box from '@mui/material/Box';
+import { icons, defaultIcon } from '../util/Icons.js';
+
+const BrandIcon = ({ brandPetname }) => {
```
turadg, line 4: it's just a `name`, right? petnames are arbitrary per user and this is keyed from source control

turadg (approved):
Some concern about "petname" handling but not blocking as I expect we'll revisit it when we restore pet name editing.

PR #6407, opened 2022-10-06
Wallet fixes

### LINKED ISSUE #6393: Wallet UI Needs Refresh After Account Creation
## Describe the bug

After creating a smart wallet from the dialog action, the dialog that shows a spinner while the wallet creation is in progress never goes away, even after the wallet is created and purses are loaded. The user needs to refresh the page to get it to go away.

## To Reproduce
1. Load the wallet ui and connect to a chain+address that doesn't have a smart wallet yet
2. Click the button in the dialog to create the wallet
3. Accept the offer in keplr
4. Observe the spinner dialog shows forever after the purses load

## Expected behavior
The dialog disappears after the wallet is created

## Screenshots
image

### LINKED ISSUE #6394: Wallet UI doesn't show error when connection fails
## Describe the bug
When the wallet UI fails to load or connect to the user's wallet, it used to show a toast and an error message. Now it just shows an empty purses box with no other indication of an error. This seems to only happen when you switch network configs. If you refresh the page starting from a bad config, it will show the error.

## To Reproduce
1. Go to the wallet UI
2. Open the network config dialog
3. Choose a network config that won't load, like localhost:1234 with no local chain running.
4. Observe the screenshot below

## Expected behavior
An error toast appears and an error icon+message shows up in place of the loading indicator.

## Screenshots
image

### LINKED ISSUE #6408: Offer status not updating after PSM trade
## Describe the bug
After making a PSM trade, the offer still shows as pending even after the assets have been swapped and the purses have updated.
image

samsiegart:
fixes https://github.com/Agoric/agoric-sdk/issues/6394
fixes https://github.com/Agoric/agoric-sdk/issues/6393
fixes https://github.com/Agoric/agoric-sdk/issues/6408

### DIFF
--- a/packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
+++ b/packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
@@ -127,7 +127,7 @@ const ConnectionSettingsDialog = ({
                 break;
               case 'localhost':
                 setConfig({
-                  href: `http://localhost:8000/wallet/network-config`,
+                  href: `http://localhost:3000/wallet/network-config`,
                 });
                 break;
               case 'custom':
@@ -136,12 +136,12 @@ const ConnectionSettingsDialog = ({
             }
           }}
         >
-          <MenuItem value="main">Main</MenuItem>
+          <MenuItem value="main">Mainnet</MenuItem>
           <MenuItem value="testnet">Testnet</MenuItem>
           <MenuItem value="devnet">Devnet</MenuItem>
-          <MenuItem value="localhost">localhost</MenuItem>
+          <MenuItem value="localhost">Localhost</MenuItem>
           <MenuItem value="custom">
-            <i>Custom url</i>
+            <i>Custom URL</i>
           </MenuItem>
         </Select>
       </FormControl>
@@ -165,9 +165,6 @@ const ConnectionSettingsDialog = ({
           return options;
         }}
         renderOption={(props, option) => <li {...props}>{option}</li>}
-        selectOnFocus
-        clearOnBlur
-        handleHomeEndKeys
         freeSolo
         renderInput={params => <TextField {...params} label="Network URL" />}
       />
--- a/packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
+++ b/packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
@@ -118,7 +118,7 @@ const ConnectionSettingsDialog = ({
             const { value } = e.target;
             setConfigSource(value);
             switch (value) {
-              case 'mainnet':
+              case 'main':
               case 'testnet':
               case 'devnet':
                 setConfig({
--- a/packages/wallet/ui/src/service/Offers.js
+++ b/packages/wallet/ui/src/service/Offers.js
@@ -62,13 +62,10 @@ export const getOfferService = (
     for await (const state of makeAsyncIterableFromNotifier(
       chainOffersNotifier,
     )) {
-      state?.forEach(offer => {
-        console.log('chain offer', offer);
-        const splitId = offer.id.split('#');
-        const rawId = splitId[splitId.length - 1];
-        if (offers.has(rawId)) {
-          offers.delete(rawId);
-          remove(publicAddress, rawId);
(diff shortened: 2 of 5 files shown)


packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -127,7 +127,7 @@ const ConnectionSettingsDialog = ({
                 break;
               case 'localhost':
                 setConfig({
-                  href: `http://localhost:8000/wallet/network-config`,
+                  href: `http://localhost:3000/wallet/network-config`,
```
samsiegart, line 130: 8000 was never used anymore so just change this to 3000

samsiegart:
All of these bugs seem to be a result of https://github.com/Agoric/agoric-sdk/commit/703eecd0af917cd432d1f6a12e645a12ab9d2c39, we should have manually tested more carefully

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
turadg: Ok great. We should update all the docs then too. 8000 is most of what I found when searching

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -118,7 +118,7 @@ const ConnectionSettingsDialog = ({
             const { value } = e.target;
             setConfigSource(value);
             switch (value) {
-              case 'mainnet':
+              case 'main':
```
turadg, line 121: Oops. Something very unit testable. I'll try to find time to write a regression test.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -136,12 +136,12 @@ const ConnectionSettingsDialog = ({
             }
           }}
         >
-          <MenuItem value="main">Main</MenuItem>
+          <MenuItem value="main">Mainnet</MenuItem>
           <MenuItem value="testnet">Testnet</MenuItem>
           <MenuItem value="devnet">Devnet</MenuItem>
-          <MenuItem value="localhost">localhost</MenuItem>
+          <MenuItem value="localhost">Localhost</MenuItem>
```
turadg, line 142: I wanted it lower caps since it's not a network name and the actual host name

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
```diff
@@ -136,12 +136,12 @@ const ConnectionSettingsDialog = ({
             }
           }}
         >
-          <MenuItem value="main">Main</MenuItem>
```
turadg, line 139: I figured it's better to match "main.Agoric.net"

turadg (approved):
I disagree on some the UX choices but defer to you. 

Thanks for diagnosing and fixing the bugs.

turadg:
> All of these bugs seem to be a result of https://github.com/Agoric/agoric-sdk/commit/703eecd0af917cd432d1f6a12e645a12ab9d2c39, we should have manually tested more carefully

Agreed, given the state of test coverage. I think it would be worthwhile to invest more in automated testing, both locally and integration with a chain.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
samsiegart: 8000 was used for ag-solo I think which is probably why

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
samsiegart: It just looked a bit jarring since the others are caps. The others are titles/labels, not hostnames, so I prefer having this consistency.

packages/wallet/ui/src/components/ConnectionSettingsDialog.jsx
samsiegart: The user knows the network as "Mainnet", I don't think these need to match the hostnames

PR #6412, opened 2022-10-06
docs(swingset): remove stale badges from README

warner:
closes #6409

PR #6413, opened 2022-10-06
build: Use larger, auto-scaling runners for our GitHub workflows

### LINKED ISSUE #6410: Use larger, auto-scaling runners for our GitHub workflows
GitHub now supports using auto-scaling numbers of larger runners for Workflows. 

Let's try it out and see if it improves our CI time.

https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

Tartuffo:
closes: #6410
refs: #XXXX

## Description

Use this new runner: 
image

### DIFF
--- a/.github/workflows/deployment-test.yml
+++ b/.github/workflows/deployment-test.yml
@@ -27,7 +27,7 @@ jobs:
   deployment-test:
     needs: pre_check
     if: needs.pre_check.outputs.should_run == 'true'
-    runs-on: ubuntu-18.04 # trusty
+    runs-on: NewAutoScaled
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/after-merge.yml
+++ b/.github/workflows/after-merge.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   build:
     if: ${{ github.repository_owner == 'agoric' }}
-    runs-on: ubuntu-latest
+    runs-on: NewAutoScaled
     strategy:
       matrix:
         node-version: ['14.x', '16.x']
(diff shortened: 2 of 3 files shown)


.github/workflows/deployment-test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
   deployment-test:
     needs: pre_check
     if: needs.pre_check.outputs.should_run == 'true'
-    runs-on: ubuntu-18.04 # trusty
+    runs-on: NewAutoScaled
```
mhofman, line 30: Yeah that's gonna be one of the complications. For various reasons must still run the deployment test on an outdated version of Ubuntu. Last time @michaelfig and I tried to update we ran into trouble and decided to punt.

warner:
Huh, I'm looking at the timing of the checks (the "Successful in NNNm" lines in the merge-is-block box above (which will be hidden once this PR lands)), comparing it against the old runner (e.g. #6412, which ran CI at about the same time), and it looks like the speedup is pretty small, and in some cases negative. Lemme collect more data, and look into the way the repo is configured (does `NewAutoScaled` mean fast*er*, or just "could be faster if you click a box that says you'll pay for it"?).

warner:
The runners config (https://github.com/organizations/Agoric/settings/actions/github-hosted-runners/1/edit for those with perms) doesn't mention anything about the speed of the machines, only the "Maximum runners" (currently set to 20). I'm guessing this would let us handle more simultaneous jobs, but wouldn't make any individual job any faster. The way we'd get more throughput would be to further break up our test suite into even smaller pieces, to use more parallelism.

Does that match other people's intuitions?

michaelfig:
Closing due to this PR languishing.  Feel free to reopen if it still seems useful.

PR #6388, opened 2022-10-04
fix: swingset amino codec / Ledger support

dckc:
closes: #6347
refs: #3628

## Description

Fix amino codec details both in `x/swingset` and `wallet/ui`.

Also, avoid collisions with `agoric` chain name in Keplr.

### Security Considerations

Being able to use a Ledger is a pretty big security feature :)

### Documentation Considerations

@rowgraus I guess this should go in the user docs?

Big thanks to the Keplr folks for diagnosing the problem!

### Testing Considerations

@michaelfig tested `agd` and the wallet UI; I also tested the wallet UI.

image

@turadg wondered about integration testing with [zemu](https://github.com/Zondax/zemu). I'm inclined to leave that out of scope.

Issue #6347, opened 2022-09-28
Can't create a Smart Wallet from Web UI using a Ledger


Tartuffo:
## Describe the bug

When trying to create a Smart Wallet using a Ledger + Keplr, you get the error message in the screenshot below.

## To Reproduce

Attempt to create a smart wallet using a Ledger with Keplr.

## Expected behavior
It should work, or give a more meaningful error message.

## Platform Environment
Unknown

## Additional context

Stack trace:
```
lockdown.umd.js:6551 Failed to accept offer (Error#9)
lockdown.umd.js:6512 Error#9: Multiple types are registered with Amino type identifier 'swingset/WalletAction': '/agoric.swingset.MsgWalletAction', '/agoric.swingset.MsgWalletSpendAction'. Thus fromAmino cannot be performed.
logError	@	lockdown.umd.js:6512
lockdown.umd.js:6534   at t.AminoTypes.fromAmino (https://wallet.agoric.app/wallet/static/js/main.ee775e33.js:2:241770)
  at https://wallet.agoric.app/wallet/static/js/main.ee775e33.js:2:304137
  at Array.map (<anonymous>)
  at E.signAmino (https://wallet.agoric.app/wallet/static/js/main.ee775e33.js:2:304113)
  at async E.signAndBroadcast (https://wallet.agoric.app/wallet/static/js/main.ee775e33.js:2:303157)
  at async Object.submitSpendAction (https://wallet.agoric.app/wallet/static/js/main.ee775e33.js:102:2166375)
DevTools failed to load source map: Could not load content for chrome-extension://dmkamcknogkgcdfhhbddcghachkejeap/browser-polyfill.js.map: System error: net::ERR_BLOCKED_BY_CLIENT
lockdown.umd.js:6551 
{accountNumber: 134, sequence: 0}
lockdown.umd.js:6551 sign spend action 
{address: 'agoric15v3haayjfj9ycyfsqj7gs4grxktgpgkndpv95h', msgs: Array(1), fee: {…}}
lockdown.umd.js:6551 Failed to accept offer (Error#10)
lockdown.umd.js:6512 Error#10: Multiple types are registered with Amino type identifier 'swingset/WalletAction': '/agoric.swingset.MsgWalletAction', '/agoric.swingset.MsgWalletSpendAction'. Thus fromAmino cannot be performed.
[...]

PR #6415, opened 2022-10-06
fix(wallet/ui): signing offers (PSM trades) with Ledger

dckc:
refs: #6388 , #3628 , #6347

## Description

spend_action is snake case, unlike powerFlags

the [.proto definitions](https://github.com/Agoric/agoric-sdk/blob/master/golang/cosmos/proto/agoric/swingset/msgs.proto) aren't parallel. oh well.

### Testing Considerations

tested manually. again. hm.

### DIFF
--- a/packages/wallet/ui/src/util/keyManagement.js
+++ b/packages/wallet/ui/src/util/keyManagement.js
@@ -191,10 +191,10 @@ const SwingsetConverters = {
   [SwingsetMsgs.MsgWalletSpendAction.typeUrl]: {
     aminoType: SwingsetMsgs.MsgWalletSpendAction.aminoType,
     toAmino: ({ spendAction, owner }) => ({
-      spendAction,
+      spend_action: spendAction,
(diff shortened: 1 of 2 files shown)


michaelfig (approved):

packages/wallet/ui/src/util/keyManagement.js
```diff
@@ -191,10 +191,10 @@ const SwingsetConverters = {
   [SwingsetMsgs.MsgWalletSpendAction.typeUrl]: {
     aminoType: SwingsetMsgs.MsgWalletSpendAction.aminoType,
     toAmino: ({ spendAction, owner }) => ({
-      spendAction,
+      spend_action: spendAction,
```
turadg, line 194: a regression test would be great. unit or static.

turadg (approved):

packages/wallet/ui/src/util/keyManagement.js
dckc: a test of the message format is pretty cheap; I'll start on it...

Issue #6416, opened 2022-10-06
Sometimes a PSM transfer succeeds, but doesn't clear the `offer` display


Chris-Hibbert:
## Describe the bug

Sometimes a PSM transfer succeeds, but doesn't clear the `offer` display

## To Reproduce
1. I did some PSM trades on ollinet. They seemed to succeed. I had less of one currency and more of another.
2.  But at the end, I had an "offer" pane that won't go away.

![OfferPaneRemains](https://user-images.githubusercontent.com/13372636/194419406-75528f1d-c895-4cac-9edf-f5f38c40e94d.png)

3. There's an "exit" button.  I'll just push that.
4. No change in the pane, but in the console, I see `TODO: cancel offer`.

## Expected behavior

The offer pane should transition to `successful`. If it doesn't, it should let me clear it.

## Platform Environment

Running in Chrome

## Additional context

https://github.com/Agoric/agoric-sdk/blob/b63ae3ab963d2c27af4c630dc73ca5f75a3e0f9e/packages/wallet/ui/src/service/Offers.js#L47

samsiegart:
@Chris-Hibbert Did this happen with every offer, or did some offers eventually show an accepted state and let you dismiss them?

Chris-Hibbert:
During today's testing, it happened for me with every offer. I think @Turadg was testing at the same time, and they all closed for him. (At least he didn't say he was seeing the same problem.)

samsiegart:
> During today's testing, it happened for me with every offer.

Interesting, we used to have https://github.com/Agoric/agoric-sdk/issues/6408 which resulted in it happening to every offer. But if @turadg never saw it, and neither have I since closing that issue, then the fact that it happens every time for you makes me wonder if you just had an older cached version of the site. Sometimes we have seen things fail to update from RPC errors but it wouldn't happen 100% of the time

samsiegart:
@Chris-Hibbert To see if it's a caching problem, can you try this version of the wallet: https://7db031d0.wallet-app.pages.dev/wallet/

You'll have to go to https://wallet.agoric.app/locator/ and change that input to "https://7db031d0.wallet-app.pages.dev" (*make sure there's no trailing slash*)

Chris-Hibbert:
This time it succeeded promptly. (once @turadg walked me through changing the URL to `ollinet`).  It worked now.

samsiegart:
@Chris-Hibbert and @turadg thanks for verifying. Can we close this issue then?

PR #6418, opened 2022-10-07
chore(wallet-ui): disable exit offer button

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6416

### DIFF
--- a/packages/wallet/ui/src/components/Offer.jsx
+++ b/packages/wallet/ui/src/components/Offer.jsx
@@ -12,6 +12,9 @@ import Proposal from './Proposal';
 
 import './Offer.scss';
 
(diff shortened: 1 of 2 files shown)


packages/wallet/ui/src/components/Offer.jsx
```diff
@@ -12,6 +12,9 @@ import Proposal from './Proposal';
 
 import './Offer.scss';
 
+// XXX this isn't implemented yet so hide the button.
+const ALLOW_EXIT_OFFER = false;
```
turadg, line 16: the `previewEnabled` could be used, but probably not worth passing it down. eventually it would make a good context.

turadg (approved):

Issue #6111, opened 2022-09-01
rationalize Inter Protocol storage node paths


turadg:
## What is the Problem Being Solved?

They've developed somewhat ad-hoc. For PSM we realized there's more than one PSM (at least one per anchor) and then we realized that even IST may not be fixed: there will someday be another PSM whose mint isn't IST.

This same insight applies to most of Inter Protocol contracts. E.g. `published.vaultFactory` shouldn't assume it's the only vault factory on chain.

## Description of the Design

Namespace under IST or something similar.
Design naming scheme going out to MN-3 permissionless.

Doc for naming of keys too. Cover this case https://github.com/Agoric/agoric-sdk/pull/5393#discussion_r882178696

If keys are an index of high cardinality, they should be under their own parent key so that they don't make siblings of other types hard to find. E.g. if `manager102` sibling to `metrics` is bad because the latter will be easy to miss. `managers` sibling to `metrics` is good. 
- [x] fix published.vaultFactory.manager0
- [ ] fix published.auction.book0

## Security Considerations

First to claim gets the path.

## Test Plan

- Manually validate that the scheme supports https://github.com/Agoric/agoric-sdk/issues/6518

dckc:
See also:
 - https://github.com/Agoric/agoric-sdk/issues/4084

And there's perhaps a more general issue of documenting not just the keys but also the data at those keys.

Speaking of which...

 - #5968

dckc:
In the Jan 23 approach (#6835) to putting vbank assets in chainStorage (#5762), `agoricNames.vbankAsset` is keyed by `denom` (`ubld`, `uist`, `ibc/32432t34362354`). Nearby NameHub collections such as `agoricNames.brand` are keyed by `issuerName` (`BLD`, `IST`, `USDC_axl`), but I couldn't find a uniqueness constraint on `issuerName` in the vbank code.

I wonder if we should add a uniqueness constraint on issuerName so that we can key this NameHub on issuerName instead.

dckc:
## idea: snapshot tests for documentation

For each contract etc. that "owns" part of chainStorage, make a [snapshot test](https://github.com/avajs/ava/blob/main/docs/04-snapshot-testing.md) with example data. Then index the resulting `.md` files, perhaps in a few places. https://github.com/Agoric/agoric-sdk/tree/master/packages/inter-protocol#reading-data-off-chain is one useful place, but it doesn't cover, for example, `agoricNames`.

[...]

PR #6426, opened 2022-10-07
fix: strike "Initial" from Economic Committee

turadg:
refs: #6111 

## Description

The "Initial Economic Committee" is a misnomer because the named entity is a persistent institution, not a particular set of people. We've called what they do Economic Governance. In the code, when we change the set of people who do the economic governing we replace one "committee" with another. But the named entity "Economic Committee" endures.

There may be a better name yet than "Economic Committee" to avoid this confusion. Meanwhile we should at least strike "Initial" as what is being named is an enduring mutable body, not an actual set of members.

### Security Considerations

none

It would be good to have something like an ER diagram of:
- storage nodes
- commiteee
- electorate
- accounts
- humans

### Testing Considerations

Requires similar change to dapp-econ-gov. Since that tool is used so rarely and changes are cheap, I'll push that change once this is deployed. Alternately I can make it support both URLs until we're sure we can kill the older.

### DIFF
--- a/packages/agoric-cli/src/commands/psm.js
+++ b/packages/agoric-cli/src/commands/psm.js
@@ -369,7 +369,7 @@ export const makePsmCommand = async logger => {
       const opts = this.opts();
 
       const questionHandleCapDataStr = await vstorage.readLatest(
-        'published.committees.Initial_Economic_Committee.latestQuestion',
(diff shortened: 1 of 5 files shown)


dckc (approved):
the future is longer than the past.

dtribble (approved):
LGTM

PR #6427, opened 2022-10-08
feat(wallet/ui): show JSON in keplr signer even without Ledger

### LINKED ISSUE #6424: Show amino encoding in Keplr

`MsgProvision` and other wallet actions are showing opaque details for offers and other interactions. 

image

That makes wallet verification mostly not useful

## Description of the Design

Keplr team proposed: "You can force to use amino encoding by using `getOfflineSignerAmino()`."

If that works, then the ~json encoding will be at least human verifiable.

## Security Considerations

None.

## Test Plan

It may be that this already works with a ledger. Make sure of that.

dckc:
fixes: #6424

## Description

Rendering of our custom messages (MsgSpendAction, MsgProvision) in the keplr signer isn't great, but with a Ledger, the amino encoding meant that it's at least about as readable as JSON, as opposed to cosmos "direct" signing, where keplr displayed hex encoded protobuf gobbledygook.

It turns out that the amino encoding works fine even without a Ledger, so we might as well use it all the time to get the nicer UX.

### Security Considerations

Being able to confirm transaction details in the signer window provides a large step up in wallet security.

re-take some screenshots, @otoole-brendan ?

### Testing Considerations

tested manually:

image

### DIFF
--- a/packages/wallet/ui/src/util/keyManagement.js
+++ b/packages/wallet/ui/src/util/keyManagement.js
@@ -380,7 +380,9 @@ export const makeInteractiveSigner = async (
 
   const key = await keplr.getKey(chainId);
 
-  const offlineSigner = await keplr.getOfflineSignerAuto(chainId);
+  // Until we have SIGN_MODE_TEXTUAL,
+  // Use Amino because Direct results in ugly protobuf in the keplr UI.
+  const offlineSigner = await keplr.getOfflineSignerOnlyAmino(chainId);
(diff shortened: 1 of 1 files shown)


dckc:
@arirubinstein points out amino is a regression in transaction mutability.

dtribble (approved):
LGTM

samsiegart (commented):
Do PSM trades work as well?

Also I do not understand what "regression in transaction mutability" means, is that a blocker?

PR #6431, opened 2022-10-10
refactor: Use new M.splitArray and M.splitRecord

erights:
Replace `M.split` and `M.partial` with `M.splitArgs` and `M.splitRecord`, where the latter two have required, optional, and rest parts.

Anticipates the compression from https://github.com/Agoric/agoric-sdk/pull/6432

### DIFF
--- a/packages/SwingSet/src/typeGuards.js
+++ b/packages/SwingSet/src/typeGuards.js
@@ -3,16 +3,17 @@ import { M } from '@agoric/store';
 
 export const ManagerType = M.or('xs-worker', 'local'); // TODO: others
 
-const Bundle = M.split({ moduleType: M.string() }, M.partial({}));
+const Bundle = M.splitRecord({ moduleType: M.string() });
 
-const p1 = M.and(
-  M.partial({ creationOptions: M.partial({ critial: M.boolean() }) }),
-  M.partial({ parameters: M.recordOf(M.string(), M.any()) }),
-);
+const SwingsetConfigOptions = harden({
+  creationOptions: M.splitRecord({}, { critial: M.boolean() }),
+  parameters: M.recordOf(M.string(), M.any()),
+});
+
 const SwingSetConfigProperties = M.or(
-  M.split({ sourceSpec: M.string() }, p1),
-  M.split({ bundleSpec: M.string() }, p1),
-  M.split({ bundle: Bundle }, p1),
+  M.splitRecord({ sourceSpec: M.string() }, SwingsetConfigOptions),
+  M.splitRecord({ bundleSpec: M.string() }, SwingsetConfigOptions),
+  M.splitRecord({ bundle: Bundle }, SwingsetConfigOptions),
 );
 const SwingSetConfigDescriptor = M.recordOf(
   M.string(),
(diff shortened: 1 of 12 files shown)


erights:
Low priority, but may be easy. Ping.

packages/store/src/types.js
```diff
@@ -568,15 +568,45 @@
...
+ * is matched against the `rest` pattern.
+ * The specimen must have all the properties that appear on the `required`
+ * pattern. It may omit properties that appear on the optional pattern.
+ * For these purposes, `undefined` is like omission --- such properties
+ * pass unconditionally.
+ *
+ * @property {(required: CopyRecord<*> | CopyArray<*>,
  *             rest?: Pattern,
  * ) => Matcher} split
+ * Deprecated. Use `M.splitArgs` or `M.splitRecord` instead.
```
turadg: ideally we'd express to TypeScript that this particular property is deprecated. Unfortunately that's [not possible with JSdoc](https://stackoverflow.com/questions/29763919/jsdoc-param-together-with-deprecated). Maybe after `https://github.com/gajus/eslint-plugin-jsdoc/issues/145`. 

Meanwhile, since this file is only types, WDYT of converting it to .ts syntax so we can express the deprecation to the linter?

packages/store/src/types.js
```diff
@@ -568,15 +568,45 @@
  *             valuePatt?: Pattern,
  *             limits?: Limits
  * ) => Matcher} mapOf
- * @property {(base: CopyRecord<*> | CopyArray<*>,
+ *
+ * @property {(required: Pattern[],
+ *             optional?: Pattern[],
+ *             rest?: Pattern,
+ * ) => Matcher} splitArgs
+ * Splits an array --- typically an arguments list --- into an initial
```
turadg: Since it's any array, why call it `splitArgs`? If it's that `splitArray` is too clunky, consider `splitList`.

turadg (commented):
Removing the need for `M.partial` is an improvement.

To make git history cleaner and reviewing easier, please consider separating the new code from the refactoring. The changes in `packages/store` could land in master without any others, since they're backwards compatible.

Also for commit history none of this is a "fix". 
`feat: splitArgs and splitRecord` would describe the new functionality.
`refactor: use new splitArgs and splitRecord` would describe the changes to make use of them.

For the refactoring commit there's one more in `casting/src/netconfig.js` (added since this PR started)

erights:
> To make git history cleaner and reviewing easier, please consider separating the new code from the refactoring. The changes in packages/store could land in master without any others, since they're backwards compatible.

Good suggestion! Since this splits the PR into two, I'm putting this PR back into Draft until it is split that way.

packages/store/src/types.js
erights: Changed to `splitArray`. Not too clunky ;)

erights:
> Removing the need for `M.partial` is an improvement.
> 
> To make git history cleaner and reviewing easier, please consider separating the new code from the refactoring. The changes in `packages/store` could land in master without any others, since they're backwards compatible.

Done. With the base extracted to #6597 , both are now Ready for Review.

> Also for commit history none of this is a "fix". `feat: splitArgs and splitRecord` would describe the new functionality. `refactor: use new splitArgs and splitRecord` would describe the changes to make use of them.

Done.

> For the refactoring commit there's one more in `casting/src/netconfig.js` (added since this PR started)

Done.

packages/store/src/types.js
erights: For type-only files, I think that's plausible. Enabling active deprecation is a big plus.

But not in this PR. Resolving.

erights:
> Done. With the base extracted to #6597 , both are now Ready for Review.

#6597 is now merged, so this PR is now a candidate for merging to master.

packages/SwingSet/src/typeGuards.js
```diff
@@ -3,16 +3,17 @@ import { M } from '@agoric/store';
 
 export const ManagerType = M.or('xs-worker', 'local'); // TODO: others
 
-const Bundle = M.split({ moduleType: M.string() }, M.partial({}));
+const Bundle = M.splitRecord({ moduleType: M.string() });
+
+const partial1 = harden({
```
turadg: a name would help. consider `swingsetConfigOpts`

turadg (approved):
LGTM. Thanks for splitting.

packages/SwingSet/src/typeGuards.js
erights: Done

PR #6432, opened 2022-10-10
WIP feat: use pattern-based compression

### LINKED ISSUE #3167: ERTP: payment ledger can just store values

The `paymentLedger` in ERTP is the mapping of payments to the current amount of digital assets they hold. All of the issuer methods regarding payments use amounts (`brand` + `value`), but that does not imply that the storage itself needs to be in terms of amounts, since the brand is the same for all. Storing just the values would potentially save space and only add a few more lines of code to get the value from an amount on entry of a method and reconstitute an amount on return for each method. 

## Description of the Design

Store values in paymentLedger, use helpers to switch from amounts to values and vice versa. 

## Security Considerations

We should make sure that the brand in amounts coming from users (uncoerced amounts) is always checked and never assumed.

## Test Plan
The tests that already exist should cover this well, but targeted unit tests would be helpful too. 

## Scheduling Considerations

It would be nice to schedule before audits.

erights:
#endo-branch: markm-pattern-based-compression-2

Won't work until updated to an endo incorporating https://github.com/endojs/endo/pull/1584 

Fixes https://github.com/Agoric/agoric-sdk/issues/3167

Modified SwingSet's virtual object stores so that if they have a `stateShape`, they also use it for compression.

Modified SwingSet's virtual mapStores so that if they have a `valueShape` option, they also use it for compression.

Modified ERTP's paymentLedger to provide the `amountShape` as the `valueShape` option for the durable ledger mapStore.

Verified in the debugger that we now only store a singleton array around the value itself, dropping the enclosing amount structure and the brand which is always the same for a given ledger mapStore. IOW, instead of storing the serialization of
```js
{
   brand: moolaBrand,
   value: 2n,
}
```
for every live payment, with this PR we store the serialization of
```js
[2n]
```
for every live payment. This is not just a savings because it is smaller. We also avoid needing to adjust the refcount of their common `moolaBrand`.

Once we've switched to smallcaps, the serialization of `[2n]` is tiny.

After rebasing on #6431 , I verified in the debugger that invitations benefit additionally from InvitationElementShape. Will be nice to look at once further combined with smallcaps.

---

@dtribble @turadg @FUDCo @gibson042 @mhofman when it is ready I'll ask you to review. Until then please ignore.

### DIFF
--- a/packages/ERTP/src/paymentLedger.js
+++ b/packages/ERTP/src/paymentLedger.js
@@ -39,6 +39,7 @@ const amountShapeFromElementShape = (brand, assetKind, elementShape) => {
       if (elementShape === undefined) {
         valueShape = M.arrayOf(M.key());
       } else {
+        // M.and compresses only according to its last conjunct
         valueShape = M.arrayOf(M.and(M.key(), elementShape));
       }
       break;
--- a/packages/swingset-liveslots/src/collectionManager.js
+++ b/packages/swingset-liveslots/src/collectionManager.js
@@ -16,6 +16,8 @@ import {
   getRankCover,
   getCopyMapEntries,
   getCopySetKeys,
+  mustCompress,
+  mustDecompress,
 } from '@endo/patterns';
 import { isCopyMap, isCopySet } from '@agoric/store';
 import { makeBaseRef, parseVatSlot } from './parseVatSlots.js';
@@ -271,19 +273,24 @@ export function makeCollectionManager(
 
     const serializeValue = value => {
       const { valueShape, label } = getSchema();
-      if (valueShape !== undefined) {
-        mustMatch(value, valueShape, makeInvalidValueTypeMsg(label));
+      if (valueShape === undefined) {
+        return serialize(value);
       }
-      return serialize(value);
+      return serialize(
+        mustCompress(value, valueShape, makeInvalidValueTypeMsg(label)),
+      );
     };
 
     const unserializeValue = data => {
       const { valueShape, label } = getSchema();
-      const value = unserialize(data);
-      if (valueShape !== undefined) {
-        mustMatch(value, valueShape, makeInvalidValueTypeMsg(label));
+      if (valueShape === undefined) {
+        return unserialize(data);
       }
-      return value;
+      return mustDecompress(
+        unserialize(data),
+        valueShape,
+        makeInvalidValueTypeMsg(label),
+      );
     };
 
     function prefix(dbEntryKey) {
--- a/packages/swingset-liveslots/src/virtualObjectManager.js
+++ b/packages/swingset-liveslots/src/virtualObjectManager.js
@@ -3,7 +3,7 @@
 
 import { environmentOptionsListHas } from '@endo/env-options';
 import { assert, Fail, q, b } from '@endo/errors';
-import { assertPattern, mustMatch } from '@agoric/store';
+import { assertPattern, mustCompress, mustDecompress } from '@endo/patterns';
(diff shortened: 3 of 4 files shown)


erights:
When combined with more use of patterns, this could result in substantial savings.
For example, if we combined this PR with proposalShapes, and if we also used proposalShape for compression.

Without compression, the proposal
```js
    {
      want: {
        Winnings: {
          brand: moolaBrand,
          value: makeCopyBagFromElements([
            { foo: 'a' },
            { foo: 'b' },
            { foo: 'c' },
          ]),
        },
      },
      give: { Bid: { brand, value: 37n } },
      exit: { afterDeadline: { deadline: 11n, timer } },
    },
```
is stored with a smallcaps body of
```js
#{"exit":{"afterDeadline":{"deadline":"+11","timer":"$0.Alleged: timer"}},"give":{"Bid":{"brand":"$1.Alleged: simoleans","value":"+37"}},"want":{"Winnings":{"brand":"$2.Alleged: moola","value":{"#tag":"copyBag","payload":[[{"foo":"c"},"+1"],[{"foo":"b"},"+1"],[{"foo":"a"},"+1"]]}}}}
```

But it compresses with the proposalShape
```js
    {
      want: {
        Winnings: {
          brand: moolaBrand,
          value: M.bagOf(harden({ foo: M.string() }), 1n),
        },
      },
      give: { Bid: { brand, value: M.nat() } },
      exit: { afterDeadline: { deadline: M.gte(10n), timer } },
    }
```
to
```js
[[['c'], ['b'], ['a']], 37n, 11n]
```
whose smallcaps body is
```js
#[[["c"],["b"],["a"]],"+37","+11"]
```
which is 12% as long.

It would take much more work, but if we were able to use matching interface guards on the sending and receiving sides, we'd get similar savings for messages. https://github.com/Agoric/agoric-sdk/pull/6355 may help get there.

packages/swingset-liveslots/test/test-collection-upgrade.js
```diff
@@ -64,7 +64,10 @@ test('durable collections survive upgrade', async t => {
   t.is(ls1.testHooks.getReachableRefCount(setVref), 1);
   // thing1 should have a refcount of 4: one for valueShape, plus one
   // for each entry
-  t.is(ls1.testHooks.getReachableRefCount(thing1Vref), 4);
+  // TODO Changed the expected number from 4 to 1
+  // when switching to using patttern-based compression.
+  // Why should that make a difference?
+  t.is(ls1.testHooks.getReachableRefCount(thing1Vref), 1);
```
erights: @FUDCo , let's discuss

erights:
@michaelfig ,

the failed CI test here seems to be because
```
Error#1: 'mustCompress' is not exported by ../../node_modules/@endo/patterns/index.js, imported by ../swingset-liveslots/src/virtualObjectManager.js
```

Is this a bug in the `#endo-branch` mechanism? The stated branch of endo does export it at
https://github.com/endojs/endo/blob/5c3e8704ae9e289bcb379929e43f690f49adeb06/packages/patterns/index.js#L74

michaelfig:
> Is this a bug in the `#endo-branch` mechanism?

My intuition says that this bug would be solved by #8715 so I will invest some time there, to follow up on your earlier suggestion to do so.

erights:
See https://github.com/Agoric/agoric-sdk/issues/9232 re CI failure

michaelfig:
@erights, it looks like CI is successful now (modulo Cloudflare pages build).  I think you can continue with testing and resolving conflicts here.

PR #6438, opened 2022-10-11
fix(deployment): Pass-through SLOGSENDER_AGENT_ env

mhofman:
#6377 introduced a new `SLOGSENDER_AGENT_` prefix for setting env used by the slog senders, but failed to pass through these environment variables in the deployment image. This PR fixes that.


packages/deployment/src/main.js
```diff
@@ -473,8 +473,13 @@ show-config      display the client connection parameters
         'SWING_STORE_TRACE',
         'XSNAP_KEEP_SNAPSHOTS',
         'NODE_HEAP_SNAPSHOTS',
-      ]) {
-        if (env[envName]) {
+      ];
+      // Use a for..in loop as env objects in node have historically allowed prototype keys
```
warner, line 477: thanks for the comment, good to call out why we're ok with `for..in` rather than `for..of`

warner (approved):
looks good

PR #6442, opened 2022-10-11
fix(swingset): only preload maxVatsOnline/2 vats

### LINKED ISSUE #6433: vat preload pathologies: reboot loads unused vats, evicts the important ones

Our vat-warehouse [defaults](https://github.com/Agoric/agoric-sdk/blob/e4aa0af9a918e0bbecf208eb198c137628f5f7ba/packages/SwingSet/src/kernel/vat-warehouse.js#L60) to allowing 50 vat workers to be active at a time. When a delivery arrives for a vat that doesn't have a worker, the least-recently used worker is evicted (the process is killed), a new worker (for the targetted vat) is started from the most recent heap snapshot, that vat's recent transcript entries are replayed, and then the delivery can be submitted to the worker. This process takes anywhere from 300ms to maybe 20s, depending upon how complex the transcript actions are, and how many have accumulated since the last heap snapshot (we currently take snapshots every 1000 deliveries, so the maximum is 999 deliveries).

When a validator is restarted, it has no workers, and we'd rather consolidate the worker-bringup latency into a single big burst at validator restart time, than have it appear randomly during later blocks, so the operator can expect timely service after a one-time restart delay has finished. So we [configured](https://github.com/Agoric/agoric-sdk/blob/e4aa0af9a918e0bbecf208eb198c137628f5f7ba/packages/SwingSet/src/kernel/vat-warehouse.js#L175-L180) the vat-warehouse to preload vats at startup. This preload code simply enumerates all active vats (in numerical `vatID` order) and pages them in.

However, once the number of active vats is greater than the max-simultaneous limit, preloading more vats than the LRU cache can hold means that we're paging in `v1` only to evict it when we reach `v51`, etc. We have roughly 20 "normal" vats, so 30 votes/issues is enough to start kicking them out, and 50 votes/issues is enough to kick them all out. After the preload phase completes, we've done a bunch of useless work for low-numbered vats (they are small, so loading one only takes about 300ms, but there may be a lot of them).

Worse, the fact that single-issue voting contracts are not terminated after their decision is made (#3907) means that the high-numbered vats are leftovers that nobody will ever talk to anymore. So preload leaves us with an LRU cache full of high-numbered unused vats, which must be evicted to make way for the low-numbered actually-important vats like Zoe and the main contracts. So we must re-spend the page-in time later, when a block arrives that contains or provokes deliveries for those vats. This page-in time appears to take about 45s, which is independent of the number of leftover voting vats.

The net result is that, on a chain where N votes have happened, any validator that restarts will take an extra N*300ms to restart. When `N > 30` the first post-restart block will take more time (and may lead to a missed block), and once `N > 50` that block will take around 45s, leading to 8-10 missed blocks.

## Possible Fixes

If we could somehow persist the LRU data (a list of vatIDs ordered by their position in the cache), we could use it to re-populate the LRU in a restarted validator, and only preload the vats that were actually paged-in by the earlier incarnation. That would remove both pathologies: the important vats would be preloaded (removing the surprising on-the-next-block-with-a-send delay), and we wouldn't waste more than 30*300ms=10s of preload time on the useless vats.

The tricky part about persisting the LRU is that it non-consensus: it only tracks the workers and deliveries that were made since validator restart, and each validator restarts at random times. So we'd need to store it somewhere that is excluded from consensus. I think we have a portion of the `kvStore` for this, but I'd have to doublecheck. The persistent writes absolutely must be kept out of the `crankHash` and `activityHash` fields, or else we'd get a rapid consensus break.

Also, it would help a lot to terminate vats that are no longer useful, or avoid creating single-use vats in the first place (#3907).

Increasing the simultaneous count from 50 to something bigger would delay the onset of the problem by a little bit, but wouldn't address the magnitude significantly.

Changing the vat-warehouse preload to only load MAX_VATS, instead of all of them, might be an easy heuristic: that would prefer the lower-numbered vats, which (in our current vat configuration) would be the important ones. In the long run, that isn't nearly as good as persisting the LRU state, but it wouldn't require any state storage.

warner:
`runtimeOptions.warehousePolicy.maxVatsOnline` is an option which limits the number of xsnap workers the kernel is willing to run at the same time. It defaults to 50, which means when a message is sent to the 51st vat, some other worker (the least-recently used one) is evicted/killed (aka "paged out") to make room for the worker being started ("paged in"). This limits the amount of RAM and processes we consume from the host OS (usually Linux).

Previously, when the kernel starts, it would preload all vats, even if there were more than `maxVatsOnline`, which had the pathological behavior of 1: paging vats in only to page them back out again before ever talking to them, and 2: leaving the most recent *defined* vats online, rather than the most likely to be *used* vats.

In the Agoric PSM chain, we'll create a new vat for each governance vote, and those vote vats don't get any traffic once the governance issue is complete, but for various reasons we don't terminate them. After 30-ish votes, there will be enough dynamic vats for the early static vats (like bootstrap and zoe) to be evicted during kernel startup, which wastes time and increases latency at some unpredictable point in the future when new Zoe/contract messages arrive.

This changes the preload process to only load `maxVatsOnline / 2` vats, not all of them. It starts with the static vats (in lexicographic order, unfortunately, not numberic), and then goes through all the dynamic vats (in creation order), stopping after we hit the limit. This should avoid loading the useless done-but-not-terminated voting vats, and should leave room in the LRU queue so the next message that arrives to a still-offline worker won't evict a useful one right away.

In the long run, we might want to persist the LRU order across a reboot, so any evictions that *do* happen are less likely to hit a vat which is probably going to get paged in again soon. We might preload everything in the LRU (restoring the same set of workers that the earlier incarnation had running), or only a subset at the top (under the assumption that the bottom is unlikely to be paged in again any time soon, and it's better to save time/memory by not loading them).

closes #6433

### DIFF
--- a/packages/SwingSet/test/vat-warehouse/test-preload.js
+++ b/packages/SwingSet/test/vat-warehouse/test-preload.js
@@ -1,0 +1,126 @@
+// eslint-disable-next-line import/order
+import { test } from '../../tools/prepare-test-env-ava.js';
+
+// import * as proc from 'child_process';
+import tmp from 'tmp';
+import { makeSnapStore, makeSnapStoreIO } from '@agoric/swing-store';
+import { provideHostStorage } from '../../src/controller/hostStorage.js';
+import { initializeSwingset, makeSwingsetController } from '../../src/index.js';
+
+test('only preload maxVatsOnline vats', async t => {
+  const bpath = new URL('vat-preload-bootstrap.js', import.meta.url).pathname;
+  const tpath = new URL('vat-preload-extra.js', import.meta.url).pathname;
+  // this combination of config and initOpts means we have only two
+  // initial vats: v1-bootstrap and v2-vatAdmin
+  const config = {
+    defaultManagerType: 'xs-worker',
+    defaultReapInterval: 'never',
+    bootstrap: 'bootstrap',
+    vats: {
+      bootstrap: {
+        sourceSpec: bpath,
+      },
+    },
+    bundles: { extra: { sourceSpec: tpath } },
+  };
+  const initOpts = {
+    addComms: false,
+    addVattp: false,
+    addTimer: false,
+  };
+  const argv = [];
+
+  const snapstorePath = tmp.dirSync({ unsafeCleanup: true }).name;
+  const snapStore = makeSnapStore(snapstorePath, makeSnapStoreIO());
+  const hostStorage = { snapStore, ...provideHostStorage() };
+
+  await initializeSwingset(config, argv, hostStorage, initOpts);
+
+  /* we use vat-warehouse's built-in instrumentation, but I'll leave
+   * this alternative here in case it's ever useful
+  // this lets us snoop on which workers are still alive
+  const workers = new Map();
+  function spawn(...args) {
+    const worker = proc.spawn(...args);
+    const name = args[1][0]; // e.g. 'v1:bootstrap'
+    workers.set(name, worker);
+    return worker;
+  }
+  const isAlive = name =>
+    workers.has(name) && workers.get(name).exitCode === null;
+  */
+
+  // we'll create a 'canary' vat, and ten extras
+  const all = ['bootstrap', 'vatAdmin', 'canary'];
+  for (let count = 0; count < 10; count += 1) {
+    all.push(`extra-${count}`);
+  }
+
+  function areLiving(c, ...expected) {
+    expected = new Set(expected);
+    const living = c.getStatus().activeVats.map(info => info.options.name);
+    for (const name of all) {
+      t.is(living.indexOf(name) !== -1, expected.has(name), name);
+    }
+  }
+
+  const maxVatsOnline = 10;
+  const warehousePolicy = { maxVatsOnline };
+  const runtimeOptions = { warehousePolicy /* , spawn */ };
+
+  const c1 = await makeSwingsetController(hostStorage, null, runtimeOptions);
+  c1.pinVatRoot('bootstrap');
+  await c1.run();
+  areLiving(c1, 'bootstrap', 'vatAdmin');
+
+  // launch v3-canary
+  c1.queueToVatRoot('bootstrap', 'launchCanary', []);
+  await c1.run();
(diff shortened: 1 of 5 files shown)


Tartuffo:
A quibble, but I'd suggest adding this to the comment:

`instantiate all static vats, in lexicographic order,` _up to a limit_.

packages/SwingSet/test/vat-warehouse/test-preload.js
```diff
@@ -0,0 +1,126 @@
...
+  const snapStore = makeSnapStore(snapstorePath, makeSnapStoreIO());
+  const hostStorage = { snapStore, ...provideHostStorage() };
+
+  await initializeSwingset(config, argv, hostStorage, initOpts);
+
+  /* we use vat-warehouse's built-in instrumentation, but I'll leave
+   * this alternative here in case it's ever useful
+  // this lets us snoop on which workers are still alive
+  const workers = new Map();
+  function spawn(...args) {
```
dckc, line 43: whee! fun with ocap discipline...

packages/SwingSet/test/vat-warehouse/test-warehouse.js
```diff
@@ -10,16 +10,26 @@ import { test } from '../../tools/prepare-test-env-ava.js';
 import fs from 'fs';
 import tmp from 'tmp';
 import { initSwingStore } from '@agoric/swing-store';
-import { loadBasedir, buildVatController } from '../../src/index.js';
+import { buildVatController } from '../../src/index.js';
```
dckc, line 13: what motivated this change?

dckc (approved):
the src/ change looks good.

I'm a little unclear on why test-warehouse.js changed, but I trust it's worthwhile.

packages/SwingSet/test/vat-warehouse/test-warehouse.js
warner: I added a few new `vat-*.js` files to the directory, for the new test, but `loadBasedir()` loads everything it can find, so the old test loaded them too, which changed the vatIDs it was looking for. The rewritten form makes it insensitive to new files.

Issue #5854, opened 2022-07-29
Expose key Oracle economic data via subscription


turadg:
## What is the Problem Being Solved?

This data can be used in Blockchain Explorers and via sites like The Graph for the community to get key information about the state of the RUN Protocol components.

From [our meeting on Feb 22](https://docs.google.com/document/d/1HkNLG8Kmm9DV6XGxvZEGVWUtHli2OfkR5ph7Q77x3QE/edit#), here is the data we want to expose from the Oracle:

For each asset pair
- Aggregate price https://github.com/Agoric/agoric-sdk/issues/4652
- Source information (per oracle node operator) TBD
~Governance parameters~ there are none

dckc:
splitting this out looks like a good idea; seems like it should inherit **MUST-HAVE** while we're at it, at least until we hear from @btulloh .

turadg:
closing per @btulloh

PR #6444, opened 2022-10-11
5791 update price authority

turadg:
refs: #5791 (doesn't close until there's a CLI that builds atop https://github.com/Agoric/agoric-sdk/issues/5854 ) 

## Description

Make it possible to grant a smart wallet holder the power to push updates to a price authority. This is implemented using the continuing invitation pattern. The main change was to the offer results to include an `invitationMakers` property with the makers that can be used to offer an update. Because of the "only data or only functions" constraint of offer results, this required a breaking change to the _admin_ methods, putting them under an `admin` key.

Most of the code is implementing an integration test atop boot-psm, in `test-oracle-integration.js`.

### Security Considerations

Grants powers push a new price for a particular authority. The power is granted by distributing an invitation to a particular smart wallet address. Presently this can only be done by JS. Since we don't know targets at boot, distributing is a CORE_EVAL level permission.

### Testing Considerations

~This has a WIP for CLI commands. If that's in scope then I should get that working before merging. It's not strictly necessary for this feature, so it could be deferred. Making it work will require chain storage work because the prices don't appear (in a quick glance) to be published there, the only way a CLI could verify it works.~

Deferring ^ to another PR.


dckc:
We have a separate issue for Oracle chain storage: #5854 , so leaving that out of this PR is straightforward. OTOH, if you want to kill two birds with one stone, that's fine too.

packages/smart-wallet/test/test-oracle-integration.js
```diff
@@ -0,0 +1,203 @@
...
+ */
+const test = anyTest;
+
+const operatorAddress = 'oracleTestAddress';
+
+const makeTestSpace = async log => {
+  const psmParams = {
+    anchorAssets: [{ denom: 'ibc/usdc1234', keyword: 'AUSD' }],
+    economicCommitteeAddresses: {
+      /* empty */
```
dckc, line 35: hm... maybe an empty record should fail the shape check for `economicCommitteeAddresses`.

packages/zoe/src/contracts/priceAggregatorChainlink.js
```diff
@@ -1,798 +0,0 @@
-// @ts-check
```
dckc, line 1: I don't recall why we don't need this any more. I wonder if that disqualifies me as a reviwer.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -47,7 +49,12 @@ import '../../tools/types.js';
 const start = async (zcf, privateArgs) => {
   const { timer, POLL_INTERVAL, brandIn, brandOut, unitAmountIn } =
     zcf.getTerms();
+  assert(brandIn, 'missing brandIn');
+  assert(brandOut, 'missing brandOut');
+  assert(POLL_INTERVAL, 'missing POLL_INTERVAL');
+  assert(timer, 'missing timer');
```
dckc, line 56: My eyes are starting to expect shape checks for this sort of thing. It would let us check that brands are remotables, for example.

not critical, but I have to keep in mind that `brandIn` might be some scary JavaScript proxy monster thing that goes bump in the night as I review this...

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -343,79 +355,93 @@ const start = async (zcf, privateArgs) => {
...
+       * @returns {Promise<{admin: OracleAdmin, invitationMakers: {makePushPriceInvitation: (price: ParsableNumber) => Promise<Invitation<void>>} }>}
        */
       const offerHandler = async (
         seat,
         { notifier: oracleNotifier, scaleValueOut = 1 } = {},
       ) => {
         const admin = await creatorFacet.initOracle(oracleKey);
+        const invitationMakers = Far('invitation makers', {
+          /** @param {ParsableNumber} price */
+          makePushPriceInvitation(price) {
```
dckc, line 433: input validation here is necessary.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -343,79 +355,93 @@ const start = async (zcf, privateArgs) => {
...
+                break;
+              }
+              default: {
+                if (value && value.data !== undefined) {
+                  result = {
+                    ...value,
+                    data: makeScaledRatioFromData(value.data),
+                  };
+                } else {
+                  result = value;
```
dckc: This much logic mixed in with I/O feels off. Consider hoisting this business of scaling and dealing with things other than ratios to module level functions and adding unit tests?

dckc (changes requested):
`makePushPriceInvitation(price)` needs input validation

packages/zoe/src/contracts/priceAggregatorChainlink.js
dckc: I think I see now.

packages/zoe/src/contracts/priceAggregator.js
turadg: Good call. I hoisted it up a bit. I could see it going out farther but I don't want to invest in this module too much because @michaelfig said it was made as a strawman and the chainlink one is the real one (so I restored it).

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -513,6 +570,7 @@ const start = async (zcf, privateArgs) => {
        */
       record.querier = async timestamp => {
         // Submit the query.
+        /** @type {ParsableNumber} */
         const result = await E(oracle).query(query);
```
dckc, line 573: Do we have reason to rely on the oracle to return a `ParseableNumber` without validating it?
If not, perhaps `@type {unknown}` is more appropriate to be sure we validate it before relying on its type.

I'm struggling to keep track of the external surface of this contract. Factoring out just that part and using patterns seems like it would be nice.

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -316,11 +324,81 @@ const start = async (zcf, privateArgs) => {
...
     return multiplyRatios(unscaled, unitPriceScale);
   };
 
+  /**
+   *
+   * @param {Notifier<OraclePriceSubmission>} oracleNotifier
+   * @param {number} scaleValueOut
+   * @param {(result: Ratio) => Promise<void>} pushResult
+   */
+  const pushFromOracle = (oracleNotifier, scaleValueOut, pushResult) => {
```
dckc, line 343: is this really new code? or just moved?
maybe I should look by commit...

dckc (commented):
I don't see any critical problems this time, though I'm not sure about one case of input validation.

I'm a bit hesitant to approve, as my understanding of the contract(s) is a bit fuzzy and there are large-ish changes to it... though maybe they're just mechanical moves.

dckc (approved):
This looks like an advance in the state of the art.

packages/zoe/src/contracts/priceAggregator.js
turadg: > Do we have reason to rely on the oracle to return a ParseableNumber without validating it?

Good point. I added an assertion instead of an unchecked cast.

PR #6450, opened 2022-10-13
Convert SwingSet to use smallcaps encoding for serialized data

### LINKED ISSUE #6326: Convert kernel/liveslots to use smallcaps

The marshal package has been upgraded to use a new and improved, but currently optional, serialization format we're calling "smallcaps" (in contrast to the existing format "capdata").  Taking advantage of this requires that the kernel and liveslots be converted to use it.

## Description of the Design

In principle this is simply a matter of adding a `serializeBodyFormat: 'smallcaps'` option to the options parameter to `makeMarshal`.  However, the source code for the kernel, liveslots, and many of their various tests incorporate strings and object literals that contain pre-serialized data in the "capdata" format, which means that various things will almost certainly break.  These need to be located and revised so that everything works properly again and all the tests pass.

FUDCo:
These changes adapt SwingSet to use the new 'smallcaps' encoding format (instead of the prior 'capdata' format) for messages and stored data.

Prior to this PR, a large number of SwingSet tests would inspect and/or generate 'capdata' formatted messages and data fragments by directly constructing or examining their 'capdata' representations.  In addition, various places in the kernel would originate or transform messages by manually generating or manipulating 'capdata' formatted objects -- the kernel did not use `marshal` directly since the serializers and deserializers all closed over vat-specific reference mapping tables.  Nearly all of this has been replaced by a new kernel module `lib/kmarshal.js` that uses the actual `marshal` package to provide a set of serialization and unserialization tools that allow messages to be worked with without engaging with the details of the underlying encoding format.  (There does remain a small number of specialized bits of code which work with the underlying representation directly, mostly for purposes of validating the serialization machinery itself).  The net result of all of this is code (especially in the tests) that is considerably cleaner and simpler, that more clearly and directly expresses what it wants to do, and is significantly less brittle as well. The cost is that this PR has a very large number of individual code changes and touches a lot of files, so review will be tedious.

Note: This will not pass CI until https://github.com/endojs/endo/issues/1309 is addressed, which requires a change to Endo and/or XS. The requisite fix to Endo is https://github.com/endojs/endo/pull/1325 which is still pending release at the time this PR is being created.

Fixes #6326

### DIFF
--- a/packages/SwingSet/src/kernel/kernel.js
+++ b/packages/SwingSet/src/kernel/kernel.js
@@ -1,9 +1,9 @@
 import { assert, details as X } from '@agoric/assert';
 import { isNat } from '@agoric/nat';
 import { importBundle } from '@endo/import-bundle';
-import { stringify } from '@endo/marshal';
 import { assertKnownOptions } from '../lib/assertOptions.js';
 import { foreverPolicy } from '../lib/runPolicies.js';
+import { kser, kslot, makeError } from '../lib/kmarshal.js';
 import { makeVatManagerFactory } from './vat-loader/manager-factory.js';
 import { makeVatWarehouse } from './vat-warehouse.js';
 import makeDeviceManager from './deviceManager.js';
@@ -40,14 +40,6 @@ function abbreviateReplacer(_, arg) {
   }
   return arg;
 }
-
-function makeError(message, name = 'Error') {
-  assert.typeof(message, 'string');
-  const err = { '@qclass': 'error', name, message };
-  return harden({ body: JSON.stringify(err), slots: [] });
-}
-
-const VAT_TERMINATION_ERROR = makeError('vat terminated');
 
 /**
  * Provide the kref of a vat's root object, as if it had been exported.
@@ -266,7 +258,7 @@ export default function buildKernel(
     if (kernelKeeper.vatIsAlive(vatID)) {
       const promisesToReject = kernelKeeper.cleanupAfterTerminatedVat(vatID);
       for (const kpid of promisesToReject) {
-        resolveToError(kpid, VAT_TERMINATION_ERROR, vatID);
+        resolveToError(kpid, makeError('vat terminated'), vatID);
       }
     }
     if (critical) {
@@ -303,12 +295,13 @@ export default function buildKernel(
   function notifyMeterThreshold(meterID) {
     // tell vatAdmin that a meter has dropped below its notifyThreshold
     const { remaining } = kernelKeeper.getMeter(meterID);
-    const methargs = {
-      body: stringify(harden(['meterCrossedThreshold', [meterID, remaining]])),
-      slots: [],
-    };
     assert.typeof(vatAdminRootKref, 'string', 'vatAdminRootKref missing');
-    queueToKref(vatAdminRootKref, methargs, 'logFailure');
+    queueToKref(
+      vatAdminRootKref,
+      'meterCrossedThreshold',
+      [meterID, remaining],
+      'logFailure',
+    );
   }
 
   // TODO: instead of using a kernel-wide flag here, consider making each
@@ -358,6 +351,7 @@ export default function buildKernel(
    *  to make vat delivery) emits one of these status events if a delivery
    *  actually happened.
    *
+   * @typedef { [string, any[]] } RawMethargs
    * @typedef { {
    *    metering?: MeterConsumption | null, // delivery metering results
    *    deliveryError?: string, // delivery failed
@@ -372,7 +366,7 @@ export default function buildKernel(
    *    meterID?: string, // deduct those computrons from a meter
    *    decrementReapCount?: { vatID: VatID }, // the reap counter should decrement
    *    terminate?: { vatID: VatID, reject: boolean, info: SwingSetCapData }, // terminate vat, notify vat-admin
-   *    vatAdminMethargs?: SwingSetCapData, // methargs to notify vat-admin about create/upgrade results
+   *    vatAdminMethargs?: RawMethargs, // methargs to notify vat-admin about create/upgrade results
    * } } CrankResults
    */
 
@@ -693,13 +687,6 @@ export default function buildKernel(
     vatKeeper.setSourceAndOptions(source, options);
     vatKeeper.initializeReapCountdown(options.reapInterval);
 
-    function makeVatAdminMessage(arg, slots) {
-      return {
-        body: JSON.stringify(['newVatCallback', [vatID, arg]]),
-        slots,
-      };
-    }
-
     // createDynamicVat makes the worker, installs lockdown and
     // supervisor, but does not load the vat bundle yet. It can fail
     // if the options are bad, worker cannot be launched, lockdown or
@@ -739,9 +726,9 @@ export default function buildKernel(
     // the new vat's root object
 
     const kernelRootObjSlot = exportRootObject(kernelKeeper, vatID);
-    const arg = { rootObject: { '@qclass': 'slot', index: 0 } };
-    const slots = [kernelRootObjSlot];
-    const vatAdminMethargs = makeVatAdminMessage(arg, slots);
+    const arg = { rootObject: kslot(kernelRootObjSlot) };
+    /** @type { RawMethargs } */
+    const vatAdminMethargs = ['newVatCallback', [vatID, arg]];
     return harden({ ...startResults, vatAdminMethargs });
   }
 
@@ -831,30 +818,22 @@ export default function buildKernel(
       upgradeMessage,
       incarnationNumber: vatKeeper.getIncarnationNumber(),
     };
-    const disconnectObjectCapData = {
-      body: JSON.stringify(disconnectObject),
-      slots: [],
-    };
     /** @type { import('../types-external.js').KernelDeliveryStopVat } */
-    const kd1 = harden(['stopVat', disconnectObjectCapData]);
+    const kd1 = harden(['stopVat', kser(disconnectObject)]);
     const vd1 = vatWarehouse.kernelDeliveryToVatDelivery(vatID, kd1);
     const status1 = await deliverAndLogToVat(vatID, kd1, vd1);
 
     // make arguments for vat-vat-admin.js vatUpgradeCallback()
+    /**
+     * @param {SwingSetCapData} _errorCD
+     * @returns {RawMethargs}
+     */
     function makeFailure(_errorCD) {
-      insistCapData(_errorCD); // capdata(Error)
-      // const args = [upgradeID, false, JSON.parse(errorCD.body)];
+      insistCapData(_errorCD); // kser(Error)
+      // const error = kunser(_errorCD)
       // actually we shouldn't reveal the details, so instead we do:
-      const error = {
-        '@qclass': 'error',
-        name: 'Error',
-        message: 'vat-upgrade failure',
-      };
-      const args = [upgradeID, false, error];
-      return {
-        body: JSON.stringify(['vatUpgradeCallback', args]),
-        slots: [],
-      };
+      const error = Error('vat-upgrade failure');
+      return ['vatUpgradeCallback', [upgradeID, false, error]];
     }
 
     // We use deliveryCrankResults to parse the stopVat status.
@@ -936,10 +915,8 @@ export default function buildKernel(
     }
 
     const args = [upgradeID, true, undefined, incarnationNumber];
-    const vatAdminMethargs = {
-      body: JSON.stringify(['vatUpgradeCallback', args]),
-      slots: [],
-    };
+    /** @type {RawMethargs} */
+    const vatAdminMethargs = ['vatUpgradeCallback', args];
     const results = harden({
       computrons,
       meterID, // for the startVat
@@ -1010,7 +987,7 @@ export default function buildKernel(
     function send(targetObject) {
       const vatID = kernelKeeper.ownerOfKernelObject(targetObject);
       if (!vatID) {
-        return splat(VAT_TERMINATION_ERROR);
+        return splat(makeError('vat terminated'));
       }
       return { vatID, target: targetObject };
     }
@@ -1048,7 +1025,7 @@ export default function buildKernel(
           const deciderVat = vatWarehouse.lookup(kp.decider);
           if (!deciderVat) {
             // decider is dead
-            return splat(VAT_TERMINATION_ERROR);
+            return splat(makeError('vat terminated'));
           }
           if (deciderVat.enablePipelining) {
             return { vatID: kp.decider, target };
@@ -1291,8 +1268,8 @@ export default function buildKernel(
       // we use terminateVat() to notify vat-admin about failed vat
       // creation, but vatAdminMethargs for successful vat creation,
       // and failed/successful vat upgrades.
-      const methargs = crankResults.vatAdminMethargs;
-      queueToKref(vatAdminRootKref, methargs, 'logFailure');
+      const [method, args] = crankResults.vatAdminMethargs;
+      queueToKref(vatAdminRootKref, method, args, 'logFailure');
     }
 
     kernelKeeper.processRefcounts();
@@ -1564,7 +1541,7 @@ export default function buildKernel(
 
     await vatWarehouse.loadTestVat(vatID, setup, actualCreationOptions);
 
-    const vpCapData = { body: stringify(harden(vatParameters)), slots: [] };
+    const vpCapData = kser(vatParameters);
     /** @type { RunQueueEventStartVat } */
     const startVatMessage = {
       type: 'startVat',
@@ -1822,12 +1799,13 @@ export default function buildKernel(
     if (!kernelKeeper.hasBundle(bundleID)) {
       kernelKeeper.addBundle(bundleID, bundle);
       if (vatAdminRootKref) {
-        const methargs = harden({
-          body: JSON.stringify(['bundleInstalled', [bundleID]]),
-          slots: [],
-        });
         // TODO: consider 'panic' instead of 'logFailure'
-        queueToKref(vatAdminRootKref, methargs, 'logFailure');
+        queueToKref(
+          vatAdminRootKref,
+          'bundleInstalled',
+          [bundleID],
+          'logFailure',
+        );
       } else {
         // this should only happen during unit tests that are too lazy to
         // build a complete kernel: test/bundles/test-bundles-kernel.js
--- a/packages/SwingSet/src/kernel/vat-admin-hooks.js
+++ b/packages/SwingSet/src/kernel/vat-admin-hooks.js
@@ -1,24 +1,18 @@
 import { assert } from '@agoric/assert';
-import { stringify, parse } from '@endo/marshal';
 import { insistVatID } from '../lib/id.js';
+import { kser, kunser } from '../lib/kmarshal.js';
 
 export function makeVatAdminHooks(tools) {
   const { kernelKeeper, terminateVat } = tools;
   return {
     createByBundle(argsCapData) {
       // first, split off vatParameters
-      const argsJSON = JSON.parse(argsCapData.body);
-      const [bundle, { vatParameters: vpJSON, ...dynamicOptionsJSON }] =
-        argsJSON;
+      const args = kunser(argsCapData);
+      const [bundle, { vatParameters, ...dynamicOptions }] = args;
       // assemble the vatParameters capdata
-      const vatParameters = {
-        body: JSON.stringify(vpJSON),
-        slots: argsCapData.slots,
-      };
-      // then re-parse the rest with marshal
-      const dynamicOptions = parse(JSON.stringify(dynamicOptionsJSON));
+      const marshalledVatParameters = kser(vatParameters);
       // incref slots while create-vat is on run-queue
-      for (const kref of vatParameters.slots) {
+      for (const kref of marshalledVatParameters.slots) {
         kernelKeeper.incrementRefCount(kref, 'create-vat-event');
       }
       const source = { bundle };
@@ -27,38 +21,29 @@ export function makeVatAdminHooks(tools) {
         type: 'create-vat',
         vatID,
         source,
-        vatParameters,
+        vatParameters: marshalledVatParameters,
         dynamicOptions,
       };
       kernelKeeper.addToAcceptanceQueue(harden(event));
       // the device gets the new vatID immediately, and will be notified
       // later when it is created and a root object is available
-      const vatIDCapData = { body: JSON.stringify(vatID), slots: [] };
-      return harden(vatIDCapData);
+      return harden(kser(vatID));
     },
 
     createByID(argsCapData) {
-      // argsCapData is marshal([bundleID, options]), and options is {
-      // vatParameters, ...rest }, and 'rest' is JSON-serializable (no
-      // slots or bigints or undefined). We get the intermediate marshal
-      // representation (with @qclass nodes), carve off vatParameters,
-      // then reassemble the rest. All slots will be associated with
-      // vatParameters.
+      // `argsCapData` is marshal([bundleID, options]), and `options` is {
+      // vatParameters, ...rest }, and `rest` is checked by vat-vat-admin.js to
+      // contain only known keys and types, none of which allow slots.  So any
+      // slots in `argsCapData` will be associated with `vatParameters`.
 
       // first, split off vatParameters
-      const argsJSON = JSON.parse(argsCapData.body);
-      const [bundleID, { vatParameters: vpJSON, ...dynamicOptionsJSON }] =
-        argsJSON;
+      const args = kunser(argsCapData);
+      const [bundleID, { vatParameters, ...dynamicOptions }] = args;
       assert(kernelKeeper.hasBundle(bundleID), bundleID);
-      // assemble the vatParameters capdata
-      const vatParameters = {
-        body: JSON.stringify(vpJSON),
-        slots: argsCapData.slots,
-      };
-      // then re-parse the rest with marshal
-      const dynamicOptions = parse(JSON.stringify(dynamicOptionsJSON));
+      // assemble the marshalled vatParameters
+      const marshalledVatParameters = kser(vatParameters);
       // incref slots while create-vat is on run-queue
-      for (const kref of vatParameters.slots) {
+      for (const kref of marshalledVatParameters.slots) {
         kernelKeeper.incrementRefCount(kref, 'create-vat-event');
       }
       const source = { bundleID };
@@ -67,25 +52,24 @@ export function makeVatAdminHooks(tools) {
         type: 'create-vat',
         vatID,
         source,
-        vatParameters,
+        vatParameters: marshalledVatParameters,
         dynamicOptions,
       };
       kernelKeeper.addToAcceptanceQueue(harden(event));
       // the device gets the new vatID immediately, and will be notified
       // later when it is created and a root object is available
-      const vatIDCapData = { body: JSON.stringify(vatID), slots: [] };
-      return harden(vatIDCapData);
+      return harden(kser(vatID));
     },
 
     upgrade(argsCapData) {
       // marshal([vatID, bundleID, vatParameters]) -> upgradeID
-      const argsJSON = JSON.parse(argsCapData.body);
-      const [vatID, bundleID, vpJSON, upgradeMessage] = argsJSON;
+      const args = kunser(argsCapData);
+      const [vatID, bundleID, vatParameters, upgradeMessage] = args;
       insistVatID(vatID);
       assert.typeof(bundleID, 'string');
       assert.typeof(upgradeMessage, 'string');
-      const vpCD = { body: JSON.stringify(vpJSON), slots: argsCapData.slots };
-      for (const kref of vpCD.slots) {
+      const marshalledVatParameters = kser(vatParameters);
+      for (const kref of marshalledVatParameters.slots) {
         kernelKeeper.incrementRefCount(kref, 'upgrade-vat-event');
       }
       const upgradeID = kernelKeeper.allocateUpgradeID();
@@ -94,34 +78,33 @@ export function makeVatAdminHooks(tools) {
         vatID,
         upgradeID,
         bundleID,
-        vatParameters: vpCD,
+        vatParameters: marshalledVatParameters,
         upgradeMessage,
       };
       kernelKeeper.addToAcceptanceQueue(harden(ev));
-      const upgradeIDCD = { body: JSON.stringify(upgradeID), slots: [] };
-      return harden(upgradeIDCD);
+      return harden(kser(upgradeID));
     },
 
-    terminate(argsCD) {
+    terminate(argsCapData) {
       // marshal([vatID, reason]) -> null
-      const argsJSON = JSON.parse(argsCD.body);
-      const [vatID, reasonJSON] = argsJSON;
+      const args = kunser(argsCapData);
+      const [vatID, reason] = args;
       insistVatID(vatID);
-      const reasonCD = { ...argsCD, body: JSON.stringify(reasonJSON) };
+      const marshalledReason = kser(reason);
       // we don't need to incrementRefCount because if terminateVat sends
       // 'reason' to vat-admin, it uses notifyTermination / queueToKref /
       // doSend, and doSend() does its own incref
-      void terminateVat(vatID, true, reasonCD);
+      void terminateVat(vatID, true, marshalledReason);
       // TODO: terminateVat is async, result doesn't fire until worker
       // is dead. To fix this we'll probably need to move termination
       // to a run-queue ['terminate-vat', vatID] event, like createVat
-      return harden({ body: stringify(undefined), slots: [] });
+      return harden(kser(undefined));
     },
 
     changeOptions(argsCapData) {
       // marshal([vatID, options]) -> null
       assert(argsCapData.slots.length === 0);
-      const args = JSON.parse(argsCapData.body);
+      const args = kunser(argsCapData);
       const [vatID, options] = args;
       insistVatID(vatID);
       const ev = {
@@ -130,7 +113,7 @@ export function makeVatAdminHooks(tools) {
         options,
       };
       kernelKeeper.addToAcceptanceQueue(harden(ev));
-      return harden({ body: stringify(undefined), slots: [] });
+      return harden(kser(undefined));
     },
   };
 }
--- a/packages/SwingSet/src/lib/capdata.js
+++ b/packages/SwingSet/src/lib/capdata.js
@@ -1,4 +1,6 @@
 import { assert, details as X } from '@agoric/assert';
+import { passStyleOf } from '@endo/marshal';
+import { kunser, krefOf } from './kmarshal.js';
 
 /* eslint-disable jsdoc/require-returns-check */
 /**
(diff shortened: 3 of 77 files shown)


FUDCo:
As an experiment, I tried switching back to `capdata` encoding and reran all the SwingSet tests.  6 tests failed, all of which contained hard-coded `smallcaps` representations (for Reasons), which we would expect to fail under these circumstances.  However, all the other tests pass, so it appears that the strategy for isolating from the marshal encoding format is a success.

packages/SwingSet/src/kernel/vat-admin-hooks.js
```diff
@@ -27,38 +21,30 @@ export function makeVatAdminHooks(tools) {
...
       // argsCapData is marshal([bundleID, options]), and options is {
-      // vatParameters, ...rest }, and 'rest' is JSON-serializable (no
-      // slots or bigints or undefined). We get the intermediate marshal
-      // representation (with @qclass nodes), carve off vatParameters,
-      // then reassemble the rest. All slots will be associated with
-      // vatParameters.
+      // vatParameters, ...rest }, and 'rest' is JSON-serializable (no slots or
+      // bigints or undefined). We get the intermediate marshal representation,
+      // carve off vatParameters, then reassemble the rest. All slots will be
+      // associated with vatParameters.
```
erights: Now that you're using kunser on it instead of JSON.parse, is this comment still appropriate?

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -72,6 +72,12 @@ function matchAny(patt) {
 }
 
 function throwNotDurable(value, slotIndex, serializedValue) {
+  assert.fail(
+    // prettier-ignore
```
erights, line 74: I didn't know we could do that.
TIL

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -15,7 +15,7 @@ import {
   makeCopyMap,
 } from '@agoric/store';
 import { Far, passStyleOf } from '@endo/marshal';
-import { decodeToJustin } from '@endo/marshal/src/marshal-justin.js';
+// import { decodeToJustin } from '@endo/marshal/src/marshal-justin.js';
```
erights: You can just delete this line. There's no reason to go out of our way to preserve a Justin rendering here specifically.

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -72,6 +72,12 @@ function matchAny(patt) {
 }
 
 function throwNotDurable(value, slotIndex, serializedValue) {
+  assert.fail(
+    // prettier-ignore
+    X`value is not durable: ${value} at slot ${q(slotIndex)} of ${serializedValue.body}`,
+  );
+  /*
+  // Currently disabled because decodeToJustin does not yet support smallcaps encoding
   const body = JSON.parse(serializedValue.body);
```
erights: Likewise here. Just delete. Nothing precious here.

erights (approved):
LGTM!

packages/SwingSet/src/kernel/vat-admin-hooks.js
FUDCo: As it turns out, yes.

packages/SwingSet/src/lib/capdata.js
```diff
@@ -30,15 +31,11 @@ export function insistCapData(capdata) {
...
-    typeof body === 'object' &&
-    body['@qclass'] === 'slot' &&
-    body.index === 0
-  ) {
-    if (data.slots.length === 1) {
-      return data.slots[0];
+  if (data.slots.length === 1) {
+    const encValue = kunser(data);
+    const slotValue = data.slots[0];
+    if (krefOf(encValue) === slotValue) {
```
warner: How hard would it be to add a `passStyleOf(encValue) === 'remotable'` or whatever to that check? I can't think of a `{ body: 'ko12', slots: ['ko12'] }` -type attack that would actually survive the `kunser()` call, and of course this cannot return any kref that wasn't already in `slotValue` (and, since callers will always translate vrefs into krefs through a c-list before calling `extractSingleSlot`, then `slotValue` must have been available to the capdata emitter), but I'm still a little nervous that this function is less strict than the one it replaces, against maliciously-malformed capdata.

packages/SwingSet/src/lib/kmarshal.js
```diff
@@ -0,0 +1,55 @@
...
+const promiseMap = new WeakMap();
+
+export const kslot = (kref, iface) => {
+  if (iface && iface.startsWith('Alleged: ')) {
+    // Encoder prepends "Alleged: " to iface string, but the decoder doesn't strip it
+    // Unclear whether it's the decoder or me who is wrong
+    iface = iface.slice(9);
+  }
+  if (
+    kref.startsWith('p') ||
```
warner, line 20: Accepting `p` means this isn't really limited to kernel-side refs. For `lp` and `rp` (which are used in the comms protocol) that's not quite as big a deal, because they're the same length as the kernel-side `kp` prefix, but `p` is shorter, and introduces a prefix-overlap hazard if we add new categories in the future (like, I dunno, "peripheral" refs / 'p-refs', with prefixes like `poNN` and `ppNN`).

Do we use this `kunser` against vref- or comms- flavored refs? Could we safely remove the `p` case? A verbose but slightly safer option would be to expand that to an `||` of both `p+` and `p-`.

packages/SwingSet/src/lib/kmarshal.js
```diff
@@ -0,0 +1,55 @@
...
+
+// Simple wrapper for serializing and unserializing marshalled values inside the
+// kernel, where we don't actually want to use clists nor actually allocate real
+// objects, but instead to stay entirely within the domain of krefs.  This is
+// used to enable syntactic manipulation of serialized values while remaining
+// agnostic about the internal details of the serialization encoding.
+
+const promiseMap = new WeakMap();
+
+export const kslot = (kref, iface) => {
```
warner, line 12: This is being used as the `convertSlotToVal` function, right? That wasn't obvious from the places where I saw it imported. What do you think about renaming it to `convertKslot`?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -42,14 +42,11 @@ function abbreviateReplacer(_, arg) {
   return arg;
 }
 
-function makeError(message, name = 'Error') {
+function makeError(message) {
   assert.typeof(message, 'string');
-  const err = { '@qclass': 'error', name, message };
-  return harden({ body: JSON.stringify(err), slots: [] });
+  return kser(Error(message));
```
warner: I like the simplicity.. what happens with stack traces when we use a real Error? I think this will be the first time `marshal.serialize` gets used from the kernel bundle, and I don't know what it will do with Errors and the whole censoring-details thing.

Oh, also, we need to make sure the serialized form doesn't have a sequence number or a "check the logs for more details about error#NNNN", because that would vary depending upon how many times `makeError` has been called within this launch of the kernel, which is way outside consensus.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -832,30 +831,22 @@ export default function buildKernel(
...
-        '@qclass': 'error',
-        name: 'Error',
-        message: 'vat-upgrade failure',
-      };
-      const args = [upgradeID, false, error];
-      return {
-        body: JSON.stringify(['vatUpgradeCallback', args]),
-        slots: [],
-      };
+      const error = kser(Error('vat-upgrade failure'));
```
warner: Something feels wrong here.. if `vatAdminMethargs` gets unpacked into `[method, args]` and passed to a `queueToKref()` later, and you've changed `queueToKref()` to take an object graph (with the special Far markers for `koid`s, and the special Promise-like markers for `kpid`s), then calling `kser()` here feels like it's going to double-serialize the `Error('vat-upgrade failure')`.

I'm wondering if our failed-upgrade test looks carefully at the error object it gets, or if it's content with just *any* kind of failure, ignoring the error object.

packages/SwingSet/src/kernel/vat-admin-hooks.js
warner: Hrm, we no longer interact with the "intermediate marshal representation" though, right?

Thinking it through, the original was more tolerant than the comment suggested. If `...rest` (e.g. `dynamicOptions`) contained a serialized BigInt, then `argsCapData.body` would have had a `{"@qclass": "bigint",..}` in it (as a string), `argsJSON` and `dynamicOptionsJSON` would have the object with a `.["@qclass"]` property, `JSON.stringify(dynamicOptionsJSON)` gets the string, `dynamicOptions = parse(JSON.stringify(dynamicOptionsJSON))` would then wind up with a real `BigInt` (because `parse` is really marshal's `parse`). It would not tolerate slots in `dynamicOptions` (those got eaten by / assigned to `vatParameters`), but if `vat-vat-admin.js` didn't prohibit it, `dynamicOptions` could include arbitrary marshallable data.

The new version is just as tolerant, but avoids touching the intermediate objects-with-`@qclass` state, because it's using placeholder `Far` objects instead.

Actually it's more tolerant: if `vat-vat-admin.js` allowed it, `dynamicOptions` could include slots, and downstream recipients of these options (e.g. vat-loader.js) would receive the `Far` marker/placeholder objects created by `kunser`, which it probably wouldn't know what to do with (also we wouldn't incref the kref, because the code below only does that for slots in `vatParameters`). I don't think that represents an attack opportunity, both because vat-vat-admin.js is doing sanitization, `kunser()` is creating pretty empty Far objects, and the code in `vat-loader.js` and friends are expecting strings and numbers, so it's not like the userspace attacker could get any more control over `.toString()` than they could have by sending, you know, a `String`.

For accuracy, how about replacing that comment with:

> argsCapData is marshal([bundleID, options]), and options is { vatParameters, ...rest },
> and 'rest' is checked by vat-vat-admin.js to contain only known keys and types, none of
> which allow slots. So any slots in `argsCapData` will be associated with `vatParameters`.

packages/SwingSet/src/lib/kdebug.js
```diff
@@ -92,20 +139,32 @@ export function legibilizeMethod(method) {
 
 export function extractMethod(methargsCapdata) {
   try {
-    const methargs = JSON.parse(methargsCapdata.body);
-    return legibilizeMethod(methargs[0]);
+    let smallcaps = false;
+    let bodyString = methargsCapdata.body;
+    if (bodyString.charAt(0) === '#') {
```
warner, line 144: This looks like a heuristic for determining the encoding format. Is it the same one that the real `marshal` uses?

If this tries to legiblize a badly-formatted methargs (say, `##`), perhaps delivered from a remote system via comms, could this crash the legibilizer, and thus the kernel? I see a call to `kdebug(legibilizeMessage(message))` in `processDeliveryMessage`, and even though `kdebug()` is disabled, its arguments will still get evaluated.

packages/SwingSet/src/vats/comms/controller.js
```diff
@@ -29,41 +25,25 @@ export function deliverToController(
...
 
     const name = args[0];
-    assert.typeof(name, 'string', X`bad addRemote name ${name}`);
-    (args[1]['@qclass'] === 'slot' && args[1].index === 0) ||
-      assert.fail(X`unexpected args for addRemote(): ${methargs.body}`);
-    (args[2]['@qclass'] === 'slot' && args[2].index === 1) ||
-      assert.fail(X`unexpected args for addRemote(): ${methargs.body}`);
-    const transmitterID = slots[args[1].index];
-    const setReceiverID = slots[args[2].index];
+    const transmitterID = `${args[1]}`;
```
warner: Eek! If someone calls `E(vats.comms).addRemote('name', 'o-12', 'o-13')`, they can forge references to transmitters or `setReceiver` objects to which they didn't have access. The old code looked for `@qclass` of `slot` first, and only looked at `.slots` if that passed. This one assumes the object is one of the Far placeholders, without performing the equivalent check.

I believe using `.toString()` on the placeholder objects is a dangerous temptation. Instead, I think `kunser` needs a way for callers to ask whether the placeholder object they're looking at (`args[1]` here, the alleged Presence-ish) is really Far, or whether it's just an attacker-controlled string.

`krefOf()` is nearly perfect for that, especially if it throws if the object is not known to the unserializer (instead of merely returning `undefined`), but only if it used a WeakMap for the object-id placeholders (like it does with Promises), not `.toString()`-bearing objects. It would be a proper inverse of `kslot()`. Then this code would use `transmitterID = krefOf(args[1])` instead of the quasiliteral.

I'm not sure if it's possible, given the new specificity of marshal, but it'd be safer (fewer temptations) if the Far / Presence-ish objects that `kunser` created were actually just hardened empty objects, and the krefs were only visible through that WeakMap, therefore only visible through a call to `krefOf()`. That way, there'd be absolutely no way for an attacker-controlled string to masquerade as a kref of their choosing, because krefs would only come from that new function, and the WeakMap is only populated by kunser itself.

packages/SwingSet/src/lib/kmarshal.js
```diff
@@ -0,0 +1,55 @@
...
+    // what the problem was thought to have been, to see if it is real and to
+    // understand it if so.  This will eventually result in either changes to
+    // the smallcaps encoding or to the marshal setup API to support the purely
+    // manipulative use case.  In the meantime, this ugliness...
+    const p = new Promise(() => undefined);
+    promiseMap.set(p, kref);
+    return harden(p);
+  } else {
+    return Far(iface, {
+      toString: () => `${kref}`,
```
warner: As described in the comment on vats/comms/controller.js, this is a dangerous temptation: code which thinks it has an alleged Presence-analogue (and uses a quasiliteral to "extract" the kref) can be tricked by a sender who merely provides a string instead of a real object reference.

It would be safer to:
* not add `toString`
  * or at least make it something that won't parse as a kref, like `placeholder for ${kref}`
* add a WeakMap that maps from this placeholder to the real kref
* read this map from `krefOf` in the object case
  * probably throwing an error if it's not in either map

So then `krefOf` is a proper inverse for `kslot`.

After making this change, we should audit all the new calls to `kunser()` (I see boatloads in tests, but the ones in vats/comms/ are the most critical) to ensure they're using `krefOf(allegedPresenceAnalogue)` rather than just a quasiliteral `${allegedPresenceAnalogue}`.

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -101,49 +69,44 @@ const testUpgrade = async (
...
   // the upgraded vat. We'll determine their krefs and later query the
   // upgraded vat to see if it's still importing them or not
-  const [impstatus, impcapdata] = await run('getImportSensors', []);
+  const [impstatus /* , impresult */] = await run('getImportSensors', []);
   t.is(impstatus, 'fulfilled');
-  const impKrefs = ['skip0'];
-  for (let i = 1; i < NUM_SENSORS + 1; i += 1) {
-    impKrefs.push(getImportSensorKref(impcapdata, i)[1]);
-  }
+  // const impKrefs = kunser(impresult);
```
warner: I know the test code that references these got commented out, but I'd love it if we could shorten the eventual path to re-enabling it, by not deleting the loop that populates `impKrefs`. I think it just needs something like `impKrefs = ['skip0', ...impresult.map(krefOf)]`. (We need the actual kref strings below, in the commented-out part, not just the placeholders, because we're querying c-lists and kernel refcount tables for them).

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -101,49 +69,44 @@ const testUpgrade = async (
...
+  const v1p2Kref = krefOf(v1result.p2);
 
   // grab exports to deduce durable/virtual vrefs
-  t.is(getRetained(v1capdata, 'dur1')[0], 'slot');
-  const dur1Kref = getRetained(v1capdata, 'dur1')[1];
-  t.is(getRetained(v1capdata, 'vir2')[0], 'slot');
-  const vir2Kref = getRetained(v1capdata, 'vir2')[1];
-  const vir5Kref = getRetained(v1capdata, 'vir5')[1];
-  const vir7Kref = getRetained(v1capdata, 'vir7')[1];
+  const dur1Kref = krefOf(v1result.retain.dur1);
```
warner, line 111: If `krefOf` throws when it sees a non-slot/placeholder, this will be just as strict as the removed `t.is(getRetained.., 'slot')`.

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -101,49 +69,44 @@ const testUpgrade = async (
...
+  const [v1status, v1resultEnc] = await run('buildV1', []);
+  const v1result = kunser(v1resultEnc);
   t.is(v1status, 'fulfilled');
-  t.deepEqual(get(v1capdata, 'version'), 'v1');
-  t.deepEqual(get(v1capdata, 'youAre'), 'v1');
-  t.deepEqual(get(v1capdata, 'marker'), ['slot', markerKref]);
-  t.deepEqual(get(v1capdata, 'data'), ['some', 'data']);
+  t.is(v1result.version, 'v1');
+  t.is(v1result.youAre, 'v1');
+  t.is(`${v1result.marker}`, `${marker}`);
```
warner: this is another test that could be tricked by a string with the same value as the kref.. not that our test program will be attacking us, but it should probably use `krefOf` instead

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -692,16 +649,15 @@ test('failed vatAdmin upgrade - bad replacement code', async t => {
   const kpid = c.upgradeStaticVat('vatAdmin', true, bundleID, {});
   await c.run();
   const vaUpgradeStatus = c.kpStatus(kpid);
-  const vaUpgradeCapdata = c.kpResolution(kpid);
+  const vaUpgradeResult = kunser(kunser(c.kpResolution(kpid)));
```
warner: why did this (and the two other instances in this file) require double unserialization?

packages/SwingSet/test/virtualObjects/collection-slots/test-collection-slots.js
```diff
@@ -10,17 +10,14 @@ import {
...
 }
 
 function getImportSensorKref(impcapdata, i) {
-  const body = JSON.parse(impcapdata.body);
-  const value = body[i];
-  if (typeof value === 'object' && value['@qclass'] === 'slot') {
-    return ['slot', impcapdata.slots[value.index]];
-  }
-  return value;
+  return ['slot', `${kunser(impcapdata)[i]}`];
```
warner: another place where `krefOf` would be safer

And if `krefOf` throws instead of returning `undefined`, then `getImportSensorKref` could be change to return just the kref, not `['slot', kref]`, and the callers elsewhere in this file could be simpler.

packages/SwingSet/test/virtualObjects/delete-stored-vo/test-delete-stored-vo.js
```diff
@@ -9,18 +9,14 @@ import {
...
 }
 
 function getImportSensorKref(impcapdata, i) {
-  const body = JSON.parse(impcapdata.body);
-  const value = body[i];
-  if (typeof value === 'object' && value['@qclass'] === 'slot') {
-    return ['slot', impcapdata.slots[value.index]];
-  }
-  return value;
+  return ['slot', `${kunser(impcapdata)[i]}`];
```
warner: ditto

packages/SwingSet/test/virtualObjects/test-virtualObjectGC.js
```diff
@@ -526,7 +466,8 @@ function validateDropStored(v, rp, what, esp, postCheck) {
   validateDone(v);
 }
 
-function validateWeakCheck(v, what) {
+function validateWeakCheck(v, vo) {
+  const what = `${vo}`;
```
warner: another job for `krefOf` I think.. several in this file

packages/SwingSet/test/virtualObjects/test-virtualObjectGC.js
```diff
@@ -150,8 +150,8 @@ function base(vref) {
   }
 }
 
-function esKey(vref) {
-  return `vom.es.${base(vref)}`;
+function esKey(what, slot = 0) {
```
warner, line 153: I can't quite tell why these helper functions are better with a serialization-then-extract-slots call, rather than just sticking to vrefs. It certainly does cause a lot of churn. What happens if you start by changing just `thingSer()`/`holderSer()`/`thingArg()`, and leave `esKey`/`rcKey`/`stateKey` alone.. is there less churn?

packages/SwingSet/test/virtualObjects/vdata-promises/test-vdata-promises.js
```diff
@@ -90,12 +88,9 @@ test('result promises in vdata', async t => {
...
   // console.log(res1); // [p4, p5, p6]
-  t.deepEqual(JSON.parse(res1.body), [
-    { '@qclass': 'slot', index: 0 },
-    { '@qclass': 'slot', index: 1 },
-    { '@qclass': 'slot', index: 2 },
-  ]);
-  const [p4kref, p5kref, p6kref] = res1.slots;
+  const res1Val = kunser(res1);
+  t.is(res1Val.length, 3);
+  const [p4kref, p5kref, p6kref] = res1Val;
```
warner, line 93: I think these would be better as `res1Val.map(krefOf)`, but if you don't want to go that way, then we should call the variables something other than `kref`, since they're now placeholders/markers/Far/Presence-analogues, not kref strings.

packages/SwingSet/test/virtualObjects/vdata-promises/test-vdata-promises.js
```diff
@@ -113,9 +109,9 @@ test('result promises in vdata', async t => {
     // returns undefined, or { vatSlot, isReachable }
     return s && parseReachableAndVatSlot(s);
   }
-  t.falsy(has(p4kref));
-  t.falsy(has(p5kref));
-  t.falsy(has(p6kref));
+  t.falsy(has(`${p4kref}`));
```
warner, line 112: more places for `krefOf`, but having `p4kref` be a string kref would be better

packages/SwingSet/src/lib/kmarshal.js
warner: On second thought, seeing how this is used to build placeholders (e.g. when building the bootstrap message), I'm not sure `convertKslot` is ideal. I don't know what to recommend, so I guess I'd recommend leaving it as is.

packages/SwingSet/src/kernel/kernel.js
warner: oh, this might explain why you have the double-unserialize in some of the tests, in which case I think both sides need to remove a serialization layer

warner (changes requested):
Ok, two main things:
* unsafe `.toString()` pattern on object-slot Presence-ish placeholders
* need to check that `kser(Error())` from within the kernel doesn't have surprising non-deterministic properties like an error number (meant for correlation with non-redacted details in logs)

plus some more cosmetic things

packages/SwingSet/src/lib/capdata.js
FUDCo: Seems like gilding the lily, since nobody is in a position to inject malformed capdata here.  Moreover, my read is that the new code is slightly stricter than what it replaces (due to the `krefOf(encValue) === slotValue` check), so I'm not sure what test you fear is being missed now that was present previously.

packages/SwingSet/src/lib/kmarshal.js
FUDCo: There's a weakness in how the functions in this module are named (i.e., they're mostly `kshortthing`), because there _are_ uses of these functions (in test code and comms code) that do involve vrefs and comms refs.  I'm very open to better names (in particular, the main reason I haven't already ditched `kslot`, a name that's arguably terrible in multiple ways, is that it is very weildy, more so than any of the competing alternatives I've considered).

However, I don't think the particular paranoia of explicitly checking for `p+` and `p-` buys us anything, since the `p` check is entirely sufficient for the current ref encodings and if we ever change or extend them we'll have to overhaul this function anyway.

packages/SwingSet/src/kernel/kernel.js
mhofman: Workaround the error id numbering issue: explicitly call `makeMarshal` with `{ errorIdNum: NaN }`. It only ever does `errorIdNum += 1;`, which is stable for `NaN` and `Infinity`

packages/SwingSet/src/lib/kmarshal.js
FUDCo: It is used as the `convertSlotToVal` function for the kernel marshaller, but that's one use out of more than 300.  As I said above, a better name would be good, but I haven't yet thought of one.  The main thing is that the name needs to be short since this is used so so much.

packages/SwingSet/src/kernel/kernel.js
FUDCo: It's a reasonable thing to be concerned about, but my take is that it's exactly the marshaling that's used going into and out of vats, so if smallcaps does that OK before the kernel marshaler became part of the story then this should be OK too.  Note that games the kernel marshaler plays with remotables don't come into play when we're dealing with Errors.

mhofman: Actually, it looks like that's what `errorTagging: 'off'` is for.

warner: I hear `errorTagging` is an option which might be useful.

FUDCo: Super!  I'll configure the kernel marshaler with error tagging off.

Removing a value-subtracted feature is a value add, right?

packages/SwingSet/src/kernel/kernel.js
FUDCo: I'm not sure.  There are places where double serialization is correct, when a capdata is part of some other message's payload.

FUDCo: I've reviewed the various calls to `makeError` and verified that the output is used in a place that's expecting capdata.  I also added a test to check that the output from `makeError` and the output from `kser(Error(stuff))` is what things expect it to be; the later test will fail if the encoding changes and thereby flag when things need to be reinspected.

packages/SwingSet/src/lib/kdebug.js
FUDCo: Yes, it's the same heuristic that marshal uses.  Note that this function is used for internal diagnostic and debug logging purposes, and any message that makes it to this point has already been vetted several times over for being kosher.

erights: Warning: I'm about to be pedantic again ;)

It is not a heuristic, because it is not just usually true. It is a rule.

FUDCo: I'll allow it.

packages/SwingSet/src/vats/comms/controller.js
FUDCo: Anyone who can send traffic directly to the comms vat's control channel is already in a position to arbitrarily mess with the comms protocol in any number of ways.  But normal vats can't do this.

Moreover, I don't follow the threat model you are positing with respect to extracting the kref from the object.  The only way anything at all can ever end up in the slots array is if `kunser` put it there, in which case it's guaranteed to be well a behaved ersatz presence.

packages/SwingSet/src/lib/kmarshal.js
FUDCo: The fact that slots might contain promises suggests that perhaps `krefOf` should be used everywhere within the kernel rather than the quasiliteral construct, since it's possible that messages might contain promises in unexpected places.  On the other hand, 99% of the uses of that construct are in tests, where it's just not an issue.  However, `krefOf` in turn getting the kref from a presence by using `toString` is fine because, as I said above, the only way these get generated is if they are the output from `kunser` or if they are hand composed (using `kslot`) inside the kernel code itself.  There exists no place where an attacker could position themselves for this to be a problem.

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: The notational convenience of the template string form is enormous, and these are tests.

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: Because the payload of a resolve, which is what the inner unserialization extracts, is itself a capdata.  We always had this double unserialization; the improved coding idioms just make it obvious.

FUDCo:
@warner and I discussed the question of having ersatz presences be self-serializing via `toString()` vs. putting them in the weak map that's used for promises.  The first approach has the problem that shenanigans could potentially sneak in via serialization.  The second approach has the problem that using it results in compartmentalized module scopes breaking a bunch of our tests in terrible ways.  We settled on a third approach, namely always requiring the use of `krefOf` to extract the kref (rather than using a template literal), then attaching a `getKref` method to the ersatz presence (rather than using `toString` for this purpose).  This allows the ersatz presences to be self-identifying but not allowing accidental bogus serialization.  This has been implemented in the latest commit (which will be squashed in once the PR is approved).

erights:
> The first approach has the problem that shenanigans could potentially sneak in via serialization. 

If this is still a problem for promises, how does having it be the same problem for presences make anything worse? I would think that either the problem is fatal for both, or the problem is something we can live with for both.

FUDCo:
> If this is still a problem for promises, how does having it be the same problem for presences make anything worse? I would think that either the problem is fatal for both, or the problem is something we can live with for both.

The problem comes up in the context of dropping messages in from the outside with `controller.queueToVatRoot`, which is not allowed to send promises.

packages/SwingSet/src/lib/capdata.js
warner: > Seems like gilding the lily, since nobody is in a position to inject malformed capdata here.

The kernel [uses](https://github.com/Agoric/agoric-sdk/blob/68ecc09f4f57ed833acbd4dcc6de6140fae3f9e7/packages/SwingSet/src/kernel/kernel.js#L1030) this `extractSingleSlot()` when deciding how to route a delivery to a resolved promise, which means that the vat that did the `syscall.resolve` gets to put whatever it wants in the resolution capdata. Liveslots might not emit something malformed, but a compromised vat worker could, and anything coming in over comms has a capdata `.body` string produced by off-kernel senders, both of which are the kind of attackers I want to defend against. Comms [does the same thing](https://github.com/Agoric/agoric-sdk/blob/68ecc09f4f57ed833acbd4dcc6de6140fae3f9e7/packages/SwingSet/src/vats/comms/delivery.js#L266), for the same reason.

> Moreover, my read is that the new code is slightly stricter than what it replaces (due to the `krefOf(encValue) === slotValue` check), so I'm not sure what test you fear is being missed now that was present previously.

I think you're right, but I just thought of a different concern. Imagine that a vat or comms remote resolves a Promise to the serialized form of `{ getKref: 'haha', presence: Far(..) }`:
* the capdata will have a single `data.slots` for the presence's kref, but the top-level object will be a record, not a Far
* when someone sends a message to this promise, the kernel will call `extractSingleSlot()` to see if the resolution is an object
* `data.slots.length === 1`, `encValue = { getKref: 'haha', presence: Far(iface, { iface: () => iface, getKref: () => kref })}`
* we call `krefOf(encValue)`, which doesn't find `obj` in `fromMap`, but *does* see a truthy `obj.getKref` property, so it calls `obj.getKref()`, which throws with a TypeError
* that throws inside the kernel as it's trying to deliver a message, which probably causes a kernel panic

We really need this to be robust against adversarial capdata, even well-formed "normal" capdata.

FUDCo: It sounds like we need `krefOf` to verify that the `getKref` property is not only present but actually a function, and if not simply return `null` like it would for any other not-just-a-single-presence-value value.

warner:
see suggested changes in https://github.com/Agoric/agoric-sdk/pull/6587

packages/SwingSet/src/lib/kmarshal.js
```diff
@@ -0,0 +1,75 @@
...
+    // want nor need a promise, nor the overhead of a map to keep track of it
+    // with.  This behavior is in service of defense against a hypothesized
+    // security issue whose exact nature has largely been forgotton in the
+    // months since it was first encountered.  MarkM is currently researching
+    // what the problem was thought to have been, to see if it is real and to
+    // understand it if so.  This will eventually result in either changes to
+    // the smallcaps encoding or to the marshal setup API to support the purely
+    // manipulative use case.  In the meantime, this ugliness...
+    const p = new Promise(() => undefined);
+    refMap.set(p, kref);
```
warner, line 39: crazy idea: `p.then = () => throw('ersatz promise placeholders are not to be then-ed')`

erights: Because own `then` methods on promises create a reentrancy hazard, a promise with such an own `then` method is not considered passable.

Actually, a promise with any own properties at all is not considered passable https://github.com/endojs/endo/blob/master/packages/marshal/src/helpers/safe-promise.js

erights: @mhofman ☝️

warner: hm, maybe not allowed by Endo/lockdown

warner: hm, but `Object.defineProperty(p, 'then', { value: ...` seems to work

erights: It is allowed by `lockdown`. It is a perfectly fine element of SES. But it is not Passable, and so rejected by marshal.

packages/SwingSet/src/lib/kmarshal.js
```diff
@@ -0,0 +1,79 @@
...
+    kref.startsWith('rp')
+  ) {
+    // TODO: temporary hack because smallcaps encodes promise references
+    // differently from remotable object references, and the current version of
+    // the smallcaps decoder annoyingly insists that if the encoded body string
+    // says a slot is a promise, then convertSlotToVal had better by damn return
+    // an actual Promise, even if, as in the intended use case here, we neither
+    // want nor need a promise, nor the overhead of a map to keep track of it
+    // with.  This behavior is in service of defense against a hypothesized
+    // security issue whose exact nature has largely been forgotton in the
```
gibson042, line 32:
```suggestion
    // security issue whose exact nature has largely been forgotten in the
```

packages/SwingSet/src/liveslots/collectionManager.js
```diff
@@ -71,32 +70,9 @@ function matchAny(patt) {
...
-      X`decodeToJustin(${serializedValue}) threw ${justinError} `,
-    );
-    throw err;
-  }
   assert.fail(
-    X`value is not durable: ${value} at slot ${q(
-      slotIndex,
-    )} of ${encodedValue}`,
+    // prettier-ignore
+    X`value is not durable: ${value} at slot ${q(slotIndex)} of ${serializedValue.body}`,
```
gibson042, line 75: Do we still have need of ```assert.fail(X`…`)```, or can it now be ``` Fail`…` ```?

gibson042 (commented):
The tiniest of nits.

packages/SwingSet/src/liveslots/collectionManager.js
warner: I didn't know we could do that, but I don't (yet) find `Fail..` to be more readable than `assert.fail(..)`.

warner (approved):

packages/SwingSet/src/liveslots/collectionManager.js
gibson042: It was introduced in https://github.com/endojs/endo/pull/1334 .

PR #6451, opened 2022-10-13
Patch bug in `stack-utils` npm package

FUDCo:
Lockdown breaks Ava under certain versions of NodeJS.  It turns out to be due to a bug in the `stack-utils` npm package which trips over the override mistake. I've filed an issue with the maintainer of that package, but in the meantime this patch deals with the problem.

It is unclear which version of Node starts manifesting the problem, but at least v16.16.0 has it.  The problem manifests as Ava test failures being reported unhelpfully -- whereas one should get a stack backtrace from the point in the test file where the failure happened, usually along with a nice diff between expected and actual result values from things like `t.deepEqual`, instead you get an "Unable to serialize error" exception with a stack trace from somewhere deep in the bowels of Ava.  The problem was the `stack-utils` npm package (which Ava uses) failing a property assignment to `Object.prototype` (which should have been a simple property assignment to an ordinary property of an ordinary object but wasn't because of the override mistake) in an environment where `Object.prototype` is frozen because of lockdown.

Without this patch (or, the eventual hoped for fix to the `stack-utils` npm package itself), debugging Ava test failures is grossly inhibited if you are using one of the newer versions of Node that manifest the problem.

See https://github.com/tapjs/stack-utils/issues/70

### DIFF
--- a/patches/stack-utils+2.0.5.patch
+++ b/patches/stack-utils+2.0.5.patch
@@ -1,0 +1,22 @@
+diff --git a/node_modules/stack-utils/index.js b/node_modules/stack-utils/index.js
+index ed14bd3..ad9eeb1 100644
+--- a/node_modules/stack-utils/index.js
++++ b/node_modules/stack-utils/index.js
+@@ -161,7 +161,7 @@ class StackUtils {
+     setFile(res, site.getFileName(), this._cwd);
+ 
+     if (site.isConstructor()) {
+-      res.constructor = true;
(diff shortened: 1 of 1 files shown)


kriskowal (approved):

erights:
PR for stack-utils now at https://github.com/tapjs/stack-utils/pull/71

(If they don't respond to an issue, they often will to a pull request. We'll see)

erights:
See also https://github.com/endojs/endo/issues/576#issuecomment-1278277646

Issue #6294, opened 2022-09-21
Dapp Wallet Bridge is Flaky


samsiegart:
## Describe the bug
We have some existing work items related to the dapp wallet bridge:
- https://github.com/Agoric/agoric-sdk/issues/6027
- https://github.com/Agoric/agoric-sdk/issues/6280

There's also other issues, like refreshing the wallet ui breaks the dapp and doesn't give any indication, requiring the user to refresh and reconnect the dapp.

The bridge signaling logic and state management with local storage likely just needs to be redesigned and cleaned up. We should investigate and design something that results in a better user experience overall.

samsiegart:
# Design Proposal

## TL;DR
Make the dapp connect directly to Keplr to load all the purse data instead of getting it from the wallet bridge. Then we can remove all the CapTP stuff from the bridge and just turn it into a one-way mechanism for dapps to write offer proposals to the wallet without a two-way connection. This also gets us closer, both technically and UX-wise, to our end goal of getting rid of the bridge.

## Design Sketch
### Current Purpose of Wallet Bridge
(1) There isn't currently a good way to see all the offer data in Keplr before signing it (<insert link to sign-mode-textual issue here>?), so we lose most of the benefits of offer safety at the UX level. To fix this, we have dapps send messages to the wallet UI with descriptions of offers that they want the user to approve. The wallet UI is the single trusted source where the user evaluates those offers, and from there they prompt keplr to sign the messages.

(2) The wallet bridge is also used for other things, such as letting the dapp view all of the user's purses and current offers. However, this functionality could easily be included in a UI component directly inside the dapp, assuming we let the dapp connect to keplr to view the user's address.

### Problems with Wallet Bridge
It has combined dependencies on local storage, iframes, and CapTP. The implementation is complex and has resulted in many bugs and UX issues such as this one. Fixing this in its current form is technically difficult, error prone, and has maintenance risk as browser behaviors around these APIs may change.

### Proposed Solution
The ultimate goal here should be to remove the need for the wallet bridge entirely. Either through sign-mode-textual, metamask snaps, or upstream changes to keplr itself, we would like the user to be able to securely interact with dapps with nothing more than a browser extension. However, due to the problem (1) above presents, we require some trusted UI to evaluate offers.

[...]

PR #6452, opened 2022-10-14
feat(web-components): add 'makeAgoricKeplrConnection' util

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6294

Creates a utility function for connecting to keplr and loading the user's smart wallet (just purses for now) without the bridge.

Short demo of it working with a PSM UI with some minimal changes:
keplrpurses

### DIFF
--- a/packages/web-components/src/keplr-connection/KeplrConnection.js
+++ b/packages/web-components/src/keplr-connection/KeplrConnection.js
@@ -5,6 +5,8 @@ import { getKeplrAddress } from './getKeplrAddress';
 import { getChainId } from './getChainId';
 import { watchWallet } from './watchWallet';
 
+// TODO: We need a way to detect the appropriate network-config, and default it
+// to mainnet.
 const DEFAULT_NETWORK_CONFIG = 'https://ollinet.agoric.net/network-config';
 
 export const makeAgoricKeplrConnection = async (
--- a/packages/web-components/src/keplr-connection/KeplrConnection.js
+++ b/packages/web-components/src/keplr-connection/KeplrConnection.js
@@ -1,0 +1,26 @@
+// @ts-check
+import { makeLeader } from '@agoric/casting';
+import { makeImportContext } from '@agoric/wallet/api/src/marshal-contexts';
+import { getKeplrAddress } from './getKeplrAddress';
+import { getChainId } from './getChainId';
+import { watchWallet } from './watchWallet';
+
+const DEFAULT_NETWORK_CONFIG = 'https://ollinet.agoric.net/network-config';
+
+export const makeAgoricKeplrConnection = async (
+  networkConfig = DEFAULT_NETWORK_CONFIG,
+) => {
+  const chainId = await getChainId(networkConfig);
+  const address = await getKeplrAddress(chainId);
+
+  const context = makeImportContext();
+  const leader = makeLeader(networkConfig);
+  const walletNotifiers = await watchWallet(leader, address, context);
+
+  return {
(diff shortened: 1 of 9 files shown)


samsiegart:
I think we should just remove this dapp-treasury build CI step. We already expect the dapp to be broken currently. Is that reasonable @michaelfig?

michaelfig:
> I think we should just remove this dapp-treasury build CI step. We already expect the dapp to be broken currently. Is that reasonable @michaelfig?

Yeah, that's fine for now.

packages/web-components/src/keplr-connection/KeplrConnection.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
+import { makeLeader } from '@agoric/casting';
+import { makeImportContext } from '@agoric/wallet/api/src/marshal-contexts';
+import { getKeplrAddress } from './getKeplrAddress';
+import { getChainId } from './getChainId';
+import { watchWallet } from './watchWallet';
+
+const DEFAULT_NETWORK_CONFIG = 'https://ollinet.agoric.net/network-config';
```
michaelfig, line 10: Deserves a TODO:
```suggestion
// TODO: We need a way to detect the appropriate network-config, and default it to mainnet.
const DEFAULT_NETWORK_CONFIG = 'https://ollinet.agoric.net/network-config';
```

packages/web-components/src/keplr-connection/errors.js
```diff
@@ -0,0 +1,7 @@
+// @ts-check
+
+export const Errors = {
+  enableKeplr: '0',
+  networkConfig: '1',
+  noSmartWallet: '2',
```
michaelfig: Suggest more searchable constants:
```suggestion
  enableKeplr: 'KEPLR_CONNECTION_ERROR_ENABLE_KEPLR',
  networkConfig: 'KEPLR_CONNECTION_ERROR_NETWORK_CONFIG',
  noSmartWallet: 'KEPLR_CONNECTION_ERROR_NO_SMART_WALLET',
```

packages/web-components/src/keplr-connection/followLatest.js
```diff
@@ -0,0 +1,48 @@
+// @ts-check
+import { iterateEach } from '@agoric/casting';
+import { AmountMath } from '@agoric/ertp';
+
+export const followLatest = async ({startingHeight, latestFollower, updatePurses, brandToPurse}) => {
```
michaelfig:
```suggestion
// TODO: Provide a way for the caller to terminate the follower loop.
export const followLatest = async ({startingHeight, latestFollower, updatePurses, brandToPurse}) => {
```

michaelfig (approved):
Approach looks good.  Since this is additive, I'm okay approving even if some of the important details are still being worked out.

packages/web-components/src/keplr-connection/followLatest.js
samsiegart: Done. Is there a good way to cancel a follower? Maybe set a flag and break on the next update?

packages/web-components/src/keplr-connection/errors.js
samsiegart: Done

packages/web-components/src/keplr-connection/KeplrConnection.js
samsiegart: Done. What were you thinking by "detect" here? I was thinking the dapp could just be responsible for providing it.

PR #6456, opened 2022-10-15
build: remove some test-dapp checks

samsiegart:
Unblocks https://github.com/Agoric/agoric-sdk/pull/6452

These dapps are known to be broken anyway so just disable the checks for now.


dckc (approved):
I see a comment from @michaelfig where he says this is ok for now. So LGTM

PR #6463, opened 2022-10-18
Dapp wallet bridge

### LINKED ISSUE #6294: Dapp Wallet Bridge is Flaky

samsiegart:
Fixes https://github.com/Agoric/agoric-sdk/issues/6294 by redesigning the dapp<->wallet bridge as described in the comment section

Tested by connecting to a local PSM version (https://github.com/Agoric/dapp-psm/pull/22), connecting to the wallet, and proposing/signing offers.


packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+ * @param {string} chainId
+ * @param {string} address
+ * @param {string} proposedPetname
+ */
+const requestDappConnection = (origin, chainId, address, proposedPetname) => {
+  const dapp = loadDapp(chainId, address, origin);
+  if (dapp) {
+    return;
+  }
+  upsertDapp(chainId, address, { origin, petname: proposedPetname });
```
turadg: Surprising to me that we up-sert instead of insert even when the dapp is known to not be present. I suppose because there's no `insert`. And we need to skip the upsert if the dapp is present because we don't want to update the petname. It would help to have this explained in the comments. At the least, have `proposedPetname` jsdoc indicate that it will only be used if the dapp identified by the other three args isn't present.

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+ * notifies the dapp.
+ *
+ * @param {string} origin
+ * @param {string} chainId
+ * @param {string} address
+ * @param {boolean} latestValue
+ */
+const watchDappApproval = (origin, chainId, address, latestValue) => {
+  watchDapps(chainId, address, dapps => {
+    console.log('something happened!!!', dapps);
```
turadg: I assume such console output will be removed before this is marked R4R

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+    const dapp = dapps.find(d => d.origin === origin);
```
turadg: how can we be sure there is at most one dapp per origin? please make the data structure or insertion enforce that and document that invariant.

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+};
+
+/**
+ * Watches for changes in local storage to the dapp's approval status and
+ * notifies the dapp.
+ *
+ * @param {string} origin
+ * @param {string} chainId
+ * @param {string} address
+ * @param {boolean} latestValue
```
turadg: too vague
```suggestion
 * @param {boolean} currentlyApproved
```

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+ */
+const watchDappApproval = (origin, chainId, address, latestValue) => {
+  watchDapps(chainId, address, dapps => {
+    console.log('something happened!!!', dapps);
+    const dapp = dapps.find(d => d.origin === origin);
+    const isDappApproved = dapp && dapp.isEnabled;
+    if (isDappApproved !== latestValue) {
```
turadg: worth distinguishing between "not yet approved" and "approval removed"?

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+      sendMessage({
+        type: BridgeProtocol.dappApprovalChanged,
+        isDappApproved,
+      });
+      latestValue = isDappApproved;
+    }
+  });
+};
+
+const createAndAddOffer = (origin, chainId, address, config) => {
```
turadg: types please.

the first three values get repeated a lot. consider a `DappId` tuple.

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+  const currentTime = new Date().getTime();
+  const id = currentTime;
```
turadg, line 89: this isn't reliable, due to different clocks. please `lastOfferId` from wallet `.current`. If you deem that out of scope, please put a TODO here. Bonus points for a ticket :)

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const offer = {
+    id,
+    instancePetname: `instance@${config.instanceHandleBoardId}`,
+    requestContext: { origin },
+    meta: {
+      id,
+      creationStamp: currentTime,
+    },
+    status: 'proposed',
```
turadg: please define and use an enum

packages/wallet/ui/src/bridge-dapp.jsx
turadg: this is conflating unknown dapp with not yet approved dapp. how intentional is that?

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * Notifies the dapp about whether it's approved or not and watches for changes
+ * to its approval status.
+ *
+ * @param {string} origin
+ * @param {string} chainId
+ * @param {string} address
+ */
+const checkIfDappApproved = (origin, chainId, address) => {
```
turadg: this name doesn't convey the watching. consider,
```suggestion
const checkAndWatchDappApproval = (origin, chainId, address) => {
```

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+  watchDappApproval(origin, chainId, address, isDappApproved);
+};
+
+const handleIncomingMessages = () => {
+  /** @type { string } */
+  let origin;
+  /** @type { string } */
+  let address;
+  /** @type { string } */
+  let chainId;
```
turadg: this affirms the case for a tuple, since these must all mutate together.

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+
+  window.addEventListener('message', ev => {
+    const type = ev.data?.type;
+    if (!type?.startsWith('agoric_')) {
```
turadg: magic string needs const and documentation

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,180 @@
...
+      return;
+    }
+    if (origin === undefined) {
```
turadg: surprising that whatever the first message is gets to define the origin, which then remains fixed. seems ripe for race conditions. can `origin` be a parameter to `handleIncomingMessages`?

packages/wallet/ui/src/components/Dapp.jsx
```diff
@@ -4,7 +4,7 @@ import { E } from '@endo/eventual-send';
 import { withApplicationContext } from '../contexts/Application';
 
 const DappWithoutContext = ({ dapp, dapps }) => {
-  const { petname, origin, dappOrigin, enable, actions } = dapp;
+  const { petname, origin, dappOrigin, isEnabled, actions } = dapp;
```
turadg, line 7: 👍

packages/wallet/ui/src/components/Offer.jsx
```diff
@@ -147,6 +149,14 @@ const OfferWithoutContext = ({
       <ErrorBoundary>
         <Proposal offer={offer} />
       </ErrorBoundary>
+      {signingError && (
+        <div className="OfferEntry">
+          <h6>Signing Error</h6>
+          <details style={{ whiteSpace: 'pre-wrap' }}>
+            {`${signingError}`}
```
turadg: I don't know if we have style guidance on this but this looks like a cast to me,
```suggestion
            {String(signingError)}
```
the string template makes me search for what's in the template and the cast makes clear more quickly what the intent is.

packages/wallet/ui/src/components/SmartWalletConnection.jsx
```diff
@@ -88,29 +87,26 @@ const SmartWalletConnection = ({
     }
 
     let cancelIterator;
-    let cleanupStorageBridge;
```
turadg, line 91: 🎉

packages/wallet/ui/src/service/Offers.js
```diff
@@ -1,36 +1,97 @@
...
+
+  const augmentOffer = async offer => {
+    const {
+      id,
+      instanceHandleBoardId,
+      invitationMaker: { method },
+      proposalTemplate: { give, want },
+    } = offer;
+
+    const mapPurses = obj =>
```
turadg: param should say what's expected, ideally with a type annotation. function name should say what comes out.

packages/wallet/ui/src/service/Offers.js
```diff
@@ -1,36 +1,97 @@
...
  * @param {Notifier<any>} chainOffersNotifier
+ * @param {Marshaller} marshaller
  */
 export const getOfferService = (
-  publicAddress,
+  chainId,
+  address,
   signSpendAction,
   chainOffersNotifier,
+  marshaller,
```
turadg: consider indicating what this marshals into. e.g. `boardIdMarshaller`

packages/wallet/ui/src/service/Offers.js
```diff
@@ -1,36 +1,97 @@
...
 
   const upsertOffer = offer => {
     offers.set(offer.id, offer);
-    add(publicAddress, offer);
+    add(chainId, address, { ...offer, spendAction: undefined });
     broadcastUpdates();
   };
 
-  const declineOffer = id => {
+  const declineOffer = (/** @type {string} */ id) => {
```
turadg, line 109: 👍

packages/wallet/ui/src/service/Offers.js
```diff
@@ -47,34 +108,81 @@ export const getOfferService = (
...
+        if ('error' in status) {
+          offers.set(status.id, {
+            ...oldOffer,
+            id: status.id,
+            status: 'rejected',
+            error: `${status.error}`,
+          });
+          remove(chainId, address, status.id);
+        } else if (
+          oldOffer.status !== 'accept' &&
```
turadg: enum

packages/wallet/ui/src/service/Offers.js
```diff
@@ -47,34 +108,81 @@ export const getOfferService = (
...
-        chainOffers = state;
         broadcastUpdates();
-      });
+      }
     }
   };
-  watchChainOffers();
+
+  const start = purseMap => {
+    pursePetnameToBrand = purseMap;
```
turadg: why would `pursePetnameToBrand` already be defined before `start`? please document how repeated calls are supposed to work and why.

packages/wallet/ui/src/service/Offers.js
```diff
@@ -47,34 +108,81 @@ export const getOfferService = (
...
+
+    watchChainOffers();
+
+    watchOffers(chainId, address, newOffers => {
+      const newOffersP = Promise.all(
+        newOffers.map(o => {
+          return augmentOffer(o).then(ao => {
+            offers.set(ao.id, {
+              ...ao,
+            });
```
turadg:
```suggestion
            offers.set(ao.id, ao);
```

packages/wallet/ui/src/service/Offers.js
```diff
@@ -1,36 +1,97 @@
...
 export const getOfferService = (
-  publicAddress,
+  chainId,
+  address,
   signSpendAction,
   chainOffersNotifier,
+  marshaller,
 ) => {
+  let pursePetnameToBrand;
   const offers = new Map();
```
turadg, line 36: consider typing this. I see that it contains "augmented" offers so it's helpful to clarify vs unaugmented.

packages/wallet/ui/src/store/Dapps.js
```diff
@@ -1,27 +1,36 @@
...
-      { origin, enable, petname, id: origin, meta: { id: origin } },
-    ],
+    [DAPPS_STORAGE_KEY, chainId, address],
+    [...dapps.filter(d => d.origin !== origin), { origin, isEnabled, petname }],
   );
 };
 
-export const removeDapp = (publicAddress, origin) => {
-  const dapps = loadDapps(publicAddress) ?? [];
+export const removeDapp = (chainId, address, origin) => {
```
turadg: place for tuple

packages/wallet/ui/src/store/Offers.js
```diff
@@ -1,22 +1,31 @@
...
+import {
+  maybeLoad,
+  maybeSave,
+  watchKey,
+  OFFERS_STORAGE_KEY,
+} from '../util/storage.js';
 
-const OFFERS_KEY_PREFIX = 'OFFERS';
+export const loadOffers = (chainId, address) =>
+  maybeLoad([OFFERS_STORAGE_KEY, chainId, address]) ?? [];
```
turadg: if `window.localStorage` is undefined or the `getItem` fails, then this will simply return `[]`. shouldn't we alert the user in that case?

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -146,9 +143,6 @@ export const makeWalletBridgeFromFollowers = (
     ]),
   );
 
-  // We assume just one cosmos purse per brand.
```
turadg, line 149: where is this assumption documented now?

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -106,24 +105,22 @@ export const makeBackendFromWalletBridge = walletBridge => {
...
 /**
+ * @param {string} chainId
+ * @param {ReturnType<import('@endo/marshal').makeMarshal>} marshaller
  * @param {import('@agoric/casting').ValueFollower<import('@agoric/smart-wallet/src/smartWallet').CurrentWalletRecord>} currentFollower
  * @param {import('@agoric/casting').ValueFollower<import('@agoric/smart-wallet/src/smartWallet').UpdateRecord>} updateFollower
- * @param {import('@agoric/casting').Leader} leader
- * @param {ReturnType<import('@endo/marshal').makeMarshal>} marshaller
  * @param {string} publicAddress
  * @param {object} keplrConnection
- * @param {string} networkConfig
```
turadg, line 115: 👍

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -323,81 +314,10 @@ export const makeWalletBridgeFromFollowers = (
...
-    publicAddress,
-    signSpendAction,
-    getNotifierMethods.getOffersNotifier(),
-  );
+  const dappService = getDappService(chainId, publicAddress);
   const { acceptOffer, declineOffer, cancelOffer } = offerService;
 
-  // We override addOffer to adapt the old proposalTemplate format to the new
-  // smart-wallet format.
-  const addOfferPSMHack = async details => {
```
turadg, line 349: 🎉

packages/wallet/ui/src/util/storage.js
```diff
@@ -1,17 +1,26 @@
...
 };
 
+/**
+ * @param {string} key
+ */
 export const maybeLoad = key => {
   if (window?.localStorage) {
     try {
-      const json = window.localStorage.getItem(key);
+      const json = window.localStorage.getItem(JSON.stringify(key));
```
turadg, line 30: `key` is supposed to be a string… why does it need to be stringified again?

turadg (commented):
Good work! There are some not-ready-to-merge aspects but I see it's in draft so I won't put a "request changes" block.

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: A dapp is basically just an origin that we've given some permission to, they're keyed by origin. I think it's enforced but just added a comment.

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: And yea it's intentional, when you delete a dapp it becomes unknown, and that's how we "unapprove" them in the UI.

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: I don't think it matters

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done (left a TODO for config type)

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: We cannot reference wallet `.current` here without completely reverting this implementation. I think this is an improbable edge case where the user will get an error and just have to try again, but will leave a TODO (there's probably ways we can avoid collision without relying on wallet `.current`).

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Added it to `BridgeProtocol`

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: We can't read the parent's origin from within this iframe, that's also why we require the separate "checkIfDappApproved" message instead of just doing that on load, because it serves as a first message so we can read the origin from the event. I don't think there's any race conditions here because we check the `agoric_` prefix first.

packages/wallet/ui/src/components/Offer.jsx
samsiegart: Sounds good, never thought of doing `String()`

packages/wallet/ui/src/service/Offers.js
samsiegart: Done

packages/wallet/ui/src/service/Offers.js
samsiegart: Just changed the method name to make it clearer for now

packages/wallet/ui/src/service/Offers.js
samsiegart: Ok changed to `mapPursePetnamesToBrands`. Not exactly "what comes out" but more descriptive I think

packages/wallet/ui/src/service/Offers.js
samsiegart: Oh yea good point it doesn't need to be. Just removed this from the outer scope and changed it to a param where needed.

packages/wallet/ui/src/util/WalletBackendAdapter.js
samsiegart: Oh looks like this was supposed to go with `brandToPurse`. Added it back.

packages/wallet/ui/src/util/storage.js
samsiegart: Just for convenience, we're JSON.stringifying arrays every time so I figured it was easier to just put this here than make helper functions and call them separately every time.

packages/wallet/ui/src/store/Offers.js
samsiegart: `utils/storage` should probably handle that. I just added some more `console.error`s for debugging purposes and a TODO for something fancier.

samsiegart (commented):
Thanks for the thorough pass! I addressed most of the feedback. The one thing I'm procrastinating on is adding types for offer configs

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: removed

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Added a jsdoc description

samsiegart:
@turadg I added a network picker to the PSM (https://github.com/Agoric/dapp-psm/pull/22), made the bridge notify the dapp of the real wallet url instead of hardcoding to wallet.agoric.app in the toasts, and I believe addressed most of your initial feedback. PTAL

packages/wallet/ui/src/bridge-dapp.jsx
turadg: > We cannot reference wallet .current here without completely reverting this implementation

We don't need to reference `.current` here. Just include `id`  in the `config` param.

Please do that and if you must use time when making the config, 1) make it use UTC time so it doesn't break when people move time zones and 2) make a TODO to use `.current`.

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,195 @@
...
+  const offer = {
+    id,
+    instancePetname: `instance@${config.instanceHandleBoardId}`,
+    requestContext: { origin },
+    meta: {
+      id,
+      creationStamp: currentTime,
+    },
+    status: OfferStatus.proposed,
+    ...config,
```
turadg: This `config` really is the "offer" and all these other fields in `const offer` are bridge communication data. Please make this something like `offerMessage` with a field `offer` that has the type of what the smart wallet is expecting, ``OfferSpec`.

packages/wallet/ui/src/bridge-dapp.jsx
turadg: Did you push it? I still see three bindings. What I'm looking for here is,
```
/** @type {DappId=} */
let dappId;
```

packages/wallet/ui/src/service/Offers.js
```diff
@@ -1,36 +1,106 @@
...
+    const {
+      id,
+      instanceHandleBoardId,
+      invitationMaker: { method },
+      proposalTemplate: { give, want },
+    } = offer;
+
+    const mapPursePetnamesToBrands = entries =>
+      Object.fromEntries(
+        Object.entries(entries).map(([kw, { pursePetname, value }]) => [
```
turadg: the param must not be `entries` or we wouldn't need to call `entries()` on it. Please name the param by what data it conveys, not its shape, the use jsdoc type to describe its shape.

packages/wallet/ui/src/store/Dapps.js
turadg:
```suggestion
export const removeDapp = (dappId) => {
```

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,195 @@
...
+import ReactDOM from 'react-dom';
+import { BridgeProtocol } from '@agoric/web-components';
+import { addOffer } from './store/Offers.js';
+import { loadDapp, upsertDapp, watchDapps } from './store/Dapps.js';
+import { OfferStatus } from './service/Offers';
+
+Error.stackTraceLimit = Infinity;
+
+/**
+ * @typedef {{ origin: string, chainId: string, address: string }} DappId
```
turadg: By tuple I meant a fixed length array, but this works too. Though I think it should be `DappKey` b/c it's a composite key, not a primary id.

packages/wallet/ui/src/util/WalletBackendAdapter.js
```diff
@@ -175,6 +170,33 @@ export const makeWalletBridgeFromFollowers = (
...
+    if (!interactiveSigner) {
+      throw new Error(
+        'Cannot sign a transaction in read only mode, connect to keplr.',
+      );
+    }
+    return interactiveSigner.submitSpendAction(data);
+  };
+
+  const getNotifierMethods = Object.fromEntries(
+    Object.entries(notifiers).map(([method, stateName]) => {
```
turadg: we have objectMap for this:
https://github.com/Agoric/agoric-sdk/blob/c3fb5f521f6667c8cbd29dd78910b545a0ca5816/packages/internal/src/utils.js#L76-L82

packages/wallet/ui/src/util/storage.js
turadg: I'm confused. `key` for `maybeSave` and `maybeLoad` is typed `string`, not array. Why would you ever need to stringify it?

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
```diff
@@ -0,0 +1,191 @@
...
+
+  render() {
+    /** @type {import('lit-html').TemplateResult<1> | undefined} */
+    let backend;
+
+    if (this.bridgeHref === null) {
+      // If we don't know the bridge href yet, render the locator so it can
+      // tell us.
+      backend = html`<agoric-iframe-messenger
+        src=${DEFAULT_LOCATOR_URL}
```
turadg: needs to be configurable for dev before merge. consider inspecting the host to get a hint or use a query param to pick

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
```diff
@@ -0,0 +1,191 @@
...
+        @message=${this.onLocateMessage}
+        @error=${this.onError}
+      ></agoric-iframe-messenger>`;
+    } else {
+      // Otherwise render the bridge.
```
turadg: It's not quickly apparent what's different between the two conditions. Consider conditionally binding `src` and `messageHandler` and then making this template call once with those values.

then you can have `const backend`. actually you won't even need to define `backend` because it can be inlined into the div.

turadg (changes requested):

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: I think its better for the wallet to control the IDs instead of relying on the dapp to provide something unique. The dapp also doesn't currently load the offers and that would be more work and computation which seems unnecessary. 

Doesn't this just have to be a unique identifier? If we don't trust the milliseconds to not collide between two offers, maybe a more sufficiently random randomness function? I don't see how different clocks/timezones could break this because the milliseconds would still have to match which is improbable.

turadg: > Doesn't this just have to be a unique identifier?

Ah, here's the misconception. No, it has to be higher than all previous ids. Each id sets a new high-water mark.

> I think its better for the wallet to control the IDs instead of relying on the dapp to provide something unique.

Is your argument that the dapp needn't know the state of the wallet? I see the case for that when considering `dapp-psm` proposing a swap, but since multiple agents can be changing wallet balances I do think dapps need to be aware of the wallet `.current` state.

packages/wallet/ui/src/service/Offers.js
samsiegart: Going to go with 'paymentProposals' but I've been thinking of these as entries in the offer, even though they're not tuples.

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
samsiegart: Made it a property. The dapp that uses this component can be responsible for changing the locator if needed. Though, this is more of an added feature, since the old bridge did not allow the locator to be configurable

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
samsiegart: done

packages/wallet/ui/src/util/storage.js
samsiegart: Hmm they're not used as strings though, yet I wasn't seeing an error in vscode. Just changed them to be `unknown`.

packages/wallet/ui/src/util/WalletBackendAdapter.js
samsiegart: This is pre-existing code and I think I'd like to get rid of this at some point since it messes with types. I'd like to defer that for now though

packages/wallet/ui/src/store/Dapps.js
samsiegart: Done

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Ah okay, done

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Just added types

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done

packages/wallet/ui/src/bridge-dapp.jsx
turadg: > A dapp is basically just an origin that we've given some permission to

If so then why does the DappKey have chainId and address? Is there a limit of one chainId,address per origin?

packages/wallet/ui/src/util/WalletBackendAdapter.js
samsiegart: done

samsiegart:
Added TODO with https://github.com/Agoric/agoric-sdk/issues/6478

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: It's so you can have different petnames or approval statuses for the same dapp across different smart wallets. I guess you could say a smart wallet is identified by `chainId,address`.

packages/wallet/ui/src/store/Dapps.js
```diff
@@ -1,27 +1,60 @@
...
   maybeSave(
-    [DAPPS_KEY_PREFIX, publicAddress],
+    [DAPPS_STORAGE_KEY, chainId, address],
     dapps.filter(d => d.origin !== origin),
   );
 };
+
+export const watchDapps = (
+  /** @type {string} */ chainId,
+  /** @type {string} */ address,
```
turadg: it would help for understanding the code if this where `SmartWalletKey` and `DappKey` was `WalletKey & {origin: string}`

packages/wallet/ui/src/bridge-dapp.jsx
turadg: Closing the loop: we agree fixing this is out of scope for this bridge work because it's an existing problem and will be fixed by https://github.com/Agoric/agoric-sdk/issues/6478

packages/wallet/ui/src/bridge-dapp.jsx
```diff
@@ -0,0 +1,190 @@
...
+const handleIncomingMessages = () => {
+  /** @type {DappKey=} */
+  let dappKey;
+
+  window.addEventListener('message', ev => {
+    const type = ev.data?.type;
+    if (!type?.startsWith(BridgeProtocol.prefix)) {
+      return;
+    }
+    if (dappKey === undefined) {
```
turadg, line 131: IIUC this is only supposed to happen for a `checkIfDappApproved` which must be the first message. If so, please move this under that case.

or… add a stronger check here like,
```
assert.same(type, BridgeProtocol.checkIfDappApproved, 'first message must be checkIfDappApproved');
```

turadg (approved):
Approving assuming you'll address my 2 new requests before merge.

packages/wallet/ui/src/bridge-dapp.jsx
samsiegart: Done

packages/wallet/ui/src/store/Dapps.js
samsiegart: Done

PR #6465, opened 2022-10-18
fix(vats): Toy USDC denom in test config

dckc:
refs: 6284

## Description

emerynet has `ibc/toyusdc` , not `ibc/usdc1234`

### Testing Considerations

 - add other denoms while we're at it? (no)
 - add `ibc/toyusdc` to faucet


dckc:
>  add other denoms while we're at it?

seems like just one is ok

arirubinstein (approved):

arirubinstein (approved):

PR #6475, opened 2022-10-19
fix: Tighten invitation proposal patterns

erights:
Staged on https://github.com/Agoric/agoric-sdk/pull/5422 "multiples"

Use tight proposal patterns on production invitations to, among other things, see if the Zoe prohibition on wash invitations is something we should bother preserving. (My assumption is not, but worth doing some of this experiment first.)

### DIFF
--- a/packages/governance/src/committee.js
+++ b/packages/governance/src/committee.js
@@ -3,6 +3,7 @@ import { makeStore, makeHeapFarInstance, M } from '@agoric/store';
 import { natSafeMath } from '@agoric/zoe/src/contractSupport/index.js';
 import { E } from '@endo/eventual-send';
 
+import { EmptyProposal } from '@agoric/zoe/src/typeGuards.js';
 import { makeHandle } from '@agoric/zoe/src/makeHandle.js';
 import {
   getOpenQuestions,
@@ -96,7 +97,12 @@ const start = (zcf, privateArgs) => {
                 return E(voteCap).submitVote(voterHandle, positions, 1n);
               };
 
-              return zcf.makeInvitation(continuingVoteHandler, 'vote');
+              return zcf.makeInvitation(
+                continuingVoteHandler,
+                'vote',
+                undefined,
+                EmptyProposal,
+              );
             },
           },
         ),
@@ -107,7 +113,12 @@ const start = (zcf, privateArgs) => {
     // This will produce unique descriptions because
     // makeCommitteeVoterInvitation() is only called within the following loop,
     // which is only called once per Electorate.
-    return zcf.makeInvitation(offerHandler, `Voter${index}`);
+    return zcf.makeInvitation(
+      offerHandler,
+      `Voter${index}`,
+      undefined,
+      EmptyProposal,
+    );
   };
 
   const { committeeName, committeeSize } = zcf.getTerms();
--- a/packages/governance/src/electorateTools.js
+++ b/packages/governance/src/electorateTools.js
@@ -1,3 +1,4 @@
+import { EmptyProposal } from '@agoric/zoe/src/typeGuards';
 import { E } from '@endo/eventual-send';
 import { deeplyFulfilled, Far } from '@endo/marshal';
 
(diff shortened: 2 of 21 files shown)


Chris-Hibbert (approved):

erights:
Reviewers, especially for the occurrences of `EmptyProposal`, let me know if each one makes sense, and where the corresponding `offer` call happens. I was not able to find many of those `offer` calls, and so could not easily tell what proposal pattern is appropriate.

packages/governance/src/committee.js
```diff
@@ -92,7 +93,12 @@ const start = (zcf, privateArgs) => {
                 return E(voteCap).submitVote(voterHandle, positions, 1n);
               };
 
-              return zcf.makeInvitation(continuingVoteHandler, 'vote');
+              return zcf.makeInvitation(
+                continuingVoteHandler,
```
Chris-Hibbert, line 101: `continuingVoteHandler` is exercised by the Wallet. If it's ever not an `EmptyProposal`, I think that's under the user's control, but I have no idea how.

packages/governance/src/committee.js
```diff
@@ -103,7 +109,12 @@ const start = (zcf, privateArgs) => {
     // This will produce unique descriptions because
     // makeCommitteeVoterInvitation() is only called within the following loop,
     // which is only called once per Electorate.
-    return zcf.makeInvitation(offerHandler, `Voter${index}`);
+    return zcf.makeInvitation(
+      offerHandler,
+      `Voter${index}`,
+      undefined,
+      EmptyProposal,
```
Chris-Hibbert, line 120: yes, always empty.

[Example](https://github.com/Agoric/agoric-sdk/blob/ffe5e7ff3dd7ec1ffbbcd86e244ef167ed70c53b/packages/inter-protocol/test/swingsetTests/governance/vat-voter.js#L16)

packages/governance/src/electorateTools.js
```diff
@@ -69,7 +70,12 @@ const getQuestion = (questionHandleP, questionStore) =>
  */
 const getPoserInvitation = (zcf, addQuestion) => {
   const questionPoserHandler = () => Far(`questionPoser`, { addQuestion });
-  return zcf.makeInvitation(questionPoserHandler, `questionPoser`);
+  return zcf.makeInvitation(
+    questionPoserHandler,
+    `questionPoser`,
+    undefined,
+    EmptyProposal,
```
Chris-Hibbert, line 94: This one is [for the use of the contractGoveror. ](https://github.com/Agoric/agoric-sdk/blob/ffe5e7ff3dd7ec1ffbbcd86e244ef167ed70c53b/packages/governance/src/contractGovernor.js#L195).  Always empty.

packages/governance/src/noActionElectorate.js
```diff
@@ -39,8 +40,12 @@ const start = zcf => {
 
   const creatorFacet = makeHeapFarInstance('creatorFacet', ElectorateCreatorI, {
     getPoserInvitation() {
-      return zcf.makeInvitation(() => {},
-      `noActionElectorate doesn't allow posing questions`);
+      return zcf.makeInvitation(
+        () => {},
+        `noActionElectorate doesn't allow posing questions`,
+        undefined,
+        EmptyProposal,
```
Chris-Hibbert, line 45: like committee.js

packages/inter-protocol/src/psm/psmCharter.js
```diff
@@ -4,7 +4,10 @@ import '@agoric/governance/src/exported.js';
 import { makeScalarMapStore, M, makeHeapFarInstance, fit } from '@agoric/store';
 import '@agoric/zoe/exported.js';
 import '@agoric/zoe/src/contracts/exported.js';
-import { InstanceHandleShape } from '@agoric/zoe/src/typeGuards.js';
+import {
+  EmptyProposal,
```
Chris-Hibbert: These are all correctly `Empty`

packages/inter-protocol/src/psm/psm.js
```diff
@@ -290,6 +292,8 @@ export const start = async (zcf, privateArgs, baggage) => {
         M.splitRecord({
           give: { In: stableAmountShape },
           want: M.or({ Out: anchorAmountShape }, {}),
+          multiples: 1n,
+          exit: M.any(),
         }),
```
erights, line 277: This now represents all the fields of the record, not just a subset. Thus, the `M.splitRecord` should be removed and the record pattern should be the proposalShape directly.

erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/7041

PR #6477, opened 2022-10-19
fix: better proposal mismatch errors

erights:
Use the description string in the proposal mismatch error, to tell which invitation was involved.


Chris-Hibbert (approved):
Nice!

erights:
@Mergifyio requeue

PR #6479, opened 2022-10-20
fix: simplify unnecessary promise

erights:
We already happen to have the resolved reference due to an earlier `await` so may as well use that.

Noticed while working on https://github.com/Agoric/agoric-sdk/pull/6475


turadg (approved):
Code LGTM.

For commit history: not a `fix` since it doesn't change behavior. Should be `refactor:` or `chore:`. Not worth another push and CI round but perhaps for the future.

PR #6482, opened 2022-10-20
fix: template fail

erights:
See https://github.com/endojs/endo/pull/1334

Stacked on https://github.com/Agoric/agoric-sdk/pull/6613

@turadg and @dtribble , once ready for review, I will ask you two to review. Until then, I cannot unlist you because of codeowners, but please ignore.


erights:
Closing because unbundled into 
#6619 
#6620 
#6621 
#6622 
#6623

PR #6483, opened 2022-10-20
feat(agops): oracle commands

turadg:
refs: #5791 (will leave open until @jacquesvcritien confirms it fits their needs)


turadg:
Ran into some problems with using core proposals so I just put it into boot-psm https://github.com/Agoric/agoric-sdk/pull/6499

PR #6484, opened 2022-10-20
chore: Sync Endo versions

kriskowal:
- chore: Sync Endo versions
- chore: Remove bundle-source patch
- chore: Roll ses-ava patch forward from 0.2.33 to 0.2.34
- chore: Update yarn.lock
- chore: Manually normalize yarn.lock
- chore: Bump meter version
- chore: Bump xsnap meter type


FUDCo (approved):
Well ok then

kriskowal:
The errors we’re observing in CI will require changes in Endo that will in turn require a new pull request. Closing until then.

Issue #6474, opened 2022-10-19
pull web UI out of agoric-sdk


turadg:
## What is the Problem Being Solved?

the Web UI packages have different dependencies, CI and deploy processes from the chain parts of agoric-sdk. Problems we could solve by pulling it out,
1. Having to release all agoric-sdk to bump versions for dapp deps ([example](https://github.com/Agoric/agoric-sdk/pull/6486#issuecomment-1286895311))
2. Needlessly running loadgen on Web UI changes, which can't possible affect the chain (~1hr merge delay, unless dev tags to skip)
3. NPM module resolution for dependencies that run in user browser VM vs chain
4. Security advisory noise for inclusion of those dependencies in different contexts
5. Having to build extra tooling for https://github.com/Agoric/wallet-app/issues/40

## Description of the Design

Move @agoric/wallet-ui package to wallet-app repo.

Update project management tooling to include tickets from that repo.

## Security Considerations
TBD
## Test Plan

TBD

PR #6486, opened 2022-10-21
build(web-components): add necessary dependencies

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6294


yarn.lock
```diff
@@ -1666,6 +1666,11 @@
   resolved "https://registry.yarnpkg.com/@endo/eventual-send/-/eventual-send-0.16.5.tgz#39f2e21b5331cd9e423a8440f9e74e28fea6701e"
   integrity sha512-tWmia2jv+beGf4oVOxUqcHc358PX3Gssz4qUD3BA7D6d7AW75Pk34ArWsi29E9w6aNMrnhnpX+EaRBWoj/Q1PA==
 
+"@endo/eventual-send@^0.16.6":
+  version "0.16.6"
```
mhofman: Duplicate versions of eventual send is probably not right. Please dedup or revert to 0.16.5

samsiegart: Fixed thanks. I blindly took the latest version not realizing the sdk was on 0.16.5

mhofman:
While you can manually bump versions, this won't produce the effect you want. This repo doesn't automatically publish a bare package if it's version changes, but only when a release is manually cut. However every commit merged into master gets a prerelease published on npm. You may be able to use that instead.

samsiegart:
> While you can manually bump versions, this won't produce the effect you want.

Okay just undid that. Would be nice to have the versions line up more logically with features in this package but correct in that my concern is just that the changes are available somehow.

mhofman:
> Would be nice to have the versions line up more logically with features in this package but correct in that my concern is just that the changes are available somehow.

This might be another argument for UI packages to live in another repo with a different release schedule from the main SDK.

cc @warner

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
```diff
@@ -188,11 +188,19 @@ export class DappWalletBridge extends LitElement {
   }
 
   render() {
-    const hasBridgeHref = this.bridgeHref !== null;
+    // Locate the bridge first.
+    if (this.bridgeHref === null) {
+      return html`<agoric-iframe-messenger
```
samsiegart, line 200: @turadg The `<agoric-iframe-messenger>` doesn't seem to play well when you dynamically change its `src`. It only checks for a loading timeout on initial render, so collating these two as before resulted in losing a valuable error state if the bridge iframe never loaded.

turadg:
> This might be another argument for UI packages to live in another repo with a different release schedule from the main SDK.

Agreed. Added to problem list in https://github.com/Agoric/agoric-sdk/issues/6474

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
turadg: good to know. I wonder if we can make `agoric-iframe-messenger` able to handle that.

turadg (approved):

packages/web-components/src/dapp-wallet-bridge/DappWalletBridge.js
samsiegart: I think we could fix it to but I'd like to keep this change contained to get it in quicker

samsiegart:
@turadg PTAL, I made some changes to the bridge component:
1. Made the locator optional by allowing the dapp to supply the `bridgeHref` directly. I discovered that brave shields seem to only break the locator, so this allows some users to not have to disable shields to use the site.
2. Added better error handling around localstorage being blocked. I discovered that firefox by default has a shield-like button that blocks cross-site cookies, so now we get a better message like:
Screen Shot 2022-10-21 at 8 43 14 AM


Tested in chrome, firefox, and brave.

turadg (approved):
Approving again but I urge you to copy these files into the dapp instead of sourcing them from the SDK, so we can begin decoupling dapp release from agoric-sdk releases.

mhofman:
I will remove myself from reviewers since the package version aspect has been resolved and this PR has morphed away from that.

PR #6488, opened 2022-10-21
fix(wallet-ui): dont overwrite all offers when adding new one

samsiegart:
refs: https://github.com/Agoric/agoric-sdk/issues/6294

### DIFF
--- a/packages/wallet/ui/src/store/Offers.js
+++ b/packages/wallet/ui/src/store/Offers.js
@@ -1,3 +1,4 @@
+// @ts-check
 import {
   maybeLoad,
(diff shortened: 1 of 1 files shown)


turadg (approved):
This looks safe to copy over. Just this file and `util/storage.js` instead of depending on an agoric-sdk rev.

samsiegart:
> This looks safe to copy over. Just this file and `util/storage.js` instead of depending on an agoric-sdk rev.

This is part of the wallet not the dapp. I can deploy this branch directly to wallet.agoric.app if need be though

PR #6492, opened 2022-10-24
omit superfluous 'SEVERED: ' from wallet spend proposal

turadg:
## Description

The SEVERED warning pertains to presences but we're sending the offer marshalled so it's distracting to the user. This removes it when the offer is being sent.

### Security Considerations

It doesn't remove it any earlier because there are no guarantees that other consumers would expect it to be a valid presence.

The code comment suffices because the distinction is relevant to Wallet UI users.

### Testing Considerations

Manually tested.

### DIFF
--- a/packages/wallet/ui/src/service/Offers.js
+++ b/packages/wallet/ui/src/service/Offers.js
@@ -33,6 +33,7 @@ export const getOfferService = (
   chainOffersNotifier,
   boardIdMarshaller,
 ) => {
+  /** @type {Map<number, Offer>} */
   const offers = new Map();
   const { notifier, updater } = makeNotifierKit();
   const broadcastUpdates = () => updater.updateState([...offers.values()]);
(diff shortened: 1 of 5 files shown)


packages/wallet/ui/src/service/Offers.js
```diff
@@ -93,10 +94,17 @@ export const getOfferService = (
       }),
     );
 
+    // XXX the SEVERED warning pertains to presences but we're sending the offer
+    // marshalled so it's just distracting.
+    const friendlySpendAction = {
+      ...spendAction,
+      body: spendAction.body.replaceAll('SEVERED: ', ''),
```
dckc: hm. are we certain that `SEVERED: ` can occur nowhere else in a `spendAction`? somebody could put it in an invitation query. Maybe that falls under "you can only foul your own nest"?

I'd be more comfortable if we didn't put `SEVERED ` in there in the 1st place. I'll try to find where the code is.

samsiegart (approved):

dckc (commented):
I'm uneasy about this approach...

packages/wallet/ui/src/service/Offers.js
dckc: What if we tweak `defaultMakePresence`?

https://github.com/Agoric/agoric-sdk/blob/1ab6c451d44e5d758713b7480cd34b8d47d04067/packages/wallet/api/src/marshal-contexts.js#L233-L236

turadg: SGTM. I was trying to keep the change isolated, but the `makeImportContext` param is a much better spot since I can still leave the default.

dtribble:
I don't see "SEVERED" appearing in the PR.  How does that get removed?

turadg:
> I don't see "SEVERED" appearing in the PR. How does that get removed?

[838f5d2](https://github.com/Agoric/agoric-sdk/pull/6492/commits/838f5d22d1e747bee2199b38c109b30e298a5fb8) overrides the `defaultMakePresence` that put in the SEVERED:
https://github.com/Agoric/agoric-sdk/pull/6492#discussion_r1003723855

dckc (approved):
I reproduced the manual testing results:

image

thanks, @samsiegart for the https://psm.inter.trade/?wallet=local for local wallet testing.

Issue #6491, opened 2022-10-24
Legal Disclaimer for Wallet and PSM UI


otoole-brendan:
## What is the Problem Being Solved?
We have a requirement for both wallet and psm.inter.trade users to be served a legal disclaimer pop-up/dialog box that will contain legal-approved copy. Users will need to take an action to dismiss and continue to use.

Requirements:
-Pop-up/dialog box that prompts user upon first time accessing 1)PSM.Inter.Trade and 2) https://wallet.agoric.app/wallet/
 
Should contain:
-Scrollable text box
-Tick box that user must tick next to 'I understand the risks and would like to proceed'[Placeholder copy] (This may change to 'I accept the terms of service' or similar. Once user tick's box, 'Proceed' button illuminates and is clickable. Clicking removes disclaimer. If user does not tick box then disclaimer cannot be dismissed (no 'x' button). 
-a means to update the disclaimer and have users re-accept should we need to (presumably we store response based on cookies/cache?)
-placeholder text for scrollable text box:

'Inter Protocol is a fully decentralized stable token protocol. No representation or warranty is made concerning any aspect of the Inter Protocol, including its suitability, quality, availability, accessibility, accuracy or safety. As more fully explained in the Terms of Use (available here) and the Risk Statement (available here), your access to and use of the Inter Protocol is entirely at your own risk and could lead to substantial losses. You take full responsibility for your use of the Inter Protocol, and acknowledge that you use it on the basis of your own enquiry, without solicitation or inducement by Contributors (as defined in the Terms of Use).'

## Description of the Design

See sample screenshots from Crescent and Shade Protocols as example![image (3).png](https://images.zenhubusercontent.com/634474599b165f9f64feb7ec/6fada59d-4d86-4eb4-a22f-3cfa542fd898)![image (4).png](https://images.zenhubusercontent.com/634474599b165f9f64feb7ec/b686687c-01b7-4fd8-9772-251723330134)s
![image (5).png](https://images.zenhubusercontent.com/634474599b165f9f64feb7ec/290f44c2-8ef7-48bc-b555-a0cba6b68168)
![image.png](https://images.zenhubusercontent.com/634474599b165f9f64feb7ec/53b667bc-bc1e-47bc-b8d8-aee02a1f5d30)

otoole-brendan:
@samsiegart think we can close this right?

PR #6495, opened 2022-10-25
feat(wallet-ui): add terms dialog on first pageview

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6491

In draft until we get final confirmation for content/behavior

image

OLD screenshot:

image


packages/wallet/ui/src/components/AppBar.jsx
```diff
@@ -63,13 +66,21 @@ const useStyles = makeStyles(theme => ({
...
 // Exported for testing only.
-export const AppBarWithoutContext = ({ connectionComponent }) => {
+export const AppBarWithoutContext = () => {
   const theme = useTheme();
   const classes = useStyles(theme);
 
-  const [dialogOpen, setDialogOpen] = useState(false);
-  const handleClosed = () => {
-    setDialogOpen(false);
+  const [areTermsAccepted, setAreTermsAccepted] = useState(
```
turadg, line 73: for future, consider sticking to `is` convention. e.g. `isDisclaimerAccepted`.

packages/wallet/ui/src/components/AppBar.jsx
```diff
@@ -118,7 +138,6 @@ export const AppBarWithoutContext = ({ connectionComponent }) => {
 };
 
 export default withApplicationContext(AppBarWithoutContext, context => ({
-  connectionComponent: context.connectionComponent,
```
turadg, line 121: 👍

packages/wallet/ui/src/store/Terms.js
```diff
@@ -0,0 +1,10 @@
+import { TERMS_KEY, maybeSave, maybeLoad } from '../util/storage.js';
+
+// The index of the current revision of the terms agreement. This should
+// increment every time the terms change.
+export const LATEST_TERMS_INDEX = 0;
+
+export const agreeToTerms = (/** @type {number} */ index) =>
+  maybeSave(TERMS_KEY, index);
+
+export const getTermsAcceptance = () => maybeLoad(TERMS_KEY) ?? -1;
```
turadg:
```suggestion
export const getTermsIndexAccepted = () => maybeLoad(TERMS_KEY) ?? -1;
```

packages/wallet/ui/src/store/Terms.js
```diff
@@ -0,0 +1,10 @@
+import { TERMS_KEY, maybeSave, maybeLoad } from '../util/storage.js';
+
+// The index of the current revision of the terms agreement. This should
+// increment every time the terms change.
+export const LATEST_TERMS_INDEX = 0;
```
turadg, line 5: I wish we had unique URLs for each set of terms but not a blocker

turadg (approved):

packages/wallet/ui/src/components/AppBar.jsx
samsiegart: In the style guide I learned from, both 'is' and 'are' are okay depending on if it's plural, also 'can', 'should', 'has', etc. I forget the grammatical term for this type of phrase though.

packages/wallet/ui/src/store/Terms.js
samsiegart: Done

PR #6497, opened 2022-10-25
chore(wallet-ui): remove smart wallet support from solo build

michaelfig:
## Description

Just remove the broken smart wallet support from the `community-dev` branch, since the actual smart wallet is available in master.


samsiegart (commented):
I think I'm not up to speed, can we just delete this code altogether instead of commenting it out? My understanding is this is a temporary branch until we have more smart wallet features to get the dapps working again.

There's also the `SmartWalletConnection` component and all the logic for code splitting in `Provider.jsx` that we can remove.

PR #6499, opened 2022-10-26
e2e demo of priceAggregator updates by smart wallet

### LINKED ISSUE #5791: [story] Oracle operators can send updates to PriceAuthority running on chain

turadg:
closes: #5791

## Description

Makes a new `decentral-oracles-config.json` and `boot-oracles.js` to demo priceAggregator instance to test e2e what was in an Ava test: https://github.com/Agoric/agoric-sdk/blob/f6898d833452c21e598f556726cc3fb7289e6936/packages/smart-wallet/test/test-oracle-integration.js#L83

Adds an `agops oracle` command to work with it and `agops-oracle-smoketest.sh` to walk through how to use it.

From discussions with @jacquesvcritien I believe this satisfies ##5791.

### Security Considerations

basic-behaviors has an installation for `priceAggregator`, though it's only instantiated by `decentral-oracles-config.json` using `boot-oracles.js` 

Integration tests demonstrate how to use.

### Testing Considerations

Ran the smoketest locally.

### DIFF
--- a/packages/agoric-cli/src/commands/oracle.js
+++ b/packages/agoric-cli/src/commands/oracle.js
@@ -1,0 +1,149 @@
+/* eslint-disable no-await-in-loop */
+/* eslint-disable @jessie.js/no-nested-await */
+// @ts-check
+/* eslint-disable func-names */
+/* global fetch, process */
+import { Command } from 'commander';
+import { inspect } from 'util';
+import {
+  boardSlottingMarshaller,
+  makeRpcUtils,
+  storageHelper,
+} from '../lib/rpc.js';
+
+const { agoricNames, fromBoard, vstorage } = await makeRpcUtils({ fetch });
+
+/**
+ *
+ * @param {import('anylogger').Logger} logger
+ */
+export const makeOracleCommand = async logger => {
+  const oracle = new Command('oracle').description('Oracle commands').usage(
+    `
+  WALLET=my-wallet
+  export AGORIC_NET=ollinet
+  
+  # provision wallet if necessary
+  agd keys add $WALLET
+  # provision with faucet, e.g.
+  open https://ollinet.faucet.agoric.net/ # and paste the key and choose client: smart-wallet
+
+  # confirm funds
+  agoric wallet list
+  agoric wallet show —from $WALLET
+
+  # prepare an offer to send
+  # (offerId is optional but best to specify as each must be higher than the last and the default value is a huge number)
+  agops oracle accept --offerId 12 > offer-12.json
+  
+  # sign and send
+  agoric wallet send --from $WALLET --offer offer-12.json
+  `,
+  );
+
+  const lookupPriceAggregatorInstance = ([brandIn, brandOut]) => {
+    const name = `${brandIn}-${brandOut} price feed`;
+    const instance = agoricNames.instance[name];
+    if (!instance) {
+      logger.debug('known instances:', agoricNames.instance);
+      throw new Error(`Unknown instance ${name}`);
+    }
+    return instance;
+  };
+  const marshaller = boardSlottingMarshaller();
+
+  /** @param {import('../lib/psm.js').BridgeAction} bridgeAction */
+  const outputAction = bridgeAction => {
+    const capData = marshaller.serialize(bridgeAction);
+    process.stdout.write(JSON.stringify(capData));
+  };
+
+  oracle
+    .command('accept')
+    .description('accept invitation to operate an oracle')
+    .requiredOption(
+      '--pair [brandIn.brandOut]',
+      'token pair (brandIn.brandOut)',
+      s => s.split('.'),
+      ['ATOM', 'USD'],
+    )
+    .option('--offerId [number]', 'Offer id', Number, Date.now())
+    .action(async function () {
+      const opts = this.opts();
+
+      const instance = lookupPriceAggregatorInstance(opts.pair);
+
+      /** @type {import('../lib/psm.js').OfferSpec} */
+      const offer = {
+        id: Number(opts.offerId),
+        invitationSpec: {
+          source: 'purse',
+          // @ts-expect-error xxx RpcRemote
+          instance,
+          description: 'oracle invitation',
+        },
+        proposal: {},
+      };
+
+      outputAction({
+        method: 'executeOffer',
+        offer,
+      });
+
+      console.warn('Now execute the prepared offer');
+    });
+
+  oracle
+    .command('pushPrice')
+    .description('add a current price sample')
+    .option('--offerId [number]', 'Offer id', Number, Date.now())
+    .requiredOption(
+      '--oracleAdminAcceptOfferId [number]',
+      'offer that had continuing invitation result',
+      Number,
+    )
+    .requiredOption('--price [number]', 'price (format TODO)', String)
+    .action(async function () {
+      const opts = this.opts();
+
+      /** @type {import('../lib/psm.js').OfferSpec} */
+      const offer = {
+        id: Number(opts.offerId),
+        invitationSpec: {
+          source: 'continuing',
+          previousOffer: opts.oracleAdminAcceptOfferId,
+          invitationMakerName: 'makePushPriceInvitation',
+          invitationArgs: harden([opts.price]),
+        },
+        proposal: {},
+      };
+
+      outputAction({
+        method: 'executeOffer',
+        offer,
+      });
+
+      console.warn('Now execute the prepared offer');
+    });
+
+  oracle
+    .command('query')
+    .description('return current aggregated (median) price')
+    .requiredOption(
+      '--pair [brandIn.brandOut]',
+      'token pair (brandIn.brandOut)',
+      s => s.split('.'),
+      ['ATOM', 'USD'],
+    )
+    .action(async function () {
+      const { pair } = this.opts();
+
+      const capDataStr = await vstorage.readLatest(
+        `published.priceFeed.${pair[0]}-${pair[1]}_price_feed`,
+      );
+      const capDatas = storageHelper.unserializeTxt(capDataStr, fromBoard);
+
+      console.log(inspect(capDatas[0], { depth: 10, colors: true }));
+    });
+  return oracle;
+};
--- a/packages/inter-protocol/src/proposals/price-feed-proposal.js
+++ b/packages/inter-protocol/src/proposals/price-feed-proposal.js
@@ -8,6 +8,9 @@ import { deeplyFulfilledObject } from '@agoric/internal';
 
 import { unitAmount } from '@agoric/zoe/src/contractSupport/priceQuote.js';
 import { reserveThenDeposit, reserveThenGetNames } from './utils.js';
+import { makeTracer } from '../makeTracer.js';
+
+const trace = makeTracer('RunPriceFeed');
 
 /** @type {(name: string) => void} */
 const sanitizePathSegment = name => {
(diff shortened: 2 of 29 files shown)


packages/vats/src/core/boot-psm.js
```diff
@@ -115,14 +118,16 @@ export const ParametersShape = M.partial({
...
+ *     demoOracleAddresses?: string[],
  *     anchorAssets: { denom: string, keyword?: string }[],
  * }} vatParameters
  */
 export const buildRootObject = (vatPowers, vatParameters) => {
   const log = vatPowers.logger || console.info;
 
   fit(harden(vatParameters), ParametersShape, 'boot-psm params');
-  const { anchorAssets, economicCommitteeAddresses } = vatParameters;
+  const { anchorAssets, demoOracleAddresses, economicCommitteeAddresses } =
```
dckc: This is useful as a spike to demonstrate that things work end-to-end, but it's important to keep demo stuff out of `boot-psm.js`, so let's not merge this as-is.

dckc (dismissed):
This is useful as a spike to demonstrate that things work end-to-end, but it's important to keep demo stuff out of `boot-psm.js`, so let's not merge this as-is.

packages/agoric-cli/src/commands/oracle.js
```diff
@@ -0,0 +1,150 @@
...
+  storageHelper,
+} from '../lib/rpc.js';
+
+const { agoricNames, fromBoard, vstorage } = await makeRpcUtils({ fetch });
+
+/**
+ *
+ * @param {import('anylogger').Logger} logger
+ */
+export const makeOracleCommand = async logger => {
```
dckc, line 20: I'm still hesitant to approve this exception to our style.

`makeOracleCommand` closes over (the output of) `fetch` (by way of `agoricNames` etc.).

> No modules should export powerful objects (for example, objects that close over `fs`). -- https://github.com/Agoric/agoric-sdk/wiki/OCap-Discipline

packages/inter-protocol/src/proposals/price-feed-proposal.js
```diff
@@ -250,3 +256,81 @@ export const getManifestForPriceFeed = async (
...
+    produce,
+    installation: {
+      consume: { priceAggregator },
+    },
+  },
+  { options: { demoOracleAddresses = [] } = {} },
+) => {
+  trace('startPriceFeeds');
+
+  const inBrandName = 'ATOM';
```
dckc, line 280: comment? "salt to taste"?

packages/smart-wallet/scripts/start-local-chain.sh
```diff
@@ -9,6 +9,8 @@ open -a /System/Applications/Utilities/Console.app $CHAIN_LOG
 # ugly way to get SDK path regardless of cwd
 SDK=$(readlink -f "$(dirname -- "$(readlink -f -- "$0")")/../../..")
 
+export CHAIN_BOOTSTRAP_VAT_CONFIG="$SDK/packages/vats/decentral-oracle-config.json"
```
dckc: :+1:

packages/smart-wallet/scripts/start-local-chain.sh
```diff
@@ -20,7 +22,10 @@ fi
 WALLET_BECH32=$(agd keys show "$WALLET" --output json | jq -r .address)
 
 # grant econ governance
-sed -i '' "s/\"voter\":.*/\"voter\": \"$WALLET_BECH32\"/" "$SDK"/packages/vats/decentral-psm-config.json
+sed -i '' "s/\"voter\":.*/\"voter\": \"$WALLET_BECH32\"/" "$CHAIN_BOOTSTRAP_VAT_CONFIG"
```
dckc, line 27: consider `jq` for editing JSON stuff.

packages/smart-wallet/scripts/start-local-chain.sh
```diff
@@ -47,7 +52,7 @@ agd query bank balances "$WALLET_BECH32" | grep ubld || exit 1
 echo "Provisioning your smart wallet..."
 agoric wallet provision --spend --account "$WALLET"
 echo "waiting for blocks"
-sleep 10
+sleep 15
```
dckc, line 57: surely that's enough! :)

packages/vats/src/core/boot-oracles.js
```diff
@@ -0,0 +1,283 @@
+// @ts-check
+/** @file Boot script for demoing oracles. Builds upon boot-psm to be used in integration tests. */
```
dckc: move to /test/ directory?

packages/smart-wallet/test/devices.js
```diff
@@ -1,25 +1,26 @@
-import bundleCentralSupply from '@agoric/vats/bundles/bundle-centralSupply.js';
-import bundleMintHolder from '@agoric/vats/bundles/bundle-mintHolder.js';
-import bundleProvisionPool from '@agoric/vats/bundles/bundle-provisionPool.js';
+// @ts-check
+import centralSupply from '@agoric/vats/bundles/bundle-centralSupply.js';
+import mintHolder from '@agoric/vats/bundles/bundle-mintHolder.js';
+import provisionPool from '@agoric/vats/bundles/bundle-provisionPool.js';
+import priceAggregator from '@agoric/zoe/bundles/bundle-priceAggregator.js';
```
dckc: maybe not needed? bundle installed elsewhere?

packages/zoe/src/contracts/priceAggregator.js
```diff
@@ -33,7 +33,8 @@ export const INVITATION_MAKERS_DESC = 'oracle invitation';
 
 /**
  * This contract aggregates price values from a set of oracles and provides a
- * PriceAuthority for their median.
+ * PriceAuthority for their median. This naive method is game-able and so this module
+ * is a stub until we complete what is now in `priceAggregatorChain.js`.
```
dckc:
```suggestion
 * is a stub until we complete what is now in `priceAggregatorChainLink.js`.
```

packages/zoe/src/contracts/priceAggregatorTypes.js
```diff
@@ -79,7 +72,7 @@
 
 /**
  * @typedef {object} OracleReply
- * @property {any} reply
+ * @property {unknown} reply
```
dckc, line 75: :+1:

packages/zoe/src/zoeService/utils.d.ts
```diff
@@ -1,10 +1,5 @@
 import { Callable } from '@agoric/eventual-send';
-import type {
-  Instance,
-  IssuerKeywordRecord,
-  Payment,
-  StandardTerms,
```
dckc, line 6: why didn't lint catch this?

packages/zoe/test/unitTests/contracts/test-priceAggregator.js
```diff
@@ -31,6 +31,9 @@ import {
   parseRatio,
 } from '../../../src/contractSupport/ratio.js';
 
+// eslint-disable-next-line no-unused-vars -- type import
+import { start } from '../../../src/contracts/oracle.js';
+
```
dckc: vestigial?

dckc (commented):
all but the cli stuff...

dckc (approved):
We did more of a black-box review of the cli code. I watched @turadg do (local) e2e integration testing.

(I'm uncomfortable looking at code that doesn't adhere to ocap discipline.)

packages/zoe/src/zoeService/utils.d.ts
turadg: I looked into it and came up empty in the time I could allot. Since it's stripped out before runtime I think it's okay to let slide.

packages/vats/src/core/boot-oracles.js
turadg: moved to `inter-protocol` since it depends on its price feed proposal and under `scripts` since it's not used in any automated tests

packages/vats/src/core/boot-psm.js
```diff
@@ -210,6 +209,8 @@ export const buildRootObject = (vatPowers, vatParameters) => {
       // to code to be evaluated after initial bootstrap.
       bridgeCoreEval(powersFor('bridgeCoreEval')),
     ]);
+    // xxx this doesn't ever resolve yet, due to a dropped promise in startPSM (datalock)
+    console.log('boot-psm fully resolved');
```
dckc, line 213: I guess I wasn't clear; the datalock showed up when I tried to fix the fact that there's a dropped promise here. The code we shipped does not suffer from datalock, but it does use a dropped promise to escape from datalock.

The caller of this function doesn't get a promise for the completion of all the work that's started here, as one might expect. We rely on the kernel's ability to sense an empty promise queue at the end of a crank for things to work out right.

Does that log message really never show up?

dckc: ah... now I remember: returning the promise resulted in things getting stuck. So yes, it makes sense that this log message doesn't show up. I'm not sure where the datalock is, though.

dckc (approved):
moving the start scripts under `inter-protocol` makes sense.

Issue #6343, opened 2022-09-28
export non-ambient types in published packages


turadg:
## What is the Problem Being Solved?

Packages that consume agoric-sdk packages cannot make use of their types. In dapps we've resorted to copying them over or hand-writing them.

## Description of the Design

Packages published to NPM have TypeScript best-practices for types.

turadg:
Solved a while ago

PR #6500, opened 2022-10-26
export types

turadg:
refs: #6343

## Description

Emit types for agoric-sdk package builds like https://github.com/endojs/endo/pull/1307

### Security Considerations

n/a

Export types are a documentation resource.

### Testing Considerations

CI. And dry release?


packages/internal/src/utils.js
```diff
@@ -139,15 +144,20 @@ export const applyLabelingError = (func, args, label = undefined) => {
...
+  const left = String(a);
+  const right = String(b);
   // eslint-disable-next-line no-nested-ternary
   return left < right ? -1 : left > right ? 1 : 0;
 };
 
 /**
- * @param {object} obj
+ * @param {Record<string | symbol, unknown>} obj
  * @returns {(string|symbol)[]}
```
mhofman, line 161: Would be neat to have an `infer` based result here, but can be left as future improvement.

packages/internal/src/utils.js
```diff
@@ -241,6 +252,7 @@ export const deeplyFulfilledObject = obj => {
  * @returns {<T>(fn: () => Promise<T>) => Promise<{ result: T, duration: number }>}
  */
 export const makeMeasureSeconds = currentTimeMillisec => {
+  /** @param {() => any} fn */
```
mhofman, line 255: For this one we should be able to easy propagate through the result by templetizing here
```suggestion
  /**
   * @template {() => any} T
   * @param {T} fn
   */

File: packages/smart-wallet/src/walletFactory.js [hunk 3]
```diff
@@ -14,6 +14,11 @@ import { E } from '@endo/far';
 import { makeSmartWallet } from './smartWallet.js';
 import { shape } from './typeGuards.js';
 
+// Ambient types. Needed only for dev but this does a runtime import.
+import '@agoric/vats/exported.js';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import '@agoric/swingset-vat/src/types-ambient.js';
```
mhofman: I don't suppose we can avoid relying on ambient types for `swingset-vat`. I know it does re-export most types explicitly at the top level.

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -158,7 +158,7 @@ const FIRST_METER_ID = 1n;
 
 /**
  * @param {HostStore} hostStorage
- * @param {KernelSlog} kernelSlog
+ * @param {KernelSlog | null} kernelSlog
```
mhofman, line 161: Oh I have a stash with this fix and other swingset ones somewhere

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -67,7 +68,7 @@ export const reserveThenDeposit = async (
   console.info('confirmed deposit for', debugName);
 };
 
-/** @type {<T>(store: ERef<MapStore>, key: string, make: () => T) => Promise<T>} */
+/** @type {<T>(store: ERef<Map<string, T>>, key: string, make: () => T) => Promise<T>} */
```
mhofman: I'm surprised we have a `Map` based store around

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -90,6 +91,7 @@ export const makeInstallCache = async (
   { installCacheKey = 'installCache', loadBundle },
 ) => {
   /** @type {CopyMap<string, {installation: Installation, boardId: string, path?: string}>} */
+  // @ts-expect-error cast
```
mhofman: Please actually cast narrowly instead of ignoring all errors on the line.

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -108,6 +110,7 @@ export const makeInstallCache = async (
   };
 
   const wrapInstall = install => async (mPath, bPath, opts) => {
+    // @ts-expect-error xxx bundle types
```
mhofman: What is the issue here?

packages/vats/src/repl.js
```diff
@@ -97,6 +98,7 @@ function dump0(value, spaces, inProgress, depth) {
 
     const props = names
       .filter(n => nonNumber.has(n))
+      // @ts-expect-error
```
mhofman: cast instead

packages/vats/src/vat-provisioning.js
```diff
@@ -50,6 +51,7 @@ export function buildRootObject(_vatPowers) {
         // Update the notifier when the chainBundle resolves.
         const { notifier, updater } = makeNotifierKit();
         chainBundle.then(clientHome => {
+          // @ts-expect-error IterationObserver not callable
           updater(harden({ clientHome, clientAddress: address }));
```
mhofman: split in multiple lines and cast as appropriate?

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -8,13 +8,13 @@ export const DEPOSIT_FACET = 'depositFacet';
 const { details: X } = assert;
 
 /**
- * @param {ERef<NameAdmin>} nameAdmin
+ * @param {ERef<import('@agoric/vats/src/nameHub.js').NameAdmin>} nameAdmin
```
mhofman: How hard would it be to export types at the root of the `vats` package?

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -44,7 +44,7 @@ export const reserveThenGetNamePaths = async (nameAdmin, paths) => {
 };
 
 /**
- * @param {ERef<NameAdmin>} nameAdmin
+ * @param {ERef<import('@agoric/vats/src/types.js').NameAdmin>} nameAdmin
```
mhofman: Yeah definitely would like to avoid deep imports here.

packages/vats/src/types.js
```diff
@@ -1,5 +1,7 @@
 // @ts-check
 
+export {};
```
mhofman, line 3: Looks like this is now exported types, so it should be easy to re-export from `index.js`

mhofman (commented):
Nice start. Not an approval until I can get some clarifications on the various `ts-expect-error`.

I'd like to at least double check the need for some of these `ts-expect-error`s. As I've expressed before, I believe they are worse than casts as they can hide multiple errors unexpectedly. If possible narrow application of casts are better (while slightly harder to read unfortunately).

Also I see there was an attempt to switch to exported types, but it was reverted. If attempted again, let's avoid deep imports cross package.

I have some type fixes for swingset that I may now try to brush off :)

packages/internal/src/utils.js
turadg: it's correctly inferred now. I don't see the value in removing the `returns` but open to it.

packages/internal/src/utils.js
turadg: What different will it make? `makeMeasureSeconds` defines the type already.

packages/inter-protocol/src/proposals/utils.js
turadg: > Property 'endoZipBase64Sha512' does not exist on type 'EndoZipBase64Bundle | GetExportBundle | NestedEvaluateBundle'.

packages/internal/src/utils.js
mhofman: Right, sorry habit of letting inference flow to the return types and not use explicit `@returns`.

packages/vats/src/vat-provisioning.js
turadg: `updater` itself has the error,
> This expression is not callable.
>  Type 'IterationObserver<any>' has no call signatures.

I've pushed what I think is a fix

packages/inter-protocol/src/proposals/utils.js
turadg: good call. in the course of the PR they were exported but these imports needed updating.

mhofman:
Nit on reviewing process: While I understand this ~is~ was still a draft PR, it's extremely difficult to track changes compared to the previously reviewed state if force pushes include both amendments and base changes. In general please do not amend, and stick to commits additions once reviews have started, and until the PR is accepted. New commits can be marked as `fixup` to indicate such amendments are to be made just before merging only. Furthermore when changing the base, please do 2 force push: one push adding commits on the old base as is, and one force push rebasing showing no changes (or only conflict resolution if warranted).

Another point specific to this PR: please include the main package name affected in the conventional commit prefix, e.g. `chore(ERTP)`. Right now there are commits with duplicate names making it impossible to understand what's going on.

I really want type improvements, but right now this PR is extremely difficult to review.

packages/inter-protocol/src/proposals/utils.js
mhofman: That seems like a legitimate issue. Why suppress it?
Or is it because of an overload in `loadBundle` which isn't typed correctly?

packages/internal/src/utils.js
mhofman: I meant with some work we could type this utility function so that the it would actually have as return type the subset of the `obj` keys for which the values are functions. Aka a type implementation of the behavior.

Sure `(string|symbol)[]` is correct, but it's fairly broad.

turadg:
I agree with the amendment policy for PRs marked R4R. I don't for PRs that aren't. This is something our eng org should agree on so let's discuss in standup.

> Another point specific to this PR: please include the main package name affected in the conventional commit prefix, e.g. chore(ERTP).

I've opted against that because IIUC it creates redundant info in READMEs. e.g. the `@agoric/ertp` changelog says "chore(ertp): foo".

> Right now there are commits with duplicate names making it impossible to understand what's going on.

I'm confused on why omission of package name makes it impossible to understand what's going on. If I had not separated the commits and instead, for example, squashed all the `chore: misc types fixes` then there would be no duplicates but I expect it would be harder to follow the changes.

> I really want type improvements, but right now this PR is extremely difficult to review.

All the changes since your review of the draft were addressing your comments. So maybe it suffices to look over your comments and see whether they were addressed to your satisfaction.

packages/smart-wallet/package.json
```diff
@@ -32,6 +34,7 @@
     "@agoric/nat": "^4.1.0",
     "@agoric/notifier": "^0.5.1",
     "@agoric/store": "^0.8.3",
+    "@agoric/swingset-vat": "^0.30.2",
```
mhofman, line 37: Ah yeah unfortunate that we have to make this a non-dev dependency.

Wondering how we're gonna audit and remove these dependencies once we switch to imported types, and away from runtime `import` statements for ambient types.

packages/inter-protocol/src/proposals/utils.js
mhofman: `provideWhen` seem to be used once with a `ERef<MapStore>` and once with a `Map`. The semantics of these Maps are different enough, which I believe does reflect on their signature. If `provideWhen` is in fact using the common subset of these 2, can we be explicit about this somehow? Since TS is structural typing only, maybe by defining an interface of the subset used? That would avoid a cast / expect-error in the place where `provideWhen` is used with the `ERef<MapStore>`.

packages/vats/src/core/boot.js
```diff
@@ -17,6 +17,13 @@ import * as behaviors from './behaviors.js';
 import * as clientBehaviors from './client-behaviors.js';
 import * as utils from './utils.js';
 
+// Ambient types
+import './types.js';
+// eslint-disable-next-line import/export -- no named exports
+export * from '../types.js';
+export * from '../nameHub.js';
+export * from '../bridge.js';
```
mhofman: These see, to be re-exporting non-type exports as well, is that intended?

If so please visually separate them. In general I prefer having runtime imports before type based ones.

mhofman: I suspect the `./bridge.js` import is for the `BridgeManager` type? In that case I believe the better approach would be to have a JSDoc `@typedef` in `../types.js` and leverage the existing re-export from there. Same the `nameHub.js`'s `NameHub` and `NameAdmin`.

packages/internal/src/utils.js
turadg: Ah got it. I think we'd run into this: https://stackoverflow.com/a/55012175/46040 (found from `https://github.com/microsoft/TypeScript/issues/41036`)

mhofman: We don't really deal with subtypes in our code base, but yeah it'd definitely be safer with exact types, if TS had those.

packages/SwingSet/tools/internal-types.js
```diff
@@ -1,13 +0,0 @@
-/**
- * @typedef {object} ManualTimerAdmin
- * @property { (when: Timestamp) => void } advanceTo
- */
-
-/**
- * @typedef {ManualTimerAdmin & TimerService} ManualTimer
- */
```
mhofman, line 8: There seems to be a definitions in Zoe for the same type names, is that what you mean by "deduping"? Those do appear to have a different definition somehow.

packages/zoe/src/internal-types.js
```diff
@@ -241,7 +241,7 @@
  * promise will fulfill to the completion value.
  * @property {ShutdownWithFailure} terminateWithFailure
  * Terminate the vat in which the contract is running as a failure.
- * @property {(bundleCap: BundleCap, options?: Record<string, any>) => Promise<RootAndAdminNode>} upgrade
+ * @property {(bundleCap: *, options?: Record<string, any>) => Promise<RootAndAdminNode>} upgrade
```
mhofman, line 244: Why the loosening here?

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -90,8 +91,12 @@ export const makeInstallCache = async (
   { installCacheKey = 'installCache', loadBundle },
 ) => {
   /** @type {CopyMap<string, {installation: Installation, boardId: string, path?: string}>} */
-  const initial = await provideWhen(E.get(homeP).scratch, installCacheKey, () =>
-    makeCopyMap([]),
+
+  const initial = await provideWhen(
+    // @ts-expect-error cast
+    E.get(homeP).scratch,
```
mhofman: While I mentioned above I'd prefer to figure out the `provideWhen`'s param type to the compatible subtype, is there a reason to use `// @ts-expect-error` for casts instead of doing an actual cast ?

Do we consider casting to be too visually onerous?

packages/vats/src/vat-provisioning.js
```diff
@@ -50,7 +51,7 @@ export function buildRootObject(_vatPowers) {
         // Update the notifier when the chainBundle resolves.
         const { notifier, updater } = makeNotifierKit();
         chainBundle.then(clientHome => {
-          updater(harden({ clientHome, clientAddress: address }));
+          updater.updateState(harden({ clientHome, clientAddress: address }));
```
mhofman: Fix LGTM, but I should probably defer to someone more familiar with this code and notifiers. I believe this is the only behavioral difference in this PR.

packages/vats/src/types.js
mhofman: This file is currently still referenced by `vats/exported.js` using `import './src/types.js';`, which is no longer relevant.

Have we also confirmed there were no use of the typedefs in this file that may have silently rolled back to an implicit `any` type?

packages/vats/src/core/boot.js
```diff
@@ -17,6 +17,13 @@ import * as behaviors from './behaviors.js';
 import * as clientBehaviors from './client-behaviors.js';
 import * as utils from './utils.js';
 
+// Ambient types
+import './types.js';
+// eslint-disable-next-line import/export -- no named exports
+export * from '../types.js';
```
mhofman: I just realized this file is the entrypoint of the package. It's definitely a little confusing.

Would you mind introducing an `index.js` at the root which `exports *` from both this file and `./src/types.js` ?

mhofman (commented):
I'd like to do another pass, but here is my review so far.

I still have some concerns regarding casts and type compatibility.

I believe the conversion of the `vats` package to exported type introduces some mistaken re-exports, and could use a better layering for the main entrypoint.

packages/zoe/src/internal-types.js
turadg: It's a type in name only.
https://github.com/Agoric/agoric-sdk/blob/8980a2d57e54d635f1d85b9d80b62ca5046b4f43/packages/SwingSet/src/types-external.js#L353

This was expedient to work around an import.

packages/vats/src/vat-provisioning.js
turadg: Yep, the only intended behavior change. `updater` comes out of `makeNotifierKit` which just has `updateState`, `fail` and `finish`. The comment says "Update the notifier" so I'm sure it's not meant to be one of the latter two.

FYI @michaelfig who I think wrote this.

packages/smart-wallet/package.json
turadg: Ticket here https://github.com/Agoric/agoric-sdk/issues/6512

packages/inter-protocol/src/proposals/utils.js
turadg: Good idea. Done.

packages/inter-protocol/src/proposals/utils.js
turadg: The type was wrong (which is what I meant by "xxx bundle types")

e6c5a934a5ab6af203151336c2eeb22aa096faf1 has a fix. It shouldn't change behavior but technically is a runtime change.

packages/vats/src/types.js
turadg: > This file is currently still referenced by vats/exported.js using import './src/types.js';, which is no longer relevant.

Good catch. Removed.

> Have we also confirmed there were no use of the typedefs in this file that may have silently rolled back to an implicit any type?

Good idea. Found some (though I don't know if they were resolving before).

packages/SwingSet/tools/internal-types.js
turadg: I believe the differences were tech debt.

packages/inter-protocol/src/proposals/utils.js
turadg: Moot with 570cc993f3cfb2eeff49a363fd10b497b100079c, but I do find casting in JSdoc to be hard to read. No strong feeling one way or the other.

packages/vats/src/core/boot.js
turadg: Good idea

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -68,7 +68,7 @@ export const reserveThenDeposit = async (
   console.info('confirmed deposit for', debugName);
 };
 
-/** @type {<T>(store: ERef<Map<string, T>>, key: string, make: () => T) => Promise<T>} */
+/** @type {<T>(store: ERef<Pick<MapStore<string, T>, 'get' | 'set'>>, key: string, make: () => T) => Promise<T>} */
 const provideWhen = async (store, key, make) => {
   const found = await E(store).get(key);
```
mhofman, line 73: I probably should have looked at the actual implementation before.

I was under the impression that `MapStore` threw when doing a `get` of an uninitialized value. That's what I meant earlier by different semantics, unfortunately the signature of `get` methods only varies on the return type between `Map` and `MapStore`, making them "compatible". Thinking more about this, it's probably a footgun that these objects have such a similar shape, yet such different semantics.

Anyway, I think `provideWhen` is faulty because of this, and I don't understand how the fallback path would have every worked with the `MapStore` variant.

packages/SwingSet/src/types-external.js
```diff
@@ -351,7 +351,7 @@ export {};
  *
  * @typedef { string } BundleID
  * @typedef {*} BundleCap
- * @typedef { { moduleFormat: 'endoZipBase64', endoZipBase64: string, endoZipBase64Sha512 } } EndoZipBase64Bundle
+ * @typedef { { moduleFormat: 'endoZipBase64', endoZipBase64: string, endoZipBase64Sha512: string } } EndoZipBase64Bundle
```
mhofman, line 354: Oh yeah that would do it!

packages/vats/index.js
```diff
@@ -0,0 +1,9 @@
+// @ts-check
+// Ambient types
+import './src/core/types.js';
+
+export * from './src/core/boot.js';
+// eslint-disable-next-line import/export -- no named exports
+export * from './src/types.js';
+export * from './src/nameHub.js';
+export * from './src/bridge.js';
```
mhofman, line 9: I still believe this will cause formerly internal runtime helpers to now be visible on the root of this package, not only types.

See https://github.com/Agoric/agoric-sdk/pull/6500#discussion_r1008709131

packages/inter-protocol/src/proposals/utils.js
mhofman: I believe runtime changes are worth when they directly address issues found by typing. improvements

packages/inter-protocol/src/proposals/utils.js
turadg: Agree with your assessment. Is there a requested change in there? An issue to make or a thread to start?

packages/vats/index.js
turadg: Thanks for the thorough review!

> I suspect the ./bridge.js import is for the BridgeManager type? In that case I believe the better approach would be to have a JSDoc @typedef in ../types.js and leverage the existing re-export from there. Same the nameHub.js's NameHub and NameAdmin.

`makeNameHubAdmin` is imported in three other modules, so arguably this top level is an improvement, but I'll err towards conservative and leave runtime exports refactoring for another day.

packages/inter-protocol/src/proposals/utils.js
mhofman: I don't know. I don't think it'd be easy enough to fix now, but at least we should create an issue?

I think reverting the `Pick` and adding a `ts-expect-error` pointing to the issue?

turadg: I made an issue but then realized a simpler fix: [bf9b05f](https://github.com/Agoric/agoric-sdk/pull/6500/commits/bf9b05ffef83d26a98df53953ee231bca637ae5b)

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -81,7 +81,7 @@ const provideWhen = async (store, key, make) => {
 
 /**
  *
- * @param {{ scratch: ERef<MapStore<string, CopyTagged>> }} homeP
+ * @param {{ scratch: ERef<Map<string, CopyTagged>> }} homeP
```
mhofman: Does `scratch` actually have `Map` semantics? If not that'd be a very unobvious cast that hide the runtime differences.

packages/inter-protocol/src/proposals/utils.js
mhofman: I fail to see how it's a fix?

packages/SwingSet/src/vats/plugin-manager.js
```diff
@@ -21,9 +21,11 @@ const DEFAULT_WALKER = Far('walker', { walk: pluginRootP => pluginRootP });
 
 /**
  * @template T
- * @typedef {T} Device
+ * @typedef {'Device' & { __deviceType__: T }} Device
```
mhofman, line 24: 🎉

packages/agoric-cli/src/init.js
```diff
@@ -2,8 +2,10 @@ import chalk from 'chalk';
 import { makePspawn } from './helpers.js';
 
 // Ambient types. Needed only for dev but this does a runtime import.
+// https://github.com/Agoric/agoric-sdk/issues/6512
 import '@endo/captp/src/types.js';
 import '@agoric/swingset-vat/exported.js';
+import '@agoric/swingset-vat/src/vats/network/types.js';
```
mhofman, line 8: Seeing this now, would it make sense to add `import './src/vats/network/types.js';` to `@agoric/swingset-vat/exported.js` instead ?

Edit: maybe not since it's used directly in other places?

mhofman (commented):
I'm not sure I understand how the `provideWhen` type change is actually a fix.

packages/inter-protocol/src/proposals/utils.js
turadg: Yes, it does https://github.com/Agoric/agoric-sdk/blob/bf9b05ffef83d26a98df53953ee231bca637ae5b/packages/solo/src/scratch.js#L17-L20

packages/agoric-cli/src/init.js
turadg: yeah, I considered that and opted for lesser change to what's exported. I figure it will be addressed in further ambients work

mhofman (approved):
LGTM

dckc:
I would have like to see more coordination around getting rid of `@ts-check`. It's a useful explicit tag, especially when moving code between repositories.

turadg:
> I would have like to see more coordination around getting rid of @ts-check. It's a useful explicit tag, especially when moving code between repositories.

@dckc it's an easy change to reverse. You want every JS file in a package to have `@ts-check` even if it's redundant with the package configuration?

dckc:
> You want every JS file in a package to have @ts-check even if it's redundant with the package configuration?

That would have been my preference, yes. At this point, I don't feel strongly that it's worth more churn in the git history. But it seems like a change to house style that's worth fairly widespread coordination.

PR #5997, opened 2022-08-18
make Zoe durable

Chris-Hibbert:
closes: #4383
refs: #6502

## Description

Make Zoe durable.

The prior version passed around many bundles of functions. Those had to be converted to durable facets to be passable.

### Security Considerations

Pervasive

### Documentation Considerations

seat notifiers (which notified on interim allocations) have been dropped and replaced with a subscriber that only notifies when the seat exits.

The way feeMintAccess is obtained changed.

The methods that allowed requests for internal allocation state (`getCurrentAllocationJig`) has been dropped.  Many of the clients can use `getFinalAllocation()` instead.

`zoe.bindDefaultFeePurse`, `zoe.makeFeePurse`, and `zcf.assert()` were  dropped.

### Testing Considerations

Test timing changed in some cases, since Promises can no longer be used.

PR #6502, opened 2022-10-27
feat(notifier): Introduce durable publish kits

### LINKED ISSUE #4567: build durable notifier

gibson042:
closes: #4567

Layered on #6627

## Description

The primary new interface is `vivifyDurablePublishKitMaker(baggage, kindName)`, which returns a factory for constructing durable objects with a { publish, finish, fail } `publisher` facet and a { subscribeAfter } `subscriber` facet (matching the publish kit interface).

### Security Considerations

None known

To be determined.

### Testing Considerations

Upgrade trauma is covered with a rather heavy-handed use of a complete SwingSet-and-vat environment.


packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +186,209 @@ export const makePublishKit = () => {
...
+      }),
+    );
+  }
+};
+
+/**
+ * @param {import('../../vat-data/src/types.js').Baggage} baggage
+ * @param {string} kindName
+ */
+export const vivifyDurablePublishKitMaker = (baggage, kindName) => {
```
Chris-Hibbert: I prefer `vivifyDurablePublishKit`.   With `vivifyFooMaker` I have trouble telling whether it's vivifying a maker, or a maker of vivifyFoos.  I've gotten used to the idea that `vivifyFoo` and `vivifyFooKit` return makers, and `vivifyFooInstance` returns an instance.

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +186,209 @@ export const makePublishKit = () => {
...
+  return vivifyKindMulti(baggage, kindName, initDurablePublishKitState, {
```
Chris-Hibbert: Why not `vivifyFarClassKit`?   The typeGuards turn out to be pretty handy.

gibson042: Ah, I think I missed that when I was looking for a multi-faceted interface.

packages/notifier/src/publish-kit.js
gibson042: So concretely, are you asking for something that supports the following?
```js
const makeDurablePublishKit = vivifyDurablePublishKit(baggage, 'publish kit');
const { publisher: publisher1, subscriber: subscriber1 } = makeDurablePublishKit();
const { publisher: publisher2, subscriber: subscriber2 } = makeDurablePublishKit();
```

Chris-Hibbert: Yeah. That's what I was looking for.

gibson042: Done.

packages/notifier/test/test-publish-kit.js
```diff
@@ -1,275 +1,310 @@
...
+    'DurablePublishKit',
+  );
+  const kit1 = makeDurablePublishKit();
+  const { publisher: pub1, subscriber: sub1 } = kit1;
+  const value = Symbol.for('value');
+  await assertTransmission(t, kit1, value);
+  // THEN A MIRACLE OCCURS...
+  // @ts-expect-error
+  // eslint-disable-next-line no-undef
+  const kit2 = recoverPublishKit(baggage);
```
gibson042, line 388: I could really use help here. It looks like [zoe/test/swingsetTests/upgradeCoveredCall](https://github.com/Agoric/agoric-sdk/blob/94625d38c3bb5333b00a69dd3086b1ac13490f62/packages/zoe/test/swingsetTests/upgradeCoveredCall/test-coveredCall-service-upgrade.js) pulls off something analogous, but it does so at the vat level, which seems much too far up the stack for this.

[prepare-test-env-ava.js](https://github.com/Agoric/agoric-sdk/blob/b43e34bd41b45de0d4771f4ade276e4b05bbffd0/packages/SwingSet/tools/prepare-test-env-ava.js) exports a globalThis.VatData populated from `{ vom, cm, wpm } = makeFakeVirtualStuff(…)` in [prepare-test-env.js](https://github.com/Agoric/agoric-sdk/blob/b43e34bd41b45de0d4771f4ade276e4b05bbffd0/packages/SwingSet/tools/prepare-test-env.js), and I have a sense that I should be doing something with [fakeVirtualSupport.js](https://github.com/Agoric/agoric-sdk/blob/b43e34bd41b45de0d4771f4ade276e4b05bbffd0/packages/SwingSet/tools/fakeVirtualSupport.js) to replace vrm and/or vom and/or crm (getting objects with new identity from reloaded durable state)—and probably an an independent test file to avoid corruption from manipulating the VatData global—but I can't see how to do it.

Chris-Hibbert: With `upgradeCoveredCall`, we wanted to upgrade the contract in a clean environment, so we could be sure there wasn't any leakage of object references from one instance to the next. 

I think for the notifier upgrade, what I'd expect to see is exercising the notifier in one vat, then allowing it to shut down. Then in a second vat, reviving the notifier, and seeing that it could still be used by its original owner, and  previously existing clients would be able to see the new updates.

The first `run()` call in `test-coveredCall-service-upgrade` does the first part by invoking the `buildV1` clause in `bootstrap-coveredCall-service-upgrade`, which starts the `v2BundleId` installation of coveredCall.  The second `run()` call invokes `upgradeV2`, which uses the adminNode from the first vat and explicitly upgrades it.

When the upgrade calls `vivifyDurablePublishKit` in a new vat, the effect is to redefine the behavior of the publishKit and associate it with objects stored in the baggage with the same kind identifier. I don't think its allowed to re-associate new behavior with an existing kind identifier, so this has to be done in a new vat with access to the existing baggage.

packages/notifier/src/publish-kit.js
```diff
@@ -102,6 +111,19 @@ export const subscribeLatest = subscriber => {
...
+  finish: M.call(M.any()).returns(),
+  fail: M.call(M.any()).returns(),
+});
+const SubscriberI = M.interface('Subscriber', {
+  subscribeAfter: M.call().optional(M.bigint()).returns(M.promise()),
+});
+const publishKitIKit = harden({
+  publisher: PublisherI,
+  subscriber: SubscriberI,
+});
```
Chris-Hibbert: Please export these.

gibson042: Done.

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,218 @@ export const makePublishKit = () => {
...
+ * @param {any} [rejection]
+ */
+const advanceDurablePublishKit = (state, done, value, rejection) => {
+  const { valueDurability, status } = state;
+  if (status !== 'live') {
+    throw new Error('Cannot update state after termination.');
+  }
+  if (done || valueDurability === 'mandatory') {
+    assertPassable(value);
+    assertPassable(rejection);
```
gibson042: Is there a more accurate test of object durability? I'm pretty sure this will incorrectly allow through some values, such as promises.

Chris-Hibbert: `canBeDurable(value)`

packages/notifier/test/vat-integration/vat-pubsub.js
```diff
@@ -0,0 +1,27 @@
...
+  const makeDurablePublishKit = vivifyDurablePublishKit(
+    baggage,
+    'DurablePublishKit',
+  );
+  // TODO: Figure out why this breaks during upgrade!
+  const { publisher, subscriber } = provide(
+    baggage,
+    'publishKitSingleton',
+    () => makeDurablePublishKit(),
+  );
```
gibson042: I'm so sorry to take off with this incomplete, but I don't have enough gas left in the tank to figure out why it doesn't work as expected. The error is coming from [`makeRepresentative`](https://github.com/Agoric/agoric-sdk/blob/94625d38c3bb5333b00a69dd3086b1ac13490f62/packages/SwingSet/src/liveslots/virtualObjectManager.js#L694) during `reanimate` because `innerSelf` is a string rather than an object, and it occurs whether I use `provide` or manually invoke `baggage.get` after a conditional `baggage.init`. I'm almost certain that I'm just trying to do something a little incorrectly.

<details><summary><code>yarn test test/test-publish-kit.js</code></summary>

```
error during vat dispatch() of startVat,[object Object] (TypeError#1)
TypeError#1: Object.getOwnPropertyNames: invalid object
TypeError: Object.getOwnPropertyNames: invalid object
 at getOwnPropertyNames ()
 at makeRepresentative (/bundled-source/.../packages/SwingSet/src/liveslots/virtualObjectManager.js:694)
 at reanimate (/bundled-source/.../packages/SwingSet/src/liveslots/virtualObjectManager.js:770)
 at convertSlotToVal (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:771)
 at decodeSlotCommon (/bundled-source/.../node_modules/@endo/marshal/src/marshal.js:254)
 at decodeFromCapData (/bundled-source/.../node_modules/@endo/marshal/src/encodeToCapData.js:414)
 at decodeEntry (/bundled-source/.../node_modules/@endo/marshal/src/encodeToCapData.js:483)
 at map ()
 at decodeFromCapData (/bundled-source/.../node_modules/@endo/marshal/src/encodeToCapData.js:485)
 at unserialize (/bundled-source/.../node_modules/@endo/marshal/src/marshal.js:359)
 at runWithoutMetering (/bundled-source/.../packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js:52)
 at buildRootObject (.../notifier/test/vat-integration/vat-pubsub.js:14)
 at startVat (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:1418)
 at ()
 at ()

UnhandledPromiseRejectionWarning: (TypeError#1)
Logging sent error stack (RemoteError#1)
RemoteError#1: vat-upgrade failure
Error: vat-upgrade failure
 at construct ()
 at Error (/bundled-source/.../node_modules/ses/src/error/tame-error-constructor.js:56)
 at makeError (/bundled-source/.../node_modules/ses/src/error/assert.js:242)
 at decodeErrorCommon (/bundled-source/.../node_modules/@endo/marshal/src/marshal.js:284)
 at decodeFromCapData (/bundled-source/.../node_modules/@endo/marshal/src/encodeToCapData.js:425)
 at map ()
 at map ()
 at unserialize (/bundled-source/.../node_modules/@endo/marshal/src/marshal.js:359)
 at deliver (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:1053)
 at dispatchToUserspace (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:1452)
 at runWithoutMetering (/bundled-source/.../packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js:52)
 at ()

RemoteError#1 ERROR_NOTE: Sent as error:liveSlots:v2#70001
Logging sent error stack (RemoteError(error:liveSlots:v2#70001)#1)
RemoteError(error:liveSlots:v2#70001)#1: vat-upgrade failure
Error: vat-upgrade failure
 at construct ()
 at Error (/bundled-source/.../node_modules/ses/src/error/tame-error-constructor.js:56)
 at makeError (/bundled-source/.../node_modules/ses/src/error/assert.js:242)
 at decodeErrorCommon (/bundled-source/.../node_modules/@endo/marshal/src/marshal.js:284)
 at decodeFromCapData (/bundled-source/.../node_modules/@endo/marshal/src/encodeToCapData.js:425)
 at unserialize (/bundled-source/.../node_modules/@endo/marshal/src/marshal.js:359)
 at notifyOnePromise (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:1189)
 at notify (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:1213)
 at dispatchToUserspace (/bundled-source/.../packages/SwingSet/src/liveslots/liveslots.js:1457)
 at runWithoutMetering (/bundled-source/.../packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js:52)
 at ()

RemoteError(error:liveSlots:v2#70001)#1 ERROR_NOTE: Sent as error:liveSlots:v1#70001
REJECTED from ava test: {
  body: '{"@qclass":"error","errorId":"error:liveSlots:v1#70001","message":"vat-upgrade failure","name":"Error"}',
  slots: []
}
  ✖ durable publish kit upgrade trauma (full-vat integration) Rejected promise returned by test
  ─

  durable publish kit upgrade trauma (full-vat integration)

  Rejected promise returned by test. Reason:

  {
    body: '{"@qclass":"error","errorId":"error:liveSlots:v1#70001","message":"vat-upgrade failure","name":"Error"}',
    slots: [],
  }
```

</details>

warner:
@FUDCo and I walked through some parts of the code, to try and get up to speed for a review. A few notes (not an actual review yet):

* enhancing swingset's vat-upgrade to re-return the root object: the implementation looks good, the `queueToKref()` or `queueToVatRoot()` will take care of adding a new refcount to the upgraded vat's root object. The changes will mostly be obviated by Chip's "smallcaps" PR (#6326), or at least they'll be a lot simpler. But we don't yet understand why the feature was necessary: it implements a "TODO: maybe" from the kernel code, but we don't yet see where the new feature gets used, and we're always inclined to minimize changes unless there's a compelling reason.
* The function named `vivifyDurablePublishKit` was confusingly named to us. Reading the conversation you had with @Chris-Hibbert , it looks like we had the opposite opinion: `vivifyDurablePublishKit` looks (to us) like something that returns you a `DurablePublishKit`, when in fact it (correctly) returns a function that *makes* durable publish kits. But I see Chris' issues too.. I think our naming patterns suffer from some "is it left-associative or right-associative" ambiguities, or we wish we could insert disambiguating parentheses into the function names, or something.
* Whatever it gets named, in an ideal world, `vivifyDurablePublishKit()` would be called exactly once per vat. In the context of ZCF and contracts, we think that ZCF should call it, and provide the resulting `makeDurablePublishKit` function to contract code (as an argument of `start()` or whatever the boundary between ZCF and contract is now). Or, it should wrap `makeDurablePublishKit` in a function that calls `provide`, which might be named (the "real") `vivifyDurablePublishKit(baggage, name) -> DurablePublishKit`: something contract code can call to get-or-create a specific DurablePublishKit.
  * If ZCF does *not* build the maker on behalf of contract code, my fear is that contract code will leave it up to (each of) some set of libraries to build themselves, and as a result we'll wind up with dozens of functionally-identical Kinds in each vat, which is just a waste of DB space. We've discussed some aspects of this concern before: the difficulty of sharing low/no-authority resources among uncoordinated libraries when you can't define/construct them at the top-level (module scope) because they need to be vivified from `baggage`.

We didn't get far enough to look at the implementation yet. We talked a bit about testing: he and I are pretty comfortable with writing up a "real-kernel" swingset-based test for things like this, the kind where you run a kernel to completion, the bootstrap vat spins up two sequential incarnations of a test vat, the test vat uses `vivify` and `provide` and friends to create a thing in the first incarnation and then make sure it's still usable in the second incarnation, the bootstrap vat writes a "yes it worked" message to `vatPowers.log`, and then the test program reads out the test log and makes sure the "yes it worked" message appears in the output.

The downsides of this sort of "integration-y" test are:
* slower: at least 5s for the bundles to be created, then usually 1-2s to launch the kernel (and doing a vat upgrade will probably cost another half second)
* debuggers are harder to use: all of the code you're interested in gets bundled and then evaluated on the fly, so you can't pre-place a breakpoint (you usually have to add a `debugger;` statement, wait for `importBundle()` to hit it, and only then can you add the real breakpoints), and generally the devtools don't know how to deal with code that gets run this way
* the AVA-side test assertions (`t.is`, `t.deepEqual`) don't get any internal state to work with, you typically wind up with a single `t.deepEquals` assertion against a "golden master" list of expected log messages inside a vat. You effectively have to write all your tests inside an unfriendly environment in the form of:

```js
  vatPowers.log(`does result 4 match?: ${res4 === expectedRes4}`);
```

which is super-annoying.

We know we want developers to have better testing tools. We spoke briefly about what those might look like, I know the structure should be something like:

```js
test('test publisher under upgrade', async t => {
  const controlObject = prepareTest();
  runVersionOne(t);
  controlObject.simulateUpgrade();
  runVersionTwo(t);
});
```

where you write `runVersionOne/Two`, but it'll take some more brainstorming to figure out something better. A big challenge is that this test environment needs to hook the `VatData` tools like `defineKind`, but we provide those as imports. We have some "fakeVirtualObject" support code already, which does at least some portion of this, so the right path probably lies in that direction.

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+  const publishCount = state.publishCount;
+  let tailP;
+  let tailR;
+
+  if (done) {
+    state.status = rejection ? 'failed' : 'finished';
+    tailP = makeQuietRejection(new Error('Cannot read past end of iteration.'));
+    tailR = undefined;
+  } else {
+    ({ promise: tailP, resolve: tailR } = makePromiseKit());
```
Chris-Hibbert, line 347: We added one here, too.
```
    tailP.catch(sink);
```

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+          publishCount,
+          tail: tailP,
+        }),
+      ),
+      tailP,
+      tailR: undefined,
+    };
+  } else if (status === 'live') {
+    const { promise: tailP, resolve: tailR } = makePromiseKit();
+    newData = {
```
Chris-Hibbert, line 295: I had problems with `unhandled rejections`,  and @mhofman tracked it down to here and recommended adding
```
    tailP.catch(sink);
```
I suspect you'll do something with `makeQuiteRejection()` instead.  I can provide a test cases if you need it.

gibson042:
> @FUDCo and I walked through some parts of the code, to try and get up to speed for a review. A few notes (not an actual review yet):

I'll provide a similarly informal out-of-office response.

> * enhancing swingset's vat-upgrade to re-return the root object: the implementation looks good, the `queueToKref()` or `queueToVatRoot()` will take care of adding a new refcount to the upgraded vat's root object. The changes will mostly be obviated by Chip's "smallcaps" PR ([Convert kernel/liveslots to use smallcaps #6326](https://github.com/Agoric/agoric-sdk/issues/6326)), or at least they'll be a lot simpler. But we don't yet understand why the feature was necessary: it implements a "TODO: maybe" from the kernel code, but we don't yet see where the new feature gets used, and we're always inclined to minimize changes unless there's a compelling reason.

In short, the test needs a reference to the new root object in order to send it messages and I couldn't find another way (but it wouldn't surprise me to find out that it exists and I just missed it).

> * The function named `vivifyDurablePublishKit` was confusingly named to us. Reading the conversation you had with @Chris-Hibbert , it looks like we had the opposite opinion: `vivifyDurablePublishKit` looks (to us) like something that returns you a `DurablePublishKit`, when in fact it (correctly) returns a function that _makes_ durable publish kits. But I see Chris' issues too.. I think our naming patterns suffer from some "is it left-associative or right-associative" ambiguities, or we wish we could insert disambiguating parentheses into the function names, or something.

/me nods. But we do need to pick _something_...

> * Whatever it gets named, in an ideal world, `vivifyDurablePublishKit()` would be called exactly once per vat. In the context of ZCF and contracts, we think that ZCF should call it, and provide the resulting `makeDurablePublishKit` function to contract code (as an argument of `start()` or whatever the boundary between ZCF and contract is now). Or, it should wrap `makeDurablePublishKit` in a function that calls `provide`, which might be named (the "real") `vivifyDurablePublishKit(baggage, name) -> DurablePublishKit`: something contract code can call to get-or-create a specific DurablePublishKit.

Agreed (and following supporting commentary therefore elided here).

> We didn't get far enough to look at the implementation yet. We talked a bit about testing: he and I are pretty comfortable with writing up a "real-kernel" swingset-based test for things like this, the kind where you run a kernel to completion, the bootstrap vat spins up two sequential incarnations of a test vat, the test vat uses `vivify` and `provide` and friends to create a thing in the first incarnation and then make sure it's still usable in the second incarnation, the bootstrap vat writes a "yes it worked" message to `vatPowers.log`, and then the test program reads out the test log and makes sure the "yes it worked" message appears in the output.

I think the current (broken) test is close to that, but I'm not sure how to push it closer.

> The downsides of this sort of "integration-y" test are [understood].

> We know we want developers to have better testing tools. We spoke briefly about what those might look like, I know the structure should be something like:
> 
> ```js
> test('test publisher under upgrade', async t => {
>   const controlObject = prepareTest();
>   runVersionOne(t);
>   controlObject.simulateUpgrade();
>   runVersionTwo(t);
> });
> ```
> 
> where you write `runVersionOne/Two`, but it'll take some more brainstorming to figure out something better. A big challenge is that this test environment needs to hook the `VatData` tools like `defineKind`, but we provide those as imports. We have some "fakeVirtualObject" support code already, which does at least some portion of this, so the right path probably lies in that direction.

Absolutely. I was hoping to get far enough in that direction here, but ultimately ran out of time.

packages/SwingSet/src/controller/controller.js
```diff
@@ -466,10 +466,14 @@ export async function makeSwingsetController(
       resultPolicy = 'ignore',
       slots = [],
     ) {
+      // TODO: Use a proper makeMarshal serialize.
```
erights: @FUDCo , does your smallcaps PR do this todo? 

@gibson042 @FUDCo is it clear what the merge of this change and @FUDCo 's upcoming smallcaps change should be? Could that be adopted here by itself even if this PR gets merged before the smallcaps PR does?

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -936,11 +936,22 @@ export default function buildKernel(
       return results;
     }
 
-    const args = [upgradeID, true, undefined, incarnationNumber];
-    const vatAdminMethargs = {
-      body: JSON.stringify(['vatUpgradeCallback', args]),
-      slots: [],
-    };
+    // TODO: Use a real makeMarshal serialize?
```
erights: Same question. The sooner we get rid of non-local knowledge of serialization formats, the better. It would also make this code easier to understand and review.

packages/notifier/src/publish-kit.js
```diff
@@ -4,10 +4,19 @@
...
+import { canBeDurable, vivifyFarClassKit } from '@agoric/vat-data';
 
 import './types.js';
 
+const { details: X, quote: q } = assert;
+
 const sink = () => {};
+const makeQuietRejection = reason => {
+  const rejection = HandledPromise.reject(reason);
+  rejection.catch(sink);
```
erights: In this particular case, the `.catch` is safe. But best to avoid bad habits and reinforce good ones.

```suggestion
  E.when(rejection, undefined, sink);
```

Alternatively, is `makeQuietRejection` a sufficiently general purpose and reusable abstraction that it should move to one of: eventual-send, promise-kit, or internal ?

packages/notifier/src/publish-kit.js
```diff
@@ -102,6 +111,19 @@ export const subscribeLatest = subscriber => {
 };
 harden(subscribeLatest);
 
+export const PublisherI = M.interface('Publisher', {
+  publish: M.call(M.any()).returns(),
+  finish: M.call(M.any()).returns(),
+  fail: M.call(M.any()).returns(),
+});
```
erights, line 117: Oh, this is a challenge. My intention is that Published has each of these as optional methods, so that a local observeFoo can feature test for their presence. But I don't think I enabled `M.interface` to express such optionality. Oops.

packages/notifier/src/publish-kit.js
```diff
@@ -102,6 +111,19 @@ export const subscribeLatest = subscriber => {
...
+export const SubscriberI = M.interface('Subscriber', {
+  subscribeAfter: M.call().optional(M.bigint()).returns(M.promise()),
+});
+export const publishKitIKit = harden({
```
erights, line 121:
```suggestion
export const publishIKit = harden({
```
or perhaps just
```suggestion
export const publishKit = harden({
```

packages/notifier/src/types.js
```diff
@@ -103,6 +103,13 @@
 
 // /////////////////////////////////////////////////////////////////////////////
 
+/**
+ * @typedef {object} DurablePublishKitOptions
+ * @property {'mandatory' | 'opportunistic' | 'ignored'} [valueDurability]
```
erights: I could guess, but I don't know what these mean. Where are they explained?

packages/notifier/src/publish-kit.js
```diff
@@ -4,10 +4,19 @@
...
+
 const sink = () => {};
+const makeQuietRejection = reason => {
+  const rejection = HandledPromise.reject(reason);
+  rejection.catch(sink);
+  return rejection;
+};
 
 /**
  * Asyncronously iterates over the contents of a PublicationList as it appears.
```
erights: On line 24 below

>  * Thus, if no one else is holding on to those, it can be garbage collected.

Please clean up my singular vs plural confusion.

packages/notifier/src/publish-kit.js
erights: Nevermind. Looking at the types, I see this is the difference between `Publisher` and a `PublishObserver`. Your interface above is exactly right for `Publisher`. Makes sense.

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+/**
+ * @typedef {ReturnType<typeof initDurablePublishKitState>} DurablePublishKitState
+ */
+
+/**
+ * @param {DurablePublishKitOptions} [options]
+ */
+const initDurablePublishKitState = (options = {}) => {
+  const { valueDurability = 'mandatory' } = options;
+  assert.equal(valueDurability, 'mandatory');
```
erights, line 218: Does this mean we do not yet support any of the other settings, `'opportunistic' | 'ignored'` ? Indeed, I don't see any other mentions of `'opportunistic'` in this file. But I do see one mention of `'ignored'`.

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+    // persisted result data
+    hasValue: false,
+    value: undefined,
+  };
+};
+
+/** @type {WeakMap<object, {currentP, tailP, tailR}>} */
+const durablePublishKitEphemeralData = new WeakMap();
+
+// XXX Does state actually work as a WeakMap key?
```
erights: Likely no. Every time the same virtual object is paged in, a fresh representative is created with a fresh state record. I think we made the exposed WeakMaps "safe" for using virtual object representatives as weakmap keys. But perhaps not the state record. In any case, even if it works, it would be "wrong" (heavy air quotes) because the representative is intended to represent a long lived (virtual) object that can be accessed from outside the object. The state record is only designed to be accessed by the object (or object kit) for which it is the state record.

Could you use a virtual object instead, rather than its state record?

@FUDCo @warner , how close did I get? What's the real truth?

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
-      const rejection = HandledPromise.reject(reason);
-      // Suppress unhandled rejection error.
-      rejection.catch(() => {});
-      advanceCurrent(true, undefined, rejection);
+      advanceCurrent(true, undefined, makeQuietRejection(reason));
     },
   });
   return harden({ publisher, subscriber });
 };
 harden(makePublishKit);
```
erights, line 207: Not urgent, but I would like to understand why we still need notifier.js . AFAICT, subscribers subsume notifiers, as we intended. Where does this fail?

erights: Likewise for subscriber.js . Can we at least deprecate these and start migrating all uses to publish-kit?

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+const provideDurablePublishKitEphemeralData = state => {
+  const foundData = durablePublishKitEphemeralData.get(state);
+  if (foundData) {
+    return foundData;
+  }
+  const { status, publishCount } = state;
+  let newData;
+  if (status === 'failed') {
+    newData = {
+      currentP: makeQuietRejection(state.value),
```
erights, line 271: Is it noted somewhere that `state.value` represents a rejection reason if `state.status === 'failed'`? I mention this only because it threw me at first, because the name suggested it would only represent fulfillments, i.e., iterator result values.

erights: What do you think of typing `status` as `'failed'`, `'finished'`, or `'live'`?

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+      tailP: makeQuietRejection(
+        new Error('Cannot read past end of iteration.'),
+      ),
+      tailR: undefined,
+    };
+  } else if (status === 'finished') {
+    // XXX Should we be harden()ing rejection-reason errors?
```
erights: yes

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+          )
+        : tailP,
+      tailP,
+      tailR,
+    };
+  } else {
+    assert.fail(X`Invalid durable promise kit status: ${q(status)}`);
+  }
+  durablePublishKitEphemeralData.set(state, harden(newData));
+  return newData;
```
erights, line 312:
```suggestion
  return harden(newData);
```
or harden each in place. Or dispense with the `newData` variable and harden and return each in place. FWIW, that's what I'd do.

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+        canBeDurable(value) ||
+          assert.fail(X`Cannot accept non-durable value: ${value}`);
+        state.value = value;
+        state.hasValue = true;
+      }
+    } catch (err) {
+      if (done || valueDurability === 'mandatory') {
+        throw err;
+      }
+    }
```
erights: So you throw the error, catch the error, and then conditionally rethrow the error? Why?

erights (commented):
So it seems there's a bunch of semi-redundant logic between the ephemeral and the durable case. What surprised me most is that you didn't reuse the internals of the ephemeral publish-kit implementation to implement the ephemeral portion of your durable publish kits.

Otherwise, everything LGTM, but I'm holding back on approval until we discuss this question.

erights (approved):
Nevermind. I see why that can't work. LGTM!

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+
+/**
+ * @param {import('../../vat-data/src/types.js').Baggage} baggage
+ * @param {string} kindName
+ */
+export const vivifyDurablePublishKit = (baggage, kindName) => {
+  /**
+   * @returns {() => PublishKit<*>}
+   */
+  return vivifyFarClassKit(
```
erights, line 388: Still kinda disturbed to have two similar but different implementations. I suspect the answer is no, but it is worth asking:

Would it help to use `defineHeapFarClassKit` for the ephemeral case, so that the ephemeral case and the durable case can share more code?

packages/notifier/src/publish-kit.js
```diff
@@ -180,12 +200,221 @@ export const makePublishKit = () => {
...
+    provideDurablePublishKitEphemeralData(state);
+
+  state.publishCount += 1n;
+  const publishCount = state.publishCount;
+  let tailP;
+  let tailR;
+
+  if (done) {
+    state.status = rejection ? 'failed' : 'finished';
+    tailP = makeQuietRejection(new Error('Cannot read past end of iteration.'));
```
Chris-Hibbert, line 344: Isn't `advanceDurablePublishKit` trying to add new state? Shouldn't this be
```suggestion
    tailP = makeQuietRejection(new Error('Cannot append past end of iteration.'));
```

packages/notifier/src/publish-kit.js
gibson042: I think this name needs to have a form like "publishKit\<NonemptySuffix>", and chose "IKit" as the suffix to match prevailing patterns. But the resulting name _is_ awkward... what about "publishKitShape"?

packages/notifier/src/publish-kit.js
gibson042: When I try to use `this.facets` rather than `this.state` I get "Invalid value used as weak map key" errors, so I'm not sure how to key the ephemeral data. Adding a unique dummy object to the state of each durable publish kit would presumably suffer the same problem as using state directly... I could add an instance counter to baggage in `vivifyDurablePublishKit` and use that as the key to a _strong_ map associated with the baggage, but that seems wrong in its own way. I welcome some guidance here!

packages/notifier/src/publish-kit.js
gibson042: > What do you think of typing `status` as `'failed'`, `'finished'`, or `'live'`?

I'm in favor of that... do you have a preferred location to put the typing? Would something like this work?
```patch
-    status: 'live', // | 'finished' | 'failed'
+    status: /** @type {'live'|'finished'|'failed'} */ ('live'),
```

packages/notifier/src/publish-kit.js
gibson042: No, this is the rejection that will be seen by consumers attempting to read a new value after receiving a terminal one.

packages/notifier/src/publish-kit.js
gibson042: I was trying to avoid drift in the validation code. Fixed by refactoring.

packages/notifier/src/publish-kit.js
gibson042: It might be possible, but I don't want to make the attempt until the new implementation is stable and confirmed to survive upgrade.

packages/notifier/src/publish-kit.js
gibson042: There was an issue with implementing notifiers on top of publish kit: https://github.com/Agoric/agoric-sdk/pull/5435#issuecomment-1158874885

But FWIW, I do consider both notifier and subscriber to be deprecated in favor of publish-kit.

packages/notifier/src/publish-kit.js
gibson042: Updated. We should keep the possibility in our back pocket, but I don't think `makeQuietRejection` is needed outside of this file for now.

gibson042:
I think I'm getting closer to the bottom of upgrade issues, which manifest as an error along the following lines:

<details open><summary>In my vat-pubsub's buildRootObject <code>provide(baggage, 'publishKitSingleton', ...)</code></summary>

→ _packages/store/src/stores/store-utils.js_:**provideLazy** `mapStore.get('publishKitSingleton')`
→ _packages/swingset-vat/src/liveslots/collectionManager.js_:**summonCollectionInternal**:**get** ```unserialize(JSON.parse(String.raw`{"body":"{\"publisher\":{\"@qclass\":\"slot\",\"iface\":\"Alleged: DurablePublishKit publisher\",\"index\":0},\"subscriber\":{\"@qclass\":\"slot\",\"iface\":\"Alleged: DurablePublishKit subscriber\",\"index\":1}}","slots":["o+10/1:0","o+10/1:1"]}`))```
→ _marshal.js_:**unserialize** `reviveFromCapData({publisher:{"@qclass":"slot",iface:"Alleged: DurablePublishKit publisher",index:0}, subscriber:{…}})`
→ _encodeToCapData.js_:**decodeFromCapData** `decodeRemotableFromCapData({"@qclass":"slot",iface:"Alleged: DurablePublishKit publisher",index:0})`
→ _packages/swingset-vat/src/liveslots/liveslots.js_:**convertSlotToVal** `val=vrm.reanimate("o+10/1")`
→ _packages/swingset-vat/src/liveslots/virtualReferences.js_:**reanimate** `return reanimator("o+10/1")`
→ _packages/swingset-vat/src/liveslots/virtualObjectManager.js_:**reanimate** `makeRepresentative({baseRef:"o+10/1",rawState:{hasValue,publishCount,…},repCount:0},false)`
...

</details>
<details open><summary>in processUpgradeVat</summary>

→ _packages/swingset-vat/src/liveslots/liveslots.js_:**convertSlotToVal** `vrm.reanimate("o+10/1")`
→ _packages/swingset-vat/src/liveslots/virtualObjectManager.js_:**reanimate** `makeRepresentative({baseRef:"o+10/1",next:undefined,prev:undefined,rawState:null,repCount:0},false)`
→ _packages/swingset-vat/src/liveslots/virtualObjectManager.js_:**makeRepresentative**:**ensureState** `cache.lookup("o+10/1",true)`
→ _packages/swingset-vat/src/liveslots/virtualObjectManager.js_:**cache**:**lookup** `innerObj.rawState=fetch("o+10/1")`
→ _packages/swingset-vat/src/kernel/kernelSyscall.js_:**vatstoreGet** `kvStore.get("v6.vs.vom.o+10/1")` // => undefined
_virtualObjectManager.js_:**makeRepresentative** `Object.getOwnPropertyNames(innerSelf.rawState)` // => Error because `rawState` is undefined

</details>


Zooming in on vat data, the pre-upgrade `kvStore` includes
* 'v6.vs.vom.dkind.10': '{"kindID":"10","tag":"DurablePublishKit","nextInstanceID":2,"facets":["publisher","subscriber"]}'
* 'v6.vs.vom.rc.o+10/1': '2'
* 'v6.vs.vc.1.|entryCount': '2'
* 'v6.vs.vc.1.sDurablePublishKit_kindHandle': '{"body":"{\\"@qclass\\":\\"slot\\",\\"iface\\":\\"Alleged: kind\\",\\"index\\":0}","slots":["o+1/10"]}'
* 'v6.vs.vc.1.spublishKitSingleton': '{"body":"{\\"publisher\\":{\\"@qclass\\":\\"slot\\",\\"iface\\":\\"Alleged: DurablePublishKit publisher\\",\\"index\\":0},\\"subscriber\\":{\\"@qclass\\":\\"slot\\",\\"iface\\":\\"Alleged: DurablePublishKit subscriber\\",\\"index\\":1}}","slots":["o+10/1:0","o+10/1:1"]}'

And during upgrade I see `vatStoreGet` invocations like
* (vatID 'v6', key 'vc.1.sDurablePublishKit_kindHandle') → actualKey 'v6.vs.vc.1.sDurablePublishKit_kindHandle' → value '{"body":"{\\"@qclass\\":\\"slot\\",\\"iface\\":\\"Alleged: kind\\",\\"index\\":0}","slots":["o+1/10"]}'
* (vatID 'v6', key 'vom.dkind.10') → actualKey 'v6.vs.vom.dkind.10' → value '{"kindID":"10","tag":"DurablePublishKit","nextInstanceID":2,"facets":["publisher","subscriber"]}'
* (vatID 'v6', key 'vc.1.spublishKitSingleton') → actualKey 'v6.vs.vc.1.spublishKitSingleton' → value '{"body":"{\\"publisher\\":{\\"@qclass\\":\\"slot\\",\\"iface\\":\\"Alleged: DurablePublishKit publisher\\",\\"index\\":0},\\"subscriber\\":{\\"@qclass\\":\\"slot\\",\\"iface\\":\\"Alleged: DurablePublishKit subscriber\\",\\"index\\":1}}","slots":["o+10/1:0","o+10/1:1"]}'
* (vatID 'v6', key 'vom.o+10/1') → actualKey 'v6.vs.vom.o+10/1' → value undefined (_which causes the error_)


So my question now is about why post-upgrade requests look for data at 'v6.vs.vom.o+10/1' while pre-upgrade requests stored data at 'v6.vs.vom.**rc**.o+10/1'. Any thoughts, @warner or others?

warner:
We tracked this down to a serious bug in the `stopVat` code: it was not flushing the virtual object manager LRU cache. The line to do a `bringOutYourDead` was accidentally commented out when we disabled a lot of the cleanup code, and as a result, the virtual object writes done by @gibson042 's first vat incarnation never made it into the DB, so the second incarnation could not find them there. I'll write up a bug, but I think @gibson042 will be making the fix as a side effect of this PR.

erights:
What's the status of this? From skimming the threads above, it seems it was blocked on a mysterious bug that has since been diagnosed. Note that https://github.com/Agoric/agoric-sdk/pull/5997 is stacked on the PR, so it would be good to get this moving again if possible.

packages/notifier/test/test-publish-kit.js
```diff
@@ -1,274 +1,476 @@
...
+import { E } from '@endo/eventual-send';
+import { makeMarshal } from '@endo/marshal';
+import {
+  buildKernelBundles,
+  initializeSwingset,
+  makeSwingsetController,
+} from '@agoric/swingset-vat';
+import { provideHostStorage } from '../../SwingSet/src/controller/hostStorage.js';
+// import { makeScalarMapStore as makeBaggage } from '@agoric/store';
+import { makeScalarBigMapStore } from '../../vat-data/src/vat-data-bindings.js';
```
turadg: this references another package so it should use package loading, not file system paths which may change.
```suggestion
import { provideHostStorage } from '@agoric/swingset-vat/src/controller/hostStorage.js';
import { makeScalarBigMapStore } from '@agoric/vat-data/src/vat-data-bindings.js';
```

packages/notifier/test/test-publish-kit.js
```diff
@@ -1,274 +1,476 @@
...
+  t.throws(() => publisher.finish(nonPassable));
+  t.throws(() => publisher.fail(nonPassable));
+  await assertTransmission(t, publishKit, Symbol.for('value'));
+});
+
+test('durable publish kit upgrade trauma (full-vat integration)', async t => {
+  const config = {
+    includeDevDependencies: true, // for vat-data
+    defaultManagerType:
+      /** @type {import('../../SwingSet/src/types-external.js').ManagerType} */ (
```
turadg:
```suggestion
      /** @type {import('@agoric/swingset-vat/src/types-external.js').ManagerType} */ (
```

packages/notifier/test/test-publish-kit.js
```diff
@@ -1,274 +1,476 @@
...
+      bootstrap: { sourceSpec: bfile('vat-integration/bootstrap-upgrade.js') },
+    },
+    bundles: {
+      pubsub: { sourceSpec: bfile('vat-integration/vat-pubsub.js') },
+    },
+  };
+  const hostStorage = provideHostStorage();
+  const { kernel: kernelBundle, ...kernelBundles } = await buildKernelBundles();
+  const initOpts =
+    /** @type {{kernelBundles: Record<string, import('../../SwingSet/src/types-external.js').Bundle>}} */ ({
```
turadg:
```suggestion
    /** @type {{kernelBundles: Record<string, import('@agoric/swingset-vat/src/types-external.js').Bundle>}} */ ({
```

packages/notifier/test/test-publish-kit.js
```diff
@@ -293,9 +495,16 @@ test('subscribeLatest', async t => {
...
   );
-});
+};
+
+for (const [type, maker] of Object.entries(makers)) {
+  test(`subscribeLatest(${type} subscriber)`, verifySubscribeLatest, maker);
+}
 
-test('subscribeEach', async t => {
+// TODO: Replace with test.macro once that works with prepare-test-env-ava.
```
turadg: try using `prepare-test-env` instead of `prepare-test-env-ava` and `import test from 'ava';`

erights: Does prepare-test-env change `'ava'` into ses-ava?

turadg: No, it doesn't do anything to Ava.
https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/tools/prepare-test-env.js

packages/notifier/test/test-publish-kit.js
gibson042: Done, except for `bfile('../../SwingSet/test/bootstrap-relay.js')` (which I don't know how to fix because `bfile` relies on `new URL`).

packages/notifier/test/test-publish-kit.js
gibson042: Success! Thanks!

gibson042:
@erights @FUDCo @warner This is ready for re-review, though it's worth noting that not much has changed outside of testing code. There's one substantial [open issue regarding map keys](https://github.com/Agoric/agoric-sdk/pull/6502#discussion_r1016136997) and some questions regarding the naming of the function that processes (baggage, kindName) into a function that makes { publisher, subscriber } publish kit instances (currently called `vivifyDurablePublishKit` per https://github.com/Agoric/agoric-sdk/pull/6502#discussion_r1007115229 ) and the naming of the object that describes the interface of such instances (currently called `publishKitIKit` but questioned in https://github.com/Agoric/agoric-sdk/pull/6502#discussion_r1014543291 ), and also a question about how to type persisted status at https://github.com/Agoric/agoric-sdk/pull/6502#discussion_r1016144198 .

packages/notifier/src/publish-kit.js
turadg: That would work. Though you'd have to do the same for `value`:
Screenshot 2022-12-06 at 10 32 42 AM
That wasn't obviously necessary. So I'd suggest having an explicit type instead, making apparent what values are typed or not. 

It would also have the advantage of being a better place for the documentation on what the values mean. E.g. this comment could be visible when hovering over an instance of the state type: https://github.com/Agoric/agoric-sdk/blob/96979ceb43e7c04e6f93707eb198405859a797d4/packages/notifier/src/publish-kit.js#L229-L232

packages/notifier/src/publish-kit.js
turadg: Agree it needs a suffix. `IKit` is the prevailing suffix. Another option, matching the param name, would be `publishKitGuardKit`. 
https://github.com/Agoric/agoric-sdk/blob/96979ceb43e7c04e6f93707eb198405859a797d4/packages/vat-data/src/far-class-utils.js#L167

Another option is to simply inline it. Why is it exported? What else might have to verify an argument matches this shape?

gibson042: It is exported per https://github.com/Agoric/agoric-sdk/pull/6502#discussion_r1008580171 ; perhaps @Chris-Hibbert can explain?

packages/notifier/src/publish-kit.js
gibson042: Done, please review: https://github.com/Agoric/agoric-sdk/pull/6502/commits/a0d56611a97894d015b2fc4a9ffa66ea514c0576

packages/notifier/src/publish-kit.js
Chris-Hibbert: I use PublisherI and SubscriberI in Zoe.  They're delivered as a pair, so I'd like to use the canonical declaration.

turadg: Got it. Since the current pattern is typename + `IKit` for the interface guard of that type, I propose it stay this way. Coming up with a different scheme is outside the scope of this PR.

packages/notifier/src/publish-kit.js
warner: @erights is correct. While using `state` as a WeakMap key cannot cause non-determinism or be used as a GC sensor (due to all the work we put into the [VirtualObjectAwareWeakMap](https://github.com/Agoric/agoric-sdk/blob/c7d422343c9fdfd173b6e756ad2a02577d7c4574/packages/SwingSet/src/liveslots/virtualObjectManager.js#L284)), it's not meant to be used that way. Using the `Publisher` facet would be more "with the grain".

I'm not sure why `this.facets` failed; unless the new class/`this`-based API means `this.facets` should be spelled `this.context.facets` or something, I'd expect it to behave basically the same as `this.state`. But neither `state` nor `facets` are meant to be used for their identity: only the facets themselves are made for that.

I think the immediate consequence of using `state` is that objects will be kept around longer than strictly necessary (i.e. increased RAM usage). However I'm not sure how much of that we can avoid. So perhaps we should land this as-is, and then continue to pursue further optimization later.

(BTW my [original sketch](https://github.com/Agoric/agoric-sdk/issues/4567#issuecomment-1088182184) for this did indeed use an integer counter as the key to a strong Map, as @gibson042 describes, and re-reading the discussion there reminds me that I didn't think we could build a high-cardinality yet low-RAM version back in April either, so yeah, trying to make this RAM efficient might not be the highest priority)

## Further Details

The `state` object is just another anonymous JS object: unlike the facets (i.e. Representatives) it lacks a `vref` identity. The same is true for the other objects we create in the course of making the Representative, like `facets` (the `{ publisher, subscriber }` container which holds them), and the `context` record which holds `{ facets, state }`. However we link these four kinds of objects (`state`, `facets`, `context`, and the facets themselves) to make sure that userspace cannot sense GC by e.g. holding onto `state`, waiting for the facets to get dropped and collected (or not), pulling `publisher` back out of a virtual collection, then comparing the new `state` object against the held copy: if they're the same, GC did not happen, if they differ, GC did happen. `context` contains `facets`, and `facets` contains the facets, and each facet closes over a reference to `state`. Then, by [maintaining an edge](https://github.com/Agoric/agoric-sdk/blob/c7d422343c9fdfd173b6e756ad2a02577d7c4574/packages/SwingSet/src/liveslots/virtualObjectManager.js#L752) from `state` to the [cohort of facets](https://github.com/Agoric/agoric-sdk/blob/c7d422343c9fdfd173b6e756ad2a02577d7c4574/packages/SwingSet/src/liveslots/virtualObjectManager.js#L742), we form a cycle, so `state` and `facets` and the facets all keep each other alive.

In addition, the `state` object is [marked as "unweakable"](https://github.com/Agoric/agoric-sdk/blob/c7d422343c9fdfd173b6e756ad2a02577d7c4574/packages/SwingSet/src/liveslots/virtualObjectManager.js#L718), so when it is used as a WeakMap/WeakSet key, we add an extra reference edge from the WeakMap to the `state`, to ensure that `state` sticks around at least as long as the WeakMap. We do the same for `context`. This prevents userspace from building a recognizer for one of these objects with `weakset.add()`, then probing it with a later copy to see if they differ, revealing that GC happened. By forcing `state` to be held strongly, we inhibit the GC, so the recognizer will always give the same answer.

The net result is that these internal objects cannot be used to sense GC, but using them as a WeakMap key will actually cause them to be held strongly, until either the WeakMap itself goes away, or someone explicitly removes it.

Getting back to publish-kit.js, with the current PR, the `publisher` and `subscriber` facets will be kept in RAM whenever `state` is being used as a key in the `durablePublishKitEphemeralData` WeakMap, and that happens the first time any method is called on either facet (and is never removed). So in practice, every cohort will be held in RAM all the time, along with all of its state, making this consume just as much RAM as a non-virtual version. They will be durable, but high cardinality will be a RAM problem.

Using `state` or `facets` or `context` as the WeakMap key doesn't let you take advantage of virtualization, because none of them have a vref identity. Things could be better if you use one of the *facets* as the WeakMap key, like `publisher`, or a third (empty) facet that you create just to serve as a key. In this case, the VirtualObjectAwareWeakMap [uses the facet's vref](https://github.com/Agoric/agoric-sdk/blob/c7d422343c9fdfd173b6e756ad2a02577d7c4574/packages/SwingSet/src/liveslots/virtualObjectManager.js#L320) (a string) as the key of a strong Map, pointing to the value. It allows the facet (the Representative) to be dropped (from RAM), which allows the `state` data to be dropped (from RAM), along with the rest of the cohort. They will be paged back in if/when someone invokes a method on either facet, but in the meantime, we only hold a vref string in RAM, not the whole `{ publisher, subscriber }` cohort. `WeakMap` is not virtual, it's just *aware* of virtual objects.

Doing that would save you the RAM of the facets, however the design is still keeping the full resolution value of all publishers in RAM. To reduce usage further, we want to move that to DB. It looks like `durablePublishKitEphemeralData` is used to hold two promises (`currentP` and `tailP`), and one bare resolver function (`tailR`). `currentP` may be resolved to some arbitrary data (the most recently published value), or it might be unresolved. I see that `value` is stored on `state.value` (thus in the durable object, thus in the DB, good), but it looks like we're also holding a copy of `value` in a resolved Promise, which means holding the resolution data in RAM (bad). So this approach will keep all the current resolution values in RAM, even though we don't need them until a new subscriber comes along.

A better approach would retain fewer Promises. I think the two dominant cases are a new subscriber to a non-empty Publisher (i.e. they should get the current value), or a subscriber who is up-to-date and will be waiting for a value. In the first case, I think we could return a new Promise (freshly resolved to the current value) to each new subscriber, rather than keeping around a pre-resolved Promise for them. (Note that promise ID `vpid`s are retired at the vat/kernel boundary upon resolution, so you don't save any syscalls by sending the same resolved promise over the wire multiple times: each one requires a new `vpid` allocation and a new `syscall.resolve()`). The current value would remain in the DB, in `state.value`, until needed.

For the second case, I think we *do* need to hold onto a single Promise in RAM, but it should be unresolved. We need the WeakMap or a BigScalarMapStore to map from something with a vref (the `publisher` facet) to a pair of the unresolved Promise and a resolver function. The RAM usage here will be O(N) in the number of publishers with at least one active subscriber, but unaffected by the total amount of resolution data, or the number of unloved publishers, all of which will only consume DB space.

packages/notifier/src/publish-kit.js
```diff
@@ -179,12 +199,219 @@ export const makePublishKit = () => {
...
-      rejection.catch(() => {});
-      advanceCurrent(true, undefined, rejection);
+      advanceCurrent(true, undefined, makeQuietRejection(reason));
     },
   });
   return harden({ publisher, subscriber });
 };
 harden(makePublishKit);
+
+// TODO: Move durable publish kit to a new file?
```
warner, line 209: yes please :)

packages/notifier/src/publish-kit.js
```diff
@@ -179,12 +199,219 @@ export const makePublishKit = () => {
...
+  }
+  if (done || valueDurability === 'mandatory') {
+    canBeDurable(value) || Fail`Cannot accept non-durable value: ${value}`;
+    canBeDurable(rejection) ||
+      Fail`Cannot accept non-durable rejection: ${rejection}`;
+  }
+  const { tailP: currentP, tailR: resolveCurrent } =
+    provideDurablePublishKitEphemeralData(state);
+
+  state.publishCount += 1n;
```
warner: Note that `state` has getters/setters, and each read and each write requires a deserialization or serialization. So this might be marginally faster if you retain the incremented value, rather than re-read it off of `state`:

```js
const oldCount = state.publishCount;
const publishCount = oldCount + 1n;
state.publishCount = publishCount;
// now use 'publishCount'
```

packages/notifier/src/types-ambient.js
```diff
@@ -101,6 +101,45 @@
...
 
+/**
+ * @typedef {'mandatory' | 'opportunistic' | 'ignored'} DurablePublishKitValueDurability
+ *
+ * Durability configures constraints on non-terminal values that can be
+ * published to a durable publish kit (terminal values sent to finish or fail
+ * must always be durable).
+ * - 'mandatory' means that each value must be durable, so it can be restored
+ *   on upgrade.
+ * - 'opportunistic' means that a durable value is persisted for post-upgrade
```
warner, line 112: do we really have a use case for this? sounds confusing, to have the value get mysteriously replaced by `undefined` after an ugprade

packages/notifier/test/vat-integration/vat-pubsub.js
warner: Does this work with the LRU fix in place? Or is the TODO now stale?

warner (approved):
Ok, I *think* I understand what this is doing. I'm going to assume that the main notifier/publisher API does what you want it to do (other people have reviewed that), and I'll restrict myself to the use of virtual/durable objects.

Assuming the `vat-pubsub.js` TODO comment is stale (and that test now works, with the LRU cache flush bug fixed), I can provisionally approve this for landing, if we can commit to doing more RAM-reducing perf work on it before it gets used in high-cardinality anger.

If you're in the mood to make some changes before landing, I'd suggest starting with changing the WeakMap key from `state` to `facets.publisher`. I believe that should exchange some RAM for some DB, by allowing the publisher/subscriber facets to be dropped from RAM until someone uses them. We'll pay some CPU and DB access to rebuild them when that happens, but we'll save the RAM costs of keeping them around forever, which is what will happen if we keep using `state` as the key.

I think the second change I listed (don't hold resolved Promises in RAM, instead keep the current value only in the DB) can wait until after this lands. I expect it won't change the API, so it shouldn't cause compatibility problems to improve it later.

packages/notifier/src/types-ambient.js
gibson042: @dtribble requested the ability to publish non-durable values to a durable publish kit, and this is the sketch of what it would look like (but note that the implementation does not actually support it and will throw upon an attempt to use it).

packages/notifier/src/publish-kit.js
gibson042: I definitely want to discuss this further face-to-face, but in the meantime I've just taken the primary suggestion to key the ephemeral-data WeakMap with a facet. Assuming CI looks good, this should be clear to merge.

PR #6504, opened 2022-10-27
feat(makeChainInfo)!: rpc param as authoritative

turadg:
closes: #6503

## Description

This condition seems to have been for backwards compatibility: https://github.com/Agoric/agoric-sdk/blob/450e937d62388adbe81e4a3c5f6f21557d2df66e/packages/wallet/ui/src/util/SuggestChain.js#L23-L30

This might break some devex but it's more urgent that we de-risk production.

### Security Considerations

n/a

### Documentation Considerations

.

### Testing Considerations

Will follow up with a test if needed.

PR #6505, opened 2022-10-27
feat(makeChainInfo): optional apiAddrs

turadg:
## Description

Follow up to https://github.com/Agoric/agoric-sdk/pull/6504

- adds static type
- adds tests
- makes it fall back to older behavior when `apiAddrs` absent in network-config

### Security Considerations

.

.
### Testing Considerations

unit

### DIFF
--- a/packages/casting/src/netconfig.js
+++ b/packages/casting/src/netconfig.js
@@ -1,0 +1,39 @@
+// @ts-check
+import { fit, M } from '@agoric/store';
+
+// NB: keep type and shape in sync manually until https://github.com/Agoric/agoric-sdk/issues/6160
+/**
+ * @typedef {object} NetworkConfig
+ * `peers` and `seeds` are only used for configuring full nodes participating in
+ * the chain’s p2p network, not for rpc clients.  The `gci` is needed for rpc
+ * “light clients” to prove the ancestry of data fetched from the rpc node.
+ *
+ * @property {string[]} rpcAddrs endpoints serving https://docs.tendermint.com/v0.34/rpc/
+ * @property {string} [chainName] e.g. agoric-3
+ * @property {string[]} [apiAddrs] overrides for REST endpoints of RPC servers https://docs.cosmos.network/main/run-node/interact-node#using-the-rest-endpoints
+ * @property {string} [gci] - "global chain identifier", a hash of the genesis block
+ * @property {string[]} [peers] - a list of nodes used to start the p2p gossip (stored in a per-node “address book”, which is a file stored in that node’s data directory)
+ * @property {string[]} [seeds] - nodes which tell you about other peers but don't gossip actual data
+ */
+export const NetworkConfigShape = M.split(
+  {
+    rpcAddrs: M.arrayOf(M.string()),
+  },
+  M.partial({
+    apiAddrs: M.arrayOf(M.string()),
+    chainName: M.string(),
+    gci: M.string(),
+    peers: M.arrayOf(M.string()),
+    seeds: M.arrayOf(M.string()),
+  }),
+);
+harden(NetworkConfigShape);
+
+/**
+ * @param {unknown} specimen
+ * @returns {asserts specimen is NetworkConfig}
+ */
+
+export const assertNetworkConfig = specimen =>
+  fit(specimen, NetworkConfigShape);
+harden(assertNetworkConfig);
--- a/packages/casting/src/netconfig.js
+++ b/packages/casting/src/netconfig.js
@@ -9,7 +9,7 @@ import { fit, M } from '@agoric/store';
  * “light clients” to prove the ancestry of data fetched from the rpc node.
  *
  * @property {string[]} rpcAddrs endpoints serving https://docs.tendermint.com/v0.34/rpc/
- * @property {string} [chainName] e.g. agoric-3
+ * @property {string} chainName e.g. agoric-3
  * @property {string[]} [apiAddrs] overrides for REST endpoints of RPC servers https://docs.cosmos.network/main/run-node/interact-node#using-the-rest-endpoints
  * @property {string} [gci] - "global chain identifier", a hash of the genesis block
  * @property {string[]} [peers] - a list of nodes used to start the p2p gossip (stored in a per-node “address book”, which is a file stored in that node’s data directory)
@@ -17,11 +17,11 @@ import { fit, M } from '@agoric/store';
  */
 export const NetworkConfigShape = M.split(
   {
+    chainName: M.string(),
     rpcAddrs: M.arrayOf(M.string()),
   },
   M.partial({
     apiAddrs: M.arrayOf(M.string()),
-    chainName: M.string(),
     gci: M.string(),
     peers: M.arrayOf(M.string()),
     seeds: M.arrayOf(M.string()),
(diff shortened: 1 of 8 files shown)


packages/wallet/ui/src/util/SuggestChain.js
```diff
@@ -5,36 +5,33 @@ export const AGORIC_COIN_TYPE = 564;
...
-  caption,
-) => {
-  const coinType = Number(
-    new URL(networkConfig).searchParams.get('coinType') || AGORIC_COIN_TYPE,
-  );
-  const hostname = new URL(networkConfig).hostname;
-  const network = hostname.split('.')[0];
+export const makeChainInfo = (io, networkConfig, caption) => {
+  const { chainName, rpcAddrs, apiAddrs } = networkConfig;
+  const index = Math.floor(io.random() * rpcAddrs.length);
```
dckc: passing in the random number generator for 1 call seems a little odd -- why not just pass in the results of the call?

packages/wallet/ui/src/util/tests/SuggestChain.test.js
```diff
@@ -0,0 +1,186 @@
...
+ */
+const ollinetExample = {
+  chainName: 'agoricollinet-44',
+  gci: 'https://ollinet.rpc.agoric.net:443/genesis',
+  peers: ['fb86a0993c694c981a28fa1ebd1fd692f345348b@34.67.167.8:26656'],
+  rpcAddrs: ['https://ollinet.rpc.agoric.net:443'],
+  seeds: ['0f04c4610b7511a64b8644944b907416db568590@34.172.146.145:26656'],
+};
+
+const io = { random: Math.random };
```
dckc: We have only 1 entry in our `rpcAddrs` array, so it doesn't matter yet, but this should be a mock random number generator with predictable outputs for testing.

packages/wallet/ui/src/util/tests/SuggestChain.test.js
```diff
@@ -0,0 +1,186 @@
...
+
+test('mainnetExample', () => {
```
dckc: I suppose this covers it, but what I had in mind was a test assertion of the form "rpcAddr parameter is not clobbered".

packages/wallet/ui/src/util/SuggestChain.js
```diff
@@ -5,36 +5,33 @@ export const AGORIC_COIN_TYPE = 564;
...
   return {
     rpc,
-    rest: apiAddr,
-    chainId,
-    chainName: caption || `Agoric ${network}`,
+    rest,
+    chainId: chainName,
+    chainName: caption,
     stakeCurrency,
-    walletUrlForStaking: `https://${network}.staking.agoric.app`,
```
dckc, line 35: No more staking link? That looks like a regression / breaking change. I have used it from time to time. I suppose I won't miss it that badly.

packages/wallet/ui/src/util/SuggestChain.js
```diff
@@ -44,41 +41,38 @@ const makeChainInfo = (
...
+  const res = await fetch(url);
   if (!res.ok) {
     throw Error(`Cannot fetch network: ${res.status}`);
   }
 
-  const { chainName: chainId, apiAddrs, rpcAddrs } = await res.json();
-  assert.equal(apiAddrs.length, rpcAddrs.length);
-  const index = Math.floor(random() * rpcAddrs.length);
+  /** @type {import('@agoric/internal').NetworkConfig} */
+  const networkConfig = await res.json();
```
dckc: I make this sort of type annotation all the time when I'm prototyping, but it's unsound. I suppose it's just documenting an assumption we were already making, so it's not a regression. But we should add a shape guard or the like. Or use type `unknown` so tsc forces us to validate each assumption we make about this input.

Please make some XXX annotation to say "this is not best practice; don't copy this idiom blindly".

packages/wallet/ui/src/util/SuggestChain.js
dckc: Hm... we did ship, so maybe it's worth more discipline about this sort of thing. For the benefit of release notes, how about a separate BREAKING CHANGE commit to take out `walletUrlForStaking` before this `feat`?

dckc (commented):
I'd like to discuss a couple things before approving.

packages/wallet/ui/src/util/SuggestChain.js
turadg: agreed. the dan voice in my head said, "pass the ocap" but that Dan is an crinkled facsimile of the real one

packages/wallet/ui/src/util/SuggestChain.js
turadg: I think this is the order of bad to good:
1. implicit `any` (what we have all over the place)
2. constrain `any` (what is here)
3. unknown (so you have to validate on use, but a pain)
4. type guard (which is why I put the TODO for that)

I'm having trouble thinking up what the unsound type exploit is here. If someone controls network config, there are much worse things they can do

packages/wallet/ui/src/util/SuggestChain.js
turadg: I found it nowhere in the Agoric org (https://github.com/Agoric/agoric-sdk/search?q=walletUrlForStaking) and I didn't think it was really part of the API of this util.

It's technically breaking the API, but I don't think it breaks anything. If we're going to get disciplined about breaking changes, then I think we also need to get more disciplined about what we export.

In any case, I'll try separating this out as a BREAKING CHANGE to see how it feels ;-)

dckc: It's not used in our code; it's used in by Keplr:

image


Again: this is a network protocol / API.

packages/wallet/ui/src/util/SuggestChain.js
dckc: Agreed on the order. I guess I was reading too fast and missed the TODO.

> I'm having trouble thinking up what the unsound type exploit is here.

I don't know that this is exploitable. But the unsoundness adds a burden when reviewing; I have to keep track of all the places where `networkConfig` is used, knowing that it's really an external input and might not be of the declared type.

Thanks for the type guard.

dckc (approved):

packages/wallet/ui/src/util/SuggestChain.js
turadg: Ah. I'll restore it.

(Fwiw, I don't see the antecedent for "again". It definitely wasn't clear from your earlier comments that this was an actual regression. )

turadg:
@dckc I'm putting this back into draft until I can document [gci, peers, seeds](https://github.com/Agoric/agoric-sdk/pull/6505/files#diff-be19ecccf4827e08edb9e57079dd7b3dabd0c37ea9fce2874b62b12e2dc97a80R10-R12)

dckc (approved):
LGTM

packages/casting/src/netconfig.js
```diff
@@ -9,19 +9,19 @@ import { fit, M } from '@agoric/store';
  * “light clients” to prove the ancestry of data fetched from the rpc node.
  *
  * @property {string[]} rpcAddrs endpoints serving https://docs.tendermint.com/v0.34/rpc/
- * @property {string} [chainName] e.g. agoric-3
+ * @property {string} chainName e.g. agoric-3
```
dckc, line 12: oops; I wasn't reviewing very carefully, was I?

turadg: type system thing, rare for human eyes to catch ime

PR #6511, opened 2022-10-31
BREAKING CHANGE: remove  @agoric/wallet-ui

turadg:
closes: #6474

## Description

Per #6474, the Wallet app code is now at https://github.com/Agoric/wallet-app. 

This PR removes it from agoric-sdk so that there is just one canonical Wallet UI. I believe that the [instructions for using the older version compatible with ag-solo ](https://github.com/Agoric/agoric-sdk/blob/cf6836259b35e5e5bfd927379825f642c3724074/packages/wallet/README.md#L12-L18)still work since they check out community-dev and not master. (Indeed, master wallet-ui is already broken for agsolo)

### Security Considerations

Improves separation of builds and dependencies.

### Documentation Considerations

End users are not affected.
Developers who wish to contribute to the Wallet app will need to know about the new repository. This commit message directs them.
Ag-solo users will need to use the instructions above to get an older version. (We've already decided agsolo doesn't need a GUI going forward).

### Testing Considerations

CI

PR #6516, opened 2022-11-01
test(getting-started): remove /wallet

turadg:
## Description

broke in https://github.com/Agoric/agoric-sdk/actions/runs/3365023455.

### Testing Considerations

would have been caught without bypass:integration ([ref](https://github.com/Agoric/agoric-sdk/pull/6511#issuecomment-1297881110)). force integration to confirm this remedies.


mhofman (approved):
Given that https://github.com/Agoric/agoric-sdk/actions/runs/3366603406/jobs/5583262644 succeeded, approving.

PR #6520, opened 2022-11-01
1254 runtime changes

turadg:
## Description

https://github.com/Agoric/agoric-sdk/pull/6500/ [failed chain deployment test](https://github.com/Agoric/agoric-sdk/actions/runs/3372290358/jobs/5595546535). It's not clear why but to help that through I've extracted all its runtime changes to this PR.


mhofman:
Strange. I don't think that error should be related. It seems something went sideways in the loadgen runner when parsing the slog. Will investigate.

mhofman:
Looks like deployment succeeded but linting failed.

michaelfig (approved):

PR #6521, opened 2022-11-01
fix: cosmic-proto exports

turadg:
## Description

https://www.npmjs.com/package/@agoric/cosmic-proto doesn’t have any code. I thought https://github.com/Agoric/agoric-sdk/pull/6510 would solve that, but [a build of agoric-sdk with it ](https://www.npmjs.com/package/@agoric/cosmic-proto/v/0.2.2-dev-0534694.0) still doesn’t have them. 

I tried removing the `swingset/*` symlinks but it broke consumers that don't understand package.json `exports`.

### Security Considerations

--

--

### Testing Considerations

CI. I see this is covered (transitively) by `test-publish-bundle.js`.

### DIFF
--- a/packages/cosmic-proto/package.json
+++ b/packages/cosmic-proto/package.json
@@ -15,8 +15,8 @@
   },
   "type": "module",
   "exports": {
-    "./swingset/msgs.js": "./swingset/msgs.js",
-    "./swingset/msgs.ts": "./swingset/msgs.ts",
+    "./swingset/msgs.js": "./gen/agoric/swingset/msgs.js",
+    "./swingset/msgs.ts": "./gen/agoric/swingset/msgs.ts",
     "./package.json": "./package.json"
   },
   "scripts": {
@@ -38,7 +38,8 @@
   "files": [
     "build",
     "LICENSE*",
-    "swingset/*"
+    "swingset",
+    "gen"
   ],
   "publishConfig": {
     "access": "public"


packages/cosmic-proto/package.json
```diff
@@ -38,7 +38,7 @@
   "files": [
     "build",
     "LICENSE*",
-    "swingset/*"
+    "gen"
```
kriskowal, line 42: This will work if we include both `swingset` (symlink sources) and `gen` (symlink targets).

packages/cosmic-proto/package.json
```diff
@@ -15,8 +15,8 @@
   },
   "type": "module",
   "exports": {
-    "./swingset/msgs.js": "./swingset/msgs.js",
-    "./swingset/msgs.ts": "./swingset/msgs.ts",
+    "./swingset/msgs.js": "./gen/agoric/swingset/msgs.js",
+    "./swingset/msgs.ts": "./gen/agoric/swingset/msgs.ts",
```
kriskowal, line 19: This is probably okay but we still need the symlinks for systems that don’t support `exports` directives, like some bundlers. It would probably be safer in the face of maintenance to leave these alone and rely on the symlinks.

kriskowal (changes requested):

kriskowal (approved):

turadg:
When I `npm install` this package, the `swingset` path is absent, even though it's in `files`. With the `exports` change I hoped my consuming package would get them from `gen` but it's not.

I though this might be sanitization in the registry, but I just noticed `npm pack` locally also omits `swingset` in its output (even though it includes it in the local tarball):
```
npm notice === Tarball Contents ===
npm notice 11.4kB  LICENSE
npm notice 184B    README.md
npm notice 17.9kB  gen/agoric/swingset/msgs.js
npm notice 24.9kB  gen/agoric/swingset/msgs.ts
npm notice 65B     gen/gogoproto/gogo.js
npm notice 65B     gen/gogoproto/gogo.ts
npm notice 92.1kB  gen/google/protobuf/descriptor.js
npm notice 102.2kB gen/google/protobuf/descriptor.ts
npm notice 1.5kB   package.json
```

Apparently this is by design: https://github.com/npm/npm/issues/3310

I'll work on a prepack to include them.

PR #6524, opened 2022-11-02
fix dev publish

turadg:
## Description

The dev NPM packages aren’t updating since [landing #6500 yesterday](https://github.com/Agoric/agoric-sdk/pull/6500) because the tsc --build is [failing in the after-merge step](https://github.com/Agoric/agoric-sdk/actions/runs/3373320829/jobs/5597817284).

One reason is that the build output confuses the other package builds. The `tsc --clean` doesn't work so this does `git clean -f`.

Another was that the `tsc build` for smart-wallet could not resolve some ambient types. This makes those explicitly imported.

### Security Considerations

--

--
### Testing Considerations

Ran `npm pack` in each directory that now has `jsconfig.build.json`.


mhofman:
I'm somewhat concerned about casually doing `git clean -f`. If I had untracked files in my local repo, I would not expect them to go away when doing `yarn clean`

turadg:
> If I had untracked files in my local repo, I would not expect them to go away when doing yarn clean

They'd only be cleared in the package in which you ran `npm pack`. Seems uncommon to me to have stray files when you're packaging a release. 

I see neither Approve nor Request Changes. Do you propose an alternative to solve the failing publish step?

mhofman:
I don't know what the solution should be but running `yarn clean` should have the semantics of cleaning only files that are build artifacts, not other files.

I would prefer to see a more targeted approach, maybe something that limits it to untracked `.d.ts` and `.d.ts.map` files?

turadg:
> I would prefer to see a more targeted approach, maybe something that limits it to untracked .d.ts and .d.ts.map files?

👍  This now limits the clean to `*.d.ts*`. Tested with `touch tmp; yarn pack` and the `tmp` file was still present.

mhofman (dismissed):
Edit: GitHub UI got stuck

mhofman (approved):
LGTM

PR #6528, opened 2022-11-02
fix: triage await safety for cli,inter,web

erights:
Extracted from #6219 as agreed at https://github.com/Agoric/agoric-sdk/pull/6219#issuecomment-1287482865

@turadg I will ask you to review when it is ready. Not ready yet.

@dtribble @turadg you're only listed as reviewers now because of code owners. Ignore until I say it's ready for review.


erights:
With the simpler `await` safety rule, this PR is obsolete. Closing.

PR #6529, opened 2022-11-02
fix: triage await safety for cache,deploy,solo

erights:
Extracted from #6219 as agreed at https://github.com/Agoric/agoric-sdk/pull/6219#issuecomment-1287482865

@michaelfig I will ask you to review when it is ready. Not ready yet.


erights:
With the simpler `await` safety rule, this PR is obsolete. Closing.

PR #6532, opened 2022-11-02
docs: include mintedPoolBalance in PSM metrics

turadg:
## Description

The code published `mintedPoolBalance` but the type lacked it.

### Security Considerations

--

--
### Testing Considerations

--


packages/inter-protocol/src/psm/psm.js
```diff
@@ -39,6 +39,8 @@ const { details: X } = assert;
  *
  * @property {Amount<'nat'>} anchorPoolBalance  amount of Anchor token
  * available to be swapped
+ * @property {Amount<'nat'>} mintedPoolBalance  amount of Minted token
+ * available to be swapped
```
Chris-Hibbert:
```suggestion
 * @property {Amount<'nat'>} mintedPoolBalance  amount of Minted token
 * outstanding, that is amount minted minus amount burned.
```

turadg: why change "available to be swapped" that `anchorPoolBalance has?

WDYT of,
```suggestion
 * @property {Amount<'nat'>} mintedPoolBalance  amount of Minted token
 * available to be swapped (equivalent to amount minted minus amount burned)
```

Chris-Hibbert: Unless I'm mis-reading it, "available to be swapped" applies to `anchorPoolBalance`, since that is necessarily the amount of the anchor currency that we actually have on hand and can sell to a trader.
In the case of the minted currency, we don't have any on hand. When we decide to sell some to a trader, we'll mint it afresh.  So instead of the amount on hand, ready to sell, this metric tracks the balance of what we have previously minted and sold and not yet received back, so it's the outstanding amount.

turadg: My misread. Thanks for elaborating.

Chris-Hibbert (approved):

Issue #6525, opened 2022-11-02
Swingset/Execution Fee transparency on Wallet UI for PSM trades


otoole-brendan:
## What is the Problem Being Solved?
We currently do not handle swingset fees on the wallet UI. The current behavior is a user will only get the batched 0.2 IST fee deducted from their balance every 10th trade with no explanation as to what is happening.

## Description of the Design
Batching execution fees is no longer required to reduce load [???]. To remove batching is a contract change [???] (too much work for now but will be handled in another ticket). To reduce the batch size is a governance vote (which we want to do but need to bundle it up with more substantial changes). The ideal happy path is to be able to preview the expected execution fee and charge users said amount once offer is accepted. That is the medium-longer term goal of this ticket.

Most logical place to show execution fee is in the offer proposal pop-up once it is sent to the wallet.

Shorter term though, we need to at the very least communicate what is happening to users so they are not surprised by the batched deduction.

I propose we add some text explaining as much to the wallet. I'm thinking either to the Offer Proposal pop-up or near the purse amounts along the lines of: 

'An execution fee of 0.02 IST applies to this trade but will only be deducted after 10 successful trades' 

## Security Considerations
Not known

PR #6535, opened 2022-11-03
feat(cosmic-proto): query swingset params etc.

dckc:
## Description

In support of showing swingset execution fee batch size (#6525), generate client stubs for `proto/agoric/swingset/query.proto` to go with `msgs.proto`.

## Security considerations

The generated stubs have been only lightly reviewed; we presume they use the same code patterns we have looked at more closely for other stubs.

## Documentation considerations

We seem to rely on the .proto files as documentation. :-/

## Testing considerations

 - sanity test is extended to import `query.js`
 - an integration test (not run in ci) shows how to use `query.js`

### DIFF
--- a/packages/cosmic-proto/gen/agoric/swingset/query.js
+++ b/packages/cosmic-proto/gen/agoric/swingset/query.js
@@ -1,0 +1,336 @@
+/* eslint-disable */
+import { Params, Egress } from './swingset.js';
+import Long from 'long';
+import _m0 from 'protobufjs/minimal.js';
+export const protobufPackage = 'agoric.swingset';
+function createBaseQueryParamsRequest() {
+  return {};
+}
+export const QueryParamsRequest = {
+  encode(_, writer = _m0.Writer.create()) {
+    return writer;
+  },
+  decode(input, length) {
+    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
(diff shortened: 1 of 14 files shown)


samsiegart (approved):

turadg:
Note that the "swingset" files aren't making it to the NPM registry.

They're packed as symlinks and I suppose the registry omits those for cross-platform. Cc @kriskowal 

UPDATE: https://github.com/Agoric/agoric-sdk/pull/6521#issuecomment-1303743451

samsiegart:
> Note that the "swingset" files aren't making it to the NPM registry.
> 
> They're packed as symlinks and I suppose the registry omits those for cross-platform. Cc @kriskowal

For the wallet we might have to use the `agoric install` method that links the packages locally instead of getting them directly from npm, e.g.  https://github.com/Agoric/dapp-card-store/blob/main/.github/workflows/deploy-ui.yml#L14-L54

samsiegart:
> For the wallet we might have to use the `agoric install` method that links the packages locally instead of getting them directly from npm, e.g. https://github.com/Agoric/dapp-card-store/blob/main/.github/workflows/deploy-ui.yml#L14-L54

Note that I believe this requires an https://github.com/Agoric/dapp-card-store/blob/main/_agstate/agoric-servers/package.json file to exist in the dapp or else `agoric install` complains

Issue #4885, opened 2022-03-21
Enable `checkJs` to switch opt-in to opt-out for tsc


turadg:
## What is the Problem Being Solved?

Many files in the repo aren't type checked at all because they don't have `@ts-check` and our tsc compiler options are just to `allowJs` (implied by using `jsconfig.json`).

## Description of the Design

Enable `checkJs` option ([ref](https://www.typescriptlang.org/tsconfig#checkJs)) which "is the equivalent of including `// @ts-check` at the top of all JavaScript files which are included in your project".

Note that this is blocked on https://github.com/Agoric/agoric-sdk/issues/4560 because anything greater than the default `maxNodeModuleJsDepth` of zero means enabling `checkJs` will make tsc check inside the node_modules directory.

## Security Considerations
--
## Test Plan

--

turadg:
More more motivation: the `@ts-check` makes js apps that have SDK packages in node_modules re-check them. I haven't devised a way to prevent this other than patching the packages: https://github.com/Agoric/wallet-app/pull/34/commits/7cad8073b4529ced764a1809351c58bc430536a2

mhofman:
Weird, I didn't think tsc reported issues in `node_modules` path. Also to clarify, I suppose this wouldn't be a problem if all SDK packages had correctly exported types using generated `.d.ts` files?

Aka if we solved https://github.com/Agoric/agoric-sdk/issues/6343

turadg:
> Weird, I didn't think tsc reported issues in node_modules path

Example:
```
Errors  Files
     4  ../../node_modules/encoding/lib/encoding.js:59
    60  ../../node_modules/node-fetch/lib/index.js:3
     4  src/main.js:4
```
Many have more than that.

>  I suppose this wouldn't be a problem if all SDK packages had correctly exported types using generated .d.ts files?

I think that's right, because the `.d.ts` files would be read instead. But I'm not certain because I think I've seen some Endo packages that were converted over still have the problem. e.g. I think iI saw an error for`bundle-source`, though that's probably due to SDK being on the latest release (not dev) which lacks the types builds.

PR #6537, opened 2022-11-03
enable checkJs for many packages

turadg:
refs: #4885 (partial completion)

## Description

For key packages, flip the _opt-in_ `ts-check` to be opt-out and check all .js files by default. 

I expect we'll want to break this up into other PRs so I'm leaving this one as Draft. I have them all gathered to see how they interplay. The commits also depend on each other somewhat so it provides an ordering for landing.

Note that many of the types are still ambient. But the removal of ts-check reduces costs to package consumers, like https://github.com/Agoric/wallet-app/commit/7cad8073b4529ced764a1809351c58bc430536a2


Chris-Hibbert:
I looked at governance, inter-protocol, and zoe, and those changes would be easy to approve.

packages/SwingSet/src/devices/lib/deviceTools.js
```diff
@@ -69,6 +69,7 @@ export function buildSerializationTools(syscall, deviceName) {
     if (devnodeSlot) {
       return devnodeSlot;
     }
+    // @ts-expect-error DetailsToken not a string
     throw Error(X`unable to convert value ${val}`);
```
erights, line 73: See https://github.com/Agoric/agoric-sdk/pull/6538

erights: Btw, it is a victory of this PR that it triggered this ts error! This is the opportune time to notice whether such newly triggered diagnostics indicate a genuine problem that we previously missed.

turadg:
All changes here have landed through other PRs

PR #6540, opened 2022-11-04
fix cosmic-proto swingset exports

turadg:
## Description

As described in https://github.com/Agoric/agoric-sdk/pull/6521#issuecomment-1303743451

Third time's a charm!

explained in README

### Testing Considerations

```
❯ npm pack

> @agoric/cosmic-proto@0.2.1 prepack
> yarn rebuild

yarn run v1.22.19
$ rm -rf gen && mkdir -p gen && yarn generate && tsc --build && yarn prettier -w swingset
$ protoc --plugin=node_modules/.bin/protoc-gen-ts_proto --ts_proto_opt='esModuleInterop=true,forceLong=long,useOptionals=messages' --ts_proto_out=./gen --ts_proto_opt=importSuffix=.js ./proto/agoric/swingset/msgs.proto ./proto/agoric/swingset/query.proto -I./proto
$ /opt/agoric/agoric-sdk/node_modules/.bin/prettier -w swingset
swingset/msgs.d.ts 243ms
swingset/msgs.js 96ms
swingset/query.d.ts 39ms
swingset/query.js 20ms
swingset/swingset.d.ts 35ms
swingset/swingset.js 35ms
✨  Done in 6.00s.
npm notice
npm notice 📦  @agoric/cosmic-proto@0.2.1
npm notice === Tarball Contents ===
npm notice 11.4kB  LICENSE
npm notice 352B    README.md
npm notice 17.7kB  dist/agoric/swingset/msgs.d.ts
npm notice 17.9kB  dist/agoric/swingset/msgs.js
npm notice 14.2kB  dist/agoric/swingset/query.d.ts
npm notice 9.5kB   dist/agoric/swingset/query.js
npm notice 16.3kB  dist/agoric/swingset/swingset.d.ts
npm notice 16.9kB  dist/agoric/swingset/swingset.js
npm notice 3.3kB   dist/cosmos/base/v1beta1/coin.d.ts
npm notice 6.2kB   dist/cosmos/base/v1beta1/coin.js
npm notice 52B     dist/gogoproto/gogo.d.ts
npm notice 65B     dist/gogoproto/gogo.js
npm notice 53B     dist/google/api/annotations.d.ts
npm notice 66B     dist/google/api/annotations.js
npm notice 83.9kB  dist/google/api/http.d.ts
npm notice 9.0kB   dist/google/api/http.js
npm notice 4.6MB   dist/google/protobuf/descriptor.d.ts
npm notice 104.5kB dist/google/protobuf/descriptor.js
npm notice 1.8kB   package.json
npm notice === Tarball Details ===
npm notice name:          @agoric/cosmic-proto
npm notice version:       0.2.1
npm notice filename:      @agoric/cosmic-proto-0.2.1.tgz
npm notice package size:  134.0 kB
npm notice unpacked size: 5.0 MB
npm notice shasum:        9e1e9c84709280241e06a4d27cfc0b8ba5536df4
npm notice integrity:     sha512-Z9LI2qzsVwnTF[...]0Ndf7WA1Zxm5A==
npm notice total files:   19
npm notice
```


kriskowal (approved):
A bit of a shame to have build products checked in.

I’m surprised no further ceremony was necessary to mute the `useOptional=messages` warnings. When I tried that, I ran into some problem or another. It looked like it produced a breaking change in the generated code.

turadg:
While looking into this more I realized what I submitted wouldn't solve my downtream problem because it has `.ts` files which will run into https://github.com/microsoft/TypeScript/issues/47387#issuecomment-1010123047

The solution is for NPM packages to distribute only `.js` and `.d.ts` files. So I made some changes here to do that. It has a `dist` directory that the `.d.ts` and `.js` get build to, from the `.ts` in `gen`. `gen` doesn't need to stay in CI because it's just an interim build artifact. By some ideal `dist` wouldn't either but by putting it in we make it so agoric-sdk consumers don't have to depend on the build state of this package.

packages/cosmic-proto/package.json
```diff
@@ -15,19 +15,25 @@
...
+      "default": "./dist/agoric/swingset/msgs.js"
+    },
+    "./swingset/query.js": {
+      "types": "./dist/agoric/swingset/query.d.ts",
+      "default": "./dist/agoric/swingset/query.js"
+    },
+    "./swingset/swingset.js": {
+      "types": "./dist/agoric/swingset/swingset.d.ts",
+      "default": "./dist/agoric/swingset/swingset.js"
+    }
```
kriskowal: Could we reduce duplication by exporting the copy in `swingset` and not include `dist` in the published artifact? Multiple physical locations could effect multiple module instances and an identity discontinuity between them in the same program.

kriskowal (approved):

packages/cosmic-proto/package.json
turadg: Good call. We don't really need `dist` in the source tree. Force-pushed this change and will merge if it passes CI.

packages/cosmic-proto/package.json
```diff
@@ -37,13 +43,12 @@
     "protobufjs": "^7.0.0"
   },
   "devDependencies": {
-    "ts-proto": "^1.121.1"
+    "ts-proto": "^1.131.0"
   },
   "files": [
-    "build",
+    "dist",
```
kriskowal, line 49:
```suggestion
```

kriskowal (approved):

packages/cosmic-proto/package.json
turadg: well I spoke too soon. integrating with master, `swingset` now depends on `cosmos` which is at a higher level such that copying it to root would require moving `swingset`. I had to revert to the symlink approach.

So all this does now is add `dist` for consumers of the package through NPM. Keeps symlinks for peer packages in the repo.

kriskowal (approved):
Whatever works.

PR #6543, opened 2022-11-04
feat: add upgrade simulation tools to test environment

### LINKED ISSUE #6523: framework to unit-test vat upgrade and durable data

Currently it is a hassle to write tests of vat/library upgradability. The one scheme we have is more of an integration test, where you use a real swingset kernel and build a pair of vats (one pre- and one post- upgrade), import the library you're testing into those, have a controller/bootstrap vat drive an upgrade and a bunch of queries, log the output, and then the only real `t.deepEquals()` -style AVA assertion is to compare the output log against a "golden master".

These tend to be slow, coarse, hard to run under a debugger, and do not provide useful coverage data.

It would be great if we could write more "unit-ish" unit tests: smaller, faster, exercising finer details.

## Description of the Design

This kind of code is importing from `@agoric/vat-data` to get the primary virtual/durable tools:
* `defineKind` (times `Durable` and `Multi` variations)
* `makeKindHandle`
* `providePromiseWatcher`/`watchPromise`
* `makeScalarBigMapStore` (times `Set` and `Weak` variations)
* `canBeDurable`

Those all come from `globalThis.VatData`, which is provided to real vats by the swingset `liveslots` layer (since they must all integrate tightly with the Presences and Representatives that liveslots manages).

(`@agoric/vat-data` also exports secondary provide/vivify tools, like `provideDurableMapStore`, `defineVirtualFarClass`, `vivifyKind`, and `vivifyFarClass`, but these are all wrappers around the core functions that come from `VatData`)

We have a "fake virtual stuff" utility already, to unit-test libraries that import from `@agoric/vat-data` (or code that uses such libraries). Your unit test starts with `import '@agoric/swingset-vat/tools/prepare-test-env.js'` (or `prepare-test-env-ava.js`, since we always use AVA), and that use a function named `makeFakeVirtualStuff` to populate `globalThis.VatData` with enough code to support `defineKind`/etc.

The backing store for `makeFakeVirtualStuff()` is a plain `Map`, which takes the place of the `syscall.vatstoreSet(key,value)` that liveslots would use to keep data in the DB. `makeFakeVirtualStuff()` then creates a VirtualObjectManager and CollectionManager around this Map, asks them for `defineKind` and friends, and populates `globalThis.VatData` with those utilities. As long as this happens before any user-level code is imported (and thus imports `@agoric/vat-data`), the utilities will be in place on the global in time to be imported by user code.

My plan is:
* change `fakeVirtualSupport.js`:
  * add an option to supply the `Map` instead of always creating a local/private one
* change `prepare-test-env-ava.js`:
  * create a single Map at import time, and pass it to `makeFakeVirtualStuff`, so the Map can be reused later
  * add a `simulateUpgrade` function, which calls `makeFakeVirtualStuff()` again but with the same Map as before, to represent the DB that is retained across an upgrade
    * this will return a `baggage` object (a BigMapStore), which uses a known/stable ID within the DB/Map
    * each time `simulateUpgrade` is called, it creates a new `baggage`, backed by the new VOM/etc, but because it uses the same ID, it will return equivalent contents
  * build wrappers around everything we add to `VatData` so they can be replaced by `simulateUpgrade()`
    * e.g. `VatData.defineKind = (...args) => currentVOM.defineKind(...args)`, etc
    * because `@agoric/vat-data` re-exports the pieces of `globalThis.VatData`, and downstream code imports those pieces from `vat-data`, so their identities are immutable

A unit test that wants to exercise an upgrade would import the code-under-test and `prepare-test-env-ava.js` as usual, then in the test case:
* `baggage1 = simulateUpgrade()`
* call the library's normal initialization function (passing in `baggage1`) which will do `defineKind`/provide/etc, and return some `apiV1` object to interact with
* interact with `apiV1` to build up some durable state
* `baggage2 = simulateUpgrade()`
* call the library's init function again (this time with `baggage2`), which will re-do the define/provides/etc, and return a new `apiV2` object
  * this time, the library's `provide()` calls will observe the pre-existing state, and refrain from repeating `defineKind` and the other "once per vat, not once per incarnation" actions
* interact with `apiV2` to confirm that the durable state is still present

I think this approach will work, but it has some significant limitations, most of which derive from our (justified) choice to expose these stateful utilities (`defineKind`/`makeScalarBigMapStore`/etc) as imports, rather than passing them into library code as arguments:

* we cannot tolerate parallel execution of multiple unit tests that want to `simulateUpgrade`, since they all reference the same global state
  * test files which use this tool will probably need `test.serial` on every test
  * forgetting `test.serial` will cause hard-to-debug random failures, or worse
* the test function must be careful to never interact with `apiV1` again after the `simulateUpgrade()`, because that version will be accessing stale data
  * we might make this safer by building the `prepare-test-env.js` wrappers to stop working after `simulateUpgrade()`, but I think that would be tricky: instead of only wrapping `defineKind`/etc, we'd need to wrap the values they return, i.e. a full membrane instead of a shallow single-level wrapper

## Security Considerations

None, this is only to support unit tests

## Test Plan

Write an example test in `packages/SwingSet/`, advertise it to authors of other libraries that use virtual/durable data.

warner:
closes #6523

### DIFF
--- a/packages/SwingSet/test/test-simulate-upgrade.js
+++ b/packages/SwingSet/test/test-simulate-upgrade.js
@@ -1,0 +1,42 @@
+import { test, simulateIncarnation } from '../tools/prepare-test-env-ava.js';
+
+// eslint-disable-next-line import/order
+import { provideDurableMapStore, prepareKind } from '@agoric/vat-data';
+
+const initFoo = () => ({ count: 0 });
+const fooBehavior = {
+  increment: ({ state }) => {
+    state.count += 1;
+  },
+  read: ({ state }) => state.count,
+};
+
+test('simulate upgrade', async t => {
+  await simulateIncarnation(async baggage => {
+    // first incarnation
+    const s = provideDurableMapStore(baggage, 'store');
+    const makeFoo = prepareKind(baggage, 'foo', initFoo, fooBehavior);
+
+    s.init('key1', 'value1');
+
+    const foo1 = makeFoo();
+    s.init('foo1', foo1);
+    foo1.increment();
+    foo1.increment();
+    t.is(foo1.read(), 2);
+  });
+
+  await simulateIncarnation(async baggage => {
+    // second incarnation
+    const s = provideDurableMapStore(baggage, 'store');
+    const makeFoo = prepareKind(baggage, 'foo', initFoo, fooBehavior);
+
+    t.is(s.get('key1'), 'value1');
+
+    const foo1 = s.get('foo1');
+    t.is(foo1.read(), 2);
+    foo1.increment();
+    t.is(foo1.read(), 3);
+    t.is(makeFoo().read(), 0);
+  });
+});
--- a/packages/SwingSet/tools/prepare-test-env-ava.js
+++ b/packages/SwingSet/tools/prepare-test-env-ava.js
@@ -13,6 +13,8 @@ import '@endo/ses-ava/exported.js';
 import { wrapTest } from '@endo/ses-ava';
 import rawTest from 'ava';
 
+export { simulateIncarnation } from './prepare-test-env.js';
+
 /** @type {typeof rawTest} */
 // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
 // @ts-ignore XXX https://github.com/endojs/endo/issues/1235
--- a/packages/SwingSet/tools/prepare-test-env.js
+++ b/packages/SwingSet/tools/prepare-test-env.js
@@ -11,39 +11,78 @@ import '@endo/init/pre-bundle-source.js';
 import './install-ses-debug.js';
 import { makeFakeVirtualStuff } from '@agoric/swingset-liveslots/tools/fakeVirtualSupport.js';
 
-const { vom, cm, wpm } = makeFakeVirtualStuff({ cacheSize: 3 });
+function makeEverything(options) {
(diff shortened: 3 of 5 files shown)


warner:
Not ready for review, but ready to experiment with. Names are in flux (maybe `runSimulatedIncarnation` instead of `simulateIncarnation`? or something with "upgrade" in the name?). Test needs to be more thorough. Docs need pointer to the test to reference as an example for downstream users to follow.

warner:
To remind myself: this also needs a safety check: a counter that is elevated while we're inside `simulateIncarnation()`, and asserts if it is not 0 upon entry, to guard against accidental reentrancy (which would confuse things horribly).

I should also decide how I want to handle the "default incarnation" that you get if you interact with `vat-data` but never call `simulateIncarnation`.

warner:
Ok that adds the safety check. I don't have an answer for interaction between the default incarnation and the `simulateIncarnation` one, maybe the best advice is "don't do that".

@gibson042 I could use the changes to `fakeVirtualSupport.js` for other testing I'm doing, so I'd like to land at least those portions. If you're comfortable with this `simulateIncarnation` (and the way it works into your other plans for userspace testing support), we can land it too, or I'm happy to remove it from this PR if it would be more harm than good.

warner:
Actually, it looks like I can do my testing with other tools, so feel free to defer this until we've got something at aligns with your plans the best.

PR #6544, opened 2022-11-04
checkJs swingset-vat

turadg:

## Description

Note that this omits `test` from tsconfig so that `checkJs` can be enabled. Only one test file has `ts-check` so it's not much loss. To add `test` back in with `checkJs` would require fixing the "Found 938 errors in 76 files" or adding `@ts-nocheck` in files that aren't deemed worthy to fix (e.g. 161 errors in `test-liveslots.js`).

### Testing Considerations

CI

### DIFF
--- a/packages/SwingSet/src/kernel/slogger.js
+++ b/packages/SwingSet/src/kernel/slogger.js
@@ -98,6 +98,7 @@ export function makeDummySlogger(slogCallbacks, dummyConsole) {
     write: () => 0,
   });
   doneRegistering(`Unrecognized makeDummySlogger slogCallbacks names:`);
+  // @ts-expect-error xxx
   return dummySlogger;
 }
 
@@ -281,5 +282,6 @@ export function makeSlogger(slogCallbacks, writeObj) {
     write,
   });
   doneRegistering(`Unrecognized makeSlogger slogCallbacks names:`);
+  // @ts-expect-error xxx
   return slogger;
 }
--- a/packages/SwingSet/src/kernel/state/deviceKeeper.js
+++ b/packages/SwingSet/src/kernel/state/deviceKeeper.js
@@ -31,15 +31,15 @@ export function initializeDeviceState(kvStore, deviceID) {
  *
  * @param {*} kvStore  The storage in which the persistent state will be kept
  * @param {string} deviceID  The device ID string of the device in question
- * @param { addKernelDeviceNode: (deviceID: string) => string,
+ * @param {{ addKernelDeviceNode: (deviceID: string) => string,
  *          incrementRefCount: (kernelSlot: string,
  *                              tag: string?,
  *                              options: {
- *                                isExport: boolean?,
- *                                onlyRecognizable: boolean?,
+ *                                isExport?: boolean,
+ *                                onlyRecognizable?: boolean,
  *                              },
- *                             ) => undefined),
- *         } tools
+ *                             ) => void,
+ *         }} tools
  * @returns {*} an object to hold and access the kernel's state for the given device
  */
 export function makeDeviceKeeper(kvStore, deviceID, tools) {
@@ -185,6 +185,7 @@ export function makeDeviceKeeper(kvStore, deviceID, tools) {
    * @returns {Array<[string, string, string]>} an array of this device's state information
    */
   function dumpState() {
+    /** @type {Array<[string, string, string]>} */
     const res = [];
     const prefix = `${deviceID}.c.`;
     for (const k of kvStore.getKeys(prefix, `${deviceID}.c/`)) {
--- a/packages/SwingSet/src/kernel/state/reachable.js
+++ b/packages/SwingSet/src/kernel/state/reachable.js
@@ -14,6 +14,7 @@ export function parseReachableAndVatSlot(value) {
   } else {
     assert(`flag (${flag}) must be 'R' or '_'`);
   }
+  // @ts-expect-error xxx TS doesn't know isReachable is assigned b/c assert() throws
   return { isReachable, vatSlot };
 }
 harden(parseReachableAndVatSlot);
--- a/packages/SwingSet/src/kernel/state/reachable.js
+++ b/packages/SwingSet/src/kernel/state/reachable.js
@@ -12,9 +12,8 @@ export function parseReachableAndVatSlot(value) {
   } else if (flag === '_') {
     isReachable = false;
   } else {
-    assert(`flag (${flag}) must be 'R' or '_'`);
+    assert.fail(`flag (${flag}) must be 'R' or '_'`);
   }
-  // @ts-expect-error xxx TS doesn't know isReachable is assigned b/c assert() throws
   return { isReachable, vatSlot };
 }
 harden(parseReachableAndVatSlot);
--- a/packages/SwingSet/src/kernel/vat-loader/manager-subprocess-node.js
+++ b/packages/SwingSet/src/kernel/vat-loader/manager-subprocess-node.js
@@ -78,7 +78,7 @@ export function makeNodeSubprocessFactory(tools) {
       } else if (type === 'dispatchReady') {
         parentLog(`dispatch() ready`);
         // wait10ms().then(dispatchIsReady); // stall to let logs get printed
-        dispatchIsReady();
+        dispatchIsReady(undefined);
       } else if (type === 'syscall') {
         parentLog(`syscall`, args);
         const [vatSyscallObject] = args;
--- a/packages/SwingSet/src/lib-nodejs/waitUntilQuiescent.js
+++ b/packages/SwingSet/src/lib-nodejs/waitUntilQuiescent.js
@@ -11,6 +11,7 @@ export function waitUntilQuiescent() {
   // lower-priority than the Promise queue on browsers and Node 11, but on
   // Node 10 it is higher. So this trick requires Node 11.
   // https://jsblog.insiderattack.net/new-changes-to-timers-and-microtasks-from-node-v11-0-0-and-above-68d112743eb3
+  /** @type {PromiseKit<void>} */
   const { promise: queueEmptyP, resolve } = makePromiseKit();
   setImmediate(() => resolve());
   return queueEmptyP;
--- a/packages/SwingSet/src/lib-nodejs/worker-protocol.js
+++ b/packages/SwingSet/src/lib-nodejs/worker-protocol.js
@@ -28,14 +28,19 @@ export function arrayEncoderStream() {
 }
 
 export function arrayDecoderStream() {
+  /**
+   *
+   * @param {Buffer} buf
+   * @param {BufferEncoding} encoding
+   * @param {*} callback
+   */
   function transform(buf, encoding, callback) {
     let err;
     try {
       if (!Buffer.isBuffer(buf)) {
         throw Error('stream expects Buffers');
       }
-      // @ts-expect-error Buffer not assignable to string
-      this.push(JSON.parse(buf));
+      this.push(JSON.parse(buf.toString()));
     } catch (e) {
       err = e;
     }
--- a/packages/SwingSet/src/lib-nodejs/worker-protocol.js
+++ b/packages/SwingSet/src/lib-nodejs/worker-protocol.js
@@ -5,6 +5,12 @@ import { Transform } from 'stream';
 // data into Buffers suitable for netstring conversion.
 
 export function arrayEncoderStream() {
+  /**
+   *
+   * @param {*} object
+   * @param {BufferEncoding} encoding
+   * @param {*} callback
+   */
   function transform(object, encoding, callback) {
     if (!Array.isArray(object)) {
       throw Error('stream requires Arrays');
@@ -28,6 +34,7 @@ export function arrayDecoderStream() {
       if (!Buffer.isBuffer(buf)) {
         throw Error('stream expects Buffers');
       }
+      // @ts-expect-error Buffer not assignable to string
       this.push(JSON.parse(buf));
     } catch (e) {
       err = e;
--- a/packages/SwingSet/src/lib/storageAPI.js
+++ b/packages/SwingSet/src/lib/storageAPI.js
@@ -137,7 +137,7 @@ function* mergeUtf16SortedIterators(it1, it2) {
  *   onPendingDelete?: (key: string) => void, // a callback invoked after a new uncommitted delete
  *   onCommit?: () => void, // a callback invoked after pending operations have been committed
  *   onAbort?: () => void, // a callback invoked after pending operations have been aborted
- * }?} listeners  Optional callbacks to be invoked when respective events occur
+ * }} listeners  Optional callbacks to be invoked when respective events occur
  *
  * @returns {{kvStore: KVStore, commit: () => void, abort: () => void}}
  */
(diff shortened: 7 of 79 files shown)


mhofman:
I have a stash of type improvements to SwingSet I had after manually enabling type checks on some files. Will bring them along here.

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -7,6 +7,7 @@ import {
 import { assert, details as X } from '@agoric/assert';
 import { isNat } from '@agoric/nat';
 import { isPromise } from '@endo/promise-kit';
+import { HandledPromise } from '@endo/eventual-send';
```
mhofman, line 10: I believe this is erroneous, and `HandledPromise` is currently expected to be available on the global, installed by `@endo/init`, which is evaluated as part of the supervisor bundle before liveslots.

`HandledPromise` should correctly be typed as a global, maybe the file where this is declared doesn't get transitively imported?

packages/SwingSet/src/kernel/vat-loader/manager-subprocess-node.js
```diff
@@ -78,7 +78,7 @@ export function makeNodeSubprocessFactory(tools) {
       } else if (type === 'dispatchReady') {
         parentLog(`dispatch() ready`);
         // wait10ms().then(dispatchIsReady); // stall to let logs get printed
-        dispatchIsReady();
+        dispatchIsReady(undefined);
```
mhofman, line 81: Ugh I hate this `void` incompatibility with `any` rule, and the runtime changes it requires. I wish there was an easy way to inform type params of generic functions in JSDoc. I may prefer if we "cast" the `makePromiseKit` call to `PromiseKit<void>`

packages/SwingSet/src/lib-nodejs/waitUntilQuiescent.js
```diff
@@ -12,6 +12,6 @@ export function waitUntilQuiescent() {
   // Node 10 it is higher. So this trick requires Node 11.
   // https://jsblog.insiderattack.net/new-changes-to-timers-and-microtasks-from-node-v11-0-0-and-above-68d112743eb3
   const { promise: queueEmptyP, resolve } = makePromiseKit();
-  setImmediate(() => resolve());
+  setImmediate(() => resolve(undefined));
```
mhofman: Same, can we cast the `makePromiseKit` result instead?

packages/SwingSet/tools/vat.js
```diff
@@ -27,10 +27,12 @@ async function main() {
     argv[0] === '--' || argv[0] === undefined ? '.' : argv.shift();
   const vatArgv = argv[0] === '--' ? argv.slice(1) : argv;
 
+  assert(basedir);
   const config = await loadBasedir(basedir);
   const { loopboxSrcPath, loopboxEndowments } = buildLoopbox('immediate');
   config.devices = [['loopbox', loopboxSrcPath, loopboxEndowments]];
 
+  // @ts-expect-error expects string[], not boolean, in second position
```
mhofman, line 35: That seems like a legitimate error that shouldn't silenced

packages/SwingSet/src/kernel/state/deviceKeeper.js
```diff
@@ -199,6 +199,7 @@ export function makeDeviceKeeper(kvStore, deviceID, tools) {
         }
       }
     }
+    // @ts-expect-error may have length less than three
```
mhofman: Please type the `res` declaration instead:

```js
    /** @type {Array<[string, string, string]>} */
    const res = [];
```

packages/SwingSet/src/kernel/state/reachable.js
```diff
@@ -14,6 +14,7 @@ export function parseReachableAndVatSlot(value) {
   } else {
     assert(`flag (${flag}) must be 'R' or '_'`);
   }
+  // @ts-expect-error xxx TS doesn't know isReachable is assigned b/c assert() throws
```
mhofman: That seems to be a legitimate mistake uncovered by tsc

```suggestion
    assert.fail(`flag (${flag}) must be 'R' or '_'`);
  }
```

packages/SwingSet/src/kernel/slogger.js
```diff
@@ -281,5 +282,6 @@ export function makeSlogger(slogCallbacks, writeObj) {
     write,
   });
   doneRegistering(`Unrecognized makeSlogger slogCallbacks names:`);
+  // @ts-expect-error xxx
```
mhofman, line 285: Ugh the console/logger mistypings were intractable last time I attempted some fixes. Hopefully someday we can sort them out.

packages/SwingSet/src/lib/storageAPI.js
```diff
@@ -163,6 +163,7 @@ export function makeBufferedStorage(kvStore, listeners = {}) {
       if (additions.has(key)) return additions.get(key);
       if (deletions.has(key)) return undefined;
       const value = kvStore.get(key);
+      // @ts-expect-error value may be undefined
       if (onGet !== undefined) onGet(key, value);
```
mhofman, line 167: That seems like a potential issue. @gibson042 ?

packages/SwingSet/src/lib-nodejs/worker-protocol.js
```diff
@@ -28,6 +28,7 @@ export function arrayDecoderStream() {
       if (!Buffer.isBuffer(buf)) {
         throw Error('stream expects Buffers');
       }
+      // @ts-expect-error Buffer not assignable to string
       this.push(JSON.parse(buf));
```
mhofman: Maybe we should be explicit and call `buf.toString()` or `String(buf)` instead of relying on the implicit conversion?

packages/SwingSet/src/liveslots/liveslots.js
```diff
@@ -1496,6 +1496,7 @@ function build(
     vom.flushCache();
     await gcTools.gcAndFinalize();
     const doMore = await scanForDeadObjects();
+    // @ts-expect-error FIXME doMore is void
```
mhofman, line 1499: @FUDCo sounds like `scanForDeadObjects` is missing a `return` ?

packages/SwingSet/src/vats/comms/state.js
```diff
@@ -704,6 +704,7 @@ export function makeState(syscall) {
     store.set(`${lpid}.data.slots`, data.slots.join(','));
     let idx = 0;
     for (const slot of data.slots) {
+      // @ts-expect-error may not be string
       incrementRefCount(slot, `resolve|s${idx}`);
```
mhofman: @warner do we actually ever use `CapData` with a `Slot` type that isn't `string`? What is the purpose of making it generic?

packages/SwingSet/test/util.js
```diff
@@ -45,7 +45,10 @@ export function dumpKT(kernel) {
   }
 }
 
-export function buildDispatch(onDispatchCallback = undefined) {
```
turadg, line 48: optionality conveyed by jsdoc so they type checker doesn't take this as a hint that the value is only ever `undefined`

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -42,6 +42,7 @@ export function buildRootObject(vatPowers, _vatParameters, baggage) {
   function noteRunningVat(vatID) {
     const prr = producePRR();
     const [doneP] = prr;
+    // @ts-expect-error may not have `catch`
```
mhofman: Please type `producePRR` to return a tuple instead of an array, e.g. with `/** @type {const} */` on the return value.

packages/SwingSet/src/vats/vat-admin/vat-vat-admin.js
```diff
@@ -330,6 +331,7 @@ export function buildRootObject(vatPowers, _vatParameters, baggage) {
     noteRunningVat(vatID);
 
     const adminNode = makeAdminNode(vatID);
+    // @ts-expect-error may not have `then`
```
mhofman: Same as above

packages/SwingSet/test/util.js
```diff
@@ -86,25 +89,37 @@ export function buildDispatch(onDispatchCallback = undefined) {
 }
 
 export function capSlot(index, iface = 'export') {
+  // @ts-expect-error undefined not assignable to type string
```
mhofman: Type the params instead to avoid `iface` inference to `string` ?

mhofman (changes requested):
Thanks for getting this started, I had meant to get better typing into SwingSet.

I would like to see some of the `ts-expect-error` handled differently. And I believe type checks actually exposed a couple legitimate issues that are not tagged as FIXME.

mhofman:
> I have a stash of type improvements to SwingSet I had after manually enabling type checks on some files. Will bring them along here.

Here is a crude rebase of that stash for future reference (not advocating for it to be taken as-is): https://github.com/Agoric/agoric-sdk/commit/94065955903e513302acb67eb25ec4f075c13c71

It may be worth including cosmic-swingset in this PR as its so interleaved with swingset for all the init flow.

packages/SwingSet/src/liveslots/liveslots.js
warner: liveslots currently gets imported by e.g. `src/supervsors/subprocess-xsnap/supervisor-subprocess-xsnap.js`, when building the supervisor bundle.

In the future, we might make a separate liveslots-only bundle, using `liveslots.js` as the entrypoint file.

Dunno how that impacts the type management. I concur with Mathieu that HandledPromise is generally expected to be a global. I don't see any problems with importing it as well, assuming we don't somehow wind up with a second version (identity discontinuity), but yeah it's probably best if it's just on the global.

packages/SwingSet/tools/vat.js
warner: BTW `vat.js` is probably the oldest file we have, and it's surely bitrotted badly by now. I haven't tried to run it in years.

packages/SwingSet/src/kernel/state/reachable.js
warner: oops, yeah, I should have written `assert.fail`

packages/SwingSet/src/liveslots/liveslots.js
warner: Hm, this looks like both functions think they're responsible for the re-scan.. I bet we didn't think this through completely when we introduced `scanForDeadObjects()`.

The `doMore` test is meant to deal with the case where `gcAndFinalize` shakes loose some in-RAM `Representative`s,  `scanForDeadObjects` looks through `possiblyDeadSet` and finds some actual vrefs it can retire (`vrm.possibleVirtualObjectDeath()`), some of those vrefs were used as keys in a (virtual) WeakMap, so deleting the keys can drop references to more in-RAM Representatives. We need a new `gcAndFinalize` pass to shake those loose, so we repeat the loop within `scanForDeadObjects()` until `possibleVirtualObjectDeath()` didn't free anything (and our other checks didn't free anything either).

The outer function(`bringOutYourDead`) looks like it was meant to do the same "repeat until nothing changes" pattern, although with recursion rather than a `do/while` loop. It does a (redundant?) extra `gcAndFinalize` at the expected time. I think the only real contribution of the outer loop, these days, is an extra call to `vom.flushCache()`, but I'm pretty sure that can't shake loose more objects like `doMore` is tracking, so I think it's not important to call from there.

@FUDCo is the authority here, but I'm thinking there are two correct fixes:

* 1: remove the `gcAndFinalize` and the recursion from `bringOutYourDead` (makie it just do `vom.flushCache(); await gcTools.gcAndFinalize(); await scanForDeadObjects(); return`, and rely upon the loop that happens inside `scanForDeadObjects`
* 2: remove the `do/while` from `scanForDeadObjects`, and modify it to return `doMore`, relying upon the recursion in `bringOutYourDead` to shake everything loose
  * if we go this way, I think I'd prefer to have `bringOutYourDead` use a do/while loop, instead of the recursion, since we can't rely on tail-call optimization

The benefit of doing the loop inside `scanForDeadObjects` (1) is that we only emit a single trio of syscalls, no matter how many times we have to loop. I can't think of a benefit for (2), other than that (1) makes `bringOutYourDead` start to look kind of useless and suggests we might refactor something

packages/SwingSet/src/vats/comms/state.js
warner: The kernel only ever uses strings, although in some places they're krefs and in others they're vrefs, (and comms also has rrefs and lrefs). I bet it'd catch some potentials bugs if we had distinct types for those.

I think `CapData` is actually defined by `marshal`, which doesn't enforce `typeof(slot)==='string'`, so `CapData` is a generic (parameterized by some type T, which the kernel always fills with `string`). I think we have a more-specialized `KernelCapData` or `SwingsetCapData` within `packages/SwingSet/`, since we can't use the generic one.

I can't think of a current use case for non-string slots, so I'd probably be ok if `marshal` went ahead and assumed they were always strings. But I also expect that the `marshal` folks (@erights) would think of that as unnecessarily narrow / overly constrained.

warner:
I'll defer to @mhofman for anything beyond the bits I mentioned above.

packages/SwingSet/src/liveslots/liveslots.js
turadg: > should correctly be typed as a global

Granted. Given that this solves the PR scope (enable `checkJs`) and that importing it doesn't create any problems, I propose deferring that as future work.

packages/SwingSet/src/kernel/vat-loader/manager-subprocess-node.js
turadg: Great idea. done

packages/SwingSet/tools/vat.js
turadg: To be clear, it was already silenced (by never being type-checked) so this is not a regression. The goal of this PR is to `checkJs` and unblock other `checkJs` PRs, not to solve every type error.

I leave it up to the package maintainers to prioritize fixing this. If you want I can change the comment to include FIXME or a link to a ticket someone requests that I include here.

packages/SwingSet/src/kernel/slogger.js
turadg: Yeah. If there's a ticket to do that I'll include it in the comment

packages/SwingSet/src/lib-nodejs/worker-protocol.js
turadg: Agreed.

packages/SwingSet/src/liveslots/liveslots.js
turadg: To unblock the other `checkJs` PRs I propose that this land with the FIXME comment and the package maintainers prioritize fixing it as they wish.

packages/SwingSet/src/vats/comms/state.js
turadg: Thanks for explaining that. A quick fix was to update the `insistCapData` assertion.

mhofman (approved):
Thanks for addressing that feedback!

PR #6546, opened 2022-11-08
4885 casting etc

turadg:

## Description

### Testing Considerations

CI


Chris-Hibbert (approved):
No opinion on .../packages/casting.

Why are there changes in `.../patches/`?

turadg:
> No opinion on .../packages/casting.

I'll merge and promise to address any requests that arrive after.

> Why are there changes in .../patches/?

Two are to eliminate the `@ts-check` that makes tsc try to validate them, and fail because ambient types are missing. The acorn one is to forward the patch to the version that got bumped.

PR #6547, opened 2022-11-08
4885 telemetry

turadg:

## Description

### Testing Considerations

CI

### DIFF
--- a/packages/telemetry/jsconfig.json
+++ b/packages/telemetry/jsconfig.json
@@ -3,23 +3,16 @@
   "compilerOptions": {
     "target": "esnext",
     "module": "esnext",
-
+    "checkJs": true,
     "noEmit": true,
-/*
-    // The following flags are for creating .d.ts files:
-    "noEmit": false,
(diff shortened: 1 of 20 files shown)


mhofman (changes requested):
Why the `yarn.lock` churn? Seems unrelated to this PR. This is my only blocking concern.

Besides that I'm not convinced by the new `@agoric/assert` dependency, especially for a fallback module, but I'm willing to let it go. I really wish JSDoc had non-null assertions to avoid such runtime cruft.

turadg:
> Why the yarn.lock churn? Seems unrelated to this PR. This is my only blocking concern.

Dunno, exactly. Maybe something with branching. I've rebased on master and cut out the lockfile changes.

> Besides that I'm not convinced by the new @agoric/assert dependency, especially for a fallback module, but I'm willing to let it go. I really wish JSDoc had non-null assertions to avoid such runtime cruft.

SGTM

mhofman (approved):

PR #6548, opened 2022-11-09
fix: simplify defineDurableHandle

erights:
Noticed while working on https://github.com/Agoric/agoric-sdk/pull/6106

### DIFF
--- a/packages/zoe/src/makeHandle.js
+++ b/packages/zoe/src/makeHandle.js
@@ -2,11 +2,7 @@
 
 import { assert } from '@agoric/assert';
 import { initEmpty, makeHeapFarInstance } from '@agoric/store';
-import {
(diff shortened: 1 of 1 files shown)


Chris-Hibbert:
I'm working on a massive refactoring of all of Zoe. I don't think it will be helpful to submit small fixes in parallel.  I'll fix this if it's still present in my branch.

Chris-Hibbert (changes requested):
'request changes' seems to be the way to block submission.

erights:
> I'm working on a massive refactoring of all of Zoe. I don't think it will be helpful to submit small fixes in parallel. I'll fix this if it's still present in my branch.

Seems like a good reason to close this PR. Thanks.

Issue #6489, opened 2022-10-21
Support Node 18


mhofman:
## What is the Problem Being Solved?

Node 18 is [reaching LTS imminently](https://github.com/nodejs/Release#release-schedule). We should make sure everything works when using node 18.

## Description of the Design

Setup a dev env with node 18 and see what breaks
Update CI to also run against node 18.

## Security Considerations
None

## Test Plan
Yes

sigv:
When trying to run the `pismoA` tag with Node.js 18.x we are observing an initialization error:

```none
launch-chain: Launching SwingSet kernel
portHandler threw (Error#1)
Error#1: You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer
  at async Promise.all (index 0)
  at async Promise.all (index 6)
  at async allValues (packages/SwingSet/src/controller/initializeSwingset.js:29:30)
  at async buildVatAndDeviceBundles (packages/SwingSet/src/controller/initializeSwingset.js:54:19)
  at async initializeSwingset (packages/SwingSet/src/controller/initializeSwingset.js:336:25)
  at async ensureSwingsetInitialized (packages/cosmic-swingset/src/launch-chain.js:113:5)
  at async buildSwingset (packages/cosmic-swingset/src/launch-chain.js:115:3)
  at async launch (packages/cosmic-swingset/src/launch-chain.js:264:52)
  at async launchAndInitializeSwingSet (packages/cosmic-swingset/src/chain-main.js:412:15)
  at async toSwingSet (packages/cosmic-swingset/src/chain-main.js:461:22)
  at async portHandlers.<computed> (packages/cosmic-swingset/src/chain-main.js:159:16)
Cannot initialize Controller Error: You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer
Error: exit status 1
```

This seems to be caused by [mozilla/source-map#454](https://github.com/mozilla/source-map/issues/454#issuecomment-1147637420), fixed with [`source-map@0.7.4`](https://www.npmjs.com/package/source-map/v/0.7.4).
[`source-map@0.7.3`](https://www.npmjs.com/package/source-map/v/0.7.3) is in place in repository:

```none
$ npm ls source-map
@agoric/sdk@ /home/agoric/source/agoric-upgrade-8
[...]

PR #6549, opened 2022-11-09
WIP experiment-only: experimentally regen yarn.lock

erights:
Just seeing what happens if I delete yarn.lock and regenerate it from scratch using `yarn && yarn build`

similar: https://github.com/endojs/endo/pull/1354

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -30,15 +30,15 @@
     "@jridgewell/trace-mapping" "^0.3.9"
 
 "@babel/cli@^7.12.13":
-  version "7.17.10"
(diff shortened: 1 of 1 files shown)


erights:
CI failures so far all due to
> Warning: patch-package detected a patch file version mismatch

which makes sense.

erights:
See also https://github.com/Agoric/agoric-sdk/issues/6489#issuecomment-1308215713 

> Curiously, in the experiment at https://github.com/Agoric/agoric-sdk/pull/6549 , when I locally run the same
> ```sh
> yarn test test/swingsetTests/makeKind/test-makeKind.js
> ```
> under Node v18, it works.

erights:
Too stale to be relevant. Closing.

PR #6550, opened 2022-11-09
export non-ambient types in ERTP

turadg:
refs: #6343

## Description

ERTP part of https://github.com/Agoric/agoric-sdk/issues/6343

Review by commit suggested.

### Security Considerations

n/a

Improves API documentation

### Testing Considerations

CI

### DIFF
--- a/packages/inter-protocol/src/feeDistributor.js
+++ b/packages/inter-protocol/src/feeDistributor.js
@@ -8,7 +8,7 @@ const { details: X } = assert;
 /**
  * @typedef {object} FeeCollector
  *
- * @property {() => ERef<Payment>} collectFees
+ * @property {() => ERef<Payment<'nat'>>} collectFees
  */
 
 /**
@@ -107,8 +107,8 @@ export const makeShareConfig = (destinations = {}, keywordShares = {}) => {
 /**
  * Split and deposit a payment given a share configuration.
  *
- * @param {Payment} payment
- * @param {ERef<Issuer>} issuer
+ * @param {Payment<'nat'>} payment
+ * @param {ERef<Issuer<'nat'>>} issuer
  * @param {ShareConfig} shareConfig
  */
 export const sharePayment = async (
@@ -158,7 +158,7 @@ export const sharePayment = async (
 };
 
 /**
- * @param {ERef<Issuer>} feeIssuer
+ * @param {ERef<Issuer<'nat'>>} feeIssuer
  * @param {{ keywordShares: Record<Keyword, bigint>, timerService: ERef<TimerService>, collectionInterval: RelativeTime}} terms
  */
 export const makeFeeDistributor = (feeIssuer, terms) => {
@@ -171,7 +171,7 @@ export const makeFeeDistributor = (feeIssuer, terms) => {
   /** @type {SetStore<FeeCollector>} */
   const collectors = makeScalarSetStore();
 
-  /** @param {Payment} payment */
+  /** @param {Payment<'nat'>} payment */
   const distributeFees = payment =>
     sharePayment(payment, feeIssuer, shareConfig);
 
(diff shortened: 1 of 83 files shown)


packages/vats/src/core/types.js
```diff
@@ -157,11 +157,11 @@
  * @typedef {{
  *   issuer: {
  *     produce: Record<WellKnownName['issuer'], Producer<Issuer>>,
- *     consume: Record<WellKnownName['issuer'], Promise<Issuer>>,
+ *     consume: Record<WellKnownName['issuer'], Promise<Issuer>> & { BLD: Promise<Issuer<'nat'>>, IST: Promise<Issuer<'nat'>> },
```
turadg, line 160: PTAL @dckc

packages/zoe/src/contractFacet/types.js
```diff
@@ -81,7 +81,7 @@
  *
  * @param {ERef<Issuer>} issuerP Promise for issuer
  * @param {Keyword} keyword Keyword for added issuer
- * @returns {Promise<IssuerRecord>} Issuer is added and ready
+ * @returns {Promise<IssuerRecord<K>>} Issuer is added and ready
```
turadg, line 84: `any` for keyword records since there's no way to know the actual type. We could keep `AssetKind` as the parameter, but that kind of means "all four of these" instead of "dunno"

packages/vats/src/core/types.js
dckc: hm. yes, I guess that's reasonable.

packages/zoe/src/contractFacet/types.js
dckc: yeah; keyword records and static types don't get along well.

packages/zoe/test/unitTests/contracts/test-coveredCall.js
```diff
@@ -1000,9 +1001,13 @@ test('zoe - coveredCall non-fungible', async t => {
     `The option was exercised. Please collect the assets in your payout.`,
   );
 
+  /** @type {Payment<any>} */
```
dckc, line 1004: Why is this needed?

packages/inter-protocol/src/feeDistributor.js
```diff
@@ -171,9 +171,13 @@ export const makeFeeDistributor = (feeIssuer, terms) => {
...
 
-  /** @param {Payment} payment */
+  /** @param {Payment<'nat'>} payment */
   const distributeFees = payment =>
-    sharePayment(payment, feeIssuer, shareConfig);
+    sharePayment(
+      payment,
+      /** @type {Issuer<'nat'>} */ (feeIssuer),
+      shareConfig,
+    );
```
dckc: this looks a little fiddly... help me understand why/if it's necessary?

packages/inter-protocol/src/my-lien.js
```diff
@@ -28,7 +28,7 @@ const XLien = /** @type { const } */ ({
  */
 export const makeStakeReporter = (bridgeManager, brand, denom = 'ubld') => {
   const { make: makeAmt } = AmountMath;
-  /** @param {string} numeral */
+  /** @type {(numeral: string) => Amount<'nat'>} */
```
dckc: I gather this isn't needed.

packages/inter-protocol/test/test-voPool.js
```diff
@@ -62,7 +62,7 @@ const voPoolTest = async (t, mutation, postTest) => {
 
   const { brand: secondaryBrand, issuer: secondaryI } =
     makeIssuerKit('secondary');
-  const quoteIssuerKit = makeIssuerKit('Quotes');
+  const quoteIssuerKit = makeIssuerKit('Quotes', 'set');
```
dckc, line 65: nifty... bug fix.

dckc (approved):
Looks like while scrubbing a bunch of stuff, the types were refined too. In some cases, it might add some burden/friction... but inter-protocol should be somewhat stable now, so that's probably worthwhile.

PR #6556, opened 2022-11-10
allow any string as offer id

### LINKED ISSUE #6551: remove server side constraint on unique offer ids

Each offer to the smart wallet has a unique offer ID, which fulfills two requirements:
1. Offer status publishing on RPC reflects actual offers pushed. (No collisions.)
2. Maps to look up offer results (`offerToUsedInvitation`, `offerToInvitationMakers`) have a stable unique key.

To avoid keeping track of every offer id, we opted for a high water mark of a serial number:
https://github.com/Agoric/agoric-sdk/blob/6cdbe3fee568f26ff13c27271bdaa3bbd751f4f4/packages/smart-wallet/src/smartWallet.js#L184-L186

## Description of the Design

Requirement #1 is more the concern of the client, so leave it to them. If you push an offer with a colliding ID, vstorage publishing is undefined. It's easy to ensure they're unique by using UUID or a high-precision timestamp.

Requirement #2 already has a collection of ids that could collide, in the keys of those maps. If an offer will collide there, through an exception.

## Security Considerations

--

## Test Plan

This should be backwards compatible, but do integration testing with the Wallet UI.

turadg:
closes: #6551

## Description

Stops validating uniqueness of offerId, leaving that to clients.

Allow it to be a string so clients can use a UUID.

Maintains `getLastOfferId` and the RPC data shape for clients that assume it, with a dummy/error value. The actual breaking change will be removing the deprecated functionality, in https://github.com/Agoric/agoric-sdk/issues/5605


dckc (approved):
looks good.

Some things I thought about:
  - Q: What's the compatibility impact of merging contract changes now?
    - A: essentially none; we'll restart the whole swingset VM before deploying this
  - Q: The screw case we saw in production that motivated this change... is it worth its own test? Or explicit mention in the release notes by way of a commit message?
    - A: not really. The audience who is aware of the constraint that's getting dropped is very small.

packages/smart-wallet/src/smartWallet.js
```diff
@@ -429,17 +423,6 @@ const behavior = {
             offerToInvitationMakers.get,
           ),
           purseForBrand: brandPurses.get,
-          lastOfferId: {
```
dckc, line 432: yay deleting code

dckc (approved):

PR #6557, opened 2022-11-10
4885 checkJs inter-protocol, smart-wallet

turadg:
refs: #4885

## Description

Adds checkJs to inter-protocol and smart-wallet.
Type improvements elsewhere to support that.
Removes `@ts-check` directives redundant with `checkJs` in enclosing packages.

### Testing Considerations

CI


packages/inter-protocol/src/proposals/price-feed-proposal.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
Chris-Hibbert: Does this deserve a comment as to why it's an exception?

Chris-Hibbert (approved):

packages/inter-protocol/src/proposals/price-feed-proposal.js
turadg: Good idea.

PR #6558, opened 2022-11-10
build: stop rebuilding for pack

turadg:
## Description

The `npm pack` [CI step](https://github.com/Agoric/agoric-sdk/actions/runs/3440489045/jobs/5739080421) for cosmic-proto started failing  after https://github.com/Agoric/agoric-sdk/pull/6540/.

The problem was it added a `prepack` that made a dependency on `protoc` which isn't available in that env. This removes the prepack, which wasn't necessary once the `dist/` files were added to SCM.

### Testing Considerations

Unlinked `protoc` locally and verified `npm pack` works.


kriskowal (approved):
Ah, aye. Another good reason for committing the artifacts. Not all engineers need to be set up for `protoc`.

Issue #3769, opened 2021-08-27
agd does not support joining with state sync


dckc:
## Describe the bug

While there is a practice of sharing [informal snapshots](https://github.com/Agoric/validator-profiles/wiki/Snapshots), the only in-protocol way to join an Agoric chain, currently, is to replay all transactions from genesis; this may take days or weeks. Contrast this with the norm in the Cosmos community:

> With block sync a node is downloading all of the data of an application from genesis and verifying it. With state sync your node will download data related to the head or near the head of the chain and verify the data. This leads to drastically shorter times for joining a network.
 -- [State Sync \| Tendermint Core](https://docs.tendermint.com/master/tendermint-core/state-sync.html)

Other blockchain systems have similar features. In Bitcoin and Ethereum, software releases include a hash of a known-good state; this way, new nodes can download a state that is not more than a few months old and start verifying from there.

## Design Notes

 - [ ] **consensus on swingset kernel DB state:** currently, the swingset DB state is not part of consensus; only the sequence of messages. Mnemonic: "you can _think_ whatever you want, as long as you _say_ the same thing that everyone else says". Fast sync most likely requires including (most of) the KB state in consensus Merkle tree proofs.
 - [ ] **consensus on xsnap snapshots**: currently, XS snapshots are not part of consensus; we don't require that all validators deterministically get exactly the same bytes in their snapshots. (In particular, #2776 is an observed case of non-determinism in snapshots). Fast sync most likely requires that we include snapshots in consensus.
- [ ] **cosmos-sdk hooks to publish swingset state**: baking our own system is undesirable; we just need some hooks to be able to leverage the Cosmos/Tendermint mechanisms for shipping states from select RPC nodes to the joining node (https://github.com/cosmos/cosmos-sdk/issues/7340#issuecomment-913092871)

cc @michaelfig @erights

rowgraus:
> Conservatively, I am putting it on phase 1. Feel free to postpone it as you see fit.

Tend to agree unless there are good arguments for postponing this

dckc:
@dtribble suggests that as long as catching up is 3x to 5x faster than the running chain, we can postpone this to a later milestone.

possible optimization:

 - [ ] replay from snapshot in parallel

(I think we only replay on-line vats, of which there is a bounded number, so that optimization doesn't seem worthwhile)

Tartuffo:
[...]

PR #6560, opened 2022-11-11
chore(deps): update `source-map@^0.7.3` to 0.7.4

sigv:
Refs: #6489 

## Description

The `source-map@0.7.4` update is important for enabling Node.js v18 compatibility. Older versions use a `typeof fetch === 'function'` check to enable browser-specific behavior. With this release however, the browser check is moved to whether `window` is defined and matches `this` value at check time. Node.js 18 has a `fetch` function, so older `source-map` versions incorrectly flag it as 'browser'. None of Node.js versions have a `this === window` situation going on.

This is a change on the path to Node.js 18 compatibility route, in anticipation of #6489 offering official support down the line.

Refs: mozilla/source-map@48f90d548f7e94ee6331e1862ca7ea2673ae9e57 

### Security Considerations

Assuming existing dependency ranges are assumed to be secure, this is not applicable.

N/A. No changes to user are expected, as the internal update is a semver patch.

### Testing Considerations

Existing regression testing cases apply here, as no new functionality is being introduced.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11931,9 +11931,9 @@ source-map@^0.6.0, source-map@^0.6.1, source-map@~0.6.1:
   integrity sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==
 
 source-map@^0.7.3:
-  version "0.7.3"
-  resolved "https://registry.yarnpkg.com/source-map/-/source-map-0.7.3.tgz#5302f8169031735226544092e64981f751750383"
-  integrity sha512-CkCj6giN3S+n9qrYiBTX5gystlENnRW5jZeNLHpe6aue+SrHcG5VYwujhW9s4dY31mEGsxBDrHR6oI69fTXsaQ==
+  version "0.7.4"
+  resolved "https://registry.yarnpkg.com/source-map/-/source-map-0.7.4.tgz#a9bbe705c9d8846f4e08ff6765acf0f1b0898656"
+  integrity sha512-l3BikUxvPOcn5E74dZiq5BGsTb5yEwhaTSzccU6t4sDOH8NWJCstKO5QT2CvtFoK6F0saL7p9xHAqHOlCPJygA==
 
 sourcemap-codec@^1.4.4:
   version "1.4.6"


sigv:
I am not sure about the existing release strategy and what gets merged where, however this can be cherry-picked into `release-pismo` for providing a new `pismo` build that should be Node.js 18 compatible.

I validated the `packages/cosmic-swingset` testsuite, as that is what I am most interested in. However I will also run the full testsuite, as it sounds likely to me that the testcases can be enabled for Node.js 18.x on the repository after this change.

sigv:
Deployed a fresh Ubuntu 22.04 (Jammy) image in Azure running on D16s_v5 (16 vCPUs, 64G RAM) for the test suite.

```bash
curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/nodesource.gpg >/dev/null
echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x $(lsb_release -s -c) main" | sudo tee /etc/apt/sources.list.d/nodesource.list >/dev/null

sudo apt update
sudo apt --yes dist-upgrade
sudo apt --yes install chromium-browser gcc g++ jq make nodejs perl

sudo corepack enable yarn

sudo wget -O /usr/local/go1.19.3.linux-amd64.tar.gz https://go.dev/dl/go1.19.3.linux-amd64.tar.gz
sudo tar -C /usr/local -xf /usr/local/go1.19.3.linux-amd64.tar.gz
eval $(echo 'export PATH=$PATH:/usr/local/go/bin' | sudo tee /etc/profile.d/golang.sh)

go version
# go version go1.19.3 linux/amd64
node --version
# v18.12.1
yarn --version
# 1.22.19

git clone -b maint/source-map-0.7.4 https://github.com/sigv/agoric-sdk.git
cd agoric-sdk

yarn install
yarn build
yarn test
# all tests reported as passed/skipped/todo (no failures)
```

Would say LGTM.

mhofman:
Thanks for the contribution, and for providing result of local tests. We definitely need to setup CI for node 18 to double check that this is the only required change.

> I am not sure about the existing release strategy and what gets merged where, however this can be cherry-picked into `release-pismo` for providing a new `pismo` build that should be Node.js 18 compatible.

As I mentioned in https://github.com/Agoric/agoric-sdk/issues/6489#issuecomment-1297453817, we do not plan on picking up such changes into the `release-pismo` branch. While it seems that the changes from [0.7.3 to 0.7.4 of source-map](https://github.com/mozilla/source-map/compare/0.7.3...v0.7.4) are limited to this bug fix and some wasm option we probably don't trigger, and we currently don't leverage source-map, it is technically a transitive dependency of endo's bundle-source, and I would be way too worried that the bundles created by the new version may differ from the bundles created by the previous version, which would cause consensus failures.

mhofman (approved):

mhofman:
@sigv apparently [Mergify cannot push to your branch](https://github.com/Agoric/agoric-sdk/pull/6560/checks?check_run_id=9463506758). Please rebase on or merge from master manually, or [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to your branch.

sigv:
Pushed a rebase via Web UI.

Doing so results in approval required for pipeline execution: _First-time contributors need a maintainer to approve running workflows._

sigv:
> I would be way too worried that the bundles created by the new version may differ from the bundles created by the previous version, which would cause consensus failures.

Understandable — I was thinking this could land together with the state-sync fix, if that is pushed out. However if you believe the risk is too high of this being a breaking change, that makes sense.

Thank you for the comment!

mhofman:
> Doing so results in approval required for pipeline execution

Approved and now running.

> this could land together with the state-sync fix, if that is pushed out

The current plan for state sync is that it'll only be part of the next major version. That said, with my limited understanding of cosmos node operations, I suggested a workaround [here](https://github.com/Agoric/agoric-sdk/issues/3769#issuecomment-1312874480) in the meantime.

sigv:
[Chain deployment test / deployment-test](https://github.com/Agoric/agoric-sdk/actions/runs/3459579649/jobs/5778952329) failed on `go mod download` with `zip: not a valid zip file` 🤔

mhofman:
It can't be related so I'll assume it's a transient network connection issue and re-ran the job.

sigv:
@mhofman, can you please reschedule the [mergify: Queue: Embarked in merge train](https://github.com/Agoric/agoric-sdk/pull/6560/checks?check_run_id=9473626821)? It seems it picked up the previous deployment test failure, but does not get triggered upon a successful re-run.

mhofman:
@mergifyio refresh

sigv:
Thank you 🎉

Issue #3087, opened 2021-05-12
explore sqlite as kernelDB


warner:
## What is the Problem Being Solved?

The swingset kernel currently encodes all of its state in a simple key-value format, with a "schema" (i.e. what we use each key for) informally defined in [a large comment](https://github.com/Agoric/agoric-sdk/blob/19954362e331a61385beb59ef7b726091d1fa9e3/packages/SwingSet/src/kernel/state/kernelKeeper.js#L27) in `kernelKeeper.js`. The kernel requires synchronous access to the DB.

We use LMDB to hold this "kernelDB" by implementing and using the `swing-store-lmdb` package. We selected LMDB because we found mature Node.js bindings which provided synchronous access. In contrast, the much richer "levelup/leveldown" ecosystem has bindings which are either async, or immature/unsupported.

@dckc and I are both fans of SQLite, and think it could be a good idea to move the kernelDB to it:

* SQLite provides actual configurable `INDEX`es, which could help a lot with the #2004 "virtual collections" that @erights is working on
* the DB is stored in a single file (vs the extra `lock.mdb` that LMDB uses)
* the size of that single file would grow only as new data is added (vs LMDB using a large-but-sparse file, currently 2GiB)
* I suspect SQLite would have less overhead, and thus would consume less space (I [estimate](https://github.com/Agoric/agoric-sdk/issues/3065#issuecomment-837484381) that 60% of the LMDB `data.mdb` is overhead)
* I can write tools to read SQLite files in other languages, whereas I've struggled to get e.g. the Python LMDB bindings to read our kernelDB files

However we haven't yet looked for synchronous SQLite Node.js bindings, so we don't know if it's really an option or not.

## Description of the Design

If we go this direction, we'd start with a `swing-store-sqlite` that has the same API as `swing-store-lmdb`, and only offers the same key-value features. Then, if it goes well, we'd abandon `swing-store-lmdb`, and start adding new features that depend upon SQLite, such as whatever indexing support #2004 needs.

## Security Considerations

Both SQLite and LMDB are pretty battle-tested libraries, at least their core C/C++ implementation. The maturity of the bindings will make a big difference, though.

dckc:
https://www.npmjs.com/package/better-sqlite3 provides synchronous bindings.

dckc:
[...]

PR #6561, opened 2022-11-13
Convert swingstore from LMDB to Sqlite, phase 1

FUDCo:
This PR is the first phase of the conversion from LMDB to Sqlite (in partial satisfaction of #3087).

This PR includes the replacement of the underlying database engine and the downstream changes that flow from it.  This is done in a series of four commits (individually reviewable for clarity), each of which realizes a different step of the transformation.

Step 1 - convert from LMDB to Sqlite, using Sqlite to realize a dumb key-value store that replicates the former LMDB semantics exactly, retaining all of the supporting implementation that assumes a dumb key-value store with LMDB's transaction model.

Step 2 - eliminate the `ephemeralSwingStore` implementation, used for testing, that realized the dumb key-value store in memory using a JavaScript Map object, replacing it with Sqlite's `:memory:` pseudo-file which does essentially the same thing (i.e., stores everything ephemerally in RAM) but does it using Sqlite directly.

Step 3 - use Sqlite savepoints for crank-level commits (rather than an in-memory change buffer), with full Sqlite transactions for block-level commits.  This eliminates a couple of layers of wrapper objects.  A necessary consequence of this is that the crank activity hash is now computed as part of the swingstore itself.

Step 4 - eliminate the rest of the storage wrappers, making the swingstore essentially self contained.

The system that results after each step is fully functional in a way that is compatible with the prior behavior of SwingSet.
If this PR is merged with the master branch, the result will be entirely usable.

I am pushing this PR now in order to get review started.  However, there is more to be done.  In the broadest possible strokes, this remaining work falls into two categories:

- Short term changes to
  - (a) rationalize the transaction model to account more gracefully for crank activity that is pre- or post-delivery, since we now have the ability to do more selective rollback. in particular, we believe there is a potential for improper commitment or non-commitment of pre-crank data mutations.  Note that this risk is not new -- it's present with the LMDB-based store as well -- but the switch to Sqlite presents us with the opportunity to ensure that the cases we are concerned about are actually handled correctly. 
  - (b) remove the use of iterators that require either the retention of a database cursor during iteration (which we can't tolerate) or fetching the entirety of a `SELECT` result into RAM before using it (which the implementation in this PR does but which exposes us to the risk of an unbounded memory suck from adversarial vat code), replacing them with something more like the vatstore's `getAfter` machinery.

- Long term changes that exploit the capabilities of the SQL query mechanism in various ways.  These include but are not limited to bulk operations for data deletion, improving GC of on-disk data, and removing the vatstore from the consensus SwingSet state.  Most of these will require SwingStore API changes or additions, which argues for taking some care planning them before proceeding, hence their absence from this PR.  In addition, we expect further exploration/experimentation to identify further optimization opportunities.


mhofman:
> The system that results after each step is fully functional in a way that is compatible with the prior behavior of SwingSet.

I'm wondering, does that mean that the behavior of node with the PR applied, and one without would be 100% consistent (long term deterministic)? Even without this, how confident would we be in applying this change to `release-pismo`? Which begs the question, do we believe there may be a way to upgrade from the current DBs to the new sqlite one? Aka export the content of LMDB in a new SQLite DB and restart from there (either as a manual script or as a built-in upgrade sequence).

FUDCo:
> I'm wondering, does that mean that the behavior of node with the PR applied, and one without would be 100% consistent (long term deterministic)?

There are three answers to this, which are respectively, yes, no, and it doesn't matter.

Yes, I've verified that the sequence of database operations are the same and that the action hashes match.

No, there is an exception to the action hashes matching, which is when there is an action in the crank containing a bundle reference and the bundle hash has changed because the code has changed.  And of course changing the database engine is a code change.  On the other hand, I observed this with tests that themselves had changes, and it's possible that since the database code changes are all in the kernel we'll be fine if the kernel bundle does not appear in any action hash (which, long term, we need it to not do if we want kernel upgrade to be possible).  @warner?

And it doesn't matter (I think) because IIUC we'd be switching database engines as part of the bulldozer upgrade.

mhofman:
> which is when there is an action in the crank containing a bundle reference and the bundle hash has changed because the code has changed

Right! This is an interesting case since this change has an impact on the kernel, mostly due to the swingstore api refactoring from what I gather. I thought however that we removed the kernel bundle from being saved in the DB? Or did this somehow impact the liveslot/supervisor bundle?

> And it doesn't matter (I think) because IIUC we'd be switching database engines as part of the bulldozer upgrade.

I was wondering in the context of potentially having to deal with state-sync before a bulldozer upgrade. It sounds like with some effort, we might be able to swap the DB implementation for release-pismo. In that case we'd need to do other surgery, e.g. extract local XS snapshot hashes and move them in a different section of the DB (after verifying they're deterministic of course), so it'd definitely not be for the faint of heart.

packages/swing-store/src/sqlStreamStore.js
```diff
@@ -54,25 +46,6 @@ export function sqlStreamStore(dbDir, io) {
     )
   `);
 
-  // We use explicit transactions to 1: not commit writes until the
```
warner, line 57: please move this "We use explicit transactions to ..." comment into the definition of `ensureTxn` in it's new place.. it's still relevant

warner:
Short take: this looks great. I'll start reviewing properly now.

We no longer store the kernel bundle anywhere. We still need to build a kernel bundle (because the kernel runs in its own Compartment, and `import-bundle` is the only tool we have for loading more than a single `eval`-ed string into a Compartment), but we now do that on every boot, rather than bundling once during `initializeSwingset` and storing the result for all subsequent reboots. This makes a kernel upgrade easier, because you just run a new version of the host application (which imports a new version of `@agoric/swingset-vat`), but of course exports the "only run compatible versions" requirement onto the author of the host app.

I'll think about whether this could cause visible behavioral changes as I review. My intention was that we make this switch without worrying about such changes, especially because of my goal to replace `vatstore.getKeyAfter` with a simpler API, more like `getNext`, and that would entail a change to liveslots.

@FUDCo and walked through what the crankBuffer commit/abort replacement will be (which can happen after this lands). We concluded that we will:
* decouple crankHash from what used to be the crankBuffer, so there's a `kernelStorage.getCrankHash()` that returns a hash of everything since the last call, and provide some way to fold that into the activityHash (or maybe that's the host's responsibility, not sure yet)
* the "crank cycle" is the thing that defines crank 1, crank 2, etc
* we pick the point on this cycle just before we pop something off the run-queue as the start, which is also the state we're in just before `controller.run()` is called, and also the state we're in just after `controller.run()` finishes
* we define two SQLite SAVEPOINTs, the first is at this "idle" point, the second is just after the delivery/etc has been popped off the run-queue (so we need two swing-store APIs, and the kernel will call each of them, once per cycle)
* then there are three possible outcomes:
  * 1: the delivery finishes normally, and the kernel wants to commit all changes. We call some swing-store API that says "delete all SAVEPOINTs, I won't use them"
  * 2: the delivery goes weird, and the kernel decides to discard the changes, and the delivery/run-queue item was consumed (so the kernel does not want to process it ever again), e.g. a `createVat` or `upgradeVat` failing. The kernel calls a swing-store API which rewinds to the second SAVEPOINT (the one taken after the pop). Then the kernel does cleanup work like pushing vat-admin messages about the failure. The swing-store API will also delete the first SAVEPOINT, since we won't be using it, and we don't want to let it clutter up RAM.
  * 3: the delivery goes weird, the kernel decides to discard the changes, but the kernel wants the delivery to be re-attempted later (e.g. a `deliver` killed the vat, and a re-delivery is the simplest way to provoke the right `VAT_TERMINATED` error). For this case, the kernel calls a third swing-store API which rewinds to the first SAVEPOINT (taken before the pop), and deletes the second SAVEPOINT (again to avoid clutter). Then the kernel does cleanup work like vat-admin messages.
* Note that all kvstore changes, committed or abandoned, get rolled into the crankhash. Previously I think we only included committed changes, but we decided/concluded that they're all part of the deterministic behavior. The commit-vs-abort is part of consensus, so the reasons for choosing one vs the other must be within-consensus too.

And @FUDCo points out:
* Note that in case 3, the rollback to the first SAVEPOINT clears out the second SAVEPOINT as it goes by with no additional action necessary.  Also, in case 2, clearing the first SAVEPOINT is not just a matter of memory savings, it’s a backstop against bugs in later code hitting it in a later rollback.

packages/swing-store/package.json
```diff
@@ -21,13 +21,13 @@
...
     "@agoric/internal": "^0.2.1",
     "better-sqlite3": "^7.5.0",
-    "lmdb": "^2.4.5",
     "tmp": "^0.2.1"
   },
   "devDependencies": {
     "@endo/init": "^0.5.52",
     "@types/better-sqlite3": "^7.5.0",
     "ava": "^5.0.1",
+    "@agoric/swingset-vat": "^0.30.2",
```
warner: Huh, I guess this is because `test-hasher.js` and `hasher.js` needs `assert`? It'd be nice to avoid this circular ref (ok, strictly speaking cycles that go through `devDependencies` aren't a problem, but it feels weird).

If `test-hasher.js` started with a prefix copied from `test/test-state.js`:
```js
import '@endo/init/debug.js';
import test from 'ava';
```

then we could remove this dependency.

It suggests we want a better story for "AVA, but under SES, and we don't need VatData or the fakeVirtualStuff", which I think is what most of our unit tests need.

packages/swing-store/src/swingStore.js
```diff
@@ -53,40 +50,56 @@ export function makeSnapStoreIO() {
...
  *   diskUsage?: () => number, // optional stats method
+ * }} SwingStoreHostStorage
+ *
+ * @typedef {{
+ *  kernelStorage: SwingStoreKernelStorage,
+ *  hostStorage: SwingStoreHostStorage,
  * }} SwingStore
  *
- * @typedef {SwingStore & { snapStore: ReturnType<typeof makeSnapStore> }} SwingAndSnapStore
+ * //at//typedef {KernelStore & { snapStore: SnapStore }} KernelAndSnapStore
```
warner: what's this `//at//` mean?

packages/swing-store/src/hasher.js
```diff
@@ -23,10 +23,13 @@ function createSHA256(initial = undefined) {
     done = true;
     return hash.digest('hex');
   }
+  function sample() {
```
warner, line 26: This new function introduces an opportunity for length-extension vulnerabilities, and doesn't appear to be used anywhere. Can we remove it?

packages/swing-store/test/test-hasher.js
```diff
@@ -1,7 +1,7 @@
-import { test } from '../tools/prepare-test-env-ava.js';
+import { test } from '@agoric/swingset-vat/tools/prepare-test-env-ava.js';
```
warner:
```js
import '@endo/init/debug.js';
import test from 'ava';
```

would let us remove the slightly-weird swingset dependency in package.json

packages/swing-store/src/sqlStreamStore.js
```diff
@@ -90,14 +90,8 @@ export function sqlStreamStore(dbDir, io) {
...
-      ensureTransaction();
-      const query = db.prepare(`
-        SELECT item
-        FROM streamItem
-        WHERE streamName = ? AND position >= ? AND position < ?
-        ORDER BY position
-      `);
-      for (const { item } of query.iterate(
+      ensureTxn();
+      for (const { item } of sqlReadStreamQuery.iterate(
```
warner, line 94: I'm sure this was a problem with previous version too, but am I correct in thinking that this iterator wouldn't be happy with other DB txns happening in the middle of iteration?

Our current use case for reading the stream is vat transcript replay, and that doesn't do a lot of other DB activity (syscalls are handled from the transcript, not executed for real), but Mathieu's #6664 XS/GC bug workaround involves allowing certain `vatstoreGet` calls to be executed during replay, and those would cause DB reads. We aren't landing that on trunk, but it suggests that we shouldn't depend too deeply on excluding other DB activity during sequential reads of a list like this.

When we do the whole `getAfter` -> `getNext` transformation, we can do something similar for the streamstore read API, either something that asks for specific deliveryNums (one at a time, using a `LIMIT 1` clause), or maybe read them in moderate (10-ish) -sized batches (via a kernel-side helper function, and an API that includes a `LIMIT` argument).

packages/swing-store/src/sqlStreamStore.js
```diff
@@ -148,23 +142,11 @@ export function sqlStreamStore(dbDir, io) {
...
-    db.close();
-  };
-
   return harden({
     writeStreamItem,
     readStream,
     closeStream,
-    commit,
-    close,
+    dumpStreams,
```
warner, line 149: Please add a short test for `dumpStreams` to `test-state.js`, since it's now part of the external API (even though it's only being used by `getAllState`).

packages/swing-store/src/swingStore.js
```diff
@@ -53,40 +50,56 @@ export function makeSnapStoreIO() {
...
  *   STREAM_START: StreamPosition,
  * }} StreamStore
  *
  * @typedef {{
- *   kvStore: KVStore, // a key-value StorageAPI object to load and store data
+ *   kvStore: KVStore, // a key-value StorageAPI object to load and store data on behalf of the kernel
  *   streamStore: StreamStore, // a stream-oriented API object to append and read streams of data
+ *   snapStore?: SnapStore,
+ *   startCrank: () => void,
+ *   establishCrankSavepoint: (savepoint: string) => void,
```
warner, line 62: I kind of want these to be named `establishSavepoint` and `rollback | rollbackToSavepoint`, because of the non-precise correlation between the crank boundaries (used for hashing) and the rollback points, and because rollback is specifically used for *deliveries* (which are a subset of all cranks). But I don't want to bikeshed on the name, at least not until I've finished reading everything else and see how the name feels in context.

packages/swing-store/src/swingStore.js
```diff
@@ -161,70 +187,92 @@ function makeSwingStore(dirPath, forceReset, options) {
...
-    // encoded as 2 bytes. lmbd-js also bumps the maxKeySize to 1791 bytes,
-    // making this a non-issue for now.
-    encoding: 'string',
-  });
+  /** @type {*} */
+  let db = sqlite3(
+    filePath,
+    // { verbose: console.log },
+  );
+  db.exec(`
```
warner, line 196: let's start this off right, and use WAL mode from the beginning:

```
db.exec(`PRAGMA journal_mode=WAL;`);
```

and then add add a checkpoint command to the `commit()` function (see below).

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +373,179 @@ function makeSwingStore(dirPath, forceReset, options) {
...
+  function endCrank() {
+    assert(inCrank);
+    if (savepoints.length > 0) {
+      sqlReleaseSavepoints.run();
+      savepoints.length = 0;
+    }
+    inCrank = false;
+  }
+
+  const sqlCommit = db.prepare('COMMIT');
```
warner, line 511: Once we're in WAL mode (https://sqlite.org/wal.html), the SQLite defaults only flush the write-ahead-log to disk (i.e. perform a "checkpoint") every 1000 blocks written, which is not safe against power failures. To fix that, we want to force a checkpoint after each commit.

```js
const sqlCheckpoint = db.prepare('PRAGMA wal_checkpoint(FULL)');
```

performance-wise, this should be fine, since we're only calling `commit()` at block boundaries, so like once every 5 seconds. This will be the only time that `fsync()` gets called.

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +373,179 @@ function makeSwingStore(dirPath, forceReset, options) {
...
     assert(db);
-    // Commit the stream-store first, before kvStore. We keep the
-    // streams' startPos/endPos in the kvStore, so if we crash after
-    // commitStreamStore() but before the kvstore doCommit(), we'll
-    // wake up with the old startPos/endPos and it's (mostly) as if
-    // the stream entry was never added.
-    commitStreamStore();
-    await doCommit(false);
+    if (db.inTransaction) {
+      sqlCommit.run();
```
warner, line 520: and then
```js
sqlCheckpoint.run();
```

packages/swing-store/src/swingStore.js
warner: After we incorporate the snapStore into the DB, we should add a small (one-entry) `version` table, and store a `1` in it, to make future upgrades/migrations more rational. But let's not do that in this PR, since we aren't finished making all the tables that V1 will hold.

packages/swing-store/src/swingStore.js
```diff
@@ -161,70 +187,92 @@ function makeSwingStore(dirPath, forceReset, options) {
...
+  );
+  db.exec(`
+    CREATE TABLE IF NOT EXISTS kvStore (
+      key TEXT,
+      value TEXT,
+      PRIMARY KEY (key)
+    )
+  `);
+
+  const sqlBeginTransaction = db.prepare('BEGIN IMMEDIATE TRANSACTION');
```
warner, line 204: What do you think about consolidating all of the `db.prepare` statements into a single block, here, just after we create the DB and the tables?

packages/swing-store/src/swingStore.js
```diff
@@ -161,70 +187,92 @@ function makeSwingStore(dirPath, forceReset, options) {
...
     assert.typeof(key, 'string');
-    let result = ensureTxn().get(key);
-    if (result == null) {
-      result = undefined;
-    }
-    return result;
+    return sqlKVGet.get(key);
   }
 
+  const sqlKVGetKeys = db.prepare(`
```
warner, line 278: style nit: most of these SQL statements could fit on a single line, and I think it'd be more readable overall if they weren't so vertically spread out

the ones that can't fit on a single line, should continue to be laid out this way, of course

packages/swing-store/src/swingStore.js
```diff
@@ -239,18 +287,14 @@ function makeSwingStore(dirPath, forceReset, options) {
...
-      rangeOptions.end = end;
+      keys = sqlKVGetKeys.all(start, end);
+    } else {
+      keys = sqlKVGetKeysNoEnd.all(start);
     }
 
-    const keys = ensureTxn().getKeys(rangeOptions);
-
-    yield* keys;
+    yield* keys.values();
```
warner, line 316: `stmt.all()` allocates and returns a giant array, and then the `yield` wraps that in an iterator, which is kind of the worst of both worlds: high memory cost, and an API that doesn't finish right away. This is high on the list of cleanups to do after this lands.

packages/swing-store/src/swingStore.js
```diff
@@ -267,6 +311,21 @@ function makeSwingStore(dirPath, forceReset, options) {
     return get(key) !== undefined;
   }
 
+  const sqlKVSet = db.prepare(`
+    INSERT INTO kvStore (key, value)
+    VALUES (?, ?)
+    ON CONFLICT DO UPDATE SET value = excluded.value
+  `);
+
+  // XXX Debug helper, remove before push
```
warner: don't forget to remove this

packages/swing-store/src/swingStore.js
```diff
@@ -161,70 +187,92 @@ function makeSwingStore(dirPath, forceReset, options) {
...
     assert(db);
-    if (!txnDone && !txnFinish) {
-      trace('begin-tx');
-      txnDone = db.transactionSync(
-        () =>
-          new Promise(resolve => {
-            txnFinish = resolve;
-          }),
-      );
+    if (!db.inTransaction) {
```
warner, line 227: let's add a comment here, for future maintainers

"It is critical to call `ensureTxn` as the first step of any API call that might modify the database (any `INSERT` or `DELETE`, etc), to prevent SQLite from creating an automatic transaction, which will commit as soon as the SQL statement finishes. This would cause partial writes to be committed to the DB, and if the application crashes before the real `hostStorage.commit()` happens, it would wake up with inconsistent state. The *only* commit point must be the `hostStorage.commit()`."

warner: er, merge this text with the text I asked to be copied over from `sqlStreamStore.js`

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +373,179 @@ function makeSwingStore(dirPath, forceReset, options) {
...
 
-  /** @param {boolean} [abort] */
-  function doCommit(abort) {
-    if (!txnFinish) {
-      return undefined;
+  function establishCrankSavepoint(savepoint) {
+    assert(inCrank);
+    const savepointOrdinal = savepoints.length;
+    savepoints.push(savepoint);
+    const sql = db.prepare(`SAVEPOINT t${savepointOrdinal}`);
```
warner, line 454: nit: collapse this to just `db.exec(..)`

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +373,179 @@ function makeSwingStore(dirPath, forceReset, options) {
...
+    sql.run();
+  }
+
+  function rollbackCrank(savepoint) {
+    assert(inCrank);
+    for (const savepointOrdinal of savepoints.keys()) {
```
warner, line 460: huh, it didn't occur to me that `.keys()` is safe to use on an Array, neat

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +373,179 @@ function makeSwingStore(dirPath, forceReset, options) {
...
-        trace(`${abort ? 'abort' : 'commit'}-tx`);
-      })
-      .finally(() => {
-        txnDone = null;
-        txnFinish = null;
-      });
+    assert.fail(`no such savepoint as "${savepoint}"`);
+  }
+
+  function emitCrankHashes() {
```
warner, line 471: nit: maybe `computeCrankHashes`, "emit" makes me think it's doing IO somehow. "Return" would be wrong, because there's definitely a side-effect going on. "close crank hash" would be close, but of course there's no corresponding open (it's automatic), plus there's also the activity hash.

packages/swing-store/src/swingStore.js
```diff
@@ -383,21 +555,20 @@ function makeSwingStore(dirPath, forceReset, options) {
...
+ *   be created.
  * @param {object?} options  Optional configuration options
  *
  * @returns {SwingStore}
  */
-export function initSwingStore(dirPath, options = {}) {
-  if (!dirPath) {
-    return initEphemeralSwingStore();
-  } else {
+export function initSwingStore(dirPath = null, options = {}) {
```
warner, line 584: not a change for now, but I'd like to revisit this creation API.. I think I want an `init` which strictly only creates a new DB (and throws if one exists already), and an `open` which strictly only opens an existing DB (and throws if one does *not* exist), and remove both the "init is willing to delete an existing one", and "open is willing to create a missing one" behaviors

I haven't through through how this should be used by `initializeSwingset`, though, especially for the `:memory:` case.

packages/cosmic-swingset/src/launch-chain.js
```diff
@@ -214,22 +214,17 @@ export async function launch({
...
-  const hostStorage = {
-    kvStore,
-    streamStore,
-    snapStore,
-  };
+  const { kernelStorage, hostStorage } = openSwingStore(kernelStateDBDir, {
+    traceFile: swingStoreTraceFile,
+    keepSnapshots,
+  });
+  const { kvStore, commit } = hostStorage;
```
warner, line 228: The PR is fine here, but the comment just below this point reminds me that we should look at `makeQueue()` (and how it is used by cosmic-swingset), and make sure that whatever writes it does into the `host.inboundQueue.NN` keys are committed at the right time, and that we understand what "crank" they appear in (q.v. our discussion about what "crank" a `controller.installBundle()` appears in).

packages/SwingSet/misc-tools/rekernelize.js
```diff
@@ -56,17 +55,17 @@ async function main() {
     fail(`can't find a database at ${dbDir}`);
   }
 
-  const swingStore = openSwingStore(kernelStateDBDir);
-  const kvStore = swingStore.kvStore;
+  const { kernelStorage, hostStorage } = openSwingStore(kernelStateDBDir);
+  const kvStore = kernelStorage.kvStore;
```
warner: PR is fine, but I realized that we should just delete this tool, since we no longer keep a kernel bundle in the DB.

packages/SwingSet/src/types-external.js
```diff
@@ -295,14 +295,16 @@ export {};
...
  * @typedef { import('@agoric/swing-store').SwingStore } SwingStore
+ * @typedef { import('@agoric/swing-store').SwingStoreKernelStorage } SwingStoreKernelStorage
+ * @typedef { import('@agoric/swing-store').SwingStoreHostStorage } SwingStoreHostStorage
  *
  * @typedef {{
  *   kvStore: KVStore,
  *   streamStore: StreamStore,
  *   snapStore?: SnapStore,
+ *   commitCrank: () => { crankhash: string, activityhash: string },
+ *   abortCrank: (wholeCrank: boolean) => void,
```
warner: I don't see `commitCrank` or `abortCrank` anywhere else.. those are leftovers from before the savepoint-based `rollbackCrank` scheme, right?

I don't see `HostStore` used anywhere else either (except for more typedef imports).. it looks like you deleted the last actual uses as part of this PR. Could you try deleting this whole typedef, plus the lines that reference it in `kernelKeeper.js` and `types-ambient.js`, and see if anything breaks?

packages/SwingSet/src/controller/controller.js
```diff
@@ -477,12 +479,13 @@ export async function makeSwingsetController(
       if (!options.upgradeMessage) {
         options.upgradeMessage = `vat ${vatName} upgraded`;
       }
-      return controller.queueToVatRoot(
+      const result = controller.queueToVatRoot(
```
warner, line 483: maybe add `// queueToVatRoot does its own emitCrankHashes()`

packages/SwingSet/src/controller/controller.js
```diff
@@ -442,7 +441,9 @@ export async function makeSwingsetController(
     },
 
     kpResolution(kpid) {
-      return kernel.kpResolution(kpid);
+      const result = kernel.kpResolution(kpid);
+      kernelStorage.emitCrankHashes();
```
warner, line 446: let's add a comment, `// that changes refcounts, so hash the DB writes`

packages/SwingSet/src/controller/controller.js
```diff
@@ -434,6 +432,7 @@ export async function makeSwingsetController(
       const vatID = kernel.vatNameToID(vatName);
       const kref = kernel.getRootObject(vatID);
       kernel.pinObject(kref);
+      kernelStorage.emitCrankHashes();
```
warner, line 435: This line is fine, but doesn't `changeKernelOptions()` need an `emitCrankHashes` too? (github won't let me add comments to lines that didn't change)

packages/SwingSet/src/controller/initializeKernel.js
```diff
@@ -152,7 +151,7 @@ export function initializeKernel(config, hostStorage, verbose = false) {
     }
   }
   kernelKeeper.setInitialized();
-  kernelKeeper.commitCrank(); // commit initialized kernel state as crank #0
+  kernelKeeper.emitCrankHashes();
```
warner: maybe add `// initialized kernel state is hashed as if it were crank #0`

packages/SwingSet/src/controller/initializeSwingset.js
```diff
@@ -340,6 +345,7 @@ export async function initializeSwingset(
 
   kvStore.set('lockdownBundle', JSON.stringify(kernelBundles.lockdown));
   kvStore.set('supervisorBundle', JSON.stringify(kernelBundles.supervisor));
+  kernelStorage.emitCrankHashes();
```
warner: Huh, this looks like we're doing two "cranks" during initialization, which seems weird. Ideally we'd only do one.

I'm guessing that these bundle writes are happening in a weird place, and it didn't feel right to defer a hashing step until `initializeKernel` does its own?

I'm guessing we should add a "these bundle writes will get hashed as part of `initializeKernel`, below" comment, and remove this `emitCrankHashes`, but I'll defer to your judgement.

packages/SwingSet/src/kernel/state/storageHelper.js
```diff
@@ -0,0 +1,47 @@
+// @ts-check
+
+import { assert } from '@agoric/assert';
+
+// We wrap a provided object implementing StorageAPI methods { has, getKeys,
```
warner: Both of these comments should be deleted: we're no longer wrapping (instead this file provides non-wrapping helper functilns), and the bit about "host realm objects" is *ancient*.

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -172,41 +168,17 @@ const FIRST_CRANK_NUMBER = 0n;
...
+  insistStorageAPI(kvStore);
 
-  const {
-    abortCrank: storeAbortCrank,
-    commitCrank: storeCommitCrank,
-    enhancedCrankBuffer: kvStore,
-  } = wrapStorage(rawKVStore, createSHA256, getKeyType);
-  insistEnhancedStorageAPI(kvStore);
-  const { streamStore, snapStore } = hostStorage;
+  const { streamStore, snapStore } = kernelStorage;
```
warner: I'd suggest `const { kvStore, streamStore, snapStore } = kernelStorage;`, nice and tidy

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -172,41 +168,17 @@ const FIRST_CRANK_NUMBER = 0n;
...
 
   function getActivityhash() {
-    return rawKVStore.get('activityhash');
+    return kvStore.get('activityhash');
```
warner: this is fine for now, but the magic key string `activityhash` appears in two different packages, so let's make a note to consider exposing it on swingstore, maybe as `kernelStorage.getActivityHash`

it probably doesn't need to be a `kernelKeeper` method, or even a `kernel.getActivityhash` method, since the only caller is `controller.activityHash` and they can access `kernelStorage` themselves

packages/SwingSet/src/kernel/state/kernelKeeper.js
```diff
@@ -242,15 +214,26 @@ export default function makeKernelKeeper(
     });
   }
 
-  function commitCrank() {
-    saveStats();
-    return storeCommitCrank();
+  function startCrank() {
```
warner, line 211: thinking about names for a minute:

| API name | touches hashes | touches savepoints | touches stats |
| ----- | --- | --- | --- |
| `startCrank` | yes | no | no |
| `establishCrankSavepoint` | no | yes | no |
| `rollbackCrank` | no | yes | yes |
| `emitCrankHashes` | yes | no | yes |
| `endCrank` | no | yes | no |

yeah, that mismatch between `startCrank` and `endCrank` is what's bugging me. Let's chat about this some more.

packages/SwingSet/test/change-parameters/test-change-parameters.js
```diff
@@ -3,8 +3,8 @@ import { test } from '../../tools/prepare-test-env-ava.js';
 
 // eslint-disable-next-line import/order
 import { assert } from '@agoric/assert';
-import { getAllState } from '@agoric/swing-store';
-import { provideHostStorage } from '../../src/controller/hostStorage.js';
+// eslint-disable-next-line import/order
```
warner: I don't think we need this extra `import/order` disabler.. the important ones were on the `prepare-test-env-ava.js` and on whatever came after it (at least that's how I convinced `prettier` to stop trying to move the `prepare-test-env-ava` out of the important first-thing-imported spot).

packages/SwingSet/test/test-gc-actions.js
```diff
@@ -169,7 +170,7 @@ test('gc actions', t => {
   rc = { ko1: [0, 0], ko2: [0, 0] };
   clistState.v1.ko2 = { exists: false, isReachable: false };
   msg = process();
-  t.deepEqual(msg, undefined);
+  t.deepEqual(msg, { type: 'negated-gc-action', vatID: undefined });
```
warner, line 173: let's add a new check, at the end of this whole sequence, to demonstrate that `processGCActionSet()` is really returning `undefined` when there's no work left to do:

```js

  msg = process();
  t.deepEqual(msg, undefined);
  t.deepEqual(newActions, []);
});
```

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -55,11 +54,11 @@ const testUpgrade = async (
     },
   };
 
-  const hostStorage = provideHostStorage();
-  const { kvStore } = hostStorage;
+  const kernelStorage = initSwingStore().kernelStorage;
```
warner: Not sure if it's worth it, but prettier might suggest `const { kernelStorage } = initSwingStore();`, and it'd be shorter too. I didn't initially like that style, but it's growing on me.

packages/SwingSet/test/virtualObjects/test-representatives.js
```diff
@@ -190,9 +190,9 @@ test.serial('exercise cache', async t => {
   await make('thing5', false, T5); // evict t2, make t5 - [t5 t4 t3 t1]
   ck('get', esKey(5), undefined);
   ck('set', esKey(5), 'r');
+  ck('set', dataKey(5), thingVal('thing5'));
```
warner, line 193: do we know why this changed? XXX-@warner

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1688,40 +1680,49 @@ export default function buildKernel(
...
-      const value = options[option];
-      switch (option) {
-        case 'defaultReapInterval': {
-          kernelKeeper.setDefaultReapInterval(value);
-          break;
-        }
-        case 'snapshotInterval': {
-          vatWarehouse.setSnapshotInterval(value);
-          break;
+    kernelKeeper.startCrank();
```
warner: huh, I don't know how `step` and `run` got separated by `changeKernelOptions`, but this is another cleanup to do later: rearrange this block so `changeKernelOptions` comes after `run` or before `step`, not in between

packages/SwingSet/src/kernel/state/kernelKeeper.js
warner: hm, and if it weren't for `stats`, we could move most of these off of `kernelKeeper` and let `kernel.js` access them directly from `kernelStorage.*` ... having them on `kernelKeeper` is no longer providing a lot of value

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1740,41 +1741,46 @@ export default function buildKernel(
...
+            break;
+          default:
+            assert.fail(`unknown policyInput type in ${policyInput}`);
+        }
+        if (!policyOutput) {
+          // console.log(`ending c.run() by policy, count=${count}`);
+          return count;
+        }
+      } finally {
+        kernelKeeper.endCrank();
```
warner, line 1788: Oh, huh, I think the previous version of this had a bug: if the policy said to stop, it did `return count`, but bypassed the `commitCrank()`. In your version, the `finally` ensures that it wil always be called.

If `commitCrank` was getting skipped on the last crank of a run, I wonder if that crank's DB writes got left over until the next run. That could be a problem, especially if the node were interrupted before the next block had a chance to fix things. I'll need to investigate that one.

warner: Oh, it's ok, we were doing `commitCrank()` inside the delivery handler. The `commitCrank()` at the end was there to catch the case where the queue was empty, so we `break` out of the loop before doing a delivery, and need to commit the now-empty run-queue. The only way for `policyOutput` to break out of the loop is if a delivery was done (to create the policyInput), and in those cases we'd always `commitCrank()` before talking to the policy.

packages/SwingSet/src/kernel/kernel.js
```diff
@@ -1210,22 +1214,9 @@ export default function buildKernel(
...
-        // delivery itself from the head of the run queue, which will
-        // result in it being delivered again on the next crank. If we
-        // don't want that, then we need to remove it again.
-
-        // eslint-disable-next-line no-use-before-define
-        getNextDeliveryMessage();
-      }
-      // other deliveries should be re-attempted on the next crank, so they
-      // get the right error: we leave those on the queue
+      kernelKeeper.rollbackCrank(
```
warner, line 1224: let's preserve a few comments:

```js
if (crankResults.abort) {
  // errors unwind any changes the vat made, by rolling back to the "deliver" savepoint
  // In addition, the crankResults will either ask for the message to be consumed
  // (without redelivery), or they'll ask for it to be attempted again (so it can go "splat"
  // against a terminated vat, and give the sender the right error). In the latter case,
  // we roll back to the "start" savepoint, established by `run()` or `step()` before
  // the delivery was pulled off the run-queue, effectively pushing it back on.
```

packages/SwingSet/test/vat-admin/test-create-vat.js
```diff
@@ -290,25 +290,27 @@ test('createVat holds refcount', async t => {
...
   // incrementing the refcount.
-  expectedRefcount += 1; // v2-bootstrap
-  expectedCLists += 1; // v2-bootstrap
+  expectedRefcount += 1; // v1-bootstrap holds resolution of 'createHeld'
+  expectedCLists += 1; // v1-bootstrap
 
   // calling getHeld doesn't immediately increment the refcount
   const kpid1 = c.queueToVatRoot('bootstrap', 'getHeld', []);
   await c.run();
+  expectedRefcount += 1; // from 'getHeld'
```
warner: Why did this change? I need to investigate more. XXX-@warner

packages/SwingSet/test/test-state.js
```diff
@@ -71,187 +70,67 @@ async function testStorage(t, s, getState, commit) {
...
-    ['foo.5', 'f5'],
-  ]);
+
+  deletePrefixedKeys(kv, 'foo.');
+  t.falsy(kv.has('foo.0'));
+  t.falsy(kv.has('foo.1'));
+  t.falsy(kv.has('foo.2'));
+  t.falsy(kv.has('foo.3'));
+  t.falsy(kv.has('foo.4'));
+  t.truthy(kv.has('foo.5'));
```
warner, line 116: let's add a comment just before the `foo.5` line, because this is testing something weird and I had to read it twice before I remembered

```js
// `getPrefixedValues` and `deletePrefixedKeys` both happen to work by counting from zero,
// so if there are any gaps (e.g. 'foo.4'), they'll stop counting
```

I think the `start=0` argument, which you correctly deleted, was a leftover from some ancient time. It's weird for a `deletePrefixedKeys` to react to a gap like this, and means we couldn't use it to delete, e.g., a vat's input queue (which, when we implement it, will be a bunch of keys that don't start at zero). But we'll implement those queues with something better anyways, so I'm happy to see that feature go.

packages/SwingSet/test/test-state.js
```diff
@@ -71,187 +70,67 @@ async function testStorage(t, s, getState, commit) {
...
-    crankBuffer,
-    () => getAllState(store).kvStuff,
-    commitCrank,
+    kernelStorage.kvStore,
+    () => getAllState(kernelStorage).kvStuff,
+    null,
   );
 });
 
-test('crankBuffer handles key iteration properly even with intra-crank data changes', t => {
```
warner, line 92: I see why this goes away (the crankBuffer is not really buffering things anymore), but I need to think about whether the `getKeys` swingstore is now providing will do the right thing w.r.t. commit points. 95% confident that it does, but XXX-@warner remind myself to double check.

warner (changes requested):
Phew, ok sorry that took a while. Some small changes to make, some items for me to investigate further, and a few things to discuss.

packages/swing-store/package.json
FUDCo: Yes, but it's Ava and not `assert` that's the culprit.  This could definitely be handled better, as it's sort of a path-dependent accident of history that some of our test infra lives in the `SwingSet` package.  However, I don't think this is the PR to address that, as there seem to be 10-20 other `agoric-sdk` packages that have the same problem.  An open question is where this test boilerplate should live instead, for which I don't have an immediate obvious answer.  Somebody should file an issue...

packages/swing-store/src/swingStore.js
FUDCo: It's debris from debugging TypeScript that I missed removing.

packages/swing-store/src/hasher.js
FUDCo: I was in the midst of writing a reply to the effect of "no, we use it quite a lot" and then went looking for examples and discovered that you're right, we don't use it at all.  I think what I was remembering was debug code, where it proved essential for diagnosing hash aggregation problems when I was working on the commit/abort changes.  Since it *was* useful in that case, I'm inclined to leave it in, but I'm open to being persuaded otherwise.  Can you explain the vulnerability potential that you are concerned about?

packages/swing-store/test/test-hasher.js
FUDCo: Hmm.  Contra my earlier comment above, that seems to have worked.  Well OK then.

packages/swing-store/src/sqlStreamStore.js
FUDCo: > I'm sure this was a problem with previous version too, but am I correct in thinking that this iterator wouldn't be happy with other DB txns happening in the middle of iteration?

Yes, the code here is essentially unchanged aside from the `ensureTxn` call and some renaming.

> Our current use case for reading the stream is vat transcript replay, and that doesn't do a lot of other DB activity (syscalls are handled from the transcript, not executed for real), but Mathieu's #6664 XS/GC bug workaround involves allowing certain `vatstoreGet` calls to be executed during replay, and those would cause DB reads. We aren't landing that on trunk, but it suggests that we shouldn't depend too deeply on excluding other DB activity during sequential reads of a list like this.

This should not be a problem, as reads do not interfere with reads.  The issue is trying to do an `UPDATE` while in the midst of a `SELECT`.  In this case that should not ever happen.

packages/swing-store/src/swingStore.js
FUDCo: I like putting them each before the place in the source code where it's used, in a manner analogous to global or module scoped variable declarations.

packages/swing-store/src/swingStore.js
FUDCo: For complex, highly parameterized statements `SELECT`, for readability I strongly prefer the multi-line format, where each clause is its own line.

packages/swing-store/src/swingStore.js
FUDCo: Yup. Missed the first time on pre-commit cleanup.

packages/swing-store/src/swingStore.js
FUDCo: No.  `db.exec` is not safe inside a transaction.  I was actually thinking of going the other way and changing our few remaining calls to `db.exec` (used in table creation) to use prepared statements instead.

packages/swing-store/src/swingStore.js
FUDCo: I learned this several years ago.  Don't ask me how it happened because the answer is embarrassing.

packages/swing-store/src/swingStore.js
FUDCo: It *is* doing IO.  It's not just computing the hashes, it's writing them to the DB.  I like "emit" because it captures both the computation and the output.

packages/swing-store/src/swingStore.js
FUDCo: I don't like that design.  Most use cases really don't want to have to care if the DB already exists or not, they just want to get a store back.  What they _do_ care about is "start from scratch, regardless of the previous state of the world" vs. "continue where you last left off from, even that if that was t0".

packages/SwingSet/misc-tools/rekernelize.js
FUDCo: I'll just take it out now, lest we forget.

packages/SwingSet/src/types-external.js
FUDCo: Yeah.  I don't know why TypeScript didn't whine about this.  I'll check what's going on and make it tidy.

FUDCo: Mystery solved: the `HostStore` type had been superseded (by `SwingStoreKernelStorage` or `SwingStoreHostStorage` depending on context) but not removed.  And nothing complains about a type that is defined but never referenced anyhere.

packages/SwingSet/src/controller/controller.js
FUDCo: It does, but (1) I'm unclear what clarification is provided by such a comment and (2) why comment this one call to `queueToVatRoot` in particular?

packages/SwingSet/src/controller/controller.js
FUDCo: Essentially same response as above re: `queueToVatRoot`.

packages/SwingSet/src/controller/controller.js
FUDCo: It does need it and indeed has it -- in `kernel.changeKernelOptions`.

packages/SwingSet/src/controller/initializeSwingset.js
FUDCo: I think this may have been a vestige of the "run the LMDB+crankBuffer version parallel to the SQLite version and make sure the hashes line up" exercise.  I will verify that it's safe to remove and do so if so.

packages/SwingSet/src/kernel/state/storageHelper.js
FUDCo: Yep.  More editing debris.

packages/SwingSet/src/kernel/state/kernelKeeper.js
FUDCo: Noice.

packages/SwingSet/src/kernel/state/kernelKeeper.js
FUDCo: That's an easy enough refactor, I just did it.

packages/SwingSet/test/change-parameters/test-change-parameters.js
FUDCo: Lint disablers for import order tend to accumulate because lint whines about import order unpredictably but never complains about being redundantly told to shut up.

packages/SwingSet/test/test-gc-actions.js
FUDCo: I added this check and it consistently fails.  Some investigation in the debugger calls into question the fundamental correctness of the test's fake `kernelKeeper`.  Since I'm not 100% clear what it *should* be doing, we need to talk.

packages/SwingSet/test/upgrade/test-upgrade.js
FUDCo: Prettier doesn't complain about the status quo, but I have no problem with the change.

packages/SwingSet/test/virtualObjects/test-representatives.js
FUDCo: I *did* but damned if I can reconstruct it.

packages/SwingSet/src/kernel/kernel.js
FUDCo: Easier to fix now than remember to do later.  Done.

packages/swing-store/src/sqlStreamStore.js
FUDCo: I think I would rather find a way for it not to be in the external API.  Since its only use is within the `swing-store` package (and for testing purposes at that), I wish there were some graceful way to avoid exposing it.

packages/swing-store/src/swingStore.js
FUDCo: I think of a crank as "a unit of work from the run queue".  Right now we only use rollback for deliveries but I could imagine non-delivery use cases in some future universe.  I could even imagine such use cases placing the rollback savepoints in different places than we do for deliveries.

packages/swing-store/src/sqlStreamStore.js
FUDCo: Upon further reflection, it strikes me that, regardless of API factoring, the useful test in `test-state.js` would be to add some stream state to the test that exercises `getAllState` and `setAllState` and verify that they do the right thing.

warner:
Oh, an update on the WAL-mode and "checkpoint" operations. I just re-read https://sqlite.org/wal.html#performance_considerations , and realized there's a difference between the `fsync()` durability action and the "checkpoint" consolidate-the-WAL performance action.

It says that if `PRAGMA synchronous` is set to `NORMAL` (the default), the DB only does an `fsync()` during a checkpoint operation. That improves performance, but is not safe against a power failure: you could lose data that was written (and committed), until a checkpoint happens. These checkpoints happen automatically once the WAL file reaches 1000 pages, or when initiated manually, but I don't think we're comfortable with a durability vulnerability window like that.

If we set `PRAGMA synchronous` to `FULL`, then it performs an `fsync()` on every `commit()`, which sounds like what we want. We only do a real `commit()` once per block (no faster than once every 5 seconds), so the performance should be fine.

Once we do that, we can stick to automatic checkpoints. These will be opportunistic: if a reader has a transaction open (e.g. to copy data out of the DB, maybe for state-sync purposes), the checkpoint/WAL-compaction might have to stop, but it will pick up where it left off at the next opportunity, and the application doesn't need to know about it.

So:

* when creating the DB, set `PRAGMA synchronous` to `FULL`, right next to where we set `PRAGMA journal_mode=WAL`
* when doing a commit, don't force a checkpoint

packages/SwingSet/test/vat-admin/test-create-vat.js
warner: I tracked this one down: the test was buggy. At this point in the execution, there should be *three* references to the "getHeld" (`ko27`) object: one from v1-bootstrap, one from `kp45` (the result promise known as `kpid1`, which we got back from the `c.queueToVatRoot(.., 'getHeld'`), and the "external" one which `kpResolution()` adds to each slot in the resolution value. The first two were applied within delivery cranks, so their updates were committed through the crankbuffer already. But the `kpResolution()` "external" ref is still sitting in the crankbuffer. I instrumented `kernelKeeper.setRefCount()` and it shows the `ko27` count being set to `3,3`, but the test's `findRefs()` reports back `2,2`: it's fetching the stale value from `kvStore`, which will be updated the next time a `commitCrank()` happens .

By adding a `kernelKeeper.commitCrank()` to the end of `kpResolution()`, I was able to make the original test fail, and adding your extra `expectedRefCount +=1` made it pass again.

In your new version, there is no crankbuffer, so when `kpResolution()` increments the refcount, the test can see it immediately.

I'll write up a patch for the test (as a PR aimed at your branch) to update the comments in that test and explain what's happening.

packages/SwingSet/test/test-state.js
warner: Ok, yes, it should work just as well as it did before. Previously, the LMDB `getKeys()` gave us an [actual iterator](https://www.npmjs.com/package/lmdb#user-content-dbgetkeysoptions-rangeoptions-iterableany), anchored in a read transaction (so subsequent writes wouldn't change the data), swing-store exposed that directly (using `yield* ensureTxn().getKeys(rangeOptions)`), and `kernelSyscall.js > vatstoreGetAfter` used [`clearVatStoreIteration()`](https://github.com/Agoric/agoric-sdk/blob/4143288b0a030309c94c7dc2e8858318f26dea48/packages/SwingSet/src/kernel/kernelSyscall.js#L54) to abandon the iterator if it saw any sign of those writes, or if a different vat shows up to iterate (since LMDB only offered one iterator at a time, I think).

This new scheme uses a SQL query to fetch *all* rows (no iterator) with `keys = sqlKVGetKeys.all(..)`, then `yield* keys.values()` to return an iterator over that saved-in-RAM copy. `kernelSyscall.js` does the same as before, so it's abandoning that iterator under the same conditions. A vat might manage to hold onto the iterator for quite a while: in the LMDB scheme, that would have kept fetching data from the same read txn (and since nothing outside `kernelSyscall.js` would modify that portion of the data, the vatstore, this wouldn't cause any problems). In the SQLite scheme, the read txn is finished promptly, because we pull all values into RAM right away.

So the new scheme will use more RAM when getKeys() is used. The use cases I can find are:
* `vatstoreGetAfter` (called from vat userspace, one key at a time)
* `cleanupAfterTerminatedVat` (in kernelKeeper): to delete/orphan/reject c-list entries, during vat termination
* `getStaticVats` / `getDevices`: at startup to prepare all vats/devices
* `dumpState` (vatKeeper): to walk whole c-list
* `makeBufferedStorage` (lib/storageAPI.js): now unused by swingset, #6704 is to remove this
* a commented-out version of `getImporters` in src/vats/comms/state.js
  * the enabled version stores all the references in a single key, as a serialized list
  * we should decide whether it's worth doing something more sophisticated: the enabled version would be slow for lrefs that are shared among lots of importers, but I don't think that's generally the case for us

The `vatstoreGetAfter` could be large: virtual collections used this for any iterators (`.keys()`, `.values()`, `.entries()`). The swing-store will load the entire virtual collection into RAM, then feed out one entry at a time to `kernelSyscall.js`. We need to replace this with a `getNext()` -like one-key-at-a-time scheme before this gets released.

Also, the c-list deletion code in `cleanupAfterTerminatedVat` could be large, it will load a significant portion of the c-list into RAM at a time (first all the exported objects, then all the imported objects, then the promises). We can start by replacing this with `getNext()`, but hopefully we can replace it with a more dedicated "delete everything" API as part of #6677.

packages/SwingSet/test/vat-admin/test-create-vat.js
warner: https://github.com/Agoric/agoric-sdk/pull/6705

packages/SwingSet/test/test-state.js
mhofman: > This new scheme uses a SQL query to fetch _all_ rows (no iterator) with `keys = sqlKVGetKeys.all(..)`, then `yield* keys.values()` to return an iterator over that saved-in-RAM copy.

> The `vatstoreGetAfter` could be large: virtual collections used this for any iterators (`.keys()`, `.values()`, `.entries()`). The swing-store will load the entire virtual collection into RAM, then feed out one entry at a time to `kernelSyscall.js`. We need to replace this with a `getNext()` -like one-key-at-a-time scheme before this gets released.

This concerns me. If there are millions of entries in a collection, we should not ever be holding all keys in RAM.

Furthermore, can a vat driven iteration cross crank boundaries? What happens to that in RAM state then?

I see 3 solutions:
- Change the iteration model so that insertions (and deletes) to the collection impact an existing iteration if the iterator has not yet reached that position. JS built-in collections will include newly inserted elements, but that's not fully relevant since they are iterated by in insertion order.
  - There ought to be an SQL query to select a single row after another one given some filter & ordering constraint?
- If we want to skip newly inserted entries (and I assume keep skipping deleted but not yet iterated on entries), SQLite in WAL mode does support read transactions like LMDB does. The problem is restoring these transactions across restarts. This is only a problem if the iteration needs to cross crank boundaries (well block boundaries really).
  - An alternative would be to have a table of "existing iterators" which references rows the iterator needs to iterate over, aka copy keys into a separate iteration table.

For context, at some point I knew how the different iterations worked, but I've completely forgotten, so I may be completely off the mark here.

FUDCo: > can a vat driven iteration cross crank boundaries?

No.

Dealing with long lived/large iterators has been an issue in our queue for a while now.

FUDCo: Which is to say, it's a big issue but not one for this PR.

FUDCo:
@warner ready for re-review pending your final analyses of the test changes.

packages/swing-store/package.json
warner: Filed #6703 about making `prepare-test-env-ava.js` into a separate package somehow.

packages/swing-store/src/hasher.js
warner: Chip and I talked through this, and decided that it was fine to leave `sample()` in place.

packages/swing-store/src/sqlStreamStore.js
warner: @FUDCo did we decide to remove `dumpStreams` from the `streamStore` returned as part of `makeSwingStore`, and move it to a `debug` component?

packages/swing-store/src/swingStore.js
warner: note to self: there's also a [`PRAGMA user_version`](https://www.sqlite.org/pragma.html#pragma_user_version) that gets you an integer without needing a whole table to hold it (the integer lives in the SQL file header), which looks like it's meant for this purpose. Also an `application_id` (32-bit signed integer).

packages/swing-store/src/swingStore.js
warner: huh, ok, didn't know that. Yeah, then switching everything else to use `db.prepare(foo).run()` sounds reasonable

packages/swing-store/src/swingStore.js
warner: Let's chat further (outside the scope of this PR). I think most applications *should* care about existing/not-existing: if they don't care, they're doing it wrong, and the result is an easier path to accidents (accidentally clobbering your precious state directory because you re-ran the `init` subcommand, or getting inconsistent behavior out of your `run` subcommand. Swingset host applications have a "initialize a blank directory" step, run once in e.g. the entire history of the chain, and then one or more "launch the application from the already-initialized directory" steps, which they do every time they reboot. That maps directly onto the API I advocated. And I think "sometimes create, sometimes re-use existing" is an anti-pattern that is more useful for test cases and docs than real applications. (Mind you, I've changed my opinion on this multiple times over the course of my career). I'll file an issue.

packages/cosmic-swingset/src/launch-chain.js
warner: For the record, we looked into this, and `inboundQueue` is manipulated within a block, so the commit happens at the end of the block, along with everything else. It's not exactly inside a crank, but in the same way that doing a `controller.installBundle()` is not exactly inside a crank: it's inside a crank-like phase, driven by the host app, before a `controller.run()`, which does not need any sort of revert-to-start-of-crank handling.

packages/SwingSet/test/test-gc-actions.js
warner: We discussed: the test was not as stateful as I was thinking. The assertion I wanted happens to be handled by the very first clause (labelled "`// idle`"), and does not really need to be exercised at the end. But Chip added an explicit clause for it with a good comment, so I'm completely satisfied.

packages/SwingSet/test/test-state.js
warner: @mhofman writes:
> Furthermore, can a vat driven iteration cross crank boundaries? What happens to that in RAM state then?

There are three layers:
* a vat, where liveslots and `collectionManager.js` [create a user-facing iterator](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/liveslots/collectionManager.js#L433-L438) around a sequence of `vatstoreGetAfter()` syscalls, which take a couple of keys (prior, start, end) and maybe return a single key/value pair
* `kernelSyscall.js`, which [implements `vatstoreGetAfter()`](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/kernel/kernelSyscall.js#L162) around a [sometimes-long-lasting iterator](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/kernel/kernelSyscall.js#L51) that it [gets back from `kvStore.getKeys()`](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/kernel/kernelSyscall.js#L209)
* swingstore's implementation of `kvStore.getKeys()`, which [used to delegate to an LMDB iterator](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/swing-store/src/swingStore.js#L238-L254), and is [now implemented by iterating through a big Array](https://github.com/Agoric/agoric-sdk/blob/0d0f8f37f0d2b3fb16ff494ee2ed339e09a2fb84/packages/swing-store/src/swingStore.js#L305-L317) that it gets by immediately and completely consuming a short-lived SQLite query iterator

The `workingKeyIterator` in `kernelSyscall.js` can outlive a single crank, but it gets invalidated upon any vatstoreSet or vatstoreDelete, or if vatstoreGetAfter is called with a `priorKey` that doesn't match the remembered value. The actual `priorKey` comparison uses the output of `vatstoreKeyKey()`, which incorporates the vatID, which prevents different vats from observing each other's iterations. (I think I remember deciding that this was sound when we wrote this code, but now it's making me anxious and I'm looking forward to removing it).

> There ought to be an SQL query to select a single row after another one given some filter & ordering constraint?

Yeah, that's what the new `getNext` API will use: it will return the key after the one you asked about, perhaps subject to some other limits (like an upper bound), and can return `undefined` if there is no such key (within the range that the particular vat is allowed to read). The plan (well, my plan, I still need to discuss it with @FUDCo) is the bottom two layers to strictly use a stateless, non-iterating API, and if anything in the liveslots layer wants to provide an iterator, it can build one itself out of a series of `getNext` calls. At that point, the question of whether iterators are killed by insertions, or see only newer keys, or whatever, is up to liveslots, and we can figure out the tradeoffs (between "does it work just like JS native iterators" and "how hard is it to implement" and performance) at that layer.

packages/SwingSet/test/virtualObjects/test-representatives.js
warner: oh, the actual diff is *moving* this line, effectively changing the previous order of:
* update esKey on 5
* check rcKey/esKey on 2
* set dataKey on 5

to be the other way around:
* update esKey on 5
* set dataKey on 5
* check rcKey/esKey on 2

(I confused myself into thinking that the "set dataKey on 5" was brand new).

And this pattern of rearranging the order of "update X, set X, check Y" appears three times in this one file.

I don't know what's happening either, but my interest in diving into that question is not high, and my belief that this represents a problem is also not high, so I'm going to let it go. The test is probably too picky (and is *certainly* hard to follow, it's fallen into that "golden master" antipattern), but my hunch is that this is not a real issue.

warner (approved):

warner:
Ok, that looks good. Maybe rebase one last time. Thanks!

packages/cosmic-swingset/src/launch-chain.js
mhofman: Yes, the inbound queue will always be manipulated during "end block", but always outside of cranks (actually outside any swingset run).

I think we can guarantee that any host access of the kvStore will always be performed outside of `controller.run()`. It might be good to add an enforcement mechanism on the host side to hold us to that guarantee.

packages/SwingSet/test/test-state.js
mhofman: > > can a vat driven iteration cross crank boundaries?
> 
> No.



> > Furthermore, can a vat driven iteration cross crank boundaries? What happens to that in RAM state then?
> 
> There are three layers:
> 
> * a vat, where liveslots and `collectionManager.js` [create a user-facing iterator](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/liveslots/collectionManager.js#L433-L438) around a sequence of `vatstoreGetAfter()` syscalls, which take a couple of keys (prior, start, end) and maybe return a single key/value pair
> * `kernelSyscall.js`, which [implements `vatstoreGetAfter()`](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/kernel/kernelSyscall.js#L162) around a [sometimes-long-lasting iterator](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/kernel/kernelSyscall.js#L51) that it [gets back from `kvStore.getKeys()`](https://github.com/Agoric/agoric-sdk/blob/02e058d619f363341398c4f345010599d6b85166/packages/SwingSet/src/kernel/kernelSyscall.js#L209)

Sorry to keep digging here, but what prevents a vat from getting an iterator on a collection, keep it in its heap, and call `next()` on it across crank boundaries? Does liveslot actively drop any alive iterator at the end of a crank?

I only see the vat side iterator being invalidated on collection insertion, and the kernel side iterator being invalidated on vatstore operations. I don't see any mechanism where these iterator are invalidated across crank boundaries.

The risk of course is that the vat iterator would be relying on a DB read transaction / in memory structure that goes away on kernel restart.

packages/swing-store/src/sqlStreamStore.js
warner: see https://github.com/Agoric/agoric-sdk/pull/6722 for the `dumpsStreams` change

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +383,186 @@ function makeSwingStore(dirPath, forceReset, options) {
...
+   * Close the database, abandoning any changes made since the last commit (if
+   * you want to save them, call commit() first).
    */
   async function close() {
     assert(db);
-    closeStreamStore();
-    await doCommit(true);
-    await db.close();
+    commit();
+    db.close();
```
mhofman, line 539: This changes the behavior on close from `abort` to `commit`.

packages/swing-store/src/swingStore.js
```diff
@@ -311,69 +383,186 @@ function makeSwingStore(dirPath, forceReset, options) {
...
+        const sql = db.prepare(`ROLLBACK TO SAVEPOINT t${savepointOrdinal}`);
+        sql.run();
+        savepoints.length = savepointOrdinal;
+        return;
+      }
     }
-    txnFinish(abort ? lmdbAbort : undefined);
-    return Promise.resolve(txnDone)
-      .then(() => {
-        trace(`${abort ? 'abort' : 'commit'}-tx`);
```
mhofman, line 334: This `trace` was not kept in the change

packages/swing-store/src/swingStore.js
warner: Oh, huh, yeah the previous `true` in `doCommit(true)` triggered an LMDB abort. I must have missed this during review (also we don't have any tests of abort-on-`close()`, and we never use it that way, applications just crash instead of calling `close`). Yeah, we should make this abort.

Issue #6553, opened 2022-11-09
all vats that will be deployed in Vaults Release must be upgradable


dckc:
## What is the Problem Being Solved?

Analogous to #5666:

> We need all of these to be upgradable, so we don't find ourselves stuck later when we want to add a feature or fix a bug. That means replacing their `Far` objects with instances of a durable `Kind`.

## Description of the Design

based on [a quick audit](https://github.com/Agoric/agoric-sdk/issues/5014#issuecomment-1309365804) it looks like the gaps are:

```[tasklist]
- [x] vat: zoe (#5283)
- [ ] https://github.com/Agoric/agoric-sdk/pull/6894
- [ ] https://github.com/Agoric/agoric-sdk/pull/6903
- [ ] https://github.com/Agoric/agoric-sdk/issues/7494
- [ ] https://github.com/Agoric/agoric-sdk/pull/7107
- [ ] https://github.com/Agoric/agoric-sdk/issues/7495
- [ ] https://github.com/Agoric/agoric-sdk/issues/5885
- [ ] https://github.com/Agoric/agoric-sdk/issues/4548
- [ ] https://github.com/Agoric/agoric-sdk/issues/7044
- [ ] https://github.com/Agoric/agoric-sdk/issues/7633
```

indicentally, not in pismoA nor vaults:
- https://github.com/Agoric/agoric-sdk/issues/5696

note related issues covering some startup vats/contracts not in the `@agoric/vats` package:
 - walletFactory #5894
 - zoe #4383
 - PSM (and presumably, psmCharter): #6099
 - governance #7248 
   - committee
   - contractGovernor
   - binaryVoteCounter

and there's perhaps some overlap in scope with #4511

## Security Considerations

_?_

## Test Plan

CI test (swingset or bootstrap) of each vat

cc @JimLarson @turadg @michaelfig @Chris-Hibbert

dckc:
@Chris-Hibbert are the `committee`, `contractGovernor` and `binaryVoteCounter` contracts in the scope of #5200?

perhaps it's better to scope this to `packages/vats/`; #4983 covers `packages/inter-protocol`. #4511 isn't scoped to any package(s)... hm...

Chris-Hibbert:
> are the committee, contractGovernor and binaryVoteCounter contracts in the scope of https://github.com/Agoric/agoric-sdk/issues/5200?

Yes.

warner:
> That means replacing their Far objects with instances of a durable Kind.

[...]

Issue #5965, opened 2022-08-15
constrain mailbox to devnet config, to protect against c-list/liveslots DoS attacks


warner:
## What is the Problem Being Solved?

One way to attack a vat is to send it messages that induce a storage obligation (many of these are enumerated in #5953). If the target vat does not deliberatly (or accidentally) hold on to the received data, then our GC implementation can help mitigate a chronic attack (in which no single message is a problem, but the long-term accumulated state might be). But our lack of adequate GC tracking of Promises means that imported `vpid`s will incur a RAM cost that cannot be shed until the attacker decides to resolve that promise.

If generalized swingset-level messaging access to a vat also enables the sender to kill a vat (or at least drive its RAM usage high enough to interfere with operations), then maybe we change our attitude towards that access, and treat it as a closely-held authority. If we go in that direction on our chain, the practical outcome would be that we stop provisioning Mailbox access for all accounts. Signed messages that are delivered to the on-chain smart wallet aren't a problem (those always arrive as strings, and the trusted smart-wallet vat decides what to send to other vats in response). The only pathway for arbitrary swingset messages (bearing `vpids` and creating Promises) to enter into the kernel from external parties is through the Mailbox.

That would mean denying `ag-solo` access by default, which would require:

* 1: most accounts to be able to do their business entirely through the smart wallet
* 2: some sort of governance mechanism to decide which accounts *are* allowed Mailbox access
  * I think we need this for contract installation, at least, but we should probably talk about the contract installing experience and see if maybe it should involve something similar, in spirit, to the smart wallet (and driven by a signed string delivered over the bridge, rather than a `E(zoe).installBundleID()` delivered over the mailbox)

This would be at odds with our general "vats everywhere" philosophy. In the long run, we want metering and fees to be the defense against resource consumption attacks (the message sender can hand you a promise, but they have to pay for it, and their cost goes up as your available resources get consumed).

I don't know that this is feasible/practical, but I wanted to start the discussion.

## Description of the Design

If implemented, we'd need to change our provisioning process to omit the `E(comms).addRemote()` and `E(vattp).addRemote()` steps.

@dckc mentioned that the process already has a flag to enable the smart wallet setup or not, so maybe we could add a flag to enable/disable the comms/mailbox portion, and then set that flag to `false`.

We'd also need some mechanism to mailbox-provision accounts when necessary, and a governance mechanism to trigger it.

## Security Considerations

[...]

Issue #5014, opened 2022-04-05
Sufficient durability / upgradability of Vat contracts for launch


warner:
This is an umbrella ticket to track the pieces needed for upgrading individual contracts and core pieces of the chain.

* Vat Upgrade:
  * #1848
  * ~#4382~ unscheduled
  * bundlecaps to name bundles:
    * #4372
    * #3269
    * #4566
    * #4326
    * ~#1346~ unscheduled
    * #4521
  * bundlecaps to load bundles:
    * #4373
    * ~#4486~ unscheduled
    * #4719
    * #4396
  * zoe using bundlecaps to name contracts:
    * #4563
    * ~#4564~ V+1
    * ~#4565~ V+1
  * vat upgrade gets contract bundlecaps:
    * #4726
    * #4588
    * #4381
* durable objects/collections so upgrade retains behavior:
  * #4495
  * #4255
  * #2004
  * "baggage":
    * #4325
    * #2910 
  * makeKind upgrade
    * #3062
* Liveslots upgrade:
  * #4376
* Kernel upgrade:
  * #4375?
* make contracts durable:
  * #4511
* contract upgrade:
  * #3272
* zoe upgrade:
  * #4383
  * #5013

dckc:
From a PSM slog, I have this list of vats; I wonder to what extent they're covered by existing issues:

 - "v7": "board", #6553
 - "v8": "chainStorage", #6553
 - "v15": "bank", #6849
 - "v6": "zoe", #4383
 - "v1": "bootstrap", #4548
 - "v4": ~~mailbox~~ vattp #5667
 - "v2": "vatAdmin", #5669
 - "v3": "comms", "is transcriptless, so it's already taken care of."
 - "v5": "timer", #5669

vats that are contracts:
 - "v9": "centralSupply", #6553
 - "v10": "mintHolder: USDC_axl", #6553
   - "v11": "mintHolder: AUSD", #6553
   - "v14": "mintHolder: BLD", #6553
 - "v12": "psmCharter", #5200
 - "v13": "committee.econ", #5200
 - "v17": "governor: PSM(USDC_axl)", #5200
   - "v18": "governor: PSM(AUSD)", #5200
   - "v19": "governor: ?", #5200
 - "v16": "walletFactory", #5894
 - "v20": "PSM(USDC_axl)", #6099
   - "v21": "PSM(AUSD)", #6099
 - "v22": "provisionPool" #6553

turadg:
[...]

PR #6568, opened 2022-11-15
Harmonise `core/boot-psm.js` with `core/boot.js`

### LINKED ISSUE #6490: how to get to PSM contract from repl / deploy script?
## Describe the bug

 @thisispalash is working on #5822 and he asked me how his code can talk to the PSM contract. I'm struggling to find a way.

## To Reproduce

1. check out the community-dev branch
2. agoric init; agoric start; agoric open --repl
3. try to look up the PSM and amm
4. find amm but no PSM:

```
command[16] E(home.agoricNames).lookup('instance', 'psm')
history[16] unresolved Promise
command[17] E(home.agoricNames).lookup('instance', 'amm')
history[17] [Object Alleged: InstanceHandle]{}
```

It looks like `decentral-demo-config.json` doesn't start the PSM.

## Expected behavior

PSM contract started in an environment where  @thisispalash can get at it.

## Platform Environment

```
$ git describe --tags --always
agoricxnet-7-412-g3482e0d98
```

## Additional context
community-dev branch

michaelfig:
closes: #6490
#loadgen-branch: mfig-diagnostics

## Description

Merge the changes from the `core/boot-psm.js` mechanism used in the first Agoric VM deployment with the manifest-based `core/boot.js` system.

### Testing Considerations

The `devnet` configuration is now tested by the cosmic-swingset integration tests.

### DIFF
--- a/packages/agoric-cli/src/start.js
+++ b/packages/agoric-cli/src/start.js
@@ -2,6 +2,7 @@
 import chalk from 'chalk';
 import { createHash } from 'crypto';
 import path from 'path';
+import { createRequire } from 'module';
 
 import { Nat, isNat } from '@agoric/nat';
 
@@ -15,6 +16,8 @@ import {
 
 import { makePspawn, getSDKBinaries } from './helpers.js';
 
+const require = createRequire(import.meta.url);
+
 const terminalOnlyFlags = (...flags) => {
   if (process.stdout.isTTY && process.stdin.isTTY) {
     return flags;
@@ -27,10 +30,10 @@ const PROVISION_COINS = [
   `5000000000000000${CENTRAL_DENOM}`,
   `100provisionpass`,
   `100sendpacketpass`,
-  `1000000000000ibc/0123456789abcdef`, // IbcATOM
-  `1000000000000ibc/123456789abcdef0`, // AUSD
-  `1000000000000ibc/23456789abcdef01`,
-  `1000000000000ibc/3456789abcdef012`,
+  `1000000000000ibc/atom1234`, // IbcATOM
+  `1000000000000ibc/toyellie`, // AUSD
+  `1000000000000ibc/usdc1234`,
+  `1000000000000ibc/usdt5678`,
 ].join(',');
 const DELEGATE0_COINS = `50000000${STAKING_DENOM}`;
 const SOLO_COINS = `13000000${STAKING_DENOM},500000000${CENTRAL_DENOM}`;
@@ -509,6 +512,32 @@ export default async function startMain(progname, rawArgs, powers, opts) {
         return exitStatus;
       }
     }
+
+    // Create the full economy chain config.
+    const agServerResolve = spec =>
+      require.resolve(spec, { paths: [agServer] });
+    const coreConfigPath = agServerResolve(
+      '@agoric/vats/decentral-core-config.json',
+    );
+    const economyTemplPath = agServerResolve(
+      '@agoric/cosmic-swingset/economy-template.json',
+    );
+    const [rawSoloAddr, coreConfigJson, economyTemplJson] = await Promise.all([
+      fs.readFile(`${agServer}/ag-cosmos-helper-address`, 'utf-8'),
+      fs.readFile(coreConfigPath, 'utf-8'),
+      fs.readFile(economyTemplPath, 'utf-8'),
+    ]);
+    const soloAddr = rawSoloAddr.trimRight();
+    const economyProposals = JSON.parse(
+      economyTemplJson.replace(/@FIRST_SOLO_ADDRESS@/g, soloAddr),
+    );
+    const economyConfig = JSON.parse(coreConfigJson);
+    economyConfig.coreProposals = economyProposals;
+    await fs.writeFile(
+      `${agServer}/decentral-economy-config.json`,
+      JSON.stringify(economyConfig, null, 2),
+    );
+
     if (!opts.restart) {
       return 0;
     }
@@ -551,9 +580,6 @@ export default async function startMain(progname, rawArgs, powers, opts) {
     let exitStatus;
 
     // Provision the ag-solo, if necessary.
-    const soloAddr = (
-      await fs.readFile(`${agServer}/ag-cosmos-helper-address`, 'utf-8')
-    ).trimRight();
     let bestRpcAddr;
     while (!bestRpcAddr) {
       for (const rpcAddr of rpcAddrs) {
--- a/packages/cosmic-swingset/Makefile
+++ b/packages/cosmic-swingset/Makefile
@@ -146,7 +146,7 @@ scenario2-run-chain-economy: t1/decentral-economy-config.json
 # We want to use the same configuration that will be deployed to the next
 # devnet.agoric.net so that developers can test their code against it locally.
 scenario2-run-chain: ../vats/decentral-devnet-config.json
-	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/../vats/decentral-devnet-config.json" \
+	CHAIN_BOOTSTRAP_VAT_CONFIG="$${CHAIN_BOOTSTRAP_VAT_CONFIG-$$PWD/../vats/decentral-devnet-config.json}" \
 		OTEL_EXPORTER_PROMETHEUS_PORT=$(OTEL_EXPORTER_PROMETHEUS_PORT) \
 		$(AGC) --home=t1/n0 start --log_level=warn $(AGC_START_ARGS)
 
--- a/packages/cosmic-swingset/Makefile
+++ b/packages/cosmic-swingset/Makefile
@@ -143,12 +143,11 @@ scenario2-run-chain-economy: t1/decentral-economy-config.json
 	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/t1/decentral-economy-config.json" \
 		$(MAKE) scenario2-run-chain
 
-scenario2-run-chain-psm: ../vats/decentral-psm-config.json
-	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/../vats/decentral-psm-config.json" \
-		$(MAKE) scenario2-run-chain
-
-scenario2-run-chain:
-	OTEL_EXPORTER_PROMETHEUS_PORT=$(OTEL_EXPORTER_PROMETHEUS_PORT) \
+# We want to use the same configuration that will be deployed to the next
+# devnet.agoric.net so that developers can test their code against it locally.
+scenario2-run-chain: ../vats/decentral-devnet-config.json
+	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/../vats/decentral-devnet-config.json" \
+		OTEL_EXPORTER_PROMETHEUS_PORT=$(OTEL_EXPORTER_PROMETHEUS_PORT) \
 		$(AGC) --home=t1/n0 start --log_level=warn $(AGC_START_ARGS)
 
 # Run a chain with an explicit halt.
--- a/packages/inter-protocol/scripts/init-core.js
+++ b/packages/inter-protocol/scripts/init-core.js
@@ -71,16 +71,15 @@ const installKeyGroups = {
  * @param {(m: string, b: string, opts?: any) => I} opts.install
  * @param {<T>(f: T) => T} [opts.wrapInstall]
  *
+ * @param {object} [options]
+ * @param {{ committeeName?: string, committeeSize?: number}} [options.econCommitteeOptions]
  * @template I
  * @template R
  */
-export const committeeProposalBuilder = async ({
-  publishRef,
-  install: install0,
-  wrapInstall,
-}) => {
-  const { ROLE = 'chain' } = process.env;
-
+export const committeeProposalBuilder = async (
+  { publishRef, install: install0, wrapInstall },
+  { econCommitteeOptions } = {},
+) => {
   const install = wrapInstall ? wrapInstall(install0) : install0;
 
   /** @param {Record<string, [string, string]>} group */
@@ -93,7 +92,7 @@ export const committeeProposalBuilder = async ({
     getManifestCall: [
       getManifestForEconCommittee.name,
       {
-        ROLE,
+        econCommitteeOptions,
         installKeys: {
           ...publishGroup(installKeyGroups.econCommittee),
         },
--- a/packages/inter-protocol/src/proposals/committee-proposal.js
+++ b/packages/inter-protocol/src/proposals/committee-proposal.js
@@ -12,26 +12,45 @@ const zip = (xs, ys) => xs.map((x, i) => [x, ys[i]]);
  * @param {{ options: { voterAddresses: Record<string, string> }}} param1
  */
 export const inviteCommitteeMembers = async (
-  {
-    consume: {
-      zoe,
-      agoricNames,
-      namesByAddressAdmin,
-      economicCommitteeCreatorFacet,
-      reserveGovernorCreatorFacet,
-      ammGovernorCreatorFacet,
-      vaultFactoryGovernorCreator,
-    },
-    produce: { econCharterStartResult },
-    installation: {
-      consume: { binaryVoteCounter: counterP },
-    },
-    instance: {
-      consume: { amm, reserve, VaultFactory },
-    },
-  },
+  { consume: { namesByAddressAdmin, economicCommitteeCreatorFacet } },
   { options: { voterAddresses } },
 ) => {
+  const invitations = await E(
+    economicCommitteeCreatorFacet,
+  ).getVoterInvitations();
+  assert.equal(invitations.length, values(voterAddresses).length);
+
+  /**
+   * @param {[string, Promise<Invitation>][]} addrInvitations
+   */
+  const distributeInvitations = async addrInvitations => {
+    await Promise.all(
+      addrInvitations.map(async ([addr, invitationP]) => {
+        await reserveThenDeposit(
+          `econ committee member ${addr}`,
+          namesByAddressAdmin,
+          addr,
+          [invitationP],
+        );
+      }),
+    );
+  };
+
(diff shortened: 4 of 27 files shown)


dckc:
I tried the steps to reproduce from #6490 ; it doesn't look fixed.

```
command[9] E(home.agoricNames).lookup('instance', 'psm')
history[9] Promise.reject("Error: \"nameKey\" not found: (a string)")
```

michaelfig:
> I tried the steps to reproduce from #6490 ; it doesn't look fixed.
> 
> ```
> command[9] E(home.agoricNames).lookup('instance', 'psm')
> history[9] Promise.reject("Error: \"nameKey\" not found: (a string)")
> ```

Ahh, yes.  That's because there's an instance per brand pair.  Try `E(E(home.agoricNames).lookup('instance')).entries()`

packages/cosmic-swingset/Makefile
```diff
@@ -143,12 +143,9 @@ scenario2-run-chain-economy: t1/decentral-economy-config.json
...
 		$(MAKE) scenario2-run-chain
 
-scenario2-run-chain-psm: ../vats/decentral-psm-config.json
-	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/../vats/decentral-psm-config.json" \
-		$(MAKE) scenario2-run-chain
-
-scenario2-run-chain:
-	OTEL_EXPORTER_PROMETHEUS_PORT=$(OTEL_EXPORTER_PROMETHEUS_PORT) \
+scenario2-run-chain: ../vats/decentral-devnet-config.json
+	CHAIN_BOOTSTRAP_VAT_CONFIG="$$PWD/../vats/decentral-devnet-config.json" \
```
turadg: a little surprising to have this default to devnet. consider a comment to call it out and say why

packages/inter-protocol/scripts/init-core.js
```diff
@@ -71,15 +71,16 @@ const installKeyGroups = {
...
-  install: install0,
-  wrapInstall,
-}) => {
-  const { ROLE = 'chain' } = process.env;
+export const committeeProposalBuilder = async (
+  { publishRef, install: install0, wrapInstall },
+  options = {},
+  { env = process.env } = {},
+) => {
+  const { ROLE = env.ROLE || 'chain', econCommitteeOptions } = options;
```
turadg: better to pass `process.env` than rely on ambient authority, but why pass it at all if it can come in through `options.ROLE`? Multiple ways to express the same configuration makes the code harder to follow. It's not apparent without reading `committeeProposalBuilder` that `env.ROLE` is valid or takes precedence over `options.ROLE`.

packages/inter-protocol/scripts/init-core.js
```diff
@@ -71,15 +71,16 @@ const installKeyGroups = {
  * @param {(m: string, b: string, opts?: any) => I} opts.install
  * @param {<T>(f: T) => T} [opts.wrapInstall]
  *
+ * @param options
```
turadg:
```suggestion
 * @param {{ ROLE: 'chain' | 'whatever else', econCommitteeOptions: ??? }} options
```

packages/inter-protocol/src/proposals/committee-proposal.js
```diff
@@ -58,15 +60,19 @@ export const inviteCommitteeMembers = async (
...
-          E(votingAPI).makeNullInvitation(),
-        ]);
         await reserveThenDeposit(
           `econ committee member ${addr}`,
           namesByAddressAdmin,
           addr,
-          [voterInvitation, nullInvitation],
+          [
+            invitationP,
+            // TODO: This null invitation is used just to identify the
```
turadg: TODO by when? Ticket?

packages/inter-protocol/src/proposals/startPSM.js
```diff
@@ -448,21 +446,36 @@ export const PSM_MANIFEST = harden({
...
+      consume: { [Stable.symbol]: 'zoe' },
     },
   },
 });
 
+export const getManifestForPsmGovernance = (
+  { restoreRef },
+  { installKeys },
+) => {
+  const { [installGovAndPSMContracts.name]: _toss, ...manifest } =
```
turadg: isn't simply `_` the convention here for ignoring a param?
```suggestion
  const { [installGovAndPSMContracts.name]: _, ...manifest } =
```

packages/inter-protocol/src/proposals/utils.js
```diff
@@ -63,8 +63,16 @@ export const reserveThenDeposit = async (
     [addr, DEPOSIT_FACET],
   ]);
   console.info('depositing to', debugName);
-  await Promise.all(payments.map(payment => E(depositFacet).receive(payment)));
-  console.info('confirmed deposit for', debugName);
+  await Promise.allSettled(
+    payments.map(async (paymentP, i) => {
+      const payment = await paymentP;
+      await E(depositFacet).receive(payment);
+      console.info(
```
turadg, line 70: nice diagnostic info

packages/smart-wallet/src/walletFactory.js
```diff
@@ -36,9 +37,16 @@ export const publishDepositFacet = async (
   namesByAddressAdmin,
 ) => {
   const { nameHub, myAddressNameAdmin } = makeMyAddressNameAdminKit(address);
-  myAddressNameAdmin.update(WalletName.depositFacet, wallet.getDepositFacet());
+  myAddressNameAdmin.reserve(WalletName.depositFacet);
 
-  return E(namesByAddressAdmin).update(address, nameHub, myAddressNameAdmin);
+  // Don't clobber an existing myAddressNameAdmin or deposit facet.
```
turadg: this will silently fail if `publishDepositFacet` is called a second time and `wallet.getDepositFacet()` changes.

This does seem like an improvement to make it immutable, but shouldn't we error if something tries to change it? At least log?

packages/vats/scripts/init-core.js
```diff
@@ -0,0 +1,30 @@
+import { makeHelpers } from '@agoric/deploy-script-support';
+
+export const defaultProposalBuilder = async ({ publishRef, install }) =>
```
turadg, line 3: is "default" here meant like "default export" or are there expected to be multiple proposal builders in a module like this?

packages/vats/src/core/chain-behaviors.js
```diff
@@ -146,8 +147,22 @@ export const bridgeProvisioner = async ({
...
-          return E(provisioning)
-            .pleaseProvision(nickname, address, powerFlags)
+          let provisionP;
+          if (powerFlags.includes(PowerFlags.SMART_WALLET)) {
+            // Only provision a smart wallet.
+            provisionP = E(bridgeManager).toBridge(
+              BRIDGE_ID.PROVISION_SMART_WALLET,
+              obj,
+            );
+          } else {
```
turadg, line 159: this is assuming the if `SMART_WALLET` isn't in the powerFlags that it should do this. why not validate the contents? I'm thinking the final `else` should be an error for an undefined case

dckc:
> Try `entries()`

I should have been clear: yes, I did that too. There are no PSMs

```
command[5] E(lookup('agoricNames', 'instance')).entries()
history[5] [...]

command[8] h[5].map(([n, _v]) => n)
history[8] ["economicCommittee","amm","ammGovernor","VaultFactory","feeDistributor","Treasury","VaultFactoryGovernor","stakeFactory","stakeFactoryGovernor","Pegasus","reserve","reserveGovernor","psmCharter","interchainPool","provisionPool","walletFactory"]
```

packages/smart-wallet/src/walletFactory.js
dckc: More like it silently succeeds, yes?
This is saying "if there's already one there, leave it alone. Otherwise, use this one"

turadg: Success or failure depends on the intent ;-)

My point is: if what's already there doesn't match, isn't that a surprise? In what cases would you expect the existing value to be different than the new one?

dckc: > In what cases would you expect the existing value to be different than the new one?

In case an ag-solo already registered it.

> isn't that a surprise?

Not any longer, IIUC.

dckc:
I see PSMs now:

```
command[0]
E(E(home.agoricNames).lookup('instance')).keys()
history[0]
["economicCommittee","amm","ammGovernor","VaultFactory","feeDistributor","Treasury","VaultFactoryGovernor","stakeFactory","stakeFactoryGovernor","Pegasus","reserve","reserveGovernor","psmCharter","interchainPool","provisionPool","walletFactory","psm-IST-USDC_axl","psm-IST-USDC_grv","psm-IST-USDT_axl","psm-IST-USDT_grv","psm-IST-AUSD"]
```

fyi, @thisispalash

packages/vats/src/core/chain-behaviors.js
michaelfig: Can you please be more specific?  What kind of validation do you expect, since the Golang side of things already enforces that `powerFlags` is an array of strings?

packages/vats/scripts/init-core.js
michaelfig: The `coreProposals` swingset config feature uses `defaultProposalBuilder` by default (i.e. if the config file does not specify a different entrypoint).

packages/smart-wallet/src/walletFactory.js
michaelfig: I'll add documentation to this effect.

packages/inter-protocol/scripts/init-core.js
michaelfig: Turns out `ROLE` was unnecessary.  Removed.

packages/inter-protocol/src/proposals/committee-proposal.js
michaelfig: #6598

packages/agoric-cli/src/start.js
```diff
@@ -509,6 +509,30 @@ export default async function startMain(progname, rawArgs, powers, opts) {
         return exitStatus;
       }
     }
+
+    // Create the full economy chain config.
+    const coreConfigPath = path.resolve(
+      'node_modules/@agoric/vats/decentral-core-config.json',
```
dckc: This seems to assume that the current working directory has `node_modules` with the `@agoric/vats` package installed. That seems to be a new assumption. I suppose it's justified by the way `agoric install` works?

Elsewhere I've see a function that resolves paths like `@agoric/vats/decentral-core-config.json`. I'm not sure why we don't use it here.

But I don't feel strongly about it. The whole `agoric-cli` package seems to be a bit of a wild west w.r.t. ocap discipline.

packages/vats/decentral-devnet-config.json
```diff
@@ -83,6 +125,18 @@
...
+    },
+    "network": {
+      "sourceSpec": "@agoric/vats/src/vat-network.js"
+    },
+    "priceAuthority": {
+      "sourceSpec": "@agoric/vats/src/vat-priceAuthority.js"
+    },
+    "provisioning": {
+      "sourceSpec": "@agoric/vats/src/vat-provisioning.js"
+    },
```
dckc, line 142: I suppose this calls for re-opening #5965 to think it over again.

packages/vats/decentral-devnet-config.json
```diff
@@ -83,6 +125,18 @@
...
     },
+    "ibc": {
+      "sourceSpec": "@agoric/vats/src/vat-ibc.js"
+    },
+    "network": {
+      "sourceSpec": "@agoric/vats/src/vat-network.js"
+    },
+    "priceAuthority": {
+      "sourceSpec": "@agoric/vats/src/vat-priceAuthority.js"
+    },
```
dckc, line 139: why does `vat-priceAuthority.js` belong here rather than in a vaults proposal?
(semi-rhetorical; I suppose I could figure it out if I looked around.)

packages/vats/decentral-devnet-config.json
```diff
@@ -83,6 +125,18 @@
     "board": {
       "sourceSpec": "@agoric/vats/src/vat-board.js"
     },
+    "ibc": {
+      "sourceSpec": "@agoric/vats/src/vat-ibc.js"
+    },
+    "network": {
+      "sourceSpec": "@agoric/vats/src/vat-network.js"
+    },
```
dckc, line 136: @ivanlei the [list of vats/contracts](https://github.com/Agoric/agoric-sdk/issues/5014#issuecomment-1309365804) that I put together Nov 9 for planning purposes didn't include these two (ibc, network). I suppose they should be added to the list of things to make upgradeable (#6553). Ah... we already have #5696 for the network vat. I don't know whether we included it in any plans.

packages/inter-protocol/scripts/init-core.js
dckc: Seems to still be there as of 2a2b2881946288ad0c7191

packages/inter-protocol/src/proposals/startPSM.js
dckc: Why are we removing `installGovAndPSMContracts` from the manifest returned by `getManifestForPsmGovernance`? It seems particularly relevant.

packages/vats/src/core/startWalletFactory.js
```diff
@@ -108,7 +108,8 @@ const startGovernedInstance = async (
  * }>} powers
  * @param {{
  *   options?: {
- *     perAccountInitialValue?: bigint
+ *     perAccountInitialValue?: bigint,
+ *     walletBridgeId?: string
```
dckc:
```suggestion
 *     walletBridgeId?: string,
```

dckc (approved):
In the interest of addressing #6490 , I'm ready to land this.

It relaxes some constraints that we introduced in `boot-psm.js`, so I re-opened #4165 and  #5965 to re-consider our approach to those.

packages/inter-protocol/scripts/init-core.js
michaelfig: > Seems to still be there

I need to push.  I'll rerequest your review when I do.

packages/agoric-cli/src/start.js
michaelfig: The dependency on `node_modules` was ugly and unnecessary, so I've now abstracted it behind `require.resolve`.

packages/inter-protocol/src/proposals/startPSM.js
michaelfig: The `installations` property below subsumes what `installGovAndPSMContracts` does.

packages/vats/decentral-devnet-config.json
michaelfig: `vat-priceAuthority.js` is not a Zoe contract, so doesn't easily fit into the current way proposals are implemented.  But yes, ideally it would be part of a proposal, and we would make a separate decision whether to publish it to solo clients.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -146,9 +147,24 @@ export const bridgeProvisioner = async ({
     async fromBridge(_srcID, obj) {
       switch (obj.type) {
         case 'PLEASE_PROVISION': {
-          const { nickname, address, powerFlags } = obj;
-          return E(provisioning)
-            .pleaseProvision(nickname, address, powerFlags)
+          const { nickname, address, powerFlags: rawPowerFlags } = obj;
+          const powerFlags = rawPowerFlags || [];
```
dckc, line 151: don't we usually write it this way?

```suggestion
          const { nickname, address, powerFlags = [] } = obj;
```

packages/vats/src/core/chain-behaviors.js
```diff
@@ -146,9 +147,24 @@ export const bridgeProvisioner = async ({
     async fromBridge(_srcID, obj) {
```
dckc, line 147: how does `fromBridge` here relate to the one in `provisionPool.js`? I'd like to see compare/contrast comments on each.

packages/vats/src/core/chain-behaviors.js
dckc: - wish: unit test
 - wish: pattern-based dynamic type check on `obj`

packages/solo/src/chain-cosmos-sdk.js
```diff
@@ -690,8 +690,7 @@ ${chainID} chain does not yet know of address ${clientAddr}${adviseEgress(
           console.debug(`helper said: ${stdout}`);
           const out = JSON.parse(stdout);
 
-          out.code ||
-            0 ||
+          out.code === 0 ||
```
dckc, line 693: yikes. that looks like it was hairy to track down.

wish: unit test

dckc (approved):
still looks good.

I have a few wishes...

packages/vats/src/core/chain-behaviors.js
michaelfig: Turns out that `obj.powerFlags` can be `null`.  The destructuring default values only work if `obj.powerFlags` is `undefined`.

Yeah, the dynamic checks would be better with patterns.

packages/vats/src/core/chain-behaviors.js
```diff
@@ -146,9 +147,24 @@ export const bridgeProvisioner = async ({
...
-          return E(provisioning)
-            .pleaseProvision(nickname, address, powerFlags)
+          const { nickname, address, powerFlags: rawPowerFlags } = obj;
+          const powerFlags = rawPowerFlags || [];
+          let provisionP;
+          if (powerFlags.includes(PowerFlags.SMART_WALLET)) {
+            // Only provision a smart wallet.
+            provisionP = E(bridgeManager).fromBridge(
+              BRIDGE_ID.PROVISION_SMART_WALLET,
+              obj,
```
mhofman, line 157: Where does this message actually end up? I do not see any bridge handler registered for it since `boot-psm` sets `walletBridgeId` to `BridgeId.PROVISION`.

PR #6572, opened 2022-11-16
fix npm prepack

turadg:
## Description

Now that packages do `tsc --build` in their `prepack`, [governance is failing to pack and publish](https://github.com/Agoric/agoric-sdk/actions/runs/3464380814/jobs/5785901498). 

The problem is that some builds were making `exported.d.ts` which prevents the propagation of ambient types. This PR stops tsc builds from creating `exported.d.ts` by omitting it from configs _and_ hunting down transitive imports.

### Security Considerations

n/a

Some tribal knowledge in ensuring that `exported.d.ts` aren't built. Maybe this git blame to this PR suffices.

### Testing Considerations

Only master builds do the publish step that was failing, so I tested locally by running `yarn prepack` in each of these packages in the order they are in the failing CI job:

1. `internal`
2. `ertp`
3. `zoe`
4. `governance`

### DIFF
--- a/packages/SwingSet/src/vats/timer/types.js
+++ b/packages/SwingSet/src/vats/timer/types.js
@@ -1,0 +1,2 @@
+// Empty JS file to correspond with types.d.ts
+export {};
--- a/packages/vats/src/core/startWalletFactory.js
+++ b/packages/vats/src/core/startWalletFactory.js
@@ -7,6 +7,9 @@ import { CONTRACT_ELECTORATE, ParamTypes } from '@agoric/governance';
 import { makeStorageNodeChild } from '../lib-chainStorage.js';
 import { Stable } from '../tokens.js';
 
+// Ambient types (globals)
+import '@agoric/swingset-vat/src/vats/timer/types.js';
+
 /**
  * @param {ERef<ZoeService>} zoe
  * @param {Installation<import('@agoric/smart-wallet/src/walletFactory').start>} inst
--- a/packages/ERTP/jsconfig.json
+++ b/packages/ERTP/jsconfig.json
@@ -12,8 +12,6 @@
     ]
   },
   "include": [
-    "*.js",
-    "*.ts",
     "src/**/*.js",
     "src/**/*.ts",
     "test"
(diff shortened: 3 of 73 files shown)


packages/vats/src/core/startWalletFactory.js
```diff
@@ -7,6 +7,10 @@ import { CONTRACT_ELECTORATE, ParamTypes } from '@agoric/governance';
 import { makeStorageNodeChild } from '../lib-chainStorage.js';
 import { Stable } from '../tokens.js';
 
+// Ambient types (globals)
+// eslint-disable-next-line import/no-unresolved -- https://github.com/typescript-eslint/typescript-eslint/issues/1624
+import '@agoric/swingset-vat/src/vats/timer/types';
```
mhofman: Missing extension

mhofman (dismissed):
I trust that none of these import removals are reverting some ambiently used  types to implicit `any`?

turadg:
> I trust none of these import removals are reverting some ambiently used types to implicit any?

I trust that too ;-)  To also verify, I tested in `zoe` which had almost all the imports changes. With `noImplicitAny` in master,
```
Found 3458 errors in 329 files.
```
And in this PR:
```
Found 3427 errors in 324 files.
```

So net improvement actually. I haven't confirmed that the latter errors are strictly a subset of the former errors, but I think as long as we have a net improvement in type coverage this should land.

packages/SwingSet/src/vats/timer/types.js
```diff
@@ -0,0 +1,2 @@
+// Empty JS file to correspond with types.d.ts
```
mhofman, line 1: I'm confused, we are runtime importing a `.js` file for ambient type defined in a `.d.ts` file? I didn't expect us to have ambient definitions in `.d.ts` files.

mhofman (approved):
If it passes tests, LGTM

packages/SwingSet/src/vats/timer/types.js
turadg: yes, here https://github.com/Agoric/agoric-sdk/blob/11d9e00ad101f4d59e316a053a69c50867bc4da4/packages/vats/src/core/startWalletFactory.js#L10-L11

ambient (aka global) definitions for this particular file were added in https://github.com/Agoric/agoric-sdk/commit/99375d09193ba4055d9a2166f07af5fc0f38ba03

PR #6577, opened 2022-11-16
fix: use atomicTransfers rather than stagings.

### LINKED ISSUE #3850: Design flaw in Zoe: stateful seats and staged allocations
This is an over-arching issue to capture the architectural design flaw introduced in https://github.com/Agoric/agoric-sdk/pull/3184. #3184 made significant usability and readability gains, at least on the surface, but has since resulted in at least 3 bugs of high severity.

Importantly, until this design flaw is corrected, we can expect more bugs involving staged allocations. 

## Design prior to #3184

The earlier design created `seatStagings` - objects which represented a particular staged allocation on a seat. `zcf.reallocate` acted on `seatStagings`, not `seats`. This meant that multiple `seatStagings` could be reallocated independently of each other, even for the same seat. Moreover, `seatStagings` could be dropped without introducing bugs within Zoe, even though this pattern is not encouraged.

Here's an example of usage in multipoolAutoswap:

```js
const seatStaging = seat.stage(
  harden({
    In: AmountMath.subtract(amountIn, reducedAmountIn),
    Out: amountOut,
  }),
);

const poolBrandInStaging = brandInPool.stageSeat({
  Secondary: AmountMath.add(
    brandInPool.getSecondaryAmount(),
    reducedAmountIn,
  ),
  Central: AmountMath.subtract(
    brandInPool.getCentralAmount(),
    reducedCentralAmount,
  ),
});

const poolBrandOutStaging = brandOutPool.stageSeat({
  Central: AmountMath.add(
    brandOutPool.getCentralAmount(),
    reducedCentralAmount,
  ),
  Secondary: AmountMath.subtract(
    brandOutPool.getSecondaryAmount(),
    amountOut,
  ),
});

zcf.reallocate(poolBrandInStaging, poolBrandOutStaging, seatStaging);  
```

## Design after #3184

What is currently on master for `multipoolAutoswap` is a lot clearer:

```js
brandOutPoolSeat.decrementBy({ Secondary: amountOut });
seat.incrementBy({ Out: amountOut });

brandInPoolSeat.decrementBy({ Central: reducedCentralAmount });
brandOutPoolSeat.incrementBy({ Central: reducedCentralAmount });

seat.decrementBy({ In: reducedAmountIn });
brandInPoolSeat.incrementBy({ Secondary: reducedAmountIn });

zcf.reallocate(brandOutPoolSeat, brandInPoolSeat, seat);
```

The code can easily be audited for rights conservation violations, whereas the previous design could not be. For example, see [this analysis of multipoolAutoswap where a rights conservation bug was found only after deep inspection](https://github.com/Agoric/agoric-sdk/blob/master/packages/zoe/src/contracts/multipoolAutoswap/docs/reallocateCatalog.md).

On the surface, this code seemed a lot more auditable, but under the surface, problems were introduced. Being able to audit for rights conservation violations at the contract level is great, but only helps if bugs aren't introduced in Zoe.

We had at least three bugs introduced in Zoe as a result of this change:

1. https://github.com/Agoric/agoric-sdk/issues/3613 (fixed)
2. https://github.com/Agoric/agoric-sdk/issues/3800 (open)
3. https://github.com/Agoric/agoric-sdk/issues/3848 (fixed)

#3613 resulted from an interaction of `zcfMint` with `stagedAllocations`. `zcfMint.mintGains` calls `reallocateInternal` after incrementing the `zcfSeat` with the assets to mint. `zcf.reallocate`  also calls `reallocateInternal` after doing some work. Thinking mostly of the `zcf.reallocate` usage of `reallocateInternal`, we had required that all seats with staged allocations should be included in `reallocateInternal`. This was intended as a safety measure to ensure that contract code does not unintentionally forget about any seats and their staged allocations. This caused a problem for `zcfMint.mintGains`, where the common usage is in the middle of preparing staged allocations, prior to the reallocation that would use them. `mintGains` was erroring if there were any prior outstanding staged allocations, but this was incorrect behavior because having outstanding staged allocations is common.  The "quick fix" was to move this check into `reallocate` only, such that the `mintGains` path never performed the check.

#3800 was the result of an expected, user-caused error in the AMM (MultipoolAutoswap) causing unexpected state. The readability and usability of the new design #3184 depended entirely on the staged allocations being cleared before an offerHandler was called. In other words, the offerHandlers were written as if the seats involved had no previous staged allocations.  This expectation was violated when an offer handler errored because the offer was not one that the contract wanted to accept, and this error left stagedAllocations in place for future offers. A potential solution would be to clear all staged allocations on errors, but more thought is required on exactly when this clearing should occur. This issue is still open, requiring more design. The design prior to #3184 would have not had this problem, because prior seatStagings would simply be dropped, and no state regarding staged allocations would continue to stay around after an error.

#3848, the latest bug, is the result of `mintGains` being able to perform a reallocate for one seat without rights conservation being checked. Before the new design in #3184, this made sense because only the seatStaging that was involved in `mintGains` would be reallocated against, and minting doesn't conserve rights anyways. Any other seatStagings, even for the same seat, would be unaffected. However, with the new design, the mintGains is adding to a stagedAllocation that may already exist. This means that staged allocations that would violate rights conservation can slip in and become the canonical allocation. Offer safety holds for this particular transaction, but because rights conservation is violated, the offer safety in other offers is violated. Even though "offer safety" might be true in terms of allocations, if rights are not conserved, someone will not be able to get a payout that matches their allocation.

### LINKED ISSUE #6116: reallocate should guarantee that stage failure cannot happen
https://github.com/Agoric/agoric-sdk/blob/757b887edd2d41960fadc86d4900ebde55729867/packages/inter-protocol/src/psm/psm.js#L192

erights:
Fixes #3850
Fixes #6116 

Alternative to https://github.com/Agoric/agoric-sdk/pull/4327

Based on https://github.com/Agoric/agoric-sdk/pull/6102#discussion_r960760639

Revisiting the example from #3850, current code like

```js
  brandOutPoolSeat.decrementBy(harden({ Secondary: amountOut }));
  seat.incrementBy(harden({ Out: amountOut }));

  brandOutPoolSeat.incrementBy(
    brandInPoolSeat.decrementBy(harden({ Central: reducedCentralAmount })));

  seat.decrementBy(harden({ In: reducedAmountIn }));
  brandInPoolSeat.incrementBy(harden({ Secondary: reducedAmountIn }));

  zcf.reallocate(brandOutPoolSeat, brandInPoolSeat, seat);
```

becomes

```js
  atomicTransfer(
    zcf,
    harden([
      [brandOutPoolSeat, seat, { Secondary: amountOut }, { Out: amountOut }],
      [brandInPoolSeat, brandOutPoolSeat, { Central: reducedCentralAmount }],
      [seat, brandInPoolSeat, { In: reducedAmountIn }, { Secondary: reducedAmountIn }],
    ]),
  );
```

No matter whether the `atomicTransfer` succeeds or fails, there are none of the stateful hazards associated with the previous staging approach.

In exchange for `atomicTransfer`, this PR deprecates the `zcf` method `reallocate, and deprecates the `zcfSeat` methods `incrementBy`, `decrementBy`, `clear`, `getStagedAllocation`, and `hasStagedAllocation`.

Currently `atomicTransfer` is a helper, to avoid collisions with ongoing work on Zoe durability. Once we're past that, I expect we'll redefine it as a replacement for `zcf.reallocate`, so the above example becomes

```js
  zcf.transfer(harden([
    [brandOutPoolSeat, seat, { Secondary: amountOut }, { Out: amountOut }],
    [brandInPoolSeat, brandOutPoolSeat, { Central: reducedCentralAmount }],
    [seat, brandInPoolSeat, { In: reducedAmountIn }, { Secondary: reducedAmountIn }],
  ]));
```

### DIFF
--- a/packages/inter-protocol/src/collectFees.js
+++ b/packages/inter-protocol/src/collectFees.js
@@ -1,3 +1,5 @@
+import { atomicTransfer } from '@agoric/zoe/src/contractSupport';
+
 /**
  * Provide shared support for providing access to fees from a service contract.
  *
@@ -14,9 +16,10 @@ export const makeMakeCollectFeesInvitation = (
 ) => {
   const collectFees = seat => {
     const amount = feeSeat.getAmountAllocated(keyword, feeBrand);
-    feeSeat.decrementBy(harden({ [keyword]: amount }));
-    seat.incrementBy(harden({ Fee: amount }));
-    zcf.reallocate(seat, feeSeat);
+    atomicTransfer(
+      zcf,
+      harden([[feeSeat, seat, { [keyword]: amount }, { Fee: amount }]]),
+    );
 
     seat.exit();
     return `paid out ${amount.value}`;
--- a/packages/inter-protocol/src/collectFees.js
+++ b/packages/inter-protocol/src/collectFees.js
@@ -16,10 +16,7 @@ export const makeMakeCollectFeesInvitation = (
 ) => {
   const collectFees = seat => {
     const amount = feeSeat.getAmountAllocated(keyword, feeBrand);
-    atomicTransfer(
-      zcf,
-      harden([[feeSeat, seat, { [keyword]: amount }, { Fee: amount }]]),
-    );
+    atomicTransfer(zcf, feeSeat, seat, { [keyword]: amount }, { Fee: amount });
 
     seat.exit();
     return `paid out ${amount.value}`;
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -7,7 +7,7 @@ import {
   ceilMultiplyBy,
   floorDivideBy,
   floorMultiplyBy,
-  atomicTransfer,
+  atomicRearrange,
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { Far } from '@endo/marshal';
 import {
@@ -163,7 +163,7 @@ export const start = async (zcf, privateArgs, baggage) => {
     const maxAnchor = floorMultiplyBy(afterFee, anchorPerMinted);
     AmountMath.isGTE(maxAnchor, wanted) ||
       Fail`wanted ${wanted} is more than ${given} minus fees ${fee}`;
-    atomicTransfer(
+    atomicRearrange(
       zcf,
       harden([
         [seat, stage, { In: afterFee }, { Minted: afterFee }],
@@ -173,7 +173,7 @@ export const start = async (zcf, privateArgs, baggage) => {
     );
     // The treatment of `burnMinted` here is different than the
     // one immediately below. This `burnMinted`
-    // happen only if the `atomicTransfer` does *not* throw.
+    // happen only if the `atomicRearrange` does *not* throw.
     burnMinted(afterFee);
     totalAnchorProvided = AmountMath.add(totalAnchorProvided, maxAnchor);
   };
@@ -192,7 +192,7 @@ export const start = async (zcf, privateArgs, baggage) => {
       Fail`wanted ${wanted} is more than ${given} minus fees ${fee}`;
     mintMinted(asStable);
     try {
-      atomicTransfer(
+      atomicRearrange(
         zcf,
         harden([
           [seat, anchorPool, { In: given }, { Anchor: given }],
@@ -203,7 +203,7 @@ export const start = async (zcf, privateArgs, baggage) => {
     } catch (e) {
       // The treatment of `burnMinted` here is different than the
       // one immediately above. This `burnMinted`
-      // happens only if the `atomicTransfer` *does* throw.
+      // happens only if the `atomicRearrange` *does* throw.
       burnMinted(asStable);
       throw e;
     }
--- a/packages/inter-protocol/src/psm/psm.js
+++ b/packages/inter-protocol/src/psm/psm.js
@@ -7,6 +7,7 @@ import {
   ceilMultiplyBy,
   floorDivideBy,
   floorMultiplyBy,
+  atomicTransfer,
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { Far } from '@endo/marshal';
 import {
@@ -48,21 +49,6 @@ const { Fail } = assert;
  * @property {Amount<'nat'>} totalMintedProvided  running sum of Minted
  * ever given by this contract
  */
-
-/**
- * Stage a transfer of a single asset from one seat to another, with an optional
- * remapping of the Keywords. Check that the remapping is for the same amount.
- *
- * @param {ZCFSeat} from
- * @param {ZCFSeat} to
- * @param {AmountKeywordRecord} txFrom
- * @param {AmountKeywordRecord} txTo
- */
-const stageTransfer = (from, to, txFrom, txTo = txFrom) => {
-  assert(AmountMath.isEqual(Object.values(txFrom)[0], Object.values(txTo)[0]));
-  from.decrementBy(txFrom);
-  to.incrementBy(txTo);
-};
 
 /** @typedef {import('@agoric/vat-data').Baggage} Baggage */
 
@@ -177,24 +163,18 @@ export const start = async (zcf, privateArgs, baggage) => {
     const maxAnchor = floorMultiplyBy(afterFee, anchorPerMinted);
     AmountMath.isGTE(maxAnchor, wanted) ||
       Fail`wanted ${wanted} is more than ${given} minus fees ${fee}`;
-    try {
-      stageTransfer(seat, stage, { In: afterFee }, { Minted: afterFee });
-      stageTransfer(seat, feePool, { In: fee }, { Minted: fee });
-      stageTransfer(
-        anchorPool,
-        seat,
-        { Anchor: maxAnchor },
-        { Out: maxAnchor },
-      );
-      zcf.reallocate(seat, anchorPool, stage, feePool);
-      burnMinted(afterFee);
-    } catch (e) {
-      stage.clear();
-      anchorPool.clear();
-      feePool.clear();
-      // TODO(#6116) someday, reallocate should guarantee that this case cannot happen
-      throw e;
-    }
+    atomicTransfer(
+      zcf,
+      harden([
+        [seat, stage, { In: afterFee }, { Minted: afterFee }],
+        [seat, feePool, { In: fee }, { Minted: fee }],
+        [anchorPool, seat, { Anchor: maxAnchor }, { Out: maxAnchor }],
+      ]),
+    );
+    // The treatment of `burnMinted` here is different than the
+    // one immediately below. This `burnMinted`
+    // happen only if the `atomicTransfer` does *not* throw.
+    burnMinted(afterFee);
     totalAnchorProvided = AmountMath.add(totalAnchorProvided, maxAnchor);
   };
 
@@ -212,15 +192,18 @@ export const start = async (zcf, privateArgs, baggage) => {
       Fail`wanted ${wanted} is more than ${given} minus fees ${fee}`;
     mintMinted(asStable);
     try {
-      stageTransfer(seat, anchorPool, { In: given }, { Anchor: given });
-      stageTransfer(stage, seat, { Minted: afterFee }, { Out: afterFee });
-      stageTransfer(stage, feePool, { Minted: fee });
-      zcf.reallocate(seat, anchorPool, stage, feePool);
+      atomicTransfer(
+        zcf,
+        harden([
+          [seat, anchorPool, { In: given }, { Anchor: given }],
+          [stage, seat, { Minted: afterFee }, { Out: afterFee }],
+          [stage, feePool, { Minted: fee }],
+        ]),
+      );
     } catch (e) {
-      stage.clear();
-      anchorPool.clear();
-      feePool.clear();
-      // TODO(#6116) someday, reallocate should guarantee that this case cannot happen
+      // The treatment of `burnMinted` here is different than the
+      // one immediately above. This `burnMinted`
+      // happens only if the `atomicTransfer` *does* throw.
       burnMinted(asStable);
       throw e;
     }
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -538,16 +538,9 @@ const helperBehavior = {
           // it. We could hold it until it crosses some threshold, then sell it
           // to the AMM, or we could transfer it to the reserve. At least it's
           // visible in the accounting.
-          atomicTransfer(
-            zcf,
-            harden([
-              [
-                vaultSeat,
-                state.retainedCollateralSeat,
-                { Collateral: collateralPost },
-              ],
-            ]),
-          );
+          atomicTransfer(zcf, vaultSeat, state.retainedCollateralSeat, {
+            Collateral: collateralPost,
+          });
         }
 
         // Reduce totalCollateral by collateralPre, since all the collateral was
--- a/packages/inter-protocol/src/vaultFactory/vaultManager.js
+++ b/packages/inter-protocol/src/vaultFactory/vaultManager.js
@@ -31,6 +31,7 @@ import {
   getAmountOut,
   makeRatio,
   makeRatioFromAmounts,
+  atomicTransfer,
 } from '@agoric/zoe/src/contractSupport/index.js';
 import { E } from '@endo/eventual-send';
 import { unitAmount } from '@agoric/zoe/src/contractSupport/priceQuote.js';
@@ -537,12 +538,16 @@ const helperBehavior = {
           // it. We could hold it until it crosses some threshold, then sell it
           // to the AMM, or we could transfer it to the reserve. At least it's
           // visible in the accounting.
-          vaultSeat.decrementBy(
-            state.retainedCollateralSeat.incrementBy({
-              Collateral: collateralPost,
-            }),
+          atomicTransfer(
+            zcf,
+            harden([
+              [
+                vaultSeat,
+                state.retainedCollateralSeat,
+                { Collateral: collateralPost },
+              ],
+            ]),
           );
-          zcf.reallocate(vaultSeat, state.retainedCollateralSeat);
         }
 
         // Reduce totalCollateral by collateralPre, since all the collateral was
--- a/packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
+++ b/packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
@@ -1,5 +1,9 @@
 import { AmountMath } from '@agoric/ertp';
-import { atomicTransfer } from '@agoric/zoe/src/contractSupport/index.js';
+import {
+  atomicRearrange,
+  fromOnly,
+  toOnly,
+} from '@agoric/zoe/src/contractSupport/index.js';
 import { Far } from '@endo/marshal';
 import { makeFeeRatio } from './constantProduct/calcFees.js';
 import {
@@ -56,27 +60,17 @@ export const makeDoublePool = (
     const inPoolSeat = collateralInPool.getPoolSeat();
     const outPoolSeat = collateralOutPool.getPoolSeat();
 
-    atomicTransfer(
+    atomicRearrange(
       zcf,
       harden([
-        [seat, seat, { In: prices.swapperGives }, { Out: prices.swapperGets }],
-        [
-          inPoolSeat,
-          inPoolSeat,
-          { Central: prices.inPoolDecrement },
-          { Secondary: prices.inPoolIncrement },
-        ],
-        [
-          outPoolSeat,
-          outPoolSeat,
-          { Secondary: prices.outPoolDecrement },
-          { Central: prices.outPoolIncrement },
-        ],
-        // This strange construction is to transfer into without saying at the
-        // same time where this amount was transfered from. But the overall
-        // atomicTransfer still has to be conserved, i.e., balance to
-        // a net zero.
-        [undefined, feeSeat, undefined, { Fee: prices.protocolFee }],
+        fromOnly(seat, { In: prices.swapperGives }),
+        fromOnly(inPoolSeat, { Central: prices.inPoolDecrement }),
+        fromOnly(outPoolSeat, { Secondary: prices.outPoolDecrement }),
+
+        toOnly(seat, { Out: prices.swapperGets }),
+        toOnly(inPoolSeat, { Secondary: prices.inPoolIncrement }),
+        toOnly(outPoolSeat, { Central: prices.outPoolIncrement }),
+        toOnly(feeSeat, { Fee: prices.protocolFee }),
       ]),
     );
 
--- a/packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
+++ b/packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
@@ -1,4 +1,5 @@
 import { AmountMath } from '@agoric/ertp';
+import { atomicTransfer } from '@agoric/zoe/src/contractSupport/index.js';
 import { Far } from '@endo/marshal';
 import { makeFeeRatio } from './constantProduct/calcFees.js';
 import {
@@ -55,15 +56,30 @@ export const makeDoublePool = (
     const inPoolSeat = collateralInPool.getPoolSeat();
     const outPoolSeat = collateralOutPool.getPoolSeat();
 
-    seat.decrementBy(harden({ In: prices.swapperGives }));
-    inPoolSeat.decrementBy(harden({ Central: prices.inPoolDecrement }));
-    outPoolSeat.decrementBy(harden({ Secondary: prices.outPoolDecrement }));
-    seat.incrementBy(harden({ Out: prices.swapperGets }));
-    inPoolSeat.incrementBy(harden({ Secondary: prices.inPoolIncrement }));
-    outPoolSeat.incrementBy(harden({ Central: prices.outPoolIncrement }));
-    feeSeat.incrementBy(harden({ Fee: prices.protocolFee }));
+    atomicTransfer(
+      zcf,
+      harden([
+        [seat, seat, { In: prices.swapperGives }, { Out: prices.swapperGets }],
+        [
+          inPoolSeat,
+          inPoolSeat,
+          { Central: prices.inPoolDecrement },
+          { Secondary: prices.inPoolIncrement },
+        ],
+        [
+          outPoolSeat,
+          outPoolSeat,
+          { Secondary: prices.outPoolDecrement },
+          { Central: prices.outPoolIncrement },
+        ],
+        // This strange construction is to transfer into without saying at the
+        // same time where this amount was transfered from. But the overall
+        // atomicTransfer still has to be conserved, i.e., balance to
+        // a net zero.
+        [undefined, feeSeat, undefined, { Fee: prices.protocolFee }],
+      ]),
+    );
 
-    zcf.reallocate(outPoolSeat, inPoolSeat, feeSeat, seat);
     seat.exit();
     collateralInPool.updateState();
     collateralOutPool.updateState();
--- a/packages/inter-protocol/src/vpool-xyk-amm/pool.js
+++ b/packages/inter-protocol/src/vpool-xyk-amm/pool.js
@@ -7,6 +7,7 @@ import {
   calcLiqValueToMint,
   calcSecondaryRequired,
   calcValueToRemove,
+  atomicTransfer,
 } from '@agoric/zoe/src/contractSupport/index.js';
 
 import { E } from '@endo/eventual-send';
@@ -252,18 +253,21 @@ export const definePoolKind = (baggage, ammPowers, storageNode, marshaller) => {
         ),
       );
 
-      poolSeat.incrementBy(
-        userSeat.decrementBy(harden({ Liquidity: liquidityIn })),
+      atomicTransfer(
+        ammPowers.zcf,
+        harden([
+          [userSeat, poolSeat, { Liquidity: liquidityIn }],
+          [
+            poolSeat,
+            userSeat,
+            {
+              Central: centralTokenAmountOut,
+              Secondary: tokenKeywordAmountOut,
+            },
+          ],
+        ]),
       );
-      userSeat.incrementBy(
-        poolSeat.decrementBy(
-          harden({
-            Central: centralTokenAmountOut,
-            Secondary: tokenKeywordAmountOut,
-          }),
-        ),
-      );
-      ammPowers.zcf.reallocate(userSeat, poolSeat);
+
       state.liqTokenSupply -= liquidityValueIn;
 
       userSeat.exit();
--- a/packages/inter-protocol/src/vpool-xyk-amm/pool.js
+++ b/packages/inter-protocol/src/vpool-xyk-amm/pool.js
@@ -7,7 +7,7 @@ import {
   calcLiqValueToMint,
   calcSecondaryRequired,
   calcValueToRemove,
-  atomicTransfer,
+  atomicRearrange,
 } from '@agoric/zoe/src/contractSupport/index.js';
 
 import { E } from '@endo/eventual-send';
@@ -253,7 +253,7 @@ export const definePoolKind = (baggage, ammPowers, storageNode, marshaller) => {
         ),
       );
 
-      atomicTransfer(
+      atomicRearrange(
         ammPowers.zcf,
         harden([
           [userSeat, poolSeat, { Liquidity: liquidityIn }],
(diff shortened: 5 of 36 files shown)


dtribble:
```javascript
const seatMgr = makeSeatMgr(zcf);

seatMgr.transfer(brandOutPoolSeat, seat, harden({ Secondary: amountOut }), harden({ Out: amountOut }));

seatMgr.transfer(brandInPoolSeat, brandOutPoolSeat, harden({ Central: reducedCentralAmount }));

seatMgr.transfer(seat, brandInPoolSeat, harden({ In: reducedAmountIn }), harden({ Secondary: reducedAmountIn }));

seatMgr.commit();
```

which doesn't look a lot better, but is much more extensible and flexible. I'd also argue the "better" is mostly because the big data structure only has one `harden`.  Compare
```javascript
const seatMgr = makeSeatMgr(zcf);
seatMgr.transfer(brandOutPoolSeat, seat, { Secondary: amountOut }, { Out: amountOut });
seatMgr.transfer(brandInPoolSeat, brandOutPoolSeat, { Central: reducedCentralAmount });
seatMgr.transfer(seat, brandInPoolSeat, { In: reducedAmountIn }, { Secondary: reducedAmountIn });
seatMgr.commit();
```

erights:
> ```js
> seatMgr.transfer(...)
> ```

To the tiny extent that any of the rest of this was done, this is now done as well. I'm putting this PR into Draft because I'm purposely accumulating too much API in the SeatMgr, for purposes of experimenting with their usage. For example, I added `transfer` without removing `incrBy` and `decrBy`.

Therefore this PR in its current state is not a candidate for merging, and won't be until SeatMgr is reduced back to a tasteful API. But it is ready for review, since the point of this intermediate state is to stimulate discussion and get feedback.

erights:
This code should now be feature complete, but it has not yet even been smoke tested. Nevertheless, it is in good shape to get feedback and provoke discussion on the API.

I did verify (only by reading code, not by running tests) that currently a failed `zcf.reallocate` does ***NOT*** clear the stagings. By contrast, a `seatMgr.commit()` clears all stagings on the `finally` path, and therefore on both success and failure.

packages/inter-protocol/src/psm/psm.js
```diff
@@ -179,24 +165,18 @@ export const start = async (zcf, privateArgs, baggage) => {
...
+      zcf,
+      harden([
+        [seat, stage, { In: afterFee }, { Minted: afterFee }],
+        [seat, feePool, { In: fee }, { Minted: fee }],
+        [anchorPool, seat, { Anchor: maxAnchor }, { Out: maxAnchor }],
+      ]),
+    );
+    // TODO The treatment of `burnMinted` here is inconsistent with the
+    // one at the corresponding TODO comment below. This `burnMinted`
+    // doesn't happen if the `atomicReallocate` throws.
```
erights: Reviewers, please look at this pair of TODOs and advise. What I tried to do here is preserve the original semantics, even though they seem inconsistent.

erights:
I'm now shifting to using only the `atomicReallocate` that I happened to see at https://github.com/Agoric/agoric-sdk/pull/6102#discussion_r960760639 . Revisiting the same example again, it would now be

```js
          atomicReallocate(
            zcf,
            harden([
              [brandOutPoolSeat, seat, { Secondary: amountOut }, { Out: amountOut }],
              [brandInPoolSeat, brandOutPoolSeat, { Central: reducedCentralAmount }],
              [seat, brandInPoolSeat, { In: reducedAmountIn }, { Secondary: reducedAmountIn }],
            ]),
          );
```
which IMO dominates all the other experiments. (Thanks @turadg ! )

If we really are reducing it all to `atomicReallocate`, then the current implementation in `delta.js` is stupidly overcomplicated. It got that way only by layering of intermediate experiments we can now cleanup. Actually doing this cleaning is now another step before R4R.

Reviewers, aside from bikeshedding the name, waddaya think of only `atomicReallocate`?

packages/inter-protocol/src/collectFees.js
```diff
@@ -14,9 +16,10 @@ export const makeMakeCollectFeesInvitation = (
 ) => {
   const collectFees = seat => {
     const amount = feeSeat.getAmountAllocated(keyword, feeBrand);
-    feeSeat.decrementBy(harden({ [keyword]: amount }));
-    seat.incrementBy(harden({ Fee: amount }));
-    zcf.reallocate(seat, feeSeat);
+    atomicReallocate(
```
Chris-Hibbert: I think, in the context of other suggestions, `atomicReallocate` should be `atomicTransfer`, since the semantics of each parameter is to decrement from the first of the pair and increment the second.

packages/inter-protocol/src/psm/psm.js
Chris-Hibbert: in giveMinted, fees are extracted from the amounts traded, and we burn fees if the transfer took place. In wantMinted, we unconditionally mint before the transaction, and burn if the transaction fails. The original logic is correct, and you have correctly reproduced it.

Chris-Hibbert (commented):
This seems an acceptable approach. It meets our two must-haves: getting rid of the stateful stagings, and at least as readable as the current form, which lets the contract writer make apparent that transfers maintain conservation.

This variant seems to get pretty verbose (in line count) when there are 2 or more transfers going on. It's still pretty readable, though

packages/inter-protocol/src/collectFees.js
erights: Done

packages/inter-protocol/src/psm/psm.js
erights: Thanks. I have revised the comments to explain this. Done.

erights:
> This variant seems to get pretty verbose (in line count) when there are 2 or more transfers going on. It's still pretty readable, though

Agreed. Usually verbosity is a good first approximation proxy for readability. In this case, the declarative data nature of the new code IMO always more than makes up for its extra line count. I think it is always at least as readable, and sometimes more so.

packages/inter-protocol/src/vaultFactory/vaultManager.js
```diff
@@ -537,12 +538,16 @@ const helperBehavior = {
           // it. We could hold it until it crosses some threshold, then sell it
           // to the AMM, or we could transfer it to the reserve. At least it's
           // visible in the accounting.
-          vaultSeat.decrementBy(
-            state.retainedCollateralSeat.incrementBy({
```
erights, line 541: Note that the original here was written backward: with the incrementBy inside the decrementBy, violating our normal convention. No such confusion in the new code.

erights:
Ready for review!

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
```diff
@@ -57,15 +58,30 @@ export const makeDoublePool = (
...
+          inPoolSeat,
+          { Central: prices.inPoolDecrement },
+          { Secondary: prices.inPoolIncrement },
+        ],
+        [
+          outPoolSeat,
+          outPoolSeat,
+          { Secondary: prices.outPoolDecrement },
+          { Central: prices.outPoolIncrement },
+        ],
```
Chris-Hibbert: I don't like the look of these transfers. I realize (as the comment below explains) that the individual lines don't need to balance. But transferring different amounts in and out of a single seat in one line is confusing.  I'd prefer something like this, though if we can create a shorthand that leaves out the `undefined`s, I'd be even happier.
```
    [seat, undefined, { In: prices.swapperGives }],
    [undefined, seat, undefined, { Out: prices.swapperGets }],
    [inPoolSeat, undefined, { Central: prices.inPoolDecrement }],
    [undefined, inPoolSeat, undefined{ Secondary: prices.inPoolIncrement }],
    [outPoolSeat, undefined{ Secondary: prices.outPoolDecrement }],
    [undefined, outPoolSeat, undefined, { Central: prices.outPoolIncrement }],
    [undefined, feeSeat, undefined, { Fee: prices.protocolFee }]
```

Chris-Hibbert (changes requested):
One more clean-up suggestion.

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
erights: Done, but somewhat differently. PTAL.

packages/zoe/src/contractSupport/atomicTransfer.js
```diff
@@ -0,0 +1,79 @@
+const { details: X, quote: q } = assert;
+
+/**
+ * @typedef {[
+ *   fromSeat?: ZCFSeat,
+ *   toSeat?: ZCFSeat,
+ *   subtrahend?: AmountKeywordRecord,
+ *   addend?: AmountKeywordRecord
+ * ]} TransferArgs
+ */
```
turadg: nit: this type is the concern of `atomicTransfer` so better to be defined there. To have a name for the type of the second arg (I see it's imported elsewhere):

```suggestion
/** @typedef {Parameters<typeof atomicTransfer>[1]} TransferArgs */
```

packages/zoe/src/contractSupport/atomicTransfer.js
```diff
@@ -0,0 +1,79 @@
...
+export const atomicTransfer = (zcf, transfers) => {
+  const uniqueSeatSet = new Set();
+  for (const [
+    fromSeat = undefined,
+    toSeat = undefined,
+    _subtrahend,
+    _addend,
+  ] of transfers) {
+    fromSeat && uniqueSeatSet.add(fromSeat);
```
turadg: why not `if`?
```suggestion
    if (fromSeat) {
      uniqueSeatSet.add(fromSeat);
    }
```

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
```diff
@@ -57,15 +58,48 @@ export const makeDoublePool = (
...
-    inPoolSeat.decrementBy(harden({ Central: prices.inPoolDecrement }));
-    outPoolSeat.decrementBy(harden({ Secondary: prices.outPoolDecrement }));
-    seat.incrementBy(harden({ Out: prices.swapperGets }));
-    inPoolSeat.incrementBy(harden({ Secondary: prices.inPoolIncrement }));
-    outPoolSeat.incrementBy(harden({ Central: prices.outPoolIncrement }));
-    feeSeat.incrementBy(harden({ Fee: prices.protocolFee }));
+    atomicTransfer(
+      zcf,
+      harden([
+        // In the typical use of `atomicTransfer`, we use each like to
```
Chris-Hibbert:
```suggestion
        // In the typical use of `atomicTransfer`, we use each line to
```

packages/inter-protocol/src/vpool-xyk-amm/singlePool.js
```diff
@@ -17,20 +18,37 @@ export const makeSinglePool = ammPowers => ({
...
+
+    atomicTransfer(
+      zcf,
+      harden([
+        [seat, seat, { In: prices.swapperGives }, { Out: prices.swapperGets }],
+        // This strange construction is to transfer into without saying at the
+        // same time where this amount was transfered from. But the overall
+        // atomicTransfer still has to be conserved, i.e., balance to
+        // a net zero.
+        [undefined, protocolSeat, undefined, { Fee: prices.protocolFee }],
```
Chris-Hibbert: This is another case where the amounts transferred are different. 
```suggestion
        // This doesn't follow the typical patter of `atomicTransfer`, in which each line
        //  transfers a single amount from one seat to another. `Prices` was previously
        // calculated as a balanced set of transfers. `atomicTransfer()` ensures that the 
        // entire transaction is in balance. 
        [seat, undefined, { In: prices.swapperGives }],
        [undefined, seat, undefined, { Out: prices.swapperGets }],
        [undefined, protocolSeat, undefined, { Fee: prices.protocolFee }],
```

packages/inter-protocol/src/vpool-xyk-amm/pool.js
```diff
@@ -252,18 +253,21 @@ export const definePoolKind = (baggage, ammPowers, storageNode, marshaller) => {
...
+      atomicTransfer(
+        ammPowers.zcf,
+        harden([
+          [userSeat, poolSeat, { Liquidity: liquidityIn }],
+          [
+            poolSeat,
+            userSeat,
+            {
+              Central: centralTokenAmountOut,
+              Secondary: tokenKeywordAmountOut,
```
Chris-Hibbert, line 265: amounts are different. I'd split into two lines.

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -51,13 +52,13 @@ export const satisfies = (zcf, seat, update) => {
...
-
-    leftSeat.decrementBy(harden(rightSeat.getProposal().want));
-    rightSeat.incrementBy(harden(rightSeat.getProposal().want));
-
-    zcf.reallocate(leftSeat, rightSeat);
+    atomicTransfer(
+      zcf,
+      harden([
+        [rightSeat, leftSeat, leftSeat.getProposal().want],
+        [leftSeat, rightSeat, rightSeat.getProposal().want],
```
Chris-Hibbert, line 64: These are different amounts and keywords

packages/zoe/src/contractSupport/zoeHelpers.js
```diff
@@ -72,13 +73,23 @@ export const swap = (zcf, leftSeat, rightSeat) => {
...
+          rightSeat,
+          leftSeat,
+          rightSeat.getProposal().give,
+          leftSeat.getProposal().want,
+        ],
+        [
+          leftSeat,
+          rightSeat,
+          leftSeat.getProposal().give,
+          rightSeat.getProposal().want,
```
Chris-Hibbert: ditto

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
Chris-Hibbert: This is not the only case in which the individual lines don't balance. The situation where `[seat1, seat2, keyedAmount1, keyedAmount2]` makes sense to me is when the amounts are the same and the keys differ. If the amounts are different I'd prefer to see it as two distinct lines, as you did with `doublePool`. 

I was hoping for more general support for the other patterns, since there's so much noise and redundancy here. The distinct cases I see are
```
atomicSwap(zcf, {
   xfer: [seat1, seat2, { key: amount }],
   xfer: [seat1, seat2, key1, key2, amount],
   from: [seat3, key3, amount3],
   to: [seat4, key4, amount4],
});
```

The labels are clearly wrong, but this seems like it could be significantly more concise. The first two vastly predominate, which is all the more reason to make the second two visually distinct. I don't have a concrete proposal, just a suggested direction.

You're right that making this change isn't necessary. I'll point out other cases where the model doesn't seem a good fit to me.

turadg: I agree with @Chris-Hibbert this could use more tweaking before landing.

Some things to try to satisfy:
- easy to re-use a {keyword: amount} object because they're often already defined
- not repeating the keyword or amount unnecessarily
- not having to bracket the array in the single transfer case
- not having `undefined` in the arguments


**single transfer on the same key**
```js
atomicTransfer(zcf, fromSeat, toSeat, { keyword: amount });
```

For other cases, @Chris-Hibbert and @turadg propose having separate functions.

**single transfer to a different key**
```js
// in this PR
atomicTransfer(zcf, fromSeat, toSeat, { sourceKeyword: amount }, { targetKeyword: amount });
// dedupe amount with same function
atomicTransfer(zcf, fromSeat, toSeat, { sourceKeyword: amount }, targetKeyword);
// new function for new signature
atomicTransferRename(zcf, fromSeat, toSeat, sourceKeyword, targetKeyword, amount );
```

**rearrangement**
codify the pattern of splitting "from" and "to", removing the need for `undefined` 
```js
atomicRearrangement(zcf, {
  from: [
    [seat, { In: prices.swapperGives }],
    [inPoolSeat, { Central: prices.inPoolDecrement }],
    [outPoolSeat, { Secondary: prices.outPoolDecrement }],
  ],
  to: [
    [seat, { Out: prices.swapperGets }],
    [inPoolSeat, { Secondary: prices.inPoolIncrement }],
    [outPoolSeat, { Central: prices.outPoolIncrement }],
    [feeSeat, { Fee: prices.protocolFee }],
  ],
});

```

dckc (commented):
I see lots of stateful stuff that always made me uneasy replaced by single atomic calls. :)

I only browsed parts of it, but all the stuff I saw looked good.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: I don't understand. AFAICT, the new code should be exactly equivalent to the old. What am I missing?

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
erights: These are great suggestions. I took big steps in these directions, but somewhat differently. PTAL and let me know what you think.

- The thing that used to be called `atomicTransfer` I renamed `atomicRearrange`.
- I then used `atomicTransfer` as you suggest above, for the special case of a single oneway transfer. I was surprised that this was the majority of cases: 22 `atomicTransfer(` vs 18 remaining `atomicRearrange`.
- For expressing from-only portions separately from to-only portions, I created to helpers `fromOnly` and `toOnly`. Both are mostly used only in `doublePool.js`. See if you're pleased by how it looks there.

packages/zoe/src/contractSupport/atomicTransfer.js
erights: This seems less clear to me.

packages/zoe/src/contractSupport/atomicTransfer.js
erights: Huh. I have no idea why I did that. In any case, done.

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
erights: Done

packages/zoe/src/contractSupport/zoeHelpers.js
erights: I now understand the similar issue you've raised elsewhere. However, I don't think it applies here.

Each line only has one AmountsKeywordRecord, and thus serves as both the fromAmounts and twoAmounts for the oneway transfer represented by that line alone. So each line is balanced by itself. Like the original, this is doing a pair of one-way transfers.

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
erights: I now understand @Chris-Hibbert 's criticisms about using one TransferArgs line with separate fromAmounts and toAmounts where the amounts are different. I think I took care of all of those, turning each into separate fromOnly and toOnly lines.

This though does suggest and error check: atomicTransfer could enforce that when a single TransferArgs line has both fromAmounts and toAmounts, that they are have the same total amounts. This ensures that each line either
   * `fromOnly` - takes only from one side
   * `toOnly` - gives only to one side
   * makes a balanced oneway transfer from one seat to another, obeying conservation by itself

Does this error check seem like a good idea? With this error check in place, what would you think of the overall API?

erights: I do want to avoid notation like `[seat1, seat2, key1, key2, amount],` because an AmountKeywordRecord can have multiple `keyword: amount` properties. It should be taken as an aggregate of all those amounts. And sometimes what we have coming in is the AmountKeywordRecord, which we should not need to take apart.

With the API in the current PR, there is still a bit of repetition left. Without the error check above (which I have not yet implemented), I can see the hazard. But with that error check, I think the repetition is harmless and actually makes things clearer. WDYT?

packages/zoe/src/contractSupport/atomicTransfer.js
erights: Btw, I also got rid of the awful names `subtrahend` and `addend`, replacing with `fromAmounts` and `toAmounts`.

packages/inter-protocol/src/vpool-xyk-amm/singlePool.js
erights: Gotcha. Now rewritten with `fromOnly` and `toOnly`. Resolving.

packages/inter-protocol/src/vpool-xyk-amm/pool.js
erights: This isn't a balance issue. Each TransferArgs line is only doing an internally balanced oneway transfer. We're just aggregating multiple specific amounts into an AmountKeywordRecord, which is what it's for. So we can consider these together as a heterogeneous group of assets.

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Yes! This one needed to split up for the reason that I initially missed. The old TransferArgs lines before were not individually balanced.

Done.

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
Chris-Hibbert: > an AmountKeywordRecord can have multiple keyword: amount properties. It should be taken as an aggregate of all those amounts. And sometimes what we have coming in is the AmountKeywordRecord, which we should not need to take apart.

That's a good argument. I expect the error check to be sufficient. The repetition I was trying to avoid is an annoyance, but in the face of existing AKRs is acceptable.

Looking...

packages/zoe/src/contracts/loan/close.js
```diff
@@ -39,20 +42,22 @@ export const makeCloseLoanInvitation = (zcf, config) => {
...
+          repaySeat,
+          {
+            Collateral: collateralSeat.getAmountAllocated(
+              'Collateral',
+              collateralBrand,
+            ),
+          },
+        ],
+        [repaySeat, lenderSeat, { Loan: debt }],
+      ]),
```
Chris-Hibbert: Would you mind pulling `collateralSeat.getAmountAllocated()` out? I think it makes the `atomicRearrange` much more readable.

```suggestion
    const collateralAmount = collateralSeat.getAmountAllocated(
      'Collateral',
      collateralBrand,
    );
    atomicRearrange(
      zcf,
      harden([
        [collateralSeat, repaySeat, { Collateral: collateralAmount }],
        [repaySeat, lenderSeat, { Loan: debt }],
      ]),
    );
```

packages/internal/src/utils.js
```diff
@@ -381,3 +381,37 @@ export const fsStreamReady = stream =>
...
+ * @typedef {object} ListAccumulator
+ * @property {(value: T) => void} push
+ * @property {() => void} clear
+ * @property {() => T[]} snapshot
+ * @property {() => T[]} takeSnapshot
+ */
+
+/**
+ * @template {any} T
+ * @param {(x: T) => T} coerce
```
turadg: can't `x` be anything and the function will return `T` or throw?
```suggestion
 * @param {(x: unknown) => T} coerce
```

packages/internal/src/utils.js
```diff
@@ -381,3 +381,37 @@ export const fsStreamReady = stream =>
     stream.on('ready', onReady);
     stream.on('error', onError);
   });
+
+/**
+ * @template T
+ * @typedef {object} ListAccumulator
+ * @property {(value: T) => void} push
```
turadg: if you're using a coerce function, can't you push non-T values?

packages/zoe/src/contractSupport/atomicTransfer.js
```diff
@@ -0,0 +1,132 @@
+const { Fail, quote: q } = assert;
+
+/**
+ * @typedef {[
+ *   fromSeat?: ZCFSeat,
+ *   toSeat?: ZCFSeat,
+ *   fromAmounts?: AmountKeywordRecord,
+ *   toAmounts?: AmountKeywordRecord
+ * ]} TransferArgs
```
turadg: it's the array of these that are the function args. consider `TransferArg` (singular), `TransferLine` or `TransferPart`

packages/zoe/src/contractSupport/atomicTransfer.js
```diff
@@ -0,0 +1,132 @@
...
+ * @param {ZCF} zcf
+ * @param {TransferArgs[]} transfers
+ */
+export const atomicRearrange = (zcf, transfers) => {
+  const uniqueSeatSet = new Set();
+  for (const [
+    fromSeat = undefined,
+    toSeat = undefined,
+    _subtrahend,
+    _addend,
```
turadg: awkward names. why destructure them at all? the type makes clear what's in the tuple

turadg (commented):
The new API LGTM. Holding approve pending some clarification on supporting utilities, but not blocking.

packages/zoe/src/contractSupport/atomicTransfer.js
```diff
@@ -0,0 +1,132 @@
...
+  ] of transfers) {
+    if (fromSeat) {
+      uniqueSeatSet.add(fromSeat);
+    }
+    if (toSeat) {
```
Chris-Hibbert: Assert that at least one of `fromSeat` adn `toSeat` is present.  I think this works naturally with an `if-else` reframing.

packages/zoe/src/contractSupport/atomicTransfer.js
Chris-Hibbert: If you're going to use `fromAmounts` and `toAmounts` below, this should too.

Chris-Hibbert (approved):
minor suggestions in `atomicRearrange`.  The PR LGTM.

packages/internal/src/utils.js
erights: turns out makeListAccumulator is vestigial from when I was using it within (what is now called) `atomicRearrange`. I no longer am, and neither is anything else, so I just deleted it from the PR.

Gone

packages/internal/src/utils.js
erights: Gone

packages/zoe/src/contractSupport/atomicTransfer.js
erights: I took your `TransferPart` suggestion.

Done

packages/zoe/src/contractSupport/atomicTransfer.js
erights: I don't understand this comment. There are four distinct things but only two types. 

In any case, I find generally find it clearer to destructure the distinct things I need into separate variables.  This might just be a bad habit taught by the weird things we need to do to be defensive in JavaScript, where a property access can have side effects and deliver different values different times. Regarding just the instability, a nice defensive practice is "If you don't want inconsistent answers, ask each question only once."

We designed our Passable types so they would not create such hazards. We're not there yet (need ref to reentrancy vulnerability, cc @mhofman ), but we plan to be, and are programming under the assumption we succeed. So, under this assumption, this defensive practice is not needed for Passables. As of this PR, this is not yet a Passable-only boundary, but it will be by the time we need it to be defensive. So my practice is overkill. But I just like it better anyway.

Fixed the names but still destructuring. Resolving.

erights: Done

packages/zoe/src/contracts/loan/close.js
erights: Much better! Done.

packages/inter-protocol/src/vpool-xyk-amm/doublePool.js
erights: Per-transferPart error check now implemented. In CI now, so we'll see if any current contracts violate it.

Done.

packages/zoe/src/contractSupport/atomicTransfer.js
erights: Resolving

packages/inter-protocol/src/vpool-xyk-amm/pool.js
erights: Resolving

packages/zoe/src/contractSupport/zoeHelpers.js
erights: Resolving

packages/zoe/src/contractSupport/atomicTransfer.js
erights: Thanks! Starting on that suggestion led to a significant improvement in the code, where I validate earlier. It also gave a natural place to do the per-transferPart conservation check.

Done

packages/zoe/src/contractFacet/types.js
```diff
@@ -22,7 +22,8 @@
  * synchronously from within the contract, and usually is referred to
  * in code as zcf.
  *
- * @property {Reallocate} reallocate - reallocate amounts among seats
+ * @property {Reallocate} reallocate - reallocate amounts among seats.
+ * Deprecated: Use atomicRearrange instead.
```
dckc, line 26: [@deprecated](https://jsdoc.app/tags-deprecated.html) would get IDEs such as vs-code to highlight uses with strikethru or the like.

(sorry I didn't notice this earlier)

turadg: but putting it here would tag the `ZCF` type.

PR #6582, opened 2022-11-17
types for `this` in far classes

turadg:
## Description

Far classes put `state` and `facets` on `this` but we never got the types there. This was allowed by our TS config because `this` was implicitly `any`. 

This PR makes `this` inferred in behaviors passed to the Far class makers. 

It also enables `noImplicitThis` across the repo to catch such type coverage loss in the future (and because the `ThisType` utility that enables the above requires it.)

### Security Considerations

--

--

### Testing Considerations

CI

### DIFF
--- a/packages/agoric-cli/src/commands/oracle.js
+++ b/packages/agoric-cli/src/commands/oracle.js
@@ -69,6 +69,7 @@ export const makeOracleCommand = async logger => {
     )
     .option('--offerId [number]', 'Offer id', Number, Date.now())
     .action(async function () {
+      // @ts-expect-error this implicit any
       const opts = this.opts();
 
       const instance = lookupPriceAggregatorInstance(opts.pair);
@@ -104,6 +105,7 @@ export const makeOracleCommand = async logger => {
     )
     .requiredOption('--price [number]', 'price (format TODO)', String)
     .action(async function () {
+      // @ts-expect-error this implicit any
       const opts = this.opts();
 
       /** @type {import('../lib/psm.js').OfferSpec} */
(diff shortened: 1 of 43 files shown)


packages/smart-wallet/src/smartWallet.js
```diff
@@ -465,29 +458,33 @@ const behavior = {
...
+      return this.state.updatePublishKit.subscriber;
     },
   },
 };
 /** @typedef {typeof behavior} SmartWalletBehavior */
+// XXX defineVirtualFarClassKit should infer this
+/** @typedef {{ facets: SmartWalletBehavior, state: State }} SmartWalletThis */
 
+// TODO type facets (runs into problems with 'this')
+/**
```
mhofman, line 484: Wondering if this is all due to the behavior, init, finish etc being defined ahead of time instead of inline during the `defineVirtualFarClassKit`.

packages/agoric-cli/src/commands/oracle.js
```diff
@@ -69,6 +69,7 @@ export const makeOracleCommand = async logger => {
     )
     .option('--offerId [number]', 'Offer id', Number, Date.now())
     .action(async function () {
+      // @ts-expect-error this implicit any
```
mhofman, line 72: Wondering if patching `commander` to type their API correctly would be sufficient? 

https://github.com/tj/commander.js/blob/1760b428553e2b84f159b1a3fdf09f0dce0aeec3/typings/index.d.ts#L494

```ts
  action(fn: (this: this, ...args: any[]) => void | Promise<void>): this;
```

mhofman (approved):
I have to admit I didn't fully grok how the interface tools typings work, but I trust they do given the tests and usages.

Couple suggestions to figure out what the cause of some of the explicit `this` typings or error suppressions may be, but feel free to leave it and merge as is.

packages/smart-wallet/src/smartWallet.js
turadg: Aha, yeah must be

PR #6587, opened 2022-11-17
6450 edits

warner:
some recommended fixes I thought of while reviewing #6450

### DIFF
--- a/packages/SwingSet/src/vats/comms/controller.js
+++ b/packages/SwingSet/src/vats/comms/controller.js
@@ -34,9 +34,9 @@ export function deliverToController(
 
     const name = args[0];
     const transmitterID = krefOf(args[1]);
-    assert(transmitterID, 'unexpected "transmitter" arg for addRemote()');
+    assert(transmitterID, 'bad "transmitter" arg for addRemote()');
     const setReceiverID = krefOf(args[2]);
-    assert(setReceiverID, 'unexpected "setReceiver" arg for addRemote()');
+    assert(setReceiverID, 'bad "setReceiver" arg for addRemote()');
 
     const { receiverID } = state.addRemote(name, transmitterID);
 
@@ -56,7 +56,7 @@ export function deliverToController(
     assert(remoteID, X`unknown remote name ${remoteName}`);
     const remoteRefID = args[1];
     const kernelRefID = krefOf(args[2]);
-    assert(kernelRefID, 'unexpected "obj0" arg for addEgress()');
+    assert(kernelRefID, 'bad "obj0" arg for addEgress()');
     const localRef = provideLocalForKernel(kernelRefID);
     addEgress(remoteID, remoteRefID, localRef);
     syscall.resolve([[result, false, kser(undefined)]]);
--- a/packages/SwingSet/src/vats/comms/controller.js
+++ b/packages/SwingSet/src/vats/comms/controller.js
@@ -34,7 +34,9 @@ export function deliverToController(
 
     const name = args[0];
     const transmitterID = krefOf(args[1]);
+    assert(transmitterID, 'unexpected "transmitter" arg for addRemote()');
     const setReceiverID = krefOf(args[2]);
+    assert(setReceiverID, 'unexpected "setReceiver" arg for addRemote()');
 
     const { receiverID } = state.addRemote(name, transmitterID);
 
@@ -53,7 +55,9 @@ export function deliverToController(
     const remoteID = state.getRemoteIDForName(remoteName);
     assert(remoteID, X`unknown remote name ${remoteName}`);
     const remoteRefID = args[1];
-    const localRef = provideLocalForKernel(krefOf(args[2]));
+    const kernelRefID = krefOf(args[2]);
+    assert(kernelRefID, 'unexpected "obj0" arg for addEgress()');
+    const localRef = provideLocalForKernel(kernelRefID);
     addEgress(remoteID, remoteRefID, localRef);
     syscall.resolve([[result, false, kser(undefined)]]);
   }
@@ -66,6 +70,9 @@ export function deliverToController(
     assert(remoteID, X`unknown remote name ${remoteName}`);
     const remoteRefID = Nat(args[1]);
     const iface = args[2];
+    if (iface) {
+      assert.typeof(iface, 'string', 'unexpected iface type in addIngress()');
+    }
     const localRef = addIngress(remoteID, remoteRefID);
     const data = kser(kslot(provideKernelForLocal(localRef), iface));
     syscall.resolve([[result, false, data]]);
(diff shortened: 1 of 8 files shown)


packages/SwingSet/src/vats/comms/controller.js
```diff
@@ -34,7 +34,9 @@ export function deliverToController(
 
     const name = args[0];
     const transmitterID = krefOf(args[1]);
+    assert(transmitterID, 'unexpected "transmitter" arg for addRemote()');
     const setReceiverID = krefOf(args[2]);
+    assert(setReceiverID, 'unexpected "setReceiver" arg for addRemote()');
```
FUDCo: Not sure I understand the error text.  Did you mean "missing" instead of "unexpected"?  We _are_ expecting a transmitter arg, after all.  That's what this assert is checking for.

packages/SwingSet/src/vats/comms/controller.js
```diff
@@ -53,7 +55,9 @@ export function deliverToController(
     const remoteID = state.getRemoteIDForName(remoteName);
     assert(remoteID, X`unknown remote name ${remoteName}`);
     const remoteRefID = args[1];
-    const localRef = provideLocalForKernel(krefOf(args[2]));
+    const kernelRefID = krefOf(args[2]);
+    assert(kernelRefID, 'unexpected "obj0" arg for addEgress()');
```
FUDCo: Echo the above comment.  I don't understand "unexpected".  We are expecting the arg.  I think you want a different word here but I'm not sure what it is.

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -522,7 +531,7 @@ async function testMultiKindUpgradeChecks(t, mode, complaint) {
   const [v2status, v2result] = await run('upgradeV2Simple', [mode]);
   if (complaint) {
     t.is(v2status, 'rejected');
-    const e = kunser(kunser(v2result));
+    const e = kunser(v2result);
```
FUDCo, line 536: Did you actually run this updated test?  I'm pretty sure the double deserialization is necessary here because the error is itself a serialized payload.

packages/SwingSet/test/upgrade/test-upgrade.js
```diff
@@ -649,7 +658,7 @@ test('failed vatAdmin upgrade - bad replacement code', async t => {
   const kpid = c.upgradeStaticVat('vatAdmin', true, bundleID, {});
   await c.run();
   const vaUpgradeStatus = c.kpStatus(kpid);
-  const vaUpgradeResult = kunser(kunser(c.kpResolution(kpid)));
+  const vaUpgradeResult = kunser(c.kpResolution(kpid));
```
FUDCo, line 663: Ditto.

FUDCo (changes requested):
Mostly these changes seem like improvements, but I have questions about some error message text and I think a couple of your test tweaks are mistaken.

packages/SwingSet/src/vats/comms/controller.js
warner: Oh, my first commits were made thinking that `krefOf` would return `null` for non-object-references instead of throwing. In that case, we need to assert that `setReceiverID` is non-null tor make sure it wasn't a non-object-reference, like `setReceiverID = 4`, or `setReceiverID = { foo: 'bar' }` (or indeed `undefined`, in which case the arg was probably missing).

After making that commit, I got bolder and added a later commit which makes `krefOf` throw. In *that* case, the `assert` is redundant, but harmless.

By "unexpected" I really should have said "unexpected type of argument", but that got too long :). How about I change it to `bad "setReceiver" arg`? That's probably less ambiguous.

packages/SwingSet/src/vats/comms/controller.js
warner: I'll say "bad" here too

packages/SwingSet/test/upgrade/test-upgrade.js
warner: Yep, and both CI and local tests pass. The matching change is in kernel.js line 835 to avoid double-*serializing* the error. This:
```js
    function makeFailure(_errorCD) {
      insistCapData(_errorCD); // kser(Error)
      // const error = kunser(_errorCD)
      // actually we shouldn't reveal the details, so instead we do:
      const error = Error('vat-upgrade failure');
      return ['vatUpgradeCallback', [upgradeID, false, error]];
    }
```

used to be:
```js
      const error = kser(Error('vat-upgrade failure'));
```

but if you trace through how the return value of `makeFailure` is used, it gets stashed in `vatAdminMethargs`, which eventually gets passed into `queueToKref()`, which (thanks to your excellent change) now performs it's own serialization (i.e. `queueToKref` now takes ersatz Presences, not capdata):

(kernel.js line 1267)
```js
    if (crankResults.vatAdminMethargs) {
      // we use terminateVat() to notify vat-admin about failed vat
      // creation, but vatAdminMethargs for successful vat creation,
      // and failed/successful vat upgrades.
      const [method, args] = crankResults.vatAdminMethargs;
      queueToKref(vatAdminRootKref, method, args, 'logFailure');
    }
```

The double-serialization was a bug: when userspace received the rejection value of `upgradeVat()`, it would get capdata of an Error, when it should have gotten an Error.

I wish typescript would help here, but the `args` input to `queueToKref()` is now quite liberal (whereas previously it wanted capdata), which is more ergonomic to use, but harder to check a type on.

The test cases (the ones I modified to change `kunser(kunser(results))` into just `kunser(results)`) don't make that clear. They tell userspace (`tests/upgrade/bootstrap-upgrade.js`) to run `upgradeV2WhichExplodes()` and return an array of events:

```js
    upgradeV2WhichExplodes: async () => {
      const bcap = await E(vatAdmin).getNamedBundleCap('ulrik2');
      const events = [];
      const handler = Far('handler', { ping: msg => events.push(msg) });
      const explode = 'kaboom';
      const vatParameters = { youAre: 'v2', marker, handler, explode };
      // vp.handler causes v2 to handler~.ping(), but that gets unwound
      const p = E(ulrikAdmin).upgrade(bcap, { vatParameters }); // throws
      await p.catch(e => events.push(e));
      await E(ulrikRoot).pingback(handler); // goes to post-rewind v1
      return events;
    },
```

and then `test-upgrade.js` inspects the (serialized) `events` that it returns, one of which is expected to be the Error. `test-upgrade.js` is not inside the vat, and the `c.kpResolution()` it uses gives back capdata, so it must do a `kunser()` to get something equivalent to what the userspace/vat would get. That kinda concealed the fact that a double `kunser` is wrong.

This would be more obvious if the userspace code (in `bootstrap-upgrade.js`) checked `e` for being an `Error` directly (e.g. if it does a check and then includes a boolean in the `events` array with that claim). That would give us an assertion that doesn't have a `kunser` in the pathway. To make *that* test pass with the incorrect double-serialization, you would have needed to add a `kunser()` into the *vat* code, and *that* would have been obviously wrong.

I'll push another commit to change the test that way, which should help make things less ambigous.

warner:
New changes pushed, see if that clears things up.

PR #6589, opened 2022-11-18
export types from notifier package

turadg:
## Description

wallet-app needs type exports from `@agoric/notifier` but the types are only available as `types-ambient`. In other packages we have made `src/types.js` the primary source and made `src/types-ambient.js` define the same types in terms of the primary.

A problem with this is that the JSDoc from the primary is not there on the secondary, so any uses of the type from global scope (ambiently) don't have JSDoc documentation.

This takes another approach, described in https://github.com/Agoric/agoric-sdk/blob/ta/types-notifier/docs/typescript.md

### Security Considerations

--

Maintains backwards compatibility

### Testing Considerations

CI


michaelfig:
> copy `types-ambient.js` to `types.js` and put an `export {}` line in the latter.

I prefer this approach where possible, but if you see advantages in using the other approach, I can be swayed.

turadg:
> I prefer this approach where possible, but if you see advantages in using the other approach, I can be swayed.

One advantage was it was a cleaner postpack, but minor enough not to go back and forth more. The PR has the "add export" approach now and I applied it in `@agoric/ertp` too.

docs/typescript.md
```diff
@@ -0,0 +1,14 @@
+# usage of TypeScript
+
+Our use of TypeScript has to accomodate both .js development in agoric-sdk (which cannot import types) and .ts development of consumers of agoric-sdk packages (which can import types). For .js development, we want ambient (global) types so that we don't have to proceed each type reference by an import. For .ts development, we want exports from modules so we don't pollute a global namespace.
```
michaelfig:
```suggestion
Our use of TypeScript has to accomodate both .js development in agoric-sdk (which cannot import types) and .ts development of consumers of agoric-sdk packages (which can import types). For .js development, we want ambient (global) types so that we don't have to precede each type reference by an import. For .ts development, we want exports from modules so we don't pollute a global namespace.
```

docs/typescript.md
```diff
@@ -0,0 +1,14 @@
...
+
+This means we need two definition files. We could keep those both in SCM, but that would cause more confusion and risk them getting out of sync. We could have one be defined in terms of the other, but since there is no way to reexport to global namespace, it would have to be a secondary definition in the primary. This would typecheck because TS is structurally typed, but the secondary would lose all the documentation on the types.
+
+So our best solution is to have one source of truth for the types and auto-generate for one case from the other. We've chosen to have the ambient types as the source of truth so the SDK development can use them. The SDK consumption can have a build step during packaging, and that's when we make the exported (non-ambient) types.
+## Best practices
+
+- `src/types-ambient.js` defines types of the package
+- `src/types.d.ts` is built automatically from types-ambient
+  - `prepack` copies types-ambient.js to types.js and appends 'export {};' to turn it into a module, then builesd
```
michaelfig:
```suggestion
  - `prepack` copies types-ambient.js to types.js and appends 'export {};' to turn it into a module, then builds
```

michaelfig (approved):
LGTM!

PR #6590, opened 2022-11-18
types improvements for Far classes

turadg:
## Description

Improves type inference for far classes and uses it in smartWallet.js to reduce boilerplate in expressing types.

### Security Considerations

--

--

### Testing Considerations

CI

### DIFF
--- a/packages/import-manager/jsconfig.json
+++ b/packages/import-manager/jsconfig.json
@@ -13,6 +13,7 @@
   */
       "downlevelIteration": true,
       "strictNullChecks": true,
+      "noImplicitThis": true,
       "moduleResolution": "node",
     },
     "include": [
--- a/packages/smart-wallet/src/smartWallet.js
+++ b/packages/smart-wallet/src/smartWallet.js
@@ -76,10 +76,10 @@ const mapToRecord = map => Object.fromEntries(map.entries());
 
 /**
  * @typedef {{
- * brand: Brand,
- * displayInfo: DisplayInfo,
- * issuer: ERef<Issuer>,
- * petname: import('./types').Petname
+ *   brand: Brand,
+ *   displayInfo: DisplayInfo,
+ *   issuer: ERef<Issuer>,
+ *   petname: import('./types').Petname
  * }} BrandDescriptor
  * For use by clients to describe brands to users. Includes `displayInfo` to save a remote call.
  */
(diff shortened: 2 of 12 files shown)


packages/smart-wallet/src/smartWallet.js
```diff
@@ -88,13 +88,28 @@ const mapToRecord = map => Object.fromEntries(map.entries());
...
+ * zoe: ERef<ZoeService>,
+ * }} SharedParams
+ *
  * @typedef {ImmutableState & MutableState} State
  * - `brandPurses` is precious and closely held. defined as late as possible to reduce its scope.
  * - `offerToInvitationMakers` is precious and closely held.
  * - `brandDescriptors` will be precious. Currently it includes invitation brand and  what we've received from the bank manager.
  * - `purseBalances` is a cache of what we've received from purses. Held so we can publish all balances on change.
  *
- * @typedef {Parameters<initState>[0] & Parameters<initState>[1]} HeldParams
```
turadg, line 97: basing the type of the params was causing a circular reference when I made `initState` to return `State`.

packages/smart-wallet/src/smartWallet.js
```diff
@@ -228,301 +235,288 @@ const behaviorGuards = {
   }),
 };
 
-// TOOD a utility type that ensures no behavior is defined that doesn't have a guard
-const behavior = {
```
turadg, line 232: this is just a move, no edits besides types

packages/smart-wallet/src/smartWallet.js
```diff
@@ -228,301 +235,288 @@ const behaviorGuards = {
...
-          facets.helper.publishCurrentState();
-        },
-      });
-      executor.executeOffer(offerSpec);
+const SmartWalletKit = defineVirtualFarClassKit(
+  'SmartWallet',
+  behaviorGuards,
+  initState,
+  {
+    helper: {
```
turadg, line 243: somewhat regrettable to have this nested now, but it lets `this` be inferred automatically. There's not really any way to get that automated without making it a param in a function call with the state init function.

erights:
Recommendation for reviewers: remember to hide whitespace before looking at diffs. Helped even more than usual.

packages/smart-wallet/src/smartWallet.js
```diff
@@ -88,13 +88,28 @@ const mapToRecord = map => Object.fromEntries(map.entries());
...
+ * }} UniqueParams
+ *
+ * @typedef {{
+ * agoricNames: ERef<import('@agoric/vats').NameHub>,
+ * invitationIssuer: ERef<Issuer<'set'>>,
+ * invitationBrand: Brand<'set'>,
+ * publicMarshaller: Marshaller,
+ * storageNode: ERef<StorageNode>,
+ * zoe: ERef<ZoeService>,
+ * }} SharedParams
```
erights: Inside comments, we're on our own for whitespace normalization.

```suggestion
 * @typedef {{
 *   address: string,
 *   bank: ERef<import('@agoric/vats/src/vat-bank').Bank>,
 *   invitationPurse: Purse<'set'>,
 * }} UniqueParams
 *
 * @typedef {{
 *   agoricNames: ERef<import('@agoric/vats').NameHub>,
 *   invitationIssuer: ERef<Issuer<'set'>>,
 *   invitationBrand: Brand<'set'>,
 *   publicMarshaller: Marshaller,
 *   storageNode: ERef<StorageNode>,
 *   zoe: ERef<ZoeService>,
 * }} SharedParams
```

packages/smart-wallet/src/smartWallet.js
```diff
@@ -228,301 +235,288 @@ const behaviorGuards = {
...
+          },
+        });
+
+        updatePublishKit.publisher.publish({ updated: 'brand', descriptor });
+
+        // deposit queued payments
+        const payments = paymentQueues.has(desc.brand)
+          ? paymentQueues.get(desc.brand)
+          : [];
+        // @ts-expect-error xxx with DataOnly / FarRef types
```
erights, line 340: "xxx"?

erights (commented):
review not yet done. In the meantime...

(As usual with static types, I'm somewhat out of my depth and still struggling.)

packages/smart-wallet/src/smartWallet.js
turadg: Acknowledgement of tech debt

https://github.com/Agoric/agoric-sdk/wiki/Coding-Style#code-comments

packages/smart-wallet/src/smartWallet.js
turadg: IMO we should allow this non-indented style until we can automate it. I'll make the change though.

packages/import-manager/jsconfig.json
```diff
@@ -13,6 +13,7 @@
   */
       "downlevelIteration": true,
       "strictNullChecks": true,
+      "noImplicitThis": true,
```
dckc, line 16: > Raise error on ‘this’ expressions with an implied ‘any’ type. -- [noImplicitThis docs](https://www.typescriptlang.org/tsconfig#noImplicitThis)

sounds good to me.

packages/vat-data/src/far-class-utils.js
```diff
@@ -108,7 +108,7 @@ harden(defineDurableFarClass);
  * @param {any} interfaceGuardKit
  * @param {(...args: A[]) => S} init
  * @param {T & { [K in keyof T]: ThisType<T[K] & { facets: T, state: S }> }} facets
- * @param {DefineKindOptions<unknown>} [options]
+ * @param {DefineKindOptions<{ facets: T, state: S }>} [options]
```
dckc: this seems to be the crux of the change

packages/vat-data/src/far-class-utils.js
```diff
@@ -195,7 +195,7 @@ harden(vivifyFarClassKit);
  * @param {string} kindName
  * @param {any} interfaceGuard
  * @param {M} methods
- * @param {DefineKindOptions<unknown>} [options]
+ * @param {DefineKindOptions<{ self: M }>} [options]
```
dckc, line 198: TIL: `self` in the swingset sense

packages/vat-data/src/far-class-utils.js
```diff
@@ -23,8 +23,8 @@ import {
  * @param {string} tag
  * @param {any} interfaceGuard
  * @param {(...args: A[]) => S} init
- * @param {T} methods
- * @param {DefineKindOptions<unknown>} [options]
+ * @param {T & ThisType<T & { state: S }>} methods
```
dckc: discussion makes us wonder if this is quite on target

dckc (approved):
looks pretty good.

PR #6597, opened 2022-11-24
feat(store): M.splitArray and M.splitRecord

erights:
As suggested at https://github.com/Agoric/agoric-sdk/pull/6431#pullrequestreview-1174164850 , this PR is extracted from #6431 just to provide the new feature. Because it is compat, this PR does not itself refactor the usage to use the feature.

So that we only bother to change most of these error messages once, this PR also borrows some code from https://github.com/Agoric/agoric-sdk/pull/6432 for generating somewhat better errors.

### DIFF
--- a/packages/store/src/patterns/patternMatchers.js
+++ b/packages/store/src/patterns/patternMatchers.js
@@ -8,7 +8,11 @@ import {
   nameForPassableSymbol,
 } from '@endo/marshal';
 import { identChecker } from '@agoric/assert';
-import { applyLabelingError, listDifference } from '@agoric/internal';
+import {
+  applyLabelingError,
+  fromUniqueEntries,
+  listDifference,
+} from '@agoric/internal';
 
 import {
   compareRank,
@@ -630,8 +634,8 @@ const makePatternKit = () => {
         // const [leftEntriesLimit, rightEntriesLimit] =
         //   getRankCover(pattEntries);
         // return harden([
-        //   fromEntries(leftEntriesLimit),
-        //   fromEntries(rightEntriesLimit),
+        //   fromUniqueEntries(leftEntriesLimit),
+        //   fromUniqueEntries(rightEntriesLimit),
         // ]);
         break;
       }
@@ -768,6 +772,16 @@ const makePatternKit = () => {
           false,
           X`${specimen} - no pattern disjuncts to match: ${patts}`,
         );
+      }
+      if (
+        patts.length === 2 &&
+        !matches(specimen, patts[0]) &&
+        checkKind(patts[0], 'match:kind', identChecker) &&
+        patts[0].payload === 'undefined'
+      ) {
+        // Worth special casing the optional pattern for
+        // better error messages.
+        return checkMatches(specimen, patts[1], check);
       }
       if (patts.some(patt => matches(specimen, patt))) {
         return true;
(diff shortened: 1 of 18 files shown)


packages/store/src/patterns/patternMatchers.js
```diff
@@ -1317,133 +1332,314 @@ const makePatternKit = () => {
...
+          checkMatches(optionalSpecimen[i], p, check, `arg ${argNum++}?`),
+        ) &&
+        checkMatches(restSpecimen, restPatt, check, '...rest')
+      );
+    },
+
+    compress: (
+      specimen,
+      [requiredPatt, optionalPatt = [], restPatt = MM.any()],
+      compress,
```
turadg: why does `compress` fn take another `compress` fn as an argument?

packages/store/src/patterns/patternMatchers.js
```diff
@@ -1317,133 +1332,314 @@ const makePatternKit = () => {
...
+   * compression distinguishing `undefined` from absence.
+   *
+   * @param {Pattern[]} optionalPatt
+   * @param {number} length
+   * @returns {Pattern[]} The partialPatt
+   */
+  const adaptArrayPattern = (optionalPatt, length) =>
+    harden(optionalPatt.slice(0, length).map(patt => MM.opt(patt)));
+
   /** @type {MatchHelper} */
```
turadg, line 1368: Looks like you've added two methods to MatchHelper but the doc hasn't been updated. Please document what `compress` and `decompress` are for, and whether they're optional on the type or not. There are at least 24 uses of the type and it looks like only this one has these new methods defined.

turadg (changes requested):
> just to provide the new feature. Because it is compat, this PR does not itself refactor the usage to use the feature.

This PR does change about a dozen test files outside `store` package so it looks like there are some changes here to adopt the feature. Maybe in `defendSyncArgs` using the new `M.splitArray`. I would have expected that change to be saved for https://github.com/Agoric/agoric-sdk/pull/6431/ .

That's not a blocker but I am requesting the documentation updates on `MatchHelper`

packages/store/src/patterns/patternMatchers.js
erights: Sorry, the `compress` and `decompress` methods were accidentally inherited from #6432 when I also moved some of the error message generation improvements from there.

To answer the question though, it is for recursive compression without creating a large static dependency cycle.

In any case, the `compress` and `decompress` methods, which are dead code in the context of this PR, are now removed. So resolving.

packages/store/src/patterns/patternMatchers.js
erights: https://github.com/Agoric/agoric-sdk/pull/6597#discussion_r1034784458 `compress` and `decompress` were dead in this PR and are now removed. Thanks for catching! Resolving.

packages/store/src/patterns/patternMatchers.js
```diff
@@ -1317,133 +1332,226 @@ const makePatternKit = () => {
...
+   * @param {CopyRecord<Pattern>} optionalPatt
+   * @returns {{
+   *   requiredSpecimen: CopyRecord<Passable>,
+   *   optionalSpecimen: CopyRecord<Passable>,
+   *   restSpecimen: CopyRecord<Passable>
+   * }}
+   */
+  const splitRecordParts = (specimen, requiredPatt, optionalPatt) => {
+    // Not frozen! Mutated in place
+    /** @type {[string, Passable][]} */
```
turadg, line 1455: nit, this style is little clearer for arrays of tuples:
```suggestion
    /** @type {Array<[string, Passable]>} */
```

packages/store/src/patterns/patternMatchers.js
```diff
@@ -1630,13 +1738,49 @@ const makePatternKit = () => {
...
+            ? [base]
+            : [base, [], rest]
+          : rest === undefined
+          ? [base, optional]
+          : [base, optional, rest],
+      ),
+    splitRecord: (base, optional = undefined, rest = undefined) =>
+      makeMatcher(
+        'match:splitRecord',
+        // eslint-disable-next-line no-nested-ternary
```
turadg: consider a helper function with conditional returns instead of this nested ternary

turadg (approved):
I see request was satisfied by a "fix" commit. That shouldn't go into master so when merging please use "squash". Or squash locally when you fix the lint error.

packages/store/src/patterns/patternMatchers.js
erights: Not sure this is good to do as a one-off in this PR. Between agoric-sdk and endo, we currently have 45 occurrences of "][]" which, by an informal skim, seem to be all or almost all types. OTOH we have 21 occurrences of "Array<[".

If we decide to canonicalize this element of style, we should coordinate across all the occurrences we wish to change.

turadg: Thanks for looking. To be clear, I wasn't requesting this change. By "nit" I meant to just suggest. (22:45 instead of 21:46). I've had a lint rule before to enforce the `Array<>` for tuples. Since we're on the topic, would you support such a rule?

erights: What does "22:45" or "21:46" mean?

We have not discussed, so this opinion is offered in ignorance of contrary arguments: I would not support. For arrays and tuples, I prefer the square bracket notation. I find it clearer as well as more concise.

packages/store/src/patterns/patternMatchers.js

packages/store/src/patterns/patternMatchers.js
turadg: > What does "22:45" or "21:46" mean?

Shorthand for ratios of instances of the two styles. (You said the current ratio is 21:45 so in this thread we've been talking about which one to increment.)

PR #6599, opened 2022-11-28
test: Add Node.js 18.x to workflows

sigv:
refs: #6489

## Description

Node.js 14.x is retained as the default, just as it was before, considering it is Maintenance LTS set to EOL next year (2023-04-30).

This change adds Node.js 18.x in all the places where workflows previously were building with Node.js 16.x - either as addition to it where multiple version are provided, or as an upgrade where only one version is used.

This finishes up the bulk of #6489, however manual testing of compatibility should be conducted further before that.

### Security Considerations

N/A. This is a change to test scenarios, to support the Current LTS of Node.js.

Backwards compatibility should not be affected. As result of manual testing in #6489, it may be decided to inform users that Node.js 18.x is now officially supported. Upgrade of Node.js versions is breaking for compiled blobs, in which case users need to rebuild project when upgrading from Node.js 16.x to 18.x, similar to how they would do when upgrading from 14.x to 16.x.

### Testing Considerations

The test suite is expanded to support Node.js 18.x. Additional manual testing is out of scope for this PR and can be evaluated/discussed in #6489.

### DIFF
--- a/.github/workflows/deployment-test.yml
+++ b/.github/workflows/deployment-test.yml
@@ -38,7 +38,7 @@ jobs:
           go-version: 1.17
       - uses: ./.github/actions/restore-node
         with:
-          node-version: 18.x
+          node-version: 16.x
           # Forces xsnap to initialize all memory to random data, which increases
           # the chances the content of snapshots may deviate between validators
           # TODO: key the node built cache on this env
--- a/.github/workflows/deployment-test.yml
+++ b/.github/workflows/deployment-test.yml
@@ -38,7 +38,7 @@ jobs:
           go-version: 1.17
       - uses: ./.github/actions/restore-node
         with:
-          node-version: 16.x
+          node-version: 18.x
           # Forces xsnap to initialize all memory to random data, which increases
           # the chances the content of snapshots may deviate between validators
           # TODO: key the node built cache on this env
--- a/packages/solo/test/captp-fixture.js
+++ b/packages/solo/test/captp-fixture.js
@@ -44,7 +44,7 @@ export async function makeFixture(PORT, noisy = false) {
       /** @type {() => void} */
       let abortCapTP;
       ws = new WebSocket(
-        `ws://localhost:${PORT}/private/captp?accessToken=${accessToken}`,
+        `ws://127.0.0.1:${PORT}/private/captp?accessToken=${accessToken}`,
         {
           origin: `http://localhost:${PORT}`,
         },
--- a/.github/workflows/after-merge.yml
+++ b/.github/workflows/after-merge.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['14.x', '16.x']
+        node-version: ['14.x', '16.x', '18.x']
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/test-all-packages.yml
+++ b/.github/workflows/test-all-packages.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['14.x', '16.x']
+        node-version: ['14.x', '16.x', '18.x']
     steps:
       - uses: actions/checkout@v2
         with:
@@ -35,7 +35,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/restore-node
         with:
-          node-version: '16.x'
+          node-version: '18.x'
 
       # first job also does repo-level linting
       - name: lint repo format
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/restore-node
         with:
-          node-version: '16.x'
+          node-version: '18.x'
 
       - name: yarn lint rest
         run: ./scripts/lint-with-types.sh rest
@@ -64,7 +64,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -132,7 +132,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -185,7 +185,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -207,7 +207,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -232,7 +232,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -253,7 +253,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -280,7 +280,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -301,7 +301,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -323,7 +323,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars
@@ -347,7 +347,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        engine: ['14.x', '16.x', 'xs']
+        engine: ['14.x', '16.x', '18.x', 'xs']
     steps:
       - name: set vars
         id: vars


sigv:
Be advised, Mergify should be adjusted to pick up on 18.x tests, and they should be also marked as "Expected"/"Required".

sigv:
[`packages/solo` `yarn test`](https://github.com/Agoric/agoric-sdk/actions/runs/3566790896/jobs/5993766480) invocation seems to have timed out after a TypeError. 🤔

sigv:
Running locally with Node.js v18.12.1, I seem to get test execution in `packages/solo`:

```plain
yarn run v1.22.19
$ ava

# connecting
Loading slog sender modules: @agoric/telemetry/src/flight-recorder.js
Loading slog sender modules: @agoric/telemetry/src/flight-recorder.js
2022-11-28T17:43:41.521Z launch-chain: Expected SwingSet kernel statistic cosmic_swingset_inbound_queue_length not found
2022-11-28T17:43:41.522Z launch-chain: Expected SwingSet kernel statistic cosmic_swingset_inbound_queue_add not found
2022-11-28T17:43:41.522Z launch-chain: Expected SwingSet kernel statistic cosmic_swingset_inbound_queue_remove not found
agoric: deploy: running /root/agoric-sdk/packages/wallet/api/deploy.js

SES_UNHANDLED_REJECTION: (TypeError#1)
TypeError#1: Cannot deliver getBridge to target; typeof target is undefined

CapTP test fixture exception: (RemoteError(error:captp:unknown#20001)#1)
RemoteError(error:captp:unknown#20001)#1: board does not have id: (a string)
  at WebSocket.<anonymous> (packages/solo/test/captp-fixture.js:60:11)
  at WebSocket.emit (node:events:513:28)
  at Receiver.emit (node:events:513:28)
  at Socket.emit (node:events:513:28)

CapTP test fixture exception: (RemoteRangeError(error:captp:unknown#20002)#2)
RemoteRangeError(error:captp:unknown#20002)#2: id is probably a typo, cannot verify CRC: (a string)
  at WebSocket.<anonymous> (packages/solo/test/captp-fixture.js:60:11)
  at WebSocket.emit (node:events:513:28)
  at Receiver.emit (node:events:513:28)
  at Socket.emit (node:events:513:28)

  ✔ home.board (7.7s)
  ✔ home.wallet - transfer funds to the feePurse (4.4s)
  ✔ home.wallet - receive zoe invite (17.4s)
    ℹ {
        contractRoot: '/root/agoric-sdk/packages/zoe/src/contracts/automaticRefund.js',
      }
  ✔ home.wallet - central issuer setup (2.6s)
  ✔ home.localTimerService makeNotifier (1.1s)
  ✔ home.localTimerService makeRepeater (895ms)
2022-11-28T17:47:58.091Z shutdown: Shutting down cleanly...
2022-11-28T17:47:58.092Z shutdown: Process terminated!
  ─

  6 tests passed
Done in 427.42s.
```

sigv:
Will try to understand what could be different between the CI worker and my local installation, resulting in `packages/solo` hanging..

sigv:
`EBADF: bad file descriptor, read` during [build (14.x)](https://github.com/Agoric/agoric-sdk/actions/runs/3575824369/jobs/6012857144) when rebasing on top of latest `master` patches. Still haven't found a way to repro this locally.

sigv:
It looks like `test-home` invoking the CLI does not return back to the Ava testcases. Must not believe the chain is ready for accepting work.

sigv:
Bug in `@endo/captp` doesn't feel likely.. but maybe something with `makeCapTP` in particular? Still doesn't explain why this would cause an issue on the CI and not locally.

sigv:
Could someone check that the runner `ubuntu-latest` actually has all packages up-to-date?
My local tests are on a fully updated Ubuntu 22.04.1 LTS (Jammy).

mhofman:
I can't look at this this week, maybe @michaelfig can lend a hand?

sigv:
Running that test on a clean Ubuntu 20.04 (Focal) image deployed in Azure on D8s_v5 (8 vCPUs, 32G RAM) seems to work. This feels more and more like a runner-specific issue, but I am not sure why this one package's test would be affected. 😐 

Here is the minimal scenario:

```bash
curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/nodesource.gpg >/dev/null
echo "deb [signed-by=/etc/apt/trusted.gpg.d/nodesource.gpg] https://deb.nodesource.com/node_18.x $(lsb_release -s -c) main" | sudo tee /etc/apt/sources.list.d/nodesource.list >/dev/null

sudo apt update
sudo apt --yes dist-upgrade
sudo apt --yes install chromium-browser gcc g++ jq make nodejs perl

sudo corepack enable yarn

sudo wget -O /usr/local/go1.19.3.linux-amd64.tar.gz https://go.dev/dl/go1.19.3.linux-amd64.tar.gz
sudo tar -C /usr/local -xf /usr/local/go1.19.3.linux-amd64.tar.gz
eval $(echo 'export PATH=$PATH:/usr/local/go/bin' | sudo tee /etc/profile.d/golang.sh)

go version
# go version go1.19.3 linux/amd64
node --version
# v18.12.1
yarn --version
# 1.22.19

git clone -b maint/test-node-18.x https://github.com/sigv/agoric-sdk.git
cd agoric-sdk

yarn install
yarn build

cd packages/solo
yarn test

# 6 tests passed
# Done in 218.96s.
```

sigv:
Slowly learning more about how this is interconnected. Right now it seems the Websocket connection in [packages/solo/test/captp-fixture.js](https://github.com/Agoric/agoric-sdk/blob/0b5df16f83629efb7cb48d54250139e082ed109c/packages/solo/test/captp-fixture.js#L92) keep failing and looping - resulting in the observable mix of dots (`.`) and some `o`s. On Node 14.x and 16.x it's a consistent line of `o`s - no waiting on connection.

[Latest run](https://github.com/Agoric/agoric-sdk/actions/runs/3585408992/jobs/6033453540) does imply that the service itself comes online - as signaled by "Deployed Wallet!" - just that the test scenario never kicks off.

---

I also noticed that on [Node 14.x](https://github.com/Agoric/agoric-sdk/actions/runs/3585408992/jobs/6033453160) the log output says
```plain
CapTP test fixture exception: (RemoteError(error:captp:http://localhost:7999#20001)#1)
RemoteError(error:captp:http://localhost:7999#20001)#1: board does not have id: (a string)
  at WebSocket.<anonymous> (packages/solo/test/captp-fixture.js:60:11)
  at WebSocket.emit (events.js:400:28)
  at Receiver.emit (events.js:400:28)
  at Socket.emit (events.js:400:28)
```

however, on [Node 16.x](https://github.com/Agoric/agoric-sdk/actions/runs/3585408992/jobs/6033453318) it does not list the origin and instead says "unknown" as
```plain
CapTP test fixture exception: (RemoteError(error:captp:unknown#20001)#1)
RemoteError(error:captp:unknown#20001)#1: board does not have id: (a string)
  at WebSocket.<anonymous> (packages/solo/test/captp-fixture.js:60:11)
  at WebSocket.emit (node:events:513:28)
  at Receiver.emit (node:events:513:28)
  at Socket.emit (node:events:513:28)
```

I am locally observing "unknown" on Node 18.x as well, but just seemed a bit strange.

sigv:
Node.js 17.0.0 enables native DNS result priority, which can result in IPv6 being preferred in many scenarios. I wonder if the IPv6 stack is the issue for the CI runner!

Node.js 16.x tests report `Open CapTP connection to ws://127.0.0.1:7999/private/captp...`
while Node.js 18.x tests fail with `connect ECONNREFUSED ::1:7999`

I did not enable IPv6 stack on my test VMs, and when testing locally the Linux box was only hooked up to IPv4 too. The test runner must have a native IPv6 address and prefers to use it then!

sigv:
Sorry for the messy investigation and various rebases.

@mhofman, @michaelfig, this is ready for review and potential merge.

packages/solo/test/captp-fixture.js
```diff
@@ -44,7 +44,7 @@ export async function makeFixture(PORT, noisy = false) {
       /** @type {() => void} */
       let abortCapTP;
       ws = new WebSocket(
-        `ws://localhost:${PORT}/private/captp?accessToken=${accessToken}`,
+        `ws://127.0.0.1:${PORT}/private/captp?accessToken=${accessToken}`,
```
mhofman, line 47: Interesting! I see node now prefers IPv6, but I'm still wondering why that wouldn't work. I guess it doesn't fallback to IPv4 if the server doesn't answer and for some reason we're not listening on IPv6?

.github/workflows/deployment-test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           go-version: 1.17
       - uses: ./.github/actions/restore-node
         with:
-          node-version: 16.x
+          node-version: 18.x
```
mhofman: I'll force integration to test if 18 works for the deployment test, but we may want to not poke the bear and stick with 16 here for now.

mhofman: Yeah looks like the version of node 18 available on GH actions is incompatible with the ancient version of ubuntu we're still using for the deployment test. I think we have an issue somewhere about upgrading the ubuntu version, or at least GH action is getting insistent, so maybe we can take care of switching to node 18 then.

mhofman:
```suggestion
          node-version: 16.x
```

mhofman (changes requested):
Thanks for investigative work you did on this PR. Once we revert to 16 on the deployment test I believe this can be merged.

Please also [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) for your branch so mergify can rebase as necessary before merging.

packages/solo/test/captp-fixture.js
sigv: Indeed - there is no way to fall back to `127.0.0.1` when it's trying to use `::1`. It fails the connection and loops back around trying `::1` again.

.github/workflows/deployment-test.yml
sigv: Ack.

packages/solo/test/captp-fixture.js
sigv: As for not listening on IPv6 - Generally requesting listen and only providing port enables all IPv4 and IPv6. I did not investigate further, but I would assume that listen here happens explicitly on `127.0.0.1` and that in turn enables only IPv4.

mhofman: Yeah afaik, there is no way to open a single listen socket for both IPv4 and IPv6 restricted to the respective loopback addresses. I believe you can bind to a specific interface without address, but that's very system dependent and I doubt node exposes that. Alternative is to open 2 listen sockets. I suppose another way may be to listen providing `localhost` and hope both the server and client use the same resolution strategies.

Really [Node should implement Happy Eyeballs](https://github.com/nodejs/node/issues/41625), exactly for these kind of cases.

mhofman (approved):

sigv:
@mhofman, don't forget about marking the 18.x tests as "Expected"/"Required" so they are visible clearly in the GitHub UI, as well as adding them to Mergify checks!

sigv:
And big thanks for the review! 🎉

mhofman:
> don't forget about marking the 18.x tests as "Expected"/"Required"

Yup, planning on making them required, and removing the 14 ones from required

mhofman:
> And big thanks for the review! 🎉

Thanks so much for your contribution.

PR #6567, opened 2022-11-15
chore: Sync Endo

kriskowal:
This change generated by mechanically following the instructions in `MAINTAINERS.md`.

- chore: Sync Endo versions
- chore: Sync Jessie versions
- chore: Carry import-bundle patch forward to v0.3.0
- chore: Carry ses-ava patch forward to v0.2.36
- chore: Update yarn.lock
- chore: Ramp the meter version
- chore(swingset-vat): Update xsnap store test snapshots

PR #6602, opened 2022-11-28
fix(swingset-runner): Relax hard dep on canvas

kriskowal:
- fix(swingset-runner): Revert hard dep on canvas
- chore: Update yarn.lock

### DIFF
--- a/packages/stat-logger/package.json
+++ b/packages/stat-logger/package.json
@@ -19,6 +19,17 @@
     "vega": "^5.10.0",
     "vega-canvas": "^1.2.1"
   },
+  "peerDependenciesMeta": {
+    "canvas": {
+      "optional": true
+    },
+    "vega": {
+      "optional": true
+    },
+    "vega-canvas": {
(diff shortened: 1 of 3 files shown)


turadg (approved):

kriskowal:
This turned out not to be viable because of an underlying change to the treatment in peer dependencies that came in with the last Endo sync #6567. The underlying change to peer dependencies in Endo was https://github.com/endojs/endo/pull/1355. We’ll need to do something different.

kriskowal:
@kumavis gave us a tip that there’s a `peerDependenciesMeta` with an `optional` field that addresses this problem and implemented with the fix for `peerDependencies` in Endo. Giving that a spin in CI now.

kriskowal:
cc @michaelfig adding peerDependenciesMeta optional: true properties alongside the peerDependencies allowed this to work with Endo. This is not an Endo precedent, but something @kumavis brought in from Browserify.

PR #6603, opened 2022-11-29
6384 casting queryAbci

### LINKED ISSUE #6384: get latest RPC data in a single HTTP request

turadg:
closes: #6384

## Description

Whenever fetching data at the latest height, Casting was first making a query for height. In some loops it was repeated.

This uses Tendermint RPC's `abciQuery` to fetch data at the same time as height. To do this, many of the helper functions had to be refactored to return the height along with the data. It no longer uses the Stargate abstraction, but to align with it the `QueryAbciResponse` type is copied and used.

### Testing Considerations

I tested some agoric-cli commands which use Casting.

### DIFF
--- a/packages/casting/src/follower-cosmjs.js
+++ b/packages/casting/src/follower-cosmjs.js
@@ -2,19 +2,28 @@
 /* eslint-disable no-await-in-loop, no-continue, @jessie.js/no-nested-await */
 
 import { E, Far } from '@endo/far';
-import * as tendermintRpcStar from '@cosmjs/tendermint-rpc';
-import * as stargateStar from '@cosmjs/stargate';
+import * as tendermint34 from '@cosmjs/tendermint-rpc';
 
 import { MAKE_DEFAULT_DECODER, MAKE_DEFAULT_UNSERIALIZER } from './defaults.js';
 import { makeCastingSpec } from './casting-spec.js';
 import { makeLeader as defaultMakeLeader } from './leader-netconfig.js';
 
-const { QueryClient } = stargateStar;
-const { Tendermint34Client } = tendermintRpcStar;
+const { Tendermint34Client } = tendermint34;
 const { details: X, quote: q } = assert;
 const textDecoder = new TextDecoder();
 
 /** @template T @typedef {import('./types.js').Follower<import('./types.js').ValueFollowerElement<T>>} ValueFollower */
+
+// Copied from https://github.com/cosmos/cosmjs/pull/1328/files until release
+/**
+ * @typedef {{
+ *   readonly value: Uint8Array;
+ *   readonly height: number;
+ * }} QueryAbciResponse
+ * The response of an ABCI query to Tendermint.
+ * This is a subset of `tendermint34.AbciQueryResponse` in order
+ * to abstract away Tendermint versions.
+ */
 
 /**
  * This is an imperfect heuristic to navigate the migration from value cells to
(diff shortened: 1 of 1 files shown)


packages/casting/src/follower-cosmjs.js
```diff
@@ -210,71 +201,64 @@ export const makeCosmjsFollower = (
...
   };
 
   /**
-   * @param {number} height
+   * @param {number} [height] desired height, or the latest height if not set
+   * @param {boolean} [prove=false]
+   * @returns {Promise<QueryAbciResponse>}
    */
-  const getUnprovenDataAtHeight = async height => {
+  const getAvailableData = async (height, prove = false) => {
```
kriskowal, line 213: I recommend using an options bag here to avoid the [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)

packages/casting/src/follower-cosmjs.js
```diff
@@ -403,53 +387,43 @@ export const makeCosmjsFollower = (
...
-    let data;
+    let lastData;
     for (;;) {
-      const currentBlockHeight = await getBlockHeight();
-      if (currentBlockHeight === blockHeight) {
-        // TODO Long-poll for next block
-        // https://github.com/Agoric/agoric-sdk/issues/6154
-        await E(leader).jitter(where);
-        continue;
-      }
```
kriskowal, line 415: This check is intended to prevent the iterable from stuttering duplicate values if the consumer side polls faster than the chain advances. Without it, we fall thru to the check below it that discards duplicates if the data are bytewise identical, which should have the same effect but leans on a code path that we might remove if we can ever drop support for legacy cells. We should at least update the comment on line 407 which is no longer true.

packages/casting/src/follower-cosmjs.js
```diff
@@ -210,71 +201,64 @@ export const makeCosmjsFollower = (
...
+  const getAvailableData = async (height, prove = false) => {
     return retryGetDataAndStripPrefix(
       async (endpoint, storeName, storeSubkey) => {
         const client = await provideTendermintClient(endpoint);
         const response = await client.abciQuery({
           path: `store/${storeName}/key`,
           data: storeSubkey,
           height,
-          prove: false,
+          prove,
```
kriskowal, line 221: I don’t think `abciQuery` checks the proofs, just includes the data necessary to verify the proofs in the response. I suspect we still need Stargate client to do the math on the proofs to verify them. But, this touches on stuff @michaelfig is far more knowledgeable about. https://github.com/cosmos/cosmjs/blob/74346b2ce6/packages/tendermint-rpc/src/tendermint34/requests.ts#L72-L78

turadg: Ah, indeed. https://github.com/cosmos/cosmjs/blob/74346b2ce6/packages/stargate/src/queryclient/queryclient.ts#L508

I'll restore that. I wonder if some regression test is feasible.

packages/casting/src/follower-cosmjs.js
turadg: Ah. PTAL at https://github.com/Agoric/agoric-sdk/pull/6607/commits/49e8fddb7da6ebb56dd39eaaec12a4e646becc65

PR #6607, opened 2022-11-29
6384 single fetch for data and block height

### LINKED ISSUE #6384: get latest RPC data in a single HTTP request

turadg:
closes: #6384

A fresh try at https://github.com/Agoric/agoric-sdk/pull/6603 with fewer changes.

## Description

Whenever fetching data at the latest height, Casting was first making a query for height. In some loops it was repeated.

This uses Tendermint RPC's `abciQuery` to fetch data at the same time as height. To do this, many of the helper functions had to be refactored to return the height along with the data (using the `QueryAbciResponse` type from  https://github.com/cosmos/cosmjs/pull/1328/ )

### Testing Considerations

I tested some agoric-cli commands which use Casting.

### DIFF
--- a/packages/casting/src/follower-cosmjs.js
+++ b/packages/casting/src/follower-cosmjs.js
@@ -476,11 +476,10 @@ export const makeCosmjsFollower = (
   }
 
   /**
-   * @param {number} cursorBlockHeight
+   * @param {number} [cursorBlockHeight]
    * @yields {ValueFollowerElement<T>}
    */
   async function* getEachIterableAtHeight(cursorBlockHeight) {
-    assert.typeof(cursorBlockHeight, 'number');
     // Track the data for the last emitted cell (the cell at the
     // cursorBlockHeight) so we know not to emit duplicates
     // of that cell.
@@ -490,7 +489,9 @@ export const makeCosmjsFollower = (
     // If the block has no corresponding data, wait for the first block to
     // contain data.
     for (;;) {
-      cursorData = (await getDataAtHeight(cursorBlockHeight)).value;
+      ({ value: cursorData, height: cursorBlockHeight } = await getDataAtHeight(
+        cursorBlockHeight,
+      ));
       if (cursorData.length !== 0) {
         const cursorStreamCell = streamCellForData(
           cursorBlockHeight,
@@ -502,7 +503,6 @@ export const makeCosmjsFollower = (
       // TODO Long-poll for next block
       // https://github.com/Agoric/agoric-sdk/issues/6154
       await E(leader).jitter(where);
-      cursorBlockHeight = await getBlockHeight();
     }
 
     // For each subsequent iteration, yield every value that has been
@@ -624,9 +624,6 @@ export const makeCosmjsFollower = (
       return getLatestIterable();
     },
     async getEachIterable({ height = undefined } = {}) {
-      if (height === undefined) {
-        height = await getBlockHeight();
-      }
       return getEachIterableAtHeight(height);
     },
     async getReverseIterable({ height = undefined } = {}) {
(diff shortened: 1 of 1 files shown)


packages/casting/src/follower-cosmjs.js
```diff
@@ -196,85 +224,89 @@ export const makeCosmjsFollower = (
...
   /**
-   * @param {number} blockHeight
+   * @param {number} [blockHeight] desired height, or the latest height if not set
+   * @returns {Promise<QueryAbciResponse>}
    */
   const tryGetDataAtHeight = async blockHeight => {
     if (proof === 'strict') {
+      // we have to know this in order to construct a valid return
+      if (blockHeight === undefined) {
+        blockHeight = await getBlockHeight();
```
turadg, line 287: for the `strict` case, this is repeating the `getBlockHeight()` more than before. I figure this all is good factoring though, because there should be a way to get the data _and height_ even for proven data.

I don't know what the performance impact of this is in practice but there are other ways to work around the Stargate deficiency with less perf regression. 

1. have `if (proof === 'strict')` to get the block height at higher levels in the call stack
2. memoize `getBlockHeight()` with an expiration of expected block time (6s)

kriskowal: I believe this is where we’ll refactor to take advantage of https://github.com/cosmos/cosmjs/pull/1328. Maybe leave a TODO link. I’m fine with this factoring until then.

packages/casting/src/follower-cosmjs.js
```diff
@@ -196,85 +224,89 @@ export const makeCosmjsFollower = (
...
       return getUnprovenDataAtHeight(blockHeight);
     } else if (proof === 'optimistic') {
-      const allegedData = await getUnprovenDataAtHeight(blockHeight);
+      const alleged = await getUnprovenDataAtHeight(blockHeight);
 
       // Prove later, since it may take time we say we can't afford.
-      getProvenDataAtHeight(blockHeight).then(provenData => {
-        if (arrayEqual(provenData, allegedData)) {
+      getProvenDataAtHeight(alleged.height).then(proven => {
+        if (arrayEqual(proven.value, alleged.value)) {
```
kriskowal, line 299: Reads well. Would read again. 13/10

packages/casting/src/follower-cosmjs.js
```diff
@@ -403,53 +434,50 @@ export const makeCosmjsFollower = (
...
         continue;
       }
-      const currentStreamCell = streamCellForData(
-        currentBlockHeight,
-        currentData,
-      );
+      const currentStreamCell = streamCellForData(latest.height, latest.value);
 
-      blockHeight = currentBlockHeight;
+      lastHeight = latest.height;
```
kriskowal, line 460: Last and Latest maybe need a Highlander moment.

kriskowal (approved):
Decadent!

packages/casting/src/follower-cosmjs.js
turadg: That PR lacks the `queryVerified` case but https://github.com/cosmos/cosmjs/pull/1278#issuecomment-1326527195 indicates that was just step 1. I've linked to your PR as a place to find whether it's resolved.

packages/casting/src/follower-cosmjs.js
turadg: Good to raise. I think they're actually distinct. `latest` is what we know is the height of the chain and `last` is the last value we got for that. Sort of `lastLatestHeight`. I'll leave a comment about that.

PR #6609, opened 2022-11-30
4383 debugging fee mint access

Chris-Hibbert:
An INTERMEDIATE state for debugging

Most tests pass. Much is working about Zoe durability.

One major sticking point is that the feeMintAccess token isn't being
robustly delivered to contracts that need it.  It appears that the
loopback mechanism from endo may be at fault.

I've turned on tracing for object IDs as built in
https://github.com/Agoric/agoric-sdk/pull/6218
This makes it possible to identify individual objects as they are
passed around.

For example, when the  'Lower Collateral' test in test-stakeFactory is
run, we get

    The object representing access to the fee brand mint was not provided

And a detailed trace using #6218 shows that the correct object was
held in econ-behavior, but when the contractGovernor is called, it
appears to have a different object.

```
EB  p.all Object [Alleged: FeeMint feeMintAccess] {
  'o+15/1|feeMintAccess': [Function: o+15/1|feeMintAccess]
}
CG  START  Object [Alleged: FeeMint feeMintAccess] {}
FeeMint Object [Alleged: FeeMint feeMintAccess] {
  'o+15/1|feeMintAccess': [Function: o+15/1|feeMintAccess]
} Object [Alleged: FeeMint feeMintAccess] {}
Temporary logging of sent error (Error#1)
Error#1: The object representing access to the fee brand mint was not provided
```

replacing `const zoe = makeFar(zoeService);` with `const zoe =
zoeService;` allows the test to pass, so removing the loopback
circumvents the problem.
### Security Considerations

None

None

### Testing Considerations

this is about testing

### DIFF
--- a/packages/ERTP/src/typeGuards.js
+++ b/packages/ERTP/src/typeGuards.js
@@ -138,6 +138,13 @@ export const DisplayInfoShape = M.partial(
   }),
 );
 
+export const IssuerKitShape = harden({
+  brand: BrandShape,
+  mint: MintShape,
+  issuer: IssuerShape,
+  displayInfo: DisplayInfoShape,
+});
+
 // //////////////////////// Interfaces /////////////////////////////////////////
 
 export const BrandI = M.interface('Brand', {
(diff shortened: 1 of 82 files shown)


Chris-Hibbert:
No longer useful.

PR #6613, opened 2022-12-02
feat(assert): export assert.Fail from endo

erights:
Now that agoric-sdk has been upgraded to depend on https://github.com/endojs/endo/pull/1334 , the `@agoric/assert` module should export `Fail` directly. Also, types.js needs to track the addition of `Fail`. See https://github.com/endojs/endo/pull/1384


turadg (approved):
LGTM 👍

PR #6614, opened 2022-12-02
refactor(types): EconCharterStartResult without runtime

turadg:
## Description

This defines the `EconCharterStartResult` without executing code and requiring lint suppressions.

### Security Considerations

--

New type, has a doc string.

### Testing Considerations

CI

### DIFF
--- a/packages/zoe/src/zoeService/utils.d.ts
+++ b/packages/zoe/src/zoeService/utils.d.ts
@@ -1,6 +1,6 @@
 import { Callable } from '@agoric/eventual-send';
 
-import type { Instance, IssuerKeywordRecord, Payment } from './types.js';
+import type { IssuerKeywordRecord, Payment } from './types.js';
 
 // XXX https://github.com/Agoric/agoric-sdk/issues/4565
 type SourceBundle = Record<string, any>;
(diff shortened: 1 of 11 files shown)


packages/zoe/src/zoeService/utils.d.ts
```diff
@@ -65,6 +65,12 @@ type StartContractInstance<C> = (
   adminFacet: AdminFacet;
 }>;
 
+/** The result of `startInstance` */
+export type StartedInstanceKit<SF> = {
+  instance: Instance;
```
dckc, line 74: Have you looked into parameterizing the `Instance` type such that `E(zoe).getPublicFacet(instance)` would get a useful type?

dckc (approved):

packages/zoe/src/zoeService/utils.d.ts
turadg: I had but ran into some problem I don't remember. Thanks for the prompt; I got it into this PR.  PTAAL.

turadg: @dckc I had expected Github to not allow merge without your second review because I clicked re-request review and that cleared the checkmark. 🤷

PR #6606, opened 2022-11-29
prettier docs

turadg:
- test: no pruning in start-local-chain
- test: consistent WALLET env var
- docs: prettier formatting in docs/ dirs

PR #6615, opened 2022-12-02
update smoketest docs

turadg:
## Description

Less ambitious than https://github.com/Agoric/agoric-sdk/pull/6606


dckc (approved):
looks like an improvement

